PR #636, opened 2016-04-22
Added protection for total_size in set_piece_hashes

aldenml:
@arvidn I added the necessary assert, otherwise all the logic get corrupt, but not sure how to handle it. Waiting for your comment.

### DIFF
--- a/include/libtorrent/create_torrent.hpp
+++ b/include/libtorrent/create_torrent.hpp
@@ -34,12 +34,10 @@ POSSIBILITY OF SUCH DAMAGE.
 #define TORRENT_CREATE_TORRENT_HPP_INCLUDED
 
 #include "libtorrent/bencode.hpp"
-#include "libtorrent/peer_id.hpp"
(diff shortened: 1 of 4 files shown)


arvidn:
that seems reasonable

arvidn:
the error seems legitimate as far as I can tell, and unrelated to your change.

aldenml:
This seems to uncover a problem in a test: https://github.com/arvidn/libtorrent/blob/master/test/test_torrent.cpp#L254

aldenml:
I decided to reduce the PR to the minimal protection that it's necessary in `set_piece_hashes`.

arvidn:
this looks like a good change. I'll merge it once appveyor completes. I have a feeling this should go into  RC_1_0 as well. the qbt quys were seeing an assert that seems related to this.

However though, I think the first patch was also good (and important). and the test should be fixed

aldenml:
I will review the original idea, and after that, the backport

PR #638, opened 2016-04-23
fix move_storage error handling bug

arvidn:


arvidn:
@glassez does this look right?

glassez:
Yes.

PR #640, opened 2016-04-24
fix x64 build with VS2015 UP2 compiler

ajax16384:
fix release build with VS2015 UP2 compiler warnings for: 
- address-model=64
- deprecated-functions=on
- link=shared

### DIFF
--- a/Jamfile
+++ b/Jamfile
@@ -286,6 +286,10 @@ rule warnings ( properties * )
 		result += <cflags>/wd4244 ;
 # disable warning C4512: assignment operator could not be generated
 		result += <cflags>/wd4512 ;
+# disable warning C4251: 'identifier' : class 'type' needs to have dll-interface to be used by clients of class 'type2'
+		result += <cflags>/wd4251 ;
+# disable warning C4275: non – DLL-interface classkey 'identifier' used as base for DLL-interface classkey 'identifier'
+		result += <cflags>/wd4275 ;
 	}
 
 	return $(result) ;
--- a/Jamfile
+++ b/Jamfile
@@ -288,7 +288,7 @@ rule warnings ( properties * )
 		result += <cflags>/wd4512 ;
 # disable warning C4251: 'identifier' : class 'type' needs to have dll-interface to be used by clients of class 'type2'
 		result += <cflags>/wd4251 ;
-# disable warning C4275: non – DLL-interface classkey 'identifier' used as base for DLL-interface classkey 'identifier'
+# disable warning C4275: non DLL-interface classkey 'identifier' used as base for DLL-interface classkey 'identifier'
 		result += <cflags>/wd4275 ;
 	}
 
--- a/examples/Jamfile
+++ b/examples/Jamfile
@@ -13,6 +13,8 @@ project client_test
    : requirements
 	<threading>multi <library>/torrent//torrent
 	<toolset>darwin:<cflags>-Wno-unused-command-line-argument
+# disable warning C4275: non – DLL-interface classkey 'identifier' used as base for DLL-interface classkey 'identifier'
+	<toolset>msvc:<cflags>/wd4275
 	: default-build
 	<link>static
    ;  
--- a/examples/Jamfile
+++ b/examples/Jamfile
@@ -13,7 +13,7 @@ project client_test
    : requirements
 	<threading>multi <library>/torrent//torrent
 	<toolset>darwin:<cflags>-Wno-unused-command-line-argument
-# disable warning C4275: non – DLL-interface classkey 'identifier' used as base for DLL-interface classkey 'identifier'
+# disable warning C4275: non DLL-interface classkey 'identifier' used as base for DLL-interface classkey 'identifier'
 	<toolset>msvc:<cflags>/wd4275
 	: default-build
 	<link>static
--- a/include/libtorrent/debug.hpp
+++ b/include/libtorrent/debug.hpp
@@ -34,6 +34,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #define TORRENT_DEBUG_HPP_INCLUDED
 
 #include "libtorrent/config.hpp"
+#include "libtorrent/assert.hpp"
 
 #if TORRENT_USE_ASSERTS && defined BOOST_HAS_PTHREADS
 #include <pthread.h>
@@ -41,7 +42,6 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #if defined TORRENT_ASIO_DEBUGGING
 
-#include "libtorrent/assert.hpp"
 #include "libtorrent/thread.hpp"
 #include "libtorrent/time.hpp"
 
@@ -169,8 +169,8 @@ namespace libtorrent
 
 #else
 
-#define ADD_OUTSTANDING_ASYNC(x) do {} while(0)
-#define COMPLETE_ASYNC(x) do {} while(0)
+#define ADD_OUTSTANDING_ASYNC(x) do {} TORRENT_WHILE_0
+#define COMPLETE_ASYNC(x) do {} TORRENT_WHILE_0
 
 #endif // TORRENT_ASIO_DEBUGGING
 
--- a/include/libtorrent/utp_socket_manager.hpp
+++ b/include/libtorrent/utp_socket_manager.hpp
@@ -90,7 +90,7 @@ namespace libtorrent
 
 		void mtu_for_dest(address const& addr, int& link_mtu, int& utp_mtu);
 		void set_sock_buf(int size);
-		int num_sockets() const { return m_utp_sockets.size(); }
+		int num_sockets() const { return int(m_utp_sockets.size()); }
 
 		void defer_ack(utp_socket_impl* s);
 		void subscribe_drained(utp_socket_impl* s);
--- a/simulation/Jamfile
+++ b/simulation/Jamfile
@@ -18,8 +18,6 @@ project
 	<source>utils.cpp
 	: default-build
 	<threading>multi
-# disable warning C4275: non DLL-interface classkey 'identifier' used as base for DLL-interface classkey 'identifier'
-	<toolset>msvc:<cflags>/wd4275
 	<invariant-checks>full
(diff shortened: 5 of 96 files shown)


Jamfile
```diff
@@ -286,6 +286,10 @@ rule warnings ( properties * )
 		result += <cflags>/wd4244 ;
 # disable warning C4512: assignment operator could not be generated
 		result += <cflags>/wd4512 ;
+# disable warning C4251: 'identifier' : class 'type' needs to have dll-interface to be used by clients of class 'type2'
+		result += <cflags>/wd4251 ;
+# disable warning C4275: non � DLL-interface classkey 'identifier' used as base for DLL-interface classkey 'identifier'
```
arvidn: what's this character? it seems like it should probably not be in here..

src/enum_net.cpp
```diff
@@ -374,7 +374,7 @@ namespace libtorrent
 			typedef boost::asio::ip::address_v4::bytes_type bytes_t;
 			bytes_t b;
 			std::memset(&b[0], 0xff, b.size());
-			for (int i = sizeof(bytes_t)/8-1; i > 0; --i)
+			for (int i = int(sizeof(bytes_t)) / 8 - 1; i > 0; --i)
```
arvidn: there's a more-or-less identical loop below, for when IPv6 is enabled. presumably it would need the same treatment

src/file.cpp
```diff
@@ -197,7 +197,7 @@ namespace
 		for (int i = 0; i < num_bufs; ++i)
 		{
 			DWORD num_read;
-			if (ReadFile(fd, bufs[i].iov_base, bufs[i].iov_len, &num_read, &ol[i]) == FALSE
+			if (ReadFile(fd, bufs[i].iov_base, DWORD(bufs[i].iov_len), &num_read, &ol[i]) == FALSE
```
arvidn: do you think it would be possible to define the iovec_t member iov_len as a DWORD? maybe it would save some of these casts

Jamfile
ajax16384: ok

src/http_seed_connection.cpp
```diff
@@ -218,7 +218,7 @@ namespace libtorrent
 
 		if (error)
 		{
-			received_bytes(0, bytes_transferred);
+			received_bytes(0, int(bytes_transferred));
```
arvidn: maybe received_bytes() should really take a size_t as well. it seems like that may also cut down on the casting, what do you think?

src/mpi.c
```diff
@@ -2650,7 +2650,7 @@ int mp_exptmod_fast (mp_int * G, mp_int * X, mp_int * P, mp_int * Y, int redmode
 {
   mp_int  M[TAB_SIZE], res;
   mp_digit buf, mp;
-  int     err, bitbuf, bitcpy, bitcnt, mode, digidx, x, y, winsize;
+  int     err, bitbuf, bitcpy, bitcnt, mode, digidx, x, y, winsize, winsize_shift, winsize_minus_shift;
```
arvidn: this file comes from libtommath. I would be more comfortable disabling warnings for this file rather than mutating it. in fact, it may even be due for an update.

ajax16384: ok, i will revert changes and inser msver coditional pragma at file header

test/Jamfile
```diff
@@ -73,6 +73,8 @@ lib libtorrent_test
 	<target-os>windows:<library>advapi32
 	<conditional>@link_libtorrent
 	<toolset>darwin:<cflags>-Wno-unused-command-line-argument
+# disable warning C4275: non � DLL-interface classkey 'identifier' used as base for DLL-interface classkey 'identifier'
```
arvidn: these non-ascii characters should probably be removed too

test/make_torrent.cpp
```diff
@@ -69,7 +69,7 @@ boost::shared_ptr<libtorrent::torrent_info> make_test_torrent(
 		std::string name = "test_file-1";
 		if (ent.find("name=") != std::string::npos)
 		{
-			int pos = ent.find("name=") + 5;
+			int pos = int(ent.find("name=") + 5);
```
arvidn: could pos be std::string::size_type instead?

test/make_torrent.cpp
```diff
@@ -103,7 +103,7 @@ boost::shared_ptr<libtorrent::torrent_info> make_test_torrent(
 			std::string name = filename;
 			if (ent.find("name=") != std::string::npos)
 			{
-				int pos = ent.find("name=") + 5;
+				int pos = int(ent.find("name=") + 5);
```
arvidn: and here

test/test_resolve_links.cpp
```diff
@@ -108,8 +108,8 @@ TORRENT_TEST(resolve_links)
 
 		std::vector<resolve_links::link_t> const& links = l.get_links();
 
-		int num_matches = std::count_if(links.begin(), links.end()
-			, boost::bind(&resolve_links::link_t::ti, _1));
+		int num_matches = int(std::count_if(links.begin(), links.end()
```
arvidn: it may be simpler to make this a std::size_t too

tools/Jamfile
```diff
@@ -28,6 +28,8 @@ rule link_libtorrent ( properties * )
 project tools
    : requirements
 	<threading>multi
+# disable warning C4275: non � DLL-interface classkey 'identifier' used as base for DLL-interface classkey 'identifier'
```
arvidn: non-ascii here too

src/file.cpp
ajax16384: iovec_t should remain compatible with *nix iovec. Where are plenty piece of code that relies on fact that iov_len is size_t.
````c++
file::iovec_t b = { random_data, size_t(16000)};
...
file::iovec_t b = { (void*)data, size_t(size) };
````
so windows specific code should internaly deal with cast

**Comment by arvidn** [comment]
looks good overall. there may be some opportunities to cast less by changing the types. maybe you could add #pragmas to mpi.c to disable warnings instead.

File: test/Jamfile (hunk 6 above)
Comment (ajax16384) [inline test/Jamfile]: ok

File: src/enum_net.cpp (hunk 2 above)
Comment (ajax16384) [inline src/enum_net.cpp]: ok

File: src/http_seed_connection.cpp (hunk 4 above)
Comment (ajax16384) [inline src/http_seed_connection.cpp]:
````c++
		virtual void sent_bytes(int bytes_payload, int bytes_protocol) = 0;
		virtual void received_bytes(int bytes_payload, int bytes_protocol) = 0;
		virtual void trancieve_ip_packet(int bytes, bool ipv6) = 0;
````
may be all of these methods should use size_t instead of int. 
there is no logic that sent or received bytes can be < 0 (special error case) ?

File: test/test_resolve_links.cpp (hunk 9 above)
Comment (ajax16384) [inline test/test_resolve_links.cpp]: ok

File: test/make_torrent.cpp (hunk 8 above)
Comment (ajax16384) [inline test/make_torrent.cpp]: ok

File: test/make_torrent.cpp (hunk 7 above)
Comment (ajax16384) [inline test/make_torrent.cpp]: ok

**Comment by ajax16384** [comment]
vs2010 seems to fail "examples" build:
 Cannot open include file: 'thread'
if it's not C++11 compiler then it should be dropped from AppVeyor script?

**Comment by arvidn** [comment]
sorry. I merged my long-standing patch to simplify the listen socket logic to master, breaking the merge with this one. hopefully there's not too much overlap

File: src/http_seed_connection.cpp (hunk 4 above)
Comment (arvidn) [inline src/http_seed_connection.cpp]: no. bytes_transferred should always be positive (including zero).

File: include/libtorrent/debug.hpp [hunk 11]
```diff
@@ -169,8 +169,8 @@ namespace libtorrent
 
 #else
 
-#define ADD_OUTSTANDING_ASYNC(x) do {} while(0)
-#define COMPLETE_ASYNC(x) do {} while(0)
```
arvidn: ah, right.. thanks

src/file.cpp
arvidn: makes sense

examples/Jamfile
```diff
@@ -13,6 +13,8 @@ project client_test
    : requirements
 	<threading>multi <library>/torrent//torrent
 	<toolset>darwin:<cflags>-Wno-unused-command-line-argument
+# disable warning C4275: non DLL-interface classkey 'identifier' used as base for DLL-interface classkey 'identifier'
+	<toolset>msvc:<cflags>/wd4275
```
arvidn: I'm not sure why this isn't working on appveyor. is it working for you locally?

arvidn: btw, isn't it enough to set this in the library build rule?

include/libtorrent/utp_socket_manager.hpp
```diff
@@ -104,8 +104,7 @@ namespace libtorrent
 		int loss_multiplier() const { return m_sett.get_int(settings_pack::utp_loss_multiplier); }
 
 		void mtu_for_dest(address const& addr, int& link_mtu, int& utp_mtu);
-		void set_sock_buf(int size);
-		int num_sockets() const { return int(m_utp_sockets.size()); }
+		int num_sockets() const { return m_utp_sockets.size(); }
```
arvidn: was this a merge issue? I would have expected m_utp_sockets.size() to have to be cast to an int

simulation/Jamfile
```diff
@@ -18,8 +18,6 @@ project
 	<source>utils.cpp
 	: default-build
 	<threading>multi
-# disable warning C4275: non DLL-interface classkey 'identifier' used as base for DLL-interface classkey 'identifier'
-	<toolset>msvc:<cflags>/wd4275
```
arvidn: ah, this is what was causing the failure.. because it was under default-build, rather than requirements (if there are warnings when building the simulation, you could probably add it back there)

src/http_seed_connection.cpp
ajax16384: i will try to make separate pull request about these methods

PR #649, opened 2016-04-26
Using aux::time_now() to pass time to dos_blocker

aldenml:


aldenml:
The error is one of these non deterministic ones

arvidn:
I'm sorry. but I'm having second thoughts about this one. seeing the patch reminds me that this code could potentially be run quite often, especially under the scenario it's trying to protect from

aldenml:
If you have doubts, let me close this, the issue is open for further discussion. I also have doubts, since the resolution is configurable and could potentially be more a lot more than 100ms.

PR #624, opened 2016-04-16
fix default release build with VS2015 UP2 compiler warnings

ajax16384:
this will allow to successfully build:
bjam toolset=msvc-14.0 release deprecated-functions=off character-set=unicode link=static warnings-as-errors=on

PR #650, opened 2016-04-26
sent_bytes,received_bytes,trancieve_ip_packet: int->size_t

ajax16384:
- change sent_bytes,received_bytes,trancieve_ip_packet arguments from int to size_t
- fix rest size_t to int conversion warns
- fix ms c++ override,final,noexcept detection

### DIFF
--- a/examples/bt-get2.cpp
+++ b/examples/bt-get2.cpp
@@ -85,7 +85,7 @@ int main(int argc, char const* argv[])
 		, std::istream_iterator<char>()};
 
 	lt::error_code ec;
-	lt::add_torrent_params atp = lt::read_resume_data(&buf[0], buf.size(), ec);
+	lt::add_torrent_params atp = lt::read_resume_data(&buf[0], int(buf.size()), ec);
 	atp.url = argv[1];
 	atp.save_path = "."; // save in current dir
 	ses.async_add_torrent(atp);
--- a/include/libtorrent/config.hpp
+++ b/include/libtorrent/config.hpp
@@ -431,23 +431,23 @@ int snprintf(char* buf, int len, char const* fmt, ...)
 #ifdef _GLIBCXX_USE_NOEXCEPT
 #define TORRENT_EXCEPTION_THROW_SPECIFIER _GLIBCXX_USE_NOEXCEPT
 #else
-#if __cplusplus <= 199711L || defined BOOST_NO_CXX11_NOEXCEPT
-#define TORRENT_EXCEPTION_THROW_SPECIFIER throw()
-#else
+#if !defined BOOST_NO_CXX11_NOEXCEPT && (__cplusplus > 199711L || (defined _MSC_VER && _MSC_VER >= 1900)) 
 #define TORRENT_EXCEPTION_THROW_SPECIFIER noexcept
+#else
+#define TORRENT_EXCEPTION_THROW_SPECIFIER throw()
 #endif
 #endif // __GLIBC__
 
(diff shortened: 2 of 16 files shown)


include/libtorrent/config.hpp
```diff
@@ -431,23 +431,23 @@ int snprintf(char* buf, int len, char const* fmt, ...)
 #ifdef _GLIBCXX_USE_NOEXCEPT
 #define TORRENT_EXCEPTION_THROW_SPECIFIER _GLIBCXX_USE_NOEXCEPT
 #else
-#if __cplusplus <= 199711L || defined BOOST_NO_CXX11_NOEXCEPT
-#define TORRENT_EXCEPTION_THROW_SPECIFIER throw()
-#else
+#if !defined BOOST_NO_CXX11_NOEXCEPT && (__cplusplus > 199711L || (defined _MSC_VER && _MSC_VER >= 1900))
```
arvidn: I intend to require c++11 in master (travis and appveoyr already dropped c++98 builds)

arvidn:
seeing this patch, I get a bit nervous about all the unsigned types.

examples/bt-get2.cpp
```diff
@@ -85,7 +85,7 @@ int main(int argc, char const* argv[])
 		, std::istream_iterator<char>()};
 
 	lt::error_code ec;
-	lt::add_torrent_params atp = lt::read_resume_data(&buf[0], buf.size(), ec);
+	lt::add_torrent_params atp = lt::read_resume_data(&buf[0], int(buf.size()), ec);
```
glassez: `static_cast`, please.

ajax16384: @glassez it was made by intention: https://github.com/arvidn/libtorrent/pull/624#discussion_r59974899
"type()" form was chosen, so any further pull requests use such style

glassez: Strange choice... It is slightly different from C-style cast (only syntactically AFAIK). In fact it is exactly the same "uncontrolled" cast. But, if @arvidn permits such a cast, I'm not going to go into it.

include/libtorrent/peer_connection.hpp
```diff
@@ -458,9 +458,9 @@ namespace libtorrent
 
 		stat const& statistics() const { return m_statistics; }
 		void add_stat(boost::int64_t downloaded, boost::int64_t uploaded);
-		void sent_bytes(int bytes_payload, int bytes_protocol);
-		void received_bytes(int bytes_payload, int bytes_protocol);
-		void trancieve_ip_packet(int bytes, bool ipv6);
```
arvidn: I'm having second thoughts about turning all these into unsigned. it feels scary

glassez: Why?

arvidn: because unsigned types are dangerous. their arithmetic operators don't behave like regular math.

glassez: >their arithmetic operators don't behave like regular math

I wonder what you mean...
>I've actually been planning a blog post about this, I may write that up today

Please ping me when you do it.

arvidn: http://blog.libtorrent.org/2016/05/unsigned-integers/

ajax16384: nice blog entry, thanks for usefull information.
according to thise info and many c++11 changes in master this pull request is obsolete and can be closed without merge

PR #657, opened 2016-04-28
Replace all occurances of thread with libtorrent::thread to avoid ambiguity

UnitedMarsupials:
Without this compiling tests on FreeBSD-10 (using clang-3.4.1) fails because unqualified "thread" is ambiguous.

### DIFF
--- a/test/dht_server.cpp
+++ b/test/dht_server.cpp
@@ -89,7 +89,7 @@ struct dht_server
 
 		fprintf(stderr, "%s: DHT initialized on port %d\n", time_now_string(), m_port);
 
-		m_thread.reset(new thread(boost::bind(&dht_server::thread_fun, this)));
+		m_thread.reset(new libtorrent::thread(boost::bind(&dht_server::thread_fun, this)));
 	}
 
 	~dht_server()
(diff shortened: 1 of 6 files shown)


arvidn:
I wouldn't expect there to be an ambiguity in libtorrent's own translation units.

however, instead of making this change, maybe it would make sense to just change it to std::thread (and remove the libtorrent::thread typedef altogether)

UnitedMarsupials:
> I wouldn't expect there to be an ambiguity in libtorrent's own translation units.

Compiler complained here -- the other candidate was `struct thread` declared in `<sysctl.h>`.
> maybe it would make sense to just change it to std::thread (and remove the libtorrent::thread typedef altogether)

Hey, fine with me -- as long as the tests compile :-)

arvidn:
this issue presumably affects the 1.1 release too.. would you mind submitting this PR against the RC_1_1 branch?

arvidn:
I moved this to RC_1_1 here https://github.com/arvidn/libtorrent/pull/664

arvidn:
I pushed this to RC_1_1 and it has been merged to master now. thanks!

PR #660, opened 2016-04-28
attempt to fix race condition in udp_tracker in test framework

arvidn:


arvidn:
@aldenml this is my attempt at fixing that race condition. I'm not too confident it will make a difference, since I don't understand why it's failing sometimes

aldenml:
Ok, this solution is interesting, if anything at least it should help

PR #669, opened 2016-04-29
remove snprintf-hack and use the standard library

arvidn:


arvidn:
gah. of all the C++11 features I've used so far, presumably the oldest of them all requires visual studio 2015!

PR #670, opened 2016-04-30
fix union storage hacks

arvidn:


arvidn:
apparently GCC didn't get aligned_union until 5.0. Which is a lot newer than what travis has

PR #671, opened 2016-04-30
Threads

arvidn:


ssiloti:
Yay! No more raiding ASIO's internals for synchronization primitives :+1:

arvidn:
yeah, now I just need to fix all the includes.. I guess clang pulls in these headers transitively somewhere

PR #675, opened 2016-04-30
Use unique_ptr where possible

ssiloti:
There are disappointingly few places where shared_ptr can be trivially replaced with unique_ptr. boost::asio is a real buzzkill in this respect, pretty much anything which is bound to an event handler needs to be a shared_ptr to prevent dangling references.

### DIFF
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -166,7 +166,8 @@ namespace libtorrent
 		// the actual sockets (TCP listen socket and UDP socket)
 		// An entry does not necessarily have a UDP or TCP socket. One of these
 		// pointers may be null!
-		// TODO: 3 make these unique_ptr<>
+		// These must be shared_ptr to avoid a dangling reference if an
+		// incoming packet is in the event queue when the socket is erased
 		boost::shared_ptr<tcp::acceptor> sock;
 		boost::shared_ptr<udp_socket> udp_sock;
 	};
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -760,7 +760,7 @@ namespace libtorrent
 
 			// a thread pool used for async_write_some calls,
 			// to distribute its cost to multiple threads
-			std::vector<boost::shared_ptr<network_thread_pool> > m_net_thread_pool;
+			std::vector<std::unique_ptr<network_thread_pool>> m_net_thread_pool;
 
 			// the bandwidth manager is responsible for
 			// handing out bandwidth to connections that
--- a/include/libtorrent/aux_/session_settings.hpp
+++ b/include/libtorrent/aux_/session_settings.hpp
@@ -77,8 +77,8 @@ namespace libtorrent { namespace aux
 		{
 			TORRENT_ASSERT((name & settings_pack::type_mask) == type);
 			if ((name & settings_pack::type_mask) != type) return;
-			int const index = name & settings_pack::index_mask;
-			TORRENT_ASSERT(index >= 0 && index < N);
+			size_t const index = name & settings_pack::index_mask;
+			TORRENT_ASSERT(index < N);
 			arr[index] = val;
 		}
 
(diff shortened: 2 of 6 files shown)


include/libtorrent/aux_/session_impl.hpp
```diff
@@ -765,7 +766,7 @@ namespace libtorrent
 
 			// a thread pool used for async_write_some calls,
 			// to distribute its cost to multiple threads
-			std::vector<boost::shared_ptr<network_thread_pool> > m_net_thread_pool;
+			std::vector<std::unique_ptr<network_thread_pool>> m_net_thread_pool;
```
arvidn: heh, running across this old code this morning put it pretty high on my list of things to kill :) or at least greatly simplify its implementation

include/libtorrent/aux_/session_settings.hpp
```diff
@@ -77,7 +77,7 @@ namespace libtorrent { namespace aux
 		{
 			TORRENT_ASSERT((name & settings_pack::type_mask) == type);
 			if ((name & settings_pack::type_mask) != type) return;
-			int const index = name & settings_pack::index_mask;
+			size_t const index = name & settings_pack::index_mask;
```
arvidn: was there a warning here? is the type of the enum unsigned? if not, I don't see how the result here would become unsigned

include/libtorrent/thread_pool.hpp
```diff
@@ -59,8 +59,8 @@ namespace libtorrent
 				while (m_num_threads < n)
 				{
 					++m_num_threads;
-					m_threads.push_back(boost::shared_ptr<thread>(
-						new thread(boost::bind(&thread_pool::thread_fun, this, int(m_num_threads)-1))));
+					m_threads.emplace_back(
```
arvidn: oh uh. this will conflict with a patch I've been baking today. I'm just putting the threads straight into the vector instead, since we have move semantics now

include/libtorrent/aux_/session_settings.hpp
ssiloti: The warning was in the assert below where index is compared to N

include/libtorrent/tracker_manager.hpp
```diff
@@ -400,7 +400,8 @@ namespace libtorrent
 		mutable mutex m_mutex;
 
 		// maps transactionid to the udp_tracker_connection
-		// TODO: this should be unique_ptr in the future
+		// These must use shared_ptr to avoid a dangling reference
+		// if a connection is erased while a timeout event is in the queue
```
arvidn: yeah, working on the listen socket patch I realized that this pattern isn't that great. It may seem nice to begin with, but I believe components may be more composable if async operations are issued and handled from outside the object itself (like all the built-in asio types). That way the user of a class can control its life-time, including destroying it before all outstanding operations have come back

include/libtorrent/udp_socket.hpp
```diff
@@ -136,10 +135,8 @@ namespace libtorrent
 
 		udp::socket m_socket;
 
-		// TODO: 2 this should probably be a scoped_ptr<> or unique_ptr
-		// with a hard coded size
-		int const m_buf_size;
-		char* m_buf;
+		using receive_buffer = std::array<char, 1500>;
```
arvidn: nice

src/disk_io_thread.cpp
```diff
@@ -241,9 +241,9 @@ namespace libtorrent
 				// the magic number 3 is also used in add_job()
 				// every 4:th thread is a hasher thread
 				if ((thread_id & 0x3) == 3) type = hasher_thread;
-				m_threads.push_back(boost::shared_ptr<thread>(
+				m_threads.emplace_back(
```
arvidn: I believe these thread objects are no longer heap-allocated either (in the "threads" branch)

src/udp_socket.cpp
```diff
@@ -162,19 +162,11 @@ struct set_dont_frag
 
 udp_socket::udp_socket(io_service& ios)
 	: m_socket(ios)
-	, m_buf_size(1500)
-	, m_buf(NULL)
+	, m_buf(new receive_buffer())
```
arvidn: if only we had make_unique<> :)

include/libtorrent/aux_/session_settings.hpp
arvidn: I see. I would have expected another warning now, testing an unsigned type for being >= 0

arvidn:
you may want to rebase now

PR #676, opened 2016-04-30
use standard array container

ssiloti:


arvidn:
nice

arvidn:
that test failure presumably is not related to this patch, right? does it pass locally for you, on windows?

ssiloti:
It passes for me locally. I doubt it's related to this patch.

PR #680, opened 2016-05-01
minor refactoring to move-storage patch

arvidn:


arvidn:
@glassez would you mind taking a look at this patch? it's an update to your patch I just pulled.

src/storage.cpp
```diff
@@ -1061,32 +1068,34 @@ namespace libtorrent
...
 		}
 
-		for (std::string subdir: subdirs)
+		for (std::string const& s : subdirs)
 		{
-			e.clear();
-			while (!subdir.empty() && !e)
+			error_code err;
+			std::string subdir = combine_path(old_save_path, s);
+			while (subdir != old_save_path && !err)
```
glassez: Why?
empty() is faster than comparison.

arvidn: combine_path isn't though. this saves a call to that inside the loop. the comparison here will most likely be quite fast too, since the strings have different length (that should be an early exit)

glassez: ok.

glassez:
>would you mind taking a look at this patch?

Looks OK. If it will take tests, then it shouldn't be a problem.
Although I haven't watched it thoroughly.

PR #681, opened 2016-05-01
add move ctor to entry

ssiloti:


include/libtorrent/entry.hpp
```diff
@@ -154,6 +155,7 @@ namespace libtorrent
 #endif
 		entry& operator=(bdecode_node const&);
 		entry& operator=(entry const&);
+		entry& operator=(entry&&);
 		entry& operator=(dictionary_type const&);
 		entry& operator=(string_type const&);
 		entry& operator=(list_type const&);
```
arvidn: this reminds me. the internal object we hold should probably be move-constructed in here too. I wonder if it would make sense to replace some of the internals here with boost.variant. I'm not sure how similar it is to what's coming in c++17 though..

arvidn:
lgtm

PR #682, opened 2016-05-01
update CMakeLists.txt to build with C++11

arvidn:


arvidn:
@jpetso does this look good (as a first step at least)?

aldenml:
@arvidn just to know about the status of this PR.

jpetso:
That looks good to me. I'll close my other PR, making the assumption that using it from a C++14/17 binary doesn't end up in (Boost) ABI issues. Sorry for having dragged this out longer than necessary, and thanks for considering it as the default.

PR #684, opened 2016-05-01
remove mutex in natpmp

arvidn:


arvidn:
merged with the upnp PR

PR #685, opened 2016-05-01
Msvc warnings

arvidn:


ajax16384:
thanks for fixing rest of msvc warns

arvidn:
there are plenty more :) this is just ehwn building examples. and when I build locally (with msvc-12.0) I get a lot more too for some reason

PR #686, opened 2016-05-01
Using address::is_multicast instead of util function

aldenml:
I wonder if you had any particular reason for the `if-else` here and if a similar reasoning is in `is_loopback`.

### DIFF
--- a/src/broadcast_socket.cpp
+++ b/src/broadcast_socket.cpp
@@ -97,20 +97,6 @@ namespace libtorrent
 #endif
 	}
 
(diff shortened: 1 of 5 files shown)


src/broadcast_socket.cpp
```diff
@@ -99,16 +99,7 @@ namespace libtorrent
 
 	bool is_multicast(address const& addr)
```
arvidn: maybe this function should be removed altogether then..

aldenml: I will do it.

PR #687, opened 2016-05-01
Moved disabled_storage and zero_storage for a private namespace in the compilation unit

aldenml:
I think this makes sense, but the original motivation is that swig is having a hard time with the `final` keyword in the class declaration.

### DIFF
--- a/include/libtorrent/storage.hpp
+++ b/include/libtorrent/storage.hpp
@@ -470,62 +470,6 @@ namespace libtorrent
 		bool m_allocate_files;
 	};
 
-	// this storage implementation does not write anything to disk
-	// and it pretends to read, and just leaves garbage in the buffers
-	// this is useful when simulating many clients on the same machine
-	// or when running stress tests and want to take the cost of the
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -318,6 +318,8 @@ if (MSVC)
 elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
 	add_definitions(-Wno-c++11-extensions)
 	add_definitions(-fcolor-diagnostics)
+	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11")
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
```
arvidn: I have this PR to add C++11 support to the CMake file. if it looks good to you, I'll merge it: https://github.com/arvidn/libtorrent/pull/682

aldenml: This two lines was a mistake, I will remove them

aldenml: Done, that was a temporary workaround

PR #688, opened 2016-05-01
remove network thread pool feature

arvidn:
(it was messy and mostly useful for bittorrent over SSL). The proper solution is to allow every peer run in its own asio strand and have multiple threads running the io service


mkatsevVR:
i'm quite interested in having multiple network threads, what would be the best way to resurrect/improve some form of this in 1.2?

mkatsevVR:
In the approach with strands, it seems that there are a lot of non-peer related operations (eg everything in session_impl) that also need to be synchronized and it would be hard to guarantee that every single one is wrapped in a strand. Would it make sense to have 2 io_services: keep a single-threaded one for all non-peer_connection operations and then add a new one multithreaded one for peer_connection specifically, with strand per connection?

Still seems risky, there are possibly plenty of places that assume only having a single thread between session and networking. The original approach seems quite reasonable to me, tbh. Are you strongly opposed to bringing it back as-is?

arvidn:
the main reason I removed the thread pool here was that it was actually not correct, and was causing memory corruption (but it was non-trivial to reproduce, it only happened under very heavy load).

Despite some efforts to make it more complex by forcing reads and writes on the same socket to be queued on the same thread, it didn't seem the complexity warranted the small gain (I can't recall whether I actually got it to work).

What made it more complex was that uTP connections, sharing the `utp_socket_manager` under the hood, could never be run concurrently anyway.

Instead I've had a long-term (slow moving admittedly) task to lower the peer <-> torrent and torrent <-> session interdependencies, to allow a future where each peer has its own strand (`boost::asio::strand`). Perhaps with an intermediate patch where each torrent has its own strand (since reducing torrent's dependency on the session is a bit simpler). The latter would at least allow peers from different torrents to run on separate threads.

Is your main use case connections over SSL? In my case that was the main motivation, as they spend a lot of CPU for the crypto.

In my mind, the ideal end-state is that everything uses strands to define what can be concurrent and what cannot, to the point where the *user* can provide an `io_service` (or `io_context` nowadays) and completely define how many threads to run, and to run other things on the same context.

I think the main challenge really is the peer callbacks and to some degree the torrent callbacks. They currently assume that they can do whatever they like, and have access to the session and peers freely.

I don't think it would be that much work to mark up all async operations in the session to use a strand, but that wouldn't be sufficient, since both peers and torrents access the session from their own handlers/callbacks.

arvidn:
so, to answer your question, I'm not strongly opposed to bringing it back, but I would like it to be simpler than it was in that case.

mkatsevVR:
> Is your main use case connections over SSL?

Yes. Maxing out at ~350MB/s (yes, I understand that this is likely not a typical use case of libtorrent :) )

> The latter would at least allow peers from different torrents to run on separate threads.

Sadly that doesn't help my case, since I only have 1 torrent anyway.

> was causing memory corruption

Do you know why? Is it because of some other operations triggered by torrent or session that touch the socket while it was being used for reads/writes? Eg if torrent decides to disconnect a peer/destroy the connection while peer_connection is reading from the socket?

> the ideal end-state is that everything uses strands

I've not worked with them before, but they look kinda limited to me. Eg if you have operations A, B and C, you can't express that A and B can be run concurrently, but C can't be run concurrently with either of A or B (eg A=read from socket, B=some torrent operation that doesn't affect the socket, C=disconnect peer from the torrent, thus affecting both the state of the torrent and the socket, so needs to be exclusive with both)

arvidn:
> Do you know why? Is it because of some other operations triggered by torrent or session that touch the socket while it was being used for reads/writes? Eg if torrent decides to disconnect a peer/destroy the connection while peer_connection is reading from the socket?

I don't recall unfortunately. I'm not 100% sure there were memory corruption with the latest version (but I think so). Perhaps SSL isn't thread safe in allowing separate threads to read and write simultaneously. I think that's most likely the case. Perhaps read and write operations were put on the same thread in the last incarnation of the network thread, but maybe not.

I also wonder if there's some possibility for intermediate operations, handled internally by the ssl asio layer, may cause threading issues too, as they will definitely run on the main thread. This situation is a bit better with the networking TS, as executors (i.e. strands) propagate down in composing operations.

arvidn:
here's a reference: https://github.com/arvidn/libtorrent/issues/3084

PR #694, opened 2016-05-03
* now calling stop_announcing() unconditionally instead of checking !…

jonathanmcdougall:
…is_paused(), which was always false when aborting a torrent and could cause infinite blocks while destroying a session

* fixed torrent invariant check that wasn't using the m_abort flag


arvidn:
thanks!

PR #698, opened 2016-05-04
added new preformatted type to bencode entry

arvidn:
to support carrying a verbatim copy of an already bencoded subtree. This is to support saving torrents in resume data and create_torrent based on existing torrents, in order to preseve key order

### DIFF
--- a/test/test_bencoding.cpp
+++ b/test/test_bencoding.cpp
@@ -58,54 +58,72 @@ entry decode(std::string const& str)
 	return bdecode(str.begin(), str.end());
 }
 
-TORRENT_TEST(bencoding)
+TORRENT_TEST(strings)
 {
-	// ** strings **
-	{
-		entry e("spam");
-		TEST_CHECK(encode(e) == "4:spam");
-		TEST_CHECK(decode(encode(e)) == e);
-	}
+	entry e("spam");
+	TEST_CHECK(encode(e) == "4:spam");
+	TEST_CHECK(decode(encode(e)) == e);
+}
 
-	// ** integers **
-	{
-		entry e(3);
-		TEST_CHECK(encode(e) == "i3e");
-		TEST_CHECK(decode(encode(e)) == e);
-	}
+TORRENT_TEST(intergers)
+{
+	entry e(3);
+	TEST_CHECK(encode(e) == "i3e");
+	TEST_CHECK(decode(encode(e)) == e);
+}
 
-	{
-		entry e(-3);
-		TEST_CHECK(encode(e) == "i-3e");
-		TEST_CHECK(decode(encode(e)) == e);
-	}
+TORRENT_TEST(intergers2)
+{
+	entry e(-3);
+	TEST_CHECK(encode(e) == "i-3e");
+	TEST_CHECK(decode(encode(e)) == e);
+}
 
-	{
-		entry e(int(0));
-		TEST_CHECK(encode(e) == "i0e");
-		TEST_CHECK(decode(encode(e)) == e);
-	}
+TORRENT_TEST(intergers3)
+{
+	entry e(int(0));
+	TEST_CHECK(encode(e) == "i0e");
+	TEST_CHECK(decode(encode(e)) == e);
+}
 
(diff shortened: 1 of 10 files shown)


sledgehammer999:
How am I supposed to test this? I compiled and run it but it still mutates the hash. Do you want me to use some specific function?

arvidn:
hm.. I thought it would fix that case. the internal entry hierarchy in create_torrent is supposed to preserve the info-dict from the torrent, when passing in a torrent_handle to it (with this patch). and bencode() is supposed to support include such pre-encoded sections.

would you mind stepping into the create_torrent constructor to make sure the metadata buffer is preserved verbatim?

the other place where this could be broken is in bencode(). You do use the bencode() function in libtorrent to encode the entry, right?

The unit test passes, so I'm a bit surprised it didn't work. I may have overlooked some code path that you're hitting.

sledgehammer999:
Sadly I cannot test further tonight. However I noticed a slight change from my previous test. I don't know if it is your patch or that I was using RC_1_0.

When the metadata are fetched from DHT we save them in our folder as `hash.torrent` and also creating `hash.fastresume`. Without your patch this hash was the mutated one. With your patch it is the initial hash. However, upon restart and loading those files libtorrent shows the mutated hash instead. Could the loading functions mutate it instead? We use this signature to load it: `torrent_info(std::string const& filename, error_code& ec, int flags = 0);`

sledgehammer999:
Well I just checked again. The initially saved torrent is mutated. The info fields are rearranged. ie the length field gets rearranged above the name field.

arvidn:
basically any time the torrent gets turned into an entry, it will get reordered, so if your load path includes a decode step into an entry, that's a problem. However, there are plenty of alternatives for loading. bdecode_node is probably the best one. It's faster, uses less memory and doesn't reorder fields.

sledgehammer999:
As I said it is written mutated to disk(create_torrent -> bencode). Also bdecode_node isn't available in RC_1_0.

arvidn:
I added a unit test and discovered a bug in create_torrent. it should work now.

sledgehammer999:
Great! Now it seems to work!!! Please backport to RC_1_0 if possible.

test/test_bencoding.cpp
```diff
@@ -58,54 +58,72 @@ entry decode(std::string const& str)
...
+	TEST_CHECK(decode(encode(e)) == e);
+}
 
-	// ** integers **
-	{
-		entry e(3);
-		TEST_CHECK(encode(e) == "i3e");
-		TEST_CHECK(decode(encode(e)) == e);
-	}
+TORRENT_TEST(intergers)
```
sledgehammer999: Did you mean `integers` ? Same typo(?) in the following tests too.

test/test_bencoding.cpp
```diff
@@ -58,54 +58,72 @@ entry decode(std::string const& str)
...
+	TEST_CHECK(encode(e) == "d3:cow3:moo4:spam4:eggse");
+	TEST_CHECK(decode(encode(e)) == e);
+}
+
+TORRENT_TEST(preformatted)
+{
+	entry e(entry::preformatted_t);
+	char const str[] = "foobar";
+	e.preformatted().assign(str, str + sizeof(str)-1);
+	TEST_EQUAL(encode(e), "foobar");
```
sledgehammer999: Unless I am mistaken the original problem occured because the encoder liked to sort the fields alphabetically and not keep the order they were appended in. So I propose to test a preformatted dictionary which has its field in non alphabetical order.

sledgehammer999:
And reading up on bencode from wikipedia, I realize that we face this problem because of some broken torrent creator program/library. Dictionaries are supposed to be sorted alphabetically but that broken implementation doesn't do that. Too bad the metadata don't record the torrent creator.

test/test_bencoding.cpp
arvidn: I added such test in test_create_torrent.cpp. This is just a unit test for the basic functionality of supporting arbitrary buffers in an entry hierarchy

PR #700, opened 2016-05-05
remove heap allocation of std::vector

ssiloti:


ssiloti:
Bah, so msvc 12 doesn't support unrestricted unions. Oh well, I'll just remove everything but the disk_buffer_pool changes.

arvidn:
maybe you're right. given that there is a free version of msvc-14 is a good enough reason to not support any older versions

ssiloti:
In this case it's not a big deal because the two job types which are affected are `file_priority` and `check_fastresume` which are not generated very often.

That failure on AppVeyor is due to the test_resume.cpp.share_mode test which has been flapping. I guess I should look into that.

src/disk_buffer_pool.cpp
```diff
@@ -67,17 +67,12 @@ namespace libtorrent
 	namespace {
 
 	// this is posted to the network thread
-	void watermark_callback(std::vector<boost::weak_ptr<disk_observer> >* cbs)
+	void watermark_callback(std::vector<boost::weak_ptr<disk_observer>> cbs)
```
arvidn: are you sure the vector will be moved into here? presumably boost::bind() allow its function object to be called more than once, in which case it cannot move any of its state in. perhaps I'm missing something

ssiloti: Good point. I've now confirmed with a simple test program that posting a std::bind and taking the parameter by const reference in the bound function does not produce any copies.

arvidn:
i suppose that error was from an RC_1_1 merge. I'll fix that up

PR #704, opened 2016-05-05
back-port preformatted entry

arvidn:


arvidn:
@sledgehammer999 do you think this is ABI compatible?

sledgehammer999:
Sorry, I wish I could answer you. My knowledge is limited in this field. Maybe @glassez knows better.

sledgehammer999:
Sidenote: I confirm that this patch works.

glassez:
I don't know much about it. IIRC it should be backward-compatible since it does not modify the existing names, but only adds a new (I mean one new entry constructor).

PR #707, opened 2016-05-07
fix peer class leak in torrent and widen peer_class_t to allow for mo…

arvidn:
…re instances


arvidn:
this needs back-porting to RC_1_1 as well

PR #720, opened 2016-05-11
Dht stats 1.1

arvidn:


aldenml:
I can confirm this patch update the counters (I can't find the original issue to comment). Do need something else to merge this?

PR #724, opened 2016-05-11
fix file_storage::m_paths memory reuse

ajax16384:
after rename_file with path delimiter - file_storage::m_paths should be searched for branch_path without torrent_separtor

### DIFF
--- a/src/file_storage.cpp
+++ b/src/file_storage.cpp
@@ -168,6 +168,10 @@ namespace libtorrent
 			// and the branch path
 			branch_path = path.c_str();
 			branch_len = leaf - path.c_str();
+
+			// trim trailing slashes
+			if (branch_len > 0 && branch_path[branch_len-1] == TORRENT_SEPARATOR)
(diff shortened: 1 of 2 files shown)


arvidn:
I take it there's no way to tell the difference through the API, right? this is just a fix to this optimization. Or can you think of a test that could exercise this?

ajax16384:
let's say we have "test.torrent" with content:

abc/file1.txt
abc/file2.txt

even ordinal parse_torrent_file before this patch will create dublicate "abc" entries in m_paths

arvidn:
yes. it's definitely a good patch. I wonder if it's possible to detect the failure or success of this optimization through the API. If so, it would be possible to have a test for it. I'm not sure the path index is exposed anywhere. maybe the raw string pointer is exposed, and they could be compared

ajax16384:
ok

arvidn:
awesome, thanks!

PR #729, opened 2016-05-14
fix assert when adding an empty file (as a torrent)

arvidn:


arvidn:
this will need back-porting to RC_1_1 as well

PR #744, opened 2016-05-21
attempt to get GCC 5 in travis

arvidn:


aldenml:
Hi @arvidn, is this related to #741. I wonder if not testing in a less modern compiler could bring potential problems in other environments (like android).

arvidn:
yes it is, and also #670. I would expect the android build environment to be c++11 capable by the time libtorrent master is stable.

aldenml:
The status of android NDK right now is depressing. They deprecated the gcc toolchain and will never upgrade it to gcc5. The clang one is buggy (judging by the tickets reported in github). Let's see

PR #745, opened 2016-05-22
replace uses of boost::bind with std::bind and lambdas

arvidn:


arvidn:
@ssiloti do you have any ideas of what may have caused the link error in mingw/gcc5?
see https://ci.appveyor.com/project/arvidn/libtorrent/build/1.0.2598/job/0sa34lbhfh5m2ayq

PR #749, opened 2016-05-22
keep debug-iterators consistent between the tests and library

ssiloti:
With GCC 5 debug-iterators changes the type of STL containers so the build fails
due to undefined symbols if the setting is not consistent

### DIFF
--- a/test/Jamfile
+++ b/test/Jamfile
@@ -73,7 +73,6 @@ lib libtorrent_test
 	<target-os>windows:<library>advapi32
 	<conditional>@link_libtorrent
 	<toolset>darwin:<cflags>-Wno-unused-command-line-argument
-	<debug-iterators>on
 
 	: # default build
 	<link>shared


arvidn:
the debug-iterators feature is supposed to be propagated to dependencies though. isn't it?

ssiloti:
In this case it's being set in the tests lib which is dependent on libtorrent, I guess boost build doesn't propagate features to reverse dependencies. I'm kind of surprised it doesn't complain due to the feature being marked as link incompatible *shrug*. If the intention was to still have debug-iterators enabled then I think we still need to set it explicitly in the `link_libtorrent` rule.

arvidn:
yeah. it's possible that the fact that I have that link_libtorrent rule may break the automatic propagation of features. But I seem to recall it working well in the python bindings (where there's a PIC feature that is propagated correctly)

arvidn:
I would prefer the debug iterators are on by default when running the tests. Did I break this recently by removing the feature from default build perhaps? maybe it should _just_ be in the default build properties.

arvidn:
yes, I broke this recently by removing the features from the default build. [8fac4e7d5cf539451e9688967d08b0735d4807f7].

ssiloti:
I've tried every permutation and the build only succeeds if debug-iterators is explicitly set to on in both link_libtorrent and the requirements for libtorrent_test.

arvidn:
this doesn't work? https://github.com/arvidn/libtorrent/pull/751

arvidn:
actually, I didn't mean to explicitly enable it for the tests, just the other builds, to use the same binary built for the tests

PR #751, opened 2016-05-22
restore debug-iterators to be enabled by default when building the tests

arvidn:


arvidn:
@ssiloti ok, I think this last version fixes it. does this work for you?

ssiloti:
Yes this works.

arvidn:
I'll pull this one instead of your patch then. thanks for bringing this up btw! I just updated my mac and got a new version of clang. I assumed these issues were related and was procrastinating looking into it :P

PR #754, opened 2016-05-25
Don't force the compiler to search in /usr/include

aldenml:
This was causing problems with android and the clang toolchain and probably other hidden issues.


arvidn:
it occurs to me that maybe it's a bad idea to add any of those system paths. if cross-compiling, I imagine they could cause trouble.

aldenml:
Do you think we can cleanup this little by little?

arvidn:
sure

arvidn:
would you mind rebasing this on top of master to try again? I had the same problem in the std::bind branch, so I switched to mingw32, which worked (and is in master now).

aldenml:
rebased

PR #755, opened 2016-05-25
Added support for x.pe parameter in parse_magnet_uri

aldenml:
`BEP 9: 16-May-2016: added peer source parameter to magnets`

### DIFF
--- a/src/magnet_uri.cpp
+++ b/src/magnet_uri.cpp
@@ -32,15 +32,12 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "libtorrent/magnet_uri.hpp"
 #include "libtorrent/session.hpp"
-#include "libtorrent/torrent_handle.hpp"
 #include "libtorrent/aux_/escape_string.hpp"
-#include "libtorrent/error_code.hpp"
 #include "libtorrent/torrent_status.hpp"
 #include "libtorrent/torrent_info.hpp"
 #include "libtorrent/announce_entry.hpp"
 #include "libtorrent/hex.hpp" // to_hex, from_hex
-
-#include <string>
+#include "libtorrent/socket_io.hpp"
 
 namespace libtorrent
 {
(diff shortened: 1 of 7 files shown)


src/magnet_uri.cpp
```diff
@@ -230,6 +230,33 @@ namespace libtorrent
 			return;
 		}
 
+		std::string::size_type peer_pos = std::string::npos;
+		std::string peer = url_has_argument(uri, "x.pe", &peer_pos);
+		while (!peer.empty())
+		{
+			std::string::size_type divider = peer.find_last_of(':');
```
arvidn: I'm pretty sure the logic to parse an IPv4/IPv6 endpoint exists somewhere else in libtorrent. let me dig it up

test/test_magnet.cpp
```diff
@@ -279,6 +279,20 @@ TORRENT_TEST(parse_space_hash)
...
 }
 
+TORRENT_TEST(parse_peer)
+{
+	error_code ec;
+	add_torrent_params p;
+	parse_magnet_uri("magnet:?xt=urn:btih:cdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcd&dn=foo&x.pe=127.0.0.1:43&x.pe=<invalid1>&x.pe=<invalid2>:100&x.pe=[::1]:45", p, ec);
+	TEST_CHECK(!ec);
+	TEST_EQUAL(p.peers.size(), 2);
+	TEST_EQUAL(p.peers[0].address(), address::from_string("127.0.0.1"));
```
arvidn: this could be shorter (and I think more readable) by using the convenience function to create endpoints, ep() https://github.com/arvidn/libtorrent/blob/master/test/setup_transfer.hpp#L109

src/magnet_uri.cpp
arvidn: I believe you could use parse_endpoint() https://github.com/arvidn/libtorrent/blob/master/src/socket_io.cpp#L86

aldenml:
I changed it to use `parse_endpoint` and `ep`. Thanks for remind me these functions.

I think I will put some changes in `parse_magnet_uri` old code and `parse_endpoint` but I will do that in RC_1_1

arvidn:
this should probably be mentioned in the Changelog as well. that change doesn't need to pass CI though :p

aldenml:
ok :) done.

arvidn:
thanks!

gubatron:
wohoo 👍

PR #757, opened 2016-05-26
Fix of parse_endpoint with invalid port character separator

aldenml:


aldenml:
Hi @arvidn, could you take a quick look at the failed mac os x build? I can't see the error

arvidn:
line 460: /bin/sh: line 9: 17407 Segmentation fault: 11  "bin/test_primitives.test/0fe1b7b619650a3d000818acf2272f80/test_primitives" > "bin/test_primitives.test/0fe1b7b619650a3d000818acf2272f80/test_primitives.output" 2>&1

it seems like test_primitives segfaults

arvidn:
I've been wanting to have a sanitize=address build as part of the CI (which probably would have made this error contain more useful information). Now that we have GCC5 on travis, that should be possible.

aldenml:
What do I do? do you think it's related to my change?

src/socket_io.cpp
```diff
@@ -105,6 +105,11 @@ namespace libtorrent
 			}
 			*port_pos = '\0';
 			++port_pos;
+			if (*port_pos != ':')
```
arvidn: you dereference the port_pos iterator here, but it's possible that it's the one-past-end by this point, which is undefined behavior.

aldenml: Right, and I will add another test for that case

arvidn:
thanks!

PR #760, opened 2016-05-26
Ported support for magnet x.pe parameter from master

aldenml:
I tried to do this with the minimal potential conflicts when merged to master, but the actual logic of adding the peers was impossible to add inside the torrent constructor due to the alerts using get_handle()

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -11579,6 +11579,7 @@ namespace libtorrent
 	{
 		if (!m_apply_ip_filter) return;
 		if (!m_peer_list) return;
+		if (!m_ip_filter) return;
 
 		torrent_state st = get_peer_list_state();
 		std::vector<address> banned;
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 1.1.1 release
 
 	* fixed parsing of IPv6 endpoint with invalid port character separator
+	* added limited support for new x.pe parameter from BEP 9
 	* fixed dht stats counters that weren't being updated
 	* make sure add_torrent_alert is always posted before other alerts for
 	  the torrent
--- a/include/libtorrent/magnet_uri.hpp
+++ b/include/libtorrent/magnet_uri.hpp
@@ -78,6 +78,9 @@ namespace libtorrent
 	// This function parses out information from the magnet link and populates the
 	// add_torrent_params object.
 	TORRENT_EXPORT void parse_magnet_uri(std::string const& uri, add_torrent_params& p, error_code& ec);
+
+	// internal, delete when merge in master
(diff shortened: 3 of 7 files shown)


src/torrent.cpp
```diff
@@ -11579,6 +11579,7 @@ namespace libtorrent
 	{
 		if (!m_apply_ip_filter) return;
 		if (!m_peer_list) return;
+		if (!m_ip_filter) return;
```
arvidn: I don't understand this. If we remove the filter, we still need to tell everyone about it

include/libtorrent/add_torrent_params.hpp
```diff
@@ -297,6 +297,10 @@ namespace libtorrent
 		// url seeds to be added to the torrent (`BEP 17`_).
 		std::vector<std::string> url_seeds;
 
+		// peers to add to the torrent, to be tried to be connected to as
+		// bittorrent peers.
+		std::vector<tcp::endpoint> peers;
```
arvidn: this breaks the ABI unfortunately :(

src/torrent.cpp
aldenml: The problem was with the line below `m_peer_list->apply_ip_filter(*m_ip_filter, &st, banned);`, it causes memory crashes

include/libtorrent/add_torrent_params.hpp
aldenml: I was afraid of this, but, is there a way to add something without breaking ABI?

src/torrent.cpp
arvidn: oh, right.. it dereferences the ip_filter.. I guess that's always expected to be set then. a bit weird..

include/libtorrent/add_torrent_params.hpp
arvidn: the only way I can think of would be to support this only via the "url" parameter in add_torrent_params. In this case, the magnet URI is parsed internally in the library, which can use new data structures. just as long as they're not user facing

arvidn: right now, that just uses parse_magnet_uri() as well though.. so it's still not obvious how to do that without splitting that function into a private and a public version..

aldenml: thinking about it

aldenml:
What do you think about this "limited support" solution?

arvidn:
seems reasonable

arvidn:
oh uh. maybe merging your other patch caused a conflict

aldenml:
oh no...the wait again :( let me rebase asap

aldenml:
Was the changelog, btw

PR #762, opened 2016-05-27
Share the DHT storage for both IPv4 and IPv6

aldenml:


src/kademlia/dht_storage.cpp
```diff
@@ -346,6 +345,7 @@ namespace
 		}
 
 		void put_immutable_item(sha1_hash const& target
+			, sha1_hash const& our_id
```
arvidn: won't this cause odd prioritizations between the two nodes? since "out ID" will be different depending on who's putting. presumably the ideal would be to prioritize based on the distance to _any_ of our node IDs. Not sure what the best way to do that is though..

aldenml: This is to remove one item in case we are full. I not familiar with the algebra of the `distance_exp(node_id)` but maybe there is a way to combine two of them `node_id ⨁ node_id -> node_id` that allows to calculate a desired "combined distance". Do you think that's possible?

arvidn: hm.. it's possible that there may be some clever hack, but I kind of doubt it. I think it would have to come down to calculating the distance to multiple node IDs and take the closest one as the "priority".

distance_exp() literally just takes a node IDs, XORs it with our node ID and then returns the xor distance between them (160 - number_of_leading_zeros)

the8472: > distance_exp() literally just takes a node IDs, XORs it with our node ID and then returns the xor distance between them (160 - number_of_leading_zeros)

Distances can differ in bits beyond the shared prefix bits. That's relevant if you want to compare multiple items relatively close to each other.

aldenml: I have a couple question I can't fully answer myself.
- One `get_immutable_item` request can come from any DHT, but what's the probability that a `put_immutable_item` with the same hash comes from multiples DHT having distinct IDs?
- Why `put_mutable_item` does not follow the same rule to discard "the least important one"?

ssiloti: 1. It is unlikely but possible, so the code shouldn't break if it happens.
2. I don't see any reason why mutable items should omit the distance metric. I think this is a bug in the code.

aldenml: Good, that was my conclusion too. If that's the case for 1), then what do you think about this?
- Attach the node's id to the `dht_immutable_item` structure when inserted, and that way I can perform a "virtual" pseudo-partitioning while looking for the last one to remove it. (I just mean to trick the comparator that for all the items stored with an id different from the current one in consideration, returns `-1`)

The downside is the extra memory.

ssiloti: I don't think it makes sense to partition the item storage by the id of the node which stored it. I would suggest having `dht_default_storage_constructor` take a reference to a distance function (probably as a `std::function`) rather than a reference to a node id. The distance function would compute the minimum distance from a given key to any of the client's node ids. The implementation should probably be as a member function of `dht_tracker`.

aldenml: Well, why do you think it doesn't make sense? Don't your reasoning in 1) implies a natural (almost-)partitioning of the incoming data?

Regarding the `dht_default_storage` implementation detail you mention, if this is precisely the "default implementation" why not put here the criteria to discard the elements? Why computing the minimum distance to any client's node ids is better, if that probably takes as an "effective" reference id, one that not properly represent the one from the "DHT network" they are coming(are requested)?

ssiloti: 1. It would consume a significant amount of memory.
2. It partially defeats the purpose of having unified storage.
3. It fails miserably if a node's id changes due to a change of IP. The existing code has this problem too.

I honestly cant figure out what you're trying to say in your second paragraph.

aldenml: Re:
1- Yes, memory is an issue
2- I'm just talking about a "virtual/logical" partitioning in the inner working of the comparator to remove the "last" one, in no way a hard partition that prevents an unified storage that can share the information to service requests from multiple DHTs.
3- You are right (this a good point)

The second paragraph is: Why your idea is better? and why the need to move the logic outside of the `dht_default_storage`?

ssiloti: There are a few ways of handling the fact that a node's id can change at any time:

1. Keep a copy of the node ids in dht_storage and provide a function to update them when an id changes. This creates duplicate storage which is undesirable both because it wastes memory and it runs the risk of the storage's id list getting out of sync with the actual ids.
2. Keep a reference to a data structure (likely a `std::vector<sha1_hash>`) which the owner of the storage instance is responsible for keeping up to date. Since we don't maintain such a list already, this has all the same problems as option 1.
3. Keep a reference to a function which has the necessary knowledge to reach into the existing data structures where the node id's are stored (i.e. `dht::node`). This is the best approach because it doesn't require duplicate storage and maintains loose coupling between the storage and dht node implementations.

aldenml: I think this (option 3) could make difficult the life of other bindings (yes... java :D) but I will give it a try in code

aldenml:
Hi @arvidn you can can cancel all builds of this PR, I just needed to push to continue the review.

@ssiloti I tried to follow your comments, but I found what I believe is a safe way to keep node ids in sync. I will try to keep the dht_storage extension point as swig-friendly as possible but I'm clear that this needs to align to the libtorrent's style.

I left to big "//HELP HERE" at `remove_least_important_immutable_item` and `remove_least_important_mutable_item`, I trust your reasoning of what should be a good criteria to remove an item given more than one node id.

include/libtorrent/kademlia/dht_storage.hpp
```diff
@@ -88,6 +88,14 @@ namespace dht
 		virtual size_t num_peers() const = 0;
 #endif
 
+		// PENDING REVIEW
+		// This method notifies the list of all node's ids of each
```
arvidn: I'm a bit of a pedantic when it comes to terminology. and in C++ it's "member function" not "method"

include/libtorrent/kademlia/dht_storage.hpp
```diff
@@ -88,6 +88,14 @@ namespace dht
...
+		// DHT running inside libtorrent. It's advisable that the
+		// concrete implementation keeps a copy of this list for an
+		// eventual prioritization when deleting an element to make
+		// room for a new one.
+		virtual void update_node_ids(std::vector<sha1_hash> const& ids) = 0;
```
arvidn: despite the fact that sha1_hash is the same type as node_id, I think there's some value in making the distinction. The dht namespace should have a typedef called "node_id"

include/libtorrent/kademlia/dht_tracker.hpp
```diff
@@ -155,6 +156,7 @@ namespace libtorrent { namespace dht
 		bdecode_node m_msg;
 
 		counters& m_counters;
+		dht_storage_interface* m_storage;
```
arvidn: a part of me feels like it would be more appropriate to make this a reference. a reference cannot be rebound and cannot be null, which I think fits the semantics of this pointer.

include/libtorrent/kademlia/node.hpp
```diff
@@ -103,7 +103,7 @@ class TORRENT_EXTRA_EXPORT node : boost::noncopyable
 		, libtorrent::dht_settings const& settings, node_id nid
 		, dht_observer* observer, counters& cnt
 		, std::map<std::string, node*> const& nodes
-		, dht_storage_constructor_type storage_constructor = dht_default_storage_constructor);
+		, dht_storage_interface* storage = NULL);
```
arvidn: oh, maybe it can be null after all? (which spells "nullptr" these days btw)

include/libtorrent/kademlia/node.hpp
```diff
@@ -285,7 +285,7 @@ class TORRENT_EXTRA_EXPORT node : boost::noncopyable
 	udp_socket_interface* m_sock;
 	counters& m_counters;
 
-	boost::scoped_ptr<dht_storage_interface> m_storage;
+	dht_storage_interface* m_storage;
```
arvidn: a reference also makes it obvious that there's no owning relationship. ideally a raw pointer would also never mean owning, but I imagine a lot of c++ programmers have a reflex of worrying about raw pointers and ownership

src/kademlia/dht_storage.cpp
```diff
@@ -565,13 +574,26 @@ namespace
 					++i;
 			}
 		}
+
+		void remove_least_important_immutable_item()
+		{
+			//HELP HERE!
```
arvidn: I think a reasonable start would be to take the code from immutable_item_comparator::operator() and put it in here. In fact, this logic should probably be identical for immutable and mutale (as steven pointed out). Perhaps these two functions could be a single one, that picks the best candidate for removal from a container (in order to work for both mutable and immutable, it may have to be a template). These functions aren't part of the interface though, right? They aren't meant to be virtual functions.

Basically what immutable_item_comparator::operator() does is that it looks at the distance of an item from our node ID, weighing it together with the number of unique IPs that have announce it (which is meant to represent popularity) and compare those. So, it prioritizes items close to our ID, because those are the ones we're supposed to store, but it also has an affinity to store popular items, to make it harder to attack and overwrite popular items.

In the world of multiple node IDs, instead of calculating the distance to our node ID, it would have to take the minimum of all the distances to all of our node IDs.

test/test_dht_storage.cpp
```diff
@@ -76,11 +76,10 @@ namespace
 
 	bool g_storage_constructor_invoked = false;
 
-	dht_storage_interface* dht_custom_storage_constructor(sha1_hash const& id
-		, dht_settings const& settings)
+	dht_storage_interface* dht_custom_storage_constructor(dht_settings const& settings)
 	{
 		g_storage_constructor_invoked = true;
-		return dht_default_storage_constructor(id, settings);
+		return dht_default_storage_constructor(settings);
```
arvidn: a part of me also feel a bit uneasy about being able to construct a storage without any node IDs. this feels like a two-stage initialization.

test/test_dht_storage.cpp
```diff
@@ -121,20 +120,21 @@ TORRENT_TEST(announce_peer)
...
+	node_id id(0);
+	boost::scoped_ptr<dht_storage_interface> s(dht_default_storage_constructor(sett));
 	TEST_CHECK(s.get() != NULL);
 
 	entry item;
 	bool r = s->get_immutable_item(n4, item);
 	TEST_CHECK(!r);
 
-	s->put_immutable_item(n4, "123", 3, address::from_string("124.31.75.21"));
+	s->put_immutable_item(n4, id, "123", 3, address::from_string("124.31.75.21"));
```
arvidn: as pointed out, the ID here is not a property of the value we're putting. it's inherently a property of all the nodes we're running

aldenml:
Hi @arvidn, I think I addressed all your comments but one. The current object-creation-flow does not allow for passing the ids at the storage creation time, and I think it makes sense, since nodes can be theoretically created/destructed but the "shared" storage can be still there or persisted.

I need some particular review in the new function `min_distance_exp`. I'm not confortable with the dynamic allocation `distances`, I have plans to make micro optimizations for the case of `1` and `2`, but in the general case, do you have any suggestion to make it more stack memory friendly?

src/kademlia/dht_storage.cpp
```diff
@@ -189,9 +201,8 @@ namespace
 
 	public:
 
-		dht_default_storage(sha1_hash const& id, dht_settings const& settings)
-			: m_id(id)
-			, m_settings(settings)
+		dht_default_storage(dht_settings const& settings)
+			: m_settings(settings)
 		{
 			memset(&m_counters, 0, sizeof(m_counters));
```
arvidn: unrelated to this review. this looks a bit suspicious. we don't own m_counters (it's a reference) but we still just erase it's storage. It should probably have a clear function on it instead that we can call here. If it's just a plain array, it should probably be a std::array<> that we can call fill() on here

src/kademlia/dht_storage.cpp
```diff
@@ -154,15 +154,15 @@ namespace
...
 
 		bool operator() (std::pair<node_id, dht_immutable_item> const& lhs
 			, std::pair<node_id, dht_immutable_item> const& rhs) const
 		{
-			int l_distance = distance_exp(lhs.first, m_our_id);
-			int r_distance = distance_exp(rhs.first, m_our_id);
+			int l_distance = min_distance_exp(lhs.first, m_node_ids);
+			int r_distance = min_distance_exp(rhs.first, m_node_ids);
 
 			// this is a score taking the popularity (number of announcers) and the
```
arvidn: I see, this isn't distance really, it's score.. so maybe the lowest score is the correct one then

arvidn: is there a test for this? :P

src/kademlia/dht_storage.cpp
```diff
@@ -178,9 +178,21 @@ namespace
 		// explicitly disallow assignment, to silence msvc warning
 		immutable_item_comparator& operator=(immutable_item_comparator const&);
 
-		node_id const& m_our_id;
+		std::vector<node_id> const& m_node_ids;
```
arvidn: this is a reference, and it's just pointing to the existing member of the storage. I don't see any heap allocation here.

src/kademlia/dht_storage.cpp
```diff
@@ -425,13 +436,8 @@ namespace
...
 				if (int(m_mutable_table.size()) >= m_settings.max_dht_items)
 				{
-					// delete the least important one (i.e. the one
-					// the fewest peers are announcing)
-					dht_mutable_table_t::iterator j = std::min_element(m_mutable_table.begin()
-						, m_mutable_table.end()
-						, [] (dht_mutable_table_t::value_type const& lhs
-							, dht_mutable_table_t::value_type const& rhs)
-						{ return lhs.second.num_announcers < rhs.second.num_announcers; });
+					dht_mutable_table_t::const_iterator j = pick_least_important_item(
```
arvidn: you can also use "auto j = ..." :)

src/kademlia/node_id.cpp
```diff
@@ -70,6 +70,19 @@ int distance_exp(node_id const& n1, node_id const& n2)
 	return (std::max)(159 - distance(n1, n2).count_leading_zeroes(), 0);
 }
 
+int min_distance_exp(node_id const& n1, std::vector<node_id> const& ids)
+{
+	if (ids.empty()) return 0;
+
+	std::vector<int> distances;
```
arvidn: oh, I see, this is the heap allocation..

src/kademlia/node_id.cpp
```diff
@@ -70,6 +70,19 @@ int distance_exp(node_id const& n1, node_id const& n2)
...
+	distances.reserve(ids.size());
+	std::transform(ids.begin()
+			, ids.end()
+			, std::back_inserter(distances)
+			, [&n1](node_id const& node_id) { return distance_exp(n1, node_id); });
+	return *std::min_element(distances.begin(), distances.end());
```
arvidn: In the future these operations can be chained, but there's no good support for that in the standard library today. boost has a transform_iterator which you could use. but it may be simpler to just write the loop manually too (which can still be pretty nice with the new range-for syntax)

aldenml:
@arvidn rebased, accommodated considering your latest comments and added two tests

arvidn:
cool, I'll take a look tomorrow. btw, it's easier to review if you don't rebase, because then I will be able to see the difference since I last reviewed.

aldenml:
Hi @arvidn, just a ping here

src/kademlia/dht_storage.cpp
```diff
@@ -189,11 +199,10 @@ namespace
...
 	public:
 
-		dht_default_storage(sha1_hash const& id, dht_settings const& settings)
-			: m_id(id)
-			, m_settings(settings)
+		dht_default_storage(dht_settings const& settings)
+			: m_settings(settings)
 		{
-			memset(&m_counters, 0, sizeof(m_counters));
+			m_counters.reset();
```
arvidn: thanks

src/kademlia/dht_storage.cpp
```diff
@@ -568,10 +572,17 @@ namespace
...
-	, dht_settings const& settings)
+void dht_storage_counters::reset()
+{
+	torrents = 0;
+	peers = 0;
+	immutable_data = 0;
+	mutable_data = 0;
+}
+
+dht_storage_interface* dht_default_storage_constructor(dht_settings const& settings)
```
arvidn: not for this patch, but at some point this function should return a unique_ptr<> to the storage, and it should probably be held by a unique_ptr<> in the session too

src/kademlia/dht_tracker.cpp
```diff
@@ -208,19 +215,19 @@ namespace libtorrent { namespace dht
 #ifndef TORRENT_NO_DEPRECATE
 	void dht_tracker::dht_status(session_status& s)
 	{
-		m_dht.status(s);
+		m_dht.status(s); //TODO: What to do with m_dht6?
```
arvidn: also not for this patch, but I imagine that the status() call should be strictly additive, and we would call it once for each node

src/kademlia/node_id.cpp
```diff
@@ -70,6 +71,25 @@ int distance_exp(node_id const& n1, node_id const& n2)
...
 
+int min_distance_exp(node_id const& n1, std::vector<node_id> const& ids)
+{
+	// specialized for cases of 0, 1 and 2 for performance reasons
+	if (ids.size() == 0) return 0;
+	if (ids.size() == 1) return distance_exp(n1, ids[0]);
+	if (ids.size() == 2)
+		return std::min(distance_exp(n1, ids[0]), distance_exp(n1, ids[1]));
+
+	int min = std::numeric_limits<int>::max();
```
arvidn: I guess this passed in msvc-15.. traditionally msvc has had problems with code like this by introducing a min and max macro from windows.h. the traditional solution is to not use identifiers called min or max and to say (std::numeric_limits<int>::max)(). But I guess this works now, maybe I should start cleaning up all the other places where I use the parenthesis trick :P

src/kademlia/node_id.cpp
```diff
@@ -70,6 +71,25 @@ int distance_exp(node_id const& n1, node_id const& n2)
 	return (std::max)(159 - distance(n1, n2).count_leading_zeroes(), 0);
 }
 
+int min_distance_exp(node_id const& n1, std::vector<node_id> const& ids)
+{
+	// specialized for cases of 0, 1 and 2 for performance reasons
+	if (ids.size() == 0) return 0;
```
arvidn: ids should never be empty, right? should there be an assert here maybe?

src/kademlia/node_id.cpp
```diff
@@ -70,6 +71,25 @@ int distance_exp(node_id const& n1, node_id const& n2)
...
+	if (ids.size() == 1) return distance_exp(n1, ids[0]);
+	if (ids.size() == 2)
+		return std::min(distance_exp(n1, ids[0]), distance_exp(n1, ids[1]));
+
+	int min = std::numeric_limits<int>::max();
+	for (const auto &node_id : ids)
+	{
+		int d = distance_exp(n1, node_id);
```
arvidn: you could also say: min = std::min(min, distance_exp(n1, node_id));

src/kademlia/node_id.cpp
```diff
@@ -70,6 +71,25 @@ int distance_exp(node_id const& n1, node_id const& n2)
 	return (std::max)(159 - distance(n1, n2).count_leading_zeroes(), 0);
 }
 
+int min_distance_exp(node_id const& n1, std::vector<node_id> const& ids)
+{
+	// specialized for cases of 0, 1 and 2 for performance reasons
+	if (ids.size() == 0) return 0;
+	if (ids.size() == 1) return distance_exp(n1, ids[0]);
+	if (ids.size() == 2)
+		return std::min(distance_exp(n1, ids[0]), distance_exp(n1, ids[1]));
```
arvidn: these special cases, is there a good reason to have them? would you imagine that they are significantly more efficient than the loop? (there's also a cost of checking the size by having them)

src/session_impl.cpp
```diff
@@ -5619,9 +5621,16 @@ namespace aux {
...
-		if (!m_dht) return;
-		m_dht->stop();
-		m_dht.reset();
+		if (m_dht)
+		{
+			m_dht->stop();
+			m_dht.reset();
+		}
+
+		if (m_dht_storage)
```
arvidn: I don't think you should check if m_dht_storage is set. just call reset() unconditionally

test/test_dht.cpp
```diff
@@ -508,10 +508,11 @@ void do_test_dht(address(&rand_addr)())
 	mock_socket s;
 	obs observer;
 	counters cnt;
+	boost::scoped_ptr<dht_storage_interface> dht_storage(dht_default_storage_constructor(sett));
```
arvidn: I would prefer std::unique_ptr over boost::scoped_ptr.. but I will have to replace them all in one big patch at some point

test/test_dht.cpp
```diff
@@ -2692,6 +2704,39 @@ TORRENT_TEST(node_id_bucket_distribution)
 	}
 }
 
+TORRENT_TEST(node_id_min_distance_exp)
+{
+	node_id n1 = to_hash("0000000000000000000000000000000000000002");
+	node_id n2 = to_hash("0000000000000000000000000000000000000004");
+	node_id n3 = to_hash("0000000000000000000000000000000000000008");
```
arvidn: I think it's nice to declare n1, n2 and n3 as const. (and I prefer "continental const placement", i.e. to the right of the type it applies to: "node_id const n1 =")

test/test_dht.cpp
```diff
@@ -2692,6 +2704,39 @@ TORRENT_TEST(node_id_bucket_distribution)
...
+
+	std::vector<node_id> ids;
+
+	TEST_EQUAL(min_distance_exp(sha1_hash::min(), ids), 0);
```
arvidn: as I pointed out, this seems like a questionable use case. does it happen in real-life?

test/test_dht_storage.cpp
```diff
@@ -308,5 +307,55 @@ TORRENT_TEST(mutable_item_limit)
...
 
+TORRENT_TEST(update_node_ids)
+{
+	dht_settings sett = test_settings();
+	boost::scoped_ptr<dht_storage_interface> s(dht_default_storage_constructor(sett));
+	TEST_CHECK(s.get() != NULL);
+
+	node_id n1 = to_hash("0000000000000000000000000000000000000200");
+	node_id n2 = to_hash("0000000000000000000000000000000000000400");
+	node_id n3 = to_hash("0000000000000000000000000000000000000800");
```
arvidn: some more opportunities to make things const

arvidn:
sorry for the delay. would you mind addressing some of my comments (or arguing against them) in a separate patch?

aldenml:
Sure! I will address them

PR #767, opened 2016-05-29
Fix compilation with cmake

evsh:


arvidn:
thanks. I'm not all that familiar with cmake, I take it you feel confident with these changes? (I think most are obvious, but not all). We should probably be a cmake build in the travis build matrix too.

evsh:
Thank you. Well, I hadn't tried it in a live install yet. Will add python bindings to the cmake build and live test everything then.

PR #768, opened 2016-05-29
remove dependency on libtommath and use boost.multiprecision instead

arvidn:


LordNyriox:
@arvidn:  Are any of these feature changes being considered for back-porting to RC_1_1?  

Thanks.

Athorcis:
i don't think it's useful since the segfault fix has already been backported to RC_1_1 and RC_1_0, this pr is not a fix but rather an improvement

arvidn:
@ssiloti does this affect your reputation plugin? I think it's a step in the right direction to not pull in libtommath and hack around in its source code. the c++ interface of boost.multiprecision is a lot nicer too, and requires less wrapping

include/libtorrent/bt_peer_connection.hpp
```diff
@@ -402,6 +397,7 @@ namespace libtorrent
 		// used during an encrypted handshake then moved
 		// into m_enc_handler if rc4 encryption is negotiated
 		// otherwise it is destroyed when the handshake completes
+		// TODO: 2 this could probably be a std::unique_ptr
```
ssiloti: I checked this before and it has to be a shared_ptr because two references to it are held in `m_enc_handler`, one for encrypting outgoing traffic and one for decrypting incoming.

ssiloti:
This doesn't affect the reputation plugin at all. I deliberately kept the crypto in the reputation plugin independent of pe crypto.

PR #769, opened 2016-05-29
fix missing includes

ssiloti:


arvidn:
hm.. I thought these were fixed by this https://github.com/arvidn/libtorrent/pull/767/files
but your PR doesn't have a conflict...

ssiloti:
Oops my bad, forgot to fetch.

PR #770, opened 2016-05-29
scalable thread pool

ssiloti:
I've kept the generic/hash thread pool sizing logic the same for now since I expect it to change anyways before we ship this.

### DIFF
--- a/include/libtorrent/disk_io_thread.hpp
+++ b/include/libtorrent/disk_io_thread.hpp
@@ -38,6 +38,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/allocator.hpp"
 #include "libtorrent/io_service.hpp"
 #include "libtorrent/sliding_average.hpp"
+#include "libtorrent/disk_io_thread_pool.hpp"
 #include "libtorrent/disk_io_job.hpp"
 #include "libtorrent/disk_job_pool.hpp"
 #include "libtorrent/block_cache.hpp"
@@ -294,7 +295,7 @@ namespace libtorrent
 		~disk_io_thread();
 
 		void set_settings(settings_pack const* sett, alert_manager& alerts);
-		void set_num_threads(int i, bool wait = true);
+		void set_num_threads(int i);
 
 		void abort(bool wait);
 
@@ -385,8 +386,7 @@ namespace libtorrent
 			hasher_thread
 		};
 
-		void thread_fun(int thread_id, thread_type_t type
-			, boost::shared_ptr<io_service::work> w);
+		void thread_fun(thread_type_t type, io_service::work w);
 
 		virtual file_pool& files() override { return m_file_pool; }
 
@@ -434,6 +434,31 @@ namespace libtorrent
 
 	private:
 
+		struct job_queue : pool_thread_interface
+		{
+			job_queue(disk_io_thread& owner, thread_type_t type)
+				: m_owner(owner), m_type(type)
+			{}
+
+			virtual void notify_all() override
+			{
+				m_job_cond.notify_all();
+			}
+
+			virtual void thread_fun(io_service::work work) override
+			{ m_owner.thread_fun(m_type, work); }
+
+			disk_io_thread& m_owner;
+			thread_type_t const m_type;
+
+			// used to wake up the disk IO thread when there are new
+			// jobs on the job queue (m_queued_jobs)
+			std::condition_variable m_job_cond;
+
+			// jobs queued for servicing
+			jobqueue_t m_queued_jobs;
+		};
+
 		enum return_value_t
 		{
 			// the do_* functions can return this to indicate the disk
@@ -444,6 +469,10 @@ namespace libtorrent
 			// queue and try again later
 			retry_job = -201
 		};
+
+		// returns true if the thread should exit
+		static bool wait_for_job(job_queue& jobq, disk_io_thread_pool& threads
+			, std::unique_lock<std::mutex>& l);
 
 		void add_completed_job(disk_io_job* j);
 		void add_completed_jobs(jobqueue_t& jobs);
@@ -515,10 +544,11 @@ namespace libtorrent
 		void immediate_execute();
 		void abort_jobs();
 
-		// this is a counter which is atomically incremented
-		// by each thread as it's started up, in order to
-		// assign a unique id to each thread
-		std::atomic<int> m_num_threads;
+		// returns the maximum number of threads
+		// the actual number of threads may be less
+		int num_threads() const;
+		job_queue& queue_for_job(disk_io_job* j);
+		disk_io_thread_pool& pool_for_job(disk_io_job* j);
 
 		// set to true once we start shutting down
 		std::atomic<bool> m_abort;
@@ -528,8 +558,16 @@ namespace libtorrent
 		// shutting down. This last thread is responsible for cleanup
 		std::atomic<int> m_num_running_threads;
 
-		// the actual threads running disk jobs
-		std::vector<std::thread> m_threads;
+		// std::mutex to protect the m_generic_io_jobs and m_hash_io_jobs lists
+		mutable std::mutex m_job_mutex;
+
+		// most jobs are posted to m_generic_io_jobs
+		// but hash jobs are posted to m_hash_io_jobs if m_hash_threads
+		// has a non-zero maximum thread count
+		job_queue m_generic_io_jobs;
+		disk_io_thread_pool m_generic_threads;
+		job_queue m_hash_io_jobs;
+		disk_io_thread_pool m_hash_threads;
 
 		aux::session_settings m_settings;
 
@@ -578,22 +616,6 @@ namespace libtorrent
 		// posted on this in order to have them execute in
 		// the main thread.
 		io_service& m_ios;
-
-		// used to wake up the disk IO thread when there are new
-		// jobs on the job queue (m_queued_jobs)
-		std::condition_variable m_job_cond;
-
-		// std::mutex to protect the m_queued_jobs list
-		mutable std::mutex m_job_mutex;
-
-		// jobs queued for servicing
-		jobqueue_t m_queued_jobs;
-
-		// when using more than 2 threads, this is
-		// used for just hashing jobs, just for threads
-		// dedicated to do hashing
-		std::condition_variable m_hash_job_cond;
-		jobqueue_t m_queued_hash_jobs;
 
 		// used to rate limit disk performance warnings
 		time_point m_last_disk_aio_performance_warning;
--- a/include/libtorrent/disk_io_thread_pool.hpp
+++ b/include/libtorrent/disk_io_thread_pool.hpp
@@ -1,0 +1,143 @@
+/*
+
+Copyright (c) 2005-2016, Arvid Norberg, Steven Siloti
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+* Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+* Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in
+the documentation and/or other materials provided with the distribution.
+* Neither the name of the author nor the names of its
+contributors may be used to endorse or promote products derived
+from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+
+*/
+
+#ifndef TORRENT_DISK_IO_THREAD_POOL
+#define TORRENT_DISK_IO_THREAD_POOL
+
+#include "libtorrent/config.hpp"
+#include "libtorrent/export.hpp"
+#include "libtorrent/deadline_timer.hpp"
+#include "libtorrent/io_service_fwd.hpp"
+#include "libtorrent/error_code.hpp"
+
+#include <thread>
+#include <mutex>
+#include <atomic>
+#include <functional>
+
+namespace libtorrent
+{
+	struct pool_thread_interface
+	{
+		virtual ~pool_thread_interface() {}
+
+		virtual void notify_all() = 0;
+		virtual void thread_fun(io_service::work /* work */) = 0;
+	};
+
+	// this class implements the policy for creating and destroying I/O threads
+	// threads are created when job_queued is called to signal the arrival of
+	// new jobs
+	// once a minute threads are destroyed if at least one thread has been
+	// idle for the entire minute
+	// the pool_thread_interface is used to spawn and notify the worker threads
+	struct TORRENT_EXTRA_EXPORT disk_io_thread_pool
+	{
+		disk_io_thread_pool(pool_thread_interface& thread_iface
+			, io_service& ios);
+		~disk_io_thread_pool();
+
+		// set the maximum number of I/O threads which may be running
+		// the actual number of threads will be <= this number
+		void set_max_threads(int const i);
+		void abort(bool wait);
+		int max_threads() const { return m_max_threads; }
+
+		// thread_idle, thread_active, and job_queued are NOT thread safe
+		// all calls to them must be serialized
+		// it is expected that they will be called while holding the
+		// job queue mutex
+
+		// theese functions should be called by the thread_fun to signal its state
+		// threads are considered active when they are started so thread_idle should
+		// be called first
+		// these calls are not thread safe
+		void thread_idle() { ++m_num_idle_threads; }
+		void thread_active();
+
+		// check if there is an outstanding request for I/O threads to stop
+		// this is a weak check, if it returns true try_thread_exit may still
+		// return false
+		bool should_exit() { return m_threads_to_exit > 0; }
+		// this should be the last function an I/O thread calls before breaking
+		// out of its service loop
+		// if it returns true then the thread MUST exit
+		// if it returns false the thread should not exit
+		bool try_thread_exit(std::thread::id id);
+
+		// get the thread id of the first thread in the internal vector
+		// since this is the first thread it will remain the same until the first
+		// thread exits
+		// it can be used to trigger maintainance jobs which should only run on one thread
+		std::thread::id first_thread_id();
+		int num_threads()
+		{
+			std::lock_guard<std::mutex> l(m_mutex);
+			return int(m_threads.size());
+		}
+
+		// this should be called whenever new jobs are queued
+		// queue_size is the current size of the job queue
+		// not thread safe
+		void job_queued(int queue_size);
+
+	private:
+		void reap_idle_threads(error_code const& ec);
+
+		// the caller must hold m_mutex
+		void stop_threads(int num_to_stop);
+
+		pool_thread_interface& m_thread_iface;
+
+		std::atomic<int> m_max_threads;
+		// the number of threads the reaper decided should exit
+		std::atomic<int> m_threads_to_exit;
+
+		// must hold m_mutex to access
+		bool m_abort;
+
+		std::atomic<int> m_num_idle_threads;
+		// the minimum number of idle threads seen since the last reaping
+		std::atomic<int> m_min_idle_threads;
+
+		// ensures thread creation/destruction is atomic
+		std::mutex m_mutex;
+
+		// the actual threads running disk jobs
+		std::vector<std::thread> m_threads;
+
+		// timer to check for and reap idle threads
+		deadline_timer m_idle_timer;
+	};
+} // namespace libtorrent
+
+#endif
--- a/simulation/test_thread_pool.cpp
+++ b/simulation/test_thread_pool.cpp
@@ -1,0 +1,215 @@
+/*
+
+Copyright (c) 2016, Steven Siloti
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+* Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+* Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in
+the documentation and/or other materials provided with the distribution.
+* Neither the name of the author nor the names of its
+contributors may be used to endorse or promote products derived
+from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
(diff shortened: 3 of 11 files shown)


include/libtorrent/disk_io_thread.hpp
```diff
@@ -434,6 +435,33 @@ namespace libtorrent
 
 	private:
 
+		friend struct job_queue;
```
arvidn: I was under the impression that nested classes automatically were friends

include/libtorrent/disk_io_thread.hpp
```diff
@@ -528,8 +557,13 @@ namespace libtorrent
 		// shutting down. This last thread is responsible for cleanup
 		std::atomic<int> m_num_running_threads;
 
-		// the actual threads running disk jobs
-		std::vector<std::thread> m_threads;
+		// std::mutex to protect the m_generic_io_jobs and m_hash_io_jobs lists
+		mutable std::mutex m_job_mutex;
```
arvidn: is there a reason why this mutex isn't split in two as well? inside job_queue

src/disk_io_thread.cpp
```diff
@@ -3179,10 +3144,14 @@ namespace libtorrent
...
-		, boost::shared_ptr<io_service::work> w)
+	void disk_io_thread::thread_fun(thread_type_t type
+		, std::shared_ptr<io_service::work> w)
 	{
-		DLOG("started disk thread %d\n", int(thread_id));
+		std::thread::id thread_id = std::this_thread::get_id();
+		std::stringstream thread_id_str;
+		thread_id_str << thread_id;
+
+		DLOG("started disk thread %s\n", thread_id_str.str().c_str());
```
arvidn: would you mind wrapping the stringstream and the loggin in #if DEBUG_DISK_THREAD?

src/disk_io_thread.cpp
```diff
@@ -3302,6 +3288,27 @@ namespace libtorrent
...
 	}
 
+	int disk_io_thread::num_threads() const
+	{
+		return m_generic_threads.max_threads() + m_hash_threads.max_threads();
+	}
+
+	disk_io_thread::job_queue& disk_io_thread::queue_for_job(disk_io_job* j)
+	{
+		if (m_hash_threads.max_threads() > 0 && j->action == disk_io_job::hash)
```
arvidn: if max_threads() == 1, is there really a hasher thread?

src/disk_io_thread_pool.cpp
```diff
@@ -0,0 +1,207 @@
...
+			if (wait)
+			{
+				// must release m_mutex to avoid a deadlock if the thread
+				// tries to aquire it
+				l.unlock();
+				t.join();
+				l.lock();
+			}
+			else
+				t.detach();
```
arvidn: hm.. did there used to be a detach? is this ever invoked? it feels a bit scary

src/disk_io_thread_pool.cpp
```diff
@@ -0,0 +1,207 @@
...
+		int current_min = m_min_idle_threads;
+		while (m_num_idle_threads < current_min
+			&& !m_min_idle_threads.compare_exchange_weak(current_min, m_num_idle_threads));
+	}
+
+	bool disk_io_thread_pool::try_thread_exit(std::thread::id id)
+	{
+		int to_exit = m_threads_to_exit;
+		while (to_exit > 0 &&
+			!m_threads_to_exit.compare_exchange_weak(to_exit, to_exit - 1));
```
arvidn: perhaps a simpler way to do this would be to unconditionally decrement m_threads_to_exit, and only act on it if the return value was > 0. I suppose you would still need to check if it's > 0 first, to avoid underflow. something like:

bool exit = false;
if (m_threads_to_exit > 0) {
   exit = (threads_to_exit--) > 0;
}

src/disk_io_thread_pool.cpp
```diff
@@ -0,0 +1,207 @@
...
+		{
+			std::unique_lock<std::mutex> l(m_mutex);
+			if (!m_abort)
+			{
+				auto new_end = std::remove_if(m_threads.begin(), m_threads.end()
+					, [id, &l](std::thread& t)
+				{
+					if (t.get_id() == id)
+					{
+						t.detach();
```
arvidn: what's scary with detach is that _any_ memory access after this may be undefined behavior. In this case, we detach ourself, right? when we access m_threads later, technically it may have been destructed. and whatever happens when the call to try_thread_exit() returns

arvidn: the same thing goes for releasing m_mutex. Ideally we would join() all threads at some point later. maybe the threads that have started to shut down could be moved into an m_undead thread queue, that we join and clear out at some later point

include/libtorrent/disk_io_thread.hpp
ssiloti: I could have sworn they weren't, but when I went to find the relevant quote in the spec I discovered that you are right! Checking further, it seems this rule was changed in C++11.

include/libtorrent/disk_io_thread.hpp
ssiloti: It could be, but I didn't see a good reason to do so.

src/disk_io_thread.cpp
ssiloti: Yes, this is checking the max specifically for the hasher thread pool, not the overall max.

src/disk_io_thread_pool.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+#include "libtorrent/disk_io_thread_pool.hpp"
+#include "libtorrent/disk_io_thread.hpp"
+#include <functional>
+
+namespace
+{
+	enum
+	{
+		reap_idle_threads_interval = 60
```
arvidn: this could be a constexpr chrono::seconds and you wouldn't need to cast it later

src/disk_io_thread_pool.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+			// this keeps the io_service::run() call blocked from returning.
+			// When shutting down, it's possible that the event queue is drained
+			// before the disk_io_thread has posted its last callback. When this
+			// happens, the io_service will have a pending callback from the
+			// disk_io_thread, but the event loop is not running. this means
+			// that the event is destructed after the disk_io_thread. If the
+			// event refers to a disk buffer it will try to free it, but the
+			// buffer pool won't exist anymore, and crash. This prevents that.
+			std::shared_ptr<io_service::work> work =
```
arvidn: technically, this could probably be just created on the stack and moved into the thread

src/disk_io_thread_pool.cpp
ssiloti: The detach is not new. The only place the detach path is invoked is `session_impl::abort_stage2`.

src/disk_io_thread_pool.cpp
```diff
@@ -0,0 +1,207 @@
...
+{
+	enum
+	{
+		reap_idle_threads_interval = 60
+	};
+}
+
+namespace libtorrent
+{
+	disk_io_thread_pool::disk_io_thread_pool(pool_thread_interface& thread_iface
```
arvidn: can you think of any ways to test disk_io_thread_pool? Perhaps a simulation would be appropriate, since it depends on time passing

include/libtorrent/disk_io_thread.hpp
arvidn: I imagine you lower the risk of contention.

src/disk_io_thread.cpp
arvidn: ah, of course

src/disk_io_thread_pool.cpp
ssiloti: My expectation is that the instance of `io_service::work` held by `disk_io_thread::thread_fun` will keep the `disk_io_thread` instance from being destructed. Based on the existing comments this seems to have been the exact reason why it was added.

src/disk_io_thread_pool.cpp
ssiloti: I really wanted to avoid having `m_threads_to_exit` become negative, but now that I look back on how it's used I see that it doesn't matter.

src/disk_io_thread_pool.cpp
arvidn: I see. that makes sense

src/disk_io_thread_pool.cpp
arvidn: yeah, once I realized you actually increment in also (and not just set it), I wasn't sure it was a good idea anymore.

src/disk_io_thread_pool.cpp
ssiloti: At least as of boost 1.60 `io_service::work` is not movable, but that's because it defines a copy constructor, which makes me wonder why it was heap allocated in the first place.

arvidn: ah, I see. it used to be a boost::optional<>, but at some point it I ended up having very few dependencies on boost::optional and got rid of it entirely, and heap allocated it instead, for simplicity. It seems like now that it's tied to a thread, it doesn't need to be anymore

src/disk_io_thread_pool.cpp
ssiloti: Yeah I can do some simulator tests since testing reaping of idle threads needs time manipulation.

ssiloti:
A quick progress update: I've written some simulator tests but I'm running into problems due to lack of thread safety in libsimulator so there's going to be an update there as well.

ssiloti:
Of course I hit send a moment too soon. Now I realize the bits I'm having problems with are not thread safe with real asio either so I'll need some more locking in the thread pool.

arvidn:
hm.. yeah. actually, it may be a bit tricky to have anything significant be done in separate threads in the simulator. it assumes every event is channeled and dispatched through the simulator object. If the threads in the pool are just there and tested for being removed within a certain amount of time, that would probably work. but to make sure it _doesn't_ remove threads once fully loaded, that might be trickier.

ssiloti:
Please let me know before you merge this so I can squash some of the commits

include/libtorrent/disk_io_thread.hpp
```diff
@@ -528,8 +558,13 @@ namespace libtorrent
...
+
+		job_queue m_generic_io_jobs;
```
arvidn: maybe it would be good with a comment here, explaining that different kinds of jobs are posted to different pools. and maybe which jobs go to which pool, and maybe even why.

include/libtorrent/disk_io_thread_pool.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include "libtorrent/error_code.hpp"
+
+#include <thread>
+#include <mutex>
+#include <atomic>
+#include <functional>
+
+namespace libtorrent
+{
+	struct pool_thread_interface
```
arvidn: this interface is here to be able to mock the internals of the thread pool in the test, right? This name isn't quite intuitive to me. My understanding is that a class implementing this interface isn't a thread, is it? it's more of a handler maybe.. the one non-test class implementing it is called job_queue.. so is it more of a job_queue_interface?

include/libtorrent/disk_io_thread_pool.hpp
```diff
@@ -0,0 +1,133 @@
...
+	{
+		virtual ~pool_thread_interface() {}
+
+		virtual void notify_all() = 0;
+		virtual void thread_fun(io_service::work /* work */) = 0;
+	};
+
+	struct TORRENT_EXTRA_EXPORT disk_io_thread_pool
```
arvidn: could you add a comment here explaining what this class is responsible for? It looks like it manages the threads, but it doesn't exactly look like a thread pool (which i would expect you to be able post jobs to). I take it you post jobs to the pool_thread_interface object you give this.

simulation/test_thread_pool.cpp
```diff
@@ -0,0 +1,215 @@
...
+	pool.set_max_threads(3);
+	pool.job_queued(3);
+	TEST_EQUAL(pool.num_threads(), 3);
+	pool.abort(true);
+	TEST_EQUAL(pool.num_threads(), 0);
+}
+
+#if 0
+// disabled for now because io_service::work doesn't work under the simulator
+// and we need it to stop this test from exiting prematurely
```
arvidn: I imagine this may be hard to fix in the simulator, especially if the work object is supposed to be released in a separate thread. the main thread would try to fast-forward time to infinity, because it doesn't know of any more events in the future.

src/disk_io_thread.cpp
```diff
@@ -193,63 +197,25 @@ namespace libtorrent
...
 		m_abort = true;
-		if (m_num_threads == 0)
+		if (num_threads() == 0)
 		{
 			abort_jobs();
 		}
 		else
 		{
-			set_num_threads(0, wait);
+			m_generic_threads.abort(wait);
```
arvidn: are you sure we don't need to call abort_jobs() here too? to cancel the ones that haven't started yet?

src/disk_io_thread.cpp
```diff
@@ -193,63 +197,25 @@ namespace libtorrent
...
-			{
-				if (wait)
-					m_threads[j].join();
-				else
-					m_threads[j].detach();
-			}
-			m_threads.resize(m_num_threads);
-		}
+		// add one hasher thread for every three generic threads
+		m_generic_threads.set_max_threads(i - i / 4);
```
arvidn: I was fooled by this expression, thinking "i - i", that's zero! I think it would be nice to say:

int const num_hash_threads = i / 4;
int const num_generic_threads = i - num_hash_threads;
...

src/disk_io_thread.cpp
```diff
@@ -1151,8 +1119,8 @@ namespace libtorrent
 
 			TORRENT_ASSERT((j->flags & disk_io_job::in_progress) || !j->storage);
 
-			bool need_sleep = m_queued_jobs.empty();
-			m_queued_jobs.push_back(j);
+			bool need_sleep = q.m_queued_jobs.empty();
```
arvidn: nit-pick: bool const need_sleep = ...

src/disk_io_thread.cpp
```diff
@@ -3108,33 +3076,24 @@ namespace libtorrent
...
-			// immediately. If add job is called internally by the disk_io_thread,
-			// we need to defer executing it. We only want the top level to loop
-			// over the job queue (as is done below)
-			if (m_num_threads == 0 && user_add)
-			{
-				l.unlock();
-				immediate_execute();
-			}
+			l.unlock();
+			immediate_execute();
```
arvidn: it looks weird to not pass in the job to this function. will it pull the job off the queue? We're just releasing the mutex, so there's no guarantee the job is still on the queue by the time we call this function, right? perhaps you'd need to pass in the lock_guard object to let immediate_execute() release it while it's performing the job.

src/disk_io_thread.cpp
```diff
@@ -3108,33 +3076,24 @@ namespace libtorrent
...
 		}
 	}
 
 	void disk_io_thread::immediate_execute()
 	{
-		while (!m_queued_jobs.empty())
+		while (!m_generic_io_jobs.m_queued_jobs.empty())
```
arvidn: related to the comment above, is it safe to do this without holding the lock?

src/disk_io_thread.cpp
```diff
@@ -3143,10 +3102,16 @@ namespace libtorrent
 	void disk_io_thread::submit_jobs()
 	{
 		std::unique_lock<std::mutex> l(m_job_mutex);
-		if (!m_queued_jobs.empty())
-			m_job_cond.notify_all();
-		if (!m_queued_hash_jobs.empty())
-			m_hash_job_cond.notify_all();
+		if (!m_generic_io_jobs.m_queued_jobs.empty())
```
arvidn: it's almost like this could be factored into a member function of the job queue.. but maybe not

src/disk_io_thread.cpp
```diff
@@ -3174,10 +3139,44 @@ namespace libtorrent
...
+		TORRENT_ASSERT(l.owns_lock());
+		while (jobq.m_queued_jobs.empty())
+		{
+			// if the number of wanted threads is decreased,
+			// we may stop this thread
+			// when we're terminating the last thread, make sure
+			// we finish up all queued jobs first
+			if (threads.should_exit()
+				&& (jobq.m_queued_jobs.empty()
+					|| threads.num_threads() > 1)
```
arvidn: since num_threads() is the upper bound. is it possible the last two threads will see max_threads() as 2 and should_exit()? (i.e. is there an interleaving that could result in both threads seeing that basically)

src/disk_io_thread.cpp
```diff
@@ -3188,34 +3187,22 @@ namespace libtorrent
...
-				// when we're terminating the last thread (id=0), make sure
-				// we finish up all queued jobs first
-				if (thread_id >= m_num_threads && !(thread_id == 0 && m_queued_jobs.size() > 0))
-				{
-					// time to exit this thread.
-					break;
-				}
-
-				j = m_queued_jobs.pop_front();
+				bool should_exit = wait_for_job(m_generic_io_jobs, m_generic_threads, l);
```
arvidn: another nitpick: bool const should_exit = ...

src/disk_io_thread_pool.cpp
```diff
@@ -0,0 +1,200 @@
...
+				});
+				TORRENT_ASSERT(new_end != m_threads.end());
+				m_threads.erase(new_end, m_threads.end());
+				if (m_threads.empty()) m_idle_timer.cancel();
+			}
+		}
+		return to_exit > 0;
+	}
+
+	std::thread::id disk_io_thread_pool::first_thread_id()
```
arvidn: could you add a comment here explaining what "first_thread" is and why it's relevant?

src/disk_io_thread_pool.cpp
```diff
@@ -0,0 +1,200 @@
...
+	{
+		std::lock_guard<std::mutex> l(m_mutex);
+		if (m_threads.empty()) return std::thread::id();
+		return m_threads.front().get_id();
+	}
+
+	void disk_io_thread_pool::job_queued(int queue_size)
```
arvidn: when is this called? the name "job_queued" sounds like it would return the size of the job queue, but it isn't doing that.
is it informing the thread pool of how many jobs are queued? and it's called "job_queued" because that's the event it's called on, another job was just queued?
I think a comment here may make sense too

src/disk_io_thread_pool.cpp
```diff
@@ -0,0 +1,200 @@
...
+			// disk_io_thread, but the event loop is not running. this means
+			// that the event is destructed after the disk_io_thread. If the
+			// event refers to a disk buffer it will try to free it, but the
+			// buffer pool won't exist anymore, and crash. This prevents that.
+			m_threads.emplace_back(&pool_thread_interface::thread_fun
+				, &m_thread_iface, io_service::work(m_idle_timer.get_io_service()));
+		}
+	}
+
+	void disk_io_thread_pool::reap_idle_threads(error_code ec)
```
arvidn: error_code is typically passed by const& (when it's an input parameter)

src/disk_io_thread_pool.cpp
```diff
@@ -0,0 +1,200 @@
...
+				, (std::max)(0, m_num_idle_threads - queue_size)));
+
+		// now start threads until we either have enough to service
+		// all queued jobs without blocking or hit the max
+		for (int i = m_num_idle_threads
+			; i < queue_size && int(m_threads.size()) < m_max_threads
+			; ++i)
+		{
+			// if this is the first thread started, start the reaper timer
+			if (m_threads.empty())
```
arvidn: I take it the number of threads will never go down to zero for an instant just to go back up to > 0 again, will it? i.e. is there a risk that multiple reaper events are created? I suppose the idle_timer would just abort the previous trigger anyway.

src/disk_io_thread_pool.cpp
```diff
@@ -0,0 +1,200 @@
...
+		// sample period and request that many threads to exit
+		if (ec) return;
+		std::lock_guard<std::mutex> l(m_mutex);
+		if (m_abort) return;
+		if (m_threads.size() == 0) return;
+		m_idle_timer.expires_from_now(reap_idle_threads_interval);
+		m_idle_timer.async_wait([this](error_code const& e) { reap_idle_threads(e); });
+		int min_idle = m_min_idle_threads.exchange(m_num_idle_threads);
+		if (min_idle <= 0) return;
+		stop_threads(min_idle);
```
arvidn: I guess I'm not collecting test coverage from sims, that's why this block is marked as not covered..

src/disk_io_thread_pool.cpp
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	void disk_io_thread_pool::stop_threads(int num_to_stop)
+	{
+		m_threads_to_exit = num_to_stop;
```
arvidn: what if this function is called when m_threads_to_exit is non-zero? it feels like there might be a race condition somewhere, but perhaps a benign one. whoever calls this, may think there are more threads than there are, and stop one thread too many for instance.

If it's never expected to be called when m_threads_to_exit != 0, there could be an assert

arvidn:
sorry for the delay. there's some nit-picks in there (about making variables "const" that can be) and I think  it would be good with a few more comments. also please rebase on top of master now. apparently travis and appveyor were running with asserts disabled, so I enabled that and fixed a few issues and warnings. master appears to be in a good place right now though

include/libtorrent/disk_io_thread_pool.hpp
ssiloti: It's handy for testing, but I originally added this interface in anticipation of the future changes outlined in the mmapped I/O document. I expect that we'll end up a separate thread function for at least one of the thread pools. The implementer of this interface is not a thread, as you suggest it's a collection of hooks to connect the thread pool with the worker threads.

More generally I see this interface as providing loose coupling between the thread pool and what the threads themselves do, which in theory might not even be processing a job queue. That abstraction has already leaked a bit with the job_queued function though.

src/disk_io_thread.cpp
ssiloti: As far as can tell this code duplicates the existing behavior. Both before and after these changes the last worker thread will keep running until the job queue is empty.

src/disk_io_thread.cpp
ssiloti: Now that I think about this, it's really subtle. If `pool_for_job(j).max_threads() == 0` then it implies that there are no worker threads and we are running on the network thread. Since only the network thread changes the max threads there's no opportunity for a race.

src/disk_io_thread.cpp
ssiloti: `num_threads()` gets decremented in `try_thread_exit()` so since we hold the job queue mutex for the entire operation it is atomic with respect to the other workers.

src/disk_io_thread_pool.cpp
ssiloti: There is a comment for this function in the header.

src/disk_io_thread_pool.cpp
ssiloti: There is an inherent race if a reaper event is queued just before the timer is canceled. This isn't a fatal error, if the reaper runs back-to-back it will just cause a request for all idle threads at that moment to exit. So at worst it's a minor performance hit.

src/disk_io_thread_pool.cpp
ssiloti: There is not a prerequisite that `m_threads_to_exit == 0`. While there are potential races here, most of them are indeed benign. I did find one potential problem when reviewing this. If `set_max_threads()` reduces the max below the current thread count then there is a chance that a later call to `reap_idle_threads()` might reduce `m_threads_to_exit` such that the number of threads remains above the max. I'll add a fix for this.

src/disk_io_thread.cpp
arvidn: sounds like a good comment

src/disk_io_thread_pool.cpp
arvidn: please make sure to leave some comments too

src/disk_io_thread_pool.cpp
```diff
@@ -186,9 +186,12 @@ namespace libtorrent
 		if (m_threads.size() == 0) return;
 		m_idle_timer.expires_from_now(reap_idle_threads_interval);
 		m_idle_timer.async_wait([this](error_code const& e) { reap_idle_threads(e); });
```
arvidn: I'm curious about the use of lambda here instead of std::bind. std::bind would be more concise, is the lambda more efficient? (I have a feeling it might be). I get the impression that the consensus in the C++ community is leaning towards using lambdas instead of std::bind.

arvidn:
are there any other issues to address before this is merged?

ssiloti:
I want to run one last round of tests, but I notice 8c8f5bb03b1c420543c670d2a85306491d3790ef removed loading of torrents from the comment line from client_test. That wan't intentional was it?

src/disk_io_thread_pool.cpp
ssiloti: The lambda happens to be less typing in this case due to the long class name, but mostly I just prefer lambdas. Bind is the quintessential example of a class which produces impenetrable error messages when you make a mistake. I never liked the use of placeholders::_# for unbound arguments either.

arvidn:
ouch.. no, I meant to delete the <info-hash>@<tracker-url> loading (which was the very original way of adding "trackerless torrents", before magnet links)

arvidn:
https://github.com/arvidn/libtorrent/pull/812

ssiloti:
Ok one last fix. If 5bbd268 looks good I'll squash this patch set and it'll be ready to merge.

arvidn:
let me know when you've squashed

ssiloti:
I already did, everything has been squashed into 5dbd9ea.

arvidn:
do you think there's a chance this was introduced in the merge?
https://travis-ci.org/arvidn/libtorrent/jobs/137959030

ssiloti:
That's a definite possibility. I saw this once before on Travis with this branch but I have never been able to reproduce it locally.

ssiloti:
I'm pretty sure that crash is due to a race which can cause `disk_io_thread::abort_jobs` to be called from both the network thread and the last disk I/O thread. The fix is non-trivial, hopefully I'll have a PR by tomorrow.

PR #775, opened 2016-06-01
Minor cleanup

aldenml:


src/session_impl.cpp
```diff
@@ -546,10 +546,6 @@ namespace aux {
 			std::bind(&session_impl::on_lsd_announce, this, _1));
 		TORRENT_ASSERT(!ec);
 
-#ifndef TORRENT_DISABLE_DHT
-		update_dht_announce_interval();
```
arvidn: what's the rationale for removing this?

aldenml: I might be wrong, but for what I can see, this call in `init` is called by `start_session` and that method is only called if a new `session_impl` is created. How this method can get pass `if (!m_dht)`

arvidn: I'm not sure I understand what you mean. This call will kick off the dht announce timer. The only other place this is called (as far as I can see) is in on_tick(), which is a second later. Not calling it here will add a second delay to starting the DHT announce, as far as I can tell. However, you're right that the DHT hasn't been constructed yet. I think it would make more sense to move this call to the end of this function. after update_dht() is called and after we've opened the listen sockets.

aldenml: Ok, I think that's better

PR #778, opened 2016-06-02
Refactor of listen alerts for separated fields address and port

aldenml:


src/alert.cpp
```diff
@@ -829,14 +829,16 @@ namespace libtorrent {
 	listen_failed_alert::listen_failed_alert(
 		aux::stack_allocator& alloc
 		, std::string const& iface
-		, tcp::endpoint const& ep
+		, libtorrent::address const& ip
```
arvidn: is the libtorrent:: qualifier necessary here?

src/session_impl.cpp
```diff
@@ -1585,7 +1585,8 @@ namespace aux {
 #endif
 
 				if (m_alerts.should_post<listen_failed_alert>())
-					m_alerts.emplace_alert<listen_failed_alert>(device, bind_ep, last_op
+					m_alerts.emplace_alert<listen_failed_alert>(device
+						, bind_ep.address(), bind_ep.port(), last_op
```
arvidn: technically there could still be a constructor taking an endpoint. it make simplify constructing the alert

src/alert.cpp
aldenml: Yes, at least in my environment with cmake, clang seems to get confused with the types

src/session_impl.cpp
```diff
@@ -1897,7 +1904,7 @@ namespace aux {
 					if (m_alerts.should_post<listen_failed_alert>())
 					{
 						m_alerts.emplace_alert<listen_failed_alert>(device
-							, tcp::endpoint(), listen_failed_alert::enum_if, ec
+							, address(), 0, listen_failed_alert::enum_if, ec
```
arvidn: and maybe there even should be a constructor that doesn't take an (address,port) pair

src/session_impl.cpp
```diff
@@ -1964,7 +1971,7 @@ namespace aux {
 							: listen_succeeded_alert::udp;
 
 						m_alerts.emplace_alert<listen_succeeded_alert>(
-							tcp::endpoint(udp_ep.address(), udp_ep.port()), socket_type);
+							udp_ep.address(), udp_ep.port(), socket_type);
```
arvidn: and there could be one taking a udp endpoint

arvidn:
could you add this to Changelog as well? as a reminder that this is an API change. actually, it would be nice to preserve the same interface as RC_1_1 but as deprecated too (it could be contained to the alert class itself)

bindings/python/src/alert.cpp
```diff
@@ -480,7 +480,8 @@ void bind_alert()
 
     class_<listen_failed_alert, bases<alert>, noncopyable>(
         "listen_failed_alert", no_init)
-        .def_readonly("endpoint", &listen_failed_alert::endpoint)
```
arvidn: you would need to preserve the endpoints in the python binding too, under the deprecated macro

include/libtorrent/alert_types.hpp
```diff
@@ -1301,17 +1341,36 @@ namespace libtorrent
 		enum socket_type_t { tcp, tcp_ssl, udp, i2p, socks5, utp_ssl };
 
 		// internal
-		listen_succeeded_alert(aux::stack_allocator& alloc, tcp::endpoint const& ep
+		listen_succeeded_alert(aux::stack_allocator& alloc
+			, libtorrent::address const& listen_addr
```
arvidn: is there a reason you qualify the namespace for address?

include/libtorrent/alert_types.hpp
```diff
@@ -1301,17 +1341,36 @@ namespace libtorrent
...
 			, socket_type_t t);
 
 		TORRENT_DEFINE_ALERT_PRIO(listen_succeeded_alert, 49)
 
 		static const int static_category = alert::status_notification;
 		virtual std::string message() const override;
 
+		// the address libtorrent ended up listening on. This address
+		// refers to the local interface.
+		libtorrent::address address;
```
arvidn: (and here)

src/alert.cpp
```diff
@@ -829,18 +829,67 @@ namespace libtorrent {
...
+		, int listen_port
 		, int op
 		, error_code const& ec
 		, socket_type_t t)
 		: error(ec)
 		, operation(op)
 		, sock_type(t)
-		, endpoint(ep)
+		, address(listen_addr)
+		, port(listen_port)
```
arvidn: you need to initialize the endpoint here too, right? (if deprecated features are enabled)

include/libtorrent/alert_types.hpp
aldenml: Yes, the compiler just get crazy, not just clang, but gcc too (I saw it in travis). I think that if the field is not called `address` it will be ok, do you want another name?

src/alert.cpp
```diff
@@ -916,17 +965,38 @@ namespace libtorrent {
 	}
 
 	listen_succeeded_alert::listen_succeeded_alert(aux::stack_allocator&
-		, tcp::endpoint const& ep, socket_type_t t)
-		: endpoint(ep)
+		, libtorrent::address const& listen_addr
+		, int listen_port
+		, socket_type_t t)
+		: address(listen_addr)
+		, port(listen_port)
```
arvidn: and initialize the endpoint here too (under the deprecated macro)

src/session_impl.cpp
```diff
@@ -1897,7 +1897,7 @@ namespace aux {
 					if (m_alerts.should_post<listen_failed_alert>())
 					{
 						m_alerts.emplace_alert<listen_failed_alert>(device
-							, tcp::endpoint(), listen_failed_alert::enum_if, ec
+							, listen_failed_alert::enum_if, ec
```
arvidn: ah, so nice :)

include/libtorrent/alert_types.hpp
```diff
@@ -1294,12 +1317,17 @@ namespace libtorrent
 
 		// the address libtorrent attempted to listen on
 		// see alert's documentation for validity of this value
-		address address;
+		libtorrent::address address;
```
arvidn: oh, I see, you have a member called "address", that's why you qualify it. another way to do it (I believe) would be to say "class address address;", not sure if that's better though

aldenml: You tell me, it's ugly either way :)

aldenml:
Do you mind cancelling all CI jobs?

include/libtorrent/alert_types.hpp
LordNyriox: @aldenml:  I think `ip` or `ipaddress` might work.  @arvidn?

src/alert.cpp
```diff
@@ -839,6 +839,9 @@ namespace libtorrent {
 		, sock_type(t)
 		, address(listen_addr)
 		, port(listen_port)
+#ifndef TORRENT_NO_DEPRECATE
+		, endpoint(tcp::endpoint(listen_addr, listen_port))
```
arvidn: you don't need to construct a temporary object here, you can just pass in the address and port directly to the constructor of endpoint

src/alert.cpp
```diff
@@ -970,6 +973,9 @@ namespace libtorrent {
 		, socket_type_t t)
 		: address(listen_addr)
 		, port(listen_port)
+#ifndef TORRENT_NO_DEPRECATE
+		, endpoint(tcp::endpoint(listen_addr, listen_port))
```
arvidn: same here

aldenml: and now...the wait... :)

PR #783, opened 2016-06-04
making dht counters additive

aldenml:


src/kademlia/routing_table.cpp
```diff
@@ -171,9 +171,13 @@ void routing_table::status(std::vector<dht_routing_bucket>& s) const
...
 {
+	int dht_nodes;
+	int dht_node_cache;
 	int ignore;
-	boost::tie(s.dht_nodes, s.dht_node_cache, ignore) = size();
-	s.dht_global_nodes = num_global_nodes();
+	boost::tie(dht_nodes, dht_node_cache, ignore) = size();
+	s.dht_nodes += dht_nodes;
+	s.dht_node_cache += dht_node_cache;
+	s.dht_global_nodes += num_global_nodes();
```
arvidn: dht_global_nodes actually doesn't make sense to have additive. if anything, it should be the average.. Well, actually, when it's across IPv4 and IPv6, adding would make sense. in the future though, where we may have one DHT node per external interface (which may be multiple of the same address family), then it becomes a bit trickier...

src/kademlia/dht_tracker.cpp
```diff
@@ -215,19 +216,47 @@ namespace libtorrent { namespace dht
...
+		c.set_value(counters::dht_torrents, dht_cnt.torrents);
+		c.set_value(counters::dht_peers, dht_cnt.peers);
+		c.set_value(counters::dht_immutable_data, dht_cnt.immutable_data);
+		c.set_value(counters::dht_mutable_data, dht_cnt.mutable_data);
+
+		c.set_value(counters::dht_nodes, 0);
+		c.set_value(counters::dht_node_cache, 0);
+		c.set_value(counters::dht_allocated_observers, 0);
+
+		m_dht.update_stats_counters(c);
```
arvidn: could you add a TODO comment here saying that the nodes should return these values and they should be added to counters in this function instead? (that way it's obvious that the clearing and adding is 1:1)

arvidn: or making that change would be fine too :)

src/kademlia/node.cpp
```diff
@@ -743,22 +743,13 @@ void node::status(std::vector<dht_routing_bucket>& table
...
-	c.set_value(counters::dht_mutable_data, dht_cnt.mutable_data);
-
 	int nodes, replacements;
 	boost::tie(nodes, replacements, boost::tuples::ignore) = size();
-	c.set_value(counters::dht_nodes, nodes);
-	c.set_value(counters::dht_node_cache, replacements);
-	c.set_value(counters::dht_allocated_observers, m_rpc.num_allocated_observers());
+	c.inc_stats_counter(counters::dht_nodes, nodes);
+	c.inc_stats_counter(counters::dht_node_cache, replacements);
+	c.inc_stats_counter(counters::dht_allocated_observers, m_rpc.num_allocated_observers());
```
arvidn: this function could just return a tuple of 3 values

PR #789, opened 2016-06-05
changed min_distance_exp for correctness and clarity

aldenml:


aldenml:
@arvidn you were right about the case of no ids.

Regarding the performance of `min_distance_exp` I removed all special cases, since this apply only when the DHT is full of items and it wasn't  a meaningful improvement anyway (if any at all). I left a note because there is a performance hit when this case arrives.

include/libtorrent/kademlia/dht_storage.hpp
```diff
@@ -76,6 +76,8 @@ namespace dht
 	// libtorrent comes with one built-in storage implementation:
 	// ``dht_default_storage`` (private non-accessible class). Its
 	// constructor function is called dht_default_storage_constructor().
+	// You should know that if this storage becomes full of DHT items,
+	// the current implementation could degrade in performance.
```
arvidn: what does this refer to exactly?

simulation/test_dht_storage.cpp
```diff
@@ -75,6 +75,19 @@ namespace
...
+
+	std::unique_ptr<dht_storage_interface> create_default_dht_storage(
+		dht_settings const& sett)
+	{
+		std::unique_ptr<dht_default_storage> s(dht_default_storage_constructor(sett));
+		TEST_CHECK(s.get() != NULL);
+
+		std::vector<node_id> node_ids;
+		node_ids.push_back(to_hash("0000000000000000000000000000000000000200"));
+		s->update_node_ids(node_ids);
```
arvidn: these days you can actually say:

s->update_node_ids({to_hash("0000000000000000000000000000000000000200")});

simulation/test_dht_storage.cpp
```diff
@@ -106,20 +119,17 @@ void test_expiration(high_resolution_clock::duration const& expiry_time
...
 }
 
-#endif // TORRENT_DISABLE_DHT
-
 TORRENT_TEST(dht_storage_counters)
 {
-#ifndef TORRENT_DISABLE_DHT
 	dht_settings sett = test_settings();
-	boost::shared_ptr<dht_storage_interface> s(dht_default_storage_constructor(sett));
+	std::unique_ptr<dht_storage_interface> s(create_default_dht_storage(sett));
```
arvidn: oops, I just made this change last night (and pulled it a few moments ago)

test/test_dht_storage.cpp
```diff
@@ -82,29 +82,41 @@ namespace
...
+
+	std::unique_ptr<dht_storage_interface> create_default_dht_storage(
+		dht_settings const& sett)
+	{
+		std::unique_ptr<dht_storage_interface> s(dht_default_storage_constructor(sett));
+		TEST_CHECK(s.get() != NULL);
+
+		std::vector<node_id> node_ids;
+		node_ids.push_back(to_hash("0000000000000000000000000000000000000200"));
+		s->update_node_ids(node_ids);
```
arvidn: same here: s->update_node_ids({to_hash("0000000000000000000000000000000000000200")});

arvidn:
I made some of the changes in here already (specifically the unique_ptr change), so you'll have to rebase.

regarding the performance, it's not obvious to me that there's a performance disadvantage having _just_ the loop.

aldenml:
It's ok, I will rebase on top of master.

The comment is not about the loop, but more about the amount of calculation is performed when an element needs to be removed. To me, is not obvious without taking a look at the implementation that when the DHT is full of items, there is a `O(num_items * num_ids)` operations to remove one. One can says that `num_ids` should be fairly small, but still it's a variable.

aldenml:
I made an unnecessary push, feel free to cancel the associated CIs

test/test_dht_storage.cpp
```diff
@@ -83,29 +83,39 @@ namespace
 		g_storage_constructor_invoked = true;
 		return dht_default_storage_constructor(settings);
 	}
+
+	std::unique_ptr<dht_storage_interface> create_default_dht_storage(
+		dht_settings const& sett)
+	{
+		std::unique_ptr<dht_storage_interface> s(dht_default_storage_constructor(sett));
+		TEST_CHECK(s.get() != NULL);
```
arvidn: for future reference, we should use nullptr instead of NULL. no need to go through another CI cycle for that now though

aldenml: It seems like I need to do one more push

arvidn:
do you understand why the errors don't happen on linux and osx?

aldenml:
I was about to ask the same. I think the simulation tests are not running in linux and osx, is that intentional?

arvidn:
aha, there's a typo in the build script. somehow this still does not return an error, which is surprising.

https://travis-ci.org/arvidn/libtorrent/jobs/135468670#L2071

simulation/test_dht_storage.cpp
```diff
@@ -88,7 +88,7 @@ namespace
 	}
 }
 
-void timer_tick(std::unique_ptr<dht_storage_interface> s
+void timer_tick(std::unique_ptr<dht_storage_interface>& s
```
arvidn: technically this should be raw pointers.. again, no need to change if this passes CI

arvidn:
btw. I fixed the .travis script. If you rebase on top of master it should build and run the simulations there as well

PR #790, opened 2016-06-05
consider macro TORRENT_USE_IOSTREAM for sha1_hash iostream operators

aldenml:


arvidn:
does this solve a problem you're having?

aldenml:
not a problem actually, but I compile with `BOOST_NO_IOSTREAM` with the hope of saving some bytes in the final binary.

Issue #781, opened 2016-06-03
crash in do_check_fastresume


aldenml:
libtorrent: RC_1_1
platform/architecture: android
compiler: GCC 4.9

We are getting quite a few errors like this:
```
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x45aa825b
    r0 ae40d748  r1 00000000  r2 125b204b  r3 45aa8258
    r4 00000002  r5 9e4e9f00  r6 00000000  r7 9f4d6d0c
    r8 ae40d6f8  r9 ad88da58  sl ae40d748  fp b2cc2d64
    ip 9e002000  sp ae40d668  lr 9f0d5658  pc 9f020854  cpsr 000f0010

backtrace:
    #00 pc 00254854  /data/app/com.frostwire.android-1/lib/arm/libjlibtorrent.so (libtorrent::bdecode_node::type() const+16)
```
At first it could seem like an error using libtorrent from java, but one particular report tell me that that's not the case:
```
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 7154383b
    r0 6275a764  r1 6275a764  r2 627dbf00  r3 71543838
    r4 6275a764  r5 627ade64  r6 626349bc  r7 6275a764
    r8 62731e60  r9 5d6651a8  sl 1aadafbf  fp 145491f6
    ip 6188525c  sp 620d4c88  lr 617c3cc0  pc 6170b854  cpsr 00000010
    d0  656761726f74732f  d1  6574616c756d652f
    d2  736f72462f302f64  d3  6f542f6572695774
    d4  02317b33fab388b9  d5  fc7160387b344506
    d6  0000000080000000  d7  0000018000000000
    d8  0000000000000000  d9  0000000000000000
    d10 0000000000000000  d11 0000000000000000
    d12 0000000000000000  d13 0000000000000000
    d14 0000000000000000  d15 0000000000000000
    d16 74696d7364726f57  d17 78742e6f69625f68
    d18 5f65676174736b63  d19 4b3032335f33504d
    d20 495754534f52465f  d21 54534f52465f4552
    d22 52435f4b43494c43  d23 435f455649544145
    d24 89da08827ccd5248  d25 85dd803928dc2e2a
    d26 f1cae3e0bd980af7  d27 e1d8c2bc6dd37a7e
    d28 5a8279995a827999  d29 5a8279995a827999
    d30 0000000000000000  d31 0000000000000000
    scr 60000010

[...]

PR #795, opened 2016-06-06
avoid double inspection of bdecode_node in piece_manager::check_fastresume

aldenml:
@arvidn this is to tackle https://github.com/arvidn/libtorrent/issues/781

I can see that `torrent::init` could be called on multiple occasions, potentially putting multiple fast resume jobs. I'm not a fan of mutable states, but in this case I couldn't see any other quick alternative.

If this is not acceptable or break some other assumption, let me know another path. Thanks.


aldenml:
Wrong branch, see https://github.com/arvidn/libtorrent/pull/796

PR #796, opened 2016-06-06
avoid double inspection of bdecode_node in piece_manager::check_fastresume

aldenml:
@arvidn this is to tackle #781

I can see that torrent::init could be called on multiple occasions, potentially putting multiple fast resume jobs. I'm not a fan of mutable states, but in this case I couldn't see any other quick alternative.

If this is not acceptable or break some other assumption, let me know another path. Thanks.

### DIFF
--- a/include/libtorrent/storage.hpp
+++ b/include/libtorrent/storage.hpp
@@ -689,6 +689,8 @@ namespace libtorrent
 		// the piece_manager destructs. This is because
 		// the torrent_info object is owned by the torrent.
 		boost::shared_ptr<void> m_torrent;
+
+		bool checked_fastresume;
 	};
 
 	// this identifies a read or write operation so that readwritev() knows
--- a/src/storage.cpp
+++ b/src/storage.cpp
@@ -1616,6 +1616,7 @@ namespace libtorrent
 		: m_files(*files)
 		, m_storage(storage_impl)
 		, m_torrent(torrent)
+		, checked_fastresume(false)
 	{
 	}
 
@@ -1691,6 +1692,9 @@ namespace libtorrent
 		, storage_error& ec)
 	{
 		TORRENT_ASSERT(m_files.piece_length() > 0);
+
+		if (checked_fastresume) return check_no_fastresume(ec);
+		checked_fastresume = true;
 
 		// if we don't have any resume data, return
 		if (rd.type() == bdecode_node::none_t) return check_no_fastresume(ec);
--- a/test/test_storage.cpp
+++ b/test/test_storage.cpp
@@ -92,6 +92,32 @@ void on_check_resume_data(disk_io_job const* j, bool* done)
 	*done = true;
 }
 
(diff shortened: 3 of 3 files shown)


include/libtorrent/storage.hpp
```diff
@@ -689,6 +689,8 @@ namespace libtorrent
 		// the piece_manager destructs. This is because
 		// the torrent_info object is owned by the torrent.
 		boost::shared_ptr<void> m_torrent;
+
+		bool checked_fastresume;
```
arvidn: I don't really understand what this solves. When a user force-recheck a torrent, do we come back through the check_fastresume() case? In either case, I think this would be a good place to describe what this variable is used for and why it's needed

src/storage.cpp
```diff
@@ -1692,6 +1693,9 @@ namespace libtorrent
 	{
 		TORRENT_ASSERT(m_files.piece_length() > 0);
 
+		if (checked_fastresume) return check_no_fastresume(ec);
```
arvidn: continuing down here, checking the resume data, isn't supposed to invalidate it. do you have a reason to believe it does? as far as I can tell, you should be able to call this function multiple times and have the same outcome. If you restore the logic here and run your test, do you get any errors?

aldenml: Let me explain a little more, since this is an important issue for jlibtorrent, we are getting a lot of errors like this:
```
backtrace:
    #00  pc 00254854  /data/app-lib/com.frostwire.android-1/libjlibtorrent.so (libtorrent::bdecode_node::type() const+16)
    #01  pc 0030ccbc  /data/app-lib/com.frostwire.android-1/libjlibtorrent.so (libtorrent::piece_manager::check_fastresume(libtorrent::bdecode_node const&, std::vector<std::string, std::allocator<std::string> > const*, libtorrent::storage_error&)+28)
```
You mentioned in https://github.com/arvidn/libtorrent/issues/781 that probably `async_check_fastresume` was called twice and the first one resulted in an error. Then the second call ends in a crash.

I think this is exactly the reason, since the potential problems of emulating a posix layer with the android SD card framework is just insane.

This just prevents a double inspection of the job `bdecode_node` pointer.

If I restore the logic, my test fail of course, but that is because I expect this logic in place, see `on_checked_resume_data` for `n==2` and `n==3` the `TEST_EQUAL(j->error.ec.value(), 0);`

I rather detect the exact logic flow that creates the two calls to `async_check_fastresume`, but I couldn't.

arvidn: the problem I would be worried about (apart from this trying to treat the symptom rather than the illness) is that there may be legitimate cases of calling async_check_fastresume() twice. For instance when force-rechecking a torrent, or when promoting a "magnet link" into a proper torrent (either by it receiving metadata or by adding a .torrent file that match the info-hash)

aldenml: I understand you. Well, I will keep looking, if you have any new idea, let us know.

PR #798, opened 2016-06-07
early resume data reset when fatal_disk_error

aldenml:
And this? issue https://github.com/arvidn/libtorrent/issues/781

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -1837,7 +1837,7 @@ namespace libtorrent
 #endif
 		}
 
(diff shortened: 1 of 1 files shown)


arvidn:
this reset() happen _after_ that crash, doesn't it?

arvidn:
hm.. you're thinking that perhaps invoking those callbacks may trigger another async_check_fastresume?

aldenml:
it appears to me that it happens _before_ the crash and yes, I'm thinking that or that during those calls, a `torrent::resume` is called for some reason.

Still this is just "mitigation", in the hope that early `reset` allow for passing a proper `NULL` parameter during the possible second call to `async_check_fastresume`.

PR #813, opened 2016-06-13
fixed bug related to flag_merge_resume_http_seeds flag in add_torrent…

arvidn:
…_params


arvidn:
This should be safe because flag_merge_resume_http_seeds was moved to a flag that was non-existent in the previous release, and should always be zero when called from clients built against the 1.1.0 ABI.

aldenml:
LGTM

PR #815, opened 2016-06-13
replace the suggest-read-cache logic with something much simpler

arvidn:
that doesn't schedule regular jobs and doesn't query the disk cache


arvidn:
this disconnects the main bittorrent engine (more) from the implementation details of the disk cache, in support for an mmap-based one.

PR #817, opened 2016-06-13
remove lazy_bitfield

arvidn:


sledgehammer999:
Looking back in history it would be nice to have a line or two in the PR description on why features are to be removed...

arvidn:
iirc, the main reason to remove this was to reduce complexity in the code. afaict, this feature is inferior to using protocol encryption/obfuscation. My impression is also devices that's detecting these kinds of messages are not widely deployed anymore. Or at least that the ones that are, are a lot more sophisticated than to be tricked by lazy bitfields.

Another reason is that the "HAVE_ALL" message is so much more efficient, protocol wise, so that's a win generally (I realize lazy_bitfields were off by default anyway).

Do you have samples to suggest that this feature makes a big difference?

sledgehammer999:
>(I realize lazy_bitfields were off by default anyway).

qbt had them turned on explicitly all these years and the user couldn't turn them off.

But to the real question: I am not particular to this feature, nor do I have better knowledge than you on this matter. I just wanted an explanation on the deprecation. Thank you.

PR #818, opened 2016-06-14
remove option to disable contiguous receive buffers

arvidn:
This greatly simplifies receive buffer management and disk buffer management in peers. Given that download performance degrade with non-contiguous receive buffers, the complexity hardly seems justified.

The only possible upside would be if low memory usage is more important than performance.

### DIFF
--- a/src/bt_peer_connection.cpp
+++ b/src/bt_peer_connection.cpp
@@ -1188,43 +1188,19 @@ namespace libtorrent
 					disconnect(errors::packet_too_large, op_bittorrent, 2);
 					return;
 				}
-
-				m_recv_buffer.assert_no_disk_buffer();
-				if (!m_settings.get_bool(settings_pack::contiguous_recv_buffer) &&
-					m_recv_buffer.can_recv_contiguous(m_recv_buffer.packet_size() - 13 - list_size))
-				{
-					if (!allocate_disk_receive_buffer(m_recv_buffer.packet_size() - 13 - list_size))
-					{
-						received_bytes(0, received);
-						return;
-					}
-				}
 			}
 		}
 		else
 		{
 			if (recv_pos == 1)
 			{
-				m_recv_buffer.assert_no_disk_buffer();
-
 				if (m_recv_buffer.packet_size() - 9 > t->block_size())
 				{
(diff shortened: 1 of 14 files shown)


arvidn:
@ssiloti I know you spent a lot of time getting this right in the crypto plugin. Do you think it makes sense to remove the receive-directly-into-disk-buffer feature? If so, would you mind taking a look?

src/pe_crypto.cpp
```diff
@@ -179,7 +179,7 @@ namespace libtorrent
 		if (recv_buffer.crypto_packet_finished())
 		{
 			std::vector<boost::asio::mutable_buffer> wr_buf;
-			recv_buffer.mutable_buffers(wr_buf, bytes_transferred);
```
arvidn: I believe we still need the vector here though, since this is used for outgoing data as well (which definitely needs to be able to mix in disk-buffers)

arvidn:
for reference: http://blog.libtorrent.org/2011/12/socket-receive-buffers/

src/bt_peer_connection.cpp
```diff
@@ -1188,43 +1188,19 @@ namespace libtorrent
...
-				m_recv_buffer.assert_no_disk_buffer();
-
 				if (m_recv_buffer.packet_size() - 9 > t->block_size())
 				{
 					disconnect(errors::packet_too_large, op_bittorrent, 2);
 					return;
 				}
-
-				if (!m_settings.get_bool(settings_pack::contiguous_recv_buffer) &&
-					m_recv_buffer.can_recv_contiguous(m_recv_buffer.packet_size() - 9))
```
ssiloti: These were the only calls to `can_recv_contiguous` so it can be removed too.

arvidn:
a somewhat related thought I had while making this change was that there are currently 3 (or 5 depending on how you count) classes for representing a range of bytes. asio::buffer/mutable_buffer, buffer::interval/const_interval and array_view. I wonder if there's a way to combine them. presumably adding free functions operating on the asio type, to provide the functionality of the other ones..

ssiloti:
This code is a pain and I'm happy to see it go. LGTM.

PR #823, opened 2016-06-15
plugin callback refactor

aldenml:
For review

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -816,18 +816,18 @@ namespace aux {
 		TORRENT_ASSERT(is_single_thread());
 		TORRENT_ASSERT_VAL(ext, ext);
 
-		boost::uint32_t features = ext->implemented_features();
+		boost::uint32_t const features = ext->implemented_features();
 
 		m_ses_extensions[plugins_all_idx].push_back(ext);
-		if (features & plugin::optimistic_unchoke_feature)
-			m_ses_extensions[plugins_optimistic_unchoke_idx].push_back(ext);
-		if (features & plugin::tick_feature)
-			m_ses_extensions[plugins_tick_idx].push_back(ext);
-		if (features & plugin::dht_request_feature)
-			m_ses_extensions[plugins_dht_request_idx].push_back(ext);
 
 		if (no_wrap)
 		{
+			if (features & plugin::optimistic_unchoke_feature)
+				m_ses_extensions[plugins_optimistic_unchoke_idx].push_back(ext);
+			if (features & plugin::tick_feature)
+				m_ses_extensions[plugins_tick_idx].push_back(ext);
+			if (features & plugin::dht_request_feature)
+				m_ses_extensions[plugins_dht_request_idx].push_back(ext);
 			if (features & plugin::alert_feature)
 				m_alerts.add_extension(ext);
 			ext->added(session_handle(this));
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -638,7 +638,7 @@ namespace aux {
 		for (auto& ext : m_ses_extensions)
 		{
 			TORRENT_TRY {
-				ext->save_state(*eptr);
+				ext.first->save_state(*eptr);
 			} TORRENT_CATCH(std::exception&) {}
 		}
 #endif
@@ -785,7 +785,7 @@ namespace aux {
 		for (auto& ext : m_ses_extensions)
 		{
 			TORRENT_TRY {
-				ext->load_state(*e);
+				ext.first->load_state(*e);
 			} TORRENT_CATCH(std::exception&) {}
 		}
 #endif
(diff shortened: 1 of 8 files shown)


src/session_impl.cpp
```diff
@@ -3145,7 +3133,8 @@ namespace aux {
 			for (auto& ext : m_ses_extensions)
```
arvidn: would it make more sense to have separate lists for separate features? the "last minute check" is a bit of an anti-pattern

aldenml: I was thinking about a map, because having multiple lists is a nightmare to maintain, but I feared throwing to much data structure could impact performance in each iteration.

If you don't mind, how the "last minute check" is an anti-pattern? just to know your opinion in the matter.

arvidn: in this case it most likely won't make a difference, but imagine you iterate over a large list of items and check if some bool is true on each of them. For the ones where it's true, you perform some work. You're going to pay the cost of pulling in a cache line for each test, making it scale O(n). If the objects instead were sorted or organized based on the property or type (and perhaps even organized contiguously in memory) it would be a lot cheaper to iterate over them.

in short, the last-minute-check indicates that data/objects may not be organized optimally in memory.

aldenml: Thank you, I understand and agree with you.

Having said that, I "literally" speculate that maybe in this case, the vector is so small, that it fits in the cache. The iteration "should" still be fast. Btw, I want to comment that with this change, there is a big saving in not calling `on_alerts` and with the `m_session_extension_features` variable, the saving is even bigger if no plugin happens to implement an specific feature (that's yours).

Well, not sure if you are willing to accept this solution, or you have something else in mind.

ssiloti:
Do you have a particular extension in mind which motivated this change?

aldenml:
to be honest, my main motivation is having a friendlier API for swig, the rest is just noise and "looking to the sky" thinking.

aldenml:
@arvidn I will give it a try to the multiple list option

aldenml:
and now?...

aldenml:
@arvidn if you have a change, can you restart the travis job: https://travis-ci.org/arvidn/libtorrent/jobs/138105795 thanks

src/session_impl.cpp
```diff
@@ -811,35 +808,29 @@ namespace aux {
...
-			e != dht_ext.end(); ++e)
+		m_ses_extensions[plugins_all_idx].push_back(ext);
+		if (features & plugin::optimistic_unchoke_feature)
+			m_ses_extensions[plugins_optimistic_unchoke_idx].push_back(ext);
+		if (features & plugin::tick_feature)
+			m_ses_extensions[plugins_tick_idx].push_back(ext);
+		if (features & plugin::dht_request_feature)
+			m_ses_extensions[plugins_dht_request_idx].push_back(ext);
+
+		if (no_wrap)
```
arvidn: it's not obvious to me what "no_wrap" refers to (even when looking at this code). Could you maybe add a comment?

src/session_impl.cpp
```diff
@@ -811,35 +808,29 @@ namespace aux {
...
+	void session_impl::add_ses_extension(boost::shared_ptr<plugin> ext, bool no_wrap)
 	{
 		TORRENT_ASSERT(is_single_thread());
 		TORRENT_ASSERT_VAL(ext, ext);
 
-		m_ses_extensions.push_back(ext);
-		m_alerts.add_extension(ext);
-		ext->added(session_handle(this));
-		m_session_extension_features |= ext->implemented_features();
+		boost::uint32_t features = ext->implemented_features();
```
arvidn: if you and up updating this PR, maybe you could make this const while you're at it

src/session_impl.cpp
```diff
@@ -811,35 +808,29 @@ namespace aux {
...
 		{
-			TORRENT_ASSERT(e->first.size() <= max_dht_query_length);
-			if (e->first.size() > max_dht_query_length) continue;
-			extension_dht_query registration;
-			registration.query_len = uint8_t(e->first.size());
-			std::copy(e->first.begin(), e->first.end(), registration.query.begin());
-			registration.handler = e->second;
-			m_extension_dht_queries.push_back(registration);
+			if (features & plugin::alert_feature)
+				m_alerts.add_extension(ext);
```
arvidn: it's not clear to me why the alert_feature is a special case here. why can't you subscribe to this unless you also have the added() function called? (maybe this is rooted in me not getting the purpose of no_wrap)

arvidn:
I actually like this change more than I expected :)

aldenml:
good to know, I will add the comment

src/session_impl.cpp
aldenml: I will add the comment of `no_wrap`, but I'm pretty sure is the name that it's wrong. Now, about the special nature of `alert_feature`, the reasoning here is that if you are a "wrap", you are 100% sure the plugin is not subscribed to alerts. And this is the same reason the call `added()` is inside this `if` too. Take a look at `session_impl::add_extension(ext_function_t ext)`:
```
boost::shared_ptr<plugin> p(new session_plugin_wrapper(ext));
add_ses_extension(p, false);
```

aldenml: and by the same reasoning, I'm moving the rest to inside the `if` too

aldenml: well, `no_wrap` is gone. It was an unnecessary optimization

PR #827, opened 2016-06-16
making disk_buffer_holder internal and include cleanup

aldenml:
`disk_buffer_holder` was exposing a pointer to a private structure (`piece_manager`) via `ref()`. I expect some comment about the new `on_piece` signature.

### DIFF
--- a/include/libtorrent/disk_buffer_holder.hpp
+++ b/include/libtorrent/disk_buffer_holder.hpp
@@ -36,12 +36,10 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/config.hpp"
 #include "libtorrent/assert.hpp"
 #include "libtorrent/disk_io_job.hpp" // for block_cache_reference
-#include <algorithm>
 
 #include "libtorrent/aux_/disable_warnings_push.hpp"
 
 #include <boost/shared_ptr.hpp>
-#include <boost/function.hpp>
 
 #include "libtorrent/aux_/disable_warnings_pop.hpp"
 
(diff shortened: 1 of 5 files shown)


include/libtorrent/disk_buffer_holder.hpp
```diff
@@ -69,7 +67,7 @@ namespace libtorrent
 	// 
 	// ``get()`` returns the pointer without transferring ownership. If
 	// this buffer has been released, ``get()`` will return nullptr.
-	struct TORRENT_EXPORT disk_buffer_holder
+	struct TORRENT_EXTRA_EXPORT disk_buffer_holder
```
arvidn: disk_buffer_holder is actually passed into one of the functions in peer_plugin, which means it needs to be exported.

include/libtorrent/extensions.hpp
```diff
@@ -426,7 +425,7 @@ namespace libtorrent
 		virtual bool on_allowed_fast(int /*index*/) { return false; }
 		virtual bool on_request(peer_request const&) { return false; }
 		virtual bool on_piece(peer_request const& /*piece*/
-			, disk_buffer_holder& /*data*/) { return false; }
+			, char* /*buf*/) { return false; }
```
arvidn: oh, I see. you changed that. what's the ownership now? (I realize that I should have updated this in my previous patch to pass this by-value too)

arvidn: actually, by-value wouldn't work. it must be passed as a non-owning pointer

arvidn: either way, this should probably be a "char const*", don't you think?

arvidn:
looks reasonable. it should probably be documented in the peer_plugin interface what this pointer points to (non-owning, 16 kiB, immutable disk buffer).

include/libtorrent/extensions.hpp
aldenml: yes

aldenml:
Ok then, I will add the documentation and see if I can add some unit test.

aldenml:
unit test requieres a lot more changes not related to this

arvidn:
lgtm

PR #828, opened 2016-06-17
deprecate functions in torrent_handle

aldenml:
feel free to discard/close this PR if you don't agree with the API change.

I think `get_storage_impl` is very dangerous and `get_full_peer_list` is too internal.

### DIFF
--- a/include/libtorrent/peer_info.hpp
+++ b/include/libtorrent/peer_info.hpp
@@ -162,7 +162,7 @@ namespace libtorrent
 			// connect via the NAT holepunch mechanism.
 			holepunched = 0x8000,
 
-			// indicates that this socket is runnin on top of the
+			// indicates that this socket is running on top of the
 			// I2P transport.
 			i2p_socket = 0x10000,
 
@@ -416,7 +416,7 @@ namespace libtorrent
 	};
 
 	// internal
-	struct TORRENT_EXPORT peer_list_entry
+	struct TORRENT_EXTRA_EXPORT peer_list_entry
 	{
 		// internal
 		enum flags_t
--- a/include/libtorrent/torrent_handle.hpp
+++ b/include/libtorrent/torrent_handle.hpp
@@ -43,7 +43,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <boost/shared_ptr.hpp>
 #include <boost/function.hpp>
 #include <boost/weak_ptr.hpp>
-#include <boost/cstdint.hpp>
 
 #ifndef TORRENT_NO_DEPRECATE
 // for deprecated force_reannounce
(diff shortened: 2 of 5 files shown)


include/libtorrent/peer_info.hpp
```diff
@@ -416,7 +416,7 @@ namespace libtorrent
 	};
 
 	// internal
-	struct TORRENT_EXPORT peer_list_entry
+	struct TORRENT_EXTRA_EXPORT peer_list_entry
```
arvidn: I mostly agree with this. I added this a long time ago for research purposes. I wonder if there's any other way of accessing the peers. maybe in resume data. I think the resume data limits the number of peers being saved though

include/libtorrent/torrent_handle.hpp
```diff
@@ -298,8 +293,11 @@ namespace libtorrent
 		// otherwise.
 		bool have_piece(int piece) const;
 
+#ifndef TORRENT_NO_DEPRECATE
 		// internal
+		TORRENT_DEPRECATED
 		void get_full_peer_list(std::vector<peer_list_entry>& v) const;
```
arvidn: I guess this is where I meant to put the peer_list_entry comment

include/libtorrent/torrent_handle.hpp
```diff
@@ -873,9 +871,12 @@ namespace libtorrent
 			, std::string const& private_key
 			, std::string const& dh_params);
 
+#ifndef TORRENT_NO_DEPRECATE
 		// Returns the storage implementation for this torrent. This depends on the
 		// storage constructor function that was passed to add_torrent.
+		TORRENT_DEPRECATED
 		storage_interface* get_storage_impl() const;
```
arvidn: this is a bit trickier I think. The reason is that libtorrent creates and owns the storage object, and even though it was user-supplied, the user doesn't have a reasonable way to talk to it (other than this function). The alternative (with the current design) would be for the storage_constructor object to keep a reference to an outside pointer, and fill that in when it creates a storage object. Another alternative would be for the user to construct the storage up-front, and pass in the storage object via add_torrent_params, rather than the storage_constructor function. However, that makes the storage object a bit messy, since it can't be properly initialized by the user, it would require two-phase initialization.

arvidn: Imagine a "RAM-disk" storage object. It might be very convenient to be able to access the stored blocks or populate it from the outside, rather than having to read or write them through libtorrent.

aldenml: I'm familiar with the problematic since providing a binding from java force me to do the two-phase initialization (not good).

I will undeprecate this and get the rest of the PR accepted, let's see if in the future another solution comes up.

PR #831, opened 2016-06-18
Fix abort_jobs being called multiple times

ssiloti:
disk_io_thread::abort_jobs must only be called once, but it was possible for
concurrent calls to be made from abort() and thread_fun() if the max threads
had been set to zero but there were still disk I/O threads running.

### DIFF
--- a/src/disk_io_thread.cpp
+++ b/src/disk_io_thread.cpp
@@ -160,6 +160,7 @@ namespace libtorrent
 		, m_outstanding_reclaim_message(false)
 #if TORRENT_USE_ASSERTS
 		, m_magic(0x1337)
+		, m_jobs_aborted(false)
 #endif
 	{
 		ADD_OUTSTANDING_ASYNC("disk_io_thread::work");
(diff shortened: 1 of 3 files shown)


src/disk_io_thread.cpp
```diff
@@ -196,16 +197,23 @@ namespace libtorrent
 
 	void disk_io_thread::abort(bool wait)
 	{
-		m_abort = true;
-		if (num_threads() == 0)
+		// abuse the job mutex to make setting m_abort and checking the thread count atomic
+		// see also the comment in thread_fun
+		std::unique_lock<std::mutex> l(m_job_mutex);
+		if (m_abort.exchange(true)) return;
```
arvidn: given that m_job_mutex is held here, does m_abort still need to be an atomic? If all access to m_abort is done while holding the m_job_mutex, presumably it doesn't. If there are accesses to it where the mutex is not held, I take it those places would never interact with aborting jobs or exiting threads. It seems a bit suspicious though

src/disk_io_thread.cpp
```diff
@@ -196,16 +197,23 @@ namespace libtorrent
...
-		}
+
+		// even if there are no threads it doesn't hurt to abort the pools
+		// it prevents threads from being started after an abort which is a good
+		// defensive programming measure
+		m_generic_threads.abort(wait);
+		m_hash_threads.abort(wait);
 	}
 
 	// TODO: 1 it would be nice to have the number of threads be set dynamically
```
arvidn: heh, I should remove this comment now :)

src/disk_io_thread.cpp
ssiloti: There are a couple of m_abort checks down in the job handlers which don't need to hold the job mutex. Now that I've looked at it though, m_num_running_threads does not need to be atomic anymore.

PR #834, opened 2016-06-18
config.h cleanup

aldenml:


Jamfile
```diff
@@ -491,6 +491,7 @@ variant test_release : release
 	: <asserts>production <debug-symbols>on
 	  <invariant-checks>full <boost-link>shared <optimization>off
 	  <export-extra>on <debug-iterators>on <threading>multi
+	  <inlining>off
```
arvidn: is this a build-time optimization? In some sense it's nice to have test coverage of optimized code, to make sure there's no broken code that just works in debug mode.

include/libtorrent/config.hpp
```diff
@@ -93,9 +89,6 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #elif defined BOOST_MSVC
 
-#pragma warning(disable: 4258)
```
arvidn: do you know if this introduced any new warnings? simulations and the python binding isn't warnings-as-errors yet (but The currently building PR on appveyor changes that). But if this patch makes it in first, it would be nice to make sure this warning is issued.

src/cpuid.cpp
```diff
@@ -30,9 +30,6 @@ POSSIBILITY OF SUCH DAMAGE.
 
 */
 
-#include "libtorrent/config.hpp"
-#include <cstring>
-
 #include "libtorrent/aux_/cpuid.hpp"
 
 #if defined _MSC_VER && TORRENT_HAS_SSE
```
arvidn: isn't TORRENT_HAS_SSE defined by config.hpp? i.e. it will always be false with this patch

.travis.yml
```diff
@@ -1,14 +1,20 @@
...
+    - env: analyze=1
+    - os: osx
+      osx_image: xcode6.4
+      env: variant=test_release docs=1
+    - os: osx
+      osx_image: xcode6.4
+      env: variant=test_debug
+    - os: osx
+      osx_image: xcode6.4
+      env: variant=test_barebones
```
arvidn: does the default osx image have a too old version of clang?
With this change, I imagine there are some conditions further down that can be cleaned up as well, for instance, currently coverage is disabled for OSX by additional comparisons, as well as analyze=1

Jamfile
aldenml: The real reason is that `boost-build` in brew is very very old, it specifies the flag `-finline-functions` in release by default, but newer versions of clang can't accept this flag. Now, I found that you already specify `<optimization>off` because of travis issues, so I thought that one less optimization is of no harm here.

You tell me if this is acceptable.

include/libtorrent/config.hpp
aldenml: I don't know, need to check that in the logs

src/cpuid.cpp
aldenml: Isn't the macro propagated via the include of `libtorrent/aux_/cpuid.hpp`? Do you prefer the explicit inclusion?

.travis.yml
aldenml: Yes, it's too old, and by the time this branch is finished, this will be old too :)

I didn't want to super clean the .travis.yml before a preliminary review, but I assume from your comment that I can do it.

Good news is that one travis job is saved (now at 7)

Jamfile
arvidn: oh. you're right about the optimizations.. So, by upgrading the image, you get a newer clang version, but you don't get a newer boost-build version, so that causes problems, is that the case?

src/cpuid.cpp
arvidn: oh, I suppose so.

Jamfile
aldenml: Yes, brew is independent of the OS X image, and `boost-build` brew formula is deprecated btw (from 2014).

CMakeLists.txt
```diff
@@ -275,8 +275,6 @@ endif()
 include_directories(${Boost_INCLUDE_DIRS})
 target_link_libraries(torrent-rasterbar ${Boost_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})
 
-# this works around a bug in asio in boost-1.39
-add_definitions(-DBOOST_ASIO_HASH_MAP_BUCKETS=1021 )
```
arvidn: I take it boost-1.39 is too old to work for libtorrent

PR #835, opened 2016-06-19
use array_view instead of std::vector in crypto_plugin interface

arvidn:
return a vector of const_buffers for prepending new buffers. use stack allocated (single buffer) iovecs for receives. general cleanup

### DIFF
--- a/include/libtorrent/bt_peer_connection.hpp
+++ b/include/libtorrent/bt_peer_connection.hpp
@@ -159,9 +159,9 @@ namespace libtorrent
 		void on_receive_impl(std::size_t bytes_transferred);
 
 #if !defined(TORRENT_DISABLE_ENCRYPTION) && !defined(TORRENT_DISABLE_EXTENSIONS)
-		// next_barrier
+		// next_barrier, buffers-to-prepend
 		virtual
-		std::tuple<int, std::vector<boost::asio::const_buffer>>
+		std::tuple<int, aux::array_view<boost::asio::const_buffer>>
 		hit_send_barrier(aux::array_view<boost::asio::mutable_buffer> iovec) override;
 #endif
 
--- a/include/libtorrent/bt_peer_connection.hpp
+++ b/include/libtorrent/bt_peer_connection.hpp
@@ -159,7 +159,10 @@ namespace libtorrent
 		void on_receive_impl(std::size_t bytes_transferred);
 
 #if !defined(TORRENT_DISABLE_ENCRYPTION) && !defined(TORRENT_DISABLE_EXTENSIONS)
-		virtual int hit_send_barrier(std::vector<boost::asio::mutable_buffer>& iovec) override;
+		// next_barrier
+		virtual
+		std::tuple<int, std::vector<boost::asio::const_buffer>>
+		hit_send_barrier(aux::array_view<boost::asio::mutable_buffer> iovec) override;
 #endif
 
 		virtual void get_specific_peer_info(peer_info& p) const override;
--- a/include/libtorrent/extensions.hpp
+++ b/include/libtorrent/extensions.hpp
@@ -173,6 +173,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/sha1_hash.hpp" // for sha1_hash
 #include "libtorrent/error_code.hpp"
 #include "libtorrent/session_handle.hpp"
+#include "libtorrent/aux_/array_view.hpp"
 
 namespace libtorrent
 {
@@ -186,6 +187,14 @@ namespace libtorrent
 	struct add_torrent_params;
 	struct peer_connection_handle;
 	struct torrent_handle;
+
+	// Functions of this type are called to handle incoming DHT requests
+	using dht_extension_handler_t = boost::function<bool(udp::endpoint const& source
+		, bdecode_node const& request, entry& response)>;
+
+	// Map of query strings to handlers. Note that query strings are limited to 15 bytes.
+	// see max_dht_query_length
+	using dht_extensions_t = std::vector<std::pair<std::string, dht_extension_handler_t> >;
 
 	// this is the base class for a session plugin. One primary feature
 	// is that it is notified of all torrents that are added to the session,
@@ -502,9 +511,9 @@ namespace libtorrent
 		// are now ready to be sent to the lower layer. This must be at least
 		// as large as the number of bytes passed in and may be larger if there
 		// is additional data to be inserted at the head of the send buffer.
-		// The additional data is retrieved from the passed in vector. The
-		// vector must be cleared if no additional data is to be inserted.
-		virtual int encrypt(std::vector<boost::asio::mutable_buffer>& /*send_vec*/) = 0;
+		// The additional data is returned as the second tupled value
+		virtual std::tuple<int, std::vector<boost::asio::const_buffer>>
+		encrypt(aux::array_view<boost::asio::mutable_buffer> /*send_vec*/) = 0;
 
 		// decrypt the provided buffers.
 		// consume is set to the number of bytes which should be trimmed from the
@@ -515,7 +524,7 @@ namespace libtorrent
 		//
 		// packet_size is set to the minimum number of bytes which must be read to
 		// advance the next step of decryption. default is 0
-		virtual void decrypt(std::vector<boost::asio::mutable_buffer>& /*receive_vec*/
+		virtual void decrypt(aux::array_view<boost::asio::mutable_buffer> /*receive_vec*/
 			, int& /* consume */, int& /*produce*/, int& /*packet_size*/) = 0;
 	};
 }
--- a/include/libtorrent/extensions.hpp
+++ b/include/libtorrent/extensions.hpp
@@ -188,14 +188,6 @@ namespace libtorrent
 	struct peer_connection_handle;
 	struct torrent_handle;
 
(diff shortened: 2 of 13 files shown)


include/libtorrent/bt_peer_connection.hpp
```diff
@@ -159,7 +159,10 @@ namespace libtorrent
 		void on_receive_impl(std::size_t bytes_transferred);
 
 #if !defined(TORRENT_DISABLE_ENCRYPTION) && !defined(TORRENT_DISABLE_EXTENSIONS)
-		virtual int hit_send_barrier(std::vector<boost::asio::mutable_buffer>& iovec) override;
+		// next_barrier
+		virtual
+		std::tuple<int, std::vector<boost::asio::const_buffer>>
+		hit_send_barrier(aux::array_view<boost::asio::mutable_buffer> iovec) override;
```
arvidn: @ssiloti this is part of the main change here. basically, instead of taking a reference to a std::vector as an in-out parameter, an array_view is used (to allow for flexible data structures being passed in, such as a stack allocated iovec) and the buffers to be prepended are returned as the second return value. This shifts the cost of heap allocation from *every time* to just when there are buffers to prepend.

This change obviously trickles through down to the crypto_plugin interface too.

src/bt_peer_connection.cpp
```diff
@@ -3545,15 +3520,19 @@ namespace libtorrent
...
+		int next_barrier;
+		std::vector<boost::asio::const_buffer> out_iovec;
+		std::tie(next_barrier, out_iovec) = m_enc_handler.encrypt(iovec);
 #ifndef TORRENT_DISABLE_LOGGING
 		if (next_barrier != 0)
 			peer_log(peer_log_alert::outgoing, "SEND_BARRIER"
 				, "encrypted block s = %d", next_barrier);
 #endif
-		return next_barrier;
+		return { next_barrier, out_iovec };
```
arvidn: pass on the out_iovec to the caller. Actually, I meant to say std::move() here. I will fix that.

src/pe_crypto.cpp
```diff
@@ -101,85 +103,99 @@ namespace libtorrent
 		m_xor_mask = h.final();
 	}
 
-	int encryption_handler::encrypt(std::vector<boost::asio::mutable_buffer>& iovec)
+	std::tuple<int, std::vector<boost::asio::const_buffer>>
+	encryption_handler::encrypt(
+		aux::array_view<boost::asio::mutable_buffer> iovec)
```
arvidn: This is probably the largest relevant hunk of this patch. I'll leave some comments.

src/pe_crypto.cpp
```diff
@@ -101,85 +103,99 @@ namespace libtorrent
...
 		int to_process = m_send_barriers.front().next;
 
+		boost::asio::mutable_buffer* bufs;
+		int num_bufs;
+		bool need_destruct = false;
 		if (to_process != INT_MAX)
 		{
-			for (std::vector<boost::asio::mutable_buffer>::iterator i = iovec.begin();
-				to_process >= 0; ++i)
+			bufs = TORRENT_ALLOCA(mutable_buffer, iovec.size());
```
arvidn: correct me if I'm wrong, but I believe we only get here if are asked to process fewer bytes than everything. In that case, the iovec we pass on to the underlying crypto-plugin is likely to be different from the one passed in. we copy it into a stack allocated array here. note that boost::asio::mutable_buffer is not a pod, those objects actually have to be constructed in place below, and destructed later.

src/pe_crypto.cpp
```diff
@@ -101,85 +103,99 @@ namespace libtorrent
...
+					new (&bufs[i]) mutable_buffer(iovec[i]);
+					to_process -= size;
 				}
-				to_process -= int(boost::asio::buffer_size(*i));
 			}
-			TORRENT_ASSERT(to_process == 0);
 		}
-
-#if TORRENT_USE_ASSERTS
-		to_process = 0;
```
arvidn: as far as I can see, the re-use of the "to_process" variable here is only used further down, in the next TORRENT_USE_ASSERTS block. I moved this down there, and also used a new variable, to not accidentally mix them or their scope up in the future

src/pe_crypto.cpp
```diff
@@ -101,85 +103,99 @@ namespace libtorrent
...
 		}
-		else
+		if (need_destruct)
 		{
-			iovec.clear();
+			for (int i = 0; i < num_bufs; ++i)
+				bufs[i].~mutable_buffer();
 		}
-		return next_barrier;
+		return { next_barrier, out_iovec };
```
arvidn: I meant to say std::move() here too.. will fix

src/pe_crypto.cpp
```diff
@@ -249,11 +265,11 @@ namespace libtorrent
...
 		rc4_init(key, len, &m_rc4_incoming);
 		// Discard first 1024 bytes
-		char buf[1024];
-		std::vector<boost::asio::mutable_buffer> vec(1, boost::asio::mutable_buffer(buf, 1024));
 		int consume = 0;
 		int produce = 0;
 		int packet_size = 0;
+		char buf[1024];
+		boost::asio::mutable_buffer vec(buf, sizeof(buf));
 		decrypt(vec, consume, produce, packet_size);
```
arvidn: array_view automatically interprets a reference as an array of 1 object

src/peer_connection.cpp
```diff
@@ -5510,15 +5510,18 @@ namespace libtorrent
 		{
 			std::vector<boost::asio::mutable_buffer> vec;
 			m_send_buffer.build_mutable_iovec(m_send_buffer.size(), vec);
-			int next_barrier = hit_send_barrier(vec);
-			for (std::vector<boost::asio::mutable_buffer>::reverse_iterator i = vec.rbegin();
-				i != vec.rend(); ++i)
+			int next_barrier;
+			std::vector<boost::asio::const_buffer> inject_vec;
+			std::tie(next_barrier, inject_vec) = hit_send_barrier(vec);
+			for (auto i = inject_vec.rbegin(); i != inject_vec.rend(); ++i)
```
arvidn: I believe the semantics are the same here as before, but I don't believe there's a test, so i can't say for sure. @ssiloti does it look OK to you? Does this interface still support what you need in the reputation plugin?

arvidn:
@ssiloti would you mind taking a look? this alters your patch originally and the crypto_plugin interface. Sorry for it being a bit diluted by unrelated cosmetic changes.

include/libtorrent/extensions.hpp
```diff
@@ -187,6 +188,14 @@ namespace libtorrent
 	struct peer_connection_handle;
 	struct torrent_handle;
 
+	// Functions of this type are called to handle incoming DHT requests
+	using dht_extension_handler_t = boost::function<bool(udp::endpoint const& source
```
aldenml: I think `dht_extension_handler_t` and `dht_extensions_t` are gone, right?

aldenml:
@arvidn I have one API related observation, `array_view` is very internal to be used in a plugin.

include/libtorrent/extensions.hpp
arvidn: oops. I think this was a rebase mis-merge. will fix

arvidn:
you're right. that's a good point. I'm sort of expecting my array_view to be replaced either by a boost library or the gsl::span. but for now maybe it need to be made public.

include/libtorrent/bt_peer_connection.hpp
ssiloti: Why not make the returned buffers an array_view as well? The buffers referenced in the returned vector are required to live until the next call to encrypt so it would make sense to apply the same rule to the vector itself.

src/pe_crypto.cpp
```diff
@@ -101,85 +103,99 @@ namespace libtorrent
...
+		else
+		{
+			bufs = iovec.data();
+			num_bufs = iovec.size();
+		}
 
 		int next_barrier = 0;
-		bool process_barrier = iovec.empty();
+		std::vector<const_buffer> out_iovec;
+		bool process_barrier = iovec.size() == 0;
```
ssiloti: This should reference `bufs` instead of `iovec`.

src/peer_connection.cpp
ssiloti: LGTM

include/libtorrent/bt_peer_connection.hpp
arvidn: seems reasonable

src/pe_crypto.cpp
arvidn: good catch

ssiloti:
lgtm

aldenml:
would you consider not make `array_view` public and instead work (or accept a PR) with a more primitive solution similar to `readv(file::iovec_t const* bufs, int num_bufs...` with a in/out parameter style? (only for the plugin API). This and `buffer::const_interval` are the remaining private types in the public plugin API.

arvidn:
for now I'm considering just merging this as-is. it's a header-only type, so it doesn't need to be exported. At some point I will look into using boost::multi_array_ref<T, 1>, and probably set up a typedef to array_view.

Not using an array view type in the API would be very unfortunate I think, since it provides a very useful abstraction (since I started using gdl::span I've been quite surprised how much such a simple class can simplify code)

PR #837, opened 2016-06-20
removed buffer::interval

aldenml:
This is soft PR, it's just a refactor to eliminate what I think is a redundant type. If you have something else in mind, feel free to discard/close it.

### DIFF
--- a/include/libtorrent/bt_peer_connection.hpp
+++ b/include/libtorrent/bt_peer_connection.hpp
@@ -286,7 +286,7 @@ namespace libtorrent
 			, char const* target, int target_size) const;
 
 		// helper to cut down on boilerplate
-		void rc4_decrypt(char* pos, int len);
+		void rc4_decrypt(boost::asio::mutable_buffer vec);
 #endif
 
 public:
--- a/include/libtorrent/buffer.hpp
+++ b/include/libtorrent/buffer.hpp
@@ -46,42 +46,8 @@ namespace libtorrent {
 class buffer
 {
 public:
-	struct interval
-	{
-	interval()
-		: begin(0)
-		, end(0)
-		{}
-
-	interval(char* b, char* e)
-		: begin(b)
-		, end(e)
-		{}
-
-		char operator[](int index) const
-		{
-			TORRENT_ASSERT(begin + index < end);
-			return begin[index];
-		}
-
-		int left() const
-		{
-			TORRENT_ASSERT(end >= begin);
-			TORRENT_ASSERT(end - begin < (std::numeric_limits<int>::max)());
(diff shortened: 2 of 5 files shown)


arvidn:
I like the spirit of this. I think the const_interval could also be replaced though. I'm also not sure about the replacement. Do you think there are good reasons not to use array_view instead?
I could even imagine using array_view instead of asio::mutable_buffer and asio::const_buffer (and create overloads for buffer_cast<>() and buffer_size(), but I'm not sure what that would look like exactly).

aldenml:
Removing `const_interval` is my second step if you accept this PR. I like `array_view` but the template nature of it keep me worried about more use of it in public API. Maybe if you create an specialized child type with a proper `TORRENT_EXPORT` macro, I would be more confortable.

btw, could you restart this: https://travis-ci.org/arvidn/libtorrent/jobs/138946951

arvidn:
what do you mean by "child type"? lots of template classes are used in the API.. std::vector, boost::optional and std::chrono::steady_clock::time_point/time_duration for instance. what specifically is your concern? that a DLL might instantiate and export part of it, while a client uses the header file, getting a duplicate definition?

aldenml:
I mean something along the lines of `struct TORRENT_EXPORT buffer_view : array_view<int8_t>`.

Remember that my point of view here is more from outside C++ and how to integrate with the native library. `std::vector` is so used (and abused) that a lot of special cases are made to handle this template, and the others you mention are a little problematic to represent (including tuple). I rather research a little more before give an opinion in the duplicate definition issue.

What do you think is the issue with the failed travis job? could it be related to this change?

aldenml:
Hi, do you want me to do something else here?

arvidn:
I'm not sure about using the boost asio type. I feel like using the array_view would be more appropriate (since it's more generic). In fact, it may even make sense to replace the asio buffers with array_view (and add buffer_cast<>() and buffer_size() overloads for it). Now, I haven't seen what it would look like, so I can't say for sure I necessarily will like it. But in general, I think it's a good idea to coalesce these types that basically do the same thing. The DHT has it's own version too (but just uses std::pair<char*, int>)

aldenml:
Ok, I think I can workaround `array_view` in swig, maybe you should move it out of `aux` namespace. Is that ok?

arvidn:
yeah, I will try to replace it with a typedef of boost's multi array view. hopefully this weekend

arvidn:
I would imagine you could bind specific template instantiations in swig, no?

aldenml:
Yes, I can do that (it could get a little tricky if the template is not simple). Closing this.

PR #838, opened 2016-06-20
don't risk use of invalid data during fast resume check

aldenml:
This is another attempt to fix https://github.com/arvidn/libtorrent/issues/781

The explanation of this change is: When you don't use `flag_duplicate_is_error` it seems to me that is possible that two successively and rapidly `add_torrent` calls with the same parameters result in calling `async_check_fastresume` twice, since the first one still has not finished the async task, the `m_resume_data` is still not cleared at the time of the second call. The second job will eventually reference an invalid pointer.

(I personally don't like the adding of another flag, can't see other option)

### DIFF
--- a/src/disk_io_thread.cpp
+++ b/src/disk_io_thread.cpp
@@ -1865,7 +1865,7 @@ namespace libtorrent
 	}
 
 	void disk_io_thread::async_check_fastresume(piece_manager* storage
-		, bdecode_node const* resume_data
+		, resume_data_t const* resume_data
 		, std::vector<std::string>& links
 		, boost::function<void(disk_io_job const*)> const& handler)
 	{
@@ -2573,12 +2573,12 @@ namespace libtorrent
 		// if this assert fails, something's wrong with the fence logic
 		TORRENT_ASSERT(j->storage->num_outstanding_jobs() == 1);
 
-		bdecode_node const* rd = j->buffer.check_resume_data;
+		resume_data_t const* rd = j->buffer.check_resume_data;
 		bdecode_node tmp;
-		if (rd == NULL) rd = &tmp;
 
 		boost::scoped_ptr<std::vector<std::string> > links(j->d.links);
-		return j->storage->check_fastresume(*rd, links.get(), j->error);
+		return j->storage->check_fastresume(rd != NULL ? rd->node : tmp
+			, links.get(), j->error);
 	}
 
 	int disk_io_thread::do_save_resume_data(disk_io_job* j, jobqueue_t& completed_jobs)
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -201,7 +201,6 @@ namespace libtorrent
 		, m_uuid(p.uuid)
 		, m_source_feed_url(p.source_feed_url)
 		, m_stats_counters(ses.stats_counters())
-		, m_check_fastresume_queued(false)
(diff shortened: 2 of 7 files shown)


arvidn:
hm.. when adding a torrent without ``duplicate_is_error``, it just returns a reference to the existing torrent, right? it doesn't re-issue the check_resume_data job or anything, does it?

aldenml:
If you follow the logic from this point: https://github.com/arvidn/libtorrent/blob/RC_1_1/src/session_impl.cpp#L4787
```
boost::shared_ptr<torrent> const torrent_ptr = add_torrent_impl(params, ec);
```
it goes on and on, without knowing if the torrent handle is new or a reference to an existing one

src/torrent.cpp
```diff
@@ -2100,9 +2101,10 @@ namespace libtorrent
 		inc_refcount("check_fastresume");
 		// async_check_fastresume will gut links
 		m_ses.disk_thread().async_check_fastresume(
-			m_storage.get(), m_resume_data ? &m_resume_data->node : NULL
+			m_storage.get(), m_resume_data && !m_check_fastresume_queued ? &m_resume_data->node : NULL
```
arvidn: I'm a bit hesitant to add another piece of state. if we no longer need the resume data in the torrent, maybe it could be moved into the job somehow. ideally without being copied. it's held by a scoped_ptr now. it could be changed to a unique_ptr, and it could be released here, and captured again in the disk job.

aldenml: ok, I will try to instrument that

aldenml:
wrong commit

arvidn:
actually, I think I may have introduced a bug when I refactored that. start() should not be called more than once, and plugins should not be added more than once either. I will have to look at that

aldenml:
Ok, cancel any CI job related to this PR if you need to

src/disk_io_thread.cpp
```diff
@@ -1865,7 +1865,7 @@ namespace libtorrent
 	}
 
 	void disk_io_thread::async_check_fastresume(piece_manager* storage
-		, bdecode_node const* resume_data
+		, resume_data_t const* resume_data
```
arvidn: this can probably not be const I'm guessing.. since this is an owning pointer now

src/torrent.cpp
```diff
@@ -2330,11 +2331,11 @@ namespace libtorrent
 
 		state_updated();
 
-		if (m_resume_data && m_resume_data->node.type() == bdecode_node::dict_t)
+		if (resume_data && resume_data->node.type() == bdecode_node::dict_t)
```
arvidn: I see. we do use it when the disk job completes.

arvidn: I think a lot, if not all, of this is done earlier in master. this change seems likely to conflict with master (i.e. difficult to bring forward)

arvidn:
I'll think a bit more about this approach. if you use the duplicate-is-not-an-error flag, the root cause is probably in the add_torrent_impl refactor though. hopefully I'll have time to look at that tomorrow

aldenml:
I rather go for the change that fix the root cause. I wasn't sure what the design in this case was, but I feel confident about the identification of the issue. Looking forward for your analysis.

aldenml:
Closing in favor of https://github.com/arvidn/libtorrent/pull/846

PR #845, opened 2016-06-22
option to use only the openssl's sha1 implementation

aldenml:
This allow me to not enable the torrent-over-ssl feature and still use the speedy SHA1 implementation. The change in `session_impl` is to no open the ssl sockets if openssl is not enable.


arvidn:
you can disable support for SSL torrents by setting the ssl listen port to 0 (or something like that, in RC_1_1). In master it's just a matter of not specifying an SSL listen port in listen_interfaces. I believe listening on SSL is disabled by default in both master and RC_1_1.

To exclude it at build time, is the concern additional code size of including a feature you don't need or use?
I'm a little bit worried about the combinatorical explosion of build configurations. I've been trying to reduce them.

aldenml:
Yes, it's the code size the main concern here. OpenSSL could be compiled to the bare minimum or even a custom build with only SHA1. Having to link with both `libssl` and `libcrypto` only to take advantage of the SHA1 requires a lot of code related to certificates/stores that can't be stripped off.

Also, I believe the current boost openssl integration is only compatible with `1.0.X`, but with this PR you are not using boost-ssl and still using the SHA1 (with the assembly code).

arvidn:
would it make sense perhaps to break this down into individual switches for libcrypto and libssl?
the Jamfile could error if libssl is enabled and libcrypto is disabled. In some sense it would fit with better with what's implemented for libgcrypt too, since there's no support for gnutls.

what do you think about these options?
  ssl=openssl | none
  crypto=libcrypto | libgcrypt | builtin | commoncrypto

maybe "commoncrypto" isn't a good name, but I believe Apples API is basically the same as libcrypto.

arvidn:
btw. I thought openssl came shipped with android, and that you can just link dynamically against it

aldenml:
I like that better, but I was trying to introduce no new bjam option :) I will change this with the new `ssl` and `crypto` options.

As for android:

> 
Android is moving away from OpenSSL to the BoringSSL library. If you’re using the Android NDK in your app, don't link against cryptographic libraries that are not a part of the NDK API, such as libcrypto.so and libssl.so. These libraries are not public APIs, and may change or break without notice across releases and devices. In addition, you may expose yourself to security vulnerabilities. Instead, modify your native code to call the Java cryptography APIs via JNI or to statically link against a cryptography library of your choice.

Jamfile
```diff
@@ -56,15 +56,52 @@ rule linking ( properties * )
 {
 	local result ;
 
+	# ssl=openssl => crypto=libcrypto
+	# this match the previous behavior
+	if <ssl>openssl in $(properties)
+		&& ! <crypto>libcrypto in $(properties)
+	{
+		EXIT "with ssl=openssl the value for crypto must be libcrypto" ;
```
arvidn: a nice feature of boost build is that properties are associated with build targets, not with invocations of bjam. what do you think of printing this message (with ECHO) and then add <build>no to result. That would simply disable the targets with invalid/unsupported combinations.

for instance, it would be possible to run: bjam ssl=openssl,none crypto=libcrypto,builtin,libgcrypt
and it would build all combinations of those features that are valid (at least that's my understanding of <build>no)

Jamfile
```diff
@@ -56,15 +56,52 @@ rule linking ( properties * )
...
+				<library>libeay32
+				<library>advapi32
+				<library>user32
+				<library>shell32
+				<library>gdi32
+			;
+		}
+		else
+		{
+			result += <library>crypto <library>ssl ;
```
arvidn: shouldn't this just add "ssl", and the crypto feature would add libcrypto?

Jamfile
```diff
@@ -56,15 +56,52 @@ rule linking ( properties * )
...
+	if <crypto>libcrypto in $(properties)
+		&& ! <ssl>openssl in $(properties)
 	{
 		# exclude gcc from a regular windows build to make mingw
 		# link against the regular unix library name
 		if <target-os>windows in $(properties)
 			&& ! <toolset>gcc in $(properties)
 		{
+			# it should be possible to cleanup this list, but this is safe for now
 			result += <library>ssleay32
```
arvidn: I imagine ssleay isn't necessary here.

Jamfile
```diff
@@ -56,15 +56,52 @@ rule linking ( properties * )
...
+			;
+		}
+		else
+		{
+			result += <library>crypto <library>ssl ;
+		}
+	}
+
+	if <crypto>libcrypto in $(properties)
+		&& ! <ssl>openssl in $(properties)
```
arvidn: I don't think you need to care about the <ssl> feature here. if you make them independent.

Jamfile
```diff
@@ -103,7 +133,7 @@ rule linking ( properties * )
 	}
 
 	# gcrypt libraries, if enabled
-	if <crypto>gcrypt in $(properties)
+	if <crypto>libgcrypt in $(properties)
```
arvidn: do you think libgcrypt is better than gcrypt? I realize I've been using the lib prefix. maybe it would be more conservative to preserve the gcrypt name and use crypto for libcrypto.

aldenml: :)

Jamfile
```diff
@@ -463,10 +493,13 @@ feature.compose <encryption>off : <define>TORRENT_DISABLE_ENCRYPTION ;
 feature mutable-torrents : on off : composite propagated link-incompatible ;
 feature.compose <mutable-torrents>off : <define>TORRENT_DISABLE_MUTABLE_TORRENTS ;
 
-feature crypto : built-in openssl openssl-sha1 gcrypt : composite propagated ;
-feature.compose <crypto>openssl : <define>TORRENT_USE_OPENSSL <define>OPENSSL_NO_SSL2 ;
-feature.compose <crypto>openssl-sha1 : <define>TORRENT_USE_OPENSSL_SHA1 ;
-feature.compose <crypto>gcrypt : <define>TORRENT_USE_GCRYPT ;
+feature crypto : built-in libcrypto libgcrypt commoncrypto : composite propagated ;
```
arvidn: interesting. I guess I didn't have commonCrypto as a build feature. maybe it would make sense to just fold that into built-in. when building for mac/iOS I imagine it's fine (if not recommended) to just use that

include/libtorrent/config.hpp
```diff
@@ -134,13 +134,6 @@ POSSIBILITY OF SUCH DAMAGE.
 # endif
 #include <AvailabilityMacros.h>
 
-#if MAC_OS_X_VERSION_MIN_REQUIRED >= 1070
-// on OSX, use the built-in common crypto for built-in
-# if !defined TORRENT_USE_OPENSSL && !defined TORRENT_USE_GCRYPT
-#  define TORRENT_USE_COMMONCRYPTO 1
```
arvidn: I'm sorry. but I think it's nicer to detect this automatically here. This seems to provide the semantics I proposed above. If we're using built-in crypto on a mac, use the official API.

aldenml: I agree

arvidn:
I like the separation of crypto library and ssl library. I think they could be made more independent. I'm also thinking that built-in should probably be anything that doesn't add any external dependencies. meaning the macOS/iOS API is fine, as well as potentially using windows API directly in the future

Jamfile
aldenml: What can I say, I code in bjam like I'm learning an alien language, let me research this

Jamfile
aldenml: I think libgcrypt is better, that's the name of the library

aldenml:
I will address your comments. If you have a chance, cancel all the outstanding CI jobs related to this PR to make room for the others

Jamfile
arvidn: oh, it seems like github ate my brackets. I meant to say: "<build>no"

arvidn: the key thing with bjam is that there is no _global_ state for "the build". There is no "the build". All state (build features/configs/properties) are associated with a specific target. You can have multiple builds of the same target at the same time, but with different configurations. like, you can build debug and release at the same time. if you would keep global state about which one you're building, it would break.

Jamfile
arvidn: alright, libgcrypt it is

Jamfile
aldenml: Thank you! some progress with this info

src/session_impl.cpp
```diff
@@ -1838,7 +1838,12 @@ namespace aux {
 		{
 			std::string const& device = m_listen_interfaces[i].device;
 			int const port = m_listen_interfaces[i].port;
-			bool const ssl = m_listen_interfaces[i].ssl;
+			bool const ssl =
+#ifdef TORRENT_USE_OPENSSL
```
arvidn: I'm not sure about silently ignoring the request to listen to an SSL socket and still open a socket but _without_ SSL? That could even be a security risk. What do you think about the following instead:

1. post a listen_alert_error with operation_not_supported error
2. skip this port, through "continue".

aldenml: You are right, I misunderstood the current logic

include/libtorrent/config.hpp
```diff
@@ -386,10 +386,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #define TORRENT_DEPRECATED
 #endif
 
-#ifndef TORRENT_USE_COMMONCRYPTO
-#define TORRENT_USE_COMMONCRYPTO 0
-#endif
-
```
arvidn: the benefit of definint macros to 0 or 1, instead of defining them or not defining them, is that the compiler can check spelling and warn if there's a mistake. Right now the convention is for configure macros set up by config.hpp are always defined, and only the ones from the outside are either defined or not. If anything, I would prefer to move towards every configuration macros always being defined to either 0 or 1 (unless you can think of a good reason not to)

Jamfile
```diff
@@ -455,9 +484,12 @@ feature.compose <encryption>off : <define>TORRENT_DISABLE_ENCRYPTION ;
 feature mutable-torrents : on off : composite propagated link-incompatible ;
 feature.compose <mutable-torrents>off : <define>TORRENT_DISABLE_MUTABLE_TORRENTS ;
 
-feature crypto : built-in openssl gcrypt : composite propagated ;
-feature.compose <crypto>openssl : <define>TORRENT_USE_OPENSSL <define>OPENSSL_NO_SSL2 ;
-feature.compose <crypto>gcrypt : <define>TORRENT_USE_GCRYPT ;
+feature crypto : built-in libcrypto libgcrypt : composite propagated ;
+feature.compose <crypto>libcrypto : <define>TORRENT_USE_LIBCRYPTO ;
+feature.compose <crypto>libgcrypt : <define>TORRENT_USE_LIBGCRYPT ;
```
arvidn: would you mind updating the documentation under docs/build.rst as well?

arvidn:
thanks! looks great. did you see bjam correctly stopping the build of unsupported configurations btw?

aldenml:
Yes, it prints and don't build, but I didn't test multiple targets at the same time, also not sure how it resolves the names in the case of multiple targets.

PR #846, opened 2016-06-22
correct error introduced in add_torrent_impl refactor, with regards t…

arvidn:
…o not having flag_duplicate_is_error set (which is not set by default)

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -4784,7 +4784,9 @@ retry:
 	{
 		// params is updated by add_torrent_impl()
 		add_torrent_params params = p;
-		boost::shared_ptr<torrent> const torrent_ptr = add_torrent_impl(params, ec);
+		boost::shared_ptr<torrent> torrent_ptr;
+		bool added;
+		boost::tie(torrent_ptr, added) = add_torrent_impl(params, ec);
 
 		torrent_handle const handle(torrent_ptr);
 		m_alerts.emplace_alert<add_torrent_alert>(handle, params, ec);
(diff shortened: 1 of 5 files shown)


src/session_impl.cpp
```diff
@@ -5086,19 +5093,19 @@ namespace aux {
 					torrent_ptr->set_url(params.url);
 				if (!params.source_feed_url.empty() && torrent_ptr->source_feed_url().empty())
 					torrent_ptr->set_source_feed_url(params.source_feed_url);
-				return torrent_ptr;
+				return std::make_pair(torrent_ptr, false);
```
gubatron: does this mean...
`if ((params.flags & add_torrent_params::flag_duplicate_is_error) == 0)`

that duplicates aren't errors?

If so, is it ok to put `false` in the second element of the pair?

gubatron: (I say this because I notice you put `false` whenever there's an error)

src/session_impl.cpp
```diff
@@ -5086,19 +5093,19 @@ namespace aux {
...
 			}
 
 			ec = errors::duplicate_torrent;
-			return ptr_t();
+			return std::make_pair(ptr_t(), false);
```
gubatron: here it looks good to me.

src/session_impl.cpp
gubatron: hmm, I see the boolean probably means `started`

aldenml:
It's not clear to me why the line `if (!added) return handle;` is placed where it is. It is after the "add_peers" logic, but before the "add_dht_nodes" logic. I don't see any significant issue here, just to left the observation.

src/session_impl.cpp
arvidn: yeah, it's mimicking std::set<>::insert()

arvidn:
yeah, adding the DHT nodes should probably be done earlier. It's important to exit here though before calling start() and adding extensions, things that actually violate the invariants.

aldenml:
@arvidn try to clean the cache for https://travis-ci.org/arvidn/libtorrent/jobs/139678216

PR #848, opened 2016-06-23
open default listen interface the same as no default

aldenml:


src/session_impl.cpp
```diff
@@ -1467,7 +1461,7 @@ namespace aux {
 					!= m_settings.get_int(settings_pack::ssl_listen))
 			||
 #endif
-			(pack.has_val(settings_pack::listen_interfaces)
+			init || (pack.has_val(settings_pack::listen_interfaces)
```
arvidn: instead of injecting "init" here, you would get the same semantics by putting it in the if (reopen_listen_port) condition, right? (that seems like it would make things a bit simpler, this is already a complex expression)

src/session_impl.cpp
```diff
@@ -1458,7 +1451,8 @@ namespace aux {
 		return ret;
 	}
 
-	void session_impl::apply_settings_pack_impl(settings_pack const& pack)
+	void session_impl::apply_settings_pack_impl(settings_pack const& pack
+		, bool const init)
 	{
 		bool reopen_listen_port =
```
arvidn: would you mind making this const while you're at it?

src/session_impl.cpp
```diff
@@ -1476,6 +1470,8 @@ namespace aux {
 
 		if (reopen_listen_port)
 		{
+			if (init)
```
arvidn: this condition could be hoisted out of the reopen_listen_port if-block, right? (you save one nesting level and get slightly simpler logic)

arvidn:
what's the motivation for this change? does it change any behavior?

src/session_impl.cpp
aldenml: sure, I wanted to save a bit of CPU instructions :)

aldenml:
It does, if the `listen_interfaces` is not the default and all of them fails (for example due to lack of `ssl` support), the deleted code below triggered the logic again, only to report twice the same failed alerts. Also, it's good to know that no matter default values or not, the listen logic happens before the set of calls to `update_*`.

arvidn:
makes sense

aldenml:
if you have chance, I need your intervention in travis, maybe the cache clear stopped some jobs

PR #851, opened 2016-06-24
fix msvc-8.0 build

### LINKED ISSUE #850: fileop compile failed

libtorrent version (or branch): RC_1_1

platform/architecture:
win10 x64

compiler and compiler version:
vc++ 2005

`struct fileop
{
virtual int file_op(int file_index, boost::int64_t file_offset, int size
			, file::iovec_t const* bufs, storage_error& ec) = 0;
}
`
declared in `include/libtorrent/storage.hpp`

but the implement in `/src/storage.cpp`  is 
`int file_op(int const file_index
			, boost::int64_t const file_offset
			, int const size
			, file::iovec_t const* bufs, storage_error& ec)`

I compile failed because the difference between them.

k1988:
fix #850 and some other error

### DIFF
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -406,7 +406,10 @@ namespace libtorrent
 #endif
 
 			torrent_handle add_torrent(add_torrent_params const&, error_code& ec);
-			boost::shared_ptr<torrent> add_torrent_impl(add_torrent_params& p, error_code& ec);
+			// second return value is true if the torrent was added and false if an
+			// existing one was found.
+			std::pair<boost::shared_ptr<torrent>, bool>
+			add_torrent_impl(add_torrent_params& p, error_code& ec);
 			void async_add_torrent(add_torrent_params* params);
 			void on_async_load_torrent(disk_io_job const* j);
 
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -1717,6 +1717,9 @@ namespace libtorrent
 	public:
 		// set to false until we've loaded resume data
 		bool m_resume_data_loaded;
+
+		// set to true when torrent is start()ed. It may only be started once
+		bool m_was_started;
 #endif
 	};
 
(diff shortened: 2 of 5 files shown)


src/crc32c.cpp
```diff
@@ -46,7 +46,7 @@ namespace libtorrent
 {
 	boost::uint32_t crc32c_32(boost::uint32_t v)
 	{
-#if TORRENT_HAS_SSE
+#if (TORRENT_HAS_SSE && defined _MSC_VER && _MSC_VER >= 1600)
```
arvidn: it seems cleaner to just consider _MSC_VER < 1600 to not have SSE, don't you think? that would be a single place to change

include/libtorrent/storage.hpp
```diff
@@ -695,7 +695,9 @@ namespace libtorrent
 	// what to do when it's actually touching the file
 	struct fileop
 	{
-		virtual int file_op(int file_index, boost::int64_t file_offset, int size
+		virtual int file_op(int const file_index
```
arvidn: I would prefer to change the .cpp file. the fact that the function doesn't change its incoming arguments should not leak out into its interface

src/bitfield.cpp
```diff
@@ -59,7 +59,7 @@ namespace libtorrent
 	{
 		int ret = 0;
 		const int words = num_words();
-#if TORRENT_HAS_SSE
+#if (TORRENT_HAS_SSE && defined _MSC_VER && _MSC_VER >= 1600)
```
aldenml: I don't think this is right, what happens to `gcc`?

aldenml:
I think you should reconsider the addition of the few `defined _MSC_VER && _MSC_VER >= 1600`

src/bitfield.cpp
arvidn: good point. this is a problem in a few other places below as well. maybe it would make more sense to make TORRENT_HAS_SSE be set to false for old versions of msvc

k1988: oh, sorry, i forgot it !
I only thought about vs.

bug only make TORRENT_HAS_SSE be set to false for old versions of msvc is not enough, because _mm_crc32_u32 is not defined。

k1988: Sorry, I didn't consid that!
i will change the PR。

PR #852, opened 2016-06-24
using bsr even with no popcnt present

aldenml:
For context, this is part of an ongoing research to include support for arm crc32c, clz and popcnt.

I don't see why `bsrl` needs to by tied to the presence of `popcnt`, any reason for that?

### DIFF
--- a/src/sha1_hash.cpp
+++ b/src/sha1_hash.cpp
@@ -76,16 +76,13 @@ namespace libtorrent
 			}
 
 #if TORRENT_HAS_SSE
-			if (aux::mmx_support)
-			{
 #ifdef __GNUC__
-				return ret + __builtin_clz(v);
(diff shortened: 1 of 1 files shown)


arvidn:
TORRENT_HAS_SSE will only be set on x86 and amd64. mmx_support is set based on the cpuinfo, indicating whether these instructions are supported or not. I don't see how this would affect arm. presumably there would have to be a separate define indicating support for such instructions there (unless all arm versions have these instructions)

aldenml:
The arm thing is for context, nothing here yet related to that. But the `bsrl` instruction has nothing to do with the `popcnt` (and I assume the name `mmx` is just a name, this have nothing to do with MMX).

arvidn:
I see. it's not an extension.

PR #853, opened 2016-06-25
simplify peer_connection's read handlers and the receive buffer.

arvidn:
removed the concept of soft-packet-size. removed the secondary null-buffer receive path. removed try_read

### DIFF
--- a/src/peer_connection.cpp
+++ b/src/peer_connection.cpp
@@ -299,6 +299,8 @@ namespace libtorrent
 
 		if (!m_outgoing)
 		{
+			m_recv_buffer.reserve(300);
+
 			tcp::socket::non_blocking_io ioc(true);
 			error_code ec;
 			m_socket->io_control(ioc, ec);
@@ -5409,10 +5411,11 @@ namespace libtorrent
 		}
 	}
 
-	int peer_connection::request_bandwidth(int channel, int bytes)
+	int peer_connection::request_bandwidth(int const channel, int bytes)
 	{
 		TORRENT_ASSERT(is_single_thread());
 		INVARIANT_CHECK;
+
 		// we can only have one outstanding bandwidth request at a time
 		if (m_channel_state[channel] & peer_info::bw_limit) return 0;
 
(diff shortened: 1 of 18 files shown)


src/receive_buffer.cpp
```diff
@@ -41,44 +41,66 @@ namespace libtorrent {
...
 }
 
+void receive_buffer::grow(int const limit)
+{
+	int const current_size = int(m_recv_buffer.size());
+	TORRENT_ASSERT(current_size < std::numeric_limits<int>::max() / 3);
+
+	// first grow to one piece message, then grow by 50% each time
+	int const new_size = std::min((current_size < m_packet_size)
+		? m_packet_size : current_size * 3 / 2, limit);
```
ssiloti: What if `m_packet_size` exceeds `limit`?

arvidn: yeah, I was going to restrict what limit can be set to. that was my original plan at least. the thing is that m_packet_size is read from the stream, so it kind of makes sense to still apply the limit to it I think

ssiloti: I don't think this is the place the enforce that limit though. I suspect that if you do get into a situation where `m_packet_size > limit` this code will cause the connection to stall. The limit on m_packet_size needs to be enforced further up the call stack where we can handle it gracefully.

The bt packet size is already capped to 1MB by `bt_peer_connection`. The one hole I'm aware of is when authenticated crypto is in use we don't cap the crypto packet size. Running over loopback I've seen crypto packets well over 1MB so the limit will need to be enforced on both the send and receive side.

src/peer_connection.cpp
```diff
@@ -5683,94 +5692,27 @@ namespace libtorrent
...
-		}
+		int const quota_left = m_quota[download_channel];
+		int const max_receive = (std::min)(buffer_size, quota_left);
 
-		int max_receive = m_recv_buffer.max_receive();
-
-		// only apply the contiguous receive buffer when we don't have any
-		// outstanding requests. When we're likely to receive pieces, we'll
-		// save more time from avoiding copying data from the socket
-		if (m_download_queue.empty())
```
arvidn: the whole shenanigan with null_buffers and looping to drain the socket and try_read() and read_data_impl() was removed and replaced by a single async callback issued from setup_read(). The callback drains the socket (only if the receive buffer was saturated) by first asking for the number of bytes available to be read, and then issuing a single synchronous read() call.

src/receive_buffer.cpp
arvidn: that's a good point. I'll make the packet size trump the limit.

arvidn:
rebased to make it build again..

PR #854, opened 2016-06-25
added arm __builtin_clz support and tests running in arm mode

aldenml:


src/sha1_hash.cpp
```diff
@@ -75,14 +75,12 @@ namespace libtorrent
 				continue;
 			}
 
-#if TORRENT_HAS_SSE
-#ifdef __GNUC__
+#if TORRENT_HAS_BUILTIN_CLZ
 			return ret + __builtin_clz(v);
-#else
+#elif TORRENT_HAS_SSE && defined _MSC_VER
```
arvidn: I got the impression from your last commit that this isn't an SSE feature, so maybe it shouldn't be tested here

tools/test_arm.cpp
```diff
@@ -0,0 +1,94 @@
...
+		{ "00000001fff00000000000000000000000000000", 31 },
+		{ "00000000fff00000000000000000000000000000", 32 },
+		{ "000000007ff00000000000000000000000000000", 33 },
+		{ "000000003ff00000000000000000000000000000", 34 },
+		{ "000000001ff00000000000000000000000000000", 35 },
+	};
+
+	for (auto const& t : clz_tests)
+	{
+		TORRENT_ASSERT(to_hash(t.first).count_leading_zeroes() == t.second);
```
arvidn: shouldn't this be part of the test_sha1_hash.cpp?

tools/test_arm.cpp
```diff
@@ -0,0 +1,94 @@
...
+{
+	std::cout << "running test arm" << std::endl;
+
+	bool has_builtin_clz =
+#if TORRENT_HAS_BUILTIN_CLZ
+		true;
+#else
+		false;
+#endif
+	TORRENT_ASSERT(has_builtin_clz);
```
arvidn: if you want to assert a compile time property, you could just do:

#ifndef TORRENT_HAS_BUILTIN_CLZ
#error "expected built-in clz"
#endif

include/libtorrent/config.hpp
```diff
@@ -518,6 +518,23 @@ POSSIBILITY OF SUCH DAMAGE.
...
+#endif // TORRENT_HAS_ARM
+
+#ifndef __has_builtin
+#define __has_builtin(x) 0  // for non-clang compilers
+#endif
+
+#if (TORRENT_HAS_SSE && __GNUC__) \
+	|| (TORRENT_HAS_ARM && defined __GNUC__ && !defined __clang__) \
+	|| (defined __clang__ && __has_builtin(__builtin_clz))
+#define TORRENT_HAS_BUILTIN_CLZ 1
```
arvidn: instead of having a complex condition here, what would you think of defining TORRENT_HAS_BUILTIN_CLZ right after TORRENT_HAS_SSE 1 is defined. then this can just be another condition under which to define it as 1. then at the end, if it's not defined, define it to 0 (a lot of other macros are defined this way, at the end they are defaulted if they haven't been set by any rule). Obviously it requires the conditions to set it to 1 to be mutually exclusive to avoid duplicate definitions.

tools/test_arm.cpp
```diff
@@ -0,0 +1,94 @@
+/*
```
arvidn: I don't really get why this test is here, separated from the regular tests. If this is a test that's handy to have around but can't run as part of the test suite, I would prefer it still lives under tests but with an explicit build target (like test_upnp and test_natpmp)

.travis-arm.sh
```diff
@@ -0,0 +1,25 @@
...
+wget -O boost.zip http://pilotfiber.dl.sourceforge.net/project/boost/boost/1.55.0/boost_1_55_0.zip
+unzip -qq boost.zip
+export BOOST_ROOT=$PWD/boost_1_55_0
+
+echo "using gcc : arm64 : aarch64-linux-gnu-gcc : <cflags>-std=c11 <cxxflags>-std=c++11 <cxxflags>-fsigned-char <linkflags>-lstdc++ <linkflags>-lm ;" >> ~/user-config.jam
+cd tools
+bjam warnings-as-errors=on variant=test_arm toolset=gcc-arm64 target-os=linux link=static location=./bin/gcc-arm64
+cd ..
+
+sudo apt-get install -y qemu-user-static debootstrap
```
arvidn: I take it the tests will be run under QEMU, will the actual build be run under the emulator too?
(if so, once travis has boost-1.59 or later, it's possible to separate the build and run-tests stages, which means the build could be done by the host)

arvidn:
there's a build property in boost.build called "launcher" which, when specified, is used as the program to run the tests via. it can be used for running tests under valgrind for instance. Maybe it could be used to launch the tests under QEMU too..

src/sha1_hash.cpp
aldenml: The name is misleading, the macro only test for essentially `x86` and `x86_64` and this days `msvc` has support for `arm` too. In fact, the redundant test is for `_MSC_VER`, I wrote it as it is for clarity.

tools/test_arm.cpp
aldenml: I'm not sure what to do here, I copy/pasted the code from `test_sha1_hash.cpp`

tools/test_arm.cpp
aldenml: (comments below)

.travis-arm.sh
aldenml: (comments below)

aldenml:
Let me explain why I coded the test this way.

First, I have little to no idea how exactly `boost.build` works :) and was so much easy to dig in `tools/Jamfile` than in `tests/Jamfile`, so I expected these observations (and some guide if possible).

The actual build runs in the host (actually any attempt to build in QEMU will fail due to time out). Then, it's only possible to test minimal code due to the nature of the (double)-virtual environment, for example, that means no networking.

Due to may lack of understanding, I just create an executable and copy it to the QEMU chroot, the execute it (in chroot too).

I will research about the boost.build launcher.

aldenml:
let me know if this looks better

.travis.yml
```diff
@@ -167,4 +167,22 @@ script:
...
+      aarch64-linux-gnu-gcc --version;
+      wget -O boost.zip http://pilotfiber.dl.sourceforge.net/project/boost/boost/1.55.0/boost_1_55_0.zip;
+      unzip -qq boost.zip;
+      export BOOST_ROOT=$PWD/boost_1_55_0;
+      sudo apt-get install -y qemu-user-static debootstrap;
+      sudo debootstrap --variant=minbase --arch arm64 --foreign --include=build-essential testing rootfs;
+      sudo cp /usr/bin/qemu-aarch64-static rootfs/usr/bin/;
+      sudo chroot rootfs /debootstrap/debootstrap --second-stage;
+      sudo chroot rootfs mount -t proc none /proc;
+      bjam arm-tests warnings-as-errors=on variant=test_arm toolset=gcc-arm64 target-os=linux link=static testing.launcher="sudo cp -R bin rootfs/; sudo chroot rootfs";
```
arvidn: nice, you actually got this to work

.travis.yml
```diff
@@ -165,3 +167,22 @@ script:
...
+      export PATH=${PWD}/gcc-linaro-5.3-2016.02-x86_64_aarch64-linux-gnu/bin:${PATH};
+      aarch64-linux-gnu-gcc --version;
+      wget -O boost.zip http://pilotfiber.dl.sourceforge.net/project/boost/boost/1.55.0/boost_1_55_0.zip;
+      unzip -qq boost.zip;
+      export BOOST_ROOT=$PWD/boost_1_55_0;
+      sudo apt-get install -y qemu-user-static debootstrap;
+      sudo debootstrap --variant=minbase --arch arm64 --foreign --include=build-essential testing rootfs;
+      sudo cp /usr/bin/qemu-aarch64-static rootfs/usr/bin/;
+      sudo chroot rootfs /debootstrap/debootstrap --second-stage;
+      sudo chroot rootfs mount -t proc none /proc;
```
arvidn: One thing I think would be nice would be to split this up into two parts, and do all the installation and set-up under the "install" travis script. That way all these logs will get collapsed and the test output will only show the build and test output from bjam.

PR #861, opened 2016-06-27
added neon support for popcount

aldenml:


Jamfile
```diff
@@ -376,6 +377,12 @@ rule building ( properties * )
 		}
 	}
 
+	if [ configure.builds /boost/architecture//x86 : $(properties) : x86 ]
+		&& ! <target-os>windows in $(properties)
```
arvidn: target-os doesn't look right here. shouldn't it check toolset gcc? is -mpopcnt a gcc option?

src/bitfield.cpp
```diff
@@ -75,6 +79,30 @@ namespace libtorrent
 		}
 #endif // TORRENT_HAS_SSE
 
+#if TORRENT_HAS_ARM_NEON
+		if (aux::arm_neon_support)
+		{
+			for (int i = 0; i < words; ++i)
+			{
+				uint8x8_t in_val;
```
arvidn: I would prefer declaring the variables on the same line they're initialized. is there a reason not to do that? also, anything that can be "const" should be

arvidn:
nice. I didn't know about the configure target in boost build. does that work when building against an "installed" version of boost? (which travis does iirc) I imagine you need to build against a source distribution for it to work.

Jamfile
aldenml: I will put `! <toolset>msvc` instead, since this is an option for gcc/clang and who knows all the name combinations

src/bitfield.cpp
aldenml: I will rearrange the code, this was to have some resemblance of actual assembly.

Jamfile
arvidn: I think it's probably better to only include toolsets known to support it.. I bet metrowerks, xlc, sunstudio don't support the option. Also, specifying this flag, won't that make GCC/clang use popcnt by itself as well, basically assuming the hardware supports it. It's important that _only_ the explicit calls to it generate this instruction, which is protected by the run-time check of support for it.

aldenml:
well, it's not working in travis :( I suppose you are right and I need to find other solution

Jamfile
aldenml: I thought about it (passing the flag) and was expecting some comment.

I couldn't find any documentation about `gcc` using it implicitly. Actually, even for the built-in call, the instruction is not generated if this flag is not passed. What's worse, the default implementation in assembly of this built-in without the flag, is slower that the pure C++ implementation (this is all anecdotal and from internet).

In any case, I think I can left the `x86` part out of this PR and the best solution is coding it in assembly.

arvidn:
should we use _mm_popcnt instead of the gcc built-in btw?

aldenml:
hmm...didn't notice gcc has that intrinsic too

PR #862, opened 2016-06-27
using popcnt asm for x86 and gcc/clang

aldenml:


arvidn:
I should probably back-port this to RC_1_1 too

aldenml:
Ok

PR #864, opened 2016-06-27
backport of popcnt asm for x86 and gcc/clang

aldenml:


arvidn:
hm.. does this mean anything to you?

`ld: in section __TEXT,__text reloc 0: R_ABS reloc but no absolute symbol at target address file '../bin/b0815893f3fa6985dade81f73181724d/src/web_connection_base.o' for architecture x86_64`

aldenml:
yes, bad cache in travis, no idea what could be the actual reason

PR #632, opened 2016-04-20
Move storage "file-by-file". Closes #435

glassez:
This prevents moving files that do not belongs to torrent.

PR #867, opened 2016-06-28
Move storage "file-by-file" (#632)

glassez:
When moving storage for a torrent, only move files belonging to the torrent, one by one.

### DIFF
--- a/src/storage.cpp
+++ b/src/storage.cpp
@@ -1158,19 +1158,6 @@ namespace libtorrent
 			}
 		}
 
-		// collect all directories in to_move. This is because we
-		// try to move entire directories by default (instead of
-		// files independently).
-		std::map<std::string, int> to_move;
-		for (int i = 0; i < f.num_files(); ++i)
-		{
-			// files moved out to absolute paths are not moved
-			if (is_complete(f.file_path(i))) continue;
-
-			std::string split = split_path(f.file_path(i));
-			to_move.insert(to_move.begin(), std::make_pair(split, i));
-		}
-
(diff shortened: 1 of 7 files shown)


src/storage.cpp
```diff
@@ -792,11 +792,11 @@ namespace libtorrent
 #endif
 			// delete the files from disk
 			std::set<std::string> directories;
-			typedef std::set<std::string>::iterator iter_t;
+			using iter_t = std::set<std::string>::iterator;
```
arvidn: RC_1_1 still supports c++98, so this has to use typedefs still

arvidn:
have you manually tested this patch in qbt as well?

src/storage.cpp
glassez: Ok. Fixed.

glassez:
>have you manually tested this patch in qbt as well?

Yes. It works as expected:
1. Any alien files are not touched when the storage has moved.
2. "Move storage into self" works as well.

arvidn:
compilation error on msvc-10.0

arvidn:
there seems to be a C++11 range-for loop in storage.cpp still too

glassez:
>compilation error on msvc-10.0
there seems to be a C++11 range-for loop in storage.cpp still too

@arvidn fixed.

arvidn:
sorry, but would you mind adding a line to Changelog too? then I'll merge this

glassez:
>sorry, but would you mind adding a line to Changelog too? then I'll merge this

Ok. Done.

PR #869, opened 2016-06-29
added support for arm crc32c and changed to test in arm 32 bits

aldenml:


aldenml:
To explain why the `TORRENT_FORCE_ARM_CRC32` and the move to 32 bits.

QEMU for arm is capable of emulate all `armv8+crc+crypto` instructions in `AArch32` state (the QEMU for arm64 does not), but the report in `getauxval` is inaccurate.

PR #872, opened 2016-06-30
clean-up more defines

arvidn:


metamystical:
This commit causes an error when autotool.sh is run on linux:

    configure.ac:304: error: possibly undefined macro: AC_DEFINE

The line in quesiton is:

    AC_DEFINE([TORRENT_NO_DEPRECATE],[1],[Define to exclude all deprecated functions from the API.])

arvidn:
Thanks! fixed

PR #873, opened 2016-06-30
dont force search of libcrypto in /usr/lib

aldenml:


Jamfile
```diff
@@ -560,7 +560,7 @@ lib libiconv : : <name>iconv <link>shared <search>/usr/local/lib ;
 # openssl on linux/bsd/macos etc.
 lib gcrypt : : <name>gcrypt <link>shared <search>/opt/local/lib <search>/usr/local/lib : : <include>/opt/local/include <include>/usr/local/include ;
 lib z : : <link>shared <name>z <search>/usr/lib ;
-lib crypto : : <name>crypto <link>shared <search>/usr/lib <use>z : : <include>/opt/local/include <include>/usr/local/include ;
```
arvidn: I take it this is to support cross-compilation. shouldn't /usr/local/ and /opt/local be removed too?
and if it makes sense to make this change for libcrypto, I imagine the other libraries probably should have the same treatment, like libz and libgcrypt, libssl etc.

aldenml: Yes, it is to support cross-compilation, specially in macOS, which is the one having problems since openssl is part of the system. Now, I don't know if people expect the current behavior or how they use this. Would you consider merge this and cleanup more in a later step?

PR #875, opened 2016-07-01
removed big-obj flag when using mingw32

aldenml:
if the binutils does not implement this flag, it fails.

### DIFF
--- a/Jamfile
+++ b/Jamfile
@@ -324,13 +324,6 @@ rule building ( properties * )
 		result += <cflags>/bigobj ;
 	}
 
-	if <toolset>gcc in $(properties)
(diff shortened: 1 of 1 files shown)


arvidn:
yeah, but then the object files will be too large building will fail anyway.. at least that was the case when I added this

aldenml:
it compiles fine in my tests, let's see in the mingw of travis, that's very new (with gcc 5.3) btw.

aldenml:
Well, it seems to be fine in appveyor (mingw+gcc), also the flag was in `<cflags>`.

arvidn:
ah. maybe it was the large tommath translation unit that caused the issue (which has been removed since)

PR #882, opened 2016-07-03
support CryptoAPI on windows, for SHA-1 hashing

arvidn:


include/libtorrent/config.hpp
```diff
@@ -235,6 +235,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #ifndef TORRENT_USE_GETIPFORWARDTABLE
 # define TORRENT_USE_GETIPFORWARDTABLE 1
 #endif
+
+#define TORRENT_USE_CRYPTOAPI 1
```
aldenml: maybe this need to be surrounded by `if !defined TORRENT_USE_LIBCRYPTO && !defined TORRENT_USE_LIBGCRYPT` similar to commoncrypto

PR #430, opened 2016-01-23
optimize the optimistic unchoke logic-1.1

arvidn:
and reduce pointer indirection in optimistic unchoke logic

PR #888, opened 2016-07-04
refactor optimistic unchoke extension API

ssiloti:
The new interface is more efficient to implement and only allows extensions to
affect the unchoke order in cases where the last unchoked timestamp is equal.
This is safer because it prevents extensions from overriding the round-robin
ordering of optimistic unchokes.

This is a revival of the discussion in #430 which was deferred until post-1.1.

### DIFF
--- a/include/libtorrent/extensions.hpp
+++ b/include/libtorrent/extensions.hpp
@@ -263,11 +263,13 @@ namespace libtorrent
 		virtual void on_tick() {}
 
 		// called when choosing peers to optimistically unchoke. The return value
-		// indicates the peer's priority for unchoking. Lower return values correspond
-		// to higher priority. If your plugin expects this to be called, make
-		// sure to include the flag ``optimistic_unchoke_feature`` in the return
-		// value from implemented_features(). If multiple plugins implement this function
-		// the lowest return value (i.e. the highest priority) is used.
+		// indicates the peer's priority for unchoking. Lower return values
+		// correspond to higher priority. Priorities above 2^63-1 are reserved.
+		// If your plugin has no priority to assign a peer it should return 2^64-1.
+		// If your plugin expects this to be called, make sure to include the flag
+		// ``optimistic_unchoke_feature`` in the return value from implemented_features().
+		// If multiple plugins implement this function the lowest return value
+		// (i.e. the highest priority) is used.
 		virtual uint64_t get_unchoke_priority(peer_connection_handle /* peer */)
 		{ return std::numeric_limits<uint64_t>::max(); }
 
--- a/include/libtorrent/extensions.hpp
+++ b/include/libtorrent/extensions.hpp
@@ -266,7 +266,8 @@ namespace libtorrent
 		// indicates the peer's priority for unchoking. Lower return values correspond
 		// to higher priority. If your plugin expects this to be called, make
 		// sure to include the flag ``optimistic_unchoke_feature`` in the return
-		// value from implemented_features().
+		// value from implemented_features(). If multiple plugins implement this function
+		// the lowest return value (i.e. the highest priority) is used.
 		virtual uint64_t get_unchoke_priority(peer_connection_handle /* peer */)
 		{ return std::numeric_limits<uint64_t>::max(); }
 
--- a/include/libtorrent/extensions.hpp
+++ b/include/libtorrent/extensions.hpp
@@ -173,6 +173,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/sha1_hash.hpp" // for sha1_hash
 #include "libtorrent/error_code.hpp"
 #include "libtorrent/session_handle.hpp"
+#include "libtorrent/peer_connection_handle.hpp"
 #include "libtorrent/aux_/array_view.hpp"
 
 namespace libtorrent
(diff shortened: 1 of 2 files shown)


src/session_impl.cpp
```diff
@@ -3732,11 +3732,29 @@ namespace aux {
...
+		{
+			if (l.peer->last_optimistically_unchoked
+				!= r.peer->last_optimistically_unchoked)
+			{
+				return l.peer->last_optimistically_unchoked
+					< r.peer->last_optimistically_unchoked;
+			}
+			else
+			{
+				return l.ext_priority < r.ext_priority;
```
arvidn: it's almost as if the ext_priority should be computed lazily

src/session_impl.cpp
```diff
@@ -3790,6 +3807,15 @@ namespace aux {
 				&& t->valid_metadata())
 			{
 				opt_unchoke.push_back(pi);
+#ifndef TORRENT_DISABLE_EXTENSIONS
+				auto const& plugins = m_ses_extensions[plugins_optimistic_unchoke_idx];
+				for (auto& e : plugins)
+				{
+					uint64_t priority = e->get_unchoke_priority(peer_connection_handle(i));
```
arvidn: a nit-pick, priority could be const here

include/libtorrent/extensions.hpp
```diff
@@ -262,15 +262,13 @@ namespace libtorrent
...
-		// to be called, make sure to include the flag
-		// ``optimistic_unchoke_feature`` in the return value from
-		// implemented_features().
-		virtual bool on_optimistic_unchoke(std::vector<peer_connection_handle>& /* peers */)
-		{ return false; }
+		// called when choosing peers to optimistically unchoke. The return value
+		// indicates the peer's priority for unchoking. Lower return values correspond
+		// to higher priority. If your plugin expects this to be called, make
+		// sure to include the flag ``optimistic_unchoke_feature`` in the return
+		// value from implemented_features().
```
arvidn: would it make sense to mention how the priorities from different plugins are combined? (i.e. higher priority/lower value wins)

arvidn:
lgtm

ssiloti:
I'm not sure about the lazy priority computation, it made the code a bit uglier than I would like.

src/session_impl.cpp
```diff
@@ -3732,11 +3732,56 @@ namespace aux {
...
+				return ext_priority;
+			}
+
+			boost::shared_ptr<peer_connection> const* peer;
+
+		private:
+			// these are mutable because comparison functors passed to std::partial_sort
+			// are not supposed to modify the elements they are sorting. Here the mutation
+			// being applied is idempotent so it should not pose a problem.
+			mutable std::vector<boost::shared_ptr<plugin>>* plugins;
```
arvidn: sorry for nit-picking. wouldn't it make more sense for this pointer to live in the comparator function object? Now it's duplicated lots of times. I realize you also use this as an indicator whether the ext_priority has been computed, would it be unreasonable to instead use a special value to indicate this, say UINT64_MAX?

src/session_impl.cpp
```diff
@@ -3732,11 +3732,56 @@ namespace aux {
...
-			, torrent_peer const* const r)
+		struct opt_unchoke_candidate
 		{
-			return l->last_optimistically_unchoked
-				< r->last_optimistically_unchoked;
+			opt_unchoke_candidate(boost::shared_ptr<peer_connection> const* tp
+				, std::vector<boost::shared_ptr<plugin>>* ps)
+				: peer(tp)
+				, plugins(ps)
+				, ext_priority(std::numeric_limits<uint64_t>::max())
```
arvidn: in a C++11 world, I think simple initialization of members should be done at member declarations (it's just more robust in general)

arvidn:
yeah. I don't really have a very strong opinion. The main thing I would be worried about would be the cost when no plugin is involved. which was low in your original patch. The reason I suggested lazy computation is that I would expect the case where the priority is equal to be somewhat rare, and that the majority of priority computation would be wasted

arvidn:
thanks! looks good. I'll just let it run through the CI

metamystical:
This commit introduced an error into client_test when built on a linux system using autotool:

    CXXLD    client_test
    g++: error: @BOOST_RANDOM_LIB@: No such file or directory
    Makefile:509: recipe for target 'client_test' failed
    make[1]: *** [client_test] Error 1
    make[1]: Leaving directory '.../libtorrent-f0d7ec395414d7d2f53639bbb05852778a307264/examples

The error persists through the latest commit.

arvidn:
are you sure it was this commit that introduced it? This commit doesn't make any changes to the build system or discovery of boost.random. Could you double check you're note getting any errors during configure?

metamystical:
It didn't make sense to me either. No errors during configure.

I think I discovered the problem, and it seems to be partly a flaw with github. As of this commit, ax_boost_random.m4 disappears from the m4 directory. Evidently, ax_boost_random.m4 was removed in commit [fef94a4](https://github.com/arvidn/libtorrent/commit/fef94a4c5254e2280b601ab3ea211cc0193fecaa#diff-fd6b6fc9220b72d21683ae8e4f50a210) on May 1, and yet it remained in m4 until this commit. The error should have begun occuring on May1, but it only began occuring on July 24. The cause of the error is that Makefile.am in examples still contains @BOOST_RANDOM_LIB@. This should have been removed on May 1.

arvidn:
I see. thanks!

PR #893, opened 2016-07-08
Modernize

arvidn:


arvidn:
I think I will abandon this in favor of applying single modernizer-checkers at a time. It feels a bit risky given how much the tool got wrong here. The obvious ones like taking arguments by value and moving them, using nullptr and range-for seems like the most obvious transformations.

PR #898, opened 2016-07-09
modernize misc

arvidn:


src/torrent_peer.cpp
```diff
@@ -244,8 +244,7 @@ namespace libtorrent
 #endif
 	}
 
-	ipv4_peer::ipv4_peer(ipv4_peer const& p)
-		: torrent_peer(p), addr(p.addr) {}
+	ipv4_peer::ipv4_peer(ipv4_peer const&) = default;
```
aldenml: hi @arvidn, I'm getting a lint warning here about not initialized base. The parent default constructor exists only under `#if TORRENT_USE_ASSERTS`. Is this correct?

arvidn: this is a copy constructor, I don't understand how ``= default`` could *not* initialize the base and all members. Or am I missing something?

aldenml: I looked at it again, I think my IDE (CLion) lint is getting confused, but I think I will put the `torrent_peer(torrent_peer const&) = default;` outside the `#if TORRENT_USE_ASSERTS` and in the translation unit for "code consistency". Do you agree with that?

arvidn: sure, that's fine. I think I put it in there because it's implied in the no-asserts case

PR #901, opened 2016-07-10
Clang tidy

arvidn:


aldenml:
How strong is this PR? I see that creating the database this way will only check with the arguments in linux, maybe collecting the arguments by hand per platform and with a simple bash script create it do the trick, and then runs the tidy per database.

arvidn:
the intention was just to run it on linux on travis.

arvidn:
I've put it on hold for now. I may get back to it later. I think a better approach is to make a boost-build toolset that runs clang-tidy (instead of compile and link)

aldenml:
feel like a cleanup (close PR) :)

PR #908, opened 2016-07-14
Changes for fixing mutable dht_get_item failure when salt is non-empty

sivachandran:
When get_item reply is received, the item target id was computed based on an uninitialised/unused salt variable. Because of this the request and reply target id didn't match.

Now the uninitialised/unused salt variable is removed and proper the salt variable is referred.

@arvidn

### DIFF
--- a/src/kademlia/get_item.cpp
+++ b/src/kademlia/get_item.cpp
@@ -78,7 +78,8 @@ void get_item::got_data(bdecode_node const& v,
 	// data can reach here, which means pk and sig must be valid.
 	if (!pk || !sig) return;
 
-	std::pair<char const*, int> salt(m_salt.c_str(), int(m_salt.size()));
+    std::string temp_copy(m_data.salt());
+	std::pair<char const*, int> salt(temp_copy.c_str(), int(temp_copy.size()));
 	sha1_hash incoming_target = item_target_id(salt, pk);
 	if (incoming_target != m_target) return;
 
--- a/include/libtorrent/kademlia/get_item.hpp
+++ b/include/libtorrent/kademlia/get_item.hpp
@@ -72,7 +72,6 @@ protected:
 
 	data_callback m_data_callback;
 	item m_data;
-	std::string m_salt;
 	bool m_immutable;
 };
 


src/kademlia/get_item.cpp
```diff
@@ -78,7 +78,7 @@ void get_item::got_data(bdecode_node const& v,
 	// data can reach here, which means pk and sig must be valid.
 	if (!pk || !sig) return;
 
-	std::pair<char const*, int> salt(m_salt.c_str(), int(m_salt.size()));
```
ssiloti: I'd like to see a copy of the salt string made here. As is, the salt string gets assigned to itself down in `m_data.assign`. While this probably wouldn't cause a problem in practice, I think it technically invokes Undefined Behavior and the aliasing here makes me nervous.

sivachandran: Now a temporary copy of the salt is taken for computing the item target id and same is used for assigning too. Thanks for highlighting aliasing.

arvidn:
I haven't looked through the patch yet, but I imagine this would apply to RC_1_1 as well. in which case I would prefer the pull request against that branch. I'll merge it into master later

sivachandran:
Replacing with PR #909 that targets RC-1-1 branch.

arvidn:
Thanks for reporting and looking at fixing this. I think I understand more of what's going on.

The original intent of this code was probably to assign m_salt in the constructor, that way it have been used correctly and not self-assigned when setting the m_data member.

However, since storing the salt twice is redundant, and that the pubic key is already stored in the m_data field and then complemented with the additional payload, maybe that's a reasonable model to extend (like you have done).

The self-assignment of salt in the m_data object bothers me a bit, but I'm not sure it's worth the extra complexity of adding another overload of assign() to avoid it.

Also, do you think you would be able to add a test that exercises this logic? You can probably be inspired by some of the existing tests. Really just sending a get message with salt and making sure we get the expected response. You could look at either test/test_dht.cpp for a unit-test or simulation/test_dht.cpp for a end-to-end system test (as a simulation).

PR #909, opened 2016-07-16
Changes for fixing mutable dht_get_item failure when salt is non-empty

sivachandran:
When get_item reply is received, the item target id was computed based on an uninitialised/unused salt variable. Because of this the request and reply target id didn't match.

Now the uninitialised/unused salt variable is removed and proper the salt variable is referred.

@arvidn


arvidn:
The dht test is in need of some refactoring. I will do some refactoring and add a test in master

arvidn:
this has been merged into master now, and a test is being added in https://github.com/arvidn/libtorrent/pull/926

PR #912, opened 2016-07-16
session.get_ip_filter() first call crash fix

XRevan86:
Fix the crash that happens when libtorrent::session.get_ip_filter() is called before any filter is set.
This is a regression caused by https://github.com/arvidn/libtorrent/commit/2bf4519.
Affected applications: Deluge.

Examples to reproduce the issue:
```c++
#include <libtorrent/session.hpp>
#include <libtorrent/ip_filter.hpp>

int main() {
    libtorrent::session session;
    session.get_ip_filter();
    return 0;
}
```
```python
python -c 'import libtorrent; libtorrent.session().get_ip_filter()'
```

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -1239,6 +1239,7 @@ namespace aux {
 	ip_filter const& session_impl::get_ip_filter()
 	{
 		TORRENT_ASSERT(is_single_thread());
+		if (!m_ip_filter) m_ip_filter = boost::make_shared<ip_filter>();
 		return *m_ip_filter;
 	}
 


arvidn:
thanks! would you mind adding a test that would have caused the crash too? (it sounds like you just need to construct a session and call get_ip_filter())

XRevan86:
@arvidn, I kind of can't figure where in tests to put that :-).
> it sounds like you just need to construct a session and call get_ip_filter()

Yeah.

arvidn:
you could put it in test/test_ip_filter.cpp, something like this perhaps:

TORRENT_TEST(session_get_ip_filter)
{
   session ses;
   ip_filter const& ipf = ses.get_ip_filter();
#if TORRENT_USE_IPV6
   TEST_CHECK(std::get<0>(ipf.export_filter()).empty());
#else
   TEST_CHECK(ipf.export_filter().empty());
#endif
}

arvidn:
added a test to this patch in https://github.com/arvidn/libtorrent/pull/914

XRevan86:
I was adding it right now but, well, it's settled now, right?

arvidn:
yeah, I'll just let it pass the tests and then merge it. thanks again!

PR #919, opened 2016-07-18
fix range of lsd::cookie

k1988:
in lsd::on_announce:
```
...
std::int32_t cookie = strtol(cookie_iter->second.c_str(), nullptr, 16);
if (cookie == m_cookie)
...
```

if m_cookie is big than 0x7fffffff, it will cause an overflow, the cookie of local will not equal with m_cookie.
then 127.0.0.1 and other local ip will be accepted (It's shoud be "ignoring packet (cookie matched our own)")。


arvidn:
please make this PR against the RC_1_1 branch instead. I will merge it into master later

PR #924, opened 2016-07-20
version to 1.2

aldenml:


arvidn:
please use the python tool set_version.py to update all places where the version is specified (like in the documentation)

aldenml:
ok, let me see it

aldenml:
@arvidn, in appveyor `The process cannot access the file because it is being used by another process.` is a new one. Any hint?

arvidn:
I've seen that in the past too, I don't think your patch affected it. It's on my list to take a look at.

PR #929, opened 2016-07-21
modernize and improve type-safety of the DHT put implementation and test

arvidn:


src/kademlia/dht_storage.cpp
```diff
@@ -206,7 +207,7 @@ namespace
 		size_t num_peers() const override
 		{
 			int ret = 0;
-			std::for_each(m_map.begin(), m_map.end(), count_peers(&ret));
+			std::for_each(m_map.begin(), m_map.end(), count_peers(ret));
```
ssiloti: You could get rid of count_peers entirely and use a lambda here.

PR #931, opened 2016-07-22
modernize hasher to use array_view

arvidn:
as well as adding array_view support to entry


aldenml:
What do you think about moving `array_view` to a public place (out of `aux`)?

arvidn:
yeah. I will make it public and probably rename it to span. perhaps at some point I will switch over to gsl::span

arvidn:
it turned out to be quite useful, including for public APIs, like the dht_storage interface

PR #932, opened 2016-07-22
fix value of current_tracker when all tracker failed

k1988:
In my project, I want to check if I have successfully connected to tracker。
So I used torrent_status::current_tracker,which is defined as
```
// the URL of the last working tracker. If no tracker request has
// been successful yet, it's set to an empty string.
std::string current_tracker;
```
then,I stop my tracker server,and run my app, at first current_tracker was empty, but soon it's value turned to be my first tracker.
But the correct value should be empty? unless it's definition as a mistake.


k1988:
Oh, I'm sorry, I forgot my commit log is Chinese.
Is there a simple way to change this log?

arvidn:
I can change it when I squash the commits

arvidn:
would you mind adding a test in simulation/test_tracker.cpp that fails without this change?

k1988:
ok，i will do it

arvidn:
thanks!

PR #934, opened 2016-07-23
using span in crypto_plugin encrypt/decrypt

aldenml:
The context for this PR is that eventually everything will move to span.


arvidn:
lgtm. @ssiloti do you have any concerns?

ssiloti:
lgtm

PR #936, opened 2016-07-23
transition part of http_parser to use span

arvidn:


src/http_connection.cpp
```diff
@@ -743,10 +743,9 @@ void http_connection::on_read(error_code const& e
 
 	if (m_bottled || !m_parser.header_finished())
 	{
-		libtorrent::buffer::const_interval rcv_buf(&m_recvbuffer[0]
-			, &m_recvbuffer[0] + m_read_pos);
+		span<char const> rcv_buf(m_recvbuffer);
 		bool error = false;
-		m_parser.incoming(rcv_buf, error);
+		m_parser.incoming(rcv_buf.subspan(0, m_read_pos), error);
```
aldenml: I wonder if `rcv_buf` could be eliminated

src/http_parser.cpp
```diff
@@ -364,7 +348,8 @@ namespace libtorrent
 						std::get<0>(ret) += int(payload);
 						incoming -= int(payload);
 					}
-					buffer::const_interval buf(recv_buffer.begin + m_cur_chunk_end, recv_buffer.end);
+					span<char const> buf(recv_buffer.data() + m_cur_chunk_end
```
aldenml: can `span:: last` be used here?

src/http_connection.cpp
arvidn: if there was a free function version of subspan() m_recvbuf could be passed in directly, but there isn't one right now

src/http_parser.cpp
arvidn: subspan() can be used

PR #937, opened 2016-07-23
update required boost version and ax_boost_base.m4

arvidn:


arvidn:
@cas-- does this seem reasonable?

arvidn:
tested to work on osx with homebrew

arvidn:
ping @cas--

cas--:
It looks to me as though can actually drop the check for atomic. I removed the atomic lib packages and libtorrent build fine with the just header file available.

I am testing again with the debian build but I expect the same outcome.

arvidn:
thanks for looking into this

PR #939, opened 2016-07-24
attempt detection of heap corruption on windows in tests

arvidn:


arvidn:
This should be back-ported to RC_1_1 as well

PR #942, opened 2016-07-24
update our id in the rpc manager when it changes

ssiloti:


arvidn:
oops, good catch! do you think it's easy to write a test for this? send a message, catch the outgoing node ID in the message, trigger a recalculation of node ID and send another message and catch it and make sure it's different.

This fix seems reasonable. I looked at where m_our_id is used and it's only in add_our_id(), which is called from two places. Once from node (which has the node ID already) and the other time from observers, which don't have our node ID normally. So I think it makes sense to have the rpc manager know the node id. An alternate fix would be to make its member a const reference to the parent's (node's) node ID.

This fix should also be mentioned in ChangeLog I think

ssiloti:
I'll add a test in a separate PR to master since the DHT tests got massively refactored since 1.1.

arvidn:
good point

arvidn:
it's been merged into master

PR #948, opened 2016-07-25
not using template/enable_if in public type entry

aldenml:
Hi @arvidn, I know this is a very particular request/PR. Swig go crazy (and fail) with these constructions and I think having them (string/char*) separated is still manageable.

### DIFF
--- a/include/libtorrent/entry.hpp
+++ b/include/libtorrent/entry.hpp
@@ -117,18 +117,8 @@ namespace libtorrent
 		// newly constructed entry
 		entry(dictionary_type);
 		entry(span<char const>);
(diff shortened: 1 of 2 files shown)


aldenml:
a lot of ambiguous errors, need to solve that

aldenml:
closing this due to lack of clear solution

arvidn:
yeah, the problem (and the reason I added those enable_ifs) is that implicit conversions to std::string have the same weight as the ones to span<char const>, which makes a lot of calls fail overload resultion. Specifically a string literal. It's more efficient to choose the span<> overload for those. The main reason to add the span<> overload was to support sha1_hash and ed25519 keys

PR #949, opened 2016-07-25
mark advanced used of template with __cplusplus

aldenml:
@arvidn, swig is having a very hard time with the template/typename syntax, but just surrounding them with the `__cplusplus` allows swig to ignore them in the parsing process. Do you think this PR is acceptable?

### DIFF
--- a/include/libtorrent/entry.hpp
+++ b/include/libtorrent/entry.hpp
@@ -117,6 +117,7 @@ namespace libtorrent
 		// newly constructed entry
 		entry(dictionary_type);
 		entry(span<char const>);
+#if __cplusplus >= 201103L
 		template <typename U, typename = typename std::enable_if<
 			std::is_same<U, entry::string_type>::value
 			|| std::is_same<U, char const*>::value >::type>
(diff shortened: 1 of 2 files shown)


arvidn:
well, this particular patch causes build failures. does it work if you give those template parameters names?

aldenml:
I will see how the naming works. Do you think the others `#if __cplusplus >= 201103L` already there could be a problem? Maybe there is an assumption that's actually not happening with msvc

arvidn:
I've been removing those from master. are there any left?
I don't know why msvc-2015 wouldn't evaluate that to true, maybe because they still don't support type sfinae yet

aldenml:
Yes, like a dozen.

aldenml:
The naming seems to work in `entry` (at least is parsed) but the `span` does not, it seems like `decltype` is a problem. Would you consider this:

- Add a conversion operator/function to `sha1_hash` to `span`
- Add a constructor to `span` to handle `vector`.
- Remove the problematic constructor in `span`.

Also, I see that you test for `data()` but not for `size()`, any particular reason?

arvidn:
as for just checking for data(), that's inspired by: https://github.com/martinmoene/gsl-lite/blob/master/include/gsl/gsl-lite.h

It's not just vector and sha1_hash that are spans. Also std::array, libtorrent::buffer, std::string.

Maybe by changing the decltype to a std::enable_if<> it would work in swig

aldenml:
Thanks for your comments and hints. I know it was a desperate request :)

I will close this and start again

PR #951, opened 2016-07-26
__cplusplus cleanup

aldenml:


include/libtorrent/span.hpp
```diff
@@ -60,10 +60,12 @@ namespace libtorrent
 		span(U (&arr)[N])
 			: m_ptr(&arr[0]), m_len(N) {}
 
+#if (defined __GNUC__ || (defined _MSC_VER && _MSC_VER >= 1900))
```
arvidn: not sure about this. libtorrent would fail to build unless that condition is true, right?

simulation/test_trackers_extension.cpp
```diff
@@ -1,162 +0,0 @@
-/*
```
arvidn: huh, I thought I had deleted this already

include/libtorrent/span.hpp
```diff
@@ -60,10 +60,12 @@ namespace libtorrent
...
 		// anything with a .data() member function is considered a container
 		template <typename Cont, typename = decltype(std::declval<Cont>().data())>
```
arvidn: could you try to give this template parameter a name too, instead?

aldenml: I tried with no success, I even tried to recreate it with another construction, the problem is that swig fail to parse if `decltype` has any non trivial expression inside

include/libtorrent/span.hpp
aldenml: Yes, it will fail, but in general, it will fail anyway if msvc < 14 and gcc/clang is not modern enough

aldenml:
I know is a hard one to accept the `#if (defined __GNUC__ || (defined _MSC_VER && _MSC_VER >= 1900))`, but even `gsl-lite` surround that constructor with the macro `gsl_HAVE_CONSTRAINED_SPAN_CONTAINER_CTOR`. Well, waiting for your comment :)

arvidn:
gsl-lite is meant to compile on pre C++11 compilers, so some features are conditionally included. I don't think that's really relevant here. If whatever condition you put around that constructor is false, libtorrent won't compile. So you might as well add that test to config.hpp and put #error unsupported compiler. I don't think that's a good idea, to encode support for specific compiler vendors like that. it would be a lot more appropriate to test for a feature-macro.

Did you try to change the condition to use std::enable_if? something like
`std::enable_if<std::is_same<decltype(std::declval<Cont>().data()), T*>::type>`

aldenml:
I underestand and yes, I tried similar `enable_if` and with other combinations too, unfortunately, is the `decltype` inner expression the problem. Whatever is there, makes swig to crash during the parsing.

I will remove the condition in span, to get this PR accepted if the rest is ok and continue in another PR

aldenml:
Thank you! I just found that in fact I don't need to include `span` in the parsing

arvidn:
great!

PR #956, opened 2016-07-27
removed const_interval, mutable_buffer and const_buffer types

aldenml:
I expect some comments related to the few needed extra calls to `subspan` due to the way `detail::read_*` works.

### DIFF
--- a/src/bt_peer_connection.cpp
+++ b/src/bt_peer_connection.cpp
@@ -62,6 +62,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/alert_types.hpp"
 #include "libtorrent/invariant_check.hpp"
 #include "libtorrent/io.hpp"
+#include "libtorrent/aux_/io.hpp"
 #include "libtorrent/socket_io.hpp"
 #include "libtorrent/version.hpp"
 #include "libtorrent/extensions.hpp"
@@ -1640,9 +1641,7 @@ namespace libtorrent
 		TORRENT_ASSERT(recv_buffer.front() == msg_extended);
 		recv_buffer = recv_buffer.subspan(1);
 
-		char const* pos = recv_buffer.begin();
-		int extended_id = detail::read_uint8(pos);
-		recv_buffer = recv_buffer.subspan(1);
+		int extended_id = aux::read_uint8(recv_buffer);
 
 		if (extended_id == 0)
 		{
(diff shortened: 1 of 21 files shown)


src/bt_peer_connection.cpp
```diff
@@ -1660,7 +1662,9 @@ namespace libtorrent
 #endif
 				return;
 			}
-			bool ul = detail::read_uint8(recv_buffer.begin) != 0;
+			char const* pos1 = recv_buffer.begin();
```
arvidn: yeah, these are kind of the messy uses of const_interval. I added a separate set of read_*() and write_*() functions that advances the start position of the span<>. I forget which namespace their in off hand

aldenml: libtorrent/aux_/io.hpp?

src/receive_buffer.cpp
```diff
@@ -317,15 +316,15 @@ int crypto_receive_buffer::advance_pos(int bytes)
...
 }
 
-buffer::const_interval crypto_receive_buffer::get() const
+span<char const> crypto_receive_buffer::get() const
 {
-	buffer::const_interval recv_buffer = m_connection_buffer.get();
+	span<char const> recv_buffer = m_connection_buffer.get();
 	if (m_recv_pos < m_connection_buffer.pos())
-		recv_buffer.end = recv_buffer.begin + m_recv_pos;
+		recv_buffer = recv_buffer.subspan(0, m_recv_pos);
```
arvidn: this is the same as recv_buffer.first(m_recv_pos), right?

aldenml: Unless I'm reading wrong, I don't think it's the same. `first` seems to discard the last `n` elements, could you confirm?

aldenml:
any suggestion of how to handle the error related to `-Wshift-count-overflow` in `../include/libtorrent/aux_/io.hpp:55:`

arvidn:
hm.. try this: http://pastebin.com/QmNht2k3

src/receive_buffer.cpp
arvidn: oh, that's not the intention... it's supposed to *pick* the first *n* items

aldenml: I rather change it (and look for `subspan(0`) in another PR, is that ok?

arvidn: yeah, of course

Issue #766, opened 2016-05-28
Fails to build with python-dbg on linux


sledgehammer999:
I am trying to make a deb package for 1.0.9 to use on qbittorrent's ppa but I have found out it errors out. I narrowed it down to this command:

```
./configure PYTHON=/usr/bin/python2.7-dbg --enable-python-binding --with-boost-python=boost_python-py27
```

It fails with:

> checking for the distutils Python package... no
> configure: error: cannot import Python module "distutils"

However, I have been able to build 1.0.7. (1.0.8 fails too).

Any ideas on what to do?
PS: You need to install package `python-dbg` to have access to python27-dbg.
PS2: Using the non-dbg python binary it works. It doesn't complain about distutils.

arvidn:
what's the error?

arvidn:
oh, the distutils..

sledgehammer999:
I found out which file causes this. It is `m4/ax_python_devel.m4`. I replaced it with the one found in 1.0.7 and run `autotool.sh`. Then the command in my 1st post worked.

arvidn:
that m4 file was updated here: https://github.com/arvidn/libtorrent/commit/2092e30852253082512974d941f74e8113efd390

which was a back-port of this PR: https://github.com/arvidn/libtorrent/pull/324
perhaps @voyageur can shed some light on this.

Is the new m4 script not compatible with older versions of python?
Is it not compatible with the debug build of python?

voyageur:
Hmm, it works for me with python 2.7 (and here for the release build apparently). So something does not go well with the python-dbg setup… Does this command work?
`python2.7-dbg -c "import distutils"`

There may be some additional env variables to set for these debug builds, that "happened to work" with the older ax_python_devel.m4?

sledgehammer999:
`python2.7-dbg -c "import distutils"` gives output `[21667 refs]` while the 3.4 and 3.5 variants don't give something. Is this normal? (I am a python noob).

PS: I use debian unstable for the above.

voyageur:
So this is probably caused by [this change](https://github.com/arvidn/libtorrent/commit/2092e30852253082512974d941f74e8113efd390#diff-808824a21167bd07857fe4cc03527a84L136). As there is some output from the command, the configure script considers there was an error.

If 3.x versions do not have this , this verbose output has robably been fixed then in Debian, but not for 2.7...

I think reverting this line should be enough to fix your case and still support new library suffix.

arvidn:
[...]

PR #958, opened 2016-07-27
[#766] Fix building bindings with python-dbg

cas--:
* In ax_python_devel grep and remove '[xxxx refs]' if found in python-dbg output so that
   it will still catch any errors redirected to stdout.
 * In ax_boost_python change the specific _BOOST_PYTHON_MODULE test for the old version that
   used a generic _main method test.


m4/ax_boost_python.m4
```diff
@@ -101,12 +101,11 @@ if test "$ac_cv_boost_python" = "yes"; then
    fi])
   BOOSTLIBDIR=`echo $BOOST_LDFLAGS | sed -e 's/@<:@^\/@:>@*//'`
   for ax_lib in $ax_python_lib $ax_boost_python_lib `ls $BOOSTLIBDIR/libboost_python*.so* $BOOSTLIBDIR/libboost_python*.dylib* $BOOSTLIBDIR/libboost_python*.a* 2>/dev/null | sed 's,.*/,,' | sed -e 's;^lib\(boost_python.*\)\.so.*$;\1;' -e 's;^lib\(boost_python.*\)\.dylib.*$;\1;' -e 's;^lib\(boost_python.*\)\.a.*$;\1;' ` boost_python boost_python3; do
-    AS_VAR_PUSHDEF([ax_Lib], [ax_cv_lib_$ax_lib''_BOOST_PYTHON_MODULE])dnl
+    AS_VAR_PUSHDEF([ax_Lib], [ax_cv_lib_$ax_lib''_main])dnl
```
arvidn: is this a python-version specific name? or does it work with both python 2 and 3?

arvidn:
I've been thinking about maybe adding a travis build configuration to run through configure && make check..

m4/ax_boost_python.m4
cas--: It generic from previous version and was tested with both 2.7 and 3.2

PR #959, opened 2016-07-27
Fix reoccurrence of py bindings -Wstrict-prototype warning

cas--:
* Recent commit df368c2d158 has allowed the -Wstrict-prototype flag to
once again be passed to gcc so this fix removes all occurrences not just from 'OPT'.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 from distutils.core import setup, Extension
-from distutils.sysconfig import get_config_var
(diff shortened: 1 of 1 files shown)


arvidn:
so, does python just expect that it's OK to pass that argument in to the compiler? it seems kind of weird..

cas--:
Seems an unintended bug using `gcc`. From python 2.7 configure:
```
case $GCC in
yes)
    if test "$CC" != 'g++' ; then
        STRICT_PROTO="-Wstrict-prototypes"
    fi
```

PR #961, opened 2016-07-28
remove DHT nodes which change their id

ssiloti:


src/kademlia/routing_table.cpp
```diff
@@ -682,13 +701,37 @@ routing_table::add_node_status_t routing_table::add_node_impl(node_entry e)
...
 			TORRENT_ASSERT(existing->id != e.id);
-			// this is the same IP and port, but with
-			// a new node ID. remove the old entry and
-			// replace it with this new ID
+			// This is the same IP and port, but with a new node ID.
+			// This may indicate a malicious node so remove the entry.
+#ifndef TORRENT_DISABLE_LOGGING
+			char hex_id_new[41];
+			char hex_id_old[41];
+			aux::to_hex(reinterpret_cast<char const*>(&e.id[0]), 20, hex_id_new);
```
arvidn: if you call e.id.data(), you don't need the reinterpret cast. hopefully at some point I'll make all different parts agree on what the type of a "byte" is.

src/kademlia/routing_table.cpp
```diff
@@ -682,13 +701,37 @@ routing_table::add_node_status_t routing_table::add_node_impl(node_entry e)
...
+			// schedule the other nodes in the bucket to be pinged soon
+			// to clean out any other malicious nodes
+			auto now = aux::time_now();
+			for (auto& node : existing_bucket->live_nodes)
+			{
+				if (now - node.last_queried > minutes(5))
+					node.last_queried = min_time();
+			}
+
+			return failed_to_add;
```
arvidn: is it possible to trick libtorrent (with this patch) to remove a bunch of nodes by sending spoofed packets, with their source (IP,port)?

ssiloti: This code can only be reached if e.pinged() is true so spoofed packets should never get here.

arvidn:
awesome, thanks!

arvidn:
Do you think there's a risk this was introduced by this patch? https://ci.appveyor.com/project/arvidn/libtorrent/build/3300/job/a8la96u3hnbwcgt0#L2857

ssiloti:
It sure looks like it. The stack trace is pointing at the log line I added, but I don't see anything obviously wrong. I'll spend more time tomorrow looking at it.

src/kademlia/routing_table.cpp
```diff
@@ -682,13 +701,37 @@ routing_table::add_node_status_t routing_table::add_node_impl(node_entry e)
...
-			// a new node ID. remove the old entry and
-			// replace it with this new ID
+			// This is the same IP and port, but with a new node ID.
+			// This may indicate a malicious node so remove the entry.
+#ifndef TORRENT_DISABLE_LOGGING
+			char hex_id_new[41];
+			char hex_id_old[41];
+			aux::to_hex(e.id.data(), 20, hex_id_new);
+			aux::to_hex(existing->id.data(), 20, hex_id_old);
+			m_log->log(dht_logger::routing_table, "evicting node (changed ID): old: %s new: %s %s"
```
arvidn: actually, I seem to recall that m_log can be NULL, especially in some tests. I realize that's only checked one out of three times in routing_table right now, but perhaps those other cases aren't covered by those tests.

arvidn: in node.cpp m_observer (which is the same object) is checked every time before use, afaict. That's probably the issue

ssiloti: If m_log being NULL was the problem I would have expected it to fail consistently on all platforms, but I can't reproduce it. In any case I've pushed a fix to always check m_log.

PR #963, opened 2016-07-28
mark encryption_handler with TORRENT_EXTRA_EXPORT and minor cleanup

aldenml:


include/libtorrent/pe_crypto.hpp
```diff
@@ -97,7 +97,7 @@ namespace libtorrent
 		sha1_hash m_xor_mask;
 	};
 
-	struct encryption_handler
+	struct TORRENT_EXTRA_EXPORT encryption_handler
```
arvidn: this is the internal encryption_handler, right? it still allows clients to define their own.

aldenml: yes, this is very internal

PR #964, opened 2016-07-28
session constructor with plugins and better api for dht storage

aldenml:
This is for review (before adding documentation and tests). The main reasons for this change are:

1- There is no opportunity to inject plugins very early in the session construction. This put custom plugins in the same level as the default ones.

2- Now that plugins are economical to add, I think customizing the dht storage via plugins is a lot better (API wise). It's a trade off here, I rather go for better API (assuming the point 1 is accepted).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+<break>
 language: cpp
 
 matrix:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,3 +1,4 @@
+<break>
 version: "{build}"
 branches:
   only:
--- a/include/libtorrent/session.hpp
+++ b/include/libtorrent/session.hpp
@@ -131,6 +131,41 @@ namespace libtorrent
 		boost::shared_ptr<aux::session_impl> m_impl;
 	};
 
+	struct TORRENT_EXPORT session_params
+	{
+		session_params()
+			: flags(default_flags)
+			, dht_storage_constructor(dht::dht_default_storage_constructor)
+		{}
+
+		session_params(int start_flags)
+			: flags(start_flags)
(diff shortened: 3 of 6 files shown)


arvidn:
I agree that the way extensions and the dht_storage are added right now is not ideal. I'm a bit hesitant for this specific fix though. In general I think it's better/simpler/more intuitive to pass down the extensions to use at construction time. But I think that really applies to the dht storage too. With this patch there's an inversion of control in a way, where, in order to customize the dht storage, you need to implement the extension interface, which will have a function called on it where you can construct the storage. And the question of which extension wins is not that great either (is it the first or the last? neither answer is great).

Do you think it would make sense for the session_settings itself to be able to hold an (optional) dht_storage object? and maybe even a vector of session plugins. I suppose putting that in session_settings could get weird when you call apply_settings(). Presumably you wouldn't want to clear and replace the existing ones every time. But on the other hand, I imagine it would be nice to have a way of removing and re-adding extensions. removing the dht storage wouldn't seem to make sense though.

Maybe it would make sense to have a separate object that holds initialization state for a session, such as extensions and the dht_storage.

One major benefit of creating the dht_storage and session plugins outside of the session itself is that you can capture custom state trivially from the outside.

aldenml:
I will go for 

>  it would make sense to have a separate object that holds initialization state for a session

let me figure out the details

aldenml:
@arvidn I created a `session_params` that potentially can hold a lot of values, but I wasn't sure about putting there the `settings_pack` because it's passed as a reference in `session`. Waiting for your comments.

aldenml:
gentle ping here, not in a rush

aldenml:
Closing to create another

PR #967, opened 2016-07-29
add ppc64le as 64-bit system libraries

michelmno:
required to avoid configure error like:
===
configure: error: Boost.System library not found. Try using --with-boost-system=lib
===

### DIFF
--- a/m4/ax_boost_base.m4
+++ b/m4/ax_boost_base.m4
@@ -92,7 +92,7 @@ if test "x$want_boost" = "xyes"; then
     libsubdirs="lib"
     ax_arch=`uname -m`
     case $ax_arch in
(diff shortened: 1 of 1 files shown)


michelmno:
the  continuous-integration/travis-ci/pr reported failure is not related to my proposed patch (that is ppc64le  specific), so should not block the pull request.

arvidn:
yeah, it's still on my list to make that test pass reliably.

arvidn:
would you mind issuing this PR against the RC_1_1 branch? that's the current stable release branch, and it will make this get released sooner.

michelmno:
it is already covered by commit 633d0621f8 in RC_1_1 branch, so expecting to have same in master.

PR #969, opened 2016-07-30
restore timeout in travis to 900

aldenml:
Don't know the reason for the lower value, but restoring it seems to improve the timeout in the test `test_web_seed_socks5_pw`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -119,7 +119,7 @@ script:
 # libtorrent is the name of the test suite target
   - cd test
   - 'if [ "$variant" != "" ]; then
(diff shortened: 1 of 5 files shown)


aldenml:
Do you think another timeout increase in `test_utp` is needed or could be another fundamental problem?

arvidn:
the only test I've seen hit this timeout is the ssl test, which I'm pretty sure is because it's flakey.
The main reason was to make flakey tests (that hand) fail earlier. Travis has a 10 minute timeout, so having a 15 minute timeout (900 seconds) in the tests is kind of pointless, it will just prevent the tests coming after to be run at all

aldenml:
I knew it couldn't be that easy, but now i see other web seed tests timeout too

PR #970, opened 2016-07-30
don't define _WIN32_WINNT if not under windows

aldenml:


arvidn:
does this actually make a difference, or is it just offensive to see this define? :) (I'm not opposed to this change)

aldenml:
sure very offensive it is :D

It does not make any difference in `libtorrent`, but if you create a bjam project and link to `libtorrent` as source code, it's injected and I think that's not good.

PR #974, opened 2016-08-01
use of BOOST_FALLTHROUGH, typos and minor cleanup

aldenml:


src/utp_stream.cpp
```diff
@@ -3147,7 +3147,7 @@ bool utp_socket_impl::incoming_packet(span<std::uint8_t const> buf
 				m_connect_handler = false;
 				utp_stream::on_connect(m_userdata, m_error, false);
 			}
-			// fall through
+			BOOST_FALLTHROUGH;
```
arvidn: do you know which boost version this macro was first introduced in?

aldenml: 1.54, can't remember what the minimum for master

arvidn: It's currently 1.53 (in configure.ac), but there's no reason it couldn't be 1.54

aldenml: I will change it then

aldenml: Is it `AX_BOOST_BASE([1.47])` in configure.ac? (can't find the string `1.53`)

aldenml:
OK, I saw it in `RC_1_1` it is `1.53`, I assume that change will come to `master` in the next merge

src/utp_stream.cpp
arvidn: ah, it's in RC_1_1, but I haven't merged that into master for a while. feel free to change it i master and I'll merge later

PR #978, opened 2016-08-02
some trivial cleanup

aldenml:


include/libtorrent/alert.hpp
```diff
@@ -250,11 +247,6 @@ namespace libtorrent {
 
 #ifndef TORRENT_NO_DEPRECATE
 
-#ifdef __GNUC__
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
arvidn: this was to cover the use of std::auto_ptr. are they all gone?

aldenml: at least in these files, yes

include/libtorrent/alert_manager.hpp
```diff
@@ -53,12 +50,6 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "libtorrent/aux_/disable_warnings_pop.hpp"
 
-#ifdef __GNUC__
-// this is to suppress the warnings for using std::auto_ptr
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
arvidn: I take it all mentions of std::auto_ptr<> are gone from this file too then?

aldenml: correct

aldenml:
I have two questions:

1 - do you want to move to `std::function`? (with the missing `clear()` method)
2 - do you want to move to `std::shared_ptr`?

PR #980, opened 2016-08-03
fixed warnings and compilation issues

aldenml:


src/utp_stream.cpp
```diff
@@ -58,8 +58,8 @@ namespace libtorrent {
 
 #if TORRENT_UTP_LOG
 
-char const* packet_type_names[] = { "ST_DATA", "ST_FIN", "ST_STATE", "ST_RESET", "ST_SYN" };
-char const* socket_state_names[] = { "NONE", "SYN_SENT", "CONNECTED", "FIN_SENT", "ERROR", "DELETE" };
+static char const* packet_type_names[] = { "ST_DATA", "ST_FIN", "ST_STATE", "ST_RESET", "ST_SYN" };
```
arvidn: generally I would prefer the anonymous namespace, but I realize it's not used pervasively right now

PR #981, opened 2016-08-03
make the random function produce proper random distributions, based o…

arvidn:
…n uniform_int_distribution of C++11. also clean up piece_picker's priority_range to return two results instead of taking pointers to out-parameters

### DIFF
--- a/include/libtorrent/bt_peer_connection.hpp
+++ b/include/libtorrent/bt_peer_connection.hpp
@@ -324,14 +324,6 @@ private:
 			read_packet
 		};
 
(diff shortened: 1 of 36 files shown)


aldenml:
I can imagine the struggle with the name `one_past_end` :)

I want to comment that I see in at least eight times that you replaced `random() % num` by `random(num)`. Is that OK?

arvidn:
I did that when the exact range of the random number was not important (or at least not as far as I could tell). Things like generating random port numbers.

aldenml:
Thanks, I didn't want to bother you, just wanted to make sure.

Have you considered removing `randint`?

arvidn:
yes, but I would have to upgrade all the uses of random_shuffle() too, so I was putting that off for this patch

PR #982, opened 2016-08-04
make settings_pack moveable

arvidn:


aldenml:
Question, do you think that with this change, I can put the settings as a field in a potential `session_params`?

arvidn:
that PR was what inspired this change. as long as the session_params itself is movable, that should be fine (and in fact beneficial)

PR #985, opened 2016-08-05
refactor for separation of clz software/hardware based

aldenml:
The idea with this is able to test individually the software based implementation. I have plans to add `ctz` later.

### DIFF
--- a/include/libtorrent/aux_/ffs.hpp
+++ b/include/libtorrent/aux_/ffs.hpp
@@ -41,6 +41,9 @@ namespace libtorrent { namespace aux
 {
 	TORRENT_EXTRA_EXPORT int clz_sw(span<std::uint32_t const> buf);
 	TORRENT_EXTRA_EXPORT int clz_hw(span<std::uint32_t const> buf);
+
+	// this function statically determines if hardware of software is used
+	TORRENT_EXTRA_EXPORT int clz(span<std::uint32_t const> buf);
 }}
 
 #endif // TORRENT_FFS_HPP_INCLUDE
--- a/include/libtorrent/aux_/ffs.hpp
+++ b/include/libtorrent/aux_/ffs.hpp
@@ -1,6 +1,6 @@
 /*
 
-Copyright (c) 2014, Arvid Norberg
+Copyright (c) 2014-2016, Arvid Norberg, Alden Torres
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
@@ -30,18 +30,17 @@ POSSIBILITY OF SUCH DAMAGE.
 
 */
 
-#ifndef TORRENT_CPUID_HPP_INCLUDED
-#define TORRENT_CPUID_HPP_INCLUDED
+#ifndef TORRENT_FFS_HPP_INCLUDE
+#define TORRENT_FFS_HPP_INCLUDE
 
-#include "libtorrent/config.hpp"
+#include <cstdint>
+#include "libtorrent/export.hpp"
+#include "libtorrent/span.hpp"
 
 namespace libtorrent { namespace aux
 {
-	// initialized by static initializers (in cpuid.cpp)
-	TORRENT_EXTRA_EXPORT extern bool const sse42_support;
-	TORRENT_EXTRA_EXPORT extern bool const mmx_support;
-	TORRENT_EXTRA_EXPORT extern bool const arm_neon_support;
-	TORRENT_EXTRA_EXPORT extern bool const arm_crc32c_support;
-} }
+	TORRENT_EXTRA_EXPORT int clz_sw(span<std::uint32_t const> buf);
+	TORRENT_EXTRA_EXPORT int clz_hw(span<std::uint32_t const> buf);
+}}
 
-#endif // TORRENT_CPUID_HPP_INCLUDED
+#endif // TORRENT_FFS_HPP_INCLUDE
--- a/include/libtorrent/aux_/ffs.hpp
+++ b/include/libtorrent/aux_/ffs.hpp
@@ -39,10 +39,14 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent { namespace aux
 {
+	// these functions expect the range to be in big-endian byte order
 	TORRENT_EXTRA_EXPORT int clz_sw(span<std::uint32_t const> buf);
+	// if this function is called in an unsupported platform, returns -1
+	// consider call always clz(buf)
 	TORRENT_EXTRA_EXPORT int clz_hw(span<std::uint32_t const> buf);
 
-	// this function statically determines if hardware of software is used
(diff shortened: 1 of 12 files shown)


src/ffs.cpp
```diff
@@ -0,0 +1,98 @@
...
+		for (int i = 0; i < num; i++)
+		{
+			if (ptr[i] == 0) continue;
+			std::uint32_t v = aux::network_to_host(ptr[i]);
+
+#if TORRENT_HAS_BUILTIN_CLZ
+			return i * 32 + __builtin_clz(v);
+#elif defined _MSC_VER
+			DWORD pos;
+			_BitScanReverse(&pos, v);
```
arvidn: my (more recent) understanding is that both __builtin_clz() and _BitScanReverse() have have software fallbacks in them too, and it may make sense to use those for clz_sw() if they're available.

test/test_ffs.cpp
```diff
@@ -0,0 +1,80 @@
...
+		{ "000000001ff00000000000000000000000000000", 35 },
+	};
+
+	for (auto const& t : tests)
+	{
+		std::fprintf(stderr, "%s\n", t.first);
+		std::uint32_t buf[5];
+		to_binary(t.first, buf);
+		TEST_EQUAL(aux::clz_sw({buf, 5}), t.second);
+		TEST_EQUAL(aux::clz_hw({buf, 5}), t.second);
```
arvidn: I would imagine that clz_hw() would require some run-time check for hardware support before being called.. no?

src/ffs.cpp
aldenml: true (to my understanding too), but then, how do you test the "pure" software implementation? I saw (don't remember where) that the fallbacks are a (very similar) variation of the `clz_sw` code compiled.

test/test_ffs.cpp
aldenml: well, `clz` is supported from `i386` the very old one, and from `armv5`, and I don't know a way to test for them. Do you?

src/ffs.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+#if (defined _MSC_VER && _MSC_VER >= 1600)
+#include <nmmintrin.h>
+#endif
+
+#include "libtorrent/aux_/disable_warnings_pop.hpp"
+
+namespace libtorrent { namespace aux
+{
+    int clz_sw(span<std::uint32_t const> buf)
```
aldenml: do you think these blanks will stop the merge :( ?

test/test_ffs.cpp
arvidn: I see. so for x86 __builtin_ctz and clz can be assumed to be efficiently implemented then

src/ffs.cpp
arvidn: yeah, that's a good point. however, wherever these functions are called, I imagine some implementation detail of this function would leak out. Specifically, the selection macros of which one to call.

maybe it would make sense to have a third function (without the _sw or _hw suffix) that does the selection as well.

src/sha1_hash.cpp
```diff
@@ -65,41 +66,11 @@ namespace libtorrent
...
-				continue;
-			}
-
-#if TORRENT_HAS_BUILTIN_CLZ
-			return ret + __builtin_clz(v);
-#elif TORRENT_HAS_SSE && defined _MSC_VER
-			DWORD pos;
-			_BitScanReverse(&pos, v);
-			return ret + 31 - pos;
+#if TORRENT_HAS_BUILTIN_CLZ || defined _MSC_VER
```
arvidn: here. everywhere clz_* is called, I imagine these macros would have to be checked. and if another implementation is added, all these places would have to be updated. another function that picks the right implementation might make sense

aldenml: I agree that's a good idea, let me add it

src/sha1_hash.cpp
```diff
@@ -63,45 +62,6 @@ namespace libtorrent
 
 #endif // TORRENT_USE_IOSTREAM
 
-	int sha1_hash::count_leading_zeroes() const
-	{
-		int ret = 0;
-		for (int i = 0; i < number_size; ++i)
-		{
-			std::uint32_t v = aux::network_to_host(m_number[i]);
```
arvidn: this byteswap, is it preserved in your new functions? That's a bit of a quirk, because it's here to support treating the sha1_hash as a byte-array as well

src/ffs.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+#if (defined _MSC_VER && _MSC_VER >= 1600)
+#include <nmmintrin.h>
+#endif
+
+#include "libtorrent/aux_/disable_warnings_pop.hpp"
+
+namespace libtorrent { namespace aux
+{
+	int clz_sw(span<std::uint32_t const> buf)
```
arvidn: it should probably be documented that these ranges are expected to be in big-endian byte order

src/sha1_hash.cpp
aldenml: hmmm...it is. I see you use it too in `bitfield::find_first_set` in your other PR. Any suggestion?

test/test_ffs.cpp
```diff
@@ -0,0 +1,85 @@
...
+		TEST_EQUAL(aux::clz({buf, 5}), t.second);
+	}
+
+#if !TORRENT_HAS_BUILTIN_CLZ && !defined _MSC_VER
+#error "expected built-in clz for unit tests architectures"
```
arvidn: maybe you could just use this to exclude that test instead. it would be nice to be able to run this test on all architectures regardless

src/ffs.cpp
aldenml: is this enough to keep the `aux::network_to_host` call inside?

include/libtorrent/aux_/ffs.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <cstdint>
+#include "libtorrent/export.hpp"
+#include "libtorrent/span.hpp"
+
+namespace libtorrent { namespace aux
+{
+	TORRENT_EXTRA_EXPORT int clz_sw(span<std::uint32_t const> buf);
+	TORRENT_EXTRA_EXPORT int clz_hw(span<std::uint32_t const> buf);
+
+	// this function statically determines if hardware of software is used
```
arvidn: s/of/or

aldenml: yes :)

test/test_ffs.cpp
aldenml: not sure about this, my idea is that you should test all variations. I imagine that if an specific platform is "strange enough" to not support these builtins (at least via GCC), then in order to properly support that platform a special consideration should be done inside `clz_hw`. But you tell me.

PR #565, opened 2016-03-23
simplification and improvement of disk_buffer_pool

arvidn:

PR #986, opened 2016-08-06
added support for flz (find last zero) and copied bitfield functions from PR #565

aldenml:


include/libtorrent/aux_/ffs.hpp
```diff
@@ -48,6 +48,17 @@ namespace libtorrent { namespace aux
 	// this function statically determines if hardware or software is used
 	// and expect the range to be in big-endian byte order
 	TORRENT_EXTRA_EXPORT int clz(span<std::uint32_t const> buf);
+
+	// these functions expect the range to be in big-endian byte order
+	// find last zero (flz)
+	TORRENT_EXTRA_EXPORT int flz_sw(span<std::uint32_t const> buf);
```
arvidn: I'm not super excited about the name "flz". Not clz either to be honest, but at least there's some wider convention there, since it's the name of the GCC builtin. I deally these functions would have a bit more intuitive names. Second, I take it this is short for "find last zero". It's not entirely obvious which bit is "last", is it the most significant or the least significant one? Also, at least bit indices are well established to start at 0 from the least significant bit. So I take it this returns the bit index of the last zero, is that right?

test/test_bitfield.cpp
```diff
@@ -262,3 +262,67 @@ TORRENT_TEST(test_resize_down)
...
+}
+
+TORRENT_TEST(find_last_clear_early)
+{
+	bitfield test1(100, true);
+	test1.clear_bit(4);
+	TEST_EQUAL(test1.find_last_clear(), 4);
+}
+
+TORRENT_TEST(find_last_clear_late)
```
arvidn: maybe it would make sense to have a case where both bit 4 and 98 are cleared, to make sure it returns the index to the last one

test/test_ffs.cpp
```diff
@@ -79,3 +79,26 @@ TORRENT_TEST(clz)
 		TEST_EQUAL(aux::clz({buf, 5}), t.second);
 	}
 }
+
+TORRENT_TEST(flz_u32)
+{
+	std::uint32_t v = 0;
+	TEST_EQUAL(aux::flz_sw(v), 31);
+	TEST_EQUAL(aux::flz_hw(v), 31);
+	TEST_EQUAL(aux::flz(v), 31);
```
arvidn: this corroborates the bit index being returned

test/test_ffs.cpp
```diff
@@ -79,3 +79,26 @@ TORRENT_TEST(clz)
...
+
+	v = 0xffffffff;
+	TEST_EQUAL(aux::flz_sw(v), -1);
+	TEST_EQUAL(aux::flz_hw(v), -1);
+	TEST_EQUAL(aux::flz(v), -1);
+
+	v = 0xff00ff00;
+	TEST_EQUAL(aux::flz_sw(v), 23);
```
arvidn: how come this doesn't return 31? if v is interpreted as big endian (assuming the test is run on a little-endian machine), the native value would be: 0x00ff00ff, and the highest index zero would be 31.

Speaking of which, these tests need to run htonl() on the values being assigned to v.

arvidn:
it's a good patch in general, just some comments

src/ffs.cpp
```diff
@@ -105,16 +105,16 @@ namespace libtorrent { namespace aux
...
-		return aux::clz_hw(buf);
+		return aux::count_leading_zeros_hw(buf);
 #else
-		return aux::clz_sw(buf);
+		return aux::count_leading_zeros_sw(buf);
 #endif
 	}
 
-	int flz_sw(span<std::uint32_t const> buf)
+	int find_last_zero_sw(span<std::uint32_t const> buf)
```
arvidn: sorry for nit-picking. normally when searching, I would expect to search from left to right, but here "last" means right-most, or highest bit-index. I would expect there to be a less ambiguous name. I can't think of any great name right now.

test/test_ffs.cpp
```diff
@@ -74,31 +75,41 @@ TORRENT_TEST(clz)
...
-	TEST_EQUAL(aux::flz_hw(v), 19);
-	TEST_EQUAL(aux::flz(v), 19);
+	TEST_EQUAL(aux::find_last_zero_sw(v), -1);
+	TEST_EQUAL(aux::find_last_zero_hw(v), -1);
+	TEST_EQUAL(aux::find_last_zero(v), -1);
+
+	v = aux::host_to_network(0xff00ff00);
+	TEST_EQUAL(aux::find_last_zero_sw(v), 31);
+	TEST_EQUAL(aux::find_last_zero_hw(v), 31);
+	TEST_EQUAL(aux::find_last_zero(v), 31);
```
arvidn: actually, this is still confusing to me. now that we see the value in host-endian order, bit-index 31 is not a zero. how does this work again?

aldenml: This is (and will be) confusing to me too :) I think the hex-literal is written like in our minds it is big-endian, regardless of the host-endianness. Since we are using `htonl` I need to write the literal in the form that we read it in little-endian. Now, the way the bits are indexed is a little odd (from left to right) but it was that way to easily accommodate to the semantic of the new `bitfield` methods. I can document the the bit index order (preferably in another PR)

arvidn: If the indexing is reversed (and the bitfield function applies 31 - x), the function could be renamed "count_trailing_ones", that would be unambiguous in how bits are indexed, and I think "trailing" is also intuitive in its meaning. What would you think about that?
Or can you think of a name that unambiguously describes the current meaning?

arvidn: oh, I guess since this applies to a span, it's not as simple to apply 31 - x

aldenml: yes, it's not as simple, actually, `count_leading_zeros` also works with the index from left-to-right, and I think it makes sense, since when you work with `span`, it's not clear the concept of endianness/bit-order to the entire buffer that seems to put the bytes as a string of bits.

aldenml: I realize that `count_leading_zeros` does not return an index, so it's clear from the name

aldenml: I think `count_trailing_ones` it is

test/test_ffs.cpp
```diff
@@ -81,35 +81,35 @@ TORRENT_TEST(count_leading_zeros)
...
-	TEST_EQUAL(aux::find_last_zero(v), -1);
+	TEST_EQUAL(aux::count_trailing_ones_sw(v), -1);
+	TEST_EQUAL(aux::count_trailing_ones_hw(v), -1);
+	TEST_EQUAL(aux::count_trailing_ones(v), -1);
 
 	v = aux::host_to_network(0xff00ff00);
-	TEST_EQUAL(aux::find_last_zero_sw(v), 31);
-	TEST_EQUAL(aux::find_last_zero_hw(v), 31);
-	TEST_EQUAL(aux::find_last_zero(v), 31);
+	TEST_EQUAL(aux::count_trailing_ones_sw(v), 31);
```
aldenml: and now it's not clear here either

test/test_ffs.cpp
aldenml: i reverted to the previous commit, the name `count_trailing_ones` is not appropriate either

arvidn:
I take it you saw the test failures on msvc

src/ffs.cpp
```diff
@@ -151,26 +151,26 @@ namespace libtorrent { namespace aux
 			std::uint32_t v = ~aux::network_to_host(ptr[i]);
 
 #if TORRENT_HAS_BUILTIN_CTZ
-			return i * 32 + 31 - __builtin_ctz(v);
+			return (num - i - 1) * 32 + __builtin_ctz(v);
 #elif defined _MSC_VER
 			DWORD pos;
 			_BitScanForward(&pos, v);
-			return i * 32 + 31 - pos;
+			return i * 32 + pos;
```
aldenml: I missed here the `(num - i - 1)` for msvc

PR #987, opened 2016-08-06
fix re-prioritizing a piece while it is downloading

ssiloti:
update_piece_state does not expect that the piece's priority has also changed
since the last time the piece was updated. Make sure the piece's priority is
updated before trying to update the state.

@arvid: I'm not very confident about the correctness of this fix since I still don't fully understand the code. I at least don't seem to be hitting the assert on line 955 any more.

### DIFF
--- a/src/piece_picker.cpp
+++ b/src/piece_picker.cpp
@@ -1834,6 +1834,18 @@ namespace libtorrent
 		p.piece_priority = new_piece_priority;
 		int new_priority = p.priority(this);
 
+		if (prev_priority != new_priority && !m_dirty)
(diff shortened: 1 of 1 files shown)


arvidn:
thanks! I'll take a closer look. I started looking at this from the unit test end. I will try to reproduce it in a test also, to get a better understanding of the problem.

arvidn:
it does look like this patch does introduce a regression in the piece priorities though, I'll keep looking

arvidn:
hm.. I'm increasingly suspecting that this is in fact the correct fix.

PR #990, opened 2016-08-06
handle short reads when exporting a file

ssiloti:
Pieces in the part file are not guaranteed to be fully written so short
reads need to be handled rather than asserting.

### DIFF
--- a/src/part_file.cpp
+++ b/src/part_file.cpp
@@ -334,7 +334,7 @@ namespace libtorrent
 				TORRENT_ASSERT(!ec);
 				if (ec || v.iov_len == 0) return;
 
-				int64_t ret = f.writev(file_offset, &v, 1, ec);
+				boost::int64_t ret = f.writev(file_offset, &v, 1, ec);
 				TORRENT_ASSERT(ec || ret == v.iov_len);
 				if (ec || ret != v.iov_len) return;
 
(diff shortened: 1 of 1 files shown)


ssiloti:
Actually I don't think this is right. I think the undersized part file was a result of a previous crash.

arvidn:
it seems reasonable to not assert on an invalid part file still though, doesn't it? I would consider asserts to be for internal consistency checks, or sometimes precondition checks.

ssiloti:
Yeah. Turns out I was wrong about it being due to a previous crash anyways. I was able to get it to reproduce even with a clean part file.

src/part_file.cpp
```diff
@@ -329,14 +329,14 @@ namespace libtorrent
...
-
-				ret = f.writev(file_offset, &v, 1, ec);
-				TORRENT_ASSERT(ec || ret == block_to_copy);
-				if (ec || ret != block_to_copy) return;
+				file::iovec_t v = { buf.get(), size_t(block_to_copy) };
+				v.iov_len = m_file.readv(slot_offset + piece_offset, &v, 1, ec);
+				TORRENT_ASSERT(!ec);
+				if (ec || v.iov_len == 0) return;
+
+				int64_t ret = f.writev(file_offset, &v, 1, ec);
```
arvidn: you need std::int64_t

PR #991, opened 2016-08-07
build python bindings with autotools on travis

arvidn:
and forward configured compiler to setup.py


arvidn:
my understanding is that distutils, by default, picks the compiler that python was built with to build extensions. This makes it non-trivial (as far as I can tell) to make it use a modern compiler. Say, one that supports C++11. Even though passing along `-std=c++11` is simple, the actual compiler binary may not be modern enough.

PR #993, opened 2016-08-08
created session_params and refactor

aldenml:


include/libtorrent/session.hpp
```diff
@@ -131,6 +121,44 @@ namespace libtorrent
...
+			, dht_storage_constructor(dht::dht_default_storage_constructor)
+#endif
+		{}
+		session_params(session_params const&) = default;
+		session_params(session_params&&) = default;
+		session_params& operator=(session_params const&) = default;
+		session_params& operator=(session_params&&) = default;
+
+		// flags to be passed in to the constructor
+		enum session_flags_t
```
arvidn: if we're breaking the API here anyway, I think this enum should be moved out from this struct and made into an "enum class". There needs to be some amount of work to make them behave as flags though. I'm envisioning something like https://www.justsoftwaresolutions.co.uk/cplusplus/using-enum-classes-as-bitfields.html

include/libtorrent/session.hpp
```diff
@@ -131,6 +121,44 @@ namespace libtorrent
 		boost::shared_ptr<aux::session_impl> m_impl;
 	};
 
+	struct TORRENT_EXPORT session_params
+	{
+		session_params(settings_pack sp = settings_pack()
+			, int flags = default_flags)
+			: settings(sp)
+			, start_flags(flags)
+#ifndef TORRENT_DISABLE_DHT
```
arvidn: Do you think it would be possible to always have the dht_storage_constructor field? In general, it makes things a lot simpler if the ABI is not affected by build options like these

include/libtorrent/session.hpp
```diff
@@ -131,6 +121,44 @@ namespace libtorrent
...
+
+			default_flags = add_default_plugins
+		};
+
+		settings_pack settings;
+
+		int start_flags;
+
+#ifndef TORRENT_DISABLE_EXTENSIONS
+		std::vector<boost::shared_ptr<plugin>> extensions;
```
arvidn: same thing here, this define should ideally not affect the ABI

include/libtorrent/session.hpp
```diff
@@ -131,6 +121,44 @@ namespace libtorrent
 		boost::shared_ptr<aux::session_impl> m_impl;
 	};
 
+	struct TORRENT_EXPORT session_params
```
arvidn: I would imagine this class would be a lot more useful if you could construct it, and add stuff to it in a chain. Something like: `session s(session_params(settings, flags).add_extension(foo).add_extension(bar));`

I'm not sure that's possible though

include/libtorrent/session.hpp
```diff
@@ -131,6 +121,44 @@ namespace libtorrent
...
+
+		int start_flags;
+
+#ifndef TORRENT_DISABLE_EXTENSIONS
+		std::vector<boost::shared_ptr<plugin>> extensions;
+#endif
+
+		libtorrent::dht_settings dht_settings;
+
+		dht::dht_storage_constructor_type dht_storage_constructor;
```
arvidn: oh, I see. the member is here unconditionally, it's just not initialized.

include/libtorrent/session.hpp
```diff
@@ -147,23 +175,37 @@ namespace libtorrent
...
-		// 
+		//
 		// The ``flags`` parameter can be used to start default features (upnp &
 		// nat-pmp) and default plugins (ut_metadata, ut_pex and smart_ban). The
 		// default is to start those features. If you do not want them to start,
 		// pass 0 as the flags parameter.
-		session(settings_pack pack = settings_pack()
-			, int flags = start_default_features | add_default_plugins)
+		session(settings_pack pack
+			, int flags = session_params::default_flags)
```
arvidn: if we're rethinking this API anyway, does it make sense to have these flags at all? The main reason I added them was because I discovered that Miro did not enable the ut_metadata extension nor the DHT, ever. So I figured the default should probably be to do that things most people would want. But maybe there's a better way. Is anything associated with default features anymore?

include/libtorrent/session.hpp
aldenml: I will check this

include/libtorrent/session.hpp
aldenml: these pseudo-building/fluently pattern style is kind of controversial, I use it, but in the context of the need for a more idiomatic client code. Let me skip this in this PR (unless you think it's necessary).

include/libtorrent/session.hpp
aldenml: thinking about it, keeping this flag in the current form could create dangerous inconsistencies at creation time, since the `params.extenstions` could have the default plugins already, re-thinking this...

include/libtorrent/session.hpp
aldenml: I rather remove this enum :D ... I imagine swig struggling big time with the idea

include/libtorrent/session.hpp
arvidn: I'm mostly thinking out loud

aldenml:
New (appveyor)?
```
ERROR: test_from_torrent_info (__main__.test_create_torrent)
  File "test.py", line 17, in test_from_torrent_info
    entry = ct.generate()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x82 in position 0: invalid start byte
```

arvidn:
yeah, this happened in the merge from RC_1_1 last night. I haven't had a chance to look at it yet

include/libtorrent/session.hpp
```diff
@@ -147,23 +158,37 @@ namespace libtorrent
...
-		// 
+		//
 		// The ``flags`` parameter can be used to start default features (upnp &
 		// nat-pmp) and default plugins (ut_metadata, ut_pex and smart_ban). The
 		// default is to start those features. If you do not want them to start,
 		// pass 0 as the flags parameter.
-		session(settings_pack pack = settings_pack()
-			, int flags = start_default_features | add_default_plugins)
+		session(settings_pack pack
+			, bool default_plugins = true)
```
aldenml: what do you think about instead changing this parameter, deprecate the entire constructor?

arvidn: oh, I didn't notice you changed the flags to a bool. In general I would prefer flags over a bool, because it makes the call site much more readable.

Regarding the constructor, I'm a bit hesitant to remove or deprecate the existing constructor. I think we should keep it for backwards compatibility and add a new one that's well thought through, and that can survive longer. Maybe instead of specifying a default plugin flag, the vector of extensions can be pre-populated with the default ones, for instance.

aldenml: ok, I will do that

arvidn:
the python test failure should be fixed now

include/libtorrent/session.hpp
```diff
@@ -131,6 +123,34 @@ namespace libtorrent
...
 
+	// The session_params is a parameters pack for configuring the session
+	// before it's started.
+	struct TORRENT_EXPORT session_params
+	{
+		// This constructor can be used to start with the default plugins
+		// (ut_metadata, ut_pex and smart_ban). The default values in the
+		// settings is to start the default features like upnp, nat-pmp,
+		// and dht for example.
+		session_params(settings_pack const sp = settings_pack());
```
arvidn: "const" should not be here. If the argument should be const, that's an implementation detail that should only be mentioned in the implementation. However, I suspect that we would like to allow this to be moved

include/libtorrent/session.hpp
```diff
@@ -131,6 +123,34 @@ namespace libtorrent
...
+		// This constructor helps to configure the set of initial plugins
+		// to be added to the session before it's started.
+		session_params(settings_pack const sp
```
arvidn: same thing with const here

include/libtorrent/session.hpp
```diff
@@ -281,7 +332,8 @@ namespace libtorrent
 
 	private:
 
-		void start(int flags, settings_pack pack, io_service* ios);
+		void start(session_params const params, io_service* ios);
+		void start(int const flags, settings_pack const sp, io_service* ios);
```
arvidn: same thing with const here

include/libtorrent/settings_pack.hpp
```diff
@@ -1332,7 +1332,7 @@ namespace libtorrent
 			// ``utp_syn_resends`` is the number of SYN packets that are sent (and
 			// timed out) before giving up and closing the socket.
 			// ``utp_num_resends`` is the number of times a packet is sent (and
-			// lossed or timed out) before giving up and closing the connection.
+			// loosed or timed out) before giving up and closing the connection.
```
arvidn: actually, I think this was meant to say "lost"

src/session.cpp
```diff
@@ -357,6 +323,33 @@ namespace libtorrent
...
 	}
 
+	namespace
+	{
+		std::vector<boost::shared_ptr<plugin>> default_plugins()
+		{
+			std::vector<boost::shared_ptr<plugin>> v;
+#ifndef TORRENT_DISABLE_EXTENSIONS
+			using wrapper = session_impl::session_plugin_wrapper;
+			boost::shared_ptr<plugin> ut_pex(new wrapper(create_ut_pex_plugin));
```
arvidn: please use make_shared<> here instead of naked new

src/session.cpp
```diff
@@ -357,6 +323,33 @@ namespace libtorrent
...
+			boost::shared_ptr<plugin> ut_metadata(new wrapper(create_ut_metadata_plugin));
+			boost::shared_ptr<plugin> smart_ban(new wrapper(create_smart_ban_plugin));
+
+			v.push_back(ut_pex);
```
arvidn: you should either move the pointers into the vector, or push_back() the temporary that make_shared<> returns

aldenml: that was my first approach, but the `explicit` keyword in `session_plugin_wrapper` was breaking the clang compilation (deprecated copy constructor). I will change it

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -222,6 +222,7 @@ namespace libtorrent
 			struct session_plugin_wrapper : plugin
 			{
 				explicit session_plugin_wrapper(ext_function_t const& f) : m_f(f) {}
+				explicit session_plugin_wrapper(session_plugin_wrapper const& p) : m_f(p.m_f) {}
```
arvidn: I'm not sure what it means for a copy constructor to be explicit.. also, I think you could just say = default; here, no?

src/session.cpp
```diff
@@ -325,29 +325,27 @@ namespace libtorrent
...
-			boost::shared_ptr<plugin> smart_ban(new wrapper(create_smart_ban_plugin));
-
-			v.push_back(ut_pex);
-			v.push_back(ut_metadata);
-			v.push_back(smart_ban);
+			return {
+				boost::make_shared<wrapper>(wrapper(create_ut_pex_plugin)),
+				boost::make_shared<wrapper>(wrapper(create_ut_metadata_plugin)),
+				boost::make_shared<wrapper>(wrapper(create_smart_ban_plugin))
+			};
```
arvidn: +1

src/session.cpp
```diff
@@ -325,29 +325,27 @@ namespace libtorrent
...
+#else
+			return {};
 #endif
-			return v;
 		}
 	}
 
-	void session::start(int const flags, settings_pack const sp, io_service* ios)
+	void session::start(int flags, settings_pack sp, io_service* ios)
```
arvidn: I do think it's reasonable for the flags to be const here though.. but I don't think sp can be const, since we're moving it (in fact, I'm a bit surprised std::move() accepts a const value)

src/session_impl.cpp
```diff
@@ -465,7 +464,7 @@ namespace aux {
 		// ---- auto-cap max connections ----
 		int max_files = max_open_files();
 		// deduct some margin for epoll/kqueue, log files,
-		// futexes, shared objects etc.
+		// mutexes, shared objects etc.
```
arvidn: actually, this was meant to say "futexes" (as in the linux kernel synchronization primitive)

src/session.cpp
aldenml: I was surprised too...not sure if going with everything `const` again or not, you tell me

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -222,7 +222,7 @@ namespace libtorrent
 			struct session_plugin_wrapper : plugin
 			{
 				explicit session_plugin_wrapper(ext_function_t const& f) : m_f(f) {}
-				explicit session_plugin_wrapper(session_plugin_wrapper const& p) : m_f(p.m_f) {}
+				explicit session_plugin_wrapper(session_plugin_wrapper const& p) = default;
```
aldenml: it fails with clang, reverting it
```
../include/libtorrent/extensions.hpp:188:11: error: definition of implicit copy constructor for 'plugin' is deprecated because it has a user-declared destructor [-Werror,-Wdeprecated]
  virtual ~plugin() {}
          ^
../src/session.cpp:335:12: note: in instantiation of function template specialization 'boost::make_shared<libtorrent::aux::session_impl::session_plugin_wrapper, libtorrent::aux::session_impl::session_plugin_wrapper>' requested here
    boost::make_shared<wrapper>(wrapper(create_ut_pex_plugin)),
           ^
../include/libtorrent/aux_/session_impl.hpp:225:14: note: implicit copy constructor for 'plugin' first required here
    explicit session_plugin_wrapper(session_plugin_wrapper const& p) = default;
             ^
/usr/local/include/boost/smart_ptr/make_shared_object.hpp:254:17: note: implicit copy constructor for 'libtorrent::aux::session_impl::session_plugin_wrapper' first required here
    ::new( pv ) T( boost::detail::sp_forward<Args>( args )... );
                ^
1 error generated.
```

src/session.cpp
```diff
@@ -357,6 +323,32 @@ namespace libtorrent
...
+	namespace
+	{
+		std::vector<boost::shared_ptr<plugin>> default_plugins(
+			bool empty = false)
+		{
+#ifndef TORRENT_DISABLE_EXTENSIONS
+			if (empty) return {};
+			using wrapper = session_impl::session_plugin_wrapper;
+			return {
+				boost::make_shared<wrapper>(wrapper(create_ut_pex_plugin)),
```
arvidn: is the wrapper constructor protected or private?

aldenml: public

PR #996, opened 2016-08-11
use boost::string_ref

arvidn:


aldenml:
Question: What do you think about creating a `string_view` that inherits from `span` and add the missing/necessary methods? It seems to me that `boost::string_ref` is not getting love
https://github.com/boostorg/utility/commit/1caa745dd72b2cffebafa904658fde0e9cac1232
https://github.com/boostorg/utility/commit/0b492bee9cca261066846216f0ba36602e2216b8

arvidn:
the reason I chose string_ref over string_view is that the latter was introduced in boost 1.61 (iirc), which I believe may be a little bit too new to have a dependency on. I would expect to transition to string_view at some point, be it boost or std

arvidn:
btw, do you know what's up with this? https://travis-ci.org/arvidn/libtorrent/jobs/151491129

aldenml:
It was some error during the `debootstrap` stage, probably internet related
```
E: Couldn't download packages: libgcrypt20 libselinux1 libsemanage1 make libmpfr4 libncursesw5
cp: cannot create regular file ‘rootfs/usr/bin/’: No such file or directory
chroot: failed to run command ‘/debootstrap/debootstrap’: No such file or directory
chroot: failed to run command ‘mount’: No such file or directory
```

aldenml:
I'm asking because of the ABI stability of `bdecode`, changing it to use something that's already out of favor feels a little strange to me.

aldenml:
Now the error is
```
E: Couldn't download packages: libgdbm3
cp: cannot create regular file ‘rootfs/usr/bin/’: No such file or directory
```
travis is broadcasting that they are having problems with the images

arvidn:
@aldenml what do you think about this? boost::string_view is too new for travis and appveyor, so I'm wrapping either in libtorrent::string_view. They are compatible enough for these uses.

aldenml:
I think it's a very good solution, and easy to transform to the `std::string_view` when the time arrives.

PR #997, opened 2016-08-12
minor cleanup

aldenml:
looking for a solution of the posix layer, btw

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -301,7 +301,7 @@ namespace aux {
 
 		return SSL_TLSEXT_ERR_OK;
 	}
-	} // anonymous namesoace
+	} // anonymous namespace
 #endif
 
 	session_impl::session_impl(io_service& ios)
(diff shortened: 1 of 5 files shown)


src/session_impl.cpp
```diff
@@ -4675,14 +4675,10 @@ namespace aux {
...
-			boost::shared_ptr<torrent_plugin>(torrent_handle const&, void*)> >
-			torrent_plugins_t;
-
-		for (torrent_plugins_t::const_iterator i = params.extensions.begin()
-			, end(params.extensions.end()); i != end; ++i)
+		for (auto& ext : params.extensions)
 		{
-			torrent_ptr->add_extension((*i)(handle, params.userdata));
+			boost::shared_ptr<torrent_plugin> tp(ext(handle, params.userdata));
+			if (tp) torrent_ptr->add_extension(tp);
```
arvidn: since you give this pointer a name here, you have to std::move() it

aldenml: ok, I will change it here and in another place. I'm not sure what's is more expensive in this case btw.

PR #999, opened 2016-08-13
no more deprecated std::random_shuffle

aldenml:
No a big fan of `extern std::mt19937 random_engine;` but I can't see any other solution

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -118,7 +118,7 @@ script:
 # libtorrent is the name of the test suite target
   - cd test
   - 'if [ "$variant" != "" ]; then
-      bjam -j3 warnings-as-errors=on ssl=$ssl crypto=$crypto debug-iterators=on picker-debugging=on invariant-checks=full $toolset variant=$variant libtorrent test_natpmp enum_if -l300 &&
+      travis_retry bjam -j3 warnings-as-errors=on ssl=$ssl crypto=$crypto debug-iterators=on picker-debugging=on invariant-checks=full $toolset variant=$variant libtorrent test_natpmp enum_if -l300 &&
       if [ "$coverage" == "1" ]; then
         codecov --root .. --gcov-exec gcov-5;
       fi;
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,8 +97,8 @@ install:
       wget -O boost.zip http://pilotfiber.dl.sourceforge.net/project/boost/boost/1.55.0/boost_1_55_0.zip;
       unzip -qq boost.zip;
       export BOOST_ROOT=$PWD/boost_1_55_0;
-      sudo apt-get install -y qemu-user-static debootstrap;
-      sudo debootstrap --variant=minbase --arch armhf --foreign --include=build-essential testing rootfs;
+      travis_retry sudo apt-get install -y qemu-user-static debootstrap;
+      travis_retry sudo debootstrap --variant=minbase --arch armhf --foreign --include=build-essential testing rootfs;
       sudo cp /usr/bin/qemu-arm-static rootfs/usr/bin/;
       sudo chroot rootfs /debootstrap/debootstrap --second-stage;
       sudo chroot rootfs mount -t proc none /proc;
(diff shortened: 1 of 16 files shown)


src/random.cpp
```diff
@@ -33,41 +33,21 @@ POSSIBILITY OF SUCH DAMAGE.
...
-
-	std::uint32_t random(std::uint32_t max)
-	{
-		// TODO: versions prior to msvc-14 (visual studio 2015) do
-		// not generate thread safe initialization of statics
-		static random_device dev;
-		static mt19937 random_engine(dev());
-		return uniform_int_distribution<std::uint32_t>(0, max)(random_engine);
+		static std::random_device dev;
+		std::mt19937 random_engine(dev());
```
arvidn: this is problematic, because the initialization order of global variables aren't defined.
It goes beyond this too though, if any global initializer happens to call aux::random_shuffle, it would end up potentially using "random_engine" before it was initialized.

The simplest way to solve this problem is to have a function that returns the random number generator instead, and it can have a local static variable which it returns a reference to. That way, whoever calls that function first will be responsible to initialize the object, and it would be guaranteed to be valid by the time it's first used. random() have this effect for instance.

aldenml: Thanks for the very illustrative explanation (actually I was expecting some observation like this). Let me accommodate that.

arvidn:
did you see the msvc error?
https://ci.appveyor.com/project/arvidn/libtorrent/build/3447/job/n5sf08a7uf8v47wr#L866

aldenml:
The failure in appveryor is very suspicious https://ci.appveyor.com/project/arvidn/libtorrent/build/3455/job/44tenmiqnc3m046w#L1840

not clear to me if it's due to `lt::aux::random_shuffle(all_nodes.begin(), all_nodes.end());`

arvidn:
I think that's a poor test, that depends on the random number outcome, there should probably be more margin added to the test

arvidn:
I'm sorry about these kinds of test failures. I believe the non-deterministic behavior comes from that the random number generator is not reset in the beginning of each test, which means the order in which tests are run will produce different results. Ideally I would fix that at some point. For now, I take it it's unlikely that this patch _actually_ worsened the routing table/bootstrapping performance. So feel free to adjust the expectation in that test to match the result on linux

arvidn:
or you could try to shuffle the nods even more frequently to see if that helps

aldenml:
well, I learned a little bit more exploring the dht tests :)
In my last commit I'm doing the shuffle more frequently and it works, do you mind restarting the two failed travis jobs?

aldenml:
@arvidn the use of `travis_retry` is necessary in the `debootstrap` due to the poor networking performance in travis, but I would like to know what do you think of using it in the tests run.

.travis.yml
```diff
@@ -118,7 +118,7 @@ script:
 # libtorrent is the name of the test suite target
   - cd test
   - 'if [ "$variant" != "" ]; then
-      bjam -j3 warnings-as-errors=on ssl=$ssl crypto=$crypto debug-iterators=on picker-debugging=on invariant-checks=full $toolset variant=$variant libtorrent test_natpmp enum_if -l300 &&
+      travis_retry bjam -j3 warnings-as-errors=on ssl=$ssl crypto=$crypto debug-iterators=on picker-debugging=on invariant-checks=full $toolset variant=$variant libtorrent test_natpmp enum_if -l300 &&
```
arvidn: I can see the other command failings because of networking issues, but how would this be affected by networking? short of the workspace directory itself being a remote drive

arvidn:
I don't see any problem with using travis_retry

.travis.yml
aldenml: I speculate that the networking issue is not limited to the communication with the outside, but deep down in the networking stack in the (crazy) vm environment they setup. Maybe network IO related calls freeze and that's part of the reason some test fail randomly.

aldenml:
Ok, do you want to keep it in the tests or not? (I have no other changes in this PR)

PR #1001, opened 2016-08-15
using string_view in node::incoming_request and plugin::on_dht_request

aldenml:
I'm a little worried about the performance impact of this change, what I want to do at the end is change the signature of `plugin::on_dht_request`

### DIFF
--- a/src/entry.cpp
+++ b/src/entry.cpp
@@ -43,15 +43,11 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/entry.hpp"
 #include "libtorrent/hex.hpp"
 
-namespace
-{
-}
-
 namespace libtorrent
 {
 	namespace detail
 	{
-		TORRENT_EXPORT char const* integer_to_str(char* buf, int size
+		char const* integer_to_str(char* buf, int size
 			, entry::integer_type val)
 		{
 			int sign = 0;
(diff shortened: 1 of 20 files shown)


arvidn:
I don't see why this would have any worse performance. It's a good change. Is there a reason to hold off on changing the plugin API? btw, that should be mentioned in the Changelog

aldenml:
good, I will go ahead with the plugin API (I wanted a quick review first)

aldenml:
btw, I was worried about the extra conversions of the static string constants during the `==` operator, but I assume the compiler could optimize this cleverly.

arvidn:
I would expect string_view to have a comparison operator against `template<size_t N> char[N]`, making it even more efficient than `strcmp()`. At the very least, I would expect a future version of string_view to have that

src/entry.cpp
```diff
@@ -43,15 +43,11 @@ POSSIBILITY OF SUCH DAMAGE.
...
 
-namespace
-{
-}
-
 namespace libtorrent
 {
 	namespace detail
 	{
-		TORRENT_EXPORT char const* integer_to_str(char* buf, int size
```
arvidn: I believe this is used from a public inlined function, isn't it? (in which case it does need to be exported)

aldenml: true, about to remove it (btw, it bothers me that this function is only used in `bencode` but declared an defined in `entry`)

arvidn: yeah, bencode and entry have an interesting (tight) relationship :p

PR #1002, opened 2016-08-15
peer_plugin refactor

aldenml:
Before going any further, I want to get review for this idea. My goal is remove the method `peer_plugin::type()` because it is hard to see a clear use of it. It seems to me that it was added mainly because of `ut_pex`.  But in order to do that I need to perform this refactor and a mayor drawback is the additional field in `bt_peer_connection`, that could be a waste of memory if no `ut_pex` is active.

Design wise, the special nature of the `ut_pex` is already captured implicitly in the cross reference via the standalone method `was_introduced_by`. With the additional field, it is explicit.

### DIFF
--- a/include/libtorrent/extensions/ut_pex.hpp
+++ b/include/libtorrent/extensions/ut_pex.hpp
@@ -48,28 +48,7 @@ POSSIBILITY OF SUCH DAMAGE.
 namespace libtorrent
 {
 	struct torrent_plugin;
-	struct peer_plugin;
 	struct torrent_handle;
-
-	struct TORRENT_EXPORT ut_pex_peer_store
-	{
-		// stores all peers this peer is connected to. These lists
-		// are updated with each pex message and are limited in size
-		// to protect against malicious clients. These lists are also
-		// used for looking up which peer a peer that supports holepunch
-		// came from.
-		// these are vectors to save memory and keep the items close
-		// together for performance. Inserting and removing is relatively
-		// cheap since the lists' size is limited
-		using peers4_t = std::vector<std::pair<address_v4::bytes_type, std::uint16_t>>;
-		peers4_t m_peers;
-#if TORRENT_USE_IPV6
-		using peers6_t = std::vector<std::pair<address_v6::bytes_type, std::uint16_t>>;
-		peers6_t m_peers6;
-#endif
-
-		bool was_introduced_by(tcp::endpoint const& ep);
-	};
 
 	// constructor function for the ut_pex extension. The ut_pex
 	// extension allows peers to gossip about their connections, allowing
--- a/include/libtorrent/extensions/ut_pex.hpp
+++ b/include/libtorrent/extensions/ut_pex.hpp
@@ -37,7 +37,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "libtorrent/config.hpp"
 #include "libtorrent/socket.hpp" // for endpoint
-#include "libtorrent/address.hpp" // for endpoint
+#include "libtorrent/address.hpp"
 
 #include "libtorrent/aux_/disable_warnings_push.hpp"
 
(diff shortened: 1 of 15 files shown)


include/libtorrent/extensions/ut_pex.hpp
```diff
@@ -37,6 +37,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "libtorrent/config.hpp"
 #include "libtorrent/socket.hpp" // for endpoint
```
arvidn: maybe you could remove the comment for this include then..

include/libtorrent/extensions/ut_pex.hpp
```diff
@@ -50,6 +51,26 @@ namespace libtorrent
...
+	{
+		// stores all peers this peer is connected to. These lists
+		// are updated with each pex message and are limited in size
+		// to protect against malicious clients. These lists are also
+		// used for looking up which peer a peer that supports holepunch
+		// came from.
+		// these are vectors to save memory and keep the items close
+		// together for performance. Inserting and removing is relatively
+		// cheap since the lists' size is limited
+		typedef std::vector<std::pair<address_v4::bytes_type, std::uint16_t>> peers4_t;
```
arvidn: I prefer: using peers4_t = std::vector<...

src/ut_pex.cpp
```diff
@@ -685,9 +679,8 @@ namespace libtorrent
 		return boost::shared_ptr<torrent_plugin>(new ut_pex_plugin(*t));
 	}
 
-	bool was_introduced_by(peer_plugin const* pp, tcp::endpoint const& ep)
+	bool was_introduced_by(ut_pex_peer_store const* p, tcp::endpoint const& ep)
```
arvidn: is there a reason for this function to not be a member of ut_pex_peer_store?

aldenml: none, I just wanted to hack the refactor with the minimal change. Does this comments mean that it's ok to add the extra member?

arvidn: at a first glance that seems like a reasonable idea at least

src/ut_pex.cpp
```diff
@@ -667,46 +653,47 @@ namespace libtorrent { namespace
 		if (pc.type() != peer_connection::bittorrent_connection)
 			return boost::shared_ptr<peer_plugin>();
 
-		return boost::shared_ptr<peer_plugin>(new ut_pex_peer_plugin(m_torrent
-			, *pc.native_handle(), *this));
+		bt_peer_connection* c = static_cast<bt_peer_connection*>(pc.native_handle().get());
+		auto p = boost::make_shared<ut_pex_peer_plugin>(ut_pex_peer_plugin(m_torrent, *c, *this));
```
arvidn: it's unusual to construct a temporary object and then copy it into the heap allocated one. Does it not work to pass the constructor arguments directly to make_shared? If the constructor is private, maybe you could make boost::make_shared a friend. Although, that may be tricky, since you may have to specify the exact overload with the right number of template arguments..

src/ut_pex.cpp
```diff
@@ -667,46 +653,47 @@ namespace libtorrent { namespace
...
 #if TORRENT_USE_IPV6
 		if (ep.address().is_v4())
 		{
 #endif
-			ut_pex_peer_plugin::peers4_t::value_type v(ep.address().to_v4().to_bytes(), ep.port());
-			ut_pex_peer_plugin::peers4_t::const_iterator i
-				= std::lower_bound(p->m_peers.begin(), p->m_peers.end(), v);
-			return i != p->m_peers.end() && *i == v;
+			peers4_t::value_type v(ep.address().to_v4().to_bytes(), ep.port());
+			peers4_t::const_iterator i
```
arvidn: "auto" is also acceptable :)

src/ut_pex.cpp
arvidn: actually, since this type has internal linkage, I don't think the private constructor buys us anything

src/ut_pex.cpp
```diff
@@ -667,46 +653,47 @@ namespace libtorrent { namespace
...
+
+	boost::shared_ptr<torrent_plugin> create_ut_pex_plugin(torrent_handle const& th, void*)
+	{
+		torrent* t = th.native_handle().get();
+		if (t->torrent_file().priv() || (t->torrent_file().is_i2p()
+			&& !t->settings().get_bool(settings_pack::allow_i2p_mixed)))
+		{
+			return boost::shared_ptr<torrent_plugin>();
+		}
+		return boost::shared_ptr<torrent_plugin>(new ut_pex_plugin(*t));
```
arvidn: I think the same argument applies here. ut_pex_plugin has internal linkage and could have a public constructor, and boost::make_shared<> could be used

arvidn:
+1

PR #1003, opened 2016-08-16
refactor in crypto_plugin to use span

aldenml:


aldenml:
@arvidn What do you think about returning a `tuple` in `crypto_plugin::decrypt` instead of using out parameters?

include/libtorrent/pe_crypto.hpp
```diff
@@ -142,9 +139,8 @@ namespace libtorrent
 		rc4_handler();
 
 		// Input keys must be 20 bytes
-		// TODO: 4 use uint768_t here instead of pointer + length
-		void set_incoming_key(unsigned char const* key, int len) override;
-		void set_outgoing_key(unsigned char const* key, int len) override;
+		void set_incoming_key(span<char const> key) override;
```
arvidn: these keys are fixed length, aren't they? Is uint768_t less convenient than a span? The internal representation is the large uint, right? but maybe it would require conversion and not be very practical for that reason

src/pe_crypto.cpp
```diff
@@ -280,10 +283,11 @@ namespace libtorrent
 		m_rc4_outgoing.y = 0;
 	}
 
-	void rc4_handler::set_incoming_key(unsigned char const* key, int len)
+	void rc4_handler::set_incoming_key(span<char const> key)
```
arvidn: I guess they are 20 bytes, so uint768_t would not work

arvidn:
seems pretty reasonable

aldenml:
Now I'm not sure what's "reasonable" :) this PR or the idea of returning a `tuple`

appveyor.yml
```diff
@@ -99,7 +99,7 @@ build_script:
 
 test_script:
 - cd %ROOT_DIRECTORY%\test
-- b2.exe -l400 --hash warnings-as-errors=on -j2 %compiler% address-model=%model% debug-iterators=on picker-debugging=on invariant-checks=full variant=%variant% %linkflags% %include% link=shared ssl=%ssl% crypto=%crypto% win-tests
+- appveyor-retry b2.exe -l400 --hash warnings-as-errors=on -j2 %compiler% address-model=%model% debug-iterators=on picker-debugging=on invariant-checks=full variant=%variant% %linkflags% %include% link=shared ssl=%ssl% crypto=%crypto% win-tests
```
arvidn: I'm a bit worried about this. I really do wish to feel the pain of flaky test to get me to fix them

arvidn: on the other hand, this is a lot more efficient than to rebuild everything on appveyor

arvidn:
you mean to return both the bytes-consumed and bytes_transferred? That seems reasonable. the recv_buffer however is mutated, it's not a pure out-parameter

appveyor.yml
aldenml: yes, it worries me too... well, I imagine the "retry" can/should be removed at some point in the future

aldenml:
it would be `tuple(consume, produce, packet_size)`, btw the current implementation of RC4 only uses `produce` but the code is prepared for all of them.

aldenml:
if this PR gets green I rather make the `tuple` change in another PR

arvidn:
lgtm

PR #1004, opened 2016-08-16
extensions refactor to use span and std::shared_ptr

aldenml:


arvidn:
looks good. please update ChangeLog as well, especially since this alters the API

aldenml:
OK, btw, something very strange is happening when  using `std::unordered_multimap` in `resolve_links` when compiling with gcc

arvidn:
it looks like there may be some subtle difference in semantics between boost::shared_ptr and std::shared_ptr

aldenml:
it's very strange, I'm testing in my account

aldenml:
This PR just went off track with a nasty memory crash running the simulation tests and a very strange behavior using `std::unordered_multimap` with `gcc`, closing it to redo in smaller parts.

PR #1005, opened 2016-08-17
extensions refactor to use span and std::shared_ptr

aldenml:


include/libtorrent/extensions.hpp
```diff
@@ -158,12 +158,11 @@ POSSIBILITY OF SUCH DAMAGE.
...
-#include "libtorrent/sha1_hash.hpp" // for sha1_hash
-#include "libtorrent/session_handle.hpp"
-#include "libtorrent/peer_connection_handle.hpp"
-#include "libtorrent/span.hpp"
-#include "libtorrent/string_view.hpp"
+
+#include <libtorrent/config.hpp>
+#include <libtorrent/span.hpp>
+#include <libtorrent/string_view.hpp>
+#include <libtorrent/socket.hpp>
```
arvidn: I would expect "" would be more appropriate here, rather than <>

test/test_session_params.cpp
```diff
@@ -58,7 +58,7 @@ namespace
 
 	struct custom_plugin : plugin
 	{
-		void added(session_handle h) override
+		void added(session_handle& h) override
```
arvidn: I thought session_handle was essentially a shared_ptr<> to a session_impl. why is it passed by reference? if it shouldn't be passed by value, I imagine it should be at least const reference.

include/libtorrent/extensions.hpp
```diff
@@ -226,7 +228,7 @@ namespace libtorrent
 		{ return boost::shared_ptr<torrent_plugin>(); }
 
 		// called when plugin is added to a session
-		virtual void added(session_handle) {}
+		virtual void added(session_handle&) {}
```
arvidn: I imagine you meant to say const reference here, no

include/libtorrent/extensions.hpp
aldenml: for some reason (probably copy&paste) it's `<>` in `dht_storage.hpp`, I will change it in both

test/test_session_params.cpp
aldenml: I wanted to change it to keep it similar (in style) to when `peer_connection_handle` is passed. I will put the `const`.

include/libtorrent/extensions.hpp
arvidn: yeah, I've sort of had this old idea of separating the DHT implementation from libtorrent, so I believe libtorrent includes are <> inside the DHT code

arvidn:
lgtm

include/libtorrent/extensions.hpp
aldenml: ok...I'm about to push the change to `std::shared_ptr`, finally found the issue (not changing boost::enable_shared_from_this<>)

include/libtorrent/bt_peer_connection.hpp
```diff
@@ -192,10 +194,10 @@ namespace libtorrent
 
 #ifndef TORRENT_DISABLE_EXTENSIONS
 		bool supports_holepunch() const { return m_holepunch_id != 0; }
-		void set_ut_pex(boost::shared_ptr<ut_pex_peer_store> ut_pex)
-		{ m_ut_pex = ut_pex; }
+		void set_ut_pex(std::weak_ptr<ut_pex_peer_store> ut_pex)
+		{ m_ut_pex.swap(ut_pex); }
```
arvidn: I would expect a: `m_ut_pex = std::move(ut_pex);` to be more appropriate

include/libtorrent/bt_peer_connection.hpp
```diff
@@ -457,9 +459,9 @@ namespace libtorrent
 		// 0 if not supported
 		std::uint8_t m_share_mode_id;
 
-		boost::shared_ptr<ut_pex_peer_store> m_ut_pex;
+		std::weak_ptr<ut_pex_peer_store> m_ut_pex;
```
arvidn: good catch

PR #1007, opened 2016-08-18
storage_interface refactor to use span and references

aldenml:


examples/connection_tester.cpp
```diff
@@ -846,7 +846,7 @@ void generate_data(char const* path, torrent_info const& ti)
 			generate_block(piece, i, j, 0x4000);
 			file::iovec_t b = { piece, 0x4000};
 			storage_error error;
-			st->writev(&b, 1, i, j, 0, error);
+			st->writev({b}, i, j, 0, error);
```
arvidn: in fact, I think you don't even need the braces

src/disk_io_thread.cpp
```diff
@@ -616,9 +616,9 @@ namespace libtorrent
 		bool failed = false;
 		for (int i = 1; i <= num_blocks; ++i)
 		{
-			if (i < num_blocks && flushing[i] == flushing[i-1]+1) continue;
-			int ret = pe->storage->get_storage_impl()->writev(iov_start
-				, i - flushing_start
+			if (i < num_blocks && flushing[i] == flushing[i - 1] + 1) continue;
+			int ret = pe->storage->get_storage_impl()->writev({iov_start
+					, size_t(i - flushing_start)}
```
arvidn: if you break the line befor { you don't need the additional indentation level

examples/connection_tester.cpp
aldenml: yes, I was thinking about it right now...I will change it in all places....

test/test_storage.cpp
```diff
@@ -232,58 +232,58 @@ void run_storage_tests(std::shared_ptr<torrent_info> info
 
 	// write piece 1 (in slot 0)
 	file::iovec_t iov = { piece1.data(), half};
-	ret = s->writev(&iov, 1, 0, 0, 0, ec);
+	ret = s->writev({iov}, 0, 0, 0, ec);
```
arvidn: same here, I don't think you need the braces

test/test_transfer.cpp
```diff
@@ -104,10 +103,10 @@ struct test_storage : default_storage
 			return 0;
 		}
 
-		for (int i = 0; i < num_bufs; ++i)
+		for (int i = 0; i < int(bufs.size()); ++i)
```
arvidn: I think you could just do: for (auto const& b : bufs) m_written += int(b.iov_len);
here

aldenml:
do you mind cancel my unnecessary travis jobs

PR #1008, opened 2016-08-18
add a default DHT bootstrap node when it's enabled on session startup

arvidn:


src/session_impl.cpp
```diff
@@ -473,6 +473,15 @@ namespace aux {
 			m_alerts.set_alert_mask(pack.get_int(settings_pack::alert_mask));
 		}
 
+		if (!pack.has_val(settings_pack::enable_dht)
+			|| pack.get_bool(settings_pack::enable_dht) == true)
+		{
+			// if the DHT is enabled on startup, there's currently no way to add
+			// router nodes early enough for them to be used by the DHT bootstrap
+			// add the default libtorrent router
+			add_dht_router(std::pair<std::string, int>("dht.libtorrent.org", 25401));
```
aldenml: what do you think of moving the dht routers to `session_params`?

aldenml: sorry, I didn't see this is for RC_1_1

arvidn: yeah, I intend to do that. I think there may be a way to do it in RC_1_1 without breaking ABI even.. I believe the interaction with the session is via the settings_pack, and not aux::session_settings. That means it's just a map of settings-indices -> values. i.e. it can be extended without altering the ABI

arvidn:
@aldenml does this make sense?

aldenml:
I think this idea (`settings_pack::dht_bootstrap_nodes`) _is_ the solution, with the additional benefit of not need of changes in master

arvidn:
I have no idea why everything is failing now.. truly mysterious

arvidn:
actually, maybe it's the DNS lookup this change implies, before the DHT starts up

aldenml:
Question: If `session_impl::add_dht_router` is async (internally calling `on_dht_router_name_lookup`) how it syncs with the start of the DHT

aldenml:
jeje, I think I was asking me the same thing

arvidn:
there's a counter, m_outstanding_router_lookups, which is incremented on initiating the dns lokup, and decremented when it completes. starting of the DHT is blocked by this > 0 https://github.com/arvidn/libtorrent/blob/master/src/session_impl.cpp#L5553

When the last DNS lookup completes, and the DHT is configured to be started, it's started from there

aldenml:
thanks for the clarification

PR #1010, opened 2016-08-18
avoid exporting integer_to_str

aldenml:
do you think this avoid the problem of "one definition rule"?

### DIFF
--- a/include/libtorrent/bencode.hpp
+++ b/include/libtorrent/bencode.hpp
@@ -88,6 +88,28 @@ namespace libtorrent
 
 	namespace detail
 	{
+		template <typename T, typename Cond = typename std::enable_if<
+			std::is_same<T, char*>::value>::type>
+		char const* integer_to_str(T buf, int size
+			, entry::integer_type val)
+		{
+			int sign = 0;
+			if (val < 0)
+			{
+				sign = 1;
+				val = -val;
+			}
+			buf[--size] = '\0';
+			if (val == 0) buf[--size] = '0';
+			for (; size > sign && val != 0;)
+			{
+				buf[--size] = '0' + char(val % 10);
+				val /= 10;
+			}
+			if (sign) buf[--size] = '-';
+			return buf + size;
+		}
+
 		template <class OutIt>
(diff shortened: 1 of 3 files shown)


include/libtorrent/bencode.hpp
```diff
@@ -88,6 +88,28 @@ namespace libtorrent
 
 	namespace detail
 	{
+		template <typename T, typename Cond = typename std::enable_if<
```
arvidn: is the main purpose here to support arbitrary iterators, and not just char pointers?

arvidn: oh, no.. you just force the template parameter to be a char*.. making this a template seems redundant. couldn't you just make it "inline"?

aldenml: yes, but I didn't know you want it inline, don't you think t's very big? (in case no -O3 is used)

arvidn:
neither version is an ODR violation as far as I can tell. what does this buy you? one symbol less in the export table? it's not obvious that picking up this symbol from the library is worse than being forced to inline the function in the client.

aldenml:
nothing substantial, just a matter of source code layout. If you think there is a disadvantage here (or that you don't like it), feel free to say it or just close it :)

include/libtorrent/bencode.hpp
arvidn: but.. making it a template makes it implicitly inline, and since you're constraining the template like this, being inline is really the only property that's left. The fact that it's inline now is why your patch works. and, yes. I wish more of the bdecode code could be moved into the library (which is why I put this function in there).

arvidn:
the disadvantage is marginally longer build times (since this function has to be parsed and compiled multiple times) and also a code size issue, where if you have a linker that isn't doing a very good job of eliminating duplicate code it may use more space. This is a very simple case of deduplication though, so I wouldn't expect that to be a problem.

It also introduced additional dependencies from other code. If you make a change in this function, you have to rebuild a lot more with it in the header vs. in a .cpp file

aldenml:
Thanks for the details and explanation...details..details :D

PR #1011, opened 2016-08-18
avoid aux::session_settings in storage_interface, plus refactor

aldenml:
this also layout a framework to start removing `aux::session_settings` when possible.

### DIFF
--- a/include/libtorrent/block_cache.hpp
+++ b/include/libtorrent/block_cache.hpp
@@ -67,7 +67,7 @@ namespace libtorrent
 	struct cache_status;
 	struct block_cache_reference;
 	struct counters;
-	namespace aux { struct session_settings; }
+	struct settings_pack;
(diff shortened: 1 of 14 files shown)


arvidn:
hm.. aux::session_settings represents the complete settings, and settings_pack represents a subset of settings. If you start using settings_pack for the latter, it will be a lot less efficient, since every lookup and update will have to traverse a tree (vs. an array index lookup). I think aux::session_settings is the more appropriate representation of settings in a lot of places, especially in the central place in session_impl, but probably also in the disk threads

arvidn:
the type to remove is `libtorrent::session_settings` (not the one under `aux`)

aldenml:
true, in fact, now I remember that I saw this difference in performance doing another refactor. I will see if I can remove it in another way from `storage_interface`, maybe with the flags

PR #1012, opened 2016-08-19
no aux::settings in storage_interface

aldenml:
what do you think about this solution?

### DIFF
--- a/include/libtorrent/storage.hpp
+++ b/include/libtorrent/storage.hpp
@@ -205,17 +205,13 @@ namespace libtorrent
 	//
 	struct TORRENT_EXPORT storage_interface
 	{
-		// hidden
-		storage_interface(): m_settings(0) {}
-
-
 		// This function is called when the storage is to be initialized. The
 		// default storage will create directories and empty files at this point.
 		// If ``allocate_files`` is true, it will also ``ftruncate`` all files to
 		// their target size.
 		//
 		// If an error occurs, ``storage_error`` should be set to reflect it.
-		virtual void initialize(storage_error& ec) = 0;
+		virtual void initialize(int flags, storage_error& ec) = 0;
 
 		// These functions should read and write the data in or to the given
 		// ``piece`` at the given ``offset``. It should read or write
@@ -238,7 +234,7 @@ namespace libtorrent
 		// bytes though. This is useful when reading and writing the last piece
 		// of a file in unbuffered mode.
 		// 
-		// The ``offset`` is aligned to 16 kiB boundaries  *most of the time*, but
+		// The ``offset`` is aligned to 16 kiB boundaries *most of the time*, but
 		// there are rare exceptions when it's not. Specifically if the read
 		// cache is disabled/or full and a peer requests unaligned data. Most
 		// clients request aligned data.
@@ -263,7 +259,7 @@ namespace libtorrent
 		// guaranteed to be the only running function on this storage
 		// when called
 		virtual void set_file_priority(std::vector<std::uint8_t> const& prio
-			, storage_error& ec) = 0;
+			, int flags, storage_error& ec) = 0;
 
 		// This function should move all the files belonging to the storage to
 		// the new save_path. The default storage moves the single file or the
@@ -364,14 +360,8 @@ namespace libtorrent
 		// off again.
 		virtual bool tick() { return false; }
 
-		// access global session_settings
-		aux::session_settings const& settings() const { return *m_settings; }
-
 		// hidden
 		virtual ~storage_interface() {}
-
-		// initialized in disk_io_thread::perform_async_job
-		aux::session_settings* m_settings;
 	};
 
 	// The default implementation of storage_interface. Behaves as a normal
@@ -403,12 +393,12 @@ namespace libtorrent
 #endif
 		virtual bool has_any_file(storage_error& ec) override;
 		virtual void set_file_priority(std::vector<std::uint8_t> const& prio
-			, storage_error& ec) override;
+			, int flags, storage_error& ec) override;
 		virtual void rename_file(int index, std::string const& new_filename
 			, storage_error& ec) override;
 		virtual void release_files(storage_error& ec) override;
 		virtual void delete_files(int options, storage_error& ec) override;
-		virtual void initialize(storage_error& ec) override;
+		virtual void initialize(int flags, storage_error& ec) override;
 		virtual int move_storage(std::string const& save_path, int flags
 			, storage_error& ec) override;
 		virtual bool verify_resume_data(add_torrent_params const& rd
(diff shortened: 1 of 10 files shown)


src/disk_io_thread.cpp
```diff
@@ -127,11 +127,19 @@ namespace libtorrent
...
-		if (coalesce_buffers) ret |= file::coalesce_buffers;
+		if (j != nullptr && !(j->flags & disk_io_job::sequential_access))
+			ret |= file::random_access;
+
+		if (settings.get_bool(settings_pack::coalesce_reads)) ret |= file::coalesce_buffers;
+		if (settings.get_bool(settings_pack::lock_files)) ret |= file::lock_file;
+		if (settings.get_bool(settings_pack::no_atime_storage)) ret |= file::no_atime;
+
+		if (settings.get_int(settings_pack::disk_io_write_mode)
+			== settings_pack::disable_os_cache) ret |= file::no_cache;
```
arvidn: is this flag still around?

include/libtorrent/storage.hpp
```diff
@@ -205,17 +205,13 @@ namespace libtorrent
...
-
-
 		// This function is called when the storage is to be initialized. The
 		// default storage will create directories and empty files at this point.
 		// If ``allocate_files`` is true, it will also ``ftruncate`` all files to
 		// their target size.
 		//
 		// If an error occurs, ``storage_error`` should be set to reflect it.
-		virtual void initialize(storage_error& ec) = 0;
+		virtual void initialize(int flags, storage_error& ec) = 0;
```
arvidn: what are the flags here? These are all file flags?

arvidn:
seems pretty reasonable..

src/disk_io_thread.cpp
aldenml: yes, even in the `test_storage` is used, but not sure if used deep down

include/libtorrent/storage.hpp
aldenml: yes, mostly to pass in `open_file`

aldenml:
good, now waiting for the ci jobs

src/disk_io_thread.cpp
```diff
@@ -2528,8 +2521,13 @@ namespace libtorrent
 		if (rd == nullptr) rd = &tmp;
 
 		std::unique_ptr<std::vector<std::string>> links(j->d.links);
+		bool no_recheck_incomplete_resume
```
arvidn: this bool should be const.. and, I wonder if there's a better way to pass this in. just a bool feels perhaps a bit untyped

src/storage.cpp
```diff
@@ -506,7 +507,7 @@ namespace libtorrent
 		}
 	}
 
-	void default_storage::initialize(storage_error& ec)
+	void default_storage::initialize(int flags, storage_error& ec)
```
arvidn: please make flags const here

src/storage.cpp
```diff
@@ -1217,9 +1217,9 @@ namespace libtorrent
 	}
 
 	file_handle default_storage::open_file(int file, int mode
-		, storage_error& ec) const
+		, int flags, storage_error& ec) const
```
arvidn: and here (and below)

src/storage.cpp
```diff
@@ -1499,37 +1493,33 @@ namespace libtorrent
 	}
 #endif
 
-	int piece_manager::check_no_fastresume(storage_error& ec)
+	int piece_manager::check_no_fastresume(int flags, storage_error& ec)
 	{
-		bool has_files = false;
-		if (!m_storage->settings().get_bool(settings_pack::no_recheck_incomplete_resume))
```
arvidn: doesn't this break the no_recheck_incomplete_resume setting?

test/test_block_cache.cpp
```diff
@@ -47,7 +47,7 @@ using namespace libtorrent;
 
 struct test_storage_impl : storage_interface
 {
-	void initialize(storage_error& ec) override {}
+	void initialize(int flags, storage_error& ec) override {}
```
arvidn: const here too

test/test_block_cache.cpp
```diff
@@ -62,7 +62,7 @@ struct test_storage_impl : storage_interface
 
 	bool has_any_file(storage_error& ec) override { return false; }
 	void set_file_priority(std::vector<std::uint8_t> const& prio
-		, storage_error& ec) override {}
+		, int flags, storage_error& ec) override {}
```
arvidn: I'm surprised there aren't unused-variable warnings here

src/storage.cpp
aldenml: no, because the logic was moved to `piece_manager::check_fastresume` where it's a little more clear and convenient

aldenml:
I know somebody complained about this a while ago
```
test_transfer.cpp(80): warning C4373: 'test_storage::set_file_priority': virtual function overrides 'libtorrent::default_storage::set_file_priority', previous versions of the compiler did not override when parameters only differed by const/volatile qualifiers
C:\projects\libtorrent\include\libtorrent/storage.hpp(395): note: see declaration of 'libtorrent::default_storage::set_file_priority'
```
that's the reason I put `const` in the signature as well

aldenml:
ignore my nonsense comment above

aldenml:
yes, that's the reason, I was looking at the wrong branch

test/test_transfer.cpp
```diff
@@ -78,7 +78,7 @@ struct test_storage : default_storage
 	{}
 
 	void set_file_priority(std::vector<std::uint8_t> const& p
-		, int const flags, storage_error& ec) override {}
+		, int flags, storage_error& ec) override {}
```
arvidn: I see. at some point it would probably make sense to silence that warning. it's just msvc saying that they used to be non-conforming.. but since master requires C++11, it seems reasonable to expect a conforming implementation (within reason, msvc-14.0 isn't all the way there yet)

arvidn:
hm.. the error on travid indicates a pretty serious issue. are you passing more flags through the disk_io_job::flags field now by any chance?

aldenml:
yes, but I checked and they are carefully tested, it's not like they go straight to the open mode, and it was green before (more than one time), can you restart it?

arvidn:
This is not the kind of test that would have false positives, but it's quite likely that chance can cause it to fail if there's breakage. I don't see anywhere in this patch where you create a disk_io_job object, and assign file::flags_t to it. The problem this assert indicates is that a disk_io_job object is either being reused incorrectly or having its flags set incorrectly. Specifically, it's supposed to have the disk_io_job::in_progress whenever it has been started, and when the torrent is being stopped in that case, the disk job has already been started (but not stopped).

aldenml:
I understand your concern, but not the actual problem, not sure how to tackle it atm

arvidn:
yeah, it is a bit mysterious. the disk_io_job is supposed to be freshly allocated and initialized at that point, in this case. It's not clear how flags can be anything other than 0, unless it's an old job that no longer hans its in_progress accounting done correctly.

aldenml:
Ok, I think it's better to follow another route, since this PR is definitively not logically equivalent to the current code and the error can be a bug (new or old). The benefits with this change is not only the "cleanup" in settings, but also the resolution of the `TODO`. I don't mind putting this on hold, but open to study the error

aldenml:
Ok, maybe next time, not happy with the amount of flags either.

PR #1016, opened 2016-08-20
minor cleanup and refactor

aldenml:


src/announce_entry.cpp
```diff
@@ -99,14 +99,14 @@ namespace libtorrent
 		min_announce = min_time();
 	}
 
-	void announce_entry::failed(aux::session_settings const& sett, int retry_interval)
+	void announce_entry::failed(int const tracker_backoff, int const retry_interval)
```
arvidn: would you mind making tracker_backoff be `libtorrent::time_duration` instead of `int`? or at least leave a todo (since I realize this would be a somewhat unrelated improvement over the existing code)

src/announce_entry.cpp
```diff
@@ -99,14 +99,14 @@ namespace libtorrent
...
 	{
 		++fails;
 		// the exponential back-off ends up being:
 		// 7, 15, 27, 45, 95, 127, 165, ... seconds
 		// with the default tracker_backoff of 250
 		int delay = (std::min)(tracker_retry_delay_min + int(fails) * int(fails)
```
arvidn: then `delay` would also naturally be a `time_duration` I believe

src/announce_entry.cpp
aldenml: ok

src/torrent_info.cpp
```diff
@@ -80,7 +80,7 @@ namespace libtorrent
 	} // anonymous namespace
 
 	// fixes invalid UTF-8 sequences
-	TORRENT_EXTRA_EXPORT bool verify_encoding(std::string& target)
+	bool verify_encoding(std::string& target)
```
arvidn: I believe this function has unit tests (if not, it should), which means the test need to be able to reach it

aldenml: yes, this is the definition, in the declaration it is exported

src/torrent_peer.cpp
```diff
@@ -82,7 +82,7 @@ namespace libtorrent
 		if (e1.address() == e2.address())
 		{
 			if (e1.port() > e2.port())
-				swap(e1, e2);
```
arvidn: the convention when using swap() is actually to not fully qualify it, to allow argument dependent lookup. You may notice that there's a `using std::swap` above, which pulls in the standard overload if there is no better one.

src/torrent_peer.cpp
```diff
@@ -107,8 +107,8 @@ namespace libtorrent
 			if (e1 > e2) swap(e1, e2);
 			address_v6::bytes_type b1 = e1.address().to_v6().to_bytes();
 			address_v6::bytes_type b2 = e2.address().to_v6().to_bytes();
-			int mask = memcmp(&b1[0], &b2[0], 4) ? 0
-				: memcmp(&b1[0], &b2[0], 6) ? 1 : 2;
+			int mask = std::memcmp(&b1[0], &b2[0], 4) ? 0
```
arvidn: would you mind making `mask` const here? also, instead of `&b1[0]`, would it make more sense to use `b1.data()`?

test/test_primitives.cpp
```diff
@@ -71,11 +71,10 @@ TORRENT_TEST(primitives)
 	// on failing announces
 	announce_entry ae("dummy");
 	int last = 0;
-	aux::session_settings sett;
-	sett.set_int(settings_pack::tracker_backoff, 250);
+	int const tracker_backoff = 250;
```
arvidn: if you wouldn't mind, this could be `auto const tracker_backoff = seconds(250);`

src/torrent_info.cpp
arvidn: ah, I see

arvidn:
looks good (apart from swap really)

aldenml:
good, let me address your comments

src/announce_entry.cpp
```diff
@@ -99,14 +99,15 @@ namespace libtorrent
...
 
-	void announce_entry::failed(aux::session_settings const& sett, int retry_interval)
+	void announce_entry::failed(time_duration const tracker_backoff, int const retry_interval)
 	{
 		++fails;
 		// the exponential back-off ends up being:
 		// 7, 15, 27, 45, 95, 127, 165, ... seconds
 		// with the default tracker_backoff of 250
+		int const tracker_backoff_seconds = total_seconds(tracker_backoff);
 		int delay = (std::min)(tracker_retry_delay_min + int(fails) * int(fails)
```
arvidn: did it turn out to be messy to maintain the `time_duration` type here? (rather than casting to int and then back again)

aldenml: Very 😀

PR #1018, opened 2016-08-21
more use of span and std::shared_ptr

aldenml:


arvidn:
lgtm

arvidn:
did you see the compilation error?

aldenml:
looking at it ...

PR #1023, opened 2016-08-21
using span in torrent_handle and no optional in torrent_info

aldenml:
Some context:
1- For what I see in this case, the `boost::optional` was essentially a comparison to `0` since the internal field was a raw `time_t`. Not using it is very swig friendly and I think it is acceptable that upper app layers (meaning UI) just filter out the value.
2- The internal `m_info_section` field was treated essentially like a `unique_ptr`, I think this simplify a little bit the logic, but needs careful review.

### DIFF
--- a/include/libtorrent/torrent_handle.hpp
+++ b/include/libtorrent/torrent_handle.hpp
@@ -51,6 +51,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "libtorrent/address.hpp"
 #include "libtorrent/socket.hpp" // tcp::endpoint
+#include "libtorrent/span.hpp"
 
 namespace libtorrent
 {
@@ -521,7 +522,13 @@ namespace libtorrent
 		// if the metadata is successfully set on the torrent, and false
 		// otherwise. If the torrent already has metadata, this function will not
 		// affect the torrent, and false will be returned.
-		bool set_metadata(char const* metadata, int size) const;
+		bool set_metadata(span<char const> metadata) const;
+
+#ifndef TORRENT_NO_DEPRECATE
+		TORRENT_DEPRECATED
+		bool set_metadata(char const* metadata, int size) const
+		{ return set_metadata({metadata, size_t(size)}); }
+#endif
 
 		// Returns true if this handle refers to a valid torrent and false if it
 		// hasn't been initialized or if the torrent it refers to has been
--- a/include/libtorrent/torrent_info.hpp
+++ b/include/libtorrent/torrent_info.hpp
@@ -36,12 +36,10 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <string>
 #include <vector>
 #include <map>
+#include <memory>
 
 #include "libtorrent/aux_/disable_warnings_push.hpp"
-
-#include <boost/optional.hpp>
 #include <boost/shared_array.hpp>
-
 #include "libtorrent/aux_/disable_warnings_pop.hpp"
 
 #include "libtorrent/config.hpp"
@@ -63,8 +61,6 @@ namespace libtorrent
 	struct announce_entry;
 	struct lazy_entry;
 
-	namespace aux { struct session_settings; }
-
 	// internal, exposed for the unit test
 	TORRENT_EXTRA_EXPORT void sanitize_append_path_element(std::string& path
 		, string_view element);
@@ -80,7 +76,7 @@ namespace libtorrent
 		// http seed spec. by John Hoffman
 		enum type_t { url_seed, http_seed };
 
-		typedef std::vector<std::pair<std::string, std::string> > headers_t;
+		typedef std::vector<std::pair<std::string, std::string>> headers_t;
 
 		web_seed_entry(std::string const& url_, type_t type_
 			, std::string const& auth_ = std::string()
@@ -399,7 +395,7 @@ namespace libtorrent
 		// returns true if this torrent_info object has a torrent loaded.
 		// This is primarily used to determine if a magnet link has had its
 		// metadata resolved yet or not.
-		bool is_valid() const { return (m_files.is_valid()) != 0; }
+		bool is_valid() const { return m_files.is_valid(); }
 
 		// returns true if this torrent is private. i.e., it should not be
 		// distributed on the trackerless network (the kademlia DHT).
@@ -469,7 +465,8 @@ namespace libtorrent
 		// 
 		// .. _`posix time`: http://www.opengroup.org/onlinepubs/009695399/functions/time.html
 		const std::string& name() const { return m_files.name(); }
-		boost::optional<time_t> creation_date() const;
+		time_t creation_date() const
+		{ return m_creation_date; }
 		const std::string& creator() const
 		{ return m_created_by; }
 		const std::string& comment() const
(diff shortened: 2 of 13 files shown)


include/libtorrent/torrent_handle.hpp
```diff
@@ -521,7 +522,7 @@ namespace libtorrent
 		// if the metadata is successfully set on the torrent, and false
 		// otherwise. If the torrent already has metadata, this function will not
 		// affect the torrent, and false will be returned.
-		bool set_metadata(char const* metadata, int size) const;
```
arvidn: would you mind preserving the existing signature, under `#ifndef TORRENT_NO_DEPRECATE`, and with the `TORRENT_DEPRECATED` attribute?

aldenml: sure

include/libtorrent/torrent_info.hpp
```diff
@@ -504,8 +497,8 @@ namespace libtorrent
 		// ``metadata()`` returns a the raw info section of the torrent file. The size
 		// of the metadata is returned by ``metadata_size()``.
 		int metadata_size() const { return m_info_section_size; }
-		boost::shared_array<char> metadata() const
-		{ return m_info_section; }
```
arvidn: It's not obvious that this is correct. previously shared ownership of this buffer was returned, but with this patch only a (non-owning) reference is returned

include/libtorrent/torrent_info.hpp
```diff
@@ -580,7 +573,7 @@ namespace libtorrent
 		// this is a copy of the info section from the torrent.
 		// it use maintained in this flat format in order to
 		// make it available through the metadata extension
-		boost::shared_array<char> m_info_section;
```
arvidn: this prevents the client from keeping this buffer alive. Given that the client and libtorrent runs in separate threads, it kind of makes sense to actually share ownership. Otherwise the client could have the rug pulled from under its feet at any moment

include/libtorrent/torrent_info.hpp
```diff
@@ -607,7 +600,7 @@ namespace libtorrent
 		sha1_hash m_info_hash;
 
 		// the number of bytes in m_info_section
-		std::uint32_t m_info_section_size = 0;
```
arvidn: is this important?

aldenml: no :)

src/torrent_info.cpp
```diff
@@ -967,8 +966,6 @@ namespace libtorrent
 
 	void torrent_info::unload()
 	{
-		TORRENT_ASSERT(m_info_section.unique());
```
arvidn: I think this should probably be a shared_ptr still. you could make it a `std::shared_ptr<char[]>` still though

test/test_torrent_info.cpp
```diff
@@ -897,13 +897,7 @@ TORRENT_TEST(copy)
 	}
 
 	// copy the torrent_info object
-	boost::shared_ptr<torrent_info> b(boost::make_shared<torrent_info>(*a));
-
-	// clear out the  buffer for a, just to make sure b doesn't have any
-	// references into it by mistake
-	int s = a->metadata_size();
-	memset(a->metadata().get(), 0, s);
```
arvidn: this test would still be relevant when allowing shared ownership I believe

arvidn:
looks good apart from the metadata buffer being a `unique_ptr`. I don't think that works well.

src/torrent_info.cpp
aldenml: `(std|boost)::shared_ptr` it's not specialized for arrays, it works but you need to provide a custom deleter (easy) and there is no operator `[]`

aldenml:
What about `TORRENT_ASSERT(m_info_section.unique());` in `torrent_info::unload`? Is that OK?

arvidn:
hm.. I suppose that if the client is in fact sharing ownership, that assert would fail. It's possible I added that to make sure no other part of libtorrent was holding ownership over it (like, the metadata extension)

aldenml:
Do I remove it? now that the metadata is using a `span`?

include/libtorrent/torrent_info.hpp
```diff
@@ -580,7 +573,7 @@ namespace libtorrent
 		// this is a copy of the info section from the torrent.
 		// it use maintained in this flat format in order to
 		// make it available through the metadata extension
-		boost::shared_array<char> m_info_section;
+		std::unique_ptr<char[]> m_info_section;
```
arvidn: this is where the metadata is owned. I'm saying that I believe this has to support shared ownership, so that the client asking for it can know that it will be valid still, immediately after it's returned. If the returned buffer is tied to the lifetime of this torrent object, especially if the torrent object is "evicted" or unloaded, it will become a dangling pointer.

arvidn: I don't think the use of `span<>` in the interface is relevant, other than returning a span is problematic as it does not transfer ownership

aldenml: I didn't ask correctly, I mean now that `ut_metadata` is using `span` (not the interface), but I can infer the response: "keep the assert"

include/libtorrent/torrent_info.hpp
```diff
@@ -497,8 +501,8 @@ namespace libtorrent
 		// ``metadata()`` returns a the raw info section of the torrent file. The size
 		// of the metadata is returned by ``metadata_size()``.
 		int metadata_size() const { return m_info_section_size; }
-		span<char const> metadata() const
-		{ return span<char const>(m_info_section.get(), size_t(m_info_section_size)); }
+		boost::shared_array<char> metadata() const
```
arvidn: I guess `std::shared_ptr<T[]>` wasn't added until C++17...

include/libtorrent/torrent_info.hpp
```diff
@@ -501,7 +497,7 @@ namespace libtorrent
 		// ``metadata()`` returns a the raw info section of the torrent file. The size
 		// of the metadata is returned by ``metadata_size()``.
 		int metadata_size() const { return m_info_section_size; }
-		boost::shared_array<char> metadata() const
+		std::shared_ptr<char> metadata() const
```
arvidn: I don't think this works. `shared_ptr<>` won't support arrays until C++17. This means a pointer to a single `char`

aldenml: ok, agree that `boost::shared_array` is the best tool for now, I will revert all the changes related to shared_array, I want to easy the review and merge

aldenml:
I rebased because in this particular case is easier to ignore the shared_array related commits

src/ut_metadata.cpp
```diff
@@ -53,8 +53,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/performance_counters.hpp" // for counters
 #include "libtorrent/aux_/time.hpp"
 
-using namespace std::placeholders;
```
arvidn: there's no use of std::bind() in here? this is used to pull in _1, _2 etc.

aldenml: no `bind`, no` _1` or `_2`

PR #1024, opened 2016-08-21
no aux::settings in storage_interface

aldenml:
I think this PR is more easy to read and safer. I didn't mix the general "flags" with the "open flags" and in fact, I discovered serious bugs mixing flags in my previous PR.

### DIFF
--- a/include/libtorrent/storage.hpp
+++ b/include/libtorrent/storage.hpp
@@ -205,17 +205,13 @@ namespace libtorrent
 	//
 	struct TORRENT_EXPORT storage_interface
 	{
-		// hidden
-		storage_interface(): m_settings(0) {}
-
-
 		// This function is called when the storage is to be initialized. The
 		// default storage will create directories and empty files at this point.
 		// If ``allocate_files`` is true, it will also ``ftruncate`` all files to
 		// their target size.
 		//
 		// If an error occurs, ``storage_error`` should be set to reflect it.
-		virtual void initialize(storage_error& ec) = 0;
+		virtual void initialize(int open_flags, storage_error& ec) = 0;
 
 		// These functions should read and write the data in or to the given
 		// ``piece`` at the given ``offset``. It should read or write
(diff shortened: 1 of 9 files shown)


aldenml:
very quiet here :) (I don't mind closing it if you don't like it)

include/libtorrent/storage.hpp
```diff
@@ -247,9 +243,9 @@ namespace libtorrent
 		// error. If there's an error, the ``storage_error`` must be filled out
 		// to represent the error that occurred.
 		virtual int readv(span<file::iovec_t const> bufs
-			, int piece, int offset, int flags, storage_error& ec) = 0;
+			, int piece, int offset, int flags, int open_flags, storage_error& ec) = 0;
```
arvidn: the thing that worries me a bit here is how all the ``int`` arguments aren't typed. It makes it easy to get the order wrong. obviously this is not a new problem, it's just compounding.

include/libtorrent/storage.hpp
```diff
@@ -594,11 +584,14 @@ namespace libtorrent
 		// when wrong in the disk access
 		int check_fastresume(add_torrent_params const& rd
 			, std::vector<std::string> const& links
+			, aux::session_settings const& settings
```
arvidn: how come the settings are necessary here?

src/storage.cpp
```diff
@@ -1498,13 +1490,13 @@ namespace libtorrent
 	}
 #endif
 
-	int piece_manager::check_no_fastresume(storage_error& ec)
+	int piece_manager::check_no_fastresume(aux::session_settings const& settings
+		, int const open_flags, storage_error& ec)
 	{
-		bool has_files = false;
-		if (!m_storage->settings().get_bool(settings_pack::no_recheck_incomplete_resume))
+		if (!settings.get_bool(settings_pack::no_recheck_incomplete_resume))
```
arvidn: ah, for the ``no_recheck_incomplete_resume``

arvidn:
looks reasonable. I'm not happy with all those ``int``s long term really, but it's possible the ``storage_interface`` may get collapsed anyway.

aldenml:
Ok, if `storage_interface` is schedule to a heavy refactor, I prefer not doing this change, this it's just a marginal improvement, at the expense of the int flags explosion (with a lot of API downsides). In the swig side, I'm able to filter out that field and method without problems. I'm closing this, but keeping the branch in my fork.

PR #1025, opened 2016-08-21
fix build

arvidn:


aldenml:
Cancel all my CI jobs if you need to

PR #1028, opened 2016-08-22
adding support for a more configurable sha512 hasher

aldenml:
This PR only adds a `hasher512` that it's very similar to the sha1 based `hasher`. For now, it helps only to test the correctness of the current sha512 implementation used in `ed25519`, since the file `sha512.cpp` is a copy from that.

If this PR is merged, the natural next step is remove the old sha512 inside `ed25519` to use the new `hasher512` which takes advantage of third party libraries if configured.

The ultimate goal is bring up to date the `ed25519` with the latest `ref10` implementation, with unit tests that easy the path for future upgrades.


arvidn:
what's your opinion on submodules? I'm increasingly tempted to pull things in by reference

include/libtorrent/hasher512.hpp
```diff
@@ -0,0 +1,123 @@
...
+	// For more info, see ``src/sha512.cpp``.
+	class TORRENT_EXPORT hasher512
+	{
+	public:
+
+		hasher512();
+
+		// this is the same as default constructing followed by a call to
+		// ``update(data, len)``.
+		hasher512(char const* data, int len);
```
arvidn: a part of me feels like this overload is unnecessary given the span<> overload. you just need one level of { and }.

include/libtorrent/hasher512.hpp
```diff
@@ -0,0 +1,123 @@
...
+		hasher512(span<char const> data);
```
arvidn: should this constructor be explicit?

include/libtorrent/sha512.hpp
```diff
@@ -0,0 +1,23 @@
+#ifndef TORRENT_SHA512_HPP_INCLUDED
+#define TORRENT_SHA512_HPP_INCLUDED
+
+#include "libtorrent/config.hpp"
+#include <cstdint>
+
+namespace libtorrent
+{
+
+	struct TORRENT_EXTRA_EXPORT sha512_ctx
```
arvidn: I don't think export will actually do anything here..

src/hasher512.cpp
```diff
@@ -0,0 +1,235 @@
...
+#endif
+		}
+		return *this;
+	}
+#else
+	hasher512::hasher512(hasher512 const&) = default;
+	hasher512& hasher512::operator=(hasher512 const&) = default;
+#endif
+
+	hasher512& hasher512::update(char const* data, int len)
```
arvidn: this overload too. the reason I kept it around in `hasher` is because it's used in a bunch of places. I think

src/hasher512.cpp
```diff
@@ -0,0 +1,235 @@
...
+		}
+#elif defined TORRENT_USE_LIBCRYPTO
+		SHA512_Update(&m_context, reinterpret_cast<unsigned char const*>(data.data()), data.size());
+#else
+		SHA512_update(&m_context, reinterpret_cast<unsigned char const*>(data.data()), data.size());
+#endif
+		return *this;
+	}
+
+	void hasher512::final(span<char> digest)
```
arvidn: would it be inconvenient to templatize the `sha1_hash` and use that as the return value here? This API is a bit questionable I'd say. If the length of the buffer is determined at run-time, I imagine it should at least be used in the memcpy() as well. But I think it isn't really a runtime variable

src/sha512.cpp
```diff
@@ -0,0 +1,283 @@
+#include <cstdio>
```
arvidn: is this better than the implementation from ed25519?

test/test_hasher512.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include "libtorrent/hasher512.hpp"
+#include "libtorrent/hex.hpp"
+
+#include "test.hpp"
+
+using namespace libtorrent;
+
+namespace
+{
+	void test_vector(std::string s, std::string output, int n = 1)
```
arvidn: `n` could be const here

include/libtorrent/hasher512.hpp
aldenml: ok, to remove it

include/libtorrent/hasher512.hpp
aldenml: I don't know, not a heavy user of that keyword :)

include/libtorrent/sha512.hpp
aldenml: k, I will remove it from `sha1` too if possible.

src/hasher512.cpp
aldenml: well, I don't like it either, I just did't figure out the best alternative. Let me see what can I do with `sha1_hash`

src/sha512.cpp
aldenml: it's exactly the same, only with a surrounding `namespace libtorrent`, the reason I made this is to minimize changes and easy the review, but if this get accepted, the very next step is remove the old file

test/test_hasher512.cpp
aldenml: k

aldenml:
well, submodules in the past was a very painful experience for us (in frostwire), we tried git native, manual submodules (with sync scripts) and now we are using a mono-repo setup (everything together). I just try to minimize dependencies as much as I can, and that's something a really like from `libtorrent`.

include/libtorrent/hasher512.hpp
arvidn: for single-parameter constructors, unless they are annotated with `explict` they allow implicit conversion. i.e. you can pass in any `span<>` to a parameter that takes a `hasher512` without mentioning `hasher512`, and a temporary one will be created for you. It's considered a bit dangerous to allow that, unless explicitly intended (like the `span<>` constructors for instance, which are intended to be used for implicit conversions)

aldenml: Thanks!

arvidn:
I'm envisioning the ed25519 code for instance, it could be a submodule. It could still be built the same way, integrated with the libtorrent build system, it would just make it easier to upgrade. As the `bdecode` function and class is being used by `bootstrap-dht` now, I've been thinking about putting that in a separate repo as well, to make them both share the implementation.

I would still imagine tarfiles to be built from the whole package, with all submodules.

aldenml:
I want to redo the last commit, the naming and new files are not the best

src/sha512.cpp
aldenml: I will remove the extra file in my next commit

src/hasher512.cpp
```diff
@@ -195,10 +183,11 @@ namespace libtorrent
 		}
 		TORRENT_ASSERT(size == digest.size());
 #elif defined TORRENT_USE_LIBCRYPTO
-		SHA512_Final(reinterpret_cast<unsigned char*>(digest.begin()), &m_context);
+		SHA512_Final(digest.begin(), &m_context);
```
arvidn: I think `data()` is a bit better than `begin()`, since in principle, `begin()` could return an iterator

aldenml: k

ed25519/src/sign.cpp
```diff
@@ -2,32 +2,31 @@
...
-    sha512_init(&hash);
-    sha512_update(&hash, signature, 32);
-    sha512_update(&hash, public_key, 32);
-    sha512_update(&hash, message, message_len);
-    sha512_final(&hash, hram);
+    hash.reset();
+    hash.update({reinterpret_cast<char const*>(signature), 32});
+    hash.update({reinterpret_cast<char const*>(public_key), 32});
+    hash.update({reinterpret_cast<char const*>(message), message_len});
+    std::memcpy(hram, hash.final().data(), 64);
```
gubatron: perhaps you want to add a method to the hasher interface that wraps this `memcpy`, just for uniformity? I imagine it'd look like this

`hash.copy_final(hram, 64)`

then if you decide there's something better than `std::memcpy` in the future also easier for refactoring

ed25519/src/verify.cpp
```diff
@@ -62,11 +64,10 @@ int ed25519_verify(const unsigned char *signature, const unsigned char *message,
         return 0;
     }
 
-    sha512_init(&hash);
-    sha512_update(&hash, signature, 32);
-    sha512_update(&hash, public_key, 32);
-    sha512_update(&hash, message, message_len);
-    sha512_final(&hash, h);
+    hash.update({reinterpret_cast<char const*>(signature), 32});
```
gubatron: no `hash.reset()` here?

ed25519/src/verify.cpp
```diff
@@ -62,11 +64,10 @@ int ed25519_verify(const unsigned char *signature, const unsigned char *message,
         return 0;
     }
```
gubatron: no need for `hash.reset()` like on the other file here?

ed25519/src/sign.cpp
aldenml: a similar API was rejected by @arvidn in the first commit, it's this way here because this code was working with a more primitive API, but it should be modernized at some point in the future, it would be nice to avoid the double memcpy btw

ed25519/src/verify.cpp
aldenml: no need

ed25519/src/sign.cpp
gubatron: ha, if it was up to me, I'd have a method:

`hash.sign(signature, public_key, message, message_len, hram);`

#DRY

😆

aldenml: looking at it, I think it's possible to remove this double `memcpy`, but I will wait for next review first and then change it

ed25519/src/keypair.cpp
```diff
@@ -2,14 +2,17 @@
...
 
+using namespace libtorrent;
 
 void ed25519_create_keypair(unsigned char *public_key, unsigned char *private_key, const unsigned char *seed) {
     ge_p3 A;
 
-    sha512(seed, 32, private_key);
+    hasher512 hash;
+    hash.update({reinterpret_cast<char const*>(seed), 32});
+    std::memcmp(private_key, hash.final().data(), 64);
```
aldenml: this will fixed in my next commit and I don't understand how it does not fail in the unit tests

Jamfile
```diff
@@ -341,15 +341,21 @@ rule building ( properties * )
...
+			result += <source>src/sha512.cpp ;
+		}
 	}
 
 	if ( <toolset>darwin in $(properties)
 		|| <toolset>gcc in $(properties)
-		|| <toolset>clang in $(propertoes)
-		|| <toolset>clang-darwin in $(propertoes) )
+		|| <toolset>clang in $(properties)
+		|| <toolset>clang-darwin in $(properties) )
```
arvidn: heh, nice typo :)

Jamfile
```diff
@@ -766,6 +770,7 @@ lib torrent
 
 	<dht>on:<source>src/kademlia/$(KADEMLIA_SOURCES).cpp
 	<dht>on:<source>ed25519/src/$(ED25519_SOURCES).cpp
+	<dht>on:<source>src/hasher512.cpp
```
arvidn: I wonder if this extra build logic is worth it. I assume the code in that .cpp file could just as well have ifdefs around it for the DHT being enabled

Jamfile
aldenml: and this `<crypo>` one too :)

Jamfile
aldenml: not sure, there are a couple of details here, 1- clang complains about an empty object, 2- `hasher512` is only relevant if `DHT` is enable, but it's not exactly part of it. You decide.

aldenml: empty object means empty object file `.o` during the compilation

include/libtorrent/sha1_hash.hpp
```diff
@@ -49,66 +49,79 @@ POSSIBILITY OF SUCH DAMAGE.
...
+
+		// returns an all-F digest. i.e. the maximum value
+		// representable by an N bit number (N/8 bytes). This is
 		// a static member function.
-		static sha1_hash max()
+		static digest32 max()
 		{
-			sha1_hash ret;
+			digest32 ret;
 			std::memset(ret.m_number, 0xff, size());
```
arvidn: is the default constructor not clearing it?

include/libtorrent/sha1_hash.hpp
```diff
@@ -49,66 +49,79 @@ POSSIBILITY OF SUCH DAMAGE.
...
 	// sequence. It implements a number of convenience functions, such
 	// as bit operations, comparison operators etc.
 	//
-	// In libtorrent it is primarily used to hold info-hashes, piece-hashes,
-	// peer IDs, node IDs etc.
-	class TORRENT_EXPORT sha1_hash
+	// This data structure is 32 bits aligned, like it's the case for
+	// each SHA-N specification.
+	template <int N>
+	class digest32
```
arvidn: should this be in a different header you think?

include/libtorrent/sha1_hash.hpp
aldenml: this is setting it to `0xff`, I'm wrong here?

include/libtorrent/sha1_hash.hpp
aldenml: at first I did it in another header, but I wanted to make the PR easy to read. This way is better to read how the `sha1_hash` was converted to a template

include/libtorrent/sha512.hpp
```diff
@@ -0,0 +1,23 @@
...
+{
+
+	struct sha512_ctx
+	{
+		std::uint64_t length, state[8];
+		size_t curlen;
+		std::uint8_t buf[128];
+	};
+
+	TORRENT_EXTRA_EXPORT int SHA512_init(sha512_ctx* context);
```
arvidn: I forget why I did this for SHA1 originally. I think it may have been to mimic libcrypto to make it easy to switch between them. At this point I can't see a good reason to introduce this "intermediate" API, under the ``hasher`` class.

can you think of a reason?

src/sha512.cpp
```diff
@@ -12,13 +17,17 @@
  * Tom St Denis, tomstdenis@gmail.com, http://libtom.org
  */
 
-#include "fixedint.h"
-#include "sha512.h"
+typedef std::uint64_t u64;
+typedef std::int64_t i64;
+typedef std::int32_t i32;
```
arvidn: would you mind using the ``using`` keyword instead? I think it's  more intuitive. i.e. ``using u64 = std::uint64_t;``

src/sha512.cpp
```diff
@@ -222,8 +231,8 @@ int sha512_update (sha512_context * md, const unsigned char *in, size_t inlen)
    @param out [out] The destination of the hash (64 bytes)
    @return 0 if successful
 */
-   int sha512_final(sha512_context * md, unsigned char *out)
-   {
+int SHA512_final(std::uint8_t* out, sha512_ctx* md)
```
arvidn: I suppose it keeps things simple for now, to leave these functions like this

include/libtorrent/sha512.hpp
aldenml: The original code, both from `sha1` and `sha512` already provide this interface, it was just slightly accommodated to mimic libcrypto, do you have any particular idea here?

test/test_ed25519.cpp
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	void test_vector(std::string seed, std::string pub, std::string signature, std::string message)
+	{
+		unsigned char s[32];
+		unsigned char sk[64];
+		unsigned char pk[32];
+		unsigned char sig[64];
+		int msg_size = int(message.size()) / 2;
+		std::unique_ptr<unsigned char[]> msg(new unsigned char[msg_size]);
```
arvidn: a std::vector<> may seem more appropriate here..

src/sha512.cpp
aldenml: k

test/test_ed25519.cpp
aldenml: k

arvidn:
I assume the main purpose of this patch is to be able to use an efficient sha512 implementation (i.e. libcrypto) right?

aldenml:
as a first step, but a second one is update the `ed25519` implementation

aldenml:
Not clear to my what do you want with ``<dht>on:<source>src/hasher512.cpp`` and the additional header for `digest32`

arvidn:
This is a safety precaution to make sure tests aren't accidentally defined incorrectly. If you have a file with no tests, you have to add an empty test.

https://travis-ci.org/arvidn/libtorrent/jobs/154930334#L1187

include/libtorrent/sha512.hpp
arvidn: long term I imagine these implementations should just be the member functions of the respective ``hasher`` class

include/libtorrent/sha1_hash.hpp
arvidn: makes sense

include/libtorrent/sha1_hash.hpp
arvidn: sorry. I mixed up min() and max()

Jamfile
gubatron: tumblr_ocfqta2bb61u8pko1o1_1280
proper toes!

Jamfile
arvidn: I see, that makes sense

include/libtorrent/hasher512.hpp
```diff
@@ -0,0 +1,124 @@
...
+	//
+	// The constructor that takes a ``char const*`` and an integer will construct the
+	// sha1 context and feed it the data passed in.
+	//
+	// If you want to reuse the hasher object once you have created a hash, you have to
+	// call ``reset()`` to reinitialize it.
+	//
+	// The built-in software version of the sha512-algorithm is from LibTomCrypt
+	// For more info, see ``src/sha512.cpp``.
+	class TORRENT_EXPORT hasher512
```
arvidn: @aldenml is there a good reason for this being public? afaict, it's only used in the ed25519 implementation. I wonder if it would make sense to make this private and move it into the `ed25519` source directory (along with `sha512.cpp`/`sha512.hpp`)

aldenml: I remember that I left the `TORRENT_EXPORT` (and left it public as `hasher`) because I wanted to have them available outside `libtorrent`, they are very convenient to use, but it turn out I never actually use them. So I guess there is no a good reason, and maybe `hasher` should be private too.

PR #1029, opened 2016-08-25
fix inconsistant prefix calulcation

ssiloti:
The special case for calculating the prefix in the last bucket was being
applied to the condidate node but not the existing nodes in the bucket.

### DIFF
--- a/src/kademlia/node_id.cpp
+++ b/src/kademlia/node_id.cpp
@@ -198,10 +198,10 @@ node_id generate_id(address const& ip)
 	return generate_id_impl(ip, random(0xffffffff));
 }
 
-bool matching_prefix(node_entry const& n, int mask, int prefix, int bucket_index)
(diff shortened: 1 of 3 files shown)


src/kademlia/node_id.cpp
```diff
@@ -198,10 +198,10 @@ node_id generate_id(address const& ip)
 	return generate_id_impl(ip, random(0xffffffff));
 }
 
-bool matching_prefix(node_entry const& n, int mask, int prefix, int bucket_index)
+bool matching_prefix(node_entry const& n, int mask, int prefix, int offset)
```
arvidn: would you mind adding a test case for this function that covers this case?

arvidn:
great catch! do you think this applies to RC_1_1 as well?

src/kademlia/node_id.cpp
ssiloti: I don't see how I could test `matching_prefix`, the special case is handled by the caller.

PR #1034, opened 2016-08-26
added string_ends_with to replace boost

aldenml:


src/string_util.cpp
```diff
@@ -146,6 +156,11 @@ namespace libtorrent
 			*begin++ = printable[random(sizeof(printable) - 2)];
 	}
 
+	bool string_ends_with(std::string const& s1, std::string const& s2)
```
arvidn: what do you think about using string_ref here instead? that way the ".i2p" string literal would have to be heap allocated into a std::string

src/string_util.cpp
```diff
@@ -113,6 +113,11 @@ namespace libtorrent
 
 	bool string_begins_no_case(char const* s1, char const* s2)
 	{
+		TORRENT_ASSERT(s1 != nullptr);
+		TORRENT_ASSERT(s2 != nullptr);
+
+		if (s1 == nullptr || s2 == nullptr) return false;
```
arvidn: this shouldn't happen, right? is it better to say false (and continue running with corrupt data) than to crash?

src/torrent.cpp
```diff
@@ -3317,11 +3314,11 @@ namespace libtorrent
 				continue;
 
 #if TORRENT_USE_I2P
-			if (r.i2pconn && boost::algorithm::ends_with(i->hostname, ".i2p"))
+			if (r.i2pconn && string_ends_with(i->hostname, ".i2p"))
 			{
-				// this is an i2p name, we need to use the sam connection
+				// this is an i2p name, we need to use the same connection
```
arvidn: this should actually be sam, or maybe better "SAM" (to make it more obvious that it's an acronym)

test/test_string.cpp
```diff
@@ -106,9 +106,21 @@ TORRENT_TEST(string_equal_no_case)
 	TEST_CHECK(string_equal_no_case("foobar", "foobar"));
 	TEST_CHECK(!string_equal_no_case("foobar", "foobar "));
 	TEST_CHECK(!string_equal_no_case("foobar", "F00"));
+	TEST_CHECK(!string_equal_no_case("foobar", "foo"));
+	TEST_CHECK(!string_equal_no_case("foo", "foobar"));
 
 	TEST_CHECK(string_begins_no_case("foobar", "FoobAR --"));
 	TEST_CHECK(!string_begins_no_case("foobar", "F00"));
+	TEST_CHECK(!string_begins_no_case("foobar", "foo"));
+	TEST_CHECK(string_begins_no_case("foo", "foobar"));
```
arvidn: I think it would be clearer if you would group the positive test cases together, and the negative test cases together

arvidn:
seems reasonable

src/string_util.cpp
aldenml: agree

PR #1035, opened 2016-08-26
better integration of ed25519

aldenml:
This improve the API, code, documentation and unit tests related to `ed25519`


aldenml:
I decided not to use the `types` from the DHT, because that create a cross reference of components, this could be seen as the "c++11" portable version of `ed25519`.

arvidn:
I thought git/github would show you the rename of the file, without deleting one and adding the other..

arvidn:
one concern I have with adapting and embedding this deeper into libtorrent is that it becomes my responsibility to maintain it. There are some benefits to being able to easily update 3rd party code

aldenml:
It does, but there is a threshold in the diff, I actually removed 6 code files and created one.

I understand your concern, but I think this is a particular case when it's relatively safe to do that, let me explain my reasons:
1- The code is so specially crafted, originally from `ref10`, that it's almost impossible to copy as is, or reference.
2- The code was copied in 2013, and the curve part, gen keys, sign and verify are the same today.
3- Every project I saw, libgcrypt, openssl, libsodium and boringssl, all do the same, copy and heavy adapt to the project.
4- Even the code from https://github.com/orlp/ed25519 is just that.

Btw, just for safety, I almost not changed the critical ones, `fe.cpp`, `ge.cpp` and `sc.cpp`.

You decide :)

aldenml:
as a note, I added support for openssl, better handled `BOOST_NO_EXCEPTIONS` and added a portable option in `ed25519_create_seed`

aldenml:
I need some help with the ref-doc part generation tool

Makefile.am
```diff
@@ -95,7 +95,6 @@ DOCS_PAGES = \
   docs/reference-Core.html        \
   docs/reference-Create_Torrents.html \
   docs/reference-Custom_Storage.html \
-  docs/reference-ed25519.html     \
```
arvidn: did you mean to delete this?

docs/makefile
```diff
@@ -22,7 +22,6 @@ REFERENCE_TARGETS = \
 	reference-Filter \
 	reference-Settings \
 	reference-Bdecoding \
-	reference-ed25519 \
```
arvidn: did you mean to delete this?

include/libtorrent/ed25519.hpp
```diff
@@ -1,34 +0,0 @@
-#ifndef ED25519_HPP
```
arvidn: ah, deleting this file is what I believe is causing the documentation generation error. these functions used to be public and exported from libtorrent. now that this file can't be found anymore, there are no symbols that get categorized ed25519, and I think the file reference-ed25519 never gets generated

arvidn:
at the top of docs/gen_reference_doc.py, there's a list of glob patterns for headers to look in for generating the documentation. you have to add ``include/libtorrent/ed25519/*.hpp``

aldenml:
Ok, let me revert those deletions and change `gen_reference_doc.py`

arvidn:
I don't think you need to do both. if you move the header file back, it will get picked up by the python script. But if you don't, you have to tell it where to look

aldenml:
@arvidn please don't merge this, let me figure out a way to better review the changes

PR #1036, opened 2016-08-27
minor cleanup/refactor

aldenml:


src/peer_connection.cpp
```diff
@@ -721,7 +720,7 @@ namespace libtorrent
 			TORRENT_ASSERT(m_have_piece.size() == t->torrent_file().num_pieces());
 			t->peer_has(m_have_piece, this);
 			bool interesting = false;
-			for (int i = 0; i < int(m_have_piece.size()); ++i)
+			for (int i = 0; i < m_have_piece.size(); ++i)
```
arvidn: wouldn't this yield a warning?

PR #1037, opened 2016-08-27
better integration and API for ed25519

aldenml:
I think this is a good approach


src/kademlia/ed25519.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+#include <libtorrent/kademlia/ed25519.hpp>
+#include <libtorrent/random.hpp>
+#include <libtorrent/ed25519.hpp>
+
+namespace libtorrent {
+namespace dht
+{
+
+	void ed25519_create_seed(std::array<char, 32>& seed)
```
arvidn: Do you think it would be prohibitively expensive for this to return the ``std::array<>``? I would expect RVO to kick in and generate as efficient code, while providing a more convenient API and making SSA analysis simpler.

src/kademlia/ed25519.cpp
```diff
@@ -0,0 +1,99 @@
...
+	{
+		aux::random_bytes(seed);
+	}
+
+	void ed25519_create_keypair(public_key& pk
```
arvidn: I guess a lot of these functions should return their values instead of taking out-parameters.

src/kademlia/ed25519.cpp
aldenml: 32 bytes in no critical path... I can't see any problem

src/kademlia/ed25519.cpp
arvidn: maybe it's unreasonable to conflate that patch with this one

src/random.cpp
```diff
@@ -48,6 +61,51 @@ namespace libtorrent
 #endif
 			return rng;
 		}
+
+		void random_bytes(span<char> buffer)
```
arvidn: at some point this should probably be replaced by ``std::random_device``

src/random.cpp
```diff
@@ -48,6 +61,51 @@ namespace libtorrent
...
+#ifndef BOOST_NO_EXCEPTIONS
+				throw system_error(error_code(ERR_get_error(), system_category()));
+#else
+				std::terminate();
+#endif
+			}
+#else
+			std::uint32_t s = random(0xffffffff);
+			std::independent_bits_engine<std::mt19937, 8, std::uint8_t> generator(s);
+			std::generate(buffer.begin(), buffer.end(), std::ref(generator));
```
arvidn: heh, I suppose I wrote this. this is pretty weak

src/random.cpp
aldenml: the problem is that `std::random_device` is not specified to generate cryptographically PSRN

src/random.cpp
aldenml: very true, actually, I wrote this, but you have a choice, use `openssl` :)

src/random.cpp
arvidn: well, nor should it be. It is a safe assumption that it provides the best possible entropy source available for the system you're running on. (and, no, I don't consider old versions of gcc to matter). My understanding is that the mainstream compilers do this correctly (possibly with an exception of GCC on Windows, but if people cared about using GCC on windows, I imagine they'd fix it).

PR #1046, opened 2016-08-31
fix exit condition when parsing nodes

ssiloti:


include/libtorrent/socket_io.hpp
```diff
@@ -56,6 +56,17 @@ namespace libtorrent
 
 	namespace detail
 	{
+		template <class Proto>
+		size_t address_size(Proto p)
+		{
```
arvidn: I guess you could put: TORRENT_UNUSED(p);

src/kademlia/node.cpp
```diff
@@ -621,7 +621,7 @@ struct ping_observer : observer
 			char const* nodes = n.string_ptr();
 			char const* end = nodes + n.string_length();
 
-			while (end - nodes >= 26)
+			while (end - nodes >= 20 + detail::address_size(protocol) + 2)
```
arvidn: do you think it would be hard to unit test the ping_observer for v6 nodes? and, is this the only observer that parses the node string btw?

arvidn: It seems this same problem exists in traversal_algorithm.cpp as well

PR #1049, opened 2016-08-31
refactor to use std::shared_ptr with http_connection

aldenml:


arvidn:
I would expect there to be a few includes of ``<boost/shared_ptr.hpp>`` that could be removed as well. Also, would you mind adding a note to ChangeLog?

aldenml:
There are a lot of includes to remove, but I can't do it until the very end of the migration to std::shared_ptr, since there are lot of boost::shared_ptr around. Do you mind if I add the note to the ChangeLog in my next commit?

arvidn:
sure

PR #1050, opened 2016-08-31
refactor to use std::shared_ptr with peer_connection, core and tests

aldenml:


bindings/c/library.cpp
```diff
@@ -60,7 +60,7 @@ namespace
 	int add_handle(libtorrent::torrent_handle const& h)
 	{
 		std::vector<libtorrent::torrent_handle>::iterator i = std::find_if(handles.begin()
-			, handles.end(), !boost::bind(&libtorrent::torrent_handle::is_valid, _1));
+			, handles.end(), !std::bind(&libtorrent::torrent_handle::is_valid, _1));
```
arvidn: I'm actually a bit surprised this works. are you sure it means the same thing as with boost::bind? boost has some lambda capabilities which did not make it into the standard, and I would expect this to be one of them.

aldenml: this is the broken c binding, it does not compile and it's not usable

aldenml: but I saw what you say in the python binding, I needed to used `boost::bind` there in a few places

arvidn: would you mind replacing this bind() with a lambda expression? [](libtorrent::torrent_handle const& h) { return !h.is_valid(); }

arvidn:
lgtm (but see comment)

bindings/c/library.cpp
arvidn: ah, so you could use a lambda there then

aldenml:
do you mind if I do it in another PR?

PR #1056, opened 2016-09-02
refactor to use std::shared_ptr with traversal_algorithm

aldenml:


include/libtorrent/kademlia/traversal_algorithm.hpp
```diff
@@ -41,10 +41,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <libtorrent/kademlia/observer.hpp>
 #include <libtorrent/address.hpp>
 
-#include "libtorrent/aux_/disable_warnings_push.hpp"
 #include <boost/noncopyable.hpp>
```
arvidn: I think it's risky to not suppress warnings for this boost header

aldenml: true, I did it on purpose

src/kademlia/traversal_algorithm.cpp
```diff
@@ -524,10 +521,10 @@ void traversal_algorithm::add_router_entries()
 			, static_cast<void*>(this), int(std::distance(m_node.m_table.router_begin(), m_node.m_table.router_end())));
 	}
 #endif
-	for (routing_table::router_iterator i = m_node.m_table.router_begin()
+	for (auto i = m_node.m_table.router_begin()
```
arvidn: a range for-loop would work here, no?

aldenml: no, because the `router_begin` is an actual method to get an internal set

aldenml: I will use `std::for_each`

src/kademlia/traversal_algorithm.cpp
```diff
@@ -521,11 +521,10 @@ void traversal_algorithm::add_router_entries()
 			, static_cast<void*>(this), int(std::distance(m_node.m_table.router_begin(), m_node.m_table.router_end())));
 	}
 #endif
-	for (auto i = m_node.m_table.router_begin()
-		, end(m_node.m_table.router_end()); i != end; ++i)
-	{
-		add_entry(node_id(), *i, observer::flag_initial);
-	}
+	std::for_each(m_node.m_table.router_begin()
```
arvidn: why not for (auto const* n : m_node.m_table) ?

aldenml: I don't think `routing_table` is iterable (it does not compile), maybe the names `router_begin` and `router_end` prevents that, maybe is a name change.

aldenml: well, arm failed again, I will change the names if that's it

PR #1061, opened 2016-09-04
refactor to avoid use of optional with piece_block_progress

aldenml:
Ok, this requires a "why" explanation. The main reason is that I'm looking for opportunities to remove `boost` dependencies. The advantages of this change are:

- Less `boost`
- `boost::optional` is not cost-free space wise.

The disadvantage is a little "less clear" API, but reading about `boost::optional`, it looks to me like a `null` version for value types. This should compose seamless in a more functional style, but that requires a lot more in the type (or even the type system).

It is of course, a tradeoff.

### DIFF
--- a/include/libtorrent/bt_peer_connection.hpp
+++ b/include/libtorrent/bt_peer_connection.hpp
@@ -284,7 +284,7 @@ namespace libtorrent
 		// block. If the peer isn't downloading
 		// a piece for the moment, the boost::optional
 		// will be invalid.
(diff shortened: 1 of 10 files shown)


arvidn:
seems like a reasonable patch, and to be honest I don't think that is a very good use of optional. Out of curiosity, what's the value of shedding boost dependencies? I see the value in replacing it with standard library facillities, but you're not likely to drop the boost dependency regardless.

aldenml:
At this point, it's mostly a strong personal preference, sometimes that results in a better build process or even a slightly better source code.

PR #1062, opened 2016-09-04
minor cleanup/refactor

aldenml:


src/torrent.cpp
```diff
@@ -10679,8 +10674,8 @@ namespace libtorrent
 				if (info[k].state == piece_picker::block_info::state_requested)
 				{
 					block = 0;
-					torrent_peer* p = static_cast<torrent_peer*>(info[k].peer);
```
arvidn: how come we don't need a cast here?

arvidn: I guess this used to be a void*, but not anymore

PR #1067, opened 2016-09-06
use of std::unordered_multimap and fix in resolve_links::match

aldenml:
Hi @arvidn, before continue with this PR, I want to confirm if in fact there is a bug (and fix) in `resolve_links::match`. The change is
```
- auto iter = m_file_sizes.find(file_size);
```
for
```
+ auto range = m_file_sizes.equal_range(file_size);
  for (auto iter = range.first; iter != range.second; ++iter)
```

### DIFF
--- a/src/resolve_links.cpp
+++ b/src/resolve_links.cpp
@@ -84,48 +84,49 @@ void resolve_links::match(std::shared_ptr<const torrent_info> const& ti
 
 		std::int64_t file_size = fs.file_size(i);
 
-		auto iter = m_file_sizes.find(file_size);
-
-		// we don't have a file whose size matches, look at the next one
-		if (iter == m_file_sizes.end()) continue;
-
-		TORRENT_ASSERT(iter->second < m_torrent_file->files().num_files());
-		TORRENT_ASSERT(iter->second >= 0);
-
-		// if we already have found a duplicate for this file, no need
-		// to keep looking
-		if (m_links[iter->second].ti) continue;
-
-		// files are aligned and have the same size, now start comparing
-		// piece hashes, to see if the files are identical
-
-		// the pieces of the incoming file
-		int their_piece = fs.map_file(i, 0, 0).piece;
-		// the pieces of "this" file (from m_torrent_file)
-		int our_piece = m_torrent_file->files().map_file(
-			iter->second, 0, 0).piece;
-
-		int num_pieces = (file_size + piece_size - 1) / piece_size;
-
-		bool match = true;
-		for (int p = 0; p < num_pieces; ++p, ++their_piece, ++our_piece)
+		auto range = m_file_sizes.equal_range(file_size);
+		for (auto iter = range.first; iter != range.second; ++iter)
 		{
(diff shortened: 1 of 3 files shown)


src/resolve_links.cpp
```diff
@@ -84,48 +84,49 @@ void resolve_links::match(std::shared_ptr<const torrent_info> const& ti
...
+		auto range = m_file_sizes.equal_range(file_size);
+		for (auto iter = range.first; iter != range.second; ++iter)
 		{
-			if (m_torrent_file->hash_for_piece(our_piece)
-				!= ti->hash_for_piece(their_piece))
-			{
-				match = false;
-				break;
+			// we don't have a file whose size matches, look at the next one
+			if (iter == m_file_sizes.end()) continue;
```
arvidn: this condition will presumably never be true..

arvidn:
it does look like a legitimate problem. could you look at the tests and see if you could add coverage for this case?

arvidn:
This should probably be fixed on RC_1_1 as well (with a comment in ChangeLog)

aldenml:
Sure, I will take a look and back port to RC_1_1

test/test_resolve_links.cpp
```diff
@@ -131,6 +133,48 @@ TORRENT_TEST(resolve_links)
...
+
+	fs1.add_file("test_resolve_links_dir/tmp1", 1024);
+	fs1.add_file("test_resolve_links_dir/tmp2", 1024);
+	fs2.add_file("test_resolve_links_dir/tmp1", 1024);
+	fs2.add_file("test_resolve_links_dir/tmp2", 1024);
+
+	libtorrent::create_torrent t1(fs1, 1024);
+	libtorrent::create_torrent t2(fs2, 1024);
+
+	t1.set_hash(0, sha1_hash::max());
```
arvidn: don't you need to set the hash for t2 as well? or any of the other pieces?

test/test_resolve_links.cpp
```diff
@@ -131,6 +133,48 @@ TORRENT_TEST(resolve_links)
 		TEST_EQUAL(num_matches, e.expected_matches);
 
 	}
-#endif // TORRENT_DISABLE_MUTABLE_TORRENTS
 }
 
+// this ensure that internally the is a range lookup
```
arvidn: s/the/there

test/test_resolve_links.cpp
aldenml: the hashes by default are `0` (an internal `m_piece_hash.resize(m_files.num_pieces())`), I only set the `t1` first piece hash to `max` and that do the trick of having the same size but different hash compared to the `t2`.

test/test_resolve_links.cpp
aldenml: do you mind if I fix the typo in my next PR?

test/test_resolve_links.cpp
arvidn: ah, I see

PR #1068, opened 2016-09-06
backport of fix in resolve_links::match

aldenml:


arvidn:
did the test not apply cleanly?

aldenml:
actually, I tried to keep it to the minimum

PR #1070, opened 2016-09-06
refactor to use std unordered containers and minor cleanup

aldenml:


include/libtorrent/torrent_handle.hpp
```diff
@@ -199,8 +196,8 @@ namespace libtorrent
 		state_t piece_state;
 	};
 
-	// for boost::hash (and to support using this type in unordered_map etc.)
-	std::size_t hash_value(torrent_handle const& h);
+	// for std::hash (and to support using this type in unordered_map etc.)
+	TORRENT_EXPORT std::size_t hash_value(torrent_handle const& h);
```
arvidn: this should be picked up by boost::unordered_map<> right? It seems good to stay backwards compatible

PR #1072, opened 2016-09-07
more boost types cleanup and handling of deprecated macos libcrypto

aldenml:


arvidn:
lgtm. it would be nice to get rid of some of those blobs of disabling warnings on mac. maybe I'll look into using the native API for hashing at least

aldenml:
I believe you already use the native API (built-in), but if you use the embedded openssl in macOS, that's the problem, because is outdated, deprecated and unmaintained :)

arvidn:
oh.. but wouldn't it make sense to leave those warnings around then? My understanding is that boost.asio only supports libssl for its TLS support, so those uses would have to be wrapped, but not anything else.

aldenml:
well...hmm....thinking about it...I suppose that for a while clients should be able to use the embedded openssl, they should be aware of the shortcomings, but at the same time be able to compile warnings free without adding `-Wno-deprecated-declarations`.

PR #1073, opened 2016-09-08
dht log optimization, consts and refactor

aldenml:


include/libtorrent/aux_/session_impl.hpp
```diff
@@ -619,10 +619,11 @@ namespace libtorrent
...
 
 #ifndef TORRENT_DISABLE_LOGGING
-			virtual void log(libtorrent::dht::dht_logger::module_t m, char const* fmt, ...)
-				override TORRENT_FORMAT(3,4);
+			virtual bool should_log(module_t m) const override;
+			virtual void log(module_t m, char const* fmt, ...)
+				const override TORRENT_FORMAT(3,4);
 			virtual void log_packet(message_direction_t dir, char const* pkt, int len
-				, udp::endpoint node) override;
+				, udp::endpoint node) const override;
```
arvidn: it's not obvious to me what this "const" buy you. for instance, it would make it less convenient to create an observer that store loglines an a std::vector<std::string> member variable (since it would have to be mutable or const_cast<>ed)

test/test_dht.cpp
```diff
@@ -497,15 +498,15 @@ struct obs : dht::dht_observer
...
+		, udp::endpoint node) const override {}
 #endif
 	bool on_dht_request(string_view query
 		, dht::msg const& request, entry& response) override { return false; }
 
 	address m_external_address = addr4("236.0.0.1");
 
 #ifndef TORRENT_DISABLE_LOGGING
-	std::vector<std::string> m_log;
+	mutable std::vector<std::string> m_log;
```
arvidn: aha! see what I mean? :)

aldenml: yes :)

arvidn:
lgtm, except I don't get the const on the dht-logger interface functions

aldenml:
can I change that in my next PR?

PR #1074, opened 2016-09-08
refactor to allow session state from the start

aldenml:
I want to know what do you think about this API change before going to the unit tests.

### DIFF
--- a/include/libtorrent/session.hpp
+++ b/include/libtorrent/session.hpp
@@ -149,6 +149,8 @@ namespace libtorrent
 		libtorrent::dht_settings dht_settings;
 
 		dht::dht_storage_constructor_type dht_storage_constructor;
+
+		bdecode_node state;
 	};
 
 	// The session holds all state that spans multiple torrents. Among other
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -56,7 +56,7 @@ namespace libtorrent
 	struct settings_pack;
 	struct bdecode_node;
 
-	TORRENT_EXTRA_EXPORT std::shared_ptr<settings_pack> load_pack_from_dict(bdecode_node const& settings);
+	TORRENT_EXTRA_EXPORT void load_pack_from_dict(bdecode_node const& settings, settings_pack& pack);
 	TORRENT_EXTRA_EXPORT void save_settings_to_dict(aux::session_settings const& s, entry::dictionary_type& sett);
 	TORRENT_EXTRA_EXPORT void apply_pack(settings_pack const* pack, aux::session_settings& sett
 		, aux::session_impl* ses = nullptr);
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -594,16 +594,23 @@ namespace aux {
 	void session_impl::load_state(bdecode_node const* e
 		, std::uint32_t const flags)
 	{
+		settings_pack sp;
+		load_state_impl(*e, flags, sp, true);
+	}
+
+	void session_impl::load_state_impl(bdecode_node const& e
+		, std::uint32_t const flags, settings_pack& pack, bool const apply)
+	{
 		TORRENT_ASSERT(is_single_thread());
 
 		bdecode_node settings;
-		if (e->type() != bdecode_node::dict_t) return;
+		if (e.type() != bdecode_node::dict_t) return;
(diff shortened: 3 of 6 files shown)


include/libtorrent/settings_pack.hpp
```diff
@@ -56,7 +56,7 @@ namespace libtorrent
 	struct settings_pack;
 	struct bdecode_node;
 
-	TORRENT_EXTRA_EXPORT std::shared_ptr<settings_pack> load_pack_from_dict(bdecode_node const& settings);
+	TORRENT_EXTRA_EXPORT void load_pack_from_dict(bdecode_node const& settings, settings_pack& pack);
```
arvidn: I haven't looked through the whole change yet, but if you want to avoid ``shared_ptr``, why not return by value? (i.e. move)

include/libtorrent/session.hpp
```diff
@@ -149,6 +149,8 @@ namespace libtorrent
 		libtorrent::dht_settings dht_settings;
 
 		dht::dht_storage_constructor_type dht_storage_constructor;
+
+		bdecode_node state;
```
arvidn: I think there is a member to hold this in ``session_impl`` already.. isn't there?

src/session_impl.cpp
```diff
@@ -594,16 +594,23 @@ namespace aux {
 	void session_impl::load_state(bdecode_node const* e
 		, std::uint32_t const flags)
 	{
+		settings_pack sp;
+		load_state_impl(*e, flags, sp, true);
+	}
+
+	void session_impl::load_state_impl(bdecode_node const& e
```
arvidn: I feel like perhaps it would be cleaner if there was a free function that could load state from a file into a ``session_params`` object, and make sure all state can fit in there.

src/session_impl.cpp
```diff
@@ -697,15 +704,15 @@ namespace aux {
...
 #endif
-
+		bool need_apply = false;
 		if (flags & session::save_settings)
 		{
-			settings = e->dict_find_dict("settings");
+			settings = e.dict_find_dict("settings");
 			if (settings)
 			{
+				load_pack_from_dict(settings, pack);
```
arvidn: are you relying in ``load_pack_from_dict`` to augment the pack? so it's an in-out parameter?

src/session_impl.cpp
```diff
@@ -697,15 +704,15 @@ namespace aux {
...
 				// apply_settings_pack will update dht and proxy
-				std::shared_ptr<settings_pack> pack = load_pack_from_dict(settings);
-				apply_settings_pack(pack);
+				need_apply = true;
```
arvidn: hm.. but you only apply it if we load it from the settings, so anything passed in through it would be ignored, unless loaded from the settings file..

arvidn:
I'm a little bit lost. as far as I can tell, there's no public API being changed here.. am I missing something?

include/libtorrent/settings_pack.hpp
aldenml: I changed it only to support the change in a DRY style

include/libtorrent/session.hpp
aldenml: nope, it does not have it

src/session_impl.cpp
aldenml: This seems like a better idea

src/session_impl.cpp
aldenml: yes

aldenml:
The only public addition is `session_params::state`, the rest was mostly to support the addition of this field. I think your idea of having the free function (plus the deprecation of `load_state` is better). I will close this to figure out a better path.

include/libtorrent/session.hpp
arvidn: I was thinking of ``m_dht_state``: https://github.com/arvidn/libtorrent/blob/master/include/libtorrent/aux_/session_impl.hpp#L924

aldenml: what do you think is the best structure to store a "dht state" in `session_params`?

arvidn: probably not in a bencoded ``entry`` object. I think something more agnostic to how it's currently stored in bencoded form. i.e. a vector of nodes and a ``node_id`` for the last session's id. etc.

PR #1077, opened 2016-09-09
refactor to load dht state with session_params

aldenml:
@arvidn what do you think of this approach? (this is only the DHT part)


aldenml:
This requires a better design. Closing this

PR #1078, opened 2016-09-10
some refactor, more const refs and span use

aldenml:


include/libtorrent/bencode.hpp
```diff
@@ -82,11 +79,6 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent
 {
-
-#ifndef TORRENT_NO_DEPRECATE
-	using invalid_encoding = system_error;
```
arvidn: why remove this?

include/libtorrent/kademlia/node_id.hpp
```diff
@@ -64,7 +64,7 @@ int TORRENT_EXTRA_EXPORT min_distance_exp(node_id const& n1, std::vector<node_id
 
 node_id TORRENT_EXTRA_EXPORT generate_id(address const& external_ip);
 node_id TORRENT_EXTRA_EXPORT generate_random_id();
-void TORRENT_EXTRA_EXPORT make_id_secret(node_id& in);
```
arvidn: I don't understand this change either. is the attribute has to come before the return type, presumably that would apply to all the other functions here too..

include/libtorrent/settings_pack.hpp
```diff
@@ -244,7 +244,7 @@ namespace libtorrent
 			// ``send_redundant_have`` controls if have messages will be sent to
 			// peers that already have the piece. This is typically not necessary,
 			// but it might be necessary for collecting statistics in some cases.
-			// Default is false.
+			// Default is true.
```
arvidn: it may even be better to remove the mention of what the default is. the script that generates the documentation will include the default value already

arvidn:
lgtm, apart from comments

include/libtorrent/bencode.hpp
aldenml: not important

include/libtorrent/kademlia/node_id.hpp
aldenml: maybe I will change all the group at some point

PR #1079, opened 2016-09-11
more dht log optimizations

aldenml:


src/kademlia/find_data.cpp
```diff
@@ -110,10 +109,12 @@ void find_data::start()
 void find_data::got_write_token(node_id const& n, std::string write_token)
 {
 #ifndef TORRENT_DISABLE_LOGGING
-	get_node().observer()->log(dht_logger::traversal
-		, "[%p] adding write token '%s' under id '%s'"
-		, static_cast<void*>(this), aux::to_hex(write_token).c_str()
-		, aux::to_hex(n).c_str());
+	auto logger = get_node().observer();
+	if (logger != nullptr && logger->should_log(dht_logger::traversal))
```
arvidn: I think it would be clearer to add the { and } here, since the statement is multiple lines

aldenml: ok, I will add in the others too

src/kademlia/get_peers.cpp
```diff
@@ -72,10 +72,12 @@ void get_peers_observer::reply(msg const& m)
 			char const* end = peers + n.list_at(0).string_length();
 
 #ifndef TORRENT_DISABLE_LOGGING
+			auto logger = get_observer();
 			bdecode_node const id = r.dict_find_string("id");
```
arvidn: technically, this lookup could be moved inside the ``should_log()`` check as well

src/kademlia/get_peers.cpp
```diff
@@ -94,18 +96,20 @@ void get_peers_observer::reply(msg const& m)
 			// assume it's uTorrent/libtorrent format
 			read_endpoint_list<tcp::endpoint>(n, peer_list);
 #ifndef TORRENT_DISABLE_LOGGING
+			auto logger = get_observer();
 			bdecode_node const id = r.dict_find_string("id");
```
arvidn: and here I think

PR #1082, opened 2016-09-12
more on log, consts, refs and cleanup in dht related code

aldenml:


include/libtorrent/kademlia/routing_table.hpp
```diff
@@ -250,7 +250,7 @@ class TORRENT_EXTRA_EXPORT routing_table
 
 	void replacement_cache(bucket_t& nodes) const;
 
-#ifndef TORRENT_DISABLE_LOGGING
+#if TORRENT_USE_IOSTREAM
```
arvidn: this is more of a debug facility, isn't it?

src/kademlia/item.cpp
```diff
@@ -69,7 +68,6 @@ namespace
 			left = out.size() - (ptr - out.data());
 			std::memcpy(ptr, salt.data(), (std::min)(salt.size(), left));
 			ptr += (std::min)(salt.size(), left);
-			left = out.size() - (ptr - out.data());
```
arvidn: ``left`` is used after this, I don't think this is safe to remove. note that ``ptr`` will have a new value in this expression than the one two lines up

include/libtorrent/kademlia/routing_table.hpp
aldenml: yes (and not relevant when logging is enable)

src/kademlia/item.cpp
aldenml: but it's updated two lines below, rendering this update null. Do you want it back regardless?

src/kademlia/routing_table.cpp
```diff
@@ -227,22 +220,22 @@ int routing_table::depth() const
...
 	{
 		--m_depth;
 	}
 
 	return m_depth;
 }
 
-#ifndef TORRENT_DISABLE_LOGGING
+#if TORRENT_USE_IOSTREAM
 void routing_table::print_state(std::ostream& os) const
```
arvidn: I'm not sure exactly where this is used, but I think, historically, it's been used as a debug logging facility. I'm a bit tempted to weed out uses of the ``TORRENT_USE_IOSTREAM`` macro.

aldenml: actually is used a lot in `dht_test`, I will remove the guards because in unit test, `TORRENT_USE_IOSTREAM` should be enable

include/libtorrent/kademlia/node.hpp
```diff
@@ -134,7 +134,7 @@ class TORRENT_EXTRA_EXPORT node : boost::noncopyable
 	int data_size() const { return int(m_storage.num_torrents()); }
 #endif
 
-#ifndef TORRENT_DISABLE_LOGGING
+#if TORRENT_USE_IOSTREAM
```
arvidn: with your new ``should_log`` it may be practical to actually make this print to the log. which would seem like a reason to keep it under TORRENT_DISABLE_LOGGING

include/libtorrent/kademlia/node.hpp
```diff
@@ -134,7 +134,7 @@ class TORRENT_EXTRA_EXPORT node : boost::noncopyable
...
 	// TODO: 3 make this print to the DHT logger instead
```
arvidn: I'm thinking of this TODO comment. this change seems to be moving in the opposite direction. I'm not married to this idea if you have a good reason to move in this direction

aldenml: I thought about it, but I don't think this `print_state` is a good fit for the logging, it seems like a potentially very big string, that's the reason I didn't tried to move it to the log.

arvidn: sure, maybe it doesn't make sense to print to the log. I'm still not sure about the rationale for this change. USE_IO_STREAM is a macro defining the environment you're building for. This function, which are just used for debugging (in the dht tests), should not be forced upon end-users I think. With this change there's no way to disable it. Arguably, it's not great to have the only way to disable it be by disabling all of logging either, but it's at least a build configuration macro. Ideally it would be moved out to a free function, maybe even living entirely within the tests

aldenml: agree, moving it to inside the `test_dht`

PR #1083, opened 2016-09-12
invalid memory access in utp_stream and torrent

milesdong:
1. packet->mtu_probe in utp_stream is not initialized, which can make mtu probe enable on small packet，lead to a lower mtu or even be attacked
2. torrent with a 0 size padding leads crash when init torrent


src/utp_stream.cpp
```diff
@@ -1888,6 +1889,7 @@ bool utp_socket_impl::send_pkt(int const flags)
 		if (payload_size)
 		{
 			p = static_cast<packet*>(malloc(sizeof(packet) + effective_mtu));
+			memset(p, 0, sizeof(packet));
```
arvidn: would it make more sense to initialize ``mtu_probe`` to false here? it seems unnecessarily expensive to clear the entire buffer, immediately before a ``memcpy()``.

src/torrent.cpp
```diff
@@ -1825,9 +1825,10 @@ namespace libtorrent
 		file_storage const& fs = m_torrent_file->files();
 		for (int i = 0; i < fs.num_files(); ++i)
 		{
+			if (!fs.pad_file_at(i) || fs.file_size(i) == 0) continue;
```
arvidn: So this makes files with 0 size that are marked as pad files still don't count as padfiles anymore. did the previous logic cause a problem for you?

arvidn:
good catch. would it be a lot of trouble to submit this PR against RC_1_1 instead? (which is the latest stable branch)

src/torrent.cpp
milesdong: need_picker() is not called and will crash when calling `picker->we_have` for padding files

milesdong:
see #1084

PR #1086, opened 2016-09-13
peer_connection log related optimization

aldenml:


include/libtorrent/alert_manager.hpp
```diff
@@ -81,13 +81,17 @@ namespace libtorrent {
 		template <class T>
 		bool should_post() const
 		{
+			if ((m_alert_mask & T::static_category) == 0)
```
arvidn: m_alert_mask is protected by the mutex. if you want to avoid the lock in this case, you could make it an atomic. I imagine the accesses to it could use ``memory_order_relaxed``.

arvidn:
lgtm

PR #1087, opened 2016-09-13
Use std::random instead of Boost.Random when in c++11 mode.

sledgehammer999:
I am sorry but I couldn't modify the different build systems.

Especially autotools require extra work. At this point configure.ac hardcodes dependencies of Boost.Chrono and Boost.Random regardless if you're in c++11 mode. I think it would be best if the script was updated to include one option to enable c++11. And then depending on that require Boost.Chrono and Boost.Random. And maybe make enabling c++11 the default?
(PS: It should be checked if c++11 is enabled anyway either by the compiler itself or by user passing the relevant CXXFLAGS to the script without using the script's dedicated option)

I have no idea how jamfiles or cmake work and what they require.

Currently I tested only on Windows with msvc 2015 (static builds) and the dependency on Boost.Random is indeed dropped. (Hint: On msvc 2015 c++11 is always on).
Due the simplicity of the patch I expect that it will work with any other compiler provided that the build system is fixed.

### DIFF
--- a/src/random.cpp
+++ b/src/random.cpp
@@ -34,6 +34,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/random.hpp"
 #include "libtorrent/assert.hpp"
 
+#ifdef BOOST_NO_CXX11_HDR_RANDOM
 #include "libtorrent/aux_/disable_warnings_push.hpp"
 
 #include <boost/random/random_device.hpp>
@@ -41,6 +42,9 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <boost/random/uniform_int_distribution.hpp>
 
 #include "libtorrent/aux_/disable_warnings_pop.hpp"
+#else
+#include <random>
+#endif
 
 #if !TORRENT_THREADSAFE_STATIC
 #include "libtorrent/thread.hpp"
@@ -48,9 +52,15 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent
 {
+#ifdef BOOST_NO_CXX11_HDR_RANDOM
 	using boost::random::random_device;
 	using boost::random::mt19937;
 	using boost::random::uniform_int_distribution;
+#else
+	using std::random_device;
+	using std::mt19937;
+	using std::uniform_int_distribution;
+#endif
 
 #ifdef TORRENT_BUILD_SIMULATOR
 


arvidn:
I get the impression that it's a bigger problem to have too many build configuration options. I've been trying to trim down that complexity by having fewer build options.

I suppose this patch is harmless though, in the sense that all it does is opt out of actually referencing the boost.random symbols.

The next major release of libtorrent will require C++11 and drop both the boost.chrono and boost.random dependencies.

src/random.cpp
```diff
@@ -48,9 +52,13 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent
 {
+#ifdef BOOST_NO_CXX11_HDR_RANDOM
 	using boost::random::random_device;
 	using boost::random::mt19937;
 	using boost::random::uniform_int_distribution;
+#else
+	using namespace std;
```
arvidn: I think this should just pull in the names that are used, just like above.

sledgehammer999: You're right. Let me test locally and I'll update it.

src/random.cpp
```diff
@@ -36,9 +36,13 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "libtorrent/aux_/disable_warnings_push.hpp"
```
arvidn: these disable_warnings include should probably be moved inside the check

src/random.cpp
```diff
@@ -36,9 +36,13 @@ POSSIBILITY OF SUCH DAMAGE.
...
 
+#ifdef BOOST_NO_CXX11_HDR_RANDOM
```
arvidn: is this reliable? I would expect that perhaps you could do something like: ``#ifdef __has_include && __has_include(<random>)

sledgehammer999: I don't know if `BOOST_NO_CXX11_HDR_RANDOM` is reliable. It is part of Boost.Config which you already use. I had thought of just checking if `__cplusplus` is c++11.
I have no preference. Whatever you tell me to put here I'll do it.

arvidn: yeah, I guess other parts of the code checks the __cplusplus define, that's probably the best to use. The reason I would be a bit worried about the boost define is that ``boost/config.hpp`` may not always have been included at this point

sledgehammer999:
Updated commit.

>I get the impression that it's a bigger problem to have too many build configuration options. I've been trying to trim down that complexity by having fewer build options.

OK. I just wanted to point out the following. Unless I am mistaken the configure script will require the presence of Boost.Chrono and Boost.Random even if c++11 is turned on.

src/random.cpp
sledgehammer999: Ok, I'll do it in a couple of hours. I gotta go now.

arvidn:
right, this will probably make it simpler when not using the configure script though

aldenml:
I'm just curious in my question (I'm for removing boost as much as possible). How is this beneficial in `RC_1_1`, since you will no drop the dependencies? I believe that if you don't use `BOOST_ASIO_HAS_STD_CHRONO` then `asio` will need that binary

sledgehammer999:
>I'm just curious in my question (I'm for removing boost as much as possible). How is this beneficial in RC_1_1, since you will no drop the dependencies? I believe that if you don't use BOOST_ASIO_HAS_STD_CHRONO then asio will need that binary 

At least on Windows with boost 1.61 asio is able to automatically tell if we're on c++11 mode and doesn't use Boost.Chrono. (on windows I use the jamfile)

On linux I did not do proper tests.

arvidn:
it depends on how you define "dependency". the ``makefile`` and the ``Jamfile`` will try to link against those libraries, but if you drop all the sources in a visual studio (2015) project and try to build it, there will be one less dependency to resolve

aldenml:
@sledgehammer999 thanks, didn't know about that feature in asio

src/random.cpp
```diff
@@ -34,23 +34,33 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/random.hpp"
 #include "libtorrent/assert.hpp"
 
+#ifdef BOOST_NO_CXX11_HDR_RANDOM
```
sledgehammer999: I tried to do it with `__cplusplus` but msvc 2015 still has a value of `199711L` for it!!!

For your worries about boost.config being included. This file at the top includes libtorrent/config.hpp. And that header includes boost/config.hpp at the top.
Or we could include boost/config.hpp here too just to be safe from future changes in libtorrent/config.hpp

sledgehammer999: Forgot to tag you @arvidn

PR #1088, opened 2016-09-13
general session, torrent and peer connection log optimization

aldenml:


src/session_impl.cpp
```diff
@@ -6865,6 +6950,7 @@ namespace aux {
 			, struct tracker_response const& resp)
 		{
 			TORRENT_UNUSED(tracker_ips);
+			if (!should_log()) return;
```
arvidn: take a look at where the tracker_logger is created. If logging is disabled, we shouldn't create it in the first place

src/session_impl.cpp
```diff
@@ -6874,47 +6960,46 @@ namespace aux {
...
-		{
-			debug_log("*** tracker timed out");
-		}
-
 		void tracker_logger::tracker_request_error(tracker_request const&
 			, int response_code, error_code const& ec, const std::string& str
 			, int retry_interval)
 		{
 			TORRENT_UNUSED(retry_interval);
+			if (!should_log()) return;
```
arvidn: ditto

src/session_impl.cpp
aldenml: do you mean surrounding the calls with the macro? all the existence of `tracker_logger` is inside the macro

include/libtorrent/torrent.hpp
```diff
@@ -956,6 +956,7 @@ namespace libtorrent
 
 		// LOGGING
 #ifndef TORRENT_DISABLE_LOGGING
+		virtual bool should_log() const override;
```
arvidn: it's not obvious to me why this is virtual. which interface does this implement? Is it something tracker_manager uses?

include/libtorrent/tracker_manager.hpp
```diff
@@ -244,6 +244,7 @@ namespace libtorrent
 			, int retry_interval) = 0;
 
 #ifndef TORRENT_DISABLE_LOGGING
+		virtual bool should_log() const = 0;
```
arvidn: I see, it's this interface

arvidn:
can you think if any way of making this less verbose? I can think of macro-magic, but that's probably not worth it.

aldenml:
I thought about it too, even lambdas, but then you need to pay for not having good support in the IDE or more runtime costs

src/session_impl.cpp
arvidn: I believe this class (tracker_logger) is instantiated in exactly one place. If instead of instantiating it, you pass in nullptr (or an empty shared_ptr perhaps) I believe it will still work, just not log anything. That way you don't even pay the cost of this object allocation

aldenml: This is the code
```java
			debug_log("==> TRACKER REQUEST \"%s\" event: %s abort: %d"
				, req.url.c_str()
				, (req.event==tracker_request::stopped?"stopped"
					:req.event==tracker_request::started?"started":"")
				, m_abort);

			// if we're not logging session logs, don't bother creating an
			// observer object just for logging
			if (m_abort && alerts().should_post<log_alert>())
			{
				std::shared_ptr<aux::tracker_logger> tl(new aux::tracker_logger(m_ses));
				m_ses.queue_tracker_request(req, tl);
			}
			else
#endif
			{
				m_ses.queue_tracker_request(req, shared_from_this());
			}
```
I see you are right, but I'm not quite sure if you want something specific

arvidn: It looks like the construction of the tracker_logger is already protected by ``should_post<>()``, I suppose you could use ``should_log()`` instead, for symmetry. and those checks inside ``tracker_logger`` itself can be removed.

arvidn:
lgtm

PR #1089, opened 2016-09-13
factor out code from alert_manager::should_post template into non-template

arvidn:


aldenml:
question, does this impact the performance?

arvidn:
probably, by generating less code

aldenml:
do you think it worth to have some trick to have that function not inlined?

arvidn:
I think if a compiler thinks it should be inlined, it's probably a good idea. my impression is that compilers are much better at inlining than outlining though, so I think this would help (although, it's a small function and probably only has a negligible impact).

PR #1090, opened 2016-09-14
check the address family of peers in get_peers

ssiloti:


src/kademlia/dht_storage.cpp
```diff
@@ -247,6 +247,10 @@ namespace
 					// peer list
 					if (noseed && iter->seed) continue;
 
+					// only include peers with the right address family
+					if (iter->addr.protocol().family() != protocol.family())
```
arvidn: wouldn't you need to take the ``want`` argument into account here too?

arvidn:
good catch!

src/kademlia/dht_storage.cpp
ssiloti: The `want` argument only applies to the `nodes`/`nodes6` keys. The `value` key always contains peers with the same address family as the request was received from.

arvidn: oh, right. of course

PR #1091, opened 2016-09-14
web_peer_connection: make a copy of the request before calling incomi…

ssiloti:
…ng_piece

This is a revision of 850eca1f21d1f6e2ae6f0177b91e077c42486534. After further
consideration I realized that it may not be safe to pop the request after
calling incoming_piece because it might lead to calling disconnect which
clears the request queue. A call to disconnect would also invalidate the
reference to the request which, although unlikely, might cause supreme
confusion for an extension author who tried to access the dangling reference.


arvidn:
would you mind capturing some of that in a comment in the code too?

PR #1092, opened 2016-09-14
avoid warnings with clang and minor dht_tracker code changes

aldenml:


src/file.cpp
```diff
@@ -60,6 +62,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #pragma GCC diagnostic pop
```
arvidn: can these pragmas be removed now?

aldenml: I didn't want to add `-Wunused-macros` to the `disable_warnings_push.hpp`, not sure if it's a good idea

PR #1093, opened 2016-09-14
consts and refactor, hash_address returns the hash

aldenml:


aldenml:
Questions: do you think that this code, passing `address()`, can affect in a meaningful way the DHT?
```cpp
// TODO: 1 report the proper address of the router as the source IP of
// this vote of our external address, instead of the empty address
set_external_address(ip, source_router, address());
```

since the `cast_vote` will returns false and the last line in `session_impl::set_external_address` is
```
if (m_dht) m_dht->update_node_id();
```

arvidn:
I would only imagine that would make a difference if you're behind a NAT and you don't have any other peers or trackers to talk to. The number of peers that tell you what your external IP would be much greater than a single router normally, so probably pretty marginal effect.

src/ip_voter.cpp
```diff
@@ -111,8 +111,7 @@ namespace libtorrent
 
 		// this is the key to use for the bloom filters
 		// it represents the identity of the voter
-		sha1_hash k;
-		hash_address(source, k);
+		sha1_hash k = hash_address(source);
```
arvidn: now this could even be const :)

aldenml:
Sure, I will put the consts

PR #1095, opened 2016-09-15
rpc_manager log optimization and minor refactor

aldenml:
Added `udp_socket:: local_endpoint()` as a best effort version of the one receiving the `error_code` parameter.

----- old------------------
The change of passing the local endpoint is necessary in order to have later:
1- `dht_tracker` identify to which DHT route the message
2- the DHT `node` specify which udp socket use to send back the response.

### DIFF
--- a/include/libtorrent/udp_socket.hpp
+++ b/include/libtorrent/udp_socket.hpp
@@ -100,6 +100,13 @@ namespace libtorrent
 		bool is_closed() const { return m_abort; }
 		udp::endpoint local_endpoint(error_code& ec) const
 		{ return m_socket.local_endpoint(ec); }
+		// best effort, if you want to know the error, use
+		// ``local_endpoint(error_code& ec)``
+		udp::endpoint local_endpoint() const
+		{
+			error_code ec;
+			return local_endpoint(ec);
+		}
 
 		typedef udp::socket::receive_buffer_size receive_buffer_size;
 		typedef udp::socket::send_buffer_size send_buffer_size;
--- a/src/kademlia/rpc_manager.cpp
+++ b/src/kademlia/rpc_manager.cpp
@@ -209,8 +209,11 @@ void rpc_manager::check_invariant() const
 void rpc_manager::unreachable(udp::endpoint const& ep)
 {
 #ifndef TORRENT_DISABLE_LOGGING
-	m_log->log(dht_logger::rpc_manager, "PORT_UNREACHABLE [ ip: %s ]"
-		, print_endpoint(ep).c_str());
+	if (m_log->should_log(dht_logger::rpc_manager))
+	{
+		m_log->log(dht_logger::rpc_manager, "PORT_UNREACHABLE [ ip: %s ]"
+			, print_endpoint(ep).c_str());
+	}
 #endif
 
 	for (auto i = m_transactions.begin(); i != m_transactions.end();)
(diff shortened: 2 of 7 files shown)


ssiloti:
`dht:node` shouldn't need to be aware of the local endpoint. The `udp_socket_interface` it is provided with should only send packets on the appropriate interface. Also, don't mix unrelated changes in the same commit, it makes reviewing the changes much harder.

aldenml:
agree with the mixing unrelated changes suggestion, I will try to get more granular in the PRs, if it's required in this one, just let me know.

I expected some comments about the way I'm passing around the "local endpoint". The problem is that in the current form, there is no way to link the response of a `dht::node` to a specific interface to get out. This PR is only one half. The other half is making the `udp_socket_interface` receiving that address in `send_packet`.

As in any API design, there are multiple ways of doing the same, the overall problem is that it's not enough to have the output packet sent via the first socket with the same protocol family. You should output the packet via the same socket the request arrived, or the one that logically is tied to the `dht::node`. If you have any idea overall, I would like to hear.

ssiloti:
A node should be bound to a specific socket at construction time, you should just store that association somehow so that all of the node's outgoing packets use that socket. It's a complete waste to include the local endpoint in every message then lookup the associated socket again for every packet.

aldenml:
hmm...that's a little more complicate to achieve in libtorrent, due to the async nature of the component's creation. I thinking about reverting this part to get accepted the rest.

arvidn:
I agree with @ssiloti. I imagine the thing that makes that a little bit more work right now is that the ``session_impl`` object itself implements the socket that's passed in to the dht. Instead, we could have a separate wrapper object, perhaps as part of the ``listen_socket_t`` object, or maybe that object itself could implement the DHT socket interface. In any case, that would be a cleaner separation of the socket outgoing packets are going over.

It's maybe a little bit hacky to make ``listen_socket_t`` implement the dht socket, but it would be simple, and I think it would work.

src/kademlia/rpc_manager.cpp
```diff
@@ -209,8 +209,11 @@ void rpc_manager::check_invariant() const
 void rpc_manager::unreachable(udp::endpoint const& ep)
 {
 #ifndef TORRENT_DISABLE_LOGGING
-	m_log->log(dht_logger::rpc_manager, "PORT_UNREACHABLE [ ip: %s ]"
-		, print_endpoint(ep).c_str());
+	if (m_log->should_log(dht_logger::rpc_manager))
```
arvidn: it would make sense to move these changes over into a separate PR

arvidn:
as for the work on improving the DHT support for multiple IPs; I think it would be really good to write a simple simulation (or perhaps a few) with a few different combinations of IPv6 and IPv4 addresses. Just to have some end-to-end coverage.

aldenml:
Ok, what I will do is transform this PR into just the log optimization and remove the DHT part. I will continue with the feedback in another branch

include/libtorrent/udp_socket.hpp
```diff
@@ -100,6 +100,13 @@ namespace libtorrent
 		bool is_closed() const { return m_abort; }
 		udp::endpoint local_endpoint(error_code& ec) const
 		{ return m_socket.local_endpoint(ec); }
+		// best effort, if you want to know the error, use
+		// ``local_endpoint(error_code& ec)``
+		udp::endpoint local_endpoint() const
```
arvidn: these ``local_endpoint()`` changes aren't needed anymore too, right?

aldenml: well, I think it improve the code in some places? do you think it's dangerous to have it?

arvidn: I see how it's used now.

PR #1096, opened 2016-09-15
created portmap_callback and refactor for natpmp and upnp log optimization

aldenml:


include/libtorrent/portmap.hpp
```diff
@@ -0,0 +1,63 @@
...
+#ifndef LIBTORRENT_PORTMAP_HPP
+#define LIBTORRENT_PORTMAP_HPP
+
+#include "libtorrent/config.hpp"
+
+namespace libtorrent
+{
+	struct TORRENT_EXTRA_EXPORT portmap_callback
+	{
+		constexpr static int map_transport_natpmp = 0;
```
arvidn: maybe this could even be an ``enum class``

src/session_impl.cpp
```diff
@@ -6670,6 +6646,20 @@ namespace aux {
 
 		m_alerts.emplace_alert<dht_pkt_alert>(pkt, len, d, node);
 	}
+
+	bool session_impl::should_log_portmap(int map_transport) const
+	{
+		TORRENT_UNUSED(map_transport);
```
arvidn: instead of using this macro, either comment out the name of the argument, or remove it

arvidn: (i.e. remove the *name*, not the parameter)

arvidn:
I like it. One thing, would it make sense to put this new file under ``include/libtorrent/_aux``? (and in the ``aux`` namespace?). I'm trying to transition all non-public headers and types there.

src/session_impl.cpp
aldenml: yes :)

aldenml:
I think it's a good idea moving it to aux

src/natpmp.cpp
```diff
@@ -170,18 +172,22 @@ bool natpmp::get_mapping(int index, int& local_port, int& external_port, int& pr
...
+bool natpmp::should_log() const
+{
+	return m_callback.should_log_portmap(aux::portmap_transport::natpmp);
+}
+
 TORRENT_FORMAT(2, 3)
 void natpmp::log(char const* fmt, ...) const
 {
 	TORRENT_ASSERT(is_single_thread());
+	if (!should_log()) return;
```
arvidn: isn't this check redundant?

aldenml: I don't think so, I don't check in every `log`, just in the expensive ones, however, if a call is called (say with a single string) I want to avoid the `va_*` and `snprinf`.

aldenml: ideally, I should forward the variadic arguments until the very end in the alerts, but that requires more refactor

arvidn: yes

PR #1100, opened 2016-09-16
created dht state and session params refactor

aldenml:
Hi @arvidn, before going to the test and more details, I want test the idea first. Do you think this is an acceptable solution for now?


src/session_impl.cpp
```diff
@@ -607,46 +720,11 @@ namespace aux {
 		bool need_update_dht = false;
 		if (flags & session::save_dht_settings)
 		{
+			m_dht_settings = read_dht_settings(settings);
```
aldenml: wrong line

aldenml:
repeat, don't get too much in the details, it's a rushed PR, just to get a quick feedback

include/libtorrent/kademlia/dht_state.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+namespace libtorrent {
+namespace dht
+{
+	struct TORRENT_EXPORT dht_state
+	{
+		node_id nid;
+		node_id nid6;
+
+		std::vector<udp::endpoint> nodes;
```
arvidn: if you're using ``endpoint``, you don't really need two separate  vectors

include/libtorrent/kademlia/dht_tracker.hpp
```diff
@@ -71,10 +72,10 @@ namespace libtorrent { namespace dht
 			, dht_settings const& settings
 			, counters& cnt
 			, dht_storage_interface& storage
-			, entry const& state);
+			, dht_state const& state);
```
arvidn: I like this, much better than passing in an ``entry``

src/kademlia/dht_tracker.cpp
```diff
@@ -198,9 +170,9 @@ namespace libtorrent { namespace dht
 
 		m_refresh_timer.expires_from_now(seconds(5), ec);
 		m_refresh_timer.async_wait(std::bind(&dht_tracker::refresh_timeout, self(), _1));
-		m_dht.bootstrap(initial_nodes, f);
+		m_dht.bootstrap(bootstrap.nodes, f);
```
arvidn: I see, it makes this convenient

src/session.cpp
```diff
@@ -287,6 +287,45 @@ namespace libtorrent
 	// configurations this will give a link error
 	void TORRENT_CFG() {}
 
+	session_params read_session_params(bdecode_node const& e, std::uint32_t const flags)
```
arvidn: I'm a bit surprised there isn't a function like this already. There's just a member function for doing this now, is there? (which probably should be deprecated).
This makes sense.

arvidn: (it would need documentation in the header though)

arvidn:
I think this is a good direction. Probably some of the dht loading/saving functions should still live under ``src/kademlia`` though.

aldenml:
good, I will continue then...

aldenml:
rebased for a complete review

src/session.cpp
aldenml: this is the new free function you mentioned. I deprecated the member one, and it was a mess, it's used a lot

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -180,7 +176,10 @@ namespace libtorrent
 		struct tracker_logger;
 #endif
 
-		TORRENT_EXPORT std::pair<bencode_map_entry*, int> settings_map();
+#ifndef TORRENT_DISABLE_DHT
+		TORRENT_EXTRA_EXPORT dht_settings read_dht_settings(bdecode_node const& e);
```
arvidn: I would have thought these two functions should live under ``src/kademlia``. they could still be made public here with a ``using`` statement.

arvidn: oh, I see ``dht_settings`` doesn't seem to live in the ``dht`` namespace either..

src/kademlia/dht_state.cpp
```diff
@@ -0,0 +1,103 @@
...
+#if TORRENT_USE_IPV6
+		ret.nid6 = extract_node_id(e, "node-id6");
+#endif
+
+		TORRENT_TRY {
+			if (bdecode_node const nodes = e.dict_find_list("nodes"))
+				detail::read_endpoint_list<udp::endpoint>(nodes, ret.nodes);
+		} TORRENT_CATCH(std::exception&) {}
+#if TORRENT_USE_IPV6
+		TORRENT_TRY{
```
arvidn: I'm not sure about the try-catch here, what kinds of errors are you trying to suppress?

src/kademlia/dht_state.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+	node_id extract_node_id(bdecode_node const& e, string_view key)
+	{
+		if (e.type() != bdecode_node::dict_t) return node_id();
+		auto nid = e.dict_find_string_value(key);
+		if (nid.size() != 20) return node_id();
+		return node_id(nid);
+	}
+
+	void save_nodes(entry& ret, std::vector<udp::endpoint> const& v
```
arvidn: what would you think about returning the ``entry`` here? I'm pretty sure it's moveable (if not, I will make it moveable)

include/libtorrent/kademlia/dht_state.hpp
```diff
@@ -0,0 +1,81 @@
...
+	// implementation of the DHT. See BEP38_
+	//
+	// .. _BEP32: http://bittorrent.org/beps/bep_0032.html
+	struct TORRENT_EXPORT dht_state
+	{
+		dht_state() = default;
+		dht_state(dht_state const&) = default;
+		dht_state(dht_state&&) = default;
+		dht_state& operator=(dht_state const&) = default;
+		dht_state& operator=(dht_state&&) = default;
```
arvidn: I'm pretty sure if you don't specify any of these, you get the same behavior. (see [rule of zero](https://rmf.io/cxx11/rule-of-zero/))

src/kademlia/dht_state.cpp
```diff
@@ -0,0 +1,103 @@
...
+#endif
+
+		return ret;
+	}
+
+	entry save_dht_state(dht_state const& state)
+	{
+		entry ret(entry::dictionary_t);
+		ret["node-id"] = state.nid.to_string();
+		save_nodes(ret, state.nodes, "nodes");
```
arvidn: what about this idea, ``save_nodes`` returns an ``entry``. instead of passing in the name of the key you want it saved under, just assign it here: ``ret["nodes"] = save_nodes(state.nodes);``

would that work? (this of course assumes there is a move-assignment operator on ``entry``)

src/kademlia/dht_tracker.cpp
```diff
@@ -585,40 +557,31 @@ namespace libtorrent { namespace dht
...
-		std::string node;
-		std::back_insert_iterator<std::string> out(node);
-		write_endpoint(e.ep(), out);
-		n->list().push_back(entry(node));
+		auto v = static_cast<std::vector<udp::endpoint>*>(userdata);
+		v->push_back(e.ep());
 	}
 
-	void save_nodes(entry& ret, node const& dht, std::string const& key)
+	void save_nodes(std::vector<udp::endpoint>& ret, node const& dht)
```
arvidn: you wouldn't even need to use ``std::move()``, since it would be an implied rvalue reference in the ``return`` statement.

src/session.cpp
arvidn: I wonder if it would make sense for this function to take an ``error_code&``. I believe ``read_resume_data()``, does, and I feel like this is the same category of operation.

src/session_impl.cpp
```diff
@@ -5636,14 +5662,21 @@ namespace aux {
...
-		return m_dht->state();
+		return m_dht ? dht::save_dht_state(m_dht->state()) : entry();
 	}
 
 	void session_impl::start_dht_deprecated(entry const& startup_state)
 	{
 		m_settings.set_bool(settings_pack::enable_dht, true);
-		start_dht(startup_state);
+		std::vector<char> tmp;
+		std::back_insert_iterator<std::vector<char>> out(tmp);
```
arvidn: instead of creating this ``out`` object, you can just say: ``bencode(std::back_inserter(tmp), startup_state);``

test/test_dht.cpp
```diff
@@ -3375,6 +3376,44 @@ TORRENT_TEST(compare_ip_cidr)
 #endif
 }
 
+TORRENT_TEST(dht_state)
+{
+	dht_state s;
+
+	s.nid = to_hash("0000000000000000000000000000000000000001");
+	s.nodes.push_back(udp::endpoint(addr("1.1.1.1"), 1));
```
arvidn: there's a utility function called ``ep()``, which makes these kinds of addresses shorter and easier to read

test/test_dht.cpp
```diff
@@ -3375,6 +3376,44 @@ TORRENT_TEST(compare_ip_cidr)
...
+	// not important that IPv6 is disabled here
+	s.nid6 = to_hash("0000000000000000000000000000000000000002");
+	s.nodes6.push_back(udp::endpoint(addr("3.3.3.3"), 3));
+	s.nodes6.push_back(udp::endpoint(addr("4.4.4.4"), 4));
+
+	entry const e = save_dht_state(s);
+
+	std::vector<char> tmp;
+	std::back_insert_iterator<std::vector<char>> out(tmp);
+	bencode(out, e);
```
arvidn: here too, you could just use ``std::back_inserter(tmp)``

test/test_session_params.cpp
```diff
@@ -52,6 +56,13 @@ namespace
 		g_storage_constructor_invoked = true;
 		return dht_default_storage_constructor(settings);
 	}
+
+	sha1_hash to_hash(char const* s)
```
arvidn: I believe this utility function should already be available. it's possible though that it has been promoted to the general test library, in which case we can coalesce them later

src/kademlia/dht_state.cpp
aldenml: no idea, that was a copy paste, I will remove it

include/libtorrent/kademlia/dht_state.hpp
aldenml: thanks for the link!

test/test_dht.cpp
aldenml: well, `ep()` is to create a `tcp::endpoint`, I will create `uep()`

arvidn: oh, right..

test/test_primitives.cpp
```diff
@@ -41,13 +41,6 @@ POSSIBILITY OF SUCH DAMAGE.
 
 using namespace libtorrent;
 
-sha1_hash to_hash(char const* s)
```
arvidn: thanks!

PR #1101, opened 2016-09-17
replace bitfield in torrent with std::bitset

arvidn:


arvidn:
I'm not sure if this is a good change. I'm trying it on. In some sense I feel like bitfields are second class citizen in c++, but on the other hand this is more verbose and less convenient in debuggers.

aldenml:
it certainly doesn't feel like a much better code, maybe with another abstraction to hide some ugliness, but C++ lacks uniform access as a language, so you need to deal with the `()`

include/libtorrent/torrent.hpp
```diff
@@ -164,10 +165,180 @@ namespace libtorrent
 		std::vector<char> restart_piece;
 	};
 
+	namespace flags {
```
aldenml: just curious, is this a common pattern? why not use `enum class`?

arvidn: and enum class would not allow using the enums as indices in the bitfield without an explicit cast to ``int``, and since the flags shared names with a lot of other member functions in torrent, so they can't be in the same scope.

PR #1107, opened 2016-09-18
dht code related refactor and cleanup

aldenml:


aldenml:
I'm having problems figuring out a good API design to be able to serialize the "dht storage". I'm thinking of something very particular (and minimal) to the default one, but at the same time, easy to use at `session` construction time.

arvidn:
this patch doesn't have any actual behavior change, right? at least I can't see anything obvious. lgtm

aldenml:
any change, just refactor and cleanup. The comment was in order to get any idea (if you happen to have one)

PR #1112, opened 2016-09-19
Fix warning C4373

pavel-pimenov:
libtorrent\src\smart_ban.cpp(77): warning C4373: 'libtorrent::`anonymous-namespace'::smart_ban_plugin::on_piece_pass': virtual function overrides 'libtorrent::torrent_plugin::on_piece_pass', previous versions of the compiler did not override when parameters only differed by const/volatile qualifiers


arvidn:
I would prefer to suppress this warning by asking the compiler to disable it. if you take a look at the Jamfile you'll see ``<toolset>msvc:<cflags>/wd4373``. If you're using cmake and you see this warning, would you mind patching the ``CMakeFiles.txt`` instead?

arvidn:
the reason I prefer to suppress these warnings as compiler flags or pragmas is that it's really nice to make everything immutable be ``const``

arvidn:
should I close this or do you intend to patch the cmake file?

PR #1115, opened 2016-09-19
support for load/save of dht storage state

aldenml:
What do you think about this idea?

### DIFF
--- a/include/libtorrent/kademlia/dht_storage.hpp
+++ b/include/libtorrent/kademlia/dht_storage.hpp
@@ -216,6 +216,10 @@ namespace dht
 
 		virtual dht_storage_counters counters() const = 0;
 
+		virtual void load_state(std::vector<char> const& state) = 0;
+
+		virtual std::vector<char> save_state() const = 0;
+
 		virtual ~dht_storage_interface() {}
 	};
 
--- a/src/kademlia/dht_state.cpp
+++ b/src/kademlia/dht_state.cpp
@@ -63,6 +63,15 @@ namespace
 	}
 } // anonymous namespace
 
+	void dht_state::clear()
+	{
+		nid.clear();
+		nid6.clear();
+		nodes.clear();
+		nodes6.clear();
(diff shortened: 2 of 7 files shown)


include/libtorrent/kademlia/dht_storage.hpp
```diff
@@ -216,6 +216,10 @@ namespace dht
 
 		virtual dht_storage_counters counters() const = 0;
 
+		virtual void load_state(std::vector<char> state) = 0;
```
aldenml: missing const& here

src/kademlia/dht_state.cpp
```diff
@@ -81,6 +90,13 @@ namespace
 			ret.nodes6 = detail::read_endpoint_list<udp::endpoint>(nodes);
 #endif
 
+		if (bdecode_node const st = e.dict_find_string("storage-state"))
+		{
+			char const* ptr = st.string_ptr();
```
arvidn: this is meant to be a "blob" defined by the storage implementation? would there need to be protection from passing in the blob from another storage class?

arvidn:
What kind of data would you expect a storage implementation to save here? What it stores internally is quite well defined, right? it can only be peer lists and data blobs

src/kademlia/dht_state.cpp
aldenml: Yes, it will be up to the implementors protect from faulty "blob"s.

aldenml:
rebased for a complete review

aldenml:
I understand this is a kind of weak API. Do you have an idea of how to improve this?

aldenml:
Closing this for a fresh start

arvidn:
sorry for not responding. I've been kind of busy the last week.

aldenml:
please, no need :)

I do want to have something robust that clients can use by default, but I don't think it's a good idea to have internal types in that "state" like the `bloom_filter`

PR #1120, opened 2016-09-19
use string_view in create_torrent

arvidn:


src/create_torrent.cpp
```diff
@@ -75,7 +76,7 @@ namespace libtorrent
 			return 0;
 #else
 			struct stat s;
-			if (lstat(convert_to_native(p).c_str(), &s) < 0) return 0;
+			if (lstat(convert_to_native(std::string(p)).c_str(), &s) < 0) return 0;
```
arvidn: is there a ``to_string()`` for ``string_view``?

aldenml: yes :) and you already use it https://github.com/arvidn/libtorrent/blame/master/src/kademlia/node.cpp#L779

arvidn: ah, it's not a free function..

src/create_torrent.cpp
```diff
@@ -665,28 +666,28 @@ namespace libtorrent
 		return dict;
 	}
 
-	void create_torrent::add_tracker(std::string const& url, int tier)
+	void create_torrent::add_tracker(string_view url, int const tier)
 	{
-		m_urls.push_back(announce_entry(url, tier));
+		m_urls.push_back(announce_entry(std::string(url), tier));
```
aldenml: here?

PR #1122, opened 2016-09-20
new peers storage policy

ssiloti:
This is based on the comment by @the8472 at https://github.com/bittorrent/bittorrent.org/pull/41#issuecomment-244620725 about how to handle get_peers/announce_peer.

### DIFF
--- a/src/kademlia/dht_storage.cpp
+++ b/src/kademlia/dht_storage.cpp
@@ -189,9 +189,8 @@ namespace
 			, bool const noseed, bool const scrape
 			, entry& peers) const override
 		{
-			auto const i = m_map.lower_bound(info_hash);
-			if (i == m_map.end()) return false;
-			if (i->first != info_hash) return false;
+			auto const i = m_map.find(info_hash);
+			if (i == m_map.end()) return int(m_map.size()) >= m_settings.max_torrents;
 
 			torrent_entry const& v = i->second;
 
@@ -257,7 +256,7 @@ namespace
 					++m;
 				}
 			}
-			return true;
+			return int(i->second.peers.size()) >= m_settings.max_peers;
 		}
 
 		void announce_peer(sha1_hash const& info_hash
--- a/src/kademlia/dht_storage.cpp
+++ b/src/kademlia/dht_storage.cpp
@@ -186,7 +186,7 @@ namespace
 		}
 
 		bool get_peers(sha1_hash const& info_hash, udp const protocol
-			, bool const noseed, bool const scrape
+			, bool const noseed, bool const scrape, address const& requester
 			, entry& peers) const override
 		{
 			auto const i = m_map.find(info_hash);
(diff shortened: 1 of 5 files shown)


src/kademlia/dht_storage.cpp
```diff
@@ -257,7 +256,7 @@ namespace
 					++m;
 				}
 			}
-			return true;
+			return int(i->second.peers.size()) >= m_settings.max_peers;
```
arvidn: it's not obvious to me that this makes sense...

arvidn: i.e. I would have expected ``return false`` here

src/kademlia/dht_storage.cpp
```diff
@@ -268,26 +267,12 @@ namespace
 			torrent_entry* v;
 			if (ti == m_map.end())
 			{
-				// we don't have this torrent, add it
-				// do we need to remove another one first?
-				if (!m_map.empty() && int(m_map.size()) >= m_settings.max_torrents)
+				if (int(m_map.size()) >= m_settings.max_torrents)
```
arvidn: would it make sense to prioritize info hashes closer to our node iD?

src/kademlia/node.cpp
```diff
@@ -859,7 +858,9 @@ void node::incoming_request(msg const& m, entry& e)
 		bool scrape = false;
 		if (msg_keys[1] && msg_keys[1].int_value() != 0) noseed = true;
 		if (msg_keys[2] && msg_keys[2].int_value() != 0) scrape = true;
-		lookup_peers(info_hash, reply, noseed, scrape);
+		bool full = lookup_peers(info_hash, reply, noseed, scrape);
```
arvidn: I think this would be a good place to copy-paste the the8472's comment, or paraphrase

src/kademlia/dht_storage.cpp
ssiloti: It definitely shouldn't be false, we want to withhold a write token if the torrent is at the max number of peers. But now that you've made me think about it again I've realized there's a problem, even if the torrent is at max_peers we need to send a token if the requester is already in the peer set.

src/kademlia/dht_storage.cpp
the8472: does capping by # of torrents even make sense? a torrent could have a single or thousands of peers announcing.

arvidn: I'm not sure. iirc, I did that just for simplicity. There is a certain cost just to have the torrent entry, so they should probably be capped to something, but the primary limit should probably be the total number of peers

ssiloti: It doesn't seem worth it to prioritize info hashes by distance. Unlike the number of peers per info hash, the distribution of info hashes themselves is of course uniform so the number of info hashes per node shouldn't vary by much. Whatever form the resource limit takes, it shouldn't be set so low as to cause a node to reject legit announces for a new info hash. Only an attacker should be getting announces for new info hashes rejected.

arvidn:
thanks steven!

PR #1123, opened 2016-09-20
allow for port 0 in listen_interfaces and logs/refactor

aldenml:
two commits here.

any idea of what's happening with appveyor?

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -1441,11 +1441,19 @@ namespace aux {
 				&& pack.get_str(settings_pack::listen_interfaces)
 					!= m_settings.get_str(settings_pack::listen_interfaces));
 
+#ifndef TORRENT_DISABLE_LOGGING
+		session_log("applying settings pack, init=%s, reopen_listen_port=%s"
+			, init ? "true" : "false", reopen_listen_port ? "true" : "false");
+#endif
+
 		apply_pack(&pack, m_settings, this);
 		m_disk_thread.set_settings(&pack, m_alerts);
 
-		if (init)
+		if (init && !reopen_listen_port)
+		{	// no need to call this if reopen_listen_port is true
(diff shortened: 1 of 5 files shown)


arvidn:
no idea about appveyor. their status page says everything is working. they did report elevated error rates from github on the 18th though, perhaps they're still catching up on the backlog they caused.

src/session_impl.cpp
```diff
@@ -1652,7 +1660,8 @@ namespace aux {
 			}
 
 			if (ec == error_code(error::address_in_use)
-				&& !(flags & listen_no_system_port))
+				&& !(flags & listen_no_system_port)
+				&& bind_ep.port() != 0)
```
arvidn: this is the key part of the patch, right?

src/string_util.cpp
```diff
@@ -279,7 +278,7 @@ namespace libtorrent
 			else
 			{
 				iface.port = std::atoi(port.c_str());
-				if (iface.port <= 0 || iface.port > 65535) iface.port = -1;
+				if (iface.port < 0 || iface.port > 65535) iface.port = -1;
```
arvidn: and here I suppose

arvidn:
I can't think of why this shouldn't be allowed, and it definitely used to be intended to be allowed.

src/string_util.cpp
aldenml: I separated the commits to easy the review (but it's trivial), it's that the PR queue is getting bigger

aldenml:
@arvidn I think you should cancel all the appveyor works, I can trigger new ones in my account

arvidn:
your account works?

aldenml:
without problems

aldenml:
maybe cancelling all the jobs first do the trick

PR #1127, opened 2016-09-21
std::count -> std::find

pavel-pimenov:
V812. Decreased performance. Ineffective use of the 'count' function. It can possibly be replaced by the call to the 'find' function


arvidn:
lgtm

PR #1128, opened 2016-09-21
remove redundant protocol parameter

ssiloti:
I realized last night that the requester parameter added in #1122 made the protocol parameter redundant. I even got up early this morning to add this change to that PR but arvid beat me to it. I blame his time zone advantage :)


arvidn:
heh, actually, I'm on the east coast for CPPCON right now. I still have a jet-lag advantage though :P

PR #1130, opened 2016-09-21
Workaround for python3 in setup.py

stuaxo:


arvidn:
I'm not a python expert, but if this function was removed in python 3, is there a better alternative that's available in 2 that should be used instead?

arvidn:
@cas-- does this look reasonable to you?

cas--:
It's overcomplicated, using exec means version check is unneeded so replace the execfile with either of these below. As it's simply running setup.py I'd be tempted to go with the latter single line version:
```
with open('setup.py) as filename:
    exec(filename.read())
```
or
```
exec(open('setup.py').read())
```

stuaxo:
Updated to use the first version.

PR #1131, opened 2016-09-21
refactor in msg verify_message and minor cleanup

aldenml:


src/kademlia/item.cpp
```diff
@@ -66,15 +66,14 @@ namespace
 		{
 			ptr += std::snprintf(ptr, left, "4:salt%d:", int(salt.size()));
 			left = out.size() - (ptr - out.data());
-			std::memcpy(ptr, salt.data(), (std::min)(salt.size(), left));
-			ptr += (std::min)(salt.size(), left);
```
arvidn: just as a reminder, had this been in a header file, it would not be safe..

src/kademlia/msg.cpp
```diff
@@ -36,11 +36,13 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent { namespace dht {
 
-namespace dht_detail {
-
-bool verify_message(bdecode_node const& message, key_desc_t const desc[]
-	, bdecode_node ret[], int size, char* error, int error_size)
+bool verify_message_impl(bdecode_node const& message, span<key_desc_t const> desc
```
arvidn: I was looking at making this exact change a while ago. I can't remember why I gave up.

include/libtorrent/kademlia/msg.hpp
```diff
@@ -82,25 +82,17 @@ struct key_desc_t
...
 
 // verifies that a message has all the required
 // entries and returns them in ret
 template <int Size>
 bool verify_message(bdecode_node const& msg, key_desc_t const (&desc)[Size]
-	, bdecode_node (&ret)[Size], char* error, int error_size)
+	, bdecode_node (&ret)[Size], span<char> error)
 {
-	return dht_detail::verify_message(msg, desc, ret, Size, error, error_size);
+	return verify_message_impl(msg, desc, ret, error);
```
arvidn: ah, right. the killer would be to actually use the version that takes a span instead of this wrapper.. I guess that might be a bit more work

arvidn:
lgtm

PR #1133, opened 2016-09-22
fix peer picking algorithm

ssiloti:
The old code was always picking the first to_pick peers from the set.

### DIFF
--- a/src/kademlia/dht_storage.cpp
+++ b/src/kademlia/dht_storage.cpp
@@ -214,18 +214,15 @@ namespace
 			else
 			{
 				tcp const protocol = requester.is_v4() ? tcp::v4() : tcp::v6();
-				int max = m_settings.max_peers_reply;
+				int to_pick = m_settings.max_peers_reply;
 				// if these are IPv6 peers their addresses are 4x the size of IPv4
 				// so reduce the max peers 4 fold to compensate
 				// max_peers_reply should probably be specified in bytes
 				if (!v.peers.empty() && protocol == tcp::v6())
-					max /= 4;
-				// we're picking "to_pick" from a list of "num" at random.
-				int const to_pick = std::min(int(v.peers.size()), max);
-				std::set<peer_entry>::const_iterator iter = v.peers.begin();
+					to_pick /= 4;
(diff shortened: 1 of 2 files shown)


src/kademlia/dht_storage.cpp
```diff
@@ -235,26 +232,20 @@ namespace
 					if (iter->addr.protocol() != protocol)
```
arvidn: there are these conditions here of peers that are excluded. afaict, these filters will skew the picking and make us potentially pick too few.

arvidn:
nice catch

arvidn:
I would imagine this would be a problem in 1.1 too..

src/kademlia/dht_storage.cpp
ssiloti: Yeah, I realized this after I went to bed last night. Note to self: always sleep on a patch before submitting a PR.

I'm debating splitting the peers set into separate IPv4 and IPv6 sets so that the max_peers limit is applied to each one separately. It would be bad to not have any IPv6 peers because the set is full of IPv4 peers or vice-versa. I'm not sure that would ever be a problem in practice though.

arvidn:
I imagine this is still a step in the right direction. The obvious way of fixing this (I think) would be to have a separate pass first that counts the filtered peers. but maybe there could be a persistent round-robin cursor that just picks peers in order, and every time it wraps, it shuffles the list

arvidn:
do you think I should merge this or are you still working on this?

ssiloti:
No don't merge it. At the very least I'm going to add a pass to count the filtered peers before picking. I'm also going to try out the cursor approach to see how ugly it gets. Using two std::vectors to store v4 and v6 peers would be significantly more memory efficient than std::set and may even be faster overall due to better cache locality.

ssiloti:
The cursor approach started looking kinda ugly, I think stateless sampling is fast enough and less error prone.

ssiloti:
I backported the unit test to 1.1 and the problem is present there too even though the code it a bit different. I'm going to rebase just the first commit in this PR to 1.1 then rebase this PR once 1.1 is merged back into master.

PR #1134, opened 2016-09-22
run undefined sanitier on travis/linux

arvidn:


aldenml:
I was thinking that with the changes you are making in the ed25519 source code, you may want to back port the vectors test from master.

arvidn:
yeah, I should probably do that

arvidn:
@aldenml do you understand why the osx build is failing in my branch, but not in RC_1_1?
I tried adding the ``osx_image`` to .travis, but it still was failing the release build.

aldenml:
no idea, but in the past, the "cache clean" in travis solved some problems

arvidn:
good point

Jamfile
```diff
@@ -505,7 +505,7 @@ feature export-extra : off on : composite propagated ;
 variant test_release : release
 	: <asserts>production <debug-symbols>on
 	  <invariant-checks>full <boost-link>shared <optimization>off
-	  <export-extra>on <debug-iterators>on <threading>multi
+	  <export-extra>on <debug-iterators>on <threading>multi <inlining>on
```
aldenml: I suppose that `boost-build` was upgraded then, it's `off` in master for that reason

arvidn: I'm not entirely sure why this turned out to be a problem now, but basically there used to be a bug in boost-build where it passed on ``-finline-functions`` to clang when ``<inlining>full``. This was fixed last year, but hasn't propagated to the package on travis. In combination with this, clang used to ignore unknown arguments, but recently started to print a warning. Since I build with ``-Werror``, that warning turned into an error

PR #1138, opened 2016-09-23
vector<T> -> vector<T> const&

pavel-pimenov:


src/torrent_info.cpp
```diff
@@ -1604,7 +1604,7 @@ namespace libtorrent
 			, auth, extra_headers));
 	}
 
-	void torrent_info::set_web_seeds(std::vector<web_seed_entry> seeds)
+	void torrent_info::set_web_seeds(std::vector<web_seed_entry> const& seeds)
```
arvidn: this will inhibit the vector being moved in, and will force a copy

src/torrent_info.cpp
```diff
@@ -1604,7 +1604,7 @@ namespace libtorrent
...
 	{
 		m_web_seeds = seeds;
```
arvidn: instead, this should say: ``m_web_seeds = std::move(seeds);``

arvidn:
also, please rebase on top of lastest master to fix the build

PR #1139, opened 2016-09-23
Fix string -> char * -> string.

pavel-pimenov:


src/file_storage.cpp
```diff
@@ -572,7 +572,7 @@ namespace libtorrent
 		else
 		{
 			if (m_files.empty())
-				m_name = split_path(path).c_str();
```
arvidn: it's actually correct as it is. and it's an unfortunate complication. ``split_path()`` will actually replace directory separators with '\0', and we just want the first, root, directory. So we want to interpret the string as a 0-terminated to get that. It could be done more efficiently with a function just extracting the root directory.

arvidn: a separate function would also be less confusing. patches welcome :)

PR #1141, opened 2016-09-23
fix peer picking algorithm

ssiloti:
The old code was always picking the first to_pick peers from the set.

This is going to conflict heavily when merging RC_1_1 to master.

### DIFF
--- a/src/kademlia/dht_storage.cpp
+++ b/src/kademlia/dht_storage.cpp
@@ -241,22 +241,28 @@ namespace
 			}
 			else
 			{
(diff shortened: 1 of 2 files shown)


arvidn:
test_dht_storage.cpp:345:73: error: invalid initialization of non-const reference

https://travis-ci.org/arvidn/libtorrent/jobs/162253662#L782

ssiloti:
oops, I had that fixed locally but I forgot to add it to the commit. No really I swear :)

PR #1143, opened 2016-09-23
considering both generic and system in close reason conversion

aldenml:


arvidn:
the proper way to do this is to use an ``error_condition`` and compare those. and that should probably be done more pervasively in the code base

aldenml:
closing this, this is only a workaround patch, but the `error_code` overall handling as it's right now is very dangerous.

src/close_reason.cpp
```diff
@@ -146,8 +138,10 @@ namespace libtorrent
 					return close_no_reason;
 			}
 		}
-		else if (ec.category() == boost::system::system_category())
-		{
+		else if (ec.category() == boost::system::system_category()
+			|| ec.category() == boost::system::generic_category())
+		{	// need to test both categories due to possible boost
```
arvidn: is this caused by boost versions? I would have expected that unix systems would use ``system_category`` and windows ``generic_category``. either way, the intention here was most likely to test for ``generic_category`` to begin with. Actually, would you mind updating this patch to just compare against ``generic_category``? (there's a typedef in the ``libtorrent`` namespace you can use). I should probably look over other places where I compare ``error_codes`` directly.

arvidn:
I understand it's error prone to compare ``error_codes`` directly. Do you know of any place where getting this wrong would lead to a serious error? i.e. segmentation fault, buffer overrun etc.

aldenml:
Not in the current form, but this comment (https://github.com/arvidn/libtorrent/blob/master/simulation/test_http_connection.cpp#L135) is an example of the potential problems.

btw, there are a few `TEST_EQUAL(ec, error_code())`.

PR #1144, opened 2016-09-23
Python bindings' fixes and addition of missing alert attributes

kaihu:


bindings/python/src/alert.cpp
```diff
@@ -441,7 +463,7 @@ void bind_alert()
 
     class_<listen_succeeded_alert, bases<alert>, noncopyable>(
         "listen_succeeded_alert", no_init)
-        .def_readonly("endpoint", &listen_succeeded_alert::endpoint)
```
arvidn: did the ``def_readonly`` not work? (if so, what error did you experience?). Or do you need to be able to write to this field?

arvidn: the reason I ask is because it's a lot more scalable to make sure the one converter for this type works, and then just bind the members, than to write a separate getter function for every type that uses it.

arvidn:
if any of these alerts are simple to trigger in ``test.py``, that might be a good sanity check that some of the fields work too.

arvidn:
I've fixed all the missing fields

PR #1145, opened 2016-09-23
modernize utf8 conversion functions to throw and move

arvidn:


include/libtorrent/utf8.hpp
```diff
@@ -42,35 +42,43 @@ POSSIBILITY OF SUCH DAMAGE.
...
-		// insuff. room in target for conversion
-		target_exhausted,
+		// hidden
+		TORRENT_EXPORT error_code make_error_code(error_code_enum e);
+	}
 
-		// source sequence is illegal/malformed
-		source_illegal
-	};
+	TORRENT_EXPORT boost::system::error_category& get_utf8_category();
```
aldenml: while you are here, I think the `get_` prefix is deprecated

include/libtorrent/string_view.hpp
```diff
@@ -41,11 +41,13 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <boost/utility/string_ref.hpp>
 namespace libtorrent {
 using string_view = boost::string_ref;
+using wstring_view = boost::wstring_ref;
```
aldenml: I think this should exists only in `#if TORRENT_USE_WSTRING || defined TORRENT_WINDOWS`

src/utf8.cpp
```diff
@@ -174,28 +176,103 @@ namespace libtorrent
...
-	utf8_conv_result_t utf8_wchar(std::string const& utf8, std::wstring &wide)
+	namespace utf8_errors
+	{
+		boost::system::error_code make_error_code(utf8_errors::error_code_enum e)
+		{
+			return error_code(e, get_utf8_category());
+		}
+	} // utf_errors namespace
+
+	boost::system::error_category& get_utf8_category()
```
aldenml: `const`?

include/libtorrent/string_view.hpp
aldenml: ping here, I think it's important

aldenml: "important" here in a relative sense...I think it's important mostly for code consistency. Also, this type does not exist in the C++ standard (the future one).

btw, I think that all the Unicode related code should be configurable, but not exposed in ABI

arvidn: this is really just a short-hand for the future ``std::wstring_view`` (which as far as I can tell i part of C++17 at this point). http://en.cppreference.com/w/cpp/string/basic_string_view

I don't really consider it a part of libtorrent in that sense, similar to ``error_code`` will be removed from libtorrent as soon as everyone can agree on using ``std::error_code``.

aldenml: ignore my comment then, just saw `std::basic_string_view<wchar_t>` in my search but not the alias type.

looking forward for the big move to `std::error_code`!! (and maybe use `asio` without boost)

aldenml: btw, at this moment, only `boost.asio` is the one that requieres `boost.system` binary dependency, all the rest is using header only parts.

PR #1146, opened 2016-09-24
make connection_type enum an enum class

arvidn:


include/libtorrent/peer_connection_handle.hpp
```diff
@@ -57,7 +58,7 @@ struct TORRENT_EXPORT peer_connection_handle
 		: m_connection(impl)
 	{}
 
-	int type() const;
+	connection_type type() const;
```
aldenml: I think this could be a problem since `connection_type` is a private type

arvidn: hm.. maybe it shouldn't be..

arvidn: actually, technically it is public. it's not in the ``aux`` namespace nor in the ``aux_`` directory. Once ``peer_connection.hpp`` gets moved, this type should remain though

aldenml: now I need to tweak the swig file :)

PR #1149, opened 2016-09-24
add proper error handling to peer connections and session

arvidn:
i.e. catch exceptions in asio handlers

### DIFF
--- a/include/libtorrent/aux_/allocating_handler.hpp
+++ b/include/libtorrent/aux_/allocating_handler.hpp
@@ -63,7 +63,7 @@ namespace libtorrent { namespace aux
 
 	struct error_handler_interface
 	{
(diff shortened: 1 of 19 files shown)


include/libtorrent/aux_/allocating_handler.hpp
```diff
@@ -61,6 +61,12 @@ namespace libtorrent { namespace aux
 		handler_storage(handler_storage const&);
 	};
 
+	struct error_handler_interface
+	{
+		virtual void on_exception(std::exception&) = 0;
```
aldenml: any reason this is not `const&`?

arvidn: it should probably be a const reference, just like the catch statements probably should be const reference as well

PR #1151, opened 2016-09-25
added support for save/load of dht storage data

aldenml:
What do you think about this API?

### DIFF
--- a/include/libtorrent/kademlia/dht_storage.hpp
+++ b/include/libtorrent/kademlia/dht_storage.hpp
@@ -34,6 +34,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #define TORRENT_DHT_STORAGE_HPP
 
 #include <functional>
+#include <vector>
+#include <array>
 
 #include <libtorrent/kademlia/node_id.hpp>
 #include <libtorrent/kademlia/types.hpp>
@@ -42,6 +44,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <libtorrent/address.hpp>
 #include <libtorrent/span.hpp>
 #include <libtorrent/string_view.hpp>
+#include <libtorrent/time.hpp>
 
 namespace libtorrent
 {
@@ -62,6 +65,44 @@ namespace dht
 
 		// This member function set the counters to zero.
 		void reset();
+	};
+
+	struct announce_peer_data
+	{
+		time_point added;
+		tcp::endpoint addr;
+		bool seed;
+	};
+
+	struct announce_data
+	{
+		sha1_hash info_hash;
+		std::string name;
+		std::vector<announce_peer_data> peers4;
+		std::vector<announce_peer_data> peers6;
+	};
+
+	struct dht_immutable_data
+	{
+		sha1_hash target;
+		std::vector<char> value;
+		std::array<char, 128> ips;
+		time_point last_seen;
+	};
+
+	struct dht_mutable_data : dht_immutable_data
+	{
+		std::array<char, signature::len> sig;
+		std::uint64_t seq;
+		std::array<char, public_key::len> key;
+		std::string salt;
+	};
+
+	struct dht_storage_data
+	{
+		std::vector<announce_data> torrents;
+		std::vector<dht_immutable_data> immutable_items;
+		std::vector<dht_mutable_data> mutable_items;
 	};
 
 	// The DHT storage interface is a pure virtual class that can
@@ -215,6 +256,10 @@ namespace dht
 		virtual void tick() = 0;
 
 		virtual dht_storage_counters counters() const = 0;
+
+		virtual dht_storage_data save_data() const = 0;
+
+		virtual void load_data(dht_storage_data const& data) = 0;
 
 		virtual ~dht_storage_interface() {}
 	};
--- a/include/libtorrent/kademlia/dht_storage.hpp
+++ b/include/libtorrent/kademlia/dht_storage.hpp
@@ -67,21 +67,6 @@ namespace dht
 		void reset();
 	};
 
-	struct announce_peer_data
-	{
-		time_point added;
-		tcp::endpoint addr;
-		bool seed;
-	};
-
-	struct announce_data
-	{
-		sha1_hash info_hash;
-		std::string name;
-		std::vector<announce_peer_data> peers4;
-		std::vector<announce_peer_data> peers6;
-	};
-
 	struct dht_immutable_data
 	{
 		sha1_hash target;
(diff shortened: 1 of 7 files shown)


ssiloti:
What's the motivation for this API? It seems like persistence should be left for each storage implementation to handle in it's own way. Whatever setup is needed should be taken care of by the factory function.

aldenml:
My motivation is that I would like to have minimal support for this directly from `libtorrent`. With the typical `dht_settings` (max items) this should be of no big impact at the `session` state and if it becomes widely used(because it comes already in libtorrent) it should be beneficial for the entire DHT.

I expect a possible "sqlite" or "AWS" storage implementation just not doing anything here, but I doubt this implementations will be popular.

Maybe I'm wrong in my assumptions, please, let me know.

aldenml:
I forgot to mention/clarify that at the end, this will be serialized in the "session::save_state" and it will be recovered with the function `read_session_params`

ssiloti:
These calls shouldn't be part of the public API then since they probably won't make sense for alternative storage implementations.

aldenml:
I'm not sure about your argument, plugins interfaces or not usually intended to be fully implemented. Do you have any suggestion of how to accomplish this? I compile with "no-rtti"

I suppose I could do a lot of gymnastics with the plugins, but it will feels like over engineering the issue.

ssiloti:
So build with RTTI enabled. Have you tested how much memory you save by omitting RTTI? I bet it isn't much, libtorrent doesn't instantiate that many classes with vtables.

arvidn:
I think what steven means is that the client that injects a custom dht storage can still hold on to a reference to it, and make whatever calls it likes. Such as loading it with data and extracting data from it to persist somewhere

aldenml:
Yes, I understand, but my intention is having this to everybody, just the minimal. And now is when I might be wrong. Isn't good if all `libtorrent` clients persist/load the ~2MB of the "DHT store" items?

ssiloti:
Frankly, I don't think it would make much of a difference. The DHT is designed with the assumption that both nodes and their storage are ephemeral. It would help to some degree of course, but I doubt anyone would notice.

aldenml:
wow...I don't know what to say :D all this time I was thinking that it would be of great benefit to the entire network, specially for retrieving old items from the DHT.

ssiloti:
It would be a bit like going back and seeding a torrent which had been dead for a while. Anyone who cared would have already moved on and the chances of someone coming across it by chance would be minuscule.

aldenml:
Ok, now I remember the problem I was facing. A long time ago I was thinking about a Pub/Sub solution (yes, another "distributed twitter" here) specially for mobile. But in the mobile, both Android and iOS, the app is shutdown by the system very aggressively.

I know that mobile in carries should move to the read-only state, but the restart happens even in a Wi-Fi network. I'm mostly concerned about the "DHT store" (no torrents) and then I reasoned that a good way to mitigate the problem is having a way to persist the items, then I created the PR for the "dht_storage".

I see a great shift to mobile devices and if the DHT does not accommodate for this new devices, the network will suffer.

ssiloti:
A device which is only online intermittently is not useful for DHT storage and will see very few incoming puts anyways.

Building a DHT out of nodes which are online less than 50% of the time is a radically more difficult problem than what the existing DHT was designed to solve. Adding a band-aid here and there is not going to get you a solution.

arvidn:
it doesn't seem like a bad idea to me to persist peers and dht blobs. perhaps peers/torrents should expire, but being able to restart the dht node and picking up the whole state seems useful and it doesn't seem like the cost is particularly high.

aldenml:
@ssiloti I hope you don't mind if I ping @the8472 here for a comment.
@arvidn does it means that this PR is not "discardable"?

the8472:
I don't have much to add. Persisting BEP44 values is desirable, but i don't think mobile clients are a good use-case.

Do phones even keep their IP in idle mode?

aldenml:
Thanks for your comment. Phones while in Wi-Fi keep their IP in the same way PCs do, the more important in this case is the external NAT one, in the carrier network I have no idea. Now that you mention it, I have some idea to keep a list of "old_node_ids", in order to reuse them if the device move between a few NATs, but that's for another PR.

I will change this to remove the torrents data.

aldenml:
Do you think this is a more acceptable API?

include/libtorrent/kademlia/dht_storage.hpp
```diff
@@ -64,6 +67,28 @@ namespace dht
...
+	{
+		sha1_hash target;
+		std::vector<char> value;
+		std::array<char, 128> ips;
+		time_point last_seen;
+	};
+
+	struct dht_mutable_data : dht_immutable_data
+	{
+		std::array<char, signature::len> signature;
```
arvidn: aren't there public typedefs for signatures and keys?

aldenml: yes, well they are full types (in `types.hpp`), but not sure if you want to put this in the `dht_state`, because then, the question will be about the `bloom_filter`.

src/kademlia/dht_storage.cpp
```diff
@@ -505,6 +505,20 @@ namespace
...
+		dht_storage_items save_items(int const /*max_items*/) const override
+		{
+			dht_storage_items ret;
+
+			// TODO: implement here
+
+			return ret;
+		}
+
+		void load_items(dht_storage_items const& /*data*/) override
```
arvidn: I would imagine these items could be moved in here. the caller of this function won't pass these items to anyone else, right?

arvidn: i.e. this could be by value

include/libtorrent/kademlia/dht_storage.hpp
```diff
@@ -64,6 +67,28 @@ namespace dht
 		void reset();
 	};
 
+	struct dht_immutable_data
+	{
+		sha1_hash target;
+		std::vector<char> value;
+		std::array<char, 128> ips;
+		time_point last_seen;
```
arvidn: this time point is not wall clock time. a ``time_t`` might make more sense.

src/kademlia/dht_storage.cpp
aldenml: I think you are right, but I need to check. That could solve the clear of the internal object in `session_impl`

aldenml:
Hi @arvidn, appreciate if you can take a look at the "movable" part. I would like to make `dht_state` pass by value, but it's used twice in `session::start_dht`. Thanks!

aldenml:
Closing this in favor of https://github.com/arvidn/libtorrent/pull/1170

PR #1152, opened 2016-09-25
compare with generic_category in close_reason

aldenml:


arvidn:
thank you! does this solve the problem you were seeing before as well?

aldenml:
the problem I was having was a monumental mistake I was making in my "posix wrapper". In Mac OS X I was calling "stat" instead of "stat64".

PR #1154, opened 2016-09-26
refactor to expose protocol_descriptor

aldenml:
This is (at least for me) a more clear code to work with and I think it will help for future refactors.

### DIFF
--- a/include/libtorrent/kademlia/node.hpp
+++ b/include/libtorrent/kademlia/node.hpp
@@ -90,6 +90,27 @@ protected:
 	~udp_socket_interface() {}
 };
 
+struct TORRENT_EXTRA_EXPORT protocol_descriptor
+{
+	protocol_descriptor(udp protocol)
+		: m_protocol(protocol) {}
+
+	udp protocol() const;
+	char const* family_name() const;
+	char const* nodes_key() const;
+
+	bool is_v4() const;
+	bool is_v6() const;
+
+	bool is_native(udp::endpoint const& ep) const;
+	bool is_native(address const& addr) const;
+
+	operator udp() const { return m_protocol; }
+
+private:
+	udp const m_protocol;
+};
+
 class TORRENT_EXTRA_EXPORT node : boost::noncopyable
 {
 public:
@@ -196,19 +217,7 @@ public:
 
 	dht_observer* observer() const { return m_observer; }
 
-	udp protocol() const { return m_protocol.protocol; }
-	char const* protocol_family_name() const { return m_protocol.family_name; }
-	char const* protocol_nodes_key() const { return m_protocol.nodes_key; }
-
-	bool native_address(udp::endpoint const& ep) const
-	{ return ep.protocol().family() == m_protocol.protocol.family(); }
-	bool native_address(tcp::endpoint const& ep) const
-	{ return ep.protocol().family() == m_protocol.protocol.family(); }
-	bool native_address(address const& addr) const
-	{
-		return (addr.is_v4() && m_protocol.protocol == m_protocol.protocol.v4())
-			|| (addr.is_v6() && m_protocol.protocol == m_protocol.protocol.v6());
-	}
+	protocol_descriptor protocol() const { return m_protocol; }
 
 private:
 
@@ -237,30 +246,12 @@ public:
 	rpc_manager m_rpc;
 
 private:
-#ifdef _MSC_VER
-#pragma warning(push)
-// warning: default constructor could not be generated
-#pragma warning(disable: 4510)
-// warning: struct can never be instantiated
-#pragma warning(disable: 4610)
-#endif
-	struct protocol_descriptor
-	{
-		udp protocol;
-		char const* family_name;
-		char const* nodes_key;
-	};
-
-#ifdef _MSC_VER
-#pragma warning(pop)
-#endif
-	static protocol_descriptor const& map_protocol_to_descriptor(udp protocol);
 
 	std::map<std::string, node*> const& m_nodes;
 
 	dht_observer* m_observer;
 
-	protocol_descriptor const& m_protocol;
+	protocol_descriptor const m_protocol;
 
 	time_point m_last_tracker_tick;
 
--- a/include/libtorrent/kademlia/node.hpp
+++ b/include/libtorrent/kademlia/node.hpp
@@ -116,7 +116,6 @@ private:
 		udp protocol;
 		char const* family_name;
 		char const* nodes_key;
-		bool is_v4;
 	};
 
 	explicit protocol_descriptor(descriptor_data const& data)
--- a/include/libtorrent/kademlia/node.hpp
+++ b/include/libtorrent/kademlia/node.hpp
@@ -92,8 +92,7 @@ protected:
 
 struct TORRENT_EXTRA_EXPORT protocol_descriptor
 {
-	protocol_descriptor(udp protocol)
-		: m_protocol(protocol) {}
+	protocol_descriptor(udp protocol);
 
 	udp protocol() const;
 	char const* family_name() const;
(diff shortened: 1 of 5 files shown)


src/kademlia/node.cpp
```diff
@@ -1191,28 +1190,44 @@ void node::write_nodes_entries(sha1_hash const& info_hash
...
+}
 
-	for (int i = 0; i < sizeof(descriptors) / sizeof(protocol_descriptor); ++i)
-	{
-		if (descriptors[i].protocol == protocol)
-			return descriptors[i];
-	}
+char const* protocol_descriptor::family_name() const
+{
+	return is_v4() ? "n4" : "n6";
```
ssiloti: I don't like going through a conditional branch every time this is called. This is something which should be resolved once at construction time and stored. Same goes for `nodes_key`.

aldenml: true, I will do that

aldenml: oh, now I remember, I didn't want to add the member fields to the struct. What do you prefer: space vs CPU?

ssiloti: Personally I would re-add the statically allocated internal descriptors and make `protocol_descriptor` wrap a pointer to one of those.

aldenml: good idea, I will add a private inner structure to `protocol_descriptor` with that trick

include/libtorrent/kademlia/node.hpp
```diff
@@ -90,6 +90,39 @@ struct udp_socket_interface
...
+	bool is_native(address const& addr) const;
+
+	static protocol_descriptor v4();
+	static protocol_descriptor v6();
+
+	operator udp() const { return m_data.protocol; }
+
+private:
+
+	struct descriptor_data
```
ssiloti: Unfortunately I think this is going to need those pragmas to disable the warnings with msvc from the old code.

aldenml: I hope that's not the case, maybe because of the updated VS in `mater`

include/libtorrent/kademlia/node.hpp
```diff
@@ -90,6 +90,39 @@ struct udp_socket_interface
 	~udp_socket_interface() {}
 };
 
+struct TORRENT_EXTRA_EXPORT protocol_descriptor
```
arvidn: would you mind adding a comment here, describing what this type is used for?

include/libtorrent/kademlia/node.hpp
arvidn: It's not clear to me why you put the state in an inner class and have a reference to it. wouldn't the code be simpler if ``protocol_descriptor`` would contain the state, and you leave ``m_protocol`` a reference?

include/libtorrent/kademlia/node.hpp
aldenml: sure, cancel the current CI jobs if you want

src/kademlia/node.cpp
```diff
@@ -1191,28 +1190,60 @@ void node::write_nodes_entries(sha1_hash const& info_hash
 		if (wanted_node == m_nodes.end()) continue;
 		std::vector<node_entry> n;
 		wanted_node->second->m_table.find_node(info_hash, n, 0);
-		r[wanted_node->second->protocol_nodes_key()] = write_nodes_entry(n);
+		r[wanted_node->second->protocol().nodes_key()] = write_nodes_entry(n);
 	}
 }
 
-node::protocol_descriptor const& node::map_protocol_to_descriptor(udp protocol)
+protocol_descriptor::protocol_descriptor(udp protocol)
```
arvidn: if you did, a lot of these functions could just not refer to ``m_data``, and maybe even be defined inline above.

include/libtorrent/kademlia/node.hpp
aldenml: mostly to safe space and use it freely without taking care of references. It's that ok?

src/kademlia/node.cpp
aldenml: well, I suppose then that 8 bytes should not be that costly

src/kademlia/node.cpp
```diff
@@ -1191,28 +1190,60 @@ void node::write_nodes_entries(sha1_hash const& info_hash
...
+
+bool protocol_descriptor::is_native(address const& addr) const
+{
+	return (addr.is_v4() && is_v4())
+		|| (addr.is_v6() && is_v6());
+}
+
+protocol_descriptor protocol_descriptor::v4()
+{
+	static descriptor_data const data = {udp::v4(), "n4", "nodes"};
```
arvidn: for instance, you could have a private constructor, and just do: ``static protocol_descripto ret{udp::v4(), "n4", "nodes"}; return ret;`` and you would also have to return const references wherever you currently return a ``protocol_descriptor``

aldenml: on it :)

arvidn:
this change seems reasonable to me. But It seems like there may be some unnecessary complexity. I would also expect those one-line getter functions to be defined inline, and save lots of lines of code. Right now this patch net adds lines of code, even though it simplifies some places to use fewer lines. I would imagine it could be a net negative lines of code.

arvidn:
btw, I'm not saying my suggesting is definitely the way to go. I may be missing some aspect

aldenml:
sure, your idea was my first approach, but then I fell in the route of saving bits probably with no real benefit.

include/libtorrent/kademlia/node.hpp
arvidn: but if you make ``protocol_descriptor`` impossible to construct, other than the two instances that need to exist, and you always pass it around by const reference, there shouldn't be any difference..

arvidn:
I don't see how you loose bits.. the steven did it, ``protocol_descriptor`` was never copied, there were never more than 2 instances allocated, and everyone referred to it by const reference. From what I can tell, all you did was rename it to ``descriptor_data`` and you added a wrapper that contained that const reference called ``protocol_descriptor``. Unless I'm missing something, I don't see a great value in wrapping the reference. As far as I know, there aren't really any traps or mistakes you can make passing around the const reference. Since you can make it non-copyable and have private constructors.

aldenml:
yes, I didn't want to use `boost::noncopyable`, btw, would you consider create an `aux::noncopyable`?

src/kademlia/node.cpp
```diff
@@ -1191,28 +1190,60 @@ void node::write_nodes_entries(sha1_hash const& info_hash
 		if (wanted_node == m_nodes.end()) continue;
 		std::vector<node_entry> n;
 		wanted_node->second->m_table.find_node(info_hash, n, 0);
-		r[wanted_node->second->protocol_nodes_key()] = write_nodes_entry(n);
+		r[wanted_node->second->protocol().nodes_key()] = write_nodes_entry(n);
 	}
 }
 
-node::protocol_descriptor const& node::map_protocol_to_descriptor(udp protocol)
```
arvidn: I think this approach of creating the instances and picking the right one kind of makes sense. your version is simpler in the sense that there's only one test and then it picks based on that, but it's more complex in the sense that you require the wrapper around the const reference. You could rewrite this function to do the same though. Although, it seems reasonably conservative to fail if there's an unexpected protocol passed in too

arvidn:
I don't understand why you don't want to use ``boost::noncopyable``, but these days just declaring the copy constructor and assignment operator as ``= deleted`` may even be better style.

aldenml:
it's just personal preference not use `boost` types unless necessary. I will rethink with all your comments, I don't want to end up with the same code, just moved

aldenml:
Well, still a net plus in code lines...

aldenml:
Sorry for this PR, I don't feel like this is a good result. Maybe in another attempt.

PR #1158, opened 2016-09-26
various cleanup and formatting updates

arvidn:


src/torrent.cpp
```diff
@@ -1047,7 +1047,9 @@ namespace libtorrent
 #ifndef TORRENT_DISABLE_LOGGING
 		if (should_log())
 		{
-			debug_log("disk error: (%d) %s in file: %s", j->error.ec.value(), j->error.ec.message().c_str()
+			debug_log("disk error: (%d) %s [%s : %s] in file: %s"
+				, j->error.ec.value(), j->error.ec.message().c_str()
+				, job_name(j->action), j->error.operation_str()
```
aldenml: `job_name` is only enable `#if TORRENT_USE_ASSERTS`, fixing it

PR #1159, opened 2016-09-27
alerts refactor to use span, don't break ABI when logging, fix compilation

aldenml:


arvidn:
are there any practical benefits to include the log alert types when logging is disabled? there is some cost associated with including the names and code in the final binary, even through they will never be used.

aldenml:
you mentioned that you were thinking about this, and I think it's a good idea. I can provide multiple binaries without changing the java side. I think it's the same for DHT, it does not break de ABI. It's just 4 types, btw.

PR #1161, opened 2016-09-28
fix adding a node with unknown id

ssiloti:
Commit a414e4c3 broke adding nodes with unknown id to the routing table by
causing them to be evicted when they responded with their real id. This should
have been caught by test_bootstrap, but a typo caused that test to silently
exit early.

### DIFF
--- a/test/test_dht.cpp
+++ b/test/test_dht.cpp
@@ -1761,7 +1761,7 @@ void test_bootstrap(address(&rand_addr)())
 			|| find_node_keys[2].string_value() == "get_peers");
 		if (find_node_keys[0].string_value() != "q"
 			|| (find_node_keys[2].string_value() != "find_node"
-				&& find_node_keys[2].string_value() == "get_peers")) return;
+				&& find_node_keys[2].string_value() != "get_peers")) return;
 	}
 	else
 	{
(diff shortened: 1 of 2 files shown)


test/test_dht.cpp
```diff
@@ -1761,7 +1761,7 @@ void test_bootstrap(address(&rand_addr)())
 			|| find_node_keys[2].string_value() == "get_peers");
 		if (find_node_keys[0].string_value() != "q"
 			|| (find_node_keys[2].string_value() != "find_node"
-				&& find_node_keys[2].string_value() == "get_peers")) return;
+				&& find_node_keys[2].string_value() != "get_peers")) return;
```
arvidn: good catch. In a framework like catch, I suppose these two tests would be ``REQUIRE()``s, making it fatal if they fail, which would have made it less error prone.

arvidn:
lgtm

ssiloti:
I just realized there is a problem. If the id has changed we may need to move the node to a different bucket so I need to change this to do what the pre-a414e4c code did in the unknown id case and remove the old entry and re-insert the node.

arvidn:
oh, right

PR #1163, opened 2016-09-28
refactor to use move with dht_state

aldenml:


src/kademlia/dht_tracker.cpp
```diff
@@ -154,9 +155,12 @@ namespace libtorrent { namespace dht
 
 		m_refresh_timer.expires_from_now(seconds(5), ec);
 		m_refresh_timer.async_wait(std::bind(&dht_tracker::refresh_timeout, self(), _1));
-		m_dht.bootstrap(bootstrap.nodes, f);
+
+		m_dht.bootstrap(m_bootstrap_nodes, f);
+		std::vector<udp::endpoint>().swap(m_bootstrap_nodes);
```
arvidn: these days you can do: ``clear();`` and ``shrink_to_fit();`` instead

src/session_impl.cpp
```diff
@@ -5640,9 +5628,14 @@ namespace aux {
 		{
 			m_dht->add_node(n);
 		}
-		m_dht_nodes.clear();
+		std::vector<udp::endpoint>().swap(m_dht_nodes);
```
arvidn: ``shrink_to_fit()``

src/kademlia/dht_tracker.cpp
```diff
@@ -79,14 +79,16 @@ namespace libtorrent { namespace dht
 		, dht_settings const& settings
 		, counters& cnt
 		, dht_storage_interface& storage
-		, dht_state const& state)
+		, dht_state state)
 		: m_counters(cnt)
 		, m_storage(storage)
 		, m_dht(udp::v4(), this, settings, state.nid
 			, observer, cnt, m_nodes, storage)
+		, m_bootstrap_nodes(std::move(state.nodes))
```
arvidn: is there a reason to not contain a ``dht_state`` instead? do we only get the node out of it?

aldenml: for the moment yes, but I think it's a good idea to keep the state and provide a method `dht_state::clear()`

src/session_impl.cpp
```diff
@@ -5641,8 +5629,14 @@ namespace aux {
...
 		m_dht_nodes.clear();
+		m_dht_nodes.shrink_to_fit();
 
-		m_dht->start(startup_state, std::bind(&on_bootstrap, std::ref(m_alerts)));
+		auto cb = [this]
+		{
+			if (m_alerts.should_post<dht_bootstrap_alert>())
+				m_alerts.emplace_alert<dht_bootstrap_alert>();
+		};
+		m_dht->start(std::bind(cb));
```
arvidn: I think the ``bind`` here should be removed. I assume that if you found it to not work without the bind, there's some argument passed in to the callback that the lambda need to declare. ``std::bind`` will just take any argument ad throw it away.

src/session_impl.cpp
```diff
@@ -5661,7 +5655,7 @@ namespace aux {
 		m_dht_settings = settings;
 	}
 
-	void session_impl::set_dht_state(dht::dht_state const& state)
+	void session_impl::set_dht_state(dht::dht_state state)
 	{
 		m_dht_state = state;
```
arvidn: did you mean to say: ``= std::move(state);`` here?

aldenml: that line cost me the read of a dozen articules :) If I understood well, all them seems to indicate that in this particular case, the compiler can infer the "implicit" call to the move assignment operator, but if you ask me, I rather make it explicit.

src/session_impl.cpp
aldenml: I wanted to avoid the verbosity, let's see how it looks

src/session_impl.cpp
arvidn: any value that has a name is an lvalue and has to be explicitly moved. another way of thinking about it is that you could, if you wanted to, still refer to the ``state`` value after the assignment, so the compiler cannot decide to move it away from under you. Since only you can make the promise that you won't use the value again later, only you can ask for the move.

aldenml: hmm...I reasoned in the same way, but I was under the impression that in this particular case, the compiler can "cleverly" detect that there is no further use of the variable and only the destructor is called. I could be wrong in that assumption.

arvidn: as far as I know, the only place compilers do that is in a ``return`` statement.

PR #1166, opened 2016-09-28
minor refactor of dht_default_storage

aldenml:
The context of this is accommodating the source code for upcoming changes.

### DIFF
--- a/src/kademlia/dht_storage.cpp
+++ b/src/kademlia/dht_storage.cpp
@@ -88,16 +88,16 @@ namespace
 	{
 		// the actual value
 		std::unique_ptr<char[]> value;
-		// size of malloced space pointed to by value
-		int size = 0;
 		// this counts the number of IPs we have seen
 		// announcing this item, this is used to determine
 		// popularity if we reach the limit of items to store
 		bloom_filter<128> ips;
-		// number of IPs in the bloom filter
-		int num_announcers = 0;
 		// the last time we heard about this
 		std::time_t last_seen;
+		// number of IPs in the bloom filter
+		int num_announcers = 0;
+		// size of malloced space pointed to by value
+		int size = 0;
 	};
 
 	struct dht_mutable_item : dht_immutable_item
--- a/src/kademlia/dht_storage.cpp
+++ b/src/kademlia/dht_storage.cpp
@@ -38,6 +38,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <map>
 #include <set>
 #include <string>
+#include <ctime>
 
 #include <libtorrent/socket_io.hpp>
 #include <libtorrent/aux_/time.hpp>
@@ -83,17 +84,15 @@ namespace
 	struct dht_immutable_item
 	{
 		// the actual value
-		std::unique_ptr<char[]> value;
+		std::vector<char> value;
 		// this counts the number of IPs we have seen
 		// announcing this item, this is used to determine
 		// popularity if we reach the limit of items to store
 		bloom_filter<128> ips;
 		// the last time we heard about this
-		time_point last_seen;
+		std::time_t last_seen;
 		// number of IPs in the bloom filter
 		int num_announcers = 0;
-		// size of malloced space pointed to by value
-		int size = 0;
 	};
 
 	struct dht_mutable_item : dht_immutable_item
@@ -106,7 +105,7 @@ namespace
 
 	void touch_item(dht_immutable_item& f, address const& addr)
 	{
-		f.last_seen = aux::time_now();
+		f.last_seen = total_seconds(aux::time_now().time_since_epoch());
 
 		// maybe increase num_announcers if we haven't seen this IP before
 		sha1_hash const iphash = hash_address(addr);
@@ -326,8 +325,8 @@ namespace
 			auto const i = m_immutable_table.find(target);
 			if (i == m_immutable_table.end()) return false;
 
-			item["v"] = bdecode(i->second.value.get()
-				, i->second.value.get() + i->second.size);
+			item["v"] = bdecode(i->second.value.begin()
+				, i->second.value.end());
 			return true;
 		}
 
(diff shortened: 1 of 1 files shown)


src/kademlia/dht_storage.cpp
```diff
@@ -83,17 +84,15 @@ namespace
 	struct dht_immutable_item
 	{
 		// the actual value
-		std::unique_ptr<char[]> value;
+		std::vector<char> value;
```
arvidn: changing this to a vector will make it use 16 more bytes (3 pointers instead of 1). the ``int`` for the size will fill out the padding caused by ``num_announcers``

src/kademlia/dht_storage.cpp
```diff
@@ -106,7 +105,7 @@ namespace
 
 	void touch_item(dht_immutable_item& f, address const& addr)
 	{
-		f.last_seen = aux::time_now();
+		f.last_seen = total_seconds(aux::time_now().time_since_epoch());
```
arvidn: unfortunately this won't work either. I forget, the idea is to persist this timestamp to disk, right? The epoch of the steady clock is likely to be something like: time of last boot. You need a real-time clock, or wall clock, here. I would imagine you'd want to use ``std::chrono::system_clock`` and use ``to_time_t``.

(i.e. ``libtorrent::clock_type`` is ``steady_clock``, which just guarantees that it is monotonically increasing, no guarantee about having a fixed epoch)

src/kademlia/dht_storage.cpp
```diff
@@ -435,14 +430,9 @@ namespace
 
 				if (item.seq < seq)
 				{
-					if (item.size != buf.size())
-					{
-						item.value.reset(new char[buf.size()]);
-						item.size = int(buf.size());
-					}
+					item.value = {buf.begin(), buf.end()};
```
arvidn: getting the memory handling here seems nice though

src/kademlia/dht_storage.cpp
aldenml: I was wondering about it too, let me rethink it

src/kademlia/dht_storage.cpp
aldenml: yes, it's to save to disk, thanks for the explanation, looking at it

src/kademlia/dht_storage.cpp
aldenml: What do you think of reducing the size of `ips` to save that number of bytes? the use of vector is very convenient.

arvidn: I agree that vector is convenient. there is a type ``buffer`` in libtorrent, which basically is a single pointer and size + the memory management. It could be a reasonable alternative, but it would also couple the dht to the core library more too.

aldenml: I don't mind testing the idea and see how it looks

aldenml: I'm surprised that `libtorrent::buffer` can overallocate, not sure if it's a good fit for this case

src/kademlia/dht_storage.cpp
```diff
@@ -84,16 +88,16 @@ namespace
 	{
 		// the actual value
 		std::unique_ptr<char[]> value;
+		// size of malloced space pointed to by value
+		int size = 0;
```
arvidn: I don't want to nit-pick, but now you got 32 bit padding here, and 32 bit padding after ``num_announcers``.

aldenml: are you saying that the order matters? didn't know that...changing it to the original version.

src/kademlia/dht_storage.cpp
```diff
@@ -473,13 +477,17 @@ namespace
...
+
+			auto const now = stime + (aux::time_now() - ctime);
 			time_duration lifetime = seconds(m_settings.item_lifetime);
 			// item lifetime must >= 120 minutes.
 			if (lifetime < minutes(120)) lifetime = minutes(120);
 
 			for (auto i = m_immutable_table.begin(); i != m_immutable_table.end();)
 			{
-				if (i->second.last_seen + lifetime > now)
+				if (system_clock::from_time_t(i->second.last_seen) + lifetime > now)
```
arvidn: I would imagine that you could just test this against the system clock, without converting the steady_clock into the system_clock domain

aldenml: I believe not, because of the simulation

arvidn: i.e. ``if (i->second.last_seen > system_clock::now())``. you could avoid the ``to_time_t`` conversion by keeping the ``last_seen`` member a ``time_point``, and convert to and from ``time_t`` when saving and restoring.

aldenml: I can change the type to `system_clock::time_point`, but because of the simulation, I need to work with a `system_clock` start point as a reference and offset with the `steady_clock`. Am I wrong?

src/kademlia/dht_storage.cpp
```diff
@@ -119,9 +119,18 @@ namespace
 		std::memcpy(item.value.get(), buf.data(), size);
 	}
 
+	system_clock::time_point clock_now()
+	{
+		static auto const stime = system_clock::now();
+		static auto const ctime = aux::time_now();
+
+		auto const secs = total_seconds(aux::time_now() - ctime);
+		return stime + seconds(secs);
```
arvidn: oh, you need to do this to get the simulations to work, right? maybe an easier way to do this would be to use the internal clock here, and just convert to and from ``time_t`` when saving and restoring

aldenml: Yes, that's the reason. Now, The type `dht_immutable_item` will be promoted to a public place in a latter commit. I think it's better and more clear to have the type `system_clock::time_point` for `last_seen`. What do you think?

aldenml: ok ok...I think I can use the old structure, and save a `timespan` as a reference when saving/loading

arvidn:
lgtm

PR #1169, opened 2016-10-01
fix error_code vs. error_condition traits

arvidn:
deprecate get_*_category(), instead use *_category()

### DIFF
--- a/src/peer_connection.cpp
+++ b/src/peer_connection.cpp
@@ -2124,7 +2124,7 @@ namespace libtorrent
 		if (t->share_mode()) return false;
 
 		if (m_upload_only && t->is_upload_only()
-			&& can_disconnect(error_code(errors::upload_upload_connection, get_libtorrent_category())))
+			&& can_disconnect(errors::upload_upload_connection))
 		{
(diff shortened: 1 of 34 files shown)


src/peer_connection.cpp
```diff
@@ -2124,7 +2124,7 @@ namespace libtorrent
 		if (t->share_mode()) return false;
 
 		if (m_upload_only && t->is_upload_only()
-			&& can_disconnect(error_code(errors::upload_upload_connection, get_libtorrent_category())))
+			&& can_disconnect(errors::upload_upload_connection))
```
aldenml: Question: If `error_code(` is not necessary (or maybe it's), is it a good idea to remove the rest?

aldenml:
(don't mind if you want to cancel my CI jobs)

src/peer_connection.cpp
arvidn: yeah, I think so. previously I was a bit confused about error_code and error_condition, and I had marked the libtorrent error codes as both, which made this not automatically pick the right constructor..

PR #1170, opened 2016-10-01
added save/load items support to DHT storage (internal feature only)

aldenml:


include/libtorrent/kademlia/dht_storage.hpp
```diff
@@ -64,6 +70,50 @@ namespace dht
...
+		public_key key;
+		std::string salt;
+	};
+
+	// This structure hold the DHT items (immutable and mutable)
+	// exported by the DHT's storage.
+	struct TORRENT_EXPORT dht_storage_items
+	{
+		// time reference to interpret the item's ``last_seen``
+		// actual last_seen time is (timestamp - total_seconds(last_seen))
```
arvidn: is this comment right? ``total_seconds()`` take a ``duration``, not a ``time_point``, doesn't it?
I take it you mean ``total_seconds(last_seen.time_since_epoch())`` or something like that.

include/libtorrent/kademlia/dht_storage.hpp
```diff
@@ -214,6 +264,24 @@ namespace dht
...
+		// for each item's type.
+		virtual dht_storage_items save_items() const = 0;
+
+		// Loads the items in the storage.
+		//
+		// For implementers:
+		// It's recommended that you take in consideration the ``last_seen``
+		// field in the items during the load phase. Similarly you should not
+		// store more than ``dht_settings::max_dht_items``.
+		virtual void load_items(dht_storage_items items) = 0;
```
arvidn: do you think there's value in being able to load items after construction? or load items more than once? In some sense it smells a little bit like two-phase initialization. Do you think it would be inconvenient to load in the constructor instead?

include/libtorrent/kademlia/dht_storage.hpp
```diff
@@ -64,6 +70,50 @@ namespace dht
 		void reset();
 	};
 
+	struct TORRENT_EXPORT dht_immutable_item
```
arvidn: I kind of liked that these types were hidden implementation details in the storage. I'm not sure what the best approach is though

include/libtorrent/kademlia/dht_storage.hpp
aldenml: no, yes, ok...I will change it :)

include/libtorrent/kademlia/dht_storage.hpp
aldenml: thinking about it

include/libtorrent/kademlia/dht_storage.hpp
aldenml: I will iterate with another idea. Let's see how it looks then

aldenml:
let me know what do you think now

src/kademlia/dht_storage.cpp
```diff
@@ -486,24 +513,67 @@ namespace
...
 			ret.timestamp = system_clock::to_time_t(system_clock::now());
 
+			auto copy = [&](dht_immutable_data& d, dht_immutable_item const& item)
+			{
+				d.value = {item.value.get(), item.value.get() + item.size};
+				d.popularity = item.ips;
+				d.last_seen = time_point(now - item.last_seen);
+			};
+
 			for (auto const& p : m_immutable_table)
```
arvidn: this loop looks like it could be replaced by a call to ``std::transform()``

include/libtorrent/kademlia/dht_storage.hpp
```diff
@@ -70,34 +70,24 @@ namespace dht
 		void reset();
 	};
 
-	struct TORRENT_EXPORT dht_immutable_item
+	struct TORRENT_EXPORT dht_immutable_data
```
arvidn: they're still public and still widen the interface to the library. They're just duplicated, I don't think that really adds that much. One potential benefit I can imagine is that you may not want to actually store the bloom filter, but rather some kind of popularity metrics. Or perhaps nothing at all, and just let the popularity be re-recorded

include/libtorrent/kademlia/dht_storage.hpp
```diff
@@ -70,34 +70,24 @@ namespace dht
...
-		dht_immutable_item& operator=(dht_immutable_item&&) = default;
-
 		// the actual value
-		std::unique_ptr<char[]> value;
+		std::vector<char> value;
 		// this counts the number of IPs we have seen
 		// announcing this item, this is used to determine
 		// popularity if we reach the limit of items to store
-		bloom_filter<128> ips;
+		bloom_filter<128> popularity;
```
arvidn: perhaps this isn't very important to save and restore

include/libtorrent/kademlia/dht_storage.hpp
arvidn: I think it's a bit confusing to have duplicate structs though, that are so similar and have such similar names. It's not obvious I think what they are used for (at least not from their names)

include/libtorrent/kademlia/dht_storage.hpp
```diff
@@ -70,34 +70,24 @@ namespace dht
...
 		// the last time we heard about this item
 		// the correct interpretation of this field
 		// requires a time reference
 		time_point last_seen;
```
arvidn: another benefit of having separate types is that this timestamp could be a ``time_t`` proper, and you wouldn't need the ``timestamp`` field below.

include/libtorrent/kademlia/dht_storage.hpp
aldenml: I will remove the `bloom_filter` field and I think that would give more credibility to the new structs. Naming can be an issue.

aldenml:
Let me know what do you think now. Do you think `last_seen` is (very) important to save?

include/libtorrent/kademlia/dht_storage.hpp
```diff
@@ -42,6 +46,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <libtorrent/address.hpp>
 #include <libtorrent/span.hpp>
 #include <libtorrent/string_view.hpp>
+#include <libtorrent/bloom_filter.hpp>
```
arvidn: I don't think you need to move this to the header anymore now

arvidn:
Thinking about the point of the dht_storage interface and what kinds of storages I can imagine, this seems very specific to the default storage. I imagine any other storage to not store the objects in RAM directly, but in some kind of database interface. As far as I can see, any other storage would not need or want to return any items via this new save/load interface. What do you think about instead letting the storage itself be responsible for exactly _how_ something is saved. The dht node could still ask it to save or flush its objects to persistent storage. What do you think of a higher level interface to achieve this where:

1. the storage object receives a string in its constructor (which could be a settings_pack or dht_settings configuration option). For the default storage this is interpreted as a save path for the dht items, but other storages may interpret it as a database/table name or filename.

2. The dht node tells the storage that it's time to save items before it shuts down (and maybe periodically too, but I don't think that would be necessary).

aldenml:
I wanted to avoid dealing with filesystem IO in the default storage. I understand your points, I will explore that route.

aldenml:
not clear to me how to address the problem in a way that does not augment the public interface in a significant way. I will take a pause on this and revisit it later.

PR #1180, opened 2016-10-03
refactor log methods in public peer_connection_handle for stable ABI

aldenml:
This breaks the ABI, and it's very internal to libtorrent, if a client insist in using the "peer log" methods, he can get the `native_handle()`.

### DIFF
--- a/include/libtorrent/peer_connection_handle.hpp
+++ b/include/libtorrent/peer_connection_handle.hpp
@@ -36,6 +36,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/config.hpp"
 #include "libtorrent/peer_id.hpp"
 #include "libtorrent/operations.hpp"
+#include "libtorrent/alert_types.hpp"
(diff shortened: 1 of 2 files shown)


arvidn:
It looks like a good patch. do you know where this came from or why it was added in the first place?

ssiloti:
My reputation plugin uses them to log info about the various extension messages it sends/receives.

aldenml:
Do you think you can call it through the native handle? I still think this is too internal, if for example, a user gets a handle he can logs

arvidn:
I suspected steven might use them. I don't really see the problem with exposing these. A user with a ``peer_connection_handle`` can do a lot of things, issuing logs doesn't seem to be the worst or most error prone thing.

aldenml:
Can i least make it ABI stable? Moving the guards to inside, similar to DHT in session handle

arvidn:
yeah, I think it makes sense to just provide empty implementations of those functions when logging is disabled

arvidn:
lgtm

ssiloti:
lgtm

PR #1188, opened 2016-10-06
Fix #1182

### LINKED ISSUE #1182: Possible duplicates: std::vector<announce_entry> m_trackers
Example tracker:
http://kino-tor.org/torrent/403102/lake-of-tears-by-the-black-sea-2014-mp3
Copy and open magnet:
magnet:?xt=urn:btih:0099d4cbf3b027f8f6e0d9bac7c57a16a56f7458&dn=rutor.info_Lake+Of+Tears+-+By+The+Black+Sea+%282014%29+MP3&tr=udp://opentor.org:2710&tr=udp://opentor.org:2710&tr=http://retracker.local/announce

libtorrent-trackers-dup

pavel-pimenov:


src/torrent.cpp
```diff
@@ -10613,15 +10614,17 @@ namespace libtorrent
 			, std::bind(&torrent::on_piece_verified, shared_from_this(), _1)
 			, reinterpret_cast<void*>(1));
 	}
-
-	announce_entry* torrent::find_tracker(tracker_request const& r)
```
arvidn: since this version of ``find_tracker()`` only looks at the URL (and not the type), it seems like it's misleading. Would you mind removing it? and have callers just use your new version instead.

src/torrent.cpp
```diff
@@ -10613,15 +10614,17 @@ namespace libtorrent
...
+	announce_entry* torrent::find_tracker(std::string const& url)
 	{
-		std::vector<announce_entry>::iterator i = std::find_if(
-			m_trackers.begin(), m_trackers.end()
-			, [&r] (announce_entry const& ae) { return ae.url == r.url; });
+		std::vector<announce_entry>::iterator i = std::find_if(m_trackers.begin(), m_trackers.end()
```
arvidn: this seems like a perfect place to use auto

src/torrent.cpp
```diff
@@ -302,7 +302,10 @@ namespace libtorrent
 			e.fail_limit = 0;
 			e.source = announce_entry::source_magnet_link;
 			e.tier = tier;
-			m_trackers.push_back(e);
+			if (!find_tracker(e.url))
```
arvidn: wouldn't it make more sense to call ``add_tracker()`` here, instead of adding it directly to the vector?

pavel-pimenov: add_tracker() side effect:
* call announce_with_tracker(), 
* k->source = announce_entry::source_client;
* ++m_last_working_tracker;

src/torrent.cpp
pavel-pimenov: find_tracker(tracker_request const& r) is public metod...
use TORRENT_NO_DEPRECATE or just remove?

src/torrent.cpp
arvidn: I see

src/torrent.cpp
arvidn: the whole class ``torrent`` is an implementation detail, so no need to deprecate

pavel-pimenov: Ok.
private:
		announce_entry* find_tracker(std::string const& url);
public:
?

src/torrent.cpp
pavel-pimenov: i.e. replace if (!find_tracker(e.url)) {m_trackers.push_back(e);}  -> add_tracker(e)?

pavel-pimenov:
I try compile and add new test (ubuntu 16.04 x64)
c++ (Ubuntu 5.4.0-6ubuntu1~16.04.2) 5.4.0 20160609

help?

ppa@ppa-P5K:~/src/libtorrent/test$ make test
g++ -std=c++11 -g -O2 -ftemplate-depth=128 -Wno-format-zero-length -fvisibility=hidden -fvisibility-inlines-hidden -I/usr/include  -L/usr/lib/x86_64-linux-gnu  -fvisibility=hidden -fvisibility-inlines-hidden  test.cpp   -o test
/usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/crt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
/tmp/ccxEZcYw.o: In function `__static_initialization_and_destruction_0':
/usr/include/boost/system/error_code.hpp:221: undefined reference to `boost::system::generic_category()'
/usr/include/boost/system/error_code.hpp:222: undefined reference to `boost::system::generic_category()'
/usr/include/boost/system/error_code.hpp:223: undefined reference to `boost::system::system_category()'
/tmp/ccxEZcYw.o: In function `boost::asio::error::get_system_category()':
/usr/include/boost/asio/error.hpp:230: undefined reference to `boost::system::system_category()'

pavel-pimenov:
https://travis-ci.org/arvidn/libtorrent/jobs/165741077#L2424
virtual libtorrent::piece_block_progress libtorrent::http_seed_connection::downloading_piece_progress()

src/torrent.cpp
arvidn: as you point out ``add_tracker`` has side effects not desirable here, so I think we'd need to stick with just push_back

arvidn:
lgtm

pavel-pimenov:
Add check for m_urls?
void create_torrent::add_tracker(string_view url, int const tier)	{
 m_urls.push_back(announce_entry(url.to_string(), tier));
...

arvidn:
I don't understand your question

pavel-pimenov:
...
lt::create_torrent t(info);
 t.add_tracker("http://3");
 t.add_tracker("http://3");

m_urls.size() == 2 -> bug?

arvidn:
I see. I think that's probably a less important case, and in either case should be a separate PR

PR #1192, opened 2016-10-06
refactor in buffer.hpp to not use anonymous namespace in header

aldenml:


include/libtorrent/buffer.hpp
```diff
@@ -75,7 +68,7 @@ class buffer
 
 		if (size == 0) return;
 
-		size = round_up8(size);
+		size = (size + 7) & (~std::size_t(0x7)); //round_up8(size);
```
arvidn: come to think of it, I don't understand why the size is aligned to 8 bytes.

arvidn: especially given that we adjust-up the size of the buffer below

aldenml: you tell me, not familiar with this `buffer`

arvidn: I want to look into the history of where this came from

arvidn: I think the rounding up does make sense, since the buffer is not resizable. I suppose it mostly makes sense for platforms of ``malloc_size()``. Is there a good reason to collapse the function down to this expression? just fewer lines? I suppose it's not that important, but maybe you could remove the comment with the previous function call

aldenml: no, it's not about fewer lines :)  The main reason for this PR is the problematic use of an anonymous namespace in a header file. It's documented to be the cause of bloated final binaries and even ODR violations. I will change the comment with your information.

arvidn: ah, that's a good point. it shouldn't be in an anonymous namespace, either collapsed as you did or in the ``detail`` namespace

include/libtorrent/buffer.hpp
```diff
@@ -68,7 +68,10 @@ class buffer
 
 		if (size == 0) return;
 
-		size = (size + 7) & (~std::size_t(0x7)); //round_up8(size);
+		// this rounds up the size to be 8 bytes aligned
+		// it mostly makes sense for platforms with support
```
arvidn: heh, maybe I miswrote. I meant to say "without". the ``malloc_size()`` rounds up to the actual slab size

arvidn:
where did that SSL error come from? seriously..

arvidn:
lgtm

aldenml:
fixing it, I think it's the same problem you are having in your other PR now. It's related to the deprecated openssl coming with the macOS

aldenml:
well, it's not the same in your other PR (I just checked)

aldenml:
I reverted the openssl fix, let me try to figure out in another PR

PR #1194, opened 2016-10-08
make single-argument constructors explicit

arvidn:


include/libtorrent/span.hpp
```diff
@@ -117,4 +117,6 @@ namespace libtorrent
 	};
 }
 
+#undef implicit
```
aldenml: just asking: this is for?

arvidn: heh, in a previous incarnation I was trying on using a GSL-style "implicit" macro (to silence warnings of not using explicit). I was using google's cpplint.py though, so I had to go with // NOLINT instead.

PR #1201, opened 2016-10-09
using arm neon inline assembly and using it only if in 32 bits

aldenml:


arvidn:
did this improve the generated code? or is it a portability issue?

aldenml:
For some reason (I don't know why), the android NDK clang with optimization enabled, make these intrinsics to crash randomly.

arvidn:
so this change produces different binary code then? and it fixes the problem?

aldenml:
I looked at it and it was actually on my side (not clang, but in debug it was not happening). Now this change makes the type of the data load more clear and the crash is not happening.

arvidn:
did you see this? https://travis-ci.org/arvidn/libtorrent/jobs/166627318#L1684

aldenml:
fixing it

aldenml:
well, now it's failing again with the intrinsics, I will revert to the inline assembly and do more tests. (android+clang+release)

aldenml:
do you have any particular concern with inline assembly?

aldenml:
what a nightmare, optimization is definitely messing up with the NEON registers but I just learned that the opcodes are not the same for arm 64 bits

PR #1206, opened 2016-10-12
Fix examples compilation with CMake

lyssdod:


examples/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ if(UNIX)
 endif ()
 set(ENV{PKG_CONFIG_PATH} "${PKG_CONFIG_CHANGED_PATH}")
 
+find_package(Threads REQUIRED)
```
arvidn: what's this "Threads" library? Is this portable across platforms and compilers?

examples/CMakeLists.txt
```diff
@@ -26,4 +27,15 @@ foreach(s ${examples})
     get_filename_component (sn ${s} NAME_WE)
     add_executable(${sn} ${s})
     target_link_libraries(${sn} ${LibtorrentRasterbar_LIBRARIES} ${Boost_LIBRARIES})
+
+    # some flags for cmake 2.6+
+    # much simpler for cmake 3.1+: https://cmake.org/cmake/help/v3.1/module/FindThreads.html
+    if(THREADS_HAVE_PTHREAD_ARG)
+        set_property(TARGET ${sn} PROPERTY COMPILE_OPTIONS "-pthread")
+        set_property(TARGET ${sn} PROPERTY INTERFACE_COMPILE_OPTIONS "-pthread")
```
arvidn: this looks GCC specific...

lyssdod: This shouldn't be triggered on Windows (it doesn't have pthreads), though I didn't try Windows compilation yet

aldenml:
I believe this PR is stalled and for a very old version. Is ir right to close it?

PR #1210, opened 2016-10-14
minor refactor, consts and cleanup in alerts

aldenml:


include/libtorrent/alert_types.hpp
```diff
@@ -1843,7 +1843,7 @@ namespace libtorrent
 		// interpret these values throughout the process' runtime.
 		//
 		// For more information, see the session-statistics_ section.
-		std::uint64_t values[counters::num_counters];
+		std::int64_t values[counters::num_counters];
```
arvidn: +1

src/alert.cpp
```diff
@@ -1900,7 +1895,7 @@ namespace libtorrent {
 		, m_alloc(alloc)
 		, m_num_peers(int(peers.size()))
 	{
-		std::size_t total_size = m_num_peers; // num bytes for sizes
+		std::size_t total_size = 0; // num bytes for sizes
 		for (int i = 0; i < m_num_peers; i++) {
 			total_size += peers[i].size();
```
arvidn: looking at this, is ``size()`` really correct here? I worry that it means ``sizeof(in_addr)`` or ``sizeof(in6_addr)``, rather than 6 or 18.

include/libtorrent/alert_types.hpp
```diff
@@ -117,10 +117,10 @@ namespace libtorrent
 		virtual std::string message() const override;
 
 		// The peer's IP address and port.
-		tcp::endpoint ip;
+		tcp::endpoint const ip;
```
aldenml: I don't think so
```cpp
void pop_alerts(std::vector<alert*>* alerts);
alert* wait_for_alert(time_duration max_wait);
```
btw, putting this constant has a significant impact in the swig interface, it does not generate the "setter"

src/alert.cpp
aldenml: I believe it is, all the info is stored (just below)

include/libtorrent/alert_types.hpp
arvidn: this should probably even be a ``std::array<>`` I imagine

include/libtorrent/alert_types.hpp
arvidn: the members that you made const, is there a reason behind picking just those? I imagine if alert members should be const, they should all be const. or what do you think?

include/libtorrent/alert_types.hpp
aldenml: do you think it's better in another PR?

include/libtorrent/alert_types.hpp
aldenml: I picked just a handful while doing the cleanup. I didn't want to go full on it before getting some review from you. If you are OK with it, I can go (in another PR) and review all the alerts.

include/libtorrent/alert_types.hpp
arvidn: sure

src/alert.cpp
arvidn: I'm pretty sure the "underlying type" is the OS type, like ``sockaddr``: http://www.boost.org/doc/libs/1_62_0/doc/html/boost_asio/reference/ip__basic_endpoint/size.html

That's going to me greater than the IP address and port.

arvidn:
although signed counters is greatly preferred, changing this will be an API change.. which has ripples into the client, as demonstrated here. https://travis-ci.org/arvidn/libtorrent/jobs/167538963#L5468

It should probably at least be mentioned in the ChangeLog

arvidn:
lgtm

src/alert.cpp
aldenml: let me review/fix this in another PR

include/libtorrent/alert_types.hpp
arvidn: the reason I hesitate is that I seem to recall that the session_view (as part of client_test) specifically relies on mutating the counters it receives. I'll take a look if that's the case

arvidn: ok, I think it may have relied on this at some point in the past, perhaps by swapping a std::vector<> out of the alert rather than copying.

arvidn:
sorry, I caused a conflict

aldenml:
no a problem, it's just the ChangeLog file

PR #1218, opened 2016-10-15
take upload speed into account in round-robin choker

### LINKED ISSUE #1171: Poor upload utilization with default choking algorithm
libtorrent version (or branch): RC_1_1, master

For a long time I've noticed that libtorrent often doesn't make full use of my available upload throughput. I've been ignoring it because the rate is usually within 80% and it's hard to tell if it's libtorrent's fault or just my ISP being crappy.

Today I had a particularly bad session where the upload rate was persistently around 60% of the limit[0] and I finally decided to look into it. After capturing some logs I've concluded that the problem is with the default round robin seed choking algorithm. Because the quota is set in terms of bytes uploaded, faster peers hit it sooner than slower ones. This tends to result in most upload slots being taken by slow peers who take a long time to hit the quota. At any given time there are typically only one or two fast peers unchoked.

To test this theory I changed the seed choking algorithm to fastest upload and saw an obvious improvement. Many fast peers were unchoked keeping the connection consistently saturated as you would expect. I noticed qBittorrent has hard coded the use of fastest upload since December 2014[1].

I thought I should allow for some discussion before going off and working on a fix. One possible way to improve the behavior of round robin would be to set the quota in time rather than bytes uploaded to remove the bias towards slow peers.

[0] Note that I do not set any rate limits in libtorrent, I enforce the limit at my router.
[1] https://github.com/qbittorrent/qBittorrent/commit/b46a0ea9b22e5707462f1006b84599e26da00e31

ssiloti:
The unchoked peers need to be sorted by upload speed so that the slowest node
will get choked as part of optimistic unchoking.

Also change the minimum quanta to 1 minute instead of 256KB.

Also remove a useless multiply in the fastest peer choker since the priorities
are guarenteed to be equal at that point.

Fixes #1171

### DIFF
--- a/src/choker.cpp
+++ b/src/choker.cpp
@@ -80,9 +80,9 @@ namespace libtorrent
 		// peers that are unchoked, but have sent more than one quota
 		// since they were unchoked, they get de-prioritized.
 
-		// if a peer is already unchoked, and the number of bytes sent since it was unchoked
-		// is greater than the send quanta, then it's done with it' upload slot, and we
-		// can de-prioritize it
+		// if a peer is already unchoked, the number of bytes sent since it was unchoked
+		// is greater than the send quanta, and it has been unchoked for at least one minute
+		// then it's done with its upload slot, and we can de-prioritize it
 		bool c1_quota_complete = !lhs->is_choked()
 			&& c1 > t1->torrent_file().piece_length() * pieces
 			&& aux::time_now() - lhs->time_of_last_unchoke() > minutes(1);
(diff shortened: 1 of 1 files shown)


src/choker.cpp
```diff
@@ -82,21 +83,30 @@ namespace libtorrent
 		// if a peer is already unchoked, and the number of bytes sent since it was unchoked
```
arvidn: this comment isn't quite right now, speaking about the quanta in terms of bytes

arvidn:
lgtm

PR #1219, opened 2016-10-16
fix buffer overrun in dht_get_peers_reply_alert

arvidn:
I wonder why address sanitizer on g++-5 didn't catch this (clang++ sanitizer caught it on macOS)


aldenml:
what?!! that was my mistake to put `0` there (I have pending the size review in another PR)

arvidn:
I feel pretty confident that ``size()`` on the endpoint returns the expected size, not ``sizeof()`` the underlying address structure

PR #1222, opened 2016-10-17
refactor in alert types to use more const fields and more clear API

aldenml:


aldenml:
Thinking that maybe I should add the old fields as deprecated, but wanted to get some review first

include/libtorrent/alert_types.hpp
```diff
@@ -117,7 +117,7 @@ namespace libtorrent
 		virtual std::string message() const override;
 
 		// The peer's IP address and port.
-		tcp::endpoint const ip;
+		tcp::endpoint const endpoint;
```
arvidn: yeah. to rename these fields you have to keep deprecated duplicates to stay API compatible

include/libtorrent/alert_types.hpp
```diff
@@ -117,10 +117,15 @@ namespace libtorrent
...
 
 		// the peer ID, if known.
 		peer_id const pid;
+
+#ifndef TORRENT_NO_DEPRECATE
+		// The peer's IP address and port.
+		tcp::endpoint const ip;
```
arvidn: here's a crazy thought: could this just be a ``const&`` to ``endpoint``?

arvidn: hm.. I guess it wouldn't be trivially copyable or movable anymore..

include/libtorrent/alert_types.hpp
```diff
@@ -2045,11 +2069,17 @@ namespace libtorrent
 		virtual std::string message() const override;
 
 		// returns the log message
+		char const* log_message() const;
```
arvidn: this doesn't seem to buy you much..

include/libtorrent/alert_types.hpp
```diff
@@ -2285,12 +2327,16 @@ namespace libtorrent
...
-		direction_t const dir;
+		direction_t const direction;
 
 		// the DHT node we received this packet from, or sent this packet to
-		// (depending on ``dir``).
+		// (depending on ``direction``).
 		udp::endpoint const node;
 
+#ifndef TORRENT_NO_DEPRECATE
+		direction_t const dir;
```
arvidn: I take it you don't like the shorter name?

simulation/test_socks5.cpp
```diff
@@ -220,7 +220,7 @@ TORRENT_TEST(socks5_tcp_listen_alert)
 		[&](lt::session&, lt::alert const* alert) {
 			if (auto* a = lt::alert_cast<lt::listen_succeeded_alert>(alert))
 			{
-				if (a->sock_type == listen_succeeded_alert::socks5)
+				if (a->socket_type == listen_succeeded_alert::socks5)
```
arvidn: If you're going to change this name (i.e. the API), what do you think about changing its type as well, to be a proper enum class? (that way it doesn't have to be a member of listen_succeeded_alert either)

include/libtorrent/alert_types.hpp
aldenml: it seems more clear to me in the public API

simulation/test_socks5.cpp
aldenml: sure, I think it's a good idea, not exactly better for swig :)

include/libtorrent/alert_types.hpp
```diff
@@ -1232,8 +1245,6 @@ namespace libtorrent
 	// listen on it.
 	struct TORRENT_EXPORT listen_failed_alert final : alert
 	{
-		enum socket_type_t { tcp, tcp_ssl, udp, i2p, socks5, utp_ssl };
```
arvidn: you still need the old enum as deprecated, just like the old name, ``sock_type``

include/libtorrent/alert_types.hpp
```diff
@@ -1259,44 +1270,45 @@ namespace libtorrent
...
 		// see alert's documentation for validity of this value
-		int port;
+		int const port;
 
 #ifndef TORRENT_NO_DEPRECATE
 		// the address and port libtorrent attempted to listen on
 		tcp::endpoint endpoint;
+
+		// the type of listen socket this alert refers to.
+		socket_type_t sock_type;
```
arvidn: ah, ok, you got this one. This can't be an enum class though, it would break the API

include/libtorrent/alert_types.hpp
arvidn: you should probably keep the name of this enum too. you can disambiguate by fully qualify them. ``::libtorrent::socket_t`` vs. ``::libtorrent::listen_failed_alert::socket_t``

arvidn:
lgtm apart from comments

aldenml:
good, I will re-add the old enums in deprecated

aldenml:
do you think the last commit (using `socket_type_t_deprecated`) is acceptable? even with the fully qualify name, gcc complains about "change the meaning"

aldenml:
same error with the different name, need to figure out another solution

src/alert.cpp
```diff
@@ -804,14 +821,15 @@ namespace libtorrent {
...
+		, libtorrent::socket_type_t t)
 		: error(ec)
 		, operation(op)
-		, sock_type(t)
+		, socket_type(t)
 		, address(listen_addr)
 		, port(listen_port)
 #ifndef TORRENT_NO_DEPRECATE
 		, endpoint(listen_addr, listen_port)
+		, sock_type(static_cast<socket_type_t>(sock_type_idx(t)))
```
arvidn: you could just say: ``sock_type(int(t))`` or, if that doesn't work: ``static_cast<int>(t)``

src/alert.cpp
```diff
@@ -774,10 +780,21 @@ namespace libtorrent {
 
 	namespace
 	{
-		static char const* const sock_type_str[] =
+		int sock_type_idx(socket_type_t type)
```
arvidn: this function is probably not necessary

src/alert.cpp
```diff
@@ -774,10 +780,21 @@ namespace libtorrent {
...
+			TORRENT_ASSERT(0 <= idx && idx < 6);
+			return idx;
+		}
+
+		char const* sock_type_str(socket_type_t type)
+		{
+			static char const* type_str[] =
+				{ "TCP", "TCP/SSL", "UDP", "I2P", "Socks5", "uTP/SSL" };
+
+			return type_str[sock_type_idx(type)];
```
arvidn: here you could just say: ``return type_str[static_cast<int>(type)];``

PR #1223, opened 2016-10-18
Try resolve #1175

### LINKED ISSUE #1175: Access Violation upnp::delete_port_mapping
platform/architecture: Win 7 sp1 x86
compiler and compiler version: VC++2015 update 3

Dumps from anonymous users:
https://drdump.com/Problem.aspx?ProblemID=238693

libtorrent-crash-1

flylinkdc!strnlen+0xdc
flylinkdc!__crt_stdio_output::output_processor<char,__crt_stdio_output::string_output_adapter<char>,__crt_stdio_output::format_validation_base<char,__crt_stdio_output::string_output_adapter<char> > >::type_case_s+0x68
flylinkdc!__crt_stdio_output::output_processor<char,__crt_stdio_output::string_output_adapter<char>,__crt_stdio_output::standard_base<char,__crt_stdio_output::string_output_adapter<char> > >::state_case_type+0x4d
flylinkdc!__crt_stdio_output::output_processor<char,__crt_stdio_output::string_output_adapter<char>,__crt_stdio_output::standard_base<char,__crt_stdio_output::string_output_adapter<char> > >::process+0xc5
flylinkdc!__acrt_lock_stream_and_call<<lambda_4f2c1eaeead2a5fc776db5b62ea0fb9b> >+0x404
flylinkdc!__stdio_common_vsprintf+0x1f
flylinkdc!vsnprintf+0x23
flylinkdc!snprintf+0x16
flylinkdc!libtorrent::upnp::delete_port_mapping+0x97
flylinkdc!std::_Func_impl<std::_Binder<std::_Unforced,void (__thiscall libtorrent::upnp::*)(libtorrent::upnp::rootdevice &,int),std::shared_ptr<libtorrent::upnp>,std::reference_wrapper<libtorrent::upnp::rootdevice>,int &>,std::allocator<int>,void,libtorrent::http_connection &>::_Do_call+0x13
flylinkdc!libtorrent::http_connection::on_connect+0x84
flylinkdc!std::invoke<void (__thiscall libtorrent::http_connection::*&)(boost::system::error_code const &),std::shared_ptr<libtorrent::http_connection> &,boost::system::error_code const &>+0x16
kernelbase!GetQueuedCompletionStatus+0x29
flylinkdc!boost::asio::detail::win_iocp_io_service::do_one+0x1cd
flylinkdc!boost::asio::detail::win_iocp_io_service::run+0xc8
flylinkdc!boost::asio::io_service::run+0x2a
flylinkdc!std::_Pad::_Call_func+0x50
flylinkdc!wcsftime+0xf5
kernel32!BaseThreadInitThunk+0xe
ntdll!__RtlUserThreadStart+0x70
ntdll!_RtlUserThreadStart+0x1b

pavel-pimenov:
- New crash dump https://drdump.com/Problem.aspx?ProblemID=241665&Login=Guest
 - Remove raw pointer and static std::string service_type;

### DIFF
--- a/src/upnp.cpp
+++ b/src/upnp.cpp
@@ -225,7 +225,7 @@ int upnp::add_mapping(portmap_protocol const p, int const external_port
 		m.external_port = external_port;
 		m.local_port = local_port;
 
-		if (d.service_namespace) update_map(d, mapping_index);
+		if (!d.service_namespace.empty()) update_map(d, mapping_index);
 	}
 
 	return mapping_index;
@@ -255,7 +255,7 @@ void upnp::delete_mapping(int mapping)
 		TORRENT_ASSERT(mapping < int(d.mapping.size()));
 		d.mapping[mapping].act = mapping_t::action::del;
 
-		if (d.service_namespace) update_map(d, mapping);
+		if (!d.service_namespace.empty()) update_map(d, mapping);
 	}
 }
 
@@ -699,7 +699,7 @@ void upnp::post(upnp::rootdevice const& d, char const* soap
 		"Soapaction: \"%s#%s\"\r\n\r\n"
 		"%s"
 		, d.path.c_str(), d.hostname.c_str(), d.port
-		, int(strlen(soap)), d.service_namespace, soap_action
+		, int(strlen(soap)), d.service_namespace.c_str(), soap_action
 		, soap);
 
 	d.upnp_connection->m_sendbuffer = header;
@@ -743,7 +743,7 @@ void upnp::create_port_mapping(http_connection& c, rootdevice& d, int i)
 		"<NewPortMappingDescription>%s at %s:%d</NewPortMappingDescription>"
 		"<NewLeaseDuration>%u</NewLeaseDuration>"
 		"</u:%s></s:Body></s:Envelope>"
-		, soap_action, d.service_namespace, d.mapping[i].external_port
+		, soap_action, d.service_namespace.c_str(), d.mapping[i].external_port
 		, (d.mapping[i].protocol == portmap_protocol::udp ? "UDP" : "TCP")
 		, d.mapping[i].local_port
 		, local_endpoint.c_str()
@@ -795,7 +795,7 @@ void upnp::update_map(rootdevice& d, int i)
 	}
 
 	TORRENT_ASSERT(!d.upnp_connection);
-	TORRENT_ASSERT(d.service_namespace);
+	TORRENT_ASSERT(!d.service_namespace.empty());
 
 #ifndef TORRENT_DISABLE_LOGGING
 	log("connecting to %s", d.hostname.c_str());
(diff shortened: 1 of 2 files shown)


src/upnp.cpp
```diff
@@ -225,7 +225,7 @@ int upnp::add_mapping(portmap_protocol const p, int const external_port
 		m.external_port = external_port;
 		m.local_port = local_port;
 
-		if (d.service_namespace) update_map(d, mapping_index);
+		if (!d.service_namespace.empty()) update_map(d, mapping_index);
```
gubatron: why not

`if (d.service_namespace && !d.service_namespace.empty()) update_map(d, mapping_index);`

and avoid a possible NPE?

The same comment applies everywhere else below, and since it happens so much perhaps I'd add a private utility function to not repeat yourself all over

```
bool valid_service_namespace(rootdevice &d) {
    return d.service_namespace && !d.service_namespace.empty();
}
```

and then have

```
if (valid_service_namespace(d)) update_map(d, mapping_index);

...

TORRENT_ASSERT(valid_service_namespace(d));

...
```

src/upnp.cpp
```diff
@@ -996,9 +998,10 @@ void upnp::on_upnp_xml(error_code const& e
 		d.disabled = true;
 		return;
 	}
-	static std::string service_type;
-	service_type.swap(s.service_type);
-	d.service_namespace = service_type.c_str();
+	d.service_namespace = s.service_type;
```
gubatron: I wonder if it's ok to do this, maybe that swap was there for a good reason.
Also, it did a `.c_str()`, and now I see you've had to put `.c_str()` everywhere because you don't do it here it seems.

gubatron: oh, I see, you do this so that you can do the `.empty()` check.

arvidn:
First of all, this fix should be made against RC_1_1, that's the version you're seeing this problem in, right?
Second; the existing code is clearly broken, I don't understand how it could get so screwed up. However, I'm tempted to fix it by restoring it instead of entirely switching to a ``std::string`` approach.
The original code (as far as I can tell from the intention of the existing code at least) meant for ``service_namespace`` to point to a string literal. There are only 3 choices of what this string can be, so that kind of makes sense. Alternatively, we could make it and index and have a lookup table for the actual namespace name.

arvidn:
going back to RC_1_0 though, it doesn't look like this was ever right. going with ``std::string`` may be the better option

arvidn:
this patch looks good to me. would you mind submitting it against RC_1_1?

src/upnp.cpp
pavel-pimenov: service_namespace -> std::string
if(string()) - impossible

int main()
{
  std::string name;
  if(name)   {
     }
}

 In function 'int main()': 8:10: error: could not convert 'name' from 'std::string {aka std::basic_string<char>}' to 'bool'

arvidn:
this has been merged into master from RC_1_1 now

PR #1226, opened 2016-10-19
ensure posix calls with LFS support are used in android

aldenml:
This was a hard one to find. There is no mention of `_FILE_OFFSET_BITS` in any header in the android toolchain. Without this, android picks the calls with `off_t` (instead of `off64_t`) and libtorrent fails while writing.

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -124,6 +124,9 @@ POSSIBILITY OF SUCH DAMAGE.
 #ifdef TORRENT_ANDROID
 #include <sys/syscall.h>
 #define lseek lseek64
+#define pread pread64
+#define pwrite pwrite64
+#define ftruncate ftruncate64
 #endif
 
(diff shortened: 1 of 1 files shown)


src/file.cpp
```diff
@@ -124,6 +124,9 @@ POSSIBILITY OF SUCH DAMAGE.
 #ifdef TORRENT_ANDROID
 #include <sys/syscall.h>
 #define lseek lseek64
+#define pread pread64
+#define pwrite pwrite64
+#define ftruncate ftruncate64
```
arvidn: you don't need to define ``off_t`` as ``off64_t``?

arvidn: I guess ``file.cpp`` doesn't use ``off_t``

arvidn:
are you sure you don't want this in RC_1_1? It seems like something that should go there

aldenml:
I'm using `master` for release already, I rather get this first in master since the merge from `RC_1_1` usually takes days. I will create a back port PR.

PR #1227, opened 2016-10-19
extend the cpplint filters used on travis

arvidn:


src/peer_connection.cpp
```diff
@@ -1093,7 +1092,7 @@ namespace libtorrent
 		{
 			TORRENT_TRY {
 				e->on_piece_pass(index);
-			} TORRENT_CATCH (std::exception const&) {}
```
aldenml: I think the linter get this wrong, since it's `} catch (`, not `} catch(`

aldenml: I wonder if using just `try` and `catch` (but still consider `BOOST_NO_EXCEPTIONS`) is better.

arvidn: I intend to retire these macros eventually

PR #1228, opened 2016-10-19
fix ambiguous call to abs() in tommath

arvidn:


nrosier:
Error remains the same:

`mpi.cpp:6711:21: error: call to 'abs' is ambiguous
    d = ((mp_digit) ::abs (MP_GEN_RANDOM())) & MP_MASK;
                    ^~~~~
/usr/include/stdlib.h:83:6: note: candidate function
int      abs(int) __pure2;
         ^
/usr/include/c++/v1/stdlib.h:115:44: note: candidate function
inline _LIBCPP_INLINE_VISIBILITY long      abs(     long __x) _NOEXCEPT {return  labs(__x);}
                                           ^
/usr/include/c++/v1/stdlib.h:117:44: note: candidate function
inline _LIBCPP_INLINE_VISIBILITY long long abs(long long __x) _NOEXCEPT {return llabs(__x);}
                                           ^
mpi.cpp:6723:41: error: call to 'abs' is ambiguous
    if ((res = mp_add_d (a, ((mp_digit) ::abs (MP_GEN_RANDOM())), a)) != MP_OKAY) {
                                        ^~~~~
/usr/include/stdlib.h:83:6: note: candidate function
int      abs(int) __pure2;
         ^
/usr/include/c++/v1/stdlib.h:115:44: note: candidate function
inline _LIBCPP_INLINE_VISIBILITY long      abs(     long __x) _NOEXCEPT {return  labs(__x);}
                                           ^
/usr/include/c++/v1/stdlib.h:117:44: note: candidate function
inline _LIBCPP_INLINE_VISIBILITY long long abs(long long __x) _NOEXCEPT {return llabs(__x);}
                                           ^
2 errors generated.`

arvidn:
it is very strange that the C++ overloads and the C function are both visible here. it's as if there is a "using namespace std;" somewhere..

arvidn:
just trying to call ``std::abs()`` won't work either, since the C header is included. Perhaps casting the argument to ``int`` would do it. Could you test this? If not, maybe the combination of qualifying the global namespace and the cast would do it

PR #1231, opened 2016-10-19
Prevent duplicate url

pavel-pimenov:
Fix #1188 (part 2)


src/create_torrent.cpp
```diff
@@ -664,11 +664,16 @@ namespace libtorrent
 
 	void create_torrent::add_tracker(string_view url, int const tier)
 	{
+		auto i = std::find_if(m_urls.begin(), m_urls.end()
+			, [&url](announce_entry const& ae) { return ae.first == url.to_string(); });
+		if (i == m_urls.end())
```
arvidn: what do you think about: ``if (i != m_urls.end()) return;``?

arvidn:
lgtm

PR #1235, opened 2016-10-20
utp_match: Reduced call boost::asio::ip::address address()

pavel-pimenov:
~30% CPU s->m_recv_id != id (false)

### DIFF
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -719,9 +719,9 @@ bool utp_incoming_packet(utp_socket_impl* s
 
 bool utp_match(utp_socket_impl* s, udp::endpoint const& ep, std::uint16_t id)
 {
-	return s->m_remote_address == ep.address()
+	return s->m_recv_id == id
 		&& s->m_port == ep.port()
-		&& s->m_recv_id == id;
+		&& s->m_remote_address == ep.address();
 }
 
 udp::endpoint utp_remote_endpoint(utp_socket_impl* s)


arvidn:
the 30% CPU must be for a debug/unoptimized build, right?

pavel-pimenov:
Thank you for merge
i use profiler VC++2015 and order by call 
bool utp_match(utp_socket_impl* s, udp::endpoint const& ep, std::uint16_t id)
{
#ifdef _DEBUG	
	if (s->m_recv_id == id)
	{
		static unsigned g_count = 0;
		++g_count;
	}
	else
	{
		static unsigned g_count = 0;
		++g_count;
	}
#endif
libtorrent-utpmath

arvidn:
my point was that in an optimized build, I would expect this change to have negligible impact

pavel-pimenov:
But C++ optimizer has no right to swap sequence for logical expression? or not?
https://msdn.microsoft.com/en-us/library/c6s3h5a7.aspx
The second operand is evaluated only if the first operand evaluates to true (nonzero). This evaluation eliminates needless evaluation of the second operand when the logical AND expression is false. ...

ssiloti:
Just because a function is being called a lot doesn't mean it is using a significant amount of CPU time. `utp_match` is a small function which I'm sure consumes a negligible amount of CPU. That being said, there's nothing wrong with this change in principle. The call to `endpoint::address` is pretty cheap but, perhaps surprisingly, not trivial so it makes sense to put the simple integer comparisons first.

arvidn:
@pavel-pimenov under the as-if rule, once these functions are inlined, it can determine that neither have any side effect and perform the comparisons in whatever order it likes, or in parallel.

That said, there is certainly value in not wasting CPU for debug builds.

Issue #1225, opened 2016-10-18
crash in android in disk_io_thread::try_flush_hashed


aldenml:
**Please provide the following information**

libtorrent version (or branch): master

platform/architecture: android

compiler and compiler version: clang NDK 13

```
pid: 21826, tid: 21936, name: Thread-7015  >>> com.frostwire.android <<<
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0xdaa00000
    r0 00000000  r1 00000010  r2 00000010  r3 da9fffe4
    r4 00000010  r5 0000000e  r6 00000080  r7 00000010
    r8 040000d6  r9 00400100  sl 00000010  fp dd8fa7f8
    ip 00000010  sp dd8fa798  lr 00000001  pc dfdeefd8  cpsr 800f0010

backtrace:
    #00 pc 00221fd8  /data/app/com.frostwire.android-2/lib/arm/libjlibtorrent.so (_ZN10libtorrent14disk_io_thread16try_flush_hashedEPNS_18cached_piece_entryEiRNS_9tailqueueINS_11disk_io_jobEEERNSt6__ndk111unique_lockINS7_5mutexEEE+340)
    #01 pc 00228a18  /data/app/com.frostwire.android-2/lib/arm/libjlibtorrent.so (_ZN10libtorrent14disk_io_thread15do_flush_hashedEPNS_11disk_io_jobERNS_9tailqueueIS1_EE+252)
    #02 pc 00223d6c  /data/app/com.frostwire.android-2/lib/arm/libjlibtorrent.so (_ZN10libtorrent14disk_io_thread11perform_jobEPNS_11disk_io_jobERNS_9tailqueueIS1_EE+144)
    #03 pc 0022947c  /data/app/com.frostwire.android-2/lib/arm/libjlibtorrent.so (_ZN10libtorrent14disk_io_thread10thread_funENS0_13thread_type_tEN5boost4asio10io_service4workE+436)
    #04 pc 0022a6a8  /data/app/com.frostwire.android-2/lib/arm/libjlibtorrent.so (_ZN10libtorrent14disk_io_thread9job_queue10thread_funEN5boost4asio10io_service4workE+56)
[...]

PR #1240, opened 2016-10-21
use of piece_refcount_holder in disk_io_thread::do_flush_piece

aldenml:
I see the code in disk_io_thread::do_flush_hashed and it seems to me like this code should be added to disk_io_thread::do_flush_piece. But I have no deep knowledge of the logic behind. Related to https://github.com/arvidn/libtorrent/issues/1225

### DIFF
--- a/src/disk_io_thread.cpp
+++ b/src/disk_io_thread.cpp
@@ -2792,11 +2792,19 @@ namespace libtorrent
 		cached_piece_entry* pe = m_disk_cache.find_piece(j);
 		if (pe == nullptr) return 0;
 
+		piece_refcount_holder refcount_holder(pe);
+
 #if TORRENT_USE_ASSERTS
 		pe->piece_log.push_back(piece_log_t(j->action));
 #endif
 		try_flush_hashed(pe, m_settings.get_int(
 			settings_pack::write_cache_line_size), completed_jobs, l);
+
+		TORRENT_ASSERT(l.owns_lock());
+
+		refcount_holder.release();
+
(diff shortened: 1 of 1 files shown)


src/disk_io_thread.cpp
```diff
@@ -2792,12 +2792,20 @@ namespace libtorrent
 		cached_piece_entry* pe = m_disk_cache.find_piece(j);
 		if (pe == nullptr) return 0;
 
+		piece_refcount_holder refcount_holder(pe);
+
 #if TORRENT_USE_ASSERTS
 		pe->piece_log.push_back(piece_log_t(j->action));
 #endif
 		try_flush_hashed(pe, m_settings.get_int(
```
arvidn: ``do_flush_hashed()`` already increments the refcount while flushing, this shouldn't be necessary.

arvidn: oh, I guess you do this in order to guarantee that ``pe`` is still around after the call, to call ``maybe_free_piece`` on it. do you have a theory for not doing that causing that segfault?

aldenml: sometimes the problem comes from `do_flush_hashed` and sometimes (much less) from `do_flush_piece`. My question (more than a theory) is: why both methods don't handle the call to `try_flush_hashed` in the same way?

Maybe this is the missing "memory pin" that prevents the segfault, but frankly, I don't have a clear idea on this.

aldenml:
reading myself here...clueless PR. Closing it.

PR #1243, opened 2016-10-22
return a span from TORRENT_ALLOCA

ssiloti:
This causes double evaluation of the size parameter which is unfortunate
but unavoidable due to the spooky voodoo that is alloca.

Alternate hack if this one is just too ugly (the alternative isn't much prettier): https://github.com/ssiloti/libtorrent/commit/b760e68e2611c667f51251ae90fa5d31aba03903

### DIFF
--- a/include/libtorrent/alloca.hpp
+++ b/include/libtorrent/alloca.hpp
@@ -33,21 +33,32 @@ POSSIBILITY OF SUCH DAMAGE.
 #ifndef TORRENT_ALLOCA
 
 #include "libtorrent/config.hpp"
+#include "libtorrent/span.hpp"
+
+// this is part of a clever (perhaps too clever) hack to safely return a span
+// over alloca'd data from an expression. Due to alloca's unique properties
+// it is not safe to directly call it in the parameter list of span's ctor.
+// Instead use a thread local temporary to store the value then use the comma
+// operator to call span's ctor and return it to the caller.
+extern thread_local void* TORRENT_ALLOCA_tmp;
 
 #if defined TORRENT_WINDOWS || defined TORRENT_MINGW
 
 #include <malloc.h>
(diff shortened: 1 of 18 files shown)


ssiloti:
Crap, you can't export thread local variables so tests are broken. On to plan B then.

include/libtorrent/alloca.hpp
```diff
@@ -33,21 +33,32 @@ POSSIBILITY OF SUCH DAMAGE.
...
 
 #include "libtorrent/config.hpp"
+#include "libtorrent/span.hpp"
+
+// this is part of a clever (perhaps too clever) hack to safely return a span
+// over alloca'd data from an expression. Due to alloca's unique properties
+// it is not safe to directly call it in the parameter list of span's ctor.
+// Instead use a thread local temporary to store the value then use the comma
+// operator to call span's ctor and return it to the caller.
+extern thread_local void* TORRENT_ALLOCA_tmp;
```
arvidn: surprisingly, the runtime on macOS (last time I checked) does not support the ``thread_local`` feature.

include/libtorrent/storage.hpp
```diff
@@ -149,7 +149,7 @@ namespace libtorrent
 	struct add_torrent_params;
 
 	TORRENT_EXTRA_EXPORT int copy_bufs(file::iovec_t const* bufs, int bytes, file::iovec_t* target);
-	TORRENT_EXTRA_EXPORT void advance_bufs(file::iovec_t*& bufs, int bytes);
+	TORRENT_EXTRA_EXPORT span<file::iovec_t> advance_bufs(span<file::iovec_t> bufs, int bytes);
```
arvidn: I think these changes to take a span still make sense though

PR #1244, opened 2016-10-22
return a span from TORRENT_ALLOCA

ssiloti:
Unfortunately this requires moving the variable declaration inside the macro.
Due to alloca’s unique properties pretty much the only safe way to call it is
in a simple assign-to-pointer expression. Therefor we need to use a temporary
pointer to store the value returned from alloca before we can call span’s ctor.

This also causes double evaluation of the size parameter which is unfortunate,
but no current callers of TORRENT_ALLOCA have a problem with this. Passing
expressions with side effects to macros is bad mojo anyways.

### DIFF
--- a/include/libtorrent/alloca.hpp
+++ b/include/libtorrent/alloca.hpp
@@ -33,21 +33,28 @@ POSSIBILITY OF SUCH DAMAGE.
 #ifndef TORRENT_ALLOCA
 
 #include "libtorrent/config.hpp"
+#include "libtorrent/span.hpp"
 
 #if defined TORRENT_WINDOWS || defined TORRENT_MINGW
 
 #include <malloc.h>
-#define TORRENT_ALLOCA(t, n) static_cast<t*>(_alloca(sizeof(t) * (n)))
+#define TORRENT_ALLOCA(v, t, n) ::libtorrent::span<t> v; { \
+	t* TORRENT_ALLOCA_tmp = static_cast<t*>(_alloca(sizeof(t) * (n))); \
+	v = ::libtorrent::span<t>(TORRENT_ALLOCA_tmp, n); }
 
 #elif defined TORRENT_BSD
 
 #include <stdlib.h>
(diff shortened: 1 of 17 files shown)


include/libtorrent/alloca.hpp
```diff
@@ -33,21 +33,28 @@ POSSIBILITY OF SUCH DAMAGE.
 #ifndef TORRENT_ALLOCA
 
 #include "libtorrent/config.hpp"
+#include "libtorrent/span.hpp"
 
 #if defined TORRENT_WINDOWS || defined TORRENT_MINGW
 
 #include <malloc.h>
-#define TORRENT_ALLOCA(t, n) static_cast<t*>(_alloca(sizeof(t) * (n)))
+#define TORRENT_ALLOCA(v, t, n) ::libtorrent::span<t> v; { \
```
arvidn: oh, you specify the name of the local variable you want the span to be assigned to.. I see. because if you ever misuse the macro as: ``auto mem = TORRENT_ALLOCA()`` you'll always end up with an empty span.

include/libtorrent/disk_buffer_pool.hpp
```diff
@@ -77,7 +78,7 @@ namespace libtorrent
 		char* allocate_buffer(bool& exceeded, std::shared_ptr<disk_observer> o
 			, char const* category);
 		void free_buffer(char* buf);
-		void free_multiple_buffers(char** bufvec, int numbufs);
+		void free_multiple_buffers(span<char*> bufvec);
```
arvidn: heh, I just encountered this a few hours ago. +1

arvidn:
lgtm. I'll take another look in the morning

include/libtorrent/alloca.hpp
ssiloti: You'll get a compilation error if you do that. The inability to support that syntax is that reason why I had to do it this way. There's no way to create a temporary from within an expression.

PR #1248, opened 2016-10-23
add unit test for 'file_sizes' entry in resume data

arvidn:


test/test_resume.cpp
```diff
@@ -192,6 +192,58 @@ void default_tests(torrent_status const& s)
...
+	lt::session ses(pack);
+	boost::shared_ptr<torrent_info> ti = generate_torrent();
+	add_torrent_params p;
+	p.ti = ti;
+	p.save_path = ".";
+	if (allocate) p.storage_mode = storage_mode_allocate;
+	torrent_handle h = ses.add_torrent(p);
+
+	std::vector<char> piece(ti->piece_length(), 0);
+	h.add_piece(0, piece.data());
```
arvidn: before saving resume data, we need to synchronize with the piece actually being written here. by waiting for the ``piece_finished_alert``

PR #1249, opened 2016-10-24
add utility function for generating peer ID fingerprint

arvidn:


aldenml:
It's like `fingerprint.cpp` was not added.

include/libtorrent/fingerprint.hpp
```diff
@@ -38,92 +38,65 @@ POSSIBILITY OF SUCH DAMAGE.
...
+	// | 'XT'     | Xan Torrent           |
+	// +----------+-----------------------+
+	// 
+	// There's an informal directory of client id's here_.
+	// 
+	// .. _here: http://wiki.theory.org/BitTorrentSpecification#peer_id
+	//
+	// The ``major``, ``minor``, ``revision`` and ``tag`` parameters are used to
+	// identify the version of your client.
+	TORRENT_EXPORT std::string gen_fingerprint(std::string name
```
aldenml: Would you consider a more explicit name, like `generate_fingerprint`?

include/libtorrent/fingerprint.hpp
```diff
@@ -38,92 +38,65 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "libtorrent/config.hpp"
```
sledgehammer999: I think you can also remove cstdio which doesn't show in this diff.

sledgehammer999:
The API seems the same pretty much for me so I am ok. I just wanted a utility function to continue using the de facto standard for declaring client version.

I cannot comment if you use correctly the TORRENT_DEPRECATED macros. I don't have that knowledge. But you seem to use both TORRENT_DEPRECATED and TORRENT_DEPRECATED_EXPORT for the struct and then again guard one of its member functions.

arvidn:
yeah, there's still some things to iron out there

sledgehammer999:
@arvidn you didn't remove the cstdio include in fingerprint.hpp

PS: Aren't you concerned about keeping a relative clean history? Why don't you use `git rebase -i` to rearrange/squash/fixup temporary commits before merging? This is a curiosity of mine. No need to answer if you're busy.

PR #1253, opened 2016-10-24
include docs/makefile in tarball

arvidn:


sledgehammer999:
Debian packages already apply a patch that creates this file that is missing from the tarball. And I was able to successfully build 1.0.10 with it.
So I suppose including it in the tarball won't hurt. The debian will need minor adjusting.
So it is your call if you want to start distributing it.

PR #1254, opened 2016-10-25
fishing for #1225

ssiloti:
Changes triggered by the hunt for the root cause of #1225

### DIFF
--- a/src/disk_io_thread.cpp
+++ b/src/disk_io_thread.cpp
@@ -2088,6 +2088,9 @@ namespace libtorrent
 
 		// save a local copy of offset to avoid concurrent access
 		int offset = ph->offset;
+#if TORRENT_USE_ASSERTS
+		int old_offset = offset;
+#endif
 
 		l.unlock();
 
@@ -2105,6 +2108,7 @@ namespace libtorrent
 
 		l.lock();
 
+		TORRENT_ASSERT(old_offset == ph->offset);
 		ph->offset = offset;
 
 		TORRENT_PIECE_ASSERT(pe->hashing, pe);
@@ -2328,6 +2332,9 @@ namespace libtorrent
 
 		// save a local copy of offset to avoid concurrent access
 		int offset = ph->offset;
+#if TORRENT_USE_ASSERTS
+		int old_offset = offset;
+#endif
 
 		l.unlock();
 
@@ -2421,6 +2428,7 @@ namespace libtorrent
 
 		l.lock();
 
+		TORRENT_ASSERT(old_offset == ph->offset);
 		ph->offset = offset;
 
 		// decrement the refcounts of the blocks we just hashed
--- a/src/disk_io_thread.cpp
+++ b/src/disk_io_thread.cpp
@@ -2086,6 +2086,9 @@ namespace libtorrent
 		DLOG("kick_hasher: %d - %d (piece: %d offset: %d)\n"
 			, cursor, end, int(pe->piece), ph->offset);
 
+		// save a local copy of offset to avoid concurrent access
+		int offset = ph->offset;
+
 		l.unlock();
 
 		time_point start_time = clock_type::now();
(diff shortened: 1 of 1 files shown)


src/disk_io_thread.cpp
```diff
@@ -2319,22 +2326,25 @@ namespace libtorrent
...
 			if (next_locked_block < num_locked_blocks
 				&& locked_blocks[next_locked_block] == i)
 			{
 				++next_locked_block;
 				TORRENT_PIECE_ASSERT(pe->blocks[i].buf, pe);
-				TORRENT_PIECE_ASSERT(ph->offset == i * block_size, pe);
-				ph->offset += int(iov.iov_len);
+				TORRENT_PIECE_ASSERT(offset == i * block_size, pe);
+				offset += int(iov.iov_len);
 				ph->h.update({pe->blocks[i].buf, iov.iov_len});
```
arvidn: there better only be a single thread calling ``update()`` here. that's guaranteed, right?

ssiloti: Yes, I verified that the hasher is not accessed concurrently.

src/disk_io_thread.cpp
```diff
@@ -2084,22 +2086,27 @@ namespace libtorrent
 		DLOG("kick_hasher: %d - %d (piece: %d offset: %d)\n"
```
arvidn: we set ``hashing = 1`` right above here. as far as I can tell, we don't release the mutex between checking ``hashing`` and setting it, meaning that there really only should be one thread getting here at a time, no?

src/disk_io_thread.cpp
arvidn: I suppose some other thread may still read ``offset``, which justifies this fix

src/disk_io_thread.cpp
ssiloti: There are other places that offset is read without checking `hashing`, in particular `try_flush_hashed`.

src/disk_io_thread.cpp
```diff
@@ -2411,6 +2421,8 @@ namespace libtorrent
 
 		l.lock();
 
+		ph->offset = offset;
```
arvidn: should there be a TORRENT_ASSERT(ph->offset == offset); here and above too?

arvidn:
awesome. nice catch!

arvidn:
is this a problem in RC_1_1 as well?

src/disk_io_thread.cpp
ssiloti: And to answer you question, yes, only one thread should be kicking the hasher at a time.

ssiloti:
I'm not sure it actually is a problem. Unless the compiler is doing something truly evil (to be fair, it wouldn't be the first time) this shouldn't cause a problem on any remotely common ISA. Reliance on 32-bit loads/stores being atomic is very common practice.

arvidn:
I think the fact that it's a race condition (or rather that the compiler can assume there's no concurrent access) opens up all kinds of possibilities that won't necessarily have intuitive behavior.

src/disk_io_thread.cpp
```diff
@@ -2084,22 +2086,31 @@ namespace libtorrent
 		DLOG("kick_hasher: %d - %d (piece: %d offset: %d)\n"
 			, cursor, end, int(pe->piece), ph->offset);
 
+		// save a local copy of offset to avoid concurrent access
+		int offset = ph->offset;
+#if TORRENT_USE_ASSERTS
+		int old_offset = offset;
```
arvidn: oh, right. my first pass reading this, I assumed ``offset`` was immutable. then I realized it wasn't, but I was still thinking that the assert I proposed would work.. oh well

arvidn:
lgtm

ssiloti:
Alright, if you want it on RC_1_1 then I'll rebase and re-submit.

PR #1259, opened 2016-10-26
removed spaces in template types and more c++11 auto/loop refactor

aldenml:


test/test_ip_filter.cpp
```diff
@@ -61,9 +61,9 @@ bool compare(ip_range<Addr> const& lhs
 }
 
 template <class T>
-void test_rules_invariant(std::vector<ip_range<T> > const& r, ip_filter const& f)
+void test_rules_invariant(std::vector<ip_range<T>> const& r, ip_filter const& f)
 {
-	typedef typename std::vector<ip_range<T> >::const_iterator iterator;
+	typedef typename std::vector<ip_range<T>>::const_iterator iterator;
```
arvidn: it seems like this would be unnecessary now too

test/test_piece_picker.cpp
```diff
@@ -225,7 +225,7 @@ bool verify_pick(std::shared_ptr<piece_picker> p
 	// make sure there are no duplicated
 	std::set<piece_block> blocks;
 	std::copy(picked.begin(), picked.end()
-		, std::insert_iterator<std::set<piece_block> >(blocks, blocks.end()));
+		, std::insert_iterator<std::set<piece_block>>(blocks, blocks.end()));
```
arvidn: hm. it seems like the iterator here could just be constructed with: std::back_inserter(blocks)

arvidn:
lgtm

PR #1260, opened 2016-10-26
use span for scatter/gather lists

ssiloti:


src/disk_buffer_pool.cpp
```diff
@@ -219,21 +219,22 @@ namespace libtorrent
...
-			if (iov[i].iov_base == nullptr)
+			i.iov_base = allocate_buffer_impl(l, "pending read");
+			i.iov_len = block_size();
+			if (i.iov_base == nullptr)
 			{
 				// uh oh. We failed to allocate the buffer!
 				// we need to roll back and free all the buffers
 				// we've already allocated
-				for (int j = 0; j < i; ++j)
+				for (auto j : iov)
```
aldenml: I know that probably the extra cycles will have zero impact, but I find more explicit the algorithm idea as it was, even if it's trivial. Just an opinion.

src/file.cpp
```diff
@@ -336,11 +336,11 @@ static_assert((libtorrent::file::sparse & libtorrent::file::attribute_mask) == 0
 
 namespace libtorrent
 {
-	int bufs_size(file::iovec_t const* bufs, int num_bufs)
+	int bufs_size(span<file::iovec_t const> bufs)
 	{
 		std::size_t size = 0;
-		for (file::iovec_t const* i = bufs, *end(bufs + num_bufs); i < end; ++i)
-			size += i->iov_len;
+		for (auto buf : bufs)
```
aldenml: I see that sometimes you use `auto` and others `auto&`, any reason not to always use `auto&` (or even `auto const&`)?

src/storage.cpp
```diff
@@ -225,18 +221,15 @@ namespace libtorrent
...
-			, int const size
-			, file::iovec_t const* bufs, storage_error& ec)
+			, span<file::iovec_t const> bufs, storage_error& ec)
 			final
 		{
 			if (m_storage.files().pad_file_at(file_index))
 			{
 				// writing to a pad-file is a no-op
-				return size;
+				return bufs_size(bufs);
```
aldenml: was this a bug?

arvidn: this function used to take the size, as pre-computed. now it doesn't and have to compute it itself

arvidn: I would expect it to be negligible, if any

aldenml: ok

src/file.cpp
ssiloti: In general I prefer to handle spans by value. I only use auto& when I need to modify the entries.

src/storage.cpp
ssiloti: If anything I expect performance to be better. Writing to pad files should happen relatively infrequently compared to writes to regular files and this approach saves an iteration over `bufs` in the regular file case.

src/file.cpp
aldenml: hmm...but I think will end up copying memory if the objects are not trivial in the general case, am I wrong? (not a problem in this case)

ssiloti: If the objects you're dealing with are large then yes, you would want to use auto const&. I was speaking specifically about spans which are small and trivial to copy so passing by value is preferable.

PR #1261, opened 2016-10-26
transition to more default member initializers

arvidn:


include/libtorrent/disk_io_job.hpp
```diff
@@ -72,6 +72,8 @@ namespace libtorrent
 	{
 		disk_io_job();
 		~disk_io_job();
+		disk_io_job(disk_io_job const&) = delete;
+		disk_io_job& operator=(disk_io_job const&) = delete;
```
aldenml: does this means `boost::noncopyable` can be removed from the type declaration?

PR #1265, opened 2016-10-27
fix deprecated warning with msvc

aldenml:


aldenml:
do you have something in mind for the expired certificates?

PR #1270, opened 2016-10-29
fix torrent_status time_since_upload/download counters for paused torrents

arvidn:


arvidn:
this (or something similar) needs to be merged into RC_1_1 as well

arvidn:
confirmed that this is not a problem in 1.1

PR #1275, opened 2016-10-30
documentation and deprecation fixes

arvidn:


include/libtorrent/settings_pack.hpp
```diff
@@ -173,11 +173,18 @@ namespace libtorrent
 			// IPs and multiple ports. Binding to port 0 will make the
 			// operating system pick the port. The default is "0.0.0.0:6881", which
 			// binds to all interfaces on port 6881.
-			//
+			// 
+			// If binding fails because the port is busy, the port number will be
+			// incremented by one, ``settings_pack::max_retry_port_bind`` times.
```
cas--: Perhaps add in brackets that default retry number is 10 for `max_retry_port_bind` as I had to look it up to verify value.

arvidn: the defaults are currently parsed out from the source, and it's not sophisticated enough to provide substitution variables anywhere in the docs. I would be worried about it becoming out of sync.

cas--: Fair enough :+1:

PR #1277, opened 2016-10-31
fix build warnings/errors, typos and minor cleanup

aldenml:


src/session_impl.cpp
```diff
@@ -5162,11 +5175,14 @@ namespace aux {
 		// proxy, and it's the same one as we're using for the tracker
 		// just tell the tracker the socks5 port we're listening on
 		if (m_socks_listen_socket && m_socks_listen_socket->is_open())
-			return m_socks_listen_socket->local_endpoint().port();
+		{
+			error_code ec;
+			return m_socks_listen_socket->local_endpoint(ec).port();
```
arvidn: is ignoring the error here the right thing to do? would you expect this to legitimately fail? without it being serious?

src/torrent.cpp
```diff
@@ -1001,7 +1000,7 @@ namespace libtorrent
 		{
 			throw;
 		}
-		catch (system_error const& err)
+		catch (boost::system::system_error const& err)
```
arvidn: I thought there was a typedef pulling ``system_error`` into the libtorrent namespace, no?

Jamfile
```diff
@@ -556,10 +547,10 @@ lib gcc : : <name>gcc <link>static ;
 lib libiconv : : <name>iconv <link>shared <search>/usr/local/lib ;
 
 # openssl on linux/bsd/macos etc.
-lib gcrypt : : <name>gcrypt <link>shared <search>/opt/local/lib <search>/usr/local/lib : : <include>/opt/local/include <include>/usr/local/include ;
+lib gcrypt : : <name>gcrypt <link>shared <search>/usr/local/lib : : <include>/usr/local/include ;
 lib z : : <link>shared <name>z <search>/usr/lib ;
-lib crypto : : <name>crypto <link>shared <use>z <search>/opt/local/lib : : <include>/opt/local/include ;
-lib ssl : : <name>ssl <link>shared <use>crypto <search>/opt/local/lib : : <include>/opt/local/include ;
```
arvidn: I agree that it's kind of weird to add these search paths. but as far as I know you need them on macOS. What's the rationale for removing them?

src/session_impl.cpp
```diff
@@ -174,14 +174,27 @@ namespace
...
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmissing-noreturn"
+#endif
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wmissing-noreturn"
+#endif
 namespace boost {
-	void throw_exception(std::exception const& e) { ::abort(); }
+	void throw_exception(std::exception const&) { ::abort(); }
```
aldenml: under `BOOST_NO_EXCEPTIONS` this functions needs to be user defined, it's actually used inside `asio`. See http://www.boost.org/doc/libs/1_62_0/libs/exception/doc/throw_exception.html

src/session_impl.cpp
aldenml: I will put the error check code.

src/torrent.cpp
aldenml: yes, but it's like it's not getting it, it just fail to compile, I will look at it

src/session_impl.cpp
arvidn: yeah, and the reason boost and asio doesn't define it as ``abort()`` or ``terminate()`` is because it's up to the application to decide what to do. it seems like it would make more sense for the client to decide, rather than libtorrent

aldenml: now I understand you... :)

Jamfile
aldenml: for me, it actually hurts, because if you install a custom `openssl` in another route, and still have one installed there, it's problematic (it happened to me), I don't think this is a standard route anymore for `openssl`. This is not critical, it's just getting the build as warn-less as possible.

arvidn: how do you specify where openssl is installed then?

src/session_impl.cpp
```diff
@@ -5162,11 +5154,21 @@ namespace aux {
 		// proxy, and it's the same one as we're using for the tracker
 		// just tell the tracker the socks5 port we're listening on
 		if (m_socks_listen_socket && m_socks_listen_socket->is_open())
-			return m_socks_listen_socket->local_endpoint().port();
+		{
+			error_code ec;
+			tcp::endpoint endp = m_socks_listen_socket->local_endpoint(ec);
+			if (ec)
```
arvidn: it's not obvious to me how this differs from the previous version. local_endpoint() throws an exception is it fails, right?

Jamfile
aldenml: I use `<cxxflags>` for the `-I` and `<linkflags>` for the `-L`

src/session_impl.cpp
aldenml: It's the same, but the method `local_endpoint()` without the parameter does not exists under `BOOST_NO_EXCEPTIONS` and fails to compile

arvidn: I see. I wonder if that should be changed instead.. it seems like that would be nicer. I take it you build with exceptions disabled then? is it a problem to have naked throw expressions? should they be ``boost::throw_exception`` calls instead? if so, does boost.asio use that, or is that also a problem?

aldenml: I'm experimenting with `BOOST_NO_EXCEPTIONS`, but still using exceptions enabled at compiler level. I'm not really that familiar with boost and asio thinking in order to say something about "naked throw expressions" vs `boost::throw_exception`.

What do you mean with "that should be changed instead"? The `socket_type` API? Do you want the `BOOST_NO_EXCEPTIONS` guard to be removed there?

arvidn: well, it's no more dangerous than before. It was just a class of errors that wasn't handled properly (and there are still outstanding issues to be fixed there). I think a reasonable thing to do when building with no exceptions is to terminate the process any time an exception would have been thrown.

I can do that for throws from within libtorrent, but if boost::asio or new, or any other part of the standard library, throws, I can't really control that, so I'm not sure what happens. It's up to the compiler/library implementation.

arvidn: and, yes. I did mean enabling the exception-throwing overloads in the ``socket_type`` interface btw.

arvidn: what would you think about deferring getting ``BOOST_NO_EXCEPTIONS`` to build to a different patch?

aldenml: sure, I will do that, let me revert the `BOOST_NO_EXCEPTIONS` changes.

I chose the wrong word, what I mean with "dangerous" is that I better get those exceptions and fail in java-land for better error reporting in the crash.

PR #1282, opened 2016-11-01
moved async-load-torrent feature out of disk-io-thread

arvidn:
into a generic worker thread

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -4502,12 +4502,12 @@ namespace aux {
 
 			m_torrent_load_thread->ios.post([params, this]
 			{
-				std::unique_ptr<add_torrent_params> holder(params);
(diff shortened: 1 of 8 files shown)


src/session_impl.cpp
```diff
@@ -4497,8 +4497,18 @@ namespace aux {
...
-			m_disk_thread.async_load_torrent(params
-				, std::bind(&session_impl::on_async_load_torrent, this, _1));
+			if (!m_torrent_load_thread)
+				m_torrent_load_thread.reset(new work_thread_t());
+
+			m_torrent_load_thread->ios.post([params, this]
+			{
+				std::unique_ptr<add_torrent_params> holder(params);
+				error_code ec;
+				params->ti.reset(new torrent_info(resolve_file_url(params->url), ec));
```
arvidn: oh right. this is a ``shared_ptr``..

PR #1287, opened 2016-11-03
treat an unknown key in a settings_pack as an error in python

arvidn:


bindings/python/src/session.cpp
```diff
@@ -110,7 +110,11 @@ namespace
 			std::string const key = extract<std::string>(iterkeys[i]);
 
 			int sett = setting_by_name(key);
-			if (sett < 0) continue;
+			if (sett < 0)
+			{
+				PyErr_SetString(PyExc_ValueError, ("unknown name in settings_pack: " + key).c_str());
```
cas--: `PyExc_ValueError` should be `PyExc_KeyError`

bindings/python/test.py
```diff
@@ -165,6 +165,13 @@ def test_post_session_stats(self):
 		self.assertTrue(isinstance(a.values, dict))
 		self.assertTrue(len(a.values) > 0)
 
+	def test_unknown_settings(self):
+		try:
+			s = lt.session({'unexpected-key-name': 42})
+			self.assertFalse('should have thrown an exception')
+		except Exception as e:
```
cas--: `Exception` is too general, it should match the thrown/raised error `KeyError` (based on change above)

PR #1299, opened 2016-11-06
reopen listen sockets when the system's IP changes

ssiloti:
Only Linux and Windows supported for now.

### DIFF
--- a/include/libtorrent/ip_notifier.hpp
+++ b/include/libtorrent/ip_notifier.hpp
@@ -53,18 +53,17 @@ namespace libtorrent
 	{
 		// cb will be invoked from within a thread calling ios.run()
 		// when a change is detected in the system's IP addresses
-		ip_change_notifier(io_service& ios, std::function<void()> cb);
+		ip_change_notifier(io_service& ios);
 		~ip_change_notifier();
 
 		// start listening for changes
-		void start();
-		void stop();
+		void async_wait(std::function<void(error_code const&)> cb);
+		void cancel();
 
 	private:
 		void on_notify(error_code const& error
-			, std::size_t bytes_transferred);
-
-		std::function<void()> m_cb;
+			, std::size_t bytes_transferred
+			, std::function<void(error_code const&)> cb);
 
 #if defined TORRENT_BUILD_SIMULATOR
 		// TODO simulator support
--- a/include/libtorrent/ip_notifier.hpp
+++ b/include/libtorrent/ip_notifier.hpp
@@ -51,12 +51,11 @@ namespace libtorrent
 {
 	struct ip_change_notifier : boost::noncopyable
 	{
-		// cb will be invoked from within a thread calling ios.run()
-		// when a change is detected in the system's IP addresses
-		ip_change_notifier(io_service& ios);
+		explicit ip_change_notifier(io_service& ios);
 		~ip_change_notifier();
 
-		// start listening for changes
+		// cb will be invoked  when a change is detected in the
+		// system's IP addresses
 		void async_wait(std::function<void(error_code const&)> cb);
 		void cancel();
 
--- a/include/libtorrent/ip_notifier.hpp
+++ b/include/libtorrent/ip_notifier.hpp
@@ -1,0 +1,81 @@
+/*
+
+Copyright (c) 2016, Steven Siloti
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+* Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+* Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in
+the documentation and/or other materials provided with the distribution.
+* Neither the name of the author nor the names of its
+contributors may be used to endorse or promote products derived
+from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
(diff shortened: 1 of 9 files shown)


include/libtorrent/ip_notifier.hpp
```diff
@@ -0,0 +1,81 @@
...
+{
+	struct ip_change_notifier : boost::noncopyable
+	{
+		// cb will be invoked from within a thread calling ios.run()
+		// when a change is detected in the system's IP addresses
+		ip_change_notifier(io_service& ios, std::function<void()> cb);
+		~ip_change_notifier();
+
+		// start listening for changes
+		void start();
```
arvidn: essentially all of libtorrent uses this pattern of "subscribing" to events, as opposed to the underlying asio model where you can only ask for one-shot events, and then re-subscribe. This has subtle implications of lifetime management on the object. I think in general, asio's way is more composable. It specifically simplifies shutting down. In this case, since the io_service is tied to the session object, the session destructor is synchronized with the io_service running out of work, which makes it all good. But in the simulator case, there can be multiple sessions sharing an io_service, which means it's critical that all handlers have been called before we can destruct the session. With a subscription interface, calling stop, there's no way of knowing whether there's an outstanding handler to be called afterwards, so it's not safe to destruct the session (and the ip_change_notifier along with it). The way I've dealt with this problem in ``lsd``, ``natpmp`` and ``upnp`` is to make those objects be heap allocated, managed by a ``shared_ptr`` and make the handler keep them alive. So, even it the session_object has destructed, the object is still alive to support receiving the last callback.

The problem with this approach is that every level of io-object needs to be heal allocated to deal with clean shutdown.

I don't think it's that big of a deal in this specific case, since the session object already suffers from this problem from code that would take a lot more to fix. However, you may want to consider moving the callback object from being passed into the constructor to the ``start()`` function and have the ``stop()`` function actually destruct it. That lets the callback keep something alive with a reference count. It probably means you need a boost::optional<std::function<void()>> too though.

just a thought.

arvidn: the main reason I mention this is because I think this is such a useful utility that it may make sense to create a separate git repository for it and use it as a submodule. And for it to be generic, I think it would need a slightly different interface.

src/ip_notifier.cpp
```diff
@@ -0,0 +1,117 @@
...
+#endif
+
+using namespace std::placeholders;
+
+namespace libtorrent
+{
+	namespace
+	{
+#if TORRENT_USE_NETLINK && !defined TORRENT_BUILD_SIMULATOR
+		netlink::endpoint ip_change_endpoint(netlink(NETLINK_ROUTE), RTMGRP_IPV4_IFADDR | RTMGRP_IPV6_IFADDR);
```
arvidn: why is this a global variable? if it truly needs to be global, could it be ``const``?

include/libtorrent/ip_notifier.hpp
ssiloti: Thanks for mentioning this. I think using the asio model for the interface makes things much cleaner.

include/libtorrent/ip_notifier.hpp
```diff
@@ -0,0 +1,79 @@
...
+			, std::size_t bytes_transferred
+			, std::function<void(error_code const&)> cb);
+
+#if defined TORRENT_BUILD_SIMULATOR
+		// TODO simulator support
+#elif TORRENT_USE_NETLINK
+		netlink::socket m_socket;
+		std::array<char, 4096> m_buf;
+#elif defined TORRENT_WINDOWS
+		OVERLAPPED m_ovl;
```
arvidn: this member doesn't seem to be initialized or cleared completely. afaict, only the event is set on it. Can you add `` = {}`` here to zero-initialize it?

src/ip_notifier.cpp
```diff
@@ -0,0 +1,117 @@
...
+#elif !TORRENT_USE_NETLINK
+		TORRENT_UNUSED(ios);
+#endif
+	}
+
+	ip_change_notifier::~ip_change_notifier()
+	{
+#if defined TORRENT_WINDOWS && !defined TORRENT_BUILD_SIMULATOR
+		cancel();
+		m_hnd.close();
```
arvidn: I would have though ``close`` would be redundant just before destructing the handle..

src/session_impl.cpp
```diff
@@ -1685,6 +1691,13 @@ namespace aux {
 		this->abort();
 	}
 
+	void session_impl::on_ip_change(error_code const& ec)
+	{
+		if (ec) return;
+		m_ip_notifier.async_wait(std::bind(&session_impl::on_ip_change, this, _1));
```
arvidn: would you mind using a lambda and call ``on_ip_change`` via ``session_impl::wrap()`` here, and in the other place you call ``async_wait``? It just wraps the callback to handle exceptions, preventing them from leaking into the ``io_service``.

src/ip_notifier.cpp
ssiloti: I thought the same thing, but it turns out `object_handle` doesn't close the handle in its destructor.

src/ip_notifier.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+namespace libtorrent
+{
+	ip_change_notifier::ip_change_notifier(io_service& ios)
+#if defined TORRENT_BUILD_SIMULATOR
+#elif TORRENT_USE_NETLINK
+		: m_socket(ios
+			, netlink::endpoint(netlink(NETLINK_ROUTE), RTMGRP_IPV4_IFADDR | RTMGRP_IPV6_IFADDR))
+#elif defined TORRENT_WINDOWS
+		: m_hnd(ios, WSACreateEvent())
```
ssiloti: If `WSACreateEvent` fails I think you would end up getting an error back from the `NotifyAddrChange` call due to passing in an invalid handle. I can check and throw in the ctor if m_hnd doesn't have a valid native handle.

src/ip_notifier.cpp
```diff
@@ -80,6 +82,10 @@ namespace libtorrent
 		{
 			m_hnd.async_wait([this,cb](error_code const& ec) { on_notify(ec, 0, cb); });
 		}
+		else
+		{
+			cb(error_code(err, system_category()));
```
arvidn: I think, technically, you need to ``post()`` the call to the callback here. To make sure it always run after the current handler (which may not expect to be recursively invoked)

src/session_impl.cpp
```diff
@@ -1694,7 +1695,8 @@ namespace aux {
 	void session_impl::on_ip_change(error_code const& ec)
 	{
 		if (ec) return;
-		m_ip_notifier.async_wait(std::bind(&session_impl::on_ip_change, this, _1));
+		m_ip_notifier.async_wait([this] (error_code const& ec)
+			{ this->wrap(&session_impl::on_ip_change, ec); });
 		reopen_listen_sockets();
```
arvidn: I suppose in this specific case, it's OK to invoke this handler "recursively", but in general I'm pretty sure the convention is to always post handlers.

src/ip_notifier.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+		// on linux we could parse the message to get information about the
+		// change but Windows requires the application to enumerate the
+		// interfaces after a notification so do that for Linux as well to
+		// minimize the difference between platforms
+
+		// Linux can generate ENOBUFS if the socket's buffers are full
+		// don't treat it as an error
+		if (ec.value() == boost::system::errc::no_buffer_space)
+			cb(error_code());
```
arvidn: the direct call here is fine, since this is already its own handler function

arvidn:
lgtm!

include/libtorrent/ip_notifier.hpp
```diff
@@ -0,0 +1,79 @@
...
+	private:
+		void on_notify(error_code const& error
+			, std::size_t bytes_transferred
+			, std::function<void(error_code const&)> cb);
+
+#if defined TORRENT_BUILD_SIMULATOR
+		// TODO simulator support
+#elif TORRENT_USE_NETLINK
+		netlink::socket m_socket;
+		std::array<char, 4096> m_buf;
```
pavel-pimenov: why 4k?

PR #1301, opened 2016-11-07
some refactor in ip change notifier source code

aldenml:


src/ip_notifier.cpp
```diff
@@ -53,17 +55,21 @@ namespace libtorrent
...
 #elif defined TORRENT_WINDOWS
 		if (!m_hnd.is_open())
+#ifndef BOOST_NO_EXCEPTIONS
 			throw system_error(WSAGetLastError(), system_category());
+#else
+			std::terminate();
+#endif // BOOST_NO_EXCEPTIONS
 		m_ovl.hEvent = m_hnd.native_handle();
-#elif !TORRENT_USE_NETLINK
+#else
```
ssiloti: This should not be changed. I don't want the TORRENT_UNUSED line on Linux, it is only for unsupported platforms.

src/ip_notifier.cpp
```diff
@@ -53,17 +55,21 @@ namespace libtorrent
...
 		TORRENT_UNUSED(ios);
 #endif
 	}
 
 	ip_change_notifier::~ip_change_notifier()
 	{
-#if defined TORRENT_WINDOWS && !defined TORRENT_BUILD_SIMULATOR
 		cancel();
+#if defined TORRENT_WINDOWS && !defined TORRENT_BUILD_SIMULATOR
```
ssiloti: This should not be changed. The cancel call is only needed on Windows.

src/ip_notifier.cpp
aldenml: ok

src/ip_notifier.cpp
aldenml: not convinced but can't discuss about it, I will revert the line

include/libtorrent/ip_notifier.hpp
```diff
@@ -7,14 +7,14 @@ Redistribution and use in source and binary forms, with or without
...
 
-* Redistributions of source code must retain the above copyright
-notice, this list of conditions and the following disclaimer.
-* Redistributions in binary form must reproduce the above copyright
-notice, this list of conditions and the following disclaimer in
-the documentation and/or other materials provided with the distribution.
-* Neither the name of the author nor the names of its
-contributors may be used to endorse or promote products derived
-from this software without specific prior written permission.
+    * Redistributions of source code must retain the above copyright
```
arvidn: I take it this is the formatting used in all other files, right?

include/libtorrent/netlink.hpp
```diff
@@ -98,7 +95,7 @@ namespace libtorrent
 
 		const data_type* data() const
 		{
-			return (struct sockaddr*)&sockaddr;
```
arvidn: I would have expected clang or gcc on travis to warn on this... hm

include/libtorrent/netlink.hpp
```diff
@@ -158,9 +155,7 @@ namespace libtorrent
 		using endpoint = basic_nl_endpoint<netlink>;
 		using socket = boost::asio::basic_raw_socket<netlink>;
 
-		netlink() : nl_family(0)
-		{
-		}
+		netlink() : netlink(0) {}
```
arvidn: there isn't a name for family 0?

src/ip_notifier.cpp
arvidn: maybe there should be a comment describing why it's only necessary on windows

include/libtorrent/ip_notifier.hpp
aldenml: yes, just formatting

include/libtorrent/netlink.hpp
aldenml: let me check

gubatron: `#define AF_UNSPEC   0` maybe?

PR #1318, opened 2016-11-13
Provide a pieces bitfield in a cache_flushed_alert.

AllSeeingEyeTolledEweSew:
This is continuing my earlier project.

For some background: I need to read completed pieces directly from disk (deluge can't really make use of read_piece_alert, since it only polls for new alerts every 1s, which murders performance). You can see the deluge plugin I've written for RPC access at: https://github.com/AllSeeingEyeTolledEweSew/Deluge-PieceIO-Plugin

The trouble is that there's no way to figure out which pieces have actually been flushed to disk. I can query the completed pieces then call flush_cache and hope that those were the pieces that got flushed, however I don't know if the resulting cache_flushed_alert was triggered by my flush_cache, or something else (e.g. another client). So I might think a piece has been flushed when it hasn't.

I propose that libtorrent include a list of pieces in cache_flushed_alert, indicating which pieces are known to have been flushed.

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -4974,14 +4974,20 @@ namespace libtorrent
 		if (m_storage.get())
 		{
 			inc_refcount("release_files");
+			boost::shared_ptr<bitfield> pieces(new bitfield);
+			fill_pieces(pieces.get(), true);
 			m_ses.disk_thread().async_stop_torrent(m_storage.get()
-				, boost::bind(&torrent::on_cache_flushed, shared_from_this(), _1));
+				, boost::bind(&torrent::on_cache_flushed, shared_from_this(), _1, pieces));
 		}
 		else
 		{
 			TORRENT_ASSERT(m_abort);
 			if (alerts().should_post<cache_flushed_alert>())
-				alerts().emplace_alert<cache_flushed_alert>(get_handle());
+			{
+				bitfield pieces;
+				pieces.resize(m_torrent_file->num_pieces(), false);
+				alerts().emplace_alert<cache_flushed_alert>(get_handle(), pieces);
+			}
 		}
 
 		m_storage.reset();
@@ -8553,8 +8559,10 @@ namespace libtorrent
 		{
 			// we need to keep the object alive during this operation
 			inc_refcount("release_files");
+			boost::shared_ptr<bitfield> pieces(new bitfield);
+			fill_pieces(pieces.get(), true);
 			m_ses.disk_thread().async_release_files(m_storage.get()
-				, boost::bind(&torrent::on_cache_flushed, shared_from_this(), _1));
+				, boost::bind(&torrent::on_cache_flushed, shared_from_this(), _1, pieces));
 		}
 
 		// this torrent just completed downloads, which means it will fall
@@ -9713,12 +9721,14 @@ namespace libtorrent
 			TORRENT_ASSERT(m_abort);
 			return;
 		}
+		boost::shared_ptr<bitfield> pieces(new bitfield);
+		fill_pieces(pieces.get(), true);
 		inc_refcount("release_files");
 		m_ses.disk_thread().async_release_files(m_storage.get()
-			, boost::bind(&torrent::on_cache_flushed, shared_from_this(), _1));
+			, boost::bind(&torrent::on_cache_flushed, shared_from_this(), _1, pieces));
 	}
 
-	void torrent::on_cache_flushed(disk_io_job const*)
+	void torrent::on_cache_flushed(disk_io_job const*, boost::shared_ptr<bitfield> pieces)
 	{
 		dec_refcount("release_files");
 		TORRENT_ASSERT(is_single_thread());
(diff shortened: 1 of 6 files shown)


arvidn:
I have some concerns with this patch:

1. It makes this alert much more expensive to post since it causes a (potentially large) heap allocation. alerts and their state is mostly allocated on a double-buffered stack right now, which makes them cheap. The bitfield could be allocated on the same stack too, then the remaining cost would be filling it in (which doesn't scale all that well with number of pieces)

2. The semantics of "piece flushed to disk" is probably ambiguously, or poorly defined. Depending on your use case, these pieces may not have been flushed enough. I don't believe there's an actual call to ``fdatasync()`` or ``FlushFileBuffers()`` for instance.

3. The use case seem somewhat questionable. This is exactly what ``read_piece`` is meant for, which also is much more reliable when renaming files, moving files or using partfiles. It seems like a much better approach to fix Deluge's alert polling performance to allow use of ``read_piece`` instead. There is a callback that can be called whenever there's a new alert, to notify the client it can ``pop_alerts`` in its main thread.

AllSeeingEyeTolledEweSew:
1. I did think about this. My rationale is that `cache_flushed_alert` shouldn't be fired very frequently, in any use case I can think of. I'll see if I can figure out how to allocate the bitfield on the stack as you suggest. Could I populate the bitfield conditionally? I could change the method signature to `flush_cache(int flags = 0)`. Would this break anything?
2. Generally POSIX requires that after a `write()` on a file has completed, `read()` should return the data written. That requirement is good enough for my use case. If a caller wanted something "stronger" like `fdatasync()` they could always do it themselves on `cache_flushed_alert`. In my reading, `fdatasync()` isn't meaningful on all environments, so it doesn't seem necessary to ensure it always happens.
3. I did look into changing deluge to use the alert callback. Sadly it wasn't as straightforward as I hoped. I ended up rewriting a few different fundamental codepaths before giving up. This was a few weeks ago, so I don't remember the details clearly.

I should explain that I'm writing a filesystem backed by bittorrent (working title "yet another torrent filesystem"), so I prefer to service read requests with minimum latency. Even if we fix deluge, it will still be a single-threaded TwistedMatrix event loop, so a `read_piece_alert` might take a while to get serviced. Note that I still don't necessarily need to call `flush_cache()` very often: I can cache the alert's bitfield, read the file directly for any pieces I know are there, and only `flush_cache()` when I want to service a read for a piece I know is downloaded but not flushed.

My hope was that this change would make `cache_flushed_alert` more meaningful generally. Can a client infer anything meaningful from it today? All it indicates is that some data was written to the backing file.

arvidn:
1. the only complication it would add to allocate the bitfield efficiently is that you wouldn't be able to reuse the bitfield class. However, this is not a strong argument, it's just a quality of implementation issue.
2. my point is that the semantics you selected are specific to your use case, and for anyone else they are arbitrary.
3. I imagine Deluge has a message queue as its main loop, where you can post messages to be executed. It sounds like right now there's a one-second timer event that calls ``pop_alerts()``. Instead of having this message be posted on a timer, it could be posted by libtorrent signalling that there are new alerts. This change would benefit Deluge in a lot of other way too presumably, being able to react quicker to events. It doesn't seem like a long-term approach to expose more internals of libtorrent instead of fixing Deluge's alert dispatch code.

My long term hope is to remove the ``cache_flushed_alert`` along with the cache itself (and use memory mapped files). You may still have similar synchronization issues though, in case the filesystem backing the files is NFS for instance. but that's what ``read_piece`` is for anyway.

If you're writing a filesystem, I would expect you would store torrent data as single blob files too, otherwise you'll be paying the cost of unaligned disk access.

AllSeeingEyeTolledEweSew:
I like the `mmap` solution (with `MAP_SHARED`, on Linux). Are you targeting that for the 1.1.x train? Could I help with it?

I still have doubts about the best-case throughput I can get with `read_piece` through deluge's event loop. If the `mmap` solution is in the future, I think I'd prefer to make no libtorrent code changes for now. I'll call `status(query_pieces)`, call `flush_cache()`, wait for `cache_flushed_alert`, declare undefined behavior if anyone else calls `flush_cache()`.

Alignment is a good thought. For my case, I'll service `read()` requests aligned to a file, not to the torrent blob, so `read_piece` results would almost always have bad alignment. I would have to implement some cache layer.

src/torrent.cpp
```diff
@@ -12205,6 +12201,25 @@ namespace libtorrent
...
 	}
 
+	void torrent::fill_pieces(bitfield* pieces, bool complete)
+	{
+		int num_pieces = m_torrent_file->num_pieces();
+		if (has_picker() && complete)
+		{
+			pieces->resize(num_pieces, false);
+			for (int i = 0; i < num_pieces; ++i)
+				if (m_picker->has_piece_passed(i)) pieces->set_bit(i);
```
arvidn: a piece having passed the hash check does not guarantee it has been flushed to disk. Even when we get a ``cache_flushed_alert``, we may have downloaded a new piece after the cache was flushed.

When a piece passes the hash check, ``async_flush_piece()`` is called to queue up the disk job to write it to disk. There's no explicit synchronization with this disk job completing and any piece picker or torrent state though.

Taking a step back. It seems like what you want to do is to query the disk cache of which pieces are dirty (i.e. not flushed to disk yet). You can do that with ``session_handle::get_cache_status()``. That's a synchronous call.

AllSeeingEyeTolledEweSew:
Ah, that does seem like what I needed. Thanks for the help. I'll abandon this PR.

Issue #291, opened 2015-11-23
v1.0.6 not disconnecting from webseeds


chrishirst:
libtorrent ver 1.0.6
compiled in with qbittorrent 3.3.0
Linux Mint 17.2 

Web seeds remain connected after completion and torrent status shows "stalled"

![completed-task-showing-stalled-in-downloading-filter](https://cloud.githubusercontent.com/assets/3680285/11336601/8e27b876-91de-11e5-9164-06767436f2ca.png)

sledgehammer999:
FYI "Stalled" status means these tests are true:
`torrent_status::downloading && m_nativeStatus.download_payload_rate == 0`

chrishirst:
Sure, but that is not the point I am making. The task IS complete, yet it is still in the "downloading" category and still connected to seeds.

sledgehammer999:
>Sure, but that is not the point I am making. The task IS complete, yet it is still in the "downloading" category and still connected to seeds.

I am just explaining to arvidn what "Stalled" means. That is a qbittorrent term, not a libtorrent one.

arvidn:
how certain are we that the torrent is completely downloaded?
I would imagine that it isn't completed given that state is still downloading.

sledgehammer999:
@chrishirst click on the "General" button it has info on how many pieces the torrent has in total how many we already have. If they match then it is completely downloaded.

chrishirst:
Did some more testing yesterday.

The combination that triggers the behaviour is:

A swarm that is exclusively comprised of "web seeds"
plus
skipped 'files' in the payload.

If there is one 'normal' seed in the swarm,  my case it was ktorrent on a LAN host machine, 'skipping' one or more files does not trigger the behaviour.

@sledgehammer999 
Without 'web seeds' being involved, "completed" == "**All required** pieces".

sledgehammer999:
@arvidn here is how we calculate the progress value:
```c++
qreal TorrentHandle::progress() const
{
    if (!m_nativeStatus.total_wanted)
        return 0.;

    if (m_nativeStatus.total_wanted_done == m_nativeStatus.total_wanted)
        return 1.;

    float progress = (float) m_nativeStatus.total_wanted_done / (float) m_nativeStatus.total_wanted;
    Q_ASSERT((progress >= 0.f) && (progress <= 1.f));
    return progress;
}
```

And then we transform the float value like this:
```c++
qreal progress = index.data().toDouble()*100.;
[...]

PR #1321, opened 2016-11-14
fix bug: urlseed never timeout

tnextday:
`m_last_sent` has updated in `peer_connection::on_send_data` function,  update in `peer_connection::keep_alive()` will case `web_connection_base` never timeout.

### DIFF
--- a/include/libtorrent/web_connection_base.hpp
+++ b/include/libtorrent/web_connection_base.hpp
@@ -87,7 +87,7 @@ namespace libtorrent
 		void write_dont_have(int) {}
 		void write_piece(peer_request const&, disk_buffer_holder)
 		{ TORRENT_ASSERT_FAIL(); }
-		void write_keepalive() {}
+		bool write_keepalive() { return false; }
 		void on_connected();
 		void write_reject_request(peer_request const&) {}
 		void write_allow_fast(int) {}
--- a/simulation/test_web_seed.cpp
+++ b/simulation/test_web_seed.cpp
@@ -449,3 +449,39 @@ TORRENT_TEST(multi_file_unaligned_redirect)
 		}
 	);
 }
+
+
+
+TORRENT_TEST(urlseed_timeout)
+{
+	using namespace libtorrent;
+
+	file_storage fs;
+	fs.add_file("timeout_test_file", 0x8000);
+	lt::add_torrent_params params = ::create_torrent(fs);
+	params.url_seeds.push_back("http://2.2.2.2:8080/");
+	bool timeout = false;
+	run_test(
+		[&params](lt::session& ses)
+		{
+			ses.async_add_torrent(params);
+		},
+		[&](lt::session& ses, lt::alert const* alert) {
+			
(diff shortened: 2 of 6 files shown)


arvidn:
This affects bittorrent peers too. it's not obvious that this is correct. I would expect something like this to have to be done in bt_peer_connection::write_keepalive() or something like that, no?

tnextday:
Ok, i updated the code, maybe  https://github.com/arvidn/libtorrent/issues/291 is the same problem.

arvidn:
this is looking better. This probably would work. However, a part of me wonders if this couldn't be done a lot simpler by just moving the update if ``m_last_sent`` to right next to ``async_write_some`` in ``peer_connection::setup_send()``. That seems like it would be a lot more robust.

It would also be nice to have a test for this. Since it's requires waiting, it should be a simulation (perhaps a test case in test_web_seed.cpp under simulation). Basically to verify that a web seed _does_ time out.

tnextday:
I have add urlseed_timeout simulator, 
It depends on https://github.com/arvidn/libsimulator/pull/17

simulation/test_web_seed.cpp
```diff
@@ -449,3 +449,39 @@ TORRENT_TEST(multi_file_unaligned_redirect)
...
+				timeout = true;
+			}
+		},
+		[&fs](sim::simulation& sim, lt::session& ses)
+		{
+			sim::asio::io_service web_server(sim, address_v4::from_string("2.2.2.2"));
+			
+			// listen on port 8080
+			sim::http_server http(web_server, 8080);
+			http.register_no_response_handler("/timeout_test_file");
```
arvidn: I see, you deliberately want it to stall

include/libtorrent/web_connection_base.hpp
```diff
@@ -87,7 +87,7 @@ namespace libtorrent
 		void write_dont_have(int) {}
 		void write_piece(peer_request const&, disk_buffer_holder)
 		{ TORRENT_ASSERT_FAIL(); }
-		void write_keepalive() {}
+		bool write_keepalive() { return false; }
```
arvidn: I don't think you meant to mix this PR in here, right?

arvidn: oh, sorry, this _is_ the patch you just added the test for.. never mind..

src/peer_connection.cpp
```diff
@@ -6518,8 +6518,9 @@ namespace libtorrent
 		peer_log(peer_log_alert::outgoing_message, "KEEPALIVE");
 #endif
 
-		m_last_sent = aux::time_now();
-		write_keepalive();
+		if (write_keepalive()) {
+			m_last_sent = aux::time_now();
```
arvidn: yeah, I think either way is fine. If the connection truly is stalled, setup_send() won't be called either, so in practice I don't think it would make a significant difference

tnextday:
ok, I will open another PR later.

arvidn:
The test in here is still useful

arvidn:
in fact, I bet it would work fine in RC_1_1 even

PR #1322, opened 2016-11-14
fixed gcc 6 va_list argument passing issue

aldenml:
Experimenting with GCC 6 and getting errors like this:
```
/home/travis/build/aldenml/frostwire-jlibtorrent/libtorrent/src/session_impl.cpp: In member function ‘virtual void libtorrent::aux::session_impl::session_log(const char*, ...) const’:
/home/travis/build/aldenml/frostwire-jlibtorrent/libtorrent/src/session_impl.cpp:4377:43: error: no matching function for call to ‘libtorrent::alert_manager::emplace_alert(const char*&, va_list)’
   m_alerts.emplace_alert<log_alert>(fmt, v);
                                           ^
In file included from /home/travis/build/aldenml/frostwire-jlibtorrent/libtorrent/include/libtorrent/aux_/session_impl.hpp:67:0,
                 from /home/travis/build/aldenml/frostwire-jlibtorrent/libtorrent/src/session_impl.cpp:77:
/home/travis/build/aldenml/frostwire-jlibtorrent/libtorrent/include/libtorrent/alert_manager.hpp:62:8: note: candidate: void libtorrent::alert_manager::emplace_alert(Args&& ...) [with T = libtorrent::log_alert; Args = {const char*&, __va_list_tag (&)[1]}]
   void emplace_alert(Args&&... args)
        ^~~~~~~~~~~~~
/home/travis/build/aldenml/frostwire-jlibtorrent/libtorrent/include/libtorrent/alert_manager.hpp:62:8: note:   no known conversion for argument 2 from ‘va_list {aka __va_list_tag [1]}’ to ‘__va_list_tag (&)[1]’
/home/travis/build/aldenml/frostwire-jlibtorrent/libtorrent/src/session_impl.cpp: In member function ‘virtual void libtorrent::aux::session_impl::log(libtorrent::dht::dht_logger::module_t, const char*, ...)’:
/home/travis/build/aldenml/frostwire-jlibtorrent/libtorrent/src/session_impl.cpp:6489:55: error: no matching function for call to ‘libtorrent::alert_manager::emplace_alert(libtorrent::dht_log_alert::dht_module_t, const char*&, va_list)’
    static_cast<dht_log_alert::dht_module_t>(m), fmt, v);
```

### DIFF
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -2101,8 +2101,7 @@ namespace libtorrent
 	struct TORRENT_EXPORT log_alert final : alert
 	{
 		// internal
-		log_alert(aux::stack_allocator& alloc, char const* log);
-		log_alert(aux::stack_allocator& alloc, char const* fmt, va_list v);
+		log_alert(aux::stack_allocator& alloc, string_view log);
 
 		TORRENT_DEFINE_ALERT(log_alert, 79)
 
@@ -2131,7 +2130,7 @@ namespace libtorrent
 	{
 		// internal
 		torrent_log_alert(aux::stack_allocator& alloc, torrent_handle const& h
-			, char const* fmt, va_list v);
+			, string_view log);
 
 		TORRENT_DEFINE_ALERT(torrent_log_alert, 80)
 
@@ -2172,7 +2171,7 @@ namespace libtorrent
 		peer_log_alert(aux::stack_allocator& alloc, torrent_handle const& h
 			, tcp::endpoint const& i, peer_id const& pi
 			, peer_log_alert::direction_t dir
-			, char const* event, char const* fmt, va_list v);
+			, char const* event, string_view log);
 
 		TORRENT_DEFINE_ALERT(peer_log_alert, 81)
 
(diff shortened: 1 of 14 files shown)


arvidn:
this adds a string copy for every log line. do you understand why it's not allowed to pass a ``va_list`` through a variadic template argument?

aldenml:
I see the drawback of this solution. I think the key is here:
```
no known conversion for argument 2 from ‘va_list {aka __va_list_tag [1]}’ to ‘__va_list_tag (&)[1]’
```
but I don't understand it more than just repeating the message.

As an additional note, I used this "passing va_list as a parameters" in my posix layer and it was a root of crashes in my JNI layer in android.

aldenml:
you know, i reverted back to gcc 5, and this is just a distraction. I'm pretty sure this will be easily resolved when the real time for gcc 6 arrives.

arvidn:
I get the impression that ``va_list`` cannot be passed by reference, and since it's an lvalue the forwarding reference will turn into be taken by non-const reference, and break. Worst case the ``va_list`` object could just be stuck inside a proper object and be passed that way.

ssiloti:
std::ref might do the trick to shepherd the va_list through the template functions.

PR #1323, opened 2016-11-15
Bind cache_status::pieces.

AllSeeingEyeTolledEweSew:
Per your suggestion to use `session_handle::get_cache_info`, I'd like to get python bindings for the pieces field.

### DIFF
--- a/bindings/python/src/session.cpp
+++ b/bindings/python/src/session.cpp
@@ -16,6 +16,7 @@
 #include <libtorrent/aux_/session_impl.hpp> // for settings_map()
 #include <libtorrent/torrent_info.hpp>
 #include <libtorrent/kademlia/item.hpp> // for sign_mutable_item
+#include <libtorrent/time.hpp>
 
 #ifndef TORRENT_NO_DEPRECATE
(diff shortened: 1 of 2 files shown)


arvidn:
looks good. would you mind adding a call to ``test.py`` to get one of these and print it out too? just to get a bit more coverage.

arvidn:
did you see this? https://ci.appveyor.com/project/arvidn/libtorrent/build/1.0.4410/job/bvl7q40tgb4s0rdj#L862

bindings/python/src/session.cpp
```diff
@@ -465,6 +466,28 @@ namespace
        return ret;
     }
 
+    list cached_piece_info_list(std::vector<cached_piece_info> const& v)
+    {
+       list pieces;
+       ptime now = time_now();
```
arvidn: I believe this is the problem. the ``ptime`` typedef is deprecated, you should use ``lt::time_point`` instead.

AllSeeingEyeTolledEweSew:
Sorry for the delay, I was out of town for a few days. Updated the PR per your advice. I think the CI failures are unrelated to my change.

PR #1324, opened 2016-11-15
fixed warning non-virtual-dtor and minor code refactor

aldenml:


arvidn:
I'm not sure about this. It feels like going a bit too far doing arbitrary things to satisfy a compiler warning. A warning like this should be issued at a problematic delete-site, not across the board.

aldenml:
I always thought that having a virtual destructor in a virtual base type is a good practice (even if nothing happens at the end). It's that way in all the rest of the code.

As for the `TORRENT_EXTRA_EXPORT` in `error_handler_interface`, once I added the virtual destructor, msvc complained of not having it. I checked all the interfaces of `session_impl` and `peer_connection` and all of them have the `TORRENT_EXTRA_EXPORT` and the virtual destructor.

I think that just for consistency, this should be added, and in the process the warning is gone.

(maybe one day all the `-Wno-*` could be removed from the `Jamfile`)

arvidn:
I see, the fact that you added the dllexport triggered the warning, that kind of makes sense, since the compiler can't know of all the places where this is used now.
Why did it need to be exported?

Jamfile
```diff
@@ -247,10 +247,6 @@ rule warnings ( properties * )
 		result += <cflags>-Wno-c++11-long-long ;
 		result += <cflags>-Wno-variadic-macros ;
 
-# in C++98 mode there's no way to silence this warning
-# in the code (without final)
-		result += <cflags>-Wno-non-virtual-dtor ;
```
arvidn: I would expect any place where this warning would be triggered can be marked as final now, since we have C++11 in master. so, this seems good

include/libtorrent/storage.hpp
```diff
@@ -565,6 +565,8 @@ namespace libtorrent
 	{
 		virtual int file_op(int const file_index, std::int64_t const file_offset
 			, span<file::iovec_t const> bufs, storage_error& ec) = 0;
+
+		virtual ~fileop() {}
```
arvidn: the problem I have with these is that these types are never deleted "polymorphically", i.e. all concrete types are instantiated on the stack. just having this destructor here, I think, is confusing. It implies that it's used or necessary.

aldenml:
I think I understand better now. I changed to a local warning suppression, another solution could be adding a protected (non-virtual) destructor.

(this is low low low priority btw)

src/storage.cpp
```diff
@@ -212,7 +212,7 @@ namespace libtorrent
 
 	} // anonymous namespace
 
-	struct write_fileop : fileop
+	struct write_fileop final : fileop
```
arvidn: this didn't help?

arvidn:
I think a protected (non-virtual) destructor would be nicer. I see I did that in a few other places

src/storage.cpp
aldenml: not for `fileop` and `error_handler_interface`

aldenml:
ok, I will do that

PR #1325, opened 2016-11-15
auto loop, consts and minor cleanup refactor

aldenml:


src/torrent.cpp
```diff
@@ -780,9 +779,8 @@ namespace libtorrent
 		if (!settings().get_bool(settings_pack::use_dht_as_fallback)) return true;
 
 		int verified_trackers = 0;
-		for (std::vector<announce_entry>::const_iterator i = m_trackers.begin()
-			, end(m_trackers.end()); i != end; ++i)
-			if (i->verified) ++verified_trackers;
+		for (auto const& tr : m_trackers)
```
arvidn: this could even be a ``std::count_if``

src/torrent.cpp
```diff
@@ -1844,20 +1842,18 @@ namespace libtorrent
 
 			std::vector<int> have_pieces;
 
-			for (std::vector<piece_picker::downloading_piece>::const_iterator i
-				= dq.begin(); i != dq.end(); ++i)
+			for (auto const& p : dq)
```
arvidn: this could be a ``std::copy_if``

arvidn: actually, no, it couldn't. never mind

PR #1327, opened 2016-11-15
fix windows-version=xp build

ajax16384:
window xp build may fail due CALG_SHA_512 definition absence. let's fallback to built-in hash routine.

### DIFF
--- a/include/libtorrent/hasher512.hpp
+++ b/include/libtorrent/hasher512.hpp
@@ -46,16 +46,9 @@ POSSIBILITY OF SUCH DAMAGE.
 #elif TORRENT_USE_COMMONCRYPTO
 #include <CommonCrypto/CommonDigest.h>
 
-#elif TORRENT_USE_CRYPTOAPI
+#elif TORRENT_USE_CRYPTOAPI_SHA_512
 #include <windows.h>
-#include <wincrypt.h> // CALG_SHA_512 conditionally defined here
-
-#ifdef CALG_SHA_512
-#define TORRENT_USE_CRYPTOAPI_SHA_512 1
-#else
-// fallback to built-in
-#include "libtorrent/sha512.hpp"
-#endif
+#include <wincrypt.h>
 
 #elif defined TORRENT_USE_LIBCRYPTO
 
@@ -65,10 +58,6 @@ extern "C" {
 
 #else
 #include "libtorrent/sha512.hpp"
-#endif
-
-#ifndef TORRENT_USE_CRYPTOAPI_SHA_512
-#define TORRENT_USE_CRYPTOAPI_SHA_512 0
 #endif
 
 #include "libtorrent/aux_/disable_warnings_pop.hpp"
--- a/include/libtorrent/hasher512.hpp
+++ b/include/libtorrent/hasher512.hpp
@@ -48,10 +48,12 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #elif TORRENT_USE_CRYPTOAPI
 #include <windows.h>
-#include <wincrypt.h>
+#include <wincrypt.h> // CALG_SHA_512 conditionally defined here
 
+#ifdef CALG_SHA_512
+#define TORRENT_USE_CRYPTOAPI_SHA_512 1
+#else
 // fallback to built-in
-#ifndef CALG_SHA_512
 #include "libtorrent/sha512.hpp"
 #endif
 
@@ -117,7 +119,7 @@ namespace libtorrent
 		gcry_md_hd_t m_context;
 #elif TORRENT_USE_COMMONCRYPTO
 		CC_SHA512_CTX m_context;
-#elif TORRENT_USE_CRYPTOAPI && defined(CALG_SHA_512)
+#elif TORRENT_USE_CRYPTOAPI_SHA_512
 		HCRYPTHASH m_context;
 #elif defined TORRENT_USE_LIBCRYPTO
 		SHA512_CTX m_context;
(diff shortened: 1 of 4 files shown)


aldenml:
according to this https://msdn.microsoft.com/en-us/library/windows/desktop/aa375549(v=vs.85).aspx it should be available in XP SP3

ajax16384:
yes it's defined for headers targeted to NTDDI_WINXPSP3 or higher.
````c++
#if (NTDDI_VERSION > NTDDI_WINXPSP2)
#define CALG_SHA_512            (ALG_CLASS_HASH | ALG_TYPE_ANY | ALG_SID_SHA_512)
````
but where is no reason to drop xp pre sp3 support

**Comment by aldenml** [comment]
question: do you see a benefit in supporting such an old OS target?

**Comment by ajax16384** [comment]
windows xp sp2 still can be found in wild, so if library can be built with xp support (which is not default option) then why not give opportunity to support outdated configuration.
(default windows-version=vista also not best windows version choice for personal use :) )

**Comment by ssiloti** [comment]
No one should be running Windows XP on an internet connected machine these days. They **definitely** should not be running pre-SP3. We should not enable people to continue to use insecure operating systems.

File: include/libtorrent/hasher512.hpp [hunk 1]
```diff
@@ -112,7 +117,7 @@ namespace libtorrent
 		gcry_md_hd_t m_context;
 #elif TORRENT_USE_COMMONCRYPTO
 		CC_SHA512_CTX m_context;
-#elif TORRENT_USE_CRYPTOAPI
+#elif TORRENT_USE_CRYPTOAPI && defined(CALG_SHA_512)
```
arvidn: I would imagine a better place for this would be in ``config.h``. Or is ``TORRENT_USE_CRYPTOAPI`` used in other places too, that do work under XP?

arvidn: i.e. make config.hpp not set ``TORRENT_USE_CRYPTOAPI`` when building for a version that doesn't support it

ajax16384: XP pre SP3 does not support only SHA512, so other cryptoapi functions should work well. To detect CALG_SHA_512 absence TORRENT_USE_CRYPTOAPI must be defined and wincrypt.h must be included, so config.hpp is not suitable place for this detection.

arvidn:
I'm a little bit hesitant about the additional preprocessor condition. Can you think of a cleaner way to do it? perhaps the ``TORRENT_USE_CRYPTOAPI`` is too broad, maybe there should be two config macros.

include/libtorrent/hasher512.hpp
```diff
@@ -48,7 +48,14 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #elif TORRENT_USE_CRYPTOAPI
 #include <windows.h>
-#include <wincrypt.h>
+#include <wincrypt.h> // CALG_SHA_512 conditionally defined here
+
+#ifdef CALG_SHA_512
+#define TORRENT_USE_CRYPTOAPI_SHA_512 1
```
arvidn: it feels like this really belongs in ``config.hpp``, where all the other configuration options are defined. I can't think of a strong argument though, other than symmetry and ease to find where they're coming from. I suppose one reason is that if any other piece of code in libtorrent would check this macro, it may not be obvious that this header needs to be included

ajax16384: it is correct to insert windows includes in config.hpp like this?
config.hpp:
````c++
# if !defined TORRENT_USE_LIBCRYPTO && !defined TORRENT_USE_LIBGCRYPT
// unless some other crypto library has been specified, default to the native
// windows CryptoAPI
#define TORRENT_USE_CRYPTOAPI 1
#endif
.....
#ifndef TORRENT_USE_CRYPTOAPI
#define TORRENT_USE_CRYPTOAPI 0
#endif

#if TORRENT_USE_CRYPTOAPI
#include <windows.h>
#include <wincrypt.h> // CALG_SHA_512 conditionally defined here

#ifdef CALG_SHA_512
#define TORRENT_USE_CRYPTOAPI_SHA_512 1
#endif

#ifndef TORRENT_USE_CRYPTOAPI_SHA_512
#define TORRENT_USE_CRYPTOAPI_SHA_512 0
#endif

#ifndef TORRENT_HAVE_MMAP
#define TORRENT_HAVE_MMAP 0
#endif

#ifndef TORRENT_COMPLETE_TYPES_REQUIRED
#define TORRENT_COMPLETE_TYPES_REQUIRED 0
#endif
.....
````

arvidn: config.hpp used to include ``windows.h``, ideally it wouldn't though, to cut down on include cost/compile time. Couldn't you make this configuration depend on the windows version the build is targeting? i.e. ``_WIN32_WINNT``.

ajax16384: ok

aldenml:
Sorry for a kind of hard line position, but supporting a target less than XP SP3 seems a little odd at the cost of more ugliness. You can always compile `openssl` and use it right?

ajax16384:
i think built-in hash implementaion is more failsafe than openssl linking

arvidn:
@ajax16384 do you have a compelling argument to support XP pre sp3?

ajax16384:
@arvidn i have no strong argument to support XP pre sp3. Personaly i wish all users should migrate at least to window 7. But if it's not demand great efforts to support outdated configuration then why not to fix it?
If you insists to drop old configurations then let's leave current source state untouched and cancel this commit.

arvidn:
I think this change is reasonably small

PR #1331, opened 2016-11-16
update stack_allocator to support 0 size allocations

mnaamani:
This fix handles the case when the session emplaces the `dht_get_peers_reply_alert` with zero peers.

The issue is that the `dht_get_peers_reply_alert` constructor doesn't handle the case when zero peers found, resulting in the allocation of 0 bytes in the stack_allocator followed by attempting to ~~access memory~~ get a pointer to location beyond the allocated storage. A `TORRENT_ASSERT` statement in `stack_allocator::ptr(int)` catches this condition and produces:

```
version: 1.1.0.0-044ee0f

file: 'include/libtorrent/stack_allocator.hpp'
line: 80
function: char *libtorrent::aux::stack_allocator::ptr(int)
expression: idx < int(m_storage.size())

stack:
1: assert_fail(char const*, int, char const*, char const*, char const*, int)
2: libtorrent::aux::stack_allocator::ptr(int)
3: libtorrent::dht_get_peers_reply_alert::dht_get_peers_reply_alert(libtorrent::aux::stack_allocator&, libtorrent::sha1_hash const&, std::__1::vector<boost::asio::ip::basic_endpoint<boost::asio::ip::tcp>, std::__1::allocator<boost::asio::ip::basic_endpoint<boost::asio::ip::tcp> > > const&)
4: libtorrent::dht_get_peers_reply_alert::dht_get_peers_reply_alert(libtorrent::aux::stack_allocator&, libtorrent::sha1_hash const&, std::__1::vector<boost::asio::ip::basic_endpoint<boost::asio::ip::tcp>, std::__1::allocator<boost::asio::ip::basic_endpoint<boost::asio::ip::tcp> > > const&)
5: void libtorrent::alert_manager::emplace_alert<libtorrent::dht_get_peers_reply_alert, libtorrent::sha1_hash&, std::__1::vector<boost::asio::ip::basic_endpoint<boost::asio::ip::tcp>, std::__1::allocator<boost::asio::ip::basic_endpoint<boost::asio::ip::tcp> > > const&>(libtorrent::sha1_hash&&&, std::__1::vector<boost::asio::ip::basic_endpoint<boost::asio::ip::tcp>, std::__1::allocator<boost::asio::ip::basic_endpoint<boost::asio::ip::tcp> > > const&&&)
.
.
```

### DIFF
--- a/include/libtorrent/stack_allocator.hpp
+++ b/include/libtorrent/stack_allocator.hpp
@@ -90,8 +90,9 @@ namespace libtorrent { namespace aux
 
 		int copy_buffer(span<char const> buf)
 		{
-			int const ret = int(m_storage.size());
 			int const size = int(buf.size());
+			if (size == 0) return -1;
+			int const ret = int(m_storage.size());
 			m_storage.resize(ret + size);
 			std::memcpy(&m_storage[ret], buf.data(), size);
 			return ret;
@@ -99,7 +100,7 @@ namespace libtorrent { namespace aux
 
 		int allocate(int const bytes)
 		{
-			TORRENT_ASSERT(bytes >= 0);
+			if (bytes < 1) return -1;
 			int const ret = int(m_storage.size());
 			m_storage.resize(ret + bytes);
 			return ret;
@@ -107,15 +108,15 @@ namespace libtorrent { namespace aux
 
 		char* ptr(int const idx)
 		{
-			TORRENT_ASSERT(idx >= 0);
-			TORRENT_ASSERT(idx < int(m_storage.size()));
+			if(idx < 0) return nullptr;
+			if(idx >= int(m_storage.size())) return nullptr;
 			return &m_storage[idx];
 		}
 
 		char const* ptr(int const idx) const
 		{
-			TORRENT_ASSERT(idx >= 0);
-			TORRENT_ASSERT(idx < int(m_storage.size()));
+			if(idx < 0) return nullptr;
+			if(idx >= int(m_storage.size())) return nullptr;
 			return &m_storage[idx];
 		}
 


mnaamani:
I don't quite understand what the failed check means:

` codecov/patch — 0.00% of diff hit (target 71.15%)`

Is it because of the two blank lines?

arvidn:
ideally there would be a test for this. I don't believe there are any existing tests really, but if you're up for it, a test would be great (I imagine you could just create a stack_allocator and instantiate the alert object directly, without an alert_manager)

arvidn:
change looks good though. I wonder if there are other alerts with similar problems. Also, I think this might affect RC_1_1 as well. If so, you might want to make this PR against that branch.

mnaamani:
Upon inspection the following alerts 

- picker_log_alert - when number of blocks is zero
- dht_pkt_alert - if buffer size is zero
- dht_direct_response_alert in dht_direct_response_alert::message(), if response data size is zero

can potentially have a similar bug because they use  `stack_allocator::copy_buffer(char*buffer, int size)` 

It might be a good idea for `stack_allocator::allocate(int bytes)` and `stack_allocator::copy_buffer(char*, int size)` to return `-1` when `bytes` argument, or `size` argument are 0. We can then check the return value before using it to get a pointer with `stack_allocator::ptr(int)` or with the `[]` operator.

aldenml:
may i suggest a different code style? (I personally don't like very much a `return` in a constructor and this is just an opinion)
something like:
```cpp
char *ptr = m_num_peers > 0 ? alloc.ptr(m_peers_idx) : nullptr;
```

arvidn:
maybe it would make more sense, as you point out, to make the stack_allocator support zero-sized allocations.

mnaamani:
Thanks for the feedback @arvidn and @aldenml 

Yes this affects RC_1_1 as well, so I will make an attempt to fix the the other alerts too, and update the stack_allocator with the suggested fix in a new PR

mnaamani:
I've pushed fixes for the remaining alerts.
`dht_direct_response_alert` didn't require any changes, there are already checks for `m_response_size` being zero.

Are there any other parts of library that utilize the `stack_allocator` or is it just the alerts?

aldenml:
I'm starting to feel that the idea of "support 0 size allocations" is not properly captured with a return of `-1` API wise. Maybe returning the size (as before) and in the method `ptr` if the index is just the size, then returns `nullptr`. What do you think?

mnaamani:
I agree, we are in fact indicating by returning `-1` that the API does NOT support 0 size allocations.

However the return value as before is the size when allocation of 0, or an index into the stack allocator buffer when allocation more than 0, giving two different meanings to the return value.

This is why I prefer returning -1.

Returning `nullptr` in `ptr` is a good idea, if the index >= size or index < 0. I think that would be better than what we have now, there are TORRENT_ASSERT statements in there and in the underlying buffer [] operator that guard against passing invalid index. In a build with asserts off we could get undefined behaviour?

Ideally we would return a `char*` from all the allocation methods, no longer requiring the `ptr` method (similar to C memory management method `alloc`, `malloc`, `calloc`) but that will break the API and require significant changes in all the alerts.

arvidn:
The reason the alerts refer to their allocation by index is:

1. an index is smaller than a pointer
2. the buffer backing the stack allocator may need to be reallocated, which would invalidate pointers

arvidn:
To be precise, the assert that fires currently, inside stack_allocator, indicates that this is fundamentally not supposed to happen, undefined and outside of the scope of the API. by returning -1, and doing sensible things when asking for the -1 pointer, I would argue the API does support zero-sized allocations. It does something sensible.

And I think that's a perfectly reasonable approach, more robust than updating all the alerts.

mnaamani:
That makes sense, especially the point about the pointers becoming invalidated after a resize of the buffer.

So going forward do you recommend only updating the stack_allocator as follows:

- copy_buffer and allocate: return -1 for zero size allocations
- ptr : remove the asserts and return `nullptr` (`NULL` in RC_1_1 branch) for out of range index

You stated "And I think that's a perfectly reasonable approach, more robust than updating all the alerts." Are you suggesting to revert the changes to the alerts?

arvidn:
> Are you suggesting to revert the changes to the alerts?

yes. In fact, it's sufficient to make this change in RC_1_1. I periodically merge RC_1_1 into master which would make it pick up this change

mnaamani:
Ok great. I prepared https://github.com/arvidn/libtorrent/pull/1337 for RC_1_1.

Yes I noticed that you periodically merge RC_1_1 into master, although you will probably have a minor merge conflict to resolve because the implementation of stack_allocator on master is slightly different.

arvidn:
given that there's a pull request against RC_1_1, this one can be closed, right?

mnaamani:
Closing.. changes implemented in  #1337

PR #1335, opened 2016-11-17
Fix alerts zero allocation for RC_1_1

mnaamani:
Implementing similar fixes addressed in: https://github.com/arvidn/libtorrent/pull/1331 to handle the 0 size allocations on the stack_allocator in the following alerts:

- picker_log_alert
- dht_get_peers_reply_alert
- dht_pkt_alert

And updating the stack_allocator::allocate and stack_allocator::copy_buffer methods to return -1 on zero allocation.

### DIFF
--- a/src/alert.cpp
+++ b/src/alert.cpp
@@ -1855,7 +1855,7 @@ namespace libtorrent {
 
 	char const* dht_pkt_alert::pkt_buf() const
 	{
-		return m_alloc.ptr(m_msg_idx);
+		return m_size > 0 ? m_alloc.ptr(m_msg_idx) : nullptr;
 	}
 
 	int dht_pkt_alert::pkt_size() const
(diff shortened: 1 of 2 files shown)


src/alert.cpp
```diff
@@ -1855,7 +1855,7 @@ namespace libtorrent {
 
 	char const* dht_pkt_alert::pkt_buf() const
 	{
-		return m_alloc.ptr(m_msg_idx);
+		return m_size > 0 ? m_alloc.ptr(m_msg_idx) : nullptr;
```
arvidn: instead of putting this in 3 places, you could just put it in the ptr() function

arvidn:
I was going to suggest adding a test for this case too, but it seems there is no test for the stack_allocator yet.

src/alert.cpp
arvidn: also, RC_1_1 still supports C++98. master is on C++14

arvidn: so nullptr has to be NULL in this branch

PR #1336, opened 2016-11-18
more use of std::make_shared, fixing warnings of over-aligned

aldenml:


include/libtorrent/proxy_base.hpp
```diff
@@ -41,11 +41,11 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent {
 
-class proxy_base : boost::noncopyable
+class proxy_base
 {
 public:
 
-	typedef std::function<void(error_code const&)> handler_type;
+	using handler_type = std::function<void(error_code const&)>;
```
arvidn: this should probably be reviewed. with c++11 the handler should probably be moved instead of heap allocated

src/kademlia/node.cpp
```diff
@@ -429,15 +429,9 @@ void node::get_peers(sha1_hash const& info_hash
...
-	std::shared_ptr<dht::get_peers> ta;
-	if (m_settings.privacy_lookups)
-	{
-		ta.reset(new dht::obfuscated_get_peers(*this, info_hash, dcallback, ncallback, noseeds));
-	}
-	else
-	{
-		ta.reset(new dht::get_peers(*this, info_hash, dcallback, ncallback, noseeds));
-	}
+	auto ta = m_settings.privacy_lookups
```
arvidn: I'm a bit surprised this works.. what's the return type of this ? : expression?

aldenml: since `obfuscated_get_peers` inherits from `get_peers`, and `make_shared` (I assume with the magic of SFINAE and `shared_ptr`) behaves like it's covariant, the type is `shared_ptr<get_peers>`. (I didn't know either)

PR #1337, opened 2016-11-18
update stack_allocator to support 0 size allocations for RC_1_1

mnaamani:
Implementing zero allocation support in `libtorrent::aux::stack_allocator` as discussed in https://github.com/arvidn/libtorrent/pull/1331

### DIFF
--- a/include/libtorrent/stack_allocator.hpp
+++ b/include/libtorrent/stack_allocator.hpp
@@ -62,6 +62,7 @@ namespace libtorrent { namespace aux
 		int copy_buffer(char const* buf, int size)
 		{
 			int ret = int(m_storage.size());
+			if (size < 1) return -1;
 			m_storage.resize(ret + size);
 			memcpy(&m_storage[ret], buf, size);
 			return ret;
@@ -69,6 +70,7 @@ namespace libtorrent { namespace aux
 
 		int allocate(int bytes)
 		{
+			if (bytes < 1) return -1;
 			int ret = int(m_storage.size());
 			m_storage.resize(ret + bytes);
 			return ret;
@@ -76,14 +78,14 @@ namespace libtorrent { namespace aux
 
 		char* ptr(int idx)
 		{
-			TORRENT_ASSERT(idx >= 0);
+			if(idx < 0) return NULL;
 			TORRENT_ASSERT(idx < int(m_storage.size()));
 			return &m_storage[idx];
 		}
 
 		char const* ptr(int idx) const
 		{
-			TORRENT_ASSERT(idx >= 0);
+			if(idx < 0) return NULL;
 			TORRENT_ASSERT(idx < int(m_storage.size()));
 			return &m_storage[idx];
 		}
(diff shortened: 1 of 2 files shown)


aldenml:
since this class is so critical for the alerts, the removal of those asserts makes me feel a little nervous for future correctness of new alerts, since the index returned in the `allocate` can be used after that, if it's miscalculated then returning `NULL` in `ptr` could potentially give the wrong impression of a "good result" or "stop condition".

aldenml:
I think the problem is returning just the index, it's like it's not enough information support the idea of 0 size allocations

arvidn:
how so? any pointer is valid for a zero size buffer.

aldenml:
you know, I'm trying to come up with a plausible hypothetical situation, but I'm getting none. If only `ptr` can receive a desired size (and in master return a `span`) then I assume you can keep the asserts (removing them is my only concern).

arvidn:
the C and C++ standard requires objects to have unique pointers, which means that even empty objects must use at least one byte. but in this case I don't think that's important

include/libtorrent/stack_allocator.hpp
```diff
@@ -62,29 +62,31 @@ namespace libtorrent { namespace aux
...
 			m_storage.resize(ret + bytes);
 			return ret;
 		}
 
 		char* ptr(int idx)
 		{
-			TORRENT_ASSERT(idx >= 0);
-			TORRENT_ASSERT(idx < int(m_storage.size()));
+			if(idx < 0) return NULL;
+			if(idx >= int(m_storage.size())) return NULL;
```
arvidn: I think this second test may still may sense to leave as an assert. I don't see a case where this would trigger without having a bug somewhere

arvidn: I'm open to be convinced otherwise though

mnaamani: Actually sorry that only applies for the first assert.

mnaamani:
@aldenml is this the kind of situation you were thinking of?

`std::string s(alloc.ptr(-1))` => `std::string s(NULL)` -> segmentation fault

aldenml:
@mnaamani thinking of something more subtle, like:
```cpp
m_idx = alloc.copy_string(error_msg);

....other method/code....

m_idx = <wrong calculation that yields a very big number>
char const* error = alloc.ptr(m_idx);
snprintf(buff, "error: %s", error != NULL ? error : "none");
```
you will have a hard time detecting the bug in `<wrong calculation`, does this makes sense?

arvidn:
as for strings, if an allocation is made via ``copy_string``, there will always be at least a 0-terminator. as for mutating indices to become invalid, maybe the type of the returned index should be fixed to prevent any arithmetic and mutation. There shouldn't really be any need to do any of that.

arvidn:
now there is a unit test for stack_allocator. would you mind adding a test for this case as well?

arvidn:
you have to rebase on top of RC_1_1

mnaamani:
Sure I can do that

mnaamani:
Does Travis CI run the unit tests ?

mnaamani:
Should I squash the commits down to a single commit?

arvidn:
yes, travis runs the tests. if you don't mind, squashing would be nice. lgtm

PR #1338, opened 2016-11-19
fixed a few warnings related to signed/unsigned conversions

aldenml:


include/libtorrent/stat.hpp
```diff
@@ -75,18 +75,18 @@ namespace libtorrent
 
 		// should be called once every second
 		void second_tick(int tick_interval_ms);
-		int rate() const { return m_5_sec_average; }
-		int low_pass_rate() const { return m_5_sec_average; }
+		int rate() const { return int(m_5_sec_average); }
+		int low_pass_rate() const { return int(m_5_sec_average); }
 
-		std::int64_t total() const { return m_total_counter; }
+		std::int64_t total() const { return std::int64_t(m_total_counter); }
```
arvidn: actually.. would changing the member variables here to be signed silence the warnings you see? If so, I think I'd prefer that

include/libtorrent/aux_/disable_warnings_push.hpp
```diff
@@ -86,5 +86,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #ifdef _MSC_VER
 #pragma warning(push, 1)
 // warning C4005: macro redefinition
-#pragma warning( disable : 4005 )
+#pragma warning(disable : 4005)
+// expression before comma has no effect; expected expression with side-effect
+#pragma warning(disable : 4548)
```
arvidn: where is this warning issued?

aldenml: if you add `/Wall` to msvc you will get a lot interesting messages :)
This helps with a boost/asio import in `socket.hpp`

include/libtorrent/stat.hpp
aldenml: ok, i will take a look at the type of this variables.

aldenml:
@arvidn what is your position/opinion in using `int` vs `std::int32_t`?

aldenml:
I know the differences but I would like to know if you have some criteria to for example: return types when the variable is of fixed width

include/libtorrent/aux_/disable_warnings_push.hpp
arvidn: oh, right, this is in the disable warnings header.

arvidn:
I don't really have opinions about ``int`` vs ``std::int32_t``. ``int`` is shorter and may be better in places where it's not important for it to be exactly 32 bits. Since both are signed, the compiler is still free to widen them when held in registers so in practice I don't think there's really any difference.

The one place would be in a 6502 compiler, where ``int`` is 16 bits :)

arvidn:
I'd say if the type has a specific width, using the types with specified widths makes sense. for generic integers, ``int`` makes sense.

include/libtorrent/stat.hpp
arvidn: the reason I used unsigned types here to begin with was because when I wrote this class (several years ago) I had not fully understood the implications and meaning of [unsigned integers](http://blog.libtorrent.org/2016/05/unsigned-integers/).

PR #1344, opened 2016-11-20
keep old listen sockets if they're still valid

ssiloti:
This is to support multi-home. We need to be able to keep track of which socket
a DHT node or UTP connection should use. We also need to generate notifications
when local endpoints come and go so that the DHT tracker knows when to create
or delete nodes. The easiest way to do this is to keep the same socket for as
long as its local endpoint is valid. This way the nodes and connections can
simply reference the socket itself and generating notifications is trivial.

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -270,6 +270,37 @@ namespace aux {
 	}
 #endif // TORRENT_DISABLE_DHT
 
+	std::list<listen_socket_t>::iterator partition_listen_sockets(
+		std::vector<listen_endpoint_t>& eps
+		, std::list<listen_socket_t>& sockets)
+	{
+		return std::partition(sockets.begin(), sockets.end()
+			, [&eps](listen_socket_t const& sock)
+		{
+			auto match = std::find_if(eps.begin(), eps.end()
+				, [&sock](listen_endpoint_t const& ep)
+			{
+				return ep.ssl == sock.ssl
+					&& ep.port == sock.original_port
+					&& ep.device == sock.device
+					&& ep.addr == sock.local_endpoint.address();
+			});
+
+			if (match != eps.end())
+			{
+				// remove the matched endpoint so that another socket can't match it
+				// this also signals to the caller that it doesn't need to create a
+				// socket for the endpoint
+				eps.erase(match);
+				return true;
+			}
+			else
+			{
+				return false;
+			}
+		});
+	}
+
 	void session_impl::init_peer_class_filter(bool unlimited_local)
 	{
 		// set the default peer_class_filter to use the local peer class
@@ -1384,6 +1415,7 @@ namespace aux {
 
 		listen_socket_t ret;
 		ret.ssl = (flags & open_ssl_socket) != 0;
+		ret.original_port = bind_ep.port();
 		int last_op = 0;
 		socket_type_t const sock_type
 			= (flags & open_ssl_socket)
@@ -1702,20 +1734,6 @@ namespace aux {
 		reopen_listen_sockets();
 	}
 
-	namespace
-	{
-		struct listen_endpoint_t
-		{
-			listen_endpoint_t(address adr, int p, std::string dev, bool s)
-				: addr(adr), port(p), device(dev), ssl(s) {}
-
-			address addr;
-			int port;
-			std::string device;
-			bool ssl;
-		};
-	}
-
 	void session_impl::reopen_listen_sockets()
 	{
 #ifndef TORRENT_DISABLE_LOGGING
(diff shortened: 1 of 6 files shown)


src/session_impl.cpp
```diff
@@ -1807,16 +1805,68 @@ namespace aux {
...
-				}
+		// sockets we are keeping get moved to this list to prevent a socket from matching
+		// multiple endpoints
+		std::list<listen_socket_t> keep;
+
+		// remove any sockets which are no longer in the set of endpoints
+		// to listen on
+		// warning: O(n^2) operation!
+		// hopefully the system doesn't have too many interfaces
+		for (auto sock = m_listen_sockets.begin()
```
arvidn: do you think there's any way to pull this logic out into its own function, so it can have tests?

src/session_impl.cpp
```diff
@@ -1807,16 +1805,68 @@ namespace aux {
...
+			// this socket's local_endpoint is not on the list of endpoints to listen on
+			// it has got to go
+			// TODO notify interested parties of this socket's demise
+#ifndef TORRENT_DISABLE_LOGGING
+			session_log("Closing listen socket for %s on device \"%s\""
+				, print_endpoint(sock->local_endpoint).c_str(), sock->device.c_str());
+#endif
+			if (sock->sock) sock->sock->close(ec);
+			if (sock->udp_sock) sock->udp_sock->close();
+			sock = m_listen_sockets.erase(sock);
```
arvidn: is there a risk there are still outstanding operations on this socket? if so, will the handler deal with the listen_socket_t object actually having been deleted?

src/session_impl.cpp
```diff
@@ -1807,16 +1805,68 @@ namespace aux {
...
+		// to listen on
+		// warning: O(n^2) operation!
+		// hopefully the system doesn't have too many interfaces
+		for (auto sock = m_listen_sockets.begin()
+			; sock != m_listen_sockets.end();)
+		{
+			auto match = std::find_if(eps.begin(), eps.end()
+				, [sock, port_retries](listen_endpoint_t const& ep)
+					{ return ep.ssl == sock->ssl
+						&& (ep.port == 0 || (sock->local_endpoint.port() >= ep.port
```
arvidn: the port comparison is not obvious to me. perhaps there should be a comment in the code. I guess the port can't be expected to be an exact match

src/session_impl.cpp
arvidn: do you think this would be simplified by using ``std::remove_if``? and the second pass could be to call close and actually erase the elements. I suppose that requires move support, or perhaps list nodes are swapped just with the chain pointers and stay stable

arvidn: actually, in order to be able to make a second pass of the elements that are to be removed, I believe you'd need to use ``std::partition``

arvidn:
looks reasonable. there is a pretty large chunk of logic that doesn't have test coverage though, which is a bit scary

src/session_impl.cpp
ssiloti: Currently handlers store a weak_ptr to the socket/acceptor so that they can detect if it has been deleted. If it turns out to be necessary to store a reference to listen_socket_t in a handler then my plan is to store listen_socket_t itself in a shared_ptr rather than the socket and acceptor.

arvidn: makes sense

src/session_impl.cpp
```diff
@@ -1701,6 +1702,20 @@ namespace aux {
 		reopen_listen_sockets();
 	}
 
+	namespace
+	{
+		struct listen_endpoint_t
+		{
+			listen_endpoint_t(address adr, int p, std::string dev, bool s)
+				: addr(adr), port(p), device(dev), ssl(s) {}
```
aldenml: no need of `std::move(dev) here? harmful or not important?

ssiloti: Not important. None of this code is remotely performance sensitive. For most people it won't be executed more than a few times a day.

arvidn:
closing and re-opening triggers the tests. appveyor seems to be having some issue right now though

test/test_listen_socket.cpp
```diff
@@ -0,0 +1,205 @@
...
+		TEST_EQUAL(e1.ssl, ssl);
+		TEST_EQUAL(e1.port, port);
+		TEST_EQUAL(e1.addr, addr);
+		TEST_EQUAL(e1.device, dev);
+	}
+}
+
+TORRENT_TEST(partition_listen_sockets)
+{
+	{
```
arvidn: would you mind turning these blocks into individual test cases? I don't think they're coupled in any way, right?

src/session_impl.cpp
```diff
@@ -270,6 +270,37 @@ namespace aux {
 	}
 #endif // TORRENT_DISABLE_DHT
 
+	std::list<listen_socket_t>::iterator partition_listen_sockets(
+		std::vector<listen_endpoint_t>& eps
+		, std::list<listen_socket_t>& sockets)
+	{
+		return std::partition(sockets.begin(), sockets.end()
+			, [&eps](listen_socket_t const& sock)
```
arvidn: do you understand why the test coverage considers all this not-covered?

arvidn: at first I thought the test wasn't being run, but it's added to the Jamfile..

test/test_listen_socket.cpp
```diff
@@ -30,8 +30,6 @@ POSSIBILITY OF SUCH DAMAGE.
 
 */
 
-#if TORRENT_USE_IPV6
```
arvidn: ah, I see. the test coverage build is configured without ipv6 then I take it..

PR #1345, opened 2016-11-21
more fixes in signed compare warnings

aldenml:


include/libtorrent/peer_class_set.hpp
```diff
@@ -56,7 +56,7 @@ namespace libtorrent {
 	private:
 
 		// the number of elements used in the m_class array
-		std::uint8_t m_size;
+		std::int8_t m_size;
```
arvidn: this is not obvious I think. since it's only 8 bits to begin with, loosing one is kind of significant

src/kademlia/dht_storage.cpp
```diff
@@ -251,7 +251,7 @@ namespace
 
 					// pick this peer with probability
 					// <peers left to pick> / <peers left in the set>
-					if (random(uint32_t(candidates--)) > to_pick)
+					if (int(random(std::uint32_t(candidates--))) > to_pick)
```
arvidn: hm.. I would have expected making ``to_pick`` unsigned would be more appropriate. ``random()`` will return independent bits as far as I understand, and there's a 50% chance the most significant bit is set. I think this needs to be an unsigned comparison (which it is currently)

include/libtorrent/peer_class_set.hpp
aldenml: it seems to my that the maximum number is `15`, am I wrong?
```
std::array<peer_class_t, 15> m_class;
```

arvidn:
lgtm

PR #1347, opened 2016-11-21
more fixes of integral type conversion warnings

aldenml:


src/lazy_bdecode.cpp
```diff
@@ -307,7 +307,7 @@ namespace libtorrent
 		m_data.start = start;
 		m_size = length;
 		m_begin = start - 1 - num_digits(length);
-		m_len = start - m_begin + length;
+		m_len = std::uint32_t(start - m_begin + length);
```
arvidn: maybe there should be an assert here too..

PR #1349, opened 2016-11-22
more integral type conversion warning fixes, now only needs shorten-64-to-32

aldenml:


src/block_cache.cpp
```diff
@@ -500,7 +500,7 @@ void block_cache::update_cache_state(cached_piece_entry* p)
 	src->erase(p);
 	dst->push_back(p);
 	p->expire = aux::time_now();
-	p->cache_state = desired_state;
```
arvidn: it seems like desired_state should have the right type here instead

src/block_cache.cpp
```diff
@@ -460,7 +460,7 @@ void block_cache::cache_hit(cached_piece_entry* p, void* requester, bool volatil
 	// move into L2 (frequently used)
 	m_lru[p->cache_state].erase(p);
 	m_lru[target_queue].push_back(p);
-	p->cache_state = target_queue;
+	p->cache_state = std::uint16_t(target_queue);
```
arvidn: it seems ``target_queue`` should have the right type here.. maybe it should even be an enum class

src/bt_peer_connection.cpp
```diff
@@ -614,8 +614,8 @@ namespace libtorrent
 		// this is an invalid setting, but let's just make the best of the situation
 		int const enc_level = m_settings.get_int(settings_pack::allowed_enc_level);
 		std::uint8_t const crypto_provide = ((enc_level & settings_pack::pe_both) == 0)
-			? settings_pack::pe_both
-			: enc_level;
+			? std::uint8_t(settings_pack::pe_both)
```
arvidn: could you fix the type the enum in ``settings_pack`` instead?

aldenml: sure

src/bt_peer_connection.cpp
```diff
@@ -1618,7 +1617,7 @@ namespace libtorrent
 	{
 		char buf[35];
 		char* ptr = buf + 6;
-		detail::write_uint8(type, ptr);
+		detail::write_uint8(std::uint8_t(type), ptr);
```
arvidn: it seems strange to say the type twice. in a sense, since the type passed to the function is explicit already, it seems like this cast should be implied. The only way I can think of doing that though would be to make ``write_uint8()`` a template

src/disk_io_thread.cpp
```diff
@@ -1767,7 +1767,7 @@ namespace libtorrent
 		j->storage = storage->shared_from_this();
 		j->buffer.string = allocate_string_copy(p.c_str());
 		j->callback = std::move(handler);
-		j->flags = flags;
+		j->flags = std::uint8_t(flags);
```
arvidn: in all of these cases of ``flags``, I think it would be better to change the type of the argument. That way the information about the truncation is propagated up the call tree and allows warnings to be triggered earlier

src/escape_string.cpp
```diff
@@ -297,10 +297,10 @@ namespace libtorrent
 			i += available_input;
 
 			// encode inbuf to outbuf
-			outbuf[0] = (inbuf[0] & 0xfc) >> 2;
-			outbuf[1] = ((inbuf[0] & 0x03) << 4) | ((inbuf [1] & 0xf0) >> 4);
-			outbuf[2] = ((inbuf[1] & 0x0f) << 2) | ((inbuf [2] & 0xc0) >> 6);
-			outbuf[3] = inbuf[2] & 0x3f;
+			outbuf[0] = std::uint8_t((inbuf[0] & 0xfc) >> 2);
```
aldenml: I don't know...I will try that

src/bt_peer_connection.cpp
aldenml: true, I will try the template idea

src/escape_string.cpp
```diff
@@ -416,17 +416,17 @@ namespace libtorrent
 			}
 
 			// decode inbuf to outbuf
-			outbuf[0] = inbuf[0] << 3;
+			outbuf[0] = std::uint8_t(inbuf[0] << 3);
```
arvidn: would ``&& 0xff`` do it here?

src/kademlia/node_id.cpp
```diff
@@ -131,9 +131,9 @@ node_id generate_id_impl(address const& ip_, std::uint32_t r)
 
 	id[0] = (c >> 24) & 0xff;
 	id[1] = (c >> 16) & 0xff;
-	id[2] = ((c >> 8) & 0xf8) | random(0x7);
+	id[2] = ((c >> 8) & 0xf8) | std::uint8_t(random(0x7));
```
aldenml: OK

src/pe_crypto.cpp
```diff
@@ -351,7 +351,7 @@ namespace libtorrent
 void rc4_init(const unsigned char* in, unsigned long len, rc4 *state)
 {
 	size_t const key_size = sizeof(state->buf);
-	unsigned char key[key_size], tmp, *s;
+	std::uint8_t key[key_size], tmp, *s;
```
aldenml: sure, I will do it

src/pe_crypto.cpp
```diff
@@ -386,15 +386,15 @@ void rc4_init(const unsigned char* in, unsigned long len, rc4 *state)
...
+	std::uint8_t x, y, *s, tmp;
 	unsigned long n;
 
 	TORRENT_ASSERT(out != nullptr);
 	TORRENT_ASSERT(state != nullptr);
 
 	n = outlen;
-	x = state->x;
-	y = state->y;
+	x = std::uint8_t(state->x);
```
arvidn: should ``state->x`` and ``state->y`` be ``uint8_t`` too?

arvidn: or is this meant to truncate? if so, I think ``&& 0xff`` would be clearer

aldenml: I will review that again, last time I tried was a problem due to some loops

src/torrent.cpp
```diff
@@ -8380,30 +8380,30 @@ namespace libtorrent
 			int const lost_seconds = seconds - m_started;
 			m_active_time += lost_seconds;
 		}
-		m_started = clamped_subtract(m_started, seconds);
+		m_started = std::uint16_t(clamped_subtract(m_started, seconds));
```
arvidn: should ``clamped_subtract`` return ``uint16_t`` directly?

src/utp_stream.cpp
```diff
@@ -1508,7 +1508,7 @@ void utp_socket_impl::parse_sack(std::uint16_t packet_ack, std::uint8_t const* p
 	// for each byte
 	for (std::uint8_t const* end = ptr + size; ptr != end; ++ptr)
 	{
-		unsigned char bitfield = unsigned(*ptr);
+		std::uint8_t bitfield = std::uint8_t(*ptr);
```
arvidn: ``ptr`` already points to ``uint8_t`` here, no?

src/torrent.cpp
aldenml: I don't think so, but maybe `clamped_subtract_s16`, or maybe changing `clamped_subtract`

src/utp_stream.cpp
aldenml: true, I see now that the problem was `unsigned` is `unsigned int`

src/bt_peer_connection.cpp
```diff
@@ -344,7 +344,7 @@ namespace libtorrent
 #endif
 		char msg[] = {0,0,0,3, msg_dht_port, 0, 0};
 		char* ptr = msg + 5;
-		detail::write_uint16(listen_port, ptr);
+		detail::write_uint16(std::uint16_t(listen_port), ptr);
```
aldenml: pending to remove `std::uint16_t(`, waiting for CI output

aldenml: and others below too

include/libtorrent/io.hpp
```diff
@@ -117,37 +117,37 @@ namespace libtorrent
 		{ return read_impl(start, type<std::uint8_t>()); }
 
 
-		template <class OutIt>
-		void write_uint64(std::uint64_t val, OutIt& start)
-		{ write_impl(val, start); }
+		template <class T, class OutIt>
+		void write_uint64(T val, OutIt& start)
+		{ write_impl<std::uint64_t, OutIt>(std::uint64_t(val), start); }
```
arvidn: The two things are ``enable_if`` and ``TORRENT_ASSERT``

src/torrent.cpp
```diff
@@ -8332,17 +8332,17 @@ namespace libtorrent
...
 	{
 		if (a < b) return 0;
-		return a - b;
+		return std::uint16_t(a - b);
 	}
 
-	int clamped_subtract_s16(int a, int b)
+	std::int16_t clamped_subtract_s16(int a, int b)
 	{
 		if (a + (std::numeric_limits<std::int16_t>::min)() < b)
```
arvidn: I think you need a test similar to this one for the signed version too... or do you?

include/libtorrent/io.hpp
aldenml: I think it's a good idea, I will apply this

src/torrent.cpp
aldenml: ok

src/bt_peer_connection.cpp
aldenml: gcc still needs the cast

src/torrent.cpp
aldenml: I don't think the test is needed in `clamped_subtract_u16` since the the min unsigned is `0`. Could you confirm this?

src/bt_peer_connection.cpp
arvidn: interesting

arvidn:
that travis error seems to be my merge.. will fix

PR #1354, opened 2016-11-26
fixed sign-compare warnings and some refactor

aldenml:
for preliminary review

### DIFF
--- a/include/libtorrent/session_handle.hpp
+++ b/include/libtorrent/session_handle.hpp
@@ -429,7 +429,7 @@ namespace libtorrent
 		// calling the callback in between is convenient.
 		void dht_put_item(std::array<char, 32> key
 			, std::function<void(entry&, std::array<char, 64>&
-				, std::uint64_t&, std::string const&)> cb
+				, std::int64_t&, std::string const&)> cb
 			, std::string salt = std::string());
 
 		void dht_get_peers(sha1_hash const& info_hash);
--- a/include/libtorrent/session_handle.hpp
+++ b/include/libtorrent/session_handle.hpp
@@ -407,7 +407,7 @@ namespace libtorrent
 		// 	To generate the signature, you may want to use the
 		// 	``sign_mutable_item`` function.
 		//
-		// std::uint64_t& seq
+		// std::int64_t& seq
 		// 	current sequence number. May be zero if there is no current value.
 		// 	The function is expected to set this to the new sequence number of
 		// 	the value that is to be stored. Sequence numbers must be monotonically
--- a/include/libtorrent/sliding_average.hpp
+++ b/include/libtorrent/sliding_average.hpp
@@ -92,9 +92,8 @@ struct average_accumulator
 
 	void add_sample(int s)
 	{
-		TORRENT_ASSERT(s >= 0);
 		++m_num_samples;
-		m_sample_sum += std::uint64_t(s);
+		m_sample_sum += s;
 	}
 
 	int mean()
@@ -106,14 +105,14 @@ struct average_accumulator
 		// let the average roll over, but only be worth a
 		// single sample
 		m_num_samples = 1;
-		m_sample_sum = std::uint64_t(ret);
+		m_sample_sum = ret;
 		return ret;
 	}
 
 private:
 
 	int m_num_samples = 0;
-	std::uint64_t m_sample_sum = 0;
+	std::int64_t m_sample_sum = 0;
 };
 
 }
--- a/include/libtorrent/sliding_average.hpp
+++ b/include/libtorrent/sliding_average.hpp
@@ -36,6 +36,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <cstdint>
 #include <cstdlib> // for std::abs
 
+#include "assert.hpp"
+
 namespace libtorrent
 {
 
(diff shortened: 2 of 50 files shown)


include/libtorrent/sliding_average.hpp
```diff
@@ -90,8 +92,9 @@ struct average_accumulator
 
 	void add_sample(int s)
 	{
+		TORRENT_ASSERT(s >= 0);
 		++m_num_samples;
-		m_sample_sum += s;
```
arvidn: I wonder if there's a good reason why this only allows positive samples. It seems a bit arbitrary. what would you think about making ``m_sample_sum`` signed instead?

src/file.cpp
```diff
@@ -1364,13 +1364,13 @@ namespace libtorrent
...
 		{
-			std::uint32_t ret = 1;
-			for (int i = 0; i < str.size(); ++i)
+			std::int32_t ret = 1;
+			for (auto ch : str)
 			{
-				if (str[i] == 0) continue;
-					ret *= int(str[i]);
+				if (ch == 0) continue;
+				ret *= int(ch);
```
arvidn: I don't think this is valid. With ``ret`` being signed, it's undefined behavior to overflow it, which this multiplication most likely will do. instead, cast the character to ``std::uint32_t``

src/kademlia/get_item.cpp
```diff
@@ -200,7 +200,7 @@ void get_item_observer::reply(msg const& m)
 	bdecode_node q = r.dict_find_int("seq");
 	if (q)
 	{
-		seq = sequence_number(q.int_value());
+		seq = sequence_number(std::uint64_t(q.int_value()));
```
arvidn: would it make sense to make ``sequence_number`` be backed by a signed integer instead?

src/kademlia/node_id.cpp
```diff
@@ -210,8 +211,8 @@ node_id generate_prefix_mask(int bits)
 	TORRENT_ASSERT(bits <= 160);
 	node_id mask(nullptr);
 	int b = 0;
```
arvidn: if you make ``b`` unsigned instead, could you avoid the cast to ``size_t``?

src/kademlia/traversal_algorithm.cpp
```diff
@@ -207,22 +207,23 @@ void traversal_algorithm::add_entry(node_id const& id
 
 	if (m_results.size() > 100)
 	{
-		for (int i = 100; i < int(m_results.size()); ++i)
+		std::for_each(m_results.begin() + 100, m_results.end()
+			, [this](std::shared_ptr<observer>& ptr)
```
arvidn: I think this could be a ``std::shared_ptr<observer> const& ptr``, right?

src/part_file.cpp
```diff
@@ -118,8 +118,9 @@ namespace libtorrent
 
 			for (int i = 0; i < num_pieces; ++i)
 			{
-				int slot = read_uint32(ptr);
-				if (slot == 0xffffffff) continue;
+				std::uint32_t const uslot = read_uint32(ptr);
+				if (uslot == 0xffffffff) continue;
+				int const slot = int(uslot);
```
arvidn: I would imagine there's very little, if any at all, arithmetic on ``slot``. perhaps it could be an unsigned type (perhaps similar to ``allocation_slot`` in stack_allocator). making it a special type should probably be a separate patch though

src/piece_picker.cpp
```diff
@@ -413,7 +413,7 @@ namespace libtorrent
 		for (int i = 0; i < num_pieces; ++i)
 		{
 			int h = have[i];
```
arvidn: would it make more sense to widen ``h`` here instead of truncating the result from ``count()``?

src/piece_picker.cpp
```diff
@@ -496,7 +496,7 @@ namespace libtorrent
 				block_info const* info = blocks_for_piece(dp);
 				for (int k = 0; k < num_blocks; ++k)
 				{
-					TORRENT_ASSERT(info[k].piece_index == dp.index);
+					TORRENT_ASSERT(info[k].piece_index == int(dp.index));
```
arvidn: hm.. why are these incompatible? is ``dp.index`` a bitfield?

arvidn:
looks reasonable for the most part. do you understand why the tests for the openssl builds are failing?

include/libtorrent/sliding_average.hpp
aldenml: I think it makes sense

src/file.cpp
aldenml: true, I will do it

src/kademlia/node_id.cpp
aldenml: it could be, I will try it.

src/piece_picker.cpp
aldenml: to review

aldenml:
no idea why `gcc+openssl` is failing, more likely I introduced a subtle bug, I will make the comments changes and rebase, but I will hunt the crash in my travis account

arvidn:
sorry about the conflicts...

aldenml:
it's OK :)

include/libtorrent/session_handle.hpp
```diff
@@ -407,7 +407,7 @@ namespace libtorrent
 		// 	To generate the signature, you may want to use the
 		// 	``sign_mutable_item`` function.
 		//
-		// std::uint64_t& seq
```
arvidn: this is an API change. it should probably be mentioned in the Changelog

PR #1360, opened 2016-11-28
Add bt_peer_connection::write_dht_port()

pavel-pimenov:


src/bt_peer_connection.cpp
```diff
@@ -326,12 +326,17 @@ namespace libtorrent
...
-		if (m_supports_dht_port && m_ses.has_dht())
-		{
-			int const port = m_ses.external_udp_port();
-			if (port >= 0) write_dht_port(port);
-		}
+        if (m_supports_dht_port && m_ses.has_dht())
+        {
+            int const port = m_ses.external_udp_port();
+            if (port >= 0) write_dht_port(port);
+        }
```
arvidn: you replaced tabs with spaces here, I think that was a mistake, no?

include/libtorrent/bt_peer_connection.hpp
```diff
@@ -272,6 +272,7 @@ namespace libtorrent
 #endif
 
 	private:
+        void write_dht_port();
```
arvidn: sorry, I didn't notice this the first time, but this indentation isn't right either, it should be tabs.

arvidn:
one more (from travis): src/bt_peer_connection.cpp:331:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]

PR #1361, opened 2016-11-28
Small optimisation identify_client

pavel-pimenov:


src/identify_client.cpp
```diff
@@ -273,7 +274,6 @@ namespace
 		, {0, "-G3", "G3 Torrent"}
 		, {0, "-FG", "FlashGet"}
 		, {0, "-ML", "MLdonkey"}
-		, {0, "-MG", "Media Get"}
```
arvidn: why did this move?

pavel-pimenov: I counted the number of hits
Revert... "Media Get" only russian locations

arvidn:
did you see this? https://travis-ci.org/arvidn/libtorrent/jobs/179440439#L1295

it's an unsigned  <-> signed comparison

src/identify_client.cpp
```diff
@@ -137,7 +137,7 @@ namespace
 
 	// only support BitTorrentSpecification
 	// must be ordered alphabetically
-	map_entry name_map[] =
+	static map_entry name_map[] =
```
arvidn: these should probably be ``const`` too

src/identify_client.cpp
```diff
@@ -367,15 +367,14 @@ namespace libtorrent
...
 		if (p.is_all_zeros()) return "Unknown";
 
 		// ----------------------
 		// non standard encodings
 		// ----------------------
 
-		int num_generic_mappings = sizeof(generic_mappings) / sizeof(generic_mappings[0]);
+		const int num_generic_mappings = sizeof(generic_mappings) / sizeof(generic_mappings[0]);
 
 		for (int i = 0; i < num_generic_mappings; ++i)
```
arvidn: to luxury it up a bit, this could even be a for-range loop over the C-array

PR #1362, opened 2016-11-28
fixed shorten-64-to-32 warnings and some refactor [dont merge, only referece]

aldenml:


include/libtorrent/file_storage.hpp
```diff
@@ -150,13 +150,13 @@ namespace libtorrent
 		// the number of characters in the name. If this is
 		// name_is_owned, name is 0-terminated and owned by this object
 		// (i.e. it should be freed in the destructor). If
-		// the len is not name_is_owned, the name pointer doesn not belong
+		// the len is not name_is_owned, the name pointer doesn't not belong
```
aldenml: to fix

src/bdecode.cpp
```diff
@@ -697,7 +697,7 @@ namespace libtorrent
 					// we push it into the stack so that we know where to fill
 					// in the next_node field once we pop this node off the stack.
 					// i.e. get to the node following the dictionary in the buffer
-					ret.m_tokens.push_back(bdecode_token(start - orig_start
+					ret.m_tokens.push_back(bdecode_token(std::uint32_t(start - orig_start)
```
aldenml: do you think it's better to make the constructor take `std::size_t` and inside make the cast?

src/disk_io_thread.cpp
```diff
@@ -658,7 +658,7 @@ namespace libtorrent
 		if (!failed)
 		{
 			TORRENT_PIECE_ASSERT(!error, pe);
-			std::uint32_t write_time = total_microseconds(clock_type::now() - start_time);
+			int const write_time = int(total_microseconds(clock_type::now() - start_time));
```
aldenml: I made all these calls the same (cast to `int`) since some of them already were `int`, but do you think it's better to take the `std::int64_t` and cast just in `add_sample`?

src/disk_io_thread.cpp
```diff
@@ -1178,7 +1178,7 @@ namespace libtorrent
 		j->ret = ret;
 
 		time_point now = clock_type::now();
-		m_job_time.add_sample(total_microseconds(now - start_time));
+		m_job_time.add_sample(int(total_microseconds(now - start_time)));
```
aldenml: I assume microseconds don't overflow the `int`, is that right?

src/lsd.cpp
```diff
@@ -242,7 +242,7 @@ void lsd::on_announce(udp::endpoint const& from, char const* buf
 	{
 		// we expect it to be hexadecimal
 		// if it isn't, it's not our cookie anyway
-		std::int32_t const cookie = strtol(cookie_iter->second.c_str(), nullptr, 16);
+		std::int32_t const cookie = int(strtol(cookie_iter->second.c_str(), nullptr, 16));
```
aldenml: is this cookie only 32 bits?

src/torrent.cpp
```diff
@@ -1770,7 +1770,7 @@ namespace libtorrent
 			// maybe this whole section should move to need_picker()
 			need_picker();
 
-			peer_request pr = m_torrent_file->map_file(i, 0, fs.file_size(i));
+			peer_request pr = m_torrent_file->map_file(i, 0, int(fs.file_size(i)));
```
aldenml: why `map_file` does not take `std::int64_t`?

src/tracker_manager.cpp
```diff
@@ -260,11 +260,11 @@ namespace libtorrent
 
 	void tracker_manager::update_transaction_id(
 		std::shared_ptr<udp_tracker_connection> c
-		, std::uint64_t tid)
+		, std::uint64_t const tid)
 	{
 		TORRENT_ASSERT(is_single_thread());
 		m_udp_conns.erase(c->transaction_id());
-		m_udp_conns[tid] = c;
+		m_udp_conns[std::uint32_t(tid)] = c;
```
aldenml: why this type mismatch?

src/bdecode.cpp
arvidn: it might be cleaner to do that, yeah. and perhaps add an assert too

src/lsd.cpp
aldenml: yes, this is OK, creation is `m_cookie((random(0x7fffffff) ^ std::uintptr_t(this)) & 0x7fffffff)`

src/torrent.cpp
aldenml: I see now the reason, `peer_request::length` can't be that long. No problem in these cases since offset is `0` (just looking for the first `peer_request`).

arvidn: is this a problem if the file size is > 2 GiB?

src/lsd.cpp
arvidn: would it make more sense to just use ``std::strtoi()``? or even better ``std::stoi()`` (which will throw an exception if the value overflows)

aldenml: after looking at this, I do think it's better to still use `stratal` and not crash, because this read from data coming from the outside

include/libtorrent/file_storage.hpp
```diff
@@ -150,13 +150,13 @@ namespace libtorrent
...
 		// (i.e. it should be freed in the destructor). If
-		// the len is not name_is_owned, the name pointer doesn not belong
+		// the len is not name_is_owned, the name pointer does not belong
 		// to this object, and it's not 0-terminated
 		std::uint64_t name_len:12;
-		std::uint64_t pad_file:1;
-		std::uint64_t hidden_attribute:1;
-		std::uint64_t executable_attribute:1;
-		std::uint64_t symlink_attribute:1;
+		bool pad_file:1;
```
arvidn: this will enlarge this type. this was previously two bitfield units backed by two uint64_ts. now it's 3, two uint64_ts and one bool. some compilers may still coalesce them, but some might not. it's not guaranteed

src/assert.cpp
```diff
@@ -175,7 +175,7 @@ TORRENT_EXPORT void print_backtrace(char* out, int len, int max_depth
 	std::array<void*, 50> stack;
 
 	STACKFRAME64 stack_frame;
-	memset(&stack_frame, 0, sizeof(stack_frame));
+	std::memset(&stack_frame, 0, sizeof(stack_frame));
```
arvidn: while you're here, would you mind removing the ``memset()`` and say ``STACKFRAME64 stack_frame = {};`` instead?

include/libtorrent/file_storage.hpp
aldenml: interesting

src/assert.cpp
aldenml: sure

src/enum_net.cpp
```diff
@@ -175,7 +173,7 @@ namespace libtorrent { namespace
 
 			if ((nl_hdr->nlmsg_flags & NLM_F_MULTI) == 0) break;
 
-		} while((nl_hdr->nlmsg_seq != seq) || (nl_hdr->nlmsg_pid != pid));
+		} while((int(nl_hdr->nlmsg_seq) != seq) || (int(nl_hdr->nlmsg_pid) != pid));
```
arvidn: it seems like perhaps the actual types of these fields should be the type of the parameters to the function, no? it seems like this may be redundant casting

src/enum_net.cpp
```diff
@@ -650,7 +648,7 @@ namespace libtorrent
 #endif
 
 		SOCKET s = socket(AF_INET, SOCK_DGRAM, 0);
-		if (s == SOCKET_ERROR)
+		if (int(s) == SOCKET_ERROR)
```
arvidn: this is weird.. ``SOCKET`` is a windows type, and ``SOCKET_ERROR`` is a windows value.. it has a different type?

src/enum_net.cpp
```diff
@@ -233,7 +231,7 @@ namespace libtorrent { namespace
 
 		if_indextoname(if_index, rt_info->name);
 		ifreq req;
-		memset(&req, 0, sizeof(req));
+		std::memset(&req, 0, sizeof(req));
```
arvidn: would you mind using ``= {};`` instead of memset here too?

src/enum_net.cpp
aldenml: this types should be signed ints, but in android, they are not :(

src/enum_net.cpp
```diff
@@ -1094,7 +1092,7 @@ namespace libtorrent
 		int seq = 0;
 
 		char msg[BUFSIZE];
```
arvidn: and ``== {};`` here too?

src/file.cpp
```diff
@@ -160,15 +160,15 @@ namespace
...
 		{
 			h += batch_size;
 			num_handles -= batch_size;
-			batch_size = (std::min)(num_handles, MAXIMUM_WAIT_OBJECTS);
-			if (batch_size <= 0) return WAIT_OBJECT_0;
+			batch_size = std::min(num_handles, MAXIMUM_WAIT_OBJECTS);
+			if (batch_size <= 0) return int(WAIT_OBJECT_0);
 		}
-		return WAIT_FAILED;
+		return int(WAIT_FAILED);
```
arvidn: maybe this function should return whatever type this is.. ``DWORD``?

src/file.cpp
```diff
@@ -209,7 +209,7 @@ namespace
 			}
 		}
 
-		if (wait_for_multiple_objects(int(h.size()), h.data()) == WAIT_FAILED)
+		if (wait_for_multiple_objects(int(h.size()), h.data()) == int(WAIT_FAILED))
```
arvidn: then you wouldn't need these casts either

src/enum_net.cpp
aldenml: yes, they are, can't remember if it was in all windows jobs

src/file.cpp
aldenml: yes, I will look the type

src/file_progress.cpp
```diff
@@ -154,18 +154,18 @@ namespace libtorrent { namespace aux
...
 		for (; size > 0; ++file_index)
 		{
-			std::int64_t file_offset = off - fs.file_offset(file_index);
+			std::int64_t const file_offset = off - fs.file_offset(file_index);
 			TORRENT_ASSERT(file_index != fs.num_files());
 			TORRENT_ASSERT(file_offset <= fs.file_size(file_index));
-			int add = (std::min)(fs.file_size(file_index)
-				- file_offset, std::int64_t(size));
-			m_file_progress[file_index] += add;
+			int const add = int(std::min(fs.file_size(file_index)
```
arvidn: if ``add`` was ``std::int64_t``, would it save casts?

aldenml: I think I can do it, I need to tweak one more type

src/file_storage.cpp
```diff
@@ -792,71 +792,73 @@ namespace libtorrent
...
-		return m_files[index].pad_file;
+		return m_files[std::size_t(index)].pad_file;
 	}
 
-	std::int64_t file_storage::file_offset(int index) const
+	std::int64_t file_storage::file_offset(int const index) const
 	{
 		TORRENT_ASSERT_PRECOND(index >= 0 && index < int(m_files.size()));
-		return m_files[index].offset;
+		return std::int64_t(m_files[std::size_t(index)].offset);
```
arvidn: is the offset field unsigned here? it's not clear why you need this cast. would it be better to make the offset field signed?

src/file_storage.cpp
```diff
@@ -865,39 +867,39 @@ namespace libtorrent
 		return m_symlinks[fe.symlink_index];
 	}
 
-	time_t file_storage::mtime(internal_file_entry const& fe) const
+	std::time_t file_storage::mtime(internal_file_entry const& fe) const
 	{
-		int index = &fe - &m_files[0];
+		int const index = int(&fe - m_files.data());
```
arvidn: the this file and the ones below, if you would make ``index`` have type ``std::size_t`` you would save a bunch of casts I believe

src/file_storage.cpp
aldenml: I'm a little afraid of making bitfields types signed (`std::uint64_t offset:48;`), is that OK?

src/file_storage.cpp
aldenml: true, I will need to tweak the preconditions

src/platform_util.cpp
```diff
@@ -73,9 +73,9 @@ namespace libtorrent
 		if (getrlimit(RLIMIT_NOFILE, &rl) == 0)
 		{
 			if (rl.rlim_cur == rlim_infinity)
-				return (std::numeric_limits<int>::max)();
+				return std::numeric_limits<int>::max();
 
-			return rl.rlim_cur;
+			return int(rl.rlim_cur);
```
arvidn: this feels like something that could easily overflow. would you mind making sure that if ``rlim_cur`` > ``INT_MAX``, that ``INT_MAX`` is returned?

src/tracker_manager.cpp
```diff
@@ -241,16 +241,16 @@ namespace libtorrent
...
 		if (i != m_http_conns.end())
 		{
 			m_http_conns.erase(i);
 			return;
 		}
 
-		udp_conns_t::iterator j = std::find_if(m_udp_conns.begin(), m_udp_conns.end()
-			, [c] (udp_conns_t::value_type const& uc) { return uc.second.get() == c; });
+		auto const j = std::find_if(m_udp_conns.begin(), m_udp_conns.end()
+			, [c] (decltype(m_udp_conns)::value_type const& uc) { return uc.second.get() == c; });
```
arvidn: I'm not sure about this use of ``decltype``. what's wrong with the typedef?

src/platform_util.cpp
aldenml: ok

src/tracker_manager.cpp
aldenml: I will use the corresponding type, I don't like to have necessary `typedef`s. You tell me if it's ok.

Any reason against the use of`decltype`?

src/file_storage.cpp
```diff
@@ -921,8 +923,8 @@ namespace libtorrent
 		return fe.offset;
 	}
 
-	file_entry file_storage::at(file_storage::iterator i) const
-	{ return at_deprecated(i - m_files.begin()); }
+	file_entry file_storage::at(file_storage::iterator const i) const
+	{ return at_deprecated(int(i - m_files.begin())); }
```
aldenml: checked it, it's worse

include/libtorrent/file_storage.hpp
```diff
@@ -134,7 +134,7 @@ namespace libtorrent
 		};
 
 		// the offset of this file inside the torrent
-		std::uint64_t offset:48;
+		std::int64_t offset:48;
```
arvidn: I see, I didn't realize this was a bitfield. I agree with you then. this causes the same problem as before, it won't necessarily merge with surrounding bitfields of different types (afaik)

aldenml: ok, back to `std::uint64_t` :)

aldenml: it seems clang knows that 48 bits can't overflow a signed of 64, don't complain without the cast, let's see msvc and gcc

arvidn:
lgtm

src/file_storage.cpp
```diff
@@ -802,7 +802,7 @@ namespace libtorrent
 	std::int64_t file_storage::file_size(int const index) const
 	{
 		TORRENT_ASSERT_PRECOND(index >= 0 && index < int(m_files.size()));
-		return std::int64_t(m_files[std::size_t(index)].size);
+		return m_files[std::size_t(index)].size;
```
arvidn: about all these casts for operator[]. it feels a bit unfortunate. is it msvc that warns on this otherwise? I seem to recall that ``-Wconversion`` doesn't warn on this

aldenml: it's really unfortunate, I'm preparing a warning free branch and this is by far the biggest source of warnings. The problem is the conversion from signed to `std::size`.

aldenml: the other one is the `resize` of containers

arvidn: is it possible to disable the specific warning that causes this? or would that disable too many cases?

aldenml: I didn't find a specific flag for this (it's very particular), the current flag `-Wno-sign-conversion` suppress many more (sometimes suspicions) conversions. I just finished my branch, no way I can send a PR from it, I will send first the headers and later the srcs in groups.

aldenml:
before merge this, I want to add minor changes and remove the added (unnecessary) `std::size_t(` casts

arvidn:
looks good. do you know why those casts in operator[] are not necessary? are there any cases where they are? (in operator[]). if so, do you understand why they are necessary sometimes?

aldenml:
yes, they are not necessary in this patch, because this is only about removing `-Wno-shorten-64-to-32`. The casts in `operator[]` was due to some mishandling of changes in another branch fixing `-Wno-sign-conversion`, I would like to have your opinion in how to handle the case of `-Wno-sign-conversion` and stl containers working with `std::size_t`.

At this moment, I rather split this mega PR in smaller ones:
1- just the fix `-Wno-shorten-64-to-32`
2- `lsd.cpp` port fix
3- minor refactor and formatting
4- adding some consts.

What do you think?

arvidn:
I don't have a strong opinion. in general I definitely think it's easier to review and push smaller patches. I believe I've gone through all of this though. let me know if you want to split it up, it would give me a bit more confidence I didn't miss something

aldenml:
I will split this in smaller PRs, keeping this open only for reference

arvidn:
is this ready to be closed?

aldenml:
sure, no need of this, I can keep track of the pending changes in my own repo (still a few of them).

arvidn:
oh, I thought the last PR with the changes to the Jamfile was the last part

PR #1366, opened 2016-11-29
Clarify documentation

markuskramerIgitt:
The current expression could mislead someone to think that the torrent is distributed (or announced) on Kademlia DHT.

The suggested expression should clarify, that Kademlia DHT, unless the torrent is private, distributes the its interested peers.

### DIFF
--- a/include/libtorrent/create_torrent.hpp
+++ b/include/libtorrent/create_torrent.hpp
@@ -347,7 +347,7 @@ namespace libtorrent
 		// and they have the same name.
 		bool m_multifile:1;
 
-		// this is true if the torrent is private. i.e., is should not
+		// this is true if the torrent is private. i.e., its interested peers should not
 		// be announced on the dht
 		bool m_private:1;
 
--- a/include/libtorrent/torrent_info.hpp
+++ b/include/libtorrent/torrent_info.hpp
@@ -397,7 +397,7 @@ namespace libtorrent
 		// metadata resolved yet or not.
 		bool is_valid() const { return m_files.is_valid(); }
 
-		// returns true if this torrent is private. i.e., it should not be
+		// returns true if this torrent is private. i.e., its interested peers should not be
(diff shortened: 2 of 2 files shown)


include/libtorrent/create_torrent.hpp
```diff
@@ -347,7 +347,7 @@ namespace libtorrent
 		// and they have the same name.
 		bool m_multifile:1;
 
-		// this is true if the torrent is private. i.e., is should not
+		// this is true if the torrent is private. i.e., its interested peers should not
```
arvidn: the typo here is that "is" should have been "it". I think it make more sense to specify whether *this* client will announce to the DHT or not. It doesn't really have control over what other peers do. also, "interested" is used in a specific meaning in the peer protocol, is that what you refer to?

markuskramerIgitt: Thank you, now I have a better understanding and agree: it should be "this client".
And no, I meant simply "other peers" with "interested peers".

Shall I close these pull requests and create new ones with "this client"?

90 lines above in `create_torrent.hpp`, I just found the comment for `set_priv()` is saying it perfectly: "... not advertise itself publicly...,"
Why not write "see comment for set_priv()" ?

markuskramerIgitt: The comment for `set_priv()` does not mention DHT.
I propose to add DHT to the comment: 

Torrents with the private flag set ask the client to not use any other
sources than the tracker for peers, and to not **use DHT to** advertise itself publicly,
**only the** tracker.

I changed "...ask clients..." to "...ask the client..." because "itself" is singular.

arvidn: please, feel free to update this PR. I will squash it at the end

markuskramerIgitt:
I close this PR in order to submit a new one.

PR #1383, opened 2016-12-04
support separate external ip for each interface

ssiloti:
Another step towards multi-home support.

### DIFF
--- a/simulation/test_dht.cpp
+++ b/simulation/test_dht.cpp
@@ -143,7 +143,7 @@ TORRENT_TEST(dht_bootstrap)
 				ses.post_session_stats();
 				std::printf("depth: %d nodes: %d\n", routing_table_depth, num_nodes);
 				TEST_CHECK(routing_table_depth >= 8);
-				TEST_CHECK(num_nodes >= 110);
+				TEST_CHECK(num_nodes >= 50);
 				dht.stop();
 				return true;
 			}
--- a/src/ip_voter.cpp
+++ b/src/ip_voter.cpp
@@ -173,14 +173,28 @@ namespace libtorrent
 		return true;
 	}
 
-	bool external_ip::cast_vote(address const& ip, int source_type, address const& source)
+	external_ip::external_ip(address local4, address global4
+		, address local6, address global6)
 	{
-		return m_vote_group[ip.is_v6()].cast_vote(ip, source_type, source);
+		m_addresses[0][0] = global4;
+		m_addresses[0][1] = global6;
+		m_addresses[1][0] = local4;
+		m_addresses[1][1] = local6;
+
+		if (m_addresses[0][1] == address_v4())
+			m_addresses[0][1] = address_v6();
+		if (m_addresses[1][1] == address_v4())
+			m_addresses[1][1] = address_v6();
+
+		TORRENT_ASSERT(m_addresses[0][0].is_v4());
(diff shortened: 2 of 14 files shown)


simulation/test_dht.cpp
```diff
@@ -143,7 +143,7 @@ TORRENT_TEST(dht_bootstrap)
 				ses.post_session_stats();
 				std::printf("depth: %d nodes: %d\n", routing_table_depth, num_nodes);
 				TEST_CHECK(routing_table_depth >= 8);
-				TEST_CHECK(num_nodes >= 110);
+				TEST_CHECK(num_nodes >= 50);
```
ssiloti: I'm a little nervous about this. As far as I can tell the new code is causing a much smaller routing table at the end of bootstrap because if the local endpoint is a global address we will use that right away to set the node ID rather than doing an initial query with a random ID and changing it when we get the first response.

src/ip_voter.cpp
```diff
@@ -173,14 +173,28 @@ namespace libtorrent
 		return true;
 	}
 
-	bool external_ip::cast_vote(address const& ip, int source_type, address const& source)
+	external_ip::external_ip(address local4, address global4
```
aldenml: any reason not to use `const&` for addresses? (IPv6 are a little heavier I think)

simulation/test_dht.cpp
arvidn: simulations should (ideally) be deterministic

test/test_peer_list.cpp
```diff
@@ -173,7 +173,6 @@ torrent_state init_state(torrent_peer_allocator& allocator
 	st.max_peerlist_size = 1000;
 	st.allow_multiple_connections_per_ip = false;
 	st.peer_allocator = &allocator;
-	st.ip = &ext_ip;
```
arvidn: oh, right. it's just a member now, constructed as part of ``torrent_state``. so the argument to this function should probably be removed then

src/ip_voter.cpp
```diff
@@ -173,14 +173,31 @@ namespace libtorrent
...
+		, address local6, address global6)
 	{
-		return m_vote_group[ip.is_v6()].cast_vote(ip, source_type, source);
+		m_addresses[0][0] = global4;
```
arvidn: would it make sense to aggregate-initialize these?

i.e. ``m_addresses{{global4, global6}, {local4, local6}}`` in the initializer list

src/ip_voter.cpp
```diff
@@ -173,14 +173,31 @@ namespace libtorrent
...
+		m_addresses[0][1] = global6;
+		m_addresses[1][0] = local4;
+		m_addresses[1][1] = local6;
+
+#if TORRENT_USE_IPV6
+		if (m_addresses[0][1] == address_v4())
```
arvidn: I wonder if it would work out nicely to have a function: ``address ensure_v6(address a) { return a == address_v4() ? address_v6() : a; }``

and use that in the aggregate initialization

simulation/test_dht.cpp
arvidn: that makes sense. it may still be a (separate) problem that the bootstrap perhaps focuses too much on depth and not so much on breadth.

arvidn:
looks good

PR #1393, opened 2016-12-07
torrent_status code cleanup

aldenml:


arvidn:
what's the motivation for this change? I consider these two things features (but I'm open to be convinced otherwise):

1. the ``torrent_info`` constructor and destructor are compiled only once, instead of every time the header is included.
2. ``torrent_handle`` supports being used as key in ``boost::unordered_map``

aldenml:
nothing in particular other than trying to remove files, and in this case I was wrong. I didn't pay attention to the actual meaning of putting the ` = default` in the header, and I forgot about `boost` :)

By the same reasoning, what's your opinion of moving
```cpp
torrent_status(torrent_status const&) = default;
torrent_status& operator=(torrent_status const&) = default;
```
to the `.cpp`?

arvidn:
I think that would make sense

PR #1395, opened 2016-12-08
Fix c4996 strcpy

pavel-pimenov:


include/libtorrent/stack_allocator.hpp
```diff
@@ -79,8 +79,9 @@ namespace libtorrent { namespace aux
 		{
 			int const ret = int(m_storage.size());
 			int const len = int(std::strlen(str));
+			TORRENT_ASSERT(len > 0);
```
arvidn: len is allowed to be 0 here, isn't it?

include/libtorrent/stack_allocator.hpp
```diff
@@ -79,8 +79,9 @@ namespace libtorrent { namespace aux
...
 			m_storage.resize(ret + len + 1);
-			std::strcpy(&m_storage[ret], str);
+			std::memcpy(&m_storage[ret], str, len);
```
aldenml: don't this need something like putting `=0` at the end?

arvidn:
yes, in fact, the tests are triggering the ``> 0`` assert in the tests

include/libtorrent/stack_allocator.hpp
pavel-pimenov: void resize (size_type n, value_type val = value_type());
m_storage.resize add 0x00?

pavel-pimenov:
if len == 0 valid parameter
m_storage.resize(ret + len+ 1);  bug?

include/libtorrent/stack_allocator.hpp
aldenml: didn't test, but a quick reading seem to support your idea.

arvidn: I still think it would be a good idea to make it explicit. Partly because it avoids questions like this and I can envision a future where ``std::vector<>`` is replaced by something that doesn't default initialize the memory (as an optimization)

arvidn:
lgtm

aldenml:
do you think the code could be improved a little with the use of `strncpy`? (just a comment)

pavel-pimenov:
char * strncpy(char *dest, const char *src, size_t n)
 {
               size_t i;
               for (i = 0; i < n && **src[i] != '\0'**; i++)
                   dest[i] = src[i];
               for ( ; i < n; i++)
                   dest[i] = '\0';
               return dest;
 }

aldenml:
what do you mean? that `strncpy` is slower or incorrect in this case?

pavel-pimenov:
slower.

aldenml:
I thought that this function is a kind of special function and that highly optimized code is emitted. am I wrong?

ssiloti:
The use case for `strncpy` is when you have a fixed sized buffer which the string's length may exceed. It is not applicable here because the buffer is sized based on the length of the string so it is guaranteed to be large enough.

PR #1396, opened 2016-12-08
don't indicate interest in peers when in upload only mode

ssiloti:
If the torrent is paused we don't want to mark peers as interesting
even if the torrent has not finished downloading.


arvidn:
lgtm

PR #1397, opened 2016-12-09
minor refactor in torrent_status and minor header cleanup in torrent_handle

aldenml:


include/libtorrent/torrent_status.hpp
```diff
@@ -545,7 +546,7 @@ namespace libtorrent
 		bool stop_when_ready = false;
 
 		// the info-hash for this torrent
-		sha1_hash info_hash{nullptr};
```
arvidn: ? I seem to recall that the default constructor leaves the ``sha1_hash`` uninitialized.

aldenml: no :) (https://github.com/arvidn/libtorrent/blob/master/include/libtorrent/sha1_hash.hpp#L76)

arvidn: I see. passing in ``nullptr`` should probably be prohibited then...

arvidn:
lgtm

PR #1404, opened 2016-12-11
refactor to more use of auto c++11 keyword

aldenml:


src/tracker_manager.cpp
```diff
@@ -241,16 +241,17 @@ namespace libtorrent
...
 		if (i != m_http_conns.end())
 		{
 			m_http_conns.erase(i);
 			return;
 		}
 
-		udp_conns_t::iterator j = std::find_if(m_udp_conns.begin(), m_udp_conns.end()
-			, [c] (udp_conns_t::value_type const& uc) { return uc.second.get() == c; });
+		auto const j = std::find_if(m_udp_conns.begin(), m_udp_conns.end()
+			, [c](std::pair<std::uint32_t, std::shared_ptr<udp_tracker_connection>> const& uc)
```
arvidn: I don't think this signature is correct. ``m_udp_conns`` is an unordered_map, so the value type will be ``std::pair<std::uint32_t const, std::shared_ptr<udp_tracker_connection>>``. it will implicitly convert to a non-const *copy* here. This is one of the benefits of referring to the type as ``::value_type``. Perhaps you could pull it out with a using statement:

``using udp_value = typename decltype(m_udp_conns)::value_type;`` (I'm not sure whether you need the ``typename`` or not)

arvidn: at a higher level, it seems a bit suspicious to search an unordered_map like this. I would expect the ``tracker_connection`` object to know its key for an efficient lookup here.

src/tracker_manager.cpp
```diff
@@ -421,28 +422,25 @@ namespace libtorrent
...
 			std::shared_ptr<request_callback> rc = c->requester();
 			if (rc) rc->debug_log("aborting: %s", req.url.c_str());
 #endif
 		}
-		for (udp_conns_t::iterator i = m_udp_conns.begin()
-			, end(m_udp_conns.end()); i != end; ++i)
+		for (auto const& p : m_udp_conns)
 		{
-			std::shared_ptr<udp_tracker_connection> c = i->second;
+			std::shared_ptr<udp_tracker_connection> c = p.second;
```
arvidn: I'm not sure we need to make a copy of ``c`` here

PR #1407, opened 2016-12-11
mark all converstion to bool operators as explicit

ssiloti:


src/kademlia/item.cpp
```diff
@@ -140,10 +140,12 @@ item::item(entry v)
...
 
 item::item(bdecode_node const& v)
-	: m_value(std::move(v))
-	, m_seq(0)
+	: m_seq(0)
 	, m_mutable(false)
-{}
+{
+	// TODO: implement ctor for entry from bdecode_node?
+	m_value = v;
```
arvidn: is ``m_value`` not a ``bdecode_node``? I'm surprised std::move() worked from a ``const&`` previously

ssiloti: m_value is an `entry`. I assume the move fell back to a copy then used the bool conversion on `bdecode_node` to construct an `entry` from integer.

arvidn:
lgtm

PR #1409, opened 2016-12-12
strengthened type safety in handling of piece and file indices

arvidn:


include/libtorrent/add_torrent_params.hpp
```diff
@@ -313,7 +315,7 @@ namespace libtorrent
 		// can be set to control the initial file priorities when adding a
 		// torrent. The semantics are the same as for
 		// ``torrent_handle::prioritize_files()``.
-		std::vector<std::uint8_t> file_priorities;
+		vector<std::uint8_t, file_index_t> file_priorities;
```
arvidn: perhaps I should limit exposure of these strong types to the public API

src/storage.cpp
```diff
@@ -907,17 +917,17 @@ namespace libtorrent
...
+				ec.file(file_index);
 				ec.operation = storage_error::check_resume;
 				return false;
 			}
 
 			// OK, this file existed, good. Now, skip all remaining pieces in
 			// this file. We're just sanity-checking whether the files exist
 			// or not.
-			peer_request const pr = fs.map_file(file_index, 0
-				, int(fs.file_size(file_index) + 1));
```
arvidn: I'm pretty sure this call had offset and length swapped. especially since it would truncate the file size to signed int.

include/libtorrent/add_torrent_params.hpp
aldenml: quite a change... This will greatly impact my `swig+java` binding, because the type will be translated as another class, and instead of passing primitives, I will be creating  and passing objects just for indexing (super slow).

aldenml:
Hi arvidn, I want to make my case to avoid this change based on my opinion. Sometimes type safety is a balance between safety, verbosity, clarity, etc.. and this particular change does not feel well balanced to me. Type safety is very nice and sometimes a must, but I think it should be a threshold in the concept to be abstracted. For example, the `kademlia/types.hpp` was definitely and improvement, especially since `secret_key` and `signature` have the same structure, but represent totally different concepts (very easy to mistakenly interchange them).

Another point is that the impact in automatic binding creation (like swig) is very high, since creating and passing objects for indexes is very costly, everyone would need to replicate the public API with `int`s and internally call the right method, rendering the "automatic" advantage null.

I still think the type indexed vector is a good idea, since it provide range checking invariants and allow for `int` index API.

src/storage.cpp
arvidn: the only output that's used from this function is the piece (ifaict). and the file size passed to it is truncated so that can't be relied on anyway. the output piece is determined by the input file and offset. the offset does support large files.

include/libtorrent/add_torrent_params.hpp
arvidn: I think it's probably best to not expose ``libtorrent::vector<>``, but the index types are trivially castable to integers (and with deprecated-functions enabled, implicitly convertible). There's no way to define converters in swig? i.e. piece_index_t <-> int etc?

arvidn:
I recognize that the cost to clients may be unreasonable by exposing too much of this through the public API. although, for C++ clients, I'm quite certain that this is the trend things are moving towards.

aldenml:
I think you gave me an idea to alleviate the problem with your comment 

> piece_index_t <-> int 

If you think this trend is the best way moving forward, well...let's accommodate the bindings :)

include/libtorrent/add_torrent_params.hpp
```diff
@@ -431,22 +433,22 @@ namespace libtorrent
 		// this is a map of partially downloaded piece. The key is the piece index
 		// and the value is a bitfield where each bit represents a 16 kiB block.
 		// A set bit means we have that block.
-		std::map<int, bitfield> unfinished_pieces;
+		std::map<piece_index_t, bitfield> unfinished_pieces;
```
aldenml: I don't think this particular case will be easy to fix in swig with a typemap

aldenml: below, in `renamed_files`, missing `file_index_t`?

include/libtorrent/aux_/vector.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+#include <vector>
+
+#include "libtorrent/assert.hpp"
+
+namespace libtorrent
+{
+
+	template <typename T, typename IndexType>
+	struct vector : std::vector<T>
```
aldenml: If you think of this as a special type of map relationship: `IndexType -> T` not sure if it feels more natural to have the `IndexType` template argument first.

Can I still index the vector with `size_t`?

include/libtorrent/block_cache.hpp
```diff
@@ -337,7 +332,7 @@ namespace libtorrent
 		struct hash_value
 		{
 			std::size_t operator()(cached_piece_entry const& p) const
-			{ return std::size_t(p.storage.get()) + std::size_t(p.piece); }
+			{ return std::size_t(p.storage.get()) + static_cast<int>(p.piece); }
```
aldenml: pretty sure this generate a double conversion, first to `signed` for the `+` and then to `unsigned` at return

src/bt_peer_connection.cpp
```diff
@@ -2176,7 +2178,7 @@ namespace libtorrent
 			std::memset(ptr, 0, packet_size - 5);
 			piece_picker const& p = t->picker();
 			int mask = 0x80;
-			for (int i = 0; i < num_pieces; ++i)
+			for (piece_index_t i(0); i != piece_index_t(num_pieces); ++i)
```
aldenml: not a problem in this particular case, but the pattern in general. Since `i` is not constant, it could be possible to increment `i` inside the loop, and since now the condition is `!=`, it will fail to stop.

aldenml:
do you have a link where I can read more about this "trend", maybe I can find some criticism that can persuade you of doing this. Sorry, I just don't see it as a good solution. It remains me of: why stl containers receive `size_t` for indices? I would say that a lot (a lot) of programers are against that.

include/libtorrent/aux_/vector.hpp
arvidn: no, this will shadow any ``operator[]`` from the base class

include/libtorrent/block_cache.hpp
arvidn: it won't actually cost anything though. ``p.piece`` is already an ``int``. it would presumably already be 64 bits wide in the register, or sign-extended when loaded from memory.

arvidn: oh, I see, the cast to ``std::size_t`` should probably be moved to encompass the ``+`` expression

arvidn:
the fact that stl containers use unsigned types a lot is a bit of a historical mistake, but it can't easily be undone at this point.

as references for the C++ trend toward stronger types, I think ``std::chrono`` is a great example. the code C++ guidelines is probably the ones best representing the community:

https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Ri-typed

Along with (afaict) the wider recognition of the dangers of unsigned types and how the arithmetic on them are unintuitive and rarely what you want. Disabling arithmetic on types that don't define sensible arithmetic is increasingly popular. enum classes is an example of this trend, classic enums can implicitly convert to int and arithmetic can be performed on it. Perhaps a more subtle example is the wider recognition of array to pointer decay is dangerous and promotion of ``std::array`` to fix it. C-arrays can decay to a pointer and arithmetic can be performed on it. ``std::byte`` as voted into C++17 is another example of disabling arithmetic for types where it does not make sense.

I'm curious what your objections are. The two types I converted in this patch, piece indices and file indices never refer to the same value domain. mixing them is almost certainly a bug. In addition, there are very few cases of performing actual arithmetic on these indices (other than increment and decrement for loops). It does not make sense to allow arithmetic on them.

arvidn:
speaking of clarity (and relevant to the bug I discovered as part of this). look at this signature and avoid passing in the wrong things:

``peer_request map_file(int, std::int64_t, int);``

:) That's fundamentally the problem. I think a good rule of thumb is that any signature should be obvious without having the parameter names (just the types)

aldenml:
your rule of thumb is indeed very good, is a common problem for a lot of frameworks and languages, some of them alleviate the problem with named parameters and others (like java) are out of luck (unless costly boxing is used).

I remember very well the line in the bug because it confused me when I added the `int` cast, but I didn't checked for the surrounding algorithm at the time.

My objection is mainly stylistic, because certainly the idea helps in the correctness and expressiveness of algorithms. But OK, lets say I'm starting to like it, I have two questions:

1 - Is "block index" another potential type?
2 - Can we make the loops more robust to avoid issues like the one I mentioned above?

aldenml:
What do you think will be the solution for `aux_/vector.hpp` and the public API? I noticed that even inside "aux_" it's not in the `aux` namespace.

arvidn:
yeah, I realized I forgot to add the ``aux`` namespace too. I should fix that. As for exposing libtorrent::aux::vector, I'm increasingly thinking that it's a bad idea. One of the benefits of using standard containers/types is easy interoperability. I will hide the ugly down-casts in the libtorrent API edge.

the loops is one of the things I'm the least happy with right now. In fact, there are at least two different conventions I use in this patch. I'm starting to lean towards ``piece_index_t i(0)`` for the start condition, rather than the ``begin_index()`` or ``begin_piece()``. using ``<`` is fine, the ``strong_typedef`` type supports all the comparison operators. I think ``block_index`` definitely is a candidate. I have maintained a list of other useful types I've come across while working on this patch. For instance, the index into the ``priority_boundaries`` vector in piece_picker, and the ``slot_index`` in part_file.

aldenml:
good, looking forward to help you here (at least with more reviews)

src/stat_cache.cpp
```diff
@@ -107,8 +104,10 @@ namespace libtorrent
 
 	void stat_cache::clear()
 	{
-		std::vector<stat_cache_t>().swap(m_stat_cache);
-		std::vector<error_code>().swap(m_errors);
+		m_stat_cache.clear();
+		m_stat_cache.shrink_to_fit();
+		m_errors.clear();
+		m_errors.shrink_to_fit();
```
pavel-pimenov: void aux::vector::truncate() 
{
  clear();
  shrink_to_fit();
}

m_stat_cache.truncate();
m_errors.truncate();

?

include/libtorrent/add_torrent_params.hpp
arvidn: yes, will fix

arvidn:
I rebased this to make travis and appveyor work. I probably shouldn't have squashed it though, so I ruined reviewing deltas. but I think this is in a good state now (assuming it builds and passes CI)

include/libtorrent/block_cache.hpp
```diff
@@ -216,7 +211,7 @@ namespace libtorrent
 		//TODO: make this 32 bits and to count seconds since the block cache was created
 		time_point expire = min_time();
 
-		std::uint64_t piece:22;
+		piece_index_t piece;
```
aldenml: no problem here with the bitfield `:22`?

include/libtorrent/block_cache.hpp
```diff
@@ -337,7 +332,7 @@ namespace libtorrent
 		struct hash_value
 		{
 			std::size_t operator()(cached_piece_entry const& p) const
-			{ return std::size_t(p.storage.get()) + std::size_t(p.piece); }
+			{ return reinterpret_cast<std::size_t>(p.storage.get()) + static_cast<int>(p.piece); }
```
aldenml: didn't you mention something about casting the entire expression?

include/libtorrent/disk_io_job.hpp
```diff
@@ -206,7 +208,10 @@ namespace libtorrent
 
 		// arguments used for read and write
 		// the piece this job applies to
-		std::uint32_t piece:24;
+		union {
```
aldenml: how this impact the size of the struct?

include/libtorrent/piece_block_progress.hpp
```diff
@@ -39,13 +39,13 @@ namespace libtorrent
 {
 	struct piece_block_progress
 	{
-		constexpr static int invalid_index = -1;
+		constexpr static const piece_index_t invalid_index{-1};
```
aldenml: I assume `const` here is redundant.

include/libtorrent/torrent_handle.hpp
```diff
@@ -944,11 +946,11 @@ namespace libtorrent
 		// marks the piece with the given index as filtered
 		// it will not be downloaded
 		TORRENT_DEPRECATED
-		void filter_piece(int index, bool filter) const;
+		void filter_piece(piece_index_t index, bool filter) const;
```
aldenml: if this is deprecated API, is it ok to change the type here?

src/disk_io_thread.cpp
```diff
@@ -361,8 +362,9 @@ namespace libtorrent
 		}
 
 		// piece range
-		int range_start = int((p->piece / cont_pieces) * cont_pieces);
-		int range_end = (std::min)(range_start + cont_pieces, p->storage->files()->num_pieces());
+		piece_index_t const range_start((static_cast<int>(p->piece) / cont_pieces) * cont_pieces);
+		piece_index_t const range_end(std::min(static_cast<int>(range_start)
+			+ cont_pieces, p->storage->files()->num_pieces()));
```
aldenml: question: when to use `num_pieces()` vs `end_piece()`?

src/file_storage.cpp
```diff
@@ -790,52 +790,52 @@ namespace libtorrent
...
-		TORRENT_ASSERT_PRECOND(index >= 0 && index < int(m_files.size()));
+		TORRENT_ASSERT_PRECOND(index < end_file());
 		return m_files[index].pad_file;
 	}
 
-	std::int64_t file_storage::file_offset(int index) const
+	std::int64_t file_storage::file_offset(file_index_t const index) const
 	{
-		TORRENT_ASSERT_PRECOND(index >= 0 && index < int(m_files.size()));
+		TORRENT_ASSERT_PRECOND(index >= file_index_t(0) && index < end_file());
```
aldenml: why keep `index >= file_index_t(0)` in some cases and not in others?

src/peer_connection.cpp
```diff
@@ -634,25 +638,20 @@ namespace libtorrent
...
-		{
-			if (*i < m_num_pieces)
-			{
-				++i;
-				continue;
-			}
-			i = m_allowed_fast.erase(i);
-		}
+		m_allowed_fast.erase(std::remove_if(m_allowed_fast.begin(), m_allowed_fast.end()
+				, [=](piece_index_t const p) { return p >= limit; })
```
aldenml: indent off?

src/peer_connection.cpp
```diff
@@ -713,7 +712,7 @@ namespace libtorrent
 			TORRENT_ASSERT(m_have_piece.size() == t->torrent_file().num_pieces());
 			t->peer_has(m_have_piece, this);
 			bool interesting = false;
-			for (int i = 0; i < m_have_piece.size(); ++i)
+			for (piece_index_t i(0); i != m_have_piece.end_index(); ++i)
```
aldenml: I bet there is a reason you change from `<` to `!=`, what is it?

src/torrent.cpp
```diff
@@ -2285,26 +2292,26 @@ namespace libtorrent
...
-				std::uint64_t file_size = st.file_size(error.file);
-				int last = st.map_file(error.file, file_size, 0).piece;
+				std::uint64_t file_size = st.file_size(error.file());
+				piece_index_t last = st.map_file(error.file(), file_size, 0).piece;
 				if (m_checking_piece < last)
 				{
-					int diff = last - m_checking_piece;
-					m_num_checked_pieces += diff;
-					m_checking_piece += diff;
+					int diff = static_cast<int>(last) - static_cast<int>(m_checking_piece);
```
aldenml: maybe `std::distance` can be specialized

aldenml:
if only the loops can be coded/expressed in a more consistent way. What do you think about creating an `aux::range` to be able to put a `for (auto : `.

It seems that `strong_typedef` is very `integral` oriented, in the sense that `UnderlyingType` should be only integral. Do you think that `enable_if` is necessary?

src/disk_io_thread.cpp
```diff
@@ -691,7 +696,7 @@ namespace libtorrent
 
 #if DEBUG_DISK_THREAD
 		DLOG("iovec_flushed: piece: %d block_offset: %d [ "
-			, int(pe->piece), block_offset);
+			, pe->piece.value(), block_offset);
```
arvidn: I see, this line doesn't get built normally, that's why this mistake went unnoticed

src/disk_io_thread.cpp
arvidn: I use ``end_piece()`` as the one-past-end piece index for loops. It probably makes sense to use here too, but this is one of few places where I do actual arithmetic on piece indices, so it's a bit of an exception since it needs to be converted to ``int`` anyway.

arvidn: actually, this expression would become more complex to use ``end_piece()``, since the first argument to ``min()`` would have to be explicitly converted to a ``piece_index_t`` otherwise

include/libtorrent/torrent_handle.hpp
arvidn: I was debating this. The reason I think it's fine is because when deprecated functions are enabled, ``piece_index_t`` is also implicitly convertible from and to ``int``, so it shouldn't make a significant difference.

include/libtorrent/block_cache.hpp
arvidn: well, this type cannot be made part of the bitfield, so this will widen the type

include/libtorrent/block_cache.hpp
arvidn: yeah, I was wrong. pointer arithmetic is not the right thing here. we want to convert the pointer to an integer and then add to that, not add ``sizeof(storage_interface)``

include/libtorrent/piece_block_progress.hpp
arvidn: yeah, probably

src/file_storage.cpp
arvidn: I think I was thinking that the asserts in ``operator[]`` would take care of it anyway, and then I realized these are ``PRECOND`` asserts, which are slightly different. I should restore them

arvidn: although, in the future I envision the check to be in the constructor of ``file_index_t``, to make sure it can never be negative

src/peer_connection.cpp
arvidn: not important. I meant to change these back

arvidn:
yes, it would probably make sense to constrain ``strong_typedef`` to integrals.

src/torrent.cpp
arvidn: I'm punting on enriching the types here. I'm thinking that perhaps there could be an implied "distance" unit, which is the result of subtraction and can be used in addition.

include/libtorrent/units.hpp
```diff
@@ -0,0 +1,137 @@
...
+	struct piece_index_tag {};
+	struct file_index_tag {};
+
+	template <typename T, typename Tag>
+	std::string to_string(strong_typedef<T, Tag> const t)
+	{ return std::to_string(static_cast<T>(t)); }
+
+	template <typename T, typename Tag>
+	strong_typedef<T, Tag> next(strong_typedef<T, Tag> v)
+	{ ++v; return v;}
```
arvidn: heh, yes. but ``++v`` didn't exist when I wrote this function

PR #1419, opened 2016-12-15
create separate function for reading a node from a list

ssiloti:
I'm going to be adding a third caller to this with the multi-home DHT changes I'm working on. It'll be used to read local IP/node id pairs from dht_state.


aldenml:
what's your opinion on putting this function in `libtorrrent` root? maybe it can be used in other parts

ssiloti:
I don't think there are any uses for this outside of the DHT code.

PR #1421, opened 2016-12-15
fix upnp

ajax16384:
fix upnp xml_parser
fix listen sockets remap

### DIFF
--- a/include/libtorrent/upnp.hpp
+++ b/include/libtorrent/upnp.hpp
@@ -120,8 +120,29 @@ struct parse_state
 	}
 };
 
+struct error_code_parse_state
+{
+	error_code_parse_state(): in_error_code(false), exit(false), error_code(-1) {}
+	bool in_error_code;
+	bool exit;
+	int error_code;
+};
+
+struct ip_address_parse_state: public error_code_parse_state
+{
+	ip_address_parse_state(): in_ip_address(false) {}
+	bool in_ip_address;
+	std::string ip_address;
+};
+
 TORRENT_EXTRA_EXPORT void find_control_url(int type, char const* string
 	, int str_len, parse_state& state);
+
+TORRENT_EXTRA_EXPORT void find_error_code(int type, char const* string
+	, int str_len, error_code_parse_state& state);
+
+TORRENT_EXTRA_EXPORT void find_ip_address(int type, char const* string
+	, int str_len, ip_address_parse_state& state);
 
 // TODO: support using the windows API for UPnP operations as well
 struct TORRENT_EXTRA_EXPORT upnp final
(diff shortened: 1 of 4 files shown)


arvidn:
would you mind adding a test case that would fail without this fix too? (see test/test_xml.cpp)
Do you think this issue exists in RC_1_1 as well?

ajax16384:
ok, a will try to add test/test_xml.cpp with next commit to this pull.

RC_1_1 contains only upnp xmp_parse error (wrong tag strcmp), remap should works well. should i made separate pull request to RC_1_1 ?

arvidn:
> should works well. should i made separate pull request to RC_1_1 ?

that would be great, thanks!

src/upnp.cpp
```diff
@@ -1115,55 +1115,40 @@ void upnp::disable(error_code const& ec)
...
 	}
+}
 
-	struct ip_address_parse_state: public error_code_parse_state
-	{
-		ip_address_parse_state(): in_ip_address(false) {}
-		bool in_ip_address;
-		std::string ip_address;
-	};
+TORRENT_EXTRA_EXPORT void find_ip_address(int type, char const* string, int str_len, ip_address_parse_state& state)
```
arvidn: I believe ``TORRENT_EXTRA_EXPORT`` is redundant here, and in fact adds the risk of a conflict with the header file

include/libtorrent/upnp.hpp
```diff
@@ -120,9 +120,30 @@ struct parse_state
...
 
+struct error_code_parse_state
+{
+	error_code_parse_state(): in_error_code(false), exit(false), error_code(-1) {}
+	bool in_error_code;
+	bool exit;
+	int error_code;
+};
+
+struct ip_address_parse_state: public error_code_parse_state
```
arvidn: the ``public`` here is redundant. for ``struct``, the default inheritance is default.

include/libtorrent/upnp.hpp
```diff
@@ -120,9 +120,30 @@ struct parse_state
...
+{
+	ip_address_parse_state(): in_ip_address(false) {}
```
arvidn: (or ``bool in_ip_address = false;``)

src/upnp.cpp
```diff
@@ -1115,55 +1115,40 @@ void upnp::disable(error_code const& ec)
 	m_socket.close();
 }
 
-namespace
+TORRENT_EXTRA_EXPORT void find_error_code(int type, char const* string, int str_len, error_code_parse_state& state)
```
arvidn: I don't think you should have the ``TORRENT_EXTRA_EXPORT`` here either

ajax16384:
just after this pull request will be squashed to master, i will add another pull request to RC_1_1 with similar changes

arvidn:
lgtm. thanks!

PR #1422, opened 2016-12-15
submit disk jobs from read_piece right away

ssiloti:
Without this the jobs get submitted on the next tick which may take a while.


arvidn:
nice catch!

Issue #1405, opened 2016-12-11
[Feature Request] Larger and less webseed requests over multiple pieces


Falcosc:
libtorrent version (or branch): head 3.3
platform/architecture: linux

Hi,
by using webseed feature per http url-list entry I see that libtorrent creates as many request as torrent has pieces. It's not so nice to create 1000's of http requests, especially because http requests creating a lot more load on a http server compared to 1000's bittorrent requests, it’s even worse if there are multiple http servers involved to handle my requests. For that reason it would be very nice to request more than 1 piece per http request like GetRight http://getright.com/seedtorrent.html
Unfortunately GetRight is not usable on a linux environemt without X. That’s the reason why I do use libtorrent.
By connecting to my own http server I did test the current behavior and got for each piece a new request in access log.

### Full Target:
The best solution would be an open http stream of the largest possible range of pieces and collecting pieces direct during stream processing to get the first pieces available before the full http response is done.
### Smaller Target:
I know that piece processing during stream receive does not fit into the current implementation style. If it is easier to implement it would be ok to have the piece processing after the response is done. But than we would need a max range, maybe 1/10 of the torrent size to get the first pieces available after 10% is downloaded.
If needed it would also ok to force the urlseed_pipeline_size to 1 to skip this logic if this makes it easier to implement larger http request ranges.
### Why this could help to improve liborrent:
An advantage of implementing a sequential Webseed processing with piece handling during response stream handling would be the preparation for FTP Webseed support. But don’t talk about FTP support, that’s something for the future. 

I hope you could understand what I would like to see. If not I would like to answer your questions.

Kind Regards
Falco

arvidn:
the default of requesting 1 MiB at a time is hardcoded here:

https://github.com/arvidn/libtorrent/blob/master/src/web_peer_connection.cpp#L84

If you want larger requests, you can try increasing that constant. A patch to make it configurable would be welcome!

Falcosc:
Thank you for your fast response.
I'm ok with changing the hardcoded value.

By such a small max size I guess that pieces are verified after the whole response is back, isn't it?

Example a 1GiB file with 1000 pieces
If I set 100MiB, and my download process is at 180MiB, how many pieces are available for the peers? 100 or 179?

If everthing is working fine I will tell you. 
[...]

PR #1423, opened 2016-12-15
add setting urlseed_max_request_bytes to handle large web seed requests #1405

Falcosc:
Pull for #1405 at RC_1_1


Falcosc:
Hi,

I would like to have a default of 16MiB because if anybody want to use webseed, they would do it with large file anyway. We don't have to care about slow speed of small files based on missing parallel connections. In many cases 1 MiB is actually slower than 16MiB because some webservers with concurrent connection limits have request delays as well which slows down the average webseed user.

But our tests can't handle 16MiB. I guess some of them are based on webseed.

Would it be possible to have a look on them?
time-critical fail with a SIGSEGV
https://travis-ci.org/arvidn/libtorrent/jobs/184390101#L1202

And the test_storage.cpp.rename_file_fastresume  fails because of an incomplete file
https://travis-ci.org/arvidn/libtorrent/jobs/184390106#L1350

Falcosc:
@arvidn I would like to help you to get 16MiB as default value. I would identify all tests which are rely on small short running tasks and extend them by setting 1MiB web seed request limit.
But I have trouble at executing the tests on my own host.

Falcosc:
Ok, it was just an temporary CI error, everthing works fine.

examples/client_test.cpp
```diff
@@ -1558,6 +1558,8 @@ int main(int argc, char* argv[])
 		, listen_port);
 	settings.set_str(settings_pack::listen_interfaces, iface_str);
 
+	settings.set_int(settings_pack::urlseed_max_request_bytes, 32 * 1024 * 1024);
```
arvidn: I think it would be better to set the default to something sensible, than to set it explicitly here. in master, client_test provides generic access to all settings via the command line too

include/libtorrent/session_settings.hpp
```diff
@@ -152,6 +152,12 @@ namespace libtorrent
 		// 5.
 		int urlseed_pipeline_size;
 
+		// the maximum request range of an url seed in bytes. This value defines
+		// the largest possible sequentiel web seed request. Default is 16777216
+		// (16 * 1024 * 1024). Lower values are possible but will be ignored if
+		// if they are lower then piece size.
+		int urlseed_max_request_bytes;
```
arvidn: this structure, ``session_settings`` is deprecated. please don't add new fields to it. instead, in the settings_pack definition, use ``SET_NOPREV`` (is in there's no previous field)

src/settings_pack.cpp
```diff
@@ -229,6 +229,7 @@ namespace libtorrent
 		SET(peer_timeout, 120, 0),
 		SET(urlseed_timeout, 20, 0),
 		SET(urlseed_pipeline_size, 5, 0),
+		SET(urlseed_max_request_bytes, 16 * 1024 * 1024, 0),
```
arvidn: you can use the ``SET_NOPREV`` macro here instead

src/web_peer_connection.cpp
```diff
@@ -86,16 +86,20 @@ web_peer_connection::web_peer_connection(peer_connection_args const& pack
...
-	// request even larger blocks at a time
-	if (!web.supports_keepalive) preferred_size *= 4;
-
-	prefer_contiguous_blocks((std::max)(preferred_size / tor->block_size(), 1));
+	// request at least 4MiB
+	if (!web.supports_keepalive)
+		min_size *= 4;
+	// we want to have at least piece size to prevent block based requests
+	min_size = (std::max)(tor->torrent_file().piece_length(), min_size);
+	preferred_size = (std::max)(preferred_size, min_size);
```
arvidn: wrapping ``std::min`` in parenthesis is a way to work around windows.h introducing a ``min`` and ``max`` macro, which interferes. It appears that this is not a problem within libtorrent translation units, but should still be used in headers (since those are exposed to client code). We've been transitioning to just use the plain ``std::min(...)`` call in .cpp files. so you don't need it here

examples/client_test.cpp
Falcosc: This was just my test to check that setting overwrite do work.

src/web_peer_connection.cpp
```diff
@@ -86,16 +86,20 @@ web_peer_connection::web_peer_connection(peer_connection_args const& pack
 	shared_ptr<torrent> tor = pack.tor.lock();
 	TORRENT_ASSERT(tor);
 
-	// we always prefer downloading 1 MiB chunks
-	// from web seeds, or whole pieces if pieces
-	// are larger than a MiB
-	int preferred_size = 1024 * 1024;
+	// we prefer downloading large chunks from web seeds,
+	// but still want to be able to split requests
+	int preferred_size = m_settings.get_int(settings_pack::urlseed_max_request_bytes);
```
arvidn: I think ``preferred_size`` should be const here, which means you have to initialize it with a ``std::max()`` expression and calculate ``min_size`` before getting the settings value. In fact, I think ``min_size`` should also be const and initialized only once, ``int const min_size = std::max((web.supports_keepalive ? 1 : 4) * 1024 * 1024, tor->torrent_file().piece_length());``

src/web_peer_connection.cpp
```diff
@@ -86,16 +86,16 @@ web_peer_connection::web_peer_connection(peer_connection_args const& pack
 	shared_ptr<torrent> tor = pack.tor.lock();
 	TORRENT_ASSERT(tor);
 
-	// we always prefer downloading 1 MiB chunks
-	// from web seeds, or whole pieces if pieces
-	// are larger than a MiB
-	int preferred_size = 1024 * 1024;
+	// if the web server is known not to support keep-alive. request 4MiB
+	// but we want to have at least piece size to prevent block based requests
+	int const min_size = std::max((web.supports_keepalive ? 1 : 4) * 1024 * 1024,
```
arvidn: sorry for not thinking this all the way through in my earlier comments, but having these constants here (1 MiB and 4 MiB) seems a bit like a smell, now that it's configurable. The thing that makes sense in my mind is the logic that uses larger contiguous blocks if the web server does not support keep alive.

The only true lower limit here is 1 (which used to be expressed explicitly before this patch). I think that was a bit clearer than now, where it's expressed indirectly by making the piece_size a minimum request size, so when divided by the block size, it's never less than 1.

What would you think about something like this?

```
int const preferred_size = (web.supports_keepalive ? 1 : 2)
    * m_settings.get_int(settings_pack::urlseed_max_request_bytes);
prefer_contiguous_blocks(std::max(preferred_size / tor->block_size(), 1));
```
?

and would it make sense to call the setting ``settings_pack::urlseed_request_bytes``? or ``urlseed_request_size``? the word "max" to me implies that something else determines the size, and this is just a cap.

Falcosc: Actually something else is determining the size. If we pick the rarest piece for a BT peer we would get a gap on the next http request.

And I don't want to multiply max request bytes for non keepalive because if you have already a large size you will get a huge new size which would lock to much pieces of the file and speed for multi connection users will be reduced because picker has no pieces left to pick.

The point is, what is the different between 20s per request and 40s per request? It would not matter because one connection each 20s does already not hurt anybody.

But real aggressive user which selects something like 1MiB or less should be forced to use larger sizes because such user would give the bittorrent client a bad reputation.

arvidn:
you need to update the python binding also, to not bind the member from the deprecated struct

Falcosc:
I did extend settings description with our discussion content.

src/web_peer_connection.cpp
Falcosc: The correct solution would be request size based on current download rate. But I guess it's not worth it to implement such a complex solution.

include/libtorrent/settings_pack.hpp
```diff
@@ -767,10 +767,19 @@ namespace libtorrent
 			// low number, like 5
 			urlseed_pipeline_size,
 
-			// the maximum request range of an url seed in bytes. This value
+			// The maximum request range of an url seed in bytes. This value
 			// defines the largest possible sequential web seed request. Default
 			// is 16 * 1024 * 1024. Lower values are possible but will be ignored
 			// if they are lower then piece size.
+			// This value should be related to your download speed to prevent
+			// libtorrent from creating to many expensive http requests per
```
arvidn: "to" should be "too"

include/libtorrent/settings_pack.hpp
```diff
@@ -767,10 +767,19 @@ namespace libtorrent
...
+			// second. You can select a value as high as you want but keep in mind
+			// that libtorrent can't create parallel requests if the first request
+			// did already select the whole file.
+			// If you combine bittorrent seeds with web seeds and pick strategies
+			// like rarest first you may find your web seed requests splitted into
```
arvidn: "splitted" -> "split"

PR #1424, opened 2016-12-15
add setting urlseed_max_request_bytes to handle large web seed requests #1405

Falcosc:
Pull for #1405 at master

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* add setting urlseed_max_request_bytes to handle large web seed requests
 	* make sequence number for mutable DHT items backed by std::int64_t
 	* tweaked storage_interface to have stronger type safety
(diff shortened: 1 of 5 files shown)


arvidn:
is this the same as the other PR for this ticket?

arvidn:
oh, I see. the other one is against RC_1_1

Falcosc:
I hope it was correct to create 2 pull requests to get both branches because I did not find multible pull requests from other user.

arvidn:
if a fix applies to both RC_1_1 and master, just putting it in RC_1_1 is enough, since I periodically merge RC_1_1 into master. However, if the branches have diverged significantly, it may be best fixed with two separate patches

Falcosc:
Ok, I will close it.

PR #1425, opened 2016-12-16
fixed more warnings of shorten-64-to-32 and sign-compare in android

aldenml:


src/enum_net.cpp
```diff
@@ -158,13 +158,20 @@ namespace libtorrent { namespace
...
 		do
 		{
-			int read_len = recv(sock, buf, bufsize - msg_len, 0);
+			int read_len = int(recv(sock, buf, bufsize - msg_len, 0));
 			if (read_len < 0) return -1;
 
 			nl_hdr = reinterpret_cast<nlmsghdr*>(buf);
 
+#if (TORRENT_ANDROID && defined __clang__)
+#pragma clang diagnostic push
```
arvidn: would you mind adding a comment here why the sign-compare needs to be disabled? I realize I haven't commented in similar situations, but I think if this is becoming a more established pattern, for more platforms, it might be a good idea

src/create_torrent.cpp
```diff
@@ -149,11 +149,11 @@ namespace libtorrent
 					&& (flags & create_torrent::symlinks))
 				{
 					std::string sym_path = get_symlink_path(f);
-					fs.add_file(l, 0, file_flags, s.mtime, sym_path);
+					fs.add_file(l, 0, file_flags, std::time_t(s.mtime), sym_path);
```
arvidn: would it make sense to change the ``mtime``, ``atime`` and ``ctime`` members to be ``std::time_t``?
The risk I can see is that some old ABIs may still use a 32 bit ``mtime``. There's still some time out though until that may become a problem.

arvidn:
lgtm, see comment

src/create_torrent.cpp
aldenml: ok

src/enum_net.cpp
aldenml: actually I now realize that this is a clang warning and it's correct since the struct types of `nlmsghdr` are unsigned. We don't see the warning in macOS because there is no `TORRENT_USE_NETLINK`.

arvidn:
lgtm

PR #1426, opened 2016-12-16
refactor in aux::write_ to avoid explicit castings

aldenml:


include/libtorrent/aux_/io.hpp
```diff
@@ -59,12 +59,16 @@ namespace libtorrent { namespace aux
 		return ret;
 	}
 
-	template <class T, class Byte>
+	template <class T, class In, class Byte, typename Cond
+		= typename std::enable_if<std::is_integral<In>::value
```
arvidn: and ``T`` should probably be integral too

aldenml: `sizeof(In) == 1`? The type `In` is the one we don't want to cast, the one that is restricted already in the `enable_if`.

I will add the condition to `T`.

include/libtorrent/aux_/io.hpp
```diff
@@ -59,12 +59,16 @@ namespace libtorrent { namespace aux
 		return ret;
 	}
 
-	template <class T, class Byte>
+	template <class T, class In, class Byte, typename Cond
+		= typename std::enable_if<std::is_integral<T>::value &&
+		(std::is_integral<In>::value || std::is_enum<In>::value)>::type>
 	inline typename std::enable_if<sizeof(Byte)==1, void>::type
-	write_impl(T val, span<Byte>& view)
+	write_impl(In data, span<Byte>& view)
```
arvidn: I meant to say ``sizeof(Byte)`` must be 1. The code in here assumes that it has 8 bits specifically.

arvidn:
I guess ``Byte`` wasn't introduced in this patch. but would you mind adding another constraint?

include/libtorrent/aux_/io.hpp
aldenml: I think that constraint is already there with the return type

aldenml:
I don't mind at all, but see my comment above , i don't want to introduce redundant conditions

include/libtorrent/aux_/io.hpp
arvidn: ah, I see

arvidn:
lgtm

PR #1430, opened 2016-12-18
minor code corrections

ajax16384:
minor code corrections

### DIFF
--- a/src/enum_net.cpp
+++ b/src/enum_net.cpp
@@ -385,6 +385,44 @@ namespace libtorrent
 		return false;
 	}
 
+#ifdef TORRENT_WINDOWS
+	HMODULE GetIPHelperHandle()
+	{
+		static bool handle_checked = false;
+		static HMODULE handle = 0;
+		if (!handle_checked) 
+		{
+			handle = LoadLibraryA("Iphlpapi.dll");
+			handle_checked = true;
+		}
+		return handle;
+	}
+
+	template <typename T>
+	bool GetIPHelperProc(LPCSTR name, T& proc_out)
+	{
+		static T proc = nullptr;
+
+		static bool failed_proc = false;
+
+		if ((proc == nullptr) && !failed_proc)
+		{
+			HMODULE iphlp = GetIPHelperHandle();
+			if (iphlp)
+			{
+				proc = (T)GetProcAddress(iphlp, name);
+			}
+			else
+			{
+				failed_proc = true;
+			}
+		}
+		proc_out = proc;
+
+		return proc_out != nullptr;
+	}	
+#endif
+
 #if TORRENT_USE_GETIPFORWARDTABLE
 	address build_netmask(int bits, int family)
 	{
@@ -590,22 +628,12 @@ namespace libtorrent
 #elif TORRENT_USE_GETADAPTERSADDRESSES
 
 #if _WIN32_WINNT >= 0x0501
-		// Load Iphlpapi library
-		HMODULE iphlp = LoadLibraryA("Iphlpapi.dll");
-		if (iphlp)
+		typedef ULONG (WINAPI *GetAdaptersAddresses_t)(ULONG,ULONG,PVOID,PIP_ADAPTER_ADDRESSES,PULONG);
+		GetAdaptersAddresses_t GetAdaptersAddresses;
+
+		// Get GetAdaptersAddresses() pointer
+		if (GetIPHelperProc("GetAdaptersAddresses", GetAdaptersAddresses))
 		{
-			// Get GetAdaptersAddresses() pointer
-			typedef ULONG (WINAPI *GetAdaptersAddresses_t)(ULONG,ULONG,PVOID,PIP_ADAPTER_ADDRESSES,PULONG);
-			GetAdaptersAddresses_t GetAdaptersAddresses = (GetAdaptersAddresses_t)GetProcAddress(
-				iphlp, "GetAdaptersAddresses");
-
-			if (GetAdaptersAddresses == nullptr)
-			{
-				FreeLibrary(iphlp);
-				ec = error_code(boost::system::errc::not_supported, generic_category());
-				return std::vector<ip_interface>();
-			}
-
 			ULONG buf_size = 10000;
 			std::vector<char> buffer(buf_size);
 			PIP_ADAPTER_ADDRESSES adapter_addresses
@@ -622,7 +650,6 @@ namespace libtorrent
 			}
 			if (res != NO_ERROR)
 			{
-				FreeLibrary(iphlp);
 				ec = error_code(WSAGetLastError(), system_category());
 				return std::vector<ip_interface>();
 			}
@@ -645,8 +672,6 @@ namespace libtorrent
 				}
 			}
 
-			// Free memory
-			FreeLibrary(iphlp);
 			return ret;
 		}
 #endif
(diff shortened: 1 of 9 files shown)


src/hasher512.cpp
```diff
@@ -160,9 +160,9 @@ namespace libtorrent
 #endif
 		}
 #elif defined TORRENT_USE_LIBCRYPTO
-		SHA512_Update(&m_context, reinterpret_cast<unsigned char const*>(data.data()), data.size());
+		SHA512_Update(&m_context, reinterpret_cast<unsigned char const*>(data.data()), std::uint32_t(data.size()));
```
aldenml: I think here it's better `std::size_t`

src/utp_stream.cpp
```diff
@@ -713,7 +713,7 @@ bool utp_incoming_packet(utp_socket_impl* s
 	, udp::endpoint const& ep, time_point receive_time)
 {
 	return s->incoming_packet(
-		span<std::uint8_t const>(reinterpret_cast<std::uint8_t const*>(p.data()), int(p.size()))
+		span<std::uint8_t const>(reinterpret_cast<std::uint8_t const*>(p.data()), p.size())
```
aldenml: What do you think about `{reinterpret_cast<std::uint8_t const*>(p.data()), p.size()}` and make this one line

src/utp_socket_manager.cpp
```diff
@@ -161,7 +161,7 @@ namespace libtorrent
 	{
 //		UTP_LOGV("incoming packet size:%d\n", size);
 
-		if (p.size() < int(sizeof(utp_header))) return false;
+		if (p.size() < sizeof(utp_header)) return false;
```
aldenml: I don't understand why this does not triggered the warning `sign-compare` before, hopefully with the new typed indexes will be more easy to enable `sign-conversion`

src/file_storage.cpp
```diff
@@ -564,7 +564,7 @@ namespace libtorrent
 		else
 		{
 			if (m_files.empty())
-				m_name = split_path(path).c_str();
+				m_name = split_path(path);
```
aldenml: I think this is the root of the problems in the CI, since internally there is a `ret.append(1, '\0');` that seems to make a difference in the internal char*.

arvidn: split path will return a string with '\0' characters embedded in it. this change will copy the entire string, rather than just the first element. is there a need to copy the entire string?
(and I admit that this is a pretty shitty function interface)

src/enum_net.cpp
```diff
@@ -980,6 +980,7 @@ namespace libtorrent
 		GetIfEntry_t GetIfEntry = (GetIfEntry_t)GetProcAddress(iphlp, "GetIfEntry");
 		if (!GetIfEntry)
 		{
+			FreeLibrary(iphlp);
```
arvidn: hm.. I was under the impression that the library was only loaded once, into a static variable. I guess that's not the case, but that also means this library is loaded for every call here, which doesn't seem very good

arvidn: basically, I think the proper solution here would be to load the library once, into a static variable and then not load it again. the same should probably be done for the actual functions we pull out of the library too.
If we need to free the library, it should really be done as a scope_guard or handle, that calls ``FreeLibrary()`` and the end of the scope

src/hasher512.cpp
ajax16384: libTorrent SHA512_Update & SHA1_update routines use std::uint32_t as length arguments. You think it's better to transfer them to std::size_t ?

src/utp_stream.cpp
ajax16384: ok, much better

src/enum_net.cpp
ajax16384: ok

src/hasher512.cpp
aldenml: hmmm...the signature from openssl is
`int SHA512_Update(SHA512_CTX *c, const void *data, size_t len);`
what version of openssl are you using?

ajax16384: i meant these functions:
https://github.com/arvidn/libtorrent/blob/master/include/libtorrent/sha512.hpp#L19
https://github.com/arvidn/libtorrent/blob/master/include/libtorrent/sha1.hpp#L32

aldenml: well, I was concerned only about the call to openssl ones :)
I don't know if changing the signature for the internal libtorrent sha* is good or not

arvidn:
src/enum_net.cpp:393:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]

arvidn:
the underlying issue in master has been resolved now. please rebase this to get the tests to pass

src/enum_net.cpp
```diff
@@ -377,6 +377,38 @@ namespace libtorrent
 		return false;
 	}
 
+#ifdef TORRENT_WINDOWS
+	HMODULE GetIPHelperHandle()
+	{
+		static bool handle_checked = false;
+		static HMODULE handle = 0;
+		if (!handle_checked)
```
arvidn: wouldn't it be sufficient to checkt ``if (handle)`` here? surely an ``HMODULE`` of 0 cannot be valid

arvidn: i.e. ``handle_checked`` seems redundant

src/enum_net.cpp
```diff
@@ -377,6 +377,38 @@ namespace libtorrent
...
+		{
+			handle = LoadLibraryA("Iphlpapi.dll");
+			handle_checked = true;
+		}
+		return handle;
+	}
+
+	template <typename T>
+	bool GetIPHelperProc(LPCSTR name, T& proc_out)
```
arvidn: I have this feeling that it might be nicer to return the function pointer, rather than taking an out-parameter. The type wouldn't be deducible anymore, but maybe that wouldn't be that big of a deal. it could be used like this:

auto GetAdaptersAddresses = get_ip_helper_proc<GetAdaptersAddresses_t>("GetAdapterAddresses");

if it fails, it could return ``nullptr``

arvidn: also, I would prefer sticking to snake_case naming.

src/enum_net.cpp
arvidn: I guess the point is to cache the failure, is it? if we fail to load the library, we don't want to retry every time it's called

ajax16384: yes, it was made to cache failed loadlibrary result

src/hasher512.cpp
arvidn: does this call ``SHA512_Update`` sometimes refer to the openssl function and sometimes to the internal one? if so, the internal one really should have a matching signature

arvidn:
yes, that's what I had in mind for loading dlls and functions. lgtm. I think eventually putting those functions in a ``win_utils.hpp`` or something like that would make sense (but it's very windows specific, so I'm not sure it should be in a generic platform_utils.hpp)

ajax16384:
ok, i will try to make separate pull request after this, which will move windows dynamic proc load to common file.

src/hasher512.cpp
arvidn: @ajax16384 as far as I can tell, this change doesn't look right. the ``#ifdef`` implies that this is in fact a call into libcrypto, which appears to be taking a ``long int``. It looks like it's the call below that end up calling the internal sha512 function, which you point out takes a ``uint32_t``.

However, I think a better solution would be to make the internal SHA512 signature match the one from libcrytpo. But in any case, I don't see how *this* call is affected.

ajax16384: ok, i will revert std::uint32_t cast from this pull request and will make separate pull request which will make internal hash declaration match libcrypt api.

arvidn:
lgtm

PR #1440, opened 2016-12-22
fixed Wrange-loop-analysis warning with Apple clang-800.0.42.1

aldenml:


arvidn:
it seems quite unnecessary to take a reference to an ``int32``.. it is a ``piece_index_t``, right? I suspect there's something wrong with this warning.

aldenml:
yes, it looks strange to me, that's the reason I made the PR just with this, maybe something is missing in the type that hint the compiler it's not necessary.

aldenml:
the warning is happening with the latest android NDK too (clang)

src/torrent.cpp
```diff
@@ -1810,7 +1810,7 @@ namespace libtorrent
 				have_pieces.push_back(p.index);
 			}
 
-			for (auto const i : have_pieces)
```
arvidn: the patch against clang that added this warning suggests that removing ``const`` here would also silence the warning. I think that's preferable. It's unfortunate, though, that it seems to discriminate primitive and other types.

aldenml: interesting...well, removed `const&` and warning is gone

arvidn:
lgtm

PR #1442, opened 2016-12-23
added explicit time_t cast to avoid warnings in android 32 bits

aldenml:


arvidn:
seems reasonable. after reading about where mingw is with this (and I suppose some android distros) I'm starting to think that perhaps ``time_t`` should be avoided, and converted to and from ``time_point`` at the edges.

PR #1445, opened 2016-12-23
added size and resize methods to aux::vector

aldenml:
OK, this is mainly for your consideration. I find this a good way to express that the vector can't be expanded beyond the natural limit of the index and in the process, a lot of `sign-conversion` warnings should be eliminated.

### DIFF
--- a/include/libtorrent/piece_picker.hpp
+++ b/include/libtorrent/piece_picker.hpp
@@ -576,12 +576,12 @@ namespace libtorrent
 			std::set<const torrent_peer*> have_peers;
 #endif
 
-			enum : std::uint32_t
+			enum : std::int32_t
 			{
 				// index is set to this to indicate that we have the
 				// piece. There is no entry for the piece in the
 				// buckets if this is the case.
-				we_have_index = 0xffffffff,
+				we_have_index = 0x7fffffff,
 				// the priority value that means the piece is filtered
 				filter_priority = 0,
 				// the max number the peer count can hold
--- a/include/libtorrent/aux_/vector.hpp
+++ b/include/libtorrent/aux_/vector.hpp
@@ -52,21 +52,38 @@ namespace libtorrent { namespace aux {
 		{
 			TORRENT_ASSERT(idx >= IndexType(0));
 			TORRENT_ASSERT(idx < end_index());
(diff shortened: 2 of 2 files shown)


arvidn:
seems reasonable. note that those ``resize()`` functions will shadow the original ones. which I suppose may be a feature.

aldenml:
yes, I consider that a feature. I think I should also add `size()`. It can't be greater than the index limit.

arvidn:
it's not entirely obvious to me that this removes more ``static_cast``s than it causes. or maybe it doesn't cause any ``new static_cast``s

include/libtorrent/piece_picker.hpp
```diff
@@ -576,12 +576,12 @@ namespace libtorrent
...
 #endif
 
-			enum : std::uint32_t
+			enum : std::int32_t
 			{
 				// index is set to this to indicate that we have the
 				// piece. There is no entry for the piece in the
 				// buckets if this is the case.
-				we_have_index = 0xffffffff,
+				we_have_index = 0x7fffffff,
```
arvidn: maybe it would make sense to make this ``-1``

aldenml:
well, reading your comment, now I think that if I just need a better "size", I can use `end_index()`. Also, I would like to review in detail `we_have_index`, this change was triggered by
```
TORRENT_ASSERT(m_piece_map.size() < piece_pos::we_have_index);
```
and having it `-1` will always assert.

Let me revert the PR just to the `resize`, this alone improve the situation in static casts because these solves a lot of `sign-conversion` warnings (still silenced in `Jamfile`).

aldenml:
Ok, i will close this, now I have a few of doubts and I don't want to introduce noise. What I'm looking for is the resolution of `sign-conversion` warnings.

I think taking a look at `we_have_index` is needed.

arvidn:
the ``we_have_index`` was mostly useful when the ``piece`` field had fewer than 32 bits, and was unsigned. All assertions against ``we_have_index`` should probably be turned into either ``==`` or ``!=``. the ``<`` check is most likely a left over from when it was unsigned.

arvidn:
I'm actually working on a patch right now that will take care of the specific issue of ``we_have_index``

PR #1446, opened 2016-12-23
Log refactoring

pavel-pimenov:


arvidn:
`../src/kademlia/traversal_algorithm.cpp:357:28: error: format ‘%s’ expects argument of type ‘char*’, but argument 5 has type ‘const void*’
`

Since there's no test for the logging, would it make sense to run one of the dht sims (under simulations) with and without the patch and diff the output?

aldenml:
I see more this functions like internal to the compilation unit (inside an anonymous namespace). Not sure what's the @arvidn's preference here.

arvidn:
well, I think anything that *can* be a free function should be one. In this case though, the purpose of these log functions is to basically dump the state of the object, so they are kind of inherently member functions. If they were made free functions (which would be a prerequisite for internal linkage) a lot of member variables would have to be passed in as arguments, instead of just pulling them out of ``this`` as is done in this patch.

I think this patch looks reasonable. One thing that might be worth thinking about is if there's any way to be able to unit test any of these functions. I think it's difficult the way the DHT code is structured right now.

src/kademlia/routing_table.cpp
```diff
@@ -1198,4 +1184,18 @@ bool routing_table::is_full(int const bucket) const
 		&& int(i->replacements.size()) >= m_bucket_size);
 }
 
+#ifndef TORRENT_DISABLE_LOGGING
+void routing_table::log_node_failed(node_id const& nid, bucket_t::iterator const j) const
```
arvidn: I think it would be better to pass in a ``node_entry const&`` instead of an iterator here

arvidn:
lgtm (but see comment)

Issue #1435, opened 2016-12-21
bugreport qBittorrent crashed in macOS


okeatime:
libtorrent branch: RC_1_1 30303c17dfce4baae84758dd2fc9c93d9165da62
platform/architecture: macOS
compiler and compiler version: Apple LLVM version 8.0.0 (clang-800.0.42.1)

qBittorrent with debug version libtorrent crashed right after launch.

I built libtorrent with:
`./configure --disable-dependency-tracking --disable-silent-rules --enable-encryption --prefix=/usr/local --with-boost=/usr/local/opt/boost CXXFLAGS=-std=c++11 --enable-debug`

```
Using a disk cache size of 0 MiB
assertion failed. Please file a bugreport at https://github.com/arvidn/libtorrent/issues
Please include the following information:

version: 1.1.1.0-1229491

file: 'session_impl.cpp'
line: 7099
function: void libtorrent::aux::session_impl::check_invariant() const
expression: m_stats_counters[counters::num_unchoke_slots] == (std::numeric_limits<int>::max)()

stack:
1: assert_fail(char const*, int, char const*, char const*, char const*, int)
2: libtorrent::aux::session_impl::check_invariant() const
3: void libtorrent::invariant_access::check_invariant<libtorrent::aux::session_impl>(libtorrent::aux::session_impl const&)
4: void libtorrent::check_invariant<libtorrent::aux::session_impl>(libtorrent::aux::session_impl const&)
5: libtorrent::invariant_checker_impl<libtorrent::aux::session_impl>::invariant_checker_impl(libtorrent::aux::session_impl const&)
6: libtorrent::invariant_checker_impl<libtorrent::aux::session_impl>::invariant_checker_impl(libtorrent::aux::session_impl const&)
7: libtorrent::invariant_checker_impl<libtorrent::aux::session_impl> libtorrent::make_invariant_checker<libtorrent::aux::session_impl>(libtorrent::aux::session_impl const&)
8: libtorrent::aux::session_impl::update_listen_interfaces()
9: libtorrent::apply_pack(libtorrent::settings_pack const*, libtorrent::aux::session_settings&, libtorrent::aux::session_impl*)
10: libtorrent::aux::session_impl::apply_settings_pack_impl(libtorrent::settings_pack const&)
11: libtorrent::aux::session_impl::apply_settings_pack(boost::shared_ptr<libtorrent::settings_pack>)
[...]

PR #1449, opened 2016-12-23
fix session invariant check fail for difficult setting dependencies

Falcosc:
fix for #1435

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -1712,7 +1712,7 @@ namespace aux {
 #endif
 		TORRENT_ASSERT(is_single_thread());
 		boost::shared_ptr<settings_pack> p = load_pack_from_struct(m_settings, s);
-		apply_settings_pack_impl(p);
+		apply_settings_pack_impl(*p);
 	}
 
 	libtorrent::session_settings session_impl::deprecated_settings() const
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -1655,6 +1655,9 @@ namespace aux {
 	// session_impl is responsible for deleting 'pack'
 	void session_impl::apply_settings_pack(boost::shared_ptr<settings_pack> pack)
 	{
+#if TORRENT_USE_INVARIANT_CHECKS
+		check_invariant();
+#endif
 		apply_settings_pack_impl(*pack);
 	}
 
@@ -1694,10 +1697,6 @@ namespace aux {
 
 		m_disk_thread.set_settings(&pack, m_alerts);
 
-#if TORRENT_USE_INVARIANT_CHECKS
-		check_invariant();
-#endif
-
 		if (reopen_listen_port)
 		{
 			error_code ec;
@@ -1708,9 +1707,12 @@ namespace aux {
 #ifndef TORRENT_NO_DEPRECATE
 	void session_impl::set_settings(libtorrent::session_settings const& s)
 	{
+#if TORRENT_USE_INVARIANT_CHECKS
+		check_invariant();
+#endif
 		TORRENT_ASSERT(is_single_thread());
 		boost::shared_ptr<settings_pack> p = load_pack_from_struct(m_settings, s);
-		apply_settings_pack(p);
+		apply_settings_pack_impl(p);
 	}
 
 	libtorrent::session_settings session_impl::deprecated_settings() const
@@ -6198,6 +6200,9 @@ retry:
 
 	void session_impl::set_local_download_rate_limit(int bytes_per_second)
 	{
+#if TORRENT_USE_INVARIANT_CHECKS
+		check_invariant();
+#endif
 		settings_pack p;
 		p.set_int(settings_pack::local_download_rate_limit, bytes_per_second);
 		apply_settings_pack_impl(p);
@@ -6205,6 +6210,9 @@ retry:
 
 	void session_impl::set_local_upload_rate_limit(int bytes_per_second)
 	{
+#if TORRENT_USE_INVARIANT_CHECKS
+		check_invariant();
+#endif
 		settings_pack p;
 		p.set_int(settings_pack::local_upload_rate_limit, bytes_per_second);
 		apply_settings_pack_impl(p);
@@ -6212,6 +6220,9 @@ retry:
 
 	void session_impl::set_download_rate_limit(int bytes_per_second)
 	{
+#if TORRENT_USE_INVARIANT_CHECKS
+		check_invariant();
+#endif
 		settings_pack p;
 		p.set_int(settings_pack::download_rate_limit, bytes_per_second);
 		apply_settings_pack_impl(p);
@@ -6219,6 +6230,9 @@ retry:
 
 	void session_impl::set_upload_rate_limit(int bytes_per_second)
 	{
+#if TORRENT_USE_INVARIANT_CHECKS
+		check_invariant();
+#endif
 		settings_pack p;
 		p.set_int(settings_pack::upload_rate_limit, bytes_per_second);
 		apply_settings_pack_impl(p);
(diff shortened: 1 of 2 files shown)


arvidn:
this is a pretty difficult problem in general. the balance between having strict invariants vs. loose invariants that "eventually" are maintained, and checking often vs. checking strictly at the edges of the API.

In general, I believe most of issues like this one can be solved by limiting the invariant checks to just be at the API edge. One problem arise from public functions being called by other functions, while the invariant may be violated. Is there a solution to this scenario involving removing some invariant checks instead? (while still preserving coverage of all external calls into the session).

It would be really nice to have a test that exposes this problem as well.

In this case, can ``apply_settings()`` be called from users directly? or is it always only called from other ``session_impl`` function? If the latter is true, perhaps ``apply_settings()`` should not have an invariant check in it.

Falcosc:
You meen I should put this deeper into the functioncall tree? My first try was to put this to the settings pack where the updatefuction vector is iterated. But this was a little bit to much overhead for this unimportand problem because I don't want to implement a public skip method. Would you like to have it  at settings_pack.apply_pack?

Testing is a good point. Even if we ignore that I'm not able to write a c++ test for it I can't even reproduce it with a client (I have still problems to execute the testsuite). I did try to test it with the client_test but I wasn't able to reproduce it. I try to copy the initialisation order from qbittorrent but didn't find the causing call order. Then I test it with qbittorrent to verify that this would work at least.

arvidn:
I would prefer a solution that doesn't disable the invariant checks and then enables them again. Ideally the fix would instead be to *only* enforce/check invariants where they matter, i.e. at the start (and end) of *public* member functions. The ``CHECK_INVARIANT`` macro will take care of running the check at the exit of the function as well.

The important distinction is that internal functions probably should *not* contain that call at all. In this case, I believe none of the ``session_impl::update_*`` functions should check the invariant (which appears to be the place where the assert you're trying to fix was triggered).

There may be situations where a public function is also called from another public function, which might make it a bit trickier. Probably in such cases, the function should be split into a public and private part, and only the public one checks the invariant.

Falcosc:
Now I get you want to tell me

Would like to remote the check invariant calls in session_impl::update_*, too. But I don't know which is a public interface and which isn't.
The idea would be to remove the calls at session_impl::update_* and add them on all public interfaces.

But I do worry that somebody is able to call the updates directly through an interface method which I did forget to amend with the check invariant call.
Could you tell me how to find all functions which are part of the public interface? Then I would be able to analyse all call stacks and move the invariant checks to a position where they are only called once per interface call.

arvidn:
I was about to say "all public functions", but ``session_impl`` isn't quite that clean, and it exposes a lot of internals. I think a good start would be to *just* remove the invariant checks from the update functions. Those are only supposed to be called by ``apply_settings`` and are definitely internal.

Also, anything that's called from ``session_handle`` is public.

Falcosc:
perfect, I will do it.

src/session_impl.cpp
```diff
@@ -3271,8 +3276,8 @@ namespace aux {
 		if (!m_paused) m_auto_manage_time_scaler--;
 		if (m_auto_manage_time_scaler < 0)
 		{
-			INVARIANT_CHECK;
 			m_auto_manage_time_scaler = settings().get_int(settings_pack::auto_manage_interval);
+			//invariant check is already part of recalculate_auto_managed_torrents
```
arvidn: no need to leave this comment here

src/session_impl.cpp
```diff
@@ -1685,6 +1689,8 @@ namespace aux {
 					!= m_settings.get_str(settings_pack::listen_interfaces));
 
 		apply_pack(&pack, m_settings, this);
+		INVARIANT_CHECK;
```
arvidn: I think it would be better to place this at the very top of the function. This macro will cause the invariant check to be run twice, once immediately and once at the end of the scope (in this case when the function returns)

src/session_impl.cpp
```diff
@@ -640,6 +640,10 @@ namespace aux {
 			update_listen_interfaces();
 			open_listen_port();
 		}
+		
+		//call invariant check again after we did repeat the update calls
+		INVARIANT_CHECK;
```
arvidn: this will cause the invariant check to be run twice. it would be better to call the function directly (check_invariant()) but it would have to be wrapped in ``#ifdef`` to only be called when invariant checks are enabled in the build

src/session_impl.cpp
```diff
@@ -1685,8 +1691,13 @@ namespace aux {
 					!= m_settings.get_str(settings_pack::listen_interfaces));
 
 		apply_pack(&pack, m_settings, this);
+
 		m_disk_thread.set_settings(&pack, m_alerts);
 
+#if TORRENT_USE_INVARIANT_CHECKS
+		check_invariant();
```
arvidn: it's not obvious to me why this should be called here. this function is an internal function (it ends with ``_impl``), so I would expect whatever calls this to enforce the invariant.

src/session_impl.cpp
```diff
@@ -1697,7 +1708,6 @@ namespace aux {
 #ifndef TORRENT_NO_DEPRECATE
 	void session_impl::set_settings(libtorrent::session_settings const& s)
 	{
-		INVARIANT_CHECK;
```
arvidn: I believe this is an external function, isn't it? I think this is accessible via ``session_handle`` and can be called directly by users.

Falcosc: `apply_settings_pack(p);` contains this check already

arvidn: I see

src/session_impl.cpp
Falcosc: some session_handle methods call the apply settings method. And many session imply set methods call the impl.

Do you want to have this call removed and added to the 10 caller functions?

Falcosc: I don't know which settings impl methods are available at the interface. I did put the checks to the deepest method call.

if apply_settings_pack is not visible too, then I would put it outside of it.

src/session_impl.cpp
Falcosc: We could make the impl method check free as you did suggest.
Then we could call apply_settings_pack_impl method instead of apply_settings_pack to keep the stille "invarient check marko at set method start)

src/session_impl.cpp
```diff
@@ -1655,6 +1655,9 @@ namespace aux {
 	// session_impl is responsible for deleting 'pack'
 	void session_impl::apply_settings_pack(boost::shared_ptr<settings_pack> pack)
 	{
+#if TORRENT_USE_INVARIANT_CHECKS
+		check_invariant();
+#endif
 		apply_settings_pack_impl(*pack);
```
Falcosc: check invariant before, or after or before and after with the macro?

arvidn: you need to use the macro here, otherwise the invariant check is not checked at the end of the function

src/session_impl.cpp
```diff
@@ -6198,41 +6200,59 @@ namespace aux {
 
 	void session_impl::set_local_download_rate_limit(int bytes_per_second)
 	{
+#if TORRENT_USE_INVARIANT_CHECKS
```
arvidn: all these should use the macro, to check the invariant at the end of the function too

arvidn: the invariant of a class is the valid states. the invariant is supposed to be maintained at all times, except while executing member functions. a (public) member function may break the invariant temporarily, to re-establish it before it returns. The invariant check must be run at the entry of a function to make sure nothing unexpected has happened to cause the object to have an invalid state, and at the end it's checked to make sure the function correctly re-established the invariant.

src/session_impl.cpp
```diff
@@ -558,6 +558,7 @@ namespace aux {
 
 	void session_impl::init(boost::shared_ptr<settings_pack> pack)
 	{
+		INVARIANT_CHECK;
```
arvidn: ``init()`` is not a public function, but it should (most likely) establish the invariant at the end of it. I'm a little bit surprised this works..

arvidn:
lgtm

PR #1453, opened 2016-12-24
introduce a prio_index_t type

arvidn:
for indexing into the priority sorted piece list, m_pieces, in piece_picker

### DIFF
--- a/include/libtorrent/piece_picker.hpp
+++ b/include/libtorrent/piece_picker.hpp
@@ -579,10 +579,19 @@ namespace libtorrent
 			std::set<const torrent_peer*> have_peers;
 #endif
 
+#ifdef _MSC_VER
+#pragma warning(push, 1)
+#pragma warning(disable : 4268)
+#endif
+
 			// index is set to this to indicate that we have the
 			// piece. There is no entry for the piece in the
 			// buckets if this is the case.
 			constexpr static prio_index_t we_have_index{-1};
+
+#ifdef _MSC_VER
+#pragma warning(pop)
+#endif
 
 			enum : std::uint32_t
 			{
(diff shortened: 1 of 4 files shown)


include/libtorrent/piece_picker.hpp
```diff
@@ -569,19 +572,29 @@ namespace libtorrent
...
+			prio_index_t index;
 
 #ifdef TORRENT_DEBUG_REFCOUNTS
 			// all the peers that have this piece
 			std::set<const torrent_peer*> have_peers;
 #endif
 
+#ifdef _MSC_VER
+#pragma warning(push, 1)
+#pragma warning(disable : 4268)
```
arvidn: as far as I can tell, all of these warnings are false positives and should probably be disabled at the project level

arvidn:
@aldenml not sure if you'd like a chance to take a look. this is more of the same, but just contained to the piece picker internals

include/libtorrent/piece_picker.hpp
aldenml: Probably, this is more of an informational(not problematic ) warning, right?

src/piece_picker.cpp
```diff
@@ -1429,58 +1424,72 @@ namespace libtorrent
...
 
-		int start = 0;
-		for (int b : m_priority_boundaries)
+		prio_index_t start(0);
+		for (auto b : m_priority_boundaries)
 		{
 			if (start == b) continue;
-			aux::random_shuffle(&m_pieces[0] + start, &m_pieces[0] + b);
+			auto r = range(m_pieces, start, b);
+			aux::random_shuffle(r.begin(), r.end());
```
aldenml: Nice

aldenml:
It looks very good to me.

sorry for the delay, I'm only with my phone for a few days :)

PR #1464, opened 2016-12-27
move LoadLibrary logic to new win_util module

ajax16384:
use get_library_procedure as common windows dyn proc load routine

### DIFF
--- a/include/libtorrent/win_util.hpp
+++ b/include/libtorrent/win_util.hpp
@@ -1,0 +1,84 @@
+/*
+
+Copyright (c) 2016, Arvid Norberg
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+    * Redistributions of source code must retain the above copyright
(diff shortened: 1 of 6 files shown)


include/libtorrent/win_util.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+	struct kernel32 {
+		static constexpr char const* library_name = "kernel32.dll";
+	};
+
+	struct advapi32 {
+		static constexpr char const* library_name = "advapi32.dll";
+	};
+
+} // namespace libtorrent
```
arvidn: sorry. would you mind moving this file to include/libtorrent/aux_/win_util.hpp? and also put the functions in namespace libtorrent::aux

arvidn:
please also add the header file to include/libtorrent/Makefile.am
otherwise, lgtm

include/libtorrent/win_util.hpp
ajax16384: ok

arvidn:
lgtm

PR #1467, opened 2016-12-28
Missing member init

pavel-pimenov:


include/libtorrent/aux_/session_impl.hpp
```diff
@@ -123,6 +123,7 @@ namespace libtorrent
 		listen_socket_t()
 			: tcp_external_port(0)
 			, udp_external_port(0)
+			, original_port(0)
```
arvidn: I think it's a lot better to initialize these below, where they are declared

include/libtorrent/piece_block.hpp
```diff
@@ -48,7 +48,7 @@ namespace libtorrent
 		{
 		}
 		piece_index_t piece_index;
-		int block_index;
+		int block_index = 0;
```
arvidn: these were kind of deliberately left uninitialized. but, if this field should be initialized, surely ``piece_index`` should too. ``piece_index{0}``

include/libtorrent/storage.hpp
```diff
@@ -286,7 +286,7 @@ namespace libtorrent
 		, boost::noncopyable
 	{
 		// hidden
-		storage_interface(): m_settings(0) {}
+		storage_interface() {}
```
arvidn: you don't actually need this here now

arvidn:
looks reasonable. see comments though

include/libtorrent/piece_block.hpp
pavel-pimenov: piece_index_t piece_index = {0};
VC++2015 error:
1>q:\vc15\r5xx\libtorrent\include\libtorrent\piece_block.hpp(50): error C2664: 'libtorrent::aux::strong_typedef<int32_t,libtorrent::aux::piece_index_tag,void>::strong_typedef(const libtorrent::aux::strong_typedef<int32_t,libtorrent::aux::piece_index_tag,void> &)': cannot convert argument 1 from 'int' to 'const libtorrent::aux::strong_typedef<int32_t,libtorrent::aux::piece_index_tag,void> &'
1>  q:\vc15\r5xx\libtorrent\include\libtorrent\piece_block.hpp(50): note: Reason: cannot convert from 'int' to 'const libtorrent::aux::strong_typedef<int32_t,libtorrent::aux::piece_index_tag,void>'
1>  q:\vc15\r5xx\libtorrent\include\libtorrent\piece_block.hpp(50): note: Constructor for struct 'libtorrent::aux::strong_typedef<int32_t,libtorrent::aux::piece_index_tag,void>' is declared 'explicit'

help?
or revert int block_index = 0 ?

arvidn: there's no ``=``. just ``piece_index{0};``

include/libtorrent/storage.hpp
pavel-pimenov: remove storage_interface() {} ?

arvidn: I think it would make sense to remove this. it's unnecessary now. (please remove the // hidden comment too)

include/libtorrent/piece_block.hpp
```diff
@@ -41,14 +41,14 @@ namespace libtorrent
 	{
 		static const piece_block invalid;
 
-		piece_block() : piece_index(0) {}
+		piece_block() {}
```
arvidn: I suppose you still need this constructor here, since the default constructor won't be automatically generated if there is another constructor. but you can say ``= default`` instead of ``{}`` which may be nicer.

include/libtorrent/storage.hpp
```diff
@@ -285,8 +285,7 @@ namespace libtorrent
 		, public storage_piece_set
 		, boost::noncopyable
 	{
-		// hidden
-		storage_interface(): m_settings(0) {}
+		storage_interface() {}
```
arvidn: did you mean to leave this in? If you keep this here, you need to keep the comment too

arvidn:
lgtm

arvidn:
I'm not sure why the dht sim is failing. is this rebased on top of my patch that adds 2000 dht nodes to that test?

arvidn:
I wouldn't expect this to make it more likely to pass. In the long run, I'm hoping to make all the sims completely deterministic. right now the random number generator is not reset in between tests, so the order in which the tests are run will give different results.

PR #1468, opened 2016-12-28
Refactoring routing_table

pavel-pimenov:
- Optimistaion ip_set::count -> ip_set::exists
 - ip_set::count add const
 - Fix copy-paste routing_table::remove_node

### DIFF
--- a/src/kademlia/routing_table.cpp
+++ b/src/kademlia/routing_table.cpp
@@ -84,7 +84,7 @@ void ip_set::insert(address const& addr)
 		m_ip4s.insert(addr.to_v4().to_bytes());
 }
 
-size_t ip_set::count(address const& addr)
+size_t ip_set::count(address const& addr) const
 {
 #if TORRENT_USE_IPV6
 	if (addr.is_v6())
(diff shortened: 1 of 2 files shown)


pavel-pimenov:
Remove unused size_t ip_set::count(address const& addr) const; ?

src/kademlia/routing_table.cpp
```diff
@@ -387,30 +397,26 @@ void routing_table::fill_from_replacements(table_t::iterator bucket)
...
+		&& n < &b[0] + b.size())
 	{
-		std::ptrdiff_t idx = n - &bucket->replacements[0];
-		TORRENT_ASSERT(m_ips.count(n->addr()) > 0);
+		std::ptrdiff_t const idx = n - &b[0];
+		TORRENT_ASSERT(m_ips.exists(n->addr()));
 		m_ips.erase(n->addr());
-		bucket->replacements.erase(bucket->replacements.begin() + idx);
+		b.erase(b.begin() + idx);
+	}
```
arvidn: is the indentation correct here?

arvidn:
yeah, if ``count()`` is not used, is should probably be removed

PR #1471, opened 2016-12-29
decrease cryptoapi aux::random_bytes() cpu usage

ajax16384:
Every Windows aux::random_bytes calls spent too much time at CryptAcquireContext
````c++
 - libtorrent::aux::random_bytes	2.92 %	0.00 %	300	0
   - CryptAcquireContextW	2.88 %	0.00 %	296	0	cryptsp.dll
   + CryptReleaseContext	0.04 %	0.00 %	4	0	cryptsp.dll
````

### DIFF
--- a/src/random.cpp
+++ b/src/random.cpp
@@ -51,6 +51,34 @@ extern "C" {
 
 #include "libtorrent/aux_/disable_warnings_pop.hpp"
 
+#if TORRENT_USE_CRYPTOAPI
+namespace
+{
+	HCRYPTPROV make_crypt_provider()
+	{
+		using namespace libtorrent;
+
+		HCRYPTPROV ret;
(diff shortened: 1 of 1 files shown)


arvidn:
lgtm. please rebase this on top of master. I believe I just fixed the merge issue causing the tests to fail

arvidn:
hm. it appears I still haven't fixed this. I'll try again tomorrow

ajax16384:
ok. what do you think of moving all make_crypt_provider() to include\libtorrent\aux_\win_util.hpp ?

arvidn:
as far as I can tell, the crypto provider code doesn't use the dll loading code, so maybe it would make more sense to put it in its own header. maybe ``win_cryptoprovider.hpp`` or something like that?

ajax16384:
ok, so i will try to make it in separate pull request.

Issue #158, opened 2015-09-14
Remove, or make optional, 1MB packet limit


bedeho:
The following code imposes a hard 1MB limit on incoming messages. It is a problem for a plugin I am working on, which requires big messages.

In a previous post on the mailing list @arvidn said that it was not necessary to strictly enforce, and was open for removal.
(http://sourceforge.net/p/libtorrent/mailman/libtorrent-discuss/?viewmonth=201504&viewday=28)

https://github.com/arvidn/libtorrent/blob/4fd56fd220cec486be516277682db9f91188ad1e/src/bt_peer_connection.cpp#L3514

// don't accept packets larger than 1 MB
if (packet_size > 1024*1024 || packet_size < 0)
{
 // packet too large
received_bytes(0, bytes_transferred);
disconnect(errors::packet_too_large, op_bittorrent, 2);
return;
}

arvidn:
The things I would like to make sure before removing it are:

1. that there's no allocation happening based on the message size
2. that unknown messages abort the connection (i.e. the one or two bytes following the length prefix)
3. that all known messages (including extensions) _do_ have an upper bound on what message sizes they allow (or a check for an exact size)

ideally there would be some spot checking in a test as well.

bedeho:
Sounds very reasonable, but how can you enforce 2. and 3. in libtorrent, given that libtorrent will not be aware of messages and corresponding lengths used by plugins?

arvidn:
I just meant the plugin that ships with it.

bedeho:
Ah, makes sense, will keep that in mind, thank you!

Falcosc:
@arvidn why did nobody else complain about this limit? And why did we not hit this limit?
I would like to remove it if it makes no sense. But I wouldn't if this is something related to the bittorrent protocoll description.

arvidn:
nobody else hit this problem because there are no bittorrent messages that are large enough to trigger it. It only happens if you write a plugin and send a message larger than 1 MiB.

It's a sanity check to make sure it's not possible to trick libtorrent into allocating unreasonable amounts of memory by sending corrupt data.

Falcosc:
Is it really only for messages? Then I did not understand the code or the protocol. Or did you mean that only messages could exceed 1MiB because other protocol communication are limited by design. 
I did guess that all bt connection data goes through `on_receive_impl() `  `if (m_state == read_packet_size)`

bedeho:
[...]

PR #1474, opened 2016-12-30
add message length limit configuration

Falcosc:


arvidn:
Although having that constant in there right now (1 MiB) is not good, I'm not super happy with this solution either. It just kicks the ball down to the user instead of fixing it properly.

I laid out my opinion in the ticket you reference. But perhaps an even better solution would be to make this limit be based on the message type. That way, each built-in bittorrent message could have an exact size-check up-front. any message that's an extension message could have the check immediately after the extension message type is known (by extending the plugin interface to ask for the max size for a specific message).

In some sense though, perhaps this is a step in the right direction. I'm just a bit worried about polluting the settings pack with another strange setting that doesn't have much meaning for most people and that is likely to go away.

Falcosc:
I got what you mean, we can improve the limits for the other messages but we can't skip the setting for the extension protocoll messages.
The main goal of the limit is the preventation of allocating unreasonable amounts of memory.
If we would just made a proper limit for all types and an unlimited for extension protocoll, the main goal would be failed. Or does we ignore the extension protocoll if nobody want to listen on it?

We could split the tasko into solving the problem of extension protocoll users and an overall enhancement for message size checks.

I would like to use a option name which is matching to the full solution including the enhancement.
Maybe max_extension_message_length or something similar, as option for protocoll extension message developers. Do you have a suggestion for a nice options which is used only by client developers?

Falcosc:
Maybe I could implement the full solution at the end of week.
But it’s quite a sensible spot to implement case specific handling because I guess this is one of the most called function libtorrent or it even is the most called one.
I would create a const m_max_message_length[] with a static bitfield length. @arvidn Or do you like a variable array with a bitfield length update during connection lifecycle?
m_max_message_length[20] would be m_ses.max_extension_message_length
m_max_message_length[9] would be 9 + 128 * 1024

arvidn:
the array of max messages sizes should be a constant. probably in the anonymous namespace in session_impl.cpp. I imagine a challenge would be to know the message type though. I'm not sure, but it's possible this check happens before the message type is received. Especially for extension messages, it's possible we have the message type, but maybe not the extension message type.

for extension messages, the limit would have to be defined by the plugin itself.

Falcosc:
I would not care about the extension message type type, I would just go for message type 20

how should the plugin define the limit? I mean it has to tell libtorrent. I thought a session setting would be the way to go.

And why in session_impl instead of bt_peer_connection? We already have an message type handler array there

arvidn:
right, the check would be in bt_peer_connection.cpp, so the constant array should be there too.
The peer connection could ask the plugins for a limit, or probably, ask if a given size is valid for a specific message type.
In case the receive buffer isn't allocated up-front when a message size prefix is received, it's possible the existing protocol for message handlers would be sufficient. Since handlers are called every time some more bytes are received, it could validate the size well before the whole message has been received.

Falcosc:
I still don't know how libtorrent could ask a libtorrent interface consumer for a valid size because the consumer don't provide an interface with a method to ask for a size.

arvidn:
the interface to built-in message handlers and to plugins is defined by libtorrent. it could be changed

Falcosc:
Ok, now I understand why you don't like the session option.

Falcosc:
I don't know how flexible the extension size should be.

I could implement a extension plugin iteration on one of the follwing places:
- on session start
- on_handshake
- on receive extended message bit

Which do you like? I like the middle way with on handshake

Plugin interface peer_plugin would look like this

```
// During handshake this function request the expected maximum extension
// messege size in bytes. The largest return value of the whole plugin
// chain gets selected for the message rejection and disconnect call
virtual boost::uint32_t const* max_allowed_message_lenght(
		span<char const>) const { return 1024 * 1024; }
```

arvidn:
It's probably best to target 1.2 I think. I suspect doing this right may be a bit involved.

I think the right way of doing this is to have the size limit depend on the kind of message. I believe most built-in messages have pretty tight checks already and that it's mostly extension messages (some of which are built-in) and maybe PIECE messages that may need some review and/or additional checks.

The interface to plugins (as well as the internal message handlers for that matter) is that the whole buffer from the start of the message to the end of the receive buffer (or end of message, if the entire message has been received) is passed to the handler function. This means the handler can sanity check the message size the first time it's called (or every time it's called if that's convenient, the semantics doesn't change). The PIECE message in merkle hash-tree mode may not be trivial for instance.

The only reason for the 1 MiB check right now is a catch-all fall-back sanity check, in case there is some handler that doesn't check its size.

What I think would be needed are:

1. tests where an oversized version of each message is sent to the bittorrent engine, making sure the connection is closed.
2. some review of the existing handlers, to make sure they perform the check correctly.
3. raising the current 1 MiB limit to, say 10 MiB (which still is way too large, but less likely to happen)
4. perhaps think of ways to make plugins less error prone in this regard. Maybe the message handler could return the maximum reasonable size of the incoming packet, and libtorrent could check the current message against it.

Falcosc:
I don't understand the direction of your comment.

I do plan to implement a max message type size vector for alle message types except for extension messages.
Some sizes are dependet on torrent information, others are dependend on protocol limitations.

The main step is to fix the extended messages size check, and my question does only target the extension messages part because I have no questions about the other message type checks.

To get a plugin specific extension message limitation. I want to ask all plugins for there max extendet message size expectation.

Ok, lets call the plugin interface function `max_allowed_extension_message_lenght` to make this more clear. I know that extension messages could be a whole message structure, but I would just ask for the worst case extension message to improve this step by step.

arvidn:
let me put it this way. This existing check is there to mitigate exploits of bugs. Instead of building new infrastructure to make this check more dynamic, we should just make sure there are no bugs in the message handlers.

Making the check more dynamic and complex also takes away some of its value, since now there's increased risk if the check itself to have bugs.

The primary mechanism for making sure (or increasing confidence) that the message handlers don't have bugs (in validating message sizes) is to have tests for them

Falcosc:
The dynamic part is only for the extension messages.

We could even define static values for the torrent related sizes to make it less complex.

I could also think of a less complex solution as step1 by just implementing a type check there:
if message type == extension: use use dynamic max extension size
else: 1MiB size

After this change we could implement the other message type size checks with the max size per type vector and the tests for all possible message types

bedeho:
Thanks for investing time in this, means a lot to us, we initiated the conversation here https://github.com/arvidn/libtorrent/issues/158.

Raising the limit from 1MB to 10MB would certainly make the issue go away for us in practice, as none of our plugin messages would likely ever exceed that, but the optimal solution would be to allow our plugin to tell libtorrent what the maximum message size is. Just as for the vanilla bittorrent protocol, our max message size is a function of metadata which may not be available right away, so we would need to be able to set this some time after the torrent has been added and all handshakes are done.

Falcosc:
I would move the max size check into this line:
https://github.com/Falcosc/libtorrent/blob/cd2e71d163f90205508280f03f1e2301e8dc9474/src/bt_peer_connection.cpp#L3298

because I make the size check message type based and will read the first byte after size to get the type
For type 20 we will select the current largest max size of all extension max sizes

@bedeho our first idea was to ask for a size, but it's may be better to store the max size of each extension because we can't force the extension to use max_allowed_extension_message_lenght as integer acces instead of a complex and slow business logic.

@arvidn would it ok to ask the new plugin interface `max_allowed_extension_message_lenght` on every read_packet_size case of type 20 (extension type)
Or should be store the max size of each extension and give them a method to update there max size? Because if the max_allowed_extension_message_lenght implementation is slow it would slow down everthing.

arvidn:
every message handler already receives the packet size every time it's invoked, including (potentially) before the whole packet has been received. The proper solution is to have every handler implement an overflow check.

The size should probably be an ``int32`` and the existing check should look for negative values.

For example, look at the message handler for ``on_have()``: https://github.com/arvidn/libtorrent/blob/master/src/bt_peer_connection.cpp#L1009

It checks to make sure the packet size matches what's expected.

What I think is missing is a test that ensures that all message handlers actually perform a check like this. and possible a review to make sure it appears they all currently are. I would imagine the simplest way to write such test would be to use a fake peer connection (like this one https://github.com/arvidn/libtorrent/blob/master/test/bittorrent_peer.hpp#L47 ) and send every message type with an oversized message size. The test could be to make sure the connection is closed immediately.

Falcosc:
I just want to repeat what I did understand: You want to remove this size check here https://github.com/arvidn/libtorrent/blob/master/src/bt_peer_connection.cpp#L3439
And check only for int32 overflow (2GiB package size)

And then we check the package size at each type handler.
We don't ask the plugin for size we just pass the recaived data to all extension and the extension decide to take the packet or to reject it because the announced size is to large.

If I did unterstand it correctly, then it would look alot cleaner. I like it.
But could it be an issue if somebody annonce a 1GiB large package? Because I don't know if we do allocate anything based on the announced package size because currently we don't allow passing such huge package size announcements to the handler, maybe we did this based on such kind of packet size based allocation issues.
And is it save to pass a 1GiB large packet announcement to all extensions or do we do any size based allocations for the extension interface call?

arvidn:
Your understanding is correct.

I don't believe there's any allocation that happens directly because of a packet size. That would be worth reviewing though. Note that message handlers are called repeatedly, each time more bytes arrive. The expectation is that the first few bytes that arrive, causing the handler to be called, will check the total message size there, before a significant portion of the message has been received

Another safety check the test could do is to override global operator new, and assert that only reasonably sized allocations are made.

Falcosc:
sounds nice, I could start with it after I did improve my debug build, which could take a while.
@bedeho Or maybe you could create the PR based on our discussion result. That would help us alot.
There are no interface changes only move the sizecheck to the handler implementation.

bedeho:
Sorry for being late on this, our next release needs this, so we will take a crack at it, thank you for being patient.

include/libtorrent/session_settings.hpp
```diff
@@ -176,6 +176,15 @@ namespace libtorrent
...
 		int tracker_maximum_response_length;
 
+		// Set the allowed incoming bt packed message length. This is used as a
+		// sanity check to make sure it’s not possible to trick libtorrent into
+		// allocating unreasonable amounts of memory by sending corrupt data. 
+		// Expected messages are quite small even piece messages are only 16KiB
+		// and some clients use 128KiB as Upper limit. But there is one exception
+		// which we want to cover: `BEP 10`_ Extension Protocol. To cover the most
+		// use-cases of custom extension messages we select 1MiB as default.
+		boost::uint32_t max_allowed_message_length;
```
arvidn: this breaks ABI. you cannot add new settings fields in RC_1_0. In RC_1_1 it's possible since the settings structure is more flexible (but only at the end)

arvidn: but anyway, I guess we've talked about not have this be a setting anyway

Falcosc: Yes, don't use this setting. Just move the size checks to the handler inlcuding the external extension handler

aldenml:
This seems like a wrong approach to start with, specially in `RC_1_0`. https://github.com/arvidn/libtorrent/issues/158 is still open, if this is an stalled PR, is it right to close it?

bedeho:
I did not pick up this point until now

> I just want to repeat what I did understand: You want to remove this size check here https://github.com/arvidn/libtorrent/blob/master/src/bt_peer_connection.cpp#L3439
And check only for int32 overflow (2GiB package size)
>
>And then we check the package size at each type handler.
We don't ask the plugin for size we just pass the recaived data to all extension and the extension decide to take the packet or to reject it because the announced size is to large.

With this fix, an attack can still induce libtorrent to allocate the 2GiB buffer by just sending a malicious extended message? Even if an extension detects, in `peer_plugin::on_extended`, that the message is too big, and invokes a disconnection, then that would still be _after_ this allocation step having been done. This seems bad.

arvidn:
could you provide some more details on this scenario?

receive buffer is allocated in two places, afaict:

https://github.com/arvidn/libtorrent/blob/master/src/peer_connection.cpp#L5636 and https://github.com/arvidn/libtorrent/blob/master/src/peer_connection.cpp#L5816

In tthe first case, the size is limited by the number of bytes already allocated. As far as I can tell at least, it seems a bit odd that the call to ``reserve`` is actually necessary. https://github.com/arvidn/libtorrent/blob/master/src/peer_connection.cpp#L5597

In the second case, the amount to grow is limited by what ``socket.available()`` returns. i.e. the number of bytes already received (and buffered) by the kernel. this is not likely to be more than a few megabytes tops.

https://github.com/arvidn/libtorrent/blob/master/src/peer_connection.cpp#L5798

I don't believe that the number of bytes of receive buffer to allocate is never directly related to the size of the bittorrent message. If message handlers *allow* large packets to be received though, it is *indirectly* related to it, because we will keep receiving (and keep allocating) until a message handler says we're done (or causes a disconnect because of an error).

bedeho:
Thanks for the quick response.

My concern was based on the assumption that there was some buffer allocated based on the length of a bittorrent message, and that the past 1MB hardcap was a way of dealing with this. I had not looked in detail into how this was being handled. From what you are saying, this is not the case directly. In particular in your last point you indicate that the receive buffer is expanded as data comes in, not preallocated to the potentially false 2GiB value a malicious peer may send. 

If I am understanding that correctly, then it means that
a) a malicious peer would actually need to send 2GiB of data to get libtorrent to reserve this much memory, its not enough to just advertise a false value.
b) before the scenario in a) can unfold the peer_plugin (extended) message handler will have the chance to detect such bad values before the receive buffer grows to big anyway.

is this correct? either way, it seems my concern is a non-issue.

arvidn:
yes, that's correct. at least that's the intention of the code. message handlers are called repeatedly, each time some more data arrives. It's up to the handler to determine when the whole packet has been received and the pop those bytes off the receive buffer

bedeho:
Great, thanks!

So just to conclude on what needs to be done

- Replace `packet_size > 1024*1024 || packet_size < 0` with `packet_size < 0`
here https://github.com/arvidn/libtorrent/blob/master/src/bt_peer_connection.cpp#L3439
- For each native handler: add a test which tries to send an oversized message of corresponding type, making sure that all handlers handles this case.

For the latter you reference using a mock peer class `peer_conn`, as in https://github.com/arvidn/libtorrent/blob/master/test/bittorrent_peer.hpp#L47. I cannot however see any existing tests which show how to actually use such a class to do this?

arvidn:
That is correct.

simulation/test_optimistic_unchoke.cpp uses that class to simulate peers. Another approach would be to do what test_fast_extensions.cpp does, but that's less reliable as it uses the system sockets to talk to libtorrent

bedeho:
Great, thanks!

PR #1480, opened 2016-12-31
fix time_point_to_python month conversion

Falcosc:


arvidn:
I bet this is a bug in older versions too. would you mind making this pull request against RC_1_1?
is there a good reason to add those print statements to the python tests?
can you think of a way to make the test assert that the date conversion is correct?

Falcosc:
No it isn't, we dont have time_point a 1.1 and 1.0.

arvidn:
ok, that's great. would you mind removing the print statements? also, I'm a bit worried that the same error exists in the opposite conversion, to C++.

In the test, you could use the create torrent class to set the date, and then read the date back to make sure it's right. Maybe both read back the number that get printed to the .torrent file itself as well as the python date you get back from asking the torrent_info/create_torrent object

arvidn:
it might make sense to just merge the fix first though, to get the tests passing today

PR #1482, opened 2016-12-31
fix force-proxy on udp_socket

arvidn:


aldenml:
I believe you should close this PR, since the fix will be a back port from `master`

PR #1487, opened 2017-01-01
fix move_storage() to its own directory (would delete the files)

arvidn:


anthonyryan1:
@arvidn I can confirm this appears to resolve the issue.

PR #1492, opened 2017-01-03
fix multible torrent status timer bugs related to move time

Falcosc:
I will make another commit with int32 instead of uint16 because the range limitation of 18 hour in past is to short for a value available over the interface. Maybe useable for last downloaded, but for last uploaded I would like a little bit more then 18 hours even if it's not persistent

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* fix last_upload and last_download overflow after 9 hours in past
 	* python binding add more add_torrent_params fields and an invalid key check
 	* introduce introduce distinct types for peer_class_t, piece_index_t and
 	  file_index_t.
--- a/bindings/python/test.py
+++ b/bindings/python/test.py
@@ -5,6 +5,7 @@ import libtorrent as lt
 
 import unittest
 import time
+import datetime
 import os
 import shutil
 import binascii
@@ -79,6 +80,18 @@ class test_torrent_handle(unittest.TestCase):
     def test_piece_deadlines(self):
         self.setup()
         self.h.clear_piece_deadlines()
+
+    def test_status_last_uploaded_dowloaded(self):
+        # we want to check at seconds precision but can't control session
+        # time, wait for next full second to prevent second increment
+        time.sleep(1 - datetime.datetime.now().microsecond / 1000000.0)
+
+        sessionStart = datetime.datetime.now().replace(microsecond=0)
+        self.setup()
+        st = self.h.status()
+        # last upload and download times are at session start time
+        self.assertEqual(st.last_upload, sessionStart)
+        self.assertEqual(st.last_download, sessionStart)
 
     def test_torrent_status(self):
         self.setup()
@@ -268,6 +281,7 @@ class test_alerts(unittest.TestCase):
                           'enable_dht': False})
         ses.async_add_torrent(
             {"ti": lt.torrent_info("base.torrent"), "save_path": "."})
+
 # this will cause an error (because of duplicate torrents) and the
 # torrent_info object created here will be deleted once the alert goes out
 # of scope. When that happens, it will decrement the python object, to allow
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -95,7 +95,7 @@ namespace libtorrent
 	class bt_peer_connection;
 	struct listen_socket_t;
 
-	typedef std::chrono::duration<std::int32_t, std::ratio<1>> seconds_t;
+	typedef std::chrono::duration<std::int32_t> seconds32;
 
 	enum class waste_reason
 	{
@@ -485,9 +485,10 @@ namespace libtorrent
 		void do_pause(bool clear_disk_cache = true);
 		void do_resume();
 
-		seconds_t finished_time() const;
-		seconds_t active_time() const;
-		seconds_t seeding_time() const;
+		seconds32 finished_time() const;
+		seconds32 active_time() const;
+		seconds32 seeding_time() const;
+		seconds32 upload_mode_time() const;
 
 		bool is_paused() const;
 		bool is_torrent_paused() const { return m_paused; }
@@ -1297,8 +1298,6 @@ namespace libtorrent
 		// m_num_verified = m_verified.count()
 		std::uint32_t m_num_verified = 0;
 
-		// TODO: waste of memomry because has a max session time offset of
-		// optimistic_disk_retry
 		time_point m_last_saved_resume = aux::time_now();
 
 		// if this torrent is running, this was the time
@@ -1408,7 +1407,7 @@ namespace libtorrent
 		// paused. specified in seconds. This only track time _before_ we started
 		// the torrent this last time. When the torrent is paused, this counter is
 		// incremented to include this current session.
-		seconds_t m_active_time;
+		seconds32 m_active_time;
 
 		// the index to the last tracker that worked
 		std::int8_t m_last_working_tracker = -1;
@@ -1417,7 +1416,7 @@ namespace libtorrent
 
 		// total time we've been finished with this torrent.
 		// does not count when the torrent is stopped or paused.
-		seconds_t m_finished_time;
+		seconds32 m_finished_time;
 
 		// in case the piece picker hasn't been constructed
 		// when this settings is set, this variable will keep
@@ -1457,7 +1456,7 @@ namespace libtorrent
 		// accounts for the time prior to the current start of the torrent. When
 		// the torrent is paused, this counter is incremented to account for the
 		// additional seeding time.
-		seconds_t m_seeding_time;
+		seconds32 m_seeding_time;
 
 // ----
 
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -483,7 +483,7 @@ namespace libtorrent
 		void do_pause(bool clear_disk_cache = true);
 		void do_resume();
 
-		int finished_time() const;
+		seconds finished_time() const;
 		int active_time() const;
 		int seeding_time() const;
 
@@ -1320,7 +1320,7 @@ namespace libtorrent
 
 		// if we're finished, this is the session time
 		// timestamp of when we finished
-		std::uint16_t m_became_finished = 0;
+		time_point m_became_finished = aux::time_now();
 
 		// when checking, this is the first piece we have not
 		// issued a hash job for
@@ -1420,7 +1420,8 @@ namespace libtorrent
 
 		// total time we've been finished with this torrent.
 		// does not count when the torrent is stopped or paused.
-		unsigned int m_finished_time:24;
+		// TODO was 24bit before
+		std::chrono::seconds m_finished_time;
 
 		// in case the piece picker hasn't been constructed
 		// when this settings is set, this variable will keep
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -1034,7 +1034,7 @@ namespace libtorrent
 		// that are not private
 		void lsd_announce();
 
-		void update_last_upload() { m_last_upload = int16_t(m_ses.session_time()); }
+		void update_last_upload() { m_last_upload = m_ses.session_time(); }
 
 		void set_apply_ip_filter(bool b);
 		bool apply_ip_filter() const { return m_apply_ip_filter; }
@@ -1548,7 +1548,7 @@ namespace libtorrent
 		// the timestamp of the last piece passed for this torrent specified in
 		// session_time. This is signed because it must be able to represent time
 		// before the session started
-		std::int16_t m_last_download = (std::numeric_limits<std::int16_t>::min)();
+		std::uint16_t m_last_download = 0;
 
 		// the number of peer connections to seeds. This should be the same as
 		// counting the peer connections that say true for is_seed()
@@ -1557,7 +1557,7 @@ namespace libtorrent
 		// the timestamp of the last byte uploaded from this torrent specified in
 		// session_time. This is signed because it must be able to represent time
 		// before the session started.
-		std::int16_t m_last_upload = (std::numeric_limits<std::int16_t>::min)();
+		std::uint16_t m_last_upload = 0;
 
 // ----
 
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -478,14 +478,14 @@ namespace libtorrent
 
 		void stop_when_ready(bool b);
 
-		int started() const { return m_started; }
+		time_point started() const { return m_started; }
 		void step_session_time(int seconds);
 		void do_pause(bool clear_disk_cache = true);
 		void do_resume();
 
 		seconds finished_time() const;
-		int active_time() const;
-		int seeding_time() const;
+		seconds active_time() const;
+		seconds seeding_time() const;
 
 		bool is_paused() const;
 		bool is_torrent_paused() const { return m_paused; }
@@ -1312,11 +1312,11 @@ namespace libtorrent
 		// in session-time. see session_impl for details.
 		// the reference point is stepped forward every 4
 		// hours to keep the timestamps fit in 16 bits
-		std::uint16_t m_started;
+		time_point m_started = aux::time_now();
 
 		// if we're a seed, this is the session time
 		// timestamp of when we became one
-		std::uint16_t m_became_seed = 0;
+		time_point m_became_seed = aux::time_now();
 
 		// if we're finished, this is the session time
 		// timestamp of when we finished
@@ -1411,7 +1411,8 @@ namespace libtorrent
 		// paused. specified in seconds. This only track time _before_ we started
 		// the torrent this last time. When the torrent is paused, this counter is
 		// incremented to include this current session.
-		unsigned int m_active_time:24;
+		// TODO was 24bit before
+		std::chrono::seconds m_active_time;
 
 		// the index to the last tracker that worked
 		std::int8_t m_last_working_tracker = -1;
@@ -1461,7 +1462,8 @@ namespace libtorrent
 		// accounts for the time prior to the current start of the torrent. When
 		// the torrent is paused, this counter is incremented to account for the
 		// additional seeding time.
-		unsigned int m_seeding_time:24;
+		// TODO was 24bit before
+		std::chrono::seconds m_seeding_time;
 
 // ----
 
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -95,6 +95,8 @@ namespace libtorrent
 	class bt_peer_connection;
 	struct listen_socket_t;
 
+	typedef std::chrono::duration<std::int32_t, std::ratio<1>> seconds_t;
+
 	enum class waste_reason
 	{
 		piece_timed_out, piece_cancelled, piece_unknown, piece_seed
@@ -483,9 +485,9 @@ namespace libtorrent
 		void do_pause(bool clear_disk_cache = true);
 		void do_resume();
 
-		seconds finished_time() const;
-		seconds active_time() const;
-		seconds seeding_time() const;
+		seconds_t finished_time() const;
+		seconds_t active_time() const;
+		seconds_t seeding_time() const;
 
 		bool is_paused() const;
 		bool is_torrent_paused() const { return m_paused; }
@@ -1406,8 +1408,7 @@ namespace libtorrent
 		// paused. specified in seconds. This only track time _before_ we started
 		// the torrent this last time. When the torrent is paused, this counter is
 		// incremented to include this current session.
-		// TODO was 24bit before
-		std::chrono::seconds m_active_time;
+		seconds_t m_active_time;
 
 		// the index to the last tracker that worked
 		std::int8_t m_last_working_tracker = -1;
@@ -1416,8 +1417,7 @@ namespace libtorrent
 
 		// total time we've been finished with this torrent.
 		// does not count when the torrent is stopped or paused.
-		// TODO was 24bit before
-		std::chrono::seconds m_finished_time;
+		seconds_t m_finished_time;
 
 		// in case the piece picker hasn't been constructed
 		// when this settings is set, this variable will keep
@@ -1457,8 +1457,7 @@ namespace libtorrent
 		// accounts for the time prior to the current start of the torrent. When
 		// the torrent is paused, this counter is incremented to account for the
 		// additional seeding time.
-		// TODO was 24bit before
-		std::chrono::seconds m_seeding_time;
+		seconds_t m_seeding_time;
 
 // ----
 
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -492,12 +492,7 @@ namespace libtorrent
 		void force_recheck();
 		void save_resume_data(int flags);
 
-		bool need_save_resume_data() const
-		{
-			// save resume data every 15 minutes regardless, just to
-			// keep stats up to date
-			return m_need_save_resume_data || m_ses.session_time() - m_last_saved_resume > 15 * 60;
-		}
+		bool need_save_resume_data() const;
 
 		void set_need_save_resume()
 		{
@@ -1300,9 +1295,9 @@ namespace libtorrent
 		// m_num_verified = m_verified.count()
 		std::uint32_t m_num_verified = 0;
 
-		// this timestamp is kept in session-time, to
-		// make it fit in 16 bits
-		std::uint16_t m_last_saved_resume;
+		// TODO: waste of memomry because has a max session time offset of
+		// optimistic_disk_retry
+		time_point m_last_saved_resume = aux::time_now();
 
 		// if this torrent is running, this was the time
 		// when it was started. This is used to have a
@@ -1369,7 +1364,7 @@ namespace libtorrent
 
 		// the session time timestamp of when we entered upload mode
 		// if we're currently in upload-mode
-		std::uint16_t m_upload_mode_time = 0;
+		time_point m_upload_mode_time = aux::time_now();
 
 		// true when this torrent should announce to
 		// trackers
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -1034,7 +1034,7 @@ namespace libtorrent
 		// that are not private
 		void lsd_announce();
 
-		void update_last_upload() { m_last_upload = m_ses.session_time(); }
+		void update_last_upload() { m_last_upload = aux::time_now(); }
 
 		void set_apply_ip_filter(bool b);
 		bool apply_ip_filter() const { return m_apply_ip_filter; }
@@ -1548,7 +1548,7 @@ namespace libtorrent
 		// the timestamp of the last piece passed for this torrent specified in
 		// session_time. This is signed because it must be able to represent time
 		// before the session started
-		std::int32_t m_last_download = 0;
+		time_point m_last_download = aux::time_now();
 
 		// the number of peer connections to seeds. This should be the same as
 		// counting the peer connections that say true for is_seed()
@@ -1557,7 +1557,7 @@ namespace libtorrent
 		// the timestamp of the last byte uploaded from this torrent specified in
 		// session_time. This is signed because it must be able to represent time
 		// before the session started.
-		std::int32_t m_last_upload = 0;
+		time_point m_last_upload = aux::time_now();
 
 // ----
 
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -1548,7 +1548,7 @@ namespace libtorrent
 		// the timestamp of the last piece passed for this torrent specified in
 		// session_time. This is signed because it must be able to represent time
 		// before the session started
-		std::uint16_t m_last_download = 0;
+		std::int32_t m_last_download = 0;
 
 		// the number of peer connections to seeds. This should be the same as
 		// counting the peer connections that say true for is_seed()
@@ -1557,7 +1557,7 @@ namespace libtorrent
 		// the timestamp of the last byte uploaded from this torrent specified in
 		// session_time. This is signed because it must be able to represent time
 		// before the session started.
-		std::uint16_t m_last_upload = 0;
+		std::int32_t m_last_upload = 0;
 
 // ----
 
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -95,7 +95,7 @@ namespace libtorrent
 	class bt_peer_connection;
 	struct listen_socket_t;
 
-	typedef std::chrono::duration<std::int32_t> seconds32;
+	using seconds32 = std::chrono::duration<std::int32_t>;
 
 	enum class waste_reason
 	{
@@ -495,7 +495,13 @@ namespace libtorrent
 		void force_recheck();
 		void save_resume_data(int flags);
 
-		bool need_save_resume_data() const;
+		bool need_save_resume_data() const
+		{
+			// save resume data every 15 minutes regardless, just to
+			// keep stats up to date
+			return m_need_save_resume_data ||
+				aux::time_now() - m_last_saved_resume > minutes(15);
+		}
 
 		void set_need_save_resume()
 		{
--- a/simulation/setup_swarm.cpp
+++ b/simulation/setup_swarm.cpp
@@ -395,13 +395,14 @@ void setup_swarm(int num_nodes
 
 		bool shut_down = terminate(tick, *nodes[0]);
 
-		if (type == swarm_test::upload)
+		if (type == swarm_test::upload || type == swarm_test::upload_slow)
 		{
 			shut_down |= std::all_of(nodes.begin() + 1, nodes.end()
 				, [](std::shared_ptr<lt::session> const& s)
 				{ return is_seed(*s); }) && num_nodes > 1;
 
-			if (tick > 88 * (num_nodes - 1) && !shut_down && num_nodes > 1)
+			if (tick > 88 * (num_nodes - 1) && !shut_down && num_nodes > 1
+				&& type != swarm_test::upload_slow)
 			{
 				TEST_ERROR("seeding failed!");
 				shut_down = true;
--- a/simulation/setup_swarm.cpp
+++ b/simulation/setup_swarm.cpp
@@ -395,14 +395,14 @@ void setup_swarm(int num_nodes
 
 		bool shut_down = terminate(tick, *nodes[0]);
 
-		if (type == swarm_test::upload || type == swarm_test::upload_slow)
+		if (type == swarm_test::upload || type == swarm_test::upload_without_timeout)
 		{
 			shut_down |= std::all_of(nodes.begin() + 1, nodes.end()
 				, [](std::shared_ptr<lt::session> const& s)
 				{ return is_seed(*s); }) && num_nodes > 1;
 
 			if (tick > 88 * (num_nodes - 1) && !shut_down && num_nodes > 1
-				&& type != swarm_test::upload_slow)
+				&& type != swarm_test::upload_without_timeout)
 			{
 				TEST_ERROR("seeding failed!");
 				shut_down = true;
--- a/simulation/setup_swarm.cpp
+++ b/simulation/setup_swarm.cpp
@@ -395,14 +395,13 @@ void setup_swarm(int num_nodes
 
 		bool shut_down = terminate(tick, *nodes[0]);
 
-		if (type == swarm_test::upload || type == swarm_test::upload_without_timeout)
+		if (type == swarm_test::upload)
 		{
 			shut_down |= std::all_of(nodes.begin() + 1, nodes.end()
 				, [](std::shared_ptr<lt::session> const& s)
 				{ return is_seed(*s); }) && num_nodes > 1;
 
-			if (tick > 88 * (num_nodes - 1) && !shut_down && num_nodes > 1
-				&& type != swarm_test::upload_without_timeout)
+			if (tick > 88 * (num_nodes - 1) && !shut_down && num_nodes > 1)
 			{
 				TEST_ERROR("seeding failed!");
 				shut_down = true;
--- a/simulation/test_torrent_status.cpp
+++ b/simulation/test_torrent_status.cpp
@@ -96,7 +96,50 @@ TORRENT_TEST(status_timers)
 	TEST_CHECK(ran_to_completion);
 }
 
+TORRENT_TEST(status_timers_last_upload)
+{
+	bool ran_to_completion = false;
 
+	lt::time_point start_time;
+	lt::torrent_handle handle;
+
+	setup_swarm(2, swarm_test::upload
+		// add session
+		, [](lt::settings_pack&) {}
+		// add torrent
+		, [](lt::add_torrent_params&) {}
+		// on alert
+		, [&](lt::alert const* a, lt::session&) {
+			if (auto ta = alert_cast<torrent_added_alert>(a))
+			{
+				TEST_CHECK(!handle.is_valid());
+				start_time = lt::clock_type::now();
+				handle = ta->handle;
+				torrent_status st = handle.status();
+				// test last upload and download state before wo go throgh
+				// torrent states
+				TEST_CHECK(st.last_download == start_time);
+				TEST_CHECK(st.last_upload == start_time);
+			}
+		}
+		// terminate
+		, [&](int ticks, lt::session&) -> bool
+		{
+			if (ticks > 10)
+			{
+				ran_to_completion = true;
+				return true;
+			}
+
+			torrent_status st = handle.status();
+			// uploadtime is 0 seconds behind now
+			TEST_EQUAL(duration_cast<seconds>(st.last_upload - lt::clock_type::now()).count(), 0);
+			// does not download in seeding mode
+			TEST_CHECK(st.last_download == start_time);
+			return false;
+		});
+	TEST_CHECK(ran_to_completion);
+}
 
 TORRENT_TEST(status_timers_time_shift_with_active_torrent)
 {
--- a/simulation/test_torrent_status.cpp
+++ b/simulation/test_torrent_status.cpp
@@ -190,7 +190,7 @@ TORRENT_TEST(status_timers_time_shift_with_active_torrent)
 					tick_is_in_active_range = false;
 					break;
 				case 64000:
-					// resum just before we hit the time shift handling
+					// resume just before we hit the time shift handling
 					// this is needed to test what happend if we want to
 					// shift more time then we have active time because
 					// we shift 4 hours and have less then 1 hours active time
--- a/simulation/test_torrent_status.cpp
+++ b/simulation/test_torrent_status.cpp
@@ -41,37 +41,15 @@ using namespace sim;
 namespace lt = libtorrent;
 
 // this is a test for torrent_status time counters are correct
-// it is a long running test, we did already try to improve it:
-// - less slow peers vs many fast peers
-// - slow peers with smal torrent vs fast peers with large torrent
-// - less peers with large torrents vs more peers with smal torrent
-// - connection limit vs parallel requests
-// as conlusion the following combination take the least amount of
-// execution time per tick:
-// - many slow peers with small torrent and connection limit
-TORRENT_TEST(status_timers_time_shift_with_paused_torrent)
+TORRENT_TEST(status_timers)
 {
 	lt::time_point start_time;
-	lt::time_point last_active_time;
 	lt::torrent_handle handle;
 	bool ran_to_completion = false;
-	seconds expected_active_duration = seconds(0);
-	bool tick_is_in_active_range = false;
 
-	// TODO: this setup takes 1 minute execution time and could be improved by
-	// simulate more time per tick, currently we execute 71000 ticks
-	setup_swarm(10, swarm_test::upload_without_timeout
+	setup_swarm(1, swarm_test::upload
 		// add session
-		, [](lt::settings_pack& pack) {
-			pack.set_int(lt::settings_pack::alert_mask, alert::status_notification
-				| alert::torrent_log_notification);
-			// slow as possible to do less compute per tick, but not too slow
-			// because simulator want to get done in 88 ticks per node
-			// 2KiB/s is very close to limit and could fail if the seed code get
-			// less efficient or if the simulator use more data by default
-			pack.set_int(settings_pack::upload_rate_limit, 2*1024);
-			pack.set_int(settings_pack::connections_limit, 2);
-		}
+		, [](lt::settings_pack&) {}
 		// add torrent
 		, [](lt::add_torrent_params&) {}
 		// on alert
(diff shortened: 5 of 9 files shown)


Falcosc:
@arvidn how could I execute the simulator tests? 
If I execute bjam at simulation/
It tells me that it miss a Jamfile at libsimulator and this folder is indeed empty

arvidn:
you need to run ``git submodule update``

src/torrent.cpp
```diff
@@ -8252,8 +8252,9 @@ namespace libtorrent
 		}
 		m_became_finished = clamped_subtract_u16(m_became_finished, seconds);
 
-		m_last_upload = clamped_subtract_s16(m_last_upload, seconds);
-		m_last_download = clamped_subtract_s16(m_last_download, seconds);
+		// don't care about overflow because it would take 68 years
+		m_last_upload = m_last_upload - seconds;
+		m_last_download = m_last_download - seconds;
```
arvidn: I wonder if all relative timestamps that are incremented every second should be replaced by absolute timetstamps. The place where the space savings are the most relevant are probably in the peer list, so it might make sense to keep it there.
another option would be to decrease the precision to 2 seconds (to double the range)

Falcosc:
Sorry, but I am not familiar enough with all modules to interpret your comment correctly.
Is the content of torrent.cpp multiplied by the peer list?
Isn’t an absolute timestamp not a little bit to large? I mean somebody had really try to save every bit in torrent.cpp, even the overflow of the sign int was used with clamped_subtract_s16, which I didn’t like at all because it’s hard to understand how a value overflow got reused as valid range. 
With creating the new fields somebody did directly add the last_upload/last_download seconds to the timestamp without handling the overflow range https://github.com/arvidn/libtorrent/blob/master/src/torrent.cpp#L10571  which results into wrong values after 9 hours.
The question is, what do you want to store in torrent structure? 32bit seconds, 64bit timestamps or the huge high precision time_point?
And what should be the default value? I did think about timestamp 0 but it’s maybe a bit confusing if the value changes after 14-18hours to 1970. I did also think about a null interpretation at the interface based on timestamp 0. Or if we could use more than 16 bit, maybe the session start time the best solution because the values are not persistent. What did you think about it?
And the other point is, we update the relative times only once every 4 hours and on every tick during downloading and uploading. So I’m not sure which relative timestamps should be absolute and which are not.

Falcosc:
@arvidn simulator seams to be broken, it does not tick anymore
the following example test does not reach ticks > 100 and passes
```
TORRENT_TEST(status_timers)
{
	lt::time_point start_time = lt::clock_type::now();
	lt::torrent_handle handle;

	setup_swarm(1, swarm_test::upload
		// add session
		, [](lt::settings_pack&) {}
		// add torrent
		, [](lt::add_torrent_params&) {}
		// on alert
		, [&](lt::alert const* a, lt::session&) {
			if (auto ta = alert_cast<torrent_added_alert>(a))
			{
				TEST_CHECK(!handle.is_valid());
				start_time = lt::clock_type::now();
				handle = ta->handle;
			}
		}
		// terminate
		, [&](int ticks, lt::session&) -> bool
		{
			printf("tick: %d \n", ticks);
			if(ticks > 100){
				TEST_CHECK(false);
			}
			return false;
		});
}
```

Falcosc:
@arvidn do you know how to raise the tick size for simulator? It's 1000ms by default I need something arround 60*1000 or 600*1000 because it takes way to much ticks until the test will finish.

I did already try 
```
// add session
, [](lt::settings_pack& pack) {
	pack.set_int(settings_pack::tick_interval, 600*1000);
}
```

arvidn:
by "tick_size" do you mean the tick interval? the tick interval is the time in between calls to tick(). I imagine you would just like the test to run for at least 18 or so hours. so just set the timer that shuts down the test to 19 hours (in sim, it's quick anyway)

arvidn:
time_point is 64 bits also btw. I believe in the torrent structure, the complication of using "session_time" (number of seconds since session start and clamped at no older than 18 hours) may not be worth the space savings. The peer-lists (where each known peer has a few timestamps) the space savings is multiplied by thousands, if not tens of thousands, of peers, which probably makes it worth the complexity.

each torrent contain a peer list.

Another benefit of moving away from using session time in the torrent object is that the step_session_time() function has less work to do, and becomes simpler.

Falcosc:
@arvidn I still have the 2 questions, which value for no last download and no last upload
and which fieldtypes do you want to have in torrect structure?

I guess you want to have time_point in torrent.hpp because it does not matter how large the torent structure fields are because we don't have more torrent structures allocated as we have torrents in session, isn't it?
And I guess you like the session start date as default because you did not disagree?

And I guess I need still to raise the time raise between simulator ticks.
It is not fast enogh to go throw 65000 ticks, because the simulator shotdown if nothing is to do, that means I have to seed 65000 times. 800 ticks does take already arround 10s
For that reason I need larger time steps per tick to reach 60000 in 1000 ticks instead of 60000 ticks.

With peerlist you mean pear info I guess, but this structure isn't optimized at all, I guess we have only as many peer info objects as we have open connections, which means less then torrents. Or did I miss something? If we release peer info on disconnect then I still not know which peer object multiplication you mean.

arvidn:
right, I think using ``time_point`` in the torrent would be reasonable. In the future, it could be optimized by turning these timestamps into ``std::chrono::time_point<libtorrent::aux::session_clock_type, std::chrono::duration<std::int32_t>>``, that way it would have second-resolution and only use 32 bits.

``libtorrent::aux::session_clock_type`` doesn't exist, but would be a clock with the session start time as its epoch.

arvidn:
I would not recommend altering the tick_interval. the problem with the swarm simulator is that it deliberately shuts down after 88 seconds.

  https://github.com/arvidn/libtorrent/blob/master/simulation/setup_swarm.cpp#L403

 the ``setup_swarm()`` utility is just meant to test transfers between two libtorrent instances under various settings. You may not need that facility, you just want to progress time and eventually ask for ``torrent_status``.

I suppose that the timers you're looking at requires some transfers to happen though..

Falcosc:
No it doesn't, swarm simulator shutdown after after 88 ticks, not after 88 realtime seconds and not after 88 simulated seconds.
If we would have 1000s per tick, we could test the 4 hours shift with 66 ticks, that was my idea.

I guess we still need a test for the 4 hours shift because during the improvement I would only change last_downloadad and last_uploaded to time_point and maybe some other fields. But there will be still fields which need the 4 hours shift.

We don't neet a timeshift test if we want to plan to remove time shifting in future. Do we plan to remove time shifting? I guess not, at least not for torrent_peer.

My next task would be to change m_last_upload and m_last_download to time_point.
Should I change m_last_saved_resume and m_upload_mode_time too? Because that would be just a waste of memory, but as I did understand correctly you would like to have time_point everywhere because we could reduce the range and precision later.
But I don't touch torrent_peer members
Should I change m_started, m_became_seed, m_became_finished as well or should we do this as a second step?

arvidn:
if the simulator (which is the code in the libsimulator submodule) has an upper limit to how long it can run for, that's a bug. But I doubt it does. the test_swarm function that the test is using may have that limitation, and it would need to be lifted/removed to support this test. If you change the tick_interval for the test, it will be less accurate since it won't do the same thing as an actual live client.

as your patch hints, the test should probably assert that it does in fact run through the whole simulated time that's intended.

The 16 bit, low resolution timestamps used in ``torrent_peer`` should remain. That means they still have to be adjusted periodically against a moving epoch (which currently is the session start time, but those should probably be untangled. there should probably be an explicit peer timestamp epoch).

All timestamps in the torrent object that's currently subject to the time shifting, should probably be made proper absolute timestamps, and they should have a proper time_point type. the nice thing about making a custom time_point and duration, is that one can specify any ``ratio<>``. so for instance, to have a 2-second resolution, you just say ``ratio<2>``.

Falcosc:
Thanks for the answer. 
That means we still need to test the time shift because even in future we want to move epoch for torrent_peer timestamps.
I'm aware of the problem with tick_intervall because the code does not expect values over 10s. For that reason a start to extent the test to make 66000 ticks instead of 30.
This could result into 20minutes runtime, if I try to adjust it to make less aktions per tick I could maybe reduce it to 5min. Which testruntime is acceptable for you or should there some option to skip this long running test during development?
I am not sure to find any solution to get 66000 tick executions under 5 minutes, maybe we should chose the middle way between tick accuracy and tick executions by using 10s except of 1s?

Falcosc:
I did fix a 0 division bug at download rate force for tick > 1000.
And I will add a unlimited upload time option to be able to pause the torent untill tick 64000 to solve the execution time issue.

Falcosc:
The test do need only 1 minute, I can't get it faster and I hope this will be fast enoght.

simulation/test_torrent_status.cpp
```diff
@@ -41,15 +41,38 @@ using namespace sim;
...
 	lt::torrent_handle handle;
 	int last_tick;
+	// at least 11s intervall because after resume it takes 11s untill we unchuck
+	// new pieces at this slow upload rate
+	int ticks_check_intervall = 20;
+	seconds active_time = seconds(0);
+	bool tick_is_in_active_range = false;
 
-	setup_swarm(2, swarm_test::upload
+	setup_swarm(10, swarm_test::upload_slow
```
arvidn: do you really need 10 nodes?

src/peer_connection.cpp
```diff
@@ -5327,8 +5327,8 @@ namespace libtorrent
 		{
 			return (std::max)((std::max)(m_outstanding_bytes
 				, m_recv_buffer.packet_bytes_remaining()) + 30
-				, int(std::int64_t(m_statistics.download_rate()) * 2
-					/ (1000 / tick_interval)));
+				, int(std::int64_t(m_statistics.download_rate() * 2
+					* tick_interval) / 1000));
```
arvidn: this was a division by zero error, right?

simulation/test_torrent_status.cpp
```diff
@@ -76,35 +99,82 @@ TORRENT_TEST(status_timers)
...
-				// torrent get ready for seeding on first tick, which means +1s
-				startseed_time = lt::clock_type::now();
+			if(tick_is_in_active_range){
+				// 1 second per tick
+				active_time++;
 			}
 
+			// resume on a tick which is far away from next tick check intervall
+			// because with very low upload limit it takes some time to unchock
+			switch(ticks)
```
arvidn: this code is not very intuitive.. I take it you set the ``tick_interval`` to large values for parts of the test to speed up the simulation, is that right?
another way to speed up the test (at least a little bit) is to pause the torrent

simulation/test_torrent_status.cpp
```diff
@@ -76,35 +99,82 @@ TORRENT_TEST(status_timers)
...
-				auto const since_start = duration_cast<seconds>(now - startseed_time);
 				torrent_status st = handle.status();
-				TEST_CHECK(st.active_duration == since_start);
-				TEST_CHECK(st.seeding_duration == since_start);
-				TEST_CHECK(st.finished_duration == since_start);
+				TEST_CHECK(st.active_duration == active_time);
+				TEST_CHECK(st.seeding_duration == active_time);
+				// TODO finished_duration seems to be broken after time shift
+				TEST_CHECK(st.finished_duration == active_time);
+				printf("active_time: %ld \n", active_time.count());
```
arvidn: instead of printing these values, you could use ``TEST_EQUAL(st.active_duration, active_time);``

simulation/test_torrent_status.cpp
Falcosc: I did head 1000 before, only one node would use 70 ticks with this torrent size and rate limit, 70 ticks are still a little bit tight for testing so much.

But there is no much difference beteween 2 and 10, only 3-5s less execution time. Lowering upload rate limit would make it worse becaus you get more ticks without upload activity. Larger torrents should help. I could change it to 2 with a 5 times larger torrent.
Maybe it would be nice to have 10 nodes for the torrent peer timestamp test. But maybe that is to complicated to test and I change it to 2 nodes with 5 times larger torrent anyway.

src/peer_connection.cpp
Falcosc: yes it is.

simulation/test_torrent_status.cpp
```diff
@@ -41,62 +41,139 @@ using namespace sim;
...
-
-	setup_swarm(1, swarm_test::upload
+	int last_tick;
+	// at least 11s intervall because after resume it takes 11s untill we unchuck
+	// new pieces at this slow upload rate
+	int ticks_check_intervall = 20;
+	seconds active_time = seconds(0);
+	bool tick_is_in_active_range = false;
+
+	setup_swarm(10, swarm_test::upload_slow
```
arvidn: could you explain again why you need 10 nodes? I would expect it to be 10 times as expensive to simulate 10 sessions instead of one. as far as I can tell, all you want is the smallest amount of upload or download activity (to set the timestamps) and then just wait. during the wait, presumably you want to do as little work as possible. e.g. as few session instances as possible.

I don't understand why you would need a larger torrent. All you need to set the timestamps is a single byte, no?

simulation/test_torrent_status.cpp
Falcosc: I don't do anything with the ticks_intervall because it does not speed up the simulation. Maybe because the download nodes still use 500ms ticks. That means the test use the default 500ms tick size.
I do stop and start the torrent before and after critiacal ticks to check them with less active torrent ticks because ticks with download and upload actions are quite expensive in terms of execution time. Running through 70000 paused torrent ticks take already 2/3 of the test exeution time and 1/3 of the execution time is used for 700 active ticks.
I could comment every case if thes existing case comments does already help to understand that.

If you mean ticks_check_intervall, this is just for prevent spamming assert errors into log, it does not save much execution time

simulation/test_torrent_status.cpp
Falcosc: I want to, but the compiler told me why the developer did this in past, TEST_EQUAL won't take them as argument. I would need to cast them into something. Maybe a .count() like for printf, but I am not sure what count return if precision is 2, for example 10s precision 2s would it return count 5 or 10?

simulation/test_torrent_status.cpp
Falcosc: I need download and upload activity over many ticks. That could be done by raising nodes, by raising torrent size or by limiting rate.
Like commented at the method header large torrents + 1 peer (2 nodes) vs 9 peers + small torrent is no difference. Because the amount of transferred blocks are equal. Actually large torrents + 1 peer was slower, I guess because it is more expensive to handle large torrents instead of the retransferring a small torrent multiple times even with the overhead of multiple peers.

simulation/test_torrent_status.cpp
```diff
@@ -41,62 +41,139 @@ using namespace sim;
...
-				}
-				else
-				{
-					// TODO: this should really be a proximity-check
-					TEST_CHECK(st.last_download == start_time + lt::seconds(1));
-				}
+				TEST_CHECK(st.active_duration == active_time);
+				TEST_CHECK(st.seeding_duration == active_time);
+				TEST_CHECK(st.finished_duration == active_time);
+				printf("active_time: %ld \n", active_time.count());
```
Falcosc: @arvidn I need to remove the printf anyway beause on windows %ld does not work because it's not a long its int64, weird.
If count is safe to use I could write TEST_EQUAL(st.active_duration.count(), active_time.count()); otherwise we need to make something to make TEST_EQUAL compatible with durations.

simulation/test_torrent_status.cpp
Falcosc: Or do you mean the comment
```
// resume on a tick which is far away from next tick check intervall
// because with very low upload limit it takes some time to unchock
```
?

Without limit I got upload aktivity on tick 1, but with rate limit the first upload activity is on tick 11. After the first start or resume unchock it get better and I get upload activity every 5 ticks.

simulation/test_torrent_status.cpp
arvidn: why do you need transfers over multiple ticks?

simulation/test_torrent_status.cpp
arvidn: well, ``%ld`` is just not portable, to print ``int64_t`` you use the ``PRId64`` macro.

simulation/test_torrent_status.cpp
arvidn: I find it unintuitive that a large torrent would be more expensive. I imagine it would require more network packets and more ticks to complete, and hence be more costly over time. but so should more peers

simulation/test_torrent_status.cpp
arvidn: you could set ``tick_interval`` for both sessions though

simulation/test_torrent_status.cpp
Falcosc: Because the same torrent need to get multible Upload time updates to check that an update does work before and after shift of timebase.

Network packet amount is the same for this setup comparation, I just need default torrentsize * 9 / 2KiB rate to get enoght upload ticks for the test. Could be one large torrent or 9 times a small torrent.
I was surprised about that fact that a large torrent * 1 takes more execution time then small torrent * 9, too.
It was not mutch more, We have 9% less piece hash checks with the small torrent because seeder has to do 90% less hash checks and we need smaller cache sizes per session with smaller torrents and we use more threads with multible nodes. In sum all these things save more execution time then we have to spend for the overhead of 8 additional nodes.

Let's make the decision large torrent * 1 vs small torrent * 9 based on the torrent_peer timestamp test implementation. If it is to difficould to test 9 peers because I don't know when the do anything, then I go for nodes 2, otherwise I would use nodes 10 to test multiple peer timer changes.

simulation/test_torrent_status.cpp
Falcosc: That could help to keep the 2/3 execution time of the 70000 paused torrent ticks a little bit lower. I could add a comment to the test that we need to test a setup where both sessions have a long tick_interval. But I don't want to priorise it. I did already spend 6 hours with improving the test execution time and another 6 hours to write the test and fix the errors. I would like to spend time for the torrent_peer timer test and for the change to time_point instead of reducing test execution time by 20s

simulation/test_torrent_status.cpp
Falcosc: I begin to understand why it takes so long to compile c++ project if we have so much macros :D
I will add a TODO comment to improve TEST_EQUALS because I'm more effective at easy tasks. This time I will use the TODO: notation to get picked up by your script

arvidn:
I'm tempted to break out the two other bugfixes in here and make separate pull requests for them, probably against RC_1_1. I will do that in a bit (unless you beat me to it :P )

arvidn:
ok, this seems to be sufficient to make the test run properly and pass: https://github.com/arvidn/libtorrent/pull/1499
That's against RC_1_1 btw. all of these bugs should probably be fixed in there

src/torrent.cpp
```diff
@@ -8244,16 +8244,16 @@ namespace libtorrent
 			m_seeding_time += lost_seconds;
 		}
 		m_became_seed = clamped_subtract_u16(m_became_seed, seconds);
-
-		if (m_finished_time < seconds && is_finished())
+		if (m_finished_time < seconds && is_finished() && !is_paused())
```
arvidn: adding ``&& !is_paused()`` here changes the semantics of finished_time, doesn't it? what's the rationale for this change?

Falcosc: I don't know what is correct because the var names does confuse me.

m_finished_time seems to be used to tell how long a torrent was in finish state and m_became_finished is the time at which we go into finish state.

That means we need to add lost seconds only if it is not paused during time switch.

I am not sure if this is correct, but m_seeding_time use is_seed() which means not paused as well and m_active_time does have `!is_paused()` too. After the change all 3 times are handled equally. If the new semantics is wrong, then m_seeding_time and m_active_time is wrong too.

Maybe it's more easy to understand what we get over the interface, and without skiping pause like at the other 2 times we got a wrong value at st.finished_duration.

Falcosc: I guess I need to create another test where we test time shift with an active torrent. I could use 1 node with unlimited online time like you did on RC_1_1 because I don't need to test upload time it would be enogh to test finish time. But I guess it would still need 30s execution time to get 70000 ticks executed. We will see.

arvidn:
do you want to rebase this on top of master? the division by zero patch and possibly some patch against the test would go away.

Falcosc:
Crazy, now my 6 days old commit is ahead of the current master without another commit :D

src/torrent.cpp
```diff
@@ -8248,15 +8249,16 @@ namespace libtorrent
 		}
 		m_became_seed = clamped_subtract_u16(m_became_seed, seconds);
 
-		if (m_finished_time < seconds && is_finished())
+		if (m_became_finished < seconds && is_finished() && !is_paused())
```
Falcosc: @arvidn I did find the bug and I could clarify why is_paused does not matter

Falcosc:
@arvidn does this match to your expectations? https://github.com/arvidn/libtorrent/pull/1492/commits/c42037a971a0eb56bda5203ff68623beedbf2053

If yes I would repeat it for all other members.

src/torrent.cpp
```diff
@@ -102,6 +102,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #endif
 
 using namespace std::placeholders;
+using namespace std::chrono;
```
arvidn: generally it's better to just pull in the names you need, like this:
using std::chrono::duration_cast;
using std::chrono::seconds;

src/torrent.cpp
```diff
@@ -8283,7 +8277,7 @@ namespace libtorrent
 
 		int ret = 0;
 
-		std::int64_t const fin_time = finished_time();
+		std::int64_t const fin_time = duration_cast<seconds>(finished_time()).count();
```
arvidn: I think you should preserve the type safe representation of std::chrono::seconds (or any duration type) for as long as possible here. no matter what duration type this returns, you can still subtract it from the duration type ``active_time()`` returns.

src/torrent.cpp
```diff
@@ -8453,7 +8447,8 @@ namespace libtorrent
 			m_seeding_time += m_ses.session_time() - m_became_seed;
 
 		if (is_finished())
-			m_finished_time += m_ses.session_time() - m_became_finished;
+			m_finished_time += duration_cast<seconds>(aux::time_now()
```
arvidn: if ``m_finished_time`` is a duration type, I don't think you need to cast here

arvidn:
yeah, that patch looks reasonable. see comments though

src/torrent.cpp
Falcosc: I cast it because the time ration except seconds. If both are the same duration I could just use count.

I would like to place a custom duration type to because at performance critical places like this sort criteria I want to skip the cast to same ration type.
During declaring a custom time I would use the opportunity to declare it with a size matching to its use-case. I'm not aware of the implications of stripping a 32bit type down by some bits or by raising a smaller 16bit type by some bits.
Something like 
`typedef std::chrono::duration<int:28, std::ratio<1, 1>> sec;`
which is enough for  -4 years to + 4 years
Or isn’t it worth it because of the byte fragmentation? And thats not a valid declaration.

src/torrent.cpp
Falcosc:
```
std::chrono::duration<_Rep, _Period>& std::chrono::duration<_Rep, _Period>::operator+=(const std::chrono::duration<_Rep, _Period>&) [with _Rep = long int; _Period = std::ratio<1l>]
  operator+=(const duration& __d)
  ^~~~~~~~
/usr/include/c++/6/chrono:313:2: note:   no known conversion for argument 1 from ‘std::__success_type<std::chrono::duration<long int, std::ratio<1l, 1000000000l> > >::type {aka std::chrono::duration<long int, std::ratio<1l, 1000000000l> >}’ to ‘const std::chrono::duration<long int>&’
```

src/torrent.cpp
arvidn: you can't have fields smaller than a byte (short of bitfields, but they're quite restricted, you can't define a time_point / duration type with smaller than one byte granularity).

src/torrent.cpp
arvidn: I'm not sure what you intended to do. but ``m_finished_time`` is a duration, ``time_now()`` is a time_point. you cannot add a time_point to a duration. If you think about it, adding two absolute times together is non-sensical.

include/libtorrent/torrent.hpp
```diff
@@ -1420,7 +1420,8 @@ namespace libtorrent
 
 		// total time we've been finished with this torrent.
 		// does not count when the torrent is stopped or paused.
-		unsigned int m_finished_time:24;
+		// TODO was 24bit before
+		std::chrono::seconds m_finished_time;
```
arvidn: do you intend to do the same thing to ``m_active_time`` too?

simulation/setup_swarm.cpp
```diff
@@ -395,13 +395,14 @@ void setup_swarm(int num_nodes
 
 		bool shut_down = terminate(tick, *nodes[0]);
 
-		if (type == swarm_test::upload)
+		if (type == swarm_test::upload || type == swarm_test::upload_slow)
```
arvidn: I don't think you need to do this anymore, do you?

arvidn: you could turn it into a ``swarm_test::download`` test

include/libtorrent/torrent.hpp
arvidn: and ``m_seeding_time``

src/torrent.cpp
```diff
@@ -10554,22 +10550,22 @@ namespace libtorrent
...
+		st->finished_time = int(duration_cast<seconds>(finished_time()).count());
 		st->active_time = active_time();
 		st->seeding_time = seeding_time();
 
-		st->time_since_upload = m_last_upload == (std::numeric_limits<std::int16_t>::min)() ? -1
-			: clamped_subtract_u16(m_ses.session_time(), m_last_upload);
-		st->time_since_download = m_last_download == (std::numeric_limits<std::int16_t>::min)() ? -1
-			: clamped_subtract_u16(m_ses.session_time(), m_last_download);
+		st->time_since_upload = int(total_seconds(m_last_upload
+			- m_ses.session_start_time()));
```
arvidn: I think you meant to say ``aux::time_now() - m_last_upload``, no? (and same for ``m_last_download``)

src/torrent.cpp
Falcosc: you missed the linebreak, its the duration between now and became finished. And this is a duration<long int, std::ratio<1l, 1000000000l>

But the + operator wants a duration<long int>

simulation/setup_swarm.cpp
Falcosc: there would be the same problem because the download does finish to fast

src/torrent.cpp
Falcosc: you are right

include/libtorrent/torrent.hpp
Falcosc: Sure, I did just start with one to get your response before I change everything

Falcosc:
Hi @arvidn 
I guess this is as far as I would go.

Please have a look on it. I'm not sure about the typedef and about all casts.

I did had a look on steady_clock and how to use it with 16bit duration for peer_list. It's not to difficould to implement a custom clock. But updating the custom clock epoch is to much work for me. There are to many things to test.

arvidn:
yeah, I think updating those timestamps to use proper chrono time_points should be a separate patch

simulation/setup_swarm.cpp
arvidn: and the download finished automatically triggers the test to exit?

bindings/python/test.py
```diff
@@ -75,6 +76,18 @@ def test_torrent_status(self):
 		st2 = self.h.status()
 		self.assertEqual(st2, st)
 
+	def test_status_last_uploaded_dowloaded(self):
+		# we want to check at seconds precision but can't control session
+		# time, wait for next full second to prevent second increment
+		time.sleep(1 - datetime.datetime.now().microsecond / 1000000.0)
+
+		sessionStart = datetime.datetime.now().replace(microsecond=0)
```
arvidn: is this reliable? wouldn't it be more robust to just take a timestamp immediately before and after constructing the session, and perhaps take the average of the two as the session start time.

regardless of how you slice it though, you'll probably need some slack in the test below. I would suggest making sure they are within some number of milliseconds of each other

include/libtorrent/torrent.hpp
```diff
@@ -95,6 +95,8 @@ namespace libtorrent
 	class bt_peer_connection;
 	struct listen_socket_t;
 
+	typedef std::chrono::duration<std::int32_t, std::ratio<1>> seconds_t;
```
arvidn: these days this spells: ``using seconds_t = std::chrono::duration<std::int32_t>;`` (I believe ``ratio<1>`` is the default)

arvidn: however, I think perhaps ``seconds32`` would be a better name

include/libtorrent/torrent.hpp
```diff
@@ -478,26 +480,21 @@ namespace libtorrent
...
 		void force_recheck();
 		void save_resume_data(int flags);
 
-		bool need_save_resume_data() const
-		{
-			// save resume data every 15 minutes regardless, just to
-			// keep stats up to date
-			return m_need_save_resume_data || m_ses.session_time() - m_last_saved_resume > 15 * 60;
-		}
+		bool need_save_resume_data() const;
```
arvidn: this seems like an unrelated change. I'm not opposed to it, it just makes it a bit harder to get to the core of the patch. Did this function become more complicated, is that why you moved it?

include/libtorrent/torrent.hpp
```diff
@@ -1300,9 +1297,9 @@ namespace libtorrent
 		// m_num_verified = m_verified.count()
 		std::uint32_t m_num_verified = 0;
 
-		// this timestamp is kept in session-time, to
-		// make it fit in 16 bits
-		std::uint16_t m_last_saved_resume;
+		// TODO: waste of memomry because has a max session time offset of
+		// optimistic_disk_retry
```
arvidn: I don't think you need this comment

simulation/setup_swarm.cpp
arvidn: as my comments may have suggested. It's not obvious to me exactly what this test mode does or why it's necessary. I get the impression there may be a better name for it. Is the main difference that this doesn't automatically end the test when the other peers are seeds?

src/torrent.cpp
```diff
@@ -121,9 +124,12 @@ namespace libtorrent
 		}
 		return ret;
 	}
-
 	constexpr int default_piece_priority = 4;
 
+	// holds seed rank priority duration in the ratio type which is used at
+	// seed rank calculation to prevent multible conversion and constructions
+	static constexpr seconds m_seed_rank_priority_duration = seconds(30 * 60);
```
arvidn: please remove ``static``. It's redundant since this definition is already in the anonymous namespace.
I use the ``m_`` prefix for members. This is not a member so it should not have the prefix.
Instead of calculating the number of second in 30 minutes, I think it would make more sense to be explicit:

``constexpr minutes seed_rank_priority_duration{30};``

src/torrent.cpp
```diff
@@ -8547,6 +8527,16 @@ namespace libtorrent
 		stop_announcing();
 	}
 
+	bool torrent::need_save_resume_data() const
+	{
+		// save resume data every 15 minutes regardless, just to
+		// keep stats up to date
+		return m_need_save_resume_data ||
+			// cast ratio<1l, 1000000000l> to seconds_t ratio
```
arvidn: this comment seems more confusing than useful

src/torrent.cpp
```diff
@@ -8873,31 +8863,39 @@ namespace libtorrent
...
+	seconds_t torrent::finished_time() const
 	{
-		// m_finished_time does not account for the current "session", just the
-		// time before we last started this torrent. To get the current time, we
-		// need to add the time since we started it
-		return m_finished_time + ((!is_finished() || is_paused()) ? 0
-			: (m_ses.session_time() - m_became_finished));
+		if(!is_finished() || is_paused())
+			return m_finished_time;
+		// cast ratio<1l, 1000000000l> to seconds_t ratio
```
arvidn: I think you should remove this comment too. that ratio you specify is implementation defined and not really relevant here.
This applies the the same comments below as well

src/torrent.cpp
```diff
@@ -8920,11 +8918,13 @@ namespace libtorrent
...
 		// leave upload mode every 10 minutes hoping that the error
 		// condition has been fixed
-		if (m_upload_mode && m_auto_managed
-			&& int(m_ses.session_time() - m_upload_mode_time)
-			>= settings().get_int(settings_pack::optimistic_disk_retry))
+		if (m_upload_mode && m_auto_managed)
 		{
-			set_upload_mode(false);
+			// cast ratio<1l, 1000000000l> to seconds_t ratio
+			std::int32_t upload_mode_duration = duration_cast<seconds_t>(
```
arvidn: maybe there should be a ``torrent::upload_mode_time()``, like there there is for seeding_time, active_time and finished_time. Then this expression would be a lot more intuitive, and it could probably be moved back into the if-condition

src/torrent.cpp
```diff
@@ -10558,22 +10558,23 @@ namespace libtorrent
 
 		// activity time
 #ifndef TORRENT_NO_DEPRECATE
-		st->finished_time = finished_time();
-		st->active_time = active_time();
-		st->seeding_time = seeding_time();
+		// cast to seconds in case that internal values doesn't have ratio<1>
+		st->finished_time = int(duration_cast<seconds>(finished_time()).count());
```
arvidn: if you duration_cast this to the type you defined above, ``seconds_t``, you don't need to wrap it in a cast to ``int``

include/libtorrent/torrent.hpp
Falcosc: That means you don't want to change it in the next optimization step? My idea was to put a todo on each position where I did at least double the memory usage.

arvidn: I'm not sure it's important to do that. but a better place to put such comment, I think, is: https://github.com/arvidn/libtorrent/projects/1

bindings/python/test.py
Falcosc: And what should I do if I got a 1 sec increment between befor and after start? I would use after start, but what to do if I got a second increment before and after add torrent? I just don't know the time. But at least session start takes less then 1 sec its reliable.
Or I could calculate a time diff and write assert(diff < seconds(2))

include/libtorrent/torrent.hpp
Falcosc: yes, we need to cast it because the diff duration type does not match to seconds32

simulation/setup_swarm.cpp
Falcosc: It does allow uploading state for more then 88 ticks

src/torrent.cpp
Falcosc: With minutes we would need a conversion on runtime, I guess the compiler is not smart enoght to move an inline cast<seconds>(seed_rank_priority_duration) to a static seconds duration, isn't it?

src/torrent.cpp
Falcosc: you did ask why it's neccesary to cast a duration between 2 timepoints to seconds. Feel free to suggest another comment.

src/torrent.cpp
Falcosc: no, that would be wrong if we change the rate to non 1.

The interface except real seconds, that why I cast to real "seconds" instead of seconds32. Like the comment already tell.

src/torrent.cpp
arvidn: you can't compare ``chrono::seconds`` < ``chrono::minutes``? I wouldn't expect you to need an explicit cast for that.

src/torrent.cpp
arvidn: I don't think the answer to that question belongs in the source code but rather as a reply to the question

src/torrent.cpp
arvidn: that's true

src/torrent.cpp
Falcosc: Even if comparation operators are able to handle it, it would still calculate a value with same base every time instead of one time on construct.

include/libtorrent/torrent.hpp
Falcosc: No it doesn't work
```
include/libtorrent/torrent.hpp:98:21: error: a template-id may not appear in a using-declaration
  using std::chrono::duration<std::int32_t> seconds32;
```

aldenml:
this comment `// TODO: this setup takes 1 minute execution time` alone makes me wonder if the test is a sound one. If there is a fundamental limitation in the simulation framework I think that warrants a separated analysis and corresponding fix.

btw, did you look at `aux::update_time_now()`? (just asking)

include/libtorrent/torrent.hpp
arvidn: I believe you forgot the ``=``-sign

arvidn:
yeah, without having looked at this, I think the test takes long because:

1. there are multiple sessions running, with torrents. this means a fair amount of work
2. the simulation framework should have a mechanism to advance time. normally the advancement of time is driven by timers and packet events, but it would be possible to add a way to advance time an hour at a time, to significantly cut down on the amount of redundant processing happening during the test.
3. I'm sure there are a few opportunities for optimizations as well, that could make it run faster

Falcosc:
Yes, currenlty its 1 minute 100% CPU Load and a optimazasion could be done if we spend 4-8 hours more hours and try to do all this checks with less ticks. Because we have resume and pause it could even work with 88 ticks limit.

Falcosc:
I would like to make a cut here. Is there anything which need to be fixed in this request?

But I see there is maybe a bug in test setup because the peers does not download sometimes: https://ci.appveyor.com/project/arvidn/libtorrent/build/5124/job/jx2rxk820d426gb8

src/torrent.cpp
```diff
@@ -8288,8 +8260,16 @@ namespace libtorrent
...
+		const seconds32 act_time = active_time();
+		static_assert(std::is_same<
+			decltype(seed_rank_priority_duration), decltype(act_time)>::value,
+			"Need to be same ratio type to prevent runtime conversion");
+
+		// use raw duration values because active time and finish time use same
+		// ratio type and don't need to be converted to a equal ratio
+		// and we don't compare them agains seconds we just use fraction value
+		std::int32_t const fin_time = finished_time().count();
+		std::int32_t const download_time = act_time.count() - fin_time;
```
arvidn: this code looks suspicious. that static assert does not look right. it seems like turning ``fin_time`` and ``download_time`` into ``duration`` types seems like the most reasonable. that way you get type-safety and nobody needs to stare at this code to be convinced it's correct.

the comparison of ``decltype()`` to make sure the ratio is the same is not clear (and not quite correct). you should just look at the actual ratio, regardless of the other types that go into a duration.

arvidn: Is it not possible to multiply and divide a ``duration`` by a scalar? like below?

src/torrent.cpp
```diff
@@ -8547,6 +8530,15 @@ namespace libtorrent
 		stop_announcing();
 	}
 
+	bool torrent::need_save_resume_data() const
+	{
+		// save resume data every 15 minutes regardless, just to
+		// keep stats up to date
+		return m_need_save_resume_data ||
+			duration_cast<seconds32>(aux::time_now() - m_last_saved_resume)
+			> seconds32(15 * 60);
```
arvidn: It would be nice to say ``minutes32(15)`` here

src/torrent.cpp
```diff
@@ -8547,6 +8530,15 @@ namespace libtorrent
 		stop_announcing();
 	}
 
+	bool torrent::need_save_resume_data() const
+	{
+		// save resume data every 15 minutes regardless, just to
+		// keep stats up to date
+		return m_need_save_resume_data ||
+			duration_cast<seconds32>(aux::time_now() - m_last_saved_resume)
```
arvidn: what happens if you don't ``duration_cast`` here?

Falcosc: boolean compare should be work without cast.

src/torrent.cpp
Falcosc: I don't want to divide inside of sort code, because it's not neccesary if we only need a ration between 2 values of equal type.
Idea is not too drop the performance. If I implement duration here it would maybe slower then before.

If you don't care about it I could just use durations instead of count() + compile time typecheck.

arvidn: I would even bet there is no performance difference.

arvidn: I don't see how you avoid the division. I mean, without caching the ratio in the peer_connection object

include/libtorrent/torrent.hpp
```diff
@@ -95,6 +95,8 @@ namespace libtorrent
 	class bt_peer_connection;
 	struct listen_socket_t;
 
+	typedef std::chrono::duration<std::int32_t> seconds32;
```
arvidn: if you'll make another change, please make this: ``using seconds32 = std::chrono...``

src/torrent.cpp
```diff
@@ -121,9 +124,12 @@ namespace libtorrent
 		}
 		return ret;
 	}
-
 	constexpr int default_piece_priority = 4;
 
+	// seed rank priority duration which need to match ratio type which is used at
+	// seed rank calculation to prevent runtime conversion and constructions
+	constexpr seconds32 seed_rank_priority_duration = seconds32(30 * 60);
```
arvidn: it would be nice to say ``minutes32`` here too :)

src/torrent.cpp
Falcosc: currently it does avoid the division because everthing has same duration ratio of type seconds32. And the compile time type check does test it.

This is the critical expression
`fin_time * 100 / download_time < s.get_int(settings_pack::seed_time_ratio_limit)`

for example 20s finish time and 20s download time and seed_time_ratio_limit 100
if we use a duration ratio 5s per count we would have
`int 4 * 100 / int 4 < s.get_int(100)`

if we use duration it would look like
`int(4*5) * 100 / int(4*5) < s.get_int(100)`
with a performance drop because the calclation is done for each element.
Oh pardon, no you are right it would be
`duration(internal4) * 100 / duration(internal4) < s.get_int(100)`
As long as the types are same duration it would be no performance drop, only some internal ratio checks

src/torrent.cpp
Falcosc: I will move it back to inline because the performance gain is not worth the extra 5 lines. I gues compiler is smart enoght to create this constexpr out of an inline class construct.

src/torrent.cpp
arvidn: right, it would still be possible to static_assert that they had the same ratio, but I think that's unnecessary. If they end up being different and it's a performance issue, it will show up in the profile and it can be addressed

bindings/python/test.py
Falcosc: @arvidn is that ok as it is or should I change it to assert(diff < seconds(2)) (2 = is max session start time 1s + 1s safety for secound wrap)  to be able to remove the sleep

simulation/test_torrent_status.cpp
```diff
@@ -41,11 +41,161 @@ using namespace sim;
...
+				TEST_CHECK(st.seeding_duration == expected_active_duration);
+				TEST_CHECK(st.finished_duration == expected_active_duration);
+				// does not download in seeding mode
+				TEST_CHECK(st.last_download == start_time);
+				// TODO: write a converter for TEST_EQUAL from duration to rep
+				// convert like this: return duration_cast<seconds>(arg1).count();
+				// test.hpp: error: no match for ‘operator<<’ operand types are
+				// std::basic_ostream<char> and std::chrono::duration<long int>
+				// then we can remove the printf statements
+				printf("expected_active_duration: %" PRId64 " \n",
```
arvidn: I think this printing should be removed to cut down the test a bit

simulation/test_torrent_status.cpp
```diff
@@ -41,11 +41,161 @@ using namespace sim;
...
+					break;
+				case 64061:
+					// pause after some some upload ticks
+					// not every tick is an upload because of the speed limit
+					handle.pause();
+					tick_is_in_active_range = false;
+					last_active_time = lt::clock_type::now();
+					break;
+				case 70001:
+					// resum after the time shift to figure out if an upload time
```
arvidn: *resume

simulation/test_torrent_status.cpp
```diff
@@ -41,11 +41,161 @@ using namespace sim;
...
-TORRENT_TEST(status_timers)
+// it is a long running test, we did already try to improve it:
+// - less slow peers vs many fast peers
+// - slow peers with smal torrent vs fast peers with large torrent
+// - less peers with large torrents vs more peers with smal torrent
+// - connection limit vs parallel requests
+// as conlusion the following combination take the least amount of
+// execution time per tick:
+// - many slow peers with small torrent and connection limit
+TORRENT_TEST(status_timers_time_shift_with_paused_torrent)
```
arvidn: since it takes a minute to run, I think this test should be disabled for now. I can take a look at speeding it up later

simulation/test_torrent_status.cpp
```diff
@@ -41,11 +41,161 @@ using namespace sim;
 namespace lt = libtorrent;
 
 // this is a test for torrent_status time counters are correct
-TORRENT_TEST(status_timers)
```
arvidn: would you mind leaving the old test in, since it's cheap to run?

simulation/test_torrent_status.cpp
```diff
@@ -59,44 +209,240 @@ TORRENT_TEST(status_timers)
...
+					// torrent get ready for seeding on first tick, means time +1s
+					tick_is_in_active_range = true;
+					break;
+				case 1:
+					// pause after we did have the first upload tick
+					handle.pause();
+					tick_is_in_active_range = false;
+					break;
+				case 64000:
+					// resum just before we hit the time shift handling
```
arvidn: *resume

simulation/test_torrent_status.cpp
```diff
@@ -59,44 +209,240 @@ TORRENT_TEST(status_timers)
...
+				TEST_CHECK(st.active_duration == expected_active_duration);
+				TEST_CHECK(st.seeding_duration == expected_active_duration);
+				TEST_CHECK(st.finished_duration == expected_active_duration);
+				// does not download in seeding mode
+				TEST_CHECK(st.last_upload == start_time);
+				// does not upload without peers
+				TEST_CHECK(st.last_download == start_time);
+
+				// TODO: write a converter for TEST_EQUAL from duration to rep
+				printf("expected_active_duration: %" PRId64 " \n",
```
arvidn: this printing should be removed before merged

simulation/test_torrent_status.cpp
```diff
@@ -59,44 +209,240 @@ TORRENT_TEST(status_timers)
...
+					// 16 bits counting seconds
+					ran_to_completion = true;
+					return true;
+			}
+
+			// verify that the timers seem correct
+			if (tick_is_in_active_range && (ticks % 3600) == 0)
+			{
+				torrent_status st = handle.status();
+				TEST_CHECK(st.active_duration == expected_active_duration);
```
arvidn: I noticed there's a lot of repetition in these tests. can you think of reasonable refactor-points to share some more logic between them?

ChangeLog
```diff
@@ -1,6 +1,7 @@
+	* fix invalid torrent state finish_duration calculation
```
arvidn: what error was this again?

simulation/test_torrent_status.cpp
Falcosc: I will try it, at least I have to refactor all the new tests because at somepoint the got messed without my notice. I just didn't understand when this did happen.

I guess whole part of them was only an unfinished draft and after some days I may find them in my repo and did check them in because the did look finish and the local build was sucessfull.
Yesterday I try to finish this branch and found out that nothing does work.
 For example the paused clamp test would never work because on pause there is no seeding.

ChangeLog
Falcosc: Finishdate did use a wrong base for calculation, I will try to find it again

ChangeLog
```diff
@@ -1,6 +1,7 @@
...
+	* fix last_upload and last_download overflow after 9 hours in past
 	* introduce introduce distinct types for peer_class_t, piece_index_t and
 	  file_index_t.
-	* fix crash caused by empty bitfield
```
arvidn: why was this removed? is it in the upstream branch?

ChangeLog
Falcosc: @arvidn here it is
https://github.com/arvidn/libtorrent/pull/1492/commits/c71611785d90ba301746b26725e40f93d12d7dfe#diff-19929347e18a6ff1fc37f09ec720306bL8259

there was a failure at timebase shift
You may want to backport this fix to RC_1_1

arvidn:
did you rebase? all my review comments are gone, and there were some unanswered questions in there

Falcosc:
I did confirm all your comments by doing what you did suggest, but did not answer them.
If Travis gives an ok, my work is done.

I don't test last_uploaded overflow anymore, I reduce this test to a 10 ticks short test. Because we don't need to test overflow anymore. But I did keep the 4 other overflow tests, each of them runs about 10s. In summary it does still take some time, but before it took 3 times longer.

ChangeLog
Falcosc: was a mistake, is now fixed

simulation/test_torrent_status.cpp
Falcosc: I don't want to share the case logic because that is the real usecase definition. And some of the asserts depend on the usecase. Overall I feel like the overhead would not improve the readablity because there are not enoght things to share.

simulation/test_torrent_status.cpp
Falcosc: is now fixed

simulation/test_torrent_status.cpp
Falcosc: is now fixed

simulation/test_torrent_status.cpp
Falcosc: I did include the old test

simulation/test_torrent_status.cpp
Falcosc: is now fixed

simulation/test_torrent_status.cpp
Falcosc: is now fixed

include/libtorrent/torrent.hpp
Falcosc: is now fixed

Falcosc: is now fixed

src/torrent.cpp
Falcosc: is now fixed

Falcosc:
I did answer all your comments to make it more easy for you.

Falcosc:
I guess you did disable parallel builds as workarround to fix test failures?
I like it, even if it take longer to get a build slot.

arvidn:
no, I think travis is just running at reduced capacity

arvidn:
ok. looks good to me. the last thing though, if the test still takes a minute to run, I think it should be disabled for now. I can look at ways to speed it up later. as far as I could see, it's still active.

you could just comment it out for now

Falcosc:
each test runs about 10s. All overflow tests are passive ones without downloaders and the active test with downloader is build to be  done after 10 ticks to save cpu time.

PR #1495, opened 2017-01-04
consolidate windows crypto provider implementation into single header

ajax16384:


include/libtorrent/aux_/win_crypto_provider.hpp
```diff
@@ -0,0 +1,162 @@
...
+				std::terminate();
+#endif
+			}
+		}
+	};
+
+	template <ALG_ID AlgId, DWORD ProviderType>
+	struct crypt_hash: crypt_provider<ProviderType>
+	{
+		void create()
```
arvidn: it seems like ``create()`` and ``destroy`` should just be the constructor and destructor.

include/libtorrent/aux_/win_crypto_provider.hpp
```diff
@@ -0,0 +1,162 @@
...
+#ifndef BOOST_NO_EXCEPTIONS
+				throw system_error(error_code(GetLastError(), system_category()));
+#else
+				std::terminate();
+#endif
+			}
+		}
+	};
+
+	template <ALG_ID AlgId, DWORD ProviderType>
```
arvidn: ``AlgId`` doesn't need to be a template parameter. just pass it in to the constructor.

include/libtorrent/aux_/win_crypto_provider.hpp
```diff
@@ -0,0 +1,162 @@
...
+#include "libtorrent/config.hpp"
+#include "libtorrent/error_code.hpp"
+
+#include <windows.h>
+#include <wincrypt.h>
+
+namespace libtorrent { namespace aux
+{
+	template <DWORD ProviderType>
+	struct crypt_provider
```
arvidn: There's no point in this being a class. in fact, ``acquire_provider`` shouldn't be a template at all

include/libtorrent/aux_/win_crypto_provider.hpp
```diff
@@ -0,0 +1,162 @@
...
+			return ret;
+		}
+		static HCRYPTPROV get_provider()
+		{
+			static HCRYPTPROV provider = acquire_provider(ProviderType);
+			return provider;
+		}
+	};
+
+	struct crypt_random: crypt_provider<PROV_RSA_FULL>
```
arvidn: there's no point in this being a class either. ``generate()`` should just be a free function

include/libtorrent/aux_/win_crypto_provider.hpp
```diff
@@ -0,0 +1,162 @@
...
+				throw system_error(error_code(GetLastError(), system_category()));
+#else
+				std::terminate();
+#endif
+			}
+		}
+	};
+
+	template <ALG_ID AlgId, DWORD ProviderType>
+	struct crypt_hash: crypt_provider<ProviderType>
```
arvidn: I don't think ``crypt_provider`` should be a class and I don't think inheriting is appropriate here

include/libtorrent/aux_/win_crypto_provider.hpp
```diff
@@ -0,0 +1,162 @@
...
+			m_initialized = true;
+		}
+
+		void destroy()
+		{
+			if (m_initialized) CryptDestroyHash(m_hash);
+			m_initialized = false;
+		}
+
+		void duplicate(crypt_hash source)
```
arvidn: this looks like it should just be a copy constructor

arvidn: but you need to take the other ``crypt_hash`` by ``const&``

include/libtorrent/hasher512.hpp
```diff
@@ -114,7 +113,7 @@ namespace libtorrent
 #elif TORRENT_USE_COMMONCRYPTO
 		CC_SHA512_CTX m_context;
 #elif TORRENT_USE_CRYPTOAPI_SHA_512
-		HCRYPTHASH m_context;
+		aux::crypt_hash<CALG_SHA_512, PROV_RSA_AES> m_context;
```
arvidn: to pass arguments into the constructor, you do this: ``... m_context{CALC_SHA_512};``

src/hasher.cpp
```diff
@@ -77,14 +49,7 @@ namespace libtorrent
...
 #elif TORRENT_USE_CRYPTOAPI
-		if (CryptCreateHash(get_crypt_provider(), CALG_SHA1, 0, 0, &m_context) == false)
-		{
-#ifndef BOOST_NO_EXCEPTIONS
-			throw system_error(error_code(GetLastError(), system_category()));
-#else
-			std::terminate();
-#endif
-		}
+		m_context.create();
```
arvidn: if you make ``create``, ``destroy`` and ``duplicate`` the constructor, destructor and copy constructor/copy assignment, you don't need to make these calls.

src/hasher.cpp
```diff
@@ -121,28 +86,14 @@ namespace libtorrent
 #elif TORRENT_USE_CRYPTOAPI
 	hasher::hasher(hasher const& h)
```
arvidn: in fact, you probably don't need to define the copy constructor explicitly here. but if you do, you can just define it as ``= default;``

src/hasher.cpp
```diff
@@ -121,28 +86,14 @@ namespace libtorrent
...
-#ifndef BOOST_NO_EXCEPTIONS
-			throw system_error(error_code(GetLastError(), system_category()));
-#else
-			std::terminate();
-#endif
-		}
+		m_context.duplicate(h.m_context);
 	}
 
 	hasher& hasher::operator=(hasher const& h)
```
arvidn: same thing for assignment

include/libtorrent/aux_/win_crypto_provider.hpp
ajax16384: ok

include/libtorrent/aux_/win_crypto_provider.hpp
ajax16384: ok

include/libtorrent/aux_/win_crypto_provider.hpp
ajax16384: ok

include/libtorrent/aux_/win_crypto_provider.hpp
ajax16384: ok

src/hasher.cpp
ajax16384: ok

src/hasher.cpp
ajax16384: ok

include/libtorrent/hasher512.hpp
ajax16384: unfortunately AlgId should be used not only at constructor routine

include/libtorrent/aux_/win_crypto_provider.hpp
```diff
@@ -0,0 +1,170 @@
...
+#include "libtorrent/config.hpp"
+#include "libtorrent/error_code.hpp"
+
+#include <windows.h>
+#include <wincrypt.h>
+
+namespace libtorrent { namespace aux
+{
+
+	static HCRYPTPROV crypt_acquire_provider(DWORD provider_type)
```
arvidn: this function should not be ``static``. it should either be ``inline`` or live in a cpp file

include/libtorrent/aux_/win_crypto_provider.hpp
```diff
@@ -0,0 +1,170 @@
...
+#ifndef BOOST_NO_EXCEPTIONS
+			throw system_error(error_code(GetLastError(), system_category()));
+#else
+			std::terminate();
+#endif
+		}
+		return ret;
+	}
+
+	static void crypt_gen_random(span<char> buffer)
```
arvidn: the same applies to this function

include/libtorrent/aux_/win_crypto_provider.hpp
```diff
@@ -0,0 +1,170 @@
...
+#ifndef BOOST_NO_EXCEPTIONS
+				throw system_error(error_code(GetLastError(), system_category()));
+#else
+				std::terminate();
+#endif
+			}
+			TORRENT_ASSERT(size == DWORD(digest_size));
+		}
+	private:
+		void create()
```
arvidn: this function is only called from the constructor. is there a reason to not just make it be the constructor?

include/libtorrent/aux_/win_crypto_provider.hpp
```diff
@@ -0,0 +1,170 @@
...
+		}
+
+		HCRYPTPROV get_provider()
+		{
+			static HCRYPTPROV provider = crypt_acquire_provider(ProviderType);
+			return provider;
+		}
+
+		HCRYPTHASH m_hash;
+		bool m_initialized = false;
```
arvidn: m_initialized is only used to assert that the constructor has been run and the destructor has not been run. this is generally a safe assumption, short of memory corruption. if you really think this assert is important, the state variable should probably be wrapped in ``#if TORRENT_USE_ASSERTS``. The problem with that though is that it changes the ABI for the different build options

src/hasher.cpp
```diff
@@ -77,14 +49,6 @@ namespace libtorrent
 #elif TORRENT_USE_COMMONCRYPTO
 		CC_SHA1_Init(&m_context);
 #elif TORRENT_USE_CRYPTOAPI
```
arvidn: this preprocessor statement is pointless now, since there's no code in here. you might as well remove it

src/hasher512.cpp
```diff
@@ -77,14 +49,6 @@ namespace libtorrent
 #elif TORRENT_USE_COMMONCRYPTO
 		CC_SHA512_Init(&m_context);
 #elif TORRENT_USE_CRYPTOAPI_SHA_512
```
arvidn: this preprocessor condition could also be removed

include/libtorrent/aux_/win_crypto_provider.hpp
ajax16384: ok

include/libtorrent/aux_/win_crypto_provider.hpp
ajax16384: ok

include/libtorrent/aux_/win_crypto_provider.hpp
ajax16384: ok

src/hasher512.cpp
```diff
@@ -203,15 +122,7 @@ namespace libtorrent
...
-		CryptDestroyHash(m_context);
-		if (CryptCreateHash(get_crypt_provider(), CALG_SHA_512, 0, 0, &m_context) == false)
-		{
-#ifndef BOOST_NO_EXCEPTIONS
-			throw system_error(error_code(GetLastError(), system_category()));
-#else
-			std::terminate();
-#endif
-		}
+		m_context = decltype(m_context)();
```
arvidn: this looks suspicious. this will create a hash context, then duplicate it and then tear down the first one. maybe there should just be a ``reset()`` member function on m_context

ajax16384: ok

src/hasher.cpp
ajax16384: last #else contains fallback to builtin crypto routine, so this #elif can not be removed

include/libtorrent/aux_/win_crypto_provider.hpp
ajax16384: Win SDK has no "INVALID_HASH_VALUE" constant for HCRYPTHASH, so m_initialized was introduced to protect calling of "destroy()" twice with same value. e.g.:
````c++
crypt_hash& crypt_hash::operator=(crypt_hash const& h) {
  if (this == &h) return *this;
  destroy(); // 1st time destroy
  duplicate(h); // throws exception, current m_hash has invalid value, ~crypt_hash must not call m_hash for 2nd time
  return *this;
}
````
i think this check goes beyond assert level and must be leaved as is

Comment (arvidn) [inline include/libtorrent/aux_/win_crypto_provider.hpp]: but this class is not movable and the default constructor creates a context. There is no way (as far as I can tell) to have the destructor be called on an object where m_hash has not been constructed.

File: include/libtorrent/aux_/win_crypto_provider.hpp (hunk 13 above)
Comment (arvidn) [inline include/libtorrent/aux_/win_crypto_provider.hpp]: yeah, that probably makes sense

File: include/libtorrent/aux_/win_crypto_provider.hpp (hunk 14 above)
Comment (arvidn) [inline include/libtorrent/aux_/win_crypto_provider.hpp]: so.. just don't call destroy() there.. as far as I can tell, there is no way to have an object that has _not_ been initialized (this is kind of the point of constructors in C++). That means, you can always rely on destroy() being safe to call _once_ as long as you restore the invariant by re-creating it.

File: include/libtorrent/aux_/win_crypto_provider.hpp [hunk 18]
```diff
@@ -123,40 +121,37 @@ namespace libtorrent { namespace aux
...
 			}
-			m_initialized = true;
+			return ret;
 		}
 
 		void destroy()
 		{
-			if (m_initialized) CryptDestroyHash(m_hash);
-			m_initialized = false;
+			CryptDestroyHash(m_hash);
```
arvidn: how about replacing all calls to ``destroy()`` with calls to ``CryptDestroyHash(m_hash);``? (and removing this function)

arvidn:
looks good

include/libtorrent/aux_/win_crypto_provider.hpp
ajax16384: ok

PR #1496, opened 2017-01-05
fix frequent dealloc/alloc for utp socket lists

ajax16384:
old implementation spent too much cpu time at std::allocator<libtorrent::utp_socket_impl * >: *

### DIFF
--- a/include/libtorrent/utp_socket_manager.hpp
+++ b/include/libtorrent/utp_socket_manager.hpp
@@ -138,6 +138,8 @@ namespace libtorrent
 		// an ack for every single packet
 		socket_vector_t m_deferred_acks;
 
+		bool m_in_socket_drained = false;
+
 		// sockets that have received or sent packets this
 		// round, may subscribe to the event of draining the
 		// UDP socket. At that point they may call the
@@ -149,6 +151,8 @@ namespace libtorrent
 		// underlying socket. They are notified when the socket
 		// becomes writable again
 		socket_vector_t m_stalled_sockets;
+
+		bool m_in_writable = false;
 
 		// the last socket we received a packet on
 		utp_socket_impl* m_last_socket = nullptr;
--- a/include/libtorrent/utp_socket_manager.hpp
+++ b/include/libtorrent/utp_socket_manager.hpp
@@ -130,23 +130,28 @@ namespace libtorrent
 		typedef std::multimap<std::uint16_t, utp_socket_impl*> socket_map_t;
 		socket_map_t m_utp_sockets;
 
+		typedef std::vector<utp_socket_impl*> socket_vector_t;
+
 		// this is a list of sockets that needs to send an ack.
 		// once the UDP socket is drained, all of these will
 		// have a chance to do that. This is to avoid sending
 		// an ack for every single packet
-		std::vector<utp_socket_impl*> m_deferred_acks;
+		socket_vector_t m_deferred_acks;
+		socket_vector_t m_deferred_acks_tock;
 
 		// sockets that have received or sent packets this
 		// round, may subscribe to the event of draining the
 		// UDP socket. At that point they may call the
 		// user callback function to indicate bytes have been
 		// sent or received.
-		std::vector<utp_socket_impl*> m_drained_event;
+		socket_vector_t m_drained_event;
+		socket_vector_t m_drained_event_tock;
 
 		// list of sockets that received EWOULDBLOCK from the
 		// underlying socket. They are notified when the socket
 		// becomes writable again
-		std::vector<utp_socket_impl*> m_stalled_sockets;
+		socket_vector_t m_stalled_sockets;
+		socket_vector_t m_stalled_sockets_tock;
 
 		// the last socket we received a packet on
 		utp_socket_impl* m_last_socket = nullptr;
--- a/include/libtorrent/utp_socket_manager.hpp
+++ b/include/libtorrent/utp_socket_manager.hpp
@@ -139,11 +139,8 @@ namespace libtorrent
 		socket_vector_t m_deferred_acks;
 
 		// storage used for saving cpu time on "push_back"
-		// by using already pre-allocated vectors
-		socket_vector_t m_socket_drained_temp;
-
-		// prevents reentrant loop
-		bool m_in_socket_drained = false;
+		// by using already pre-allocated vector
+		socket_vector_t m_temp_sockets;
 
 		// sockets that have received or sent packets this
 		// round, may subscribe to the event of draining the
@@ -156,9 +153,6 @@ namespace libtorrent
 		// underlying socket. They are notified when the socket
 		// becomes writable again
 		socket_vector_t m_stalled_sockets;
-
-		socket_vector_t m_writable_temp;
-		bool m_in_writable = false;
 
 		// the last socket we received a packet on
 		utp_socket_impl* m_last_socket = nullptr;
--- a/include/libtorrent/utp_socket_manager.hpp
+++ b/include/libtorrent/utp_socket_manager.hpp
@@ -130,14 +130,13 @@ namespace libtorrent
 		typedef std::multimap<std::uint16_t, utp_socket_impl*> socket_map_t;
 		socket_map_t m_utp_sockets;
 
-		typedef std::vector<utp_socket_impl*> socket_vector_t;
+		using socket_vector_t = std::vector<utp_socket_impl*>;
 
 		// this is a list of sockets that needs to send an ack.
 		// once the UDP socket is drained, all of these will
 		// have a chance to do that. This is to avoid sending
 		// an ack for every single packet
 		socket_vector_t m_deferred_acks;
-		socket_vector_t m_deferred_acks_tock;
 
 		// sockets that have received or sent packets this
 		// round, may subscribe to the event of draining the
@@ -145,13 +144,11 @@ namespace libtorrent
 		// user callback function to indicate bytes have been
 		// sent or received.
 		socket_vector_t m_drained_event;
-		socket_vector_t m_drained_event_tock;
 
 		// list of sockets that received EWOULDBLOCK from the
 		// underlying socket. They are notified when the socket
 		// becomes writable again
 		socket_vector_t m_stalled_sockets;
-		socket_vector_t m_stalled_sockets_tock;
 
 		// the last socket we received a packet on
 		utp_socket_impl* m_last_socket = nullptr;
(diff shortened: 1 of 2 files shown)


arvidn:
what's your observation motivating this change?

ajax16384:
````c++
 - libtorrent::utp_socket_manager::incoming_packet	5.10 %	0.10 %	1735	34
   - libtorrent::utp_incoming_packet	4.68 %	0.01 %	1592	3
     - libtorrent::utp_socket_impl::incoming_packet	4.61 %	0.12 %	1568	40
       - std::vector<libtorrent::utp_socket_impl *,std::allocator<libtorrent::utp_socket_impl *> >::push_back	2.21 %	0.01 %	753	2
         - std::vector<libtorrent::utp_socket_impl *,std::allocator<libtorrent::utp_socket_impl *> >::_Reserve	2.20 %	0.02 %	749	6
           - std::vector<libtorrent::utp_socket_impl *,std::allocator<libtorrent::utp_socket_impl *> >::_Reallocate	2.16 %	0.02 %	734	7
             + std::allocator<libtorrent::utp_socket_impl *>::allocate	1.84 %	0.01 %	626	2
             + std::allocator<libtorrent::utp_socket_impl *>::deallocate	0.29 %	0.00 %	100	1


     - libtorrent::aux::session_impl::on_udp_packet 20.11 % 0.08 %  6845  26
       - libtorrent::utp_socket_manager::socket_drained 5.79 %  0.03 %  1971  9
         + libtorrent::utp_socket_impl::send_pkt  4.30 %  0.09 %  1463  30
         + std::allocator<libtorrent::utp_socket_impl *>::deallocate  1.42 %  0.01 %  483 4
````
old implemetation had frequent alloc from system heap for push_back(utp_socket_impl *)
new one use capacity of persistent std::vector

File: include/libtorrent/utp_socket_manager.hpp [hunk 1]
```diff
@@ -130,23 +130,28 @@ namespace libtorrent
 		typedef std::multimap<std::uint16_t, utp_socket_impl*> socket_map_t;
 		socket_map_t m_utp_sockets;
 
+		typedef std::vector<utp_socket_impl*> socket_vector_t;
```
arvidn: please say:

``using socket_vector_t = std::vector<utp_socket_impl*>;``

instead

include/libtorrent/utp_socket_manager.hpp
```diff
@@ -130,23 +130,28 @@ namespace libtorrent
...
+
 		// this is a list of sockets that needs to send an ack.
 		// once the UDP socket is drained, all of these will
 		// have a chance to do that. This is to avoid sending
 		// an ack for every single packet
-		std::vector<utp_socket_impl*> m_deferred_acks;
+		socket_vector_t m_deferred_acks;
+		socket_vector_t m_deferred_acks_tock;
```
arvidn: it's not obvious what ``_tock`` refers to. it's also not obvious why you need it. I take it the purpose is to double buffer..

src/utp_socket_manager.cpp
```diff
@@ -255,10 +255,10 @@ namespace libtorrent
 
 	void utp_socket_manager::writable()
 	{
-		std::vector<utp_socket_impl*> stalled_sockets;
-		m_stalled_sockets.swap(stalled_sockets);
-		for (std::vector<utp_socket_impl*>::iterator i = stalled_sockets.begin()
-			, end(stalled_sockets.end()); i != end; ++i)
+		m_stalled_sockets_tock.clear();
+		m_stalled_sockets.swap(m_stalled_sockets_tock);
+		for (socket_vector_t::iterator i = m_stalled_sockets_tock.begin()
```
arvidn: this should be a range-based for-loop. Something like: ``for (auto const& s : m_stalled_sockets_tock)``.

src/utp_socket_manager.cpp
```diff
@@ -269,19 +269,19 @@ namespace libtorrent
 	{
 		// flush all deferred acks
 
-		std::vector<utp_socket_impl*> deferred_acks;
-		m_deferred_acks.swap(deferred_acks);
-		for (std::vector<utp_socket_impl*>::iterator i = deferred_acks.begin()
-			, end(deferred_acks.end()); i != end; ++i)
+		m_deferred_acks_tock.clear();
+		m_deferred_acks.swap(m_deferred_acks_tock);
+		for (socket_vector_t::iterator i = m_deferred_acks_tock.begin()
```
arvidn: these two should also be range-for loops

include/libtorrent/utp_socket_manager.hpp
arvidn: in fact, all of the ``_tock`` vectors could be collapsed to a single one, called something like ``temp``. The only purpose of it is to make it possible for events happening while iterating over it to add new sockets to the main vector. In fact, that temporary vector could be ``static``, since uTP sockets are single threaded, only one at a time will be using that temporary vector.

include/libtorrent/utp_socket_manager.hpp
ajax16384: ok

src/utp_socket_manager.cpp
ajax16384: ok

src/utp_socket_manager.cpp
ajax16384: ok

src/utp_socket_manager.cpp
```diff
@@ -269,21 +268,19 @@ namespace libtorrent
...
-		for (std::vector<utp_socket_impl*>::iterator i = deferred_acks.begin()
-			, end(deferred_acks.end()); i != end; ++i)
+		for (auto const &s : deferred_acks)
 		{
-			utp_socket_impl* s = *i;
 			utp_send_ack(s);
 		}
 
-		std::vector<utp_socket_impl*> drained_event;
+		static socket_vector_t drained_event;
```
arvidn: is the reason for having 3 different static temporary vectors that they would get mixed up with idle sockets and not provide enough pre-allocated space?

also, I imagine there should be conditions to skip this logic if ``m_deferred_acks`` and ``m_drained_events`` are empty. we don't want to swap in a pre-allocated vector from the static instance into an inactive socket.

ajax16384: you right there is not particular reason to have separate temp storages.

arvidn:
it's not obvious why the previous version would cause that error. there aren't any iterators kept around across the swapping of the vectors, right? I suppose this has chance of working better.

btw, would you mind adding comments there to describe that those static vectors are kept around to save on re-allocations?

arvidn:
it didn't make a difference...

src/utp_socket_manager.cpp
```diff
@@ -255,12 +255,13 @@ namespace libtorrent
...
+
+		static socket_vector_t stalled_sockets;
+		stalled_sockets.clear();
 		m_stalled_sockets.swap(stalled_sockets);
-		for (std::vector<utp_socket_impl*>::iterator i = stalled_sockets.begin()
-			, end(stalled_sockets.end()); i != end; ++i)
+		for (auto const &s : stalled_sockets)
 		{
-			utp_socket_impl* s = *i;
 			utp_writable(s);
```
arvidn: perhaps one of these functions are "reentrant". i.e. utp_socket::writable may be called while calling a handler.

ajax16384:
test_utp runs two session which tries to use same static storage which leads to iterator error 
````c++
	lt::session ses1(pack);
....
	lt::session ses2(pack);
````

File: include/libtorrent/utp_socket_manager.hpp [hunk 7]
```diff
@@ -130,23 +130,35 @@ namespace libtorrent
...
 		// this is a list of sockets that needs to send an ack.
 		// once the UDP socket is drained, all of these will
 		// have a chance to do that. This is to avoid sending
 		// an ack for every single packet
-		std::vector<utp_socket_impl*> m_deferred_acks;
+		socket_vector_t m_deferred_acks;
+
+		// storage used for saving cpu time on "push_back"
+		// by using already pre-allocated vectors
+		socket_vector_t m_socket_drained_temp;
```
arvidn: I was going to suggest going back to a member here. This should fix the problem, since each utp_socket_manager has its own temporary storage.

include/libtorrent/utp_socket_manager.hpp
```diff
@@ -130,23 +130,35 @@ namespace libtorrent
...
+
+		// prevents reentrant loop
+		bool m_in_socket_drained = false;
```
arvidn: you shouldn't need this anymore

include/libtorrent/utp_socket_manager.hpp
```diff
@@ -130,23 +130,35 @@ namespace libtorrent
...
-		std::vector<utp_socket_impl*> m_drained_event;
+		socket_vector_t m_drained_event;
 
 		// list of sockets that received EWOULDBLOCK from the
 		// underlying socket. They are notified when the socket
 		// becomes writable again
-		std::vector<utp_socket_impl*> m_stalled_sockets;
+		socket_vector_t m_stalled_sockets;
+
+		socket_vector_t m_writable_temp;
```
arvidn: I would expect it to be sufficient to just have a single temporary vector in here, that they all can use. The problem, as you pointed out, was that there were more than one libtorrent session going, which means the same static vector will get swapped from multiple threads and not be thread safe. By moving the temporary vector into here, you're back to thread-safety

include/libtorrent/utp_socket_manager.hpp
```diff
@@ -130,23 +130,35 @@ namespace libtorrent
...
+		bool m_in_writable = false;
```
arvidn: you shouldn't need this anymore either

src/utp_socket_manager.cpp
```diff
@@ -255,37 +255,48 @@ namespace libtorrent
 
 	void utp_socket_manager::writable()
 	{
-		std::vector<utp_socket_impl*> stalled_sockets;
-		m_stalled_sockets.swap(stalled_sockets);
-		for (std::vector<utp_socket_impl*>::iterator i = stalled_sockets.begin()
-			, end(stalled_sockets.end()); i != end; ++i)
+		if (m_in_writable) return;
```
arvidn: I don't think this is necessary anymore. It's also not obvious that this is correct (on the off-chance that this would actually get called again during handling). I would prefer to have as simple code as possible, and just not expect re-entrancy

arvidn:
good catch btw. This is shaping up now I think. having a single temp vector and not supporting/expecting re-entrancy I think would be better

ajax16384:
ok

PR #1499, opened 2017-01-07
fix torrent_status test

arvidn:


arvidn:
pulled out from https://github.com/arvidn/libtorrent/pull/1492

arvidn:
it turns out fixing this test uncovered some other issues as well

arvidn:
@Falcosc this patch makes the sim for the timers run to completion, and pass. it doesn't expose the overflow issue you found, but I believe this is a good first step in that direction

arvidn:
fixing the test also exposed errors in the tracker announce interval test which I also had to fix

Falcosc:
I did already notice that you did change the upload simulator tor inifinit runtime. I will use that for the status_timers_time_shift_with_active_torrent test to test the finish timer during overflow

arvidn:
ok, great. This is against RC_1_1, so once I land this, I'll merge it into master too

PR #1500, opened 2017-01-07
fix division by zero when setting tick_interval > 1000

arvidn:


arvidn:
@Falcosc does this change make sense? this makes the test run to completion, testing what it intended to test (without failing). The issue you found though was specifically the last upload/download timestamps, right? (which are a bit more difficult to test)

Falcosc:
Yes looks right but to make sure that the testsetup was correct you should try it out with the wrong code, you should get a SIGFPE and a stacktrace.

arvidn:
I did. actually, I meant to comment on the other patch. the one that fixes the torrent_status test

Falcosc:
Sorry, then I didn't get it, could try it with another wording?

My next task would be to create a test for torrent status with active torrent during timeshift. That would be more easy to test.

PR #1503, opened 2017-01-07
fixing sign-conversion warnings, part 1

aldenml:


include/libtorrent/aux_/vector.hpp
```diff
@@ -52,14 +52,14 @@ namespace libtorrent { namespace aux {
 		{
 			TORRENT_ASSERT(idx >= IndexType(0));
 			TORRENT_ASSERT(idx < end_index());
-			return this->base::operator[](static_cast<underlying_index>(idx));
+			return this->base::operator[](static_cast<std::size_t>(idx));
```
arvidn: I don't think this will work. the conversion operator to the underlying type is explicit, you can only cast to that specific type. you could do two ``static_cast``s though

aldenml: Do you have an idea of why some configurations compile without problems?

include/libtorrent/bencode.hpp
```diff
@@ -163,7 +163,7 @@ namespace libtorrent
 				ret += 2;
 				break;
 			case entry::string_t:
-				ret += write_integer(out, e.string().length());
+				ret += write_integer(out, entry::integer_type(e.string().length()));
```
arvidn: what would you think of giving ``write_integer()`` a similar treatment as the ``write_int*()``? i.e. have built-in conversions from other integral types

include/libtorrent/bencode.hpp
```diff
@@ -328,7 +328,7 @@ namespace libtorrent
 			// ----------------------------------------------
 			// string
 			default:
-				if (is_digit(std::uint8_t(*in)))
+				if (is_digit(char(*in)))
```
arvidn: isn't ``*in`` already a ``char``?

include/libtorrent/bencode.hpp
aldenml: ok

include/libtorrent/bencode.hpp
aldenml: no, it's a template type`InIt`

include/libtorrent/aux_/vector.hpp
arvidn: yeah, if ``deprecated-functions=on`` (which it is by default) the conversion operator is implicit, to be backwards compatible with using ``int`` everywhere

include/libtorrent/bencode.hpp
arvidn: I see. would you mind adding a ``static_assert`` that ``sizeof(*in) == 1`` too then?

include/libtorrent/aux_/vector.hpp
aldenml: thanks!

include/libtorrent/bencode.hpp
aldenml: sure, will do

include/libtorrent/aux_/suggest_piece.hpp
```diff
@@ -61,7 +61,7 @@ struct suggest_piece
 		// back and then reverse the items we put there.
 		for (int i = int(m_priority_pieces.size()) - 1; i >= 0; --i)
 		{
-			piece_index_t const piece = m_priority_pieces[i];
+			piece_index_t const piece = m_priority_pieces[std::size_t(i)];
```
arvidn: in the long run I would prefer to not have to do this. perhaps there could be another ``vector`` wrapper that accepts any integral type as index and hides the cast

include/libtorrent/bt_peer_connection.hpp
```diff
@@ -323,7 +323,7 @@ namespace libtorrent
 				// if we're encrypting this buffer, we need to make a copy
 				// since we'll mutate it
 				std::unique_ptr<char[]> buf(new char[size]);
-				std::memcpy(buf.get(), buffer.get(), size);
+				std::memcpy(buf.get(), buffer.get(), std::size_t(size));
```
arvidn: what do you think about doing this instead?
``std::copy(buffer.get(), buf.get(), buf.get() + size);``

include/libtorrent/bencode.hpp
```diff
@@ -86,9 +86,12 @@ namespace libtorrent
 
 	namespace detail
 	{
-		template <class OutIt>
-		int write_integer(OutIt& out, entry::integer_type val)
+		template <class OutIt, class In, typename Cond
+			= typename std::enable_if<std::is_integral<In>::value>::type>
+		int write_integer(OutIt& out, In data)
 		{
+			entry::integer_type val = entry::integer_type(data);
```
arvidn: you should make this ``const``, and actually, there's a shorter form:

``entry::integer_type const val(data);``

arvidn: hm.. it doesn't look like you're using ``val`` anywhere though. should it perhaps just say:

``TORRENT_ASSERT(data == In(entry::integer_type(data)));``

include/libtorrent/aux_/suggest_piece.hpp
aldenml: I will do it in this PR, it's still of reasonable size

include/libtorrent/bt_peer_connection.hpp
aldenml: better

include/libtorrent/bencode.hpp
aldenml: it's used below, at `for (char const* str = integer_to_str(buf, 21, val);`, I will do the `const` change

aldenml:
what do you think about the changes in `aux::vector`?

include/libtorrent/aux_/suggest_piece.hpp
```diff
@@ -117,7 +118,7 @@ struct suggest_piece
 	// read from disk (and are likely in our read cache).
 	// pieces closer to the end were inserted into the cache more recently and
 	// have higher priority
-	std::vector<piece_index_t> m_priority_pieces;
```
arvidn: seems reasonable

include/libtorrent/aux_/vector.hpp
```diff
@@ -39,11 +39,22 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent { namespace aux {
 
+	template <typename IndexType>
+	struct vector_index_t
```
arvidn: I think this could be a lot simpler. something like this (I may get the syntax wrong):

template <typename T> struct underlying_type { using type = T; };
template <>
template<typename I> struct underlying_type<strong_typedef<I>> { using type = I; };

aldenml: true, I wanted to capture the fact that only integral types are allowed as early as possible, I will change it to this, it will fail with the internal conversion in the `[]` operators.

include/libtorrent/bencode.hpp
```diff
@@ -86,9 +86,12 @@ namespace libtorrent
 
 	namespace detail
 	{
-		template <class OutIt>
-		int write_integer(OutIt& out, entry::integer_type val)
+		template <class OutIt, class In, typename Cond
+			= typename std::enable_if<std::is_integral<In>::value>::type>
+		int write_integer(OutIt& out, In data)
 		{
+			entry::integer_type const val(data);
```
arvidn: looking at this again. it feels like there should be something like this: ``static_assert(sizeof(data) <= 8, "no more than 64 bits supported");``. Maybe it's a bit paranoid though.

arvidn: or I suppose ``8`` should be replaced by ``sizeof(entry::integer_type)``

arvidn:
looks good. do you have any thoughts on the static assert?

aldenml:
good, I'm ok with the `static_assert`, I will add it

arvidn:
lgtm

PR #1504, opened 2017-01-07
simplify disk thread a little bit

arvidn:


arvidn:
@ssiloti does this look reasonable?

ssiloti:
The change to how the first thread is tracked won't work because the first thread to be spawned isn't necessarily the last to stop.

ssiloti:
lgtm

PR #1511, opened 2017-01-09
fixing sign-conversion warnings, part 2

aldenml:


src/broadcast_socket.cpp
```diff
@@ -276,12 +276,12 @@ namespace libtorrent
 		for (auto& s : m_unicast_sockets)
 		{
 			if (!s.socket) continue;
-			s.socket->send_to(boost::asio::buffer(buffer, size), m_multicast_endpoint, 0, e);
+			s.socket->send_to(boost::asio::buffer(buffer, std::size_t(size)), m_multicast_endpoint, 0, e);
```
arvidn: if you're doing this, is it possible to omit ``boost::asio::buffer`` and just use ``{}``?

aldenml: it's not possible, because `send_to` templates the first argument (`ConstBufferSequence`) and it requires to explicit the type.

aldenml:
btw, if you are considering merging this and my other PR (https://github.com/arvidn/libtorrent/pull/1514), I'm pretty sure that together will cause master to fail due to warnings. If you merge one, I need to rebase the other.

include/libtorrent/bencode.hpp
```diff
@@ -90,7 +90,7 @@ namespace libtorrent
 			= typename std::enable_if<std::is_integral<In>::value>::type>
 		int write_integer(OutIt& out, In data)
 		{
-			entry::integer_type const val(data);
+			entry::integer_type const val = entry::integer_type(data);
```
arvidn: sorry, did you explain why not type ``entry::integer_type const val(data);`` instead?

arvidn:
your other patch undoes at least some of those ``operator[std::size_t(...)]`` casts, right?

arvidn:
lgtm, just that one comment

include/libtorrent/bencode.hpp
aldenml: (I did not) I needed to restore this because when I compiled in my warning free branch, this direct construction didn't count like an explicit cast, only doing this the warning goes away.

aldenml:
yes, both patch can't be merge at the same time, once you merge one, I will rebase the other

include/libtorrent/bencode.hpp
arvidn: weird. that's not typical. it should count as a ``static_cast<>``.. anyway..

Issue #1510, opened 2017-01-09
'module' object has no attribute 'session_settings'


NetSharkDon:
Hi arvidn 
I use libtorrent python binding ,version '1.2.0.0'
when i  run  python client.py
ERROR
“AttributeError: 'module' object has no attribute 'session_settings'”
HELP ME PLEASE! thanks

Falcosc:
I could confirm that this is currenlty broken.

You could try test.py to check that your compilation does work.

NetSharkDon:
Thank You Very Much
libtorrent 1.1.1 can work
hope for 1.2

demonskull:
This problem is still alive with the Deluge torrent daemon.

arvidn:
session_settings were deprecated a long time ago in favour of settings_pack. I seem to recall it’s still there though, as long as you build with deprecated functions enabled (which is the default)

arvidn:
Do you have any more details on version of libtorrent and how it was built?

marcan:
I'm getting this on upgrade to libtorrent-rasterbar-1.2.0. Deluge fails due to the missing session_settings. It's definitely not there:

```python
Python 2.7.15 (default, Jan 15 2019, 23:07:09) 
[GCC 8.2.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import libtorrent
>>> libtorrent.session_settings
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'session_settings'
```

On another machine still on 1.1.11:

```python
Python 2.7.15 (default, Jan  2 2019, 21:38:32) 
[GCC 8.2.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import libtorrent
>>> libtorrent.session_settings
<class 'libtorrent.session_settings'>
```

marcan:
The build *does* enable deprecated functions:

```
Compiler and linker flags:
  CPPFlags:             -I/usr/include 
  CFlags:               -lpthread -O2 -pipe -march=haswell -fvisibility=hidden
  CXXFlags:             -O2 -pipe -march=haswell -std=c++11 -ftemplate-depth=512 -Wno-format-zero-length -fvisibility=hidden -fvisibility-inlines-hidden
  LDFlags:              -L/usr/lib64 -Wl,-O1 -Wl,--as-needed -fvisibility=hidden -fvisibility-inlines-hidden
[...]

PR #1512, opened 2017-01-09
Fix python client

Falcosc:
for https://github.com/arvidn/libtorrent/issues/1510


bindings/python/client.py
```diff
@@ -110,11 +110,12 @@ def print_peer_info(console, peers):
         else: out += 'r'
         out += ' '
 
-        if p.downloading_piece_index >= 0:
-            assert(p.downloading_progress <= p.downloading_total)
-            out += progress_bar(float(p.downloading_progress) / p.downloading_total, 15)
-        else:
-            out += progress_bar(0, 15)
+        # TODO TypeError: No Python class registered for C++ class libtorrent::aux::strong_typedef<int, libtorrent::aux::piece_index_tag, void>
```
arvidn: you can fix this by adding the converter in ``src/converters.cpp`` under bindings/python.

arvidn: there seems to be one. I'm not sure why it's not working

Falcosc:
I would create a testmethod which calls client.py with a torrent to check if the main functionality of the example is outdated or not.

bindings/python/client.py
```diff
@@ -306,13 +303,7 @@ def main():
...
         write_line(console, 76 * '-' + '\n')
 
-        while 1:
-            a = ses.pop_alert()
-            if not a: break
-            alerts.append(a)
-
-        if len(alerts) > 8:
-            del alerts[:len(alerts) - 8]
+        alerts = ses.pop_alerts()
```
arvidn: this is a great fix btw. I'm surprised I've left that old code around for so long. I don't even get why it's deleting alerts beyond index 8.

bindings/python/client.py
```diff
@@ -192,14 +193,10 @@ def main():
     if options.max_download_rate <= 0:
         options.max_download_rate = -1
 
-    settings = lt.session_settings()
-    settings.user_agent = 'python_client/' + lt.version
-
-    ses = lt.session()
+    ses = lt.session({'user_agent':  'python_client/' + lt.version})
```
arvidn: +1

bindings/python/client.py
arvidn: ah, I think I know what the problem is. the ``downloading_piece_index`` member need to be bound as a property, to be able to be convertible with custom converters.

bindings/python/client.py
Falcosc: Now we have tests for all examples, make_torrent has an binding issue too

arvidn:
Here's an example of this fix. the regular ``.def_readonly`` or even ``.add_property`` won't invoke converters. it has to be ``.add_property()`` with a ``make_getter()`` function to properly use converters:

https://github.com/arvidn/libtorrent/commit/a7dcdf30ef97ed27360dd5d08ae4be439a83e5bc

arvidn:
here's the structure. any field that's a non-trivial type need to use ``.add_property()`` and ``make_getter()``, as far as I understand:
https://github.com/arvidn/libtorrent/blob/master/bindings/python/src/peer_info.cpp#L52

Falcosc:
@arvidn I can't fix create torrent because I dont unterstand the binding with storage create torrent and the functions addFiles and addFile. But looks like there was a refactoring to remove the size requriement at the interface.

bindings/python/src/peer_info.cpp
```diff
@@ -82,7 +83,7 @@ void bind_peer_info()
         .def_readonly("download_queue_length", &peer_info::download_queue_length)
         .def_readonly("upload_queue_length", &peer_info::upload_queue_length)
         .def_readonly("failcount", &peer_info::failcount)
-        .def_readonly("downloading_piece_index", &peer_info::downloading_piece_index)
+        .add_property("downloading_piece_index", make_getter(&peer_info::downloading_piece_index, by_value()))
```
arvidn: would you mind giving the same treatments to other fields that is a piece-index or file-index too?

bindings/python/test.py
```diff
@@ -266,6 +268,34 @@ def test_apply_settings(self):
 		self.assertEqual(s.get_settings()['num_want'], 66)
 		self.assertEqual(s.get_settings()['user_agent'], 'test123')
 
+class test_example_client(unittest.TestCase):
+
+	def test_execute_client(self):
```
arvidn: does this actually work? it just runs the program for a few seconds and make sure it shuts down cleanly?

bindings/python/test.py
```diff
@@ -266,6 +268,34 @@ def test_apply_settings(self):
...
+		with open(os.devnull, "w") as DEVNULL:
+			process = subprocess.Popen([sys.executable,"client.py","url_seed_multi.torrent"], stdout=DEVNULL, stderr=subprocess.STDOUT)
+		# python2 has no Popen.wait() timeout
+		time.sleep(10)
+		returncode = process.poll()
+		if returncode == None:
+			process.kill()
```
arvidn: in this case, you skip the assert. shouldn't this fail the test?

bindings/python/test.py
```diff
@@ -266,6 +268,34 @@ def test_apply_settings(self):
...
+			self.assertEqual(returncode, 0)
+
+	def test_execute_simple_client(self):
+		with open(os.devnull, "w") as DEVNULL:
+			process = subprocess.Popen([sys.executable,"simple_client.py","url_seed_multi.torrent"], stdout=DEVNULL, stderr=subprocess.STDOUT)
+		# python2 has no Popen.wait() timeout
+		time.sleep(10)
+		returncode = process.poll()
+		if returncode == None:
+			process.kill()
```
arvidn: same thing here.. presumably a stall should be considered an error

bindings/python/test.py
```diff
@@ -266,6 +268,34 @@ def test_apply_settings(self):
...
 		self.assertEqual(s.get_settings()['user_agent'], 'test123')
 
+class test_example_client(unittest.TestCase):
+
+	def test_execute_client(self):
+		with open(os.devnull, "w") as DEVNULL:
+			process = subprocess.Popen([sys.executable,"client.py","url_seed_multi.torrent"], stdout=DEVNULL, stderr=subprocess.STDOUT)
+		# python2 has no Popen.wait() timeout
+		time.sleep(10)
+		returncode = process.poll()
```
arvidn: don't you need to send a 'q' to the process to make it shut down?

bindings/python/test.py
```diff
@@ -266,6 +268,34 @@ def test_apply_settings(self):
...
+			process.kill()
+		else:
+			self.assertEqual(returncode, 0)
+
+	def test_execute_simple_client(self):
+		with open(os.devnull, "w") as DEVNULL:
+			process = subprocess.Popen([sys.executable,"simple_client.py","url_seed_multi.torrent"], stdout=DEVNULL, stderr=subprocess.STDOUT)
+		# python2 has no Popen.wait() timeout
+		time.sleep(10)
+		returncode = process.poll()
```
arvidn: I'm pretty confident you could call: ``process.wait()`` here. in the case of ``client.py`` though, you still want to wait some time before sending a `q`. But I just realized you're running ``simple_client.py``, not ``client.py``. It would probably make sense to run client.py too

bindings/python/src/peer_info.cpp
Falcosc: I did not try to understand which was causing the issue.
I will add
https://github.com/Falcosc/libtorrent/blob/bd72ea64c508e2406d5e6ed1abe2f29b76f1341b/bindings/python/src/peer_info.cpp#L87
https://github.com/arvidn/libtorrent/blob/4f89b397b1d53c9043f28b04757950399111400b/bindings/python/src/torrent_handle.cpp#L476
https://github.com/arvidn/libtorrent/blob/8800cae5854cd6993664128570ea8a8f1a9efb44/bindings/python/src/torrent_info.cpp#L221

And all piece and block index in the alert binding as well?
https://github.com/arvidn/libtorrent/blob/13f9c110789df67c10f223d02eb9ed8633f6b7c7/bindings/python/src/alert.cpp

bindings/python/test.py
Falcosc: this does just check for interface issues, python will crash if a binding failed.

bindings/python/test.py
Falcosc: I just want to know that it is still running after 10s to check that there were no bad interface calls. I guess 5s would be enoght

bindings/python/test.py
Falcosc: both clients don't stop, that means no returncode is the expected result.

bindings/python/test.py
Falcosc: How do I check for stall? The client does not have an end it runs until somebody close it.

bindings/python/test.py
Falcosc: I did use the same test for both clients. And both don't close without user action which means I can't use wait.
client.py has it's own test.

Falcosc:
@arvidn do you know why we got sometimes a returncode -6?

And sometimes the libs are not there, I guess I have to care about LD_LIBRARY_PATH as well, maybe I copy it or a export LD_LIBRARY_PATH is visible in subprocess.

arvidn:
keep in mind that there's limited bandwidth on appveyor and travis.

Falcosc:
which implications should I keep in mind?
I try to find out if there are shared libs because sometimes  lt.version disappers and I try to find out when python subprocesses get aborted.

Do you have any idea for both of them?
I could ignore -6 but It's very difficould to ignore interface failures.
Any suggestions?

Falcosc:
@arvidn where did you get this from? https://github.com/arvidn/libtorrent/blob/master/bindings/python/client.py#L60

There is no console on windows, even not on latest python. I would replace it with windows logic with stdout calls and cls calls if you don't remember where console is from.

arvidn:
oh.. Console is a third party library. I guess that might make it difficult to run on appveyor. if you think it's not too much trouble to remove that dependency, please feel free (but that should probably be a separate patch)

arvidn:
regarding travis/appveyor. each time you push to a branch that's a push request, travis will add a new job, and tie up the 4 or so machines for quite a while. on appveyor, there are rolling builds, so a single PR has less impact there.

Falcosc:
I did still not figured out why I'm loosing modules. Do the travis builds share something? I did guess not because they use all the same directory and they don't see the files of the other process (always need a full git checkout) which tells me that they run all on different virtual boxes.
The normal test does work fine because they are executed just after the compilation. But the subprocesses are load there models some time after compilation.

arvidn:
what do you mean by "loosing modules"?

Falcosc:
I did guess we load some outdated ones, but I found out that were deprecated functions.

Falcosc:
Now I did the cleanup, if everthing does work and we have only the console module error and make torrent error left, then I'm ready.

Falcosc:
Now it's done.

I figured out that travis python does not print error messages on lib binding failures. Instead it crashes with returncode -6
Is it ok to skip -6 crashes because they are useless without error message and just use the output of windows or darwin configuration?
I am not sure but I guess I did seen travis gcc jobs which print an error, which would be even worse to properly fix a unpredictive result.

arvidn:
I'm not sure I understand why it would be reasonable to "skip" -6 errors. By skip, you mean ignore them, and not treat them as test failures, right? If there is an issue with the bindings, and they fail with -6, why isn't that an error?

Falcosc:
I guess -6 is an acutall error, but many times the error stream got lost or the missing module function did not get logged and it just crashed. Do you want to have the build errors without error message for the example code check?

arvidn:
well.. I don't think two wrongs make a right. If there's something broken with the binding preventing ``client.py`` from running, that seems like it should make the test fail, regardless of whether we can get travis to report details on what failed or not.

Do you understand why it's failing? presumably, at this point, it's not actually a problem with the binding, but some setup issue.

arvidn:
as far as I can tell, the error in ``make_torrent.py`` seems legitimate.

bindings/python/make_torrent.py
```diff
@@ -19,10 +19,13 @@
 
 parent_input = os.path.split(input)[0]
 
+if os.path.isfile(input):
+	# TODO bind the updated create torrent interface
+	fs.add_files(input)
```
arvidn: why did you add this? the loop below adds the files already. If anything, I think the commented out lines above should be removed.

If it's a matter of testing the ``add_files`` function, perhaps that could be done separately in ``test.py``?

also btw, ``add_files()`` is a free function, not a member of ``file_storage``.

bindings/python/test.py
```diff
@@ -7,6 +7,11 @@
 import os
 import shutil
 import binascii
+import subprocess as sub
+import sys
+
+if os.name != 'nt':
+	import pty
```
arvidn: what's this? would you mind adding a comment?

arvidn:
lgtm, apart from ``make_torrent.py``. In order to make this pass for now, perhaps you could just not run the ``client.py`` test on windows. But in the long run it would be nice to drop the Console dependency and be able to run that test on windows too

Falcosc:
> Do you understand why it's failing? presumably, at this point, it's not actually a problem with the binding, but some setup issue.

No, but after some more builds I could confirm, a missing module member error without output on stderr is always returncode -6 on the travis setup. Maybe the stderr pipe did get lost or something similar. I will add a message how to get the error message.

And I skip all failed test with a todo comment.

bindings/python/test.py
Falcosc: You can't use the default terminal for multible parallel client.py executions, we need to create a new pty for them and pty does not work on windows.

bindings/python/test.py
```diff
@@ -266,9 +273,80 @@ def test_apply_settings(self):
...
+			if returncode != 0:
+				print("stdout:\n" + process.stdout.read().decode("utf-8"))
+			self.assertEqual(returncode, 0, "returncode: " + str(returncode) + "\n"
+				+ "stderr: empty\n"
+				+ "some configuration does not output errors like missing module members,"
+				+ "try to call it manually to get the error message\n")
+
+	def test_execute_make_torrent(self):
+		# TODO: fix deprecated calls of make_torrent.py
+		return
```
arvidn: if you rebase, you shouldn't need this anymore

ChangeLog
```diff
@@ -1,3 +1,5 @@
+	* remove deprecated functions from python examples
+	* add piece index and block index python bindings
```
arvidn: I believe there are already changelog entries for deprecating the functions. I don't think updating the python binding needs its own entry. The same thing applies to the piece_index_t type. I believe it's mentioned in a separate entry

bindings/python/test.py
Falcosc: are you sure? the commit from yesterday did contain only the deprecated add_torrent with size option instead of the new one without size
But I will see after I build it with no-deprecated

arvidn: well, I got make_torrent.py to work with deprecated functions disabled.

Falcosc:
fix done

bindings/python/make_torrent.py
```diff
@@ -33,6 +37,7 @@
 		fname = os.path.join(root[len(parent_input)+1:], f)
 		size = os.path.getsize(os.path.join(parent_input, fname))
 		print('%10d kiB  %s' % (size / 1024, fname))
+		# TODO: is deprecated, bind the updated create torrent interface
```
arvidn: are you sure this comment is correct? I was convinced that I fixed all deprecated calls in ``make_torrent.py``

bindings/python/make_torrent.py
```diff
@@ -19,10 +19,14 @@
 
 parent_input = os.path.split(input)[0]
 
+# if we have a single file, use it because os.walk does not work on a single files
+if os.path.isfile(input):
+	size = os.path.getsize(input)
+	fs.add_file(input, size)
```
arvidn: I don't understand why this was added, would you mind explaining this?

bindings/python/simple_client.py
```diff
@@ -8,14 +8,14 @@
...
+ses = lt.session({'listen_interfaces':'0.0.0.0:6881'})
 
 info = lt.torrent_info(sys.argv[1])
 h = ses.add_torrent({'ti': info, 'save_path': '.'})
-print('starting', h.name())
+s = h.status()
+print('starting', s.name)
 
-while (not h.is_seed()):
+while (not s.is_seeding):
```
arvidn: +1

bindings/python/test.py
```diff
@@ -276,9 +283,78 @@ def test_apply_settings(self):
...
+		# in case of error return: output stdout if nothing was on stderr
+		if returncode != 0:
+			print("stdout:\n" + process.stdout.read().decode("utf-8"))
+		self.assertEqual(returncode, 0, "returncode: " + str(returncode) + "\n"
+			+ "stderr: empty\n"
+			+ "some configuration does not output errors like missing module members,"
+			+ "try to call it manually to get the error message\n")
 
 if __name__ == '__main__':
-	print(lt.__version__)
```
arvidn: why did you remove this? It's meant to verify the version is exported correctly

arvidn:
looks good. just some small questions

arvidn:
I just tested make_torrent.py again on mac, with deprecated functions disabled, and it works for me.

bindings/python/test.py
Falcosc: just one failed commit of 26 ;)

bindings/python/make_torrent.py
Falcosc: os.walk does not work on a single files at least not at python 2.7

bindings/python/make_torrent.py
Falcosc: Not sure why I did miss that, I hope there is not a full commit missing.

bindings/python/make_torrent.py
arvidn: ah, I see. this was just a bug in make_torrent.py

Falcosc:
The merge did already cover all review finidngs.
I will call it a success because it took less then 30 commits ;)

arvidn:
this patch is in a good state now, right?

Falcosc:
Yes it is.

PR #1514, opened 2017-01-10
added custom resize methods to aux::vector

aldenml:
The context of this change is:
1 - safety
2 - help with the upcoming changes to remove more warnings

Now, do you think is it better to receive `IndexType` instead of `underlying_index` in `resize` methods?

### DIFF
--- a/include/libtorrent/aux_/vector.hpp
+++ b/include/libtorrent/aux_/vector.hpp
@@ -86,6 +86,18 @@ namespace libtorrent { namespace aux {
 			TORRENT_ASSERT(s >= 0);
 			this->base::resize(std::size_t(s), v);
 		}
+
+		void resize(std::size_t s)
+		{
+			TORRENT_ASSERT(s <= std::size_t(std::numeric_limits<underlying_index>::max()));
+			this->base::resize(s);
+		}
+
+		void resize(std::size_t s, T const& v)
+		{
+			TORRENT_ASSERT(s <= std::size_t(std::numeric_limits<underlying_index>::max()));
+			this->base::resize(s, v);
+		}
 	};
 
 	template <typename Iter>
--- a/include/libtorrent/aux_/vector.hpp
+++ b/include/libtorrent/aux_/vector.hpp
@@ -34,6 +34,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #define TORRENT_VECTOR_HPP
 
 #include <vector>
+#include <type_traits>
 
 #include "libtorrent/units.hpp"
 #include "libtorrent/assert.hpp"
@@ -75,12 +76,16 @@ namespace libtorrent { namespace aux {
 			return IndexType(static_cast<underlying_index>(this->size()));
 		}
 
+		template <typename U = underlying_index, typename Cond
+			= typename std::enable_if<std::is_signed<U>::value>::type>
 		void resize(underlying_index s)
 		{
 			TORRENT_ASSERT(s >= 0);
 			this->base::resize(std::size_t(s));
 		}
 
+		template <typename U = underlying_index, typename Cond
+			= typename std::enable_if<std::is_signed<U>::value>::type>
 		void resize(underlying_index s, T const& v)
 		{
 			TORRENT_ASSERT(s >= 0);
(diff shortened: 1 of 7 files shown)


arvidn:
I'm not sure. underlying type seems more reasonable since it isn't referring to an element. will you keep the existing ``resize()`` too? in case underlying type is unsigned or something like that

aldenml:
hmm...not sure about your `resize()` question. This `resize()` captures whatever the `underlying_index` is, either from `aux::strong_typedef` or plain integral type.

include/libtorrent/aux_/vector.hpp
```diff
@@ -70,7 +70,22 @@ namespace libtorrent { namespace aux {
...
+			return IndexType(static_cast<underlying_index>(this->size()));
+		}
+
+		void resize(underlying_index s)
+		{
+			TORRENT_ASSERT(s >= 0);
+			this->base::resize(std::size_t(s));
+		}
+
+		void resize(underlying_index s, T const& v)
```
arvidn: hm.. I wonder if you want to do: ``T&& v`` here, and then pass it on with: ``std::forward<T>(v)``. I'm not sure, it might be worth considering

src/piece_picker.cpp
```diff
@@ -3076,7 +3076,7 @@ namespace libtorrent
 		TORRENT_ASSERT(m_seeds >= 0);
 		INVARIANT_CHECK;
 
-		avail.resize(m_piece_map.size());
```
arvidn: this is what I meant by preserving the original ``resize()`` function. I think this should work without a cast

include/libtorrent/aux_/vector.hpp
```diff
@@ -70,7 +70,22 @@ namespace libtorrent { namespace aux {
 		}
 
 		IndexType end_index() const
-		{ return IndexType(static_cast<underlying_index>(this->size())); }
+		{
+			TORRENT_ASSERT(this->size() <= std::size_t(std::numeric_limits<underlying_index>::max()));
+			return IndexType(static_cast<underlying_index>(this->size()));
+		}
+
```
arvidn: if you add a line: ``using base::resize;`` here, you'll pull in the overloads from the base class too

aldenml: ok, I will do that

include/libtorrent/aux_/vector.hpp
aldenml: I'm not sure either, I want to stay as close as possible to the original `base`

src/piece_picker.cpp
aldenml: ok

aldenml:
I think it's better if you merge my other PR first, then I will rebase this one and make the changes

arvidn:
done

include/libtorrent/aux_/vector.hpp
aldenml: I think I will rewrite the `resize()`, I want to put some asserts

aldenml:
do you think the CI's failure has anything to do with the change? I notice that at some hours of the day, the failures are more frequent.

arvidn:
I doubt it. I've noticed this test in particular being flaky

include/libtorrent/aux_/vector.hpp
```diff
@@ -70,7 +71,38 @@ namespace libtorrent { namespace aux {
...
+			= typename std::enable_if<std::is_signed<U>::value>::type>
+		void resize(underlying_index s, T const& v)
+		{
+			TORRENT_ASSERT(s >= 0);
+			this->base::resize(std::size_t(s), v);
+		}
+
+		void resize(std::size_t s)
+		{
+			TORRENT_ASSERT(s <= std::size_t(std::numeric_limits<underlying_index>::max()));
```
arvidn: since this is in a header, I would expect you would have to do the trick to avoid macro expansion for max. i.e. ``(std::numeric_limits<underlying_index>::max)()``.

include/libtorrent/aux_/vector.hpp
```diff
@@ -70,7 +71,38 @@ namespace libtorrent { namespace aux {
...
+			this->base::resize(s);
+		}
+
+		void resize(std::size_t s, T const& v)
+		{
+			TORRENT_ASSERT(s <= std::size_t(std::numeric_limits<underlying_index>::max()));
```
arvidn: same thing here

aldenml:
if the CI pass, do you mind if I add the `(..)` in my next PR?

arvidn:
sure, that's fine

PR #1518, opened 2017-01-10
fix of internal session_impl logic regarding TOS

aldenml:


arvidn:
from address sanitizer, it looks like my patch is the culprit. could you try backing that part out?

aldenml:
ok, doing it, but isn't that the fix?

arvidn:
this change looks reasonable. I don't think this applies to RC_1_1 though, does it? my recollection is that RC_1_1 doesn't have a UDP socket per listen_socket structure.

Issue #1520, opened 2017-01-10
Add AM_MAINTAINER_MODE([enable]) to configure.ac


cas--:
The maintainer mode is enabled by default without the macro. However debian build system passes  `--disable-maintainer-mode` which cannot be acted upon because the macro is missing resulting in this error:
```
configure: WARNING: unrecognized options: --disable-maintainer-mode
```

Adding the following macro with parameter ensures maintainer mode is still enabled by default but can be disabled if requested.
```
AM_MAINTAINER_MODE([enable]) 
```

References: 
https://autotools.io/automake/maintainer.html
https://blogs.gnome.org/desrt/2011/09/08/am_maintainer_mode-is-not-cool/

cas--:
Could this go in 1.1.2 milestone instead?

arvidn:
sure. I was planning on making one last 1.0.x release though, and it seems like a simple patch

cas--:
Oops not applied to 1.0.x yet

Issue #955, opened 2016-07-27
Configure error: sed: can't read include/libtorrent/version.hpp


cas--:
I have seen this into the configure output over quite a few versions now:

```
checking for iconv declaration... 
         extern size_t iconv (iconv_t cd, char * *inbuf, size_t *inbytesleft, char * *outbuf, size_t *outbytesleft);
sed: can't read include/libtorrent/version.hpp: No such file or directory

Generating Makefiles:
```

I've not checked the code for the reason it is occurring but felt I should at least log it in case there is an easy fix... :)

One small detail I missed is that this is with debian package build so this an example configure line:

```
CPPFLAGS="-D_FORTIFY_SOURCE=2 -I/usr/include/python3.2dmu -I/usr/include/python3.2dmu" \
    PYTHON_LDFLAGS="-lpthread -ldl -lutil -lm -lpython3.2dmu" \
    dh_auto_configure --builddirectory build-py3.2-dbg -- \
        PYTHON=/usr/bin/python3.2-dbg PYTHON_INSTALL_PARAMS="--root=/home/ubuntu/Projects/lt-PPA-cas/lt_1.1git/libtorrent-RC_1_1/debian/tmp --install-layout=deb" \
        --enable-python-binding --with-boost-python=boost_python-py32 \
        --with-libiconv
```

Perhaps the SVN version test should be removed?

Falcosc:
What did you take as base of your compilation? Because I can't reproduce it on 1.1, maybe it is already fixed?

On master there is a GIT revision check and on RC_1_1 an svn revision check. If the revision check does not match your checkout client, it will just return an empty string but no errors because include/libtorrent/version.hpp is always there.

Falcosc:
@cas-- did you still able to reproduce the issue? Because I would like to have the possibility to test a change somewere where the error appears.

cas--:
Every debian pkg build has this error:

https://launchpadlibrarian.net/301686651/buildlog_ubuntu-zesty-amd64.libtorrent-rasterbar_1.1.2dev8436+5d1fb97+201701081817~ubuntu17.04.1_BUILDING.txt.gz

Falcosc:
Thanks

Falcosc:
@cas-- where did you get your sources from? Because you did not build from version controll which means version have to be taken from the version file but your sources doesn't include them. I did check the gitbranches and the released sources but all have the missing file.

Could you check why your source archive is incomplete?
```
Local sources
-------------

[...]

PR #1521, opened 2017-01-10
[#1520] Add AM_MAINTAINER_MODE macro to configure.ac

cas--:
* '--disable-maintainer-mode' option is now available for configure.
 * This fixes the warning 'unrecognized options: --disable-maintainer-mode' from debian pkg builds.
 * Setting the macro to 'enable' as default is equivalent to no macro call so no change in default functionality.

Could this also be applied to RC_1_1, thanks!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -79,7 +79,7 @@ AC_CANONICAL_TARGET
 AS_ECHO
 AS_ECHO "Initializing Automake:"
 AM_INIT_AUTOMAKE([1.11 foreign])
-
+AM_MAINTAINER_MODE([enable])
 
 AS_ECHO
 AS_ECHO "Initializing Libtool:"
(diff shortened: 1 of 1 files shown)


Falcosc:
@cas-- please commit this to the oldest branch then arvid could merge it during the normal tasks.
If you start with master it will be extra work

And if you have time you could remove the svn check of https://github.com/arvidn/libtorrent/issues/955 because I have already a full shedule.

Thanks

cas--:
Not sure I follow what you mean?

Merging with master and then backporting is the sane workflow.

Falcosc:
@arvidn is it the same for you? I didn't have merged anything between libtorrent branches

arvidn:
what about distros other than debian. will this have any effect when not building with ``--disable-maintainer-mode``?

Falcosc:
According to:

> Setting the macro to 'enable' as default is equivalent to no macro call so no change in default functionality.

there should be no effects for other distros

PR #1524, opened 2017-01-11
fix file rename issue with name prefix matching torrent name

arvidn:


sledgehammer999:
Fix confirmed.

arvidn:
the original fix broke some other test. updated to be a proper fix. the new unit test still passes.

PR #1526, opened 2017-01-11
fix bitfield m_buf null pointer

Falcosc:
I didn't apply it to 1.0 or 1.1 because on the change on master is larger then on the old branches.
Do you want to have pullrequests for them as well or would you merge and test it?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -49,7 +49,7 @@
 	* resume data no longer has timestamps of files
 	* require C++11 to build libtorrent
 
-	* fix bitfield m_buf null pointer
+	* fix crash caused by empty bitfield 
 	* fix division by zero when setting tick_interval > 1000
 	* fix move_storage() to its own directory (would delete the files)
 	* fix socks5 support for UDP
--- a/ChangeLog
+++ b/ChangeLog
@@ -49,6 +49,7 @@
 	* resume data no longer has timestamps of files
 	* require C++11 to build libtorrent
 
+	* fix bitfield m_buf null pointer
 	* fix division by zero when setting tick_interval > 1000
 	* fix move_storage() to its own directory (would delete the files)
 	* fix socks5 support for UDP
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* fix crash caused by empty bitfield 
 	* removed disk-access-log build configuration
 	* removed mmap_cache feature
 	* strengthened type safety in handling of piece and file indices
@@ -49,7 +50,6 @@
 	* resume data no longer has timestamps of files
 	* require C++11 to build libtorrent
 
-	* fix crash caused by empty bitfield 
 	* fix division by zero when setting tick_interval > 1000
 	* fix move_storage() to its own directory (would delete the files)
 	* fix socks5 support for UDP
--- a/src/bitfield.cpp
+++ b/src/bitfield.cpp
@@ -41,6 +41,8 @@ namespace libtorrent
 {
 	bool bitfield::all_set() const
 	{
+		if(m_buf == nullptr) return false;
+
 		int const words = size() / 32;
 		for (int i = 1; i < words + 1; ++i)
 		{
--- a/src/bitfield.cpp
+++ b/src/bitfield.cpp
@@ -41,7 +41,7 @@ namespace libtorrent
 {
 	bool bitfield::all_set() const
 	{
-		if(m_buf == nullptr) return false;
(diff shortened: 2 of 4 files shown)


test/test_bitfield.cpp
```diff
@@ -355,3 +355,56 @@ TORRENT_TEST(find_last_clear_misc)
...
+
+	test1.clear();
+	TEST_EQUAL(test1.size(), 0);
+
+	test1.set_all();
+	TEST_EQUAL(test1.size(), 0);
+
+	// don't test the following methods because they are optimized and would be
+	// only called after initialization to skip the null check
+	// get_bit, clear_bit, set_bit
```
arvidn: more specifically, they aren't defined for empty sets. just like you can't use ``operator[]`` on an empty vector

test/test_bitfield.cpp
```diff
@@ -355,3 +355,56 @@ TORRENT_TEST(find_last_clear_misc)
...
+}
+
+TORRENT_TEST(not_initialized_assign)
+{
+	// in some cases for example peer disconnect before receiving bitfield
+	// we try to check a not initialized bitfield
```
arvidn: I don't think these comments are great. they're prone to become outdated. instead, just describe at a high level what's being tested

src/bitfield.cpp
```diff
@@ -41,6 +41,8 @@ namespace libtorrent
 {
 	bool bitfield::all_set() const
 	{
+		if(m_buf == nullptr) return false;
```
arvidn: would it make more sense to say: ``if (size() == 0) return false;`` ?
it feels like that would be more intuitive and direct

ChangeLog
```diff
@@ -49,6 +49,7 @@
 	* resume data no longer has timestamps of files
 	* require C++11 to build libtorrent
 
+	* fix bitfield m_buf null pointer
```
arvidn: m_buf null pointer is an implementation detail. say "empty bitfield" instead

Col-blimp:
Implemented the changes, now `std::cout` looks like so:

```
url_seed_alert (error) libtorrent:24
url_seed_alert (error) libtorrent:24
url_seed_alert (error) system:11001
url_seed_alert (error) system:11001
url_seed_alert (error) system:11002
url_seed_alert (error) system:11001
url_seed_alert (error) system:11001
url_seed_alert (error) system:11001
url_seed_alert (error) system:11002
url_seed_alert (error) system:11001
url_seed_alert (error) system:11001
url_seed_alert (error) system:11001
u
```
then crashes before finishing writing out the message from `url_seed_alert` although with the fixes above no asserts.log is generated.

Not doing anything in particular with `url_seed_alert`:

```
void do_url_error_alert(const libtorrent::url_seed_alert *u , const libtorrent::alert *a) {
	if ( u->error ) {
		std::cout << "url_seed_alert (error) " << u->error << "\n";
	}
}
```

except only writing out errors as I was not interested in status messages.

This is a bit hit and miss as you can download the torrent fine several times then you get the unstable behaviour.

Col-blimp:
With the latest changes asserts=production build still crashes with no explanation.

Was looking to get more info with the variant=debug build however that build runs fine and it appears to run fine as it strips out all the web seeds from that torrent so gets no errors.

Is it the case that none of these web seeds are compatible and should all be removed on variant=release build?

arvidn:
you can build with ``release`` and ``debug-symbols=on``. That will give you more useful information in the debugger. You may even want to try ``optimization=off``, if that still exhibits the problem you'll have even more legible crash dumps

arvidn:
this patch lgtm

Col-blimp:
My issues are resolved (threading issues) and this patch does the job, thanks.

ChangeLog
```diff
@@ -49,6 +49,7 @@
 	* resume data no longer has timestamps of files
 	* require C++11 to build libtorrent
 
+	* fix crash caused by empty bitfield
```
arvidn: I didn't notice this until just now. this is in the wrong place. you're in master, this should be at the top of the fil

Falcosc: Wait, then I didn't understand the changelog. I guess changes on top fixes on bottom.
could you discribe what is on top and what on bottom?

Falcosc: I guess I got it, you use the bottom list for RC_1_1 merges

arvidn: yeah, the blank is a placeholder for the release marker from the RC_1_1 branch

PR #1529, opened 2017-01-12
enforce asserts on invariant_check

ajax16384:
invariant_check needs asserts to be enabled

### DIFF
--- a/Jamfile
+++ b/Jamfile
@@ -111,7 +111,6 @@ rule linking ( properties * )
 	if <target-os>windows in $(properties)
 		&& ( <asserts>on in $(properties)
 			|| <asserts>production in $(properties)
-			|| ! <invariant-checks>off in $(properties)
 			|| <pool-allocators>debug in $(properties)
 			|| <allocator>debug in $(properties)
 			|| <asio-debugging>on in $(properties) )
@@ -172,7 +171,6 @@ rule linking ( properties * )
 		&& <target-os>linux in $(properties)
 		&& ( <asserts>on in $(properties)
 			|| <asserts>production in $(properties)
-			|| ! <invariant-checks>off in $(properties)
 			|| <pool-allocators>debug in $(properties)
 			|| <allocator>debug in $(properties)
 			|| <asio-debugging>on in $(properties) )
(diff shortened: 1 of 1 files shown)


ajax16384:
it seems "<asserts>on" should be included only if "<asserts>off"

arvidn:
you have to update the travis and appveyor scripts to not attempt to build an inconsistent configuration too

arvidn:
I don't think this is the right path to go down. it won't scale. I think it's better to have the boost-build  features correspond 1:1 with the build features in libtorrent.

The error you got after your first patch essentially said: asserts=off and invariant-checks=on are incompatible. Which in some sense is reasonable. That's essentially what your original point was, right?

ajax16384:
I meant " asserts=off and invariant-checks=on " is illegal combination, so i want to add jamfile rule which looks like this pseduo code:
````c++
if ((invariant-checks != off) && (asserts == off)) { asserts=on }
````

if you think that developer should not specify " asserts=off and invariant-checks=on " at all, then simply don't bother and drop this pull request.

arvidn:
that logic is not possible. if the property asserts=on is present, you can't remove it or modify it (at least not as far as I know) and also, it would be a questionable thing to do, since properties are associated with build targets and are propagated up and down the dependency tree. if you could (easily) make arbitrary modifications to it, you may end up building objects with different configurations. For instance, a client that links against libtorrent may build with asserts=off invariant-checks=on, but then once libtorrent builds asserts are enabled and libtorrent is built with a different configuration

arvidn:
I think there's still value in explicitly forbidding such combinations though, to make b2 print a helpful error message, rather than having surprising runtime behavior

arvidn:
if you make invariant-checks=on also add the asserts=on property, it does effectively make the invalid combinations an error. the build will have asserts=on *and* asserts=off, which will fail (iiuc)

arvidn:
lgtm

aldenml:
I was under the impression that an `ECHO` with a `result += <build>no ;` was preferred

arvidn:
that's a good point. I think you're right

ajax16384:
should i made another pull request?

arvidn:
if you don't mind, please do

Issue #1528, opened 2017-01-11
Crash on torrent load


ngosang:
qBittorrent version: v3.3.5
Libtorrent version: 1.0.9.0
Qt version: 5.6.1
Boost version: 1.61.0
OS version: Windows 10 10.0.10586 x86_64

Could you test the torrent file provided? https://github.com/qbittorrent/qBittorrent/issues/5480#issuecomment-230411200
Thank you

ngosang:
Reported in https://github.com/qbittorrent/qBittorrent/issues/5480

sledgehammer999:
It crashes with latest RC_1_0.

Relevant backtrace:

```
#  8 qbittorrent.exe      0x000000013fbd1710 memcpy()[ f:\dd\vctools\crt\vcruntime\src\string\amd64\memcpy.asm : 138 ]
#  9 qbittorrent.exe      0x000000013f313dac libtorrent::torrent_info::hash_for_piece(index)[ g:\qbittorrent\libtorrent\include\libtorrent\torrent_info.hpp : 561 ]
# 10 qbittorrent.exe      0x000000013f32764a libtorrent::create_torrent::create_torrent(ti)[ g:\qbittorrent\libtorrent\src\create_torrent.cpp : 360 ]
# 11 qbittorrent.exe      0x000000013f1d67f1 BitTorrent::TorrentHandle::saveTorrentFile(path)[ base\bittorrent\torrenthandle.cpp : 1315 ]
# 12 qbittorrent.exe      0x000000013f1bf622 BitTorrent::Session::createTorrentHandle(nativeHandle)[ base\bittorrent\session.cpp : 3295 ]
```
@ngosang I think that's why I didn't report. The backtrace needs a little bit investigation on our part. But now I see it still a libtorrent problem.

RC_1_1 fails much more gracefully. When adding the torrent we always check if the obtained torrent_info is valid. RC_1_1 tells that it isn't with error: `incorrect number of piece hashes in torrent file`
While RC_1_0 tells that this is a valid torrent_info

Falcosc:
It does not crash in our test client. 
And head of RC_1_0 test client https://github.com/arvidn/libtorrent/blob/RC_1_0/examples/client_test.cpp does return incorrect number of piece hashes in torrent file as well

Falcosc:
I did manually check the
info.length = 2196780193
that are 524 pieces because of info.piece length 4194304
[...]

PR #1531, opened 2017-01-12
test invalid torrent info

Falcosc:
https://github.com/arvidn/libtorrent/issues/1528

### DIFF
--- a/test/test_create_torrent.cpp
+++ b/test/test_create_torrent.cpp
@@ -41,6 +41,8 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace lt = libtorrent;
 
+int test_invalid_piece_length();
+
 // make sure creating a torrent from an existing handle preserves the
 // info-dictionary verbatim, so as to not alter the info-hash
 int test_main()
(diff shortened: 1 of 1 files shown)


arvidn:
RC_1_0 and RC_1_1 can be built without exceptions, in which case errors are reported by ``error_code`` output parameters

arvidn:
please add the actual fix too, to the loading of the torrent in ``torrent_info``.

also, would you mind taking a look at why create_torrent() crashes? it shouldn't do that, even with a bad torrent as input.

Falcosc:
I didn't find the broken code, because I can't create a broken torrent info.
Do you have an example for the exception less handling? I don't know where to get the error_code without exceptions after I did a wrong aktion.

arvidn:
This constructor takes a reference to an ``error_code``. In case it fails, that error code is set to indicate which error occurred:

https://github.com/arvidn/libtorrent/blob/RC_1_0/include/libtorrent/torrent_info.hpp#L351

Falcosc:
No you can't, that is the point of my problem because each **torrent_info::torrent_info** constructor calls parse_torrent_file() https://github.com/arvidn/libtorrent/blob/RC_1_0/src/torrent_info.cpp#L1187
which does all the checks.

You need to clone a broken torrent info, or adjust a already validated torrent info. Or you construct torrent_info with another torrent info.

For that reason I need to know wich constructor was used to crash libtorrent. Maybe one of the constructors are able to skip parse_torrent_file on special conditions.

arvidn:
ok. so what does this patch do?

Falcosc:
just test something which does already work

arvidn:
it's probably better to extend test coverage in RC_1_1 (or no older than RC_1_1 at least)

PR #1537, opened 2017-01-12
fixing sign-conversion warnings, part 3

aldenml:


include/libtorrent/link.hpp
```diff
@@ -48,7 +50,7 @@ namespace libtorrent
 		void clear() { index = -1; }
 
 		template <class T>
-		void unlink(std::vector<T*>& list, int link_index)
+		void unlink(aux::vector<T*, int>& list, int link_index)
```
arvidn: maybe ``int`` should be the default template argument for ``Index``

aldenml: I think it's a good idea, I will do that

arvidn:
lgtm

PR #1540, opened 2017-01-13
added std:: prefix, end of file cleanup, formatting, consts

aldenml:


include/libtorrent/aux_/merkle.hpp
```diff
@@ -45,4 +45,3 @@ namespace libtorrent
 }
 
 #endif
-
```
arvidn: you know, it wasn't that long ago when this was undefined behavior :)

aldenml: true :) (only that I'm removing the extra ones)

Issue #606, opened 2016-04-10
python binding: read_resume_data() bad return type


lawnmowerwcd:
Using the latest code from master. In the python binding `read_resume_data()` has return type `add_torrent_params` but this neither does have a converter to a python type nor is consistent with the `[async_]add_torrent()` methods. It should probably return an appropriate dict that can be passed to those methods.

```
TypeError: No to_python (by-value) converter found for C++ type: libtorrent::add_torrent_params
```

arvidn:
ah, yes, that's a good point

arvidn:
looking into this a bit, I think it would make sense to promote add_torrent_params to a properly exported type. Primarily because it's now passing essential resume-data information via python, and with the current design that would require two conversion steps, which seems unnecessarily error prone.

Does that make sense?

Basically, if some information is inaccessible from python, or poorly bound, at least it will be preserved for the add_torrent() call.

arvidn:
does this solve the problem for you? https://github.com/arvidn/libtorrent/pull/609

lawnmowerwcd:
Thank, that works pretty well. Several attributes of `add_torrent_params` cannot be accessed due to missing converters but the most important attributes seem to work (for me at least).

Falcosc:
Some of them are not allowed to use, others does already work. We take care of parameters which are left.

arvidn:
I believe the problem with those fields is that they're exposed as ``def_readwrite`` rather than ``add_property``. like this: https://github.com/arvidn/libtorrent/blob/master/bindings/python/src/session.cpp#L634

PR #1541, opened 2017-01-13
add check and test for torrent parameter

Falcosc:
https://github.com/arvidn/libtorrent/issues/606

I did try to find the binding misstake, but wasn't sucessfull. At least we have now a test an I did group the errors and put some open questions into it about what should be exposed and what shouldn't.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* python binding add more add_torrent_params fields and an invalid key check
 	* introduce introduce distinct types for peer_class_t, piece_index_t and
 	  file_index_t.
 	* fix crash caused by empty bitfield
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,8 +2,6 @@
 	* introduce introduce distinct types for peer_class_t, piece_index_t and
 	  file_index_t.
 	* fix crash caused by empty bitfield
-	* python binding check for invalid torrent parameter
-	* python binding check add_torrent_params fields
 	* removed disk-access-log build configuration
 	* removed mmap_cache feature
 	* strengthened type safety in handling of piece and file indices
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,8 @@
 	* introduce introduce distinct types for peer_class_t, piece_index_t and
 	  file_index_t.
 	* fix crash caused by empty bitfield
+	* python binding check for invalid torrent parameter
+	* python binding check add_torrent_params fields
 	* removed disk-access-log build configuration
 	* removed mmap_cache feature
 	* strengthened type safety in handling of piece and file indices
--- a/bindings/python/src/converters.cpp
+++ b/bindings/python/src/converters.cpp
@@ -110,19 +110,6 @@ struct tuple_to_pair
 };
 
 template<class T1, class T2>
-struct map_to_dict
-{
-    static PyObject* convert(const std::map<T1, T2>& m)
-    {
-        bp::dict dictionary;
-        for (auto& iter : m) {
-            dictionary[iter.first] = iter.second;
-        }
-        return incref(dictionary.ptr());
-    }
-};
-
-template<class T1, class T2>
 struct dict_to_map
 {
     dict_to_map()
@@ -262,8 +249,6 @@ void bind_converters()
     to_python_converter<std::vector<lt::tcp::endpoint>, vector_to_list<lt::tcp::endpoint>>();
     to_python_converter<std::vector<lt::udp::endpoint>, vector_to_list<lt::udp::endpoint>>();
     to_python_converter<std::vector<std::pair<std::string, int>>, vector_to_list<std::pair<std::string, int>>>();
-
-    to_python_converter<std::map<lt::file_index_t, std::string>, map_to_dict<lt::file_index_t, std::string>>();
 
     to_python_converter<lt::piece_index_t, from_strong_typedef<lt::piece_index_t>>();
     to_python_converter<lt::file_index_t, from_strong_typedef<lt::file_index_t>>();
--- a/bindings/python/src/converters.cpp
+++ b/bindings/python/src/converters.cpp
@@ -252,7 +252,6 @@ void bind_converters()
     to_python_converter<lt::udp::endpoint, endpoint_to_tuple<lt::udp::endpoint>>();
     to_python_converter<lt::address, address_to_tuple>();
     to_python_converter<std::pair<std::string, int>, pair_to_tuple<std::string, int>>();
-    to_python_converter<std::map<lt::file_index_t, std::string>, map_to_dict<lt::file_index_t, std::string>>();
 
     to_python_converter<std::vector<lt::stats_metric>, vector_to_list<lt::stats_metric>>();
     to_python_converter<std::vector<lt::pool_file_status>, vector_to_list<lt::pool_file_status>>();
@@ -263,6 +262,8 @@ void bind_converters()
     to_python_converter<std::vector<lt::tcp::endpoint>, vector_to_list<lt::tcp::endpoint>>();
     to_python_converter<std::vector<lt::udp::endpoint>, vector_to_list<lt::udp::endpoint>>();
     to_python_converter<std::vector<std::pair<std::string, int>>, vector_to_list<std::pair<std::string, int>>>();
+
+    to_python_converter<std::map<lt::file_index_t, std::string>, map_to_dict<lt::file_index_t, std::string>>();
 
     to_python_converter<lt::piece_index_t, from_strong_typedef<lt::piece_index_t>>();
     to_python_converter<lt::file_index_t, from_strong_typedef<lt::file_index_t>>();
--- a/bindings/python/src/converters.cpp
+++ b/bindings/python/src/converters.cpp
@@ -109,6 +109,54 @@ struct tuple_to_pair
     }
 };
 
+template<class T1, class T2>
+struct map_to_dict
+{
+    static PyObject* convert(const std::map<T1, T2>& m)
+    {
+        bp::dict dictionary;
+        for (auto& iter : m) {
+            dictionary[iter.first] = iter.second;
+        }
+        return incref(dictionary.ptr());
+    }
+};
+
+template<class T1, class T2>
+struct dict_to_map
+{
+    dict_to_map()
+    {
+        converter::registry::push_back(
+            &convertible, &construct, type_id<std::map<T1, T2>>()
+        );
+    }
+
+    static void* convertible(PyObject* x)
+    {
+        return PyDict_Check(x) ? x: nullptr;
+    }
+
+    static void construct(PyObject* x, converter::rvalue_from_python_stage1_data* data)
+    {
+        void* storage = ((converter::rvalue_from_python_storage<
+            std::map<T1, T2>>*)data)->storage.bytes;
+
+        dict o(borrowed(x));
+        std::map<T1, T2> m;
+
+        list iterkeys = (list)o.keys();
+        int const len = int(boost::python::len(iterkeys));
+        for (int i = 0; i < len; i++)
+        {
+            object key = iterkeys[i];
+            m[extract<T1>(key)] = extract<T2>(o[key]);
+        }
+        new (storage) std::map<T1, T2>(m);
+        data->convertible = storage;
+    }
+};
+
 template<class T>
 struct vector_to_list
 {
@@ -203,6 +251,8 @@ void bind_converters()
     to_python_converter<lt::tcp::endpoint, endpoint_to_tuple<lt::tcp::endpoint>>();
     to_python_converter<lt::udp::endpoint, endpoint_to_tuple<lt::udp::endpoint>>();
     to_python_converter<lt::address, address_to_tuple>();
+    to_python_converter<std::pair<std::string, int>, pair_to_tuple<std::string, int>>();
+    to_python_converter<std::map<lt::file_index_t, std::string>, map_to_dict<lt::file_index_t, std::string>>();
 
     to_python_converter<std::vector<lt::stats_metric>, vector_to_list<lt::stats_metric>>();
     to_python_converter<std::vector<lt::pool_file_status>, vector_to_list<lt::pool_file_status>>();
@@ -223,6 +273,7 @@ void bind_converters()
     tuple_to_endpoint<lt::tcp::endpoint>();
     tuple_to_endpoint<lt::udp::endpoint>();
     tuple_to_pair<lt::piece_index_t, int>();
+    dict_to_map<lt::file_index_t, std::string>();
     list_to_vector<int>();
     list_to_vector<std::uint8_t>();
     list_to_vector<std::string>();
--- a/bindings/python/src/session.cpp
+++ b/bindings/python/src/session.cpp
@@ -27,12 +27,12 @@
 
 namespace boost
 {
-	// this fixes mysterious link error on msvc
-	libtorrent::alert const volatile*
-	get_pointer(libtorrent::alert const volatile* p)
-	{
-		return p;
-	}
+    // this fixes mysterious link error on msvc
+    libtorrent::alert const volatile*
+    get_pointer(libtorrent::alert const volatile* p)
+    {
+        return p;
+    }
 }
 
 #include "gil.hpp"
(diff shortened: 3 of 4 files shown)


bindings/python/test.py
```diff
@@ -57,6 +57,53 @@ def test_torrent_handle(self):
 		self.h.prioritize_pieces([(0, 1)])
 		self.assertEqual(self.h.piece_priorities(), [1])
 
+	def test_unknown_torrent_parameter(self):
+		self.ses = lt.session({'alert_mask': lt.alert.category_t.all_categories,
+			'enable_dht': False})
+		try:
+			self.h = self.ses.add_torrent({'storage': ''})
```
arvidn: it's not obvious to me why this should be inaccessible. sure, it can't actually be set to a function pointer because of the problems of calling back into python code from libtorrent, but it would be possible for it to be set to some other object to select between the ``default_storage``, ``zero_storage`` and ``discard_storage``. similar to how plugins can be selected from python

bindings/python/test.py
```diff
@@ -57,6 +57,53 @@ def test_torrent_handle(self):
...
+	def test_torrent_parameter(self):
+		self.ses = lt.session({'alert_mask': lt.alert.category_t.all_categories,
+			'enable_dht': False})
+		self.ti = lt.torrent_info('url_seed_multi.torrent');
+		self.h = self.ses.add_torrent({
+			'ti': self.ti,
+			'save_path': os.getcwd(),
+			# TODO: Expecting an object of type list; got an object of type str instead
+			# I don't know what is wrong at the binding
+			#'trackers': ['http://test.com/announce'],
```
arvidn: your comment here suggests that "trackers" is expected to be a list of list of strings. which kind of makes sense. On the C++ side, the ``tracker_tiers`` is used to indicate which "list" each torrent belongs to, and ``trackers`` is flat instead. We should make sure though that ``load_resume_data`` is compatible with this

bindings/python/test.py
```diff
@@ -57,6 +57,53 @@ def test_torrent_handle(self):
...
+		self.assertEqual(self.st.save_path, os.getcwd())
+		self.assertEqual(self.h.trackers(), [])
+		self.assertEqual(self.h.file_priorities(), [4,4])
+		self.assertEqual(self.h.http_seeds(),[])
+		self.assertEqual(self.h.url_seeds(),['http://test.com/file/'])
+		self.assertEqual(self.h.piece_priorities(),[4])
+		self.assertEqual(self.ti.merkle_tree(),[])
+		self.assertEqual(self.st.verified_pieces,[])
+		# TODO: Do we need to expose one of them?
+		# print(self.h.tracker_tiers)
```
arvidn: tracker_tiers doesn't actually exist once the torrent is created. it's just a mechanism to implement a list of lists in a flat list

bindings/python/test.py
```diff
@@ -57,6 +57,53 @@ def test_torrent_handle(self):
...
+		# print(self.st.dht_nodes)
+		# print(self.st.banned_peers)
+		# print(self.st.peers)
+		# print(self.st.unfinished_pieces)
+		# print(self.st.have_pieces)
+		# TODO make_getter did not find a converter
+		# print(self.ti.renamed_file)
```
arvidn: asking the torrent for its torrent file may reflect whether files have been renamed or not

bindings/python/test.py
Falcosc: Make sense, wasn't sure about it.
And do wee need some of the other mentioned fileds?

bindings/python/test.py
Falcosc: That is only one example, all list based members of torrent parameters does have a binding issue which I couldn't solve

bindings/python/test.py
Falcosc: Ok, wasn't sure about it.

bindings/python/src/session.cpp
```diff
@@ -57,61 +57,61 @@ namespace lt = libtorrent;
 namespace
 {
 #ifndef TORRENT_NO_DEPRECATE
-    void listen_on(lt::session& s, int min_, int max_, char const* interface, int flags)
```
arvidn: would you mind reverting the white-space change? It dilutes the actual diff. At some point the spaces in here should be turned into tabs, but that should be a white-space-only patch

ChangeLog
```diff
@@ -1,3 +1,4 @@
+	* python binding check for invalid torrent parameter
```
arvidn: instead of "invalid torrent parameter", don't you mean "add_torrent_params fields"?

bindings/python/src/session.cpp
Falcosc: I did it multible commits, one for spaces and one for changes
here is the change: https://github.com/arvidn/libtorrent/pull/1541/commits/2dc55dffc9736449d1561a2b2c237012d153c2ff

bindings/python/test.py
Falcosc: do you mean I should remove it from binding?

bindings/python/src/session.cpp
arvidn: the whitespace changes should be a separate review/PR. I think it would be best to fix *all* python binding source files to conform to the code conventions, not just session.cpp

bindings/python/src/session.cpp
```diff
@@ -100,174 +100,229 @@ namespace
...
-		}
-
-		for (int i = settings_pack::bool_type_base;
-			i < settings_pack::max_bool_setting_internal; ++i)
-		{
-			ret[name_for_setting(i)] = sett.get_bool(i);
-		}
-		return ret;
-	}
+    void make_settings_pack(lt::settings_pack& p, dict const& sett_dict)
```
arvidn: this whole hunk appears identical (as if it's just white-space changes)

bindings/python/test.py
arvidn: no, I was just commenting on how it could be tested. ``torrent_handle`` doesn't have the same fields as ``add_torrent_params``.

bindings/python/src/session.cpp
```diff
@@ -206,68 +206,123 @@ namespace
 
     void dict_to_add_torrent_params(dict params, add_torrent_params& p)
     {
-        // torrent_info objects are always held by a shared_ptr in the python binding
-        if (params.has_key("ti") && params.get("ti") != boost::python::object())
+        list iterkeys = (list)params.keys();
+        int const len = int(boost::python::len(iterkeys));
+        for (int i = 0; i < len; i++)
```
arvidn: I see, you turned the logic around a bit. I take it this is more efficient.

bindings/python/src/session.cpp
```diff
@@ -206,68 +206,123 @@ namespace
...
+            {
+                p.url_seeds = extract<std::vector<std::string>>(params[key]);
+                continue;
+            }
+            else if(key == "http_seeds")
+            {
+                // TODO: is it possible to create map["http_seeds"]=add_torrent_params::http_seeds?
+                // I don't how how to do it in c++ because we don't have reflections
+                // but I would like a map instead of if else and decltype does help us
+                p.http_seeds =
```
arvidn: I'm not keen on this use of ``decltype``. It would be a lot clearer to just spell out the actual type I think.
as long as http seeds are just strings, this will work because there's a ``std::vector<std::string>`` converter defined in bindings/python/src/converters.cpp

I don't understand what your comment means exactly. But it doesn't look like it should be in the code, it looks like it should be in the code review.

bindings/python/src/session.cpp
```diff
@@ -206,68 +206,123 @@ namespace
...
+                    extract<decltype(add_torrent_params::http_seeds)>(params[key]);
+                continue;
+            }
+            else if(key == "dht_nodes")
+            {
+                // TODO: did not find converter
+                //p.dht_nodes =
+                //    extract<std::vector<std::pair<std::string, int>>>(params[key]);
```
arvidn: I think it would be better to add the converter to bindings/python/src/converters.cpp
It should be straight-forward of how to do that. just follow the example of the other ones

bindings/python/src/session.cpp
Falcosc: I just want report not applied keys because it's a flexible map.
But would like to get rid of the if complexity after we fix all binding issues.

bindings/python/src/session.cpp
```diff
@@ -206,68 +206,123 @@ namespace
...
+                        // TODO: did not find converter either
+                        extract<std::pair<std::string, int>>(l[i]));
+                continue;
+            }
+            else if(key == "banned_peers")
+            {
+                // TODO: maybe the conversion of main or sub type of
+                // boost::asio::ip::basic_endpoint<boost::asio::ip::tcp>
+                // is missing because the other vector<type> definitions do work
+                p.banned_peers = extract<std::vector<lt::tcp::endpoint>>(params[key]);
```
arvidn: same here. add it to converters.cpp. However, this reminds me, the field "peers" is even more important to include here. would you mind adding it? it should use the same converter as this field

bindings/python/src/session.cpp
arvidn: actually, I see you just look up the value using ``operator[]`` anyway, so this is probably slower than the previous version. is there a reason why you don't just iterate over ``iteritems`` instead?

arvidn: I see, printing a warning or failing for unknown keys makes sense

Falcosc: Is it possible to get keyvalue of python dict? I guess you always have to do it like this. I did just copy the existing dict loop

bindings/python/src/session.cpp
Falcosc: It would be lost after the review is done because I am not able to do it anyway.

bindings/python/src/session.cpp
Falcosc: The converter is already there, I dont know which converter is missing. We have pair conversion and we have vector conversion. Thats why I have to put a todo on there, I can't fix it

bindings/python/src/session.cpp
Falcosc: set peers manually is importand?

bindings/python/src/session.cpp
arvidn: in python itself you can iterate over (key, value) - pairs, I would imagine it's possible in boost.python too.

bindings/python/src/session.cpp
arvidn: but I could answer the question in the code review.. I just don't know what you mean. what would ``map`` hold?

Falcosc: the idea is to replace ifelse with a map
p.*map[key].member = extract<map[key].type>(value)

bindings/python/src/session.cpp
arvidn: "peers" store the peer list from the last session. But I suppose when they're restored from resume data they won't be coming back this path, they will end up being passed to the function that takes an ``add_torrent_params``

bindings/python/src/session.cpp
arvidn: I see. it would be possible to build a table of member pointers and names. That's basically what boost.python does when exposing a class (like add_torrent_params for instance)

bindings/python/src/session.cpp
arvidn: the converter seems to work for me. how are you testing it?
https://github.com/arvidn/libtorrent/pull/1566

Falcosc: `[('5.6.7.8', 6881)]` did the trick.
I did just check `[['5.6.7.8', 6881]]` and `[{'5.6.7.8', 6881}]`

arvidn: ah, right. endpoints are always tuples in python. at least as far as I know.

bindings/python/src/session.cpp
Falcosc: I’m just not used to look at code with more than 50 lines per method, 1000 lines per class and more than complexity 5. ;)
Actually it will lower complexity by using less complex but more advanced assignments logics. But you are right it’s not worth it in any case because I don’t plan to refactor the 1 000 000 existing violations :D
I did just want to see how it’s look like to have a field mapping map without reflection to implement new things in an accredited code style.
Maybe we could also use a exposed add_torrent_params via boost python instead of use a handcrafted dict mapping. I guess the conversion issues where the reason for this handcrafted solution.

arvidn: in previous versions of the bindings, add_torrent_params was exposed as a dictionary to make it more accessible to python, and make it possible to do things like: ``add_torrent({'ti': ..., 'save_path': ...})``

in master though, I added a ``read_resume_data`` function that populates the add_torrent_params. so now it doesn't really scale that well to expose it manually. In fact, most of the fields you're adding here didn't used to be available in add_torrent_params in previous versions

Falcosc:
I did miss the make getter for rename_file. There is something missing but I don't understand the message. 

file_index_t conversion is already registered: `to_python_converter<lt::file_index_t, from_strong_typedef<lt::file_index_t>>();`

```
/usr/include/boost/python/object/py_function.hpp:38:33:   required from ‘PyObject* boost::python::objects::caller_py_function_impl<Caller>::operator()(PyObject*, PyObject*) [with Caller = boost::python::detail::caller<boost::python::detail::member<void(libtorrent::aux::strong_typedef<int, libtorrent::aux::file_index_tag>, const std::__cxx11::basic_string<char>&), libtorrent::torrent_info>, boost::python::return_value_policy<boost::python::return_by_value>, boost::mpl::vector2<void (&)(libtorrent::aux::strong_typedef<int, libtorrent::aux::file_index_tag>, const std::__cxx11::basic_string<char>&), libtorrent::torrent_info&> >; PyObject = _object]’
src/torrent_info.cpp:337:1:   required from here
/usr/include/boost/python/converter/registered.hpp:90:35: error: no matching function for call to ‘register_shared_ptr1(void (*)(libtorrent::aux::strong_typedef<int, libtorrent::aux::file_index_tag>, const std::__cxx11::basic_string<char>&))’
       detail::register_shared_ptr1((T*)0);
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
/usr/include/boost/python/converter/registered.hpp:81:3: note: candidate: template<class T> void boost::python::converter::detail::register_shared_ptr1(const volatile T*)
   register_shared_ptr1(T const volatile*)
   ^~~~~~~~~~~~~~~~~~~~
/usr/include/boost/python/converter/registered.hpp:81:3: note:   template argument deduction/substitution failed:
/usr/include/boost/python/converter/registered.hpp:90:35: note:   types ‘const volatile T’ and ‘void(libtorrent::aux::strong_typedef<int, libtorrent::aux::file_index_tag>, const std::__cxx11::basic_string<char>&)’ have incompatible cv-qualifiers
       detail::register_shared_ptr1((T*)0);
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~

```

Falcosc:
fix done, I did not have anymore improvements for this request in schedule.

bindings/python/src/session.cpp
```diff
@@ -206,113 +206,115 @@ namespace
...
-        list iterkeys = (list)params.keys();
-        int const len = int(boost::python::len(iterkeys));
+        list items = params.items();
+        int const len = int(boost::python::len(items));
         for (int i = 0; i < len; i++)
         {
-            std::string const key = extract<std::string>(iterkeys[i]);
+            boost::python::api::object_item item = items[i];
+            std::string const key = extract<std::string>(item[0]);
+            object const value = item[1];
```
arvidn: nice

bindings/python/src/converters.cpp
```diff
@@ -214,6 +263,8 @@ void bind_converters()
     to_python_converter<std::vector<lt::udp::endpoint>, vector_to_list<lt::udp::endpoint>>();
     to_python_converter<std::vector<std::pair<std::string, int>>, vector_to_list<std::pair<std::string, int>>>();
 
+    to_python_converter<std::map<lt::file_index_t, std::string>, map_to_dict<lt::file_index_t, std::string>>();
```
arvidn: are there any API function that deals with ``std::map``?

bindings/python/test.py
```diff
@@ -57,6 +57,46 @@ def test_torrent_handle(self):
 		self.h.prioritize_pieces([(0, 1)])
 		self.assertEqual(self.h.piece_priorities(), [1])
 
+	def test_unknown_torrent_parameter(self):
```
arvidn: +1

bindings/python/src/converters.cpp
Falcosc: not anymore for python->C++ after I figured out that rename_files was just a setter. But c++->python is used for rename_files

arvidn:
looks good. what is the converter for ``std::map`` used for? and in the Changelog, you say "invalid torrent parameter", but you refer to an invalid key in the add torrent dict, right? I think that could be said a bit clearer

ChangeLog
Falcosc: need to be changed

Falcosc:
I hope I will find the windows bug after I changed the asserts, because this could be a little bit difficult to debug.

bindings/python/src/converters.cpp
arvidn: if this is not needed anymore, is there a reason to leave it in?

arvidn:
lgtm, apart from the std::map python converter, not sure what it's doing there

Falcosc:
I thought about leaving it in because we may need it in future but the map_to_dict converter was untestet and this makes it useless.

bindings/python/src/converters.cpp
```diff
@@ -109,6 +109,41 @@ struct tuple_to_pair
     }
 };
 
+template<class T1, class T2>
+struct dict_to_map
```
arvidn: what about dict_to_map, is it used?

Falcosc: yes, for rename: https://github.com/arvidn/libtorrent/pull/1541/files#diff-9a8153ff01a5fbbc984a9b19dcb49e22R313

That was the storry where I wonder why the getter is not working after I fugured out that it was just a method call.

I have to say my work with c++ is very unproductive, but at least something got accomplished :)

Falcosc:
@arvidn this PR is solves https://github.com/arvidn/libtorrent/issues/606

arvidn:
I believe that ticket refers to RC_1_1. in master this is already solved by actually exposing the ``add_torrent_params`` type directly.

arvidn:
just re-read that ticket. I solved most of that issue by exposing the ``add_torrent_params`` type, but apparently some parts of it is inaccessible. presumably because fields are not exposed in a way that supports converters.

Falcosc:
Now I lost the knowlage why we have the class binding https://github.com/arvidn/libtorrent/blob/master/bindings/python/src/session.cpp#L631 and the python dict https://github.com/arvidn/libtorrent/blob/master/bindings/python/src/session.cpp#L349

Do you remember the reason for that?

arvidn:
the dictionary was the original python binding for ``add_torrent_params``. when I added ``read_resume_data()`` (which is a much better way of separating resume data and adding the torrent), it contained types that were hard to bind manually to a dict, so I exposed the type properly. Now both are supported, but the proper type (not the dict) supports everything from the resume data, the dict doesn't.

Issue #1543, opened 2017-01-13
master branch python bindings not building due to missing c++11 flag


cas--:
This is on `master` branch a31ba79

The following from [build logs](https://launchpadlibrarian.net/302365971/buildlog_ubuntu-xenial-i386.libtorrent-rasterbar_1.2.0dev8739+a31ba79+201701131732~ubuntu16.04.1_BUILDING.txt.gz) when trying to build the python bindings:
```
/usr/include/c++/5/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
```

The rest of libtorrent is building fine and due can see `-std=c++11` and `-DHAVE_CXX11=1` are in the compiler arguments.

Falcosc:
I can't reproduce it with my gcc version because c++11 is already enabled by default. I try to guess something but you may need to apply and test some of my patches.

cas--:
Looks like the `CXXFLAGS` is somehow not setup correctly in the python bindings `Makefile`:

```
CXX = g++ -std=c++11
CXXCPP = g++ -E -std=c++11
CXXDEPMODE = depmode=gcc3
CXXFLAGS = -g -O2 -ftemplate-depth=128 -Wno-format-zero-length -fvisibility=hidden -fvisibility-inlines-hidden
```

Falcosc:
Did you try the patch? I did a hardcoded CXXFLAGS change which I did not see in your config.report.txt
But never mind I will change it to CXX anyway because AX_CXX_COMPILE_STDCXX_11 does put it into CXX if your compiler need this option.

cas--:
The patch is a hack and just guessing so no. Obviously if I manually modify the compile flags it works but I think it's `CFLAGS` that needs settings for C++11

Falcosc:
I guess CXX and CXXFLAGS are enoght, python did just not see them.

cas--:
This explains the whole problem: https://github.com/arvidn/libtorrent/issues/870

The fix only was for using `setup.py`...

cas--:
@arvidn does this look ok to add to `configure.ac`

```
AX_CHECK_COMPILE_FLAG([-std=c++11], [
  CPPFLAGS+=" -std=c++11"
], [
  echo "CPP compiler cannot compile C11 code"
  exit -1
])
```

Falcosc:
[...]

PR #1545, opened 2017-01-13
fix cxxflags for python binding build

Falcosc:
#1543

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,5 @@
 	* fix cxxflags for python binding build
+	* fix crash caused by empty bitfield 
 	* removed disk-access-log build configuration
 	* removed mmap_cache feature
 	* strengthened type safety in handling of piece and file indices
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,3 @@
-	* fix cxxflags for python binding build
 	* fix crash caused by empty bitfield 
 	* removed disk-access-log build configuration
 	* removed mmap_cache feature
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* fix cxxflags for python binding build
 	* removed disk-access-log build configuration
 	* removed mmap_cache feature
 	* strengthened type safety in handling of piece and file indices
--- a/bindings/python/Makefile.am
+++ b/bindings/python/Makefile.am
@@ -33,7 +33,7 @@ EXTRA_DIST = \
 if ENABLE_PYTHON_BINDING
 
 all-local:
-	$(PYTHON) $(srcdir)/setup.py build
+	CXXFLAGS="$(CXXFLAGS)" $(PYTHON) $(srcdir)/setup.py build
 
 install-exec-local:
 	$(PYTHON) $(srcdir)/setup.py install @PYTHON_INSTALL_PARAMS@
--- a/bindings/python/Makefile.am
+++ b/bindings/python/Makefile.am
@@ -33,7 +33,7 @@ EXTRA_DIST = \
 if ENABLE_PYTHON_BINDING
 
 all-local:
-	CXXFLAGS="$(CXXFLAGS)" $(PYTHON) $(srcdir)/setup.py build
+	CXX="$(CXX)" $(PYTHON) $(srcdir)/setup.py build
 
 install-exec-local:
 	$(PYTHON) $(srcdir)/setup.py install @PYTHON_INSTALL_PARAMS@
--- a/configure.ac
+++ b/configure.ac
@@ -99,7 +99,7 @@ AX_PTHREAD()
 LIBS="$PTHREAD_LIBS $LIBS"
 CFLAGS="$PTHREAD_CFLAGS $CFLAGS"
 CC="$PTHREAD_CC"
-CXXFLAGS="$CXXFLAGS -ftemplate-depth=128 -Wno-format-zero-length -std=c++11"
+CXXFLAGS="$CXXFLAGS -ftemplate-depth=128 -Wno-format-zero-length"
 
 AS_ECHO "Checking for visibility support:"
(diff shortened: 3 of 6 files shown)


bindings/python/Makefile.am
```diff
@@ -33,7 +33,7 @@ EXTRA_DIST = \
 if ENABLE_PYTHON_BINDING
 
 all-local:
-	$(PYTHON) $(srcdir)/setup.py build
+	CXXFLAGS="$(CXXFLAGS)" $(PYTHON) $(srcdir)/setup.py build
```
arvidn: I was under the impression that the python dist-utils already took care of specifying ``-std=c++11``. Would it still break without this change?

configure.ac
```diff
@@ -99,7 +99,7 @@ AX_PTHREAD()
 LIBS="$PTHREAD_LIBS $LIBS"
 CFLAGS="$PTHREAD_CFLAGS $CFLAGS"
 CC="$PTHREAD_CC"
-CXXFLAGS="$CXXFLAGS -ftemplate-depth=128 -Wno-format-zero-length"
+CXXFLAGS="$CXXFLAGS -ftemplate-depth=128 -Wno-format-zero-length -std=c++11"
```
arvidn: This is hardcoding GCC specifics. I was under the impression that there already are some C++11 detection logic in place in here, by auto conf

arvidn: line 122

ChangeLog
```diff
@@ -1,3 +1,4 @@
+	* fix cxxflags for python binding build
```
arvidn: I don't think this warrants a Changelog entry. There hasn't been a release where this was broken

bindings/python/Makefile.am
arvidn: I guess not, I can't find anything. However, I think it would be better to forward the result of ``AX_CXX_COMPILE_STDCXX_11([noext], [mandatory])`` instead.

cas--:
I have updated my issue with more details. I am not sure these changes are correct

bindings/python/Makefile.am
Falcosc: But AX_CXX_COMPILE_STDCXX_11 would not include the flags which are configured with configure.ac. For example `-ftemplate-depth=128 -Wno-format-zero-length"` would be missing in python because we leave the make scope by calling $(PYTHON)
I would guess we may need even more then CXXFLAGS="$(CXXFLAGS)" instead of less, isn't it?

And if I use AX_CXX_COMPILE_STDCXX_11 I would need to write
`CXX="$(CXX)" $(PYTHON) $(srcdir)/setup.py build`
because it add the flag to CXX https://github.com/sandstorm-io/capnproto/blob/master/c%2B%2B/m4/ax_cxx_compile_stdcxx_11.m4 and python does not see it because it is not in make scope

Falcosc: Or should I write 
`CXX="$(CXX)" CXXFLAGS="$(CXXFLAGS)" $(PYTHON) $(srcdir)/setup.py build`
without hardcoding gcc specfifics at CXXFLAGS

cas--: Please read my update to the original issue, this is not the issue. The bindings Makefile has the correct flags.

Falcosc: and python does not find them because it leaves the make scope. I just want to know how much of the make scope should be exposed to python by adding them in front of the python call.

ChangeLog
Falcosc: Sure, didn't note that our c++11 update was just so recently

configure.ac
Falcosc: thanks, I did wonder how it did work in past.

arvidn:
I think a better fix would be to include ``CXX`` in the ``compile_flags.in`` file. then ``setup.py`` could also be run manually an behave correctly

arvidn:
also, there is a travis test that build using the makefiles. does it not enable the python build? if not, it probably should

Falcosc:
You can't include CXX in compile_flags.in because it starts with g++ which cant be inserted mid line.

```
x86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -fno-strict-aliasing -D_FORTIFY_SOURCE=2 -g -fstack-protector-strong -Wformat -Werror=format-security -fPIC -I../../include -I/usr/include -I/usr/include -I/usr/include/python2.7 -c /root/fas/libtorrent/bindings/python/src/utility.cpp -o build/temp.linux-x86_64-2.7/root/fas/libtorrent/bindings/python/src/utility.o -DTORRENT_USE_OPENSSL -DTORRENT_USE_LIBCRYPTO -DBOOST_ASIO_HAS_STD_CHRONO=1 -DBOOST_EXCEPTION_DISABLE -DBOOST_ASIO_ENABLE_CANCELIO -DTORRENT_LINKING_SHARED g++
x86_64-linux-gnu-gcc: error: g++: Datei oder Verzeichnis nicht gefunden
```
We have to use CXX as environment var.
Export or include_env.in or extract and std=c++11 to a CXXFLAGS/CPPFLAGS

cas--:
The only reason that works is because there is a workaround in setup.py to copy `CXX` to `CC` so it still doesn't seem correct to me.

cas--:
I think supplying `CXX` to `setup.py` via a file is the correct course. So we always set `CC` in setup.py like so: `os.environ['CC'] = (CXX from file)`.

Also should add `CXXFLAGS` to `compile_flags.in`. Should `CPPFLAGS` be in there at all?

arvidn:
sure. so some other variable that ``-std=c++11`` ends up in could be included in  ``compile_flags.in``. I'm sure the autoconf macro sets some other output too

arvidn:
I'm tempted to close this, as @cas-- has another PR now

PR #1548, opened 2017-01-14
Fix autotool.sh warning if builddir exists

cas--:
Trivial fix for ```mkdir: cannot create directory `build-aux`: File exists```


arvidn:
lgtm

PR #1552, opened 2017-01-14
fix make_torrent py 1.1

arvidn:


arvidn:
@Falcosc would you mind if I pull this ahead of your fix, and you can rebase and re-enable the make_torrent.py test?

Falcosc:
sure

Falcosc:
@arvidn the test will not work on master, because add_file is deprecated there. Are you sure that I should enable it there?

Would be make more sense to enable it with a binding fix, unfortunately I wasn't able to change the binding to accept the new add_files method.

PR #1553, opened 2017-01-14
[#1543] Fix missing C++11 flag for building python bindings

cas--:


cas--:
I just realised I forgot to remove your `os.environ` additions to setup.py. What do you think?

arvidn:
if you think they're redundant, I'm not opposed to removing them.

cas--:
Done. If users want to change the compiler there is still the setup.py `--compiler` option.

PR #1554, opened 2017-01-14
Fix ar 'modifier ignored' build warning

cas--:
Another trivial fix but nice to squash the random build warnings :stuck_out_tongue: 

 * This only applies to AR_FLAGS and not ARFLAGS.
 * afaik ARFLAGS is the new convention and this warning is fixed
   upstream in libtool but may be a while before OSes catchup.


arvidn:
should this go into RC_1_1 as well?

Issue #1217, opened 2016-10-15
Runtime warning, then importing python libtorrent module (RC_1_1)


barmalej2:
libtorrent version (or branch):RC_1_1 git branch
boost version: 1.61.0
python version: 2.7.12
platform/architecture: armv5tel GNU/Linux, kernel 2.6.31.8
compiler and compiler version: gcc 4.9.2, 
Step to reproduce case. Import python libtorrent:
`python -c "import libtorrent; print libtorrent.version"`
Expected output is only libtorrent version.
Actual output is:

> -c:1: RuntimeWarning: to-Python converter for boost::shared_ptrlibtorrent::alert already registered; second conversion method ignored.
> 1.1.1.0

cas--:
A little research turned up this project encountering the [same issue](https://github.com/BVLC/caffe/pull/4069/files) and seems to due to boost version differences.

arvidn:
iirc, the underlying problem is that different version of boost.python or perhaps when building against different versions of python (2 vs. 3) has different behavior here. A while ago I spent quite some time trying to resolve this. as far as I know, at least I got it to work everywhere.

However, the long term solution is to get rid of the ``shared_ptr<>`` as part of the API, which puts all of this to reset once and for all. This is already the case in master and will be the fix in the next major release. I am open to patches that fixes this warning in the 1.1 branch, but I'm not sure I'm prepared to put much more work into it myself.

Falcosc:
@barmalej2 What did you think about it? I don't like the idea of changing the RC_1_1 api to solve this issue. And it would may be take more effort than switching your python scripts to libtorrent 1.2.

cas--:
The problem here is why is this occuring only in certain Linux configurations... I.e. I don't see this on my Ubuntu systems but iirc it's seen on Arch

Falcosc:
I would guess because on the seen Arch system there was a boost version difference. But I don't know.

arvidn:
would it make sense to remove the problematic alert-binding for the upcoming RC_1_1 release? That way at least some other deprecated functions could be preserved

cas--:
It seems this is occurring with boost >=1.61 as I can replicate this on Ubuntu Yaketty.

@arvidn did you see this workaround that `caffe` used: https://github.com/BVLC/caffe/pull/3866/files

barmalej2:
[...]

Issue #448, opened 2016-02-02
Cant build libtorrent 1.0.81 / 1.1.0 - error: Boost.Python library not found.


lethanhviet:
Attempting to build libtorrent on clean debian 8 server with boost 1.55. I have installed libboost-python-dev, but libtorrent during configure cannot find it.

root@Debian-83-jessie-64-minimal ~/libtorrent # ./configure --enable-python-binding --with-libgeoip --with-libiconv --disable-debug --enable-logging=default --with-boost-libdir=/usr/lib/x86_64-linux-gnu
configure: WARNING: unrecognized options: --with-libgeoip
checking whether make supports nested variables... yes

Building libtorrent-rasterbar 1.1.0

Checking for a C/C++ compiler to use:
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables...
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking whether gcc understands -c and -o together... yes
checking how to run the C preprocessor... gcc -E
checking whether gcc and cc understand -c and -o together... yes
checking for g++... g++
checking whether we are using the GNU C++ compiler... yes
checking whether g++ accepts -g... yes
checking how to run the C++ preprocessor... g++ -E
checking whether g++ understands -c and -o together... yes

Checking system type:
checking build system type... x86_64-unknown-linux-gnu
checking host system type... x86_64-unknown-linux-gnu
checking target system type... x86_64-unknown-linux-gnu

Initializing Automake:
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /bin/mkdir -p
checking for gawk... no
checking for mawk... mawk
checking whether make sets $(MAKE)... yes
checking for style of include used by make... GNU
checking dependency style of gcc... gcc3
checking dependency style of g++... gcc3

Initializing Libtool:
checking how to print strings... printf
checking for a sed that does not truncate output... /bin/sed
checking for grep that handles long lines and -e... /bin/grep
checking for egrep... /bin/grep -E
checking for fgrep... /bin/grep -F
checking for ld used by gcc... /usr/bin/ld
checking if the linker (/usr/bin/ld) is GNU ld... yes
[...]

PR #1555, opened 2017-01-14
[#1217] Fix RuntimeWarning registering pointer with Python

cas--:
For Boost >=1.60 the following warning is output importing libtorrent in Python:

    RuntimeWarning: to-Python converter for boost::shared_ptrlibtorrent::alert already registered; second conversion method ignored.

### DIFF
--- a/bindings/python/src/alert.cpp
+++ b/bindings/python/src/alert.cpp
@@ -170,9 +170,11 @@ void bind_alert()
     using boost::noncopyable;
 #ifndef TORRENT_NO_DEPRECATE
     typedef boost::shared_ptr<alert> alert_holder;
-#if BOOST_VERSION >= 106000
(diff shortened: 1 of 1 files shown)


cas--:
This should also be applied to RC_1_0

arvidn:
do you think the boost version check is still necessary?

cas--:
I think so as there is an else condition that I didn't see before.

arvidn:
the "else" is for TORRENT_NO_DEPRECATE

cas--:
Hah they always catch me out! 

Ok I see that was added originally to fix the warning in #448

cas--:
Removed, correctly indented code and updated commit details.

arvidn:
thanks! lgtm

PR #1557, opened 2017-01-14
fix torrent peer class index overflow

Falcosc:
this resolves:
- session start fast resume assert
- fast resume data save error on session close
- throtteling accuracy get improved

### DIFF
--- a/include/libtorrent/peer_class.hpp
+++ b/include/libtorrent/peer_class.hpp
@@ -47,7 +47,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent
 {
-	typedef boost::uint16_t peer_class_t;
+	typedef boost::uint32_t peer_class_t;
 
 	struct peer_class_info
 	{
--- a/include/libtorrent/peer_class.hpp
+++ b/include/libtorrent/peer_class.hpp
@@ -47,7 +47,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent
 {
-	typedef boost::uint8_t peer_class_t;
+	typedef boost::uint16_t peer_class_t;
 
 	struct peer_class_info
 	{
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,5 @@
+	* fix torrent connection index overflow, this fix several errors related to
+	  working with more then 256 torrents like resume data and connection handling
 	* don't load user_agent and peer_fingerprint from session_state
 	* fix file rename issue with name prefix matching torrent name
 	* fix division by zero when setting tick_interval > 1000
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,4 @@
-	* fix torrent connection index overflow, this fix several errors related to
-	  working with more then 256 torrents like resume data and connection handling
+	* fix per-torrent rate limits for >256 peer classes
 	* don't load user_agent and peer_fingerprint from session_state
 	* fix file rename issue with name prefix matching torrent name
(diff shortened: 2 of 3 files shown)


arvidn:
this essentially only affects you when you set per-torrent upload/download rate limits. right? I think the Changelog entry can be simplified to "fix per-torrent rate limits and >256 peer classes". what do you think?

Falcosc:
No, I can't start a session if asserts are enabled because the overflow let the assert of new peers fail: https://github.com/arvidn/libtorrent/blob/RC_1_1/src/peer_class.cpp#L94

And when I close the session I got illegal pointer because without the assert multible torrents use same vector entries and during close session something got confused.

As long as you don't stop or start the session all will be ok without the bugfix.

arvidn:
how and why are 256 peer classes created?
presumably, if no peer classes were created, you would not have a problem, right?

Falcosc:
On session create with torrent resume it creates one class per torrent
Name is the torrent name:
https://github.com/arvidn/libtorrent/blob/a31ba7972d31714518abac157206fa662420bb5c/src/torrent.cpp#L8047

But I didn't know why it does this.

Falcosc:
I see, it's because all these torrents have torrent based limits

include/libtorrent/peer_class.hpp
```diff
@@ -47,7 +47,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent
 {
-	typedef boost::uint8_t peer_class_t;
+	typedef boost::uint16_t peer_class_t;
```
arvidn: sorry for not mentioning this earlier. But this should probably be a uint32_t.

arvidn:
actually.. thinking about this a bit more. this patch will alter the ABI, and I try to maintain ABI compatibility within minor versions (i.e. 1.1.1 -> 1.1.2).
This change should definitely be made, but perhaps not in RC_1_1. I think the short term fix need to maintain ABI compatibility. I think there are 2 options:

1. leave this as ``uint8_t`` but make sure torrents gracefully handle the case where they fail to create a peer-class. i.e. make the rate limits stop working, but no asserts firing.

2. leave *peer_class_t* as a ``uint8_t``, but introduce a new peer class index that's only used internally, which is 32 bits. That way the ABI stays the same, users just can't refer to peer classes >255.

option 2 is probably preferable, but may be a little bit more work.

include/libtorrent/peer_class.hpp
Falcosc: sure? because of having more then 65536 torrents with per torrent limit or what is the matter?

Falcosc:
Compatibility is a difficult topic, But I do just worry about how long we are stuck with RC_1_1 because the API change to 1.2 is alot to do for the clients.
Where did the user get access to peer_class_t, maybe it's possible to create a 8 bit getter.

include/libtorrent/peer_class.hpp
arvidn: right 2^16 is an arbitrary limit, and most likely the most restrictive one when adding a lot of torrents

arvidn:
actually, it looks like it's not exposed directly to the public API. (perhaps I thought of this when I implemented it). The public representation of a ``peer_class_t`` is just an ``int``, so the easiest would probably be to use that internally as well.

in master, peer_class_t should be a strong_typedef of an ``int`` and probably exposed to the user.

So, this patch most likely does not alter the ABI then. I still think it should be an ``int`` internally.

arvidn:
you need to rebase on top of the Changelog too

arvidn:
lgtm

PR #1558, opened 2017-01-14
use /dev/random as random number source on macOS

arvidn:


aldenml:
Well, this PR have a few unintended consequences :)

now it requires `libssl` if using `libcrypto` due to missing `SSL_library_init` and others....

arvidn:
I see. perhaps I misunderstand the libcrypto error handling. is ``ERR_get_error()`` a libssl call?
Do I need a separate error_category for libcrypto?

aldenml:
looking at it (to fix it), can't say anything yet

aldenml:
see comments at https://github.com/arvidn/libtorrent/pull/1570

arvidn:
the reason libssl ended up being pulled in was because I assumed the ssl error_category from boost.asio was the appropriate one

aldenml:
just arrived to the same conclusion :)

PR #1559, opened 2017-01-15
fixing sign-conversion warnings, part 5, minor refactor

aldenml:


src/string_util.cpp
```diff
@@ -161,9 +161,9 @@ namespace libtorrent
 	{
 		if (str == nullptr) return nullptr;
 		int const len = int(std::strlen(str));
-		char* tmp = static_cast<char*>(std::malloc(len + 1));
+		char* tmp = static_cast<char*>(std::malloc(std::size_t(len + 1)));
```
aldenml: I will refactor this

include/libtorrent/kademlia/routing_table.hpp
```diff
@@ -58,7 +59,7 @@ namespace libtorrent { namespace dht
 {
 struct dht_logger;
 
-typedef std::vector<node_entry> bucket_t;
+typedef aux::vector<node_entry> bucket_t;
```
aldenml: I will revert this, seems to me like a pseudo-public type

arvidn: hm.. I don't think it's supposed to be..

arvidn:
sorry. I broke master in my merge. would you mind rebasing?

include/libtorrent/kademlia/routing_table.hpp
aldenml: ok, I will keep it as `aux::`

aldenml:
sure, in a minute

src/http_tracker_connection.cpp
```diff
@@ -504,7 +504,7 @@ namespace libtorrent
 			else
 #endif
 			{
-				resp.peers4.reserve(len / 6);
+				resp.peers4.reserve(std::size_t(len / 6));
```
arvidn: thinking a bit about why this is a warning, presumably it's because ``len`` could be negative, which can't be represented by the unsigned variable. Maybe it would make sense to have a function perform these kinds of conversions, that also assert that the incoming value is not negative.

src/string_util.cpp
```diff
@@ -47,12 +47,12 @@ namespace libtorrent
 	// lexical_cast's result depends on the locale. We need
 	// a well defined result
 	std::array<char, 4 + std::numeric_limits<std::int64_t>::digits10>
-		to_string(std::int64_t n)
+		to_string(std::int64_t const n)
 	{
 		std::array<char, 4 + std::numeric_limits<std::int64_t>::digits10> ret;
 		char *p = &ret.back();
 		*p = '\0';
-		std::uint64_t un = n;
```
arvidn: can you not do: ``std::uint64_t un(n);`` here?

src/storage.cpp
```diff
@@ -864,7 +864,7 @@ namespace libtorrent
 
 		// indices of all files we ended up copying. These need to be deleted
 		// later
-		aux::vector<bool, file_index_t> copied_files(f.num_files(), false);
+		aux::vector<bool, file_index_t> copied_files(std::size_t(f.num_files()), false);
```
arvidn: hm.. couldn't this just be using ``f.file_end()``?

arvidn: or maybe I argued against using the index type for size, or maybe the constructor isn't overridden

src/http_tracker_connection.cpp
aldenml: not sure, but...do you have a name suggestion and a place to put it?

src/storage.cpp
aldenml: both :) you didn't like the index type for size and no constructor is overridden

src/string_util.cpp
aldenml: last time in another case (PR), clang was still complaining.

src/http_tracker_connection.cpp
arvidn: well, not for the PR, but it would be something very similar to ``numeric_cast<>()``, but an assert instead of a run-time check

PR #1560, opened 2017-01-15
fix python binding issue with string_view

arvidn:


arvidn:
@Falcosc it turned out the python binding issue was because of the transition to string_view. Since it's fundamentally a reference type, it has to be captured in a ``std::string`` in the binding to be passed on. There may be other places with the same problem.

Falcosc:
Sorry, I didn't understand it without c++ knowledge.
Is it possible to describe when to use a fuction reference like `.def("num_pieces", &create_torrent::num_pieces)` and when to use `.def("add_tracker", add_tracker, (arg("announce_url"), arg("tier") = 0))`

Thanks

arvidn:
the symptom of it not working is the error you get when trying to call it in master right now. the second form introduces a wrapper function that takes a ``std::string``, which captures the python string for the duration of the function invocation to be able to pass it in as a string_view.

the reason why python's own internal representation of the string cannot be passed in as the string_view is because it's likely not stored as a utf-8 string internally, nor is the internal representation exposed. We can only get at the strings by asking for a copy in a specific encoding.

PR #1563, opened 2017-01-15
call set_proxy_settings during session_impl::setup_listener

aldenml:


src/session_impl.cpp
```diff
@@ -1715,6 +1715,7 @@ namespace aux {
 		}
 
 		ret.udp_sock->set_force_proxy(m_settings.get_bool(settings_pack::force_proxy));
+		ret.udp_sock->set_proxy_settings(proxy());
```
arvidn: would you mind adding a comment explaining that this has to be called for the ``force_proxy`` settings to take effect? (assuming my understanding is correct). perhaps add more details too if you have a deeper understanding

aldenml: sure, but let me get the confirmation that this fix the problem first, no need to triggers more CI jobs :)

arvidn:
should I merge this? I expect this patch should go into RC_1_1 as well, right?

aldenml:
I rather wait for confirmation (and the comment), but it could take a while. Your decision.

arvidn:
I was under the impression that @alex-druzhinin had confirmed.

aldenml:
I didn't release any beta, so maybe it was "source code" diff confirmation.

aldenml:
@arvidn added the comment and was confirmed it works

PR #1564, opened 2017-01-15
add python binding for generate_fingerprint

arvidn:


cas--:
Thought it might be an easy one then forgot about it! :+1:

PR #1565, opened 2017-01-15
Issue/silence py depr warnings

cas--:
There is also a commit dealing with debug flags.


arvidn:
does this need to be back-ported too?

cas--:
The suppressing of warning yep. 

A few issues with debug commit. The debug flag `-Og` is a gcc 4.8 onwards feature so that would need changed back. Also switching to non-debug python build might catch out packagers, unless a changelog entry would suffice? I could create a separate PR?

PR #1568, opened 2017-01-16
refactor entry::dictionary_type

ajax16384:
consolidate conditional std::map string code
enable C++14 string_view std::map::find for VS2015

### DIFF
--- a/include/libtorrent/entry.hpp
+++ b/include/libtorrent/entry.hpp
@@ -99,11 +99,9 @@ namespace libtorrent
 		template<class T> using map_string = std::map<std::string, T, aux::strview_less>;
 #else
 		template<class T>
-		class map_string : public std::map<std::string, T>
+		struct map_string : std::map<std::string, T>
 		{
-		private:
 			using base = std::map<std::string, T>;
-		public:
 
 			typename base::iterator find(const string_view& key)
 			{
--- a/include/libtorrent/entry.hpp
+++ b/include/libtorrent/entry.hpp
@@ -83,9 +83,9 @@ namespace libtorrent
 #endif
 	struct bdecode_node;
 
-#if __cplusplus > 201103
 	namespace aux
 	{
+#if (__cplusplus > 201103) || (defined _MSC_VER && _MSC_VER >= 1900)
 		// this enables us to compare a string_view against the std::string that's
 		// held by the std::map
 		struct strview_less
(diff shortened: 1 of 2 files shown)


include/libtorrent/entry.hpp
```diff
@@ -95,8 +95,30 @@ namespace libtorrent
...
 			{ return rhs < lhs; }
 		};
-	};
+
+		template<class T>
+		class map_string : public std::map<std::string, T, aux::strview_less>
+		{
+		};
+#else
+		template<class T>
```
arvidn: since this is master, C++11 can be assumed, and this else-block should probably just be removed. In fact, just removing the conditionals from the previous version is probably even simpler

ajax16384: ok, i'll wait till appveyor and travis finish their jobs and try to remove condition

ajax16384: original std::map::find string_view commit ( 48a647a1696e3029eed2fdd68fa9a691b9c79f10 ) assumes that it will work only at C++14

arvidn: oh, I see

include/libtorrent/entry.hpp
```diff
@@ -95,8 +95,30 @@ namespace libtorrent
 			bool operator()(T1 const& rhs, T2 const& lhs) const
 			{ return rhs < lhs; }
 		};
-	};
+
+		template<class T>
+		class map_string : public std::map<std::string, T, aux::strview_less>
```
arvidn: this could be a ``using`` statement.
``template <typename T>
map_string = std::map<std::string, T, aux::strview_less>;``

include/libtorrent/entry.hpp
```diff
@@ -95,8 +95,30 @@ namespace libtorrent
...
+		{
+		};
+#else
+		template<class T>
+		class map_string : public std::map<std::string, T>
+		{
+		private:
+			using base = std::map<std::string, T>;
+		public:
+			typename base::iterator find(const string_view& key)
```
arvidn: you should probably pull in the ``find()`` overloads from the base too: ``using base::find;``

include/libtorrent/entry.hpp
ajax16384: ok

include/libtorrent/entry.hpp
ajax16384: ok

arvidn:
lgtm

arvidn:
I guess pulling in the base class' ``find()`` was poor advice, that cause ambiguities.

include/libtorrent/entry.hpp
```diff
@@ -95,8 +95,28 @@ namespace libtorrent
...
+		template<class T> using map_string = std::map<std::string, T, aux::strview_less>;
+#else
+		template<class T>
+		class map_string : public std::map<std::string, T>
+		{
+		private:
+			using base = std::map<std::string, T>;
+		public:
+
+			typename base::iterator find(const string_view& key)
```
arvidn: I'm not sure I understand why you get that compilation error. does it build for you locally? maybe you should try taking ``string_view`` by value instead

include/libtorrent/entry.hpp
```diff
@@ -95,8 +95,28 @@ namespace libtorrent
 			bool operator()(T1 const& rhs, T2 const& lhs) const
 			{ return rhs < lhs; }
 		};
-	};
+
+		template<class T> using map_string = std::map<std::string, T, aux::strview_less>;
+#else
+		template<class T>
+		class map_string : public std::map<std::string, T>
```
arvidn: if you make it a ``struct`` instead you don't have to say ``public`` here nor below (provided you put the public members first)

include/libtorrent/entry.hpp
ajax16384: builds without problem. 
the insert to emplace change was made for optimisation purpose.

include/libtorrent/entry.hpp
ajax16384: ok

PR #1570, opened 2017-01-16
fix openssl build on windows

arvidn:


aldenml:
Just including `<boost/asio/ssl.hpp>` or `<boost/asio/ssl/error.hpp>` requires linking to `libssl` because of the internal machinery of `boost.asio` to setup `get_ssl_category`. Strictly speaking, using directly `libcrypto` is not related to `boost.asio` so I think the right solution is to create a custom category. But since it will only be for "random", I find it kind of a weak reason to do it.

If you remove this include and revert to `throw system_error(error_code(int(::ERR_get_error()), system_category()));`, this will solve all the issues in the meantime

arvidn:
I should probably not use ``ERR_get_error()`` though, but perhaps hardcode some ``errno`` error and use ``generic_category()``

aldenml:
I suppose you can do it in this PR, what do you think?

arvidn:
yes, I will update

PR #1577, opened 2017-01-18
created numeric_cast<> and using it in stack_allocator

aldenml:


aldenml:
ping here :)

include/libtorrent/aux_/vector.hpp
```diff
@@ -143,6 +143,20 @@ namespace libtorrent { namespace aux {
 		return {vec.data() + static_cast<type>(begin), vec.data() + static_cast<type>(end)};
 	}
 
+	// TODO: find a better place for this function
```
arvidn: I think there may be a better name too. there's a boost function called ``numeric_cast`` which does runtime checking

aldenml: interesting, didn't know about `boost::numeric_cast`, do you want to do something else about it? any other name suggestion?

PR #1579, opened 2017-01-19
torrent_info::is_valid fix

arvidn:


sledgehammer999:
I confirm that it works.

PR #1583, opened 2017-01-19
make handler exchange without copy->clone mem allocation

ajax16384:


aldenml:
What's the difference/advantage?

ajax16384:
according to profiler old implementation copies whole handler content (with redudant heap allocation):
````c++
 - libtorrent::utp_socket_impl::maybe_trigger_receive_callback	2.16 %	0.03 %	171	2	
   - libtorrent::utp_stream::on_read	2.11 %	0.00 %	167	0	
     - std::bind<void,std::function<void __cdecl(boost::system::error_code const &,unsigned int)> &,boost::system::error_code const &,unsigned int &>	1.18 %	0.04 %	93	3	
       - std::_Func_impl<libtorrent::aux::allocating_handler<std::_Binder<std::_Unforced,void (__thiscall libtorrent::peer_connection::*)(boost::system::error_code const &,unsigned int),std::shared_ptr<libtorrent::peer_connection>,std::_Ph<1> const &,std::_Ph<2> const &>,336>,std::allocator<int>,void,boost::system::error_code const &,unsigned int>::_Copy	1.14 %	0.00 %	90	0	
         - std::_Func_impl<libtorrent::aux::allocating_handler<std::_Binder<std::_Unforced,void (__thiscall libtorrent::peer_connection::*)(boost::system::error_code const &,unsigned int),std::shared_ptr<libtorrent::peer_connection>,std::_Ph<1> const &,std::_Ph<2> const &>,336>,std::allocator<int>,void,boost::system::error_code const &,unsigned int>::_Clone<void>	1.14 %	0.00 %	90	0	
           - operator new	1.13 %	0.01 %	89	1	
             + _malloc_base	1.10 %	0.00 %	87	0	
````
the new one just tries to move it to bind reducing copy and allocate

**Comment by aldenml** [comment]
Thanks, I see it. Why the need of the temporary variable? ins't calling `std::move(...)` possible?

**Comment by ajax16384** [comment]
s->m_*_handler should definetly have nullptr value, and after std::move object may stay at valid but not specified state.

**Comment by aldenml** [comment]
true, but the ` = nullptr;` was already there, is that causing a heap work too?

**Comment by ajax16384** [comment]
= nullptr might delete allocated funct copy

**Comment by aldenml** [comment]
yes, but it was already moved, right? I expect the handler have none inside

File: src/utp_stream.cpp [hunk 1]
```diff
@@ -898,8 +898,9 @@ void utp_stream::on_read(void* self, size_t bytes_transferred
 
 	TORRENT_ASSERT(s->m_read_handler);
 	TORRENT_ASSERT(bytes_transferred > 0 || ec || s->m_impl->m_null_buffers);
-	s->m_io_service.post(std::bind<void>(s->m_read_handler, ec, bytes_transferred));
-	s->m_read_handler = nullptr;
+	read_handler_t read_handler{nullptr};
```
arvidn: you don't need to pass in ``nullptr`` here

arvidn:
@aldenml I don't believe it's possible to move an object into a ``bind``-capture, and in C++11 it's not possible to move into a lambda.

aldenml:
@arvidn thanks for the comment, my questions are because I don't see the difference between
```cpp
s->m_io_service.post(std::bind<void>(std::move(s->m_read_handler), ec, bytes_transferred));
s->m_read_handler = nullptr;
```
and
```cpp
read_handler_t read_handler{nullptr};
read_handler.swap(s->m_read_handler);
s->m_io_service.post(std::bind<void>(std::move(read_handler), ec, bytes_transferred));
```
but I could be missing something :) maybe some by-copy detail

arvidn:
ah, I see your point. the only difference I can think of (which may be a bit far fetched), is that if ``post()`` throws an exception ``m_read_handler`` may not have been cleared. Whether that matter or not is a different story.

@ajax16384 what do you think? I kind of like the idea of saving a line of code.

ajax16384:
@aldenml two line variant looks compact (and have less cpu commands), but as already mentioned by @arvidn raised exception may leave object in undefined state.

ajax16384:
current old variant has same sideeffect, so let's try two line variant

aldenml:
I hope it has the desired effect now. If that's the case, what do you think of reverting the `using ...`s, they are not really necessary.

ajax16384:
np

PR #1585, opened 2017-01-19
make sure torrent_info::is_valid returns false when the file fails to load

arvidn:


arvidn:
hm.. already pulled..

PR #1586, opened 2017-01-20
added support for infohashes sample in dht storage

aldenml:


src/kademlia/dht_storage.cpp
```diff
@@ -532,6 +556,36 @@ namespace
...
+			time_point const now = aux::time_now();
+			int const interval = clamp(m_settings.sample_infohashes_interval, 0, 21600);
+
+			std::size_t const max_count
+				= std::size_t(clamp(m_settings.max_infohashes_sample_count, 0, 20));
+			std::size_t const count = std::min(max_count, m_map.size());
+
+			if (interval > 0
+				&& m_infohashes_sample.created + seconds(interval) > now
+				&& m_infohashes_sample.samples.size() >= max_count)
```
aldenml: this line should be `m_infohashes_sample.samples.size() >= max_count * 20`

src/kademlia/dht_storage.cpp
```diff
@@ -532,6 +556,36 @@ namespace
...
+				return;
+
+			std::vector<node_id> keys;
+			for (auto const& t : m_map)
+				keys.push_back(t.first);
```
ssiloti: Rather than copy all in info hashes into a vector and shuffle that, it would be more efficient to pick hashes with the appropriate probability like get_peers does.

aldenml: thanks for the review, I will definitely evaluate that

src/kademlia/dht_storage.cpp
```diff
@@ -453,6 +464,15 @@ namespace
 			touch_item(i->second, addr);
 		}
 
+		int get_infohashes_sample(entry& item) const override
```
arvidn: I don't really like ``entry`` all that much. Do you think it would be inconvenient to return these values in a more "decoded" form, and defer decoding to the caller?

src/kademlia/dht_storage.cpp
```diff
@@ -453,6 +464,15 @@ namespace
...
+		{
+			item["interval"] = clamp(m_settings.sample_infohashes_interval, 0, 21600);
```
arvidn: this literal integer, I take it, is defined by the bep? what do you think about giving it a name (as a constant) and use that instead?

src/kademlia/dht_storage.cpp
```diff
@@ -532,6 +556,36 @@ namespace
 			if (!peers.empty() && peers.capacity() / peers.size() >= 4u)
 				peers.shrink_to_fit();
 		}
+
+		void refresh_infohashes_sample()
```
arvidn: I get the impression that the code in here would be simpler if ``infohashes_sample.sample`` was ``std::vector<sha1_hash>`` instead of a string.

src/kademlia/dht_storage.cpp
aldenml: well, part of this is for historical reasons, but right now it's very convenient due to the way the storage is used in other parts, even performance wise. I would say that a refactor is warranted at later point.

src/kademlia/dht_storage.cpp
aldenml: sure

src/kademlia/dht_storage.cpp
aldenml: I need `infohashes_sample.sample` to be a string due to performance reasons. This way that data is added as is to `entry["samples"]` without the need of conversion to the "blob" string each time. I could refactor a little though.

arvidn: isn't there a way to assign a ``span`` to an entry? Either way, I would consider this primarily a problem in ``entry``. It seems like it would be cleaner for this class to not worry _too_ much about how it's encoded. But the bytes in memory *would* be the same as with the string, right?

src/kademlia/dht_storage.cpp
arvidn: is ``entry`` used in other parts of this interface?

aldenml: if you are talking about the pattern of passing an entry reference, yes, all over the interface

src/kademlia/dht_storage.cpp
aldenml: sorry, but I'm not sure I follow you correctly. Could you rephrase the second question?

arvidn: I was just thinking of ways to use a ``std::vector<sha1_hash>`` instead of a ``std::string`` when putting it into an ``entry`` dictionary. ``entry`` has an ``operator=(span<char onst>)``, so if only there was a way to turn an arbitrary ``std::vector<>`` into a span of bytes, you could still assign it as efficiently.

perhaps something like: ``template <typename T> span<char const> to_span(std::vector<T>& v) { return {reinterpret_cast<char const*>(v.data()), v.size() * sizeof(T)}; }``. or something like that

aldenml: thanks, now I understand, well, the problem is the exact nature of `T`, in this particular case is a pointer, but assuming `T` has `data()` and `size()` I think is doable, I want to give it a try in another PR to use it here

aldenml: actually `sha1_hash` is in the stack

aldenml:
I made the `span` trick here because I'm not sure of the "general" use of `to_span`, let me know

aldenml:
ping here, in case you have a chance

src/kademlia/dht_storage.cpp
```diff
@@ -171,6 +171,22 @@ namespace
...
+	{
+		std::vector<sha1_hash> samples;
+		time_point created = min_time();
+
+		int count() const { return int(samples.size()); }
+	};
+
+	int clamp(int v, int lo, int hi)
+	{
+		return (v < lo) ? lo : (hi < v) ? hi : v;
```
arvidn: looking at this makes me envision someone getting ``lo`` and ``hi`` mixed up, with very surprising results. should there be an assert ``lo <= hi`` maybe?

src/kademlia/dht_storage.cpp
```diff
@@ -453,6 +469,19 @@ namespace
 			touch_item(i->second, addr);
 		}
 
+		int get_infohashes_sample(entry& item) const override
+		{
+			item["interval"] = clamp(m_settings.sample_infohashes_interval
+				, 0, sample_infohashes_interval_max);
+			item["num"] = int(m_map.size());
+
+			std::vector<sha1_hash> const& samples = m_infohashes_sample.samples;
```
arvidn: I think it would be slightly nicer if the samples were computed lazily. i.e. right before this line, check if the samples have expired or not. if they have, call ``refresh_infohashes_sample()``.

Or call that function unconditionally here, and it checks to see if the sample is still "fresh".

arvidn: that way, the test wouldn't need this (somewhat awkward) call to ``tick()``

test/test_dht_storage.cpp
```diff
@@ -437,4 +437,83 @@ TORRENT_TEST(update_node_ids)
...
+
+	entry item;
+	int r = s->get_infohashes_sample(item);
+	TEST_EQUAL(r, 2);
+	TEST_EQUAL(item["interval"].integer(), 10)
+	TEST_EQUAL(item["num"].integer(), 4);
+	TEST_EQUAL(item["samples"].string().size(), 2 * 20);
+
+	// get all of them
+	sett.sample_infohashes_interval = 0;
```
arvidn: do you set this to zero to force the sample to be generated immediately?

test/test_dht_storage.cpp
```diff
@@ -437,4 +437,83 @@ TORRENT_TEST(update_node_ids)
...
+	sett.max_infohashes_sample_count = 5;
+
+	s->tick();
+
+	item = entry();
+	r = s->get_infohashes_sample(item);
+	TEST_EQUAL(r, 4);
+	TEST_EQUAL(item["interval"].integer(), 0)
```
arvidn: it wasn't obvious to me why this would be zero at first, but I guess you have to do that to make ``tick`` generate the sample

src/kademlia/dht_storage.cpp
aldenml: sure, why not :)

src/kademlia/dht_storage.cpp
aldenml: I think it is a good idea, I will rework for that

test/test_dht_storage.cpp
aldenml: yes, but that should go away with your idea

PR #1589, opened 2017-01-20
fix torrent name in alerts of builds with deprecated functions

Falcosc:
and use non deprecated for internal calls

This did not work in constructor:
`name = torrent_name();`


arvidn:
lgtm

Falcosc:
Don't ask about this, I did spend 5 hours into the investigation. This bug made me crazy.

And there is still a string buffer mistake somewere, it does cut the front of the lines sometimes.

PR #1590, opened 2017-01-20
remove dependency on aligned_storage and aligned_union

arvidn:
to support older versions of GCC


aldenml:
the ubuntu toolchain PPA supports gcc 5 for very old dist versions, don't you think it's better to simply drop support for gcc <= 5.1? gcc 4 is no longer maintained

arvidn:
I don't know what PPA means.. just that sometimes systems have old toolchains. cas-- points out that gcc 4.8 is supported in Ubuntu Trusty until 2019. I take it that's the LTS version perhaps.

aldenml:
sorry I assumed some familiarity...no need of complicating this if you already made it

arvidn:
it seems msvc makes this a bit trickier

src/cpuid.cpp
```diff
@@ -46,7 +46,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <cstring> // for std::memset
 #endif
 
-#if TORRENT_HAS_ARM
+#if TORRENT_HAS_ARM && TORRENT_HAS_ARM_NEON
```
arvidn: @aldenml does this seem reasonable? as far as I can tell, this function is only used if ``TORRENT_HAS_ARM_NEON``, and in my cross compiler for ``arm-linux-gnueabihf`` does not have this header

aldenml: It is used for `TORRENT_HAS_ARM_CRC32` too, I think you can add it to this condition and solve the issue. What toolchain/sysroot are you using?

arvidn: linaro for raspberry pi. arm-linux-gnueabihf-raspbian
Additionally, this is a linux header, so presumably it should only be used on ``__linux__`` as well, right?

aldenml: true.

btw, linaro is the same you have in travis and raspberry pi is a linux based system :)

arvidn: this header wasn't introduced until glibc 2.16 as far as I can tell, so I added that condition. It doesn't seem linux specific, just glibc specific

src/cpuid.cpp
```diff
@@ -46,7 +46,14 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <cstring> // for std::memset
 #endif
 
-#if TORRENT_HAS_ARM && defined __linux__ && (TORRENT_HAS_ARM_NEON || TORRENT_HAS_ARM_CRC32)
+#if defined __GLIBC__ && (__GLIBC__ > 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 16))
```
aldenml: do you mind adding here `|| TORRENT_ANDROID`?

my understanding is that it was added in `glibc` but it's _not_ `glibc` specific, in fact, that's the documented way in android to get the CPU features.

arvidn: isn't android using glibc though?

aldenml: no, it's `bionic`

PR #1593, opened 2017-01-21
fix truncation of peer_class field in torrent

arvidn:


arvidn:
@Falcosc I believe you missed this one in your patch

Falcosc:
Thanks, how did you find this?

arvidn:
in master we have value conversion warnings enabled

PR #1596, opened 2017-01-21
Typo in comment

Col-blimp:


arvidn:
would you mind just remove the mention of any default value instead? I'm pretty sure the actual default is included in the generated documentation

Col-blimp:
Shall all references to defaults be removed from that file?

arvidn:
yes, they should

arvidn:
as you can see [here](http://libtorrent.org/reference-Settings.html#settings_pack), the defaults are extracted from the source.

PR #1600, opened 2017-01-22
packet allocation pool

ajax16384:
- no more direct malloc/free with cast
- reuse allocated memory for 3 common allocation case (with limit)

### DIFF
--- a/include/libtorrent/packet_pool.hpp
+++ b/include/libtorrent/packet_pool.hpp
@@ -79,7 +79,7 @@ namespace libtorrent
 	};
 
 
-	struct packet_mem_manager
+	struct packet_deleter
 	{
 		// deleter for std::unique_ptr
 		void operator()(packet* p) const
@@ -87,16 +87,16 @@ namespace libtorrent
 			p->~packet();
 			std::free(p);
 		}
-
-		static packet *create_packet(int size)
-		{
-			packet *p = static_cast<packet*>(std::malloc(sizeof(packet) + size));
-			new (p) packet();
-			return p;
-		}
 	};
 
-	using packet_ptr = std::unique_ptr<packet, packet_mem_manager>;
+	static packet *create_packet(int size)
+	{
+		packet *p = static_cast<packet*>(std::malloc(sizeof(packet) + size));
+		new (p) packet();
+		return p;
+	}
+
+	using packet_ptr = std::unique_ptr<packet, packet_deleter>;
 
 	template<int ALLOCATE_SIZE>
 	struct TORRENT_EXTRA_EXPORT packet_slab
@@ -116,7 +116,7 @@ namespace libtorrent
 		packet_ptr alloc()
 		{
 			if (m_storage.empty())
-				return packet_ptr(packet_mem_manager::create_packet(allocate_size));
+				return packet_ptr(create_packet(allocate_size));
 			else
 			{
 				auto ret = std::move(m_storage.back());
@@ -176,11 +176,11 @@ namespace libtorrent
 			if (allocate <= m_syn_slab.allocate_size) { return m_syn_slab.alloc(); }
 			else if (allocate <= m_mtu_floor_slab.allocate_size) { return m_mtu_floor_slab.alloc(); }
 			else if (allocate <= m_mtu_ceiling_slab.allocate_size) { return m_mtu_ceiling_slab.alloc(); }
-			return packet_ptr(packet_mem_manager::create_packet(allocate));
+			return packet_ptr(create_packet(allocate));
 		}
-		packet_slab<sizeof(utp_header)> m_syn_slab{ 100 };
-		packet_slab<(TORRENT_INET_MIN_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_floor_slab{ 512 };
-		packet_slab<(TORRENT_ETHERNET_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_ceiling_slab{ 256 };
+		packet_slab<sizeof(utp_header)> m_syn_slab{ 50 };
+		packet_slab<(TORRENT_INET_MIN_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_floor_slab{ 100 };
+		packet_slab<(TORRENT_ETHERNET_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_ceiling_slab{ 50 };
 	};
 }
 
--- a/include/libtorrent/packet_pool.hpp
+++ b/include/libtorrent/packet_pool.hpp
@@ -83,55 +83,55 @@ namespace libtorrent
 	{
 		void operator()(packet* handle) const
 		{
+			handle->~packet();
 			std::free(handle);
 		}
 	};
 
 	using packet_ptr = std::unique_ptr<packet, packet_deleter>;
 
-	template<std::size_t ALLOCATE_SIZE>
+	template<int ALLOCATE_SIZE>
 	struct TORRENT_EXTRA_EXPORT packet_slab : private aux::vector<packet_ptr, std::size_t>
 	{
-		static const std::size_t allocate_size{ ALLOCATE_SIZE };
+		static const int allocate_size{ ALLOCATE_SIZE };
 
 		packet_slab(std::size_t limit) : m_limit(limit) { reserve(m_limit); }
 
-		void try_free(packet_ptr &p)
+		void try_push_back(packet_ptr &p)
 		{
 			if (size() < m_limit)
 				push_back(std::move(p));
 		}
 
-		void try_alloc(packet_ptr &p)
+		packet_ptr alloc()
 		{
 			if (empty())
-				p.reset(static_cast<packet*>(std::malloc(sizeof(packet) + allocate_size)));
+				return packet_ptr(static_cast<packet*>(std::malloc(sizeof(packet) + allocate_size)));
 			else
 			{
-				p = std::move(back());
+				auto ret = std::move(back());
 				pop_back();
+				return ret;
 			}
 		}
+
+		void flush() { resize(0); }
 	private:
-		std::size_t m_limit;
+		const std::size_t m_limit;
 	};
 
 	// single thread packet allocation packet pool
 	// can handle common cases of packet size by 3 pools
 	struct TORRENT_EXTRA_EXPORT packet_pool : private single_threaded
 	{
-		packet *acquire(std::size_t allocate)
+		packet *acquire(int allocate)
 		{
 			TORRENT_ASSERT(is_single_thread());
+			TORRENT_ASSERT(allocate >= 0);
 			TORRENT_ASSERT(allocate <= std::numeric_limits<std::uint16_t>::max());
 
-			packet_ptr p;
-
-			if (allocate <= m_syn_slab.allocate_size) { m_syn_slab.try_alloc(p); }
-			else if (allocate <= m_mtu_floor_slab.allocate_size) { m_mtu_floor_slab.try_alloc(p); }
-			else if (allocate <= m_mtu_ceiling_slab.allocate_size) { m_mtu_ceiling_slab.try_alloc(p); }
-			else { m_rest_slab.try_alloc(p); }
-
+			packet_ptr p{ alloc(allocate) };
+			new (p.get()) packet();
 			p->allocated = static_cast<std::uint16_t>(allocate);
 			return p.release();
 		}
@@ -142,18 +142,35 @@ namespace libtorrent
 
 			if (p == nullptr) return;
 
-			packet_ptr pp(p); // take ownership and auto free
+			packet_ptr pp(p); // takes ownership and may auto free if slab container does not move it
 			std::size_t allocated = pp->allocated;
 
-			if (allocated <= m_syn_slab.allocate_size) { m_syn_slab.try_free(pp); }
-			else if (allocated <= m_mtu_floor_slab.allocate_size) { m_mtu_floor_slab.try_free(pp); }
-			else if (allocated <= m_mtu_ceiling_slab.allocate_size) { m_mtu_ceiling_slab.try_free(pp); }
+			if (allocated <= m_syn_slab.allocate_size) { m_syn_slab.try_push_back(pp); }
+			else if (allocated <= m_mtu_floor_slab.allocate_size) { m_mtu_floor_slab.try_push_back(pp); }
+			else if (allocated <= m_mtu_ceiling_slab.allocate_size) { m_mtu_ceiling_slab.try_push_back(pp); }
 		}
+
+		//TODO: call from utp manager to flush mem
+		void flush()
+		{
+			TORRENT_ASSERT(is_single_thread());
+
+			m_syn_slab.flush();
+			m_mtu_floor_slab.flush();
+			m_mtu_ceiling_slab.flush();
+		}
+
 	private:
-		packet_slab<sizeof(utp_header)> m_syn_slab{ 0x400 };
+		packet_ptr alloc(int allocate)
+		{
+			if (allocate <= m_syn_slab.allocate_size) { return m_syn_slab.alloc(); }
+			else if (allocate <= m_mtu_floor_slab.allocate_size) { return m_mtu_floor_slab.alloc(); }
+			else if (allocate <= m_mtu_ceiling_slab.allocate_size) { return m_mtu_ceiling_slab.alloc(); }
+			return packet_ptr(static_cast<packet*>(std::malloc(sizeof(packet) + allocate)));
+		}
+		packet_slab<sizeof(utp_header)> m_syn_slab{ 0x100 };
 		packet_slab<(TORRENT_INET_MIN_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_floor_slab{ 0x200 };
-		packet_slab<(TORRENT_ETHERNET_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_ceiling_slab{ 0x200 };
-		packet_slab<0> m_rest_slab{ 0 };
+		packet_slab<(TORRENT_ETHERNET_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_ceiling_slab{ 0x100 };
 	};
 }
 
--- a/include/libtorrent/packet_pool.hpp
+++ b/include/libtorrent/packet_pool.hpp
@@ -79,16 +79,24 @@ namespace libtorrent
 	};
 
 
-	struct packet_deleter
+	struct packet_mem_manager
 	{
-		void operator()(packet* handle) const
+		// deleter for std::unique_ptr
+		void operator()(packet* p) const
 		{
-			handle->~packet();
-			std::free(handle);
+			p->~packet();
+			std::free(p);
+		}
+
+		static packet *create_packet(int size)
+		{
+			packet *p = static_cast<packet*>(std::malloc(sizeof(packet) + size));
+			new (p) packet();
+			return p;
 		}
 	};
 
-	using packet_ptr = std::unique_ptr<packet, packet_deleter>;
+	using packet_ptr = std::unique_ptr<packet, packet_mem_manager>;
 
 	template<int ALLOCATE_SIZE>
 	struct TORRENT_EXTRA_EXPORT packet_slab
@@ -108,7 +116,7 @@ namespace libtorrent
 		packet_ptr alloc()
 		{
 			if (m_storage.empty())
-				return packet_ptr(static_cast<packet*>(std::malloc(sizeof(packet) + allocate_size)));
+				return packet_ptr(packet_mem_manager::create_packet(allocate_size));
 			else
 			{
 				auto ret = std::move(m_storage.back());
@@ -117,7 +125,7 @@ namespace libtorrent
 			}
 		}
 
-		void flush() { m_storage.resize(0); }
+		void flush() { m_storage.clear(); }
 	private:
 		const std::size_t m_limit;
 		aux::vector<packet_ptr, std::size_t> m_storage;
@@ -134,7 +142,6 @@ namespace libtorrent
 			TORRENT_ASSERT(allocate <= std::numeric_limits<std::uint16_t>::max());
 
 			packet_ptr p{ alloc(allocate) };
-			new (p.get()) packet();
 			p->allocated = static_cast<std::uint16_t>(allocate);
 			return p.release();
 		}
@@ -169,11 +176,11 @@ namespace libtorrent
 			if (allocate <= m_syn_slab.allocate_size) { return m_syn_slab.alloc(); }
 			else if (allocate <= m_mtu_floor_slab.allocate_size) { return m_mtu_floor_slab.alloc(); }
 			else if (allocate <= m_mtu_ceiling_slab.allocate_size) { return m_mtu_ceiling_slab.alloc(); }
-			return packet_ptr(static_cast<packet*>(std::malloc(sizeof(packet) + allocate)));
+			return packet_ptr(packet_mem_manager::create_packet(allocate));
 		}
-		packet_slab<sizeof(utp_header)> m_syn_slab{ 0x100 };
-		packet_slab<(TORRENT_INET_MIN_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_floor_slab{ 0x200 };
-		packet_slab<(TORRENT_ETHERNET_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_ceiling_slab{ 0x100 };
+		packet_slab<sizeof(utp_header)> m_syn_slab{ 100 };
+		packet_slab<(TORRENT_INET_MIN_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_floor_slab{ 512 };
+		packet_slab<(TORRENT_ETHERNET_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_ceiling_slab{ 256 };
 	};
 }
 
--- a/include/libtorrent/packet_pool.hpp
+++ b/include/libtorrent/packet_pool.hpp
@@ -78,24 +78,62 @@ namespace libtorrent
 		std::uint8_t buf[1];
 	};
 
+
+	struct packet_deleter
+	{
+		void operator()(packet* handle) const
+		{
+			std::free(handle);
+		}
+	};
+
+	using packet_ptr = std::unique_ptr<packet, packet_deleter>;
+
+	template<std::size_t ALLOCATE_SIZE>
+	struct TORRENT_EXTRA_EXPORT packet_slab : private aux::vector<packet_ptr, std::size_t>
+	{
+		static const std::size_t allocate_size{ ALLOCATE_SIZE };
+
+		packet_slab(std::size_t limit) : m_limit(limit) { reserve(m_limit); }
+
+		void try_free(packet_ptr &p)
+		{
+			if (size() < m_limit)
+				push_back(std::move(p));
+		}
+
+		void try_alloc(packet_ptr &p)
+		{
+			if (empty())
+				p.reset(static_cast<packet*>(std::malloc(sizeof(packet) + allocate_size)));
+			else
+			{
+				p = std::move(back());
+				pop_back();
+			}
+		}
+	private:
+		std::size_t m_limit;
+	};
+
 	// single thread packet allocation packet pool
 	// can handle common cases of packet size by 3 pools
-	struct TORRENT_EXTRA_EXPORT packet_pool : single_threaded
+	struct TORRENT_EXTRA_EXPORT packet_pool : private single_threaded
 	{
 		packet *acquire(std::size_t allocate)
 		{
 			TORRENT_ASSERT(is_single_thread());
 			TORRENT_ASSERT(allocate <= std::numeric_limits<std::uint16_t>::max());
 
-			packet *p;
+			packet_ptr p;
 
-			if (allocate <= m_syn_slab.allocate_size) { p = m_syn_slab.try_alloc(); }
-			else if (allocate <= m_mtu_floor_slab.allocate_size) { p = m_mtu_floor_slab.try_alloc(); }
-			else if (allocate <= m_mtu_ceiling_slab.allocate_size) { p = m_mtu_ceiling_slab.try_alloc(); }
-			else { p = static_cast<packet*>(malloc(sizeof(packet) + allocate)); }
+			if (allocate <= m_syn_slab.allocate_size) { m_syn_slab.try_alloc(p); }
+			else if (allocate <= m_mtu_floor_slab.allocate_size) { m_mtu_floor_slab.try_alloc(p); }
+			else if (allocate <= m_mtu_ceiling_slab.allocate_size) { m_mtu_ceiling_slab.try_alloc(p); }
+			else { m_rest_slab.try_alloc(p); }
 
 			p->allocated = static_cast<std::uint16_t>(allocate);
-			return p;
+			return p.release();
 		}
 
 		void release(packet *p)
@@ -103,53 +141,19 @@ namespace libtorrent
 			TORRENT_ASSERT(is_single_thread());
 
 			if (p == nullptr) return;
-			std::size_t allocated = p->allocated;
 
-			if (allocated <= m_syn_slab.allocate_size) { m_syn_slab.try_free(p); }
-			else if (allocated <= m_mtu_floor_slab.allocate_size) { m_mtu_floor_slab.try_free(p); }
-			else if (allocated <= m_mtu_ceiling_slab.allocate_size) { m_mtu_ceiling_slab.try_free(p); }
-			else { free(p); }
+			packet_ptr pp(p); // take ownership and auto free
+			std::size_t allocated = pp->allocated;
+
+			if (allocated <= m_syn_slab.allocate_size) { m_syn_slab.try_free(pp); }
+			else if (allocated <= m_mtu_floor_slab.allocate_size) { m_mtu_floor_slab.try_free(pp); }
+			else if (allocated <= m_mtu_ceiling_slab.allocate_size) { m_mtu_ceiling_slab.try_free(pp); }
 		}
 	private:
-
-		template<std::size_t PACKET_LIMIT, std::size_t ALLOCATE_SIZE>
-		struct packet_slab
-		{
-			aux::vector<packet*, std::size_t> m_storage;
-			std::size_t idx{ 0 };
-			static const std::size_t allocate_size{ ALLOCATE_SIZE };
-
-			packet_slab() : m_storage(PACKET_LIMIT, nullptr) { }
-
-			~packet_slab() { flush(); }
-
-			void flush()
-			{
-				while (idx != 0) {
-					free(m_storage[--idx]);
-				}
-			}
-
-			void try_free(packet *p)
-			{
-				if (idx >= PACKET_LIMIT)
-					free(p);
-				else
-					m_storage[idx++] = p;
-			}
-
-			packet *try_alloc()
-			{
-				if (idx == 0)
-					return static_cast<packet*>(malloc(sizeof(packet) + allocate_size));
-				else
-					return m_storage[--idx];
-			}
-		};
-
-		packet_slab<0x400, sizeof(utp_header)> m_syn_slab;
-		packet_slab<0x200, (TORRENT_INET_MIN_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_floor_slab;
-		packet_slab<0x200, (TORRENT_ETHERNET_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_ceiling_slab;
+		packet_slab<sizeof(utp_header)> m_syn_slab{ 0x400 };
+		packet_slab<(TORRENT_INET_MIN_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_floor_slab{ 0x200 };
+		packet_slab<(TORRENT_ETHERNET_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_ceiling_slab{ 0x200 };
+		packet_slab<0> m_rest_slab{ 0 };
 	};
 }
 
--- a/include/libtorrent/packet_pool.hpp
+++ b/include/libtorrent/packet_pool.hpp
@@ -89,9 +89,9 @@ namespace libtorrent
 		}
 	};
 
-	inline packet *create_packet(int size)
+	inline packet* create_packet(int size)
 	{
-		packet *p = static_cast<packet*>(std::malloc(sizeof(packet) + size));
+		packet* p = static_cast<packet*>(std::malloc(sizeof(packet) + size));
 		new (p) packet();
 		return p;
 	}
@@ -135,7 +135,7 @@ namespace libtorrent
 	// can handle common cases of packet size by 3 pools
 	struct TORRENT_EXTRA_EXPORT packet_pool : private single_threaded
 	{
-		packet *acquire(int allocate)
+		packet* acquire(int allocate)
 		{
 			TORRENT_ASSERT(is_single_thread());
 			TORRENT_ASSERT(allocate >= 0);
@@ -146,7 +146,7 @@ namespace libtorrent
 			return p.release();
 		}
 
-		void release(packet *p)
+		void release(packet* p)
 		{
 			TORRENT_ASSERT(is_single_thread());
 
--- a/include/libtorrent/packet_pool.hpp
+++ b/include/libtorrent/packet_pool.hpp
@@ -91,33 +91,36 @@ namespace libtorrent
 	using packet_ptr = std::unique_ptr<packet, packet_deleter>;
 
 	template<int ALLOCATE_SIZE>
-	struct TORRENT_EXTRA_EXPORT packet_slab : private aux::vector<packet_ptr, std::size_t>
+	struct TORRENT_EXTRA_EXPORT packet_slab
 	{
 		static const int allocate_size{ ALLOCATE_SIZE };
 
-		packet_slab(std::size_t limit) : m_limit(limit) { reserve(m_limit); }
+		explicit packet_slab(std::size_t limit) : m_limit(limit) { m_storage.reserve(m_limit); }
+
+		packet_slab(const packet_slab&) = delete;
 
 		void try_push_back(packet_ptr &p)
 		{
-			if (size() < m_limit)
-				push_back(std::move(p));
+			if (m_storage.size() < m_limit)
+				m_storage.push_back(std::move(p));
 		}
 
 		packet_ptr alloc()
 		{
-			if (empty())
+			if (m_storage.empty())
 				return packet_ptr(static_cast<packet*>(std::malloc(sizeof(packet) + allocate_size)));
 			else
 			{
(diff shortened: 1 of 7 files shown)


src/utp_socket_manager.cpp
```diff
@@ -63,6 +63,7 @@ namespace libtorrent
 		, m_ssl_context(ssl_context)
 	{
 		m_restrict_mtu.fill(65536);
+		m_packet_pool = construct_packet_pool();
```
arvidn: please don't have free functions for this. This should be the constructor

src/utp_socket_manager.cpp
```diff
@@ -72,6 +73,7 @@ namespace libtorrent
 		{
 			delete_utp_impl(i->second);
 		}
+		delete_packet_pool(m_packet_pool);
```
arvidn: please use the destructor instead of a free function like this

include/libtorrent/utp_socket_manager.hpp
```diff
@@ -188,6 +188,8 @@ namespace libtorrent
 		// this is  passed on to the instantiate connection
 		// if this is non-nullptr it will create SSL connections over uTP
 		void* m_ssl_context;
+
+		packet_pool *m_packet_pool;
```
arvidn: since this is constructed unconditionally. I should just be part of the class, not a pointer

arvidn: and if it really needs to be heap allocated (which I haven't seen any evidence of yet), please make it a ``std::unique_ptr``

src/utp_stream.cpp
```diff
@@ -210,6 +210,152 @@ struct packet
...
+			{
+				packet_stash_item *pi = reinterpret_cast<packet_stash_item *>(p);
+				top = pi->next;
+			}
+			return p;
+		}
+	};
+
+#if TORRENT_USE_ASSERTS
+	void check_thread()
```
arvidn: there is a class that does this already, calld ``single_threaded``. to use it you derive from it and call ``TORRENT_ASSERT(is_single_thread());``

src/utp_stream.cpp
```diff
@@ -210,6 +210,152 @@ struct packet
...
+		else
+		{
+			TORRENT_ASSERT(current_thread_id == check_thread_id);
+		}
+	}
+
+	std::thread::id check_thread_id;
+#endif
+
+	const std::uint16_t utp_header_size = sizeof(utp_header);
```
arvidn: these should probably be ``static`` (there's no need to have member per instance, right?). But perhaps even better, if they are just referenced in the .cpp file, they should be defined in there

src/utp_stream.cpp
```diff
@@ -210,6 +210,152 @@ struct packet
 	std::uint8_t buf[1];
 };
 
+// single thread packet allocation packet pool
+// can handle 3 common cases of packet allocation size
+struct packet_pool
```
arvidn: this class needs a unit test, which I believe requires it to be defined in a header file. ideally it's own header file

src/utp_stream.cpp
```diff
@@ -210,6 +210,152 @@ struct packet
...
+		void flush()
+		{
+			packet *p = top;
+			while (p != nullptr)
+			{
+				packet_stash_item *pi = reinterpret_cast<packet_stash_item *>(p);
+				packet* next = pi->next;
+				free(p);
+				p = next;
+			}
```
arvidn: I would have expected this function to set ``top`` to ``nullptr`` here. is this correct?

src/utp_stream.cpp
```diff
@@ -210,6 +210,152 @@ struct packet
...
+		m_mtu_ceiling_stash.flush();
+	}
+private:
+	struct packet_stash_item
+	{
+		packet *next;
+		std::size_t index;
+	};
+
+	struct packet_stash
```
arvidn: I think this type would normally be referred to as a "slab". But it also looks like this basically does exactly what ``boost::pool`` does. boost.pool actually does the allocations in blocks as well, so it may provide a slightly more compact pool

src/utp_stream.cpp
```diff
@@ -210,6 +210,152 @@ struct packet
...
+		{
+			TORRENT_ASSERT(current_thread_id == check_thread_id);
+		}
+	}
+
+	std::thread::id check_thread_id;
+#endif
+
+	const std::uint16_t utp_header_size = sizeof(utp_header);
+	const std::size_t utp_header_limit = 0x4000;
```
arvidn: at first read, this may be interpreted as a *size* limit, which looks wrong (perhaps there is a clearer name). Ideally it would be configurable. That way the test would be a bit quicker too

src/utp_stream.cpp
```diff
@@ -261,8 +407,9 @@ struct packet
 struct utp_socket_impl
 {
 	utp_socket_impl(std::uint16_t recv_id, std::uint16_t send_id
-		, void* userdata, utp_socket_manager* sm)
+		, void* userdata, utp_socket_manager* sm, packet_pool* pp)
 		: m_sm(sm)
+		, m_packet_pool(pp)
```
arvidn: what would you think about making the packet allocation and freeing functions available via the ``utp_socket_manager`` instead? that way each individual socket would not need to carry the extra pointer

src/utp_stream.cpp
```diff
@@ -341,6 +488,9 @@ struct utp_socket_impl
 
 	void set_state(int s);
 
+	packet *acquire_packet(std::uint16_t allocate) { return m_packet_pool->acquire(allocate); }
```
arvidn: in the future, in a different patch, it would be nice to wrap this pointer in a holder with a private constructor that only the packet_pool can invoke, and make it move-only. That way the only way to make it compile is to return the packet to the pool.

arvidn: actually, never mind, I don't think that would work.

However, this function should not take a ``uint16_t``. It should either take a ``std::size_t`` or ``int``. that way the callers don't have to cast

src/utp_stream.cpp
```diff
@@ -210,6 +210,152 @@ struct packet
...
+		return p;
+	}
+
+	void release(packet *p)
+	{
+#if TORRENT_USE_ASSERTS
+		check_thread();
+#endif
+		if (p == nullptr) return;
+		if (p->allocated <= utp_header_size)
```
arvidn: I'm pretty sure the ``<=`` here is a bug. if you return a buffer that's *smaller*, the allocation function may hand you a buffer that's smaller than you asked for. Surely these comparisons must be ``==``

src/utp_stream.cpp
```diff
@@ -210,6 +210,152 @@ struct packet
...
+private:
+	struct packet_stash_item
+	{
+		packet *next;
+		std::size_t index;
+	};
+
+	struct packet_stash
+	{
+		packet *top{ nullptr };
```
arvidn: are you confident a linked list is the most efficient data structure? compared to, say, a vector of pointers?

src/utp_stream.cpp
```diff
@@ -210,6 +210,152 @@ struct packet
...
+		else if (p->allocated <= mtu_floor_size)
+		{
+			if (m_mtu_floor_stash.push(p, mtu_floor_limit)) return;
+		}
+		else if (p->allocated <= mtu_ceiling_size)
+		{
+			if (m_mtu_ceiling_stash.push(p, mtu_ceiling_limit)) return;
+		}
+
+		free(p);
```
arvidn: speaking of tests. this line is not executed by any test right now (i.e. there needs to be a unit test that deliberately makes sure this class operates correctly, not that the whole program appears to behave correctly)

arvidn:
this is a good feature. I believe it can be simplified (by reusing the ``single_threaded`` class, and perhaps by using boost.pool). It definitely needs thorough test coverage though

src/utp_stream.cpp
```diff
@@ -210,6 +210,152 @@ struct packet
...
+			should_allocate = mtu_floor_size;
+		}
+		else if (allocate <= mtu_ceiling_size)
+		{
+			p = m_mtu_ceiling_stash.pop();
+			should_allocate = mtu_ceiling_size;
+		}
+
+		if (p == nullptr)
+			p = static_cast<packet*>(malloc(sizeof(packet) + should_allocate));
```
arvidn: it's not an obvious win, in my mind, to make the ``sizeof(packet)`` implied in the request for memory. From the users point of view, it may look like a bug to request less memory than what is used

src/utp_stream.cpp
```diff
@@ -210,6 +210,152 @@ struct packet
...
+			}
+		}
+
+		bool push(packet *p, std::size_t limit)
+		{
+			packet_stash_item *tpi = reinterpret_cast<packet_stash_item *>(top);
+			std::size_t top_index = (tpi == nullptr) ? 0 : tpi->index;
+			if (limit < top_index) return false;
+
+			packet_stash_item *pi = reinterpret_cast<packet_stash_item *>(p);
```
ssiloti: This is undefined behavior, you cannot dereference a pointer after casting it from an unrelated type.

ssiloti:
I am _really_ skeptical that this is worth implementing a custom pool allocator for. Even using boost.pool is iffy IMO. If heap management overhead is an issue you are much better off just dropping in a highly optimized malloc implementation like jemalloc.

arvidn:
it's not just how efficient the allocator is, but also having the exact right slab sizes, not have no waste. (I realize this patch doesn't do that though). In addition, the win32 allocator (at least historically) is very inefficient, so it seems pragmatic to use pool allocators

src/utp_socket_manager.cpp
ajax16384: ok

src/utp_socket_manager.cpp
ajax16384: ok

include/libtorrent/utp_socket_manager.hpp
ajax16384: ok

src/utp_stream.cpp
ajax16384: ok

src/utp_stream.cpp
ajax16384: ok

src/utp_stream.cpp
ajax16384: it will automatically set to nullptr by last packet which next field points to nullptr

src/utp_stream.cpp
ajax16384: ok. i'll try to reuse boost::pool

src/utp_stream.cpp
ajax16384: ok

src/utp_stream.cpp
ajax16384: boost::pool should cover this issue

ajax16384:
@ssiloti jemalloc is good choice for whole program mem management optimization (like nosql "redis" can switch to alt malloc) and libTorrent should have such build variant too (may be in next pull request). But in case of packet management it obvious that same size packets should only be rotated without deallocation.  i'll try to use well tested boost::pool instead of custom allocator (not to reinvent same thing).

src/utp_stream.cpp
ajax16384: ok

src/utp_stream.cpp
ajax16384: "p" and "pi" are pointers to different structs (reinterpret_cast "Allows any pointer to be converted into any other pointer type. "  https://msdn.microsoft.com/en-us/library/e0w9f63b.aspx ).

ssiloti: Quote from that page: "The result of a reinterpret_cast cannot safely be used for anything other than being cast back to its original type. Other uses are, at best, nonportable."

ajax16384: use can easily check all possible cast options in this place:
````c++
packet_stash_item *pi = reinterpret_cast<packet_stash_item *>(p);
packet_stash_item *pi1 = (packet_stash_item *)p; // same as above
packet_stash_item *pi2 = static_cast<packet_stash_item *>(p); // will not compile: note: Types pointed to are unrelated; conversion requires reinterpret_cast, C-style cast or function-style cast
````
in case you have missed the point of this cast - i need to reuse same data storage for linked item and it;s index. this was made by intention (pointer to same memory)

Comment (ajax16384) [inline src/utp_stream.cpp]: pretty sure you may find reinterpret_cast all over libTorrent for same purpose - use pointer for one data struct for access it as another type of data structure

Comment (ssiloti) [inline src/utp_stream.cpp]: Existing uses of reinterpret_cast are casting to the original type of the data, or to (unsigned) char*. The latter is a special case, char* is the one type permitted to alias any other data.

Comment (ajax16384) [inline src/utp_stream.cpp]: please take a look at:
````c++
heterogeneous_queue::grow_capacity(int const size) 
///
			uintptr_t* new_storage = new uintptr_t[m_capacity + amount_to_grow];

			uintptr_t* src = m_storage;
			uintptr_t* dst = new_storage;
///
				header_t* src_hdr = reinterpret_cast<header_t*>(src);
				header_t* dst_hdr = reinterpret_cast<header_t*>(dst);
````
and libtorrent::socket_type and libtorrent::entry implementations which use reinterpret_cast with completely different types to access same data locations.

i understand that it might be dangerous to "to treat the sequence of bits (object representation) of expression as if it had the type new_type", but it's what i want to be in that case of organizing linked list (reuse memory of released data to store link to next slab).

Comment (ssiloti) [inline src/utp_stream.cpp]: Yes, there are a couple of places where libtorrent invokes undefined behavior using reinterpret_cast. Some are sort-of ok, like `entry`, because they use placement new and are careful to make sure alignment requirements are met. You still shouldn't be doing it, if you want to re-use storage like this you need to use a union type.

File: include/libtorrent/packet_pool.hpp [hunk 17]
```diff
@@ -0,0 +1,158 @@
...
+			p->allocated = static_cast<std::uint16_t>(allocate);
+			return p;
+		}
+
+		void release(packet *p)
+		{
+			TORRENT_ASSERT(is_single_thread());
+
+			if (p == nullptr) return;
+			if (p->allocated <= utp_header_size)
```
arvidn: I still maintain that the ``<=`` comparison is a bug. it must be ``==``

arvidn: and I think there should be a test that exercises this bug (along with a fix obviously)

arvidn:
I have some thoughts. the risk with boost.pool is that it will become fragmented and difficult to compact/free up blocks. There is a certain elegance to the previous solution, in that it simply caches packets for reuse. I think it probably had too large a cache though, and I think a boost pool is likely to grow too big (and fragmented) as well.

The case I'm worried about is where you have a very active session and then you go idle. There should be some way to reclaim that memory. Perhaps especially if you don't go entirely idle, so some buffers are used at any given time. with boost pool, that means you have to use ``ordered_free()`` and ``ordered_malloc()``, which is slow instead.

ajax16384:
ok. i will make different tests and post results with sources.

ajax16384:
replace boost.pool with fixed size aux::vector

include/libtorrent/packet_pool.hpp
```diff
@@ -0,0 +1,156 @@
...
+	private:
+
+		template<std::size_t PACKET_LIMIT, std::size_t ALLOCATE_SIZE>
+		struct packet_slab
+		{
+			aux::vector<packet*, std::size_t> m_storage;
+			std::size_t idx{ 0 };
+			static const std::size_t allocate_size{ ALLOCATE_SIZE };
+
+			packet_slab() : m_storage(PACKET_LIMIT, nullptr) { }
```
arvidn: ``PACKET_LIMIT`` should not be a template parameter. it's encoded at run-time in the vector here, so instantiating this template for each size is redundant. it should just be a run-time parameter. I kind of think that the slab_size should be a runtime parameter too. at least I don't see a great gain in making that a template parameter

include/libtorrent/packet_pool.hpp
```diff
@@ -0,0 +1,156 @@
...
+			else if (allocated <= m_mtu_floor_slab.allocate_size) { m_mtu_floor_slab.try_free(p); }
+			else if (allocated <= m_mtu_ceiling_slab.allocate_size) { m_mtu_ceiling_slab.try_free(p); }
+			else { free(p); }
+		}
+	private:
+
+		template<std::size_t PACKET_LIMIT, std::size_t ALLOCATE_SIZE>
+		struct packet_slab
+		{
+			aux::vector<packet*, std::size_t> m_storage;
```
arvidn: this should probably be a ``std::vector<std::unique_ptr<>>`` with an appropriate deleter type to free the packet.

include/libtorrent/packet_pool.hpp
```diff
@@ -0,0 +1,156 @@
...
+			std::size_t idx{ 0 };
```
arvidn: you already have a ``std::vector`` that has the ability to track how many elements are in it. no need to track that separately

include/libtorrent/packet_pool.hpp
```diff
@@ -0,0 +1,156 @@
...
+			static const std::size_t allocate_size{ ALLOCATE_SIZE };
+
+			packet_slab() : m_storage(PACKET_LIMIT, nullptr) { }
+
+			~packet_slab() { flush(); }
+
+			void flush()
+			{
```
arvidn: this should call ``m_storage.clear()``

include/libtorrent/packet_pool.hpp
```diff
@@ -0,0 +1,156 @@
...
+				while (idx != 0) {
+					free(m_storage[--idx]);
+				}
+			}
+
+			void try_free(packet *p)
+			{
+				if (idx >= PACKET_LIMIT)
```
arvidn: I think this function should do:
```
if (m_storage.size() >= m_limit)
   std::free(p);
else
   m_storage.push_back(p);
```

it's probably a good idea to ``TORRENT_ASSERT`` the packet size is greater than or equal to the slab size as well

arvidn: actually, I think ``p`` should be a ``std::unique_ptr``, so the first if-block should be removed. just:

if (m_storage.size() < m_limit) m_storage.push_back(std::move(p));

include/libtorrent/packet_pool.hpp
```diff
@@ -0,0 +1,156 @@
...
+					free(p);
+				else
+					m_storage[idx++] = p;
+			}
+
+			packet *try_alloc()
+			{
+				if (idx == 0)
```
arvidn: I think this function should return a ``std::unique_ptr<>``

```
if (m_storage.empty()) return static_cast<packet*>(malloc(sizeof(packet) + allocate_size));
auto ret = std::move(m_storage.back());
m_storage.pop_back();
return ret;
```

include/libtorrent/packet_pool.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+			if (allocated <= m_syn_slab.allocate_size) { m_syn_slab.try_free(p); }
+			else if (allocated <= m_mtu_floor_slab.allocate_size) { m_mtu_floor_slab.try_free(p); }
+			else if (allocated <= m_mtu_ceiling_slab.allocate_size) { m_mtu_ceiling_slab.try_free(p); }
+			else { free(p); }
+		}
+	private:
+
+		template<std::size_t PACKET_LIMIT, std::size_t ALLOCATE_SIZE>
+		struct packet_slab
```
arvidn: I think it would be better to not make this an inner class. It's easier to unit test and you improve encapsulation by moving it out.

include/libtorrent/packet_pool.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+		// the actual packet buffer
+		std::uint8_t buf[1];
+	};
+
+	// single thread packet allocation packet pool
+	// can handle common cases of packet size by 3 pools
+	struct TORRENT_EXTRA_EXPORT packet_pool : single_threaded
+	{
+		packet *acquire(std::size_t allocate)
```
arvidn: I think this should return a ``std::unique_ptr`` (you should probably have a type alias in here for the unique_ptr, since you need a custom deleter.) Make the deleter have a static ``operator()`` so you don't need to allocate any space for it, since all you do is call ``std::free()``.

arvidn: I suppose at this API boundary you could return a raw pointer for now and I can update the utp code to actually use ``std::unique_ptr`` later. you just call ``release()`` on it to get the raw pointer out

include/libtorrent/utp_socket_manager.hpp
```diff
@@ -119,6 +120,9 @@ namespace libtorrent
 		// the counter is the enum from ``counters``.
 		void inc_stats_counter(int counter, int delta = 1);
 
+		packet *acquire_packet(std::uint16_t allocate) { return m_packet_pool.acquire(allocate); }
```
arvidn: I don't think ``allocate`` should be ``uint16_t``. I think it should be ``int`` because it will be an arithmetic type. you can have a ``TORRENT_ASSERT`` internally to ensure it fits in a ``uint16_t``

arvidn:
you need a test for ``packet_pool`` and ``packet_slab`` too

include/libtorrent/packet_pool.hpp
ajax16384: ok

include/libtorrent/packet_pool.hpp
ajax16384: ok

include/libtorrent/packet_pool.hpp
ajax16384: vector::capacity() and vector::size() may be a little slow than vector::resize() and external last_free_slot_index. Anyway i'll try capacity variant

include/libtorrent/packet_pool.hpp
ajax16384: ok

include/libtorrent/utp_socket_manager.hpp
ajax16384: ok

include/libtorrent/packet_pool.hpp
ajax16384: ok

include/libtorrent/packet_pool.hpp
ajax16384: where is not way to declare static operator(), but it's not a problem. there will be no space waste - release build will deduce it to std::free call by compiler.

include/libtorrent/packet_pool.hpp
ajax16384: still leave allocate_size as template argument (it compiles as non mem touching comparision)

include/libtorrent/packet_pool.hpp
ajax16384: ok

include/libtorrent/packet_pool.hpp
ajax16384: ok

include/libtorrent/packet_pool.hpp
```diff
@@ -78,78 +78,82 @@ namespace libtorrent
 		std::uint8_t buf[1];
 	};
 
+
+	struct packet_deleter
+	{
+		void operator()(packet* handle) const
+		{
+			std::free(handle);
```
arvidn: awesome. one more thing I forgot to mention: would you mind also destructing the packet object?
``handle->~packet();`` just before freeing it

include/libtorrent/packet_pool.hpp
```diff
@@ -78,78 +78,82 @@ namespace libtorrent
...
+	using packet_ptr = std::unique_ptr<packet, packet_deleter>;
+
+	template<std::size_t ALLOCATE_SIZE>
+	struct TORRENT_EXTRA_EXPORT packet_slab : private aux::vector<packet_ptr, std::size_t>
+	{
+		static const std::size_t allocate_size{ ALLOCATE_SIZE };
+
+		packet_slab(std::size_t limit) : m_limit(limit) { reserve(m_limit); }
+
+		void try_free(packet_ptr &p)
```
arvidn: I don't think there's a need for this to *try* to free anymore. It could just be ``free()`` and it could take the ``packet_ptr`` by value, and let it free itself in the case where it's not added back into the cache

include/libtorrent/packet_pool.hpp
```diff
@@ -78,78 +78,82 @@ namespace libtorrent
...
+		{
+			if (size() < m_limit)
+				push_back(std::move(p));
+		}
+
+		void try_alloc(packet_ptr &p)
```
arvidn: this function seems to allocate the packet unconditionally. it should probably just be called ``alloc()``.
also, would you mind returning the ``packet_ptr`` instead?

include/libtorrent/packet_pool.hpp
```diff
@@ -78,78 +78,82 @@ namespace libtorrent
...
-			if (allocate <= m_syn_slab.allocate_size) { p = m_syn_slab.try_alloc(); }
-			else if (allocate <= m_mtu_floor_slab.allocate_size) { p = m_mtu_floor_slab.try_alloc(); }
-			else if (allocate <= m_mtu_ceiling_slab.allocate_size) { p = m_mtu_ceiling_slab.try_alloc(); }
-			else { p = static_cast<packet*>(malloc(sizeof(packet) + allocate)); }
+			if (allocate <= m_syn_slab.allocate_size) { m_syn_slab.try_alloc(p); }
+			else if (allocate <= m_mtu_floor_slab.allocate_size) { m_mtu_floor_slab.try_alloc(p); }
+			else if (allocate <= m_mtu_ceiling_slab.allocate_size) { m_mtu_ceiling_slab.try_alloc(p); }
+			else { m_rest_slab.try_alloc(p); }
 
 			p->allocated = static_cast<std::uint16_t>(allocate);
```
arvidn: I think it would be a bit more robust if the ``allocated`` member was filled in by the ``malloc()`` function, no?
I guess that wouldn't quite work with the "rest" slab. it seems a bit redundant to implement the general case via a slab though. it holds state that's never used

include/libtorrent/packet_pool.hpp
```diff
@@ -78,78 +78,82 @@ namespace libtorrent
...
+			if (empty())
+				p.reset(static_cast<packet*>(std::malloc(sizeof(packet) + allocate_size)));
+			else
+			{
+				p = std::move(back());
+				pop_back();
+			}
+		}
+	private:
+		std::size_t m_limit;
```
arvidn: this could be ``const``

include/libtorrent/packet_pool.hpp
```diff
@@ -78,78 +78,82 @@ namespace libtorrent
...
+		void try_free(packet_ptr &p)
+		{
+			if (size() < m_limit)
+				push_back(std::move(p));
+		}
+
+		void try_alloc(packet_ptr &p)
+		{
+			if (empty())
+				p.reset(static_cast<packet*>(std::malloc(sizeof(packet) + allocate_size)));
```
arvidn: would you mind calling the ``packet`` constructor here too?
``new (p.get()) packet();``

include/libtorrent/packet_pool.hpp
```diff
@@ -78,78 +78,82 @@ namespace libtorrent
...
-			std::size_t allocated = p->allocated;
 
-			if (allocated <= m_syn_slab.allocate_size) { m_syn_slab.try_free(p); }
-			else if (allocated <= m_mtu_floor_slab.allocate_size) { m_mtu_floor_slab.try_free(p); }
-			else if (allocated <= m_mtu_ceiling_slab.allocate_size) { m_mtu_ceiling_slab.try_free(p); }
-			else { free(p); }
+			packet_ptr pp(p); // take ownership and auto free
+			std::size_t allocated = pp->allocated;
+
+			if (allocated <= m_syn_slab.allocate_size) { m_syn_slab.try_free(pp); }
```
arvidn: so, I think this is the 3rd time I'm pointing this out, and I don't think you've responded yet. but these <= conditions must be a bug, right? it will let you put back a smaller allocation than what's supposed to be in the slab!

arvidn: I think you need to use ==

include/libtorrent/packet_pool.hpp
```diff
@@ -78,78 +78,82 @@ namespace libtorrent
...
-				else
-					return m_storage[--idx];
-			}
-		};
-
-		packet_slab<0x400, sizeof(utp_header)> m_syn_slab;
-		packet_slab<0x200, (TORRENT_INET_MIN_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_floor_slab;
-		packet_slab<0x200, (TORRENT_ETHERNET_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_ceiling_slab;
+		packet_slab<sizeof(utp_header)> m_syn_slab{ 0x400 };
+		packet_slab<(TORRENT_INET_MIN_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_floor_slab{ 0x200 };
```
arvidn: is this a common packet size btw? I wouldn't expect so. Did you collect any data on packet sizes, a histogram perhaps?
Also, these limits feel pretty high to me still. on a small system this could be way too much heap to keep around.
I also think we need some mechanism to gradually shrink these caches. perhaps some periodic tick in ``utp_socket_manager`` that frees some fraction of the cached packets. That could be a separate patch though, we should just keep it in mind

include/libtorrent/packet_pool.hpp
```diff
@@ -78,78 +78,82 @@ namespace libtorrent
...
+		packet_slab<(TORRENT_ETHERNET_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_ceiling_slab{ 0x200 };
+		packet_slab<0> m_rest_slab{ 0 };
```
arvidn: if you instead factor out the function that allocates memory, calls the constructor and puts it into a ``packet_ptr``, you could just use that instead of adding an extra (redundant) member. what do you think?

src/utp_stream.cpp
```diff
@@ -2444,7 +2409,7 @@ void utp_socket_impl::incoming(std::uint8_t const* buf, int size, packet* p
 	if (!p)
 	{
 		TORRENT_ASSERT(buf);
-		p = static_cast<packet*>(std::malloc(sizeof(packet) + size));
+		p = acquire_packet(std::uint16_t(size));
```
arvidn: again, I don't think this should be an unsigned type. please use ``int`` for the parameter here

arvidn:
I think this is starting to shape up

include/libtorrent/packet_pool.hpp
ajax16384: ok

include/libtorrent/packet_pool.hpp
ajax16384: ok

include/libtorrent/packet_pool.hpp
ajax16384: slab container can lease packets of smaller size.

include/libtorrent/packet_pool.hpp
ajax16384: i have made packet alloc/free recording of downloading of aprox 7gb torrent conntent. i'll try to decrease slab size.

src/utp_stream.cpp
ajax16384: ok

include/libtorrent/packet_pool.hpp
```diff
@@ -0,0 +1,160 @@
...
+	struct TORRENT_EXTRA_EXPORT packet_pool : private single_threaded
+	{
+		packet *acquire(std::size_t allocate)
+		{
+			TORRENT_ASSERT(is_single_thread());
+			TORRENT_ASSERT(allocate <= std::numeric_limits<std::uint16_t>::max());
+
+			packet_ptr p;
+
+			if (allocate <= m_syn_slab.allocate_size) { m_syn_slab.try_alloc(p); }
```
arvidn: you say a slab can have allocations *smaller* than its ``allocate_size``, right? In that case, you could get back a buffer that's smaller than ``allocate`` here, no? Which obviously would not work

include/libtorrent/packet_pool.hpp
arvidn: I think it makes sense to have a full MTU slab still. I consider it a problem that the majority of the packets you sent were 600 Bytes, I would like them to be larger than that

include/libtorrent/packet_pool.hpp
arvidn: I added a response at the point where packets are allocated out of the slabs

include/libtorrent/packet_pool.hpp
ajax16384: it still allocates full slab.allocate_size
 e.g. "pool.acquire" gets allocate == 10, by comparision we falls at first slab which allocates full slab size ==20, (packet->allocate fills with 10)
at "pool.release" of this packet we look at packet->allocate and safely return to first slab container cause it's really have 20 bytes allocated.

include/libtorrent/packet_pool.hpp
ajax16384: it;s 548== (TORRENT_INET_MIN_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)  (google sheets chart adds some axis number representation shift)

include/libtorrent/packet_pool.hpp
ajax16384: it is used at "release" (packet->allocated will determine to wich slab container packet should be "returned")

include/libtorrent/packet_pool.hpp
arvidn: I see, the assumption is that the packet must have been allocated from this pool to begin with. I think the ``allocated`` field in the packet struct should be set correctly to 20 in that case. Sometimes packets are extended (via nagle's or by inserting selective ACK headers), and having the allocation size accurate may save re-allcations

arvidn: doing that, you no longer need to assume that packets freed by this pool were allocated by it, and you can have == comparisons, and I think it becomes a bit more robust

include/libtorrent/packet_pool.hpp
```diff
@@ -0,0 +1,180 @@
...
+
+	private:
+		packet_ptr alloc(int allocate)
+		{
+			if (allocate <= m_syn_slab.allocate_size) { return m_syn_slab.alloc(); }
+			else if (allocate <= m_mtu_floor_slab.allocate_size) { return m_mtu_floor_slab.alloc(); }
+			else if (allocate <= m_mtu_ceiling_slab.allocate_size) { return m_mtu_ceiling_slab.alloc(); }
+			return packet_ptr(static_cast<packet*>(std::malloc(sizeof(packet) + allocate)));
+		}
+		packet_slab<sizeof(utp_header)> m_syn_slab{ 0x100 };
```
arvidn: In general I think the number of packets you cache here are too many. How many half-open connections do you expect to have at any given time? 256 seems like quite a lot, that's more than most people have total connections (at least that's my impression).
Is there a good reason to specify these counts in hex btw?

include/libtorrent/packet_pool.hpp
```diff
@@ -0,0 +1,180 @@
...
+			m_mtu_ceiling_slab.flush();
+		}
+
+	private:
+		packet_ptr alloc(int allocate)
+		{
+			if (allocate <= m_syn_slab.allocate_size) { return m_syn_slab.alloc(); }
+			else if (allocate <= m_mtu_floor_slab.allocate_size) { return m_mtu_floor_slab.alloc(); }
+			else if (allocate <= m_mtu_ceiling_slab.allocate_size) { return m_mtu_ceiling_slab.alloc(); }
+			return packet_ptr(static_cast<packet*>(std::malloc(sizeof(packet) + allocate)));
```
arvidn: when you allocate the packet here, you need to run the constructor too

arvidn: oh, I see. you do that in acquire, never mind

include/libtorrent/packet_pool.hpp
```diff
@@ -0,0 +1,180 @@
...
+	struct TORRENT_EXTRA_EXPORT packet_pool : private single_threaded
+	{
+		packet *acquire(int allocate)
+		{
+			TORRENT_ASSERT(is_single_thread());
+			TORRENT_ASSERT(allocate >= 0);
+			TORRENT_ASSERT(allocate <= std::numeric_limits<std::uint16_t>::max());
+
+			packet_ptr p{ alloc(allocate) };
+			new (p.get()) packet();
```
arvidn: There is an imbalance in the constructors and destructors now. You don't call the destructor until the memory is freed, but you call the constructor every time a buffer is returned, even if it's reused from a slab. I think you should move the call to the constructor into the ``alloc()`` function, and only call it immediately after ``malloc()``. That would restore balance

include/libtorrent/packet_pool.hpp
```diff
@@ -0,0 +1,180 @@
...
+				return packet_ptr(static_cast<packet*>(std::malloc(sizeof(packet) + allocate_size)));
+			else
+			{
+				auto ret = std::move(m_storage.back());
+				m_storage.pop_back();
+				return ret;
+			}
+		}
+
+		void flush() { m_storage.resize(0); }
```
arvidn: please use ``m_storage.clear()`` instead

include/libtorrent/packet_pool.hpp
```diff
@@ -0,0 +1,180 @@
...
+			packet_ptr pp(p); // takes ownership and may auto free if slab container does not move it
+			std::size_t allocated = pp->allocated;
+
+			if (allocated <= m_syn_slab.allocate_size) { m_syn_slab.try_push_back(pp); }
+			else if (allocated <= m_mtu_floor_slab.allocate_size) { m_mtu_floor_slab.try_push_back(pp); }
+			else if (allocated <= m_mtu_ceiling_slab.allocate_size) { m_mtu_ceiling_slab.try_push_back(pp); }
+		}
+
+		//TODO: call from utp manager to flush mem
+		void flush()
```
arvidn: do you have any ideas of how to call this?
One way (which may be a bit involved) would be to have a exponential decay inside each slab. flush is called periodically and it frees some portion of the buffers in there. The slabs only contain the "head room", so it would not interfere with heavy churn of the buffers

arvidn:
looks good (apart from the constructor comment). Do you want to address the flushing/maintenance  in this PR too?

include/libtorrent/packet_pool.hpp
```diff
@@ -0,0 +1,180 @@
...
+				auto ret = std::move(m_storage.back());
+				m_storage.pop_back();
+				return ret;
+			}
+		}
+
+		void flush() { m_storage.resize(0); }
+	private:
+		const std::size_t m_limit;
+		aux::vector<packet_ptr, std::size_t> m_storage;
```
aldenml: any particular reason to use `aux::vector` instead of `std::vector`?

ajax16384: i thought all std::vector inside libtorrent should migrate to these implementation (it has some usefull asserts?)

include/libtorrent/packet_pool.hpp
ajax16384: there is not reason, let's use dec and decrease it

include/libtorrent/packet_pool.hpp
ajax16384: ok

include/libtorrent/packet_pool.hpp
ajax16384: ok

include/libtorrent/packet_pool.hpp
arvidn: most standard library implementations can have asserts/debug-iterators enabled too. The main purpose of this wrapper is to enforce specific types to index into it. In your case you use the same type as the standard vector, so it doesn't add anything.

include/libtorrent/packet_pool.hpp
arvidn: in fact, the size of these is not directly related to how popular a packet size is. You can have thousands of packets allocated, sitting in the send buffer, with a very small slab limit. This size only determines how many to keep around when churn fluctuates. I would think you would get the majority of performance wins by making these be limited to 10 or 20 or so.

ajax16384:
packet flush can be made in separate PR

include/libtorrent/packet_pool.hpp
```diff
@@ -79,16 +79,24 @@ namespace libtorrent
...
-			std::free(handle);
+			p->~packet();
+			std::free(p);
+		}
+
+		static packet *create_packet(int size)
+		{
+			packet *p = static_cast<packet*>(std::malloc(sizeof(packet) + size));
+			new (p) packet();
+			return p;
```
arvidn: It's not clear to me what the benefit of conflating these two is. This static member function does not need access to any members in this class. it should probably just be a free function. I think calling this struct a deleter is very helpful too, because that's what the standard refers to by.

include/libtorrent/packet_pool.hpp
```diff
@@ -169,11 +176,11 @@ namespace libtorrent
...
 			else if (allocate <= m_mtu_ceiling_slab.allocate_size) { return m_mtu_ceiling_slab.alloc(); }
-			return packet_ptr(static_cast<packet*>(std::malloc(sizeof(packet) + allocate)));
+			return packet_ptr(packet_mem_manager::create_packet(allocate));
 		}
-		packet_slab<sizeof(utp_header)> m_syn_slab{ 0x100 };
-		packet_slab<(TORRENT_INET_MIN_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_floor_slab{ 0x200 };
-		packet_slab<(TORRENT_ETHERNET_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_ceiling_slab{ 0x100 };
+		packet_slab<sizeof(utp_header)> m_syn_slab{ 100 };
+		packet_slab<(TORRENT_INET_MIN_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_floor_slab{ 512 };
+		packet_slab<(TORRENT_ETHERNET_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER)> m_mtu_ceiling_slab{ 256 };
```
arvidn: still, this seems too large. these limits don't determine how many packets you can have in flight, it's how many packets you'll keep around in between bursts. I don't think it's important that these are large numbers. In fact, I think the risk of keeping too much memory tied up for too long is a greater risk

arvidn: I would think these could be one order of magnitude less

ajax16384: ok

include/libtorrent/packet_pool.hpp
ajax16384: ok

include/libtorrent/packet_pool.hpp
```diff
@@ -79,24 +79,24 @@ namespace libtorrent
...
-		static packet *create_packet(int size)
-		{
-			packet *p = static_cast<packet*>(std::malloc(sizeof(packet) + size));
-			new (p) packet();
-			return p;
-		}
 	};
 
-	using packet_ptr = std::unique_ptr<packet, packet_mem_manager>;
+	static packet *create_packet(int size)
```
arvidn: sorry for nit-picking. I don't think you should define a static function in a header file. Either put it in a .cpp file or make it ``inline``.

arvidn: since there is no .cpp file in this case, I think you should make it inline

ajax16384: ok

include/libtorrent/packet_pool.hpp
```diff
@@ -89,7 +89,7 @@ namespace libtorrent
 		}
 	};
 
-	static packet *create_packet(int size)
+	inline packet *create_packet(int size)
```
arvidn: this broke the documentation generation tool/rst.

https://travis-ci.org/arvidn/libtorrent/jobs/199527166#L1455

My guess is because you put the asterisk to the right instead of to the left (where it really belongs). It seems a little bit far fetched though, but this didn't break before, right?

arvidn:
lgtm

PR #1601, opened 2017-01-22
fix missed static array specifier

ajax16384:


arvidn:
lgtm

arvidn:
in fact, if you don't mind, would you back-port this to RC_1_1?

ajax16384:
np

PR #1604, opened 2017-01-23
fixing sign-conversion warnings, part 7

aldenml:


src/alert.cpp
```diff
@@ -1952,10 +1952,11 @@ namespace libtorrent {
 	}
 #endif
 	std::vector<tcp::endpoint> dht_get_peers_reply_alert::peers() const {
-		std::vector<tcp::endpoint> peers(m_num_peers);
+		std::size_t num_peers = aux::numeric_cast<std::size_t>(m_num_peers);
+		std::vector<tcp::endpoint> peers(num_peers);
```
arvidn: I imagine both of these could be const

aldenml: yes

arvidn:
lgtm

PR #1615, opened 2017-01-24
Add  web seed name lookup retry minutes to session_settings

ximply:
It will take 30 minutes to retry name lookup when web seed name lookup
failed default. You can restart libtorrent client to retry but it is not
convenient, so just add a setting to session_settings to control it is a
good idea.

### DIFF
--- a/include/libtorrent/session_settings.hpp
+++ b/include/libtorrent/session_settings.hpp
@@ -229,6 +229,9 @@ namespace libtorrent
 		// time to wait until a new retry takes place
 		int urlseed_wait_retry;
 		
+		// if unavailable, retry in these minutes
+		int web_seed_name_lookup_retry_minutes;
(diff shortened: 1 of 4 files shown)


include/libtorrent/session_settings.hpp
```diff
@@ -229,6 +229,9 @@ namespace libtorrent
 		// time to wait until a new retry takes place
 		int urlseed_wait_retry;
 		
+		// if unavailable, retry in these minutes
+		int web_seed_name_lookup_retry_minutes;
```
arvidn: unfortunately, adding this breaks the ABI. You can add a new setting at the end to RC_1_1 though.

ximply: Well, I will commit a new PR.

PR #1616, opened 2017-01-24
fixing sign-conversion warnings, part 8, created aux::unique_ptr, refactor

aldenml:


include/libtorrent/aux_/unique_ptr.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <memory>
+#include <type_traits>
+
+#include "libtorrent/units.hpp"
+#include "libtorrent/assert.hpp"
+
+namespace libtorrent { namespace aux {
+
+	template <typename T, typename IndexType = int>
+	struct unique_ptr : std::unique_ptr<T>
```
arvidn: since this only makes sense for ``unique_ptr``s holding an array, maybe this should be an undefined type. i.e. remove ``{}``. would that work?

arvidn:
lgtm

include/libtorrent/aux_/unique_ptr.hpp
aldenml: good idea, and it works

PR #1617, opened 2017-01-24
1.1.2 set queue position by number

Falcosc:
Something of my local extensions.

If you like it I will create a test for it.

This is needed to sort queue without dropping connections. Currently I use queue top which will drop connections because I have to set queue top for all entries starting at lowest priority. The low entries queue top calls kicks active entries out of active range.

Offtopic
These connection drops cased by moving queue entries every 10 minutes does maybe cause the illegal torrent piece state, but we will see it after we found the illegal state description.

### DIFF
--- a/include/libtorrent/torrent_handle.hpp
+++ b/include/libtorrent/torrent_handle.hpp
@@ -833,9 +833,13 @@ namespace libtorrent
 		int queue_position() const;
 		void queue_position_up() const;
 		void queue_position_down() const;
-		void queue_position_set(uint p) const;
 		void queue_position_top() const;
 		void queue_position_bottom() const;
+		// set the new position in queue starting by 0. If given position does
+		// already exist your torrent will be set on top of it and remaining once
+		// get posion + 1. If the given position is greater then queue it will be
+		// added on bottom
+		void queue_position_set(uint p) const;
 
 #ifndef TORRENT_NO_DEPRECATE
 		// deprecated in 1.1
--- a/include/libtorrent/torrent_handle.hpp
+++ b/include/libtorrent/torrent_handle.hpp
@@ -835,6 +835,7 @@ namespace libtorrent
 		void queue_position_down() const;
 		void queue_position_top() const;
 		void queue_position_bottom() const;
+
 		// set the new position in queue starting by 0. If given position does
 		// already exist your torrent will be set on top of it and remaining once
 		// get posion + 1. If the given position is greater then queue it will be
--- a/include/libtorrent/torrent_handle.hpp
+++ b/include/libtorrent/torrent_handle.hpp
@@ -840,7 +840,7 @@ namespace libtorrent
 		// already exist your torrent will be set on top of it and remaining once
 		// get posion + 1. If the given position is greater then queue it will be
 		// added on bottom
-		void queue_position_set(uint p) const;
+		void queue_position_set(int p) const;
 
 #ifndef TORRENT_NO_DEPRECATE
 		// deprecated in 1.1
--- a/include/libtorrent/torrent_handle.hpp
+++ b/include/libtorrent/torrent_handle.hpp
@@ -833,6 +833,7 @@ namespace libtorrent
 		int queue_position() const;
 		void queue_position_up() const;
 		void queue_position_down() const;
+		void queue_position_set(uint p) const;
 		void queue_position_top() const;
 		void queue_position_bottom() const;
 
--- a/include/libtorrent/torrent_handle.hpp
+++ b/include/libtorrent/torrent_handle.hpp
@@ -836,10 +836,9 @@ namespace libtorrent
 		void queue_position_top() const;
 		void queue_position_bottom() const;
 
-		// set the new position in queue starting by 0. If given position does
-		// already exist your torrent will be set on top of it and remaining once
-		// get posion + 1. If the given position is greater then queue it will be
-		// added on bottom
+		// updates the position in the queue for this torrent. The relative order
+		// of all other torrents remain intact but their numerical queue position
+		// shifts to make space for this torrent's new position
 		void queue_position_set(int p) const;
 
 #ifndef TORRENT_NO_DEPRECATE
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -9239,13 +9239,13 @@ namespace libtorrent
 	void torrent::set_queue_position(int p)
 	{
 		TORRENT_ASSERT(is_single_thread());
-		TORRENT_ASSERT((p == -1) == is_finished()
-			|| (!m_auto_managed && p == -1)
-			|| (m_abort && p == -1));
-		if (is_finished() && p != -1) return;
+		if(is_finished() || is_seed() || !m_auto_managed || m_abort)
+		{
+			p = -1;
+		} else {
+			p = p < 0 ? 0 : p;
+		}
 		if (p == m_sequence_number) return;
-
-		TORRENT_ASSERT(p >= -1);
 
 		state_updated();
 
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -9227,25 +9227,34 @@ namespace libtorrent
 
 	void torrent::queue_up()
 	{
+		// fix async position change calls for race conditions on changed torrents
+		if(!m_auto_managed || m_abort || is_finished()) return;
+
 		set_queue_position(queue_position() == 0
 			? queue_position() : queue_position() - 1);
 	}
 
 	void torrent::queue_down()
 	{
+		// fix async position change calls for race conditions on changed torrents
+		if(!m_auto_managed || m_abort || is_finished()) return;
+
 		set_queue_position(queue_position() + 1);
 	}
 
 	void torrent::set_queue_position(int p)
 	{
 		TORRENT_ASSERT(is_single_thread());
-		if(is_finished() || is_seed() || !m_auto_managed || m_abort)
-		{
-			p = -1;
-		} else {
-			p = p < 0 ? 0 : p;
-		}
+
+		// fix async position change calls for race conditions on changed torrents
+		if ((!m_auto_managed || m_abort || is_finished()) && p != -1) return;
+
+		TORRENT_ASSERT((p == -1) == is_finished()
+			|| (!m_auto_managed && p == -1)
+			|| (m_abort && p == -1));
 		if (p == m_sequence_number) return;
+
+		TORRENT_ASSERT(p >= -1);
 
 		state_updated();
 
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -9227,7 +9227,7 @@ namespace libtorrent
 
 	void torrent::queue_up()
 	{
-		// fix async position change calls for race conditions on changed torrents
+		// fix race conditions on async position change calls (from handler)
 		if(!m_auto_managed || m_abort || is_finished()) return;
 
 		set_queue_position(queue_position() == 0
@@ -9236,9 +9236,6 @@ namespace libtorrent
 
 	void torrent::queue_down()
 	{
-		// fix async position change calls for race conditions on changed torrents
-		if(!m_auto_managed || m_abort || is_finished()) return;
-
 		set_queue_position(queue_position() + 1);
 	}
 
@@ -9246,7 +9243,7 @@ namespace libtorrent
 	{
 		TORRENT_ASSERT(is_single_thread());
 
-		// fix async position change calls for race conditions on changed torrents
+		// fix race conditions on async position change calls (from handler)
 		if ((!m_auto_managed || m_abort || is_finished()) && p != -1) return;
 
 		TORRENT_ASSERT((p == -1) == is_finished()
--- a/src/torrent_handle.cpp
+++ b/src/torrent_handle.cpp
@@ -363,7 +363,7 @@ namespace libtorrent
 
 	void torrent_handle::queue_position_set(int p) const
 	{
-		TORRENT_ASYNC_CALL1(set_queue_position, p);
+		TORRENT_ASYNC_CALL1(set_queue_position, p < 0 ? 0 : p);
 	}
 
 	void torrent_handle::queue_position_top() const
--- a/src/torrent_handle.cpp
+++ b/src/torrent_handle.cpp
@@ -363,6 +363,7 @@ namespace libtorrent
 
 	void torrent_handle::queue_position_set(int p) const
 	{
+		TORRENT_ASSERT_PRECOND(p >= 0);
 		TORRENT_ASYNC_CALL1(set_queue_position, p);
 	}
 
--- a/src/torrent_handle.cpp
+++ b/src/torrent_handle.cpp
@@ -361,6 +361,11 @@ namespace libtorrent
 		TORRENT_ASYNC_CALL(queue_down);
 	}
 
+	void torrent_handle::queue_position_set(uint p) const
+	{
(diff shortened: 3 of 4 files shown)


arvidn:
seems reasonable

Falcosc:
I can't initialize a torrent with queue position because the queue position is set somewhere during session boot, it's not enough to start a session and add some torrents there.

But I don’t want to create a simulator test because it would be not part of our code coverage.
What do you suggest?

I did test already existing code to exclude my incompetence, but same problem there, for example https://github.com/arvidn/libtorrent/blob/3f1084d63d591ec9afd64c07b65c9fec98b40eaa/test/test_torrent.cpp#L93 a `TEST_EQUAL(h.queue_position(), 0);`
Sometimes it is 0, sometimes it is -1
I didn't even understand why it is not predictable because I thought it is a single thread execution or does session have it's own thread? (if it's a own thread it would make sense because result would be runtime specific)

arvidn:
the session is single threaded, as in it has exactly one of its own threads. iirc, the queue number is assigned sequentially by the order torrents are added. if you're adding torrents by ``file://`` path, they will be loaded by the disk thread(s), which will then add them to the session in an unspecified order

Falcosc:
That should be the reason for the behaivor.

But I did have trouble with the waiting for threads too.
I did try https://github.com/arvidn/libtorrent/blob/RC_1_1/test/setup_transfer.cpp#L408 without success result was still unpredictable, somtimes no queue (-1) and sometimes 0.
This method was removed in master, should the new one
`std::this_thread::sleep_for(lt::milliseconds(500));` work better or is there another way to wait for the correct torrent state?

arvidn:
I tried to replace sleeps with functions that wait for specific alerts. I imagine you could use that too (as most states are announced via alerts)

Falcosc:
Is it possible to wait for an alert without using the simulator?
Or should I create this test at simulator?

arvidn:
This test utility function waits for an alert: https://github.com/arvidn/libtorrent/blob/master/test/setup_transfer.cpp#L178

Could you expand the documentation a bit? I'm not sure exactly how this function affects the queue position for other torrents. If you want to explicitly set the queue position, do you need to start at 0 and go up? I suspect if you start by setting the last queue position, it will keep changing as you set the ones before it, right?

Falcosc:
I guess you did implement it by using the number as target because the wrapper queueTop() use QueueSet(0)
And queueBotton() use maxint
and existing ones get a position +1

Is that description ok?:

`// set the new position in queue starting by 0. If given position does already exist your torrent will be set on top of it and remaining once get posion + 1. If the given position is greater then queue it will be added on botton`

Falcosc:
I don't know what was wrong yesterday. It works fine even without wait.

arvidn:
I'm not sure I understand that description. I would imagine (assuming this is how it works) to mention that torrents with a higher queue position will shift down

include/libtorrent/torrent_handle.hpp
```diff
@@ -835,6 +835,11 @@ namespace libtorrent
 		void queue_position_down() const;
 		void queue_position_top() const;
 		void queue_position_bottom() const;
+		// set the new position in queue starting by 0. If given position does
```
arvidn: I believe you may need a blank line here, for the documentation tool to pick this up correctly

test/test_torrent.cpp
```diff
@@ -392,4 +392,64 @@ TORRENT_TEST(torrent_total_size_zero)
 	TEST_CHECK(ec);
 }
 
+TORRENT_TEST(queue)
```
arvidn: lgtm

include/libtorrent/torrent_handle.hpp
```diff
@@ -835,6 +835,7 @@ namespace libtorrent
 		void queue_position_down() const;
 		void queue_position_top() const;
 		void queue_position_bottom() const;
+
 		// set the new position in queue starting by 0. If given position does
```
arvidn: please say: "updates the position in the queue for this torrent. The relative order of all other torrents remain intact but their numerical queue position shifts to make space for this torrent's new position"

src/torrent_handle.cpp
```diff
@@ -361,6 +361,11 @@ namespace libtorrent
 		TORRENT_ASYNC_CALL(queue_down);
 	}
 
+	void torrent_handle::queue_position_set(uint p) const
```
arvidn: you noticed the compilation error here, right? use ``int``

Falcosc: No, I didn't.

Falcosc:
https://github.com/arvidn/libtorrent/blob/master/src/torrent.cpp#L7943-L7968

remove the asserts from set_queue_position or use uint at handle with int conversion at method call

What would you like more?

arvidn:
> remove the asserts from set_queue_position or use uint at handle with int conversion at method call or should the libtorrent user care about wrong values?

I think that you should use ``int``.  (``uint`` isn't a type, as the compilation error indicated). http://blog.libtorrent.org/2016/05/unsigned-integers/

as for the asserts. what is the problems with the asserts? I suppose now that this function is exposed to users, the wrapper function on ``torrent_handle`` should validate input and also use ``TORRENT_ASSERT_PRECOND()``. Would you mind adding that?

I suppose the client side doesn't know whether the torrent is completed or not, so that has to be handled at run-time on the libtorrent side.

Falcosc:
Yes, validation at torrent handle is fine. But I don't know how to use TORRENT_ASSERT_PRECOND()? I guess it will destry the session like the other ones, does it?

arvidn:
``TORRENT_ASSERT_PRECOND()`` is just like ``TORRENT_ASSERT()`` or ``assert()``. It aborts the process in debug builds and disappears in release builds.

The main distinction between PRECOND and not is that the message indicates that the client has a problem, as opposed to libtorrent.

include/libtorrent/torrent_handle.hpp
```diff
@@ -840,7 +840,7 @@ namespace libtorrent
 		// already exist your torrent will be set on top of it and remaining once
 		// get posion + 1. If the given position is greater then queue it will be
 		// added on bottom
```
arvidn: this comment does not seem to recognize that every torrent (unless it's a seed) already has a queue position. when you change the queue position of a torrent, you're not only _adding_ a torrent to the queue, you're also _removing_ it from the queue first. i.e. your example of how other torrents get position + 1 is misleading because that's only true when you're moving a torrent up the queue (closer to the beginning). If you're moving the first torrent to the last position, every torrent in between will get the position - 1.

This is why the way I suggested phrasing it seems more accurate (and succinct).

Falcosc: Sorry, I forgott your commet "updates the position in the queue for this torrent. The relative order of all other torrents remain intact but their numerical queue position shifts to make space for this torrent's new position"

src/torrent_handle.cpp
```diff
@@ -361,6 +361,11 @@ namespace libtorrent
 		TORRENT_ASYNC_CALL(queue_down);
 	}
 
+	void torrent_handle::queue_position_set(int p) const
+	{
+		TORRENT_ASYNC_CALL1(set_queue_position, p < 0 ? 0 : p);
```
arvidn: you should have a ``TORRENT_ASSERT_PRECOND(p >= 0, "queue position must be positive");`` here

test/test_torrent.cpp
```diff
@@ -392,4 +392,93 @@ TORRENT_TEST(torrent_total_size_zero)
...
+		bencode(std::back_inserter(buf), t.generate());
+		boost::shared_ptr<torrent_info> ti = boost::make_shared<torrent_info>(&buf[0], buf.size());
+		add_torrent_params p;
+		p.ti = ti;
+		p.save_path = ".";
+		torrents[i] = ses.add_torrent(p);
+	}
+
+	// add_torrent should be ordered
+	TEST_EQUAL(torrents[0].queue_position(), 0);
```
arvidn: as you mentioned the asserts. I would imagine currently this call will fail if the torrent is ``finished`` or ``seed``. Because there's an assert that such torrents can only have queue position -1 (iirc). It might make sense to test that too. You could just set the priority of all pieces to 0 to make it finished.

And also turn the assert into a run-time check to set it to -1 unconditionally if it's finished. The reason that assert can't stay with this function exposed is that there's an inherent race condition. The client cannot know whether the torrent will be finished by the time the message reaches the libtorrent thread.

Falcosc: You mean something like if seed or finished: return?

thanks for the priority 0 suggestion.

arvidn: no. if you would just return, seeding torrents would never be able to have their queue position set to -1.

Falcosc: you are right, I forget that.
Thanks.

src/torrent.cpp
```diff
@@ -9239,14 +9239,14 @@ namespace libtorrent
...
-		if (is_finished() && p != -1) return;
+		if(is_finished() || is_seed() || !m_auto_managed || m_abort)
+		{
+			p = -1;
+		} else {
+			p = p < 0 ? 0 : p;
+		}
 		if (p == m_sequence_number) return;
 
-		TORRENT_ASSERT(p >= -1);
```
arvidn: I don't see why this assert needs to go

src/torrent.cpp
```diff
@@ -9239,14 +9239,14 @@ namespace libtorrent
...
 		TORRENT_ASSERT(is_single_thread());
-		TORRENT_ASSERT((p == -1) == is_finished()
-			|| (!m_auto_managed && p == -1)
-			|| (m_abort && p == -1));
-		if (is_finished() && p != -1) return;
+		if(is_finished() || is_seed() || !m_auto_managed || m_abort)
+		{
+			p = -1;
+		} else {
+			p = p < 0 ? 0 : p;
```
arvidn: this seems to alter the semantics of this function. why should -1 turn into 0?

src/torrent.cpp
```diff
@@ -9239,14 +9239,14 @@ namespace libtorrent
 	void torrent::set_queue_position(int p)
 	{
 		TORRENT_ASSERT(is_single_thread());
-		TORRENT_ASSERT((p == -1) == is_finished()
-			|| (!m_auto_managed && p == -1)
-			|| (m_abort && p == -1));
-		if (is_finished() && p != -1) return;
```
arvidn: this early exit condition seems to still make sense. was there a case where this would break anything?

src/torrent_handle.cpp
```diff
@@ -363,7 +363,7 @@ namespace libtorrent
 
 	void torrent_handle::queue_position_set(int p) const
 	{
-		TORRENT_ASYNC_CALL1(set_queue_position, p < 0 ? 0 : p);
+		TORRENT_ASYNC_CALL1(set_queue_position, p);
```
arvidn: I see, you moved this check into the libtorrent side. I'm not convinced re-interpreting negative numbers as 0 is reasonable. I think having an assert (TORRENT_ASSERT_PRECOND) and just ignore the call when made with an invalid argument, might make more sense

test/test_torrent.cpp
```diff
@@ -399,8 +399,8 @@ TORRENT_TEST(queue)
 	pack.set_bool(settings_pack::disable_hash_checks, true);
 	lt::session ses(pack);
 
-	torrent_handle torrents[5];
-	for(int i = 0; i < 5; i++)
+	torrent_handle torrents[6];
```
arvidn: I think it would be nicer to make this a ``std::vector<torrent_handle> torrents;``. and then use ``torrents.push_back()`` when adding a new torrent, rather than directly indexing into it

test/test_torrent.cpp
```diff
@@ -474,7 +491,9 @@ TORRENT_TEST(queue)
 	// test set pos on not existing pos
 	torrents[3].queue_position_set(10);
 	torrents[2].queue_position_set(-10);
+	finished.queue_position_set(-10);
```
arvidn: negative numbers are never valid in calls to this function. I think a more realistic test would be to set it to some valid queue position, and still assert that it remains -1, since it's finished

src/torrent.cpp
Falcosc: to catch user faults

Or should this moved back to torrent handler?

src/torrent_handle.cpp
Falcosc: you are right

src/torrent.cpp
Falcosc: the return could not reached anymore because -1 is forced and catched by `if (p == m_sequence_number) return;`

Falcosc: I see, queue up has an problem with finished torrents

arvidn:
compilation error: https://travis-ci.org/arvidn/libtorrent/jobs/196359073#L1163

Falcosc:
is c++11 missing on build configuration?

arvidn:
RC_1_0 supports C++98

src/torrent.cpp
```diff
@@ -9227,22 +9227,31 @@ namespace libtorrent
 
 	void torrent::queue_up()
 	{
+		// fix async position change calls for race conditions on changed torrents
+		if(!m_auto_managed || m_abort || is_finished()) return;
```
arvidn: can you explain this? where's the race condition? this is a function on ``torrent`` which is run in the libtorrent thread. there's no opportunity for the torrent to change state between the call to ``queue_position()`` and ``set_queue_position()``

src/torrent.cpp
```diff
@@ -9227,22 +9227,31 @@ namespace libtorrent
...
+
 		set_queue_position(queue_position() + 1);
 	}
 
 	void torrent::set_queue_position(int p)
 	{
 		TORRENT_ASSERT(is_single_thread());
+
+		// fix async position change calls for race conditions on changed torrents
+		if ((!m_auto_managed || m_abort || is_finished()) && p != -1) return;
```
arvidn: I think you meant to add something like this to ``torrent_handle``. but clearly you can't know the state of the torrent at that point. the only thing you can know is that queue positions < 0 are invalid.

arvidn: oh, sorry, this comment was meant on the functions above. for ``set_queue_position()``, this check might make sense

src/torrent_handle.cpp
```diff
@@ -361,6 +361,12 @@ namespace libtorrent
 		TORRENT_ASYNC_CALL(queue_down);
 	}
 
+	void torrent_handle::queue_position_set(int p) const
+	{
+		TORRENT_ASSERT_PRECOND(p >= 0);
```
arvidn: when built in release mode, you should have a second check too. ``if (p < 0) return;``

src/torrent.cpp
arvidn: at a second though, I suppose this makes sense. was this a problem before exposing ``set_queue_position()`` then?

Falcosc: Jepp, 
You can't call up on finished torrent, it will run into the old set_queue_position assert

arvidn: @Falcosc do you remember why ``m_auto_managed`` is part of this condition? This is the thing that prevents paused torrents (or any non-auto-managed torrent) from having its queue position updated. https://github.com/arvidn/libtorrent/issues/2097

Falcosc: I would guess that we tought that m_auto_managed is always true if queue management is enabled. 
We could have an issue if m_auto_managed is false on paused torrents.
Or we did worry that we stop the queue if 5 paused torrents are on top and only 3 queue entries could be active.

PR #1618, opened 2017-01-24
Add web seed name lookup retry minutes to session_settings 1.1

ximply:


include/libtorrent/session_settings.hpp
```diff
@@ -155,6 +155,9 @@ namespace libtorrent
 		// time to wait until a new retry takes place
 		int urlseed_wait_retry;
 
+		// if unavailable, retry in these minutes
+		int web_seed_name_lookup_retry_minutes;
```
arvidn: this structure is deprecated. I would prefer to not add anything to it

src/settings_pack.cpp
```diff
@@ -231,6 +231,7 @@ namespace libtorrent
 		SET(urlseed_pipeline_size, 5, 0),
 		SET_NOPREV(urlseed_max_request_bytes, 16 * 1024 * 1024, 0),
 		SET(urlseed_wait_retry, 30, 0),
+		SET(web_seed_name_lookup_retry_minutes, 30, 0),
```
arvidn: instead, you can use the ``SET_NOPREV`` macro. this will all be cleaned up once ``session_settings`` is removed

include/libtorrent/settings_pack.hpp
```diff
@@ -792,6 +792,9 @@ namespace libtorrent
 			// time to wait until a new retry of a web seed takes place
 			urlseed_wait_retry,
 
+			// if unavailable, retry in these minutes
+			web_seed_name_lookup_retry_minutes,
```
arvidn: please add a little bit more comprehensive documentation for this setting. I also believe all (afaict) times are specified in seconds. I think it would make sense to specify this in seconds as well.

arvidn:
also, please add a line to Changelog indicating this feature was added

arvidn:
I hope to cut 1.1.2 tomorrow. so you know in case you wanted this in.

PR #1622, opened 2017-01-25
fixed finished-time calculation

arvidn:


arvidn:
@Falcosc does this look right?

Falcosc:
Yes, it does.

PR #1624, opened 2017-01-25
fix issue with unloading torrents

arvidn:


Falcosc:
Would be nice if you could fix the feature description, because I don't understand the feature but would like to undertand it to know what this fix did improve.
https://github.com/arvidn/libtorrent/blob/RC_1_1/include/libtorrent/torrent.hpp#L1635

Looks like it got removed in master, if this gets removed, we don't need to improve the description.

Thanks

arvidn:
yeah, I did remove this feature in master. This basically was an attempt at scalability by "unloading"/evicting torrents from memory when they were inactive, using an LRU. It did not improve scalability enough to warrant the additional complexity (such as this bug).

PR #1625, opened 2017-01-25
fix merge issue in python endpoint converter

arvidn:


Falcosc:
Here is a build result with error stream output:

https://travis-ci.org/arvidn/libtorrent/jobs/195074594#L3507
`KeyError: 'unknown name in torrent params: auto_managed'`

arvidn:
I wonder why this passed the test before the merged

Falcosc:
I guess because the pull requests did not get a rebase on buildhost.
That means the pull request did not include the make torrent test.

If you could confirm thing tought I need to rebase all my PRs because I guess my suspicious test issues could have a similar reason.

Falcosc:
AttributeError: type object 'add_torrent_params' has no attribute 'flag_duplicate_is_error'

You could execute only one test class to speed up the fix confirmation
`bjam -j4 libtorrent-link=static boost-link=static stage_module && python test.py test-examples test_example_client`

And it would be nice if you could move the last 4 tests  to class test_session(unittest.TestCase)
- test_add_torrent
- test_unknown_settings
- test_apply_settings
- test_fingerprint

arvidn:
that seems to have fixed it

PR #1629, opened 2017-01-25
finished non-seed torrents can also be in super-seeding mode (1.0)

arvidn:


sledgehammer999:
I think there is another minor issue. It seems that `post_torrent_updates()`/`state_update_alert` doesn't consider this change in status and doesn't report it in the alert.
Explicitly requesting the status seems to report the new value back.
Can you confirm?

arvidn:
thanks, I'll look into that

arvidn:
I think that update should fix it

sledgehammer999:
I confirm that it works.

PR #1630, opened 2017-01-26
making strong_typedef::UnderlyingType constexpr and warnigs cleanup

aldenml:


Jamfile
```diff
@@ -276,6 +276,9 @@ rule warnings ( properties * )
 		result += <cflags>/wd4275 ;
 # disable warning C4373: virtual function overrides, previous versions of the compiler did not override when parameters only differed by const/volatile qualifiers
 		result += <cflags>/wd4373 ;
+		# C4268: 'identifier' : 'const' static/global data initialized
+		#        with compiler generated default constructor fills the object with zeros
+		result += <cflags>/wd4268 ;
```
arvidn: do you understand this warning? I get the impression it's a bug in msvc. the tests pass, suggesting these do get initialized correctly

arvidn:
lgtm

Jamfile
aldenml: I found in internet that in a very similar situation, one msvc's developer documented it was a bug.

PR #1631, opened 2017-01-26
fixing sign-conversion warnings, part 9

aldenml:


arvidn:
lgtm

include/libtorrent/aux_/unique_ptr.hpp
```diff
@@ -50,6 +50,9 @@ namespace libtorrent { namespace aux {
 		using base = std::unique_ptr<T[]>;
 		using underlying_index = typename underlying_index_t<IndexType>::type;
 
+		unique_ptr() {}
+		explicit unique_ptr(T arr[]) : base(arr) {}
```
arvidn: I think it's a bit misleading to use the ``T arr[]`` syntax. It means the same as ``T* arr``, but that might not be obvious to everyone

arvidn: in fact, I think it would be better to just pull in the base class' constructors, instead of defining new ones.

I believe this should do it: ``using base::base;``

aldenml: I tried, it's not possible due to some templating issue

PR #1633, opened 2017-01-27
build single-page version of documentation

arvidn:


arvidn:
is there an easy way to make aafigure optional? detect whether the tool is installed and not include those targets?

PR #1634, opened 2017-01-27
moved alloca.hpp to private aux_ and size refactor

aldenml:
The context for this is that I want to use private api `aux::numeric_cast` and an eventual `aux::typed_span`

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -67,7 +67,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/aux_/disable_warnings_pop.hpp"
 
 #include "libtorrent/config.hpp"
-#include "libtorrent/alloca.hpp"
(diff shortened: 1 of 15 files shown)


arvidn:
lgtm

aldenml:
good (I need to fix error)

src/file.cpp
```diff
@@ -67,7 +67,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/aux_/disable_warnings_pop.hpp"
 
 #include "libtorrent/config.hpp"
-#include "libtorrent/alloca.hpp"
+#include "libtorrent/aux_/alloca.hpp"
```
arvidn: it appears file.cpp is not missing ``aux::numeric_cast`` for some reason. not obvious why that happened

PR #1636, opened 2017-01-27
created aux::typed_span and related refactor

aldenml:


include/libtorrent/aux_/typed_span.hpp
```diff
@@ -0,0 +1,126 @@
...
+#include <type_traits>
+
+#include "libtorrent/units.hpp"
+#include "libtorrent/span.hpp"
+#include "libtorrent/assert.hpp"
+
+namespace libtorrent { namespace aux {
+
+	template <typename T, typename IndexType = int>
+	struct typed_span : span<T>
```
arvidn: ah, I see. this is what you meant by "typed_span". I guess that follow the precedence of "typed_bitfield".

arvidn:
lgtm

include/libtorrent/aux_/typed_span.hpp
aldenml: yes

PR #1642, opened 2017-01-29
fixing sign-conversion warnings, part 10

aldenml:


src/disk_io_thread.cpp
```diff
@@ -2575,9 +2576,9 @@ namespace libtorrent
 			? cached_piece_info::volatile_read_cache
 			: cached_piece_info::read_cache;
 		int blocks_in_piece = i->blocks_in_piece;
-		info.blocks.resize(blocks_in_piece);
+		info.blocks.resize(aux::numeric_cast<std::size_t>(blocks_in_piece));
 		for (int b = 0; b < blocks_in_piece; ++b)
-			info.blocks[b] = i->blocks[b].buf != nullptr;
+			info.blocks[std::size_t(b)] = i->blocks[b].buf != nullptr;
```
arvidn: could ``info.blocks`` be an ``aux::unique_ptr<>``?

aldenml: I don't think so, because it's part of public API (in deprecated mode) in `session_handle`
```cpp
TORRENT_DEPRECATED
void get_cache_info(sha1_hash const& ih, std::vector<cached_piece_info>& ret) const;
```

PR #1643, opened 2017-01-29
test fail related to bt_peer_connection::on_suggest_piece

aldenml:
Hi @arvidn, this simple numeric cast makes `[test_fast_extension.cpp.invalid_suggest]` fail. I'm not very familiar with this part of the code, what do you think it could be?

### DIFF
--- a/src/bt_peer_connection.cpp
+++ b/src/bt_peer_connection.cpp
@@ -1335,7 +1335,7 @@ namespace libtorrent
 		span<char const> recv_buffer = m_recv_buffer.get();
 
 		const char* ptr = recv_buffer.begin() + 1;
-		piece_index_t const piece(detail::read_uint32(ptr));
+		piece_index_t const piece(aux::numeric_cast<int>(detail::read_uint32(ptr)));
 		incoming_suggest(piece);
 	}
 


aldenml:
Well, if this is a valid situation (and following the code it appears to be), then the explicit cast like `static_cast<int>` should be valid here (maybe with a comment)?

src/bt_peer_connection.cpp
```diff
@@ -1335,7 +1335,7 @@ namespace libtorrent
 		span<char const> recv_buffer = m_recv_buffer.get();
 
 		const char* ptr = recv_buffer.begin() + 1;
-		piece_index_t const piece(detail::read_uint32(ptr));
+		piece_index_t const piece(aux::numeric_cast<int>(detail::read_uint32(ptr)));
```
arvidn: I'm pretty sure there's a run-time check in ``incoming_suggest()`` to validate the piece index. the test that's failing, most likely sends a piece index > ``INT_MAX``. In this case, it's first read as unsigned, then cast to signed, then checked. your ``numeric_cast<>`` here introduces an assert in front of the run-time validation, and hence the test for it will trigger your assert instead of exercising the validation.

arvidn: I think the proper fix here is to use ``detail::read_int32(ptr)`` instead, without a cast

aldenml: ok, I will do that

aldenml: actually, I will close this and add the change in my other put PR

PR #1645, opened 2017-01-29
fix issue where num_seeds could be greater than num_peers in torrent_status

arvidn:


src/torrent.cpp
```diff
@@ -12253,7 +12269,7 @@ namespace libtorrent
 		TORRENT_ASSERT(is_single_thread());
 		INVARIANT_CHECK;
 
-		return (std::max)(0, int(m_connections.size()) - m_num_seeds - m_num_connecting);
```
arvidn: this needs to deduct the number of connecting *non-seed* peers as well:

return std::max(0, int(m_connections.size()) - m_num_seeds - m_num_connecting + m_num_connecting_seeds);

arvidn: and perhaps that ``std::max`` statement can be removed now, replaced by an assert

PR #1647, opened 2017-01-30
Add missing announce_entry properties to python bindings 1.1

cas--:
There is no documentation on which of the announce entries are rw and which are ro so I made a guess with the new properties...

Also I was not sure about how useful `announce_entry::failed()` was, it seems more of an internal method? I had problems compiling that addition anyway.


bindings/python/src/torrent_info.cpp
```diff
@@ -307,6 +307,14 @@ void bind_torrent_info()
 
     class_<announce_entry>("announce_entry", init<std::string const&>())
         .def_readwrite("url", &announce_entry::url)
+        .def_readonly("trackerid", &announce_entry::trackerid)
+        .def_readonly("message", &announce_entry::message)
+        .def_readonly("last_error", &announce_entry::last_error)
+        .def_readonly("next_announce", &announce_entry::next_announce)
+        .def_readwrite("min_announce", &announce_entry::min_announce)
```
arvidn: they should all be read-only. iirc, this is the response from the tracker, saying you should not re-announce before this time

bindings/python/src/torrent_info.cpp
```diff
@@ -307,6 +307,14 @@ void bind_torrent_info()
 
     class_<announce_entry>("announce_entry", init<std::string const&>())
         .def_readwrite("url", &announce_entry::url)
```
arvidn: interesting that this is read_write. maybe in 1.1 this class can be used to add trackers as well, is that the case?

bindings/python/src/torrent_info.cpp
```diff
@@ -307,6 +307,14 @@ void bind_torrent_info()
...
+        .def_readonly("trackerid", &announce_entry::trackerid)
+        .def_readonly("message", &announce_entry::message)
+        .def_readonly("last_error", &announce_entry::last_error)
```
arvidn: there's a risk that non-trivial types must be bound with ``add_property`` instead, to have converters applied

arvidn:
can you think of some simple way of making sure these fields are accessible, in test.py?

bindings/python/src/torrent_info.cpp
cas--: Most are `int` or `bool` so I wouldn't imagine there is an issue? Though `next_announce` and `min_announce` are `lt::time_point`.

cas--:
Ok I misunderstood the way the exposed class works, I shall refactor it and add a test :)

cas--:
So in python we do use `lt.announce_entry` class to create an object and populate it with trackers. This means that `url`, `tier` and `fail_limit` are able to be set. However there is also an alternative in `torrent_handle` binding that takes a dict and converts it to announce_entry:

```
   ae.url = extract<std::string>(d["url"]);
   if (d.has_key("tier"))
      ae.tier = extract<int>(d["tier"]);
   if (d.has_key("fail_limit"))
      ae.fail_limit = extract<int>(d["fail_limit"]);
   if (d.has_key("source"))
      ae.source = extract<int>(d["source"]);
   if (d.has_key("verified"))
      ae.verified = extract<bool>(d["verified"]);
   if (d.has_key("send_stats"))
      ae.send_stats = extract<bool>(d["send_stats"]);
```

I don't think `verified` should be part of this code as this is a setter not a getter. However as I mentioned before there is nothing in the documentation or code to indicate what is and what isn't valid to be set for a c++ announce_entry. Only tier is mentioned in `replace_trackers` but the above list minus validate seems correct.

cas--:
Just thought, I'm not sure if you use the `--fixup` option in git but I use it a lot in branches before merging so can easily track the changes, especially in PRs. When you are happy with this PR I shall rebase them out.

cas--:
@arvidn I need a little help creating a convertor from `time_point` to python for next_announce and min_announce.

bindings/python/src/torrent_handle.cpp
```diff
@@ -191,10 +191,6 @@ void dict_to_announce_entry(dict d, announce_entry& ae)
       ae.tier = extract<int>(d["tier"]);
    if (d.has_key("fail_limit"))
       ae.fail_limit = extract<int>(d["fail_limit"]);
-   if (d.has_key("source"))
-      ae.source = extract<int>(d["source"]);
```
arvidn: if this is in fact used for adding new trackers, the "source" fields needs to be forwarded

bindings/python/src/torrent_info.cpp
```diff
@@ -126,17 +126,23 @@ namespace
        return result;
     }
 
-    int get_tier(announce_entry const& ae) { return ae.tier; }
-    void set_tier(announce_entry& ae, int v) { ae.tier = v; }
-    int get_fail_limit(announce_entry const& ae) { return ae.fail_limit; }
-    void set_fail_limit(announce_entry& ae, int l) { ae.fail_limit = l; }
+    // Create getters for announce_entry data members with non-trivial types which need converting.
+    libtorrent::time_point get_next_announce(announce_entry const& ae) { return ae.next_announce; }
```
arvidn: you don't need to define these, you can use ``make_getter`` from boost.python

bindings/python/src/torrent_info.cpp
```diff
@@ -126,17 +126,23 @@ namespace
...
+    libtorrent::time_point get_min_announce(announce_entry const& ae) { return ae.min_announce; }
+    // announce_entry data member bit-fields.
     int get_fails(announce_entry const& ae) { return ae.fails; }
     int get_source(announce_entry const& ae) { return ae.source; }
     bool get_verified(announce_entry const& ae) { return ae.verified; }
     bool get_updating(announce_entry const& ae) { return ae.updating; }
     bool get_start_sent(announce_entry const& ae) { return ae.start_sent; }
     bool get_complete_sent(announce_entry const& ae) { return ae.complete_sent; }
     bool get_send_stats(announce_entry const& ae) { return ae.send_stats; }
+    void set_send_stats(announce_entry& ae, bool b) { ae.send_stats = b; }
```
arvidn: I don't believe these were ever meant to be accessible from the client and I believe I have removed them in master

bindings/python/src/torrent_handle.cpp
cas--: it doesn't make sense for the user to set source?

arvidn: I see, I suppose the source is always set to ``client`` when added like this

arvidn:
I think this is good. you don't have any more changes in the works, do you?

cas--:
I was in the middle of creating a test but need help creating a convertor from time_point to python for next_announce and min_announce. They are currently broken.

arvidn:
Ah, right. Maybe you have an opinion about how to convert this. ``time_point`` represents an absolute point in time. However, the *epoch* (the reference point) is not statically defined. It's whatever ``libtorrent::clock_type`` says it is. libtorrent uses ``high_resolution_clock``, which basically is like ``QueryPerformanceCounter()``, the epoch is boot-time.

The way this is used in c++ is to compare these points with the current time, or some previous current time that's been saved in a time point variable. Which means calling ``libtorrent::clock_type::now()`` to get the current time.

I don't believe there's a well defined and/or reasonable way to find the epoch relative to wall clock time in order to convert it to a ``time_t`` or something like that.

So, perhaps it comes down to returning an integer, representing seconds, along with a counterpart to ``libtorrent::clock_type::now()``, which returns the current time in the same unit.

arvidn:
I suppose they could be exposed as relative times too. i.e. how long until the next announce

arvidn:
actually, there is a ``time_point`` converter in master. in RC_1_1 it would have to use ``boost::chrono::system_clock`` instead of ``std::chrono::system_clock`` though.

https://github.com/arvidn/libtorrent/blob/master/bindings/python/src/datetime.cpp#L61

arvidn:
hm.. I'm not sure the converter in master is quite correct though. It doesn't specify the timezone to be UTC. I'm not sure if that's assumed or not.

arvidn:
ok, the converter in master should be good. I just fixed it.

cas--:
I have backported it successfully but the value for `lt.announce_entry.next_announce` doesn't look quite correct: `1724-10-26 22:59:37`

arvidn:
would you mind pushing the patch? and I'll take a look. It seems like there's some system clock epoch mixup somewhere. perhaps the posix time (with epoch january 1st 1970) is applied to window's native epoch of 1600 something

cas--:
This is on linux... I have rebased the branch

I am also uncertain how to use `make_getter` with data-member bit-field. I'm clearly too spoiled by Python :P

arvidn:
ah, yeah. you can't use ``make_getter`` for bitfields. You can't take a reference to a bit, so those have to be custom functions.

bindings/python/src/datetime.cpp
```diff
@@ -55,6 +55,29 @@ struct time_duration_to_python
...
+    static PyObject* convert(lt::time_point pt)
+    {
+        using boost::chrono::system_clock;
+        using boost::chrono::duration_cast;
+        time_t const tm = system_clock::to_time_t(system_clock::now()
+            + duration_cast<system_clock::duration>(pt - lt::clock_type::now()));
+
+        std::tm* date = std::localtime(&tm);
+        object result = datetime_datetime(
+            (int)1900 + date->tm_year
```
arvidn: if you get a year in 1700-something, date->tm_year must be negative... hm.

bindings/python/src/datetime.cpp
```diff
@@ -55,6 +55,29 @@ struct time_duration_to_python
...
 };
 
+struct time_point_to_python
+{
+    static PyObject* convert(lt::time_point pt)
+    {
+        using boost::chrono::system_clock;
+        using boost::chrono::duration_cast;
+        time_t const tm = system_clock::to_time_t(system_clock::now()
+            + duration_cast<system_clock::duration>(pt - lt::clock_type::now()));
```
arvidn: I can't see anything wrong with this. maybe looking at ``duration_cast<lt::seconds>(pt - lt::clock_type::now()).count()`` might be useful. that should be some relatively small number of seconds (likely less then 1800 at least). If that works, maybe it would make sense to cut out ``system_clock`` and just do:

``time_t const tm = time(NULL) + duration_cast<lt::seconds>(pt - lt::clock_type::now()).count()``

bindings/python/src/datetime.cpp
```diff
@@ -55,6 +55,29 @@ struct time_duration_to_python
     }
 };
 
+struct time_point_to_python
+{
+    static PyObject* convert(lt::time_point pt)
+    {
+        using boost::chrono::system_clock;
+        using boost::chrono::duration_cast;
```
arvidn: my guess of why msvc is failing here is that you may need to include the ``boost/chrono.hpp`` header

arvidn: https://ci.appveyor.com/project/arvidn/libtorrent/build/1.0.5389/job/8lhtyxqkrxpdgm3x#L881

arvidn:
trying this patch out on mscOS, I can't even compile. It seems like boost uses ``std::chrono`` to some degree when available, but it's still failing with saying the boost::chrono::system_clock::duration is not a duration type. The patch I suggested (using ``time(NULL)``) and cast to ``boost::chrono::seconds`` and add them appears to work around that immediate problem.

However, there are still some python unit tests that fail

bindings/python/src/datetime.cpp
cas--: I added this to code:
```
std::cout << tm << "\n";
std::cout << now << "\n";
std::cout << pt << "\n";
```
resulting in:
```
10709678762
100238916604617 nanoseconds since boot
-9223372036854775808 nanoseconds since boot
```
So it's a rather large negative number?

cas--: I think a timedelta would be better than datetime for the announce value. That is what is currently used for torrent_status.next_announce.

arvidn: the reason to use absolute timestamps here (or at least in general) is to avoid having to post ``state_update_alert`` continuously. If these counters are relative, and constantly updates, a client might expect to be notified of those changes.

I'm not sure what printing ``pt`` will print. It's not really useful without also printing ``lt::clock_type::now()`` I think.

btw, do you understand why this isn't causing test failures on travis?

arvidn: oh, wait a minute. ``next_announce`` and ``min_announce`` are initialized to ``min_time()``, which means they are as far back in time as possible. these timestamps are by design.

Now, is that a good design? probably not.. I would imagine the proper solution would be to use something like ``std::optional<time_point>``. Perhaps in the python binding you could make them ``None`` if they are ``INT64_MIN``.

cas--: Oh I changed the code so `now` is `lt::clock_type::now()`. 

> counters are relative, and constantly updates, a client might expect to be notified of those changes
Ah ok that's fine, just thought it might reduce complexity :)

> btw, do you understand why this isn't causing test failures on travis?
Not looked into that, I see now the errors it's raising and then skipping the test: https://travis-ci.org/arvidn/libtorrent/jobs/198695098#L1890

I shall check and correct those errors tomorrow.

cas--: > next_announce and min_announce are initialized to min_time()

aha! I wondered that but seemed a bit odd :D

cas--:
Slightly off-topic but how hard would it be for the bindings makefile to only re-build the changed source files?

bindings/python/src/datetime.cpp
cas--: > do you understand why this isn't causing test failures on travis?

Afaik the travis builds only fail on non-zero exit code so perhaps that is not occurring?

arvidn:
I'm not sure. I would have expected that to be the case already.

cas--:
I am just about there, the test is updated and I have modified the code to account for minimum int but really need a test for when an actual announce time_point value is set...

I need some suggestion because the use of `None` for next_announce appears to have broken can_announce method:
```
======================================================================
ERROR: test_announce_entry (__main__.test_torrent_info)
  File "test.py", line 157, in test_announce_entry
    self.assertEquals(ae.can_announce(False), True)
ArgumentError: Python argument types in
    announce_entry.can_announce(announce_entry, bool)
did not match C++ signature:
    can_announce(libtorrent::announce_entry {lvalue}, boost::chrono::time_point<boost::chrono::steady_clock, boost::chrono::duration<long, boost::ratio<1l, 1000000000l> > >, bool)

```

bindings/python/src/datetime.cpp
```diff
@@ -5,6 +5,7 @@
 #include "boost_python.hpp"
 #include <boost/date_time/posix_time/posix_time_types.hpp>
 #include "optional.hpp"
+#include <boost/chrono.hpp>
```
cas--: Oh this is probably not needed now

bindings/python/src/datetime.cpp
```diff
@@ -55,6 +56,33 @@ struct time_duration_to_python
...
 };
 
+struct time_point_to_python
+{
+    static PyObject* convert(lt::time_point tpt)
+    {
+        object result;
+        if (tpt > lt::min_time()) {
+            lt::time_point now = lt::clock_type::now();
+            time_t const tm = lt::total_seconds(tpt - now);
```
arvidn: hm.. I don't see how this could be correct... you're just looking at the number of seconds from now.. std::localtime() expects the number of seconds since 00:00 january 1st, 1970

cas--: This is a refactor of your earlier suggested code, I did wonder that before but I couldn't get a valid `tpt` value so wasn't able to verify

arvidn: what I suggested was to also call ``time(NULL)`` and add the number of seconds to that. However, if the version with ``system_clock`` works, I think I would prefer that

cas--: Oh damn I forgot I removed that `time(NULL)`...

I guess the `system_clock` version was working so yeah. Is there a quick way to get a proper `next_announce` value when testing?

cas--:
I have cleaned up the branch so that leaves just the issue with `can_announce()`

arvidn:
it still looks like msvc is mixing in std::chrono with boost::chrono.

https://ci.appveyor.com/project/arvidn/libtorrent/build/1.0.5410/job/scs9ck2ktwo585ly#L881

cas--:
Looks like an issue in c++11 run too: https://travis-ci.org/arvidn/libtorrent/jobs/199318802#L1603

arvidn:
wow, that's pretty bad. I guess the last command determines the return value maybe. I should move those ``cd`` out

arvidn:
ok, I'm fixing the travis script here: https://github.com/arvidn/libtorrent/pull/1679

arvidn:
I think you should try to fully qualify ``system_clock`` and ``duration_cast`` with the ``boost::chrono`` namespace. I suspect something else somewhere is maybe ``using namespace std;`` or ``using namespace std::chrono;`` (finding those locations and eliminating them would be preferable, but that could be separate patch)

src/sha1.cpp
```diff
@@ -75,12 +75,11 @@ namespace
 	template <class BlkFun>
 	void SHA1transform(u32 state[5], u8 const buffer[64])
 	{
-		using namespace std;
-		u32 a, b, c, d, e;
+		std::u32 a, b, c, d, e;
```
arvidn: I think you meant ``std::uint32_t``

arvidn:
those instances of ``using namespace std;`` shouldn't affect the context where the python binding is built, since it's in a separate translation unit (and in function scope)

src/sha1.cpp
cas--: Actually it's a typedef at top of file `typedef boost::uint32_t u32;` so I should revert that change.

cas--:
Hmm ok, I thought I would change it as it's the only `namespace std` reference in code. I shall just delete that commit, as you say outside scope of PR :)

I shall fully qualify as you suggest with the chrono namespace.

cas--:
Oh should it perhaps only be `using namespace boost::chrono;` in datetime?

arvidn:
this looks good. I think I understand where the problem comes from. the python binding has ``using namespace libtorrent;``, and libtorrent has its own aliases for ``duration_cast`` and ``time_point`` etc. This is because boost.asio can be built both using boost.chrono and std.chrono, so libtorrent picks the appropriate one for its aliases.

If this patch doesn't work, maybe it would make more sense to use ``time(NULL) + ...`` trick to sidestep it. But I would expect ``boost::chrono::duration_cast<>`` to support ``std::duration``.

arvidn:
you could also use this same macro to determine whether to use boost or std.

https://github.com/arvidn/libtorrent/blob/RC_1_1/include/libtorrent/time.hpp#L67

arvidn:
let me know if you want me to take over. I believe this is the last thing I want in 1.1.2

cas--:
Yes that would be good as you understand the issues a bit better than I. :)

arvidn:
If only I could figure out how to pull your commits into a branch in my clone, without causing a merge... ``git pull --rebase`` rebases the commits from *my* clone onto your changes. I would like to do the opposite. (since the commits I have are public)

aldenml:
if it's not that important to keep this commits separated, I assume you can squash this PR (in a fork branch or not) and cherrypick

cas--:
If you checkout my branch and then `checkout -b <new-branch>` that creates a local branch with my commits.

cas--:
If you squash you will lose all the independent commit details...

arvidn:
@aldenml I'm mostly trying to preserve cas-- as the author, I do intend to squash too
@cas-- I need to get the commits into my clone though, right? I don't think I can push the changes to your clone. The only way I know of to get them into my clone is by following the instructions on github, which say: ``git pull https://github.com/cas--/libtorrent.git Issue/missing_announce_entry_properties``. and if I add ``--rebase`` there, it puts the existing ``RC_1_1`` commits on top of yours.

cas--:
You need to create a local branch that tracks my remote so no need to rebase. Once you have the local branch with my commits then rebase RC_1_1 to get the branch up to date.

arvidn:
but I don't know which commit your branch sits on top of right now. I would need to create by branch from that same commit to avoid a merge, I imagine

cas--:
You don't need to know those details. Just create a branch locally of mine. The rebase later on will take care of placing my commits at tip.

cas--:
```
git checkout --track -b <local branch> <remote>/<tracked branch>
```

```
git rebase origin/RC_1_1
```

arvidn:
I don't believe I can put a URL in place of <remote>.. do I add a remote first?

```
git remote add cas https://github.com/cas--/libtorrent.git
git checkout --track -b cas---Issue/missing_announce_entry_properties cas/Issue/missing_announce_entry_properties
```

fails with:
```
fatal: Cannot update paths and switch to branch 'cas---Issue/missing_announce_entry_properties' at the same time.
Did you intend to checkout 'cas/Issue/missing_announce_entry_properties' which can not be resolved as commit?
```

cas--:
Sorry yes you would need to add my repo as remote. Have you fetched the repo?

arvidn:
aha, I found the commit (from the github page of your clone) and created my branch from that. that seems to have worked

arvidn:
replaced by https://github.com/arvidn/libtorrent/pull/1685

PR #1649, opened 2017-01-31
fixing sign-conversion warnings, part 11, created aux::array and refactor

aldenml:


arvidn:
visual studio doesn't like that.. :/

aldenml:
fixed :) not having a problem with gcc

src/bt_peer_connection.cpp
```diff
@@ -2166,10 +2166,11 @@ namespace libtorrent
 		if (should_log(peer_log_alert::outgoing_message))
 		{
 			std::string bitfield_string;
-			bitfield_string.resize(num_pieces);
-			for (int k = 0; k < num_pieces; ++k)
+			std::size_t const n_pieces = aux::numeric_cast<std::size_t>(num_pieces);
+			bitfield_string.resize(n_pieces);
+			for (std::size_t k = 0; k < n_pieces; ++k)
```
arvidn: loops with unsigned counters variables make me nervous.. this is for indexing into the string?

aldenml: yes, me too, I take a close inspection when I use the `std::size_t` for loops. And yes, this for string indexing but also `n_pieces` is used in  `.resize(`

src/escape_string.cpp
```diff
@@ -336,7 +336,7 @@ namespace libtorrent
 		};
 		char const *base32_table = 0 != (flags & string::lowercase) ? base32_table_lowercase : base32_table_canonical;
 
-		static aux::array<int const, 6> input_output_mapping({{0, 2, 4, 5, 7, 8}});
+		static aux::array<int, 6> const input_output_mapping{{{0, 2, 4, 5, 7, 8}}};
```
arvidn: shouldn't the members be const too here?

aldenml: do you mean like `aux::array<int const, 6> const`?

arvidn: I suppose the whole vector is const, so it would be redundant

src/kademlia/routing_table.cpp
```diff
@@ -127,7 +127,7 @@ int routing_table::bucket_limit(int bucket) const
 {
 	if (!m_settings.extended_routing_table) return m_bucket_size;
 
-	static const aux::array<int, 4> size_exceptions({{16, 8, 4, 2}});
+	static const aux::array<int, 4> size_exceptions{{{16, 8, 4, 2}}};
```
arvidn: it's a bit unfortunate.. there's no way to make the constructor transparent? did you try ``using`` the base class' constructor?

src/settings_pack.cpp
```diff
@@ -99,7 +99,7 @@ namespace libtorrent
 
 	using aux::session_impl;
 
-	aux::array<str_setting_entry_t, settings_pack::num_string_settings> str_settings
+	aux::array<str_setting_entry_t, settings_pack::num_string_settings> const str_settings
```
arvidn: +1

src/kademlia/routing_table.cpp
aldenml: yes, I tried and it doesn't work, `std::array` has no explicit constructor and I was unable to achieve the same using `std:: initializer_list`

arvidn:
lgtm

aldenml:
good, and travis seems to be working again

aldenml:
as soon as you merge the travis related PR, I will rebase this one

arvidn:
done

PR #1651, opened 2017-02-01
fix web_seed redirects

ajax16384:
- make all redirects ephemeral (to avoid resume serialization)
- remove dup slash for redirects in proxy mode

### DIFF
--- a/simulation/test_web_seed.cpp
+++ b/simulation/test_web_seed.cpp
@@ -36,11 +36,13 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/torrent_info.hpp"
 #include "libtorrent/alert_types.hpp"
 #include "simulator/http_server.hpp"
+#include "simulator/http_proxy.hpp"
 #include "settings.hpp"
 #include "libtorrent/create_torrent.hpp"
 #include "simulator/simulator.hpp"
 #include "setup_swarm.hpp"
 #include "utils.hpp"
+#include "make_proxy_settings.hpp"
 #include "simulator/utils.hpp"
 #include <iostream>
 
@@ -358,6 +360,67 @@ TORRENT_TEST(multi_file_redirect)
 		},
 		[&fs](sim::simulation& sim, lt::session& ses)
 		{
+			// http1 is the root web server that will just redirect requests to
+			// other servers
+			sim::asio::io_service web_server1(sim, address_v4::from_string("2.2.2.2"));
+			sim::http_server http1(web_server1, 8080);
+			// redirect file 1 and file 2 to different servers
+			http1.register_redirect("/foo/1", "http://3.3.3.3:4444/bla/file1");
+			http1.register_redirect("/foo/2", "http://4.4.4.4:9999/bar/file2");
+
+			// server for file 1
+			sim::asio::io_service web_server2(sim, address_v4::from_string("3.3.3.3"));
+			sim::http_server http2(web_server2, 4444);
+			serve_content_for(http2, "/bla/file1", fs, file_index_t(0));
+
+			// server for file 2
+			sim::asio::io_service web_server3(sim, address_v4::from_string("4.4.4.4"));
+			sim::http_server http3(web_server3, 9999);
+			serve_content_for(http3, "/bar/file2", fs, file_index_t(1));
+
+			sim.run();
+		}
+	);
+
+	TEST_EQUAL(seeding, true);
+}
+// test web_seed redirect through proxy
+TORRENT_TEST(multi_file_redirect_through_proxy)
+{
+	using namespace libtorrent;
+	file_storage fs;
+	fs.add_file(combine_path("foo", "1"), 0xc000);
+	fs.add_file(combine_path("foo", "2"), 0xc030);
+	lt::add_torrent_params params = ::create_torrent(fs);
+	params.url_seeds.push_back("http://2.2.2.2:8080/");
+
+	bool seeding = false;
+
+	run_test(
+		[&params](lt::session& ses)
+		{
+			settings_pack pack;
+
+			pack.set_int(settings_pack::proxy_type, settings_pack::http);
+			pack.set_str(settings_pack::proxy_hostname, "50.50.50.50");
+			pack.set_str(settings_pack::proxy_username, "testuser");
+			pack.set_str(settings_pack::proxy_password, "testpass");
+			pack.set_int(settings_pack::proxy_port, 4445);
+			pack.set_bool(settings_pack::proxy_hostnames, true);
+			//TODO: ses.apply_settings(pack);
+
+			ses.async_add_torrent(params);
+		},
+		[&](lt::session& ses, lt::alert const* alert) {
+			if (lt::alert_cast<lt::torrent_finished_alert>(alert)) {
+				seeding = true;
+			}
+		},
+		[&fs](sim::simulation& sim, lt::session& ses)
+		{
+			sim::asio::io_service proxy_ios(sim, address_v4::from_string("50.50.50.50"));
+			sim::http_proxy http_p(proxy_ios, 4445);
+
 			// http1 is the root web server that will just redirect requests to
 			// other servers
 			sim::asio::io_service web_server1(sim, address_v4::from_string("2.2.2.2"));
--- a/simulation/test_web_seed.cpp
+++ b/simulation/test_web_seed.cpp
@@ -409,7 +409,7 @@ TORRENT_TEST(multi_file_redirect_through_proxy)
 			pack.set_str(settings_pack::proxy_password, "testpass");
 			pack.set_int(settings_pack::proxy_port, 4445);
 			pack.set_bool(settings_pack::proxy_hostnames, true);
-			//TODO: ses.apply_settings(pack);
+			ses.apply_settings(pack);
 
 			ses.async_add_torrent(params);
 		},
--- a/simulation/test_web_seed.cpp
+++ b/simulation/test_web_seed.cpp
@@ -448,56 +448,6 @@ TORRENT_TEST(multi_file_redirect_through_proxy)
 	TEST_EQUAL(seeding, true);
 }
 
-// Test location name prediction with unaligned files
-// Without this prediction edge piece of "foo/file1" will be excluded
-// from further original server request by range pick
-// logic of web_peer_connection::on_connected().
-TORRENT_TEST(predict_unaligned_redirect)
-{
-	using namespace libtorrent;
-	file_storage fs;
-	fs.add_file(combine_path("foo", "file1"), piece_size / 2); // piece 0
-	fs.add_file(combine_path("foo", "file2"), piece_size);     // piece 0 & 1
-	fs.add_file(combine_path("foo", "file3"), 0x1000);         // piece 1
-	lt::add_torrent_params params = ::create_torrent(fs);
-	params.url_seeds.push_back("http://2.2.2.2:8080/");
-
-	bool seeding = false;
-
-	run_test(
-		[&params](lt::session& ses)
-		{
-			ses.async_add_torrent(params);
-		},
-		[&](lt::session& ses, lt::alert const* alert) {
-			if (lt::alert_cast<lt::torrent_finished_alert>(alert))
-				seeding = true;
-		},
-		[&fs](sim::simulation& sim, lt::session& ses)
-		{
-			// http1 is the root web server that will just redirect requests to
-			// other servers
-			sim::asio::io_service web_server1(sim, address_v4::from_string("2.2.2.2"));
-			sim::http_server http1(web_server1, 8080);
-			// redirect all files to the same server
-			http1.register_redirect("/foo/file1", "http://3.3.3.3:4444/new/foo/file1");
-			http1.register_redirect("/foo/file2", "http://3.3.3.3:4444/new/foo/file2");
-			http1.register_redirect("/foo/file3", "http://3.3.3.3:4444/new/foo/file3");
-
-			// server for serving the content
-			sim::asio::io_service web_server2(sim, address_v4::from_string("3.3.3.3"));
-			sim::http_server http2(web_server2, 4444);
-			serve_content_for(http2, "/new/foo/file1", fs, file_index_t(0));
-			serve_content_for(http2, "/new/foo/file2", fs, file_index_t(1));
-			serve_content_for(http2, "/new/foo/file3", fs, file_index_t(2));
-
-			sim.run();
-		}
-	);
-
-	TEST_EQUAL(seeding, true);
-}
-
 // this is expected to fail, since the files are not aligned and redirected to
 // separate servers, without pad files
 TORRENT_TEST(multi_file_unaligned_redirect)
--- a/simulation/test_web_seed.cpp
+++ b/simulation/test_web_seed.cpp
@@ -224,7 +224,7 @@ std::string generate_content(lt::file_storage const& fs, file_index_t file
 	, std::int64_t offset, std::int64_t len)
 {
 	std::string ret;
-	ret.reserve(len);
+	ret.reserve(lt::aux::numeric_cast<std::size_t>(len));
 	std::int64_t const file_offset = fs.file_offset(file);
 	int const piece_size = fs.piece_length();
 	for (std::int64_t i = offset + file_offset; i < offset + file_offset + len; ++i)
--- a/simulation/test_web_seed.cpp
+++ b/simulation/test_web_seed.cpp
@@ -51,9 +51,10 @@ using namespace libtorrent;
 
 namespace lt = libtorrent;
 
+int const piece_size = 0x4000;
+
 add_torrent_params create_torrent(file_storage& fs, bool const pad_files = false)
 {
-	int const piece_size = 0x4000;
 	libtorrent::create_torrent t(fs, piece_size
 		, pad_files ? piece_size : -1
 		, pad_files ? create_torrent::optimize_alignment : 0);
@@ -384,6 +385,7 @@ TORRENT_TEST(multi_file_redirect)
 
 	TEST_EQUAL(seeding, true);
 }
+
 // test web_seed redirect through proxy
 TORRENT_TEST(multi_file_redirect_through_proxy)
 {
@@ -438,6 +440,56 @@ TORRENT_TEST(multi_file_redirect_through_proxy)
 			sim::asio::io_service web_server3(sim, address_v4::from_string("4.4.4.4"));
 			sim::http_server http3(web_server3, 9999);
 			serve_content_for(http3, "/bar/file2", fs, file_index_t(1));
+
+			sim.run();
+		}
+	);
+
+	TEST_EQUAL(seeding, true);
+}
+
+// Test location name prediction with unaligned files
+// Without this prediction edge piece of "foo/file1" will be excluded
+// from further original server request by range pick
+// logic of web_peer_connection::on_connected().
+TORRENT_TEST(predict_unaligned_redirect)
+{
+	using namespace libtorrent;
+	file_storage fs;
+	fs.add_file(combine_path("foo", "file1"), piece_size / 2); // piece 0
+	fs.add_file(combine_path("foo", "file2"), piece_size);     // piece 0 & 1
+	fs.add_file(combine_path("foo", "file3"), 0x1000);         // piece 1
+	lt::add_torrent_params params = ::create_torrent(fs);
+	params.url_seeds.push_back("http://2.2.2.2:8080/");
+
+	bool seeding = false;
+
+	run_test(
+		[&params](lt::session& ses)
+		{
+			ses.async_add_torrent(params);
+		},
+		[&](lt::session& ses, lt::alert const* alert) {
+			if (lt::alert_cast<lt::torrent_finished_alert>(alert))
+				seeding = true;
+		},
+		[&fs](sim::simulation& sim, lt::session& ses)
+		{
+			// http1 is the root web server that will just redirect requests to
+			// other servers
+			sim::asio::io_service web_server1(sim, address_v4::from_string("2.2.2.2"));
+			sim::http_server http1(web_server1, 8080);
+			// redirect all files to the same server
+			http1.register_redirect("/foo/file1", "http://3.3.3.3:4444/new/foo/file1");
+			http1.register_redirect("/foo/file2", "http://3.3.3.3:4444/new/foo/file2");
+			http1.register_redirect("/foo/file3", "http://3.3.3.3:4444/new/foo/file3");
+
+			// server for serving the content
+			sim::asio::io_service web_server2(sim, address_v4::from_string("3.3.3.3"));
+			sim::http_server http2(web_server2, 4444);
+			serve_content_for(http2, "/new/foo/file1", fs, file_index_t(0));
+			serve_content_for(http2, "/new/foo/file2", fs, file_index_t(1));
+			serve_content_for(http2, "/new/foo/file3", fs, file_index_t(2));
 
 			sim.run();
 		}
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -375,7 +375,16 @@ namespace libtorrent
 		{
 			std::string sep_path { path };
 			std::replace(sep_path.begin(), sep_path.end(), '/', '\\');
-			prepared_path = "\\\\?\\" + (is_complete(sep_path) ? sep_path : combine_path(current_working_directory(), sep_path));
+			if (is_complete(sep_path))
+				prepared_path = "\\\\?\\" + sep_path;
+			else
+			{
+				std::string cwd = current_working_directory();
+				// cwd may have already contain UNC path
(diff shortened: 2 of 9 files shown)


src/web_peer_connection.cpp
```diff
@@ -411,7 +411,16 @@ void web_peer_connection::write_request(peer_request const& r)
 			auto redirection = m_web->redirects.find(f.file_index);
 			if (redirection != m_web->redirects.end())
 			{
-				request += redirection->second;
+				auto const& redirect = redirection->second;
+				if (using_proxy && !redirect.empty() && (redirect[0] == '/'))
```
arvidn: a more concise way of expressing this would be:

```
bool const trailing_slash = using_proxy && !redirect.empty() && redirect[0] == '/';
request.append(redirect, trailing_slash, std::string::npos);
```

arvidn: why is ``using_proxy`` a condition here btw?

src/http_seed_connection.cpp
```diff
@@ -315,7 +315,7 @@ namespace libtorrent
 					}
 
 					// add the redirected url and remove the current one
-					t->add_web_seed(location, web_seed_entry::http_seed);
+					t->add_web_seed(location, web_seed_entry::http_seed, std::string(), web_seed_entry::headers_t(), true);
```
arvidn: I take it that last ``true`` means ephemeral, right? why do you need that here?

arvidn: iirc, it just means it won't be saved in the resume data, but you still remove the original web seed URL below, won't that be a problem?

arvidn:
could you think of a way to test this patch? I believe there's a simulation you could extend to make sure you don't get duplicate ``/``

src/web_peer_connection.cpp
ajax16384: "m_url" will contain "http://example.com/" and in http proxy case will be added to "request" variable. if "redirection->second" contains "/redirect_file" then "request" may contain "http://example.com//redirect_file".

src/http_seed_connection.cpp
ajax16384: redirected web_seed should not be stored in resume because it is only serialize url part (without "redirects" map") e.g original url points to "http://example1.com/file1", web server response with location "http://example2.com/subpath/file2". (m_url=="http://example2.com/", redirects[0]=="/subpath/file2"). if we try to load old save resume then library tryies to request wrong path "http://example2.com/file1" ("redirects" map is lost).  alternative approach is to serialize "redirects" maps, but it will definetly increase resume data size.

src/web_peer_connection.cpp
ajax16384: "trailing_slash" - ok

ajax16384:
- test\web_server.py:http_handler should validate income request url (dissalow duplicate slashes after "protocol://" and raise exception in case of rule violation
- test\test_web_seed_redirect.cpp should be extended with web_seed_redirect_through_proxy

src/web_peer_connection.cpp
arvidn: but even if ``using_proxy`` is false, concatenating those two paths may result in a double slash, no?

src/http_seed_connection.cpp
arvidn: what do you mean by that? done in serial?

src/web_peer_connection.cpp
```diff
@@ -640,7 +643,7 @@ void web_peer_connection::handle_redirect(int const bytes_left)
 
 		// add_web_seed won't add duplicates. If we have already added an entry
 		// with this URL, we'll get back the existing entry
-		web_seed_t* web = t->add_web_seed(redirect_base, web_seed_entry::url_seed, m_external_auth, m_extra_headers);
+		web_seed_t* web = t->add_web_seed(redirect_base, web_seed_entry::url_seed, m_external_auth, m_extra_headers, true);
```
arvidn: could you please explain why these two web seeds should be ephemeral? and confirm that all it means is that they won't be saved in the resume data. (ideally as a brief comment in the code)

arvidn:
if ``web_server.py`` checks for double slashes, how come the tests aren't failing without this patch?
In general it would be better to add tests as simulations, as the python web server (and socks proxy) isn't as reliable and I'm trying to move away from having those kinds of tests

src/web_peer_connection.cpp
ajax16384: redirected web_seed should not be stored in resume because it is only serialize url part (without "redirects" map") e.g original url points to "http://example1.com/file1", web server response with location "http://example2.com/subpath/file2". (m_url=="http://example2.com/", redirects[0]=="/subpath/file2"). if we try to load old save resume then library tryies to request wrong path "http://example2.com/file1" ("redirects" map is lost).

ajax16384: brief comment should be placed at "bool ephemeral" declaration?

ajax16384:
if web_server.py will find duplicate slash then it should throw exception that fail/abort current unit test.

arvidn:
> if web_server.py will find duplicate slash then it should throw exception that fail/abort current unit test.

Right. If it would and a test would trigger the case that causes a double slash, that would cover this bug. none of those conditions are true right now though.

> simulations: you talking about adding checks to simulation\test_web_seed.cpp ?

yes. in fact, the check is implicit, since you register handlers for specific paths. if the path is wrong your handle won't get run. so it should just be a matter of triggering the double slash in a test.

src/web_peer_connection.cpp
arvidn: I would think a brief comment where you add the new (redirected) urls would make sense

arvidn: at some point I will change that ``bool`` parameter to be a ``enum class ephemeral { persist, dont_persist };``, to make the call site readable.

ajax16384:
upon implementation of web_seed redirect through proxy simulation i have found test framework error.  look at following pseudo workflow:
````c++
create_dir('tmp123');
change_dir('tmp123');
  test_a();
    storage.add_file('foo1');
    try_to_download_generated_content_to('foo1');
    wait_till_seed_mode();
  test_b();
    storage.add_file('foo1'); // same file name as test_a()
    try_to_download_generated_content_to('foo1'); // we have already file from test_a(), so libTorrent reuse it's content without simulation network access, ANY possible network logic error checks will be passed
    wait_till_seed_mode(); 
 ````
to fix such behaviour i fixed test/main.cpp to add sub directory for every unit tests. this will help isolate unit test file access.

**Comment by ajax16384** [comment]
Next error was at libsimulator (wrong memmove usage) which was fixed at https://github.com/arvidn/libsimulator/pull/22

**Comment by ajax16384** [comment]
@arvidn now proxy simulator falls at:
Assertion failed: false, file libsimulator\src\acceptor.cpp, line 175
````c++
tcp::acceptor::incoming_packet(p) , p.type == aux::packet::payload
````
to repro just apply libsimulator commit, this commit and uncomment: https://github.com/arvidn/libtorrent/pull/1651/commits/75e239d6d539d3db0f70f6cf9c817c88f0d83902#diff-154d5e15c0604c37b9a2b4316318441bR432

File: test/main.cpp [hunk 4]
```diff
@@ -357,6 +368,17 @@ EXPORT int main(int argc, char const* argv[])
 		if (filter && tests_to_run.count(_g_unit_tests[i].name) == 0)
 			continue;
 
+		std::snprintf(dir, sizeof(dir), "sub_%i", i);
+		std::string unit_dir = combine_path(current_working_directory(), dir);
```
arvidn: this looks questionable. you overwrite "dir" here (as far as I can tell) with the per-test sub directory, the you add that to the current working directory, then you set the new current working directory to that.

I just notice that you "cd .." at the end of the test. What would you think about saving the root test directory you generate above, and reuse that to append the per-test subdirectory to. that way each way through the loop you would just cd straight to the new subdirectory.

simulation/test_web_seed.cpp
```diff
@@ -404,6 +406,67 @@ TORRENT_TEST(multi_file_redirect)
...
+		{
+			settings_pack pack;
+
+			pack.set_int(settings_pack::proxy_type, settings_pack::http);
+			pack.set_str(settings_pack::proxy_hostname, "50.50.50.50");
+			pack.set_str(settings_pack::proxy_username, "testuser");
+			pack.set_str(settings_pack::proxy_password, "testpass");
+			pack.set_int(settings_pack::proxy_port, 4445);
+			pack.set_bool(settings_pack::proxy_hostnames, true);
+			//TODO: ses.apply_settings(pack);
```
arvidn: thanks! I hope to get a chance to check this out this weekend

test/main.cpp
arvidn: in fact, since this is in master (and you have C++11) you could just do:

auto unit_dir = combine_path(dir, "sub_" + std::to_string(i));

arvidn:
did you see this btw? https://travis-ci.org/arvidn/libtorrent/jobs/198212458#L1706

test/main.cpp
ajax16384: ok

ajax16384:
yes, will fix in next commit

src/file.cpp
```diff
@@ -503,6 +503,29 @@ namespace libtorrent
 #endif
 	}
 
+	void change_directory(std::string const& f, error_code& ec)
```
arvidn: I would actually prefer if this function stays in the test util library. libtorrent should never update current working directory. But it does make sense for the tests

test/main.cpp
```diff
@@ -368,16 +338,21 @@ EXPORT int main(int argc, char const* argv[])
...
-			std::printf("Failed to create unit directory: %s\n", ec.message().c_str());
+			std::printf("Failed to create unit test directory: %s\n", ec.message().c_str());
 			return 1;
 		}
-		if (!change_dir(dir)) return 1; // every unit test must have separate sandbox
+		change_directory(unit_dir, ec);
+		if (ec)
+		{
+			std::printf("Failed to change unit test directory: %s\n", ec.message().c_str());
+			return 1;
```
arvidn: could this (and the return above) be a ``continue`` instead?

test/main.cpp
```diff
@@ -368,16 +338,21 @@ EXPORT int main(int argc, char const* argv[])
...
+		}
+		unit_dirs.push_back(unit_dir);
```
arvidn: the directory should be put in here immediately after it's been (successfully) created. imagine you can create the directory, but fail to cd into it. you still want to delete it later.

In fact, I'm kind of thinking that it would be more robust to delete the test directory inside the loop, right after each individual test passes. Would that make sense you think? Then you wouldn't need to store them in a vector.

In addition, the way I think they should be deleted is by having a scope guard delete it (so it happens regardless of early exits, continues and exceptions). Something like this:
``struct dir_guard { std::string dir; ~dir_guard() { error_code ec; remove_all(dir, ec); } } guard = { unit_dir };``

arvidn: I'd be happy to make this change after yours is in though. because I also don't understand why the directories are only cleaned up if ``TORRENT_LOGGING`` is not defined. that doesn't seem to make any sense

ajax16384: ok

src/file.cpp
ajax16384: ok

test/main.cpp
ajax16384: i think it is unrecoverable fail state of test framework (some base functionality that must prepare environment for unit test). if it cannot change dir then there is no sense to continue.

ajax16384:
@arvidn if you decrease file sizes of old "unaligned_file_redirect" test so both files will fall in one piece, then it will fail.

test/main.cpp
arvidn: true, it's critical the process exits with an error code if this happens

src/web_peer_connection.cpp
```diff
@@ -649,14 +658,39 @@ void web_peer_connection::handle_redirect(int const bytes_left)
...
+			{
+				std::string redirect_path_prefix = redirect_path.substr(0, redirect_path_idx);
+				file_storage const& fs = info.files();
+				for (file_index_t fi(0); fi < fs.end_file(); ++fi)
+				{
+					if (info.orig_files().pad_file_at(fi))
+						continue;
+					if (web->redirects.find(fi) != web->redirects.end())
+						continue;
+					web->have_files.set_bit(fi);
```
arvidn: why are you doing this? it seems incorrect. you're assuming that all files will be redirected to this server. It seems to defeat the purpose of the ``have_files`` bitmask, which records which files have *actually* ben redirected here.

arvidn: btw, is this patch a prerequisite for your other fix? If this change is independent, I would prefer it be a separate PR

src/web_peer_connection.cpp
```diff
@@ -780,9 +814,8 @@ void web_peer_connection::on_receive(error_code const& error
 				peer_log(peer_log_alert::info, "STATUS"
 					, "%d %s", m_parser.status_code(), m_parser.message().c_str());
 				std::multimap<std::string, std::string> const& headers = m_parser.headers();
-				for (std::multimap<std::string, std::string>::const_iterator i = headers.begin()
-					, end(headers.end()); i != end; ++i)
-					peer_log(peer_log_alert::info, "STATUS", "   %s: %s", i->first.c_str(), i->second.c_str());
+				for (auto const &i : headers)
```
arvidn: +1

src/web_peer_connection.cpp
```diff
@@ -380,11 +385,8 @@ void web_peer_connection::write_request(peer_request const& r)
 		std::vector<file_slice> files = info.orig_files().map_block(req.piece, req.start
 			, req.length);
 
-		for (std::vector<file_slice>::iterator i = files.begin();
-			i != files.end(); ++i)
+		for (auto const &f : files)
```
arvidn: +1

test/main.cpp
```diff
@@ -217,6 +219,51 @@ void print_usage(char const* executable)
...
+#endif
+}
+
+struct unit_directory_guard
+{
+	std::string dir;
+	~unit_directory_guard()
+	{
+		error_code ec;
+		change_directory(root_dir, ec); // windows will not allow to remove current dir, so let's change it to root
```
arvidn: instead of making ``root_dir`` a global variable, could you just do:

``change_directory(parent_path(dir));``?

test/main.cpp
```diff
@@ -217,6 +219,51 @@ void print_usage(char const* executable)
...
+		if (ec)
+		{
+			std::printf("Failed to change directory: %s\n", ec.message().c_str());
```
arvidn: this should probably be a fatal error. You should be able to use the ``TEST_ERROR`` macro for that, I would imagine.

test/main.cpp
```diff
@@ -217,6 +219,51 @@ void print_usage(char const* executable)
...
+			return;
+		}
+		if (_g_test_failures == 0 && !keep_files)
```
arvidn: would you mind always deleting the files unless ``keep_files`` is set?

test/test.cpp
```diff
@@ -36,7 +36,8 @@ POSSIBILITY OF SUCH DAMAGE.
 
 unit_test_t _g_unit_tests[1024];
 int _g_num_unit_tests = 0;
-int _g_test_failures = 0;
+int _g_test_failures = 0; // flushed at start of every unit
+int _g_test_total_failures = 0;
```
arvidn: did this used to return incorrect return code from the test process? if the last test passed, would it shadow previous failures?
Is there a good reason to clear ``_g_test_failures`` each turn through the loop? I suppose there needs to be some way to tell whether a test had failures or not. an alternative way would be to copy this counter at the beginning of the test and then compare at the end. I think that would be a bit nicer, as it would require one less global variable.

arvidn:
btw, I merged your PR against libsimulator, so I imagine you could update the submodule reference in this patch

src/web_peer_connection.cpp
ajax16384: this is optimisitic assumption that all the rest of files will be at these new predicted locations (it's done only once if redirects[file_index] is not filled). if assumption is incorrect then have_file.clear_bit occured for this web_seed.
anyway it's the current only way to download unaligned files from plain redirect (new location for all files are equal by file_path string endings). without this prediction download simply stuck cause web_peer_connection::on_connect cut requests for edge pieces.

test/main.cpp
ajax16384: ok

test/main.cpp
ajax16384: ok

test/main.cpp
ajax16384: ok

test/test.cpp
ajax16384: you right, it shadowed real failures count by last successfull result.
test framework already had some total_failures counter so i just moved it to global scope.

src/web_peer_connection.cpp
ajax16384: with optimistic prediction web_seed requests will be issued to new location at first and only if it failures to origin server.
just as i said the real problem may be observed even at old "unaligned_file_redirect" with several files which content resides inside first piece.

ajax16384:
ok, but i still waiting for how to pass libsimulator\src\acceptor.cpp assert

test/test.cpp
ajax16384: and print_failures() is called in the end of all test and must use real total number of failures which may be located only at it's file definition (test/test.cpp)

arvidn:
oh, right

arvidn:
I believe the problem with the simulation is that the http proxy only supports a single request at a time. It was made this way to keep it simple, and it covered all the use cases at the time.

This test specifically want to redirect files to separate servers though, right? It can't be changed to only have a single proxy connection at a time?

src/web_peer_connection.cpp
arvidn: that case is supposed to be handled by storing partial pieces in the ``web_seed_entry``. I suppose it's a problem when you combine that with redirects, since you have multiple different ``web_seed_entry`` objects. I would prefer to extend the existing solution than to do this though. It seems a bit hacky, and doesn't solve the general problem (afaict).

test/test.cpp
arvidn: actually, aren't the number of failures stored per unit test in the ``_g_unit_tests`` array?
I would expect the ``test_total_failures`` state is redundant, and could be recomputed from the test array. I haven't looked into this recently, but I suspect there's a smaller/simpler patch to solve the problem. Btw, what was the symptom you experienced?

ajax16384:
unfortunately no

test/test.cpp
ajax16384: oh. you right - it can be recalculated

ajax16384: symptom - unexpected test return code

arvidn:
I'm still working on libsimulator to get this to work

arvidn:
I've fixed libsimulator here https://github.com/arvidn/libsimulator/pull/24
You could test with that branch, but as soon as it passes CI, I will merge it into master

arvidn:
that fix is in master of libsimulator now

arvidn:
I'm tempted to pull out your fixes to the test framework and back-port them to RC_1_1.

ajax16384:
libsimulator commit fix proxy unit test, now it web_seed redirect through proxy work as intended. (with commented out dup slash checking unit test throw error)

ajax16384:
@arvidn i can backport only test framework changes by myself

arvidn:
cool. I looked into it last night, and as far as I can tell, there's nothing wrong with counting or reporting the number of errors. However, your patch to have each individual test run in its own directory would probably improve reliability of the tests

ajax16384:
https://github.com/arvidn/libtorrent/pull/1707

ajax16384:
waiting for another rebase round

arvidn:
please rebase on top of master again now

test/test.cpp
```diff
@@ -65,7 +62,7 @@ int print_failures()
 	}
 
 	std::printf("\n\n");
-	int total_num_failures = 0;
+	int total_num_failures { 0 };
```
arvidn: is there a good reason to make this change?

test/test.cpp
```diff
@@ -51,7 +49,6 @@ void report_failure(char const* err, char const* file, int line)
 	char buf[500];
 	std::snprintf(buf, sizeof(buf), "\x1b[41m***** %s:%d \"%s\" *****\x1b[0m\n", file, line, err);
 	std::fprintf(stderr, "\n%s\n", buf);
-	failure_strings.push_back(buf);
```
arvidn: I used to print all of these at the end. I take it that's not the case anymore? I think it's a pretty convenient feature, to potentially avoid digging through a bunch of log output

test/main.cpp
```diff
@@ -514,6 +507,8 @@ EXPORT int main(int argc, char const* argv[])
 	if (redirect_stdout) fflush(stdout);
 	if (redirect_stderr) fflush(stderr);
 
-	return print_failures() ? 333 : 0;
+	int total_num_failures = print_failures();
+
+	return total_num_failures ? 333 : 0;
```
arvidn: is there a good reason for this change? If so, ``total_num_failures`` should be const

test/main.cpp
```diff
@@ -361,10 +359,8 @@ EXPORT int main(int argc, char const* argv[])
 #else
 	process_id = getpid();
 #endif
-	std::string root_dir = current_working_directory();
-	char dir[40];
-	snprintf(dir, sizeof(dir), "test_tmp_%u", process_id);
-	std::string unit_dir_prefix = combine_path(root_dir, dir);
+	std::string root_dir { current_working_directory() };
```
arvidn: if anything, this should be made const. I don't see a strong reason to use uniform initialization here, instead of regular initialization.

test/main.cpp
```diff
@@ -221,20 +221,18 @@ void print_usage(char const* executable)
 void change_directory(std::string const& f, error_code& ec)
 {
 	ec.clear();
+	native_path_string n = convert_to_native_path_string(f);
```
arvidn: I believe this ``n`` can be ``const``, right?

src/web_peer_connection.cpp
arvidn: I would prefer to not do this. A better approach would be to factor out the "partial piece" from ``web_seed_entry`` into the torrent object, or something at the torrent level. keyd by the piece. That way, a partial piece downloaded from one web seed can be picked up an resume by another

simulation/test_web_seed.cpp
```diff
@@ -383,6 +386,118 @@ TORRENT_TEST(multi_file_redirect)
...
+		{
+			settings_pack pack;
+
+			pack.set_int(settings_pack::proxy_type, settings_pack::http);
+			pack.set_str(settings_pack::proxy_hostname, "50.50.50.50");
+			pack.set_str(settings_pack::proxy_username, "testuser");
+			pack.set_str(settings_pack::proxy_password, "testpass");
+			pack.set_int(settings_pack::proxy_port, 4445);
+			pack.set_bool(settings_pack::proxy_hostnames, true);
+			ses.apply_settings(pack);
```
arvidn: I see you enabled this now. has the submodule been updated to the latest version already?

src/web_peer_connection.cpp
```diff
@@ -126,6 +122,15 @@ web_peer_connection::web_peer_connection(peer_connection_args const& pack
 #endif
 }
 
+std::string web_peer_connection::escape_file_path(file_storage const& storage, file_index_t index)
```
arvidn: as far as I can tell, this does not need to be a member function. I think it should be a free function with internal linkage (anonymous namespace). and its declaration should be removed from the header

test/test.cpp
ajax16384: no, just remained from old commits, will revert to minmize patch

test/test.cpp
ajax16384: ok, will revert

test/main.cpp
ajax16384: ok

simulation/test_web_seed.cpp
ajax16384: yes

test/main.cpp
ajax16384: ok

ajax16384:
@arvidn build failed due: https://github.com/arvidn/libsimulator/pull/25

ajax16384:
if cwd returned in UNC format then many tests failed. reverting from current commit. will make separate PR after this one.

arvidn:
I'm not sure I understand the cwd point. But internally in libtorrent, I believe the assumption is that paths are utf8 in posix format. is that right?

ajax16384:
i meant case then current_working_directory() returns something like "\\?\c:\abc\def"

src/web_peer_connection.cpp
arvidn: I still consider this a blocker for this PR btw.

ajax16384: ok

arvidn:
lgtm

PR #1652, opened 2017-02-01
implemented support for BEP 51

aldenml:
For preliminary review

### DIFF
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -2604,7 +2604,7 @@ namespace libtorrent {
 
 		virtual std::string message() const override;
 
-		udp::endpoint const endpoint;
+		aux::noexcept_movable<udp::endpoint> endpoint;
 
 		time_duration const interval;
 
@@ -2616,7 +2616,13 @@ namespace libtorrent {
 		int num_samples() const;
 		std::vector<sha1_hash> samples() const;
 
+		// The total number of nodes returned by ``nodes()``.
 		int num_nodes() const;
+
+		// This is the set of more DHT nodes returned by the request.
+		//
+		// The information is included so that indexing nodes can perform a keyspace
+		// traversal with a single RPC per node by adjusting the target value for each RPC.
 		std::vector<std::pair<sha1_hash, udp::endpoint>> nodes() const;
 
 	private:
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -2590,11 +2590,50 @@ namespace libtorrent {
 		virtual std::string message() const override;
 	};
 
+	struct TORRENT_EXPORT dht_sample_infohashes_alert final : alert
+	{
+		dht_sample_infohashes_alert(aux::stack_allocator& alloc
+			, udp::endpoint const& endp
+			, time_duration interval
+			, int num
+			, std::vector<sha1_hash> const& samples
+			, std::vector<std::pair<sha1_hash, udp::endpoint>> const& nodes);
+
+		static const int static_category = alert::dht_operation_notification;
+		TORRENT_DEFINE_ALERT(dht_sample_infohashes_alert, 93)
+
+		virtual std::string message() const override;
+
+		udp::endpoint const endpoint;
+
+		time_duration const interval;
+
+		// This field indicates how many infohash keys are currently in the node's storage.
+		// If the value is larger than the number of returned samples it indicates that the
+		// indexer may obtain additional samples after waiting out the interval.
+		int const num_infohashes;
+
+		int num_samples() const;
+		std::vector<sha1_hash> samples() const;
+
+		int num_nodes() const;
+		std::vector<std::pair<sha1_hash, udp::endpoint>> nodes() const;
+
+	private:
+		std::reference_wrapper<aux::stack_allocator> m_alloc;
+		int const m_num_samples;
+		aux::allocation_slot m_samples_idx;
+		int m_v4_num_nodes = 0;
+		int m_v6_num_nodes = 0;
+		aux::allocation_slot m_v4_nodes_idx;
+		aux::allocation_slot m_v6_nodes_idx;
+	};
+
 #undef TORRENT_DEFINE_ALERT_IMPL
 #undef TORRENT_DEFINE_ALERT
 #undef TORRENT_DEFINE_ALERT_PRIO
 
-	enum { num_alert_types = 93 }; // this enum represents "max_alert_index" + 1
+	enum { num_alert_types = 94 }; // this enum represents "max_alert_index" + 1
 }
 
 #endif
--- a/include/libtorrent/session_handle.hpp
+++ b/include/libtorrent/session_handle.hpp
@@ -452,6 +452,14 @@ namespace libtorrent {
 		// posted, regardless of the alert mask.
 		void dht_live_nodes(sha1_hash const& nid);
 
+		// Query the DHT to retrieve a sample of the infohashes that the specific
+		// node currently have in their storage.
+		// The ``target`` is included for iterative lookups so that indexing nodes
+		// can perform a keyspace traversal with a single RPC per node by adjusting
+		// the target value for each RPC. It has no effect on the returned sample value.
+		// The result is posted as a ``dht_sample_infohashes_alert``.
+		void dht_sample_infohashes(udp::endpoint const& ep, sha1_hash const& target);
+
 		// Send an arbitrary DHT request directly to the specified endpoint. This
 		// function is intended for use by plugins. When a response is received
 		// or the request times out, a dht_direct_response_alert will be posted
--- a/include/libtorrent/session_handle.hpp
+++ b/include/libtorrent/session_handle.hpp
@@ -452,8 +452,8 @@ namespace libtorrent {
 		// posted, regardless of the alert mask.
 		void dht_live_nodes(sha1_hash const& nid);
 
-		// Query the DHT to retrieve a sample of the infohashes that the specific
-		// node currently have in their storage.
+		// Query the DHT node specified by ``ep`` to retrieve a sample of the
+		// infohashes that the node currently have in their storage.
 		// The ``target`` is included for iterative lookups so that indexing nodes
 		// can perform a keyspace traversal with a single RPC per node by adjusting
 		// the target value for each RPC. It has no effect on the returned sample value.
--- a/src/alert.cpp
+++ b/src/alert.cpp
@@ -2275,4 +2275,62 @@ namespace {
 		return stats_header;
 	}
 
+	dht_sample_infohashes_alert::dht_sample_infohashes_alert(aux::stack_allocator& alloc
+		, udp::endpoint const& endp
+		, time_duration _interval
+		, int _num
+		, std::vector<sha1_hash> const& samples
+		, std::vector<std::pair<sha1_hash, udp::endpoint>> const& nodes)
+		: endpoint(endp)
+		, interval(_interval)
+		, num_infohashes(_num)
+		, m_alloc(alloc)
+		, m_num_samples(aux::numeric_cast<int>(samples.size()))
+	{
+		m_samples_idx = alloc.allocate(m_num_samples * 20);
+
+		char *ptr = alloc.ptr(m_samples_idx);
+		std::memcpy(ptr, samples.data(), samples.size() * 20);
+
+		std::tie(m_v4_num_nodes, m_v4_nodes_idx, m_v6_num_nodes, m_v6_nodes_idx)
+			= write_nodes(alloc, nodes);
+	}
+
+	std::string dht_sample_infohashes_alert::message() const
+	{
+		char msg[200];
+		std::snprintf(msg, sizeof(msg)
+			, "incoming dht sample_infohashes reply from: %s, samples %d"
+			, print_endpoint(endpoint).c_str(), m_num_samples);
+		return msg;
+	}
+
+	int dht_sample_infohashes_alert::num_samples() const
+	{
+		return m_num_samples;
+	}
+
+	std::vector<sha1_hash> dht_sample_infohashes_alert::samples() const
+	{
+		aux::vector<sha1_hash> samples;
+		samples.resize(m_num_samples);
+
+		const char *ptr = m_alloc.get().ptr(m_samples_idx);
+		std::memcpy(samples.data(), ptr, samples.size() * 20);
+
+		return samples;
+	}
+
+	int dht_sample_infohashes_alert::num_nodes() const
+	{
+		return m_v4_num_nodes + m_v6_num_nodes;
+	}
+
+	std::vector<std::pair<sha1_hash, udp::endpoint>> dht_sample_infohashes_alert::nodes() const
+	{
+		return read_nodes(m_alloc.get()
+			, m_v4_num_nodes, m_v4_nodes_idx
+			, m_v6_num_nodes, m_v6_nodes_idx);
+	}
+
 } // namespace libtorrent
--- a/src/kademlia/dht_tracker.cpp
+++ b/src/kademlia/dht_tracker.cpp
@@ -322,6 +322,20 @@ namespace libtorrent { namespace dht {
 			n.second.dht.announce(ih, listen_port, flags, f);
 	}
 
+	void dht_tracker::sample_infohashes(udp::endpoint const& ep, sha1_hash const& target
+		, std::function<void(time_duration
+			, int, std::vector<sha1_hash>
+			, std::vector<std::pair<sha1_hash, udp::endpoint>>)> f)
+	{
+		for (auto& n : m_nodes)
+		{
+			if (ep.protocol() != (n.first->get_external_address().is_v4() ? udp::v4() : udp::v6()))
+				continue;
+			n.second.dht.sample_infohashes(ep, target, f);
+			break;
+		}
+	}
+
 	namespace {
 
 	struct get_immutable_item_ctx
--- a/src/kademlia/node.cpp
+++ b/src/kademlia/node.cpp
@@ -64,6 +64,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/kademlia/get_item.hpp"
 #include "libtorrent/kademlia/msg.hpp"
 #include <libtorrent/kademlia/put_data.hpp>
+#include <libtorrent/kademlia/sample_infohashes.hpp>
 
 using namespace std::placeholders;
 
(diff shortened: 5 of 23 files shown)


include/libtorrent/alert_types.hpp
```diff
@@ -2506,11 +2506,38 @@ namespace libtorrent
...
+		static const int static_category = alert::dht_operation_notification;
+		TORRENT_DEFINE_ALERT(dht_sample_infohashes_alert, 91)
+
+		virtual std::string message() const override;
+
+		sha1_hash const target;
+
+		time_duration const interval;
+
+		int const num;
```
arvidn: is this ``num`` different from the ``num_samples`` below?

src/kademlia/sample_infohashes.cpp
```diff
@@ -0,0 +1,151 @@
...
+		timeout();
+		return;
+	}
+
+	std::int64_t const num = r.dict_find_int_value("num", -1);
+	if (num < 0 || num > std::numeric_limits<int>::max())
+	{
+#ifndef TORRENT_DISABLE_LOGGING
+		get_observer()->log(dht_logger::traversal, "[%p] wrong or missing num value"
+			, static_cast<void*>(algorithm()));
```
arvidn: please use ``search_id`` from the traversal_algorithm for logging instead of the pointer (I changed this a while ago to make the log output deterministic across simulation runs)

include/libtorrent/session_handle.hpp
```diff
@@ -443,6 +443,14 @@ namespace libtorrent
 		void dht_get_peers(sha1_hash const& info_hash);
 		void dht_announce(sha1_hash const& info_hash, int port = 0, int flags = 0);
 
+		// Query the DHT to retrieve a sample of the infohashes that other nodes
+		// currently have in their storage.
+		// The ``target`` is included for iterative lookups so that indexing nodes
+		// can perform a keyspace traversal with a single RPC per node by adjusting
+		// the target value for each RPC. It has no effect on the returned sample value.
+		// The result is posted as a ``dht_sample_infohashes_alert``.
```
arvidn: I would expect the resulting info hashes to have an affinity to be close to ``target`` no? your comment here seem to suggest they are unrelated

arvidn:
looks reasonable. it's a big patch, so it feels like some more test coverage would be appropriate.

include/libtorrent/session_handle.hpp
ssiloti: `target` is only used to influence the `nodes` field. The set of info-hashs returned should be fixed and regenerated periodically so that repeated querying does not yield additional info-hashes.

arvidn: it affects which nodes you ask, right? and presumably nodes tend to hold info-hashes close to their own node-id

ssiloti: Hrm, that's true with the way it is implemented here. I think the original intent was to adjust the target with each query to effectively do a linear scan of the key space rather that do a typical iterative lookup.

include/libtorrent/alert_types.hpp
aldenml: yes, it's different, I will add some documentation about it

src/kademlia/sample_infohashes.cpp
aldenml: ok

include/libtorrent/session_handle.hpp
aldenml: @ssiloti I think you are right, but I would like to revisit the BEP and the implementation by @the8472 and change accordingly.

aldenml:
@arvidn I can add trivial tests for the new alert, but I appreciate if you give me some guide in order to test more the core DHT feature

the8472:
Is the question about how to respond to a single request or how a whole-keyspace-traversal should be implemented?

aldenml:
Second: 

> how a whole-keyspace-traversal should be implemented

the8472:
Ah, that's non-trivial. I tried a random approach first but that performed poorly, only visiting a small fraction of the nodes.

You can find the current implementation which manages to visit approximately all reachable nodes in the DHT at https://github.com/the8472/mldht/blob/master/src/lbms/plugins/mldht/kad/tasks/KeyspaceSampler.java

I'll do a writeup of the algorithm later today. The rough idea is that it uses a routing table with a moving home bucket, forgetting things left of the home bucket through bucket merges.

aldenml:
Do you think that using "iterative lookup" as implemented in this PR is a wrong idea?

the8472:
A targeted lookup would only give you a tiny sample of the keyspace. Which might be good enough for some usecases. But for the "index all the things" case you will need something broader.

aldenml:
thanks! I will need to review this.

@ssiloti do you think I can/should implement this in terms of "direct DHT message" and left the "key space traversal" problem to the client?

gubatron:
>I tried a random approach first but that performed poorly, only visiting a small fraction of the nodes.

@the8472 could you expand a bit more on what the desired result would be. what you mean here by poor performance and a small fraction of the nodes, does it mean that the random function you used gave you a very narrow/clustered distribution as the resulting sample obtained?

gubatron:
another simple idea on how to pick who to query:

imagine all your node-keys laid out in an X axis line.

1. calculate median distance between each node.
2. pick the center-most node-key as a candidate.
3. then pick a node from each side, jumping at least (`mean_distance * factor`) until you have the number of candidates you desire.

then adjust factor to see what the performance is

the8472:
@gubatron 

> could you expand a bit more on what the desired result would be.

ultimately the goal is to get as many infohashes as we can with as little resources (mostly traffic) as possible. since node generally host infohashes close to their node ID that means we need to traverse the whole keyspace. traffic can be reduced in two ways, on by making the traversal efficient, i.e. sending no more than 1 query per node (this needs efficient `target` selection), the second is by skipping nodes, i.e. making the traversal more sparse.


> what you mean here by poor performance and a small fraction of the nodes, does it mean that the random function you used gave you a very narrow/clustered distribution as the resulting sample obtained?

a randomized approach simply does not work well because you cannot get an unbiased list of nodes from other nodes by the nature of their routing tables. it's not a matter of the local random function, it's an issue of selecting the `target` to balance keyspace traversal with locality. nodes only have good accuracy near their own id.

ssiloti:
@aldenml I think we should leave the traversal strategy up to the client since crawlers are likely going to want specific policies around polling intervals and those may require persistent storage of ``interval`` values for individual nodes.

the8472:
Well, even if you don't seem to need it anymore, here's the promised writeup:

https://github.com/the8472/mldht/blob/master/docs/keyspace-traversal.md

arvidn:
I think what you have right now seems reasonable. you just need to invoke it multiple times with ids spread across the keyspace. doing traversal would be difficult on the client side as far as I can tell

arvidn:
actually, as @the8472's point, you don't want to traverse in the traditional way, because you'll increase the risk of getting duplicates the farther along you get.

src/alert.cpp
```diff
@@ -2105,4 +2105,39 @@ namespace libtorrent {
 		return buf;
 	}
 
+	dht_sample_infohashes_alert::dht_sample_infohashes_alert(aux::stack_allocator& alloc
```
arvidn: I think it would be good to have some test coverage of the alert

src/kademlia/node.cpp
```diff
@@ -1118,6 +1135,30 @@ void node::incoming_request(msg const& m, entry& e)
 				, reply);
 		}
 	}
+	else if (query == "sample_infohashes")
+	{
+		key_desc_t const msg_desc[] = {
+			{"target", bdecode_node::string_t, 20, 0},
+			{"want", bdecode_node::list_t, 0, key_desc_t::optional},
+		};
```
arvidn: can you think of a way to test this without having to set up a full simulation of the DHT? I suppose you could have the test send this message directly to the dht node

src/alert.cpp
aldenml: ok

src/kademlia/node.cpp
aldenml: I believe that's the one (and only one) test I added here. Could you check at it to see if I'm wrong?

arvidn: I guess the code coverage report does not include the simulations.. that's why this appears to not be covered

aldenml:
for review again, thanks

the8472:
I don't see anything relating to processing the nodes/nodes6 response fields. A traversal implementation would want to use those, including from responses that don't contain the required fields.
Is that handled by some common logic?

aldenml:
@the8472 thanks for pointing this out, there is a common logic for this in `find_data_observer::reply(m);`. I removed the call when implementing this without "actual traversal", I will re-add it, but first, I want to factor out some changes that not belong to this PR for easy the review.

include/libtorrent/session_handle.hpp
```diff
@@ -449,6 +449,14 @@ namespace libtorrent
 		// posted, regardless of the alert mask.
 		void dht_live_nodes(sha1_hash const& nid);
 
+		// Query the DHT to retrieve a sample of the infohashes that the specific
+		// node currently have in their storage.
+		// The ``target`` is included for iterative lookups so that indexing nodes
+		// can perform a keyspace traversal with a single RPC per node by adjusting
+		// the target value for each RPC. It has no effect on the returned sample value.
+		// The result is posted as a ``dht_sample_infohashes_alert``.
+		void dht_sample_infohashes(udp::endpoint const& ep, sha1_hash const& target);
```
arvidn: sorry for the delay. what did we conclude here? Without having been too involved in how this traversal should work, my guess (from looking at this API and documentation) would be that I am (as a client) supposed to call this function repeatedly, adjusting the ``target`` argument to be evenly distributed across the key space.

The conflict with that notion is that you say "It has no effect on the returned sample value.". If that is indeed true, why do I need to pass it in here? Why wouldn't libtorrent randomize it (or whatever else it could do to determine it)?

Furthermore, thinking about how I imagine this to be implemented is that instead of narrowing in on a certain info-hash when traversing the DHT, I would expect every lookup to adjust the target info-hash some way, to try to spread out the nodes being visited. In either case, I don't understand why I'm passing in ``target`` here.

aldenml: If I understand correctly, while adjusting the `target`, nodes will return the field `nodes/nodes6` populated and the DHT will include those nodes. If after that, you use `dht_live_nodes` I expect that to be enough for the entire feature of "DHT indexing". Well, it could be a little convoluted, but the constraint is expose as little as possible in the public API.

`target` has no influence in `dht_sample_infohashes_alert::samples()`, the data returned by the nodes.

If it feels like I'm rephrasing the same, let me know :)

ssiloti: Indexing the DHT will involve traversing many times more nodes than libtorrent keeps in its routing table. The client would only use `dht_live_nodes` to get the initial batch of nodes to query. The list of `nodes` returned for the query needs to be included in `dht_sample_infohashes_alert` so that the client can implement its own traversal algorithm.

aldenml: OK, then, if I include that "nodes" list in the alert, do I let the DHT process "nodes/nodes6" internally, or not?

ssiloti: Yes, you should still go through the usual `nodes` processing.

aldenml: interesting...I thought that calling `dht_live_nodes` successively would be the same at the end, but I assume that returning "nodes" in the alert is more direct and clear to use. Is my reasoning correct?

ssiloti: Not all nodes will be added to the routing table. In fact most of the time few if any will be because most buckets will be full.

aldenml: Thanks for the clarification!

aldenml:
@arvidn I need to clarify a few things before continue this. For reference: https://github.com/arvidn/libtorrent/issues/1810

arvidn:
@aldenml I'm starting to think about a 1.2 release. Is this patch good to go?

aldenml:
I believe it is, I made the changes following the last review

include/libtorrent/session_handle.hpp
```diff
@@ -456,6 +456,14 @@ namespace libtorrent {
 		// posted, regardless of the alert mask.
 		void dht_live_nodes(sha1_hash const& nid);
 
+		// Query the DHT to retrieve a sample of the infohashes that the specific
+		// node currently have in their storage.
```
arvidn: "the specific node", I take it this should say something like: "the node specified by ``ep``". Is that right, or am I misunderstanding this call?

include/libtorrent/alert_types.hpp
```diff
@@ -2569,11 +2569,50 @@ namespace libtorrent {
...
+		// This field indicates how many infohash keys are currently in the node's storage.
+		// If the value is larger than the number of returned samples it indicates that the
+		// indexer may obtain additional samples after waiting out the interval.
+		int const num_infohashes;
+
+		int num_samples() const;
+		std::vector<sha1_hash> samples() const;
+
+		int num_nodes() const;
+		std::vector<std::pair<sha1_hash, udp::endpoint>> nodes() const;
```
arvidn: would you mind documenting the ``nodes()`` function too? As far as I understand, this returns the *node-id* and endpoints for more DHT nodes returned by the request.

include/libtorrent/session_handle.hpp
arvidn: actually, this call will only send a single request to a single node, right? Perhaps the wording here can be clearer to say "Query the DHT node specified by ``ep``". This isn't querying the DHT as a whole, right? Since there's no traversal or any use of our routing table. (assuming I understand this correctly)

src/kademlia/dht_tracker.cpp
```diff
@@ -316,6 +316,20 @@ namespace libtorrent { namespace dht {
...
+	void dht_tracker::sample_infohashes(udp::endpoint const& ep, sha1_hash const& target
+		, std::function<void(time_duration
+			, int, std::vector<sha1_hash>
+			, std::vector<std::pair<sha1_hash, udp::endpoint>>)> f)
+	{
+		for (auto& n : m_nodes)
+		{
+			if (ep.protocol() != (n.first->get_external_address().is_v4() ? udp::v4() : udp::v6()))
+				continue;
+			n.second.dht.sample_infohashes(ep, target, f);
```
arvidn: I guess it sends (at most) one request per DHT node we run locally

src/kademlia/dht_tracker.cpp
```diff
@@ -316,6 +316,20 @@ namespace libtorrent { namespace dht {
 			n.second.dht.announce(ih, listen_port, flags, f);
 	}
 
+	void dht_tracker::sample_infohashes(udp::endpoint const& ep, sha1_hash const& target
+		, std::function<void(time_duration
+			, int, std::vector<sha1_hash>
+			, std::vector<std::pair<sha1_hash, udp::endpoint>>)> f)
```
arvidn: I feel like this function taking another function may warrant a comment to how it is used

src/kademlia/node.cpp
```diff
@@ -565,6 +566,39 @@ void node::put_item(public_key const& pk, std::string const& salt
...
+	if (!o) return;
+#if TORRENT_USE_ASSERTS
+	o->m_in_constructor = false;
+#endif
+
+	entry e;
+	entry& a = e["a"];
+
+	e["q"] = "sample_infohashes";
+	a["target"] = target.to_string();
```
arvidn: you could simplify this a bit by saying: ``e["a"]["target"] = target.to_string()``. In fact, I don't believe you need to say ``to_string()``, since target will also fit the ``span`` overload (unless I'm mistaken), which may also avoid the heap allocation.

src/kademlia/node.cpp
```diff
@@ -1119,6 +1153,30 @@ void node::incoming_request(msg const& m, entry& e)
...
+		bdecode_node msg_keys[2];
+		if (!verify_message(arg_ent, msg_desc, msg_keys, error_string))
+		{
+			m_counters.inc_stats_counter(counters::dht_invalid_sample_infohashes);
+			incoming_error(e, error_string);
+			return;
+		}
+
+		m_counters.inc_stats_counter(counters::dht_sample_infohashes_in);
+		sha1_hash target(msg_keys[0].string_ptr());
```
arvidn: I believe you could make ``target`` const here

src/kademlia/sample_infohashes.cpp
```diff
@@ -0,0 +1,144 @@
...
+	: traversal_algorithm(dht_node, target)
+	, m_data_callback(dcallback) {}
+
+char const* sample_infohashes::name() const { return "sample_infohashes"; }
+
+void sample_infohashes::got_samples(time_duration interval
+	, int num, std::vector<sha1_hash> samples
+	, std::vector<std::pair<sha1_hash, udp::endpoint>> nodes)
+{
+	if (m_data_callback)
```
arvidn: is there ever any use case where this callback may not be set? If so, would it make sense to even issue the request?

src/kademlia/sample_infohashes.cpp
```diff
@@ -0,0 +1,144 @@
...
+	{
+		m_data_callback(interval, num, std::move(samples), std::move(nodes));
+		m_data_callback = nullptr;
```
arvidn: oh, I see, we ignore all responses after the first one

src/kademlia/sample_infohashes.cpp
```diff
@@ -0,0 +1,144 @@
...
+	{
+#ifndef TORRENT_DISABLE_LOGGING
+		get_observer()->log(dht_logger::traversal, "[%u] wrong or missing samples value"
+			, algorithm()->id());
+#endif
+		timeout();
+	}
+
+	traversal_observer::reply(m);
+	flags |= flag_done;
```
arvidn: I don't recall how this flag works off-hand. doesn't it need to be set before calling the base class' function in order to have done() be called? If you remember why this is set after, would you mind leaving a comment?

arvidn:
sorry for the delay of re-reviewing this. it looks good. would you mind addressing the last few comments? (assuming I understand it correctly)

aldenml:
thank you! I will take care of them, I also need to refresh some details too :)

include/libtorrent/session_handle.hpp
aldenml: you are right, I will change the comment

include/libtorrent/alert_types.hpp
aldenml: adding it

src/kademlia/dht_tracker.cpp
aldenml: true

src/kademlia/node.cpp
aldenml: Ok

src/kademlia/node.cpp
aldenml: Ok

src/kademlia/sample_infohashes.cpp
aldenml: it have no effects if called before `traversal_observer::reply` but it's necessary for the rest of the the logic works with `observer::done()`

aldenml:
@arvidn added comments/changes and rebased

PR #1655, opened 2017-02-03
using std:: for integral types

aldenml:


arvidn:
wow, I guess I picked up some of these in merges

arvidn:
lgtm

PR #1656, opened 2017-02-03
avoid double saving of replacements nodes, removed replacement_cache, test

aldenml:


include/libtorrent/kademlia/node.hpp
```diff
@@ -162,8 +162,10 @@ class TORRENT_EXTRA_EXPORT node : boost::noncopyable
 	// bucket is not full.
 	void add_node(udp::endpoint const& node);
 
+#ifndef TORRENT_NO_DEPRECATE
```
arvidn: if this is not actually exposed in the API, I imagine you could just remove it, no?

arvidn:
good catch! I would imagine you could just remove that code though

include/libtorrent/kademlia/node.hpp
aldenml: yes, agree, I will remove it...removing code :)

arvidn:
back-ported here: https://github.com/arvidn/libtorrent/pull/1661

PR #1659, opened 2017-02-04
fix issue where updates to super seeding state-1.1

arvidn:
would not include the torrent in the state_update_alert


sledgehammer999:
I confirm that it works.

arvidn:
thanks!

PR #1660, opened 2017-02-04
cut down on travis testing a bit to make it quicker

arvidn:


aldenml:
Question: What happens with all the flags in `gcc-sanitizer`, no code will be running with these flags?

arvidn:
The sims still run with sanitizer. Maybe the regular tests should too though

aldenml:
I think it's a good idea to run regular tests with sanitizer too.

arvidn:
@aldenml I think this is good. it cuts down on some of the redundant building and testing. you want to double check?

aldenml:
LGTM

PR #1662, opened 2017-02-05
use 32 bit time points in torrent

arvidn:
@Falcosc does this look reasonable?


Falcosc:
Looks nice.

Would be even better if we could get a meanfull assert failure for time points in future. As long as we have no time point bugs it does not matter, but if you try to fix a time related bug it's a little bit difficult if you don't know how large the difference between the 2 non matching timepoints is.

include/libtorrent/torrent.hpp
```diff
@@ -1287,7 +1285,7 @@ namespace libtorrent
 		// the posix time this torrent was added and when
 		// it was completed. If the torrent isn't yet
 		// completed, m_completed_time is 0
-		time_t m_added_time;
+		time_t m_added_time = time(nullptr);
```
Falcosc: Do you know what the difference between 0 and time(nullptr) is?

arvidn: this was not a semantic change. The initialization was just moved from torrent.cpp:183

arvidn:
So, while fixing the off-by-one time comparison in the python test, I realized that on my Mac, I had a timezone issue. I'm not sure why this hasn't showed up before. But the time I get from ``datetime.now()`` in python, is local time. and ``std::gmtime()`` returns UTC. So the ``datetime`` objects in python would compare 5 hours apart (UTC - EST).

I changed the binding to use ``std::localtime`` which solved it, but I'm not sure I understand why this change triggered this difference for me. @Falcosc Any ideas?

arvidn:
I actually started seeing the timezone issue in master too. It doesn't seem related to this time_point patch

include/libtorrent/torrent.hpp
Falcosc: I don't complain, I just want to know if there is a difference.

Falcosc:
Sorry, I have no idea about the timezone issue.

PR #1664, opened 2017-02-05
slightly improve proxy settings backwards compatibility-1.1

arvidn:


src/session_handle.cpp
```diff
@@ -807,12 +812,14 @@ namespace libtorrent
 
 	proxy_settings session_handle::tracker_proxy() const
 	{
-		return proxy();
+		settings_pack const sett = get_settings();
+		return sett.get_bool(settings_pack::proxy_tracker_connections)
+			? proxy_settings(sett) : proxy();
```
cas--: Do these not both do the same thing? Where instead it should return essentially proxy_settings(type=None)...

arvidn: oops. I meant to return a default-constructed ``proxy_settings``

PR #1670, opened 2017-02-06
fixing sign-conversion warnings, part 12, minor refactor, eol cleanup

aldenml:


src/read_resume_data.cpp
```diff
@@ -165,13 +166,14 @@ namespace libtorrent
...
 		{
 			int const num_files = file_priority.list_size();
-			ret.file_priorities.resize(num_files, 4);
+			ret.file_priorities.resize(aux::numeric_cast<std::size_t>(num_files), 4);
 			for (int i = 0; i < num_files; ++i)
 			{
-				ret.file_priorities[i] = std::uint8_t(
+				std::size_t const idx = std::size_t(i);
+				ret.file_priorities[idx] = aux::numeric_cast<std::uint8_t>(
 					file_priority.list_int_value_at(i, 1));
```
arvidn: I just noticed this. the integer literal "1" here should really be using the default_priority constant (which is 4 these days)

arvidn: I also wonder if this value should be clamped with run-time checks, rather than asserted

aldenml: something like `aux::numeric_clamp`?

arvidn: I would think just ``clamp`` would be a reasonable name

aldenml: ok

arvidn: I think fundamentally, it's risky to assert things that come from untrusted/uncontrolled sources. such as the network or disk. I think those cases need runtime (i.e. release mode) checks instead.

src/read_resume_data.cpp
```diff
@@ -166,12 +167,13 @@ namespace libtorrent
...
 			int const num_files = file_priority.list_size();
-			ret.file_priorities.resize(aux::numeric_cast<std::size_t>(num_files), 4);
+			ret.file_priorities.resize(aux::numeric_cast<std::size_t>(num_files)
+				, default_piece_priority);
 			for (int i = 0; i < num_files; ++i)
 			{
 				std::size_t const idx = std::size_t(i);
-				ret.file_priorities[idx] = aux::numeric_cast<std::uint8_t>(
-					file_priority.list_int_value_at(i, 1));
+				ret.file_priorities[idx] = aux::clamp(
```
arvidn: awesome, thanks!

arvidn:
lgtm

PR #1671, opened 2017-02-06
minor cleanup refactor

aldenml:
@arvidn I'm not sure these `using`s are good addition to `master`, I understand that that could smooth the migration from `1.1` to `1.2` but in order to be effective, it would requires more than this (I think)

### DIFF
--- a/include/libtorrent/torrent_handle.hpp
+++ b/include/libtorrent/torrent_handle.hpp
@@ -77,9 +77,6 @@ namespace libtorrent
 	void TORRENT_NO_RETURN throw_invalid_handle();
 #endif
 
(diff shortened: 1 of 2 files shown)


arvidn:
I'm also not particularly happy with them. On the other hand, they seem fairly cheap and may help some. What else do you have in mind would be required to be effective?

aldenml:
putting them in a more "common" header and remove a lot of `std::` in public api :(

arvidn:
well, since they're just type aliases, the API doesn't *need* to use them, as long as they're the same type. But yes, putting them in a different place is probably a good idea.

PR #1674, opened 2017-02-07
add web_seed_name_lookup_retry to session_settings

ximply:
add web_seed_name_lookup_retry to session_settings to control web seed
name lookup retry in seconds, default is 1800 seconds.

### DIFF
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -795,8 +795,7 @@ namespace libtorrent
 			// time to wait until a new retry of a web seed takes place
 			urlseed_wait_retry,
 			
-			// time to wait until a new retry of a web seed name lookup. Default
-			// is 1800 seconds
+			// time to wait until a new retry of a web seed name lookup
 			web_seed_name_lookup_retry,
 			
 			// sets the upper limit on the total number of files this session will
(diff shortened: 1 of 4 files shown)


include/libtorrent/settings_pack.hpp
```diff
@@ -794,7 +794,11 @@ namespace libtorrent
 
 			// time to wait until a new retry of a web seed takes place
 			urlseed_wait_retry,
-
+			
+			// time to wait until a new retry of a web seed name lookup. Default
```
arvidn: the default value will be extracted from the code by the documentation generation tool. It's better not to mention it in the comment.

ximply: Ok.

aldenml:
Hi @arvidn, I don't know how you want to consider this, but this PR seriously broke some sort of "ABI compatibility". If you have a session state saved with `session_handle::save_state` with a previous version, loading the session with `session_handle::load_state` after this, results in all sort of problems.

arvidn:
great catch! the settings_pack enum must be moved to the end

arvidn:
there should probably be some tests for these enum values too. just to make it harder to do this by accident again

arvidn:
I believe the same thing applies to ``urlseed_max_request_bytes``. I will fix.

arvidn:
I don't really understand how this affects the state saved with ``load_state`` and ``save_state`` though. I thought those would key the settings values by name, not by enum value.

aldenml:
they are saved by name, but if ABI is broken, in `load_pack_from_dict` and `apply_pack`, there is a mismatch with `settings_counts_t` and indexes. I actually suffered the problem, but looking at it again, it looks very subtle, if I have time I will find an example

PR #1675, opened 2017-02-07
fixing sign-conversion warnings, part 13, minor refactor

aldenml:


include/libtorrent/lazy_entry.hpp
```diff
@@ -378,7 +378,7 @@ namespace libtorrent
 
 		// the number of bytes this entry extends in the
 		// bencoded buffer
-		std::uint32_t m_len;
+		std::int32_t m_len;
```
arvidn: this might be a bit delicate. there used to be a bug in the bdecoder where overflowing a 32 bit and effectively circumvent overflow checks

aldenml: ok, I will revert all the `lazy_entry` changes for a later PR, I already have a few changes that need to be pushed alone for careful review.

arvidn:
lgtm

PR #1676, opened 2017-02-07
fix disabled "close_redundant_connections" web_seed disconnect exception

ajax16384:
with  disabled "close_redundant_connections" web_seed disconnect may occure by unexpected http response which raise exception:
````c++
 0: 0155BDC5 std::_Debug_message +37   f:\dd\vctools\crt\crtw32\stdcpp\stdthrow.cpp:17
 1: 014C8ED0 std::_Deque_const_iterator<std::_Deque_val<std::_Deque_simple_types<libtorrent::web_peer_connection::file_request_t> > >::operator* +80   c:\program files (x86)\microsoft visual studio 14.0\vc\include\deque:329
 2: 014C904F std::_Deque_iterator<std::_Deque_val<std::_Deque_simple_types<libtorrent::web_peer_connection::file_request_t> > >::operator* +15   c:\program files (x86)\microsoft visual studio 14.0\vc\include\deque:591
 3: 014CA92C std::deque<libtorrent::web_peer_connection::file_request_t,std::allocator<libtorrent::web_peer_connection::file_request_t> >::front +76   c:\program files (x86)\microsoft visual studio 14.0\vc\include\deque:1438
 4: 014C2825 libtorrent::web_peer_connection::on_receive +1861 libtorrent\src\web_peer_connection.cpp:792
 5: 013CF311 libtorrent::peer_connection::on_receive_data +2273 libtorrent\src\peer_connection.cpp:5850
````
commit adds fix and simulation test

### DIFF
--- a/simulation/test_web_seed.cpp
+++ b/simulation/test_web_seed.cpp
@@ -483,3 +483,47 @@ TORRENT_TEST(urlseed_timeout)
 	TEST_EQUAL(timeout, true);
 }
 
+// check for correct handle of unexpected http status response.
+// with disabled "close_redundant_connections" alive web server connection
+// may be closed in such manner.
+TORRENT_TEST(no_close_redudant_webseed)
+{
+	using namespace libtorrent;
+
+	file_storage fs;
+	fs.add_file("file1", 1);
+	lt::add_torrent_params params = ::create_torrent(fs);
+	params.url_seeds.push_back("http://2.2.2.2:8080/");
+
+	bool expected = false;
+	run_test(
+		[&params](lt::session& ses)
+		{
+			lt::settings_pack pack;
+			pack.set_bool(settings_pack::close_redundant_connections, false);
+			ses.apply_settings(pack);
+			ses.async_add_torrent(params);
+		},
+		[](lt::session& ses, lt::alert const* alert) {},
+		[&expected](sim::simulation& sim, lt::session& ses)
+		{
+			sim::asio::io_service web_server(sim, address_v4::from_string("2.2.2.2"));
+			// listen on port 8080
+			sim::http_server http(web_server, 8080);
+
+			http.register_handler("/file1"
+				, [&expected](std::string method, std::string req
+				, std::map<std::string, std::string>&)
+			{
+				expected = true;
+				char const* extra_headers[4] = { "Content-Range: bytes 0-0/1\r\n", "", "", ""};
+
+				return sim::send_response(206, "Partial Content", 1, extra_headers).append("A").append(sim::send_response(408, "REQUEST TIMEOUT", 0));
+			});
+
+			sim.run();
+		}
+	);
+
+	TEST_CHECK(expected);
+}
--- a/simulation/test_web_seed.cpp
+++ b/simulation/test_web_seed.cpp
@@ -520,7 +520,7 @@ TORRENT_TEST(no_close_redudant_webseed)
 
 				return sim::send_response(206, "Partial Content", 1, extra_headers).
 					append("A").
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/web_peer_connection.cpp [hunk 1]
```diff
@@ -789,9 +789,12 @@ void web_peer_connection::on_receive(error_code const& error
 			// if the status code is not one of the accepted ones, abort
 			if (!is_ok_status(m_parser.status_code()))
 			{
-				file_request_t const& file_req = m_file_requests.front();
-				m_web->have_files.resize(t->torrent_file().num_files(), true);
-				m_web->have_files.clear_bit(file_req.file_index);
+				if (!m_file_requests.empty())
+				{
+					file_request_t const& file_req = m_file_requests.front();
+					m_web->have_files.resize(t->torrent_file().num_files(), true);
```
arvidn: this resize() doesn't look right. it marks all files with lower indices as existing on this web seed..

simulation/test_web_seed.cpp
```diff
@@ -483,3 +483,47 @@ TORRENT_TEST(urlseed_timeout)
...
+			sim::http_server http(web_server, 8080);
+
+			http.register_handler("/file1"
+				, [&expected](std::string method, std::string req
+				, std::map<std::string, std::string>&)
+			{
+				expected = true;
+				char const* extra_headers[4] = { "Content-Range: bytes 0-0/1\r\n", "", "", ""};
+
+				return sim::send_response(206, "Partial Content", 1, extra_headers).append("A").append(sim::send_response(408, "REQUEST TIMEOUT", 0));
```
arvidn: would you mind breaking up this line to be a bit shorter?

ajax16384: ok

src/web_peer_connection.cpp
ajax16384: ok

src/web_peer_connection.cpp
```diff
@@ -789,9 +789,11 @@ void web_peer_connection::on_receive(error_code const& error
 			// if the status code is not one of the accepted ones, abort
 			if (!is_ok_status(m_parser.status_code()))
 			{
-				file_request_t const& file_req = m_file_requests.front();
-				m_web->have_files.resize(t->torrent_file().num_files(), true);
-				m_web->have_files.clear_bit(file_req.file_index);
+				if (!m_file_requests.empty())
+				{
+					file_request_t const& file_req = m_file_requests.front();
+					m_web->have_files.clear_bit(file_req.file_index);
```
arvidn: do you think it would make sense to only do this on certain errors? For instance, this is an obvious reaction to a 404. But if you get a (503) internal error it may not make sense to assume the server is lacking this file. It seems safer to only do this on error codes specifically whitelisted to indicate the file is missing.

arvidn: hm.. I didn't realize this was the previous behavior. including the resize(.., true); I apologize, I remember what the semantics of this field it. It does make sense that the fundamental assumption is that the web server has every file, and only when we fail to get a file from it do we mark this as false. I still think it would be a bit more conservative to look for specific error codes, but I don't expect your patch to do that.

Would you mind restoring this. Your patch was just to protect this against an empty ``m_file_requests``, right?

ajax16384: ok

ajax16384: in next PR can add logic to exclude file idx bit only for (>=400 && < 500) ?  HTTP Status 4XX shows that resource can not be reached by current request

arvidn:
lgtm

PR #1677, opened 2017-02-07
- add windows guid listen_interfaces settings example

ajax16384:
- rearrange comments for corresponding methods

@arvidn  is it ok to toss comment closer to its method so IDE will show appropriate method hint?

### DIFF
--- a/include/libtorrent/torrent_info.hpp
+++ b/include/libtorrent/torrent_info.hpp
@@ -240,14 +240,14 @@ namespace libtorrent
 
 		// ``add_tracker()`` adds a tracker to the announce-list. The ``tier``
 		// determines the order in which the trackers are to be tried.
-		//
+		void add_tracker(std::string const& url, int tier = 0);
+
 		// The ``trackers()`` function will return a sorted vector of
 		// ``announce_entry``. Each announce entry contains a string, which is
 		// the tracker url, and a tier index. The tier index is the high-level
 		// priority. No matter which trackers that works or not, the ones with
 		// lower tier will always be tried before the one with higher tier
 		// number. For more information, see announce_entry_.
-		void add_tracker(std::string const& url, int tier = 0);
 		std::vector<announce_entry> const& trackers() const { return m_urls; }
 
 		// These two functions are related to BEP38_ (mutable torrents). The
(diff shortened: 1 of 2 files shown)


include/libtorrent/torrent_info.hpp
```diff
@@ -418,13 +418,15 @@ namespace libtorrent
 		// to peers over anything other than the i2p network.
 		bool is_i2p() const { return (m_flags & i2p) != 0; }
 
+		int piece_size(piece_index_t index) const { return m_files.piece_size(index); }
+
 		// ``hash_for_piece()`` takes a piece-index and returns the 20-bytes
 		// sha1-hash for that piece and ``info_hash()`` returns the 20-bytes
 		// sha1-hash for the info-section of the torrent file.
+		sha1_hash hash_for_piece(piece_index_t index) const;
```
arvidn: the two ways of getting a hash for a piece I think should be grouped together. I suppose ``piece_size`` could have it's own entry

include/libtorrent/torrent_info.hpp
```diff
@@ -450,37 +452,36 @@ namespace libtorrent
...
 		// libtorrent, the merkle tree will be persisted in the resume data.
-		std::vector<sha1_hash> const& merkle_tree() const { return m_merkle_tree; }
 		void set_merkle_tree(std::vector<sha1_hash>& h)
 		{ TORRENT_ASSERT(h.size() == m_merkle_tree.size() ); m_merkle_tree.swap(h); }
 
 		// ``name()`` returns the name of the torrent.
-		//
-		// ``comment()`` returns the comment associated with the torrent. If
-		// there's no comment, it will return an empty string.
+		// Both the name and the comment is UTF-8 encoded strings.
```
arvidn: this statement would make more sense if ``name()`` and ``comment()`` were grouped together

ajax16384: will use separate UTF8 comment

include/libtorrent/torrent_info.hpp
```diff
@@ -450,37 +452,36 @@ namespace libtorrent
 
 		// ``merkle_tree()`` returns a reference to the merkle tree for this
 		// torrent, if any.
-		//
+		std::vector<sha1_hash> const& merkle_tree() const { return m_merkle_tree; }
```
ajax16384: my idea was to have separate comments for every methods. While I type class/struct method IDE usually popup hint with description. It takes from comment above method declaration.

include/libtorrent/torrent_info.hpp
ajax16384: ok

include/libtorrent/torrent_info.hpp
```diff
@@ -240,14 +240,14 @@ namespace libtorrent
 
 		// ``add_tracker()`` adds a tracker to the announce-list. The ``tier``
 		// determines the order in which the trackers are to be tried.
-		//
+		void add_tracker(std::string const& url, int tier = 0);
```
arvidn: I actually prefer to keep ``trackers()`` and ``add_tracker()`` grouped together

include/libtorrent/torrent_info.hpp
arvidn: I think the makes for worse documentation though. especially functions that belong together, like ``merkle_tree()`` and ``set_merkle_tree()``

arvidn:
lgtm

PR #1680, opened 2017-02-08
back-port ajax16384's fix to enum_net_interfaces on windows

arvidn:


arvidn:
@ajax16384 looks right?

PR #1682, opened 2017-02-10
packet pool fixups

arvidn:
@ajax16384 does this make sense?

### DIFF
--- a/include/libtorrent/packet_pool.hpp
+++ b/include/libtorrent/packet_pool.hpp
@@ -35,7 +35,9 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "libtorrent/config.hpp"
 
+#include "libtorrent/aux_/throw.hpp"
 #include "libtorrent/aux_/vector.hpp"
+#include "libtorrent/assert.hpp"
 
 namespace libtorrent
 {
@@ -84,15 +86,18 @@ namespace libtorrent
 		// deleter for std::unique_ptr
 		void operator()(packet* p) const
 		{
+			TORRENT_ASSERT(p != nullptr);
 			p->~packet();
 			std::free(p);
 		}
 	};
 
-	inline packet* create_packet(int size)
+	inline packet* create_packet(int const size)
 	{
 		packet* p = static_cast<packet*>(std::malloc(sizeof(packet) + size));
+		if (p == nullptr) aux::throw_ex<std::bad_alloc>();
 		new (p) packet();
+		p->allocated = static_cast<std::uint16_t>(size);
 		return p;
 	}
 
@@ -147,7 +152,6 @@ namespace libtorrent
 			TORRENT_ASSERT(allocate <= std::numeric_limits<std::uint16_t>::max());
 
 			packet_ptr p{ alloc(allocate) };
-			p->allocated = static_cast<std::uint16_t>(allocate);
 			return p.release();
 		}
 
--- a/include/libtorrent/packet_pool.hpp
+++ b/include/libtorrent/packet_pool.hpp
@@ -37,6 +37,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "libtorrent/aux_/throw.hpp"
 #include "libtorrent/aux_/vector.hpp"
+#include "libtorrent/aux_/numeric_cast.hpp"
 #include "libtorrent/assert.hpp"
 
 namespace libtorrent
@@ -97,7 +98,7 @@ namespace libtorrent
 		packet* p = static_cast<packet*>(std::malloc(sizeof(packet) + size));
 		if (p == nullptr) aux::throw_ex<std::bad_alloc>();
 		new (p) packet();
-		p->allocated = static_cast<std::uint16_t>(size);
+		p->allocated = aux::numeric_cast<std::uint16_t>(size);
 		return p;
 	}
 
--- a/include/libtorrent/packet_pool.hpp
+++ b/include/libtorrent/packet_pool.hpp
@@ -98,12 +98,16 @@ namespace libtorrent
 
 	using packet_ptr = std::unique_ptr<packet, packet_deleter>;
 
-	template<int ALLOCATE_SIZE>
 	struct TORRENT_EXTRA_EXPORT packet_slab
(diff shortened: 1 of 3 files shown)


include/libtorrent/packet_pool.hpp
```diff
@@ -84,26 +86,33 @@ namespace libtorrent
...
 		}
 	};
 
-	inline packet* create_packet(int size)
+	inline packet* create_packet(int const size)
 	{
 		packet* p = static_cast<packet*>(std::malloc(sizeof(packet) + size));
+		if (p == nullptr) aux::throw_ex<std::bad_alloc>();
 		new (p) packet();
+		p->allocated = static_cast<std::uint16_t>(size);
```
arvidn: this makes sure the ``allocated`` member is accurate.

include/libtorrent/packet_pool.hpp
```diff
@@ -84,26 +86,33 @@ namespace libtorrent
...
 		return p;
 	}
 
 	using packet_ptr = std::unique_ptr<packet, packet_deleter>;
 
-	template<int ALLOCATE_SIZE>
 	struct TORRENT_EXTRA_EXPORT packet_slab
 	{
-		static const int allocate_size{ ALLOCATE_SIZE };
+		int const allocate_size;
```
arvidn: I made this a run-time value instead, so there's only a single (compil-time) instantiation of ``packet_slab``

include/libtorrent/packet_pool.hpp
aldenml: why not use `aux::numeric_cast`?

include/libtorrent/packet_pool.hpp
```diff
@@ -115,17 +124,18 @@ namespace libtorrent
...
-			}
+			if (m_storage.empty()) return packet_ptr(create_packet(allocate_size));
+			auto ret = std::move(m_storage.back());
+			m_storage.pop_back();
+			return ret;
+		}
+
+		void decay()
+		{
+			if (m_storage.empty()) return;
```
arvidn: I remove one packet from here per second, meaning that it will be emptied 10 seconds after a burst ends

include/libtorrent/packet_pool.hpp
arvidn: I suppose that would make sense

include/libtorrent/packet_pool.hpp
ajax16384: i prefered template argument due it's capability to be inlined as value at comparision (cmp ecx, 100), now it's always reads same value from struct insance.

include/libtorrent/packet_pool.hpp
```diff
@@ -153,34 +158,36 @@ namespace libtorrent
...
 
 			packet_ptr pp(p); // takes ownership and may auto free if slab container does not move it
-			std::size_t allocated = pp->allocated;
+			int const allocated = pp->allocated;
 
-			if (allocated <= m_syn_slab.allocate_size) { m_syn_slab.try_push_back(pp); }
-			else if (allocated <= m_mtu_floor_slab.allocate_size) { m_mtu_floor_slab.try_push_back(pp); }
-			else if (allocated <= m_mtu_ceiling_slab.allocate_size) { m_mtu_ceiling_slab.try_push_back(pp); }
+			if (allocated == m_syn_slab.allocate_size) { m_syn_slab.try_push_back(pp); }
+			else if (allocated == m_mtu_floor_slab.allocate_size) { m_mtu_floor_slab.try_push_back(pp); }
```
ajax16384: packets may be allocated with not exact mtu_floor size.
like this sequence(548,548,529,529,529,529,529,529,529,529,529,529,529,528,528 (taken from my packet allocation profile log)). less or equal compare gave opportunity to allocate those packets.

include/libtorrent/packet_pool.hpp
arvidn: sure. it's is *possible* for the same optimization to happen now, since this variable is declared const, the compiler can assume it never changes, and constant fold it. whether that happens or not, I don't know.

In either case, all of the other functions in this class do not need to be instantiated per slab size though, that seems to be unnecessary.

include/libtorrent/packet_pool.hpp
arvidn: if I return a packet whose allocation is 529 into this slab, and next someone needs a packet that can fir 548 bytes, it will return an allocation that's *smaller* than the requested size. The only reason this code worked was because we would only return packets allocated with this same pool, which would have rounded up to 548, for small allocations.

What my patch does is to encode the fact that it was rounded up in the ``allocated`` member (which also allows the packet to grow without reallocating it). It also makes the freeing robust, since we don't make any assumptions about where the packet came from.

Am I misunderstanding something?

include/libtorrent/packet_pool.hpp
```diff
@@ -84,26 +87,33 @@ namespace libtorrent
...
 		}
 	};
 
-	inline packet* create_packet(int size)
+	inline packet* create_packet(int const size)
 	{
 		packet* p = static_cast<packet*>(std::malloc(sizeof(packet) + size));
+		if (p == nullptr) aux::throw_ex<std::bad_alloc>();
 		new (p) packet();
+		p->allocated = aux::numeric_cast<std::uint16_t>(size);
```
arvidn: @ajax16384 this is a critical part of the patch. ``allocated`` is set to actually say how much memory is available

ajax16384: packet->allocated showed how many memory caller of acquire_packet wanted to allocate, but in reality packet occupied sizeof(packet) + slab.allocate_size (which may be great or equal to caller argument). 
i made assumption that only packets acquired from packet_pool will be released to same pool and less or equal logic was based on this fact. if you think that this "optimisation" will constrain somehow packets usage then replacing comparision with strict equal seems right.

arvidn: I think this assumption made the code less robust/more fragile, because it wasn't encoded in the code/type-system.

ajax16384: ok

PR #1686, opened 2017-02-10
fixing sign-conversion warnings, part 15, torrent class

aldenml:


include/libtorrent/torrent.hpp
```diff
@@ -1351,10 +1351,10 @@ namespace libtorrent
 		file_index_t m_error_file;
 
 		// the average time it takes to download one time critical piece
-		std::uint32_t m_average_piece_time = 0;
+		std::int32_t m_average_piece_time = 0;
```
arvidn: hm.. this should really be a ``std::chrono::seconds``. but I guess that's for a separate patch. this is a step in the right direction though, since those are signed

arvidn: ``seconds32`` really I suppose

include/libtorrent/torrent.hpp
```diff
@@ -1481,7 +1481,7 @@ namespace libtorrent
 // ----
 
 		// the maximum number of uploads for this torrent
-		unsigned int m_max_uploads:24;
+		std::uint32_t m_max_uploads:24;
```
arvidn: the adjacent fields in this bitfield need to have the same type too. Otherwise they will be separate "objects"/members.

include/libtorrent/torrent.hpp
aldenml: yes, I rather not do that in this PR

include/libtorrent/torrent.hpp
aldenml: I made some change, but maybe you want to take a look at the entire layout again. I have the impression that maybe something is not up to date

arvidn: yeah, you're right. these bitfields aren't right currently.

src/torrent.cpp
```diff
@@ -9246,9 +9246,9 @@ namespace libtorrent
 			return;
 
 		// now, pick one of the rarest pieces to download
-		int const pick = random(std::uint32_t(rarest_pieces.size() - 1));
+		int const pick = int(random(aux::numeric_cast<std::uint32_t>(rarest_pieces.end_index() - 1)));
 		bool const was_finished = is_finished();
-		m_picker->set_piece_priority(rarest_pieces[pick], 1);
+		m_picker->set_piece_priority(rarest_pieces[pick], default_piece_priority);
```
arvidn: good catch. there's another call on line 9218 that probably should use ``default_piece_priority`` instead of 1.

I will back-port these to 1.1

src/torrent.cpp
```diff
@@ -10754,11 +10754,11 @@ namespace libtorrent
 		m_stats_counters.inc_stats_counter(counters::waste_piece_timed_out + static_cast<int>(reason), b);
 	}
 
-	void torrent::add_failed_bytes(int b)
+	void torrent::add_failed_bytes(int const b)
 	{
 		TORRENT_ASSERT(is_single_thread());
 		TORRENT_ASSERT(b > 0);
-		m_total_failed_bytes += b;
+		m_total_failed_bytes += std::uint32_t(b);
```
arvidn: it seems a bit questionable to use unsigned types for these members (``m_failed_bytes`` and ``m_total_redundant_bytes``). In some sense these warnings is a reminder of that

arvidn:
looks good overall

src/torrent.cpp
aldenml: ok, I will change it

src/torrent.cpp
aldenml: I will change them to signed and put some asserts for overflow

PR #1689, opened 2017-02-10
complete UNC path support

ajax16384:
all internal file api functions now handle long size paths (> MAX_PATH) and special names (CON,PRN ...)
added UNC path unit test

### DIFF
--- a/include/libtorrent/file.hpp
+++ b/include/libtorrent/file.hpp
@@ -176,6 +176,17 @@ namespace libtorrent
 	TORRENT_EXTRA_EXPORT std::string canonicalize_path(string_view f);
 #endif
 
+#ifdef TORRENT_WINDOWS
+// export to be used at unit tests only
+	using win_path_string =
+#if TORRENT_USE_WSTRING
+		std::wstring;
+#else
+		std::string;
+#endif
+	TORRENT_EXTRA_EXPORT  win_path_string convert_to_win_path_string(std::string const& path);
+#endif
+
 	// TODO: move this into a separate header file, TU pair
 	class TORRENT_EXTRA_EXPORT directory : public boost::noncopyable
 	{
--- a/include/libtorrent/file.hpp
+++ b/include/libtorrent/file.hpp
@@ -187,6 +187,14 @@ namespace libtorrent
 // internal export should be used at unit tests only
 	TORRENT_EXTRA_EXPORT native_path_string convert_to_native_path_string(std::string const& path);
 
+// internal export should be used at unit tests only
+	TORRENT_EXTRA_EXPORT std::string convert_from_native_path(char const* s);
+
+#if defined TORRENT_WINDOWS && TORRENT_USE_WSTRING
+// internal export should be used at unit tests only
+	TORRENT_EXTRA_EXPORT std::string convert_from_native_path(wchar_t const* s);
+#endif
+
 	// TODO: move this into a separate header file, TU pair
 	class TORRENT_EXTRA_EXPORT directory : public boost::noncopyable
 	{
--- a/include/libtorrent/file.hpp
+++ b/include/libtorrent/file.hpp
@@ -176,16 +176,16 @@ namespace libtorrent
 	TORRENT_EXTRA_EXPORT std::string canonicalize_path(string_view f);
 #endif
 
-#ifdef TORRENT_WINDOWS
-// export to be used at unit tests only
-	using win_path_string =
-#if TORRENT_USE_WSTRING
+// internal type alias export should be used at unit tests only
+	using native_path_string =
+#if TORRENT_USE_WSTRING && defined TORRENT_WINDOWS
 		std::wstring;
 #else
 		std::string;
 #endif
-	TORRENT_EXTRA_EXPORT  win_path_string convert_to_win_path_string(std::string const& path);
-#endif
+
+// internal export should be used at unit tests only
+	TORRENT_EXTRA_EXPORT native_path_string convert_to_native_path_string(std::string const& path);
 
 	// TODO: move this into a separate header file, TU pair
 	class TORRENT_EXTRA_EXPORT directory : public boost::noncopyable
--- a/src/alert.cpp
+++ b/src/alert.cpp
@@ -175,10 +175,12 @@ namespace libtorrent {
 
 	std::string file_completed_alert::message() const
 	{
-		char msg[200 + TORRENT_MAX_PATH];
-		std::snprintf(msg, sizeof(msg), "%s: file %d finished downloading"
-			, torrent_alert::message().c_str(), static_cast<int>(index));
-		return msg;
+		std::string ret { torrent_alert::message() };
+		char msg[200];
+		std::snprintf(msg, sizeof(msg), ": file %d finished downloading"
+			, static_cast<int>(index));
+		ret.append(msg);
+		return ret;
 	}
 
 	file_renamed_alert::file_renamed_alert(aux::stack_allocator& alloc
@@ -198,10 +200,13 @@ namespace libtorrent {
 
 	std::string file_renamed_alert::message() const
 	{
-		char msg[200 + TORRENT_MAX_PATH * 2];
-		std::snprintf(msg, sizeof(msg), "%s: file %d renamed to %s"
-			, torrent_alert::message().c_str(), static_cast<int>(index), new_name());
-		return msg;
+		std::string ret { torrent_alert::message() };
+		char msg[200];
+		std::snprintf(msg, sizeof(msg), ": file %d renamed to "
+			, static_cast<int>(index));
+		ret.append(msg);
+		ret.append(new_name());
+		return ret;
 	}
 
 	file_rename_failed_alert::file_rename_failed_alert(aux::stack_allocator& alloc
@@ -215,10 +220,12 @@ namespace libtorrent {
 
 	std::string file_rename_failed_alert::message() const
 	{
-		char ret[200 + TORRENT_MAX_PATH * 2];
-		std::snprintf(ret, sizeof(ret), "%s: failed to rename file %d: %s"
-			, torrent_alert::message().c_str(), static_cast<int>(index)
-			, convert_from_native(error.message()).c_str());
+		std::string ret { torrent_alert::message() };
+		char msg[200];
+		std::snprintf(msg, sizeof(msg), ": failed to rename file %d: "
+			, static_cast<int>(index));
+		ret.append(msg);
+		ret.append(convert_from_native(error.message()));
 		return ret;
 	}
 
--- a/src/escape_string.cpp
+++ b/src/escape_string.cpp
@@ -231,12 +231,21 @@ namespace libtorrent
 		if (!need_encoding(path.c_str(), int(path.size())))
 			return url;
 
-		char msg[TORRENT_MAX_PATH*4];
-		std::snprintf(msg, sizeof(msg), "%s://%s%s%s%s%s%s", protocol.c_str(), auth.c_str()
-			, auth.empty() ? "" : "@", host.c_str()
-			, port == -1 ? "" : ":"
-			, port == -1 ? "" : to_string(port).data()
-			, escape_path(path).c_str());
+		std::string msg { protocol };
+		msg.append("://");
+		if (!auth.empty())
+		{
+			msg.append(auth);
+			msg.append("@");
+		}
+		msg.append(host);
+		if (port != -1)
+		{
+			msg.append(":");
+			msg.append(to_string(port).data());
+		}
+		msg.append(escape_path(path));
+
 		return msg;
 	}
 
--- a/src/escape_string.cpp
+++ b/src/escape_string.cpp
@@ -231,7 +231,16 @@ namespace libtorrent
 		if (!need_encoding(path.c_str(), int(path.size())))
 			return url;
 
-		std::string msg { protocol };
+		std::string msg;
+		std::string escaped_path { escape_path(path) };
+		// reserve enough space so further append will
+		// only copy values to existing location
+		msg.reserve(protocol.size() + 3 + // protocol part
+			auth.size() + 1 + // auth part
+			host.size() + // host part
+			1 + 5 + // port part
+			escaped_path.size());
+		msg.append(protocol);
 		msg.append("://");
 		if (!auth.empty())
 		{
@@ -244,7 +253,7 @@ namespace libtorrent
 			msg.append(":");
 			msg.append(to_string(port).data());
 		}
-		msg.append(escape_path(path));
+		msg.append(escaped_path);
 
 		return msg;
 	}
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -337,6 +337,29 @@ namespace libtorrent
 		return int(size);
 	}
 
+	template <typename T>
+	std::string convert_from_native_path(T* s);
+
+	template <>
+	std::string convert_from_native_path(char const* s) { return convert_from_native(s); }
+
+	template <>
+	std::string convert_from_native_path(char* s) { return convert_from_native(s); }
+
+#if TORRENT_USE_WSTRING
+	template <>
+	std::string convert_from_native_path(wchar_t const* s) { return convert_from_wstring(s); }
+
+	template <>
+	std::string convert_from_native_path(wchar_t* s) { return convert_from_wstring(s); }
+#endif
+
+	template <typename T>
+	std::unique_ptr<T, decltype(&std::free)> make_free_holder(T* ptr)
+	{
+		return std::unique_ptr<T, decltype(&std::free)>(ptr, &std::free);
+	}
+
 #ifdef TORRENT_WINDOWS
 	time_t file_time_to_posix(FILETIME f)
 	{
@@ -348,9 +371,11 @@ namespace libtorrent
 		// convert to seconds
 		return time_t(ft / 10000000 - posix_time_offset);
 	}
+#endif
 
-	win_path_string convert_to_win_path_string(std::string const& path)
+	native_path_string convert_to_native_path_string(std::string const& path)
 	{
+#ifdef TORRENT_WINDOWS
 #if TORRENT_USE_UNC_PATHS
 		std::string prepared_path;
 		// UNC paths must be absolute
@@ -373,19 +398,19 @@ namespace libtorrent
 #else
 		return convert_to_native(prepared_path);
 #endif
-	}
-
+#else // TORRENT_WINDOWS
+		return convert_to_native(path);
 #endif
+	}
 
 	void stat_file(std::string const& inf, file_status* s
 		, error_code& ec, int const flags)
 	{
 		ec.clear();
+		native_path_string const& f = convert_to_native_path_string(inf);
 #ifdef TORRENT_WINDOWS
 
 		TORRENT_UNUSED(flags);
-
-		win_path_string file_name = convert_to_win_path_string(inf);
 
 #if TORRENT_USE_WSTRING
 #define CreateFile_ CreateFileW
@@ -394,7 +419,7 @@ namespace libtorrent
 #endif
 
 		// in order to open a directory, we need the FILE_FLAG_BACKUP_SEMANTICS
-		HANDLE h = CreateFile_(file_name.c_str(), 0, FILE_SHARE_DELETE | FILE_SHARE_READ
+		HANDLE h = CreateFile_(f.c_str(), 0, FILE_SHARE_DELETE | FILE_SHARE_READ
 			| FILE_SHARE_WRITE, nullptr, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS, nullptr);
 #undef CreateFile_
 		if (h == INVALID_HANDLE_VALUE)
@@ -426,8 +451,6 @@ namespace libtorrent
 #else
 
 		// posix version
-
-		std::string const& f = convert_to_native(inf);
 
 		struct stat ret;
 		int retval;
@@ -461,13 +484,8 @@ namespace libtorrent
 	{
 		ec.clear();
 
-#ifdef TORRENT_WINDOWS
-		win_path_string f1 = convert_to_win_path_string(inf);
-		win_path_string f2 = convert_to_win_path_string(newf);
-#else
-		std::string const& f1 = convert_to_native(inf);
-		std::string const& f2 = convert_to_native(newf);
-#endif
+		native_path_string const& f1 = convert_to_native_path_string(inf);
+		native_path_string const& f2 = convert_to_native_path_string(newf);
 
 #if TORRENT_USE_WSTRING && defined TORRENT_WINDOWS
 #define RenameFunction_ _wrename
@@ -502,10 +520,8 @@ namespace libtorrent
 	{
 		ec.clear();
 
+		native_path_string const &n = convert_to_native_path_string(f);
 #ifdef TORRENT_WINDOWS
-
-		win_path_string n = convert_to_win_path_string(f);
-
 #if TORRENT_USE_WSTRING
 #define CreateDirectory_ CreateDirectoryW
 #else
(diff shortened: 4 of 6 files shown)


src/file.cpp
```diff
@@ -354,6 +347,47 @@ namespace libtorrent
...
+#if TORRENT_USE_UNC_PATHS
+		std::string prepared_path;
+		// UNC paths must be absolute
+		// network paths are already UNC paths
+		if (path.substr(0,2) == "\\\\")
+			prepared_path = path;
+		else
+		{
+			std::string sep_path { path };
+			replace_separators(sep_path);
```
arvidn: instead of this function, please use: ``std::replace(sep_path.begin(), sep_path.end(), '/', '\\');``. That would work, right?

src/alert.cpp
```diff
@@ -175,10 +175,12 @@ namespace libtorrent {
...
 	{
-		char msg[200 + TORRENT_MAX_PATH];
-		std::snprintf(msg, sizeof(msg), "%s: file %d finished downloading"
-			, torrent_alert::message().c_str(), static_cast<int>(index));
-		return msg;
+		std::string ret { torrent_alert::message() };
+		char msg[200];
+		std::snprintf(msg, sizeof(msg), ": file %d finished downloading"
+			, static_cast<int>(index));
+		ret.append(msg);
```
arvidn: This is to avoid truncating the message?

arvidn: Do you think there's some of this code that can be refactored, to perhaps have a net decrease of lines of code?

src/escape_string.cpp
```diff
@@ -231,12 +231,21 @@ namespace libtorrent
...
 			return url;
 
-		char msg[TORRENT_MAX_PATH*4];
-		std::snprintf(msg, sizeof(msg), "%s://%s%s%s%s%s%s", protocol.c_str(), auth.c_str()
-			, auth.empty() ? "" : "@", host.c_str()
-			, port == -1 ? "" : ":"
-			, port == -1 ? "" : to_string(port).data()
-			, escape_path(path).c_str());
+		std::string msg { protocol };
+		msg.append("://");
```
arvidn: what's the rationale here? I would expect this change to make things slower

src/file.cpp
```diff
@@ -354,6 +347,47 @@ namespace libtorrent
...
 	}
+
+	using win_path_string =
+#if TORRENT_USE_WSTRING
+		std::wstring;
+#else
+		std::string;
+#endif
+
+	void replace_separators(std::string &path)
```
arvidn: was the moving of this function significant?

src/file.cpp
```diff
@@ -946,21 +981,21 @@ namespace libtorrent
 	{
 #if defined TORRENT_WINDOWS && !defined TORRENT_MINGW
 #if TORRENT_USE_WSTRING
-		wchar_t cwd[TORRENT_MAX_PATH];
-		_wgetcwd(cwd, sizeof(cwd) / sizeof(wchar_t));
+		wchar_t* cwd { ::_wgetcwd(nullptr, 0) };
```
arvidn: does this allocate a string that we're expected to free?

src/file.cpp
```diff
@@ -946,21 +981,21 @@ namespace libtorrent
...
-#endif
-#if defined TORRENT_WINDOWS && !defined TORRENT_MINGW && TORRENT_USE_WSTRING
-		return convert_from_wstring(cwd);
-#else
-		return convert_from_native(cwd);
+		char* cwd { ::getcwd(nullptr, 0) };
+		if (cwd == nullptr) return "/";
+		std::string ret { convert_from_native(cwd) };
 #endif
+		std::free(cwd);
```
arvidn: ah, here it is. this is not exception safe. could you capture the ``cwd`` pointer in a RAII deleter? ``std::shared_ptr<char*> holder(cwd, &std::free);`` should work I believe

src/file.cpp
```diff
@@ -1059,33 +1094,32 @@ namespace libtorrent
...
 		// windows does not allow trailing / or \ in
 		// the path when removing files
-		std::string pruned;
-		if (inf[inf.size() - 1] == '/'
-			|| inf[inf.size() - 1] == '\\')
-			pruned = inf.substr(0, inf.size() - 1);
-		else
-			pruned = inf;
+		while (!file_path.empty() && (
+			file_path.back() == '/' ||
```
arvidn: when ``win_path_string`` is ``std::wstring``, I would expect this comparison, of ``wchar_t`` and ``char`` to yield a warning. no?

test/test_file.cpp
```diff
@@ -408,3 +399,139 @@ TORRENT_TEST(stat_file)
...
+		"COM5", "COM6", "COM7", "COM8",
+		"COM9", "LPT1", "LPT2", "LPT3",
+		"LPT4", "LPT5", "LPT6", "LPT7",
+		"LPT8", "LPT9"
+	};
+
+	for (std::string special_name : special_names)
+	{
+		TEST_EQUAL(touch_file(special_name, 10), 0);
+
```
arvidn: it seems there's perhaps a bit too many blank lines here, no?

test/test_file.cpp
```diff
@@ -408,3 +399,139 @@ TORRENT_TEST(stat_file)
...
 	TEST_EQUAL(ec, boost::system::errc::no_such_file_or_directory);
 }
+
+// specificaly UNC tests
+#if TORRENT_USE_UNC_PATHS
+
+int maximum_component_length { 0 };
+bool support_hard_links { false };
+
+bool fill_current_directory_caps()
```
arvidn: what does this function do?

src/file.cpp
ajax16384: ok

src/alert.cpp
ajax16384: was changed to remove MAX_PATH cap dependency

src/file.cpp
ajax16384: yes, it's allocate by std::malloc and must be free after usage. according msdn examples it's a proper way to retrieve cwd > MAX_PATH

src/file.cpp
ajax16384: ok

test/test_file.cpp
ajax16384: It retrieves file system capabilities for current working directory. It was made skip certain test on host file system that can handle it (not to false ASSERT).

src/file.cpp
ajax16384: no, it does not through any warn

ajax16384: where are no warn at all for such code:
````c++
 		std::string str;
		while (!str.empty() && (
			str.back() == '/' ||
			str.back() == '\\'
			)) str.pop_back();

		std::wstring wstr;
		while (!wstr.empty() && (
			wstr.back() == '/' ||
			wstr.back() == '\\'
			)) wstr.pop_back();
````

File: src/escape_string.cpp (hunk 3 above)
Comment (ajax16384) [inline src/escape_string.cpp]: i'll think how to solve it

File: src/file.cpp (hunk 6 above)
Comment (arvidn) [inline src/file.cpp]: I meant to say ``std::unique_ptr`` there btw.

File: test/test_file.cpp (hunk 9 above)
Comment (arvidn) [inline test/test_file.cpp]: ah, I see. that's what "caps" refers to

File: test/test_file.cpp (hunk 8 above)
Comment (ajax16384) [inline test/test_file.cpp]: ok

**Comment by ajax16384** [comment]
@arvidn  is it OK to move win_path_string and convert_to_win_path_string declaration to file.hpp (to be used in unit test internal file api (change_dir and etc)) ?

File: src/file.cpp [hunk 10]
```diff
@@ -1153,23 +1179,26 @@ namespace libtorrent
 		m_inode = 0;
 		// the path passed to FindFirstFile() must be
 		// a pattern
-		std::string f = convert_separators(path);
-		if (!f.empty() && f[f.size()-1] != '\\') f += "\\*";
-		else f += "*";
+		std::string pattern_path{ path };
+		if (!pattern_path.empty() && pattern_path.back() != '\\')
```
arvidn: I'm thinking of ways to make this patch a bit less verbose. what do you think about this?
``patter_path += (pattern_path.empty() && pattern_path.back() != '\\') ? "\\*" : "*";``
perhaps it would even make sense to have a utility function to check if a path ends with a "/" or a "\".

src/file.cpp
```diff
@@ -946,21 +974,20 @@ namespace libtorrent
...
-		char cwd[TORRENT_MAX_PATH];
-		if (getcwd(cwd, sizeof(cwd)) == nullptr) return "/";
-#endif
-#if defined TORRENT_WINDOWS && !defined TORRENT_MINGW && TORRENT_USE_WSTRING
-		return convert_from_wstring(cwd);
-#else
-		return convert_from_native(cwd);
+		char* cwd { ::getcwd(nullptr, 0) };
+		std::shared_ptr<char> holder(cwd, &std::free);
+		std::string ret { cwd == nullptr ? "/" : convert_from_native(cwd) };
```
arvidn: I thinking that the last two lines here are nearly identical in all 3 cases, and may be a candidate for factoring out of the #ifdef check. The only difference is how to treat ``getcwd()`` returning null. Do you know under which conditions it does that? And is it really reasonable to silently ignore it, returning an empty string or "/"?

I realize this is the current behavior though. But would you mind throwing an ``lt::system_error()`` exception in case this fails? The error code (I believe) should be ``error_code(errno, generic_category());``

ajax16384: ok

src/file.cpp
ajax16384: ok

src/file.cpp
```diff
@@ -974,20 +968,28 @@ namespace libtorrent
 	{
 #if defined TORRENT_WINDOWS && !defined TORRENT_MINGW
 #if TORRENT_USE_WSTRING
-		wchar_t* cwd { ::_wgetcwd(nullptr, 0) };
-		std::shared_ptr<wchar_t> holder(cwd, &std::free);
-		std::string ret { cwd == nullptr ? "" : convert_from_wstring(cwd) };
+#define GetCurrentDir_ ::_wgetcwd
+#define PathCharType_ wchar_t
```
arvidn: this change made it even more verbose. Also, for type aliases, please use: ``using <name> = <type>;``

src/file.cpp
```diff
@@ -974,20 +968,28 @@ namespace libtorrent
...
-		char* cwd { ::getcwd(nullptr, 0) };
-		std::shared_ptr<char> holder(cwd, &std::free);
-		std::string ret { cwd == nullptr ? "/" : convert_from_native(cwd) };
-#endif
+#define GetCurrentDir_ ::getcwd
+#define PathCharType_ char
+#define ConvertFrom_ convert_from_native
+#endif
+		PathCharType_* cwd { GetCurrentDir_(nullptr, 0) };
+		if (cwd == nullptr)
```
arvidn: I guess you added this error handling

include/libtorrent/file.hpp
```diff
@@ -176,6 +176,17 @@ namespace libtorrent
 	TORRENT_EXTRA_EXPORT std::string canonicalize_path(string_view f);
 #endif
 
+#ifdef TORRENT_WINDOWS
+// export to be used at unit tests only
+	using win_path_string =
```
arvidn: I think you should move this into each #ifdef block.

src/file.cpp
```diff
@@ -974,20 +968,28 @@ namespace libtorrent
...
 #else
-		char* cwd { ::getcwd(nullptr, 0) };
-		std::shared_ptr<char> holder(cwd, &std::free);
-		std::string ret { cwd == nullptr ? "/" : convert_from_native(cwd) };
-#endif
+#define GetCurrentDir_ ::getcwd
+#define PathCharType_ char
+#define ConvertFrom_ convert_from_native
+#endif
+		PathCharType_* cwd { GetCurrentDir_(nullptr, 0) };
```
arvidn: instead of specifying this type explicitly, you could say: ``auto cwd = GetCurrentDir_(nullptr, 0);``

src/file.cpp
```diff
@@ -974,20 +968,28 @@ namespace libtorrent
...
+		if (cwd == nullptr)
+			aux::throw_ex<system_error>(error_code(errno, generic_category()));
+		std::shared_ptr<PathCharType_> holder(cwd, &std::free);
```
arvidn: for this you could add a function like this: ``template <typename T> std::unique_ptr<T, decltype(&std::free)> make_free_holder(T* ptr) { return std::unique_ptr<T, decltype(&std::free)>(ptr, &std::free); }``.

With such function, you could say: ``auto holder = make_free_holder(cwd);``, without having to explicitly mention the pointer type

src/file.cpp
```diff
@@ -974,20 +968,28 @@ namespace libtorrent
...
+		std::string ret { ConvertFrom_(cwd) };
```
arvidn: you could save a line of code here by saying: ``return ConvertFrom_(cwd);``

src/file.cpp
```diff
@@ -1180,10 +1182,8 @@ namespace libtorrent
 		// the path passed to FindFirstFile() must be
 		// a pattern
 		std::string pattern_path{ path };
-		if (!pattern_path.empty() && pattern_path.back() != '\\')
-			pattern_path.append("\\*");
-		else
-			pattern_path.append("*");
+		pattern_path.append(
```
arvidn: you like ``append()`` better than ``+=``? :)

arvidn:
> @arvidn is it OK to move win_path_string and convert_to_win_path_string declaration to file.hpp (to be used in unit test internal file api (change_dir and etc)) ?

Sure. I mostly prefer to not put functions that are only used by tests in the library itself.

src/file.cpp
ajax16384: yes :)

src/file.cpp
ajax16384: ok

src/file.cpp
ajax16384: ok

ajax16384:
@arvidn i did understand why MingW cwd needs special treatment and use ::getcwd instead of MSVC variant cause vanilla MingW has same _getcwd(io.h) and _wgetcwd(direct.h) declarations.  (blame shows this refs: https://github.com/arvidn/libtorrent/commit/27485a95698c9d95b68ce594b36550d72be88718#diff-65f7f206953a6602cd00af2548a850b5R558 ). May be it is about __MSVCRT__  dependency inside direct.h?

ajax16384:
rebase

arvidn:
I broke master temporarily when merging RC_1_1 into it, and it seems like you rebased this on top of the broken master.  I just landed the fix for master, so if you rebase some of the build errors you see will go away

arvidn:
I don't know about ``getcwd`` on mingw. Is it not just using the posix code path? If so, explicitly qualifying the global scope is just a precaution to not make it mix up with any other names

ajax16384:
may be old mingw was supplied without widechar _wgetcwd? appveyor using mingw environment for gcc build? i think to remove !TORRENT_MINGW for cwd and check how it's build.

ajax16384:
( ::getcwd for MINGW may narrow unicode directory names even WSTRING support is enabled)

ajax16384:
travis: autotools=1 toolset=gcc failed by some other reason? should i rebase again?

src/file.cpp
```diff
@@ -1330,6 +1351,7 @@ namespace libtorrent
 	bool file::open(std::string const& path, int mode, error_code& ec)
 	{
 		close();
+		native_path_string const &file_path = convert_to_native_path_string(path);
```
arvidn: how can this function return a reference to the string? where is the storage for it?

arvidn: I see, it doesn't return a reference. you're relying on the fact that holding a reference to an object extends its lifetime. I don't think there's any benefit of doing this over just storing this directly in a local variable. is there?
I think it would be a lot clearer, in all cases where you call this function, to store it in a string by value.

src/file.cpp
```diff
@@ -336,14 +337,30 @@ namespace libtorrent
 		return int(size);
 	}
 
-#ifdef TORRENT_WINDOWS
-	std::string convert_separators(std::string p)
+	template <typename T>
+	std::string convert_from_native_path(T* s);
+
+	template <>
+	std::string convert_from_native_path(char const* s) { return convert_from_native(s); }
```
arvidn: It doesn't seem right to me that these are template specializations. Is there a reason for not just making them overloads? That typically makes for much more intuitive behavior

src/file.cpp
```diff
@@ -336,14 +337,30 @@ namespace libtorrent
 		return int(size);
 	}
 
-#ifdef TORRENT_WINDOWS
-	std::string convert_separators(std::string p)
+	template <typename T>
+	std::string convert_from_native_path(T* s);
```
arvidn: Having this declaration without a definition, will delay the error. Instead of being a compilation error you'll get a link error. it seems unnecessary

src/file.cpp
```diff
@@ -336,14 +337,30 @@ namespace libtorrent
...
+
+	template <>
+	std::string convert_from_native_path(char const* s) { return convert_from_native(s); }
+
+	template <>
+	std::string convert_from_native_path(char* s) { return convert_from_native(s); }
```
arvidn: how does the overload taking a non-const char pointer differ from the one that take a regular const string? it doesn't actually do any in-place editing, does it?

src/file.cpp
```diff
@@ -336,14 +337,30 @@ namespace libtorrent
...
+
+#if defined TORRENT_WINDOWS && TORRENT_USE_WSTRING
+	template <>
+	std::string convert_from_native_path(wchar_t const* s) { return convert_from_wstring(s); }
+
+	template <>
+	std::string convert_from_native_path(wchar_t* s) { return convert_from_wstring(s); }
```
arvidn: again, I think the non-const overload (or, what should be an overload) is suspect. As far as I can tell, it's redundant and mostly dangerous

test/test_file.cpp
```diff
@@ -408,3 +399,120 @@ TORRENT_TEST(stat_file)
...
+		ec.assign(GetLastError(), system_category());
+		std::printf("GetVolumeInformation: [%s] %s\n"
+			, ec.category().name(), ec.message().c_str());
+		return false;
+	}
+	maximum_component_length = int(dw_maximum_component_length);
+	support_hard_links = ((dw_file_system_flags & FILE_SUPPORTS_HARD_LINKS) != 0);
+	return maximum_component_length > 0;
+#else
+	return false;
```
arvidn: it's not obvious to me why this returns false on non-windows. that excludes a bunch of tests from running.

Is this ever expected to fail on windows? I'm kind of thinking that it might be reasonable to throw on failures.

test/test_file.cpp
```diff
@@ -408,3 +399,120 @@ TORRENT_TEST(stat_file)
...
+	DWORD dw_file_system_flags;
+	if (!GetVolumeInformationA(nullptr, nullptr, 0, nullptr, &dw_maximum_component_length, &dw_file_system_flags, nullptr, 0))
+	{
+		ec.assign(GetLastError(), system_category());
+		std::printf("GetVolumeInformation: [%s] %s\n"
+			, ec.category().name(), ec.message().c_str());
+		return false;
+	}
+	maximum_component_length = int(dw_maximum_component_length);
+	support_hard_links = ((dw_file_system_flags & FILE_SUPPORTS_HARD_LINKS) != 0);
```
arvidn: instead of writing to global variables, I think it would be a lot nicer if you would return a ``std::pair<bool, int>``. Then here you could say:

``return {support_hard_links, maximum_component_length};``

and in the test:

```
int maximum_component_length;
bool support_hard_links;
std::tie(support_hard_links, maximum_component_length) = current_directory_caps();
```

That way you also don't need any global variables

test/test_file.cpp
```diff
@@ -408,3 +399,120 @@ TORRENT_TEST(stat_file)
...
+	for (std::string special_name : special_names)
+	{
+		TEST_EQUAL(touch_file(special_name, 10), 0);
+		TEST_CHECK(lt::exists(special_name));
+		lt::remove(special_name, ec);
+		TEST_EQUAL(ec, error_code());
+		TEST_CHECK(!lt::exists(special_name));
+	}
+
+	if (fill_current_directory_caps())
```
arvidn: the whole ``true`` block below, does it really only make sense to run that on windows?

src/file.cpp
ajax16384: just used old style fron non windows implementation:
````c++
std::string const& f1 = convert_to_native(inf);
```
will remove it

src/file.cpp
ajax16384: ok

src/file.cpp
ajax16384: ok

test/test_file.cpp
ajax16384: ok, will fix non windows

test/test_file.cpp
ajax16384: ok

test/test_file.cpp
ajax16384: this unc tests are usefull particularly on windows - to check if files api correctly handles very long file name. ( fill_current_directory_caps checks if unit test host machine filesystem supports hardlinks and gets max component len)

ajax16384: TORRENT_USE_UNC_PATHS = 1 only in windows build, so this test will be executed only with TORRENT_WINDOWS defined

include/libtorrent/file.hpp
```diff
@@ -187,6 +187,14 @@ namespace libtorrent
 // internal export should be used at unit tests only
 	TORRENT_EXTRA_EXPORT native_path_string convert_to_native_path_string(std::string const& path);
 
+// internal export should be used at unit tests only
+	TORRENT_EXTRA_EXPORT std::string convert_from_native_path(char const* s);
```
arvidn: if ``TORRENT_USE_WSTRING`` is defined, is this function used? it feels like these functions should be mutually exclusive

test/test_file.cpp
```diff
@@ -417,13 +414,13 @@ bool fill_current_directory_caps()
 		ec.assign(GetLastError(), system_category());
 		std::printf("GetVolumeInformation: [%s] %s\n"
 			, ec.category().name(), ec.message().c_str());
-		return false;
+		return std::make_tuple(0, false);
```
arvidn: ``std::pair<>`` would let you just say ``{0, false}``, but either way is fine

include/libtorrent/file.hpp
ajax16384: in MINGW cwd case we should convert from "char* " path even if WSTRING defined

ajax16384: if we could only check MINGW build - to use MSVC way of getting cwd, then convert_from_native_path could include only one type version

ajax16384: i want to setup vanilla MINGW and new MinGW x86-x64 to check if this mingw cwd workaround can be dropped

arvidn: I was under the impression that mingw doesn't have ``std::wstring`` and that ``TORRENT_USE_WSTRING`` would not be defined in that case. But maybe I'm mixing it up with cygwin.

ajax16384: checked vanilla MinGW (https://sourceforge.net/projects/mingw/) and new MinGW (https://sourceforge.net/projects/mingw-w64/).
old MinGW just could not compile libTorrent due lack of c++11 support (std::thread not found, std::snprintf not found and so on).
MinGW-w64 works well out of box (with little tweak for sha1,sha512 source support). 
Removing workaround for CWD gives no error and test_file passed OK.

ajax16384: may i remove it from current_working_directory?

arvidn: sure. there is a mingw build as part of appveyor, right?

ajax16384:
@arvidn  https://ci.appveyor.com/project/arvidn/libtorrent/build/5520/job/2g4uywmti5jm4ef8 it's ok?

PR #1691, opened 2017-02-11
use unique_ptr for packets in utp_stream

arvidn:
@ajax16384 does this make sense? just pulling out to use ``packet_ptr`` in the utp implementation too

### DIFF
--- a/src/packet_buffer.cpp
+++ b/src/packet_buffer.cpp
@@ -42,7 +42,7 @@ namespace libtorrent {
 		, std::uint32_t mask);
 
 #if TORRENT_USE_INVARIANT_CHECKS
-	void packet_buffer_impl::check_invariant() const
+	void packet_buffer::check_invariant() const
 	{
 		int count = 0;
 		for (int i = 0; i < int(m_capacity); ++i)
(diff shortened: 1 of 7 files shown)


ajax16384:
ok

src/packet_buffer.cpp
```diff
@@ -53,15 +53,15 @@ namespace libtorrent {
...
+	packet_ptr packet_buffer::insert(index_type idx, packet_ptr value)
 	{
 		INVARIANT_CHECK;
 
 		TORRENT_ASSERT_VAL(idx <= 0xffff, idx);
 		// you're not allowed to insert NULLs!
 		TORRENT_ASSERT(value);
 
-		if (value == nullptr) return remove(idx);
+		if (!value) return remove(idx);
```
aldenml: @arvidn I have a question here: a couple of lines above, there is an assert for null `TORRENT_ASSERT(value);`, is it OK, valid?

arvidn: I don't think you can compare a ``unique_ptr`` against a raw pointer or ``nullptr`` directly. But it has a bool conversion operator

arvidn: I see what you mean. yes, I believe the call to ``remove()`` is defensive here

PR #1699, opened 2017-02-12
fixing sign-conversion warnings, part 16, utp_stream

aldenml:


src/utp_stream.cpp
```diff
@@ -3613,7 +3614,7 @@ void utp_socket_impl::check_invariant() const
 		i != int((m_outbuf.cursor() + m_outbuf.span()) & ACK_MASK);
 		i = (i + 1) & ACK_MASK)
 	{
-		packet* p = m_outbuf.at(i);
+		packet* p = m_outbuf.at(aux::numeric_cast<packet_buffer::index_type>(i));
```
arvidn: oh, I guess this conflicts with master

PR #1702, opened 2017-02-12
fix ABI compatibility issue introduced with preformatted entry type

arvidn:


sledgehammer999:
Just to clarify: Do I need to use the new function if libtorrent supports it during compilation or is this transparent and qbt compiled with older version will use new code in newer versions?

arvidn:
I could not figure out how to make it transparent (partly because this is a constructor). For you it means that if you don't do anything, you will get the same behavior as 1.0.0 (i.e. malencoded torrents will fail). If you want to fix support for those, you have to detect at compile time whether libtorrent is sufficiently new, and if so, use this new constructor (just pass in ``true`` as the second argument). In which case you'll get this support.

Unfortunately, it won't be until 1.2 I can break the ABI and make this transparent to the client.

sledgehammer999:
@arvidn do you have any rules for when to bump the version? If not, could you do bump to 1.0.12? I do, builds using the git branches instead of the tarballs and it will be impossible to detect the new constructor during compilation.

arvidn:
well, I normally bump before I'm about to cut a release. You could probably detect it with something like: ``template <typename = decltype(libtorrent::create_torrent(torrent_info(), true))>`` (if the constructor isn't there, the substitution failure will remove the overload from the overload set)

arvidn:
``decltype`` is assuming C++11 though

PR #1705, opened 2017-02-13
Add const peer_request p -> peer_request const& p

pavel-pimenov:


src/identify_client.cpp
```diff
@@ -239,7 +239,7 @@ namespace
 
 	struct generic_map_entry
 	{
-		int offset;
+		int const offset;
```
arvidn: this is a bit inconsistent. all fields should really be const, but iirc, the array that these objects are in is already const, so I think this is redundant

src/identify_client.cpp
```diff
@@ -290,7 +290,7 @@ namespace
 			|| ((lhs.id[0] == rhs.id[0]) && (lhs.id[1] < rhs.id[1]));
 	}
 
-	std::string lookup(fingerprint const& f)
+	static std::string lookup(fingerprint const& f)
```
arvidn: if you meant to give this internal linkage, please put it in the anonymous namespace instead of using ``static``

arvidn:
lgtm apart from comments

arvidn:
https://travis-ci.org/arvidn/libtorrent/jobs/201127357#L859

PR #1710, opened 2017-02-14
updated libsimulator module reference

aldenml:
I believe this was pushed in some commit in `RC_1_1` but maybe it was lost in the merge


arvidn:
I did not deliberately update the submodule in RC_1_1, but submodules are notoriously "sticky" and switching branch back and forth can easily cause accidental updates I've noticed.

arvidn:
on a side note; do you know if there's a way to make ``travis_retry`` wait longer in between retries?

aldenml:
agree, I will close this

aldenml:
no idea, I was asking myself the same :)

PR #1711, opened 2017-02-14
more cleanup of load torrent function

aldenml:


arvidn:
I left this in the public API, to provide API compatibility with 1.1.x. I like to deprecate functions at least in one release before removing them

include/libtorrent/session_handle.hpp
```diff
@@ -65,9 +65,6 @@ namespace libtorrent
 	struct session_status;
 #endif
 
-	typedef std::function<void(sha1_hash const&, std::vector<char>&
```
arvidn: I suppose this should be inside a deprecated ifdef though

aldenml:
I understand, but in this particular case, I rather completely remove it due to the following reasoning: I expect a "deprecated" feature to be a _functional_ one. I assume it's better for the client to just remove the feature once he finds that the function is not there anymore.

It's not like this will be the problem in the migration to 1.2 :) but if you still prefer to keep the public function, just let me know to change this PR.

arvidn:
well, I don't have a strong opinion here and I'm open to being convinced. The use case I have in mind are clients that maintain support for being built against multiple versions of libtorrent.

This particular case is a bit special, since I would be surprised if anyone actually used or relied on this feature. Also, not doing anything (and effectively pinning all torrents) is backwards compatible with supporting unloading/loading on the fly, so it could be argued that this is still "functionally compatible".

As a separate issue, the underlying problem still remains, of having a feature that can load torrents on-demand as peers connect and ask for them.

aldenml:
I see, and thanks for the explanation. I will change the PR

aldenml:
Done here https://github.com/arvidn/libtorrent/pull/1715

PR #1712, opened 2017-02-14
added support for enumerating the internal DHT live nodes

aldenml:
Before going any further with tests (and some details), I would like to have some feedback in this public API addition. The context is be able to access internal DHT nodes information in order to use in https://github.com/arvidn/libtorrent/pull/1652.

I believe that this approach allows for a flexible collection (and recreation) of relevant data of nodes. One interesting trick is that if one wants to avoid a big alert in the queue, a crafted "callback" function could be used.

### DIFF
--- a/include/libtorrent/session_handle.hpp
+++ b/include/libtorrent/session_handle.hpp
@@ -442,6 +442,19 @@ namespace libtorrent
 		void dht_get_peers(sha1_hash const& info_hash);
 		void dht_announce(sha1_hash const& info_hash, int port = 0, int flags = 0);
 
+		// Iterate all the live DHT (identified by ``nid``) nodes and pass the
+		// nodes id and endpoint to the callback ``cb``. If the callback function
+		// returns true, the pair will be returned in the list of nodes in the
+		// alert ``dht_live_nodes_alert``.
+		// Since this alert is a response to an explicit call, it will always be
+		// posted, regardless of the alert mask.
+		//
+		// Given that the function is called from an internal thread of libtorrent,
+		// it's important to not stall or make calls to the public API
+		// that can result in race conditions.
+		void dht_live_nodes(sha1_hash const& nid
+			, std::function<bool(sha1_hash const&, udp::endpoint const&)> cb);
+
 		// Send an arbitrary DHT request directly to the specified endpoint. This
 		// function is intended for use by plugins. When a response is received
 		// or the request times out, a dht_direct_response_alert will be posted
--- a/include/libtorrent/session_handle.hpp
+++ b/include/libtorrent/session_handle.hpp
@@ -442,18 +442,12 @@ namespace libtorrent
 		void dht_get_peers(sha1_hash const& info_hash);
 		void dht_announce(sha1_hash const& info_hash, int port = 0, int flags = 0);
 
-		// Iterate all the live DHT (identified by ``nid``) nodes and pass the
-		// nodes id and endpoint to the callback ``cb``. If the callback function
-		// returns true, the pair will be returned in the list of nodes in the
+		// Retrieve all the live DHT (identified by ``nid``) nodes. All the
+		// nodes id and endpoint will be returned in the list of nodes in the
 		// alert ``dht_live_nodes_alert``.
 		// Since this alert is a response to an explicit call, it will always be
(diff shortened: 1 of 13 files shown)


src/alert.cpp
```diff
@@ -2105,4 +2107,62 @@ namespace libtorrent {
...
+	{
+		std::size_t const num_nodes = aux::numeric_cast<std::size_t>(m_num_nodes);
+		std::vector<std::pair<sha1_hash, udp::endpoint>> nodes(num_nodes);
+
+		const char *ptr = m_alloc.get().ptr(m_nodes_idx);
+		for (std::size_t i = 0; i < num_nodes; i++) {
+			std::size_t size = sha1_hash::size();
+			std::memcpy(nodes[i].first.data(), ptr, size);
+			ptr += size;
+			size = detail::read_uint8(ptr);
```
arvidn: instead of mixing IPv4 and IPv6 addresses, requiring a size-prefix, would it make sense to have two separate allocations, one for IPv4 and one for IPv6?

src/alert.cpp
```diff
@@ -2105,4 +2107,62 @@ namespace libtorrent {
...
+
+	int dht_live_nodes_alert::num_nodes() const
+	{
+		return m_num_nodes;
+	}
+
+	std::vector<std::pair<sha1_hash, udp::endpoint>> dht_live_nodes_alert::nodes() const
+	{
+		std::size_t const num_nodes = aux::numeric_cast<std::size_t>(m_num_nodes);
+		std::vector<std::pair<sha1_hash, udp::endpoint>> nodes(num_nodes);
```
arvidn: I think it would be a bit nicer to use ``reserve()`` here and ``emplace_back()`` in the loop. That way you may be able to cut down on the ``memcpy()`` too. actually, I think it would be beneficial to store the node endpoints in a more compact format. i.e. 4-16 bytes for the IP address and 2 bytes for the port. The ``sockaddr_in`` and ``sockaddr_in6`` are likely a lot larger than that

src/kademlia/dht_tracker.cpp
```diff
@@ -536,19 +536,39 @@ namespace libtorrent { namespace dht
...
+#else
+			: nullptr
+#endif
+
+		if (dht == nullptr) return ret;
+
+		dht->m_table.for_each_node([&cb, &ret](node_entry const& e)
+		{
+			udp::endpoint const endp = e.endpoint;
+			if (cb(e.id, endp)) ret.emplace_back(e.id, endp);
```
arvidn: Is this filter function important? I'm a bit hesitant, because it makes it a lot easier to screw things up and make it look like a bug in libtorrent

include/libtorrent/session_handle.hpp
```diff
@@ -443,6 +443,19 @@ namespace libtorrent
...
 		void dht_announce(sha1_hash const& info_hash, int port = 0, int flags = 0);
 
+		// Iterate all the live DHT (identified by ``nid``) nodes and pass the
+		// nodes id and endpoint to the callback ``cb``. If the callback function
+		// returns true, the pair will be returned in the list of nodes in the
+		// alert ``dht_live_nodes_alert``.
+		// Since this alert is a response to an explicit call, it will always be
+		// posted, regardless of the alert mask.
+		//
+		// Given that the function is called from an internal thread of libtorrent,
```
arvidn: this kind of API makes me nervous. do you think there's a simpler way of specifying filters (without a callback) that would cover the majority of use cases?

src/alert.cpp
aldenml: I will think about it

src/alert.cpp
aldenml: ok

src/kademlia/dht_tracker.cpp
aldenml: I understand

include/libtorrent/session_handle.hpp
aldenml: well, I'm thinking something like "distance" and "max", I will work on it

aldenml:
@arvidn I removed the "filtering" thing, I just notice that that's not critical, just saving the session's sate is as memory costly as storing all this info in the alert.

PR #1714, opened 2017-02-15
optimized dht_get_peers_reply_alert storage and created alerts unit test

aldenml:


src/alert.cpp
```diff
@@ -1915,24 +1915,30 @@ namespace libtorrent
...
 		for (auto const& endp : peers)
 		{
-			std::size_t const size = endp.size();
-			TORRENT_ASSERT(size < 0x100);
-			detail::write_uint8(size, ptr);
-			std::memcpy(ptr, endp.data(), size);
-			ptr += size;
+#if TORRENT_USE_IPV6
+			int const size = endp.protocol() == tcp::v4() ? 6 : 18;
+			detail::write_int8(size, ptr);
```
arvidn: I was suggesting having separate allocations for IPv4 and IPv6, to avoid having the size. it would be more compact

test/test_alert_types.cpp
```diff
@@ -0,0 +1,59 @@
...
+	std::vector<tcp::endpoint> const v = {ep1, ep2};
+
+	mgr.emplace_alert<dht_get_peers_reply_alert>(ih, v);
+
+	auto const* a = alert_cast<dht_get_peers_reply_alert>(mgr.wait_for_alert(seconds(0)));
+	TEST_CHECK(a != nullptr);
+
+	TEST_EQUAL(a->info_hash, ih);
+	TEST_EQUAL(a->num_peers(), 2);
+	TEST_CHECK(a->peers() == v);
```
arvidn: in which case this assertion would not necessarily be right, but sorting both vectors before comparing should make it pass

src/alert.cpp
aldenml: ok, I will do that

src/alert.cpp
```diff
@@ -1965,21 +1966,16 @@ namespace libtorrent
 	std::vector<tcp::endpoint> dht_get_peers_reply_alert::peers() const
```
arvidn: interesting. I wonder why this is ``tcp::endpoint``...

test/setup_transfer.cpp
```diff
@@ -121,11 +121,11 @@ address rand_v6()
 
 static std::uint16_t g_port = 0;
 
-tcp::endpoint rand_tcp_ep()
+tcp::endpoint rand_tcp_ep(libtorrent::address(&rand_addr)())
 {
-	// make sure we don't procude the same "random" port twice
+	// make sure we don't procure the same "random" port twice
```
arvidn: I think this was meant to say "produce", but it's not a big deal :)

test/test_alert_types.cpp
```diff
@@ -44,16 +46,27 @@ TORRENT_TEST(dht_get_peers_reply_alert)
...
 	TEST_CHECK(a != nullptr);
 
 	TEST_EQUAL(a->info_hash, ih);
-	TEST_EQUAL(a->num_peers(), 2);
-	TEST_CHECK(a->peers() == v);
+	TEST_EQUAL(a->num_peers(), 5);
+
+	std::vector<tcp::endpoint> const peers = a->peers();
+	TEST_EQUAL(peers.size(), 5);
+	TEST_CHECK(std::includes(peers.begin(), peers.end(), v.begin(), v.end()));
```
arvidn: my understanding is that ``std::includes()`` requires the ranges to be sorted

src/alert.cpp
aldenml: historical reasons...coming from the DHT, that does not make sense, right? Do you have an alternative in mind?

arvidn:
thanks! looks good. I think you need to sort the vectors in the test still

test/setup_transfer.cpp
aldenml: I can change it if another commit is necessary :)

aldenml:
you are right about the sorting. I will change that

src/alert.cpp
arvidn: oh right, these are bittorrent peers. that makes sense

aldenml: thinking again, I believe `tcp::endpoint` is right, since this is endpoint the peer is announced (per hash) in the DHT

aldenml: 👍

arvidn:
lgtm

arvidn:
https://travis-ci.org/arvidn/libtorrent/jobs/202377833#L1672

aldenml:
What?!! Now the endless wait again... :)

aldenml:
I mean the Travis failing constantly for mac

arvidn:
I wonder if there's a way to not rely on brew..

src/alert.cpp
```diff
@@ -1915,37 +1915,44 @@ namespace libtorrent
...
+#endif
 		}
 
-		m_peers_idx = alloc.allocate(int(total_size));
+		m_v4_peers_idx = alloc.allocate(m_v4_num_peers * 6);
+		m_v6_peers_idx = alloc.allocate(m_v6_num_peers * 18);
 
-		char *ptr = alloc.ptr(m_peers_idx);
+		char* v4_ptr = alloc.ptr(m_v4_peers_idx);
+		char* v6_ptr = alloc.ptr(m_v6_peers_idx);
```
arvidn: would you mind wrapping ``v6_ptr`` in ``TORRENT_USE_IPV6``?

aldenml: sure, in a minute, I didn't have the time

arvidn:
from the errors on travis, it looks like you rebased this on top of the broken master. would you mind rebasing it again?

PR #1724, opened 2017-02-17
fix build

arvidn:


arvidn:
@aldenml how do you normally solve these? https://ci.appveyor.com/project/arvidn/libtorrent/build/5533/job/9plya8didscf7jkq#L305

and, I'm not sure why so many of these popped up now... :/

aldenml:
I see some "fix" commits, did you fixed already?

src/bt_peer_connection.cpp
```diff
@@ -1991,7 +1991,7 @@ namespace libtorrent
 		// only, since they might disconnect immediately when
 		// they have downloaded a single piece, although we'll
 		// make another piece available
-		detail::write_uint8(t->is_upload_only() && !t->super_seeding(), ptr);
+		detail::write_uint8(std::uint8_t(t->is_upload_only() && !t->super_seeding()), ptr);
```
aldenml: I expect this cast to be no necessary

aldenml:
Just speculating, maybe they updated `msvc` and now it is more aggressive with the warnings detection. I know some potential warnings `clang` is able to ignore due to the type's size.

src/bt_peer_connection.cpp
arvidn: maybe I misunderstood the warning. msvc was saying that a uint8 was being implicitly cast to a bool (which I believe would happen in ``write_uint8()`` when passed a ``bool``)

arvidn:
it seems far more likely that this is due to the merge I just made. I'm just a bit surprised because I don't recall any significant changes in RC_1_1 that could affect this. but maybe I'm mistaken

PR #1725, opened 2017-02-17
fix build from RC_1_1 merge

aldenml:
Hi @arvidn, something must be changed with the compiler. These warnings are new (the code is old) and both `clang` and `gcc` are able to detect that the bitfields will not overflow and avoid the warnings.

Feel free to discard this PR, cherry-pick or just copy the useful parts

### DIFF
--- a/include/libtorrent/io.hpp
+++ b/include/libtorrent/io.hpp
@@ -80,14 +80,18 @@ namespace libtorrent
 				|| std::is_enum<In>::value>::type>
 		inline void write_impl(In data, OutIt& start)
 		{
-			T val = T(data);
-			TORRENT_ASSERT(data == In(val));
+			T val = static_cast<T>(data);
+			TORRENT_ASSERT(data == static_cast<In>(val));
 			for (int i = int(sizeof(T)) - 1; i >= 0; --i)
 			{
 				*start = static_cast<unsigned char>((val >> (i * 8)) & 0xff);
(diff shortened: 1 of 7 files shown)


aldenml:
also, it seems the compiler is not widening the native types in some places.

include/libtorrent/io.hpp
```diff
@@ -80,15 +80,19 @@ namespace libtorrent
...
 			for (int i = int(sizeof(T)) - 1; i >= 0; --i)
 			{
 				*start = static_cast<unsigned char>((val >> (i * 8)) & 0xff);
 				++start;
 			}
 		}
 
+		template <class T, class OutIt>
+		inline void write_impl(bool val, OutIt& start)
+		{ write_impl<std::uint8_t>(val ? 1 : 0, start); }
```
arvidn: was this necessary in addition to the change above?

aldenml: yes

arvidn:
thanks for fixing this!

include/libtorrent/io.hpp
aldenml: you know, I was thinking that maybe it's better to silence the `msvc` warning above, since it's only inside the `TORRENT_ASSERT(...)`, and remove this new `write_impl(bool val, OutIt& start)` to avoid new code.

What do you think?

PR #1728, opened 2017-02-18
fixing sign-conversion warnings, part 18, packet_buffer

aldenml:


arvidn:
I was thinking of turning the sequence number/ack number into a strong_typedef. Would that cancel some of this out?

aldenml:
probably, and that will probably solve others (related) problems I'm delaying. Cancel the CI jobs if you want, I will close if no useful parts remain

aldenml:
hmm...a lot of error, this is not good. I will close this to avoid noise.

PR #1732, opened 2017-02-19
use auto increment alert_type definition

ajax16384:
make use compiler preprocessor __COUNTER__ to gen alert_type.
i thought about c++11 compile const counter but it has several dissadvantages(recursion depth, slow down to overcome recursion).

### DIFF
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -1712,7 +1712,7 @@ namespace libtorrent
 		static const int static_category = alert::dht_notification;
 		virtual std::string message() const override;
 	};
-	
+
 	TORRENT_DEFINE_ALERT_TYPE_SKIP
 
 	// This is posted whenever a torrent is transitioned into the error state.
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -66,6 +66,15 @@ namespace libtorrent
 	// user defined alerts should use IDs greater than this
 	static const int user_alert_id = 10000;
 
+	static const int base_alert_id_counter = __COUNTER__ + 1;
+
+#define TORRENT_DEFINE_COUNTER_INT(NAME) \
+	static const int NAME = __COUNTER__ - base_alert_id_counter;
+
+#define TORRENT_DEFINE_ALERT_TYPE TORRENT_DEFINE_COUNTER_INT(alert_type)
+
+#define TORRENT_DEFINE_ALERT_TYPE_SKIP static_assert(__COUNTER__ > base_alert_id_counter, "");
+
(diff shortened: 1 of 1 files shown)


include/libtorrent/alert_types.hpp
```diff
@@ -66,6 +66,15 @@ namespace libtorrent
 	// user defined alerts should use IDs greater than this
 	static const int user_alert_id = 10000;
 
+	static const int base_alert_id_counter = __COUNTER__ + 1;
```
arvidn: this is not standard C++. I think it's fine to use this as long as there is a standard fallback for when it's not supported.

ajax16384: so if __COUNTER__ is not supported then constexpr compile counter can be used instead?

arvidn: with a constexpr counter, is the ordering and assignment of ids guaranteed to be deterministic? I imagine it would depend on the order of instantiating templates. I think it's important, especially for ABI compatibility, that the assignment of IDs is deterministic and identical to the ``COUNTER`` preprocessor.

aldenml:
I just saw you include a "dummy" `#include BOOST_PP_UPDATE_COUNTER()` for the deprecated case.

Still not find a clear advantage of using this.

ajax16384:
the advantage was automaticaly calc slot number (not accidently reuse same number, finding next free number e.t.c). the disadvantage is an extra code line for counter increment.

aldenml:
sure, as always, it's a tradeoff. It will not be my decision anyway, I just wanted to give you my opinion and I just hope that if merged, it will not make my `swig` wrapper more difficult to maintain. I also have a strong bias to use `boost` less and less, btw.

aldenml:
Hi @ajax16384, I wonder if with https://github.com/arvidn/libtorrent/pull/1840 do you still think that this brings an advantage over having the number explicitly there.

ajax16384:
test_alert_types.cpp explicitly checks if alert has corresponding integer value. it is especially good for checking autoincrement  :)

aldenml:
what else can I do!! :D

aldenml:
@arvidn what is your position here?

arvidn:
I have a hard time justifying this patch, because it:

* adds a net 100 lines of code
* increases macro use
* increases risk of accidentally insert a new alert that breaks the ABI (but I suppose the test would catch this)

I would much rather look at replacing the ``TORRENT_DEFINE_ALERT`` macro with a template. Maybe make the base class provide some of what it does.

aldenml:
@ajax16384 I think this PR has stalled and unlikely to fly, do you think it's appropriate to close it?

ajax16384:
ok, will think of template base solution

PR #1736, opened 2017-02-19
 Move include into implementation instead(from #1702). 

sledgehammer999:


arvidn:
lgtm, as long as it still builds :)

sledgehammer999:
Bummer, it seems I didn't test correctly. Travis fails, but I don't understand why. My local copy of `create_torrent.cpp` doesn't contain `void bind_create_torrent()` nor any `LIBTORRENT_VERSION_NUM`.
What is happening here!?!?!?

sledgehammer999:
Hmm, it probably means `bindings/python/src/create_torrent.cpp`, right?
Should I insert the include in that file or drop my PR?

sledgehammer999:
I pushed a new commit with different name and I added the version header into the python .cpp.
I was not sure if I should use `<header.hpp>` or `"header.hpp"`. That file contains both...

arvidn:
does this apply to RC_1_1 too? I don't merge RC_1_0 into RC_1_1 anymore (they have diverged too far)

sledgehammer999:
I guess you missed #1737. I hope you merged it before official 1.1.2 release.

PR #1737, opened 2017-02-20
Move include into implementation instead(from #1703).

sledgehammer999:


sledgehammer999:
Is it normal for appveyor to fail? To me the errors seem irrelevant to my changes. I could be wrong. I am not familiar with your tests.

arvidn:
all tests are no deterministic, and sometimes the CI system itself causes failures.

PR #1740, opened 2017-02-20
fixing sign-conversion warnings, part 18

aldenml:


src/session_impl.cpp
```diff
@@ -5504,18 +5504,18 @@ namespace aux {
...
-		s.utp_stats.packets_out = m_stats_counters[counters::utp_packets_out];
-		s.utp_stats.fast_retransmit = m_stats_counters[counters::utp_fast_retransmit];
-		s.utp_stats.packet_resend = m_stats_counters[counters::utp_packet_resend];
-		s.utp_stats.samples_above_target = m_stats_counters[counters::utp_samples_above_target];
-		s.utp_stats.samples_below_target = m_stats_counters[counters::utp_samples_below_target];
-		s.utp_stats.payload_pkts_in = m_stats_counters[counters::utp_payload_pkts_in];
-		s.utp_stats.payload_pkts_out = m_stats_counters[counters::utp_payload_pkts_out];
-		s.utp_stats.invalid_pkts_in = m_stats_counters[counters::utp_invalid_pkts_in];
-		s.utp_stats.redundant_pkts_in = m_stats_counters[counters::utp_redundant_pkts_in];
+		s.utp_stats.packet_loss = std::uint64_t(m_stats_counters[counters::utp_packet_loss]);
```
arvidn: I kind of feel like it would make more sense to change the utp_stats struct to be signed int64 instead. what do you think?

aldenml: I didn't do that because `utp_status` is deprecated and it's not obvious to me if braking ABI in deprecated API is a good idea. You tell me.

arvidn: I see, makes sense

PR #1741, opened 2017-02-21
fixing sign-conversion warnings, part 19

aldenml:


src/lazy_bdecode.cpp
```diff
@@ -561,6 +563,7 @@ namespace libtorrent
 
 	void print_string(std::string& ret, char const* str, int const len, bool single_line)
```
arvidn: ideally this should take a ``string_view`` instead, and ``str.size()`` would not require a cast below. maybe that's a big change though

Issue #1738, opened 2017-02-20
RC_1_1: Pickling of "libtorrent.error_code" instances is not enabled


cas--:
I am seeing an issue using latest RC_1_1 code when trying to pickle the torrent state. It's a serious error in Deluge because torrents with trackers are not saved on shutdown.

I suspected it's last_error added to trackers that's causing the issue:

```
import libtorrent as lt
import pickle
ae = lt.announce_entry('')
pickle.dumps(ae.fail_limit)
pickle.dumps(ae.last_error)
```

```
Traceback (most recent call last):
  File "<pyshell#8>", line 5, in <module>
    pickle.dumps(ae.last_error)
  File "/usr/lib/python2.7/pickle.py", line 1380, in dumps
    Pickler(file, protocol).dump(obj)
  File "/usr/lib/python2.7/pickle.py", line 224, in dump
    self.save(obj)
  File "/usr/lib/python2.7/pickle.py", line 306, in save
    rv = reduce(self.proto)
RuntimeError: Pickling of "libtorrent.error_code" instances is not enabled (http://www.boost.org/libs/python/doc/v2/pickle.html)
```

I am not sure the best solution yet, we don't really need to pickle that entry but it does need fixed as it breaks backward compatibility. Probably best to have a convertor for `error_code` to dict.

arvidn:
would it be sufficient to have a conversion in one direction? from C++ -> python. I suppose a converter in the other direction could just test a few known categories and fail if it's not known.
I imagine the python version would be the error value and the name of the category.

cas--:
I think a one way conversion should suffice. Offhand I can't think where we would pass back EC but I can look at it properly tomorrow...

arvidn:
don't you un-pickle the objects? wouldn't that require a conversion in the other direction?

cas--:
Hmm I don't think so as it's a dict within a dict of a deluge object rather than libtorrent object

arvidn:
@cas-- would you mind writing a test that I can test a converter against?

cas--:
I can do, did you also want to test pickling or just that the dict is as expected?

cas--:
Could simply add to `test_torrent_info.test_announce_entry`:

```
self.assertEquals(ae.last_error['value'], 0)
```

arvidn:
if possible, making sure pickling works would be ideal I think. partly because I'm not confident that it will just because the converter works

cas--:
[...]

PR #1742, opened 2017-02-21
Add a python bindings test for pickling trackers

cas--:
* If ae.last_error (lt.error_code) is not converted in bindings to a dict
 then pickling the trackers with fail.

For #1738

### DIFF
--- a/bindings/python/test.py
+++ b/bindings/python/test.py
@@ -8,6 +8,7 @@ import os
 import shutil
 import binascii
 import inspect
+import pickle
 
 class test_create_torrent(unittest.TestCase):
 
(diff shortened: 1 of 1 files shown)


arvidn:
there are bindings for ``error_code`` and ``error_category`` already. it just says pickling isn't enabled (and you can't subscript it). do you know why that isn't enough?

arvidn:
reading this gives some hints, http://www.boost.org/doc/libs/1_46_1/libs/python/doc/v2/pickle.html
but why is ``error_code`` the problem? no libtorrent object is pickle-enabled as far as I know

cas--:
Yes I know Deluge does not pickle any libtorrent object (intentionally or not) but the problem here is backward compatibility because the old code in 1.3 pickles the trackers but with the recent changes exposing more attributes to the trackers it has created a serious bug that will lose data in Deluge 1.3 when it tries to pickle the trackers containing last_error which is a lt object.

I was just thinking about the problem again and instead of the convertor what is needed is in torrent_handle that the last_error attributes are mapped to a dict. Leaving the lt.announce_entry in torrrent_status as is, since it's not the problem, just the trackers list from torrent_handle suddenly containing a lt error_code object.

arvidn:
There are a few cases in the past where my decision to use python dicts instead of a proper boost.python object has come back and bite me later (because of state that fails to be fully representable in the python version). I think it's a safer direction to enable pickle of error_code and embrace the boost.python binding.

I added another commit on top of yours here: https://github.com/arvidn/libtorrent/pull/1744

PR #1744, opened 2017-02-22
Cas   test conv ec pickled

arvidn:


arvidn:
@cas-- does this make sense?

cas--:
lgtm :)

PR #1745, opened 2017-02-22
fixing sign-conversion warnings, part 20

aldenml:


include/libtorrent/io.hpp
```diff
@@ -80,11 +81,16 @@ namespace libtorrent
 				|| std::is_enum<In>::value>::type>
 		inline void write_impl(In data, OutIt& start)
 		{
+			using OutItT = typename std::iterator_traits<OutIt>::value_type;
+			using Byte = typename std::conditional<
+				std::is_same<OutItT, void>::value, char, OutItT>::type;
```
arvidn: I don't see how ``OutItT`` could ever be ``void``. Is there an example of that now?

arvidn: I take it this is to avoid signed <-> unsigned conversions, is it? It feels a bit questionable to use bit-wise operations on signed integers. doesn't ``static_cast<>`` suppress the conversion warning?

arvidn: I think I get it now. you want this function to accept a ``void*`` as the output iterator, so users don't have to cast. I'm thinking of defining a ``byte`` type [1] instead, and try to use that pervasively for generic memory buffers. Maybe it would make sense to *require* the ``value_type`` be ``byte`` instead (as the final goal).

[1] http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0298r0.pdf

aldenml: I suppose this could be improved in next versions of C++ (or with the use of a new `byte` type), but for now, this is to allow the case when `std::back_insert_iterator<std::string>` is used, in C++11 `::value_type` is `void`. See https://github.com/arvidn/libtorrent/blob/master/src/bt_peer_connection.cpp#L2215
```cpp
std::string remote_address;
std::back_insert_iterator<std::string> out(remote_address);
detail::write_address(remote().address(), out);
```

arvidn: I see, interesting. it seems odd that ``value_type`` would be ``void``

arvidn: would you mind adding a comment with this rationale for the special case of ``void``?

aldenml: sure

PR #1747, opened 2017-02-22
fix in DHT counters, alert types test, typos, formatting

aldenml:


src/session_handle.cpp
```diff
@@ -59,7 +59,7 @@ namespace libtorrent
 #ifndef BOOST_NO_EXCEPTIONS
 			try {
 #endif
-				(m_impl->*f)(a...);
```
arvidn: it's not clear that this will make any difference, since the arguments are already copied into the lambda object (not moved)

arvidn: probably makes sense still though

PR #1750, opened 2017-02-23
fixing sign-conversion warnings, part 21, file_flags

aldenml:


src/file_storage.cpp
```diff
@@ -363,7 +363,7 @@ namespace libtorrent
 	void file_storage::add_file(std::wstring const& file, std::int64_t file_size
 		, int file_flags, std::time_t mtime, string_view symlink_path)
```
arvidn: does it not work to make ``file_flags`` unsigned here?

src/torrent_info.cpp
```diff
@@ -528,7 +528,7 @@ namespace libtorrent
 		}
 		else
 		{
-			file_flags &= ~file_storage::flag_symlink;
+			file_flags &= ~(file_storage::flag_symlink & 0xffffffff);
```
arvidn: is the problem here that the enum value is signed? Since these are flags, it probably would make sense to make the enum unsigned: ``enum file_flags_t : std::uint32_t`` in file_storage.hpp

src/file_storage.cpp
aldenml: this method is deprecated, I like to keep it as it is, but you tell me

src/torrent_info.cpp
aldenml: yes, that's the problem. I wanted to make the minimal amount of changes, to make `file_flags_t` unsigned, I need to change `file_storage::file_flags` too.  I will give it a try

src/file_storage.cpp
arvidn: I think it's probably *more* OK to change the signature of a deprecated function than of a non-deprecated one. This change would at most give a conversion warning in the client though, so I think it's not too bad. It's still an ABI change, but this is for a major release

aldenml: 👍

arvidn:
lgtm

PR #1752, opened 2017-02-23
fix recent settings_pack ABI regression

arvidn:
@aldenml do you mind sanity checking?


aldenml:
LGTM, hopefully you can merge this to master soon :)

arvidn:
it's less urgent in master, I think. since it won't maintain ABI compatibility.. or are you saying these enums leak into the actual saved state on disk?

aldenml:
it's not urgent at all, but the solution I have for saved states (I use `master`) is keeping a marker of the "version" and reset it if it's different. The more "ABI stable" is this, the better.

PR #1755, opened 2017-02-24
fixing sign issues in peer_connection::send_allowed_set

aldenml:
@arvidn I'm not entirely clear how the logic works here, I just made it semantically the same (bitwise speaking)

### DIFF
--- a/src/peer_connection.cpp
+++ b/src/peer_connection.cpp
@@ -547,7 +547,7 @@ namespace libtorrent
 		}
 
 		int const num_allowed_pieces = m_settings.get_int(settings_pack::allowed_fast_set_size);
-		if (num_allowed_pieces == 0) return;
+		if (num_allowed_pieces <= 0) return;
 
 		if (!t->valid_metadata()) return;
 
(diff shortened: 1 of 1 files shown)


src/peer_connection.cpp
```diff
@@ -605,7 +605,8 @@ namespace libtorrent
 			for (int i = 0; i < int(hash.size() / sizeof(std::uint32_t)); ++i)
 			{
 				++loops;
-				piece_index_t const piece(detail::read_uint32(p) % num_pieces);
+				TORRENT_ASSERT(num_pieces > 0);
+				piece_index_t const piece(int(detail::read_uint32(p) % std::uint32_t(num_pieces)));
```
arvidn: could you use ``read_int32()`` instead, and not need the cast?
modulo should still yield a negative value if the input is negative, it it won't do any harm (as far as I can tell)

aldenml: it fails miserable in the tests If I do that, I appreciate if you take a look, because it could be some hidden problem

arvidn: which test(s) specifically? perhaps it would make sense to filter negative piece indices. It's not immediately obvious why modulus is used here either.

arvidn: I think I see now. the modulus need to be applied unsigned types, to get a positive value.

aldenml: do you prefer that I make the change and make it crash to see it? (stop my CI if needed)

arvidn: do you really need the ``int`` cast though? Since you're explicitly constructing the ``piece_index_t``, I would expect you to get an implicit cast there, without issues

aldenml: yes, I just tested in my warn free branch
```
src/peer_connection.cpp:609:55: error: implicit conversion changes signedness: 'unsigned int' to 'int' [-Werror,-Wsign-conversion]
```

arvidn:
lgtm

PR #1756, opened 2017-02-24
in python binding, make announce_entry::last_error a dict instead of error_code

arvidn:


bindings/python/src/torrent_handle.cpp
```diff
@@ -241,7 +241,10 @@ list trackers(torrent_handle& h)
         d["url"] = i->url;
         d["trackerid"] = i->trackerid;
         d["message"] = i->message;
-        d["last_error"] = i->last_error;
+        dict last_error;
+        last_error["value"] = i->last_error.value();
+        last_error["category"] = i->last_error.category().name();
```
cas--: Could also add `last_error["message"]`

arvidn: the problem with that is that it's likely in the language of the current locale. in my mind the message is something that should be generated as late as possible

PR #1758, opened 2017-02-25
default TOS marking to 0x20

arvidn:


ssiloti:
Are we sure this does what we think it does? I'm having a hard time finding clear, authoritative information on what the Class Selector values actually do in practice. The most authoritative source appears to be RFC2474 Section 4.2.2.2 which states that:

> We refer to a Class Selector Codepoint with a larger numerical value than another Class Selector Codepoint as having a higher relative order while a Class Selector Codepoint with a smaller numerical value than another Class Selector Codepoint is said to have a lower relative order.  The set of PHBs mapped to by the eight Class Selector Codepoints MUST yield at least two independently forwarded classes of traffic, and PHBs selected by a Class Selector Codepoint SHOULD give packets a probability of timely forwarding that is not lower than that given to packets marked with a Class Selector codepoint of lower relative order, under reasonable operating conditions and traffic loads.

Which implies that CS1 has precedence greater-than-or-equal-to the default. This is certainly not what we want.

arvidn:
@chromi would you mind chiming in?

chromi:
Thanks for forwarding me to the correct repo.

RFC2474 is very old - it is the first RFC defining the Diffserv field as such, replacing the old Precedence and TOS fields in the TOS byte.  Note that it is dated 1998; a lot happens in two decades on the Internet.  At that time, retaining backwards compatibility with the existing uses of the TOS field was paramount.  Nowadays, the original Precedence meanings have largely died out.

Somewhat more recent is RFC4594, in which Section 4.10 describes a "Low Priority Data" class with the CS1 codepoint being recommended.  I consider this to be the best available summary of DSCP meanings.

The definition of CS1 as low-priority is also enshrined in typical implementations of IEEE 802.11e, which defines four priority classes for Wifi MACs.  The mapping of DSCPs to Wifi classes is generally very crude, but Precedence 0 and 3 map to "Best Effort", Precedence 1 and 2 map to "Bulk" (which has the lowest priority), Precedence 4 and 5 map to "Video", and Precedence 6 and 7 map to "Voice" (which has the highest priority but the least aggregation permitted).

ssiloti:
@chromi I see. Thanks for the info.

chromi:
Incidentally, I had already spotted that there was a TOS setting documented, which did make me more confident in including that in my suggestions.

Diffserv *is* rather a mess of a specification family.  It's pretty hard to implement correctly just by reading the RFCs, especially since they often partially contradict each other without actually getting obsoleted.  There's a whole industry of Highly Paid Consultants specialising in setting this sort of thing up, generally revolving around expensive Cisco kit.  At some point, you have to punt and code to the *spirit* of the specs, rather than the letter of them.

In my own QoS system, I've defaulted to a simple three-class system: one for best-effort traffic, one for bulk traffic (which catches CS1 and pretty much nothing else), and one for latency-sensitive traffic (CS7, CS6, EF, VA, etc).  It seems to work as expected, and tries to provide the right incentives to use the correct DSCP.  Very few applications do so far - but libtorrent using CS1 by default should move the needle significantly in the right direction.

PR #1761, opened 2017-02-26
fixing sign-conversion warnings, part 22

aldenml:


src/torrent.cpp
```diff
@@ -1421,14 +1421,14 @@ namespace libtorrent
...
-			std::size_t name_length = domain->length;
+			std::size_t name_length = aux::numeric_cast<std::size_t>(domain->length);
 
 #ifndef TORRENT_DISABLE_LOGGING
 			if (i > 1) names += " | n: ";
 			names.append(torrent_name, name_length);
 #endif
-			if (strncmp(torrent_name, "*", name_length) == 0
-				|| strncmp(torrent_name, m_torrent_file->name().c_str(), name_length) == 0)
+			if (std::strncmp(torrent_name, "*", name_length) == 0
```
arvidn: if these string comparisons are the only place ``name_length`` is used in,  it should probably be changed to use a ``string_view``

PR #1763, opened 2017-02-27
fixing sign-conversion warnings, part 23, final

aldenml:


aldenml:
Well, this is it. I just need help in one line (marked with `TODO`)

src/utp_stream.cpp
```diff
@@ -2640,7 +2640,7 @@ bool utp_socket_impl::incoming_packet(span<std::uint8_t const> buf
 		m_reply_micro = timestamp - ph->timestamp_microseconds;
 		std::uint32_t prev_base = m_their_delay_hist.initialized() ? m_their_delay_hist.base() : 0;
 		their_delay = m_their_delay_hist.add_sample(m_reply_micro, step);
-		int base_change = m_their_delay_hist.base() - prev_base;
+		int base_change = int(m_their_delay_hist.base()) - int(prev_base); // TODO: review this sign conversion
```
aldenml: here, I need some help. Below there is a line like `if (prev_base && base_change < 0 && base_change > -10000...`, but how `base_change` could be negative without overflow?

include/libtorrent/packet_pool.hpp
```diff
@@ -198,8 +200,8 @@ namespace libtorrent
 			else if (allocate <= m_mtu_ceiling_slab.allocate_size) { return m_mtu_ceiling_slab.alloc(); }
 			return create_packet(allocate);
 		}
-		static int const mtu_floor_size = TORRENT_INET_MIN_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER;
-		static int const mtu_ceiling_size = TORRENT_ETHERNET_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER;
+		static constexpr int mtu_floor_size = TORRENT_INET_MIN_MTU - TORRENT_IPV4_HEADER - TORRENT_UDP_HEADER;
```
arvidn: does the ``constexpr`` make a difference here? I would expect both to be equally "constant".

src/utp_stream.cpp
arvidn: right, I don't think your change here is correct (and perhaps the original code isn't either). It's critical that the subtraction here is on *unsigned* types with wrapping semantics, and *then* cast to signed. the -10000 threshold seems a bit suspicious (as in, it should be a named constant). I imagine that's a threshold to avoid enabling this logic too eagerly.

please make ``base_change`` and ``prev_base`` const too (I believe they can be)

arvidn: I suppose, technically, this would assume modulo-2 signed integer representation. Perhaps there's a more correct way of doing it, but modulo-2 is a pretty safe assumption still

src/utp_stream.cpp
```diff
@@ -3042,11 +3042,11 @@ bool utp_socket_impl::incoming_packet(span<std::uint8_t const> buf
 				// sure to clamp it as a sanity check
 				if (delay > min_rtt) delay = min_rtt;
 
-				do_ledbat(acked_bytes, delay, prev_bytes_in_flight);
-				m_send_delay = delay;
+				do_ledbat(acked_bytes, aux::numeric_cast<int>(delay), prev_bytes_in_flight);
```
arvidn: I'm a little bit worried that these numeric_casts will cause a lot of spurious asserts, as I don't think there are any runtime checks for these values to stay within bounds.

Jamfile
```diff
@@ -244,7 +244,6 @@ rule warnings ( properties * )
 
 # enable these warnings again, once the other ones are dealt with
 		result += <cflags>-Wno-weak-vtables ;
-		result += <cflags>-Wno-sign-conversion ;
```
arvidn: yay :) is there an msvc counterpart flag too?

src/utp_stream.cpp
aldenml: ok, I will do that

include/libtorrent/packet_pool.hpp
aldenml: none at all, I will revert this

src/utp_stream.cpp
aldenml: ok, I will do the simple casts then

Jamfile
aldenml: :) yes it is, C4365 and it is activated with `/Wall`, I have a pending review in `Jamfile` and I will check this (and others flags).

PR #1765, opened 2017-02-27
making torrent::we_have a little more robust while iterating connections

aldenml:
This is to partially address this issue https://github.com/frostwire/frostwire-jlibtorrent/issues/157


aldenml:
taking another look at this, I'm not 100% if this is indeed better, closing it

PR #1767, opened 2017-02-28
optimize bdecode_node offset calculation

ajax16384:
save couple of cpu commands for offset calculcation

### DIFF
--- a/include/libtorrent/bdecode.hpp
+++ b/include/libtorrent/bdecode.hpp
@@ -202,6 +202,9 @@ struct bdecode_token
 
 	int start_offset() const { TORRENT_ASSERT(type == string); return int(header) + 2; }
 
+	// diff between current and next item offset
+	int diff_size() const { bdecode_token const* next(this); ++next; return next->offset - offset; }
+
 	// offset into the bdecoded buffer where this node is
 	std::uint32_t offset:29;
 
--- a/include/libtorrent/bdecode.hpp
+++ b/include/libtorrent/bdecode.hpp
@@ -202,9 +202,6 @@ struct bdecode_token
 
 	int start_offset() const { TORRENT_ASSERT(type == string); return int(header) + 2; }
 
-	// diff between current and next item offset
-	int diff_size() const { bdecode_token const* next(this); ++next; return next->offset - offset; }
-
 	// offset into the bdecoded buffer where this node is
 	std::uint32_t offset:29;
 
--- a/src/bdecode.cpp
+++ b/src/bdecode.cpp
@@ -112,7 +112,7 @@ namespace libtorrent
 
 	// diff between current and next item offset
 	// should only be called for non last item in array
-	int calc_token_in_array_size(bdecode_token const& t)
+	int token_source_span(bdecode_token const& t)
 	{
 		return (&t)[1].offset - t.offset;
 	}
@@ -486,7 +486,7 @@ namespace libtorrent
 		{
 			bdecode_token const& t = tokens[token];
 			TORRENT_ASSERT(t.type == bdecode_token::string);
-			int const size = calc_token_in_array_size(t) - t.start_offset();
+			int const size = token_source_span(t) - t.start_offset();
 			if (int(key.size()) == size
 				&& std::equal(key.data(), key.data() + size, m_buffer
 					+ t.offset + t.start_offset()))
@@ -561,7 +561,7 @@ namespace libtorrent
 	{
 		TORRENT_ASSERT(type() == int_t);
 		bdecode_token const& t = m_root_tokens[m_token_idx];
-		int const size = calc_token_in_array_size(t);
+		int const size = token_source_span(t);
 		TORRENT_ASSERT(t.type == bdecode_token::integer);
 
 		// +1 is to skip the 'i'
@@ -581,7 +581,7 @@ namespace libtorrent
 	{
 		TORRENT_ASSERT(type() == string_t);
 		bdecode_token const& t = m_root_tokens[m_token_idx];
-		std::size_t const size = aux::numeric_cast<std::size_t>(calc_token_in_array_size(t) - t.start_offset());
+		std::size_t const size = aux::numeric_cast<std::size_t>(token_source_span(t) - t.start_offset());
 		TORRENT_ASSERT(t.type == bdecode_token::string);
 
 		return string_view(m_buffer + t.offset + t.start_offset(), size);
@@ -600,7 +600,7 @@ namespace libtorrent
 		TORRENT_ASSERT(type() == string_t);
 		bdecode_token const& t = m_root_tokens[m_token_idx];
 		TORRENT_ASSERT(t.type == bdecode_token::string);
-		return calc_token_in_array_size(t) - t.start_offset();
+		return token_source_span(t) - t.start_offset();
 	}
 
 	void bdecode_node::reserve(int tokens)
--- a/src/bdecode.cpp
+++ b/src/bdecode.cpp
@@ -194,6 +194,8 @@ namespace libtorrent
 	}
 
 	// diff between current and next item offset
+	int tokens_offset_diff(bdecode_token const* t);
+
 	int tokens_offset_diff(bdecode_token const* t)
 	{
 		bdecode_token const* next(t);
--- a/src/bdecode.cpp
+++ b/src/bdecode.cpp
@@ -477,22 +477,22 @@ namespace libtorrent
 
 		while (tokens[token].type != bdecode_token::end)
 		{
-			bdecode_token const* t = &tokens[token];
-			TORRENT_ASSERT(t->type == bdecode_token::string);
-			int const size = (t + 1)->offset - t->offset - t->start_offset();
+			bdecode_token const& t = tokens[token];
+			TORRENT_ASSERT(t.type == bdecode_token::string);
+			int const size = (&t)[1].offset - t.offset - t.start_offset();
 			if (int(key.size()) == size
 				&& std::equal(key.data(), key.data() + size, m_buffer
-					+ t->offset + t->start_offset()))
+					+ t.offset + t.start_offset()))
 			{
 				// skip key
-				token += t->next_item;
+				token += t.next_item;
 				TORRENT_ASSERT(tokens[token].type != bdecode_token::end);
 
 				return bdecode_node(tokens, m_buffer, m_buffer_size, token);
 			}
 
 			// skip key
-			token += t->next_item;
+			token += t.next_item;
 			TORRENT_ASSERT(tokens[token].type != bdecode_token::end);
 
 			// skip value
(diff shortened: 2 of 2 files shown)


aldenml:
question: how significant is the optimization? (if any)

ajax16384:
for x86 release build indexed item access "m_root_tokens[token + 1].offset" ([eax+esi*8+8]) was replaced with next item address "(t +1)->offset" ([edi+8])

arvidn:
I don't like that pointer. could you instead take the address of the reference and index from there? Something like: ``(&t)[1].offset``

src/bdecode.cpp
```diff
@@ -477,22 +477,22 @@ namespace libtorrent
 
 		while (tokens[token].type != bdecode_token::end)
 		{
-			bdecode_token const& t = tokens[token];
-			TORRENT_ASSERT(t.type == bdecode_token::string);
-			int const size = m_root_tokens[token + 1].offset - t.offset - t.start_offset();
+			bdecode_token const* t = &tokens[token];
+			TORRENT_ASSERT(t->type == bdecode_token::string);
+			int const size = (t + 1)->offset - t->offset - t->start_offset();
```
arvidn: it's not obvious from just looking at the code here that m_root_tokens[token + 1] is the same as (t + 1). If it is, it should also be the same as ``tokens[token+1]``, I would think compilers would have an easier time to optimize that

ajax16384: "m_root_tokens" reference instead of local "tokens" comes from original "bdecode land" commit. from "bdecode_node::string_length()" implementation i figure out that "tokens" should definilty be used to access next item.

aldenml:
for what I see, performance is not really an issue here. If anything, a refactor for a more clear code could be done, something like:
```cpp
bdecode_token const& t = tokens[token];
bdecode_token const& next = tokens[token + 1];
```
and let the compiler do their work. I don't see how this `(&t)[1]` is actually better.

ajax16384:
@aldenml gcc can not deduce that "next" address should be calculated by only adding constant to "&t".

aldenml:
@ajax16384 well, it should be :) I rather let the CPU spend a few more ticks if the path is not critical, than having a source code less clear.
sorry if I sound a little harsh, btw, not my intention

ajax16384:
@aldenml no problem. i tried to consolidate calculcation to diff_size for better readability

include/libtorrent/bdecode.hpp
```diff
@@ -202,6 +202,9 @@ struct bdecode_token
 
 	int start_offset() const { TORRENT_ASSERT(type == string); return int(header) + 2; }
 
+	// diff between current and next item offset
+	int diff_size() const { bdecode_token const* next(this); ++next; return next->offset - offset; }
```
arvidn: this is a very dangerous function. it makes assumptions about where its object is allocated, or rather, that another identical object is allocated immediately following it. For every array, this assumption is false for one element.

I don't think this belongs as a member function, as it cannot satisfy the basic assumption that the following code would be valid:

```
bdecode_token a;
a.diff_size();
```

src/bdecode.cpp
```diff
@@ -479,7 +479,7 @@ namespace libtorrent
 		{
 			bdecode_token const& t = tokens[token];
 			TORRENT_ASSERT(t.type == bdecode_token::string);
-			int const size = m_root_tokens[token + 1].offset - t.offset - t.start_offset();
```
arvidn: did you try replacing ``m_root_tokens`` by ``tokens`` instead?

arvidn: also, could you try to make ``tokens`` const? my assumption is that it won't actually change in this function.

`bdecode_token const* const tokens = m_root_tokens;`

arvidn:
@ajax16384 would you mind providing some more rationale for this change? I take it you have the generated x86 assembly before and after for instance, right?

include/libtorrent/bdecode.hpp
ajax16384: ok

src/bdecode.cpp
ajax16384: ok

ajax16384:
before:
````asm
			bdecode_token const& t = tokens[token];
			TORRENT_ASSERT(t.type == bdecode_token::string);
			int const size = m_root_tokens[token + 1].offset - t.offset - t.start_offset();
50A32520  mov         eax,dword ptr [tokens]  
50A32523  mov         edx,dword ptr [ebx]  
50A32525  mov         edi,dword ptr [ebx+4]  
50A32528  and         edx,1FFFFFFFh  
50A3252E  mov         ecx,edi  
50A32530  mov         eax,dword ptr [eax+esi*8+8]  
50A32534  and         eax,1FFFFFFFh  
50A32539  shr         ecx,1Dh  
50A3253C  sub         eax,edx  
50A3253E  mov         dword ptr [ebp-1Ch],ecx  
50A32541  sub         eax,ecx  
50A32543  sub         eax,2  
````
after:
````asm
			bdecode_token const& t = tokens[token];
			TORRENT_ASSERT(t.type == bdecode_token::string);
			int const size = tokens_offset_diff(&t) - t.start_offset();
0F532520  mov         ecx,dword ptr [edi]  
0F532522  mov         edx,dword ptr [edi+8]  
0F532525  and         ecx,1FFFFFFFh  
0F53252B  mov         ebx,dword ptr [edi+4]  
0F53252E  and         edx,1FFFFFFFh  
0F532534  mov         eax,ebx  
0F532536  sub         edx,ecx  
0F532538  shr         eax,1Dh  
0F53253B  sub         edx,eax  
0F53253D  sub         edx,2  
````

File: src/bdecode.cpp (hunk 3 above)
Comment (arvidn) [inline src/bdecode.cpp]: did you try replacing ``m_root_tokens`` by ``tokens``? Does that not generate the efficient code?

Comment (ajax16384) [inline src/bdecode.cpp]: for msvc release replacing "m_root_tokens" with "tokens" generates good code (ms deduce how to optimal calculate address), but mingw gcc still remains unoptimized. i will post later gcc mingw disasm

**Comment by aldenml** [comment]
and this asm is with which optimization flag?

File: src/bdecode.cpp [hunk 4]
```diff
@@ -193,6 +193,15 @@ namespace libtorrent
 		}
 	}
 
+	// diff between current and next item offset
+	int tokens_offset_diff(bdecode_token const* t);
```
aldenml: this is weird

ajax16384: apple gcc validator needs proto even for local functions

ajax16384: msvc /O2

aldenml: why this does not apply to the other local functions?

aldenml:
You already said that `msvc` is fine, what was the optimization flag with mingw gcc?

ajax16384:
g++.exe (i686-posix-dwarf-rev1, Built by MinGW-W64 project) 6.3.0
 (bjam release)
"g++"  -ftemplate-depth-128 -O3 -finline-functions -Wno-inline -Wall -mthreads -Wc++11-compat -std=c++11 -Wall -Wextra -Wno-format-zero-length -Wno-long-long -Wno-unused-variable -Wparentheses -Wpedantic -Wvla 

original:
````asm
    1b2e:	0f b6 46 03          	movzx  eax,BYTE PTR [esi+0x3]
    1b32:	8b 6f 1c             	mov    ebp,DWORD PTR [edi+0x1c]
    1b35:	8b 0e                	mov    ecx,DWORD PTR [esi]
    1b37:	83 e0 e0             	and    eax,0xffffffe0
    1b3a:	8b 6c dd 08          	mov    ebp,DWORD PTR [ebp+ebx*8+0x8]
    1b3e:	81 e1 ff ff ff 1f    	and    ecx,0x1fffffff
    1b44:	81 e5 ff ff ff 1f    	and    ebp,0x1fffffff
    1b4a:	29 cd                	sub    ebp,ecx
    1b4c:	3c 60                	cmp    al,0x60
````

"m_root_tokens" replaced with "tokens":
````asm
    1b2b:	0f b6 45 03          	movzx  eax,BYTE PTR [ebp+0x3]
    1b2f:	8b 7d 00             	mov    edi,DWORD PTR [ebp+0x0]
    1b32:	8b 4c de 08          	mov    ecx,DWORD PTR [esi+ebx*8+0x8]
    1b36:	83 e0 e0             	and    eax,0xffffffe0
    1b39:	81 e7 ff ff ff 1f    	and    edi,0x1fffffff
    1b3f:	81 e1 ff ff ff 1f    	and    ecx,0x1fffffff
    1b45:	29 f9                	sub    ecx,edi
    1b47:	3c 60                	cmp    al,0x60
````

tokens_offset_diff:
````asm
    1b1c:	0f b6 46 03          	movzx  eax,BYTE PTR [esi+0x3]
    1b20:	8b 16                	mov    edx,DWORD PTR [esi]
    1b22:	8b 4e 08             	mov    ecx,DWORD PTR [esi+0x8]
    1b25:	83 e0 e0             	and    eax,0xffffffe0
    1b28:	81 e2 ff ff ff 1f    	and    edx,0x1fffffff
    1b2e:	81 e1 ff ff ff 1f    	and    ecx,0x1fffffff
    1b34:	29 d1                	sub    ecx,edx
    1b36:	3c 60                	cmp    al,0x60
````

**Comment by ajax16384** [comment]
just want to find elegant solution that will help produce compiler better code

File: src/bdecode.cpp (hunk 3 above)
Comment (arvidn) [inline src/bdecode.cpp]: I'd be happy with the patch to use ``tokens`` instead of ``m_root_tokens``. I think the patch as it stands now adds complexity and makes it less clear what the code is trying to achieve. If you think it's important that gcc 6.3 also generates the more efficient code, I would hope there would be a solution that's simple and clear as well

File: src/bdecode.cpp [hunk 5]
```diff
@@ -193,6 +193,13 @@ namespace libtorrent
 		}
 	}
 
+	// diff between current and next item offset
+	inline int tokens_offset_diff(bdecode_token const* t)
```
arvidn: first: there's no need to use pointers here (unless you tell me that the optimizer somehow can optimize a pointer but not a reference). A pointer, semantically, mean that something can be null. a reference is the same thing, but without pointer arithmetic and without being null. It seems like a clear win here. ``t`` cannot be null.

However, you do need the pointer arithmetic here. I suggested you try: ``(&t)[1].offset`` instead of ``(t + 1)->offset``. Did you try that, and did you get a different result? I would be very surprised if it generates different code.

I can't say I'm keen on having a function like this at all, but here are some comments about it:

1. the name does not really describe what it does (at least not as far as I can tell)
2. It should not be ``inline`` (it won't appear in any other translation units)
3. It should have internal linkage (i.e. be in the anonymous namespace)

ajax16384: i have already tried (&t)[1] - should work ok

src/bdecode.cpp
```diff
@@ -109,6 +109,14 @@ namespace libtorrent
 		// reading a key or a vale. 0 means key 1 is value
 		std::uint32_t state:1;
 	};
+
+	// diff between current and next item offset
+	// should only be called for non last item in array
+	int calc_token_in_array_size(bdecode_token const& t)
```
arvidn: this looks a lot more reasonable. I want to try to come up with a better name

arvidn: how about ``token_source_span()``? something to indicate it's the number of bytes this token covers in the source encoding.

ajax16384: ok

PR #1769, opened 2017-03-01
more robust iteration on torrent peer connections

aldenml:


src/torrent.cpp
```diff
@@ -1715,9 +1719,9 @@ namespace libtorrent
 			// is available
 			// copy the peer list since peers may disconnect and invalidate
 			// m_connections as we initialize them
-			std::vector<peer_connection*> peers = m_connections;
-			for (auto* pc :  peers)
+			for (auto c : clone(m_connections))
 			{
+				auto pc = c->self();
```
aldenml: this is necessary (I think) because `pc` is used after `pc->on_metadata_impl();`

src/torrent.cpp
```diff
@@ -3804,13 +3806,10 @@ namespace libtorrent
 		// since this piece just passed, we might have
 		// become uninterested in some peers where this
 		// was the last piece we were interested in
-		for (peer_iterator i = m_connections.begin();
-			i != m_connections.end();)
+		// update_interest may disconnect the peer and
+		// invalidate the iterator
+		for (auto p : clone(m_connections))
 		{
-			peer_connection* p = *i;
```
aldenml: here, `->self()` is not necessary (I think), because `p->update_interest();` is at the end of the loop

src/torrent.cpp
```diff
@@ -124,6 +124,10 @@ namespace libtorrent
 		return ret;
 	}
 
+	template <typename T, typename IndexType = int>
+	aux::vector<T, IndexType> clone(aux::vector<T, IndexType> const& v)
+	{ return {v.begin(), v.end()}; }
```
arvidn: I think I would prefer just: ``return v;`` here.

arvidn:
lgtm. just one comment

src/torrent.cpp
aldenml: sure, me too

aldenml:
I discovered other loops that requires the `clone`, but that's for another PR

PR #1770, opened 2017-03-01
use inline is_digit

pavel-pimenov:


arvidn:
the reason I have a custom ``numeric`` function in there is to make ``bdecode`` less tied to libtorrent utility functions. Every few years or so I copy the bdecoder out into its own github repo.

at some point in the future I'm hoping to be able to make libtorrent use it via a submodule (to improve modularity)

arvidn:
this is not the case for lazy_decode though, and I'm fine with that patch

src/bdecode.cpp
```diff
@@ -48,7 +48,7 @@ namespace libtorrent
 
 	namespace
 	{
-	bool numeric(char c) { return c >= '0' && c <= '9'; }
+	inline bool numeric(char c) { return c >= '0' && c <= '9'; }
```
arvidn: this "inline" is redundant. this function has internal linkage already

PR #1772, opened 2017-03-02
mark connection as upload only manually instead of with set_upload_on…

ssiloti:
…ly()

We can't call set_upload_only() in this context because it can trigger an invariant
check in the torrent and m_num_connecting is not consistent.

### DIFF
--- a/src/web_connection_base.cpp
+++ b/src/web_connection_base.cpp
@@ -113,7 +113,11 @@ namespace libtorrent
 
 	void web_connection_base::start()
 	{
-		set_upload_only(true);
+		// avoid calling torrent::set_seed because it calls torrent::check_invariant
+		// which fails because the m_num_connecting count is not consistent until
+		// after we call peer_connection::start
+		m_upload_only = true;
+		disconnect_if_redundant();
 		if (is_disconnecting()) return;
 		peer_connection::start();
 	}


src/web_connection_base.cpp
```diff
@@ -113,7 +113,11 @@ namespace libtorrent
 
 	void web_connection_base::start()
 	{
-		set_upload_only(true);
```
arvidn: I seem to recall there's some accounting in the torrent of the number of connected seeds, won't this cause discrepancies there?

arvidn: actually, I don't see where check_invariant() is called from when calling set_seed(), in RC_1_1

arvidn: but I suspect the invariant check either shouldn't be called there, or this invariant shouldn't be checked

ssiloti: The seed accounting is done based on `torrent_peer::seed`, as long as we only change `m_upload_only` it doesn't affect that. Eventually everything gets straitened out when the connection is established and `web_connection_base::on_connected` calls `incoming_have_all` which calls `set_seed`.

The call to `check_invariant` is way down the call stack:

`set_upload_only` -> `set_seed` -> `update_auto_sequential` -> `num_downloaders`

PR #1774, opened 2017-03-02
more review of m_connections loops when the iterator could be invalidated

aldenml:


arvidn:
your last patch made me think about this. the clone function is clearly a lot nicer than what's there right now, but in some sense it's still error prone. You have to remember to call it when iterating over peers and they may be disconnected.

I think a more robust solution would be to simply not remove peers from the peer list immediately when disconnected. I can think of two (somewhat similar) ways of doing this.

1. in peer_connection::disconnect(), factor out the piece of code that actually removes the peer from ``m_connections`` and probably from the session connection list. Then post a call to this function to the message queue.

2. To better handle cases where a lot of peers are getting disconnected in one go, a slightly more optimized variant would be to keep a set of peers to be disconnected in the torrent object. Every time a peer is added to this list, post a call to a function to actually go through the list and disconnect them. Unless such call has already been posted, and is still in-flight, in this case we don't need to do anything more. (this is actually a common pattern, it's used for recalculating unchoke_slots, interest, queuing, and probably some other things too).

There's a risk that there are other operations that can mutate the peer list while iterating over it, and I think those should be fixed as they're encountered. I think this approach would be more robust going forward, as it would be less likely that bugs like the ones ``clone()`` solves will be introduced.

aldenml:
agree with your comment, in fact I needed to jump inside the functions in every loop to see if indeed disconnect was possible. I want to explore option 2) but that could take some time.

Do you want to close this PR, or use it in the meantime? (I don't mind closing it).

arvidn:
I think it would make sense to close this then.

For (2), I would encourage you to try something a little bit more reusable, like:

```
template <typename Handler>
struct handler_wrapper
{
   handler_wrapper(bool& in_flight, Handler&& h) : m_handler(std::move(h)), m_in_flight(in_flight) {}
   template <typename... Args>
   void operator()(Args&&... a)
   {
      TORRENT_ASSERT(m_in_flight);
      m_in_flight = false;
      m_handler(std::forward<Args>(a)...);
   }
   Handler m_handler;
   bool& in_flight;
};

struct deferred_handler
{
   template <typename Handler>
   void post(io_service& ios, Handler&& h)
   {
      ios.post(handler_wrapper<Handler>(m_in_flight, h));
   }
   bool m_in_flight = false;
};
```

I didn't try to compile that, but I think you get the idea. With a utility like this, this pattern could be made a bit more regular and recognizable.

aldenml:
hehe, I think I _will_ get the idea

arvidn:
I missed some key pieces of logic there. I'm updating it in-place

aldenml:
Hi @arvidn, I reopened the PR with another alternative: just iterate in reverse order.

This should be enough, because if an element is removed, it's from the end and the vector reverse iterator is not invalidated.

Now, I studied your `deferred_handler` idea, I think it's a very good idea, but the refactor will be very very dangerous.

I detected more loops in need of this, but I wanted to know if this is an aceptable solution.

src/torrent.cpp
```diff
@@ -124,9 +124,9 @@ namespace libtorrent
 		return ret;
 	}
 
-	template <typename T, typename IndexType = int>
-	aux::vector<T, IndexType> clone(aux::vector<T, IndexType> const& v)
-	{ return v; }
+	template <typename T, typename IndexType>
+	aux::reverse_const_wrapper<T> safe(aux::vector<T, IndexType> const& v)
```
arvidn: this should probably be called ``reverse``, no?

aldenml:
Also, still remain the issue that the call to `->self()` is at discretion.

src/torrent.cpp
aldenml: I wanted to differentiate, this "`safe`" from `aux::reverse`, but I know it's a little odd the naming

aldenml:
I see some iterators errors in the CI jobs, but I would like to have your opinion before fixing them

arvidn:
I don't see how my proposed approach is dangerous. Obviously I haven't tried to make the change, so I'm sure there's something I'm missing. I would think that it would make all of these loops automatically safe, with only a fairly local change (and then the loops could be simplified to just be a plain range-for loop).

aldenml:
Your approach is better, "dangerous" is the refactor I could do with my lack of understanding of what exactly is needed to defer. I get lost with the amount of logic in `peer_connection::disconnect` and `torrent::remove_peer`. I appreciate if you take a look at it.

Closing this again.

arvidn:
while looking at this, there is an edge case when removing a torrent or shutting down, where previously all peers would be gone instantly, now they linger a bit. I'm still optimistic I'll come up with a robust way of solving it though (and shutdown is tricky by itself either way)

arvidn:
ok, shutdown fixed. next issue is fast-reconnect, which also expect peers to be disconnected immediately

aldenml:
ok....

arvidn:
alright, finally: https://github.com/arvidn/libtorrent/pull/1782

PR #1775, opened 2017-03-03
more sign warnings fixes

aldenml:


ssiloti:
Why disable netlink on Android? It's rather important that we get network change notifications on Android since obviously they happen rather frequently on that platform.

aldenml:
For what I know, netlink support in android is very limited due to (not only) security concerns. I also see that the headers differ from the linux standard.

Before reverting that change, do you test it in android?

src/file.cpp
```diff
@@ -593,7 +593,7 @@ namespace libtorrent
 	{
 		ec.clear();
 		error_code e;
-		file_status s;
+		file_status s = {};
```
arvidn: what would you think about adding initializers to the members of this class instead?

arvidn:
In some sense it seems more conservative to leave it in until there's a good reason to remove. And, if it should be removed, I think there should be a comment in ``config.hpp``, where the macro used to be, explaining why it's not enabled for android.

Otherwise I imagine someone will just put it back at some point

ssiloti:
I don't test on Android but I'd rather keep it on unless someone reports that it doesn't work.

src/file.cpp
aldenml: Ok, I will do it

aldenml:
Ok, I will revert the `TORRENT_USE_NETLINK` change.

Thinking about adding something related to android in travis.

src/enum_net.cpp
```diff
@@ -1052,8 +1052,8 @@ namespace libtorrent
 		nl_msg->nlmsg_len = NLMSG_LENGTH(sizeof(rtmsg));
 		nl_msg->nlmsg_type = RTM_GETROUTE;
 		nl_msg->nlmsg_flags = NLM_F_DUMP | NLM_F_REQUEST;
-		nl_msg->nlmsg_seq = seq++;
-		nl_msg->nlmsg_pid = getpid();
+		nl_msg->nlmsg_seq = std::uint32_t(seq++);
```
arvidn: would it work to make ``seq`` ``uint32_t``?

aldenml: that would require change in `read_nl_sock`, do you prefer that?

arvidn: only if it would lower the total amount of casting I suppose

aldenml: I will remove this cast to speed up the merge

arvidn:
lgtm

PR #1776, opened 2017-03-04
add write_resume_data() function

arvidn:
Make resume data alert use an add_torrent_params object

### DIFF
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -975,7 +975,7 @@ namespace libtorrent
 	{
 		// internal
 		save_resume_data_alert(aux::stack_allocator& alloc
-			, std::shared_ptr<entry> const& rd
+			, add_torrent_params params
 			, torrent_handle const& h);
 
 		TORRENT_DEFINE_ALERT_PRIO(save_resume_data_alert, 37)
@@ -983,8 +983,15 @@ namespace libtorrent
 		static const int static_category = alert::storage_notification;
 		virtual std::string message() const override;
 
+		// the ``params`` structure is populated with the fields to be passed to
+		// add_torrent() or async_add_torrent() to resume the torrent. To
+		// save the state to disk, you may pass it on to write_resume_data().
+		add_torrent_params params;
+
+#ifndef TORRENT_NO_DEPRECATE
 		// points to the resume data.
-		std::shared_ptr<entry> const resume_data;
+		std::shared_ptr<entry> resume_data;
+#endif
 	};
 
 	// This alert is generated instead of ``save_resume_data_alert`` if there was an error
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -5937,9 +5937,9 @@ namespace libtorrent
 		using namespace libtorrent::detail; // for write_*_endpoint()
 
 		// cast to seconds in case that internal values doesn't have ratio<1>
-		ret.active_time = total_seconds(active_time());
-		ret.finished_time = total_seconds(finished_time());
-		ret.seeding_time = total_seconds(seeding_time());
+		ret.active_time = static_cast<int>(total_seconds(active_time()));
+		ret.finished_time = static_cast<int>(total_seconds(finished_time()));
+		ret.seeding_time = static_cast<int>(total_seconds(seeding_time()));
 		ret.last_seen_complete = m_last_seen_complete;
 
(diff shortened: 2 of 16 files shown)


arvidn:
This closes the circle, creating symmetry with the ``read_resume_data()`` function. With this patch resume data is returned and accepted in the form of an ``add_torrent_params`` object. This makes it easier to access and build programmatically, giving more control over both writing and reading resume data.

aldenml:
it's painful the sign warnings, right :)

arvidn:
I'm just surprised that my version of clang seems to emit fewer warnings than the version on travis. I have a newer version.

arvidn:
heh, those warnings are nothing compared to how annoying submodules are in git (i.e. that they don't update when you switch branch)

arvidn:
do you think this change makes sense? It just occurred to me a few days ago that resume data would be a lot more useful if easily accessible programmatically, not just when loading but also when saving. Specifically modifying it before saving and after loading. This also makes it possible to save resume data in different formats (in a database for instance), which may be more efficient

aldenml:
sure, it looks like a good idea to me (I didn't review the PR, btw).

include/libtorrent/alert_types.hpp
```diff
@@ -975,16 +975,23 @@ namespace libtorrent
...
 
 		TORRENT_DEFINE_ALERT_PRIO(save_resume_data_alert, 37)
 
 		static const int static_category = alert::storage_notification;
 		virtual std::string message() const override;
 
+		// the ``params`` structure is populated with the fields to be passed to
+		// add_torrent() or async_add_torrent() to resume the torrent. To
+		// save the state to disk, you may pass it on to write_resume_data().
+		add_torrent_params params;
```
aldenml: can (or should) be this `const`?

src/torrent.cpp
```diff
@@ -5934,148 +5934,105 @@ namespace libtorrent
...
+		ret.version = LIBTORRENT_VERSION_NUM;
+		ret.storage_mode = storage_mode();
+		ret.total_uploaded = m_total_uploaded;
+		ret.total_downloaded = m_total_downloaded;
 
-		ret["active_time"] = total_seconds(active_time());
-		ret["finished_time"] = total_seconds(finished_time());
-		ret["seeding_time"] = total_seconds(seeding_time());
-		ret["last_seen_complete"] = m_last_seen_complete;
+		using namespace libtorrent::detail; // for write_*_endpoint()
```
aldenml: I imagine this is not needed anymore in this function

src/write_resume_data.cpp
```diff
@@ -0,0 +1,269 @@
...
+			std::back_insert_iterator<entry::string_type> ptr(ret["peers"].string());
+#if TORRENT_USE_IPV6
+			std::back_insert_iterator<entry::string_type> ptr6(ret["peers6"].string());
+#endif
+			for (auto const& p : atp.peers)
+			{
+#if TORRENT_USE_IPV6
+				if (p.address().is_v6())
+				{
+					write_address(p.address().to_v6(), ptr6);
```
aldenml: I was asking myself the other day if these address writes can be replaced by `detail::write_endpoint` even if there is a redundant `if` inside

aldenml:
Question: What's the reason for not store all the params's vectors and bitfields in the allocator?

include/libtorrent/alert_types.hpp
arvidn: it could be const, but I've avoided it in other alerts since, technically, it may be beneficial for clients to *move* values out of alerts, and that may apply here too

arvidn:
I think there's value in literally having an ``add_torrent_params`` object, since that's what you pass to add a torrent. I can't think of a way to have both (an actual ``add_torrent_params`` object and have its members be allocated by the alert's stack allocator)

src/write_resume_data.cpp
arvidn: you mean to move the decision to write to ``ptr6`` or ``ptr`` into ``write_endpoint()``? (that seems like it could be a bit messy). Or do you mean to replace the two calls ``write_address()`` and ``write_uint16()`` with a single call to ``write_endpoint()``? Which seems reasonable to me. Assuming the function is inlined, that condition would be optimized out anyway.

aldenml: the second

> write_address() and write_uint16() with a single call to write_endpoint()

PR #1777, opened 2017-03-04
use span<> to return the block_info for a piece

arvidn:
to improve range checks and loops

### DIFF
--- a/include/libtorrent/span.hpp
+++ b/include/libtorrent/span.hpp
@@ -111,13 +111,6 @@ namespace libtorrent
 			return m_ptr[idx];
 		}
 
-		T& operator[](int const idx) const
-		{
-			TORRENT_ASSERT(idx >= 0);
-			TORRENT_ASSERT(static_cast<std::size_t>(idx) < m_len);
-			return m_ptr[static_cast<std::size_t>(idx)];
-		}
-
 	private:
 		T* m_ptr;
 		std::size_t m_len;
--- a/include/libtorrent/span.hpp
+++ b/include/libtorrent/span.hpp
@@ -105,10 +105,17 @@ namespace libtorrent
 			return { data() + offset, count };
 		}
 
-		T& operator[](std::size_t const idx)
+		T& operator[](std::size_t const idx) const
 		{
 			TORRENT_ASSERT(idx < m_len);
 			return m_ptr[idx];
+		}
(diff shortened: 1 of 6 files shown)


include/libtorrent/span.hpp
```diff
@@ -105,12 +105,19 @@ namespace libtorrent
...
 		}
 
-		T& operator[](std::size_t const idx)
+		T& operator[](std::size_t const idx) const
 		{
 			TORRENT_ASSERT(idx < m_len);
 			return m_ptr[idx];
 		}
 
+		T& operator[](int const idx) const
```
aldenml: I didn't review completely the PR, but I see `span<T>` like a very interoperable type, similar to `std::vector<T>`, and that means classic `std::size_t` index. Maybe you can use `typed_span<T, I>` in the place you need to index by `int`, similar to `aux::vector`.

arvidn: yeah, I thought about that a bit. you mean that ``span`` should be ready to be replaced by gsl or some other standard type? so this utility feature may get in the way. I agree with that, they should be sepate, I just wasn't sure it was worth separating them up-front, rather then put it off until it's actually needed

aldenml: yes, and `aux::typed_span` is already there

arvidn: oh, I see

aldenml: looking at it, I believe I missed the `IndexType` in the return type of the methods.

src/torrent.cpp
```diff
@@ -1855,30 +1855,25 @@ namespace libtorrent
 			{
 				std::vector<std::shared_ptr<torrent>> ts = m_ses.find_collection(c);
 
-				for (std::vector<std::shared_ptr<torrent>>::iterator k = ts.begin()
-					, end2(ts.end()); k != end2; ++k)
+				for (auto const& t : ts)
```
aldenml: sometimes I wonder how exactly lists of `std::shared_ptr` should be iterated, by ref, or by copy (just a comment)

arvidn: by reference. there's no need to incur the cost of atomic increment and decrement for the loop

aldenml: thanks!

PR #1782, opened 2017-03-06
don't remove peers from the connection list immediately when disconnecting

arvidn:
but defer it until later. This makes it less error prone to disconnect peers in loops over the connections

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -6417,7 +6417,8 @@ namespace libtorrent
 		{
 			// this asserts that we don't have duplicates in the peer_list's peer list
 			peer_iterator i_ = std::find_if(m_connections.begin(), m_connections.end()
-				, [peerinfo] (peer_connection const* p) { return p->remote() == peerinfo->ip(); });
+				, [peerinfo] (peer_connection const* p)
+				{ return !p->is_disconnecting() && p->remote() == peerinfo->ip(); });
 #if TORRENT_USE_I2P
 			TORRENT_ASSERT(i_ == m_connections.end()
 				|| (*i_)->type() != connection_type::bittorrent
@@ -7168,8 +7169,8 @@ namespace libtorrent
 		for (auto p : m_connections)
 			TORRENT_ASSERT(m_ses.has_peer(p));
 #endif
-		std::vector<peer_connection*> to_disconnect;
-		to_disconnect.resize(static_cast<std::size_t>(num));
+		aux::vector<peer_connection*> to_disconnect;
+		to_disconnect.resize(num);
 		auto end = std::partial_sort_copy(m_connections.begin(), m_connections.end()
 			, to_disconnect.begin(), to_disconnect.end(), compare_disconnect_peer);
 		for (auto p : range(to_disconnect.begin(), end))
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -95,6 +95,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/disk_interface.hpp"
 #include "libtorrent/broadcast_socket.hpp" // for is_ip_address
 #include "libtorrent/hex.hpp" // to_hex
+#include "libtorrent/aux_/range.hpp"
 // TODO: factor out cache_status to its own header
 #include "libtorrent/disk_io_thread.hpp" // for cache_status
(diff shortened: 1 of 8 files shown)


arvidn:
@aldenml heh, this was a bit tricker than I though. I think it's a more long-term approach though. I could (had to really) un-hack a few loops that assumed ``m_connections`` to be one smaller every time a peer got disconnected. I think the new code is a lot more robust though.

There are still some wrinkles around the accounting, it would be nice to improve that at some point

src/torrent.cpp
```diff
@@ -5399,8 +5368,27 @@ namespace libtorrent
...
+		TORRENT_ASSERT(is_single_thread());
+		INVARIANT_CHECK;
+
+		std::vector<peer_connection*> peers;
+		m_peers_to_disconnect.swap(peers);
+		for (peer_connection* p : peers)
+		{
+			TORRENT_ASSERT(p != nullptr);
+
+			auto const i = sorted_find(m_connections, p);
```
aldenml: I wonder if having both sorted `m_connections` and `m_peers_to_disconnect` reduces the `O(n^2)`, or if it does't worth the effort.

src/torrent.cpp
```diff
@@ -7199,35 +7159,25 @@ namespace libtorrent
...
-			TORRENT_ASSERT(m_ses.has_peer(*i));
-		}
+		// make sure we don't have any dangling pointers
+		for (auto p : m_connections)
+			TORRENT_ASSERT(m_ses.has_peer(p));
 #endif
-		int ret = 0;
-		while (ret < num && !m_connections.empty())
+		std::vector<peer_connection*> to_disconnect;
+		to_disconnect.resize(static_cast<std::size_t>(num));
```
aldenml: you can save the `static_cast` using `aux::vector`

src/torrent.cpp
```diff
@@ -5334,30 +5304,29 @@ namespace libtorrent
...
 
-		auto const i = sorted_find(m_connections, p);
-		if (i == m_connections.end())
-		{
-			TORRENT_ASSERT_FAIL();
-			return;
-		}
+		std::weak_ptr<torrent> weak_t = shared_from_this();
+		m_peers_to_disconnect.push_back(p);
+		m_deferred_disconnect.post(m_ses.get_io_service(), [=]() {
```
aldenml: Now that I see the code, maybe `deferred_handler` could have some `should_post` and save the creation of the handler, that I suppose it's not free, or is there some magic in the `&&`?

aldenml:
Now, I have a big question (at least for me): When to call (or not) `->self()` in the connection inside the loop(s)? because `m_ses.close_connection(this);` can destroy the object, or am I wrong?

And this is definitely a lot better :)

src/torrent.cpp
arvidn: I would expect the ``post()`` function to be inlined, and most work of creating the handler could probably be moved by the optimizer. The one thing I don't think could be done conditionally though is the creation of the ``shared_ptr`` and ``subsequently a ``weak_ptr``

src/torrent.cpp
arvidn: yeah, this will be a bit less efficient. The lookup here though is *log n*, so it's not quite as bad as n^2

arvidn:
yeah, that can happen if a peer_connection is not associated with a torrent, so it should be rare. Especially inside torrent.cpp itself, it shouldn't be possible

arvidn:
what I mean is that ``m_ses.close_connection()`` could be called when disconnecting a peer, and, yes, that does destroy the peer_connection object. However, it only happens for peers that aren't associated with a torrent

aldenml:
I see, and I also see that you moved `m_ses.close_connection(this);` to inside the `else` (instead of calling it always at the end).

Would you say that it's safe to remove all `->self()` in the `m_connections` loops inside `torrent.cpp`?

arvidn:
I believe that would be safe with this patch, yes.

With this patch, the call to ``m_ses.close_connection()`` is also deferred (when it can be). That's why it moved

src/torrent.cpp
arvidn: you're right that keeping ``m_peers_to_disconnect`` sorted would improve the complexity to O(n) though. maybe that's a worthwhile future patch.

PR #1783, opened 2017-03-06
loop conversion #1

ajax16384:


src/torrent.cpp
```diff
@@ -9586,7 +9558,7 @@ namespace libtorrent
 		// now, iterate over all time critical pieces, in order of importance, and
 		// request them from the peers, in order of responsiveness. i.e. request
 		// the most time critical pieces from the fastest peers.
-		for (std::vector<time_critical_piece>::iterator i = m_time_critical_pieces.begin()
+		for (auto i = m_time_critical_pieces.begin()
```
arvidn: if you would introduce a ``bool first`` variable, I think you could make this a range-for loop too

src/torrent.cpp
```diff
@@ -7690,16 +7665,14 @@ namespace libtorrent
 			if (p.peer_info_struct() && p.peer_info_struct()->seed)
 				++seeds;
 
-			for (std::vector<pending_block>::const_iterator j = p.request_queue().begin()
-				, end(p.request_queue().end()); j != end; ++j)
+			for (auto const& j : p.request_queue())
```
arvidn: there are some conflicts here I believe, with a patch that I have in flight. and I'm afraid I intend to land mine first

src/torrent.cpp
```diff
@@ -3419,11 +3411,10 @@ namespace libtorrent
 		std::vector<file_slice> files = fs.map_block(
 			p.piece_index, offset, (std::min)(piece_size - offset, block_size()));
 		std::int64_t ret = 0;
-		for (std::vector<file_slice>::iterator i = files.begin()
-			, end(files.end()); i != end; ++i)
+		for (auto const& i :  files)
```
arvidn: would you mind removing one of the spaces in front of "files"?

src/torrent.cpp
ajax16384: ok. i have already noticed your PR and skipped peer_connection

src/torrent.cpp
ajax16384: ok

src/torrent.cpp
ajax16384: oops. ok

src/torrent.cpp
ajax16384: this one https://github.com/arvidn/libtorrent/pull/1782 ?

arvidn:
yes, that was the PR. I just merged it, please rebase

ajax16384:
ok

src/torrent.cpp
```diff
@@ -9515,8 +9492,8 @@ namespace libtorrent
 		// now, iterate over all time critical pieces, in order of importance, and
 		// request them from the peers, in order of responsiveness. i.e. request
 		// the most time critical pieces from the fastest peers.
-		for (std::vector<time_critical_piece>::iterator i = m_time_critical_pieces.begin()
-			, end(m_time_critical_pieces.end()); i != end; ++i)
+		bool non_first_piece{false};
```
arvidn: this is a bit of a nit-pick, but would you mind negating the name and meaning of this? It just feels like a double-negation right now

arvidn:
lgtm, apart from comment

src/torrent.cpp
ajax16384: ok

arvidn:
lgtm

PR #1785, opened 2017-03-06
added log to on_ip_change and notify if ip_notifier is not supported

aldenml:


src/ip_notifier.cpp
```diff
@@ -91,7 +91,7 @@ namespace libtorrent
 				{ cb(error_code(err, system_category())); });
 		}
 #else
-		TORRENT_UNUSED(cb);
+		cb(make_error_code(boost::system::errc::not_supported));
```
aldenml: @ssiloti do you think this is a good idea?

src/session_impl.cpp
```diff
@@ -1763,6 +1763,10 @@ namespace aux {
 
 	void session_impl::on_ip_change(error_code const& ec)
 	{
+#ifndef TORRENT_DISABLE_LOGGING
+		session_log("received ip change from internal ip_notifier");
```
arvidn: I would prefer a single log, either an error or a success.

arvidn:
looks reasonable to me

aldenml:
@arvidn Are you open to add
```
target_link_libraries(torrent-rasterbar "-framework CoreFoundation" "-framework SystemConfiguration")
```
in a possible support for ip change notification in macOS?

arvidn:
yes. But I think it's important to be added to the makefile and Jamfile too.

src/ip_notifier.cpp
ssiloti: LGTM. It should probably do the same thing in the simulator case for consistency.

aldenml: I hope you don't mind another PR for that, CI is green already :)

btw, I think I have a solution for macOS

PR #1792, opened 2017-03-08
added support for ip_change_notifier in macOS

aldenml:


arvidn:
hm.. this looks more complicated than I recall

aldenml:
how...did you do something with the same goal in the past? I couldn't find any other solution that doesn't involve manual polling.

arvidn:
I think you could use ``SCNetworkReachabilityCreateWithAddress`` and ``SCNetworkReachabilitySetCallback``.

aldenml:
For what I see in the signature
```cpp
SCNetworkReachabilityRef SCNetworkReachabilityCreateWithAddress(CFAllocatorRef allocator, const struct sockaddr *address);
```
what should I use for `address`?

arvidn:
iirc, INADDR_ANY should work

aldenml:
Ok, I will try that in another branch

aldenml:
Hi @arvidn, before coding anything new (because I'm not a fan of `Cocoa`), I researched again and I want to confirm something first. What I'm trying to mirror is what's coded for Linux and Windows, and I believe is a notification on the routing table, not a notification on the networking (up or down).

What I want to confirm is if in fact, this is the desired behavior (monitor for routing table instead of networking). Also, maybe I'm wrong, but from the documentation and internet examples, `SCNetworkReachabilityCreateWithAddress` seems that is only about networking.

Btw, I just saw that for Windows, `NotifyAddrChange` only deals with `IPv4`, and...

> On Windows Vista and later, the NotifyIpInterfaceChange function can be used to register to be notified for changes to IPv4 and IPv6 interfaces on the local computer.

arvidn:
I'm not sure exactly which one is more accurate to subscribe to. I would expect if any NIC is enabled or disabled, both would trigger, and I would expect that to be the common scenario. Isn't there a way to subscribe to routing table changes though? instead of polling? (I haven't looked too closely at your code yet)

aldenml:
In my code, what I'm doing is subscribing to "something" that change when the routing table change, that is `SCDynamicStoreKeyCreateNetworkServiceEntity`. I did that because for Linux and Windows, that's the subscription to, and my "google" research only gave me this solution. Are you willing to change all? any?

I also hope that you can have some time to live test this notification in your environment, because the CI jobs don't do that.

aldenml:
@ssiloti @arvidn, the more I think about this feature, the more doubts come to my mind. Two of them are:

1- In case of multi NIC, If I listen to an specific address, the notify happens even if another NIC's status change, because the subscription (both in Windows and Linux) is to the routing table. Why do I need to reconnect to my actual connection? (event looking at the code is not clear to me that it will even succeed)

2- If the user is connected via VPN (with bind 0.0.0.0:0), when the notify happens because the VPN is disconnected, this automatic reopen isn't a security problem?

ssiloti:
The notification is only meant to signal that *something* changed. It's up to the callback to check what the state of things is and react accordingly. If nothing relevant has changed then the notification should be a no-op. Similarly, policies like forced VPN usage must be enforced by some other mechanism, it is outside of the scope of what the change notifier does.

aldenml:
I understand, but right now in `session_impl` (https://github.com/arvidn/libtorrent/blob/master/src/session_impl.cpp#L1750) it just reopen the sockets. No opportunity to ignore, or logic to do exists. What do you think it's appropriated to do here?

ssiloti:
Look in the body of `reopen_listen_sockets`. It doesn't blindly re-open all the sockets, it leaves existing sockets alone if they still match a valid endpoint.

aldenml:
Right, I didn't notice the detail that `listen_interface_t::device` could be an "address". Together with `partition_listen_sockets` I see it now. Thanks!

One more question: Now that `libtorrent` have this feature embedded, do you think creating an alert for clients to use is convenient?

ssiloti:
There's already a `listen_succeeded_alert`, I'm not sure what else you would want.

aldenml:
Well, there is this comment `// TODO only post alerts for new sockets?`, if that ever is implemented, then there is no opportunity to know what happened in some cases (assuming reusing the internal notification is a good idea). Is a different alert for "reopen_listen_sockets was called" reasonable? Or maybe some extension point via plugin is better?

ssiloti:
No, there is no need to notify the client unconditionally when `reopen_listen_sockets` is called.

src/ip_notifier.cpp
```diff
@@ -74,10 +222,29 @@ namespace libtorrent
...
 #if defined TORRENT_BUILD_SIMULATOR
-		TORRENT_UNUSED(cb);
+		// TODO: simulator support
+		cb(make_error_code(boost::system::errc::not_supported));
 #elif TORRENT_USE_NETLINK
+		using namespace std::placeholders;
 		m_socket.async_receive(boost::asio::buffer(m_buf)
 			, std::bind(&ip_change_notifier::on_notify, this, _1, _2, cb));
+#elif TORRENT_USE_SYSTEMCONFIGURATION
+		m_cb = std::move(cb);
```
ssiloti: It would be better to capture the callback directly rather than assigning it to a member and capturing `this`. That way if there are multiple callers they won't clobber each other's callbacks.

aldenml: I will try to do that, I need to look more into the safety of passing `std::function` as pointers

aldenml: I looked at this, and I can't convert a lambda with capture to a function pointer. And I coded this in a way that once the subscription is done, you can only change the "callback member", not safe to use from multiple clients.

arvidn:
I tested this on macos, and I can't get it to notify me at all. It occurs to me, don't you need to spin up a thread to actually run the Cocoa runLoop?

aldenml:
Well, now that you mention it, I tested in a UI app that I know it runs a regular Cocoa loop (main default). Let me review this again to take that in consideration.

aldenml:
Back to this issue, I need more feedback from you.

@arvidn I assembled a `SCNetworkReachabilityCreateWithAddress, SCNetworkReachabilitySetCallback` based solution, working with a separated `std::thread` to run `CFRunLoopRun()`. It works, but as expected, it only notify if the "default NIC" changes status. I tested it in my PC with two links (wired and wireless). The advantage of using these functions is that they are supported in `iOS`, but those devices don't have multi-home. What do you think is a good approach here?

@ssiloti Looking at the windows implementation, I see this in msdn:

> If the NotifyAddrChange function is called asynchronously, a notification will be sent on the next IPv4 address change until either the application cancels the notification by calling the CancelIPChangeNotify function or the application terminates.

But I see a call `NotifyAddrChange(&hnd, &m_ovl);` every time `ip_change_notifier::async_wait` is called with no corresponding `CancelIPChangeNotify`. I want to ask you if you think this is a potential problem or not.

arvidn:
I'm not really familiar with the API you're using in this patch now, but a cursory stack overflow search suggests that it's the right one.

aldenml:
To confirm (and clarify): do you mean using `SCNetworkReachabilityCreateWithAddress` and `SCNetworkReachabilitySetCallback` is the right path?

ssiloti:
Calling `NotifyAddrChange` multiple times with the same handle is probably a no-op, but `ip_change_notifier::on_notify` should still call `CancelIPChangeNotify`.

aldenml:
Closing this in favor of https://github.com/arvidn/libtorrent/pull/1891

PR #1793, opened 2017-03-08
using aux::array in metrics array and aux::vector in session_stats_metrics

aldenml:


src/session_stats.cpp
```diff
@@ -44,8 +47,8 @@ namespace libtorrent
 	};
 
 #define METRIC(category, name) { #category "." #name, counters:: name },
-	static const stats_metric_impl metrics[] =
-	{
+	aux::array<stats_metric_impl, counters::num_counters> const metrics
```
arvidn: if you remove ``static`` (which seems reasonable to do), you need to put it in the anonymous namespace

aldenml: right

src/session_stats.cpp
```diff
@@ -545,15 +548,14 @@ namespace libtorrent
...
+	}});
 #undef METRIC
 
 	std::vector<stats_metric> session_stats_metrics()
 	{
-		std::vector<stats_metric> stats;
-		int const num = sizeof(metrics) / sizeof(metrics[0]);
-		stats.resize(num);
-		for (std::size_t i = 0; i < num; ++i)
+		aux::vector<stats_metric> stats;
```
arvidn: will the return statement call the move constructor of ``std::vector``, despite slicing?

arvidn:
lgtm

src/session_stats.cpp
aldenml: I don't know, how can I verify this? (since this is not the only place when `aux::*` is in the return for an `std::*` type)

aldenml: I kind of assumed static cast to a base type

arvidn: yeah, it's probably fine. I imagine ``std::vector<stats_metric>(std::vector&& rhs);`` will be invoked, and ``aux::vector<stats_metric>`` is cast to a reference to its baseclass.

arvidn:
lgtm

Issue #1788, opened 2017-03-07
Question about "session stats header" log alert


aldenml:
libtorrent version (or branch): master
With the latest change
```cpp
if (m_alerts.should_post<log_alert>()
	|| m_alerts.should_post<session_stats_alert>())
```
now a `log_alert` is coming even if we don't expect one. I know this is a kind of especial log, but I find this printing of the header very fragile.

Is it posible to remove this "header construction" from `session_impl` and moving it to the tool that prints the stats? Where is that?

arvidn:
yeah, this is a bit of a hack. the tool parses this header to know what values to expect in the columns. It would be fragile to not tie this mapping to the values in the same log.

The fundamental problem is that this should only be logged when ``session_stats_alert``s are enabled, but it has to be logged as a ``log_alert``, since it's text.

I imagine it would be possible to only log it when session_stats_alerts are enabled, regardless of whether logs are enabled or not

aldenml:
Why not remove the code from `session_impl`? If that's possible, where to put it?

arvidn:
I consider it important that the name-mapping of the stats counters (i.e. this alert) is part of the same alert-stream, or log file, as the stats itself. If that's accepted, it needs to be logged somewhere, on session startup seems reasonable. I suppose the code to build the string, and maybe even create and post the alert could be factored out into the session_stats code

aldenml:
What do you think of having a method `session_stats_alert::headers_message()` that whatever tool is logging the alerts can use (at will)?

aldenml:
That would not requires any space in the "stack allocator"

PR #1794, opened 2017-03-08
created new session_stats_header_alert

aldenml:
@arvidn, before going to the tests, do you think this is a reasonable change to address #1788?

### DIFF
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -1875,6 +1875,12 @@ namespace libtorrent
 	// calling post_session_stats() on the session object. Its category is
 	// ``status_notification``, but it is not subject to filtering, since it's only
 	// manually posted anyway.
+	//
+	// the ``message()`` member function returns a string representation of the values that
+	// properly match the line returned in ``session_stats_header_alert::message()``.
+	//
+	// this specific output is parsed by tools/parse_session_stats.py
+	// if this is changed, that parser should also be changed
 	struct TORRENT_EXPORT session_stats_alert final : alert
 	{
 		session_stats_alert(aux::stack_allocator& alloc, counters const& cnt);
@@ -2498,7 +2504,7 @@ namespace libtorrent
 
 	// this alert is posted when the session encounters a serious error,
 	// potentially fatal
-	struct TORRENT_EXPORT session_error_alert : alert
+	struct TORRENT_EXPORT session_error_alert final : alert
 	{
 		// internal
 		session_error_alert(aux::stack_allocator& alloc, error_code err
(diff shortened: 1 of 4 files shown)


src/session_impl.cpp
```diff
@@ -588,30 +587,7 @@ namespace aux {
 		TORRENT_ASSERT(is_single_thread());
 
 #ifndef TORRENT_DISABLE_LOGGING
-		// this alert is a bit special. The stats headers aren't very useful
```
arvidn: I take it the purpose of this patch is to get rid of the posting of this alert, right?
I really think there's a benefit of having this be an alert that gets posted through the same path as the stats counters. The main reason is that any client that can log alert messages to a file will support being instrumented. There's no extra work on the client's side. I think this is especially useful because (in my experience), being able to instrument and trouble-shoot is something I'm interested in, but not client developers to the same extent. or at least it significantly lowers the hurdle to be able to provide stats.

I'm not sure I have a correct understanding of your concern of having this here. Is it that it creates noise in your alert logs? is it that it blows up the stack allocators by allocating a large string?

What would you think about introducing a separate alert just for this. It would have these benefits:

* It could belong to the stats category and not be mixed up with enabling logging.
* It could be completely stateless and its ``message()`` function could just generate this string (and most of this code would move in there)
* It could easily be identified by a client that wants to ignore it

aldenml: I want to get rid of the posting, because of the noise. In my particular case, I have code that it's prepared to print "log alerts" always, and I control that by putting the right value in the mask (or not), but now I always get this "headers". No other problem.

I kind of like the benefits of a new alert, even at the cost of creating a "new alert" type. Let me work on it and let's see how it looks.

Thanks for taking your time to explain me the benefits of having the message in the log stream, now I understand better how this "messages logging" can be used.

include/libtorrent/alert_types.hpp
```diff
@@ -2541,11 +2547,29 @@ namespace libtorrent
 		aux::allocation_slot m_v6_nodes_idx;
 	};
 
+	// The session_stats_header alert is posted during the init of the
+	// session thread.
+	//
+	// the ``message()`` method returns a string representation of the
```
arvidn: a bit of a nit-pick. in C++ terminology it's called "member function" no method :P

arvidn:
lgtm

aldenml:
good, fixing the `explicit` issue and the comment :)

arvidn:
great, thanks!

arvidn:
I'd be happy to merge this once you rebase

aldenml:
done

PR #1798, opened 2017-03-09
Improve Documantation

Falcosc:
If you don't have any references to http://www.rasterbar.com/products/libtorrent/manual.html anymore, you should delete it otherwise you should update this page with the manual of the master branch.

Maybe this should be marged to RC_1_1 as well

### DIFF
--- a/docs/tuning.rst
+++ b/docs/tuning.rst
@@ -163,31 +163,11 @@ send buffer watermark
 The send buffer watermark controls when libtorrent will ask the disk I/O thread
 to read blocks from disk, and append it to a peer's send buffer.
 
-``settings_pack::send_buffer_low_watermark`` defines the minimum send buffer
-target size (send buffer includes bytes pending being read from disk). For good
-and snappy seeding performance, set this fairly high, to at least fit a few
-blocks. This is essentially the initial window size which will determine how
-fast we can ramp up the send rate
-
 When the send buffer has fewer than or equal number of bytes as
 ``settings_pack::send_buffer_watermark``, the peer will ask the disk I/O thread
 for more data to send. The trade-off here is between wasting memory by having too
 much data in the send buffer, and hurting send rate by starving out the socket,
 waiting for the disk read operation to complete.
-
-The actual watermark may be lower than this in case the upload rate is low,
-``settings_pack::send_buffer_watermark`` defines the upper limit.
-The current upload rate to a peer is multiplied by this factor to get the send buffer
-watermark. The factor is specified as a percentage. i.e. 50 -> 0.5 This product is
-clamped to the ``settings_pack::send_buffer_watermark`` setting to not exceed the
-max. 100 -> factor 1 mean you never allocate more buffer than payload got
-transfered during last second.
-For high speed upload, this should be set to a greater value than 100. For high
-capacity connections, setting this higher can improve upload performance and disk
-throughput. Setting it too high may waste RAM and create a bias towards read jobs
-over write jobs. For example with factor 150 you could put 1.5 seconds worth of data in
-the send buffer this gives the disk I/O more heads-up on disk reads, and can maximize
-throughput.
 
 If your main objective is memory usage and you're not concerned about being able
 to achieve high send rates, you can set the watermark to 9 bytes. This will guarantee
@@ -196,11 +176,6 @@ all peers. This is the least amount of memory possible for the send buffer.
 
 You should benchmark your max send rate when adjusting this setting. If you have
 a very fast disk, you are less likely see a performance hit.
-
-If your main objective is maximum seeding performance you should consider a usecase
-where your harddisk get overloaded by your connection pool size. For example you could
-calculate with 500ms read time and a send rate of 6 MiB/s, the upper limit should be
-3 MiB.
 
 reduce executable size
 ----------------------
@@ -296,6 +271,11 @@ libtorrent uses a watermark for send buffers to determine when a new piece shoul
 be requested from the disk I/O subsystem, to be appended to the send buffer. The low
 watermark needs to be large enough to not draining the socket's send buffer before the
 disk operation completes.
+``settings_pack::send_buffer_low_watermark`` defines the minimum send buffer
+target size (send buffer includes bytes pending being read from disk). For good
+and snappy seeding performance, set this fairly high, to at least fit a few
+blocks. This is essentially the initial window size which will determine how
+fast we can ramp up the send rate
 But by carefull with this setting if you have many connections because it does define
 an absolut part of the memory footprint per connection.
 
@@ -306,6 +286,25 @@ The watermark is bound to a max value per connection, to avoid buffer sizes grow
 of control. The default max send buffer size might not be enough to sustain very high
 upload rates, and you might have to increase it. It's specified in bytes in
 ``settings_pack::send_buffer_watermark``.
+
+The actual watermark may be lower than this in case the upload rate is low,
+``settings_pack::send_buffer_watermark`` defines the upper limit.
+The current upload rate to a peer is multiplied by this factor to get the send buffer
+watermark. The factor is specified as a percentage. i.e. 50 -> 0.5 This product is
+clamped to the ``settings_pack::send_buffer_watermark`` setting to not exceed the
+max. 100 -> factor 1 mean you never allocate more buffer than payload got
+transfered during last second.
+For high speed upload, this should be set to a greater value than 100. For high
+capacity connections, setting this higher can improve upload performance and disk
+throughput. Setting it too high may waste RAM and create a bias towards read jobs
+over write jobs. For example with factor 150 you could put 1.5 seconds worth of data in
+the send buffer this gives the disk I/O more heads-up on disk reads, and can maximize
+throughput.
+
+If your main objective is maximum seeding performance you should consider a usecase
+where your harddisk get overloaded by your connection pool size. For example you could
+calculate with 500ms read time and a send rate of 6 MiB/s, the upper limit should be
+3 MiB.
 
 It's a little bit difficould to calculate overall maximum possible usage of send buffer.
 Because your low watermark definition does maybe waste memory because you hold more buffer
@@ -386,6 +385,14 @@ minimal set of torrents you are actually interested in.
 
 To get an update with only the torrents that have changed since last time, call
 ``session::post_torrent_updates()``.
+
+high latency networks
+=====================
+
+High network latencies have a huge impact on very fast connections. If you have a environment
+with high network latencies, you should check if they have an impact of your maximum
+throughput. Here are some information about using libtorrent in high latency networks:
+https://github.com/vandenheuvel/libtorrent-latency-benchmark
 
 benchmarking
 ============
--- a/docs/tuning.rst
+++ b/docs/tuning.rst
@@ -163,11 +163,31 @@ send buffer watermark
 The send buffer watermark controls when libtorrent will ask the disk I/O thread
 to read blocks from disk, and append it to a peer's send buffer.
 
+``settings_pack::send_buffer_low_watermark`` defines the minimum send buffer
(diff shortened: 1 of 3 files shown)


docs/tuning.rst
```diff
@@ -350,6 +386,14 @@ minimal set of torrents you are actually interested in.
 To get an update with only the torrents that have changed since last time, call
 ``session::post_torrent_updates()``.
 
+high latency networks
+=====================
+
```
Falcosc: I'm not sure about this. It's a very interesting topic but are there any fast networks with high latency out there? I know only networks with extrem proxy setups for privacy and I don't know any non suspicious usecases for large file transfere in these kind of networks.

docs/tuning.rst
```diff
@@ -267,16 +267,52 @@ part be squashed by the TCP traffic.
...
-libtorrent uses a low watermark for send buffers to determine when a new piece should
+libtorrent uses a watermark for send buffers to determine when a new piece should
 be requested from the disk I/O subsystem, to be appended to the send buffer. The low
-watermark is determined based on the send rate of the socket. It needs to be large
-enough to not draining the socket's send buffer before the disk operation completes.
-
-The watermark is bound to a max value, to avoid buffer sizes growing out of control.
-The default max send buffer size might not be enough to sustain very high upload rates,
-and you might have to increase it. It's specified in bytes in
+watermark needs to be large enough to not draining the socket's send buffer before the
```
thalieht: to avoid draining?

docs/tuning.rst
```diff
@@ -267,16 +267,52 @@ part be squashed by the TCP traffic.
...
+disk operation completes.
+``settings_pack::send_buffer_low_watermark`` defines the minimum send buffer
+target size (send buffer includes bytes pending being read from disk). For good
+and snappy seeding performance, set this fairly high, to at least fit a few
```
thalieht: to fit at least a few blocks?

docs/tuning.rst
```diff
@@ -267,16 +267,52 @@ part be squashed by the TCP traffic.
...
+blocks. This is essentially the initial window size which will determine how
+fast we can ramp up the send rate
+But by carefull with this setting if you have many connections because it does define
```
thalieht: "careful"
"because it defines an absolute part"

docs/tuning.rst
```diff
@@ -267,16 +267,52 @@ part be squashed by the TCP traffic.
...
+an absolut part of the memory footprint per connection.
+
+``settings_pack::send_buffer_watermark_factor`` does raise the minimum memory footprint
```
thalieht: does raise -> raises
Also something else seems wrong with this sentence but i don't understand the context.

docs/tuning.rst
```diff
@@ -267,16 +267,52 @@ part be squashed by the TCP traffic.
...
 ``settings_pack::send_buffer_watermark``.
 
+The actual watermark may be lower than this in case the upload rate is low,
+``settings_pack::send_buffer_watermark`` defines the upper limit.
+The current upload rate to a peer is multiplied by this factor to get the send buffer
+watermark. The factor is specified as a percentage. i.e. 50 -> 0.5 This product is
+clamped to the ``settings_pack::send_buffer_watermark`` setting to not exceed the
+max. 100 -> factor 1 mean you never allocate more buffer than payload got
+transfered during last second.
+For high speed upload, this should be set to a greater value than 100. For high
```
thalieht: set to a value greater than 100?

docs/tuning.rst
```diff
@@ -267,16 +267,52 @@ part be squashed by the TCP traffic.
...
+capacity connections, setting this higher can improve upload performance and disk
+throughput. Setting it too high may waste RAM and create a bias towards read jobs
+over write jobs. For example with factor 150 you could put 1.5 seconds worth of data in
+the send buffer this gives the disk I/O more heads-up on disk reads, and can maximize
+throughput.
+
+If your main objective is maximum seeding performance you should consider a usecase
+where your harddisk get overloaded by your connection pool size. For example you could
```
thalieht: harddisk (i think the "hard" is redundant)

docs/tuning.rst
```diff
@@ -267,16 +267,52 @@ part be squashed by the TCP traffic.
...
+calculate with 500ms read time and a send rate of 6 MiB/s, the upper limit should be
+3 MiB.
+
+It's a little bit difficould to calculate overall maximum possible usage of send buffer.
```
thalieht: difficult

docs/tuning.rst
```diff
@@ -267,16 +267,52 @@ part be squashed by the TCP traffic.
...
+Because your low watermark definition does maybe waste memory because you hold more buffer
+than needed by a slow connection. If you don't wast much memory the accumulated send buffer
```
thalieht: I don't think i've seen the documentation addressing the reader directly elsewhere. So:
"Because a low watermark (definition required?) can allocate a bigger buffer in memory than is required by a slow connection"?
I don't know how to rephrase the next sentence..

docs/tuning.rst
```diff
@@ -350,6 +386,14 @@ minimal set of torrents you are actually interested in.
 To get an update with only the torrents that have changed since last time, call
 ``session::post_torrent_updates()``.
 
+high latency networks
+=====================
+
+High network latencies have a huge impact on very fast connections. If you have a environment
```
thalieht: **an** environment

include/libtorrent/settings_pack.hpp
```diff
@@ -321,7 +321,9 @@ namespace libtorrent
 
 			// prefer seeding torrents when determining which torrents to give
 			// active slots to, the default is false which gives preference to
-			// downloading torrents
+			// downloading torrents. Be carefull if you want to prefer seeds,
```
thalieht: careful

include/libtorrent/settings_pack.hpp
```diff
@@ -1045,8 +1048,15 @@ namespace libtorrent
 			// torrents.
 			//
 			// ``active_limit`` is a hard limit on the number of active (auto
-			// managed) torrents. This limit also applies to slow torrents.
-			//
+			// managed) torrents. This limit also applies to slow torrents and it
+			// does apply to all announcement protocols as well. If you have a
+			// large amount of seeds, only a limited amount get annonced based on
```
thalieht: anno**u**nced

include/libtorrent/settings_pack.hpp
```diff
@@ -1045,8 +1048,15 @@ namespace libtorrent
...
+			// the seed rank. Use low values if you want to work with
+			// ``seed_time_ratio_limit``, ``seed_time_limit`` or
+			// ``share_ratio_limit`` to controll your active torrents
+			// ``auto_manage_prefer_seeds`` controlls which torrents will drain
```
thalieht: controls. Same for the above line

include/libtorrent/settings_pack.hpp
```diff
@@ -1093,7 +1103,9 @@ namespace libtorrent
 
 			// this is the limit on the time a torrent has been an active seed
 			// (specified in seconds) before it is considered having met the seed
-			// limit criteria. See queuing_.
+			// limit criteria. See queuing_. It does not disable the torrent, it
+			// does only move to a lower queue position which may result into an
```
thalieht: it only moves it to a lower queue?
it is only moved to a lower queue?

include/libtorrent/settings_pack.hpp
```diff
@@ -1503,7 +1515,9 @@ namespace libtorrent
 			// bytes down) or the seed time ratio (seconds as seed / seconds as
 			// downloader) or the seed time limit (seconds as seed) it is
 			// considered done, and it will leave room for other torrents these
-			// are specified as percentages
+			// are specified as percentages. They does not disable the torrent,
```
thalieht: **do** not disable the torrent

include/libtorrent/settings_pack.hpp
```diff
@@ -1503,7 +1515,9 @@ namespace libtorrent
...
+			// they do only controll the queue position which may result into an
```
thalieht: they only control the queue

docs/tuning.rst
arvidn: I would interpret "high latency" as latency / bandwidth. There are plenty of high-bandwidth networks where the latency poses a similar problem (as in the bandwidth delay product is higher than normal)

include/libtorrent/settings_pack.hpp
arvidn: I think the "be careful if you want to prefer seeds, because" is redundant and can be removed.

include/libtorrent/settings_pack.hpp
```diff
@@ -1045,8 +1048,15 @@ namespace libtorrent
 			// torrents.
 			//
 			// ``active_limit`` is a hard limit on the number of active (auto
-			// managed) torrents. This limit also applies to slow torrents.
-			//
+			// managed) torrents. This limit also applies to slow torrents and it
+			// does apply to all announcement protocols as well. If you have a
```
arvidn: this term is not obvious: "announcement protocols as well". I take it you refer to the limits on where to announce torrents. Maybe there's a more concise way to say that. Perhaps you could just drop " protocols as well" (since the start of the sentence is "This limit also applies to"). I would also shorten: "and it does apply to all announcements" -> "and to announcements"

include/libtorrent/settings_pack.hpp
arvidn: seeds come in "numbers" not "amounts". i.e. s/amount/number
also: "only a limited amount get announced based on" -> "only some are announced based on"

include/libtorrent/settings_pack.hpp
```diff
@@ -1045,8 +1048,15 @@ namespace libtorrent
...
+			// large amount of seeds, only a limited amount get annonced based on
+			// the seed rank. Use low values if you want to work with
```
arvidn: in general, I would stay away from "you". I think "Use low values if you want to work with" should be something like "Use low values when using".

include/libtorrent/settings_pack.hpp
arvidn: this sentence is a bit odd. Perhaps something like: "When this limit is hit, the torrent is not disabled, instead its seed rank is adjusted give priority to torrents that have not met their seed_time_limit"

include/libtorrent/settings_pack.hpp
arvidn: I don't think this should be described as a "queue" position. queue position is already a term used for the order in which downloading torrents are started. This affects the seed rank. This sentence should probably be phrased similar to the one above, for seed_time_limit

docs/tuning.rst
arvidn: I think "and snappy" is redundant and can be removed

docs/tuning.rst
arvidn: or "to not drain the socket's"

arvidn:
see: https://travis-ci.org/arvidn/libtorrent/jobs/209417002#L1044

Col-blimp:
Can I suggest something like this to the top of each file:

```
/*

Required/Optional

Description/purpose

*/
```

arvidn:
@Col-blimp what would it mean for a file to be optional or required?

Falcosc:
And which files do you mean?


By the way, my changes could be delayed a week or so, but I did look over all of you comments, thanks for the help.

Col-blimp:
Well it's quite a big library covering a lot of platforms and if your new to it it's quite daunting so what I was meaning was a short synopsis at the top of each .cpp/.h file saying what it's for would go a long way to help newcomers for example session.hpp is clear enough but move in to the aux_ folder you do wonder what cpuid.hpp is for as it's not included in examples to figure out.

aldenml:
can this PR be rescued for a potential 1.2 release?

PR #1804, opened 2017-03-11
loop conversion #2

ajax16384:
use auto loop
replace "erase(iter++)" with "iter = erase(iter)"

### DIFF
--- a/src/file_pool.cpp
+++ b/src/file_pool.cpp
@@ -247,8 +247,9 @@ namespace libtorrent
 		while (begin != end)
 		{
 			to_close.push_back(std::move(begin->second.file_ptr));
-			m_files.erase(begin++);
+			begin++;
 		}
+		m_files.erase(begin, end);
 		l.unlock();
 		// the files are closed here while the lock is not held
 	}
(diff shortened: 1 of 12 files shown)


src/file_pool.cpp
```diff
@@ -239,16 +239,18 @@ namespace libtorrent
 	{
 		std::unique_lock<std::mutex> l(m_mutex);
 
-		auto begin = m_files.lower_bound(std::make_pair(st, file_index_t(0)));
+		auto const begin = m_files.lower_bound(std::make_pair(st, file_index_t(0)));
+		auto current = begin;
```
arvidn: if you're introducing a loop variable, I think the loop below should be turned into a for-loop. (and ``current`` could be named ``i``)

arvidn:
lgtm apart from comment

src/file_pool.cpp
ajax16384: ok

arvidn:
lgtm

arvidn:
https://travis-ci.org/arvidn/libtorrent/jobs/210113639#L2096

arvidn:
lgtm

PR #1805, opened 2017-03-11
improve parallelism of connection_tester gen-torrent

arvidn:


examples/connection_tester.cpp
```diff
@@ -46,8 +46,9 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <thread>
 #include <functional>
 #include <iostream>
+#include <thread>
```
thalieht: 3 lines up :)

Issue #1803, opened 2017-03-11
boost 1.64: error: string_view from a temporary string


pavel-pimenov:
1>q:\vc15\r5xx\libtorrent\src\storage_utils.cpp(266): error C2280: 'boost::basic_string_view<char,std::char_traits<char>>::basic_string_view<std::allocator<char>>(std::basic_string<char,std::char_traits<char>,std::allocator<char>> &&)': attempting to reference a deleted function
1>q:\vc15\r5xx\boost\boost\utility\string_view.hpp(100): note: see declaration of 'boost::basic_string_view<char,std::char_traits<char>>::basic_string_view'
1>q:\vc15\r5xx\boost\boost\utility\string_view.hpp(100): note: 'boost::basic_string_view<char,std::char_traits<char>>::basic_string_view<std::allocator<char>>(std::basic_string<char,std::char_traits<char>,std::allocator<char>> &&)': function was explicitly deleted
1>session_impl.cpp

![default](https://cloud.githubusercontent.com/assets/920705/23820838/920e3b4e-0632-11e7-82ce-bd3bd96b3b44.png)

boost\boost\utility\string_view.hpp:

#if !defined(BOOST_NO_CXX11_RVALUE_REFERENCES) && !defined(BOOST_NO_CXX11_DELETED_FUNCTIONS)
      // Constructing a string_view from a temporary string is a bad idea
      template<typename Allocator>
        basic_string_view(      std::basic_string<charT, traits, Allocator>&&)
          = delete;
#endif

arvidn:
is this new in boost-1.64? is this the only place you get this error?

pavel-pimenov:
yes, only one error.
#define BOOST_NO_CXX11_DELETED_FUNCTIONS - all ok.

i use
https://sourceforge.net/projects/boost/files/boost/snapshots/master/boost_1_64_0-snapshot.zip/download
boost-git-dev
https://github.com/boostorg/utility/blob/00f02167e35861bd11b8a9d6585444e9ef0b5172/include/boost/utility/string_view.hpp#L98

SysMan-One:
Got the same compilation problem.

ajax16384:
https://github.com/arvidn/libtorrent/pull/1814

ajw107:
[...]

PR #1814, opened 2017-03-15
fix boost 1.64: error: string_view from a temporary string #1803

ajax16384:


ajax16384:
still needs "tests" to be fixed.
@arvidn what do you think about introduction combine_path alias with std::string args to reduce changes?

aldenml:
Don't this defeat the whole purpose of optimization using `string_view`? What about calling `.c_str()` ?

ajax16384:
overloaded "сombine_path" will allow not change many following lines:
````c++
combine_path("test", combine_path("c", "a")), 30000)
````
calling ".c_str()" still needs fix many "comine_path(..comine_path(..).c_str()"

**Comment by arvidn** [comment]
isn't there a way to allow the conversions? is the constructor explicit perhaps?
As far as I can tell, it *is* safe to construct a ``string_view`` from temporaries in all of these cases. The temporaries will live until the end of the statement.

**Comment by arvidn** [comment]
a possible other pattern would be to have overloads of combine_path that take ``std::string const&``

**Comment by ajax16384** [comment]
so i will try to add overload "combine_path"

**Comment by aldenml** [comment]
@ajax16384 it's good the overload `combine_path`, but what about the rest, isn't it better to just call `.c_str()`?

File: src/file.cpp [hunk 1]
```diff
@@ -937,6 +937,20 @@ namespace libtorrent
...
 	}
 
+	std::string combine_path(std::string const& ls, std::string const& rs)
+	{
+		string_view lhs{ls};
+		string_view rhs{rs};
+		return combine_path(lhs, rhs);
+	}
+
+	std::string combine_path(char const* ls, char const* rs)
```
aldenml: is this necessary?

ajax16384: yes it's necessary, comine_path("a", "b") shows ambiguity (can not decide which of string_view or std::string cont&  to use)

ajax16384:
instead of
```c++
std::string hash_str{ aux::to_hex(t->info_hash()) };
m_name_idx = alloc.copy_string(hash_str);
````
use:
````c++
m_name_idx = alloc.copy_string(aux::to_hex(t->info_hash()).c_str());
````
?

**Comment by aldenml** [comment]
yes

**Comment by ajax16384** [comment]
just checked - string_view from c_str() will use unoptimal length calculation (strlen). string_view from local std::string just get size()

**Comment by aldenml** [comment]
Well, what can I say, those temporary string creations just seem very odd. I would like to test, but `boost 1.64` is still beta and that will mess my env.

**Comment by ajax16384** [comment]
ok, let's wait for 1.64 release

File: src/file.cpp [hunk 2]
```diff
@@ -375,7 +375,12 @@ namespace libtorrent
 		{
 			std::string sep_path { path };
 			std::replace(sep_path.begin(), sep_path.end(), '/', '\\');
-			prepared_path = "\\\\?\\" + (is_complete(sep_path) ? sep_path : combine_path(current_working_directory(), sep_path));
+			if (!is_complete(sep_path))
+			{
+				std::string cwd_path { current_working_directory() };
+				sep_path = combine_path(cwd_path, sep_path);
+			}
+			prepared_path = "\\\\?\\" + sep_path;
```
arvidn: I think the new version here is objectively worse than the previous. Primarily because it's longer, but it also introduces a mutable variable. I take it the problem with the current version is that ``combine_path()`` is being called with one argument that's an rvalue string, and string_view won't construct from that anymore.

what I tried to ask in the comments earlier was, is that constructor deleted or just explicit? would this work?

``combine_path(string_view(current_working_directory()), sep_path)``

If so, I think that would be preferable, as it's optimal.

src/file.cpp
arvidn: this overload will also be sub-optimal as constructing the string views will have to run ``strlen()``. The proper overload would be:

``template <std::size_t Len1, std::size_t len2> std::string combine_path(char const (&ls)[Len1], char const (&rs)[Len2]);``

But doing this for you is the whole point of string_view in the first place.

src/file.cpp
ajax16384: no, it will not work (gives same "attempting to reference a deleted function")

arvidn: what do you think about a function like this then:

``string_view make_sv(std::string const& str) { return string_view(str); }`` ?

and say: ``combine_path(make_sv(current_working_directory()), sep_path)``

ajax16384: it's ok, will try in next commit

include/libtorrent/file.hpp
```diff
@@ -165,6 +165,10 @@ namespace libtorrent
 	TORRENT_EXTRA_EXPORT std::string filename(std::string const& f);
 	TORRENT_EXTRA_EXPORT std::string combine_path(string_view lhs
 		, string_view rhs);
+	TORRENT_EXTRA_EXPORT std::string combine_path(std::string const& ls
```
arvidn: with the ``make_sv`` function, do you still need these overloads?

src/storage_utils.cpp
```diff
@@ -263,8 +263,9 @@ namespace libtorrent { namespace aux
 			// files moved out to absolute paths are not moved
 			if (f.file_absolute_path(i)) continue;
 
-			std::string const old_path = combine_path(save_path, f.file_path(i));
-			std::string const new_path = combine_path(new_save_path, f.file_path(i));
+			std::string cur_file_path{ f.file_path(i) };
```
arvidn: cur_file_path can be const here

src/storage_utils.cpp
```diff
@@ -325,8 +326,9 @@ namespace libtorrent { namespace aux
 				// roll-back
 				if (copied_files[i]) continue;
 
-				std::string const old_path = combine_path(save_path, f.file_path(i));
-				std::string const new_path = combine_path(new_save_path, f.file_path(i));
+				std::string cur_file_path{ f.file_path(i) };
```
arvidn: this can also be const

src/storage_utils.cpp
```diff
@@ -355,7 +357,8 @@ namespace libtorrent { namespace aux
 			// need to delete the source.
 			if (copied_files[i] == false) continue;
 
-			std::string const old_path = combine_path(save_path, f.file_path(i));
+			std::string cur_file_path{ f.file_path(i) };
+			std::string const old_path = combine_path(save_path, cur_file_path);
```
arvidn: I imagine you could use ``make_sv`` here instead

test/test_bdecode.cpp
```diff
@@ -201,7 +201,8 @@ TORRENT_TEST(dict_null_key)
 	int ret = bdecode(b, b + sizeof(b)-1, e, ec);
 	TEST_EQUAL(ret, 0);
 	TEST_CHECK(e.dict_size() == 1);
-	bdecode_node d = e.dict_find(std::string("a\0b", 3));
+	std::string find_str("a\0b", 3);
+	bdecode_node d = e.dict_find(find_str);
```
arvidn: I think you can just skip the ``std::string`` entirely now. ``string_view`` should have a constructor from a string literal

test/test_bdecode.cpp
```diff
@@ -848,9 +849,12 @@ TORRENT_TEST(dict_find_funs)
 	TEST_CHECK(!e.dict_find_dict("c"));
 
 	// variants taking std::string
-	TEST_EQUAL(e.dict_find_dict(std::string("d")).dict_find_int_value("x"), 1);
-	TEST_CHECK(!e.dict_find_dict(std::string("c")));
-	TEST_CHECK(!e.dict_find_dict(std::string("x")));
+	std::string d_str("d");
+	std::string c_str("c");
+	std::string x_str("x");
+	TEST_EQUAL(e.dict_find_dict(d_str).dict_find_int_value("x"), 1);
```
arvidn: here too, I don't think you need the ``std::string`` at all

test/test_bdecode.cpp
```diff
@@ -949,7 +953,8 @@ TORRENT_TEST(dict_find_funs2)
 
 	// try finding the last item in a dict (to skip all the other ones)
 	TEST_EQUAL(e.dict_find("d").type(), bdecode_node::dict_t);
-	TEST_EQUAL(e.dict_find(std::string("d")).type(), bdecode_node::dict_t);
+	std::string d_str("d");
+	TEST_EQUAL(e.dict_find(d_str).type(), bdecode_node::dict_t);
```
arvidn: and here, try to just remove the ``std::string``

test/test_stack_allocator.cpp
```diff
@@ -44,7 +44,8 @@ TORRENT_TEST(copy_string)
 	// attempt to trigger a reallocation
 	a.allocate(100000);
 
-	allocation_slot const idx2 = a.copy_string(std::string("foobar"));
+	std::string foobar_str("foobar");
+	allocation_slot const idx2 = a.copy_string(foobar_str);
```
arvidn: and here. ``string_view`` should have a constructor from string literal

arvidn:
looks better. I think there are still a few improvements to be made

test/test_bdecode.cpp
ajax16384: this test block was made especially for std::string api call  "// variants taking std::string"

test/test_bdecode.cpp
ajax16384: here it also made for testing of passing std::string as argument (or else it must be completly removed cause line above do the same test but with "d" passed as const literal)

src/storage_utils.cpp
```diff
@@ -357,7 +357,7 @@ namespace libtorrent { namespace aux
 			// need to delete the source.
 			if (copied_files[i] == false) continue;
 
-			std::string cur_file_path{ f.file_path(i) };
+			std::string const cur_file_path{ f.file_path(i) };
```
arvidn: ``cur_file_path`` is only used once here. I think capturing this in a named variable just adds noise. It makes sense to capture above because the variable is used twice (and the work done by ``file_path(i)`` only has to be done once).

I would suggest:
```std::string const old_path = combine_path(save_path, make_sv(f.file_path(i)));
```

does that make sense?

test/test_bdecode.cpp
```diff
@@ -201,8 +201,7 @@ TORRENT_TEST(dict_null_key)
 	int ret = bdecode(b, b + sizeof(b)-1, e, ec);
 	TEST_EQUAL(ret, 0);
 	TEST_CHECK(e.dict_size() == 1);
-	std::string find_str("a\0b", 3);
-	bdecode_node d = e.dict_find(find_str);
+	bdecode_node d = e.dict_find(string_view("a\0b", 3));
```
arvidn: do you need to say ``string_view`` here? I am pretty sure it has a constructor like this:

```
template <std::size_t N>
string_view(char const (&str)[N]) { ... }
```
Allowing to construct from a string literal, and deducing the length of of it automatically (so you don't have to say 3 either)

test/test_bdecode.cpp
arvidn: ah, I see. that makese sense

include/libtorrent/file.hpp
arvidn: I still don't think these additional overloads should be here.

test/test_bdecode.cpp
ajax16384: unfortunately without "stringview(..)" it uses strlen and throws assert "expression: type() == int_t"

test/test.cpp
```diff
@@ -94,3 +94,15 @@ int print_failures()
 	return total_num_failures;
 }
 
+std::string combine_path(std::string const& ls, std::string const& rs)
```
arvidn: why do you need these?

ajax16384: many "combine_path(combine_path(" relies on this overload

ajax16384: without these overloads - patch for test block will increase with many make_sv

ajax16384:
something strange with #pragma GCC push

arvidn:
let's keep an eye at this: https://svn.boost.org/trac/boost/ticket/12917

OlafvdSpek:
Why use make_sv(something) rather then string_view(something)?

arvidn:
@OlafvdSpek because that constructor was deleted (at least according to ajax16384). Had it just been made explicit that would have worked.

ajax16384:
boost 1.64 released.
it seems they have revert disallowing of construction : https://github.com/boostorg/utility/commit/6c4ab93573904f6d37f74c8819ac39cd230118e9

arvidn:
should this be closed then?

ajax16384:
yes. just need confirmation from https://github.com/arvidn/libtorrent/issues/1803 that release boost works with master.

Issue #1797, opened 2017-03-09
VC++ 2017 error C2760: syntax error: expected ')' not ']'


pavel-pimenov:
master
![default](https://cloud.githubusercontent.com/assets/920705/23738494/62282dce-04aa-11e7-9e2e-363913ba162e.png)

arvidn:
is there a reason to believe that this is not caused by a bug in visual studio? I take it this is not a final release, right?

pavel-pimenov:
VC++ 2017 release
C:\Program Files (x86)\Microsoft Visual Studio\2017\Community>cl
Microsoft (R) C/C++ Optimizing Compiler Version 19.10.25017 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

usage: cl [ option... ] filename... [ /link linkoption... ]

Hot fix aux::* operator []

![default](https://cloud.githubusercontent.com/assets/920705/23752088/004adfa8-04e5-11e7-8f0e-b9ba8f304f0e.png)

arvidn:
that change is not correct. ``std::vector<bool>::operator[]`` does not return a ``bool&``.
is this a regression in visual studio then?

pavel-pimenov:
I use std:;vector<bool> 

(https://github.com/pavel-pimenov/libtorrent/commit/38c3a7dea601f0169eb54dab7d36385fddc530dd)

Test for operator [] VC++online
http://webcompiler.cloudapp.net/

#include <iostream>
#include <vector>
using namespace std;
int main()
{
    vector<bool> v;
    v.push_back(true);
    cout << "v[0] = " << v[0] << endl; 
    v[0] = false;
    cout << "v[0] = " << v[0] << endl; 
}
Compiled with  /EHsc /nologo /W4
main.cpp
Compilation successful!
Total compilation time: 31ms
v[0] = 1
v[0] = 0
Total execution time: 750ms

arvidn:
I don't understand what you're saying

arvidn:
try this program, it does not print "bool":

```
#include <iostream>
#include <vector>
#include <typeinfo>

using namespace std;

int main()
{
   cout << typeid(decltype(std::vector<bool>()[0])).name() << endl; 
}
```

pavel-pimenov:
so i temporary replaced `aux::vector<bool, xxx>` to `std::vector<bool>`

ajax16384:
cout << typeid(decltype(std::vector<bool>()[0])).name() << endl; 
[...]

PR #1815, opened 2017-03-15
use decltype(auto) return types for MSVC

ajax16384:
should fix https://github.com/arvidn/libtorrent/issues/1797 issue

### DIFF
--- a/include/libtorrent/config.hpp
+++ b/include/libtorrent/config.hpp
@@ -100,7 +100,10 @@ POSSIBILITY OF SUCH DAMAGE.
 
 // auto and decltype(auto) return types supports since MSVS2015
 // https://msdn.microsoft.com/en-us/library/hh567368.aspx
+// we need to force C++14 feature due VS2017 inability to parse C++11 syntax
+#if defined(_MSC_VER) && (_MSC_VER > 1900)
 #define TORRENT_AUTO_RETURN_TYPES 1
+#endif
 #endif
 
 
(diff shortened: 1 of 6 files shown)


ajax16384:
@aldenml can TORRENT_AUTO_RETURN__TYPES be defined with other c++14 capable compilers or just use it as MSVC workaround?

aldenml:
To be honest, I haven't had the opportunity to review the issue at hand. I would like to know why the new `msvc` does not support what the previous version did. It seems weird to me. Nothing else to say at the moment :(

ajax16384:
ok, let's leave it as MSVC workaround

arvidn:
aren't microsoft switching to use clang as their front-end? (it would still be weird, as clang definitely supports this syntax)

include/libtorrent/config.hpp
```diff
@@ -98,6 +98,9 @@ POSSIBILITY OF SUCH DAMAGE.
 # define TORRENT_DEPRECATED __declspec(deprecated)
 #endif
 
+// auto and decltype(auto) return types supports since MSVS2015
+// https://msdn.microsoft.com/en-us/library/hh567368.aspx
+#define TORRENT_AUTO_RETURN_TYPES 1
```
arvidn: please mention why we need this too. i.e. msvc-15 doesn't support the C++11 syntax.
Also, ideally this would only get enabled for that version of msvc

arvidn:
looks good apart from comment

ajax16384:
ok

arvidn:
lgtm

PR #1816, opened 2017-03-15
fix memory corruption during DHT simulation tests

aldenml:
I detected this working in another PR that triggers the call to `node::write_nodes_entries` during the bootstrap.

### DIFF
--- a/simulation/setup_dht.cpp
+++ b/simulation/setup_dht.cpp
@@ -89,17 +89,18 @@ struct dht_node final : lt::dht::udp_socket_interface
 		, m_socket(m_io_service)
 		, m_dht((flags & dht_network::bind_ipv6) ? udp::v6() : udp::v4()
 			, this, sett, id_from_addr(m_io_service.get_ips().front())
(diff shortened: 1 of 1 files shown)


simulation/setup_dht.cpp
```diff
@@ -89,17 +89,18 @@ struct dht_node final : lt::dht::udp_socket_interface
 		, m_socket(m_io_service)
 		, m_dht((flags & dht_network::bind_ipv6) ? udp::v6() : udp::v4()
 			, this, sett, id_from_addr(m_io_service.get_ips().front())
-			, nullptr, cnt, std::map<std::string, lt::dht::node*>(), *m_dht_storage)
+			, nullptr, cnt, m_nodes, *m_dht_storage)
```
arvidn: this object was captured by reference and stored I take it?

arvidn:
lgtm

arvidn:
I really wish there was a way to run brew and ask it to wait for the file lock, rather than just fail

simulation/setup_dht.cpp
aldenml: Yes

Issue #1810, opened 2017-03-13
Question about DHT, want, nodes/nodes6


aldenml:
libtorrent version (or branch): master

Hi @ssiloti, I have a few question about this, in the code `rcp_manage.cpp:L468`
```cpp
node& n = o->algorithm()->get_node();
if (!n.native_address(o->target_addr()))
{
	a["want"].list().push_back(entry(n.protocol_family_name()));
}
```
what I see is that you put only `want` if the target is IPv4 vs IPv6, but how is that possible? don't the DHT communicate only with nodes of the same protocol?

Later in `traversal_algorithm.cpp:L583`
```cpp
bdecode_node const n = r.dict_find_string(nodes_key);
```
what I see is that you get only nodes with the same protocol, with only `libtorrent` nodes, how it's supposed to work?

In `node::write_nodes_entries` you always write nodes, don't you risk always getting a message too big for UDP?

Thanks for your time

ssiloti:
Most requests are only sent to nodes in the same address family, but bootstrap queries can be sent to any address.

I'm not sure what your confusion is with the second question.

The number of nodes entries written should be small enough that the packet never exceeds the typical MTU of UDP over Ethernet (i.e. slightly less than 1500 bytes). This is something which is enforced purely through convention so it's not as robust as it could be.

aldenml:
Right, I missed the bootstrap queries, but looking at the code again, I don't see where a "bootstrap" could mix protocols for the same node. Could you help me here? Is `dht_state` bad designed?

I think I clarified myself my second question/confusion, I just want to confirm that the fact that `node::write_nodes_entries` writes nodes even if `want` **IS NOT THERE** is a key factor. Is that right?

ssiloti:
Yes, if the `want` key is not present then it is implied that the requester wants nodes of the same address family as the request was sent to.

Bootsrap was intended to take nodes of any address family via the `nodes` parameter of `node::bootstrap` but I see that got blocked by 6c31ea3d44.

aldenml:
Do you consider a priority to revert part of that to allow mixing protocols again in the bootstrap?

ssiloti:
It should definitely be fixed, but it's not a big enough deal for me to drop what I'm working on to fix it if that's what you're asking.

aldenml:
no, of course not :) it's for myself, to continue in the other PRs or put some time on this.

PR #1818, opened 2017-03-16
allow DHT bootstrap to use nodes of both IP protocols

aldenml:
To address https://github.com/arvidn/libtorrent/issues/1810

@ssiloti appreciate if you have a chance to take a look at this

### DIFF
--- a/src/kademlia/dht_tracker.cpp
+++ b/src/kademlia/dht_tracker.cpp
@@ -69,6 +69,14 @@ namespace libtorrent { namespace dht
 		c.inc_stats_counter(counters::dht_allocated_observers, allocated_observers);
 	}
 
+	std::vector<udp::endpoint> concat(std::vector<udp::endpoint> const& v1
+		, std::vector<udp::endpoint> const& v2)
+	{
+		std::vector<udp::endpoint> r = v1;
+		r.insert(r.end(), v2.begin(), v2.end());
+		return r;
+	}
+
 	} // anonymous namespace
 
 	// class that puts the networking and the kademlia node in a single
@@ -155,9 +163,10 @@ namespace libtorrent { namespace dht
 		m_refresh_timer.expires_from_now(seconds(5), ec);
 		m_refresh_timer.async_wait(std::bind(&dht_tracker::refresh_timeout, self(), _1));
 
-		m_dht.bootstrap(m_state.nodes, f);
+		// bootstrap with mix of IP protocols via want/nodes/nodes6
+		m_dht.bootstrap(concat(m_state.nodes, m_state.nodes6), f);
 #if TORRENT_USE_IPV6
-		m_dht6.bootstrap(m_state.nodes6, f);
+		m_dht6.bootstrap(concat(m_state.nodes6, m_state.nodes), f);
 #endif
 		m_state.clear();
 	}
--- a/src/kademlia/node.cpp
+++ b/src/kademlia/node.cpp
@@ -220,7 +220,10 @@ void node::bootstrap(std::vector<udp::endpoint> const& nodes
 
 	for (auto const& n : nodes)
 	{
-		if (n.protocol() != protocol()) continue;
+#if !TORRENT_USE_IPV6
+		if (n.protocol() == udp::v6()) continue;
+#endif
+
 #ifndef TORRENT_DISABLE_LOGGING
 		++count;
 #endif
--- a/test/test_dht.cpp
+++ b/test/test_dht.cpp
@@ -1824,6 +1824,75 @@ TORRENT_TEST(bootstrap_v6)
 }
 #endif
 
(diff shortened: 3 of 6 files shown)


ssiloti:
lgtm

src/kademlia/dht_tracker.cpp
```diff
@@ -155,9 +163,10 @@ namespace libtorrent { namespace dht
 		m_refresh_timer.expires_from_now(seconds(5), ec);
 		m_refresh_timer.async_wait(std::bind(&dht_tracker::refresh_timeout, self(), _1));
 
-		m_dht.bootstrap(m_state.nodes, f);
+		// bootstrap with mix of IP protocols via want/nodes/nodes6
+		m_dht.bootstrap(concat(m_state.nodes, m_state.nodes6), f);
```
arvidn: does the order matter? If not, you could save a bit by capturing the concatenated vector and pass it in by const reference to both calls. i.e.

```auto nodes = m_state.nodes;
nodes.insert(nodes.end(), m_state.nodes6.begin(), m_state.node6.end());
m_dht.bootstrap(nodes);
...
```

aldenml: well, since the "sends" are FIFO (and that's my assumption), I wanted to prioritize the ones with the same protocol. You tell me if that's a relevant factor here.

src/kademlia/node.cpp
```diff
@@ -220,7 +220,10 @@ void node::bootstrap(std::vector<udp::endpoint> const& nodes
 
 	for (auto const& n : nodes)
 	{
-		if (n.protocol() != protocol()) continue;
+#if !TORRENT_USE_IPV6
+		if (n.protocol() == udp::v6()) continue;
```
arvidn: if you don't concatenate ``nodes6``, presumably this wouldn't happen, right?
I mean, maybe you could have an ifdef to not add in the v6 nodes in the first place, if ``TORRENT_USE_IPV6`` wasn't set.

test/test_dht.cpp
```diff
@@ -1824,6 +1824,75 @@ TORRENT_TEST(bootstrap_v6)
...
+		nodesv.push_back(rand_udp_ep(rand_v6));
+	else
+		nodesv.push_back(rand_udp_ep(rand_v4));
+
+	t.dht_node.bootstrap(nodesv, std::bind(&nop_node));
+
+	TEST_EQUAL(g_sent_packets.size(), 1);
+	TEST_EQUAL(g_sent_packets.front().first, nodesv[0]);
+
+	lazy_from_entry(g_sent_packets.front().second, response);
```
arvidn: I was under the impression that ``lazy_entry`` only existed for backwards compatibility now. Is it still used here?

src/kademlia/node.cpp
aldenml: The problem is that the state (`dht_state`) have a vector of `udp::endpoint` (two to be precise), there is no really a way to make sure they are all IPv4. This is a safety measure.

src/kademlia/dht_tracker.cpp
arvidn: I don't think sending is FIFO. as far as I can tell, the traversal algorithm inserts nodes sorted. There's a ``std::lower_bound()`` in ``traversal_algorithm::add_entry()``.

In fact, looking at the bootstrap algorithm, it doesn't look quite right. It will just generate random node IDs for the IPs we bootstrap from, so it will be shuffled. But then we call ``trim_seed_nodes()``, which assumes that the nodes are sorted and have correct node IDs. The rationale being that we want to start farther away when bootstrapping.

Starting farther away makes sense, but I don't think this call makes sense, since we know that we just created random node IDs, and the first 32 nodes are just some random nodes, not the ones closest to us.

test/test_dht.cpp
aldenml: The name `lazy_from_entry` is an old one (it's just the name), it's working with `bdecode_node` now.

src/kademlia/dht_tracker.cpp
aldenml: hmmm...out of my understanding at the moment. Is it OK if I reuse the vector then to get this merged, or is the review of the algorithm a requisite?

test/test_dht.cpp
arvidn: ah :)

src/kademlia/node.cpp
arvidn: I see, that makes sense

src/kademlia/dht_tracker.cpp
arvidn: no, they're separate. and actually, perhaps it would make sense to not insert nodes with the ``flag_no_id`` flag set in a sorted location, but to just append to the end.

PR #1819, opened 2017-03-16
fix IndexType template parameter in typed_span member functions

aldenml:


include/libtorrent/aux_/typed_span.hpp
```diff
@@ -64,56 +64,56 @@ namespace libtorrent { namespace aux {
 
 		template <typename U = underlying_index, typename Cond
 			= typename std::enable_if<std::is_signed<U>::value>::type>
-		typed_span<T> first(underlying_index n) const
+		typed_span<T, IndexType> first(underlying_index n) const
```
arvidn: actually, you should be able to just say ``typed_span``. iirc, the default is to imply the same template arguments as the class' own arguments. would you mind trying that?

arvidn:
lgtm

include/libtorrent/aux_/typed_span.hpp
aldenml: sure, it works too. Actually, I prefer the implicit way, but I wasn't sure about yours. I will change it.

PR #1820, opened 2017-03-16
refactor of read/write nodes in dht_live_nodes_alert to reusable functions

aldenml:
I will need these functions in https://github.com/arvidn/libtorrent/pull/1652

### DIFF
--- a/src/alert.cpp
+++ b/src/alert.cpp
@@ -2136,28 +2136,32 @@ namespace libtorrent
 		return buf;
 	}
 
-	dht_live_nodes_alert::dht_live_nodes_alert(aux::stack_allocator& alloc
-		, sha1_hash const& nid
-		, std::vector<std::pair<sha1_hash, udp::endpoint>> const& nodes)
-		: node_id(nid)
-		, m_alloc(alloc)
+	namespace
 	{
+	using nodes_slot = std::tuple<int, aux::allocation_slot, int, aux::allocation_slot>;
+
+	nodes_slot write_nodes(aux::stack_allocator& alloc
+		, std::vector<std::pair<sha1_hash, udp::endpoint>> const& nodes)
+	{
+		int v4_num_nodes = 0;
+		int v6_num_nodes = 0;
+
 		for (auto const& n : nodes)
 		{
 			if (n.second.protocol() == udp::v4())
-				m_v4_num_nodes++;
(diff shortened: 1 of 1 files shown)


src/alert.cpp
```diff
@@ -2136,28 +2136,32 @@ namespace libtorrent
...
 			else
-				m_v6_num_nodes++;
+				v6_num_nodes++;
 #endif
 		}
 
-		m_v4_nodes_idx = alloc.allocate(m_v4_num_nodes * (20 + 6));
-		m_v6_nodes_idx = alloc.allocate(m_v6_num_nodes * (20 + 18));
+		aux::allocation_slot const v4_nodes_idx = alloc.allocate(v4_num_nodes * (20 + 6));
+		aux::allocation_slot const v6_nodes_idx = alloc.allocate(v6_num_nodes * (20 + 18));
```
arvidn: does it make sense to allocate space for IPv6 nodes if ``TORRENT_USE_IPV6`` is not set?
I wonder if this was a deliberate decision or an oversight

src/alert.cpp
```diff
@@ -2136,28 +2136,32 @@ namespace libtorrent
...
 		, std::vector<std::pair<sha1_hash, udp::endpoint>> const& nodes)
-		: node_id(nid)
-		, m_alloc(alloc)
 	{
+		int v4_num_nodes = 0;
+		int v6_num_nodes = 0;
+
 		for (auto const& n : nodes)
 		{
 			if (n.second.protocol() == udp::v4())
```
arvidn: if you invert this check you can avoid it entirely when IPv6 is disabled

src/alert.cpp
aldenml: it was deliberate, and it's `0`, no actual allocation

src/alert.cpp
aldenml: jeje, this is intentional, a safety feature, since it's `udp::endpoint`

arvidn:
iiuc, you'r just factoring out the work previously done in the constructor to a separate function, as well as the work in assembling the node vector. is that right?

aldenml:
That's correct, because I have plans to use these functions in my other PR (https://github.com/arvidn/libtorrent/pull/1652) to return `nodes`.

Issue #1821, opened 2017-03-16
To upgrade to travis xcode7.3 image


aldenml:
libtorrent version (or branch): `master`

Finally I discovered the reason `brew` is failing with such a frequency in travis. The relevant documentation is here: https://github.com/travis-ci/travis-ci/issues/7456

Summary:
The `cache` feature uses `md5deep`. That program is not included in all images of old Mac OS (who knows why aren't they all equals). They install it with `brew` and that triggers a `brew update`. But the entire process of downloading the "cache" fails for multiple reasons (size one of them), leaving garbage and locks.

I checked the behavior in my account.

Easy solution is use the next (and default) version xcode7.3, but that requires changing the include of `openssl`, since it's already deprecated in that version. Future versions of xcode simply lack any `openssl` headers.

@arvidn what do you think about that?

arvidn:
thanks for looking into this! I will merge your PR

PR #1824, opened 2017-03-17
using travis xcode7.3 image to avoid brew update issue

aldenml:
To address https://github.com/arvidn/libtorrent/issues/1821 in case you think it's a good idea.

I removed one redundant "brew update" and the invalid option "--quiet"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,13 @@ matrix:
     - env: variant=test_release coverage=1 tests=1 toolset=gcc-coverage python=1
     - env: autotools=1 toolset=gcc
     - os: osx
-      osx_image: xcode6.4
+      osx_image: xcode7.3
       env: variant=test_release docs=1 tests=1 toolset=darwin
     - os: osx
-      osx_image: xcode6.4
+      osx_image: xcode7.3
       env: variant=test_debug ssl=openssl tests=1 crypto=libcrypto toolset=darwin
     - os: osx
-      osx_image: xcode6.4
+      osx_image: xcode7.3
       env: variant=test_barebones tools=1 examples=1 toolset=darwin python=1
     - dist: trusty
       env: arch=arm toolset=gcc-arm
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -55,17 +55,17 @@ before_install:
...
-    travis_retry brew install --quiet docutils;
+    travis_retry brew install docutils;
     mkdir -p /Users/travis/Library/Python/2.7/lib/python/site-packages;
     echo ''import site; site.addsitedir("/usr/local/lib/python2.7/site-packages")'' >> /Users/travis/Library/Python/2.7/lib/python/site-packages/homebrew.pth;
     sudo easy_install Pygments;
     sudo easy_install -U aafigure;
-    travis_retry brew install --quiet graphviz;
-    travis_retry brew install --quiet Homebrew/python/pillow;
+    travis_retry brew install graphviz;
+    travis_retry brew install Homebrew/python/pillow;
```
arvidn: I wonder if merging these two calls into a singe ``brew install`` would make it more likely to pass too

.travis.yml
```diff
@@ -8,13 +8,13 @@ matrix:
...
-      osx_image: xcode6.4
+      osx_image: xcode7.3
       env: variant=test_release docs=1 tests=1 toolset=darwin
     - os: osx
-      osx_image: xcode6.4
+      osx_image: xcode7.3
       env: variant=test_debug ssl=openssl tests=1 crypto=libcrypto toolset=darwin
     - os: osx
-      osx_image: xcode6.4
+      osx_image: xcode7.3
```
arvidn: does the new version of xcode improve anything related to brew?

aldenml: yes, it's in the summary of https://github.com/arvidn/libtorrent/issues/1821

Jamfile
```diff
@@ -27,7 +27,6 @@ if $(BOOST_ROOT)
 else
 {
 	local boost-lib-search-path =
-		<search>/usr/lib
```
arvidn: is your assertion here that this is always a default search path?

.travis.yml
aldenml: hardly a significant factor

Jamfile
aldenml: yes, and in fact, it gets in the way, since it's explicitly passed, and it takes higher priority, event if a custom toolchain does not depend on the system libraries

aldenml: To rephrase the above: If you use a custom toolchain or specify --sysroot, then explicitly passing `/usr/lib` is a problem, if you use a system toolchain, then it takes higher priority.

PR #1825, opened 2017-03-17
fixing remaining clang sign warnings

aldenml:


include/libtorrent/netlink.hpp
```diff
@@ -63,7 +63,7 @@ namespace libtorrent
 			std::memset(&m_sockaddr, 0, sizeof(sockaddr_nl));
 			m_sockaddr.nl_family = AF_NETLINK;
 			m_sockaddr.nl_groups = group;
-			m_sockaddr.nl_pid = pid;
+			m_sockaddr.nl_pid = std::uint32_t(pid);
```
arvidn: would it be safer to cast to ``pid_t``? I would be worried about a system that uses 64 bit numbers for pids.

arvidn:
lgtm apart from comment

include/libtorrent/netlink.hpp
aldenml: @arvidn I need some help to decide here. In the android headers, `sockaddr_nl::nl_pid` is hardcoded to `__u32`, in the documentation http://man7.org/linux/man-pages/man7/netlink.7.html it is `pid_t`, but the actual meaning of the field is "Port ID" (not "Process ID").

Then you have `nlmsghdr:: nlmsg_pid` coded to `__u32` in the documentation. Not clear to me what to do

arvidn: I see, ``pid`` is already of type ``pid_t``. So really what this assignment should be is: "cast to whatever type ``nl_pid`` has", I assume.

arvidn: I believe the use of ``pid_t`` is a typo in the man page. If you look at the actual header, they're both u32. http://lxr.free-electrons.com/source/include/uapi/linux/netlink.h#L35

Which means that the use of ``pid_t`` here is probably incorrect as well.

arvidn: also, as far as I can tell, there's no place where the "port ID" is actually specified. how about you just make the parameter a ``uint32_t`` instead? That seems to be the most correct solution.

aldenml: done

PR #1828, opened 2017-03-18
is_ip_address refactor and minor code cleanup

aldenml:


src/broadcast_socket.cpp
```diff
@@ -56,10 +56,10 @@ using namespace std::placeholders;
 
 namespace libtorrent
 {
-	bool is_ip_address(char const* host)
+	bool is_ip_address(string_view host)
 	{
 		error_code ec;
-		address::from_string(host, ec);
+		address::from_string(host.data(), ec);
```
arvidn: what's returned from ``data()`` is not necessarily 0-terminated, so I don't think this is right

arvidn: there is an overload of ``from_string`` that takes a ``std::string const&`` though, so maybe the best solution for now is to make ``is_ip_address`` take that as its parameter

aldenml: ok

src/session_impl.cpp
```diff
@@ -633,9 +633,7 @@ namespace aux {
 		update_dht();
 		update_peer_fingerprint();
 		update_dht_bootstrap_nodes();
-#ifndef TORRENT_DISABLE_DHT
 		update_dht_announce_interval();
```
arvidn: is this change significant? I wouldn't expect this function to do anything when DHT is disabled

aldenml: not, but just for aesthetics.

src/torrent.cpp
```diff
@@ -5790,8 +5790,8 @@ namespace libtorrent
 			return;
 		}
 
-		bool const is_ip = is_ip_address(hostname.c_str());
-		if (is_ip) a.address(address::from_string(hostname.c_str(), ec));
+		bool const is_ip = is_ip_address(hostname);
+		if (is_ip) a.address(address::from_string(hostname, ec));
```
arvidn: heh, this code is a bit weird (unrelated to your patch). ``is_ip_address()`` just calls ``from_string`` and checks the error, it seems redundant to just call it twice here

aldenml: true

arvidn:
lgtm. just one question about the disable DHT define

PR #1830, opened 2017-03-18
fix setup of DHT logic in session_impl::init

aldenml:
Due to the way `settings_pack` works when applied, special care should be taken when creating the session and the DHT. Three cases should be considered:
1-) `dht_bootstrap_nodes` setting not touched
2-) `dht_bootstrap_nodes` changed but not empty (`""`)
3-) `dht_bootstrap_nodes` set to empty (`""`)

### In `master`
Currently, with head `master`, if you see the log stream, this is what you get:
1-) `dht_bootstrap_nodes` setting not touched
**This is the worst case, the DHT is twice created and bootstrapped.**
```
log_alert - start session
...
log_alert - about to start DHT, running: false, router lookups: 0, aborting: false
log_alert - about to stop DHT, running: false
dht_log_alert - DHT tracker: starting IPv4 DHT tracker with node id: XXXX
dht_log_alert - DHT tracker: starting IPv6 DHT tracker with node id: XXXX
...
log_alert - about to start DHT, running: true, router lookups: 0, aborting: false
log_alert - about to stop DHT, running: true
dht_log_alert - DHT tracker: starting IPv4 DHT tracker with node id: XXX
dht_log_alert - DHT tracker: starting IPv6 DHT tracker with node id: XXX
...
```

2-) `dht_bootstrap_nodes` changed but not empty (`""`)
_unnecessary call to `update_dht_announce_interval()`_
```
log_alert - start session
...
log_alert - about to start DHT, running: false, router lookups: 4, aborting: false
log_alert - about to stop DHT, running: false
log_alert - not starting DHT announce timer: m_dht == nullptr
...
log_alert - about to start DHT, running: false, router lookups: 0, aborting: false
log_alert - about to stop DHT, running: false
dht_log_alert - DHT tracker: starting IPv4 DHT tracker with node id: XXXX
dht_log_alert - DHT tracker: starting IPv6 DHT tracker with node id: XXXX
...
```
3-) `dht_bootstrap_nodes` set to empty (`""`)
_it is fine_
```
log_alert - start session
...
log_alert - about to start DHT, running: false, router lookups: 0, aborting: false
log_alert - about to stop DHT, running: false
dht_log_alert - DHT tracker: starting IPv4 DHT tracker with node id: XXXX
dht_log_alert - DHT tracker: starting IPv6 DHT tracker with node id: XXXX
...
```

### With this PR
1-) `dht_bootstrap_nodes` setting not touched
```
log_alert - start session
...
log_alert - about to start DHT, running: false, router lookups: 0, aborting: false
log_alert - about to stop DHT, running: false
dht_log_alert - DHT tracker: starting IPv4 DHT tracker with node id: XXXX
dht_log_alert - DHT tracker: starting IPv6 DHT tracker with node id: XXXX
...
```

2-) `dht_bootstrap_nodes` changed but not empty (`""`)
```
log_alert - start session
...
log_alert - about to start DHT, running: false, router lookups: 0, aborting: false
log_alert - about to stop DHT, running: false
dht_log_alert - DHT tracker: starting IPv4 DHT tracker with node id: XXXX
dht_log_alert - DHT tracker: starting IPv6 DHT tracker with node id: XXXX
...
```
3-) `dht_bootstrap_nodes` set to empty (`""`)
```
log_alert - start session
...
log_alert - about to start DHT, running: false, router lookups: 0, aborting: false
log_alert - about to stop DHT, running: false
dht_log_alert - DHT tracker: starting IPv4 DHT tracker with node id: XXXX
dht_log_alert - DHT tracker: starting IPv6 DHT tracker with node id: XXXX
...
```

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -632,23 +632,42 @@ namespace aux {
 		update_lsd();
 		update_peer_fingerprint();
 
+		init_dht();
+	}
+
+	void session_impl::init_dht()
+	{
+		// the need of this elaborated logic is because if the value
+		// of settings_pack::dht_bootstrap_nodes is not the default,
+		// then update_dht_bootstrap_nodes is called. For this reason,
+		// three different cases should be considered.
+		// 1-) dht_bootstrap_nodes setting not touched
+		// 2-) dht_bootstrap_nodes changed but not empty
+		// 3-) dht_bootstrap_nodes set to empty ("")
+		// TODO: find a solution and refactor to avoid potentially stalling
+		// for minutes due to the name resolution
+
 #ifndef TORRENT_DISABLE_DHT
-		// setup DHT
 		if (m_outstanding_router_lookups == 0)
 		{
 			// this can happens because either the setting value was untouched
 			// or the value in the initial settings is empty
 			if (m_settings.get_str(settings_pack::dht_bootstrap_nodes).empty())
 			{
+				// case 3)
 				update_dht();
 				update_dht_announce_interval();
 			}
 			else
 			{
+				// case 1)
 				// eventually update_dht() is called when all resolves are done
 				update_dht_bootstrap_nodes();
 			}
 		}
+		// else is case 2)
+		// in this case the call to update_dht_bootstrap_nodes() by the apply settings
+		// will eventually call update_dht() when all resolves are done
 #endif
 	}
 
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -630,10 +630,26 @@ namespace aux {
 		update_upnp();
 		update_natpmp();
 		update_lsd();
-		update_dht();
 		update_peer_fingerprint();
-		update_dht_bootstrap_nodes();
-		update_dht_announce_interval();
+
+#ifndef TORRENT_DISABLE_DHT
+		// setup DHT
+		if (m_outstanding_router_lookups == 0)
+		{
+			// this can happens because either the setting value was untouched
+			// or the value in the initial settings is empty
+			if (m_settings.get_str(settings_pack::dht_bootstrap_nodes).empty())
+			{
+				update_dht();
+				update_dht_announce_interval();
+			}
+			else
+			{
+				// eventually update_dht() is called when all resolves are done
+				update_dht_bootstrap_nodes();
+			}
+		}
+#endif
 	}
 
 	void session_impl::async_resolve(std::string const& host, int flags
--- a/test/test_session.cpp
+++ b/test/test_session.cpp
@@ -394,3 +394,73 @@ TORRENT_TEST(save_state_peer_id)
 	TEST_CHECK(pid3[5] == 'r');
 }
 
+TORRENT_TEST(init_dht)
+{
+	auto count_dht_inits = [](session& ses)
+	{
+		int count = 0;
+		int num = 70; // this number is adjusted per version, an estimate
+		time_point const end_time = clock_type::now() + seconds(15);
+		while (true)
+		{
+			time_point const now = clock_type::now();
+			if (now > end_time) return count;
+
+			ses.wait_for_alert(end_time - now);
+			std::vector<alert*> alerts;
+			ses.pop_alerts(&alerts);
+			for (auto a : alerts)
+			{
+				std::printf("%d: [%s] %s\n", num, a->what(), a->message().c_str());
+				if (a->type() == log_alert::alert_type)
+				{
+					std::string const msg = a->message();
+					if (msg.find("about to start DHT") != std::string::npos)
+						count++;
+				}
+				num--;
+			}
+			if (num <= 0) return count;
+		}
+		return count;
+	};
+
+	{
+		settings_pack p = settings();
+		p.set_bool(settings_pack::enable_dht, true);
+		p.set_int(settings_pack::alert_mask, alert::all_categories);
+		// default value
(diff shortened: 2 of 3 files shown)


src/session_impl.cpp
```diff
@@ -630,10 +630,26 @@ namespace aux {
...
 		update_natpmp();
 		update_lsd();
-		update_dht();
 		update_peer_fingerprint();
-		update_dht_bootstrap_nodes();
-		update_dht_announce_interval();
+
+#ifndef TORRENT_DISABLE_DHT
+		// setup DHT
+		if (m_outstanding_router_lookups == 0)
```
arvidn: I don't understand how ``session_impl::init()`` can be called more than once. Is ``m_outstanding_router_lookups`` ever non-zero?

arvidn: I suppose ``init()`` doesn't necessarily have to be called twice for ``m_outstanding_router_lookups`` to be non-zero, but if it is, do you know how it may get set to non-zero before getting here?

arvidn: does ``update_dht`` get called via ``apply_settings_pack`` perhaps?

arvidn:
I don't fully understand how the original problem happens, but I would have expected to be fine to call ``update_dht`` without the DHT actually restarting, assuming the DHT settings didn't change, is that not the case?

aldenml:
No, that's not the case, `update_dht` only test for `settings_pack::enable_dht` and if enable, goes directly to `start_dht` and the first line there is `stop_dht();`.

Now, the fundamental problem here is that `update_dht_bootstrap_nodes` can be called (or not) during the `apply_settings_pack_impl`, and if not, then calling `update_dht_bootstrap_nodes` after `update_dht` (as it's in master) triggers a double creation of the DHT.

src/session_impl.cpp
aldenml: If the value of `settings_pack::dht_bootstrap_nodes` is not the default, then `update_dht_bootstrap_nodes` is called and makes `m_outstanding_router_lookups > 0` and when the resolve is done, `update_dht` is called.

arvidn:
I think I understand what the problem is. This definitely seem like a step in the right direction. I wonder though if a longer term solution would be to do something similar to disk-related settings, and listen socket settings. Where all settings are applied first, and then if one affected the listen socket(s), the update function is called.

This has the additional dimension of having bootstrap nodes lookup potentially taking a long time. I think this makes for a difficult trade-off. If we don't start the DHT because we have bootstrap node lookups outstanding, how long to we wait if they get stalled? It's possible we could bootstrap the node one the node cache from last time still, in which case we're just stalling, potentially for minutes.

arvidn:
sorry for the delay in looking at this, but would you mind adding some comments around those conditionals to describe what's going on? I don't think it's obvious when reading the code

arvidn:
can you think of a good way to test that the DHT is not started twice? would it make sense to capture the ``session_log_alerts``? it seems a bit fragile perhaps.

aldenml:
I thought of it, but I wanted to wait until the problem was clarified (I didn't expect an immediate merge :)

I will read your comments and try to assemble something.

aldenml:
@arvidn let me know what you think now. I can confirm that with the old logic in `session_impl::init`, the test fail.

test/test_session.cpp
```diff
@@ -394,3 +394,73 @@ TORRENT_TEST(save_state_peer_id)
...
+			ses.wait_for_alert(end_time - now);
+			std::vector<alert*> alerts;
+			ses.pop_alerts(&alerts);
+			for (auto a : alerts)
+			{
+				std::printf("%d: [%s] %s\n", num, a->what(), a->message().c_str());
+				if (a->type() == log_alert::alert_type)
+				{
+					std::string const msg = a->message();
+					if (msg.find("about to start DHT") != std::string::npos)
```
arvidn: @aldenml I don't think this test is quite right. it looks for the "about to start DTH" log, but looking at the code where that's emitted, that message is printed even when starting the DHT is suppressed/delayed because there are outstanding lookups. see [here](https://github.com/arvidn/libtorrent/blob/master/src/session_impl.cpp#L5716).

aldenml: That’s is true if update_dht is called like it was before, or like it is in RC_1_1, but the new init_dht here take care of the special cases.

arvidn: I bring this up as I've been merging RC_1_1 into master, with the changes to how ``m_settings`` are initialized. The test for this patch failed, but if I move the log to when the DHT is *actually* started, they all pass. As far as I can tell, the intention of the test was to make sure the DHT is only constructed and started once on startup, regardless of those settings.

As far as I can tell, the merged master (which I still only have in my working copy) doesn't need the ``init_dht()`` function. what happens is that the router nodes are set first, and they kick off a host name lookup. Even if this is a cache hit, it won't return "in-line", it will post the result to the io_service and effectively defer starting the DHT. Next, ``update_dht()`` is called, but it won't actually construct or start the DHT, since there's at least one outstanding lookup.

Once the lookup comes back, the DHT will be started. Only once.

The reason this works in my merged version is because ``update_dht_bootstrap_nodes()`` is called before ``update_dht()``, which is an implementation detail of ``run_all_updates()``, but at least these tests would catch it if it was broken.

Trying to understand all the cases in this patch, I can't think of a way if a case that would break the single-dht-init.

I have not gone back to look at current master to see what would happen there if I moved the log call.

PR #1832, opened 2017-03-19
make DHT bootstrapping more robust by not throwing away nodes

arvidn:


arvidn:
@aldenml does this make sense to you? the diff looks a bit more complex than it is (because of some indentation changes). I'll mark it up with some comments in a bit

include/libtorrent/kademlia/refresh.hpp
```diff
@@ -51,8 +51,6 @@ class bootstrap : public get_peers
 	observer_ptr new_observer(udp::endpoint const& ep
 		, node_id const& id);
 
-	void trim_seed_nodes();
```
arvidn: this was the main thing I wanted to get rid of

include/libtorrent/kademlia/traversal_algorithm.hpp
```diff
@@ -105,15 +105,27 @@ struct TORRENT_EXTRA_EXPORT traversal_algorithm : boost::noncopyable
 	int num_timeouts() const { return m_timeouts; }
 
 	node& m_node;
+
+	// this vector is sorted by node-id distance from our node id. Closer nodes
+	// are earlier in the vector. However, not the entire vector is necessarily
+	// sorted, the tail of the vector may contain nodes out-of-order. This is
+	// used when bootstrapping. The ``m_sorted_results`` member indicates how
+	// many of the first elements are sorted.
```
arvidn: this describes the main change in this patch

src/kademlia/traversal_algorithm.cpp
```diff
@@ -118,85 +140,110 @@ void traversal_algorithm::add_entry(node_id const& id
 		done();
 		return;
 	}
+
+	o->flags |= flags;
+
 	if (id.is_all_zeros())
```
arvidn: the observation I made was that bootstrap nodes (i.e. the ones we load from our state file) never have node IDs associated with them. it's unnecessary to generate bogus node IDs and then sort the nodes based on them. Instead, allow the tail of ``m_results`` to be unsorted, and function as back-up nodes when lookups fails (and don't yield more nodes)

src/kademlia/traversal_algorithm.cpp
```diff
@@ -118,85 +140,110 @@ void traversal_algorithm::add_entry(node_id const& id
...
 	{
 		o->set_id(generate_random_id());
 		o->flags |= observer::flag_no_id;
-	}
 
-	o->flags |= flags;
+		m_results.push_back(o);
```
arvidn: I just push back here, causing most of this function to end up in the "else" block, and gaining one level of indentation (which makes the patch look ugly)

src/kademlia/traversal_algorithm.cpp
```diff
@@ -118,85 +140,110 @@ void traversal_algorithm::add_entry(node_id const& id
...
-				, distance_exp(m_target, id), m_invoke_count, name());
-		}
-#endif
-		iter = m_results.insert(iter, o);
-
-		TORRENT_ASSERT(libtorrent::dht::is_sorted(m_results.begin(), m_results.end()
-			, [this](observer_ptr const& lhs, observer_ptr const& rhs)
-			{ return compare_ref(lhs->id(), rhs->id(), m_target); }));
-	}
+	TORRENT_ASSERT(libtorrent::dht::is_sorted(m_results.begin(), end
```
arvidn: the new invariant is that the head of this vector is sorted

aldenml:
@arvidn I think I can vaguely grasp the idea, but at this moment I'm far from be able to emit an opinion about correctness.

PR #1838, opened 2017-03-22
fix python binding for parse_magnet_uri

arvidn:


bindings/python/test.py
```diff
@@ -256,6 +256,14 @@ def test_sha1hash(self):
...
 		self.assertEqual(h, str(s))
 
+class test_magnet_link(unittest.TestCase):
+
+	def test_parse_magnet_uri(self):
+		ses = lt.session({})
+		magnet = 'magnet:?xt=urn:btih:C6EIF4CCYDBTIJVG3APAGM7M4NDONCTI'
+		p = lt.parse_magnet_uri(magnet)
+		p['save_path'] = '.'
+		h = ses.add_torrent(p)
```
cas--: Should have an assert that will also show up the second issue; incorrect hash calculated:

```
self.assertEqual(str(h.info_hash()), '178882f042c0c33426a6d81e0333ece346e68a68')
```

cas--:
That's a bit strange that it is passing the test...

arvidn:
the magnet link uses base32 encoding of the info-hash. This is unusual (the original specification used this, but as far as I recall it was changed to hex 10 years ago). However, base32 is still supported (and apparently used sometimes).

The info_hash strings used in the python bindings are always hex

PR #1842, opened 2017-03-23
deprecated (undocumented) file:// urls, added torrent_file_path altertive

arvidn:


arvidn:
I'm a bit mixed about the utility of this feature. on the one hand, it makes it really convenient to add lots of torrents very quickly and efficiently. but in practice, I imagine the client would need to load the resume file anyway, so you're already stuck performing an expensive load from disk, so unless that could be done on the libtorrent side as well, is it really that valuable?

arvidn:
I suppose it's still a step in the right direction. I may remove ``torrent_file_path`` before cutting a release from master

PR #1843, opened 2017-03-23
optimize access to ssl_cert

arvidn:
specifically the case where there is none, avoid pessimisation of the common case

### DIFF
--- a/include/libtorrent/torrent_info.hpp
+++ b/include/libtorrent/torrent_info.hpp
@@ -400,7 +400,7 @@ namespace libtorrent
 		// Returns the SSL root certificate for the torrent, if it is an SSL
 		// torrent. Otherwise returns an empty string. The certificate is
 		// the the public certificate in x509 format.
-		std::string ssl_cert() const;
+		string_view ssl_cert() const;
 
 		// returns true if this torrent_info object has a torrent loaded.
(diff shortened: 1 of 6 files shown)


include/libtorrent/torrent_info.hpp
```diff
@@ -643,7 +643,11 @@ namespace libtorrent
 			// domain in its hostname. This means the DHT and LSD
 			// features are disabled for this torrent (unless the
 			// settings allows mixing i2p peers with regular peers)
-			i2p = 4
+			i2p = 4,
+
+			// this flag is set if we found an ssl-cert field in the info
+			// dictionary
+			ssl_torrent = 8,
```
aldenml: I suspect this `,` at the end could be a problem, in that case try https://blog.travis-ci.com for auto cancelation of PRs (new)

arvidn: you mean cppcheck.py may complain about it?

arvidn: c++ allows this (for good reasons I think)

aldenml: interesting, I remember in the past getting a clang warning about this in another enum, anyway, if you have a chance, take a look at the link to enable "auto cancelation" in PRs.

PR #1860, opened 2017-03-28
minor general code cleanup and refactor

aldenml:


src/resolver.cpp
```diff
@@ -73,9 +71,8 @@ namespace libtorrent
 		// oldest entries
 		if (int(m_cache.size()) > m_max_size)
 		{
-			cache_t::iterator oldest = m_cache.begin();
-			for (cache_t::iterator k = m_cache.begin();
-				k != m_cache.end(); ++k)
+			auto oldest = m_cache.begin();
+			for (auto k = m_cache.begin(); k != m_cache.end(); ++k)
```
arvidn: this should really be a ``std::min_element()`` call

PR #1861, opened 2017-03-28
removed (broken) support for incoming connections over socks5

arvidn:


arvidn:
@cas-- This patch seems to have uncovered an unrelated issue in the python binding. Specifically the test you added to make sure ``announce_entry`` could be serialized as json. It seems it was passing by accident. If ``min_announce`` or ``next_announce`` is set to a valid date object, it fails. (see https://travis-ci.org/arvidn/libtorrent/jobs/215811845#L2494 ).

Do you have any ideas of what to do here?
I can't recall if JSON was relevant here, or whether it really tested some other property.

cas--:
So json is relevant to ensure backward compatibility that the tracker did not contain any python objects. I think I did ask at the time about testing with valid date object but forgot to follow up on that.

Coincidentally this tracker issue does actually break Deluge 1.3 and I'm in the middle of applying a workaround for lt `1.1.2`: http://dev.deluge-torrent.org/ticket/3008

What I was going to propose is that an epoch time in secs is returned rather than datetime for `min_announce` and `next_announce` because I will need to do that conversion in Deluge code anyway.

arvidn:
did these fields used to be posix time counters? in 1.0?

arvidn:
https://github.com/arvidn/libtorrent/pull/1865

cas--:
Those fields were just added in 1.1.2. So up until then objects in the
tracker fields were not expected in Deluge code

PR #1865, opened 2017-03-29
restore announce_entry's timestamp fields to posix time in python binding

arvidn:


arvidn:
@cas-- does this look right to you? Does this restore the 1.0 behavior?

cas--:
Yeah looks about right

PR #1871, opened 2017-03-30
fix m_host_resolver

pavel-pimenov:
V670 The uninitialized class member 'm_host_resolver' is used to initialize the 'm_tracker_manager' member.
  Remember that members are initialized in the order of their declarations inside a class. session_impl.cpp


arvidn:
would you mind pack-porting this to RC_1_1 as well?

pavel-pimenov:
RC_1_1 https://github.com/arvidn/libtorrent/pull/1877

Issue #1841, opened 2017-03-23
peer_connection related crashes


MassaRoddel:
master, recently
Windows, 64Bit
VS2015 64Bit Builds

I get crashes because of pointers to invalid peer_connection objects (likely already freed).

stack starts from torrent::remove_peer():
m_deferred_disconnect.post(m_ses.get_io_service(), [=]() {
			std::shared_ptr<torrent> t = weak_t.lock();
			if (t) t->on_remove_peers();
		});

more specific: it crashes in m_ses.close_connection(p); in on_remove_peer()

arvidn:
does it happen under any specific circumstances? shutdown or removing a torrent?
just to clarify, it's in the deferred removal callback it happens, right?

MassaRoddel:
Yes, it happens in the deferred callback so the call stack does help to find the reason. It seems that it happens when a torrent download is finished.

arvidn:
could it be that the torrent object is deleted before the callback fires perhaps?

MassaRoddel:
I don' think so. The last crash happened during normal download. Here is a call stack. Maybe there is a reference count problem and the object get already deleted in m_connections.erase(i) in on_remove_peers();

MooPolice.exe!std::_Ref_count_base::_Incref_nz() Zeile 75	C++
MooPolice.exe!std::_Ptr_base<libtorrent::peer_connection>::_Reset(libtorrent::peer_connection * _Other_ptr, std::_Ref_count_base * _Other_rep, bool _Throw) Zeile 386	C++
MooPolice.exe!std::_Ptr_base<libtorrent::peer_connection>::_Reset<libtorrent::peer_connection>(const std::_Ptr_base<libtorrent::peer_connection> & _Other, bool _Throw) Zeile 356	C++
MooPolice.exe!std::shared_ptr<libtorrent::peer_connection>::shared_ptr<libtorrent::peer_connection><libtorrent::peer_connection>(const std::weak_ptr<libtorrent::peer_connection> & _Other, bool _Throw) Zeile 510	C++
MooPolice.exe!std::enable_shared_from_this<libtorrent::peer_connection>::shared_from_this() Zeile 1126	C++
MooPolice.exe!libtorrent::peer_connection::self() Zeile 676	C++
MooPolice.exe!libtorrent::aux::session_impl::close_connection(libtorrent::peer_connection * p) Zeile 2888	C++
MooPolice.exe!libtorrent::torrent::on_remove_peers() Zeile 5387	C++
[...]

Issue #1806, opened 2017-03-12
https webseed


Col-blimp:
I seen a reference that says https web-seeds are supported so longs as built with openssl, have done so but https web-seeds get rejected as unsupported, is there a setting or a define that's needs for this to work?

arvidn:
yes, when building with ``b2``, specify ``crypto=openssl``. What it boils down to is to define ``TORRENT_USE_OPENSSL`` when building, as well as linking against libssl and libcrypto

Col-blimp:
```
error: "openssl" is not a known value of feature <crypto>
error: legal values: "built-in" "libcrypto" "libgcrypt"
```

Col-blimp:
Built as `crypto=libcrypto` then the libtorrent.lib will compile without error but then building the project you get:

```
1>client.obj : error LNK2001: unresolved external symbol "class boost::system::error_category const & __cdecl boost::asio::error::get_ssl_category(void)" (?get_ssl_category@error@asio@boost@@YAAEBVerror_category@system@3@XZ)
1>client.obj : error LNK2001: unresolved external symbol "class boost::system::error_category const & __cdecl boost::asio::ssl::error::get_stream_category(void)" (?get_stream_category@error@ssl@asio@boost@@YAAEBVerror_category@system@4@XZ)
```

if you remove `#define BOOST_ASIO_SEPARATE_COMPILATION` the error goes away but no sll support

arvidn:
it sounds like you're building master, in which case you say ``ssl=openssl crypto=libcrypto``

from where do you remove ``BOOST_ASIO_SEPARATE_COMPILATION``? the config.hpp header?

Col-blimp:
Building as: `toolset=msvc-14.0 boost-link=static asserts=production address-model=64 runtime-link=static deprecated-functions=off encryption=on ssl=openssl crypto=libcrypto link=static dht=on logging=off i2p=on variant=release`

gives the following:

```
CXXFLAGS =
LDFLAGS =
OS = NT
building boost from source directory:  C:\Boost-1.63-64(vs2015)\boost_1_63_0
Performing configuration checks

    - 32-bit                   : yes
    - arm                      : no
    - mips1                    : no
    - power                    : no
    - sparc                    : no
    - x86                      : yes
    - symlinks supported       : yes
...patience...
...patience...
...found 2677 targets...
...updating 163 targets...
```

basically the console closes with no explanation, it is master Im using yes.

Col-blimp:
[...]

PR #1872, opened 2017-03-30
schedule peer removal only if properly attached and handle special case of attach_peer

aldenml:


aldenml:
this should solve https://github.com/arvidn/libtorrent/issues/1841 and https://github.com/arvidn/libtorrent/issues/1806

arvidn:
oh, nice. you figured it out?

aldenml:
I hope so :)

src/torrent.cpp
```diff
@@ -5291,12 +5291,20 @@ namespace libtorrent
...
 			, m_peers_to_disconnect.end(), p) == 0);
 
-		std::weak_ptr<torrent> weak_t = shared_from_this();
-		m_peers_to_disconnect.push_back(p);
-		m_deferred_disconnect.post(m_ses.get_io_service(), [=]() {
-			std::shared_ptr<torrent> t = weak_t.lock();
-			if (t) t->on_remove_peers();
-		});
+		// we only schedule the peer for actual removal if in fact
+		// it is attached to this torrent, since this method can
```
arvidn: sorry for nit-picking. would you mind changing "method" to "function"? (or "member function")

aldenml: sure, I will add other typos

src/torrent.cpp
```diff
@@ -6769,7 +6778,7 @@ namespace libtorrent
 			// TODO: 2 if peer is a really good peer, maybe we shouldn't disconnect it
 			// perhaps this logic should be disabled if we have too many idle peers
 			// (with some definition of idle)
-			if (peer && peer->peer_rank() < p->peer_rank())
+			if (peer != nullptr && peer->peer_rank() < p->peer_rank())
```
arvidn: how can peer be null here?

aldenml: no idea, didn't inspect that logic, just changed it to `!= nullptr` because it's a raw pointer

src/torrent.cpp
```diff
@@ -6797,10 +6806,13 @@ namespace libtorrent
...
 #endif
 				p->disconnect(errors::too_many_connections, op_bittorrent);
-				// we have to do this here because from the peer's point of
+				// we have to do this here because from the peer's point of view
 				// it wasn't really attached to the torrent, but we do need
 				// to let peer_list know we're removing it
 				remove_peer(p);
+				// the peer is already inserted and needs to be removed
+				TORRENT_ASSERT(sorted_find(m_connections, p) != m_connections.end());
+				m_connections.erase(sorted_find(m_connections, p));
```
arvidn: doesn't this defeat the purpose of the original patch? I was under the impression that we want to be able to iterate over ``m_connection`` and disconnect peers. but erasing from it will break any such loop. won't the peer be removed from this list in the deferred function?

src/torrent.cpp
```diff
@@ -5291,12 +5291,20 @@ namespace libtorrent
...
-		m_peers_to_disconnect.push_back(p);
-		m_deferred_disconnect.post(m_ses.get_io_service(), [=]() {
-			std::shared_ptr<torrent> t = weak_t.lock();
-			if (t) t->on_remove_peers();
-		});
+		// we only schedule the peer for actual removal if in fact
+		// it is attached to this torrent, since this method can
+		// be called from the attach_peer path and fail to do so
+		// because of too many connections
+		if (p->associated_torrent().lock().get() == this)
```
arvidn: I don't really understand this comment. My understanding is that the peer ``p`` is part of the ``m_connections`` vector at this point, no? Is that not a prerequisite for this function?
would it be a more accurate test, to not add the peer to the deferred list if it hasn't been added to that list? iirc, m_connections is sorted, and fairly cheap to do a lookup in

src/torrent.cpp
aldenml: a few lines above L6748 you have `sorted_insert(m_connections, p);`. I thought that if this is a safe place to add to it (and invalidate the iterators) it should be safe to remove from it. This is important because at the end, the peer connection is not associated to the torrent, it's destroyed by the `session_impl::tick`, but when the times comes to call `torrent::disconnect_all`, some pointers in the loop are invalid

src/torrent.cpp
aldenml: indeed, `p` is part of the `m_connections`, but when coming via `peer_connection::attach_to_torrent`->`torrent::attach_peer`-->`maybe_replace_peer is true`, and due to ranking call `remove_peer(p);`, the `peer_connection* p` is left without associated torrent. Then `session_impl::tick` could clean it, and we end up with an invalid pointer in `m_connections`

aldenml:
@arvidn I hope may last two comments make sense, if not let me know

src/torrent.cpp
arvidn: oh, sorry. I missed that this check was already there

src/torrent.cpp
arvidn: I see. really the property to test for here is whether this peer connection will be kept alive until the deferred function is called or not. could you make it clearer in the comment that before a peer_connection is attached to a torrent, the session_impl object may remove it at any time, which may be while the non-owning pointer is waiting for the deferred function to be called.

I have a feeling that there's a simpler way to express this patch. I have two ideas:

1. Instead of removing the peer from ``m_connections`` in attach_peer(), could you do that in an else block to this if-statement? If that would work, that would seems a bit more robust.

2. Perhaps ``m_peers_to_disconnect`` should hold shared_ptr to the peers

src/torrent.cpp
arvidn: yes, I misread the context here. I think it would be nicer if remove_peer() would do this also

src/torrent.cpp
aldenml: at first, I did it like your idea in 1) but was like "too much magic" in `torrent::remove_peer`, but once the problem is clear, I will redo it that way.

about 2), I'm not sure, I like the idea of immediate destruction if not needed, if you keep a `shared_ptr` they will stored in `m_undead_peers` for next tick. Maybe for another PR.

aldenml:
@arvidn what do you think of this now?

arvidn:
lgtm

PR #1878, opened 2017-03-31
added resolver_cache_timeout setting

aldenml:
@arvidn do you think this is a reasonable setting addition?

### DIFF
--- a/src/resolver.cpp
+++ b/src/resolver.cpp
@@ -132,4 +132,10 @@ namespace libtorrent
 	{
 		m_resolver.cancel();
 	}
+
+	void resolver::set_cache_timeout(seconds timeout)
+	{
+		if (timeout >= seconds(0))
+			m_timeout = timeout;
+	}
 }
--- a/src/resolver.cpp
+++ b/src/resolver.cpp
@@ -137,5 +137,7 @@ namespace libtorrent
 	{
 		if (timeout >= seconds(0))
 			m_timeout = timeout;
+		else
+			m_timeout = seconds(0);
 	}
 }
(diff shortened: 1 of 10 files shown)


src/resolver.cpp
```diff
@@ -132,4 +132,10 @@ namespace libtorrent
 	{
 		m_resolver.cancel();
 	}
+
+	void resolver::set_cache_timeout(seconds timeout)
+	{
+		if (timeout >= seconds(0))
```
arvidn: I think it might be a bit confusing to ignore negative values, rather than interpreting them as 0 (or something).

Imagine I set the value to 100, then to -1. I can read the settings to say -1, but the effective timeout would still be 100.

arvidn:
seems reasonable to me

src/resolver.cpp
```diff
@@ -132,4 +132,10 @@ namespace libtorrent
...
+			m_timeout = timeout;
```
arvidn: where is ``m_timeout`` used? I would have expected a special case of it being zero, where the cache would be circumvented too.

src/resolver.cpp
aldenml: true, I recognize this is a bad idea, I will default to 0 for negative values

src/resolver.cpp
aldenml: it is used in `resolver::async_resolve`
```cpp
// keep cache entries valid for m_timeout seconds
if ((flags & resolver_interface::prefer_cache)
	|| i->second.last_seen + m_timeout >= aux::time_now())
{ ... }
```

PR #1879, opened 2017-04-01
use inline abi-version namespace

arvidn:
* renames the actual namespace to "lt".
* adds another interior, inline, namespace with ABI version
* removes the previous (inferior) link-time ABI check
* introduces a namespace alias "libtorrent" to preserve backwards compatibility


aldenml:
well...this is a really really bold bold change :)

Do you have some link when I can read more about this pattern of namespacing? thanks

aldenml:
and what is the goal or benefit here from the client point of view? better tracking of version? mix of versions? (just asking)

arvidn:
libstdc++ is an example, but this is really what inline namespaces were meant for. The main benefit for clients will be a much clearer error on abi incompatibility along with better abilities to introduce abi changes while staying backwards compatible.
Can you think of any significant risks with this change?
One issue is that clients can no longer (easily) forward declare libtorrent names

arvidn:
here's a stackoverflow post about it: http://stackoverflow.com/questions/11016220/what-are-inline-namespaces-for

arvidn:
perhaps it's misleading to put the internal functions and types in the versioned namespace, since those aren't part of the ABI anyway

aldenml:
@arvidn thanks for the information, I also read your email in sf.net, but I rather write here.

The namespace rename (to `lt`) is a matter of taste, convention, tradition, etc..., nothing material here.

Regarding the introduction of the ABI inline namespace. I think that _NOT_ using it for "aux" is important, because otherwise you are implicitly giving a guarantee of ABI compatibility and will be easier to have code repetition by not using the new "private" functions to reimplement old features (internally). Maybe doing the `lt` rename in a separated PR is better and will reduce the amount of changes in the PRs.

I was doing the mental exercise with `add_torrent_params`. Let say we are in `1.2` and in `1.3` you want to introduce significant changes. `add_torrent_params` is know to have very deep relationships with `session_impl`, in a way that even if `session_handle` does not change the interface, the ABI is incompatible. My point and questions is: Are we about to end up with big code copies of big classes, like `session` and `session_impl` just for `add_torrent_params`?

More generally, what kind of changes, in terms of size and ABI break ramifications, are tolerated with this technique? what will be the version policy to accommodate for the not aceptable changes?

arvidn:
the only reason I wanted to rename the top-level namespace was to try to keep the symbol names from getting too long, since the additional inline namespace will make them longer. But, right, there's no material change there (except it occupies a new name in the global namespace).

You are right that the intention is not to version internal types and functions (i.e. the ones in the ``aux`` namespace). By default (at least when building with boost-build) all non-public functions have hidden visibility. However, there are some symbols in the aux-namespace that are part of the ABI, but wrapped by a template (or some other inline function). Any such ``aux`` function/type *would* have to be versioned. I believe ``set_piece_hashes()`` used to do this. In general though, things in the ``aux`` namespace are not part of the library interface.

However, I don't think there's any cost in applying the version namespace uniformly everywhere, their names presumably wouldn't be generated as part of the interface of a shared library anyway. I don't expect any internal types would need duplicate definintions, like ``session_impl`` for instance. Not even ``session`` or ``session_handle`` would need to change, since the session is a pimpl, so there's no ABI breaking changes (just potentiallly additional functions)

aldenml:
Agree with the `aux` issue but, I'm not sure if you are addressing my last two paragraphs/questions already. Are you?

arvidn:
> More generally, what kind of changes, in terms of size and ABI break ramifications, are tolerated with this technique? what will be the version policy to accommodate for the not aceptable changes?

I'm not sure I understand exactly what you're asking, but: For now I still expect to continue a similar pattern of which versions that are allowed to break the ABI. Since it's always possible to duplicate symbols, in theory any ABI upgrades can provide backwards-compatible versions. This makes ABI-stable upgrades more flexible too.

Perhaps it will prove flexible enough to not have any ABI breaking releases.

I would not expect to preserve ABI backwards compatibility indefinitely though, and probably to have some build switches to select which versions to support.

aldenml:
I'm starting to see the benefits of this, but I would like to clarify some details. Let me try with another hypothetical example.

Let's suppose that in `1.3` you find a way to better implement `span<T>` but it breaks the ABI, If you want to keep ABI compatibility with `1.2` in the same library, I assume that a duplicate is necessary, but how that affects the rest of the classes? does it requires a duplication of all affected symbols (because `1.2::add_torrent_params` will be no compatible with `1.3::add_torrent_params`)?

Is the above change too big? what's the size that will be acceptable  to duplicate?

If no practical benefits exists of mixing `1.2::span<T>` with `1.3::span<T>`, are we talking only about better detect "linking problems"?

arvidn:
perhaps it doesn't make sense to have a single ABI version for everything in the library. If each class could have its own version (which certainly is possible, but the macros in this PR wouldn't be helpful), there would be a pretty straight-forward solution to your case.

1. a new ``span`` class is defined in a new abi version namespace.
2. all existing functions that take a span is overloaded to take the old span as well, and forward to the new-span overload.

However, if something fundamental causes almost all of the ABI to change, it's not obvious is worth trying to maintain backwards compatibility in the same library. Bumping the "global" abi version would still improve error messages though.

aldenml:
right, I didn't see the overload solution, mainly because I was thinking about member variables like in `add_torrent_params`.

Now, I think overall this is an improvement and I'm pretty sure common sense will prevail when moving from `x.y` to `x.(y+1)`.

As a final suggestion, maybe you can do this in parts (separated PRs)? like:
1-rename `libtorrent` to `lt`.
2-inline ABI only for well defined public API,
3-inline ABI for those internals that are needed by public templates (not convinced that all private in inline ABI is a good idea).

evsh:
@sledgehammer999 raised in the mailing list the question with forward declarations that are used in qBittorrent. While I myself compile on Linux and with libtorrent headers (1.5 MiB) compilation speed is not an issue here, this indeed might further hurt those on Windows. Could it be possible to add forward declarations for public API to libtorret itself, producing a file similar to `<iosfwd>`?

glassez:
@arvidn, can we forward declare libtorrent types this way:
```c++
namespace lt {
LIBTORRENT_VERSION_NAMESPACE {
    class some_libtorrent_class;
}}
```
?

glassez:
```c++
#define QBT_LIBTORRENT_NAMESPACE_BEGIN namespace lt { LIBTORRENT_VERSION_NAMESPACE {
#define QBT_LIBTORRENT_NAMESPACE_END }}

QBT_LIBTORRENT_NAMESPACE_BEGIN
    class some_libtorrent_class;
QBT_LIBTORRENT_NAMESPACE_END

// and then
namespace libt = lt::LIBTORRENT_VERSION_NAMESPACE_NAME;
libt::some_libtorrent_class obj;
```

glassez:
>Could it be possible to add forward declarations for public API to libtorret itself, producing a file similar to <iosfwd>?

IMO, this is a solution too.

evsh:
Forward declarations in this style become so verbose, that we will likely join them into a single file on one day. Why not distribute that with libtorret itself then?

arvidn:
I think a better approach than a sweeping change like this is one where the version namespace is introduced as needed. see https://github.com/arvidn/inline-namespace-abi-versioning

Though, I can't think of a way to allow forward declarations in clients, and still support robust ABI versioning. I think forward declaration headers is probably a reasonable approach for this (until we have modules in C++20 :) )

Would it make sense to have a single forward declaration header for every type in libtorrent? Presumably only types should be forward declared, not functions. So there shouldn't be any external dependencies for this header (other than ``config.hpp`` for build config, ``export.hpp`` for ``TORRENT_EXPORT`` attribute and ``version.hpp`` for abi namespace).

Any volunteers? :)

arvidn:
I'm closing this as making all these changes at once is moving too fast. I will make less radical changes for the next major release and the gradually get to something like this.

PR #1880, opened 2017-04-01
enable warnings when building tests

arvidn:


Jamfile
```diff
@@ -237,6 +237,7 @@ rule warnings ( properties * )
 		result += <cxxflags>-Wno-c++98-compat-pedantic ;
 		result += <cflags>-Wno-padded ;
 		result += <cflags>-Wno-global-constructors ;
+		result += <cflags>-Wheader-hygiene ;
```
aldenml: I would expect this already enable by `-Weverything`

arvidn: hm.. yeah, that makes sense. I tried building with this warning enabled specifically yesterday, and I saw it fire in the tests. but I realize now that the tests were never built with all the warnings enabled to begin with

PR #1881, opened 2017-04-02
bump version number

arvidn:


arvidn:
@cas-- do you know of any outstanding issues that should be fixed before 1.1.3?

cas--:
Not that I currently know of :)

PR #1883, opened 2017-04-03
python bindings: fix "TypeError: 'dict_keys' object does not support

chutz:
This fixes a problem when using the bindings with python3 when passing
python dictionaries will result in the error:
```
TypeError: 'dict_keys' object does not support indexing
```
This is because the keys() method of a dictionary no longer returns an
object that is indexable. Instead of indexing, use the iterator's
`__next__` method to get the next object the appropiate number of times.

### DIFF
--- a/bindings/python/src/converters.cpp
+++ b/bindings/python/src/converters.cpp
@@ -132,11 +132,16 @@ struct dict_to_map
         dict o(borrowed(x));
         std::map<T1, T2> m;
 
-        list iterkeys = (list)o.keys();
-        int const len = int(boost::python::len(iterkeys));
-        for (int i = 0; i < len; i++)
+        const object dictkeys = o.keys();
+        const size_t len = size_t(boost::python::len(dictkeys));
+        const object dictkeys_iter = dictkeys.attr("__iter__")();
+        for (size_t i = 0; i < len; i++)
         {
-            object key = iterkeys[i];
+#if PY_VERSION_HEX >= 0x03000000
+            object key = dictkeys_iter.attr("__next__")();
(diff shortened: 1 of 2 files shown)


arvidn:
great! would you mind adding a test case to ``test.py`` that exercises this too? (otherwise it's likely to break again in the future). I'm not sure any of the CI runs python3 now, I'm under the impression that appveyor does, but just making sure the functions are exercised would be good either way.

chutz:
This code is exercised by existing tests, I can write more test code to exercise this stuff if you like. We have been having trouble reproducing the error we have been seeing in a minimal test case. When we run the full application, we see the error but a minimal test case seems to work fine.

bindings/python/src/converters.cpp
```diff
@@ -132,11 +132,16 @@ struct dict_to_map
         dict o(borrowed(x));
         std::map<T1, T2> m;
 
-        list iterkeys = (list)o.keys();
-        int const len = int(boost::python::len(iterkeys));
-        for (int i = 0; i < len; i++)
+        const object dictkeys = o.keys();
```
arvidn: is this a better (version agnostic) way of doing the same thing?
``stl_input_iterator<object> begin(o), end;``
and iterate over them

arvidn:
does this patch against 1.1 look reasonable? I will extend it to cover the new code when I merge it into master. https://github.com/arvidn/libtorrent/pull/2017

arvidn:
#2017 was pushed to RC_1_1, merged to master and extended to cover both places this PR addresses.

PR #1885, opened 2017-04-04
add full multi-home support to DHT, UTP, and tracker announces

ssiloti:
This is a big one because once the `session_impl` changes went in everything else had to come too. I'm most unsure about the tracker announce logic. It's not obvious to me what the re-announce policy should be when announcing to trackers on multiple interfaces.


ssiloti:
Hmm, looks like a template depth limit of 128 is a little too aggressive for this. Building locally I'm seeing the offending line in dht_tracker.cpp use a depth of 126, but I'm using a different gcc and libstdc++ version so that could account for the variation.

@arvidn do you know where that option is getting set?

arvidn:
boost-build sets it to something like 128 by default. I believe this comes from the times when the default was a lot lower.

From this [patch](https://svn.boost.org/trac/boost/changeset/47727) it looks like there is a feature called ``c++-template-depth``. It would probably make sense to add this to the ``user-config.jam`` for gcc and clang in ``.travis``

arvidn:
don't forget this too: https://travis-ci.org/arvidn/libtorrent/jobs/218310921#L746

include/libtorrent/announce_entry.hpp
```diff
@@ -37,12 +37,41 @@ POSSIBILITY OF SUCH DAMAGE.
...
+
+		// if this tracker has returned an error or warning message
+		// that message is stored here
+		std::string message;
+
+		// if this tracker failed the last time it was contacted
+		// this error code specifies what error occurred
+		error_code last_error;
+
+		aux::session_listen_socket* socket;
```
aldenml: this leaks a private type in public API

aldenml:
(this is quite a big PR and not very easy to follow in the edge cases)

@ssiloti I have a question, in the past, having one IPv4 nid and one IPv6 nid facilitated the reuse of the node id if it was still valid for the interface, but I fail to see a similar logic here. I expected something like assigning/selecting the "nid" from the state that improve such a possibility. Does my idea makes sense?

include/libtorrent/http_connection.hpp
```diff
@@ -102,7 +103,7 @@ struct TORRENT_EXTRA_EXPORT http_connection
 	void get(std::string const& url, time_duration timeout = seconds(30)
 		, int prio = 0, aux::proxy_settings const* ps = 0, int handle_redirects = 5
 		, std::string const& user_agent = std::string()
-		, address const& bind_addr = address_v4::any()
+		, boost::optional<address> const& bind_addr = boost::optional<address>()
```
arvidn: +1. I've recently started to use ``optional`` again, and part of me regret removing them from libtorrent

include/libtorrent/announce_entry.hpp
arvidn: assuming this isn't meant to be public, you could derive an internal class from the public one

arvidn: I mean, create another, internal, type that derives from ``announce_endpoint`` and add any private field into the derived class.

include/libtorrent/announce_entry.hpp
```diff
@@ -62,13 +91,7 @@ namespace libtorrent
...
 		std::string trackerid;
 
-		// if this tracker has returned an error or warning message
-		// that message is stored here
-		std::string message;
-
-		// if this tracker failed the last time it was contacted
-		// this error code specifies what error occurred
-		error_code last_error;
+		std::vector<announce_endpoint> endpoints;
```
arvidn: Do you think it would make sense (and simplify anything) to flatten this structure, and elevate each individual endpoint to the top level list of trackers? In some way endpoints can be treated as trackers in the same tier. But I suppose there's some special handling of endpoints with different endpoint families.

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -118,8 +118,40 @@ namespace libtorrent
 		class item;
 	}
 
-	struct listen_socket_t
+	struct session_udp_socket : utp_socket_interface
 	{
+		session_udp_socket(io_service& ios)
+			: sock(ios) {}
+
+		virtual udp::endpoint local_endpoint() override { return sock.local_endpoint(); }
```
arvidn: The cppCoreGuidelines recommends using only one of "virtual" or "override". Since "override" implies virtual. In fact, I would have thought cpplint.py would have flagged this

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -118,8 +118,40 @@ namespace libtorrent
...
+	{
+		outgoing_udp_socket(io_service& ios)
+			: session_udp_socket(ios) {}
+
+		// the name of the device the socket is bound to, may be empty
+		// if the socket is not bound to a device
+		std::string device;
+	};
+
+	struct listen_socket_t : dht::dht_socket, aux::session_listen_socket
```
arvidn: reading this, it's not obvious what the different socket types are for. would you mind adding some comments? or do you think it's possible to collapse any of them?

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -558,7 +585,17 @@ namespace libtorrent
 			void set_peer_id(peer_id const& id);
 			void set_key(std::uint32_t key);
 			std::uint16_t listen_port() const override;
+			std::uint16_t listen_port(listen_socket_t* sock) const;
 			std::uint16_t ssl_listen_port() const override;
+			std::uint16_t ssl_listen_port(listen_socket_t* sock) const;
+
+			void for_each_listen_socket(std::function<void(aux::session_listen_socket*)> f) override
```
arvidn: could this function object take a reference instead of a pointer? It can never be null, right?

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -899,6 +942,12 @@ namespace libtorrent
 			// socket fails.
 			std::vector<std::string> m_outgoing_interfaces;
 
+			// sockets used for outoing utp connections
+			// sockets are created for each device name or IP address in
+			// m_outgoing_interfaces
+			// they are used in a round-robin fashion
+			std::vector<std::shared_ptr<outgoing_udp_socket>> m_outgoing_sockets;
```
arvidn: do you think it would make sense to wrap this, along with the round-robin indices into its own class, and have that as a member here? (I've been trying to think of ways to make session_impl have fewer lines of code). It should also make it easier to unit test.

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -1050,21 +1102,43 @@ namespace libtorrent
...
-			void send_udp_packet_hostname(char const* hostname
+			void send_udp_packet_hostname(std::weak_ptr<utp_socket_interface> sock
+				, char const* hostname
 				, int port
 				, span<char const> p
 				, error_code& ec
 				, int flags);
 
-			void send_udp_packet(bool ssl
+			void send_udp_packet_hostname_listen(aux::session_listen_socket* sock
```
arvidn: what is the listen socket for here?

arvidn:
very exciting patch!
if it helps, keep in mind that I've removed the Socks5 "listen" support

aldenml:
Why is it better for utp to have round-robin instead of always use the same interface? Should this be configurable?

simulation/setup_dht.cpp
```diff
@@ -182,6 +185,18 @@ struct dht_node final : lt::dht::udp_socket_interface
 		return true;
 	}
 
+	address get_external_address() override
+	{
+		return get_local_address();
```
arvidn: this looks unintuitive. external address returns local address?

include/libtorrent/http_connection.hpp
aldenml: I hope that feeling fades without any action :)

arvidn: do you prefer sentinel values? my experience is that sentinel values encourages half-constructed objects and possibly two-phase initialization and then it's just downhill from there :)

simulation/test_tracker.cpp
```diff
@@ -262,6 +262,8 @@ TORRENT_TEST(announce_interval_1200)
 
 struct sim_config : sim::default_config
 {
+	explicit sim_config(bool ipv6 = true) : ipv6(ipv6) {}
```
arvidn: would you mind doing something like this instead? (or maybe even use ``boost::asio::ip::protocol``)

```
enum class ip_protocol { ipv4, ipv6 };

explicit sim_config(ip_protocol) ...
```

arvidn: actually, I just realized this is local to this test, it probably doesn't matter

src/http_connection.cpp
```diff
@@ -530,13 +530,28 @@ void http_connection::on_resolve(error_code const& e
...
-	// sort the endpoints so that the ones with the same IP version as our
-	// bound listen socket are first. So that when contacting a tracker,
-	// we'll talk to it from the same IP that we're listening on
-	if (m_bind_addr != address_v4::any())
-		std::partition(m_endpoints.begin(), m_endpoints.end()
+	// if we have been told to bind to a particular address
+	// only connect to addresses of the same family
+	if (m_bind_addr)
+	{
+		auto new_end = std::partition(m_endpoints.begin(), m_endpoints.end()
```
arvidn: do we still need this logic? I got the impression that you would expand all these endpoints into distinct announces, promoted to the announce_entry object.

src/kademlia/dht_state.cpp
```diff
@@ -40,12 +40,40 @@ namespace dht
...
+			return ret;
+		}
+		auto nids = e.dict_find_list(key);
+		if (!nids) return ret;
+		for (int i = 0; i < nids.list_size(); i++)
+		{
+			bdecode_node nid = nids.list_at(i);
+			if (nid.type() != bdecode_node::string_t) continue;
+			if (nid.string_length() < 20) continue;
+			char const* in = nid.string_ptr();
```
arvidn: on a separate note, looking at this, I wonder if there should be a way to get a ``span<char const>`` back here. I believe there's already a way to get a ``string_view``, but since this is not actually a string, and we parse stuff out from it, a ``span`` probably would make more sense.

src/kademlia/dht_state.cpp
```diff
@@ -40,12 +40,40 @@ namespace dht
 {
 namespace
 {
-	node_id extract_node_id(bdecode_node const& e, string_view key)
+	node_ids_t extract_node_ids(bdecode_node const& e, string_view key)
```
arvidn: I wonder if it would make sense to give this function external linkage, so we could have a test for it.

arvidn:
would it be possible to slice this patch up into smaller independent units? I haven't fully grasped it yet, but perhaps separating fundamental session_impl changes, DHT, uTP and trackers into separate patches.

include/libtorrent/http_connection.hpp
ssiloti: In this case optional is needed to distinguish between "no specific binding requested" and "binding to 0.0.0.0 requested".

ssiloti:
@aldenml Node ids are still saved and restored. See references to `dht_state::nids`. Node ids are associated with interfaces based on the local address. See code in `dht_tracker::new_socket`.

include/libtorrent/announce_entry.hpp
ssiloti: My intent was that `aux::session_listen_socket` would be an opaque type to clients. I can change the type of the pointer here if you're worried about clients "peeking behind the curtains" by including `aux/session_interface.hpp` to get the complete type definition.

include/libtorrent/aux_/session_impl.hpp
ssiloti: I used a pointer here because other places in the code use such pointers as handles. I could pass a reference here but then the function would just end up converting it to a pointer.

include/libtorrent/aux_/session_impl.hpp
ssiloti: Because we need to specify which socket to send the packet with. Previously we just picked the first socket with a suitable address family.

simulation/setup_dht.cpp
ssiloti: It's just for testing. The local and external addresses only differ if there's NAT and that's not what we're testing here.

src/http_connection.cpp
ssiloti: We generate separate announces for each one of the _client's_ listen sockets. This is concerning the set of addresses the tracker(s) are listening on so the logic still applies. If we generated separate announces for each tracker address there would be M*N announces for each listen socket and tracker address which I don't think is what we want.

include/libtorrent/announce_entry.hpp
ssiloti: I'm still weighing this. On the upside flattening the list makes it easier to maintain API compatibility in the python bindings, but it also makes the public API more awkward since clients are likely to want to group trackers by url for displaying to the user. It creates some weird corner cases, like if for some reason all listen sockets went away momentarily we might lose the tracker information. It would also result in the duplication of tracker specific fields like url, trackerid, teir, and source which annoys the DBA in me. If I don't flatten the list I think I'll at least move all of the dynamic state to `announce_endpoint` and treat them as separate trackers for the purpose of announcing.

ssiloti: A problem with treating every (tracker, listen socket) pair as a separate tracker is it interacts poorly with the `announce_to_all_tiers` setting. If it is set to true we will only announce to a single tracker with a single listen socket per tier, which is not what we want. Perhaps the setting should be deprecated?

include/libtorrent/aux_/session_impl.hpp
ssiloti: The main reason I make a separate type for `dht_socket` was to maintain loose coupling between the dht and session implementations. Now that I look back at it I could just make a separate type not tied to either module.

include/libtorrent/announce_entry.hpp
arvidn: I didn't realize the list was (tracker, listen_socket) pairs, I thought it was just a list of each endpoint a tracker hostname resolved to. I think you make good points to keep these as a list on the tracker then.
regarding the ``announce_to_all_tiers``, I don't believe it's possible to deprecate that. My impression is that people creating torrents don't understand the announce rules defined by the multi-tracker BEP enough to create reasonable torrents, and since torrents won't work well without it, clients default to announcing to all in parallel.
(unless of course you mean to deprecate the original behavior of announcing to trackers one tier at a time. I would be a bit worried about people actually relying on that behavior, since it's quite useful and pretty well specified).

That said, the existing logic in libtorrent to iterate over trackers and decide which one to announce to is a mess (and quite error prone), and I apologize if you've had to read that :/ I've made a few attempts to clean it up a bit.

include/libtorrent/announce_entry.hpp
arvidn: I don't think there's necessarily a problem to expose the pointer value to clients, but I would prefer if you could define a new ``strong_typedef`` for the public type (in fact, I should probably add another kind of typedef that completely disables arithmetic for things like this). That way the fact that the type is opaque is well encoded in the type system.

include/libtorrent/aux_/session_impl.hpp
arvidn: I see, that makes sense. This will affect the source port, but not necessarily the source IP, right? because which IP interface to use will still (in the normal case) be determined by the system's routing table, right? I suppose if we have a working bind-to-device, that might not be the case.

.travis.yml
```diff
@@ -146,21 +146,21 @@ script:
 # libtorrent is the name of the test suite target
   - cd test
   - 'if [ "$tests" == "1" ]; then
-      travis_retry bjam -j3 warnings-as-errors=on ssl=$ssl crypto=$crypto debug-iterators=on picker-debugging=on invariant-checks=full $toolset variant=$variant libtorrent test_natpmp enum_if -l300 &&
+      travis_retry bjam -j3 c++-template-depth=512 warnings-as-errors=on ssl=$ssl crypto=$crypto debug-iterators=on picker-debugging=on invariant-checks=full $toolset variant=$variant libtorrent test_natpmp enum_if -l300 &&
```
arvidn: if you add this to the "~/user-config.jam" file instead, you only have to mention it in 3 places. For ``darwin``, ``gcc`` and ``gcc-sanitizer``. I think it might make this script read a bit easier too (come to think of it, I should probably move some other common features into the bjam config file)

arvidn: i.e. you would put something like this in the ``echo`` statements:
``<c++-template-depth>512``

examples/client_test.cpp
```diff
@@ -1696,14 +1696,17 @@ MAGNETURL is a magnet link
 			{
 				for (announce_entry const& ae : h.trackers())
 				{
+					auto best_ae = std::min_element(ae.endpoints.begin(), ae.endpoints.end()
```
arvidn: afaict, this prints the first endpoint with the lowest fail counter. I imagine in the future it might be reasonable to print all of them, right?

include/libtorrent/announce_entry.hpp
```diff
@@ -37,12 +37,41 @@ POSSIBILITY OF SUCH DAMAGE.
...
 
 #include <string>
 #include <cstdint>
+#include <vector>
 
 namespace libtorrent
 {
+	namespace aux { struct session_listen_socket; }
+
+	struct TORRENT_EXPORT announce_endpoint
```
arvidn: since this is a public type, would you mind adding a comment documenting it, above this declaration? The documentation tool will pick this up

include/libtorrent/aux_/session_impl.hpp
ssiloti: I understand that without bind-to-device the interface may be selected based on the routing table, but I thought the source IP was still controlled by the socket's bound address. In the general case the source IP is all we care about here.

examples/client_test.cpp
ssiloti: Yeah, I must admit I wrote this just to make the problem go away for now with the least amount of effort. We can go back later and enhance the tracker display to handle multiple endpoints better.

ssiloti:
@arvidn I'll take a shot at splitting this up into separate commits for the DHT, UTP, and tracker parts once all the other issues have been settled.

include/libtorrent/announce_entry.hpp
ssiloti: I think I've found a reasonably clean solution for this. Conceptually each (torrent, listen socket) pair is treated like a separate torrent which does it's own traversal of the tracker list. For efficiency reasons the code traverses all listen sockets for a single torrent at once while keeping a vector of states for each socket.

ssiloti:
Now that all unit tests are passing, I'm going to split this up into separate PRs for each subsystem (session, UTP, DHT, trackers).

PR #1886, opened 2017-04-04
minor code cleanup, std:: prefix, consts, typos

aldenml:


aldenml:
@arvidn did you have an opportunity to enable "auto cancelation" in travis? (https://blog.travis-ci.com/2017-03-22-introducing-auto-cancellation)

arvidn:
Yes, I definitely attempted to enable it

src/utp_stream.cpp
```diff
@@ -158,7 +158,7 @@ enum
 // into account. if lhs is close to UINT_MAX and rhs
 // is close to 0, lhs is assumed to have wrapped and
 // considered smaller
-TORRENT_EXTRA_EXPORT bool compare_less_wrap(std::uint32_t lhs
+bool compare_less_wrap(std::uint32_t lhs
```
arvidn: there's no unit test for this function?

arvidn:
lgtm, apart from ``compare_less_wrap``. I would prefer to add tests for it

aldenml:
Ok, I just only removed the `TORRENT_EXTRA_EXPORT` from the `.cpp`, I don't see any test already, I will add one.

arvidn:
thanks!

PR #1888, opened 2017-04-05
using unique_ptr for m_ip_notifier in session_impl

aldenml:
@arvidn the context of this PR is because I need to prepare the code first, in order to be able to hide all the platform dependent code in the compilation unit. This way it should be easy to split `ip_change_notifier` per platform and easy the review of new code.


arvidn:
lgtm. what do you think about moving: ``include/libtorrent/ip_notifier.hpp`` into the ``include/libtorrent/aux_`` directory in the future?

aldenml:
you mean, like right now? :)

PR #1890, opened 2017-04-06
moved current ip_change_notifier implementation to compilation unit

aldenml:
following a conservative approach here

### DIFF
--- a/include/libtorrent/aux_/ip_notifier.hpp
+++ b/include/libtorrent/aux_/ip_notifier.hpp
@@ -34,50 +34,21 @@ POSSIBILITY OF SUCH DAMAGE.
 #define TORRENT_IP_NOTIFIER_HPP_INCLUDED
 
 #include <functional>
-#include <array>
+#include <memory>
 
 #include "libtorrent/error_code.hpp"
-#include "libtorrent/address.hpp"
 #include "libtorrent/io_service.hpp"
-
-#if defined TORRENT_BUILD_SIMULATOR
-// TODO: simulator support
-#elif TORRENT_USE_NETLINK
-#include "libtorrent/netlink.hpp"
-#elif defined TORRENT_WINDOWS
-#include "libtorrent/aux_/disable_warnings_push.hpp"
-#include <boost/asio/windows/object_handle.hpp>
-#include "libtorrent/aux_/disable_warnings_pop.hpp"
-#endif
 
 namespace libtorrent { namespace aux
 {
-	struct ip_change_notifier
+	struct TORRENT_EXTRA_EXPORT ip_change_notifier
 	{
-		explicit ip_change_notifier(io_service& ios);
-		~ip_change_notifier();
-		ip_change_notifier(ip_change_notifier const&) = delete;
-		ip_change_notifier& operator=(ip_change_notifier const&) = delete;
-
 		// cb will be invoked  when a change is detected in the
 		// system's IP addresses
-		void async_wait(std::function<void(error_code const&)> cb);
-		void cancel();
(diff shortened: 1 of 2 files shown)


src/ip_notifier.cpp
```diff
@@ -63,21 +107,21 @@ namespace libtorrent { namespace aux
...
 
-	void ip_change_notifier::async_wait(std::function<void(error_code const&)> cb)
+	void ip_change_notifier_impl::async_wait(std::function<void(error_code const&)> cb)
 	{
 #if defined TORRENT_BUILD_SIMULATOR
 		TORRENT_UNUSED(cb);
 #elif TORRENT_USE_NETLINK
 		m_socket.async_receive(boost::asio::buffer(m_buf)
-			, std::bind(&ip_change_notifier::on_notify, this, _1, _2, cb));
+			, std::bind(&ip_change_notifier_impl::on_notify, this, _1, _2, cb));
```
arvidn: would it work to say: ``std::move(cb)`` here?
 suspect it would

src/ip_notifier.cpp
```diff
@@ -106,7 +150,7 @@ namespace libtorrent { namespace aux
 #endif
 	}
 
-	void ip_change_notifier::on_notify(error_code const& ec
+	void ip_change_notifier_impl::on_notify(error_code const& ec
 		, std::size_t bytes_transferred
 		, std::function<void(error_code const&)> cb)
```
arvidn: and if you would take ``cb`` by ``const&``, it would likely avoid a copy too

src/ip_notifier.cpp
```diff
@@ -63,21 +107,21 @@ namespace libtorrent { namespace aux
 #endif
 	}
 
-	ip_change_notifier::~ip_change_notifier()
+	ip_change_notifier_impl::~ip_change_notifier_impl()
 	{
 #if defined TORRENT_WINDOWS && !defined TORRENT_BUILD_SIMULATOR
 		cancel();
 		m_hnd.close();
```
arvidn: it seems a bit redundant to call ``close`` immediately before destructing the handle

include/libtorrent/aux_/ip_notifier.hpp
```diff
@@ -34,50 +34,21 @@ POSSIBILITY OF SUCH DAMAGE.
...
-#elif defined TORRENT_WINDOWS
-#include "libtorrent/aux_/disable_warnings_push.hpp"
-#include <boost/asio/windows/object_handle.hpp>
-#include "libtorrent/aux_/disable_warnings_pop.hpp"
-#endif
-
 namespace libtorrent { namespace aux
 {
-	struct ip_change_notifier
+	struct TORRENT_EXTRA_EXPORT ip_change_notifier
```
arvidn: ``TORRENT_EXTRA_EXPORT`` would only be necessary for a test to use this class. I can't think of a way to have an automated test for this though, so I'm not sure why this would be here

aldenml: I hope to provide something in the simulator and in macOS

aldenml:
I read your comments, but I'm trying to introduce as less changes in this PR as possible, you tell me if those are required for the merge. But any comment here I will keep it for the separation per platform. You tell me.

arvidn:
yeah, sorry. none of my comments are related to your patch :/

aldenml:
OK!

arvidn:
lgtm

aldenml:
@arvidn can you take a look at https://travis-ci.org/arvidn/libtorrent/jobs/219113180#L2644? thanks

arvidn:
that's a flaky assert :(

PR #1891, opened 2017-04-06
implemented ip_change_notifier_macos

aldenml:
After more research on this, I came to the conclusion that listening to the routing/interface changes is the more adequate solution for macOS (instead of listening for reachability). Now I'm using the Grand Central Dispatch (GCD) framework, much better than loops.

I wasn't able to provide unit tests for it (left the `TORRENT_EXTRA_EXPORT` just in case), but I added a new tool `session_log_alerts` that easy the task of manually testing the feature. I tested in my environment with a physical link, Wi-Fi, and a software VPN, and all combinations of disconnect/connect reported the appropriate log `"log_alert - received ip change"`

If this get's merged, I have plans to implement `ip_change_notifier_ios` with a setting to force it's use in macOS. This will be based on reachability since iOS does not support real multi-home, just the extension "Multipath TCP" for some apps .


src/ip_notifier.cpp
```diff
@@ -55,11 +58,105 @@ namespace
...
+
+	return success ? store : nullptr;
+}
+
+struct ip_change_notifier_macos final : ip_change_notifier
+{
+	explicit ip_change_notifier_macos(io_service& ios)
+		: m_ios(ios)
+	{
+		m_queue = dispatch_queue_create("libtorrent.IPChangeNotifierQueue", nullptr);
```
arvidn: you create the queue here, but you don't free it anywhere. is that right?
If responsibility to free it is handed over to the DynamicStore, what if constructing that fails?
perhaps I'm a bit pedantic, but I think this would be simpler if each thing was owned by a handle

aldenml: it's my understanding that ARC takes care of that, in fact, you can't call `CFRelease` on it. Do you mind elaborate the "owned by a handle" idea?

arvidn: I was under the impression that ARC was an Objective-C feature, not implemented by C++ in clang/gcc. by *handle* I mean something like:

```
template <typename T>
struct CFRef {
   explicit CFRef(T h) : m_h(h) {} // take ownership
   ~CFRef() { CFRelease(m_h); }
   CFRef(CFRef const& rhs) : m_h(rhs.m_h) { CFAcquire(m_h); }
   CFRef& operator=(CFRef const& rhs) {
      if (rhs == *this) return *this;
      CFRelease(m_h);
      m_h = rhs.m_h;
      CFAcquire(m_h);
   }
};
```

src/ip_notifier.cpp
```diff
@@ -55,11 +58,105 @@ namespace
 	// TODO: ip_change_notifier_sim
 #elif TORRENT_USE_NETLINK
 	// TODO: ip_change_notifier_nl
+#elif TORRENT_USE_SYSTEMCONFIGURATION
+// see https://developer.apple.com/library/content/technotes/tn1145/_index.html
+
+CFArrayRef create_keys_array()
+{
+	CFMutableArrayRef keys = CFArrayCreateMutable(nullptr
```
arvidn: As far as I can tell, ``key`` follow the "Create rule", meaning you need to call ``CFRelease()`` on it at some point

src/ip_notifier.cpp
```diff
@@ -55,11 +58,105 @@ namespace
...
+	SCDynamicStoreContext context = {0, nullptr, nullptr, nullptr, nullptr};
+	context.info = context_info;
+	SCDynamicStoreRef store = SCDynamicStoreCreate(nullptr
+		, CFSTR("libtorrent.IPChangeNotifierStore"), callback, &context);
+	if (store == nullptr)
+		return nullptr;
+
+	CFArrayRef keys = create_keys_array();
+	bool success = SCDynamicStoreSetNotificationKeys(store, nullptr, keys);
+	CFRelease(keys);
```
arvidn: I see, you release it here

src/ip_notifier.cpp
arvidn: if ``m_h`` can have a null-state, it would be simple to make it movable as well.

src/ip_notifier.cpp
```diff
@@ -55,11 +58,105 @@ namespace
...
+
+	return success ? store : nullptr;
```
arvidn: as far as I can tell, ``store`` is never released.

src/ip_notifier.cpp
aldenml: OK, let me review the memory handling of this objects and more about ARC in generic clang/gcc.

aldenml:
@arvidn thanks for your `CFRef` idea, I applied it and I think it's a lot more robust now, waiting for your review.

src/ip_notifier.cpp
```diff
@@ -61,17 +61,53 @@ namespace
...
+
+private:
+	T m_h = nullptr; // handle
+
+	void retain() { if (m_h != nullptr) Retain(m_h); }
+	void release() { if (m_h != nullptr) Release(m_h); m_h = nullptr; }
+};
+
+void CFDispatchRetain(dispatch_queue_t q) { dispatch_retain(q); }
+void CFDispatchRelease(dispatch_queue_t q) { dispatch_release(q); }
```
arvidn: oh, they have different kind of refcounting for dispatch queues?

arvidn:
lgtm

PR #1892, opened 2017-04-06
steps towars making file_pool private to disk_io_thread

arvidn:


aldenml:
will `file_pool` eventually be inside `aux::`?

arvidn:
as long as it's part of the storage customization point, it has to be public. But once 1.2 is cut (probably shortly after steven's multi-home patch lands) I intend for the disk_interface to be the new customization point. This is prepping for making ``file_pool``an implementation detail of ``disk_io_thread``.

Unless it's obvious, the main purpose of this is to raise the abstraction level of the disk I/O, to be able to encompass mmap I/O.

PR #1896, opened 2017-04-09
added limited support for iOS build, minor code refactor

aldenml:


.travis.yml
```diff
@@ -205,3 +218,9 @@ script:
       bjam arm-tests warnings-as-errors=on ssl=$ssl crypto=$crypto variant=test_arm $toolset target-os=linux link=static testing.launcher="sudo cp -R bin rootfs/; sudo chroot rootfs";
     fi'
   - cd ..
+
+# test build for iOS, using x86_64 to avoid the need to build boost from source for arm
```
arvidn: since the ``BOOST_ROOT`` environment variable is set, the Jamfile will automatically use that boost directory, and build dependencies from it. To force a build from the "system" boost, you have to unset it. But, does brew's boost package really contain arm binaries?

arvidn: oh, I iphonesim is not arm, so isn't this comment misleading? I'm not sure I understand what you mean.

src/ip_notifier.cpp
```diff
@@ -54,11 +54,29 @@ namespace libtorrent { namespace aux
 {
 namespace
 {
+template <typename...>
+struct ip_change_notifier_none final : ip_change_notifier
```
arvidn: why is this a variadic template?

.travis.yml
```diff
@@ -205,3 +218,9 @@ script:
...
+  - 'if [ "$iphone" == "1" ]; then
+      brew upgrade boost;
+      bjam -j3 variant=test_debug warnings-as-errors=on asserts=on toolset=darwin-x86_64~iphonesim link=static;
```
arvidn: this is just building. shouldn't you ``cd`` into tests, to run the tests as well?
Perhaps the binaries you get from iphonesim aren't executable directly from a macOS shell, is that the case?

src/ip_notifier.cpp
```diff
@@ -54,11 +54,29 @@ namespace libtorrent { namespace aux
...
+
+	void cancel() override {}
+
+private:
+	io_service& m_ios;
+};
+
 #if defined TORRENT_BUILD_SIMULATOR
-	// TODO: ip_change_notifier_sim
+using ip_change_notifier_sim = ip_change_notifier_none<>;
```
arvidn: it seems it would be simpler if all of these platform proprocessor blocks would produce the same name of the ip_change_notifier type, or at least provide an alias with the same name. That way you wouldn't need to repeat the if-chain below, to pick the right one to construct.

.travis.yml
aldenml: I don't see where `BOOST_ROOT` is set, are you talking about the `if "$arch" == "arm"`? I will rephrase the comment, but basically, I'm using a x86_64 arch with the iOS simulation toolchain, to avoid the need of "arm boost build" (they are not provided by brew).

src/ip_notifier.cpp
aldenml: a trick to avoid unused warnings, is like allow to take any "Tag", but I will change that.

.travis.yml
aldenml: you know, I didn't tried that, I was only testing that iOS build are not broken, but it the tests run, it would be nice, I till test.

src/ip_notifier.cpp
aldenml: Ok, I will do that

aldenml:
btw, the context of this change is have a more robust codebase for when I add the `ip_change_notifier` based on reachability (the only one that's suitable for iOS).

.travis.yml
arvidn: oh, right. yes, that's just for the arm build, that makes sense

src/ip_notifier.cpp
aldenml: doing the simplification here https://github.com/arvidn/libtorrent/pull/1899

aldenml:
I give up with the iOS build, maybe with more time in the future

PR #1902, opened 2017-04-11
added reachability based ip_change_notifier_impl for iOS

aldenml:
Possible to test in macOS adding `define=TORRENT_FORCE_IOS_IP_NOTIFIER` to the bjam command, to test with `session_log_alerts` tool.

This only detect "actual" reachability of ANY in the default routing table selected by the OS.

### DIFF
--- a/src/ip_notifier.cpp
+++ b/src/ip_notifier.cpp
@@ -98,7 +98,7 @@ void CFDispatchRetain(dispatch_queue_t q) { dispatch_retain(q); }
 void CFDispatchRelease(dispatch_queue_t q) { dispatch_release(q); }
 using CFDispatchRef = CFRef<dispatch_queue_t, CFDispatchRetain, CFDispatchRelease>;
 
-#if TARGET_OS_IPHONE || defined TORRENT_FORCE_IOS_IP_NOTIFIER
+#if TORRENT_USE_SC_NETWORK_REACHABILITY
 CFRef<SCNetworkReachabilityRef> create_reachability(SCNetworkReachabilityCallBack callback
 	, void* context_info)
 {
@@ -130,7 +130,11 @@ struct ip_change_notifier_impl final : ip_change_notifier
 			[](SCNetworkReachabilityRef /*target*/, SCNetworkReachabilityFlags /*flags*/, void *info)
 			{
 				auto obj = static_cast<ip_change_notifier_impl*>(info);
-				obj->m_ios.post([obj]() { if (obj->m_cb) obj->m_cb(error_code()); });
+				if (!obj->m_cb) return;
+
+				auto cb = std::move(obj->m_cb);
+				obj->m_cb = nullptr;
+				obj->m_ios.post([cb]() { cb(error_code()); });
 			}, this);
 
 		if (!m_queue || !m_reach
@@ -216,7 +220,11 @@ struct ip_change_notifier_impl final : ip_change_notifier
 			[](SCDynamicStoreRef /*store*/, CFArrayRef /*changedKeys*/, void *info)
 			{
 				auto obj = static_cast<ip_change_notifier_impl*>(info);
-				obj->m_ios.post([obj]() { if (obj->m_cb) obj->m_cb(error_code()); });
+				if (!obj->m_cb) return;
+
+				auto cb = std::move(obj->m_cb);
+				obj->m_cb = nullptr;
+				obj->m_ios.post([cb]() { cb(error_code()); });
 			}, this);
 
 		if (!m_queue || !m_store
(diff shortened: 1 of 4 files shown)


src/ip_notifier.cpp
```diff
@@ -99,6 +98,80 @@ void CFDispatchRetain(dispatch_queue_t q) { dispatch_retain(q); }
 void CFDispatchRelease(dispatch_queue_t q) { dispatch_release(q); }
 using CFDispatchRef = CFRef<dispatch_queue_t, CFDispatchRetain, CFDispatchRelease>;
 
+#if TARGET_OS_IPHONE || defined TORRENT_FORCE_IOS_IP_NOTIFIER
```
arvidn: the way this is normally done is by having a define that controls the use of this feature, and then have ``config.hpp`` map platforms to sets of features to enable. This way you can still enable a feature on the command line. i.e. I think having a: ``TORRENT_USE_SC_NETWORK_REACHABILITY`` that enables this feature would be nicer than the ``TORRENT_FORCE_IOS_IP_NOTIFIER``

aldenml: ok!

src/ip_notifier.cpp
```diff
@@ -99,6 +98,80 @@ void CFDispatchRetain(dispatch_queue_t q) { dispatch_retain(q); }
...
+			m_cb = std::move(cb);
+		else
+			m_ios.post([cb]()
+			{ cb(make_error_code(boost::system::errc::not_supported)); });
+	}
+
+	void cancel() override
+	{
+		if (m_reach)
+			SCNetworkReachabilitySetDispatchQueue(m_reach.get(), nullptr);
```
arvidn: if ``m_cb`` is set here, is it guaranteed to be called exactly once? Even if there is an event that has happend but hasn't triggered the callback yet? (the 1:1 correspondence is important)

aldenml: do you mind rephrasing the question?(I'm not sure of what are you asking)

arvidn: an important property of boost.asio I/O objects is that one invocation of ``async_*()`` results exactly in one invocation of the callback function, under all circumstances. There may be some tricky edge cases around cancelation. Consider the case where we cancel the ip_notifier, but we're too late, a change has already happened and there's a notification queued up to be delivered to us. If we would unconditionally call post a call to ``m_cb`` here, when we cancel, it may be invoked twice.

aldenml: I think I understand now. Let's see, this `ip_change_notifier_impl` is not strictly a boost.asio I/O object, but I presume you are willing to give him the same behavior. Does it work if right after calling m_cb (in `if (obj->m_cb) obj->m_cb(error_code());`, set `m_cb = nullptr`?

arvidn: I would expect so, and ignore any notification you get if m_cb is null. Maybe the notification function need to clear the callback too.
This property is important for managing lifetime management.

aldenml:
@arvidn I think I followed your comments as much as possible, let me know

src/ip_notifier.cpp
```diff
@@ -130,7 +130,11 @@ struct ip_change_notifier_impl final : ip_change_notifier
 			[](SCNetworkReachabilityRef /*target*/, SCNetworkReachabilityFlags /*flags*/, void *info)
 			{
 				auto obj = static_cast<ip_change_notifier_impl*>(info);
-				obj->m_ios.post([obj]() { if (obj->m_cb) obj->m_cb(error_code()); });
+				if (!obj->m_cb) return;
+
+				auto cb = std::move(obj->m_cb);
```
arvidn: I can't wait to bump C++ version to 14 to get ``std::exchange()`` :)

src/ip_notifier.cpp
```diff
@@ -130,7 +130,11 @@ struct ip_change_notifier_impl final : ip_change_notifier
 			[](SCNetworkReachabilityRef /*target*/, SCNetworkReachabilityFlags /*flags*/, void *info)
 			{
 				auto obj = static_cast<ip_change_notifier_impl*>(info);
-				obj->m_ios.post([obj]() { if (obj->m_cb) obj->m_cb(error_code()); });
+				if (!obj->m_cb) return;
```
arvidn: actually, looking at this I get a bit nervous about threads. the dispatch_queue you create, that's basically a thread pool, right? so, I suspect this callback will be called from some arbitrary thread. This is why you have to post it back onto the ``io_service`` message queue, to serialize.

The manipulation you do to ``obj`` below likely needs to happen *inside* the handler you post, right? Because this would likely cause a race condition (assuming my assumption about the threads is right)

arvidn: not just the clearing it, but also testing it. any access to ``obj`` probably should be done on the boost.asio ``io_service`` object, and hence in the main libtorrent thread.

arvidn: I think you should probably just pass ``obj`` straight through to the inner lambda, and then you can access it.

aldenml: I'm not an expert in dispatch_queue, but I'm sure it is another thread/pool.  I see your point with the multithread issue :) let me do that right now

arvidn:
lgtm

PR #1903, opened 2017-04-12
refactored netlink based ip_notifier in a separated class

aldenml:
This is essentially a copy/paste with the `std::move(cb)` and `const& cb` observations

### DIFF
--- a/src/ip_notifier.cpp
+++ b/src/ip_notifier.cpp
@@ -48,8 +48,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/aux_/disable_warnings_pop.hpp"
 #endif
 
-using namespace std::placeholders;
-
 namespace libtorrent { namespace aux
 {
 namespace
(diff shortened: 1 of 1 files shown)


src/ip_notifier.cpp
```diff
@@ -339,6 +379,7 @@ struct ip_change_notifier_impl final : ip_change_notifier
 		TORRENT_UNUSED(&ip_change_notifier_impl::on_notify);
 		TORRENT_UNUSED(cb);
 #elif TORRENT_USE_NETLINK
+		using namespace std::placeholders;
```
arvidn: all mentions of ``TORRENT_USE_NETLINK`` and associated snippets of code can be removed now, right?
this function and the constructor specifically

arvidn:
lgtm, but I think some old netlink stuff could be cleaned up

src/ip_notifier.cpp
aldenml: yes, but I want to do that at end, when I moved/refactored the windows impl, then delete the old `ip_change_notifier_impl`

aldenml:
Are you OK with the cleanup at the end, I need like two more PRs

arvidn:
sure

PR #1904, opened 2017-04-12
refactored windows based ip_notifier in a separated class

aldenml:


src/ip_notifier.cpp
```diff
@@ -311,13 +311,60 @@ struct ip_change_notifier_impl final : ip_change_notifier
...
+	// noncopyable
+	ip_change_notifier_impl(ip_change_notifier_impl const&) = delete;
+	ip_change_notifier_impl& operator=(ip_change_notifier_impl const&) = delete;
+
+	~ip_change_notifier_impl() override
+	{
+		cancel();
+		// the reason to call close() here is because the
+		// object_handle destructor doesn't close the handle
+		m_hnd.close();
```
arvidn: I'm a bit surprised the destructor doesn't close the handle. That means its life time isn't managed by anything. If any code are introduced in the constructor that may throw, the handle will leak. Currently the only thing that can throw in there is failing to create the event itself.

arvidn:
lgtm

PR #1905, opened 2017-04-12
provide a forward declaration header

arvidn:
document that forward declaring is discouraged


arvidn:
I merged this patch in from RC_1_1 instead

PR #1908, opened 2017-04-12
introduce lt namespace alias

arvidn:


arvidn:
I'm prepared to pull the trigger on this. @aldenml do you have any concerns about this?
It shouldn't (at least it isn't meant to) alter the API not ABI, just introduce an additional namespace in the API (that I will encourage going forward). Specifically, it is still compatible with forward declarations in the ``libtorrent`` namespace.

aldenml:
none, me 👍

Issue #1862, opened 2017-03-28
Clearing / configuring resolver cache


jberkel:
**Please provide the following information**

libtorrent version (or branch):

1.1.2

platform/architecture:

iOS / OSX

compiler and compiler version:

Apple LLVM version 8.0.0 (clang-800.0.42.1)

please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

I could not find a way to configure or clear the [resolver cache](https://github.com/arvidn/libtorrent/blob/1f9f588e75384df41beee5bfab58e143d8a38d54/src/resolver.cpp#L45-L45). The concrete use case are network configuration changes: I connect on a Nat64 enabled network, the resolver caches IPv6 addresses, I switch back to a IPv4 network. The next request will be made with the (now invalid) IPv6 address.

Therefore I'd like to disable the resolver cache completely or purge the contents on network config change. It does not seem to be possible to do this without changing the source code at the moment however.

aldenml:
do you really need `1.1.2`? I'm asking because I would like to do something at it, but I'm not using `RC_1_1` anymore.

jberkel:
@aldenml not necessarily. thanks for your offer to help! i also noticed your work on #1792 which is something I'm interested as well (my main platform is iOS).

aldenml:
interesting to hear from someone in iOS, well, that PR still requires serious work

jberkel:
Yes looks like it. I think a simple solution with `SCNetworkReachabilitySetCallback` would be preferable there, even if you don't get fine-grained network configuration details. Separate code paths for iOS / OSX would add substantial complexity.

arvidn:
would it make sense to automatically clear the cache when network changes are discovered?
I recognize that the cache is a bit weird to have at all, but my experience is that it's quite common to have lots of tracker host names that can stall to lookup, and the conventional dns APIs don't really support aborting with a custom (short) timeout. I believe the cache mitigates these problems.

jberkel:
@arvidn yes that would work. but being able to switch the cache off completely (setting max size = 0) would already help, I don't really need it since I run in a trusted environment with good OS caching support.

aldenml:
@jberkel a PR for configuration is merged https://github.com/arvidn/libtorrent/pull/1878, I decided to use the timeout instead the size.

There is an open question about what to do on network changes, but I think that's outside the scope of this issue.

jberkel:
@aldenml great, thanks, i'll give it a go.

[...]

PR #1911, opened 2017-04-12
added clear_resolver_cache_on_ip_change setting

aldenml:
for your consideration

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* added clear_resolver_cache_on_ip_change setting for internal host name resolver
 	* make the file_status interface explicitly public types
 	* added resolver_cache_timeout setting for internal host name resolver
 	* make parse_magnet_uri take a string_view instead of std::string
--- a/include/libtorrent/resolver.hpp
+++ b/include/libtorrent/resolver.hpp
@@ -59,6 +59,8 @@ struct TORRENT_EXTRA_EXPORT resolver final : resolver_interface
 
 	virtual void set_cache_timeout(seconds timeout) override;
 
+	virtual void clear() override;
+
 private:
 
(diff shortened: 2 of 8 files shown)


arvidn:
is the assumption behind this feature that when you change IP, you may implicitly have changed geolocation, and load balancing DNS servers may give you new (and presumably better) results?

Other than that, and for the most part, I wouldn't expect hostnames to resolve to different IPs depending on your IP.

aldenml:
This is a follow up of your comment https://github.com/arvidn/libtorrent/issues/1862#issuecomment-290741582, the main issue being IPv4 vs IPv6 name resolution (and I don't mind if you feel like it's not necessary).

arvidn:
ah, I see. Rereading that, do you know if we filter endpoints of the "wrong" family in the cache? I would have expected both IPv6 and IPv4 endpoints would be cached in the resolver cache.

aldenml:
There is no such filtering at the cache level, it stores all the addresses, and there are a fair amount of logic all over other places to deal with that. Now, is it possible that while connected via NAT64, you get _only_ IPv6 addresses? @jberkel?

arvidn:
hm, perhaps

jberkel:
@aldenml that is my understanding, you get ipv4 addresses embedded in ipv6, tagged with a well known-prefix:

https://tools.ietf.org/html/rfc6052#section-2.2

aldenml:
Thanks for the link, to be honest, I'm not a big fan of this setting, but it took me 10 mins. It's @arvidn call

arvidn:
yeah, I'm not a big fan of this patch either. it feels like it's addressing a much more specific case than it triggers on (i.e. perhaps it would make sense to do this when changing between all IPs being v4 to all IPs being v6 or vice versa). Still though, being able to disable the cache may be good enough.

PR #1915, opened 2017-04-15
fix memory leak in the disk cache

arvidn:
if a cached_piece_entry would stick around in a ghost list (ARC), it would keep the torrent object itself alive

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* fix memory leak in the disk cache
 	* fix double free in disk cache
 	* forward declaring libtorrent types is discouraged. a new fwd.hpp header is provided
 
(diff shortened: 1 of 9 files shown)


arvidn:
This was a subtle bug related to the ARC (advanced replacement cache) cache eviction algorithm. ARC fundamentally 2 kinds of lists, when an object is evicted from the list of items in the cache, a dummy placeholder is placed in a "ghost" list. If an item is brought back into the cache that is in the ghost list, the item is promoted to "frequently used".

in the disk cache, I used a ``cached_piece_entry`` object, with no buffers associated with it, as that dummy entry. The problem was that it still held an owning pointer to the storage, and hence the torrent object. So, even though a torrent was stopped and all it's disk jobs aborted and all its cached pieces evicted, there were (may) still be ghost pieces hanging around, preventing the torrent from being removed.

This fix is to have the option to evict pieces and force them to be removed, rather than demoted to a ghost entry.

arvidn:
I imagine this affects 1.0.x as well. Perhaps I should back-port this

arvidn:
heh, forgot to clean up and fix all the other tests..

PR #1917, opened 2017-04-15
add const

pavel-pimenov:


src/bt_peer_connection.cpp
```diff
@@ -2026,7 +2026,7 @@ namespace libtorrent {
 		// for the metadata extension.
 		TORRENT_ASSERT(m_sent_handshake);
 
-		char msg[] = {0,0,0,0};
+        const char msg[] = {0,0,0,0};
```
aldenml: indentation with spaces here, it should be `\t\t` (and 4 more below)

src/bt_peer_connection.cpp
```diff
@@ -366,7 +366,7 @@ namespace libtorrent {
 #ifndef TORRENT_DISABLE_LOGGING
 		peer_log(peer_log_alert::outgoing_message, "HAVE_ALL");
 #endif
-		char msg[] = {0,0,0,1, msg_have_all};
+		const char msg[] = {0,0,0,1, msg_have_all};
```
arvidn: for all the arrays used for messages, would you mind making them: ``static char const msg[]``?

pavel-pimenov: ok

src/bt_peer_connection.cpp
pavel-pimenov: fixed. sorry

pavel-pimenov:
@arvidn approve add private:
void send_buffer_internal(message_type const type, counters::stats_counter_t const counter);
?
default

arvidn:
this PR looks good to me. the proposed refactor for sending single-byte messages also seem reasonable. Perhaps a better name would be something along the lines of: ``send_simple_message()`` (it should still be a private member though).

If you're up for it, I also think ``send_message()`` should take a ``span<char const>`` instead of a pointer and length.

arvidn:
@pavel-pimenov do you intend to update this PR or should I merge this one?

pavel-pimenov:
@arvidn i create separate PR(after merge). ok?

pavel-pimenov:
PR for send_*_message https://github.com/arvidn/libtorrent/pull/1922

PR #1918, opened 2017-04-15
fix typo in natpmp::start

ssiloti:
Calling update_mapping() on a mapping with protocol set to none is a no-op so
I'm pretty sure this check was meant to skip over such mappings.


arvidn:
lgtm

PR #1919, opened 2017-04-15
remove calls to RAND_add

ssiloti:
This function has been removed from LibreSSL and offers dubious value.
The PRNGs in modern operating systems are robust enough that these little extra
bits aren't going to make a difference.


arvidn:
lgtm

PR #1922, opened 2017-04-17
add send_*_message

pavel-pimenov:


src/bt_peer_connection.cpp
```diff
@@ -358,6 +358,51 @@ namespace libtorrent {
 		stats_counters().inc_stats_counter(counters::num_outgoing_dht_port);
 	}
 
+	void bt_peer_connection::send_piece_message(message_type const type,
+			counters::stats_counter_t const counter,
```
arvidn: please just use one additional level of indentation, and when breaking a line, please put the "," at the beginning of the new line, for consistency

src/bt_peer_connection.cpp
```diff
@@ -366,24 +411,17 @@ namespace libtorrent {
...
 		peer_log(peer_log_alert::outgoing_message, "HAVE_NONE");
 #endif
-		static const char msg[] = {0,0,0,1, msg_have_none};
-		send_buffer(msg, sizeof(msg));
-
-		stats_counters().inc_stats_counter(counters::num_outgoing_have_none);
+		send_simple_message(msg_have_none, counters::num_outgoing_have_none);
 	}
 
 	void bt_peer_connection::write_reject_request(peer_request const& r)
```
arvidn: write_reject_request() could use the ``send_request_message()`` function too, right?

src/bt_peer_connection.cpp
```diff
@@ -358,6 +358,51 @@ namespace libtorrent {
...
+		TORRENT_ASSERT(m_sent_handshake);
+		TORRENT_ASSERT(m_sent_bitfield);
+
+		char msg[] = { 0,0,0,1,static_cast<char>(type) };
+		send_buffer(msg, sizeof(msg));
+
+		stats_counters().inc_stats_counter(counter);
+	}
+
+	void bt_peer_connection::send_request_message(message_type const type,
```
arvidn: it seems like you could merged all these 3 functions into a single template. They all has 0, 1 or 3 int32 arguments.

pavel-pimenov: template? different sizes: char msg[5], char msg[9], char msg[17]... and detail::write_int32

include/libtorrent/bt_peer_connection.hpp
```diff
@@ -273,6 +273,13 @@ namespace libtorrent {
 
 	private:
 		void write_dht_port();
+		void send_simple_message(message_type const type,
+			counters::stats_counter_t const counter);
+		void send_piece_message(message_type const type,
+			counters::stats_counter_t const counter, piece_index_t const index);
+		void send_request_message(message_type const type,
+			counters::stats_counter_t const counter, peer_request const& r,
+			int const flag);
```
arvidn: sorry, I missed this in the first pass. please remove all "top level" ``const`` here. (i.e. for the ``const&`` parameters you should leave them in).

arvidn: these don't have any meaning in the function declaration. they *do* have meaning in the definition in the .cpp file though

src/bt_peer_connection.cpp
```diff
@@ -366,24 +408,17 @@ namespace libtorrent {
 #ifndef TORRENT_DISABLE_LOGGING
```
arvidn: the assert that ``m_sent_handshake`` is set, 2 lines above, can be removed now. it's checked in ``send_simple_message()`` already

src/bt_peer_connection.cpp
```diff
@@ -2306,23 +2298,14 @@ namespace libtorrent {
 		TORRENT_ASSERT(m_sent_bitfield);
```
arvidn: The two asserts above can be removed I believe, right? (since they're done in send_simple_message())

src/bt_peer_connection.cpp
```diff
@@ -358,6 +358,48 @@ namespace libtorrent {
 		stats_counters().inc_stats_counter(counters::num_outgoing_dht_port);
 	}
 
+	void bt_peer_connection::send_piece_message(message_type const type
```
arvidn: as far as I can tell, all callers of this function also asserts or check: ``TORRENT_ASSERT(m_sent_handshake);`` and ``TORRENT_ASSERT(m_sent_bitfield);`` too. I believe you could factor those into this function.

src/bt_peer_connection.cpp
```diff
@@ -358,6 +358,48 @@ namespace libtorrent {
...
+		char msg[17] = { 0,0,0,13, static_cast<char>(type) };
+		char* ptr = msg + 5;
+
+		detail::write_int32(static_cast<int>(r.piece), ptr); // index
+		detail::write_int32(r.start, ptr); // begin
+		detail::write_int32(r.length, ptr); // length
+		send_buffer(msg, sizeof(msg), flag);
+
+		stats_counters().inc_stats_counter(counter);
+
```
arvidn: would you mind removing this blank line?

src/bt_peer_connection.cpp
arvidn: I can turn it into a template as a second phase. it could be a variadic template, assuming ``int32`` parameters.

src/bt_peer_connection.cpp
```diff
@@ -358,9 +358,12 @@ namespace libtorrent {
 		stats_counters().inc_stats_counter(counters::num_outgoing_dht_port);
 	}
 
-	void bt_peer_connection::send_piece_message(message_type const type
```
arvidn: these are the function definitions. the ``const`` *do* have meaning here. would you mind putting them back?

src/bt_peer_connection.cpp
arvidn: what about these two asserts?

src/bt_peer_connection.cpp
arvidn: and this assert?

arvidn: what about this assert?

arvidn:
did you leave the last assert in there deliberately?

arvidn:
lgtm

PR #1925, opened 2017-04-18
magnet links: unescape hash parameter

jberkel:
Unescape `xt` parameter.

### DIFF
--- a/src/magnet_uri.cpp
+++ b/src/magnet_uri.cpp
@@ -215,21 +215,23 @@ namespace libtorrent {
 			url = uri.substr(pos, find(uri, "&", pos) - pos);
 		}
 
-		std::string btih = unescape_string(url_has_argument(uri, "xt"), ec);
-		if (ec)
-		{
-			return;
-		}
-		else if (btih.empty())
+		string_view btih = url_has_argument(uri, "xt");
+		std::string unescaped_btih;
+		if (btih.empty())
 		{
 			ec = errors::missing_info_hash_in_uri;
 			return;
 		}
-
-		if (btih.substr(0, 9) != "urn:btih:")
+		else if (btih.substr(0, 9) != "urn:btih:")
 		{
-			ec = errors::missing_info_hash_in_uri;
-			return;
+			unescaped_btih = unescape_string(btih, ec);
+			if (ec) return;
+			else if (unescaped_btih.substr(0, 9) != "urn:btih:")
+			{
+				ec = errors::missing_info_hash_in_uri;
+				return;
+			}
+			else btih = unescaped_btih;
 		}
 
 		std::string::size_type peer_pos = std::string::npos;
--- a/src/magnet_uri.cpp
+++ b/src/magnet_uri.cpp
@@ -215,7 +215,7 @@ namespace libtorrent {
 			url = uri.substr(pos, find(uri, "&", pos) - pos);
 		}
 
-		string_view btih = unescape_string(url_has_argument(uri, "xt"), ec);
+		std::string btih = unescape_string(url_has_argument(uri, "xt"), ec);
 		if (ec)
 		{
 			return;
--- a/src/magnet_uri.cpp
+++ b/src/magnet_uri.cpp
@@ -215,8 +215,12 @@ namespace libtorrent {
 			url = uri.substr(pos, find(uri, "&", pos) - pos);
 		}
 
-		string_view btih = url_has_argument(uri, "xt");
-		if (btih.empty())
+		string_view btih = unescape_string(url_has_argument(uri, "xt"), ec);
+		if (ec)
+		{
+			return;
+		}
+		else if (btih.empty())
 		{
 			ec = errors::missing_info_hash_in_uri;
 			return;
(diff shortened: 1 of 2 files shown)


src/magnet_uri.cpp
```diff
@@ -215,8 +215,12 @@ namespace libtorrent {
 			url = uri.substr(pos, find(uri, "&", pos) - pos);
 		}
 
-		string_view btih = url_has_argument(uri, "xt");
-		if (btih.empty())
+		string_view btih = unescape_string(url_has_argument(uri, "xt"), ec);
```
arvidn: you can't assign a temporary string to a ``string_view``. But, we don't really need to decode the string here, as far as I can tell. We just need to match the prefix of "urn%3abtih%3a" (and I suppose the upper case versions too).

test/test_magnet.cpp
```diff
@@ -224,6 +224,23 @@ TORRENT_TEST(magnet)
 	p2 = s->abort();
 }
 
+TORRENT_TEST(parse_escaped_hash_parameter)
+{
+	error_code ec;
+	add_torrent_params p;
+	parse_magnet_uri("magnet:?xt=urn%3Abtih%3Acdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcd", p, ec);
```
arvidn: although correct, is there an actual use case for this? as far as I can tell, the intention behind the magnet link syntax is to not need to escape the colons.

src/magnet_uri.cpp
arvidn: I would expect the vast majority of magnet links to not escape this. is there a simple way to only pay the cost of allocating a new string when you need to unescape it?

arvidn: by the way, assigning a temporary to the string view is what's causing all the failures in CI

test/test_magnet.cpp
jberkel: I wrote a Python script which generates magnet links. Those links don't work with libtorrent right now because everything gets escaped.

```python
>>> from urllib import urlencode
>>> urlencode({'xt': 'urn:btih:1234'})
'xt=urn%3Abtih%3A1234'
```

After re-reading [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.3) it seems that `:` is a valid character in the query part and does not need to be escaped:

```
query       = *( pchar / "/" / "?" )
pchar       = unreserved / pct-encoded / sub-delims / ":" / "@"
unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
sub-delims  = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="
```

I think the implementation should still handle the case when it is escaped, though.

arvidn: I agree (that supporting escaped characters should be supported). What I'm hesitant about is to have everyone pay for this support, that's only needed very rarely (presumably). Can you think of a way to only copy the string in case there is any unescaping to be done?
The first thing that comes to my mind would be something like this, but maybe there's simpler/better way:

```
std::string info_hash_copy;
string_view btih = url_has_argument(uri, "xt");
// ... existing code
if (btih.find('%') != string_view::npos)
{
   info_hash_copy = unescape_string(btih);
   btih = info_hash_copy;
}
// ... more existing code
```

src/magnet_uri.cpp
```diff
@@ -215,21 +215,23 @@ namespace libtorrent {
...
 		}
-
-		if (btih.substr(0, 9) != "urn:btih:")
+		else if (btih.substr(0, 9) != "urn:btih:")
 		{
-			ec = errors::missing_info_hash_in_uri;
-			return;
+			unescaped_btih = unescape_string(btih, ec);
+			if (ec) return;
+			else if (unescaped_btih.substr(0, 9) != "urn:btih:")
```
arvidn: I would prefer to not have this ``else`` here (and below).

src/magnet_uri.cpp
```diff
@@ -215,21 +215,23 @@ namespace libtorrent {
...
+		string_view btih = url_has_argument(uri, "xt");
+		std::string unescaped_btih;
+		if (btih.empty())
 		{
 			ec = errors::missing_info_hash_in_uri;
 			return;
 		}
-
-		if (btih.substr(0, 9) != "urn:btih:")
+		else if (btih.substr(0, 9) != "urn:btih:")
```
arvidn: technically, it's possible for part of the hex-encoded info-hash to be escaped as well (redundantly, just like the ":"). So, although I appreciate the attempt to avoid extra work, I think scanning for a "%" may be unavoidable to support this properly

test/test_magnet.cpp
```diff
@@ -224,6 +224,23 @@ TORRENT_TEST(magnet)
...
+	TEST_CHECK(!ec);
+	TEST_EQUAL(aux::to_hex(p.info_hash), "cdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcd");
+}
+
+TORRENT_TEST(parse_invalid_escaped_hash_parameter)
+{
+	error_code ec;
+	add_torrent_params p;
+	parse_magnet_uri("magnet:?xt=urn%%3A", p, ec);
```
arvidn: in fact, perhaps there should be a test where the "urn:btih:" part is not escaped, but a character in the info-hash is

arvidn:
great. thanks!

PR #1929, opened 2017-04-19
merge send_message utility functions in bt_peer_connections into a template

arvidn:
@pavel-pimenov does this make sense?

### DIFF
--- a/include/libtorrent/bt_peer_connection.hpp
+++ b/include/libtorrent/bt_peer_connection.hpp
@@ -53,6 +53,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/piece_block_progress.hpp"
 #include "libtorrent/config.hpp"
 #include "libtorrent/pe_crypto.hpp"
+#include "libtorrent/io.hpp"
 
 namespace libtorrent {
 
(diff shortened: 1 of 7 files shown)


include/libtorrent/bt_peer_connection.hpp
```diff
@@ -272,14 +273,30 @@ namespace libtorrent {
...
+		void send_message(message_type const type
+			, counters::stats_counter_t const counter
+			, std::uint32_t flags
+			, Args... args)
+		{
+			TORRENT_ASSERT(m_sent_handshake);
+			TORRENT_ASSERT(m_sent_bitfield);
+
+			char msg[5 + sizeof...(Args) * 4]
+				= { 0,0,0,1 + sizeof...(Args) * 4, static_cast<char>(type) };
```
arvidn: the evaluation order is not defined here. this doesn't work

include/libtorrent/bt_peer_connection.hpp
```diff
@@ -272,14 +273,30 @@ namespace libtorrent {
...
+			char* ptr = msg + 5;
+			TORRENT_UNUSED(ptr);
+
+			int tmp[] = {0, (detail::write_int32(args, ptr), 0)...};
```
arvidn: the evaluation order is not defined here. this doesn't work

PR #1932, opened 2017-04-19
avoid connections to trackers when the event is stopped and stop_tracker_timeout <= 0

aldenml:
@arvidn the context of this change is that I was having "session destruction" issues and the problem was having these trackers communications during the "torrent abort". Now, this problem was not easy to replicate and it probably depends on the underlying networking conditions and OS. Having a small `stop_tracker_timeout` does not guarantee a small duration in the overall shutdown.

I think that this addition helps, since setting `stop_tracker_timeout = 0` is an explicit intention of having no slowness at all during the `stopped` event, the only way to guarantee that is not doing the communication.

I hope it really makes sense, because it fix my problem :)

### DIFF
--- a/include/libtorrent/announce_entry.hpp
+++ b/include/libtorrent/announce_entry.hpp
@@ -127,7 +127,7 @@ namespace libtorrent {
 		{
 			// the tracker was part of the .torrent file
 			source_torrent = 1,
-			// the tracker was added programatically via the add_troacker()_ function
+			// the tracker was added programatically via the add_tracker()_ function
 			source_client = 2,
 			// the tracker was part of a magnet link
 			source_magnet_link = 4,
--- a/include/libtorrent/announce_entry.hpp
+++ b/include/libtorrent/announce_entry.hpp
@@ -127,7 +127,7 @@ namespace libtorrent {
 		{
 			// the tracker was part of the .torrent file
 			source_torrent = 1,
-			// the tracker was added programatically via the add_tracker()_ function
+			// the tracker was added programatically via the add_tracker() function
 			source_client = 2,
 			// the tracker was part of a magnet link
 			source_magnet_link = 4,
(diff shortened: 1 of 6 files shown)


arvidn:
I'm not opposed to this feature, but I would ask for 2 things:

1. a test that make sure we don't contact the tracker when the stop_timeout is 0 (and maybe even make sure there's no name lookup)

2. documentation at this setting that this is what 0 means.

My assumption/impression is that the *cause* of this potentially taking a long time is the hostname lookup. Did you also get that impression? This was the main reason to introduce the dns cache, to perform the lookups at torrent *startup*, so it won't delay *shutdown*.

Perhaps the cache need to store lookup failures as well, to avoid re-issuing such lookups on shutdown.

aldenml:
I got the same impression but I couldn't confirm that in fact, that was the issue. I didn't have the time to refactor to allow for better log in the `http_connection`.

I will do 1) and 2)

include/libtorrent/announce_entry.hpp
```diff
@@ -127,7 +127,7 @@ namespace libtorrent {
 		{
 			// the tracker was part of the .torrent file
 			source_torrent = 1,
-			// the tracker was added programatically via the add_troacker()_ function
+			// the tracker was added programatically via the add_tracker()_ function
```
arvidn: hm.. I don't think that underscore should be there actually..

include/libtorrent/settings_pack.hpp
```diff
@@ -729,6 +729,8 @@ namespace libtorrent {
 			// ``stop_tracker_timeout`` is the number of seconds to wait when
 			// sending a stopped message before considering a tracker to have
 			// timed out. This is usually shorter, to make the client quit faster.
+			// If the value is set to 0, the connections to trackers with the
+			// stopped event are avoided.
```
arvidn: "avoided" sounds like there's some technique to make it unnecessary to send an ``event=stopped`` announce. What do you think about "suppressed"?

arvidn:
lgtm otherwise

arvidn:
on an unrelated note, I'm also considering adding an option to only perform the event=stopped announce if the hostname is in the cache. I think there's something like this right now, to force it to use the cache, but maybe it just makes more sense to never do any hostname lookups when shutting down

PR #1933, opened 2017-04-20
minor code cleanup

aldenml:


src/storage.cpp
```diff
@@ -301,7 +301,10 @@ namespace libtorrent {
 				m_part_file->export_file([&f, &ec](std::int64_t file_offset, span<char> buf)
 				{
 					iovec_t const v = {buf.data(), buf.size()};
-					std::int64_t const ret = f->writev(file_offset, v, ec.ec);
+#if TORRENT_USE_ASSERTS
```
arvidn: if you instead to ``TORRENT_UNUSED(ret);`` you only add 1 line, instead of 4.

src/peer_connection.cpp
```diff
@@ -5675,7 +5675,6 @@ namespace libtorrent {
 		}
 		if (size <= 0) return;
 
-		int i = 0;
```
arvidn: I thought I made this change already. is this from top of master?

src/storage.cpp
aldenml: yes, but for some reason I thought this was more "clear"

src/peer_connection.cpp
aldenml: yes, on top of master

aldenml:
I will do the changes, but I will wait for your `RC_1_1` merge is done, but you can put the changes in your PR if you want :)

arvidn:
yeah, I exposed a bug in the storage refcounting in master, and fixing that uncovered another issue with (what appears to be) releasing the storage too early. This has made the merge take a bit longer, but I've already been bit again by the fact that you cannot rebase a merge commit, so I would like to hold off on new commits on top of master until I can sort this out and get my merge in

arvidn:
I'm ready to merge this. did you intend to change that unused variable hunk?

aldenml:
yes, in a few minutes I will push an update

include/libtorrent/buffer.hpp
```diff
@@ -157,4 +157,4 @@ class buffer
 
 }
 
-#endif // BTORRENT_BUFFER_HPP_INCLUDED
```
arvidn: heh.. I wonder how this snuck in there..

PR #1935, opened 2017-04-21
Add keep_redundant_connections, a per-torrent override to close_redundant_connections.

AllSeeingEyeTolledEweSew:
Hi,

I added a "keep_redundant_connections" torrent setting, which is a per-torrent override to the session-wide "close_redundant_connections" setting.

Existing unit tests all ran successfully. I will fix any errors from Jenkins once it runs. I end-to-end tested this against my code and it seems to work, though I would appreciate guidance on how to write unit tests for this.

Motivation:

I've been developing a bittorrent-backed FUSE filesystem, which downloads pieces on demand as files are read.

This mechanism means the backing torrents flap between "downloading" and "seeding" state fairly often, under many different access patterns. So I had a significant performance problem due to the fact that libtorrent automatically disconnects seeds once all outstanding pieces are downloaded. It would often take several seconds to reconnect seeds, and some seeds would appear to start blocking new connections due to the flapping.

It helps to have a readahead buffer of outstanding pieces. This keeps the torrent in a downloading state more often, but it's no guarantee.

I tried disabling the "close_redundant_connections" session setting, but it's too broad. Once I touch a few torrents (particularly with large swarms) I quickly reach my maximum number of connections, with no way to selectively close connections for least-recently-used torrents. The "keep_redundant_connections" setting lets me create a policy like that.

My filesystem code is here: https://github.com/AllSeeingEyeTolledEweSew/yatfs ; requires deluge with this plugin https://github.com/AllSeeingEyeTolledEweSew/Deluge-PieceIO-Plugin . My code is still super experimental. TL;DR for testing I currently set keep_redundant_connections to true when a file is read, and set to false 30s after it's last closed. I have more complex LRU-based schemes in mind though.

### DIFF
--- a/bindings/python/src/torrent_handle.cpp
+++ b/bindings/python/src/torrent_handle.cpp
@@ -510,6 +510,8 @@ void bind_torrent_handle()
         .def("set_download_limit", _(&torrent_handle::set_download_limit))
         .def("download_limit", _(&torrent_handle::download_limit))
         .def("set_sequential_download", _(&torrent_handle::set_sequential_download))
+        .def("set_keep_redundant_connections", _(&torrent_handle::set_keep_redundant_connections))
+        .def("keep_redundant_connections", _(&torrent_handle::keep_redundant_connections))
 #ifndef TORRENT_NO_DEPRECATE
         .def("set_peer_upload_limit", &set_peer_upload_limit)
         .def("set_peer_download_limit", &set_peer_download_limit)
--- a/bindings/python/test.py
+++ b/bindings/python/test.py
@@ -59,6 +59,11 @@ class test_torrent_handle(unittest.TestCase):
 		self.h.prioritize_pieces([(0, 1)])
 		self.assertEqual(self.h.piece_priorities(), [1])
 
+	def test_redundant_connections(self):
+		self.assertFalse(self.h.keep_redundant_connections())
+		self.h.set_keep_redundant_connections(True)
+		self.assertTrue(self.h.keep_redundant_connections())
+
 	def test_replace_trackers(self):
 		self.setup()
 		trackers = []
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -417,6 +417,10 @@ namespace libtorrent
 		bool is_sequential_download() const
 		{ return m_sequential_download || m_auto_sequential; }
 
+		void set_keep_redundant_connections(bool keep);
+		bool keep_redundant_connections() const;
+		void maybe_close_redundant_connections();
+
 		void queue_up();
 		void queue_down();
 		void set_queue_position(int p);
@@ -1512,6 +1516,8 @@ namespace libtorrent
 		// there's mostly seeds in the swarm, download the files sequentially
 		// for improved disk I/O performance.
 		bool m_auto_sequential:1;
+
+		bool m_keep_redundant_connections:1;
 
 		// this means we haven't verified the file content
 		// of the files we're seeding. the m_verified bitfield
--- a/include/libtorrent/torrent_handle.hpp
+++ b/include/libtorrent/torrent_handle.hpp
@@ -1180,6 +1180,9 @@ namespace libtorrent
 		void connect_peer(tcp::endpoint const& adr, int source = 0
 			, int flags = 0x1 + 0x4 + 0x8) const;
 
+                void set_keep_redundant_connections(bool keep) const;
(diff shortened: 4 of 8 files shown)


arvidn:
The patch looks reasonable. Although it is missing a test (it would ideally be a simulation with two torrents, one configured to keep redundant connections and one not).

However, I'm not completely sold on this approach just yet. When I build something similar I found that padding requests with read-ahead (say, 1 MiB beyond the end of the request) would serve the purpose to let the requesting application come back to request more data without loosing any connections or "momentum" as well as actually providing read-ahead, which is very likely the next piece of data that will be requested anyway.

Another approach with using the global setting for redundant connections would be to dynamically adjust the max_connections per torrent. Granted, it won't be ideal as you would still keep a small number of redundant connections around for such torrents. But the client-side logic would be similar to your current plan to dynamically update this setting.

Lastly, something that I've been wanting to do for a while now is to replace all of these per-torrent settings with a single entry-point function to set and get *flags* for a torrent. These flags would be the same flags as are set in ``add_torrent_params::flags`` and would make it trivial to add more per-torrent features like these without even affecting the ABI. Adding another pair of functions like this increases the inertia of refactoring this in the future.

Would you, by any chance, be interested in folding this feature into a refactor of how to set *pause*, *auto_managed*, *seed_mode*, *upload_mode*, *share_mode*, *apply_ip_filter*, *stop_when_ready*, *apply_ip_filter* on a torrent, and deprecate the current individual functions?

(perhaps "folding" isn't the best word, I think the refactor and the new feature should be separate commits still).

AllSeeingEyeTolledEweSew:
(NB: I messed up my branch and had to reopen this PR as #1937)

Thanks for looking. I'll work on adding a test.

Interesting that you already wrote something similar to this! Did you ever post any code for it? I'd love to take a look.

I already use a readahead buffer, but I found it doesn't eliminate the problem. I prioritize "reading now" pieces to priority 7 and "readahead" pieces to 4, but it still often happens that the "reading now" piece is the last to finish downloading, so the torrent goes into the Seeding state before the next read request comes in, and disconnects all seeds. (I'm not sure if these priorities are ideal, it was just my first guess based on the text descriptions in the documentation). It seems to murder performance even if this happens to <1% of reads, since some seeds seem to aggressively block flapping connections.

I thought about some other approaches, like aggressively polling torrent status and ensuring I always have at least N pieces outstanding, or just setting all unused pieces to priority 1. However I really want to avoid wasting transfer. I realize my motivations probably seem really obscure, and I forget if I summarized my project before, so I wrote a little more below for motivation/clarity.

I really like the idea of tweaking ``max_connections``. That would allow graceful LRU "degredation" rather than eviction. I'll definitely incorporate that. However as you mention it doesn't really solve my problem. In particular I want to be able to keep thousands of torrents seeding (i.e., available for connections from downloaders) in the same libtorrent instance where I use filesystem-backing torrents -- since it makes sense to manage bandwidth all in one place. I can't do this with ``close_redundant_connections`` = ``false`` and low ``max_connections`` per torrent -- they would just each connect to a few other seeds and block downloaders.

The refactor sounds like a good idea. I'd love to work on it. Do you envision the API working like ``session_handle::apply_settings(settings_pack)``? That could still be included in the 1.1 train, correct? I'd love to have ``keep_redundant_connections`` available before too long.

---

About my project:

There have been a lot of projects for sequential / on-demand access to bittorrent files (popcorn time, etc). Many people consider them harmful to public swarms, for reasons summarized [here](https://wiki.vuze.com/w/Sequential_downloading_is_bad). I think those reasons are correct, but don't apply much to private trackers. On private trackers, seeds are rewarded for uploading and being online as much as possible, so they don't quit after uploading some ratio; and bad behavior is punished (hit-and-runs, etc). So there should be no inherent harm in sequential downloading on private trackers. Also, private trackers have copious user-supplied metadata about their content, so it's a good fit for a system to browse all content on a tracker, and access it on-demand. This would also open the door to seeding content based on swarm health needs, rather than user needs (I find I'll often delete content A to make room for content B that I might use ... then never use it).

So I want to build at least one system like the above, and my hope above all is to lift the taboo on sequential downloading and generate productive discussion about how we could do more interesting things with the bittorrent ecosystem. Maybe the benefits of private trackers show that a reputation system would be useful, though it's a compromise to anonymity. Maybe the more conservative torrent client developers can be convinced to add piece-level prioritization.

arvidn:
for future reference, you can just force push an update to the branch the PR points to, and github will pick it up.

I'll post my response in the other PR.

PR #1937, opened 2017-04-22
Add torrent_handle::keep_redundant_connections and implementation.

AllSeeingEyeTolledEweSew:
Hi,

I added a "keep_redundant_connections" torrent setting, which is a per-torrent override to the session-wide "close_redundant_connections" setting.

Existing unit tests all ran successfully. I will fix any errors from Jenkins once it runs. I end-to-end tested this against my code and it seems to work, though I would appreciate guidance on how to write unit tests for this.

Motivation:

I've been developing a bittorrent-backed FUSE filesystem, which downloads pieces on demand as files are read.

This mechanism means the backing torrents flap between "downloading" and "seeding" state fairly often, under many different access patterns. So I had a significant performance problem due to the fact that libtorrent automatically disconnects seeds once all outstanding pieces are downloaded. It would often take several seconds to reconnect seeds, and some seeds would appear to start blocking new connections due to the flapping.

It helps to have a readahead buffer of outstanding pieces. This keeps the torrent in a downloading state more often, but it's no guarantee.

I tried disabling the "close_redundant_connections" session setting, but it's too broad. Once I touch a few torrents (particularly with large swarms) I quickly reach my maximum number of connections, with no way to selectively close connections for least-recently-used torrents. The "keep_redundant_connections" setting lets me create a policy like that.

My filesystem code is here: https://github.com/AllSeeingEyeTolledEweSew/yatfs ; requires deluge with this plugin https://github.com/AllSeeingEyeTolledEweSew/Deluge-PieceIO-Plugin . My code is still super experimental. TL;DR for testing I currently set keep_redundant_connections to true when a file is read, and set to false 30s after it's last closed. I have more complex LRU-based schemes in mind though.

### DIFF
--- a/src/bt_peer_connection.cpp
+++ b/src/bt_peer_connection.cpp
@@ -1906,7 +1906,7 @@ namespace libtorrent
 		// if we're finished and this peer is uploading only
 		// disconnect it
 		if (t->is_finished() && upload_only()
-			&& m_settings.get_bool(settings_pack::close_redundant_connections)
+			&& !t->keep_redundant_connections()
 			&& !t->share_mode())
 			disconnect(errors::upload_upload_connection, op_bittorrent);
 
@@ -2006,7 +2006,7 @@ namespace libtorrent
 		// if we send upload-only, the other end is very likely to disconnect
 		// us, at least if it's a seed. If we don't want to close redundant
 		// connections, don't sent upload-only
-		if (!m_settings.get_bool(settings_pack::close_redundant_connections)) return;
+		if (t->keep_redundant_connections()) return;
 
 #ifndef TORRENT_DISABLE_LOGGING
 		peer_log(peer_log_alert::outgoing_message, "UPLOAD_ONLY", "%d"
--- a/src/torrent_handle.cpp
+++ b/src/torrent_handle.cpp
@@ -413,6 +413,11 @@ namespace libtorrent
 		TORRENT_ASYNC_CALL1(set_sequential_download, sd);
 	}
 
+	void torrent_handle::set_keep_redundant_connections(bool keep) const
+	{
+		TORRENT_ASYNC_CALL1(set_keep_redundant_connections, keep);
+	}
+
 	void torrent_handle::piece_availability(std::vector<int>& avail) const
 	{
 		TORRENT_SYNC_CALL1(piece_availability, boost::ref(avail));
@@ -510,6 +515,12 @@ namespace libtorrent
 	bool torrent_handle::is_sequential_download() const
 	{
 		TORRENT_SYNC_CALL_RET(bool, false, is_sequential_download);
+		return r;
+	}
+
+	bool torrent_handle::keep_redundant_connections() const
+	{
+		TORRENT_SYNC_CALL_RET(bool, false, keep_redundant_connections);
 		return r;
 	}
 
--- a/src/torrent_handle.cpp
+++ b/src/torrent_handle.cpp
@@ -418,6 +418,12 @@ namespace libtorrent
 		TORRENT_ASYNC_CALL1(set_keep_redundant_connections, keep);
 	}
 
+	bool torrent_handle::keep_redundant_connections() const
+	{
+		TORRENT_SYNC_CALL_RET(bool, false, keep_redundant_connections);
+		return r;
+	}
+
 	void torrent_handle::piece_availability(std::vector<int>& avail) const
 	{
 		TORRENT_SYNC_CALL1(piece_availability, boost::ref(avail));
@@ -515,12 +521,6 @@ namespace libtorrent
 	bool torrent_handle::is_sequential_download() const
 	{
 		TORRENT_SYNC_CALL_RET(bool, false, is_sequential_download);
-		return r;
-	}
-
-	bool torrent_handle::keep_redundant_connections() const
-	{
-		TORRENT_SYNC_CALL_RET(bool, false, keep_redundant_connections);
 		return r;
 	}
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -132,6 +132,8 @@ script:
     bjam --hash -j3 warnings-as-errors=on debug-iterators=on picker-debugging=on invariant-checks=full variant=$variant $coverage_toolset stage_module libtorrent-link=shared install-type=LIB dll-path=../../lib;
     LD_LIBRARY_PATH=../../lib DYLD_LIBRARY_PATH=../../lib python test.py;
     fi'
+  - objdump -T libtorrent.so | grep keep_redundant_connections
+  - objdump -T ../../lib/libtorrent.so.1.1.4 | grep keep_redundant_connections
   - cd ../..
 
   - cd simulation
--- a/.travis.yml
+++ b/.travis.yml
@@ -132,8 +132,6 @@ script:
     bjam --hash -j3 warnings-as-errors=on debug-iterators=on picker-debugging=on invariant-checks=full variant=$variant $coverage_toolset stage_module libtorrent-link=shared install-type=LIB dll-path=../../lib;
     LD_LIBRARY_PATH=../../lib DYLD_LIBRARY_PATH=../../lib python test.py;
     fi'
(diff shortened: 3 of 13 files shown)


AllSeeingEyeTolledEweSew:
I don't understand why these CI checks are failing. I'd appreciate any guidance.

src/torrent_handle.cpp
```diff
@@ -513,6 +518,12 @@ namespace libtorrent
 		return r;
 	}
 
+	bool torrent_handle::keep_redundant_connections() const
```
arvidn: you added this function to a section of deprecated functions. so this won't be included when building with ``TORRENT_NO_DEPRECATE``. In fact, I think this state should be reflected in the ``torrent_status::flags`` instead, where these other states were moved.

arvidn:
I just realized this patch is against RC_1_1. My previous comment about refactoring this doesn't apply to this branch.

However, There are some changes I would want to see:

1. Instead of exposing the state via the ``torrent_handle::close_redundant_connections()`` function, it should be a ``bool`` in ``torrent_status``.

2. There should be a flag in ``add_torrent_params`` to enable this when a torrent is added (without any race conditions).

3. There should be a test making sure it works as expected. Keep in mind that the main value of a test for this is to make sure nobody accidentally breaks the feature in the future. I would suggest you take a look at ``simulation/test_auto_manage.cpp``. You could probably just add a test case to that file. In the simulation, you can get the torrent handles by calling ``session::get_torrents()``, and  the test could be that connections are still open after a download completes. (make sure both clients enable redundant connections).

arvidn:
> I prioritize "reading now" pieces to priority 7 and "readahead" pieces to 4, but it still often happens that the "reading now" piece is the last to finish downloading

You may want to take a look at ``set_piece_deadline()`` instead of using piece priorities. See: http://libtorrent.org/streaming.html

> I thought about some other approaches, like aggressively polling torrent status and ensuring I always have at least N pieces outstanding, or just setting all unused pieces to priority 1. However I really want to avoid wasting transfer.

Apart from wasting transfer (which I suppose could be debated, unused bandwidth capacity just expires and is wasted regardless) the major cost of having all pieces enabled "in the background" so to speak, is that they clog up the request queues at the peers, increasing the delay to request time critical pieces. One way to mitigate that would be to set the request queue very small. Time critical pieces has an effecting request queue length twice as long as the configured one anyway. Perhaps that ratio may not be enough though. (i.e. you could set ``max_out_request_queue`` to 1). If you end up experimenting with this, you may want to try increasing the factor, [here](https://github.com/arvidn/libtorrent/blob/master/src/peer_connection.cpp#L3366).

> The refactor sounds like a good idea. I'd love to work on it. Do you envision the API working like session_handle::apply_settings(settings_pack)? That could still be included in the 1.1 train, correct? I'd love to have keep_redundant_connections available before too long.

At first I was thinking that this refactor should be done in master, but now that I think about it, I can't see any ABI breakage with it. I haven't really formed a strong opinion either way. The benefits of doing it in RC_1_1 is that we don't need to introduce a new function call just to deprecated it immediately in the next major release.

arvidn:
> Maybe the benefits of private trackers show that a reputation system would be useful, though it's a compromise to anonymity. Maybe the more conservative torrent client developers can be convinced to add piece-level prioritization.

You may be interested in @ssiloti 's [reputation extension](https://github.com/ssiloti/libtorrent-reputation):

AllSeeingEyeTolledEweSew:
> You may want to take a look at set_piece_deadline() instead of using piece priorities

Ah, I hadn't grokked how the deadline piece picker worked until I read that. It seems like it would suffice to set deadline to 0 for pieces being read, and 1 for readahead pieces. I'll experiment.

Would you also recommend disabling `strict_end_game_mode`?

> There should be a flag in add_torrent_params to enable this when a torrent is added (without any race conditions).

Sounds like this would fit with the refactor.

My current picture of the refactor is to add:
```
void torrent_handle::set_flags(uint64_t flags);
uint64_t torrent_handle::get_flags();
```
where the layout of flags is the same as `add_torrent_params`.

Should there also be `set_flag(add_torrent_params::flag_t, bool)` and `get_flag(add_torrent_params::flag_t)`, or will the above be sufficient?

> You may be interested in @ssiloti 's reputation extension:

That looks really interesting. Thanks!

arvidn:
> Would you also recommend disabling strict_end_game_mode?

I would recommend experimenting with that at least. It will make it more likely to waste bandwidth, but for time critical data, it will likely make it less likely to get stuck on a slow peer. On the other hand, if your download link is the bottleneck, it will make it worse.

> My current picture of the refactor is to add:
> 
> void torrent_handle::set_flags(uint64_t flags);
> uint64_t torrent_handle::get_flags();
> 
> where the layout of flags is the same as add_torrent_params.
> 
> Should there also be set_flag(add_torrent_params::flag_t, bool) and  get_flag(add_torrent_params::flag_t), or will the above be sufficient?

Hm. I hadn't thought that much about it. As for querying the flags, I imagine adding a flags field to ``torrent_status`` would be better. But that would affect the ABI, so that won't really work for RC_1_1.

For setting them though, it seems inconvenient to have to first query the flags (synchronously) first just to set or clear a single flag.

How about a setter like this?

```
set_flags(std::uint64_t mask, uint64_t flags = 0xffffffffffffffff);
```

Where only flags specified in ``mask`` are affected.

AllSeeingEyeTolledEweSew:
I'm still working on this, but I'm having a lot of trouble reproducing the link error found in the CI runs.

AllSeeingEyeTolledEweSew:
Actually -- re-reading your comments from earlier, you advise adding `keep_redundant_connections` to `torrent_status`. But wouldn't this break the ABI, therefore be inappropriate for RC_1_1?

If so, should I do this change against master, after #2119 is in?

arvidn:
yeah, the new field can not go into ``torrent_status``. Perhaps that specific piece could be deferred until master.

arvidn:
I think this patch looks reasonable for RC_1_1. As you point out though, the ``torrent_status`` change should probably be dropped, and introduced in master.

Would you mind writing a test for it as well? The simplest would probably be to extend simulations/test_transfer.cpp or test_swarm.cpp. Set the priority of all pieces to 0, wait one second, make sure your seed is still connected.

Note that there's a risk the seed will still disconnect you though. And this may be a deeper issue with this approach. My understanding is that what you really want is to pretend that we're still interested in peers, we're just not requesting anything. With this patch, _you_ won't disconnect redundant connections, but your peers may still disconnect you if they consider you redundant. For instance, there's no point for a seed to keep a connection to a peer that isn't interested in it.

AllSeeingEyeTolledEweSew:
I'll write a simulation test. Thanks for looking!

Once this is done, is the preferred way for me to make a second branch, rebase it onto master, and do a second PR?

True, nothing stops a peer from disconnecting from us. In experimentation though, I see most peers stay connected for at least a few minutes, and some appear to stay connected forever. In fact if all peer connections expired fairly, then I could try to rely on `settings_pack::close_redundant_connections` and not do this patch.

You make an interesting point though. I didn't check to see when libtorrent sends not-interested. Perhaps I should call this something more generic like `keep_interested`, and ensure we stay interested to peers, in addition to not disconnecting them.

arvidn:
I regularly merge all of RC_1_1 into master, so no need for a second PR against master.

arvidn:
when you update this, feel free to squash and rebase

src/bt_peer_connection.cpp
```diff
@@ -1906,7 +1906,7 @@ namespace libtorrent
 		// if we're finished and this peer is uploading only
 		// disconnect it
 		if (t->is_finished() && upload_only()
-			&& m_settings.get_bool(settings_pack::close_redundant_connections)
```
arvidn: don't you need to check both of these? or do you update the torrent settings if the global change?

arvidn:
Now I'm starting to want to cut a release from master. I think it might be best to put this in master at this point, before I cut. Note that the torrent flags have been refactored in master to actually be flags in a flag-field, rather than separate fields. It has unified the way they are set in ``add_torrent_params`` and on a live ``torrent_handle``.

PR #1938, opened 2017-04-22
Multi-home support part 1/3 - DHT

ssiloti:


ssiloti:
I question the accuracy of the Codecov report. For example, it claims `peer_connection::snub_peer` is no longer being covered, but I checked locally and it is definitely still being called in multiple simulation tests.

.travis.yml
```diff
@@ -146,21 +146,21 @@ script:
 # libtorrent is the name of the test suite target
   - cd test
   - 'if [ "$tests" == "1" ]; then
-      travis_retry bjam -j3 warnings-as-errors=on ssl=$ssl crypto=$crypto debug-iterators=on picker-debugging=on invariant-checks=full $toolset variant=$variant libtorrent test_natpmp enum_if -l300 &&
+      travis_retry bjam -j3 c++-template-depth=512 warnings-as-errors=on ssl=$ssl crypto=$crypto debug-iterators=on picker-debugging=on invariant-checks=full $toolset variant=$variant libtorrent test_natpmp enum_if -l300 &&
```
arvidn: I think this setting belongs in the ``user-config.jam`` file. would you mind putting it there instead? or is there a good reason to keep it on the command line?

aldenml: Is this a necessary condition for the library itself? if that's the case, I think it belongs to `Jamfile`. And what's the piece of code that triggers this?

ssiloti: That's a good point actually. I didn't think to add it to the Jamfile because I don't experience this problem locally, but I believe that's because I use gcc 6. There's surely still lots of people using gcc 5 so this setting needs to be changed in the default build.

The offending statement is on line 130 in dht_tracker.cpp. The second `forward_as_tuple` call apparently triggers a rather deep nesting of template instantiations.

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -1054,15 +1065,28 @@ namespace aux {
 				, error_code& ec
 				, int flags);
 
+			void send_udp_packet_hostname_listen(aux::session_listen_socket* sock
```
arvidn: the hostname overload is only used for http and socks5 proxies, right? I would have expected that case to not depend on a source NIC. Or is it possible to have multiple connections to the same proxy, via different NICs?

include/libtorrent/aux_/session_listen_socket.hpp
```diff
@@ -0,0 +1,60 @@
...
+	{
+		virtual address get_external_address() = 0;
+		virtual tcp::endpoint get_local_endpoint() = 0;
+
+		session_listen_socket() = default;
+		session_listen_socket(session_listen_socket const&) = default;
+		session_listen_socket& operator=(session_listen_socket const&) = default;
+
+	protected:
+		~session_listen_socket() {}
```
arvidn: would it work to say ``= default;`` instead of ``{}``? I don't believe there's any meaningful difference, but I think it may communicate intent a bit better.

include/libtorrent/aux_/session_impl.hpp
ssiloti: I'm honestly not sure what is and isn't possible when it comes to proxy configurations. I just did this for constancy with the regular udp send function and because unless `m_force_proxy` is set we can still end up sending directly.

include/libtorrent/aux_/session_udp_sockets.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include <boost/asio/io_service.hpp>
+#include <vector>
+
+namespace libtorrent { namespace aux {
+
+	struct session_udp_socket
+	{
+		explicit session_udp_socket(io_service& ios)
+			: sock(ios) {}
+		virtual ~session_udp_socket() {}
```
arvidn: does anyone derive from this type? I don't see any. If not, does this need to be virtual?

include/libtorrent/kademlia/dht_tracker.hpp
```diff
@@ -69,14 +71,20 @@ namespace libtorrent { namespace dht {
 			, counters& cnt
 			, dht_storage_interface& storage
 			, dht_state state);
-		virtual ~dht_tracker();
+
+		dht_tracker(dht_tracker const&) = delete;
+
+		virtual ~dht_tracker() = default;
```
arvidn: I think you can just remove this destructor declaration. This class is final, so nobody can derive from it, and if ``socket_manager`` has a virtual destructor, this destructor will be virtual automatically anyway.

include/libtorrent/kademlia/node.hpp
```diff
@@ -75,22 +76,25 @@ class announce_observer : public observer
...
 
-struct udp_socket_interface
+struct socket_manager
 {
 	virtual bool has_quota() = 0;
-	virtual bool send_packet(entry& e, udp::endpoint const& addr) = 0;
+	virtual bool send_packet(aux::session_listen_socket* s, entry& e, udp::endpoint const& addr) = 0;
 protected:
-	~udp_socket_interface() {}
+	~socket_manager() {}
```
arvidn: ``= default;`` here too perhaps?

arvidn: I don't actually have a strong opinion here, if you think ``{}`` is better, that's fine (but you changed a few other ``{}`` -> ``= default`` I noticed)

include/libtorrent/kademlia/node.hpp
```diff
@@ -75,22 +76,25 @@ class announce_observer : public observer
...
 };
 
+// get the closest node to the id with the given family_name
+using get_foreign_node_t = std::function<node*(node_id const&, std::string const&)>;
```
arvidn: can this return ``nullptr``? If not, should it return a reference?

simulation/setup_dht.cpp
```diff
@@ -257,7 +273,8 @@ struct dht_node final : lt::dht::udp_socket_interface
 private:
 	asio::io_service m_io_service;
 	std::shared_ptr<dht::dht_storage_interface> m_dht_storage;
-	std::map<std::string, lt::dht::node*> m_nodes;
+	bool m_add_dead_nodes;
+	bool m_ipv6;
```
arvidn: I believe this member can be ``const``.

src/kademlia/dht_state.cpp
```diff
@@ -38,12 +38,40 @@ POSSIBILITY OF SUCH DAMAGE.
 namespace libtorrent { namespace dht {
 namespace {
 
-	node_id extract_node_id(bdecode_node const& e, string_view key)
+	node_ids_t extract_node_ids(bdecode_node const& e, string_view key)
```
arvidn: would you mind adding a unit test for this function?

src/kademlia/dht_tracker.cpp
```diff
@@ -91,95 +91,104 @@ namespace libtorrent { namespace dht {
...
+		if (n != m_nodes.end())
+			n->second.dht.update_node_id();
 		update_storage_node_ids();
+	}
+
+	void dht_tracker::new_socket(aux::session_listen_socket* s)
+	{
+		address local_address = s->get_local_endpoint().address();
+#if TORRENT_USE_IPV6
+		// don't try to start dht nodes on non-global addresses
```
arvidn: shouldn't there be a check like this for IPv4 addresses as well?
do you think you could use ``is_local()`` from broadcast_socket.hpp? (that's supposed to work for both IPv4 and IPv6)

src/kademlia/dht_tracker.cpp
```diff
@@ -91,95 +91,104 @@ namespace libtorrent { namespace dht {
...
+		if (local_address.is_v6()
+			&& (local_address.to_v6().is_link_local() || local_address.to_v6().is_site_local()))
+			return;
+#endif
+		auto stored_nid = std::find_if(m_state.nids.begin(), m_state.nids.end()
+			, [&](node_ids_t::value_type const& nid) { return nid.first == local_address; });
+		node_id nid = stored_nid != m_state.nids.end() ? stored_nid->second : node_id();
```
arvidn: I believe ``nid`` could be ``const``

aldenml:
I have a question, previously there was an explicit bias towards IPv4, in the sense that all calls were made first to the IPv4 DHT, but now, it's just an iteration on the map. I not sure if that was the way for historical reasons or not. Could this impact the performance of the IPv4 DHT network, if peers start to have 3 or 4 IPv6 addresses?

include/libtorrent/aux_/session_udp_sockets.hpp
ssiloti: That's coming in the UTP part of this patch series which adds the outgoing sockets. I'll make it non-virtual for now and change it in that PR.

src/kademlia/dht_tracker.cpp
```diff
@@ -284,30 +280,37 @@ namespace libtorrent { namespace dht {
...
-		ids.push_back(m_dht.nid());
-#if TORRENT_USE_IPV6
-		ids.push_back(m_dht6.nid());
-#endif
+		for (auto& n : m_nodes)
+			ids.push_back(n.second.dht.nid());
 		m_storage.update_node_ids(ids);
 	}
 
+	node* dht_tracker::get_node(node_id const& id, std::string const& family_name)
```
arvidn: is this used to select which interface to used for a specific lookup? I suppose that makes sense, at least once for each address family.

include/libtorrent/kademlia/node.hpp
ssiloti: Yes it can return a `nullptr` if there is no node with the requested address family.

src/kademlia/dht_tracker.cpp
```diff
@@ -284,30 +280,37 @@ namespace libtorrent { namespace dht {
...
 
 	void dht_tracker::announce(sha1_hash const& ih, int listen_port, int flags
 		, std::function<void(std::vector<tcp::endpoint> const&)> f)
 	{
-		m_dht.announce(ih, listen_port, flags, f);
-#if TORRENT_USE_IPV6
-		m_dht6.announce(ih, listen_port, flags, f);
-#endif
+		for (auto& n : m_nodes)
+			n.second.dht.announce(ih, listen_port, flags, f);
```
arvidn: I see, announcing from each interface makes sense

src/kademlia/dht_tracker.cpp
ssiloti: The problem with IPv4 is that there might be a NAT so we can't trust that the local address reflects the interface's true scope.

src/kademlia/dht_tracker.cpp
ssiloti: `dht_tracker::get_node` is used for the case where a node needs to reach into another node's routing table to get nodes to satisfy a `want` parameter.

src/kademlia/dht_tracker.cpp
```diff
@@ -538,29 +530,35 @@ namespace libtorrent { namespace dht {
...
-		// TODO: figure out a better solution when multi-home is implemented
-		node const* dht = m_dht.nid() == nid ? &m_dht
-#if TORRENT_USE_IPV6
-			: m_dht6.nid() == nid ? &m_dht6 : nullptr;
-#else
-			: nullptr;
-#endif
-
-		if (dht == nullptr) return ret;
+		for (auto& n : m_nodes)
```
arvidn: this looks like a ``std::find_if()`` followed by a call to ``for_each_node()``. would that be a more intuitive structure?

src/kademlia/dht_tracker.cpp
```diff
@@ -639,7 +635,30 @@ namespace libtorrent { namespace dht {
 		m_send_quota -= int(m_send_buf.size());
 
 		error_code ec;
-		m_send_fun(addr, m_send_buf, ec, 0);
+		if (s->get_local_endpoint().protocol().family() != addr.protocol().family())
+		{
+			// the node is trying to send a packet to a different address family
+			// than its socket, this can happen during bootstrap
+			// pick a node with the right address family and use its socket
+			bool found = false;
```
arvidn: could this be simplified with a ``std::find_if()``?

src/kademlia/node.cpp
```diff
@@ -101,22 +102,23 @@ void incoming_error(entry& e, char const* msg, int error_code = 203)
 
 } // anonymous namespace
 
-node::node(udp proto, udp_socket_interface* sock
+node::node(aux::session_listen_socket* sock, socket_manager* sock_man
```
arvidn: you had a consistency argument for using a pointer here instead of a reference, right?
(I think eventually these should be turned into references)

ssiloti: In the case of the DHT the problem is we use the pointer as a key into a `std::map` in `dht_tracker` so we can't store it purely as a reference. Although I guess we could use a reference wherever possible and add ref/deref operators as necessary.

src/kademlia/dht_tracker.cpp
ssiloti: It ends up being rather terse because I have to spell out the type of m_nodes. I could typedef it but it doesn't seem worth the effort. We can revisit this once we move the codebase to C++14 and can use `auto` parameters in lambdas.

src/kademlia/node.cpp
```diff
@@ -1164,11 +1166,11 @@ void node::write_nodes_entries(sha1_hash const& info_hash
 		bdecode_node wanted = want.list_at(i);
 		if (wanted.type() != bdecode_node::string_t)
 			continue;
-		auto wanted_node = m_nodes.find(wanted.string_value().to_string());
-		if (wanted_node == m_nodes.end()) continue;
+		node* wanted_node = m_get_foreign_node(info_hash, wanted.string_value().to_string());
```
arvidn: I see, it can return ``nullptr``, never mind then.

src/session_handle.cpp
```diff
@@ -760,12 +760,14 @@ namespace {
 
 	unsigned short session_handle::listen_port() const
 	{
-		return sync_call_ret<unsigned short>(&session_impl::listen_port);
+		return sync_call_ret<unsigned short, unsigned short(session_impl::*)() const>
```
arvidn: do you envision deprecating this API once we switch over to everything being multi-homed?

src/session_impl.cpp
```diff
@@ -1763,6 +1763,55 @@ namespace {
 		reopen_listen_sockets();
 	}
 
+	void session_impl::interface_to_endpoints(std::string const& device, int const port
+		, bool const ssl, std::vector<listen_endpoint_t>& eps)
```
arvidn: do callers of this function rely on ``eps`` being appended to, and not replaced? if not, it seems it would be better to return the vector.
Really, unless you expect this to have a material performance impact, that interface probably would make sense regardless (and callers can make it obvious that they append the results)

src/session_impl.cpp
```diff
@@ -1763,6 +1763,55 @@ namespace {
...
+	{
+		// First, check to see if it's an IP address
+		error_code err;
+		address const adr = address::from_string(device.c_str(), err);
+		if (!err)
+		{
+#if !TORRENT_USE_IPV6
+			if (adr.is_v4())
+#endif
+				eps.emplace_back(adr, port, std::string(), ssl);
```
arvidn: what if the device string is an IPv6 address?

src/session_handle.cpp
ssiloti: We can certainly deprecate this. The only potential issue is that it would leave the `listen_succeeded_alert` as the only way for the client to find out what port(s) it is listening on.

src/session_impl.cpp
ssiloti: Yes this function is expected to append to the vector.

src/session_impl.cpp
```diff
@@ -1813,55 +1862,18 @@ namespace {
...
 		}
 
 		auto remove_iter = partition_listen_sockets(eps, m_listen_sockets);
 
 		while (remove_iter != m_listen_sockets.end())
 		{
-			// TODO notify interested parties of this socket's demise
+#ifndef TORRENT_DISABLE_DHT
+			if (m_dht)
+				m_dht->delete_socket(&*remove_iter);
```
arvidn: if only this callback took a reference :)

src/session_impl.cpp
ssiloti: Then we ignore it if IPv6 is disabled.

src/session_impl.cpp
```diff
@@ -2092,29 +2108,40 @@ namespace {
...
+		, span<char const> p
+		, error_code& ec
+		, int const flags)
+	{
+		auto s = static_cast<listen_socket_t*>(sock)->udp_sock;
+
+		s->sock.send_hostname(hostname, port, p, ec, flags);
+
+		if ((ec == error::would_block
+				|| ec == error::try_again)
```
arvidn: you probably don't need to break this line

src/session_impl.cpp
```diff
@@ -2092,29 +2108,40 @@ namespace {
...
+	void session_impl::send_udp_packet_hostname_listen(aux::session_listen_socket* sock
+		, char const* hostname
+		, int const port
+		, span<char const> p
+		, error_code& ec
+		, int const flags)
+	{
+		auto s = static_cast<listen_socket_t*>(sock)->udp_sock;
+
+		s->sock.send_hostname(hostname, port, p, ec, flags);
```
arvidn: I guess each udp socket sets up its own connection to the socks5 proxy, so the source address will match each individual socks5 connection. I suppose it might make sense to talk to the proxy over all interfaces in case we have different bandwidth on different links, but I'm not confident each of these will be able to bind to the same port on the proxy and actually have the interface be transparent. not a big deal for now though.

src/session_impl.cpp
```diff
@@ -6545,24 +6591,44 @@ namespace {
...
+			set_external_address(i, ip, source_type, source);
 			break;
 		}
+	}
+
+	void session_impl::set_external_address(
+		tcp::endpoint const& local_endpoint, address const& ip
+		, int const source_type, address const& source)
+	{
+		for (auto& sock : m_listen_sockets)
```
arvidn: unless this is placeholder logic, do you think it would be simpler to use ``std::find_if()`` followed by the call to ``set_external_address()``?

arvidn:
lgtm overall. some comments

src/kademlia/dht_tracker.cpp
arvidn: I see, that makes sense. (would you mind adding a comment to that effect?)

include/libtorrent/aux_/session_udp_sockets.hpp
arvidn: ok, as I was going over the patch I realized that some things may be preparations for the coming patches. makes sense

include/libtorrent/aux_/session_impl.hpp
arvidn: my suspicion is that each udp socket will contact the socks5 proxy from its interface and try to set up a UDP associate tunnel, and presumably only the first one will succeed, as it tries to bind to a specific port. so, whatever the other ones do, probably doesn't matter. Maybe at some point someone will want to improve this, but I think this is fine for now.

.travis.yml
arvidn: I'm fine with putting it in the Jamfile, in the default build as you suggest. Although, it is a gcc specific option, I suppose appveyor will tell us if it breaks msvc to have it there

arvidn:
code coverage is only collected from the unit tests (i.e. ``tests`` directory), no from the simulations (iirc). The idea is that code coverage is not as useful of a metric for simulations, as lots of code can be executed without any test actually verifying that it's doing the right thing (this is true for a lot of the non-unit tests under ``tests`` as well though)

src/kademlia/dht_state.cpp
ssiloti: It already gets good coverage via the test for `read_dht_state()`.

src/session_handle.cpp
arvidn: my assumption here is that the 2nd template argument is necessary to disambiguate an overload, and that the other overload (where you don't specify the interface) is meant to be deprecated. If so, perhaps it would make sense to deprecate the public API that doesn't specify an interface as well. I can't say I've thought it through any more than that.

src/session_impl.cpp
arvidn: oh, right

src/session_handle.cpp
ssiloti: This function is calling the overload that doesn't specify an interface. The public API right now doesn't expose any sort of interface handle so there's no way to have a public function which specifies one.

src/kademlia/dht_tracker.cpp
```diff
@@ -117,14 +117,16 @@ namespace libtorrent { namespace dht {
 	{
 		address local_address = s->get_local_endpoint().address();
 #if TORRENT_USE_IPV6
-		// don't try to start dht nodes on non-global addresses
+		// don't try to start dht nodes on non-global IPv6 addresses
+		// with IPv4 the interface might be behind NAT so we can't skip them based on the scope of the local address
+		// and we might not have the external address yet
```
arvidn: thinking some more about this; it won't be a problem for 127.0.0.0/8, will it? I would imagine we don't want to run a DHT over loopback

src/session_handle.cpp
ssiloti: I guess you could use the local endpoint we expose in `announce_endpoint` but that would be subject to race conditions.

src/kademlia/dht_tracker.cpp
ssiloti: We won't normally create a listen socket on loopback. If someone does that explicitly it's probably because they're testing something in which case we probably should run the DHT in case that's what they're testing.

arvidn: makes sense. I just seem to recall some default behavior to enumerate all interfaces and create sockets on them, but maybe that's gone now.

ssiloti: That logic will be going into the last patch in the series, but it still won't listen on loopback by default. Now that I think about it that may confuse people trying to test transfers locally because `localhost` often resolves to ::1 these days. I may have to modify the final patch to disallow DHT and tracker announces on loopback.

arvidn:
lgtm. @ssiloti do you want to squash any of these commits before I merge?

include/libtorrent/aux_/session_listen_socket.hpp
```diff
@@ -0,0 +1,60 @@
...
+	// abstract interface for a listen socket owned by session_impl
+	// pointers to this type serve as a handle for the listen socket
+	// use a separate abstract type to prohibit outside access to private fields of listen_socket_t
+	// and because some users of these handles should not be coupled to session_impl
+	struct TORRENT_EXTRA_EXPORT session_listen_socket
+	{
+		virtual address get_external_address() = 0;
+		virtual tcp::endpoint get_local_endpoint() = 0;
+
+		session_listen_socket() = default;
```
aldenml: why do you need to provide these `default`s? I don't recall them in any other "interface"

include/libtorrent/kademlia/dht_tracker.hpp
```diff
@@ -69,14 +71,18 @@ namespace libtorrent { namespace dht {
 			, counters& cnt
 			, dht_storage_interface& storage
 			, dht_state state);
-		virtual ~dht_tracker();
+
+		dht_tracker(dht_tracker const&) = delete;
```
aldenml: Is this a "noncopyable" thing? if that's the case, don't you need `operator=(dht_tracker const&) = delete` for consistency?

src/kademlia/dht_state.cpp
```diff
@@ -36,16 +36,44 @@ POSSIBILITY OF SUCH DAMAGE.
...
 	{
-		if (e.type() != bdecode_node::dict_t) return node_id();
-		auto nid = e.dict_find_string_value(key);
-		if (nid.size() != 20) return node_id();
-		return node_id(nid);
+		if (e.type() != bdecode_node::dict_t) return node_ids_t();
+		node_ids_t ret;
+		// first look for an old-style nid
```
aldenml: how is that possible that in this DHT's state, there is an old-style nid? shouldn't that be an error?

src/kademlia/dht_state.cpp
```diff
@@ -78,29 +106,28 @@ namespace {
...
-#if TORRENT_USE_IPV6
-		ret.nid6 = extract_node_id(e, "node-id6");
-#endif
+		ret.nids = extract_node_ids(e, "node-id");
 
 		if (bdecode_node const nodes = e.dict_find_list("nodes"))
 			ret.nodes = detail::read_endpoint_list<udp::endpoint>(nodes);
-#if TORRENT_USE_IPV6
-		if (bdecode_node const nodes = e.dict_find_list("nodes6"))
-			ret.nodes6 = detail::read_endpoint_list<udp::endpoint>(nodes);
```
aldenml: is this right? I expect `nodes6` property to not be used and removed, but I still see that that's not the case and it's still used in `save_dht_state`

src/kademlia/dht_tracker.cpp
```diff
@@ -91,95 +91,106 @@ namespace libtorrent { namespace dht {
...
+
+	void dht_tracker::new_socket(aux::session_listen_socket* s)
+	{
+		address local_address = s->get_local_endpoint().address();
+#if TORRENT_USE_IPV6
+		// don't try to start dht nodes on non-global IPv6 addresses
+		// with IPv4 the interface might be behind NAT so we can't skip them based on the scope of the local address
+		// and we might not have the external address yet
+		if (local_address.is_v6()
+			&& (local_address.to_v6().is_link_local() || local_address.to_v6().is_site_local()))
```
aldenml: If `is_site_local()` is another signal for "is local", does makes sense to add that `||` to `bool is_local(address const& a)` and use that function instead?

Jamfile
```diff
@@ -766,6 +766,7 @@ lib torrent
 
 	: # default build
 	<threading>multi
+	<c++-template-depth>512
```
aldenml: if only `gcc-5` is giving problems, why put this for all compilers?

aldenml:
@arvidn @ssiloti unless you are eager to merge this in a short time, I appreciate if I can have your response to my comments, specially about the DHT state serialization.

I have an observation design wise. I see the need for `new_socket` to be called by `session_impl`, that's obvious, but the need for `delete_socket` looks dubious to me, because that means that a client of `dht_tracker` (in a potential out of `libtorrent` scenario) needs to be aware of that and properly pair those calls. There should be a way to automatically handle the lifetime of the internal `dht_node`s inside.

src/kademlia/dht_state.cpp
ssiloti: The state might have been saved by an older version of libtorrent.

src/kademlia/dht_state.cpp
ssiloti: Good catch. I waffled a bit on whether to consolidate these into a single nodes list or not. ultimately I left them separate but missed reverting this change.

Jamfile
ssiloti: Other versions only pass because they just barely manage to stay below the depth limit. In the future something else may trip the limit everywhere so we might as well raise it now.

ssiloti:
@aldenml 
I suppose you could hang a thunk somewhere to call `dht_tracker::delete_socket` as part of the `listen_socket_t` destructor, but that seems like an unnecessary complication at this point.

include/libtorrent/aux_/session_listen_socket.hpp
ssiloti: Because the destructor is declared protected the default constructor is not automatically generated.

src/kademlia/dht_tracker.cpp
ssiloti: Site local addresses are deprecated so it's probably not necessary to check for them.

ssiloti:
@arvidn I'll squash this down to a clean patch series now

src/kademlia/dht_state.cpp
aldenml: I doubt it, since this was introduced in `master`, not that important anyway

arvidn:
@ssiloti https://ci.appveyor.com/project/arvidn/libtorrent/build/6117/job/6mrtbmw6oucn0p7f#L1177

include/libtorrent/kademlia/dht_tracker.hpp
```diff
@@ -128,18 +132,33 @@ namespace libtorrent { namespace dht {
...
+				, libtorrent::dht_settings const& settings
+				, node_id const& nid
+				, dht_observer* observer, counters& cnt
+				, get_foreign_node_t get_foreign_node
+				, dht_storage_interface& storage);
+
+			node dht;
+			deadline_timer connection_timer;
+		};
+		using tracker_nodes_t = std::map<aux::session_listen_socket*, tracker_node>;
```
arvidn: it looks like msvc is trying to generate a copy constructor for this map, failing because ``tracker_node`` is not copyable.

include/libtorrent/kademlia/dht_tracker.hpp
ssiloti: Now I remember why I did this. It's a workaround for a bug in msvc 14.0 which is fixed in msvc 14.1. I've re-added the line with an appropriate `#if` condition and comment.

test/test_dht.cpp
```diff
@@ -3480,13 +3512,13 @@ TORRENT_TEST(dht_state)
 {
 	dht_state s;
 
-	s.nid = to_hash("0000000000000000000000000000000000000001");
+	s.nids.emplace_back(address::from_string("1.1.1.1"), to_hash("0000000000000000000000000000000000000001"));
 	s.nodes.push_back(uep("1.1.1.1", 1));
 	s.nodes.push_back(uep("2.2.2.2", 2));
 	// not important that IPv6 is disabled here
-	s.nid6 = to_hash("0000000000000000000000000000000000000002");
-	s.nodes6.push_back(uep("3.3.3.3", 3));
```
aldenml: if the deletions of tests related to `.nodes6` correct?

test/test_dht.cpp
```diff
@@ -3480,13 +3512,13 @@ TORRENT_TEST(dht_state)
...
-	s.nodes6.push_back(uep("4.4.4.4", 4));
+	s.nids.emplace_back(address::from_string("1::1"), to_hash("0000000000000000000000000000000000000002"));
+	s.nodes.push_back(uep("1::1", 3));
```
aldenml: this does not looks right, if `.nodes` is intended to store IPv4

test/test_dht.cpp
ssiloti: I had to remove those because they no longer work if the host system doesn't have IPv6 support. This can only be reliably tested in the simulator.

include/libtorrent/kademlia/dht_observer.hpp
```diff
@@ -64,14 +68,13 @@ namespace libtorrent { namespace dht {
 #endif
 
 	protected:
-		~dht_logger() {}
+		~dht_logger() = default;
```
aldenml: just a question, going forward, what is the preferred style here?

simulation/test_dht.cpp
```diff
@@ -65,13 +65,7 @@ void bootstrap_session(std::vector<dht_network*> networks, lt::session& ses)
...
 		auto router_nodes = dht->router_nodes();
 
-		char const* nodes_key;
-		if (router_nodes.front().address().is_v6())
-			nodes_key = "nodes6";
-		else
-			nodes_key = "nodes";
-
-		lt::entry::list_type& nodes = state["dht state"][nodes_key].list();
+		lt::entry::list_type& nodes = state["dht state"]["nodes"].list();
```
aldenml: Is it ok to just use "nodes" here?

src/kademlia/dht_tracker.cpp
```diff
@@ -91,95 +92,105 @@ namespace libtorrent { namespace dht {
...
 
-	void dht_tracker::update_node_id()
+	void dht_tracker::delete_socket(aux::session_listen_socket* s)
 	{
-		m_dht.update_node_id();
-#if TORRENT_USE_IPV6
-		m_dht6.update_node_id();
-#endif
-		update_storage_node_ids();
+		m_nodes.erase(s);
```
aldenml: I think it's a very good idea to put `TORRENT_ASSERT(m_nodes.contains(s))` here

aldenml:
@ssiloti I see no real protection in `dht_tracker::new_socket(aux::session_listen_socket* s)` to avoid IPv6 if not `TORRENT_USE_IPV6`. I expect some protection coming from `session_impl.cpp` but I think it's a good idea to handle the case, and return some error. That would help even if `TORRENT_USE_IPV6` and the IPv6 is local, instead of the silent `return;`

ssiloti:
@aldenml 

If an IPv6 listen socket gets created when IPv6 is disabled that is a bug in session_impl and it should be handled there.

aldenml:
I remember, back before IPv6 DHT support, some similar assumption that resulted in a runtime bug if exceptions were disabled, easily avoided, and being a release binary, was not obvious where exactly was the error

arvidn:
is this good to be merged?

ssiloti:
Yes

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -117,8 +118,14 @@ namespace dht {
 		class item;
 	}
 
-	struct listen_socket_t
+	struct listen_socket_t final : aux::session_listen_socket
```
arvidn: I just realized this class is defined directly in the ``libtorrent`` namespace, not under ``aux``. It doesn't look like this changed anytime recently. I will attempt to move it into ``aux`` where I think it belongs

PR #1939, opened 2017-04-22
fix MSVC 14.1 warnings-as-errors=on build

ajax16384:
fix type narrowing warns
use span random_bytes

### DIFF
--- a/src/bt_peer_connection.cpp
+++ b/src/bt_peer_connection.cpp
@@ -2084,7 +2084,7 @@ namespace libtorrent {
 
 		if (t->is_seed())
 		{
-			std::fill_n(ptr, packet_size - 5, static_cast<std::uint8_t>(0xff));
+			std::fill_n(ptr, packet_size - 5, std::uint8_t{0xff});
 
 			// Clear trailing bits
 			msg.back() = (0xff << ((8 - (num_pieces & 7)) & 7)) & 0xff;
--- a/src/bt_peer_connection.cpp
+++ b/src/bt_peer_connection.cpp
@@ -2716,7 +2716,7 @@ namespace libtorrent {
 					disconnect(errors::no_memory, op_encryption);
 					return;
 				}
-				std::fill(m_sync_vc.get(), m_sync_vc.get() + 8, static_cast<char>(0));
+				std::fill(m_sync_vc.get(), m_sync_vc.get() + 8, char{0});
 				rc4_decrypt({m_sync_vc.get(), 8});
 			}
 
--- a/src/bt_peer_connection.cpp
+++ b/src/bt_peer_connection.cpp
@@ -2084,7 +2084,7 @@ namespace libtorrent {
 
 		if (t->is_seed())
 		{
-			std::fill_n(ptr, packet_size - 5, 0xff);
+			std::fill_n(ptr, packet_size - 5, static_cast<std::uint8_t>(0xff));
 
 			// Clear trailing bits
 			msg.back() = (0xff << ((8 - (num_pieces & 7)) & 7)) & 0xff;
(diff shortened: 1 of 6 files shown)


src/receive_buffer.cpp
```diff
@@ -125,7 +125,7 @@ void receive_buffer::cut(int const size, int const packet_size, int const offset
 		m_recv_end -= size;
 
 #if TORRENT_USE_ASSERTS
-		std::fill(m_recv_buffer.begin() + m_recv_end, m_recv_buffer.end(), 0xcc);
+		std::fill(m_recv_buffer.begin() + m_recv_end, m_recv_buffer.end(), static_cast<unsigned char>(0xcc));
```
arvidn: as far as I can tell, ``m_recv_buffer.begin()`` returns a ``char*``, not ``unsigned char``. Am I missing something?

src/bt_peer_connection.cpp
```diff
@@ -2084,7 +2084,7 @@ namespace libtorrent {
 
 		if (t->is_seed())
 		{
-			std::fill_n(ptr, packet_size - 5, 0xff);
+			std::fill_n(ptr, packet_size - 5, static_cast<std::uint8_t>(0xff));
```
arvidn: if this works: ``std::uint8_t{0xff}``, I think I would prefer that syntax

src/bt_peer_connection.cpp
```diff
@@ -2716,7 +2716,7 @@ namespace libtorrent {
 					disconnect(errors::no_memory, op_encryption);
 					return;
 				}
-				std::fill(m_sync_vc.get(), m_sync_vc.get() + 8, 0);
+				std::fill(m_sync_vc.get(), m_sync_vc.get() + 8, static_cast<char>(0));
```
arvidn: here as well, if ``char{0}`` works

src/i2p_stream.cpp
```diff
@@ -130,7 +131,7 @@ namespace libtorrent {
 		m_state = sam_connecting;
 
 		char tmp[20];
-		std::generate(tmp, tmp + sizeof(tmp), &std::rand);
+		aux::random_bytes(tmp);
```
arvidn: +1

arvidn:
lgtm (apart from comments).
also, if this actually eliminates all the warnings in the tests, please add the flag to the appveyor build command line as well!

src/torrent.cpp
```diff
@@ -4865,14 +4865,11 @@ namespace libtorrent {
...
-
-		// initialize pad files to priority 0
-		for (file_index_t i(0); i < limit; ++i)
+		auto si = files.begin();
+		for (file_index_t i(0); i < limit; ++i, ++si)
 		{
-			if (!fs.pad_file_at(i)) continue;
-			m_file_priority[i] = 0;
+			// initialize pad files to priority 0
+			m_file_priority[i] = fs.pad_file_at(i) ? 0 : aux::numeric_cast<std::uint8_t>(*si);
```
aldenml: I suggest something like `aux::clamp(*si, 0, 7) & 0xff` here, what do you think?

arvidn: is ``& 0xff`` there to silence some warning? I don't see how ``clamp(..., 0, 7)`` could return anything outside of the ``0xff`` mask

arvidn: but yes, ``clamp()`` would have higher accuracy in fixing incorrect priority values. The distinction, though, that ``numeric_cast`` is only an assert and not a runtime check. Should the input vector be considered trusted? I'm not sure.

src/receive_buffer.cpp
ajax16384:
````c++
vc\tools\msvc\14.10.25017\include\xutility(2690): warning C4244: '=': conversion from 'const int' to 'char', possible loss of data
````

src/bt_peer_connection.cpp
ajax16384: ok

ajax16384:
appveyor already has "warnings-as-errors=on"

arvidn:
how come the appveyor build passes without this patch, if warnings are treated as errors and this patch fixes warnings?

arvidn:
lgtm btw. I will merge steven's multi-home patch ahead of this though

ajax16384:
patch fixes 14.1 warns (vs2017)

src/torrent.cpp
aldenml: yes, `& 0xff` is a warning thing

Issue #1940, opened 2017-04-22
next_announce retry interval is not respected


jberkel:
libtorrent version (or branch): 1.1.3
platform/architecture: iOS

please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

Not 100% sure if this is a bug or just works as expected, so wanted to clarify:
I'm implementing failure handling at the moment and adjusted the tracker timeout / backoff settings.

1. When the tracker announce request is sent, `min_announce` is initialised to now+10 seconds:
[torrent.cpp#L3312-L3312](https://github.com/arvidn/libtorrent/blob/f95f5b05c268b31c691e31b46b3f86e55f88f7cc/src/torrent.cpp#L3312-L3312)

2. Now, when a failure occurs `next_announce` is calculated to happen in now+7 seconds. In my case the failure happens immediately (connection refused): [announce_entry.cpp#L107-L107](https://github.com/arvidn/libtorrent/blob/1f9f588e75384df41beee5bfab58e143d8a38d54/src/announce_entry.cpp#L107-L107)

However, the `min_announce` value set previously prevents the request from getting sent out after 7 seconds, it is only sent after 10 seconds have passed. Since `tracker_retry_delay_min` is set to 5 seconds ([announce_entry.cpp#L43-L43](https://github.com/arvidn/libtorrent/blob/1f9f588e75384df41beee5bfab58e143d8a38d54/src/announce_entry.cpp#L43-L43)) I would expect `min_announce` in step 1 to be set to now+(tracker_retry_delay_min) = 5 seconds, not 10.

arvidn:
yes, I think you're right. I can't recall whether setting those timestamps to now+20s and now+10s is relevant or necessary. I would say it looks suspicious to have those literals there in general. Do you know what would happen if they're just set to now? Presumably there won't be any re-announces while it's being announced to anyway, and immediately when ``updating`` is set to false, those timestamps should have been set to something reasonable anyway.

Would you mind submitting a PR to fix this (against the RC_1_1 branch)? I think ideally those timestamps would just be set to ``now``. Presumably there wouldn't be a busy loop re-announcing because of that.

jberkel:
Sure, I'll do some testing with the timestamps set to `now` and submit a PR.

jberkel:
Closed w/ #1943

PR #1943, opened 2017-04-23
Set next/min_announce to now

jberkel:
Prevents unnecessary tracker announce delays #1940


jberkel:
@arvidn works as expected now

PR #1944, opened 2017-04-24
add support for utp over multiple interfaces

ssiloti:


include/libtorrent/aux_/session_udp_sockets.hpp
```diff
@@ -56,6 +59,42 @@ namespace libtorrent { namespace aux {
 		bool write_blocked = false;
 	};
 
+	struct outgoing_udp_socket final : session_udp_socket
+	{
+		explicit outgoing_udp_socket(io_service& ios, std::string const& dev, bool ssl_)
```
arvidn: I don't think you need "explicit" here, since the constructor takes more than one argument

include/libtorrent/aux_/session_udp_sockets.hpp
```diff
@@ -56,6 +59,42 @@ namespace libtorrent { namespace aux {
...
+			: session_udp_socket(ios), device(dev), ssl(ssl_) {}
+
+		// the name of the device the socket is bound to, may be empty
+		// if the socket is not bound to a device
+		std::string device;
+
+		// set to true if this is an SSL socket
+		bool ssl = false;
```
arvidn: are these two members meant to change during the lifetime of the socket, or can the be ``const``?

include/libtorrent/aux_/session_udp_sockets.hpp
```diff
@@ -56,6 +59,42 @@ namespace libtorrent { namespace aux {
...
+		partition_outgoing_sockets(std::vector<listen_endpoint_t>& eps);
+
+		tcp::endpoint bind(socket_type& s, address const& remote_address) const;
+
+		void update_proxy(proxy_settings const& settings);
+
+		// close all sockets
+		void close();
+
+		std::vector<std::shared_ptr<outgoing_udp_socket>> sockets;
```
arvidn: I would have expected these members to be private. are they accessed directly?

include/libtorrent/utp_socket_manager.hpp
```diff
@@ -49,9 +49,18 @@ namespace libtorrent {
...
 
+	// interface/handle to the underlying udp socket
+	struct TORRENT_EXTRA_EXPORT utp_socket_interface
+	{
+		virtual udp::endpoint local_endpoint() = 0;
+	protected:
+		virtual ~utp_socket_interface() = default;
+	};
+
 	struct utp_socket_manager final
```
arvidn: did this class used to implement an interface? I wonder why it's ``final``

src/session_udp_sockets.cpp
```diff
@@ -0,0 +1,117 @@
...
+				return ep.device == sock->device
+					&& ep.addr == sock->sock.local_endpoint().address();
+			});
+
+			if (match != eps.end())
+			{
+				// remove the matched endpoint so that another socket can't match it
+				// this also signals to the caller that it doesn't need to create a
+				// socket for the endpoint
+				eps.erase(match);
```
arvidn: this looks a bit odd. I expect the general case of this comparison function to not provider strict ordering. Perhaps it's not a problem in practice. Bu if there are two sockets that match the same endpoint, the result here is non-deterministic, right?

arvidn: presumably all sockets have unique listen endpoints. I take it this function is used for finding the sockets that need closing and opening, right? to "sync" with the endpoint list.

src/session_impl.cpp
```diff
@@ -1973,6 +1984,136 @@ namespace {
...
+			eps.emplace_back(address_v6(), 0, "", true);
+#endif
+#endif
+		}
+
+		auto remove_iter = m_outgoing_sockets.partition_outgoing_sockets(eps);
+
+		for (auto i = remove_iter; i != m_outgoing_sockets.sockets.end(); ++i)
+		{
+			auto& remove_sock = *i;
```
arvidn: this takes a reference, then you remove it from the utp_socket_manager, doesn't that invalidate the reference?

arvidn: oh, right, the actual object is owned by ``m_outgoing_sockets``

src/session_impl.cpp
```diff
@@ -1973,6 +1984,136 @@ namespace {
...
+				if (ec)
+				{
+#ifndef TORRENT_DISABLE_LOGGING
+					if (should_log())
+					{
+						session_log("bind to device failed (device: %s): %s"
+							, ep.device.c_str(), ec.message().c_str());
+					}
+#endif // TORRENT_DISABLE_LOGGING
+					continue;
```
arvidn: do you think it would make sense to post a ``udp_error_alert`` or a ``listen_failed_alert`` here, and the error case below?

src/session_udp_sockets.cpp
ssiloti: Yes, this function is part of synchronizing the set of outgoing sockets with the set of outgoing interfaces.

The predicate in `std::partition` is not an ordering function so there's no requirement to produce a strict weak ordering. In the case of multiple possible matches then yes the results are non-deterministic, but any of the possible results should still satisfy the specified interface configuration. That comment was originally written for listen sockets though, multiple matches shouldn't happen with outgoing sockets so I'll update the comment to remove that bit.

include/libtorrent/aux_/session_udp_sockets.hpp
ssiloti: `sockets` is accessed directly by `session_impl::reopen_outgoing_sockets`, index can be private.

arvidn:
lgtm

arvidn:
I notice this review is concerned with *outgoing* sockets. for UDP/uTP, listen sockets and outgoing sockets are the same thing, right? How come there's a separate list of outgoing UDP sockets, rather than being part of the listen socket list?

ssiloti:
The outgoing sockets are handled differently from listen sockets in several ways which I think justifies keeping them separate:

They are configured with the `outoing_interfaces` setting rather than `listen_interfaces`.
The `listen_socket_t` struct contains many members which are not applicable to outgoing sockets, such as an `ip_voter` and `tcp::acteptor`.
Outgoing sockets should not have DHT nodes running on them or tracker announces sent over them.

Basically outgoing sockets are _only_ for outgoing UTP connections. This mirrors how `outgoing_interfaces` works with TCP connections.

src/session_impl.cpp
```diff
@@ -1973,6 +1984,143 @@ namespace {
...
+#ifdef TORRENT_USE_OPENSSL
+			interface_to_endpoints(iface, 0, true, eps);
+#endif
+		}
+
+		// if no outgoing interfaces are specified, create sockets to use
+		// any interface
+		if (eps.empty())
+		{
+			eps.emplace_back(address_v4(), 0, "", false);
```
arvidn: @ssiloti I can't remember anymore why the `outgoing_sockets` must be initialized and used. I can see that the utp_socket_manager can only use sockets from the outgoing_sockets list. It seems to me that it's overlapping a lot with the listen sockets list, especially for UDP.

Me looking at this is related to the SOCKS5 issues people are having, and this [comment/observation](https://github.com/qbittorrent/qBittorrent/issues/11735#issuecomment-568610944) I made.

Even when I set a single listen socket, I get 3 SOCKS5 connections. 1 for the listen socket UDP socket, and 2 for the outgoing socket UDP sockets (because I don't specify any outgoing interface). I would have expected that `m_outgoing_sockets` would be empty, and that the listen socket would be used as the UDP socket for the `utp_socket_manager`.

What would break if `m_outgoing_sockets` were removed entirely, and `utp_socket_manager` was changed to use the UDP sockets from the `m_listen_sockets` array instead?

iirc, the original use of `outgoing_interfaces` was to bind outgoing TCP sockets to specific interfaces. It's not obvious to me what purpose they serve for UDP (that doesn't overlap with listen sockets).

ssiloti: IIRC ougoing_sockets is used with UTP because with IPv6 listen sockets are bound to a specific address by default and libtorrent doesn't necessarily know which interface should be used for an outgoing connection. Using the outgoing socket(s), which are bound to the unspecified address by default, allows the OS to decide how to route outgoing connections.

arvidn: I think one important property is that DHT and UDP trackers use the same socket as uTP. since those will open up the NAT pinhole that incoming uTP connections will be received through. as far as I can tell now, utp_socket_manager doesn't use the listen sockets at all, but the DHT and UDP trackers do, so there is a disconnect there

PR #1950, opened 2017-04-26
remove void* m_userdata

pavel-pimenov:


include/libtorrent/disk_io_thread.hpp
```diff
@@ -516,10 +516,6 @@ namespace aux {
 
 		aux::session_settings m_settings;
 
-		// userdata pointer for the complete_job function, which
-		// is posted to the network thread when jobs complete
-		void* m_userdata;
```
arvidn: how is the context for the complete_job function passed along now?

arvidn:
?

arvidn:
why did you close this?

pavel-pimenov:
i thought m_userdata is used somewhere.

#ifndef TORRENT_DISABLE_UDATA_COMPLETE_JOB
 void* m_userdata;
#endif
?

arvidn:
implied in my question was: is it a mistake that this member is not used? (probably not). Are there any other things that should be removed too? like other occurrences of that macro perhaps?

pavel-pimenov:
hm. i found m_userdata by accident :( maybe my mistake
where to find an example of use complete_job and m_userdata?

PR #1951, opened 2017-04-27
add upnp::create_soap

pavel-pimenov:


include/libtorrent/upnp.hpp
```diff
@@ -240,8 +240,10 @@ struct TORRENT_EXTRA_EXPORT upnp final
 	void get_ip_address(rootdevice& d);
 	void delete_port_mapping(rootdevice& d, int i);
 	void create_port_mapping(http_connection& c, rootdevice& d, int i);
-	void post(upnp::rootdevice const& d, char const* soap
+	void post(upnp::rootdevice const& d, std::string const& soap
```
arvidn: please use ``string_view`` instead of ``std::string`` here

src/upnp.cpp
```diff
@@ -1106,7 +1109,7 @@ void find_error_code(int const type, string_view string, error_code_parse_state&
 	}
 	else if (type == xml_string && state.in_error_code)
 	{
-		std::string error_code_str(string.begin(), string.end());
+		std::string const error_code_str(string.begin(), string.end());
 		state.error_code = std::atoi(error_code_str.c_str());
```
arvidn: the ``error_code_str`` variable could be removed entirely here by just calling: ``string.to_string().c_str()`` and pass that into ``std::atoi()`` directly.

include/libtorrent/upnp.hpp
```diff
@@ -240,8 +240,10 @@ struct TORRENT_EXTRA_EXPORT upnp final
...
 		, char const* soap_action);
+	std::string create_soap(char const* soap_action
```
arvidn: all incoming strings here should probably be ``string_view`` too. Technically, returning a ``std::string`` here will be slightly more costly than the previous version, since it didn't heap allocate the string. not sure if anything need to be done about that though

include/libtorrent/upnp.hpp
```diff
@@ -275,6 +277,11 @@ struct TORRENT_EXTRA_EXPORT upnp final
 
 		// the number of times this mapping has failed
 		int failcount = 0;
+
+		const char* get_protocol_name() const
```
arvidn: I think I would prefer this function be called: ``protocol_name()``.

src/upnp.cpp
```diff
@@ -698,8 +698,8 @@ void upnp::post(upnp::rootdevice const& d, char const* soap
 		"Soapaction: \"%s#%s\"\r\n\r\n"
 		"%s"
 		, d.path.c_str(), d.hostname.c_str(), d.port
-		, int(strlen(soap)), d.service_namespace.c_str(), soap_action
-		, soap);
+		, int(soap.length()), d.service_namespace.c_str(), soap_action
+		, soap.c_str());
```
arvidn: the way to print a ``string_view`` without allocating a new string is like this: ("%*s", int(soap.size()), soap.data())

src/upnp.cpp
```diff
@@ -741,13 +754,16 @@ void upnp::create_port_mapping(http_connection& c, rootdevice& d, int const i)
...
 		, d.mapping[i].local_port
 		, local_endpoint.c_str()
-		, m_user_agent.c_str(), local_endpoint.c_str(), d.mapping[i].local_port
-		, d.lease_duration, soap_action);
+		, m_user_agent.c_str()
+		, local_endpoint.c_str()
+		, d.mapping[i].local_port
+		, d.lease_duration);
+
+	auto const soap = create_soap(soap_action, d.service_namespace, soap_body);
```
arvidn: I'm not super excited about the way this is split up into 2 steps.

src/upnp.cpp
```diff
@@ -708,6 +708,22 @@ void upnp::post(upnp::rootdevice const& d, char const* soap
 #endif
 }
 
+std::string upnp::create_soap(char const* soap_action, std::string const& service_namespace, char const* part)
```
arvidn: maybe you could take a ``std::initialization_list`` of (tag-name, value) pairs and format the whole message in here

arvidn: also, I imagine one benefit of factoring this out into its own function is that it can more easily be tested directly.
and, should this be a member function? (or free function?)

pavel-pimenov: help with std::initialization_list? (example)

src/upnp.cpp
pavel-pimenov: remove \r\n ?

, soap_action, d.service_namespace.c_str(), d.mapping[i].external_port
...
, m_user_agent.c_str(), local_endpoint.c_str(), d.mapping[i].local_port

include/libtorrent/upnp.hpp
arvidn: I meant ``soap_action``, ``service_namespace`` and ``part`` too

src/upnp.cpp
arvidn: for example, take this as a parameter: ``std::initializer_list<std::pair<string_view, string_view>> args``.

and build the string like something like this (in a loop):

```
span<char> buf = buffer;
for (auto a : args)
{
   int written = std::snprintf(buf.data(), buf.size(), "<%*s>%*s</%*s>\n"
      , int(a.first.size()), a.first.data()
      , int(a.second.size()), a.second.data()
      , int(a.first.size()), a.first.data());
   buf = buf.subspan(written);
}
```

something like that. Then you can call it and pass in: ``{ {"NewLeaseDuration"_sv, string_view(duration), { ... }} }``.

For the string view literal, use this: https://github.com/arvidn/libtorrent/blob/master/include/libtorrent/string_view.hpp#L60

pavel-pimenov: for string :
span<char> create_xml_tag(std::initializer_list<std::pair<string_view, string_view>> const args)
{
	span<char> buf;
	for (auto const a : args)
	{
		int written = std::snprintf(buf.data(), buf.size(), "<%*s>%*s</%*s>\n"
			, int(a.first.size()), a.first.data()
			, int(a.second.size()), a.second.data()
			, int(a.first.size()), a.first.data());
		buf = buf.subspan(written);
	}
	return buf;
}
`
"<NewRemoteHost></NewRemoteHost>"
		"<NewExternalPort>%u</NewExternalPort>"
		"<NewProtocol>%s</NewProtocol>"
		"<NewInternalPort>%u</NewInternalPort>"
		"<NewInternalClient>%s</NewInternalClient>"
		"<NewEnabled>1</NewEnabled>"
		"<NewPortMappingDescription>%s at %s:%d</NewPortMappingDescription>"
		"<NewLeaseDuration>%u</NewLeaseDuration>"
`
%u  - unsigned int

PR #1952, opened 2017-04-27
Set connection timeout when next endpoint is tried

jberkel:
After a timeout on the first endpoint the timer won't get reset, so there's no timeout active for the next endpoint on the list, which means the connection can potentially hang forever.

### DIFF
--- a/simulation/test_http_connection.cpp
+++ b/simulation/test_http_connection.cpp
@@ -232,7 +232,7 @@ void run_suite(lt::aux::proxy_settings ps)
 	if (ps.type != settings_pack::socks5
 		&& ps.type != settings_pack::http)
 	{
-		auto expected_code = ps.type == settings_pack::socks4 ?
+		const auto expected_code = ps.type == settings_pack::socks4 ?
 			boost::system::errc::address_family_not_supported :
 			boost::system::errc::address_not_available;
 
@@ -449,6 +449,9 @@ TORRENT_TEST(http_connection_socks5_proxy_names)
 	run_suite(ps);
 }
 
+// tests the error scenario of a http server listening on two sockets (ipv4/ipv6) which
+// both accept the incoming connection but never send anything back. we test that
+// both ip addresses get tried in turn and that the connection attempts time out as expected.
 TORRENT_TEST(http_connection_timeout)
 {
 	sim_config network_cfg;
@@ -486,8 +489,8 @@ TORRENT_TEST(http_connection_timeout)
 	error_code e;
 	sim.run(e);
 	TEST_CHECK(!e);
-	TEST_EQUAL(2, connect_counter);
-	TEST_EQUAL(1, handler_counter);
+	TEST_EQUAL(2, connect_counter); // both endpoints are connected to
+	TEST_EQUAL(1, handler_counter); // the handler only gets called once with error_code == timed_out
 }
 
 void test_proxy_failure(lt::settings_pack::proxy_type_t proxy_type)
--- a/simulation/test_http_connection.cpp
+++ b/simulation/test_http_connection.cpp
@@ -453,12 +453,14 @@ TORRENT_TEST(http_connection_timeout)
 {
 	sim_config network_cfg;
 	sim::simulation sim{network_cfg};
-
+	// server has two ip addresses (ipv4/ipv6)
 	sim::asio::io_service server_ios(sim, address_v4::from_string("10.0.0.2"));
 	sim::asio::io_service server_ios_ipv6(sim, address_v6::from_string("ff::dead:beef"));
-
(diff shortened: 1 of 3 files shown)


arvidn:
I would love to have a test for this. would you mind taking a look at the other tests under simulation?

jberkel:
Sure, will do

arvidn:
and please add a note to ``ChangeLog`` too

simulation/test_http_connection.cpp
```diff
@@ -225,8 +232,12 @@ void run_suite(lt::aux::proxy_settings ps)
 	if (ps.type != settings_pack::socks5
 		&& ps.type != settings_pack::http)
 	{
+		auto expected_code = ps.type == settings_pack::socks4 ?
+			boost::system::errc::address_family_not_supported :
+			boost::system::errc::address_not_available;
+
 		run_test(ps, "http://[ff::dead:beef]:8080/test_file", 0, -1
-			, error_condition(boost::system::errc::address_family_not_supported, generic_category())
+			, error_condition(expected_code, generic_category())
```
jberkel: needed because of changes in libsimulator

simulation/test_http_connection.cpp
```diff
@@ -225,8 +232,12 @@ void run_suite(lt::aux::proxy_settings ps)
 	if (ps.type != settings_pack::socks5
 		&& ps.type != settings_pack::http)
 	{
+		auto expected_code = ps.type == settings_pack::socks4 ?
```
arvidn: slight nit: you could make this ``const``

simulation/test_http_connection.cpp
```diff
@@ -438,6 +449,47 @@ TORRENT_TEST(http_connection_socks5_proxy_names)
 	run_suite(ps);
 }
 
+TORRENT_TEST(http_connection_timeout)
+{
```
arvidn: would you mind adding a comment here explaining the scenario being tested?

simulation/test_http_connection.cpp
```diff
@@ -438,6 +449,47 @@ TORRENT_TEST(http_connection_socks5_proxy_names)
...
+
+	auto c = test_request(client_ios, resolver
+		, "http://dual-stack.test-hostname.com:8080/timeout", data_buffer, -1, -1
+		, timed_out, lt::aux::proxy_settings()
+		, &connect_counter, &handler_counter);
+
+	error_code e;
+	sim.run(e);
+	TEST_CHECK(!e);
+	TEST_EQUAL(2, connect_counter);
```
arvidn: I take it the reasoning is that we tried to connect twice, but succeeded once, right?

src/http_connection.cpp
```diff
@@ -449,12 +449,14 @@ void http_connection::on_timeout(boost::weak_ptr<http_connection> p
...
 			if (!c->m_connecting) c->connect();
+			c->m_last_receive = now;
+			c->m_start_time = c->m_last_receive;
 		}
 		else
 		{
 			c->callback(boost::asio::error::timed_out);
+			return;
 		}
-		return;
```
arvidn: I guess ``c->connect()`` will reset ``c->m_start_time``. And if ``c->m_connecting == true``, I suppose the handler that gets called as a result of ``close()`` will issue the retry. is that right?

arvidn:
looks good, see comments

src/http_connection.cpp
jberkel: yes, but `m_start_time` isn't reset until `on_connected()` is called.

in this test case `m_connecting` is always `false` (the timeout happens after a successful connection), but  we probably need another test covering the case where the connect never succeeds and just hangs in the air (often the case for firewalled connections where packets are dropped).

simulation/test_http_connection.cpp
```diff
@@ -438,6 +449,50 @@ TORRENT_TEST(http_connection_socks5_proxy_names)
...
+
+	auto c = test_request(client_ios, resolver
+		, "http://dual-stack.test-hostname.com:8080/timeout", data_buffer, -1, -1
+		, timed_out, lt::aux::proxy_settings()
+		, &connect_counter, &handler_counter);
+
+	error_code e;
+	sim.run(e);
+	TEST_CHECK(!e);
+	TEST_EQUAL(2, connect_counter); // both endpoints are connected to
```
jberkel: not sure if there's an easy way to test that different endpoints got tried here – maybe we could add some counters to the fake http_servers so one could say
```c++
TEST_EQUAL(1, http.count());
TEST_EQUAL(1, http_ipv6.count());
```

jberkel:
doesn't look like the build failure is related to the changes, since the other jobs passed?

arvidn:
thanks!

PR #1956, opened 2017-04-28
announce to trackers for all listen interfaces

ssiloti:


test/test_tracker.cpp
```diff
@@ -678,16 +680,29 @@ void test_stop_tracker_timeout(bool nostop)
...
 
 	int const count = count_stopped_events(s);
-	TEST_EQUAL(count, nostop ? 0 : 1);
+	if (nostop)
+	{
+		TEST_EQUAL(count, 0);
+	}
+	else
+	{
+		// the exact value will depend on the number of interfaces the host system has
```
arvidn: if the session is bound to a specific interface/IP, only that is used for the announce, right?

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -563,6 +563,14 @@ namespace aux {
 			std::uint16_t ssl_listen_port() const override;
 			std::uint16_t ssl_listen_port(listen_socket_t* sock) const;
 
+			void for_each_listen_socket(std::function<void(aux::session_listen_socket*)> f) override
```
arvidn: since this is in the header already, perhaps it would make sense to make it a template and avoid type-erasure on the function object and allow inlining..

simulation/test_tracker.cpp
```diff
@@ -491,11 +516,14 @@ TORRENT_TEST(test_error)
 		}
 		, [](announce_entry const& ae)
 		{
-			TEST_EQUAL(ae.is_working(), false);
-			TEST_EQUAL(ae.message, "test");
 			TEST_EQUAL(ae.url, "http://tracker.com:8080/announce");
-			TEST_EQUAL(ae.last_error, error_code(errors::tracker_failure));
-			TEST_EQUAL(ae.fails, 1);
+			for (auto const& aep : ae.endpoints)
```
arvidn: should you check that there are exactly 2 endpoints as well?

src/http_connection.cpp
```diff
@@ -530,13 +530,28 @@ void http_connection::on_resolve(error_code const& e
 
 	aux::random_shuffle(m_endpoints.begin(), m_endpoints.end());
 
-	// sort the endpoints so that the ones with the same IP version as our
-	// bound listen socket are first. So that when contacting a tracker,
-	// we'll talk to it from the same IP that we're listening on
-	if (m_bind_addr != address_v4::any())
-		std::partition(m_endpoints.begin(), m_endpoints.end()
+	// if we have been told to bind to a particular address
```
arvidn: this is the meat of the patch, right?

src/session_impl.cpp
```diff
@@ -1151,27 +1151,43 @@ namespace {
...
-		// just _one_ SSL listen port, which one we use depends on which interface
-		// we announce from.
-		if (req.ssl_ctx) req.listen_port = ssl_listen_port();
+		bool use_ssl = req.ssl_ctx != nullptr;
 		req.ssl_ctx = &m_ssl_ctx;
 #endif
 
-#if TORRENT_USE_I2P
-		if (!m_settings.get_str(settings_pack::i2p_hostname).empty())
+		if (req.outgoing_socket)
```
arvidn: something else, upstream, has picked a specific listen socket here. If one has not been picked, we'll announce once for each listen socket. in which case do we just pick one?

arvidn: I can't find any place in this patch where this member is assigned (other than in the else-block below)

src/session_impl.cpp
```diff
@@ -1876,6 +1892,43 @@ namespace {
 			interface_to_endpoints(device, port, ssl, eps);
 		}
 
+#if TORRENT_USE_IPV6
+		std::vector<ip_interface> const ifs = enum_net_interfaces(m_io_service, ec);
+		if (!ec)
+		{
+			// if any listen endpoints are on [::] remove them and add endpoints for
+			// each interface that isn't already included in eps
```
arvidn: would you mind explaining the rationale for this (ideally by extending this comment)

arvidn: it looks like this would be a good candidate for factoring out into a free function and unit test

src/torrent.cpp
```diff
@@ -2775,40 +2825,51 @@ namespace libtorrent {
 			}
 #endif
 
-#ifndef TORRENT_DISABLE_LOGGING
-			debug_log("==> TRACKER REQUEST \"%s\" event: %s abort: %d"
-				, req.url.c_str()
-				, (req.event == tracker_request::stopped ? "stopped"
-					: req.event == tracker_request::started ? "started" : "")
-				, m_abort);
+				req.outgoing_socket = aep.socket;
```
arvidn: I see. this diff wasn't shown by default by github

arvidn: is the indentation right here?

include/libtorrent/aux_/session_impl.hpp
ssiloti: This can't be a template function because it is virtual

src/torrent.cpp
ssiloti: No it's not. I left it this way intentionally to keep the diff clean. I meant to create a separate commit to fix it but forgot. Thanks for reminding me.

include/libtorrent/aux_/session_impl.hpp
arvidn: ah

include/libtorrent/announce_entry.hpp
```diff
@@ -107,20 +97,73 @@ namespace libtorrent {
...
+
+		// tracker URL as it appeared in the torrent file
+		std::string url;
+
+		// the current ``&trackerid=`` argument passed to the tracker.
+		// this is optional and is normally empty (in which case no
+		// trackerid is sent).
+		std::string trackerid;
+
+		std::vector<announce_endpoint> endpoints;
```
arvidn: we need to be careful with this I think. Keep in mind that there are ``torrent_handle::replace_trackers()`` and ``torrent_handle::add_tracker()``, where the client (deliberately or accidentally) could provide a pre-populated list of endpoints. I think it would need to be cleared in that case.

Now that ``announce_entry`` contains a vector, it would probably make sense to make sure it can be moved in places where that makes sense (at least ``add_tracker()`` should probably take the announce_entry by value, not sure about ``replace_trackers()``).

include/libtorrent/announce_entry.hpp
```diff
@@ -159,40 +194,13 @@ namespace libtorrent {
...
-				&& !updating;
-		}
-#endif
-
-		// returns true if we can announce to this tracker now.
-		// The current time is passed in as ``now``. The ``is_seed``
-		// argument is necessary because once we become a seed, we
-		// need to announce right away, even if the re-announce timer
-		// hasn't expired yet.
-		bool can_announce(time_point now, bool is_seed) const;
```
arvidn: for API backwards compatibility, do you think there's a reasonable way to add this and and ``is_working()`` back as deprecated?

include/libtorrent/announce_entry.hpp
```diff
@@ -107,20 +97,73 @@ namespace libtorrent {
...
 		// the max number of failures to announce to this tracker in
 		// a row, before this tracker is not used anymore. 0 means unlimited
 		std::uint8_t fail_limit = 0;
 
-		// the number of times in a row we have failed to announce to this
-		// tracker.
-		std::uint8_t fails:7;
-
-		// true while we're waiting for a response from the tracker.
-		bool updating:1;
```
arvidn: in fact, perhaps all these fields that were moved, should be left in here as well, deprecated, and probably not set to anything. just to simplify upgrading.

arvidn:
Thinking about ``announce_entry`` some more, I think there are a lot of internal state that should not be exposed to the client in there. In some sense you're not really adding much in terms of exposed internal state, but it's more complex internal state now, it being a vector.

I wonder if there's a way to hide this. My first thought was similar to how the internal ``web_seed_t`` derives from the public ``web_seed_entry``. It would mean that ``torrent_handle::trackers()`` would need to construct a copy of just the internal state (i.e. base class) before returning it.

It already needs to copy the vector since it's returned by value, so maybe that's not a big deal.

anyway, thinking about it even more, this refactor should probably be a separate patch anyway. I can do this once this patch has landed

include/libtorrent/announce_entry.hpp
ssiloti: Moving `announce_entry` is not trivial because it gets passed through `torrent_handle::async_call` which will copy any value parameters. We would need to use generalized lambda capture to move the parameters. That is a C++14 feature, although I think it's supported by all compilers we care about.

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -195,6 +195,11 @@ namespace aux {
 			listen_endpoint_t(address adr, int p, std::string dev, bool s)
 				: addr(adr), port(p), device(dev), ssl(s) {}
 
+			bool operator==(listen_endpoint_t const& o)
```
arvidn: this function should probably be const

include/libtorrent/announce_entry.hpp
```diff
@@ -187,18 +187,45 @@ namespace libtorrent {
 
 #ifndef TORRENT_NO_DEPRECATE
 		// deprecated in 1.2
-		// this is false the stats sent to this tracker will be 0
+		// all of these will be set to false or 0
+		// use the corresponding members in announce_endpoint
+		std::uint8_t fails:7;
```
arvidn: I introduced a macro recently to mark members as deprecated as well. would you mind adding it? like this:
``std::uint8_t TORRENT_DEPRECATED_MEMBER fails:7;``

test/test_listen_socket.cpp
```diff
@@ -203,3 +203,73 @@ TORRENT_TEST(partition_listen_sockets_op_ports)
 	TEST_EQUAL(eps.size(), 2);
 }
 
+bool listen_endpoint_cmp(aux::listen_endpoint_t const& l, aux::listen_endpoint_t const& r)
```
arvidn: you added a comparison operator to ``aux::listen_endpoint_t``. I would expect the test to be able to use that instead of this one. Perhaps it was failing because it wasn't marked ``const`` correctly.

ssiloti: Oh damnit. I added this before I decided to implement operator== and just forgot to delete it.

arvidn:
please mention these new features in the ``ChangeLog`` too!

arvidn:
lgtm

PR #1960, opened 2017-04-30
use enum class for ssl bool in listen socket

arvidn:
use std::array instead of C array. make test_listen_socket a bit more compact.

### DIFF
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -114,9 +114,24 @@ namespace libtorrent {
 
 namespace dht {
 
-		struct dht_tracker;
-		class item;
-	}
+	struct dht_tracker;
+	class item;
+
+}
+
+namespace aux {
+
+		struct session_impl;
+		struct session_settings;
+
+#ifndef TORRENT_DISABLE_LOGGING
+		struct tracker_logger;
+#endif
+
+#ifndef TORRENT_DISABLE_DHT
+		TORRENT_EXTRA_EXPORT dht_settings read_dht_settings(bdecode_node const& e);
+		TORRENT_EXTRA_EXPORT entry save_dht_settings(dht_settings const& settings);
+#endif
 
 	struct listen_socket_t final : aux::session_listen_socket
 	{
@@ -164,8 +179,8 @@ namespace dht {
 		int tcp_port_mapping[2];
 		int udp_port_mapping[2];
 
-		// set to true if this is an SSL listen socket
-		bool ssl = false;
+		// indicates whether this is an SSL listen socket or not
+		transport ssl = transport::plaintext;
 
 		// the actual sockets (TCP listen socket and UDP socket)
 		// An entry does not necessarily have a UDP or TCP socket. One of these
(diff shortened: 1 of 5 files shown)


include/libtorrent/aux_/session_impl.hpp
```diff
@@ -114,9 +114,24 @@ namespace libtorrent {
...
+#ifndef TORRENT_DISABLE_LOGGING
+		struct tracker_logger;
+#endif
+
+#ifndef TORRENT_DISABLE_DHT
+		TORRENT_EXTRA_EXPORT dht_settings read_dht_settings(bdecode_node const& e);
+		TORRENT_EXTRA_EXPORT entry save_dht_settings(dht_settings const& settings);
+#endif
 
 	struct listen_socket_t final : aux::session_listen_socket
```
arvidn: I moved ``listen_socket_t`` into ``libtorrent::aux`` namespace

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -164,8 +179,8 @@ namespace dht {
 		int tcp_port_mapping[2];
 		int udp_port_mapping[2];
 
-		// set to true if this is an SSL listen socket
-		bool ssl = false;
+		// indicates whether this is an SSL listen socket or not
+		transport ssl = transport::plaintext;
```
arvidn: increased type-safety by using an enum class here instead of raw ``bool``

include/libtorrent/aux_/session_udp_sockets.hpp
```diff
@@ -93,7 +95,7 @@ namespace libtorrent { namespace aux {
 	private:
 		// round-robin index into sockets
 		// one dimention for IPv4/IPv6 and a second for SSL/non-SSL
-		mutable std::uint8_t index[2][2] = { {0, 0}, {0, 0} };
+		mutable std::array<std::array<std::uint8_t, 2>, 2> index = {{ {{0, 0}}, {{0, 0}} }};
```
arvidn: ``std::array<>`` instead of C-array

test/test_listen_socket.cpp
```diff
@@ -37,209 +37,218 @@ using namespace lt;
...
-	test_equal(sockets.back(), address_v4(), 6881, "", false);
-	test_equal(eps.front(), address_v4::from_string("4.4.4.5"), 6881, "", false);
+	test_equal(sockets.front(), address_v4::from_string("4.4.4.4"), 6881, "", tp::plaintext);
+	test_equal(sockets.back(), address_v4(), 6881, "", tp::plaintext);
+	test_equal(eps.front(), address_v4::from_string("4.4.4.5"), 6881, "", tp::plaintext);
 }
 
 TORRENT_TEST(partition_listen_sockets_port_change)
 {
-	std::list<listen_socket_t> sockets;
```
arvidn: generally just replaced series of ``emplace_back`` into initializer lists

arvidn: @ssiloti does this look right to you?

src/session_impl.cpp
```diff
@@ -5223,8 +5218,13 @@ namespace {
 		}
 
 		std::vector<listen_interface_t> new_ifaces;
+		for (auto& ls : current_ifaces)
+		{
+			ls.ssl = true;
+		}
```
ssiloti: This means there will only be ssl listen sockets, was that the intent?

test/test_listen_socket.cpp
ssiloti: lgtm

src/session_impl.cpp
arvidn: oh. good catch. when I first looked at the ``transform`` below I thought this is what it did, but I forgot to remove this

Issue #1953, opened 2017-04-27
Unable to access specific values in the alert object


MaChengxin:
_libtorrent version (or branch):_
1.1.3.0

_platform/architecture:_
Ubuntu 16.04, x64

_compiler and compiler version:_
gcc (GCC) 5.4.0

_Python version:_
Python 2.7.12

_please describe what symptom you see, what you would expect to see instead and
how to reproduce it._
I am trying to access session statistics and I am only interested in a few of them. It is suggested by the [doc](http://www.libtorrent.org/manual-ref.html#session-statistics) that "Always query the metrics first, to find out the index at which the value is stored".
However, the `alert` object returned by calling `post_session_stats()` doesn't support indexing.

> TypeError: 'session_stats_alert' object has no attribute '\__getitem\__'

So here come two questions:

1. How do we use the returned index of `lt.find_metric_idx("NAME_OF_STATS")`?
2. Can we just use `alert.values["NAME_OF_STATS"]` to get the stats? (e.g. `alert.values["ses.num_incoming_choke"]`)

arvidn:
right, the python binding turns the array into a dictionary mapping the names to the values. i.e. in python you don't have to look up any index up-front

arvidn:
This should probably be documented in the python binding documentation. if  you're up for it, please submit pull request!

arvidn:
yes

MaChengxin:
So I tried to use `alert.values["NAME_OF_STATS"]` to get the stats that I am interested in: 

```
alerts = ses.pop_alerts()
for a in alerts:
	if a.category() & lt.alert.category_t.error_notification:
		print(a)
	elif a.category() & lt.alert.category_t.stats_notification:
		'''
		total number of bytes sent and received by the session
		'''
		print("net.sent_bytes: ", a.values["net.sent_bytes"])
		print("net.sent_payload_bytes: ", a.values["net.sent_payload_bytes"])
		print("net.sent_ip_overhead_bytes: ", a.values["net.sent_ip_overhead_bytes"])
		print("net.sent_tracker_bytes: ", a.values["net.sent_tracker_bytes"])

		print("net.recv_bytes: ", a.values["net.recv_bytes"])
		print("net.recv_payload_bytes: ", a.values["net.recv_payload_bytes"])
		print("net.recv_ip_overhead_bytes: ", a.values["net.recv_ip_overhead_bytes"])
[...]

PR #1961, opened 2017-05-01
Explain retrieving session statistics

MaChengxin:
Based on the discussion here: https://github.com/arvidn/libtorrent/issues/1953

### DIFF
--- a/docs/python_binding.rst
+++ b/docs/python_binding.rst
@@ -107,6 +107,14 @@ To get a python dictionary of the settings, call ``session::get_settings``.
 
 .. _`library reference`: reference.html
 
+
+Retrieving session statistics in Python is more convenient than that in C++.
+
+In C++, one should call ``post_session_stats()`` in the ``session`` object to get ``session_stats_alert``, in which the statistics is stored as an array.
+In order to interpret this array, it is required to call ``session_stats_metrics()`` to get the the indices of these metrics.
+
+In Python it can be done using ``alert.values["NAME_OF_METRICS"]``, where ``alert`` is triggered by ``post_session_stats()``, and ``NAME_OF_METRICS`` is the name of a metric.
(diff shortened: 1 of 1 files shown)


docs/python_binding.rst
```diff
@@ -107,6 +107,14 @@ To get a python dictionary of the settings, call ``session::get_settings``.
 
 .. _`library reference`: reference.html
 
+
+Retrieving session statistics in Python is more convenient than that in C++.
+
+In C++, one should call ``post_session_stats()`` in the ``session`` object to get ``session_stats_alert``, in which the statistics is stored as an array.
```
arvidn: this gives the impression that calling ``post_session_stats()`` is a C++ thing (I think). Would it improve things to move the "in C++" to the next sentence?

docs/python_binding.rst
```diff
@@ -107,6 +107,14 @@ To get a python dictionary of the settings, call ``session::get_settings``.
...
+In order to interpret this array, it is required to call ``session_stats_metrics()`` to get the the indices of these metrics.
+
+In Python it can be done using ``alert.values["NAME_OF_METRICS"]``, where ``alert`` is triggered by ``post_session_stats()``, and ``NAME_OF_METRICS`` is the name of a metric.
```
arvidn: since you can only look up one counter at a time, perhaps the singular form would be better here, i.e. "NAME_OF_METRIC"

MaChengxin: Thanks for pointing out!

docs/python_binding.rst
MaChengxin: Yes, that's better.

MaChengxin:
Just made several changes to make the documentation more concise.
Maybe eventually a git squash is needed?

PR #1964, opened 2017-05-02
remove unique.count(url)

pavel-pimenov:


src/torrent_info.cpp
```diff
@@ -1475,8 +1475,8 @@ namespace libtorrent {
 				web_seed_entry ent(maybe_url_encode(url.string_value().to_string())
 					, web_seed_entry::url_seed);
 				if ((m_flags & multifile) && ent.url[ent.url.size() - 1] != '/') ent.url += '/';
-				if (unique.count(ent.url)) continue;
-				unique.insert(ent.url);
+				auto const uniq = unique.insert(ent.url);
+				if (uniq.second == false) continue;
```
arvidn: I wonder if it would make sense to not even name ``uniq`` here, like this:

``if (!unique.insert(ent.url)) continue;``

arvidn:
are there no tests exercising instertion of duplicate URLs?

pavel-pimenov:
https://github.com/arvidn/libtorrent/blob/d138f0242453bf36d7e5744c35f4452c4dce4467/test/test_torrent_info.cpp#L656 
?

arvidn:
it doesn't seem to exercise the case of duplicate web seeds according to the test coverage.

would you mind adding two explicit tests, similar to these: https://github.com/arvidn/libtorrent/blob/d138f0242453bf36d7e5744c35f4452c4dce4467/test/test_torrent_info.cpp#L181

that add the same URL twice and make sure there's only one added?

pavel-pimenov:
ok

pavel-pimenov:
need help
TORRENT_ASSERT(ti.web_seeds().size() == 2) or 
TORRENT_ASSERT(ti.web_seeds().size() == 4)?

Code:
		entry info;
		info["pieces"] = "aaaaaaaaaaaaaaaaaaaa";
		info["name.utf-8"] = "test1";
		info["name"] = "test__";
		info["piece length"] = 16 * 1024;
		info["length"] = 3245;
		entry::list_type l;
		l.push_back(entry("http://foo.com/bar1"));
		l.push_back(entry("http://foo.com/bar1"));
		l.push_back(entry("http://foo.com/bar2"));
		entry e(l);
		entry torrent;
		torrent["url-list"] = e;
		torrent["httpseeds"] = e;
		torrent["info"] = info;
		std::vector<char> buf;
		bencode(std::back_inserter(buf), torrent);
		torrent_info ti(&buf[0], int(buf.size()), ec);
		TORRENT_ASSERT(ti.web_seeds().size() == 2);

default

arvidn:
that looks right. the logic in ``parse_info_section()`` checks uniqueness per url-type. So you get two URL seeds and two HTTP seeds.

PR #1968, opened 2017-05-03
avoid extra sha1_hash memory copy, minor formatting

aldenml:


include/libtorrent/create_torrent.hpp
```diff
@@ -366,8 +366,7 @@ namespace libtorrent {
 		bool m_include_symlinks:1;
 	};
 
-namespace detail {
-
+	namespace detail {
```
arvidn: I'm actually hoping to go the other direction, where namespaces don't add additional indentation. (in some way to adopt the of C++17 feature of being able to open nested namespaces in a single namespace definition)

src/create_torrent.cpp
```diff
@@ -179,7 +179,7 @@ namespace {
 				st->iothread.abort(true);
 				return;
 			}
-			st->ct.set_hash(piece, sha1_hash(piece_hash));
```
arvidn: this was just a copy constructor, it didn't actually hash the digest, right?

include/libtorrent/create_torrent.hpp
aldenml: ok, I will move the `}` to the start line then :)

src/create_torrent.cpp
aldenml: yes, just a copy constructor

include/libtorrent/create_torrent.hpp
aldenml: actually, `inline void nop(piece_index_t) {}` is already double indented, do you really prefer remove one indentation level here?

arvidn: well, I would prefer no more than one indentation level as the end goal (and probably just make namespaces not indent at all). Small corrections in that direction are fine, I'm procrastinating a bit about changing all the indentations in all files. I don't think it's urgent and it comes with some (merge) risk, so I don't know about that the best approach is there.

in this case, if it's just the ``nop`` function, please fix its indentation

aldenml: done

PR #1969, opened 2017-05-04
add trailing slash

pavel-pimenov:


include/libtorrent/string_util.hpp
```diff
@@ -53,6 +53,11 @@ namespace libtorrent {
 	// internal
 	inline bool is_digit(char c)
 	{ return c >= '0' && c <= '9'; }
+	inline void add_trailing_slash(std::string& url)
```
arvidn: I think a better name would be ``ensure_trailing_slash``, since it's not unconditionally adding one (which I think the current name implies)

PR #1978, opened 2017-05-06
HTTP proxy: avoid connecting to itself

jberkel:
Just wanted to sanity check something before I go off and write tests for this: I don't understand the current logic in `http_stream::connected`:

1. `m_hostname` is set (the proxy hostname)
2. `http_stream::async_connect` sets `m_endpoint` and resolves `m_hostname`.
3. after resolving the hostname a connection to the proxy is established
4. the `http_stream::connected` callbacks sends a proxy `CONNECT` to `m_hostname:m_endpoint.port`, therefore connecting to itself

To me, the correct thing to do would be to just `CONNECT` to `m_endpoint`, but maybe I've overlooked another case.

I run into this particular problem when using a SSL connection + http proxy to download the `.torrent` file.

### DIFF
--- a/simulation/test_http_connection.cpp
+++ b/simulation/test_http_connection.cpp
@@ -489,8 +489,8 @@ TORRENT_TEST(http_connection_timeout_server_stalls)
 	error_code e;
 	sim.run(e);
 	TEST_CHECK(!e);
-	TEST_EQUAL(2, connect_counter); // both endpoints are connected to
-	TEST_EQUAL(1, handler_counter); // the handler only gets called once with error_code == timed_out
+	TEST_EQUAL(connect_counter, 2); // both endpoints are connected to
+	TEST_EQUAL(handler_counter, 1); // the handler only gets called once with error_code == timed_out
 }
 
 // tests the error scenario of a http server listening on two sockets (ipv4/ipv6) neither of which
@@ -541,8 +541,8 @@ TORRENT_TEST(http_connection_timeout_server_does_not_accept)
 	error_code e;
 	sim.run(e);
 	TEST_CHECK(!e);
-	TEST_EQUAL(0, connect_counter); // no connection takes place
-	TEST_EQUAL(1, handler_counter); // the handler only gets called once with error_code == timed_out
+	TEST_EQUAL(connect_counter, 0); // no connection takes place
+	TEST_EQUAL(handler_counter, 1); // the handler only gets called once with error_code == timed_out
 }
 
 void test_proxy_failure(lt::settings_pack::proxy_type_t proxy_type)
@@ -639,8 +639,8 @@ TORRENT_TEST(http_connection_ssl_proxy)
 	error_code e;
 	sim.run(e);
 
-	TEST_EQUAL(1, client_counter);
-	TEST_EQUAL(1, proxy_counter);
+	TEST_EQUAL(client_counter, 1);
+	TEST_EQUAL(proxy_counter, 1);
 	if (e) std::cerr << " run failed: " << e.message() << std::endl;
 	TEST_EQUAL(e, error_code());
 }
--- a/simulation/test_http_connection.cpp
+++ b/simulation/test_http_connection.cpp
@@ -598,6 +598,55 @@ TORRENT_TEST(http_connection_http_error)
 	test_proxy_failure(settings_pack::http);
 }
 
+// Requests a proxied SSL connection. This test just ensures that the correct CONNECT request
+// is sent to the proxy server.
+TORRENT_TEST(http_connection_ssl_proxy)
+{
+	using sim::asio::ip::address_v4;
+	sim_config network_cfg;
+	sim::simulation sim{network_cfg};
+
+	sim::asio::io_service client_ios(sim, address_v4::from_string("10.0.0.1"));
+	sim::asio::io_service server_ios(sim, address_v4::from_string("10.0.0.2"));
+	sim::asio::io_service proxy_ios(sim, address_v4::from_string("50.50.50.50"));
+	lt::resolver res(client_ios);
+
+	sim::http_server http(server_ios, 8080);
+	sim::http_server http_proxy(proxy_ios, 4445);
+
+	lt::aux::proxy_settings ps = make_proxy_settings(settings_pack::http);
+
+	int handler_counter = 0;
+	int server_counter = 0;
+
+	http_proxy.register_handler("/10.0.0.2:8080"
+		, [&handler_counter](std::string method, std::string req, std::map<std::string, std::string>& headers)
+		{
+			handler_counter++;
+			TEST_EQUAL(method, "CONNECT");
+			return sim::send_response(403, "Not supported", 1337);
+		});
+
+	auto h = boost::make_shared<http_connection>(client_ios
+		, res
+		, [&server_counter](error_code const& ec, http_parser const& parser
+		, char const* data, const int size, http_connection& c)
+		{
+			server_counter++;
+			TEST_EQUAL(ec, boost::asio::error::operation_not_supported);
+		});
+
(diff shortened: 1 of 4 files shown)


arvidn:
off the top of my head, I don't believe ``m_hostname`` is the name of the proxy. The proxy hostname is part of the ``settings_pack``. I would expect ``m_hostname`` to be set when the proxy is configured to proxy hostname lookups (i.e. not perform the name lookup locally, but to make the proxy do the name lookup)

jberkel:
`m_hostname` is set here, from `proxy_base::set_proxy()`: 
```c++
void set_proxy(std::string hostname, int port)
{
    m_hostname = hostname;
    m_port = port;
}
```

https://github.com/arvidn/libtorrent/blob/RC_1_1/include/libtorrent/proxy_base.hpp#L63-L67

It comes from `aux::proxy_settings` and gets applied in `instantiate_connection()`:

```c++
bool instantiate_connection(io_service& ios
  , aux::proxy_settings const& ps, socket_type& s ... 
{
    // ...
    if (ssl_context)
    {
        s.instantiate<ssl_stream<http_stream> >(ios, ssl_context);
        str = &s.get<ssl_stream<http_stream> >()->next_layer();
    }
    // ...
   str->set_proxy(ps.hostname, ps.port);  
```

https://github.com/arvidn/libtorrent/blob/1f9f588e75384df41beee5bfab58e143d8a38d54/src/instantiate_connection.cpp#L102-L114

and then, later:
```c++
void http_stream::connected(error_code const& e, boost::shared_ptr<handler_type> h) 
{
    // ...
    if (!m_hostname.empty())
    {
        endpoint = m_hostname + ':' + to_string(m_remote_endpoint.port()).elems;
    }
    else
    {
        endpoint = print_endpoint(m_remote_endpoint);
    }
    write_string("CONNECT " + endpoint + " HTTP/1.0\r\n", p);
```

https://github.com/arvidn/libtorrent/blob/RC_1_1/src/http_stream.cpp#L49-L73

arvidn:
yes, it looks like you're right. I guess the http proxy doesn't support to resolve hostnames on the proxy itself

arvidn:
please add a comment by ``m_hostname`` to make it clear that it's the proxy hostname. Also, please add an entry to the ``ChangeLog``

jberkel:
do the simulator tests support https? i can add a testcase if it's not too much effort

arvidn:
it's probably sufficient to make sure the correct request get generated and sent to the http server (i.e. proxy). and there is support for accepting http requests in the simulator. you just add a lambda to be called when a request is received

arvidn:
or, when you say https, do you means specifically SSL? I don't believe that's supported by the simulator (because it relies on some blocking calls, and the simulator only supports async. calls)

jberkel:
What I meant was if the simulator tests can be compiled with `TORRENT_USE_OPENSSL`. But maybe the code in question can be tested without (`http_connection::start` takes `ssl` as a parameter).

simulation/test_http_connection.cpp
```diff
@@ -598,6 +598,55 @@ TORRENT_TEST(http_connection_http_error)
...
+	auto h = boost::make_shared<http_connection>(client_ios
+		, res
+		, [&server_counter](error_code const& ec, http_parser const& parser
+		, char const* data, const int size, http_connection& c)
+		{
+			server_counter++;
+			TEST_EQUAL(ec, boost::asio::error::operation_not_supported);
+		});
+
+	h->start("10.0.0.2", 8080, seconds(1), 0, &ps, true /*ssl*/);
```
arvidn: I don't understand *ssl* here. is it required to exercise the problem? The suspicious thing about it is that saying true here isn't really supposed to happen when ssl is disabled. if the simulator will support ssl at some point, I would expect this to fail, since the server isn't accepting ssl.

simulation/test_http_connection.cpp
```diff
@@ -598,6 +598,55 @@ TORRENT_TEST(http_connection_http_error)
...
+// Requests a proxied SSL connection. This test just ensures that the correct CONNECT request
+// is sent to the proxy server.
+TORRENT_TEST(http_connection_ssl_proxy)
+{
+	using sim::asio::ip::address_v4;
+	sim_config network_cfg;
+	sim::simulation sim{network_cfg};
+
+	sim::asio::io_service client_ios(sim, address_v4::from_string("10.0.0.1"));
+	sim::asio::io_service server_ios(sim, address_v4::from_string("10.0.0.2"));
```
arvidn: one way to simplify this test would be to remove the server and make the proxy just a http handler that verifies it gets the appropriate request to proxy. i.e. that the address is correct. Now you're also testing that the test-proxy implementation is working (which I don't think is strictly necessary to test for this)

arvidn: I guess there are a bunch of other tests that do this, so it's not a big deal

simulation/test_http_connection.cpp
```diff
@@ -598,6 +598,55 @@ TORRENT_TEST(http_connection_http_error)
...
+
+	sim::http_server http(server_ios, 8080);
+	sim::http_server http_proxy(proxy_ios, 4445);
+
+	lt::aux::proxy_settings ps = make_proxy_settings(settings_pack::http);
+
+	int handler_counter = 0;
+	int server_counter = 0;
+
+	http_proxy.register_handler("/10.0.0.2:8080"
```
arvidn: is there really supposed to be a slash in front there? it doesn't look right to me. My recollection is that the CONNECT command just takes the endpoint as parameter, not a URL or anything.

jberkel: A slash is automatically added by the http server, in `normalize` (https://github.com/arvidn/libsimulator/blob/9e1948731b42b32b41e14ba3f16bd828d53c8a42/src/http_server.cpp#L97), so it won't match the request without the slash.

simulation/test_http_connection.cpp
jberkel: if `ssl` is false, `http_connection` won't exercise the CONNECT  at all, and just rewire the GET to go via the proxy:

* https://github.com/arvidn/libtorrent/blob/fe9f877087c66c746f36eaf1547bd8872a256b55/src/http_connection.cpp#L191-L191)
* https://github.com/arvidn/libtorrent/blob/fe9f877087c66c746f36eaf1547bd8872a256b55/src/http_connection.cpp#L331

I find this part of the code highly confusing (parameters get re-assigned), it could probably be made simpler.

An alternative test would be to call `instantiate_connection` connection directly, but it would require more setup code.

simulation/test_http_connection.cpp
jberkel: The server wasn't actually used, and the variable was named incorrectly. fixed.

simulation/test_http_connection.cpp
jberkel: (i'll submit a separate PR in libsimulator to fix this)

jberkel: 👉 https://github.com/arvidn/libsimulator/pull/32

simulation/test_http_connection.cpp
```diff
@@ -598,6 +598,53 @@ TORRENT_TEST(http_connection_http_error)
...
+			TEST_EQUAL(ec, boost::asio::error::operation_not_supported);
+		});
+
+	h->start("10.0.0.2", 8080, seconds(1), 0, &ps, true /*ssl*/);
+
+	error_code e;
+	sim.run(e);
+
+	TEST_EQUAL(1, client_counter);
+	TEST_EQUAL(1, proxy_counter);
```
arvidn: would you mind swapping the arguments here? I just realized this macro prints "expect" about the second one

jberkel: sure, old junit habits....

arvidn:
lgtm!

PR #1982, opened 2017-05-07
add traversal_algorithm::look_for_nodes

pavel-pimenov:


src/kademlia/traversal_algorithm.cpp
```diff
@@ -593,6 +593,28 @@ void traversal_algorithm::status(dht_lookup& l)
 	l.last_sent = last_sent;
 }
 
+void traversal_algorithm::look_for_nodes(bdecode_node const& r,bool const is_traverse)
```
arvidn: I don't think a raw ``bool`` is a good idea. it's impossible to tell what it means at the call sites. Also, it seems like this function is special-tailored to the two places it's called from and not necessarily a reasonable API boundary.
It would probably be reasonable to take a ``std::function<void(node_endpoint)>`` instead of a bool, and let the call sites decide exactly what to do with the nodes. then you could just feed the node_endpoint straight into the callback function, and use a one-line lambda at the call sites

pavel-pimenov: ok

include/libtorrent/kademlia/observer.hpp
```diff
@@ -122,6 +123,7 @@ struct TORRENT_EXTRA_EXPORT observer : boost::noncopyable
 protected:
 
 	void done();
+	void reply_r(msg const&, bdecode_node const& r_id);
```
arvidn: please call this function ``reply`` as well

include/libtorrent/kademlia/observer.hpp
```diff
@@ -38,6 +38,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include <libtorrent/time.hpp>
 #include <libtorrent/address.hpp>
+#include <libtorrent/bdecode.hpp>
```
arvidn: please forward declare ``bdecode_node`` instead of including this header. (you can include the header in the .cpp file)

src/kademlia/traversal_algorithm.cpp
```diff
@@ -664,5 +651,22 @@ void traversal_observer::reply(msg const& m)
 	// it. For instance if it's a bootstrap node.
 	set_id(node_id(id.string_ptr()));
 }
+void traversal_observer::reply(msg const& m)
```
arvidn: please add a blank line between these functions

src/kademlia/traversal_algorithm.cpp
```diff
@@ -607,10 +607,7 @@ void traversal_algorithm::look_for_nodes(bdecode_node const& r,bool const is_tra
 		while (end - nodes >= 20 + protocol_size + 2)
 		{
 			node_endpoint const nep = read_node_endpoint(protocol, nodes);
-			if(is_traverse)
-				traverse(nep.id, nep.ep);
-			else
-				get_node().m_table.heard_about(nep.id, nep.ep);
+			f(nep);
```
arvidn: I don't think you should give ``nep`` a name. just pass it straight into the function

src/kademlia/traversal_algorithm.cpp
```diff
@@ -593,7 +593,7 @@ void traversal_algorithm::status(dht_lookup& l)
 	l.last_sent = last_sent;
 }
 
-void traversal_algorithm::look_for_nodes(bdecode_node const& r,bool const is_traverse)
+void traversal_algorithm::look_for_nodes(bdecode_node const& r, std::function<void(const node_endpoint&)> f)
```
arvidn: does this function need to be a member of ``traversal_algorithm``? If not, I think it should be made a free function

arvidn: and a unit test for that free function would be awesome!

pavel-pimenov: free function? look_for_nodes use traversal_algorithm::get_node()

arvidn: I think you should pass that in as a parameter. The work this function perform does not really affect or relate to the node object. it's parsing a dht message

arvidn:
this is a pretty scary looking error, https://travis-ci.org/arvidn/libtorrent/jobs/230737724#L1345
can you think of how this could be related to these changes?

src/kademlia/traversal_algorithm.cpp
pavel-pimenov: ok.

pavel-pimenov:
no :(

src/kademlia/traversal_algorithm.cpp
```diff
@@ -593,6 +593,24 @@ void traversal_algorithm::status(dht_lookup& l)
 	l.last_sent = last_sent;
 }
 
+void traversal_algorithm::look_for_nodes(bdecode_node const& r, std::function<void(const node_endpoint&)> f)
+{
+	udp const protocol = get_node().protocol();
+	int const protocol_size = int(detail::address_size(protocol));
+	char const* nodes_key = get_node().protocol_nodes_key();
```
arvidn: I really think this should be a parameter instead of the whole function being a member function. (i.e. ``nodes_key``)

arvidn:
I think this is a reasonable patch apart from it being a member function. that seems gratuitous.

pavel-pimenov:
thank you. remove member function (monday.)

src/kademlia/traversal_algorithm.cpp
```diff
@@ -593,6 +593,24 @@ void traversal_algorithm::status(dht_lookup& l)
 	l.last_sent = last_sent;
 }
 
+void look_for_nodes(node const& node, bdecode_node const& r, std::function<void(const node_endpoint&)> f)
```
arvidn: I really think the "nodes_key" should be the parameter here, not the whole node. Imagine writing a unit test for this function. As it stands now you would have to construct a whole DHT node object, just in order for this function to be able to query for the ``protocol_nodes_key()`` from it. It makes the function have a lot more dependencies than it needs.

If you instead just pass in a ``char const*`` or ``string_view`` for the actual key name, it would be a lot simpler to test.

arvidn: I suppose you would need the ``protocol`` object as well. perhaps the nodes key could be derived from it too..

PR #1994, opened 2017-05-14
close listen sockets after closing all connections

ssiloti:
This fixes an assertion failure on shutdown if an IPv6 UTP socket tries
to send a FIN to a closed UDP socket. Because the closed socket reports
its local_endpoint() as udp::endpoint() it trips the protocol check in
send_udp_packet.

I'm a bit nervous about this change because I don't see why it wasn't done this way in the first place.

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -969,25 +969,6 @@ namespace aux {
 		}
 		m_incoming_sockets.clear();
 
-		// close the listen sockets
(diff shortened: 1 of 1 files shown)


arvidn:
lgtm. probably an oversight.

technically this still doesn't let utp sockets close gracefully though, since they aren't waiting for the syn+ack, and aren't re-sending any data in case there's been packet loss.

PR #1999, opened 2017-05-16
don't abort the existing torrent when attempting to add it again

ssiloti:


src/session_impl.cpp
```diff
@@ -4794,7 +4794,11 @@ namespace {
 
 		// if this was an existing torrent, we can't start it again, or add
 		// another set of plugins etc. we're done
-		if (!added) return handle;
+		if (!added)
+		{
+			abort_torrent.disarm();
```
arvidn: ah, this is the case where we set "duplicate is not an error" and we add the same torrent twice back-to-back. Previously the second torrent would end up aborting the first one. This should be easy to test. I can whip something up

PR #2001, opened 2017-05-16
Fix busy loop when a torrent has multiple trackers

ssiloti:


src/torrent.cpp
```diff
@@ -2753,7 +2753,7 @@ namespace libtorrent {
 				continue;
 
 			if (ae.tier > state.tier && state.sent_announce
-				&& !settings().get_bool(settings_pack::announce_to_all_tiers)) break;
+				&& !settings().get_bool(settings_pack::announce_to_all_tiers)) continue;
```
arvidn: It's not obvious what this change does

ssiloti: The problem is that if the first endpoint in the list triggers this condition, it will prevent all of the other endpoints from being examined. If one of those endpoints should have an announce sent on it there will be a loop because `update_tracker_timer `sees that the announce is due, but `announce_with_tracker `doesn't send it.

PR #2007, opened 2017-05-18
stack overflow libtorrent::natpmp::update_mapping 

### LINKED ISSUE #1370: [Stack Overflow] libtorrent::natpmp::update_mapping
https://drdump.com/Problem.aspx?ProblemID=249669&login=guest

flylinkdc_x64!libtorrent::natpmp::update_mapping+0x1fb

Repeated 10726 times:
flylinkdc_x64!libtorrent::natpmp::update_mapping+0x200

flylinkdc_x64!libtorrent::natpmp::mapping_expired+0x4f
flylinkdc_x64!boost::asio::detail::wait_handler<std::_Binder<std::_Unforced,void (__cdecl libtorrent::natpmp::*)(boost::system::error_code const & __ptr64,int) __ptr64,std::shared_ptr<libtorrent::natpmp>,std::_Ph<1> const & __ptr64,int & __ptr64> >::do_complete+0xd2
flylinkdc_x64!boost::asio::detail::win_iocp_io_service::do_one+0x25d
flylinkdc_x64!boost::asio::detail::win_iocp_io_service::run+0xba
flylinkdc_x64!std::_LaunchPad<std::unique_ptr<std::tuple<<lambda_b36f97b57e062a41dc0986f1cc6f4630> >,std::default_delete<std::tuple<<lambda_b36f97b57e062a41dc0986f1cc6f4630> > > > >::_Run+0x91
flylinkdc_x64!std::_Pad::_Call_func+0x9

pavel-pimenov:
Fix #1370

### DIFF
--- a/include/libtorrent/natpmp.hpp
+++ b/include/libtorrent/natpmp.hpp
@@ -111,6 +111,12 @@ private:
 
 		// set to true while we're waiting for a response
 		bool outstanding_request = false;
+
+		void set_none()
+		{
+			protocol = aux::portmap_protocol::none;
+			act = action::none;
+		}
 	};
 
 	aux::portmap_callback& m_callback;
--- a/src/natpmp.cpp
+++ b/src/natpmp.cpp
@@ -277,8 +277,7 @@ void natpmp::try_next_mapping(int const i)
 
 	auto const m = std::find_if(
 		m_mappings.begin(), m_mappings.end()
-		, [] (mapping_t const& ma) { return ma.act != mapping_t::action::none &&
-									ma.protocol != portmap_protocol::none; });
+		, [] (mapping_t const& ma) { return ma.act != mapping_t::action::none; });
 
 	if (m == m_mappings.end())
 	{
--- a/src/natpmp.cpp
+++ b/src/natpmp.cpp
@@ -564,7 +564,7 @@ void natpmp::on_reply(error_code const& e
 	if (result != 0)
 	{
 		// TODO: 3 it would be nice to have a separate NAT-PMP error category
-		errors::error_code_enum const errors[] =
+		static errors::error_code_enum const errors[] =
 		{
 			errors::unsupported_protocol_version,
 			errors::natpmp_not_authorized,
--- a/src/natpmp.cpp
+++ b/src/natpmp.cpp
@@ -200,7 +200,7 @@ void natpmp::disable(error_code const& ec)
 	{
 		if (i->protocol == portmap_protocol::none) continue;
 		portmap_protocol const proto = i->protocol;
-		i->protocol = portmap_protocol::none;
+		i->set_none();
 		int const index = int(i - m_mappings.begin());
 		m_callback.on_port_mapping(index, address(), 0, proto, ec
 			, aux::portmap_transport::natpmp);
@@ -219,8 +219,7 @@ void natpmp::delete_mapping(int const index)
 	if (m.protocol == portmap_protocol::none) return;
 	if (!m.map_sent)
 	{
-		m.act = mapping_t::action::none;
-		m.protocol = portmap_protocol::none;
+		m.set_none();
 		return;
 	}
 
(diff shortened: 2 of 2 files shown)


src/natpmp.cpp
```diff
@@ -278,7 +277,8 @@ void natpmp::try_next_mapping(int const i)
 
 	auto const m = std::find_if(
 		m_mappings.begin(), m_mappings.end()
-		, [] (mapping_t const& ma) { return ma.act != mapping_t::action::none; });
+		, [] (mapping_t const& ma) { return ma.act != mapping_t::action::none &&
```
arvidn: since we clear the action everywhere, presumably we don't need to check the protocol here, right?

src/natpmp.cpp
```diff
@@ -564,13 +564,13 @@ void natpmp::on_reply(error_code const& e
 	if (result != 0)
 	{
 		// TODO: 3 it would be nice to have a separate NAT-PMP error category
-		errors::error_code_enum errors[] =
+		errors::error_code_enum const errors[] =
```
arvidn: this should probably be static as well

arvidn:
do you understand how the code can end up in the infinite loop/recursion?
does this apply to the upnp code as well?

src/natpmp.cpp
pavel-pimenov: https://github.com/arvidn/libtorrent/issues/1370#issuecomment-301741160

arvidn: @pavel-pimenov so what do you think? can you explain why this change is necessary?

pavel-pimenov: i can not reproduce stack overflow  :-(

int natpmp::add_mapping(portmap_protocol const p, int const external_port
	, tcp::endpoint const local_ep)
{
#ifdef _DEBUG
	static bool is_first = false;
	if (!is_first)
	{
		is_first = true;
		m_mappings.resize(4);

		m_mappings[0].protocol = portmap_protocol::none;
		m_mappings[1].protocol = portmap_protocol::none;
		m_mappings[2].protocol = portmap_protocol::tcp;
		m_mappings[3].protocol = portmap_protocol::udp;

		m_mappings[0].act = mapping_t::action::add;
		m_mappings[1].act = mapping_t::action::none;
		m_mappings[2].act = mapping_t::action::none;
		m_mappings[3].act = mapping_t::action::none;

		m_mappings[0].external_port = 8999;
		m_mappings[1].external_port = 8999;
		m_mappings[2].external_port = 8999;
		m_mappings[3].external_port = 8999;

		m_mappings[0].local_port = 8999;
		m_mappings[1].local_port = 8999;
		m_mappings[2].local_port = 8999;
		m_mappings[3].local_port = 8999;

		m_mappings[0].map_sent = true;
		m_mappings[1].map_sent = true;
		m_mappings[2].map_sent = true;
		m_mappings[3].map_sent = true;

		m_mappings[0].outstanding_request = false;
		m_mappings[1].outstanding_request = false;
		m_mappings[2].outstanding_request = false;
		m_mappings[3].outstanding_request = false;

		m_mappings[0].expires = aux::time_now();
		m_mappings[1].expires = aux::time_now();
		m_mappings[2].expires = aux::time_now();
		m_mappings[3].expires = aux::time_now();
	}

arvidn: I haven't dug into this problem or analyzed the code and the details you posted in the issue. That's why I'm asking, if you understand the problem it would save me some time if you could explain it.

include/libtorrent/natpmp.hpp
```diff
@@ -111,6 +111,12 @@ struct TORRENT_EXTRA_EXPORT natpmp
 
 		// set to true while we're waiting for a response
 		bool outstanding_request = false;
+
+		void set_none()
```
arvidn: I wonder if "clear()" is a better name for this

arvidn:
as far as I can tell, this affects RC_1_1 as well. The problem is if there end up being a mapping with an action set on it, but with the protocol cleared (to indicate the mapping has been removed).

At the end of a scan of mappings to send any updates we need, try_next_mapping() will scan for updates to any mapping, to catch any changes made during the sweep. It will find this mapping with an action on it and call ``update_mapping`` for it. However, update mapping will realize that its protocol field is cleared, and hence deleted, and will just move on to the next mapping. At the end of the sweep, the same check will be made again and the "corrupt" mapping will be tried again.

arvidn:
replaced by https://github.com/arvidn/libtorrent/pull/2012
thanks for your report!

PR #2012, opened 2017-05-20
fix NAT-PMP crash when removing a mapping at the wrong time

arvidn:


pavel-pimenov:
please, push fix for master?

arvidn:
it should be in master already. https://github.com/arvidn/libtorrent/commit/604904c80cecd8fe14085fe8c9af08147651483b

PR #2018, opened 2017-05-21
Fix Debug build when building with CMake

mnaamani:
Flags added to CMAKE_C_FLAGS_DEBUG are not applied to C++ files, CMAKE_CXX_FLAGS_DEBUG should be used instead.


arvidn:
I don't think there's a CI job to test the cmake build

PR #2021, opened 2017-05-22
when stopping a torrent, never perform a name lookup on the tracker

arvidn:
only announce to trackers whose IP we already know. This is expected to make shutdowns not hang

### DIFF
--- a/src/http_tracker_connection.cpp
+++ b/src/http_tracker_connection.cpp
@@ -229,8 +229,8 @@ namespace libtorrent
 			, ps.proxy_tracker_connections ? &ps : NULL
 			, 5, user_agent, bind_interface()
 			, tracker_req().event == tracker_request::stopped
-				? resolver_interface::prefer_cache
-				: resolver_interface::abort_on_shutdown
+				? resolver_interface::cache_only : 0
+				| resolver_interface::abort_on_shutdown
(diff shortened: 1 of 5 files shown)


src/http_tracker_connection.cpp
```diff
@@ -229,8 +229,8 @@ namespace libtorrent
 			, ps.proxy_tracker_connections ? &ps : NULL
 			, 5, user_agent, bind_interface()
 			, tracker_req().event == tracker_request::stopped
-				? resolver_interface::prefer_cache
-				: resolver_interface::abort_on_shutdown
+				? resolver_interface::cache_only : 0
+				| resolver_interface::abort_on_shutdown
```
aldenml: a little confused by the use of `0|` here :(

aldenml: sorry, I see, now you `|` the flag `resolver_interface::abort_on_shutdown`, nvm

arvidn:
I suspect in master the tracker endpoint list could be used directly instead of relying on this cache

arvidn:
@aldenml do you think this patch makes sense? Or do you think it's too loose about announcing to trackers?

aldenml:
@arvidn I find the idea very reasonable

PR #2025, opened 2017-05-23
add base_mapping_t

pavel-pimenov:


include/libtorrent/aux_/portmap.hpp
```diff
@@ -66,6 +66,31 @@ namespace libtorrent { namespace aux {
 	protected:
 		~portmap_callback() {}
 	};
+	struct base_mapping_t
```
arvidn: would you mind adding a blank line in between here, and rename ``base_mapping_t`` to just ``base_mapping``

include/libtorrent/aux_/portmap.hpp
```diff
@@ -66,6 +66,31 @@ namespace libtorrent { namespace aux {
...
+		action act = action::none;
+
+		// the external (on the NAT router) port
+		// for the mapping. This is the port we
+		// should announce to others
+		int external_port = 0;
+
+		aux::portmap_protocol protocol = aux::portmap_protocol::none;
+
+		char const* protocol_name() const
```
arvidn: I think these functions really should be free functions, something like this:

``char const* to_string(aux::portmap_protocol);`` and char const* to_string(aux::action);``

include/libtorrent/aux_/portmap.hpp
```diff
@@ -66,6 +66,32 @@ namespace libtorrent { namespace aux {
 	protected:
 		~portmap_callback() {}
 	};
+
+		struct base_mapping
```
arvidn: this indentation seems wrong

include/libtorrent/aux_/portmap.hpp
```diff
@@ -66,6 +66,32 @@ namespace libtorrent { namespace aux {
...
+	{
+		enum class action : std::uint8_t { none, add, del };
```
arvidn: please hoist this enum class out of the ``base_mapping`` struct. I think it should belong with the other enum classes (``portmap_protocol`` and ``portmap_transport``). Would it cause a lot of trouble to do this? perhaps ``action`` is a bit too broad of a name. ``portmap_action`` seems to fit in with the other enums. what do you think?

my thinking is that since it's a proper type (enum class) it doesn't need to be "potected" by being scoped by the class

include/libtorrent/aux_/portmap.hpp
```diff
@@ -66,6 +66,32 @@ namespace libtorrent { namespace aux {
...
+		time_point expires;
+
+		action act = action::none;
+
+		// the external (on the NAT router) port
+		// for the mapping. This is the port we
+		// should announce to others
+		int external_port = 0;
+
+			portmap_protocol protocol = portmap_protocol::none;
```
arvidn: it looks like the indentation is wrong from here on. At first I thought the ``to_string()`` functions were still members

src/natpmp.cpp
```diff
@@ -315,12 +314,11 @@ void natpmp::update_mapping(int const i)
 	if (should_log())
 	{
 		log("update-mapping: proto: %s port: %d local-port: %d action: %s ttl: %" PRId64
-			, (m.protocol == portmap_protocol::none
-				? "none" : m.protocol == portmap_protocol::tcp ? "tcp" : "udp")
+			, m.protocol == portmap_protocol::none
+				? "none" : aux::to_string(m.protocol)
```
arvidn: I don't believe you need to specify ``aux::`` when calling ``to_string``. It should pick the right function by ADL

src/natpmp.cpp
```diff
@@ -176,6 +176,21 @@ bool natpmp::should_log() const
...
 }
 
+void natpmp::mapping_log(char const* op, mapping_t const& m) const
+{
+	if (should_log())
+	{
+		log("%s-mapping: proto: %s port: %d local-port: %d action: %s ttl: %" PRId64
+			, op
+			, m.protocol == portmap_protocol::none
+			? "none" : to_string(m.protocol)
```
arvidn: it would probably make sense to bake-in support for "none" into ``to_string()`` and not check for it here. what do you think?

arvidn:
lgtm, just one comment

src/natpmp.cpp
pavel-pimenov: "none" using only for #ifndef TORRENT_DISABLE_LOGGING
default

PR #2026, opened 2017-05-24
restore find_plugin API

ssiloti:
My reputation plugin uses this API.

@aldenml For future reference, the interfaces in `extensions.hpp` and `peer_connection_handle` are part of the public API. Do not remove parts of them just because libtorrent itself isn't using them.

### DIFF
--- a/include/libtorrent/extensions.hpp
+++ b/include/libtorrent/extensions.hpp
@@ -370,6 +370,10 @@ namespace libtorrent {
 		// hidden
 		virtual ~peer_plugin() {}
 
+		// This function is expected to return the name of
+		// the plugin.
+		virtual span<char const> type() const { return span<char const>(); }
+
 		// can add entries to the extension handshake
 		// this is not called for web seeds
 		virtual void add_handshake(entry&) {}
--- a/include/libtorrent/extensions.hpp
+++ b/include/libtorrent/extensions.hpp
@@ -372,7 +372,7 @@ namespace libtorrent {
 
 		// This function is expected to return the name of
 		// the plugin.
-		virtual span<char const> type() const { return span<char const>(); }
+		virtual string_view type() const { return {}; }
 
 		// can add entries to the extension handshake
 		// this is not called for web seeds
--- a/src/peer_connection.cpp
+++ b/src/peer_connection.cpp
@@ -518,6 +518,16 @@ namespace libtorrent {
 		TORRENT_ASSERT(is_single_thread());
 		m_extensions.push_back(ext);
 	}
+
+	peer_plugin const* peer_connection::find_plugin(span<char const> type)
+	{
+		TORRENT_ASSERT(is_single_thread());
+		for (auto p : m_extensions)
+		{
+			if (p->type() == type) return p.get();
+		}
+		return nullptr;
+	}
 #endif
 
 	void peer_connection::send_allowed_set()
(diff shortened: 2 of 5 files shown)


aldenml:
That was removed ages ago and had appropriated review, I believe the plugin API is public but without guarantee of stability. In any case, in isolation, it's hard to see the use case of this and a case could be made against dangerous acccess to any plugin through the public handle API.

Is the problem maintaining the bookkeeping of connections?

ssiloti:
Yes I missed this change when it was reviewed. It surely didn't help that the PR had the nebulous title "peer_plugin refactor". While obviously the public API is permitted to change between major releases, changes should not be made without a good cause.

I need this API to get the reputation plugin associated with a `peer_connection_handle`. Without it I would have to maintain a mapping of reputation peer_plugin instances to peer connection handles which would be absurdly wasteful. I don't see there being much of safety concern here. Plugins are already granted a great deal of trust, and getting the `peer_plugin` pointer doesn't grant access to the plugin's internals unless you cast it to the underlying type, at which point the author is clearly responsible for whatever happens.

include/libtorrent/extensions.hpp
```diff
@@ -370,6 +370,10 @@ namespace libtorrent {
 		// hidden
 		virtual ~peer_plugin() {}
 
+		// This function is expected to return the name of
+		// the plugin.
+		virtual span<char const> type() const { return span<char const>(); }
```
arvidn: I think ``string_view`` might be a better return type. and to default construct one, you can just do ``{}``.

src/peer_connection.cpp
```diff
@@ -518,6 +518,16 @@ namespace libtorrent {
 		TORRENT_ASSERT(is_single_thread());
 		m_extensions.push_back(ext);
 	}
+
+	peer_plugin const* peer_connection::find_plugin(span<char const> type)
+	{
+		TORRENT_ASSERT(is_single_thread());
+		for (auto p : m_extensions)
```
arvidn: this looks like a ``std::find_if()``

src/peer_connection.cpp
```diff
@@ -518,6 +518,16 @@ namespace libtorrent {
 		TORRENT_ASSERT(is_single_thread());
 		m_extensions.push_back(ext);
 	}
+
+	peer_plugin const* peer_connection::find_plugin(span<char const> type)
```
arvidn: I think ``string_view`` would be better here too

arvidn:
I can't recall what (or if) the rationale for the removal was. It's possible I was thinking that plugins _could_ do their own tracking of which objects have been created. I don't have a strong opinion against it though.
The previous version took a C-string though, didn't it? Is it OK to change it to a ``string_view``?

aldenml:
The main idea was remove that method, it does not have a clear use other than allow some form of coding for some plugins, or even a clear semantic (what's a `type`? what if other plugins use the same "string"?). I think plugins _should_ do all the tracking on their own.

Having said that, by no means this is critical or a big deal, just give it a second thought API design wise, that's all.

ref: https://github.com/arvidn/libtorrent/pull/1002

ssiloti:
Plugins _could_ do their own tracking, but it would mean maintaining a `std::map<peer_connection_handle, peer_plugin>` which is both more error prone and a waste of resources because it would be duplicating the mapping which already exists in libtorrent.

Yes this interface originally took C strings, I'll swap in `string_view` instead of `span`.

arvidn:
lgtm

PR #2028, opened 2017-05-25
use session_params to load dht state before creating the session

arvidn:
in client_test


arvidn:
@aldenml since you added the ``session_params``, would you mind double checking this?

aldenml:
LGTM

PR #2031, opened 2017-05-26
make file open mode flags be an enum class for type-safety

arvidn:
The main purpose of this patch is to lay the foundation for transitioning flags over to enum classes and make new flags enum classes. This is based on Anthony William's [blog post](https://www.justsoftwaresolutions.co.uk/cplusplus/using-enum-classes-as-bitfields.html).


arvidn:
It does alter the ``storage_interface``, since it replaces ``uint32_t`` with ``open_mode_t``.

PR #2034, opened 2017-05-27
make session_handle copyable instead of movable

ssiloti:
This class wraps a raw pointer so there's no point making it movable and
non-copyable.


arvidn:
this seems reasonable. but is it reasonable for this to represent a non-owning raw pointer? Would it perhaps make more sense for it to be a ``weak_ptr``? (in which case it should probably be both movable and copyable)

arvidn:
oh cool, I take it this works?

include/libtorrent/error_code.hpp
```diff
@@ -92,6 +92,8 @@ namespace libtorrent {
 			invalid_escaped_string,
 			// Operation is not permitted since the session is shutting down
 			session_is_closing,
+			// The session_handle is not referring to a valid session_impl
+			invalid_session_handle,
```
arvidn: I've tried to keep the error codes stable across releases. That's why I've left gaps. Would you mind moving this to an empty slot further down?

arvidn: the reason for keeping them stable is it may be a way to translate error messages in user code.

arvidn:
lgtm (apart from the error code comment)

PR #2036, opened 2017-05-28
make move constructors and move assignment operators noexcept

arvidn:


include/libtorrent/socket.hpp
```diff
@@ -81,6 +81,26 @@ namespace libtorrent {
 	using boost::asio::null_buffers;
 #endif
 
+	// this is a silly wrapper for address and endpoint to make their move
+	// constructors be noexcept (because boost.asio is incorrectly making them
+	// potentially throwing). This can be removed once libtorrent uses the
+	// networking TS
+	template <typename T>
+	struct noexcept_move_constructible : T
```
arvidn: this is kind of an unfortunate hack, but the alternative would be to manually spell out a lot of alert type constructors, to work around that the defaulted move constructor can't be noexcept

aldenml:
I have a minor technical question: Is `digest32` really a "movable" type in practice (since it's stack allocated)?

arvidn:
Yeah, it probably doesn't make sense to give it a move constructor and move assignment operator. I believe it would be enough to have a no-throw copy constructor and copy assignment operator.

There's also some mystery to be solved for GCC, not sure why its implicit noexcept specifier is empty.

PR #2040, opened 2017-05-29
sort input files

bmwiedemann:
when building packages (e.g. for openSUSE Linux)
(random) filesystem order of input files
influences ordering of functions in the output,
thus without the patch, builds (in disposable VMs) would differ.

See https://reproducible-builds.org/ for why this matters.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -127,7 +127,7 @@ else:
 
         ext = [Extension(
             'libtorrent',
-            sources=source_list,
+            sources=sorted(source_list),
             language='c++',
             include_dirs=flags.include_dirs,
             library_dirs=flags.library_dirs,


arvidn:
does this actually make a difference? I would expect the link order would be the thing that matters. doesn't distutils compile translation units in parallel?

arvidn:
I'm all for reproducible builds. But I would imagine there are a lot more things to go into it than this. For instance, the libtorrent assert macro uses ``__FILE__``, which means the (debug) build will include the path to the build directory.

Can you confirm that this change fixes some aspect of the build that is non-deterministic right now?

bmwiedemann:
It does make it reproducible, when building under a constant path (as we do in openSUSE).
The trick with the list of .c files is that they are turned into .o files one-by-one and those are linked into libtorrent.so in the same (random) order

arvidn:
I see, so the order in which the .cpp files are specified is preserved for the .o files when linked. Do you know what the link order affects specifically? All I can think of is the linked list of global constructors. Is that it?

bmwiedemann:
I publish resulting diffs at
http://rb.zq1.de/compare.factory-20170523/libtorrent-rasterbar-compare.out

and there you can see that it is mostly re-ordering functions, but sometimes the optimizer will also do slight variations (using different registers or different types of jumps)

arvidn:
great, thanks! I would be happy to accept this patch against the current stable branch as well, ``RC_1_1``

PR #2042, opened 2017-05-30
only listen on preferred IPv6 addresses on Windows

ssiloti:
We shouldn't listen on addresses which are duplicates, deprecated,
or tentative unless explicitly requested.

This is only supported on Windows for now because the Linux interface
to get this information is much less convenient for libtorrent to use.
As far as I've been able to determine, Linux only exposes this information
via the RTM_NEWADDR and RTM_GETADDR netlink messages. Currently libtorrent
uses getifaddrs() to enumerate network interfaces on Linux so this will
probably need to be changed to use netlink if available.

MacOS: ???

### DIFF
--- a/include/libtorrent/enum_net.hpp
+++ b/include/libtorrent/enum_net.hpp
@@ -62,6 +62,9 @@ namespace libtorrent {
 		address netmask;
 		char name[64];
 		int mtu;
+		// an interface is preferred if its address is
+		// not tenative/duplicate/deprecated
+		bool preferred = true;
 	};
 
 	struct ip_route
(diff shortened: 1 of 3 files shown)


include/libtorrent/enum_net.hpp
```diff
@@ -62,6 +62,9 @@ namespace libtorrent {
 		address netmask;
 		char name[64];
 		int mtu;
+		// an interface is preferred if it's address is
```
arvidn: nit: "it's" -> "its"

arvidn:
lgtm

Issue #2044, opened 2017-05-30
crash libtorrent::aux::session_impl::on_tick


pavel-pimenov:
libtorrent version (or branch): master
platform/architecture: windows x86/x64
compiler and compiler version: VC++2017

https://drdump.com/Problem.aspx?ProblemID=298184&login=guest
https://drdump.com/Problem.aspx?ProblemID=297638&login=guest

steps to reproduse:
1. add sleep std::string entry::to_string()
![default](https://cloud.githubusercontent.com/assets/920705/26599689/28baae36-4582-11e7-9e15-6dff276d3f70.png)

2. start download magnet
3. wait...

![default](https://cloud.githubusercontent.com/assets/920705/26599772/71a0ed36-4582-11e7-8460-d8513aec9f27.png)

_DEBUG_ERROR("vector iterator not incrementable");

![default](https://cloud.githubusercontent.com/assets/920705/26599350/033d71f8-4581-11e7-870e-a446d4205b35.png)

![default](https://cloud.githubusercontent.com/assets/920705/26599401/2c86787a-4581-11e7-899a-188b228eb1b5.png)

arvidn:
that patch doesn't attempt to fix this issue, right?

pavel-pimenov:
yes :(

ajax16384:
````c++
m_iterating_connections == 1
     !assert_fail(const char * expr, int line, const char * file, const char * function, const char * value, int kind) Line 357	C++
>	!libtorrent::torrent::connect_to_peer(libtorrent::torrent_peer * peerinfo, bool ignore_limit) Line 6537	C++
 	!libtorrent::peer_connection::connect_failed(const boost::system::error_code & e) Line 4026	C++
 	!libtorrent::peer_connection::second_tick(int tick_interval_ms) Line 4735	C++
 	!libtorrent::torrent::second_tick(int tick_interval_ms) Line 8998	C++
 	!libtorrent::aux::session_impl::on_tick(const boost::system::error_code & e) Line 3408	C++
.... 	
````

arvidn:
removing peers is deferred already, I suppose adding new peers need to be deferred as well

ajax16384:
some kind of journal of m_connnections changes?

ajax16384:
current quick workaround is to use clone

````c++
	template <typename T, typename IndexType = int>
	aux::vector<T, IndexType> clone(aux::vector<T, IndexType> const& v)
	{
		return v;
	}
...
[...]

PR #2046, opened 2017-05-31
torrent::remove_connection refactor

pavel-pimenov:
add TORRENT_INCREMENT(m_iterating_connections) (issue #2044)

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -159,6 +159,13 @@ namespace libtorrent {
 		, m_state(torrent_status::checking_resume_data)
 	{}
 
+	void torrent_hot_members::remove_connection(peer_connection const* p)
+	{
+		auto const i = sorted_find(m_connections, p);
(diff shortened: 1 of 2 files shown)


src/torrent.cpp
```diff
@@ -159,6 +159,13 @@ namespace libtorrent {
 		, m_state(torrent_status::checking_resume_data)
 	{}
 
+	void torrent_hot_members::remove_connection(peer_connection const* p)
+	{
+		auto const i = sorted_find(m_connections, p);
```
arvidn: this function should assert ``TORRENT_ASSERT(m_iterating_connections == 0);``. In fact, it probably should not be a member of ``torrent_hot_members``, but just ``torrent``, right?

pavel-pimenov:
@arvidn help me fix crash https://github.com/arvidn/libtorrent/issues/2044?

PR #2047, opened 2017-05-31
fixed compilation warnings with newer versions of clang

aldenml:


include/libtorrent/aux_/typed_span.hpp
```diff
@@ -49,6 +49,9 @@ namespace libtorrent { namespace aux {
 
 		// pull in constructors from base class
 		using base::base;
+		// implicit constructor needed for template based member functions
+		typed_span() = default;
+		typed_span(span<T> const& v) : base(v) {} // NOLINT
```
arvidn: The problem with this (afaict) is that the ``typed_span`` now has implicit conversions in both directions to ``span`` and also from ``span``. I believe this defeats some of the purpose of having the extra type in the first place. My first attempt at fixing this was to make this constructor private, since (I believe) this implicit conversion is only necessary from within the member functions themselves. Would that work?

arvidn: also, I think both the default- and copy constructor should be made ``noexcept``

include/libtorrent/upnp.hpp
```diff
@@ -258,11 +258,6 @@ struct TORRENT_EXTRA_EXPORT upnp final
 		// to 0, the mapping is not in use
 		tcp::endpoint local_ep;
 
-		// the external (on the NAT router) port
-		// for the mapping. This is the port we
-		// should announce to others
-		int external_port = 0;
```
arvidn: this was an oversight from the ``base_mapping``-patch, right?

include/libtorrent/http_tracker_connection.hpp
```diff
@@ -84,7 +84,6 @@ namespace libtorrent {
 
 		virtual void on_timeout(error_code const&) {}
 
-		tracker_manager& m_man;
```
arvidn: the base class has this reference too, right? that's why this isn't necessary, is it?

include/libtorrent/aux_/typed_span.hpp
aldenml: I understand you, and this is an interesting edge case of the type system, because it goes to the root of how template member functions are implemented in template based types. I think I will try something else.

include/libtorrent/upnp.hpp
aldenml: yes (detected by the compiler)

include/libtorrent/http_tracker_connection.hpp
aldenml: yes (and detected by the compiler)

aldenml:
@arvidn what do you think now?

arvidn:
thanks! lgtm

PR #2050, opened 2017-06-01
one more fix related to typed_span conversion

aldenml:


src/storage_utils.cpp
```diff
@@ -174,7 +174,8 @@ namespace libtorrent { namespace aux {
 			if (ec) return -1;
 
 			// advance our position in the iovec array and the file offset.
-			current_buf = advance_bufs(current_buf, bytes_transferred);
```
arvidn: would you mind making ``advance_bufs`` just use ``typed_span<iovec_t, int>`` instead? I believe it's only used for that type anyway

aldenml: yes, I saw that option too, didn't do it because it involve more code, let me do that

aldenml:
I think it looks better now, let me know

arvidn:
lgtm

Issue #2041, opened 2017-05-29
Compile error


dbermond:
I'm getting the following error when trying to compile:
```
CXX      chained_buffer.lo
In file included from ../include/libtorrent/sha1_hash.hpp:45:0,
                 from ../include/libtorrent/hasher.hpp:37,
                 from ../include/libtorrent/block_cache.hpp:44,
                 from block_cache.cpp:34:
../include/libtorrent/aux_/typed_span.hpp: In instantiation of ‘libtorrent::aux::typed_span<T, IndexType> libtorrent::aux::typed_span<T, IndexType>::first(libtorrent::aux::typed_span<T, IndexType>::underlying_index) const [with U = int; Cond = void; T = char*; IndexType = int; libtorrent::aux::typed_span<T, IndexType>::underlying_index = int]’:
block_cache.cpp:597:54:   required from here
../include/libtorrent/aux_/typed_span.hpp:75:43: error: could not convert ‘((const libtorrent::aux::typed_span<char*>*)this)->libtorrent::aux::typed_span<char*>::<anonymous>.libtorrent::span<char*>::first(((std::size_t)n))’ from ‘libtorrent::span<char*>’ to ‘libtorrent::aux::typed_span<char*>’
    return this->base::first(std::size_t(n));
                                           ^
make[1]: *** [Makefile:959: block_cache.lo] Error 1
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory '/storage/linux/abs/libtorrent-rasterbar-git/src/libtorrent-rasterbar-git/src'
make: *** [Makefile:617: all-recursive] Error 1
```

System Information:
**OS:** Arch Linux x86_64
**libtorrent-rasterbar:** git master (currently at 1.1.3.r1058.g98a034419)
**compiler**: gcc 7.1.1
**boost**/**boost-libs:** 1.64.0

It was compiling perfectly with gcc 6.3.1 and boost/boost-libs 1.63.
This error showed after the packages update.

arvidn:
I don't understand why that line would require a conversion from ``span`` to ``typed_span``. The conversion that happens there is the opposite.

arvidn:
actually, since ``first()`` returns a ``span``, there shouldn't be any conversion at all

ssiloti:
[...]

PR #2053, opened 2017-06-02
avoid netlink.hpp compilation errors due to lack of comparison operators for sockaddr_nl

aldenml:
See https://github.com/arvidn/libtorrent/issues/2041#issuecomment-305662317

### DIFF
--- a/include/libtorrent/netlink.hpp
+++ b/include/libtorrent/netlink.hpp
@@ -106,6 +106,9 @@ namespace libtorrent {
 			return sizeof(m_sockaddr);
 		}
 
+		// commented the comparison operators for now, until the
+		// same operators are implemented for sockaddr_nl
+		/*
 		friend bool operator==(const basic_nl_endpoint<Protocol>& l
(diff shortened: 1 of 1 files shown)


include/libtorrent/netlink.hpp
```diff
@@ -106,6 +106,9 @@ namespace libtorrent {
 			return sizeof(m_sockaddr);
 		}
 
+		// commented the comparison operators for now, until the
+		// same operators are implemented for sockaddr_nl
+		/*
```
arvidn: I would imagine these operators would need to be implemented as ``std::tie(l.m_sockaddr. X, ...) < std::tie(r.m_sockaddr. X, ...)`` or something like that. assuming we need them at all

arvidn: is ``sockaddr_nl`` a posix type?

aldenml: not sure if it's strictly "POSIX", http://man7.org/linux/man-pages/man7/netlink.7.html

PR #2055, opened 2017-06-03
fix for what appears to be an clang/llvm miscompilation

arvidn:


aldenml:
do you mind commenting what happened? (to learn)

src/storage_utils.cpp
```diff
@@ -63,19 +63,20 @@ namespace libtorrent { namespace aux {
 		return ret;
 	}
 
-	typed_span<iovec_t> advance_bufs(typed_span<iovec_t> bufs, int const bytes)
+	typed_span<iovec_t> advance_bufs(typed_span<iovec_t> const& bufs, int const bytes)
 	{
```
arvidn: when stepping past this line "{", ``&bufs`` changed from being a reasonable pointer to being a pointer to 0xa (which happens to be the value of the``m_len`` field). At first I thought it was some subtle conversion issue, where it ended up copying a temporary that had gone out of scope. But I could not come up with an explanation for that. Stepping through the debugger indicated that it was all good at the time of entering this function.

aldenml: interesting

ssiloti: I would sooner believe there's bug in the debugger than a bug in the compiler, especially if you're debugging optimized code. I would take a look at the assembly output to be sure of what's going on.

arvidn: ok, maybeI can get some help to look at it. I will post some more details of what I see in the main comments

arvidn:
This is my notes/transcript from digging into this error. I feel more confident that this is in fact a compiler bug. If you're impatient and just want to see my conclusion, search for "ERROR:".

https://pastebin.com/MDMHpbQP

It appears that this is an issue where a span<> object is split up and passed as in two registers (one for the pointer and one for length), and the function body then needs to call a member function on this object, but it fails to correctly construct a complete object on the stack, to which it can create a this-pointer for the call.

The function prolog does in fact emit a few extra instructions and use one more stack slot in order to form a valid span<> object on the stack, but the member function call appears to pick the wrong address off the stack, and passes the length field instead of the this pointer.

arvidn:
```
$ clang++ --version
Apple LLVM version 8.1.0 (clang-802.0.42)
Target: x86_64-apple-darwin16.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

ssiloti:
My assembly experience is mainly with ARM, but I'll give it a shot. In the first snippet you note that you expected `iov` to be at `-0xa90(%rbp)` but it wasn't. Doesn't that `movq` instruction load the value stored at `-0xa90(%rbp)` into `%rdi` rather than the address itself? So the value at `-0xa90(%rbp)` should be the address of `iov` shouldn't it?

ssiloti:
Yeah, looks like a compiler bug, but I think things may have gone wrong even earlier. At line 211 you comment on the layout of the variables on the stack. In fact they are ordered as (len, ptr, ptr) so there is no valid span object there.

arvidn:
you're right. I interpreted the first ``movq`` as a ``leaq``. I wasn't expecting the pointer and length fields to be passed in separate registers until later. But that's what it's doing there, pulling out the two fields into ``rdi`` and ``rsi``. It's not obvious to me though where the this-pointer comes from in the first call to the span<> constructor.

PR #2057, opened 2017-06-05
fix import error in python bindings

ssiloti:
TypeError: No to_python (by-value) converter found for C++ type:
libtorrent::deprecated_move_flags_t


arvidn:
interesting. how is this enum referenced? I would have expected that any references would just (silently) start referring to the new enum values, no?

ssiloti:
For example, it's referenced in the definition of `move_storage1` in bindings/python/src/torrent_handle.cpp.

PR #2058, opened 2017-06-05
refactor static asserts related to nothrow in macros

aldenml:
The main motivation of this change is that SWIG is having a hard time with `static_assert` at namespace level. This allows me to pass empty macro definitions and avoid the problem. But I think overall it is a good DRY refactor.

### DIFF
--- a/include/libtorrent/add_torrent_params.hpp
+++ b/include/libtorrent/add_torrent_params.hpp
@@ -525,18 +525,13 @@ namespace libtorrent {
 
 	};
 
-	static_assert(std::is_nothrow_move_constructible<add_torrent_params>::value
-		, "should be nothrow move constructible");
-
+	TORRENT_ASSERT_NOTHROW_MOVE_CONSTRUCTIBLE(add_torrent_params);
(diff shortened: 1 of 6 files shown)


arvidn:
hm.. I think avoiding macros may rank higher than some repetition (especially since there's only 1 line less). Also given that a lot of the repetition will be fixed with c++17 (and static_assert without message).

Do you think there's some other place the static asserts could be moved to, to not be at the namespace level? I suppose if they were moved into a .cpp file swig wouldn't even see them.

aldenml:
agree, and I can move them to the cpp, only problem is `add_torrent_params` that no have `.cpp`. If you are ok with adding one more cpp, tell me and I do that. Also, if you are ok, I can move all `add_torrent_params ` constructors to the cpp.

Other option is move the asserts to inside one of the constructors (it works because the type is complete there).

What do you think?

arvidn:
I think adding another .cpp file for ``add_torrent_params`` seems reasonable, along with moving some of the non-trivial member functions into there

aldenml:
Closed in favor of https://github.com/arvidn/libtorrent/pull/2059

PR #2059, opened 2017-06-06
added add_torrent_params.cpp and moved is_nothrow asserts to cpp files

aldenml:


aldenml:
@arvidn do you mind taking a look at the error with `gcc`? I think it's the same problem with `noexcept` that you commented in other places, but it's working with `msvc`

arvidn:
hm. the fact that GCC fails to compile this but succeeded with the previous form definitely seem to suggest it's buggy to some extent. As far as I understand, this error means that when GCC generates the default move assignment operator, by calling ``operator=(T&&)`` on all members, there's at least one that is not ``noexcept``.

actually, ``std::string`` is not nothrow-move assignable in GCC (despite C++11 requiring it). It's possible this type also can only be nothrow move constructible.

aldenml:
and `mingw-gcc` pass :) well, I'm removing the `noexcept` on that operator.

aldenml:
if you have a minute, appreciate if you restart the macOS job

arvidn:
I don't understand why ``aafigure`` has stopped working..

aldenml:
sometimes is the travis cache, if it fails again, maybe clean it up could help

PR #2060, opened 2017-06-06
fix bandwith rate limit calculation

ajax16384:
c9a2fed2c95ff4d42d216d4b19248ca751e7edd6 introduces regression:
any *rate_limit value above 4294966 with "tick_count"==500 leads to wrong quota calculation (unable to set upload/dowload limit e.t.c).

### DIFF
--- a/include/libtorrent/bandwidth_limit.hpp
+++ b/include/libtorrent/bandwidth_limit.hpp
@@ -43,7 +43,7 @@ namespace libtorrent {
 // member of peer_connection
 struct TORRENT_EXTRA_EXPORT bandwidth_channel
 {
-	static constexpr int inf = std::numeric_limits<int>::max();
+	static constexpr int inf = std::numeric_limits<std::int32_t>::max();
 
 	bandwidth_channel();
 
@@ -93,7 +93,7 @@ private:
 
 	// the limit is the number of bytes
 	// per second we are allowed to use.
-	int m_limit;
+	std::int32_t m_limit;
 };
 
 }
--- a/include/libtorrent/bandwidth_limit.hpp
+++ b/include/libtorrent/bandwidth_limit.hpp
@@ -51,8 +51,9 @@ struct TORRENT_EXTRA_EXPORT bandwidth_channel
 	void throttle(int limit);
 	int throttle() const
 	{
+		TORRENT_ASSERT_VAL(m_limit >= 0, m_limit);
 		TORRENT_ASSERT_VAL(m_limit < inf, m_limit);
-		return int(m_limit);
+		return m_limit;
 	}
 
 	int quota_left() const;
@@ -92,7 +93,7 @@ private:
 
 	// the limit is the number of bytes
 	// per second we are allowed to use.
-	std::int64_t m_limit;
+	int m_limit;
 };
 
 }
--- a/src/bandwidth_limit.cpp
+++ b/src/bandwidth_limit.cpp
@@ -61,16 +61,12 @@ namespace libtorrent {
 	{
 		if (m_limit == 0) return;
 
+		m_quota_left += (m_limit * dt_milliseconds + 500) / 1000;
+		if (m_quota_left / 3 > m_limit) m_quota_left = m_limit * 3;
+
 		// avoid integer overflow
-		if (m_limit >= (std::int64_t(std::numeric_limits<int>::max()) * 1000) / dt_milliseconds)
-		{
-			m_quota_left = std::numeric_limits<int>::max();
-		}
-		else
-		{
-			m_quota_left += (m_limit * dt_milliseconds + 500) / 1000;
-			if (m_quota_left / 3 > m_limit) m_quota_left = m_limit * 3;
-		}
+		m_quota_left = std::min(m_quota_left, std::int64_t(std::numeric_limits<int>::max()));
+
 		distribute_quota = int(std::max(m_quota_left, std::int64_t(0)));
 	}
 
--- a/src/bandwidth_limit.cpp
+++ b/src/bandwidth_limit.cpp
@@ -65,7 +65,7 @@ namespace libtorrent {
 		if (m_limit == 0) return;
 
 		// "to_add" should never have int64 overflow: "m_limit" contains < "<int>::max"
-		std::int64_t to_add = (m_limit * dt_milliseconds + 500) / 1000;
+		std::int64_t to_add = (std::int64_t(m_limit) * dt_milliseconds + 500) / 1000;
 
 		if (to_add > inf - m_quota_left)
 		{
@@ -74,7 +74,7 @@ namespace libtorrent {
 		else
 		{
 			m_quota_left += to_add;
-			if (m_quota_left / 3 > m_limit) m_quota_left = m_limit * 3;
+			if (m_quota_left / 3 > m_limit) m_quota_left = std::int64_t(m_limit) * 3;
 			// "m_quota_left" will never have int64 overflow but may exceed "<int>::max"
 			m_quota_left = std::min(m_quota_left, std::int64_t(inf));
 		}
--- a/src/bandwidth_limit.cpp
+++ b/src/bandwidth_limit.cpp
@@ -61,7 +61,7 @@ namespace libtorrent {
 	{
 		TORRENT_ASSERT_VAL(m_limit >= 0, m_limit);
 
-		TORRENT_ASSERT_VAL(m_limit < inf, m_limit);
+		TORRENT_ASSERT_VAL(m_limit <= inf, m_limit);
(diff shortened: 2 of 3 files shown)


src/bandwidth_limit.cpp
```diff
@@ -62,7 +62,7 @@ namespace libtorrent {
 		if (m_limit == 0) return;
 
 		// avoid integer overflow
-		if (m_limit >= std::numeric_limits<int>::max() / dt_milliseconds)
+		if (m_limit >= (std::int64_t(std::numeric_limits<int>::max()) * 1000) / dt_milliseconds)
```
arvidn: if you multiply by 1000 after the division, you don't need to extend to 64bits, I imagine the loss of precision is negligible, no?
Also, I'm not sure this is quite right. the intermediate value of ``m_limit`` * ``dt_milliseconds`` may still overflow, before dividing by 1000.

src/bandwidth_limit.cpp
```diff
@@ -62,7 +62,7 @@ namespace libtorrent {
...
 		{
 			m_quota_left = std::numeric_limits<int>::max();
```
arvidn: think a better fix would be ti divide this by 1000

arvidn:
do you think it would be a lot of work to extend the test introduced for the original patch to catch this bug?

arvidn:
In case you missed this too: https://travis-ci.org/arvidn/libtorrent/jobs/240054652#L1155

src/bandwidth_limit.cpp
ajax16384: i'll try more robust variant - do check after all calculation

src/bandwidth_limit.cpp
```diff
@@ -61,16 +61,12 @@ namespace libtorrent {
...
-		if (m_limit >= std::numeric_limits<int>::max() / dt_milliseconds)
-		{
-			m_quota_left = std::numeric_limits<int>::max();
-		}
-		else
-		{
-			m_quota_left += (m_limit * dt_milliseconds + 500) / 1000;
-			if (m_quota_left / 3 > m_limit) m_quota_left = m_limit * 3;
-		}
+		m_quota_left = std::min(m_quota_left, std::int64_t(std::numeric_limits<int>::max()));
```
arvidn: I'm not sure how you intended this to work. but the overflow would happen on line 64, right? It doesn't matter what you do *after* undefined behavior is triggered, you need to avoid it from happening in the first place

ajax16384: ok

ajax16384:
and added regression test case to test_bandwidth_limiter.cpp

src/bandwidth_limit.cpp
```diff
@@ -59,18 +59,27 @@ namespace libtorrent {
...
+		TORRENT_ASSERT_VAL(m_limit >= 0, m_limit);
+
+		TORRENT_ASSERT_VAL(m_limit < inf, m_limit);
+
 		if (m_limit == 0) return;
 
-		// avoid integer overflow
-		if (m_limit >= std::numeric_limits<int>::max() / dt_milliseconds)
+		// "to_add" should never have int64 overflow: "m_limit" contains < "<int>::max"
+		std::int64_t to_add = (m_limit * dt_milliseconds + 500) / 1000;
```
arvidn: I don't think it's necessary to widen to 64 bits here. you seem to prefer that, is there a compelling reason?
also, the multiplication is still performed in 32 bits, you need to widen one of the operands before multiplying if you want it to be 64 bits.

src/bandwidth_limit.cpp
```diff
@@ -59,18 +59,27 @@ namespace libtorrent {
 
 	void bandwidth_channel::update_quota(int dt_milliseconds)
 	{
+		TORRENT_ASSERT_VAL(m_limit >= 0, m_limit);
+
+		TORRENT_ASSERT_VAL(m_limit < inf, m_limit);
```
arvidn: ``m_limit`` can be == ``inf``, right?

src/bandwidth_limit.cpp
```diff
@@ -59,18 +59,27 @@ namespace libtorrent {
...
+
 		if (m_limit == 0) return;
 
-		// avoid integer overflow
-		if (m_limit >= std::numeric_limits<int>::max() / dt_milliseconds)
+		// "to_add" should never have int64 overflow: "m_limit" contains < "<int>::max"
+		std::int64_t to_add = (m_limit * dt_milliseconds + 500) / 1000;
+
+		if (to_add > inf)
```
arvidn: actually, this test would have to be ``to_add - m_quota_left > inf``, right?

src/bandwidth_limit.cpp
ajax16384: m_limit was already defined as int64_t so this calculation was always made with 64bit variables

src/bandwidth_limit.cpp
ajax16384: ok

src/bandwidth_limit.cpp
ajax16384: i meant "to_add + m_quota_left > inf" - if so then it will be covered at "m_quota_left = std::min(m_quota_left, std::int64_t(inf));"

arvidn: heh, actually I meant ``to_add > inf - m_quota_left``

arvidn: to avoid an overflow in the addition on line 77

src/bandwidth_limit.cpp
arvidn: I see. would you mind making this ``const`` btw?

arvidn: actually, it seems weird that ``m_limit`` would be a 64 bit int, if it can only hold values <= ``INT32_MAX``.. it seems like it would make more sense to widen these calculations instead..

src/bandwidth_limit.cpp
ajax16384: it's ok

src/bandwidth_limit.cpp
ajax16384: rest asserts checks "limit < inf" so i revert "<=" checking

src/bandwidth_limit.cpp
ajax16384: so i can change m_limit to int and wide to int64 only this calculation?

src/bandwidth_limit.cpp
```diff
@@ -59,18 +59,26 @@ namespace libtorrent {
 
 	void bandwidth_channel::update_quota(int dt_milliseconds)
 	{
+		TORRENT_ASSERT_VAL(m_limit >= 0, m_limit);
+		TORRENT_ASSERT_VAL(m_limit < inf, m_limit);
```
arvidn: should not ``INT32_MAX`` be allowed?

include/libtorrent/bandwidth_limit.hpp
```diff
@@ -92,7 +93,7 @@ struct TORRENT_EXTRA_EXPORT bandwidth_channel
 
 	// the limit is the number of bytes
 	// per second we are allowed to use.
-	std::int64_t m_limit;
+	int m_limit;
```
arvidn: it may even make sense to say ``std::int32_t``

src/bandwidth_limit.cpp
```diff
@@ -59,18 +59,26 @@ namespace libtorrent {
...
+			m_quota_left = inf;
 		}
 		else
 		{
-			m_quota_left += (m_limit * dt_milliseconds + 500) / 1000;
-			if (m_quota_left / 3 > m_limit) m_quota_left = m_limit * 3;
+			m_quota_left += to_add;
+			if (m_quota_left / 3 > m_limit) m_quota_left = std::int64_t(m_limit) * 3;
+			// "m_quota_left" will never have int64 overflow but may exceed "<int>::max"
+			m_quota_left = std::min(m_quota_left, std::int64_t(inf));
```
arvidn: I don't understand this line

arvidn:
looks reasonable

src/bandwidth_limit.cpp
ajax16384: original has several asserts checking for "< inf"

src/bandwidth_limit.cpp
ajax16384: this will prevent m_quota_left exceed "inf" ( it definetly not overflow int64 but may be greater than "inf")

ajax16384:
BTW RC_1_1 contains same regression: https://github.com/arvidn/libtorrent/blob/RC_1_1/src/bandwidth_limit.cpp#L65

PR #2063, opened 2017-06-08
fix operator() ==, add const

pavel-pimenov:


include/libtorrent/peer_connection_handle.hpp
```diff
@@ -110,7 +110,7 @@ struct TORRENT_EXPORT peer_connection_handle
 	bool operator==(peer_connection_handle const& o) const
 	{ return !lt(m_connection, o.m_connection) && !lt(o.m_connection, m_connection); }
 	bool operator!=(peer_connection_handle const& o) const
-	{ return lt(m_connection, o.m_connection) || lt(o.m_connection, m_connection); }
```
arvidn: it's not obvious that this change improves anything, is it? If the compiler fails to inline all of this, it will probably make more calls to ``owner_before`` for instance

include/libtorrent/sha1_hash.hpp
```diff
@@ -161,7 +161,7 @@ namespace aux {
 		}
 		bool operator!=(digest32 const& n) const
 		{
-			return !std::equal(n.m_number, n.m_number + number_size, m_number);
```
arvidn: I'm not necessarily sold on this transformation being an improvement. could you argue for it?

pavel-pimenov: http://www.learncpp.com/cpp-tutorial/96-overloading-the-comparison-operators/
http://courses.cms.caltech.edu/cs11/material/cpp/donnie/cpp-ops.html 

revert all !(*this == x); ?

pavel-pimenov: hm... C++ Core Guidelines:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c86-make--symmetric-with-respect-to-operand-types-and-noexcept
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c161-use-nonmember-functions-for-symmetric-operators

pavel-pimenov:
CI error..
https://travis-ci.org/arvidn/libtorrent/jobs/240713485#L759

include/libtorrent/sha1_hash.hpp
arvidn: I'm on-board with making the operators non-members, and to make them all ``noexcept``.
Specifically what I'm questioning is to always express ``!=`` in terms of ``==``. Obviously the semantics should be the opposite, but you're not changing the semantics, just how it's expressed.

PR #2068, opened 2017-06-13
defer reconnecting peers to after the second_tick loop

arvidn:
to avoid mutating m_connections while iterating over it


arvidn:
perhaps this should be back-ported to RC_1_1

PR #2088, opened 2017-06-18
handle invalid arguments to set_piece_deadline()

arvidn:


src/torrent.cpp
```diff
@@ -5235,7 +5235,13 @@ namespace libtorrent
...
 		INVARIANT_CHECK;
 
-		if (m_abort)
+		TORRENT_ASSERT_PRECOND(piece >= 0);
+		TORRENT_ASSERT_PRECOND(valid_metadata());
+		TORRENT_ASSERT_PRECOND(valid_metadata() && piece < m_torrent_file->num_files());
+
+		if (m_abort || !valid_metadata()
+			|| piece < 0
+			|| piece >= m_torrent_file->num_files())
```
ssiloti: Should be `num_pieces` instead of `num_files`. Also in the assert above.

arvidn: doh! at least I would have discovered it in master, where those are different units!

src/torrent.cpp
```diff
@@ -5235,7 +5235,13 @@ namespace libtorrent
 	{
 		INVARIANT_CHECK;
 
-		if (m_abort)
+		TORRENT_ASSERT_PRECOND(piece >= 0);
+		TORRENT_ASSERT_PRECOND(valid_metadata());
+		TORRENT_ASSERT_PRECOND(valid_metadata() && piece < m_torrent_file->num_files());
```
ssiloti: Typo here too.

PR #2093, opened 2017-06-20
added const and auto refactor in torrent, piece_picker

aldenml:
The context of this change is that I'm looking for a potential bug in `torrent::set_piece_deadline`

### DIFF
--- a/src/piece_picker.cpp
+++ b/src/piece_picker.cpp
@@ -979,7 +979,7 @@ namespace libtorrent {
 		}
 	}
 
-	void piece_picker::shuffle(int priority, prio_index_t elem_index)
+	void piece_picker::shuffle(int const priority, prio_index_t const elem_index)
 	{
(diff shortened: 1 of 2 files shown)


src/piece_picker.cpp
```diff
@@ -2774,26 +2774,26 @@ namespace {
 				TORRENT_ASSERT(find_dl_piece(i, index) == m_downloads[i].end());
 			return false;
 		}
-		std::vector<downloading_piece>::const_iterator i = find_dl_piece(state, index);
+		auto const i = find_dl_piece(state, index);
 		TORRENT_ASSERT(i != m_downloads[state].end());
-		return i->passed_hash_check;
+		return bool(i->passed_hash_check);
```
arvidn: this is a bitfield, right? that's why it's an integer instead of bool

aldenml: yes

arvidn:
lgtm!

PR #2095, opened 2017-06-20
added consts and minor auto refactor in upnp.cpp

aldenml:


include/libtorrent/upnp.hpp
```diff
@@ -184,7 +184,7 @@ struct TORRENT_EXTRA_EXPORT upnp final
 	void discover_device();
 	void close();
 
-	// This is only available for UPnP routers. If the model is advertized by
+	// This is only available for UPnP routers. If the model is advertised by
```
arvidn: I added hunspell to RC_1_1 :) I will soon merged it into master

src/upnp.cpp
```diff
@@ -1079,8 +1077,7 @@ void upnp::disable(error_code const& ec)
 	m_disabled = true;
 
 	// kill all mappings
-	for (std::vector<global_mapping_t>::iterator i = m_mappings.begin()
-		, end(m_mappings.end()); i != end; ++i)
+	for (auto i = m_mappings.begin(), end(m_mappings.end()); i != end; ++i)
```
arvidn: this could probably even be a range-for loop

arvidn:
lgtm

src/upnp.cpp
aldenml: this is a few lines below `int(i - m_mappings.begin()`

PR #2101, opened 2017-06-24
invert logic in read_nl_sock for better clarity and reliability

ssiloti:
The seq and pid parameter are apparently used as a safety check in case
a stream of netlink messages is not properly terminated. I initially thought
that they represented the expected values of the incoming messages.
Instead seq is set to one-past the expected value and the loop aborts when
seq and pid match the message.

Inverting the logic so that reading continues as long as the seq and pid match
the incoming messages is, to me at least, more intuitive. It is also more
reliable since it does not rely on the next seq being strictly sequential. It
also catches unexpected messages preceding or interleaved with the expected
messages.

### DIFF
--- a/src/enum_net.cpp
+++ b/src/enum_net.cpp
@@ -160,13 +160,13 @@ namespace libtorrent {namespace {
 
 #if TORRENT_USE_NETLINK
 
-	int read_nl_sock(int sock, char *buf, int bufsize, std::uint32_t const seq, int const pid)
+	int read_nl_sock(int sock, char *buf, int bufsize, std::uint32_t const seq, std::uint32_t const pid)
 	{
 		nlmsghdr* nl_hdr;
 
 		int msg_len = 0;
 
-		do
+		for (;;)
 		{
 			int read_len = int(recv(sock, buf, std::size_t(bufsize - msg_len), 0));
 			if (read_len < 0) return -1;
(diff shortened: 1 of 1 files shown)


aldenml:
I hope the error is not related to the new `trusty` images in travis

src/enum_net.cpp
```diff
@@ -1052,7 +1052,7 @@ namespace libtorrent {
 		nl_msg->nlmsg_len = NLMSG_LENGTH(sizeof(rtmsg));
 		nl_msg->nlmsg_type = RTM_GETROUTE;
 		nl_msg->nlmsg_flags = NLM_F_DUMP | NLM_F_REQUEST;
-		nl_msg->nlmsg_seq = seq++;
+		nl_msg->nlmsg_seq = seq;
```
arvidn: is ``seq`` a constant now? or is it passed in by reference to ``read_nl_sock``?
if it is, would you mind making it ``const``?
I would also appreciate if you could add some comments about what you've learned about what ``seq`` and ``pid`` do

arvidn:
I restarted the travis job, in case it's ephemeral. but the failure definitely does not look like it has anything to do with this patch

src/enum_net.cpp
ssiloti: I went to double check my info on `nlmsg_pid` and went down a bit of a rabbit hole. I'll add another commit with some more changes and my findings.

PR #2103, opened 2017-06-24
refactoring stack_frame

pavel-pimenov:
Visual C++ 2017 release build:
default

### DIFF
--- a/include/libtorrent/bdecode.hpp
+++ b/include/libtorrent/bdecode.hpp
@@ -166,7 +166,7 @@ struct bdecode_token
 	enum type_t
 	{ none, dict, list, string, integer, end };
 
-	enum limits_t
+	enum limits_t : std::uint32_t
 	{
 		max_offset = (1 << 29) - 1,
 		max_next_item = (1 << 29) - 1,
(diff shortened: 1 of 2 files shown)


include/libtorrent/bdecode.hpp
```diff
@@ -166,7 +166,7 @@ struct bdecode_token
 	enum type_t
 	{ none, dict, list, string, integer, end };
 
-	enum limits_t
+	enum limits_t : std::uint32_t
```
arvidn: since these are quantaties, meant for arithmetic, I don't think they should be unsigned. see this [blog post](http://blog.libtorrent.org/2016/05/unsigned-integers/).

Also, this is a bit of an abuse of enums. Ideally these constants should be ``constexpr`` constants.

src/bdecode.cpp
```diff
@@ -102,13 +102,25 @@ namespace {
 
 	struct stack_frame
 	{
-		stack_frame() : token(0), state(0) {}
-		explicit stack_frame(int const t): token(std::uint32_t(t)), state(0) {}
+		stack_frame() : m_token_state(0) {}
```
arvidn: what's the motivation for this change? it seems to increase complexity, and I don't know what it buys

PR #2105, opened 2017-06-26
Use netlink to enumerate network interfaces on linux

ssiloti:
The motivation for this is to be able to set the `preferred` field based on `ifa_flags`.


src/enum_net.cpp
```diff
@@ -199,6 +201,38 @@ namespace libtorrent {namespace {
 		return msg_len;
 	}
 
+	enum { NL_BUFSIZE = 8192 };
+
+	int nl_dump_request(int sock, int type, std::uint32_t seq, char family, char msg[NL_BUFSIZE], std::size_t msg_len)
```
arvidn: I'm pretty sure ``msg`` will decay to just a pointer. Would it make sense to use a span<char> here instead?

arvidn:
lgtm

PR #2106, opened 2017-06-27
broadcast_socket: remove obsolete special case

ssiloti:
`enum_net_interfaces` now fills in `scope_id` correctly so there's no need for
a special case to set it here.

### DIFF
--- a/src/broadcast_socket.cpp
+++ b/src/broadcast_socket.cpp
@@ -194,25 +194,6 @@ namespace libtorrent {
 
 			ec = error_code();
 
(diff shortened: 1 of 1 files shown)


arvidn:
does this depend on https://github.com/arvidn/libtorrent/pull/2105 ? or is this redundant on other systems than linux too now?

ssiloti:
This is not dependent on #2105. It's been redundant on all platforms ever since `sockaddr_to_address` started setting the scope_id.

PR #2108, opened 2017-06-29
fix backwards compatibility issue

arvidn:
when loading the torrent info dict from resume data

### DIFF
--- a/src/session_handle.cpp
+++ b/src/session_handle.cpp
@@ -201,121 +201,128 @@ namespace libtorrent {
 #ifndef TORRENT_NO_DEPRECATE
 namespace {
 
-		void handle_backwards_compatible_resume_data(add_torrent_params& atp)
+	void handle_backwards_compatible_resume_data(add_torrent_params& atp)
+	{
+		// if there's no resume data set, there's nothing to do. It's either
+		// using the previous API without resume data, or the resume data has
+		// already been parsed out into the add_torrent_params struct.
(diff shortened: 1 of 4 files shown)


src/session_handle.cpp
```diff
@@ -201,121 +201,128 @@ namespace libtorrent {
...
-				atp.trackers.insert(atp.trackers.end()
-					, resume_data.trackers.begin()
-					, resume_data.trackers.end());
-				atp.tracker_tiers.insert(atp.tracker_tiers.end()
-					, resume_data.tracker_tiers.begin()
-					, resume_data.tracker_tiers.end());
-				if ((resume_data.flags & add_torrent_params::flag_merge_resume_trackers) == 0)
-					atp.flags |= add_torrent_params::flag_override_trackers;
-			}
+		if (!atp.ti)
```
arvidn: this is the only material change (the rest of this hunk is just reducing the indentation)

test/test_resume.cpp
```diff
@@ -751,19 +751,80 @@ TORRENT_TEST(zero_file_prio_deprecated)
...
+namespace libtorrent {
+namespace flags {
+
+template <>
+struct enable_flag_operators<test_mode_t> : std::true_type {};
+
+}
+}
+
+void test_seed_mode(test_mode_t const flags)
```
arvidn: tests should be treated as production code, and so, fixed the signature to not use raw bools.

PR #2109, opened 2017-06-29
fixed compilation warnings in enum_net netlink code

aldenml:


src/enum_net.cpp
```diff
@@ -225,7 +225,7 @@ namespace libtorrent {namespace {
 		// get the socket's port ID so that we can verify it in the repsonse
 		sockaddr_nl sock_addr;
 		socklen_t sock_addr_len = sizeof(sock_addr);
-		if (getsockname(sock, (sockaddr*)&sock_addr, &sock_addr_len) < 0)
+		if (getsockname(sock, reinterpret_cast<sockaddr*>(&sock_addr), &sock_addr_len) < 0)
```
arvidn: I thought these would be errors on travis, aren't they?

aldenml: no, because they are detected by `clang`, but in the `Xcode` job the macro for `TORRENT_USE_NETLINK` is `0`. I detect those while compiling for android.

PR #2115, opened 2017-06-30
added block_uploaded_alert for upload

toinetoine:
Ability for client to track [upload activity](https://stackoverflow.com/a/13938080/1053362).


include/libtorrent/alert_types.hpp
```diff
@@ -2663,11 +2663,27 @@ namespace libtorrent {
...
+			, tcp::endpoint const& ep, peer_id const& peer_id, int block_num
+			, int piece_num);
+
+		TORRENT_DEFINE_ALERT(block_uploaded_alert, 94)
+
+		static const int static_category = alert::progress_notification;
+		virtual std::string message() const override;
+
+		int block_index;
+		int piece_index;
```
arvidn: please use ``piece_index_t`` as the type for ``piece_index``

include/libtorrent/alert_types.hpp
```diff
@@ -2663,11 +2663,27 @@ namespace libtorrent {
 		aux::allocation_slot m_v6_nodes_idx;
 	};
 
+	struct TORRENT_EXPORT block_uploaded_alert final : peer_alert
```
arvidn: Please document this alert (using comments on the lines above the declaration). It's worth mentioning that it's posted when a block is placed in the send buffer to a peer, which may or may not have been sent yet (and if the connection is closed, the alert may be posted without the bytes having been sent to the peer). It's also probably worth mentioning which category it belongs to.

arvidn:
I think this change warrants a line in the ``ChangeLog`` as well

include/libtorrent/alert_types.hpp
arvidn: when deprecated functions are disabled, the type-safety is enforced, and ``piece_index_t`` does not implicitly convert to ``int`` anymore. https://ci.appveyor.com/project/arvidn/libtorrent/build/6561/job/9vcf5vjw7ubjmntq#L204

src/alert.cpp
```diff
@@ -2414,4 +2414,22 @@ namespace {
 			, m_v6_num_nodes, m_v6_nodes_idx);
 	}
 
+	block_uploaded_alert::block_uploaded_alert(aux::stack_allocator& alloc, torrent_handle h
+		, tcp::endpoint const& ep, peer_id const& peer_id, int block_num
+		, int piece_num)
```
arvidn: this is still an ``int`` instead of ``piece_index_t``

src/alert.cpp
```diff
@@ -2414,4 +2414,22 @@ namespace {
...
+		: peer_alert(alloc, h, ep, peer_id)
+		, block_index(block_num)
+		, piece_index(piece_num)
+	{
+		TORRENT_ASSERT(block_index >= 0 && piece_index >= 0);
```
arvidn: the second comparison here need to be against ``piece_index_t{0}`` instead of just 0.
If you build with ``TORRENT_NO_DEPRECATE`` defined, this should be an error for you.

src/alert.cpp
```diff
@@ -2414,4 +2414,22 @@ namespace {
...
+	}
+
+	std::string block_uploaded_alert::message() const
+	{
+		char ret[200];
+		snprintf(ret, sizeof(ret), "%s block uploaded to a peer (piece: %u block: %u)"
+			, torrent_alert::message().c_str(), piece_index, block_index);
```
arvidn: to get the ``int`` out of the ``piece_index_t``, use ``static_cast<int>(piece_index)``

arvidn:
did you see this?
test_alert_types.cpp:169 "TEST_ERROR: num_alert_types: 95 expected: 94

include/libtorrent/alert_types.hpp
```diff
@@ -2663,11 +2663,31 @@ namespace libtorrent {
...
+			, tcp::endpoint const& ep, peer_id const& peer_id, int block_num
+			, piece_index_t piece_num);
+
+		TORRENT_DEFINE_ALERT(block_uploaded_alert, 94)
+
+		static const int static_category = alert::progress_notification;
+		virtual std::string message() const override;
+
+        int const block_index;
+        piece_index_t const piece_index;
```
arvidn: sorry, I just noticed this. Could you indent these two lines with tabs, like the other ones?

include/libtorrent/alert_types.hpp
```diff
@@ -2663,11 +2663,31 @@ namespace libtorrent {
 		aux::allocation_slot m_v6_nodes_idx;
 	};
 
+    // This alert is posted when a block intended to be sent to a peer is placed in the
+    // send buffer. Note that if the connection is closed before the send buffer is sent,
+    // the alert may be posted without the bytes having been sent to the peer.
+    // It belongs to the ``progress_notification`` category.
```
arvidn: you use spaces instead of tabs here too

arvidn:
I think the merge may be screwing with githubs ability to rebase this on top of master. In general I think it's better to rebase patches instead of merging master into them

ChangeLog
```diff
@@ -1,4 +1,4 @@
-	* drop support for windows compilers without std::wstring
```
arvidn: This change is no longer correct. I would suggest you squash and rebase your commits on top of master, and while doing so, drop both of the merge commits

PR #2118, opened 2017-07-01
reject put messages with incorrect bencoding

ssiloti:


test/test_bdecode.cpp
```diff
@@ -206,6 +209,69 @@ TORRENT_TEST(dict_null_key)
...
+TORRENT_TEST(dict_unordered_keys)
+{
+	char error_string[200];
+	{
+		char b[] = "d2:abi1e2:aai2ee";
+		bdecode_node e;
+		error_code ec;
+		int ret = bdecode(b, b + sizeof(b) - 1, e, ec);
+		TEST_EQUAL(ret, 0);
+		TEST_CHECK(e.has_soft_error(error_string));
```
arvidn: I'm not sure about returning an error message via a span. Is there a benefit over returning an error_code?

src/kademlia/node.cpp
```diff
@@ -966,8 +966,12 @@ void node::incoming_request(msg const& m, entry& e)
 		};
 
 		// attempt to parse the message
+		// also reject the message if it has any non-fatal encoding errors
+		// because put messages contain a signed value they must have correct bencoding
+		// otherwise re-encoding the value will break the signature
```
arvidn: It feels a bit wasteful to do a second pass over the buffer to detect soft errors. Could these issues be discovered during the main parsing pass?
I think referring to the problem as "not round-tripping" helps a reader understand it quicker too

ssiloti: I can certainly do the checking as part of `bdecode`. In fact my original version worked that way. I moved the checking to a separate function so that the added cost would only be incurred when necessary.

arvidn: yeah, perhaps it does make sense to keep the main decode path as fast as possible.

arvidn: Do we define a "canonical encoding" in the BEP? (if not, perhaps that would make sense). The leading zeroes hadn't occurred to me before

ssiloti: The description in BEP 3 is pretty terse, but it does explicitly call out leading zeros as an error.

src/bdecode.cpp
```diff
@@ -289,6 +289,93 @@ namespace {
...
+						// skip to the first key's value
+						std::uint32_t const v1 = k1 + tokens[k1].next_item;
+						// then to the next key
+						std::uint32_t const k2 = v1 + tokens[v1].next_item;
+
+						// check if k1 was the last key in the dict
+						if (k2 == token)
+							break;
+
+						std::uint32_t const v2 = k2 + tokens[k2].next_item;
```
arvidn: in general I like signed integers for arithmetic. Is there a good reason to use unsigned here?

src/bdecode.cpp
```diff
@@ -289,6 +289,93 @@ namespace {
...
+
+						std::uint32_t const k1_start = tokens[k1].offset + std::uint32_t(tokens[k1].start_offset());
+						std::uint32_t const k1_len = tokens[v1].offset - k1_start;
+						std::uint32_t const k2_start = tokens[k2].offset + std::uint32_t(tokens[k2].start_offset());
+						std::uint32_t const k2_len = tokens[v2].offset - k2_start;
+
+						std::uint32_t const min_len = std::min(k1_len, k2_len);
+
+						int cmp = std::memcmp(m_buffer + k1_start, m_buffer + k2_start, min_len);
```
arvidn: I have a feeling that this test would be simpler using span's ``operator==``

arvidn: oh, I see, this is enforcing the lexicographical ordering of keys, so you actually compare greater than

src/bdecode.cpp
ssiloti: Not really, it seemed like the nature choice since the source is unsigned. I'll change all these locals to ints.

src/bdecode.cpp
```diff
@@ -289,6 +289,93 @@ namespace {
...
+
+		// we don't know what the original depth_limit was
+		// so this has to go on the heap
+		std::vector<int> stack;
+		// make the initial allocation the default depth_limit
+		stack.reserve(100);
+
+		do
+		{
+			switch (tokens[token].type)
```
arvidn: this just occurred to me; is it possible to have 0 tokens? i.e. have this be an out-of-bound access?

src/bdecode.cpp
```diff
@@ -289,6 +289,93 @@ namespace {
...
+					std::snprintf(error.data(), error.size(), "leading zero in string length");
+					return true;
+				}
+				break;
+			case bdecode_token::dict:
+			case bdecode_token::list:
+				stack.push_back(token);
+				break;
+			case bdecode_token::end:
+				auto const parent = stack.back();
```
arvidn: hm, I'm surprised it's OK to declare a new variable here, without enclosing braces.

src/bdecode.cpp
arvidn: I think this function should have an early exit for the case of the ``bdecode_node`` object being default constructed (or in the state after ``clear()`` is called).

When the node is not the root, ``m_tokens`` will be empty, so it's not is simple as just checking that. Perhaps a check that ``m_buffer`` is set? or perhaps ``m_token_idx`` >= 0.

arvidn:
lgtm, apart from the comment about supporting default constructed objects.

PR #2119, opened 2017-07-02
Add get_flags()/set_flags() to torrent_handle, deprecate individual functions

AllSeeingEyeTolledEweSew:
Per our discussions a while ago, I refactored several boolean fields into a single "flags" bitfield.

There's now `torrent_handle::get_flags()/set_flags()`. I wasn't quite sure if I should add `torrent_status::flags` in this commit -- I believe it would break the ABI so it shouldn't be in the 1.1 train. I included it in this PR so I could ask what you thought about it.

I also considered that maybe we should move the official definitions of `flags_t` to `torrent_handle`. The `add_torrent_params::flag_foo` definitions could be aliases for `torrent_handle::flag_foo`. Thoughts?

I refactored all tests to use flags, rather than the deprecated fields. Building and testing with `deprecated-functions=off` seems to work.

### DIFF
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -401,8 +401,10 @@ namespace libtorrent {
 		// before disconnecting
 		bool graceful_pause() const { return m_graceful_pause_mode; }
 
-		boost::uint64_t get_flags() const;
-		void set_flags(boost::uint64_t mask, boost::uint64_t flags);
+		boost::uint64_t flags() const;
+		void set_flags(boost::uint64_t flags, boost::uint64_t mask);
+		void set_flags(boost::uint64_t flags) { set_flags(0xffffffffffffffff, flags); }
+		void unset_flags(boost::uint64_t flags) { set_flags(0, flags); }
 
 		void set_upload_mode(bool b);
 		bool upload_mode() const { return m_upload_mode || m_graceful_pause_mode; }
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -403,8 +403,6 @@ namespace libtorrent {
 
 		boost::uint64_t flags() const;
 		void set_flags(boost::uint64_t flags, boost::uint64_t mask);
-		void set_flags(boost::uint64_t flags) { set_flags(0xffffffffffffffff, flags); }
-		void unset_flags(boost::uint64_t flags) { set_flags(0, flags); }
 
 		void set_upload_mode(bool b);
 		bool upload_mode() const { return m_upload_mode || m_graceful_pause_mode; }
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -401,6 +401,9 @@ namespace libtorrent {
 		// before disconnecting
 		bool graceful_pause() const { return m_graceful_pause_mode; }
 
+		boost::uint64_t get_flags() const;
+		void set_flags(boost::uint64_t mask, boost::uint64_t flags);
+
 		void set_upload_mode(bool b);
 		bool upload_mode() const { return m_upload_mode || m_graceful_pause_mode; }
 		bool is_upload_only() const { return is_finished() || upload_mode(); }
--- a/include/libtorrent/torrent_status.hpp
+++ b/include/libtorrent/torrent_status.hpp
@@ -460,6 +460,7 @@ namespace libtorrent {
 		// was saved.
 		bool need_save_resume = false;
 
+#ifndef TORRENT_NO_DEPRECATE
 		// true if the session global IP filter applies
 		// to this torrent. This defaults to true.
 		bool ip_filter_applies = false;
@@ -495,6 +496,16 @@ namespace libtorrent {
 		// true when the torrent is in sequential download mode. In this mode
 		// pieces are downloaded in order rather than rarest first.
 		bool sequential_download = false;
+#else
+		// hidden
+		bool deprecated_ip_filter_applies = false;
+		bool deprecated_upload_mode = false;
+		bool deprecated_share_mode = false;
+		bool deprecated_super_seeding = false;
+		bool deprecated_paused = false;
+		bool deprecated_auto_managed = false;
+		bool deprecated_sequential_download = false;
+#endif
 
 		// true if all pieces have been downloaded.
 		bool is_seeding = false;
@@ -515,10 +526,15 @@ namespace libtorrent {
 		// torrent.
 		bool has_incoming = false;
 
+#ifndef TORRENT_NO_DEPRECATE
 		// true if the torrent is in seed_mode. If the torrent was started in
 		// seed mode, it will leave seed mode once all pieces have been checked
 		// or as soon as one piece fails the hash check.
 		bool seed_mode = false;
+#else
+		// hidden
+		bool deprecated_seed_mode = false;
+#endif
 
 		// this is true if this torrent's storage is currently being moved from
 		// one location to another. This may potentially be a long operation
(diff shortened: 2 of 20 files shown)


include/libtorrent/torrent_status.hpp
```diff
@@ -511,13 +527,22 @@ namespace libtorrent
...
+		// hidden
+		bool _dummy_stop_when_ready;
+#endif
 
 		// the info-hash for this torrent
 		sha1_hash info_hash;
+
+		// reflects several of the torrent's flags. For more
+		// information, see ``torrent_handle::get_flags()``.
+		boost::uint64_t flags;
```
arvidn: this would break the ABI from 1.1.x

src/torrent.cpp
```diff
@@ -1094,6 +1094,82 @@ namespace libtorrent
 #endif
 	}
 
+	boost::uint64_t torrent::get_flags() const
```
arvidn: I wonder if ``get_flags()`` and ``set_flags()`` should be modeled after http://en.cppreference.com/w/cpp/io/ios_base/flags

include/libtorrent/torrent_status.hpp
```diff
@@ -468,6 +469,16 @@ namespace libtorrent
 		// true when the torrent is in sequential download mode. In this mode
 		// pieces are downloaded in order rather than rarest first.
 		bool sequential_download;
+#else
+		// hidden
+		bool _dummy_ip_filter_applies;
```
arvidn: in the master branch, there's a macro ``TORRENT_DEPRECATED_MEMBER`` to annotate members as deprecated.

test/test_flags.cpp
```diff
@@ -0,0 +1,172 @@
...
+	TEST_EQUAL(h.get_flags() & flags, flags);
+}
+
+void test_set_after_add(boost::uint64_t flags)
+{
+	session ses(settings());
+	add_torrent_params p;
+	p.save_path = ".";
+	error_code ec;
+	p.ti = boost::shared_ptr<torrent_info>(
```
arvidn: I think it would be better to use ``make_shared`` here (and the other places).

test/test_resume.cpp
```diff
@@ -572,137 +573,153 @@ TORRENT_TEST(plain)
 {
 	lt::session ses(settings());
 
-	torrent_status s = test_resume_flags(ses, 0).status();
+	torrent_handle h = test_resume_flags(ses, 0);
+	torrent_status s = h.status();
+	boost::uint64_t flags = h.get_flags();
```
arvidn: I believe this can be ``const``

arvidn: (and the other versions)

src/torrent.cpp
arvidn: I suppose these functions are closer to ``setf()`` and ``unsetf()``, which they already model pretty closely. It also clearly would be bad to make the setter be blocking just to return the previous flags

arvidn:
I am in favor of this patch, but I think it should go into master.

I would also like to make the flags (in ``add_torrent_params``) to be an ``enum class`` (treated with the enum bitfield operators in [flags.hpp](https://github.com/arvidn/libtorrent/blob/master/include/libtorrent/flags.hpp)). But perhaps that's a separate patch.

As for naming, I think the c++ way is to drop the "get_" and "set_". However, the setter is a bit special in this case, since it ORs in the flags.

Looking at ``<iostream>``, perhaps there should be another convenience function, ``unsetf()`` seting the flags to 0 and the mask to whatever is passed in.

AllSeeingEyeTolledEweSew:
Sorry, my C++ experience is pretty old, so I don't have a ton of intuition about naming.

Based on your comments, my plan is to have:
`uint64_t flags()`
`void flags(uint64_t flags, uint64_t mask)`
`void unset_flags(uint64_t flags)`

If you recall, we originally started talking about this patch because I wanted to add one more torrent-level boolean flag. If we push this to master instead of 1.1, will that single-flag patch also move to master, or should it be implemented with boolean fields and functions only to get deprecated quickly?

This is why I thought to add the setter/getter functions now, but save the change to `torrent_status` until master.

AllSeeingEyeTolledEweSew:
Now that I'm looking at the changes in `examples/*` that I forgot about, it seems very awkward to deprecate the boolean fields from `torrent_status`, but only have `flags` available on `torrent_handle`. The examples are set up so that `torrent_status` is a nice read-only view of a torrent, it would probably be awkward for clients to have to jump between `torrent_status` and `torrent_handle` so much.

Now I'm also in favor of moving this change to master. However I would really like to get that earlier single boolean field change in 1.1 if possible. What do you think?

arvidn:
I think ``uint64_t flags()``, ``void set_flags(uint64_t fl, uint64_t mask)`` and ``unset_flags(uint64_t fl)``. Where ``unset_flags()`` really just is an alias for ``set_flags(0, fl)``.

AllSeeingEyeTolledEweSew:
I updated the names per your advice. Now working on rebasing this branch to master.

arvidn:
there are still some build errors, for instance: https://ci.appveyor.com/project/arvidn/libtorrent/build/6641/job/9sym205efe5fq046#L761

src/torrent_handle.cpp
```diff
@@ -265,6 +265,29 @@ namespace libtorrent {
...
 	}
 
+	boost::uint64_t torrent_handle::flags() const
+	{
+		return sync_call_ret<boost::uint64_t>(0, &torrent::flags);
+	}
+
+	void torrent_handle::set_flags(boost::uint64_t flags, boost::uint64_t mask) const
+	{
+		void (torrent::*set_flags)(boost::uint64_t, boost::uint64_t) = &torrent::set_flags;
```
arvidn: it might be simpler if the internal type (``torrent``) only has the ``set_flags`` that takes a mask, and both of these call it. Then you don't need to cast to pick the right overload here

arvidn: in fact, ``unset_flags()`` could call that same function too, but just forward the flags as the mask

include/libtorrent/torrent.hpp
```diff
@@ -401,6 +401,11 @@ namespace libtorrent {
 		// before disconnecting
 		bool graceful_pause() const { return m_graceful_pause_mode; }
 
+		boost::uint64_t flags() const;
+		void set_flags(boost::uint64_t flags, boost::uint64_t mask);
+		void set_flags(boost::uint64_t flags) { set_flags(0xffffffffffffffff, flags); }
+		void unset_flags(boost::uint64_t flags) { set_flags(0, flags); }
```
arvidn: I think these two convenience functions should just exist on the user-facing ``torrent_handle``. I think that would simplify things.

include/libtorrent/torrent_status.hpp
```diff
@@ -460,6 +460,7 @@ namespace libtorrent {
 		// was saved.
 		bool need_save_resume = false;
 
+#ifndef TORRENT_NO_DEPRECATE
```
arvidn: where you having second thoughts about making this change to torrent_status? I understand your argument that ``torrent_status`` should be convenient and readable to use, and picking out flags may make that less readable and convenient. In my mind, the main arguments in favor of this change are:

1. making it possible to extend the API with new flags without breaking ABI
2. making the interface to libtorrent smaller, with fewer functions, and possibly simpler to some degree.

At least the ABI argument would apply here

arvidn:
lgtm (with one minor request for change). After this patch, I think it's important to make the flags a proper enum class as well.

include/libtorrent/torrent_status.hpp
AllSeeingEyeTolledEweSew: I was waffling about making the change to torrent_status because I really wanted my other change (#1937) to go to RC_1_1, not master. If I could get *this* change to RC_1_1 I could rebase #1937 after it.

src/torrent_handle.cpp
AllSeeingEyeTolledEweSew: Done

AllSeeingEyeTolledEweSew:
Not sure why test_ssl.cpp.utp_config7 failed on darwin. I don't think I changed anything that affects it since the last build where it succeeded. Is this test flaky sometimes?

arvidn:
yeah, the ssl test are a bit flaky

arvidn:
I squashed, rebased and added another commit on top of this change here: https://github.com/arvidn/libtorrent/pull/2135

AllSeeingEyeTolledEweSew:
Looks like #2135 supersedes this? Should it get closed?

PR #2121, opened 2017-07-02
use UNC paths pervasively on windows

arvidn:


src/path.cpp
```diff
@@ -163,21 +162,16 @@ namespace libtorrent {
...
-		else
-		{
-			std::string sep_path = path;
-			std::replace(sep_path.begin(), sep_path.end(), '/', '\\');
-			prepared_path = "\\\\?\\" + (is_complete(sep_path) ? sep_path : complete(sep_path));
-		}
+		std::string prepared_path =
+			(path.substr(0,2) == "\\\\")
+			? path
+			: "\\\\?\\" + (is_complete(path) ? path : complete(path));
```
ajax16384:
````c++
std::string complete(string_view f)
{
	if (is_complete(f)) return f.to_string();
...
````
so last line can be:
````c++
: "\\\\?\\" + complete(path);
````

Comment (arvidn) [inline src/path.cpp]: true

File: src/path.cpp [hunk 2]
```diff
@@ -165,23 +167,13 @@ namespace {
...
-		{
-			std::string sep_path = path;
-			std::replace(sep_path.begin(), sep_path.end(), '/', '\\');
-			prepared_path = "\\\\?\\" + (is_complete(sep_path) ? sep_path : complete(sep_path));
-		}
-#else
-		std::string prepared_path = path;
+		std::string prepared_path = complete(path);
+		if (prepared_path.substr(0,2) != "\\\\")
+			prepared_path = "\\\\?\\" + prepared_path;
```
LordNyriox: @arvidn:  Something I wondered while looking at Libtorrent's existing UNC path implementation: 

Why are there 4 backslashes in the prefix here, rather than 2?

arvidn: in C and C++, `\` is an escape character, for `\n` for instance. To represent one backslash, you have to escape it `\\`.

PR #2123, opened 2017-07-02
added block_upload_alert to allow client to track upload activity

toinetoine:
Ability for client to track [upload activity](https://stackoverflow.com/a/13938080/1053362).

https://github.com/arvidn/libtorrent/pull/2115

### DIFF
--- a/simulation/test_swarm.cpp
+++ b/simulation/test_swarm.cpp
@@ -537,6 +537,51 @@ TORRENT_TEST(torrent_completed_alert)
 	TEST_EQUAL(num_file_completed, 1);
 }
 
+TORRENT_TEST(block_uploaded_alert)
+{
+	// blocks[piece count][number of blocks per piece] (each block's element will
+	// be set to true when a block_uploaded_alert alert is received for that block)
+	std::vector<std::vector<bool>> blocks;
+
+	setup_swarm(2, swarm_test::upload
+		// add session
+		, [](lt::settings_pack& pack)
+		{
+			pack.set_int(lt::settings_pack::alert_mask,
+				alert::progress_notification | alert::status_notification);
+		}
+		// add torrent
+		, [](lt::add_torrent_params&) {}
+		// on alert
+		, [&](lt::alert const* a, lt::session&) {
+			if (auto at = lt::alert_cast<lt::add_torrent_alert>(a))
+			{
+				// init blocks vector, MUST happen before any block_uploaded_alert alerts
+				int blocks_per_piece = at->handle.torrent_file()->piece_length() / 0x4000;
+				blocks.resize(at->handle.torrent_file()->num_pieces(), std::vector<bool>(blocks_per_piece, false));
+			}
+			else if (auto at = lt::alert_cast<lt::block_uploaded_alert>(a))
+			{
+				TEST_EQUAL(blocks[at->piece_index][at->block_index], false);
+				blocks[at->piece_index][at->block_index] = true;
+			}
+		}
+		// terminate
+		, [](int, lt::session&) -> bool
+		{ return false; });
+
+		// ensure a block_uploaded_alert was received for each block in the torrent
+		TEST_CHECK(std::all_of(blocks.begin(), blocks.end(),
+			[](std::vector<bool> const& piece_row) {
+				return std::all_of(piece_row.begin(), piece_row.end(),
+					[](bool upload_alert_received) {
+						return upload_alert_received;
+					}
+				);
+			}
+		));
+}
+
 // template for testing running swarms with edge case settings
 template <typename SettingsFun>
 void test_settings(SettingsFun fun)
--- a/src/alert.cpp
+++ b/src/alert.cpp
@@ -2414,4 +2414,22 @@ namespace {
 			, m_v6_num_nodes, m_v6_nodes_idx);
 	}
 
+	block_uploaded_alert::block_uploaded_alert(aux::stack_allocator& alloc, torrent_handle h
+		, tcp::endpoint const& ep, peer_id const& peer_id, int block_num
+		, piece_index_t piece_num)
+		: peer_alert(alloc, h, ep, peer_id)
+		, block_index(block_num)
+		, piece_index(piece_num)
+	{
+		TORRENT_ASSERT(block_index >= 0 && piece_index >= piece_index_t{0});
+	}
+
+	std::string block_uploaded_alert::message() const
+	{
+		char ret[200];
+		snprintf(ret, sizeof(ret), "%s block uploaded to a peer (piece: %u block: %u)"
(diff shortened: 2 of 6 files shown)


src/alert.cpp
```diff
@@ -2414,4 +2414,22 @@ namespace {
...
 	}
 
+	block_uploaded_alert::block_uploaded_alert(aux::stack_allocator& alloc, torrent_handle h
+		, tcp::endpoint const& ep, peer_id const& peer_id, int block_num
+		, piece_index_t piece_num)
+		: peer_alert(alloc, h, ep, peer_id)
+		, block_index(block_num)
+		, piece_index(piece_num)
+	{
+		TORRENT_ASSERT(block_index >= 0 && piece_index_t{0});
```
arvidn: you have a typo here. ``block_index <``

arvidn:
Can you think of a good way to test this?
I'm thinking you could add a test to ``simulation/test_transfer.cpp`` where you can collect all ``block_upload_alerts`` and make sure you get exactly one for every block, from the seed.

toinetoine:
A test in `simulation/test_transfer.cpp` is a good idea. Will require small change to 'run_test' in `simulation/test_transfer.cpp` since as of now it only allows attaching an alert handler to the downloader's session.

arvidn:
ah, ``test_swarm`` may be easier to extend then. It has the option to have the session under test to either be the seeder or the downloader

arvidn:
@toinetoine I'm under the impression that you intend to update this PR with a test, is that right? (I think it would be great if you would)

toinetoine:
@arvidn Yes sorry, was figuring out bjam. I had to include `torrent.hpp` in `test_swarm` since that's the only way to access `block_size` (needed to calculate the number of blocks per piece).

simulation/test_swarm.cpp
```diff
@@ -537,6 +538,56 @@ TORRENT_TEST(torrent_completed_alert)
...
+		, [](lt::add_torrent_params&) {}
+		// on alert
+		, [&](lt::alert const* a, lt::session&) {
+			if (auto at = lt::alert_cast<lt::add_torrent_alert>(a))
+			{
+				// init blocks vector, MUST happen before any block_uploaded_alert alerts
+				std::shared_ptr<lt::torrent> torrent = at->handle.native_handle();
+				std::shared_ptr<const torrent_info> torrent_info = torrent->get_torrent_copy();
+				int blocks_per_piece = torrent_info->piece_length() / torrent->block_size();
+				for (int i = 0; i < torrent_info->num_pieces(); ++i)
```
arvidn: I believe you could simplify this loop to:
``blocks.resize(torrent_info->num_pieces(), std::vector<bool>(blocks_per_piece, false));``

simulation/test_swarm.cpp
```diff
@@ -537,6 +538,56 @@ TORRENT_TEST(torrent_completed_alert)
...
+				{
+					std::vector<bool> piece_row(blocks_per_piece, false);
+					blocks.push_back(piece_row);
+				}
+			}
+			else if (auto at = lt::alert_cast<lt::block_uploaded_alert>(a))
+			{
+				blocks[at->piece_index][at->block_index] = true;
```
arvidn: I think it would also make sense to ``TORRENT_EQUAL(blocks[at->piece_index][at->block_index], false);

To make sure there are no duplicates

simulation/test_swarm.cpp
```diff
@@ -537,6 +538,56 @@ TORRENT_TEST(torrent_completed_alert)
...
+			}
+		}
+		// terminate
+		, [](int, lt::session&) -> bool
+		{ return false; });
+
+		// ensure a block_uploaded_alert was received for each block in the torrent
+		TEST_CHECK(std::all_of(blocks.begin(), blocks.end(),
+			[](std::vector<bool> piece_row) {
```
arvidn: it would feel better to take this by ``const&``

simulation/test_swarm.cpp
```diff
@@ -537,6 +538,56 @@ TORRENT_TEST(torrent_completed_alert)
...
+		}
+		// add torrent
+		, [](lt::add_torrent_params&) {}
+		// on alert
+		, [&](lt::alert const* a, lt::session&) {
+			if (auto at = lt::alert_cast<lt::add_torrent_alert>(a))
+			{
+				// init blocks vector, MUST happen before any block_uploaded_alert alerts
+				std::shared_ptr<lt::torrent> torrent = at->handle.native_handle();
+				std::shared_ptr<const torrent_info> torrent_info = torrent->get_torrent_copy();
```
arvidn: you don't need to copy the ``torrent_info`` object here. just call ``torrent_file()``. Also, feel free to use ``auto``

simulation/test_swarm.cpp
```diff
@@ -558,17 +558,14 @@ TORRENT_TEST(block_uploaded_alert)
 			if (auto at = lt::alert_cast<lt::add_torrent_alert>(a))
 			{
 				// init blocks vector, MUST happen before any block_uploaded_alert alerts
-				std::shared_ptr<lt::torrent> torrent = at->handle.native_handle();
-				std::shared_ptr<const torrent_info> torrent_info = torrent->get_torrent_copy();
+				auto torrent = at->handle.native_handle();
+				auto torrent_info = &torrent->torrent_file();
```
arvidn: sorry I wasn't very clear here (and I missed that you're calling ``native_handle``). But you can get the ``torrent_info`` object by just calling: ``st->handle.torrent_file()``. There's no need to pull out the internal object.

arvidn:
lgtm. just one small fix

simulation/test_swarm.cpp
toinetoine: Ok so remove `auto torrent_info = &torrent->torrent_file();` and just call `at->handle.torrent_file()` the 2 times `torrent_info` is used:

```c++
...
auto torrent_info = &torrent->torrent_file();
int blocks_per_piece = torrent_info->piece_length() / torrent->block_size();
blocks.resize(torrent_info->num_pieces(), std::vector<bool>(blocks_per_piece, false));
...
```

Becomes:

```c++
...
int blocks_per_piece = at->handle.torrent_file()->piece_length() / torrent->block_size();
blocks.resize(at->handle.torrent_file()->num_pieces(), std::vector<bool>(blocks_per_piece, false));
...
```


?

arvidn: hm. do you still need the ``torrent`` object because of ``block_size()``?
you could actually hard code the block size to ``0x4000`` as well. Originally I intended to make the block size configurable, but the value of that ended up not warrant the complexity, so by now that block size of 0x4000 is quite engrained in libtorrent (and hard coding it here would not contribute significantly to the difficulty of changing this in the future)

arvidn:
lgtm

Issue #2128, opened 2017-07-05
Crashes in DHT, possibly related to invalid reference to node in dht_tracker::connection_timeout


aldenml:
libtorrent version (or branch): `master`
platform/architecture: `all`

Hi, I'm getting a lot of crashes in the DHT like this
```
native: pc 0000000000393b5f  (libtorrent::dht::dht_tracker::send_packet(libtorrent::aux::session_listen_socket*, libtorrent::entry&, boost::asio::ip::basic_endpoint<boost::asio::ip::udp> const&)+271)
native: pc 00000000003b5fc6  (libtorrent::dht::rpc_manager::invoke(libtorrent::entry&, boost::asio::ip::basic_endpoint<boost::asio::ip::udp> const&, std::__ndk1::shared_ptr<libtorrent::dht::observer>)+1446)
native: pc 00000000003a678c  (libtorrent::dht::bootstrap::invoke(std::__ndk1::shared_ptr<libtorrent::dht::observer>)+668)
native: pc 00000000003b91a7  (libtorrent::dht::traversal_algorithm::add_requests()+567)
native: pc 00000000003b8de9  (libtorrent::dht::traversal_algorithm::start()+73)
native: pc 000000000043ecc7  (libtorrent::dht::find_data::start()+39)
native: pc 000000000039ffd0  (libtorrent::dht::node::tick()+192)
native: pc 000000000039100a  (libtorrent::dht::dht_tracker::refresh_timeout(boost::system::error_code const&)+74)
```
and
```
native: pc 00000000003dbbe0  (_ZN10libtorrent8ip_voter9cast_voteERKN5boost4asio2ip7addressEiS6_+160)
native: pc 00000000002b5094  (_ZN10libtorrent3aux12session_impl20set_external_addressERNS0_15listen_socket_tERKN5boost4asio2ip7addressEiS9_+244)
native: pc 00000000002b519d  (_ZThn4_N10libtorrent3aux12session_impl20set_external_addressEPNS0_21session_listen_socketERKN5boost4asio2ip7addressES9_+45)
native: pc 000000000039a0b0   (_ZN10libtorrent3dht4node8incomingERKNS0_3msgE+1008)
native: pc 0000000000393022 (_ZN10libtorrent3dht11dht_tracker15incoming_packetERKN5boost4asio2ip14basic_endpointINS4_3udpEEENS_4spanIKcEE+770)
[...]

PR #2132, opened 2017-07-07
don't move listen_socket_t when deleting sockets

ssiloti:
Pointers to listen_socket_t are stored in dht_tracker, so they
cannot be copied or moved once they are added to m_listen_sockets.

listen_socket_t should really be made non-copyable and non-moveable, but the
socket creation path needs to be refactored before that can be done.

I believe this is the primary cause of the crashes reported in #2128

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -260,35 +260,12 @@ namespace aux {
 	}
 #endif // TORRENT_DISABLE_DHT
 
-	std::list<listen_socket_t>::iterator partition_listen_sockets(
-		std::vector<listen_endpoint_t>& eps
-		, std::list<listen_socket_t>& sockets)
+	bool operator==(listen_endpoint_t const& ep, listen_socket_t const& sock)
 	{
-		return std::partition(sockets.begin(), sockets.end()
-			, [&eps](listen_socket_t const& sock)
-		{
-			auto match = std::find_if(eps.begin(), eps.end()
-				, [&sock](listen_endpoint_t const& ep)
-			{
-				return ep.ssl == sock.ssl
-					&& ep.port == sock.original_port
-					&& ep.device == sock.device
-					&& ep.addr == sock.local_endpoint.address();
-			});
-
-			if (match != eps.end())
-			{
-				// remove the matched endpoint so that another socket can't match it
-				// this also signals to the caller that it doesn't need to create a
-				// socket for the endpoint
-				eps.erase(match);
-				return true;
-			}
-			else
-			{
-				return false;
-			}
-		});
+		return ep.ssl == sock.ssl
+			&& ep.port == sock.original_port
+			&& ep.device == sock.device
+			&& ep.addr == sock.local_endpoint.address();
 	}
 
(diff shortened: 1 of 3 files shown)


src/session_impl.cpp
```diff
@@ -1887,6 +1864,50 @@ namespace {
 		}
 	}
 
+	void session_impl::remove_listen_sockets(std::vector<listen_endpoint_t>& eps)
+	{
+		m_listen_sockets.remove_if([&eps, this](listen_socket_t const& sock)
```
arvidn: this will only consolidate the elements not removed in the beginning of the container, and return an iterator to the first element to erase. you still need to call ``m_listen_sockets.erase(iter, m_listen_sockets.end())``, or am I missing something subtle?

arvidn: oh, I am. this is a ``std::list``, and ``remove_if``-member function means what you think it does (as opposed to the free function ``remove_if``). never mind

aldenml:
@ssiloti I confirmed that doing `std::partition` in a `std::list` does not invalidate the address of listen_socket_t

src/session_impl.cpp
ssiloti: This is the special `std::list` version of `remove_if` which removes the items as it goes.

src/session_impl.cpp
```diff
@@ -1887,6 +1864,50 @@ namespace {
 		}
 	}
 
+	void session_impl::remove_listen_sockets(std::vector<listen_endpoint_t>& eps)
+	{
```
arvidn: it would be nice if this could be a free function, and unit testable. but I guess the access to ``m_dht`` makes that a bit tricky

ssiloti:
@aldenml `std::partition` swaps the items by value which doesn't invalidate the pointers themselves, but it does mean that the pointers now point to different sockets so the wrong `node` can get removed.

aldenml:
yes, but since `std::list` is a linked list, the swap (in this case, an internal `iter_swap`) only change the links, and the address (pointers) to values stay the same (I tested it and looked at the implementation), or I'm understanding something different?

aldenml:
I'm not saying the code is not correct, but if they are indeed equivalent in result, the drawback of this version is the lack of unit testing I think.

ssiloti:
I don't think that `iter_swap` optimization is guaranteed. If I try to build with `listen_socket_t` made non-copyable and non-movable it generates an error, at least with gcc.

aldenml:
I was expecting a comment like that :) and I think you are right.

Still, I believe my other PR (https://github.com/arvidn/libtorrent/pull/2133) is necessary.

src/session_impl.cpp
ssiloti: I could of course hoist the `session_impl` specific code into a function parameter, but I'm not convinced it's worth the extra boilerplate. This function is a pretty straightforward application of `remove_if`. The most likely place for bugs is the comparison function between `listen_socket_t` and `listen_endpoint_t`, and that is still being unit tested.

PR #2133, opened 2017-07-07
avoid executing timed async task if the dht node is already removed

aldenml:
An attempt to fix https://github.com/arvidn/libtorrent/issues/2128#issuecomment-313575564

### DIFF
--- a/src/kademlia/dht_tracker.cpp
+++ b/src/kademlia/dht_tracker.cpp
@@ -149,8 +149,8 @@ namespace libtorrent { namespace dht {
 			error_code ec;
 			n.first->second.connection_timer.expires_from_now(seconds(1), ec);
 			n.first->second.connection_timer.async_wait(
-				std::bind(&dht_tracker::connection_timeout, self(), std::ref(n.first->second), _1));
-			n.first->second.dht.bootstrap(std::vector<udp::endpoint>(), find_data::nodes_callback());
+				std::bind(&dht_tracker::connection_timeout, self(), n.first->first, _1));
+			n.first->second.dht.bootstrap({}, find_data::nodes_callback());
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


src/kademlia/dht_tracker.cpp
```diff
@@ -644,6 +648,17 @@ namespace libtorrent { namespace dht {
...
+		// via dht.tick, if the pair (socket, node) is removed from the map while this operation
+		// is right to go executed, this address becomes invalid
+		if (m_nodes.find(s) == m_nodes.end()) // socket already removed
+		{
+#ifndef TORRENT_DISABLE_LOGGING
+			m_log->log(dht_logger::tracker, "ignoring send_packet due to removed listen socket");
+#endif
+			return false;
+		}
+
```
ssiloti: This check is unnecessary because `dht_tracker::refresh_timeout` calls tick by iterating over `m_nodes` so the socket is guaranteed to be in `m_nodes`. It might be good to keep this as an assert though.

arvidn:
I haven't had a chance to fully review this patch, but you might want to reproduce the issue in a simulation. That would increase confidence that (1) we understand what the problem is and (2) that this fixes it. It would also ensure we don't introduce a problem like this again.

arvidn:
I haven't had a chance to review this patch yet, but you might want to try to reproduce the issue in a simulation. To be confident we understand what the problem is, that this patch fixes it and that we won't introduce it again in the future

src/kademlia/dht_tracker.cpp
aldenml: Looking at it again, yes, I will change that to assert.

To confirm, do you think that the `dht_tracker::send_packet` crash is then addressed by your other PR?

ssiloti: I suspect #2132 will fix that, but I don't have hard evidence since I can't repo the crash.

aldenml:
@arvidn agree, but at this moment I have no idea how to do it, let me see

src/kademlia/dht_tracker.cpp
```diff
@@ -242,15 +242,21 @@ namespace libtorrent { namespace dht {
...
-	void dht_tracker::connection_timeout(tracker_node& n, error_code const& e)
+	void dht_tracker::connection_timeout(aux::session_listen_socket* s, error_code const& e)
 	{
 		if (e || !m_running) return;
 
-		time_duration d = n.dht.connection_timeout();
+		auto const it = m_nodes.find(s);
+		// this could happen if the task is about to be executed (and not cancellable) and
+		// the socket is just removed
+		if (it == m_nodes.end()) return; // node already destroyed
```
aldenml: @arvidn I added the simulation test, this line is the critical one, if you comment it, the test will fail

PR #2135, opened 2017-07-09
torrent flags refactor

arvidn:
this amends https://github.com/arvidn/libtorrent/pull/2119 with making the flags strongly typed

### DIFF
--- a/include/libtorrent/torrent_flags.hpp
+++ b/include/libtorrent/torrent_flags.hpp
@@ -1,0 +1,268 @@
+/*
+
+Copyright (c) 2017, Arvid Norberg
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in
+      the documentation and/or other materials provided with the distribution.
+    * Neither the name of the author nor the names of its
+      contributors may be used to endorse or promote products derived
+      from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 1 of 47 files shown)


arvidn:
@AllSeeingEyeTolledEweSew would you mind taking a look at this? does it look ok?

include/libtorrent/torrent_flags.hpp
```diff
@@ -0,0 +1,267 @@
...
+	// sets the sequential download state for the torrent. It has the same
+	// effect as calling ``torrent_handle::sequential_download(true)`` on
+	// the torrent handle immediately after adding it.
+	constexpr torrent_flags_t sequential_download{0x800};
+
+#ifndef TORRENT_NO_DEPRECATE
+	// indicates that this torrent should never be unloaded from RAM, even
+	// if unloading torrents are allowed in general. Setting this makes
+	// the torrent exempt from loading/unloading management.
+	constexpr torrent_flags_t TORRENT_DEPRECATED_MEMBER pinned{0x1000};;
```
aldenml: there are a few double semicolon in this patch like `};;`

AllSeeingEyeTolledEweSew:
Looks good overall. I actually started writing this myself: see https://github.com/AllSeeingEyeTolledEweSew/libtorrent/tree/asetes-torrent-flags-enum-class . I branched this from my #2119 branch. I thought it might go in a separate PR, but couldn't figure out how to create a new PR based off of a pending PR (this instinct comes from big-software-house source control processes, I guess it doesn't fit github as well), so I was waiting for #2119 to land.

I had a few differences: I ended up adding `operator<<(std::ostream&, x)` to make it easier to log the flags.

Would you mind also adding a python binding for the torrent_flags class? I have one in my branch.

I kept `add_torrent_params::flags_t` separate from `torrent_flags`, but kept the numeric spaces the same. I like your approach more though; not all flags can be set in all cases (`seed_mode`) so there's no surprise that some flags only apply to `add_torrent`. May as well have a single class.

arvidn:
@AllSeeingEyeTolledEweSew I added the ``operator<<`` too, but I made that a separate PR (which I have rebased on top of).

This patch includes python bindings for the flags. Both the new and the deprecated ones.
I actually deprecated all of the flags in ``add_torrent_params`` and moved them into the ``torrent_flags`` namespace. You may have noticed that I also abandoned the ``enum class`` approach in favor of a proper class. again, I did that in a separate PR which this has been rebased on top of.

arvidn:
The reason I folder your change into a single PR is that I really think exposing the flags wider really makes the problem of just using enums more prominent, so I wanted to avoid having that at all.

include/libtorrent/torrent_flags.hpp
arvidn: thanks. I fixed this

PR #2136, opened 2017-07-09
fix inconsistency in file_priorities and override_resume_data behavior

arvidn:
file_priorities are not subject to the override_resume_data flag

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -5637,7 +5637,7 @@ namespace libtorrent
 		{
 			inc_refcount("file_priority");
 			m_ses.disk_thread().async_set_file_priority(m_storage.get()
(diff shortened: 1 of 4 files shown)


src/torrent.cpp
```diff
@@ -7044,9 +7044,7 @@ namespace libtorrent
 		if (m_completed_time != 0 && m_completed_time < m_added_time)
 			m_completed_time = m_added_time;
 
-		// load file priorities except if the add_torrent_param file was set to
-		// override resume data
-		if (!m_override_resume_data || m_file_priority.empty())
+		if (m_file_priority.empty())
```
arvidn: hm.. I need a test for this. It's not obvious this is correct. This looks like the priorities in add_torrent_params always take priority.

PR #2137, opened 2017-07-09
make listen_socket_t non-copyable and non-movable

ssiloti:
The DHT stores a pointer to this struct so it must not be copied or moved once
it is stored in m_listen_sockets.


ssiloti:
I did that for convenience. This way you can create a `listen_socket_impl` and still be able to copy it around. Only when it is added to `m_listen_sockets` does it get pinned in place.

PR #2138, opened 2017-07-10
fix storage destruction order issue

d-komarov:
back-ported from b553cb32f785775b527bd53ea0153400c488e45e

### DIFF
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -1168,6 +1168,10 @@ namespace libtorrent
 		void on_file_renamed(disk_io_job const* j);
 		void on_cache_flushed(disk_io_job const* j);
 
+		// this is used when a torrent is being removed.It synchronizes with the
+		// disk thread
+		void on_torrent_aborted();
+
 		// upload and download rate limits for the torrent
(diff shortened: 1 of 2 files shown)


d-komarov:
Could you restart tests please?

arvidn:
lgtm. would you mind including the ChangLog comment as well (I believe I put one for master).
This patch is picked from the master branch, right? I just want to make sure this will merge relatively cleanly when merged forward again.

arvidn:
there's something weird with appveyor. when I ask it to rebuild, it triggers a build for a different commit

d-komarov:
Changes originate from https://github.com/arvidn/libtorrent/pull/1934. I've manually extracted changes and put em to my branch. Any extra changes were omitted to grant code compatibility and deliver the fix.
No questions, I'll add a line to the changelog. But to be honest there is no mention of this fix in a master branch. Double checked it.

d-komarov:
I'm an expert in appveynor, so I can just believe 
 that this strange behavior is not a consequence of my PR.

d-komarov:
I've taken a look at some report that contained test results and noticed that all fails were caused by inability to delete previous test files.

arvidn:
lgtm

PR #2141, opened 2017-07-10
cmake cleanups and enhancements

ssiloti:


libtorrent-rasterbarConfig.cmake
```diff
@@ -0,0 +1 @@
+include("${CMAKE_CURRENT_LIST_DIR}/libtorrent-rasterbarTargets.cmake")
```
arvidn: where does this file come from? "libtorrent-rasterbarTargets.cmake"

arvidn:
lgtm, but I'm not a cmake expert

libtorrent-rasterbarConfig.cmake
ssiloti: It's generated by cmake when you do a "make install" or equivalent. You can see the reference to it on line 382 in CMakeLists.txt

arvidn:
do you think there's a chance cmake could replace autoconf and automake for building tarballs?
I believe that is the main purpose of the makefiles right now.

ssiloti:
Cmake comes with a tool called Cpack which can generate binary and source archives, I'm not sure how much of an improvement it would be over autotools. On Linux autotools is still often the preferred build system. I think it's mostly due to familiarity. Distro maintainers all know how to use autotools and often have their own tooling built around it.

PR #2148, opened 2017-07-12
add reserve entry::to_string()

pavel-pimenov:


src/entry.cpp
```diff
@@ -637,6 +637,7 @@ namespace {
 	std::string entry::to_string() const
 	{
 		std::string ret;
+		ret.reserve(256);
```
arvidn: It seems reasonable to only do this if there's more than one element in the entry, but this will pessimize printing an integer. What do you think about only reserving this if the type is a list or dict?

ssiloti: If this is done at all it should definitely only be done for lists and dicts. This is effectively defeating the Short String Optimization for small entries.

pavel-pimenov:
change 256 -> 280
simple runtime statistics:
		static std::map<data_type, std::map<size_t, unsigned> > g_entry_stat;
		auto& item = g_entry_stat[type()];
		item[ret.size()]++;

default

arvidn:
Are you counting elements in the lists and dicts, or characters when printing?

arvidn:
at a second look, that screenshot implies that you're counting the number of bytes used by  dictionaries. I don't think that's necessarily a good proxy for how many characters it will turn into when printing. Perhaps a better test would be to capture 1000 DHT packets, print them and look at the distribution of their sizes, and pick some reasonable default allocation from that.

I think DHT packets are reasonable because they are common and probably some of the smallest bencoded structures in use.

src/entry.cpp
```diff
@@ -637,6 +637,7 @@ namespace {
 	std::string entry::to_string() const
 	{
 		std::string ret;
+		if(type() == dictionary_t) ret.reserve(280);
```
arvidn: As I pointed out in the other comment, the rationale for "280" seems suspect. But it starting at a larger size will still probably significantly cut down on the number of times it will have to grow.
would you mind adding a space after the "if" though?

PR #2164, opened 2017-07-20
Fix bug where the resume data would fail to load the piece bitmask

arvidn:
for seeds when suggest_cache was enabled


arvidn:
I wonder what happened to openssl in mingw all of a sudden

aldenml:
maybe something with the cache (like in travis)...and wasn't picking the right library to start with?

PR #2165, opened 2017-07-20
use span in torrent_info constructors

arvidn:
and deprecate the redundant flags argument

### DIFF
--- a/include/libtorrent/torrent_info.hpp
+++ b/include/libtorrent/torrent_info.hpp
@@ -110,6 +110,12 @@ namespace libtorrent {
 		std::uint8_t type;
 	};
 
+	// hidden
+	class from_span_t {};
+
+	// used to disambiguate a bencoded buffer and a filename
+	extern TORRENT_EXPORT from_span_t from_span;
+
 	// TODO: there may be some opportunities to optimize the size if torrent_info.
 	// specifically to turn some std::string and std::vector into pointers
 	class TORRENT_EXPORT torrent_info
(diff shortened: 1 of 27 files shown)


include/libtorrent/torrent_info.hpp
```diff
@@ -146,30 +152,46 @@ namespace libtorrent {
...
-		explicit torrent_info(bdecode_node const& torrent_file, int flags = 0);
-		explicit torrent_info(char const* buffer, int size, int flags = 0);
-		explicit torrent_info(std::string const& filename, int flags = 0);
+		explicit torrent_info(bdecode_node const& torrent_file);
+#ifndef TORRENT_NO_DEPRECATE
+		TORRENT_DEPRECATED
+		torrent_info(char const* buffer, int size, int = 0)
+			: torrent_info(span<char const>{buffer, std::size_t(size)}, from_span) {}
+#endif
+		explicit torrent_info(span<char const> buffer, from_span_t);
```
aldenml: this is a little strange from the client API point of view (I understand the reason). Is keeping the one with `std::string const&` really necessary?

arvidn: yes, I agree. I think the string overload is quite commonly used. In the long run, I would like to disambiguate them by using ``span<std::byte const>`` instead of ``char``.Perhaps that's even feasible in the short term

arvidn: there is precedence for these kinds of tags though. ``std::nothrow`` for instance

aldenml: yeah...languages are not perfect :)

PR #2180, opened 2017-07-24
remove redundant m_disk_cache.set_settings

pavel-pimenov:


arvidn:
It's not obvious that this is redundant. could you please explain?

pavel-pimenov:
https://github.com/arvidn/libtorrent/blob/c4eb4c8b5f829215a50aaf2722ab19f8bf628ca1/src/create_torrent.cpp#L285

https://github.com/arvidn/libtorrent/blob/c4afb5e9496c00a752b9f129bd9d8ef44568d0af/src/disk_io_thread.cpp#L309

arvidn:
right, there are other places settings are applied too. but that doesn't mean this is redundant. You're implying that no logic in between construction and one of those other times the settings are applied depend on the settings. Can you explain why you believe that's the case?

PR #2196, opened 2017-07-30
read_piece_alert bindings for piece

rllola:
The `read_piece_alert` has attribute piece which is a `piece_index_t` which cannot be used in python if returned as it is.

I guess we expect it to have the same behavior like the others `piece_index_t` attributes.


arvidn:
lgtm. can you think of a simple way to exercise this conversion from test.py too?

PR #2203, opened 2017-08-04
fixed -Wshorten-64-to-32 in 32 bits and explicit nullptr check

aldenml:


arvidn:
lgtm

ssiloti:
Wait, something is wrong if the type of `m_inode` does not match `d_ino`.

ssiloti:
What platform are you seeing the warning on?

ssiloti:
No! Don't merge this. There's a bug here that the cast is just covering up.

arvidn:
oops

arvidn:
@ssiloti I assume the warning is on android

arvidn:
@aldenml what is the warning you get exactly? http://pubs.opengroup.org/onlinepubs/009695399/basedefs/dirent.h.html seems to suggest that member definitely should have the ``ino_t`` type.

aldenml:
@arvidn @ssiloti the warning is in android: https://travis-ci.org/aldenml/frostwire-jlibtorrent/jobs/260765521#L448

but even in macOS it does not use the same "type", only happens to be the same at the end. Let me look in android headers

aldenml:
This is what I have in the android 32 bits (similar to macOS)
```cpp
#define __DIRENT64_BODY \
    uint64_t         d_ino; \
    int64_t          d_off; \
    unsigned short   d_reclen; \
    unsigned char    d_type; \
    char             d_name[256]; \

struct dirent { __DIRENT64_BODY };
struct dirent64 { __DIRENT64_BODY };
```
Now, I don't know if `m_inode` should be changed to that type, or if in 32 bits `d_ino` can carry a bigger than `int32` value. Waiting for your comments...

ssiloti:
What does the definition of `ino_t` look like? If `d_ino` is defined as `uint64_t` then `ino_t` should be as well.

aldenml:
```
sysroot/usr/include/asm-generic/posix_types.h:typedef __kernel_ulong_t __kernel_ino_t;
sysroot/usr/include/sys/types.h:typedef __kernel_ino_t __ino_t;
sysroot/usr/include/sys/types.h:typedef __ino_t ino_t;
```
and
```cpp
typedef unsigned long __kernel_ulong_t;
```
and that is a 32bit size type in 32 bit archs, I believe same problem exists in linux 32, but with `gcc` it's not visible, this warning is from `clang`.

ssiloti:
I found a [bug report](https://issuetracker.google.com/issues/37011207) which indicates that this is a known issue with android which they wont fix because it would break ABI compatibility. I think the best solution for us is to make a special case for andoid where we define `m_inode` as `uint64_t`. There should be a comment explaining why with a link to that bug report.

aldenml:
I will do that

aldenml:
See https://github.com/arvidn/libtorrent/pull/2204

PR #2221, opened 2017-08-07
more use of emplace_back in write_resume_data

aldenml:


src/write_resume_data.cpp
```diff
@@ -126,7 +126,7 @@ namespace libtorrent {
 		if (!atp.trackers.empty())
 		{
 			entry::list_type& tr_list = ret["trackers"].list();
-			tr_list.push_back(entry::list_type());
+			tr_list.emplace_back(entry::list_type());
```
arvidn: when using ``emplace_back()``, I don't think you need to pass a default-initialized object. By not passing any arguments, it should just default initialize the ``list_type``, right?

aldenml: yes, but below, it needs to be `entry::list_type`, can't be a generic default one (unless the default is a list, or...is it?)

src/write_resume_data.cpp
```diff
@@ -232,7 +232,7 @@ namespace libtorrent {
 			// write file priorities
 			entry::list_type& prio = ret["file_priority"].list();
```
arvidn: the more significant improvement here would probably be a ``prio.reserve()`` call

aldenml: true

src/write_resume_data.cpp
arvidn: oh, right

arvidn:
lgtm

PR #2234, opened 2017-08-10
using separated sockets in netlink based enum_net_interfaces

aldenml:
Hi @ssiloti, I appreciate if you can take a look at this.

The main problem is that for some reason, if the same socket is used for `RTM_GETLINK` and `RTM_GETADDR` it fails to work the second time (the one with `RTM_GETADDR `) in android.

The second problem was that `nl_dump_request` can return a negative number (`-1`) but not necessarily with an `errno` (see deep down in `read_nl_sock`).

I tried to touch as less as possible to easy the review, but if this PR is right, maybe two functions are better in a later refactor.

### DIFF
--- a/src/enum_net.cpp
+++ b/src/enum_net.cpp
@@ -608,21 +608,22 @@ int _System __libsocket_sysctl(int* mib, u_int namelen, void *oldp, size_t *oldl
 			ret.push_back(wan);
 		}
 #elif TORRENT_USE_NETLINK
-		int sock = socket(PF_ROUTE, SOCK_DGRAM, NETLINK_ROUTE);
-		if (sock < 0)
-		{
-			ec = error_code(errno, system_category());
-			return ret;
-		}
 
 		std::uint32_t seq = 0;
 		std::map<int, ip_interface> link_info;
 
 		{
+			int sock = socket(PF_ROUTE, SOCK_DGRAM, NETLINK_ROUTE);
+			if (sock < 0)
+			{
+				ec = error_code(errno, system_category());
+				return ret;
+			}
+
 			char msg[NL_BUFSIZE] = {};
 			nlmsghdr* nl_msg = reinterpret_cast<nlmsghdr*>(msg);
 			int len = nl_dump_request(sock, RTM_GETLINK, seq++, AF_UNSPEC, msg, sizeof(ifinfomsg));
-			if (len < 0)
+			if (errno)
 			{
 				ec = error_code(errno, system_category());
 				close(sock);
@@ -635,7 +636,7 @@ int _System __libsocket_sysctl(int* mib, u_int namelen, void *oldp, size_t *oldl
 			// NLMSG_OK uses signed/unsigned compare in the same expression
 #pragma clang diagnostic ignored "-Wsign-compare"
 #endif
-			for (; NLMSG_OK(nl_msg, len); nl_msg = NLMSG_NEXT(nl_msg, len))
+			for (; len > 0 && NLMSG_OK(nl_msg, len); nl_msg = NLMSG_NEXT(nl_msg, len))
 			{
 				ip_interface iface;
 				int ifi_index = parse_nl_link(nl_msg, &iface);
@@ -644,13 +645,21 @@ int _System __libsocket_sysctl(int* mib, u_int namelen, void *oldp, size_t *oldl
 #ifdef __clang__
 #pragma clang diagnostic pop
 #endif
+			close(sock);
 		}
 
 		{
+			int sock = socket(PF_ROUTE, SOCK_DGRAM, NETLINK_ROUTE);
+			if (sock < 0)
+			{
+				ec = error_code(errno, system_category());
+				return ret;
+			}
+
 			char msg[NL_BUFSIZE] = {};
 			nlmsghdr* nl_msg = reinterpret_cast<nlmsghdr*>(msg);
 			int len = nl_dump_request(sock, RTM_GETADDR, seq++, AF_PACKET, msg, sizeof(ifaddrmsg));
-			if (len < 0)
+			if (errno)
 			{
 				ec = error_code(errno, system_category());
 				close(sock);
@@ -663,7 +672,7 @@ int _System __libsocket_sysctl(int* mib, u_int namelen, void *oldp, size_t *oldl
 			// NLMSG_OK uses signed/unsigned compare in the same expression
 #pragma clang diagnostic ignored "-Wsign-compare"
 #endif
-			for (; NLMSG_OK(nl_msg, len); nl_msg = NLMSG_NEXT(nl_msg, len))
+			for (; len > 0 && NLMSG_OK(nl_msg, len); nl_msg = NLMSG_NEXT(nl_msg, len))
 			{
 				ip_interface iface;
 				if (parse_nl_address(link_info, nl_msg, &iface)) ret.push_back(iface);
@@ -671,9 +680,8 @@ int _System __libsocket_sysctl(int* mib, u_int namelen, void *oldp, size_t *oldl
 #ifdef __clang__
 #pragma clang diagnostic pop
 #endif
+			close(sock);
 		}
-
-		close(sock);
 #elif TORRENT_USE_IFADDRS
 		int s = socket(AF_INET, SOCK_DGRAM, 0);
 		if (s < 0)
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -1943,6 +1943,12 @@ namespace {
 		{
 			expand_unspecified_address(ifs, eps);
 		}
+#ifndef TORRENT_DISABLE_LOGGING
+		if (ec)
+			session_log("failed to enum net interfaces: %s", ec.message().c_str());
+		else
+			session_log("enum net found %d interfaces", int(ifs.size()));
+#endif
 #endif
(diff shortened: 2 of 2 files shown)


ssiloti:
Is this is known bug in android? I get the feeling this is papering over a bug in our code rather than fixing it.

aldenml:
to be honest, I'm not an expert in `netlink` related stuff, I find hard to find the right constants for the right thing in the linux documentation, and android has a lot of limitations (and probably not documented) related to this due to security reasons.

I can't find any obvious bug in the current code

aldenml:
to your question: I don't know, it wasn't working, and it works with this PR.

ssiloti:
Can you at least describe how it is failing? Does sending the second request fail, or reading the response? If reading the response fails, was any data returned? What was that data? If it's a syscall failing, what's the error code?

aldenml:
at any moment a `syscall` fails, so no `errno` is set. The call with `RTM_GETLINK` fails to return any data (not luck with google), and after that, a call to `RTM_GETADDR` in the same socket fails to return data too.

If I comment the call with `RTM_GETLINK`, then `RTM_GETADDR` works perfect, that's the reason I use two sockets, because it seems that `RTM_GETLINK` is problematic in android (couldn't find why, I'm just assuming is a security issue).

ssiloti:
I'm not talking about errno, if a syscall fails it will return an error code as a negative number.

aldenml:
I didn't pin which one was returning a negative number, I was assuming that each one of them set the `errno` to a meaningful code. Crossing fingers you don't demand that :), do you?

ssiloti:
What I demand is that we have some level of understanding about this issue before we push a half-baked workaround. If `RTM_GETLINK` doesn't work at all on android then we'll have to fall back to `getifaddrs` there. Tomorrow I'll ask the resident android expert at BT Inc about this. I seem to remember him commenting once that netlink on android is "weird".

aldenml:
Thanks

btw `getifaddrs` requires API >= 24 since the latest NDK changes (and 16 is still popular), and iconf code path does not compile in android

ssiloti:
Unfortunately our android expert isn't an expert when it comes to netlink. I also can't find any clear information on the usability of `RTM_GETLINK` on android. It would really help to know where in the code the `RTM_GETLINK` request is failing and what error code is being generated if it's a failing syscall.

aldenml:
ok, I will try to get that information

aldenml:
Hi @ssiloti with https://github.com/arvidn/libtorrent/pull/2241 I get many instances of these log in android (different runs)

```
read_nl_sock, returning -1
read_nl_sock, NLMSG_OK(nl_hdr, read_len)=0
read_nl_sock, nl_hdr->nlmsg_type == NLMSG_ERROR is 0
enum_net_interfaces, RTM_GETLINK returned negative:-1, errno=0
```
but only in one instance I got one with the `errno` set to `EDEADLK`
```
read_nl_sock, returning -1
read_nl_sock, NLMSG_OK(nl_hdr, read_len)=0
read_nl_sock, nl_hdr->nlmsg_type == NLMSG_ERROR is 0
enum_net_interfaces, RTM_GETLINK returned negative:-1, errno=11
```
As you can see, it looks that `RTM_GETLINK` is problematic (with `RTM_GETADDR` and `RTM_GETROUTE` it works fine). The problem is that `NLMSG_OK(nl_hdr, read_len)` always yields `0` (but `nl_hdr->nlmsg_type` is not `NLMSG_ERROR`).

I doubt this will help to fully understand the problem, but still think that using separated sockets allows to one operation to safely fail and still get the `RTM_GETADDR` result. Waiting for your comments

ssiloti:
Ok, `NLMSG_OK` looks at the length of the received data and the message length in the header. What were the values of `read_len` and `nl_hdr->nlmsg_len`?

Also, isn't there a way to attach gdb or some other debugger to a process on android? I think that'd be easier than debugging by logging.

aldenml:
I will get those values and post again. Yes :), there is a way to debug native code in android, but other than in trivial cases, the platform is a real nightmare.

aldenml:
It is `read_nl_sock, read_len=0, nl_hdr->nlmsg_len=0`

ssiloti:
A zero return value from `recv` implies the socket was closed. That doesn't give us much to go on. I don't think closing the socket is a normal response to any kind of error. This smells like android specific kludgery. I'd like to try one last experiment before we abandon `RTM_GETLINK`. What happens if you change the address family of the dump request from `AF_UNSPEC` to `AF_INET6`?

ssiloti:
Curiously, I checked the implementation of `getifaddrs` in bionic (android's C standard library) and it issues an `RTM_GETLINK` request like we do. I wonder if what @aldenml is seeing is specific to his device?

arvidn:
@ssiloti does it use a fresh, temporary socket though?

Is the theory that the issue is that the block that reads the response from ``RTM_GETLINK`` doesn't fully exhaust the socket, and when it gets around to read the response from ``RTM_GETADDR`` there's some garbage at the beginning of the buffer causing it to fail?

if the implementation of ``getifaddrs()`` closes the socket when it's done, perhaps it also doesn't read the response correctly.

ssiloti:
Bionc uses the same socket for both the `RTM_GETLINK` and `RTM_GETADDR`.

The impression I got from @aldenml is that the failing sequence is:

1. send(RTM_GETLINK)
2. recv() returns zero bytes, presumably because the socket is closed
3. send(RTM_GETADDR) fails
4. recv() returns zero bytes

Is that right?

aldenml:
@ssiloti let my try in another device and also with the change you suggest.

The failing sequence is `1.` and `2.` only, because after that, the method returns and it doesn't try the part of `RTM_GETADDR` (so no `3.` and `4.` are executed).

I will also try re-read more than one time, for what I interpret in the documentation of `recv`, `0` could be returned even if the socket is not closed (am I wrong here?)

ssiloti:
The man page for [recv(2)](http://man7.org/linux/man-pages/man2/recv.2.html) mentions three cases where zero may be returned:

1. The socket has performed an orderly shutdown.
2. A zero length datagram was received.
3. The requested number of bytes is zero.

Case 3 is obviously not applicable here. Case 2 shouldn't be happening because the kernel shouldn't be generating zero length netlink datagrams. That leaves the first case.

That being said, it's still a good experiment to see what happens if you retry the recv call.

aldenml:
Well, never easy with android. Here are the results:
1) using `AF_INET6`, no difference.
2) "retry the recv", no difference.
3) Testing other devices (with the original code):
- Emulator with API 19: works perfect.
- Device with API 23: doesn't work
- Device with API 24: doesn't work and here, a device log is always posted:
```
W: type=1400 audit(0.0:804): avc: denied { read } for uid=10156 name="net" dev="sysfs" ino=19231 scontext=u:r:untrusted_app:s0:c512,c768 tcontext=u:object_r:sysfs:s0 tclass=dir permissive=0
```
Waiting for your comments

ssiloti:
It looks like that audit log line means the process tried to access `/sys/class/net` but was denied by SELinux. I wouldn't expect an `RTM_GETLINK` request to trigger access to `/sys`, are you sure the log line is associated with the `RTM_GETLINK` request?

aldenml:
@ssiloti I can confirm that in fact, that log is _not_ related to the `RTM_GETLINK` query. It was a line in other part of the application that triggers it (in my Java side).

Well, you tell me if you want other test of you have any conclusion :)

ssiloti:
Try with this code: https://github.com/ssiloti/libtorrent/tree/android-getlink-experiment

I copy-pasted the code from bionc and made minimal modifications to it. If this doesn't work I give up.

aldenml:
@ssiloti I just tested, unfortunately, it doesn't work either. You tell me if this PR needs something else.

aldenml:
btw, returning to the my first comment, if I tweak the code, to not return even if `RTM_GETLINK` return `-1` due to no data returned, then, reusing the same socket, `RTM_GETADDR` doesn't work. That was my main motivation to use different sockets for each query.

ssiloti:
We cannot just ignore the failing `RTM_GETLINK` request. We need to find an alternative method of getting the names of the interfaces. Perhaps we'd have better luck with `SIOCGIFNAME`. I'm willing to do the work to refactor the code if you don't want to spend more time on this.

aldenml:
I see. You tell me how I can help, first time I see `SIOCGIFNAME`. If this is a problem only in android, I think a different specific method for it is better.

aldenml:
OK, I see the path of `TORRENT_USE_IFCONF` is using `SIOCGIFCONF`, but it seems out of date and does not compile, looking at it.

aldenml:
@ssiloti I just tested in an emulator arm with API 25 and it works, so it's not a matter of permission (or restricted feature). It's probably a problem in OEM implementations (all my devices are Motorola).

I have another idea: keep this PR as is, but `link_info.find(int(addr_msg->ifa_index));` is `end()`, then use  `if_indextoname` to get the interface name, or is the `MTU` value critical? what do you think?

ssiloti:
I'd rather use `if_indextoname` all the time if `RTM_GETLINK` is not reliable. The MTU is not critical, I don't think anything even uses it right now.

aldenml:
Done here https://github.com/arvidn/libtorrent/pull/2253, waiting for your review

PR #2236, opened 2017-08-11
remove #include (doxygen)

pavel-pimenov:
before:
include-graph-before

after:
include-graph-after

### DIFF
--- a/include/libtorrent/aux_/ffs.hpp
+++ b/include/libtorrent/aux_/ffs.hpp
@@ -34,7 +34,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #define TORRENT_FFS_HPP_INCLUDE
 
 #include <cstdint>
-#include "libtorrent/export.hpp"
 #include "libtorrent/span.hpp"
 
 namespace libtorrent { namespace aux {
--- a/include/libtorrent/aux_/typed_span.hpp
+++ b/include/libtorrent/aux_/typed_span.hpp
@@ -33,11 +33,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #ifndef TORRENT_TYPED_SPAN_HPP
 #define TORRENT_TYPED_SPAN_HPP
 
-#include <type_traits>
-
 #include "libtorrent/units.hpp"
 #include "libtorrent/span.hpp"
-#include "libtorrent/assert.hpp"
 
 namespace libtorrent { namespace aux {
 
--- a/include/libtorrent/bitfield.hpp
+++ b/include/libtorrent/bitfield.hpp
@@ -40,7 +40,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/aux_/ffs.hpp"
 
 #include <cstring> // for memset and memcpy
-#include <cstdint> // uint32_t
 
 namespace libtorrent {
 
--- a/include/libtorrent/config.hpp
+++ b/include/libtorrent/config.hpp
@@ -37,7 +37,6 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #define _FILE_OFFSET_BITS 64
 
-#include <boost/config.hpp>
 #include <boost/asio/detail/config.hpp>
 
 #include "libtorrent/aux_/disable_warnings_pop.hpp"
--- a/include/libtorrent/peer_info.hpp
+++ b/include/libtorrent/peer_info.hpp
@@ -39,7 +39,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/config.hpp"
 #include "libtorrent/bitfield.hpp"
 #include "libtorrent/time.hpp"
-#include "libtorrent/units.hpp"
 #include "libtorrent/flags.hpp"
 
 namespace libtorrent {
--- a/include/libtorrent/sha1_hash.hpp
+++ b/include/libtorrent/sha1_hash.hpp
@@ -35,10 +35,8 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include <cctype>
 #include <algorithm>
-#include <string>
 #include <cstring>
 
-#include "libtorrent/config.hpp"
 #include "libtorrent/assert.hpp"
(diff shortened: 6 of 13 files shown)


src/session.cpp
```diff
@@ -34,8 +34,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/extensions/ut_pex.hpp"
 #include "libtorrent/extensions/ut_metadata.hpp"
 #include "libtorrent/extensions/smart_ban.hpp"
-#include "libtorrent/session.hpp"
```
aldenml: just a comment: I'm in favor of cleanup as much as possible, but in my opinion, the `.cpp` should include the corresponding `.hpp` if only for coding style.

arvidn: I had the same thought.

src/session_handle.cpp
```diff
@@ -30,8 +30,6 @@ POSSIBILITY OF SUCH DAMAGE.
 
 */
 
-#include "libtorrent/session_handle.hpp"
```
aldenml: same here

include/libtorrent/aux_/typed_span.hpp
```diff
@@ -33,11 +33,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #ifndef TORRENT_TYPED_SPAN_HPP
 #define TORRENT_TYPED_SPAN_HPP
 
-#include <type_traits>
```
arvidn: this header is required for ``std::enable_if``

include/libtorrent/aux_/typed_span.hpp
```diff
@@ -33,11 +33,8 @@ POSSIBILITY OF SUCH DAMAGE.
...
-
 #include "libtorrent/units.hpp"
 #include "libtorrent/span.hpp"
-#include "libtorrent/assert.hpp"
```
arvidn: this header is required for ``TORRENT_ASSERT``

include/libtorrent/aux_/ffs.hpp
```diff
@@ -34,7 +34,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #define TORRENT_FFS_HPP_INCLUDE
 
 #include <cstdint>
-#include "libtorrent/export.hpp"
```
arvidn: this header is required for ``TORRENT_EXTRA_EXPORT``

include/libtorrent/bitfield.hpp
```diff
@@ -40,7 +40,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/aux_/ffs.hpp"
 
 #include <cstring> // for memset and memcpy
-#include <cstdint> // uint32_t
```
arvidn: this header is required for ``std::uint32_t``

include/libtorrent/config.hpp
```diff
@@ -37,7 +37,6 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #define _FILE_OFFSET_BITS 64
 
-#include <boost/config.hpp>
```
arvidn: I believe this is required for ``BOOST_MSVC``

include/libtorrent/peer_info.hpp
```diff
@@ -39,7 +39,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/config.hpp"
 #include "libtorrent/bitfield.hpp"
 #include "libtorrent/time.hpp"
-#include "libtorrent/units.hpp"
```
arvidn: I believe this is required for ``piece_index_t``

include/libtorrent/sha1_hash.hpp
```diff
@@ -35,10 +35,8 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include <cctype>
 #include <algorithm>
-#include <string>
```
arvidn: this is required for ``std::string``

include/libtorrent/sha1_hash.hpp
```diff
@@ -35,10 +35,8 @@ POSSIBILITY OF SUCH DAMAGE.
...
 #include <cstring>
 
-#include "libtorrent/config.hpp"
```
arvidn: this is required for ``TORRENT_USE_IOSTREAM`` and ``TORRENT_EXTRA_EXPORT`` and probably a bunch of other macros

arvidn:
I didn't go through all of the changes, but it seems this breaks header independency. headers should be self contained, as in, they should not yield any errors if they are the first to be included in a translation unit.

include/libtorrent/aux_/typed_span.hpp
pavel-pimenov: libtorrent/span.hpp
...
#include <type_traits> 

default

include/libtorrent/aux_/typed_span.hpp
pavel-pimenov: span.hpp :
...
#include "libtorrent/assert.hpp"

include/libtorrent/aux_/ffs.hpp
pavel-pimenov: default

include/libtorrent/bitfield.hpp
pavel-pimenov: default

include/libtorrent/config.hpp
pavel-pimenov: default

include/libtorrent/peer_info.hpp
pavel-pimenov: default

include/libtorrent/sha1_hash.hpp
pavel-pimenov: default

include/libtorrent/aux_/ffs.hpp
arvidn: sure, there are transitive dependencies on that header elsewhere, that doesn't make it a good idea to depend on them. This is what include guards are for.

PR #2243, opened 2017-08-13
python binding fixes

arvidn:
upgrade travis to trusty and newer xcode. python binding fixes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: cpp
+dist: trusty
 
 matrix:
   fast_finish: true
@@ -16,8 +17,7 @@ matrix:
     - os: osx
       osx_image: xcode7.3
       env: variant=test_barebones tools=1 examples=1 toolset=darwin python=1
-    - dist: trusty
-      env: arch=arm toolset=gcc-arm
+    - env: arch=arm toolset=gcc-arm
 
 notifications:
   email: false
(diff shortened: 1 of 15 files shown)


.travis.yml
```diff
@@ -41,16 +41,21 @@ cache:
...
-    - libboost1.55-all-dev
-    - libboost1.55-tools-dev
     - python2.7-dev
     - g++-5
 
 before_install:
-
+  # on linux, pull down a recent version of boost and build boost-build
+  - 'if [[ $TRAVIS_OS_NAME == "linux" ]]; then
+    wget https://dl.bintray.com/boostorg/release/1.64.0/source/boost_1_64_0.zip;
```
aldenml: are you willing to increase the minimal boost version? is there something you need in particular?

.travis.yml
```diff
@@ -8,16 +9,15 @@ matrix:
     - env: variant=test_release coverage=1 tests=1 toolset=gcc-coverage python=1
     - env: autotools=1 toolset=gcc
     - os: osx
-      osx_image: xcode7.3
+      osx_image: xcode8.3
```
aldenml: testing with the minimal version of Xcode is useful, unless you want to stay always with the latest version, is that the case?

arvidn: that's a fair point

.travis.yml
arvidn: I don't need it yet, but at some point I will need move support in boost::optional. perhaps I shouldn't bump boost version prematurely

PR #2246, opened 2017-08-14
making disk_buffer_pool::set_settings virtual and block_cache final

aldenml:
I wonder if there is a reason not to

### DIFF
--- a/include/libtorrent/disk_buffer_pool.hpp
+++ b/include/libtorrent/disk_buffer_pool.hpp
@@ -68,7 +68,7 @@ namespace libtorrent {
 			, std::function<void()> const& trigger_trim);
 		disk_buffer_pool(disk_buffer_pool const&) = delete;
 		disk_buffer_pool& operator=(disk_buffer_pool const&) = delete;
-		~disk_buffer_pool();
+		virtual ~disk_buffer_pool();
 
 #if TORRENT_USE_ASSERTS
 		bool is_disk_buffer(char* buffer
@@ -96,7 +96,7 @@ namespace libtorrent {
 		}
 		int num_to_evict(int num_needed = 0);
 
-		void set_settings(aux::session_settings const& sett);
+		virtual void set_settings(aux::session_settings const& sett);
 
 	protected:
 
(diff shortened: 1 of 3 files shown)


include/libtorrent/disk_buffer_pool.hpp
```diff
@@ -96,7 +96,7 @@ namespace libtorrent {
 		}
 		int num_to_evict(int num_needed = 0);
 
-		void set_settings(aux::session_settings const& sett);
```
arvidn: what does it buy you to make this virtual?

aldenml: I personally prefer `virtual` on methods that are overridden, otherwise it can be tricky to infer the right "polymorphic" behavior if a variable is of the base type.

arvidn: I see. this isn't really a polymorphic type in that sense. but I see what you mean. Both functions are called ``set_settings``.

include/libtorrent/disk_buffer_pool.hpp
```diff
@@ -68,7 +68,7 @@ namespace libtorrent {
 			, std::function<void()> const& trigger_trim);
 		disk_buffer_pool(disk_buffer_pool const&) = delete;
 		disk_buffer_pool& operator=(disk_buffer_pool const&) = delete;
-		~disk_buffer_pool();
+		virtual ~disk_buffer_pool();
```
arvidn: I think an alternative here is to make the destructor protected

include/libtorrent/disk_buffer_pool.hpp
arvidn: I think I might prefer renaming ``disk_buffer_pool::set_settings`` to something else instead. Say, ``disk_buffer_pool::set_cache_size()`` or something like that

aldenml: Ok, I will change the name

include/libtorrent/disk_buffer_pool.hpp
aldenml: OK

aldenml:
thinking about this again, I will close this for now, I don't want to break anything related

PR #2248, opened 2017-08-14
minor trivial code optimization

aldenml:


src/http_tracker_connection.cpp
```diff
@@ -586,10 +586,10 @@ namespace libtorrent {
 		if (ip_ent)
 		{
 			char const* p = ip_ent.string_ptr();
-			if (ip_ent.string_length() == int(address_v4::bytes_type().size()))
+			if (ip_ent.string_length() == std::tuple_size<address_v4::bytes_type>::value)
 				resp.external_ip = detail::read_v4_address(p);
 #if TORRENT_USE_IPV6
-			else if (ip_ent.string_length() == int(address_v6::bytes_type().size()))
+			else if (ip_ent.string_length() == std::tuple_size<address_v6::bytes_type>::value)
```
arvidn: I thought these types were ``boost::array`` or ``std::array``, not tuples. How come this works?

aldenml: it's `std::array` because of the `c++11`, now, the right way to get the `size` at compile time, is with the specialization of `std::tuple_size` for array. Don't blame me for the name selection :D (http://en.cppreference.com/w/cpp/container/array/tuple_size)

arvidn:
lgtm

PR #2253, opened 2017-08-16
using if_indextoname instead of RTM_GETLINK to fill the interface name in enum_net_interfaces

aldenml:


src/enum_net.cpp
```diff
@@ -408,14 +376,11 @@ namespace {
...
-		auto ifi_info = link_info.find(int(addr_msg->ifa_index));
-		if (ifi_info != link_info.end())
+		if (ip_info->name[0] == '\0')
 		{
-			ip_info->mtu = ifi_info->second.mtu;
 			// for some reason IPv6 entries don't include an IFA_LABEL attribute
 			// so get it from the link in that case
-			if (ip_info->name[0] == '\0')
-				strncpy(ip_info->name, ifi_info->second.name, sizeof(ip_info->name));
+			if_indextoname(addr_msg->ifa_index, ip_info->name);
```
ssiloti: There should be a static assert that `sizeof(ip_info->name) >= IF_NAMESIZE`.

aldenml:
Hi @arvidn, I wonder if there is any particular detail you want to address here

src/enum_net.cpp
```diff
@@ -406,14 +374,12 @@ namespace {
 #pragma clang diagnostic pop
 #endif
 
-		auto ifi_info = link_info.find(int(addr_msg->ifa_index));
-		if (ifi_info != link_info.end())
+		if (ip_info->name[0] == '\0')
 		{
-			ip_info->mtu = ifi_info->second.mtu;
 			// for some reason IPv6 entries don't include an IFA_LABEL attribute
 			// so get it from the link in that case
```
arvidn: this comment may be slightly outdated now. I also wonder if it would make sense to use this path to get the device name unconditionally, and not bother parsing the ``IFA_LABEL`` message.

include/libtorrent/enum_net.hpp
```diff
@@ -61,6 +61,7 @@ namespace libtorrent {
 		address interface_address;
 		address netmask;
 		char name[64];
+		// TODO: remove this?
 		int mtu;
```
arvidn: I suppose this was the only thing set from the link. I take it this isn't used anywhere?  believe it was used in the uTP implementation at some point, to initialize the PMTUd. Can you confirm it's not used anymore? (and if so, please remove it)

arvidn:
sorry for the delay. looks good overall, just some minor comments

src/enum_net.cpp
aldenml: Ok, I will remove the `IFA_LABEL` code

include/libtorrent/enum_net.hpp
aldenml: I will check and remove it if not used

src/enum_net.cpp
```diff
@@ -701,7 +622,6 @@ int _System __libsocket_sysctl(int* mib, u_int namelen, void *oldp, size_t *oldl
 
 					// ignore errors here. This is best-effort
 					ioctl(s, siocgifmtu, &req);
```
arvidn: the ``siocgifmtu`` ioctl request here is redundant now. ``req`` can be removed

arvidn: I can clear out the remaining MTU stuff. I feel bad for "feature creeping" your PR

arvidn:
some minor stuff here: https://github.com/arvidn/libtorrent/pull/2260

PR #2257, opened 2017-08-18
remove support for using a pool allocator for disk buffers

### LINKED ISSUE #2251: RC_1_1 uses too much memory
I am comparing RC_1_0 and RC_1_1 memory usage here.
Specifically when `cache_size` is set to `-1`, which has the same meaning on both branches but different results in real life.

To rule out bad usage from qBt I used the `example/client_test` program to download the same torrent with each branch. The only modification to the `.cpp` file was that I specifically hardcoded `-1` for `cache_size`.

For a 320MB file RC_1_0 ram usage hovers at around 12MB, while RC_1_1 goes above 250MB ( at about 70% progress).

This was done on Windows. But then I compiled under linux to use valgrind and memcheck. It has the same problem but it doesn't appear to be a memleak. Valgrind reports only a few KB of leaks. I'll attach the log at the end.

qbt+RC_1_1 gives the feel that it just forgets to empty the cache and keeps unnecessary space. Even when I remove the torrent from the session the memory footprint doesn't go down.

I can't believe that this is by design. I hope that you have overlooked something.
If it is by design, is there any setting that will give me previous behavior?

Valgrind log: https://gist.github.com/sledgehammer999/3adfdb4e9c3b353b0d8379f9a7e28160

ssiloti:
The pool allocator has known issues with not releasing memory when it should
and is of dubious benefit.

Fixes #2251

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -18,6 +18,7 @@
 	* added support for retrieval of DHT live nodes
 	* complete UNC path support
 	* add packets pool allocator
+	* remove disk buffer pool allocator
 	* fix last_upload and last_download overflow after 9 hours in past
 	* python binding add more add_torrent_params fields and an invalid key check
 	* introduce introduce distinct types for peer_class_t, piece_index_t and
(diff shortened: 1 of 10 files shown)


arvidn:
this should probably be mentioned in the ChangeLog

arvidn:
It's still used by ``high_performance_seed()`` https://travis-ci.org/arvidn/libtorrent/jobs/265821710#L1276

https://github.com/arvidn/libtorrent/blob/master/src/session.cpp#L244

arvidn:
lgtm

PR #2261, opened 2017-08-20
minor code refactor and cleanup in bt_peer_connection

aldenml:


src/bt_peer_connection.cpp
```diff
@@ -664,7 +660,7 @@ namespace {
 		aux::write_uint32(crypto_field, write_buf);
 		aux::write_uint16(pad_size, write_buf); // len (pad)
 
-		std::generate(write_buf.data(), write_buf.data() + pad_size, random_byte);
+		aux::random_bytes({write_buf.data(), pad_size});
```
arvidn: I think this span would be better written as: ``write_buf.first(pad_size)``

arvidn:
lgtm, just one minor nit

src/bt_peer_connection.cpp
aldenml: OK

aldenml: Hi @arvidn, do you mind taking a look at the two lines before this one? I see a couple of `aux::write_uint16`, but then, this overwrite whatever is there, is that right? or am I missing something?

arvidn: those writes updates the span to refer to the remaining bytes. Perhaps it's not the most intuitive interface, but I think it's at least an improvement over the older versions that just increments a pointer/iterator. i.e. ``write_buf`` is passed by non-const reference there.

aldenml: true, true, now I remember :) "the pointer is incremented under the hood"

PR #2262, opened 2017-08-20
fix that RAND_bytes is actually used under libcrypto, moved dev_random to compilation unit

aldenml:


ssiloti:
I'm not sure if this was done intentionally, but to me it makes sense to favor `/dev/random` over `RAND_bytes`. The former is the more direct interface and `RAND_bytes` likely uses `/dev/random` anyways.

aldenml:
It is my understanding that even if `/dev/random` is secure in most implementations, it doesn't need to be. Having `RAND_bytes` as an extra layer, allows for a more formal CSPRNG specification/implementation, that in the case of `openssl` I know it uses `/dev/random` as an input, but it applies some transformations.

I know this could sound like theoretical fantasy, but `openssl` allows for hardware providers and alternative implementations like `BoringSSL` might have something different.

(btw, not critical for me)

ssiloti:
`/dev/random` is very much intended to be cryptographically secure. I am quite sure there are no implementations which intentionally do not back it with a secure entropy source. Given the choice between trusting the kernel or trusting openssl to have a secure RNG, I'll take the kernel any day.

aldenml:
jeje, well, interesting...I pretty much prefer to use third party libraries for crypto routines in general, specially if I embed the desired library version in the binaries, and don't have any reason to trust the user installed OS kernel more.

ssiloti:
Userspace code has very limited ability to collect entropy other than by getting it from the kernel. The output of a PRNG is only as good as the entropy that's fed to it, so layering openssl over the kernel's RNG doesn't really buy you much.

aldenml:
good point!, I would like to read some opinion from @arvidn

arvidn:
I don't think I really can contribute much here. The factors not mentioned so far that I can think of are:

1. libtorrent makes some half-assed attempts to adding entropy to openssl's pool based on timestamps of receiving incoming connections
2. ``/dev/random`` may block, whereas I would imagine OpenSSL degrades instead of blocking.

arvidn:
@aldenml is it possible that openssl is more efficient on embedded systems or phones? Perhaps it batches the system calls to refill entropy. I don't have a strong opinion here, but it also doesn't sound like there's a strong reason to make this change.

aldenml:
No, I don't think it's more efficient and I don't think this is necessary either, closing and moving on

PR #2268, opened 2017-08-23
wrap destination string in i2p_peer in string_ptr handler

arvidn:
and minor cleanup

### DIFF
--- a/include/libtorrent/torrent_peer.hpp
+++ b/include/libtorrent/torrent_peer.hpp
@@ -37,6 +37,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/address.hpp"
 #include "libtorrent/socket.hpp"
 #include "libtorrent/peer_info.hpp" // for peer_source_flags_t
+#include "libtorrent/string_util.hpp" // for string_ptr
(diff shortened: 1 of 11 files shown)


include/libtorrent/torrent_peer.hpp
```diff
@@ -213,12 +215,13 @@ namespace libtorrent {
...
-		~i2p_peer();
-		i2p_peer& operator=(i2p_peer const&);
+		i2p_peer(string_view dst, bool connectable, peer_source_flags_t src);
+		i2p_peer(i2p_peer const&) = delete;
+		i2p_peer& operator=(i2p_peer const&) = delete;
+		i2p_peer(i2p_peer&&) = default;
+		i2p_peer& operator=(i2p_peer&&) = default;
 
-		char* destination;
+		string_ptr destination;
```
aldenml: question: what's the practical benefit of having this vs `std::string`? (since `string_ptr` copy the string)

arvidn:
I should put ``string_ptr`` in its own header and add a test for it

include/libtorrent/torrent_peer.hpp
aldenml: I see

PR #2273, opened 2017-08-25
cmake: build examples in the same solution as libtorrent

ssiloti:
Keep the search script for reference


arvidn:
lgtm. it would be great if travis could verify cmake builds work too :)

PR #2274, opened 2017-08-26
remove session pool allocator and use the buffer class instead

arvidn:


include/libtorrent/disk_buffer_holder.hpp
```diff
@@ -86,6 +86,7 @@ namespace libtorrent {
 		char* release() noexcept;
 
 		// return a pointer to the held buffer
+		char* data() const noexcept { return m_buf; }
 		char* get() const noexcept { return m_buf; }
```
aldenml: any reason to keep `get()`?

arvidn: the disk cache still uses ``get()`` and I thought it might not be worth changing all of its uses

include/libtorrent/aux_/non_owning_handle.hpp
```diff
@@ -1,49 +0,0 @@
...
-
-*/
-
-#ifndef TORRENT_NON_OWNING_HANDLE_HPP_INCLUDED
-#define TORRENT_NON_OWNING_HANDLE_HPP_INCLUDED
-
-namespace libtorrent { namespace aux {
-
-	// internal
-	struct non_owning_handle
```
arvidn: this is like a ``span``, but less sophisticated.

include/libtorrent/disk_buffer_holder.hpp
```diff
@@ -109,6 +110,8 @@ namespace libtorrent {
 		// buffer
 		explicit operator bool() const noexcept { return m_buf != nullptr; }
 
+		std::size_t size() const { return 0x4000; }
```
arvidn: this constant should have a name

PR #2275, opened 2017-08-26
add span ctor from const containers

ssiloti:
This was found to be useful while working on implementing BEP 52.

### DIFF
--- a/include/libtorrent/span.hpp
+++ b/include/libtorrent/span.hpp
@@ -90,6 +90,14 @@ namespace aux {
 		span(Cont& c) // NOLINT
 			: m_ptr(c.data()), m_len(c.size()) {}
 
+		// allow construction from const containers if T is const
+		template <typename Cont
+			, typename U = typename std::remove_reference<decltype(*std::declval<Cont>().data())>::type
+			, typename = typename std::enable_if<aux::compatible_type<U, T>::value
+				&& std::is_const<T>::value>::type>
+		span(Cont const& c) // NOLINT
+			: m_ptr(c.data()), m_len(c.size()) {}
+
 		std::size_t size() const noexcept { return m_len; }
 		bool empty() const noexcept { return m_len == 0; }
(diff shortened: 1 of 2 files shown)


include/libtorrent/span.hpp
```diff
@@ -90,6 +90,14 @@ namespace aux {
 		span(Cont& c) // NOLINT
 			: m_ptr(c.data()), m_len(c.size()) {}
 
+		// allow construction from const containers if T is const
+		template <typename Cont
+			, typename U = typename std::remove_reference<decltype(*std::declval<Cont>().data())>::type
+			, typename = typename std::enable_if<aux::compatible_type<U, T>::value
+				&& std::is_const<T>::value>::type>
```
arvidn: Isn't this exactly the same condition as the overload above, except with the ``&& std::is_const<>`` added?
I don't really understand which cases require this overload but doesn't match the existing one. The intention with ``aux::compatible_type`` was to capture all the const conversions correctly.

arvidn: There's a ``test_span.cpp`` which exercises the cases I had thought of. Could you add the missing one(s) there too?

include/libtorrent/span.hpp
```diff
@@ -90,6 +90,15 @@ namespace aux {
 		span(Cont& c) // NOLINT
 			: m_ptr(c.data()), m_len(c.size()) {}
 
+		// allow construction from const containers if T is const
+		// this allows const spans to be constructed from a temporary container
+		template <typename Cont
+			, typename U = typename std::remove_reference<decltype(*std::declval<Cont>().data())>::type
+			, typename = typename std::enable_if<aux::compatible_type<U, T>::value
+				&& std::is_const<T>::value>::type>
+		span(Cont const& c) // NOLINT
```
arvidn: oh, now I see. this takes a ``const&``, that's the difference.

PR #2278, opened 2017-08-28
more use of std::tuple_size<> and empty()

aldenml:


include/libtorrent/socket_io.hpp
```diff
@@ -63,10 +63,10 @@ namespace detail {
 			TORRENT_UNUSED(p);
 #if TORRENT_USE_IPV6
 			if (p == Proto::v6())
-				return address_v6::bytes_type().size();
+				return std::tuple_size<address_v6::bytes_type>::value;
```
arvidn: does this make a difference in practice? Since ``size()`` is ``constexpr``, I would assume both versions generate the same code.

aldenml: I didn't check the generated code, but I will say yes, because in order to the compiler get the "constexpr" value of `.size()`, I believe the instance should be `constexpr` too, I don't see a reason why the compiler should assume that. I also doubt that `std::tuple_size` is there as a redundant feature.

arvidn: right, ``constexpr`` isn't really relevant. The fact that the function returns a constant and that the ``bytes_type`` object is never used, I would expect a decent optimizer to replace by a constant. In compile-time contexts you would need the metafunction (until more things are constexpr I suppose)

Issue #2140, opened 2017-07-10
Crash in block_cache


aldenml:
libtorrent version (or branch): `master`
platform/architecture: `android/arm` but possible `all`

I'm getting this crash:
```
native: pc 0000000000398950   (_ZN10libtorrent11block_cache11evict_pieceEPNS_18cached_piece_entryERNS_9tailqueueINS_11disk_io_jobEEENS0_13eviction_modeE+104)
native: pc 00000000003988c8  (_ZN10libtorrent11block_cache16maybe_free_pieceEPNS_18cached_piece_entryE+156)
native: pc 000000000024ffd8   (_ZN10libtorrent14disk_io_thread11flush_rangeEPNS_18cached_piece_entryEiiRNS_9tailqueueINS_11disk_io_jobEEERNSt6__ndk111unique_lockINS7_5mutexEEE+456)
native: pc 000000000024fc3c   (_ZN10libtorrent14disk_io_thread16try_flush_hashedEPNS_18cached_piece_entryEiRNS_9tailqueueINS_11disk_io_jobEEERNSt6__ndk111unique_lockINS7_5mutexEEE+1292)
native: pc 0000000000255c8c  (_ZN10libtorrent14disk_io_thread14do_flush_pieceEPNS_11disk_io_jobERNS_9tailqueueIS1_EE+88)
native: pc 000000000025168c  (_ZN10libtorrent14disk_io_thread11perform_jobEPNS_11disk_io_jobERNS_9tailqueueIS1_EE+136)
native: pc 00000000002568f0   (_ZN10libtorrent14disk_io_thread10thread_funERNS0_9job_queueERNS_19disk_io_thread_poolE+856)
native: pc 000000000025f600   (_ZNSt6__ndk114__thread_proxyINS_5tupleIJNS_10unique_ptrINS_15__thread_structENS_14default_deleteIS3_EEEEMN10libtorrent21pool_thread_interfaceEFvRNS7_19disk_io_thread_poolEN5boost4asio10io_service4workEEPS8_NS_17reference_wrapperIS9_EESE_EEEEEPvSL_+156)
native: pc 0000000000048677  /system/lib/libc.so (_ZL15__pthread_startPv+22)
native: pc 000000000001a411  /system/lib/libc.so (__start_thread+6)
```
(unfortunately, no more info)

I think what's happening is that in `disk_io_thread::flush_range:L799` it is:
```cpp
[...]

Issue #2089, opened 2017-06-19
crash disk_buffer_pool::free_multiple_buffers


pavel-pimenov:
libtorrent version (or branch): master
platform/architecture: x86
compiler and compiler version: vc++2017

https://drdump.com/Problem.aspx?ProblemID=304135&login=guest

flylinkdc!libtorrent::disk_buffer_pool::free_multiple_buffers+0xb1
flylinkdc!libtorrent::block_cache::evict_piece+0x168
flylinkdc!libtorrent::block_cache::maybe_free_piece+0x73
flylinkdc!libtorrent::disk_io_thread::flush_range+0x1bc
flylinkdc!libtorrent::disk_io_thread::try_flush_hashed+0x5cd
flylinkdc!libtorrent::disk_io_thread::do_flush_piece+0x83
flylinkdc!libtorrent::disk_io_thread::perform_job+0x95
flylinkdc!libtorrent::disk_io_thread::thread_fun+0x2f4
flylinkdc!libtorrent::disk_io_thread::job_queue::thread_fun+0x36
flylinkdc!std::_LaunchPad<std::unique_ptr<std::tuple<void (__thiscall libtorrent::pool_thread_interface::*)(libtorrent::disk_io_thread_pool &,boost::asio::io_service::work),libtorrent::pool_thread_interface *,std::reference_wrapper<libtorrent::disk_io_thread_pool>,boost::asio::io_service::work>,std::default_delete<std::tuple<void (__thiscall libtorrent::pool_thread_interface::*)(libtorrent::disk_io_thread_pool &,boost::asio::io_service::work),libtorrent::pool_thread_interface *,std::reference_wrapper<libtorrent::disk_io_thread_pool>,boost::asio::io_service::work> > > >::_Run+0x50
flylinkdc!std::_Pad::_Call_func+0x14
flylinkdc!wcsftime+0xf5
kernel32!BaseThreadInitThunk+0xe
ntdll!__RtlUserThreadStart+0x70
ntdll!_RtlUserThreadStart+0x1b

arvidn:
most of the useful information to understand what happened here will likely be in that minidump, and only you can look at that, since that requires the original binary and debug information.

Could you look around for hints and post any findings here?

aldenml:
is this crash still happening to you? I think it was solved with https://github.com/arvidn/libtorrent/pull/2440

PR #2282, opened 2017-08-29
hold an owning reference to storage objects in try_flush_write_blocks

ssiloti:
It is possible for all other references to a storage object to be
destroyed while try_flush_write_blocks is running. If the storage is
destroyed then find_piece will crash when trying to re-aquire the
shared_ptr. To prevent this, keep the storage alive by holding a
shared_ptr to it in try_flush_write_blocks.

Normally the fence job when stopping a torrent would prevent the storage
object from being destroyed until all flush jobs are complete.
try_flush_write_blocks can be run after every disk job though so it has
the potential to "stradle the fence".

If the associated torrent does get unloaded then it is expected that
find_piece will return NULL thus causing the entry to be ignored.

### DIFF
--- a/src/disk_io_thread.cpp
+++ b/src/disk_io_thread.cpp
@@ -899,22 +899,22 @@ namespace libtorrent
 		DLOG("try_flush_write_blocks: %d\n", num);
 
 		list_iterator<cached_piece_entry> range = m_disk_cache.write_lru_pieces();
-		std::vector<std::pair<piece_manager*, int> > pieces;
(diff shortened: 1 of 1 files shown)


aldenml:
do you think that this could be the root of weird memory crashes related to the storage IO? (there are at least two of them in the issues)

ssiloti:
I see #2140 and #2089 which I don't think are related to this because they involves a regular flush piece request. Did you have any others in mind?

aldenml:
they are the ones, I was just asking, because sometimes memory errors manifest in a strange ways

arvidn:
good catch!

PR #2285, opened 2017-08-29
straighten out some type-safety in port mapping enums

arvidn:


include/libtorrent/aux_/portmap.hpp
```diff
@@ -35,9 +35,8 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "libtorrent/config.hpp"
 
-namespace libtorrent { namespace aux {
-	// TODO: move this for a better place and integrate it with
-	// portmap error alerts
+namespace libtorrent {
+
```
aldenml: if this will be in the public API, it should be moved to a place out of `aux_`

src/session_impl.cpp
```diff
@@ -5415,46 +5415,44 @@ namespace {
...
-		bool find_udp_port_mapping(int transport, int mapping, std::shared_ptr<listen_socket_t> const& ls)
+		bool find_udp_port_mapping(portmap_transport const transport
+			, int mapping, std::shared_ptr<listen_socket_t> const& ls)
 		{
-			return ls->udp_port_mapping[transport] == mapping;
+			return ls->udp_port_mapping[static_cast<int>(transport)] == mapping;
 		}
 	}
 
 	// transport is 0 for NAT-PMP and 1 for UPnP
```
arvidn: this comment could be removed now too

include/libtorrent/aux_/portmap.hpp
arvidn: yeah, I was thinking about maybe getting away with having ``alert_types.hpp`` and ``session.hpp`` be the public way of getting to these (a bit like how ``windows.h`` is the way to get to all other header files on windows basically). But it's probably better to break them out into a separate header.

PR #2291, opened 2017-08-30
use a strong typedef for port mapping id

arvidn:


include/libtorrent/portmap.hpp
```diff
@@ -47,6 +51,9 @@ namespace libtorrent {
 		none, tcp, udp
 	};
 
+	// this type represents an index referring to a port mapping
+	using port_mapping_t = aux::strong_typedef<int, aux::port_mapping_tag>;
```
aldenml: and the suffering with swig begins :D

src/session_impl.cpp
```diff
@@ -6593,19 +6594,19 @@ namespace {
 		return m_upnp.get();
 	}
 
-	int session_impl::add_port_mapping(portmap_protocol const t
+	port_mapping_t session_impl::add_port_mapping(portmap_protocol const t
 		, int const external_port
 		, int const local_port)
 	{
-		int ret = 0;
+		port_mapping_t ret{-1};
```
aldenml: do you think this could be a reason for some subtle bug?

arvidn: I suppose so. if both natpmp and upnp are disabled and the user tries to issue add a mapping. I don't think it would have any severe consequences though.

PR #2292, opened 2017-08-31
ensure settings pack enums are 16bit long

gubatron:
This is just me being paranoid (and still not knowing enough about C++ compilers)

When I read these declarations on how setting_packs stores settings in vector of `pairs<enum_for_type, value_of_type>`

```
std::vector<std::pair<std::uint16_t, std::string>> m_strings;
std::vector<std::pair<std::uint16_t, int>> m_ints;
std::vector<std::pair<std::uint16_t, bool>> m_bools;
```
I wondered why the programmer assumed that enums are by default 16-bits long, what if you had just a few enums and wanted less bits to be used, say `std::uint8_t` for less than 255 possibilities.

This assumes that all compilers will use 16bits for enums, but perhaps there's a guy out there that somehow has his compiler set to build enums of 8bits. I wonder if this helps in any way, or if in the Makefiles the size of enums has been specified to be by default 16bits.

### DIFF
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -111,7 +111,7 @@ namespace libtorrent {
 
 		// setting names (indices) are 16 bits. The two most significant
 		// bits indicate what type the setting has. (string, int, bool)
-		enum type_bases
+		enum type_bases : std::uint16_t
 		{
 			string_type_base = 0x0000,
 			int_type_base =    0x4000,
@@ -120,7 +120,7 @@ namespace libtorrent {
 			index_mask =       0x3fff
 		};
 
-		enum string_types
+		enum string_types : std::uint16_t
 		{
 			// this is the client identification to the tracker. The recommended
 			// format of this string is: "ClientName/ClientVersion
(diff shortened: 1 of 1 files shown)


arvidn:
I'm not entirely sure why this fails, I suspect it has something to do with ``operator+`` promoting the type to ``int``.

https://travis-ci.org/arvidn/libtorrent/jobs/270260346#L4819

aldenml:
It's my understanding that the compiler has a lot of freedom to pick whatever integral type is suitable, as long as it is big enough.

I see the fact that internally the index is stored in a `std::uint16_t` as an implementation detail, but the public API uses `int`, maybe it's better to specify these enums to `int` and put some tests for the maximum.

gubatron:
I just read that in clang there's indeed 2 options on how to handle enum sizing
`fshort-enums` (as short as possible) `-fno-short-enums` (starting at 32bits).
They say going for short ones (8bits) can save memory but might end up performing slower because 
> narrow memory accesses can be less efficient than full register-width accesses.

I've to check what's up with travis, I had no issues building and forgot to check travis output here.

PR #2296, opened 2017-08-31
avoid port mapping of local addresses

aldenml:


ssiloti:
lgtm

src/session_impl.cpp
```diff
@@ -2169,6 +2169,10 @@ namespace {
 		void map_port(MapProtocol& m, ProtoType protocol, EndpointType const& ep
 			, int& map_handle)
 		{
+			address const addr = ep.address();
+			if (addr.is_v6() && is_local(addr))
```
arvidn: it's not obvious to me why the ``is_local`` constraints only applies to IPv6.
But if this is indeed v6 specific, please wrap it in ``#if TORRENT_USE_IPV6`` too

arvidn: also, it seems like it would be more robust to move this condition to after the ``delete_mapping()`` call, just to ensure we don't have a mapping we don't want to have

aldenml: OK, I will do that

aldenml: btw, this is the explanation copied from `dht_tracker`:

> with IPv4 the interface might be behind NAT so we can't skip them based on the scope of the local address

arvidn:
lgtm (although, that comment would be great to have in the code)

aldenml:
:) required for merge? I can add it in another general minor change

arvidn:
lgtm

PR #2302, opened 2017-09-02
add_file_borrow: use string_view instead of pointer, size parameters

ssiloti:


include/libtorrent/file_storage.hpp
```diff
@@ -268,7 +268,7 @@ namespace libtorrent {
 		// That is, the first path element of all files must be the same.
 		// This shared path element is also set to the name of the torrent. It
 		// can be changed by calling ``set_name``.
```
arvidn: I believe this is part of the public API, so there still needs to be a deprecated overload with the old signature, right?

PR #2316, opened 2017-09-08
select which DHT port to report based on the connection's local endpoint

ssiloti:


src/session_impl.cpp
```diff
@@ -2151,6 +2151,22 @@ namespace {
 #endif
 	}
 
+#ifndef TORRENT_DISABLE_DHT
+	int session_impl::external_udp_port(tcp::endpoint const& local_endpoint) const
+	{
+		auto ls = std::find_if(m_listen_sockets.begin(), m_listen_sockets.end()
+			, [&](std::shared_ptr<listen_socket_t> const& e)
+		{
+			return e->local_endpoint == local_endpoint;
```
arvidn: for outgoing peer connections, I would expect `local_endpoint` to have an arbitrary port in it. Would it be better to just compare addresses here?

ssiloti: Right. I thought even outgoing connections got bound to one of the listen ports, but after double checking I see that is not the case.

arvidn:
lgtm

PR #2321, opened 2017-09-09
Autotools: Check if std::chrono and std::random are available.

sledgehammer999:
If yes, don't even try to find their Boost equivalents.

This way you avoid installing dev packages for Boost chrono/random.

I tested only on linux.

**I wanted to fix bjam too**. However, for the life of me, I can't find a suitable solution. The Boost.Build docs seem horribly inadequate. And bjam/b2 syntax/capabilities don't seem so powerful after all.
Please tell me that you can find a solution to this.
On Windows with msvc2017, **b2** copies boost.chrono and boost.random in the installation dir even though they aren't really needed. MSVC2017 has c++11 support always on.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -123,17 +123,27 @@ AX_BOOST_SYSTEM()
 AS_IF([test -z "$BOOST_SYSTEM_LIB"],
       [AC_MSG_ERROR(Boost.System library not found. Try using --with-boost-system=lib)])
 
-AX_BOOST_CHRONO()
-AS_IF([test -z "$BOOST_CHRONO_LIB"],
-      [AC_MSG_ERROR(Boost.Chrono library not found. Try using --with-boost-chrono=lib)])
+AC_LANG_PUSH([C++])
+AC_CHECK_HEADERS([chrono])
+AC_CHECK_HEADERS([random])
+AC_LANG_POP
 
-AX_BOOST_RANDOM()
-AS_IF([test -z "$BOOST_RANDOM_LIB"],
-      [AC_MSG_ERROR(Boost.Random library not found. Try using --with-boost-random=lib)])
+AS_IF([test "x$ac_cv_header_chrono" = "xno"],
+      [AX_BOOST_CHRONO()
+       AS_IF([test -z "$BOOST_CHRONO_LIB"],
+             [AC_MSG_ERROR(Boost.Chrono library not found. Try using --with-boost-chrono=lib)])
+       LIBS="$BOOST_CHRONO_LIB $LIBS"
+      ])
+
+AS_IF([test "x$ac_cv_header_random" = "xno"],
+      [AX_BOOST_RANDOM()
+       AS_IF([test -z "$BOOST_RANDOM_LIB"],
+             [AC_MSG_ERROR(Boost.Random library not found. Try using --with-boost-random=lib)])
+       LIBS="$BOOST_RANDOM_LIB $LIBS"
+      ])
 
 CPPFLAGS="$BOOST_CPPFLAGS $CPPFLAGS"
 LDFLAGS="$BOOST_LDFLAGS $LDFLAGS"
-LIBS="$BOOST_CHRONO_LIB $BOOST_RANDOM_LIB $LIBS"
 
 ###############################################################################
 # Checking for functions and other stuffs
@@ -642,9 +652,19 @@ Boost libraries:
   CPPFlags:             ${BOOST_CPPFLAGS}
   LDFlags:              ${BOOST_LDFLAGS}
   boost.system:         ${BOOST_SYSTEM_LIB}
+END
+
+AS_IF([test "x$ac_cv_header_chrono" = "xno"], [
+cat >> config.report << END
   boost.chrono:         ${BOOST_CHRONO_LIB}
+END
+])
+
+AS_IF([test "x$ac_cv_header_random" = "xno"], [
+cat >> config.report << END
   boost.random:         ${BOOST_RANDOM_LIB}
 END
+])
 
 AS_IF([test "x$ARG_ENABLE_PYTHON_BINDING" = "xyes"], [
 cat >> config.report << END


sledgehammer999:
In case you missed it, this targets RC_1_1.

arvidn:
> Please tell me that you can find a solution to this.

What's the problem exactly?

Looking at the places where libtorrent use boost.chrono and boost.random, it seems to be conditioned by ``BOOST_ASIO_HAS_STD_CHRONO`` and ``BOOST_NO_CXX11_HDR_RANDOM``.

I take it you're suggesting detecting whether or not those are defined at compile time and determine whether or not the dependency on boost.chrono and boost.random is necessary or not.

I find that a bit risky. In your patch you assume that the presence of ``<chrono>`` correlates exactly 1:1 with boost.asio defining ``BOOST_ASIO_HAS_STD_CHRONO``, which I have some doubts about being a safe assumption. Likewise about the ``<random>`` header. Unless boost config headers use [__has_include](http://en.cppreference.com/w/cpp/experimental/feature_test) that is. But I think that may be a bit too new.

Is the cost of the boost.random and boost.chrono dependencies that great? The purpose of boost is to provide exactly this abstraction, to provide these APIs unconditionally on whether the compiler/system supports the standard APIs.

sledgehammer999:
Sorry for the following wall of text. I want to discuss my observations and how I came to this solution. The solution might be naive and incomplete.

It all started when in qBittorrent we required c++11. After some commits travis-ci linux started erroring out during linking. Long story short: The ubuntu libtorrent wasn't compiled with c++11 enabled in the compiler while qBittorrent was. So qbittorrent was seeing the libtorrent headers differently than what libtorrent saw them. So the symbols qbittorrent expected weren't there in the to-be-linked library.
(IIRC that symbol was `libtorrent::address_v4::bytes_type`)
After providing custom packages for libtorrent which were built with `CXXFLAGS=-std=c++11` in the configure script everything started working again.
Also the end library when observed via `ldd` didn't have dependencies on boost.chrono and boost.random anymore. However, the configure script still insisted to search for them.

That's on linux. On macOS it is a slightly different story. There the end library(observed via `otool -L`) still linked to boost.chrono and boost.random even though the compiler was in c++11 mode. But I have the hunch that they weren't actually used. Probably the macOS linker links all the libs you tell it to link even if they aren't used. However, I have very little knowledge of macOS.
However, after I apply this patch, the end library will not have dependencies on boost.chrono and boost. random on macOS with c++11 mode.

On Windows with b2 when I do in libtorrent `b2 --prefix=/some/path install`, it also copies the boost.chrono and boost.random libs unconditionally. Then if I delete those 2 libs and build qBittorrent it links without an error, so they are indeed useless in c++11 mode.

You should understand that this patch doesn't enforce c++11. If c++11 isn't enabled (eg -std=c++98) then the checks fail and it uses the boost equivalents. Since I couldn't find a specific test for c++11 availability I opted to check for those 2 headers which are only available in c++11 mode.

I truly don't know what is their relationship with boost.asio, but I do know that our [PPA](https://launchpad.net/~qbittorrent-team/+archive/ubuntu/qbittorrent-stable) worked this way as far back as Ubuntu 15.04 (Vivid Vervet)

Summing up: AFAIK this patch shouldn't affect compiling. At least on linux/windows, if c++11 is enabled nothing changes, it just stops requiring unused things.

PS: I think RC_1_0 has same problem. I'll provide patch, depending on what we do with this.

arvidn:
I think the fundamental problem is to have libtorrent headers alter their declarations based on the c++ version of the translation unit that included them. I kind of think it would be better to stick to boost for 1.1.x and have a clean switch-over to C++11 in 1.2

sledgehammer999:
>I kind of think it would be better to stick to boost for 1.1.x

So you're going to push changes to RC_1_1 that only use boost types?

I am tagging @glassez @Chocobo1 @evsh in case they have something to say about this.

sledgehammer999:
>So you're going to push changes to RC_1_1 that only use boost types?

Answering my own question: I think that it will not achieve much. An example: `libtorrent::address_v4::bytes_type`. It is ultimately defined in `boost/asio/detail/array.hpp` as:
```c++
#if defined(BOOST_ASIO_HAS_STD_ARRAY)
using std::array;
#else // defined(BOOST_ASIO_HAS_STD_ARRAY)
using boost::array;
#endif // defined(BOOST_ASIO_HAS_STD_ARRAY)
```

In this case, no matter which libs you require, the headers will change signatures depending on the C++ standard used by the includer.

evsh:
I would like to point out that we already made significant efforts to switch qBittorrent builds over  to C++11, enforcing the same move for the libtorrent build/package. And yet there same bug reports coming. I hardly imagine we now would ask people to change back.

sledgehammer999:
Any chance in merging this?

arvidn:
I'm sorry I've been slow at responding to this. I've been procrastinating and putting it off, since it's a difficult problem.

I recognize there will be/are difficulties either way. I wish there was a clear solution. In one way, I think it would be best if the libtorrent API did not depend on whether it was built with C++11 or not, but on the other hand, it would be quite limiting to not be able to use the new standard types.

One way forward could be to have explicit build options in libtorrent to enable std.chrono and std.random. And recognize that these would be ABI breaking options. The problem there is that people in general have a very hard time with such build options, because clients would then have to use something like pkg-config to know which flags to use when linking against libtorrent. On windows there isn't even a tool like pkg-config afaik.

But I'm willing to accept that it's worth it in this case. Perhaps the default could be to use boost (to have the ABI stay backwards compatible). I suppose boost asio has a similar issue, that its ABI (for ``address_v4::byte_type`` specifically) depend on whether it's built with C++11 or not. Perhaps libtorrent build options would need to forcefully configure boost as well, to use the same C++ version.

sledgehammer999:
I agree with your concerns but just in case I'll reiterate something about this PR.
This PR is useful in 2 situations:
1. `./configure CXXFLAGS=-std=c++11` is used
2. gcc6+ is used where c++11 is enabled by default.

In those cases the boost dependencies are wrong and won't be used by the compiler anyway. So this PR lets you have less build dependencies if you go the c++11 route. Also on macOS the linker doesn't detect if the generated lib actually needs boost.chrono/random and references them anyway. So on macOS you have less deployment dependencies too.

In short, this PR in a sense automates some manual work if you build in c++11 mode. It doesn't change internals.

**Unless you wanted to somehow patch RC_1_1 and not have changing ABI**

arvidn:
my concern is not limited to just this change, but perhaps primarily the previous change (which was merged) to use std.random instead of boost.random, when building with C++11. i think they are connected

sledgehammer999:
>but perhaps primarily the previous change (which was merged) to use std.random instead of boost.random, when building with C++11

+1
Now I understood you completely.

arvidn:
I'll merge this for 1.1.6. The situation will be a bit better in master (1.2) since there will be fewer compiled boost dependencies. The problem of reverting the older patch, that use C++11 when available, is that it also may break the ABI. The main problem of leaving things as they are is that the libtorrent ABI now depends on a subtle (and implicitly detected) feature of the build environment, which is like asking for trouble.

arvidn:
I'm going to have to revert this. this actually breaks building on mac (at least with one of my setups). The reason is that boost.asio's configuration header that detecs whether std::chrono is available, is different than the configuration script's. In my case, boost.asio does not think std::chrono is available, but the configuration script does. the result is not linking against boost.chrono, but still using it.

sledgehammer999:
In your case, did you remember to also enable c++11 in the app linking libtorrent/boost ?
For me, boost from `homebrew` behaves as expected with this patch.

arvidn:
well, it was all part of ``./configure`` and ``make check``. I suppose it's possible the ``-std=c++11`` flag would only apply to the library, but not the tests or examples, but I don't believe that's the case.

sledgehammer999:
Its weird for libtorrent to "see std::chrono" during its configuration+building and then not see it. AFAIK, boost.asio is a header lib so its types should be evaluated/resolved each time compiling/linking takes place.
Anyway, I am going to be applying this manually for my macOS and linux(PPA) builds.

arvidn:
but boost.asio infers whether or not the compiler has chrono or not, whereas the configure script literally looks for the <chrono> header. Had boost.asio used the C++17 feature ``__has_include(<chrono>)`` I would have expected them to always agree, but now it's just a bunch of heuristics of compiler and standard library version checks.

sledgehammer999:
>but boost.asio infers whether or not the compiler has chrono or not, whereas the configure script literally looks for the header.

I am under the impression that if the compiler is run in c++98 mode boost.asio won't use std::chrono. And I used the same logic for the configure script.
In any case I am not pushing further for this. But in case you wanted to look at it again, make sure that your final program/tests/examples are also build in c++11 mode.

sledgehammer999:
>And I used the same logic for the configure script.

I realized that this may not be obvious:
```
AC_CHECK_HEADERS([chrono])
```
will fail if g++ is in c++98 mode. Behind the scenes autotools compiles a small test program that includes the header. So in c++98 mode the header isn't available and compilation fails and then AC_CHECK fails too.

arvidn:
I don't think supporting std.chrono is completely aligned with supporting C++11 or not. typically features like these are introduced incrementally.

PR #2322, opened 2017-09-09
minor cleanup and code refactor in web_peer_connection.cpp

aldenml:


src/web_peer_connection.cpp
```diff
@@ -514,7 +514,7 @@ namespace {
 				ec = errors::no_content_length;
 			}
 		}
-		return std::tuple<std::int64_t, std::int64_t>(range_start, range_end);
+		return {range_start, range_end};
```
arvidn: this is actually not supported by C++11. It has been fixed in later versions, but I'm trying to stick to C++11 for now. libc++ supports it as an extension.

arvidn: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4387.html

arvidn:
lgtm apart from comment

src/web_peer_connection.cpp
aldenml: right, now I remember that, read about it a long time ago

arvidn:
lgtm

PR #2325, opened 2017-09-09
use the official boost.config header

sledgehammer999:
Just to have RC_1_0 work with boost 1.65+

(cherry-picked from RC_1_1)


arvidn:
lgtm

PR #2334, opened 2017-09-12
fix Windows build

ssiloti:


src/file.cpp
```diff
@@ -250,10 +276,6 @@ namespace libtorrent {
 #endif
 	}
 
-#ifndef INVALID_HANDLE_VALUE
-#define INVALID_HANDLE_VALUE (-1)
-#endif
```
arvidn: I think this might be used for non-windows builds.

arvidn:
it also looks like the terms in between ``|`` needs to be parenthesized: https://github.com/arvidn/libtorrent/pull/2334/files#diff-216c6cad716c26d79607e6e3c7effa0fR96

arvidn:
lgtm

arvidn:
the failure looks like it might be related to the ``lock_files`` flag, preventing concurrent access.

ssiloti:
At least one of the problems is that because we release the mutex in `file_view file_view_pool::open_file` while opening the file it, is possible for multiple threads to attempt to open the same file. Obviously only the first will succeed.

ssiloti:
Even if we serialize opening files somehow, file locking still conflicts with the path which swaps a read only mapping for a read/write one. I don't see any way around this without making the swap non-atomic. Maybe file locking isn't worth the trouble?

arvidn:
yeah, maybe not

arvidn:
so, I guess file locking is just broken in master right now then.

ssiloti:
On master `file_pool::open_file` holds its lock in the regular open file path so that's not a problem, but the read-only -> read/write upgrade path is broken if `lock_files` is enabled.

ssiloti:
Ahhh, on master there is a fallback which re-opens the file without locking if the initial open fails. That would explain what we're not seeing failures.

arvidn:
in order to land this patch, how about we just disable ``lock_files`` by default?

arvidn:
looks like a mingw file handle mixup. https://ci.appveyor.com/project/arvidn/libtorrent/build/7241/job/f26a2yr90fb3p72o#L130
(I'm sorry I made a mess)

PR #2340, opened 2017-09-15
add try_evict_blocks

pavel-pimenov:


arvidn:
I'm not sure I feel this duplication is worth factoring out. Mostly because I don't intend to develop the block cache/disk cache further in master. I intend to remove it in the major release after 1.2

PR #2341, opened 2017-09-15
removed leading :: in posix_fadvise to allow for fix in old android API build

aldenml:
Hi @arvidn, this little change makes all the difference in the workaround I need to do for android API 16

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -651,7 +651,7 @@ static_assert(!(open_mode::sparse & open_mode::attribute_mask), "internal flags 
 		if ((mode & open_mode::random_access))
 		{
 			// disable read-ahead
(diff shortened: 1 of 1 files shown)


arvidn:
what's the issue on android? is this a macro there?

arvidn:
I think it might make sense to leave a comment too, that there's something fishy right there on android

aldenml:
No, is that `POSIX_FADV_RANDOM` is defined, but the function is not there, I redefine it in the compiler command line to nullify the call. This is due to bugs in the headers.

arvidn:
thanks. lgtm

arvidn:
the arm failures here are transient, right?

PR #2344, opened 2017-09-16
deprecate lock_files

ssiloti:
File locking interacts poorly with the path to replace a read only file handle
with a read/write one. For performance reasons the old handle is not closed
until the new one is put in place. With file locking this obviously cannot
work because the first handle is holding a lock on the file. As a result,
file_pool::open_file fails in this case. Even if we dropped the reference to
the read only file handle before attempting to re-open it, the open could still
fail if another thread is holding a reference to the read only handle.

There doesn't seem to be a good way around this. The simple solution would be
to always open files in read/write mode, but that has obvious safety downsides.
Any other solution would significantly complicate the file pool code.
Deprecating file locking seems to be the least bad option.

### DIFF
--- a/bindings/python/src/torrent_handle.cpp
+++ b/bindings/python/src/torrent_handle.cpp
@@ -596,7 +596,9 @@ void bind_torrent_handle()
     s.attr("sparse") = file_open_mode::sparse;
     s.attr("no_atime") = file_open_mode::no_atime;
     s.attr("random_access") = file_open_mode::random_access;
-    s.attr("locked") = file_open_mode::locked;
+#ifndef TORRENT_NO_DEPRECATE
+    s.attr("locked") = 0;
+#endif
     }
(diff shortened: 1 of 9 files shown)


arvidn:
yeah. always opening in read/write would prevent seeding from read-only media/mounts as well as seeding files that *other* processes have a write lock on. I agree, the problem of 3rd party processing corrupting files behind the back is probably not big enough to warrant these complications.

now, if only there was a way to have the *process* own the lock, and not have it tied to individual file handles..

arvidn:
https://travis-ci.org/arvidn/libtorrent/jobs/276332971#L457

arvidn:
and https://ci.appveyor.com/project/arvidn/libtorrent/build/7246/job/b2u90dffy9ig1796#L953

it's the python binding

PR #2346, opened 2017-09-17
use new instead of malloc

arvidn:


src/string_util.cpp
```diff
@@ -168,9 +168,9 @@ namespace libtorrent {
 	{
 		if (str == nullptr) return nullptr;
 		std::size_t const len = std::strlen(str);
-		char* tmp = static_cast<char*>(std::malloc(len + 1));
+		char* tmp = new char[len + 1];
 		if (tmp == nullptr) return nullptr;
```
aldenml: Hi @arvidn, I wanted to ask something because I already saw this refactor in other places, in order to this line `if (tmp == nullptr) return nullptr;` be meaningful, don't you need `new (std::nothrow)`?

arvidn: ah, yes. That's right. I was considering the semantic change in throwing an exception here and decided that it was likely a safer response. I forgot to remove the nullptr check though

PR #2348, opened 2017-09-17
remove redundant null check

arvidn:


aldenml:
The others two are in `file.cpp`, L789 and L807, `if (!buf) return false;` (if you want to add them in this commit)

arvidn:
thanks

PR #2351, opened 2017-09-19
add set_peer_class_filter method in python bindings

chaoaero:
Change-Id: I3fc300dcd2704e4b0b643bbfef29f067b0d8d1e5


arvidn:
great, thanks! would you mind making ``test.py`` exercise this function as well (and maybe some other peer class related functions)?

Otherwise it's likely that these functions get neglected (again), down the road.

arvidn:
if you're interested in getting this released sooner, I would also be willing to merge this into ``RC_1_1`` (which is the current stable branch)

chaoaero:
I just want the global rate limit to apply to all local peers in LAN.  After testing the python code snippets corresponding to [peer class](https://github.com/arvidn/libtorrent/blob/master/docs/manual.rst#peer-classes), you mentioned, I found that It only works when download rate less than 4.3MB.  While I want to set the rate limit to 100MB in LAN, but failed.

All the efforts such as set_local_download_rate_limit or anything else is in vain. Now I am trying to investigate tc-cgroup.

If I can only use libtorrent to achieve this goal, that could be better!

Any ideas?

Sincerely!

arvidn:
what happens if the download rate limit is > 4.3 MB/s?

arvidn:
rate limits are definitely meant to work for higher limits. 4.3 seems like an odd limit too. It looks a bit like it would be a ``uint32_t`` limit, but in that case it would be limited at 4.3 GB/s.
Although, I believe rate limits are multiplied by the order of 1000 milliseconds during calculations though, so that could explain 3 orders of magnitude lower limit. All that is supposed to be done with ``in64_t`` arithmetic though

arvidn:
which version of libtorrent are you testing this with?

chaoaero:
The release version libtorrent-1.1.4

chaoaero:
When the download rate limit exceed 4.3MB/s, the download speed will increase sharply. Nearly up to 300MB/s. It doesn't work at all.

chaoaero:
I use the client.py located in bindings/python. Which the download speed will be multiplied by 1000. So in order to gain download speed of 4.3MB/s, I offerred the option *-d 4300* .  You use int in the C++ code snippet is below, which doesn't matter
```
	void session_impl::set_local_download_rate_limit(int bytes_per_second)
	{
		INVARIANT_CHECK;
		settings_pack p;
		p.set_int(settings_pack::local_download_rate_limit, bytes_per_second);
		apply_settings_pack_impl(p);
	}
```

arvidn:
I found this, which may be related to the issue you see with download rate limits. https://github.com/arvidn/libtorrent/pull/2359/files

I would still be interested in a patch adding this missing python binding though

chaoaero:
That would be nice if you could merge the python patch.

chaoaero:
It would be nice that you could merge [2359](https://github.com/arvidn/libtorrent/pull/2359) and the python binding ASAP.

arvidn:
would you mind updating this PR with a test of this function in ``test.py``?

arvidn:
interestingly, I can reproduce the download issue in the python binding, but not in client_test. I'm suspecting there's an issue in the binding

chaoaero:
I will finish and re-submit it tomorrow. Thanks for your reminder

arvidn:
I'm not sure what I was seeing earlier trying to reproduce this. In current ``RC_1_1`` (granted, with the integer overflow fix) I cannot reproduce the download rate limit issue. I've tried 5MB/s and 6 MB/s.

I'm using this patch to rate limit local peers. Note that this is a deprecated function, and the way it *should* be done requires the peer class API to be exposed to python.

```
diff --git a/bindings/python/client.py b/bindings/python/client.py
index 994a3d3eb..cfcff59f0 100755
--- a/bindings/python/client.py
+++ b/bindings/python/client.py
@@ -215,6 +215,7 @@ def main():
         settings['proxy_port'] = options.proxy_host.split(':')[1]

     ses = lt.session(settings)
+    ses.set_local_download_rate_limit(int(options.max_download_rate))

     # map torrent_handle to torrent_status
     torrents = {}
```

chaoaero:
Hi @arvidn , RC_1_1 works to me. Thanks

While compiling the master branch, some error occurred. Hopefully, you can check the master branch again.

Which branch you use when releasing v1.14? It seems that that release tag doesn't work yet.

chaoaero:
I found that If I want to test the peer_class_filter, I need to expose more API such as get_peer_class_filter. Or use another way like: bringing a existing torrent into the test script, and continuously checking the download rate is below the threshold. While I don't have a valid public torrent to use. I have checked the torrents in test folder. They cannot satisfy my need. What's the proper way to test such kinds of issue?

arvidn:
I don't think the python test need to test the underlying functionality, really just the binding itself. say, could you set some settings and try to read them back and make sure that makes sense?

chaoaero:
get_peer_class_filter is not exposed to session_handler.cpp. So it calls for adding so many wrappers   of get_peer_class_filter to achieve this.

arvidn:
@chaoaero could you give this a try? https://github.com/arvidn/libtorrent/pull/2366

arvidn:
I'll close this PR since it's replaced by #2366

PR #2355, opened 2017-09-21
DHT nodes should only handle requests on their socket

ssiloti:
Go through the trouble of passing the listen socket through
session_impl::on_udp_packet because otherwise we'd have to match by comparing
a tcp::endpoint with a udp::endpoint which is surprisingly expensive.

### DIFF
--- a/include/libtorrent/kademlia/node.hpp
+++ b/include/libtorrent/kademlia/node.hpp
@@ -110,7 +110,7 @@ public:
 	void add_router_node(udp::endpoint const& router);
 
 	void unreachable(udp::endpoint const& ep);
-	void incoming(msg const& m);
+	void incoming(aux::listen_socket_handle const& s, msg const& m);
 
 #ifndef TORRENT_NO_DEPRECATE
 	int num_torrents() const { return int(m_storage.num_torrents()); }
@@ -241,6 +241,7 @@ private:
 public:
 	routing_table m_table;
 	rpc_manager m_rpc;
+	aux::listen_socket_handle const m_sock;
 
 private:
 
@@ -252,6 +253,8 @@ private:
 	};
 
 	static protocol_descriptor const& map_protocol_to_descriptor(udp protocol);
+
+	socket_manager* m_sock_man;
 
 	get_foreign_node_t m_get_foreign_node;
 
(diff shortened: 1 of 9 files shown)


src/kademlia/node.cpp
```diff
@@ -317,7 +317,8 @@ void node::incoming(msg const& m)
 			// responds to 'query' messages that it receives.
 			if (m_settings.read_only) break;
 
-			if (!native_address(m.addr)) break;
+			// only respond to requests if they're addressed to this node
```
arvidn: is there a good reason to only filter requests (``q``) and not responses (``r``) based on which socket we received it on?

test/test_dht.cpp
```diff
@@ -288,7 +288,7 @@ void send_dht_request(node& node, char const* msg, udp::endpoint const& ep
 	if (ec) std::printf("bdecode failed: %s\n", ec.message().c_str());
 
 	dht::msg m(decoded, ep);
-	node.incoming(m);
+	node.incoming(node.m_sock, m);
```
arvidn: is this the only reason ``m_sock`` was made ``public``?

include/libtorrent/kademlia/node.hpp
```diff
@@ -241,6 +241,7 @@ class TORRENT_EXTRA_EXPORT node
 public:
 	routing_table m_table;
 	rpc_manager m_rpc;
+	aux::listen_socket_handle const m_sock;
```
arvidn: I can't find a strong reason to make this public. perhaps I'm missing something.

include/libtorrent/kademlia/node.hpp
```diff
@@ -253,6 +254,8 @@ class TORRENT_EXTRA_EXPORT node
 
 	static protocol_descriptor const& map_protocol_to_descriptor(udp protocol);
 
+	socket_manager* m_sock_man;
```
arvidn: did this move just to order it among other pointers?

arvidn:
would it be a lot of trouble to set up a test that exercise this? a unit test might be simplest

src/kademlia/node.cpp
ssiloti: Nodes can send requests using a listen socket other than their own for cross-network bootstrap. They need to be able to receive the response.

test/test_dht.cpp
ssiloti: Yes. Having made `m_sock` `const` I didn't see any harm in doing so.

include/libtorrent/kademlia/node.hpp
ssiloti: I moved it up to keep it next to `m_sock` because they usually get accessed together.

PR #2358, opened 2017-09-21
fix issue where peers could be disconnected while iterating over the peer list in the torrent

arvidn:
Specifically while marking peers as not-interesting


aldenml:
is this still needed in `master`, or the deferred disconnect solves the problem there?

arvidn:
this is not needed in master. I'm hoping to merge soon so I don't forget that this should not make it in.

PR #2359, opened 2017-09-21
fix integer overflow in whole_pieces_threshold logic

arvidn:


arvidn:
@chaoaero did you have a chance to test this patch? does it make any difference to the download rate issue you experience?

chaoaero:
I build this patch with release 1.1.4, While the python client cannot run throughly. It seems that something might be wrong. 
I will verify it torrow morning again.

src/request_blocks.cpp
```diff
@@ -112,10 +112,12 @@ namespace libtorrent
...
 		{
-			int blocks_per_piece = t.torrent_file().piece_length() / t.block_size();
-			prefer_contiguous_blocks = c.statistics().download_payload_rate()
-				* t.settings().get_int(settings_pack::whole_pieces_threshold)
-				> t.torrent_file().piece_length() ? blocks_per_piece : 0;
+			int const blocks_per_piece = t.torrent_file().piece_length() / t.block_size();
+			prefer_contiguous_blocks
+				= (c.statistics().download_payload_rate()
+				> t.torrent_file().piece_length()
+				/ std::max(1, t.settings().get_int(settings_pack::whole_pieces_threshold)))
```
arvidn: instead of using ``max`` here, there should be a check for ``whole_pieces_threshold`` to be 0, and set ``prefer_contiguous_blocks`` to false in that case.

chaoaero: I will try it today and fix potential issue. Thanks!

chaoaero: I will try it today, and fix potential issue if possible. Thanks

PR #2360, opened 2017-09-21
fix kludge and down-cast to send upload_only

arvidn:
Promote upload-only to a top-level peer feature

### DIFF
--- a/src/bt_peer_connection.cpp
+++ b/src/bt_peer_connection.cpp
@@ -315,13 +315,17 @@ namespace {
 		// connections that are still in the handshake
 		// will send their bitfield when the handshake
 		// is done
+		std::shared_ptr<torrent> t = associated_torrent().lock();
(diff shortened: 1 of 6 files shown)


src/bt_peer_connection.cpp
```diff
@@ -315,13 +315,17 @@ namespace {
 		// connections that are still in the handshake
 		// will send their bitfield when the handshake
 		// is done
+		std::shared_ptr<torrent> t = associated_torrent().lock();
 #ifndef TORRENT_DISABLE_EXTENSIONS
-		write_upload_only();
+		if (!t->share_mode())
+		{
+			bool const upload_only_enabled = t->is_upload_only() && !t->super_seeding();
+			write_upload_only(upload_only_enabled);
```
arvidn: maybe this should be ``send_upload_only()``

PR #2362, opened 2017-09-22
back-port aldenml's travis brew fix

arvidn:


aldenml:
I believe this can be solved in the same way like in `master`:
https://github.com/arvidn/libtorrent/pull/1824 (PR)
https://github.com/arvidn/libtorrent/issues/1821 (issue with explanation)

arvidn:
thanks for reminding me. I had forgotten about that!

PR #2368, opened 2017-09-24
code refactor in torrent_handle::info_hash

aldenml:
Hi @arvidn, unless I'm missing something, I think that this code save the thread guarantee from the compiler to `empty`, and of course save the tiny `empty` memory space. The CPU cost of memory copy can't be avoided and RVO should help.

### DIFF
--- a/src/torrent_handle.cpp
+++ b/src/torrent_handle.cpp
@@ -180,9 +180,7 @@ namespace libtorrent {
 	sha1_hash torrent_handle::info_hash() const
 	{
 		std::shared_ptr<torrent> t = m_torrent.lock();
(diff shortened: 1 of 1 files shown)


src/torrent_handle.cpp
```diff
@@ -180,9 +180,7 @@ namespace libtorrent {
 	sha1_hash torrent_handle::info_hash() const
 	{
 		std::shared_ptr<torrent> t = m_torrent.lock();
-		static const sha1_hash empty;
-		if (!t) return empty;
-		return t->info_hash();
```
arvidn: it looks like this function used to return a reference

PR #2374, opened 2017-09-27
fix windows file preallocation issue

ajax16384:
GetFileInformationByHandleEx and GetCompressedFileSize will lie about real space allocation so "set_file_valid_data" will never be called for just preallocated file.
Let's always try to call "set_file_valid_data" on file::set_size change

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -1153,35 +1153,7 @@ namespace {
 				ec.assign(GetLastError(), system_category());
 				return false;
 			}
-		}
-
-#if _WIN32_WINNT >= 0x0600 // only if Windows Vista or newer
-		if (!(m_open_mode & open_mode::sparse))
-		{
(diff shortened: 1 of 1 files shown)


ajax16384:
https://gist.github.com/ajax16384/c0a9b1e50a307bc8f656622f801badea#file-setfilevaliddata-c-L78 this one gives "virtually" preallocated value (not 0 for new file) so old libtorrent code will never called it's "SetFileValidData"

arvidn:
it seems like this should be back ported to RC_1_1 too, does that make sense?

ssiloti:
Yeah. I realized after hitting merge this should have gone to RC_1_1. Sorry.

ajax16384:
NP: https://github.com/arvidn/libtorrent/pull/2376

PR #2382, opened 2017-09-29
add const

pavel-pimenov:


src/http_seed_connection.cpp
```diff
@@ -279,10 +279,10 @@ namespace libtorrent {
 					// temporarily unavailable, retry later
 					t->retry_web_seed(this, retry_time);
 
-					std::string error_msg = to_string(m_parser.status_code()).data()
-						+ (" " + m_parser.message());
 					if (t->alerts().should_post<url_seed_alert>())
 					{
+						std::string const error_msg = to_string(m_parser.status_code()).data()
+							+ (" " + m_parser.message());
```
arvidn: interesting. the ``url_seed_alert`` should probably contain the actual ``error_code`` object too.

src/kademlia/node.cpp
```diff
@@ -270,7 +270,7 @@ void node::incoming(aux::listen_socket_handle const& s, msg const& m)
 		return;
 	}
 
-	char y = *(y_ent.string_ptr());
+	const char y = *(y_ent.string_ptr());
```
arvidn: I prefer the const to the right of the type it applies to, i.e. [continental const placement](https://kuhllib.com/2012/01/17/continental-const-placement/).

PR #2393, opened 2017-10-01
Fix piece size overflow check

timniederhausen:
Previously torrents larger than (INT_MAX - piece_size) were rejected incorrectly
(i.e. torrents larger than ~2GiB)

This fixes the check that was introduced in #2381


arvidn:
ah, yes. thanks. I prefer to have a single check instead of these two. There should be a test for this to prevent a mistake like this again too.

arvidn:
I was thinking something like this. does it look reasonable? https://github.com/arvidn/libtorrent/pull/2395

arvidn:
thanks for bringing this to my attention!

PR #2395, opened 2017-10-01
fix regression in torrent size check

arvidn:


src/torrent_info.cpp
```diff
@@ -1013,7 +1013,8 @@ namespace {
 		// we want this division to round upwards, that's why we have the
 		// extra addition
 
-		if (files.total_size() >= std::numeric_limits<int>::max()
+		if (files.total_size() >=
+			(std::numeric_limits<std::int64_t>::max() / files.piece_length())
```
arvidn: this is also wrong :(
It should be: "files.total_size() >= static_cast<std::int64_t>(std::numeric_limits<int>::max() - files.piece_length()) * files.piece_length())"

PR #2400, opened 2017-10-01
try to link against dbghelp on mingw

arvidn:


silverqx:
Some travis test failed, don't know if its ok

arvidn:
oh, I guess RC_1_1 doesn't have a mingw build on appveyor

arvidn:
it turns out this change has already been made in master though

PR #2405, opened 2017-10-03
Fix building on FreeBSD

timniederhausen:
These changes were necessary to build libtorrent on FreeBSD 10/11.

The default ip_change_notifier (which will be used on FreeBSD) always fails with not_supported.


arvidn:
thanks!

PR #2406, opened 2017-10-03
don't assert on an empty mapping

ssiloti:
If the underlying file is empty then m_mapping is expected to be null.
It is the caller's responsibilty to notice that the region's size is
zero and not dereference it in that case.

Otherwise there is no good way for users of `default_storage::open_file`
to determine that the opened file is empty.


arvidn:
I seem to recall that ``mmap()`` requires a size > 0. but maybe the constructor has a special case for this already

PR #2408, opened 2017-10-04
revert automatically pick of openssl from brew in macOS

aldenml:
This makes the explicit location requirement the same for all platforms and avoid get in the way for custom openssl builds and even multiple openssl brew installation versions


aldenml:
I wonder if you have some comments here

arvidn:
yeah, I'm trying to think of ways to still make the default "just work", but still allow customization.
I'm thinking of maybe introducing separate jam features to customize the ssl libraries. That would allow still allow the brew openssl libraries to be used by default.

I'm also thinking that this should be done in RC_1_1 before I cut 1.1.5

aldenml:
if the solution involve some working/time (specially if going to `RC_1_1`), is it a problem to "temporarily" merge this? I prefer very much work with an upstream checkout than a fork.

aldenml:
Closing since this will resolved in the right way when `RC_1_1` is merge into `master`

PR #2409, opened 2017-10-04
making use of openssl API 1.1 compatible

aldenml:
as a note, the macro `OPENSSL_API_COMPAT` is to be defined in the `openssl` compilation and use the value in the client code.


arvidn:
so you're saying ``OPENSSL_API_COMPAT`` is defined by an openssl header, based on how it was build?

If this is the correct way of doing this, I should back-port this to RC_1_1, right?

aldenml:
Yes, and ... yes

PR #2410, opened 2017-10-05
Fixes to copy_ptr

fmatthew5876:
* Add move constructor
* Add move assignment
* Fix exception safety bug in copy assignment


fmatthew5876:
Just rebased it to fix tests complaining about `(T*)nullptr`

arvidn:
lgtm. I suppose the pointer could really just be held by a ``unique_ptr`` in there. I should add some tests too

PR #2411, opened 2017-10-05
Fix build on FreeBSD 11.1

fmatthew5876:
Build fails on FreeBSD 11.1-p1 using cmake and base clang compiler.
I'm not sure what the correct convention is here for this iconv for
all BSD version, but certainly casting it to non-const and then
passing it in should be safe even if some versions of iconv take
a const pointer.

With current master I get this build error:
```
[ 24%] Building CXX object CMakeFiles/torrent-rasterbar.dir/src/i2p_stream.cpp.o
/home/matt/libtorrent/include/libtorrent/config.hpp:176:31: note: expanded from macro 'TORRENT_ICONV_ARG'
# define TORRENT_ICONV_ARG(x) (x)
                              ^~~
/usr/local/include/iconv.h:85:43: note: passing argument to parameter 'inbuf' here
extern size_t iconv (iconv_t cd,  char* * inbuf, size_t *inbytesleft, char* * outbuf, size_t *outbytesleft);
```

### DIFF
--- a/include/libtorrent/config.hpp
+++ b/include/libtorrent/config.hpp
@@ -173,7 +173,7 @@ POSSIBILITY OF SUCH DAMAGE.
 // FreeBSD has a reasonable iconv signature
 // unless we're on glibc
 #ifndef __GLIBC__
-# define TORRENT_ICONV_ARG(x) (x)
+# define TORRENT_ICONV_ARG(x) (const_cast<char**>(x))
 #endif
 #endif // __APPLE__
 


fmatthew5876:
Thinking about it again. Instead of this macro, its probably better to just wrap the `iconv()` function with a wrapper function which takes `const char**` and have all of the libtorrent code call that instead.

include/libtorrent/config.hpp
```diff
@@ -173,7 +173,7 @@ POSSIBILITY OF SUCH DAMAGE.
 // FreeBSD has a reasonable iconv signature
 // unless we're on glibc
 #ifndef __GLIBC__
-# define TORRENT_ICONV_ARG(x) (x)
+# define TORRENT_ICONV_ARG(x) (const_cast<char**>(x))
```
arvidn: If this patch works, the comment should be removed as well, in fact, this ``#define`` should be removed. The default for this macro is already to perform the ``const_cast``.

arvidn:
I'm not sure I understand how the solution with using an overload would work. The fundamental problem is that many ``iconv()`` implementations have a signature that take a pointer to a mutable C string as input (which is weird, but probably dates back to before C had ``const``).

Some other implementations take the more reasonable ``char const**`` as their input. IIRC, a ``char**`` cannot be implicitly converted to a ``char const**``, because of the additional pointer indirection. So always performing the ``const_cast`` isn't really an option.

It *would* technically be possible to deduce the argument type through template trickery, but that could easily get a bit nasty.

arvidn:
actually, maybe it isn't such a bad idea after all. What do you think about this? https://github.com/arvidn/libtorrent/pull/2412

fmatthew5876:
Killing this since you fixed it with a better solution.

PR #2412, opened 2017-10-05
deduce the second argument to iconv()

arvidn:
instead of requiring it to be configured via a macro

### DIFF
--- a/src/escape_string.cpp
+++ b/src/escape_string.cpp
@@ -515,6 +515,17 @@ namespace libtorrent
 
 #if TORRENT_USE_ICONV
 namespace {
+
+	// this is a helper function to deduce the type of the second argument to
+	// the iconv() function.
+
+	template <typename Input>
+	size_t call_iconv(size_t (&fun)(iconv_t, Input**, size_t*, char**, size_t*)
+		, iconv_t cd, char const** in, size_t* insize, char** out, size_t* outsize)
+	{
+		return fun(cd, const_cast<Input**>(in), insize, out, outsize);
+	}
+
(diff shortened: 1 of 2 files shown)


arvidn:
it seems GCC didn't fix ``decltype(function)`` until 6.1, which might be a bit of a strict compiler requirement.

fmatthew5876:
Nice.

That's way better than the #defines. The only suggestion I would make is to wrap the `call_iconv()` with an inline `libtorrent::iconv()` function one more time so clients don't need to understand how the function pointer trick works.

arvidn:
the only place ``iconv()`` is called from is ``iconv_convert_impl()`` afaict.

Now I just need to figure out how to suppress the unused function warning in gcc.

arvidn:
@fmatthew5876 ok, I think this patch is in pretty good shape now. Would you mind testing it to make sure it fixes the original issue you experienced?

src/escape_string.cpp
```diff
@@ -523,11 +534,10 @@ namespace {
 		ret.resize(outsize);
 		char const* in = s.c_str();
 		char* out = &ret[0];
-		// posix has a weird iconv signature. implementations
-		// differ on what this signature should be, so we use
-		// a macro to let config.hpp determine it
-		size_t retval = iconv(h, TORRENT_ICONV_ARG(&in), &insize,
-			&out, &outsize);
+		// posix has a weird iconv() signature. implementations
+		// differ on the trype of the second parameter. We use a helper template
```
aldenml: no important typo `trype`

fmatthew5876:
Works for me

PR #2414, opened 2017-10-05
cleanup of field torrent_need_cert_alert::error

aldenml:


include/libtorrent/alert_types.hpp
```diff
@@ -1810,8 +1810,6 @@ namespace libtorrent {
 
 		static constexpr alert_category_t static_category = alert::status_notification;
 		std::string message() const override;
-
-		error_code const error;
```
arvidn: I think removing this would alter the API compared to 1.1, wouldn't it?
If so, I think it should be deprecated instead.

aldenml: right, will do that

PR #2415, opened 2017-10-06
back-port openssl API compat patch from master

arvidn:
@aldenml does this look ok?


aldenml:
maybe you can delete
```
# boost.asio use the pre-1.1 OpenSSL API
result += <define>OPENSSL_API_COMPAT=0x10000000L ;
```
from Jamfile, other than that LGTM

arvidn:
I don't think that define ever made it into ``RC_1_1``

aldenml:
true, your are right

PR #2417, opened 2017-10-06
check the write buffer queue in do_read

ssiloti:


arvidn:
this makes sense. However, do you think it would work to check the write buffer in ``async_read()`` instead of ``do_read()``? I would expect there not to be any ordering issues, since all ``async_*`` functions are sequenced in the network thread anyway.
I would expect it to be more efficient since the thread communication and job queueing is avoided.

PR #2418, opened 2017-10-06
simplify copy_ptr

arvidn:


arvidn:
@fmatthew5876 what do you think of this?

include/libtorrent/copy_ptr.hpp
```diff
@@ -33,46 +33,34 @@ POSSIBILITY OF SUCH DAMAGE.
...
-			T* c = p.m_ptr ? new T(*p.m_ptr) : nullptr;
-			delete m_ptr;
-			m_ptr = c;
-			return *this;
-		}
-		copy_ptr& operator=(copy_ptr&& p) noexcept {
-			if(m_ptr == p.m_ptr) return *this;
-			delete m_ptr;
-			m_ptr = p.m_ptr;
-			p.m_ptr = nullptr;
```
aldenml: this line: is it not needed?

aldenml: `p.m_ptr = nullptr`

include/libtorrent/copy_ptr.hpp
```diff
@@ -33,46 +33,34 @@ POSSIBILITY OF SUCH DAMAGE.
...
-		void swap(copy_ptr<T>& p)
-		{
-			T* tmp = m_ptr;
-			m_ptr = p.m_ptr;
-			p.m_ptr = tmp;
-		}
-		explicit operator bool() const { return m_ptr != nullptr; }
-		~copy_ptr() { delete m_ptr; }
+		void swap(copy_ptr<T>& p) { std::swap(*this, p); }
+		explicit operator bool() const { return m_ptr.get() != nullptr; }
```
aldenml: just `return m_ptr.get()`?

include/libtorrent/copy_ptr.hpp
arvidn: since ``std::unique_ptr`` already is movable, the default move assignment operator is fine.

include/libtorrent/copy_ptr.hpp
arvidn: I wanted to avoid an implicit pointer -> integral conversion

fmatthew5876:
This is much better than implementing it manually.

I would also add a constructor which takes a unique_ptr since people should not be passing around owning raw pointers to begin with.

PR #2420, opened 2017-10-07
don't hard-code the paths to find openssl on mac

arvidn:
default to picking up brew. Add two new jam features, openssl-lib and openssl-include that can be used to specify which openssl to link against.

@aldenml does this look good to you?
I will update this PR with documentation as well

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,30 +10,20 @@ environment:
     compiler: msvc-14.0
     os: Visual Studio 2015
     sim: 1
-    linkflags: '"/LIBPATH:C:\\openssl-1.0.1p-vs2015\\lib"'
-    include: '"c:\\openssl-1.0.1p-vs2015\\include"'
   - variant: test_debug
     compiler: msvc-12.0
     os: Visual Studio 2013
     x64: 1
-    linkflags: '"/LIBPATH:C:\\openssl-1.0.1p-vs2013\\lib64"'
-    include: '"c:\\openssl-1.0.1p-vs2013\\include"'
   - variant: test_debug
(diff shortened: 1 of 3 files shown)


aldenml:
@arvidn this looks sophisticated for my basic understanding of `bjam`

arvidn:
yeah, I couldn't resist to clean up some related aspects of the way openssl is linked. For instance, I added an additional abstraction for the library names on windows vs. all other systems with the ``openssl-libraries`` alias target. I also made the libraries properly specify their header paths as *usage-requirements* to correctly propagate to targets linking against them.

Anyway, with this patch there's a clean way to specify a different library- and include path for the openssl libraries (libssl, libcrypto, libssleay32 and libeay32). It's also more correct since the search paths are associated with the libraries themselves, not with the libtorrent target.

I have will add documentation and squash this, but as far as I can tell, this should be sufficient for cross compiling against any custom openssl build.

appveyor.yml
```diff
@@ -83,17 +75,21 @@ install:
...
   )
 - cd %ROOT_DIRECTORY%
 - set BOOST_ROOT=c:\Libraries\boost_1_63_0
 - set BOOST_BUILD_PATH=%BOOST_ROOT%\tools\build
 - echo %BOOST_ROOT%
 - echo %BOOST_BUILD_PATH%
 - set PATH=%PATH%;%BOOST_BUILD_PATH%\src\engine\bin.ntx86
+- dir c:\openssl
+- dir c:\openssl\lib
+- dir c:\openssl\include
```
arvidn: debug-leftovers

PR #2423, opened 2017-10-08
don't try to connect to a global address with a local source address

ssiloti:


src/http_connection.cpp
```diff
@@ -538,6 +538,11 @@ void http_connection::on_resolve(error_code const& e
 			if (ep.address().is_v4() && m_bind_addr->is_v4())
 				return true;
 			TORRENT_ASSERT(ep.address().is_v6() && m_bind_addr->is_v6());
+			// don't try to connect to a global address with a local source address
+			// this is mainly needed to prevent attempting to connect to a global
+			// address using a ULA as the source
+			if (!is_local(ep.address()) && is_local(*m_bind_addr))
```
arvidn: won't this break the case of being behind a NAT? ``is_local()`` returns true for ``10.0.0.1`` for instance, right? but behind a NAT it still makes sense to try to connect to public IPs. or am I misunderstanding something?

aldenml: I believe at this point both addresses are IPv6

arvidn: ah, you're right

PR #2424, opened 2017-10-08
missing use of piece_refcount_holder before flush_iovec in unlocked scope

aldenml:
Hi @arvidn, I'm not sure this is necessary (or right), that's why I'm skipping the CI, but take a look when you have a chance and let me know

### DIFF
--- a/src/disk_io_thread.cpp
+++ b/src/disk_io_thread.cpp
@@ -544,6 +544,7 @@ constexpr disk_job_flags_t disk_interface::cache_hit;
 		{
 			// unlock while we're performing the actual disk I/O
 			// then lock again
+			piece_refcount_holder refcount_holder(first_piece);
 			auto unlock = scoped_unlock(l);
 			flush_iovec(first_piece, iov, flushing, iov_len, error);
 		}


arvidn:
yeah, as far as I can tell, this does look like a bug (or a fix for a bug). I'll look if it's also present in RC_1_1

aldenml:
Ok, to rebase and run CI

arvidn:
looking a bit closer at this, in RC_1_1 as well, it doesn't appear to be a problem (other than the lack of exception safety).

There are loops where multiple piece refcounters are incremented before and after this flush call.

here:https://github.com/arvidn/libtorrent/pull/2424/files#diff-32675a6b33f593e15ce3afb854579882L520
and here: https://github.com/arvidn/libtorrent/pull/2424/files#diff-32675a6b33f593e15ce3afb854579882L569

aldenml:
I saw them, but wasn't sure about the entire logic, I will close this then. Still looking for a reason for the crashes.

arvidn:
however, speaking of disk_io_thread issues, I think I've identified a race condition when aborting torrents. It seems storages for torrents can be removed despite there still being active disk jobs on it.

https://pastebin.com/yDQqTv45

aldenml:
👍

PR #2427, opened 2017-10-09
don't delete pieces from cache with refcount > 0

arvidn:


aldenml:
Do you think this could be "the" reason for the storage crashes?

arvidn:
Not sure. Only if it happens on shutdown probably

aldenml:
shutdown in general, or torrent removal too?

arvidn:
I think the issue this PR addresses is only for full shutdown, where the disk threads terminate

PR #2436, opened 2017-10-11
send one last state update when a paused torrent's rates reach zero

ssiloti:


arvidn:
ah, the stats got stuck at the penultimate value, right before hitting 0?

arvidn:
does this warrant a changelog entry?

ssiloti:
Yeah, our state on the client side gets stuck at one when you pause a torrent.

PR #2437, opened 2017-10-12
added dbghelp lib and bigobj to cmake for msvc

silverqx:
I can't compile without bigobj and RC_1_1 branch depends on dbghelp library, functions in `assert.cpp` file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -268,12 +268,16 @@ target_link_libraries(torrent-rasterbar ${Boost_LIBRARIES} ${CMAKE_THREAD_LIBS_I
 
 if (WIN32)
 	target_link_libraries(torrent-rasterbar wsock32 ws2_32 Iphlpapi)
+	target_link_libraries(torrent-rasterbar debug dbghelp)
 	add_definitions(-D_WIN32_WINNT=0x0600)
 	# prevent winsock1 to be included
 	add_definitions(-DWIN32_LEAN_AND_MEAN)
 	if (MSVC)
 		add_definitions(-DBOOST_ALL_NO_LIB)
-		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP") # multicore compilation
+		# for multicore compilation
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
+		# increase the number of sections for obj files
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")
 	endif()
 endif()
 


ssiloti:
I haven't seen any problems compiling without `/bigobj`. What version of msvc and boost are you using? Which source file is causing the problem?

silverqx:
b2 builds are using bigobj too https://github.com/arvidn/libtorrent/blob/master/Jamfile#L251

boost 1.64 and msvc 2015, what versions do you use? If i remember good, I have to used bigobj everytime to compile, don't know which file right know.

ssiloti:
We compile libtorrent 1.1 with that exact configuration (boost 1.64 + msvc 2015 + cmake) every day and don't see any problems. I'd really like to see the error in case there's something unusual going on with your build.

silverqx:
of course, np, here is output, it's on master:
```
[ 46%] Building CXX object CMakeFiles/torrent-rasterbar.dir/src/session_impl.cpp.obj
session_impl.cpp
E:\c\lt-test\src\session_impl.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj
NMAKE : fatal error U1077: 'C:\PROGRA~2\MICROS~1.0\VC\BIN\amd64\cl.exe' : return code '0x1'
Stop.
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\nmake.exe"' : return code '0x2'
Stop.
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\nmake.exe"' : return code '0x2'
Stop.
22:41:42: The process "C:\Program Files\CMake\bin\cmake.exe" exited with code 2.
Error while building/deploying project libtorrent (kit: Desktop Qt 5.9.1 MSVC2015 64bit)
When executing step "CMake Build"
22:41:42: Elapsed time: 02:19.
```

Nothing is wrong, bigobj is needed for __x64__ builds, as x64 builds needs more memory / space, x32 builds are ok, in this case x64 build hits the section limit, but nothing wrong, bigobj is normal in big files.

ssiloti:
Ok that explains it. We only do 32-bit builds. I'm kind of surprised 64-bit requires significantly more sections, but whatever. LGTM

CMakeLists.txt
```diff
@@ -267,13 +267,16 @@ target_link_libraries(torrent-rasterbar ${Boost_LIBRARIES} ${CMAKE_THREAD_LIBS_I
 #add_definitions(-DBOOST_ASIO_HASH_MAP_BUCKETS=1021  -D__USE_W32_SOCKETS -DWIN32_LEAN_AND_MEAN )
 
 if (WIN32)
-	target_link_libraries(torrent-rasterbar wsock32 ws2_32 Iphlpapi)
+	target_link_libraries(torrent-rasterbar wsock32 ws2_32 Iphlpapi dbghelp)
```
arvidn: ``dbghlp`` is only needed for certain kinds of builds. when asserts are enabled or asio-debugging. Is it OK to add a dependency on it unconditionally here?
I know some systems may not actually have dbghlp installed (wine for instance). But maybe all of the dynamic library dependencies are loaded lazily anyway so it would never be a problem

CMakeLists.txt
```diff
@@ -267,13 +267,16 @@ target_link_libraries(torrent-rasterbar ${Boost_LIBRARIES} ${CMAKE_THREAD_LIBS_I
...
 	add_definitions(-D_WIN32_WINNT=0x0600)
 	# prevent winsock1 to be included
 	add_definitions(-DWIN32_LEAN_AND_MEAN)
 	if (MSVC)
 		add_definitions(-DBOOST_ALL_NO_LIB)
-		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP") # multicore compilation
+		# for multicore compilation
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
```
arvidn: my understanding of what's considered "modern" cmake style is to never define any ABI-affecting flags in libraries, but just to assert that they were set by the top level target. I don't have a strong opinion, and this might make sense. Just curious if you have any opinion on how practical that recommendation would be.

arvidn: I guess this isn't actually a change in this patch anyway

CMakeLists.txt
silverqx: I check this and let know

CMakeLists.txt
silverqx: What do you exactly mean, to use `target_compile_options()` for this two flags? I can move the `if` statement to root block and use `target_compile_options()` as it is in master, its better than modify this flags by `set()`.

arvidn: this isn't actually a comment on your patch, but just how to use cmake in general. My understanding is that since a target cannot propagate requirements *up* the dependency graph in cmake, there's nothing preventing ABI-incompatible objects from being linked together. for instance, if the target that depends on libtorrent does *not* build with ``/MP`` but still links against libtorrent built with ``/MP``, that would cause an ABI incompatibility and subtle memory corruption.

heh, I figured I should look up ``/MP`` just to be sure, and it turns out I had mixed it up with ``/MD`` (to link against the multi-threaded runtime library). So, my comment is even more abstract now, since it doesn't even apply to this line.

silverqx: Ohh I understand, don't know if /MP somehow affects ABI, I think there will be no problem ( with static or dynamic linkage ) if the lib is compiled with /MP and an app which use this lib isn't compiled with /MP. I can check it later, I looked at [msdn docs](https://msdn.microsoft.com/en-us/library/bb385193.aspx) and it is more aggresive option than I expected, should be checked :)

CMakeLists.txt
silverqx: you are right, dbghelp is only needed during debug build, but cmake doesn't provide `FindDbgHelp.cmake` module, we can use:
```
target_link_libraries(torrent-rasterbar debug dbghelp)

Comment (arvidn) [reply CMakeLists.txt]: I would kind of expect the linker to detect that there aren't any references to the library and just drop it.

Comment (silverqx) [reply CMakeLists.txt]: you are right, dbghelp is only needed during debug build, but cmake doesn't provide `FindDbgHelp.cmake` module, we can use:
```
target_link_libraries(torrent-rasterbar debug dbghelp)
```
or add `FindDbgHelp.cmake` too.
What do you mean?
The other thing is, that I can build on master without this lib :/ How it's possible :)

arvidn: perhaps "dbghelp.h" contains a "#pragma lib" directive to link against it or something like that

CMakeLists.txt
ssiloti: I think the current best practice with cmake is to use the `target_*` commands which _do_ propagate requirements up the dependency graph. Hence the change in master to use those.

arvidn: I was under the impression they just propagate down (i.e. to dependencies of the target, not targets depending on this)

silverqx: As described here https://cmake.org/cmake/help/v3.9/manual/cmake-buildsystem.7.html#target-usage-requirements

CMakeLists.txt
silverqx: I think that asserts are disabled for debug builds too in master, `TORRENT_USE_ASSERTS` is not defined, i'm debugging this now. 
As you can see here https://github.com/arvidn/libtorrent/blob/master/include/libtorrent/config.hpp#L531, even the comment message says something else and that is the reason, why libtorrent builds without dbghelp, but i'm not 100% sure now.

arvidn: ah, yes. The logic to enable asserts in debug mode was moved to the Jamfile, but not into the CMakeList.txt iirc. ``TORRENT_USE_ASSERTS`` should be defined when building in debug mode

silverqx: ok, thx

arvidn:
So, this patch looks good in principle. I think it would be better if the ``dbghelp`` dependency was only added when needed. Basically under the same conditions as it is in the [Jamfile](https://github.com/arvidn/libtorrent/blob/master/Jamfile#L56).

Would you be up for setting ``TORRENT_USE_ASSERTS`` by default for debug builds too? It overlaps a bit with this dependency.

silverqx:
I know what you mean, boost build is solved best, cmake build should have assert option and on the base of this option should be added eg dbghelp lib and set up compile macros like TORRENT_RELEASE_ASSERTS, TORRENT_USE_ASSERTS, ... But its to much testing :).

TORRENT_USE_ASSERTS is set up by default for debug build now [https://github.com/arvidn/libtorrent/blob/RC_1_1/CMakeLists.txt#L179](here) and [https://github.com/arvidn/libtorrent/blob/RC_1_1/include/libtorrent/config.hpp#L674](here), course its not optimal :) and in master its not optimal too

PR #2438, opened 2017-10-12
enabled bigobj for msvc cmake build

silverqx:
bigobj is needed for msvc win builds


silverqx:
As you wrote, it's safe, thx
I tried to build static lib with bigobj and paradoxically has smaller sizer.

PR #2439, opened 2017-10-12
added .editorconfig file

silverqx:
what do you think? :)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+# editorconfig.org
+root = true
+
+[*]
+charset = utf-8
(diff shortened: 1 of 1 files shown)


silverqx:
Can be merged to RC_1_1 too?

arvidn:
it's an intriguing idea. I've considered putting a vim config file in the root.

silverqx:
What do you mean by vim config? I'm using it in all projects, its pretty standard now and every ide supports it, clion, eclipse, netbeans, vim too, if you don't like it, np to close this, thx

PR #2440, opened 2017-10-12
make disk_buffer_holder know the size of the buffer it holds

arvidn:
to fix buffer overrun in chained_buffer. The problem was that sometimes the disk subsystem would return a buffer with an offset applied to it, in which case the total buffer size is no longer 0x4000, which previously was assumed to be the case for all disk buffers


arvidn:
I'm having second thoughts about this. It is unnecessary to store the size in the holder type. I was thinking that it's a step in the direction of extending adoption of ``span<>``

aldenml:
I thought of that too (the use of `span<>`), but assumed you have your reasons, well, in this particular case, whatever is faster for merge :)

PR #2441, opened 2017-10-13
fix GCC-4.9 build

arvidn:


aldenml:
Hi @arvidn I don't understand the need for `<link>shared` but now I'm getting this error
```
error: No best alternative for /Users/aldenml/Development/libtorrent/crypto
    next alternative: required properties: <link>shared <openssl-version>pre1.1 <target-os>windows
        not matched
    next alternative: required properties: <link>shared
        not matched
error: No best alternative for /Users/aldenml/Development/libtorrent/ssl
    next alternative: required properties: <link>shared <openssl-version>pre1.1 <target-os>windows
        not matched
    next alternative: required properties: <link>shared
        not matched
```

arvidn:
oh. I should probably revert that. sorry. It was meant as a requirement for libssl and libcrypto to be linked dynamically. But I think it ends up as a requirement on the dependent target

aldenml:
👍

arvidn:
fixed

aldenml:
it's working now, thanks!

PR #2447, opened 2017-10-14
pick standard std::aligned_union if using clang

aldenml:
Since in clang `__GNUC__` is `4`, the custom defined `aux::aligned_union` was used.

btw, this code `unsigned char __data[(_Len + _Align - 1)/_Align * _Align];` from the standard `type_traits` seems a little more sophisticated than the one defined in `aligned_union.hpp`

### DIFF
--- a/include/libtorrent/aux_/aligned_union.hpp
+++ b/include/libtorrent/aux_/aligned_union.hpp
@@ -37,7 +37,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent { namespace aux {
 
-#if defined __GNUC__ && __GNUC__ < 5
+#if defined __GNUC__ && __GNUC__ < 5 && !defined(__clang__)
 
 constexpr std::size_t max(std::size_t a)
 { return a; }
--- a/include/libtorrent/aux_/aligned_union.hpp
+++ b/include/libtorrent/aux_/aligned_union.hpp
@@ -37,7 +37,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent { namespace aux {
 
-#if defined __GNUC__ && __GNUC__ < 5 && !defined(__clang__)
+#if defined __GNUC__ && __GNUC__ < 5 && !defined(_LIBCPP_VERSION)
 
 constexpr std::size_t max(std::size_t a)
 { return a; }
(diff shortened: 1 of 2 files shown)


include/libtorrent/aux_/aligned_union.hpp
```diff
@@ -37,7 +37,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent { namespace aux {
 
-#if defined __GNUC__ && __GNUC__ < 5
+#if defined __GNUC__ && __GNUC__ < 5 && !defined(__clang__)
```
arvidn: technically, this should check for libc++. it's possible to use libstdc++ with clang too

arvidn:
is that expression better in some respect than using ``alignas``?

include/libtorrent/aux_/aligned_union.hpp
aldenml: true, but I think it's very unlikely that a `libstdc++` corresponding to such and old `gcc` will play nice with `clang` and c++11. What do you think?

aldenml:
no idea, but I'm afraid that the custom one is not entirely accurate, since I'm getting weird errors in `entry::copy`. I just discovered the same issue for `aligned_storage`, but before doing more, please tell me if this is OK, or I need to figure out a way to detect `libstdc++`

arvidn:
would you mind using ``_LIBCPP_VERSION`` instead?

what error do you see in ``entry::copy``?

include/libtorrent/aux_/aligned_union.hpp
arvidn: when clang first appeared on mac, it used libstdc++ by default. and clang has always had strong c++11 support, so I don't think it's that unlikely

aldenml:
Ok, I will use `_LIBCPP_VERSION` instead.

The error happens with very low frequency, `SWIG` generates a line like this:
```
*(libtorrent::entry **)&jresult = new libtorrent::entry((const libtorrent::entry &)result); 
```
And sometimes in that line happens a crash that I can trace always to:
```
entry.cpp:L551: 
		case preformatted_t:
			new (&data) preformatted_type(e.preformatted());
			break;
		}
```
But some random errors happens in `utp_stream.cpp` too, I just need more information

arvidn:
with ``-fsanitize=undefined``, you actually get reports of any unaligned memory access. The crash you see, is it a SIGBUS?

aldenml:
Last time I checked, android NDK had very limited support for `-fsanitize`, but I will check again, also, I need to look again at the crash to see if it's a `SIGBUS`, for now, my priority is the crash with UPnP

PR #2449, opened 2017-10-15
Expose torrent_alert::torrent_name in python bindings

aresch:


arvidn:
thanks! would you mind adding a call to this function in ``test.py`` too?

aresch:
I added a test.  You may want to consider formatting your test.py in a pep-8 compatible mode, it looks like it's full of tabs (I got ~377 errors in my linter when modifying it).

arvidn:
I believe master has a compatible formatting

PR #2452, opened 2017-10-16
close connection in upnp::on_upnp_xml and added abort/closing returns

aldenml:
@arvidn this is another attempt to improve correctness in upnp, but I can't say it is related to the recursion crash

### DIFF
--- a/src/upnp.cpp
+++ b/src/upnp.cpp
@@ -955,6 +955,8 @@ void upnp::on_upnp_xml(error_code const& e
 		d.upnp_connection.reset();
 	}
 
+	if (m_closing) return;
+
 	if (e && e != boost::asio::error::eof)
 	{
 #ifndef TORRENT_DISABLE_LOGGING
@@ -1059,6 +1061,7 @@ void upnp::on_upnp_xml(error_code const& e
 		return;
 	}
 
+	if (d.upnp_connection) d.upnp_connection->close();
 	d.upnp_connection = std::make_shared<http_connection>(m_io_service
 		, m_resolver
 		, std::bind(&upnp::on_upnp_get_ip_address_response, self(), _1, _2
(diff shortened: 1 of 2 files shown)


src/upnp.cpp
```diff
@@ -955,6 +955,8 @@ void upnp::on_upnp_xml(error_code const& e
 		d.upnp_connection.reset();
 	}
 
+	if (m_closing) return;
```
arvidn: what worries me a bit with the placement of this is that ``d.disabled = true`` may be skipped. Specifically in the case of shutting down and the error may be ``eof``

arvidn:
if you have reason to believe that the crash happens on shutdown, is it possible that perhaps some async. operation outlives the upnp object itself? or perhaps one of the mapping objects.

src/upnp.cpp
aldenml: I put it there because `on_upnp_get_ip_address_response` has it in the same place, do you prefer I move both to below the `eof` logic (or remove both)? or just this one?

aldenml:
I honestly have no idea, the crash was always there, but happening  just once or twice a day, but around the time the `ip_notifier` started to work, the crash is happening about 400X times

aldenml:
@arvidn I'm starting to beleive that putting this returns is part of the solution, but I would like to know what do you think. The problem that I see is that when
```
m_io_service.post(std::bind(&session_impl::abort_stage2, this));
```
is executed, there is no guarantee that no more async operations queued in the upnp logic, are put after it, precisely due to the lack of this "returns". Then the `work` is reset and I imagine that could be problematic. I also detected the lack of this return in `upnp::on_reply`, do you think this idea makes sense?

ssiloti:
The only way I can see an outstanding async event causing a problem is if it holds a non-owning reference to the `upnp` object. An async event being outstanding when `work` is reset is not a problem. The reason to have `work` is to keep the `io_service` alive even when there are no oustanding events. If there is an event in the queue when `work` is destroyed it will simply delay destruction of the `io_service`.

aldenml:
@ssiloti thank you for the comment and clarification, do you have any idea of how I can test for it or force the situation, even if that means putting some custom code in abort?

PR #2455, opened 2017-10-18
fix python binding for torrent_status::error_file

arvidn:


aresch:
Yup, that works but it looks like the same issue exists for flags too:

>           value = getattr(struct, key)
E           TypeError: No Python class registered for C++ class libtorrent::flags::bitfield_flag<unsigned long, libtorrent::torrent_flags_tag, void>

I tested the other keys and they seem to work.

PR #2456, opened 2017-10-19
added and moved some me(self()) in upnp code

aldenml:
@arvidn @ssiloti do you think this could be it?

### DIFF
--- a/src/upnp.cpp
+++ b/src/upnp.cpp
@@ -639,6 +639,9 @@ void upnp::map_timer(error_code const& ec)
 	TORRENT_ASSERT(is_single_thread());
 	COMPLETE_ASYNC("upnp::map_timer");
 	if (ec) return;
+
+	std::shared_ptr<upnp> me(self());
+
 	if (m_closing) return;
 
 	try_map_upnp(true);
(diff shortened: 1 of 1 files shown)


src/upnp.cpp
```diff
@@ -1566,6 +1569,8 @@ void upnp::on_expire(error_code const& ec)
 	COMPLETE_ASYNC("upnp::on_expire");
 	if (ec) return;
 
+	std::shared_ptr<upnp> me(self());
```
arvidn: I don't believe there's anything in this function that may cause the last reference to this object to be released. do you?

src/upnp.cpp
```diff
@@ -796,6 +799,8 @@ void upnp::next(rootdevice& d, port_mapping_t const i)
 void upnp::update_map(rootdevice& d, port_mapping_t const i)
 {
 	TORRENT_ASSERT(is_single_thread());
+	std::shared_ptr<upnp> me(self());
```
arvidn: as far as I can tell, all async operations in here correctly captures a refcounted shared_ptr to the upnp object. I don't see a way for the last reference to be released in here.

aldenml:
Closing this, not well thought

PR #2459, opened 2017-10-20
missing return if error in session_impl::on_port_mapping, doc typo

aldenml:


ssiloti:
The return should not be conditional on the alert being enabled.

aldenml:
correct, I will change that

src/session_impl.cpp
```diff
@@ -5417,6 +5417,7 @@ namespace {
 		{
 			m_alerts.emplace_alert<portmap_error_alert>(mapping
 				, transport, ec);
+			return;
```
arvidn: the idea is to return in case there's an error, right? this only returns if the alert is enabled as well.
However, the code right now seems to deliberately run the whole function unconditionally of an error or not. I can't see anything that needs to happen on errors though, at a glance.

It just cast an external IP vote and updates the external port. presumably those may not be set correctly in case of an error, so perhaps it does make more sense to return here. However, if so, the check for ``!ec`` at the end of the function should be removed too I think

arvidn: and also this if-block should be:
```
if (ec) {
   if (m_alerts.should_post<portmap_error_alert>()) {
      m_alerts.emplace_alert<portmap_error_alert>(mapping
        , transport, ec);
   }
   return;
}
```

aldenml: yes, I will change to that and remove the check at the end

aldenml:
can you say what happened in appveyor?

PR #2461, opened 2017-10-22
remove some questionable code from disk_io_thread when deleting files

arvidn:
extend transfer test to cover deleting files while seeding


aldenml:
do you think this was causing some problems? or it's just a cleanup?

arvidn:
I hit an issue suggesting that this was a problem. specifically, when this code was aborting jobs, I don't think it took into account that some of those jobs may be executing on a separate thread concurrently. I think this fixes that issue, but I couldn't tease out the race in testing.

PR #2464, opened 2017-10-22
making  default_storage::m_part_file creation thread safe

aldenml:


src/storage.cpp
```diff
@@ -103,11 +103,18 @@ namespace libtorrent {
...
 		if (m_part_file) return;
 
-		m_part_file.reset(new part_file(
-			m_save_path, m_part_file_name
-			, files().num_pieces(), files().piece_length()));
+		// the need of this synchronization is that writev runs in
+		// a multithreaded environment with the mutex unlocked
+		std::call_once(m_part_file_flag, [this]()
+		{
+			m_part_file.reset(new part_file(
```
arvidn: you would also need to reset the ``m_part_file_flag`` every time ``m_part_file`` is reset too, right?
it's also not obvious that the access to ``m_part_file`` in the if-statement above is race-free. There's no atomic operation when accessing it, but it may still be written to concurrently. I would imagine it would need to be a ``std::atomic<std::unique_ptr<>>`` or something.

aldenml: true, I didn't see that other `reset` in `delete_files`.

For what I read `std::call_once` is thread safe, could it be written concurrently?

aldenml: I think I understand your point better now, what are you saying is that even calling `.reset()` could be problematic, in the sense that counts as a write? am I right?

aldenml:
Well, given that there is not possible such `std::atomic<std::unique_ptr<>>`, would you consider the atomic of the raw pointer?

arvidn:
Actually, the calls to ``reset`` are already synchronized, they can't happen concurrently with any other disk job. But the check at the start of ``need_partfile()`` also reads the pointer value. Another thread could be writing to it concurrently with no synchronization, causing a race condition.

note that synchronization needs to happen both in both threads.

arvidn:
perhaps the simplest solution would be to unconditionally construct the partfile in ``set_file_priorities`` (if at least one non-pad-file has priority 0, or something like that).

since that's a fence job, it's already synchronized and the concurrent jobs would always have the partfile object if they needed it.

aldenml:
How it looks now?

aldenml:
Hi @arvidn, do you mind (if you have the time), explain a little more about how a "fence job" is actually fenced and how it relates (if any) to the mutex that is unlocked before `writev`? particularly, I want to understand how the data during the `writev` is not affected or how it's protected if another "fence job" start to run, if that it's possible? Thanks

arvidn:
I'm not sure why there's a new warning in the mingw build now. it looks like some ssl include needs to be wrapped in disabling warnings..

aldenml:
jeje, that's new, looks like an update in mingw
```
g++ (Rev1, Built by MSYS2 project) 7.2.0
```

aldenml:
I will rebase once https://github.com/arvidn/libtorrent/pull/2466 is merge

arvidn:
the way fence jobs work is to effectively introduce a "fence" or "barrier" in front of the job *and* after it. i.e. it marks the torrent to redirect any subsequent disk jobs into a separate queue, where they're held up instead of posted to the thread. It also add itself to this queue. When the last job on the torrent completes (there's a counter in the storage object) it detects that it is currently in "fenced" mode and it will post the oldest job from the queue (i.e. the fenced job). this means no other job is running concurrently. Once the fence job completes it will post the queued up jobs.

Some care is taken to allow fence jobs to be queued up as well, so only jobs up until the next fence job are posted to the disk thread.

aldenml:
Thanks, I still need some info, how `m_cache_mutex` plays into the equation? for what I see, it is eventually unlocked right before calling `flush_iovec` (locked after that). If all this jobs are fenced, how is that this "unlock" is necessary, or how is that this "unlock/lock" of `m_cache_mutex` improve concurrency? Thanks again

arvidn:
the cache mutex is not related to the part file. It protects the block cache object, specifically updates to it. The reason it's unlocked during actual disk operations is to allow other jobs to access it while waiting for the disk. To avoid having pieces and blocks evicted during disk operations, pieces are pinned into the cache temporarily, while the I/O is happening.

aldenml:
I see...much better now...then the problem with `m_part_file` was that: even if the jobs are fenced, the fact that they were executed in different threads, made the use of a non atomic "unique_ptr" problematic. Is that correct?

if my understanding is correct now? how is this patch an improvement under repeated calls to `set_file_priority`?

arvidn:
well, the reason this is a problem (prior to this patch) is that ``async_read`` and ``async_write`` are *not* fence jobs. those would call ``need_partfile()`` and introduce a race. Now, those jobs can safely run concurrently since they don't modify the state of the storage object, since the part file is only ever created and destructed inside a fence job.

aldenml:
Thanks!

arvidn:
thanks for the patch and PR!

arvidn:
hm. it just occurred to me that this problem may need to be addressed in RC_1_1 as well.

PR #2468, opened 2017-10-23
revert early return in session_impl::on_port_mapping and only check for error if cast_vote

aldenml:
Sorry I missed a bit of logic here, what I was trying to avoid was a `cast_vote` under an error, but I missed the fact that the ports are set too. And this is important, since under errors, they are set to `0`.

This essentially reverts the previous PR and only check for the error before calling `cast_vote`.

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -5413,14 +5413,10 @@ namespace {
 	{
 		TORRENT_ASSERT(is_single_thread());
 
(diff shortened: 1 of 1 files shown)


arvidn:
looks reasonable. what do you think about a comment to avoid making this mistake again?

aldenml:
sure, adding comment

PR #2470, opened 2017-10-23
add limit of max 50 upnp mappings and recover free global mappings

aldenml:
This PR is to protect the "not so robust" UPnP mapping code from going crazy in case mapping/unmapping is called indiscriminately.

The number `50` is arbitrary.

### DIFF
--- a/src/upnp.cpp
+++ b/src/upnp.cpp
@@ -235,7 +235,9 @@ port_mapping_t upnp::add_mapping(portmap_protocol const p, int const external_po
 		TORRENT_ASSERT(m_mappings.size() <= max_global_mappings);
 		if (m_mappings.size() >= max_global_mappings)
 		{
+#ifndef TORRENT_DISABLE_LOGGING
 			log("too many mappings registered");
+#endif
 			return port_mapping_t{-1};
 		}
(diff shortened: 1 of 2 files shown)


src/upnp.cpp
```diff
@@ -227,6 +232,11 @@ port_mapping_t upnp::add_mapping(portmap_protocol const p, int const external_po
 
 	if (mapping_it == m_mappings.end())
 	{
+		if (m_mappings.end_index() >= max_global_mappings)
+		{
+			log("too many mappings registered");
```
arvidn: do you think it makes sense to add an assert here too? This isn't really expected to happen, right? (I suppose the client can add its own mappings and exceed this limit, but it doesn't seem all that likely)

aldenml: yes, I will the assert

pavel-pimenov:
@aldenml thank you
https://drdump.com/Problem.aspx?ProblemID=324929&login=guest

arvidn:
does the call to ``log()`` need ifdefs around it? or some additional namespace qualifiers?

https://travis-ci.org/arvidn/libtorrent/jobs/291838118#L366

aldenml:
missing ifdefs

PR #2471, opened 2017-10-24
upnp: reduce size NewPortMappingDescription

pavel-pimenov:
TP-Link archer c2
default


arvidn:
I don't see what the value is here exactly. Skype seems to be just as verbose for instance.
is the local endpoint that's included in the description wrong? I don't see any other columns saying which internal host the mapping belongs to

pavel-pimenov:
1. truncate description = 31 char (TP-Link bug?)
2. skype description:
default
3. endpoint in the description - redundant 

uTorrent:
default

PR #2472, opened 2017-10-24
add http_parser::header_int

pavel-pimenov:


src/http_parser.cpp
```diff
@@ -130,6 +130,23 @@ namespace libtorrent {
 		return url;
 	}
 
+	std::string const& http_parser::header(const string_view key) const
+	{
+		static std::string empty;
+		auto const i = m_header.find(key.to_string());
```
arvidn: would you mind leaving a TODO comment here that with C++14 the string construction should be removed?

i.e. when using a transparent (``is_transparent``) comparator, a ``string_view`` can be used directly here

src/http_parser.cpp
```diff
@@ -130,6 +130,23 @@ namespace libtorrent {
...
+		if (i == m_header.end()) return empty;
+		return i->second;
+	}
+
+	std::int64_t http_parser::header_int(const string_view key, std::int64_t def_value) const
+	{
+		auto const i = m_header.find(key.to_string());
```
arvidn: and here

arvidn:
looks good otherwise

arvidn:
actually, I would you mind adding a test for this new function in the http_parser test?

src/http_seed_connection.cpp
```diff
@@ -274,8 +274,7 @@ namespace libtorrent {
 				// if the status code is not one of the accepted ones, abort
 				if (!is_ok_status(m_parser.status_code()))
 				{
-					int retry_time = atoi(m_parser.header("retry-after").c_str());
-					if (retry_time <= 0) retry_time = 5 * 60;
+					int const retry_time = static_cast<int>(m_parser.header_int("retry-after", 5 * 60));
```
aldenml: I think you should use `aux::numeric_cast`

src/web_peer_connection.cpp
```diff
@@ -578,8 +578,8 @@ void web_peer_connection::handle_error(int const bytes_left)
 	// associated with the file we just requested. Only
 	// when it doesn't have any of the file do the following
 	// pad files will make it complicated
-	int retry_time = atoi(m_parser.header("retry-after").c_str());
-	if (retry_time <= 0) retry_time = m_settings.get_int(settings_pack::urlseed_wait_retry);
+	int const retry_time = static_cast<int>(m_parser.header_int("retry-after"
```
aldenml: same here

arvidn:
@pavel-pimenov did you abandon this patch?

pavel-pimenov:
no.

src/http_parser.cpp
```diff
@@ -130,6 +130,25 @@ namespace libtorrent {
 		return url;
 	}
 
+	std::string const& http_parser::header(const string_view key) const
+	{
+		static std::string empty;
```
arvidn: I just noticed that this should be ``const``.

src/http_parser.cpp
```diff
@@ -130,6 +130,25 @@ namespace libtorrent {
...
+		// TODO: remove to_string() if we're in C++14
+		auto const i = m_header.find(key.to_string());
+		if (i == m_header.end()) return empty;
+		return i->second;
+	}
+
+	std::int64_t http_parser::header_int(const string_view key, std::int64_t def_value) const
```
arvidn: ideally ``def_value`` would be ``const`` too

include/libtorrent/http_parser.hpp
```diff
@@ -60,14 +61,8 @@ namespace libtorrent {
...
-		std::string const& header(char const* key) const
-		{
-			static std::string empty;
-			auto const i = m_header.find(key);
-			if (i == m_header.end()) return empty;
-			return i->second;
-		}
-
+		std::string const& header(string_view const key) const;
+		std::int64_t header_int(string_view const key, std::int64_t const def_value) const;
```
arvidn: there shouldn't be top-level ``const``s in declarations. Does your compiler suggest you do this?  They don't have any meaning in declarations and only add noise.

src/web_peer_connection.cpp
```diff
@@ -578,8 +578,8 @@ void web_peer_connection::handle_error(int const bytes_left)
 	// associated with the file we just requested. Only
 	// when it doesn't have any of the file do the following
 	// pad files will make it complicated
-	int retry_time = atoi(m_parser.header("retry-after").c_str());
-	if (retry_time <= 0) retry_time = m_settings.get_int(settings_pack::urlseed_wait_retry);
+	int const retry_time = aux::numeric_cast<int>(m_parser.header_int("retry-after"
+						,m_settings.get_int(settings_pack::urlseed_wait_retry)));
```
arvidn: this indented so far that it wraps. I think a single level is sufficient

arvidn:
thanks!

PR #2475, opened 2017-10-24
using NETLINK_NO_ENOBUFS to ignore ENOBUFS in netlink based ip notifier

aldenml:


src/ip_notifier.cpp
```diff
@@ -75,7 +76,12 @@ struct ip_change_notifier_impl final : ip_change_notifier
 	explicit ip_change_notifier_impl(io_service& ios)
 		: m_socket(ios
 			, netlink::endpoint(netlink(NETLINK_ROUTE), RTMGRP_IPV4_IFADDR | RTMGRP_IPV6_IFADDR))
-	{}
+	{
+		// Linux can generate ENOBUFS if the socket's buffers are full
+		// don't treat it as an error
```
arvidn: receiving this message indicates that there was an event that got dropped, right?
I would also imagine that once the socket has been drained, you wouldn't get these anymore, until there's a spike of events overrunning the buffer again.

Are you sure getting these events isn't an indication of the event handler failing to drain the socket?
does the handler currently keep reading from the socket until it returns ``EAGAIN``, or does it only read a single message?

aldenml: well, I'm not sure to be honest, I will need some help from someone that has more knowledge with netlink.

Since I don't know for sure if this is a broadcast or unicast listener, then I'm not sure of what buffer we are dealing with.

It seems to me like this is at least an improvement, by avoiding possible CPU spinning under stress or faulty netlink behavior.

ssiloti: The current handler only reads one message at a time, but the handle immediately issues a new receive request so the socket's buffer _should_ eventually be drained.

@aldenml Once the ip notifier starts getting `ENOBUFS` does it ever stop?

aldenml: at least not before making the application crash (and the phone gets warm). It happened a handful of times, but I can't find a way to reproduce it, probably some special condition in the phone.

include/libtorrent/socket.hpp
```diff
@@ -63,6 +63,13 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "simulator/simulator.hpp"
 #endif
 
+#if TORRENT_USE_NETLINK
+#include <linux/netlink.h>
+#ifndef SOL_NETLINK
+#define SOL_NETLINK 270
```
arvidn: this looks suspicious. I take it this is android specific, is it?

aldenml: agree, actually, the problem is in linux, I will try adding more headers

aldenml: I tried multiple variants/includes in my travis account, and still fails, it looks like I'm not the only one. @ssiloti what do you think?

ssiloti: `sys/socket.h` is the include you want to get the definition of `SOL_NETLINK`. This is with glibc though, Android may be different.

aldenml: I tried that already, but I will try again, just to make sure

aldenml: yes, my latest travis job has that include, and it fails, what other option do I have?

ssiloti: Looks like glibc is not consistent with defining `SOL_NETLINK` and we're not the only ones resorting to defining it ourselves. It's unfortunate, but seems to be necessary.

arvidn: is netlink a linux specific API? is it safe to assume it uses a stable value?

aldenml: I don't think it's linux specific, but more to linux kernel based OSes, but it's not mandatory. I think the value is pretty stable, I mean, fixed and copied everywhere.

ssiloti: Yes, netlink is a Linux specific API. It is safe to assume the value is stable given Linux's strong commitment to maintaining a stable userspace ABI.

PR #2476, opened 2017-10-24
added reopen_network_sockets method to allow manual reopen of listen/outgoing sockets

aldenml:


include/libtorrent/session_handle.hpp
```diff
@@ -1014,6 +1014,18 @@ namespace libtorrent {
...
+		//
+		// It's useful in the case your platform doesn't support the built in
+		// IP notifier mechanism, or if you have a better more reliable way to
+		// detect changes in the IP routing table.
+		//
+		// The ``map_ports`` indicates if the ports are mapped using natpmp
+		// and upnp. If mapping was already made, they are deleted and added
+		// again. This only works if natpmp and/or upnp are configured to be
+		// enable.
+		void reopen_network_sockets(bool map_ports = true);
```
arvidn: instead of taking a bool here, I think you should take flags. Specifically a "type-safe" flag. The benefits are extensibility with preserved ABI, easier to read call sites and stronger type safety

test/test_session.cpp
```diff
@@ -510,3 +510,70 @@ TORRENT_TEST(init_dht)
...
+			std::vector<alert*> alerts;
+			ses.pop_alerts(&alerts);
+			for (auto a : alerts)
+			{
+				std::printf("%d: [%s] %s\n", num, a->what(), a->message().c_str());
+				std::string const msg = a->message();
+				if (msg.find("successfully listening on") != std::string::npos)
+					count_listen++;
+				// upnp
+				if (msg.find("adding port map:") != std::string::npos)
```
arvidn: it's a bit dangerous to look for strings in the log :) I can't think of a better way though, short of implementing a dummy UPnP- and NATPMP router

test/test_session.cpp
```diff
@@ -510,3 +510,70 @@ TORRENT_TEST(init_dht)
 		TEST_EQUAL(count, 1);
 	}
 }
+
+TORRENT_TEST(reopen_network_sockets)
+{
+	auto count_alerts = [](session& ses, int listen, int portmap)
```
arvidn: both ``listen`` and ``portmap`` can be const

test/test_session.cpp
```diff
@@ -510,3 +510,70 @@ TORRENT_TEST(init_dht)
...
+	{
+		int count_listen = 0;
+		int count_portmap = 0;
+		int num = 120; // this number is adjusted per version, an estimate
+		time_point const end_time = clock_type::now() + seconds(15);
```
arvidn: does it really take 15 seconds? I'm a bit worried about the amount of time this test takes to run.
If this was a simulation it wouldn't be a problem, but this is wall clock time.

arvidn:
do you use this function from java? I take it you have a more reliable notification there, is that right?

aldenml:
I have plans to use this function from java in android as soon as it's merged, the `netlink` approach is very unreliably, and android has a supported way to detect IP/network changes, but in the java land.

include/libtorrent/session_handle.hpp
aldenml: ok, will do that

test/test_session.cpp
aldenml: what?!! :D

test/test_session.cpp
aldenml: OK

test/test_session.cpp
aldenml: I will reduce both `num` and `end_time` to a minimum

aldenml:
btw, the other PR is making the "ip notifier" configurable by settings

include/libtorrent/session_handle.hpp
```diff
@@ -1014,6 +1014,24 @@ namespace libtorrent {
...
+		// enable.
+		static constexpr reopen_network_flags_t reopen_map_ports = 0_bit;
+
+		// Instructs the session to reopen all listen and outgoing sockets.
+		//
+		// It's useful in the case your platform doesn't support the built in
+		// IP notifier mechanism, or if you have a better more reliable way to
+		// detect changes in the IP routing table.
+		//
+		// The ``map_ports`` indicates if the ports are mapped using natpmp
```
aldenml: need to remove this comment

bindings/python/src/session.cpp
```diff
@@ -791,6 +792,11 @@ void bind_session()
     }
 
     {
+        scope s = class_<dummy12>("reopen_network_flags_t");
+        s.attr("reopen_map_ports") = lt::session::reopen_map_ports;
```
arvidn: please add this as an attribute directly on the ``session`` class. this hackery with dummy classes is something I've done mostly to stay API compatible when converting enums into type-safe constants.

aldenml: OK

aldenml:
can't believe hunspell is failing with "natpmp" word. @arvidn travis is having problems, is it critical for the merge?

aldenml:
nvm, I will add the word

aldenml:
@arvidn if you have a minute, can you restart the stale job in travis? thanks

aldenml:
closing to restart

PR #2477, opened 2017-10-24
avoid listening in addresses that are not suitable for bittorrent protocol

aldenml:
Hi @ssiloti, I agree with the rationale of keeping the link local ones, but I find these particular addresses (the ones that I'm filtering out) unsuitable for bittorrent. What do you think?

### DIFF
--- a/src/broadcast_socket.cpp
+++ b/src/broadcast_socket.cpp
@@ -139,6 +139,32 @@ namespace libtorrent {
 #endif
 	}
 
+	TORRENT_EXTRA_EXPORT bool is_non_useful(address const& a)
+	{
+		TORRENT_TRY {
+#if TORRENT_USE_IPV6
+			error_code ec;
+			std::string s = a.to_string(ec);
+			if (ec) return true;
+			if (s.find("%dummy") != std::string::npos) return true;
+			// Android Wi-Fi Direct
+			if (s.find("%p2p") != std::string::npos) return true;
+			// Apple Wireless Direct Link
+			if (s.find("%awdl") != std::string::npos) return true;
+			// TODO: study more about
+			// - rmnet_data
+			// - sit
+			// - upnlink
+			// - svnet
+			// - ip6tnl
+			return false;
+#else
+			TORRENT_UNUSED(a);
+			return false;
+#endif
+		} TORRENT_CATCH(std::exception const&) { return false; }
+	}
+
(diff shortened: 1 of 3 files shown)


src/broadcast_socket.cpp
```diff
@@ -139,6 +139,32 @@ namespace libtorrent {
 #endif
 	}
 
+	TORRENT_EXTRA_EXPORT bool is_non_useful(address const& a)
+	{
+		TORRENT_TRY {
+#if TORRENT_USE_IPV6
+			error_code ec;
+			std::string s = a.to_string(ec);
```
arvidn: isn't there a way to do this without converting to a string? I'm a bit ignorant on IPv6 zones unfortunately.

aldenml: it was an easy way I found to get that suffix, do you think this is a costly operation in this context?

ssiloti:
I'm not a fan of trying to enumerate a bunch of non-standard scope names which might not be useful. It's not obvious to me that the direct wifi protocols will never be useful. Couldn't the client potentially find a peer over one of those interfaces using local peer discovery? It's also not clear to me that there is significant benefit from not listening on these interfaces even if they are never used. I'm assuming these interfaces are using local IPv6 addresses. There is very little overhead to listening on local addresses. We won't try to start a DHT node on them or send tracker requests using them.

aldenml:
It's my opinion that is very annoying to have something that will never work in the pseudo-long term, like `Wi-Fi Direct` which is mostly for smaller devices, like a camera (incapable of running `libtorrent`). Putting it in a different way, if a device is capable of running `libtorrent`, more likely it will have a full capable network interface, and communicating them via these "trimmed" protocols will only hurt the actual communication.

This is more acute in android, where devices put USB, Bluetooth, Wi-Fi, to work with whatever they find is the new cool wireless protocol and put IPv6 over it.

ssiloti:
Why is it annoying? I don't think clients typically even show the full list of listen interfaces. If they do it's certainly not featured prominently. As I said, the backend overhead is negligible, so I don't see what there is to be annoyed about.

aldenml:
sorry, I meant "annoying" for me, I don't like the idea of having open ports that will be not used, even if they don't pose any overhead. Well, maybe next time I can convince you of putting some setting to avoid all link-locals :)

PR #2478, opened 2017-10-24
minor and general code cleanup/refactor

aldenml:


include/libtorrent/upnp.hpp
```diff
@@ -308,7 +308,6 @@ struct TORRENT_EXTRA_EXPORT upnp final
 #if TORRENT_USE_ASSERTS
 		int magic = 1337;
 #endif
-		void close() const;
```
arvidn: there's no definition of this function anywhere that also should be removed?

PR #2479, opened 2017-10-25
introduce type for torrent queue_position

arvidn:


arvidn:
there should be named constants for ``queue_position_t{-1}`` and ``queue_position_t{INT_MAX}``

PR #2484, opened 2017-10-26
keep updating aux::time_now() while there are announces

ssiloti:
This fixes an infinite loop when an announce during shutown fails and
`aux::time_now()` has fallen behind `clock_type::now()` by more than the
delay calculated in `announce_entry::failed()`. If this happens,
`announce_entry::can_announce()` can forever return true when called from
`torrent::announce_with_tracker()` because that function uses
`clock_type::now()`.

Also change `torrent::announce_with_tracker()` to use `aux::time_now()` just
for consistency with the reset of the announce related functions.

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -3052,10 +3052,13 @@ retry:
 
 		// we have to keep ticking the utp socket manager
 		// until they're all closed
+		// we also have to keep updating the aux time while
+		// there are outstanding announces
 		if (m_abort)
 		{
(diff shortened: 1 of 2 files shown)


arvidn:
great catch! at some point I was considering using a different clock (and so different types for duration and time_point) for the actual clock and this cheap clock that gets updated regularly. That would avoid some of the issues of comparing time_points from the two different sources.

iirc though, in master I just removed this cached clock value and use the proper clock everywhere.

arvidn:
hm. this should probably be mentioned in the changelog

PR #2487, opened 2017-10-28
enable/disable the internal ip notifier with new setting

aldenml:
This helps to avoid the issue that in some android phones, netlink notifies with high frecuency even if no actual changes happened.

### DIFF
--- a/simulation/test_session.cpp
+++ b/simulation/test_session.cpp
@@ -67,7 +67,7 @@ TORRENT_TEST(seed_mode)
 TORRENT_TEST(ip_notifier_setting)
 {
 	int s_tick = 0;
-	bool working = false;
+	int working_count = 0;
 
 	setup_swarm(1, swarm_test::upload
 		// add session
@@ -79,13 +79,13 @@ TORRENT_TEST(ip_notifier_setting)
 		// add torrent
 		, [](lt::add_torrent_params& params) {}
 		// on alert
-		, [&s_tick, &working](lt::alert const* a, lt::session& ses)
+		, [&s_tick, &working_count](lt::alert const* a, lt::session& ses)
 		{
 			std::string const msg = a->message();
 			if (msg.find("received error on_ip_change:") != std::string::npos)
 			{
 				TEST_CHECK(s_tick == 0 || s_tick == 2);
-				working = true;
+				working_count++;
 			}
 		}
 		// terminate
@@ -110,5 +110,5 @@ TORRENT_TEST(ip_notifier_setting)
 			return ticks > 3;
 		});
 
-	TEST_CHECK(working);
+	TEST_EQUAL(working_count, 2);
 }
--- a/simulation/test_session.cpp
+++ b/simulation/test_session.cpp
@@ -64,3 +64,51 @@ TORRENT_TEST(seed_mode)
 		});
 }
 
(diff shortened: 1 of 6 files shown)


src/session_impl.cpp
```diff
@@ -6599,10 +6614,19 @@ namespace {
...
+
+		m_ip_notifier->cancel();
+		m_ip_notifier.reset();
+	}
+
 	void session_impl::stop_lsd()
 	{
-		if (m_lsd)
-			m_lsd->close();
+		if (!m_lsd) return;
```
aldenml: this change should be no necessary, but something happened in `stop_ip_notifier`, waiting for CI jobs to finish/fail again

src/session_impl.cpp
```diff
@@ -6599,10 +6614,19 @@ namespace {
 		if (m_natpmp) m_natpmp->delete_mapping(handle);
 	}
 
+	void session_impl::stop_ip_notifier()
+	{
+		if (!m_ip_notifier) return;
+
+		m_ip_notifier->cancel();
+		m_ip_notifier.reset();
```
ssiloti: This means the notifier object can be deleted while there is a notification event is queued. It doesn't look like this poses a problem for any of the notifier implementations, but the netlink version should be changed to make `ip_change_notifier_impl::on_notify` a static function because `this` is not used and with this change may be a dangling pointer.

simulation/test_session.cpp
```diff
@@ -64,3 +64,51 @@ TORRENT_TEST(seed_mode)
...
+		// add torrent
+		, [](lt::add_torrent_params& params) {}
+		// on alert
+		, [&s_tick, &working](lt::alert const* a, lt::session& ses)
+		{
+			std::string const msg = a->message();
+			if (msg.find("received error on_ip_change:") != std::string::npos)
+			{
+				TEST_CHECK(s_tick == 0 || s_tick == 2);
+				working = true;
```
arvidn: so this is expected to happen twice, right? this test only asserts that it happens at least once. perhaps "working" should be a counter.

src/session_impl.cpp
```diff
@@ -6527,6 +6531,17 @@ namespace {
...
 #endif
 
+	void session_impl::start_ip_notifier()
+	{
+		INVARIANT_CHECK;
+
+		if (m_ip_notifier) return;
+
+		m_ip_notifier = create_ip_notifier(m_io_service);
+		m_ip_notifier->async_wait([this](error_code const& e)
```
arvidn: is ``m_ip_notifier`` a ``shared_ptr``? if so, perhaps a copy of the shared_ptr should be captured here, just to not have dangling pointers appear at all.

src/session_impl.cpp
arvidn: isn't the handler object itself held by the notifier object?

ssiloti: No, it's captured in the `std::bind` object.

aldenml: thanks for the review, I will accommodate the code for it

simulation/test_session.cpp
aldenml: yes, it's better

aldenml:
Reopening to restart ci

aldenml:
hmm...definitely something odd with the arm CI job

Issue #2489, opened 2017-10-29
crash libtorrent::part_file::readv


pavel-pimenov:
libtorrent version (or branch): master
platform/architecture: windows
compiler and compiler version: vc++2017

13+16 crash dump :-(
https://drdump.com/Problem.aspx?ProblemID=338340&login=guest
https://drdump.com/Problem.aspx?ProblemID=338265&login=guest

possibly after 
https://github.com/arvidn/libtorrent/commit/1ce3483a9e726e780c155911490799e171869874
help?

![default](https://user-images.githubusercontent.com/920705/32145602-68f42212-bcdc-11e7-97f5-2adf2a4617c1.png)

arvidn:
@aldenml do you think we missed a case of where priority may be set to zero? what about the constructor?

aldenml:
definitely some missing case, looking at it at this moment

aldenml:
@pavel-pimenov I think this issue is already fixed in `master`

PR #2490, opened 2017-10-29
fix issue of null m_part_file in default_storage readv/writev

aldenml:
@arvidn this is the missing part in the constructor (as you mentioned in https://github.com/arvidn/libtorrent/issues/2489).

In case you want to merge it, I still have plans to create some unit tests to cover the issue

### DIFF
--- a/src/storage.cpp
+++ b/src/storage.cpp
@@ -89,6 +89,13 @@ namespace libtorrent {
 		TORRENT_ASSERT(files().num_files() > 0);
 		m_save_path = complete(params.path);
 		m_part_file_name = "." + aux::to_hex(params.info_hash) + ".parts";
+
+		file_storage const& fs = files();
+		for (file_index_t i(0); i < m_file_priority.end_index(); ++i)
+		{
+			if (m_file_priority[i] == 0 && !fs.pad_file_at(i))
+				need_partfile();
+		}
 	}
 
 	default_storage::~default_storage()
--- a/src/storage.cpp
+++ b/src/storage.cpp
@@ -94,7 +94,10 @@ namespace libtorrent {
 		for (file_index_t i(0); i < m_file_priority.end_index(); ++i)
 		{
 			if (m_file_priority[i] == 0 && !fs.pad_file_at(i))
+			{
 				need_partfile();
+				break;
+			}
 		}
 	}
 
@@ -389,7 +392,6 @@ namespace libtorrent {
 		{
 			error_code ignore;
 			m_part_file->flush_metadata(ignore);
-			m_part_file.reset();
 		}
 
 		// make sure we don't have the files open
--- a/test/test_priority.cpp
+++ b/test/test_priority.cpp
@@ -123,7 +123,7 @@ void test_transfer(settings_pack const& sett, bool test_deprecated = false)
 	file.close();
 
 	wait_for_listen(ses1, "ses1");
-	wait_for_listen(ses2, "ses1");
+	wait_for_listen(ses2, "ses2");
 
 	peer_disconnects = 0;
 


src/storage.cpp
```diff
@@ -89,6 +89,13 @@ namespace libtorrent {
 		TORRENT_ASSERT(files().num_files() > 0);
 		m_save_path = complete(params.path);
 		m_part_file_name = "." + aux::to_hex(params.info_hash) + ".parts";
+
+		file_storage const& fs = files();
+		for (file_index_t i(0); i < m_file_priority.end_index(); ++i)
+		{
+			if (m_file_priority[i] == 0 && !fs.pad_file_at(i))
+				need_partfile();
```
arvidn: you could even break the loop if you get here

aldenml: will do that

aldenml:
The problem seems to require more than this

aldenml:
@arvidn when you mentioned that `readv/writev` are not fenced jobs, does that mean that if I modify `m_file_priority` inside `set_file_priority`, those methods can read `m_file_priority` while still `set_file_priority` is running?

arvidn:
no, since ``set_file_priority`` *is* fenced, it means no other jobs (on that torrent) are allowed to run concurrently with it. Similarly, obviously no other job (on a specific torrent) can run concurrently with its constructor.

arvidn:
I don't believe the failing jobs on travis are related to this patch. it started happening consistently on master recently

aldenml:
Ok, btw, the failing CI jobs is another problem, the issue of having `readv/writev` fail is big, now that `need_partfile()` is not called inside them, the crashes are out of control, even with this patch

arvidn:
are you sure this patch made it into the version with the issue still? I can't see any other ways of the pad file or the priority to change.

aldenml:
yes, I'm sure

aldenml:
@arvidn in case you have some hints to reproduce it, is it possible that `default_storage::release_files` is called because of some torrent operations, and then `readv/writev` happens?

arvidn:
yes, definitely. ``release_files`` is called whenever a torrent completes. The idea is that all files should be re-opened in read-only mode. Does release_files currently destruct the part_file object? if it does, the rationale was probably that when seeding, we shouldn't need it anyway.. But clearly we do need it. ``release_files`` should not delete the part_file.

aldenml:
it does
```cpp
	void default_storage::release_files(storage_error&)
	{
		if (m_part_file)
		{
			error_code ignore;
			m_part_file->flush_metadata(ignore);
--->>>>			m_part_file.reset();
		}

		// make sure we don't have the files open
		m_pool.release(storage_index());

		// make sure we can pick up new files added to the download directory when
		// we start the torrent again
		m_stat_cache.clear();
	}
```

aldenml:
What about `default_storage::delete_files`? This is called "eventually" by `session_impl::remove_torrent`, but the call sequence is:
```cpp
remove_torrent_impl(tptr, options);

tptr->abort();
```

Could it be that between those two lines, some `readv/writev` happens?

arvidn:
I think this PR should be amended to either not do that.

arvidn:
``delete_files`` should be a fence job, and it shouldn't be posted until after the torrent has stopped, so there shouldn't be anything issuing new reads or writes once it's run.

aldenml:
Ok, will do that ASAP

arvidn:
I guess the priority test really should do some seeding after the download completes too, to cover this case

aldenml:
I will give it a try to that test, btw, do you mind when you have a time, to write a little more about how `delete_files` is guarantee to be posted until after the torrent has stopped?

arvidn:
The only call to ``delete_files()`` on torrent is [here](https://github.com/arvidn/libtorrent/blob/master/src/session_impl.cpp#L5105). This is when a torrent is removed from the session. [delete_files](https://github.com/arvidn/libtorrent/blob/master/src/torrent.cpp#L8117) will in turn disconnect all peers, stop announcing to trackers and call ``m_ses.disk_thread().async_delete_files()``. Immediately following, it sets m_deleted = true [here](https://github.com/arvidn/libtorrent/blob/master/src/torrent.cpp#L8134).

m_deleted will prevent:

1. the user from reading pieces ([here](https://github.com/arvidn/libtorrent/blob/master/src/torrent.cpp#L815))
2. the user from writing to files (https://github.com/arvidn/libtorrent/blob/master/src/torrent.cpp#L1285).
3. resume hashing during file checks ([here](https://github.com/arvidn/libtorrent/blob/master/src/torrent.cpp#L2325), and [here](https://github.com/arvidn/libtorrent/blob/master/src/torrent.cpp#L3776)).

The [disconnect_all()](https://github.com/arvidn/libtorrent/blob/master/src/torrent.cpp#L7162) call will call disconnect() on all peers, which will set [m_disconnecting = true](https://github.com/arvidn/libtorrent/blob/master/src/peer_connection.cpp#L4276) on the peers which should prevent them from issuing any more disk jobs.

aldenml:
Thank you!

And so far, the crash is not happening, waiting more time to have a bigger sample.

PR #2500, opened 2017-11-06
fix IPv6 tracker announce issue

arvidn:


src/torrent.cpp
```diff
@@ -3196,9 +3196,8 @@ namespace {
 			&& m_torrent_file
 			&& m_torrent_file->priv())
 		{
-			tcp::endpoint ep;
-			ep = m_ses.get_ipv6_interface();
-			if (ep != tcp::endpoint()) req.ipv6 = ep.address().to_v6();
+			auto ep = m_ses.get_ipv6_interface();
```
ssiloti: Aren't we still maintaining C++98 compatibility on RC_1_1?

arvidn: Oops. Habit...

PR #2501, opened 2017-11-06
move shared_ptr instead of copying them, when constructing DHT observers

arvidn:


ssiloti:
I don't think this saves a copy unless you pass the `shared_ptr` by value and move the parameter in the caller too.

arvidn:
Yeah you’re right. There are a few more changes required here.

arvidn:
I don't think there's any significant number of places where the traversal_algorithm pointer can be moved in to the top constructor, but this is the more idiomatic way of passing shared_ptr down

PR #2503, opened 2017-11-08
minor cleanup and modernization of torrent_peer

arvidn:


src/peer_list.cpp
```diff
@@ -1113,16 +1093,15 @@ namespace libtorrent {
 #endif
 				new (p) ipv4_peer(remote, true, src);
 
-#if TORRENT_USE_ASSERTS
-			p->in_use = true;
-#endif
-
-			if (!insert_peer(p, iter, flags, state))
+			try {
```
aldenml: other than the missing `\n`, can `insert_peer` throw an exception?

arvidn: yes, since it may allocate memory

PR #2511, opened 2017-11-10
factor out handler allocator

arvidn:
and reuse it for posting deferred peer removals on torrents


aldenml:
how this affects the previous logic of deferred removals? was some corner case still? is keeping a `shared_ptr` an actual improvement? (trying to understand the rationale behind the change)

arvidn:
this change was pulled out of a larger local patch I'm working on where I'm testing error handling paths. Specifically, if the session destruct by an exception, these pointers may become dangling unless they're owning.

 The other (main) aspect of this patch is to pre-allocate the deferred handler, to make sure we can always issue that without risking a memory allocation failure (since peers may be disconnected in response to an exception).

Eventually, I'm hoping that all these failure paths will be tested by simulation/test_error_handling.cpp. But there are more fixes to be made to both libtorrent and the simulator to have that pass.

PR #2514, opened 2017-11-10
add API to query whether alerts have been dropped or not

arvidn:


include/libtorrent/alert_manager.hpp
```diff
@@ -83,6 +87,11 @@ namespace libtorrent {
 
 			maybe_notify(&alert, lock);
 		}
+		catch (std::bad_alloc const&)
+		{
+			// record that we dropped an alert of this type
+			m_dropped.set(T::alert_type);
```
ssiloti: The stack has been unwound at this point, so `m_mutex` needs to be re-acquired before setting `m_dropped`.

PR #2516, opened 2017-11-11
lint guided refactor in dht source code

aldenml:


src/kademlia/find_data.cpp
```diff
@@ -147,8 +147,7 @@ void find_data::done()
 
 	std::vector<std::pair<node_entry, std::string>> results;
 	int num_results = m_node.m_table.bucket_size();
-	for (std::vector<observer_ptr>::iterator i = m_results.begin()
-		, end(m_results.end()); i != end && num_results > 0; ++i)
+	for (auto i = m_results.begin(), end(m_results.end()); i != end && num_results > 0; ++i)
```
arvidn: this is just un-wrapping the line, right? it looks like this exceeds 80 columns now

aldenml:
I will break that line again, I rather keep the 80 columns limit

PR #2523, opened 2017-11-12
fix tutorial examples bt-get.cpp and bt-get2.cpp

arvidn:


arvidn:
this behavior changed in master, these changes can't be made in 1.1 and have already been made in master.

PR #2524, opened 2017-11-12
fix some macro expansions of 'max' in headers

arvidn:
to improve compatibility with windows.h


LiMinggang:
There are two lines left in libtorrent\include\libtorrent\announce_entry.hpp
Line 71 and 74
		// the time of next tracker announce
		time_point32 next_announce = time_point32::min();

		// no announces before this time
		time_point32 min_announce = time_point32::min();
image

arvidn:
fixed here https://github.com/arvidn/libtorrent/pull/2526

PR #2529, opened 2017-11-15
make tracker announces happen even if there are no open listen sockets

arvidn:
@ssiloti I'm not sure this is a reasonable approach. The fundamental feature I want is that there should be at least one tracker announce even when ``m_listen_interfaces`` is empty. (there is a simulation that covers the case as part of this patch).

one think I'm worried about is the removal of the checks in the tracker code whether the designated listen socket is still open. actually, perhaps that could be made conditional.

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -1348,7 +1348,10 @@ namespace {
 					!= m_settings.get_int(settings_pack::ssl_listen))
 			||
 #endif
-			(pack.has_val(settings_pack::listen_interfaces)
+			(pack.has_val(settings_pack::force_proxy)
+				&& !pack.get_bool(settings_pack::force_proxy)
+				&& m_settings.get_bool(settings_pack::force_proxy))
+			|| (pack.has_val(settings_pack::listen_interfaces)
 				&& pack.get_str(settings_pack::listen_interfaces)
 					!= m_settings.get_str(settings_pack::listen_interfaces));
 
@@ -1757,7 +1760,7 @@ namespace {
 	}
 
 	void session_impl::interface_to_endpoints(std::string const& device, int const port
-		, bool const ssl, std::vector<listen_endpoint_t>& eps)
+		, transport const ssl, duplex const incoming, std::vector<listen_endpoint_t>& eps)
 	{
 		// First, check to see if it's an IP address
 		error_code err;
@@ -1767,8 +1770,7 @@ namespace {
 #if !TORRENT_USE_IPV6
 			if (adr.is_v4())
 #endif
-				eps.emplace_back(adr, port, std::string()
-					, ssl ? transport::ssl : transport::plaintext);
+				eps.emplace_back(adr, port, std::string(), ssl, incoming);
 		}
 		else
 		{
@@ -1801,8 +1803,7 @@ namespace {
 				// (which must be of the same family as the address we're
 				// connecting to)
 				if (device != ipface.name) continue;
-				eps.emplace_back(ipface.interface_address, port, device
-					, ssl ? transport::ssl : transport::plaintext);
+				eps.emplace_back(ipface.interface_address, port, device, ssl, incoming);
 			}
 		}
 	}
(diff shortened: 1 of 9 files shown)


src/torrent.cpp
```diff
@@ -2792,6 +2792,11 @@ namespace libtorrent {
 			TORRENT_ASSERT(valid_endpoints <= ae.endpoints.size());
 			ae.endpoints.erase(ae.endpoints.begin() + int(valid_endpoints), ae.endpoints.end());
 
+			// even if we're not listening on any sockets, we want to announce to
+			// trackers
+			if (ae.endpoints.empty())
```
arvidn: is there a legitimate case of when all endpoints get erased? should this have a different condition?

arvidn:
I don't intend to merge this as is (obviously). It's meant as a request for comments

src/torrent.cpp
ssiloti: `ae.endpoints` will be empty if and only if there are no applicable listen interfaces, so this is the correct condition. It means either there are no listen interfaces, or this is an ssl torrent and there are no ssl listen interfaces or vice-versa.

ssiloti:
I think you're right that there will need to be a flag so that the code can distinguish between "this announce's listen socket went away" and "this is an announce without a listen socket".

Announcing to UDP trackers is going to be a problem because it currently needs a listen socket to send on. We could use one of the outgoing UDP sockets, but I suspect the way the code is structured will make that difficult.

arvidn:
hm, yeah. I'll make sure there's a test covering UDP as well

src/torrent.cpp
arvidn: actually, I just realized that if there's an error connecting to a tracker, the endpoint is removed and ``announce_with_tracker()`` is called again. So the retry logic relies on an empty list here being the "base-case". I detected this while working on the udp test, where a failure effectively caused an infinite loop (of immediately retrying the announce over and over again)

arvidn: I think the way this used to work was by recording the fail count in the ``announce_entry`` itself. Now it's stored in the announce_endpoint instead, and is lost (on line 2793). once it's erased, there's no trace that it failed, it will be added back and fail instantly again

ssiloti: That sounds right. We probably should not remove the "no listen socket" endpoint in the code above.

arvidn: another approach would be to have a fake listen socket in the session.

ssiloti: Then you'd have to worry about handling the not-really-a-listen-socket everywhere else that deals with listen sockets. Seems like it could be a bit of a land mine.

ssiloti: If we're aiming to enable tracker announces _and_ keep the DHT going without a configured listen socket then it probably does make sense to create a "default" listen socket which actually does have a UDP socket but no TCP socket. The code should be able to handle this case anyways if it doesn't already. The only special case would be to hide the default socket from the UTP connection manager.

arvidn: yeah, I'm leaning towards doing that. At least for the UDP case, a lot of places assume a ``listen_socket_t``. I could just add a flag to say we should accept incoming connections on it (which applies both for uTP and TCP).

src/session_impl.cpp
```diff
@@ -1856,6 +1864,14 @@ namespace {
 		}
 #endif
 
+		// if no listen interfaces are specified, create sockets to use
+		// any interface
+		if (eps.empty())
+		{
+			eps.emplace_back(address_v4(), 0, "", transport::plaintext
+				, duplex::only_outgoing);
```
ssiloti: There should be a socket for IPv6 as well, otherwise we can't announce to IPv6 only trackers. Binding to `::` is fine in this case.

src/session_impl.cpp
```diff
@@ -1151,6 +1151,12 @@ namespace {
 #endif
 			m_tracker_manager.queue_request(get_io_service(), req, c);
 		}
+		else if (m_listen_sockets.empty())
+		{
+			// we don't have a listen socket and no port
+			req.listen_port = 0;
+			m_tracker_manager.queue_request(get_io_service(), req, c);
+		}
```
ssiloti: I don't think this code is necessary anymore. Now we should always have at least one listen socket.

arvidn:
The original motivation for this change was that RC_1_1 currently clears the m_listen_sockets when force-proxy mode is enabled (and now I wonder if this behavior is right).

In master however, it seems more correct to simply close the listen sockets, but leave the UDP sockets intact (but enable force proxy on them).

The question of what to do when setting the listen_interfaces to an empty string though may still be an interesting one, independently. Does it make sense or is it intuitive that not opening any sockets also stops DHT and tracker announces? perhaps the part of this patch that always adds outgoing-only interfaces still makes sense.

arvidn:
right, the reason why ``m_listen_sockets`` can be cleared in RC_1_1 is because the UDP sockets aren't part of it. in master, it should not be cleared when ``force_proxy`` is set

PR #2531, opened 2017-11-15
use a proper optional value instead of a sentiel value for bytes_left()

arvidn:


Krakean:
Why to use boost::optional if there is std::optional?

arvidn:
master is targeting c++11

Krakean:
Oh, I see.
Curious, why not bump it to C++17 to lower dependency on boost as much as possible?
Current compilers(VS2017, GCC7, CLang 4) support mostly all important C+17 (and completely all C++14) features. It doesn't cost much to update to VS2017 Community Edition (which is free), same for other two. For those who wont or can't - can still use previous version of libtorrent.
Makes sense?

arvidn:
@Krakean The version of GCC shipping in ubuntu LTS is more relevant than the latest release

PR #2532, opened 2017-11-15
use seconds32 as the retry-time for web seeds

arvidn:


include/libtorrent/http_parser.hpp
```diff
@@ -62,7 +67,7 @@ namespace libtorrent {
 		explicit http_parser(int flags = 0);
 		~http_parser();
 		std::string const& header(string_view key) const;
-		std::int64_t header_int(string_view key, std::int64_t def_value) const;
+		boost::optional<seconds32> header_duration(string_view key) const;
```
aldenml: this change looks strange, not all "int result" keys are times, don't you think? (like "content-length" in the test)

arvidn: this isn't an "int header" anymore. this is a "duration header". content-length never used ``header_int()``, perhaps it should. If so, we can add it back. Currently, the only place that use ``header_int()`` use if for duration.

PR #2540, opened 2017-11-21
fix force-proxy regression (udp sockets would not be opened)

arvidn:


aldenml:
is this a problem in `master` too?

arvidn:
No, I don't believe so. master is a lot simpler and more clear in its handling of listen interfaces. It always creates them

calvilasboasjr:
Arvidn, When do you think that a new verson, maybe qb 4.0.1 will be avaliable? 
And, Thank you for your work.

PR #2542, opened 2017-11-21
fix issue in test udp tracker

arvidn:


arvidn:
I don't understand why binding to [::1] works in master but not here. fundamentally I get the impression travis isn't meant to support IPv6 loopback in its containers. but then why isn't it failing in master?

aldenml:
what was the reason at the end? why isn't it failing in master? are you sure travis doesn't support IPv6 loopback?

arvidn:
yeah, my understanding is that the hosting providers travis use don't support it (for some reason). the reason it worked in master was because the test was conditioned on ipv6 being supported in the first place (so the test doesn't run on travis/linux in master). It does run on travis/macOS and windows on appveyor though.

So I back-ported that change as well as an improvement to the detection of IPv6 support.

aldenml:
well, is good to know that it is tested on other CIs, what about doozer.io?

arvidn:
not sure about its x86 environment. that's only in master too though

PR #2544, opened 2017-11-22
more use of emplace_back in torrent.cpp

aldenml:


src/torrent.cpp
```diff
@@ -6365,7 +6365,7 @@ namespace libtorrent {
 			// not be included in this list
 			if (peer->associated_torrent().expired()) continue;
 
-			v->push_back(peer_info());
+			v->emplace_back();
```
thomasyuan: v->emplace_back(peer_info());

aldenml: something wrong with `v->emplace_back()`?

thomasyuan: Needn’t put peer_info() as parameter?

aldenml: no, `emplace_back` builds the object in place with whatever parameters are passed, in this case none, then it uses `peer_info()`

PR #2545, opened 2017-11-23
don't early move shared_ptr plugin in torrent::add_extension_fun, added test for it

aldenml:


arvidn:
good catch! I'm a bit disappointed that no compiler caught this

PR #2546, opened 2017-11-23
fix iOS compile: <net/route.h> not available on iOS

joycepg:
for a while, <net/route.h> was in the simulator header files but not in the device header files. a lot of people used to copy it from the simulator header files to make things compile. but since Xcode 9, it has been removed altogether. a recent [discussion](https://forums.developer.apple.com/thread/84607) on the apple developer forums


arvidn:
does this mean the iOS build won't support NAT-PMP with this patch?
what does the iOS build of ``enum_net`` print with this patch?

arvidn:
addressed here: https://github.com/arvidn/libtorrent/pull/2699

PR #2554, opened 2017-11-24
more use of auto, const and minor refactor in torrent.cpp

aldenml:


src/torrent.cpp
```diff
@@ -4470,7 +4470,7 @@ namespace libtorrent {
 			avail_vec.push_back(i);
 		}
 
-		if (avail_vec.empty()) return piece_index_t(-1);
+		if (avail_vec.empty()) return {-1};
```
arvidn: I believe this only works if the implicit constructor is available, which is deprecated.

aldenml: true

PR #2557, opened 2017-11-25
fix issue with receiving interested before metadata

arvidn:


arvidn:
unfortunately testing issues like this require a significant effort into a peer simulator. This is on my TODO list, but shouldn't block getting this patch in. I'm pretty confident this won't cause any problems at least.

arvidn:
@ishirshov if you have a chance to test this patch, please report back your result!

Issue #2255, opened 2017-08-16
Simultaneously announcing IPv4 and IPv6 addresses


lovesyk:
**Please provide the following information**

libtorrent version (or branch):

1.1.4

platform/architecture:

Debian Stretch x86-64 (Deluge)

compiler and compiler version:

Those used by the debian package / pre-compiled binaries for Ubuntu

please describe what symptom you see, what you would expect to see instead and
how to reproduce it:

After postponing the update to 1.x as long as possible (0.16.18 on Debian Jessie works fine) but now Debian stable having reached 1.1.1 with the issue still in place, I hope it can be solved. The issue seems similar to https://github.com/arvidn/libtorrent/issues/219 but its resolution did not solve this one.

I'm trying to announce both my IPv4 and IPv6 addresses to trackers which resulted in me trying the following configurations:

-  Binding to all addresses results in the client connecting to the tracker using the address it resolves to and not supplying either "ip" or "ipv6" parameters. If the tracker resolves to an v4 address, it only receives my IPv4 address and the same with IPv6.

- Binding to only IPv4 results in the same outcome except the tracker will obviously only see my IPv4 address.

- Binding to only IPv6 results in the client appending the "ipv6" parameter and the tracker receiving only the IPv6 address (two times).

- When binding to all addresses it's possible to use announce_ip which seems to work for both v4 and v6 according to NexusPHP. If one guesses the address family the tracker resolves to (or just forces it using /etc/hosts), it's possible to add the "missing" other address that way. This seems to be the only way to currently pass both v4 and v6 to it while also listening to both addresses.

To be sure the culprit are not these conditions:
https://github.com/arvidn/libtorrent/pull/597/files#diff-19929347e18a6ff1fc37f09ec720306bR3147 I compiled 1.1.4 without them without any change.

A bit off-topic but I don't understand the reason for the latter condition. In a regular scenario I'd want to connect to as many peers as possible, private torrent or not. This obviously includes peers having only an IPv6 address (which at some point in the future will be the whole swarm).

arvidn:
I'm not sure what scenario you have. My understanding is:

* Your computer has one IPv4 address and one IPv6 address.
* The tracker you're announcing to can only be reached via an IPv4 address
* in order for the tracker to track your IPv6 address you need the ``&ipv6=`` parameter to be passed along to the tracker.

Is that right?
What's the actual symtom you experience? Is it that the ``&ipv6=`` parameter is not passed to the tracker?

[...]

PR #2561, opened 2017-11-25
generate random keys for use in tracker announces

arvidn:
keys are unique for each torrent and the listen interface they announce via

### DIFF
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -198,6 +198,10 @@ namespace aux {
 		// which alias the listen_socket_t shared_ptr
 		std::shared_ptr<tcp::acceptor> sock;
 		std::shared_ptr<aux::session_udp_socket> udp_sock;
+
+		// the key is an id that is used to identify the
+		// client with the tracker only.
+		std::uint32_t tracker_key = 0;
 	};
 
 		struct TORRENT_EXTRA_EXPORT listen_endpoint_t
@@ -583,6 +587,8 @@ namespace aux {
 			std::uint16_t ssl_listen_port() const override;
 			std::uint16_t ssl_listen_port(listen_socket_t* sock) const;
 
+			std::uint32_t get_tracker_key(address const& iface) const;
+
 			void for_each_listen_socket(std::function<void(aux::listen_socket_handle const&)> f) override
 			{
 				for (auto& s : m_listen_sockets)
@@ -905,11 +911,6 @@ namespace aux {
 
 			// the peer id that is generated at the start of the session
 			peer_id m_peer_id;
-
-			// the key is an id that is used to identify the
-			// client with the tracker only. It is randomized
-			// at startup
-			std::uint32_t m_key = 0;
 
 			// posts a notification when the set of local IPs changes
 			std::unique_ptr<ip_change_notifier> m_ip_notifier;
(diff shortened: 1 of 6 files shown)


arvidn:
@ssiloti I would appreciate if you would take a look. this is related to https://github.com/arvidn/libtorrent/issues/2255. I generate stable keys for the listen_sockets (based on the local address) and XOR them with the per-torrent key. Perhaps there are better ways to combine them but I figure it's not that big of a deal.

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -906,11 +912,6 @@ namespace aux {
 			// the peer id that is generated at the start of the session
 			peer_id m_peer_id;
 
-			// the key is an id that is used to identify the
-			// client with the tracker only. It is randomized
-			// at startup
-			std::uint32_t m_key = 0;
```
arvidn: no session wide key anymore

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -198,6 +198,10 @@ namespace aux {
 		// which alias the listen_socket_t shared_ptr
 		std::shared_ptr<tcp::acceptor> sock;
 		std::shared_ptr<aux::session_udp_socket> udp_sock;
+
+		// the key is an id that is used to identify the
+		// client with the tracker only.
+		std::uint32_t tracker_key = 0;
```
arvidn: the session wide keys are generated per listen socket now

include/libtorrent/session_handle.hpp
```diff
@@ -586,10 +586,13 @@ namespace libtorrent {
 		// the peer ID is randomized per peer.
 		peer_id id() const;
 
+#ifndef TORRENT_NO_DEPRECATE
+		// deprecated in 1.2
 		// sets the key sent to trackers. If it's not set, it is initialized
 		// by libtorrent. The key may be used by the tracker to identify the
 		// peer potentially across you changing your IP.
 		void set_key(std::uint32_t key);
```
arvidn: I'm not sure what people used this for anyway

src/session_impl.cpp
```diff
@@ -1384,6 +1394,25 @@ namespace {
 			reopen_outgoing_sockets();
 	}
 
+	std::uint32_t session_impl::get_tracker_key(address const& iface) const
```
arvidn: generates a (stable) key for a given listen socket address

src/session_impl.cpp
```diff
@@ -1384,6 +1394,25 @@ namespace {
...
+	{
+		uintptr_t const ses = reinterpret_cast<uintptr_t>(this);
+		hasher h(reinterpret_cast<char const*>(&ses), sizeof(ses));
+		if (iface.is_v4())
+		{
+			auto const b = iface.to_v4().to_bytes();
+			h.update({reinterpret_cast<char const*>(b.data()), b.size()});
```
ssiloti: I'm not sure why anyone would do it, but in theory there could be multiple listen sockets with the same address but different ports. To get unique keys for every listen socket in this case the `original_port` would also need to be incorporated into the hash.

ssiloti:
Just one comment, otherwise lgtm

arvidn:
I’m thinking of the case where the client changes listen port, but only has one interface say. It would be strange for the key to change.

PR #2577, opened 2017-11-30
fix issue with initializing settings on session construction

arvidn:


src/session_impl.cpp
```diff
@@ -448,19 +447,15 @@ namespace aux {
 		TORRENT_ASSERT_VAL(!ec, ec);
 
 		update_time_now();
+		m_disk_thread.set_settings(&pack, m_alerts);
```
aldenml: is this the issue?

src/session_impl.cpp
```diff
@@ -604,26 +598,11 @@ namespace aux {
...
-		update_download_rate();
-		update_upload_rate();
-		update_connections_limit();
-		update_unchoke_limit();
-		update_disk_threads();
-		update_network_threads();
-		update_upnp();
-		update_natpmp();
-		update_lsd();
-		update_dht();
```
aldenml: what would happen in `master`? there is a fix there to avoid double creation of DHT

arvidn:
the underlying issue was that in the ``session_impl`` constructor's initialization list, the member ``m_settings`` was used *before* it had been initialized, and even after it had been initialized, it hadn't had its settings set to the correct values yet.

This patch properly constructs ``m_settings`` first, and it also initializes it with the settings values passed in to the constructor by the user (passed to the ``session`` constructor).

This means (among other things) the ``alert_manager`` can now be properly initialized with the correct mask and queue size, on construction. In general, i prefer to initialize things to their correct value in the constructor, so I think this is a step forward.

The consequence of doing this was that when applying the settings later, in ``session_impl::init``, none of the ``session_impl::update_*()`` functions were called, because all the settings were already set to the same value. So the ``apply_settings_pack()`` call was redundant. So the settings pack could be removed from the ``init()`` call.

Now, the ``update_*()`` functions need to be called on startup anyway, to actually execute the settings. That's why there was a long list of calls to them. Not all of them though, because, by accident, some update functions are already called by the ``apply_settings_pack`` call. If you're unlucky and you initialize the session with settings that happen to match the default, those ``update_*()`` functions wouldn't be called.

So, the other aspect of this patch is to properly make sure all of the update functions are called on initialization. This should fix some subtle and rare issues.

src/session_impl.cpp
arvidn: previously this would be called by ``apply_settings_pack()``, which I removed (because it ended up being redundant when ``m_settings`` was already initialized to the same settings, from the constructor)

src/session_impl.cpp
arvidn: I'm not sure. do you know if there's a test to detect the double-creation?

arvidn: I would expect this to be the only place the DHT be created on startup though

aldenml:
agree with the general idea, let' see how it plays when merge into `master`

src/session_impl.cpp
aldenml: yes, there is a test for that

aldenml:
Hi @arvidn, I recommend you taking a look at https://github.com/arvidn/libtorrent/pull/1830 (both your comments and code), I have the impression that this PR could be problematic in `master` or a considerable change in `master` will be necessary.

PR #2578, opened 2017-11-30
implemented support magnet URI extension, select specific file indices for download, BEP53

aldenml:


test/test_magnet.cpp
```diff
@@ -476,3 +476,82 @@ TORRENT_TEST(invalid_web_seed_escaping)
...
 
+TORRENT_TEST(parse_magnet_select_only)
+{
+	error_code ec;
+	add_torrent_params p = parse_magnet_uri("magnet:?xt=urn:btih:cdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcd"
+		"&dn=foo&so=0,2,4,6-8", ec);
+	TEST_CHECK(!ec);
+
+	TEST_EQUAL(p.file_priorities.size(), 9);
+	for (std::size_t i = 0; i < p.file_priorities.size(); i++)
```
arvidn: I think this would be clearer by doing:
```
auto const yes = default_priority;
auto const no = dont_download;
TEST_CHECK(p.file_priorities == std::vector<download_priority_t>{
   yes, no, yes, no, yes, no, yes, yes, yes});
```
what do you think?

test/test_magnet.cpp
```diff
@@ -476,3 +476,82 @@ TORRENT_TEST(invalid_web_seed_escaping)
...
+		{
+			TEST_EQUAL(p.file_priorities[i], default_priority);
+		}
+		else
+		{
+			TEST_EQUAL(p.file_priorities[i], dont_download);
+		}
+	}
+
+	p = parse_magnet_uri("magnet:?xt=urn:btih:cdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcd"
```
arvidn: if it's not too much trouble, I think these should be separated into individual test cases. If something breaks, it's a lot easier to pin-point which case broke in that case

test/test_magnet.cpp
```diff
@@ -476,3 +476,82 @@ TORRENT_TEST(invalid_web_seed_escaping)
...
```
arvidn: does the BEP defined multiple arguments being additive?

ChangeLog
```diff
@@ -1,3 +1,5 @@
+	* implemented support magnet URI extension, select specific file indices
+	  for download, BEP53
```
arvidn: would you mind adding a bullet for this BEP on the ``docs/features.rst`` page as well? ideally with a link to bittorrent.org

src/magnet_uri.cpp
```diff
@@ -258,14 +258,68 @@ namespace libtorrent {
...
 
+		auto select_pos = std::string::npos;
+		string_view select = url_has_argument(uri, "so", &select_pos);
+		while (!select.empty())
+		{
+			// parse the ranges or indices
+			do
+			{
+				string_view token;
+				std::tie(token, select) = split_string(select, ',');
```
arvidn: technically this would accept quotes. say: ``&so="1,2"`` would not be interpreted the way you would expect, but also not fail. i.e. ``split_string`` doesn't split quoted substrings.

arvidn: maybe this isn't the end of the world, but i think there should be some negative tests problably, of invalid input.

src/magnet_uri.cpp
```diff
@@ -258,14 +258,68 @@ namespace libtorrent {
...
+
+				int idx1, idx2;
+				// TODO: what's the right number here?
+				constexpr int max_index = 10000; // can't risk out of memory
+
+				auto const divider = token.find_first_of('-');
```
arvidn: I would imagine ``split_string()`` is an even better fit here, since you only expect there to be a single "-", so you get both sides back right away.

arvidn: to know whether it was found or not, see if the second result is an empty string

src/magnet_uri.cpp
```diff
@@ -258,14 +258,68 @@ namespace libtorrent {
...
+				if (divider != std::string::npos) // it's a range
+				{
+					if (divider == 0) // no start index
```
arvidn: ideally there would be a test case for every failure that's handled here

arvidn: including indices > 10000

arvidn:
thanks for this!

ChangeLog
aldenml: OK

src/magnet_uri.cpp
aldenml: OK

src/magnet_uri.cpp
aldenml: I'm not sure I follow you here, could you rephrase please?

src/magnet_uri.cpp
aldenml: OK

test/test_magnet.cpp
aldenml: yes, better :)

test/test_magnet.cpp
aldenml: OK

test/test_magnet.cpp
aldenml: no, that was my interpretation, it could be a reset, you tell me

src/magnet_uri.cpp
arvidn: if you look at the test for ``split_string()``, you'll see that it supports quotes. i.e. it won't split quoted strings

test/test_magnet.cpp
arvidn: I don't have an opinion. perhaps it makes sense to bring it back to bittorrent.org and see if anyone else does

aldenml: @dessalines what is your position here?

src/magnet_uri.cpp
aldenml: I will add some protection here

src/magnet_uri.cpp
aldenml: looking at this again, I need the index to know if the `-` is at the beginning or right at the end

Col-blimp:
`add_torrent_params` has `url` depreciated so there is no way to tell if a magnet link is a BEP53 link or not once it hits the session.

arvidn:
@Col-blimp is that a question or a statement?
in principle that's right, it's definitely the direction of the new ``read_resume_data()`` and ``write_resume_data()``. However, as long as you build with deprecated functions enabled, that ``url`` field is still there, and should still be set to the original magnet link when you get the ``add_torrent_alert``. But maybe you had a different case in mind.

Col-blimp:
It was pointing out a potential problem, BEP53 is like me deciding you take your coffee black when you want it with milk and sugar so it should be optional and there is nothing in the API to give you the choice, `url ` from  `add_torrent_params` would do the job as you could just look for `&so=` in the string but as it's depreciated you can't do that, BEP53 is not expected behaviour so I'm thinking you should be able to know if a link is or is not BEP53.

arvidn:
ah, right. but with the non-deprecated behavior, you get to see the file priorities in ``add_torrent_params`` and you can alter them any way you'd like before adding it to the session

Col-blimp:
non-deprecated is fine but enabling depreciated features to use a new feature doesn't seem right, put it another way, once a BEP53 link has been added to the session given `add_torrent_params` `url` is unavailable and `torrent_handle` `file_priorities()` is unavailable how can you on the `metadata_received_alert` determine if this is or is not a BEP53 torrent?

aldenml:
I fail to see the nature of the problem (but I'm sure I'm missing something in your reasoning). I see this `&so=` similar to any other optional parameter like `&x.pe=`, if you discard the actual `add_torrent_params`, you will never know if peers connected were provided in the magnet.

I expect the UI can hold some information if relevant, also `add_torrent_alert` provide you with a copy of the original `add_torrent_params` (or at least I think it should)

Col-blimp:
The problem was that the magnet link was removed from the `add_torrent_params` as was `file_priorities()` from `torrent_handle` so I could not figure out how to ignore this BEP without adding an external list of magnet links, `get_file_priorities()` has been added in place of `file_priorities()` so that can be gone through to see if anything is not `default_priority`.

aldenml:
if you want to ignore it, just `(add_torrent_params)::file_priorities.clear()` should do the trick right after parsing any magnet.

arvidn:
it's still not clear to me exactly what you want to achieve that might require an external list of magnet links.

was it just the querying priorities of files? (if so, referring to it as a "bep53-torrent" is a bit misleading, since file priorities exists independently of this feature). Even with the ``get_file_priorities()``, you can't tell whether those priorities originated from a magnet link or not. In fact, once the metadata has been received, I don't believe you can tell that the torrent originated from a magnet link either.

Col-blimp:
Yes, I just wanted to know if the file priorities had been set by this BEP at the point of getting the `metadata_received_alert` so I could then set the corresponding UI dialog according to user choice.

PR #2579, opened 2017-11-30
added const to static arrays, minor refactor in alert.cpp

aldenml:


src/alert.cpp
```diff
@@ -271,7 +271,7 @@ namespace libtorrent {
 
 	std::string performance_alert::message() const
 	{
-		static char const* warning_str[] =
+		static char const* const warning_str[] =
```
aldenml: I wonder if this actually makes any difference in the generated code

PR #2580, opened 2017-12-01
reduce boilerplate in magnet link test

arvidn:
@aldenml looks ok?


aldenml:
yes, and a lot  better!

PR #2582, opened 2017-12-03
fix part-file creation race condition (back-port from master)

arvidn:


arvidn:
@aldenml looks reasonable?

aldenml:
I think part of this PR (in the constructor) is necessary, https://github.com/arvidn/libtorrent/pull/2490/files

arvidn:
ah, of course. thanks for the reminder!

PR #2583, opened 2017-12-03
merge RC_1_1 into master

arvidn:
it looks like the merge history disappeared again. this makes me miss subversion :(

this is what I did:

``git merge RC_1_1``
*fix merge conflicts and make sure all tests pass*
``git checkout -b merge-1.1.into-master``
``git commit -a -m "merge RC_1_1 into master"``
``git push origin merge-1.1.into-master``

Am I not allowed to create a branch for my merge before I commit? Should I have committed to master first then created the branch and pushed it there?

this really sucks

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -81,6 +81,7 @@
 	* resume data no longer has timestamps of files
 	* require C++11 to build libtorrent
 
+	* fix issue with initializing settings on session construction
 	* fix issue with receiving interested before metadata
 	* fix IPv6 tracker announce issue
 	* restore path sanitization behavior of ":"
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -266,10 +266,10 @@ namespace aux {
 			using connection_map = std::set<std::shared_ptr<peer_connection>>;
 			using torrent_map = std::unordered_map<sha1_hash, std::shared_ptr<torrent>>;
 
-			explicit session_impl(io_service& ios);
+			session_impl(io_service& ios, settings_pack const& pack);
 			~session_impl() override;
 
-			void start_session(settings_pack pack);
+			void start_session();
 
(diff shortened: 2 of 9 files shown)


arvidn:
@aldenml do you know how I'm supposed to do this? Is there any way I can recover the work I put into fixing conflicts?

arvidn:
it seems you can't rebase merge commits at all, so as far as I can tell, there's no way to stage or review merges. I have to push them straight into master.

aldenml:
subversion??? :) I honestly don’t have any idea

arvidn:
heh, in subversion the branch/merge metadata is stored in regular files and show up in diffs and doesn't get stripped so easily.

arvidn:
i guess going forward, I will create a local merge commit, but before I push it to master, I push it to a separate branch just to get the CI running on it.

aldenml:
That sounds like a good idea

ssiloti:
It looks like switching branches while a merge commit is staged in the index causes git to forget it was doing a merge. I would do this by switching to a new branch before starting the merge.

arvidn:
yeah, I think even in that case though, you can't land the merge-commit into master via a rebase, but you have to introduce *another* merge. As someone who really likes linear history, it doesn't seem quite worth it.

I could be wrong about that, but this PR right here does have the correct history (I re-did the merge preserving the changes I had already made), and it cannot be rebased into master

ssiloti:
If you haven't added any commits to master since the merge then you can move the head of master to the head of the temp branch with `rebase --hard`, but otherwise yeah you have to either create another merge commit or re-do the merge. I think the basic problem is that git can't pick apart a merge commit into the parts which came from each parent branch, so there's no way to automatically back out one parent and replace it with another.

If you were really determined you could use some low-level git incantations to manually re-write a rebased commit's parents, but that's probably not worth the effort.

PR #2591, opened 2017-12-04
Fix mailing list archives link. Closes #2588.

### LINKED ISSUE #2588: Mailing list archive links aren't working
From the main website the archive link correctly loads: http://dir.gmane.org/gmane.network.bit-torrent.libtorrent

But none of the links in that page work for me.
I don't know what the issue is, but in case you can't fix it you can point the initial link to the sf.net archives instead: https://sourceforge.net/p/libtorrent/mailman/libtorrent-discuss/

sledgehammer999:
I did only a string search in the repo. I don't actually know if those are the correct places.


arvidn:
I will merge rc_1_1 into master

PR #2593, opened 2017-12-06
minor cleanup, refactor and typos

aldenml:


src/gzip.cpp
```diff
@@ -180,8 +179,7 @@ namespace {
 	}
 	} // anonymous namespace
 
-	TORRENT_EXTRA_EXPORT void inflate_gzip(
-		span<char const> in
+	void inflate_gzip(span<char const> in
```
arvidn: this is exported in the header file already?

PR #2599, opened 2017-12-13
restore the parse_magnet_uri overload that amends an add_torrent_params object

arvidn:


aldenml:
is this due to some practical use case?

arvidn:
yeah, as I was converting some ``libtorrent-webui`` code especially, it was quite inconvenient as I had to copy over fields one-by-one. [Here's an example](https://github.com/arvidn/libtorrent-webui/blob/master/src/transmission_webui.cpp#L177).

PR #2606, opened 2017-12-16
fix reopening of listen sockets when disabling force-proxy

arvidn:


arvidn:
It seems I forgot to push to libsimulator

PR #2616, opened 2017-12-20
Fix local network address mappings.

sledgehammer999:
I assume your mappings aren't intentional.

About IPv4: Wikipedia shows different end address for the range. [Link](https://en.wikipedia.org/wiki/Private_network#Private_IPv4_address_spaces)
About IPv6: It is somewhat equivalent to the IP4 private ranges. [Link](https://en.wikipedia.org/wiki/IPv6_address#Unique_local_addresses)


arvidn:
I wonder where all those compiler warnings came from all of a sudden. maybe travis upgraded compiler

PR #2628, opened 2017-12-25
handle errors in peer_connection

arvidn:


src/peer_connection.cpp
```diff
@@ -3024,7 +3061,8 @@ namespace libtorrent {
 				{
 					// when this returns, all outstanding jobs to the
 					// piece are done, and we can restore it, allowing
-					// new requests to it
+					// new req3uests to it
```
aldenml: typo?

src/peer_connection.cpp
```diff
@@ -3024,7 +3061,8 @@ namespace libtorrent {
...
+					auto conn = self();
 					m_disk_thread.async_clear_piece(t->storage(), p.piece
```
aldenml: missing use of lambda with `conn->wrap` here?

arvidn: I started wrapping this call, but this is actually a call on the torrent, so I back-tracked. not all the way though it turns out :P, but maybe I should use the torrent wrap function instead.

PR #2630, opened 2017-12-26
fix empty state_update_alert, dht_stats_alert

pavel-pimenov:
log message: "state updates for 0 torrents"

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -4567,6 +4567,8 @@ namespace {
 		std::vector<torrent*>& state_updates
 			= m_torrent_lists[aux::session_impl::torrent_state_updates];
 
+		if (state_updates.empty()) return;
+
 #if TORRENT_USE_ASSERTS
 		m_posting_torrent_updates = true;
 #endif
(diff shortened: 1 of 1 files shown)


aldenml:
isn't the information of empty values useful in itself?

src/session_impl.cpp
```diff
@@ -4567,6 +4567,8 @@ namespace {
 		std::vector<torrent*>& state_updates
 			= m_torrent_lists[aux::session_impl::torrent_state_updates];
 
+		if (state_updates.empty()) return;
```
arvidn: if the user asks for an update, and doesn't get one, the invariant for this call becomes impossibly difficult to use.

should the client expect a response? there's no way of knowing, so should the client have some time-out? if it hasn't received a response in 3 seconds, assume there has been no updates. That has several problems. (1) update frequency may be difficult to keep high, as you may want to time-out your first request before issuing another (2) it's wrong, since on some devices under certain load it may take more than 3 seconds for a response to come back

arvidn:
a test to make sure that both of these calls unconditionally return a response would be useful. i.e. something to make this patch fail the test. The most difficult case is when the alert queue is full. alerts that are posted unconditionally are subject to a much higher alert queue limit to give them higher priority, but even they may be lost if alerts aren't being popped.

PR #2641, opened 2017-12-28
improve error handling during session shutdown

arvidn:


include/libtorrent/aux_/session_impl.hpp
```diff
@@ -277,6 +277,13 @@ namespace aux {
 
 			void init_peer_class_filter(bool unlimited_local);
 
+			void call_abort()
+			{
+				auto ptr = shared_from_this();
+				m_io_service.dispatch(make_handler([ptr] { ptr->abort(); }
+				, m_abort_handler_storage, *this));
```
aldenml: missing `\t`

src/torrent.cpp
```diff
@@ -8399,15 +8399,14 @@ namespace libtorrent {
 	}
 	catch (...) { handle_exception(); }
 
-	void torrent::on_torrent_aborted() try
+	void torrent::on_torrent_aborted()
```
aldenml: does it makes sense to remove the `try` above in `on_cache_flushed` too?

arvidn: ``on_cache_flushed`` is still doing things that could throw an exception though

PR #2644, opened 2017-12-28
complete the error handling test and make it part of the default simulation test suite

arvidn:
It will run a file transfer between two clients repeatedly, each time cause another memory allocation fail, until every single memory allocation has failed once. Any invariant check failure, assertion or signal will cause the test to fail

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -4361,8 +4361,23 @@ namespace libtorrent {
 		// the torrent object from there
 		if (m_storage)
 		{
-			m_ses.disk_thread().async_stop_torrent(m_storage
-				, std::bind(&torrent::on_torrent_aborted, shared_from_this()));
+			try {
+				m_ses.disk_thread().async_stop_torrent(m_storage
+					, std::bind(&torrent::on_torrent_aborted, shared_from_this()));
+			}
+			catch (std::exception const& e)
+			{
+				TORRENT_UNUSED(e);
+				m_storage.reset();
+#ifndef TORRENT_DISABLE_LOGGING
+				debug_log("Failed to flush disk cache: %s", e.what());
+#endif
+				// clients may rely on this alert to be posted, so it's probably a
+				// good idea to post it here, even though we failed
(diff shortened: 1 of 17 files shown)


src/torrent.cpp
```diff
@@ -4386,8 +4386,15 @@ namespace libtorrent {
...
 		{
-			m_ses.disk_thread().async_stop_torrent(m_storage
-				, std::bind(&torrent::on_torrent_aborted, shared_from_this()));
+			try {
+				m_ses.disk_thread().async_stop_torrent(m_storage
+					, std::bind(&torrent::on_torrent_aborted, shared_from_this()));
+			} catch (std::exception const&)
+			{
+				m_storage.reset();
+				if (alerts().should_post<cache_flushed_alert>())
```
aldenml: other than the missing `\n`s in `try` and `catch`, I wonder why `cache_flushed_alert` is the right alert here

arvidn: yeah, good question. perhaps there's a better approach here. I just did the same thing as the "else" block, but failing to flush the cache is kind of serious

arvidn:
I could use some help understanding these errors. I will try to reproduce the windows error, but the linux error is puzzling. It's aborting because of an exception (it doesn't say it was uncaught, but I suppose that's what they mean). As far as I can tell, there's no ``noexcept`` function on the stack, nor a destructor (in which case ``std::terminate`` would be expected). But there is a catch handler further down the stack. (wrapping all of ``run_test()``).

aldenml:
Do you think this gcc bug is relevant? https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80683

arvidn:
yeah, it might be. I just tracked down the issue one windows to having debug iterators enabled. From what I can tell, the ``std::vector`` default constructor is marked ``noexcept``, even when debug iterators are enabled, and the default constructor allocates memory. When this allocation fails, it gets to the noexcept constructor and terminates.

Hopefully that bug is related to member initializers (which could be worked around) and the issue here is also related to debug iterators.

PR #2650, opened 2017-12-29
cleanup examples a bit

arvidn:


aldenml:
are you moving to trailing `{`?

arvidn:
I don't have any plans on changing formatting of libtorrent itself, but I think it makes sense to make examples a but more succinct.

PR #2655, opened 2017-12-31
upgrade OSX image on travis

arvidn:


aldenml:
I’m a little surprised that you are using the latest image, I was under the impression that you wanted to use the oldest possible/reasonable one, also, I remember that starting from 8.x, having problems with the code signing. Just a comment.

arvidn:
Well, in the new-disk-io branch I need a newer version of clang, but xcode8.3 is sufficient. Perhaps going straight to the latest version is a bit overkill, but on the other hand, I won't have to update it in a while then presumably.

arvidn:
Perhaps I'm inconsistent in my opinions on which version to run CI on. There are good arguments for running on the oldest supported version as well as the latest version. I think running on the latest version is a pretty strong argument, that (as a user) upgrading should be safe, and things should (ideally) be more likely to work the later version.

PR #2659, opened 2018-01-03
using make_address instead of deprecated from_string when boost>=1.66

aldenml:
I wanted to change the code only inside core libtorrent, and not in tests and bindings to easy the diff

### DIFF
--- a/include/libtorrent/address.hpp
+++ b/include/libtorrent/address.hpp
@@ -65,13 +65,10 @@ namespace libtorrent {
 // the from_string member functions are deprecated starting
 // in boost 1.66.0
 #if BOOST_VERSION >= 106600 && !defined TORRENT_BUILD_SIMULATOR
-	inline address make_address(string_view str, boost::system::error_code& ec)
-	{ return boost::asio::ip::make_address(str.data(), ec); }
-	inline address_v4 make_address_v4(string_view str, boost::system::error_code& ec)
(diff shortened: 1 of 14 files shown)


include/libtorrent/address.hpp
```diff
@@ -58,6 +61,28 @@ namespace libtorrent {
 	using address_v6 = boost::asio::ip::address_v6;
 #endif
 #endif // SIMULATOR
+
+// the from_string member functions are deprecated starting
+// in boost 1.66.0
+#if BOOST_VERSION >= 106600 && !defined TORRENT_BUILD_SIMULATOR
+	inline address make_address(string_view str, boost::system::error_code& ec)
```
arvidn: Instead of having inline functions here, I think you could just do:

```
using boost::asio::ip::make_address;
using boost::asio::ip::make_address_v4;

```
etc. That way you would get the overloads that throw too (which I expect exist).

aldenml: one question: I thought about it, but I was asking myself if pulling the overloads that throw is a good idea. Don't  you prefer the ones that use the `error_code` to avoid  unexpected exceptions?

arvidn: I prefer exceptions for unexpected errors (and ``error_code`` for expected errors)

aldenml: OK, I will do the `using`

PR #2664, opened 2018-01-03
propagate the openssl include directory to clients via pkg-config

arvidn:


sledgehammer999:
This doesn't work. In fact OPENSSL_INCLUDES is empty. But somehow libtorrent manages to compile. How is that possible? Does it pick up the system's deprecated openssl?

If you can't reproduce on your system, please give some instructions on how to debug this on macOS.

arvidn:
I'm far from an expert on autoconf. I don't know how it's supposed to work. I suppose brew puts libraries in ``/usr/local/lib``, which probably is a default search path though.

sledgehammer999:
>I'm far from an expert on autoconf. I don't know how it's supposed to work

Ok.
If you want to allow propagation of openssl flags, then I can take a look into it.

PR #2666, opened 2018-01-04
fix noexcept marking on entry, and make move assignment

arvidn:


aldenml:
do you think this could help with some weird memory crashes when creating `entry`?

arvidn:
I doubt it, the main effect of this being wrong is that *if* there's an exception when constructing the underlying object, std::terminate is called instead of the exception propagating up.

Issue #1757, opened 2017-02-25
Merkle


Col-blimp:
```
assertion failed. Please file a bugreport at https://github.com/arvidn/libtorrent/issues
Please include the following information:

version: 1.2.0.0-62cdec8

#: 1
file: 'src\torrent.cpp'
line: 334
function: libtorrent::torrent::torrent
expression: <unconditional>

stack:
 0: 00007FF796139297 print_backtrace +375 
```

Assertion occurs when adding a torrent created as a merkle torrent but without passing in the merkle tree, I assumed that in the absence of passing in the tree LibTorrent would assume it's downloading and would get what's necessary from peers.

asserts.log when this error is generated is created in the location of the .torrent file not the location of the application .exe, is this expected behaviour?

`settings_pack::support_merkle_torrents` is set to true, when set to false `add_torrent_alert` adds the torrent without the assertion and error is `error_code_enum::no_error` and that looks wrong given merkle is unsupported?

arvidn:
If you have a chance, it would be really helpful if you could write a test that would reproduce this issue (probably by extending ``simulations/test_swarm.cpp``)

Col-blimp:
The simple client and client test produce the error, a sample torrent is here: http://s000.tinyupload.com/index.php?file_id=96868270061824021153

it's just LibTorrent Master created as a merkle torrent, with the default test the tree is not added in but I do get other errors once it's added in and resumed but I think they are related to this issue.

arvidn:
an automated test would be very helpful. I tested merkle torrents a few years ago, and they worked. The reason why they don't anymore is because there's no (automated) test.

Col-blimp:
I shall make one.

Col-blimp:
Out of curiosity is there any ready made visual studio projects for the testing and simulation?

arvidn:
no, and the CMakeLists.txt in the root, I don't think, builds the simulator.

PR #2669, opened 2018-01-05
raise auto piece size selection limit to 16 MB in create_torrent()

Chocobo1:
Related issue in qbt: https://github.com/qbittorrent/qBittorrent/issues/8205

I've read various guides on the net, however they doesn't put out suggestions for large torrents (size in tens/hundreds of GB).
I'm only partially aware of its consequences, I definitely need some advice and comments welcome!

ps. This one is targeting RC_1_1 branch, I'll open another one for master branch after merged

### DIFF
--- a/src/create_torrent.cpp
+++ b/src/create_torrent.cpp
@@ -332,16 +332,20 @@ namespace libtorrent
 		// a piece_size of 0 means automatic
 		if (piece_size == 0 && !m_merkle_torrent)
 		{
-			const int target_size = 40 * 1024;
-			piece_size = int(fs.total_size() / (target_size / 20));
+			// size_table is computed from the following:
+			//   target_list_size = sqrt(total_size) * 2;
+			//   target_piece_size = total_size / (target_list_size / hash_size);
+			// Given hash_size = 20 bytes, target_piece_size = (16*1024 * pow(2, i))
+			// we can determine size_table = (total_size = pow(2 * target_piece_size / hash_size, 2))
+			boost::int64_t const size_table[] = {2684355, 10737418, 42949673, 171798692, 687194767,
+				2748779069LL, 10995116278LL, 43980465111LL, 175921860444LL, 703687441777LL};
 
-			int i = 16*1024;
-			for (; i < 2*1024*1024; i *= 2)
+			int i = 0;
+			for (int max = sizeof(size_table) / sizeof(size_table[0]); i < max; ++i)
 			{
-				if (piece_size > i) continue;
-				break;
+				if (size_table[i] >= fs.total_size()) break;
 			}
-			piece_size = i;
+			piece_size = 0x4000 << i;
 		}
 		else if (piece_size == 0 && m_merkle_torrent)
 		{


Chocobo1:
I just realized that BEP 30 (Merkle hash torrent extension) is supposed to remedy this kind of issue...
However is the support widespread enough so we can recommend it to users?
Does major torrent clients support it yet?

arvidn:
I agree that 2 MiB piece sizes are perhaps a bit small, but 16 MiB feels quite large as well.
Here's my model for reasoning about what a reasonable piece size is:

The larger piece size, the longer it will take a peer to complete downloading a piece. The whole piece need to be downloaded in order to check its hash and checking its hash is required before offering it up to other peers. Larger piece sizes mean longer delays for payload to propagate from one peer to the next, slowing down the dissemination in general.

So, how much worse does it gets as the piece size increases? It's hard to say without proper measurements, but it probably depends on the general download speed of the peers in the swarm. i.e. the most relevant property is probably not the piece size directly, but the *piece download time*.

Do you have any references to what people on the internet recommend?

Another aspect to take into account is whether torrents are expected to primarily be distributed via magnet links or not. If they are, a large torrent file may not be that big of a deal.

Another aspect is to keep bittorrent v2 in mind, although that's probably some way out, I would expect the piece size to be a lot smaller with a merkle tree. But perhaps these concerns are disconnected.

Col-blimp:
8mb and 16mb are useful for large file compilations, I've got some users that needed 32mb as they were sending torrent files via email for transferring large files over SSL torrents and in these cases it's not really about the swarm as only certain people could get the torrents anyway . Merkle is a solution but its not a default build in Libtorrent and there is a bug in Libtorrent where you can create Merkle torrents just fine but they wont get added to the [session](https://github.com/arvidn/libtorrent/issues/1757).

ssiloti:
Bittorrent v2 doesn't change the calculus for piece size selection that much. The only concerns v2 eliminates are the up-front overhead of downloading the piece hashes when using magnet links and the amount of data wasted on hash fails.

It seems to me that the current algorithm is too aggressive about increasing the piece size as torrents grow into the gigabytes and beyond. I wonder if we should scale the target piece list size with something like the square root of the total size. Maybe put a floor on it so that small torrents don't sacrifice piece size for an insignificant savings on metadata size. For example, if we used `target_size = sqrt(total_size) * 2`:

4GB torrent = 1MB piece size = 80KB piece list
16GB = 2MB = 160KB
128GB = 4MB = 640KB
1TB = 16MB = 1.25MB

I think it's a safe assumption that anyone on on 1TB torrent has a connection fast enough for 16MB pieces to be reasonable.

Chocobo1:
>Do you have any references to what people on the internet recommend?

Did a quick search on google and found these:
https://wiki.vuze.com/w/Torrent_Piece_Size
http://torrentinvites.org/f29/piece-size-guide-167985/

>8mb and 16mb are useful for large file compilations to get around torrent file size limits on sites

Yeah, in the second guide says 1 MB torrent files is a common limitation, maybe it should be taken into consideration in this PR?
Yet I also think 1MB limit is a bit silly for a mega size content (1 TB), however this is tracker operators problem.

>I wonder if we should scale the target piece list size with something like the square root of the total size.  Maybe put a floor on it so that small torrents don't sacrifice piece size for an insignificant savings on metadata size.

Agree, I was thinking about it too.

>For example, if we used target_size = sqrt(total_size) * 2:

Seems you used `1.25` instead of `2` in the examples?
I think the results is not bad.

arvidn:
I also like the square root formula. @Chocobo1 what do you think about updating this PR to something like what @ssiloti suggested?

One thing to keep in mind, I think it's best to keep piece sizes to even powers of 2. Technically the specification doesn't mandate this and libtorrent supports arbitrary piece sizes, but it's a bit funky.

The question then is what should the coefficient be? is 2 right?

ssiloti:
@Chocobo1 It looks like the coefficient is smaller than 2 because we round the piece size up to the nearest power of 2, so the actual piece list size comes out smaller than the target. Bittorrent v2 requires that the piece size be a power of 2.

Chocobo1:
>@Chocobo1 what do you think about updating this PR to something like what @ssiloti suggested?

Okay, I've updated the PR.

Some results for small size torrents:
torrent content, determined piece size, piece list size, piece count
1MB, 16KB, 1KB, 64
10MB, 64KB, 3KB, 160
100MB, 256KB, 7KB, 400
200MB, 256KB, 15KB, 800
700MB, 512KB, 27KB, 1400

arvidn:
https://ci.appveyor.com/project/arvidn/libtorrent/build/1.0.8084/job/n2hjudil8ojof93d#L376

src/create_torrent.cpp
```diff
@@ -332,14 +332,15 @@ namespace libtorrent
 		// a piece_size of 0 means automatic
 		if (piece_size == 0 && !m_merkle_torrent)
 		{
-			const int target_size = 40 * 1024;
-			piece_size = int(fs.total_size() / (target_size / 20));
+			const int hash_size = 20;
+
+			double target_list_size = sqrt(fs.total_size()) * 2;
```
arvidn: this should be ``std::sqrt()``

Chocobo1: fixed.

arvidn: https://ci.appveyor.com/project/arvidn/libtorrent/build/1.0.8089/job/3r5ac86p4a47qhmw#L238

Chocobo1: >Would it make sense to just have a table instead?

Indeed, PR updated.
I try keeping `sqrt()` for master branch.

src/create_torrent.cpp
```diff
@@ -332,16 +332,20 @@ namespace libtorrent
...
+
+			int i = 0;
+			for (int max = sizeof(size_table) / sizeof(size_table[0]); i < max; ++i)
 			{
-				if (piece_size > i) continue;
-				break;
+				if (size_table[i] >= fs.total_size()) break;
 			}
-			piece_size = i;
+			piece_size = 16*1024 * pow(2.0, i);
```
arvidn: I would really prefer to not have any floating point operations, especially if they're unnecessary.
In this case for instance, you could just do: ``piece_size = 0x4000 << i``, if I'm not mistaken.

arvidn:
I think this table-approach is better and simpler than using ``std::sqrt()`` in master too.

src/create_torrent.cpp
Chocobo1: Changed.

PR #2675, opened 2018-01-07
fix V821 Decreased performance.

pavel-pimenov:
The 't' variable can be constructed in a lower level scope. bt_peer_connection.cpp 417

### DIFF
--- a/src/bt_peer_connection.cpp
+++ b/src/bt_peer_connection.cpp
@@ -414,14 +414,19 @@ namespace {
 
 		if (!m_supports_fast) return;
 
+#if TORRENT_USE_ASSERTS
+		std::shared_ptr<torrent> t = associated_torrent().lock();
+		TORRENT_ASSERT(t);
+		TORRENT_ASSERT(t->valid_metadata());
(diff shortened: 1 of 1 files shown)


src/bt_peer_connection.cpp
```diff
@@ -414,13 +414,13 @@ namespace {
 
 		if (!m_supports_fast) return;
 
-		std::shared_ptr<torrent> t = associated_torrent().lock();
```
aldenml: I don't think this is correct, because the asserts are not related to the logging. Maybe what do you need here is a `#if TORRENT_USE_ASSERTS` block

pavel-pimenov: default

?

arvidn:
as is, this would break the build with ``logging=on`` and ``asserts=off``. The second #ifdef block is wrong I believe

pavel-pimenov:
hm...
#if !TORRENT_USE_ASSERTS wrong?

arvidn:
oh, I see. it's negated

PR #2677, opened 2018-01-07
fix V560 & V547

pavel-pimenov:
V560 A part of conditional expression is always false: idx < 0. alert.cpp 1557
V560 A part of conditional expression is always false: j < 0. block_cache.cpp 232
V560 A part of conditional expression is always false: pe->num_dirty == 0. disk_io_thread.cpp 1018
V560 A part of conditional expression is always true: m_connecting. peer_connection.cpp 374
V547 Expression 'recv_pos >= 13' is always true. bt_peer_connection.cpp 1050
V547 Expression 't' is always true. peer_connection.cpp 5294
V547 Expression 'm_abort' is always false. session_impl.cpp 5707

### DIFF
--- a/src/alert.cpp
+++ b/src/alert.cpp
@@ -1548,7 +1548,7 @@ namespace {
 		};
 
 		int const idx = static_cast<int>(op);
-		if (idx < 0 || idx >= int(sizeof(names) / sizeof(names[0])))
+		if (idx >= int(sizeof(names) / sizeof(names[0])))
 			return "unknown operation";
 
 		return names[idx];
--- a/src/alert.cpp
+++ b/src/alert.cpp
@@ -1548,7 +1548,7 @@ namespace {
 		};
 
 		int const idx = static_cast<int>(op);
-		if (idx >= int(sizeof(names) / sizeof(names[0])))
+		if (idx < 0 || idx >= int(sizeof(names) / sizeof(names[0])))
 			return "unknown operation";
 
 		return names[idx];
--- a/src/block_cache.cpp
+++ b/src/block_cache.cpp
@@ -229,7 +229,7 @@ static_assert(int(job_action_name.size()) == static_cast<int>(job_action_t::num_
 	char const* job_name(job_action_t const job)
 	{
 		int const j = static_cast<int>(job);
-		if (j >= piece_log_t::last_job)
+		if (j < 0 || j >= piece_log_t::last_job)
 			return "unknown";
 
 		if (j < piece_log_t::flushing)
--- a/src/block_cache.cpp
+++ b/src/block_cache.cpp
@@ -229,7 +229,7 @@ static_assert(int(job_action_name.size()) == static_cast<int>(job_action_t::num_
 	char const* job_name(job_action_t const job)
 	{
 		int const j = static_cast<int>(job);
-		if (j < 0 || j >= piece_log_t::last_job)
+		if (j >= piece_log_t::last_job)
 			return "unknown";
 
 		if (j < piece_log_t::flushing)
--- a/src/bt_peer_connection.cpp
+++ b/src/bt_peer_connection.cpp
@@ -1047,21 +1047,20 @@ namespace {
 				received_bytes(0, received);
 				return;
 			}
+			TORRENT_ASSERT(recv_pos >= 13);
+			char const* ptr = recv_buffer.begin() + 9;
+			int const list_size = detail::read_int32(ptr);
+
+			if (list_size > m_recv_buffer.packet_size() - 13)
 			{
-				char const* ptr = recv_buffer.begin() + 9;
-				int const list_size = detail::read_int32(ptr);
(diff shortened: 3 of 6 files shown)


src/alert.cpp
```diff
@@ -1548,7 +1548,7 @@ namespace {
 		};
 
 		int const idx = static_cast<int>(op);
-		if (idx < 0 || idx >= int(sizeof(names) / sizeof(names[0])))
```
arvidn: I don't the warning you see here is correct. The cast of ``op`` to an ``int`` can produce a negative value, as far as I can see.

src/bt_peer_connection.cpp
```diff
@@ -1047,7 +1047,6 @@ namespace {
 				received_bytes(0, received);
 				return;
 			}
-			if (recv_pos >= 13)
```
arvidn: if this is always true, it should probably be an assert

src/peer_connection.cpp
```diff
@@ -4872,7 +4872,7 @@ namespace libtorrent {
 			&& m_reading_bytes == 0
 			&& !m_choked
 			&& m_peer_interested
-			&& t && t->is_upload_only()
```
arvidn: where is ``t`` checked?

pavel-pimenov: https://github.com/arvidn/libtorrent/blob/88643dd121d8d06c805fefae20216e346b2f9794/src/peer_connection.cpp#L4740

src/alert.cpp
pavel-pimenov: enum class operation_t : std::uint8_t

src/peer_connection.cpp
arvidn: ah, line 4740

src/session_impl.cpp
```diff
@@ -5699,9 +5699,8 @@ namespace {
 		}
 
 #ifndef TORRENT_DISABLE_LOGGING
-		session_log("starting DHT, running: %s, router lookups: %d, aborting: %s"
```
arvidn: I'm not sure about this. it makes the log line look a bit deceitful.

src/block_cache.cpp
```diff
@@ -229,7 +229,7 @@ static_assert(int(job_action_name.size()) == static_cast<int>(job_action_t::num_
 	char const* job_name(job_action_t const job)
 	{
 		int const j = static_cast<int>(job);
-		if (j < 0 || j >= piece_log_t::last_job)
```
arvidn: I think the same thing applies here. there are ways for ``j`` to be negative. my understanding is that casting a large ``unsigned int`` into an ``int`` is not undefined, just implementation defined

pavel-pimenov: auto const j = static_cast<std::uint8_t>(job);
?

src/session_impl.cpp
pavel-pimenov: revert?

arvidn: I think that would be best

src/alert.cpp
arvidn: I see. However, I like that the current logic here does not *depend* on the underlying storage type of the enum class. I worry about the commit where the 256th operation enum is added, and this type has to be widened, it would be virtually impossible to find all places that depend on it being no more than 8 bits.

In addition, I would expect the compiler to optimise-away this test, so this change is kind of only to please a warning.

I suppose this change would make sense combined with:

``static_assert(std::numeric_limits<std::underlying_type<operation_t>::value>::min() >= 0
&& std::numeric_limits<std::underlying_type<operation_t>::value>::max() <= std::numeric_limits<int>::max(), "an check for negative values is necessary");``.

But it's not obvious to me that that would be an improvement to the code. As it is now, a reader does not need to worry about any of this.

src/block_cache.cpp
arvidn: I think the reasoning here should be the same as the one above, about ``operation_t``

src/session_impl.cpp
```diff
@@ -5700,7 +5700,7 @@ namespace {
 
 #ifndef TORRENT_DISABLE_LOGGING
 		session_log("starting DHT, running: %s, router lookups: %d, aborting: %s"
-			, m_dht ? "true" : "false", m_outstanding_router_lookups
+			, m_dht ? "true" : "false", m_outstanding_router_lookups);
```
arvidn: you added an extra ");" here.
syntax error: https://ci.appveyor.com/project/arvidn/libtorrent/build/8105/job/s6l498mo8axjsclp#L243

arvidn:
there's this now. is there perhaps some threading subtlety being broken now?
https://travis-ci.org/arvidn/libtorrent/jobs/326163185#L1783

pavel-pimenov:
no versions :-(

arvidn:
what do you mean by "no versions"?

pavel-pimenov:
https://travis-ci.org/arvidn/libtorrent/jobs/326163185#L1783

PR #2680, opened 2018-01-08
Test merge

arvidn:


src/create_torrent.cpp
```diff
@@ -328,16 +328,20 @@ namespace {
...
-			piece_size = int(fs.total_size() / (target_size / 20));
-
-			int i = 16*1024;
-			for (; i < 2*1024*1024; i *= 2)
+			// size_table is computed from the following:
+			//   target_list_size = sqrt(total_size) * 2;
+			//   target_piece_size = total_size / (target_list_size / hash_size);
+			// Given hash_size = 20 bytes, target_piece_size = (16*1024 * pow(2, i))
+			// we can determine size_table = (total_size = pow(2 * target_piece_size / hash_size, 2))
+			boost::int64_t const size_table[] = {2684355, 10737418, 42949673, 171798692, 687194767,
```
aldenml: if for any reason you need to add something, maybe you can change it to use `std:: int64_t`

arvidn: oh, yes. thanks!

Issue #2674, opened 2018-01-07
Priority-0 files are not checked when `resume_data` is used


fbriere:
libtorrent version (or branch): 1.1.5 and 1.1.6
platform/architecture: amd64

There seems to be a bug when loading from resume data a torrent that was partially selected and downloaded without a part file, where libtorrent will not properly check any file unselected for download.

In other words, under these conditions:

- A torrent is initiated with `resume_data`;
- Some files within that torrent have been set to priority 0;
- Some of those files are present on disk, with (partial) data; and
- No part file is present;

Then libtorrent will not properly check any of the files with priority 0.

I'm guessing this is a common scenario when migrating partially-selected torrents from 1.0 to 1.1, and is probably one of two possible causes for qbittorrent/qBittorrent#7802.

In case this might be of use, I've uploaded (https://gist.github.com/fbriere/ed25285dacd14d842ee98391e8e3866e) a (crappy) Python script and torrent that demonstrates this issue.  Simply run it three times, with `1`, `2` and `3` as respective argument; the third run will not start at 100% like it should.

arvidn:
if the files have priority set to zero, libtorrent will pretend they don't exist and won't check whether they exist. But, as long as they remain having priority set to zero, is that a problem?
Even if their priorities are updated to > 0, will the effect be worse than re-downloading the data?

arvidn:
is it the case that the resume data indicates the edge-pieces exist, even though they don't? i.e. the pieces that would have been written to a partfile.

fbriere:
It is indeed a problem if the torrent was already at 100% and seeding; things will work fine for a while, but as soon as a peer requests a piece that overlaps with one of those files, libtorrent will raise an error.  Re-downloading the (not actually missing) data will temporarily fix the issue, but it will reappear the next time the torrent is loaded.

(I myself eventually resorted to recompiling qBittorrent with 1.0 because of this issue; every partial torrent I was seeding became an "I/O Error" ticking bomb under 1.1.)

As for analyzing the resume data, I currently lack the knowledge and expertise for this.  I'd rather not invest too much time if you can easily recreate this bug for yourself; if not, well, let me know and I'll see what I can do. :smile:

fbriere:
I have to slightly amend my initial description of this issue: the lack of a part file is not actually required to trigger it.  (I was under the misguided impression that unselecting a downloaded file would bring it back into the part file.)  I have amended the example script correspondingly.

fbriere:
[...]

PR #2695, opened 2018-01-14
fix issue resuming 1.0.x downloads with a file priority 0

arvidn:


sledgehammer999:
This PR is related I think so I'll state my problem here.
It isn't easy to replicate but I encounter a problem with multi-file torrents occasionally.
1. Download **some** files from a multi-file torrent completely.
2. Then set 0 priority to one of them. Make sure that this file shares edge pieces with another one.
3. Start seeding.
4. When an edge piece from the 0 priority file is requested `file_error_alert` is reported. The `message()` of the alert returns this:
```
<torrent name> partfile read (<path to 0 priority file>) error: No such file or directory
```
Notice that although it talks about partfile read it points to the requested file instead.
Because I don't trigger the bug often and easily my step 4 observations might be off. For example, it might be triggered because it tries to **uncoditionally** look into the parts file for 0 priority files even if they are downloaded fully and regardless if an edge piece is requested.

**Another quirk about qbt:** When a user unselects a file(0 priority) we call `torrent_handle::rename_file()` on that file moving it into a hidden subdirectory called `.unwanted`. That's where the alert points to (along with the filename).

@arvidn it would be nice if you could check the code to verify if it does "the right thing" in the above scenarios too.

arvidn:
I believe libtorrent does not do the right thing in this case. the logic to find the block is tied to the file priority. If the file priority changes, it expects to find the block in a different place. But as you may have noticed, importing files into the part file is not supported.

arvidn:
I think this patch as-is though, still is a useful improvement over the current behavior, especially when upgrading.

sledgehammer999:
>I think this patch as-is though, still is a useful improvement over the current behavior, especially when upgrading.

Ok, then consider at least merging this PR today. I am planning to do a release myself so it would be nice if I could include this. (I build libtorrent RC_1_1 for windows/macOS release, and for Ubuntu PPA).

Unless, you aren't yet confident in this PR. I don't want to rush you.

sledgehammer999:
@arvidn I should have run it earlier. Anyway, today as I prepared my new release I used latest RC_1_1 and then did a test run on my machine using my torrents.
2 of them came back with 0% progress as if the fastresume was rejected. I force rejected and they went to 100%. After client restart same problem. Even though it looks like a fastresume rejection I don't the relevant alert.
In any case, I test with this commit and the previous one. This is the offending one. The previous commit doesn't reject the fastresume after client restart.

Unfortunately one of the torrents during the downgrade finished as 99% and I can seem to get the rest of pieces from peers. The other one still fine. But I fail to see any connection between the 2. Both of them were multifile torrents with some files unselected(0 priority).

How should I proceed with this? Any idea why a fastresume rejection alert isn't posted?

arvidn:
ok, so basically you're seeing the same behavior as documented here: https://github.com/arvidn/libtorrent/issues/2674#issuecomment-364246854

I'll take a look.

PR #2699, opened 2018-01-15
build for iOS on travis

arvidn:


aldenml:
don't you think that if this is something not supported, it could eventually trigger the banning of apps using this feature? I rather use a more "platform accepted" solution

arvidn:
This is the only solution i’ve found to the question of enumerating routes on iOS

aldenml:
I have the impression that deep down in the discussion link of https://github.com/arvidn/libtorrent/pull/2546, there is an iOS specific solution, but I didn't take a close look at it

aldenml:
I want to add that I have no idea how the rejection policy works for iOS and I just found everyone seems to do this very same hack.

PR #2700, opened 2018-01-15
fixing warnings in tests code, part1

aldenml:


test/test_upnp.cpp
```diff
@@ -115,14 +117,14 @@ struct callback_info
...
-namespace
+namespace // TODO: remove this nested namespace
 {
 	struct upnp_callback : aux::portmap_callback
 	{
 		void on_port_mapping(port_mapping_t const mapping
 			, address const& ip, int port
 			, portmap_protocol const protocol, error_code const& err
-			, portmap_transport const transport) override
+			, portmap_transport const /*transport*/) override
```
arvidn: I prefer just removing the name over commenting it out. (if the comment would help, it suggests the type isn't specific enough, which I would consider the underlying problem). Here it's quite clear what the parameter means without the name

aldenml: OK

aldenml:
travis is having a really hard time lately :)

PR #2705, opened 2018-01-19
fix python bindings

arvidn:


bindings/python/test.py
```diff
@@ -204,6 +204,11 @@ def test_read_resume_data(self):
 
         ses = lt.session({'alert_mask': lt.alert.category_t.all_categories})
         h = ses.add_torrent(tp)
+        print('add_torrent_params:')
```
aresch: Maybe it would be best to try accessing all of the properties to ensure none of them break in the future?

for attr in dir(tp):
  print('%s: %s' % (attr, getattr(tp, attr)))

bindings/python/src/session.cpp
```diff
@@ -760,7 +760,7 @@ void bind_session()
         .add_property("url_seeds", PROP(&add_torrent_params::url_seeds))
         .add_property("peers", PROP(&add_torrent_params::peers))
         .add_property("banned_peers", PROP(&add_torrent_params::banned_peers))
-        .add_property("unfinished_pieces", PROP(&add_torrent_params::unfinished_pieces))
+        .add_property("unfinished_pieces",PROP(&add_torrent_params::unfinished_pieces))
```
aresch: This looks unintentional.

arvidn: yes it was

aresch:
I just tested this and it works for me.

PR #2708, opened 2018-01-19
fix announce_endpoint::reset

pavel-pimenov:


src/announce_entry.cpp
```diff
@@ -91,8 +91,7 @@ namespace libtorrent {
 	void announce_endpoint::reset()
 	{
 		start_sent = false;
-		next_announce = time_point32::min();
-		min_announce = time_point32::min();
+		min_announce = next_announce = time_point32::min();
```
aldenml: just a comment, my personal preference is the more explicit nature of per field value set

pavel-pimenov: next_announce = time_point32::min();
min_announce = next_announce; 
?

aldenml: I rather do nothing, but is this a problem of performance? or just not to repeat the expression?

pavel-pimenov: https://www.viva64.com/en/w/v656/print/

V656 Variables 'next_announce', 'min_announce' are initialized through the call to the same function. It's probably an error or un-optimized code. Consider inspecting the 'time_point32::min()' expression. Check lines: 94, 95. announce_entry.cpp 95

default

aldenml: well...it's not my call :)

ssiloti:
This is an optimization, not a fix. In fact I don't think it's even an optimization. `time_point32::min` is a trivial inline function which constructs a `time_point` from a static constant. It's very likely that the compiler will optimize both the existing code and the proposed optimization down to exactly the same machine code.

src/announce_entry.cpp
arvidn: ``time_point32::min()`` is a constant (``constexpr`` function). There's no call. The diagnostic suggesting it's "unoptimized" is just not sophisticated enough. As for it being a potential mistake, I can see why that might justify the warning, but it isn't a mistake in this case.

Issue #2702, opened 2018-01-18
Disable anonymous mode for private torrents


sledgehammer999:
Hey guys,

one tracker operator contacted me saying that anonymous mode is causing problems and they will be forced to ban qbittorrent if no action is taken.

From further discussions, it seems their problem is the obfuscating effects of anonymous mode, which garble the peer-id. They claim that in qbittorrent v3_3_x (which uses RC_1_0) there was no problem.
So I am assuming that this problem targets RC_1_1 only.

They basically need to identify the clients properly in order to keep correct stats and keep cheaters out. This concerns **private** torrents.

My request is to disable anonymous mode for torrents with the private flag set. Is that possible to implement? Is it wise to implement? Will it create other problems?

arvidn:
this only applies to the HTTP user agent, right? or do they think it's important to advertise client version to peers too?

sledgehammer999:
From what I gather their tracker looks at the peer-id, which they see as garbled/random data. eg `&!H"GHJ"`

arvidn:
I realise my question wasn't quite what I meant. As far as you know, do these trackers only look at the HTTP tracker request? (which does include the peer id) or do they also connect back over the bittorrent protocol to look at the peer-id advertised there? (at least back in the day, some trackers would detect if peers were behind NAT by connecting back).

If it's just the HTTP request, it wouldn't be too much of an privacy intrusion.

sledgehammer999:
Hmm, good question. I just emailed the person who contacted me with your question. I'll return with their answer.

sledgehammer999:
From the reply they gave they seem to also analyze the swarm itself too.
They also mentioned that sometimes even when anon-mode is explicitly disabled via the settings, qbt keeps sending random bytes as peer-id. Could this be a bug?

arvidn:
Yes, if you enable anonymous mode, the peer id is clobbered with random bytes. if you disable anonymous mode, it won't be recreated (if you change the client fingerprint setting it will be though).

This is a bug that I'll fix.

As for the main topic of this thread, I found this [related patch](https://github.com/arvidn/libtorrent/commit/e37e4bfe8abccd0af275b9131cf0ddcb9cc15e80).

sledgehammer999:
>As for the main topic of this thread, I found this related patch.

As far as I can tell this patch affects user-agent only and not the peer-id. These people are messaged me about the peer-id.

arvidn:
[...]

PR #2715, opened 2018-01-21
use unique peer_ids per connection

arvidn:
attempt to address https://github.com/arvidn/libtorrent/issues/2702

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,5 @@
 
+	* use unique peer_ids per connection
 	* fix iOS build on recent SDK
 	* fix tracker connection bind issue for IPv6 trackers
 	* fix error handling of some merkle torrents
--- a/bindings/python/src/session.cpp
+++ b/bindings/python/src/session.cpp
@@ -973,7 +973,6 @@ void bind_session()
         .def("pause", allow_threads(&lt::session::pause))
         .def("resume", allow_threads(&lt::session::resume))
         .def("is_paused", allow_threads(&lt::session::is_paused))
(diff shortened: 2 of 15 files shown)


sledgehammer999:
Will the announces to trackers use the same peer-id or unique each time?
How do other peers decide how to allocate to us up/down speeds and slots? Based on IP, peer-id or some other metric? I am concerned about peers not recognizing us "as good citizens" when we ask for more pieces.

arvidn:
each torrent will generate a "torrent-level" peer-id which is used for all tracker announces. If the tracker includes peer-ids in the response (this is the original, verbose, tracker response format, and i2p may do this as well) this peer-id will be used to avoid connecting to ourself.

Each peer connection then generates a unique peer-id which is primarily used to settle race conditions of two peers connecting to each other at the same time. As a way for both peers to agree on which connection should be closed.

Almost all peer "management" is done at the IP level, for all major clients. Since the peer-id or source port etc. all are trivial to generate, the typical setting allow only a single peer connection per IP. Similarly, transfer stats, optimistic unchoke timers, and choking logic is all based on peer connection and/or IP address. (again, since otherwise it would be trivial to "spoof" peer-ids to get an advantage).

I'm not aware of anything important the major clients use peer-id for. The most "important" I can think of is uTorrent using it to detect clients spoofing it's fingerprint.

arvidn:
(but in the case of uT, the only effect of this detection is to append "(fake)" in the peer list in the GUI, so not terribly important from the network's point of view)

PR #2727, opened 2018-01-25
fixing warnings in tests code, part4

aldenml:


arvidn:
thanks for doing this!

aldenml:
👍

aldenml:
do you recognize the ssl error in macOS in travis?

arvidn:
I think this is normal flakiness of the tests that use real sockets

PR #2729, opened 2018-01-26
avoid repeated calls to .address(), minor refactor

aldenml:


include/libtorrent/torrent_peer.hpp
```diff
@@ -276,4 +276,3 @@ namespace libtorrent {
 }
 
 #endif
-
```
d-komarov: Don't remove newline at the end of file, git won't like this.
I would recommend reading [Newlines and Version Control](https://mrkiffie.com/2016/newlines-and-version-control/)

arvidn: heh, it also used to be undefined behavior, pre C++11 :)

arvidn:
does this actually make a difference in a profiler?
you're not profiling debug builds, are you?

include/libtorrent/torrent_peer.hpp
aldenml: I only remove _extra_ newlines at the end

aldenml:
to be honest, I didn't take a look at the profiler (no even in debug :D ), but it's my impression that under IPv6, this memory copies become non trivial and that call can't be easily optimized by the compiler. Do you have any concern? code style?

include/libtorrent/torrent_peer.hpp
d-komarov: Line 279 is not an `extra` one. If it was followed by another empty(new) line then it could be removed. In other words if the file ends with a "\n\n" or "\r\n\r\n" then it has no extra new lines.

aldenml: I'm not sure I understand  what are you talking about

arvidn: traditionally, every line was required to end with a '\n' character in C and C++

aldenml: for future reference, when doing github web based code review, the web interface give you a nice hint of that missing newline at the end
screen shot 2018-01-27 at 11 02 10 am

PR #2736, opened 2018-01-27
fixing warnings in tests code, part5

aldenml:


test/test_checking.cpp
```diff
@@ -329,4 +333,3 @@ TORRENT_TEST(force_recheck)
 {
 	test_checking(force_recheck);
 }
-
```
arvidn: do you get warnings for these?

aldenml: no, that's me again removing _extra_ '\n's

PR #2738, opened 2018-01-28
make coalesce-buffers work on windows

arvidn:
and systems with pwritev()/preadv()


Chocobo1:
There are build errors:
https://ci.appveyor.com/project/arvidn/libtorrent/build/1.0.8237/job/40umbsyv10ifjtte#L302

arvidn:
right, I tested it on mac, but mac doesn't have ``preadv()`` so it wasn't a good test. Should be fixed now

PR #2745, opened 2018-01-28
fixing warnings in tests code, part6

aldenml:


test/test_buffer.cpp
```diff
@@ -152,7 +152,7 @@ void free_buffer(char* m)
 
 char* allocate_buffer(int size)
 {
-	char* mem = (char*)std::malloc(size);
+	char* mem = static_cast<char*>(std::malloc(std::size_t(size)));
```
arvidn: I wonder if there's a reason not to use ``new char[size]`` here instead, and avoid the casts

test/test_dht.cpp
```diff
@@ -2583,11 +2594,11 @@ TORRENT_TEST(traversal_done)
...
 	// invert the ith most significant byte so that the test nodes are
 	// progressively closer to the target item
-	for (int i = 0; i < num_test_nodes; ++i)
+	for (std::size_t i = 0; i < num_test_nodes; ++i)
 		nodes[i].id[i] = ~nodes[i].id[i];
 
 	// add the first k nodes to the subject's routing table
-	for (int i = 0; i < 8; ++i)
+	for (std::size_t i = 0; i < 8; ++i)
 		t.dht_node.m_table.add_node(nodes[i]);
```
arvidn: could this be a range-for over ``nodes``?

test/test_dht.cpp
```diff
@@ -3305,7 +3316,7 @@ TORRENT_TEST(invalid_error_msg)
 	node.incoming(node.m_sock, m);
 
 	bool found = false;
-	for (int i = 0; i < int(observer.m_log.size()); ++i)
+	for (std::size_t i = 0; i < observer.m_log.size(); ++i)
```
arvidn: this looks like it could be a range-for loop over ``observer.m_log``.

aldenml:
sure, I will use the for-loop and `new char[size]`

test/test_buffer.cpp
arvidn: if the ``chained_buffer`` frees with ``free``, this would still have to be ``malloc``

aldenml: OK

aldenml: yes, `std::free` is used with the `free_buffer` above, no change here

test/test_dht.cpp
aldenml: can't modify this one because nodes is of `9` elements, but using it in others

test/test_dht.cpp
aldenml: done

PR #2758, opened 2018-01-31
fixing warnings in tests code, part9

aldenml:


test/test_buffer.cpp
```diff
@@ -188,14 +188,16 @@ struct holder
...
+	// -Wunused-member-function
+	/*holder& operator=(holder&& rhs) noexcept
 	{
 		if (m_buf) free_buffer(m_buf);
 		m_buf = rhs.m_buf;
 		m_size = rhs.m_size;
 		rhs.m_buf = nullptr;
 		return *this;
-	}
+	}*/
```
arvidn: hm. this doesn't look right. all the calls to ``append_buffer()`` below, that construct a temporary ``holder``, are  they must be moved in, right?

test/test_ssl.cpp
```diff
@@ -459,7 +460,7 @@ bool try_connect(lt::session& ses1, int port
 	{
 		std::string name = aux::to_hex(t->info_hash());
 		std::printf("SNI: %s\n", name.c_str());
-		SSL_set_tlsext_host_name(ssl_sock.native_handle(), name.c_str());
+		aux::openssl_set_tlsext_hostname(ssl_sock.native_handle(), name.c_str());
```
arvidn: was there a warning here?

test/test_buffer.cpp
arvidn: oh, right. that's move constructed, not move assigned

aldenml: do you think it's a good idea to put an static assert with `std::is_move_constructible` for `Holder` in `chained_buffer` methods?

test/test_ssl.cpp
aldenml: yes, there is an old style cast there:
```cpp
# define SSL_set_tlsext_host_name(s,name) \
SSL_ctrl(s,SSL_CTRL_SET_TLSEXT_HOSTNAME,TLSEXT_NAMETYPE_host_name,(char *)name)
```

test/test_buffer.cpp
arvidn: well, I don't think it makes sense to test the test class just for the sake of it. Maybe ``chained_buffer`` would never invoke the move assignment of holders (in which case this should probably be ``= delete``). Or maybe there should be a test that attempts to move assign a ``chained_buffer`` to ensure it moves the holder correctly. Either way, those things seems like out of the scope of this patch. I'm just worried about maybe forgetting following up on this by just commenting it out

aldenml: Well, I would like to put `= delete`, but I need to take a look in detail, since there is a `#if TORRENT_CPP98_DEQUE` in `chained_buffer`, I want to know if that is still necessary.

aldenml: do you want me to move out this from the PR?

arvidn: if it works with ``= delete``, I would prefer that

PR #2761, opened 2018-01-31
avoid endianness test in peer_priority function

aldenml:
for you review, I still wonder if I need to call `aux::host_to_network(` on the ports


arvidn:
this looks good. I didn't realize the existing code did all those endianness conditionals. New code is a lot cleaner I think. ``write_uint16()`` writes network byte order already.

PR #2763, opened 2018-02-01
Improve SSL performance by avoiding excessive memory allocations

mkatsevVR:
Boost.Asio moves the io_op object multiple times per async SSL read/write operation: https://github.com/boostorg/asio/blob/develop/include/boost/asio/ssl/detail/io.hpp#L187
Move constuctor makes a **copy** of the internal op_ object: https://github.com/boostorg/asio/blob/develop/include/boost/asio/ssl/detail/io.hpp#L118
That in turn involves copying the container with memory buffers: https://github.com/boostorg/asio/blob/develop/include/boost/asio/ssl/detail/write_op.hpp#L59
With the container being an std::list, this causes a very large number of memory allocations for each write operation.
Replacing the container with a vector reduces the overhead by a factor of O(iovec.size()).
In my application this improved upload speed for SSL-enabled torrents by 7x (from 50MB/s to 350MB/s).

### DIFF
--- a/include/libtorrent/chained_buffer.hpp
+++ b/include/libtorrent/chained_buffer.hpp
@@ -91,7 +91,7 @@ namespace libtorrent
 		// enough room, returns 0
 		char* allocate_appendix(int s);
 
-		std::list<asio::const_buffer> const& build_iovec(int to_send);
+		std::vector<asio::const_buffer> const& build_iovec(int to_send);
 
 		~chained_buffer();
 
@@ -103,7 +103,7 @@ namespace libtorrent
 
 		// this is the vector of buffers used when
 		// invoking the async write call
-		std::list<asio::const_buffer> m_tmp_vec;
+		std::vector<asio::const_buffer> m_tmp_vec;
 
 		// this is the number of bytes in the send buf.
 		// this will always be equal to the sum of the
--- a/src/chained_buffer.cpp
+++ b/src/chained_buffer.cpp
@@ -115,7 +115,7 @@ namespace libtorrent
 		return insert;
 	}
 
-	std::list<asio::const_buffer> const& chained_buffer::build_iovec(int to_send)
+	std::vector<asio::const_buffer> const& chained_buffer::build_iovec(int to_send)
 	{
 		m_tmp_vec.clear();
 
--- a/src/peer_connection.cpp
+++ b/src/peer_connection.cpp
@@ -4987,7 +4987,7 @@ namespace libtorrent
 #ifdef TORRENT_VERBOSE_LOGGING
 		peer_log(">>> ASYNC_WRITE [ bytes: %d ]", amount_to_send);
 #endif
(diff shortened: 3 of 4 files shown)


aldenml:
is there any reason you are still using this old version `RC_1_0` (just asking)

mkatsevVR:
Upgrading is non-trivial due to many API and internal behavior changes in 1.0, haven't had time to do the work.

mkatsevVR:
FWIW, I totally understand if you don't care about it anymore and don't want patches for an old version. This is just something that was a major improvement for me and thought worth kicking upstream.

mkatsevVR:
I don't think the test failures are from my change, but who knows... I did build locally (macOS) and run the test_buffer test successfully.

aldenml:
those tests in `RC_1_0` are very problematic, I don't remember the last time I saw them fully green. You will need to way for arvidn for the acceptance criteria.

arvidn:
yeah, the tests were quite flaky back in 1.0.x (they still are in 1.1 to some extent). I seem to recall the reason for this to be a ``list`` was to allow appending to it while the list is used in an asynchronous write call. With this being a vector, the "iovec" could potentially be freed under the feed of ``async_write_some()``. Now, is that a problem? It's unclear. Chances are that the write operation will copy the iovec itself, and that it's really just the buffers that matter. iirc, the ambiguity in the documentation led me to go with the conservative approach.

The performance gains you cite are quite encouraging though, and maybe it's not worth to the this conservative.

This would affect ``RC_1_1`` as well, but I would expect in master (requiring C++11 and having movable handler object), it may not make a difference.

@mkatsevVR Do you have any insight into the safety of possibly deallocating the vector passed into asio?

I suppose the underlying OS primitive (i.e. ``sendmsg``) doesn't take asio buffers anyway, so it's probably a safe assumption the whole iovec is copied.

aldenml:
it is `std::vector` already in both `RC_1_1` and `master`

mkatsevVR:
When you say "deallocating", are you talking about next call to build_iovec clearing the vector and freeing the memory while the previous async operation is still using it? Why does build_iovec reuse the same member variable instead of simply returning a new object each time? This would avoid any issues like this, no?

Boost doc says "Although the buffers object may be copied as necessary, ownership of the underlying memory blocks is retained by the caller, which must guarantee that they remain valid until the handler is called." which suggests that it the container ownership is not an issue. 
At the very least in the SSL path the container is definitely copied many times (both by looking at the code I've linked in the 1st message and the profiling data that led me to this patch).

arvidn:
I get the impression that this is essentially a "back-port" then, if this is already the case in RC_1_1 and later

arvidn:
@mkatsevVR what I was referring to was using ``std::vector<>::push_back()`` causing the underlying array to be reallocated. i.e. the previous one freed. But it seems quite clear this is not a problem.

PR #2764, opened 2018-02-02
fixing warnings in tests code, part10, final

aldenml:


test/test_dht.cpp
```diff
@@ -642,10 +642,11 @@ dht::key_desc_t const sample_infohashes_desc[] = {
 
 void print_state(std::ostream& os, routing_table const& table)
 {
+#define BUFFER_CURSOR_POS &buf[std::size_t(cursor)], buf.size() - std::size_t(cursor)
```
arvidn: hm. not super excited about this. I think a more reasonable approach would be to use ``std::stringstream`` instead. That's a bit off-topic here though. The warnings you got here, is it more than ``cursor`` being ``int`` (instead of ``std::size_t``)?

aldenml: yes, I agree with you on the use of the macro, and the warnings is about it, `cursor` not being `std::size_t`. Do you want another solution? (I just wanted to keep the diff to the minimum)

arvidn: would it work to make the ``cursor`` ``std::size_t`` instead of the macro?

arvidn: I mean, to fix the warnings

aldenml: no, because `std::snprintf` returns `int`, and then I will need to wrap all the calls with a cast, like
```cpp
cursor += std::size_t(std::snprintf(&buf[cursor], buf.size() - cursor....
```
do you prefer that?

arvidn: I see, let's go with the macro then

PR #2767, opened 2018-02-03
added include files to CMakeLists.txt and added minimal support for cmake build in CI

aldenml:


CMakeLists.txt
```diff
@@ -4,8 +4,278 @@ project(libtorrent)
 set (SOVERSION "10")
 set (VERSION "1.2.0")
 
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
```
ssiloti: I think `target_compile_features(torrent-rasterbar PUBLIC cxx_std_11)` is the current preferred way of specifying that C++11 is required.

CMakeLists.txt
```diff
@@ -212,6 +482,31 @@ if (dht)
...
+		-Wall
+		-Wextra
+		-Wpedantic
+		-Wparentheses
+		-Wvla
+		-Wc++11-compat
+		-Wno-format-zero-length
+		-ftemplate-depth=512)
+endif()
+
```
ssiloti: These should probably use `target_compile_options`.

CMakeLists.txt
aldenml: that's nice, didn't know about it, and it propagates the requirement

CMakeLists.txt
ssiloti: Looking at this again I realize that these options shouldn't be propagated to dependents, but we still want them to apply to examples/tests/etc. So you had it right the first time, `add_compile_options` was correct here.

aldenml: right, but before reverting, what do you think of `-ftemplate-depth=512`? (to propagate, or not)

ssiloti: I don't think it should. We're setting it lower than the default specified by C++11 (1024) to catch excessive template recursion and that shouldn't be imposed on dependents.

arvidn:
next challenge: build with cmake on appveyor :)

PR #2768, opened 2018-02-03
defer truncating existing files until the first time we write to them

arvidn:


src/storage.cpp
```diff
@@ -605,20 +605,16 @@ namespace libtorrent
 					}
 				}
 				ec.ec.clear();
+
+				// just creating the file is enough to make it zero-sized. If
+				// there's a race here an some other process truncates the file,
+				// it's not a problem, we won't access emty files ever again
```
arvidn: some typos here

arvidn:
I should really add a test for this

PR #2788, opened 2018-02-12
more user_load_torrent cleanup in internal API

aldenml:
My reasoning here is that, since this action is effectively nil, the API deprecated feature is only important at the public API boundary.

### DIFF
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -47,7 +47,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/ssl_stream.hpp"
 #endif
 
-#include "libtorrent/session.hpp" // for user_load_function_t
+#include "libtorrent/session.hpp"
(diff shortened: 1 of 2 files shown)


include/libtorrent/aux_/session_impl.hpp
```diff
@@ -352,10 +352,6 @@ namespace aux {
 #ifndef TORRENT_NO_DEPRECATE
 			//deprecated in 1.2
 
-			TORRENT_DEPRECATED
-			void set_load_function(user_load_function_t fun)
-			{ m_user_load_torrent = fun; }
```
arvidn: the reason I left this here is that the client may store handles in the function object, and rely on them being held alive. Perhaps this is a bit far fetched, but it could break some code.

arvidn: handles, like a ``shared_ptr<>`` for instance

aldenml: now I remember, you mentioned this before

PR #2793, opened 2018-02-15
minor lint oriented cleanup and refactor in headers

aldenml:


include/libtorrent/packet_pool.hpp
```diff
@@ -168,7 +168,7 @@ namespace libtorrent {
 			, m_mtu_floor_slab(mtu_floor_size)
 			, m_mtu_ceiling_slab(mtu_ceiling_size)
 		{}
-		packet_pool(packet_pool&&) = default;
+		packet_pool(packet_pool&&) = delete;
```
arvidn: I don't understand this change

arvidn: is there a non-movable member in here?

arvidn: I see, I would prefer making ``packet_slab`` movable instead, by just defaulting its move constructor.

aldenml: yes, the packet_slab members, they have `packet_slab(const packet_slab&) = delete;`

aldenml: ok, I will change that

PR #2795, opened 2018-02-17
more code cleanup, typos, minor refactor in headers

aldenml:


include/libtorrent/piece_picker.hpp
```diff
@@ -450,9 +450,9 @@ namespace libtorrent {
 			, typed_bitfield<piece_index_t> const& have
 			, picker_options_t options) const;
 
-		// only defined when TORRENT_PICKER_LOG is defined, used for debugging
-		// unit tests
+#ifdef TORRENT_PICKER_LOG
```
arvidn: this define is not really a project wide configuration option. It's just defined in piece_picker.cpp when debugging. So I don't think this will make any real difference

aldenml: true, I will revert that

PR #2796, opened 2018-02-17
fix tie-break in duplicate peer connection disconnect logic

arvidn:


src/peer_list.cpp
```diff
@@ -671,57 +671,39 @@ namespace libtorrent
...
+						|| ((our_port == other_port) && randint(2));
 
-					if (our_port < other_port)
-					{
 #ifndef TORRENT_DISABLE_LOGGING
-						c.peer_log(peer_log_alert::info, "DUPLICATE_PEER_RESOLUTION"
-							, "\"%d\" < \"%d\"", our_port, other_port);
-						i->connection->peer_log(peer_log_alert::info, "DUPLICATE_PEER_RESOLUTION"
-							, "\"%d\" < \"%d\"", our_port, other_port);
+					c.peer_log(peer_log_alert::info, "DUPLICATE_PEER_RESOLUTION"
```
mkatsevVR: probably also want to include disconnect1 in the log message

arvidn:
@mkatsevVR does this fix the problem for you?

PR #2798, opened 2018-02-18
undo the patch to restore incoming TCP connections over SOCKS5

arvidn:


aldenml:
any comment for this revert?

arvidn:
yeah, over at the tribler project they said their download through tunnels has broken, and that they were making libtorrent connect to their tunnel via a SOCKS5 interface. I got the impression that they relied on this feature, of extending ``BIND`` to support incoming TCP connections (more generic than old school FTP connections). After a while it turned out they only use uTP connections, and only rely on the ``UDP ASSOCIATE`` socks5 command. Which is well defined and worked all along (or at least supposed to work).

I prefer to not have all this code here, since I'm not aware of a single socks5 proxy that actually implements this more generic interpretation of ``BIND``.

arvidn:
Here's the reference: https://github.com/Tribler/tribler/issues/3353

aldenml:
thanks for the explanation!

PR #2799, opened 2018-02-18
added friendly_name and description to ip_interface

aldenml:
The context of this change is:

1-) It's very useful in Windows to have this information to inspect while experimenting/debugging, because the device name is an ugly UUID.
2-) This `enum_net` code is very robust, in terms of multiplatform and security, compared to everything I see in other projects. Just adding this minimal change, allows me to use it in a particular way (to VPN detection heuristic) without repeating code.

It's possible to get a similar information in macOS with a combined use of `SCNetworkInterfaceGetLocalizedDisplayName` and `SCNetworkInterfaceGetBSDName`. I'm not changing any public API, and I'm fully aware of the risks if I insist in using it. Also, I don't discard that in the future this can be part of the public API.

### DIFF
--- a/include/libtorrent/enum_net.hpp
+++ b/include/libtorrent/enum_net.hpp
@@ -59,6 +59,8 @@ namespace libtorrent {
 		address interface_address;
 		address netmask;
 		char name[64];
+		char friendly_name[128];
+		char description[128];
 		// an interface is preferred if its address is
 		// not tentative/duplicate/deprecated
 		bool preferred = true;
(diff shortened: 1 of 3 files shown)


arvidn:
at times I've been quite interested in cleaning up enum_net (specifically to give it an iterator interface), pull it out as separate library and perhaps propose it as an addition to boost.asio.

I think it would make sense to extend the ``test/enum_if.cpp`` command line tool as well, to print these new fields.

aldenml:
ok, I will add that tool

PR #2800, opened 2018-02-19
CMake script improvements

d-komarov:


d-komarov:
Also fix numerous test failures on Windows platforms

aldenml:
I think that you should at least separate the code changes from the cmake ones in two PRs.

test/setup_transfer.cpp
```diff
@@ -252,36 +253,38 @@ void save_file(char const* filename, char const* data, int size)
 		fprintf(stderr, "ERROR writing file '%s': %s\n", filename, ec.message().c_str());
 		return;
 	}
-
 }
 
 bool print_alerts(lt::session& ses, char const* name
 	, bool allow_disconnects, bool allow_no_torrents, bool allow_failed_fastresume
-	, boost::function<bool(libtorrent::alert const*)> predicate, bool no_output)
+	, boost::function<bool(libtorrent::alert const*)> const& predicate, bool no_output)
```
arvidn: this will prevent ``predicate`` from being moved into this function. was that intentional?

test/test_upnp.cpp
```diff
@@ -176,12 +176,12 @@ void run_upnp_test(char const* root_filename, char const* router_model, char con
...
 			break;
 		}
-		if (upnp_handler->router_model() != "") break;
+		if (!upnp_handler->router_model().empty()) break;
 		test_sleep(100);
 	}
 
 	std::cerr << "router: " << upnp_handler->router_model() << std::endl;
-	TEST_EQUAL(upnp_handler->router_model(), router_model);
+	//TEST_EQUAL(upnp_handler->router_model(), router_model);
```
arvidn: the idea is that we should pick up the fake UPnP device the test spins up. I take it you find this unreliable, is that right?

test/test_tracker.cpp
```diff
@@ -316,8 +316,11 @@ TORRENT_TEST(extract_peer_missing_port)
 bool connect_alert(libtorrent::alert const* a, tcp::endpoint& ep)
 {
 	if (peer_connect_alert const* pc = alert_cast<peer_connect_alert>(a))
+	{
 		ep = pc->ip;
-	return true;
```
arvidn: ouch

test/setup_transfer.cpp
d-komarov: Yes, I intended it. C++98 doesn't have move semantics.

arvidn:
These are really good fixes. I'm not an expert on CMake though and would defer judging these changes to @ssiloti and @aldenml .

CMakeLists.txt
```diff
@@ -286,6 +286,10 @@ if (NOT pool-allocators)
 	add_definitions(-DTORRENT_DISABLE_POOL_ALLOCATOR)
 endif()
 
+if (NOT resolve-countries)
+	add_definitions(-DTORRENT_DISABLE_RESOLVE_COUNTRIES)
```
arvidn: This define is not in use anymore. I think your first patch to remove it was correct

test/CmakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+file(GLOB PYTHON_ASSETS "${CMAKE_CURRENT_SOURCE_DIR}/*.py")
+file(COPY ${PYTHON_ASSETS} DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
+
+file(GLOB XML_ASSETS "${CMAKE_CURRENT_SOURCE_DIR}/*.xml")
+file(COPY ${XML_ASSETS} DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
+
+file(COPY "utf8_test.txt" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
+file(COPY "mutable_test_torrents" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
+file(COPY "test_torrents" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
+file(COPY "ssl" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
```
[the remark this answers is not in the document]
d-komarov: I don't like the idea to run tests with ```WORKING_DIRECTORY``` that point to the source tree, at least not in someones development environment. The reason for that is very straightforward, interrupted or failed tests may leave junk files/directories after them. After all I don't want to mess source tree with test output.
I agree that copying test assets might have some drawbacks, but at the moment it is the lesser of two evils.

test/test_upnp.cpp
d-komarov: Exactly. The problem here is that ```lt::upnp``` constructor ignores interface passed to it and thus upnp instance will update ```m_model``` each time it receives upnp xml. Since it is neither possible to predict order in which routers are discovered nor to select router by the interface it makes no sense to expect ```upnp_handler->router_model()``` to return fake device name.
I beleive ```TEST_CHECK(!upnp_handler->router_model().empty());``` might be used instead. What do you think?

test/test_tracker.cpp
d-komarov: Something wrong with it?

arvidn: no, I was just commenting on me writing this bug :P (that you just fixed)

test/test_upnp.cpp
arvidn: that seems reasonable, yes. (at least until the upnp implementation can stick to a specific interface)

test/CmakeLists.txt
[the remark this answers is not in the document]
arvidn: does CMake have (platform independent) functions for creating symlinks?

CMakeLists.txt
d-komarov: But it is still better to remove this along with source code cleanup, right?

This could be in a separate PR as finding all ```TORRENT_DISABLE_RESOLVE_COUNTRIES``` references gives:
>Matching lines: 18    Matching files: 10

arvidn: oh, sorry. Yes. I must have git grepped in the wrong directory or something.

arvidn:
thanks!

PR #2804, opened 2018-02-20
Add file test workaround

d-komarov:
File test compiled with TORRENT_USE_UNC_PATHS produces file that cannot
be removed.

### DIFF
--- a/test/test_file.cpp
+++ b/test/test_file.cpp
@@ -393,8 +393,10 @@ TORRENT_TEST(unc_paths)
 {
 	std::string const reserved_name = "con";
 	error_code ec;
-	file f;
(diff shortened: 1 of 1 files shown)


d-komarov:
I'm afraid there are more cases that gonna fail with UNC file paths.

arvidn:
This still creates a file where th a reserved name. If the original test failed because of the file name, the new test would fail too, no?

d-komarov:
Yes, file with reserved name is created flawlessly. The file test actually gets passed because it doesn't care of cleanup that would never succeed because `lt::remove` can't handle reserved file names.

arvidn:
ah, I see. I guess that's a separate issue (which I tried to address a while ago here: https://github.com/arvidn/libtorrent/pull/2121 ) but I got stuck at a few of those errors.

PR #2805, opened 2018-02-20
add modified docs with more detailed explanation on how to create the…

David-Luis:
… bindings for windows

remove std:: from snprintf for windows compiler compatibility


include/libtorrent/session.hpp
```diff
@@ -308,7 +308,7 @@ namespace aux {
 			char if_string[100];
 
 			if (listen_interface == nullptr) listen_interface = "0.0.0.0";
-			std::snprintf(if_string, sizeof(if_string), "%s:%d", listen_interface, listen_port_range.first);
```
aldenml: why remove `std::`?

David-Luis: baaically this doesn't compile on windows (msvc 14.0)

d-komarov: I believe that it's not about msvc 14.0, but because of these lines:
```
#ifndef TORRENT_NO_DEPRECATE
#include "libtorrent/fingerprint.hpp"
#include <cstdio> // for snprintf
#endif
```
Instead of removing `std::`, try moving `#include <cstdio>` out of `#ifndef` block.

docs/python_binding.rst
```diff
@@ -14,37 +14,39 @@ building
...
 
-There is a ``setup.py`` shipped with libtorrent that can be used on windows.
-On windows the setup.py will invoke ``bjam`` and assume that you have boost
-sources at ``$BOOST_PATH``. The resulting executable is self-contained, it does
-not depend any boost or libtorrent dlls.
+Download and install `Visual C++ 2015 Build Tools <http://landinghub.visualstudio.com/visual-cpp-build-tools/>`_
 
-On other systems, the setup.py is generated by running
-``./configure --enable-python-binding``.
+Download `Boost libraries <http://www.boost.org/users/history/>`_ Extract it to c:\\Libraries\\boost_1_63_0 and create these environmental vars:
```
arvidn: are the double backslashes required by restructured text?

arvidn:
thanks for helping improve the docs!

include/libtorrent/session.hpp
David-Luis: I tried moving it out of the ifndef with no luck :(

docs/python_binding.rst
```diff
@@ -14,37 +14,39 @@ building
...
+	
+building using boost build (others)
+-----------------------------------
 To set up your build environment, you need to add some settings to your
 ``$BOOST_BUILD_PATH/user-config.jam``.
 
 Make sure your user config contains the following line::
 
-	using python : 2.3 ;
+	#using python : 2.3 ;
```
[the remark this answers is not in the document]
David-Luis: yes, but when you open the file for the first time this line is commented. I explain in the next lines that you should change this to your python version

docs/python_binding.rst
```diff
@@ -14,37 +14,39 @@ building
...
-	python setup.py install
-
-
-building using boost build
---------------------------
+Navigate to binding\\torrents and execute::
+	python setup.py build --bjam
+	
+Note: If you are using 64bits python you should edit setup.py and add this to the b2 command:
+``$address-model=64``
```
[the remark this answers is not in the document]
David-Luis: yes!

docs/python_binding.rst
```diff
@@ -14,37 +14,39 @@ building
...
-On other systems, the setup.py is generated by running
-``./configure --enable-python-binding``.
+Download `Boost libraries <http://www.boost.org/users/history/>`_ Extract it to c:\\Libraries\\boost_1_63_0 and create these environmental vars:
 
-To build the Python bindings do:
+1. BOOST_BUILD_PATH: "c:\\Libraries\\boost_1_63_0\\tools\\build\\"
+2. BOOST_ROOT: "c:\\Libraries\\boost_1_63_0\\"
 
-1. Run::
+Navigate to BOOST_ROOT, execute "bootstrap.bat" and add to the path "c:\\Libraries\\boost_1_63_0\\tools\\build\\src\\engine\\bin.ntx86\\"
```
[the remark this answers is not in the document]
David-Luis: I'm following this tutorial but it can be wrong https://theboostcpplibraries.com/introduction-installation
"Enter bootstrap on Windows and ./bootstrap.sh on other platforms, such as Linux, to compile bjam"

docs/python_binding.rst
David-Luis: nope, I changed all of them to forward slashes

docs/python_binding.rst
```diff
@@ -14,37 +14,39 @@ building
...
-building using setup.py
------------------------
+building using boost build (windows)
+------------------------------------
 
-There is a ``setup.py`` shipped with libtorrent that can be used on windows.
-On windows the setup.py will invoke ``bjam`` and assume that you have boost
-sources at ``$BOOST_PATH``. The resulting executable is self-contained, it does
-not depend any boost or libtorrent dlls.
+Download and install `Visual C++ 2015 Build Tools <http://landinghub.visualstudio.com/visual-cpp-build-tools/>`_
```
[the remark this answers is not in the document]
David-Luis: ok!

docs/python_binding.rst
```diff
@@ -14,37 +14,39 @@ building
...
+1. BOOST_BUILD_PATH: "c:\\Libraries\\boost_1_63_0\\tools\\build\\"
+2. BOOST_ROOT: "c:\\Libraries\\boost_1_63_0\\"
 
-1. Run::
+Navigate to BOOST_ROOT, execute "bootstrap.bat" and add to the path "c:\\Libraries\\boost_1_63_0\\tools\\build\\src\\engine\\bin.ntx86\\"
+	
+Move the file ``$user-config.jam`` from ``$BOOST_BUILD_PATH\example\`` to ``$BOOST_BUILD_PAT\\user-config.jam`` and add this at the end:
+1. using msvc : 14.0 ;
+2. using python : 3.5 : C:/Users/<UserName>/AppData/Local/Programs/Python/Python35 : C:/Users/<UserName>/AppData/Local/Programs/Python/Python35/include : C:/Users/<UserName>/AppData/Local/Programs/Python/Python35/libs ;
+(change the python path for yours)
```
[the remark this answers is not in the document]
David-Luis: changed

docs/python_binding.rst
```diff
@@ -17,32 +17,40 @@ which is a python module that can be imported in a python program.
...
+
+1. BOOST_BUILD_PATH: "c:/Libraries/boost_1_63_0/tools/build/"
+2. BOOST_ROOT: "c:/Libraries/boost_1_63_0/"
+
+Navigate to BOOST_ROOT, execute "bootstrap.bat" and add to the path "c:/Libraries/boost_1_63_0/tools/build/src/engine/bin.ntx86/"
 	
-Move the file ``$user-config.jam`` from ``$BOOST_BUILD_PATH\example\`` to ``$BOOST_BUILD_PAT\\user-config.jam`` and add this at the end:
-1. using msvc : 14.0 ;
-2. using python : 3.5 : C:/Users/<UserName>/AppData/Local/Programs/Python/Python35 : C:/Users/<UserName>/AppData/Local/Programs/Python/Python35/include : C:/Users/<UserName>/AppData/Local/Programs/Python/Python35/libs ;
+Move the file ``$user-config.jam`` from ``BOOST_BUILD_PATH/example/`` to ``BOOST_BUILD_PAT/user-config.jam`` and add this at the end:
```
arvidn: there's no "$" in ``user-config.jam`` either

arvidn: since this is the windows section, perhaps you should put "%" around ``BOOST_BUILD_PATH`` and ``BOOST_ROOT``. Also, there's a typo here, you're missing an "H" at the end of ``BOOST_BUILD_PATH``

docs/python_binding.rst
```diff
@@ -17,32 +17,40 @@ which is a python module that can be imported in a python program.
...
+``address-model=64``
 
-This will create the file libtorrent.pyd inside build\\lib\\ that contains the binding.
+This will create the file libtorrent.pyd inside build/lib/ that contains the binding.
 	
 building using boost build (others)
 -----------------------------------
 To set up your build environment, you need to add some settings to your
-``$BOOST_BUILD_PATH/user-config.jam``.
+``BOOST_BUILD_PATH/user-config.jam``.
```
arvidn: in this case I think it would make sense to keep the "$", to make it clear that this is referring to the environment variable

docs/python_binding.rst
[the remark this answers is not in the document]
arvidn: I think it's risky to document what the default ``user-config.jam`` file will look like, since it's likely to change in a future version of boost. It may also be somewhat misleading, for a reader skimming, it may not be obvious that what you're showing is what *not* to have in the file. I think it would be better to just say what *should* be in the file.

include/libtorrent/session.hpp
arvidn: are you sure you're building in C++11 mode? ``std::snprintf()`` was introduced with C++11.

David-Luis: wow you are right, chaning for using msvc : 14.0 : : /std:c++11 fix the issue, I will update the docs accordly :)

PR #2810, opened 2018-02-23
Fix bandwidth allocation

d-komarov:
Improve accuracy of expected transfer quota calculation by using
floating point ratio when possible.

Don't allow web peer connection to ask quota from bandwidth manager
if send buffer is empty.

* Use __debugbreak in TORRENT_ASSERT if compiled with msvc


src/peer_connection.cpp
```diff
@@ -5460,23 +5460,22 @@ namespace libtorrent
 	int peer_connection::wanted_transfer(int channel)
 	{
 		TORRENT_ASSERT(is_single_thread());
-		shared_ptr<torrent> t = m_torrent.lock();
 
-		const int tick_interval = (std::max)(1, m_settings.get_int(settings_pack::tick_interval));
+#if TORRENT_NO_FPU
+		int const tick_interval = (std::max)(1, m_settings.get_int(settings_pack::tick_interval) / 1000);
```
aldenml: This new expression is ill-conditioned from the numerical point of view and not equivalent to the previous ones. Also, the use of floating point arithmetic is only used in deprecated features in `master`.

src/peer_connection.cpp
```diff
@@ -5460,23 +5460,22 @@ namespace libtorrent
 	int peer_connection::wanted_transfer(int channel)
 	{
 		TORRENT_ASSERT(is_single_thread());
-		shared_ptr<torrent> t = m_torrent.lock();
```
aldenml: are you sure this line is not necessary? (it has the secondary effect of keeping the torrent alive)

src/peer_connection.cpp
d-komarov: Well, then could you please explain me the meaning of both original expressions?

aldenml: other than the missing `* 2`, it all comes down to the fact that precision in integer division works better if the numerator have bigger module.

src/peer_connection.cpp
d-komarov: I was thinking of it, but didn't find any practical reason why torrent object's lifetime must be prolonged during this call.

src/peer_connection.cpp
d-komarov: I wouldn't be so sure. Consider the following input:
upload_rate = 1 000 000 000(1 GB/s)
tick_interval = 4500(4.5s)
Results would be:
Original expression: 9 000 000 000 = 9 GB
New expression: 4 000 000 000 = 4 GB
Precise result : 4.5 GB
In my opinion 4 is a way better approximation to 4.5 than 9 gigs.

aldenml: to be honest, I didn't pay attention to the correctness of the value returned in the method (you can be right about the dropping of the `*2`), only to the fact that you never do this kind of refactor in a numeric sensitive operation, you are definitely losing precision. Also, If I remember correctly, arvidn was moving away from float operations, which is a good idea IMO.

ssiloti: I'm not sure why the rates are multiplied by 2, but I don't see any benefit to using floating point here. If you simply removed the `* 2` from the pure integer version it would yield exactly 4.5GB.

d-komarov: That's exactly what it evaluates to if there is no FPU support.
```
int const tick_interval = (std::max)(1, m_settings.get_int(settings_pack::tick_interval) / 1000);
```
That is meant to be max of 1 second and floor of settings_pack::tick_interval in seconds.

d-komarov: Ouch, it also appears that int(4.5GB) would give ~205MB. I think function's arithmetic should be covered with a test, to check it's accuracy for wide range of hypothetical input data.

arvidn: I think it's risky to use ``int`` operands here. It's actually worse than yielding ~205 MB, it yields undefined behaviour because of signed overflow.
I would prefer to avoid floating point arithmetic if it can be, and it looks like it's not necessary or particularly important in this case.

src/peer_connection.cpp
```diff
@@ -5564,6 +5563,9 @@ namespace libtorrent
 		TORRENT_ASSERT(is_single_thread());
 		if (m_disconnecting) return;
 
+		// guard BEP17 and BEP19 seeds with depleted send buffer
+		if (type() != bittorrent_connection && m_send_buffer.empty()) return;
```
arvidn: could you elaborate a bit on this? If it makes sense to not request bandwidth for a web seed that doesn't have anything in the send buffer, wouldn't it make sense for normal bittorrent connections too?

src/assert.cpp
```diff
@@ -352,10 +352,14 @@ TORRENT_EXPORT void assert_fail(char const* expr, int line
 
 	// if production asserts are defined, don't abort, just print the error
 #ifndef TORRENT_PRODUCTION_ASSERTS
+	#ifdef _MSC_VER
+	__debugbreak();
```
arvidn: +1

src/peer_connection.cpp
```diff
@@ -5460,23 +5460,22 @@ namespace libtorrent
...
+#endif
 
 		if (channel == download_channel)
 		{
-			return std::max((std::max)(m_outstanding_bytes
-				, m_recv_buffer.packet_bytes_remaining()) + 30
-				, int(boost::int64_t(m_statistics.download_rate()) * 2
-					* tick_interval / 1000));
+			return std::max((std::max)(m_outstanding_bytes, m_recv_buffer.packet_bytes_remaining())
+				+ 30, static_cast<int>(m_statistics.download_rate() * tick_interval));
```
arvidn: I think the idea behind the ``* 2`` is to have some slack, to not have run out entirely of quota by the time the next bandwidth request is made (i.e. stall) and also to allow connections that have just started up to increase their download rate, and not be stuck to the download rate they had at startup just because they don't get any more bandwidth assigned to them.

Do you see some improved results in practice removing the ``* 2``?

src/peer_connection.cpp
d-komarov: For regular bt connections it could also make sense if  `m_reading_bytes == 0'. I.e. if bittorrent connection has an empty send buffer and has no pending read operations it should be safe to skip bandwidth request as well.

src/peer_connection.cpp
d-komarov: Well, definitely web seed connections will benefit from this change as they will be sharing available quota better then before. For regular bt connections it is not that obvious.

d-komarov: Also I was thinking of web connections to be restricted to send complete HTTP requests only. Now it is possible that, because of quota limitations, web connection will send partial HTTP request. I have been watching situation when web connection was sending 1 byte requests. This definitely must be avoided, especially when communicating to NGINX. The last one will append access.log with bad request records when it receives 1 byte.

arvidn: I suppose quota could be accrued in a counter here, or the bandwidth manager could get another parameter of the size of a "quanta". That way a bittorrent connection could ask for enough space to send whole messages at a time.

src/peer_connection.cpp
arvidn: I was suggesting that this conditional could be simplified to:

``if (m_send_buffer.empty()) return;``

src/peer_connection.cpp
arvidn: the problem I fear of not using any multiplier here is that a peer with a low upload rate, will only get enough quota for a slow download rate, which means it will request quota for a low download rate. There's a feed-back loop that a peer could get stuck in.

I think there should be some multiplier, if not `2`, perhaps `* 3 / 2` or `* 4 / 3`.

Does that make sense?

src/peer_connection.cpp
```diff
@@ -5460,23 +5460,22 @@ namespace libtorrent
...
 		}
 		else
 		{
-			return std::max((std::max)(m_reading_bytes
-				, m_send_buffer.size())
-				, int((boost::int64_t(m_statistics.upload_rate()) * 2
-					* tick_interval) / 1000));
+			return std::max((std::max)(m_reading_bytes, m_send_buffer.size()),
+				static_cast<int>(m_statistics.upload_rate() * tick_interval));
```
arvidn: before you multiply by `tick_interval` you should widen the upload rate to a `boost::int64_t`. Also, the tick_interval is specified in milliseconds, you still need to divide by `1000` here

arvidn: I just noticed you divide it by 1000 before you assign `tick_interval`. This is problematic since the default interval is < 1 second, assigning it to an integer will give you 0.

d-komarov: Nope, integer is evaluated as:
```int const tick_interval = (std::max)(1, m_settings.get_int(settings_pack::tick_interval) / 1000);```

The case you mentioned a `float` one.

Comment (arvidn) [reply src/peer_connection.cpp]: ok, so the tick_interval can never be less than 1 second. The existing code lets it be less than one second, which I think may be important

Comment (d-komarov) [reply src/peer_connection.cpp]: Ah, I realized what's wrong with it. I will revert all this. But I still wish this function to be much simpler... Maybe the next PR will dedicated to this.

File: src/peer_connection.cpp [hunk 7]
```diff
@@ -5560,7 +5557,8 @@ namespace libtorrent
 	void peer_connection::setup_send()
 	{
 		TORRENT_ASSERT(is_single_thread());
-		if (m_disconnecting) return;
+
+		if (m_disconnecting || m_send_buffer.empty()) return;
```
arvidn: I notice the request bandwidth call is just below here. Is there a risk that if we don't have any quota, and the send buffer is empty, we won't ever request send quota?

d-komarov: There are numerous traces where `m_send_buffer` is being modified, but the only function to consider quota is `peer_connection::setup_send()`. 

The following methods are known to modify send buffer regardless of upload quota:
1. `peer_connection::fill_send_buffer()`
2. `peer_connection::send_buffer()`
3. `peer_connection::append_send_buffer()`
4. `peer_connection::append_const_send_buffer()`

The only case I can't justify is related to send barrier.

arvidn:
I still don't understand why you think `bindings/python/setup.py` needs to be checked in. You say python likes to build packages using distutils, but the cmake file generates this file, to (I assume) be able to build or package the python module using distutils. Why does it need to exist before running cmake?

If the `setup.py` generated by cmake is broken somehow, why generate it at all?
If it works just fine, then it seems redundant to have a slightly different version as well.

But perhaps more importantly, if there is a genuine use-case to have both, they should probably have different names. It's quite inconvenient to have the builds script mutate a file that's under version control.

PR #2815, opened 2018-03-01
utp retransmit and window size fix

ichorid:
This patch is the result of [work done by Tribler team](https://github.com/Tribler/tribler/issues/2620). It solves two problems:

1. uTP stream implementation has a bug that incorrectly counts `m_bytes_in_flight` twice when checking it against`m_adv_wnd`:
```
if (m_bytes_in_flight + payload_size > (std::min)(int(m_cwnd >> 16)
	, int(m_adv_wnd - m_bytes_in_flight)))
```
This results in _advertised window size_ effectively being cut in half, significantly reducing uTP performance on single-peer low-latency links. The patch fixes that problem. [Performance tests](https://github.com/Tribler/tribler/issues/2620#issuecomment-367688634) show immediate increase in uTP bandwidth usage.

2. uTP implementation parameters  `m_in_buf_size` and `max_packets_reorder` are too low for high-latency links. These set the limits of growth for `std::vector<packet*> m_receive_buffer`. 
As `m_receive_buffer` does not allocate any space by default, and is not affected in any way by `recv_socket_buffer_size` and `send_socket_buffer_size`, it is safe to increase `m_in_buf_size` and `max_packets_reorder` to better suit the high-latency, high-bandwidth links (such as satellite links and TOR-like tunnels). The [tests](https://github.com/Tribler/tribler/issues/2620#issuecomment-368549367) show immediate improvement of uTP performance on these kind of links.

### DIFF
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -1788,16 +1788,6 @@ bool utp_socket_impl::send_pkt(int const flags)
 			m_fast_resend_seq_nr = (m_fast_resend_seq_nr + 1) & ACK_MASK;
 	}
 
-	int sack = 0;
-	if (m_inbuf.size())
-	{
-		// the SACK bitfield should ideally fit all
-		// the pieces we have successfully received
-		sack = (m_inbuf.span() + 7) / 8;
-		if (sack > 32) sack = 32;
-	}
-
-	boost::uint32_t const close_reason = m_close_reason;
 
 	// MTU DISCOVERY
 
@@ -1811,13 +1801,30 @@ bool utp_socket_impl::send_pkt(int const flags)
 		&& m_write_buffer_size >= m_mtu_floor * 3
 		&& m_seq_nr != 0
 		&& (m_cwnd >> 16) > m_mtu_floor * 3);
+	// for non MTU-probes, use the conservative packet size
+	int const effective_mtu = mtu_probe ? m_mtu : m_mtu_floor;
+
+	boost::uint32_t const close_reason = m_close_reason;
+
+
+	int sack = 0;
+	if (m_inbuf.size())
+	{
+        const int max_sack_size = effective_mtu 
+            - sizeof(utp_header)
+            - 2 // for sack padding/header
+            - (close_reason ? 6 : 0);
+
+		// the SACK bitfield should ideally fit all
+		// the pieces we have successfully received
+		sack = (m_inbuf.span() + 7) / 8;
+		if (sack > max_sack_size) sack = max_sack_size;
+	}
 
 	int const header_size = sizeof(utp_header)
 		+ (sack ? sack + 2 : 0)
 		+ (close_reason ? 6 : 0);
 
-	// for non MTU-probes, use the conservative packet size
-	int const effective_mtu = mtu_probe ? m_mtu : m_mtu_floor;
 	int payload_size = (std::min)(m_write_buffer_size
 		, effective_mtu - header_size);
 	TORRENT_ASSERT(payload_size >= 0);
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -136,8 +136,6 @@ enum
 {
 	ACK_MASK = 0xffff,
 
-	// the number of packets that'll fit in the reorder buffer
-	max_packets_reorder = 4096,
 
 	// if a packet receives more than this number of
 	// duplicate acks, we'll trigger a fast re-send
@@ -2850,6 +2848,10 @@ bool utp_socket_impl::incoming_packet(boost::uint8_t const* buf, int size
 
 	if (ph->get_type() == ST_DATA)
 		m_sm->inc_stats_counter(counters::utp_payload_pkts_in);
+
+	// the number of packets that'll fit in the reorder buffer
+    // incoming MTU of 1100 and minimal size of 16 packets are chosen arbitrarly
+	const boost::uint32_t max_packets_reorder = std::max(16, m_in_buf_size / 1100);
 
 	if (m_state != UTP_STATE_NONE
 		&& m_state != UTP_STATE_SYN_SENT
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -1827,8 +1827,7 @@ bool utp_socket_impl::send_pkt(int const flags)
 	// if we have one MSS worth of data, make sure it fits in our
 	// congestion window and the advertised receive window from
 	// the other end.
-	if (m_bytes_in_flight + payload_size > (std::min)(int(m_cwnd >> 16)
-		, int(m_adv_wnd - m_bytes_in_flight)))
+	if (m_bytes_in_flight + payload_size > (std::min)(int(m_cwnd >> 16) , int(m_adv_wnd)))
 	{
 		// this means there's not enough room in the send window for
 		// another packet. We have to hold off sending this data.
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -137,7 +137,7 @@ enum
 	ACK_MASK = 0xffff,
 
 	// the number of packets that'll fit in the reorder buffer
-	max_packets_reorder = 512,
+	max_packets_reorder = 4096,
 
 	// if a packet receives more than this number of
 	// duplicate acks, we'll trigger a fast re-send
@@ -281,7 +281,7 @@ struct utp_socket_impl
 		, m_written(0)
 		, m_receive_buffer_size(0)
 		, m_read_buffer_size(0)
-		, m_in_buf_size(1024 * 1024)
+		, m_in_buf_size(8 * 1024 * 1024)
 		, m_in_packets(0)
 		, m_out_packets(0)
 		, m_send_delay(0)
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -136,7 +136,6 @@ enum
 {
 	ACK_MASK = 0xffff,
 
-
 	// if a packet receives more than this number of
 	// duplicate acks, we'll trigger a fast re-send
 	dup_ack_limit = 3,
@@ -279,7 +278,7 @@ struct utp_socket_impl
 		, m_written(0)
 		, m_receive_buffer_size(0)
 		, m_read_buffer_size(0)
-		, m_in_buf_size(8 * 1024 * 1024)
+		, m_receive_buffer_capacity(1024 * 1024)
 		, m_in_packets(0)
 		, m_out_packets(0)
 		, m_send_delay(0)
@@ -528,7 +527,7 @@ public:
 	boost::int32_t m_read_buffer_size;
 
 	// max number of bytes to allocate for receive buffer
-	boost::int32_t m_in_buf_size;
+	boost::int32_t m_receive_buffer_capacity;
 
 	// this holds the 3 last delay measurements,
 	// these are the actual corrected delay measurements.
@@ -1810,10 +1809,10 @@ bool utp_socket_impl::send_pkt(int const flags)
 	int sack = 0;
 	if (m_inbuf.size())
 	{
-        const int max_sack_size = effective_mtu 
-            - sizeof(utp_header)
-            - 2 // for sack padding/header
-            - (close_reason ? 6 : 0);
+		const int max_sack_size = effective_mtu
+			- sizeof(utp_header)
+			- 2 // for sack padding/header
+			- (close_reason ? 6 : 0);
 
 		// the SACK bitfield should ideally fit all
 		// the pieces we have successfully received
@@ -2050,7 +2049,7 @@ bool utp_socket_impl::send_pkt(int const flags)
 	}
 
 	h->timestamp_difference_microseconds = m_reply_micro;
-	h->wnd_size = (std::max)(m_in_buf_size - m_buffered_incoming_bytes
+	h->wnd_size = (std::max)(m_receive_buffer_capacity - m_buffered_incoming_bytes
 		- m_receive_buffer_size, boost::int32_t(0));
 	h->ack_nr = m_ack_nr;
 
@@ -2551,7 +2550,7 @@ bool utp_socket_impl::consume_incoming_data(
 	}
 
 	if (m_read_buffer_size == 0
-		&& m_receive_buffer_size >= m_in_buf_size - m_buffered_incoming_bytes)
+		&& m_receive_buffer_size >= m_receive_buffer_capacity - m_buffered_incoming_bytes)
 	{
 		// if we don't have a buffer from the upper layer, and the
 		// number of queued up bytes, waiting for the upper layer,
@@ -2559,7 +2558,7 @@ bool utp_socket_impl::consume_incoming_data(
 		// more data packets
 		UTP_LOG("%8p: ERROR: our advertized window is not honored. "
 			"recv_buf: %d buffered_in: %d max_size: %d\n"
-			, static_cast<void*>(this), m_receive_buffer_size, m_buffered_incoming_bytes, m_in_buf_size);
+			, static_cast<void*>(this), m_receive_buffer_size, m_buffered_incoming_bytes, m_receive_buffer_capacity);
 		return false;
 	}
 
@@ -2567,7 +2566,7 @@ bool utp_socket_impl::consume_incoming_data(
 	{
 		TORRENT_ASSERT(m_inbuf.at(m_ack_nr) == 0);
 
-		if (m_buffered_incoming_bytes + m_receive_buffer_size + payload_size > m_in_buf_size)
+		if (m_buffered_incoming_bytes + m_receive_buffer_size + payload_size > m_receive_buffer_capacity)
 		{
 			UTP_LOGV("%8p: other end is not honoring our advertised window, dropping packet\n"
 				, static_cast<void*>(this));
@@ -2624,7 +2623,7 @@ bool utp_socket_impl::consume_incoming_data(
 			return true;
 		}
 
-		if (m_buffered_incoming_bytes + m_receive_buffer_size + payload_size > m_in_buf_size)
+		if (m_buffered_incoming_bytes + m_receive_buffer_size + payload_size > m_receive_buffer_capacity)
 		{
 			UTP_LOGV("%8p: other end is not honoring our advertised window, dropping packet %d\n"
 				, static_cast<void*>(this), int(ph->seq_nr));
@@ -2857,8 +2856,7 @@ bool utp_socket_impl::incoming_packet(boost::uint8_t const* buf, int size
 		m_sm->inc_stats_counter(counters::utp_payload_pkts_in);
 
 	// the number of packets that'll fit in the reorder buffer
-    // incoming MTU of 1100 and minimal size of 16 packets are chosen arbitrarly
-	const boost::uint32_t max_packets_reorder = std::max(16, m_in_buf_size / 1100);
+	const boost::uint32_t max_packets_reorder = std::max(16, m_receive_buffer_capacity / 1100);
 
 	if (m_state != UTP_STATE_NONE
 		&& m_state != UTP_STATE_SYN_SENT


src/utp_stream.cpp
```diff
@@ -281,7 +281,7 @@ struct utp_socket_impl
 		, m_written(0)
 		, m_receive_buffer_size(0)
 		, m_read_buffer_size(0)
-		, m_in_buf_size(1024 * 1024)
+		, m_in_buf_size(8 * 1024 * 1024)
```
arvidn: the problem of not connecting this back to the socket option (which I suppose is broken right now, is that right?) is that this is the mechanism used to provide back-pressure to senders, when the next layer isn't consuming data quick enough. Specifically, when the user sets a download rate limit, libtorrent won't read from the socket as fast as it can, but only as fast as it receives download quota. In that situation, you can end up allocating a lot of memory for these buffers.

I know, the download rate limiter should limit the bittorrent requests instead, and still read as fast as possible. It's on my list, but the list is pretty long.

src/utp_stream.cpp
```diff
@@ -137,7 +137,7 @@ enum
 	ACK_MASK = 0xffff,
 
 	// the number of packets that'll fit in the reorder buffer
-	max_packets_reorder = 512,
+	max_packets_reorder = 4096,
```
arvidn: Do you know if this is used in any context where it needs to be a compile-time constant?
If not, perhaps you could remove this constant and replace it with ``std::max(3, m_in_buf_size / 1400)`` (or something like that)

ichorid: `max_packets_reorder` is only used once, and only in`utp_stream.cpp`.
I assume it is safe to do as you suggest and make it depend on `m_in_buf_size`.
The only potential downside is addition of tiny computational overhead, but that is neglible.
I will redo the patch and upload new version.

src/utp_stream.cpp
ichorid: As I understand, there are 2 levels of buffering of incoming packets (not counting the user-provided `m_read_buffer`):
`[OS-managed UDP socket buffer] -> [utp_stream_impl receive buffer]`.

`settings_pack::recv_socket_buffer_size` sets the maximum size of the OS-managed UDP socket buffer. The maximum size for the `utp_socket_impl.m_receive_buffer` is stored in `utp_socket_impl.m_in_buf_size`, and it is only set once at initialization of `utp_socket_impl`. However, the name `utp_socket_impl.m_receive_buffer_size` denotes:
```
    // the sum of all packets stored in m_receive_buffer
    boost::int32_t m_receive_buffer_size;
```
This creates ambiguity in naming.

| Level        | OS socket                                | utp_socket_impl                       |
|--------------|------------------------------------------|-----------------------------------------|
| Current size | ???                                      | `m_receive_buffer_size` |
| Maximum size | `recv_socket_buffer_size` | `m_in_buf_size`         |

If the current implementation of download rate control in libtorrent results in `m_receive_buffer` always filled up to `m_in_buf_size` limit, we really want to set it as low as possible to keep the memory usage reasonable. But we need to provide at least some way to increase it for high-latency connections.

To do this, we could take one of the following approaches:
1. Bound it to the OS socket size, and increase it proportionally if user changes `recv_socket_buffer_size`.
2. Create a new option in `settings_pack`.
3. Make it scale automatically, dependent on latency.

Which one of these you find the best for libtorrent?

arvidn: I think (1) is the simplest and safest option. I'm hesitant to add another option, as there are so many already, and I would be a bit worried about (3) because there would likely need to be a lot of heuristics and instability in practice. Latency isn't just caused by long distances, but also congestion, and if it's caused by congestion you really *don't* want to increase the buffer sizes. That would just make the problem worse.

src/utp_stream.cpp
```diff
@@ -2852,6 +2849,10 @@ bool utp_socket_impl::incoming_packet(boost::uint8_t const* buf, int size
 	if (ph->get_type() == ST_DATA)
 		m_sm->inc_stats_counter(counters::utp_payload_pkts_in);
 
+	// the number of packets that'll fit in the reorder buffer
+    // incoming MTU of 1100 and minimal size of 16 packets are chosen arbitrarly
+	boost::uint32_t max_packets_reorder = std::max(16, m_in_buf_size / 1100);
```
arvidn: this is a bit of a nit-pick, but I like to make anything that is not mutated ``const``

ichorid: changed to `const` and amend/force pushed the result. Is it even OK to --ament/ --force push things while targeting a pull request?

arvidn: yeah, that's great. It can make thinks more difficult to review if there are large changes, that's all

src/utp_stream.cpp
ichorid: There are two problems with implementing (1):

1. `utp_socket_impl` does not have access to neither `m_sm->m_sett`, nor `m_sm->m_sock` - both are declared private in `utp_socket_manager.hpp`. These are the only sources of information about `recv_socket_buffer_size`. Everything is so well isolated in that code... Breaking this isolation would make it ugly.

2. If we increase `m_in_buf_size` on increasing `recv_socket_buffer_size`, we increase it for **every** uTP stream out there, regardless of the real necessity to do that. This is clearly an overkill. We only want it increased on high-latency connections.

Therefore, I advocate to reconsider option (3). We could do that on basis of current latency data in LEDBAT.

arvidn: iirc, some settings are exposed via the utp socket manager (via function calls). So this could be exposed similarly. I agree with (2), it would be ideal to set it adaptively. I would like to make three points though:

1. The receive buffer size (which I suppose may not work right for uTP sockets right now) already has the problem you describe in (2). i.e. it would not be new. It affects TCP sockets the same way for instance.

2. The delay measurements from LEDBAT are only *relative* *one-way* delays. You can't estimate the bandwidth delay product from them. You could use RTT though, which is also maintained (as a running/sliding average). The problem there is still that the total delay you experience on a line may just as likely be due to congestion as to speed-of-light. I think this is fundamentally a difficult problem and a bit risky to mess too much with. The best solution I can think of is to track the *minimum* RTT ever experienced on a connection, and use that for the "physical" latency.

3. I think it's clear that the reorder count and the receive buffer size are intimately coupled. They both represent the number of incoming packet's we're willing to hold while waiting, either for the upper layer to consume more payload, or a lost packet to be re-sent. If the receive buffer (and the number of "reorder packets") adapts to network conditions, it's critical to ensure an adversarial counterpart can't DoS the client.

ichorid: Added method `adjust_receive_buf_size`. It adjusts `m_in_buf_size` linearly by adding 1Mbyte to it for each 100ms RTT latency. It would always clamp `1Mbyte<m_in_buf_size<8Mbytes`.
Testing show it works for high and low latencies as intended.

ichorid: However, during testing I noticed that `min_rtt` is indeed a local variable, and could not only shrink, but actually grow. And logs show it does, regularly. Is it by design? I thought it was for keeping the smallest seen latency during connection with a single peer (that is a single uTP stream, right?).

Besides, we have another serious problem: when I add some _variation_ to latency, both patched and unpatched versions become very slow, even for moderate latencies, like 200ms RTT. I will investigate this issue further.

src/utp_stream.cpp
```diff
@@ -3408,6 +3420,22 @@ bool utp_socket_impl::incoming_packet(boost::uint8_t const* buf, int size
 	return true;
 }
 
+boost::int32_t utp_socket_impl::adjust_receive_buf_size(boost::uint32_t const& min_rtt)
+{
+	// maximum value for RTT we would still scale for
+	const boost::int32_t rtt_ceil = 800000;
```
arvidn: the comment makes it sound like the effective RTT to be used in this function is ``min(min_rtt, rtt_ceil)``, but that's not the case. ``rtt_ceil`` is just used to calculate the coefficient. I suppose the clamping at the end makes any ``min_rtt`` that's higher still not scale the buffer. There's a potential robustness issue though, if ``min_rtt`` is large enough the multiplication with ``k`` might overflow.

src/utp_stream.cpp
```diff
@@ -3408,6 +3420,22 @@ bool utp_socket_impl::incoming_packet(boost::uint8_t const* buf, int size
...
+
+	// linear buffer increase coefficient. Defines how fast
+	// m_in_buf_size would increase/decrease with increase/decrease of minimum RTT
+	const boost::int64_t k = (boost::int64_t(m_in_buf_size_max - m_in_buf_size_min)<<16) / rtt_ceil;
+
+	//calculate new size of the buffer and clamp it to max allowed value
+	const boost::int32_t adjusted_buffer_size = m_in_buf_size_min + boost::int32_t((min_rtt * k)>>16);
+	const boost::int32_t clamped_buffer_size = std::min(adjusted_buffer_size, int(m_in_buf_size_max));
+
+	return size_t(clamped_buffer_size);
```
arvidn: I think the ``size_t`` conversion here was a mistake. ``clamped_buffer_size`` already has the right type

src/utp_stream.cpp
```diff
@@ -3408,6 +3420,22 @@ bool utp_socket_impl::incoming_packet(boost::uint8_t const* buf, int size
 	return true;
 }
 
+boost::int32_t utp_socket_impl::adjust_receive_buf_size(boost::uint32_t const& min_rtt)
```
arvidn: as far as  can tell, this function doesn't use any members of ``utp_socket_impl`` so it should be a free function (and ideally have a unit test)

src/utp_stream.cpp
```diff
@@ -2438,7 +2444,12 @@ void utp_socket_impl::ack_packet(packet* p, time_point const& receive_time
 		, static_cast<void*>(this), seq_nr, p->size - p->header_size, rtt / 1000);
 
 	m_rtt.add_sample(rtt / 1000);
-	if (rtt < min_rtt) min_rtt = rtt;
+	if (rtt < min_rtt)
+	{
+		min_rtt = rtt;
+		// tune receive buffer size according to the new RTT value
+		m_in_buf_size = adjust_receive_buf_size(min_rtt);
```
arvidn: my first thought here was: "an attacker can deliberately add delay, to make the RTT long and force the other endpoint to allocate a larger receive window". 8 MiB per connection may not be the end of the world on a desktop, but on a raspberry PI it might. I think the upper end still should be configurable via the same socket option used to configure the TCP connections. I suppose that's an independent change from this though, so I can do that later.

arvidn: I have another nagging feeling. This will adjust the advertised receive window, right? Which means that the other end may currently believe that we have a larger advertised receive window and has sent packets to fill it. If we shrink it now, while those packets are in-flight, we may experience an unexpected error, where we sometimes receive more packets than our m_in_buf_size allows. Maybe that's not a big deal. I think the proper way of handling that would be to shrink the advertised receive window only by the number of bytes that we ACK, as we ACK them. That way we'll never "take back" a receive window we've already advertised.

If we instead just accept that sometimes we receive more packets than the receive window allows (rather than treating that as a fatal error), we may be exposed to an attack.

src/utp_stream.cpp
arvidn: yeah, the current use for min_rtt is as a sanity check for the one-way delay measurements. As a protection against a malicious end-point, the one way delay measurement is clamped to the minimum RTT of any of the packets that were just ACKed. To keep a per-connection min_rtt you would need to add a new member (e.g. ``m_min_rtt``).

As for the issue you see with jitter, this is most likely because the one-way delay measurements are used to infer congestion. If you have non-congestion jitter, uTP will still interpret the high delay measurements as a signal to slow down, and the low delays as a signal to speed up.

Since you may expect jitter, you could increase the target delay. That's a session-wide setting. It defaults to 100ms, if you expect to have a range of [0,200] ms additional jitter, you should raise it by 200. Keep in mind the one-way delay measurement does not take into account any fixed latency. Like switch-processing or speed-of-light. It *only* looks at the variation in delays.

src/utp_stream.cpp
ichorid: Regarding the potential attack: a single stream corresponds to a single connection. It means that even if the attacker sends a high RTT packet, it would increase the buffer size only for this single connection. 8 MiB is not a big deal in this case, even for Raspberry Pi. To truly exploit this feature, the attacker needs to create dozens of individual connections.
If it is possible for attacker to trick victim to open multiple independent uTP streams to his host, the attack becomes real (i.e. emulate several seeds from a single ip). But in this case, it does not matter what the buffer limit is: as long as the attacker can induce the victim to open unlimited number of connections, he could bring down _any_ victim, regardless of her buffers size.

So, it would be a good safety measure to give the user some control on the maximum size of her buffer. But this matter is closely related to other safety limits, an I am not an expert on those.

But regardless of safety measures, if we want users to really benefit from this fix, we have to make `in_buf_size_max` 8 MiB by default.

ichorid: About the problem with reduction of advertised window size, I had the same feeling when writing this patch. I'll need to do some more thinking/working on this problem. As we have seen, packet drop have non-trivial consequences at high latencies.

src/utp_stream.cpp
ichorid: moved to free function, fixed clamping and types

arvidn:
you mentioned something about making SACK support more bits as well. That sounds like an important fix, but it's not part of this patch.

ichorid:
I've just made a commit for the SACK problem. Still need to address other things:

1. Advertised window shrink / packet drop
2. RTT min not stable
3. when applied on top of RC_1_1 this patch does not produce the expected result. Work OK on top of 1.1.6 though.
4. When at least some variance in latency is applied, the patch loses its effect.

ichorid:
I rechecked everything from scratch for RC_1_1. Patch works as intended. However, something introduced by commit `ad82be145` breaks very-high-latency performance on `master` branch.

src/utp_stream.cpp
ichorid: When the other side sends us a packet that don't fit into our receive buffer, we drop it and blame the other side for "not honoring our advertised window". But it could be really a consequence of our own actions (or inaction):
1. Our upper layer didn't read bytes fast enough. Maybe it really could not cope with all the data, or maybe we did it intentionally, as a form of bandwidth control.
2. We shrinked our buffer size. This is the new condition that this pull request introduces.

But in both cases it is wrong to blame the partner, because _at the moment of sending_ the partner honored our congestion window! And if we want to know the real reason for his action, we need to take this fact into account. And we only want to drop packets if our partner _really_ didn't honored our advertised window size _at the moment of sending_ of the offending packet.
To do that, we need to keep the history of our advertised window, and consult it every time we receive a packet.
One side effect of introducing this design would be the smoother implicit incoming bandwidth control. It would be based not on packet drop, but on _signaling_ our partner that we approach congestion by reducing our advertised window size.

ichorid: Using `min_rtt` as a source of RTT data for `adjust_receive_buffer_size`  was wrong, because that is a local variable. In the latest commit I switched it to use sliding average (`m_rtt.mean()`) instead. It includes latency added by queuing, but that is OK because we have to consider delays introduced by any kind of source.

The only potential pitfall is the possibility that by including another feedback loop we could potentially create another oscillation cycle based on packet drop (on abrupt shrinking of buffer). The only way to prevent this effect completely, while keeping buffer adjusting heuristics, is to use historical data on advertised window size, as I suggested above.

@arvidn, adding that would be more than a slight change in `utp_stream.cpp`, so I ask your blessing on this ))

ichorid:
The most important remaining problem is that even 1ms of latency variance drops performance severely, regardless of base latency. I will look into this issue, it should not be like this.

src/utp_stream.cpp
arvidn: I'll move this conversation out to top-level

arvidn:
> Our upper layer didn't read bytes fast enough. Maybe it really could not cope with all the data, or maybe we did it intentionally, as a form of bandwidth control.

Actually, this is not a problem. When the receiver ACKs, the receive window it will advertise is the *remaining* size of its receive buffer, along with the sequence number of the most recent received packet. When the sending side receives this, it may already have bytes in-flight, but it will make sure that it won't send more bytes such that its total in-flght bytes exceeds the receive window. Even though the other end may have consumed more or received some of the in-flight bytes by now, it doesn't matter. The sender assumes that it hasn't, which is a conservative assumption and will ensure that it never exceeds the receive buffer. This mechanism is important for reporting back-pressure to the sender.

> To do that, we need to keep the history of our advertised window, and consult it every time we receive a packet.

I believe only (2) would need fixing, and deferring the shrinkage of the receive window seems simpler. That would just be one more counter, and whenever it's > 0, we decrement it instead of replenishing the receive window when incoming data is consumed by the upper layer.

> One side effect of introducing this design would be the smoother implicit incoming bandwidth control. It would be based not on packet drop, but on signaling our partner that we approach congestion by reducing our advertised window size.

There's an important distinction between *congestion* and *back-pressure*. The advertised receive window provides the latter, the congestion window provides the former (and is controller by one-way delays, packet loss and time outs).

> The only potential pitfall is the possibility that by including another feedback loop we could potentially create another oscillation cycle based on packet drop (on abrupt shrinking of buffer). The only way to prevent this effect completely, while keeping buffer adjusting heuristics, is to use historical data on advertised window size, as I suggested above.

I agree that this is a risk of using the continuously measured RTT. In my mind the problem is that a higher RTT invites more packets in-flight, which in turn can cause the RTT to increase (by adding queueing delay). I feel pretty confident that tracking the lowest ever measured RTT would perfectly solve that problem. Note that the adjustment we're talking about is to the advertised receive window, right? Not the congestion window.

arvidn:
> The most important remaining problem is that even 1ms of latency variance drops performance severely, regardless of base latency. I will look into this issue, it should not be like this.

Yes, that definitely sounds wrong. I'd be happy to look at graphs for this scenario

ichorid:
These are the plots for 100ms +/- 1ms latency on each side of connection (200+/-2ms RTT).
As you can see, it is a pure primordial chaos of white noise.
It looks like LEDBAT goes crazy, as if it is using raw delay data, instead of smoothing it with sliding average or something.
Comments are very welcome.

utp out0x7f7ab8005980-our-delay
utp out0x7f7ab8005980-cwnd
utp out0x7f7ab8005980-send-packet-size
utp out0x7f7ab8005980-our_delay_base
utp out0x7f7ab8005980-their_delay_base
utp out0x7f7ab8005980-their_delay
utp out0x7f7ab8005980-slow-start
utp out0x7f7ab8005980-uploading_packets
utp out0x7f7ab8005980-uploading
utp out0x7f7ab8005980 packet_sizes

ichorid:
On the logs side, there is a heavy packet reordering:
```
[000388430181] 0x7f7ab8005980: ERROR: incoming packet ack_nr:10369 our seq_nr:10384 our acked_seq_nr:10375 (ignored)
[000389232441] 0x7f7ab8005980: ERROR: incoming packet ack_nr:10392 our seq_nr:10412 our acked_seq_nr:10398 (ignored)
[000389233361] 0x7f7ab8005980: ERROR: incoming packet ack_nr:10394 our seq_nr:10412 our acked_seq_nr:10398 (ignored)
[000389233511] 0x7f7ab8005980: ERROR: incoming packet ack_nr:10394 our seq_nr:10412 our acked_seq_nr:10398 (ignored)
[000390431649] 0x7f7ab8005980: ERROR: incoming packet ack_nr:10424 our seq_nr:10441 our acked_seq_nr:10433 (ignored)
[000392029249] 0x7f7ab8005980: ERROR: incoming packet ack_nr:10457 our seq_nr:10472 our acked_seq_nr:10461 (ignored)
[000392429235] 0x7f7ab8005980: ERROR: incoming packet ack_nr:10467 our seq_nr:10479 our acked_seq_nr:10471 (ignored)
[000395227841] 0x7f7ab8005980: ERROR: incoming packet ack_nr:10546 our seq_nr:10559 our acked_seq_nr:10551 (ignored)
[000395227850] 0x7f7ab8005980: ERROR: incoming packet ack_nr:10546 our seq_nr:10559 our acked_seq_nr:10551 (ignored)
[000396031742] 0x7f7ab8005980: ERROR: incoming packet ack_nr:10567 our seq_nr:10584 our acked_seq_nr:10571 (ignored)
[000397231484] 0x7f7ab8005980: ERROR: incoming packet ack_nr:10606 our seq_nr:10624 our acked_seq_nr:10613 (ignored)
[000398832850] 0x7f7ab8005980: ERROR: incoming packet ack_nr:10656 our seq_nr:10668 our acked_seq_nr:10660 (ignored)
```

ichorid:
Small part of sender's log:
```
[000005607404] 0x7f7ab8005980: incoming packet reply_micro:199262 base_change:0
[000005607406] 0x7f7ab8005980: updating timeout to: now + 500
[000005607407] 0x7f7ab8005980: acked packet 634 (1378 bytes) (rtt:400)
[000005607408] 0x7f7ab8005980: acked packet 637 (1378 bytes) (rtt:398)
[000005607409] 0x7f7ab8005980: incoming packet seq_nr:22111 ack_nr:637 type:ST_STATE id:47698 size:0 timestampdiff:200736 timestamp:3813688237 our ack_nr:22110 our seq_nr:644 our acked_seq_nr:637 our state:CONNECTED
[000005607410] 0x7f7ab8005980: do_ledbat delay:57 off_target: 99943 window_factor:0.250000 target_factor:0.999420 scaled_gain:749.542236 cwnd:12196 slow_start:0
[000005607414] 0x7f7ab8005980: mtu:1398 in_flight:11024 adv_wnd:4713077 cwnd:12945 acked_bytes:2756 cwnd_full -> 0
[000005607420] 0x7f7ab8005980: sending packet seq_nr:644 ack_nr:22110 type:ST_DATA id:47697 target:192.168.0.14:6881 size:1398 error:Success send_buffer_size:7853 cwnd:12945 adv_wnd:4713077 in-flight:8268 mtu:1398 timestamp:3813887515 time_diff:199262 mtu_probe:0 extension:0
[000005607433] 0x7f7ab8005980: sending packet seq_nr:645 ack_nr:22110 type:ST_DATA id:47697 target:192.168.0.14:6881 size:1398 error:Success send_buffer_size:6475 cwnd:12945 adv_wnd:4713077 in-flight:9646 mtu:1398 timestamp:3813887528 time_diff:199262 mtu_probe:0 extension:0
[000005607452] 0x7f7ab8005980: sending packet seq_nr:646 ack_nr:22110 type:ST_DATA id:47697 target:192.168.0.14:6881 size:1398 error:Success send_buffer_size:5097 cwnd:12945 adv_wnd:4713077 in-flight:11024 mtu:1
398 timestamp:3813887546 time_diff:199262 mtu_probe:0 extension:0
[000005607455] 0x7f7ab8005980: no space in window send_buffer_size:5097 cwnd:12945 adv_wnd:4713077 in-flight:12402 mtu:1398
[000005607457] 0x7f7ab8005980: skipping send seq_nr:647 ack_nr:22110 id:47697 target:192.168.0.14:6881 header_size:20 error:Success send_buffer_size:5097 cwnd:12945 adv_wnd:4713077 in-flight:12402 mtu:1398 effective-mtu:1398
[000005607460] 0x7f7ab8005980: actual_delay:200736 our_delay:0.057000 their_delay:0.127000 off_target:99.943001 max_window:12945 upload_rate:0 delay_base:199087 delay_sum:0.184000 target_delay:100 acked_bytes:2756 cur_window:12402 scaled_gain:0.000000 rtt:368 rate:35179 quota:0 wnduser:4713077 rto:500 timeout:500 get_microseconds:3813887499 cur_window_packets:10 packet_size:1398 their_delay_base:199135 their_actual_delay:199262 seq_nr:647 acked_seq_nr:637 reply_micro:199262 min_rtt:398 send_buffer:5097 recv_buffer:0 fast_resend_seq_nr:638 ssthres:4742 
[000005607468] 0x7f7ab8005980: calling write handler written:4134
[000005607469] 0x7f7ab8005980: calling write handler written:4134 ec:Success kill:0
[000005607480] 0x7f7ab8005980: add_write_buffer 5097 bytes
[000005607480] 0x7f7ab8005980: new write handler. 5097 bytes to write
[000005607482] 0x7f7ab8005980: no space in window send_buffer_size:5097 cwnd:12945 adv_wnd:4713077 in-flight:12402 mtu:1398
[000005607483] 0x7f7ab8005980: skipping send seq_nr:647 ack_nr:22110 id:47697 target:192.168.0.14:6881 header_size:20 error:Success send_buffer_size:5097 cwnd:12945 adv_wnd:4713077 in-flight:12402 mtu:1398 effective-mtu:1398
[000005608217] 0x7f7ab8005980: their_delay::add_sample:199101 prev_base:199135 new_base:199101
[000005608220] 0x7f7ab8005980: incoming packet reply_micro:199101 base_change:-34
[000005608222] 0x7f7ab8005980: updating timeout to: now + 500
[000005608223] 0x7f7ab8005980: incoming packet seq_nr:22111 ack_nr:637 type:ST_STATE id:47698 size:0 timestampdiff:199547 timestamp:3813689211 our ack_nr:22110 our seq_nr:647 our acked_seq_nr:637 our state:CONNECTED
[000005608225] 0x7f7ab8005980: no space in window send_buffer_size:5097 cwnd:12945 adv_wnd:4718589 in-flight:12402 mtu:1398
[000005608230] 0x7f7ab8005980: skipping send seq_nr:647 ack_nr:22110 id:47697 target:192.168.0.14:6881 header_size:20 error:Success send_buffer_size:5097 cwnd:12945 adv_wnd:4718589 in-flight:12402 mtu:1398 effective-mtu:1398
[000005608232] 0x7f7ab8005980: subscribe drained
[000005608439] 0x7f7ab8005980: their_delay::add_sample:200365 prev_base:199101 new_base:199101
[000005608440] 0x7f7ab8005980: incoming packet reply_micro:200365 base_change:0
[000005608440] 0x7f7ab8005980: ERROR: incoming packet ack_nr:633 our seq_nr:647 our acked_seq_nr:637 (ignored)
[000005610826] 0x7f7ab8005980: their_delay::add_sample:199971 prev_base:199101 new_base:199101
[000005610828] 0x7f7ab8005980: incoming packet reply_micro:199971 base_change:0
[000005610830] 0x7f7ab8005980: updating timeout to: now + 500
[000005610831] 0x7f7ab8005980: incoming packet seq_nr:22111 ack_nr:637 type:ST_STATE id:47698 size:0 timestampdiff:200553 timestamp:3813690949 our ack_nr:22110 our seq_nr:647 our acked_seq_nr:637 our state:CONNECTED
[000005610834] 0x7f7ab8005980: no space in window send_buffer_size:5097 cwnd:12945 adv_wnd:4718589 in-flight:12402 mtu:1398
[000005610838] 0x7f7ab8005980: skipping send seq_nr:647 ack_nr:22110 id:47697 target:192.168.0.14:6881 header_size:20 error:Success send_buffer_size:5097 cwnd:12945 adv_wnd:4718589 in-flight:12402 mtu:1398 effective-mtu:1398
```

ichorid:
These errors are triggered by this condition:
```
            || compare_less_wrap(ph->ack_nr, m_acked_seq_nr
                - dup_ack_limit, ACK_MASK)))
```

src/utp_stream.cpp
```diff
@@ -799,6 +802,23 @@ void utp_socket_drained(utp_socket_impl* s)
 	s->maybe_trigger_send_callback();
 }
 
+boost::int32_t adjust_receive_buf_size(boost::uint32_t const& rtt)
```
arvidn: the ``rtt`` should not be taken by ``const&`` here, just by value (but it still makes sense to declare it ``const``)

src/utp_stream.cpp
```diff
@@ -799,6 +802,23 @@ void utp_socket_drained(utp_socket_impl* s)
...
+{
+	// maximum value for RTT we would still scale for. We use it to clamp min_rtt 
+	// and to calculate the linear buffer increase coefficient
+	const boost::int32_t rtt_ceil = 800000;
```
arvidn: maybe the RTT type can be changed to use milliseconds here, to match the rest of the utp code (for RTT). In master I hope to make all these timestamps use chrono to fix type safety

src/utp_stream.cpp
```diff
@@ -1813,22 +1823,38 @@ bool utp_socket_impl::send_pkt(int const flags)
...
+	{
+        const int max_sack_size = effective_mtu 
+            - sizeof(utp_header)
+            - 2 // for sack padding/header
+            - (close_reason ? 6 : 0);
+
+		// the SACK bitfield should ideally fit all
+		// the pieces we have successfully received
+		sack = (m_inbuf.span() + 7) / 8;
+		if (sack > max_sack_size) sack = max_sack_size;
```
arvidn: is this really sufficient? I seem to recall the code that actually generates the SACK header also to be constrained to a unt32

ichorid:
When using variable latency, `netem` implicitly introduces reorder in sequence of UDP packets, when they are written to socket as a burst. This happens because the time between two sockets put into buffer could be less than the variance between latencies, and it could send the second one faster than the first one.
This problem is fixed by including bandwidth rate limiting command in netem settings line. It introduces TBF-like service, that "spaces out" packets to fit to some rate. As a nice side effect, it solves our problem ))
`tc qdisc add dev eth0 root netem delay 100ms 1ms rate 100Mbit`

Nonetheless, when we add even 1ms latency variance (to 200ms RTT), uTP perfromance drops to 300KByte/s. TCP does about 15MBytes/s on this connection, and patched uTP does 5Mbytes/s, when no latency variance is introduced. Need to look further into LEDBAT behaviour, I guess.

src/utp_stream.cpp
ichorid: Yes, this is really sufficient. `write_sack` uses `sack` variable, that denotes the SACK size in bytes. It writes the size of SACK in the packet itself. That is where `parse_sack` and `remove_sack_header` use that.
```
void utp_socket_impl::write_sack(boost::uint8_t* buf, int size) const
{
    INVARIANT_CHECK;

    TORRENT_ASSERT(m_inbuf.size());
    int ack_nr = (m_ack_nr + 2) & ACK_MASK;
    boost::uint8_t* end = buf + size;

    for (; buf != end; ++buf)
    {
        *buf = 0;
        int mask = 1;
        for (int i = 0; i < 8; ++i)
        {
            if (m_inbuf.at(ack_nr)) *buf |= mask;
            mask <<= 1;
            ack_nr = (ack_nr + 1) & ACK_MASK;
        }
    }
}
```

Anyway, it would be nice to add huge SACK test to unit tests.

arvidn: yeah, I took some notes. the ``parse_sack()``, ``write_sack()`` and ``remove_sack()`` could most likely be made free functions and easily unit tested. I'm mainly working on a bigger issue in RC_1_1 right now though.

arvidn:
with this option to rate limit, do you eliminate all reordering?

ichorid:
Almost all reordering is eliminated with this option. Only a few reordering notices remain in leecher's log.

src/utp_stream.cpp
ichorid: When I solve the variance problem and other more important issues with this patch, I could take this problem, and write unit tests.

src/utp_stream.cpp
```diff
@@ -555,6 +571,9 @@ struct utp_socket_impl
 	// average RTT
 	sliding_average<16> m_rtt;
 
+    // lowest RTT observed during the whole existence time of connection
+    boost::int32_t m_min_rtt;
```
arvidn: this is signed. when you initialize it in the constructor you set it to the *unsigned* max. they should be consistent (and I think signed may be slightly less error prone)

src/utp_stream.cpp
```diff
@@ -1129,7 +1148,7 @@ size_t utp_stream::read_some(bool clear_buffers)
 		target->buf = static_cast<char*>(target->buf) + to_copy;
 		TORRENT_ASSERT(int(target->len) >= to_copy);
 		target->len -= to_copy;
-		m_impl->m_receive_buffer_size -= to_copy;
+        m_impl->decrease_receive_buffer(to_copy);
```
arvidn: indentation..

src/utp_stream.cpp
```diff
@@ -242,7 +242,14 @@ struct packet
...
 // is none, is "m_receive_buffer" and "m_receive_buffer_size"
-// respectively.
+// respectively. 
+
+// "m_receive_buffer_capacity" marks the maximum amount of
+// bytes receive buffer should store at once. If we
+// want to reduce this limit during runtime, we must be careful
+// not to reduce it below the window size we advertised
+// to our peer earlier. The number of bytes we want to 
+// reduce receive buffer by is stored in "m_receive_buffer_surplus".
```
arvidn: every time we send an ACK, we advertise the whole window. In order to decrease it, we must always wait for the upper layer to consume bytes (which normally would free up space in the receive buffer). Instead of freeing it up, we shrink the receive buffer by the consumed amount (or less if we're not shrinking it that much).

As I see it, there are two ways of implementing this. (1) is to have two separate window sizes, one that we're aiming for and one that we've advertised (the effective window), and then at the appropriate time inch the effective window closer to the target.

However, I think the (2) is a bit simpler (and less intrusive). We keep the current effective window size, but in addition we have a *delta*, indicating how much we want to lower it by. Every time the upper layer consumers bytes, we reduce the window size by min(consumed, want_to_reduce_wnd) and also reduce the ``want_to_reduce_wnd`` by the same amount.

When we want to increase the size, we first "eat into" the ``want_to_reduce`` and then increase the effective window size.

The main reason I think that's simpler is that the existing window size variable does not change meaning, and any code interacting with it can stay the same. My first reading of this code was that you've implemented (2) and ``m_receive_buffer_surplus`` is this counter I called ``want_to_reduce_wnd`` above.

src/utp_stream.cpp
```diff
@@ -1524,6 +1543,24 @@ std::size_t utp_socket_impl::available() const
 	return m_receive_buffer_size;
 }
 
+
+void utp_socket_impl::decrease_receive_buffer(int const bytes)
+{
```
arvidn: as far as I can tell, all this function should do is to increment ``m_receive_buffer_surplus`` by ``bytes``. The actual shrinkage of ``m_receive_buffer_capacity`` must be deferred.

src/utp_stream.cpp
```diff
@@ -2468,8 +2505,22 @@ void utp_socket_impl::ack_packet(packet* p, time_point const& receive_time
...
+        m_min_rtt = min_rtt;
+        // and tune receive buffer size according to it
+        boost::int32_t const new_capacity = get_optimized_receive_buf_capacity(m_min_rtt/1000);
+        // only use surplus mechanism if capacity should really
+        // descrease. In case it should increse, like after first RTT,
+        // change it directly;
+        if (new_capacity <= m_receive_buffer_capacity - m_receive_buffer_surplus)
+            m_receive_buffer_surplus += (m_receive_buffer_capacity - m_receive_buffer_surplus - new_capacity);
+        else
+            m_receive_buffer_capacity = new_capacity;
```
arvidn: I don't understand this logic. as far as I can tell, this would still make is pull receive window "from under the feet" of the sender. If we have advertised 1 MiB, the sender may already have sent that megabyte. The only way the receive window is "replenished" is by our upper layer consuming bytes, in which case the advertised receive window is increased. This is the only opportunity we have to decrease the window and preserve consistency.

arvidn: I think at this point in the code, you should either adjust the ``surplus`` variable (in any direction) *OR* increase ``m_receive_buffer_capacity``.

arvidn:
it looks like all the pieces are in place now. just some tweaking and moving things around left I think.
Sorry this is taking a while, but it's pretty sensitive code (and not the most well-written, I'll be the first to admit)

src/utp_stream.cpp
```diff
@@ -799,6 +821,23 @@ void utp_socket_drained(utp_socket_impl* s)
 	s->maybe_trigger_send_callback();
 }
 
+boost::int32_t get_optimized_receive_buf_capacity(boost::uint32_t const rtt)
```
arvidn: the build error here is that you should either put this in the anonymous namespace (to get internal linkage) or add a forward declaration in the header. In order to have a unit test for it it has to have the forward declaration.

src/utp_stream.cpp
```diff
@@ -2046,7 +2109,7 @@ bool utp_socket_impl::send_pkt(int const flags)
 	}
 
 	h->timestamp_difference_microseconds = m_reply_micro;
-	h->wnd_size = (std::max)(m_in_buf_size - m_buffered_incoming_bytes
+	h->wnd_size = (std::max)(m_receive_buffer_capacity - m_buffered_incoming_bytes
```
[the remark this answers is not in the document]
ichorid: There are two buffers for incoming data:

1. `m_read_buf` - the primary buffer. The upper layer reads data from it directly.
2. `m_receive_buffer` - the secondary buffer. It is used as temporary storage when there is no more space available in the primary buffer.

My patch only touches the second one, while the window size is affected by both.  When data is read from `m_receive_buffer` by the upper layer, we replenish the window here:
```
       target->len -= to_copy;
        m_impl->decrease_receive_buffer(to_copy);
```
This is where we call `decrease_receive_buffer()`, to chop a piece of `m_receive_buffer_surplus`, and if it is not enough - a piece of real `m_receive_buffer_capacity`.
I will change the things around a little bit to make them more clear.

src/utp_stream.cpp
ichorid: The problem is, we can't start with the buffer of maximal possible capacity (for the sake of not blowing away Raspberry Pi's with lots of saved torrents), and only decrease it afterward. We must start with 1 MByte buffer, _increase_ it after the first round trip, and only then start shrinking it on further `m_min_rtt` updates.
This construction enables us to handle the increase in buffer capacity after first `m_min_rtt` update. As it is written now, it does not account for all possible sequences of increase and decrease of buffer capacity. But it does not have to because the only possible scenario is:
`default -> increase -> decrease -> decrease ...`,

When the buffer capacity must be decreased, we first look if the new capacity is less than the capacity we had already _planned_ our buffer to have (`m_receive_buffer_capacity - m_receive_buffer_surplus`). If it is, we add the difference between our current _planned_ capacity and our new _planned_ capacity to `m_receive_buffer_surplus`.

ichorid:
I added another patch important for supporting high-latency links: a2836f3814c5fee03ccbd657fd879507e9d510ae. It makes `target_delay` increase if `m_min_rtt` becomes more than 200ms. It only does so in case the user has not changed the default `target_delay`. If user provides target delay value through `settings_pack::target_delay`, the code would always prefer the provided value over the one calculated from RTT.

ichorid:
This is really all we could do to better support higher latency links (without messing with error recovery mechanisms, but that's completely different story).

src/utp_stream.cpp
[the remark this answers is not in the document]
arvidn: right, I've referred to ``m_read_buf`` to the upper layer. This is the same buffer that's used to read from TCP sockets. You say the window size is affected by both buffers? I was under the impression that once the bytes have been passed on the the upper layer, they don't cont anymore, and the advertised receive buffer is replenished.

are you saying your code already worked the way I described I would have expected it to? I will take another look.

The comment about how min_rtt only can decrease is a good point though, please add that in a comment in the code.

ichorid:
You're right, it deserves a separate branch. And I haven't thought about the problem of balancing multiple connections to a single peer.
I will move it out of this pull request and create a new one when I solve the multiple connections problem.

src/utp_stream.cpp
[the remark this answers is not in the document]
ichorid: I have added the comment on buffer size evolution.

src/utp_stream.cpp
```diff
@@ -2468,8 +2505,32 @@ void utp_socket_impl::ack_packet(packet* p, time_point const& receive_time
...
+		// +---------------------------+------------------------------+-----------+----------------+
+		// | m_receive_buffer_capacity | 1 MByte                      | increase  | decrease       |
+		// +---------------------------+------------------------------+-----------+----------------+
+		// We only use surplus mechanism if the capacity should really
+		// decrease. In case it should increase (e.g. after the first RTT),
+		// we change it directly.
+		if (new_capacity <= m_receive_buffer_capacity - m_receive_buffer_surplus)
+			m_receive_buffer_surplus += (m_receive_buffer_capacity - m_receive_buffer_surplus - new_capacity);
+		else
+			m_receive_buffer_capacity = new_capacity;
```
arvidn: if ``new_capacity`` is 1000, ``m_receive_buffer_capacity`` is 1100 and ``m_receive_buffer_surplus`` is 200. i.e. we want ``m_receive_buffer_capacity`` to be 900, we're just waiting for the right moment to adjust it. This logic doesn't look right:

``new_capacity <= m_receive_buffer_capacity - m_receive_buffer_surplus``, i.e.:

(1000 <= 1100 - 200) -> false

which means we'll set ``m_receive_buffer_capacity`` to 1000 (i.e. not deferring the decrement), and perhaps more importantly, without adjusting the surplus, so now we effectively are looking to decrement the capacity to 800.

I think the second case (the else-block) here should have these asserts:

``TORRENT_ASSERT(m_receive_buffer_capacity < new_capacity);``
``TORRENT_ASSERT(m_receive_buffer_surplus == 0);``

and I think the first case (the if-block) should have this assert after the adjustment has been made:

``TORRENT_ASSERT(m_receive_buffer_surplus >= 0);``

src/utp_stream.cpp
arvidn: I see now. the ``bytes`` arguments essentially is how much we're allowed to decrement the capacity, i.e. the number of bytes consumed by the upper layer.

src/utp_stream.cpp
ichorid: (1000 <= 1100 - 200) -> false
case really cannot happen with the current implementation of `get_optimized_receive_buf_capacity()`, because the latter would never recommend buffer size lower than its starting value (that is 1000 in this example). So, if everything goes as planned, the asserts would never trigger.

However, I agree that making pieces interdependent on each other is a bad design pattern. I will redo this procedure to be safe and will add the proposed asserts.

arvidn: I see. yeah, encoding knowledge about the internals of that function there is risky. Imagine a future where the *minimum* RTT is changed to a running average RTT for instance. (I can think of other reasons that's a bad idea, but still)

ichorid: By coincidence, this is almost exactly what I planned to implement today. Real-life test runs of Tribler showed that connection circuit could switch intermediate hops several times during its lifetime, resulting in dramatic changes in connection properties (RTT, bandwidth). Therefore, to support this kind of mutations, we have to use something like a circular buffer of lowest seen RTTs in last 1 minute. Maybe `timestamp_history` object could be modified to fit this purpose.

src/utp_stream.cpp
```diff
@@ -290,9 +290,9 @@ struct utp_socket_impl
 		, m_write_buffer_size(0)
 		, m_written(0)
 		, m_receive_buffer_size(0)
-		, m_receive_buffer_surplus(0)
 		, m_read_buffer_size(0)
 		, m_receive_buffer_capacity(in_buf_size_min)
+		, m_desired_receive_buffer_capacity(in_buf_size_min)
```
arvidn: moving this caused the build failure I believe

src/utp_stream.cpp
```diff
@@ -1547,18 +1565,12 @@ std::size_t utp_socket_impl::available() const
...
-	{
-		m_receive_buffer_capacity -= m_receive_buffer_surplus;
-		m_receive_buffer_surplus = 0;
-	}
+	TORRENT_ASSERT(bytes > 0);
+	const int delta = m_receive_buffer_capacity - m_desired_receive_buffer_capacity;
+	if (delta == 0)
+		return;
+	TORRENT_ASSERT(delta > 0);
+	m_receive_buffer_capacity -= std::min(bytes, delta);
```
arvidn: I think this is much easier to understand!

arvidn:
it also looks like ``utp_stream.cpp`` is exposed to ``windows.h`` and expansion of the ``max`` macro needs to be suppressed. See: https://ci.appveyor.com/project/arvidn/libtorrent/build/1.0.8536/job/u6jbidbrufwx7vq8#L3045

the solution is to either figure out where ``windows.h`` gets included from and remove it, or to say: ``(std::max)(a, b)``

ichorid:
Stack guys advise putting `#define NOMINMAX` before `windows.h` inclusion. I'll try that.

ichorid:
It didn't work. Travis spew: `error: extended initializer lists only available with -std=c++11 or -std=gnu++11 [-Werror]`.
I'll redo that as `std::max( ,std::max( )) for compatibility.
\*sigh\*

arvidn:
this last commit makes me a bit nervous. I think up until that though, this patch looks good and is ready to be merged. Changing min_rtt to mean is a bit scary, and I think that should be a separate PR. I think it would be less scary to use the min_rtt in the last 5 minutes, or something like that instead. As you suggested, similar to delay base history.

ichorid:
Ok, I removed that last one, and changed max reachable window size to 4 MBytes. In my tests the window never reached even 2,5 Mbytes on 800ms RTT, so that is more than enough. I guess we could consider the work done here.

However, there is one thing that irks me: in real-life high RTT tests the window never exceeds even 400Kbytes. Moreover, using min seen RTT for Tribler tunnels means it would eventually only account for RTT to the closest hop, and that defies the whole point of having RTT-based buffer capacity tuning.
These points means that all my code relating to dynamical change of buffer capacity makes no sense, because in real life everything works good enough as it is. And we should remove it, because overengineering is a Bad Thing (TM).
On the other hand, I put much effort in making it readable, and someone could make use of it in the future. And I plan to return to it at some point, to do more proper tests and really decide if we need this dynamic capacity tuning.

@arvidn, if you decide, I would clean dynamic buffer capacity tuning out of this pull request and keep it in my own fork.

ichorid:
You not mistaken about how Tribler tunnels work. Libtorrent only runs at the leecher, and intermediate nodes  just forward packets. The problem is, during lifetime of the session, the hops constituting the session could change several times. And that means that connection properties could radically change as well - several times. RTT, packet loss, bandwidth - everything change during lifetime of a single session, because hops behind the exit node could change. This effect could be seen on plots (unfortunately, I could only post them here after the weekend, I left them at work).
For min RTT, this means that initially our total min RTT is sum of RTT to the exit node + sum of RTTs of initial circuit. But when the circuit behind the exit node changes, it's new RTT could be much higher that from the previous circuit. And for the leecher and the seeder it would still be the same session - but the min RTT would still be pushed down by history of sending packets through the previous circuit.
And it could be too small, at least in theory.
In practice, it does not matter, because real-life Tribler connections are congested far before reaching even 500KBytes window.
Maybe when we fix our packet forwarding and change LEDBAT to something more suitable for long fat links (like FastTCP), we would require window size of more than 1MByte.

arvidn:
I see. thanks for the explanation! I'm inclined to not include the "dynamic buffer capacity tuning" aspect of this patch for now then, as it seems a bit safer.

src/utp_stream.cpp
```diff
@@ -281,7 +278,7 @@ struct utp_socket_impl
 		, m_written(0)
 		, m_receive_buffer_size(0)
 		, m_read_buffer_size(0)
-		, m_in_buf_size(1024 * 1024)
+		, m_receive_buffer_capacity(8 * 1024 * 1024)
```
arvidn: did you mean to set this lower?

src/utp_stream.cpp
```diff
@@ -1813,22 +1800,38 @@ bool utp_socket_impl::send_pkt(int const flags)
...
+	boost::uint32_t const close_reason = m_close_reason;
+
+
+	int sack = 0;
+	if (m_inbuf.size())
+	{
+        const int max_sack_size = effective_mtu 
+            - sizeof(utp_header)
+            - 2 // for sack padding/header
+            - (close_reason ? 6 : 0);
```
arvidn: the indentation isn't quite right here

arvidn:
looks good (apart from 2 comments)

ichorid:
Oops, sorry for that one. Had to move/rebase a lot of stuff, waited to see when all the checks are done.
Fixed the buffers size back to 1 MByte, fixed the indentation.

src/utp_stream.cpp
```diff
@@ -1813,22 +1800,38 @@ bool utp_socket_impl::send_pkt(int const flags)
...
 	int payload_size = (std::min)(m_write_buffer_size
 		, effective_mtu - header_size);
 	TORRENT_ASSERT(payload_size >= 0);
 
 	// if we have one MSS worth of data, make sure it fits in our
 	// congestion window and the advertised receive window from
 	// the other end.
-	if (m_bytes_in_flight + payload_size > (std::min)(int(m_cwnd >> 16)
-		, int(m_adv_wnd - m_bytes_in_flight)))
+	if (m_bytes_in_flight + payload_size > (std::min)(int(m_cwnd >> 16) , int(m_adv_wnd)))
```
arvidn: this formatting change seems a bit unnecessary. This doesn't push the line past 80 columns?
If you think this change should be made, please remove the space in front of the last comma

src/utp_stream.cpp
```diff
@@ -1813,22 +1800,38 @@ bool utp_socket_impl::send_pkt(int const flags)
 		&& m_write_buffer_size >= m_mtu_floor * 3
 		&& m_seq_nr != 0
 		&& (m_cwnd >> 16) > m_mtu_floor * 3);
+	// for non MTU-probes, use the conservative packet size
+	int const effective_mtu = mtu_probe ? m_mtu : m_mtu_floor;
+
+	boost::uint32_t const close_reason = m_close_reason;
+
+
```
arvidn: would you mind removing this blank line?

arvidn:
meged here https://github.com/arvidn/libtorrent/pull/2917

PR #2819, opened 2018-03-04
minimal optimization by using node_id::is_all_zero

aldenml:


src/kademlia/node.cpp
```diff
@@ -85,7 +85,7 @@ node_id calculate_node_id(node_id const& nid, aux::listen_socket_handle const& s
 		return generate_random_id();
 	}
 
-	if (nid == node_id::min() || !verify_id(nid, external_address))
+	if (nid.is_all_zeros() || !verify_id(nid, external_address))
```
arvidn: thinking a bit about this, I think it's a bit nicer to use ``operator==``. It could still be nicer than on the left hand side I think, like: ``nid == node_id{}``

arvidn: or, is this actually faster? I suppose the default constructor isn't ``constexpr``, but it still seems a bit odd that the compiler wouldn't see that all the writes in ``clear()`` are dead.

aldenml: interesting...

Issue #2825, opened 2018-03-07
Libtorrent is not compiling with older android sdk version


alexandre-janniaux:
**libtorrent version (or branch):**

Branch master, commit f98a7f3cf9698de10b721f49d3cf93b9771ef78d

**platform/architecture:**

Android sdk 15 / armv7

**compiler and compiler version:**

arm-linux-androideabi-gcc, generated with sdk version 15

**issue:**

Compiling libtorrent for android doesn't work with old sdk version, although it can. The problem is the config.hpp assume that it should use netlink and cpuid on Android whatever the sdk version is. However, NETLINK_ENOBUFS is not available until sdk21.

Moreover, ``sys/auxv.h`` doesn't exist, so I had to disable TORRENT_HAS_AUXV too but I didn't find if it has been available since a more recent sdk version yet. 

Is it possible to test the android sdk version so as to disable these features if the sdk is too old ? I can manage some time to write the patch if it's ok.

aldenml:
In order to compile `master` you need a toolchain with support for c++11. You will need to migrate to a clang based NDK, there are still some quirks with API 16, but easy to solve

alexandre-janniaux:
I'm sorry, actually it is arm-linux-androideabi-clang, both were available and I got confused. I'm able to build libtorrent with the fix I ask and make it works on android devices. The toolchain is the one generated by google's make_toolchain script with ``--api android-15``.

I don't use API 16 because it defines ``tdestroy`` but doesn't declare it, so I have to patch the configuration file to make it works.

aldenml:
well, never heard of `tdestroy` before :) how is that it works with API 15 and not API 16? (strange). Also, make sure you are using the latest NDK

In order to make it work with API 16, I only need to do this hack:
```
echo "unsigned long int getauxval(unsigned long int type) __attribute__((weak));" >> $PWD/android-toolchain/sysroot/usr/include/sys/auxv.h;
...and...
sed -i 's/using ::fgetpos;//g' $PWD/android-toolchain/include/c++/4.9.x/cstdio;
sed -i 's/using ::fsetpos;//g' $PWD/android-toolchain/include/c++/4.9.x/cstdio;
```

alexandre-janniaux:
[...]

PR #2828, opened 2018-03-08
fix compilation with android SDK < 21

alexandre-janniaux:
As explained in #2825, ``sys/auxv.h`` and ``NETLINK_ENOBUFS`` are not available until API 21. They were added quite late in the SDK, see for example this patch:
https://android.googlesource.com/platform/bionic/+/2c5153b043b44e9935a334ae9b2d5a4bc5258b40

This PR adds checks so that it enables netlink and auxv only if ``ANDROID_API`` is greater or equal than 21.

Maybe the part about defining ``TORRENT_HAS_AUXV`` should be moved to the ``include/libtorrent/config.hpp`` file. If so, I can make the change.

### DIFF
--- a/src/cpuid.cpp
+++ b/src/cpuid.cpp
@@ -48,7 +48,8 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #if defined __GLIBC__ && (__GLIBC__ > 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 16))
 #define TORRENT_HAS_AUXV 1
-#elif defined TORRENT_ANDROID
+#elif defined __ANDROID_API__ && __ANDROID_API__ >= 21
+// sys/auxv.h has been added in sdk 21
 #define TORRENT_HAS_AUXV 1
 #else
 #define TORRENT_HAS_AUXV 0
--- a/include/libtorrent/config.hpp
+++ b/include/libtorrent/config.hpp
@@ -193,11 +193,17 @@ POSSIBILITY OF SUCH DAMAGE.
 #endif
 
 #define TORRENT_HAVE_MMAP 1
-#define TORRENT_USE_NETLINK 1
 #define TORRENT_USE_IFADDRS 0
 #define TORRENT_USE_IFCONF 1
 #define TORRENT_HAS_SALEN 0
 #define TORRENT_USE_FDATASYNC 1
+
+// Android doesn't have NETLINK_ENOBUFS until API 21
+#if !defined __ANDROID_API__ || __ANDROID_API__ >= 21
+#define TORRENT_USE_NETLINK 1
+#else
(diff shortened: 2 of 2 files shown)


aldenml:
wait, both features work perfectly in android with API 16 both at toolchain level and real android devices. The problem is that the NDK was not very "stable" and "correct" until much later API. Things have improved with latest NDK and Unified Headers.

alexandre-janniaux:
Yes, at least for netlinks. But headers won't define NETLINK_ENOBUFS. This patch is exactly to fix what unified header are fixing. Maybe there is a better way to fix this ?  I though defining NETLINK_ENOBUFS and adding sys/auxv.h ourselves wasn't very clean.

aldenml:
There is a precedent in which @arvidn and @ssiloti accepted one macro definition, give it a try: https://github.com/arvidn/libtorrent/pull/2475#discussion-diff-146615119R67

as for `auxv`, I rather use my hack (minimal in the toolchain), the function is there for ages, and not having it prevents the use of useful optimizations in android

alexandre-janniaux:
I'll write another PR to write a better patch for the sys/auxv.h problem and another one for the NETLINK_NO_ENOBUFS define. 

But when you say that it works with SDK >= 16, did you try enough devices ?

aldenml:
a fair amount of them, and `NETLINK_NO_ENOBUFS` is pretty basic stuff in linux kernel, so it's a rare possibility. But it's buggy on some cases. Do you have a device API >= 16 that fails to have them?

alexandre-janniaux:
>  Do you have a device API >= 16 that fails to have them? 

Well, no, but I can't really afford to enable netlink if it fails on a few devices. I'll check for some clean workaround.

src/cpuid.cpp
```diff
@@ -48,7 +48,8 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #if defined __GLIBC__ && (__GLIBC__ > 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 16))
 #define TORRENT_HAS_AUXV 1
-#elif defined TORRENT_ANDROID
+#elif defined __ANDROID_API__ && __ANDROID_API__ >= 21
+// sys/auxv.h has been added in sdk 21
```
arvidn: with the other patch of loading getauxv() dynamically, this is redundant, right?

arvidn:
is the main reason to disable netlink that the android SDK *headers* are missing the ``NETLINK_NO_ENOBUFS`` define? This is part of the kernel interface, so it seems safe to just define it in that case, no?
In either case, if the kernel doesn't support this option, the socket option will fail and we'll ignore the error.

alexandre-janniaux:
Done, this PR is replaced by #2830 and #2831.

PR #2830, opened 2018-03-09
dynamically load getauxval so as to support older android devices

alexandre-janniaux:
Dynamically load getauxval just like android cpufeatures module does. It allows cpuid module to work on android even on older android sdk versions for which the ``sys/auxv.h`` header is not present.

See #2825

### DIFF
--- a/src/cpuid.cpp
+++ b/src/cpuid.cpp
@@ -55,8 +55,21 @@ POSSIBILITY OF SUCH DAMAGE.
 #endif
 
 
-#if TORRENT_HAS_ARM && TORRENT_HAS_AUXV
+#if defined(TORRENT_ANDROID) && TORRENT_HAS_AUXV
+#include <dlfcn.h>
+namespace {
+unsigned long int helper_getauxval(unsigned long int type)
+{
+    using getauxval_t = unsigned long int(*)(unsigned long int);
+    getauxval_t pf_getauxval = reinterpret_cast<getauxval_t>(dlsym(RTLD_DEFAULT, "getauxval"));
+    if (pf_getauxval == nullptr)
+        return 0;
+    return pf_getauxval(type);
+}
+}
+#elif TORRENT_HAS_ARM && TORRENT_HAS_AUXV
 #include <sys/auxv.h>
+#define helper_getauxval getauxval
 #endif
 
 namespace libtorrent { namespace aux {
@@ -107,7 +120,7 @@ namespace libtorrent { namespace aux {
 #if TORRENT_HAS_ARM_NEON && TORRENT_HAS_AUXV
 #if defined __arm__
 		//return (getauxval(AT_HWCAP) & HWCAP_NEON);
-		return (getauxval(16) & (1 << 12));
+		return (helper_getauxval(16) & (1 << 12));
 #elif defined __aarch64__
 		//return (getauxval(AT_HWCAP) & HWCAP_ASIMD);
 		//return (getauxval(16) & (1 << 1));
@@ -126,10 +139,10 @@ namespace libtorrent { namespace aux {
 		return true;
 #elif defined __arm__
 		//return (getauxval(AT_HWCAP2) & HWCAP2_CRC32);
-		return (getauxval(26) & (1 << 4));
+		return (helper_getauxval(26) & (1 << 4));
 #elif defined __aarch64__
 		//return (getauxval(AT_HWCAP) & HWCAP_CRC32);
-		return (getauxval(16) & (1 << 7));
+		return (helper_getauxval(16) & (1 << 7));
 #endif
 #else
 		return false;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -619,6 +619,10 @@ if (WIN32)
 	endif()
 endif()
 
+if(ANDROID)
+    target_link_libraries(torrent-rasterbar dl)
+endif()
+
 if(APPLE)
 	# for ip_notifier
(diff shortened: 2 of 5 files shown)


src/cpuid.cpp
```diff
@@ -56,7 +56,15 @@ POSSIBILITY OF SUCH DAMAGE.
 
 
 #if TORRENT_HAS_ARM && TORRENT_HAS_AUXV
-#include <sys/auxv.h>
+#include <dlfcn.h>
+static unsigned long int _getauxval(unsigned long int type)
```
arvidn: names starting with an underscore in the global namespace are reserved. please choose a better name

arvidn: in fact, it may be enough to replace the ``static`` with putting this function in the anonymous namespace

src/cpuid.cpp
```diff
@@ -56,7 +56,15 @@ POSSIBILITY OF SUCH DAMAGE.
...
+{
+    using getauxval_t = unsigned long int(unsigned long int);
+    getauxval_t* getauxval = (getauxval_t*) dlsym(RTLD_DEFAULT, "getauxval");
```
arvidn: please use ``static_cast<>`` instead of a C-style cast

arvidn:
please note: https://travis-ci.org/arvidn/libtorrent/jobs/351222528#L1461
`src/cpuid.cpp:63: undefined reference to 'dlsym'`

src/cpuid.cpp
```diff
@@ -56,7 +56,15 @@ POSSIBILITY OF SUCH DAMAGE.
 
 
 #if TORRENT_HAS_ARM && TORRENT_HAS_AUXV
```
arvidn: in your other patch, you define ``TORRENT_HAS_AUXV`` conditionally based on the target version. With this patch, could that define be removed entirely?

arvidn: I bring it up because it seems these two patches can't be combined cleanly right now, because if you built against an older SDK ``TORRENT_HAS_AUXV`` won't be defined, and you won't try to load the function dynamically.

src/cpuid.cpp
alexandre-janniaux: No problem, do you still want to change the name ?

src/cpuid.cpp
alexandre-janniaux: No, I should have specified it. The two last patch I made are aimed at replacing the first patch, so as to be able to compile without disabling the netlink and cpuid features. I'll clean both patch and check again.

src/cpuid.cpp
alexandre-janniaux: do you mean ``reintrepret_cast<>`` ?

alexandre-janniaux:
I still have to update the three build systems (if I understood correctly) for the ``-ldlsym`` issue.

src/cpuid.cpp
arvidn: ``dlsym()`` returns a ``void*``, I'm pretty sure ``static_cast<>`` suffice.

arvidn:
please only add a dependency on the dynamic linker for android builds (it's called ``libdl``, right?). (in the Jamfile you can use the ``<target-os>andoid`` condition, iirc)

src/cpuid.cpp
alexandre-janniaux: With the following code:
```cpp
getauxval_t p_getauxval = static_cast<getauxval_t> (dlsym(RTLD_DEFAULT, "getauxval"));
```

I stumble across the following error 
```
cpuid.cpp:64:31: error: static_cast from 'void *' to 'getauxval_t' (aka 'unsigned long (*)(unsigned long)') is not allowed
```

According to https://linux.die.net/man/3/dlsym, conversion from ``void*`` to pointer-to-function is an undefined behaviour. In fact, ``void*`` wasn't even meant to store pointer-to-function in ANSI C, but is instead a pointer-to-object storage. It seems to be the same in the C++ standard, but we can stick to POSIX for Android and assume that reinterpret_cast-ing it in a pointer-to-function is ok.

alexandre-janniaux:
> (it's called libdl, right?)

Oops, yes it is, sorry for mistyping.

Thank you for the indication.

src/cpuid.cpp
arvidn: ah. In that case, yes, I meant ``reinterpret_cast<>`` :)

aldenml:
not a fan of that new `libdl`dependency, but I like this, it will remove one hack in my build

src/cpuid.cpp
```diff
@@ -55,8 +55,21 @@ POSSIBILITY OF SUCH DAMAGE.
...
+    using getauxval_t = unsigned long int(*)(unsigned long int);
+    getauxval_t p_getauxval = reinterpret_cast<getauxval_t>(dlsym(RTLD_DEFAULT, "getauxval"));
+    if (p_getauxval == nullptr)
+        return 0;
+    return p_getauxval(type);
+}
+}
+#elif TORRENT_HAS_ARM && TORRENT_HAS_AUXV
 #include <sys/auxv.h>
+#define _getauxval getauxval
```
arvidn: now, if you introduce a define, you're definitely in trouble using a reserved word. I think it's probably best to use a name that doesn't start with an underscore.

arvidn: in fact, even in the anonymous namespace, it would overload with the global namespace, so it seems questionable even without the define

alexandre-janniaux: You're definitely right, I will rename it ``helper_getauxval``.

alexandre-janniaux: fixed

PR #2831, opened 2018-03-09
Add missing define for old android sdk

alexandre-janniaux:
On old android SDK, before unified header, NETLINK_NO_ENOBUFS and IFA_D_DADFAILED don't exist, but seems to be supported. Define them with their known values in that case.

See #2825

### DIFF
--- a/include/libtorrent/socket.hpp
+++ b/include/libtorrent/socket.hpp
@@ -68,6 +68,11 @@ POSSIBILITY OF SUCH DAMAGE.
 #ifndef SOL_NETLINK
 #define SOL_NETLINK 270
 #endif
+
+// NETLINK_NO_ENOBUFS exists at least since android 2.3, but is not exposed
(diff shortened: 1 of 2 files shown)


arvidn:
looks good. could you add a condition to only define these for android f they're missing?
The constant is only known for that specific kernel.

alexandre-janniaux:
What is your conventions for ``#if defined`` ? Do you prefer ``#if defined(value)`` or ``#if defined value`` ?

arvidn:
oh, I've tried to stick to ``#if defined macro``, i.e. no parenthesis

alexandre-janniaux:
Fixed !

aldenml:
👍

alexandre-janniaux:
I made a mistake in the ``TORRENT_ANDROID`` check, it should have been ``#if defined TORRENT_ANDROID`` instead of ``#if TORRENT_ANDROID``.

PR #2834, opened 2018-03-10
add windows socket libraries in Makefile.am

### LINKED ISSUE #2833: Some libraries are missing in the autotools scripts for windows
**libtorrent version (or branch):**

master

**issue:**

It seems that the autotools scripts are missing the following link to libraries:
```
 -liphlpapi -lws2_32 -lwsock32
```

They are correctly linked in the CMake and Jamfile scripts.

alexandre-janniaux:
Fix link issues when cross compiling for windows with mingw.

I don't know if autotools' scripts are also expected to work with MSVC, but as they are linked in the other build scripts, I think it's ok.

Fixes #2833

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -104,6 +104,13 @@ AS_IF([test "$SYS" = linux],[
 ])
 AM_CONDITIONAL(HAVE_ANDROID, test "${HAVE_ANDROID}" = "1")
 
+HAVE_WINDOWS=0
+case "${host_os}" in
+    *mingw32* | *cygwin*)
+        HAVE_WINDOWS=1
+        ;;
+esac
+AM_CONDITIONAL(HAVE_WINDOWS, test "${HAVE_WINDOWS}" = "1")
 
 ###############################################################################
 # Checking for needed base libraries
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -171,15 +171,21 @@ libtorrent_rasterbar_la_SOURCES = \
   \
   $(BUILTIN_CRYPTO_SOURCES)
 
+AM_CFLAGS = -I$(top_srcdir)/ed25519/src -std=c99
+AM_CPPFLAGS = -DTORRENT_BUILDING_LIBRARY -I$(top_srcdir)/include -I$(top_srcdir)/ed25519/src @DEBUGFLAGS@ @OPENSSL_INCLUDES@
+AM_LDFLAGS = @OPENSSL_LDFLAGS@
+
 libtorrent_rasterbar_la_LDFLAGS = -version-info $(INTERFACE_VERSION_INFO)
 libtorrent_rasterbar_la_LIBADD = @OPENSSL_LIBS@
+libtorrent_rasterbar_la_CPPFLAGS = $(AM_CPPFLAGS)
 
 if HAVE_ANDROID
 libtorrent_rasterbar_la_LIBADD += -ldl
 endif
 
-AM_CPPFLAGS = -DTORRENT_BUILDING_LIBRARY -I$(top_srcdir)/include -I$(top_srcdir)/ed25519/src @DEBUGFLAGS@ @OPENSSL_INCLUDES@
-AM_CFLAGS = -I$(top_srcdir)/ed25519/src -std=c99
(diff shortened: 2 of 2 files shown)


alexandre-janniaux:
Actually, it seems that this PR is still incomplete, there is no HAVE_WINDOWS

configure.ac
```diff
@@ -104,6 +104,13 @@ AS_IF([test "$SYS" = linux],[
 ])
 AM_CONDITIONAL(HAVE_ANDROID, test "${HAVE_ANDROID}" = "1")
 
+HAVE_WINDOWS=0
+case "${host_os}" in
```
arvidn: perhaps this is a bit of a nit-pick, but this looks like it may break cross-compilation. i.e. the *host OS* may be different from the *target OS*. But I'm also quite ignorant of how autotools work, so maybe there's no hope to support cross compilation anyway.

alexandre-janniaux: Actually, I'm trying to push this PR so as to cross compile for windows. ${host_os} references the system on which the executable will run, in contrast with ${build_os} which is the system on which it is build. I also checked before right here, because it was confusing: https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Specifying-Target-Triplets.html

But you may be right in the case of a Canadian cross, and I know absolutely nothing about it. If you think it should be supported, I'll do some research and fix this.

Currently, I don't differentiate WIN32 from WIN64 too, I don't know if you want to as nothing depends on this in libtorrent.

alexandre-janniaux: Also, the build with MinGW currently needs morganz/mingw-std-threads header files because there is no C++ thread feature support in the mingw compiler with win32 thread model. I tried to alias boost thread types but I can't alias std::chrono types as they are directly used in the code (and already defined). Are you using features that std has but boost hasn't ?

arvidn: I don't really have a strong opinion, if it's fine with you you can leave it as is.

I actually build mingw on appveyor without any additional stuff (as far as I know). But I only got it to build with mingw64.

alexandre-janniaux: Would you accept a patch to use boost types (aliased in a header) if C++ thread library is not available and if it is possible to write ?

About this PR, I hope I'll manage to finish it until the end of the week. It doesn't seems correct to me yet.

alexandre-janniaux:
Ok, after fighting with autotools and a misunderstanding of the manual, it seems better.
Maybe I can add a mingw64 test build with link tests in travis ? 

I'll check one last time that the link is ok on Windows and it will be ready.

alexandre-janniaux:
LGTM. There is no ``TORRENT_WINDOWS`` exported to the source code but as it is not useful for you I believe it's ok.

configure.ac
arvidn: I would be a bit hesitant to fake C++11 headers. Are you suggesting they are not available for mingw? I believe there would be a somewhat high maintenance burden to keep it up. I believe boost threads isn't quite compatible with std threads (but I could be wrong, I haven't looked in a long time). the ``thread`` class itself is probably the simplest part to support, but ``atomic<>``, ``mutex<>``, ``lock_guard<>``, ``unique_lock<>``, ``this_thread::yield()`` and ``this_thread::slee_for()`` may be trickier. And to put them in the ``std`` namespace is technically not allowed.

arvidn:
please feel free to add a cross compilation test to travis, that would be the best way of ensuring nobody breaks this in the future.

arvidn:
``TORRENT_WINDOWS`` is defined in ``libtorrent/config.hpp`` anyway, it doesn't rely on a configure pass

configure.ac
alexandre-janniaux: The C++11 thread library as a whole is not available for mingw64 with win32 thread model. There are header-only implementations of these (meganz/mingw-std-threads) but you have to include and distribute them with the library so I think it's also a no-go. 

I don't find any other clean solution so I believe it's better if I only keep a local patch for these headers.

alexandre-janniaux:
Oops, I must have tested my patch with the ``HAVE_WINDOWS`` at the top of config.hpp for testing purpose, and didn't notice it. Thank you !

Issue #2724, opened 2018-01-23
fails to pin to the working tracker within a tier


sledgehammer999:
Some users are reporting that tracker announcements with correct upload/download stats aren't happening correctly.
Libtorrent version used is 1.1.5+gitd52763805c

Relevant qbt bug: https://github.com/qbittorrent/qBittorrent/issues/8108
The first post has some pictures showing the discrepancy between client and tracker stats. It is quite big.
Others report that correct stats are reported after torrent completion and during seeding, but the uploaded stats that happened during download aren't reported and are "lost".

arvidn:
there are simulation tests that assert that announces happen at the right times. I will look into those to make sure they are correct and would cover this case, but I feel pretty confident they do.

It would be really helpful if someone experiencing this problem could post a log exposing this issue 
(that means announce logs with the relevant query parameters and torrent state transitions)

ignitionnight:
@arvidn I emailed you with the an attached log, the email subject is "issue #2724 log"

I downloaded a torrent from private tracker broadcasthe.net and seeded for about two and a half hours. qbittorrent reports 39.5 MiB uploaded, but the tracker only reports 11.1 MB. This experience is consistent with the issues that I originally reported in the qbittorrent issue tracker https://github.com/qbittorrent/qBittorrent/issues/8108.

I downloaded a torrent almost immediately after it was loaded to the site. There were already 30ish seeds before I started leeching, but I was seeding in about a minute. After the first update on the tracker it showed only 3 MB of upload despite the qbittorrent reporting ~30 MiB uploaded. 

I tested this against utorrent 2.2.1 and under similar circumstances the upload credit was 100% correct, any additional upload I contributed after an update on the tracker was reflected on the site within 30 minutes following the next update. 

Thank you, and let me know if you need anything else.

arvidn:
just to rule things out; are you confident that the number qBittorrent told you was correct?
There are more than one way to report uploads. Perhaps qBittorrent shows you *total* uploaded bytes, whereas what's reported to the tracker is only *payload* upload. i.e. all the protocol overhead for downloading does not count from the tracker's point of view.

greenbench:
Is it possible that the difference between total uploaded bytes and only payload upload can be hundreds of megabytes - almost gigabyte? Because that's the difference i'm getting with popular torrents. Then again, before qB 4.0.x that has never happened before. All upload was 100% matched.

arvidn:
[...]

PR #2837, opened 2018-03-12
fix issue where tracker would be skipped for the next tracker in the tier

arvidn:
This is believed to fix https://github.com/arvidn/libtorrent/issues/2724


arvidn:
since I haven't heard anything on this, I will assume this fixes it and merge.

PR #2842, opened 2018-03-13
Fix resuming partially checked torrent

d-komarov:
Torrent can't resume downloading when file priorities prevent the full
check. If file priorities are changed and torrent has unverified piece
hashes, then it will stuck in downloading state. The only way to handle
this is to perform a `force_recheck`.

* Add .editorconfig file to enforce code style
* Fix TORRENT_TEST(checking) <=> TORRENT_TEST(read_only)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+root = true
+
+[*]
+indent_size = 4
+indent_style = tab
+insert_final_newline = true
+trim_trailing_whitespace = true
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -8646,6 +8646,10 @@ namespace {
 		// to be in downloading state (which it will be set to shortly)
 //		INVARIANT_CHECK;
 
+		// workaround for torrent getting stuck in `downloading` state
+		if (!are_files_checked())
+			set_state(torrent_status::checking_files);
+
 		if (m_state == torrent_status::checking_resume_data
 			|| m_state == torrent_status::checking_files
 			|| m_state == torrent_status::allocating)
--- a/test/test_checking.cpp
+++ b/test/test_checking.cpp
@@ -76,7 +76,7 @@ enum
 	force_recheck = 8,
 };
 
-void test_checking(int flags = 0)
+void test_checking(int flags)
 {
 	using namespace libtorrent;
 	namespace lt = libtorrent;
@@ -272,7 +272,7 @@ void test_checking(int flags = 0)
 
 TORRENT_TEST(checking)
 {
-	test_checking();
+	test_checking(0);
 }
 
 TORRENT_TEST(read_only_corrupt)
--- a/test/test_checking.cpp
+++ b/test/test_checking.cpp
@@ -40,10 +40,43 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/torrent_info.hpp"
 #include "libtorrent/torrent_status.hpp"
 
-static const int file_sizes[] =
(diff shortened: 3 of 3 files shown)


d-komarov:
I'll fix debug configuration test failures tomorrow. Adding wait timeout argument to `wait_for_alert` routine should do the job.

src/torrent.cpp
```diff
@@ -8646,6 +8646,10 @@ namespace {
 		// to be in downloading state (which it will be set to shortly)
 //		INVARIANT_CHECK;
 
+		// workaround for torrent getting stuck in `downloading` state
+		if (!are_files_checked())
```
arvidn: This change is not intuitive to me. I wonder, if checking torrents always ignored the file- and piece priorities, would that also fix it?

test/test_checking.cpp
```diff
@@ -40,10 +40,43 @@ POSSIBILITY OF SUCH DAMAGE.
...
+	bool is_checking(int const state)
+	{
+		return state == lt::torrent_status::checking_files
+#ifndef TORRENT_NO_DEPRECATE
+			|| state == lt::torrent_status::queued_for_checking
+#endif
+			|| state == lt::torrent_status::checking_resume_data;
+	};
+
+	lt::settings_pack test_settings()
```
arvidn: there's already a function called ``settings()`` in ``test/settings.hpp`` that does essentially the same thing. would it work to use that instead?

test/test_checking.cpp
```diff
@@ -63,7 +96,7 @@ enum
 	force_recheck = 8,
 };
 
-void test_checking(int flags = read_only_files)
+void test_checking(int flags = 0)
```
arvidn: I don't see this change affecting anything here. is it left-over from a previous iteration?

test/test_checking.cpp
d-komarov: I think so.

test/test_checking.cpp
d-komarov: Well, default value seemed incorrect to me. I was pretty sure that "checking" and "read_only" tests must be different. Before this change they were doing exactly the same.

d-komarov: TORRENT_TEST(checking)
{
	test_checking();
}

TORRENT_TEST(read_only)
{
	test_checking(read_only_files);
}

```test_checking() == test_checking(read_only_files)```

Comment (arvidn) [reply test/test_checking.cpp]: ah, I see. maybe the default value should be removed even, if it's only relied on in one place

Comment (d-komarov) [reply test/test_checking.cpp]: I like the idea to remove default argument from ‘test_checking’.

File: .editorconfig [hunk 4]
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*]
+indent_size = 4
+indent_style = tab
+insert_final_newline = true
+trim_trailing_whitespace = true
```
arvidn: what do you think about moving this to a separate pull request?
I think it should be a little bit more sophisticated:

* limit this style to [*.{cpp,hpp,c,h}]
* don't specify an ``indent_size``
* add style for [*.py] to use spaces and match the indent size that's used there
* maybe add a style to .travis and appveyor.yml as well

arvidn:
I misunderstood the problem here initially. The issue is changing file- or piece priority while checking, breaks the checking. I will submit a pull request against this pull request with suggested changes

arvidn:
since I haven't heard anything, I did this. I will probably merge it this weekend unless you have any comments: https://github.com/arvidn/libtorrent/pull/2869

arvidn:
merged https://github.com/arvidn/libtorrent/pull/2869 instead

PR #2846, opened 2018-03-14
CIS-2081 - skip symlinks initialization (fix a bug when a size of 0 is set on symlink erasing a target).

vkhorozvs:


arvidn:
Do you have a more detailed description of this problem? Did it turn out to not be a problem at all?

vkhorozvs:
A problem is quite simple:

1. Create a file with some content e.g. /tmp/myfile.txt
2. Create a symlink to that file (in our case it's /tmp/myfolder/mylink.txt -> ../myfile.txt).
3. Create a torrent with a symlink (in our case we added in entire myfolder.
4. Start seeding on this host.
5. Start download from another host.

A problem is: /tmp/myfile.txt gets a size of 0 on seeder node i.e. all its content is effectively deleted.
This is constantly reproduced on 1.1.6 and there was no problem in 1.1.1.
So, we've created this patch in our private repo.

There is a bigger problem: libtorrent doesn't seem to replicate symlinks at all - it just creates empty files instead, but we've solved it with our own code.

PR #2864, opened 2018-03-21
cmake: restore change to make test_common a propper library

ssiloti:
This was originally done in 87832ce706506b5b568e50e52751a6bd0cbb4f9e but
it got clobbered by a merge from RC_1_1.


arvidn:
sorry, that was my bad

PR #2865, opened 2018-03-21
some code refactor and cleanup in broadcast_socket

aldenml:


include/libtorrent/broadcast_socket.hpp
```diff
@@ -79,10 +79,10 @@ namespace libtorrent {
 
 		struct socket_entry
 		{
-			explicit socket_entry(std::shared_ptr<udp::socket> const& s)
-				: socket(s), broadcast(false) { std::memset(buffer, 0, sizeof(buffer)); }
-			socket_entry(std::shared_ptr<udp::socket> const& s
-				, address_v4 const& mask): socket(s), netmask(mask), broadcast(false)
+			explicit socket_entry(std::shared_ptr<udp::socket> s)
```
arvidn: the places where ``socket_entry`` is constructed, do we already move the shared_ptr into it?

aldenml: not possible, because it's used after it:
```cpp
m_sockets.emplace_back(s);
socket_entry& se = m_sockets.back();
ADD_OUTSTANDING_ASYNC("broadcast_socket::on_receive");
s->async_receive_from(boost::asio:....
```

PR #2866, opened 2018-03-21
more use of move and code refactor in bandwidth related code

aldenml:


aldenml:
do you think it makes sense to add the default move constructors to `bw_request`?

arvidn:
is ``bw_request`` not movable already?

PR #2869, opened 2018-03-22
Changing file priorities while checking interrupts checking.

arvidn:
Also, if the last file has zero priority, checking is interrupted prematurely.


arvidn:
@d-komarov I took your patch, made some changes, rebased and squashed it here

PR #2871, opened 2018-03-23
using move with bw_request and make sure it is movable

aldenml:


aldenml:
not sure if you prefer the asserts or adding the explicit default ones.

PR #2875, opened 2018-03-24
don't perform DNS lookups for the DHT bootstrap unless DHT is enabled

arvidn:


Chocobo1:
Pardon... the review was sent to fast, now I observed qbt won't quit after shutdown: keeping the CPU busy, currently not sure whether is related to this PR.

arvidn:
it hangs the tests though, because of the additional lookup I think

Chocobo1:
>currently not sure whether is related to this PR.

Now I confirm it is caused by this PR.

arvidn:
Ok, I just updated the patch (force pushed). I believe it should work now. at least the tests no longer hang

aldenml:
does this change the logic in `master`? there, there is some logic in place to account for the fact that `update_dht_bootstrap_nodes` is called before `update_dht` due to the settings order.

arvidn:
I haven't looked at master yet. but the test results suggests that this would need to be a bit more sophisticated. perhaps having a test makes sense after all, and deferring it to master

aldenml:
yes please, take a look at it, maybe the way it's now is wrong, but with this change, I feel the logic of operations is somehow inverted and not necessarily better

Chocobo1:
>I believe it should work now. at least the tests no longer hang

Still no good, it keeps CPU busy all the time.

arvidn:
@aldenml as far as I can tell, master has the same problem. The code looks pretty similar between master and RC_1_1, I believe your fix was back-ported to RC_1_1.

arvidn:
@Chocobo1 I feel pretty confident with the patch as it is now. Would you mind testing it again?

aldenml:
Ok, now it looks like this is fixing something :)

Chocobo1:
>Ok, now it looks like this is fixing something :)

Indeed, just tested, now it works as expected.

arvidn:
Do we think this is reliable enough to make it into 1.1.7?
specifically, enabling and disabling the DHT back and forth would be the risky scenario I think

aldenml:
I can't see any obvious logic flaw but didn't verify in practice. I can give it a try to add more related unit tests once it's merge in `master`

PR #2877, opened 2018-03-25
path error in python_binding docs

David-Luis:


arvidn:
is this not a typo in RC_1_1 too?

arvidn:
would you mind making the pull request against RC_1_1 instead?

arvidn:
oh, I see

PR #2880, opened 2018-03-26
support building python bindings for python 3 and 64 bit

ssiloti:
Don't merge yet, just testing on appveyor


arvidn:
is this different than what would be merged in from RC_1_1 now?

ssiloti:
No, other than changes to get it to apply cleanly on master this is the same at what went into RC_1_1.

PR #2884, opened 2018-03-27
separate cmake and autotools travis build

arvidn:


aldenml:
at first, I didn't want to add another job, but now that cmake is another one, I think I could add more tasks to it

PR #2886, opened 2018-03-27
Add alert when piece passes hash check and is written to disk

fernando-rodriguez:
Add a new alert, piece_ready_alert, that is generated when a piece has
passed the hash check and has been written to disk. This alert may be
generated more than once for some pieces.

### DIFF
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -2516,9 +2516,9 @@ namespace libtorrent
 		piece_ready_alert(aux::stack_allocator& alloc, torrent_handle h
 			, int piece_num);
 
-		TORRENT_DEFINE_ALERT(piece_ready_alert, 90)
+		TORRENT_DEFINE_ALERT_PRIO(piece_ready_alert, 90)
 
-		static const int static_category = alert::progress_notification;
+		static const int static_category = alert::piece_progress_notification;
 		virtual std::string message() const TORRENT_OVERRIDE;
 
 		int piece_index;
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -2507,12 +2507,30 @@ namespace libtorrent
 #endif // TORRENT_DISABLE_LOGGING
 	};
 
+	// This alert is generated when a piece has passed the hash check and has
+	// been written to disk (so it can be safely read). This alert may be generated
+	// more than once for each piece.
+	struct TORRENT_EXPORT piece_ready_alert TORRENT_FINAL : torrent_alert
+	{
+		// internal
+		piece_ready_alert(aux::stack_allocator& alloc, torrent_handle h
+			, int piece_num);
+
+		TORRENT_DEFINE_ALERT(piece_ready_alert, 90)
+
+		static const int static_category = alert::progress_notification;
+		virtual std::string message() const TORRENT_OVERRIDE;
+
+		int piece_index;
+	};
+
+
 #undef TORRENT_DEFINE_ALERT_IMPL
 #undef TORRENT_DEFINE_ALERT
 #undef TORRENT_DEFINE_ALERT_PRIO
 #undef TORRENT_CLONE
 
-	enum { num_alert_types = 90 }; // this enum represents "max_alert_index" + 1
+	enum { num_alert_types = 91 }; // this enum represents "max_alert_index" + 1
 }
 
 
--- a/include/libtorrent/piece_picker.hpp
+++ b/include/libtorrent/piece_picker.hpp
@@ -385,7 +385,7 @@ namespace libtorrent
 		// or if we have the piece
 		bool is_piece_finished(int index) const;
 
-		// returns true if all blocks in this piece are finished
+		// returns true if the piece has passed the hash check
 		// and have been written to disk
 		bool is_piece_ready(int index) const;
 
--- a/include/libtorrent/piece_picker.hpp
+++ b/include/libtorrent/piece_picker.hpp
@@ -385,6 +385,10 @@ namespace libtorrent
 		// or if we have the piece
 		bool is_piece_finished(int index) const;
 
+		// returns true if all blocks in this piece are finished
+		// and have been written to disk
+		bool is_piece_ready(int index) const;
+
 		// returns true if we have the piece or if the piece
 		// has passed the hash check
 		bool has_piece_passed(int index) const;
--- a/src/peer_connection.cpp
+++ b/src/peer_connection.cpp
@@ -3064,6 +3064,15 @@ namespace libtorrent
 //			, peer_info_struct(), block_finished.piece_index, block_finished.block_index);
 		picker.mark_as_finished(block_finished, peer_info_struct());
 
+		if (picker.is_piece_ready(block_finished.piece_index))
+		{
+			if (t->alerts().should_post<piece_ready_alert>())
+			{
+				t->alerts().emplace_alert<piece_ready_alert>(t->get_handle(),
+					int(block_finished.piece_index));
+			}
+		}
+
 		t->maybe_done_flushing();
 
 		if (t->alerts().should_post<block_finished_alert>())
--- a/src/peer_connection.cpp
+++ b/src/peer_connection.cpp
@@ -3064,9 +3064,9 @@ namespace libtorrent
 //			, peer_info_struct(), block_finished.piece_index, block_finished.block_index);
 		picker.mark_as_finished(block_finished, peer_info_struct());
 
-		if (picker.is_piece_ready(block_finished.piece_index))
+		if (t->alerts().should_post<piece_ready_alert>())
 		{
-			if (t->alerts().should_post<piece_ready_alert>())
+			if (picker.is_piece_ready(block_finished.piece_index))
 			{
 				t->alerts().emplace_alert<piece_ready_alert>(t->get_handle(),
 					int(block_finished.piece_index));
--- a/src/piece_picker.cpp
+++ b/src/piece_picker.cpp
@@ -2923,18 +2923,24 @@ get_out:
 
 	bool piece_picker::is_piece_ready(int index) const
 	{
-		if (!is_piece_finished(index))
-			return false;
+		TORRENT_ASSERT(index < int(m_piece_map.size()));
+		TORRENT_ASSERT(index >= 0);
 
 		piece_pos const& p = m_piece_map[index];
 		int state = p.download_queue();
 
-		if (state < piece_pos::num_download_categories)
+		if (p.index != piece_pos::we_have_index && state == piece_pos::piece_open)
+		{
+			for (int i = 0; i < piece_pos::num_download_categories; ++i)
+				TORRENT_ASSERT(find_dl_piece(i, index) == m_downloads[i].end());
+			return false;
+		}
+		else if (state < piece_pos::num_download_categories)
 		{
 			std::vector<downloading_piece>::const_iterator i =
 				find_dl_piece(state, index);
 			TORRENT_ASSERT(i != m_downloads[state].end());
-			if (int(i->finished) < blocks_in_piece(index))
+			if (!i->passed_hash_check || int(i->finished) < blocks_in_piece(index))
 				return false;
 		}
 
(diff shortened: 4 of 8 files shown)


src/piece_picker.cpp
```diff
@@ -2921,6 +2921,26 @@ namespace libtorrent
 		return true;
 	}
 
+	bool piece_picker::is_piece_ready(int index) const
```
arvidn: I think this function would be simpler (and more efficient) by not calling ``is_piece_finished`` here, but to embed just the interesting case directly in this function. You will likely save a few lookups

include/libtorrent/piece_picker.hpp
```diff
@@ -385,6 +385,10 @@ namespace libtorrent
 		// or if we have the piece
 		bool is_piece_finished(int index) const;
 
+		// returns true if all blocks in this piece are finished
+		// and have been written to disk
```
arvidn: did you mean to say "and the piece has passed the hash-check"?
if all blocks are ``finished``, that means the piece has been written to disk already, right? otherwise it would be in ``writing``, no?

ssiloti: The finished state does *not* imply the piece has been verified. Hash checking and writing to disk are done asynchronously so to make this work there need to be checks in both the write complete and piece verified paths.

ssiloti: Strike that last sentence, I missed that there is already a check in the verified path.

arvidn: I was also under the impression that the ``state_finished`` block state did not imply having passed the hash check. The ``downloading_piece`` has a ``passed_hash_check`` member, which I would expect would have to checked. But I haven't looked very deep

include/libtorrent/alert_types.hpp
```diff
@@ -2507,12 +2507,30 @@ namespace libtorrent
...
+	// more than once for each piece.
+	struct TORRENT_EXPORT piece_ready_alert TORRENT_FINAL : torrent_alert
+	{
+		// internal
+		piece_ready_alert(aux::stack_allocator& alloc, torrent_handle h
+			, int piece_num);
+
+		TORRENT_DEFINE_ALERT(piece_ready_alert, 90)
+
+		static const int static_category = alert::progress_notification;
```
arvidn: I would suggest inventing a new alert category, say, ``alert::piece_progress_notification`` and maybe raise the priority of this alert as well (as one may imagine it's important if this category is enabled)

src/peer_connection.cpp
```diff
@@ -3064,6 +3064,15 @@ namespace libtorrent
 //			, peer_info_struct(), block_finished.piece_index, block_finished.block_index);
 		picker.mark_as_finished(block_finished, peer_info_struct());
 
+		if (picker.is_piece_ready(block_finished.piece_index))
+		{
+			if (t->alerts().should_post<piece_ready_alert>())
```
arvidn: would it make sense to check ``should_post<>`` first, before ``is_piece_ready()``? The former should be a lot cheaper

src/piece_picker.cpp
```diff
@@ -2921,6 +2921,26 @@ namespace libtorrent
...
+		piece_pos const& p = m_piece_map[index];
+		int state = p.download_queue();
+
+		if (state < piece_pos::num_download_categories)
+		{
+			std::vector<downloading_piece>::const_iterator i =
+				find_dl_piece(state, index);
+			TORRENT_ASSERT(i != m_downloads[state].end());
+			if (int(i->finished) < blocks_in_piece(index))
+				return false;
```
arvidn: I think you need to checkt ``i->passed_hash_check`` here too. But I could be wrong. If I am, maybe there should be a comment here explaining why

src/torrent.cpp
```diff
@@ -4382,6 +4392,14 @@ namespace {
 		if (m_ses.alerts().should_post<piece_finished_alert>())
 			m_ses.alerts().emplace_alert<piece_finished_alert>(get_handle(), index);
 
+		if (has_picker() && picker().is_piece_ready(index))
+		{
+			if (alerts().should_post<piece_ready_alert>())
```
arvidn: here too, I think it may make sense to check ``should_post<>`` first.

arvidn:
``test_piece_picker.cpp`` should be amended to test the new function as well.

include/libtorrent/piece_picker.hpp
fernando-rodriguez: I'm posting the alert right after piece_finished_alert, that alert implies that the piece has passed the hash check. And I *think* piece_picker::is_piece_finished() also implies that even though is not documented as such, but I'm not 100% sure. So I don't think I need to check passed_hash_check right now but I may have to after I implement your previous suggestion.

src/piece_picker.cpp
fernando-rodriguez: Will do. Thanks

include/libtorrent/alert_types.hpp
fernando-rodriguez: Will do.

src/peer_connection.cpp
fernando-rodriguez: Yes, that's what I had done initially and changed it alleviate the contention problem over the mutex. I will change it back as we found other ways to deal with that and it probably makes little difference.

include/libtorrent/piece_picker.hpp
arvidn: I see. although, if a piece being "ready" means that it's written to disk and has passed the hash-check, it would be an error prone interface of the piece picker to say a piece is ready even though it hasn't passed the check. i.e. I think piece_picker should have a sensible interface, and not rely on specific external logic to behave well

fernando-rodriguez: It's not. Right now calling is_piece_finished() guarantees that the piece has passed the hash check AFAIU. It would've been clearer if I had called has_piece_passed(). It will be much clearer after I implement your previous advice of removing the is_piece_finished() call as then I will probably have to check passed_hash_check too.

fernando-rodriguez: Actually, you're absolutely right, I get it now. There is a change that is_piece_ready() will return a false positive. It just will never happen when I'm calling it. I fixed it already, will test later and push.

src/piece_picker.cpp
```diff
@@ -2923,18 +2923,24 @@ namespace libtorrent
...
+		TORRENT_ASSERT(index >= 0);
 
 		piece_pos const& p = m_piece_map[index];
 		int state = p.download_queue();
 
-		if (state < piece_pos::num_download_categories)
+		if (p.index != piece_pos::we_have_index && state == piece_pos::piece_open)
+		{
+			for (int i = 0; i < piece_pos::num_download_categories; ++i)
+				TORRENT_ASSERT(find_dl_piece(i, index) == m_downloads[i].end());
```
arvidn: I think it always feels scary to have an assert as the only statement in a loop that doesn't have braces. Imagine if this assert statement would simply disappear in release builds, then the return statement would end up in the loop.

I think you could probably just remove the loop and the assert, to keep in simple.

include/libtorrent/alert_types.hpp
```diff
@@ -2507,12 +2507,30 @@ namespace libtorrent
 #endif // TORRENT_DISABLE_LOGGING
 	};
 
+	// This alert is generated when a piece has passed the hash check and has
+	// been written to disk (so it can be safely read). This alert may be generated
+	// more than once for each piece.
```
arvidn: why may it be generated more than once per piece? is there anything we can do to avoid that?

arvidn:
I'm a bit worried about introducing a new alert in the stable branch, because of ABI compatibility. I haven't thought it all the way through, of which kinds of errors can happen, but ``alert_cast<>`` may have issues for instance.

include/libtorrent/alert_types.hpp
fernando-rodriguez: I don't think it will now that I'm getting a better picture of how libtorrent works so I'll change that.

src/torrent.cpp
```diff
@@ -4392,9 +4392,9 @@ namespace {
 		if (m_ses.alerts().should_post<piece_finished_alert>())
 			m_ses.alerts().emplace_alert<piece_finished_alert>(get_handle(), index);
 
-		if (has_picker() && picker().is_piece_ready(index))
+		if (alerts().should_post<piece_ready_alert>())
 		{
-			if (alerts().should_post<piece_ready_alert>())
+			if (has_picker() && picker().is_piece_ready(index))
```
fernando-rodriguez: I'm not sure under what circumstances we won't have a picker here. You think it would make sense to call need_picker() to ensure the alert is never missed?

Likewise on the `on_write_completed` callbacks would bail out before sending the alert if there's no picker. `on_piece_verified` calls `need_picker()`

fernando-rodriguez:
Anything I can do to help get it in sooner just let me know. Should I send a PR for master after the review changes are done?

src/torrent.cpp
arvidn: when the torrent becomes a seed, the picker is deleted (as there is no need to keep the additional state associated with picking pieces anymore).

If a picker is expected to always exist through some code path, it probably makes sense to add an assert to ensure that assumption holds

fernando-rodriguez:
Thanks for all the help. I think this alert may not be needed after all. I'm testing the following: Instead of keeping just an integer flag for each piece I keep a `blocks` counter, a `check_passed` bit flag, and a `piece_ready` bit. On `block_finished_alert` I increment the block counter, on `hash_failed_alert` I reset it to 0, and on `piece_finished_alert` I set `check_passed`. On both `block_finished_alert` and `piece_finished_alert` after updating the counters I check if `check_passed && blocks == blocks_in_piece` and if it is set `piece_ready` and notify any waiting threads. To check if a piece is ready I just check the piece_ready bit after that. It seems to work so far. Do you see any problems with that approach?

It generates a lot more alerts but as long as I can continue to pop the alerts on the notify function that won't be a problem. Also I may need to take special care to update the counters  when resuming a download. I haven't implemented that yet as I'm mostly just streaming for now.

arvidn:
in the case a piece fails the hash check, you should probably clear the "block-finished" bits. I think there's a risk you'll get the piece_passed alert before all blocks have hit disk, and if you've already marked all blocks as having been written (from the first attempt) you may be too early.

fernando-rodriguez:
> in the case a piece fails the hash check, you should probably clear the "block-finished" bits. I think there's a risk you'll get the piece_passed alert before all blocks have hit disk, and if you've already marked all blocks as having been written (from the first attempt) you may be too early.

I'm not keeping a bit for each block, just a 15-bit counter (I chose that size because it's what the picker uses and I got bits to spare), and I clear it on `hash_failed_alert`. Do you see a need to track each block?

The only thing I don't like is that currently I need to do a lookup to find the torrent state for each alert and now there's a lot more alerts. As a C guy my solution would be to add an opaque field (ie. `user_data`) to the torrent handle so I can store a pointer to the state there. Perhaps there's a better way in C++ or in libtorrent's API already? Any suggestions?

fernando-rodriguez:
Well this blew up last night after tunning some settings. It looks like after `hash_failed_alert` is posted the piece is locked until the network thread is in-sync with the IO thread. But before the sync completes the `block_finished_alert` for the failed pieces are still posted. So clearing the counter on `hash_failed_alert` is not enough to synchronize with libtorrent.

I worked around it by tracking the number of pieces left to be written when I receive the `hash_failed_alert` and adding that to the number of expected pieces,  so a piece is finished writting when we have received `blocks_in_piece + no_of_blocks_remaining_when_hash_failed`. This sort of works but it's broken because if a block write fails for any reason I'm fried.

I can think of two possible solutions that would be faster than a new `piece_ready_alert` since they don't require the lookup for each block:

1) Delay `hash_failed_alert` until after the sync with the IO thread is done. It would still be within specs (documentation) so I doubt it'll break anything, what are you thoughts?

2) Add a new alert, `piece_sync_complete`, that is posted after the sync completes. That way I can ignore the blocks finished while syncing.

arvidn:
(1) seems reasonable and probably yields more intuitive results

fernando-rodriguez:
Great. I'll close this one and submit a new pull request then.

PR #2887, opened 2018-03-27
removed unnecessary loop in request_a_block

aldenml:


aldenml:
This was introduced here: https://github.com/arvidn/libtorrent/commit/65fef972a285d8403e934f0b61a9218746b41561#diff-a63adc2c6def826dd1db59b48f67f5e2L206

arvidn:
ouch!

PR #2892, opened 2018-03-29
avoid calls to .address() when looking for endpoint protocol

aldenml:
I detected quite a few of these, but before going full in, I would like to have some opinions

### DIFF
--- a/include/libtorrent/broadcast_socket.hpp
+++ b/include/libtorrent/broadcast_socket.hpp
@@ -53,6 +53,18 @@ namespace libtorrent {
 	TORRENT_EXTRA_EXPORT bool is_teredo(address const& addr);
 	TORRENT_EXTRA_EXPORT bool is_ip_address(std::string const& host);
 
+	// TODO: refactor these out too
+	template <typename Endpoint>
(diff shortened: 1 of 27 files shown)


arvidn:
this looks like a reasonable approach.

aldenml:
do you think it's better to merge this and do others PR(s)?

arvidn:
what do you mean?

aldenml:
that if you prefer that I do the other changes(more use of these new functions) in a separate PR, or just amend this one?

arvidn:
I think amending this one makes sense

aldenml:
OK, will do that

include/libtorrent/broadcast_socket.hpp
```diff
@@ -53,6 +53,18 @@ namespace libtorrent {
 	TORRENT_EXTRA_EXPORT bool is_teredo(address const& addr);
 	TORRENT_EXTRA_EXPORT bool is_ip_address(std::string const& host);
 
+	// TODO: refactor these out too
```
arvidn: is this comment still relevant?

PR #2894, opened 2018-03-30
deprecated alert::progress_notification alert category

arvidn:
split into finer grained categories

### DIFF
--- a/test/test_alert_types.cpp
+++ b/test/test_alert_types.cpp
@@ -65,9 +65,16 @@ TORRENT_TEST(alerts_types)
 #else
 	++count_alert_types;
 #endif
+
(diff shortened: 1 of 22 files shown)


test/test_alert_types.cpp
```diff
@@ -65,9 +65,16 @@ TORRENT_TEST(alerts_types)
 #else
 	++count_alert_types;
 #endif
+
+#ifndef TORRENT_NO_DEPRECATE
+#define PROGRESS_NOTIFICATION alert::progress_notification |
+#else
+#define PROGRESS_NOTIFICATION
```
aldenml: maybe you should add `#undef PROGRESS_NOTIFICATION` after the tests (like `#undef TEST_ALERT_TYPE`)

PR #2899, opened 2018-04-01
Test merge

arvidn:


arvidn:
@aldenml @ssiloti I can't understand where this (consistent) call to a pure virtual function in the python binding comes from. I cannot reproduce it on my mac.

aldenml:
looked at it and I can't help :( not familiar enough with the python binding

ssiloti:
I am able to repro this on my linux machine. Here's a [backtrace](https://gist.github.com/ssiloti/bc357169136980a4c66bd977f99c8c82). This is with Boost 1.65.1.

arvidn:
thanks. I didn't even try it on my linux machine (since it happened on osx on travis). I will try that too

ssiloti:
4e7cb26 fixed the crash for me.

arvidn:
the remaining problem appears to be that dependency according to boost-build is boost_python-vc140-mt-1_63.dll whereas the actual dependency (as reported by dumpbin) is boost_python3-vc140-mt-1_63.dll

(i.e. "python3" instead of "python"). I'll take a look at how RC_1_1 differs here

src/session_impl.cpp
```diff
@@ -5795,20 +5815,23 @@ namespace aux {
...
-			if (m_outstanding_router_lookups == 0) update_dht();
+			if (m_outstanding_router_lookups == 0) start_dht();
 			return;
 		}
 
 
 		for (auto const& addr : addresses)
 		{
+#if !TORRENT_USE_IPV6
+			if (!i->is_v4()) continue;
```
aldenml: where is this `i` defined?, could it be `addr` instead?

PR #2903, opened 2018-04-02
fix potential fd leak in enum_net_interfaces

arvidn:


src/enum_net.cpp
```diff
@@ -1132,6 +1132,7 @@ namespace libtorrent
 		if (s < 0)
 		{
 			ec = error_code(errno, system_category());
+			close(sock);
```
aldenml: why not call this right above `int s = socket(AF_INET, SOCK_DGRAM, 0);` and remove the `close` at the end?

PR #2911, opened 2018-04-04
make socket_entry buffer std::array

aldenml:


arvidn:
if you rebase, the mac build should be fixed

aldenml:
OK, done

PR #2915, opened 2018-04-07
Delay hash_failed_alert until on_piece_sync()

fernando-rodriguez:
Delay the hash_failed_alert until after synchronizing with the IO thread
so that it can be used to synchronize piece progress.

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -4525,9 +4525,6 @@ namespace {
 
 		inc_stats_counter(counters::num_piece_failed);
 
-		if (m_ses.alerts().should_post<hash_failed_alert>())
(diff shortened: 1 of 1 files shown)


src/torrent.cpp
```diff
@@ -4728,6 +4725,9 @@ namespace {
 		// ever downloaded for it.
 		m_picker->restore_piece(j->piece);
 
+		if (m_ses.alerts().should_post<hash_failed_alert>())
```
arvidn: did you make sure ``on_piece_sync`` isn't used as a callback in any other situation? like where there was no piece failure?

fernando-rodriguez: Yes.

PR #2916, opened 2018-04-07
Make extension alerts reliable

fernando-rodriguez:
Add feature to make alerts sent to session extensions reliable (ie. the
alerts will be delivered even if the queue is full). Also make should_post
lockless.

### DIFF
--- a/include/libtorrent/alert_manager.hpp
+++ b/include/libtorrent/alert_manager.hpp
@@ -191,7 +191,7 @@ namespace libtorrent {
 		boost::function<void()> m_notify;
 
 		// the number of resume data alerts  in the alert queue
-		boost::atomic<int> m_num_queued_resume;
+		int m_num_queued_resume;
 
 		// this is either 0 or 1, it indicates which m_alerts and m_allocations
 		// the alert_manager is allowed to use right now. This is swapped when
--- a/include/libtorrent/alert_manager.hpp
+++ b/include/libtorrent/alert_manager.hpp
@@ -114,7 +114,7 @@ namespace libtorrent {
 			T alert(m_allocations[m_generation], std::forward<Args>(args)...);
 			m_alerts[m_generation].push_back(alert);
 
-			maybe_notify(&alert, lock);
+			maybe_notify(&alert);
 		}
 
 #else
@@ -167,7 +167,7 @@ namespace libtorrent {
 		alert_manager(alert_manager const&);
 		alert_manager& operator=(alert_manager const&);
 
-		void maybe_notify(alert* a, mutex::scoped_lock& lock);
+		void maybe_notify(alert* const a);
 
 		mutable mutex m_mutex;
 		condition_variable m_condition;
--- a/include/libtorrent/alert_manager.hpp
+++ b/include/libtorrent/alert_manager.hpp
@@ -46,6 +46,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #endif
 #include <boost/function/function0.hpp>
 #include <boost/shared_ptr.hpp>
+#include <boost/atomic.hpp>
 #include <boost/config.hpp>
 #include <list>
 #include <utility> // for std::forward
@@ -97,23 +98,15 @@ namespace libtorrent {
 				* (1 + T::priority))
 			{
 #ifndef TORRENT_DISABLE_EXTENSIONS
+				lock.unlock();
+
 				if (m_ses_extensions_reliable.empty())
 					return;
 
-				// after we have a reference to the current allocator it
-				// is safe to unlock the mutex because m_allocations is protected
-				// by the fact that the client needs to pop alerts *twice* before
-				// it can free it and that's impossible until we emplace
-				// more alerts.
-				aux::stack_allocator& alloc = m_allocations[m_generation];
-				lock.unlock();
-
-				// save the state of the active allocator so that
-				// we can restore it when we're done
-				aux::stack_allocator_state_t state = alloc.save_state();
-				T alert(alloc, std::forward<Args>(args)...);
+				mutex::scoped_lock lock(m_mutex_reliable);
+				T alert(m_allocator_reliable, std::forward<Args>(args)...);
 				notify_extensions(&alert, m_ses_extensions_reliable);
-				alloc.restore_state(state);
+				m_allocator_reliable.reset();
 #endif
 				return;
 			}
@@ -219,6 +212,8 @@ namespace libtorrent {
 		void notify_extensions(alert * const a, ses_extension_list_t& extensions);
 		ses_extension_list_t m_ses_extensions;
 		ses_extension_list_t m_ses_extensions_reliable;
+		aux::stack_allocator m_allocator_reliable;
+		mutex m_mutex_reliable;
 #endif
 	};
 }
--- a/include/libtorrent/alert_manager.hpp
+++ b/include/libtorrent/alert_manager.hpp
@@ -52,6 +52,11 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "libtorrent/aux_/disable_warnings_pop.hpp"
 
+#ifndef TORRENT_DISABLE_EXTENSIONS
+#include "libtorrent/extensions.hpp"
+#endif
+
+
 #ifdef __GNUC__
 // this is to suppress the warnings for using std::auto_ptr
 #pragma GCC diagnostic push
@@ -94,7 +99,36 @@ namespace libtorrent {
 			// for high priority alerts, double the upper limit
 			if (m_alerts[m_generation].size() >= m_queue_size_limit
 				* (1 + T::priority))
+			{
+#ifndef TORRENT_DISABLE_EXTENSIONS
+				if (m_reliable_ext_alerts) {
+					// after we have a reference to the current allocator it
+					// is safe to unlock the mutex because m_allocations is protected
+					// by the fact that the client needs to pop alerts *twice* before
+					// it can free it and that's impossible until we emplace
+					// more alerts.
+					aux::stack_allocator& alloc = m_allocations[m_generation];
+					lock.unlock();
+
+					// save the state of the active allocator so that
+					// we can restore it when we're done
+					aux::stack_allocator_state_t state = alloc.save_state();
+					T alert(alloc, std::forward<Args>(args)...);
+
+					for (ses_extension_list_t::iterator i = m_ses_extensions.begin()
+						, end(m_ses_extensions.end()); i != end; ++i)
+					{
+						if (((*i)->implemented_features() & plugin::reliable_alerts_feature) != 0)
+						{
+							(*i)->on_alert(&alert);
+						}
+					}
+
+					alloc.restore_state(state);
+				}
+#endif
 				return;
+			}
 
 			T alert(m_allocations[m_generation], std::forward<Args>(args)...);
 			m_alerts[m_generation].push_back(alert);
@@ -116,12 +150,6 @@ namespace libtorrent {
 		template <class T>
 		bool should_post() const
 		{
-			mutex::scoped_lock lock(m_mutex);
-			if (m_alerts[m_generation].size() >= m_queue_size_limit
-				* (1 + T::priority))
-			{
-				return false;
-			}
 			return (m_alert_mask & T::static_category) != 0;
 		}
 
@@ -129,13 +157,11 @@ namespace libtorrent {
 
 		void set_alert_mask(boost::uint32_t m)
 		{
-			mutex::scoped_lock lock(m_mutex);
 			m_alert_mask = m;
 		}
 
 		boost::uint32_t alert_mask() const
 		{
-			mutex::scoped_lock lock(m_mutex);
 			return m_alert_mask;
 		}
 
@@ -203,6 +229,7 @@ namespace libtorrent {
 #ifndef TORRENT_DISABLE_EXTENSIONS
 		typedef std::list<boost::shared_ptr<plugin> > ses_extension_list_t;
 		ses_extension_list_t m_ses_extensions;
+		bool m_reliable_ext_alerts;
 #endif
 	};
 }
--- a/include/libtorrent/alert_manager.hpp
+++ b/include/libtorrent/alert_manager.hpp
@@ -103,7 +103,7 @@ namespace libtorrent {
 				if (m_ses_extensions_reliable.empty())
 					return;
 
-				mutex::scoped_lock lock(m_mutex_reliable);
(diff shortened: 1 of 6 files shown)


fernando-rodriguez:
I'm not sure why [this](https://github.com/arvidn/libtorrent/pull/2916/filesl) fails and I can't build locally with C++98 ATM. The on_alert() callback is not being called after the queue fills up so there must be something wrong with the implemented_features() checks or with the plugin on the test.

arvidn:
test_alert_manager.cpp is failing

fernando-rodriguez:
I know, the test I added is failing. I can't figure out why. It only happens for the c++98 build. Can you spot anything wrong with the two checks in alert_manager.hpp:121 or alert_manager.cpp:170 ... or with the changes I made to the plugin on test_alert_manager.cpp?

fernando-rodriguez:
I protected the new allocator with it's own mutex so that part is thread-safe. But `maybe_notify()` is still not thread-safe because the mutex is unlocked before the callbacks. The simplest solution is to extend the lock through the callbacks but there's a comment on line 155 saying never to do that. I'm not sure why. Anything else I can think of turns out very complicated.

include/libtorrent/alert_manager.hpp
```diff
@@ -181,7 +188,7 @@ namespace libtorrent {
 		boost::function<void()> m_notify;
 
 		// the number of resume data alerts  in the alert queue
-		int m_num_queued_resume;
+		boost::atomic<int> m_num_queued_resume;
```
arvidn: I think you meant to change ``m_alert_mask``, not ``m_num_queued_resume``, right?

arvidn: or, I suppose both

include/libtorrent/alert_manager.hpp
```diff
@@ -202,7 +209,11 @@ namespace libtorrent {
 
 #ifndef TORRENT_DISABLE_EXTENSIONS
 		typedef std::list<boost::shared_ptr<plugin> > ses_extension_list_t;
+		void notify_extensions(alert * const a, ses_extension_list_t& extensions);
```
arvidn: the ``extensions`` parameter is not mutated by this call, right? It should be a ``const&``

src/alert_manager.cpp
```diff
@@ -162,9 +158,25 @@ namespace libtorrent
...
 	void alert_manager::add_extension(boost::shared_ptr<plugin> ext)
 	{
-		m_ses_extensions.push_back(ext);
+		if ((ext->implemented_features() & plugin::reliable_alerts_feature) != 0)
+		{
+			m_ses_extensions_reliable.push_back(ext);
+		}
+		else
+		{
+			m_ses_extensions.push_back(ext);
```
arvidn: you still need to add the alert to ``m_ses_extensions`` unconditionally

src/alert_manager.cpp
```diff
@@ -97,11 +96,8 @@ namespace libtorrent
 		}
 
 #ifndef TORRENT_DISABLE_EXTENSIONS
-		for (ses_extension_list_t::iterator i = m_ses_extensions.begin()
-			, end(m_ses_extensions.end()); i != end; ++i)
-		{
-			(*i)->on_alert(a);
-		}
+		notify_extensions(a, m_ses_extensions);
+		notify_extensions(a, m_ses_extensions_reliable);
```
arvidn: oh, I see. you call it twice. I would prefer to have ``m_ses_extensions`` be a superset and have `notify_extensions()` just called once. I can't think of a good argument right now though

test/test_alert_manager.cpp
```diff
@@ -223,6 +255,42 @@ TORRENT_TEST(extensions)
...
+
+	TEST_EQUAL(plugin_alerts[0], 100);
+	TEST_EQUAL(plugin_alerts[1], 100);
+	TEST_EQUAL(plugin_alerts[2], 140);
+
+	bool running = false;
+	int num_resume = 0;
+	std::vector<alert*> alerts;
+	mgr.get_all(alerts, num_resume);
+	libtorrent::thread t(boost::bind(&alert_popper, boost::ref(mgr), boost::ref(running)));
```
arvidn: this is a race condition. ``running`` would need to be an atomic

test/test_alert_manager.cpp
```diff
@@ -223,6 +255,42 @@ TORRENT_TEST(extensions)
...
+
+	/* make sure the thread is started */
+	while (!running) test_sleep(10);
```
arvidn: but it seems that involving a thread that pops here is unnecessary. you don't need to pop the alerts. In fact, by doing so you risk not exercising the interesting case (where the queue is full)

test/test_alert_manager.cpp
```diff
@@ -223,6 +255,42 @@ TORRENT_TEST(extensions)
...
+
+	for (int i = 0; i < 100000000; ++i)
```
arvidn: since this literal is greater than what ``int`` can represent, it may yield warnings, or possibly errors. Do you really need this many iterations?

test/test_alert_manager.cpp
```diff
@@ -223,6 +255,42 @@ TORRENT_TEST(extensions)
 	TEST_EQUAL(plugin_alerts[0], 70);
 	TEST_EQUAL(plugin_alerts[1], 70);
 	TEST_EQUAL(plugin_alerts[2], 70);
+
+	for (int i = 0; i < 35; ++i)
+		mgr.emplace_alert<torrent_finished_alert>(torrent_handle());
```
arvidn: I think it would make sense to split this up in multiple test cases. Right now there are just a bunch of different aspects tested  in one sequence, making it harder to run them in isolation.

fernando-rodriguez:
Do we need to support any targets smaller than 32-bit? If so I don't think m_queued_resume needs to be atomic either. My intention was to prevent 2 threads from incrementing at once but that part is protected by a mutex.

fernando-rodriguez:
I meant to say if *not* above.

arvidn:
When you say 32 but, are you referring to the address model or the width of int?
Either way i don’t see how that’s relevant

fernando-rodriguez:
I mean the CPU register and data bus size, that is the largest amount of memory that the CPU can move atomically. But that usually coincides with the addressing model (sizeof(int) is usually at most that big too). If it can be less than 32-bit then when incrementing `m_num_queued_resume` it would be read from the CPU in two 16-bits halves, incremented, and written back to memory in two 16-bits halves. So if we call `num_queued_resume()` after the first half has been written we would read the wrong value. Making it atomic protects against that. But if we only support 32-bit systems then it only needs to be atomic if we want to protect against two threads modifying it at once.

That is at least how I understand it. I may be wrong about some of it or there may be some architectures where it doesn't exactly apply.

arvidn:
I see what you mean. However, C++ is not concerned about any of those things. If one thread reads a (non-atomic) value and another thread writes it, without synchronisation, it's a race condition and you invoke undefined behaviour. These rules aren't just reflecting how hardware may work, it also reflects the assumptions compilers are allowed to make (about non-atomic) variable access.

fernando-rodriguez:
This is what I found on the standard (ss. 1.10.21):

> The execution of a program contains a data race if it contains two conflicting actions in different threads,
at least one of which is not atomic, and neither happens before the other. Any such data race results in
undefined behavior.

In this case the read operation would be atomic, so I don't think there is a race or undefined behaviour. I get the point about compiler variable access but I think the right tool for that is the volatile keyword as in my failed test with apple's compiler. I think in that case the compiler may still optimize the atomic type out unless the boost::atomic template does something to prevent it like making it volatile.

So I can make them both atomic to be safe, but I don't think it's necessary.

fernando-rodriguez:
I found out why that test had failed on Apple. There's nothing wrong with the code but clang's over-agressive warnings. You can reproduce the warning with this code snippet:
```
bool running = false;
bool& running2 = running;
for (running = true; running;)
{
        fprintf(stdout, "Hello\n");
        running2 = false;
}
```

arvidn:
you mean it would be atomic because you made `m_num_queued_resume` an atomic variable? Both the write and the read to that variable will be atomic, so that's not a problem.

I'm more concerned about `m_alert_mask`, which is not an atomic variable. No write nor read of that variable will be atomic, hence, if done from different threads will be a race condition. This has nothing to do with word-size or memory bus width. What counts as atomic operations are defined [here](http://eel.is/c++draft/atomics).

volatile is not related to synchronisation. It turns reads and writes to a variable into observable side-effects of the program (just like system calls are I/O). They are not allowed to be omitted, re-ordered (within a thread). You can still introduce race conditions by writing to a volatile variable from different threads though.

test/test_alert_manager.cpp
```diff
@@ -206,9 +210,9 @@ TORRENT_TEST(extensions)
 	memset(plugin_alerts, 0, sizeof(plugin_alerts));
 	alert_manager mgr(100, 0xffffffff);
 
-	mgr.add_extension(boost::make_shared<test_plugin>(0));
-	mgr.add_extension(boost::make_shared<test_plugin>(1));
-	mgr.add_extension(boost::make_shared<test_plugin>(2));
+	mgr.add_extension(boost::make_shared<test_plugin>(0, false));
+	mgr.add_extension(boost::make_shared<test_plugin>(1, false));
+	mgr.add_extension(boost::make_shared<test_plugin>(2, false));
```
arvidn: maybe `reliable_alerts` could have a default value of `false`

src/alert_manager.cpp
```diff
@@ -71,16 +70,14 @@ namespace libtorrent
...
-	void alert_manager::maybe_notify(alert* a, mutex::scoped_lock& lock)
+	void alert_manager::maybe_notify(alert* const a)
 	{
 		if (a->type() == save_resume_data_failed_alert::alert_type
 			|| a->type() == save_resume_data_alert::alert_type)
 			++m_num_queued_resume;
 
 		if (m_alerts[m_generation].size() == 1)
 		{
-			lock.unlock();
```
arvidn: the reason to unlock the mutex before calling the user function is to avoid a deadlock in case the user function makes a call into libtorrent that in turn tries to post an alert (or pop alerts).

I would expect a call to `pop_alerts()` would deadlock with this patch for instance, if not, do you understand why?

fernando-rodriguez:
I meant that both operations would be atomic on 32-bits, even if not declared as such. We're not concerned with the increment, that is done under lock, only with the write back to memory and our read. In pure C++ they cannot be considered atomic because standard C++ should have well defined behaviour on any architecture.

In the case of `m_alert_mask` in pure C++ there would be undefined behaviour but it would not be harmful to our program. For example, if one thread is updating it from 0x00000000 to 0xFFFFFFFF and we read it from another thread then it is undefined whether we will see 0x0000FFFF or 0xFFFF0000. But we don't care about that because we're checking single bits, only the integer representation would be incorrect. It won't happen on 32-bits though because store operation is atomic.

I meant using volatile to prevent the compiler from making assumptions about non-atomic types. AFAICT the atomic type just overrides all the operations with the built-in atomic C functions if supported or wraps them in a mutex if not supported. Does it gets any special treatment by the compiler?

Let's just make them both atomic to be safe though.

arvidn:
> I meant that both operations would be atomic on 32-bits, even if not declared as such. We're not concerned with the increment, that is done under lock, only with the write back to memory and our read. In pure C++ they cannot be considered atomic because standard C++ should have well defined behaviour on any architecture.

I think "pure C++" (i.e. the standard) is the only thing that matters.

> In the case of m_alert_mask in pure C++ there would be undefined behaviour but it would not be harmful to our program.

I think undefined behaviour is always harmful. Compilers are under no obligation to make *any* part of the program behave the way you want if it invokes undefined behaviour (including the parts that come "before" the UB).

> For example, if one thread is updating it from 0x00000000 to 0xFFFFFFFF and we read it from another thread then it is undefined whether we will see 0x0000FFFF or 0xFFFF0000. But we don't care about that because we're checking single bits, only the integer representation would be incorrect. It won't happen on 32-bits though because store operation is atomic.

Another example would be that: the compiler can tell that the only way a variable could change its value would introduce a race condition. Since that would be UB it then assumes it cannot and will not happen. It can then proceed to not generate any read instructions, since the value is already known, and we know it cannot change.

> I meant using volatile to prevent the compiler from making assumptions about non-atomic types. AFAICT the atomic type just overrides all the operations with the built-in atomic C functions if supported or wraps them in a mutex if not supported. Does it gets any special treatment by the compiler?

Exactly what's going on is an implementation detail. But, yes, compilers do quite a lot to optimise code, including atomics (but especially non-atomics, since they can be assumed to be "single-threaded"). for example, [N4455](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4455.html).

include/libtorrent/alert_manager.hpp
```diff
@@ -116,27 +131,19 @@ namespace libtorrent {
...
 		bool should_post() const
 		{
-			mutex::scoped_lock lock(m_mutex);
-			if (m_alerts[m_generation].size() >= m_queue_size_limit
-				* (1 + T::priority))
-			{
-				return false;
-			}
-			return (m_alert_mask & T::static_category) != 0;
+			return (m_alert_mask.load(boost::memory_order_relaxed) & T::static_category) != 0;
```
arvidn: I would not use a relaxed load here. if you change the alert mask, which alerts that are posted after that should be deterministic and reliable. a relaxed load disconnects it from any other memory updates.

Just `(m_alert_mask & T::static_category) != 0` should be fine.

src/alert_manager.cpp
```diff
@@ -52,8 +52,7 @@ namespace libtorrent
 
 	int alert_manager::num_queued_resume() const
 	{
-		mutex::scoped_lock lock(m_mutex);
-		return m_num_queued_resume;
+		return m_num_queued_resume.load(boost::memory_order_consume);
```
arvidn: I think `consume` here is clearly wrong. I think anything other than "sequential consistency" requires very strong arguments.

src/alert_manager.cpp
```diff
@@ -75,7 +74,7 @@ namespace libtorrent
 	{
 		if (a->type() == save_resume_data_failed_alert::alert_type
 			|| a->type() == save_resume_data_alert::alert_type)
-			++m_num_queued_resume;
+			m_num_queued_resume.fetch_add(1, boost::memory_order_relaxed);
```
arvidn: I don't understand why you would use relaxed here. This counter is definitely tied to the actual alerts stored in the queue, it's not independent. This counter better be consistent with the queue itself, by making it relaxed it won't synchronise *any* other memory. Again, I think anything other than "sequential consistency" requires strong arguments and good rationale.

PR #2918, opened 2018-04-07
remove the global cache of the current time

arvidn:
just use clock::now() instead


arvidn:
I thought I had removed this already. I don't think there's really any point

PR #2921, opened 2018-04-08
Logo design

baranpirincal:
Hello!

I uploaded .png format of my logo design with different colors.

Best Regards,
Baran Pirincal
Visual Communication Designer


aldenml:
Questions:
- Is this `readme.rst` really necessary?
- There is an space after the dash in each image file name, intentionally?
- Since these are pure media files, do you think it makes sense to explicitly put a license to them?
- Do you think it's also good to include the vectorial file in this PR?

baranpirincal:
- I deleted readme.txt it was unnecessary
- There was a mistake in names of files, corrected.
- I removed vectoral files on pull-request and shared only with project owner.

Sorry for that problems, I am new on Github...

arvidn:
@aldenml I like to keep all source in the repository (in fact, I believe there is a policy for BSD source packages to include all source, including documentation source files).

I can't find the commit with the .ai file in it, is it unreasonably large for the repository or impractical for some other reason?

@baranpirincal what would you think about putting the logo file in the docs folder? Then it can be integrated with the documentation and website that's built from the rst files and templates in there (but integration should probably be a separate commit as that may require some tweaking of style-sheets and color palette of the site in general, to go together).

What do you think of putting the attribution of the design in the ``AUTHORS`` file as well as the "Acknowledgements" section in ``docs/index.rst``?

baranpirincal:
@arvidn It is okay for me, I'm talking about your both question. Also, if you want me to do something, I can

aldenml:
To clarify :) I was asking precisely for the inclusion of the vectorial file. I believe it should be included (only 700KB).

 Now, as for the license, source code is easier because of the header and common conventions, but for media files, I suggest adding a more explicit one next to them (I like Creative Commons).

baranpirincal:
@aldenml @arvidn 

I uploaded vectoral files and created license of logo design.

arvidn:
would you mind putting the ``logo`` folder under ``docs``? and I think the pdf version is probably unnecessary.

I also had some thoughts.

Did you experiment with a monochrome version? I imagine perhaps the one of the colours would just become transparent, and the shadow would be dropped.
I'm also curious about what it would look like to flatten it. i.e. remove the shadow and maybe the shadow along the rim. In fact, it might actually work to remove the middle "band" along the left and bottom side to just be transparent too (which would unify a monochrome and colour version).

If you don't mind I might experiment a bit. I'll probably use the svg version as source though, as I don't have adobe illustrator.

baranpirincal:
@arvidn I will add monochrome version of my logo design here to show as soon as possible.
Also, I will add the logo folder under docs and delete pdf version if it is unnecessary.

baranpirincal:
@arvidn here is the monochrome versions

libtorrent monochrome

baranpirincal:
@arvidn Hello again, I put "logo" folder under "docs" and removed .pdf format. Also, I added monochrome versions. Could you please check it?

PR #2931, opened 2018-04-12
pass default constructed boost::function instead of NULL

ssiloti:
Clang 6 defines `NULL` as `nullptr` which breaks implicit construction
of `boost::function` from `NULL`.


bentley:
Works for me. Thanks!

jbeich:
FWIW, `NULL` is defined by OS vendors, not Clang e.g.,
https://github.com/freebsd/freebsd/commit/c8ed04c26b6758354853a6bed4629f71d0d01a7d
https://github.com/openbsd/src/commit/6ecde746dea9a5d17abf3bafa06c232b9189b33b

PR #2938, opened 2018-04-13
Improve CMake build

evsh:
This PR proposes two improvements:
1. Specify public dependencies explicitly and use them to generate a proper pkg-config file. The list of public libraries is taken from pkg-config (produced by the autotools build).
2. Extends cmake into bindings/python subdirectory. The extension module is built by CMake itself, without setuptools, thus avoiding passing flags and options from one build system to another.  Setuptools are used to generate egg-info through.
3. Provide build-time dependencies in the generated `libtorrent-rasterbarConfig.cmake` file.


evsh:
Thanks, corrected (threads and openssl libraries are part of the interface now).

bindings/python/CMakeLists.txt
```diff
@@ -0,0 +1,85 @@
+find_package(PythonInterp REQUIRED)
+find_package(PythonLibs REQUIRED)
+
+if (NOT boost-python)
+	set(_boost_python "python-${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}")
```
ssiloti: I've never seen the Boost Python libraries follow this naming convention. By default they are named `boost_python` for Python 2 and `boost_python3` for Python 3. The default should be to follow upstream's naming.

CMakeLists.txt
```diff
@@ -583,7 +586,7 @@ if (encryption)
 		FIND_PACKAGE(OpenSSL REQUIRED)
 	endif()
 	target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_OPENSSL TORRENT_USE_LIBCRYPTO)
-	target_include_directories(torrent-rasterbar PUBLIC ${OPENSSL_INCLUDE_DIR})
+	target_include_directories(torrent-rasterbar SYSTEM PUBLIC ${OPENSSL_INCLUDE_DIR})
```
ssiloti: This is causing gcc to use `-isystem` to specify the include dir which is messing up the include search order and causing build failures for me. I don't think we should be using `SYSTEM` here, nor for the boost include dir below.

evsh: > …and causing build failures for me.

How that's possible?

ssiloti: The exact error is:

> In file included from /usr/x86_64-pc-linux-gnu/include/boost/config/platform/linux.hpp:15:0,
>                  from /usr/x86_64-pc-linux-gnu/include/boost/config.hpp:57,
>                  from /home/steve/projects/libtorrent/include/libtorrent/config.hpp:40,
>                  from /home/steve/projects/libtorrent/src/web_connection_base.cpp:33:
> /usr/x86_64-pc-linux-gnu/include/c++/6.4.0/cstdlib:75:25: fatal error: stdlib.h: No such file or directory
>  #include_next <stdlib.h>

The GNU C/C++ headers are very sensitive to the system include dir search order so it's best not to mess with it.

evsh: Thanks. So, that is the old #include_next problem... Probably pkg-config for OpenSSL on your system lists /usr/include?

ssiloti: Yep. To be exact it's `/usr/x86_64-pc-linux-gnu/include` due to multiarch.

evsh: So neither shall we put standard include directories in pkg-config (PR updated).

evsh: > Yep. To be exact it's /usr/x86_64-pc-linux-gnu/include due to multiarch.

Aha! And could you test, please, does CMAKE_C_IMPLICIT_INCLUDE_DIRECTORIES contain that path?

bindings/python/CMakeLists.txt
evsh: Done

bindings/python/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
+find_package(PythonInterp REQUIRED)
+find_package(PythonLibs REQUIRED)
+
+if (NOT boost-python)
+	# use active python
+	set(_boost_python "python${PYTHON_VERSION_MAJOR}")
```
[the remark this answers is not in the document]
evsh: I don't know what to do. I'm on Gentoo and here we have python3.6m and boost_python-3.6, so the naming is up to package maintainers. I have no preference for the default name, the error message from CMake is pretty clear when the name is incorrect.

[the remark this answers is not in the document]
ssiloti: There needs to be a special case for when `${PYTHON_VERSION_MAJOR} == 2`.

CMakeLists.txt
evsh: I wanted to silence warnings from the boost headers, but if there are still distros with errors in their package configurations, it's not a big deal to revert that.

bindings/python/CMakeLists.txt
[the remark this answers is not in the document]
evsh: I disagree. It seems simpler for me to pass -Dboost-python option to CMake than complicate the code here.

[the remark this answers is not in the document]
ssiloti: Simpler for us sure, but not for people trying to build libtorrent. It'd be pretty silly to require explicitly setting boost python's library name even with a default boost install.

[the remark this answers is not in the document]
evsh: But python2? Isn't it too old to provide a special handling for it? Perhaps there are more systems with non-default library names (like Gentoo) than those interested in bindings for python 2 only? Because those who need bindings for 3 and 2 have to read this fragment and the shorter it is the better.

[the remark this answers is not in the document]
evsh: From the other hand, the current default does not make sense at all, I could just write "python3" instead...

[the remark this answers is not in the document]
evsh: Arch:
```
libboost_python.a
libboost_python3.a
```
Fedora:
```
libboost_python.so.1.66.0
libboost_python3.so.1.66.0
```
Gentoo:
```
libboost_python-3.6-mt.so
libboost_python-3.6.a
libboost_python-3.6.so
```
Ubuntu:
```
libboost_python-py27.so.1.62.0
libboost_python-py35.so.1.62.0
libboost_python-py36.so.1.62.0
```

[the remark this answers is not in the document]
ssiloti: Put it as a comment at the top of the file.

The default suffix should be `python` if the major version is 2 else `python${major}`. I'm not too worried about what all the different Linux distros do. They have package maintainers who can figure out what needs to be done to make it work with their unique naming schemes. I'm worried about Windows users. They all use the boost build naming, they're more likely to be building libtorrent manually, and they're the most likely to post a github issue when they get an error.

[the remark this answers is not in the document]
evsh: Done

bindings/python/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+else()
+	set(_boost_python_suffix "${boost-python-suffix}")
+endif()
+
+find_package(Boost REQUIRED COMPONENTS python${_boost_python_suffix})
+
+execute_process(COMMAND
+  ${PYTHON_EXECUTABLE} -c "import distutils.sysconfig;
+print(distutils.sysconfig.get_python_lib(plat_specific=True, prefix=''));
+print(distutils.sysconfig.get_config_var('EXT_SUFFIX'))"
```
ssiloti: Python 2.7 doesn't define `EXT_SUFFIX` so the output file of `python-libtorrent` ends up being named `libtorrentNone`.

evsh: Refactored this piece, thanks! I did not test with python 2 :(

evsh:
Thanks for the review, @ssiloti!

CMakeLists.txt
```diff
@@ -600,40 +613,37 @@ endif()
...
+	target_link_libraries(torrent-rasterbar
+		PRIVATE
+			wsock32 ws2_32 Iphlpapi
+			debug dbghelp
+	)
+
+	target_compile_definitions(torrent-rasterbar
+		PUBLIC
+			_WIN32_WINNT=0x0600 # target Windows Vista or later
+			WIN32_LEAN_AND_MEAN # prevent winsock1 to be included
```
arvidn: I recognise this wasn't introduced by this change, but does anyone have any thoughts on exporting this define? In one way it seems a bit intrusive. It's essentially saying no clients linking against libtorrent is allowed to use microsoft's macros. On the other hand, most likely only software written in the 90's use those anyway.

arvidn:
does `bindings/python/CMakeLists.txt` supplant the need for `setup.py`? (it would be exciting to be able to build python packages with cmake directly). I suspect not, given the windows compiler version logic @ssiloti put in there recently

arvidn:
Runnin this: ``$ cmake -Dpython-bindings=1 .`` I get this error:

```
  File "<string>", line 4, in <module>
TypeError: sequence item 1: expected string, NoneType found
CMake Error at bindings/python/CMakeLists.txt:73 (list):
  list GET given empty list


CMake Error at bindings/python/CMakeLists.txt:74 (list):
  list GET given empty list


-- Python site packages:
-- Python extension suffix:
CMake Error at bindings/python/CMakeLists.txt:82 (set_target_properties):
  set_target_properties called with incorrect number of arguments.


CMake Error at bindings/python/CMakeLists.txt:101 (install):
  install TARGETS given no LIBRARY DESTINATION for module target
  "python-libtorrent".


-- Configuring incomplete, errors occurred!
See also "/Users/arvid/Documents/dev/libtorrent/CMakeFiles/CMakeOutput.log".
```

The output log is quite verbose and doesn't contain anything that appears relevany

aldenml:
maybe you add the python build to the cmake travis task

evsh:
> Running this: $ cmake -Dpython-bindings=1 . I get this error:
It means the following python snippet fails:
```python
import distutils.sysconfig;
print(';'.join([
	distutils.sysconfig.get_python_lib(plat_specific=True, prefix=''),
	distutils.sysconfig.get_config_var('EXT_SUFFIX')
]))
```
Could you, please, run it with your python interpreter manually?
Likely, it is python 2 and the second call returns Nobe, which is unacceptable by the list.join.
The snippet should probably be changed to:
```python
import distutils.sysconfig;
print(';'.join([
	distutils.sysconfig.get_python_lib(plat_specific=True, prefix=''),
	str(distutils.sysconfig.get_config_var('EXT_SUFFIX'))
])
```
Pardon me, am on vacation until Friday without access to my computer.

arvidn:
Yeah, `EXT_SUFFIX` is empty for me. I think this is the default python on OSX.

```
>>> distutils.sysconfig.get_python_lib(plat_specific=True, prefix='')
'lib/python2.7/site-packages'
>>> distutils.sysconfig.get_config_var('EXT_SUFFIX')
>>>
```

arvidn:
@evsh  I just realized that the logic @ssiloti put in here to pick the right version of visual studio based on python version. Is that preserved? the `setup.py.in` will overwrite all that, right?

evsh:
> Yeah, EXT_SUFFIX is empty for me.

My bad, had to test that change. So, we probably better use ';'.join(map(str, [ . I will update the PR on Friday.

> ...pick the right version of visual studio based on python version.

Sorry, can't understand that.

arvidn:
@evsh This logic [here](https://github.com/arvidn/libtorrent/blob/master/bindings/python/setup.py#L95)

When you build compiled python modules on windows, you have to use the same version of visual studio as python itself was compiled with (for ABI compatibility). There are conventions for different python versions.

CMakeLists.txt
ssiloti: Defining `WIN32_LEAN_AND_MEAN` has been SOP for 20 years, I doubt anyone is going to complain.

evsh:
Thanks. I can't imagine how can that be applied here: we select compiler when start cmake build. Why would we build bindings with another compiler. On the opposite hand, one can point cmake to a python executable of our choice.

arvidn:
I’m not sure what would be best to do either. Perhaps fail the build with a friendly error message if the python version doesn’t match the compiler that’s already been selected. @ssiloti, do you have any ideas?

ssiloti:
We should set the `Python_ADDITIONAL_VERSIONS` variable prior to loading the `FindPythonInterp` module when we detect an msvc toolset. That variable is just a guideline though, we'll still need to check the found Python version against the toolset version and show an error if there is a mismatch.

evsh:
Thank you, I will add that.

CMakeLists.txt
```diff
@@ -655,52 +665,68 @@ else()
...
-	# disable bogus deprecation warnings on msvc8
-	target_compile_definitions(torrent-rasterbar PUBLIC _SCL_SECURE_NO_DEPRECATE _CRT_SECURE_NO_DEPRECATE)
-	# these compiler settings just make the compiler standard conforming
-	target_compile_options(torrent-rasterbar PUBLIC /Zc:wchar_t /Zc:forScope)
-	# for multi-core compilation
-	target_compile_options(torrent-rasterbar PUBLIC /MP)
-	# increase the number of sections for obj files
-	target_compile_options(torrent-rasterbar PUBLIC /bigobj)
+	target_compile_definitions(torrent-rasterbar
+		PUBLIC
```
evsh: These ones probably should be private?

evsh:
> We should set the Python_ADDITIONAL_VERSIONS

Done. I don't use MSVC through, thus can't test that checks properly (did a mockup run with GCC). Also, I'm not sure about MSVC 14: am I right rejecting versions 14.1 and higher as incompatible with MSVC 14.0 runtime?

.travis.yml
```diff
@@ -10,7 +10,7 @@ matrix:
     - env: variant=test_debug sim=1 crypto=openssl toolset=gcc-sanitizer
     - env: variant=test_release coverage=1 tests=1 toolset=gcc-coverage python=1
     - env: autotools=1 toolset=gcc
-    - env: cmake=1 toolset=gcc
+    - env: cmake=1 toolset=gcc python=1
```
arvidn: this makes me a bit nervous. Don't you think it's enough to just always build python bindings when invoking cmake? we only do it in one configuration anyway

.travis.yml
```diff
@@ -207,7 +207,7 @@ script:
 # a binary built with sanitizer does not interact well with python
 # on OSX we need to use the brew version of python, for reasons explained above
   - cd bindings/python
-  - 'if [[ "$python" == "1" ]]; then
+  - 'if [[ "$python" == "1" && "${cmake+0}" == "0" ]]; then
```
arvidn: because now you need complexity like this

.travis.yml
evsh: I have no opinion on that. If your suggestion is to build the bindings always, that makes things easier, of course. Should I change that?

arvidn: I think the `cmake` invocation could always build the bindings, yes.

ssiloti:
MS sort of follows semver with msvc version numbers. Version 14 and 14.1 are ABI compatible so both will work with Python 3.5 and 3.6.

.travis.yml
evsh: OK, done.

evsh:
> Version 14 and 14.1 are ABI compatible so both will work with Python 3.5 and 3.6.

Thanks!

evsh:
Not yet.

evsh:
I'm puzzled. It does work find on my local machine with a default boost installation. Here it fails to find libboost_python3. But it finds boost system library...

evsh:
Now cmake bindings build fails in the very same way as the b2 one.

evsh:
If I comment out [this block](https://github.com/arvidn/libtorrent/blob/master/bindings/python/src/boost_python.hpp#L27), MSVS 2017 compiles the bindings. @arvidn, could you suggest how to proceed, please?

appveyor.yml
```diff
@@ -12,6 +12,7 @@ environment:
     compiler: msvc-14.0
     model: 64
     sim: 1
+    python: 1
```
arvidn: please restore this as well. now the python bindings are built twice

arvidn:
the `snprintf` error looks like it's not built as C++11 for some reason

arvidn:
I don't understand how your patch can have affected the `b2` build...

evsh:
> the snprintf error looks like it's not built as C++11 for some reason

I know little to nothing about MSVC, but Google says MSVS 2017, which I use in my local tests, is in C++14 mode by default. How could I turn it off?

evsh:
Removing that superfluous 'python: 1' fixed the b2 build. A magic?

evsh:
With boost 1.67 MSVS 2017 compiles it successfully.

appveyor.yml
```diff
@@ -83,7 +83,7 @@ build_script:
 - cd %ROOT_DIRECTORY%\bindings\python
 # we use 64 bit python builds
 - if defined python (
-  b2.exe --hash openssl-version=pre1.1 warnings-as-errors=on -j2 %compiler% address-model=%model% picker-debugging=on invariant-checks=full variant=%variant% %linkflags% %include% link=shared crypto=%crypto% libtorrent-link=shared stage_module stage_dependencies
+  b2.exe -d+2 --hash openssl-version=pre1.1 warnings-as-errors=on -j2 %compiler% address-model=%model% picker-debugging=on invariant-checks=full variant=%variant% %linkflags% %include% link=shared crypto=%crypto% libtorrent-link=shared stage_module stage_dependencies
```
arvidn: oh, this looks like debugging left-overs too

evsh: Right, I'm still on it. I did something last night to make my local MSVS compile it, and left a comment about Boost 1.67 above. Now I can't reproduce that.

evsh: @arvidn, would you accept the workaround from the last commit? I'm about to give up on finding the real cause of the problem. Also, what to do with the warnings which are treated as errors?

appveyor.yml
```diff
@@ -96,7 +96,8 @@ build_script:
 - cd %ROOT_DIRECTORY%
 - mkdir build && cd build
 - if defined cmake (
-  cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS="/WX" -G "Visual Studio 14 2015 Win64" .. &&
+  set "PATH=c:\Python35-x64;%PATH%" &&
+  cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS="/WX" -Dpython-bindings=%python% -Dboost-python-module-name="python" -G "Visual Studio 14 2015 Win64" .. &&
```
evsh: `-DCMAKE_BUILD_TYPE=Release` makes no sense with the Visual Studio generator. To run release build we have to either run build using msbuild invocation and specify the configuration name, or generate makefiles.

evsh:
Now it compiles but linking fails because all boost::python binaries in AppVeyor images are linked against python 2.7

appveyor.yml
aldenml: Do you mind linking documentation to such information (that `-DCMAKE_BUILD_TYPE=Release` does not make sense here)?

evsh: https://cmake.org/cmake/help/v3.11/variable/CMAKE_BUILD_TYPE.html

aldenml: thanks

evsh: I’m happy to help

evsh:
@arvidn, @ssiloti, review these changes again, please; AppVeyor build passes.

CMakeLists.txt
```diff
@@ -449,6 +451,11 @@ option(libiconv "enable linking against system libiconv" OFF)
 option(logging "build with logging" ON)
 option(build_tests "build tests" OFF)
 option(build_examples "build examples" OFF)
+option(python-bindings "build python bindings" OFF)
+
+if (build_tests OR (WIN32 AND python-bindings))
```
ssiloti: The python bindings should not be using extra exported symbols. Is there an issue for this?

CMakeLists.txt
ssiloti: Eh, I think of it as a community service to disable that BS.

CMakeLists.txt
evsh: On Windows the bindings library links to openssl symbols.

evsh:
PR updated: generator expressions are taken into account for pkg-config. For that we generate file in two steps: using `configure_file()` we expand variables and at the second step `file(GENERATE)` is used to expand generator expressions.

evsh:
Update once again: include `INTERFACE_COMPILE_OPTIONS` property into the pkgconfig.

evsh:
Hm, that method for handling generator expressions is not perfect: since they are expanded at the very last stage of the processing, I can't remove empty items, which results in empty '-D' switches in the pkg-config file.

arvidn:
@evsh did you see this? https://ci.appveyor.com/project/arvidn/libtorrent/build/8925/job/wf6amftj1hlt5o29#L1106

evsh:
Thanks, I saw that, yes. It's tool late here already today, I will fix that tomorrow. Apparently, multi-configuration generator (MSVS) + a package with multiple available configurations (boost with debug and release libraries installed) requires handling those configurations explicitly.

evsh:
I think its done now: the cmake build can be used to install the package into a system and build libtorrent clients written in C++ or Python.

I would like to make a few more little improvements and clean/beautify code in CMakeLists. Would you like me to push those changes to this PR, or wait for a decision on this one and open another PR?

arvidn:
please polish this PR to prepare to land it in master. please also squash any commits you think are appropriate to. I will pull this down at test it locally

CMakeLists.txt
```diff
@@ -567,23 +579,38 @@ else()
...
 if (encryption)
 	if(NOT DEFINED OPENSSL_INCLUDE_DIR OR NOT DEFINED OPENSSL_LIBRARIES)
 		FIND_PACKAGE(OpenSSL REQUIRED)
 	endif()
-	target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_OPENSSL TORRENT_USE_LIBCRYPTO)
+	list(APPEND _package_dependencies "OpenSSL")
 	target_include_directories(torrent-rasterbar PUBLIC ${OPENSSL_INCLUDE_DIR})
+	target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_OPENSSL TORRENT_USE_LIBCRYPTO)
+	if (_libtorrent_additional_exports)
+		target_link_libraries(torrent-rasterbar PUBLIC ${OPENSSL_LIBRARIES})
```
arvidn: shouldn't this be specifying the dependency on proper targets instead? like "OpenSSL::SSL OpenSSL::Crypto", and remove the call to `target_include_directory()` above, where the include directory is added manually.

CMakeLists.txt
```diff
@@ -567,23 +579,38 @@ else()
...
+	list(APPEND _package_dependencies "Iconv")
 	target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_ICONV)
-	target_include_directories(torrent-rasterbar SYSTEM PUBLIC ${ICONV_INCLUDE_DIR})
+	target_include_directories(torrent-rasterbar PUBLIC ${ICONV_INCLUDE_DIR})
+	target_link_libraries(torrent-rasterbar PRIVATE ${ICONV_LIBRARIES})
 endif (libiconv)
 
 if (encryption)
 	if(NOT DEFINED OPENSSL_INCLUDE_DIR OR NOT DEFINED OPENSSL_LIBRARIES)
 		FIND_PACKAGE(OpenSSL REQUIRED)
```
arvidn: is `FIND_PACKAGE` different from `find_package` in cmake? if not, I suppose this should be lower case too, like all other invocations.

CMakeLists.txt
arvidn: also, it doesn't seem like `find_package()` works to find openssl from homebrew on mac. But I suppose that's a bug/missing feature in cmake itself.

arvidn:
should `bindings/python/setup.py` be removed from the repository, now that it's generated?
it seems a bit odd to have the build process modify a file that's under version control.

CMakeLists.txt
arvidn: something like this
```
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -600,16 +600,13 @@ if (libiconv)
 endif (libiconv)

 if (encryption)
-       if(NOT DEFINED OPENSSL_INCLUDE_DIR OR NOT DEFINED OPENSSL_LIBRARIES)
-               FIND_PACKAGE(OpenSSL REQUIRED)
-       endif()
+       find_package(OpenSSL REQUIRED)
        list(APPEND _package_dependencies "OpenSSL")
-       target_include_directories(torrent-rasterbar PUBLIC ${OPENSSL_INCLUDE_DIR})
        target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_OPENSSL TORRENT_USE_LIBCRYPTO)
        if (_libtorrent_additional_exports)
-               target_link_libraries(torrent-rasterbar PUBLIC ${OPENSSL_LIBRARIES})
+               target_link_libraries(torrent-rasterbar PUBLIC OpenSSL::SSL OpenSSL::Crypto)
        else()
-               target_link_libraries(torrent-rasterbar PRIVATE ${OPENSSL_LIBRARIES})
+               target_link_libraries(torrent-rasterbar PRIVATE OpenSSL::SSL OpenSSL::Crypto)
        endif()
 else()
        target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_DISABLE_ENCRYPTION)
```

except, I think the public headers of libtorrent may include openssl headers, so the link probably has to be unconditionally `PUBLIC`.

appveyor.yml
```diff
@@ -96,8 +96,9 @@ build_script:
 - cd %ROOT_DIRECTORY%
 - mkdir build && cd build
 - if defined cmake (
-  cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS="/WX" -G "Visual Studio 14 2015 Win64" .. &&
-  cmake --build .
+  set "PATH=c:\Python27-x64;%PATH%" &&
+  cmake -DCMAKE_CXX_FLAGS="/WX /wd4244 /wd4459" -Dpython-bindings=%python% -Dboost-python-module-name="python" -Dskip-python-runtime-test=true -DPython_ADDITIONAL_VERSIONS="2.7" -G "Visual Studio 14 2015 Win64" .. &&
+  cmake --build . --config Release
```
arvidn: this `--config` flag, is it an msvc-only option? It looks a lot better than `-DCMAKE_BUILD_TYPE=`. Could it be used on travis as well?

CMakeLists.txt
evsh: Yeah, imported targets (OpenSSL::SSL OpenSSL::Crypto here) would make a cleaner code, of course. I was puzzled by the fact, already mentioned by you just above, that headers are always public dependency while the library is either private or public.  When you say let's make them all unconditionally public, I will make the change.

CMake needs a [hint](https://github.com/qbittorrent/qBittorrent/blob/master/.travis.yml#L154) to find the openssl brew package. I guess the brew package installs it in an uncommon location.

The canonical way for CMake (and as far as I understand this is what package maintainers got used to) is to make the `find_package()` call  first (without `REQUIRED` argument) and enable or disable option basing on the outcome. CMake provides a standard way to make `find_package()` fail for a given package name. I was looking into the sources to learn how do OpenSSL headers are used, and found that GCrypt is also an option. Thus I would replace the `option(encryption ...)` call with option `find_package()` calls for OpenSSL and then GCrypt (or in the opposite order; don't know which one is preferred).

CMakeLists.txt
evsh: There is no case sensitivity here.

CMakeLists.txt
arvidn: while looking at this earlier I noticed that enabling encryption (bittorrent protocol obfuscation) and linking against openssl are tied together here. They shouldn't be and there's no reason for them to be.

I don't think it's important to be able to disable encryption (but that's just controlled by the define `TORRENT_DISABLE_ENCRYPTION`). linking against openssl enables HTTPS support. HTTPS is not supported by gcrypt (it's up to boost.asio). The gcrypt option is just to use gcrypt instead of the built-in functions for RC4 and SHA1.

I was going to make a separate patch once this lands, but if you're up for it, I think renaming this option to just openssl or something like that would make sense. And have it not affect the encryption settings.

appveyor.yml
evsh: No, they differ. There are two types of generators in CMake: single configuration and multi-configuration. The first ones (e.g. Makefile) are controlled by the regular CMake variable during configuration phase (`CMAKE_BUILD_TYPE`, actually just selects a predefined set of compiler and linker flags), those of the second type (e.g. MSVS) have their own configurations inside and the `--config` switch is used when invoking build (i.e. when configuration stage has finished).

CMakeLists.txt
evsh: So, I should then make the `encryption`option to control the `TORRENT_DISABLE_ENCRYPTION` flag only, and the `find_package(OpenSSL)` to be independent? Similarly, the `libiconv` option is also useless and can be replace by `find_package()`. Then we get optional dependencies and it would make sense to notify user at the end of configure stage about results. CMake provides a [standard way](https://cmake.org/cmake/help/latest/module/FeatureSummary.html) for that. But then it makes sense to put all the options into that, which will be quite a big change and let's leave it for another PR, please (I already have those changes adding `feature_summary()` in my local repo)?

evsh:
> should `bindings/python/setup.py` be removed from the repository, now that it's generated?

Sorry, can't understand is that question related to this PR? As far as I understood, this file is required to launch `b2` from distutils and pass required flags to the compiler.

arvidn:
it was a question, asking if it makes sense to delete that file. It seems problematic to generate a file that overwrites a file in the repository. It seems it would ask for someone to accidentally commit the changes.

afaik, `setup.py` is there to build an egg file (or whatever kinds of packages python uses). You can still build the python bindings with `b2`.

The `setup.py` file that cmake writes, does it do something else?

CMakeLists.txt
ssiloti: There should be options for all optional dependencies. Automagic dependencies are annoying for Linux package maintainers.

evsh: Disagree. `find_package()` call is a nice way to signal optional dependence. In case of CMake, I think we can follow the style used by the KDE project, which is probably the biggest CMake consuming community, and package maintainers know what to do in that case very well. Also, time to time I observe Gentoo maintainers insist on this very model in KDE bugzilla/phabrikator.

evsh:
Thanks for the clarification. The `setup.py` from the CMake build does only generate the egg.

evsh:
Fixed a few little mistakes and updated openssl linking. I think I do not want to squash commits any further.

CMakeLists.txt
ssiloti: I see, I did not realize there is a standard way to tell `find_package` not to search for a particular package. That's fine then.

bindings/python/src/boost_python.hpp
```diff
@@ -5,6 +5,12 @@
 #ifndef BOOST_PYTHON_HPP
 #define BOOST_PYTHON_HPP
 
+#ifdef _MSC_VER
+// to work around a mysterious msvc bug (see snprinf defs below) that results in
+// C2039: 'snprintf' is not a member of 'std'
+#include <cstdio>
```
arvidn: it's possible this file isn't included anywhere before ``snprintf()`` is used, but on clang and gcc maybe it got pulled in as a side effect. it should probably be included unconditionally

bindings/python/src/error_code.cpp
```diff
@@ -64,7 +64,7 @@ namespace {
 	struct ec_pickle_suite : boost::python::pickle_suite
 	{
 		static boost::python::tuple
-		getinitargs(error_code const& ec)
+		getinitargs(error_code const& /*ec*/)
```
arvidn: the typical way of dealing with these warnings is to just delete the argument name

evsh: Pardon, but this is precisely what is done here, no?

arvidn: oh, I meant deleted instead of just commented out. not a big deal. I have a rule-of-thumb, if a parameter is difficult to understand what it does when its name is deleted, it should probably have a more specific type

CMakeLists.txt
```diff
@@ -567,23 +576,33 @@ else()
...
 endif (libiconv)
 
 if (encryption)
-	if(NOT DEFINED OPENSSL_INCLUDE_DIR OR NOT DEFINED OPENSSL_LIBRARIES)
-		FIND_PACKAGE(OpenSSL REQUIRED)
+	if(NOT TARGET OpenSSL::SSL)
+		find_package(OpenSSL REQUIRED)
 	endif()
+	list(APPEND _package_dependencies "OpenSSL")
+	target_link_libraries(torrent-rasterbar PUBLIC OpenSSL::SSL)
```
arvidn: is `OpenSSL::SSL` automatically pulling in the dependency on `OpenSSL::Crypto`?

evsh: It does in CMake 3.11, and I believe it does the same in other versions too.

evsh: Yep, checked with their Git.

CMakeLists.txt
```diff
@@ -597,45 +616,36 @@ if (NOT dht)
 endif()
 
 # Boost
-if(NOT DEFINED Boost_INCLUDE_DIR OR NOT DEFINED Boost_LIBRARIES)
-	FIND_PACKAGE(Boost REQUIRED COMPONENTS system)
-endif()
-include_directories(${Boost_INCLUDE_DIRS})
-target_link_libraries(torrent-rasterbar ${Boost_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})
+find_package(Boost REQUIRED COMPONENTS system)
+target_link_libraries(torrent-rasterbar PUBLIC Boost::system)
```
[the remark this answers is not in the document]
evsh: What are the boost and cmake version you are using, please? To speedy up information exchange, the most likely reason of the problem is give below.

Most likely this is due to a problem which I can't solve, unfortunately, and we might want to return to using `${Boost_LIBRARIES}`. Every Boost release can change inter-component dependencies and as such the `FindBoost.cmake` module does list imported target (with correct dependencies) only for release it is aware about. Therefore with each Boost release CMake guys need to change that file adding new boost version. For example, right now CMake 3.11 know about boost versions up to 1.67. If you take a newer boost code, the boost imported targets are absent and you see the following message from FindBoost: "New Boost version may have incorrect or missing dependencies and imported targets".

evsh:
@arvidn, thanks!

CMakeLists.txt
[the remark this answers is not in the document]
aldenml: that's exactly the case, I'm using `cmake 3.10` from CLion, and `boost 1.67`, and getting the message "New Boost version may have incorrect or missing dependencies and imported targets", I will send the PR with the use of `${Boost_LIBRARIES}`, Thanks!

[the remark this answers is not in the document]
evsh: You can always get a patch for the latest released boost from https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake

Issue #2934, opened 2018-04-12
Fix race condition on alert_manager::maybe_notify()


fernando-rodriguez:
**Please provide the following information**

libtorrent version (or branch): RC_1_1

platform/architecture: All

compiler and compiler version: All

There is a race condition on `alert_manager::maybe_notify()` caused by unlocking the mutex before calling `plugin::on_alert()`. If another thread pops alerts twice the alert allocations can become invalid before `on_alert()` completes. It can also result in `on_alert()` being called after the bit for that alert has been cleared.

This cannot normally happen because all alerts are posted from the same thread, but there is a deprecated alert being posted from the IO thread.

fernando-rodriguez:
It would also affect the `set_alert_notify()` callback.

fernando-rodriguez:
I think we can use a separate recursive mutex just to protect the allocator.

arvidn:
I’m in the camp of thinking that recursive mutexes are a symptom of a design issue. Thanks for reporting this, my intention and hope is that there will be more threads in the future, so I’ll try to fix this

arvidn:
there is a fundamental problem here. If the mutex is held while calling the callback, the callback can't do anything that may post another alert (since it's already locked, it would deadlock). That's clearly not a good situation since it's quite likely the callback would want to do something interesting there.

On the other hand, if the mutex is unlocked, not only can another thread cause the passed in alert to become dangling, having a sufficient number of alerts be posted will also cause the queue to grow and possibly reallocate the alert, also making it a dangling pointer.

fernando-rodriguez:
Actually strike the first part. In this case I think it's safe because it's never called with the mutex locked.

arvidn:
I hadn't considered that a recursive mutex may have odd interactions with condition variables. But, as you say, as long as the thread calling `wait()` has only locked the mutex once, it should be safe.

As for the second part, the future I envision would decouple peer connections and not have them block on each other, so I don't think it would be important to be re-entrant from different threads. I consider it a hack to need a recursive_mutex at all, that comes out of having arbitrary user code run in-thread. Normally this would be avoided by posting messages and deferring the handler (how alerts are meant to be handled).

fernando-rodriguez:
This is not part of this report or your PR but I think there's another race condition on `condition_var::wait()` and `wait_for()` on Windows and Beos because the mutex is unlocked before waiting for the semaphore, no?

[...]

PR #2940, opened 2018-04-15
fix alert_manager on_alert() race

arvidn:
make sure the alert_manager is still locked when calling the on_alert() hook on plugins, since they receive the alert, the storage of that alert is protected by the mutex and might otherwise be destroyed from another thread while the on_alert() callback is still running.

This is intended to fix https://github.com/arvidn/libtorrent/issues/2934

### DIFF
--- a/src/alert_manager.cpp
+++ b/src/alert_manager.cpp
@@ -77,6 +77,14 @@ namespace libtorrent
 			|| a->type() == save_resume_data_alert::alert_type)
 			++m_num_queued_resume;
 
+#ifndef TORRENT_DISABLE_EXTENSIONS
+		for (ses_extension_list_t::iterator i = m_ses_extensions.begin()
+			, end(m_ses_extensions.end()); i != end; ++i)
+		{
+			(*i)->on_alert(a);
+		}
+#endif
+
 		if (m_alerts[m_generation].size() == 1)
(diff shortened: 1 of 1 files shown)


ssiloti:
This re-introduces a problem I had when working on my reputation plugin: if a plugin's `on_alert` handler does something which triggers an alert it will deadlock. This is very easy to do when all of the logging alerts are enabled.

ssiloti:
Maybe it's enough to put a big warning on `on_alert` that plugins should `post()` to the `io_service` if they need to do anything non-trivial.

arvidn:
I agree that this feels a bit risky. I was thinking that maybe there's a way to "lock" on alerts being popped just once, and if popped a second time (while in this callback) it blocks until it's done. somehow

arvidn:
actually, I don't think it's quite as simple as I thought. If enough alerts are posted from within this callback, the alert queue and/or the stack allocator will grow and re-allocate. Leaving the `alert*` passed in to `on_alert()` dangling. It's not clear what the best solution would be. As long as you look at the alert first, then do stuff and promise to never look at the alert again, things should be safe. This also seems like a reasonable common way of doing it, but how do you shape the code to communicate that alert will become invalid or enforce that it won't be used again?

PR #2944, opened 2018-04-16
Fix storage initialization

d-komarov:
If `default_storage::has_any_file` fails during fastresume data check,
`piece_manager::check_no_fastresume` will skip storage initialization.
In such case, any storage operation that require part file will cause
an application crash.

### DIFF
--- a/src/storage.cpp
+++ b/src/storage.cpp
@@ -1792,19 +1792,10 @@ namespace libtorrent
 
 	int piece_manager::check_no_fastresume(storage_error& ec)
 	{
-		bool has_files = false;
 		if (!m_storage->settings().get_bool(settings_pack::no_recheck_incomplete_resume))
 		{
 			storage_error se;
-			has_files = m_storage->has_any_file(se);
-
-			if (se)
-			{
-				ec = se;
-				return fatal_disk_error;
-			}
-
-			if (has_files)
+			if (m_storage->has_any_file(se))
 			{
 				// always initialize the storage
 				int ret = check_init_storage(ec);
--- a/src/assert.cpp
+++ b/src/assert.cpp
@@ -352,10 +352,14 @@ TORRENT_EXPORT void assert_fail(char const* expr, int line
 
 	// if production asserts are defined, don't abort, just print the error
 #ifndef TORRENT_PRODUCTION_ASSERTS
+    #ifdef _MSC_VER
+    __debugbreak();
+    #else
 	// send SIGINT to the current process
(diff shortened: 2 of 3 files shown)


arvidn:
would it be difficult to add a test that reproduces the issue?

d-komarov:
No problem.

src/storage.cpp
```diff
@@ -1792,19 +1792,10 @@ namespace libtorrent
...
-			has_files = m_storage->has_any_file(se);
-
-			if (se)
-			{
-				ec = se;
-				return fatal_disk_error;
-			}
-
-			if (has_files)
+			if (m_storage->has_any_file(se))
```
arvidn: it's not obvious to me why it's safe to ignore errors from `has_any_file()`. shouldn't an `EPERM` be a fatal error for instance?

d-komarov: Because otherwise `check_init_storage()` won't be called that is incorrect.

d-komarov: If you strongly believe that storage implementation is correct, then the following code is error-prone:
```
void torrent::on_resume_data_checked(disk_io_job const* j)
{
...
	if (j->ret == piece_manager::fatal_disk_error)
	{
		m_resume_data.reset();
		handle_disk_error(j);
		auto_managed(false);
		pause();
		set_state(torrent_status::checking_files);
		if (should_check_files()) start_checking();
		return;
	}
...
}
```
Torrent shouldn't enter `checking_files` state with uninitialized storage, that is the case when `has_any_file` reports an error.

d-komarov: This is how `torrent::force_recheck` handles similar error:
```
void torrent::on_force_recheck(disk_io_job const* j)
{
    ...
    if (j->ret == piece_manager::fatal_disk_error)
    {
        handle_disk_error(j);
        return;
    }
    ...
}
```
torrent's state doesn't change here, it is still `checking_resume_data`, preventing `start_checking()` calls.

d-komarov: At the end of the day I believe final fix would be:

	int piece_manager::check_no_fastresume(storage_error& ec)
	{
        // always initialize the storage
        m_storage->initialize(ec);

        if (ec) return fatal_disk_error;
            
        return (m_storage->settings().get_bool(settings_pack::no_recheck_incomplete_resume)
            || !m_storage->has_any_file(ec) && !ec) ? no_error : need_full_check;
    }

Only storage initialization error should be treated as fatal_disk_error.

PR #2948, opened 2018-04-18
Use settings_pack::urlseed_wait_retry for default retry with http seeds.

TheOriginalWinCat:
Currently there is no way to configure the default retry time for http_seeds.
I looked through the history and can see no signs this is intentional.

This change will result in a default of 30 seconds instead of 5 minutes, but since the example in the BEP17 spec uses a 30 second retry time as the default I don't think it's a major issue.

Potentially this would be better as a separate setting, but I see http_seeds already share a few of the url_seed settings.

### DIFF
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -812,7 +812,7 @@ namespace libtorrent {
 			// low number, like 5
 			urlseed_pipeline_size,
 
-			// time to wait until a new retry of a url-seed takes place.
+			// number of seconds until a new retry of a url-seed takes place.
 			// Default retry value for http-seeds that don't provide a valid 'retry-after' header.
 			urlseed_wait_retry,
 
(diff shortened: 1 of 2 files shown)


arvidn:
If you feel up for it, it would probably make sense to mention this use of the setting in the documentation too (which is generated from the comment in the header file where the enum is defined)

TheOriginalWinCat:
I have updated the documentation, I also updated urlseed_pipeline_size comment to mention it's use with http seeds.

include/libtorrent/settings_pack.hpp
```diff
@@ -806,13 +806,14 @@ namespace libtorrent {
...
+			// controls the pipelining size of url and http seeds. i.e. the number of HTTP
 			// request to keep outstanding before waiting for the first one to
 			// complete. It's common for web servers to limit this to a relatively
 			// low number, like 5
 			urlseed_pipeline_size,
 
-			// time to wait until a new retry of a web seed takes place
+			// time to wait until a new retry of a url-seed takes place.
+			// Default retry value for http-seeds that don't provide a valid 'retry-after' header.
 			urlseed_wait_retry,
```
d-komarov: Probably it should be clarified that `urlseed_wait_retry` is measured in seconds.

PR #2968, opened 2018-04-23
cleanup and refactor in storage source code

aldenml:


src/storage.cpp
```diff
@@ -535,7 +535,6 @@ namespace libtorrent {
 
 			if (e)
 			{
-				ec.file(torrent_status::error_file_partfile);
```
arvidn: can you explain this?

arvidn: oh, never mind

PR #2970, opened 2018-04-24
fix disk_io_thread::do_check_fastresume storage initialize logic from RC_1_1 merge

aldenml:
Hi @arvidn, I'm looking hard at storage related changes and if I follow correctly this PR https://github.com/arvidn/libtorrent/pull/2944, I think the merge changed the logic.

Now, I can't assure this fix the new reappearance of the old storage crash, appreciate your review.


arvidn:
good catch!

PR #2971, opened 2018-04-24
Fix magnet URIs handling in client.py

evsh:


bindings/python/client.py
```diff
@@ -156,7 +156,8 @@ def print_download_queue(console, download_queue):
 def add_torrent(ses, filename, options):
     atp = lt.add_torrent_params()
     if filename.startswith('magnet:'):
-        atp = lt.parse_magnet_uti(filename)
+        for key, value in lt.parse_magnet_uri(filename).items():
+            setattr(atp, key, type(getattr(atp, key))(value))
```
arvidn: did the return value from `parse_magnet_uri()` not work when passing into `add_torrent()`?

evsh: No, it did not.

arvidn: it's supposed to work. both `add_torrent()` and `async_add_torrent()` are overloaded to support both a dict and the `add_torrent_params` object. what's the error you got?

evsh: The problem is not in the `add_torrent()`. The code below the changed fragment expects `atp` to hold an `add_torrent_params` but not `dict` object:
```
  File "client.py", line 360, in <module>
    main()
  File "client.py", line 237, in main
    add_torrent(ses, f, options)
  File "client.py", line 167, in add_torrent
    atp.save_path = options.save_path
AttributeError: 'dict' object has no attribute 'save_path'
```

evsh: P.S. This is after fixing the typo in `parse_magnet_uTi`

arvidn: I see, `parse_magnet_uri()` returns a dict. I suppose this is historical. I will make it return `add_torrent_params`

arvidn:
I fixed this in RC_1_1 and merged it into master [here](https://github.com/arvidn/libtorrent/commit/fc56ec194ae1e189cf7ae9f22d857a807e49a439). I think this supersedes this patch

evsh:
yep. thanks!

PR #2973, opened 2018-04-25
fix empty outgoing interfaces for UDP sockets

arvidn:
@ssiloti would you mind taking a look?
I ran into this problem when setting `outgoing_interfaces` to something that failed to resolve to any valid address, so (I think) no UDP sockets were opened

### DIFF
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -691,6 +691,7 @@ namespace aux {
 			mutable std::condition_variable cond;
 
 			// implements session_interface
+			bool has_udp_outgoing_sockets() const override;
 			tcp::endpoint bind_outgoing_socket(socket_type& s, address
 				const& remote_address, error_code& ec) const override;
 			bool verify_incoming_interface(address const& addr);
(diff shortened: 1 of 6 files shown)


ssiloti:
This will result in dropping the connection and giving up on it right? I think we should prevent outgoing connections from being assigned UTP in `connect_to_peer`, basically treating a lack of outgoing sockets as if `enable_outgoing_utp` is force disabled.

arvidn:
I agree. It would be best to stop earlier in this case. Do you think there's a drawback with also making this a runtime error?

ssiloti:
I'm not sure what you mean by making this a runtime error. You mean posting an alert when there are no outgoing sockets? If so, that seems reasonable.

evsh:
No download/upload activity with this patch if outgoing interface is specified. DHT nodes counter is non-empty.

arvidn:
@ssiloti I meant to make `outgoing_sockets::bind()` able to report errors for this case, rather than asserting (or segfaulting). the next step would be to do something sensible with that error and also to catch the error case earlier to not get into it in the first place. I'm thinking that all of it makes sense.

ssiloti:
sgtm

arvidn:
I'm having second thoughts about making `SO_BINDTODEVICE` best effort, and just log errors. If it's OK to not bind to a specific NIC, the client can just not configure the `outgoing_interfaces` setting. But perhaps this failing *is* a fatal error for some clients.

PR #2975, opened 2018-04-26
replace the TORRENT_CFG ABI check with inline namespaces

arvidn:
around the symbol that can change ABI based on build configuration options


sledgehammer999:
>means you would have to drop support for libtorrent-1.0.x

This is fine by me. At most we support 2 major versions of libtorrent. So once we start supporting master (aka 1.2 ?) then 1.0.x support will be dropped.
Currently we don't support master (since it isn't released). So I assume, once we do, it wouldn't be much pain to start adding `libtorrent/fwd.hpp`

As for the ramifications of this PR, I invite @glassez and @Chocobo1 to comment on this. (if any comment is warranted)

aldenml:
@arvidn in your previous iteration, I got to understand the general idea behind this, but now, I wonder if this is really that useful. For example, if in the future you go for `c++17` in `1.3.X`, can you really provide an API and ABI compatible binary with both code from the old `1.2.X` and the new `master`? Because if at the end, the client needs to recompile, I don't see the point, but again, I might be completely missing the point. Is this only to handle minor changes inside the same major version?

I feel that providing a namespace for assert is strange. Related to the support of IPv6, I'm starting to think that maybe drop support for compilation of only IPv4 and handle errors is more viable solution going forward.

arvidn:
keep in mind the existing function that I'm replacing. the `TORRENT_CFG` function is a way to cause link errors when a client uses an incompatible build configuration from the libtorrent library was built with.

The `ipv6`, `assrt` and `v1_1` are there to indicate which configuration those specific symbols are built with. If a client use them with a different configuration, they should have different link-names to cause a link error. The link error is the feature.

This opens the door for exporting providing both versions of functions, but there isn't really any infrastructure for that right now.

arvidn:
There was a very small and unnecessary part of the public ABI that depended on asserts, so I just removed that.

This new approach replaces TORRENT_NO_DEPRECATE with an ABI version macro, which should simplify updating in future versions

aldenml:
and the nightmare of the SWIG wrapper begins :D

arvidn:
how does it affect the wrapper? the API is unchanged, right?

aldenml:
the `swig` command choke at parsing the files, but I'm just starting

aldenml:
after minor modifications in my swig generation scripts, it's working again

aldenml:
@arvidn to let you know that SWIG fails to parse `inline namespace`, I was able to work with it because you use a macro that I can redefine, but if you opt for plain `inline namespace`, it will be a struggle to work around.

PR #2976, opened 2018-04-26
Use _CFLAGS_OTHER for compiler definitions

W4RH4WK:
`PC_LIBTORRENT_RASTERBAR_CFLAGS` not only gives you the definitions, but
also includes the include path flags (`-I...`).

This causes problesm when passing the result to
`target_compile_definitions`. The resulting compiler command would then
look something like this:

    /usr/bin/c++   -D-I/home/...

This change uses `PC_LIBTORRENT_RASTERBAR_CFLAGS_OTHER` instead which
does not contain the include path flags.

An example of this can be seen here:
https://github.com/IT-Syndikat/arr-torrent/tree/8703fc63354f1e3e0de2e8ee05f23e550b335a86

### DIFF
--- a/examples/cmake/FindLibtorrentRasterbar.cmake
+++ b/examples/cmake/FindLibtorrentRasterbar.cmake
@@ -24,7 +24,7 @@ if(LibtorrentRasterbar_USE_STATIC_LIBS)
 endif()
 
 if(PC_LIBTORRENT_RASTERBAR_FOUND)
-    set(LibtorrentRasterbar_DEFINITIONS ${PC_LIBTORRENT_RASTERBAR_CFLAGS})
+    set(LibtorrentRasterbar_DEFINITIONS ${PC_LIBTORRENT_RASTERBAR_CFLAGS_OTHER})
 else()
     if(LibtorrentRasterbar_CUSTOM_DEFINITIONS)
         set(LibtorrentRasterbar_DEFINITIONS ${LibtorrentRasterbar_CUSTOM_DEFINITIONS})


arvidn:
Is there any overlap with https://github.com/arvidn/libtorrent/pull/2938 ?

W4RH4WK:
No, I don't think so. The pkg-config file already seems to be correct. `PC_LIBTORRENT_RASTERBAR_CFLAGS` should indeed contain the include path flags as this corresponds to pkg-config's `--cflags` option. pkg-config provides also `--cflags-only-I` to output only include path flags, as well as `--cflags-only-other`. The latter corresponds to `PC_LIBTORRENT_RASTERBAR_CFLAGS_OTHER` which seems to be what we actually want for compile definitions.

ssiloti:
This change is fine, although `LibtorrentRasterbar_DEFINITIONS`'s semantics still don't match what the comments claim. The `_CFLAGS_OTHER` variable can also contain things other than preprocessor definitions so it still can't safely be passed to `target_compile_definitions`.

evsh:
Why don't we do (did not test):
```cmake
string(REPLACE " " ";" tmpList "${flags_from_pkg-config}") # make cmake list out of the flags
set(definitions "${tmpList}")
list(FILTER definitions INCLUDE REGEX "-D.+") # leave only those starting with '-D'
string(REGEX REPLACE "(^|;)(-D)" "\\1" definitions "${definitions}") # remove those '-D's
set(options ${tmpList})
list(FILTER options EXCLUDE REGEX "-D.+") # put everything else into `COMPILE_OPTIONS`
```

W4RH4WK:
@evsh I guess this would work (haven't tested it). You may still want to use `PC_LIBTORRENT_RASTERBAR_CFLAGS_OTHER`, otherwise the include paths would be part of `COMPILE_OPTIONS` which you probably do not want.

W4RH4WK:
Alternatively you could provide something like `LibtorrentRasterbar_CFLAGS_OTHER` which can be used in conjunction with `target_compile_options` as mentioned [here](https://stackoverflow.com/a/29316084).

It saddens me that CMake does not provide better integration with pkg-config. I mean, it's rather obvious what we want to do, yet we still have to do everything manually.

evsh:
@W4RH4WK, I wanted to properly separate compile options and definitions and replied to https://github.com/arvidn/libtorrent/pull/2976#issuecomment-385191115. If we do not want to complicate the split by accounting for quotes and spaces in paths, it is clearly better to take the include directories list from the  `_CFLAGS_OTHER` as you do suggest.

PR #2978, opened 2018-04-26
fix use after free in flush_range and flush_iovec

ssiloti:
Calling `blocks_flushed` can cause the piece entry to be freed so its
callers need to be aware of that and avoid dereferencing the pointer if
the entry is freed.


arvidn:
nice catch!

aldenml:
is this problem present in `master`?

ssiloti:
Yes, this bug also exists in master.

PR #2981, opened 2018-04-28
Test merge

arvidn:


src/disk_io_thread.cpp
```diff
@@ -802,9 +806,8 @@ constexpr disk_job_flags_t disk_interface::cache_hit;
 			flush_iovec(pe, iov, flushing, iov_len, error);
 		}
 
-		iovec_flushed(pe, flushing.data(), iov_len, 0, error, completed_jobs);
-
-		m_disk_cache.maybe_free_piece(pe);
+		if (iovec_flushed(pe, flushing.data(), iov_len, 0, error, completed_jobs))
```
ssiloti: This should be `if (!iovec_flushed...`.

arvidn: thanks

PR #2986, opened 2018-04-29
more message-sent hooks for extensions

ghazel:
It's useful for a `peer_plugin` to know when messages of certain types have been sent. There was already `sent_unchoke`; I needed `sent_interested` and `sent_request`, so I added all the ones that I could imagine would be useful.

It's a fairly small patch, but maybe there is a simpler way than the `*e.*` part.

### DIFF
--- a/include/libtorrent/bt_peer_connection.hpp
+++ b/include/libtorrent/bt_peer_connection.hpp
@@ -226,6 +226,9 @@ namespace libtorrent {
 		void on_extended_handshake();
 #endif
 
+		template<class F, typename... Args>
+		void extension_notify(F message, Args... args);
+
 		// the following functions appends messages
 		// to the send buffer
 		void write_choke() override;
--- a/include/libtorrent/extensions.hpp
+++ b/include/libtorrent/extensions.hpp
@@ -449,7 +449,7 @@ namespace libtorrent {
 		virtual void sent_unchoke() {}
 		virtual void sent_interested() {}
 		virtual void sent_not_interested() {}
-		virtual void sent_have(int) {}
+		virtual void sent_have(piece_index_t) {}
 		virtual void sent_piece(peer_request const&) {}
 
 		// called after piece data has been sent to the peer
--- a/include/libtorrent/extensions.hpp
+++ b/include/libtorrent/extensions.hpp
@@ -439,18 +439,18 @@ namespace libtorrent {
 
 		virtual void sent_have_all() {}
 		virtual void sent_have_none() {}
-		virtual void sent_reject_request(piece_index_t, int, int) {}
+		virtual void sent_reject_request(peer_request const&) {}
 		virtual void sent_allow_fast(piece_index_t) {}
 		virtual void sent_suggest(piece_index_t) {}
-		virtual void sent_cancel(piece_index_t, int, int) {}
-		virtual void sent_request(piece_index_t, int, int) {}
+		virtual void sent_cancel(peer_request const&) {}
+		virtual void sent_request(peer_request const&) {}
 		virtual void sent_choke() {}
 		// called after a choke message has been sent to the peer
 		virtual void sent_unchoke() {}
 		virtual void sent_interested() {}
 		virtual void sent_not_interested() {}
 		virtual void sent_have(int) {}
-		virtual void sent_piece(piece_index_t, int, int) {}
+		virtual void sent_piece(peer_request const&) {}
 
 		// called after piece data has been sent to the peer
 		// this can be used for stats book keeping
--- a/include/libtorrent/extensions.hpp
+++ b/include/libtorrent/extensions.hpp
@@ -437,8 +437,20 @@ namespace libtorrent {
 		virtual bool on_reject(peer_request const&) { return false; }
 		virtual bool on_suggest(piece_index_t) { return false; }
 
+		virtual void sent_have_all() {}
+		virtual void sent_have_none() {}
+		virtual void sent_reject_request(piece_index_t, int, int) {}
+		virtual void sent_allow_fast(piece_index_t) {}
+		virtual void sent_suggest(piece_index_t) {}
+		virtual void sent_cancel(piece_index_t, int, int) {}
+		virtual void sent_request(piece_index_t, int, int) {}
+		virtual void sent_choke() {}
 		// called after a choke message has been sent to the peer
 		virtual void sent_unchoke() {}
+		virtual void sent_interested() {}
+		virtual void sent_not_interested() {}
+		virtual void sent_have(int) {}
+		virtual void sent_piece(piece_index_t, int, int) {}
 
 		// called after piece data has been sent to the peer
 		// this can be used for stats book keeping
--- a/src/bt_peer_connection.cpp
+++ b/src/bt_peer_connection.cpp
@@ -348,6 +348,17 @@ namespace {
 		stats_counters().inc_stats_counter(counters::num_outgoing_dht_port);
 	}
 
+	template<class F, typename... Args>
+	void bt_peer_connection::extension_notify(F message, Args... args)
+	{
+#ifndef TORRENT_DISABLE_EXTENSIONS
+		for (auto const& e : m_extensions)
+		{
+			(*e.*message)(args...);
+		}
+#endif
+	}
+
 	void bt_peer_connection::write_have_all()
 	{
 		INVARIANT_CHECK;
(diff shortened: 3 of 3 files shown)


ghazel:
@ssiloti

include/libtorrent/extensions.hpp
```diff
@@ -437,8 +437,20 @@ namespace libtorrent {
...
+		virtual void sent_suggest(piece_index_t) {}
+		virtual void sent_cancel(piece_index_t, int, int) {}
+		virtual void sent_request(piece_index_t, int, int) {}
+		virtual void sent_choke() {}
 		// called after a choke message has been sent to the peer
 		virtual void sent_unchoke() {}
+		virtual void sent_interested() {}
+		virtual void sent_not_interested() {}
+		virtual void sent_have(int) {}
+		virtual void sent_piece(piece_index_t, int, int) {}
```
arvidn: a concern I have with this is that there will be a lot of unnecessary virtual function calls to plugins that don't care, and it will scale poorly if most plugins are uninterested in most of these hooks. The approach I've taken with some other hooks on the (session-level) `plugin` is to have plugins "subscribe" to certain events via the `implemented_features()` bitmask. The session keeps separate lists of *just* the plugins that care about certain hooks. Perhaps this would be overkill at the peer-level plugins. I believe both the `ut_metadata` and `smart_ban` plugins installs peer connection plugins though, so there would be some more looping and making indirect function calls just by introducing these. Is there a simpler way to mitigate this concern?

include/libtorrent/bt_peer_connection.hpp
```diff
@@ -226,6 +226,9 @@ namespace libtorrent {
 		void on_extended_handshake();
 #endif
 
+		template<class F, typename... Args>
+		void extension_notify(F message, Args... args);
```
arvidn: it should probably be: `Args&&... args`

src/bt_peer_connection.cpp
```diff
@@ -348,6 +348,17 @@ namespace {
 		stats_counters().inc_stats_counter(counters::num_outgoing_dht_port);
 	}
 
+	template<class F, typename... Args>
+	void bt_peer_connection::extension_notify(F message, Args... args)
+	{
+#ifndef TORRENT_DISABLE_EXTENSIONS
+		for (auto const& e : m_extensions)
+		{
+			(*e.*message)(args...);
```
arvidn: oh, I guess since we make multiple calls here, we can't move the arguments. so, never mind. they're all primitive types anyway

include/libtorrent/extensions.hpp
```diff
@@ -437,8 +437,20 @@ namespace libtorrent {
 		virtual bool on_reject(peer_request const&) { return false; }
 		virtual bool on_suggest(piece_index_t) { return false; }
 
+		virtual void sent_have_all() {}
+		virtual void sent_have_none() {}
+		virtual void sent_reject_request(piece_index_t, int, int) {}
+		virtual void sent_allow_fast(piece_index_t) {}
+		virtual void sent_suggest(piece_index_t) {}
+		virtual void sent_cancel(piece_index_t, int, int) {}
+		virtual void sent_request(piece_index_t, int, int) {}
```
arvidn: the `peer_request` type is public, so you may as well pass it as the argument for `sent_reject_request()`, `sent_cancel()` and `sent_request()`

src/bt_peer_connection.cpp
```diff
@@ -348,6 +348,17 @@ namespace {
 		stats_counters().inc_stats_counter(counters::num_outgoing_dht_port);
 	}
 
+	template<class F, typename... Args>
+	void bt_peer_connection::extension_notify(F message, Args... args)
```
arvidn: I like this function. I imagine it could be made to replace some of the existing loops and calls to plugins too (in the future).

arvidn:
I think it would make more sense to make the calls into the plugins in `peer_connection.cpp`, to cover web seeds as well. Are there any issues with doing that? There may be some slight duplication, there are two calls to `write_unchoke()` for instance, but I don't think it would be that bad.

include/libtorrent/extensions.hpp
ghazel: Is it significantly costly to call virtual functions which haven't been overridden?

If there's a way to check if the virtual function is overridden at subscription time, that might be easier than requiring the `implemented_features()` bitmask. (something like `&peer_plugin::sent_choke != &foo::sent_choke`)

ssiloti: Unfortunately C++ does not provide any way to compare individual pointers in an object's vtable. You can only test for the presence of an entire class interface with `dynamic_cast`.

ghazel: Per [Stackoverflow](https://stackoverflow.com/a/37588766):

```c++
template <class derived_peer_plugin>
struct peer_plugin {
    virtual void on_event() {}

    peer_plugin() {
        if (&derived_peer_plugin::on_event == &peer_plugin::on_event) {
            std::cout << "not overridden" << std::endl;
        } else {
            std::cout << "overridden" << std::endl;
        }
    }
};

struct a_peer_plugin : peer_plugin<a_peer_plugin> {
    virtual void on_event() {}
};

struct b_peer_plugin : peer_plugin<b_peer_plugin> {
    // no override
};

int main()
{
    a_peer_plugin a;
    b_peer_plugin b;
}
```

=>

```
overridden
not overridden
```

aldenml: this will not fly, what do you think is the relationship between `a_peer_plugin` and `b_peer_plugin` in terms of pure OOP?

ghazel: :P

```c++
struct base_peer_plugin {
public:
    virtual ~base_peer_plugin() {}
};

template <class derived_peer_plugin>
struct peer_plugin : public base_peer_plugin {
    virtual void on_event() {}

    peer_plugin() {
        if (&derived_peer_plugin::on_event == &peer_plugin::on_event) {
            std::cout << "not overridden" << std::endl;
        } else {
            std::cout << "overridden" << std::endl;
        }
    }
};

struct a_peer_plugin : peer_plugin<a_peer_plugin> {
    virtual void on_event() {}
};

struct b_peer_plugin : peer_plugin<b_peer_plugin> {
    // no override
};

int main()
{
    std::list<std::shared_ptr<base_peer_plugin>> m_extensions;

    a_peer_plugin a;
    b_peer_plugin b;

    m_extensions.push_back(std::make_shared<base_peer_plugin>(a));
    m_extensions.push_back(std::make_shared<base_peer_plugin>(b));
}
```

ghazel: Anyway, I'm just goofing off here. If unused virtual function calls are a concern, then probably it's better to not hook so many functions. Remember I only needed three, and one of them already exists in the code.

aldenml: I don't mind :) C++ can be a little tricky at times, specially under heavy optimization

include/libtorrent/extensions.hpp
```diff
@@ -437,8 +437,20 @@ namespace libtorrent {
...
+		virtual void sent_allow_fast(piece_index_t) {}
+		virtual void sent_suggest(piece_index_t) {}
+		virtual void sent_cancel(peer_request const&) {}
+		virtual void sent_request(peer_request const&) {}
+		virtual void sent_choke() {}
 		// called after a choke message has been sent to the peer
 		virtual void sent_unchoke() {}
+		virtual void sent_interested() {}
+		virtual void sent_not_interested() {}
+		virtual void sent_have(int) {}
```
arvidn: this `int` should be `piece_index_t`. it causes a build failure right now: https://ci.appveyor.com/project/arvidn/libtorrent/build/8983/job/m28a9uvqy8r2w33d#L210

arvidn:
lgtm (apart from the typo)

arvidn:
somehow `peer_plugin` is an incomplete type in `bt_peer_connection.cpp`. https://travis-ci.org/arvidn/libtorrent/jobs/378876090#L1405 not sure what's going on

ghazel:
Odd. I don’t get that warning, but I’m also using `clang`.

arvidn:
the stronger types were introduced in master but are still backwards compatible (as in, they allow being implicitly converted to and from `int`) unless deprecated functions are disabled. That's when the type-safety is enforced.

arvidn:
there's still this build failure: https://travis-ci.org/arvidn/libtorrent/jobs/378876090#L1406

PR #2988, opened 2018-04-30
revert to use Boost variables  in CMakeLists.txt

aldenml:
See https://github.com/arvidn/libtorrent/pull/2938/files#r185007457 for more information

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -617,7 +617,8 @@ endif()
 
 # Boost
 find_package(Boost REQUIRED COMPONENTS system)
-target_link_libraries(torrent-rasterbar PUBLIC Boost::system)
+target_include_directories(torrent-rasterbar PUBLIC ${Boost_INCLUDE_DIRS})
+target_link_libraries(torrent-rasterbar PUBLIC ${Boost_SYSTEM_LIBRARY})
 list(APPEND _package_dependencies "Boost COMPONENTS system")
 
 if (WIN32)
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -88,10 +88,11 @@ target_include_directories(python-libtorrent
 		${PYTHON_INCLUDE_DIRS}
 )
 
+string(TOUPPER "${boost-python-module-name}" boost_python_module_name_uppercase)
 target_link_libraries(python-libtorrent
 	PRIVATE
 		torrent-rasterbar
-		Boost::${boost-python-module-name}
+		${Boost_${boost_python_module_name_uppercase}_LIBRARY}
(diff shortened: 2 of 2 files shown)


arvidn:
isn't it odd that `Boost::system` doesn't work though? that library has been around for almost 10 years!

aldenml:
yes, it's odd, not sure if having boost from source (and not installed via brew) is a factor. I can hack my environment, but this way seems to works in any situation.

arvidn:
does the `Boost_LIBRARIES` variable contain every boost library?

arvidn:
it looks like `Boost_LIBRARIES` aren't *targets*, and that you have to add the include directory manually as well.

aldenml:
I believe it's only the ones resolved by `find_package(Boost REQUIRED COMPONENTS system)`

arvidn:
https://ci.appveyor.com/project/arvidn/libtorrent/build/8987/job/k0qf0mj7poaur8vh#L1121

aldenml:
I'm adding `target_include_directories(torrent-rasterbar PUBLIC ${Boost_INCLUDE_DIRS})` but I'm open to a better solution

evsh:
You might want to consider `Boost_SYSTEM_LIBRARY` variable as well as making the same changes in `bindings/python/CMakeLists.txt`

aldenml:
@evsh I appreciate some help with that (or even a PR to my branch) since I'm not familiar with the python build part, and I know very little of `cmake`.

evsh:
I suggest (please note those variable names are case sensitive):
```cmake
target_include_directories(torrent-rasterbar PUBLIC {Boost_INCLUDE_DIRS})
target_link_libraries(torrent-rasterbar PUBLIC ${Boost_SYSTEM_LIBRARY})
```
and then in `bindings/python/CMakeLists.txt` instead of `target_link_libraries(python-libtorrent Boost::${boost-python-module-name})` you write 
```cmake
string(TOUPPER "${boost-python-module-name}" boost_python_module_name_uppercase)
target_link_libraries(python-libtorrent ${Boost_${boost_python_module_name_uppercase}_LIBRARY})
```

You may want to use
```cmake
include(CMakePrintHelpers)
cmake_print_properties(TARGETS python-libtorrent
	PROPERTIES LINK_LIBRARIES)
```
to check afterwards that everything is OK.

aldenml:
interesting...I will explore that, thanks

evsh:
Sorry for breaking your workflow with those imported targets.

arvidn:
lgtm. would you mind squashing this?

aldenml:
sure, I will wait for the completion of CIs

PR #2994, opened 2018-05-01
Test merge

arvidn:


include/libtorrent/alert_manager.hpp
```diff
@@ -105,12 +105,7 @@ namespace libtorrent {
 		template <class T>
 		bool should_post() const
 		{
-			if (!(m_alert_mask.load(std::memory_order_relaxed) & T::static_category))
-			{
-				return false;
-			}
-
-			return should_post_impl(T::priority);
+			return bool(m_alert_mask.load(std::memory_order_relaxed) & T::static_category);
```
aldenml: I suppose this is an intentional relaxation, right?

arvidn: yes. it relies on there being some other synchronization of the threads, which should be a pretty safe assumption (it just doesn't feel very modular).

PR #2995, opened 2018-05-02
avoid unnecessary call to to_string()

aldenml:


src/file_storage.cpp
```diff
@@ -252,7 +252,7 @@ namespace {
 		executable_attribute = fe.executable_attribute;
 		symlink_attribute = fe.symlink_attribute;
 		no_root_dir = fe.no_root_dir;
-		set_name(fe.filename().to_string().c_str());
+		set_name(fe.filename().data());
```
arvidn: `set_name()` expects a 0-terminated string, `data()` returns a pointer to the first character, it does not 0-terminate it.

arvidn: I suppose `set_name()` could be made to take a string view. iirc, it copies the string anyway

aldenml:
I'm having second thoughts here, and this is not critical

PR #2997, opened 2018-05-02
Enhance CMake build

evsh:
1. Add tools to CMake build
1. Replace custom 'shared' option with the standard `BUILD_SHARED_LIBS`  flag
2. Replace `foreach()` calls for source items with `target_sources()` and the generator expression `$<JOIN:>`.
3. Remove build types definition: makes no sense for  single-configuration generators, and is populated automatically for multi-configuration ones.
4. Add feature summary, printed at the end of the configure stage. Example:
```
-- The following features have been enabled:

 * BUILD_SHARED_LIBS, build libtorrent as a shared library
 * build_tests, build tests
 * build_examples, build examples
 * build_tools, build tools
 * python-bindings, build python bindings
 * dht, enable support for Mainline DHT
 * encryption, Enables encryption in libtorrent
 * exceptions, build with exception support
 * extensions, Enables protocol extensions
 * i2p, build with I2P support
 * ipv6, build with IPv6 support
 * logging, build with logging

-- The following features have been disabled:

 * static_runtime, build libtorrent with static runtime

-- The following RECOMMENDED packages have been found:

 * Iconv, GNU encoding conversion library, <https://www.gnu.org/software/libiconv/>
   Convert strings between various encodings
 * OpenSSL, Full-strength general purpose cryptography library, <https://www.openssl.org/>
   Provides HTTPS support to libtorrent
```

5. Enahnce compiler flags management for static runtime by utilizing functions from the [ucm project](https://github.com/onqtam/ucm)
6. Copy almost all options from the Jamfile.
7. Use C++14 mode If compiler supports that.
8. Rename variables in FindIconv.cmake to match those in the same module  appeared in CMake 3.11.
9. Try to use config mode first while looking for the package in the example FindLibtorrentRasterbar.cmake.
10. Rename targets and config files for consistency:  CMake does check both suffixes anyway (-config and Config), but since the project name is "libtorrent-rasterbar", the "-config" suffix looks  better.
11. Install the FindLibtorrentRasterbar.cmake module.


CMakeLists.txt
```diff
@@ -633,24 +583,133 @@ if (WIN32)
...
+target_optional_compile_definitions(torrent-rasterbar PUBLIC FEATURE NAME extensions DEFAULT ON
+	DESCRIPTION "Enables protocol extensions" DISABLED TORRENT_DISABLE_EXTENSIONS)
+target_optional_compile_definitions(torrent-rasterbar PUBLIC FEATURE NAME i2p DEFAULT ON
+	DESCRIPTION "build with I2P support" DISABLED TORRENT_USE_I2P=0)
+target_optional_compile_definitions(torrent-rasterbar PUBLIC FEATURE NAME ipv6 DEFAULT ON
+	DESCRIPTION "build with IPv6 support" DISABLED TORRENT_USE_IPV6=0)
+target_optional_compile_definitions(torrent-rasterbar PUBLIC FEATURE NAME logging DEFAULT ON
+	DESCRIPTION "build with logging" DISABLED TORRENT_DISABLE_LOGGING)
+
+find_public_dependency(Iconv)
```
evsh: Maybe this call has to be made only on UNIX? From the other hand MinGW might use libiconv, if I'm not mistaken. `if (NOT MSVC)` perhaps?

CMakeLists.txt
```diff
@@ -538,89 +506,71 @@ elseif(MSVC)
...
+	PROPERTIES
+		CXX_VISIBILITY_PRESET "hidden"
+		VISIBILITY_INLINES_HIDDEN "true"
+		VERSION ${PROJECT_VERSION}
+		SOVERSION ${SOVERSION}
+)
+
+if (cxx_std_14 IN_LIST CMAKE_CXX_COMPILE_FEATURES)
+	target_compile_features(torrent-rasterbar PUBLIC cxx_std_14)
+	message(STATUS "Building in C++14 mode")
```
ssiloti: Is there any benefit to building as C++14? I don't recall if libtorrent enables any code based on C++ version. Unless there's a good reason to do this I'd rather stick with C++11 so that we catch any accidental use of C++14 or latter features.

aldenml: yes, there is, in `entry.hpp`, to use `strview_less`

CMakeLists.txt
ssiloti: In theory I think you could use iconv even with MSVC, but I'm not sure why you would want to since the Windows SDK provides the same functionality. Maybe make the dep `OPTIONAL` instead of `RECOMMENDED` on MSVC.

CMakeLists.txt
```diff
@@ -633,24 +583,133 @@ if (WIN32)
...
 
-if (NOT deprecated-functions)
-	target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_NO_DEPRECATE)
+if(OPENSSL_FOUND)
+	target_link_libraries(torrent-rasterbar PUBLIC OpenSSL::SSL)
+	target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_OPENSSL)
+	target_sources(torrent-rasterbar PRIVATE src/pe_crypto)
+else()
+	if (NOT WIN32 AND NOT APPLE)
+		target_sources(torrent-rasterbar PRIVATE src/sha1)
```
ssiloti: libtorrent can also use libgcrypt for sha1 so the internal implementation should only be linked in if neither openssl nor libgcrypt are being used.

CMakeLists.txt
```diff
@@ -633,24 +583,133 @@ if (WIN32)
...
+	target_sources(torrent-rasterbar PRIVATE
+		src/kademlia/$<JOIN:${kademlia_sources}, src/kademlia/>
+		ed25519/src/$<JOIN:${ed25519_sources}, ed25519/src/>
+		include/libtorrent/kademlia/$<JOIN:${libtorrent_kademlia_include_files}, include/libtorrent/kademlia/>
+		src/hasher512
+	)
+	target_include_directories(torrent-rasterbar PRIVATE ed25519/src)
+
+	if (NOT OpenSSL_FOUND AND NOT WIN32 AND NOT APPLE)
+		target_sources(torrent-rasterbar PRIVATE src/sha512)
```
ssiloti: This should also only be pulled in if gcrypt wasn't found either.

CMakeLists.txt
```diff
@@ -706,39 +788,49 @@ string(REGEX REPLACE "([^;]+)" "find_dependency(\\1)" _find_dependency_calls "${
 string(REPLACE ";" "\n" _find_dependency_calls "${_find_dependency_calls}")
 
 write_basic_package_version_file(
-	"${CMAKE_CURRENT_BINARY_DIR}/libtorrent-rasterbar/libtorrent-rasterbarConfigVersion.cmake"
+	"${CMAKE_CURRENT_BINARY_DIR}/libtorrent-rasterbar/libtorrent-rasterbar-config-version.cmake"
 	VERSION ${libtorrent_VERSION}
-	COMPATIBILITY ExactVersion
+	COMPATIBILITY SameMajorVersion
```
ssiloti: libtorrent breaks ABI compatibility with minor version changes, so this should be `SameMinorVersion`.

examples/cmake/FindLibtorrentRasterbar.cmake
```diff
@@ -10,83 +12,133 @@
...
-                                  Boost_SYSTEM_FOUND
-                                  Boost_RANDOM_FOUND)
-
-mark_as_advanced(LibtorrentRasterbar_INCLUDE_DIR LibtorrentRasterbar_LIBRARY
-    LibtorrentRasterbar_ORIG_CMAKE_FIND_LIBRARY_SUFFIXES
-    LibtorrentRasterbar_ENCRYPTION_INDEX)
+	set(LibtorrentRasterbar_LIBRARIES ${LibtorrentRasterbar_LIBRARY} ${CMAKE_THREAD_LIBS_INIT})
+	set(LibtorrentRasterbar_INCLUDE_DIRS ${LibtorrentRasterbar_INCLUDE_DIR})
+
+	if(NOT Boost_SYSTEM_FOUND OR NOT Boost_RANDOM_FOUND)
```
ssiloti: `OR NOT Boost_RANDOM_FOUND` should be removed.

CMakeLists.txt
evsh: Moreover, it is required to use the latest available standard, because tests in libtorrent headers rely on the `__cplusplus` macro version, and when you build a client with a newer standard, libtorrent declarations follow that, so must do implementations.

CMakeLists.txt
evsh: OK.

CMakeLists.txt
evsh: `SameMinorVersion` appeared in CMake 3.11.

ssiloti: Ah ok, that would explain why it was set to `ExactVersion` before. I would rather be too strict than too lenient here so I think it should stay `ExactVersion`.

ssiloti: lgtm

CMakeLists.txt
```diff
@@ -633,23 +583,138 @@ if (WIN32)
...
+				DESCRIPTION "A general purpose cryptographic library"
+				TYPE RECOMMENDED
+				PURPOSE "Use GCrypt instead of the built-in functions for RC4 and SHA1"
+		)
+		if (LibGcrypt_FOUND)
+			target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_LIBGCRYPT)
+			target_link_libraries(torrent-rasterbar PRIVATE LibGcrypt::LibGcrypt)
+		else()
+			if (NOT WIN32 AND NOT APPLE)
+				target_sources(torrent-rasterbar PRIVATE src/sha1)
```
ssiloti: sha1.cpp is needed even if encryption is disabled as long as there is no crypto library found and we're not building for WIN32 or APPLE.

evsh: Thanks! Carefully copied features from the Jamfile this time.

evsh:
@ssiloti, thanks for the review. Commits were squashed.

arvidn:
I already forgot what the deal with `bindings/python/setup.py` is. This change was made in the last patch, but it's still related. This file is updated/generated by running cmake, why do we need to have it in the repository to begin with?

arvidn:
I also note that the new `name` of the module in `setup.py` is `name='libtorrent'` where it used to be `name='python-libtorrent'`. Is this relevant or significant?

arvidn:
since `find_package()` of OpenSSL doesn't seem to work on mac (with openssl brew), the new logic will search for gcrypt and fail like this:

```
CMake Warning at CMakeLists.txt:455 (find_package):
  By not providing "FindLibGcrypt.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "LibGcrypt", but CMake did not find one.

  Could not find a package configuration file provided by "LibGcrypt" with
  any of the following names:

    LibGcryptConfig.cmake
    libgcrypt-config.cmake

  Add the installation prefix of "LibGcrypt" to CMAKE_PREFIX_PATH or set
  "LibGcrypt_DIR" to a directory containing one of the above files.  If
  "LibGcrypt" provides a separate development package or SDK, be sure it has
  been installed.
Call Stack (most recent call first):
  CMakeLists.txt:675 (find_public_dependency)
```

evsh:
> …why do we need to have it in the repository to begin with?

Distutils documentation says the the proffered way to build a cpython extension module is via `setup.py`. Then distutils ensure passing right compilation parameters, installing file into the right place, etc. Python world is so Python-centric...

> I also note that the new name of the module in setup.py is name='libtorrent' where it used to be name='python-libtorrent'. Is this relevant or significant?

The extension module is named 'libtorrent'. This name is used in the example python scripts. Why the egg should contain another name? 

> since find_package() of OpenSSL doesn't seem to work on mac (with openssl brew)…

One have to help CMake (Brew installs OpenSSL into a non-standard prefix, I believe). See here: https://github.com/qbittorrent/qBittorrent/blob/master/.travis.yml#L154. 

> … the new logic will search for gcrypt and fail like this

Ops… I'm sorry, I did not check where does cmake gcrypt module comes from on my system. And it comes from not from GCrypt itself but from KTorrent (and they borrowed it from Kopete). I have to import it here. Will do.

evsh:
Added FindLibGcrypt.cmake from KTorrent project.

CMakeLists.txt
```diff
@@ -705,40 +794,56 @@ set(ConfigPackageLocation ${CMAKE_INSTALL_LIBDIR}/cmake/libtorrent-rasterbar)
...
 	VERSION ${libtorrent_VERSION}
-	COMPATIBILITY ExactVersion
+	COMPATIBILITY ${_compatibility}
 )
 
-export(EXPORT torrent-rasterbarTargets
-	FILE "${CMAKE_CURRENT_BINARY_DIR}/libtorrent-rasterbar/libtorrent-rasterbarTargets.cmake"
+export(EXPORT libtorrent-rasterbar-targets
+	FILE "${CMAKE_CURRENT_BINARY_DIR}/libtorrent-rasterbar/libtorrent-rasterbar-targets.cmake"
 	NAMESPACE Libtorrent::
```
evsh: @arvidn, @ssiloti: there is a bit of inconvenience here: we export target "torrent-rasterbar" into namespace "Libtorrent". At the same time, package name of the library is "libtorrent-rasterbar". Thus the cmake find module is named "FindLibtorrentRasterbar.cmake". However, these export commands create imported target in "Libtorrent" namespace. For consistency "rasterbar" should be moved into the namespace name, and the full name of the exported target would become "LibtorrentRasterbar::torrent-rasterbar" then. 

Let's look at the library file names now. Consider Linux only. The CMake build creates output library named "libtorrent-rasterbar.so" At the same time, b2 build seems to create output "libtorrent.so" (at least I found no mention for "rasterbar" in the Jamfile), autotools build creates "libtorrent-rasterbar.la" (and .so, naturally). 

As you see, there are various naming schemes and I would suggest to unify them. My question is: what are the proper names for the package and output library?

ssiloti: Both the package and library should be named "libtorrent-rasterbar".

evsh: Since cmake default to camelCased names, "FindLibtorrentRasterbar.cmake" and target named "LibtorrentRasterbar::torrent-rasterbar", i.e. exporting to namespace "LibtorrentRasterbar" should be OK?

arvidn: Seems reasonable

evsh: OK, commited these changes.

arvidn:
on macOS, Iconv is found (`Iconv_FOUND` is set) but `ICONV_LIBRARIES` is empty. This results in libtorrent including iconv headers and calling those functions, but failing to link against iconv.

Furthermore, if I try to link against `Iconv::Iconv` it just breaks entirely. I don't think that target is set up by the find call.

Is there an important distinction between `ICONV_INCLUDE_DIR` and `ICONV_INCLUDE_DIRS`?
Looking here https://cmake.org/cmake/help/git-stage/module/FindIconv.html it seems like the plural form is preferred.

Btw, are variable names case insensitive?

Techically, the definition of `TORRENT_USE_ICONV` should be set to 1, not just defined. In the code it's used in `#if TORRENT_USE_ICONV` (i.e. not `#ifdef`).

I would think you'd have to change it along these lines:

```
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -648,10 +648,10 @@ set_package_properties(Iconv
                PURPOSE "Convert strings between various encodings"
 )

-if(Iconv_FOUND)
-       target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_ICONV)
-       target_include_directories(torrent-rasterbar PRIVATE ${ICONV_INCLUDE_DIR})
-       target_link_libraries(torrent-rasterbar PRIVATE ${ICONV_LIBRARIES})
+if(Iconv_LIBRARIES)
+       target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_ICONV=1)
+       target_include_directories(torrent-rasterbar PRIVATE ${Iconv_INCLUDE_DIR})
+       target_link_libraries(torrent-rasterbar PRIVATE ${Iconv_LIBRARIES})
 endif()

 find_public_dependency(OpenSSL)
```

But maybe that will break if iconv is built into libc.

arvidn:
also, when I'm enabling building the tools, I get this error:
```
/Users/arvid/Documents/dev/libtorrent/tools/parse_access_log.cpp:33:10: fatal error: 'libtorrent/config.hpp' file not found
         ^~~~~~~~~~~~~~~~~~~~~~~
```

Which is a bit weird. There is clearly a dependency on `torrent-rasterbar`, but I guess I don't fully understand how the `target_include_directories` work exactly on the `torrent-rasterbar` target. It doesn't seem to quite work though, at least not for the tools directory.

arvidn:
> Or we can raise minimum required cmake version to 3.11 and remove the bundled file.

I think it's fine to raise the requirement to the latest version. It's easy to get the latest version on most platforms anyway.

> However, since this tool does not link to the library, it makes sense to introduce an additional interface target for includes only, say torrent-includes, make torrent-rasterbar depend on it, and then use it for the tool. Would you agree?

Is there a benefit of doing this other than possibly not have to build libtorrent or depend on it?

I think it's fine to depend on the whole built library even if it isn't used, mostly to keep the CMakeList.txt simpler, and this tool isn't that important anyway.

evsh:
> I think it's fine to raise the requirement to the latest version.

OK.

> Is there a benefit of doing this other than possibly not have to build libtorrent or depend on it?

I don't know what happens on other platforms, but on Linux linker does not strip unneeded libraries and the executable loads libtorrent-rasterbar.so.10 and all its dependencies, and to avoid that I proposed to add an interface-only target. But of course, the executable is not that important on itself, and the dependencies question is even less important.

evsh:
Updated: raised minimum cmake version to 3.11, fixed tools build.

arvidn:
I still don't understand why you think bindings/python/setup.py needs to be checked in. You say python likes to build packages using distutils, but the cmake file generates this file, to (I assume) be able to build or package the python module using distutils. Why does it need to exist before running cmake?

If the setup.py generated by cmake is broken somehow, why generate it at all?
If it works just fine, then it seems redundant to have a slightly different version as well.

But perhaps more importantly, if there is a genuine use-case to have both, they should probably have different names. It's quite inconvenient to have the builds script mutate a file that's under version control.

arvidn:
you probably need to pull down `cmake` with `brew install cmake` on mac.

https://travis-ci.org/arvidn/libtorrent/jobs/378108068#L4748

evsh:
>  I still don't understand why you think bindings/python/setup.py needs to be checked in.

This file has nothing to do with the CMake build.

> …but the cmake file generates this file…

No, it generates another file with different content and not to build the extension module.

> If the setup.py generated by cmake is broken somehow,…

It does not build the extension module, just generates egg-info. The file you are referring to is used in autotools build. 

> But perhaps more importantly, if there is a genuine use-case to have both, they should probably have different names. It's quite inconvenient to have the builds script mutate a file that's under version control.

Agree, but on the other hand, this name clearly shows what the file is for, and CMake builds are out-fo-source anyway (mostly), thus the generated file does not interfere with any source file. Regardless, do you have a suggestion how to name the template and the generated files, please?

> you probably need to pull down cmake with brew install cmake on mac.

The quoted build fail is because Travis installs and outdated cmake in Linux container into /usr/local/cmake-3.9.2, and cmake from a PPA is installed into /usr, so I blindly wrote /usr/bin/cmake in the common part of the script, thus breaking the OSX build, because despite no activated cmake on OSX, there is "cmake --version" call there. I'm waiting for the AppVeyour build to complete before pushing the fix.

evsh:
> you probably need to pull down cmake with brew install cmake on mac.

Or did you want to enable cmake build on OSX too?

arvidn:
> Or did you want to enable cmake build on OSX too?

I thought that's what you tried to do, causing the build issue. I'm hoping that eventually cmake can replace the autotools build, and then I think it should be part of the CI for mac too

arvidn:
>> If the setup.py generated by cmake is broken somehow,…
>
> It does not build the extension module, just generates egg-info. The file you are referring to is used in autotools build.

The reason the cmake build generates it is so that it can package the module, right? And the module is built by cmake. That's my understanding, and if that works, I see no reason to also be able to build it some other way. For windows right now for example, `bjam` is used to build the binary, and then packaged using distutils.

evsh:
> then I think it should be part of the CI for mac too.

I would be happy to have it right away or you to replace ["branches only" in the .travis.yml](https://github.com/arvidn/libtorrent/blob/master/.travis.yml#L27). Both options give possibility to test cmake builds on OSX

> That's my understanding, and if that works, I see no reason to also be able to build it some other way.

Well, I use that to install bindings on my system and they seem to work. AFAIK nobody else tested the bindings module and installation.

evsh:
Pardon, but cmake build on OSX is absent. I was talking about the build system testing, and PR does not help me with that. Everything is OK, although, for code changes.

arvidn:
I think I'm a bit lost. When you say "build system testing" what do you refer to exactly?
I would imagine that introducing cmake to the Mac Travis build, one would simply add the necessary commands to the `.travis.yml` file and submit a PR. That would trigger a travis build and tell you what went wrong, then you can update the PR to fix it until it works.

As far as I can tell, the only thing removing the branches-only list in `.travis.yml` buys you, is it would let you do that without creating a PR.

evsh:
With these changes to the cmake build scripts, we encountered errors in build on OSX discovered by people, not a CI. I don't have Mac hardware and was unable to test that. Sure, I can make a commit in my fork that includes all the changes I need to trigger OSX build, various compilers and so on, then cherry-pick it on top of the feature branch, and remove it together with the WIP tag. 

I don't quite understand how do travis settings work, but it seems to me that you can disable building pushed branches using settings on travic-ci.org and remove those branch-only specification in the .travis.yml file, giving collaborators d freedom to set their travis builds up as they like. Because, yes, you are right, I would like to be able to use Travis without publishing a PR.

arvidn:
> it seems to me that you can disable building pushed branches using settings on travic-ci.org and remove those branch-only specification in the .travis.yml file,

Yeah, I believe you can.

> giving collaborators d freedom to set their travis builds up as they like. Because, yes, you are right, I would like to be able to use Travis without publishing a PR.

Wouldn't those settings on travis-ci.org still affect those branches? or is the distinction that it wouldn't affect other forks?

evsh:
I guess (can't  find that particular setting in the Travis documentation) the setting on the travis-ci.org controls only a single repository, so forks are unaffected.

evsh:
Great! I thank the reviewers!

PR #2999, opened 2018-05-02
[WIP] Hide thread implementation and header

vktr:
This is a simple change (I think) which moves the `#include <thread>` from `session.hpp` to `session.cpp`.

The reason is that I want to wrap libtorrent in C++/CLI and make it available to .NET, but C++/CLI doesn't support directly including headers which include `<thread>`.

I think there are a few other places where `<thread>` is included in the header, and if this change is something you'd consider accepting then I can make the change for the other headers as well.

### DIFF
--- a/src/session.cpp
+++ b/src/session.cpp
@@ -30,6 +30,8 @@ POSSIBILITY OF SUCH DAMAGE.
 
 */
 
+#include <thread>
+
 #include "libtorrent/config.hpp"
 #include "libtorrent/extensions/ut_pex.hpp"
 #include "libtorrent/extensions/ut_metadata.hpp"
@@ -60,6 +62,12 @@ namespace aux {
 	constexpr torrent_list_index_t session_interface::torrent_downloading_auto_managed;
 	constexpr torrent_list_index_t session_interface::torrent_seeding_auto_managed;
 	constexpr torrent_list_index_t session_interface::torrent_checking_auto_managed;
+
+	struct std_thread_impl : public std::thread
+	{
+	public:
+		using std::thread::thread;
+	};
 }
 
 #ifndef TORRENT_DISABLE_EXTENSIONS
@@ -369,7 +377,7 @@ namespace {
 		if (internal_executor)
 		{
 			// start a thread for the message pump
-			m_thread = std::make_shared<std::thread>(
+			m_thread = std::make_shared<aux::std_thread_impl>(
 				[&] { m_io_service->run(); });
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/session.cpp
```diff
@@ -60,6 +62,12 @@ namespace aux {
 	constexpr torrent_list_index_t session_interface::torrent_downloading_auto_managed;
 	constexpr torrent_list_index_t session_interface::torrent_seeding_auto_managed;
 	constexpr torrent_list_index_t session_interface::torrent_checking_auto_managed;
+
+	struct std_thread_impl : public std::thread
+	{
+	public:
```
arvidn: this `public:` is redundant

arvidn:
I'm in favour of hiding implementation details, especially if it decreases the header "footprint" of what's exposed. I'm not super happy with having to introduce a new type, but I can't think of a better way. Clearly both of those classes (`session` and `session_proxy`) must be defined in the header.

src/session.cpp
arvidn: actually, both `public` are redundant

aldenml:
My question is: is this the only problematic header? Which one are the others? I'm a little worry this is a start of a slippery slope. But again, at the time, I was worry with the typed indexed `aux::` containers.

vktr:
@aldenml This was the first header that gave me trouble and was an easy fix. I know <mutex> is included indirectly by alert_types.hpp which also makes C++/CLI complain. 

I’ll see if that is as easy as this and will push a fix if so.

I’m not looking to do major changes, and am in favor of keeping this to a minimum and then I can solve the the rest by having a wrapper library

arvidn:
@vktr Is this the extent of this patch? I got the impression you were going to complete it until `<thread>` is not included by any public header. Does it go for `<mutex>` too btw?

vktr:
Sorry, haven’t had the time to finish this. I’ll see if I can get to it
this weekend!

On Sat, May 12, 2018, at 13:54, Arvid Norberg wrote:
> @vktr[1] Is this the extent of this patch? I got the impression you
> were going to complete it until <thread> is not included by any public
> header. Does it go for <mutex> too btw?> — You are receiving this because you were mentioned. Reply to this
> email directly, view it on GitHub[2], or mute the thread[3].> 


Links:

  1. https://github.com/vktr
  2. https://github.com/arvidn/libtorrent/pull/2999#issuecomment-388549992

arvidn:
no rush, I just want to make sure we're not deadlocking, waiting on each other.

vktr:
Closing this due to menot needing the change - however if C++/CLI is something nice to have, then I can put some more energy in looking into what changes would be necessary.

PR #3004, opened 2018-05-05
Fix compilation with GCC 8

evsh:


include/libtorrent/lazy_entry.hpp
```diff
@@ -354,6 +354,9 @@ namespace libtorrent {
 			swap(m_len, e.m_len);
 		}
 
+		lazy_entry(lazy_entry&&) = default;
+		lazy_entry& operator=(lazy_entry&&) = default;
```
arvidn: could you explain this?

src/lazy_bdecode.cpp
```diff
@@ -266,7 +267,7 @@ namespace {
 			int const capacity = this->capacity() * lazy_entry_grow_factor / 100;
 			auto* tmp = new (std::nothrow) lazy_dict_entry[capacity + 1];
 			if (tmp == nullptr) return nullptr;
-			std::memcpy(tmp, m_data.dict, sizeof(lazy_dict_entry) * (m_size + 1));
+			std::move(&m_data.dict[0], &m_data.dict[m_size + 1], tmp);
```
arvidn: you have to apply +1 to the pointer, not the offset, otherwise you dereference the one-past-end slot

include/libtorrent/lazy_entry.hpp
evsh: Due to explicitly declared constructor and copy assignment (private), move constructor and move assignment will not be generated automatically. Hence the  declaration.

src/lazy_bdecode.cpp
evsh: `std::move()` operates with elements, not bytes. End iterator should point one past the last, and is never dereferenced.

include/libtorrent/lazy_entry.hpp
evsh: > I think after you move-from I think you need to call release(). I think there should be a test of moving a lazy_entry as well, to make sure there aren't any double-frees

Agree and will add a test.

src/lazy_bdecode.cpp
arvidn: `operator[]` forms a reference to the item at the given index. If that's past the end, I'm pretty sure that is considered a dereference and undefined behavior. if `m_data.dict` was a `std::vector` for instance, it would be invalid.

That said, why do you need the +1 at all? `m_size` is already the index of one-past-end, right?

arvidn: I would suggest you don't use `operator[]` and `operator&`, but just operate directly on the pointers

include/libtorrent/lazy_entry.hpp
arvidn: right, I'm suggesting that the automatically generated move constructor and move assignment may be incorrect

src/lazy_bdecode.cpp
arvidn: you could say that. I was referring to the definition of the expression `E1[E2]` is identical to `*((E1) + (E2))` [[expr.sub]](http://eel.is/c++draft/expr.sub). And the `operator*` is defined in [[expr.unary.op]](http://eel.is/c++draft/expr.unary#op-1) as:

> The unary * operator performs indirection:

I don't think the fact that you then apply the `operator&` avoids the fact that you've already invoked undefined behaviour.

evsh: > operator[] forms a reference to the item at the given index. If that's past the end, I'm pretty sure that is considered a dereference and undefined behavior.

Can't understand how can obtaining a reference to a element (i.e. loading object address into a register)  be considered dereferencing? Even should `m_data.dict` be an `std::vector` instance, `std::vector::operator[]` may not perform bounds checking.

> I was referring to the definition of the expression `E1[E2]` is identical to `*((E1) + (E2))`... 

I simply can't understand how can the description in [expr.sub](http://eel.is/c++draft/expr.sub) hold true, specifically the words " The result is of type “T”. Perhaps that is why I don't understand how does `&aPointer[anyPositiveNumber]`  have anything to do with dereferencing. Example:
```c++
int* p = nullptr;
printf("Pointer value is: %p\n", static_cast<void*>(&p[-reinterpret_cast<ptrdiff_t>(p)]));
```
This peacefully prints out '(nil)' as I would expect.

> That said, why do you need the +1 at all? `m_size` is already the index of one-past-end, right?

Because (you may clearly see that a few lines above, in the other `if` branch) the length of the allocated block is of `m_size + 1` . The first element contains the length value (equal to `m_size`) and `m_size` elements follow.  We have to move them all, i.e. `m_size + 1` elements, thus index of the last valid element is equal to `m_size`, and index of the one past the last one is `m_size + `1.

evsh:
PR updated: `lazy_entry` moving constructor and assignment operator operate as one would expect from a public function, a few move operations added to the `test_bencoding.cpp` file, although it crashes perfectly fine without that if the moving assignment does not clean the `other` object because of move calls during its normal operation.

src/torrent_info.cpp
```diff
@@ -1094,7 +1094,6 @@ namespace {
 			int const num_nodes = merkle_num_nodes(num_leafs);
 			m_merkle_first_leaf = num_nodes - num_leafs;
 			m_merkle_tree.resize(num_nodes);
-			std::memset(m_merkle_tree.data(), 0, aux::numeric_cast<std::size_t>(num_nodes * 20));
```
arvidn: is it not important that all the nodes in the merkle tree are cleared? I recognize that ``memset()`` isn't such a great way of clearing it, but now you're not clearing it at all

arvidn: I suppose if this is just called once, when `m_merkle_tree` is empty, it's fine. It will be cleared by `resize()`.

src/lazy_bdecode.cpp
arvidn: > I simply can't understand how can the description in expr.sub hold true, specifically the words " The result is of type “T”. Perhaps that is why I don't understand how does &aPointer[anyPositiveNumber] have anything to do with dereferencing

In short, it's true because it says so in the standard, which is kind of the definition. (assuming my reading is correct, that this *does* invoke undefined behavior) Your example just illustrates one valid outcome of undefined behaviour. Compilers don't have to do anything special when you invoke it, they are just allowed to assume that you don't.

src/torrent_info.cpp
evsh: Yes, I believe the call to `resize()` just above clears the elements.

evsh:
Thanks!

PR #3007, opened 2018-05-06
Fix unit tests build

d-komarov:
Export `recursive_mutex` because `emplace_alert` template injects
dependency on that type during instatiation.


d-komarov:
Fixes `test_alert_manager` build.

arvidn:
I wonder why this was passing on appveyor.

d-komarov:
Static build configuration?

PR #3013, opened 2018-05-09
remove redundant set_piece_hashes template. take a std::function instead

arvidn:


include/libtorrent/create_torrent.hpp
```diff
@@ -408,8 +408,8 @@ namespace detail {
 		set_piece_hashes(t, p, detail::nop, ec);
 		if (ec) throw system_error(ec);
 	}
-	template <class Fun>
-	void set_piece_hashes(create_torrent& t, std::string const& p, Fun f)
+	inline void set_piece_hashes(create_torrent& t, std::string const& p
+		, std::function<void(piece_index_t)> const& f)
```
evsh: Don't understand why is f `const` here, but not in line 444?

arvidn: it's a `const&` here because this function does not need a copy of the function object, so it just takes a reference. The reason it *isn't* a reference in the other case is that it's deprecated and I figured I might as well leave it as backwards compatible as possible.

evsh: Thanks for the clarification!

PR #3014, opened 2018-05-10
don't specialize standard type trait is_integral

### LINKED ISSUE #3012: build errors with VS 2017 15.7

platform/architecture: Windows x86/x64

compiler and compiler version: msvc 14.14

With the new version of msvc compilation fails with:

```
libtorrent\src\torrent.cpp(4903): error C2338: we need assert p.second >= dont_download
libtorrent\src\torrent.cpp(4913): error C2338: we need additional condition: p.second < dont_download
libtorrent\src\torrent.cpp(4956): error C2338: we need assert prio >= dont_download
```

The problem is related to the specialization of `std::is_intregral` for `strong_typedef`. See [this issue](https://developercommunity.visualstudio.com/content/problem/247889/vs-2017-1570-type-traits-behaviour-changed.html) on MS's support forum. MSVC's implementation of `std::is_unsigned` uses `std::is_intregral_v` internally. Note the solution pointing out a clause in the standard stating that user specializations of standard type traits are generally undefined behavior. I don't see anything in the standard which excepts `is_intregral` from this.

ssiloti:
Concerning type traits, the C++ standard states: "The behavior of a program
that adds specializations for any of the templates defined in this subclause is
undefined unless otherwise specified."

There is no exception for is_integral.

Fixes #3012

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -4900,7 +4900,7 @@ bool is_downloading_state(int const st)
 		bool const was_finished = is_finished();
 		for (auto const& p : pieces)
 		{
-			static_assert(std::is_unsigned<decltype(p.second)>::value
+			static_assert(std::is_unsigned<decltype(p.second)::underlying_type>::value
 				, "we need assert p.second >= dont_download");
 			TORRENT_ASSERT(p.second <= top_priority);
 			TORRENT_ASSERT(p.first >= piece_index_t(0));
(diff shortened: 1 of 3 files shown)


src/torrent.cpp
```diff
@@ -4900,7 +4900,7 @@ bool is_downloading_state(int const st)
 		bool const was_finished = is_finished();
 		for (auto const& p : pieces)
 		{
-			static_assert(std::is_unsigned<decltype(p.second)>::value
+			static_assert(std::is_unsigned<decltype(p.second)::underlying_type>::value
```
arvidn: I would have thought you need to say `typename` here (to indicate `underlying_type` is a type, and not a value)

PR #3020, opened 2018-05-11
restore aspects of the previous behavior of force-proxy

arvidn:
Instead of not opening any TCP listen sockets at all, open them but reject any incoming connection. This is because in RC_1_1, the UDP and TCP sockets are tied in subtle and unintuitive ways. This is much cleaner in the next major release and this patch will not need to be merged

### DIFF
--- a/simulation/test_session.cpp
+++ b/simulation/test_session.cpp
@@ -79,8 +79,9 @@ TORRENT_TEST(force_proxy)
 	// create session
 	std::shared_ptr<lt::session> ses = std::make_shared<lt::session>(pack, *ios);
 
-	// disable force proxy in 3 seconds (this should make us open up listen
-	// sockets)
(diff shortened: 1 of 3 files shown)


simulation/test_session.cpp
```diff
@@ -79,8 +79,9 @@ TORRENT_TEST(force_proxy)
 	// create session
 	std::shared_ptr<lt::session> ses = std::make_shared<lt::session>(pack, *ios);
 
-	// disable force proxy in 3 seconds (this should make us open up listen
-	// sockets)
+	// disable force proxy in 3 seconds. this won't make us open up listen
+	// sockets, since force proxy rejects incoming connections, it doesn't
+	// precent the listen sockets from opening
```
sledgehammer999: Typo. "prevent"

PR #3024, opened 2018-05-12
fix uTP and SOCKS5 simulations and update libsimulator

arvidn:


include/libtorrent/utp_stream.hpp
```diff
@@ -300,12 +299,6 @@ struct TORRENT_EXTRA_EXPORT utp_stream
 	template <class Handler>
 	void async_connect(endpoint_type const& endpoint, Handler const& handler)
 	{
-		if (!is_v4(endpoint))
-		{
-			m_io_service.post(std::bind<void>(handler, boost::asio::error::operation_not_supported));
-			return;
-		}
```
arvidn: @ssiloti I was surprised to find this. I can't recall any good reason to block IPv6 uTP connections, can you?
When enabling IPv6 and the commented out simulation, it failed because the path MTU discovery didn't work (which I fixed in the sim and updated below)

simulation/test_transfer.cpp
```diff
@@ -218,7 +218,8 @@ TORRENT_TEST(no_proxy_utp_ipv6)
 {
 	using namespace lt;
 	run_test(
-		[](lt::session&, lt::session&) {},
+		[](lt::session& ses0, lt::session& ses1)
+		{ utp_only(ses0); utp_only(ses1); },
```
arvidn: utp is blocked by default by the test, so both sessions need utp re-enabled for any uTP test.

include/libtorrent/utp_stream.hpp
ssiloti: No, we should definitely be able to support uTP over IPv6. This code is very old (committed in November of 2010) so I can only assume it is a vestigial workaround for some hopefully long since resolved issue.

PR #3028, opened 2018-05-13
move export.hpp into aux_ folder

arvidn:
put assert symbols in libtorrent namespace


Col-blimp:
This patch gives you error when using:
`
assert.hpp(39): fatal error C1017: invalid integer constant expression`

libtorrent.lib itself builds fine.

arvidn:
could you provide any more details? I take it this is with msvc, right? appveyor seems to work though

Col-blimp:
I'm away from desktop at the minute but yes it's msvc and the error is completely non specific only stating line 39.

ssiloti:
It would help to know what version of msvc you're using and what you are trying to compile.

Col-blimp:
VS 15.7.

Issue is here:

capture

the three 'defined' are not defined so you can't perform arithmetic on them and defining them just throws other errors.

You can simulate this error by trying to change the path of `libtorrent_assert_log` somewhere in you're code.

arvidn:
you mean that none of `TORRENT_ASIO_DEBUGGING`, `TORRENT_PROFILE_CALLS` or `TORRENT_DEBUG_BUFFERS` are defined? I wouldn't expect any arithmetic to be performed on them. Where is that?

Are you sure the problem isn't that the `aux_/export.hpp` isn't terminated properly and it somehow expects a constant integer expression to follow.

Which configuration defines do you build with?

arvidn:
this should probably be reported in a separate issue though

Col-blimp:
Ill post this in a new issue, thanks.

PR #3031, opened 2018-05-14
minor code cleanup in torrent.cpp

aldenml:


src/torrent.cpp
```diff
@@ -2759,16 +2759,14 @@ bool is_downloading_state(int const st)
...
-				for (peer_iterator i = m_connections.begin();
-					i != m_connections.end(); ++i)
+				for (auto c : m_connections)
 				{
 					TORRENT_INCREMENT(m_iterating_connections);
-					peer_connection* p = *i;
-					boost::optional<piece_block_progress> pbp = p->downloading_piece_progress();
-					if (pbp && pbp->bytes_downloaded > 0)
+					auto const pbp = c->downloading_piece_progress();
+					if (pbp.bytes_downloaded > 0)
```
arvidn: Don’t you still need to check that the optional is set?

aldenml: `downloading_piece_progress` does not return an optional since last year in `master`

PR #3032, opened 2018-05-14
fixed signed/unsigned warnings with latest version of Xcode clang

aldenml:


src/kademlia/dht_storage.cpp
```diff
@@ -112,7 +112,7 @@ namespace {
 		int const size = int(buf.size());
```
arvidn: If size becomes size_t here instead, maybe you get fewer casts

src/lazy_bdecode.cpp
```diff
@@ -436,7 +436,7 @@ namespace {
 		else if (int(m_size) == this->capacity())
 		{
 			int const capacity = this->capacity() * lazy_entry_grow_factor / 100;
```
arvidn: Perhaps capacity should be size_t here too. There is a miltiplication, so widening might make sense

src/kademlia/dht_storage.cpp
aldenml: OK

src/lazy_bdecode.cpp
aldenml: OK

src/kademlia/dht_storage.cpp
aldenml: well, not really, because `item.size` is `int`, it's about the same, but I can use `buf.size()` if you think it looks better, you tell me

PR #3039, opened 2018-05-16
fix address of point-to-point interfaces

### LINKED ISSUE #3038: Binding to a P-t-P interface shows wrong IP in error message

platform/architecture: Gentoo Linux/amd64

compiler and compiler version: GCC 8

Seems like libtorrent incorrectly determines local IP for a P-t-P tunnel interface:
```
client.py -i tun0 -o tun0 -s /tmp 'magnet:?xt=urn:btih:c1e797e823900db479b3ec2fb3bd18fe44c89f30&dn=archlinux-2018.05.01-x86_64.iso&tr=udp://tracker.archlinux.org:6969&tr=http://tracker.archlinux.org:6969/announce'
```
shows in console:
```
listening on 10.25.0.5:6881 (device: tun0) failed: [sock_bind_to_device] [TCP] Operation not permitted
```
but:
```
$ ip -4 addr show dev tun0
4: tun0: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UNKNOWN group default qlen 100
    inet 10.25.0.6 peer 10.25.0.5/32 brd 10.25.0.6 scope global noprefixroute tun0
       valid_lft forever preferred_lft forever
```

ssiloti:
Fixes #3038

### DIFF
--- a/src/enum_net.cpp
+++ b/src/enum_net.cpp
@@ -352,6 +352,7 @@ namespace {
 			}
 		}
 
+		ip_info->interface_address = address();
 		int rt_len = int(IFA_PAYLOAD(nl_hdr));
 #ifdef __clang__
 #pragma clang diagnostic push
--- a/src/enum_net.cpp
+++ b/src/enum_net.cpp
@@ -363,6 +363,13 @@ namespace {
 			switch(rt_attr->rta_type)
 			{
 			case IFA_ADDRESS:
+				// if this is a point-to-point link then IFA_LOCAL holds
+				// the local address while IFA_ADDRESS is the destination
+				// don't overwrite the former with the latter
+				if (!ip_info->interface_address.is_unspecified())
+					break;
+				BOOST_FALLTHROUGH;
+			case IFA_LOCAL:
 #if TORRENT_USE_IPV6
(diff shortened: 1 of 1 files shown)


src/enum_net.cpp
```diff
@@ -363,6 +363,13 @@ namespace {
 			switch(rt_attr->rta_type)
 			{
 			case IFA_ADDRESS:
+				// if this is a point-to-point link then IFA_LOCAL holds
+				// the local address while IFA_ADDRESS is the destination
+				// don't overwrite the former with the latter
+				if (!ip_info->interface_address.is_unspecified())
```
arvidn: the only thing that makes me a little bit nervous is that this function now has a new (subtle) precondition that this `interface_address` field be in a default constructed state or cleared somehow.
Perhaps we should just clear this field at the top of the function, to simplify the interface.

arvidn:
does this not apply to `RC_1_1`?

ssiloti:
RC_1_1 doesn't use netlink to enumerate network interfaces.

evsh:
Thanks!

PR #3043, opened 2018-05-18
refactor to use the utility function for bind_socket_to_device()

arvidn:
instead of using the bind_to_device socket option directly, to make it best effort.

### DIFF
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -1326,6 +1326,7 @@ TORRENT_VERSION_NAMESPACE_2
 		udp_error_alert(
 			aux::stack_allocator& alloc
 			, udp::endpoint const& ep
+			, operation_t op
 			, error_code const& ec);
 
 		TORRENT_DEFINE_ALERT(udp_error_alert, 46)
@@ -1335,6 +1336,9 @@ TORRENT_VERSION_NAMESPACE_2
 
 		// the source address associated with the error (if any)
 		aux::noexcept_movable<udp::endpoint> endpoint;
+
+		// the operation that failed
+		operation_t operation;
 
 		// the error code describing the error
 		error_code const error;
--- a/include/libtorrent/enum_net.hpp
+++ b/include/libtorrent/enum_net.hpp
@@ -105,7 +105,7 @@ namespace libtorrent {
 	// TODO: 3 use string_view for device_name
 	template <class Socket>
 	address bind_socket_to_device(io_service& ios, Socket& sock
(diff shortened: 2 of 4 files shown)


ssiloti:
The behavior of `bind_socket_to_device` when bind_to_device fails is not what we want here. For listen sockets `bind_ep` has already been selected in `reopen_listen_sockets`. To use `bind_socket_to_device` here it needs to be able to take `bind_ep` as a parameter.

arvidn:
@evsh I'm eager to hear back :)

evsh:
The patch did not help, there is no download activity.

arvidn:
Interesting. Any relevant error alerts? Bind to device failures shouldn’t prevent the socket from being opened now. However, even if the listen sockets were failing earlier, outgoing sockets should still have allowed for downloading. I suppose the outgoing sockets really are the main problem.

Do you have any observations that hint of what’s going on?

arvidn:
(Please double check you ended up applying and linking the patched binary correctly too)

evsh:
The log contains:
```
listening on <correct ipv4 address>:6881 (device: tun0) failed: [sock_bind] [UDP] Operation not permitted
listening on [<correct ipv6 address>%tun0]:6881 (device: tun0) failed: [sock_bind] [UDP] Operation not permitted
UDP error: Operation not permitted from: <correct ipv4 address>
UDP error: Operation not permitted from:<correct ipv6 address>%tun0
UDP error: Operation not permitted from: <correct ipv4 address>
UDP error: Operation not permitted from: <correct ipv6 address>%tun0
```
Downloading starts if I pass outgoing interface only; console output contains the same error messages.

> (Please double check you ended up applying and linking the patched binary correctly too)

Not sure what did you mean here, but I need to just drop a file with patch into a predefined directory, and the Gentoo package management system applies it when building and installing a package. It reports back that that the patch was applied.

Does the `client.py -i <interface name> -o <interface name>` work for you?

evsh:
`client.py -o <interface name>` (omitting the listen_interfaces option) make it work.

evsh:
> I could hard-code it to fail and see what happens I suppose.

I would appreciate that.

arvidn:
it's interesting that you get those `UDP Error:` messages. That suggests that it's not just the socket option that fails, but subsequent attempts to read from the socket also fail.

arvidn:
oh, also. It's not the `BIND_TO_DEVICE` socket option that's failing. it's `bind()`.

arvidn:
please test again now

evsh:
It works! I'm very grateful!

PR #3049, opened 2018-05-20
only start a new accept request on new listen sockets

ssiloti:
Existing sockets already have an outstanding accept.
Also fix mapping new listen sockets when re-mapping is not requested.

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -1895,6 +1895,12 @@ namespace aux {
 				if (m_dht)
 					m_dht->new_socket(m_listen_sockets.back());
 #endif
+
+				TORRENT_ASSERT((s->incoming == duplex::accept_incoming) == bool(s->sock));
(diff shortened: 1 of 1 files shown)


src/session_impl.cpp
```diff
@@ -1895,6 +1895,12 @@ namespace aux {
 				if (m_dht)
 					m_dht->new_socket(m_listen_sockets.back());
 #endif
+
+				TORRENT_ASSERT((s->incoming == duplex::accept_incoming) == bool(s->sock));
+				if (s->sock) async_accept(s->sock, s->ssl);
+				// since this is a new socket it needs to map ports
+				// even if the caller did not request re-mapping
+				if (!map_ports) remap_ports(remap_natpmp_and_upnp, *s);
```
arvidn: the comment says "even if the caller did not request re-mapping", but the code says: "only if the caller did not request remapping".

I see, the logic to remap all ports id `map_ports` is true remains at the end of the function.

PR #3051, opened 2018-05-21
set the hidden attribute when creating the part file

arvidn:


src/part_file.cpp
```diff
@@ -212,7 +212,7 @@ namespace libtorrent
 
 		int slot = i->second;
 
-		open_file(file::read_write, ec);
+		open_file(file::read_writ | file::attribute_hiddene, ec);
```
aldenml: i bet you already notice the typo :)

arvidn:
please test now

sledgehammer999:
I confirm that it works now.

PR #3054, opened 2018-05-21
some cleanup and refactor in torrent_info code

aldenml:


src/torrent_info.cpp
```diff
@@ -324,9 +324,9 @@ namespace {
 		bdecode_node const attr = dict.dict_find_string("attr");
 		if (attr)
 		{
-			for (int i = 0; i < attr.string_length(); ++i)
+			for (char const c : attr.string_value())
```
arvidn: this makes me a little bit nervous. This code is a bit sensitive because there are torrents out there with tens of thousands of files, and sessions with thousands of torrents that need loading on startup. I believe this will construct a `std::string`, no?
Isn't there a function to get a `string_view`?

aldenml: don't be :)
`bedecode::string_value()` returns a `string_view`

src/torrent_info.cpp
```diff
@@ -1239,18 +1238,16 @@ namespace {
...
 		{
-			bdecode_node link = torrent_file.dict_find_string("magnet-uri");
-			if (link)
+			bdecode_node const uri = torrent_file.dict_find_string("magnet-uri");
+			if (uri)
 			{
-				auto uri = link.string_value();
-
-				add_torrent_params p = parse_magnet_uri(uri.to_string(), ec);
+				add_torrent_params p = parse_magnet_uri(uri.string_value(), ec);
```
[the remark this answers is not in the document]
aldenml: same, `string_view bdecode_node::string_value() const`
I will add the `const`

PR #3059, opened 2018-05-25
Reflect Boost version and C++11/14 interconnection in CMake build

### LINKED ISSUE #2966: Should require C++11/14 when building against boost >= 1.66
Since https://github.com/arvidn/libtorrent/commit/d1a63cd781e79e1e3792a258c0bc7f9455f2272f, building libtorrent-rasterbar with boost >= 1.66 actually requires C++11 or newer standard due to the use of "auto", which seems to be hard to workaround due to different return types from "buffer_sequence_begin". Ideally one can check whether boost version is 1.66 or newer and add "-std=gnu++0x" (or appropriate flag) to ${CXXFLAGS}, but it can be tricky to implement in m4. Since some of the ${CXXFLAGS} (e.g. -ftemplate-depth=n) requires GCC 4.5 which had sufficient support, maybe it's easier to directly enable C++11 all together. I can help with PR if this is acceptable.

This was discovered when building libtorrent-rasterbar 1.1.6 on OpenBSD, which uses gcc 4.9.4 which still defaults to "-std=gnu++98".

libtorrent version (or branch): master & RC_1_1

platform/architecture: all/OpenBSD

compiler and compiler version: GCC 4.9.4

how to reproduce it.

More like a wish list.

### LINKED ISSUE #3011: Unexpected Boost dependencies when compile in c++11 mode
libtorrent version (or branch): RC_1_1

When c++11 is used libtorrent use std::chrono and std::random, but CMake still require boost::chrono and boost::random. Can you make these requirements to be conditional?

evsh:
C++ standard and Boost requirements are connected:
1. With C++11 onward, we require Boost system component, with
C++03 we need chrono and random components too
2. When building against boost 1.66 and newer, C++11 is required.

Closes #3011, partially closes #2966.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.6)
+cmake_minimum_required(VERSION 3.8)
 project(libtorrent)
 set (SOVERSION "8")
 set (VERSION "1.1.7")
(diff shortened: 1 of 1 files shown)


arvidn:
Wouldn't you need to reflect which version of C++ libtorrent was built with in the `pkconfig` file in order for this to work across a binary-install and an independent client build?

evsh:
Important point! Addressed, thank you!

evsh:
Thanks for the review!

PR #3061, opened 2018-05-25
remove use of deprecated function readdir_r

evsh:
Ports 140b8ace onto RC_1_1 branch

Compilation with GCC 8 fails without such change.

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -1185,9 +1185,6 @@ namespace libtorrent
 		}
 #else
 
-		memset(&m_dirent, 0, sizeof(dirent));
-		m_name[0] = 0;
-
 		// the path passed to opendir() may not
 		// end with a /
 		std::string p = path;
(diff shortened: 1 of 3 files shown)


src/file.cpp
```diff
@@ -1185,7 +1185,6 @@ namespace libtorrent
 		}
 #else
 
-		memset(&m_dirent, 0, sizeof(dirent));
 		m_name[0] = 0;
```
arvidn: I notice you left this line in (`m_name[0] = 0;`) while it was removed in master. is it necessary in `RC_1_1`?

evsh: Overlooked, sorry. Of course, it is a mistake to assign to a empty `std::string`.

evsh: Assignment removed.

evsh:
thanks!

PR #3062, opened 2018-05-25
Fix pkg-config branch in the example CMake find module

evsh:
1. The imported target properties may not contain -D switches in the
INTERFACE_COMPILE_DEFINITIONS property, because CMake will always
prepend them to each item. Therefore we have to remove those switches
and also pass other options (e.g. -fexceptions) in INTERFACE_COMPILE_OPTIONS
property, otherwise CMake creates "mutants" '-D-fexceptions' alike.

2. FindOpenSSL.cmake variable for include directory is
OPENSSL_INCLUDE_DIR, not OPENSSL_INCLUDE_DIRS.

### DIFF
--- a/examples/cmake/FindLibtorrentRasterbar.cmake
+++ b/examples/cmake/FindLibtorrentRasterbar.cmake
@@ -108,7 +108,7 @@ else()
 	if(LibtorrentRasterbar_ENCRYPTION_INDEX GREATER -1)
 		find_package(OpenSSL QUIET REQUIRED)
 		set(LibtorrentRasterbar_LIBRARIES ${LibtorrentRasterbar_LIBRARIES} ${OPENSSL_LIBRARIES})
-		set(LibtorrentRasterbar_INCLUDE_DIRS ${LibtorrentRasterbar_INCLUDE_DIRS} ${OPENSSL_INCLUDE_DIRS})
+		set(LibtorrentRasterbar_INCLUDE_DIRS ${LibtorrentRasterbar_INCLUDE_DIRS} ${OPENSSL_INCLUDE_DIR})
 		set(LibtorrentRasterbar_OPENSSL_ENABLED ON)
 	endif()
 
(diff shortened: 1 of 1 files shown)


evsh:
PR Updated:

2. FindOpenSSL.cmake variable for include directory is
OPENSSL_INCLUDE_DIR, not OPENSSL_INCLUDE_DIRS.

examples/cmake/FindLibtorrentRasterbar.cmake
```diff
@@ -128,13 +128,23 @@ else()
...
 		add_library(LibtorrentRasterbar::torrent-rasterbar SHARED IMPORTED)
 
+		# LibtorrentRasterbar_DEFINITIONS var contains a mix of -D, -f, and possible -std options
+		# let's split them into definitions and options (that are not definitions)
+		set(LibtorrentRasterbar_defines "${LibtorrentRasterbar_DEFINITIONS}")
+		set(LibtorrentRasterbar_options "${LibtorrentRasterbar_DEFINITIONS}")
+		list(FILTER LibtorrentRasterbar_defines INCLUDE REGEX "(^|;)-D.+")
+		list(FILTER LibtorrentRasterbar_options EXCLUDE REGEX "(^|;)-D.+")
+		# remove '-D' from LibtorrentRasterbar_defines
+		string(REGEX REPLACE "(^|;)(-D)" "\\1" LibtorrentRasterbar_defines "${LibtorrentRasterbar_defines}")
```
arvidn: this all looks very GCC specific. I suppose msvc (or any non GCC-compatible compilers) don't run on a system that has `pkgconfig` (or does it?). Is matching `-D` (like a caveman) really the best/only way to know what's a preprocessor definition and not?

arvidn: I guess *we* generate this string, and we use `-D` prefixes for the `pkgconfig` script

evsh: That's correct, although there might be a straightforward way to learn those switches (-I and -D) for the current compiler.

evsh:
Thanks!

examples/cmake/FindLibtorrentRasterbar.cmake
evsh: I don't know where can I find an overview of command line switches for existing compilers, but checking CMake sources show that for all known to CMake C/C++ compilers (ARMCC, Bruce, Cray, GNU, IAR, Intel, NVIDIA, PGI, PathScale, SunPro, TinyCC, XL) it uses `-D`.

PR #3063, opened 2018-05-26
cmake: pass target C++ standard to pkgconfig generator

evsh:


cmake/Modules/GeneratePkgConfig.cmake
```diff
@@ -70,6 +89,7 @@ function(_expand_targets _targets _libraries_var _include_dirs_var _compile_opti
 				_get_target_property_merging_configs(_iface_include_dirs ${_dep} INTERFACE_INCLUDE_DIRECTORIES)
 				_get_target_property_merging_configs(_iface_compile_options ${_dep} INTERFACE_COMPILE_OPTIONS)
 				_get_target_property_merging_configs(_iface_definitions ${_dep} INTERFACE_COMPILE_DEFINITIONS)
+				get_target_property(_iface_compile_features ${_dep} INTERFACE_COMPILE_FEATURES)
```
arvidn: is it deliberate that this does not have an underscore in front of it?

arvidn: I guess it's a different function as it also doesn't have the `_merging_configs` suffix either

evsh: Yep, this is the standard CMake function, and I use underscore to mark module's private functions.

evsh:
Thanks!

PR #3064, opened 2018-05-26
deliver notification of alerts being dropped via alerts_dropped_alert

arvidn:


sledgehammer999:
I suppose this is a good start to have some kind of diagnostic and at least make the program realize that there might be a problem.

sledgehammer999:
I can't think of any use case, but I wonder if it would be useful to also have a counter per dropped alert type.

test/test_alert_types.cpp
```diff
@@ -275,6 +276,17 @@ TORRENT_TEST(session_stats_alert)
 	TEST_CHECK(v->message().find("session stats (") != std::string::npos);
 }
 
+TORRENT_TEST(alert_name)
+{
+	TEST_EQUAL(alert_name(tracker_announce_alert::alert_type), "tracker_announce"_sv);
```
aldenml: I think this test can be included in the macro `TEST_ALERT_TYPE` of `TORRENT_TEST(alerts_types)` to cover all tests

src/alert.cpp
```diff
@@ -2540,6 +2540,83 @@ namespace {
...
 	}
 
+	alerts_dropped_alert::alerts_dropped_alert(aux::stack_allocator&
+		, std::bitset<num_alert_types> const& dropped)
+		: dropped_alerts(dropped)
+	{}
+
+	char const* alert_name(int const alert_type)
+	{
+		static std::array<char const*, num_alert_types> names = {{
```
aldenml: can this be `const`?

PR #3066, opened 2018-05-26
Add set_alert_notify to python bindings

aresch:


bindings/python/src/session.cpp
```diff
@@ -388,6 +388,22 @@ namespace
     }
 #endif // TORRENT_ABI_VERSION
 
+    object alert_notify_cb;
```
arvidn: if you have more than one session instance, this won't work. Instead, bind the python `object` into the function object. Like, `s.set_alert_notify(std::bind(&alert_notify, cb));`

and make `alert_notify()` take the `object` as a parameter.

aresch:
Yup, sorry, forgot about multiple sessions. I also added a test and cleaned up the test.py a bit since it made my linter go crazy. Let me know if you want me to split the clean-up into a separate PR.

bindings/python/test.py
```diff
@@ -593,69 +610,70 @@ def test_execute_client(self):
...
             if returncode != 0:
                 print("stdout:\n" + process.stdout.read().decode("utf-8"))
-            self.assertEqual(returncode, 0, "returncode: " + str(returncode) + "\n"
-                + "stderr: empty\n"
-                + "some configuration does not output errors like missing module members,"
-                + "try to call it manually to get the error message\n")
+            self.assertEqual(returncode, 0, "returncode: " + str(returncode) + "\n" +
+                             "stderr: empty\n" +
+                             "some configuration does not output errors like missing module members," +
+                             "try to call it manually to get the error message\n")
```
arvidn: is there some standard formatting convention that defines this?

arvidn:
I kind of think formatting changes (especially larger ones) should be isolated in separate PRs or at least commits

bindings/python/test.py
aresch: It looks like we should keep the original based on this: https://lintlyci.github.io/Flake8Rules/rules/W503.html

I guess flake8 has changed but I still got the warning in my linter. Ultimately, I wouldn't really do either but I figured just moving the operator would quiet the linter. This is how I would do it:

```python
self.assertEqual(returncode, 0, ("returncode: {}\n"
"stderr: empty\n"
"some configuration does not output errors like missing module members,"
"try to call it manually to get the error message\n".format(returncode))
```

aresch:
I've reverted the style changes and I can re-do them in a separate PR.

bindings/python/src/session.cpp
```diff
@@ -388,6 +388,20 @@ namespace
     }
 #endif // TORRENT_ABI_VERSION
 
+    void alert_notify(object cb)
+    {
+        if (cb)
```
arvidn: is it OK to check to see if `cb` is set without holding the GIL?

aresch: I'm not sure. Let me see what I can find out. I have read in the past that whenever you deal with Python objects you should always acquire the GIL, so maybe it would be prudent to just do that in this case. I do worry a bit about blocking the libtorrent main thread while waiting to acquire the lock though.

aresch: Oh whoops. We grab the gil anyways if cb is not None. I'll just move the lock to the top of the function just to be safe.

aresch:
Is there anything else that needs to be done before this can be merged?

PR #3069, opened 2018-05-28
Test merge

arvidn:
@evsh could you have a look to make sure I merged the `CMakeLists.txt` correctly?


evsh:
@arvidn, the changes are unneeded because the pkgconfig generating code in the master branch takes all the required information (including the c++ standard version) from target properties).

PR #3075, opened 2018-05-30
Clean-up python code

aresch:
Used a combination of autopep8 and output from flake8 to fix any linting errors.

```
$ flake8 --max-line-length=120 *.py
```

This comes back clean with these changes.


arvidn:
would it be hard to run that tool on travis?
I fear the formatting will diverge over time (again) otherwise

aresch:
I found a whole bunch of python in the tools/ and tests/ directories, so I'm going through and cleaning all of those files up and making sure they are Python 3 compatible. There were A LOT of lint errors though: 4000+.

aresch:
@arvidn I think this is ready for review now. Sorry for the huge change but the scope grew as I found more python in the tree.

Changes include:
* pep8 compliance with a 120 character line-limit
* python 2 and 3 compatible code
* travis CI testing both python 2 & 3 style
* fixed a couple of real bugs in the code that were highlighted by the linter (missing local variables or typos)

arvidn:
this still looks like a show-stopper: https://travis-ci.org/arvidn/libtorrent/jobs/387539907#L4704

.travis.yml
```diff
@@ -79,6 +80,12 @@ before_install:
   - ulimit -a
 
 install:
+  - sudo pip install flake8
+  - flake8 --version
+  - flake8 --max-line-length=120
+  - sudo pip3 install flake8
+  - python3 -m flake8 --version
+  - python3 -m flake8 --max-line-length=120
```
arvidn: I think you should wrap these commands under a condition, and only enable it on a single build (in the build matrix). Specifically, this doesn't seem to work as-is under macOS. But it's also a waste of time to run it 4 times in parallel. So, set a new environment variable, say: `flake8` and set it to true in one of the build configurations that currently is quick. Use the top one, the one with `sonar_scan=1`.

bindings/python/client.py
```diff
@@ -153,15 +153,16 @@ def print_download_queue(console, download_queue):
...
     atp = lt.add_torrent_params()
     if filename.startswith('magnet:'):
         atp = lt.parse_magnet_uri(filename)
     else:
         atp.ti = lt.torrent_info(filename)
         try:
-            at.resume_data = open(os.path.join(options.save_path, info.name() + '.fastresume'), 'rb').read()
-        except:
+            atp.resume_data = open(os.path.join(options.save_path, atp.info.name() + '.fastresume'), 'rb').read()
+        except BaseException:
```
arvidn: is this a style convention or is it a correctness issue?

bindings/python/client.py
```diff
@@ -225,13 +227,13 @@ def main():
...
-        'alert_mask': lt.alert.category_t.all_categories,
-        'outgoing_interfaces' : options.outgoing_interface
-    }
+    settings = {'user_agent': 'python_client/' + lt.__version__,
+                'listen_interfaces': '%s:%d' % (options.listen_interface, options.port),
+                'download_rate_limit': int(options.max_download_rate),
+                'upload_rate_limit': int(options.max_upload_rate),
+                'alert_mask': lt.alert.category_t.all_categories,
+                'outgoing_interfaces': options.outgoing_interface
+                }
```
arvidn: this formatting is enforced by flake8?

aresch:
Yea, I think I should put the linting behind an env variable like 'lint' so that we don't try to run it on every run.

docs/gen_reference_doc.py
```diff
@@ -35,690 +36,784 @@
...
-	if name in category_fun_mapping:
-		return category_fun_mapping[name]
+    if f in category_mapping:
+        return category_mapping[f]
 
-	if f in category_mapping:
-		return category_mapping[f]
+    return 'Core'
 
-	return 'Core'
```
arvidn: this additional blank line, does it serve a purpose?

docs/gen_reference_doc.py
```diff
@@ -35,690 +36,784 @@
...
-	return len(line) == 0
+    line = line.split('//')[0]
+    line = line.replace('{', '')
+    line = line.replace('}', '')
+    line = line.replace('[', '')
+    line = line.replace(']', '')
+    line = line.replace(';', '')
+    line = line.strip()
+    return len(line) == 0
+
```
arvidn: there are a few of these. does the formatting style require two blank lines between functions?

docs/gen_reference_doc.py
```diff
@@ -729,266 +824,311 @@ def consume_ifdef(lno, lines, warn_on_ifdefs = False):
...
+            'name'.ljust(
+                width[0]) +
+            ' | ' +
+            'value'.ljust(
+                width[1]) +
+            ' | ' +
+            'description'.ljust(
+                width[2]) +
+            ' |',
+            file=out)
```
arvidn: this line seems to have been chopped up in quite small pieces. was this deliberate?

docs/gen_reference_doc.py
```diff
@@ -729,266 +824,311 @@ def consume_ifdef(lno, lines, warn_on_ifdefs = False):
...
+                v['name'].ljust(
+                    width[0]) +
+                ' | ' +
+                v['val'].ljust(
+                    width[1]) +
+                ' | ' +
+                d[0].ljust(
+                    width[2]) +
+                ' |',
+                file=out)
```
arvidn: here too

tools/gprof2dot.py
```diff
@@ -35,24 +35,23 @@
 # Python 2.x/3.x compatibility
 if sys.version_info[0] >= 3:
```
arvidn: this is a 3rd party program. I'm not even sure this is hooked up to any of the other tools in here

.travis.yml
```diff
@@ -6,7 +6,7 @@ matrix:
   include:
     - if: repo = arvidn/libtorrent
       env: variant=release sonar_scan=1 toolset=gcc
-    - env: variant=test_debug lint=1 tests=1 toolset=gcc-sanitizer
+    - env: variant=test_debug lint=1 pylint=1 tests=1 toolset=gcc-sanitizer
```
arvidn: I would prefer to do this in the same job as `sonar_scan=1`, as it only takes 2-3 minutes right now

arvidn:
do you want to squash any of these commits?

bindings/python/setup.py
```diff
@@ -145,13 +153,13 @@ def target_specific():
...
     cfg_vars = get_config_vars()
-    for key, value in cfg_vars.items():
+    for key, value in list(cfg_vars.items()):
         if isinstance(value, str):
             cfg_vars[key] = value.replace('-Wstrict-prototypes', '')
 
     source_list = os.listdir(os.path.join(os.path.dirname(__file__), "src"))
     source_list = [os.path.abspath(os.path.join(os.path.dirname(__file__),
-                   "src", s)) for s in source_list if s.endswith(".cpp")]
+                                                "src", s)) for s in source_list if s.endswith(".cpp")]
```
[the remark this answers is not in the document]
aresch: Yea, it's because of where the split happens. I could clean this up a bit by maybe removing the list comprehension here as it's quite long.

bindings/python/client.py
aresch: Some behaviour changed with bare except clauses in newer versions of python, basically a bare except will now act as 'except Exception:' which has a different behaviour than 'except BaseException:' (there are a few exceptions that are a direct subclass of BaseException). In the past there was no BaseException and everything was a subclass of Exception.

In order to not change the behaviour of the code an except BaseException was added here by autopep8. See https://www.python.org/dev/peps/pep-0348/ for some more details on the change and the hierarchy of Exceptions.

In this case, it's probably best to just catch Exception which was probably the original intent. In general, it is discouraged to ever use a blank except as a convention of style.

https://docs.python.org/3.1/howto/doanddont.html#except

bindings/python/client.py
aresch: It looks like because the original statement split after the first line, the next lines were indented to align with the {. If we move the first item to a new line then we get something like this:

```python
    settings = {                                  
        'user_agent': 'python_client/' + lt.__version__,                                                              
        'listen_interfaces': '%s:%d' % (options.listen_interface, options.port),
        'download_rate_limit': int(options.max_download_rate),    
        'upload_rate_limit': int(options.max_upload_rate),                                                            
        'alert_mask': lt.alert.category_t.all_categories,           
        'outgoing_interfaces': options.outgoing_interface,               
    }                                                               
```
This passes the linter and looks better to me.

docs/gen_reference_doc.py
aresch: There should be 2 blank lines between top level classes and functions.

https://lintlyci.github.io/Flake8Rules/rules/E302.html

docs/gen_reference_doc.py
aresch: Yes, note that this only occurs with top-level functions and would not apply to functions within a class definition.

docs/gen_reference_doc.py
aresch: Nope, looks like autopep8 went a bit wild with this. It was probably trying to deal with the max-line-length.  I'll clean this up.

tools/gprof2dot.py
aresch: Yea, in hindsight maybe I should have just had the linter ignore this file, but as it stands this change will make it pass the linter and it should be Python 3 ready.

If this is not used by anything, should it maybe be removed from the repository?

aresch:
Feel free to just squash the whole PR.

tools/gprof2dot.py
arvidn: yeah, I've taken some notes of reviewing whether all the log parsers still need to be here

PR #3076, opened 2018-05-30
add support for multi-home NAT-PMP and PCP

### LINKED ISSUE #2961: Port mappings for IPv6 firewalls
Skimming the UPnP code looks like it is IPv4-only. And NAT-PMP doesn't support IPv6 anyway. So unless I have missed something libtorrent does not appear to talk to IPv6 firewalls present on lots of CPEs. Which means IPv6 listening ports may experience reachability issues similar to the effects of being behind an IPv4 NAT.

Both UPnP-IGD and PCP (the NAT-PMP successor) support creating pinholes in ipv6 firewalls.

ssiloti:
Fixes #2961

### DIFF
--- a/include/libtorrent/enum_net.hpp
+++ b/include/libtorrent/enum_net.hpp
@@ -94,7 +94,9 @@ namespace libtorrent {
 	TORRENT_EXTRA_EXPORT bool in_local_network(std::vector<ip_interface> const& net
 		, address const& addr);
 
-	TORRENT_EXTRA_EXPORT address get_default_gateway(io_service& ios, error_code& ec);
+	// returns the first default gateway found if device is empty
+	TORRENT_EXTRA_EXPORT address get_default_gateway(io_service& ios
+		, string_view device, bool v6, error_code& ec);
 
 	// attempt to bind socket to the device with the specified name. For systems
 	// that don't support SO_BINDTODEVICE the socket will be bound to one of the
--- a/include/libtorrent/error_code.hpp
+++ b/include/libtorrent/error_code.hpp
@@ -326,17 +326,19 @@ namespace libtorrent {
 			// the listen socket associated with this request was closed
 			invalid_listen_socket,
 
+			// these error codes are deprecated, NAT-PMP/PCP error codes have
+			// been moved to their own category
 
 			// The NAT-PMP router responded with an unsupported protocol version
-			unsupported_protocol_version = 120,
+			unsupported_protocol_version TORRENT_DEPRECATED_ENUM = 120,
 			// You are not authorized to map ports on this NAT-PMP router
-			natpmp_not_authorized,
+			natpmp_not_authorized TORRENT_DEPRECATED_ENUM,
 			// The NAT-PMP router failed because of a network failure
-			network_failure,
+			network_failure TORRENT_DEPRECATED_ENUM,
 			// The NAT-PMP router failed because of lack of resources
-			no_resources,
+			no_resources TORRENT_DEPRECATED_ENUM,
 			// The NAT-PMP router failed because an unsupported opcode was sent
-			unsupported_opcode,
+			unsupported_opcode TORRENT_DEPRECATED_ENUM,
 
 
 
--- a/include/libtorrent/natpmp.hpp
+++ b/include/libtorrent/natpmp.hpp
@@ -120,6 +120,8 @@ private:
 		version_pcp = 2,
 	};
 
+	static char const* version_to_string(protocol_version version);
+
 	// See RFC 6887 Section 19.2
 	enum pcp_opcode
 	{
--- a/include/libtorrent/natpmp.hpp
+++ b/include/libtorrent/natpmp.hpp
@@ -45,13 +45,47 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent {
 
+	namespace errors
+	{
+		// See RFC 6887 Section 7.4
+		enum pcp_errors
+		{
+			pcp_success = 0,
+			pcp_unsupp_version,
+			pcp_not_authorized,
+			pcp_malformed_request,
+			pcp_unsupp_opcode,
+			pcp_unsupp_option,
+			pcp_malformed_option,
+			pcp_network_failure,
+			pcp_no_resources,
+			pcp_unsupp_protocol,
+			pcp_user_ex_quota,
+			pcp_cannot_provide_external,
+			pcp_address_mismatch,
+			pcp_excessive_remote_peers,
+		};
+
+		boost::system::error_code make_error_code(pcp_errors e);
+	}
+
+	boost::system::error_category& pcp_category();
+}
+
+namespace boost { namespace system {
+	template<> struct is_error_code_enum<libtorrent::errors::pcp_errors>
+	{ static const bool value = true; };
+} }
+
+namespace libtorrent {
+
 struct TORRENT_EXTRA_EXPORT natpmp
 	: std::enable_shared_from_this<natpmp>
 	, single_threaded
 {
 	natpmp(io_service& ios, aux::portmap_callback& cb);
 
-	void start(address const& local_address, std::string device);
+	void start(address local_address, std::string device);
 
 	// maps the ports, if a port is set to 0
 	// it will not be mapped
@@ -63,6 +97,7 @@ struct TORRENT_EXTRA_EXPORT natpmp
 	void close();
 
 private:
+	static error_code from_result_code(int version, int result);
 
 	std::shared_ptr<natpmp> self() { return shared_from_this(); }
 
@@ -79,8 +114,28 @@ private:
 
 	void disable(error_code const& ec);
 
+	enum protocol_version
+	{
+		version_natpmp = 0,
+		version_pcp = 2,
+	};
+
+	// See RFC 6887 Section 19.2
+	enum pcp_opcode
+	{
+		opcode_announce = 0,
+		opcode_map,
+		opcode_peer,
+	};
+
 	struct mapping_t : aux::base_mapping
 	{
+		// random identifier, used by PCP
+		std::array<char, 12> nonce;
+
+		// only valid if the router supports PCP
+		address external_address;
+
 		// the local port for this mapping. If this is set
 		// to 0, the mapping is not in use
 		int local_port = 0;
@@ -100,6 +155,8 @@ private:
 
 	aux::portmap_callback& m_callback;
 
+	protocol_version m_version = version_pcp;
+
 	aux::vector<mapping_t, port_mapping_t> m_mappings;
 
 	// the endpoint to the nat router
@@ -114,9 +171,12 @@ private:
 	int m_retry_count = 0;
 
 	// used to receive responses in
-	char m_response_buffer[16];
+	// 1100 octets is the maximum size of a PCP packet
+	char m_response_buffer[1100];
 
 	// router external IP address
+	// this is only used if the router does not support PCP
+	// with PCP the external IP is stored with the mapping
 	address m_external_ip;
 
 	// the endpoint we received the message from
--- a/include/libtorrent/natpmp.hpp
+++ b/include/libtorrent/natpmp.hpp
@@ -51,7 +51,7 @@ struct TORRENT_EXTRA_EXPORT natpmp
 {
 	natpmp(io_service& ios, aux::portmap_callback& cb);
 
-	void start();
+	void start(address const& local_address, std::string device);
 
 	// maps the ports, if a port is set to 0
 	// it will not be mapped
--- a/src/enum_net.cpp
+++ b/src/enum_net.cpp
@@ -173,6 +173,52 @@ namespace {
 		);
 	}
 
+#if TORRENT_USE_GETIPFORWARDTABLE || TORRENT_USE_NETLINK
+	address build_netmask(int bits, int family)
+	{
+		if (family == AF_INET)
+		{
+			using bytes_t = boost::asio::ip::address_v4::bytes_type;
+			bytes_t b;
+			std::memset(&b[0], 0xff, b.size());
+			for (int i = int(b.size()) - 1; i > 0; --i)
+			{
+				if (bits < 8)
+				{
+					b[i] <<= bits;
+					break;
+				}
+				b[i] = 0;
+				bits -= 8;
+			}
+			return address_v4(b);
+		}
+#if TORRENT_USE_IPV6
+		else if (family == AF_INET6)
+		{
+			using bytes_t = boost::asio::ip::address_v6::bytes_type;
+			bytes_t b;
+			std::memset(&b[0], 0xff, b.size());
+			for (int i = int(b.size()) - 1; i > 0; --i)
+			{
+				if (bits < 8)
+				{
+					b[i] <<= bits;
+					break;
+				}
+				b[i] = 0;
+				bits -= 8;
+			}
+			return address_v6(b);
+		}
+#endif
+		else
+		{
+			return address();
+		}
+	}
+#endif
+
 #if TORRENT_USE_NETLINK
 
 	int read_nl_sock(int sock, span<char> buf, std::uint32_t const seq, std::uint32_t const pid)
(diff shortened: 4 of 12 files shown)


include/libtorrent/enum_net.hpp
```diff
@@ -94,7 +94,9 @@ namespace libtorrent {
 	TORRENT_EXTRA_EXPORT bool in_local_network(std::vector<ip_interface> const& net
 		, address const& addr);
 
-	TORRENT_EXTRA_EXPORT address get_default_gateway(io_service& ios, error_code& ec);
+	// returns the first default gateway found if device is empty
+	TORRENT_EXTRA_EXPORT address get_default_gateway(io_service& ios
+		, string_view device, bool v6, error_code& ec);
```
arvidn: I wonder if it would be better to pass in the `boost::asio::udp::v6` or `boost::asio::udp::v4` instead of this `bool`. The `udp` or `tcp` part is a bit odd I suppose, but the call site would probably be easier to understand than `true` or `false`

include/libtorrent/natpmp.hpp
```diff
@@ -51,7 +85,7 @@ struct TORRENT_EXTRA_EXPORT natpmp
 {
 	natpmp(io_service& ios, aux::portmap_callback& cb);
 
-	void start();
+	void start(address local_address, std::string device);
```
arvidn: I imagine `device` could be passed as a `string_view`

include/libtorrent/natpmp.hpp
```diff
@@ -79,8 +113,28 @@ struct TORRENT_EXTRA_EXPORT natpmp
 
 	void disable(error_code const& ec);
 
+	enum protocol_version
```
arvidn: I'm a fan of `enum class`. Then you can remove the `version_` and `opcode_` prefixes of the enum values too

src/enum_net.cpp
```diff
@@ -173,6 +173,52 @@ namespace {
...
+				bits -= 8;
+			}
+			return address_v4(b);
+		}
+#if TORRENT_USE_IPV6
+		else if (family == AF_INET6)
+		{
+			using bytes_t = boost::asio::ip::address_v6::bytes_type;
+			bytes_t b;
+			std::memset(&b[0], 0xff, b.size());
```
arvidn: are we guaranteed that the bytes type is a `std::array`? if so you could use `b.fill(0)` (with the main benefit of not using `memset()`)

arvidn: I'm also thinking. It feels like there should be a simpler way of generating these bitmasks. stackoverflow suggests a table of 32 entries. But that won't be quite as nice for IPv6. It seems at least that the core of this algorithm could be factored out, and operate on a span.

the `sha1_hash` has function that do things like this too. I'm not sure if they are trivially reusable though. I suppose that's more plain shifting, so a little bit more generic.

src/natpmp.cpp
```diff
@@ -485,35 +717,55 @@ void natpmp::on_reply(error_code const& e
...
+			? portmap_protocol::udp
+			: portmap_protocol::tcp;
+	}
 
 #ifndef TORRENT_DISABLE_LOGGING
 	char msg[200];
 	int const num_chars = std::snprintf(msg, sizeof(msg), "<== port map ["
 		" protocol: %s local: %u external: %u ttl: %u ]"
-		, (cmd - 128 == 1 ? "udp" : "tcp")
+		, (protocol == portmap_protocol::udp ? "udp" : "tcp")
```
arvidn: https://travis-ci.org/arvidn/libtorrent/jobs/385519966#L1310

src/enum_net.cpp
```diff
@@ -509,52 +571,6 @@ int _System __libsocket_sysctl(int* mib, u_int namelen, void *oldp, size_t *oldl
 		return false;
 	}
 
-#if TORRENT_USE_GETIPFORWARDTABLE
-	address build_netmask(int bits, int family)
```
arvidn: oh, I see. This was just moved.

test/enum_if.cpp
```diff
@@ -42,7 +42,7 @@ int main()
 	io_service ios;
 	error_code ec;
 
-	address def_gw = get_default_gateway(ios, ec);
+	address def_gw = get_default_gateway(ios, "", false, ec);
```
arvidn: in order to manually test this function, do you think it would make sense to add a column (or maybe two, one for v4 and one for v6) to the enumeration of interfaces below?

include/libtorrent/enum_net.hpp
ssiloti: In this case it would be pretty awkward because both the caller and callee are dealing with `address`es so it would mean converting bool -> protocol -> bool. I think it's clear enough what's going on at the one-and-only call site of this function.

include/libtorrent/natpmp.hpp
ssiloti: The function might need to fill-in for an empty `device` so the code is cleaner if it takes the parameter by value.

include/libtorrent/natpmp.hpp
ssiloti: The reason I didn't use `enum class` for these is that they primarily serve as aliases for their integral values. If I used `enum class` I'd have to litter the code with casts to `int`.

arvidn:
@ssiloti do you have a PCP enabled router to test this on?

ssiloti:
Yes, my home router supports PCP. One thing I didn't test, because I lack the requisite hardware, is macos. Seeing as I had to fix `enum_net` on both Windows and Linux to get this working I wouldn't be surprised if the macos version needs some work too.

include/libtorrent/error_code.hpp
```diff
@@ -326,6 +326,8 @@ namespace libtorrent {
 			// the listen socket associated with this request was closed
 			invalid_listen_socket,
 
+			// these error codes are deprecated, NAT-PMP/PCP error codes have
+			// been moved to their own category
```
arvidn: there's a `TORRENT_DEPRECATED_ENUM` macro that expands to the appropriate deprecated attribute for compilers that support it. It seems like it would make sense to mark these up, or would it trigger tons of internal warnings?

src/natpmp.cpp
```diff
@@ -55,11 +55,79 @@ POSSIBILITY OF SUCH DAMAGE.
...
+	{
+		return boost::system::error_code(e, pcp_category());
+	}
+}
+
+namespace
+{
+	error_code from_result_code(int const version, int result)
+	{
+		if (version == 0)
```
arvidn: instead of `0`, should you use `version_natpmp`?

src/natpmp.cpp
```diff
@@ -76,19 +144,77 @@ natpmp::natpmp(io_service& ios
...
+			ec.clear();
+		}
+	}
+
+	// we really want a device name to get the right default gateway
+	// try to find one even if the listen socket isn't bound to a device
+	if (device.empty())
+	{
+		device = device_for_address(local_address, m_socket.get_io_service(), ec);
+		// if this failes fall back to using the first default gateway in the
```
arvidn: "failes" -> "fails"

src/natpmp.cpp
```diff
@@ -344,15 +478,73 @@ void natpmp::send_map_request(port_mapping_t const i)
...
+		out = std::copy(local_bytes.begin(), local_bytes.end(), out);
+		out = std::copy(m.nonce.begin(), m.nonce.end(), out);
+		// translate portmap_protocol to an IANA protocol number
+		int protocol = 0;
+		switch (m.protocol)
+		{
+			case portmap_protocol::tcp:protocol = 6; break;
+			case portmap_protocol::udp:protocol = 17; break;
+			case portmap_protocol::none:protocol = 0; break;
+		}
```
arvidn: this could be a bit shorter, and `protocol` could be const if it's initialized with the ternary operator instead.

```
int const protocol =
   (m.protocol == portmap_protocol::tcp) ? 6
   : (m.protocol == portmap_protocol::udp) ? 17
   : 0;
```

arvidn:
I think this patch warrants a note in `ChangeLog` too

src/session_impl.cpp
```diff
@@ -586,7 +586,7 @@ namespace aux {
 
 		// apply all m_settings to this session
 		run_all_updates(*this);
-		reopen_listen_sockets();
+		reopen_listen_sockets(false);
```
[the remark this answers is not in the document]
ssiloti: Without it reopen_listen_sockets sends all mapping requests twice on startup. First when the sockets are created then again in the `map_ports` loop.

arvidn:
regarding mac support. This is what I get from `enum_if`:
```
Default gateway: 10.0.1.1
0.0.0.0           0.0.0.0           10.0.1.1                           259    en0
127.0.0.0         255.0.0.0         127.0.0.1                          259    lo0
address                           netmask                                      name                flags               default gateway                   description
127.0.0.1                         255.0.0.0                                    lo0                 local loopback            0.0.0.0
::1                               ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff      lo0                 local loopback            0.0.0.0
fe80::1%lo0                       ffff:ffff:ffff:ffff::                        lo0                 local                     0.0.0.0
fe80::8d4:44ac:1cc0:5660%en0      ffff:ffff:ffff:ffff::                        en0                 local                     0.0.0.0
10.0.1.5                          255.255.255.0                                en0                 local                     10.0.1.1
fe80::b830:70ff:fecf:f5dd%awdl0   ffff:ffff:ffff:ffff::                        awdl0               local                     0.0.0.0
fe80::1283:87b1:e0ad:a21c%utun0   ffff:ffff:ffff:ffff::                        utun0               local                     0.0.0.0
fe80::e464:8de3:d332:4ed0%utun1   ffff:ffff:ffff:ffff::                        utun1               local                     0.0.0.0
```

and when running with VPN:

```
Default gateway: 10.18.11.5
0.0.0.0           128.0.0.0         10.18.11.5                         259    utun2
0.0.0.0           0.0.0.0           10.0.1.1                           259    en0
0.0.0.0           0.0.0.0           10.18.11.5                         259    utun2
10.18.11.1        255.255.255.255   10.18.11.5                         259    utun2
46.246.123.145    255.255.255.255   10.0.1.1                           259    en0
127.0.0.0         255.0.0.0         127.0.0.1                          259    lo0
128.0.0.0         128.0.0.0         10.18.11.5                         259    utun2
address                           netmask                                      name                flags               default gateway                   description
127.0.0.1                         255.0.0.0                                    lo0                 local loopback            0.0.0.0
::1                               ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff      lo0                 local loopback            0.0.0.0
fe80::1%lo0                       ffff:ffff:ffff:ffff::                        lo0                 local                     0.0.0.0
10.0.1.5                          255.255.255.0                                en0                 local                     10.0.1.1
fe80::b830:70ff:fecf:f5dd%awdl0   ffff:ffff:ffff:ffff::                        awdl0               local                     0.0.0.0
fe80::1283:87b1:e0ad:a21c%utun0   ffff:ffff:ffff:ffff::                        utun0               local                     0.0.0.0
fe80::e464:8de3:d332:4ed0%utun1   ffff:ffff:ffff:ffff::                        utun1               local                     0.0.0.0
10.18.11.6                        0.0.0.0                                      utun2               local                     10.18.11.5
```

looks reasonable I think

ssiloti:
My main concern for `enum_routes` is whether the scope_id is set correctly for IPv6 gateways. If the scope_id is not correct then responses will be rejected because they appear to be coming from a different address than the default gateway's. Unfortunately your system does not have global IPv6 connectivity so I can't tell if it is working based on your results.

Looking at the code it _should_ be working on macos since the gateway address is parsed out of a `sockaddr` which contains a scope_id. However there is a similar situation on Windows and it does not set the scop_id for some reason.

arvidn:
I suppose I should land this. I'm a bit tempted to test it against my NAT-PMP router just to make sure nothing broke. @ssiloti you feel confident about this, right?

arvidn:
do you want to squash any of the commits?

ssiloti:
I would certainly appreciate anyone else testing this against their router. PCP and NAT-PMP are pretty simple protocols so I don't anticipate there being any problems, but of course there's always a chance of surprises.

aldenml:
hi @ssiloti , this the result of IPv6 test in my environment, if you tell me specific the specific steps you need, I can give it a try, but not sure if my router supports these protocols
screen shot 2018-06-08 at 8 49 49 am

ssiloti:
Build and run client_test (or any other client) then use an [IPv6 port scanner](http://www.ipv6scanner.com/cgi-bin/main.py) to see if the client's listen port (6881 by default) is reachable. If it doesn't work then I'd want to see a wireshark/tcpdump capture of the client's startup.

If you have a mac it would be interesting to see the output of the `enum_if` tool which you can find in the test directory.

arvidn:
I don't see any other attempts at mapping the port though. is it supposed to first try PCP and then fail over to NAT-PMP?

ssiloti:
Yes, the code tries PCP first then falls back to NAT-PMP if it gets an unsupported version response. I was only able to simulate this code path so I'm glad your router doesn't support PCP. I've pushed a fix for the invalid size check.

arvidn:
I can confirm that the new version correctly falls back to NAT-PMP. @ssiloti what's your opinion on this small extension?

```
diff --git a/include/libtorrent/portmap.hpp b/include/libtorrent/portmap.hpp
index 9a1ddfd9f..9b672f06b 100644
--- a/include/libtorrent/portmap.hpp
+++ b/include/libtorrent/portmap.hpp
@@ -40,8 +40,8 @@ namespace libtorrent {

 	enum class portmap_transport : std::uint8_t
 	{
-		// natpmp can be NAT-PMP or PCP
-		natpmp, upnp
+		// pcp is normally considered natpmp, except for logging
+		natpmp, upnp, pcp
 	};

 	enum class portmap_protocol : std::uint8_t
diff --git a/simulation/libsimulator b/simulation/libsimulator
--- a/simulation/libsimulator
+++ b/simulation/libsimulator
@@ -1 +1 @@
-Subproject commit ffaccc122691d10131a95bbe085ab325f57a602b
+Subproject commit ffaccc122691d10131a95bbe085ab325f57a602b-dirty
diff --git a/src/alert.cpp b/src/alert.cpp
index 1dd9233ec..9c70152e9 100644
--- a/src/alert.cpp
+++ b/src/alert.cpp
@@ -819,7 +819,7 @@ namespace {
 		return type_str[sock_type_idx(type)];
 	}

-	char const* const nat_type_str[] = {"NAT-PMP", "UPnP"};
+	char const* const nat_type_str[] = {"NAT-PMP", "UPnP", "PCP"};

 	char const* const protocol_str[] = {"none", "TCP", "UDP"};

diff --git a/src/natpmp.cpp b/src/natpmp.cpp
index cabbec1b7..aec2bdece 100644
--- a/src/natpmp.cpp
+++ b/src/natpmp.cpp
@@ -331,7 +331,8 @@ void natpmp::log(char const* fmt, ...) const
 	va_start(v, fmt);
 	std::vsnprintf(msg, sizeof(msg), fmt, v);
 	va_end(v);
-	m_callback.log_portmap(portmap_transport::natpmp, msg);
+	m_callback.log_portmap(m_version == version_natpmp
+		? portmap_transport::natpmp : portmap_transport::pcp, msg);
 }
 #endif

@@ -677,6 +678,9 @@ void natpmp::on_reply(error_code const& e

 	if (result == errors::pcp_unsupp_version)
 	{
+		log("unsupported version");
+#endif
 		if (m_version == version_pcp && !is_v6(m_socket.local_endpoint()))
 		{
 			m_version = version_natpmp;
```

It does make the `portmap_transport` enum a bit weird, but it makes the logging easier to follow. Maybe another approach would be to include the version number with the log alert, to still say NAT-PMP, but with version 1.

arvidn:
@aldenml are you testing this patch too? If so, I'll wait for you before landing it. Otherwise I think this is ready to land.

aldenml:
@arvidn I have plans to do it later on the day, but I can't really commit, if you think it's ready, don't wait for me. I will do the tests and report the results here anyway.

ssiloti:
Can you zip it up and email it to me? Just the alert log is good enough to start with.

aldenml:
sure, I can see your email in your profile, in a minute

ssiloti:
@aldenml You're on macos right? The problem is that `enum_routes` isn't finding any IPv6 routes. Unfortunately I don't have a mac here so someone else is going to have to debug it.

aldenml:
that's correct, macOS. @arvidn do you get IPv6 routes?

PR #3077, opened 2018-05-30
fix integer overflow in alert_manager

arvidn:


sledgehammer999:
Yes, it seems to work now with the `state_update_alert` alert.

PR #3080, opened 2018-06-02
improve the API for iterating over all files and pieces

arvidn:
with the new strong index types.

### DIFF
--- a/include/libtorrent/file_storage.hpp
+++ b/include/libtorrent/file_storage.hpp
@@ -45,6 +45,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/sha1_hash.hpp"
 #include "libtorrent/string_view.hpp"
 #include "libtorrent/aux_/vector.hpp"
+#include "libtorrent/index_range.hpp"
 #include "libtorrent/flags.hpp"
 
 namespace libtorrent {
@@ -106,7 +107,7 @@ namespace libtorrent {
 #endif // TORRENT_ABI_VERSION
 
 	// internal
-	struct TORRENT_DEPRECATED_EXPORT internal_file_entry
+	struct internal_file_entry
 	{
 		friend class file_storage;
 #if TORRENT_USE_INVARIANT_CHECKS
@@ -172,6 +173,7 @@ namespace libtorrent {
 		// to the file
 		int path_index;
 	};
+
 
 	// represents a window of a file in a torrent.
 	//
@@ -341,12 +343,7 @@ namespace libtorrent {
 		TORRENT_DEPRECATED
 		reverse_iterator rend() const { return m_files.rend(); }
 		TORRENT_DEPRECATED
-		internal_file_entry const& internal_at(int index) const
-		{
-			TORRENT_ASSERT(index >= 0);
-			TORRENT_ASSERT(index < int(m_files.size()));
-			return m_files[file_index_t(index)];
-		}
+		internal_file_entry const& internal_at(int const index) const;
 		TORRENT_DEPRECATED
 		file_entry at(iterator i) const;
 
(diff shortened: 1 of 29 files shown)


include/libtorrent/file_storage.hpp
```diff
@@ -105,73 +106,7 @@ namespace libtorrent {
 	};
 #endif // TORRENT_ABI_VERSION
 
-	// internal
-	struct TORRENT_DEPRECATED_EXPORT internal_file_entry
```
arvidn: given that this is an internal type, I moved it into the library to be secret.

include/libtorrent/torrent_info.hpp
```diff
@@ -342,6 +342,8 @@ namespace libtorrent {
 			TORRENT_ASSERT(m_files.num_pieces() > 0);
 			return piece_index_t(m_files.num_pieces());
 		}
+		aux::index_range<piece_index_t> piece_range() const
```
aldenml: I think this leak `aux::` to the public API

arvidn: I would like the return type to be an *unspecified type* that can be used in a range-for loop. But my simple documentation generation script can't hide such concrete types.

aldenml: I think that if this `index_range` is intended to be public API, it should be outside `aux`

include/libtorrent/aux_/range.hpp
```diff
@@ -44,6 +44,33 @@ namespace libtorrent { namespace aux {
 		Iter end() { return _end; }
 	};
 
+	template <typename Index>
+	struct index_iter
+	{
+		index_iter(Index i) : m_idx(i) {}
```
evsh: The constructor should be `explicit` to complete type isolation.

include/libtorrent/aux_/range.hpp
```diff
@@ -44,6 +44,33 @@ namespace libtorrent { namespace aux {
...
+		friend inline bool operator==(index_iter lhs, index_iter rhs)
+		{ return lhs.m_idx == rhs.m_idx; }
+		friend inline bool operator!=(index_iter lhs, index_iter rhs)
+		{ return lhs.m_idx != rhs.m_idx; }
+	private:
+		Index m_idx;
+	};
+
+	template <typename Index>
+	struct index_range
```
evsh: The `rbegin()/rend()` pair would be nice to have too.

include/libtorrent/torrent_info.hpp
aldenml: well, not sure if "unspecified" has some special semantic here (modern c++ has a lot of them lately :), but maybe `using piece_range_t = aux::index_range<piece_index_t>` inside `torrent_info` do the trick of not having `aux::` in public API.

arvidn: Is there some practical (or otherwise) benefit of returning an alias for an internal type over the internal type itself?
The thing I can think of is that the `aux::index_range` becomes part of the ABI here, and it living in the `aux` namespace may be misleading to make it look like it doesn't affect library ABI.

aldenml: yes, I would say that's the only reason, the `aux::` happening in public ABI (at first I was concerned only about API)

include/libtorrent/aux_/range.hpp
zeule: Doesn't `index_iter` need `operator--` then?

include/libtorrent/torrent_info.hpp
arvidn: I moved `index_range` into its own header and out of the `aux` namespace

zeule:
Thanks, with these changes clients should be able to use the index types with ease, I believe. I did not find the typed bitfield class in this PR through, which gives a similar complication right now, as I remember.

aldenml:
do you mean `typed_bitfield`? it is here https://github.com/arvidn/libtorrent/blob/master/include/libtorrent/bitfield.hpp#L262

zeule:
Yes, that type.  And the problem, as it seems to me, is that type isolation here transformed into value isolation: to creating this type out of another bit container and to coping values from it is unnecessary complicated.

aldenml:
do you mind elaborate? I don't think I follow you

arvidn:
@zeule I added a `range()` function on `typed_bitfield` too, but this doesn't really address the core of your point. Do you have any suggestions of how to improve this situation?
maybe some kind of `void import(span<std::uint8_t> bits)`?

zeule:
@arvidn, can't see how `span<std::uint8_t>` would make importing easier, maybe `template <class t> void import(span<T> bits)`? 

In my humble opinion, iterators would provide a better solution, because `<algorithm>` provides tools for all the required operations. I ended up with [this](https://github.com/zeule/qbit/blob/master/src/base/range.h#L80) helper class.

arvidn:
I see. That only works for copying bits out of a bitfield, since `operator[]` doesn't return an lvalue.

zeule:
Yes, I did not finish that properly with a `BoolValue` class to set values too.

arvidn:
even without iterators on `bitfield`, this patch seems like a step forward. I'm inclined to land it.

PR #3088, opened 2018-06-05
add bool is_save_resume()

pavel-pimenov:


include/libtorrent/torrent_handle.hpp
```diff
@@ -694,7 +694,7 @@ TORRENT_IPV6_NAMESPACE_END
 		//		if (!h.is_valid()) continue;
 		//		torrent_status s = h.status();
 		//		if (!s.has_metadata) continue;
-		//		if (!s.need_save_resume_data()) continue;
+		//		if (!h.need_save_resume_data()) continue;
```
arvidn: I believe this function is deprecated, in favor of the torrent flag (on torrent_status::flags)

arvidn: it doesn't seem to be deprecated, but it's more efficient to check `s.flags & torrent_flags_t::need_save_resume_data` instead

include/libtorrent/torrent_status.hpp
```diff
@@ -78,7 +78,12 @@ TORRENT_VERSION_NAMESPACE_2
 		// only the torrent_handle field is compared.
 		bool operator==(torrent_status const& st) const
 		{ return handle == st.handle; }
-
+	
+		bool is_save_resume() const
+		{
+		  return handle.is_valid() && has_metadata && need_save_resume;
```
arvidn: Are you taking the addresses of these member functions and converting them to bools?

examples/client_test.cpp
```diff
@@ -1302,11 +1302,7 @@ example alert_masks:
 						, file.c_str(), ec.message().c_str());
 					continue;
 				}
-
-				// we're loading this torrent from resume data. There's no need to
-				// re-save the resume data immediately.
-				p.flags &= ~lt::torrent_flags::need_save_resume;
```
arvidn: Can you explain why this was removed?

include/libtorrent/torrent_status.hpp
arvidn: I think this should be a free function, and be defined in client_test.cpp

src/read_resume_data.cpp
```diff
@@ -336,6 +336,8 @@ namespace {
 			}
 		}
 
+		// we're loading this torrent from resume data. There's no need to
+		// re-save the resume data immediately.
 		ret.flags &= ~torrent_flags::need_save_resume;
```
arvidn: Aha, I see

include/libtorrent/torrent_status.hpp
arvidn: I don't understand the naming here either. what does it refer to, that *is* save resume?
It seems it would make much more sense for the function to be called: `need_save_resume()`

include/libtorrent/torrent_handle.hpp
```diff
@@ -690,11 +690,9 @@ TORRENT_IPV6_NAMESPACE_END
...
 		//	for (torrent_handle i : handles)
 		//	{
-		//		torrent_handle& h = *i;
-		//		if (!h.is_valid()) continue;
-		//		torrent_status s = h.status();
-		//		if (!s.has_metadata) continue;
-		//		if (!s.need_save_resume_data()) continue;
+		//		const torrent_handle& h = *i;
+		//		const torrent_status s = h.status();
+		//		if (!h.is_need_save_resume()) continue;
```
arvidn: this is still using `h` here, I think you meant to use `s`

include/libtorrent/torrent_status.hpp
arvidn: as far as I can tell, `torrent_status::need_save_resume` should be deprecated. This state is part of the `flags` now.

arvidn: Do you disagree with this being a free function?

arvidn:
closing in favour of https://github.com/arvidn/libtorrent/pull/3161

PR #3090, opened 2018-06-06
Test merge

arvidn:


src/random.cpp
```diff
@@ -62,8 +64,9 @@ namespace libtorrent { namespace aux {
 			// make sure random numbers are deterministic. Seed with a fixed number
 			static std::mt19937 rng(0x82daf973);
 #else
+			static std::mutex mutex;
```
aldenml: why the mutex is needed here, if `rng` is `thread_local`, why not make `dev` too?

arvidn: aha, left-over from RC_1_1. Will fix

arvidn: it turns out; in order to support old macOS systems the mutex is still required where `thread_local` isn't

aldenml: interesting

PR #3105, opened 2018-06-12
some refactor, more use of auto and const in peer_connection.cpp

aldenml:


src/peer_connection.cpp
```diff
@@ -1319,8 +1318,11 @@ namespace libtorrent {
 				}
 				// find the lowest ranking peer and disconnect that
 				peer_connection* p = other_t->find_lowest_ranking_peer();
-				p->disconnect(errors::too_many_connections, operation_t::bittorrent);
-				peer_disconnected_other();
+				if (p != nullptr)
```
aldenml: @arvidn do I need to put an `else` here? or is this correct?

arvidn: one of the connections need to be disconnected, so there needs to be an else block doing:
```
 disconnect(errors::too_many_connections, operation_t::bittorrent);
 return;
```

I believe

src/peer_connection.cpp
```diff
@@ -1488,10 +1490,11 @@ namespace libtorrent {
 
 		if (is_disconnecting()) return;
 
+		int const block_size = t->block_size();
 		auto const dlq_iter = std::find_if(
 			m_download_queue.begin(), m_download_queue.end()
-			, std::bind(match_request, std::cref(r), std::bind(&pending_block::block, _1)
-			, t->block_size()));
+			, [&r, block_size](pending_block const& pb)
+			{ return match_request(r, pb.block, block_size); });
```
arvidn: is `match_request` used in other places too? if not, should it just be moved directly into the lambda instead?

src/peer_connection.cpp
aldenml: Ok, I will add that

src/peer_connection.cpp
aldenml: no, I will move it here then

PR #3107, opened 2018-06-13
fixed sign implicit conversion warnings in new enum_net code

aldenml:


src/enum_net.cpp
```diff
@@ -178,10 +178,9 @@ namespace {
 	{
 		if (family == AF_INET)
 		{
-			using bytes_t = boost::asio::ip::address_v4::bytes_type;
-			bytes_t b;
-			std::memset(&b[0], 0xff, b.size());
-			for (int i = int(b.size()) - 1; i > 0; --i)
+			address_v4::bytes_type b;
+			b.fill(0xff);
+			for (std::size_t i = b.size() - 1; i > 0; --i)
```
aldenml: hi @ssiloti, I want to confirm that `i > 0` (with `>`) is the correct condition, thanks

ssiloti: No, it's not. The loop needs to run when i = 0 to set the least significant byte's value correctly if `bits` is less than 8 bits smaller than the size of the address.

aldenml: I will change that then

src/enum_net.cpp
```diff
@@ -174,39 +174,37 @@ namespace {
...
 	{
 		if (family == AF_INET)
 		{
-			using bytes_t = boost::asio::ip::address_v4::bytes_type;
-			bytes_t b;
-			std::memset(&b[0], 0xff, b.size());
-			for (int i = int(b.size()) - 1; i > 0; --i)
+			address_v4::bytes_type b;
+			b.fill(0xff);
+			for (int i = int(b.size()) - 1; i >= 0; --i)
```
aldenml: @ssiloti do it looks good now? (with `i >= 0`)

ssiloti: lgtm

PR #3110, opened 2018-06-14
Fix parts file i/o errors

d-komarov:
Since i/o operations are not synchronized, it is possible that one
thread closes file handle, opened by another thread. With addition
of `hidden` file attribute, an old bug in `part_file::open_file()`
revealed both problems, causing a lot of random i/o errors in parts
file. Fixing `open_file` bug should reduce the number of such errors
back to 'normal'.

* Add more const
* Extract slot offset calculation to separate function

### DIFF
--- a/src/part_file.cpp
+++ b/src/part_file.cpp
@@ -44,7 +44,7 @@ POSSIBILITY OF SUCH DAMAGE.
   // the size of the torrent (and can be used to calculate the size
   // of the file header)
   uint32_t num_pieces;
-  
+
   // the number of bytes in each piece. This determines the size of
   // each slot in the part file. This is typically an even power of 2,
   // but it is not guaranteed to be.
@@ -58,7 +58,7 @@ POSSIBILITY OF SUCH DAMAGE.
   // unused, n is defined as the number to align the size of this
   // header to an even multiple of 1024 bytes.
   uint8_t padding[n];
- 
+
 */
 
 #include "libtorrent/part_file.hpp"
@@ -72,6 +72,9 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace
 {
+	inline boost::int64_t calc_offset(boost::int64_t slot, int header_size, int piece_size)
+	{ return header_size + slot * piece_size; }
+
 	// round up to even kilobyte
 	int round_up(int n)
 	{ return (n + 1023) & ~0x3ff; }
@@ -192,7 +195,7 @@ namespace libtorrent
 
 		l.unlock();
 
-		boost::int64_t slot_offset = boost::int64_t(m_header_size) + boost::int64_t(slot) * m_piece_size;
+		boost::int64_t const slot_offset = calc_offset(slot, m_header_size, m_piece_size);
 		return m_file.writev(slot_offset + offset, bufs, num_bufs, ec);
 	}
 
@@ -210,24 +213,24 @@ namespace libtorrent
 			return -1;
 		}
 
-		int slot = i->second;
+		int const slot = i->second;
 
 		open_file(file::read_write | file::attribute_hidden, ec);
 		if (ec) return -1;
 
 		l.unlock();
 
-		boost::int64_t slot_offset = boost::int64_t(m_header_size) + boost::int64_t(slot) * m_piece_size;
+		boost::int64_t const slot_offset = calc_offset(slot, m_header_size, m_piece_size);
 		return m_file.readv(slot_offset + offset, bufs, num_bufs, ec);
 	}
 
 	void part_file::open_file(int mode, error_code& ec)
 	{
-		if (m_file.is_open()
-			&& ((m_file.open_mode() & file::rw_mask) == mode
-				|| mode == file::read_only)) return;
+		if (m_file.is_open() && mode == file::read_only
+			|| (m_file.open_mode() & file::rw_mask) == (mode & file::rw_mask))
+			return;
 
-		std::string fn = combine_path(m_path, m_name);
+		std::string const fn = combine_path(m_path, m_name);
 		m_file.open(fn, mode, ec);
 		if (((mode & file::rw_mask) != file::read_only)
 			&& ec == boost::system::errc::no_such_file_or_directory)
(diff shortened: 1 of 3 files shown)


src/part_file.cpp
```diff
@@ -210,24 +213,24 @@ namespace libtorrent
...
 	}
 
 	void part_file::open_file(int mode, error_code& ec)
 	{
-		if (m_file.is_open()
-			&& ((m_file.open_mode() & file::rw_mask) == mode
-				|| mode == file::read_only)) return;
+		if (m_file.is_open() && mode == file::read_only
+			|| (m_file.open_mode() & file::rw_mask) == (mode & file::rw_mask))
+			return;
```
d-komarov: Here, wrong `mode` argument comparison caused file to be re-opened most of time.

d-komarov:
Further consideration make me worry about 'export' problem. Reading from a parts file could break `part_file::export_file` running in different thread.

Maybe it worth using `read_write` mode everywhere in `part_file`, except constructor? Or using `read_only` mode in `part_file::readv`?

src/part_file.cpp
```diff
@@ -210,24 +213,24 @@ namespace libtorrent
...
 		return m_file.readv(slot_offset + offset, bufs, num_bufs, ec);
 	}
 
 	void part_file::open_file(int mode, error_code& ec)
 	{
-		if (m_file.is_open()
-			&& ((m_file.open_mode() & file::rw_mask) == mode
-				|| mode == file::read_only)) return;
+		if (m_file.is_open() && (mode == file::read_only
+			|| (m_file.open_mode() & file::rw_mask) == (mode & file::rw_mask)))
```
arvidn: I think the original intention for that expression was to say: `(m_file.open_mode() & mode) == mode`, i.e. is every bit in `mode` is set in `m_file.open_mode()`? If so, we have at least the capabilities we need.

arvidn: I think that expression has the same behavior as yours, but I'd say is a bit simpler (assuming you expect the mode to be a bitmask)

src/part_file.cpp
```diff
@@ -72,6 +72,9 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace
 {
+	inline boost::int64_t calc_offset(boost::int64_t slot, int header_size, int piece_size)
```
arvidn: it almost seems warranted to make this a member, so it can pick up `m_header_size` and `m_piece_size` by itself.

src/part_file.cpp
arvidn: ah, I suppose it would have to look like this:
`(m_file.open_mode() & mode & file::rw_mask) == (mode & file::rw_mask)`

src/part_file.cpp
d-komarov: Don't you mind given name?

arvidn: `calc_offset()` seems fine, or maybe `slot_offset()`, unless it clashes with some existing names already

arvidn:
btw, the reason I think it's good to use read_only when it's possible is that sometimes people mount drives read-only, and want to seed from it

d-komarov:
In this case `part_file::readv` should be opening file in `read_only` mode.

src/part_file.cpp
d-komarov: Now I'm starting to hesitate if any of given expressions are correct for this function.
Maybe we should list constraints that come from possible use cases?

arvidn: well, in practice I don't think that's important because libtorrent doesn't actually use that mode. In fact, in a future branch (not master, but `new-disk-io`) that mode has been removed

d-komarov: Next let's define compatibility of different file modes in terms of `part_file` operations.

d-komarov: Does it make sense to consider any of these values: `sparse`, `no_atime`, `random_access`, `lock_file`, `no_cache`?
I didn't include file attributes, because their modification doesn't require file handle.

arvidn: I don't think so. I would imagine `random_access` would always be set (but maybe it isn't). Otherwise I don't think really any of those flags are critical.

arvidn: I would think it's more reasonable to ignore `random_access` and just make sure the read-write permissions are compatible.

what I think you want to check is to make sure that all bits set in `mode` are also set in `m_file.open_mode()`

arvidn:
yes, I believe that's right

src/part_file.cpp
d-komarov: So my initial change in `part_file::open_file` is correct?

arvidn:
I posted a pull request against this pull request, [here](https://github.com/d-komarov/libtorrent/pull/3).

arvidn:
also btw, I think this fix warrants a note in the Changelog. would you mind adding that?

d-komarov:
Sure, I'll drop a line to the Changelog.

src/part_file.cpp
```diff
@@ -314,9 +314,6 @@ namespace libtorrent
 
 				if (!buf) buf.reset(new char[m_piece_size]);
 
-				// don't hold the lock during disk I/O
-				l.unlock();
```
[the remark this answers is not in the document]
d-komarov: For me it's more like blocking i/o could be neglectable here, because the only case for this is changing file priorities.
No other operations would suffer from this change.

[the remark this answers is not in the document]
arvidn: in principle, the `readv()` function could break for the same reason. Imagine you have one thread in a `readv()` call, where the partfile was opened in read-only. Another thread comes around and wants to *write* to the partfile. It will then re-open the file in read-write mode. Possibly breaking the outstanding read operation in the first thread.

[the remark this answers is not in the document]
arvidn: this is a separate issue though, and should be fixed in a separate PR I think

arvidn:
thanks for figuring this out and submitting this patch! I will merge this as soon as it's green

Hlsgs:
Is this included in v4.1.8? I'm still getting I/O errors on that.

arvidn:
this was released in libtorrent-1.1.8, which was released over a year ago: https://github.com/arvidn/libtorrent/releases/tag/libtorrent-1_1_8

Hlsgs:
@arvidn I mistook this for a qBittorrent issue, not libtorrent. Sorry for that!

PR #3115, opened 2018-06-18
fix race condition in part_file

arvidn:
@d-komarov does this looks reasonable?

### DIFF
--- a/src/part_file.cpp
+++ b/src/part_file.cpp
@@ -94,13 +94,13 @@ namespace libtorrent
 
 		error_code ec;
 		std::string fn = combine_path(m_path, m_name);
-		m_file.open(fn, file::read_only, ec);
+		boost::shared_ptr<file> const f = boost::make_shared<file>(fn, file::read_only, boost::ref(ec));
 		if (!ec)
 		{
 			// parse header
 			boost::scoped_array<boost::uint32_t> header(new boost::uint32_t[m_header_size / 4]);
 			file::iovec_t b = {header.get(), size_t(m_header_size) };
-			int n = m_file.readv(0, &b, 1, ec);
+			int const n = f->readv(0, &b, 1, ec);
 			if (ec) return;
 
 			// we don't have a full header. consider the file empty
@@ -142,8 +142,7 @@ namespace libtorrent
 			{
 				if (free_slots[i]) m_free_slots.push_back(i);
 			}
-
-			m_file.close();
+			m_file = f;
 		}
 	}
 
@@ -186,8 +185,9 @@ namespace libtorrent
 		boost::unordered_map<int, int>::iterator const i = m_piece_map.find(piece);
 		int const slot = i == m_piece_map.end() ? allocate_slot(piece) : i->second;
 
+		boost::shared_ptr<file> const f = m_file;
 		l.unlock();
-		return m_file.writev(slot_offset(slot) + offset, bufs, num_bufs, ec);
+		return f->writev(slot_offset(slot) + offset, bufs, num_bufs, ec);
 	}
 
 	int part_file::readv(file::iovec_t const* bufs, int num_bufs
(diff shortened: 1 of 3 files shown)


src/part_file.cpp
```diff
@@ -310,19 +315,27 @@ namespace libtorrent
 			{
 				int const slot = i->second;
 				open_file(file::read_only, ec);
+				boost::shared_ptr<file> local_file = m_file;
 				if (ec) return;
 
 				if (!buf) buf.reset(new char[m_piece_size]);
 
+				// don't hold the lock during disk I/O
+				l.unlock();
```
arvidn: I added these `unlock()` and `lock()` back again, as I believe they are safe now

src/part_file.cpp
```diff
@@ -208,18 +207,20 @@ namespace libtorrent
...
 	}
 
 	void part_file::open_file(int mode, error_code& ec)
 	{
-		if (m_file.is_open() && ((mode & file::rw_mask) == file::read_only
-			|| (m_file.open_mode() & file::rw_mask) == file::read_write))
+		if (m_file && m_file->is_open()
+			&& ((mode & file::rw_mask) == file::read_only
+			|| (m_file->open_mode() & file::rw_mask) == file::read_write))
 			return;
```
d-komarov: Adding assertion for modes containing `lock_file` flag, could make sense, because it will block subsequent open operations if they request `read_write` access.

src/part_file.cpp
```diff
@@ -310,19 +315,27 @@ namespace libtorrent
 			{
 				int const slot = i->second;
 				open_file(file::read_only, ec);
+				boost::shared_ptr<file> local_file = m_file;
```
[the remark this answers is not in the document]
arvidn: not trivially, this line depends on `open_file()` to actually initialize `m_file`. If this is moved, `open_file()` has to be moved as well, and I think it's in here to avoid the pessimisation of assuming we need to open the file even in cases were we don't.

src/part_file.cpp
```diff
@@ -186,8 +184,9 @@ namespace libtorrent
 		boost::unordered_map<int, int>::iterator const i = m_piece_map.find(piece);
 		int const slot = i == m_piece_map.end() ? allocate_slot(piece) : i->second;
 
+		boost::shared_ptr<file> f = m_file;
```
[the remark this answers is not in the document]
arvidn: I'm a proponent of "const by default", so it seems reasonable to make this const too

src/part_file.cpp
```diff
@@ -94,13 +94,13 @@ namespace libtorrent
...
 		std::string fn = combine_path(m_path, m_name);
-		m_file.open(fn, file::read_only, ec);
+		m_file = boost::make_shared<file>(fn, file::read_only, ec);
 		if (!ec)
 		{
 			// parse header
 			boost::scoped_array<boost::uint32_t> header(new boost::uint32_t[m_header_size / 4]);
 			file::iovec_t b = {header.get(), size_t(m_header_size) };
-			int n = m_file.readv(0, &b, 1, ec);
+			int n = m_file->readv(0, &b, 1, ec);
```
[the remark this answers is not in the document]
arvidn: do you think I should close and reset `m_file` if `ec`?

src/part_file.cpp
```diff
@@ -208,18 +207,20 @@ namespace libtorrent
 		open_file(file::read_write | file::attribute_hidden, ec);
```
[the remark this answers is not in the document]
arvidn: oh. I thought this would have opened the file in `read_only` mode. I missed this

src/part_file.cpp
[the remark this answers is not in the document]
d-komarov: Yes, any read operation won't success in that case. Write operations would require reopening file in different mode.
Also an empty/invalid file could be removed here.

src/part_file.cpp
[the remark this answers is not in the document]
d-komarov: Same goes to me.

src/part_file.cpp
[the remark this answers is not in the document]
d-komarov: Yes, I also just realized that exporting a single file shouldn't require more than 1 piece.

src/part_file.cpp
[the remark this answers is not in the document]
arvidn: ok, I updated this to only set `m_file` in case everything succeeds

src/part_file.cpp
```diff
@@ -94,13 +94,13 @@ namespace libtorrent
...
 		if (!ec)
 		{
 			// parse header
 			boost::scoped_array<boost::uint32_t> header(new boost::uint32_t[m_header_size / 4]);
 			file::iovec_t b = {header.get(), size_t(m_header_size) };
-			int n = m_file.readv(0, &b, 1, ec);
+			int const n = f->readv(0, &b, 1, ec);
 			if (ec) return;
 
 			// we don't have a full header. consider the file empty
```
d-komarov: I can't see previous comments on this.
I'll duplicate the main point, I think it is fine to release m_file, because subsequent reads from an empty/invalid file will fail. Removing such file from disk when possible is also fine.

d-komarov: No problem.

Releasing the file is not that important on most modern operating systems. At the same time I suggest to avoid wasting of system resources when possible.

Since `m_piece_map` is always verified before reading, `open_file` won't be called for empty/invalid file.

arvidn: in fact, since the part file is only expected to be used rarely, perhaps it doesn't make sense to keep an open file at all, but to just open it on demand. it would make it even "more" thread safe. maybe something for the master branch

arvidn: oh, but this comment still applies to the constructor. your point has been addressed. Note how the file is opened and owned by a local variable. Only after everything succeeded will the `m_file` be assigned

d-komarov: Do you think invalid/empty file could be removed in constructor?

arvidn: you mean if there is a file on disk already, but it's empty?
as long as any potential races could be worked out, I suppose so. There shouldn't be any race in the constructor I wouldn't think. I think that's for another PR though

d-komarov: Exactly.

I agree, this question if out of current scope.

PR #3117, opened 2018-06-19
minor code refactor, cleanup and typo in peer related code

aldenml:


src/peer_connection.cpp
```diff
@@ -6012,17 +6012,13 @@ namespace libtorrent {
...
-#endif
-
 		INVARIANT_CHECK;
 
 #ifndef TORRENT_DISABLE_LOGGING
 		{
 			std::shared_ptr<torrent> t = m_torrent.lock();
 			if (t) t->debug_log("END connect [%p]", static_cast<void*>(this));
-			m_connect_time = completed;
+			m_connect_time = aux::time_now();
```
aldenml: hi @arvidn, now that `aux::time_now()` is not cached, do you prefer `clock_type::now()` or `aux::time_now()`?

arvidn: I think `clock_type::now()` is probably better now

PR #3118, opened 2018-06-20
Test merge

arvidn:


aldenml:
I don't know if it's the diff that's too messy to read, but I barely get what the change is. Is it that you read an entire piece at once, instead of block by block?

arvidn:
yes. the difficulty reading the patch has to do with a bunch of code getting indented one more level. It's easier to read ignoring whitespace. Just append `?w=1` to the URL when looking at the diff

src/disk_io_thread.cpp
```diff
@@ -2257,91 +2277,142 @@ constexpr disk_job_flags_t disk_interface::cache_hit;
...
+
+				time_point const start_time = clock_type::now();
+				int const read_ret = j->storage->readv(iov
+					, j->piece, offset, file_flags, j->error);
 
-				if (iov.data() == nullptr)
+				if (read_ret >= 0)
 				{
-					l.lock();
+					boost::int64_t const read_time = total_microseconds(clock_type::now() - start_time);
```
aldenml: this should be `std::int64_t`

src/disk_io_thread.cpp
```diff
@@ -2257,91 +2277,142 @@ constexpr disk_job_flags_t disk_interface::cache_hit;
...
+					slow_path = false;
 
-					m_disk_cache.maybe_free_piece(pe);
+					l.lock();
+					m_disk_cache.insert_blocks(pe, first_block, iov, j);
+					l.unlock();
+				}
+				else
+				{
+					m_disk_cache.free_iovec(iov);
```
aldenml: Do you mind commenting why `l.lock()/l.unlock()` is not necessary here?

aldenml: thanks

src/peer_connection.cpp
```diff
@@ -5253,6 +5253,13 @@ namespace libtorrent {
 
 		if (error)
 		{
+#ifndef TORRENT_DISABLE_LOGGING
+			peer_log(peer_log_alert::info, "SEED_MODE_HASH"
```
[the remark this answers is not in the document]
arvidn: the commit on top of the merge (that adds logging) is just there to gather some hints on why this is failing. It's not failing on macOS for me, so I can't see it fail other than on travis and appveyor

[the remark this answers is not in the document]
aldenml: oh, ignore then

arvidn:
Sweden may have just lost against Germany (and will likely not make it further in the world cup), but at least I found this coalesce-read bug that's been haunting me for the last week or so while watching it.

PR #3119, opened 2018-06-21
set the minimum number of checking jobs based on the number of hasher…

ssiloti:
… threads

To effectively mask the latency of generating more hash jobs there need to be
at least two jobs in-flight for each hasher thread.

### DIFF
--- a/src/disk_io_thread.cpp
+++ b/src/disk_io_thread.cpp
@@ -220,9 +220,8 @@ namespace libtorrent
 				boost::shared_ptr<io_service::work> work =
 					boost::make_shared<io_service::work>(boost::ref(m_ios));
 
-				// the magic number 3 is also used in add_job()
 				// every 4:th thread is a hasher thread
-				if ((thread_id & 0x3) == 3) type = hasher_thread;
+				if ((thread_id & hasher_thread_mask) == hasher_thread_mask) type = hasher_thread;
 				m_threads.push_back(boost::shared_ptr<thread>(
 					new thread(boost::bind(&disk_io_thread::thread_fun, this
 						, thread_id, type, work))));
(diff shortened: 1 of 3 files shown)


src/disk_io_thread.cpp
```diff
@@ -3246,7 +3245,7 @@ namespace libtorrent
 		// if there are at least 3 threads, there's a hasher thread
 		// and the hash jobs go into a separate queue
 		// see set_num_threads()
-		if (m_num_threads > 3 && j->action == disk_io_job::hash)
+		if (m_num_threads > hasher_thread_mask && j->action == disk_io_job::hash)
```
arvidn: I was thinking of another condition as well. ``&& !m_queued_jobs.empty()``

If the main purpose of separating hash jobs from other disk jobs is to avoid starvation, if the normal queue is empty, it should be ok to put one hash job on there, to get at least a little bit utilisation of those threads too

ssiloti: That would potentially lead to all non-hasher threads being occupied with hash jobs. I'm worried about the impact that would have on the latency of sporadic reads that don't keep the queue consistently populated.

arvidn: fair enough. It seems a bit wasteful though, in the case of re-checking torrents, that only 1 in 4 threads are used. I wonder if the logic should be inverted. 1 in 4 threads *only* accepts non-hash jobs, and the other threads accept all kinds of jobs. I guess it's not quite that simple, since the logic now requires the threads to know which queue to pull from. Maybe 1 in 4 threads only pull from the normal queue, 2 in 4 pulls from the normal queue first, but falls back to pull from the hash queue if it's empty, and the last thread pulls only from the hash queue. It's probably not worth making this too complex either.

arvidn:
btw. I'll land this patch as soon as I get my merge to work. I'd prefer to not have to re-merge on top of this

Issue #3131, opened 2018-06-26
On startup download speed stuck at 0Kbytes/sec for 10 seconds


dimitriss:
libtorrent version (or branch): RC_1_1
platform/architecture: Linux-arm64
compiler and compiler version: GCC6

New torrent download problems with commit c55bc7dd420f5f79b571e85d0984049a6fa28307. Transfer speed is at 0Kbytes/sec for the first 10 to 15 seconds. 
Same branch, same build, same torrent with this commit reverted works fine.

arvidn:
does this happen consistently? Do you have any instructions on how to reproduce it?

arvidn:
would you be able to produce a log (enable session, torrent and peer logging and print alerts to a file)?

dimitriss:
Here is all log with alerts set to all_categories
[4164790068.zip](https://github.com/arvidn/libtorrent/files/2140598/4164790068.zip)

arvidn:
does this only happen with magnet links? i.e. is it possible the problem is it taking longer to get the metadata from the swarm?

arvidn:
If that commit really is the one causing the problem, presumably it has something to do with file priorities. Are you setting any file priorities? or piece priorities?

perhaps the returned priority vector gets corrupted somehow, and we think all files have priority 0 for a little while, resulting in not being interested in any peer.

dimitriss:
Yes I'm setting both file, piece priorities and using sequential download. Issue occurs only with magnets.

arvidn:
can you describe when and to what you set the priorities to? also how you interleave setting of both piece- and file priorities. Normally they don't mix well, as file priorities overwrite piece priorities.

There's no log emitted for updating priorities so I can't tell from the log

dimitriss:
After getting metadata I'm setting file priority 1 to largest file and 0 to the other files of the torrent

d-komarov:
#3133 might be fixing this issue.

dimitriss:
Yes issue resolved with https://github.com/arvidn/libtorrent/pull/3133

PR #3132, opened 2018-06-26
Bump `cache_flushed_alert` priority

d-komarov:
A call to `torrent_handle::flush_cache` must result in `cache_flushed_alert` being posted.

### DIFF
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -1630,7 +1630,7 @@ namespace libtorrent
 		// internal
 		cache_flushed_alert(aux::stack_allocator& alloc, torrent_handle const& h);
 
-		TORRENT_DEFINE_ALERT(cache_flushed_alert, 58)
+		TORRENT_DEFINE_ALERT_PRIO(cache_flushed_alert, 58, alert_priority_critical)
 
 		static const int static_category = alert::storage_notification;
 	};


d-komarov:
Seems to be fair enough, isn't it?

d-komarov:
test_torrent fail is related to #3131

arvidn:
is this alert only posted as a result of the client requesting it?

d-komarov:
I'm afraid that there is a bug in c55bc7d. I'm currently working on fix and hope to present it soon.

d-komarov:
Actually, it could also be posted from `torrent::abort()` and `torrent::finished()` when `async_release_files` is being called.

d-komarov:
Use case: perform some action on file that just finished downloading.

Currently there is no other way to know that torrent finished downloading and has all bytes written to disk.

For instance, unpacking a small archive might fail if it was started in between `torrent_finished_alert` and `cache_flushed_alert`.

arvidn:
I wonder if maybe this alert shouldn't be posted other than when a torrent completes. Or maybe the `torrent_finished_alert` should be moved, to be posted after storage has been flushed. I kind of like that more, as it provides a lightly higher level abstraction.

d-komarov:
Yes, posting `torrent_finished_alert` when all files has been flushed to disk, would be a perfect solution. Also, I think that alert's priority should be raised in order to lower the chance of loosing it. Then posting `cache_flushed_alert` can stay unchanged.

d-komarov:
> I wonder if maybe this alert shouldn't be posted other than when a torrent completes.
But what about `torrent_handle::flush_cache()`, do you think there is no sense in posting `cache_flush_alert` as response?

arvidn:
right, if it's manually triggered it should be posted. maybe it would be reasonable then to *only* post `cache_flush_alert` when it's manually triggered, and always post `torrent_finished_alert` after the cache has been flushed.

d-komarov:
Yes, I like this idea. It's only left to figure out how many changes must be done in order to do this.

arvidn:
I'm having second thoughts about moving altering the meaning of `torrent_finished_alert`. It's currently tied to updating the state, which would be risky to move and make conditional on a bunch of things.

arvidn:
do you think this would do? https://github.com/arvidn/libtorrent/pull/3156/files

PR #3133, opened 2018-06-27
Fix torrent files prioritization

d-komarov:
After setting file priority, any subsequent attempt to set different
priorities will fail if there is a `file_priority` job running in in
disk thread. This happens because `torrent::m_file_priority` is being
updated before adding disk thread job. The problem is gone if the file
priority vector owned by the torrent object, is modified in the
`torrent::on_file_priority` callback, when disk job finishes.

* Fix torrent test
* Minor `torrent` class refactoring

### DIFF
--- a/bindings/python/test.py
+++ b/bindings/python/test.py
@@ -50,6 +50,8 @@ class test_torrent_handle(unittest.TestCase):
 		self.assertEqual(self.h.piece_priorities(), [4])
 
 		self.h.prioritize_files([0,1])
+		# workaround for asynchronous priority update
+		time.sleep(1)
 		self.assertEqual(self.h.file_priorities(), [0,1])
 
 		self.h.prioritize_pieces([0])
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -552,7 +552,6 @@ namespace libtorrent
 		void set_piece_deadline(int piece, int t, int flags);
 		void reset_piece_deadline(int piece);
 		void clear_time_critical();
-		void update_piece_priorities();
 
 		void status(torrent_status* st, boost::uint32_t flags);
 
@@ -1149,7 +1148,7 @@ namespace libtorrent
 
 		void ip_filter_updated();
 
-		void inc_stats_counter(int c, int value = 1);
+		void inc_stats_counter(int c, int value = 1) const;
 
 		// initialize the torrent_state structure passed to peer_list
 		// member functions. Don't forget to also call peers_erased()
@@ -1182,6 +1181,7 @@ namespace libtorrent
 		bool request_bandwidth_from_session(int channel) const;
 
 		void update_peer_interest(bool was_finished);
+		void update_piece_priorities();
 		void prioritize_udp_trackers();
 
 		void update_tracker_timer(time_point now);
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -5650,9 +5650,9 @@ namespace {
 
 	namespace
 	{
-		std::vector<uint8_t> fix_priorities(std::vector<int> const& input, file_storage const& fs)
+		std::vector<boost::uint8_t> fix_priorities(std::vector<int> const& input, file_storage const& fs)
 		{
-			std::vector<uint8_t> files(input.begin(), input.end());
+			std::vector<boost::uint8_t> files(input.begin(), input.end());
 
 			for (int i = 0; i < std::min<int>(fs.num_files(), files.size()); ++i)
 			{
@@ -5663,7 +5663,6 @@ namespace {
 			}
 
 			files.resize(fs.num_files(), 4);
-
 			return files;
 		}
 
@@ -5711,7 +5710,7 @@ namespace {
 			return;
 		}
 
-		std::vector<uint8_t> const new_priority = fix_priorities(files, m_torrent_file->files());
+		std::vector<boost::uint8_t> const new_priority = fix_priorities(files, m_torrent_file->files());
 
 		if (!m_torrent_file->num_pieces() || is_seed() || new_priority == m_file_priority) { return; }
 
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -5656,7 +5656,7 @@ namespace {
 
 			for (int i = 0; i < std::min<int>(fs.num_files(), files.size()); ++i)
 			{
-				if (files[i] < 0 || fs.pad_file_at(i))
+				if (files[i] > 0 && fs.pad_file_at(i))
 					files[i] = 0;
 				else if (files[i] > 7)
 					files[i] = 7;
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -400,7 +400,7 @@ namespace libtorrent
 		}
 	}
 
-	void torrent::inc_stats_counter(int c, int value)
+	void torrent::inc_stats_counter(int c, int value) const
 	{ m_ses.stats_counters().inc_stats_counter(c, value); }
 
 #if 0
@@ -1142,17 +1142,13 @@ namespace libtorrent
 #ifndef TORRENT_DISABLE_LOGGING
 		debug_log("*** set-share-mode: %d", s);
 #endif
-
-		// in share mode, all pieces have their priorities initialized to 0
-		if (m_share_mode && valid_metadata())
+		if (m_share_mode)
 		{
-			m_file_priority.clear();
-			m_file_priority.resize(m_torrent_file->num_files(), 0);
+			int const num_files = valid_metadata() ? m_torrent_file->num_files() : m_file_priority.size();
+			// in share mode, all pieces have their priorities initialized to 0
+			prioritize_files(std::vector<int>(num_files));
+			recalc_share_mode();
 		}
-
-		update_piece_priorities();
-
-		if (m_share_mode) recalc_share_mode();
 	}
 
 	void torrent::set_upload_mode(bool b)
@@ -2022,8 +2018,7 @@ namespace libtorrent
 
 		// in case file priorities were passed in via the add_torrent_params
 		// and also in the case of share mode, we need to update the priorities
-		if (!m_file_priority.empty() && std::find(m_file_priority.begin()
-				, m_file_priority.end(), 0) != m_file_priority.end())
+		if (std::find(m_file_priority.begin(), m_file_priority.end(), 0) != m_file_priority.end())
 		{
 			update_piece_priorities();
 		}
@@ -5655,6 +5650,23 @@ namespace {
 
 	namespace
 	{
+		std::vector<uint8_t> fix_priorities(std::vector<int> const& input, file_storage const& fs)
+		{
+			std::vector<uint8_t> files(input.begin(), input.end());
+
+			for (int i = 0; i < std::min<int>(fs.num_files(), files.size()); ++i)
+			{
+				if (files[i] < 0 || fs.pad_file_at(i))
+					files[i] = 0;
+				else if (files[i] > 7)
+					files[i] = 7;
+			}
+
+			files.resize(fs.num_files(), 4);
+
+			return files;
+		}
+
 		void set_if_greater(int& piece_prio, int file_prio)
 		{
 			if (file_prio > piece_prio) piece_prio = file_prio;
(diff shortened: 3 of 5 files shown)


include/libtorrent/torrent.hpp
```diff
@@ -1182,6 +1181,7 @@ namespace libtorrent
 		bool request_bandwidth_from_session(int channel) const;
 
 		void update_peer_interest(bool was_finished);
+		void update_piece_priorities();
```
d-komarov: Moved to private section

bindings/python/test.py
```diff
@@ -50,6 +50,8 @@ def test_torrent_handle(self):
 		self.assertEqual(self.h.piece_priorities(), [4])
 
 		self.h.prioritize_files([0,1])
+		# workaround for asynchronous priority update
+		time.sleep(1)
```
arvidn: good find. I thought the priorities were still updated immediately, but then corrected when the disk job come back, is that not the case?

src/torrent.cpp
```diff
@@ -5664,99 +5675,66 @@ namespace {
...
+				boost::bind(&torrent::on_file_priority, shared_from_this(), _1));
 		}
-
-		update_piece_priorities();
 	}
 
 	void torrent::set_file_priority(int index, int prio)
 	{
-		INVARIANT_CHECK;
+		if (index < 0 || (valid_metadata() && index >= m_torrent_file->num_files())) { return; }
```
arvidn: I would like to preserve the formatting style. and was there something wrong with the comment?
also, can you comment on why the invariant check and check for `is_seed()` was removed?

bindings/python/test.py
arvidn: I would prefer to restore the previous behavior here

src/torrent.cpp
```diff
@@ -1142,17 +1142,13 @@ namespace libtorrent
...
-		// in share mode, all pieces have their priorities initialized to 0
-		if (m_share_mode && valid_metadata())
+		if (m_share_mode)
 		{
-			m_file_priority.clear();
-			m_file_priority.resize(m_torrent_file->num_files(), 0);
+			int const num_files = valid_metadata() ? m_torrent_file->num_files() : m_file_priority.size();
+			// in share mode, all pieces have their priorities initialized to 0
+			prioritize_files(std::vector<int>(num_files));
+			recalc_share_mode();
```
arvidn: this seems like a reasonable refactor, but unrelated to the issue, right?

src/torrent.cpp
```diff
@@ -2022,8 +2018,7 @@ namespace libtorrent
 
 		// in case file priorities were passed in via the add_torrent_params
 		// and also in the case of share mode, we need to update the priorities
-		if (!m_file_priority.empty() && std::find(m_file_priority.begin()
-				, m_file_priority.end(), 0) != m_file_priority.end())
+		if (std::find(m_file_priority.begin(), m_file_priority.end(), 0) != m_file_priority.end())
```
arvidn: I would prefer to keep lines shorter than this

src/torrent.cpp
```diff
@@ -5664,99 +5675,66 @@ namespace {
...
 		boost::scoped_ptr<std::vector<boost::uint8_t> > p(j->buffer.priorities);
-		if (m_file_priority == *p) return;
 
-		// in this case, some file priorities failed to get set
-		m_file_priority = *p;
-		update_piece_priorities();
+		if (j->ret == piece_manager::no_error)
+		{
+			TORRENT_ASSERT(!j->error);
+			m_file_priority = *p;
```
arvidn: do we even need to assign the priorities in this case?

src/torrent.cpp
```diff
@@ -5664,99 +5675,66 @@ namespace {
 	void torrent::on_file_priority(disk_io_job const* j)
```
arvidn: is this the actual fix?

src/torrent.cpp
```diff
@@ -5664,99 +5675,66 @@ namespace {
...
+			return;
 		}
 
+		std::vector<boost::uint8_t> const new_priority = fix_priorities(files, m_torrent_file->files());
+
+		if (!m_torrent_file->num_pieces() || is_seed() || new_priority == m_file_priority) { return; }
+
 		// storage may be NULL during shutdown
-		if (m_torrent_file->num_pieces() > 0 && m_storage)
+		if (boost::shared_ptr<piece_manager> const tmp = m_storage)
```
arvidn: why do you need to make a copy of the storage pointer? surely `m_storage` won't get destroyed from under our feet here, before we issue the disk call

arvidn:
would you mind splitting these changes up into smaller, independent, pieces?
It would make it much simpler to see which changes belong together.

Does this patch make updating of priorities asynchronous? or did I do that in my patch? (I would prefer them not to be).

Perhaps a better solution would be for the disk job to send back just the diff of priorities. Or maybe we would only overwrite the file priorities if there's a disk error, and they deviate.

my understanding is that the fundamental case that's failing is:

1. set priorities -> issues async disk job in the background
2. set priorities to something else
3. the first disk job comes back and overwrites the priorities from (2)

src/torrent.cpp
d-komarov: This method is only about to construct priority vector and pass it to `torrent::prioritize_files` that has all checks preserved. Logic for adding disk job was also removed from here.

src/torrent.cpp
d-komarov: I'm not aware of current line length limitation so I assumed to align this to the longest line(# 2061) of the function.

src/torrent.cpp
d-komarov: This change actually fixes the issue.

src/torrent.cpp
d-komarov: Yes, it is in this function, also in `torrent::prioritize_files` and `torrent::set_file_priority`.

src/torrent.cpp
d-komarov: I'll overview lifetime management of storage object again, to give the correct answer.
But at the moment I was writing this, I thought that it could be possible to destroy storage from other thread by external plugins.

src/torrent.cpp
d-komarov: Each piece of code that deals with file priorities is presumably related.

d-komarov:
I'll try to handle this, but it'll take some time, because I'm going to leave for a week.

Yes, this patch makes the file priorities change asynchronous.

How that priority diff will be used? I'm not sure what solution is better. Need to consider both cons and pros to make up a decision.

Your understanding is correct.

bindings/python/test.py
d-komarov: I've added this to get python tests passed. This workaround depends on `torrent::prioritize_files` implementation that is considered in this PR.

src/torrent.cpp
d-komarov: Creating actual '.clang-format' file may be a good idea.

src/torrent.cpp
arvidn: so the relevant change here is to call `prioritize_files()` regardless of whether we have metadata or not, right?

src/torrent.cpp
arvidn: yes, I've considered that. I haven't had time to learn how to describe the existing formatting rules in libtorrent though. I didn't realise you had change this line other than reformatting it. This is also not a functional change, right?

src/torrent.cpp
```diff
@@ -5664,99 +5675,66 @@ namespace {
...
-		// initialize pad files to priority 0
-		file_storage const& fs = m_torrent_file->files();
-		for (int i = 0; i < (std::min)(fs.num_files(), limit); ++i)
+		// store file priorities until we get the metadata
+		if (!valid_metadata())
 		{
-			if (!fs.pad_file_at(i)) continue;
-			m_file_priority[i] = 0;
+			m_file_priority.assign(files.begin(), files.end());
+			return;
```
arvidn: I see, you're saving off the priorities here, so you can apply them immediately when we receive the metadata. is that right? (if so, please explain that in a comment)

src/torrent.cpp
```diff
@@ -5664,99 +5675,66 @@ namespace {
...
 		}
 
+		std::vector<boost::uint8_t> const new_priority = fix_priorities(files, m_torrent_file->files());
+
+		if (!m_torrent_file->num_pieces() || is_seed() || new_priority == m_file_priority) { return; }
```
arvidn: it seems reasonable to move the checks that are independent of `new_priority` before doing the work of `fix_priorites()`, no?

src/torrent.cpp
arvidn: `m_storage` is only supposed to be changed from the main (network) thread. most of libtorrent is single threaded in a sense. The exceptions are the `alert_manager` (which can be called directly from the user thread) and the disk threads (which is communicated with via disk jobs and post results back to the main thread, via the `io_service`)

bindings/python/test.py
arvidn: maybe you're right. It does make sense to apply file priorities asynchronously, since the alternative is to (potentially) lie about the priorities. If we can't apply it, we shouldn't ever say it was applied.

The problem with making it async though is that there really should be an alert to indicate that it was applied.

src/torrent.cpp
arvidn: I think we need to assign this unconditionally here. (i.e. hoist it out of the if-block). Even if there is an error reported from the disk thread, some priorities may have been applied.

src/torrent.cpp
arvidn: oh, I see you did leave a comment

src/torrent.cpp
arvidn: please expand the `{ return; }` block to match the existing formatting, or remove the `{` and `}`

src/torrent.cpp
```diff
@@ -5784,17 +5762,11 @@ namespace {
 	{
 		INVARIANT_CHECK;
 
-		if (!valid_metadata())
-		{
-			files->resize(m_file_priority.size());
-			std::copy(m_file_priority.begin(), m_file_priority.end(), files->begin());
-			return;
-		}
+		files->assign(m_file_priority.begin(), m_file_priority.end());
```
arvidn: this new code is not quite equivalent. previously the size of `files` was preserved. Now it isn't. Was that deliberate?

test/test_torrent.cpp
```diff
@@ -52,110 +52,117 @@ POSSIBILITY OF SUCH DAMAGE.
...
+			if (h.file_priorities() == prio) { return true; }
 
-	std::vector<boost::uint8_t> zeroes;
-	zeroes.resize(1000, 0);
-	add_torrent_params p;
-	p.flags &= ~add_torrent_params::flag_paused;
-	p.flags &= ~add_torrent_params::flag_auto_managed;
-	p.ti = info;
-	p.save_path = ".";
+			test_sleep(100);
```
arvidn: ideally this should be waiting for an alert. but I'm not sure it's possible to add a new alert without risking ABI incompatibilities. maybe it has to be added to `master`

arvidn:
I think your approach of setting file priorities async. is better. It makes more sense compared to the alternatives

bindings/python/test.py
d-komarov: Yes, I like the idea of introducing new alert for this operation. This what I thought of, when posted an issue for improvement.

src/torrent.cpp
d-komarov: No problem.

src/torrent.cpp
d-komarov: Yes, I believe that priorities must be handled in one place, that is `torrent::prioritize_files`. 'Empty metadata' case is covered there.

src/torrent.cpp
d-komarov: I'll try helping with this.

The reason why I changed this line, is that I thought `!m_file_priority.empty()` expression could be redundant here. `std::find` will work for an empty container just as well.

d-komarov: If you don't like it, I'll revert this change easily.

arvidn: no, this is fine

src/torrent.cpp
d-komarov: I'm not quite sure that this must be done, since `file_error_alert` and `torrent_error_alert` are not __critical__. That mean they could be lost and user won't have a chance to handle priority deviation.

d-komarov: I found a problem with this, `disk_io_thread::do_file_priority` always returns `no_error`, thus second branch will never be executed.

src/torrent.cpp
d-komarov: I believe that most compilers will optimize this code, i.e. `fix_priorities` is about to execute only when `!m_torrent_file->num_pieces() || is_seed()` checks are passed.

However I don't mind moving those checks up.

test/test_torrent.cpp
d-komarov: Good idea, but for now this should stay, isn't it?

src/torrent.cpp
arvidn: oh, I think it was intended to propagate file errors.
Whether the user receives the alert or not is somewhat separate. I think it's important for libtorrent's internal state to stay consistent regardless.
The user can always ask for the priorities.

src/torrent.cpp
arvidn: only if `fix_priorities()` and `torrent_file::files()` can be proven to be pure.

test/test_torrent.cpp
arvidn: yeah

src/torrent.cpp
```diff
@@ -5655,6 +5650,22 @@ namespace {
 
 	namespace
 	{
+		std::vector<boost::uint8_t> fix_priorities(std::vector<int> const& input, file_storage const& fs)
```
arvidn: factoring out this logic into a pure function is clearly a good thing. However, I'm a bit worried about merging this patch (correctly) into master. I'm a bit tempted to factor this function out in master instead of here, to keep this patch smaller

src/torrent.cpp
```diff
@@ -5664,99 +5675,66 @@ namespace {
...
+			pause();
+		}
 	}
 
 	void torrent::prioritize_files(std::vector<int> const& files)
 	{
 		INVARIANT_CHECK;
 
-		// this call is only valid on torrents with metadata
-		if (!valid_metadata() || is_seed()) return;
```
arvidn: I take it it's a deliberate change to allow changing file priorities even when you're a seed. What's the rationale there? the problem of settings files to priority 0 once they're downloaded is that it may be very expensive to copy into the partfile.

src/torrent.cpp
arvidn: I suppose this is required, because we need to convert `std::vector<int>` to `std::vector<std::uint8_t>` in order to make the disk thread call

src/torrent.cpp
```diff
@@ -5664,99 +5675,66 @@ namespace {
...
 	}
 
 	void torrent::set_file_priority(int index, int prio)
 	{
-		INVARIANT_CHECK;
+		if (index < 0 || (valid_metadata() && index >= m_torrent_file->num_files())) { return; }
 
-		if (is_seed()) return;
+		// any unallocated slot is assumed to be 4
+		if (int(m_file_priority.size()) <= index && prio == 4) { return; }
```
arvidn: I don't this is safe anymore. there may be a set-priority-job in flight, that sets this file priority to something else. A subsequent set-priority must update it, even if it's called back-to-back

test/test_torrent.cpp
```diff
@@ -248,63 +260,26 @@ TORRENT_TEST(added_peers)
...
 
-		// calculate the hash for all pieces
-		sha1_hash ph = hasher(&piece[0], piece.size()).final();
-		int num = t.num_pieces();
-		TEST_CHECK(t.num_pieces() > 0);
-		for (int i = 0; i < num; ++i)
-			t.set_hash(i, ph);
+	// calculate the hash for all pieces
+	TEST_CHECK(t.num_pieces() == 1);
+	t.set_hash(0, hasher(&piece[0], piece.size()).final());
```
arvidn: was it deliberate to only set the piece hash for the first piece? You made the torrent only have a single please. what's the rationale?

d-komarov: Previously there was only one piece too. Also I thought there was a mistake, torrent with only 1KB file and a 128 KB piece to calculate hash for. The last 127 KB of the piece should be padded with zeroes, and shouldn't participate in piece hash calculation. Does this make sense?

src/torrent.cpp
d-komarov: Yes, also to reassign invalid priorities with correct values.

d-komarov: Hmm, will it make patch's size much smaller? However it's not a problem to move this code back to the `torrent::prioritize_files`.

d-komarov: I left some comments on it. Please take a look.

src/torrent.cpp
d-komarov: This is true.

src/torrent.cpp
d-komarov: This check was relocated to the line 5715.

d-komarov:
Updated comments on https://github.com/d-komarov/libtorrent/pull/4

test/test_torrent.cpp
```diff
@@ -285,18 +295,15 @@ TORRENT_TEST(torrent)
...
+		std::vector<char> piece(1024);
 		for (int i = 0; i < int(piece.size()); ++i)
 			piece[i] = (i % 26) + 'A';
 
 		// calculate the hash for all pieces
-		sha1_hash ph = hasher(&piece[0], piece.size()).final();
-		int num = t.num_pieces();
-		TEST_CHECK(t.num_pieces() > 0);
-		for (int i = 0; i < num; ++i)
-			t.set_hash(i, ph);
```
arvidn: I thought I reverted the changes to this test. As far as I can tell, the logic change does not require making these changes to the test (and it will be harder to merge in master). would you mind reverting these changes? (obviously leave in the new tests you added and the required changes like `wait_priority()`)

test/test_torrent.cpp
```diff
@@ -81,48 +101,31 @@ void test_running_torrent(boost::shared_ptr<torrent_info> info, boost::int64_t f
...
 		TEST_EQUAL(st.total_wanted_done, 0);
-		if (!st.is_seeding)
-		{
-			TEST_EQUAL(int(h.file_priorities().size()), info->num_files());
-			TEST_EQUAL(h.file_priorities()[0], 0);
-			if (info->num_files() > 1)
-				TEST_EQUAL(h.file_priorities()[1], 0);
-			if (info->num_files() > 2)
-				TEST_EQUAL(h.file_priorities()[2], 1);
-		}
```
arvidn: I think these tests should still be here, or is there a reason to remove them?

arvidn:
would you mind squashing these commits to?

arvidn:
please include a not in the ChangeLog too

d-komarov:
> please include a not in the ChangeLog too

I'm not quite sure what do you mean.

test/test_torrent.cpp
d-komarov: Everything that is being tested here is now covered by `TEST_CHECK(prioritize_files(h, prio))`, plus it also checks the rest of priorities that should have been set.

arvidn:
I meant to say "note". i.e. line describing the bug fix in the `ChangeLog` file

test/test_torrent.cpp
```diff
@@ -285,16 +299,16 @@ TORRENT_TEST(torrent)
 		file_storage fs;
 
 		fs.add_file("test_torrent_dir2/tmp1", 1024);
-		libtorrent::create_torrent t(fs, 128 * 1024, 6);
+		libtorrent::create_torrent t(fs, 1024, 6);
```
arvidn: this used to be 128 blocks

test/test_torrent.cpp
```diff
@@ -263,7 +278,6 @@ TORRENT_TEST(torrent)
 		std::vector<char> piece(128 * 1024);
 		for (int i = 0; i < int(piece.size()); ++i)
 			piece[i] = (i % 26) + 'A';
-
```
arvidn: please add back this blank line too. these tests have changed in master (I'm pretty sure) and this seems like an unnecessary conflict.

d-komarov:
Oh, sure.

test/test_torrent.cpp
d-komarov: This is commented code, that was removed in original PR. Are you sure it really make sense?

test/test_torrent.cpp
d-komarov: The second argument is a `piece_size`. Setting it back to 128 * 1024 will require additional changes in `test_running_torrent()`.

test/test_torrent.cpp
arvidn: it will still generate a conflict with master.

test/test_torrent.cpp
```diff
@@ -52,10 +52,34 @@ POSSIBILITY OF SUCH DAMAGE.
...
+{
+	h.prioritize_files(prio);
+	return wait_priority(h, prio);
+}
+
 void test_running_torrent(boost::shared_ptr<torrent_info> info, boost::int64_t file_size)
 {
 	settings_pack pack = settings();
-	pack.set_int(settings_pack::alert_mask, alert::storage_notification);
+	pack.set_int(settings_pack::alert_mask, alert::progress_notification | alert::storage_notification);
```
arvidn: what's the `progress_notification` needed for?

test/test_torrent.cpp
d-komarov: Alternative fix 13e92c31.

arvidn: A lot of this buffer juggling is a lot nicer in master, with `span`

d-komarov: The problem is not in particular buffer type, it is about how piece size for the single-file torrent is being obtained.

arvidn: I don't think I actually understand this (yet). Is this an independent problem from setting the file priorities?

d-komarov: From this test I discovered the problem for the first time. At that moment I didn't have a clue on what's causing test failure. So I began to investigate and had found some issues in it. When I fixed em', test still was failing randomly. Soon I found a root cause and it was file priorities, so I fixed it too.

TL;DR: Test issue appears to be independent, and could be submitted as a separate PR.

d-komarov: Without that changes test should be passing, but will take extra 10 seconds to finish. Also it is error prone, because of incorrect piece handling logic.

d-komarov: You decide, whether fixing test issues goes to this PR or another one. I just want this test be correct in the end.

PR #3136, opened 2018-06-30
qualify some assignment operators to disallow assignment to temporaries

arvidn:


aldenml:
just to let you know, that this change (like others) is pushing the swig parsers to it's limits, need to solve this one

PR #3137, opened 2018-06-30
fix some unintentional copies (with explicit moves)

arvidn:


aldenml:
I was under the impression that RVO was at play here. Do you mind commenting a little about why this is needed? Thanks

arvidn:
Yeah, NRVO and implicit move only happens when the return type match the type of the returned value. In these cases the objects are deliberately sliced, which inhibits implicit move.

aldenml:
Thanks! didn't know that restriction

PR #3139, opened 2018-07-01
Proposed changes to resulting patch.

d-komarov:


src/torrent.cpp
```diff
@@ -5707,6 +5714,8 @@ namespace {
 
 		std::vector<boost::uint8_t> const new_priority = fix_priorities(files, m_torrent_file->files());
 
+		if (m_file_priority == new_priority) return;
```
arvidn: see my comment on this in my code review

src/torrent.cpp
```diff
@@ -5737,6 +5743,8 @@ namespace {
 		if (prio < 0) prio = 0;
 		else if (prio > 7) prio = 7;
 
+		if (m_file_priority.size() > index && m_file_priority[index] == prio) return;
```
arvidn: and this

src/torrent.cpp
d-komarov: You're right, let's drop those checks for now.

src/torrent.cpp
```diff
@@ -5684,12 +5686,17 @@ namespace {
 		{
 			m_file_priority = *p;
 			update_piece_priorities();
+
+			if (m_share_mode)
+			{
+				TORRENT_ASSERT(boost::algorithm::all_of_equal(m_file_priority, 0));
```
arvidn: this assert is not a safe assumption, since `m_share_mode` is updated immediately when entering share mode, it doesn't first wait to make sure all outstanding set-priority jobs complete.

src/torrent.cpp
```diff
@@ -5726,9 +5735,6 @@ namespace {
 
 		if (is_seed()) return;
 
-		// setting file priority on a torrent that doesn't have metadata yet is
-		// similar to having passed in file priorities through add_torrent_params.
-		// we store the priorities in m_file_priority until we get the metadata
```
arvidn: I think this comment is useful. it should probably be moved down to the assignment to `m_file_priority` in the else-block. But I would prefer to do that in master to make this simpler to merge

src/torrent.cpp
d-komarov: Let's drop it too

src/torrent.cpp
d-komarov: Agreed.

arvidn:
I have incorporated your feedback into my PR

d-komarov:
Yes, I see.

PR #3143, opened 2018-07-02
Fix torrent test

d-komarov:


arvidn:
was this intended to sit on top of the `file_priority` patch?

d-komarov:
Not really, this is a temporary branch to show you how much changes is required to fix 'torrent' test with `128 * 1024` piece size. This was your question from initial PR.
Consider only commit b151351.

d-komarov:
I'm closing this PR, because similar one has been merged already.

PR #3144, opened 2018-07-04
deprecate (and disable) the force-proxy setting

arvidn:
Instead, always use the proxy when set, never fall back on circumventing it


arvidn:
@sledgehammer999 @cas-- Do you have any strong opinions about this?
The reason `force_proxy` existed in the first place was because in the very beginning of adding support for proxies, it was done as best effort. If the proxy appeared to be down, try without a proxy. (this was treating proxies as the traditional deployment where you have a local network that's firewalled and only allow access to the outside world via a proxy).

My impression since than is that people want to use a proxy, and never circumvent it if it doesn't work. So force-proxy was a way to indicate that all communication had to go through a proxy.

This patch is undoing some of this mess (which probably was a mistake from the beginning), by always honouring the proxy setting, whether the proxy works or not. The new semantics is as if `force_proxy` is set whenever a proxy is configured, and not set when a proxy hasn't been configured.

I also discovered some issues around `proxy_tracker_connections` and `proxy_peer_connections`. If one exempts one kind of traffic from the proxy, `force_proxy` would still block it, because it wasn't going through a proxy.

cas--:
That seems like a reasonable implementation to me. I think most users have
that assumption.

PR #3146, opened 2018-07-04
Test merge

arvidn:


d-komarov:
OK, I'l take a look tomorrow.

src/torrent.cpp
```diff
@@ -979,17 +980,15 @@ bool is_downloading_state(int const st)
...
+		if (m_share_mode)
 		{
-			m_file_priority.clear();
-			m_file_priority.resize(m_torrent_file->num_files(), dont_download);
+			std::size_t const num_files = valid_metadata()
+				? std::size_t(m_torrent_file->num_files())
+				: m_file_priority.size();
+			// in share mode, all pieces have their priorities initialized to
+			// dont_download
+			prioritize_files(aux::vector<download_priority_t, file_index_t>(num_files, dont_download));
```
d-komarov: I guess this line might be exceeding length limit.

src/torrent.cpp
```diff
@@ -5017,15 +5016,46 @@ bool is_downloading_state(int const st)
...
+				// initialize pad files to priority 0
+				if (files[i] > dont_download && fs && fs->pad_file_at(i))
+					files[i] = dont_download;
+				else if (files[i] > top_priority)
+					files[i] = top_priority;
+			}
+
+			if (fs)
+			{
+				files.resize(fs->num_files(), default_priority);
```
d-komarov: This line should be placed before loop, in order to reset pad files priorities.

src/torrent.cpp
```diff
@@ -5078,33 +5096,33 @@ bool is_downloading_state(int const st)
 		// setting file priority on a torrent that doesn't have metadata yet is
 		// similar to having passed in file priorities through add_torrent_params.
 		// we store the priorities in m_file_priority until we get the metadata
-		file_storage const& fs = m_torrent_file->files();
-		if (index < file_index_t(0) || (valid_metadata() && index >= fs.end_file()))
+		if (index < file_index_t(0)
```
d-komarov: Is `file_index_t` type allowed to have negative values? What does it represent with negative value?

src/torrent.cpp
```diff
@@ -5078,33 +5096,33 @@ bool is_downloading_state(int const st)
...
 		if (m_storage)
 		{
 			ADD_OUTSTANDING_ASYNC("file_priority");
 			m_ses.disk_thread().async_set_file_priority(m_storage
-				, m_file_priority, std::bind(&torrent::on_file_priority, shared_from_this(), _1, _2));
+				, std::move(new_priority), std::bind(&torrent::on_file_priority, shared_from_this(), _1, _2));
+		}
+		else
+		{
+			m_file_priority = new_priority;
```
d-komarov: Using `swap` is more effective then copy assignment. Using move assignment is also fine.

src/torrent.cpp
arvidn: it depends on what you mean by "allowed". It refers to an invalid index, similar to an index thats >= the number of files

d-komarov: Ah, got it.

test/test_torrent.cpp
```diff
@@ -81,48 +103,32 @@ void test_running_torrent(std::shared_ptr<torrent_info> info, std::int64_t file_
 		return;
 	}
 
-	aux::vector<download_priority_t, file_index_t> ones(std::size_t(info->num_files()), 1_pri);
```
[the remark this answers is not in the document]
d-komarov: No problem.

arvidn:
@d-komarov I updated this patch too

d-komarov:
Yes. Though some comments still need your attention.

d-komarov:
Looks good to me now.

PR #3147, opened 2018-07-05
fix recent patch to support setting file priorities on torrents that don't have metadata

arvidn:
@d-komarov could you please take a look? I broke this in one of my last changes on that previous patch. I discovered this when merging it into master

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -5650,20 +5650,21 @@ namespace {
 
 	namespace
 	{
-		std::vector<boost::uint8_t> fix_priorities(std::vector<int> const& input, file_storage const& fs)
+		std::vector<boost::uint8_t> fix_priorities(std::vector<int> const& input
+			, file_storage const* fs)
 		{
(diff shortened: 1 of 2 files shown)


src/torrent.cpp
```diff
@@ -5650,20 +5650,22 @@ namespace {
 
 	namespace
 	{
-		std::vector<boost::uint8_t> fix_priorities(std::vector<int> const& input, file_storage const& fs)
+		std::vector<boost::uint8_t> fix_priorities(std::vector<int> const& input
+			, file_storage const* fs)
 		{
 			std::vector<boost::uint8_t> files(input.begin(), input.end());
+			if (fs && files.size() > fs->num_files()) files.resize(fs->num_files());
```
d-komarov: There could a problem here, if `files` vector size is less then a number of files in torrent, we'll fix it as is, and then resize it, filling with 4.
What if some of that tail files are paddings? Resulting vector will have non-zero paddings priorities.
I think line 5668 should be moved here to replace this one.

PR #3149, opened 2018-07-06
Fix deadlock when loading libtorrent DLL

d-komarov:
Definition of `bool file::has_manage_volume_privs` involves a call to
`get_manage_volume_privs()`, causing restricted tasks to be performed
from within `DllMain` function. They introduce possibility that client
application deadlocks or crashes.

You should never perform the following tasks from within DllMain:

* Call `LoadLibrary` or `LoadLibraryEx` (either directly or indirectly).
  This can cause a deadlock or a crash.

* Call the registry functions. These functions are implemented in
  'Advapi32.dll'. If not initialized before your DLL, it can access
  uninitialized memory and cause the process to crash.

See [Dynamic-Link Library Best Practices](https://docs.microsoft.com/en-us/windows/desktop/dlls/dynamic-link-library-best-practices).

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -79,6 +79,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <stdio.h>
 #include "libtorrent/assert.hpp"
 
+#include <boost/scope_exit.hpp>
 #include <boost/scoped_ptr.hpp>
 #include <boost/static_assert.hpp>
 
@@ -1314,9 +1315,6 @@ namespace libtorrent
 
 #ifdef TORRENT_WINDOWS
 	bool get_manage_volume_privs();
-
-	// this needs to be run before CreateFile
-	bool file::has_manage_volume_privs = get_manage_volume_privs();
 #endif
 
 	file::file()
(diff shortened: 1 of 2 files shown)


src/file.cpp
```diff
@@ -1996,38 +2002,23 @@ typedef struct _FILE_ALLOCATED_RANGE_BUFFER {
...
 		return ret;
 	}
 
 	void set_file_valid_data(HANDLE f, boost::int64_t size)
 	{
 		typedef BOOL (WINAPI *SetFileValidData_t)(HANDLE, LONGLONG);
-		static SetFileValidData_t pSetFileValidData = NULL;
-		static bool failed_kernel32 = false;
+		static SetFileValidData_t const pSetFileValidData = (SetFileValidData_t)
+				GetProcAddress(GetModuleHandleA("kernel32"), "SetFileValidData");
```
d-komarov: "kernel32" will always be present since it exports `LoadLibrary()`.

src/file.cpp
```diff
@@ -1996,38 +2002,23 @@ typedef struct _FILE_ALLOCATED_RANGE_BUFFER {
...
-			if (pSetFileValidData == NULL)
-			{
-				failed_kernel32 = true;
-				return;
-			}
-		}
+		if (pSetFileValidData == NULL) return;
 
-		TORRENT_ASSERT(pSetFileValidData);
+		static bool const has_manage_volume_privs = get_manage_volume_privs();
```
d-komarov: Postpone initialization of `has_manage_volume_privs`.

src/file.cpp
```diff
@@ -1956,10 +1954,15 @@ typedef struct _FILE_ALLOCATED_RANGE_BUFFER {
 		static LookupPrivilegeValue_t pLookupPrivilegeValue = NULL;
 		static AdjustTokenPrivileges_t pAdjustTokenPrivileges = NULL;
 		static bool failed_advapi = false;
+		HMODULE advapi = NULL;
+
+		BOOST_SCOPE_EXIT(&advapi) {
+			if (advapi) FreeLibrary(advapi);
+		} BOOST_SCOPE_EXIT_END
```
d-komarov: Used `RAII` to always free the library.

d-komarov:
That was done already )

arvidn:
I was just pointing out that the original code was questionable even without this bug, and another reason in favour of your patch

src/file.cpp
arvidn: technically this is a race condition in C++98, as static initializations aren't thread safe. But I would expect it to be a small problem in practice, especially given that mot popular compilers support C++11 by now.

d-komarov: Also multiple calls to `get_manage_volume_privs()` from different threads aren't a problem, since they would return the same value.

PR #3151, opened 2018-07-07
Test merge

arvidn:


src/file.cpp
```diff
@@ -476,9 +476,6 @@ static_assert(!(open_mode::sparse & open_mode::attribute_mask), "internal flags
 
 #ifdef TORRENT_WINDOWS
 	bool get_manage_volume_privs();
-
-	// this needs to be run before CreateFile
```
arvidn: @d-komarov this comment here suggests that the recent change runs it too late. In fact, in `master`, the result of `get_manage_volume_privs()` is not used, but I believe whatever is done in there is still important.

d-komarov: Yeah I see, I'll submit another patch to fix this.

arvidn: against `RC_1_1`, right? (please add the comment there too)

d-komarov: Right, no problem.

arvidn: I will land this patch and fix it by merging your fix in here again. Maybe running the function on the `file` constructor would be reasonable. In fact, in master it should be called with `std::call_once()`

PR #3153, opened 2018-07-07
fix some validation issues in read_resume_data()

arvidn:


src/read_resume_data.cpp
```diff
@@ -290,31 +296,31 @@ namespace {
 		if (bdecode_node const peers_entry = rd.dict_find_string("peers"))
 		{
 			char const* ptr = peers_entry.string_ptr();
-			for (int i = 0; i < peers_entry.string_length(); i += 6)
+			for (int i = 0; i < peers_entry.string_length() - 6; i += 6)
```
aldenml: are you really sure of these changes in the loop condition? doesn't this prevent the reading of the last one?

arvidn: yeah, probably. that might explain why the tests are failing too

PR #3156, opened 2018-07-08
raise priority of cache_flushed_alert

arvidn:
and post it unconditionally when triggered explicitly by the client. This gives the alert the same priority as `torrent_finished_alert`


d-komarov:
Yes, I think this could be a solution for 1.1.x, but master branch should probably overhaul  `torrent_finished_alert` logic, to make this alert to be triggered after all data has been flushed to disk.

PR #3158, opened 2018-07-08
Fix previous patch fixing deadlock in DllMain

d-komarov:
Make `get_manage_volume_privs()` to acquire privelege only once.
Subsequent calls will be reporting value that was evaluated on 1st run.

* Add namespace ending comment
* Update ChangeLog

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -1959,7 +1959,7 @@ typedef struct _FILE_ALLOCATED_RANGE_BUFFER {
 
 		static BOOL result = -1;
 
-		if (result != -1) return !!result;
+		if (result != -1) return result == TRUE;
 
 		result = FALSE;
 		HMODULE advapi = LoadLibraryA("advapi32");
@@ -2006,7 +2006,7 @@ typedef struct _FILE_ALLOCATED_RANGE_BUFFER {
 		result = pAdjustTokenPrivileges(token, FALSE, &privs, 0, NULL, NULL)
 			&& GetLastError() == ERROR_SUCCESS;
 
-		return !!result;
+		return result == TRUE;
 	}
 
 	void set_file_valid_data(HANDLE f, boost::int64_t size)
@@ -2015,8 +2015,10 @@ typedef struct _FILE_ALLOCATED_RANGE_BUFFER {
 		static SetFileValidData_t const pSetFileValidData = (SetFileValidData_t)
 				GetProcAddress(GetModuleHandleA("kernel32"), "SetFileValidData");
 
-		if (pSetFileValidData && get_manage_volume_privs())
+		if (pSetFileValidData)
 		{
+			// we don't necessarily expect to have enough
+			// privilege to do this, so ignore errors.
 			pSetFileValidData(f, size);
 		}
 	}
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -1314,7 +1314,7 @@ namespace libtorrent
 
 
 #ifdef TORRENT_WINDOWS
-	void get_manage_volume_privs();
+	void acquire_manage_volume_privs();
 #endif
 
 	file::file()
@@ -1430,7 +1430,7 @@ namespace libtorrent
 		{
 			// Enable privilege required by SetFileValidData()
 			// https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-setfilevaliddata
-			get_manage_volume_privs();
+			acquire_manage_volume_privs();
 		}
 
 		handle_type handle = CreateFile_(file_path.c_str(), m.rw_mode
@@ -1937,7 +1937,7 @@ typedef struct _FILE_ALLOCATED_RANGE_BUFFER {
 	}
 
 #ifdef TORRENT_WINDOWS
-	void get_manage_volume_privs()
+	void acquire_manage_volume_privs()
 	{
 		static bool called_once = false;
 
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -1314,7 +1314,7 @@ namespace libtorrent
 
 
 #ifdef TORRENT_WINDOWS
-	bool get_manage_volume_privs();
+	void get_manage_volume_privs();
 #endif
 
 	file::file()
@@ -1937,8 +1937,14 @@ typedef struct _FILE_ALLOCATED_RANGE_BUFFER {
 	}
 
 #ifdef TORRENT_WINDOWS
-	bool get_manage_volume_privs()
+	void get_manage_volume_privs()
 	{
+		static bool called_once = false;
+
+		if (called_once) return;
+
+		called_once = true;
+
 		typedef BOOL (WINAPI *OpenProcessToken_t)(
 			HANDLE ProcessHandle,
 			DWORD DesiredAccess,
@@ -1957,14 +1963,9 @@ typedef struct _FILE_ALLOCATED_RANGE_BUFFER {
 			PTOKEN_PRIVILEGES PreviousState,
 			PDWORD ReturnLength);
 
-		static BOOL result = -1;
-
-		if (result != -1) return result == TRUE;
-
-		result = FALSE;
 		HMODULE advapi = LoadLibraryA("advapi32");
 
-		if (advapi == NULL) return false;
+		if (advapi == NULL) return;
 
 		BOOST_SCOPE_EXIT(&advapi) {
 			FreeLibrary(advapi);
@@ -1981,13 +1982,13 @@ typedef struct _FILE_ALLOCATED_RANGE_BUFFER {
 			|| pLookupPrivilegeValue == NULL
 			|| pAdjustTokenPrivileges == NULL)
 		{
-			return false;
+			return;
 		}
 
 		HANDLE token = NULL;
 		if (!pOpenProcessToken(GetCurrentProcess()
 			, TOKEN_ADJUST_PRIVILEGES | TOKEN_QUERY, &token))
-			return false;
+			return;
 
 		BOOST_SCOPE_EXIT(&token) {
 			CloseHandle(token);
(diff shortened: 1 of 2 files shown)


src/file.cpp
```diff
@@ -2011,11 +2015,7 @@ typedef struct _FILE_ALLOCATED_RANGE_BUFFER {
 		static SetFileValidData_t const pSetFileValidData = (SetFileValidData_t)
 				GetProcAddress(GetModuleHandleA("kernel32"), "SetFileValidData");
 
-		if (pSetFileValidData == NULL) return;
-
-		static bool const has_manage_volume_privs = get_manage_volume_privs();
-
-		if (has_manage_volume_privs)
+		if (pSetFileValidData && get_manage_volume_privs())
```
arvidn: I don't think you have to call this function here. It's already been called once before you get here. The approach in master was to ignore the result from it, and try the call to `SetFileValidData()` anyway, and let it fail, so I don't think there's any need for extra checking here either

src/file.cpp
```diff
@@ -1426,6 +1426,13 @@ namespace libtorrent
 			| FILE_FLAG_OVERLAPPED
 			| ((mode & no_cache) ? FILE_FLAG_WRITE_THROUGH : 0);
 
+		if ((mode & sparse) == 0)
+		{
+			// Enable privilege required by SetFileValidData()
+			// https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-setfilevaliddata
+			get_manage_volume_privs();
```
arvidn: Is there a good reason to move the "call once" logic into the `get_manage_volume_privs()` function?
I would prefer the keep this patch as small as possible. Something like just initializing a `static` variable here with its call. I'm also not sure it's a good idea to condition this on whether the `sparse` flag is set or not. It could be unset later, which would complicate the logic for calling this function.

I think it would be better (and simpler) to keep `get_manage_volume_privs()` as it is, and just make sure it's called once in the constructor.

src/file.cpp
```diff
@@ -1950,34 +1957,31 @@ typedef struct _FILE_ALLOCATED_RANGE_BUFFER {
...
 			PDWORD ReturnLength);
 
-		static OpenProcessToken_t pOpenProcessToken = NULL;
-		static LookupPrivilegeValue_t pLookupPrivilegeValue = NULL;
-		static AdjustTokenPrivileges_t pAdjustTokenPrivileges = NULL;
-		static bool failed_advapi = false;
-		HMODULE advapi = NULL;
+		static BOOL result = -1;
+
+		if (result != -1) return !!result;
```
arvidn: I think it's better to use `bool` than `BOOL`, and I also think an explicit conversion to bool, `bool(X)` is better than the double negation trick `!!X`. But by just using the standard type, this conversion isn't necessary

src/file.cpp
d-komarov: Why should we ask system to do something that is known to fail? This call is cheap because we already know result. Why not using that result to avoid call that would fail?

src/file.cpp
d-komarov: OK, I'll remove that condition.

src/file.cpp
d-komarov: I'll replace `!!` with explicit cast to bool.

src/file.cpp
arvidn: because I'm not confident there is a 1:1 relationship between the result and whether the call will succeed. I have not found very clear documentation around this.

d-komarov: I've added link to related doc in the comment. See `open_file()`.

d-komarov: Here is also a good reading on this from Raymond Chen:
https://blogs.msdn.microsoft.com/oldnewthing/20160603-00/?p=93565

arvidn: that example code doesn't call `AdjustTokenPrivileges()` nor does it suggest that the `SetFileValidData()` can only succeed if the call to `AdjustTokenPrivileges()` succeeds. For instance, I would expect that `AdjustTokenPrivileges()` might fail for some reason other than lack of privileges.

d-komarov: Please pay attention to [SetPrivelege](https://msdn.microsoft.com/library/windows/desktop/aa446619(v=vs.85).aspx), it hides a hyperlink.

d-komarov: Possible corner cases also should include this, enabling required privilege once doesn't make sense since other piece of code could simply disable that privilege at any time.

src/file.cpp
d-komarov: > Is there a good reason to move the "call once" logic into the get_manage_volume_privs() function?

Since I'm not sure that this function should really be called only once, moving that logic to the function implementation makes sense. Docs tell us that `SetValidFileData()` require file handle to be initially opened with `SeManageVolumePrivilege` acquired. If we enable it once, there is no guaranty that other code won't disable this privilege afterwards. Thus `get_manage_volume_privs()` is called each time before `CreateFile`, but the hidden 'call once' logic makes it work as before.

src/file.cpp
```diff
@@ -1950,34 +1957,31 @@ typedef struct _FILE_ALLOCATED_RANGE_BUFFER {
...
-			pOpenProcessToken = (OpenProcessToken_t)GetProcAddress(advapi, "OpenProcessToken");
-			pLookupPrivilegeValue = (LookupPrivilegeValue_t)GetProcAddress(advapi, "LookupPrivilegeValueA");
-			pAdjustTokenPrivileges = (AdjustTokenPrivileges_t)GetProcAddress(advapi, "AdjustTokenPrivileges");
-			if (pOpenProcessToken == NULL
-				|| pLookupPrivilegeValue == NULL
-				|| pAdjustTokenPrivileges == NULL)
-			{
-				failed_advapi = true;
-				return false;
-			}
```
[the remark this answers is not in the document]
arvidn: Ah, because `failed_advapi == true` doesn't make the function return with a failure (I think that was the intention).

You may have noticed that this function is quite different in `master`, with a utility function for looking up functions out of DLLs, that's why I'm interested in keeping these changes small and simple. But This issue clearly needs to be fixed.

src/file.cpp
d-komarov: Also I believe that `(mode & sparse) == 0` condition is correct and should persist, because file just can't change sparse mode on the fly.

arvidn: that's a good point. You don't think calling it on every file-open will add an unacceptable slow down?

d-komarov: Everything is possible. For now it won't because of hidden call-once logic. Disabling that logic is definitely requires further consideration, that is out of current scope.

Also out of scope, but I think that using that API should be restricted by introducing special flag, that the user must explicitly specify. In such case user becomes responsible for related security flaws.

arvidn: the security risks are mitigated somewhat by libtorrent itself tracking which parts of a file has been written, and won't serve up "garbage" data (unless coerced by invalid resume data). It turns out people run bittorrent clients as administrator just to get access to this feature.

d-komarov: Did you consider the case when application crash or spontaneous reboot happens after that API call but before libtorrent manage to write any data to file. Any user will be able to explore sensible data that previously belonged to some file that was removed.

arvidn: yes, in all such cases the resume file will indicate that we don't have those pieces, and they won't be advertised to peers and requests for them will be rejected.

d-komarov: No, that's not what I worry about. What you call a "garbage data" potentially could be a wide variety of sensitive information such as stored passwords, crypto-keys, etc. All this could become visible to any local user with limited permissions. Someone would call it a __vulnerability__.

arvidn: right, if the person with administrator privileges make the file accessible to other users. Currently you need to run the client with administrator access though, which hasn't been the default since XP, so it's quite deliberate.

src/file.cpp
```diff
@@ -1999,10 +2003,10 @@ typedef struct _FILE_ALLOCATED_RANGE_BUFFER {
 		privs.PrivilegeCount = 1;
 		privs.Privileges[0].Attributes = SE_PRIVILEGE_ENABLED;
 
-		bool ret = pAdjustTokenPrivileges(token, FALSE, &privs, 0, NULL, NULL)
+		result = pAdjustTokenPrivileges(token, FALSE, &privs, 0, NULL, NULL)
 			&& GetLastError() == ERROR_SUCCESS;
```
arvidn: it looks a bit suspicious to not use a `bool` here for `result`. The resulting type of `operator&&` is `bool`, not `BOOL`. perhaps the result from `pAdjustTokenPrivileges()` should be checked against `TRUE` though.

Also, it seem reasonable to just return directly, rather than first storing the result in `result`, that would make the question of type a non-issue. the `master` branch is more strict with types as all implicit conversion warnings are enabled there

d-komarov: Standard conformant. `bool` to `int` conversion is implicit, and `BOOL` stands for `int` in our case.

§4.7/4 from the C++ Standard says (Integral Conversion)

> If the source type is bool, the value false is converted to zero and the value true is converted to one.

arvidn: I'm not suggesting the code is invalid. I would just prefer to not rely on implicit conversions. That's why all conversion warnings have been enabled in `master`

d-komarov:
Do you think it's a problem, if this PR goes to RC_1_1 without merging into master?

src/file.cpp
arvidn: I agree. I think it's bonkers that Microsoft still support this feature

arvidn:
well, it has to be fixed in master too, and I would like to be able to (easily) merge other fixes from RC_1_1 into master, which means git should record that everything in RC_1_1 has made it into master (even if the merge commit completely rewrites this code)

src/file.cpp
d-komarov: Can you confirm that there is a warning caused by this conversion? I supposed there shouldn't be any.

d-komarov:
Yeah, I have some understanding how git merge works. But whatever these changes in RC_11, there still gonna be a conflict while merging to master and I suggest resolving it by choosing master's version of change.

d-komarov:
OK, I think I've got your point. I'll make another commit to reduce the number of changes.

arvidn:
returning void seems reasonable to me. maybe it shouldn't be called `get_` though. or maybe `get` is good name for it, or maybe `install` or `adjust`.

d-komarov:
Maybe it should be "acquire"?

arvidn:
sounds good

src/file.cpp
arvidn: https://clang.llvm.org/extra/clang-tidy/checks/readability-implicit-bool-conversion.html

d-komarov:
Do you think anything else should be changed?

PR #3159, opened 2018-07-08
fix port parameter in DHT announces on multi-homed hosts

ssiloti:
Use the correct listen port for each DHT node's external address instead of using the same port for all nodes.

### DIFF
--- a/include/libtorrent/kademlia/dht_tracker.hpp
+++ b/include/libtorrent/kademlia/dht_tracker.hpp
@@ -96,7 +96,7 @@ namespace libtorrent { namespace dht {
 
 		dht_state state() const;
 
-		enum flags_t { flag_seed = 1, flag_implied_port = 2 };
+		enum flags_t { flag_seed = 1, flag_implied_port = 2, flag_ssl_torrent = 4 };
 		void get_peers(sha1_hash const& ih
 			, std::function<void(std::vector<tcp::endpoint> const&)> f);
 		void announce(sha1_hash const& ih, int listen_port, int flags
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -5438,6 +5438,24 @@ namespace aux {
 		return 0;
 	}
 
+	int session_impl::get_listen_port(transport const ssl, aux::listen_socket_handle const& s)
+	{
+		auto socket = s.get();
+		if (socket->ssl != ssl)
+		{
+			auto alt_socket = std::find_if(m_listen_sockets.begin(), m_listen_sockets.end()
+				, [&](std::shared_ptr<listen_socket_t> const& e)
+			{
+				return e->ssl == ssl
+					&& e->external_address.external_address()
+						== socket->external_address.external_address();
+			});
+			if (alt_socket != m_listen_sockets.end())
+				socket = alt_socket->get();
+		}
+		return socket->udp_external_port;
+	}
+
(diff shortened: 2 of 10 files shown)


src/session_impl.cpp
```diff
@@ -5988,7 +6006,14 @@ namespace aux {
...
 	{
 		if (!m_dht) return;
-		m_dht->announce(info_hash, port, flags, std::bind(&on_dht_get_peers, std::ref(m_alerts), info_hash, _1));
+		if (port == 0)
+			m_dht->announce(info_hash, flags
+				, std::bind(&session_impl::get_listen_port, this, transport::plaintext, _1)
+				, std::bind(&on_dht_get_peers, std::ref(m_alerts), info_hash, _1));
+		else
+			m_dht->announce(info_hash, flags
+				, [port](aux::listen_socket_handle const&) { return port; }
```
arvidn: is this case still necessary?

arvidn: it seems `port` could just be dropped instead

include/libtorrent/kademlia/dht_tracker.hpp
```diff
@@ -99,7 +99,8 @@ namespace libtorrent { namespace dht {
 		enum flags_t { flag_seed = 1, flag_implied_port = 2 };
 		void get_peers(sha1_hash const& ih
 			, std::function<void(std::vector<tcp::endpoint> const&)> f);
-		void announce(sha1_hash const& ih, int listen_port, int flags
+		void announce(sha1_hash const& ih, int flags
+			, std::function<int(aux::listen_socket_handle const& s)> listen_port_fn
```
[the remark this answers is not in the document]
ssiloti: Good call, I didn't think to use `dht_observer`.

src/session_impl.cpp
ssiloti: I kept the `port` parameter so that a client can announce non-bittorrent services to the DHT.

arvidn:
what's up with the SSL link errors? perhaps travis updated something. That doesn't really explain appveyor though

aldenml:
sounds to me like a version of boost not playing nice with the version of openssl used, in appveyor is picking `boost_1_63_0`, very old

src/torrent.cpp
```diff
@@ -2603,12 +2597,17 @@ bool is_downloading_state(int const st)
...
 		// likely more accurate when behind a NAT
-		if (settings().get_bool(settings_pack::enable_incoming_utp))
+		// don't set this for SSL torrents because peers need
+		// to connect to the SSL listen port
+		if (settings().get_bool(settings_pack::enable_incoming_utp)
+			&& !is_ssl_torrent())
 			flags |= dht::dht_tracker::flag_implied_port;
 
+		if (is_ssl_torrent())
+			flags |= dht::dht_tracker::flag_ssl_torrent;
```
arvidn: does this do the correct thing for SSL torrents where `settings_pack::enable_incoming_utp` is set to `false`? It probably doesn't matter that much though.

It seems like there may be a simpler way of phrasing this logic. Perhaps by checking `is_ssl_torrent()` first, something along these lines:

```
if (is_ssl_torrent())
{
   flags |= dht::dht_tracker::flag_ssl_torrent;
}
else
{
   if (settings().get_bool(settings_pack::enable_incoming_utp))
      flags |= dht::dht_tracker::flag_implied_port;
}
```

PR #3162, opened 2018-07-11
Windows async read EOF handling

ajax16384:
Windows "preadv" emulation uses overlapped ReadFile API.
This async read API may throw ERROR_HANDLE_EOF (https://docs.microsoft.com/en-us/windows/desktop/fileio/testing-for-the-end-of-a-file).
There are several issues with this error code handling:
- if "preadv" has multiple input bufs and only one of them reaches EOF then all data will be discarded and windows sepcific system_error will be raised
- *nix "preadv" and old windows "readv" does not generate error upon reaching EOF (they simply returns 0 while reading after EOF)

PR should fix these issues.

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -212,7 +212,7 @@ namespace {
 				{
 					num_waits = i;
 					break;
-				} 
+				}
 				else if (last_error != ERROR_IO_PENDING
 #ifdef ERROR_CANT_WAIT
 					&& last_error != ERROR_CANT_WAIT
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -201,42 +201,60 @@ namespace {
 		}
 
 		int ret = 0;
+		int num_waits = num_bufs;
 		for (int i = 0; i < num_bufs; ++i)
 		{
 			DWORD num_read;
-			if (ReadFile(fd, bufs[i].iov_base, DWORD(bufs[i].iov_len), &num_read, &ol[i]) == FALSE
-				&& GetLastError() != ERROR_IO_PENDING
-#ifdef ERROR_CANT_WAIT
-				&& GetLastError() != ERROR_CANT_WAIT
-#endif
-				)
+			if (ReadFile(fd, bufs[i].iov_base, DWORD(bufs[i].iov_len), &num_read, &ol[i]) == FALSE)
 			{
-				ret = -1;
-				goto done;
+				DWORD last_error = GetLastError();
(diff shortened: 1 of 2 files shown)


src/file.cpp
```diff
@@ -201,42 +201,60 @@ namespace {
...
+		}
+
+		if (wait_for_multiple_objects(num_waits, h.data()) == WAIT_FAILED)
 		{
 			ret = -1;
 			goto done;
 		}
 
-		for (auto& o : ol)
+		for (int i = 0; i < num_waits; ++i)
```
arvidn: what would you think about making this a range-for loop like this?

`for (auto& o : span<OVERLAPPED>(ol).first(num_waits))`

src/file.cpp
```diff
@@ -201,42 +201,60 @@ namespace {
...
 		int ret = 0;
+		int num_waits = num_bufs;
 		for (int i = 0; i < num_bufs; ++i)
 		{
 			DWORD num_read;
-			if (ReadFile(fd, bufs[i].iov_base, DWORD(bufs[i].iov_len), &num_read, &ol[i]) == FALSE
-				&& GetLastError() != ERROR_IO_PENDING
+			if (ReadFile(fd, bufs[i].iov_base, DWORD(bufs[i].iov_len), &num_read, &ol[i]) == FALSE)
+			{
+				DWORD last_error = GetLastError();
```
arvidn: could you make `last_error` `const` please?

arvidn:
lgtm! would you mind adding a test or two demonstrating this case as well?

arvidn:
also, I imagine this fix really should go into `RC_1_1`

src/file.cpp
ajax16384: ok

src/file.cpp
ajax16384: ok

ajax16384:
add test of reading into two buffers beyong EOF

ajax16384:
will backport to RC_1_1 after merging it to master

test/test_file.cpp
```diff
@@ -1,4 +1,4 @@
-/*
+ /*
```
arvidn: would you mind removing this space?

ajax16384: ok

arvidn:
I will land this after the merge from RC_1_1 (which is in CI right now)

ajax16384:
https://github.com/arvidn/libtorrent/pull/3167

PR #3163, opened 2018-07-11
Fix unserialization of "unfinished" resume field

ajax16384:
It seems a34ce0278e0ce4c3d62a632425add5b7fcdf5551 changed how "unfinished" pieces blocks bitmasks stored, but forgot to fix how to read these bitmasks.
One of negative effects due this error was unexpected income_piece->verify_piece->do_hash->read ->file_error_alert. Because after torrent resume some piece blocks were falsely marked as finished.

### DIFF
--- a/src/read_resume_data.cpp
+++ b/src/read_resume_data.cpp
@@ -338,7 +338,15 @@ namespace {
 				bdecode_node const bitmask = e.dict_find_string("bitmask");
 				if (!bitmask || bitmask.string_length() == 0) continue;
 				bitfield& bf = ret.unfinished_pieces[piece];
-				bf.assign(bitmask.string_ptr(), bitmask.string_length());
+				char const* bitmask_str = bitmask.string_ptr();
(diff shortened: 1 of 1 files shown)


arvidn:
I don't think the format change was deliberate! The documentation still states the bitmask field is binary. I think it would be better to fix the writing of metadata to be compatible with `RC_1_1` (and itself) rather than changing the reader

arvidn:
ideally there would be a test to make sure an `add_torrent_params` object can round-trip through `write_resume_data()` and `read_resume_data()` too.

arvidn:
I submitted a pull request against your patch, to add some tests. https://github.com/ajax16384/libtorrent/pull/1

arvidn:
@ajax I think this is a better fix. what do you think? https://github.com/arvidn/libtorrent/pull/3165

ajax16384:
yes, it's looks better
so my test PR should be closed

PR #3169, opened 2018-07-12
simplify preadv and pwritev emulation functions

arvidn:


arvidn:
@ajax16384 does this look OK?

ajax16384:
yes, with scope exit routine looks much better

PR #3172, opened 2018-07-14
set port in handshake based on source address

ssiloti:


src/bt_peer_connection.cpp
```diff
@@ -2084,9 +2084,14 @@ namespace {
 		entry handshake;
 		entry::dictionary_type& m = handshake["m"].dict();
 
+		std::shared_ptr<torrent> t = associated_torrent().lock();
+		TORRENT_ASSERT(t);
+
 		// if we're using a proxy, our listen port won't be useful
 		// anyway.
-		auto const port = m_ses.listen_port();
+		auto const port = m_ses.listen_port(
```
arvidn: do you think it makes sense to gate the whole lookup logic behind `is_outgoing()` here? something like:

```
if (is_outgoing())
{
   auto const port = m_ses.listen_port(...);
   if (port == 0) handshake["p"] = port;
}
```

PR #3173, opened 2018-07-15
make the DHT announce flags a strong type

arvidn:


src/kademlia/dht_tracker.cpp
```diff
@@ -321,10 +321,11 @@ namespace libtorrent { namespace dht {
 		, std::function<void(std::vector<tcp::endpoint> const&)> f)
 	{
 		for (auto& n : m_nodes)
-			n.second.dht.get_peers(ih, f, {}, false);
+			n.second.dht.get_peers(ih, f, {}, {});
```
aldenml: do you think it makes sense to add `::none` to `flags::bitfield_flag`at some point?

arvidn: I kind of like `{}`

arvidn: ( https://www.youtube.com/watch?v=NGrnKr9rSz4 )

aldenml: anja, :), looking at it

PR #3174, opened 2018-07-15
minor code cleanup, more use of const

aldenml:


src/session_impl.cpp
```diff
@@ -880,13 +878,8 @@ namespace aux {
 		session_log(" aborting all connections (%d)", int(m_connections.size()));
 #endif
 		// abort all connections
-		for (connection_map::iterator i = m_connections.begin();
-			i != m_connections.end();)
-		{
-			peer_connection* p = (*i).get();
-			++i;
+		for (auto& p : m_connections)
 			p->disconnect(errors::stopping_torrent, operation_t::bittorrent);
```
arvidn: this is problematic. the call to `disconnect()` may erase an item from `m_connections`. That's why `i` is incremented before the call.

arvidn: in fact, there should probably be a similar assert for `session_impl` as there is in `torrent` for iterating the peer connections`

src/session_impl.cpp
```diff
@@ -276,8 +276,8 @@ namespace aux {
 	{
 		// set the default peer_class_filter to use the local peer class
 		// for peers on local networks
-		std::uint32_t lfilter = 1 << static_cast<std::uint32_t>(m_local_peer_class);
-		std::uint32_t gfilter = 1 << static_cast<std::uint32_t>(m_global_class);
+		std::uint32_t lfilter = 1u << static_cast<std::uint32_t>(m_local_peer_class);
+		std::uint32_t gfilter = 1u << static_cast<std::uint32_t>(m_global_class);
```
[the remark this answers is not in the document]
aldenml: my CLion linter, but if you don't like it, I can revert, actually, I was about to ask for your preference

src/session_impl.cpp
aldenml: I thought that with the latest changes, all actual removal of the map are done deferred, am I wrong?

arvidn:
looks good, except I'm not sure about the literals needing to be explicitly made unsigned

src/session_impl.cpp
arvidn: oh, yeah, you're right

aldenml:
hi @arvidn, can you take a look at the errors in `test_lsd`: https://travis-ci.org/arvidn/libtorrent/jobs/404249975#L2116

aldenml:
if that's the case, what _should_ it be, deferred or not?

arvidn:
perhaps it's this case that causes the immediate erase: https://github.com/arvidn/libtorrent/blob/master/src/peer_connection.cpp#L4405

aldenml:
you are absolutely right, reverting the loop change

PR #3178, opened 2018-07-17
run clang tidy on travis

arvidn:


arvidn:
I'm not excited about how `clang-tidy` has compiler errors (which I believe is caused by a very old version of libc++), and how this relies on `clang-tidy` not considering compile errors as failures

.travis.yml
```diff
@@ -5,8 +5,18 @@ matrix:
...
+      env: variant=release sonar_scan=1 toolset=gcc
+    - env: variant=debug toolset=gcc lint=1 pylint=1 clang_tidy=1
+      addons:
+        apt:
+          sources:
+          - ubuntu-toolchain-r-test
+          packages:
+          - libc++-dev
+          - python3-pip
+          - libboost1.55-all-dev
```
aldenml: question: are these packages really needed here? (they are already included below)

arvidn: I originally thought these packages were added to the ones mentioned below, but after trying, it appears they *replace* the packages below.

PR #3179, opened 2018-07-18
Test merge

arvidn:


src/xml_parse.cpp
```diff
@@ -113,7 +113,6 @@ namespace libtorrent {
 			{
 				start += 3;
 				callback(xml_comment, {start, std::size_t(tag_name_end - start - 2)}, {});
-				tag_end = p - 2;
```
aldenml: hi @arvidn, do you mind commenting how you discovered this? it was a bug, a warning, code review? thanks

arvidn: it's not a bug really, but it's a dead-store. nothing looks at this variable after it's assigned here.
clang-tidy reported it. [here's](https://travis-ci.org/arvidn/libtorrent/builds/405227223#L2342) my work in progress.

aldenml: right, `tag_end` is local to the loop, thanks

Issue #3148, opened 2018-07-06
interface rebind events for every radvd multicast message


zeule:
libtorrent version (or branch): master

platform/architecture: linux/amd64

compiler and compiler version: GCC 8.1.0

With every radvd multicast message received, libtorrent generates `listen_succeeded_alert`. These two stacktraces repeat one after another:
```
#0 libtorrent::v1_2::listen_succeeded_alert::listen_succeeded_alert() at /home/eugene/develop/qbt/libtorrent/src/alert.cpp:1058
#1 libtorrent::v1_2::listen_succeeded_alert::listen_succeeded_alert() at /home/eugene/develop/qbt/libtorrent/src/alert.cpp:1066
#2 libtorrent::heterogeneous_queue<libtorrent::alert>::emplace_back<libtorrent::v1_2::listen_succeeded_alert, libtorrent::aux::stack_allocator&, boost::asio::ip::basic_endpoint<boost::asio::ip::tcp> const&, libtorrent::v1_2::socket_type_t const&>() at /home/eugene/develop/qbt/libtorrent/include/libtorrent/heterogeneous_queue.hpp:110
#3 libtorrent::alert_manager::emplace_alert<libtorrent::v1_2::listen_succeeded_alert, boost::asio::ip::basic_endpoint<boost::asio::ip::tcp> const&, libtorrent::v1_2::socket_type_t const&>() at /home/eugene/develop/qbt/libtorrent/include/libtorrent/alert_manager.hpp:84
#4 libtorrent::aux::session_impl::reopen_listen_sockets() at /home/eugene/develop/qbt/libtorrent/src/session_impl.cpp:1917
#5 libtorrent::aux::session_impl::reopen_network_sockets() at /home/eugene/develop/qbt/libtorrent/src/session_impl.cpp:2095
#6 libtorrent::aux::session_impl::on_ip_change() at /home/eugene/develop/qbt/libtorrent/src/session_impl.cpp:1716
#7 libtorrent::aux::session_impl::wrap<void (libtorrent::aux::session_impl::*)(boost::system::error_code const&), boost::system::error_code const&>() at /home/eugene/develop/qbt/libtorrent/src/session_impl.cpp:458
#8 libtorrent::aux::session_impl::<lambda(const error_code&)>::operator()(const libtorrent::error_code &) const() at /home/eugene/develop/qbt/libtorrent/src/session_impl.cpp:1715
[...]

PR #3191, opened 2018-07-21
only post alerts for newly opened listen sockets, and only attempt to…

arvidn:
… map ports for newly opened sockets. track has_incoming_connections per listen socket

### DIFF
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -207,6 +207,10 @@ namespace aux {
 		// the key is an id that is used to identify the
 		// client with the tracker only.
 		std::uint32_t tracker_key = 0;
+
+		// set to true when we receive an incoming connection from this listen
+		// socket
+		bool incoming_connection = false;
 	};
 
 		struct TORRENT_EXTRA_EXPORT listen_endpoint_t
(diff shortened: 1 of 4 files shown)


arvidn:
@ssiloti and @zeule does this make sense? an attempt to address https://github.com/arvidn/libtorrent/issues/3148

src/session_impl.cpp
```diff
@@ -1938,7 +1949,7 @@ namespace aux {
 
 		if (map_ports)
 		{
-			for (auto const& s : m_listen_sockets)
+			for (auto const& s : new_sockets)
```
ssiloti: Based on the documentation of `session_handle::reopen_network_sockets` I take it that the `map_ports` is intended to re-map all listen sockets, including existing ones. In that case this should loop over the existing sockets rather than the new ones. That would allow the `if (!map_ports)` condition to be removed above when mapping uPNP for new sockets, and fix a bug I just noticed which causes NATPMP to be mapped twice for new sockets.

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -207,6 +207,10 @@ namespace aux {
 		// the key is an id that is used to identify the
 		// client with the tracker only.
 		std::uint32_t tracker_key = 0;
+
+		// set ot true when we receive an incoming connection from this listen
```
zeule: typo (ot → to)

arvidn:
@zeule do you think you'd have a chance to test this?

zeule:
@arvidn yes, I'll do that today.

zeule:
@arvidn, thank you, seems to be working fine.

PR #3199, opened 2018-07-26
fix doozer script

arvidn:


arvidn:
@ssiloti I think this may be a possible issue with the new PCP support. In the doozer sandboxes, it appears IPv6 is disabled, which prevents NAT-PMP from firing (because PCP fails in an unexpected way I think).

https://doozer.io/arvidn/libtorrent/buildlog/797

arvidn:
actually, it looks more like the default route in the sandbox is the problem:

```
30: [portmap_log] NAT-PMP: failed to find default route for 172.0.0.1: Success
29: [portmap_log] NAT-PMP: closing
```

ssiloti:
I suspect libtorrent is picking the wrong local address. Since the listen address is unspecified, `natpmp::start` will pick the first suitable address returned by `enum_net_interfaces`. We may need to enhance the heuristics for picking that address.

In any case it would be helpful to see the output of `enum_if` running on doozer.

arvidn:
```
Default gateway: 172.31.38.1
0.0.0.0           255.255.255.255   172.31.38.1                        1460   eth0
172.31.38.0       192.0.0.0         0.0.0.0                            1460   eth0
127.0.0.0         0.0.0.0           0.0.0.0                            65536  lo
127.0.0.0         255.255.255.0     0.0.0.0                            65536  lo
127.0.0.1         0.0.0.0           0.0.0.0                            65536  lo
127.255.255.255   0.0.0.0           0.0.0.0                            65536  lo
172.0.0.0         0.0.0.0           0.0.0.0                            65536  lo
172.0.0.0         255.255.255.0     0.0.0.0                            65536  lo
172.0.0.1         0.0.0.0           0.0.0.0                            65536  lo
172.31.38.0       0.0.0.0           0.0.0.0                            1460   eth0
172.31.38.2       0.0.0.0           0.0.0.0                            1460   eth0
172.31.38.3       0.0.0.0           0.0.0.0                            1460   eth0
172.255.255.255   0.0.0.0           0.0.0.0                            65536  lo
address                           netmask                                      name                flags               default gateway                   description
127.0.0.1                         255.0.0.0                                    lo                  local loopback            0.0.0.0                           �ő1� 
172.0.0.1                         255.0.0.0                                    lo                                      0.0.0.0                           �ő1� 
172.31.38.2                       255.255.255.252                              eth0                local                     172.31.38.1                       �ő1� 
```

arvidn:
I think the problem is that it picks `172.0.0.1` as the local address. I think it should have picked `172.31.38.2`. I will attempt a slight refactor of the `natpmp::start()` function. I think it's taking a few unnecessary detours right now.

src/natpmp.cpp
```diff
@@ -156,76 +156,102 @@ void natpmp::start(address local_address, std::string device)
...
-	// PCP requires reporting the source address at the application
-	// layer so the socket MUST be bound to a specific address
-	// if the caller didn't specify one then get the first suitable
-	// address from the OS
-	if (local_address.is_unspecified())
+	std::vector<ip_interface> const net = enum_net_interfaces(
+		m_socket.get_io_service(), ec);
+
+	address gateway;
+	if (!device.empty())
```
arvidn: the main difference here is that if we have specified a device name, just look up that device, use the local address of that device and then look up if there's a default route over that device.

Otherwise, look up the default route, and then find the local address for the device in the interface list (that we already queried above).

it feels more reliable than to first find (or guess) the local address and then ask to get the the default route for that address.

arvidn:
@ssiloti does this make sense?

ssiloti:
It doesn't correctly handle the case of having a specified `local_address` but no specified `device` name, which is the common case with IPv6. Looking at this I realized there is an even simpler approach which should handle all cases correctly. See #3206.

PR #3207, opened 2018-07-28
fix typo in #if tests for TORRENT_DISABLE_ENCRYPTION

ssiloti:


arvidn:
doh! this is why macros should be checked with just `#if ..` instead of `#ifdef`. I take it you grepped the whole repo for this, right?

ssiloti:
Yes, this patch was generated by a global search and replace.

PR #3221, opened 2018-08-04
using constexpr in recently added constants

aldenml:


arvidn:
does this make a difference?

aldenml:
I really don't think there is any, I just made it because I like what could be `constexpr` make it so. But I don't mind closing this PR, just let me know

PR #3225, opened 2018-08-06
fix buffer offset bug recently introduced to disk_io_thread::do_read()

arvidn:


src/disk_io_thread.cpp
```diff
@@ -1307,7 +1307,8 @@ constexpr disk_job_flags_t disk_interface::cache_hit;
 		// if this is the last piece, adjust the size of the
 		// last buffer to match up
 		iov[iov_len - 1] = iov[iov_len - 1].first(aux::numeric_cast<std::size_t>(
-				std::min(piece_size - (iov_len - 1) * default_block_size, default_block_size)));
+				std::min(piece_size - int(adjusted_offset) - (iov_len - 1)
```
arvidn: the offset was not taken into account here.

PR #3232, opened 2018-08-08
remove dead code from piece picker

arvidn:


src/piece_picker.cpp
```diff
@@ -1882,13 +1882,11 @@ namespace {
...
 		// scale clients, we would have more peers, and allow a higher
 		// threshold for the number of partials
-		// deduct pad files because they case partial pieces which are OK
 		// the second condition is to make sure we cap the number of partial
 		// _bytes_. The larger the pieces are, the fewer partial pieces we want.
 		// 2048 corresponds to 32 MiB
 		// TODO: 2 make the 2048 limit configurable
-		const int num_partials = int(m_downloads[piece_pos::piece_downloading].size())
-			- m_num_pad_files;
+		const int num_partials = int(m_downloads[piece_pos::piece_downloading].size());
```
arvidn: this feature was a bit of a hack, back when pad-files would create `download_piece`s just to mark the blocks as finished. That's not done anymore though.

PR #3240, opened 2018-08-13
dial back the connect boost a bit from recent update

arvidn:


d-komarov:
Is 80 connections too much?

arvidn:
with the default `connect_speed` of 30 / s, 80 connections will cause the regular connection logic to be disabled for 2.6 seconds per torrent. This is especially costly when starting up a new session from scratch with lots of torrents, it may take several minutes for late torrents to even get a single connection.

this change limits it to 1 second per torrent by default.

PR #3245, opened 2018-08-14
remove the IPv6 enable/disable build config

arvidn:
@aldenml does this look right?


include/libtorrent/socket_io.hpp
```diff
@@ -61,30 +61,24 @@ namespace detail {
 		std::size_t address_size(Proto p)
 		{
 			TORRENT_UNUSED(p);
```
aldenml: I guess this line could be removed `TORRENT_UNUSED(p);`

src/enum_net.cpp
```diff
@@ -546,9 +525,8 @@ int _System __libsocket_sysctl(int* mib, u_int namelen, void *oldp, size_t *oldl
 				b1[i] &= m[i];
 				b2[i] &= m[i];
 			}
-			return std::memcmp(b1.data(), b2.data(), b1.size()) == 0;
+			return b1 == b2;
```
aldenml: :+

src/natpmp.cpp
```diff
@@ -504,8 +504,6 @@ void natpmp::send_map_request(port_mapping_t const i)
 	else if (m_version == version_pcp)
 	{
 		// PCP requires the use of IPv6 addresses even for IPv4 messages
-		// reference asio's address_v6 class directly so that we can use it
-		// even if TORRENT_USE_IPV6 is false
 		using boost::asio::ip::address_v6;
```
aldenml: I assume this line can be removed `using boost::asio::ip::address_v6;`

aldenml:
LGTM!!

arvidn:
anything else I should consider before landing this?

aldenml:
none on my part, but I would like to hear from @ssiloti

src/peer_list.cpp
```diff
@@ -1043,13 +1037,11 @@ namespace libtorrent {
 		if (remote_address == address() || remote.port() == 0)
 			return nullptr;
 
-#if TORRENT_USE_IPV6
 		// don't allow link-local IPv6 addresses since they
 		// can't be used like normal addresses, they require an interface
 		// and will just cause connect() to fail with EINVAL
 		if (remote_address.is_v6() && remote_address.to_v6().is_link_local())
```
ssiloti: Not really related to these changes, but link-local addresses could be allowed as long as `remote_address.to_v6().scope_id()` is non-zero.

PR #3249, opened 2018-08-18
fix issue in self-connection detection

arvidn:
introduced with the change to generate unique peer-ids for each connection. Now, the torrent keeps track of all of our peer-ids generated for outgoing (bittorrent) connections, and we check them against incoming peers' peer-ids


arvidn:
with this change, all the previous mechanisms to detect self-connections can probably be removed (the ones looking at source and remote IPs for instance)

arvidn:
on the other hand, it seems reasonable to have short-cuts for detecting self-connections based on IP addresses, to avoid the handshake

Issue #3226, opened 2018-08-06
libtorrent 1.1.{7,8,9} + Deluge 1.3.15, connecting only to small fraction of peers over dozens and only on certain trackers


quarreldazzle:
**Please provide the following information**

libtorrent version (or branch):

- 1.1.7
- 1.1.8
- 1.1.9

platform/architecture: Debian GNU/Linux 9, x86_64

compiler and compiler version:

gcc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516

please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

This issue was originally [posted on Reddit r/seedboxes](https://www.reddit.com/r/seedboxes/comments/937hct/deluge_w_libtorrent_11x_and_ipt_potentially_other/). I am summarizing it here.

Deluge with the 1.1 branch of libtorrent (confirmed on .7, .8, and .9). 
Snatching torrents with autodl irssi on IPT and TL trackers.

After seeing the first _Announce OK_ message (not related to the _unregistered torrent_ issue in #315; workaround on that works fine), Deluge connects to only a small fraction of peers. Usually 2 to 5 over confirmed swarms of dozens (also 60+). Often, the uploader does not seem to be included in the swarm, or the uploader is not shown on Deluge because of the bug itself. Download and upload start and performance is (of course) terrible.

There are no other trackers that seem affected by the issue.

**Pausing and resuming a torrent mid-race "fixes" the issue**. From 2-5 peers, Deluge starts seeing the entire swarm. Most of which are finished already. Download goes to the max, upload is almost none as most of them are at 100% already.

A catch: the issue happens most of the times, but not all the times.

**Downgrading to libtorrent 1.0.11.0 fixes the issue completely**.

Somebody on Reddit suggested disabling UTP, but it does not solve the issue.

arvidn:
is it possible that the main reason pausing and resuming solves the problem is because of re-announcing with the tracker?

could you try just re-announcing to the tracker instead of a full pause?

Do you have access to the `connect_candidates` counter for the torrent? I would expect it to either be 0 or libtorrent to attempt more connections.

quarreldazzle:
> is it possible that the main reason pausing and resuming solves the problem is because of re-announcing with the tracker?

Pause/resume fixes the issue, update tracker does not. Unless you want me to try something else?

> Do you have access to the connect_candidates counter for the torrent? 

How do I access the counter? I have root access and deluge-console is installed, if it helps.

[...]

PR #3250, opened 2018-08-18
Option to ignore min interval

arvidn:


LordNyriox:
@arvidn:  What do you expect the use-case for this feature, to be?

arvidn:
@LordNyriox I think my post in the other thread answers that: https://github.com/arvidn/libtorrent/issues/3226#issuecomment-415342266

Issue #2663, opened 2018-01-03
update documentation for anonymous mode


Piccirello:
libtorrent version (or branch): 1.1.6

platform/architecture: n/a

compiler and compiler version: n/a

I'm attempting to better understand the libtorrent codebase and have been looking into the anonymous_mode feature. I've found various instances of `settings_pack::anonymous_mode` being used to determine user-agents, peer ids, and client versions throughout the code. The settings [reference](https://www.libtorrent.org/reference-Settings.html) also states that when anonymous mode is on:

```
* Trackers will only be used if they are using a proxy server
* The listen sockets are closed, and incoming connections will only be accepted through a SOCKS5 or I2P proxy
* NAT-PMP, UPnP, DHT and local peer discovery are all turned off
```

I haven't been able to find where any of these things happen based on the value of `settings_pack::anonymous_mode`. Could someone please point me to where UPnP, DHT, or local peer discovery are disabled; where the listen sockets are closed; or where any of these other items happen? Thank you.

arvidn:
that documentation is outdated. originally anonymous mode did all those things, but the setting was then split into ``anonymous_mode`` and ``force_proxy``.

Piccirello:
> originally anonymous mode did all those things, but the setting was then split into anonymous_mode and force_proxy.

Do you know what version of libtorrent was responsible for this change?

Piccirello:
It appears the functionality was migrated from `anonymous_mode` to `force_proxy` for the 1.0.0 release. The change was documented in `session_settings.hpp` but alas libtorrent.org hasn't been updated. Should I create an issue for updating the web documentation?

arvidn:
let's just use this ticket

mhertz:
@arvidn 

There's a discussion currently going on in qBittorrent bugtracker, so could you please kindly confirm for me that force_proxy doesn't disable DHT so we can lay this to rest thanks :) (I know the option is deprceated now and "built-in" but nonetheless)

Thanks in advance.

arvidn:
that's right; `force_proxy` doesn't disable DHT. However, if you don't have a proxy that supports UDP (only the socks5 protocol does) it will effectively be disabled, just like uTP and UDP trackers.

Not all socks5 proxy software supports UDP though.

Piccirello:
[...]

PR #3251, opened 2018-08-19
force proxy no longer disables the DHT

arvidn:
issue raised in https://github.com/arvidn/libtorrent/issues/2663

### DIFF
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -609,9 +609,7 @@ namespace libtorrent
 			// proxy_type and proxy_hostname settings. The listen sockets are
 			// closed, and incoming connections will only be accepted through a
 			// SOCKS5 or I2P proxy (if a peer proxy is set up and is run on the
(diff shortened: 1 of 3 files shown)


include/libtorrent/settings_pack.hpp
```diff
@@ -609,9 +609,7 @@ namespace libtorrent
 			// proxy_type and proxy_hostname settings. The listen sockets are
 			// closed, and incoming connections will only be accepted through a
 			// SOCKS5 or I2P proxy (if a peer proxy is set up and is run on the
-			// same machine as the tracker proxy). This setting also disabled peer
-			// country lookups, since those are done via DNS lookups that aren't
-			// supported by proxies.
+			// same machine as the tracker proxy).
```
Piccirello: Are peer country lookups happening regardless of this setting?

arvidn: they are deprecated, disabled by default, not documented and has been removed for the next major version

PR #3253, opened 2018-08-19
Pad blocks performance

arvidn:


include/libtorrent/torrent.hpp
```diff
@@ -113,6 +113,8 @@ namespace libtorrent {
 		, max
 	};
 
+	std::int64_t TORRENT_EXTRA_EXPORT calc_bytes(file_storage const& fs, piece_count const& pc);
```
aldenml: just for code consistency, maybe you want to change this to `TORRENT_EXTRA_EXPORT std::int64_t` instead

Issue #3246, opened 2018-08-16
Freeze when deleting or pausing a torrent using the cache


Rootax:
**Please provide the following information**

**libtorrent version (or branch):**

1.1.X

**platform/architecture:**

Windows 10 x64 (1803)

**compiler and compiler version:**

The version used bit Qbittorent ?

Hello. I already opened a issue on the qbittorent git, but it seems that no one cares, or I believe the cause is libtorrent since all is fine with 1.0 branch.

So... 

qBittorrent version and Operating System:
Any version with libtorrent 1.1.x / Windows 10 x64

What is the problem:
When I delete or pause a torrent which use the cache, the application is not responding, and is using a whole cpu thread. The more cache is used, the more time Qbittorent is doing this. In the end, it doesn't crash and the app is responsive again, but during this time, the download and upload stops. It's only doing that with libtorrent 1.1 builds. Whith 1.0 it's ok (but I got much worse upload speeds...)

What is the expected behavior:
Not freezing like that...

Steps to reproduce:
Enable cache (size doesn't matter), download a torrent, then when it's over, deleting it (with or without deleting the file too, doesn't make a difference). Qbittorent will "freeze", using a lot of cpu, and take a long time deleting the files/torrent.

Extra info(if any):
Sorry for my english.

It seems that, when I shutdown qbittorent, It sometimes doing the same thing. Like, I exit it, but in the task manager, I see the process still running, using a lot of cpu, and slowly using less and less memory, until it really shutdown.

And no errors are shown in the log file.

My hardware 
CPU : Intel I3 4010u, 16gb of ram, a Crucial SSD (Sata)
-No error on disk
-Pagefile is enabled; 32gb
-During those cpu spike, the drive does nothing. The only thing working is the cpu (1 thread), and the memory used by qbitorrent cache slowly goes down. When all the useless data in cache is "purged", then qbittorent respond again. And it's not data waiting to be writing on disk, since my cache is mostly a read cache (I seed a lot).

All is ok with libtorrent 1.0, so I guess it's a libtorrent 1.1x bug. I tried various cache settings, but the result is always the same.

For clarity, I showed the problem in this youtube video : https://youtu.be/xeucCbjILP8 By the end of it, you see qbittorrent not responding. The video cut before it's responsive again, but it's ok when the cache is "flushed". Before that, a thread fully used, and qbittorrent not responding or very laggy.

[...]

PR #3255, opened 2018-08-19
improve performance of flush_cache in disk_io_thread

arvidn:
by reducing piece lookups

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* improve performance of freeing disk cache for a specific torrent
 	* fix redundant flushes of partfile metadata
 	* add option to ignore min-interval from trackers on force-reannounce
 	* raise default setting for active_limit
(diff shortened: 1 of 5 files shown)


arvidn:
@ssiloti do you have any other ideas about https://github.com/arvidn/libtorrent/issues/3246 and does this patch look reasonable?

arvidn:
I suspect this is a legitimate issue: https://travis-ci.org/arvidn/libtorrent/jobs/427861699#L1471

arvidn:
I can't get this to work. Some invariants that I expect don't seem to be maintained in the cache. I would much rather work on the patch where the cache is removed

ssiloti:
I can understand that. Just reviewing this patch made me look forward to the day when all this code is gone.

arvidn:
closing again. if someone else feels strongly that this should be improved, feel free to revive and take over the branch.

PR #3256, opened 2018-08-19
fall back to copy+remove if rename_file fails

arvidn:


arvidn:
@oxguy3 any chance you could try this patch out?

oxguy3:
@arvidn Hey, thanks for the patch! Been super busy with the start of the semester but I will try to give this a try ASAP.

PR #3273, opened 2018-08-31
WIP: experiment with always want tick for all torrents

arvidn:


LordNyriox:
@arvidn:  Near as I can tell, this PR simply changes the fallback, for a setting called `torrent::want_tick()`, to `true` instead of `false`.  

Perhaps you can explain what `want_tick` is supposed to do?

arvidn:
as an optimization, libtorrent keeps various lists of pointers to torrents with certain properties. For instance, torrents that want to have it's `tick()` function called once per second (for performing maintenance tasks), or torrents that want to connect to more peers. This is used to not have to iterate over all torrents when doing those things, so idle torrents become cheap.

This patch effectively disable the optimization to not call `tick()` on some torrents, by making all torrents always wanting to have that called.

The reason I suspect this may make a difference is because it's in this `tick()` function the logic lives to determine whether a torrent is active or inactive.

LordNyriox:
@arvidn:  Was the behavior of this "optimization" different in Libtorrent 1.0.x?  

Because the report suggests that the problems started with qBittorrent 4.0—which happens to also be the first qBT release using Libtorrent 1.1.x.

LordNyriox:
@arvidn:  Was the behavior of this "optimization" different in Libtorrent 1.0.x?  

Because [this report](https://github.com/qbittorrent/qBittorrent/issues/7754#issuecomment-351638589) suggests that the problem started with qBittorrent 3.4—which happens to also be the first qBT release using Libtorrent 1.1.x.

LordNyriox:
@arvidn:  Ah.  I suppose that explains why this patch might work.  

Perhaps you can poke @sledgehammer999 to make a testing build of qBittorrent, using this PR?

PR #3276, opened 2018-09-04
fix some warnings from sonar-cube

arvidn:


include/libtorrent/stat_cache.hpp
```diff
@@ -84,7 +84,7 @@ namespace libtorrent {
 
 		struct stat_cache_t
 		{
-			stat_cache_t(std::int64_t s): file_size(s) {} // NOLINT
+			explicit stat_cache_t(std::int64_t s): file_size(s) {} // NOLINT
```
aldenml: no more need of `// NOLINT` I guess

src/torrent.cpp
```diff
@@ -4971,10 +4971,11 @@ bool is_downloading_state(int const st)
 			position += size;
 
 			// pad files always have priority 0
-			download_priority_t const file_prio
-				= fs.pad_file_at(i) ? dont_download
-				: i >= file_prios.end_index() ? default_priority
-				: file_prios[i];
+			download_priority_t const file_prio = [&](){
```
aldenml: I wonder what happens if I send a PR with a refactor like this :D

arvidn: heh, I'm not super happy about this. I might revert it.

PR #3279, opened 2018-09-04
improve type-safety of the severity parameter to peer_connection::disconnect()

arvidn:


src/peer_connection.cpp
```diff
@@ -4155,21 +4160,10 @@ namespace libtorrent {
...
-			case 1:
-				peer_log(peer_log_alert::info, "CONNECTION_FAILED", "op: %d error: %s"
-					, static_cast<int>(op), ec.message().c_str());
-				break;
-			case 2:
-				peer_log(peer_log_alert::info, "PEER_ERROR" ,"op: %d error: %s"
-					, static_cast<int>(op), ec.message().c_str());
-				break;
-			}
+			aux::array<char const*, 3, disconnect_severity_t> str{{{
```
aldenml: `static const` is appropriate here or not?

PR #3282, opened 2018-09-07
Fire checked_alert before finished_alert

mj-p:
QBittorrent makes use of both those alerts to schedule a recheck on download completion.
But in case you want to seed a new file QBittorrent ends in a loop because a finished_alert
gets fired before checked_alert which schedules another recheck.


thalieht:
qBt uses the RC_1_1 branch if it matters to you.

PR #3284, opened 2018-09-07
Fire checked_alert before finished_alert

mj-p:
QBittorrent makes use of both those alerts to schedule a recheck on download completion.
But in case you want to seed a new file QBittorrent ends in a loop because a finished_alert
gets fired before checked_alert which schedules another recheck.

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -8858,6 +8858,14 @@ namespace {
 			m_ses.trigger_auto_manage();
 		}
 
+		INVARIANT_CHECK;
+
+		if (m_ses.alerts().should_post<torrent_checked_alert>())
+		{
+			m_ses.alerts().emplace_alert<torrent_checked_alert>(
+				get_handle());
+		}
+		
 		if (!is_seed())
 		{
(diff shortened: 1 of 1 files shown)


arvidn:
I’m not opposed to this logic per-se, but it seems a bit suspicious to solve that problem by changing the behaviour of libtorrent. What if some other client relies on the current order, you might break someone else.

It also seems like qbt should be a bit more robust

LordNyriox:
@arvidn?

arvidn:
how about only trigger the "recheck-on-completion" logic if the torrent is not already in a "checking"-state. and perhaps only trigger it if the number of downloaded bytes in the current session is greater than 0.

arvidn:
my understanding is that this issue has been addressed with a different approach, is that right?

LordNyriox:
@arvidn:  

>my understanding is that this issue has been addressed with a different approach, is that right? 

The qBT PR relating to this issue has not been merged yet.  

See [<https://github.com/qbittorrent/qBittorrent/pull/9481>], for more info on the proposed solution.

PR #3287, opened 2018-09-09
add new resumedata_accepted_alert

arvidn:


arvidn:
the part I'm not all that confident about with this is whether the `fastresume_accepted_alert` is posted at all the right times.

glassez:
>the part I'm not all that confident about with this is whether the fastresume_accepted_alert is posted at all the right times.

Shouldn't be only one place where resume data is fully checked?
You can reject it at any point when error is detected. But it can be accepted only in the end of the checking process if there is no error occurred, isn't it?

arvidn:
@LordNyriox does this patch solve the problems in qbt it was meant to solve?

LordNyriox:
@arvidn:  I do not know what is sufficient to @glassez's purposes.  

And I do not know C++ well enough to form any kind of meaningful opinion, based on the code alone.

glassez:
>does this patch solve the problems in qbt it was meant to solve?

This patch itself can't solve the qbt. It's needed as a base for qbt patch to solve meantioned problem.
I thought it's not so bad idea to have this alert anyway so I'm waiting for it to be merged to create following qbt changes.
I can test it before if you want.

glassez:
@arvidn, along the way, I'm thinking of a backwards compatible way that doesn't require these changes. I have one question for you.
Is there any guarantee that checking_resume_data can't be non-initial torrent state (it cannot go into this state after any other)?
Or is there any other reliable "torrent state" order?

arvidn:
sorry, I forgot to respond. I believe that a "force recheck" may go via the checking_resume_data state also.

arvidn:
I'll close this, as I'm a bit worried about how this would be merged into master, where resume data is handled a bit differently.

PR #3291, opened 2018-09-11
fix redundant flushes of partfile metadata

arvidn:


aldenml:
is this an issue in `master`?

arvidn:
I would expect so

PR #3294, opened 2018-09-15
remove dead code udp_tracker_connection::pick_target_endpoint

ssiloti:
Endpoints are filtered by address family in name_lookup() so this
function will always return the first endpoint.

### DIFF
--- a/src/udp_tracker_connection.cpp
+++ b/src/udp_tracker_connection.cpp
@@ -154,7 +154,7 @@ namespace libtorrent {
 #endif
 
 		// pick another target endpoint and try again
-		m_target = pick_target_endpoint();
+		m_target = udp::endpoint(m_endpoints.front().address(), m_endpoints.front().port());
 
 #ifndef TORRENT_DISABLE_LOGGING
 		if (cb && cb->should_log())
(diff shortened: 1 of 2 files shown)


src/udp_tracker_connection.cpp
```diff
@@ -154,7 +154,7 @@ namespace libtorrent {
 #endif
 
 		// pick another target endpoint and try again
-		m_target = pick_target_endpoint();
+		m_target = udp::endpoint(m_endpoints.front().address(), m_endpoints.front().port());
```
arvidn: i seem to recall there is a utility function already for converting between udp and tcp endpoints.

src/udp_tracker_connection.cpp
```diff
@@ -252,49 +252,11 @@ namespace libtorrent {
 			return;
 		}
 
-		m_target = pick_target_endpoint();
```
arvidn: `pick_target_endpoint()` seems to support the case `m_endpoints.empty()` as well. are you sure that can't happen?
there should probably be an assert for that.

ssiloti: On the contrary, `pick_target_endpoint` assumes `!m_endpoints.empty()` by unconditionally dereferencing `m_endpoints.begin()` on its second line.

There is a check just above this line which returns if `m_endpoints.empty()` and a similar one above other other callsite so neither line can be reached if `m_endpoints.empty()`.

src/udp_tracker_connection.cpp
ssiloti: I cannot find such a function.

PR #3295, opened 2018-09-17
add make_tcp() and make_udp() to convert endpoint types

arvidn:


aldenml:
you might want to change/use/refactor here too: https://github.com/arvidn/libtorrent/blob/28c94b54edfda24c5ad70c2ae06248be3e642753/src/session_impl.cpp#L2108

include/libtorrent/socket.hpp
```diff
@@ -93,6 +93,12 @@ namespace libtorrent {
 	using null_buffers = boost::asio::null_buffers;
 #endif
 
+	inline udp::endpoint make_udp(tcp::endpoint const ep)
```
aldenml: do you think it's better to take `const&` instead?

arvidn: I don't really think it matters. the function is making a copy anyway, and it will most likely be inlined and probably disappear entirely

PR #3301, opened 2018-09-22
back-port the patch to split up the progress_notification

arvidn:
into three new categories, file-, piece- and block progress

### DIFF
--- a/examples/client_test.cpp
+++ b/examples/client_test.cpp
@@ -1539,6 +1539,8 @@ int main(int argc, char* argv[])
 	settings.set_str(settings_pack::user_agent, "client_test/" LIBTORRENT_VERSION);
 	settings.set_int(settings_pack::alert_mask, alert::all_categories
 		& ~(alert::dht_notification
+		+ alert::piece_progress_notification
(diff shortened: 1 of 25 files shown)


examples/client_test.cpp
```diff
@@ -1539,6 +1539,8 @@ int main(int argc, char* argv[])
 	settings.set_str(settings_pack::user_agent, "client_test/" LIBTORRENT_VERSION);
 	settings.set_int(settings_pack::alert_mask, alert::all_categories
 		& ~(alert::dht_notification
+		+ alert::piece_progress_notification
+		+ alert::block_progress_notification
```
glassez: Shouldn't file_progress_notification be here too?

arvidn: it should in order to have the same behavior as before. I figured that file progress alerts aren't as frequent and that it may be ok to print them

arvidn: however, constructing masks by *excluding* flags is risky for when I add new flags with new alerts. in master I've changed this to specify the flags to *include*

PR #3309, opened 2018-09-24
cmake: add option to build the C bindings library

zeule:
I somehow overlooked this library when did the work with the CMake build scripts, so here it is.


arvidn:
My understanding is that the C bindings don't build or at least don't really work. If they do, it would seem reasonable to build them on travis (and maybe appveyor too). However, I'm not really ready to support a C binding, so maybe it's better to not build it on CI.

PR #3316, opened 2018-10-01
factor out the documentation header into separate rst file

arvidn:


LordNyriox:
@arvidn:  Can you also backport this change for the `1.1.10` release? 

This seems like an effective way of reducing the number of change-files for each "bump version" commit.

arvidn:
yeah, the intention of this change is to make the merges simpler when bumping the version.

that was a deliberate change for that one document. the ones that use the include still render correctly  locally, not on github though.

LordNyriox:
@arvidn:  

>the intention of this change is to make the merges simpler when bumping the version. 

Which is itself an effective argument for backporting this documentation refactor. 

Unless Libtorrent 1.2 is ready for release branching?

LordNyriox:
@arvidn:  

>the intention of this change is to make the merges simpler when bumping the version. 

Which is, itself, an effective argument for backporting this documentation refactor. 

Please do so?

arvidn:
I'm hoping 1.1.10 will be the last release out of `RC_1_1` and that master will be the next release, as 1.2

PR #3339, opened 2018-10-08
Avoid having CMakeList.txt alter the C++ standard version

arvidn:
a bunch of cmake fixes. specifically to enable building with cmake out of a tarball generated with automake.

@zeule would you mind taking a look?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -543,12 +543,10 @@ set_target_properties(torrent-rasterbar
 		SOVERSION ${SOVERSION}
 )
 
-if (cxx_std_14 IN_LIST CMAKE_CXX_COMPILE_FEATURES)
-	target_compile_features(torrent-rasterbar PUBLIC cxx_std_14)
-	message(STATUS "Building in C++14 mode")
-else()
-	target_compile_features(torrent-rasterbar PUBLIC cxx_std_11)
-	message(STATUS "Building in C++11 mode")
+# make the CXX_STANDARD property public to ensure it makes it into the pkg-config file
+get_target_property(_std torrent-rasterbar CXX_STANDARD)
+if (${_std})
+	target_compile_features(torrent-rasterbar PUBLIC cxx_std_${_std})
 endif()
 
 target_include_directories(torrent-rasterbar PUBLIC


CMakeLists.txt
```diff
@@ -4,12 +4,18 @@ project(libtorrent
...
 
 list(APPEND CMAKE_MODULE_PATH ${libtorrent_SOURCE_DIR}/cmake/Modules)
 include(FeatureSummary)
 include(GNUInstallDirs)
 include(GeneratePkgConfig)
+include(CPack)
+
+add_custom_target(dist COMMAND ${CMAKE_MAKE_PROGRAM} package_source)
+
+set(CMAKE_CXX_EXTENSIONS OFF)
```
arvidn: this fixes the `-std=gnu++<version>` to instead be `-std=c++<version>`

CMakeLists.txt
```diff
@@ -543,14 +549,6 @@ set_target_properties(torrent-rasterbar
 		SOVERSION ${SOVERSION}
 )
 
-if (cxx_std_14 IN_LIST CMAKE_CXX_COMPILE_FEATURES)
-	target_compile_features(torrent-rasterbar PUBLIC cxx_std_14)
-	message(STATUS "Building in C++14 mode")
-else()
-	target_compile_features(torrent-rasterbar PUBLIC cxx_std_11)
-	message(STATUS "Building in C++11 mode")
-endif()
```
arvidn: it's not safe to set the c++ version just for a single target. The c++ version must be set at the top level, not at this library.

CMakeLists.txt
```diff
@@ -712,8 +712,8 @@ else()
 endif()
 
 # Boost
-find_public_dependency(Boost REQUIRED COMPONENTS system)
-target_include_directories(torrent-rasterbar PUBLIC ${Boost_INCLUDE_DIRS})
+find_package(Boost REQUIRED COMPONENTS system)
+target_include_directories(torrent-rasterbar PUBLIC ${Boost_INCLUDE_DIR})
```
arvidn: I couldn't get this to work right. For some reason both `find_public_dependency` and `find_package` would find the system installed boost library but headers from some boost repository somewhere in my home directory. setting clearing `BOOST_ROOT` did not help. @zeule do you have any ideas of how to fix this? It results in a serious link error that's hard to diagnose.

test/CMakeLists.txt
```diff
@@ -16,12 +16,10 @@ add_library(test_common main.cpp test.cpp setup_transfer.cpp dht_server.cpp udp_
 	peer_server.cpp web_seed_suite.cpp swarm_suite.cpp test_utils.cpp make_torrent.cpp settings.cpp
 )
 target_link_libraries(test_common PUBLIC torrent-rasterbar)
-target_compile_features(test_common PUBLIC cxx_std_11)
 
 foreach(TARGET_SRC ${tests})
 	get_filename_component(TARGET ${TARGET_SRC} NAME_WE)
 	add_executable(${TARGET} ${TARGET_SRC})
-	target_compile_features(${TARGET} PRIVATE cxx_std_11)
```
arvidn: both of these cause link incompatibilities

CMakeLists.txt
zeule: It was here to propagate this value into the exported targets and the pkg-config file. As a `PUBLIC` feature it will be picked up by all dependent targets. I don't understand therefore why it is unsafe.

CMakeLists.txt
```diff
@@ -581,8 +580,9 @@ if (WIN32)
 	)
 
 	target_compile_definitions(torrent-rasterbar
-		PUBLIC
+		PRIVATE
```
zeule: I know little about Windows includes, but it seems strange to me that these two control definitions have different visibility.

CMakeLists.txt
```diff
@@ -712,8 +712,8 @@ else()
 endif()
 
 # Boost
-find_public_dependency(Boost REQUIRED COMPONENTS system)
-target_include_directories(torrent-rasterbar PUBLIC ${Boost_INCLUDE_DIRS})
+find_package(Boost REQUIRED COMPONENTS system)
```
zeule: `find_public_dependency` ensures that libtorrent clients will look for Boost too. Can't understand this change.

CMakeLists.txt
zeule: I suppose you tried cleaning the cmake cache in the build dir? Where does the `Boost_DIR` variable point to?

zeule: Forgot to ask: which Boost installation you want the build to use?

CMakeLists.txt
arvidn: This define determines which version of the windows runtime to compile against. If this is public, it forces all clients of libtorrent to also target vista, which I don't think is reasonable. IIUC, private means it only affects the libtorrent build itself

CMakeLists.txt
arvidn: I see. I thought `PUBLIC` did that. I will change this back. However, the fundamental problem of it not working right is still there..

CMakeLists.txt
arvidn: either is fine. The problem is that the library and the include path point to different ones, and then cause link errors

zeule: If you set `Boost_DIR`, the `FindBoost.cmake` module should not go outside of it, IIRC. So, could you try to set it and run cmake again? You have to clean cmake cache (or use a new build directory, obviously simpler for testing).

CMakeLists.txt
arvidn: picking which C++ version to used *just* based on what the compiler supports is really dangerous and fails in lots of ways, especially when linking against pre-built system libraries. Even if this property is propagated up to dependents, will the build fail if two dependencies propagate conflicting properties up?

my understanding of the cmake model is that these kinds of ABI altering options should always only be propagated down the dependency tree, and only from the top level. (and libtorrent, being a library, won't ever really be the top level)

arvidn: I didn't realize this affected the `pkg-cfg` though, but is it important that it does? it seems suspicious to put these options in there too (for the same reason)

zeule: How can we propagate that down if libtorrent is built already and we are compiling an application using it? libtorrent ABI depends on the C++ version, right? If so, we have to set something for clients, otherwise the library has to be rebuild with compiler upgrade/downgrade which changes the default C++ standard.

> …will the build fail if two dependencies propagate conflicting properties up?

Errr… It was you who coded it in that way to fail linking in such cases. Why do you ask?

zeule: Or am I wrong here and it fails in run time?

arvidn: you're right, for ``pkg-cfg`` it can't be propagated down, and you may as well propagate it up.
I don't think I've really contributed much to the cmake file. My understanding is that detecting conflicting properties being propagated up is a limitation of cmake, that it can't do that, and that's why the convention is to only propagate down, from top level.

does the package config file look at the build options for the `torrent-rasterbar` target to build the file? if so, this may still be correct. By removing this, the build options are all derived from the top level `cmake` invocation of `CMAKE_CXX_VERSION`, right?

CMakeLists.txt
arvidn: on second thought I think this should be controlled from top-level as well

CMakeLists.txt
zeule: But `WIN32_LEAN_AND_MEAN` casts restrictions on the API too. Why these two macros differ in scopes? My point is purely about their conceptual similarity (both cancel some declaration out from Windows include files).

arvidn: that's true. However, `WIN32_LEAN_AND_MEAN` is (likely) required to not break *inluding* a libtorrent header. I'm not sure but I seem to recall that there are boost.asio headers that in turn include windows.h. It's also less controversial and restrictive, since (my impression at least) is that everyone building on windows has defined this for decades to get a slightly saner build environment. So it could be considered a service to clients to provide this for them :)

CMakeLists.txt
arvidn: how can I verify that the pkg-config file is correct? I don't actually see the `libtorrent-rasterbar.pc` be updated at all by cmake. Is it put in some other place?

zeule: > how can I verify that the pkg-config file is correct?

After the generation stage (i.e. just before the build starts) it can be found inside `torrent-rasterbar-pkgconfig` subdir in the build directory. 

>  I don't actually see the libtorrent-rasterbar.pc be updated at all by cmake. Is it put in some other place?

Do you mean the install target does not actually install the pkg-config file into the proper location? It should do that (qlist lists contents of package on Gentoo):
```
$ qlist libtorrent-rasterbar | grep pkgconfig
/usr/lib64/pkgconfig/libtorrent-rasterbar.pc
```

arvidn: I see, the default `CXX_STANDARD` is private. I agree that whatever `CXX_STANDARD` resolves to should be put in the pkg-config file. However, pkg-config will most likely fail in the case of a client linking against two separate libraries, both which have conflicting c++ versions exported. But I don't think there's anything that can be done about that in pkg-config. However, in cmake things can be a bit better by never setting this explicitly, but only rely on the target pulling in libtorrent passing down its settings.

do you think it would work to do this unconditionally?

`target_compile_features(torrent-rasterbar PUBLIC cxx_std_${CXX_STANDARD})`

arvidn: > Do you mean the install target does not actually install the pkg-config file into the proper location? It should do that (qlist lists contents of a package on Gentoo):

Oh, I see. The automake build scripts puts the `.pc` file in the root directory (based on the `libtorrent-rasterbar.pc.in` also in the root directory). I thought cmake would write it to the same place.

CMakeLists.txt
zeule: Thanks for the explanations!

CMakeLists.txt
zeule: > I thought cmake would write it to the same place.

The file is accompanied by a number of others (intermidiates), as you see. I decided to put them together into a subdir, so the root build dir looks cleaner.

zeule: > but I would not use the variable directly

Psychological self-analysis showed :) that I expect the property to be affected by other ways, and I think to maintain consistency it is safer to read the final property value instead of relying on one of the way to initiate it.

arvidn: in the first version of this patch I set `CXX_STANDARD_REQUIRED`, but I figured it's better to let it be set at the top level. you're suggesting it should be exported in the `pkg-config` file as well?

zeule: > you're suggesting it should be exported in the pkg-config file as well?

No. As I understand documentation for the `CXX_STANDARD` property and `CMAKE_CXX_STANDARD` variable, the property can have a value, but it is not enforcing unless `CXX_STANDARD_REQUIRED` is set. From the other hand, compile  features are always enforcing.

I guess we have to test `CMAKE_CXX_COMPILE_FEATURES` before copying the value from the property into the compile feature:
```cmake
if (cxx_std_${_std} IN_LIST CMAKE_CXX_COMPILE_FEATURES)
    target_compile_features(torrent-rasterbar PUBLIC cxx_std_${_std})
endif()
```

arvidn: what does it mean for the `CXX_STANDARD` to not be enforced?
is it not reasonable to require this property to be present? Would it not be present if the top-level build command does not specify `CXX_STANDARD`? If so, it should probably still be set by whatever is the default in the compiler, no?

zeule: If compiler does not support the requested standard version and  `CXX_STANDARD_REQUIRED` is not set, CMake switches to the latest supported by the compiler. See https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html

arvidn: do you think it would be appropriate to set `CXX_STANDARD_REQUIRED` and `CXX_EXTENSIONS` in the libtorrent CMakeLists.txt? What happens if the top-level project sets conflicting values for those? will it still be guaranteed that every part of the program is built with identical flags?

arvidn: it doesn't seem right to set the `target_compile_features` based on `CMAKE_CXX_COMPILE_FEATURES`. The only thing that should be done (as far as I can tell) is to find the `cxx_std_` property in the `torrent-rasterbar` target, and make it `PUBLIC`. (and it seems really odd that cmake doesn't make it public by default, it's clearly part of the ABI).

The documentation for `CXX_STANDARD` says that it will be set on all targets, so it seems like a fairly safe assumption that it will be set on the `torrent-rasterbar` target. But I suppose there's a risk that the `torrent-rastebar` target has some other, conflicting, property set, in which case there are conflicting PUBLIC and PRIVATE properties. Would cmake fail in this case? (one would hope so).

But maybe it's clearer to explicitly list the properties from the target. Is there a way to do that?

CMakeLists.txt
```diff
@@ -543,12 +543,10 @@ set_target_properties(torrent-rasterbar
 		SOVERSION ${SOVERSION}
 )
 
-if (cxx_std_14 IN_LIST CMAKE_CXX_COMPILE_FEATURES)
-	target_compile_features(torrent-rasterbar PUBLIC cxx_std_14)
-	message(STATUS "Building in C++14 mode")
-else()
-	target_compile_features(torrent-rasterbar PUBLIC cxx_std_11)
-	message(STATUS "Building in C++11 mode")
+if (NOT "${CXX_STANDARD}" STREQUAL "")
```
zeule: The name of the variable is `CMAKE_CXX_STANDARD`, `CXX_STANDARD` is the name of the property.

arvidn: oh. I'm confused. I was seeing an error on appveyor (before I added the `-DCMAKE_CXX_VERSION` to the command line) where `_std` expanded into some weird "NOTFOUND" string. I thought this was a way to detect whether the property was set or not. I suppose it isn't.

It looks like the value "NOTFOUND" is well defined. So I could just check for that then, right?

zeule: For a property that can have default value you set it via variable named "CMAKE_<property name>";  CMake copies values from such variables into properties when creating a target.

zeule: > It looks like the value "NOTFOUND" is well defined. So I could just check for that then, right?

You can even write 
```cmake
if (NOT ${_std})
```
`NOTFOUND` is one of the boolean "false" values for CMake.

zeule: Pardon, `(if ${_std})`, of course.

arvidn: > For a property that can have default value you set it via variable named CMAKE_<property name>; CMake copies values from such variables into properties when creating a target.

makes sense. so my understanding then is to never set any `CMAKE_` variables inside a CMakeList.txt, but only set them at the top level `cmake` invocation, otherwise you risk different targets to have different properties.

I'll simplify the check.

CMakeLists.txt
```diff
@@ -543,12 +543,10 @@ set_target_properties(torrent-rasterbar
...
-if (cxx_std_14 IN_LIST CMAKE_CXX_COMPILE_FEATURES)
-	target_compile_features(torrent-rasterbar PUBLIC cxx_std_14)
-	message(STATUS "Building in C++14 mode")
-else()
-	target_compile_features(torrent-rasterbar PUBLIC cxx_std_11)
-	message(STATUS "Building in C++11 mode")
+# make the CXX_STANDARD property public to ensure it makes it into the pkg-config file
+get_target_property(_std torrent-rasterbar CXX_STANDARD)
+if (${_std})
+	target_compile_features(torrent-rasterbar PUBLIC cxx_std_${_std})
```
zeule: Let me quote the manual once again:
`CXX_STANDARD`

This property specifies the C++ standard whose features are **requested**
to build this target. 

If the value requested **does not result** in a compile flag being added for
the compiler in use, **a previous standard flag will be added instead**.  This
means that using:
```cmake
 set_property(TARGET tgt PROPERTY CXX_STANDARD 11)
```
with a compiler which does not support ``-std=gnu++11`` or an equivalent
flag will not result in an error or warning, but will instead add the
``-std=gnu++98`` flag if supported.  This "decay" behavior may be controlled
with the ``CXX_STANDARD_REQUIRED`` target property.
Additionally, the ``CXX_EXTENSIONS`` target property may be used to
control whether compiler-specific extensions are enabled on a per-target basis.

*end of the quote*

But `target_compile_features()` sets features that can't be ignored. In other words, you change here the behaviour of ` CMAKE_CXX_STANDARD` variable and `CXX_STANDARD` making them enforcing as if `CXX_STANDARD_REQUIRED` was always set to `true`. Was it your intention?

arvidn: > But target_compile_features() sets features that can't be ignored. In other words, you change here the behaviour of CMAKE_CXX_STANDARD variable and CXX_STANDARD making them enforcing as if CXX_STANDARD_REQUIRED was always set to true. Was it your intention?

I don't see how this code does that. If the target doesn't have a value for the `CXX_STANDARD` property, this won't do anything. Presumably, if the target *does* have a `CXX_STANDARD` property set, surely it's *after* this decay logic has happened already, right? How else was the property set on the target?

zeule: > Presumably, if the target does have a CXX_STANDARD property set, surely it's after this decay logic has happened already, right?

I think no. The decay, as far as I understand, happens right when CMake compares the value of `CXX_STANDARD` with the list of standards supported by the compiler. If `CXX_STANDARD_REQUIRED` is not set, CMake silently drops the requirement and shows an error otherwise.

zeule: > so, you're saying the *effective* `CXX_STANDARD` property is determined at some point later? 

Yes.

> Is there a way to query that property?

I'm not aware of that. But we can test if the requested standard is on the `CMAKE_CXX_COMPILE_FEATURES` list. If it is there, `CXX_STANDARD` will be used. And we can safely enforce it via compile features. If not, we need to downgrade as well. It can be done taking a sublist of `CMAKE_CXX_KNOWN_FEATURES` (those prefixed with cxx_std_) and selecting previous records. We would need `list(FILTER)`, `list(TRANSFORM)`, `list(GET)` and a for-loop. I could write a snippet for doing that, if you like (albeit in a few hours).

> Or do you have an idea for how to do this differently?

I believe the test above does what is needed here.

zeule: https://gitlab.kitware.com/cmake/cmake/issues/18446

We can wait for their suggestions.

arvidn: perhaps simpler solution would be to check to see if `CMAKE_CXX_STANDARD_REQUIRED` is set to `ON`, and if it isn't, fail the build.

arvidn:
@zeule btw, do you have any ideas of why `test_common.lib` isn't being built (or at least not found) on windows using cmake? As far as I can tell, the dependencies are set up correctly.

zeule:
No ideas. Can you, please, try to change the add_library call for test_common to make it static?
add_library(test_common STATIC sources)

arvidn:
I pulled out the non-controversial/simple parts from the PR into a new one: https://github.com/arvidn/libtorrent/pull/3347

CMakeLists.txt
zeule: Maybe we'd better pack this mess into a macro and hide in its own file.

arvidn: @zeule would you mind posting this as a new PR?

zeule: OK

zeule: #3377

PR #3351, opened 2018-10-13
Retire autotools

arvidn:


zeule:
@arvidn, will this be merged into the 1.2? If so, does b2 toolchain generate pkg-config?

arvidn:
I think this is too big of a change to make it into 1.2. b2 does not create a `pkg-config` file (yet).

zeule:
If I backport CMake fixes for pkg-config would you accept the changes?

arvidn:
you mean back-port to `RC_1_1`? sure

zeule:
OK, thanks. Will do.

aldenml:
Is the ios compilation done with 64 bits by default? (I don't know)

arvidn:
it's the ios simulator, and it's probably configured as the old simulator, back when iOS was 32 bit

aldenml:
I think then that for iOS switching to 64 bits is reasonable and this problem should go away

PR #3375, opened 2018-10-20
remove params.url.clear();

pavel-pimenov:


arvidn:
Can you explain why you would like to make this change?

pavel-pimenov:
https://github.com/arvidn/libtorrent/blob/master/src/session_impl.cpp#L4883

PR #3377, opened 2018-10-20
cmake: allow selecting C++ standard version during build

zeule:
Also moves helper functions into a module file and replaces JOIN
generator expressions with list(TRANSFORM) commands in order to get
the correct sources list at the configure stage.


cmake/Modules/LibtorrentMacros.cmake
```diff
@@ -0,0 +1,83 @@
...
+		list(GET _supported_standards -1 _std) # if all tests in the loop below fail, we will continue with the oldest one
+		foreach(_s IN LISTS  _supported_standards)
+			if (cxx_std_${_s} IN_LIST CMAKE_CXX_COMPILE_FEATURES)
+				set(_std ${_s})
+				break()
+			endif()
+		endforeach()
+	else()
+		if (NOT ${_std} IN_LIST _supported_standards)
+		message(FATAL_ERROR "Sorry, C++${_std} is not supported by libtorrent")
```
ssiloti: Any standard version greater-than-or-equal-to C++11 should be permitted, not just what is in `_supported_standards`.

zeule: Makes sense. Updated the macro. @arvidn , this differs from what we have discussed.

CMakeLists.txt
```diff
@@ -580,8 +546,9 @@ if (WIN32)
 			debug dbghelp
 	)
 
+	add_definitions(-D_WIN32_WINNT=0x0600) # target Windows Vista or later
```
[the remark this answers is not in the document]
zeule: All targets in this directory and subdirectories as private definitions. Seems to be what we need?

sledgehammer999:
@arvidn (or @zeule ) please consider backporting this into RC_1_1. Possibly before 1.1.11 is released.

zeule:
@sledgehammer999, I'm working on a PR.

PR #3385, opened 2018-10-26
ssl listen port fix and improved tracker announce logging

arvidn:


arvidn:
I think the main reason this hasn't been caught in testing is because the simulator can't be built with ssl enabled.

src/torrent.cpp
```diff
@@ -3713,6 +3707,9 @@ namespace {
 
 				req.private_torrent = m_torrent_file->priv();
 				req.second_announce = true;
+#ifdef TORRENT_USE_OPENSSL
+				req.ssl_ctx = NULL;
```
arvidn: this is the main fix

src/session_impl.cpp
```diff
@@ -1320,7 +1320,7 @@ namespace aux {
 #ifdef TORRENT_USE_OPENSSL
 		// SSL torrents use the SSL listen port
 		if (req.ssl_ctx) req.listen_port = ssl_listen_port();
-		req.ssl_ctx = &m_ssl_ctx;
+		else req.ssl_ctx = &m_ssl_ctx;
```
arvidn: the previous logic here seemed suspicious too. surely if the torrent has a cert, that's what we should authenticate the tracker against

Issue #3382, opened 2018-10-24
Random offset writes happen in sequential download mode


zvirja:
**Environment**

_libtorrent version (or branch):_ I use the latest version of [libretorrent app](https://gitlab.com/proninyaroslav/libretorrent). Expects it uses the latest version of the library.
_platform/architecture:_ Android 7, ARMv7
_compiler and compiler version:_ Don't know.

**Disclaimer**

First of all please notice that I never used this library for the development, so might miss some details. However, I've examined the API shape and am very confident you are the right target for this issue. Will try to describe the issue from the user's perspective, so we can together discuss what could be done here.

**Scenario**

I'm an owner of the Galaxy Tab S2 tablet and have an external SD Card installed. That card is not super fast - for sequential write it can give only 10-15 MB/s, for random it's much worse. It's formatted with `exFAT` file system and I cannot format it to the `ext3/ext4/ext?` - device simply doesn't recognize it then. Android uses FUSE and creates a virtual file system on top of that - on my Android device it's named as `sdcardfs` if I check that via `mount`. The issue is that at the end of this chain the `sparse` files are not supported (I suspect that's because of the underlying file system, but never had a chance to prove that).

When I start to download a torrent file to my SD Card, I _always_ enable the sequential mode. Otherwise, any write (seek/access?) to the advanced file position (e.g. to middle) forces kernel to fill file with zeros. The sequential mode does the trick and torrent starts to download at stable 9-10MB/s speed. However, after short time something goes wrong and download gets stuck. When monitoring the file during that time, I can see that it starts to grow and becomes filled with zeros, which indicates that seek to the end occurred. It happens randomly - sometimes soon after start, sometimes after it downloaded quarter or even half of the file - but happens almost for each time. After the file is fully created on the file system, the download resumes back at the previous speed, which indicates that normally write happens pretty linearly. Obviously, this stuck significantly increases the download time. For instance, for 10GB file it takes ~20min to fill it with zeros. Later it takes the same time to download the actual file content. Basically, this issue doubles the download time.

I noticed that behavior with the all torrent clients I tested. All of them suffer from the issue and I know that a lot of them use `libtorrent` library under the hood (e.g. libretorrent or Flud). Therefore, I made a conclusion that it's library related issue.

As far as I understand, the only ways we could prevent kernel from filling files with zeros are:
- use sparse files, like we do on modern file systems. Here it's not available due to SD Card file system.
- perform only linear writes without seeking too far (or without seeking at all).

[...]

PR #3386, opened 2018-10-26
don't pick from the end in sequential mode

arvidn:
only make snubbed peers invert the piece picking strategy when we're doing rarest first (i.e. snubbed peers do common-first). Specifically, this prevents snubbed peers from picking from the end of the torrent when in sequential mode.

This specifically would address https://github.com/arvidn/libtorrent/issues/3382. I can't think of any other reasonable use cases for sequential mode, so perhaps this is a reasonable change in behaviour (i.e. without introducing it as a new option).

@ssiloti do you have any thoughts?

### DIFF
--- a/src/peer_connection.cpp
+++ b/src/peer_connection.cpp
@@ -880,15 +880,16 @@ namespace libtorrent
 		else
 		{
 			ret |= piece_picker::rarest_first;
-		}
 
-		if (m_snubbed)
-		{
-			// snubbed peers should request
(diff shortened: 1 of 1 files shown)


zvirja:
@ssiloti @arvidn Be aware of [my answer](https://github.com/arvidn/libtorrent/issues/3382#issuecomment-433659227). This fix is not enough to fix the problem completely (while it still makes it one step closer).

PR #3390, opened 2018-10-27
Provide virtual destructors for classes with virtual methods

janisozaur:


arvidn:
I don't think it's a good idea to do this. Would you care to explain your reasoning?
If the purpose is to silence "LGTM", I see two problems. (1) I don't think it's not good to make arbitrary style changes based on a tool (unless those style changes makes code less error prone). Because every tool may have a slightly different opinion on style. (2) The rule in LGTM is wrong. There is no way to accidentally delete a derived class through a base class, because the destructors are protected. This is a well established pattern.

In sort, as far as I can tell, making these destructors virtual *only* has downsides (of additional cost in time and space) and not a single upside (other than making one arbitrary tool happy). As far as I can tell, there is no safety improvement in doing this.

I am open minded about it though, and I would be happy to be convinced otherwise.

arvidn:
related to the LGTM tool, I reported this issue [here](https://discuss.lgtm.com/t/false-positive-c-base-classes-with-a-virtual-function-must-define-a-virtual-destructor/1422).

PR #3399, opened 2018-11-03
make span's and sha1_hash's index and size types signed

arvidn:
instead of unsigned `std::size_t`. Specifically `std::ptrdiff_t`. In line with the direction of gsl span and reduce signed->unsigned casts.

This was prompted by the (WIP) [patch](https://github.com/qbittorrent/qBittorrent/pull/9704/files#diff-b369e532eff50253387c9d203da4b928R207) to bring qBT to be compatible with libtorrent-1.2 RC. Specifically, where indices into the counters array was changed from `int` to `std::size_t` to avoid a type conversion warning as the counters changed from a raw pointer to a `lt::span`.

There is some movement in the C++ community to try to fix the problem of all containers using unsigned `std::size_t` as size and index. Herb explains it quite thoroughly in this [PR against the core guidelines](https://github.com/isocpp/CppCoreGuidelines/pull/1115).

I subscribe to gabby's view, as he summarises [here](https://github.com/Microsoft/GSL/issues/171#issuecomment-156202752).

The GSL `span` has a `ptrdiff_t` index type (see [here](https://github.com/Microsoft/GSL/blob/master/include/gsl/span#L361)), and it has had it since 2016.

So, why are signed integers better than unsigned? I've covered this in a [lightning talk](https://www.youtube.com/watch?v=y_0KTXuSOJ4) a while ago, as well as [Jon Kalb's lightning talk from CPPCON](https://www.youtube.com/watch?v=wvtFGa6XJDU).

@aldenml since you did the bulk of the work to rid libtorrent of `-Wconversion` warnings, would you mind taking a look? Do you think this makes sense?

@zeule Since this would affect your patch for libtorrent-1.2, do you have an opinion?

The core changes are in `span.hpp` and `sha1_hash.hpp`.

### DIFF
--- a/include/libtorrent/aux_/io.hpp
+++ b/include/libtorrent/aux_/io.hpp
@@ -37,6 +37,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <string>
 #include <type_traits>
 #include "libtorrent/span.hpp"
+#include "libtorrent/aux_/numeric_cast.hpp"
 
 namespace libtorrent { namespace aux {
 
@@ -55,7 +56,7 @@ namespace libtorrent { namespace aux {
 			ret <<= 8;
 			ret |= static_cast<std::uint8_t>(b);
 		}
-		view = view.subspan(sizeof(T));
+		view = view.subspan(int(sizeof(T)));
 		return ret;
 	}
 
@@ -73,7 +74,7 @@ namespace libtorrent { namespace aux {
 			shift -= 8;
 			b = static_cast<Byte>((val >> shift) & 0xff);
 		}
-		view = view.subspan(sizeof(T));
+		view = view.subspan(int(sizeof(T)));
 	}
 
 	// the single-byte case is separate to avoid a warning on the shift-left by
@@ -167,8 +168,9 @@ namespace libtorrent { namespace aux {
 	template<typename Byte>
 	inline int write_string(std::string const& str, span<Byte>& view)
 	{
+		TORRENT_ASSERT(view.size() >= numeric_cast<int>(str.size()));
 		std::copy(str.begin(), str.end(), view.begin());
-		view = view.subspan(str.size());
+		view = view.subspan(int(str.size()));
 		return int(str.size());
 	}
 
--- a/include/libtorrent/receive_buffer.hpp
+++ b/include/libtorrent/receive_buffer.hpp
@@ -155,7 +155,7 @@ private:
 
 	// keep track of how much of the receive buffer we use, if we're not using
 	// enough of it we shrink it
-	sliding_average<std::int64_t, 20> m_watermark;
+	sliding_average<std::ptrdiff_t, 20> m_watermark;
 
 	buffer m_recv_buffer;
 };
@@ -207,7 +207,7 @@ struct crypto_receive_buffer
 
 	span<char const> get() const;
 
-	span<char> mutable_buffer(std::size_t bytes);
+	span<char> mutable_buffer(int bytes);
 
 private:
 	// explicitly disallow assignment, to silence msvc warning
--- a/test/test_bdecode.cpp
+++ b/test/test_bdecode.cpp
@@ -500,7 +500,7 @@ TORRENT_TEST(item_limit)
 {
 	char b[10240];
 	b[0] = 'l';
-	std::size_t i = 1;
+	std::ptrdiff_t i = 1;
 	for (i = 1; i < 10239; i += 2)
 		memcpy(&b[i], "0:", 2);
 	b[i] = 'e';
@@ -782,18 +782,18 @@ TORRENT_TEST(parse_int_overflow)
 
 TORRENT_TEST(parse_length_overflow)
 {
-	char const* b[] = {
-		"d1:a1919191010:11111",
-		"d2143289344:a4:aaaae",
-		"d214328934114:a4:aaaae",
-		"d9205357638345293824:a4:aaaae",
-		"d1:a9205357638345293824:11111",
+	string_view const b[] = {
+		"d1:a1919191010:11111"_sv,
+		"d2143289344:a4:aaaae"_sv,
+		"d214328934114:a4:aaaae"_sv,
+		"d9205357638345293824:a4:aaaae"_sv,
+		"d1:a9205357638345293824:11111"_sv
 	};
 
-	for (int i = 0; i < int(sizeof(b)/sizeof(b[0])); ++i)
+	for (auto const& buf : b)
 	{
 		error_code ec;
-		bdecode_node e = bdecode({b[i], strlen(b[i])}, ec);
+		bdecode_node e = bdecode(buf, ec);
 		TEST_EQUAL(ec, error_code(bdecode_errors::unexpected_eof));
 	}
 }
--- a/ed25519/src/sign.cpp
+++ b/ed25519/src/sign.cpp
@@ -9,7 +9,7 @@
 namespace libtorrent
 {
 
-void ed25519_sign(unsigned char *signature, const unsigned char *message, size_t message_len, const unsigned char *public_key, const unsigned char *private_key) {
+void ed25519_sign(unsigned char *signature, const unsigned char *message, std::ptrdiff_t message_len, const unsigned char *public_key, const unsigned char *private_key) {
     ge_p3 R;
 
     hasher512 hash;
--- a/ed25519/src/verify.cpp
+++ b/ed25519/src/verify.cpp
@@ -50,7 +50,7 @@ static int consttime_equal(const unsigned char *x, const unsigned char *y) {
     return !r;
 }
 
-int ed25519_verify(const unsigned char *signature, const unsigned char *message, size_t message_len, const unsigned char *public_key) {
+int ed25519_verify(const unsigned char *signature, const unsigned char *message, std::ptrdiff_t message_len, const unsigned char *public_key) {
     unsigned char checker[32];
     ge_p3 A;
     ge_p2 R;
--- a/examples/connection_tester.cpp
+++ b/examples/connection_tester.cpp
@@ -676,7 +676,7 @@ struct peer_conn
 
 	void write_piece(piece_index_t const piece, int start, int length)
 	{
-		generate_block({write_buffer, static_cast<std::size_t>(length / 4)}
+		generate_block({write_buffer, length / 4}
 			, piece, start);
 
 		if (corrupt)
@@ -860,7 +860,7 @@ void generate_data(char const* path, torrent_info const& ti)
 			generate_block(piece, i, j);
 			int const left_in_piece = ti.piece_size(i) - j;
 			iovec_t const b = { reinterpret_cast<char*>(piece)
-				, size_t(std::min(left_in_piece, 0x4000))};
+				, std::min(left_in_piece, 0x4000)};
 			storage_error error;
 			st->writev(b, i, j, open_mode::write_only, error);
 			if (error)
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -2599,7 +2599,7 @@ TORRENT_VERSION_NAMESPACE_2
 	private:
 		std::reference_wrapper<aux::stack_allocator> m_alloc;
 		aux::allocation_slot m_msg_idx;
-		std::size_t const m_size;
+		int const m_size;
 #if TORRENT_ABI_VERSION == 1
 	public:
 		direction_t TORRENT_DEPRECATED_MEMBER dir;
--- a/include/libtorrent/aux_/alloca.hpp
+++ b/include/libtorrent/aux_/alloca.hpp
@@ -78,8 +78,8 @@ struct alloca_destructor
 
 #include <malloc.h>
 #define TORRENT_ALLOCA(v, t, n) ::libtorrent::aux::typed_span<t> v; { \
-	auto TORRENT_ALLOCA_size = ::libtorrent::aux::numeric_cast<std::size_t>(n); \
-	auto* TORRENT_ALLOCA_tmp = static_cast<t*>(_alloca(sizeof(t) * TORRENT_ALLOCA_size)); \
+	auto TORRENT_ALLOCA_size = ::libtorrent::aux::numeric_cast<std::ptrdiff_t>(n); \
+	auto* TORRENT_ALLOCA_tmp = static_cast<t*>(_alloca(sizeof(t) * static_cast<std::size_t>(TORRENT_ALLOCA_size))); \
 	v = ::libtorrent::aux::typed_span<t>(TORRENT_ALLOCA_tmp, TORRENT_ALLOCA_size); \
 	::libtorrent::aux::uninitialized_default_construct(v.begin(), v.end()); \
 	} \
@@ -89,8 +89,8 @@ struct alloca_destructor
 
 #include <stdlib.h>
 #define TORRENT_ALLOCA(v, t, n) ::libtorrent::aux::typed_span<t> v; { \
-	auto TORRENT_ALLOCA_size = ::libtorrent::aux::numeric_cast<std::size_t>(n); \
-	auto* TORRENT_ALLOCA_tmp = static_cast<t*>(alloca(sizeof(t) * TORRENT_ALLOCA_size)); \
+	auto TORRENT_ALLOCA_size = ::libtorrent::aux::numeric_cast<std::ptrdiff_t>(n); \
+	auto* TORRENT_ALLOCA_tmp = static_cast<t*>(alloca(sizeof(t) * static_cast<std::size_t>(TORRENT_ALLOCA_size))); \
 	v = ::libtorrent::aux::typed_span<t>(TORRENT_ALLOCA_tmp, TORRENT_ALLOCA_size); \
 	::libtorrent::aux::uninitialized_default_construct(v.begin(), v.end()); \
 	} \
@@ -100,8 +100,8 @@ struct alloca_destructor
 
 #include <alloca.h>
 #define TORRENT_ALLOCA(v, t, n) ::libtorrent::aux::typed_span<t> v; { \
-	auto TORRENT_ALLOCA_size = ::libtorrent::aux::numeric_cast<std::size_t>(n); \
-	auto* TORRENT_ALLOCA_tmp = static_cast<t*>(alloca(sizeof(t) * TORRENT_ALLOCA_size)); \
+	auto TORRENT_ALLOCA_size = ::libtorrent::aux::numeric_cast<std::ptrdiff_t>(n); \
+	auto* TORRENT_ALLOCA_tmp = static_cast<t*>(alloca(sizeof(t) * static_cast<std::size_t>(TORRENT_ALLOCA_size))); \
 	v = ::libtorrent::aux::typed_span<t>(TORRENT_ALLOCA_tmp, TORRENT_ALLOCA_size); \
 	::libtorrent::aux::uninitialized_default_construct(v.begin(), v.end()); \
 	} \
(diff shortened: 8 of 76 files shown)


aldenml:
It totally makes sense to me. I would like to review this PR in detail completely, but so far, only half

zeule:
> @zeule Since this would affect your patch for libtorrent-1.2, do you have an opinion?

Thank for the invitation. I'm not going to make a review, just leave a short comment. These changes and all the hassle about unsigned arithmetic make no sense to me, neither theoretical nor practical.  Indices are not integer numbers, as natural numbers ≠ integer numbers. Period. The fact that arithmetic operations are not complete on the set of cardinals does not mean we want to use complex numbers everywhere. As to the practical side, nobody cares that `x/y*y != x` when using integers, why should we care about index operations more than that? And you don't want to use floats for indexing just because you can divide and multiply index values. Any compiler provides an option to separate signed and unsigned numbers in the very same way as you can separate integers and floating points numbers. Isn't that enough?

arvidn:
@zeule thanks for your comment. Unfortunately I'm not quite following all of it.

> all the hassle about unsigned arithmetic make no sense to me

Are you referring to preferring signed over unsigned? I'm suggesting *avoiding* unsigned arithmetic (i.e. modulo arithmetic), i.e. getting rid of the "hassle". But perhaps you're referring to *caring* about the arithmetic in the first place.

> Indices are not integer numbers, as natural numbers ≠ integer numbers. Period.

C++ doesn't model natural numbers though (I think it's a mistake to treat the >= 0 as the most important property of unsigned types, the most important property is that you get modulo 2 arithmetic). Also, I think it's a bit of a stretch to say that an *index* (in the C and C++ context) is a natural number. You've always been able to index elements *before* a pointer, not just after.

> The fact that arithmetic operations are not complete on the set of cardinals does not mean we want to use complex numbers everywhere.

Just like nobody (most of the time) care about whether results of integer arithmetic wrap around or not, so we should be using the simpler type. Just like complex numbers have more complicated arithmetic than scalars, so do unsigned integers compared to signed integers. The compiler sometimes need to add extra instructions to ensure that the result of an arithmetic operation wraps correctly. Mostly you don't care, but you still pay for it.

> As to the practical side, nobody cares that x/y*y != x when using integers, why should we care about index operations more than that?

I'm not following this. currently, the operations to "index" are made up of `static_cast<std::size_t>(...)`, because `operator[]` takes a `std::size_t`. All those casts are what I'm trying to get rid of (or at least minimise the need for them). The reason I brought up signed vs. unsigned in this context is that almost all computations are done with signed types, and every now and then the results from those computations are used as indices. That requires casts. The most reasonable way to avoid the casts is to make the `operator[]`'s take signed types.

> And you don't want to use floats for indexing just because you can divide and multiply index values. Any compiler provides an option to separate signed and unsigned numbers in the very same way as you can separate integers and floating points numbers. Isn't that enough?

I don't know what you mean by this. By separating `signed` and `unsigned`, do you mean something like `-Wsign-conversion` and `-Wconversion`? If so, those flags are fundamental to highlight where the implicit conversions happen, so we can make them explicit. I don't get the analogy with `float`, I think the implicit conversion to and from `float` are problematic too (and addressed by those warning flags).

Are you saying that converting a signed integer to an unsigned index should have the same "weight" as a conversion from a float to an unsigned index?

aldenml:
@zeule  The problem of having signed and unsigned "integer" types have some theoretical implications. The fact that unsigned "integers" are not closed under subtraction, makes for really complicated math for formal proofs. From the practical point of view, it gets even worse if you have automatic promotion (or automatic truncation) of integral types, which is just a special case of implicit conversion. I also learned of the particular issues in cpp here https://blog.libtorrent.org/2016/05/unsigned-integers/

@arvidn Now that `span` takes signed, what do you think of putting `TORRENT_ASSERT`s to test that the passing size is positive?

Also, I remember creating `typed_span` for workaround this specific issue, and I don't recall using it for any other type of index. Do you think it is something to remove?

zeule:
Please don't mix engineering and science. There is well known definition for natural numbers. Numeral system is orthogonal to that. Indeed, the set is closed only against summation and multiplication, the other 2 arithmetic operations produce negative and rational numbers. Perhaps that complicates life for automatic proof systems, I don't know, as I'm physicist but not a mathematician and used to handle numbers with a light heart. Nevertheless it is obvious to me that pointer subscripts are not indices I was talking about because no one (obviously) is ready to have a map from natural numbers to integers in mind every time  a pointer is used. Heh, people trouble sometimes with much simpler mapping --- naturals to non-negative! 

Approaching from he other side, I would venture to say that `sizeof()` will always be unsigned, as well as `::size()` of standard containers, simply because C++ is used for close to metal things. As such, the PR against the CppCoreGuidelines cited above, is a pure left-wing populism, if you allow me to use a political terminology. Singed index for span classes goes straight against their very definition, don't you see this? The suggestion to put an assertion that test index value against zero perfectly highlights the nonsense.   

Concluding, I want to say that natural numbers (and unsigned numbers in a CPU) should be distinguished from integer numbers in the very same way as we try to not mix floating point numbers with integers. If compiler shows a warning about signed/unsigned mismatch that doesn't mean you should put an explicit cast at that very place, no, you'd better rewrite the code around to avoid signed numbers. And think about iterators. 

I'm surprised that a C-era concept of raw pointers now changes fundamentals of the C++ library, while all the previous years we were told that raw pointers are better to be avoided.

arvidn:
I think it's well accepted that the problem with unsigned integers is not (primarily) that they can't represent negative values, but that they have modulo arithmetic (which is very surprising to humans and typically a very undesirable property). Again, I think it's a mistake to suggest that an unsigned integer (in C and C++) represent natural numbers. They primarily represent numbers with modulo arithmetic.

regarding `sizeof()` and containers `::size()` returning `size_t` is closer to the metal, I think an argument can be made of the opposite. In practice (as I understand it) extremely few systems support objects larger than half the address space.

I think we all agree that `sizeof()` always returns a positive number. The argument against that value be represented by an unsigned integer is that the size of an object should not have modulo arithmetic. That last bit you lose by using signed is a much smaller problem than the problems introduced by modulo arithmetic. Note that when I say "unsigned integer", I don't mean an integer with no sign, I *specifically* mean the C/C++ integral types with the `unsigned` type modifier. i.e. something with modulo arithmetic.

> If compiler shows a warning about signed/unsigned mismatch that doesn't mean you should put an explicit cast at that very place, no, you'd better rewrite the code around to avoid signed numbers. And think about iterators.

But virtually every expression in all programs *need* normal arithmetic to work. It's unrealistic to use *more* modulo arithmetic. It's not just more error prone and more difficult to work with, you ask the compiler to generate code to do more work for things you don't want (specifically to mask results to ensure correct wrapping behavior). In fact, I would venture to claim the vast majority of arithmetic operations on unsigned integers in C and C++ programs are bugs. (I don't count bitwise logic- or shift operators as arithmetic operators).

> I'm surprised that a C-era concept of raw pointers now changes fundamentals of the C++ library, while all the previous years we were told that raw pointers are better to be avoided.

I don't see how this affects or interacts with avoiding owning raw pointers

Chocobo1:
@arvidn 
Just want to say thanks to the summary in the opening post, I never know why `gsl::span::index_type == std::ptrdiff_t` and the signature `index_type span::size() {}` until now.
>The argument against that value be represented by an unsigned integer is that the size of an object should not have modulo arithmetic. 

The examples at http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-subscripts make sense to me, although I've not get used to it

zeule:
Numbers in computers wrap due to the limited number of digits in the CPU, and not due to a decision  was made to design unsigned or signed numbers in that or another way. Are you stating that the way signed numbers wrap is less surprising? Can't believe you do, because they wrap in the very same way, just adding additional complication because of the sign. You do repeat "modulo arithmetic" so many times, but I can't understand what do you blame it for. It is the nature of the device we use for the computation; nothing can be done about that. Besides, each person who uses measurement systems is familiar with modulo arithmetic. Clocks and calendars are good examples.

> But virtually every expression in all programs need normal arithmetic to work.

This statement is non falsifiable. I can object that virtually all manipulations with pointers that go beyond  `+=` and `-=` are signs of logical mistakes.

> I don't see how this affects or interacts with avoiding owning raw pointers

The signed indices you are advocating make sense only for pointers (i.e. arrays with undefined lower bound). But all the containers you design have a well defined lower bound. And then one needs negative numbers only for temporals, which is very likely is a design mistake. I beliebe they are mistakes because that need can be avoided by simplifying the design. BTW, this statement is falsifiable and if you want to continue arguing, I would appreciate an example where signed arithmetic produces less error-prone code (but of course I don't mean stupid examples with loop counters like in the Sutter's PR. They are stupid because a) each an every compiler loudly warns about such mistypes, even if ones forgot about auto-typed variables, and b) these loops have to use iterators).

include/libtorrent/receive_buffer.hpp
```diff
@@ -155,7 +155,7 @@ struct TORRENT_EXTRA_EXPORT receive_buffer
 
 	// keep track of how much of the receive buffer we use, if we're not using
 	// enough of it we shrink it
-	sliding_average<std::int64_t, 20> m_watermark;
+	sliding_average<std::ptrdiff_t, 20> m_watermark;
```
aldenml: I wonder that the warning was here

aldenml:
Hi @arvidn, I would like to hear your opinion about removing `typed_span` and the use of `TORRENT_ASSERT` inside `span`

include/libtorrent/aux_/io.hpp
```diff
@@ -50,12 +51,12 @@ namespace libtorrent { namespace aux {
 	read_impl(span<Byte>& view, type<T>)
 	{
 		T ret = 0;
-		for (std::size_t i = 0; i < sizeof(T); ++i)
+		for (int i = 0; i < int(sizeof(T)); ++i)
```
zeule: What's the point of this change?

aldenml: I bet it's to avoid the cast at `ret |= static_cast<std::uint8_t>(view[i]);` (but I know, two new casts instead)

zeule: Precisely.

include/libtorrent/receive_buffer.hpp
arvidn: the mean value is used to specify a new size for a buffer, which means it should be a "buffer size" type, not `int64_t`. On second though though, I wonder if this increases the risk of overflow on 32 bit systems

include/libtorrent/aux_/io.hpp
arvidn: in my mind `sizeof()` is the operator at fault here. It should really return `std::ptrdiff_t`. I think this makes that clearer.
Btw, the second cast (on line 59) would have to be there regardless of whether `i` is signed or not. So changing the type of the counter does not introduce a net new cast, it just moves one.

It makes me wonder if it would make sense, in the future, to do something like:
```
template <typename T>
inline std::ptrdiff_t sizeof_v(sizeof(T));
```
(with some appropriate static asserts perhaps)

In addition to that, I prefer [signed loop counters](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es107-dont-use-unsigned-for-subscripts-prefer-gslindex).

Perhaps this would be better as a range-for loop though, over: `view.first(sizeof(T))`

arvidn:
@aldenml 

> Now that span takes signed, what do you think of putting TORRENT_ASSERTs to test that the passing size is positive?

yes, that makes sense. I'll add that.

> Also, I remember creating typed_span for workaround this specific issue, and I don't recall using it for any other type of index. Do you think it is something to remove?

I'll take a look. It seems like a reasonable thing to have in general though, as a span could be a range of something indexed by pieces. Although, it becomes a bit weird in that specific case, as you're really addressing it with "delta"-piece indices. at that point. or "piece index offsets".

arvidn:
yeah, using strong types for indices in subspans seems wrong. piece_t{0} will mean something else. I will see if it's easy to remove.

arvidn:
> Numbers in computers wrap due to the limited number of digits in the CPU, and not due to a decision was made to design unsigned or signed numbers in that or another way. Are you stating that the way signed numbers wrap is less surprising?

Yes. Especially around 0, but also to a degree at the outer boundaries (but only in the sense that the compiler is more likely to generate the code you would expect).

> Can't believe you do, because they wrap in the very same way, just adding additional complication because of the sign.

The main difference is that they wrap at different boundary values.

but they also don't wrap the same *way* in C and C++. signed integer overflow has undefined behavior. In other words, they don't wrap. It lets you communicate more precisely with the compiler that your integers are not supposed to overflow (which I would claim is the most common types of integers).

> You do repeat "modulo arithmetic" so many times, but I can't understand what do you blame it for. It is the nature of the device we use for the computation; nothing can be done about that.

Oh I think something *can* be done about it. You can make it undefined behavior (just like signed integers). This still wouldn't solve the risk of having negative intermediate values (or negative results for that matter) in expressions.

I think fundamentally the issue is that values in programs are very often near 0, even for only positive values. With signed integers, the boundaries where bad things happen are far away from the values you normally represent, with unsigned, one boundary is *very* close to your typical numbers. In addition, since the wrapping is well defined, you can accurately diagnose them as errors, even though they most likely are.

> > But virtually every expression in all programs need normal arithmetic to work.
> 
> This statement is non falsifiable.

It *is* falsifiable. But I should probably have said signed integers, as the the ability to represent negative values is what makes them correct. I perhaps also should have been explicit about how I assume no implicit or explicit conversions (or at least minimal casts).

You could survey a vary large sample of programs.

expressions that I would consider needing normal arithmetic:

(1) An expression where negative numbers are expected
(2) An expression whose result is used in (1)
(3) An expression with any input being the result of (1)

I would also include expressions where overflow is an error, as the only way to communicate this to the compiler is via signed integers, this primarily affects quality of generated code though.

And I would probably exclude expressions that require the top bit of the machine word to fit its workload, as the only way to get that bit is via unsigned.

(2) and (3) comes from the expectation that we're building with `-Wsign-conversion` and `-Werror`, we're on the same page there, right?

The scenario I can think of (let's call it A), which it sounds like you're suggesting, is that the vast majority of expressions only deal with positive numbers, both results and intermediate values. And there is a relatively small surface area where that world interacts with expressions that deal with negative numbers, and you would only need a relatively small number of casts there.

With that assumption my criteria above would have to change, clearly. I can be very hard to determine whether expressions legitimately have negative intermediate values though. iirc, I was bitten by this a few times in the rate limiter code.

A response to that though (let's call it B), perhaps the number expressions need the 64th bit to be able to fit its workload is probably a lot smaller than the cases where you have negative numbers. So, if you use *signed* integer everywhere, except in those rare cases, you most likely have a much smaller surface area where you need to cast.

(B) is definitely true in my experience, in the domains that I have worked in. It's quite likely there are domains where (A) is true, just none of the ones I have experience with (nor the authors I read, but perhaps there's selection bias).

> The signed indices you are advocating make sense only for pointers (i.e. arrays with undefined lower bound).

I disagree. The signed indices makes sense because they let you use values that come from expressions supporting negative numbers, without having to cast to unsigned. i.e. you can use `-Wsign-conversion` and `-Werror` without an explicit cast. They also provide normal arithmetic, instead of modulo arithmetic.

> But all the containers you design have a well defined lower bound. And then one needs negative numbers only for temporals, which is very likely is a design mistake.

Only in scenario (A) above (which it sounds like you subscribe to). I live firmly in (B).

> I believe they are mistakes because that need can be avoided by simplifying the design.

Perhaps you have tried this, but if you haven't I would suggest you build any non-trivial C++ program with `-Wconversion` and `-Wsign-conversion`. I tried bitcoin core recently for example. Then start looking over the code that triggers the warnings and try to come up with ways to silence the warnings without explicit casts (especially between signed and unsigned). It's extremely difficult, because values that cannot be negative are mixed with values that *can* be negative a lot.

You end up with a lost of casting. @aldenml did the bulk of the work to get libtorrent to build with `-Wconversion`, `-Wsign-conversion` and `-Werror`, and there were a lot of casts. Other people have had the same experience, and also come to the conclusion that it was a mistake to use `std::size_t` as index- and size type for standard containers.

> BTW, this statement is falsifiable and if you want to continue arguing, I would appreciate an example where signed arithmetic produces less error-prone code (but of course I don't mean stupid examples with loop counters like in the Sutter's PR. They are stupid because a) each an every compiler loudly warns about such mistypes, even if ones forgot about auto-typed variables, and b) these loops have to use iterators).

I haven't indexed my own bugs based on this category, I could probably dig up a few from the libtorrent revision history, but they would be quite old. There are a few good ones on reddit though, for example:

https://www.reddit.com/r/cpp/comments/8r5b3k/this_from_gslspan_is_a_good_example_of_why_the_c/e0qn9kw/

the compiler warnings you get sometimes only help you so much. If the solution to 95% of them is to insert an explicit cast, there's a risk of "cast fatigue", and just silence the warnings by habit.

arvidn:
rebased

arvidn:
I proposing expressions in programs can be classified in 3 categories:

1. those that need negative values, intermediate or otherwise. (this is where you must use signed integers)
2. those that only need positive values, always (this could be either signed or unsigned)
3. those that need all 64 bits of positive values, or that rely on `UINT64_MAX` + 1 wrapping to 0. (this is where you must use unsigned integers)

I'm also assuming that all these classes of expressions interact sometimes to some degree (i.e. values from one class are used in expression in another class).

What I referred to as scenario (A) was where the set of (3) is larger than (1). Or at least where the interaction surface of (2) and (3) against (1) is very small. Smaller than the surface between (1) and (2) against (3). Then it would require fewer casts to use unsigned by default.

What I referred to as scenario (B) was where (1) is larger than (3), or more specifically, the surface area of (1) and (2) against (3) is really small compared to the surface area of (2) and (3) against (1).

In the domains I have worked (3) has been very small, and (1) has been relatively large.

Currently, libtorrent hasn't lives somewhere in between scenario (A) and (B). It accepts (to some degree) that indices are unsigned and use unsigned loop counters sometimes as a result. This has resulted in a lot of signed/unsigned casts, to the point where we have wrapped some containers to give them signed indices. Moving towards scenario (B) should definitely have a net decrease of the number of casts. The set of (3) in libtorrent is quite small. The only parts I can think of are gzip, sha1, rc4, and uTP sequence numbers.

zeule:
> In the domains I have worked (3) has been very small, and (1) has been relatively large.

I could have argued that it depends on your own style more than on the domain. But we seem to reach a dead-end here.

You want to freely intermix pointers and indices, and hence, IMO, the need for negative values. Libtorrent API supports shows that exporting a lot via indexed collections. On the other hand, I'm scared by your idea written somewhere above to add `sizeof_v()`, because that is close to what macros can do: changing fundamental language semantic, which makes harm for those who are not limited to a single codebase for their whole life. The same applies to the containers in the libtorrent namespace with signed index types but named after `std::` ones.

aldenml:
Hi @zeule, quick question, is `libtorrent` leaking the "indexed collections" in the public API? that shouldn't be the case, they are only for internal use

zeule:
@aldenml, sorry, I probably used too common wording. What I meant is excessive usage of indices to address objects in collections. For example, https://github.com/arvidn/libtorrent/blob/master/include/libtorrent/torrent_info.hpp#L237. And consider [this](https://github.com/arvidn/libtorrent/blob/master/include/libtorrent/session_stats.hpp#L71) example. Why does it return integers? Am I supposed to subtract metric indices? Another example is [torrent_handle::file_progress()](https://github.com/arvidn/libtorrent/blob/master/include/libtorrent/torrent_handle.hpp#L444) simply says that index is a "connection point" between files and their order. The same is true for pieces and blocks.

test/test_bdecode.cpp
```diff
@@ -782,18 +782,18 @@ TORRENT_TEST(parse_int_overflow)
...
 TORRENT_TEST(parse_length_overflow)
 {
-	char const* b[] = {
-		"d1:a1919191010:11111",
-		"d2143289344:a4:aaaae",
-		"d214328934114:a4:aaaae",
-		"d9205357638345293824:a4:aaaae",
-		"d1:a9205357638345293824:11111",
+	string_view const b[] = {
+		string_view("d1:a1919191010:11111"),
```
aldenml: I think you have `string_view operator""_sv` for this

PR #3402, opened 2018-11-05
Remove typed span

arvidn:


test/setup_transfer.cpp
```diff
@@ -115,7 +115,7 @@ sha1_hash rand_hash()
 {
 	sha1_hash ret;
 	for (int i = 0; i < 20; ++i)
-		ret[static_cast<std::size_t>(i)] = std::uint8_t(lt::random(0xff));
+		ret[i] = std::uint8_t(lt::random(0xff));
```
aldenml: not for this PR, but this can be replaced with
```cpp
sha1_hash ret;
aux::random_bytes(ret);
return ret;
```
like `generate_next` in `test_dht.cpp`

PR #3403, opened 2018-11-05
some cleanup of memcmp()/memcpy()

arvidn:
by turning them into span comparisons and std::copy(). also some more use of random_bytes() instead of loops.


arvidn:
@aldenml I pulled out some of the unrelated fixes from https://github.com/arvidn/libtorrent/pull/3399 into a separate pull request. and I included your suggesting of using `random_bytes()`.

aldenml:
👍 LGTM

PR #3409, opened 2018-11-07
convert classic for-loops to range-for loops

arvidn:
revert use of CHAR_BIT macro (we use fixed width types, not char)

### DIFF
--- a/include/libtorrent/sha1_hash.hpp
+++ b/include/libtorrent/sha1_hash.hpp
@@ -38,7 +38,10 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <string>
 #include <cstring>
 #include <array>
-#include <climits> // for CHAR_BIT
+
+#include "libtorrent/aux_/disable_warnings_push.hpp"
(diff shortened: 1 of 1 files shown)


arvidn:
@aldenml is that better?

aldenml:
Well, it's a matter of taste I guess, getting the argument to an extreme: "Now I need to look around to see how `boost::combine` was used to determine what `boost::get<>` exactly is."

and I rather not use boost when possible :)

arvidn:
`boost::get<>` is the predecessor to `std::get<>`, i.e. to get values out of tuples.

aldenml:
I know, what I meant is `boost::get<N>`, the index `N`, I need to look around to see which one is. But this is _not_ a strong argument.

arvidn:
I considered the index range too, but I'm hesitant as I'm trying to use range-for to avoid having to use an index in the first place (to avoid having to specify the type of the index)

aldenml:
I see...LGTM

PR #3413, opened 2018-11-08
bump verson to 1.1.11.0

arvidn:


zeule:
A change [here](https://github.com/arvidn/libtorrent/blob/RC_1_1/CMakeLists.txt#L5) is required too.

PR #3420, opened 2018-11-10
run tests built by cmake on appveyor

arvidn:


arvidn:
todo:
* disable `test_lsd` for cmake
* look into optimizing the tests that take > 50 seconds

PR #3422, opened 2018-11-11
remove w3 compiler flag in cmake build

silverqx:
In cmake is added `/W4` compilation option, so `/W3` and `/W4` are defined at once in CMAKE_CXX_FLAGS, `/W3` is defined by default.

Because of that, every compilation command generates warning:
```
[1/158 0.7/sec] Building CXX object CMakeFiles\torrent-rasterbar.dir\src\entry.cpp.obj
cl : Command line warning D9025 : overriding '/W3' with '/W4'
[2/158 0.7/sec] Building CXX object CMakeFiles\torrent-rasterbar.dir\src\disk_buffer_holder.cpp.obj
cl : Command line warning D9025 : overriding '/W3' with '/W4'
[3/158 1.0/sec] Building CXX object CMakeFiles\torrent-rasterbar.dir\src\chained_buffer.cpp.obj
cl : Command line warning D9025 : overriding '/W3' with '/W4'
[4/158 0.7/sec] Building CXX object CMakeFiles\torrent-rasterbar.dir\src\generate_peer_id.cpp.obj
cl : Command line warning D9025 : overriding '/W3' with '/W4'
[5/158 0.8/sec] Building CXX object CMakeFiles\torrent-rasterbar.dir\src\block_cache.cpp.obj
cl : Command line warning D9025 : overriding '/W3' with '/W4'
[6/158 0.9/sec] Building CXX object CMakeFiles\torrent-rasterbar.dir\src\lazy_bdecode.cpp.obj
cl : Command line warning D9025 : overriding '/W3' with '/W4'
[7/158 1.0/sec] Building CXX object CMakeFiles\torrent-rasterbar.dir\src\choker.cpp.obj
cl : Command line warning D9025 : overriding '/W3' with '/W4'
[8/158 1.1/sec] Building CXX object CMakeFiles\torrent-rasterbar.dir\src\web_connection_base.cpp.obj
cl : Command line warning D9025 : overriding '/W3' with '/W4'
```

This PR removes this warnings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -462,7 +462,9 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES GNU)
 		-Wno-format-zero-length
 		-ftemplate-depth=512)
 elseif(MSVC)
-	add_compile_options(
+    # remove default warning level from CMAKE_CXX_FLAGS
+    string (REGEX REPLACE " +/(W|w)[0-4] +" " " CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+    add_compile_options(
 		/W4
 		# C4251: 'identifier' : class 'type' needs to have dll-interface to be
 		#        used by clients of class 'type2'


zeule:
Why are AppVeyour builds free from this options clash?   https://ci.appveyor.com/project/arvidn/libtorrent/builds/20196738/job/d66s79y0je0bx576

silverqx:
> Why are AppVeyour builds free from this options clash? 

I looked at it more carefully and it is default setting in msvc toolchain, it's default in visual studio, cmake and in qt creator too.
AppVoyer cmake build doesn't have this /W3 setting because is setup by few simple commands on command line.

Of course this setting can be removed from ide gui manually, but I think it is good practice to remove previous `/W[0-4]` flag, when it changes to another value, because of D9025 warning.

arvidn:
please excuse my ignorance about CMake. Are you saying that visual studio (when using CMake) and QT creator invoke `cmake` with a non-empty `CMAKE_CXX_FLAGS` variable?

Maybe it does make more sense for the "upper layer" to set the warning level. I suppose that's something you normally want project wide. Would it make more sense to move the `/W4` command line in libtorrent to only be set for the *top-level* targets? i.e. tests, tools, examples and python bindings.

If the warning level is set for a top level target, will it propagate down to all its dependencies?

zeule:
> I looked at it more carefully and it is default setting in msvc toolchain, it's default in visual studio, cmake and in qt creator too.

Excuse me, could you clarify where does the /W3 option comes from and where it appears? AppVeyour builds use the same MSVC toolchain (what else could it be?). Do you mean that the /W3 option is set by default in the MSVS GUI?

> If the warning level is set for a top level target, will it propagate down to all its dependencies?

And into the clients too, and that is hardly what they want.

arvidn:
is there any convention in cmake on how to handle this? Perhaps something like a higher level property than a raw command line argument added to a variable. It seems like cmake itself should be aware of the warning level and set one, and only one, of these flags.

zeule:
> is there any convention in cmake on how to handle this? 

AFAIK, there is: by default CMAKE_XXX_FLAGS do not contain any warning related options. Another convention for compilers is that the latter option *silently* overrides the former one.

silverqx:
> please excuse my ignorance about CMake. Are you saying that visual studio (when using CMake) and QT creator invoke cmake with a non-empty CMAKE_CXX_FLAGS variable?

Yes, that it's true, it only happens with msvc tool chain ( I only tried 2017 x64 ) and only with GUIs like qt creator, cmake gui.



CMAKE_CXX_FLAGS are empty with mingw or msys tools chain.

I also tried to build with `x64 Native Tools Command Prompt for VS 2017` from cmd and also CMAKE_CXX_FLAGS was empty.

> Excuse me, could you clarify where does the /W3 option comes from and where it appears?

I don't know how this value is pre-populated or how it internally works.

> is there any convention in cmake on how to handle this?

I didn't find better solution for now.

silverqx:
Yesterday I build in `x64 Native Tools Command Prompt for VS 2017` and found `cmake -LA` command, so I can fastly examine cmake variables and discovered, that `CMAKE_XXX_FLAGS ` is also preset with /W3 flag, but any warning during build :/ 
So I propose to close this PR and we will see later

zeule:
Aha! We use the MSVC generator in AppVeyour, and that is why, probably, we do not see these messages.

arvidn:
I've noticed that the msvc generator is *extremely* verbose. Unless it seems like a bad idea, I'm tempted to change to ninja (once I have some time to spare)

arvidn:
I assume the ninja generator can use msvc as the compiler still..

zeule:
We can pass `-verbosity:quiet` to the msbuild instead.

silverqx:
@zeule I think so too

> I assume the ninja generator can use msvc as the compiler still..

of course, ninja is better choice, much clearer build dir

PR #3425, opened 2018-11-11
add rvalue reference overloads for add_torrent_params, settings_pack and session_params

arvidn:
Those structures are pretty large so being able to take them by rvalue reference may be beneficial

### DIFF
--- a/bindings/python/src/session.cpp
+++ b/bindings/python/src/session.cpp
@@ -979,8 +979,8 @@ void bind_session()
 #endif // TORRENT_DISABLE_DHT
         .def("add_torrent", &add_torrent)
         .def("async_add_torrent", &async_add_torrent)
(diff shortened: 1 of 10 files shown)


arvidn:
The idea here is that these objects are quite large, and even though we can move all of their members individually, it's still more efficient to take the whole object by reference. For public facing functions, this patch replaces a single function taking these types by value by two overloads, one for `T&&` and one `T const&`. This let's us take the whole object as a reference in all cases, whether we take ownership of the copy or not. We can then internally create the copy if we have to.

For internal functions, instead I deleted the overload that takes a `T const&`, to make it a compiler error to not move it into the call.

This change was inspired by the LGTM alerts about passing these large objects by value. There are a few more places I would like to give this treatment too.

PR #3429, opened 2018-11-11
require tracker_request objects be moved into queue_manager. also rep…

arvidn:
…lace non-const reference parameter with rvalue reference

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -1035,7 +1035,7 @@ namespace aux {
 		return ret;
 	}
 
-	void session_impl::queue_tracker_request(tracker_request& req
+	void session_impl::queue_tracker_request(tracker_request&& req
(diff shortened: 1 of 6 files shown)


src/session_impl.cpp
```diff
@@ -1082,7 +1082,7 @@ namespace aux {
 				// them consistent and unique per torrent and interface
 				req.key ^= ls->tracker_key;
 				req.outgoing_socket = ls;
-				m_tracker_manager.queue_request(get_io_service(), req, c);
+				m_tracker_manager.queue_request(get_io_service(), std::move(req), c);
```
ssiloti: This is repeatedly moving `req` in a loop, I don't think this is safe.

arvidn: in fact, the logic for setting `key` isn't right either. It isn't supposed to include the keys from the trackers earlier in the loop

PR #3431, opened 2018-11-11
cmake: strip directories and library prefixes for pkg-config

zeule:
Replace full library names with -L commands and -llib-name pairs. This
way it resembles what autotools produce and additionally -l can not be
used with GNU linker and full paths (as it prepends "lib" to such
arguments).


zeule:
Generated .pc file:
```
prefix=/usr
libdir=${prefix}/lib64

Name: libtorrent
Description: Bittorrent library
Version: 1.2.0
Libs: -L${libdir} -ltorrent-rasterbar -lpthread -lssl -lcrypto -lboost_system-mt
Cflags: -fexceptions -std=c++14 -I/usr/include -DTORRENT_LINKING_SHARED -DTORRENT_USE_ASSERTS -DTORRENT_USE_ICONV -DTORRENT_USE_OPENSSL -DTORRENT_USE_LIBCRYPTO
```

PR #3433, opened 2018-11-11
fix some warnings in utp_stream

arvidn:


src/utp_stream.cpp
```diff
@@ -1605,15 +1596,21 @@ void utp_socket_impl::parse_sack(boost::uint16_t packet_ack, boost::uint8_t cons
...
 		experienced_loss(m_fast_resend_seq_nr);
 		int num_resent = 0;
-		while (m_fast_resend_seq_nr != last_ack)
+
+		// only re-sending a single packet per sack
+		// appears to improve performance by making it
+		// less likely to loose the re-sent packet. Because
+		// when that happens, we must time-out in order
+		// to continue, which takes a long time.
+		if (m_fast_resend_seq_nr != last_ack)
```
arvidn: since the limit was 1 anyway, there's no need for a loop and a constant (with a check that's always true)

src/utp_stream.cpp
```diff
@@ -2985,14 +2982,7 @@ bool utp_socket_impl::incoming_packet(boost::uint8_t const* buf, int size
 			return true;
 		}
 		int next_extension = *ptr++;
-		int len = *ptr++;
-		if (len < 0)
```
arvidn: `len` is read from an unsigned 8 bit integer, it's always >= 0

PR #3434, opened 2018-11-11
perfect forward async handlers for udp_socket to underlying asio socket

arvidn:


aldenml:
do you think this can improve the performance a noticeable way?

PR #3444, opened 2018-11-13
exit natpmp::on_reply earlier if we're shutting down

arvidn:


aldenml:
Hi @arvidn, I'm curious if you detected some bad behavior and this fix it. Thanks

arvidn:
I saw a failure in a pull request against master in travis, where it looked like the handle was called after something it referenced had been destructed. Unfortunately I did not record the error log. I think there need to be something like this patch, if not exactly this one. If `m_abort` is set we should not issue more async operations at least.

PR #3449, opened 2018-11-14
cmake: allow selecting C++ standard version during build

zeule:
This backports commit 8e72fad3c to the RC_1_1 branch, taking into
account that C++98 is allowed for RC_1_1 with older boost versions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,9 +8,9 @@ project(libtorrent
 set (SOVERSION "9")
 
 list(APPEND CMAKE_MODULE_PATH ${libtorrent_SOURCE_DIR}/cmake/Modules)
-include(FeatureSummary)
 include(GNUInstallDirs)
 include(GeneratePkgConfig)
+include(LibtorrentMacros)
 
 set(sources
 	web_connection_base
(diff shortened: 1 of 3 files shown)


glassez:
Please describe how it works. How should I select c++ mode: via CXXFLAGS or via appropriate CMake option? What mode will be used by default?

zeule:
You have to pass the desired standard version in [CMAKE_CXX_STANDARD](https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_STANDARD.html) variable:
```
cmake -DCMAKE_CXX_STANDARD=11 ...
```

sledgehammer999:
You also need to include `LibtorrentMacros.cmake` into `Makefile.am` in the `EXTRA_DIST` variable. Otherwise it won't be included in the release tarball.

glassez:
@zeule, what about my second question?
>What mode will be used by default?

If I don't pass CMAKE_CXX_STANDARD...

zeule:
> you also need to include LibtorrentMacros.cmake into Makefile.am

Thanks, corrected.

> If I don't pass CMAKE_CXX_STANDARD...

The compiler default if sufficient, the minimal required otherwise.

arvidn:
@sledgehammer999 do you want to test this patch before I land it?

sledgehammer999:
I have tested it on debian unstable. The compiler default is c++14 which seems to be detected by cmake too.
It also seems to correctly enable C++11 mode if I pass `-DCMAKE_CXX_STANDARD=11`
But at the moment I don't have a gcc version that doesn't default to c++14 in order to test it.

PR #3452, opened 2018-11-14
Fix setting ipv6 interface

airium:
The loop stops once we set an ipv4 interface or an ipv6 interface (depends on the sequence of interfaces), leaving the other interface unset. If it results in empty ipv6 interface, then in `http_tracker_connection.cpp` we will fail to include `&ipv6=` segment to private trackers. That's why people reported failure to announce ipv6 address.

I would also advise to revert some previous changes about ipv6 announcement. Since with this fix, if a tracker is accessible via ipv4 and ipv6 at the same time, currently we will announce twice to it via ipv4 and ipv6 respectively. Then in the ipv6 one, the tracker will fail to obtain the ipv4 address. This is especially unexpected by private trackers.

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -2055,7 +2055,7 @@ retry:
 			// used to send to the tracker
 			std::vector<ip_interface> ifs = enum_net_interfaces(m_io_service, ec);
 			for (std::vector<ip_interface>::const_iterator i = ifs.begin()
-					, end(ifs.end()); i != end && (want_v4 && want_v6); ++i)
+					, end(ifs.end()); i != end && (want_v4 || want_v6); ++i)
 			{
 				address const& addr = i->interface_address;
 				if (want_v4 && addr.is_v4() && !is_local(addr) && !is_loopback(addr))


arvidn:
looks right to me. @ssiloti ?

ssiloti:
I'm not clear on what you want with the refactoring you mention. In general the `ipv6` parameter should not be relied on if it can be avoided because there is no way to verify that it belongs to the announcing peer. If a tracker has both IPv4 and IPv6 addresses then libtorrent absolutely should announce to each address. If a tracker is operating with dual stacks and cannot handle this then that is an issue with the tracker.

airium:
The problem is, before libtorrent 1.1 introducing the second announce on the other interface, using `&ipv6=` in ipv4 report to private trackers has become a conventional way to announce ipv4+ipv6 at the same time. Due to that, on receiving the second announce, private tracker has no idea if you are changing current or adding another ip interface. The practice shows the former will be chosen. As a result, if the second announce is via ipv6, the first ipv4 announce with `&ipv6=` will be overridden and therefore your ipv4 address will no longer be recorded.

Indeed, tracker admins should adapt the new style of dual stack announce, but I also think libtorrent should preserve the compatibility with the old style. There are 2 solutions:

1. Ensure the ipv4 announce is sent **after** the ipv6 announce, so that the ipv4 one with `&ipv6=` will override the first ipv6 one without ipv4 information.
2. Use ipv4 in priority. That is, if the tracker is accessible via ipv4, then just connect via ipv4 and uses `&ipv6=` to report ipv6; otherwise, connect via ipv6.

> In general the ipv6 parameter should not be relied on if it can be avoided because there is no way to verify that it belongs to the announcing peer

Since ipv6 no longer recommends NAT, I think commonly the ipv6 reported by `&ipv6=` is reliable enough (at least compared with ipv4). Before trackers are mostly changed to accept the new style, using ipv6 parameter should be better kept.

arvidn:
since the `&ipv6=` and `&ipv4=` parameters also pose a privacy risk, they were altered a while ago to only be used for private trackers, where they seem to matter.

PR #3458, opened 2018-11-16
Decrease MSBuild verbosity level for AppVeyour cmake builds

zeule:


zeule:
@arvidn, take a look at the output, please. I think it is fine in terms of verbosity. I can add a commit with an error in addition to warnings, if you like.

PR #3461, opened 2018-11-16
Fix redundant bytes overflow

airium:
`uint32_t` = 4GB is too small for redundant bytes if people disable `strict_end_game_mode` on seedboxes. Every time `m_total_redundant_bytes` overflows 4GB,  it is reset to 0, which results in incorrectly reported download bytes.
<https://github.com/arvidn/libtorrent/blob/6381b35891f7fce1c71a798eaae4eb3172f569f4/src/torrent.cpp#L3219-L3222>


arvidn:
thanks. interestingly, the field in `torrent_status` was already a 64 bit integer.

PR #3470, opened 2018-11-18
doozer: download boost using curl on macOS

Kolcha:
each macOS system has `curl` preinstalled, so why not to use it download anything? so I changed doozer's build script to use `curl` instead of `wget` installed from `homebrew`, to download boost.
moreover, I replaced download/extract/remove archive sequence only with one download, i.e. no archive saved, everything extracted on the fly.

used `curl` options are common and well-known:
`-L` - follow redirects
`-s` - be quiet

### DIFF
--- a/.doozer.json
+++ b/.doozer.json
@@ -14,11 +14,8 @@
     },
     "macOS": {
       "buildenv": "osx",
-      "homebrew": { "formulae": ["wget"] },
       "buildcmd": [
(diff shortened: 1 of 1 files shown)


arvidn:
don't you need to pass `z` to `tar` as well? to make it unzip

Kolcha:
for some strange reason unclear to me my command doesn't work on doozer, so I close PR

Kolcha:
I passed `z` for `tar` means `gz` archive format, I used `bz2` archive, so I passed `j`.
I don't understand why doozer build fails...

PR #3473, opened 2018-11-18
port of cmake c++ standard detection from RC_1_1

aldenml:
This fix my environment, when `${CMAKE_CXX_STANDARD_DEFAULT} EQUAL 98` is `true`


arvidn:
`CMAKE_CXX_STANDARD_DEFAULT` is what the compiler uses if nothing is specified?
Is this relevant because we assume this is the version clients will use? How do clients specify which version they use?

arvidn:
ping @zeule

aldenml:
@zeule I wasn't aware of that code in `RC_1_1`, I can do the port to `master`

aldenml:
@arvidn I rebase from `master`, @zeule let me know what do you think of this change

cmake/Modules/LibtorrentMacros.cmake
```diff
@@ -54,20 +54,35 @@ macro(find_public_dependency _name)
...
 			set(_std ${CMAKE_CXX_STANDARD_DEFAULT})
 		else()
 			set(_std ${_minimal_supported_version})
 		endif()
 	else()
-		if (${_std} LESS ${_minimal_supported_version})
-			message(FATAL_ERROR "Sorry, C++${_std} is not supported by libtorrent")
+		_cxx_standard_to_year(std_year ${_std})
+		if (${std_year} LESS ${minimal_supported_version_year})
+			message(SEND_ERROR "Sorry, C++${_std} is not supported by libtorrent with the given Boost version")
```
zeule: AFAIR, the master does not connect boost version and required C++ standard, hence this message would be misleading.

aldenml: true, removing it

PR #3486, opened 2018-11-21
optimize disk buffer allocation/deallocation

arvidn:
by introducing a free-list of recently released buffers, for recycling

### DIFF
--- a/src/disk_io_thread.cpp
+++ b/src/disk_io_thread.cpp
@@ -2591,6 +2591,8 @@ constexpr disk_job_flags_t disk_interface::cache_hit;
 			, completed_jobs, l);
 		l.unlock();
 
+		m_disk_cache.trim_freelist();
+
 		j->storage->release_files(j->error);
 		return j->error ? status_t::fatal_disk_error : status_t::no_error;
 	}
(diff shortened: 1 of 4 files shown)


src/disk_io_thread.cpp
```diff
@@ -3148,6 +3150,13 @@ constexpr disk_job_flags_t disk_interface::cache_hit;
 						m_file_pool.close_oldest();
 					}
 				}
+
+				if (now > m_next_trim)
+				{
+					m_disk_cache.trim_freelist();
+					m_next_trim = now + seconds(10);
```
arvidn: every 10 seconds, free 1/8th of the free list

arvidn:
I'm hesitating to land this as it appears the problem of slow releasing of the buffer went away when transitioning from `_aligned_free()` to `free()`. However, at very high upload rates I did observe a noticeable amount of time (about 8-9%) spent in the allocator (specifically using `tcmalloc` and its CPU profiler).

ssiloti:
I agree this is unnecessary now that the buffer freeing performance has been resolved. I think it would be neat to recycle evicted buffers directly into the job which caused the eviction when possible, but even that might not be justified.

Issue #3467, opened 2018-11-18
Slow disk cache flushing on Windows


sandersaares:
libtorrent version (or branch): libtorrent-rasterbar-1.1.10 built according to instructions in [qBittorrent wiki](https://github.com/qbittorrent/qBittorrent/wiki/Compiling-with-MSVC-2017-x64-%28static-linkage%29)

platform/architecture: win64

compiler and compiler version: VS2017 Visual C++

When using a large disk cache (e.g. 3-4 GB) and pausing torrents, libtorrent spends a long time calling `_aligned_free` on all the allocated buffers. This causes undesirable "hangs" in some clients such as qBittorrent during the period it is "flushing cache" (really just deallocating buffers). In my case, it can be up to 10 minutes of just spinning the CPU.

Here is a screenshot where I let the RAM usage build up and then pause all torrents (indicated by RAM downslope). There is only a single thread active during the flushing interval, doing some `_aliged_free()` calls:

![procexp64_2018-11-18_11-36-50](https://user-images.githubusercontent.com/9914262/48670776-56106600-eb26-11e8-99f0-d0b7ea524a8e.png)

The above shows a quite fast decline (relatively speaking). Sometimes it is much slower:

![image](https://user-images.githubusercontent.com/9914262/43340298-8a8dbd88-91e4-11e8-8347-64df4ba984f4.png)

Seems to correlate with uptime - the longer I run the client the slower it gets (RAM fragmentation effects?).

Stack trace of the active thread:

```
ntoskrnl.exe!KeSynchronizeExecution+0x5be6
ntoskrnl.exe!KeWaitForSingleObject+0x12e6
ntoskrnl.exe!KeWaitForSingleObject+0xadb
ntoskrnl.exe!KeWaitForSingleObject+0x1ff
ntoskrnl.exe!KeCheckProcessorGroupAffinity+0x98a
ntoskrnl.exe!ExIsResourceAcquiredSharedLite+0x7cb
ntoskrnl.exe!KeSynchronizeExecution+0x2dec
ntdll.dll!RtlFreeAnsiString+0x1f2d
ntdll.dll!RtlFreeAnsiString+0x1b7c
ntdll.dll!RtlFreeAnsiString+0x1495
ntdll.dll!RtlReAllocateHeap+0x21d5
ntdll.dll!RtlFreeHeap+0x3fc
qbittorrent.exe!_free_base+0x1c
qbittorrent.exe!_aligned_free+0x16
[...]

PR #3488, opened 2018-11-23
don't use page aligned disk buffers

arvidn:
there isn't anything requiring these buffers to be aligned, and it is suspected the aligned allocation functions on windows are slow.


arvidn:
@ssiloti This seems like a good idea. Can you think of any reasons not do land this?
https://github.com/arvidn/libtorrent/issues/3467#issuecomment-441171259

PR #3490, opened 2018-11-23
fix Jamfile for linking against openssl 1.1+ on windows

arvidn:


Jamfile
```diff
@@ -568,41 +568,19 @@ lib shell32 : : <name>shell32 ;
...
+lib crypto : ssl-deps : <target-os>windows <openssl-version>pre1.1 <name>libeay32 <conditional>@openssl-lib-path : : <conditional>@openssl-include-path ;
+lib ssl : ssl-deps : <target-os>windows <openssl-version>pre1.1 <name>ssleay32 <use>crypto <conditional>@openssl-lib-path : : <conditional>@openssl-include-path ;
 
+# OpenSSL 1.1+ windows
+lib crypto : ssl-deps : <target-os>windows <name>libcrypto <conditional>@openssl-lib-path : : <conditional>@openssl-include-path ;
+lib ssl : ssl-deps : <target-os>windows <name>libssl <use>crypto <conditional>@openssl-lib-path : : <conditional>@openssl-include-path ;
 
-lib crypto : : <name>crypto <use>z <conditional>@openssl-lib-path : : <conditional>@openssl-include-path ;
+# generic OpenSSL
+lib crypto : dl : <name>crypto <use>z <conditional>@openssl-lib-path : : <conditional>@openssl-include-path ;
```
aldenml: is `dl` necessary here?

arvidn: well, good question. it was on my linux system, but I suppose that's a bit linux specific

Jamfile
```diff
@@ -568,40 +568,18 @@ lib shell32 : : <name>shell32 ;
...
-	: # usage-requirments
-	<conditional>@openssl-include-path
-	<library>advapi32
-	<library>user32
-	<library>shell32
-	<library>gdi32
-	;
+# pre OpenSSL 1.1
+lib crypto : ssl-deps : <target-os>windows <openssl-version>pre1.1 <name>libeay32 <conditional>@openssl-lib-path : : <conditional>@openssl-include-path ;
+lib ssl : ssl-deps : <target-os>windows <openssl-version>pre1.1 <name>ssleay32 <use>crypto <conditional>@openssl-lib-path : : <conditional>@openssl-include-path ;
```
cas--: Could we split the lines as these are very long :)

cas--:
Yes this works

PR #3502, opened 2018-11-26
convert some for-loops into std algorithms

arvidn:


src/kademlia/routing_table.cpp
```diff
@@ -177,10 +177,8 @@ std::tuple<int, int, int> routing_table::size() const
 	for (auto const& i : m_buckets)
 	{
 		nodes += int(i.live_nodes.size());
-		for (auto const& k : i.live_nodes)
-		{
-			if (k.confirmed()) ++confirmed;
-		}
+		confirmed += std::count_if(i.live_nodes.begin(), i.live_nodes.end()
```
aldenml: c++ syntax is not helping here, I hope that with `ranges` it could be improved.

arvidn: yeah, there were a few places where using an algorithm would be too much of an obfuscation that I chose not to convert it. I think this is reasonable though

src/session_impl.cpp
```diff
@@ -5068,17 +5062,12 @@ namespace aux {
...
 	{
-		for (auto const& s : m_listen_sockets)
-		{
-			if (s->local_endpoint.address() == addr)
-			{
-				return s->incoming == duplex::accept_incoming;
-			}
-		}
-		return false;
+		return std::any_of(m_listen_sockets.begin(), m_listen_sockets.end()
```
arvidn: this should be a `find_if`

src/torrent.cpp
```diff
@@ -2565,9 +2562,8 @@ bool is_downloading_state(int const st)
 
 				if (settings().get_bool(settings_pack::use_dht_as_fallback))
 				{
-					int verified_trackers = 0;
-					for (auto const& t : m_trackers)
-						if (t.verified) ++verified_trackers;
+					int const verified_trackers = std::accumulate(m_trackers.begin(), m_trackers.end(), 0
```
arvidn: this should be a `count_if`

src/session_impl.cpp
aldenml: why?

arvidn: the `any_of` logic will keep searching, even after if finds `addr`, but the previous loop would terminate immediately after finding the interface with the given address

aldenml: hmm...I find it strange, since I would expect `any_of` to be implemented in terms of `find_if` and avoid the iterator check, but OK (maybe that's not specified).

aldenml: Sorry, disregard my previous comment, now I understand

arvidn: right, but there are two conditions here, and in my first patch (which I just replaced with a force-push) I baked both tests together into one. The current behavior use only one condition for the search, and then the second condition for the return value

PR #3503, opened 2018-11-26
fix some cppcheck warnings

arvidn:


src/peer_connection.cpp
```diff
@@ -2308,8 +2308,8 @@ namespace libtorrent {
 		{
 			if (e->on_request(r)) return;
 		}
-#endif
 		if (is_disconnecting()) return;
+#endif
```
aldenml: Hi @arvidn, sorry to bother you, but do you mind a brief explanation here? I can't identify the potential warning here

arvidn: moving the `is_disconnecting()` check inside the `#ifdef` is because when that code is disabled, there will be two identical checks back-to-back, and the second one is clearly redundant. It only makes sense to have the second check when the plugins have a chance to initiate a disconnect

aldenml: Thank you! I can see it now and it's clear!

PR #3510, opened 2018-11-30
using toolset instead of target-os to select openssl libraries when using openssl >= 1.1

aldenml:
Hi @cas-- , do you mind looking at this? Thanks

### DIFF
--- a/Jamfile
+++ b/Jamfile
@@ -557,9 +557,9 @@ lib ssl : ssl-deps : <target-os>windows <openssl-version>pre1.1 <name>ssleay32
 	<use>crypto <conditional>@openssl-lib-path : : <conditional>@openssl-include-path ;
 
 # OpenSSL 1.1+ windows
-lib crypto : ssl-deps : <target-os>windows <name>libcrypto
+lib crypto : ssl-deps : <toolset>msvc <openssl-version>1.1 <name>libcrypto
 	<conditional>@openssl-lib-path : : <conditional>@openssl-include-path ;
-lib ssl : ssl-deps : <target-os>windows <name>libssl <use>crypto
+lib ssl : ssl-deps : <toolset>msvc <openssl-version>1.1 <name>libssl <use>crypto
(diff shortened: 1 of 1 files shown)


arvidn:
the mingw build seems to have a problem with this.

arvidn:
I wonder if this is not a function of the toolset but rather whether you're linking against an msvc-build of openssl or a msys-build of openssl.

arvidn:
I believe I made this change in RC_1_1 too, but I can back-port it later

cas--:
I don't think it should be replacing target for toolset for pre-1.1. 

This needs appveyor to test with OpenSSL 1.1, especially since that is now the recommended release.

arvidn:
> I don't think it should be replacing target for toolset for pre-1.1.

I think it depends on whether one links against an "msvc-build" or an msys-build. I would expect the msys builds to always have been named "libcrypto" and "libssl", even before 1.1.

cas--:
I mentioned it because the pre-1.1 with target was working fine before the libcrypto name changes.

The change for 1.1 from target to toolset is supported by the change here: https://github.com/radiator-software/p5-net-ssleay/pull/29/files

aldenml:
I also don't think it's wise to mix a `libtorrent` msys build with an `openssl` msvc build, I will take a look at this later

aldenml:
@arvidn I know this fix is the absolute minimum to get it work for me, but making appveyor to work with openssl-1.1+ is a big time consuming task.

@cas-- do you see any problem with this minimal change for now?

aldenml:
any idea of what happened here? https://ci.appveyor.com/project/arvidn/libtorrent/builds/20690878/job/4y2n46fl23ctpexa

aldenml:
it looks like `bjam` find that both conditions apply and can't select one...interesting...

cas--:
lgtm :+1:

aldenml:
sure, done

PR #3514, opened 2018-12-02
 Change Jamfile default openssl path on Windows 

cas--:
The widely used Windows OpenSSL package[1] uses C:\OpenSSL-Win[32|64] as
the install location so set this as default in Jamfile for Windows. It
also means that the appveyor builds can be simplified by using their
installed OpenSSL.

Mingw is set to the same location so remove seperate specification.

[1]: https://slproweb.com/products/Win32OpenSSL.html

Also includes a tidy-up commit of appveyor config

### DIFF
--- a/Jamfile
+++ b/Jamfile
@@ -359,20 +359,13 @@ rule openssl-lib-path ( properties * )
 		# brew install openssl
 		OPENSSL_LIB = /usr/local/opt/openssl/lib ;
 	}
-	else if <target-os>windows in $(properties)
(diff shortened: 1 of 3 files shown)


cas--:
Just remembered I need to update docs for this change. Did you also want a changelog entry?

arvidn:
nah, I don't think this needs a changelog entry

arvidn:
were you planning to update this patch with documentation updates?

PR #3522, opened 2018-12-05
removed repeated code and minor code refactor

aldenml:


src/http_tracker_connection.cpp
```diff
@@ -327,12 +327,6 @@ namespace libtorrent {
 			return;
 		}
 
-		if (ec && ec != boost::asio::error::eof)
```
aldenml: this code is the same as in L311

src/tracker_manager.cpp
```diff
@@ -292,9 +292,9 @@ namespace libtorrent {
 		}
 
 		// we need to post the error to avoid deadlock
-		if (std::shared_ptr<request_callback> r = c.lock())
+		if (auto r = c.lock())
 			ios.post(std::bind(&request_callback::tracker_request_error, r, std::move(req)
-				, error_code(errors::unsupported_url_protocol)
+				, errors::unsupported_url_protocol
```
aldenml: @arvidn do you have a preference about the use of the implicit constructor of `error_code`?

arvidn: I like the implicit constructor

PR #3524, opened 2018-12-06
Python Jamfile

arvidn:
attempt to parse out the version of boost when building (boost) from source. This can then be used to know whether to attempt to link against the `boost_python3` or `boost_python` target (which changed in boost-1.68.0).

### DIFF
--- a/bindings/python/Jamfile
+++ b/bindings/python/Jamfile
@@ -31,16 +31,16 @@ if $(LIBTORRENT_PYTHON_INTERPRETER)
 # copied from boost 1.63's boost python jamfile
 rule find-py3-version
 {
-    local versions = [ feature.values python ] ;
-    local py3ver ;
-    for local v in $(versions)
-    {
-        if $(v) >= 3.0
-        {
-            py3ver = $(v) ;
-        }
-    }
-    return $(py3ver) ;
+	local versions = [ feature.values python ] ;
+	local py3ver ;
+	for local v in $(versions)
+	{
+		if $(v) >= 3.0
+		{
+			py3ver = $(v) ;
+		}
+	}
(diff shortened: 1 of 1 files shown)


arvidn:
sorry about the first commit. I couldn't stand the mixed tabs and spaces. To review the patch, either just look at the second commit, or enabled "ignore white space changes" in github

bindings/python/Jamfile
```diff
@@ -31,16 +33,26 @@ if $(LIBTORRENT_PYTHON_INTERPRETER)
...
-    for local v in $(versions)
-    {
-        if $(v) >= 3.0
-        {
-            py3ver = $(v) ;
-        }
-    }
-    return $(py3ver) ;
+	local py3ver ;
+	# expect the path to be something like boost_1_68_0
```
cas--: Perhaps avoid using path, it seems fragile and a workflow I have used renamed boost directory

What about parsing version.cpp?

arvidn: yeah, that was my next thought

cas--: Can you `import boost` and get the version?

arvidn:
how about this? does it work?

bindings/python/Jamfile
```diff
@@ -5,6 +5,8 @@ import project ;
 import targets ;
 import "class" : new ;
 import modules ;
+import path ;
+import regex ;
```
arvidn: I should revert this too

ssiloti:
Wait, no, this isn't going to work as-is. The key part though is the boost version check, I could never figure out the right boost build incantation to do that. I'll push a commit to fix up this PR.

PR #3531, opened 2018-12-09
added more TORRENT_DEPRECATED_ENUM and deprecated unused aio_max in settings_pack

aldenml:


include/libtorrent/settings_pack.hpp
```diff
@@ -1697,7 +1700,7 @@ namespace libtorrent {
 #if TORRENT_ABI_VERSION == 1
 			disable_os_cache_for_aligned_files TORRENT_DEPRECATED_ENUM = 2,
```
aldenml: @arvidn there is something odd with this number `2`, the deprecated value is `1`, is that right that mismatch?

arvidn: well, yeah, it looks a bit odd. I don't think it matters though, as you won't use the `deprecated_` enum, and the next enumeration has its value explicitly assigned.

PR #3540, opened 2018-12-13
Add magnet uri support for tr.x tracker

### LINKED ISSUE #1302: Add magnet uri support for tr.x tracker
Do you think libtorrent could add support for tracker tiers included in some magnets. I am in the middle of adding support for it to our code but realised it would be useful for libtorrent to also have it.

The specs are outlined in this Deluge ticket: http://dev.deluge-torrent.org/ticket/2716

I am not sure whether creating magnet uris should contain the `tr.x=` param or simply stick with `tr=` as default. Alternatively only use `tr.x=` if trackers share a tier...

DjLegolas:
This will close #1302.

As for adding tests, I saw in `test_magnet.cpp` the tests that contains the `tr=` part and edited one of them to check my changes.
The qustion is, how do you want to test this? with its own test or add to the current tests or even run the tests for both `tr` and `tr.x`?

### DIFF
--- a/src/magnet_uri.cpp
+++ b/src/magnet_uri.cpp
@@ -190,7 +190,8 @@ namespace libtorrent {
 				error_code e;
 				display_name = unescape_string(value, e);
 			}
-			else if (name == "tr"_sv) // tracker
+			else if (name == "tr"_sv ||
+			    (name.substr(0, 3) == "tr."_sv && is_numeric(name.substr(3)))) // tracker
 			{
 				// since we're about to assign tiers to the trackers, make sure the two
 				// vectors are aligned
--- a/src/string_util.cpp
+++ b/src/string_util.cpp
@@ -387,6 +387,13 @@ namespace libtorrent {
 		return {last.substr(0, pos), last.substr(pos + found_sep)};
 	}
 
+	bool is_numeric(string_view s)
+	{
+		if (s.empty()) return false;
+
+        return std::all_of(s.begin(), s.end(), [](char c) { return is_digit(c); } );
+	}
+
 #if TORRENT_USE_I2P
 
 	bool is_i2p_url(std::string const& url)
--- a/include/libtorrent/string_util.hpp
+++ b/include/libtorrent/string_util.hpp
@@ -117,6 +117,9 @@ namespace libtorrent {
 	// return value is an empty string_view.
 	TORRENT_EXTRA_EXPORT std::pair<string_view, string_view> split_string(string_view last, char sep);
 
+    // checks whether the string_view is a numeric or not
+	TORRENT_EXTRA_EXPORT bool is_numeric(string_view s);
+
 #if TORRENT_USE_I2P
 
 	TORRENT_EXTRA_EXPORT bool is_i2p_url(std::string const& url);
--- a/test/test_magnet.cpp
+++ b/test/test_magnet.cpp
@@ -227,6 +227,31 @@ TORRENT_TEST(magnet)
 	p2 = s->abort();
 }
 
+TORRENT_TEST(magnet_tr_x_uri)
+{
+
+    add_torrent_params p = parse_magnet_uri("magnet:"
+        "?tr.0=udp://1"
+        "&tr.1=http://2"
+        "&tr=http://3"
+        "&xt=urn:btih:c352cdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcd");
+    TEST_EQUAL(p.trackers.size(), 3);
+    TEST_EQUAL(p.trackers[0], "udp://1");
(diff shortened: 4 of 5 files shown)


src/string_util.cpp
```diff
@@ -387,6 +387,16 @@ namespace libtorrent {
 		return {last.substr(0, pos), last.substr(pos + found_sep)};
 	}
 
+	bool is_numeric(string_view s)
```
arvidn: this function should have tests, in `test/test_string.cpp`

src/magnet_uri.cpp
```diff
@@ -190,7 +190,8 @@ namespace libtorrent {
 				error_code e;
 				display_name = unescape_string(value, e);
 			}
-			else if (name == "tr"_sv) // tracker
+			else if (name == "tr"_sv ||
+			    (name.substr(0, 3) == "tr."_sv && is_numeric(name.substr(3, name.size())))) // tracker
```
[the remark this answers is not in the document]
arvidn: surely the number following the `tr.` name has some significance, no? Does it not specify which tier a tracker belongs to? If it doesn't, what's the point?

Can you point to any documentation of what this is supposed to mean?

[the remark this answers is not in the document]
arvidn: This would need a test as well, please add another `TORRENT_TEST()` to `test/test_magnet.cpp` to ensure that trackers are parsed out of magnets using this syntax correctly.

There is a large test case in there that really should be split up into smaller ones. Don't use that as inspiration, write a small test that just tests all aspects of this one feature.

DjLegolas:
As there is no documentation of any sort for this feature, I can't really say.
I don't think it comes to replace `tr=`, but to extend it.
Honestly, I didn't find any site that uses this tag, but if it **IS** being used somewhere, I think `libtorrent` should support it.

arvidn:
I would love to hear more arguments in favour of this btw, If anyone else want to chime in. There may very well be a really good reason to add this number to this tag, I just can't imagine what it is.

cas--:
@arvidn I echo and understand your concerns about long-term support so I shall address some of the concerns.

> I don't think it comes to replace `&tr=`, but to extend it

Yes it is definitely to extend not to replace it. This would be utilising [magnet specification group settings](https://en.wikipedia.org/wiki/Magnet_URI_scheme#Group_settings)

```
Multiple files and their URNs, names and hashes in the Magnet link can be included 
by adding a count number preceded by a dot (".") to each link parameter.

    magnet:?xt.1=[ URN of the first file]&xt.2=[ URN of the second file]
```

To me using multiple `tr=` in a magnet URI seems out of spec, i.e. there should only be one and for multiple we use `tr.1=, tr.2=` etc.

> I didn't find any site that uses this tag

There are magnets in the wild with `tr.n` options so I felt we do need to support it. A google search for `btih  "tr.0"` or `btih  "tr.1"` returns results. This was the reason for the [feature request](https://dev.deluge-torrent.org/ticket/2716) in the first place by a Deluge user.

---

After doing all this research again, a sweeping assumption can be made that the order of the trackers in the magnet URI is the tier order so multiple `tr` options would be placed in separate tiers, starting at 0. This would also apply to `tr.n` (dot with single digit) group setting, with code that ignores the actual number and just places trackers in separate tiers in the order they are parsed. 

There is very little code in this PR and we could reduce the overhead further. e.g. do without the new `is_numeric` function when `is_digit` would suffice.

A separate issue that does need addressed is with Deluge and libtorrent parsing magnet trackers into tier 0 (I should double-check this) so that magnets with multiple trackers are not being tiered as they are with torrent files. e.g. [SuperUser question](https://superuser.com/questions/1271085/how-to-specify-multiple-tiered-trackers-in-magnet-uri).

src/magnet_uri.cpp
```diff
@@ -190,7 +190,8 @@ namespace libtorrent {
 				error_code e;
 				display_name = unescape_string(value, e);
 			}
-			else if (name == "tr"_sv) // tracker
+			else if (name == "tr"_sv ||
+			    (name.substr(0, 3) == "tr."_sv && is_numeric(name.substr(3)))) // tracker
```
cas--: Lets simplify by only checking for a single digit and remove the need for `is_numeric`

cas--: Since we don't care or use the value following the dot should we even check if it's numeric?

arvidn:
@cas-- libtorrent-1.2 assigns incrementing tiers to trackers in magnet links, in the order the `&tr=` parameters appear

arvidn:
does this look right? https://github.com/arvidn/libtorrent/pull/3647
It's kind of messy to do this in 1.1

cas--:
That seems good to me :+1:

arvidn:
closed in favour of https://github.com/arvidn/libtorrent/pull/3647

PR #3543, opened 2018-12-20
don't cut cwnd multiple times per congestion window

arvidn:
and adjust the swnd multiplier to be in line with uTorrent


src/utp_stream.cpp
```diff
@@ -2342,7 +2342,8 @@ void utp_socket_impl::experienced_loss(int const seq_nr)
 	// cut window size in 2
 	m_cwnd = std::max(m_cwnd * m_sm->loss_multiplier() / 100
 		, boost::int64_t(m_mtu) * (1 << 16));
-	m_loss_seq_nr = m_seq_nr;
+	int const packets_in_flight = (m_seq_nr - m_acked_seq_nr) & ACK_MASK;
+	m_loss_seq_nr = m_seq_nr + packets_in_flight;
```
arvidn: this is the relevant change

arvidn:
I'm splitting this up into smaller PRs, and dropping the 100ms cwnd drop limit patch.

PR #3552, opened 2018-12-31
remove old (incompatible) sequence number build option

arvidn:
Issue MTU probes with less size margin

### DIFF
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -44,10 +44,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <boost/cstdint.hpp>
 #include <limits>
 
-// the behavior of the sequence numbers as implemented by uTorrent is not
-// particularly regular. This switch indicates the odd parts.
-#define TORRENT_UT_SEQ 1
-
 #if TORRENT_UTP_LOG
 #include <stdarg.h>
 #include "libtorrent/socket_io.hpp"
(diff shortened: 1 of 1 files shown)


src/utp_stream.cpp
```diff
@@ -2850,7 +2839,9 @@ bool utp_socket_impl::incoming_packet(boost::uint8_t const* buf, int size
 
 	// if the socket is closing, always ignore any packet
 	// with a higher sequence number than the FIN sequence number
-	if (m_eof && compare_less_wrap(m_eof_seq_nr, ph->seq_nr, ACK_MASK))
+	// ST_STATE messages always include the next seqnr.
+	if (m_eof && (compare_less_wrap(m_eof_seq_nr, ph->seq_nr, ACK_MASK)
+		|| (m_eof_seq_nr == ph->seq_nr && ph->get_type() != ST_STATE)))
```
arvidn: it's a bit awkward, but `ST_STATE` messages use the "next" sequence number. So, if the packet you receive is `ST_STATE`, the sequence number indicates what will be used in the next `ST_DATA` message. So, when shutting down, it's OK to receive more messages with the same sequence number as the `FIN`, as long as they are `ST_STATE` messages

src/utp_stream.cpp
```diff
@@ -1795,10 +1791,11 @@ bool utp_socket_impl::send_pkt(int const flags)
 	// for the cwnd condition is to make sure the probe is surrounded by non-
 	// probes, to be able to distinguish a loss of the probe vs. just loss in
 	// general.
-	bool const mtu_probe = (m_mtu_seq == 0
-		&& m_write_buffer_size >= m_mtu_floor * 3
+	bool const mtu_probe = m_mtu_floor < m_mtu_ceiling
```
arvidn: when `m_mtu_floor` == `m_mtu_ceiling`, we're done, and we don't need to issue any more probes

PR #3560, opened 2019-01-05
make sure we reset the duplicate ack counter every time we don't receive a duplicate ack

arvidn:
Also fix logging of outstanding packets


arvidn:
basically, once we don't have any outstanding packets, the duplicate ack-counter should be reset, as we expect to just receive acks for that last packet repeatedly at that point.

PR #3566, opened 2019-01-06
removed unused file::sparse_end

aldenml:
again, I couldn't find any use of this


arvidn:
hm. This used to be used to speed up checking of files. Specifically, sparse files. This optimization must have been dropped in some refactoring. Perhaps it's not worth it.

PR #3569, opened 2019-01-07
Fix Boost handling with CMake

evpobr:
* Add Boost dependencies as imported libraries. This properly adds Boost
includes, libraries and also definitions to target
* Add forgotten Boost::date_time library

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,28 @@ include(GNUInstallDirs)
 include(GeneratePkgConfig)
 include(CMakeDependentOption)
 
-# Boost
+# Options
+
+# these options control target creation and thus have to be declared before the add_library() call
+feature_option(BUILD_SHARED_LIBS "build libtorrent as a shared library" ON)
+feature_option(static_runtime "build libtorrent with static runtime" OFF)
+
+# Configuration
+
+if(static_runtime)
+	include(ucm_flags)
+	ucm_set_runtime(STATIC)
+	set(Boost_USE_MULTITHREADED ON)
+	set(Boost_USE_STATIC_RUNTIME ON)
+	set(OPENSSL_USE_STATIC_LIBS TRUE)
+	set(OPENSSL_MSVC_STATIC_RT TRUE)
+endif()
+
+if (NOT BUILD_SHARED_LIBS)
+	set(Boost_USE_STATIC_LIBS ON)
+endif()
+
+# Packages
 find_public_dependency(Boost REQUIRED COMPONENTS system)
 find_public_dependency(Boost COMPONENTS date_time)
 
@@ -441,10 +462,6 @@ list(TRANSFORM libtorrent_extensions_include_files PREPEND "include/libtorrent/e
 list(TRANSFORM libtorrent_aux_include_files PREPEND "include/libtorrent/aux_/")
 list(TRANSFORM libtorrent_kademlia_include_files PREPEND "include/libtorrent/kademlia/")
 
-# these options control target creation and thus have to be declared before the add_library() call
-feature_option(BUILD_SHARED_LIBS "build libtorrent as a shared library" ON)
-feature_option(static_runtime "build libtorrent with static runtime" OFF)
-
 find_public_dependency(Threads REQUIRED)
 
 if(CMAKE_CXX_COMPILER_ID MATCHES Clang)
@@ -483,19 +500,6 @@ elseif(MSVC)
 		/wd4268
 		# C4503: 'identifier': decorated name length exceeded, name was truncated
 		/wd4503)
-endif()
-
-if(static_runtime)
-	include(ucm_flags)
-	ucm_set_runtime(STATIC)
-	set(Boost_USE_MULTITHREADED ON)
-	set(Boost_USE_STATIC_RUNTIME ON)
-	set(OPENSSL_USE_STATIC_LIBS TRUE)
-	set(OPENSSL_MSVC_STATIC_RT TRUE)
-endif()
-
-if (NOT BUILD_SHARED_LIBS)
-	set(Boost_USE_STATIC_LIBS ON)
 endif()
 
 add_library(torrent-rasterbar
@@ -541,6 +545,8 @@ target_compile_definitions(torrent-rasterbar
 target_link_libraries(torrent-rasterbar
 	PUBLIC
 		Threads::Threads
+	PRIVATE
+		Boost::system
 )
 
 # Unconditional platform-specific settings
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,11 @@ set (SOVERSION "10")
 
 include(GNUInstallDirs)
 include(GeneratePkgConfig)
+include(CMakeDependentOption)
+
+# Boost
+find_public_dependency(Boost REQUIRED COMPONENTS system)
+find_public_dependency(Boost REQUIRED COMPONENTS date_time)
 
 set(libtorrent_include_files
 	add_torrent_params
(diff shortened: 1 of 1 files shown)


arvidn:
`posix_time_duration` is necessary when building with deprecated functions enabled. Perhaps this dependency should be conditional too. But perhaps that would add too much complexity.

ssiloti:
I think it's worth it to make the date_time dep condition on the deprecated functions option. Requiring libraries which aren't actually required is obnoxious if you're trying to do a space optimized build (e.g. for a consumer router).

CMakeLists.txt
```diff
@@ -678,8 +678,12 @@ else()
 endif()
 
 # Boost
-find_public_dependency(Boost REQUIRED COMPONENTS system date_time)
-target_link_libraries(torrent-rasterbar PUBLIC Boost::system Boost::date_time)
+find_public_dependency(Boost REQUIRED COMPONENTS system)
+target_link_libraries(torrent-rasterbar PUBLIC Boost::system)
+if (TORRENT_NO_DEPRECATE)
```
arvidn: This condition looks inverted. Also, where does this definition come from?
A brief look at this file, I can't see an option to disable deprecated functions in cmake.

arvidn: oh, I see. It's this one `deprecated-functions`. I was thrown off by the conditional around that option. I don't think the comment (or logic) there is right. the tests *can* be built without deprecated functions

evpobr: Can you tell me what exactly i need to put in `if ()`? I see no useful options to start from.

evpobr: The name is really confusing. I understand `NO_DEPRECATE` as "do not deprecate, allow old ABI".

Probably good name is `NO_DEPRECATED`.

arvidn:
@zeule does this look right to you?

zeule:
> I don't use macros from `LibtorrentMacros`, they make me angry.

Could you explain why, please? With your choice the option identifier and the description string are duplicated for `option()` and `add_feature_info()` calls.

CMakeLists.txt
```diff
@@ -13,6 +13,11 @@ set (SOVERSION "10")
 
 include(GNUInstallDirs)
 include(GeneratePkgConfig)
+include(CMakeDependentOption)
+
+# Boost
+find_public_dependency(Boost REQUIRED COMPONENTS system)
```
zeule: Why don't you modify https://github.com/arvidn/libtorrent/blob/master/CMakeLists.txt#L681?

zeule: Never mind, pardon.

CMakeLists.txt
```diff
@@ -13,6 +13,11 @@ set (SOVERSION "10")
...
+find_public_dependency(Boost REQUIRED COMPONENTS date_time)
```
zeule: The `REQUIRED` keyword has to be removed.

zeule: `date_time` is not an unconditional public dependency, but it is only when `deprecated-functions` is set (according to your code below).

CMakeLists.txt
```diff
@@ -677,11 +685,6 @@ else()
 	target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_DISABLE_DHT)
 endif()
 
-# Boost
-find_public_dependency(Boost REQUIRED COMPONENTS system)
-target_include_directories(torrent-rasterbar PUBLIC ${Boost_INCLUDE_DIRS})
```
zeule: This and the next line were here for a reason: every new boost release breaks imported targets in `FindBoost.cmake`. Please revert this change.

zeule:
@evpobr: Take a look at the Travis build log. The build fails.

evpobr:
> @evpobr: Take a look at the Travis build log. The build fails.

Failed to link Boost::system.

zeule:
> Failed to link Boost::system.

target_link_libraries(torrent-rasterbar PUBLIC ${Boost_SYSTEM_LIBRARY})

evpobr:
> > Failed to link Boost::system.
> 
> target_link_libraries(torrent-rasterbar PUBLIC ${Boost_SYSTEM_LIBRARY})

Actually i forgot to `add target_link_libraries(torrent-rasterbar PUBLIC Boost::system)`.

evpobr:
`Boost::date_time` is linked as you can see as target. No problems.

zeule:
> `Boost::system`

I've told you already why this is a bad idea.

evpobr:
> > `Boost::system`
> 
> I've told you already why this is a bad idea.

Explain please. If you're right i will fix my code.

zeule:
Please: every time boost version increases, cmake needs an update in FindBoost.cmake. Until the update imported targets do not exist for the new boost version. Now suppose your changes go public, and one day boost gets an update. Then the cmake build is broken until the next cmake version bump.

zeule:
Example of the required cmake change: https://gitlab.kitware.com/cmake/cmake/commit/364fb69adc8ef0dfe59e5209735561d451c2ec9d

evpobr:
> Please: every time boost version increases, cmake needs an update in FindBoost.cmake. Until the update imported targets do not exist for the new boost version. Now suppose your changes go public, and one day boost gets an update. Then the cmake build is broken until the next cmake version bump.

Is it? I thought `FindBoost` module parses `version.hpp`:

```cmake
# ------------------------------------------------------------------------
#  Extract version information from version.hpp
# ------------------------------------------------------------------------

# Set Boost_FOUND based only on header location and version.
# It will be updated below for component libraries.
if(Boost_INCLUDE_DIR)
  if(Boost_DEBUG)
    message(STATUS "[ ${CMAKE_CURRENT_LIST_FILE}:${CMAKE_CURRENT_LIST_LINE} ] "
                   "location of version.hpp: ${Boost_INCLUDE_DIR}/boost/version.hpp")
  endif()
...
```

evpobr:
Well, fails again. What happens, GitHub problems? They activate private repos.

CMakeLists.txt
```diff
@@ -536,6 +545,7 @@ target_compile_definitions(torrent-rasterbar
 target_link_libraries(torrent-rasterbar
 	PUBLIC
 		Threads::Threads
+		Boost::system
```
zeule: ${Boost_SYSTEM_LIBRARY}

CMakeLists.txt
```diff
@@ -577,17 +587,20 @@ if (APPLE)
...
 # these options require existing target
 feature_option(dht "enable support for Mainline DHT" ON)
-if(NOT build_tests) # tests require deprecated symbols
-	target_optional_compile_definitions(torrent-rasterbar PUBLIC FEATURE NAME deprecated-functions DEFAULT ON
-		DESCRIPTION "enable deprecated functions for backwards compatibility" DISABLED TORRENT_NO_DEPRECATE)
-endif()
+cmake_dependent_option(deprecated-functions "enable deprecated functions for backwards compatibility" ON "Boost_DATE_TIME_FOUND" OFF)
+add_feature_info(deprecated-functions deprecated-functions "enables deprecated functions for backwards compatibility")
+target_compile_definitions(torrent-rasterbar PUBLIC $<$<NOT:$<BOOL:${deprecated-functions}>>:TORRENT_NO_DEPRECATE>)
+target_link_libraries(torrent-rasterbar PUBLIC $<$<BOOL:${deprecated-functions}>:Boost::date_time>)
```
zeule: ${Boost_DATE_TIME_LIBRARY}

arvidn:
@evpobr what was the conclusion?

evpobr:
No sense to develop this PR further with changes @zeule requested.

Issue #3571, opened 2019-01-08
Local IPv4/6 is reported to private trackers


airium:
This is a bug further exposed after the introduce of `&ipv4=` parameter to private trackers, as described by <https://github.com/qbittorrent/qBittorrent/issues/10051>. I have confirmed it on both Windows and Linux.

If an IPv4-NAT end user selects a particular interface, then this interface's local IPv4 may be set to `m_ipv4_interface` at here:
	
<https://github.com/arvidn/libtorrent/blob/e53bc71f63fabfd7d0cf502b4534f96e3d582d00/src/session_impl.cpp#L1993-L1999>

This leads to `false` for `want_v4` and `want_v6` and skips the followed setting part traversing all IPv4/6 interfaces. Finally, either `m_ipv4_interface` or `m_ipv6_interface` may hold a local IP address. Since there is no more check in torrent.cpp, the announce message to private tracker will include this invalid local IP and raise tracker's error report, like this:

![4](https://user-images.githubusercontent.com/38249940/50819605-7cec5b00-1365-11e9-99d7-98733e60dd31.jpg)
where the `&ipv6=` part is a link-local address.

A simple fix is to add a sanity check in torrent.cpp, as <https://github.com/airium/libtorrent/commit/831901a61cfec4f45559e644e60715db768254e8>. I can PR that if you think it's ok. But I am wondering how can it be fixed just in the above shown code, which may be a better fix.

arvidn:
I think `m_ipv4_interface` and `m_ipv6_interface` are not supposed to contain local IPs. would it be practical to add the check earlier, when those fields are assigned?

airium:
I created another fix <https://github.com/airium/libtorrent/commit/b7efb92894b43240d67cfa0e5283ec8829ac2f77>. This one places the check when `m_ipv4_interface` and `m_ipv6_interface` are firstly assigned. Does it implement "add the check earlier" correctly?

arvidn:
ok, I had some time to look at this. At a closer look it seems I was wrong. those members *are* expected to contain local IPs as well. Perhaps their most important use is to announce to trackers over both IPv4 and IPv6.

Would you mind creating a PR with your first patch?

airium:
Thanks. I have created it.

Rootax:
Why not use the second patch instead ? https://github.com/airium/libtorrent/commit/b7efb92894b43240d67cfa0e5283ec8829ac2f77

arvidn:
[...]

PR #3572, opened 2019-01-09
Avoid announcing local ip to private tracker

airium:
As discussed in <https://github.com/arvidn/libtorrent/issues/3571>.
Solve <https://github.com/qbittorrent/qBittorrent/issues/10051>.


arvidn:
thanks!

PR #3573, opened 2019-01-11
affinity for nodes with node IDs derived from their IP

arvidn:
similar to how low RTT DHT nodes are preferred over high RTT nodes

### DIFF
--- a/include/libtorrent/kademlia/dht_settings.hpp
+++ b/include/libtorrent/kademlia/dht_settings.hpp
@@ -163,7 +163,15 @@ namespace dht {
 		// same as the tcp interface
 		int service_port = 0;
 #endif
+	};
 
+	// internal
+	struct settings : dht_settings
+	{
+		// when this is true, nodes whose IDs are derived from their source IP
+		// according to BEP 42 (http://bittorrent.org/beps/bep_0042.html) are
+		// preferred in the routing table.
+		bool prefer_verified_node_ids = true;
 	};
 
 
--- a/src/kademlia/routing_table.cpp
+++ b/src/kademlia/routing_table.cpp
@@ -68,7 +68,7 @@ namespace {
 		container.erase(i);
 	}
 
-	bool verify_node_address(dht_settings const& settings
+	bool verify_node_address(dht::settings const& settings
 		, node_id const& id, address const& addr)
 	{
 		// only when the node_id pass the verification, add it to routing table.
@@ -100,8 +100,16 @@ void ip_set::erase(address const& addr)
 		erase_one(m_ip4s, addr.to_v4().to_bytes());
 }
 
+bool mostly_verified_nodes(bucket_t const& b)
+{
+	int const num_verified = static_cast<int>(std::count_if(b.begin(), b.end()
+		, [](node_entry const& e) { return e.verified; }));
+	if (num_verified == 0 && b.size() > 0) return false;
+	return num_verified >= static_cast<int>(b.size()) * 2 / 3;
+}
+
 routing_table::routing_table(node_id const& id, udp proto, int bucket_size
-	, dht_settings const& settings
+	, dht::settings const& settings
 	, dht_logger* log)
 	:
 #ifndef TORRENT_DISABLE_LOGGING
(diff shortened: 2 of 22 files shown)


src/kademlia/routing_table.cpp
```diff
@@ -626,6 +631,7 @@ routing_table::add_node_status_t routing_table::add_node_impl(node_entry e)
 	// split the last bucket
 	bool const can_split = (std::next(i) == m_buckets.end()
 		&& m_buckets.size() < 159)
+		&& e.verified
```
ssiloti: This is probably fine, but it feels wrong for the contents of the routing table to depend on the order in which nodes enter it. Maybe set a threshold based on the number of verified nodes in the last bucket?

src/kademlia/routing_table.cpp
```diff
@@ -772,11 +778,11 @@ routing_table::add_node_status_t routing_table::add_node_impl(node_entry e)
 			{
 				j = std::max_element(b.begin(), b.end()
 					, [](node_entry const& lhs, node_entry const& rhs)
-					{ return lhs.rtt < rhs.rtt; });
+					{ return lhs < rhs; });
```
ssiloti: You don't need the lambda function here anymore.

src/kademlia/routing_table.cpp
arvidn: yeah, I think I will let this patch sit for a while. I won't block cutting 1.2. The simulation tests also need to be updated

src/kademlia/routing_table.cpp
```diff
@@ -758,9 +764,7 @@ routing_table::add_node_status_t routing_table::add_node_impl(node_entry e)
 				// from these nodes, pick the one with the highest RTT
 				// and replace it
 
-				auto k = std::max_element(nodes.begin(), nodes.end()
-					, [](bucket_t::iterator lhs, bucket_t::iterator rhs)
-					{ return lhs->rtt < rhs->rtt; });
+				auto k = std::max_element(nodes.begin(), nodes.end());
```
arvidn: this is not correct. this will compare the iterators, not the values the iterators point to

src/kademlia/routing_table.cpp
arvidn: you mean that whether the last node (causing the bucket to split) is verified or not, should not matter so much as how many verified nodes there are in the bucket? that kind of makes sense. Although, you would imagine there would be a verified node eventually that would come around to causing the split.

ssiloti: Right. With the way this PR is now a bucket full of unverified nodes can be split by a single verified node, but a bucket full of verified nodes will not be split when an unverified node is added. So an attacker only needs to come up with a single verified ID in the appropriate range to force a split. This might let them trigger marginally more splits for a given number of attacker controlled IPs than if they needed to insert, say, `K / 2` verified peers in the home bucket before it will be split.

arvidn:
I think this is ready to be merged

include/libtorrent/alert_types.hpp
```diff
@@ -97,7 +97,9 @@ namespace libtorrent {
 		// the total number of nodes and replacement nodes
 		// in the routing table
 		int num_nodes;
+		int valid_id_nodes;
 		int num_replacements;
+		int valid_id_replacements;
```
arvidn: this would break the ABI :( I think I'll have to put this off until master

include/libtorrent/kademlia/dht_settings.hpp
```diff
@@ -116,6 +116,11 @@ namespace dht {
 		// error message is returned with a message saying "invalid node ID".
 		bool enforce_node_id = false;
 
+		// when this is true, nodes whose IDs are derived from their source IP
+		// according to BEP 42 (http://bittorrent.org/beps/bep_0042.html) are
+		// preferred in the routing table.
+		bool prefer_verified_node_ids = true;
```
arvidn: hm. this will also alter the ABI.

arvidn:
I think the `dht_settings` should probably be phased out into `settings_pack`, to be more ABI-robust.

PR #3575, opened 2019-01-11
refactor in file_progress to avoid dependency of alert_manager

aldenml:
Hi @arvidn, I want to code the unit test, but I would like to have your opinion on this first


aldenml:
Ok, about to push it

PR #3579, opened 2019-01-12
disk-I/O subsystem revamp to use memory mapped files

arvidn:
This is a major change, replacing the `storage_interface` as a customization point with the whole `disk_io_thread` implementing the entire disk subsystem, as the customization point.


.travis.yml
```diff
@@ -35,6 +35,7 @@ branches:
     - master
     - RC_1_1
     - RC_1_0
+    - new-disk-io
```
arvidn: this should probably be removed

appveyor.yml
```diff
@@ -1,9 +1,10 @@
 version: "{build}"
 branches:
   only:
-    - master
-    - RC_1_1
-    - RC_1_0
+  - master
+  - RC_1_1
+  - RC_1_0
+  - new-disk-io
```
arvidn: this should be removed

bindings/python/src/alert.cpp
```diff
@@ -135,7 +135,7 @@ dict session_stats_values(session_stats_alert const& alert)
     return d;
 }
 
-#ifndef TORRENT_NO_DEPRECATE
+#if TORRENT_ABI_VERSION == 1
```
arvidn: this should be fixed upstream

examples/client_test.cpp
```diff
@@ -299,7 +300,7 @@ FILE* g_log_file = nullptr;
 int peer_index(lt::tcp::endpoint addr, std::vector<lt::peer_info> const& peers)
 {
 	using namespace lt;
-	std::vector<peer_info>::const_iterator i = std::find_if(peers.begin(), peers.end()
+	auto i = std::find_if(peers.begin(), peers.end()
```
arvidn: this could be fixed upstream

include/libtorrent/disk_io_thread.hpp
```diff
@@ -41,240 +41,34 @@ POSSIBILITY OF SUCH DAMAGE.
...
-		// compare its hash against the hashes in the .torrent file.
-		int next_to_hash;
-
-		// the piece index for this cache entry.
-		piece_index_t piece;
+#if TORRENT_HAVE_MMAP || TORRENT_HAVE_MAP_VIEW_OF_FILE
 
-		enum kind_t { read_cache = 0, write_cache = 1, volatile_read_cache = 2 };
+// TODO: 3 move this class into the .cpp file and just expose the constructor.
+// the header/cpp-file pair should be renamed mmap_disk_io.hpp/.cpp
```
arvidn: this should be done before landing

include/libtorrent/aux_/file_view_pool.hpp
```diff
@@ -30,87 +30,124 @@ POSSIBILITY OF SUCH DAMAGE.
...
+		>;
+
+		// maps storage pointer, file index pairs to the lru entry for the file
+		files_container m_files;
 		mutable std::mutex m_mutex;
+
+		// the boost.multi-index container is not no-throw move constructable. In
+		// order to destruct m_files without holding the mutex, we need this
+		// separate pre-allocated container to move it into before releasing the
+		// mutex and clearing it.
```
ssiloti: I don't understand this comment. What problem does moving `m_files` then clearing the moved-to container avoid, and what does the move ctor not being no-throw have to do with it? It looks to me like this code is just trying to avoid holding the main mutex while doing the potentially long-running job of closing all of the mappings.

It's also not clear to me why `m_deferred_destruction` needs to be a member instead of a local variable inside of `file_view_pool::release()`.

arvidn: (from memory): You're right that the ultimate goal is to avoid holding the mutex while destructing files, as it may be expensive (especially on MacOS/HFS+). To achieve this, I move the main container (`m_files`) into another, temporary, one. release the mutex and then destruct the temporary container. Traditionally this is done by just constructing a local container, on the stack, and move the member container into that one.

The problem with that is that (default) constructing that local temporary container may throw an exception. I imagine this happens at a point in the program where correctly handling errors makes things very complex, so instead I pre-allocate the temporary container here, to always have it available. If the default construction will throw, it at least happens at a better time, when the file_view_pool is being constructed.

arvidn: I suppose one could argue that locking the mutex also may fail, and I'm not handling those errors correctly. I don't have those instrumented in my simulation that triggers exceptions though.

PR #3587, opened 2019-01-19
marked final error category classes, avoid -Wnon-virtual-dtor

aldenml:


src/natpmp.cpp
```diff
@@ -62,7 +62,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent {
 
-struct TORRENT_EXPORT pcp_error_category : boost::system::error_category
```
arvidn: this raises an interesting question (although, independent of your patch). Are the concrete error category classes implementation details, not available to clients?

arvidn: I suppose none of them are defined in a header

aldenml: I had a similar question when I realized they are confined to the `.cpp`, but... I don't know

PR #3598, opened 2019-01-23
remove cpp98 deque support

arvidn:


arvidn:
it seems msvc still implements C++98-style containers

PR #3605, opened 2019-02-02
update asio usage to networking TS compatible API

arvidn:


arvidn:
yeah, I haven't updated the handler allocators in libtorrent to use the standard API yet. I think that's the issue

elgatito:
@arvidn you can extend travis build with this ppa: https://launchpad.net/~mhier/+archive/ubuntu/libboost-latest

```
sudo add-apt-repository ppa:mhier/libboost-latest
sudo apt-get update
```

https://stackoverflow.com/a/50862419/1551009

elgatito:
```
wget https://dl.bintray.com/boostorg/release/1.68.0/source/boost_1_68_0.zip;
(cd boost_1_68_0/tools/build/src/engine; ./build.sh);
```

If it building in that way, ./bootstrap.sh + ./b2 is not working in travis?

arvidn:
I believe I would not only have to build boost, but install it as well. I haven't tried that yet, I'm worried I don't have sudo on the travis machines.

the boost-build based tests work fine, they just build the parts of boost they depend on. but the `cmake` script uses `find_package()` which I believe requires that dependency to be built already.

aldenml:
what are the implications for `boost` minimum version requirement with this?

arvidn:
boost-1.66.0 If I understand correctly. That's when boost.asio first included the networking TS API

arvidn:
that's still too new for xenial, which is the newest dist on travis..

arvidn:
I'm not sure why the `make check` stalls. All tests passed, but then nothing happened. I just took it out for now.

I think this is in a mergeable state now

PR #3606, opened 2019-02-03
pull in ed25519 dependency as a submodule

arvidn:


arvidn:
@aldenml one caveat with this is that it regresses your patch to use openSSL for sha512.

arvidn:
I'm having second thoughts about this.

aldenml:
good to know you are having them :)

PR #3610, opened 2019-02-08
don't copy a vector into the async_write operation for iovec

arvidn:
Since the chained_buffer will own the vector anyway, just make it return a span and copy that

### DIFF
--- a/src/peer_connection.cpp
+++ b/src/peer_connection.cpp
@@ -5616,7 +5616,7 @@ namespace libtorrent {
 #ifndef TORRENT_DISABLE_LOGGING
 		peer_log(peer_log_alert::outgoing, "ASYNC_WRITE", "bytes: %d", amount_to_send);
 #endif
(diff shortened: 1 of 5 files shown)


src/peer_connection.cpp
```diff
@@ -5616,7 +5616,7 @@ namespace libtorrent {
 #ifndef TORRENT_DISABLE_LOGGING
 		peer_log(peer_log_alert::outgoing, "ASYNC_WRITE", "bytes: %d", amount_to_send);
 #endif
-		std::vector<boost::asio::const_buffer> const& vec = m_send_buffer.build_iovec(amount_to_send);
+		auto const& vec = m_send_buffer.build_iovec(amount_to_send);
```
aldenml: Is it better to keep the `&` here or not?

arvidn: Oh, it should probably make a copy, but it probably doesn’t matter in practice, because of lifetime extension, and it being copied into the handler op

PR #3633, opened 2019-02-12
CMakeLists.txt: `libtorrent_include_files/ed25519` into `libtorrent_aux_include_files/`

gubatron:
note: cmake build is still broken due to
```
CMake Error at CMakeLists.txt:507 (add_library):
Cannot find source file:

    deps/try_signal/try_signal

 Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
  .hpp .hxx .in .txx
```
I thought it was maybe a path issue, but I cannot find a try_signal.* file, is it auto-generated or was it removed?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,7 +52,6 @@ set(libtorrent_include_files
 	disk_job_pool
 	disk_observer
 	download_priority
-	ed25519
 	entry
 	enum_net
 	error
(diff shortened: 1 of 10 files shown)


gubatron:
In relationship to work started in https://github.com/arvidn/libtorrent/pull/3632

arvidn:
in master you have to get submodules too:

```
git submodule update --init
```

PR #3649, opened 2019-02-16
Minor fixes - remove abort, atoi, and atol

latot:
Hi, some minor fixes to remove that functions because have undefined behaviors, and are deprecated (at least atoi and atol).

There is a section in the code where i don't know what to do if fails, for now have comments.

Bye.

### DIFF
--- a/include/libtorrent/error_code.hpp
+++ b/include/libtorrent/error_code.hpp
@@ -256,6 +256,8 @@ namespace libtorrent {
 			redirecting,
 			// The HTTP range header is invalid
 			invalid_range,
+			// The HTTP response did have a invalid content length
+			invalid_content_length,
 			// The HTTP response did not have a content length
 			no_content_length,
 			// The IP is blocked by the IP filter
--- a/src/http_parser.cpp
+++ b/src/http_parser.cpp
@@ -147,7 +147,7 @@ namespace libtorrent {
 		auto const i = m_header.find(key.to_string());
 		if (i == m_header.end()) return boost::none;
         char *str;
-		auto const val = std::strtod(i->second.c_str(), &str);
+		auto const val = std::strtol(i->second.c_str(), &str, 10);
 		if (val <= 0 || *str) return boost::none;
 		return seconds32(val);
 	}
@@ -211,7 +211,7 @@ restart_response:
 			if (m_protocol.substr(0, 5) == "HTTP/")
 			{
                 char *str;
-				m_status_code = strtol(read_until(line, ' ', line_end).c_str(), &str);
+				m_status_code = strtol(read_until(line, ' ', line_end).c_str(), &str, 10);
 			    if (*str)
 			    {
 				    m_state = error_state;
--- a/src/http_parser.cpp
+++ b/src/http_parser.cpp
@@ -146,8 +146,9 @@ namespace libtorrent {
 		// TODO: remove to_string() if we're in C++14
 		auto const i = m_header.find(key.to_string());
 		if (i == m_header.end()) return boost::none;
-		auto const val = std::atol(i->second.c_str());
-		if (val <= 0) return boost::none;
+        char *str;
+		auto const val = std::strtod(i->second.c_str(), &str);
+		if (val <= 0 || *str) return boost::none;
 		return seconds32(val);
 	}
 
@@ -209,7 +210,14 @@ restart_response:
 			m_protocol = read_until(line, ' ', line_end);
 			if (m_protocol.substr(0, 5) == "HTTP/")
 			{
-				m_status_code = atoi(read_until(line, ' ', line_end).c_str());
+                char *str;
+				m_status_code = strtol(read_until(line, ' ', line_end).c_str(), &str);
+			    if (*str)
+			    {
+				    m_state = error_state;
+				    error = true;
+				    return ret;
+			    }
 				m_server_message = read_until(line, '\r', line_end);
 
 				// HTTP 1.0 always closes the connection after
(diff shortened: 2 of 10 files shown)


include/libtorrent/error_code.hpp
```diff
@@ -256,6 +256,8 @@ namespace libtorrent {
 			redirecting,
 			// The HTTP range header is invalid
 			invalid_range,
+			// The HTTP response did have a invalid content length
+			invalid_content_length,
```
arvidn: you can't insert another value here, it will shift all the following error codes one step up and break ABI. Either use `invalid_range` or add another error code at the end of a section (where the next code is assigned a specific value)

src/http_parser.cpp
```diff
@@ -146,8 +146,9 @@ namespace libtorrent {
 		// TODO: remove to_string() if we're in C++14
 		auto const i = m_header.find(key.to_string());
 		if (i == m_header.end()) return boost::none;
-		auto const val = std::atol(i->second.c_str());
-		if (val <= 0) return boost::none;
+        char *str;
+		auto const val = std::strtol(i->second.c_str(), &str, 10);
+		if (val <= 0 || *str) return boost::none;
```
arvidn: you're not handling the case where the converted value won't fit in the returned variable. `errno` is set to `ERANGE` and the return value is one of  `LONG_MAX`, `LONG_MIN`, `LLONG_MAX` or `LLONG_MIN`

src/http_parser.cpp
```diff
@@ -209,7 +210,14 @@ namespace libtorrent {
 			m_protocol = read_until(line, ' ', line_end);
 			if (m_protocol.substr(0, 5) == "HTTP/")
 			{
-				m_status_code = atoi(read_until(line, ' ', line_end).c_str());
+                char *str;
+				m_status_code = strtol(read_until(line, ' ', line_end).c_str(), &str, 10);
+			    if (*str)
```
arvidn: the tests are failing, and I'm suspecting that this assumption that `str` won't be set to anything isn't correct. there may be white space or a newline following the number. It's not an error.

src/session_impl.cpp
```diff
@@ -182,7 +182,7 @@ using namespace std::placeholders;
 #ifdef BOOST_NO_EXCEPTIONS
 namespace boost {
 
-	void throw_exception(std::exception const& e) { std::abort(); }
+	void throw_exception(std::exception const& e) { exit(0); }
```
arvidn: why this change?

src/http_parser.cpp
arvidn: I would suggest wrapping the use of `strtol()` in a function (probably in `string_utils.cpp`) that does all the error handling and has a simpler interface.

In fact, it may even be simpler to implement the string to number conversion "manually" than to handle all the error cases from the standard functions.

arvidn:
If you're up for it, I would really like to toss `http_parser` and gut `http_connection`/`web_seed_connection` and use boost.beast instead. That's a bigger project though, and should be done in master

src/session_impl.cpp
latot: seems some systems have undefined behavior with abort....

src/http_parser.cpp
latot: mm, i think first we/i (maybe you already know this) need know what format we should support, how there was trying to get numbers i suppose the strings would be that, but if not i think is necessary define the format to avoid weird things.

arvidn: HTTP is fairly well defined

src/session_impl.cpp
arvidn: how so?

src/http_parser.cpp
latot: Hi, in that case we should be getting just a number, maybe we just need check what we are getting for the error.

arvidn: `strtol()` is documented here: https://en.cppreference.com/w/cpp/string/byte/strtol
"The functions sets the pointer pointed to by str_end to point to the character past the last character interpreted".

Your check `if (*str)` does *not* indicate an error, you're just checking whether there *is* a character after the number

latot: right, the idea is check if there is something more than a number, if there is something more i think we should read until that chracter and from there get the number.

src/session_impl.cpp
latot: i found this in sonarcloud, one of the code checkers you have, the description provided is:
```<stdlib.h>'s abort, exit, getenv, and system have undefined, and implementation-defined behaviors, and should therefore be used.```

 if you think is not neccessary i can remove it, but from atol and atoi i can confirm is deprecated.

arvidn: I find it hard to believe that `abort` has *undefined* behavior. It has *implementation* defined behavior, specifically the behavior that's intended here

arvidn: abort (on most normal computers) will raise a `SIGABRT` which will be caught by debuggers. That's the intended behavior. Not all systems have signals, but it is likely to do something else that also will be caught by debuggers

src/http_parser.cpp
arvidn: I'm supposing there's white space *after* the number, which is perfectly fine in HTTP.

vktr:
@arvidn have you started on the boost.beast conversion? Or is it something I can start on perhaps?

latot:
Hi!, sadly i don't have too much time now, i would appreciate if you can work on it.

Thx and sorry to keep this all this time.

vktr:
Hey @latot, no worries! I'll pick it up 😃

PR #3651, opened 2019-02-16
move hasher512 into the aux-namespace and make it private

arvidn:


arvidn:
@aldenml this looks reasonable?

aldenml:
yes, do you have intentions to move `hasher` too?

arvidn:
I was considering this, but I'm a bit hesitant, as it's been public for longer and is actually used by `connection_tester.cpp`

PR #3654, opened 2019-02-17
fix handler sizes for msvc debug builds

ssiloti:


arvidn:
in the error message you got, the handler type should be included. I'm curious if you know what part of the capture/closure changes size in debug mode.

Previously the write handler would bind a `std::vector<iovec_t>`, which changes size with debug iterators. I fixed that by changing it to a `span<>`. I', curious if there's another fix involving changing the types in the capture/closure.

ssiloti:
The offending handlers are:

- The lambda passed to `async_read_some` in `peer_connection::setup_receive`
- The lambda passed to `async_write_some` in `peer_connection::setup_send`

I'm not sure why these handlers change size. The lambdas only capture an `std::shared_ptr` and I confirmed that both lambda objects are 16 bytes in both Debug and Release builds.

arvidn:
interesting. It's probably some member of boost.asio's internal operation class that changes size then. Sort of how enabling handler tracking adds another 64 bits

PR #3658, opened 2019-02-17
make random_shuffle() take a range instead of two iterators

arvidn:
use random_bytes() instead of std::generate() and random_byte(). Remove unused hasher.hpp includes


aldenml:
nice cleanup!

PR #3660, opened 2019-02-19
remove the (unused) block_cache_reference to simplify disk I/O

arvidn:


src/disk_io_thread.cpp
```diff
@@ -399,12 +399,8 @@ TORRENT_EXPORT std::unique_ptr<disk_interface> mmap_disk_io_constructor(
 
 	void disk_io_thread::remove_torrent(storage_index_t const idx)
 	{
-		auto& pos = m_torrents[idx];
-		if (pos->dec_refcount() == 0)
```
arvidn: this is the only questionable part of this patch, but I don't see how this could be correct anyway. If it `dec_refcount()` can ever be > 0 here, the storage is never freed, as far as I can tell. And I think it's a safe assumption that by the time the torrent object destructs (and the `storage_holder` holder, with it), there are no more references to buffers in this storage. All peer objects should have been destructed by then

ssiloti: I suppose you could replace the `if` with a `TORRENT_ASSERT` if there's any doubt.

arvidn: yeah, except that this patch removes the reference-counting of holder keeping the storage object alive. I would really want to have this guarantee, that all buffer holders for a storage have been destructed by the time the storage itself is destructed (and I'm pretty sure it's maintained by the torrent object containing the `storage_holder`, which means the `torrent` object must be destructing before the disk io object is asked to remove the corresponding storage object).

i.e. all peers belonging to the torrent should have been destructed by then

PR #3662, opened 2019-02-20
upgrade compiler version to C++14

arvidn:


arvidn:
boost.asio appears to be having problems with C++14, as it assumes <experimental/string_view> exists. which it doesn't anymore on clang. Perhaps it's better to just go all the way to C++17

hijxf:
Is C++17 still the route?

arvidn:
I haven't really spent a lot of time considering the benefits of switching to C++17 yet. I imagine it goes something like this:

* configuration header could be simplified with `__has_include`
* some code could be cleaned up a bit with structured bindings
* (most of) custom path-handling code can be replaced by std::filesystem (but then again, it could also be replaced by boost.filesystem in C++14)
* `boost::variant` could be replaced by `std::variant`, which may be a bit more efficient, since it doesn't use double buffering
* `boost::optional` could be replaced by `std::optional`
* `std::byte` could be used pervasively (but a custom `byte` type could also be used with similar benefits, except in the public API)
* there may be a few places `to_chars()` and `from_chars()` could be used instead of custom conversion code, or instead locale-dependent conversions
* standard `[[nodiscard]]` markup (just saves a macro)
* standard `[[maybe_unused]]` markup (also just saves a macro)

I don't think these are huge benefits. The first item is probably the one I like the most. It would probably improve the accuracy of the configuration header.

Can you think of other benefits?

gubatron:
better template deductions
```cpp
std::pair<int, double> p(10, 0.0);
// same as
std::pair p(10, 0.0); // deduced automatically!
```

lots of code would probably be simplified

xavier2k6:
May or may not be relevant to you.....
https://blog.qt.io/blog/2019/08/01/precompiled-headers-and-unity-jumbo-builds-in-upcoming-cmake/

hijxf:
> May or may not be relevant to you.....
> https://blog.qt.io/blog/2019/08/01/precompiled-headers-and-unity-jumbo-builds-in-upcoming-cmake/

i'm not personally a fan of pch, but i see why they exist or why people would want them. i guess you have them in your build folder anyhow, but we want to be careful and not design the software to cater to them.

PR #3663, opened 2019-02-20
upgrade compilers on CI to C++14 compatible ones

arvidn:


ssiloti:
In Boost Build MSVC 2017 is known as msvc-14.1 (major stayed the same because 2017 is ABI compatible with 2015).

arvidn:
oh, I was tricked by https://en.cppreference.com/w/cpp/compiler_support , it's talking about 19.0.
Also, this is talking about version 15.x https://en.wikipedia.org/wiki/Microsoft_Visual_Studio#2019

I'll try changing

ssiloti:
Versioning with Visual Studio/Visual C++ is a mess. There are actually three different version numbers at play here. 15.x is the Visual Studio version number. 19.0 is the C/C++ compiler version number which goes back to the original Microsoft C compiler from the 80s. When Visual C++ was introduced it got it's own version number which is where 14.1 comes from.

arvidn:
@ssiloti @zeule I need some cmake help. The last remaining issue in this patch is that cmake (on windows) doesn't find `boost_system`.

I'm building and staging boost_system (as you can see in the output). `BOOST_ROOT` is set as an environment variable and I set `BOOST_LIBRARYDIR` on the cmake command line (with `-D`), to point specifically to the %BOOST_ROOT%\stage\lib directory, where `libboost_system.lib` has been placed.

it still can't find it. Any hints on what else I can do to help cmake find the library?

https://ci.appveyor.com/project/arvidn/libtorrent/builds/22542972/job/vo0wuq2mxwvmsr5u#L213

ssiloti:
It's because you're building libtorrent as a shared lib but only built boost as a static lib. You should be able to set `Boost_USE_STATIC_LIBS=ON` from the command line to get a shared libtorrent with static boost. Or just set `BUILD_SHARED_LIBS=OFF`.

PR #3670, opened 2019-02-22
fix build with no mmap support

arvidn:
fix posix_disk_io to make the tests pass with it, and build it on travis to get CI coverage. Add build option to disable mmap support to enable testing the posix_disk_io.


src/posix_disk_io.cpp
```diff
@@ -128,7 +128,7 @@ namespace {
 			storage_index_t const idx = m_free_slots.empty()
 				? m_torrents.end_index()
 				: pop(m_free_slots);
-			auto storage = std::make_shared<posix_storage>(std::move(params));
+			auto storage = std::unique_ptr<posix_storage>(new posix_storage(std::move(params)));
```
aldenml: not 100% sure if you are finally requiring c++14, but if that's the case, then `std::make_unique` here?

arvidn: oh, I thought I fixed this in a separate commit. and yes, I just switched master over to C++14 (and I may go to 17 at some point)

PR #3679, opened 2019-02-24
Add binaries to gitignore and remove duplicate words in comments

AmirAbrams:
- Remove redundant words in comments like ` to to ` and ` and and `

- Add binaries to git ignore file to prevent untracked files:
libtorrent-gitignore-files


arvidn:
I'm surprised hunspell doesn't catch those duplicate words

PR #3688, opened 2019-02-28
allow padfiles of equal size to share the same filename

arvidn:
@ssiloti does this look right?

### DIFF
--- a/src/torrent_info.cpp
+++ b/src/torrent_info.cpp
@@ -59,6 +59,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/lazy_entry.hpp"
 #endif
 
+#include <unordered_map>
 #include <unordered_set>
 #include <cstdint>
 #include <iterator>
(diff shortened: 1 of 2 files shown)


src/torrent_info.cpp
```diff
@@ -665,23 +665,30 @@ namespace {
...
-					char new_ext[50];
-					std::snprintf(new_ext, sizeof(new_ext), ".%d%s", cnt, ext.c_str());
-					filename = base + new_ext;
-				}
-				while (!files.insert(filename).second);
+			auto const ret = files.insert({filename, i});
+			if (ret.second) continue;
+			// pad files are allowed to collide with each-other, as long as they have
+			// the same size.
+			file_index_t const other_idx = ret.first->second;
```
ssiloti: There needs to be a check for `other_idx == -1` before calling `file_flags` etc.

arvidn: ah, right! I'll add a test for that too

PR #3697, opened 2019-03-05
more message-sent hooks for extensions redux

ssiloti:
This is a revival of #2986 rebased on RC_1_2


arvidn:
since this changes the interface of `peer_plugin`, I think it should probably be put in master

PR #3702, opened 2019-03-10
minor lint oriented refactor, moved a couple of functions to aux

aldenml:


appveyor.yml
```diff
@@ -4,7 +4,7 @@ branches:
     - master
     - RC_1_2
     - RC_1_1
-os: Visual Studio 2015
+image: Visual Studio 2015
```
aldenml: a few days back I found that `os` is not documented here: https://www.appveyor.com/docs/appveyor-yml

arvidn:
I'm not sure why, but the macOS travis instance seems to fail consistently.

aldenml:
it looks like some change in the pre-installed `brew` packages, `which python2` fails: https://travis-ci.org/arvidn/libtorrent/jobs/504415430#L5090

I will look at it

PR #3709, opened 2019-03-14
update symlinks to conform to BEP 47

arvidn:


src/torrent_info.cpp
```diff
@@ -526,6 +529,8 @@ namespace {
 				, info_ptr_diff, false, pad_file_cnt, ec))
 				return false;
 		}
+		// this rewrites invalid symlinks to point to themselves
+		target.validate_symlinks();
```
aldenml: Hi @arvidn, I don't see the return of `validate_symlinks` used in this PR, is that ignore correct?

arvidn: good point. `validate_symlinks()` should probably not return anything, and should probably be called something like `sanitize_symlinks()`

PR #3710, opened 2019-03-15
fix python bindings for peer_info

arvidn:


devos50:
Yes, that fixes the issue, thanks! 👍

PR #3718, opened 2019-03-18
remove old, hacky and unused split_string() function

arvidn:


Dobatymo:
What about the `replace_extension` tests?

arvidn:
https://github.com/arvidn/libtorrent/pull/3720

PR #3731, opened 2019-03-22
deprecate file_name_ptr() and file_name_len() in file_storage

arvidn:


src/torrent_info.cpp
```diff
@@ -1545,17 +1545,18 @@ namespace {
...
-				TORRENT_ASSERT(m_files.file_name_ptr(i) < m_info_section.get() + m_info_section_size);
+				TORRENT_ASSERT(m_files.file_name(i).data() >= m_info_section.get());
+				TORRENT_ASSERT(m_files.file_name(i).data() < m_info_section.get() + m_info_section_size);
 			}
 			else
 			{
-				// name must be a valid string
-				TORRENT_ASSERT(strlen(m_files.file_name_ptr(i)) < 2048);
+				// name must be a null terminated string
+				string_view const name = m_files.file_name(i);
```
aldenml: I wonder if this is not a problem (unused variable warning) when compiling with not asserts

arvidn: I don't think invariant-checks without asserts is a valid configuration though

aldenml: true, didn't notice that this is inside `torrent_info::check_invariant()`

PR #3738, opened 2019-03-23
add Python binding bep38 (add_similar_torrent & add_collection)

joriscarrier:


arvidn:
I just added these in here: https://github.com/arvidn/libtorrent/pull/3733 (along with another fix). there's still something not working right no mac though

Issue #3739, opened 2019-03-24
disk_io_thread race condition


arvidn:
@ssiloti I'm including you since you were involved a bit in some of the threading logic here. I'm working on a patch, maybe you can review it later. Please let me know if you have any ideas.

libtorrent version (or branch): `RC_1_2`

platform/architecture: at least amd64

compiler and compiler version: GCC 6, clang 7

There is a subtle race condition in the disk io thread in the logic of starting/stopping threads and whether to post jobs or to execute them immediately.

(The main reason it's possible to have zero disk I/O threads is for small systems as well as the simulator).

The symptom is that `test_torrent.cpp.async_load_deprecated` is reporting memory leaks when running under leak sanitizer. For example:

```
Direct leak of 96 byte(s) in 3 object(s) allocated from:
    #0 0x7fd4fa729ec0 in operator new(unsigned long)
    #1 0x556c62ea2217 in std::_Function_base::_Base_manager<std::_Bind<std::_Mem_fn<void (libtorrent::torrent::*)()> (std::shared_ptr<libtorrent::torrent>)> >::_M_init_functor(std::_Any_data&, std::_Bind<std::_Mem_fn<void (libtorrent::torrent::*)()> (std::shared_ptr<libtorrent::torrent>)>&&, std::integral_constant<bool, false>) 
    #2 0x556c62e64226 in std::_Function_base::_Base_manager<std::_Bind<std::_Mem_fn<void (libtorrent::torrent::*)()> (std::shared_ptr<libtorrent::torrent>)> >::_M_init_functor(std::_Any_data&, std::_Bind<std::_Mem_fn<void (libtorrent::torrent::*)()> (std::shared_ptr<libtorrent::torrent>)>&&)
    #3 0x556c62e1aaa3 in std::function<void ()>::function<std::_Bind<std::_Mem_fn<void (libtorrent::torrent::*)()> (std::shared_ptr<libtorrent::torrent>)>, void, void>(std::_Bind<std::_Mem_fn<void (libtorrent::torrent::*)()> (std::shared_ptr<libtorrent::torrent>)>) 
    #4 0x556c62d0ca92 in libtorrent::torrent::abort() ../src/torrent.cpp:4316
    #5 0x556c62a31030 in libtorrent::aux::session_impl::abort() ../src/session_impl.cpp:866
[...]

PR #3740, opened 2019-03-24
fix subtle race condition in disk I/O thread (for short-lived sessions)

arvidn:
@ssiloti does this look reasonable? related to https://github.com/arvidn/libtorrent/issues/3739

### DIFF
--- a/include/libtorrent/disk_io_thread_pool.hpp
+++ b/include/libtorrent/disk_io_thread_pool.hpp
@@ -88,7 +88,11 @@ namespace libtorrent {
 		// check if there is an outstanding request for I/O threads to stop
 		// this is a weak check, if it returns true try_thread_exit may still
 		// return false
-		bool should_exit() { return m_threads_to_exit > 0; }
+		bool should_exit() const
+		{
+			std::unique_lock<std::mutex> l(m_mutex);
+			return m_max_threads > int(m_threads.size());
+		}
 		// this should be the last function an I/O thread calls before breaking
 		// out of its service loop
 		// if it returns true then the thread MUST exit
@@ -100,7 +104,7 @@ namespace libtorrent {
 		// thread exits
 		// it can be used to trigger maintenance jobs which should only run on one thread
 		std::thread::id first_thread_id();
-		int num_threads()
+		int num_threads() const
 		{
 			std::lock_guard<std::mutex> l(m_mutex);
 			return int(m_threads.size());
@@ -120,8 +124,6 @@ namespace libtorrent {
 		pool_thread_interface& m_thread_iface;
 
 		std::atomic<int> m_max_threads;
-		// the number of threads the reaper decided should exit
-		std::atomic<int> m_threads_to_exit;
 
 		// must hold m_mutex to access
 		bool m_abort;
@@ -131,7 +133,7 @@ namespace libtorrent {
 		std::atomic<int> m_min_idle_threads;
 
 		// ensures thread creation/destruction is atomic
-		std::mutex m_mutex;
+		mutable std::mutex m_mutex;
 
 		// the actual threads running disk jobs
 		std::vector<std::thread> m_threads;
--- a/src/disk_io_thread.cpp
+++ b/src/disk_io_thread.cpp
@@ -270,7 +270,8 @@ constexpr disk_job_flags_t disk_interface::cache_hit;
 		// see also the comment in thread_fun
 		std::unique_lock<std::mutex> l(m_job_mutex);
 		if (m_abort.exchange(true)) return;
-		bool const no_threads = m_num_running_threads == 0;
+
+		bool const no_threads = num_threads() == 0;
 		// abort outstanding jobs belonging to this torrent
 
 		for (auto i = m_hash_io_jobs.m_queued_jobs.iterate(); i.get(); i.next())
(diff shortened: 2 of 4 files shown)


src/disk_io_thread.cpp
```diff
@@ -270,7 +270,8 @@ constexpr disk_job_flags_t disk_interface::cache_hit;
 		// see also the comment in thread_fun
 		std::unique_lock<std::mutex> l(m_job_mutex);
 		if (m_abort.exchange(true)) return;
-		bool const no_threads = m_num_running_threads == 0;
+
+		bool const no_threads = num_threads() == 0;
```
arvidn: `m_num_running_threads` may be zero here because none of the threads have started yet. Instead, look at the number of threads we're *configured* to have.

src/disk_io_thread.cpp
```diff
@@ -3100,7 +3101,6 @@ constexpr disk_job_flags_t disk_interface::cache_hit;
 		DLOG("started disk thread %s\n", thread_id_str.str().c_str());
 
 		std::unique_lock<std::mutex> l(m_job_mutex);
-		if (m_abort) return;
```
arvidn: even if we have aborted at this point, we still need to enter the main loop, process jobs and exit gracefully

src/disk_io_thread_pool.cpp
```diff
@@ -70,8 +70,6 @@ namespace libtorrent {
 	void disk_io_thread_pool::abort(bool wait)
 	{
 		std::unique_lock<std::mutex> l(m_mutex);
-		if (m_abort) return;
-		m_max_threads = 0;
```
arvidn: in this patch, `m_max_threads` is used to start up threads and to configure the number of threads once started up. It's not used to synchronize the shutdown, as that's the problem, you can't tell the difference between not having started up yet and wanting to shut down. `m_abort` is used to signal that we want to shut down, not setting the number of threads to 0

src/disk_io_thread_pool.cpp
```diff
@@ -103,6 +101,8 @@ namespace libtorrent {
 
 	bool disk_io_thread_pool::try_thread_exit(std::thread::id id)
 	{
+		if (m_abort) return true;
```
arvidn: I'm not sure about this one. the other case where we return true, we also remove the thread from `m_threads`. I should try remove this

arvidn: there would need to be some special case below though, to remove threads when aborting

arvidn: the important part is that when we exit via this mechanism, we always make sure the job queue is empty first

src/disk_io_thread_pool.cpp
```diff
@@ -158,7 +157,7 @@ namespace libtorrent {
 			; ++i)
 		{
 			// if this is the first thread started, start the reaper timer
-			if (m_threads.empty())
+			if (m_threads.empty() && !m_abort)
```
arvidn: just some defensive programming to make sure we don't reactivate the idle timer during shut down, as we may get stuck in `io_context::run()` waiting for it to trigger before exiting the last thread

src/disk_io_thread_pool.cpp
```diff
@@ -142,7 +142,6 @@ namespace libtorrent {
 		// but do it to avoid acquiring the mutex in the trivial case
 		if (m_num_idle_threads >= queue_size) return;
 		std::lock_guard<std::mutex> l(m_mutex);
-		if (m_abort) return;
```
arvidn: even though we're aborting, we still need to service disk jobs, as long as there are any in the queue

src/disk_io_thread_pool.cpp
arvidn: it's also accessing `m_abort` without holding `m_mutex`...

arvidn:
On the one hand I would like to keep this patch as small as possible, but I would also like the end-result be a bit more intuitive exactly how it works. I don't believe I fully understand all the possible interleavings of events right now. The last commit tried to simplify things a bit, but it appears to be incorrect

include/libtorrent/disk_io_thread_pool.hpp
```diff
@@ -120,8 +124,6 @@ namespace libtorrent {
 		pool_thread_interface& m_thread_iface;
 
 		std::atomic<int> m_max_threads;
-		// the number of threads the reaper decided should exit
-		std::atomic<int> m_threads_to_exit;
```
ssiloti: I don't think we can remove `m_threads_to_exit`. This PR tries to use `m_max_threads` instead but that breaks the scalable thread pool. At it is when the thread pool goes from max threads to idle it will reduce `m_max_threads` to zero which will prevent any new threads from being created. There needs to be a separate control for stopping threads due to lack of demand vs due to the configured max threads being reduced.

arvidn: yeah, this patch clearly doesn't work. Do you have any ideas of how to solve this in a simple way?

One other aspect is that when the session terminates immediately, the disk I/O threads don't come up until *after* `abort()` is called on `disk_io_thread`, so the destructor has to abort (and wait for the thread) again

ssiloti: I was able to repro the memory leak. I found that, after removing the eager abort bail out in `thread_fun` the other issue which was preventing the queue from being processed was deferred job submission. The session can shutdown before jobs have been submitted in which case no thread is created to handle the deferred jobs. Simple solution is to always submit queued jobs just before initiating an abort. See #3744

PR #3742, opened 2019-03-24
improve disk I/O logging

arvidn:


src/disk_io_thread.cpp
```diff
@@ -94,6 +98,16 @@ namespace libtorrent {
...
 		static const time_point start = clock_type::now();
+		// map thread IDs to low numbers
+		static std::unordered_map<std::thread::id, int> thread_ids;
+
+		std::thread::id const self = std::this_thread::get_id();
+
+		std::unique_lock<std::mutex> l(log_mutex);
+		auto it = thread_ids.find(self);
+		if (it == thread_ids.end())
+			it = thread_ids.insert({self, int(thread_ids.size())}).first;
```
ssiloti: Why not just do `insert` instead of `find` then `insert`?

arvidn: good question, I was thinking about that, but I guess I didn't do a second pass over the patch

PR #3744, opened 2019-03-25
always flush disk I/O job queue before shutdown

### LINKED ISSUE #3739: disk_io_thread race condition

ssiloti:
Leaving jobs on the queue can lead to memory leaks

Fixes #3739

### DIFF
--- a/src/disk_io_thread.cpp
+++ b/src/disk_io_thread.cpp
@@ -3211,7 +3211,7 @@ constexpr disk_job_flags_t disk_interface::cache_hit;
 	void disk_io_thread::abort_jobs()
 	{
 		TORRENT_ASSERT(m_magic == 0x1337);
-		TORRENT_ASSERT(!m_jobs_aborted.exchange(true));
+		if (m_jobs_aborted.test_and_set()) return;
 
 		jobqueue_t jobs;
 		m_disk_cache.clear(jobs);
--- a/src/disk_io_thread.cpp
+++ b/src/disk_io_thread.cpp
@@ -261,16 +261,23 @@ constexpr disk_job_flags_t disk_interface::cache_hit;
 
 		TORRENT_ASSERT(m_magic == 0x1337);
 		m_magic = 0xdead;
+		TORRENT_ASSERT(m_generic_io_jobs.m_queued_jobs.empty());
+		TORRENT_ASSERT(m_hash_io_jobs.m_queued_jobs.empty());
 	}
 #endif
 
 	void disk_io_thread::abort(bool const wait)
 	{
+		// first make sure queued jobs have been submitted
+		// otherwise the queue may not get processed
+		submit_jobs();
+
 		// abuse the job mutex to make setting m_abort and checking the thread count atomic
 		// see also the comment in thread_fun
 		std::unique_lock<std::mutex> l(m_job_mutex);
 		if (m_abort.exchange(true)) return;
-		bool const no_threads = m_num_running_threads == 0;
+		bool const no_threads = m_generic_threads.num_threads() == 0
+			&& m_hash_threads.num_threads() == 0;
 		// abort outstanding jobs belonging to this torrent
 
 		for (auto i = m_hash_io_jobs.m_queued_jobs.iterate(); i.get(); i.next())
@@ -3100,7 +3107,6 @@ constexpr disk_job_flags_t disk_interface::cache_hit;
 		DLOG("started disk thread %s\n", thread_id_str.str().c_str());
 
 		std::unique_lock<std::mutex> l(m_job_mutex);
-		if (m_abort) return;
 
 		++m_num_running_threads;
 		m_stats_counters.inc_stats_counter(counters::num_running_threads, 1);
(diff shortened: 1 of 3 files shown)


arvidn:
```
assertion failed. Please file a bugreport at https://github.com/arvidn/libtorrent/issues
Please include the following information:

version: 1.2.0.0-c76e9cc5e

file: '../src/disk_io_thread.cpp'
line: 3214
function: abort_jobs
expression: !m_jobs_aborted.exchange(true)

stack:
1: libtorrent::assert_fail(char const*, int, char const*, char const*, char const*, int)
2: libtorrent::disk_io_thread::abort_jobs()
3: libtorrent::disk_io_thread::thread_fun(libtorrent::disk_io_thread::job_queue&, libtorrent::disk_io_thread_pool&)
```

I think this suggests that `abort_jobs` is called twice, and it doesn't like that

src/disk_io_thread.cpp
```diff
@@ -261,16 +261,23 @@ constexpr disk_job_flags_t disk_interface::cache_hit;
...
+		// otherwise the queue may not get processed
+		submit_jobs();
+
 		// abuse the job mutex to make setting m_abort and checking the thread count atomic
 		// see also the comment in thread_fun
 		std::unique_lock<std::mutex> l(m_job_mutex);
 		if (m_abort.exchange(true)) return;
-		bool const no_threads = m_num_running_threads == 0;
+		bool const no_threads = m_generic_threads.num_threads() == 0
+			&& m_hash_threads.num_threads() == 0;
```
arvidn: I think the assert in CI is caused by this being `true`, even though there are threads that have been spawned, they just don't "count" yet. Leading to the last thread also calling `abort_jobs()`, just as is done a few lines below here. And there seems to be an assert that it's only called once.

ssiloti: When threads are spawned they are added to the thread vector while holding `m_job_mutex` so I don't think it could be the startup path. I think the problem is the thread shutdown path. The last thread calls `abort_jobs` after it has decremented `m_num_running_threads` and released `m_job_mutex`, so I don't see any easy way to detect a thread which might be lurking at the end of `thread_fun` right before the call to `abort_jobs`. We may need to remove the assert and rely on `m_jobs_aborted` to prevent a double invocation.

arvidn:
it looks like `test_resume` hangs now..

PR #3745, opened 2019-03-25
some minor refactoring of disk I/O utility functions

arvidn:
to reduce duplicate code across storage and posix_storage

### DIFF
--- a/include/libtorrent/aux_/storage_utils.hpp
+++ b/include/libtorrent/aux_/storage_utils.hpp
@@ -103,6 +103,8 @@ namespace aux {
 		, std::string const& save_path
 		, stat_cache& stat
 		, storage_error& ec);
+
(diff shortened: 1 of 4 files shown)


src/part_file.cpp
```diff
@@ -188,12 +194,18 @@ namespace libtorrent {
 		auto const f = m_file;
 		l.unlock();
 
-		return int(f->writev(slot_offset(slot) + offset, bufs, ec));
+		int const ret = int(f->writev(slot_offset(slot) + offset, bufs, ec.ec));
+		if (ec)
+		{
+			ec.operation = operation_t::partfile_write;
+			ec.file(torrent_status::error_file_partfile);
+			ec.operation = operation_t::partfile_read;
```
aldenml: this double `ec.operation = ` looks strange

PR #3753, opened 2019-03-28
request contiguous pieces from fast peers

arvidn:
extend the whole_pieces_threshold setting to also request contiguous pieces from fast peers

### DIFF
--- a/src/request_blocks.cpp
+++ b/src/request_blocks.cpp
@@ -116,12 +116,24 @@ namespace libtorrent {
 			&& !time_critical_mode
 			&& t.settings().get_int(settings_pack::whole_pieces_threshold) > 0)
 		{
+			// if our download rate lets us download a whole piece in
+			// "whole_pieces_threshold" seconds, we prefer to pick an entire piece.
+			// If we can download multiple whole pieces, we prefer to download that
+			// many contiguous pieces.
+
+			// download_rate times the whole piece threshold (seconds) gives the
+			// number of bytes downloaded in one window of that threshold, divided
+			// by the piece size give us the number of (whole) pieces downloaded
+			// in the window.
+			int const contiguous_pieces =
+				std::min(c.statistics().download_payload_rate()
+				* t.settings().get_int(settings_pack::whole_pieces_threshold)
+				, 8 * 1024 * 1024)
+				/ t.torrent_file().piece_length();
+
 			int const blocks_per_piece = t.torrent_file().piece_length() / t.block_size();
-			prefer_contiguous_blocks
-				= (c.statistics().download_payload_rate()
-				> t.torrent_file().piece_length()
-				/ t.settings().get_int(settings_pack::whole_pieces_threshold))
-				? blocks_per_piece : 0;
+
+			prefer_contiguous_blocks = contiguous_pieces * blocks_per_piece;
 		}
 
 		// if we prefer whole pieces, the piece picker will pick at least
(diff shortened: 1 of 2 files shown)


arvidn:
maybe the number of contiguous blocks should be capped at 4MiB or something like that

ssiloti:
I don't think this is going to be sufficient. Disk throughput is most likely to be an issue with heavily over-seeded torrents. In this case libtorrent ends up downloading from many seeds at a relatively slow rate per-seed, so the logic in this PR would not kick in or at least not as much as it should.

arvidn:
right, but it's quite hard to know whether the bottleneck is my *download* connection or everyone else's *upload* connection. In the examples posted by @sserdda-liamE, this *would* be sufficient, right? as those download rates are still high

ssiloti:
> it's quite hard to know whether the bottleneck is my download connection or everyone else's upload connection

That is true, but I don't see how it is relevant. If I understand @sserdda-liamE's concern correctly, they are worried about the disk becoming the bottleneck. Libtorrent already detects this situation: when allocating a disk buffer for an incoming piece exceeds the buffer pool limit the connection is paused while buffers are flushed to disk. IMO this is what should be the signal for the piece picker to favor larger contiguous blocks. The picking policy should be applied globally too, not just within a single picker invocation.

arvidn:
yeah, that's an interesting approach. It doesn't address the same problem if it's happening to the seed though.

arvidn:
actually, if the local disk is the bottleneck, it's not obvious that requesting contiguous blocks from peers would help. I can imagine it making it *worse*, as you spread out the blocks you're flushing to disk even more. In an ideal world, you would download every block in a contiguous range simultaneously, from separate peers, and then flush it as one unit. If you instead download *more* sequentially from all peers, you'll increase the pressure on the disk cache, to keep more bytes in RAM in order to flush equally large ranges.

ssiloti:
Yes, that's what I meant with my last sentence about applying the policy globally. The piece sequencing needs to be done per-torrent, not per-peer.

arvidn:
I'm not quite following. Do you mean that if the disk is the bottleneck on the peers we're downloading from, it doesn't really matter what rate we're getting from them, if we're requesting individual blocks we may contribute to the problem?

If that's the case, a larger piece size wouldn't help, right? We have an affinity to complete pieces which will improve our local disk throughput. I think I know what you mean now. To improve our own local disk throughput, we would want a way to say "this range of pieces is now the ones everyone picks from, so we can flush it as one large write". Is that what you mean?

ssiloti:
Yes. You could think of it as an extension of `prioritize_partials` so that it would prioritize partially downloaded blocks of n pieces instead of just single pieces.

arvidn:
right now, the piece picker maintains a sorted list of all pieces we don't have, sorted by priority, and shuffled within equal priorities. I wonder if this list instead should represent "piece groups". i.e. adjacent pieces of at least one piece, but maybe more.

src/request_blocks.cpp
```diff
@@ -116,12 +116,18 @@ namespace libtorrent {
 			&& !time_critical_mode
 			&& t.settings().get_int(settings_pack::whole_pieces_threshold) > 0)
 		{
+			// if our download rate lets us download a whole piece in
+			// "whole_pieces_threshold" seconds, we prefer to pick an entire piece.
+			// If we can download multiple whole pieces, we prefer to download that
+			// many contiguous pieces.
+			int const contiguous_pieces =
```
arvidn: this should probably have an upper bound, of 4 or 8 MiB

PR #3754, opened 2019-03-28
fix seeding of random number generator on mingw

arvidn:


sledgehammer999:
Out of curiosity: What's the issue with mingw here? Any links?

arvidn:
libstdc++ does not support windows entropy sources, so random_device is not random

PR #3765, opened 2019-04-02
don't run the SHA-1 and SHA-512 test vector tests by default

arvidn:
as they are slow


aldenml:
interesting, are they slower with `built-in` or it doesn't matter?

arvidn:
I didn't measure it that closely, but SHA-1 takes about 8 seconds of full CPU usage, and SHA-512 takes about 30 seconds

PR #3766, opened 2019-04-02
make get_download_queue() return the queue

arvidn:


aldenml:
Are you planning to deprecate the other one?

arvidn:
I'm not sure. do you think I should?
There may be edge cases where having to reallocate the vector for every call may be too expensive

aldenml:
Good point, I will take a look at the code to clarify something first

aldenml:
Hi @arvidn, I checked and I agree that it should stay for the cases where performance is a concern

PR #3768, opened 2019-04-04
optimize resolve_duplicate_filenames_slow()

arvidn:


src/torrent_info.cpp
```diff
@@ -640,58 +644,117 @@ namespace {
...
 			{
 				++cnt;
 				char new_ext[50];
 				std::snprintf(new_ext, sizeof(new_ext), ".%d%s", cnt, ext.c_str());
 				filename = base + new_ext;
+
+				boost::crc_optimal<32, 0x1EDC6F41, 0xFFFFFFFF, 0xFFFFFFFF, true, true> crc;
+				process_string_lowercase(crc, filename);
+				std::uint32_t const new_hash = crc.checksum();
+				if (files.find(new_hash) == files.end())
```
arvidn: I'm not super happy about this one. it's a bit sloppy to *just* check for a hash collision, and not also check whether the name collides. The idea is that it doesn't really matter that much if someone can construct names that will collide, since the worst case is that the sequence numbers appended to the filenames aren't contiguous.

and I'm pretty sure you can't trigger an infinite loop here, CRC32 is really good at covering all the bits, so you would need near `UINT32_MAX` file entries for that

arvidn:
the main cost previously was building the hash table of all existing files, especially if you had a very very long path, where each path segment would have to be added individually. Specifically, copying strings into the hash table was very costly. This solution is to not copy any strings, but just record which file or path the hash represents, and reproduce the string if we get a hash collision, which should be rare.

However, it's easy to deliberately produce hash collisions with CRC32, that's why I'm considering an upper limit on how many collisions to allow, before just rejecting the torrent as malicious

arvidn:
I should probably review the test cases for `resolve_duplicate_filenames` to see if any should be added

PR #3772, opened 2019-04-06
add fuzzers to the main repository

arvidn:


arvidn:
The python scripts to manage the corpus should also be added

arvidn:
I think this is ready to land now

PR #3782, opened 2019-04-15
change m_torrents to a vector and add a map for lookups on infohash

arvidn:
With v2, torrents can now have two infohashes. Simply adding multiple
    entries in m_torrents is undesirable because many code paths use
    m_torrents to iterate over every torrent. To keep m_torrents usable for
    iterating over, make it an unsorted vector storing all torrents exactly
    once.

    A new map is created to support fast lookups based on v1 or v2
    infohash.

@ssiloti does this look reasonable? I'm thinking that maybe I should experiment with `boost.multi_index` before I merge

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -2603,7 +2603,7 @@ namespace aux {
 				{
 					// now, disconnect a random peer
 					auto const i = std::max_element(m_torrents.begin(), m_torrents.end()
-						, [](torrent_array::value_type const& lhs, torrent_array::value_type const& rhs)
+						, [](std::shared_ptr<torrent> const& lhs, std::shared_ptr<torrent> const& rhs)
 						{ return lhs->num_peers() < rhs->num_peers(); });
 
 					if (m_alerts.should_post<performance_alert>())
@@ -3479,7 +3479,7 @@ namespace aux {
 					// every 90 seconds, disconnect the worst peers
 					// if we have reached the connection limit
 					auto const i = std::max_element(m_torrents.begin(), m_torrents.end()
-						, [] (torrent_array::value_type const& lhs, torrent_array::value_type const& rhs)
+						, [] (std::shared_ptr<torrent> const& lhs, std::shared_ptr<torrent> const& rhs)
 						{ return lhs->num_peers() < rhs->num_peers(); });
 
 					TORRENT_ASSERT(i != m_torrents.end());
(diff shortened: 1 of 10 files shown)


ssiloti:
LGTM. Once v2 support is introduced I think a couple of operations would be slightly awkward with Boost.MultiIndex:

- Adding the other infohash of a hybrid torrent which was originally only known by one of its infohashes. This would require removing the torrent from the multiIndex set and re-adding it. Otherwise this would break the container's invariant.
- Looking up a torrent by an infohash of unknown version. With the current approach this only requires one hash tree lookup. With a multiIndex set it would take up to two, one for each protocol version.

It still might be worth it though, I leave it up to you.

arvidn:
right. I think it could be done with having one *sequential* index and one *multi-hash* index. Hm, but you're right, an important property is that there are multiple entries in the hash index pointing to the same object in the container. And if you have multiple entries in the hashed index, you'll have multiple entries in the sequential index too, which clearly wouldn't work.

I'll experiment with wrapping the three containers into a class though, that can be unit tested and provide a bit better encapsulation.

I think `m_obfuscated_torrents` can be removed too, and have those hashes just be inserted in the normal index. Is that your impression too?

On a side note, the reason I brought this up is that I rebased the `v2` branch onto master, and encountered a few conflicts around this patch. There were a few other conflicts too. I have a branch with the rebase, a few fixes but still one failing unit test. I will probably land this before completing that rebase.

src/session_impl.cpp
```diff
@@ -5135,14 +5113,15 @@ namespace aux {
...
-		if (i == m_next_dht_torrent)
-			++m_next_dht_torrent;
-#endif
-		if (i == m_next_lsd_torrent)
-			++m_next_lsd_torrent;
 
-		m_torrents.erase(i);
+		auto array_iter = std::find(m_torrents.begin(), m_torrents.end(), i->second);
+		TORRENT_ASSERT(array_iter != m_torrents.end());
+		if (array_iter != m_torrents.end())
```
arvidn: actually, this should be `array_iter != m_torrents.end() - 1`, right?

arvidn: I don't think it's OK to self-swap

ssiloti: You're right, self-swap isn't safe here. We still wan to check for `array_iter != m_torrents.end()` though, because swapping with `end()` is definitely not safe and we don't want to pop some random torrent out of the vector. The `array_iter != m_torrents.end() - 1` check should be nested inside the current check and only cover the call to `std::swap`.

arvidn: Oh, i was under the impression the iterator cannot be end(), given the assert

ssiloti: Right, but asserts aren't typically enabled in release builds so I added the check to prevent a crash if the condition arises in the field. Perhaps that's a bit much.

PR #3792, opened 2019-04-20
fix windows build with boost 1.70

ssiloti:


arvidn:
Is this not built by CI btw? I wonder how it snuck through

ssiloti:
The AppVeyor builds use Boost 1.63 and this error doesn't show up until 1.70.

arvidn:
ah, right. of course

PR #3793, opened 2019-04-20
Track the mapped port for each NAT mapping transport

### LINKED ISSUE #3785: Failure on upnp port forwarding will reset the listening port to 0

**libtorrent version (or branch):**
latest master or 1.2 branch, tested with qbittorrent latest master
**platform/architecture:**
linux, ubuntu 18.04 LTS
**compiler and compiler version:**
GCC 7.3, C++14
**please describe what symptom you see, what you would expect to see instead and
how to reproduce it:**
Conventionally, end-users just leave the upnp function in enabled status to adapt arbitrary scenarios. However, since 1.2, the failure on upnp port forwarding (may result from no router, router upnp off, etc.) will reset the listening port to 0. This is done by:
https://github.com/arvidn/libtorrent/blob/8506f9350710db6276e7a28b90b960a6f466d722/src/upnp.cpp#L1103-L1104
where the positional argument `0` in calling `on_port_mapping` will inevitably reset the value of the listening port to 0:
https://github.com/arvidn/libtorrent/blob/8506f9350710db6276e7a28b90b960a6f466d722/src/session_impl.cpp#L5575-L5576
Since then, all announces to trackers will include `port=0`, which triggers the `invalid port` error message from trackers.

It should not reset the port to 0. If port reset, the end user has to manually turn on or off the upnp port forwarding function to avoid invalid port issue, which brings a certain inconvenience.

ssiloti:
Each transport needs to be tracked separately so that a failed mapping request
doesn't cause the listen port for a socket to be cleared to zero.

Fixes #3785

### DIFF
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -133,15 +133,15 @@ namespace aux {
 		only_outgoing
 	};
 
+	struct listen_port_mapping
+	{
+		port_mapping_t mapping = port_mapping_t{-1};
+		int port = 0;
+	};
+
 	struct listen_socket_t
 	{
-		listen_socket_t()
-		{
-			tcp_port_mapping[0] = port_mapping_t{-1};
-			tcp_port_mapping[1] = port_mapping_t{-1};
-			udp_port_mapping[0] = port_mapping_t{-1};
-			udp_port_mapping[1] = port_mapping_t{-1};
-		}
+		listen_socket_t() = default;
 
 		// listen_socket_t should not be copied or moved because
 		// references to it are held by the DHT and tracker announce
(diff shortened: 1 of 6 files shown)


src/session_impl.cpp
```diff
@@ -5408,7 +5406,7 @@ namespace aux {
 			// connections. We may be able to accept uTP connections though, so
 			// announce the UDP port instead
 			if (m_settings.get_int(settings_pack::proxy_type) != settings_pack::none)
-				return std::uint16_t(sock->udp_external_port);
+				return std::uint16_t(sock->udp_external_port());
 			else
 				return std::uint16_t(sock->tcp_external_port);
```
arvidn: this needs to be a function call too, right?

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -169,19 +169,35 @@ namespace aux {
...
-		// to be published to peers, since this is the port
-		// the client is reachable through.
-		int tcp_external_port = 0;
-		int udp_external_port = 0;
+		// tcp_external_port and udp_external_port return the port which
+		// should be published to peers/trackers for this socket
+		// If there are active NAT mappings the return value will be
+		// the external port returned by the NAT router, otherwise the
+		// local listen port is returned
+		int tcp_external_port()
```
arvidn: It looks like every call to these two functions immediately casts the results to `uint16_t`. Do you think it would make sense to just return that type right away?

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -169,19 +169,35 @@ namespace aux {
...
+			return 0;
+		}
 
 		// 0 is natpmp 1 is upnp
-		port_mapping_t tcp_port_mapping[2];
-		port_mapping_t udp_port_mapping[2];
+		// the order of these arrays determines the priorty in
+		// which their ports will be announced to peers
+		listen_port_mapping tcp_port_mapping[2];
+		listen_port_mapping udp_port_mapping[2];
```
arvidn: it occurs to me that this could probably be `aux::array<listen_port_mapping, 2, portmap_transport>`, and we would probably save some `static_cast`s. But that might be better as a separate patch

include/libtorrent/aux_/session_impl.hpp
ssiloti: There are a couple of uses without immediate casts in `session_impl::get_listen_port` and `session_impl::listen_port`. This hits on the larger issue of port numbers using a mix of `std::uint16_t` and `int`. My preference would be to make a separate PR to convert all port numbers to use `std::uint16_t`.

PR #3805, opened 2019-04-25
initial documentation for bundled support for fuzzing

arvidn:
@pauldreik does this look reasonable? I'm thinking more hand made corpus files could be put in the python script.

I also think the OSSFuzz project only should run "large" fuzz targets.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -59,6 +59,7 @@ DOCS_PAGES = \
   docs/projects.html              \
   docs/python_binding.html        \
   docs/tuning.html                \
+  docs/fuzzing.html               \
   docs/settings.rst               \
   docs/stats_counters.rst         \
   docs/troubleshooting.html       \
@@ -80,6 +81,7 @@ DOCS_PAGES = \
   docs/projects.rst               \
   docs/python_binding.rst         \
   docs/tuning.rst                 \
+  docs/fuzzing.rst                \
   docs/troubleshooting.rst        \
   docs/udp_tracker_protocol.rst   \
   docs/utp.rst                    \
--- a/docs/fuzzing.rst
+++ b/docs/fuzzing.rst
@@ -1,0 +1,77 @@
+==================
+fuzzing libtorrent
+==================
+
+.. include:: header.rst
+
+.. contents:: Table of contents
+  :depth: 1
+  :backlinks: none
+
+overview
+========
+
+Libtorrent comes with a set of fuzzers. They are not included in the distribution
+tar ball, instead download the `repository snapshot`_ or clone the repository_.
+
+The fuzzers can be found in the `fuzzers` subdirectory and come with a `Jamfile`
+to build them, and a `run.sh` bash script to run them.
+
+.. _`repository snapshot`: https://github.com/arvidn/libtorrent/releases
+.. _repository: https://github.com/arvidn/libtorrent
+
+building
+--------
+
+The fuzzers use clang's libFuzzer, which means they can only be built with clang.
+Clang must be configured in your `user-config.jam`, for example::
+
+	using clang : 7 : clang++-7 ;
+
+When building, you most likely want to stage the resulting binaries into a
+well known location. Invoke `b2` like this::
+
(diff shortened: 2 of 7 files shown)


pauldreik:
I like the documentation!
I like the split in large and small targets!
This pull request is about documentation, so I am all in favour of merging that!

But there is more - the generation of seed corpus.
I don't like the approach of generating the seed corpus with a python script. It had made sense if, as an example, xml was created dynamically, but now it's just putting strings into files. Why can't it just be files in the repository?
Also, does the python script require a special working directory to work as intended? in that case, it should be enforced somehow (inspecting argv[0], perhaps?).

As a side note, while working with both the seed corpus and corpora from other sources (like those downloaded from oss-fuzz), it would be great if one could easily generate coverage. I had the shell script that invoked the fuzzers with kcov, which I think is very easy to work with. The workflow is
- run kcov through some helper script, using the seed corpus (or what have you)
- view the results in the browser
- think about what needs to be added (that's how I came up with the weird xml cases) and add it
- repeat

Have a look at [fuzz.sh](https://github.com/pauldreik/libtorrent/blob/ossfuzz/fuzzers/fuzz.sh). It's not pretty but it worked for the above workflow.

pauldreik:
It would be great to have some kind of central script for working the the fuzzers, so it possible to easily verify that the corpus goes through without problems. It is perhaps even desirable to have a CI build which runs each fuzzer for say, 20 seconds, to make sure nothing obvious has broken. That way small mistakes could be caught immediately, without having to go through oss fuzz first.

arvidn:
> I don't like the approach of generating the seed corpus with a python script. It had made sense if, as an example, xml was created dynamically, but now it's just putting strings into files. Why can't it just be files in the repository?

I'm not excited about having sample inputs in the script either, and I expect to move them out. I just haven't decided how yet. I would like to avoid the temptation of keeping a large corpus in the repo, and if we go to the trouble of hand-generating good test cases, they should really be used by the unit tests as well.

You may notice that the seed corpus for `torrent_info` is just copied from the unit test cases. I think it probably would make sense to do something similar for xml test cases. (I'm just not ready to commit to that work quite yet).

Additionally, I have another fuzzer in the pipeline, which actually *does* generate a seed corpus in a python script, so I plan on extending this script in that patch.

> Also, does the python script require a special working directory to work as intended? in that case, it should be enforced somehow (inspecting argv[0], perhaps?).

It does. Maybe a better approach would be to deduce the directory from the location of the script. Maybe that's what you meant.

pauldreik:
> > I don't like the approach of generating the seed corpus with a python script. It had made sense if, as an example, xml was created dynamically, but now it's just putting strings into files. Why can't it just be files in the repository?
> 
> I'm not excited about having sample inputs in the script either, and I expect to move them out. I just haven't decided how yet. I would like to avoid the temptation of keeping a large corpus in the repo, and if we go to the trouble of hand-generating good test cases, they should really be used by the unit tests as well.
> 
> You may notice that the seed corpus for `torrent_info` is just copied from the unit test cases. I think it probably would make sense to do something similar for xml test cases. (I'm just not ready to commit to that work quite yet).

OK
> 
> Additionally, I have another fuzzer in the pipeline, which actually _does_ generate a seed corpus in a python script, so I plan on extending this script in that patch.
> 
> > Also, does the python script require a special working directory to work as intended? in that case, it should be enforced somehow (inspecting argv[0], perhaps?).
> 
> It does. Maybe a better approach would be to deduce the directory from the location of the script. Maybe that's what you meant.

I usually try to make my scripts either warn on being invoked from the wrong place, chdir to right place, or generate the output path correctly. Any method works. I think the ideal is such that the script can be invoked from anywhere in the tree and does the right thing.

pauldreik:
While I remember it: sometimes it is practical to let  fuzzers selects flags etc based on the input. Such cases make sense only for fuzzing, not unit tests. Those files are of interest to keep around separately since they are connected to the particular fuzzer. Of course you can keep them as strings inside the python script, even if I don't like it :-)

Issue #3808, opened 2019-04-26
python hash_for_piece can segfault


joriscarrier:
**Please provide the following information**

libtorrent version (or branch): RC_1_2
platform/architecture: MacOS / Debian python 3.7.x
compiler and compiler version: boost 1.69

Hi, 
when I create a torrent of similar torrents my python interpreter crash and pytest show me this:

`malloc: Incorrect checksum for freed object 0x7f89849bba00: probably modified after being freed.`
`Corrupt value: 0x4e9b44024e601d2b`
`malloc: *** set a breakpoint in malloc_error_break to debug`

Any Idea ?

Here my code:

```python
    file_storage = lt.file_storage()

    for ti in torrents_info:
        for _file in ti.files():
            file_storage.add_file(_file.path, _file.size)

    file_storage.set_name(name)

    create_torrent = lt.create_torrent(file_storage)

    for torrent_indice, ti in enumerate(torrents_info):
        create_torrent.add_similar_torrent(ti.info_hash())

        for file_indice, _file in enumerate(ti.files()):
            for piece_indice in range(ti.num_pieces()):
                create_torrent.set_hash(
                    piece_indice * torrent_indice,
                    ti.hash_for_piece(piece_indice)  # if I delete this line I don't get a segfault
                )
            create_torrent.set_file_hash(
                file_indice * torrent_indice,
                _file.filehash.to_bytes()
            )
```

arvidn:
what's the reason to pass in `piece_indice * torrent_indice`. That looks like it will pass in a piece index that's out-of-bounds, and hence undefined. If you build libtorrent in debug mode, I believe this would trigger an assert with a helpful error message.

Similarly, `file_indice * torrent_indice` looks like it will cause file indices past the last file as well.

joriscarrier:
thx & sorry

PR #3809, opened 2019-04-27
Test similar torrent

joriscarrier:
#3808

### DIFF
--- a/bindings/python/tests/conftest.py
+++ b/bindings/python/tests/conftest.py
@@ -1,0 +1,27 @@
+from uuid import uuid4
+import os
+
+import pytest
+import libtorrent as lt
+
+
+@pytest.yield_fixture
+def torrent_info(tmpdir):
+    uid = str(uuid4())
+    save_path = tmpdir.mkdir('save_path')
+    directory_source_file = save_path.mkdir(uid)
+    for _ in range(10):
+        source_file = directory_source_file.join(uid + str(_))
+        source_file.write(os.urandom(1024**2 * 10), mode='wb')  # 10M
+
+    fs = lt.file_storage()
+    lt.add_files(fs, str(directory_source_file))
+    ct = lt.create_torrent(fs)
+    lt.set_piece_hashes(ct, str(save_path))
+
+    yield lt.torrent_info(ct.generate())
+
+    try:
+        save_path.remove()
+    except Exception:
+        pass
--- a/bindings/python/tests/test_torrent.py
+++ b/bindings/python/tests/test_torrent.py
@@ -1,0 +1,37 @@
+import libtorrent as lt
+
+
+def test_simimlar_torrent(torrent_info):
+    from builtins import range
+
+    torrents_info = [torrent_info for _ in range(10)]
+
+    file_storage = lt.file_storage()
+
+    for ti in torrents_info:
+        for _file in ti.files():
+            file_storage.add_file(_file.path, _file.size)
+
+    file_storage.set_name('similar_torrent')
+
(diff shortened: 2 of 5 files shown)


arvidn:
If you want to use a different test framework, I think all tests should be transitioned over, and I don't think such change should be conflated with adding a test

arvidn:
What are the benefits of `pytest` over the unit test framework currently used?

bindings/python/tests/conftest.py
```diff
@@ -0,0 +1,27 @@
+from uuid import uuid4
+import os
+
+import pytest
+import libtorrent as lt
+
+
+@pytest.yield_fixture
+def torrent_info(tmpdir):
```
arvidn: what does this function do? it doesn't look like it's testing anything.

bindings/python/tests/test_torrent.py
```diff
@@ -0,0 +1,37 @@
...
+                )
+            create_torrent.set_file_hash(
+                file_indice * torrent_indice,
+                _file.filehash.to_bytes()
+            )
+
+    entry = create_torrent.generate()
+    ti = lt.torrent_info(entry)
+
+    assert ti.name() == 'similar_torrent'
```
arvidn: this test doesn't really seem to capture the logic that's being exercised here. surely you would need to ensure that the torrent has a field for similar torrents, and maybe some properties of that field, like which info-hashes are in there and no duplicates etc.

bindings/python/tests/test_torrent.py
```diff
@@ -0,0 +1,37 @@
...
+    create_torrent = lt.create_torrent(file_storage)
+
+    for torrent_indice, ti in enumerate(torrents_info):
+        create_torrent.add_similar_torrent(ti.info_hash())
+
+        for file_indice, _file in enumerate(ti.files()):
+            for piece_indice in range(ti.num_pieces()):
+                create_torrent.set_hash(
+                    piece_indice * torrent_indice,
+                    ti.hash_for_piece(piece_indice)
```
arvidn: this line is non-sensical. you're looking over the number of pieces of *some other* torrent, and assigning it to the torrent you're creating.

bindings/python/tests/test_torrent.py
```diff
@@ -0,0 +1,37 @@
...
+                )
+            create_torrent.set_file_hash(
+                file_indice * torrent_indice,
```
arvidn: this line is also non-sensical. you're looping over the files of *some other* torrent, and assigning to the one you're creating

bindings/python/tests/test_torrent.py
```diff
@@ -0,0 +1,37 @@
...
+
+    file_storage.set_name('similar_torrent')
+
+    create_torrent = lt.create_torrent(file_storage)
+
+    for torrent_indice, ti in enumerate(torrents_info):
+        create_torrent.add_similar_torrent(ti.info_hash())
+
+        for file_indice, _file in enumerate(ti.files()):
+            for piece_indice in range(ti.num_pieces()):
```
arvidn: what's the point of looping over all pieces once per file? the pieces are the same every time

arvidn:
it appears you have misunderstood how to create torrents and how "similar torrents" work. would you mind pointing out parts of the documentation that you find unclear, ambiguous, or hard to follow?

bindings/python/tests/conftest.py
joriscarrier: https://docs.pytest.org/en/latest/fixture.html

PR #3833, opened 2019-05-13
choke dishonest peer in anti-leech seeding algorithm

holymonson:
As discussed in https://github.com/qbittorrent/qBittorrent/issues/10258, some clients (e.g. XunLei) always claim they have zero pieces, even we have uploaded a lot to them. They take most advantage in the anti-leech seeding algorithm but won't share any pieces. Choke them.

### DIFF
--- a/src/choker.cpp
+++ b/src/choker.cpp
@@ -162,8 +162,8 @@ namespace {
 		TORRENT_ASSERT(t);
 
 		std::int64_t const total_size = t->torrent_file().total_size();
-		std::int64_t have_size = peer->num_have_pieces() * t->torrent_file().piece_length();
-		have_size = std::max(have_size, peer->statistics().total_payload_upload());
+		std::int64_t const have_size = std::max(peer->statistics().total_payload_upload()
+			, std::int64_t(t->torrent_file().piece_length()) * peer->num_have_pieces());
 		return int(std::abs((have_size - total_size / 2) * 2000 / total_size));
 	}
 
--- a/src/choker.cpp
+++ b/src/choker.cpp
@@ -59,6 +59,14 @@ namespace {
 		return 0;
 	}
 
+	// return true if peer is suspected of dishonesty
+	bool dishonest_peer(peer_connection const* peer, int piece_length)
+	{
+		// some clients (e.g. XunLei) always claim they have zero pieces as fresh starter,
+		// no matter how many we have uploaded.
+		return peer->num_have_pieces() == 0 && peer->uploaded_at_last_round() >= piece_length;
+	}
+
 	// return true if 'lhs' peer should be preferred to be unchoke over 'rhs'
 	bool unchoke_compare_rr(peer_connection const* lhs
 		, peer_connection const* rhs, int pieces)
@@ -168,6 +176,13 @@ namespace {
 		//   |             V             |
 		//   +---------------------------+
 		//   0%    num have pieces     100%
+
+		// the algorithm presupposes peers reporting their info honestly.
+		// those obviously cheating for a high score should be banned in advance.
+		bool const dishonest1 = dishonest_peer(lhs, t1->torrent_file().piece_length());
+		bool const dishonest2 = dishonest_peer(rhs, t2->torrent_file().piece_length());
+		if (dishonest1 != dishonest2) return !dishonest1;
+
 		int const t1_total = t1->torrent_file().num_pieces();
 		int const t2_total = t2->torrent_file().num_pieces();
 		int const score1 = (lhs->num_have_pieces() < t1_total / 2
--- a/src/choker.cpp
+++ b/src/choker.cpp
@@ -59,14 +59,6 @@ namespace {
 		return 0;
 	}
 
-	// return true if peer is suspected of dishonesty
-	bool dishonest_peer(peer_connection const* peer, int piece_length)
-	{
-		// some clients (e.g. XunLei) always claim they have zero pieces as fresh starter,
-		// no matter how many we have uploaded.
-		return peer->num_have_pieces() == 0 && peer->uploaded_at_last_round() >= piece_length;
-	}
-
 	// return true if 'lhs' peer should be preferred to be unchoke over 'rhs'
 	bool unchoke_compare_rr(peer_connection const* lhs
 		, peer_connection const* rhs, int pieces)
@@ -144,18 +136,8 @@ namespace {
 		return lhs->time_of_last_unchoke() < rhs->time_of_last_unchoke();
 	}
 
-	// return true if 'lhs' peer should be preferred to be unchoke over 'rhs'
-	bool unchoke_compare_anti_leech(peer_connection const* lhs
-		, peer_connection const* rhs)
+	int anti_leech_score(peer_connection const* peer)
 	{
-		int const cmp = compare_peers(lhs, rhs);
-		if (cmp != 0) return cmp > 0;
-
-		std::shared_ptr<torrent> const t1 = lhs->associated_torrent().lock();
-		std::shared_ptr<torrent> const t2 = rhs->associated_torrent().lock();
-		TORRENT_ASSERT(t1);
-		TORRENT_ASSERT(t2);
-
 		// the anti-leech seeding algorithm is based on the paper "Improving
 		// BitTorrent: A Simple Approach" from Chow et. al. and ranks peers based
 		// on how many pieces they have, preferring to unchoke peers that just
@@ -176,20 +158,24 @@ namespace {
 		//   |             V             |
 		//   +---------------------------+
 		//   0%    num have pieces     100%
+		std::shared_ptr<torrent> const t = peer->associated_torrent().lock();
+		TORRENT_ASSERT(t);
 
-		// the algorithm presupposes peers reporting their info honestly.
-		// those obviously cheating for a high score should be banned in advance.
-		bool const dishonest1 = dishonest_peer(lhs, t1->torrent_file().piece_length());
-		bool const dishonest2 = dishonest_peer(rhs, t2->torrent_file().piece_length());
-		if (dishonest1 != dishonest2) return !dishonest1;
+		std::int64_t const total_size = t->torrent_file().total_size();
+		std::int64_t have_size = peer->num_have_pieces() * t->torrent_file().piece_length();
+		have_size = std::max(have_size, peer->uploaded_at_last_round());
+		return int(std::abs((have_size - total_size / 2) * 2000 / total_size));
+	}
 
-		int const t1_total = t1->torrent_file().num_pieces();
-		int const t2_total = t2->torrent_file().num_pieces();
-		int const score1 = (lhs->num_have_pieces() < t1_total / 2
-			? t1_total - lhs->num_have_pieces() : lhs->num_have_pieces()) * 1000 / t1_total;
-		int const score2 = (rhs->num_have_pieces() < t2_total / 2
-			? t2_total - rhs->num_have_pieces() : rhs->num_have_pieces()) * 1000 / t2_total;
+	// return true if 'lhs' peer should be preferred to be unchoke over 'rhs'
+	bool unchoke_compare_anti_leech(peer_connection const* lhs
+		, peer_connection const* rhs)
+	{
+		int const cmp = compare_peers(lhs, rhs);
+		if (cmp != 0) return cmp > 0;
 
+		int const score1 = anti_leech_score(lhs);
+		int const score2 = anti_leech_score(rhs);
 		if (score1 != score2) return score1 > score2;
 
 		// prioritize the one that has waited the longest to be unchoked
--- a/src/choker.cpp
+++ b/src/choker.cpp
@@ -163,7 +163,7 @@ namespace {
 
 		std::int64_t const total_size = t->torrent_file().total_size();
 		std::int64_t have_size = peer->num_have_pieces() * t->torrent_file().piece_length();
-		have_size = std::max(have_size, peer->uploaded_at_last_round());
+		have_size = std::max(have_size, peer->statistics().total_payload_upload());
 		return int(std::abs((have_size - total_size / 2) * 2000 / total_size));
 	}
 
--- a/include/libtorrent/peer_connection.hpp
+++ b/include/libtorrent/peer_connection.hpp
@@ -645,6 +645,9 @@ namespace aux {
 		std::time_t last_seen_complete() const { return m_last_seen_complete; }
 		void set_last_seen_complete(int ago) { m_last_seen_complete = ::time(nullptr) - ago; }
 
+		std::int64_t uploaded_at_last_round() const
+		{ return m_uploaded_at_last_round; }
+
 		std::int64_t uploaded_in_last_round() const
 		{ return m_statistics.total_payload_upload() - m_uploaded_at_last_round; }
 
--- a/include/libtorrent/peer_connection.hpp
+++ b/include/libtorrent/peer_connection.hpp
@@ -645,9 +645,6 @@ namespace aux {
 		std::time_t last_seen_complete() const { return m_last_seen_complete; }
 		void set_last_seen_complete(int ago) { m_last_seen_complete = ::time(nullptr) - ago; }
 
-		std::int64_t uploaded_at_last_round() const
-		{ return m_uploaded_at_last_round; }
-
 		std::int64_t uploaded_in_last_round() const
 		{ return m_statistics.total_payload_upload() - m_uploaded_at_last_round; }
 


src/choker.cpp
```diff
@@ -59,6 +59,14 @@ namespace {
 		return 0;
 	}
 
+	// return true if peer is suspected of dishonesty
+	bool dishonest_peer(peer_connection const* peer, int piece_length)
+	{
+		// some clients (e.g. XunLei) always claim they have zero pieces as fresh starter,
+		// no matter how many we have uploaded.
+		return peer->num_have_pieces() == 0 && peer->uploaded_at_last_round() >= piece_length;
```
arvidn: I think you'll need to add some factor to the piece length here. I can think of three considerations:

1. The blocks we've uploaded may not have been part of the same piece, and the peer may still be stuck waiting for the last stragglers to complete those pieces.
2. There will be some delay between the peer completing the piece and us receiving its "HAVE" message. It will have to check its hash, maybe write it to disk and send back the message over the wire. There will be *at least* one round-trip time of delay, but probably more.
3. The peer may chose to batch HAVE messages it sends back, which will contribute to the latency we experience from uploading and receiving the HAVE notification.

Another consideration is that this heuristic is probably too simplistic. All it would take would be for xunlei (or whichever client is trying to cheat) would be to always report having 1 piece.

I think more accurate way of defeating this would be to remember how much we've uploaded to a peer and use the max of that and `num_have_pieces()` in this anti-leech choker.

I believe we already store the total amount of payload we've uploaded to a peer, and we preserve it between reconnects even.

ssiloti:
I don't think there's much value in targeting these ham-fisted attempts at making a pure leach client. If a mitigation like this PR ever became widespread, the offending clients would promptly switch to an alternative method. The obvious solution would be to simply never unchoke any peers (nor allow fast sets). The never unchoke method cannot be detected by individual clients, so we'd be right back were we started.

src/choker.cpp
holymonson: 1. Correct me if wrong, we prefer uploading blocks in the same piece, unless part of it has gotten from others. In that situation, peer would have at least 1 piece eventually.
2. Yes, it may take several rounds to comfirm (by trip delay and batching messages). But they will get nearly hightest score in next round, I think it's worth if we choke those pure leechers always higher than them. Even there is no cheater, innocently choked for several round-time won't hurt too much.

Anyway, I'm OK with an extra factor. @aresch , how many would you think is appropriate?

3. XunLei is the largest download company in China, which has it' own download network and cache servers. It won't share any, the high score by cheeting is merely a side-effect. I don't think it would likely change the cheating means.
If situation change, we can improve `dishonest_peer()` or even the algorithm in the future, for now it looks good enough to me.
4. @ssiloti , no algorithm is perfert, people choose this anti-leech choker because they want anti-leech, and this patch fixes a most obvious cheating loophole. For those believing other unchoke algorithm, it won't hurt anyway.

ssiloti:
@holymonson What good does closing a loophole do when there are similarly obvious loopholes which achieve the same result and are much harder to detect? This PR is putting a padlock on a door which has another unlocked door right next to it.

There are also downsides to implementing this:

1. Increased size and complexity of libtorrent. While the complexity increase from this feature may seem small, there are many such features one may want to add and the complexity adds up. Libtorrent is already a large and complex codebase, so the line must to drawn somewhere.
2. Increased protocol overhead. In a small way, the XunLei client's current behavior is doing the rest of the swarm a favor. Not sending HAVE messages saves on protocol overhead. If XunLei started sending HAVE messages while never unchoking any peers it would leave swarms slightly worse off than they are now.

holymonson:
@ssiloti If you don't like the anti-leech unchoking method, you can file a PR to remove it. As long as it exists, I think we should fix it, rather than worry about complexity of libtorrent, which is a different-scope topic. Let's be pragmatic and discuss within the implement of this PR.

As I said, the cheating method would unlikely change, no need to worry now. Even it does, the worse case is the patch becoming useless and we back to origin, then we drop it.
Except fresh starters may be chroked for several extra rounds, I don't see other cons. Unless someone is working on a better solution, I don't think this PR should be blocked.

arvidn:
a simpler and more robust approach is to compute the max of pieces the peer has and the number of pieces we've uploaded in total to the peer. The upload metric should be available for peers already.

holymonson:
@aresch , do you mean this?
```cpp
	bool dishonest_peer(peer_connection const* peer, int piece_length)
	{
		int const on_trip = 3;
		return (peer->num_have_pieces() + on_trip) * piece_length < peer->uploaded_at_last_round();
	}
```
But if peer loses too many data on transmission exceeding the `on_trip` buffer, he will be hard to catch up, especially we are the only seeder. And it seems not simpler and more robust then the prejudge `peer->num_have_pieces() == 0`.

arvidn:
No. I mean drop the concept of "dishonest peer" and just make sure that the current logic, instead of just looking at the number of pieces it has advertised to have, take the max of that and the number of pieces uploaded to the peer.

arvidn:
you can find the total uploaded by:

```
peer->statistics().m_statistics.total_payload_upload()
```

src/choker.cpp
```diff
@@ -168,13 +158,24 @@ namespace {
...
-		int const t2_total = t2->torrent_file().num_pieces();
-		int const score1 = (lhs->num_have_pieces() < t1_total / 2
-			? t1_total - lhs->num_have_pieces() : lhs->num_have_pieces()) * 1000 / t1_total;
-		int const score2 = (rhs->num_have_pieces() < t2_total / 2
-			? t2_total - rhs->num_have_pieces() : rhs->num_have_pieces()) * 1000 / t2_total;
+		std::shared_ptr<torrent> const t = peer->associated_torrent().lock();
+		TORRENT_ASSERT(t);
+
+		std::int64_t const total_size = t->torrent_file().total_size();
+		std::int64_t have_size = peer->num_have_pieces() * t->torrent_file().piece_length();
```
arvidn: this looks like it would overflow. `num_have_pieces()` returns an `int`, right? and I imagine `piece_length()` does. This multiplication may overflow. You could either divide `uploaded_at_last_round()` by the piece length, or widen one of the factors before multiplying.

Also, I prefer the style of writing this as a single expression, so `have_size` can be made `const`.

src/choker.cpp
```diff
@@ -168,13 +158,24 @@ namespace {
...
+		have_size = std::max(have_size, peer->uploaded_at_last_round());
+		return int(std::abs((have_size - total_size / 2) * 2000 / total_size));
```
arvidn: you changed the scale from `1000` to `2000`, was that deliberate? if so, what's the reason?

holymonson: To make the score falls into [0, 1000], neater then the origin [0, 500].

src/choker.cpp
```diff
@@ -168,13 +158,24 @@ namespace {
...
-		int const score1 = (lhs->num_have_pieces() < t1_total / 2
-			? t1_total - lhs->num_have_pieces() : lhs->num_have_pieces()) * 1000 / t1_total;
-		int const score2 = (rhs->num_have_pieces() < t2_total / 2
-			? t2_total - rhs->num_have_pieces() : rhs->num_have_pieces()) * 1000 / t2_total;
+		std::shared_ptr<torrent> const t = peer->associated_torrent().lock();
+		TORRENT_ASSERT(t);
+
+		std::int64_t const total_size = t->torrent_file().total_size();
+		std::int64_t have_size = peer->num_have_pieces() * t->torrent_file().piece_length();
+		have_size = std::max(have_size, peer->uploaded_at_last_round());
```
arvidn: why is `uploaded_at_last_round()` better than `statistics().total_payload_upload()`?

holymonson: `statistics().total_payload_upload()` is assigned to `m_uploaded_at_last_round` in `peer_connection::reset_choke_counters()` before unchoking, so they should be exactly the same.

Ryrynz:
Backport to 1.1.13?

xavier2k6:
@arvidn is this patch going to be backdated to 1.1.13 or a 1.1.14 release?

arvidn:
I'm not planning on doing so, but patches are welcome

xavier2k6:
@arvidn it seems this patch is not fully effective.
choker_broken

arvidn:
@xavier2k6 can you explain what that screenshot is showing? and what you would expect it to say instead? (and ideally why you would expect it to differ)

xavier2k6:
It is showing uploading(never downloading) to -XL0012- clients which remain at 0.00% & the basis of this PR was to remove the workaround rather than banning the client.

I don't know if there has been a workaround for your workaround found?! or if the expected behaviour is to upload for a certain period of time & then cut them off or whether that client should be cut-off from instantaneously.

arvidn:
The expected behavior is to always upload as long as there are peers interested in downloading. That's always the top-level rule. Unused bandwidth is wasted bandwidth.

Is there a reasonable rationale for banning a client that's staying at %0? (i.e. not sending redundant HAVE messages). I don't think I understand what the actual problem is. You're making some assumption about it being a bad idea to stay connected to some peers, but I'm not clear on why you think that's a bad idea. Every peer starts at 0%. If everyone reject such peers, the whole network breaks down.

The anti-leech choker will prefer to upload to peers that appear to behave well. Not sending redundant HAVE-messages isn't necessarily bad behavior/ libtorrent can be configured not to do that (since there typically isn't a need to send such messages). This patch makes the anti-leech choker work for clients that don't send redundant HAVE messages by also taking into account how many blocks have been uploaded to that peer.

For example, if you have more unchoke slots than peers, the unchoke-algorithm will never come into play, everyone will always be unchoked.

xavier2k6:
> As discussed in [qbittorrent/qBittorrent#10258](https://github.com/qbittorrent/qBittorrent/issues/10258), some clients (e.g. XunLei) always claim they have zero pieces, even we have uploaded a lot to them. They take most advantage in the anti-leech seeding algorithm but won't share any pieces. Choke them.

This was the original reason from OP & although I do believe this has worked in the past - have seen uploading recently to this client with it status of 0.00% never changing!

arvidn:
> > As discussed in qbittorrent/qBittorrent#10258, some clients (e.g. XunLei) always claim they have zero pieces, even we have uploaded a lot to them. They take most advantage in the anti-leech seeding algorithm but won't share any pieces. Choke them.
> 
> This was the original reason from OP & although I do believe this has worked in the past - have seen uploading recently to this client with it status of 0.00% never changing! & they NEVER upload/share - hence why i thought your work around may have been circumvented somehow...

I feel pretty confident that this has in fact been fixed by this patch. I have not seen any evidence suggesting that it has not.

Again, the symptom of this problem is that there are other peers, that are better upload candidates that are choked, in favor of the peers that "stay at 0%". The fact that peers with 0% is being uploaded to is *not* evidence of a problem.

The trait of "staying at 0%" is not in itself a problem, nor a signal indicating bad behavior. This used to be the default in libtorrent, to not send HAVE messages to peers that already had that piece, as it would be a waste of bandwidth. The trait of never uploading anything is also not necessarily a sign of bad behavior, it could just be that the peer doesn't have any piece that you're interested in, which is likely to be the case for any peer that just joined the swarm.

If you have any reason to believe that there is still a problem, after applying the patch of this PR, please be precise in what symptom you experience and what you would expect instead and why.

xavier2k6:
https://github.com/qbittorrent/qBittorrent/issues/10258#issuecomment-557846831

another user has confirmed what I've been saying, the workaround/patch was to stop -XL0012- clients from being complete leeches but doesn't seem to be fully working.

arvidn:
> you could be sharing a 1GiB file & upload 100% of that file to those clients & they will remain at 0.00% & will never share data/bandwidth.

Which is how bittorrent is supposed to work. If you're a seed, nobody will upload to you but everyone (you stay connected to) will download from you. This observation is not (by itself) indicative of a problem. The fact that they stay at 0% is also not a problem or an indication of misbehavior (as I've explained in previous posts).

If your aim is to disconnect/ban certain peers, a much more effective tool is to use the IP-filter. But if you start banning peers based on a vague idea of "misbehaving" or being "complete leeches", chances are pretty good you're have a *lot* of false positives, banning legitimate clients, crippling the network.

When you unchoke and upload to these peers, do you also have other peers that are interested and choked that you have uploaded less to?

That would be an indication of a problem. The anti-leech choker attempts to evenly distribute upload across peers, and also to penalize peers that are close to 50% complete (unrelated to whether reporting it or not).

xavier2k6:
The only issue I have is with the -XL0012- clients & them being a "dishonest peer" - all other peers share normally - I can download from them & vice-versa upload to them.

with the -XL0012- clients, these are leeches - you can upload to them all the damn day long & not seem to get 1kb back.

As I said previously this seemed to work previously.....they would only get a few kb's & nothing more but now they seem to be able to do what they were able to do which was fully leech before this patch was implemented.

>This PR is putting a padlock on a door which has another unlocked door right next to it.

I'll just have to resort to banning this client, I'd say. Thanks.

Chocobo1:
@xavier2k6 
>I'll just have to resort to banning this client, I'd say. Thanks.

Or you can search for "qBittorrent Enhanced Edition" which bans those clients by default.

Ryrynz:
I couldn't find any notes on specifically what was different with qBT EE but the installer provides notes.
For anyone reading this interested..

1. Auto Ban Xunlei, QQ, Baidu, Xfplay, DLBT and Offline downloader client
2. Temporary IP Filter Web API for advanced user
3. Logging Failed WebUI/API Login Attempts
4. Update Message if NEW version is available
5. Auto Ban Unknown Peer from China Option(Default: OFF)
6. Show Tracker Authentication Window(Default: ON)
7. Auto Update Public Trackers List(Default: OFF)
8. Multiple qBittorrent instances

arvidn:
@xavier2k6 

> As I said previously this seemed to work ....they would only get a few kb's & nothing more but now they seem to be able to do what they were able to do which was fully leech before this patch was implemented.

I missed this. You're saying that this patch caused a regression?

xavier2k6:
> @xavier2k6
> 
> > As I said previously this seemed to work ....they would only get a few kb's & nothing more but now they seem to be able to do what they were able to do which was fully leech before this patch was implemented.
> 
> I missed this. You're saying that this patch caused a regression?

No, I don't think this patch caused a regression - it worked, it's just that I believe the Xunlei -XL0012- clients have found another loophole/workaround as @ssiloti had mentioned above that they probably would.

@Chocobo1 Thanks for the suggestion - I had come across [qBittorrent Enhanced Edition](https://github.com/c0re100/qBittorrent-Enhanced-Edition/releases) long time ago but would rather stay for now with trusted than 3rd party. (I will test it out some time on a spare machine)

mr-cn:
So, it means we will share few data to those dishonest peers unless we have more unused bandwidth? If so, this may be a wiser way.

We just don't want to see those dishonests peer are downloading happily, while others are suffering extreme low speed.

arvidn:
Right, that's how the choker works, unless you've configured to always unchoke everyone (but that's not the default setting).

I think it's important to have a good understanding of what we mean by "dishonest". So far I have not seen a compelling definition. I don't think clients that don't send redundant HAVE messages should be penalized.

Does anyone have any other properties of a dishonest peer?

Is there evidence that they manipulate the clock they use in the uTP transport to get a larger share of upload capacity? This is something uTorrent has some protection against, but not libtorrent.

AoiSeaghdha:
I don't know if this patch is *(or was)* effective at fixing the issue, as I rarely see the XunLei client *(and when I do it's as a seeder)* enough to test, but I can answer this at least:
> I think it's important to have a good understanding of what we mean by "dishonest". So far I have not seen a compelling definition. I don't think clients that don't send redundant HAVE messages should be penalized.
> 
> Does anyone have any other properties of a dishonest peer?

The fear the other users are trying to express is that, as XunLei clients show 0% download, it also must mean they're not contributing to the swarm in any way as a seeder. Akin to a drain, they're worried that XunLei users will occupy all their upload bandwidth and then leave the swarm, acting as if Users were a simple filehosting service.

If enough XunLei clients connected to a swarm and acted in this manner, the fear is that this could kill the swarm *(whether by leading new peers to believe the torrent is dead as XunLei clients receive the pieces over them, by having the last few seeders believe they've "given enough" ratio back and stop seeding, or by some similar circumstance)*, and even if it did not kill the swarm it's seen as unfair that these XunLei clients are not contributing in any way, forcing Users to contribute more for longer.

---

That being said, I'm not familiar enough with the protocol nor with the XunLei client to know if that's a valid fear. For the former, is it possible for a client to upload pieces while simultaneously showing as 0%? For the latter, it's possible the XunLei client does allow seeding & that their users do indeed contribute, and this behaviour simply isn't visible as a seeder ourselves. That would need to be tested.

If they intentionally aren't seeding/helping the swarm, then I'd certainly consider that to be "dishonest", but I don't see any other way to mitigate the issue beyond **a)** the `total_payload_uploaded` code merged above *(at least, I believe it was)* or **b)** blanket-banning IP ranges originating from "problem" countries, e.g. China. However, the latter would punish users of said countries that do not use "dishonest" clients. At any rate, while it's wordy and I'm not the best at explaining things, I hope that helps clarify @arvidn.

Artoria2e5:
We might be seeing some more dishonest behavior in another kind of client, as reported in https://github.com/anacrolix/torrent/discussions/891. The big concerning things are:

1. Instead of going out to download stuff by leeching, the clients seem to be leeching for the purpose of just consuming bandwidth. Choking will still help allocate bandwidth to the right people, but the total bandwidth out might still make your carrier unhappy. Nevertheless, the choke will help somewhat.

2. Most of the suspects report near 0% HAVE, which is fine because that's similar to Xunlei behavior addressed by this PR. However, several of the screenshots give amounts like 25% and 37%, which would MASSIVELY increase the weight they get in our current implementation, as these are closer to 50%.

For the second point, a simple `std::max` will no longer suffice. I recommend doing something more verbose:

```C++
		// Calculate two scores: one based on what we've given, the other based on what they report having.
		std::int64_t const given_size = peer->statistics().total_payload_upload();
		std::int64_t const have_size = std::int64_t(t->torrent_file().piece_length()) * peer->num_have_pieces();

		int const given_score = std::abs((given_size - total_size / 2) * 2000 / total_size);
		int const have_score = std::abs((have_size - total_size / 2) * 2000 / total_size);

		// Use the more conservative score if we suspect the peer of being dishonest
		return (have_size < given_size) ? given_score : have_score;
```

Should I make that a PR?

PR #3838, opened 2019-05-18
don't redirect stdout when running in a debugger

ssiloti:


arvidn:
it's failing here: https://travis-ci.org/arvidn/libtorrent/jobs/534275344#L5804

PR #3844, opened 2019-05-31
Misc. fixes 

Chocobo1:
* Fix unsafe type narrowing
  MSVC was emitting warning C4244 before this commit.
* Enlarge read/write handler size
  This is required when building with MSVC 2019 v16.1.1, boost 1.70, OpenSSL 1.0.2r

### DIFF
--- a/src/storage_utils.cpp
+++ b/src/storage_utils.cpp
@@ -607,7 +607,7 @@ namespace libtorrent { namespace aux {
 		int ret = 0;
 		for (auto buf : bufs)
 		{
-			ret += buf.size();
+			ret += static_cast<int>(buf.size());
 			std::fill(buf.begin(), buf.end(), 0);
 		}
 		return ret;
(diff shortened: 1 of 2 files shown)


src/storage_utils.cpp
```diff
@@ -607,7 +607,7 @@ namespace libtorrent { namespace aux {
 		int ret = 0;
 		for (auto buf : bufs)
 		{
-			ret += buf.size();
+			ret += static_cast<int>(buf.size());
```
arvidn: I think `aux::numeric_cast<>` might be more appropriate here, as that also contains an assert that there is no overflow

Chocobo1: perhaps, I haven't really given thought to the overflow...

BTW the TORRENT_ASSERT looks weird there, shouldn't it be using `&&` instead of `||`?
```c++
TORRENT_ASSERT(std::is_unsigned<In>::value && std::is_signed<T>::value && std::int64_t(v) >= 0);
TORRENT_ASSERT(std::is_signed<In>::value && std::is_unsigned<T>::value && std::size_t(v) <= std::size_t((std::numeric_limits<T>::max)()));
```
https://github.com/arvidn/libtorrent/blob/RC_1_2/include/libtorrent/aux_/numeric_cast.hpp#L49-L52

Chocobo1: Never mind...

PR #3848, opened 2019-06-05
workaround a race condition when (un)mapping files on Windows

### LINKED ISSUE #3842: File corruption when downloading on Windows with sparse files

platform/architecture: Windows x64

compiler and compiler version: msvc-14.2

When downloading multi-file torrents on Windows with sparse files enabled files can be written incorrectly due to an apparent bug in Windows. I've reported the issue to MS [here](https://social.msdn.microsoft.com/Forums/windowsdesktop/en-US/44e9ddea-672f-441f-83ed-169f19fc48b8/concurrently-calling-createfilemapping-and-unmapviewoffile-with-a-mapped-sparse-file-causes-data?forum=windowsgeneraldevelopmentissues), but I'm not holding my breath for a fix. In any case we'll need to work around this by serializing opening/closing of mappings for the same file. I still need to figure out the best way to do this.

ssiloti:
Fixes #3842 

I'm not really happy with the amount of preprocessor noise added for this, but I don't see a clear alternative.

### DIFF
--- a/include/libtorrent/storage.hpp
+++ b/include/libtorrent/storage.hpp
@@ -194,6 +194,18 @@ namespace aux {
 		mutable std::mutex m_file_created_mutex;
 		mutable typed_bitfield<file_index_t> m_file_created;
 
+#if TORRENT_HAVE_MAP_VIEW_OF_FILE
+		// Windows has a race condition when unmapping a view while a new
+		// view or mapping object is being created in a different thread.
+		// The race can cause a page of written data to be zeroed out before
+		// it is written out to disk. To avoid the race these calls must be
+		// serialized on a per-file basis. See github issue #3842 for details.
+
+		// This array stores a mutex for each file in the storage object
+		// It must be aquired before calling CreateFileMapping or UnmapViewOfFile
+		mutable std::shared_ptr<std::mutex> m_file_open_unmap_lock;
+#endif
+
 		bool m_allocate_files;
 	};
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -106,7 +106,7 @@ build_script:
     mkdir build &&
     cd build &&
     set "PATH=c:\Python27-x64;%PATH%" &&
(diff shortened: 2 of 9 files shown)


arvidn:
https://ci.appveyor.com/project/arvidn/libtorrent/builds/25049823/job/7oqkspcsnwp9077x#L806

include/libtorrent/storage.hpp
```diff
@@ -194,6 +194,18 @@ namespace aux {
...
+#if TORRENT_HAVE_MAP_VIEW_OF_FILE
+		// Windows has a race condition when unmapping a view while a new
+		// view or mapping object is being created in a different thread.
+		// The race can cause a page of written data to be zeroed out before
+		// it is written out to disk. To avoid the race these calls must be
+		// serialized on a per-file basis. See github issue #3842 for details.
+
+		// This array stores a mutex for each file in the storage object
+		// It must be aquired before calling CreateFileMapping or UnmapViewOfFile
+		mutable std::shared_ptr<std::mutex[]> m_file_open_unmap_lock;
```
arvidn: storing arrays in `shared_ptr` is a c++17 feature. afaik, master currently builds with C++14. Do you think we should just go straight to 17?

arvidn:
it looks like the test needs to have a similar treatment to take a 3rd parameter. Patch looks reasonable, I can't see any obvious way of making it simpler/less noisy.

One way to make it simpler though, and cause more contention, would be to have a single mutex for the whole process (it could be a singleton even). That's assuming there aren't that many calls to `CreateFileMapping()` and `UnmapViewOfFile()`

arvidn:
actually, is there a reason the array of mutexes isn't owned by the `file_view_pool`? that would also just make it one instance per file

arvidn:
and the mutex wouldn't have to be passed around in the `file_view_pool` interface, like `open_file()`

arvidn:
the build failure happens in release builds apparently. Both the bjam and cmake builds. I have a feeling this is some msvc compiler quirk

include/libtorrent/storage.hpp
ssiloti: I believe the use of a custom deleter is enough to make this work with C++14, is it not?

arvidn: I think you may have to hide that it's an array too, like `std::shared_ptr<std::mutex>`, with the custom deleter knowing that it's an array.

arvidn:
it's quite unclear to me how this patch can have affected the callback closure size on macOS in `peer_connection.cpp`. Types should only have changed size and functions should only have gotten more arguments on windows/mingw.

arvidn:
would you mind rebasing this on master now? I fixed the handler sizes

PR #3859, opened 2019-06-27
set HTTP Host to request-target when using proxy

patch3proxyheaders915360:
To make the Host field carry the request-target it is enough to set
the Host information before changing the connection destination.

Many proxies rely on the Host field-value to route requests
in a violation of RFC7230 section 5.4.

https://tools.ietf.org/html/rfc7230#section-5.4

A request with the Host field set to the proxy hostname and port
in place of the request-target produces error responses and otherwise
confuses such misconfigured middleboxes.


patch3proxyheaders915360:
Here's a relevant bug discussion started by affected users https://github.com/qbittorrent/qBittorrent/issues/8972

arvidn:
this patch looks correct. but it should be made against `RC_1_2`, right?

arvidn:
it looks like some tests may have to be updated as well

patch3proxyheaders915360:
So what do I do, rebase it against RC_1_2 and file another PR?

arvidn:
you can actually update the PR to be against `RC_1_2`, but most likely you'll have to rebase the branch too

patch3proxyheaders915360:
I found that git-cherry-pick works better than rebase, but I can't change the head repository branch.

I decided to file a new PR. https://github.com/arvidn/libtorrent/pull/3863

PR #3863, opened 2019-06-30
set HTTP Host to request-target when using proxy

patch3proxyheaders915360:
To make the Host field carry the request-target it is enough to set
the Host information before changing the connection destination.

Many proxies rely on the Host field-value to route requests
in a violation of RFC7230 section 5.4.

https://tools.ietf.org/html/rfc7230#section-5.4

A request with the Host field set to the proxy hostname and port
in place of the request-target produces error responses and otherwise
confuses such misconfigured middleboxes.

### DIFF
--- a/src/http_connection.cpp
+++ b/src/http_connection.cpp
@@ -184,12 +184,12 @@ void http_connection::get(std::string const& url, time_duration timeout, int pri
 			request << "Proxy-Authorization: Basic " << base64encode(
 				ps->username + ":" + ps->password) << "\r\n";
 
-		hostname = ps->hostname;
-		port = ps->port;
-
 		request << "Host: " << hostname;
 		if (port != default_port) request << ":" << port << "\r\n";
 		else request << "\r\n";
+
+		hostname = ps->hostname;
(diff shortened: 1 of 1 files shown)


arvidn:
I'm a little bit worried about `m_hostname` being updated, causing the next attempt to still use the incorrect hostname. if a hostname resolves into multiple IPs, and the first request fails, it will try again with the next IP, and then the `http_connection` object preserves some state.

patch3proxyheaders915360:
Nye. Works like a charm on my ancient 1.1.x. Ten bogus addresses (fail by TCP reset, not 4xx or 5xx) and one real attached to a proxy's name. I looked in Wireshark and found the real IP was the third in the list given by the local DNS cache. HTTP Host was the desired one in an each little test case. Do I need to safeguard against future changes to ``http_connection`` or some other code disruptions?

arvidn:
thanks for testing!

Chocobo1:
Any plans to backport to RC_1_1?

arvidn:
I would welcome a PR!

PR #3871, opened 2019-07-04
feature to disable DHT, PEX and LSD per torrent

arvidn:


aldenml:
Hi! is this something that warrants more unit tests?

arvidn:
yes, probably. I was looking at the DHT tests, but it's non-trivial to make one for this. However, the torrent flags test should be extended at least

include/libtorrent/torrent.hpp
```diff
@@ -1548,7 +1548,11 @@ namespace libtorrent {
 		// whenever something is downloaded
 		bool m_need_save_resume_data:1;
 
-		// 2 bits here
+		// when this is true, this torrent participates in the DHT
+		bool m_enable_dht:1;
```
gubatron: style: why not keep it consistent with the flag name and avoid the negations everywhere else?

gubatron: this everywhere seems like a sign it should be named the other way:
`if (!m_enable_dht)`

:)

arvidn: well, I prefer to not have "built-in" negations in the names. The reason the flags have to have the negation is to be backwards compatible, since they won't be set by default in older versions.

PR #3873, opened 2019-07-04
Protocol version 2

ssiloti:


aldenml:
The use of `Copyright (c) ... BitTorrent Inc.` is interesting

include/libtorrent/peer_connection.hpp
```diff
@@ -432,6 +432,11 @@ namespace aux {
 		std::weak_ptr<torrent> associated_torrent() const
 		{ return m_torrent; }
 
+		// get the info hash associated with this peer
+		// this will be a sha1 hash or truncated sha256 hash depending
+		// on which protcol version this connection is using
```
aldenml: typo in `protcol`

aldenml:
(quite a big patch :)

@arvidn or @ssiloti, `info_hash_t` looks like a very important public type, is it possible to have an explanation for what's the rationale behind `get_best`? Is it for have a "unique id/number"? I feel the API could become a little confusing or even error prone with the truncated sha256, imo

ssiloti:
The `get_best` function is used to get a unique identifier for a torrent without having to worry about whether it is a v1 only, v2 only, or hybrid torrent. It needs to return a `sha1_hash` for compatibility with the status quo of using the v1 info-hash directly, hence v2 info-hashes must be truncated to fit.

arvidn:
https://travis-ci.org/arvidn/libtorrent/jobs/554373209#L849
https://travis-ci.org/arvidn/libtorrent/jobs/554373210#L1437
https://travis-ci.org/arvidn/libtorrent/jobs/554373213#L2660
https://travis-ci.org/arvidn/libtorrent/jobs/554373213#L2903
https://travis-ci.org/arvidn/libtorrent/jobs/554373214#L2424
https://travis-ci.org/arvidn/libtorrent/jobs/554373217#L644
https://travis-ci.org/arvidn/libtorrent/jobs/554373217#L708

this one may be unrelated to your patch: https://travis-ci.org/arvidn/libtorrent/jobs/554373211#L1110

ssiloti:
https://travis-ci.org/arvidn/libtorrent/jobs/554481893#L5561

This appears to be a bug in clang, https://bugs.llvm.org/show_bug.cgi?id=21629

arvidn:
@ssiloti I ended up putting my changes in a patch: https://github.com/ssiloti/libtorrent/pull/50

examples/client_test.cpp
```diff
@@ -1728,16 +1728,16 @@ COLUMN OPTIONS
 				for (lt::announce_entry const& ae : h.trackers())
 				{
 					auto best_ae = std::min_element(ae.endpoints.begin(), ae.endpoints.end()
-						, [](lt::announce_endpoint const& l, lt::announce_endpoint const& r) { return l.fails < r.fails; } );
+						, [](lt::announce_endpoint const& l, lt::announce_endpoint const& r) { return l.info_hashes[1].fails < r.info_hashes[1].fails; } );
```
arvidn: it's not obvious to me why info_hash 1 is used here. Is `info_hashes` guaranteed to always have 2 elements?

examples/dump_torrent.cpp
```diff
@@ -100,7 +142,9 @@ int main(int argc, char* argv[]) try
 		std::printf("%2d: %s\n", i.tier, i.url.c_str());
 
 	std::stringstream ih;
-	ih << t.info_hash();
+	ih << t.info_hash().v1;
+	if (t.info_hash().has_v2())
+		ih << ", " << t.info_hash().v2;
```
arvidn: should `info_hash_t` just do this in its `operator<<`?

examples/client_test.cpp
ssiloti: Yes, `info_hashes` has a fixed size of 2 with with first entry being for protocol v1 and the second v2.

examples/dump_torrent.cpp
ssiloti: The problem with defining `operator<<` for `info_hash_t` is that there is more than one thing it could do. Does it print one or both hashes as done here? Or only the "best" hash? Is the v2 hash truncated?

examples/client_test.cpp
arvidn: so, for a v1 torrent, everything under `info_hashes[1]` will be cleared, right? so this `min_element` won't do anything for v1 torrents, right?

include/libtorrent/alert_types.hpp
```diff
@@ -241,12 +241,12 @@ TORRENT_VERSION_NAMESPACE_2
...
 		// internal
 		torrent_removed_alert(aux::stack_allocator& alloc
-			, torrent_handle const& h, sha1_hash const& ih);
+			, torrent_handle const& h, info_hash_t const& ih);
 
 		TORRENT_DEFINE_ALERT_PRIO(torrent_removed_alert, 4, alert_priority_critical)
 		static constexpr alert_category_t static_category = alert::status_notification;
 		std::string message() const override;
-		sha1_hash info_hash;
+		info_hash_t info_hash;
```
arvidn: I wonder if it would be worth it to improve backwards compatibility here, by deprecating the `info_hash` member and adding the `info_hash_t` member under a different name.

`info_hash_t` *is* implicitly convertible to a `sha1_hash`, so in some cases the current patch would just work.

include/libtorrent/announce_entry.hpp
```diff
@@ -142,6 +126,51 @@ namespace libtorrent {
...
+
+		// the local endpoint of the listen interface associated with this endpoint
+		tcp::endpoint local_endpoint;
+
+		// torrents can be announced using multiple info hashes
+		// for different protocol versions
+
+		// info_hashes[0] is the v1 info hash (SHA1)
+		// info_hashes[1] is the v2 info hash (truncated SHA-256)
+		announce_infohash info_hashes[2];
```
arvidn: I wonder if it would make sense to make this `aux::array<announce_infohash, 2, protocol_version> info_hashes` instead. It might make it clearer which one is which version.

arvidn: I'll test this out

include/libtorrent/disk_interface.hpp
```diff
@@ -170,8 +173,13 @@ namespace file_open_mode {
 			, char const* buf, std::shared_ptr<disk_observer> o
 			, std::function<void(storage_error const&)> handler
 			, disk_job_flags_t flags = {}) = 0;
-		virtual void async_hash(storage_index_t storage, piece_index_t piece, disk_job_flags_t flags
+		// if v2 is non-empty it must be at least large enough to hold all v2 blocks in the piece
+		virtual void async_hash(storage_index_t storage, piece_index_t piece, span<sha256_hash> v2
+			, disk_job_flags_t flags
 			, std::function<void(piece_index_t, sha1_hash const&, storage_error const&)> handler) = 0;
+		// async_hash2 computes the v2 hash of a single block
+		virtual void async_hash2(storage_index_t storage, piece_index_t piece, int offset, disk_job_flags_t flags
```
arvidn: is `offset` a byte offset or block offset? (I think the comment should say, and in the future it should be part of the type). I think all other offsets in the disk I/O code refers to byte offsets as just "offset"

include/libtorrent/disk_io_job.hpp
```diff
@@ -151,7 +154,12 @@ namespace libtorrent {
 		{
 			un() {}
 			// result for hash jobs
-			sha1_hash piece_hash;
+			struct hash_args
```
arvidn: did you mean for this to be a union? or is the `piece_hash` used at the same time as `block_hashes`?

arvidn: as far as I can tell, they're not used at the same time, but I just did a very shallow search.

include/libtorrent/piece_picker.hpp
```diff
@@ -372,6 +380,10 @@ namespace libtorrent {
 		// or if we have the piece
 		bool is_piece_finished(piece_index_t index) const;
 
+		// returns true if at least one block in this piece is being hashed
+		// only valid for v2 torrents
+		bool is_hashing(piece_index_t piece) const;
```
arvidn: the piece_picker unit test should be extended to cover this

src/torrent.cpp
```diff
@@ -1492,7 +1510,7 @@ bool is_downloading_state(int const st)
 			m_save_path,
 			static_cast<storage_mode_t>(m_storage_mode),
 			m_file_priority,
-			m_info_hash
+			m_info_hash.has_v1() ? m_info_hash.v1 : m_info_hash.get(protocol_version::V2)
```
arvidn: this basically looks like "get_worst()". would it not work to just use `get_best()` instead?

src/torrent.cpp
```diff
@@ -1697,7 +1715,7 @@ bool is_downloading_state(int const st)
 
 			for (auto const& ih : m_torrent_file->similar_torrents())
 			{
-				std::shared_ptr<torrent> t = m_ses.find_torrent(ih).lock();
+				std::shared_ptr<torrent> t = m_ses.find_torrent(info_hash_t(ih)).lock();
```
arvidn: does this assume `ih` to be a SHA-1 hash? or would it work if similar-torrents had truncated SHA-256 hashes in it also? Whether similar torrents support v2 hashes or not should be deliberate I think

src/torrent.cpp
```diff
@@ -6102,6 +6501,37 @@ bool is_downloading_state(int const st)
 					ret.piece_priorities.push_back(m_picker->piece_priority(i));
 			}
 		}
+
+		if (m_torrent_file->info_hash().has_v2())
```
arvidn: I think there should be a test for the this. It could probably be as simple as loading a merkle tree and verified_leaf_hashes from resume data, then save resume data again and make sure we get back what we put in

src/torrent.cpp
```diff
@@ -6102,6 +6501,37 @@ bool is_downloading_state(int const st)
...
+				for (file_index_t f(0); f != m_torrent_file->files().end_file(); ++f)
+				{
+					if (m_torrent_file->files().pad_file_at(f))
+					{
+						ret.verified_leaf_hashes.emplace_back();
+						continue;
+					}
+					ret.verified_leaf_hashes.emplace_back(m_torrent_file->files().file_num_blocks(f));
+					std::fill(ret.verified_leaf_hashes.back().begin()
+						, ret.verified_leaf_hashes.back().end(), false);
```
arvidn: instead of a `std::fill()` here, couldn't the `emplace_back()` call just have a second argument of `false`?

src/hash_picker.cpp
```diff
@@ -0,0 +1,639 @@
...
+				m_merkle_trees[req.file][dest_start_idx - 1] = proof.first;
+				m_merkle_trees[req.file][dest_start_idx] = proof.second;
+			}
+			dest_start_idx = merkle_get_parent(dest_start_idx);
+		}
+
+		if (req.base == 0)
+		{
+			std::fill_n(m_hash_verified[req.file].begin() + req.index, req.count, true);
+			// TODO: add passing pieces to ret.hash_passed
```
arvidn: doing this is not critical for the hash_picker's function I take it?

src/hash_picker.cpp
```diff
@@ -0,0 +1,639 @@
...
+				if (done) continue;
+
+				merkle_fill_tree(m_merkle_trees[req.file], num_leafs, file_first_leaf + first_leaf);
+				if (m_merkle_trees[req.file][merkle_to_flat_index(base_layer_index, req.index + i)] != hashes[i])
+				{
+					merkle_clear_tree(m_merkle_trees[req.file], num_leafs / 2, merkle_get_parent(file_first_leaf + first_leaf));
+					m_merkle_trees[req.file][merkle_to_flat_index(base_layer_index, req.index + i)] = hashes[i];
+					TORRENT_ASSERT(num_leafs == m_files.piece_length() / default_block_size);
+					//verify_block_hashes(m_files.file_offset(req.file) / m_files.piece_length() + req.index);
+					// TODO: add to failed hashes
```
arvidn: this also sounds like an important thing to do

arvidn: this is the case where the block-hashes we got from a peer don't match what already know about the merkle tree. is that right?

examples/client_test.cpp
ssiloti: Right. I'm not sure why index 1 is hard coded here. At the very least it should be picking the index based on the best protocol version supported by the torrent.

include/libtorrent/disk_interface.hpp
ssiloti: `offset` is bytes from the start of the piece, same as the other disk I/O functions.

include/libtorrent/disk_io_job.hpp
ssiloti: Both `piece_hash` and `block_hashes` are used at the same time in hash jobs which compute both v1 and v2 hashes in the same job.

src/torrent.cpp
ssiloti: I was probably thinking of backwards compatibility when I wrote this. In hindsight that's not an issue here since hybrid torrents will have a different v1 info-hash from an equivalent v1 only torrent.

src/torrent.cpp
ssiloti: Right. Good catch.

src/hash_picker.cpp
ssiloti: This is for the case were the client downloaded piece data before it knew the corresponding piece hashes. Now it has the hashes and discovered that some of the piece data was bad.

src/torrent.cpp
ssiloti: Hybrid torrents are indexed by both their SHA1 and truncated SHA256 hashes, so `find_torrent` will work with both.

src/bt_peer_connection.cpp
```diff
@@ -1230,6 +1139,231 @@ namespace {
...
+		{
+			disconnect(errors::invalid_hash_reject, operation_t::bittorrent, peer_connection_interface::peer_error);
+			return;
+		}
+		if (!m_recv_buffer.packet_finished()) return;
+
+		std::shared_ptr<torrent> t = associated_torrent().lock();
+		TORRENT_ASSERT(t);
+
+		auto files = t->torrent_file().files();
```
arvidn: I think this should be `auto const& files = `, right? otherwise we'll get a copy of `file_storage`.

src/bt_peer_connection.cpp
```diff
@@ -1576,6 +1711,149 @@ namespace {
...
+		else
+		{
+			// comput the number of layers between the piece layer
+			// and the leaves
+			int blocks_per_piece = ti.piece_length() / (16 * 1024);
+			int layer_index = 0;
+			while (blocks_per_piece >>= 1) ++layer_index;
+			detail::write_uint32(layer_index, ptr);
+		}
+#endif
```
arvidn: we should decide what to do with this block I think. Is this here as a proposed future improvement? (if so, there should probably be a comment to that effect)

src/bt_peer_connection.cpp
```diff
@@ -1230,6 +1139,231 @@ namespace {
...
+			+ (count + proof_hashes) * int(sha256_hash::size()))
+		{
+			disconnect(errors::invalid_hashes, operation_t::bittorrent, peer_connection_interface::peer_error);
+			return;
+		}
+
+		if (!m_recv_buffer.packet_finished()) return;
+
+		auto new_end = std::remove(m_hash_requests.begin(), m_hash_requests.end(), hr);
+		m_hash_requests.erase(new_end, m_hash_requests.end());
```
arvidn: I think there's a risk here, that `hr` is not in `m_hash_requests`, and that will prevent us from sending more than two requests. Maybe that's fine, because this is just a single peer, and eventually maybe we'll disconnect the peer. and we can always request from other peers, right?

In fact, looking into this, it seems like we (probably) could replace the `m_hash_requests` vector with just a counter of outstanding hash requests. The only difference (as far as I can tell) would be how we handle the error cases of peers sending responses that we didn't request. But it doesn't seem important that it's handled by not requesting any more from the peer.

src/bt_peer_connection.cpp
```diff
@@ -3136,14 +3397,37 @@ namespace {
...
+				attach_to_torrent(info_hash_t(info_hash));
 				if (is_disconnecting()) return;
+
+				// this must go after the connection is attached to a torrent because that is what
+				// adds the peer info for incoming connections
+				if (recv_buffer[7] & 0x10)
+				{
+					if (!associated_torrent().lock()->info_hash().has_v2())
+					{
+						// the peer claims to support the v2 protocol with a non-v2 torrent
```
arvidn: is it not possible that the "v2" aspect of a torrent is stripped away via a magnet link? and that this could happen legitimately (i.e. non-maliciously)

src/create_torrent.cpp
```diff
@@ -650,42 +746,72 @@ namespace {
...
+							}
+						}
+					}
+
+					if (!file_e_ptr->dict().empty())
+					{
+						// path conflict
+						// there is already a directory with this name
+						// refuse to generate a torrent with such a conflict
+						// TODO: 3 report error
```
arvidn: I think an exception is reasonable here, and the case above. What do you think? I can prepare a patch

src/magnet_uri.cpp
```diff
@@ -75,10 +87,22 @@ namespace libtorrent {
...
+			ret += aux::to_hex(ih);
+		}
+
+		if (info.info_hash().has_v2())
+		{
+			if (info.info_hash().has_v1()) ret += '&';
+			sha256_hash const& ih = info.info_hash().v2;
+			ret += "xt=urn:btmh:1220";
+			ret += aux::to_hex(ih);
+		}
```
arvidn: there should be a test for `make_magnet_uri()` that creates a v2-only as well as a hybrid magnet URI

src/peer_connection.cpp
```diff
@@ -1255,7 +1267,8 @@ namespace libtorrent {
 			if (t && should_log(peer_log_alert::info))
 			{
 				peer_log(peer_log_alert::info, "ATTACH"
-					, "Delay loaded torrent: %s:", aux::to_hex(ih).c_str());
+					, "Delay loaded torrent: %s:"
+					, aux::to_hex(t->torrent_file().info_hash().get_best()).c_str());
```
arvidn: why isn't this just `ih.get_best()`?

src/peer_connection.cpp
```diff
@@ -2943,6 +2965,18 @@ namespace libtorrent {
...
+		if (t->info_hash().has_v2())
+		{
+			t->picker().started_hash_job(p.piece);
+			m_disk_thread.async_hash2(t->storage(), p.piece, p.start, {}
+				, [conn = self(), p](piece_index_t, sha256_hash const& h, storage_error const& e)
+			{
+				conn->wrap(&peer_connection::on_hash2_complete, e, p, h);
+			});
+		}
+#endif
```
arvidn: we should decide what to do with this block. I don't quite understand it at first glance, but it looks kind of important

src/piece_picker.cpp
```diff
@@ -1060,21 +1060,43 @@ namespace libtorrent {
...
+		{
+			auto const binfo = mutable_blocks_for_piece(*i);
+			for (int block : blocks)
+			{
+				block_info& info = binfo[block];
+				TORRENT_ASSERT(info.state >= block_info::state_writing);
+				if (info.state == block_info::state_writing)
+					--i->writing;
+				else if (info.state == block_info::state_finished)
+					--i->finished;
```
arvidn: don't you need to check for `block_info::state_requested` and decrement `i->requested` also?

arvidn: (this should be covered by tests in `test_piece_picker.cpp`)

src/read_resume_data.cpp
```diff
@@ -86,27 +86,34 @@ namespace {
 		}
 
 		auto info_hash = rd.dict_find_string_value("info-hash");
-		if (info_hash.size() != static_cast<std::size_t>(sha1_hash::size()))
+		auto info_hash2 = rd.dict_find_string_value("info-hash2");
+		if (info_hash.size() != std::size_t(sha1_hash::size())
+			&& info_hash2.size() != std::size_t(sha256_hash::size()))
 		{
 			ec = errors::missing_info_hash;
```
arvidn: this should be covered by additional tests in `test_read_resume.cpp`

src/sha1_hash.cpp
```diff
@@ -48,16 +48,32 @@ namespace libtorrent {
 		return os << aux::to_hex(peer);
 	}
 
+	std::ostream& operator<<(std::ostream& os, sha256_hash const& peer)
+	{
+		return os << aux::to_hex(peer);
+	}
+
 	// read 40 hexadecimal digits from an istream into a sha1_hash
```
arvidn: this comment should be updated

src/torrent_info.cpp
```diff
@@ -980,12 +1107,47 @@ namespace {
 		swap(m_info_section, ti.m_info_section);
 		swap(m_piece_hashes, ti.m_piece_hashes);
 		m_info_dict.swap(ti.m_info_dict);
-		swap(m_merkle_tree, ti.m_merkle_tree);
```
arvidn: the new member `m_merkle_trees` should be swapped here, right?

src/torrent_info.cpp
```diff
@@ -980,12 +1107,47 @@ namespace {
...
+				auto const leafs = merkle_num_leafs(orig_files().file_num_blocks(i));
+				f.resize(merkle_num_nodes(leafs));
+				f[0] = orig_files().root(i);
+			}
+		}
+
+		return m_merkle_trees;
+	}
+
+	aux::vector<sha256_hash>& torrent_info::file_merkle_tree(file_index_t file) const
```
arvidn: I notice this function is `const`. this works because `m_merkle_trees` is `mutable`. Is this the only reason `m_merkle_trees` has to be mutable? the function above, is *not* `const` (even though I suspect it could be).

If it's possible, I would prefer to make this non-const, and make `m_merkle_trees` non-mutable. Since `torrent_info` *is* shared between threads, this is a race condition asking to happen.

test/test_hash_picker.cpp
```diff
@@ -0,0 +1,548 @@
...
+	TEST_EQUAL(picked[1].count, 512);
+	TEST_EQUAL(picked[1].index, 1024);
+	TEST_EQUAL(picked[1].proof_layers, 10);
+	TEST_EQUAL(picked[2].file, 1_file);
+	TEST_EQUAL(picked[2].base, 0);
+	TEST_EQUAL(picked[2].count, 512);
+	TEST_EQUAL(picked[2].index, 1536);
+	TEST_EQUAL(picked[2].proof_layers, 10);
+}
+#endif
```
arvidn: is this test failing?

arvidn:
I think the default behaviour of `create_torrent` to create hybrid torrents will make all transfer tests exercise that case. It seem like there should be some transfer tests (i.e. end-to-end test).

arvidn:
I've gone through the whole patch now. Here are some of my notes of things I think are missing. I have a larger list of things I will propose as patches against the PR, and I will get to these eventually too:

* unit test for write_resume_data with "info-hash2"
* unit test making sure add_torrent_params::merkle_trees round-trips correctly through write_resume_data() and read_resume_data()
* unit tests for started_hash_job() and completed_hash_job() on piece_picker
* unit test for new argument to restore_piece() on piece_picker.
* unit test for is_hashing() on piece_picker
* unit test for v2 torrents with symlinks
* a review of the "smart_ban" logic for v2 torrents. Specifically, that a single failed block will immediately identify the peer sending the corrupt data. Maybe even a test. It might make sense to not instantiate the smart_ban extension object for v2 connections.
* a test_transfer test for v2-only torrents, in a few different scenarios. Especially where the hashes are missing on the downloader.
* a unit test exercising the m_priority_block_requests feature in the hash_picker
* the `#if 0`-blocks should be decided about, whether they should be kept or not
* the error cases in create_torrent::generate() should be handled
* unit test for make_magnet_uri() for a v2-only torrent

src/bt_peer_connection.cpp
ssiloti: This is obsolete, I'll remove it.

src/bt_peer_connection.cpp
ssiloti: This code is modeled on what the ut_metadata plugin does. If we don't stop sending requests when a peer fails to respond to a request then there is the potential to get in a loop where we keep sending a request for the same hashes over-and-over again without making any progress towards completion.

In the event of a failed peer, `hash_picker::pick_hashes` will re-send the request to a different peer once all hashes have been requested at least once.

src/bt_peer_connection.cpp
ssiloti: Good catch. This check should only be done if the torrent has valid metadata. I'll fix this.

src/create_torrent.cpp
ssiloti: sgtm

src/peer_connection.cpp
ssiloti: For hybrid torrents `ih` might only have the v1 hash while the torrent probably has both.

src/peer_connection.cpp
ssiloti: I originally intended to compute v2 block hashes incrementally as the blocks where downloaded, but I ended up changing to computing v2 hashes in the same job as the v1 hashes. Eventually I want to support setting deadlines with block granularity which will require downloading leaf hashes and checking individual blocks. This code remains as a reference for when I implement that.

src/piece_picker.cpp
ssiloti: This loop is resetting the state of blocks which have failed hash check so I don't think it's possible for them to be in `state_requested`, hence the assert than `info.state >= block_info::state_writing`. If it is possible for a failed block to be in `state_requested` then it should be left in that state and the request allowed to complete, hopefully the new data will turn out to be correct.

test/test_hash_picker.cpp
ssiloti: It needs a re-write because the signature of `pick_hashes` changed and it now only returns one hash request at a time.

src/peer_connection.cpp
arvidn: would you mind adding a comment in the code explaining this?

include/libtorrent/piece_picker.hpp
```diff
@@ -188,9 +188,14 @@ namespace libtorrent {
...
 			std::uint16_t requested:15;
 
-			// set to true while there is an outstanding
-			// hash check for this piece
-			std::uint16_t outstanding_hash_check:1;
+			// available for future use
+			std::uint16_t unused:1;
+
+			// number of outstanding hash jobs for this piece
+			std::uint16_t hashing:15;
```
arvidn: is there ever more than one outstanding hash job for the same piece at the same time?

arvidn: I suppose the user can call `add_piece()` at any time, to trigger overlapping calls to `torrent::verify_piece()`.
However, would it be unreasonable to, at the top of that function, check for `is_hashing()` and just exit if it's already has an outstanding hash job?

The reason I ask is because it would be really nice to save 16 bits here, and fit all the state in perfectly. This is one of those data structures that's instantiated *a lot* of times.

src/piece_picker.cpp
```diff
@@ -1060,21 +1060,43 @@ namespace libtorrent {
...
+				info.peer = nullptr;
+				info.state = block_info::state_none;
+			}
+			//p.download_state = piece_pos::piece_downloading;
+			update_piece_state(i);
+		}
+
+		if (blocks.empty() || i->requested + i->finished + i->writing == 0)
```
arvidn: if this block is not executed because one of these counters is non-zero, isn't there a risk that this piece will just get left open indefinitely?
It's not clear to me why it's safe to remove the piece under this condition for a v1 torrent, but not a v2 torrent. I would expect it to either be a problem today, or be fine to run this code unconditionally.

src/piece_picker.cpp
```diff
@@ -3470,7 +3535,7 @@ namespace {
 			if (i->finished < blocks_in_piece(i->index))
 				return;
 
-			if (i->passed_hash_check)
+			if (i->passed_hash_check && i->hashing == 0)
```
arvidn: is it possible for `passed_hash_check` to be true while `i->hashing` > 0?
I have a feeling that maybe this should be an assert instead.

arvidn: it seems that if this can happen, there need to be a check when we decrement `hashing` as well, to maybe call `we_have()` at that point.

include/libtorrent/hash_picker.hpp
```diff
@@ -0,0 +1,217 @@
...
+		aux::vector<aux::vector<sha256_hash>, file_index_t>& m_merkle_trees;
+		aux::vector<aux::vector<bool>, file_index_t> m_hash_verified;
+		aux::vector<aux::vector<piece_hash_request>, file_index_t> m_piece_hash_requested;
+		// blocks are only added to this list if there is a time critial block which
+		// has been downloaded but we don't have its hash or if the initial request
+		// for the hash was rejected
+		// this block hash will be requested from every peer possible until the hash
+		// is received
+		// the vector is sorted by the number of requests sent for each block
+		aux::vector<priority_block_request> m_priority_block_requests;
```
arvidn: I wonder if it would make sense to make this a `std::priority_queue<>`. that could easily be a later optimization too though.

src/hash_picker.cpp
```diff
@@ -0,0 +1,640 @@
...
+
+		int const blocks_per_piece = m_files.piece_length() / default_block_size;
+		piece_index_t const file_first_piece(int(m_files.file_offset(f) / m_files.piece_length()));
+		return { int(leafs_index - static_cast<int>(piece - file_first_piece) * blocks_per_piece)
+			, std::min(leafs_size, m_files.file_num_pieces(f) * blocks_per_piece - leafs_index) };
+	}
+
+	void hash_picker::hashes_rejected(hash_request const& req)
+	{
+		if (req.base == m_piece_layer && req.index % 512 == 0)
```
arvidn: I'm trying to add a test that exercises the *else* block in this condition. but I don't see how you can ever end up with a `has_request` that *doesn't* satisfy this case. My understanding is that all requests originate from line 237:

```
				return hash_request(fidx
					, m_piece_layer
					, i * 512
					, std::min(512, int(m_files.file_num_pieces(fidx) - i * 512))
					, layers_to_verify({ fidx, piece_tree_root }) + piece_tree_num_layers);
```

so, `req.base` is set to `m_piece_layer` and `req.index` is set to a multiple of 512. How is this condition ever false?

include/libtorrent/piece_picker.hpp
ssiloti: With the current state of the code there can only be one hash job per piece. This counter is to handle the future case where we support block level deadlines. That code will use `async_hash2` to compute an individual block hash and there may be multiple such jobs active at a time.

src/piece_picker.cpp
ssiloti: Calling `erase_download_piece` will cause the entire piece to be re-downloaded, which is exactly what we are trying to avoid when `blocks` is not empty. Block hashes are never known for a v1 torrent so the piece will always be erased, whereas for v2 torrents where block hashes are known we try to only reset the states of the failing blocks and otherwise leave the downloading piece as-is.

The counters check is assuming that at least one block has had its state reset to `state_none` so the piece should always be able to make progress by re-downloading those blocks.

arvidn: right, of course. typically we don't want to run this for v2 torrents, since we just reset the bad blocks

include/libtorrent/piece_picker.hpp
arvidn: would you mind if I propose a change to use a single bit for now, and wrap the counter code in `#if 0`, like the other logic for block level requests?

src/hash_picker.cpp
```diff
@@ -0,0 +1,640 @@
...
+		{
+			for (int i = req.index; i < req.index + req.count; i += 512)
+			{
+				m_piece_hash_requested[req.file][i / 512].last_request = min_time();
+				--m_piece_hash_requested[req.file][i / 512].num_requests;
+			}
+		}
+		else if (req.base == 0)
+		{
+			priority_block_request block_req(req.file, req.index);
```
arvidn: I must admit, I don't understand the logic here. Why are hash requests high priority under these conditions?
Is this also related to block-level requests for future streaming support?

arvidn:
I think this patch is in a good place now. @ssiloti would you mind squashing a few of the commits belonging together? I'd be happy to do it too.

ssiloti:
I'll take care of it. I know how much you like wrangling commits with git :)

arvidn:
https://ci.appveyor.com/project/arvidn/libtorrent/builds/26577613/job/7w5x5kaschyx3nu8#L530

The `info_hash_t` constructor that takes a `sha1_hash` is explicit. I think this line has to explicitly create an `info_hash_t` object.

arvidn:
did `v2_symlinks.torrent` get removed? `make dist` failed because of this file missing, I think.

PR #3877, opened 2019-07-14
add torrent_info constructor overloads to control torrent file limits

arvidn:


src/torrent_info.cpp
```diff
@@ -539,16 +547,21 @@ namespace {
 	}
 
 	int load_file(std::string const& filename, std::vector<char>& v
-		, error_code& ec)
+		, error_code& ec, int const max_buffer_size = 80000000)
```
FranciscoPombal: If `max_buffer_size` is forced to this value here, what is the point of setting the default to `6000000` in `torrent_info.hpp:119`? Also, is this supposed to be the file size limit in bytes? It should be at least configurable, like the piece limit and such.

examples/dump_torrent.cpp
```diff
@@ -58,40 +58,37 @@ std::vector<char> load_file(std::string const& filename)
...
 
 	std::vector<char> buf = load_file(argv[1]);
-	lt::bdecode_node e;
 	int pos = -1;
 	lt::error_code ec;
-	std::cout << "decoding. recursion limit: " << depth_limit
-		<< " total item count limit: " << item_limit << "\n";
-	int const ret = lt::bdecode(&buf[0], &buf[0] + buf.size(), e, ec, &pos
-		, depth_limit, item_limit);
+	std::cout << "decoding. recursion limit: " << cfg.max_decode_depth
```
FranciscoPombal: Shouldn't `cfg.max_pieces` also be printed here?

include/libtorrent/torrent_info.hpp
```diff
@@ -111,6 +111,20 @@ namespace libtorrent {
 	// used to disambiguate a bencoded buffer and a filename
 	extern TORRENT_EXPORT from_span_t from_span;
 
+	// this object holds configuration options for limits to use when loading
+	// torrents. They are meant to prevent loading potentially malicious torrents
+	// that cause excessive memory allocations.
+	struct load_torrent_limits
+	{
+		int max_buffer_size = 6000000;
```
FranciscoPombal: Is this supposed to be the file size limit in bytes?

examples/dump_torrent.cpp
```diff
@@ -58,40 +58,37 @@ std::vector<char> load_file(std::string const& filename)
...
+		<< " total item count limit: " << cfg.max_decode_tokens << "\n";
+	lt::bdecode_node const e = lt::bdecode(buf, ec, &pos, cfg.max_decode_depth
```
FranciscoPombal: Can you please explain the reason why `cfg.max_pieces` is also not passed here, like the others? Why is it not a field in `bdecode_node`?

For example, in `torrent_info:1376`, it is also passed separately:
```
bool torrent_info::parse_torrent_file(bdecode_node const& torrent_file
		, error_code& ec, int const piece_limit)
```

examples/dump_torrent.cpp
arvidn: well, I don't think any of these limits need to be printed. I just left these here because they were already being printed. I'm tempted to remove this print statement also, but that's a separate change.

examples/dump_torrent.cpp
arvidn: this is a call to a different function, `bdecode()`. All this function does is to parse the bdecoded structure. It doesn't do anything related to the number of pieces, just the recursion depth and number of tokens to parse.

The reason this example, `dump_torrent`, calls `bdecode()` itself, instead of just relying on the higher level `torrent_info` constructor, is that it also prints the raw bdecoded file first. In a way this is unnecessary, but it's really useful for debugging stuff. I should probably separate that out to a separate program though.

include/libtorrent/torrent_info.hpp
arvidn: yes. the .torrent-file size limit

src/torrent_info.cpp
arvidn: it's not forced, it's just the default, for all the calls to this function that doesn't specify it. It is configurable via that `load_torrent_limits::max_buffer_size`

FranciscoPombal: Ok, understood.

examples/dump_torrent.cpp
FranciscoPombal: Ok, thanks for the explanation

PR #3890, opened 2019-07-24
improve file_storage::sanitize_symlinks

arvidn:
What prompted this change was that I realized a MacOS .framework directory did not work in a .torrent file. The reason is that it has a concrete directory `Versions/A`, and a symlink `Versions/Current` pointing to it. Then it has shortcut symlinks in the root to `Versions/Current/Headers` etc.

In other words, it has (valid) symlinks whose targets don't exist in the torrent. The only way to validate that they are correct is to traverse the other symlinks. This is not trivial unfortunately, especially as I also would like to keep the common case (where there are no symlinks) cheap.

The current solution to this is to run two passes. One where "simple" symlinks are validated (i.e. ones that just point to a concrete file or directory), and all symlinks that potentially point to directories are put in a map.

Any symlinks that couldn't be validated in the first pass, are gone over in a second pass, where paths are deconstructed and validated one directory element at a time.


include/libtorrent/file_storage.hpp
```diff
@@ -446,7 +446,7 @@ namespace libtorrent {
 		// where this file starts. It can be used to map the file to a piece
 		// index (given the piece size).
 		sha1_hash hash(file_index_t index) const;
-		std::string const& symlink(file_index_t index) const;
```
arvidn: This also alters the abi

PR #3893, opened 2019-07-26
fix dht_stats_alert routing table stats for multi-homed clients

arvidn:
(temporary hack for 1.2.x series)


arvidn:
@ssiloti I'm thinking that really `dht_status_alert` needs to be extended to report stats for all tables individually, but that would be an ABI break for 1.2

arvidn:
https://travis-ci.org/arvidn/libtorrent/jobs/564244815#L1746
https://travis-ci.org/arvidn/libtorrent/jobs/564244817#L1540

PR #3894, opened 2019-07-28
factor out and unit test parts of the DHT routing table logic

arvidn:
the part that handles the affinity of nodes to insert. Specifically, make sure the even distribution within routing table buckets works correctly

### DIFF
--- a/src/kademlia/routing_table.cpp
+++ b/src/kademlia/routing_table.cpp
@@ -108,7 +108,139 @@ bool mostly_verified_nodes(bucket_t const& b)
 	return num_verified >= static_cast<int>(b.size()) * 2 / 3;
 }
 
-routing_table::routing_table(node_id const& id, udp proto, int bucket_size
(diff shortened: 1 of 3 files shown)


src/kademlia/routing_table.cpp
```diff
@@ -108,6 +108,138 @@ bool mostly_verified_nodes(bucket_t const& b)
...
+	// yet, so it will contain entries from both "sides", so we need to include
+	// the top bit.
+	nid <<= bucket_idx + int(!last_bucket);
+	std::uint8_t const ret = (nid[0] & mask) >> mask_shift;
+	TORRENT_ASSERT_VAL(ret >= 0, ret);
+	TORRENT_ASSERT_VAL(ret < bucket_size, ret);
+	return ret;
+}
+
+// TODO: unit test!
```
arvidn: this needs to be done

PR #3903, opened 2019-07-31
enhance API: Clear trackers

ajax16384:
allow clear torrent_info trackers list

### DIFF
--- a/include/libtorrent/torrent_info.hpp
+++ b/include/libtorrent/torrent_info.hpp
@@ -268,6 +268,7 @@ namespace libtorrent {
 		// number. For more information, see announce_entry_.
 		void add_tracker(std::string const& url, int tier = 0);
 		std::vector<announce_entry> const& trackers() const { return m_urls; }
+		void clear_trackers();
 
 		// These two functions are related to `BEP 38`_ (mutable torrents). The
 		// vectors returned from these correspond to the "similar" and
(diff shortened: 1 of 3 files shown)


include/libtorrent/torrent_info.hpp
```diff
@@ -268,6 +268,7 @@ namespace libtorrent {
 		// number. For more information, see announce_entry_.
 		void add_tracker(std::string const& url, int tier = 0);
 		std::vector<announce_entry> const& trackers() const { return m_urls; }
+		void clear_trackers();
```
arvidn: could you include `clear_trackers()` in the documentation above as well?

src/torrent_info.cpp
```diff
@@ -1525,6 +1525,10 @@ namespace {
 			{ return lhs.tier < rhs.tier; });
 	}
 
+	void torrent_info::clear_trackers() {
```
arvidn: please put the open brace on the next line

arvidn:
could you please add a unit test for `clear_trackers()` to `test/test_torrent_info.cpp` too?

include/libtorrent/torrent_info.hpp
ajax16384: Ok

src/torrent_info.cpp
ajax16384: Ok

ajax16384:
Ok

PR #3905, opened 2019-07-31
enhance API: Force LSD announce

ajax16384:
allow to force torrent LSD announce

### DIFF
--- a/include/libtorrent/torrent_handle.hpp
+++ b/include/libtorrent/torrent_handle.hpp
@@ -1044,8 +1044,12 @@ namespace aux {
 		//
 		// ``force_dht_announce`` will announce the torrent to the DHT
 		// immediately.
+		//
+		// ``force_lsd_announce`` will announce the torrent on LSD
+		// immediately.
 		void force_reannounce(int seconds = 0, int tracker_index = -1, reannounce_flags_t = {}) const;
 		void force_dht_announce() const;
+		void force_lsd_announce() const;
 
 #if TORRENT_ABI_VERSION == 1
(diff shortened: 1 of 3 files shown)


include/libtorrent/torrent_handle.hpp
```diff
@@ -1044,8 +1044,12 @@ namespace aux {
 		//
 		// ``force_dht_announce`` will announce the torrent to the DHT
 		// immediately.
+		//
+		// ``force_lsd_announce`` will announce the torrent on LSD
+		// immediately.
```
arvidn: Do you thin it's worth mentioning that this is ignored if:

1. the torrent is in *checking* state
2. LSD is not enabled on the session
3. the torrent has LSD disabled (via `torrent_flags`)
4. the torrent is "private"
5. it's an i2p torrent (and mixed mode is not enabled)
6. the torrent is paused

?
I suppose the other calls aren't documented to this detail, but it's never too late to raise the bar

ajax16384: it has same rule of execution as force_dht_announce - it will do nothing if some condition are not satisfied

arvidn: "lsd" is not in the spell checking dictionary. please spell it "LSD". https://travis-ci.org/arvidn/libtorrent/jobs/566118037#L5568

arvidn: it seems the spell checker is still expecting it to be lower case, I'm not sure why. maybe it picks up part of the function name.

PR #3907, opened 2019-08-01
update more GCC warnings to suppress for 3rd party headers

arvidn:


arvidn:
Well, at least it builds now. The unit tests still have to be made more reliable

PR #3909, opened 2019-08-02
unban 51.x.x.x from dht darknet list

ajax16384:
according to geoip isp database 51.x.x.x network is using by normal ISP


arvidn:
I'll merge this in from `RC_1_2` later

PR #3911, opened 2019-08-02
fixing appveyor test

arvidn:


arvidn:
I don’t really understand why the output redirection of the python script works. It’s possible python just fails to launch, but it seems like createprocess succeeds.

PR #3913, opened 2019-08-05
add feature to create an affinity to pick adjecent pieces

arvidn:
aligned to 4MiB extents. It's an attempt to improve disk I/O, by writing larger contiguous ranges of bytes. It's off by default.


src/settings_pack.cpp
```diff
@@ -208,6 +208,7 @@ constexpr int CLOSE_FILE_INTERVAL = 0;
 		SET(proxy_tracker_connections, true, nullptr),
 		SET(enable_ip_notifier, true, &session_impl::update_ip_notifier),
 		SET(dht_prefer_verified_node_ids, true, &session_impl::update_dht_settings),
+		SET(piece_extent_affinity, false, nullptr),
```
arvidn: it's off by default

PR #3916, opened 2019-08-07
Add dht_live_nodes and dht_sample_infohashes to Python bindings

fonic:
Add `dht_live_nodes` and `dht_sample_infohashes` functionality and `dht_bootstrap_alert` to Python bindings:
- add function `dht_live_nodes` to session
- add function `dht_sample_infohashes` to session
- add alert `dht_live_nodes_alert` to alerts
- add alert `dht_sample_infohashes_alert` to alerts
- add alert 'dht_bootstrap_alert' to alerts

**I'm still having some trouble with `dht_sample_infohashes_alert` (help requested):**
- property `endpoint` only contains garbage when inspected in Python; not sure if that's an issue with my code or existing code
- property `interval` throws a missing converter error in Python, however a converter for `time_duration` exists in `datetime.cpp`, so this should work, but doesn't; what am I missing?

And thanks by the way for an absolutely great library!

### DIFF
--- a/bindings/python/src/alert.cpp
+++ b/bindings/python/src/alert.cpp
@@ -135,6 +135,34 @@ dict session_stats_values(session_stats_alert const& alert)
     return d;
 }
 
+list dht_live_nodes_nodes(dht_live_nodes_alert const& alert)
+{
+    list result;
+    std::vector<std::pair<sha1_hash, udp::endpoint>> const nodes = alert.nodes();
+    for (std::pair<sha1_hash, udp::endpoint> const& node : nodes)
+    {
+        dict d;
+        d["nid"] = node.first;
+        d["endpoint"] = node.second;
+        result.append(d);
+    }
+    return result;
+}
+
+list dht_sample_infohashes_nodes(dht_sample_infohashes_alert const& alert)
+{
+    list result;
+    std::vector<std::pair<sha1_hash, udp::endpoint>> const nodes = alert.nodes();
+    for (std::pair<sha1_hash, udp::endpoint> const& node : nodes)
+    {
+        dict d;
+        d["nid"] = node.first;
+        d["endpoint"] = node.second;
+        result.append(d);
+    }
+    return result;
+}
+
 #if TORRENT_ABI_VERSION == 1
 entry const& get_resume_data_entry(save_resume_data_alert const& self)
 {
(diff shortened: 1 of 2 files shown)


bindings/python/src/alert.cpp
```diff
@@ -1029,6 +1057,27 @@ void bind_alert()
...
+        .add_property("nodes", &dht_live_nodes_nodes)
+        ;
+
+    std::vector<sha1_hash> (dht_sample_infohashes_alert::*samples)() const = &dht_sample_infohashes_alert::samples;
+
+    class_<dht_sample_infohashes_alert, bases<alert>, noncopyable>(
+       "dht_sample_infohashes_alert", no_init)
+        //.add_property("endpoint", &dht_sample_infohashes_alert::endpoint)
+        .add_property("endpoint", make_getter(&dht_sample_infohashes_alert::endpoint, by_value())) // TODO: should be correct like this, but in Python endpoint contains only garbage
+        //.add_property("interval", &dht_sample_infohashes_alert::interval) // TODO: throws error in Python; there is a converter for 'time_duration' in 'datatime.cpp', so this should work, but doesn't; type/class: dht_sample_infohashes_alert::interval == time_duration == std::chrono::nanoseconds
```
arvidn: iirc, in order to invoke the converter, you have to expose this field like this:
`.add_property("interval", make_getter(&dht_sample_infohashes_alert::interval), by_value())`

bindings/python/src/alert.cpp
```diff
@@ -1029,6 +1057,27 @@ void bind_alert()
...
+        .add_property("num_nodes", &dht_live_nodes_alert::num_nodes)
+        .add_property("nodes", &dht_live_nodes_nodes)
+        ;
+
+    std::vector<sha1_hash> (dht_sample_infohashes_alert::*samples)() const = &dht_sample_infohashes_alert::samples;
+
+    class_<dht_sample_infohashes_alert, bases<alert>, noncopyable>(
+       "dht_sample_infohashes_alert", no_init)
+        //.add_property("endpoint", &dht_sample_infohashes_alert::endpoint)
+        .add_property("endpoint", make_getter(&dht_sample_infohashes_alert::endpoint, by_value())) // TODO: should be correct like this, but in Python endpoint contains only garbage
```
arvidn: this is supposed to invoke the `endpoint_to_tuple` converter, right? when you say "garbage", do you at least get a tuple?

bindings/python/src/alert.cpp
```diff
@@ -135,6 +135,34 @@ dict session_stats_values(session_stats_alert const& alert)
...
 
+list dht_live_nodes_nodes(dht_live_nodes_alert const& alert)
+{
+    list result;
+    std::vector<std::pair<sha1_hash, udp::endpoint>> const nodes = alert.nodes();
+    for (std::pair<sha1_hash, udp::endpoint> const& node : nodes)
+    {
+        dict d;
+        d["nid"] = node.first;
+        d["endpoint"] = node.second;
```
arvidn: does this correctly convert the endpoint to a tuple?

arvidn:
is there an easy way of testing these new bindings in `test.py`?

bindings/python/src/alert.cpp
arvidn: would you mind testing this change?

fonic:
@arvidn: I'm currently on vacation, I'll continue and finish this once I'm back (about a week from now).

bindings/python/src/alert.cpp
fonic: works!

bindings/python/src/alert.cpp
fonic: I ported the patch to libtorrent 1.2.5 and now it works flawlessly. I guess some other bugfix fixed this as well.

bindings/python/src/alert.cpp
fonic: See comment above.

arvidn:
would you mind rebasing your branch on to of latest `RC_1_2`?

fonic:
Done, I hope. I hate rebasing - if you only do it once in a while, you just don't feel comfortable doing it. I need more practice ;)

fonic:
Added another alert (`dht_bootstrap_alert`).

Also, added node_id required for PR #4517.

arvidn:
I meant to have landed this. sorry about the delay. However, you now added `node_id` to the `dht_sample_infohashes_alert` binding, which doesn't exist, so I don't think this can land now.
adding `dht_bootstrap_alert` seems good though.

fonic:
Sorry, I'l then remove the `node_id` and rebase once again, ok?

arvidn:
yes. I imagine this branch doesn't build for you as is either. https://travis-ci.org/github/arvidn/libtorrent/jobs/672436576#L642

fonic:
This should be it.

PR #3933, opened 2019-08-12
post a dht_stats_alert per node we have running

arvidn:
Include node ID and local socket address


ssiloti:
The python bindings for `dht_stats_alert` need to be updated to expose `nid` and `local_endpoint`.

PR #3939, opened 2019-08-17
fix undefined behavior in torrent_info where an offset was applied to…

arvidn:
… pointers to move them from one object to another

### DIFF
--- a/src/bdecode.cpp
+++ b/src/bdecode.cpp
@@ -408,6 +408,14 @@ namespace detail {
 		return {m_buffer + t.offset, static_cast<std::ptrdiff_t>(next.offset - t.offset)};
 	}
 
+	std::ptrdiff_t bdecode_node::data_offset() const noexcept
+	{
+		TORRENT_ASSERT_PRECOND(m_token_idx != -1);
+		if (m_token_idx == -1) return -1;
+		bdecode_token const& t = m_root_tokens[m_token_idx];
+		return t.offset;
+	}
+
 	bdecode_node bdecode_node::list_at(int i) const
 	{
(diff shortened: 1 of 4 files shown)


arvidn:
http://www.cs.utah.edu/~regehr/oopsla18.pdf

src/bdecode.cpp
```diff
@@ -408,6 +408,14 @@ namespace detail {
 		return {m_buffer + t.offset, static_cast<std::ptrdiff_t>(next.offset - t.offset)};
 	}
 
+	std::ptrdiff_t bdecode_node::data_offset() const noexcept
```
arvidn: this needs a unit test

src/bdecode.cpp
```diff
@@ -535,10 +543,18 @@ namespace detail {
 		TORRENT_ASSERT(tokens[token].type != bdecode_token::end);
 
 		return std::make_pair(
-			bdecode_node(tokens, m_buffer, m_buffer_size, token).string_value()
+			bdecode_node(tokens, m_buffer, m_buffer_size, token)
 			, bdecode_node(tokens, m_buffer, m_buffer_size, value_token));
 	}
 
+	std::pair<string_view, bdecode_node> bdecode_node::dict_at(int const i) const
```
arvidn: this needs a unit test

src/bdecode.cpp
```diff
@@ -692,6 +708,14 @@ namespace detail {
 		return string_view(m_buffer + t.offset + t.start_offset(), size);
 	}
 
+	std::ptrdiff_t bdecode_node::string_offset() const
```
arvidn: and this

PR #3945, opened 2019-08-19
fix regression in left= reporting to trackers when adding a seeding t…

arvidn:
…orrent

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* fix regression in &left= reporting when adding a seeding torrent
 	* fix integer overflow in http parser
 	* improve sanitation of symlinks, to support more complex link targets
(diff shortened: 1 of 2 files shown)


an0n666:
Can this be backported to RC1_1? I’m facing similar issue with libtorrent 1.1.14. 
After restart, seeding torrent doesn’t report left = 0. Have to stop and start to make it work again.

arvidn:
@arvidn do you know if the patch applies cleanly to `RC_1_1`?

arvidn:
if it does, I imagine it would be simple to back-port. Testing might be trickier though.
If it doesn't apply cleanly, there's a risk that the issue you experience is caused by something else (but not necessarily)

an0n666:
Sorry I think it's a different issue.

PR #3947, opened 2019-08-20
remove empty statements and fix integral promotion warning

arvidn:


src/bdecode.cpp
```diff
@@ -942,7 +942,8 @@ namespace detail {
 				&& ret.m_tokens[stack[current_frame - 1].token].type == bdecode_token::dict)
 			{
 				// the next item we parse is the opposite
-				stack[current_frame - 1].state = ~stack[current_frame - 1].state;
```
arvidn: the problem here was that `state` being an unsigned single bit is promoted to an `int` via normal integer promotion rules. when the bitwise-not operator is applied, the value will change from 0 to -1. This -1 is then *implicitly* converted back to an `unsigned int`, which changes its interpretation of the current value (-1 -> 1). This is the problem.

With this change, the promoted `int` is always 0 or 1, the result of the addition is always 1 or 2. Both of those values are representable by an `unsigned int`, so the implicit conversion back is not a problem. When assigning it to a 1-bit bitfield, the value is truncated back to 0 or 1.

PR #3956, opened 2019-08-25
unify the DHT settings to use the main settings_pack

arvidn:
The main benefit of moving the dht settings into the main `settings_pack` is that new configuration options can be added without affecting the ABI. It also simplifies the overall API a bit.

The reason `settings_interface` was introduced was to avoid exposing the internal `aux::session_settings` to the `dht_storage_constructor` interface. Now `settings_pack` and `aux::session_interface` share a polymorphic interface. In practice, I would expect that dynamic dispatch for these calls to *only* be used inside the dht storage itself. both `settings_pack` and `aux::session_settings` are final, so de-virtualizing those calls should be an easy job for the compiler.

removing `dht::dht_settings` also impacts the `session_params` object (and simplifies it). But since this causes an ABI difference between a build with deprecated functions and one without, it's been put in an inline namespace to encode this.

@aldenml since this affects the dht storage API, would you mind taking a look?


aldenml:
@arvidn this looks like a really good improvement 👍

src/session_impl.cpp
```diff
@@ -5189,10 +5243,37 @@ namespace aux {
...
+		SET_INT(block_timeout) * 60;
+		SET_INT(block_ratelimit);
+		SET_BOOL(read_only);
+		SET_INT(item_lifetime);
+		SET_INT(upload_rate_limit);
+		SET_INT(sample_infohashes_interval);
+		SET_INT(max_infohashes_sample_count);
+		SET_BOOL(prefer_verified_node_ids);
+#undef SET_BOOL
+#undef SET_INT
```
arvidn: I wonder if it might make sense to scrap `dht::settings` altogether, and just pass in `aux::session_settings` or `settngs_interface` into the dht nodes. It would save this awkward conversion of all the DHT settings. The abstraction/separation of the DHT from the main library is already compromised, and even more so with this patch.

PR #3962, opened 2019-08-28
add back close_file_interval feature

arvidn:


arvidn:
I'm under the impression that this is still necessary on windows. @ssiloti is that your impression too?

ssiloti:
No, I don't think this is necessary.

Looking at https://github.com/qbittorrent/qBittorrent/issues/8328 and https://github.com/qbittorrent/qBittorrent/issues/5804 it appears these are cases of users freaking out unnecessarily over the disk cache using memory which would otherwise be unused. This would only be a problem if the cached data was not promptly evicted when the memory is needed for something else, and I don't see any evidence that is happening.  I am not aware of any legitimate issues with Windows 7 or newer which would justify this "feature".

arvidn:
that ticket I reference is just the last in a long string of reports like this, dating back many years. At least some of them have seen legitimate (as in the symptom was clearly described as system slowing down, not just "using lots of RAM"). I actually seem to recall we struggled with similar issues with uTorrent, using `FILE_FLAG_NO_BUFFERING` for a while at least.

One would hope that new versions of windows wouldn't suffer as much from this anymore though.

So, Maybe I'll leave this patch here in case someone files a similar report against the new libtorrent version.

@sledgehammer999 @Chocobo1 do you have recent experience with this from qbt?

Chocobo1:
>@sledgehammer999 @Chocobo1 do you have recent experience with this from qbt?

I don't recall users complaining about it recently.

arvidn:
I'm thinking that maybe I should land this but disable it by default on windows too. That at least leaves the possibility to enable it if someone runs into similar problems, to collect more samples of whether it ever makes a difference.

arvidn:
I don't quite feel confident it's safe to remove yet, but as long as more people can test it, and show it's unnecessary, it can be removed entirely in a later version.

PR #3965, opened 2019-08-29
make aux::session_settings thread-safe

arvidn:
this allows all threads to share the same settings object, which has low contention and is rarely updated

### DIFF
--- a/include/libtorrent/disk_io_thread.hpp
+++ b/include/libtorrent/disk_io_thread.hpp
@@ -279,7 +279,7 @@ namespace aux { struct block_cache_reference; }
 		, disk_interface
 		, buffer_allocator_interface
 	{
-		disk_io_thread(io_service& ios, counters& cnt);
+		disk_io_thread(io_service& ios, aux::session_settings const&, counters&);
 #if TORRENT_USE_ASSERTS
 		~disk_io_thread();
 #endif
@@ -290,7 +290,7 @@ namespace aux { struct block_cache_reference; }
 			hasher_thread_divisor = 4
 		};
 
-		void set_settings(settings_pack const* sett);
+		void settings_updated();
 
 		void abort(bool wait);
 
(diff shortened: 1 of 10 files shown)


arvidn:
Perhaps this is a bit overkill, but I think I want to go in this direction in master anyway, I can’t think of a much simpler solution.

arvidn:
@aldenml is that a sign of disapproval?

include/libtorrent/disk_io_thread.hpp
```diff
@@ -279,7 +279,7 @@ namespace aux { struct block_cache_reference; }
 		, disk_interface
 		, buffer_allocator_interface
 	{
-		disk_io_thread(io_service& ios, counters& cnt);
+		disk_io_thread(io_service& ios, aux::session_settings const&, counters&);
```
arvidn: I think everything in this file is a great win. No duplication of the settings object

aldenml:
a neutral comment :D, I would love to have more time to provide a serious opinion

src/create_torrent.cpp
```diff
@@ -294,7 +294,14 @@ namespace {
...
 		counters cnt;
-		disk_io_thread disk_thread(ios, cnt);
+		aux::session_settings sett;
+
+		sett.set_int(settings_pack::cache_size, 0);
+		int const num_threads = disk_io_thread::hasher_thread_divisor - 1;
+		int const jobs_per_thread = 4;
+		sett.set_int(settings_pack::aio_threads, num_threads);
+
+		disk_io_thread disk_thread(ios, sett, cnt);
```
arvidn: it's nicer (I think) to pass in the settings in the constructor, rather than setting it later

include/libtorrent/storage.hpp
```diff
@@ -277,7 +277,7 @@ namespace libtorrent {
 		virtual ~storage_interface() {}
 
 		// initialized in disk_io_thread::perform_async_job
-		aux::session_settings* m_settings = nullptr;
+		aux::session_settings const* m_settings = nullptr;
```
arvidn: technically, this could be cleaned up and simplified now too, where this member could be initialized in the constructor (but that would be an API break). But it could at least be initialized by the main thread immediately after construction

PR #4006, opened 2019-09-22
provide the previous filename in file_renamed_alert

arvidn:


src/alert.cpp
```diff
@@ -247,14 +248,22 @@ namespace libtorrent {
...
+		return m_alloc.get().ptr(m_old_name_idx);
+	}
+
 	std::string file_renamed_alert::message() const
 	{
 		std::string ret { torrent_alert::message() };
 		char msg[200];
-		std::snprintf(msg, sizeof(msg), ": file %d renamed to "
-			, static_cast<int>(index));
+		std::snprintf(msg, sizeof(msg), ": file %d renamed from \"",
```
aldenml: not an issue by any mean, but are you moving to trailing `,`?

arvidn: unintentional :)

aldenml: 😀

an0n666:
arvidn can you please backport this to RC1_2?

PR #4014, opened 2019-09-29
Cleanup example

pixelherodev:


arvidn:
this will cause a delay when exiting, right? not only will it iterate over the remaining alerts, but it will also put the thread to sleep, when it's already decided to exit.

arvidn:
I agree that this is better! However, I don't think the `done` variable is necessary anymore.
Also, there's some whitespace at the end of lines 57 and 59. https://travis-ci.org/arvidn/libtorrent/jobs/591245691#L1285

arvidn:
oh, right. you still need the `done` in order to exit two loops. I'm not sure additional mutable state is easier to reason about than a forward-goto out of nested loops.

arvidn:
I think exiting nested loops is a perfect use of `goto`. Any alternative I can think of is worse. Certainly additional mutable program state and additional tests is worse. I'm more than happy to discuss merits of different solutions or maybe different ways to lay out the code to avoid the goto, but I don't think "avoiding goto at all costs" is a reasonable goal. The goal should be to make code easy to follow and reason about, and having intent clearly expressed.

PR #4020, opened 2019-10-08
Windows Cryptography Next Generation

ajax16384:
This pull request add support for Windows Cryptography API: Next Generation (https://docs.microsoft.com/windows/desktop/SecCNG/cng-portal).
According to my simple test (https://gist.github.com/ajax16384/3adbb2be258366065753fddaa6de8b8f) this support may give some performance boost at multiple hash object recreation area like piece-block checking
(it seems old CryptoApi has some heavy lift code at hash creation/destroy routines).

My Windows 10 shows following numbers:
````
Start
CNG Random small buf:       2.51 sec
CNG Random large buf:       0.44 sec
CNG SHA1 small buf:       4.76 sec
CNG SHA1 large buf:       1.84 sec
CNG SHA512 small buf:       5.78 sec
CNG SHA512 large buf:       2.68 sec
CryptoAPI Random small buf:       2.76 sec
CryptoAPI Random large buf:       0.43 sec
CryptoAPI SHA1 small buf:       8.76 sec
CryptoAPI SHA1 large buf:       1.84 sec
CryptoAPI SHA512 small buf:       9.73 sec
CryptoAPI SHA512 large buf:       2.67 sec
````

Upon completion of reviewing and merging PR to master - it would be nice to back-port it to 1.2.

### CODE REVIEW

File: include/libtorrent/aux_/win_cng.hpp [hunk 1]
```diff
@@ -0,0 +1,206 @@
...
+		{
+			static std::size_t object_size =
+				static_cast<std::size_t>(
+					cng_get_algorithm_object_size(get_algorithm_handle()));
+			return object_size;
+		}
+
+		BCRYPT_HASH_HANDLE m_hash;
+		std::vector<UCHAR> m_hash_object = std::vector<UCHAR>(get_algorithm_object_size());
+	};
```
arvidn: it seems like the code here should be put in a .cpp file, instead of being inlined. would that cause problems?

include/libtorrent/hasher.hpp
```diff
@@ -119,6 +122,8 @@ TORRENT_CRYPTO_NAMESPACE
 		gcry_md_hd_t m_context;
 #elif TORRENT_USE_COMMONCRYPTO
 		CC_SHA1_CTX m_context;
+#elif TORRENT_USE_CNG
```
arvidn: the the `TORRENT_CRYPTO_NAMESPACE` above (line 72). The definition of that macro should also depend on whether `TORRENT_USE_CNG` is defined or not. This protects against the ABI effect of using different crypto libraries.

include/libtorrent/aux_/win_cng.hpp
```diff
@@ -0,0 +1,206 @@
...
+
+#ifndef TORRENT_WIN_CNG_HPP
+#define TORRENT_WIN_CNG_HPP
+
+#include <vector>
+
+#include "libtorrent/config.hpp"
+#include "libtorrent/error_code.hpp"
+#include "libtorrent/aux_/throw.hpp"
+#include "libtorrent/aux_/windows.hpp"
```
arvidn: is this file not supposed to be included in case we're not on windows? it seems like it would be safer to add those checks here, in this file

include/libtorrent/aux_/win_cng.hpp
```diff
@@ -0,0 +1,206 @@
...
+
+#include "libtorrent/aux_/disable_warnings_push.hpp"
+#include <bcrypt.h>
+#include "libtorrent/aux_/disable_warnings_pop.hpp"
+
+namespace libtorrent { namespace aux {
+
+	static void throw_ntstatus_error(char const* name, NTSTATUS status) {
```
arvidn: if this function is going to be defined in a header, it should be `inline`. and I don't really think you want to give it internal linkage (with `static`)

include/libtorrent/aux_/win_cng.hpp
```diff
@@ -0,0 +1,206 @@
...
+		char message[1024];
+		std::snprintf(message, sizeof(message),
+			"Error 0x%x returned by %s", status, name);
+		throw_ex<std::runtime_error>(message);
```
arvidn: instead of converting the `NTSTATUS` error message into a string, isn't there an `error_category` for NT error codes? Ideally it would be preserved in the error message (and ideally the string would only be generated on demand, when `what()` is called for instance)

include/libtorrent/aux_/win_cng.hpp
```diff
@@ -0,0 +1,206 @@
...
+	inline BCRYPT_ALG_HANDLE cng_open_algorithm_handle(cng_algorithm alg_id)
+	{
+		static LPCWSTR algorithm_str[] =
+		{
+			BCRYPT_RNG_ALGORITHM,
+			BCRYPT_SHA1_ALGORITHM,
+			BCRYPT_SHA256_ALGORITHM,
+			BCRYPT_SHA512_ALGORITHM
+		};
+		LPCWSTR str = algorithm_str[static_cast<uint8_t>(alg_id)];
```
arvidn: the portable spelling is `std::uint8_t`

include/libtorrent/aux_/win_cng.hpp
```diff
@@ -0,0 +1,206 @@
...
+			&data_size, 0);
+		if (status < 0) {
+			throw_ntstatus_error("BCryptGetProperty BCRYPT_OBJECT_LENGTH",
+				status);
+		}
+
+		return object_size;
+	}
+
+	inline void cng_gen_random(span<char> buffer)
```
arvidn: this function isn't used anywhere, as far as I can tell. what's its purpose?

include/libtorrent/hasher.hpp
```diff
@@ -119,6 +122,8 @@ TORRENT_CRYPTO_NAMESPACE
...
+		aux::cng_hash<aux::cng_algorithm::sha1> m_context;
```
arvidn: it seems like perhaps `aux::cng_hash` should be given an API compatible to `hasher`, and it could just be an alias here, instead of interleaving the cng code with the rest of `hasher`. In fact, given how many `hasher` implementations there are, that would probably be better for all of them

include/libtorrent/hasher.hpp
ajax16384: thanks, forget to define macro as 0 in case of undefined

include/libtorrent/aux_/win_cng.hpp
ajax16384: win_cng.hpp (as an old win_crypto_provider.hpp ) must only be included in windows builds. aux_\win* have no such checking.

include/libtorrent/aux_/win_cng.hpp
ajax16384: ok

include/libtorrent/aux_/win_cng.hpp
ajax16384: I thought NTSTATUS is not compatible with generic_category or system_category(windows getlasterror results), but have found MS related source that uses ntstatus as system_error code https://github.com/Azure/azure-storage-fuse/blob/43e82df5d85a4c082dc67af8131bcf05f4d9270a/azure-storage-cpp-lite/src/hash.cpp#L16

Will fix this

include/libtorrent/aux_/win_cng.hpp
ajax16384: ok, will try to move to LPCWSTR typename template approach.

include/libtorrent/aux_/win_cng.hpp
ajax16384: ok

include/libtorrent/hasher.hpp
ajax16384: just used same approach as TORRENT_USE_CRYPTOAPI.

include/libtorrent/aux_/win_cng.hpp
arvidn: great! then you could use the `system_error` exception too. unfortunately I don't think you can use `std::error_code` as `boost::system::error_code` is used pervasively throughout boost.asio

include/libtorrent/aux_/win_cng.hpp
arvidn: I can see it being useful in `random.cpp` perhaps. but you'd think `std::random_device` already uses this under the hood on windows

include/libtorrent/hasher.hpp
arvidn: yeah, it's fine. I think as this grows in number of alternatives, it would make sense to split them out into individual classes (with compatible interfaces). but that's separate from this patch

arvidn:
you got this one already, to define the macro to 0 when it's not active: https://travis-ci.org/arvidn/libtorrent/jobs/595231077#L5732

This one is a bit mysterious. it looks like perhaps the header you include defines some macro screwing with the code: https://ci.appveyor.com/project/arvidn/libtorrent/builds/27967183/job/6i56s1dipmkcapx2#L347
https://ci.appveyor.com/project/arvidn/libtorrent/builds/27967183/job/6i56s1dipmkcapx2#L397

actually, maybe it's some openssl header that fails to be included: https://ci.appveyor.com/project/arvidn/libtorrent/builds/27967183/job/w8m9jbd2qc3v70yy#L159

ajax16384:
only  strange appveyor errors left ...

arvidn:
the patch looks good, but it can't land breaking mingw and windows XP builds. My suspicion is that there's a case where CNG is not supported, and some header that used to be included in that case is not anymore

ajax16384:
pre-Vista "windows-version=" mode will try to use good old CRYPTOAPI (checked by myself with windows-version=xp succefull build). As of mingw - it should build without a problem cause mingw contains bcrypt.h  (https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-headers/include/bcrypt.h )

arvidn:
the mingw `bcrypt.h` is probably different from the one that comes with windows. so perhaps it doesn't include some other header that the code currently relies one. that's included if `TORRENT_USE_CNG` is defined. my suspicion is that there's a missing include.

arvidn:
for instance, does `bcrypt.h` define `X509_get_subject_name()` and `X509_NAME*`? (those are the names OpenSSL uses). So, perhaps by excluding some preprocessor path if an `#if TORRENT_USE_CNG` you fail to include the openssl header, which defines those.

ajax16384:
- if you insist then TORRENT_USE_CNG detection might be disabled for MINGW path
- CGN will be detected/used only if any other (TORRENT_USE_LIBCRYPTO or TORRENT_USE_LIBGCRYPT) crypto engines not defined, so there is no name definition/declaration ambiguity

arvidn:
my guess is that in your patch `#if TORRENT_USE_OPENSSL` is true, you fail to include the openssl header by intercepting that preprocessor check. just look at the error in torrent.cpp: https://github.com/arvidn/libtorrent/blob/master/src/torrent.cpp#L1407 it's clear the X509 macros/types are defined there.

as far as I can tell, it's included from https://github.com/arvidn/libtorrent/blob/master/include/libtorrent/aux_/openssl.hpp#L64

you could add an `#error` there or around it to understand if it's getting there, and why if not

ajax16384:
oh, i see

ajax16384:
done

PR #4037, opened 2019-10-21
add Windows "Cryptography API: Next Generation" support

ajax16384:
backport CNG support


arvidn:
this is problematic since `hasher` is part of the public API and this alters its ABI

ajax16384:
https://travis-ci.org/arvidn/libtorrent/jobs/601500649#L858 crashed for unknown reason. There is any way to bump recheck?

PR #4042, opened 2019-10-23
fix storing udp port mappings in listen_socket_t

### LINKED ISSUE #4039: tcp_port_mapping - then statement is equivalent to the else
libtorrent version: master

https://github.com/arvidn/libtorrent/blob/master/src/session_impl.cpp#L5487-L5488

ssiloti:
Fixes #4039


arvidn:
this looks like a correct fix. did you close it because you intend to resubmit it against `RC_1_2`?

PR #4046, opened 2019-10-27
use less space for timestamps in peer_connection

arvidn:
there are 13 timestamp fields in peer_connection. Make 12 of them relative the connect timestamp, and use narrower types


aldenml:
interesting, I like it :)

PR #4050, opened 2019-10-29
include: GCC 10 support

mooninite:
Red Hat is testing GCC 10 and found libtorrent needed to add a direct include on `<string>`.

This is a relatively common failure we're seeing with gcc-10 and occurs
as a result of header file cleanups within libstdc++.  For example
various C++ runtime headers indirectly included <string> which in turn
included <local> and <cerrno>.  Those indirect inclusions have been
eliminated which in turn forces packages to include the C++ headers they
actually need.


arvidn:
thanks! I will back-port this to `RC_1_2` as well

PR #4051, opened 2019-10-30
client_test: add set_torrent_params

pavel-pimenov:
add using std::stoi;
remove p.userdata = static_cast<void*>(new std::string(torrent)) (memory leak?)


arvidn:
yes, I think that `userdata` field is some left-over from a long time ago.
This looks good, would you mind making the PR against `RC_1_2` though?
I merge `RC_1_2` into master regularly, so it will make it in here too

pavel-pimenov:
https://github.com/arvidn/libtorrent/pull/4055

PR #4067, opened 2019-11-03
clean up type for socket type index

arvidn:


src/peer_connection.cpp
```diff
@@ -168,7 +168,8 @@ namespace libtorrent {
 		, m_exceeded_limit(false)
 		, m_slow_start(true)
 	{
-		m_counters.inc_stats_counter(counters::num_tcp_peers + socket_type_idx(*m_socket) - 1);
+		m_counters.inc_stats_counter(counters::num_tcp_peers
+			+ static_cast<std::uint8_t>(socket_type_idx(*m_socket)));
```
arvidn: removing the `-1` here is subtle. In the new enum, there is no value representing an "empty" socket, so the numbers shifted down one step

include/libtorrent/socket_type.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+#ifndef TORRENT_SOCKET_TYPE_HPP
+#define TORRENT_SOCKET_TYPE_HPP
+
+#include "libtorrent/config.hpp"
+#include <cstdint>
+
+namespace libtorrent {
+
+enum class socket_type_t : std::uint8_t {
```
arvidn: it's important that this new enum matches the order of the previous socket index definitions: https://github.com/arvidn/libtorrent/pull/4067/files#diff-845c1357857c1caeb44d4beec382177fL53

Issue #4058, opened 2019-10-31
Handler buffer not large enough, please increase it


MassaRoddel:
compile error in peer_connection.cpp

libtorrent version (or branch): master
compiler and compiler version: VC2015 64Bit (v140)
Where/how do I increase the buffer?

ssiloti:
Handler sizes are set in `include/libtorrent/aux_/allocating_handler.hpp`. Increase the size of the offending handler to the actual size you observe. Be sure to mind the `#ifdefs`, handler sizes vary per platform and build config.

MassaRoddel:
Changing the debug_read_iter from 16*.. to 22*... fixed the compile problem.
Thanks.

arvidn:
Is it the debug iterators that use up the space? After your change, do release builds still work?

arvidn:
I imagine this change should go into master, right?

MassaRoddel:
The change affects debug builds only (per #ifndef NDEBUG). The release version compiled without problems and seems to work.
I just started from the top to change sizes and increasing debug_read_iter fixed it. Could be that the actual problem is a different handler size (i.e. openssl_read_cost).
The change should go into master.

PR #4070, opened 2019-11-04
Increase debug_read_iter/debug_write_iter

vktr:
Here's what I've been using on Windows to fix the error *Handler buffer not large enough* for debug builds. Release builds have been fine.

This builds in both Debug and Release, however I am not knowledgeable enough to make out if we should use something less than 24. An increase by 8 felt good.

Relates to #4058


arvidn:
thanks!

PR #4072, opened 2019-11-05
Increase read/write handler max size for non-Windows

vktr:
This will increase the read/write handler max size for non-Windows platforms to resolve the error *Handler buffer not large enough* which I got on a clean Ubuntu 19.10 x64 system.

### DIFF
--- a/include/libtorrent/aux_/allocating_handler.hpp
+++ b/include/libtorrent/aux_/allocating_handler.hpp
@@ -102,8 +102,8 @@ namespace libtorrent { namespace aux {
 	constexpr std::size_t openssl_read_cost = 0;
 	constexpr std::size_t openssl_write_cost = 0;
 #endif
-	constexpr std::size_t write_handler_max_size = tracking + debug_write_iter + openssl_write_cost + 136;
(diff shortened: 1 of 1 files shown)


arvidn:
was this for a release build with openssl disabled?

vktr:
No, this was for a regular build, generated with

`cmake -Ddeprecated-functions=off`

arvidn:
It seems to me that GCC 9 is involved, but I can’t reproduce it on my Ubuntu system with GCC 9.

ssiloti:
I am able to reproduce it using GCC 9.2.0 on Linux. Make sure you're using openssl for crypto. I can also confirm that this PR fixes the build.

arvidn:
if openssl is enabled, it sounds like this additional size should be added to the `openssl_read_cost` and `openssl_write_cost`. does that sound right?

vktr:
I’ll double check when I’m at my computer again

vktr:
Tested, and I get the error regardless of whether encryption is enabled or disabled.

arvidn:
thanks! @vktr if you're in Stockholm, you should come to the C++ meetups! https://www.meetup.com/StockholmCpp/

PR #4076, opened 2019-11-08
Increase UPNP sendbuffer length

vktr:
This will increase the UPNP sendbuffer with 512 bytes to a total of 2560. Building with `-Werror` on GCC9 gave the following warning,

```
../src/upnp.cpp: In member function ‘void libtorrent::upnp::get_ip_address(libtorrent::upnp::rootdevice&)’:
../src/upnp.cpp:719:4: warning: ‘%s’ directive output may be truncated writing up to 2047 bytes into a region of size between 1920 and 1923 [-Wformat-truncation=]
  719 |   "%s"
      |    ^~
......
 1096 |  post(d, soap, soap_action);
```

I don't know if we should keep the buffer 2k aligned and in that case perhaps increase it to 4096? Having it at 2560 fixes the warning at least.

### DIFF
--- a/src/upnp.cpp
+++ b/src/upnp.cpp
@@ -710,7 +710,7 @@ void upnp::post(upnp::rootdevice const& d, char const* soap
 	TORRENT_ASSERT(d.magic == 1337);
 	TORRENT_ASSERT(d.upnp_connection);
 
-	char header[2048];
+	char header[2560];
 	std::snprintf(header, sizeof(header), "POST %s HTTP/1.1\r\n"
 		"Host: %s:%d\r\n"
 		"Content-Type: text/xml; charset=\"utf-8\"\r\n"


arvidn:
I don't know where GCC gets the 2047 number from. one of the strings is a `std::string`, so surely there is no bound on how large the output can be. I don't think the case where the UPNP header would end up being greater than 2kiB matters. I don't think it would legitimately be that large.

vktr:
As I understand it, we allocate a buffer for the SOAP body in (for example) `void upnp::get_ip_address(rootdevice& d)` which is 2048 bytes. This buffer is then passed to `void upnp::post(upnp::rootdevice const& d, char const* soap, char const* soap_action)` as the  `soap` parameter.

Inside the `post` function is where this PR increases the `header` buffer to which the SOAP body buffer copied, since both were 2048 bytes before and if the SOAP body is filled, it would be truncated.

So the `header` variable actually ends up with both the HTTP headers as well as the SOAP body.

The 2047 number could be from GCC trying to not count a null-terminating character?

If we shouldn't increase the buffer length, then maybe we could disable the warning with a `#pragma` directive.

arvidn:
My understanding is that GCC assumes that the first `snprintf()` can produce a full 2048 byte output, since that's the buffer size, and if it would do that, the second `snprintf()` may truncate.

This buffer is probably way too large for what's necessary. I wonder if it would make sense to make the soap buffer smaller instead, to silence this warning.

arvidn:
how does this look? https://github.com/arvidn/libtorrent/pull/4131

vktr:
Looks good!

PR #4081, opened 2019-11-12
python-binding: attempt to use numpy's multi-threaded compiler if available

damolp:
Gives similar behaviour to building the python bindings with `boost-build` if `numpy` is present on the system.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -9,6 +9,15 @@ import sys
 import shutil
 import multiprocessing
 
+# try to replace the default distutils compiler with the numpy one (which is multi-threaded)
+try:
+    import numpy.distutils.ccompiler
+    import distutils.ccompiler
+    distutils.ccompiler.CCompiler.compile = numpy.distutils.ccompiler.CCompiler_compile
+    os.environ['NPY_NUM_BUILD_JOBS'] = str(multiprocessing.cpu_count())
+except ImportError:
+    pass
+
 
 class flags_parser:
     def __init__(self):


arvidn:
this is not replacing the C++ compiler, right? (because I don't think this will, libtorrent is built with `bjam` further down). Is it replacing the python source -> bytecode compiler?
There is virtually no (possibly none at all) python code that's part of the binding. Did you measure a speed-up with this change?

I assume it doesn't affect the resulting binary in anyway too.

You say "similar", but if this affects ABI, it cannot be enabled implicitly. There would need to be an explicit switch for it.

damolp:
> this is not replacing the C++ compiler, right?

No, it calls the C++ compiler in parallel. The compilation of the C++ is done *from* python which is why the `-j` flag on `make` is ignored.

> Did you measure a speed-up with this change?

Yep:
With:
```
ubuntu@ip-172-31-16-160:~/libtorrent-s3/bindings/python$ time make > /dev/null
real	0m19.989s
user	2m3.829s
sys	0m12.034s
```

Without:
```
ubuntu@ip-172-31-16-160:~/libtorrent-s3/bindings/python$ time make > /dev/null
real	2m4.076s
user	1m52.451s
sys	0m11.576s
```



> I assume it doesn't affect the resulting binary in anyway too.

Nope, still the same binary.


> You say "similar", but if this affects ABI, it cannot be enabled implicitly. There would need to be an explicit switch for it.

No it doesn't change the ABI. I said similar because I saw [this branch](https://github.com/arvidn/libtorrent/blob/RC_1_2/bindings/python/setup.py#L84) in `setup.py` which compiles the python bindings in parallel

arvidn:
I had some time to look at this again. I think I see what this is doing. It's not affecting the `--bjam` path, where the python module is built by `b2`, but it's affecting the path where distutils is actually compiling the C++ code. (which, frankly, I'm surprised works at all).

I'm not confident that distutils will be building the module correctly. Specifically, ABI compatible with the main library. This is the main reason I build it with `bjam`, especially on windows.

I'm not opposed to this patch, I worry a bit about it encouraging more use of this build path, which I'm not confident in is correct.

PR #4082, opened 2019-11-12
remove include

pavel-pimenov:


examples/client_test.cpp
```diff
@@ -30,7 +30,6 @@ POSSIBILITY OF SUCH DAMAGE.
 
 */
 
-#include <cstdio> // for snprintf
```
arvidn: `std::snprintf()` is used in this file. It seems inappropriate to remove the include and instead rely on some other header including this "by accident".

arvidn:
how did you determine which includes should be removed? It doesn't look like the algorithm is all the way right.

pavel-pimenov:
Doxygen include graph for client_test.cpp
...
#include "print.hpp" 
...

изображение

arvidn:
I don't think it's good practice to rely on other headers to pull in dependencies the file itself has. it makes the headers and cpp files less independent. includes that don't have any direct dependecies in the file though, those are good candidates for removal.

test/web_seed_suite.cpp
```diff
@@ -30,14 +30,12 @@ POSSIBILITY OF SUCH DAMAGE.
 
 */
 
-#include "libtorrent/session.hpp"
```
arvidn: this file uses `session`. it constructs one. I don't see why you wouldn't want to include the header

pavel-pimenov: изображение

arvidn: I don't understand your point. `setup_transfer.hpp` *also* uses `session`, right? so it makes sense for it to also include it.
To put it another way, if `setup_transfer.hpp` for some reason, in the future, stops using `session` and removes its include; it doesn't make any sense for `web_seed_suite.cpp` to fail.

arvidn:
@pavel-pimenov I'm not really interested in manually reviewing all includes you've removed, to make sure they are sensible. Instead, can you tell me what algorithm you've used to determine what to remove?

PR #4090, opened 2019-11-13
test new appveyor image

arvidn:


arvidn:
@zeule or @ssiloti could one of you could help me with a cmake issue?

I think this is related to updating to the latest boost version. cmake cannot find boost.system (but it appears to be finding other boost libraries). My suspicion is that perhaps boost.system is header only now, and that find_package perhaps can only find compiled libraries.

here's the error from appveyor: https://ci.appveyor.com/project/arvidn/libtorrent/builds/28839301/job/27v0qr1xmrm15927#L81

arvidn:
it happens with boost-1.69 too
https://ci.appveyor.com/project/arvidn/libtorrent/builds/28841036/job/ucsu87qs4avkx0cv#L250

ssiloti:
The install script is calling `bootstrap.bat` to build bjam, but not building any libraries. Are you sure the library files exist? They would probably be in a `lib` or `stage` directory under `%BOOST_ROOT%`.

arvidn:
it's possible appveyor changed to not have binaries pre-installed anymore, but in previous images they did. I would be surprised if the libraries aren't there anymore.

ssiloti:
It could be because you're building with Visual Studio 2015 on a 2017 image. There may only be boost libraries built with 2017. Such libraries are usable by 2015, but the cmake boost script may not be smart enough to use the 2017 libraries when building with 2015.

ssiloti:
Ok found the relevant bit in FindBoost.cmake:

```
    elseif(MSVC_TOOLSET_VERSION GREATER_EQUAL 140)
      # MSVC toolset 14.x versions are forward compatible.
      set(_boost_COMPILER "")
      foreach(v 9 8 7 6 5 4 3 2 1 0)
        if(MSVC_TOOLSET_VERSION GREATER_EQUAL 14${v})
          list(APPEND _boost_COMPILER "-vc14${v}")
        endif()
      endforeach()
```

So msvc 14.x will link to libraries built with 14.y where x >= y, but not the other way around.

arvidn:
that makes sense. is it the cmake generator that selects the msvc version?

PR #4112, opened 2019-11-23
fixed restoring of trackers, comment, creation date and created-by in resume data

arvidn:


src/write_resume_data.cpp
```diff
@@ -92,6 +92,12 @@ namespace libtorrent {
 			auto const info = atp.ti->metadata();
 			int const size = atp.ti->metadata_size();
 			ret["info"].preformatted().assign(&info[0], &info[0] + size);
+			if (!atp.ti->comment().empty())
+				ret["comment"] = atp.ti->comment().empty();
```
j1warren: .empty() seems like a typo

sledgehammer999:
The `comment` isn't preserved for some reason. `created_by` and `created date` are preserved.

I am going to test the fix for the trackers too. But I imagine you'll have to do the same fix for `http_seeds` and `url_seeds`.
Also can the values of `peers` and `banned_peers` come the .torrent file too? If yes, can those be manipulated by the API? If yes, then the same fix is in order for them too.

sledgehammer999:
The tracker problem seems fixed too: Aka not passing `torrent_handle::save_info_dict` to `save_resume_data()` while simultaneously having deleted all the trackers. Then loading the torrent using the previously generated fastresume AND the .torrent file separately (.torrent file containing values for trackers). Now, the torrent loads correctly without having any trackers in the list.

arvidn:
there's a bug regarding the "comment". I'm fixing that along with adding a few tests.

arvidn:
as far as I can tell (and also confirmed by my tests) the web seeds are already covered by the `override_web_seeds` flag, and it already works as expected.

arvidn:
I think this is ready to land. Anyone else want to try it out? maybe in qbt?

test/test_resume.cpp
```diff
@@ -276,6 +276,66 @@ TORRENT_TEST(piece_priorities)
...
+	save_resume_data_alert const* ra = alert_cast<save_resume_data_alert>(a);
+	TEST_CHECK(ra);
+	if (ra)
+	{
+		auto const& atp = ra->params;
+		TEST_CHECK(atp.trackers == std::vector<std::string>{"http://torrent_file_tracker2.com/announce"});
+		TEST_CHECK(atp.url_seeds == std::vector<std::string>{"http://torrent.com/"});
+		TEST_CHECK(atp.ti);
+		TEST_EQUAL(atp.ti->comment(), "test comment");
+		TEST_EQUAL(atp.ti->comment(), "test comment");
```
sledgehammer999: duplicate line

sledgehammer999:
I am going to test now.

>he web seeds are already covered by the override_web_seeds flag

Are you sure that's the correct way? Consider this use case: We add .torrent file for the first time which contains both trackers and web_seeds. If we use the override flags won't libtorrent use the empty trackers/seeds list from the default constructed add_torrent_params object? Thus not loading any trackers/seeds?
In this case, from our POV should we always use the override flags when restoring from fastresume and not set them when adding it for the 1st time? Is there any case, that we should use override flags when restoring from fastresume? Does fastresume store the full list of trackers/seeds when nothing is removed from it from the original list in the .torrent file?

arvidn:
> Consider this use case: We add .torrent file for the first time which contains both trackers and web_seeds. If we use the override flags won't libtorrent use the empty trackers/seeds list from the default constructed add_torrent_params object? Thus not loading any trackers/seeds?

That's correct. But I would expect you to *not* set the `override_web_seeds` flag. This flag is set by libtorrent in `read_resume_data()`, which means we *know* the torrent is being restored from resume data, and not added for the first time.

> In this case, from our POV should we always use the override flags when restoring from fastresume and not set them when adding it for the 1st time?

Yes, I think that's right. The `read_resume_data()` function essentially does that. It actually bases it on whether the fields exist at all in the resume data file. And empty list of trackers sets the override flag, whereas a missing list doesn't.

> Is there any case, that we should use override flags when restoring from fastresume? Does fastresume store the full list of trackers/seeds when nothing is removed from it from the original list in the .torrent file?

Yes, that is the intention. And again, you (the client) isn't supposed to set the override flag.

sledgehammer999:
>And again, you (the client) isn't supposed to set the override flag.
>This flag is set by libtorrent in read_resume_data()

Oh sorry. I did not know that. This doesn't seem to be mentioned in the docs.
Anyway, this commit seems solid now. I am currently building and testing it.

sledgehammer999:
url/http seeds are broken. The bug seems to be that there isn't an empty list written out in the fastresume when all the url/http seeds are deleted.

The `comment` field is now correctly restored.

arvidn:
thanks for testing. I will add this case to the unit test.

arvidn:
@sledgehammer999 fixed now

sledgehammer999:
FYI, I'll be able to test again in ~12 hours.

sledgehammer999:
I confirm that all issues seem fixed now.

PR #4113, opened 2019-11-23
fixed warning about explicitly calling std::move to avoid copying on older compilers

aldenml:
With the latest version of Apple Clang, I'm getting this warning:
```
warning: prior to the resolution of a defect report against ISO C++11, local variable '<name>' would have been copied despite being returned by name, due to its not matching the function return type ('shared_ptr<base>' vs 'shared_ptr<child>') [-Wreturn-std-move-in-c++11]

note: call 'std::move' explicitly to avoid copying on older compilers
```
I know it's not relevant, but it's just one thing less to ignore.

### DIFF
--- a/src/kademlia/find_data.cpp
+++ b/src/kademlia/find_data.cpp
@@ -131,7 +131,7 @@ observer_ptr find_data::new_observer(udp::endpoint const& ep
 #if TORRENT_USE_ASSERTS
 	if (o) o->m_in_constructor = false;
 #endif
-	return o;
+	return std::move(o);
 }
 
 char const* find_data::name() const { return "find_data"; }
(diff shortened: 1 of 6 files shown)


arvidn:
the problem with this patch is that it triggers another warning, on compilers that have implemented the resolution of the defect report, that the `std::move()` is redundant and may inhibit RVO.

aldenml:
Do you prefer add `-Wreturn-std-move-in-c++11` to the ignored ones instead?

aldenml:
with your comment and more thinking about this, I think ignoring it is the best approach

aldenml:
and I can see it's already ignored in `Jamfile` :)
```
result += <cflags>-Wno-return-std-move-in-c++11 ;
```

PR #4117, opened 2019-11-24
add support for wolfSSL

aldenml:
@arvidn I cherry picked from your other PR


.travis.yml
```diff
@@ -220,6 +221,14 @@ install:
       cmake --version;
     fi
 
+  - if [[ "$crypto" == "wolfssl" ]]; then
+      wget --quiet -O wolfssl-4.2.0-stable.zip https://github.com/wolfSSL/wolfssl/archive/v4.2.0-stable.zip;
+      unzip wolfssl-4.2.0-stable.zip;
+      cd wolfssl-4.2.0-stable;
+      ./autogen.sh && ./configure --enable-asio && make && sudo make install;
```
arvidn: what does `--enable-asio` do? If the installed version of wolfSSL needs to be built with a specific configuration, it's probably a good idea to mention this in the documentation somewhere

include/libtorrent/aux_/hasher512.hpp
```diff
@@ -58,6 +58,14 @@ POSSIBILITY OF SUCH DAMAGE.
 	#include <openssl/sha.h>
 	}
 
+#elif defined TORRENT_USE_WOLFCRYPT
+
+// when using wolfcrypt, we need to enable the SHA-512 hasher
+#define WOLFSSL_SHA512 1
+
+#define WC_NO_HARDEN
```
arvidn: could you add a comment explaining what this does?

arvidn:
I think the most important test is disabled in this new configuration. See `test/test_ssl.cpp`, it's wrapped in an `#if TORRENT_USE_OPENSSL` check. I think you should add `|| defined TORRENT_USE_WOLFSSL` to that condition.

Other tests that also test SSL functionality (and need to be updated to exercise it when using WolfSSL are):

* test_http_connection.cpp
* test_url_seed.cpp
* test_web_seed_ban.cpp
* test_web_seed_chunked.cpp
* test_web_seed_http.cpp
* test_web_seed_http_pw.cpp
* test_web_seed_socks4.cpp
* test_web_seed_socks5.cpp
* test_web_seed_socks5_no_peers.cpp
* test_web_seed_socks5_pw.cpp

I also notice that the `Jamfile` defines `TORRENT_USE_WOLFSSL` when building with `crypto=wolfssl`, but there's nothing in the code that tests for this defined, just `TORRENT_USE_WOLFCRYPT`.

.travis.yml
aldenml: I was surprised by the need of this option, per wolfSSL documentation, it's needed to put all the OpenSSL layer in place in order to ASIO be able to work with it, but in fact it just enable "opensslextra" and "opensslall", and that makes `libwolfssl` quite big, but still significantly smaller than openssl

include/libtorrent/aux_/hasher512.hpp
aldenml: will do

aldenml:
I will take care of your comments

aldenml:
I think this can be greatly simplified if we rely on the emulation layer alone

Jamfile
```diff
@@ -396,6 +400,52 @@ rule openssl-include-path ( properties * )
 	return $(result) ;
 }
 
+# the search path to pick up the wolfssl libraries from. This is the <search>
+# property of those libraries
+rule wolfssl-lib-path ( properties * )
```
aldenml: a few things are wrong in these rules

PR #4119, opened 2019-11-24
fix handling of torrents with too large pieces

arvidn:


src/torrent.cpp
```diff
@@ -1860,7 +1870,6 @@ bool is_downloading_state(int const st)
 				TORRENT_ASSERT((pr.start & (block_size() - 1)) == 0);
 
 				int block = block_size();
-				int blocks_per_piece = m_torrent_file->piece_length() / block;
```
arvidn: there should be a test covering this fix

src/torrent.cpp
```diff
@@ -1776,6 +1777,15 @@ bool is_downloading_state(int const st)
 			return;
 		}
 
+		int const blocks_per_piece
+			= (m_torrent_file->piece_length() + default_block_size - 1) / default_block_size;
+		if (blocks_per_piece > piece_picker::max_blocks_per_piece)
```
arvidn: there should be a test covering this

src/torrent.cpp
```diff
@@ -8183,7 +8194,7 @@ bool is_downloading_state(int const st)
 		}
 #endif
 		// if the error happened during initialization, try again now
-		if (!m_connections_initialized && valid_metadata()) init();
```
arvidn: there should be a test covering this fix

PR #4120, opened 2019-11-24
[WIP] added support for wolfSSL via openssl compatibility layer

aldenml:


aldenml:
hi @arvidn, still facing some errors running `test_ssl`, but just wanted to have your first opinion on this approach

include/libtorrent/aux_/export.hpp
```diff
@@ -69,9 +69,12 @@ POSSIBILITY OF SUCH DAMAGE.
 #elif TORRENT_USE_CRYPTOAPI
 # define TORRENT_CRYPTO_NAMESPACE inline namespace capi {
 # define TORRENT_CRYPTO_NAMESPACE_END }
-#elif defined TORRENT_USE_LIBCRYPTO
+#elif defined TORRENT_USE_LIBCRYPTO && !defined TORRENT_USE_WOLFSSL
```
arvidn: could you avoid this `&& !`-condition by moving the `USE_WOLFSSL` check ahead of it?

Jamfile
```diff
@@ -511,10 +553,20 @@ feature.compose <encryption>off : <define>TORRENT_DISABLE_ENCRYPTION ;
...
-feature crypto : built-in openssl libcrypto gcrypt : composite propagated ;
+feature crypto : built-in openssl libcrypto gcrypt wolfssl : composite propagated ;
 feature.compose <crypto>openssl : <define>TORRENT_USE_LIBCRYPTO <define>TORRENT_USE_OPENSSL <define>OPENSSL_NO_SSL2 ;
 feature.compose <crypto>libcrypto : <define>TORRENT_USE_LIBCRYPTO ;
 feature.compose <crypto>gcrypt : <define>TORRENT_USE_LIBGCRYPT ;
+feature.compose <crypto>wolfssl
+	: <define>TORRENT_USE_WOLFSSL
+	<define>TORRENT_USE_LIBCRYPTO
+	<define>TORRENT_USE_OPENSSL
+	<define>OPENSSL_NO_SSL2
```
arvidn: what does this do? I assume it disables support for obsolete SSL2. Should this be set for normal openssl builds as well?

Jamfile
```diff
@@ -511,10 +553,20 @@ feature.compose <encryption>off : <define>TORRENT_DISABLE_ENCRYPTION ;
...
 
-feature crypto : built-in openssl libcrypto gcrypt : composite propagated ;
+feature crypto : built-in openssl libcrypto gcrypt wolfssl : composite propagated ;
 feature.compose <crypto>openssl : <define>TORRENT_USE_LIBCRYPTO <define>TORRENT_USE_OPENSSL <define>OPENSSL_NO_SSL2 ;
 feature.compose <crypto>libcrypto : <define>TORRENT_USE_LIBCRYPTO ;
 feature.compose <crypto>gcrypt : <define>TORRENT_USE_LIBGCRYPT ;
+feature.compose <crypto>wolfssl
+	: <define>TORRENT_USE_WOLFSSL
+	<define>TORRENT_USE_LIBCRYPTO
+	<define>TORRENT_USE_OPENSSL
```
arvidn: so, wolfSSL provides openssl headers? doesn't that conflict with openssl? can these libraries not be installed at the same time?
Do you think it would be impractical to, instead of defining both `TORRENT_USE_OPENSSL` and `TORRENT_USE_WOLFSSL`, change the code to check for both macros? part of me feels like that would be a bit cleaner, just in case there are deviations at some places.

Jamfile
```diff
@@ -511,10 +553,20 @@ feature.compose <encryption>off : <define>TORRENT_DISABLE_ENCRYPTION ;
...
 feature.compose <mutable-torrents>off : <define>TORRENT_DISABLE_MUTABLE_TORRENTS ;
 
-feature crypto : built-in openssl libcrypto gcrypt : composite propagated ;
+feature crypto : built-in openssl libcrypto gcrypt wolfssl : composite propagated ;
 feature.compose <crypto>openssl : <define>TORRENT_USE_LIBCRYPTO <define>TORRENT_USE_OPENSSL <define>OPENSSL_NO_SSL2 ;
 feature.compose <crypto>libcrypto : <define>TORRENT_USE_LIBCRYPTO ;
 feature.compose <crypto>gcrypt : <define>TORRENT_USE_LIBGCRYPT ;
+feature.compose <crypto>wolfssl
+	: <define>TORRENT_USE_WOLFSSL
+	<define>TORRENT_USE_LIBCRYPTO
```
arvidn: I don't think we would need to rely on libcrypto emulation, right? does anything fail if you don't define this?

include/libtorrent/aux_/export.hpp
aldenml: yes

Jamfile
aldenml: it is already set when using openssl

Jamfile
aldenml: well, they only achieved that compatibility with `asio` (and others) providing the emulation layer, so for this case, the openssl headers will (and need) to be there. I also think that checking for both macros is very aggressive on the code, and we can always check for `TORRENT_USE_WOLFSSL` for special cases.

Jamfile
aldenml: well, is either this or incorporate your commit :), and the main benefit of using wolfSSL is using it all (for SSL and Crypto)

aldenml: and for what I see there is no "libcrypto emulation", just "extra" and "all" openssl emulation, which both are needed for ASIO.

arvidn: I see. but it emulates libcrypto API in the sense that the SHA-1 and SHA-256 etc. algorithm can be used as if we had libcrypty, right? because you're not linking against libcrypto in your branch

Jamfile
arvidn: I see. and so far there is not a single need to check `TORRENT_USE_WOLFSSL`, is that right? If the openssl emulation is accurate enough to "just work", we should probably not even define `TORRENT_USE_WOLFSSL` either, right?

Jamfile
aldenml: true

Jamfile
aldenml: right, except that I assume `define TORRENT_CRYPTO_NAMESPACE inline namespace wcrypto {` should be accurate (or clear) for clients, or not?

aldenml:
Hi @arvidn, do you have any idea of how to troubleshoot this? maybe I'm missing a flag that allows the peer certificate to be accepted? Thanks
```
MALICIOUS PEER TEST: valid-certificate valid-SNI-hash valid-bittorrent-hash  port: 17831
set_password_callback
use_certificate_file "../ssl/peer_certificate.pem"
use_private_key_file "../ssl/peer_private_key.pem"
use_tmp_dh_file "../ssl/dhparams.pem"
connecting 127.0.0.1:17831
00:00:01.100: ses1: [torrent_checked] temporary checked
00:00:01.100: ses1: [torrent_log] temporary: *** announce: no trackers
SNI: dd6c5dfd1a00881f8137f26f38444332cd2580f1
SSL handshake
00:00:01.101: ses1: [log]  *** peer SSL handshake done [ ip: 127.0.0.1:54316 ec: non-blocking socket write buffer full socket: SSL/TCP ]
00:00:01.101: ses1: [peer_error]  -  peer [ 127.0.0.1:54316 client: Unknown ] peer error [ssl_handshake] [asio.ssl]: non-blocking socket write buffer full
--- peer_errors: 1 ssl_disconnects: 1
Failed SSL handshake: Connection reset by peer

***** test_ssl.cpp:606 "TEST_ERROR: success: 0 expected: 1" *****
```

arvidn:
as you may have figured out, "non-blocking socket write buffer full" is the relevant error. This doesn't happen with openssl. Maybe there's an option to increase the buffer size wolfSSL uses internally?
I imagine that if the buffer is smaller than the size of one TLS frame, there would be a problem, but it can't be that obvious. The OpenSSL interface to read and write to a socket are quite non-obvious, and a read can sometimes trigger a write of the underlying socket and vice versa. The logic in `boost::asio::ssl::stream` deals with all that, and I suspect that's where this error is originating from.

aldenml:
@arvidn can you indicate to me how you generate the pem files for the tests, after some debugging I have the impression that could be a problem if it's not compatible with wolfSSL, Thanks!

arvidn:
yes, I generate those with this script: https://github.com/arvidn/libtorrent/blob/RC_1_2/test/ssl/regenerate_test_certificate.sh
and I haven't figured out how to produce certificates that last longer than a year, so every november (or so) I have to regenerate them :)

aldenml:
well, definitely something is wrong while loading `peer_private_key.pem`, it's like wolfSSL is unable to parse the ASN.1 inside.

For reference, these are the flags that I need so far: `--enable-asio  --enable-des3 --enable-sha`, I can hunt the error to the point it starts parsing and fails.

aldenml:
Hi @arvidn, I just notice this comment
```cpp
// accept connections initializing the SSL connection to
// use the generic m_ssl_ctx context. However, since it has
// the servername callback set on it, we will switch away from
// this context into a specific torrent once we start handshaking
return socket_type(ssl_stream<tcp::socket>(tcp::socket(std::move(s)), m_ssl_ctx));
```
at https://github.com/arvidn/libtorrent/blob/b61c492b376b5b0d2f5037f8bf4d63f0c8d0cc7c/src/session_impl.cpp#L2723

Can you help me find where that switch is made? Thanks

arvidn:
sure. It should be in the vicinity of interpreting the SNI. SNI is normally used on the web to specify which hostname the browser is connecting to, so a multi-tenant web server can host multiple sites and serve up the correct certificate. So this happens very early on in the TLS handshake.

arvidn:
right here: https://github.com/arvidn/libtorrent/blob/0c2d3a0dac38276f5c5518bc35b7c0c84a25bf06/src/session_impl.cpp#L387

aldenml:
interesting, I just put a breakpoint there and it's not called. This is in the context of `test_malicious_peer`. I'm able to load the private keys after moving to `DER` and `ASN1`, but the switch of the context does not happen before `m_sock->async_handshake(ssl::stream_base::server....`, any idea? any other place that I can force that switching?

arvidn:
ah, it's probably because WolfSSL has a different way of setting the SNI of an outgoing connection. And maybe a different way of setting the SNI-callback too. That's where I would look.
Both of those are "extensions" in openssl, but they are critical and must exist (in some form)

arvidn:
i.e. you probably don't get here: https://github.com/arvidn/libtorrent/blob/0c2d3a0dac38276f5c5518bc35b7c0c84a25bf06/src/session_impl.cpp#L487

nor here: https://github.com/arvidn/libtorrent/blob/057c6684f20065176926afa21e01a4473a551e88/src/socket_type.cpp#L127

Note that the test use these functions too

PR #4133, opened 2019-11-30
 added support for wolfssl with new bjam option

aldenml:


CMakeLists.txt
```diff
@@ -54,7 +54,7 @@ set(libtorrent_include_files
 	enum_net
 	error
 	error_code
-	extensions
+	extensions.hpp
```
arvidn: is this right?

docs/manual.rst
```diff
@@ -923,7 +923,7 @@ libtorrent's point of view, it doesn't matter what it is. libtorrent only makes
 sure the peer certificates are signed by the correct root certificate.
 
 One way to create the certificates is to use the ``CA.sh`` script that comes
-with openssl, like thisi (don't forget to enter a common Name for the
+with openssl, like these (don't forget to enter a common Name for the
```
arvidn: I just landed a patch in RC_1_2 fixing these typos. this will conflict with that

src/peer_connection.cpp
```diff
@@ -84,10 +84,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <set>
 #endif
 
-#ifdef TORRENT_USE_OPENSSL
-#include <openssl/rand.h>
-#endif
-
```
arvidn: was this just unused in here?

arvidn:
I'm thinking that there's a simpler way of doing this, perhaps some other way to define the macros to make them a bit more orthogonal.

Also, I don't think this actually solves your problem with WolfSSL all the way. You didn't address setting the SNI to the hostname for HTTPS connections

CMakeLists.txt
aldenml: yes, it's something that have been bothering me for a long time. CLion (and `cmake`) pick the folder `include/extensions` as project element, this way only `extensions.hpp` is picked up.

docs/manual.rst
aldenml: will revert this typo fix then

src/peer_connection.cpp
aldenml: yes, unusued :)

CMakeLists.txt
arvidn: interesting. The purpose of this list of header files is only for creating a package, right?
I just realized there's no logic that adds the ".hpp" suffix to these names. is that implied by cmake?
when building a package with cmake, does it still include the `extensions` directory? I imagine it would need to be included too

aldenml: Yes, `cmake` performs all kind of extensions guessing until it finds (or not) the file. The files under `extensions` are still included with the logic associated to `libtorrent_extensions_include_files`.

.travis.yml
```diff
@@ -23,6 +23,7 @@ matrix:
     - env: variant=test_debug tests=1 toolset=gcc sanitizer=on fuzzers=1
     - env: variant=test_debug tests=1 mmap=off toolset=gcc
     - env: variant=test_debug sim=1 crypto=openssl toolset=gcc sanitizer=on
+    - env: variant=test_debug tests=1 crypto=openssl-https toolset=gcc sanitizer=on
```
aldenml: btw, I have plans to reuse this job for wolfSSL

CMakeLists.txt
arvidn: it seems reasonable then to add the `.hpp` extension to all files here, what do you think?

aldenml: hmm... maybe in the future :)

arvidn: I think it would make sense to do this as a separate patch. there's nothing in this PR that depends on this fix, is there?

arvidn:
have you verified that the SNI is set on outgoing HTTPS connections?
currently those calls are conditioned by having a new enough version of openssl, so I suspect they are just compiled out using WolfSSL

arvidn:
I think you have to add "https" to the dictionary: https://travis-ci.org/arvidn/libtorrent/jobs/618948960#L764

CMakeLists.txt
aldenml: no, removing this `.hpp` from this PR

aldenml:
I'm attaching images of my IDE debug stepping into SNI code
Screen Shot 2019-11-30 at 2 37 31 PM
and (see `ret = 1`)
Screen Shot 2019-11-30 at 2 39 38 PM

What do you think about the new `TORRENT_USE_OPENSSL_HTTPS_OR_SSL`?

aldenml:
@arvidn do you have an idea of what's happening with https://travis-ci.org/arvidn/libtorrent/jobs/619124602?utm_medium=notification&utm_source=github_status

arvidn:
some of those tests are flaky. at some point I should try to make the more reliable

arvidn:
`TORRENT_USE_OPENSSL_HTTPS_OR_SSL` seems a bit long, and yet not very clear what it means.
What would you think of instead separating out the support for bittorrent over SSL as a separate build configuration that requires SSL? (there could even be a friendly error message if SSL wasn't enabled when trying to enable torrents over SSL).

Something like `TORRENT_BITTORRENT_SSL` or `TORRENT_SSL_PEERS`. That way, we could leave the existing `TORRENT_USE_OPENSSL` as is (I think), to mean to include the basic support for SSL for trackers and web seeds.

Another alternative would be to make it mandatory to build with SSL support, and remove the `TORRENT_USE_OPENSSL` define entirely.

aldenml:
Hi @arvidn, what do you think now with `TORRENT_SSL_PEERS` and `openssl-no-peers`? I know `openssl-no-peers` is not the most beautiful option name, but I fear to break a lot of builds setups and also complicate the build options. Btw, I expect most people to just use `openssl` and not bother with other options.

aldenml:
@arvidn, before fixing the conflict and trigger another CI, what do you think about now that we have a way to separate the Torrents over SSL code,  keeping `openssl` for everything, and moving the `wolfssl` to this PR without the Torrents over SSL?

arvidn:
yes, I think this is a good approach. Would you mind adding `#error` if `TORRENT_SSL_PEERS` is enabled but the openSSL version is too old to support SNI? maybe with a friendly message suggesting this new build configuration `openssl-no-peers`.

Also, if `TORRENT_SSL_PEERS` is set, but `TORRENT_USE_OPENSSL` is *not*, that should also trigger a `#error` with some friendly message.

arvidn:
I would also be OK with making it mandatory to build with a new enough OpenSSL version where SNI is supported.

aldenml:
btw, and sorry if I wasn't clear enough, I want to remove `openssl-no-peers`, and bring `wolfssl`, so it will be only:
`openssl` - everything
`wolfssl` - only https
and the `#error` conditions of your comments. Thoughts?

arvidn:
ah, I see. So, you're thinking a separate define, next to `TORRENT_USE_OPENSSL` for WolfSSL?
so some of the `#ifdef TORRENT_USE_OPENSSL` would also check for your new define?

arvidn:
oh, maybe I misunderstood again. You want to keep the current defines/macros, but you want to rename just the build feature?

aldenml:
correct :), just trying to keep the changes to the minimum, but you tell me

arvidn:
that's fine

aldenml:
rebased, added the wolfSSL here, and added the two #error messages, waiting for CI to complete

arvidn:
it looks like there's something in the random number generator in Wolf causing the current code to fail:

https://travis-ci.org/arvidn/libtorrent/jobs/621878109#L3916

Maybe there's some missing initialization or something.

aldenml:
Fixed the random number generator, there is a subtle API incompatibility, I documented that in the code.

arvidn:
> Now, the arm build is failing consistently, and I can't see the reason for it (maybe the new boost?)

Interestingly, the first test passes. I wonder if it leaves the container in some state that causes the next test to fail.

https://travis-ci.org/arvidn/libtorrent/jobs/622050005#L771

aldenml:
I can confirm that using boost 1.68 builds without problem. Now this is a completely unrelated problem to the wolfSSL integration :( I will take a look at that

aldenml:
travis is green!, appveyor should be fine, waiting for your review

.travis.yml
```diff
@@ -341,5 +352,7 @@ script:
   - cd $TRAVIS_BUILD_DIR/test;
   - 'if [[ "$arch" == "arm" ]];
     then
-      ${B2} arm-tests warnings-as-errors=on warnings=all crypto=$crypto variant=test_arm $toolset target-os=linux link=static testing.launcher="sudo cp -R bin rootfs/; sudo chroot rootfs";
+      ${B2} arm-tests warnings-as-errors=on warnings=all crypto=$crypto variant=test_arm $toolset target-os=linux link=static testing.execute=off;
+      sudo cp -R bin rootfs/;
+      ${B2} arm-tests warnings-as-errors=on warnings=all crypto=$crypto variant=test_arm $toolset target-os=linux link=static testing.launcher="sudo chroot rootfs";
```
arvidn: now that building and running tests are separated, you could even add a `travis_retry` here. (not important though, and I think only deterministic tests are part of the ARM tests)

include/libtorrent/aux_/openssl.hpp
```diff
@@ -101,6 +105,6 @@ inline GENERAL_NAME* openssl_general_name_value(GENERAL_NAMES* gens, int i)
 
 #include "libtorrent/aux_/disable_warnings_pop.hpp"
 
-#endif // TORRENT_USE_OPENSSL
+#endif // TORRENT_SSL_PEERS
```
arvidn: This comment doesn't look right. I would expect the corresponding `#if` to also have changed in this file

src/session_impl.cpp
```diff
@@ -543,6 +544,8 @@ void apply_deprecated_dht_settings(settings_pack& sett, bdecode_node const& s)
 #ifdef TORRENT_USE_OPENSSL
 		error_code ec;
 		m_ssl_ctx.set_verify_mode(boost::asio::ssl::context::verify_none, ec);
+#endif
+#ifdef TORRENT_SSL_PEERS
 #if OPENSSL_VERSION_NUMBER >= 0x90812f
```
arvidn: this test can be removed now, right?

arvidn: and any similar tests, except for the one you added for the `#error`

src/socket_type.cpp
```diff
@@ -279,7 +279,7 @@ namespace aux {
 		TORRENT_UNUSED(holder);
 		error_code e;
 		s.close(e);
-#endif // TORRENT_USE_OPENSSL
+#endif // TORRENT_SSL_PEERS
```
arvidn: this comment changed, but the corresponding `#if` didn't

test/test_session.cpp
```diff
@@ -333,7 +333,7 @@ TORRENT_TEST(save_state_fingerprint)
 auto const count_dht_inits = [](session& ses)
 {
 	int count = 0;
-	int num = 120; // this number is adjusted per version, an estimate
+	int num = 200; // this number is adjusted per version, an estimate
```
arvidn: this seems unrelated

arvidn:
lgtm. please squash these. However, I think the independent changes should go into separate commits. specifically removing the blank lines at the end of the tests should be a separate commit, and also the independent removal of unused openssl includes as well as the unused `placeholders::_1` using declaration.

.travis.yml
aldenml: yes, I will add it

src/socket_type.cpp
aldenml: your are right

src/session_impl.cpp
aldenml: yes, I will remove them

include/libtorrent/aux_/openssl.hpp
aldenml: you are right

test/test_session.cpp
aldenml: I was having problems in my env that required this change, will revert it

aldenml:
good, actually, I will create separate PRs for a lot of changes here to strip this one from the non essentials ones.

aldenml:
hi @arvidn, can you restart the travis job? thanks

docs/manual.rst
```diff
@@ -886,7 +886,7 @@ This setting is only taken into account when the normal listen socket is opened
 socket). To not listen on an SSL socket at all, set ``ssl_listen`` to 0.
 
 This feature is only available if libtorrent is build with openssl support
-(``TORRENT_USE_OPENSSL``) and requires at least OpenSSL version 1.0, since it
+(``TORRENT_SSL_PEERS``) and requires at least OpenSSL version 1.0, since it
```
arvidn: it seems like this should probably mention both defines, or the text should say something like "built with SSL torrent support"

include/libtorrent/config.hpp
```diff
@@ -116,7 +116,7 @@ POSSIBILITY OF SUCH DAMAGE.
 // on OSX, use the built-in common crypto for built-in
 # if !defined TORRENT_USE_LIBCRYPTO && !defined TORRENT_USE_LIBGCRYPT
 #  define TORRENT_USE_COMMONCRYPTO 1
-# endif // TORRENT_USE_OPENSSL
+# endif
```
arvidn: was this comment incorrect?

arvidn:
one thing that makes me a little bit nervous is that the headers included when building against wolfSSL have the same name as when linking against openssl. It seems like begging for including the wrong headers.

I imagine if you build against wolfSSL on a system that has openssl "installed", it just won't work. Is that right?

docs/manual.rst
aldenml: will change

include/libtorrent/config.hpp
aldenml: I think it's incorrect, the if is not with that macro

aldenml:
agree with the fact that this is the perfect recipe for build problems, I looked at this issue from the beginning, but found that the problem begins in `asio`, if you look at https://github.com/boostorg/asio/blob/develop/include/boost/asio/ssl/detail/openssl_types.hpp there is nothing a client library can do (to the best of my knowledge).

I guess this just a "common problem" with C/C++ :)

aldenml:
@arvidn, I don't want to take CI bandwidth for a change in `manual.srt`, do you mind if I do that in another PR?

aldenml:
one travis job restart when you have a chance...thanks

arvidn:
are you sure `test_url_seed` passes? it's failed 6 times in a row now

aldenml:
what can I say :) (green now), but I noticed more random errors lately

PR #4135, opened 2019-11-30
move session_params into its own header/cpp pair

arvidn:


include/libtorrent/session_params.hpp
```diff
@@ -0,0 +1,131 @@
...
+	dht::dht_state dht_state;
+
+	dht::dht_storage_constructor_type dht_storage_constructor;
+
+	disk_io_constructor_type disk_io_constructor;
+
+	// this container can be used by extensions/plugins to store settings. It's
+	// primarily here to make it convenient to save and restore state across
+	// sessions, using read_session_params() and write_session_params().
+	std::map<std::string, std::string> ext_state;
```
aldenml: 👍

arvidn: heh. technically, this part belongs in a future patch

PR #4154, opened 2019-12-05
fix buffer overflow in SOCKS5 UDP logic

arvidn:


sledgehammer999:
I have a user on macOS that has experienced a crash with very similar stacktrace. And now he faces a crash each time he adds a magnet and this happens only when he uses SOCKS5. Do you think the "adding a magnet" codepath (under SOCKS5) is different than the one you fix here? In any case I'll provide to him a test build of this PR probably tomorrow.

PS: Here is his magnet crash stacktrace(https://github.com/qbittorrent/qBittorrent/issues/11564#issuecomment-562231453). **Caution:** That bug report contains many huge stacktraces. Go to the comment I linked.

arvidn:
This PR specifically fixes the SOCKS5 path when sending a UDP packet to a hostname (as opposed to an IP address). This happens for UDP trackers mostly. Whether the tracker came from a magnet link or a .torrent file shouldn't matter

PR #4165, opened 2019-12-08
rename disk_io_thread -> mmap_disk_io and storage -> mmap_storage

arvidn:


arvidn:
files like `disk_io_job.hpp`, `disk_io_thread_pool.hpp` and `aux_/disk_job_fence.hpp` are also specific to the `mmap_disk_io`. Perhaps they should also be renamed to indicate this.

PR #4166, opened 2019-12-09
only trigger default listen sockets if the listen_interfaces setting …

ssiloti:
…is empty

The old code would add default listen sockets when listen_interfces was
non-empty but only contained invalid interface specifications.

In the future it may be possible to support operation with no listen sockets
at all, but this will require additional work since the announce code currently
requires a listen socket to function.

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -1822,16 +1822,6 @@ namespace aux {
 			expand_unspecified_address(ifs, eps);
 		}
 
-		// if no listen interfaces are specified, create sockets to use
-		// any interface
-		if (eps.empty())
-		{
-			eps.emplace_back(address_v4(), 0, "", transport::plaintext
-				, duplex::only_outgoing);
-			eps.emplace_back(address_v6(), 0, "", transport::plaintext
-				, duplex::only_outgoing);
-		}
-
 		auto remove_iter = partition_listen_sockets(eps, m_listen_sockets);
 
 		while (remove_iter != m_listen_sockets.end())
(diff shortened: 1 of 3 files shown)


src/session_impl.cpp
```diff
@@ -5227,7 +5217,16 @@ namespace aux {
...
-		m_listen_interfaces = parse_listen_interfaces(net_interfaces);
+		if (!net_interfaces.empty())
+		{
+			m_listen_interfaces = parse_listen_interfaces(net_interfaces);
+		}
+		else
+		{
+			// if no listen interfaces are specified, create sockets to use
+			// any interface
+			m_listen_interfaces = parse_listen_interfaces("0.0.0.0:0,[::]:0");
```
arvidn: I think the `duplex::only_outgoing` flag in the previous code is relevant.

arvidn: I seem to recall that the old behavior tried to mimic, as closely as possible, not having any listen sockets. But I think the name is a bit misleading as it is. Since we use these for outgoing connections also. And I think it makes sense to preserve this attempt to not accept incoming connections for an empty string.

But doesn't this still raise the question of what happens when a torrent attempts to announce when there are literally no sockets to use?

arvidn:
as far as I can tell, opening the "default listen sockets" is as close as we can get to not opening any at all, since we use these sockets for outgoing connections as well. But the existing code sets the `duplex` to outgoing only, which (I believe) effectively makes them not listen for incoming connections.

Hence, I think the current code is more correct; if you don't specify any listen interfaces, you won't receive incoming connections. If you have invalid syntax for the listen interfaces, you also won't receiving incoming connections, but at least you'll get errors now.

PR #4167, opened 2019-12-09
bump C++ version to 17

arvidn:


arvidn:
@aldenml In order to get C++17 on ARM, I upgraded the linaro package version, which seems to have made a change in the signature of some intrinsics. Do you have any thoughts about how to best fix this?

https://travis-ci.org/arvidn/libtorrent/jobs/623470301#L515

I imagine we would stay backwards compatible with the previous signature too.

aldenml:
@arvidn at first sight, I have no idea, I will look at it in details

arvidn:
it seems like it's fixed for GCC 8.2. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81497#c2

aldenml:
maybe use this https://developer.arm.com/-/media/Files/downloads/gnu-a/8.3-2019.03/binrel/gcc-arm-8.3-2019.03-x86_64-arm-linux-gnueabihf.tar.xz?revision=e09a1c45-0ed3-4a8e-b06b-db3978fd8d56&la=en&hash=93ED4444B8B3A812B893373B490B90BBB28FD2E3 from https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-a/downloads

arvidn:
@aldenml did you figure anything out when you had this symptom? https://travis-ci.org/arvidn/libtorrent/builds/624169970#L684

aldenml:
@arvidn I was looking at this, and I think I can do some tests today. I think the problem starts at L667 with
`cp: cannot overwrite non-directory 'rootfs/bin' with directory 'bin'`
maybe the new base image has some `bin` that should be deleted first

aldenml:
You managed to fix the arm test!! I was not even close to considering all the combinations you tested :)

arvidn:
yeah, the mingw build seems hopeless though...

paullouisageneau:
Can I help with the transition somehow?

arvidn:
I will rebase this on top of latest master soon. I need to merge a few more patches in from `RC_1_2` first. I suspect most of the remaining work on this is to ensure all the CI pass. And I think that might include not going all the way to C++17. I get the impression Apple is lagging behind in their support still.

PR #4171, opened 2019-12-11
fix erroneous event=completed tracker announce when checking files

arvidn:


arvidn:
@ssiloti does this look reasonable?
As I explained in the associated ticket, the problem is that the `announce_endpoint`s are added lazily, so when the `complete_sent` are set to false, they aren't there yet.

include/libtorrent/announce_entry.hpp
```diff
@@ -194,14 +194,16 @@ TORRENT_VERSION_NAMESPACE_2
 		// from this tracker.
 		bool verified:1;
 
+		// set to true if we've already sent an event=completed to this tracker
+		bool complete_sent:1;
```
ssiloti: This should be a per-torrent rather than a per-tracker property. As-is this flag will not be set on trackers which get added  later via `torrent::add_tracker` or `torrent::replace_trackers`. Also, 5e54d13dec93368884a59c7633b1fe66155e2512 indicates that the intent is for trackers added after the torrent has completed by any means to not send a completed event. To honor that the flag should be set when the torrent completes downloading.

arvidn: You're right that often you'd want this state to be saved per torrent. I don't think it's quite that clear-cut. Maybe it should really be per tracker *tier*. At least the way tiers are used, they are independent, and we probably *should* send `event=completed` to each of them

arvidn: I think what makes this less obvious is that trackers serve two purposes; (1) maintain the swarm and introduce peers to each other. (2) report statistics to the tacker.

(1) is really about a listen socket that wants to join the swarm, whereas (2) is really per client.

For example: if I have one IPv4 and one IPv6 listen socket, I announce to the tracker from each. Should both of those announces contain the `event=completed` message?

Right now, libtorrent says "yes". Which I can imagine is surprising to some. However, if one would say "no" instead, you may run into the opposite issue that, you may accidentally drop an announce.

But it probably makes sense to alter how this is handled in some way.

arvidn:
I'm interested in opinions on this. I'm leaning towards this patch for the 1.2.3 release. Mostly because it's the most similar to 1.1 behavior. Even though this is not ideal behavior, I think it probably shouldn't hold up cutting 1.2.3

include/libtorrent/announce_entry.hpp
an0n666: Event_complete should be sent only once and that is when the torrent completes downloading. If I had to change the tracker for some reason i.e. tracker passkey change/passkey added after uploading a torrent- it would send event_complete as I just added a new tracker which is actually the same as previous one except the passkey.

ssiloti: I think it makes sense to send the event on all listen sockets. If the tracker really cares about keeping accurate stats it can simply ignore redundant events from the same peer.

The change I suggested was really about whether the completion event is edge or level triggered. The flag this PR adds only takes effect when an announce_endpoint is created, so it does not suppress completed events on trackers/endpoints created before the flag is set. This means that in the common case where the set of tackers and listen sockets is static the flag will only suppress the completed event if it is set immediately on torrent creation. Setting the flag on download completion would only suppress the completed event on trackers/endpoints added after the fact.

arvidn: @ssiloti I'm not sure I quite follow. Note that I force-pushed an update to this patch after your first comment "fixing" the case of the user calling `replace_trackers()` too. Fixing as in making it behave the same as libtorrent-1.1.x.

Is your suggestion: instead of introducing the `complete_sent` to the `announce_entry`, introduce it to the `torrent` object. Other than that, maintain the same behavior? The main difference (afaict) is that there need not be a special case for `replace_trackers()`.

Actually, I think I see one of your points; the `announce_entry::complete_sent` should be set to true after at least one tracker announce with `event=completed` comes back successful. And in your suggestion, that variable would live in `torrent` instead of `announce_entry`. Do I understand you correctly?

I think that makes sense, and in fact, re-introducing `announce_entry::complete_sent` is a bit shady as it would have been an ABI break if it wasn't for the bitfields having spare bits.

ssiloti: Ah sorry about that, I missed your force-push. You did address my issue with adding trackers.

The other issue has to do with how new listen sockets are handled. If a new listen socket is added while a torrent is seeding presumably it should mirror the behavior when adding a tracker and not send a completion event if the torrent is already seeding. Right now if a tracker was added before the download completed its `complete_sent` flag will not be set, so any new listen sockets will send `event=completed` in their first announce. If however the tracker was added while seeding no completion event will be sent on the new socket. This is a strange inconsistency which IMO should be addressed. Setting the flag on all trackers immediately upon torrent completion would work. The flag only affects new `announce_endpoint`s so it will not suppress `event=completed` on any of the existing sockets.

Whether the flag lives in `announce_entry` or `torrent` doesn't make a big difference. If the behavior I describe above is implemented then the value should always be the same for all trackers, so it would be a bit odd to have it in each `announce_entry`.

arvidn: I think the case of adding a new listen socket is addressed in this patch. As far as I understand, a new listen socket will cause a new `announce_endpoint` to be added right before sending the announce, at which point the new `announce_endpoint` is constructed to copy the `complete_sent` from the `announce_entry`, here: https://github.com/arvidn/libtorrent/pull/4171/files#diff-19929347e18a6ff1fc37f09ec720306bR2928

If you add a new listen socket *before* completing, the first announce wouldn't have `event=completed` in it, but once you complete it would. I don't really see how this could be an issue. Unless you're talking about tiers and groups of trackers, which I think is still behaving strange. If a tracker hasn't been used, because we started using another tracker in the same tier, and then we switch over, the `complete_sent` will be false on the previously unused tracker and it would send `event=completed` (assuming we're seeding).

I think I should move this state to the torrent, and make *any* successful announce with `event=completed` sets it to true. That would also address my concern of altering the supposed ABI-stable `announce_endpoint`.

arvidn: oh, I get it now, the `complete_sent` in `announce_entry` isn't updated by a successful announce with `event=completed`. You're right.

an0n666: @arvidn 
I justed tested with the qbit build @sledgehammer999 provided and it seem to be working as expected now.
Added torrent > qBit rechecked > didnt send event_complete
Edited tracker url > didnt send event_complete
Added new tracker > didnt send event_complete

arvidn: Thanks for testing! I’m working on a refinement of this patch that will fix the issue of multiple trackers in the same tier as well, and add unit tests

src/announce_entry.cpp
```diff
@@ -47,13 +47,13 @@ namespace libtorrent {
...
 
-	announce_endpoint::announce_endpoint(aux::listen_socket_handle const& s)
+	announce_endpoint::announce_endpoint(aux::listen_socket_handle const& s, bool const completed)
 		: local_endpoint(s ? s.get_local_endpoint() : tcp::endpoint())
 		, socket(s)
 		, fails(0)
 		, updating(false)
 		, start_sent(false)
-		, complete_sent(false)
+		, complete_sent(completed)
```
arvidn: technically this isn't necessary, since the condition for sending `event=completed` now also requires `torrent::m_complete_sent` to be false. But since this field is user-facing, it might be a good idea to keep it up-to-date

src/torrent.cpp
```diff
@@ -2925,7 +2926,7 @@ bool is_downloading_state(int const st)
 					return;
 				}
 
-				ae.endpoints.emplace_back(s);
+				ae.endpoints.emplace_back(s, bool(m_complete_sent));
```
arvidn: a bitfield cannot bind to a universal reference, but a temporary `bool` can

src/torrent.cpp
```diff
@@ -2986,7 +2987,12 @@ bool is_downloading_state(int const st)
 				if (req.event == tracker_request::none)
 				{
 					if (!aep.start_sent) req.event = tracker_request::started;
-					else if (!aep.complete_sent && is_seed()) req.event = tracker_request::completed;
+					else if (!m_complete_sent
```
arvidn: this is a material change, requiring `m_complete_sent` to be false.

src/torrent.cpp
```diff
@@ -3228,7 +3238,13 @@ bool is_downloading_state(int const st)
 				if (!aep->start_sent && r.event == tracker_request::started)
 					aep->start_sent = true;
 				if (!aep->complete_sent && r.event == tracker_request::completed)
+				{
 					aep->complete_sent = true;
+					// we successfully reported event=completed to one tracker. Don't
+					// send it to any other ones from now on (there may be other
+					// announces outstanding right now though)
+					m_complete_sent = true;
```
arvidn: this is a material change, setting the torrent-wide `m_complete_sent` when *any* tracker successfully announces with `event=completed`

arvidn:
ok, I'm pretty happy with this patch now

PR #4188, opened 2019-12-18
add userdata to torrent_alert

ghbplayer:
"add userdata to torrent_alert to skip searching for client data associated with torrent on every alert"

LT clients have data associated with torrents. Without that data, the alert system would be nearly worthless. Clients can associate their data with LT torrents via lt::torrent_handle, but the reverse association requires searching by comparing pointers or hashes. Efficient systems such as Windows Overlapped I/O and Linux epoll will accept a user pointer and provide that user pointer on events to help clients find their associated data.

With this change, LT alerts have a similar user data pointer so that every alert doesn't result in the client performing a binary search for associated data. This is nearly free in most cases and thus the client doesn't have to make a synchronous call to retrieve the pointer from the main LT thread.

### DIFF
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -112,7 +112,7 @@ TORRENT_VERSION_NAMESPACE_2
 	struct TORRENT_EXPORT torrent_alert : alert
 	{
 		// internal
-		torrent_alert(aux::stack_allocator& alloc, torrent_handle const& h);
+		torrent_alert(aux::stack_allocator& alloc, torrent_handle const& h, void* userdata);
 		torrent_alert(torrent_alert&&) noexcept = default;
 
 #if TORRENT_ABI_VERSION == 1
@@ -127,6 +127,11 @@ TORRENT_VERSION_NAMESPACE_2
 		torrent_handle handle;
 
 		char const* torrent_name() const;
+
+		// User data field from add_torrent_params. May be used for any
+		// purpose including finding client data associated with this
+		// torrent without performing a search
+		void* userdata;
 
 	protected:
 		std::reference_wrapper<aux::stack_allocator const> m_alloc;
@@ -144,7 +149,7 @@ TORRENT_VERSION_NAMESPACE_2
 	{
 		// internal
 		peer_alert(aux::stack_allocator& alloc, torrent_handle const& h,
-			tcp::endpoint const& i, peer_id const& pi);
+			tcp::endpoint const& i, peer_id const& pi, void* userdata);
 		peer_alert(peer_alert&& rhs) noexcept = default;
 
 #if TORRENT_ABI_VERSION == 1
@@ -172,7 +177,7 @@ TORRENT_VERSION_NAMESPACE_2
 	{
 		// internal
 		tracker_alert(aux::stack_allocator& alloc, torrent_handle const& h
-			, tcp::endpoint const& ep, string_view u);
+			, tcp::endpoint const& ep, string_view u, void* userdata);
 
 #if TORRENT_ABI_VERSION == 1
 		static const int TORRENT_DEPRECATED_MEMBER alert_type = 2;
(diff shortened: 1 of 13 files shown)


arvidn:
A change that alters the ABI cannot be made against `RC_1_2`. It's the current stable branch, and is meant to stay ABI compatible with previous 1.2 releases. A patch like this could be made against `master` though.

I have a few comments:

Using a `void*` does not seem appropriate as it invites type- and resource errors. Did you consider a safer alternative like `std::shared_ptr<void>` or `std::any`? I imagine `std::any` is probably the most appropriate (technically it would have to be `boost::any` for now, since libtorrent is still on C++14).

I don't think the alternative to having a user data pointer is a binary search. You can use the `torrent_handle` as the key in a hash map, to get O(1) look-ups. That said, there may still be good reasons to have even more efficient (and perhaps simpler) ways of associating data with torrents.

You add the user data field to `torrent_alert`, but you don't pass it along with all such alerts. Why is that? It would seem reasonable to pass along the user data in every alert that derives from `torrent_alert`, since the space is allocated in there already anyway.

Passing the user data pointer along with alerts adds to the memory pressure of the alert queues, it's not clear to me that it has to be passed in in that channel. You say it's to avoid an extra round-trip, but since libtorrent doesn't actually use or look at the user-data, it doesn't have to be synchronized with the libtorrent thread. I think it would be sufficient for this data to be accessed via atomic operations by client threads, thus avoiding the round-trip. This would be similar to how one associates user data with an `HWND` using `GetWindowLongPtr()` and `SetWindowLongPtr()` and the slot `GWLP_USERDATA`.

As far as I can tell, there is no way to *set* the user data in this patch. I imagine that should be done via the `add_torrent_params` object. Would it be feasible to only be able to set the user data when adding it? It would create fewer questions around synchronization, and it could probably be entirely non-synchronized that way. If a client use the user data from multiple threads, it would be responsible for synchronization.

@aldenml Do I recall correctly that you were talking about a feature along these lines a long time ago as well? Do you have any thoughts or opinions?

ghbplayer:
> A change that alters the ABI cannot be made against RC_1_2.

Okay. It's in the 'userdata' branch on my repo. Do I have to re-create the PR to change the target branch?

> Using a void* does not seem appropriate as it invites type- and resource errors.

I planned to add it to ATP and found it already had a void* userdata member. I can see the advantage of `std::shared_ptr<void>` for memory management but I don't see how `boost::any` helps because you're still casting in the end.

> Would it be feasible to only be able to set the user data when adding it?

I think so. It's usually a pointer to a struct and the user has a lot of control over changing the contents of the struct.

-----------

In summary, target the master branch, switch to `std::shared_ptr<void>`, and keep it out of the alerts. Sound good?

arvidn:
> Okay. It's in the 'userdata' branch on my repo. Do I have to re-create the PR to change the target branch?

No, you should be able to change the target branch on the PR. You'd have to rebase your branch too.

> I planned to add it to ATP and found it already had a void* userdata member.

Right, that's been there for quite a while as a channel into the `torrent_plugin`
 constructor, since it's constructed by libtorrent.

> I can see the advantage of std::shared_ptr<void> for memory management but I don't see how boost::any helps because you're still casting in the end.

`shared_ptr<>` can also be problematic, since it will force you to hold the shared state in a `shared_ptr`. If you have an object with static storage duration, you can't just pass in a pointer to that.

The benefit of `std::any` is that it's type safe. I take it you consider the `dynamic_cast` implied in an `any_cast` to be unreasonably costly compared to a `static_cast` you would use with a raw pointer or the (also static) `pointer_cast` of a `shared_ptr<void>`.

The benefits I see though, being the counter-weight to the more expensive cast, is that you're more flexible in what you can store. You can store just a raw pointer in there (but a pointer to the correct type, not type-punned). It doesn't impose any memory management, but you can also store a `shared_ptr<>` to your specific type if you'd like.

> In summary, target the master branch, switch to std::shared_ptr<void>, and keep it out of the alerts. Sound good?

Yes, except I'm not fully sold on the `shared_ptr`.

ghbplayer:
I just wrote the `shared_ptr` version and I'm testing it locally. It works well and it's a much smaller change. I tried to make it `unique_ptr` but it's too hard a change as ATP is explicitly copied in at least one place. If you're not sold on memory management (and I'm not either) then void* is still on the table.

I'm not very familiar with boost::any, but last I heard dynamic_cast uses string comparisons from RTTI. I'm not sure how expensive that is.

I imagine clients would only have exactly one data type. One cast when creating a torrent, one cast when polling messages. Not much logic, easy to get right and any problems will be obvious. Paying a runtime cost to add a safety layer on top of that sounds dubious.

It sooks like this PR closed automatically when I deleted the branch. I'll open another PR when I'm satisfied with my changes. Let me know your thoughts on void* though.

arvidn:
it's fine to use `void*`, I can add a light-weight type-check on top of that

aldenml:
@arvidn I remember there was some similar discussion, but nothing else :(

PR #4193, opened 2019-12-19
docs improvement

arvidn:


docs/makefile
```diff
@@ -18,6 +18,8 @@ REFERENCE_TARGETS = \
 	upgrade_to_1.2-ref \
 	reference \
 	reference-Core \
+	reference-DHT \
+	reference-Session \
```
arvidn: these files must be added to the tarball too

PR #4195, opened 2019-12-20
add clientdata to torrent

ghbplayer:
"add clientdata to torrent to skip searching for client data associated with torrent on every alert"

LT clients have data associated with torrents. Without that data, the alert system would be nearly worthless. Clients can associate their data with LT torrents via lt::torrent_handle, but the reverse association requires searching by comparing pointers or hashes. Efficient systems such as Windows Overlapped I/O and Linux epoll will accept a user pointer and provide that user pointer on events to help clients find their associated data.

With this change, LT alerts have a similar user data pointer so that every alert doesn't result in the client performing a search for associated data. This is nearly free in most cases and thus the client doesn't have to make a synchronous call to retrieve the pointer from the main LT thread.

There was a [previous version](https://github.com/arvidn/libtorrent/pull/4188) of this PR. This version doesn't add clientdata to any alerts except the remote torrent alert, and it's pulled into the add torrent alert because it's in add_torrent_params.

The clientdata type is void*. This isn't expected to cause much grief because it's 0 casts to set the ATP clientdata parameter and 1 cast when receiving an alert. It should always be the same cast, probably done by a single function. There's no lifetime management (smart pointer) involved, so the client probably needs to take some action on remove and add failure alerts. It's entirely optional.

Since only two alerts have grown, this has minimal impact on alert memory pressure. However, the clientdata parameter on remove is largely redundant with the 20 byte hash parameter. They serve the same purpose, perhaps the hash parameter could go.


arvidn:
there has to be a way to get the clientdata pointer from the `torrent_handle` too. Something like:

```
void* torrent_handle::clientdata() const
{
   return t ? t->get_clientdata() : nullptr;
}
```

include/libtorrent/add_torrent_params.hpp
```diff
@@ -179,6 +179,9 @@ TORRENT_VERSION_NAMESPACE_2
 		// (see torrent_handle::add_extension()).
 		void* userdata = nullptr;
 
+		// The ``clientdata`` parameter is optional and will be stored in the torrent.
```
arvidn: this comment should probably mention how to get it back. e.g. `torrent_handle::clientdata()`, and maybe that it's included in `torrent_removed_alert` too

include/libtorrent/alert_types.hpp
```diff
@@ -274,12 +274,14 @@ TORRENT_VERSION_NAMESPACE_2
...
 		TORRENT_UNEXPORT torrent_removed_alert(aux::stack_allocator& alloc
-			, torrent_handle const& h, info_hash_t const& ih);
+			, torrent_handle const& h, info_hash_t const& ih, void* clientdata);
 
 		TORRENT_DEFINE_ALERT_PRIO(torrent_removed_alert, 4, alert_priority::critical)
 		static constexpr alert_category_t static_category = alert::status_notification;
 		std::string message() const override;
+
 		info_hash_t info_hash;
+		void* clientdata;	// largely redundant with info_hash
```
arvidn: please put the comment above the `clientdata` field, and make it actually explain what it is and where it comes from. This is used to generate the documentation.

ghbplayer:
Good feedback, I've made the changes. I'm trying to understand why the travis CI mac build failed up above. Do you know if that's new or something I've caused?

ghbplayer:
Oh look, this time travis passed.

arvidn:
@aldenml does this seem reasonable to you too? Or perhaps more specifically with this patch: https://github.com/ghbplayer/libtorrent/pull/1

ghbplayer:
What do you think about putting the type check in an assert?

arvidn:
at first I had an assert in addition, but it makes it a lot harder to unit test.
The cost of the additional test in production is most likely negligible in the context. Even if it ends up in an inner loop, a condition that's always true is extremely cheap (with branch prediction).

arvidn:
could you please mention this new feature in the Changelog? (just a one-liner, just below the `release 2.0`). and also rebase on top of master (apparently there are some merge conflicts right now) and squash commits that belong together.

arvidn:
I'm sorry, I must have merged something that conflicts with this again :(

arvidn:
ok, I tried to use githubs "resolve conflicts", but that was probably a mistake. It should be rebased and squashed before landing in `master`

aldenml:
@arvidn I'm starting to remember the conversation, and yes, this looks good to me. Now, if I remember correctly, I was asking about having this "userdata" added to all torrent alerts, that way I could have avoided this filter https://github.com/frostwire/frostwire/blob/master/common/src/main/java/com/frostwire/bittorrent/BTDownload.java#L704

arvidn:
@aldenml what would you have done instead? I'm guessing you would have filtered out alerts belonging to that torrent handle some other way, by looking at your userdata object. It's not obvious to me how that code could be simplified with userdata associated with the alerts.

With this patch, where you can ask torrent_handle for the userdata, would it simplify that code?

aldenml:
I was thinking of pining the UI java object pointer and pass that as the "userdata", then reacting to my thread listening to events by just forwarding the alert data to the UI thread or something like that, and the goal was removing O(n) lookups. In any case, I wasn't dealing with more than a thousand torrents, so it was just a theoretical exercise.

arvidn:
@ghbplayer do you want me to squash and rebase this?

arvidn:
cleaned up here: https://github.com/arvidn/libtorrent/pull/4207

ghbplayer:
Thank you Arvid

PR #4200, opened 2019-12-21
fix typo in file_open_mode conversion

arvidn:


src/file_view_pool.cpp
```diff
@@ -143,7 +143,7 @@ namespace {
 	file_open_mode_t to_file_open_mode(open_mode_t const mode)
 	{
 		return ((mode & open_mode::write)
-				? file_open_mode::read_write : file_open_mode::read_write)
+				? file_open_mode::read_write : file_open_mode::read_only)
```
aldenml: I wonder if there should be a test that fails because of the typo

arvidn: yeah, that would be reasonable

PR #4202, opened 2019-12-21
add new socks5_alert to trouble shoot SOCKS5 proxies

arvidn:


arvidn:
I should expose the new alert to the python binding as well

sledgehammer999:
Don't forget to add the new alert in `fwd.hpp` too.
In any case, I have created a test build and a new issue on qbt bug tracker to track all relevant progress. If you like, follow user reports here: https://github.com/qbittorrent/qBittorrent/issues/11735

PR #4207, opened 2019-12-27
associate arbitrary client data with torrent_handles

arvidn:


arvidn:
@ghbplayer does this look good?

include/libtorrent/torrent.hpp
```diff
@@ -1179,6 +1179,8 @@ namespace libtorrent {
 		}
 		void add_suggest_piece(piece_index_t index);
 
+		client_data_t get_clientdata() const { return m_clientdata; }
```
aldenml: what about `get_client_data` vs `get_clientdata`, personal preference?

arvidn: well, not really. I suppose it should be consistent. and I think it is now. the type is called `client_data` and the variables are all called `clientdata`.
maybe `client_data` is nicer. Maybe there's a better name still. afaict, the only reason this isn't called `userdata` (which is a well established term) is because there is already an (inferior) userdata field in `add_torrent_params`, that most likely nobody uses.
An alternative name would be `cookie`, which is also well understood I think.

ssiloti: > there is already an (inferior) userdata field in add_torrent_params, that most likely nobody uses.

uTorrent Web uses the existing `userdata` field to track what UI actions should be taken when an `add_torrent_alert` is received. E.g. whether a new browser tab should be launched.

arvidn: ah yeah, I guess it is passed all the way back to the `add_torrent_alert`. I was thinking that it was only used for plugins

arvidn: actually, the `client_data_t` type is *almost* API compatible with a `void*`. If it would have an explicit template conversion operator, (instead of `get()`), the only use case (I can think of) it wouldn't support is being passed along as or assigned to a `void*`. i.e. only the use case where someone would `static_cast<>` it directly to the actual type would be allowed.

Maybe that's close enough to update the existing `userdata` to use the slightly safer type.

arvidn:
something like these last two commits

include/libtorrent/client_data.hpp
```diff
@@ -0,0 +1,96 @@
...
+	// identical to the type of the pointer assigned earlier, including
+	// cv-qualifiers.
+	template <typename T>
+	T* get() const
+	{
+		if (m_type_ptr != type<T>()) return nullptr;
+		return static_cast<T*>(m_client_ptr);
+	}
+	template <typename T>
+	operator T() const
```
arvidn: oops. I forgot to make this `explicit`

arvidn:
@ssiloti @aldenml @ghbplayer 
This updated version actually replaces the `void* userdata` in `add_torrent_params` with `client_data_t`, as well as what's passed to plugins. It also adds the new functionality from @ghbplayer's original patch where the userdata is stored in torrents and can (cheaply) be retrieved by clients from a `torrent_handle`.

The main *risk* with this patch is that it's not fully backwards compatible with the `void*` version. Specifically, it prevents void pointers to be assigned to it, as well as extracted from it. So, an existing user may be storing the userdata as a void pointer themselves, which this patch would break. However, I think the idiomatic and best-practice use of a `void*` userdata, of assigning to it from a concrete pointer type, and immediately when using, casting it to that same concrete pointer type, *is* supported.

Do you think this API change is small enough to warrant avoiding another userdata field, with a different name? (I'm leaning towards: yes)

include/libtorrent/client_data.hpp
```diff
@@ -0,0 +1,105 @@
...
+	}
+	template <typename T>
+	explicit operator T() const
+	{
+		static_assert(std::is_pointer<T>::value, "client_data_t can only be cast to a pointer");
+		if (m_type_ptr != type<typename std::remove_pointer<T>::type>()) return nullptr;
+		return static_cast<T>(m_client_ptr);
+	}
+
+#if TORRENT_ABI_VERSION <= 2
```
arvidn: I meant to turn this the other way around. To deprecate type-unsafe operations, but to allow it in older versions

ssiloti:
I'm fine with dropping support for void pointers.

PR #4210, opened 2019-12-29
simplify dump_torrent example. add new example

arvidn:
to convert torrent to magnet link

### DIFF
--- a/examples/dump_torrent.cpp
+++ b/examples/dump_torrent.cpp
@@ -44,19 +44,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/magnet_uri.hpp"
 #include "libtorrent/span.hpp"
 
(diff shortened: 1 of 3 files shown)


examples/dump_torrent.cpp
```diff
@@ -117,28 +104,22 @@ int main(int argc, char const* argv[]) try
...
+	{
+		std::printf("nodes:\n");
+		for (auto const& i : t.nodes())
+			std::printf("%s: %d\n", i.first.c_str(), i.second);
+	}
 
-	puts("trackers:\n");
-	for (auto const& i : t.trackers())
-		std::printf("%2d: %s\n", i.tier, i.url.c_str());
+	if (t.trackers().empty())
```
FranciscoPombal: @arvidn 
Missing negation?

arvidn: heh, yes...

PR #4213, opened 2019-12-31
suppress announces with incompatible address family

arvidn:
suppress future attempts to announce to a tracker from an interface with an incompatible address family


arvidn:
oh, `announce_endpoint` is public, this would break ABI

PR #4220, opened 2020-01-02
WIP: don't start DHT on loopback

arvidn:
if a loopback IP address, or the loopback device, is explicitly specified in `listen_interfaces`, don't start a DHT node on it.

### DIFF
--- a/src/kademlia/dht_tracker.cpp
+++ b/src/kademlia/dht_tracker.cpp
@@ -119,8 +119,13 @@ namespace libtorrent { namespace dht {
 		if (s.is_ssl()) return;
 
 		address const local_address = s.get_local_endpoint().address();
+
+		// don't start DHT nodes on loopback address
+		if (local_address.is_loopback()) return;
+
 		// don't try to start dht nodes on non-global IPv6 addresses
-		// with IPv4 the interface might be behind NAT so we can't skip them based on the scope of the local address
+		// with IPv4 the interface might be behind NAT so we can't skip them
+		// based on the scope of the local address
 		// and we might not have the external address yet
 		if (local_address.is_v6() && is_local(local_address))
 			return;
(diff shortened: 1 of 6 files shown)


src/kademlia/dht_tracker.cpp
```diff
@@ -119,8 +119,13 @@ namespace libtorrent { namespace dht {
 		if (s.is_ssl()) return;
 
 		address const local_address = s.get_local_endpoint().address();
+
+		// don't start DHT nodes on loopback address
+		if (local_address.is_loopback()) return;
```
arvidn: this is the material change

test/test_direct_dht.cpp
```diff
@@ -99,9 +99,9 @@ TORRENT_TEST(direct_dht_request)
 	sp.set_bool(settings_pack::enable_upnp, false);
 	sp.set_str(settings_pack::dht_bootstrap_nodes, "");
 	sp.set_int(settings_pack::max_retry_port_bind, 800);
-	sp.set_str(settings_pack::listen_interfaces, "127.0.0.1:42434");
+	sp.set_str(settings_pack::listen_interfaces, "0.0.0.0:42434");
 	lt::session responder(sp, {});
-	sp.set_str(settings_pack::listen_interfaces, "127.0.0.1:45434");
+	sp.set_str(settings_pack::listen_interfaces, "0.0.0.0:45434");
```
ssiloti: Hrm, if we start expanding `0.0.0.0` these tests will break again.

arvidn: you're right. what is the right thing to do? maybe if one explicitly binds to loopback, it's OK to start a DHT node on it, as long as it stays there, and doesn't try to reach out to non-loopback addresses.

arvidn: on the other hand, if one binds to `0.0.0.0`, I think it's expected to be able to reach "the one IPv4" DHT node via `127.0.0.1`

arvidn:
maybe this doesn't make sense after all.

PR #4223, opened 2020-01-04
WIP: only expand :: to IP on the default route device

arvidn:


arvidn:
If this makes sense, the new `find_default_device()` should have tests as well

arvidn:
I don't really think this makes a difference. I will test it some more, but I don't think the non-tunnel interface works while a tunnel is in effect.

PR #4225, opened 2020-01-04
fix null termination of description and friendly_name in ip_interface on linux

arvidn:


ssiloti:
Maybe these fields should be default initialized to all zeros. Then all this code would be unnecessary.

PR #4226, opened 2020-01-05
listen_socket_t in async_accept()

arvidn:
pass the whole listen_socket_t object to async_accept() and on_accept_connection(), instead of just the acceptor socket


arvidn:
this is probably a good idea, but not urgent. maybe it should target master

PR #4233, opened 2020-01-07
minor fix to socks5 simulation

arvidn:


simulation/test_socks5.cpp
```diff
@@ -279,5 +277,9 @@ TORRENT_TEST(socks5_udp_retry)
 	sim.run();
 
 	// number of UDP ASSOCIATE commands invoked on the socks proxy
-	TEST_EQUAL(socks5.cmd_counts()[2], 12);
+	// We run for 60 seconds. The sokcks5 retry interval is expected to be 5
+	// seconds, meaning there should have been 12 connection attempts since we
+	// have an outgoing_udp_socket as well, it will also attempt to establish a
+	// SOCKS5 UDP tunnel, so it will bring it to 24 attempts
+	TEST_EQUAL(socks5.cmd_counts()[2], 24);
```
arvidn: the reason this was just 12 previously (which is what one would expect) was that binding the listen socket to "50.50.50.50" failed (because that's not the address of the bittorrent client node, it's "50.0.0.1"). The listen socket failing, left only the "outgoing socket" to issue the SOCKS5 tunnel. Fixing the listen socket caused both to attempt the SOCKS5 tunnel.

PR #4248, opened 2020-01-11
Backport dishonest peer antileech choking to RC_1_1

alanswanson:
Backport the choking of dishonest peers in the anti-leech seeding algorithm #3833 to libtorrent 1.1 for any potential future release, or for others who just want it themselves. Still using Deluge 1.3 pending some Deluge 2.0 issues.

Per https://github.com/arvidn/libtorrent/pull/3833#issuecomment-538783555 patches welcomed.

### DIFF
--- a/src/choker.cpp
+++ b/src/choker.cpp
@@ -38,6 +38,14 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include <boost/bind.hpp>
 
+// Workaround due to MSVC's absence of std::abs (int64_t) before C++11
+template <typename T>
+inline T
+prec11_abs (T x)
+{
+	return (x >= 0 ? x : -x);
+}
+
 namespace libtorrent {
 
 namespace {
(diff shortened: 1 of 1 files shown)


arvidn:
since travis doesn't work for RC_1_1, which platform did you try this on?

arvidn:
this is a problem: https://ci.appveyor.com/project/arvidn/libtorrent/builds/30048019/job/501oexp1w8e5nq2u#L4766

do you need `std::` qualification?

alanswanson:
Built on Linux with GCC 9.2. Maybe MS Visual Studio needs <cstdlib> specifically included? Grepping the code, it is commented as needed for std::abs in src/sliding_average.hpp. Will spin a new patch.

arvidn:
you may have to include `<cinttypes>` to get the overload of `abs()` that take a `std::int64_t`

arvidn:
oh, maybe the problem is that 1.1 is pre C++11

alanswanson:
Changed patch to workaround MSVC's absence of std::abs (int64_t) before C++11 with an inline template.

Issue #4216, opened 2020-01-01
expanding [::] to all IPv6 interfaces (multi-homed support)


arvidn:
libtorrent version (or branch): RC_1_2

Currently, if one has multiple network interfaces and listens on `[::]` (which is done by default). That address is expanded to all IPv6 addresses the machine has. One reason for having multiple IPv6 addresses is by running a VPN, where there's a `tun0` device with an IPv6 address assigned to it.

The current logic opens a listen socket for both the VPN interface *and* the underlying "direct" interface. I don't think this is intuitive behavior. It's explained in the code with this comment:

```
      // To comply with BEP 45 multi homed clients must run separate DHT nodes
      // on each interface they use to talk to the DHT. For IPv6 this is enforced
      // by prohibiting creating a listen socket on [::]. Instead the list of
      // interfaces is enumerated and sockets are created for each of them.
      // This is not enforced for 0.0.0.0 because multi homed IPv4 configurations
      // are much less common and the presence of NAT means that we cannot
      // automatically determine which interfaces should have DHT nodes started on
      // them.
```

This is why it's not done with the IPv4 counterpart, 0.0.0.0.

My first thought of making this more intuitive was to only expand it to the single *main* IPv6 interface. To find the main interface:

1. find the first entry in the routing table with `[::]` as the destination. remember the associated device name.
2. find the first IP address in the interfaces list associated with this device, and whose gateway is specified.

If no entry in the routing table can be found, ignore the `[::]` entry in the `listen_interfaces` setting. This same algorithm could be used to treat the unspecified IPv4 address the same as well.

The idea is that in the (presumably rare) case of having a machine actually hooked up to two or more internet connections, one would have to explicitly add both to the `listen_interfaces` setting.

After having implemented this behavior though, it occurred to me that it really seems that the behavior I'm looking for is to preserve the unspecified listen addresses, and let the OS figure it out. The question in that scenario is what to do about the DHT nodes.

Perhaps the DHT node could be made to allow the unspecified IPv6 address, but use the algorithm above to find its concrete IP.

In either case, I should also add a section in the documentation talking about multi-homed support and explain this logic in detail, to make it clear how it works.

@ssiloti I'm hoping you have some thoughts or opinions on this. I'm not sure what the best solution is, but I really hope it involves reducing complexity.

arvidn:
actually, the comment about prohibiting DHT nodes on `[::]` seem outdated. I don't see anything prohibiting this.

[...]

PR #4249, opened 2020-01-13
improve multi-homing support

arvidn:
This is a fairly significant change to how listen sockets and outgoing sockets are assigned to network interfaces and source IP addresses.

The commits all represent (mostly) independent changes all having the effect that:

* Trackers are only announced from a source IP address where it makes sense. e.g. only if the tracker is running on loopback does it make sense to announce to it from a socket bound to loopback.
* SOCKS5 UDP tunnels are only set up for listen sockets where it makes sense (same logic as for trackers)
* the UDP sockets dedicated for outgoing packets have been removed and folded into the listen sockets (for UDP). This simplifies things a bit and possibly improves reachability throught NATs.
* when listening on 0.0.0.0, multiple IPv4 sockets will be opened, one for each local IPv4 address, however, they will all be restricted to be used for their respective subnet. Like the example with the tracker above, a socket listening to loopback, will only be used for destinations over loopback. IPv6 addresses were already treated this way, but not IPv4.

I think this will improve SOCKS5 support, mainly in that only a single UDP tunnel will be opened (in the common case). But if you have both IPv4 and IPv6 connectivity for instance, and the SOCKS5 tunnel resolves to both IPv4 and IPv6, two tunnels will be opened. This is to take full advantage of all network paths.

There's some more discussion on this topic in [this ticket](https://github.com/arvidn/libtorrent/issues/4216#issuecomment-572535958).


arvidn:
Improvements I intend to build on top of this:

* simplify UPnP by also restricting it only to the subnets it's started for (as is done for NAT-PMP/PCP already)
* simplify local service discovery by restricting it to the subnet it's started for

arvidn:
@ssiloti Would you mind reviewing this patch?

arvidn:
I just rebased on top of `RC_1_2`

src/session_impl.cpp
```diff
@@ -283,6 +278,66 @@ namespace aux {
...
+
+			// also record whether the device has a gateway associated with it
+			// (which indicates it can be used to reach the internet)
+			bool const has_gateway
+				= std::find_if(routes.begin(), routes.end(), [&](ip_route const& r)
+				{
+					return r.destination.is_unspecified()
+						&& r.destination.is_v4() == iface->interface_address.is_v4()
+						&& strcmp(r.name, iface->name) == 0;
+				}) != routes.end();
```
ssiloti: `has_gateway`should not be set to true for listen endpoints on local IPv6 addresses

src/http_tracker_connection.cpp
```diff
@@ -272,6 +272,20 @@ namespace libtorrent {
...
+		auto const ls = bind_socket();
+		endpoints.erase(std::remove_if(endpoints.begin(), endpoints.end()
+			, [&](tcp::endpoint const& ep) { return !ls.can_route(ep.address()); })
+			, endpoints.end());
+
+		if (endpoints.empty())
+		{
+			fail(error_code(boost::system::errc::host_unreachable, system_category()));
+			return;
+		}
```
ssiloti: There's similar code in `http_connection::on_resolve` which should probably be folded into this.

src/session_impl.cpp
```diff
@@ -283,6 +278,66 @@ namespace aux {
...
+		}
+	}
+
+	bool listen_socket_t::can_route(address const& addr) const
+	{
+		if (is_v4(local_endpoint) != addr.is_v4()) return false;
+
+		if (local_endpoint.address().is_v6()
+			&& local_endpoint.address().to_v6().scope_id() != addr.to_v6().scope_id())
+			return false;
```
ssiloti: This scope_id check should go in `match_addr_mask` since it is needed other places that function is called such as `session_impl::bind_outgoing_socket`

arvidn: good point. I think this check still needs to be here though, right? the call to `match_addr_mask()` is kind of the "last resort" in this function, but the scope ID check must happen first. At least that's my understanding

src/session_impl.cpp
```diff
@@ -283,6 +278,72 @@ namespace aux {
...
+
+			ep.netmask = iface->netmask;
+
+			bool const v4 = iface->interface_address.is_v4();
+
+			// also record whether the device has a gateway associated with it
+			// (which indicates it can be used to reach the internet)
+			// only gateways inside the interface's network count
+			bool const has_gateway =
+				(v4 || !iface->interface_address.to_v6().is_link_local())
```
ssiloti: Use `is_local` instead of `is_link_local`, link-local is just one of several local address ranges.

src/session_impl.cpp
ssiloti: Yes you're right, this check is still needed

src/session_impl.cpp
arvidn: the more I think about this, it's not obvious to me that this actually matters. with the check that the gateway is inside the network specified in the `ip_interface`, I don't think it will be mis-associated with a local interface. It would only happen if the gateway's IP address also falls within the local IP range. like, `127.1.1.1` or something. Which would be weird, but would it necessarily be wrong?
`is_local()` is pretty broad and covers IPv4 as well, and I think a lot of the cases it covers wouldn't be necessary.

I'm not very familiar with the semantics of scope IDs in IPv6. But is it possible to have an IP address that looks global but is local just by the scope ID?

arvidn: actually, `is_local()` checks `is_site_local()`. I'm not an expert on this obviously, but I would imagine you can have a local "site" that consists of multiple subnets, where you actually need a gateway between them.

In the current use of `is_local()`this makes sense (I think), since it's used to determine whether to apply rate limits or not (at least it did historically), I suppose this is mostly based on IP filters now.

In fact, I get the impression that there shouldn't be a special case here. I think it's possible, and perhaps even likely, that some sites run a private IPv6 network, with all nodes in the `fc00::/7` block (i.e. local addresses), but still have this private network connected to the internet, via NAT. In that case, you would have an interface with a local IPv6 address, and still have a gateway to the internet. As long as that gateway also live within the same private network, and is reachable from this interface, that seems like a legitimate set up.

src/enum_net.cpp
```diff
@@ -764,28 +767,27 @@ int _System __libsocket_sysctl(int* mib, u_int namelen, void *oldp, size_t *oldl
...
 
+	boost::optional<address> get_gateway(ip_interface const& iface, span<ip_route const> routes)
+	{
+		auto const it = std::find_if(routes.begin(), routes.end()
+			, [&](ip_route const& r) -> bool
+			{
+				return r.destination.is_unspecified()
+					&& r.destination.is_v4() == iface.interface_address.is_v4()
+					&& !r.gateway.is_unspecified()
+					&& match_addr_mask(r.gateway, iface.interface_address, iface.netmask)
```
ssiloti: This will not work for IPv6. The gateway is typically specified by its link-local address, so this will cause only the link-local address on an interface to have `has_gateway` set which is the exact opposite of what it should be.

src/session_impl.cpp
ssiloti: What you describe is NAT66 which is rare and considered bad practice. Even if we wanted to support it, which I oppose, we'd have to actively probe each local address to see if it can reach outside the local network before flagging it with `has_gateway`. Otherwise in the common case we'd constantly be trying to send global traffic on local addresses which cannot reach the internet.

ssiloti: Site-local addresses are specified as non-routable, so they should not be used across subnets. Anyways they were deprecated a long time ago (RFC 3879) so I doubt they're actually used today.

src/enum_net.cpp
arvidn: what does it mean for an address to be "link local" in IPv6? My mental model is based on IPv4, where I would expect all addresses reachable by a switch on the local network would all be considered link local. Just like the IP address associated with the network would be a "local" address (not local as in loopback, but reserved for a private network). The gateway would be on that same private network too. In IPv6, if you have a local network does every node have both a local address and a global address?

I don't have IPv6 connectivity so I don't really have an example of what it looks like. what would be the correct way of ascribing a gateway route to a subnet? where "subnet" means address on the interface + netmask.

arvidn:
I realized I had one machine with IPv6 connectivity. running `enum_if` on that machine give me this:

```
$ ./enum_if
Default gateway: 185.157.221.1
0.0.0.0           0.0.0.0           185.157.221.1                      1500   ens1
185.157.221.0     255.255.255.0     -                                  1500   ens1
185.157.221.1     255.255.255.255   -                                  1500   ens1
127.0.0.0         255.255.255.255   -                                  65536  lo
127.0.0.0         255.0.0.0         -                                  65536  lo
127.0.0.1         255.255.255.255   -                                  65536  lo
127.255.255.255   255.255.255.255   -                                  65536  lo
185.157.221.0     255.255.255.255   -                                  1500   ens1
185.157.221.247   255.255.255.255   -                                  1500   ens1
185.157.221.255   255.255.255.255   -                                  1500   ens1
::1               ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff-                                  65536  lo
fe80::            ffff:ffff:ffff:ffff::-                                  1500   ens1
::                ::                fe80::5:73ff:fea0:a%ens1           1500   ens1
::1               ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff-                                  65536  lo
2a02:752:0:18::128ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff-                                  1500   ens1
fe80::1005:57ff:fe51:1201ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff-                                  1500   ens1
ff00::            ff00::            -                                  1500   ens1
address                           netmask                                      name                flags               gateway                           description
127.0.0.1                         255.0.0.0                                    lo                  local loopback            -
185.157.221.247                   255.255.255.0                                ens1                                    185.157.221.1
::1                               ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff      lo                  local loopback            -
2a02:752:0:18::128                ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff      ens1                                    -
fe80::1005:57ff:fe51:1201%ens1    ffff:ffff:ffff:ffff::                        ens1                local                     fe80::5:73ff:fea0:a%ens1
```

Which looks like it's ascribing the gateway to the correct interface.

arvidn:
do you have an example setup where it gets this wrong?

ssiloti:
Link local addresses are a special range in IPv6: `fe80::/10`. Typically yes, every node will have a link local and one-or-more global addresses. Link local addresses are non-rotatable so in that sense they are similar to RFC 1918 addresses. In the example output you show the gateway should not be associated with `fe80::1005:57ff:fe51:1201` because that address cannot be used to reach other networks. Trying to send packets to the gateway using that address will probably just elicit an error from the gateway, or be silently dropped. The `2a02:752:0:18::128` address is the one which should have the gateway associated with it since that is a global address which can be used to reach the internet.

arvidn:
I see. I imagine the `local` flag could also be used to determine an interface should not be associated with a router then. At least for IPv6.

So, if one has multiple routes to the internet, both over IPv6 networks. Is there a way to determine which gateway is for which network? (if they are on the same NIC say). Or is the network interface name the only distinguishing feature?

ssiloti:
With SLAAC the Router Advertisement message specifies both the gateway address and the network prefix it is valid for, so in theory multiple gateways can be supported. The problem is getting this information from the OS. If source-specific routing is configured I *think* this would manifest as an `RTA_SRC` attribute in the routing table which could be matched to the interface address. I don't have a multi-homed network to test this theory on though.

It's not immediately clear how Windows would report this, it may not even support source-specific routing.

ssiloti:
For our purposes we can probably assume that any device with at least one gateway has a valid gateway for every global address assigned to it.

arvidn:
yeah. I was thinking about adding support for the source-address hint linux has. That would probably be a slight improvement.

arvidn:
With this version of the patch, I get:

```
$ ./enum_if
Default gateway: 185.157.221.1
0.0.0.0           0.0.0.0           185.157.221.1                      1500   ens1
185.157.221.0     255.255.255.0     -                                  1500   ens1
185.157.221.1     255.255.255.255   -                                  1500   ens1
127.0.0.0         255.255.255.255   -                                  65536  lo
127.0.0.0         255.0.0.0         -                                  65536  lo
127.0.0.1         255.255.255.255   -                                  65536  lo
127.255.255.255   255.255.255.255   -                                  65536  lo
185.157.221.0     255.255.255.255   -                                  1500   ens1
185.157.221.247   255.255.255.255   -                                  1500   ens1
185.157.221.255   255.255.255.255   -                                  1500   ens1
::1               ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff-                                  65536  lo
fe80::            ffff:ffff:ffff:ffff::-                                  1500   ens1
::                ::                fe80::5:73ff:fea0:a%ens1           1500   ens1
::1               ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff-                                  65536  lo
2a02:752:0:18::128ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff-                                  1500   ens1
fe80::1005:57ff:fe51:1201ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff-                                  1500   ens1
ff00::            ff00::            -                                  1500   ens1
address                           netmask                                      name                flags               gateway                           description
127.0.0.1                         255.0.0.0                                    lo                  local loopback            -
185.157.221.247                   255.255.255.0                                ens1                                    185.157.221.1
::1                               ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff      lo                  local loopback            -
2a02:752:0:18::128                ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff      ens1                                    fe80::5:73ff:fea0:a%ens1
fe80::1005:57ff:fe51:1201%ens1    ffff:ffff:ffff:ffff::                        ens1                local                     -
```

PR #4251, opened 2020-01-14
deprecate (and remove) upnp_ignore_nonrouters setting

arvidn:


src/upnp.cpp
```diff
@@ -441,50 +436,6 @@ void upnp::on_reply(udp::endpoint const& from, span<char const> buffer)
 		return;
 	}
 
-	bool non_router = false;
-	if (m_ignore_non_routers)
-	{
-		auto const routes = enum_routes(m_io_service, ec);
```
aldenml: do you remember if `enum_routes` is still used in other parts of the code?

arvidn: yes, it's definitely used. In fact, I intend to make it mandatory for building libtorrent. See https://github.com/arvidn/libtorrent/pull/4249

arvidn:
The reason I'm removing this is that I'm working on a patch to do this much better, where the UPnP logic is tied to a specific subnet, and knows exactly which IP addresses are part of it.

src/upnp.cpp
aldenml: 👍

aldenml:
👍

PR #4254, opened 2020-01-15
restrict DHT nodes to the network they are associated with (multi-homed)

arvidn:


arvidn:
@ssiloti can you think if any issues of filtering packets to this degree? I imagine there's some rationale for not having done so previously

ssiloti:
Responses aren't filtered to allow for cross-address-family bootstrapping as allowed by BEP 32.

arvidn:
I see. So maybe I should just filter the “external up” call then

PR #4255, opened 2020-01-16
populate ip_interface.netmask on Windows

ssiloti:


src/enum_net.cpp
```diff
@@ -699,10 +699,30 @@ int _System __libsocket_sysctl(int* mib, u_int namelen, void *oldp, size_t *oldl
...
 				{
-					if (!valid_addr_family(unicast->Address.lpSockaddr->sa_family))
+					auto const family = unicast->Address.lpSockaddr->sa_family;
+					if (!valid_addr_family(family))
 						continue;
 					r.preferred = unicast->DadState == IpDadStatePreferred;
 					r.interface_address = sockaddr_to_address(unicast->Address.lpSockaddr);
+					// OnLinkPrefixLength is only present on Vista and newer
+					int const max_prefix_len = family == AF_INET ? 32 : 128;
+					if (unicast->Length >= 64 && unicast->OnLinkPrefixLength <= max_prefix_len)
```
arvidn: I take it the `Length >= 64` is a check to make sure we're on vista or later. Wouldn't it be a bit clearer to say: `unicast->Length >= sizeof(IP_ADAPTER_UNICAST_ADDRESS)`? (I assume that's effectively what it does now)

arvidn: oh, but I guess if you compile for a newer version of windows, with a larger structure, you still want to support running on vista..

src/enum_net.cpp
```diff
@@ -699,10 +699,30 @@ int _System __libsocket_sysctl(int* mib, u_int namelen, void *oldp, size_t *oldl
...
+					}
+
+					if (family == AF_INET6 && (unicast->PrefixOrigin == IpPrefixOriginDhcp
+						|| unicast->SuffixOrigin == IpSuffixOriginRandom))
+					{
+						// DHCPv6 does not specify a subnet mask (it should be taken from the RA)
+						// but apparently MS didn't get the memo and incorrectly reports a
+						// prefix length of 128 for DHCPv6 assigned addresses
+						// 128 is also reported for privacy addresses despite claiming to
+						// have gotten the prifix length from the RA *shrug*
```
aldenml: in case you need to change this PR, there is a typo in `prifix`

src/enum_net.cpp
arvidn: in [qt](https://code.woboq.org/qt5/qtbase/src/network/kernel/qnetworkinterface_win.cpp.html#218) they check against `offsetof` instead of a hard coded struct size.

```
if (unicast->Length > offsetof(IP_ADAPTER_UNICAST_ADDRESS, OnLinkPrefixLength) ...
```

I can't find any reasons (on the internet) to expect windows to return entries with invalid `OnLinkPrefixLength`. So it's probably fine to always add the `ip_interface`

PR #4256, opened 2020-01-17
fix clearing of netmask if it's not known, on windows

arvidn:


ssiloti:
One other change I meant to make is to add a check that the prefix length really is 128 before applying the workaround. That way if Windows is fixed the workaround won't be applied unnecessarily.

arvidn:
ah, so maybe the reordering of these blocks won't make sense then

PR #4258, opened 2020-01-18
fix test_lsd to pass reliably

arvidn:


test/test_lsd.cpp
```diff
@@ -53,8 +53,12 @@ void test_lsd()
 	session_proxy p2;
 
 	settings_pack pack;
-	pack.set_bool(settings_pack::allow_multiple_connections_per_ip, true);
```
arvidn: removing this line was the fix

PR #4273, opened 2020-01-20
set thread memory priority to below-normal for disk I/O threads.

arvidn:
The memory priority of a thread is assigned to pages it caused the allocation of. Presumably this includes pages allocated as part of a fault against a memory mapped file.

Presumably this means the disk cache will be first in line to be reclaimed if the system is running low on memory.

### DIFF
--- a/src/mmap_disk_io.cpp
+++ b/src/mmap_disk_io.cpp
@@ -64,6 +64,11 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/aux_/file_view_pool.hpp"
 #include "libtorrent/aux_/scope_end.hpp"
 
+#ifdef _WIN32
+#include "libtorrent/aux_/windows.hpp"
+#include "libtorrent/aux_/win_util.hpp"
+#endif
+
 #include <functional>
 #include <condition_variable>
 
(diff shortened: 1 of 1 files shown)


src/mmap_disk_io.cpp
```diff
@@ -1374,11 +1378,16 @@ TORRENT_EXPORT std::unique_ptr<disk_interface> mmap_disk_io_constructor(
...
 	}
 
-	void mmap_disk_io::thread_fun(job_queue& queue
-		, disk_io_thread_pool& pool)
+	void mmap_disk_io::thread_fun(job_queue& queue, disk_io_thread_pool& pool)
 	{
 		std::thread::id const thread_id = std::this_thread::get_id();
 
+#ifdef _WIN32
+		MEMORY_PRIORITY_INFORMATION const info{MEMORY_PRIORITY_BELOW_NORMAL};
```
Chocobo1: This struct is only available on Windows 8 and later, so you might want to add some guard here.
https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/ns-processthreadsapi-memory_priority_information

FranciscoPombal: Windows 7 was recently EOL'd (yes, I realize this does not magically make it disappear). Not sure what libtorrent's support policy is, but I think this is worth mentioning.

arvidn: I should probably use the helpers I already have to load the shared library at run-time and detect whether this function exists or not. `get_library_procedure()` in `win_util.hpp`

PR #4275, opened 2020-01-20
limit the number of DHT nodes to run

arvidn:


ssiloti:
There's existing filtering code in `dht_tracker::new_socket` which is obsoleted by this PR.

PR #4276, opened 2020-01-21
cleanup and lint refactor in ip_filter code

aldenml:


include/libtorrent/ip_filter.hpp
```diff
@@ -41,21 +41,12 @@ POSSIBILITY OF SUCH DAMAGE.
...
-#include <iterator> // for next
 #include <limits>
 
 #include "libtorrent/address.hpp"
-#include "libtorrent/assert.hpp"
 
 namespace libtorrent {
 
-// hidden
-inline bool operator<=(address const& lhs
```
aldenml: hi @arvidn, how one can be sure this is not used (or is it?)

arvidn: I imagine if it builds it's not used.

arvidn:
is this related to this change? https://ci.appveyor.com/project/arvidn/libtorrent/builds/30246831/job/bf6hgy3n2jg4jgqm#L354

arvidn:
https://ci.appveyor.com/project/arvidn/libtorrent/builds/30246831/job/8o50o23j97i0f9oj#L211

arvidn:
`std::set<>` is not defined to be `noexcept`. You could make the `noexcept` you added conditional, something like this:

```
ip_filter(ip_filter&&) noexcept(std::is_nothrow_move_constructible<std::set<int>>::value)
```

The value type of the set is a private type, so you can't use that in the header, so `int` could work as a stand in.

aldenml:
(always good to learn something) I can see that in `clang` headers I have
```cpp
set(set&& __s)
        _NOEXCEPT_(is_nothrow_move_constructible<__base>::value)
```
do you think it makes sense to try to get `filter_impl::range` _nothrow_move_constructible_?

arvidn:
it's a prerequisite to make `ip_filter` and `port_filter` nothrow_move_constructible. Whether that's worth it? I'm not sure. In a way I think everything that can should be, but I think the real value from that property is when you stick the type in a container. libtorrent doesn't use `ip_filter` or `port_filter` in any container, so the practical benefit will probably not be that great.

not that currently `filter_impl` inherits its nothrow_move_construciability from its `std::set` member, which *is* nothrow move constructible on clang and GCC, but not on msvc.

The recent patch of mine where I moved the implementations into the cpp file caused the `ip_filter` to no longer be nothrow move constructible even on clang and GCC, since it's not known outside of the cpp file anymore.

aldenml:
OK, I will revert that

aldenml:
sure, done

PR #4279, opened 2020-01-22
Add vcpkg installation instructions

PhoebeHui:
Libtorrent is available as a port in VCPKG , documenting the install process here will help users get started by providing a single set of commands to build libtorrent, ready to be included in their projects.

VCPKG is a C++ library manager that simplifies installation for libtorrent and other project dependencies, we also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and here is what the port script looks like. We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/docs/building.rst
+++ b/docs/building.rst
@@ -483,7 +483,7 @@ If you enabled test in the configuration step, to run them, run::
 
 building with VCPKG
 
-You can download and install libtorrent  using the [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
+You can download and install libtorrent using the [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
 
     git clone https://github.com/Microsoft/vcpkg.git
     cd vcpkg
--- a/docs/building.rst
+++ b/docs/building.rst
@@ -481,6 +481,19 @@ If you enabled test in the configuration step, to run them, run::
 
 	ctest -j8
 
+building with VCPKG
+
+You can download and install libtorrent  using the [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
+    ./vcpkg install libtorrent
+
+The libtorrent port in vcpkg is kept up to date by Microsoft team members and community contributors.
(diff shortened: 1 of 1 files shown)


docs/building.rst
```diff
@@ -481,6 +481,19 @@ If you enabled test in the configuration step, to run them, run::
 
 	ctest -j8
 
+building with VCPKG
```
arvidn: this should probably be a heading, and the same level as the headings using `-`. Could you add a line of 19 `-` under this?

docs/building.rst
```diff
@@ -481,6 +481,19 @@ If you enabled test in the configuration step, to run them, run::
...
+
+You can download and install libtorrent using the [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
```
arvidn: if you add another  colon ":" here, the block below will be a fixed with, preformatted, block.

docs/building.rst
```diff
@@ -481,6 +481,19 @@ If you enabled test in the configuration step, to run them, run::
...
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
+    ./vcpkg install libtorrent
```
arvidn: this indented block should work as it is (with the colon I suggest above). But the convention I have in the documentation is to use a single tab character for indentation.

arvidn:
you have to add "vcpkg" to `docs/hunspell/libtorrent.dic`. See this CI failure: https://travis-ci.org/arvidn/libtorrent/jobs/640341720#L900

docs/building.rst
PhoebeHui: @arvidn, thank you for your review!

I have updated the format.

docs/building.rst
PhoebeHui: Updated.

docs/building.rst
PhoebeHui: Updated, thanks!

arvidn:
the spell check is still failing: https://travis-ci.org/arvidn/libtorrent/jobs/640736251#L898

arvidn:
replaced by https://github.com/arvidn/libtorrent/pull/4301

PhoebeHui:
@arvidn, thanks for your help!

arvidn:
no problem, thanks yourself!

PR #4280, opened 2020-01-22
remove broadcast_socket from main library

arvidn:
it's only used by `test_upnp` now


include/libtorrent/address.hpp
```diff
@@ -73,6 +73,29 @@ namespace libtorrent {
 	using boost::asio::ip::make_address_v4;
 	using boost::asio::ip::make_address_v6;
 #endif
+
+	TORRENT_EXTRA_EXPORT bool is_local(address const& a);
```
aldenml: what do you think about putting these inside `aux_/address.hpp` instead?

arvidn: yeah, that probably makes sense. If I'm adding a new header, I might as well call it something else too.

arvidn:
@aldenml does this look reasonable?

aldenml:
Sometimes I do wonder if using `aux::` all over the place (like `std::`) instead of using `using namespace aux` is the right approach, when done incrementally it goes unnoticed, but it's adding up.

arvidn:
I'm thinking that once all the internal code is actually *defined* in the `aux` namespace, a lot of it could be removed

PR #4286, opened 2020-01-25
fix infinite loop when announcing

ssiloti:
In torrent::announce_with_tracker there is a check after announcing all of
a tracker's endpoints to see if all required announces have already been sent
according to the tier settings.

```
	if (std::all_of(listen_socket_states.begin(), listen_socket_states.end()
		, [](announce_state const& s) { return s.done; }))
		break;
```

Commit ab07eceead added a check which can cause listen_socket_states to not be
populated if all endpoints for the first tracker are disabled. This causes the
early break out code above to trigger and prevent any announces from being sent.

The extra check is not in torrent::update_tracker_timer though, so if it thinks there
is at least one endpoint ready to announce it will immediately trigger another
on_tracker_announce, again, and again.

Simple fix is to move the offending check below the code which populates
listen_socket_states so the the latter will execute unconditionally.

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -2937,10 +2937,11 @@ bool is_downloading_state(int const st)
 
 			for (auto& aep : ae.endpoints)
 			{
-				// if we haven't sent an event=start to the tracker, there's no
-				// point in sending an event=stopped
-				if (!aep.enabled || (!aep.start_sent && req.event == tracker_request::stopped))
-					continue;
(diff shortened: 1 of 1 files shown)


arvidn:
thanks! Sometimes I think about rewriting the whole announce logic to more directly express the logic, and be bit more robust (this is not the first time there's been an infinite loop bug in this code). But there's quite a lot of behavior that would need to preserved, so it's not that easy.

PR #4290, opened 2020-01-29
lower TCP keepalive timeout for SOCKS5 UDP tunnel

arvidn:
to detect network failures and retry


arvidn:
I was hoping someone would test this patch. But I feel pretty confident it won't make things worse

PR #4296, opened 2020-01-31
don't use memory mapped files when building as 32 bit

arvidn:
since the address space is too small. Fall back to the plain posix disk IO


arvidn:
@ssiloti this seems reasonable, right? Can you think of a better solution?
I'm sort of thinking that a clever compiler may dead-code remove all of `mmap_disk_io` for 32 bit builds.

ssiloti:
I think it's an acceptable solution. The mmap code definitely will not be optimized away though because  `mmap_disk_io_constructor` is a public symbol.

Can the default value of `mmap-disk-io` be changed based on the value of `address-model`? In cmake it should be possible to change the default based on the `CMAKE_SIZEOF_VOID_P` variable.

arvidn:
good point. on a second thought, I'm a bit hesitant to disable the *option* to use mmap disk IO on 32 bit builds.

Issue #4302, opened 2020-02-01
libtorrent not honouring announce_to_all_trackers settings


an0n666:
libtorrent version (or branch):
1.2.3
platform/architecture:
Windows 10
compiler and compiler version:

please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

In Qbit latest version 4.2.1 using libtorrent 1.2.3 if you set Announce to all trackers to false libtorrent still announcing to all trackers in same tier. If downgraded to 4.1.9.1 which use libtorrent 1.1x this problem doesn't exist. So I'm suspecting the problem is with libtorrent 1.2.3.

arvidn:
I'm trying to reproduce this by [extending the unit tests](https://github.com/arvidn/libtorrent/pull/4306), and I cannot. There's a possibility this is an issue on qBT. Could you provide some more details on the behavior and the expected behavior?
Are you certain of the tiers the trackers belong to?

arvidn:
for instance, are you sure all trackers are not part of different tiers? perhaps the `announce_to_all_tiers` changed its default between qbt-4.1.9 and qbt-4.2.1. Could you try changing the `announce_to_all_tiers` setting as well?

an0n666:
Ok I found an interesting thing. When I bind Qbit to a specific interface and specific IP address it honours the settings. But when we bind to all interfaces it announces to all trackers.

arvidn:
interesting. This really overlaps with what I've been working on fixing for 1.2.4. It would be great if you would have a chance to try a recent build. However, I don't think this issue is fixed, but I think I know what's going on.

The way multi-homing support works is by *attempting* to announce from all source IP addresses, but under normal circumstances, only one will work and the other ones will be considered "no route to host". I bet this failure of specific endpoints for a tracker spills over into the whole tracker being considered not working. If a tracker isn't working, the next one in the tier is tried.

I think the logic needs to be that *all* endpoints for a tracker fail before the tracker is considered failed.

an0n666:
So who's considering the tracker as failed when one of the endpoints fail? Qbit or lib?
If Qbit decides that then the issue has to be fixed on their end. I tried with Qbit master as well which contains this PR but it doesn't fix the multi announce issue. 
https://github.com/qbittorrent/qBittorrent/pull/11637

arvidn:
Here's an attempt to fix this: https://github.com/arvidn/libtorrent/pull/4310
[...]

PR #4310, opened 2020-02-03
fix announce_to_all_trackers and announce_to_all_tiers behavior

arvidn:
This solves a problem described here: https://github.com/arvidn/libtorrent/issues/4302

When an endpoint for a tracker fails, previously the whole tracker was considered having failed (for purposes of `update_tracker_timer()` and `announce_with_tracker()`) causing it to move on to the next tracker, regardless of whether the main endpoint worked.

In the test I added a multi-homed peer (IPv4 and IPv6) and IPv4-only trackers. When the IPv6 announce failed, it would move on to the next tracker, even though it shouldn't have.

I think this behavior was deliberate to search for another tracker that would support IPv6 (in this case), however, it would also cause an IPv4 announce to the next tracker, even though IPv4 was working fine with the main tracker.

With the existing logic, `announce_to_all_trackers` would effectively always be true (when having a listen_socket that couldn't reach any tracker).

This patch attempts to fix this by just not moving forward in the tracker list, as long as at least one endpoint is working. It may be ideal to fix this by allowing each endpoint to search for a tracker that works for it independently, but that would require a lot more complexity and/or an overhaul of the current tracker announce logic. It's not obvious to me that it's worth it.

@ssiloti would you mind taking a look?

A lost of the changes to the unit test were just simplifications. Since it's only testing tracker announce logic, there's no reason to have a second peer or transfer any data.

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -2999,7 +2999,15 @@ bool is_downloading_state(int const st)
 				if (!aep.can_announce(now, is_seed(), ae.fail_limit))
 				{
 					// this counts
-					if (aep.is_working()) state.sent_announce = true;
+					if (aep.is_working())
+					{
+						state.sent_announce = true;
+						if (!settings().get_bool(settings_pack::announce_to_all_trackers)
+							&& !settings().get_bool(settings_pack::announce_to_all_tiers))
+						{
+							state.done = true;
+						}
+					}
 					continue;
 				}
 
@@ -3274,7 +3282,7 @@ bool is_downloading_state(int const st)
 				aep->last_error.clear();
 				aep->message = !resp.warning_message.empty() ? resp.warning_message : std::string();
 				int tracker_index = int(ae - m_trackers.data());
-				m_last_working_tracker = std::int8_t(prioritize_tracker(tracker_index));
+				m_last_working_tracker = std::int8_t(tracker_index);
 
 				if ((!resp.trackerid.empty()) && (ae->trackerid != resp.trackerid))
 				{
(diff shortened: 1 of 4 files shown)


arvidn:
actually, this might not work right if there's a tracker that resolves *only* to IPv6 (for a client that doesn't have IPv6 connectivity). The endpoint would be disabled and considered "done", and maybe prevent moving on to the next tracker. I should extend the unit test with this case

arvidn:
I think maybe the `done` state shouldn't be per endpoint, but per tracker

src/torrent.cpp
```diff
@@ -2999,7 +3005,11 @@ bool is_downloading_state(int const st)
 				if (!aep.can_announce(now, is_seed(), ae.fail_limit))
 				{
 					// this counts
-					if (aep.is_working()) state.sent_announce = true;
+					if (aep.is_working())
+					{
+						state.sent_announce = true;
+						state.done = true;
```
ssiloti: I think this needs only set `done` when neither `announce_to_all_tiers` nor `announce_to_all_trackers` are set (i.e. same as the existing logic for setting `done`). I also think this is the only change need to fix #4302. The other `state.done = true` instances added by this PR should not be there. We should not abandon announcing on an endpoint just because one tracker isn't reachable via that endpoint.

arvidn: I’ll test that. Regarding stopping without trying to announce all endpoints; right now content busing to the new tracker ends up announcing the first endpoint again. I didn’t feel confident I fully understood this logic to fix that. But maybe I should just make sure *all* cases are covered by the test

arvidn: I implemented your suggestion. it turns out the call to `prioritize_tracker()` had to be taken out as well, to support the case where different endpoints end up using different trackers. I think the call to `deprioritize_tracker()` should be enough though

PR #4339, opened 2020-02-20
Attempt to improve reliability of enumerating routes by not filtering by routing table ID

arvidn:
improve error message if enumerating routes fail. Add special case if there's only a single non-loopback listen socket, and just assume it has a gateway (to make it work even if we fail to enumerate routes).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* fix issue with knowing which interfaces to announce to trackers and DHT
 	* undeprecate settings_pack::dht_upload_rate_limit
 
 1.2.4 release
--- a/docs/hunspell/libtorrent.dic
+++ b/docs/hunspell/libtorrent.dic
@@ -556,3 +556,4 @@ netmask
 fe80
 vcpkg
 leecher
+6881l
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -776,7 +776,10 @@ unless the following conditions are met:
 * the IP address is not in a known loopback range
 * the item the IP address was expanded from was not marked local (``l``)
 * the IP address is in a globally reachable IP address range OR the routing
-  table contains a default route with a gateway for the corresponding network
+  table contains a default route with a gateway for the corresponding network.
+  This bullet only applies when expanding from an unspecified IP address. When
+  explicitly specifying a device, we don't need to find a route to treat it as
+  external.
 
 The NAT-PMP/PCP and UPnP port mapper objects are only created for networks that
 are expected to be externally available (i.e. not "local network"). If there are
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -733,12 +733,11 @@ setting.
 Multi-homed hosts
 =================
 
-libtorrent has solid support for multi-homed hosts, starting with version 1.2.4.
-the settings_pack::listen_interfaces setting is used to specify which interfaces/IP addresses
+The settings_pack::listen_interfaces setting is used to specify which interfaces/IP addresses
 to listen on, and accept incoming connections via.
 
-Each entry in ``listen_interfaces`` is an IP address or a device name, followed
-by a listen port number. Each entry (called ``listen_socket_t``) will have the
+Each item in ``listen_interfaces`` is an IP address or a device name, followed
+by a listen port number. Each item (called ``listen_socket_t``) will have the
 following objects associated with it:
 
 * a listen socket accepting incoming TCP connections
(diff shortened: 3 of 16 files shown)


arvidn:
@ssiloti Do you have any thoughts on this? It turns out that enumerating the routing table is unreliable (at least on exotic setups on linux). This is my attempt to rely less on it working. This is the gist of this change (at least the intention of it).

* instead of marking a `listen_socket_t` as having a gateway, they are all assumed to effectively have one (or be independent IPs that should be announced to trackers).
* any IP address that's specified explicitly in `listen_interfaces` will always be announced to trackers
* there is special logic to guess whether an IP address should be announce to trackers or not *only* for IP addresses that are expanded from the unspecified address, or from a device name.

The special logic assumes that all IPs are private (link-local, connects to a local container or a private local network not connected to the internet). This is indicated with the `local_network` flag. These `listen_socket_t`s are not announced to trackers (unless the tracker happens to be in that network).
If the IP address from an expanded entry (unspecified or device name) is a known globally reachable IP address OR it's a known private network address with a gateway that we found in the routing table, then we clear the `local_network` flag, and we announce it.

The user can specify the `l` flag at the end of an (IP, port)-pair in `listen_interfaces` to explicitly say it's a local network. This can be used for private/internal networks, or for tests.

arvidn:
@FranciscoPombal I invite anyone to review the logic that I described above and/or the code that's meant to implement it!

arvidn:
these are the problems still:

1. If you use the default `listen_interfaces` with a regular VPN, you will "listen" on both the plain IP *and* the VPN IP. Both will be announced to the tracker (but the plain announce will still go through the VPN, I'm pretty sure). The problem is you end up announcing twice.

2. If you use a wireguard VPN, libtorrent fails to detect that the VPN device is connected to anything. It's a point-to-point connection, but there's nothing in the routing table or interface list that suggests you can route over it (at least as far as I can tell)

arvidn:
I keep coming back to the solution of not (automatically) expanding unspecified addresses, but treat them as a *single* interface. The idea being if you have a more sophisticated setup, you can change the `listen_interfaces` setting to accommodate.

@ssiloti I recall you argued against this. But as far as I can see, doing so has a non trivial cost to the common cases (of using these two kinds of VPNs), just to support a rare case (at least afaict). It seem hard to justify.

arvidn:
actually, I did some more testing and discovered:

1. when using OpenVPN style routing, where the unspecified address is expanded to both the "plain" and tunnel IP, only the tunnel tracker announce will go through, the other announce will be blocked.

2. By looking for a default route instead of a gateway, the heuristic seem to do the right thing for wireguard-style routing as well, where there isn't a gateway, but a point-to-point interface.

FranciscoPombal:
> I keep coming back to the solution of not (automatically) expanding unspecified addresses, but treat them as a _single_ interface. The idea being if you have a more sophisticated setup, you can change the `listen_interfaces` setting to accommodate.
> 
> @ssiloti I recall you argued against this. But as far as I can see, doing so has a non trivial cost to the common cases (of using these two kinds of VPNs), just to support a rare case (at least afaict). It seem hard to justify.

Just chiming in to say that IMO, generically speaking the default setup/most obvious options should support as many common cases as possible OOB. For other oddball or complicated setups, the options for them should be available and well-documented.

For me, the "most common cases" here means for example someone using qBittorrent from a "normal" home/university/work connection with or without a ("simple") VPN (wireguard or other)/SOCKS5 setup.

I'm afraid I can't help with any technical suggestion about this though.

ssiloti:
Listening on the unspecified address is a problem for the DHT. There would need to be a filter restricting incoming traffic to a single address and a heuristic for picking that address (e.g. the most common destination address of the last n incoming packets). On multi-homed systems there would probably be some spurious outgoing traffic with the wrong source, but I'm not sure how much of a problem that would be.

Safari77:
If I use qbittorrent with this route-error-message branch , I get only incoming connections (to the specified listening port, e.g. 9999 in this case), even though PEX works with this branch and v1.2.3.  If I run ```tcpdump -i wg not port 9999``` I don't get any other traffic (besides tracker and dns)!

With v1.2.3 I get more connections and starting up the torrent download goes much faster.

arvidn:
@Safari77 thanks for testing! Those other connections, not on port 9999, can you describe those in some more detail?

* are they incoming connections? (i.e. are you not receiving incoming connections with this PR?)
* do they go over the `wg` interface? (perhaps there's a problem in 1.2.3 where connections over the normal network interface is allowed).
* Are all your connections TCP or do you have uTP connections as well? I would expect all uTP traffic to go through a single UDP socket bound to port 9999 on your end, so you wouldn't be seeing any of that.

So, I'm curious if the failure in this branch is that you can't make outgoing connections, or that you can't make TCP connections.

Also, I recall your routing table from the other github ticket. Do you still not have a default route to the `wg` interface?
Do you specify an IP or interface name in `listen_interfaces`? I'm interested in whether it makes a difference for you to leave that as `0.0.0.0:9999` or `wg:9999`.

Safari77:
With v1.2.3 I have both incoming and PEX, all over wg. 
I have configured qB to listen on wg device, all IPv4 addresses. I have TCP connections only, uTP is disabled.  I have route on wg interface and it works fine.  ```iptables --list``` shows 60000 accepted outgoing TCP connections for uid of the qbittorrent user.
Well, I don't have to debug this this week, I have ten other things I should be doing.

arvidn:
`iptables --list` doesn't answer the question of whether you have a default route for the `wg` interface though. I assume you still don't, so I pushed a commit that removes the condition that a default route must be found for the interface, when an interface is explicitly configured.

arvidn:
I was hoping more people could test this. But I don't think it makes much sense to delay much longer

PR #4343, opened 2020-02-21
Fix possible output truncation in test_dht

FranciscoPombal:
NOTE: I was not sure which branch to target so I just chose `RC_1_2`, but let me know if it should be `master`.

From the `printf(3)` man page (emphasis mine, the same applies to `std::snprintf`):

> int snprintf(char *str, size_t size, const char *format, ...);

> The functions **snprintf()** and vsnprintf() **write at most size bytes (including the terminating null byte** ('\0')) to str

Thus, to write something that is at most 10 chars in length to str, `sizeof(str)` must be at least 11, to account for the terminating null byte.

This fixes the warning I was getting when compiling (C++14 mode).

```
/home/francisco/Documents/libtorrent/test/test_dht.cpp: In function ‘void unit_test_traversal_done()’:
/home/francisco/Documents/libtorrent/test/test_dht.cpp:2614:35: warning: ‘__builtin___snprintf_chk’ output may be truncated before the last format character [-Wformat-truncation=]
   std::snprintf(tok, sizeof(tok), "%02d", i);
                                   ^~~~~~
In file included from /usr/include/stdio.h:862,
                 from /usr/include/c++/8/cstdio:42,
                 from /usr/include/c++/8/ext/string_conversions.h:43,
                 from /usr/include/c++/8/bits/basic_string.h:6400,
                 from /usr/include/c++/8/string:52,
                 from /opt/boost/include/boost/utility/string_view_fwd.hpp:21,
                 from /opt/boost/include/boost/utility/string_view.hpp:24,
                 from /home/francisco/Documents/libtorrent/include/libtorrent/string_view.hpp:73,
                 from /home/francisco/Documents/libtorrent/include/libtorrent/address.hpp:39,
                 from /home/francisco/Documents/libtorrent/test/test.hpp:36,
                 from /home/francisco/Documents/libtorrent/test/test_dht.cpp:33:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:64:35: note: ‘__builtin___snprintf_chk’ output between 3 and 11 bytes into a destination of size 10
   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
          ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        __bos (__s), __fmt, __va_arg_pack ());
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
```
$ gcc --version
gcc (Ubuntu 8.3.0-6ubuntu1~18.04.1) 8.3.0
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

### DIFF
--- a/test/test_dht.cpp
+++ b/test/test_dht.cpp
@@ -2610,7 +2610,7 @@ TORRENT_TEST(traversal_done)
 			TEST_ERROR(t.error_string);
 			continue;
 		}
-		char tok[10];
+		char tok[11];
 		std::snprintf(tok, sizeof(tok), "%02d", i);
 
 		msg_args args;


FranciscoPombal:
I just noticed something. In the case I fixed, the value being copied is that of the variable being iterated:

```cpp
for (int i = 1;; ++i)
{
     /*...*/
    char tok[10];
    std::snprintf(tok, sizeof(tok), "%02d", i);
    /*...*/
}
```

But there are other places that might have the same issue, that the compiler is not complaining about for some reason. Here the value being copied can potentially be incremented up to a sufficiently large value to trigger the truncation. However, the compiler is not detecting that, maybe due to it being declared outside of the loop, incremented in the middle of it, or a combination of both. I think these should also be changed to `11`.

```cpp
int idx = -1;
for (auto& node : nodes)
{
    /*...*/
    ++idx;
    char tok[10];
    std::snprintf(tok, sizeof(tok), "%02d", idx);
    /*...*/
}
```

Link: https://github.com/arvidn/libtorrent/blob/68a57218e7bb0d1bc73df7c8cf04f86b7b53e4d6/test/test_dht.cpp#L2373

@arvidn Do you think my suspicion is correct?

arvidn:
It’s unclear what problem you’re trying to solve. I don’t think your reading of the snprintf man page is correct. A buffer of 10 bytes is sufficient if specifying n as 10.

See this version of the man page: http://www.skrenta.com/rt/man/snprintf.3.html

However, in your patch you just increase the buffer size, you still pass in the size of the buffer as n to snprintf.

Is the problem that the integer might have too large a value to fit in 9 digits?

FranciscoPombal:
@arvidn 

> It’s unclear what problem you’re trying to solve. I don’t think your reading of the snprintf man page is correct. A buffer of 10 bytes is sufficient if specifying n as 10.
> 
> See this version of the man page: http://www.skrenta.com/rt/man/snprintf.3.html

Any particular reason for using that man page specifically as reference? The date at the bottom reads "16 September 1995". My quote was from the man pages included in Ubuntu 18.04 (dated 2017-09-15). Turns out there the latest version at the time of writing is dated 2019-10-10 (available here: http://man7.org/linux/man-pages/man3/snprintf.3.html, which I believe is the most authoritative source). However, the relevant part has not changed from the version I read from to the latest, and I believe they are clearer about this matter than the 1995 page.

> However, in your patch you just increase the buffer size, you still pass in the size of the buffer as n to snprintf.
> 
> Is the problem that the integer might have too large a value to fit in 9 digits?

An `int` (32 bit) has at most 10 digits. Take `INT_MAX`, for example: `2,147,483,647`; When it is converted to a string with the `%02d` format specifier (in this case it is functionally identical to just "%d"), the string will be: `2147486647\0`. This is 11 bytes in total, notice the `\0`.

For the same reasons the smallest string generated with a 32-bit int and the `%02d` format specifier will be 3 bytes wide. For example the number 2 becomes `02\0`.

That's why the error says `output between 3 and 11 bytes into a destination of size 10`.

Since `snprintf` writes at most n (in this case 10, the size of the buffer) bytes **including** the `\0` (meaning `\0` is the nth byte, in this case the 10th byte) to the destination, the string at the destination would become `214748664\0` (notice the last digit was truncated by the `\0`).

Increasing the buffer size to 11 solves the (potential) problem by making room for 10 digits + terminating `\0`. Passing the size of the destination buffer as n to `snprinf` is the correct thing to do regardless.

arvidn:
I see. At first I thought you suggested the string might not be 0-terminated currently. But, yes, an`int` can have values large enough to cause truncation. In practice that won't happen in this test, but I'm fine with this patch since it seem to fix a warning.

FranciscoPombal:
@arvidn Thanks. Can you please also address this comment: https://github.com/arvidn/libtorrent/pull/4343#issuecomment-589898006?

arvidn:
> I am pretty sure this code is also technically incorrect even though it does not currently produce a warning.
> @arvidn Do you think my suspicion is correct?

Yeah, any time an `int` is printed into a 10 character buffer, there's potential for truncation. I would expect that almost none of the integers are greater than 2 billion though, so I don't think it's an issue in practice.

FranciscoPombal:
@arvidn 

> Yeah, any time an `int` is printed into a 10 character buffer, there's potential for truncation. I would expect that almost none of the integers are greater than 2 billion though, so I don't think it's an issue in practice.

Yes, like the case addressed in this PR, it's more about correctness, and it prevents a compiler warning. I'm going to submit a follow-up PR to address the other cases, for both correctness and preemptive reasons, since a smarter compiler in the future might issue warnings about them as well.

PR #4344, opened 2020-02-22
update features docs with bittorrent v2

arvidn:


docs/features.rst
```diff
@@ -17,12 +17,44 @@ desktops. It boasts a well documented library interface that is easy to
...
+`BEP 52`_). BitTorrent V2 introduces a new format for .torrent files, whish is
+smaller than the original format.
+
+1. Files are organized in a directory structure, instead of listing full paths.
+   Torrents that have a lot of files in deep directory structures will use a lot
+   less space to represent that structure in a v2 torrent.
+
+2. Piece hashes are organized in a merkle hash trees per file, and only the
+   roots of the trees are included in the .torrent file. The actual hashes are
+   delivered by peers.
```
ssiloti: Not just the roots are in the .torrent file, the piece layer hashes are stored as well, albeit outside of the info-dict. This also impacts the claim above about v2 torrent files being smaller. In the common case where the torrent only contains a handful of files the v2 torrent will be larger due to the larger hashes.

arvidn: Good point

PR #4353, opened 2020-02-27
explore not expanding unspecified addresses

arvidn:


FranciscoPombal:
What is this intended to fix/improve and how does it impact BEP 45 compliance?

arvidn:
I think the impact would be that if one has two routes to the internet with the same address family, one would incorrectly share one DHT node for those external IPs, by default. But it would be possible to configure libtorrent to not do that, by explicitly specifying those interfaces/IPs.

FranciscoPombal:
> I think the impact would be that if one has two routes to the internet with the same address family, one would incorrectly share one DHT node for those external IPs, by default. But it would be possible to configure libtorrent to not do that, by explicitly specifying those interfaces/IPs.

I suppose that's not too bad, since at least one would still be sharing a DHT node for the correct addresses. But still, would this not mean that libtorrent would leak addresses that it should not be leaking if the user is using a VPN and wishes that libtorrent only uses the VPN? If that is the case it should be documented that VPN users most likely will want to explicitly specify interfaces/IPs.

arvidn:
After my discoveries in #4339, I'm less inclined to pursue this path

Issue #4361, opened 2020-02-29
Windows x86 - error posix_storage - fix utf8_wchar


pavel-pimenov:
libtorrent version (or branch): master
platform/architecture: Win /x86
compiler and compiler version: VC++ 2019

![изображение](https://user-images.githubusercontent.com/920705/75605434-e1cbbd00-5af3-11ea-93f8-7194679cc2c1.png)

Fix:

![изображение](https://user-images.githubusercontent.com/920705/75605779-e0e85a80-5af6-11ea-8f31-b2d9050afa7b.png)

PR #4368, opened 2020-02-29
add utf8_wchar

pavel-pimenov:
https://github.com/arvidn/libtorrent/issues/4361


src/posix_storage.cpp
```diff
@@ -512,10 +515,18 @@ namespace aux {
 	{
 		std::string const fn = files().file_path(idx, m_save_path);
 
-		char const* mode_str = (mode & open_mode::write)
+		const auto mode_str = (mode & open_mode::write)
```
arvidn: do you have objections to explicitly decay the array to a pointer her? i.e. `auto const* `

arvidn:
I would prefer a single commit. Could you squash these?

arvidn:
there's whitespace at the end of this line: https://travis-ci.org/arvidn/libtorrent/jobs/656753515#L617

PR #4376, opened 2020-03-03
add support for TCP_NOTSENT_LOWAT

arvidn:


src/session.cpp
```diff
@@ -236,8 +236,7 @@ namespace {
 
 		set.set_int(settings_pack::max_rejects, 10);
 
-		set.set_int(settings_pack::recv_socket_buffer_size, 1024 * 1024);
-		set.set_int(settings_pack::send_socket_buffer_size, 1024 * 1024);
```
the8472: Do these settings also apply to UDP? The kernel's socket buffer auto sizing and TCP_NOTSENT_LOWAT only applies to TCP. https://github.com/torvalds/linux/blob/98d54f81e36ba3bf92172791eba5ca5bd813989b/Documentation/networking/ip-sysctl.txt#L567-L571
For UDP you would still want to set explicit buffer sizes.

arvidn: I'm pretty sure I don't apply this to the UDP socket, but I'll double check. Worst case though it would just fail I suppose

arvidn: It's only used for peer connections, and the utp_stream will ignore all socket options set on it

PR #4396, opened 2020-03-06
Specify port=1 in announces via proxy

Snawoot:
In case when proxy is enabled, libtorrent sends port=0 parameter in announce request. Some trackers (like [rutracker.org](https://rutracker.org)) reject announce requests when specified port is zero.

I've noticed there are already exception made for I2P to send fake port number, because non-zero port is required. As it appears to me in case when only passive downloads possible, just like in proxy case, and port number in announce is not meaningful, zero value of port doesn't help much and it can cause problems on it's own. Instead of it, some trivial value like `1` may pass tracker sanity checks where `0` don't.

Actual PR ensures port parameter in announce always has non-zero value, catching up all these cases with proxies and absence of real port number.

I'm not sure this patch is accurate, but that is what I have to use, so I've decided to share.

### DIFF
--- a/src/http_tracker_connection.cpp
+++ b/src/http_tracker_connection.cpp
@@ -121,7 +121,7 @@ namespace libtorrent {
 				, escape_string({tracker_req().pid.data(), 20}).c_str()
 				// the i2p tracker seems to verify that the port is not 0,
 				// even though it ignores it otherwise
-				, i2p ? 1 : tracker_req().listen_port
+				, tracker_req().listen_port == 0 ? 1 : tracker_req().listen_port
 				, tracker_req().uploaded
 				, tracker_req().downloaded
 				, tracker_req().left
--- a/simulation/test_tracker.cpp
+++ b/simulation/test_tracker.cpp
@@ -353,7 +353,7 @@ void test_ipv6_support(char const* listen_interfaces
 
 	// if we're not listening we'll just report port 0
 	std::string const expect_port = (listen_interfaces && listen_interfaces == ""_sv)
-		? "&port=0" : "&port=6881";
+		? "&port=1" : "&port=6881";
 
 	http_v4.register_handler("/announce"
 	, [&v4_announces,expect_port](std::string method, std::string req


src/http_tracker_connection.cpp
```diff
@@ -121,7 +121,7 @@ namespace libtorrent {
 				, escape_string({tracker_req().pid.data(), 20}).c_str()
 				// the i2p tracker seems to verify that the port is not 0,
 				// even though it ignores it otherwise
-				, i2p ? 1 : tracker_req().listen_port
```
arvidn: is there logic elsewhere that makes sure that `listen_port` is always zero for i2p torrents?
It's not obvious that this change is correct.

arvidn: it seems simpler to remove the condition here, and make sure we set `listen_port` to 1, wherever we set it.

Snawoot: Hello, @arvidn 

> is there logic elsewhere that makes sure that listen_port is always zero for i2p torrents?

If it wasn't the case, you wouldn't have to ensure port is non-zero in the first place.

> it seems simpler to remove the condition here, and make sure we set listen_port to 1, wherever we set it.

I'm not sure about that. port=1 is a valid port and hypothetically we can have real listen_port equal to 1. On the other hand, announces is the only place where port number has to be "spoofed". We need that change only here (as i2p did before).

arvidn: > If it wasn't the case, you wouldn't have to ensure port is non-zero in the first place.

The existing code doesn't *just* make sure the port is non-zero, it also makes sure that whatever value it might have is not announced. i.e. it protects against information leaks.

I'm thinking it would likely be an overall simplification if the code building the `tracker_request` object would set the listen port to 1, and not have any special case here.

For instance, it would make UDP trackers behave the same.

> I'm not sure about that. port=1 is a valid port and hypothetically we can have real listen_port equal to 1. On the other hand, announces is the only place where port number has to be "spoofed". We need that change only here (as i2p did before).

I'm not suggesting making the *actual* listen port 1, but to fill in the `tracker_request` object with a listen port of 1, when it is supposed to not be there.

Btw, do you know what other popular clients do in this case? uTorrent, Transmission or rTorrent?

Snawoot: > The existing code doesn't just make sure the port is non-zero, it also makes sure that whatever value it might have is not announced. i.e. it protects against information leaks.

I'm not familiar with existing code base that well, so I'm lost about how to implement it properly. But I don't think port number is a much of a leak as far as real client IP address is hidden. I was pretty sure `tracker_request.listen_port` tends to be `0` when proxy-only operation of client is enabled, so I had thought it is sufficient criteria and main goal is to make port number in announce look always viable.

> Btw, do you know what other popular clients do in this case? uTorrent, Transmission or rTorrent?

Transmission has no proxy support at all (it had, but proxy support was eliminated long time ago). I haven't tried uTorrent for a long time because it lost credibility long time ago when they bundled application with a background cryptocurrency miner.

qBittorrent is the client I use. It relies on libtorrent-rasterbar, so I had to make this patch to make qBittorrent actually work behind proxy. Otherwise, it published `port=0` and such announces were mostly rejected.

Also I've tried Deluge which is also based on `python-libtorrent` which inherently is a binding to libtorrent-rasterbar library. And it had similar issue: `port=0`, announces rejected, downloads not happening.

I don't know any more torrent clients with viable proxy support. Maybe rtorrent also does, but I haven't tried it yet, last time I've checked it appeared to me too complex. Probably I should revisit it and give it another chance.

BTW, qbittorrent has issues with proxy on its own: qbittorrent/qBittorrent#12114 Latest versions of qbittorrent have terrible IP address leak with older versions of libtorrent-rasterbar (pre-1.2.0).

arvidn:
@Snawoot what do you think of this patch? https://github.com/arvidn/libtorrent/pull/4400
Would you mind testing it?
It's a little bit bigger, but it fixes this same issue for UDP trackers as well, and makes it a bit clearer what port is being announced.

Snawoot:
@arvidn Thank you! I'll build and test it soon. I'm not sure I'll be able to test UDP trackers because I use SOCKS5 proxy (Tor) and I'm not sure there is UDP binding provides by it's SOCKS interface.

Snawoot:
@arvidn I've built and tested branch from #4400 and it works as expected with and without proxy :thumbsup:

arvidn:
thanks!

PR #4398, opened 2020-03-07
run move_storage disk jobs in the hasher threads

arvidn:
to force them to be serialized to a greater extent


arvidn:
come to think of it, this is probably not that great..

PR #4406, opened 2020-03-08
change default lease duration for UPnP port maps from indefinite to 1…

arvidn:
… hour

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* make UPnP port-mapping lease duration configurable
 	* deprecate the bittyrant choking algorithm
 	* add build option to disable streaming
 
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -829,7 +829,6 @@ namespace aux {
 
 			void update_socket_buffer_size();
 			void update_dht_announce_interval();
-			void update_anonymous_mode();
 			void update_download_rate();
 			void update_upload_rate();
 			void update_connections_limit();
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -1735,6 +1735,13 @@ namespace aux {
 			// option.
 			send_not_sent_low_watermark,
 
+			// The expiration time of UPnP port-mappings, specified in seconds. 0
+			// means permanent lease. Some routers do not support expiration times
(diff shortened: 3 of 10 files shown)


arvidn:
there should be a setting for the default lease time, so people with routers that don't support it can set it to 0.

ChangeLog
```diff
@@ -1,5 +1,8 @@
+	* make UPnP port-mapping lease duration configurable
+
 1.2.5 release
 
+	* chagne default lease duration of UPnP port maps to 1 hour. Fall back to indefinite if not supported
```
NotTsunami: Typo in change

NotTsunami:
Is there a reliable way to test whether this is supported by a router? If it should just be a basic tunable on the qBit side, I'm more than happy to send an accompanying PR to qBit if it isn't already in the works.

FranciscoPombal:
@NotTsunami See https://github.com/qbittorrent/qBittorrent/issues/12122#issuecomment-596192433.
The problem is that some routers don't support the temporary leases feature nor a way to report that they don't. On those routers, using temporary leases will fail silently.

arvidn:
has anyone had a chance to try this out? (I only have one UPnP router, and they're all different)

FranciscoPombal:
What case were you able to test with your router?
Did you test with or without leases, and if you tested with leases, did the connection work again post-restart after the remapping?

arvidn:
I've only tested that the periodic remapping happens at all (and that the mapping requests look good). I found a bug in the process, fixed in the top commit in this patch.

I should try permanent leases too

arvidn:
permanent leases seem to work too

FranciscoPombal:
Ok, so the only remaining cases to test are:
1. that upnp works with leases disabled in routers that fail silently with leases enabled
2. that libtorrent falls back to not use leases if the router reports it does not support them
3. that the port mappings that are re-established after a router restart work (and that libtorrent doesn'þ think the old ones are still in effect).

I don't have a router that I can test the first 2 cases with, and restarting my home router for testing this in the near future is not an option due to the quarantine.

@whoodini @Seeker2

if any of you could help test this patch, it would be appreciated.

Seeker2:
I might if I had a working client that uses the latest changes, but I'm not going to try to compile it myself.
I have a ComCast modem-router-wifi-gateway device (the XB3 model) that enables UPnP even when it's disabled in settings...that I know ignores the UPnP leash duration period by days.
I also have another router I might test.

PR #4415, opened 2020-03-13
Tracker ssl validation

arvidn:


src/openssl.cpp
```diff
@@ -0,0 +1,88 @@
...
+#endif // OPENSSL_VERSION_NUMBER
+
+ssl::context::method ssl_version(int const v)
+{
+	switch (v)
+	{
+		case settings_pack::ssl3: return ssl::context::sslv3;
+		case settings_pack::ssl23: return ssl::context::sslv23;
+		case settings_pack::tls11: return ssl::context::tlsv11;
+		case settings_pack::tls12: return ssl::context::tlsv12;
```
NotTsunami: Is TLS v1.3 not supported?

arvidn: hm. I must have looked at old documentation. it seems to be here: https://www.boost.org/doc/libs/1_72_0/doc/html/boost_asio/reference/ssl__context/method.html

NotTsunami: Looks like it was added in v1.69, so still fairly recent.

src/openssl.cpp
```diff
@@ -78,7 +78,8 @@ ssl::context::method ssl_version(int const v)
 		case settings_pack::ssl23: return ssl::context::sslv23;
 		case settings_pack::tls11: return ssl::context::tlsv11;
 		case settings_pack::tls12: return ssl::context::tlsv12;
-		default: return ssl::context::tlsv12;
+		case settings_pack::tls13: return ssl::context::tlsv13;
```
NotTsunami: Maybe guard against older boost versions? I'm fine if not but seems recent enough.

include/libtorrent/settings_pack.hpp
```diff
@@ -1747,6 +1759,18 @@ namespace aux {
...
+		enum ssl_version_t : std::uint8_t
+		{
+			// SSL version 3
+			ssl3,
+			// SSL version 2 or 3
+			ssl23,
+			// TLS version 1.1
+			tls11,
+			// TLS version 1.2
+			tls12
```
NotTsunami: CI is failing because tls13 needs to be added here too

include/libtorrent/settings_pack.hpp
```diff
@@ -1769,6 +1769,8 @@ namespace aux {
 			tls11,
 			// TLS version 1.2
 			tls12
```
NotTsunami: Comma

include/libtorrent/settings_pack.hpp
```diff
@@ -1747,6 +1759,22 @@ namespace aux {
...
+			ssl23,
+			// TLS version 1.1
+			tls11,
+			// TLS version 1.2
+			tls12,
+#if BOOST_VERSION >= 106900
+			// TLS version 1.3
+			tls13,
+#endif
+		};
```
FranciscoPombal: I don't think supporting SSL v3 or v2 is a good idea at all. And TLS <= 1.1 is also on its way out (https://hacks.mozilla.org/2020/02/its-the-boot-for-tls-1-0-and-tls-1-1/), but I think completely dropping support for it can be done in a few months from now.

arvidn: my experience (which may be a bit dated by now, but not more than 3 years probably) is that the web server that comes with python 2 *requires* clients to support SSL 2 or 3 (I think that's what the asio value means at least)

FranciscoPombal: Well python2 is EOL'd as well but even so, it defaults to disabling SSL 2 and 3, and recommends not using it at all as well:
https://docs.python.org/2/library/ssl.html#ssl-contexts (See the bit about `Changed in version 2.7.16`)
https://docs.python.org/2/library/ssl.html#protocol-versions

arvidn: it seems python supports at least TLS 1.1. Now, how confident are we that no trackers require an old SSL version?

FranciscoPombal: @arvidn 
> it seems python supports at least TLS 1.1.

Yes, like I said, totally removing support TLS 1.0 and 1.1 will have to wait for later. Even Firefox is simply doing a "soft" deprecation, where you can still enable usage at the press of a button (see the link above and the release notes for FF 74).

> Now, how confident are we that no trackers require an old SSL version?

Very. If there are any, they should have upgraded long ago anyway.

PR #4417, opened 2020-03-13
WIP: improve rate based choker

arvidn:


arvidn:
There should probably be a lower limit on the number of unchoke slots still. Say, never fewer than 2 unchoke slots (+ the optimistic unchoke slot)

src/choker.cpp
```diff
@@ -317,28 +313,28 @@ namespace {
...
-				if (rate < rate_threshold) break;
+				// always have at least 1 unchoke slot
+				if (rate < rate_threshold && i >= 1) break;
 
 				++upload_slots;
 
-				// TODO: make configurable
-				rate_threshold += 1024;
+				// double the threshold
+				rate_threshold += rate_threshold;
```
an0n666: Maybe check 
if ( rate_threshold <= 1048576)
rate_threshold += rate_threshold;
to prevent the rate being increased infinitely.

arvidn: could you provide some more rationale for that? That specific logic you propose would have a pretty weird singularity right at that point..

an0n666: Let’s say it’d take 8 MiB/s to unchoke the next peer. But the next peer can only reach 7.9 MiB/s because of the bottleneck on my end. Then I’ll be wasting that 7.9MiB/s upload capacity. Whereas if we had a max rate limit of lets say 1 MiB/s we would lose max 0.99MiB/s upload capacity.

arvidn: but you have to weigh that against the negative effects of having too many unchoke slots too. If all we care about is using all our upload capacity, unlimited unchoke slots would waste even less. but if your storage is spinning rust, of course, you may have worse performance with too many unchoke slots. But you may also drop below other peer's reciprocation rate.

I'm not saying the current algorithm is perfect, but I think more analysis is required.

For example, perhaps instead of doubling the threshold, it could increase by 50%.

arvidn: in fact, this might be an interesting telemetry metric for qbt. What is the typical reciprocation rate? But I also think it's important to have an algorithm that's robust to faster internet connections

an0n666: Maybe after reaching a certain rate you can change the multiplier to be 0.5x instead of 2x.

arvidn:
maybe we can simplify the problem to assuming that we can upload the same rate to all peers, i so what function should `f()` be:

*number_of_unchoke_slots* = f( *peer_upload_rate* )

In this patch it's: max(1, log2(*peer_upload_rate*) - 11)

(11 comes from the initial value of 1 kiB, 2^11). Here's a table to illustrate this function:

```
up-rate   unchoke-slots
0.5 kiB/s 1
  1       1
  2       2
  4       3
  8       4
 16       5
 32       6
 64       7
128       8
512       9
  1 MiB/s 10
  2       11
  4       12
  8       13
 16       14
```

in practice, all peers won't be able to receive at the max rate, so it's going to skew towards having fewer upload slots, since the threshold is not compared against the fastest peer, but the *Nth* fastest, moving down the table.

Perhaps this existing function would tend to open too few upload slots. I would think so.

FranciscoPombal:
In the example above of the 7.9 MiB/s peer, wouldn't that peer get optimistically unchoked at some point? Then it could replace a slower peer in the already-opened slots.

It would technically not fully solve the problem; after all unchoked peers are replaced by 7.9 MiB/s peers, no more 7.9 MiB/s peers would be added. In practice, however, at that point you would have 12 peers each at 7.9 MiB/s, so the bottleneck would most likely be elsewhere.

By the way, I believe there's a typo in the current function, it should be `max(2, log2(peer_upload_rate) - 11)` because of the hard limit of at least 2 slots.

What about comparing the threshold against the fastest peer instead of the nth fastest?

arvidn:
> In the example above of the 7.9 MiB/s peer, wouldn't that peer get optimistically unchoked at some point? Then it could replace a slower peer in the already-opened slots.

Yes, it would be optimistically unchoked (that's how we would discover faster peers). But it would still not make the rate_based_choker open up another slot, because the nth fastest peer would still have to be faster than 8 MiB/s.

> It would technically not fully solve the problem; after all unchoked peers are replaced by 7.9 MiB/s peers, no more 7.9 MiB/s peers would be added. In practice, however, at that point you would have 12 peers each at 7.9 MiB/s, so the bottleneck would most likely be elsewhere.

7.9 MiB/s *or faster*. if the nth fastest peer is 7.9 MiB/s, chances are than peers 0 through n are faster.

> By the way, I believe there's a typo in the current function, it should be `max(2, log2(peer_upload_rate) - 11)` because of the hard limit of at least 2 slots.

the lower limit is 1. See [here](https://github.com/arvidn/libtorrent/pull/4417/files#diff-5090984e616405485b194a4a42569f90R332). Only when `i == 0` will the condition be ignored.

> What about comparing the threshold against the fastest peer instead of the nth fastest?

I would worry that such logic would not be very stable, but the number of unchoke slots would vary wildly over time, and possibly cause a feed-back loop affecting the unchoking logic.

FranciscoPombal:
> Yes, it would be optimistically unchoked (that's how we would discover faster peers). But it would still not make the rate_based_choker open up another slot, because the nth fastest peer would still have to be faster than 8 MiB/s.

> 7.9 MiB/s _or faster_. if the nth fastest peer is 7.9 MiB/s, chances are than peers 0 through n are faster.

> the lower limit is 1. See [here](https://github.com/arvidn/libtorrent/pull/4417/files#diff-5090984e616405485b194a4a42569f90R332). Only when `i == 0` will the condition be ignored.

I see. I misunderstood the problem.

> I would worry that such logic would not be very stable, but the number of unchoke slots would vary wildly over time, and possibly cause a feed-back loop affecting the unchoking logic.

Then would about time-averaging the "fastest" speed? would that be feasible?

FranciscoPombal:
Ok so if the nth fastest is 7.9 MiB/s, should we care about adding another 7.x MiB/s peer? If we already have n peers 7.9 MiB/s or faster, then the bottleneck is now probably elsewhere.

Still with fast SSDs there would probably be wasted potential in this case.

How about an "extra unchoke slots" bucket for peers whose speed `x` is `speed_of_nth_fastest / 2 < x < speed_of_nth_fastest`?

arvidn:
one thing I'm tempted to do is to make the threshold increase less steep:

```
rate_threshold += rate_threshold / 2;
```

That would give the following table:

```
up-rate   unchoke-slots
   0.5 kiB/s 1
   1         1
   1.5       2
   2.3       3
   3.4       4
   5.2       5
   7.8       6
  11.7       7
  17.5       8
  26.2       9
  39.4      10
  59        11
  88.6      12
 132.9      13
 199.3      14
 298.9      15
 448.4      16
 672.6      17
1008.9      18
1513.3      19
```

I think that's more reasonable. Keep in mind that the rate threshold is for the *nth* fastest uploader. i.e. the further down that table, the less likely you are to have fast peers left.

FranciscoPombal:
Seems like it could work better.

Related note: how would you suggest testing this kind of stuff, so that we can compare the performance of each variant more objectively? It seems like we are shooting in the dark, but due to the enormous variability of real-world swarms, it's probably very difficult to test this in a meaningful way.

an0n666:
why 50% though? Why not 75 or 80 or 90. It’d be better if we could keep it configurable.

arvidn:
I don't think it's *terribly* important to keep the `rate_based_choker` particularly configurable. If you configure things, you most likely want to use the `fixed_slot_choker`

TiraelSedai:
@arvidn 

>  one thing I'm tempted to do is to make the threshold increase less steep

I'm obviously do not know what is the best approach here, but I want to offer two even steeper curves. Here is yours: https://i.imgur.com/spoeBCY.png

Now here are my options to consider:
(i is unchoke slots or iteration counter)
`rate_threshold += 3*rate_threshold / i;`
https://i.imgur.com/nvVkTzL.png

Or even **inverted**, which I theoretically like more but I have no idea how sensible it is but at least it looks logical in my head:
`rate_threshold += 200 / Math.Pow(i, 0.5);`
https://i.imgur.com/6qLDS2f.png

arvidn:
I welcome experiments with different curves. I think one thing that makes it difficult to evaluate these is that there are so many factors at play.

* utilizing ones upload capacity fully
* unchoking enough peers that our upload rate is above their reciprocation rate, but not by too much
* not unchoking so many peers that our I/O becomes the bottleneck

Most importantly, getting optimal reciprocation behavior depends almost entirely on what the rest of the swarm does. So, something working well now, may not a year from now.

I think one goal should be at least that this logic should behave well if all peers in the swarm implement it.

FranciscoPombal:
@arvidn Is this going to make it to 1.2.6? I think using `rate_threshold += rate_threshold / 2;
`, even if sub-optimal is better than nothing. A future PR can be dedicated to finding and using a better curve.

arvidn:
there are a few other fixes in there that definitely should make the next cut. That specific change to the logic is the one I'm the least confident about. I think restoring it to be linear would be a way to expedite landing this patch. It should probably be a steeper curve than 1kB/s. But I haven't had time to build simulations for all the scenarios I would like to be confident in this change of logic

NotTsunami:
Do you think there's any value in allowing the user to pick from different preset curves, or would the nature of people using different curves be harmful to the swarm? I do want to collect some analytics on different curves but that would take longer than ideal, and I'd like to see some form of this patch land in the near future. Also related, can we get a rebase?

arvidn:
I'm not sure to what extent it should be configurable. Ideally it would "just work" (that's kind of why it's here). I think configuration options would mostly be useful for users to experiment and update the defaults. So, only to the extent people are believed to do that.

And, yes, I will rebas

slrslr:
FranciscoPombal referred to this pull request from the qbittorrent client issue above. The problem was signifficant active (seeding & uploading) torrents decrease when the mode was changed from Fixed slots to "Upload rate based" choker. It happened on my computer, so in case it is needed and important i can i can supply information needed, though i am noob in these things so i would need step by step instructions. I am nto saying the issue that happened was not fixed already. It was libtorrent 1.2.10.0.

Pentaphon:
I think this should be a higher priority task than it is now because optimizing performance for people that don't change default settings is the absolute best way to make users happy. Most people just want their client to "just work" and "run fast" they aren't going to tweak anything themselves.

arvidn:
the default `choking_algorithm` in libtorrent is `fixed_slots_choker`

TheBestPessimist:
Should this be auto-closed?

Pentaphon:
None of these topics should be closed. Too many longstanding issues.

goffy59:
Still a major issue. The way I understand Upload Rate based is "provide whatever you can most optimally provide to the upload rate to peers/etc". But instead what it does is, reduce upload speed/etc to all peers/etc. I was hoping this setting would do what it sounds like which is upload as much as optimally as possible.

arvidn:
The upload_rate based choker (just like any choker) is to pick an appropriate number of peers to upload to to maximise download rate.

If you don't care about download rate, but just upload rate, you can use the fixed_slots choker and set a very high limit.

Pentaphon:
> If you don't care about download rate, but just upload rate, you can use the fixed_slots choker and set a very high limit.

I would think most people, if not all, would want the download rate to be as high as possible when downloading a torrent to 100% and then want the upload rate to be as high as possible when the torrent reaches 100% and begins seeding, no? I mean, it is possible to care about both your download and your upload rate when torrenting. Shouldn't the rate-based choker adjust itself to optimize for both download and upload rate after a torrent goes from leeching status to seeding status?

arvidn:
> Shouldn't the rate-based choker adjust itself to optimize for both download and upload rate after a torrent goes from leeching status to seeding status?

The rate_based_choker is only engaged with downloading torrents. When seeding the [seed_choking_algorithm](https://libtorrent.org/reference-Settings.html#seed_choking_algorithm) is used.

PR #4419, opened 2020-03-14
Invalid linker flag in genereated .pc file

### LINKED ISSUE #4122: Invalid linker flag in genereated .pc file

RC_1_2

platform/architecture:
Linux x64

compiler and compiler version:
gcc version 9.2.0
boost 1.71.0
cmake version 3.15.5

how to reproduce it.

I'm building libtorrent via cmake and I noticed this linker flag `-lvals-NOTFOUND` is appearing in the generated libtorrent-rasterbar.pc file which cause the linker to fail when building other programs.
I would expect that flag not to appear.

Build script: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=libtorrent-rasterbar-1_2-git#n29
Generated libtorrent-rasterbar.pc file:
```
prefix=/usr
libdir=${prefix}/lib

Name: libtorrent
Description: Bittorrent library
Version: 1.2.2
Libs: -L${libdir} -ltorrent-rasterbar -lpthread -lssl -lcrypto -lvals-NOTFOUND -lboost_system
Cflags: -fexceptions -std=c++14 -I/usr/include -DTORRENT_LINKING_SHARED -DBOOST_ASIO_ENABLE_CANCELIO -DTORRENT_USE_ICONV -DTORRENT_USE_OPENSSL -DTORRENT_USE_LIBCRYPTO -DBOOST_SYSTEM_DYN_LINK -DBOOST_ALL_NO_LIB
```

[build log.txt](https://github.com/arvidn/libtorrent/files/3890094/build.log.txt)

Also this might be relevant, from https://cmake.org/cmake/help/v3.16/command/get_target_property.html :
>  Non-inherited properties will set VAR to NOTFOUND, whereas inherited properties will search the relevant parent scope as described for the define_property() command and if still unable to find the property, VAR will be set to an empty string.

silverqx:
Fixes invalid linker flag `vals-NOTFOUND` in generated `libtorrent-rasterbar.pc`.

Fixes #4122


arvidn:
thanks!

Chocobo1:
Thanks for this, it fixes the issue.
@silverqx 
Care to backport (actually cherry-pick) it to RC_1_1?

PR #4420, opened 2020-03-14
set default build type for cmake

### LINKED ISSUE #4416: pkgconfig file doesn't contain boost_system lib with cmake
If I don't pass `-DCMAKE_BUILD_TYPE=Release|Debug` to cmake build command, I don't have `-lboost_system` in generated `libtorrent-rasterbar.pc` file.

I think there should be some defaults, if this option is not passed.

Cmake docs for [CMAKE_BUILD_TYPE](https://cmake.org/cmake/help/v3.16/variable/CMAKE_BUILD_TYPE.html?highlight=cmake_build_type)

silverqx:
Sets CMake build type, if none was selected with `-DCMAKE_BUILD_TYPE=` option, if the folder contains `.git` folder `Debug` is selected, otherwise `Release` is selected.

This is a good alternative, because if you download the package from GitHub from Release section eg `https://github.com/silverqx/libtorrent/archive/libtorrent-1_1_4.zip` or you download branch as zip file, in this cases package archive doesn't contain `.git` folder, so Linux distributions which are downloading this package this way eg for preparing this package for distro repositories, will not be affected to accidentally compile debug build.

I used code from this blog post https://blog.kitware.com/cmake-and-the-default-build-type/ where is everything described in detail.

I have tried shadow builds, builds with and without `.git` folder and of course with and without `-DCMAKE_BUILD_TYPE=` option and it works as expected.

Fixes #4416

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,8 @@ project(libtorrent
 )
 set (SOVERSION "10")
 
+set_default_build_type()
+
 include(GNUInstallDirs)
 include(GeneratePkgConfig)
 
--- a/cmake/Modules/LibtorrentMacros.cmake
+++ b/cmake/Modules/LibtorrentMacros.cmake
@@ -107,7 +107,7 @@ function(read_version _verFile _outVarMajor _outVarMinor _outVarTiny)
 	# the verFileContents variable contains something like the following:
 	# #define LIBTORRENT_VERSION_MAJOR 1;#define LIBTORRENT_VERSION_MINOR 2;#define LIBTORRENT_VERSION_TINY 0
 	set(_regex ".+_MAJOR +([0-9]+);.+_MINOR +([0-9]+);.+_TINY +([0-9]+)")
-	 # note quotes around _regex, they are needed because the variable contains semicolons
+	# note quotes around _regex, they are needed because the variable contains semicolons
 	string(REGEX MATCH "${_regex}" _tmp "${verFileContents}")
 	if (NOT _tmp)
 		message(FATAL_ERROR "Could not detect project version number from ${_verFile}")
@@ -118,4 +118,21 @@ function(read_version _verFile _outVarMajor _outVarMinor _outVarTiny)
 	set(${_outVarMajor} ${CMAKE_MATCH_1} PARENT_SCOPE)
 	set(${_outVarMinor} ${CMAKE_MATCH_2} PARENT_SCOPE)
 	set(${_outVarTiny} ${CMAKE_MATCH_3} PARENT_SCOPE)
+endfunction()
+
+# Set a default build type if none was specified
+function(set_default_build_type)
+	set(default_build_type "Release")
+	if(EXISTS "${CMAKE_SOURCE_DIR}/.git")
+		set(default_build_type "Debug")
+	endif()
+
+	if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+		message(STATUS "Setting build type to '${default_build_type}' as none was specified")
+		set(CMAKE_BUILD_TYPE "${default_build_type}" CACHE STRING
+			"Choose the type of build, options are: Debug Release RelWithDebInfo MinSizeRel ..." FORCE)
+		# Set the possible values of build type for cmake-gui
+		set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
+			"Debug" "Release" "MinSizeRel" "RelWithDebInfo")
+	endif()
 endfunction()


arvidn:
it seems reasonable to not set the build type when `CMAKE_CONFIGURATION_TYPES` is set (I think). Doesn't that still mean that the generated `.pc` file will be incorrect for such generators?

If so, should the logic that generates the `.pc` file have a whitelist of generators and fail if it's not supported? Perhaps more reasonable, the function generating the `.pc` file should fail if `CMAKE_BUILD_TYPE` is not set.

I'm not very experienced with cmake, so I may be talking nonsense.

arvidn:
@zeule what are your thoughts?

Chocobo1:
>If I don't pass -DCMAKE_BUILD_TYPE=Release|Debug to cmake build command, I don't have -lboost_system in generated libtorrent-rasterbar.pc file.

From this statement, I would expect `-lboost_system` to be generated regardless of what value `CMAKE_BUILD_TYPE` has. And while I think having a smart default value for `CMAKE_BUILD_TYPE` could be a good idea on its own, sadly I don't think it is solving the linked issue directly.
https://cmake.org/cmake/help/v3.17/variable/CMAKE_BUILD_TYPE.html#variable:CMAKE_BUILD_TYPE
Notice that it says an empty value is also a possibility.

zeule:
> @zeule what are your thoughts?

Can't reproduce #4416 on my system.

I doubt these changes will work well with the multi-configuration generators (i.e. MSVC). Maybe we better try to fix the pkg-config generator?

silverqx:
Today I looked again on this and I understood multi-config, it's great and I can tell, that I will use multi-config ( currently only on Windows ) instead of single-config generators.

I try to answer some questions, which was asked.

> it seems reasonable to not set the build type when `CMAKE_CONFIGURATION_TYPES` is set (I think). Doesn't that still mean that the generated `.pc` file will be incorrect for such generators?
> 

No, because pkgconf isn't generated for MSVC toolchain, look [here](https://github.com/arvidn/libtorrent/blob/RC_1_2/CMakeLists.txt#L741) and multi-config is only supported for visual studio generators, but from cmake 3.17 multi-config will be supported for [ninja generators](https://blog.kitware.com/multi-config-ninja-generator-in-cmake-3-17/) too, after 3.17 release I can test it.

> If so, should the logic that generates the `.pc` file have a whitelist of generators and fail if it's not supported? Perhaps more reasonable, the function generating the `.pc` file should fail if `CMAKE_BUILD_TYPE` is not set.
> 

I don't like to fail when `CMAKE_BUILD_TYPE` is not set, when we can easily recover from this, better than fail ( from my view ) is to offer the user some meaningful default for this variable.

pkg-config generator works fine, but he needs to have `CMAKE_BUILD_TYPE` to function properly, the problem occurs exactly [here](https://github.com/arvidn/libtorrent/blob/RC_1_2/cmake/Modules/GeneratePkgConfig.cmake#L34) and [here](https://github.com/arvidn/libtorrent/blob/RC_1_2/cmake/Modules/GeneratePkgConfig.cmake#L39) and is only relevant for single-config generators, multi-config generators are not affected by this problem, because `CMAKE_CONFIGURATION_TYPES` variable has reasonable defaults, is always prepopulated with a default value if CMake is executed with a multi-config generator.
Because of this behavior, you can check if CMake uses multi-config generator like this:
```
if(CMAKE_CONFIGURATION_TYPES)
    message("|| Multi-config")
endif()
```

arvidn:
> > Doesn't that still mean that the generated .pc file will be incorrect for such generators?
>
> No, because pkgconf isn't generated for MSVC toolchain, look here and multi-config is only supported for visual studio generators, but from cmake 3.17 multi-config will be supported for ninja generators too, after 3.17 release I can test it.

So, it sounds like it's still a problem then. Whether .pc files are relevant for MSVC or not doesn't really seem relevant. That's not the only multi-config generator. Also, are you asserting that a .pc file not being generated is correct for every possible multi-config generator? It doesn't seem intuitive for linux, where .pc files are actually used, and where people expect ABI compatibility across compilers.

On the other hand, it doesn't make sense for a .pc file to be generated for a "generic" build, before knowing which configuration was chosen for the build

silverqx:
> Also, are you asserting that a .pc file not being generated is correct for every possible multi-config generator? It doesn't seem intuitive for linux, where .pc files are actually used, and where people expect ABI compatibility across compilers.

I don't exactly understand what you mean, but multi-config generators don't have this problem with missing `-lboost_system`.

> On the other hand, it doesn't make sense for a .pc file to be generated for a "generic" build, before knowing which configuration was chosen for the build

It's all about how are you looking at it, you can use .pc file even when `-DCMAKE_BUILD_TYPE=` is empty, generally `-DCMAKE_BUILD_TYPE=` means build without debug info and without optimizations.

silverqx:
I think that `-DCMAKE_BUILD_TYPE=` should be initialized by default, like in autotools build, `none` value isn't a good idea in general, it is confusing.

arvidn:
> It's all about how are you looking at it, you can use .pc file even when -DCMAKE_BUILD_TYPE= is empty, generally -DCMAKE_BUILD_TYPE= means build without debug info and without optimizations.

The .pc file is a poor-man's extension of a build system, in a half-assed attempt to make libraries ABI compatible. If the .pc file has a configuration different from the actual binary, chances are there will be ODR violations and memory corruption.

> I think that -DCMAKE_BUILD_TYPE= should be initialized by default, like in autotools build, none value isn't a good idea in general, it is confusing and especially for newcomers.

I agree that it's weird that there isn't default for this. But I also don't fully understand a lot of the logic behind cmake's design decisions. I suppose it works like this deliberately, and that cmake users would expect it to work like this. Is it better to make it behave better than to make it behave as people expect?

silverqx:
> > I think that -DCMAKE_BUILD_TYPE= should be initialized by default, like in autotools build, none value isn't a good idea in general, it is confusing and especially for newcomers.
> 
> I agree that it's weird that there isn't default for this. But I also don't fully understand a lot of the logic behind cmake's design decisions. I suppose it works like this deliberately, and that cmake users would expect it to work like this. Is it better to make it behave better than to make it behave as people expect?

Ohh, I understand what you mean, but there is not any special need for none configuration, I never need it, you never need it and I compile apps few years:), of course, there is some mechanic or logic behind how this configuration works and how are CMake variables connected to every configuration.

Anyway, can I close this PR? You want none configuration to stay?

arvidn:
so, to be clear. Regardless of this being applicable to solving the .pc file generation, it seems reasonable to have a default build type. But since I'm not really familiar with the principles behind cmake, I'd like to hear from someone that is. I can see it causing trouble too.

For example, if I a client links against libtorrent, and that client is built without specifying the build type, if then one of its dependencies (libtorrent) decides to build with debug or release, you may have an ABI issue in the final link.

arvidn:
actually, the more I think about it, it would probably make more sense to *fail* the build if no build type is specified (if there's any value in that). We must assume that libtorrent is not at the top of the build tree, and it must be built with whatever configuration the top level decides

silverqx:
But every configuration should be ABI compatible.

silverqx:
eg it doesn't matter if I link against debug or release build of openssl, it works everytime, the same should be true for every other library, ABI compatibility is not based on configuration, but by SOVERSION.

arvidn:
> eg it doesn't matter if I link against debug or release build of openssl, it works everytime,

perhaps openssl makes this guarantee.

> the same should be true for every other library,

you mean you *wish* this to be true for other libraries

> ABI compatibility is not based on configuration, but by SOVERSION.

SOVERSION is one blunt tool to indicate when ABI breaks happen deliberately. It doesn't encode ABI breaks for other, more subtle, reasons. Perhaps the best example is that in debug mode STL iterators may be debug iterators, not ABI compatible with the release iterators (which could be raw pointers sometimes). This would make Debug and Release builds not ABI compatible.

But there are essentially an infinite number of knobs to turn on compilers that break ABI compatibility.

arvidn:
> Ohh, I understand what you mean, but there is not any special need for none configuration, I never need it, you never need it and I compile apps few years:), of course, there is some mechanic or logic behind how this configuration works and how are CMake variables connected to every configuration.

I agree that it's very strange that cmake even have this concept of no build type. But I don't know enough about cmake's rationale to say whether it should be addressed in cmake files or not.

> Anyway, can I close this PR? You want none configuration to stay?

I would like to hear from more experienced cmake people, but I think now I'm leaning towards not landing this PR.

silverqx:
The none configuration connects to CMAKE_C_FLAGS, Debug configuration connects to CMAKE_C_FLAGS_DEBUG and cmake is responsible to copy values to CMAKE_C_FLAGS, that is how none configuration works, I don't think that there was any meaningful decision for none configuration, I think that because of this mechanic none configuration exists, but I never needed this none configuration.

silverqx:
And none configuration is set like, no debug symbols and no compiler optimizations, there will not be more to say about none configuration.

silverqx:
Ok, you can close this, if you want this none configuration, thx

silverqx:
@arvidn can I close this? @zeule fixed all problems at #4456

arvidn:
sure

PR #4422, opened 2020-03-15
Disable bittyrant

arvidn:


arvidn:
it turns out the bittyrant choker actually doesn't work anymore either. I removed the support for setting per-connection rate limits a long time ago, not realizing that bittyrant depends on that. I'm tempted deprecate and remove the bittyrant choker entirely in the next major release.

arvidn:
in fact, maybe deprecating is better than this patch

arvidn:
superceded by https://github.com/arvidn/libtorrent/pull/4423

PR #4423, opened 2020-03-15
deprecate bittyrant choking algorithm

arvidn:
the bittyrant choker relies on being able to set upload rate limits *per connection*. This feature was removed a long time ago in an attempt to simplify things. The bittyrant dependency on this feature was overlooked and effectively broke this choking algorithm. Rather than fixing it, I'm tempted to remove it. If there is interest (and effort) to research choking algorithms, I'm open to add something back in the future.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* deprecate the bittyrant choking algorithm
 	* add build option to disable streaming
 
 1.2.5 release
--- a/bindings/python/src/alert.cpp
+++ b/bindings/python/src/alert.cpp
@@ -842,7 +842,9 @@ void bind_alert()
         .value("download_limit_too_low", performance_alert::download_limit_too_low)
         .value("send_buffer_watermark_too_low", performance_alert::send_buffer_watermark_too_low)
         .value("too_many_optimistic_unchoke_slots", performance_alert::too_many_optimistic_unchoke_slots)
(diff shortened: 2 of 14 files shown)


FranciscoPombal:
In https://github.com/arvidn/libtorrent/pull/4422 you mention per-connection rate limits, but in the documentation (https://libtorrent.org/single-page-ref.html#choking_algorithm) it says:

> For this choker to be efficient, you need to set a global upload rate limit (`settings_pack::upload_rate_limit`).

which makes it seem like only global limits are needed to make it work. So I think the docs should be updated, even if this choker is being removed soon.

arvidn:
The global upload rate limit is referring to the setting. My explanation of why this choker doesn’t work refers to the feature itself. There is no way to set the upload rate limit per connection anymore. It’s an implementation detail.

FranciscoPombal:
I meant the docs make it sound like the feature works as long as there is a global upload rate limit set - and it won't, because what is needed is an upload rate limit per connection, which does not exist anymore, as you say.

arvidn:
I see what you mean. This patch will remove the mention of the bittyrant choker from the documentation though

FranciscoPombal:
> I see what you mean. This patch will remove the mention of the bittyrant choker from the documentation though

Hadn't noticed that - that works as well :)

PR #4439, opened 2020-03-19
adding BOOST_NO_EXCEPTIONS around file_storage add files

aldenml:


arvidn:
are you actually building libtorrent with `-fno-exceptions`? does it work?

aldenml:
not really, I just use it as a safeguard for not use the API that throws while using the SWIG interface. Do you want to remove these checks?

arvidn:
well, I've been tempted to either remove them or fix support for building without exceptions (where "support" means terminating)

aldenml:
iirc, the problem is not only in `libtorrent` but in `boost` also, meaning that even using `BOOST_NO_EXCEPTIONS`, boost requires exceptions support due to the use of some related keywords, but I haven't tried it recently.

include/libtorrent/file_storage.hpp
```diff
@@ -306,6 +307,7 @@ namespace aux {
 			, file_flags_t file_flags = {}
 			, std::time_t mtime = 0, string_view symlink_path = string_view()
 			, char const* root_hash = nullptr);
+#endif
```
arvidn: you would have to surround the definitions of these functions with this preprocessor condition too, right?

arvidn: otherwise I imagine you would get a compile error

aldenml: that's right!

aldenml: done

PR #4440, opened 2020-03-19
Option to disable super seeding

arvidn:


FranciscoPombal:
In this comment (https://github.com/qbittorrent/qBittorrent/pull/12162#discussion_r394967414) I said, most likely erroneously, that the non-strict super seeding could be enabled with a `torrent_flags_t` flag. Now I realize that most likely, that flag is the strict flag, that is enabled by the corresponding setting. If this is indeed the case, please mention that in the flag's documentation (even if this is all now deprecated). Otherwise, it leads to confusion because the flag's name does not include the word "strict", and the setting name does.

arvidn:
no, strict super seeding is controlled by a session-wide [setting](https://libtorrent.org/reference-Settings.html#strict_super_seeding), and is disabled by default

FranciscoPombal:
> no, strict super seeding is controlled by a session-wide [setting](https://libtorrent.org/reference-Settings.html#strict_super_seeding), and is disabled by default

So the `super_seeding` flag in `torrent_flags_t` really is different from the strict super seeding setting? In that case, why is the `super_seeding` flag not controllable by a setting as well?

arvidn:
the `super_seeding` flag on the torrent indicates whether that torrent is being super seeded or not.

The `strict_super_seeding` flag on the session defines what it means for a torrent to be super seeding

FranciscoPombal:
> the `super_seeding` flag on the torrent indicates whether that torrent is being super seeded or not.
> 
> The `strict_super_seeding` flag on the session defines what it means for a torrent to be super seeding

Right, the latter is being deprecated, but there is no way to configure the former independently.

arvidn:
> Right, the latter is being deprecated, but there is no way to configure the former independently.

Independent from what? You mean have one torrent strict super seeding and another relaxed super seeding? No, that's not possible.

FranciscoPombal:
> Independent from what? You mean have one torrent strict super seeding and another relaxed super seeding? No, that's not possible.

I did not know it was all or nothing. Why is that?

arvidn:
> I did not know it was all or nothing. Why is that?

All or nothing what?

FranciscoPombal:
> All or nothing what?

Super seeding. I though you could have super seeding without being strict.

arvidn:
> Super seeding. I though you could have super seeding without being strict.

Right, that's the default. https://github.com/arvidn/libtorrent/pull/4440#issuecomment-601490415

FranciscoPombal:
> > Super seeding. I though you could have super seeding without being strict.
> 
> Right, that's the default. [#4440 (comment)](https://github.com/arvidn/libtorrent/pull/4440#issuecomment-601490415)

Ok, but even if you have the strict super seeding disabled via that option, you can't turn off the regular one (at runtime).

arvidn:
> Ok, but even if you have the strict super seeding disabled via that option, you can't turn off the regular one (at runtime).

A torrent that doesn't have the `super_seeding` flag set, is not super seeding. A torrent that has the flag set is. Whether the torrents that are super seeding or not use the strict definition or not cannot be configured per-torrent, but only per session, via the `strict_super_seeding` *setting*. This setting is being deprecated and defaults to off.

arvidn:
both the torrent flag and the setting are changed at runtime. Or are you talking about the new *build* option to disable the super seeding feature altogether?

FranciscoPombal:
I am talking about the torrent `super_seeding` flag at the torrent level, not the strict mode which is enabled by the setting. How do you enable/disable the `super_seeding` flag at the torrent level? There is no setting for that.

arvidn:
before adding the torrent: https://libtorrent.org/reference-Core.html#add_torrent_params (`add_torrent_params::flags`)
after adding the torrent: https://libtorrent.org/reference-Core.html#set_flags

FranciscoPombal:
> before adding the torrent: https://libtorrent.org/reference-Core.html#add_torrent_params (`add_torrent_params::flags`)
> after adding the torrent: https://libtorrent.org/reference-Core.html#set_flags

Right, so this can be done even if strict mode is disabled. However, this PR introduces a build option to disable super seeding completely (both strict and non strict), right? So there is no build-time option to disabled _just_ strict super seeding.

arvidn:
> However, this PR introduces a build option to disable super seeding completely (both strict and non strict), right?

That's right.

> So there is no build-time option to disabled just strict super seeding.

Correct

FranciscoPombal:
Ok thanks for the clarification, I understand now.

But please add a note in the documentation that the default build-time option is to support super seeding (i.e. with b2 the `super-seeding` build feature defaults to `on` and in the CMake build `TORRENT_DISABLE_SUPERSEEDING` is not defined by default).

In fact I would suggest you update the docs to be more consistent in this regard. Some build features explicitly mention what the default is and some don't.

Seeker2:
On public trackers, newly-made torrents are occasionally seeded using qBitTorrent's super seeding.
This is because it's very handy for creating new seeds and making sequential downloading and streaming possible earlier in a torrent's life-cycle.

It would also be helpful if non-strict super seeding automatically enabled and disabled (if manually set as enabled):
Enabled when connected peers are >9 and piece availability between connected peers is <0.5 (not counting the seed.)
Disabled if connected peers falls below 5 for more than 10 minutes or if availability is >1 between all the connected peers.

I am not suggesting making non-strict super seeding default behavior or even globally enabled.

Please don't remove the non-strict implementation from the regular libtorrent code base.

arvidn:
> Disabled if connected peers falls below 5 for more than 10 minutes or if availability is >1 between all the connected peers.

I think I was considering disabling `super_seeding` mode as soon as another seed was observed in the swarm. It doesn't look that that's done now though.

> Please don't remove the non-strict implementation from the regular libtorrent code base.

This patch is only preparing to remove the *strict mode*. The relaxed super seeding is still in. There's just a build option to not have that feature, just like there's a build option to disable SSL and other extensions. They are still built-in by default.

Seeker2:
I have a lot of 5+ year old torrents that I reseed from time-to-time, far too many to effective seed at once.
The really problematic torrents among those have multiple peers stuck on exactly the same percentage.
I've seen torrent trackers reporting all kinds of numbers for seeds+peers that weren't really there, or there was a slow+firewalled seed that only showed up briefly.

That's why I think checking reported availability among connected peers (assuming there's at least 5) is a good idea.

Seeker2:
There is also the problem of qBitTorrent (possibly caused by something in libtorrent) misreporting peers as seeds:
Incorrect connected seeds/peers numbers being shown  https://github.com/qbittorrent/qBittorrent/issues/12575

PR #4446, opened 2020-03-21
fix -Wmissing-braces warnings for older Apple LLVM version 9.1.0 (default in travis)

aldenml:


arvidn:
I think these warnings is a bug in LLVM, that was fixed in later versions

aldenml:
yes, that's the case, but the `{{{` is already in some places, and others might want to still compile with older versions. If you rather not merge this, I'm ok, just let me know :)

arvidn:
I'm a bit hesitant. I seem to recall that some compilers warn on the redundant pair of braces too. But if you say there are triple-braces already, I suppose not.

aldenml:
Yes, a few `{{{` (added  by me a while ago :)

https://github.com/arvidn/libtorrent/blob/92c1418992f6326ea96b17f029c7953f415d6d39/src/kademlia/routing_table.cpp#L275

https://github.com/aldenml/libtorrent/blob/ef81e5814e14bdd97a1eeae44dde392a01527c4c/src/escape_string.cpp#L321

PR #4452, opened 2020-03-22
moving bandwidth related types and files to aux

aldenml:


arvidn:
this doesn't look related to your patch.. https://travis-ci.org/github/arvidn/libtorrent/jobs/665640384#L1312

PR #4455, opened 2020-03-23
moving resolver related types and files to aux

aldenml:


aldenml:
I wonder what's happening to travis status here

arvidn:
yeah, since their maintenance they seem to have broken a few things

PR #4456, opened 2020-03-23
Fix PkgConfig generation on Arch

### LINKED ISSUE #4416: pkgconfig file doesn't contain boost_system lib with cmake

zeule:
Use first available configuration from the IMPORTED_CONFIGURATIONS list
if no valid property value was found via current or mapped
configurations.

Fixes #4416.


cmake/Modules/GeneratePkgConfig.cmake
```diff
@@ -28,28 +28,35 @@ function(_compile_features_to_gcc_flags _res _features)
 endfunction()
 
 function(_get_target_property_merging_configs _var_name _target_name _propert_name)
-	get_target_property(vals ${_target_name} ${_propert_name})
+	get_property(vals TARGET ${_target_name} PROPERTY ${_propert_name})
 	if (NOT vals)
-		set(vals "")
```
zeule: Unneeded because `get_property()` returns empty strings if property does not exist (unlike `get_target_property()`, which returns "-NOTFOUND").

silverqx: Empty string is what we needed, so its ok

zeule: Perhaps a cleaner way would be to use `get_property(vals TARGET ${_target_name} PROPERTY ${_propert_name} SET)` to test if there is a value?

zeule:
> ... by the way, you changed case style for variables to camelCase.

Oh, will fix. Thanks!

cmake/Modules/GeneratePkgConfig.cmake
silverqx: You are right, `SET` is ideal in this case, because all other code dependent on this `vals` variable is closed in `if` statements.

silverqx:
I tested it and it works like charm, it solves all problems, even problem with empty / none `CMAKE_BUILD_TYPE`. But I posted some review and it lost? Maybe I forgot to post it or something, I'm gona to review it again.

cmake/Modules/GeneratePkgConfig.cmake
```diff
@@ -28,27 +28,39 @@ function(_compile_features_to_gcc_flags _res _features)
...
+			if (prop_set)
+				get_property(val_for_cfg TARGET ${_target_name} PROPERTY ${_propert_name}_${target_cfg})
 				list(APPEND vals "$<$<CONFIG:${cfg}>:${val_for_cfg}>")
+				break()
 			endif()
 		endforeach()
+		if (NOT prop_set)
+			get_property(imported_cfgs TARGET ${_target_name} PROPERTY IMPORTED_CONFIGURATIONS)
+			# CMake docs say we can use any of the imported configs
+			list(GET imported_cfgs 0 imported_config)
```
silverqx: It would not be better to prefer `Release` configuration, if it exists in  `IMPORTED_CONFIGURATIONS`? Something like, if `Release` exists in `imported_cfgs`, then use release, otherwise use an item at 0 index.
I'm thinking about this, like everytime is better to use release build, if there will be more values, like RelWithDebInfo or MinSizeRel or something custom.
Or prefer best choice and if not exists, than use 0 index.

silverqx: Or that's too much :)

zeule: I should probably check what does CMake itself do and copy its logic.

zeule: CMake takes the first available: https://gitlab.kitware.com/cmake/cmake/-/blob/master/Source/cmTarget.cxx#L2107

silverqx: cmake doesn't prioritize Release config, it only populates `IMPORTED_CONFIGURATIONS` with all available configurations, so example, if at 0 index will be MinSizeRel configuration and at 1 index will be Release configuration, so of course MinSizeRel config will be picked now

silverqx: I meant something like this:
```
list(FIND imported_cfgs "RELEASE" found)
if(found)
  set(imported_cfg "RELEASE")
else()
  list(GET imported_cfgs 0 imported_config)
endif()
```

silverqx: but if not, its ok, i'm happy with this PR, it solves all problems which we found 👌

zeule: Sorry, don't really understand you. I replied to you that replicating the internal CMake behavior is the best approach, as far as I cab see. From the CMake sources (linked above) I conclude it chooses the first available configuration, and so should do we. Why do you want to prioritize the release configuration? The .pc file may not differ from the Targets.cmake file, and therefore we have to replicate the internal CMake logic here.

silverqx: You are right, replicate internal cmake logic is good idea anyway, I agree, there is no reason to change this logic

zeule: OK, thanks.

arvidn:
thanks guys!

arvidn:
should this go into `RC_1_2` btw?

zeule:
Is there a reason why it shouldn't?

arvidn:
I imagine this problem exists there too. If it applies to `RC_1_2`, I think it should go in there instead. I'll merged it into master later

zeule:
#4457

arvidn:
thanks! I assume the patch against `RC_1_2` I just merged is the same as this, and that I can merge `RC_1_2` into master instead of landing this PR. Is that right?

I will want to merge `RC_1_2` into master anyway, to pick up other patches, so it would be simpler to get this patch from there as well, rather than sorting out a conflict.

silverqx:
yep, its ok, thx

zeule:
@arvidn, yes, the second PR contains this commit, cherry-picked to the RC_1_2 branch.

PR #4458, opened 2020-03-24
moved internal disk types and files to aux, minor cleanup

aldenml:


include/libtorrent/disk_observer.hpp
```diff
@@ -37,7 +37,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent {
 
-	struct TORRENT_EXTRA_EXPORT disk_observer
+	struct TORRENT_EXPORT disk_observer
```
arvidn: is this really part of the public API? I don't think so, it's just used for peers to be notified when under back-pressure from the disk threads, isn't it?

arvidn:
this seems like a reasonable direction to go. I've also been thinking about taking this to the next level, and maybe keep things a bit cleaner in the process. By:

* creating a clean separation between public headers and private headers. I'm not sure exactly where private headers would go, but some other directory that isn't reachable by clients (unless export-extra is enabled, for tests).

* only internal headers that need to be reachable from public headers would need to use the `aux` namespace in that case. Which I suppose would be somewhat common during a transition period.

* a lot of the classes in this PR really belong to the `mmap_disk_io` class. I've been thinking about whether it makes sense to organize them in a separate directory too, and maybe even its own sub-namespace.

I don't have a clear idea of what this future would look like. Do you have any thoughts?

include/libtorrent/disk_observer.hpp
aldenml: I initially moved it to `aux_/`, but then I found that it's used as part of the public API in `disk_interface`, at https://github.com/arvidn/libtorrent/blob/master/include/libtorrent/disk_interface.hpp#L176

Not sure how do you want to handle this.

aldenml:
I don't have a clear idea either, but I know it's a very laborious task :)

What I'm thinking is slowly moving what's "internal" to `aux_/` and free the public headers from those headers as much as possible. Then from there is easier to see what's strictly private and what needs to be exposed anyway.

include/libtorrent/disk_observer.hpp
arvidn: that's a good point. This will make this type show up in the documentation too.

PR #4461, opened 2020-03-25
fix compilation for mingw under linux

aldenml:


include/libtorrent/aux_/windows.hpp
```diff
@@ -42,7 +42,12 @@ POSSIBILITY OF SUCH DAMAGE.
 #ifndef STRICT
 #define STRICT
 #endif
+
+#if defined __MINGW32__ || defined __MINGW64__
```
arvidn: are these not defined on windows?

aldenml: I imagine, but I don't think the lower case is a problem in windows, is it?

arvidn: clang (on windows) warns about it. I think because NTFS *can* be formatted to be case sensitive.

arvidn: I could imagine another reason would be if one cross compiles for windows under windows subsystem for linux, maybe the system headers are used from main windows but case sensitive. I don't know that that's a problem.

In the future this could be solved with `__has_include <Windows.h>`

aldenml: maybe I will just hack my build setup, this is a very particular situation

arvidn: I can see two approaches:

1. Consider this a bug in mingw's cross compilation environment, because the windows headers don't have the correct names.

2. add some additional logic here to detect when compiling on linux (or I suppose any non-windows host)

I'm actually a little bit inclined to lean towards (1). But I'm also open to supporting it properly in C++17 with `__has_include`

aldenml: Ok, I decided to tweak my build env, closing this

PR #4464, opened 2020-03-26
fixed -Wextra-semi-stmt warning

aldenml:


arvidn:
huh. I wonder why I don't get these warnings. Are they not part of `-Wall -Wextra`?

aldenml:
that warning happens only with the most recent version of the Android NDK clang.

PR #4465, opened 2020-03-26
moved alert_manager to aux

aldenml:


include/libtorrent/aux_/session_impl.hpp
```diff
@@ -902,7 +902,7 @@ namespace aux {
 #endif
 
 			// handles delayed alerts
-			mutable alert_manager m_alerts;
+			mutable aux::alert_manager m_alerts;
```
arvidn: these two changes shouldn't be necessary. This is already within the `aux` namespace.

include/libtorrent/aux_/session_interface.hpp
```diff
@@ -158,7 +158,7 @@ namespace aux {
 
 		virtual disk_interface& disk_thread() = 0;
 
-		virtual alert_manager& alerts() = 0;
+		virtual aux::alert_manager& alerts() = 0;
```
arvidn: Is this namespace qualification really necessary?

aldenml:
@arvidn changes and rebased

PR #4466, opened 2020-03-26
fix ssl_version configuration issue in session constructor

arvidn:
introduce a generic tls version to use as default

### DIFF
--- a/include/libtorrent/aux_/openssl.hpp
+++ b/include/libtorrent/aux_/openssl.hpp
@@ -101,9 +101,6 @@ TORRENT_EXTRA_EXPORT GENERAL_NAME* openssl_general_name_value(GENERAL_NAMES* gen
 
 #endif // OPENSSL_VERSION_NUMBER
 
-// converts setting_pack::ssl_version_t enum into asio version
-ssl::context::method ssl_version(int const v);
-
 } // aux
 } // libtorrent
 
--- a/include/libtorrent/config.hpp
+++ b/include/libtorrent/config.hpp
@@ -568,14 +568,6 @@ constexpr std::size_t TORRENT_WRITE_HANDLER_MAX_SIZE = 342;
 #endif
 #endif // TORRENT_HAS_ARM_CRC32
 
-#ifndef TORRENT_USE_TLS13
-#if BOOST_VERSION >= 106900
-#define TORRENT_USE_TLS13 1
-#else
-#define TORRENT_USE_TLS13 0
-#endif
-#endif
-
 namespace libtorrent {}
 
 // create alias
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -1752,12 +1752,6 @@ namespace aux {
 			// lower than 5 minutes.
 			upnp_lease_duration,
 
-			// the SSL/TLS version to use for HTTPS trackers and SSL torrents. Set
-			// it to one of the ssl_version_t values. This setting only takes
-			// effect when passed in to the session constructor. It cannot be
-			// changed once the session has been constructed.
-			ssl_version,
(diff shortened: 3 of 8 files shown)


FranciscoPombal:
What does this "generic tls version" default to? 1.2?

arvidn:
it's a bit unclear in the boost.asio documentation. The answer can probably be found in the openssl docs. But I suspect it means *any* TLS version.

FranciscoPombal:
> it's a bit unclear in the boost.asio documentation. The answer can probably be found in the openssl docs. But I suspect it means _any_ TLS version.

Isn't there an alternative to doing things this way then? If there isn't, please make it abundantly clear i n the docs that libtorrent does not make any guarantees about the TLS version used when `tls` is specified and that people should double check what version will actually used, otherwise it could be a security risk for users.

arvidn:
actually, reading about this ([here](https://www.openssl.org/docs/manmaster/man3/SSL_CTX_new.html)). It seems the version specific "methods" are deprecated, and that maybe this shouldn't be configurable at all..

arvidn:
I think this is the right approach. I'm undoing the `ssl_version` setting and defer to OpenSSL to do the right thing. If anyone can think of a good reason *not* to apply this patch, please let me know.

FranciscoPombal:
> I think this is the right approach. I'm undoing the `ssl_version` setting and defer to OpenSSL to do the right thing. If anyone can think of a good reason _not_ to apply this patch, please let me know.

Seems reasonable, but I sill think you should explicitly disable TLS <1.2. This is to prevent TLS < 1.2 from being auto selected if the OpenSSL version used is old enough to default to TLS < 1.2, but recent enough to have TLS 1.2 support. TLS 1.2 support was added in OpenSSL 1.0.1, released back in 2012, so I don't think you need to care about OpenSSL versions that don't support TLS 1.2.

arvidn:
> Seems reasonable, but I sill think you should explicitly disable TLS <1.2.

I don't know of a way to do that. OpenSSL doesn't even have a deprecated API to do this, as far as I can tell. I assume the reason for this is for OpenSSL itself to have full control of disabling any insecure features, without giving the option to turn them back on again.

rwasef1830:
@arvidn The latest version of this pull request fixes the tracker connection problem I described to you in email. Thanks!

FranciscoPombal:
@arvidn 
What about this?
https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_options.html

or this (the `no-*` options):
https://www.boost.org/doc/libs/1_72_0/doc/html/boost_asio/reference/ssl__context.html

> OpenSSL itself to have full control of disabling any insecure features, without giving the option to turn them back on again.

This only makes sense on OpenSSL's part if they always default to the latest supported version. Not sure if this is the case or not, I know nothing about their API.

NotTsunami:
The macro in `include/libtorrent/config.hpp` should be removed as well.

arvidn:
@FranciscoPombal it seems reasonable to disable old versions of TLS. However:

* I'm not an authority on this so deferring this to what OpenSSL does by default seems reasonable.
* The steps of authenticating trackers *at all* is already the most significant step, so I think this is sufficient for one PR. Tuning of which TLS versions to support could always be added later, but I also have faith in OpenSSL disabling insecure features by default, as they come up.

FranciscoPombal:
I suppose you're right. Now that I think about it, a better way of achieving essentially the same result is to bump the dependency version requirement of OpenSSL. For example, in this case you could force libtorrent to require that the dependency on OpenSSL be `OpenSSL >= 1.0.1`. As protocols become insecure in the future, you just bump this version requirement accordingly, to a version of OpenSSL that doesn't default to them or disables them entirely.

PR #4467, opened 2020-03-26
Fix [-Wsign-compare] warning

zywo:


src/platform_util.cpp
```diff
@@ -74,7 +74,7 @@ namespace libtorrent {
 			if (rl.rlim_cur == rlim_infinity)
 				return std::numeric_limits<int>::max();
 
-			return rl.rlim_cur <= std::numeric_limits<int>::max()
+			return rl.rlim_cur <= std::numeric_limits<unsigned long long>::max()
```
arvidn: `rlim_t` is not guaranteed to be an `unsigned long long`.

also, this changes the semantics of the code, to be incorrect. The reason this is compared against the largest value an `int` can hold is because this function returns an `int`. If anything, you can *cast* the result to `rlim_t`.

The warning you get on this line is a false positive, and I think only old compilers warn on this. The right hand side in this comparison is a *constant*. The compiler can tell that this constant fits in `rlim_t` without losing precision.

zywo: The compiler version is gcc 9.3.0, and here's the warning:

```
/home/zywo/libtorrent/src/platform_util.cpp: In function ‘int libtorrent::max_open_files()’:
/home/zywo/libtorrent/src/platform_util.cpp:77:23: warning: comparison of integer expressions of different signedness: ‘rlim_t’ {aka ‘long unsigned int’} and ‘int’ [-Wsign-compare]
   77 |    return rl.rlim_cur <= std::numeric_limits<int>::max()
      |           ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

arvidn: ok, so it's a false positive in GCC then. I assume GCC wouldn't warn if the max constant was changed to `INT_MAX`. This patch still breaks the code though.

zywo: By using `INT_MAX` and including `climits` header, there's no warning.

If the change is useless, we can close this PR.

arvidn: the *current* patch is wrong. But you could force push to it to make it compare against the correct value again, and also fix the warning

arvidn:
looking into this a bit, it seems like a GCC9 bug: https://www.mail-archive.com/gcc-bugs@gcc.gnu.org/msg615033.html

zywo:
We can wait for gcc fix.

src/platform_util.cpp
```diff
@@ -74,8 +74,8 @@ namespace libtorrent {
 			if (rl.rlim_cur == rlim_infinity)
 				return std::numeric_limits<int>::max();
 
-			return rl.rlim_cur <= std::numeric_limits<int>::max()
-				? int(rl.rlim_cur) : std::numeric_limits<int>::max();
+			return static_cast<int>(rl.rlim_cur) <= std::numeric_limits<int>::max()
```
arvidn: the reason this comparison is here is because `rl.rlim_cur` may be greater than the largest value that can be represented by `int`. If it is, this function just returns that largest possible value. Since this function returns an `int`, it cannot return the true (larger) number in that case.

By casting the `rl.rlim_cur`, you effectively truncate that value *first*, making the test meaningless. to make the test meaningful, you'd have to cast the *right* hand side to `rlim_t`.

zywo:
Forgive me, I'm not experienced in programming.

arvidn:
no problem. thanks for the patch!

PR #4475, opened 2020-03-28
update Jamfiles to default to C++11

arvidn:


Jamfile
```diff
@@ -272,6 +272,13 @@ rule building ( properties * )
 		result += <build>no ;
 	}
 
+	local VERSION = [ feature.get-values <cxxstd> : $(properties) ] ;
+	if ! $(VERSION) || $(VERSION) < 11
+	{
+		ECHO "libtorrent requires at least C++14. Specify <cxxstd>11 or higher" ;
```
aldenml: typo here `C++11`

arvidn: oops :). I'm not sure the boost version is new enough for this also. this is mostly an experiment

arvidn:
it's probably not worth the hassle of bringing all the CI up to a new version of boost for this

PR #4484, opened 2020-03-31
make sure attempting to bind to an unknown device fails (on MacOS/BSD)

arvidn:


FranciscoPombal:
Is this related to https://github.com/qbittorrent/qBittorrent/issues/12253#issuecomment-605922645?

If so, has this feature been working on Linux/Windows? Or do those also need a future fix?

an0n666:
> If so, has this feature been working on Linux/Windows? Or do those also need a future fix?
>
I don't think so. In Windows the listen interface switches to 127.0.0.1 but outgoing remains on 0.0.0.0

arvidn:
This patch affects systems that use the `IP_BOUND_IF` socket option for binding a socket to an interface. It doesn't take a name of the interface, but the index. prior to this fix, failure to resolve the name to an index would not be propagated as an error.

I get the impression that there is no interface 0, so it was still failing, just with a confusing error message.

PR #4488, opened 2020-03-31
moved torrent and choker to aux

aldenml:


arvidn:
do you think this will cause a difficult conflict with https://github.com/arvidn/libtorrent/pull/4417 ?
I'm hoping to wrap that up for 1.2.6

aldenml:
I don't think it will be a big problem, but it's better to hold this until you get to merge `RC_1_2` in `master`, then I will rebase

arvidn:
would you mind rebasing this on top of master now? I'm holding off on the more ambitious changes to the rate-based choker, and I will put that in master laster

aldenml:
sure, I will do it probably later today

aldenml:
Hi @arvidn, finally had some time to do the rebase, but do you know why travis-ci is not starting

aldenml:
well, the jobs start

aldenml:
sure, but just ran it and nothing changed, I think it was never in `fwd.hpp` (not sure how it was filtered)

aldenml:
@arvidn do you think this could be a missing header https://travis-ci.org/github/arvidn/libtorrent/jobs/705057699#L919

arvidn:
it looks more like a missing library being included in the link step. Maybe a header added an additional link-time dependency.

aldenml:
strange, I haven't touch anything in particular here, just moving files around, and this is with `gnutls`

aldenml:
do you think a restart of the job can help?

arvidn:
it might, but my travis fix for OSX hasn't made it to master yet. I'm working on the merge now

arvidn:
it has landed now, the tests should pass if you rebase now, I would think.

arvidn:
`ERR_reason_error_string`, that's an openssl symbol, right? which seems weird in a gnutls build

src/torrent.cpp
```diff
@@ -61,7 +61,16 @@ POSSIBILITY OF SUCH DAMAGE.
...
-#include "libtorrent/torrent.hpp"
+#include "libtorrent/aux_/torrent.hpp"
+
+#ifdef TORRENT_SSL_PEERS
+#include "libtorrent/ssl_stream.hpp"
+#include "libtorrent/aux_/disable_warnings_push.hpp"
+#include <boost/asio/ssl/context.hpp>
+#include <boost/asio/ssl/verify_context.hpp>
+#include "libtorrent/aux_/disable_warnings_pop.hpp"
+#endif // TORRENT_SSL_PEERS
```
arvidn, line 72: perhaps this section is not taking gnutls builds into account.

src/torrent.cpp
```diff
@@ -61,7 +61,16 @@ POSSIBILITY OF SUCH DAMAGE.
...
 #include <functional>
 
-#include "libtorrent/torrent.hpp"
+#include "libtorrent/aux_/torrent.hpp"
+
+#ifdef TORRENT_SSL_PEERS
+#include "libtorrent/ssl_stream.hpp"
+#include "libtorrent/aux_/disable_warnings_push.hpp"
+#include <boost/asio/ssl/context.hpp>
+#include <boost/asio/ssl/verify_context.hpp>
```
arvidn, line 70: yes, these asio SSL headers are only for openssl, not gnutls. These are abstracted by ssl headers in libtorrent

aldenml: I see, I will review it and put the necessary `#if`

arvidn: I think this is what you want: https://github.com/arvidn/libtorrent/blob/master/include/libtorrent/ssl.hpp

aldenml:
rebased and using `libtorrent/ssl.hpp`

PR #4492, opened 2020-04-01
clarify documentation for outgoing_interfaces

arvidn:


include/libtorrent/settings_pack.hpp
```diff
@@ -195,14 +195,14 @@ namespace aux {
...
-			// Setting this to an empty string will disable binding of outgoing
-			// connections.
+			// be a string containing one or more, comma separated, adapter
+			// names or IP addresses. Adapter names on Unix systems are of the
+			// form "eth0", "eth1", "tun0", etc. When specifying multiple
+			// interfaces, they will be assigned in round-robin order. This may
+			// be useful for clients that are multi-homed. Binding an outgoing
+			// connection to a local IP does not necessarily make the connection
+			// via the associated NIC/Adapter. Setting this to an empty string
+			// will disable binding of outgoing connections.
```
sledgehammer999, line 205: There is a discrepancy here. The unchanged text says:
```
An empty string binds outgoing connections to INADDR_ANY and port 0 (i.e. let the OS decide).
```

while the changed text now says:
```
Setting this to an empty string will disable binding of outgoing connections.
```

arvidn: that's a good point. Both are kind of true; there will still be a call to `bind()` either way, but binding to the unspecified address and port 0 is effectively the same as not calling `bind()`. I agree this should be clarified though. The main difference is that there won't be any attempts to *bind-to-device* which is different from a regular `bind()` call.

sledgehammer999:
Also maybe clarify what happens when a given name does not exist?
I suspect you have this flow: Given a random string-> check if it is an IPv4/IPv6-> If not, assume interface name->check if interface name exists-> if not -> ???

PR #4498, opened 2020-04-03
setting a proxy overrides listen_interfaces

arvidn:
The idea with this patch is to simplify the proxy handling. Instead of a proxy being treated as something independent from the network interfaces, it's treated as a network interface itself.

It makes sense that a single proxy really acts as an access point to the internet and when configuring one, whether your machine is multi-homed or not is not relevant, everything is funneled through the proxy.

It simplifies things like:

* UPnP, NAT-PMP, PCP and local service discovery can easily be disabled entirely when using a proxy
* Trackers will only announce once, via the proxy
* There will only be a single DHT running, via the proxy

This change, while cleaning up some common cases, also bans some rare cases. With this patch it will not be possible to combine running SSL torrents with a proxy. The SSL listen socket is also treated as a `listen_socket_t`. It will no longer be possible to combine multiple listen sockets with a proxy (mostly because there's only one proxy that can be configured).

To make this fully generic, each listen interface (i.e. entry in `listen_interfaces` setting) would have to have its own proxy configuration. That hardly seems worth the trouble though.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* simplify proxy handling. A proxy now overrides listen_interfaces
 	* fix issues when configured to use a non-default choking algorithm
 	* fix issue in reading resume data
 	* revert NXDOMAIN change from 1.2.4
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -151,6 +151,10 @@ namespace aux {
 		// listen on an unspecified address (either IPv4 or IPv6)
 		static constexpr listen_socket_flags_t was_expanded = 2_bit;
 
+		// there's a proxy configured, and this is the only one interface
+		// representing that one proxy
+		static constexpr listen_socket_flags_t proxy = 3_bit;
+
 		listen_socket_t() = default;
 
 		// listen_socket_t should not be copied or moved because
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -284,7 +284,14 @@ namespace aux {
 			listen_interfaces,
 
 			// when using a proxy, this is the hostname where the proxy is running
-			// see proxy_type.
+			// see proxy_type. Note that when using a proxy, the
+			// settings_pack::listen_interfaces setting is overridden and only a
+			// single interface is created, just to contact the proxy. This
+			// means a proxy cannot be combined with SSL torrents or multiple
+			// listen interfaces. This proxy listen interface will not accept
+			// incoming TCP connections, will not map ports with any gateway and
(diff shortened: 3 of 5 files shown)


FranciscoPombal:
> With this patch it will not be possible to combine running SSL torrents with a proxy.

Does "SSL torrents" refer to using HTTPS trackers or to [BEP 32](https://www.bittorrent.org/beps/bep_0035.html)?

arvidn:
The latter. The reason is that to open an SSL listen socket (to accept incoming SSL peer connections), you need a `listen_socket_t` entry for that ssl socket. This patch makes a proxy also be its own kind of `listen_socket_t`. And there is no support for a `listen_socket_t` combining both features. Almost nobody uses SSL torrents (afaik), so I don't expect it to be an issue.

mederel:
As mentioned in the QBittorrent issue #11735, I compiled the content of the PR and it "fixed" the issue in QBittorrent.

I am running Manjaro Linux and QBittorrent 4.2.3 (community build for Manjaro).
To install the proxy-interface branch, I used the already existing AUR package libtorrent-rasterbar-git and just switch the source download to checkout the branch proxy-interface. But that is Linux, compiling is of course easier.

I think we are yet to see a Windows user compiling it too and corroborating my findings.

Cheers and kudos for the work provided, fantastic!

After that, I am sorry, I am really not well-versed enough in C++ to be in a position to do any kind of code reviews... I am a dev too, only in another language...

arvidn:
thanks for testing! I will still need to figure out the issue in CI before landing this (and other, more urgent, issues have come up the last few days)

FranciscoPombal:
@arvidn the situation with https://github.com/qbittorrent/qBittorrent/issues/11735 seems to have improved, but there are still UDP tracker issues apparently.

arvidn:
@xavier2k6 I believe that last commit may have addressed the "already open" issue

xavier2k6:
> @xavier2k6 I believe that last commit may have addressed the "already open" issue

ok, will download from [here](https://github.com/arvidn/libtorrent/tree/proxy-interface) & compile a new qBittorrent file for testing in #11735

arvidn:
those last force pushes were just rebasing this on top of latest `RC_1_2`. it should be essentially the same otherwise.

I think this is in a good state now, ready to land. Anyone wanting to test would be appreciated!

FranciscoPombal:
A build of qBittorrent with this patch will soon land in this thread https://github.com/qbittorrent/qBittorrent/issues/11735, hopefully some of the people affected by the proxy issue will test it out.

mederel:
Hi again, I recompiled the latest of proxy-interface and tested again. Working correctly here.
Still QBittorrent 4.2.3 and adapted libtorrent-rasterbar-git package of Manjaro/Arch linux.

PR #4505, opened 2020-04-04
don't open any listen sockets if listen_interfaces is empty

arvidn:
or misconfigured.

@ssiloti I keep forgetting why it's important to open the "default" listen sockets in case none are configured (or if the configured interface doesn't exist). I seem to recall it's because `listen_interfaces` ought to be just for *listening*, and this would prevent outgoing connections as well. I'm having second thoughts on these semantics.

I'm imagining heading in a direction where `listen_interfaces` and `outgoing_interfaces` are merged, and the behavior of actually performing a bind-to-device can be controlled by a boolean setting instead. But there wouldn't be two separate lists of interfaces.

In such future, it *would* make sense to allow having no listen sockets. The main rationale for this is if `listen_interfaces` is configured with a VPN interface that is currently offline (not available).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* don't open any listen sockets if listen_interfaces is empty or misconfigured
 	* fix bug in auto disk cache size logic
 	* fix issue with outgoing_interfaces setting, where bind() would be called twice
 	* add build option to disable share-mode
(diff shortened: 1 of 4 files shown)


ssiloti:
IIRC we should still be able to make outgoing TCP connection with no listen sockets, but uTP will be disabled completely because it depends on the UDP listen sockets. I'm not sure how much this matters since, as this PR points out, if there are no listen interfaces there will be no DHT nodes nor tracker announces.

arvidn:
it's quite likely, still, that the resume data has peers that *will* be connected. In practice, though, `outgoing_interfaces` is also set to the same interface names, so that will fail as well.

PR #4506, opened 2020-04-04
fix bug in auto disk cache size logic

arvidn:


src/platform_util.cpp
```diff
@@ -112,7 +112,7 @@ namespace libtorrent {
 		MEMORYSTATUSEX ms;
 		ms.dwLength = sizeof(MEMORYSTATUSEX);
 		if (GlobalMemoryStatusEx(&ms))
-			ret = int(ms.ullTotalPhys);
+			ret = ms.ullTotalPhys;
```
FranciscoPombal, line 115: Technically, isn't there still a possibility of unsigned -> signed conversion which would return a negative number? If the number is bigger than `std::numeric_limits<int64_t>::max()`, just truncate to that value, so that a negative number is never returned.

arvidn: you're right. I assumed `DWORDLONG` was signed

arvidn: interestingly, there is no warning for this. I think it's safe to assume the amount of memory will not exceed the max for `int64_t` (given that current hardware only can address 48 bits)

FranciscoPombal: Still I think explicitly enforcing the the truncation to `std::numeric_limits<int64_t>::max()` better conveys the intent for humans, even if the machines don't need it.

PR #4509, opened 2020-04-05
changed boost array to shared_ptr in torrent_info

fireuse:
Changed boost type to C++11 type


arvidn:
this is a C++17 feature, and apple doesn't support it (on the version available on travis-ci).
I tried to make this change in the `c++17` branch, but had to back it out because darwin hasn't caught up yet.

arvidn:
https://github.com/arvidn/libtorrent/pull/4167

If you want to pursue this, I would suggest putting it on top the that patch for now (but that will be rebased a few times until I get all the tests to pass as well)

fireuse:
OK, moved to cxx-17

Issue #3917, opened 2019-08-07
Alert 'dht_sample_infohashes_alert' should contain 'node_id'/'nid'


fonic:
libtorrent version (or branch): 1.2.1

Alert `dht_sample_infohashes_alert` should contain an additional property `node_id` (or `nid`) so one knows which node sent the info hash samples. Currently, only property `endpoint` can be used for identification, which I don't think is enough.

arvidn:
this would be an ABI breaking change for 1.2.x though. It could go into master

FranciscoPombal:
Commenting so this does not auto close.

@arvidn Do you know if simply reacting to comments (with :+1:, for instance) counts as activity for the bot? It would be convenient, as otherwise issues might get spammed overtime with "bump" comments.

Anyway, you probably want to exempt issues that are part of milestones and projects from being marked as stale, and prevent PRs from being auto-closed (but not from being marked as stale).

Relevant bits from the official docs:

```
# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
daysUntilClose: 7

# Set to true to ignore issues in a project (defaults to false)
exemptProjects: false

# Set to true to ignore issues in a milestone (defaults to false)
exemptMilestones: false

# Optionally, specify configuration settings that are specific to just 'issues' or 'pulls':
# pulls:
#   daysUntilStale: 30
#   markComment: >
#     This pull request has been automatically marked as stale because it has not had
#     recent activity. It will be closed if no further activity occurs. Thank you
#     for your contributions.

# issues:
#   exemptLabels:
#     - confirmed
```

fonic:
What happened to this? This is still something that should be added/changed, in my opinion.

arvidn:
yes, I agree this is a good idea. I will do this in master, since the alerts are part of the ABI for 1.2.x already.

I wonder if it would make sense to have a `dht_alert` which all dht related alerts derive from, which can tell you which node posted the alert.

Patches are welcome too!

fonic:
> I wonder if it would make sense to have a `dht_alert` which all dht related alerts derive from, which can tell you which node posted the alert.
That might be a good idea. Whenever you do something DHT-related, you almost always need to know who sent what. I can't think of a case where this information would not be relevant.

> Patches are welcome too!
[...]

PR #4517, opened 2020-04-08
Add node_id to dht_sample_infohashes_alert

fonic:
This adds `node_id` to `dht_sample_infohashes_alert` to enable receiver of alert to determine which node sent the reply to an earlier request for info hash samples.

Tested via Python bindings, seems to work as expected.

I'm not sure I got the naming of attributes/arguments right - there are so many varieties in use, e.g. `endpoint` appears as `endpoint/ep/endp`... I opted for `node_id` for the alert attribute and `nid` for method arguments.

Related issue: #3917

### DIFF
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -2771,6 +2771,7 @@ TORRENT_VERSION_NAMESPACE_2
 	{
 		// internal
 		TORRENT_UNEXPORT dht_sample_infohashes_alert(aux::stack_allocator& alloc
+			, sha1_hash const& nid
 			, udp::endpoint const& endp
 			, time_duration interval
 			, int num
@@ -2781,6 +2782,9 @@ TORRENT_VERSION_NAMESPACE_2
 		TORRENT_DEFINE_ALERT(dht_sample_infohashes_alert, 93)
 
 		std::string message() const override;
+
+		// id of the node the request was sent to (and this response was received from)
+		sha1_hash node_id;
 
 		// the node the request was sent to (and this response was received from)
 		aux::noexcept_movable<udp::endpoint> endpoint;
--- a/include/libtorrent/kademlia/dht_tracker.hpp
+++ b/include/libtorrent/kademlia/dht_tracker.hpp
@@ -117,7 +117,8 @@ namespace dht {
 			, std::function<void(std::vector<tcp::endpoint> const&)> f);
 
 		void sample_infohashes(udp::endpoint const& ep, sha1_hash const& target
-			, std::function<void(time_duration
+			, std::function<void(node_id
+				, time_duration
 				, int, std::vector<sha1_hash>
 				, std::vector<std::pair<sha1_hash, udp::endpoint>>)> f);
 
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -5993,12 +5993,13 @@ namespace {
 	void session_impl::dht_sample_infohashes(udp::endpoint const& ep, sha1_hash const& target)
 	{
 		if (!m_dht) return;
-		m_dht->sample_infohashes(ep, target, [this, ep](time_duration const interval
+		m_dht->sample_infohashes(ep, target, [this, ep](sha1_hash const& nid
+			, time_duration const interval
 			, int const num, std::vector<sha1_hash> samples
 			, std::vector<std::pair<sha1_hash, udp::endpoint>> nodes)
 		{
-			m_alerts.emplace_alert<dht_sample_infohashes_alert>(ep
-				, interval, num, std::move(samples), std::move(nodes));
+			m_alerts.emplace_alert<dht_sample_infohashes_alert>(nid
+				, ep, interval, num, std::move(samples), std::move(nodes));
 		});
 	}
 
--- a/test/test_dht.cpp
+++ b/test/test_dht.cpp
@@ -3687,6 +3687,7 @@ TORRENT_TEST(sample_infohashes)
 
 	g_sent_packets.clear();
 
+	node_id nid = generate_random_id();
 	udp::endpoint initial_node = rand_udp_ep();
 	t.dht_node.m_table.add_node(node_entry{initial_node});
 
@@ -3697,10 +3698,13 @@ TORRENT_TEST(sample_infohashes)
 	udp::endpoint const ep2 = rand_udp_ep(rand_v4);
 
 	t.dht_node.sample_infohashes(initial_node, items[0].target,
-		[h1, ep1, h2, ep2](time_duration interval, int num
+		[nid, h1, ep1, h2, ep2](node_id const& node_id
+			, time_duration interval
+			, int num
 			, std::vector<sha1_hash> samples
 			, std::vector<std::pair<sha1_hash, udp::endpoint>> const& nodes)
 	{
+		TEST_EQUAL(node_id, nid);
 		TEST_EQUAL(total_seconds(interval), 10);
 		TEST_EQUAL(num, 2);
 		TEST_EQUAL(samples.size(), 1);
@@ -3740,6 +3744,7 @@ TORRENT_TEST(sample_infohashes)
 	g_sent_packets.clear();
 	send_dht_response(t.dht_node, response, initial_node
 		, msg_args()
+			.nid(nid)
 			.interval(seconds(10))
 			.num(2)
 			.samples({to_hash("1000000000000000000000000000000000000001")})
--- a/include/libtorrent/kademlia/node.hpp
+++ b/include/libtorrent/kademlia/node.hpp
@@ -169,7 +169,8 @@ public:
 		, std::function<void(item&)> data_cb);
 
 	void sample_infohashes(udp::endpoint const& ep, sha1_hash const& target
-		, std::function<void(time_duration
+		, std::function<void(sha1_hash
+			, time_duration
 			, int, std::vector<sha1_hash>
 			, std::vector<std::pair<sha1_hash, udp::endpoint>>)> f);
 
--- a/include/libtorrent/kademlia/sample_infohashes.hpp
+++ b/include/libtorrent/kademlia/sample_infohashes.hpp
@@ -46,7 +46,8 @@ class sample_infohashes final : public traversal_algorithm
 {
 public:
 
-	using data_callback = std::function<void(time_duration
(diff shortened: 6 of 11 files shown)


include/libtorrent/session_handle.hpp
```diff
@@ -496,7 +496,7 @@ namespace libtorrent {
 		// can perform a key space traversal with a single RPC per node by adjusting
 		// the target value for each RPC. It has no effect on the returned sample value.
 		// The result is posted as a ``dht_sample_infohashes_alert``.
-		void dht_sample_infohashes(udp::endpoint const& ep, sha1_hash const& target);
+		void dht_sample_infohashes(sha1_hash const& nid, udp::endpoint const& ep, sha1_hash const& target);
```
arvidn, line 499: I don't understand what this is meant to be used for. It's just passed through into the alert. I was under the impression you wanted the alert to either contain the node-ID of the node responding, or the node ID of our local node that received the response. I don't see how this is useful.

If it's important to add, I think it should either be as the last parameter with a default value, or as a separate overload. In order to stay backwards compatible. It should also be mentioned in the documentation.

src/session_impl.cpp
```diff
@@ -6070,14 +6070,14 @@ namespace {
 		m_alerts.emplace_alert<dht_live_nodes_alert>(nid, nodes);
 	}
 
-	void session_impl::dht_sample_infohashes(udp::endpoint const& ep, sha1_hash const& target)
+	void session_impl::dht_sample_infohashes(sha1_hash const& nid, udp::endpoint const& ep, sha1_hash const& target)
 	{
 		if (!m_dht) return;
-		m_dht->sample_infohashes(ep, target, [this, ep](time_duration const interval
+		m_dht->sample_infohashes(ep, target, [this, nid, ep](time_duration const interval
```
arvidn, line 6076: I would have expected this callback to get passed in the node ID instead of just capturing *some* node ID passed in by the called

include/libtorrent/alert_types.hpp
```diff
@@ -2864,6 +2865,9 @@ TORRENT_VERSION_NAMESPACE_2
 
 		std::string message() const override;
 
+		// id of the node the request was sent to (and this response was received from)
```
arvidn, line 2868: as far as I can tell from the code, this comment is not correct. It's the ID that was passed to the `dht_sample_infohashes()` call.

include/libtorrent/alert_types.hpp
```diff
@@ -2864,6 +2865,9 @@ TORRENT_VERSION_NAMESPACE_2
...
+		sha1_hash node_id;
```
arvidn, line 2869: this change cannot be made in `RC_1_2` since it changes the ABI. It has to go in `master`

include/libtorrent/session_handle.hpp
arvidn: if this is meant to be the node ID of the *target*, how is the caller supposed to know this? The DHT response message will include it, we should just pick it up from there.

fonic: Well... the endpoint is passed through, so I figured the node id should be passed through as well. Of course the caller has to know the node id in the first place, but why wouldn't he/it?

fonic: Well... the endpoint is passed through, so I figured the node id should be passed through as well. Of course the caller has to know the node id in the first place, but why wouldn't he/it? When dealing with DHT, you always know the other nodes' ids, that's in fact what you'll always use to identify other nodes.

The list of live nodes contains node id + endpoint. Info hash sample replies contain node id + endpoint. Thus, the caller always known the node id. Or am I missing something?

arvidn: you can know the endpoint of a DHT node without knowing its node ID. But more importantly, you don't *need* to know its node ID to send a message to it. Its node ID will be returned in its response. Your comment suggest that *that's* the ID that will come back in the alert, but in your patch that's not the case. The node ID you get back is what you pass in, it doesn't matter what you pass in, because it isn't used for anything.

arvidn: The nodes you send this message to are not necessarily in your routing table. In fact, mostly they probably won't be, you will have learned about them in the response of a previous `dht_sample_infohashes_alert`

arvidn: here's an example use, how I envision it: https://github.com/arvidn/libtorrent/pull/4518

fonic: Let's look at it from a different angle: to determine if you got a reply to a previously sent request for info hash samples, you need a reference. A node can change its node id, right? So in this case it would actually be beneficial to have the node id unchanged from request to reply. 

> The nodes you send this message to are not necessarily in your routing table. In fact, mostly they probably won't be, you will have learned about them in the response of a previous `dht_sample_infohashes_alert`

That's my point - the node list of a dht_sample_infohashes_alert contains both node id and endpoint.

arvidn: > to determine if you got a reply to a previously sent request for info hash samples, you need a reference to match them

This is typically handled at a lower layer, with the transaction ID. If you need some kind of "user data" to tie a response to the request, I don't think using a `sha1_hash` is appropriate. Especially not suggesting that it somehow reflects the node ID the response was received from (since it doesn't, right now).

> That's my point - the node list of a dht_sample_infohashes_alert contains both node id and endpoint for each node. So you know the node id.

I don't think that's a good reason to require it to be passed in. It's not used for anything, so why require the user to pass it in? If you need a "user data" cookie, it should be optional.

wouldn't the endpoint work just as well as such cookie?

fonic:
Changed this to retrieve node id directly from reply.

include/libtorrent/kademlia/dht_tracker.hpp
```diff
@@ -111,7 +111,8 @@ namespace libtorrent { namespace dht {
 			, std::function<void(std::vector<tcp::endpoint> const&)> f);
 
 		void sample_infohashes(udp::endpoint const& ep, sha1_hash const& target
-			, std::function<void(time_duration
+			, std::function<void(sha1_hash
```
arvidn, line 114: iirc, there's an alias called `dht::node_id` which I think would be better suited here. They are both the same underlying type, but I think it would be better if they were distinct. at least in the API

include/libtorrent/alert_types.hpp
arvidn: I notice I'm not using `node_id` consistently either...

include/libtorrent/kademlia/dht_tracker.hpp
fonic: changed

arvidn:
could you rebase this branch on top of master instead? (and move the PR to target master as well).

`git rebase --onto master <parent-of-first-commit>`

fonic:
> could you rebase this branch on top of master instead? (and move the PR to target master as well).
> 
> `git rebase --onto master <parent-of-first-commit>`

Yes, but could you help to fix the test first? I can't figure it out. libtorrent is a very complex library, my knowledge about its inner workings are limited, although I'm trying my best.

test/test_dht.cpp
```diff
@@ -3687,6 +3687,7 @@ TORRENT_TEST(sample_infohashes)
 
 	g_sent_packets.clear();
 
+	node_id nid = t.dht_node.nid();
```
arvidn, line 3690: this is *our* node ID. not the node ID that will respond to the sample-infohashes request

fonic: The node we're sending to is `udp::endpoint initial_node = rand_udp_ep();`, correct? So how can we determine its id?

test/test_dht.cpp
```diff
@@ -3697,10 +3698,13 @@ TORRENT_TEST(sample_infohashes)
...
 
 	t.dht_node.sample_infohashes(initial_node, items[0].target,
-		[h1, ep1, h2, ep2](time_duration interval, int num
+		[nid, h1, ep1, h2, ep2](node_id const& node_id
+			, time_duration interval
+			, int num
 			, std::vector<sha1_hash> samples
 			, std::vector<std::pair<sha1_hash, udp::endpoint>> const& nodes)
 	{
+		TEST_EQUAL(node_id, nid);
```
arvidn, line 3707: I think you want to compare `node_id` against the node ID that was set up as part of the `dht_test_setup`. I think you can get to this as: `t.dht_node.nid()`

arvidn:
> Yes, but could you help to fix the test first? I can't figure it out. libtorrent is a very complex library, my knowledge about its inner workings are limited, although I'm trying my best.

certainly. see comments. I haven't had time to test these suggestions locally yet

test/test_dht.cpp
arvidn: oh, I didn't look at this closely enough... this is what you're doing already..

arvidn: ok line 3741, that's where we send the response to this message. after `msg_args()` add a line `.nid(id)`, (with whatever `id` you'd want)

fonic: Ok, now I'm using this to generate a random id
```
node_id nid = generate_random_id();
```
keeping this as is
```
	t.dht_node.sample_infohashes(initial_node, items[0].target,
		[nid, h1, ep1, h2, ep2](node_id const& node_id
			, time_duration interval
			, int num
			, std::vector<sha1_hash> samples
			, std::vector<std::pair<sha1_hash, udp::endpoint>> const& nodes)
	{
		TEST_EQUAL(node_id, nid);
		TEST_EQUAL(total_seconds(interval), 10);
```
and changing this to include the randomly generated nid
```
	send_dht_response(t.dht_node, response, initial_node
		, msg_args()
			.nid(nid)
			.interval(seconds(10))
```

fonic:
> could you rebase this branch on top of master instead? (and move the PR to target master as well).
> 
> `git rebase --onto master <parent-of-first-commit>`

Looks about right. Can't test it though, as this branch won't configure properly on my machine.

test/test_dht.cpp
fonic: It did build fine and checks were passed before rebasing, so it should be fixed now.

fonic:
As far as I can tell from testing with my crawler, this works as expected.

arvidn:
the windows CI build wasn't triggered properly. Just triggered that

PR #4518, opened 2020-04-08
add a DHT tool to issue sample_infohashes DHT messages

arvidn:
to collect samples from the DHT

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,8 @@ matrix:
   fast_finish: true
   include:
     - if: repo = arvidn/libtorrent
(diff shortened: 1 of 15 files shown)


fonic:
Just as a side note, all replies from nodes I've seen so far set an interval of 6h, so you're not to ask them again for info hash samples unless 6h have passed. You should use the interval and not a fixed interval of 1h.

arvidn:
ah, yes. I thought I was doing that. Maybe this is why I hadn't pushed this yet (I just dug it out of the reflog now, to push it)

PR #4526, opened 2020-04-11
fix issue in reading resume data

arvidn:


src/session_impl.cpp
```diff
@@ -4829,6 +4829,14 @@ namespace {
 			return std::make_pair(ptr_t(), false);
 		}
 
+		if (params.ti
+			&& params.info_hash != sha1_hash(0)
```
aldenml, line 4833: what do you think about using `is_all_zeros` instead?

arvidn: probably a good idea

PR #4527, opened 2020-04-11
reduce include directories

orivej:
This removes -I src -I test -I tools from the autotools build and -I ed25519/src from all build systems.

---

libtorrent consistently uses either relative includes or includes from its `include` directory, yet the build systems, especially autotools whose DEFAULT_INCLUDES by default includes the directory of the `Makefile.am` file, pass additional -I flags that make it look like libtorrent needs them.

Tested with autotools (both in the source tree and from a separate build directory), cmake, and bjam.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -669,7 +669,6 @@ if (dht)
 		src/hasher512
 		src/sha512
 	)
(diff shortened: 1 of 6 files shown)


arvidn:
thanks!

PR #4530, opened 2020-04-12
remove (flags & session_handle::save_dht_settings)

pavel-pimenov:


arvidn:
I have a feeling that one of those conditions were meant to check `session_handle::save_settings`. But I'm not sure. It's a bit complicated because the DHT settings were moved from that class into the `settings_pack`, with all the other settings

pavel-pimenov:
if (flags & session_handle::save_settings)
		|| (flags & session_handle::save_dht_settings))
?

arvidn:
looking closer at this, I think your original patch was right

PR #4542, opened 2020-04-14
support boost version older than 1.64

arvidn:
by falling back to tlsv12_client (instead of tls_client)

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -454,8 +454,13 @@ namespace aux {
 		: m_settings(pack)
 		, m_io_service(ios)
 #ifdef TORRENT_USE_OPENSSL
+#if BOOST_VERSION >= 106400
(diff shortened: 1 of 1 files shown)


src/session_impl.cpp
```diff
@@ -454,7 +454,11 @@ namespace aux {
 		: m_settings(pack)
 		, m_io_service(ios)
 #ifdef TORRENT_USE_OPENSSL
+#if BOOST_VERSION >= 106400
 		, m_ssl_ctx(ssl::context::tls_client)
+#else
+		, m_ssl_ctx(ssl::context::tlsv12_client)
+#endif
 		, m_peer_ssl_ctx(ssl::context::tls)
```
cas--, line 462: `tls` also needs an alternative for < 1.64, `tlsv12`

PR #4546, opened 2020-04-15
support newer python versions in python bindings

FranciscoPombal:
Hopefully this can make it to 1.2.6


arvidn:
seems reasonable. I was expecting a URL in there pointing to something like this: https://packaging.python.org/guides/packaging-binary-extensions/

It supports this patch, as it says:

> Note that from Python 3.5 onwards, Visual Studio works in a backward compatible way, which means that any future version of Visual Studio will be able to build Python extensions for all Python versions from 3.5 onwards.

PR #4547, opened 2020-04-15
fix static builds of examples and tools on windows

FranciscoPombal:
Related to https://github.com/microsoft/vcpkg/pull/10686.

~This is a draft PR because I am not sure if this is a good general solution. This is simply what I had to do to get it to build (static windows x64 builds) with `vcpkg`. @zeule comments appreciated.~ fixed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -633,6 +633,12 @@ set_package_properties(OpenSSL
 )
 
 if(OPENSSL_FOUND)
+
+	# TODO: needed until https://gitlab.kitware.com/cmake/cmake/issues/19263 is fixed
+	if(WIN32 AND OPENSSL_USE_STATIC_LIBS)
+		target_link_libraries(OpenSSL::Crypto INTERFACE crypt32)
+	endif()
+
 	target_link_libraries(torrent-rasterbar PUBLIC OpenSSL::SSL)
 	target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_OPENSSL)
 	target_sources(torrent-rasterbar PRIVATE src/pe_crypto)
--- a/docs/building.rst
+++ b/docs/building.rst
@@ -484,6 +484,12 @@ Other build options are:
 
 Options are set on the ``cmake`` command line with the ``-D`` option or later on using ``ccmake`` or ``cmake-gui`` applications. ``cmake`` run outputs a summary of all available options and their current values.
 
+.. note::
+
+	If you are linking statically against OpenSSL on Windows and not using ``-Dstatic_runtime=ON``,
+	you should additionally use the option ``-DOPENSSL_USE_STATIC_LIBS=ON``.
+	If you use ``-Dstatic_runtime=ON``, ``-DOPENSSL_USE_STATIC_LIBS=ON`` is implied.
+
 Step 2: Building libtorrent
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -647,4 +653,3 @@ in a command shell::
 
 This will also install the headers and library files in the visual studio directories to
 be picked up by libtorrent.
-
--- a/docs/hunspell/libtorrent.dic
+++ b/docs/hunspell/libtorrent.dic
@@ -559,6 +559,5 @@ leecher
 6881l
 NOTSENT
 LOWAT
-tls11
-tls12
-tls13
+Dstatic
+DOPENSSL


examples/CMakeLists.txt
```diff
@@ -20,3 +20,13 @@ add_executable(client_test
...
 target_link_libraries(client_test PRIVATE torrent-rasterbar)
+
+if(WIN32 AND NOT BUILD_SHARED_LIBS)
+    target_link_libraries(simple_client PRIVATE crypt32.lib)
+    target_link_libraries(custom_storage PRIVATE crypt32.lib)
+    target_link_libraries(dump_torrent PRIVATE crypt32.lib)
+    target_link_libraries(make_torrent PRIVATE crypt32.lib)
+    target_link_libraries(connection_tester PRIVATE crypt32.lib)
+    target_link_libraries(upnp_test PRIVATE crypt32.lib)
+    target_link_libraries(client_test PRIVATE crypt32.lib)
```
arvidn, line 31: `crypt32.lib` looks like a dependency libtorrent has. I imagine this dependency still boils down to a dependency on a shared library (say, `crypt32.dll`), and this is the link library. And the reason you need this when building libtorrent as a static library is because static libraries cannot propagate dependencies on other shared libraries, the way a shared library can.

However, it seems more appropriate to propagate this dependency via the build system instead of like this (where the dependency is not propagated at all).

Couldn't you make the `libtorrent` target have a `PUBLIC` dependency on `crypt32.lib` or `crypt32.dll`, and have everyone linking against it pick that up?

zeule: Although I do not use Windows, CMake should handle the ".lib" suffix automatically, because it handles the "lib" prefix on POSIX platforms.[Documentation](https://cmake.org/cmake/help/latest/command/target_link_libraries.html) seems to agree with that.

zeule: Ah, yes: https://github.com/arvidn/libtorrent/blob/RC_1_2/CMakeLists.txt#L545
So, I imagine you just need to add something like `$<$<NOT:$<BOOL:BUILD_SHARED_LIBS>>:crypt32>`

arvidn: @FranciscoPombal could you give that a try?

FranciscoPombal: @arvidn yes, I should have that ready today

FranciscoPombal: @arvidn fixed.

arvidn:
@FranciscoPombal I take it this fixes the issue of building libtorrent as a static library on windows, right?

examples/CMakeLists.txt
arvidn: @zeule it's not clear to me why this dependency should not exist when building a shared library. I assume `crypt32` is a cmake target which refers to a DLL. A dependency libtorrent would have regardless of whether it's built as a static library or shared, no?

CMakeLists.txt
```diff
@@ -544,6 +544,7 @@ target_link_libraries(torrent-rasterbar
 if (WIN32)
 	target_link_libraries(torrent-rasterbar
 		PRIVATE
+			$<$<NOT:$<BOOL:${BUILD_SHARED_LIBS}>>:crypt32>
```
zeule, line 547: PRIVATE? How do this propagates to the test executables? Does this change fix the problem?

examples/CMakeLists.txt
zeule: For me too. Did not code for Windows for the last 10 years, have no idea what is libcrypto.lib in the linking error messages.

CMakeLists.txt
FranciscoPombal: It does work, yes. I see in the output dump of the compiler/linker commands that when `BUILD_SHARED_LIBS` is true, `crypt32` is not linked, and when it is false, `crypt32` gets linked.

zeule: Linked to the examples?

FranciscoPombal: Yes, both examples and tools.

zeule: Must be something special about static libs on Windows. Can't understand both the problem and the solution :D

FranciscoPombal: @zeule my initial solution was an attempt to solve this problem: https://github.com/microsoft/vcpkg/pull/10686#issuecomment-612817769. Then I simply made the change you suggested which makes sense, as it's doing the same thing but "upstream". However, I also don't understand why _only_ static builds would need this either. Could it be the case that this is only needed if using static libcrypto? I ask this because this happened with `vcpkg`, and a static build there pulls in static-everything.

FranciscoPombal:
@arvidn 
> @FranciscoPombal I take it this fixes the issue of building libtorrent as a static library on windows, right?

Just the examples and tools, the rest library itself already built just fine even without this fix.

CMakeLists.txt
zeule: …> static build there pulls in static-everything.

Aha, thanks! that helps to understand it a bit better. Seems like there is no need to make changes to libtorrent target. You might need to play with `OPENSSL_USE_STATIC_LIBS` and `OPENSSL_MSVC_STATIC_RT` in vcpkg, or it's a bug in cmake (FindOpenSSL module). Please check FindOpenSSL.cmake for details.

zeule: Wait, we have them already under the `static_runtime` flag. Doe vcpkg sets this option?

FranciscoPombal: > Does vcpkg set this option?

Not sure, but I don't think so. I will check that.

zeule: There is no "crypt32" string in the FindOpenSSL.cmake on my machine. The log in microsoft/vcpkg#10686 shows it is the OpenSSL libcrypto who needs crypt32.lib. Here is OpenSSL issue: https://github.com/openssl/openssl/issues/1061. I'd you encountered a  bug in FindOpenSSL.cmake.

zeule: Finally the root of it: https://gitlab.kitware.com/cmake/cmake/-/issues/19263

FranciscoPombal: @zeule nice find. I haven't yet checkedif passing in `static_runtime` fixes the issue, but if I understand correctly, it won't, right?

zeule: Yes, it won't. You need to patch FindOpenSSL.cmake instead.

FranciscoPombal: @zeule Until the problem in in `FindOpenSSL.cmake` is fixed upstream, how about I add a comment like `TODO: needed until https://gitlab.kitware.com/cmake/cmake/-/issues/19263 is fixed` next to the change?

zeule: Me? It's up to @arvidn.

zeule: If put the patch here, here is my suggestion: 
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index c8e284480..5f1fa52c5 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,6 +262,7 @@ if (encryption)
    TYPE RECOMMENDED
    PURPOSE "Provides encryption support to libtorrent"
  )
+ target_link_libraries(OpenSSL::SSL INTERFACE $<$<BOOL:${OPENSSL_USE_STATIC_LIBS}>:crypt32>)
  target_sources(torrent-rasterbar PRIVATE src/mpi src/pe_crypto)
  target_link_libraries(torrent-rasterbar PUBLIC OpenSSL::SSL)
  target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_OPENSSL)
```

zeule: I forgot to add the `WIN32` condition, but the idea is clear, I believe

FranciscoPombal: > If put the patch here, here is my suggestion:
> 
> ...

What is the advantage of doing it that way?

Also personally I'm not a fan of putting the `WIN32` condition in the generator expression as well, it will become really unreadable, but if it is technically the much better option, well...

zeule: We went from unconditionally appending crypt32 to the executables to fixing a bug in cmake modifying the target which is has incorrect dependencies.

FranciscoPombal: @zeule I noticed in the CMake file, `OPENSSL_USE_STATIC_LIBS` only gets set to `TRUE` if `static-runtime` is used, but not if `BUILD_SHARED_LIBS` is false; I don't think that's what we want, because then if `BUILD_SHARED_LIBS` is false but `static-runtime` is not specified, the build will still fail.

zeule: Pardon me, but it can't be true. Whether a build fails depends only on the OpenSSL libraries. FindOpenSSL.cmake does not account for BUILD_SHARED_LIBS. If there are only static OpenSSL libs available, the libtorrent build will fail always.

FranciscoPombal: @zeule which `CMakeLists.txt` file is that patch for? I grep'd "Provides encryption support to libtorrent" but found nothing.

zeule: Sorry, that was RC_1_1 branch. [Here](https://github.com/arvidn/libtorrent/blob/RC_1_2/CMakeLists.txt#L636)

FranciscoPombal: @zeule 
Still, I don't understand why you suggest `if (WIN32 AND OPENSSL_USE_STATIC_LIBS)` and not `if (WIN32 AND BUILD_SHARED_LIBS)`. The build fails with the former, unless I pass in `static_runtime`, whereas with the latter, the build succeeds regardless of `static_runtime`.

zeule: This is puzzling. Could you, please, share contents of the `INTERFACE_LINK_LIBRARIES` property for the `OpenSSL::SSL` target in both cases?

```cmake
get_property(ill TARGET OpenSSL::SSL PROPERTY INTERFACE_LINK_LIBRARIES)
message(STATUS "ill: ${ill}")
```

zeule: And `IMPORTED_LOCATION` too, please.

FranciscoPombal: @zeule 
I appended this at the end of the `CMakeLists.txt`:

```
get_property(ill TARGET OpenSSL::SSL PROPERTY INTERFACE_LINK_LIBRARIES)
message(STATUS "ill: ${ill}")
get_property(imp TARGET OpenSSL::SSL PROPERTY IMPORTED_LOCATION)
message(STATUS "imp: ${imp}")
```

This was the output at the end of the configuring step (in `vcpkg` this goes to a log file), building `libtorrent[examples,tools]:x64-windows-static` from my `vcpkg` PR, except with your patch from https://github.com/arvidn/libtorrent/pull/4547#discussion_r409779283 instead of my "force link"`crypt32` patches:
```
ill: OpenSLL::Crypto
itt: C:/Users/User/Documents/vcpkg/installed/x64-windows-static/lib/libssl.lib
```

FranciscoPombal: @zeule 
I guess I can just pass `-DOPENSSL_USE_STATIC_LIBS=ON` manually to the libtorrent build. That way I can have that without using `-Dstatic_runtime`. Was this what you meant?

zeule: No, that is not what I meant. You told us the libtorrent build fails if you define `BUILD_SHARED_LIBS` to `False`. But the linking error comes from `libcrypt` (OpenSLL::Crypto). How that one can depend on `BUILD_SHARED_LIBS`? If we want to understand what's going on, we need `INTERFACE_LINK_LIBRARIES` and `IMPORTED_LOCATION` for OpenSLL::Crypto in the vcpkg environment for these three cases, please:

1. `BUILD_SHARED_LIBS=True`, `OPENSSL_USE_STATIC_LIBS=False`
2. `BUILD_SHARED_LIBS=False`, `OPENSSL_USE_STATIC_LIBS=False`
3. `BUILD_SHARED_LIBS=False`, `OPENSSL_USE_STATIC_LIBS=True`

FranciscoPombal: Actually, thinking about it a bit more, I think the correct thing to do is to just pass `-DOPENSSL_USE_STATIC_LIBS=ON` if the library linkage is static, but not the CRT linkage, and if both are static, just pass `-Dstatic_runtime`.

FranciscoPombal: > Actually, thinking about it a bit more, I think the correct thing to do is to just pass `-DOPENSSL_USE_STATIC_LIBS=ON` if the library linkage is static, but not the CRT linkage, and if both are static, just pass `-Dstatic_runtime`.

Please note I wrote this comment before seeing your previous (the conversation did not update on my end for some reason).

FranciscoPombal: > You told us the libtorrent build fails if you define `BUILD_SHARED_LIBS` to `False`

No, I said that's when the build succeeded (because when building `x64-windows-static`, `BUILD_SHARED_LIBS` gets set to `FALSE`) with the previous patch (`$<$<NOT:$<BOOL:${BUILD_SHARED_LIBS}>>:crypt32>`). In the same circusmstances, the build was failing with your patch:

```
if (WIN32 AND OPENSSL_USE_STATIC_LIBS)
    target_link_libraries(OpenSSL::SSL INTERFACE crypt32)
endif()
```

zeule: > think the correct thing to do is to just pass -DOPENSSL_USE_STATIC_LIBS=ON

But that should not help, because of the bug in CMake-supplied OpenSSL module, right?

FranciscoPombal: @zeule 
The whole point of setting `OPENSSL_USE_STATIC_LIBS` is to satisfy the condition imposed by your patch (https://github.com/arvidn/libtorrent/pull/4547#discussion_r409779283), which is now what the PR is.

The way the PR is now, everything seems to work correctly.

- static library linkage + static CRT -> set `-Dstatic_runtime=ON`, build succeeds
- dynamic library linkage + dynamic CRT -> do nothing, build succeeds
- static library linkage + dynamic CRT -> set `-DOPENSSL_USE_STATIC_LIBS=ON`, build succeeds.

FranciscoPombal: I suppose this is the correct thing to do, the OpenSSL docs have this to say:

> If you **link with static OpenSSL libraries then you're expected to
 additionally link** your application with WS2_32.LIB, GDI32.LIB,
 ADVAPI32.LIB, **CRYPT32.LIB** and USER32.LIB.

(emphasis mine, source: https://github.com/openssl/openssl/blob/6f892296038490a7fa24b32ac6f7305687634fb0/NOTES.WIN#L127)

which is what this patch achieves now, provided the user passes `-Dstatic_runtime=ON` or `-DOPENSSL_USE_STATIC_LIBS=ON`, depending on what they need.

CMakeLists.txt
```diff
@@ -633,6 +633,12 @@ set_package_properties(OpenSSL
 )
 
 if(OPENSSL_FOUND)
+
+	# TODO: needed until https://gitlab.kitware.com/cmake/cmake/issues/19263 is fixed
+	if(WIN32 AND OPENSSL_USE_STATIC_LIBS)
+		target_link_libraries(OpenSSL::SSL INTERFACE crypt32)
```
zeule, line 639: Shouldn't it be OpenSSL::Crypto?

FranciscoPombal: @zeule 
No idea. This way, it works. What is the difference? If it is better that way, I could test it out I suppose.

CMakeLists.txt
FranciscoPombal: Actually, I just noticed that currently `advapi32` is not mentioned anywhere, and it is another library that it is recommended to link against (ignoring the GUI-related libraries). Currently, there are no build errors due to the lack of it, but it could be needed due to some future change. @arvidn @zeule is it needed?

CMakeLists.txt
FranciscoPombal: Ok, I have tested with `OpenSSL::Crypto` and it also works fine. I have no idea why it works _both_ ways, but I changed to `OpenSSL::Crypto` anyway because it seems to be "more correct".

zeule: It works both ways because libtorrent links to OpenSSL::SSL, which in turn links to OpenSSL::Crypto, as you showed yourself earlier.

FranciscoPombal:
@arvidn if you don't have additional technical changes to suggest, let me know if I should update `building.rst` in this PR or in a new one. I think it's important to document the need to use `-DOPENSSL_USE_STATIC_LIBS=ON` (or `-Dstatic_runtime=ON`) when linking against static OpenSSL on Windows.

FranciscoPombal:
@arvidn updated docs.

arvidn:
thanks!

arvidn:
@FranciscoPombal it took me a while to figure out why the CI kept failing :/

https://travis-ci.org/github/arvidn/libtorrent/jobs/676139972#L702

```
spellcheck failed, fix words or add to dictionary:
building.html: Dstatic
building.html: DOPENSSL
building.html: Dstatic
building.html: DOPENSSL
make: *** [spell-check] Error 1
```

if you have `hunspell` you should be able to run `make spell-check` in the `docs` directory to test.

I think the most reasonable solution is to add these words to the `docs/hunspell/libtorrent.dic` dictionary file. Ideally you would add the whole words (i.e. `-Dstatic_runtime=ON`), but I'm not sure that works. It's possible hunspell splits the words first, before looking them up.

arvidn:
yeah, I think this has to be fixed like this:
```
index c3835fcb7..499a638bc 100644
--- a/docs/hunspell/libtorrent.dic
+++ b/docs/hunspell/libtorrent.dic
@@ -562,3 +562,5 @@ LOWAT
 tls11
 tls12
 tls13
+Dstatic
+DOPENSSL
```

arvidn:
https://github.com/FranciscoPombal/libtorrent/pull/1

FranciscoPombal:
@arvidn why does it attempt to translate stuff that's between \`\`? Isn't stuff like that supposed to be non-translatable?

PR #4553, opened 2020-04-18
check flags when matching listen sockets

ssiloti:
If the flags change the socket needs to be re-opened. This is
particularly important when the local_network flag changes so that
the socket will correctly be treated as global even if the gateway
was not configured when the address was first detected.

Don't just modify the flags of the existing listen socket because
some flags may influence how the socket is opened. Such flags could
be special cased, but it doesn't seem worth the trouble.


arvidn:
thanks!

PR #4555, opened 2020-04-18
fix quoting of path in cmake generate-pkg-config

arvidn:


zeule:
There are other places where unquoted paths are used:

https://github.com/arvidn/libtorrent/blob/RC_1_2/CMakeLists.txt#L523
https://github.com/arvidn/libtorrent/blob/RC_1_2/CMakeLists.txt#L678
https://github.com/arvidn/libtorrent/blob/RC_1_2/CMakeLists.txt#L679
https://github.com/arvidn/libtorrent/blob/RC_1_2/CMakeLists.txt#L796
https://github.com/arvidn/libtorrent/blob/RC_1_2/bindings/python/CMakeLists.txt#L137

arvidn:
I'm kind of surprised that cmake can expand a single variable into multiple arguments (at least if I understand correctly). It knows what's a path and what isn't, right? You'd think it could just do the quoting itself. It's not obvious to me which arguments need quoting.

When adding quotes to [here](https://github.com/arvidn/libtorrent/blob/RC_1_2/CMakeLists.txt#L523),  cmake failed. The path got double quoted and interpreted as a relative path (it was an absolute path, but I suppose they only strip one level of quotes and then look at the first character, which was still `"` and not `/`)

zeule:
My guess is that has something to do with lists. Lists in cmake are strings with semicolons separating (or should I say joining?) elements. If at particular place cmake expects a list (like, probably,  target_include_directories), it scans for semicolons and is happy to handle spaces in strings, joining multiple arguments back into strings with spaces. If it expects multiple arguments and no lists, spaces remain, unless enquoted, argument separators (like in the original error).

arvidn:
patching all those places failed on the windows CI (but passed locally on ubuntu for me). I don't think it's worth trying to fix potential issues right now. If someone runs into them later and there's a bug report it should be addressed then

PR #4558, opened 2020-04-18
fix IPv6 address change notification on Windows

ssiloti:
The old NotifyAddrChange only detects IPv4 address changes. Use the newer
NotifyUnicastIpAddressChange function instead which supports both v4 and v6.

### DIFF
--- a/src/ip_notifier.cpp
+++ b/src/ip_notifier.cpp
@@ -44,8 +44,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #elif defined TORRENT_WINDOWS
 #include "libtorrent/aux_/throw.hpp"
 #include "libtorrent/aux_/disable_warnings_push.hpp"
-#include <boost/asio/windows/object_handle.hpp>
 #include <iphlpapi.h>
+#include <mutex>
 #include "libtorrent/aux_/disable_warnings_pop.hpp"
 #endif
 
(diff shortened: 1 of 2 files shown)


src/ip_notifier.cpp
```diff
@@ -360,40 +359,52 @@ struct ip_change_notifier_impl final : ip_change_notifier
...
 	}
 
 private:
-	OVERLAPPED m_ovl = {};
-	boost::asio::windows::object_handle m_hnd;
+	static void address_change_cb(void* ctx, MIB_UNICASTIPADDRESS_ROW*, MIB_NOTIFICATION_TYPE)
+	{
+		ip_change_notifier_impl* impl = static_cast<ip_change_notifier_impl*>(ctx);
+		std::lock_guard<std::mutex> l(impl->m_cb_mutex);
+		impl->m_ios.post([cbs = std::move(impl->m_cb)]()
```
arvidn, line 396: `RC_1_2` only requires C++11, so you can't move into lambda (yet) :/

arvidn: (if you leave a comment mentioning C++14, it will be easier to find it once this lands in master)

src/ip_notifier.cpp
```diff
@@ -360,40 +359,52 @@ struct ip_change_notifier_impl final : ip_change_notifier
...
+		{
+			for (auto& cb : cbs) cb(error_code());
+		});
+	}
+
+	io_service& m_ios;
+	HANDLE m_hnd = NULL;
```
arvidn, line 403: I prefer `nullptr`. I think there's a clang-tidy check for that in CI too

src/ip_notifier.cpp
```diff
@@ -348,10 +348,9 @@ struct ip_change_notifier_impl final : ip_change_notifier
 struct ip_change_notifier_impl final : ip_change_notifier
 {
 	explicit ip_change_notifier_impl(io_service& ios)
-		: m_hnd(ios, WSACreateEvent())
+		: m_ios(ios)
 	{
-		if (!m_hnd.is_open()) aux::throw_ex<system_error>(WSAGetLastError(), system_category());
-		m_ovl.hEvent = m_hnd.native_handle();
+		NotifyUnicastIpAddressChange(AF_UNSPEC, address_change_cb, this, false, &m_hnd);
```
arvidn, line 353: since this class now owns `m_hwnd`, it needs to either implement move constructor and move assignment (or delete them)

ssiloti: Right. This class is non-movable now anyways due to the mutex, will delete them.

src/ip_notifier.cpp
```diff
@@ -360,40 +359,52 @@ struct ip_change_notifier_impl final : ip_change_notifier
...
 	}
 
 	void cancel() override
 	{
-		CancelIPChangeNotify(&m_ovl);
-		m_hnd.cancel();
+		std::vector<std::function<void(error_code const&)>> cbs;
+		{
+			std::lock_guard<std::mutex> l(m_cb_mutex);
+			cbs = std::move(m_cb);
```
arvidn, line 386: I forget, is `std::vector` guaranteed to be empty after move? I think it needs to be here, right?

src/ip_notifier.cpp
ssiloti: Wait no, mutex is moveable. I'll implement the move functions.

src/ip_notifier.cpp
ssiloti: Yes, moved-from vectors are always empty.

arvidn:
I'm about to cut 1.2.6. Do you think this is reliable enough to include? I take it you have tested it to some degree

src/ip_notifier.cpp
ssiloti: Haha no I had it right the first time, mutexes aren't movable so neither is this class.

ssiloti:
I've tested the happy path with adding/removing IPv6 addresses as well as whole network interfaces. I don't think this is important enough to push out right away though.

gtumanyan:
Windows XP doesn't support [NotifyUnicastIpAddressChange](https://docs.microsoft.com/en-us/windows/win32/api/netioapi/nf-netioapi-notifyunicastipaddresschange)

FranciscoPombal:
>  doesn't support [NotifyUnicastIpAddressChange](https://docs.microsoft.com/en-us/windows/win32/api/netioapi/nf-netioapi-notifyunicastipaddresschange)

I don't think XP is supported anymore, and doubt Vista is as well. In fact, at this point, even 7 is officially EOL.

Issue #4567, opened 2020-04-22
HTTP tracker announce fails due to read timeout on slow trackers


qix67:
**Please provide the following information**

libtorrent version (or branch):
1.2.5

platform/architecture:
linux/x86_64

compiler and compiler version:
gcc 9.2.0

please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

I have encounted a private HTTP tracker reported as "not working" by qBittorrent. The problem occurs during announce request. The client closes the connection after ~5 seconds.

I picked up announce request using tcpdump and then tried to send the same announce request manually using curl. The request works properly. I noticed the tracker immediatly accepts connection but then takes more than 11s to reply.

In src/http_connection.cpp line 239, replacing:
m_read_timeout = seconds(5);
by
m_read_timeout = seconds(20);
makes the tracker available however it is surely not a good solution.

It whould be great to have an option to adjust this timeout.

FranciscoPombal:
You can adjust with the `stop_tracker_timeout` setting
https://github.com/arvidn/libtorrent/blob/a9968916ca82366f1c236af59aaecb9bc94ffe73/include/libtorrent/settings_pack.hpp#L860

qix67:
I have tried this variable but I think the description is misleading because it speaks about "stopped message" and in my case, it is a "start event". Moreover, it is nowhere written this value is divided by 5 (line 246).

FranciscoPombal:
You're right, that is only for `stop` events. Otherwise, this is the timeout you want to configure:
https://github.com/arvidn/libtorrent/blob/a9968916ca82366f1c236af59aaecb9bc94ffe73/include/libtorrent/settings_pack.hpp#L848

Which as a default value of `30`:
https://github.com/arvidn/libtorrent/blob/a9968916ca82366f1c236af59aaecb9bc94ffe73/src/settings_pack.cpp#L218

You can see how the distinction between event types here:
https://github.com/arvidn/libtorrent/blob/a9968916ca82366f1c236af59aaecb9bc94ffe73/src/http_tracker_connection.cpp#L215

See also the related setting `tracker_receive_timeout`.

But yes, I have no idea why this happens:
https://github.com/arvidn/libtorrent/blob/a9968916ca82366f1c236af59aaecb9bc94ffe73/src/http_connection.cpp#L240

[...]

PR #4569, opened 2020-04-23
simplify and increase default HTTP tracker timeout

arvidn:
Specifically, the "read timeout" is removed. The remaining timeout logic is simply checking to see if the whole announce is complete after the timeout, and closes the connection if it isn't.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* increase default tracker timeout
 	* retry failed socks5 server connections
 	* allow UPnP lease duration to be changed after device discovery
 	* fix IPv6 address change detection on Windows
--- a/include/libtorrent/http_connection.hpp
+++ b/include/libtorrent/http_connection.hpp
@@ -178,7 +178,6 @@ private:
 	http_filter_handler m_filter_handler;
 	deadline_timer m_timer;
 
-	time_duration m_read_timeout;
 	time_duration m_completion_timeout;
 
 	// the timer fires every 250 millisecond as long
--- a/src/http_connection.cpp
+++ b/src/http_connection.cpp
@@ -83,7 +83,6 @@ http_connection::http_connection(io_service& ios
 	, m_connect_handler(ch)
 	, m_filter_handler(fh)
 	, m_timer(ios)
-	, m_read_timeout(seconds(5))
 	, m_completion_timeout(seconds(5))
 	, m_limiter_timer(ios)
 	, m_last_receive(aux::time_now())
@@ -236,11 +235,8 @@ void http_connection::start(std::string const& hostname, int port
 	std::shared_ptr<http_connection> me(shared_from_this());
 
 	m_completion_timeout = timeout;
-	m_read_timeout = seconds(5);
-	if (m_read_timeout < timeout / 5) m_read_timeout = timeout / 5;
 	error_code ec;
-	m_timer.expires_from_now(std::min(
-		m_read_timeout, m_completion_timeout), ec);
+	m_timer.expires_from_now(m_completion_timeout, ec);
 	ADD_OUTSTANDING_ASYNC("http_connection::on_timeout");
 	m_timer.async_wait(std::bind(&http_connection::on_timeout
 		, std::weak_ptr<http_connection>(me), _1));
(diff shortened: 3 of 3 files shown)


FranciscoPombal:
Shouldn't this be done at the `http_tracker_connection.cpp` level? `http_connection` is used for other things like HTTP webseeding, right?

Anyway I still think exposing a setting to control the "the timeout is shorter if nothing is being read" logic would be better. No matter what you set the divisor to, it will still be unexpected from the library user's perspective that under certain circumstances the timeout will be shorter than they expect.

arvidn:
perhaps the timeout setting should control the read-timeout, and the completion timeout should be a factor away from that. It seems excessive to configure these independently

an0n666:
@arvidn this looks like an invisible timeout. Since users will probably set a completion timeout and take it for granted, is it necessary to make the timeout shorter in case the tracker is too slow? I mean if someone sets the completion timeout to 5 seconds it would actually time out lot faster. Because tracker responses don’t come in slowly because of network. Usually it takes the time in processing the request. So there’d be no data sent in between.

arvidn:
@an0n666 are you suggesting I remove the "read timeout"? This was added quite a while ago, and I'm pretty sure I *did* experience trackers trickling in data. But if the read-timeout is believed to be unnecessary and unintuitive, I'd be happy to remove it.

an0n666:
Yes I think it’s best to not keep it. 
Trackers don’t necessarily trickle data in all cases. It could just be taking a long time to process the request due poor design/large database etc.

FranciscoPombal:
> It seems excessive to configure these independently

To be fair, my suggestion was not to have both timers independently configurable. Only the completion timer would be configurable, and then there would be a configurable boolean setting to enable/disable the use of the shorter read timer. But your idea:

> perhaps the timeout setting should control the read-timeout, and the completion timeout should be a factor away from that.

seems better. This way, the libtorrent errs on the side of caution.

arvidn:
on the other hand, it doesn't ever make sense for the read-timeout to be longer than the completion timeout. That's effectively removing it.

Do I get another vote for removing the read timeout entirely? :)

FranciscoPombal:
I posted my comment before seeing @an0n666 's comments. I somewhat agree. For me, either of the two would be sensible:
- remove the read timeout, at least for http tracker connections
- expose the read timeout as the configurable timeout instead of the completion timeout, and have the completion timeout be obtained by multiplying the read timeout by some constant >1

arvidn:
I don't see any objections to this. so I will probably land it soon

FranciscoPombal:
@arvidn 
Actually, I believe this has not yet been addressed: https://github.com/arvidn/libtorrent/issues/4567#issuecomment-618390368
> Since you're removing the read timer completely, it would probably then make sense to set the default of the completion timer a bit lower, to maybe 10 or less.

Along with the corresponding docs update, if you do decide to change it.

arvidn:
> Since you're removing the read timer completely, it would probably then make sense to set the default of the completion timer a bit lower, to maybe 10 or less.

One of the motivating issue reports had a tracker take longer than 10 seconds to respond though, or do I misremember?

FranciscoPombal:
> One of the motivating issue reports had a tracker take longer than 10 seconds to respond though, or do I misremember?

Yes, but as it stands, with this change the default completion timeout is jumping all the way to 30 seconds. The default should be maybe 10 or 15, no? People with slower tracker can then bump the setting.

arvidn:
well, the main reason to keep the timeout short is for (likely public) trackers where most of them fail. then you want them to fail fast to move on to the next one

FranciscoPombal:
@arvidn 
> well, the main reason to keep the timeout short is for (likely public) trackers where most of them fail. then you want them to fail fast to move on to the next one

Still, if you think 30 seconds default won't be too bad even taking that into account, I don't have other objections.

On another note, why aren't they tried asynchronously?

arvidn:
> On another note, why aren't they tried asynchronously?

Because the specification makes the assumption that people publishing torrents also make sure their trackers work, and not random people using other random people's trackers that probably won't work. i.e. that trackers are likely to work and the other trackers are there for back-up in case the first one goes down.

Although, there are settings in libtorrent to announce to all trackers and tiers in parallel

FranciscoPombal:
@arvidn 
> > On another note, why aren't they tried asynchronously?
> 
> Because the specification makes the assumption that people publishing torrents also make sure their trackers work, and not random people using other random people's trackers that probably won't work. i.e. that trackers are likely to work and the other trackers are there for back-up in case the first one goes down.
> 
> Although, there are settings in libtorrent to announce to all trackers and tiers in parallel

Well, since there are already setting to configure that, that's ok. But the assumption that 
> people publishing torrents also make sure their trackers work

is slightly outdated IMO. Nowadays, people stuff torrents with 'known" general-purpose public trackers for resilience, with the expectation that some of them might go down, possibly soon.

PR #4572, opened 2020-04-23
make upnp lease duration configurable on the fly

arvidn:


chchia:
thanks man, this work!

PR #4574, opened 2020-04-23
Improve python cmake

FranciscoPombal:
Basically changes to use the new FindPython3: `https://cmake.org/cmake/help/v3.17/module/FindPython3.html`. Also drops EOL python versions.

---

Warning: slightly outdated info below. See https://github.com/arvidn/libtorrent/pull/4574#issuecomment-709481112 for updated info.

The first commit introduces changes that make it work the same as before (on Ubuntu 18.04, I have to specify `-Dboost-python-module-name=python36`).

The second commit makes further use of the new variables. One benefit is eliminating the build-time dependency on `python3-dev` and `python3-setuptools` packages. It requires the latest CMake version as of writing, but I don't think that's an issue, given how easy it is to install the latest version on basically any system. But I haven't managed to get the same results as before - the site-packages directory does not seems to respect the `CMAKE_INSTALL_PREFIX`.

Without this commit, the discovered site-packages directory is relative, while with it, it becomes absolute. Here are the differences in the configure output as well as the `install_manifest.txt`, for a run with the install prefix set to `/usr/local`:

- In the configure step:

`Python site packages: lib/python3/dist-packages` vs `Python site packages: /usr/lib/python3/dist-packages`

- Diff of the install manifests:

```diff
--- install_manifest_old.txt
+++ install_manifest_new.txt
@@ -253,8 +253,8 @@
 /usr/local/lib/cmake/LibtorrentRasterbar/LibtorrentRasterbarConfig.cmake
 /usr/local/lib/cmake/LibtorrentRasterbar/LibtorrentRasterbarConfigVersion.cmake
 /usr/local/share/cmake/Modules/FindLibtorrentRasterbar.cmake
-/usr/local/lib/python3/dist-packages/libtorrent.cpython-38-x86_64-linux-gnu.so
-/usr/local/lib/python3/dist-packages/libtorrent.egg-info/SOURCES.txt
-/usr/local/lib/python3/dist-packages/libtorrent.egg-info/top_level.txt
-/usr/local/lib/python3/dist-packages/libtorrent.egg-info/PKG-INFO
-/usr/local/lib/python3/dist-packages/libtorrent.egg-info/dependency_links.txt
\ No newline at end of file
+/usr/lib/python3/dist-packages/libtorrent.cpython-38-x86_64-linux-gnu.so
+/usr/lib/python3/dist-packages/libtorrent.egg-info/SOURCES.txt
+/usr/lib/python3/dist-packages/libtorrent.egg-info/top_level.txt
+/usr/lib/python3/dist-packages/libtorrent.egg-info/PKG-INFO
+/usr/lib/python3/dist-packages/libtorrent.egg-info/dependency_links.txt
\ No newline at end of file
```

Also, is it a problem if the `boost_python` module version is different from the python library version? e.g. `libboost_python36.so` with `libpython3.8.so`? I have found that this build script allows that, with or without this PR.

Finally, should `Python3_USE_STATIC_LIBS=ON` depend directly on the value of `BUILD_SHARED_LIBS`, or does it make sense to leave them independently configurable?
@zeule any tips?

---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -153,7 +153,16 @@ install:
   - ccache --show-stats
   - ccache --zero-stats
 
+  # CMake from Kitware is installed in /usr/bin. TravisCI installs its own cmake to another location
+  # which overrides other installations if they don't call the new binary directly
+  # Install and prepend a symlink to the latest version to PATH so that it gets used instead.
   - if [[ "$cmake" == "1" ]]; then
+      wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | sudo tee /etc/apt/trusted.gpg.d/kitware.gpg >/dev/null;
+      sudo apt-add-repository 'deb https://apt.kitware.com/ubuntu/ focal main';
+      sudo apt install cmake;
+      sudo mkdir -p /usr/local/cmake-latest/bin;
+      sudo ln -s /usr/bin/cmake /usr/local/cmake-latest/bin/cmake;
+      PATH=/usr/local/cmake-latest/bin:$PATH;
       which cmake;
       cmake --version;
     fi
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -623,6 +623,7 @@ feature_option(build_tests "build tests" OFF)
 feature_option(build_examples "build examples" OFF)
 feature_option(build_tools "build tools" OFF)
 feature_option(python-bindings "build python bindings" OFF)
+feature_option(python-egg-info "generate python egg info" OFF)
 
 # these options require existing target
 feature_option(dht "enable support for Mainline DHT" ON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -624,6 +624,7 @@ feature_option(build_examples "build examples" OFF)
 feature_option(build_tools "build tools" OFF)
 feature_option(python-bindings "build python bindings" OFF)
 feature_option(python-egg-info "generate python egg info" OFF)
+feature_option(python-install-system-dir "Install python bindings to the system installation directory rather than the CMake installation prefix" OFF)
 
 # these options require existing target
 feature_option(dht "enable support for Mainline DHT" ON)
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -90,21 +90,23 @@ endif()
 message(STATUS "Python 3 site packages: ${Python3_SITEARCH}")
 message(STATUS "Python 3 extension suffix: ${Python3_SOABI}")
 
-set(SETUP_PY_IN "${CMAKE_CURRENT_SOURCE_DIR}/setup.py.cmake.in")
-set(SETUP_PY    "${CMAKE_CURRENT_BINARY_DIR}/setup.py")
-set(OUTPUT      "${CMAKE_CURRENT_BINARY_DIR}/timestamp")
-set(DEPS        python-libtorrent "${SETUP_PY}")
-
-configure_file(${SETUP_PY_IN} ${SETUP_PY} @ONLY)
-
-add_custom_command(OUTPUT ${OUTPUT}
-                   COMMAND ${Python3_EXECUTABLE} ${SETUP_PY} build -b "${CMAKE_CURRENT_SOURCE_DIR}"
-                   COMMAND ${Python3_EXECUTABLE} ${SETUP_PY} egg_info -b "${CMAKE_CURRENT_SOURCE_DIR}"
-                   COMMAND ${CMAKE_COMMAND} -E touch ${OUTPUT}
-                   DEPENDS ${DEPS})
-
-add_custom_target(python_bindings ALL DEPENDS ${OUTPUT})
-
-
 install(TARGETS python-libtorrent DESTINATION "${Python3_SITEARCH}")
-install(DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/libtorrent.egg-info" DESTINATION "${Python3_SITEARCH}")
+
+if (python-egg-info)
+	set(SETUP_PY_IN "${CMAKE_CURRENT_SOURCE_DIR}/setup.py.cmake.in")
+	set(SETUP_PY    "${CMAKE_CURRENT_BINARY_DIR}/setup.py")
+	set(OUTPUT      "${CMAKE_CURRENT_BINARY_DIR}/timestamp")
+	set(DEPS        python-libtorrent "${SETUP_PY}")
+
+	configure_file(${SETUP_PY_IN} ${SETUP_PY} @ONLY)
+
+	add_custom_command(OUTPUT ${OUTPUT}
+	                   COMMAND ${Python3_EXECUTABLE} ${SETUP_PY} build -b "${CMAKE_CURRENT_SOURCE_DIR}"
+	                   COMMAND ${Python3_EXECUTABLE} ${SETUP_PY} egg_info -b "${CMAKE_CURRENT_SOURCE_DIR}"
+	                   COMMAND ${CMAKE_COMMAND} -E touch ${OUTPUT}
+	                   DEPENDS ${DEPS})
+
+	add_custom_target(python_bindings ALL DEPENDS ${OUTPUT})
+
+	install(DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/libtorrent.egg-info" DESTINATION "${Python3_SITEARCH}")
+endif()
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -1,7 +1,8 @@
+cmake_minimum_required(VERSION 3.17.0 FATAL_ERROR)
+
 # To build python bindings we need a python executable and boost python module. Unfortunately,
 # their names might not be interlinked and we can not implement a general solution.
-# The code below assumes default boost installation, when the module for python 2 is named
-# 'python' and the module for python 3 is named 'python3'.
+# The code below assumes default boost installation, when the module for python 3 is named 'python3'.
 # To customize that one can provide a name for the Boost::python module via
 # 'boost-python-module-name' variable when invoking cmake.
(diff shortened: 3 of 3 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12.0 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.17.0 FATAL_ERROR)
```
zeule, line 1: Are you sure?

zeule:
What's the Python 3 status on MacOS?

FranciscoPombal:
@zeule AFAIK macOS does not come with any 3rd party language runtime installed by default other than Python 2, which is also deprecated there and supposed to be for "Apple use only" until it is fully removed. So users are supposed to install it manually, using the official installers or brew or something.

bindings/python/CMakeLists.txt
```diff
@@ -15,12 +14,9 @@
 # Sets _ret to a list of python versions (major.minor) that use the same MSVC runtime as this build does
 # assumes MSVC was detected already
 # See https://en.wikipedia.org/wiki/Microsoft_Visual_C++#Internal_version_numbering
+# See https://devguide.python.org/#status-of-python-branches for supported python versions
 function(_get_compatible_python_versions _ret)
-	if (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 15 AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 16)
```
zeule, line 19: Why?

bindings/python/CMakeLists.txt
```diff
@@ -107,25 +95,10 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
...
-if (NOT "${PYTHON_EXT_SUFFIX}" STREQUAL "None")
-	message(STATUS "Python extension suffix: ${PYTHON_EXT_SUFFIX}")
-	# we mimic the name, created by setuptools
-	# example: libtorrent.cpython-36m-x86_64-linux-gnu.so
-	set_target_properties(python-libtorrent PROPERTIES SUFFIX ${PYTHON_EXT_SUFFIX})
-endif()
+message(STATUS "Python site packages: ${Python3_SITEARCH}")
+
+get_target_property(PYTHON_EXT_SUFFIX python-libtorrent SUFFIX)
+message(STATUS "Python extension suffix: ${PYTHON_EXT_SUFFIX}")
```
zeule, line 101: This was more of a test message, maybe it's unneeded now?

CMakeLists.txt
FranciscoPombal: @zeule Only needed for the changes introduced by the second commit. See https://cmake.org/cmake/help/v3.17/release/3.17.html#modules (the part about `FindPython3`).

If it were possible to set the minimum required version just for the python part of the build that would be ideal, but AFAIK that's not possible, so the version requirement has to be bumped for the whole project.

bindings/python/CMakeLists.txt
FranciscoPombal: In the end it can be removed, sure, but for now it is still useful for debugging purposes.

bindings/python/CMakeLists.txt
FranciscoPombal: @zeule Python <3.5 has EOL'd. If support for older versions is still desired, I can add that later. To make things easier though, I'd like to get everything working first with just Python 3 (see my concerns in the OP), and then go from there.

CMakeLists.txt
NotTsunami: You can set `cmake_minimum_required(VERSION 3.17.0)` in `bindings/python/CMakeLists.txt` but 3.12->3.17 is still a huge jump, that might be better for master branch instead. People shouldn't _need_ to be bleeding edge to build libtorrent.

zeule: @FranciscoPombal  uses `WITH_SOABI` option to `python_add_module()`, which appeared in 3.17. Without the option this PR makes little sense.

NotTsunami: Which is also used within `bindings/python/CMakeLists.txt`, so you can still get away with setting the minimum in that file versus bumping it globally.

FranciscoPombal: @NotTsunami

> You can set `cmake_minimum_required(VERSION 3.17.0)` in `bindings/python/CMakeLists.txt`

You're right, it works after all. Fixed.

> but 3.12->3.17 is still a huge jump, that might be better for master branch instead. People shouldn't need to be bleeding edge to build libtorrent.

the version bump is required for the changes in the second commit, and only when building python bindings. As for the bleeding edge argument, I don't think it is a convincing one. I think there is no good reason for anyone _not_ to use the latest CMake version. See a previous comment I made about that: https://github.com/qbittorrent/qBittorrent/pull/12480#issuecomment-612514368, which explains my reasoning in detail.

arvidn:
cmake version: https://travis-ci.org/github/arvidn/libtorrent/jobs/678690372#L787

FranciscoPombal:
@arvidn 
Seems fairly easy to support the latest version, thanks to the APT repo:
https://travis-ci.com/github/Farwaykorse/test_Travis_CMake/jobs/213053641/config

Kolcha:
> What's the Python 3 status on MacOS?

macOS 10.15 finally comes with Python 3, it has Python 3.7, but Python 2.7 is also available

FranciscoPombal:
bump, will deal with this later.

FranciscoPombal:
@arvidn 

OK, I'm finally happy with this. There should be no difference in behavior from the current implementation, except for the fact that the egg info is no longer generated by default.

There are now two additional features related to the python bindings:
- generate egg info and install it (OFF by default)
- install to system site packages instead of the prefix, (OFF by default, which preserves current behavior)

Let me know if you want me to squash/reword/reorder the commits. Once this is in, I'll continue with https://github.com/arvidn/libtorrent/pull/5197

arvidn:
this seems like a problem: https://travis-ci.com/github/arvidn/libtorrent/jobs/400301284#L1227

FranciscoPombal:
@arvidn 

Ugh. Alright, I'll try to either install a newer CMake version on Travis or just rebase https://github.com/arvidn/libtorrent/pull/5197 on top of this in a test branch and see if it passes.

FranciscoPombal:
> @arvidn
> 
> Ugh. Alright, I'll try to either install a newer CMake version on Travis or just rebase #5197 on top of this in a test branch and see if it passes.

Bad news: https://github.com/FranciscoPombal/libtorrent/runs/1261779636?check_suite_focus=true

Might be a port bug in vcpkg... I have to try without it.

.travis.yml
```diff
@@ -162,6 +162,9 @@ install:
       sudo apt-add-repository 'deb https://apt.kitware.com/ubuntu/ focal main';
       sudo apt update;
       sudo apt install cmake;
+      sudo mkdir /usr/local/cmake-latest/bin
```
zeule: This works for me:
```
apt:
    sources:
      - sourceline: 'deb https://apt.kitware.com/ubuntu/ bionic main'
        key_url: 'https://apt.kitware.com/keys/kitware-archive-latest.asc'

before_install:
  - |
    if [ "$TRAVIS_OS_NAME" = "linux" ]; then
      # CMake from Kitware is installed in /usr/bin
      # TravisCI installs its own cmake to another location which overrides other installations
      # if they don't call the new binary directly
      alias cmake="/usr/bin/cmake"
    fi
  - |
```

FranciscoPombal: @zeule 

Thanks! However, we will have to wait to see the results, since Travis is so slow and only does one build at a time...

.travis.yml
```diff
@@ -48,6 +48,9 @@ cache:
 # packages list: https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise
 addons:
   apt:
+    sources:
+      - sourceline: 'deb https://apt.kitware.com/ubuntu/ bionic main'
```
zeule, line 52: @FranciscoPombal bionic -> focal

FranciscoPombal: @zeule :+1:

.travis.yml
FranciscoPombal: @zeule 
Sadly, your method did not work: https://travis-ci.com/github/arvidn/libtorrent/jobs/400759993#L1323. Thanks anyway. I will now wait for the results of v3.

zeule: Why did it pool 3.16.8 from the kitware repo? In any case, that alias command it enough, you don't need to mess with directories.

FranciscoPombal: @zeule 

> Why did it pull 3.16.8 from the kitware repo? In any case, that alias command it enough, you don't need to mess with directories.

It didn't, it pulled 3.18.4: https://travis-ci.com/github/arvidn/libtorrent/jobs/400759993#L1218

zeule: Ah, yes. I was wrong. Could you add, please, `alias` call (without parameters) to check that the alias was created?

FranciscoPombal: @zeule I'll do that after getting a result from https://travis-ci.com/github/arvidn/libtorrent/jobs/400784318. I screwed up the `mkdir` invocation (missed the `-p`) so yeah...

FranciscoPombal: The initial method worked out: https://travis-ci.com/github/arvidn/libtorrent/jobs/400784318. Not sure why yours failed, but this is good enough for me.

FranciscoPombal:
@arvidn 

Alright, CMake CI passes now. Here is a run of https://github.com/arvidn/libtorrent/pull/5197 on top of this: https://github.com/FranciscoPombal/libtorrent/runs/1266542035?check_suite_focus=true - it also works, except for debug builds and static linking, which is its own problem that will look into in https://github.com/arvidn/libtorrent/pull/5197.

FranciscoPombal:
@arvidn ping, the remaining CI failure seems unrelated: https://travis-ci.com/github/arvidn/libtorrent/jobs/401080845 (or if it is related, I don't know what it's about).

arvidn (approved):

PR #4583, opened 2020-04-25
disable human readable strings in alerts with TORRENT_DISABLE_ALERT_MSG

arvidn:
The idea is to save a bit more space for builds that don't need to produce human readable logs


sledgehammer999:
This affects all alerts, right? IMO, this is a bad idea. Up until now the builds I provide are with disabled logging (for reduced binary size). But we use the `message()` method of some alerts (usually failed alerts) to give more info to the user.

@Chocobo1 @glassez you may want to look at this too.

glassez:
IMO, logging and alerts is different features. If you want disable messages in regular alerts you should introduce another definition for it, e.g. `TORRENT_DISABLE_ALERT_MESSAGES`.

arvidn:
makes sense. Thanks for the feedback, that's partly why I put this up here. I may abandon this or put it under a different define.

PR #4585, opened 2020-04-26
fix setting lease duration to 0

arvidn:


sledgehammer999:
I confirm that this works.

Issue #2858, opened 2018-03-19
Rewrite the SSL engine part so as to support other SSL libraries, likely GnuTLS


alexandre-janniaux:
Hi,

I'm currently writing support for gnutls in libtorrent, but as I wrote my patch I was thinking libtorrent would need quite a big rework of the SSL part so as to produce a clean interface.
I open this issue to have feedback from other developers about what should be done.

Here is the problem:

libtorrent uses boost API for the transport part, which only support openSSL library, without abstracting the context (it is actually a SSL_CTX* object), and directly uses openSSL functions for certificate verification and TLS sni extension.

It is handled by adding ``#TORRENT_USE_OPENSSL`` surrounding the SSL code. While it is quite handy and provides an easy way to integrate other SSL library, it means that the whole logic has to be written multiple times, tests has to be written multiple times and the code is starting to look like spaghetti, with defines everywhere.

Moreover, the three files mainly using SSL, namely ``ssl_stream.cpp``, ``torrent.cpp`` and ``session_impl.cpp`` are already more than multiple thousand lines long.

As libtorrent is using ``boost::asio``, it is acceptable to provide a boost-like compatibility wrapper around ``ssl::context`` and ``ssl::stream`` classes and write a ``using gnutls::context;`` or ``using boost::asio::ssl::context;`` when appropriate, but I think a wrapper around session context and session initialization, certificate verification, hostname settings and ssl enabling in libtorrent should be made.

Maybe libtorrent can be linked against many multiple SSL libraries at once and the user could choose the SSL engine he wants to use for a libtorrent session thanks to a template parameter, defaulted to openssl.

If you agree with these change, we can discuss what this abstraction should look like and I can make the patch so as to implement it.

aldenml:
I had similar general ideas a while ago, but at the end I concluded that writing a custom transport for a new SSL library was too much effort for what it's used. Are you actually using the "torrents over SSL" feature? or just to support HTTPS trackers?

alexandre-janniaux:
I'd like to use the torrent over SSL feature, i.e. authenticating peers and tracker with a certificate and use weak encryption for exchanges between peers.

About the transport, it's quite verbose with boost interface but I believe it can be simplified a lot. It is actually the more factorized part of the library.

The core library will also benefit from moving SSL code to an abstraction, because the code will be a lot simpler and shorter. 

[...]

PR #4588, opened 2020-04-27
Refactor SSL engine and add optional GnuTLS support

paullouisageneau:
This PR refactors SSL handling to hide OpenSSL specifics and introduces an option to compile against GnuTLS instead of OpenSSL.

The motivation is to ease integration by allowing to avoid different TLS backends in a single project as well as potential license issues with OpenSSL (the OpenSSL license is an Apache License 1.0 which is incompatible with the GPL).

GnuTLS support can be enabled with the compilation flag `TORRENT_USE_GNUTLS`. SSL code is made generic where possible (for both trackers and SSL torrents) and SSL tests are modified accordingly.

The GnuTLS boost.asio wrapper is imported from [paullouisageneau/boost-asio-gnutls](https://github.com/paullouisageneau/boost-asio-gnutls) as submodule.

This is based on the preliminary work of @alexandre-janniaux and implements https://github.com/arvidn/libtorrent/issues/2858


alexandre-janniaux:
Hi, thank you for integrating my work! As they were WIP, I can rewrite my commits or you can rewrite and author them to your name with an additional `Co-authored-by` if you prefer rewriting them yourself.

arvidn:
This is quite a patch. I think there is quite a lot of interest in being able to use GnuTLS instead of OpenSSL. I'm a little bit hesitant to put the GnuTLS stream wrapper in libtorrent itself. It would seem more reasonable to either contribute it to boost.asio, or put it in a separate repository that libtorrent could pull in.

That said, obviously there would still need to be new abstractions in libtorrent itself.

Do you know if boost.asio has unit tests for the openssl stream? If so, hooking this stream wrapper into that test would probably make a lot of sense too.

paullouisageneau:
@alexandre-janniaux I'll rewrite and author them as you suggest!

@arvidn I'm not sure about contributing it to boost.asio for now, even if it completely covers the use case here, it is not a full featured wrapper. However putting it in another repository and only letting the libtorrent abstractions here makes sense. It could still be merged in boost.asio one day anyway given enough interest and development.

I'll start working on moving the wrapper and see about boost.asio's unit tests. In the meantime feel free to comment on the rest.

include/libtorrent/ssl/context.hpp
```diff
@@ -165,9 +165,8 @@ class TORRENT_EXTRA_EXPORT context : public boost::asio::ssl::context
 
 	error_code set_verify_dn(std::string name, error_code& ec)
 	{
-		using namespace std::placeholders;
 		m_impl->expected_dn = std::move(name);
-		set_verify_callback(std::bind(&context::verify_dn, this, _1, _2), ec);
+		set_verify_callback(std::bind(&context::verify_dn, this, std::placeholders::_1, std::placeholders::_2), ec);
```
alexandre-janniaux, line 169: This could probably use some local `use std::placeholders` !

paullouisageneau: @alexandre-janniaux That's right, I used it at first but eventually reverted because it caused an ambiguity with `boost::placeholders` in some of the CI environment (for instance the macOS build on travis)...

paullouisageneau:
@arvidn I've moved the GnuTLS wrapper in the independant repository [paullouisageneau/boost-asio-gnutls](https://github.com/paullouisageneau/boost-asio-gnutls) as you suggested. I added unit tests equivalent to the ones of asio/ssl, which are compile-time tests, I'm looking at the proper way to add runtime tests also.

include/libtorrent/ssl/context.hpp
```diff
@@ -0,0 +1,279 @@
...
+
+#if defined TORRENT_USE_OPENSSL
+namespace base = boost::asio::ssl;
+class stream_base;
+#elif defined TORRENT_USE_GNUTLS
+namespace base = boost::asio::gnutls;
+using base::stream_base;
+#endif
+
+class TORRENT_EXTRA_EXPORT context : public base::context
```
arvidn, line 110: It would seem more appropriate to just make `context` a typedef to either the boost asio one or the gnutls one. As far as I can tell, all OpenSSL/GnuTLS specific code could still be done in free functions (ideally implemented in a .cpp file)

include/libtorrent/ssl/context.hpp
```diff
@@ -0,0 +1,279 @@
...
+		m_impl->servername_callback = std::move(cb);
+		return ec;
+#else // OPENSSL_VERSION_NUMBER
+		return ec = {boost::asio::error::operation_not_supported, error::get_ssl_category()};
+#endif
+	}
+#endif
+
+private:
+	static void global_init();
```
arvidn, line 226: this `global_init()` function doesn't seem all that useful. Couldn't you just do the openSSL initialization in a global constructor, as it used to be? Now it looks like you run it every time a `context` is created which happens multiple times. It also doesn't seem to benefit from being a member function here.

include/libtorrent/ssl/context.hpp
```diff
@@ -0,0 +1,279 @@
...
+		// Warning: returns the number of certificates processed or a negative error code on error
+		int ret = gnutls_certificate_set_x509_trust_mem(native_handle()
+				, &ca
+				, format == pem ? GNUTLS_X509_FMT_PEM : GNUTLS_X509_FMT_DER);
+		return ec = ret >= 0 ? error_code() : error_code(ret, error::get_ssl_category());
+#endif
+	}
+
+#if defined TORRENT_USE_OPENSSL
+	error_code set_servername_callback(std::function<bool(stream_base& s, std::string name)> cb, error_code& ec)
```
arvidn, line 205: does this not have an implementation for GnuTLS?
Either way, this used to be a free function with the implementation hidden in a .cpp file, I think it's desirable to put everything that needs to include an openssl header in a .cpp file. I at least had attempted to do so previously, with the `openssl.cpp`

src/openssl.cpp
```diff
@@ -1,82 +0,0 @@
...
-void openssl_set_tlsext_hostname(SSL* s, char const* name)
-{
-#if OPENSSL_VERSION_NUMBER >= 0x90812f
-	SSL_set_tlsext_host_name(s, name);
-#endif
-}
-
-#if OPENSSL_VERSION_NUMBER >= 0x90812f
-
-void openssl_set_tlsext_servername_callback(SSL_CTX* ctx
```
arvidn, line 54: I haven't thought this through all the way, but it seems like you could keep these functions and just make them take an `lt::ssl::context&` instead of the openssl specific type.

arvidn: oh, I guess the callback function signature is still openssl specific. If there isn't a GnuTLS implementation for this yet though, maybe that's OK for a first pass.

include/libtorrent/ssl/context.hpp
```diff
@@ -0,0 +1,279 @@
...
+namespace {
+	// openssl requires this to clean up internal
+	// structures it allocates
+	struct cleanup
+	{
+#ifdef TORRENT_MACOS_DEPRECATED_LIBCRYPTO
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#endif
+		~cleanup() { CRYPTO_cleanup_all_ex_data(); }
```
arvidn, line 264: I would prefer this code also lives in a .cpp file. `session_impl.cpp` is probably *not* the best place, but some counterpart to the previous `openssl.cpp` would make sense I think

include/libtorrent/ssl/stream.hpp
```diff
@@ -0,0 +1,217 @@
...
+	void init_base();
+
+private:
+	static int ex_index;
+	std::shared_ptr<context::impl> m_context_impl;
+	friend class context;
+};
+
+template<typename NextLayer>
+class stream : public base::stream<NextLayer>, public stream_base
```
arvidn, line 124: is there a reason `stream` cannot be a typedef to either `boost::asio::ssl::context` or the gnutls counterpart?

src/openssl.cpp
```diff
@@ -1,82 +0,0 @@
...
-
-namespace libtorrent {
-namespace aux {
-
-#ifdef TORRENT_USE_OPENSSL
-
-// all of OpenSSL causes warnings, so we just have to disable them
-#include "libtorrent/aux_/disable_warnings_push.hpp"
-
-void openssl_set_tlsext_hostname(SSL* s, char const* name)
```
arvidn, line 45: I think having this free function here (or in some .cpp file) is really nice. It hides the openssl/gnutls includes, to not leak into a dependency passed on to clients of libtorrent. It keeps the places where the warnings are disabled to a single place, because of these openssl macros.

There could be another version of this that take whatever the native handle is for a GnuTLS connection. the counterpart to `SSL*`.

src/socket_type.cpp
```diff
@@ -114,33 +105,17 @@ namespace aux {
 		}
 #undef CASE
 
-#if OPENSSL_VERSION_NUMBER >= 0x90812f
-		if (ctx)
-		{
-			aux::openssl_set_tlsext_servername_callback(ctx, nullptr);
-			aux::openssl_set_tlsext_servername_arg(ctx, nullptr);
```
arvidn, line 121: I have a feeling clearing this callback may be important

src/socket_type.cpp
```diff
@@ -88,23 +85,17 @@ namespace aux {
...
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
 #endif
 		// for SSL connections, make sure to authenticate the hostname
 		// of the certificate
 #define CASE(t) case socket_type_int_impl<ssl_stream<t>>::value: \
-		s.get<ssl_stream<t>>()->set_verify_callback( \
-			boost::asio::ssl::rfc2818_verification(hostname), ec); \
-		ssl = s.get<ssl_stream<t>>()->native_handle(); \
-		ctx = SSL_get_SSL_CTX(ssl); \
+		s.get<ssl_stream<t>>()->set_host_name(hostname, ec); \
```
arvidn, line 96: I think it would be nice if the gnutls classes were compatible with boost.asio's ones as much as possible. iirc, asio actually has this verify callback with a generic signature, not dependent on OpenSSL

src/socket_type.cpp
```diff
@@ -88,23 +85,17 @@ namespace aux {
...
 		break;
 
-		SSL* ssl = nullptr;
-		SSL_CTX* ctx = nullptr;
-
 		switch(s.type())
```
arvidn, line 99: one relatively big change in master is that `socket_type` now behaves more like a `boost::variant`, so instead of these switch statements and saving off SSL pointers, there are visitors. So I think this could be done a bit cleaner in the most recent `master` branch

arvidn: specifically, where the concrete type of the native handles don't necessarily need to be mentioned, but just passed in to free functions that have two implementations, one for openssl and one for GnuTLS

src/ssl/context.cpp
```diff
@@ -0,0 +1,287 @@
...
+{
+	// this is needed for openssl < 1.0 to decrypt keys created by openssl 1.0+
+#if !defined(OPENSSL_API_COMPAT) || (OPENSSL_API_COMPAT < 0x10100000L)
+	OpenSSL_add_all_algorithms();
+#else
+	OPENSSL_init_crypto(OPENSSL_INIT_ADD_ALL_CIPHERS | OPENSSL_INIT_ADD_ALL_DIGESTS, nullptr);
+#endif
+}
+
+bool context::verify_dn(std::string const& expected, bool preverified, boost::asio::ssl::verify_context& ctx) {
```
arvidn, line 87: This function is specific to running bittorrent over SSL. I don't think it belongs as a member function of the SSL context. I think it would make more sense for it to be a free function (or probably two free functions, on for OpenSSL and one for GnuTLS). But I would think `torrent.cpp` is a pretty reasonable place for it still.

src/ssl/context.cpp
```diff
@@ -0,0 +1,287 @@
...
+
+#if DEBUG_SSL_VERIFY
+#include <cstdarg> // for va_list
+#include <cstdio> // for vsnprintf
+#endif
+
+namespace {
+
+#if DEBUG_SSL_VERIFY
+void debug_log(char const* fmt, ...)
```
arvidn, line 49: this used to print logsas alerts, and make them available to clients. I see the verification functions used to be member functions on `torrent`, was the alert the only thing it needed the torrent object for?

src/ssl/stream.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include "libtorrent/ssl/stream.hpp"
+
+namespace libtorrent {
+namespace ssl {
+
+#ifdef TORRENT_USE_OPENSSL
+void stream_base::init_base()
+{
+    if (ex_index < 0) ex_index = SSL_get_ex_new_index(0, NULL, NULL, NULL, NULL);
+    SSL_set_ex_data(native_handle(), ex_index, this); // expected by contex
```
arvidn, line 44: what does this do? I think this is a new call

src/torrent.cpp
```diff
@@ -1530,141 +1527,27 @@ bool is_downloading_state(int const st)
...
-#if !defined(OPENSSL_API_COMPAT) || (OPENSSL_API_COMPAT < 0x10100000L)
-		OpenSSL_add_all_algorithms();
-#else
-		OPENSSL_init_crypto(OPENSSL_INIT_ADD_ALL_CIPHERS | OPENSSL_INIT_ADD_ALL_DIGESTS, nullptr);
-#endif
-
 		// create the SSL context for this torrent. We need to
 		// inject the root certificate, and no other, to
 		// verify other peers against
 #if BOOST_VERSION >= 106400
```
arvidn, line 1536: in the latest `master`, boost older than 1.64 is no longer supported, so this was simplified a bit

src/torrent.cpp
```diff
@@ -1675,54 +1558,28 @@ bool is_downloading_state(int const st)
 		// the verification function verifies the distinguished name
 		// of a peer certificate to make sure it matches the info-hash
 		// of the torrent, or that it's a "star-cert"
-		ctx->set_verify_callback(std::bind(&torrent::verify_peer_cert, this, _1, _2), ec);
```
arvidn, line 1678: I think it would be nicer if the gnutls context would have `set_verify_callback()` and `verify_dn` would be a free function (or two free functions, but only one would be in use at a time).

I think it's an improvement to decouple it from being a `torrent` member function as you have done, but I think it should be decoupled from the context too.

arvidn:
one part of this patch that I think could land in `master` right away is the `TORRENT_USE_OPENSSL` and `TORRENT_USE_SSL` macro change. However, what do you think of making `TORRENT_USE_SSL` be defined to 0 or 1 and tested like `#if TORRENT_USE_SSL` instead of `#ifdef`?

I think in general that's a better practice, at least for internal configuration macros, as the compiler can warn on misspelled macros.

include/libtorrent/ssl/context.hpp
paullouisageneau: A significant amount of code can be moved to .cpp files for sure. 
Free functions would have a drawback I think: I see no way to implement generic functions for `set_servername_callback` (because you need to keep the generic callback somewhere and openssl takes a raw function pointer) and `get_context` because you need to retrieve the wrapped context somehow).

include/libtorrent/ssl/context.hpp
paullouisageneau: The init was done every time a torrent context was created before: https://github.com/arvidn/libtorrent/blob/48627d7dfe47fb50a99fbd429de25bebec129dfc/src/torrent.cpp#L1636
I'll move it in a global constructor just like the cleanup is in a global destructor.

include/libtorrent/ssl/context.hpp
paullouisageneau: This is already implemented in the GnuTLS wrapper itself (I put it there since for GnuTLS, SNI is not an extension API, it's in the core lib).
I'll put everything I can in .cpp files.

src/openssl.cpp
paullouisageneau: Yes, that's exactly the problem, `SSL*` is OpenSSL specific, GnuTLS uses opaque handles and the native callback signatures are totally different.

include/libtorrent/ssl/stream.hpp
paullouisageneau: The only reason is the implementation of `get_context` in that case. However, I think it's doable by storing the reference to it in ex data of the corresponding `SSL_CTX`.

src/openssl.cpp
paullouisageneau: The set hostname function per se is no problem. If I manage to implement the callbacks with free functions then this one will follow.

src/socket_type.cpp
paullouisageneau: The gnutls classes are compatible, for instance the signature is the same for `set_verify_callback`. The issue is that the `verify_context` is just a dummy container for an OpenSSL specific type.
However, I see now that I could add a `gnutls::rfc2818_verification` for this case.

src/socket_type.cpp
paullouisageneau: I see, that'd be nice. I'm not 100% sure this is doable with free functions for now because of the callbacks but I'll look into it.

src/ssl/context.cpp
paullouisageneau: Yes it was the only thing. So the alerts were only available in debug mode by design?

src/ssl/context.cpp
paullouisageneau: I'll move them there then.

src/ssl/stream.cpp
paullouisageneau: This is OpenSSL's way of setting a user pointer. It is used in `ssl::context` to retrieve a pointer to the `stream_base` from the `SSL*` native handle in the servername callback.

src/torrent.cpp
paullouisageneau: OK I'll see what I can do.

include/libtorrent/ssl/context.hpp
arvidn: oh, I see. I think I confused it with how I used to initialize openssl a long time ago, where it was in the same global static-storage duration object that does the cleanup ( https://github.com/arvidn/libtorrent/blob/master/src/session_impl.cpp#L172 ). It's possible I moved it to be done more often for a reason. It could be risky to call it over and over again though

include/libtorrent/ssl/context.hpp
arvidn: but boost.asio's `ssl::context` already keeps that function object. You just have to do the same thing in the gnutls counterpart, no?

paullouisageneau: It does for `set_verify_callback`, but it has no support for servername callback sadly.

include/libtorrent/ssl/context.hpp
arvidn: Whether it's in the core API or not, the API *is* a callback still, right? I don't think it's important that callback itself is generic. Or rather, it's not clear to me the additional abstraction is worth making this callback SSL-library-agnostic.

src/ssl/context.cpp
arvidn: it was enabled unless `TORRENT_DISABLE_LOGGING` was defined, which actually makes it available by default (but only when enabled in the `alert_mask`). I don't really have strong opinions about preserving this logging though. I imagine I added it while trouble shooting certs issues in SSL swarms

include/libtorrent/ssl/context.hpp
paullouisageneau: I think you are right, I'll reduce the abstraction by having two versions of the callback.

paullouisageneau:
@arvidn I've refactored everything into free functions, I think a managed to do something clean and generic enough. I also changed `TORRENT_USE_SSL` to be defined to be either 0 or 1 as suggested. Would you like me to extract this part and open a new PR targeting master?

include/libtorrent/proxy_base.hpp
```diff
@@ -124,6 +124,18 @@ class proxy_base
 		m_sock.async_write_some(buffers, handler);
 	}
 
+#if BOOST_VERSION >= 106600 && !defined TORRENT_BUILD_SIMULATOR
```
arvidn, line 127: I imagine I would have to add `async_wait` to uTP connections as well. and I really should add it to libsimulator (but I suppose the simulator doesn't support SSL right now anyway).

I take it this is used by the gnutls wrapper?

include/libtorrent/utp_stream.hpp
```diff
@@ -493,6 +518,33 @@ struct TORRENT_EXTRA_EXPORT utp_stream
 		issue_write();
 	}
 
+#if BOOST_VERSION >= 106600
+	// Compatiblity with the async_wait method introduced in boost 1.66
```
arvidn, line 522: oh, I see, you added it :)

arvidn: I think this function, along with the `proxy_base` support for `async_wait` is a good candidate for a separate PR (that can land ahead of this big one, to make it just a little bit smaller)

include/libtorrent/utp_stream.hpp
```diff
@@ -420,11 +421,35 @@ struct TORRENT_EXTRA_EXPORT utp_stream
...
+			add_write_buffer(i->data(), i->size());
+#if TORRENT_USE_ASSERTS
+			buf_size += i->size();
+#endif
+		}
+		std::size_t ret = write_some(true);
+		TORRENT_ASSERT(ret <= buf_size);
+		if(ret == 0)
+		{
+			ec = boost::asio::error::would_block;
```
arvidn, line 449: this isn't quite right. if `write_some()` returns 0, it just means it couldn't send any packets *right now*. However, since you called `add_write_buffer()` already, the socket is remembering those pointers and will read from them later, when it *can* send more packets on the socket.

The normal behavior of a blocking write is that the socket/kernel will copy some of the data into its own send buffer, which later will be sent. If that send buffer is full, the call fails with `EWOULDBLOCK`. There isn't a send buffer in the uTP sockets though. Since they only support asynchronous writes, all they support is to remember where the user's buffers are, and read directly from them.

Have you seen this work? does the ssl unit test pass with the GnuTLS backend?

I can't think of a simple way of supporting this. I think we'd either need a send buffer, or have this never buffer anything but just consume any data that can be sent as packets *right now* (which most likely will be 0 most of the time)

src/utp_stream.cpp
```diff
@@ -1084,25 +1084,43 @@ std::size_t utp_stream::read_some(bool const clear_buffers)
 	return ret;
 }
 
+std::size_t utp_stream::write_some(bool const clear_buffers)
```
arvidn, line 1087: I think this patch (along with support for syncrhonous write) is also a good candidate for a separate PR that could land ahead of this one

arvidn:
I think this is looking good. I haven't reviewed everything yet, but what I like what I've seen so far.

In the interest of making it more manageable to review the core of this patch, I think it would be beneficial to break out some chunks.

The `TORRENT_USE_SSL` define is a good candidate I think. The broken-out patch should probably be before GnuTLS is introduced, so mostly a new internal name (`TORRENT_USE_SSL`) with the same external name (the one build systems define, `TORRENT_USE_OPENSSL`).

I point out two other candidates too, which seem somewhat independent

arvidn:
I just noticed this is targeting `RC_1_2`. I'm not sure that's possible, ABI wise. It might, but it would also create quite a headache merging this into master.

Do you think it's important it makes it into `RC_1_2`?
I try to limit changes to bug fixes, and simple changes.

include/libtorrent/utp_stream.hpp
paullouisageneau: You are right, the `write_some` I implemented only supports non-blocking immediate writes due to the limitations you mentioned (similar to socket behavior with the non-blocking option set to true), which is OK as long as you don't mix it with asynchronous writes of course. I'll comment about that.
Luckily, it's only what I needed for the GnuTLS wrapper. It allows to make it much simpler as relying on async writes would require buffering data on the wrapper level.

include/libtorrent/proxy_base.hpp
paullouisageneau: Yes it is. I observed strange behaviors with null buffers, so since the are now deprecated anyway, I used the recommended `async_wait` which worked fine.

paullouisageneau:
It's not important that it makes it into `RC_1_2`, therefore I'll target master and break it down into 3 parts as you suggest: `TORRENT_USE_SSL` renaming, `async_wait` and `read_some` methods, and eventually the actual GnuTLS support.

arvidn:
in `master`, there's a new abstraction to represent the `info-hash` of a torrent, with bittorrent v2 support, it may be both sha-1 and sha-256. In most cases it can be handled the same though.
https://ci.appveyor.com/project/arvidn/libtorrent/builds/32746464/job/y4i5agrgi5x2m1px#L225

paullouisageneau:
@arvidn Thanks for merging the prerequisites, I rebased on master and fixed the info_hash issue.

arvidn:
great! I'm hoping to do a more careful review later today. One thing I think might be a good idea is to add a build in `.travis.yml` with `crypto=gnutls`, to demonstrate it working. I imagine getting it to work on windows and MacOS might be a bit challenging, but a linux build would be doable I hope.

does that make sense?

paullouisageneau:
Yes, a Linux build should be doable, I'm adding it.

CMakeLists.txt
```diff
@@ -651,35 +652,67 @@ if(Iconv_FOUND)
...
 
-	target_link_libraries(torrent-rasterbar PUBLIC OpenSSL::SSL)
-	target_compile_definitions(torrent-rasterbar
-		PUBLIC
-			TORRENT_USE_LIBCRYPTO
-			TORRENT_USE_OPENSSL
-			TORRENT_SSL_PEERS
-			OPENSSL_NO_SSL2)
-	target_sources(torrent-rasterbar PRIVATE src/pe_crypto)
+if(gnutls OR NOT OPENSSL_FOUND)
```
arvidn, line 688: I'm not quite following this. `gnutls` is set by the first call to `find_public_dependency(GnuTLS)`, is it?

If it's set and `OPENSSL_FOUND` is *not* set, then you call `find_public_dependency(GnuTLS)` again? Why does it need to be called twice?

arvidn: are the first calls to `find_public_dependency(GnuTLS)` and `set_package_properties(GnuTLS, ...` lef-overs that aren't needed anymore?

paullouisageneau: No, `gnutls` is actually the cmake option set by the user (disabled by default). So the logic is to use GnuTLS if either set by the user or if OpenSSL is not found.

paullouisageneau: However the second call is a rebase mistake I think, it's indeed useless :facepalm:

CMakeLists.txt
```diff
@@ -651,35 +652,67 @@ if(Iconv_FOUND)
...
+		install(DIRECTORY deps/asio-gnutls/include/ DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+	elseif(gnutls)
+		message(FATAL_ERROR "GnuTLS library not found")
+	endif()
 endif()
 
 if(OPENSSL_FOUND)
 	target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_LIBCRYPTO)
-else(OPENSSL_FOUND)
+	target_sources(torrent-rasterbar PRIVATE src/pe_crypto)
```
arvidn, line 705: `pe_crypto` implements the bittorrent "encryption"/obfuscation protocol. It uses diffie hellman and RC4. It doesn't require libcrypto, but I think it will use its routines if it's available. It's already being added below, if `encryption` is enabled. I don't think it needs to be added here, or in the else-clause

Jamfile
```diff
@@ -407,6 +411,43 @@ rule openssl-include-path ( properties * )
...
+# usage-requirement for the gnutls-related libraries
+rule gnutls-include-path ( properties * )
+{
+	local GNUTLS_INCLUDE = [ feature.get-values <gnutls-include> : $(properties) ] ;
+
+	if <target-os>darwin in $(properties) && $(GNUTLS_INCLUDE) = ""
+	{
+		# on macOS, default to pick up gnutls from the homebrew installation
+		# brew install gnutls
+		GNUTLS_INCLUDE = /usr/local/opt/openssl/include ;
```
arvidn, line 442: this include looks wrong. did you test this on MacOS?

include/libtorrent/ssl_stream.hpp
```diff
@@ -34,14 +34,14 @@ POSSIBILITY OF SUCH DAMAGE.
...
 
-#include "libtorrent/config.hpp" // for TORRENT_USE_SSL
-
-#if TORRENT_USE_SSL
-
 #include "libtorrent/socket.hpp"
 #include "libtorrent/error_code.hpp"
 #include "libtorrent/io_context.hpp"
-#include "libtorrent/aux_/openssl.hpp"
+#include "libtorrent/ssl.hpp"
```
arvidn, line 40: should the `ssl.hpp` header really be included regardless of whether `TORRENT_USE_SSL` is 1?

include/libtorrent/ssl_stream.hpp
```diff
@@ -34,14 +34,14 @@ POSSIBILITY OF SUCH DAMAGE.
...
+
+#if TORRENT_USE_SSL
```
arvidn, line 42: actually, it's not clear to me why you moved this down. if `TORRENT_USE_SSL == 0`, this whole file will turn into nothing, right?

src/hasher.cpp
```diff
@@ -194,7 +193,7 @@ TORRENT_CRYPTO_NAMESPACE
 
 	hasher256& hasher256::operator=(hasher256 const& h) &
 	{
-		if (this == &h) return;
+		if (this == &h) return *this;
```
arvidn, line 196: ouch

Jamfile
paullouisageneau: No, I don't have such a machine on hand. I'll try on travis when the Linux build works.

include/libtorrent/ssl_stream.hpp
paullouisageneau: Indeed, I just moved it down because it was shorter than including `config.hpp` beforehand.

src/session_impl.cpp
```diff
@@ -452,51 +452,45 @@ void apply_deprecated_dht_settings(settings_pack& sett, bdecode_node const& s)
...
-		const char* servername = SSL_get_servername(s, TLSEXT_NAMETYPE_host_name);
-
-		if (!servername || std::strlen(servername) < 40)
-			return SSL_TLSEXT_ERR_ALERT_FATAL;
+		if (name.size() < 40)
+			return false;
 
 		info_hash_t info_hash;
-		bool valid = aux::from_hex({servername, 40}, info_hash.v1.data());
+        bool valid = aux::from_hex({name.c_str(), 40}, info_hash.v1.data());
```
arvidn, line 467: spaces instead of tabs here, for indentation

Jamfile
arvidn: it probably shouldn't say "openssl" either way :)

src/ssl.cpp
```diff
@@ -0,0 +1,257 @@
...
+	SSL_CTX_set_cert_store(c, cert_store);
+
+#elif defined TORRENT_USE_GNUTLS
+	std::string str(pem);
+    gnutls_datum_t ca;
+    ca.data = reinterpret_cast<unsigned char*>(const_cast<char*>(str.c_str())); // must be null terminated
+    ca.size = unsigned(str.size());
+
+	// Warning: returns the number of certificates processed or a negative error code on error
+	int ret = gnutls_certificate_set_x509_trust_mem(c.native_handle(), &ca, GNUTLS_X509_FMT_PEM);
```
arvidn, line 90: I thought it looked a bit weird that the `data` field needed to be null terminated, given that you pass in the length as well. So, I guess [this](https://www.gnutls.org/manual/html_node/Core-TLS-API.html#index-gnutls_005fcertificate_005fset_005fx509_005ftrust_005fmem) is the authoritative documentation.

I don't see it say anything about it needing to be 0-terminated. But I see `ca` is a *list* of certificates. How does it know how many certs are in the list? You pass in 1, but what stops the function from assuming there are more after? Are you sure it's not the list of certs that's null terminated? Say, the last `data` field is `nullptr` to indicate the end of the list.

If that is the case, you wouldn't have to copy the certificate above.

src/torrent.cpp
```diff
@@ -1384,16 +1383,87 @@ bool is_downloading_state(int const st)
 #endif
 
 #ifdef TORRENT_SSL_PEERS
+	void torrent::init_ssl(string_view cert)
```
arvidn, line 1386: it looks like `torrent::init_ssl()` moved up, so it no longer lines up with where the previous function was. This makes it a bit more difficult to see what changed in here. Would you mind moving it down to where it currently is in `master`?

src/ssl.cpp
```diff
@@ -0,0 +1,257 @@
...
+				{
+					return cb(s, name, arg);
+				}
+				, ec);
+	else
+		c.set_server_name_callback(nullptr);
+#endif
+}
+
+void set_host_name(stream_handle_type s, std::string const& name, error_code& ec)
```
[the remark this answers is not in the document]
paullouisageneau: Yes you are right!

The underlying issue is that GnuTLS does not have the same distinction of context and stream, so the native types exposed by `gnutls::context` and `gnutls::stream` can't be used as their OpenSSL counterpart. Especially for switching contexts, since the `gnutls::context` is actually only a credentials handler, the rest is emulated by the wrapper.

include/libtorrent/ssl.hpp
```diff
@@ -0,0 +1,182 @@
...
+using boost::asio::gnutls::context;
+using boost::asio::gnutls::stream_base;
+using boost::asio::gnutls::stream;
+using boost::asio::gnutls::verify_context;
+using boost::asio::gnutls::rfc2818_verification;
+
+using native_context_type = context::native_handle_type;
+using native_stream_type = stream_base::native_handle_type;
+using context_handle_type = context&;
+using stream_handle_type = stream_base&;
```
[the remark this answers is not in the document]
paullouisageneau: Very good point, I'm changing for pointers.

As you guessed in the next comment, the goal of the handle is not to make callbacks uniform, it is to be able to pass either the native handle or the wrapper itself the to the `ssl` free functions.

src/ssl.cpp
```diff
@@ -0,0 +1,257 @@
...
+	void* user_arg = SSL_CTX_get_ex_data(SSL_get_SSL_CTX(s), server_name_arg_ex_index);
+	return cb(s, name ? std::string(name) : "", user_arg)
+			? SSL_TLSEXT_ERR_OK
+			: SSL_TLSEXT_ERR_ALERT_FATAL;
+}
+
+} // anonymous namespace
+#endif
+
+void set_server_name_callback(context_handle_type c, server_name_callback_type cb, void* arg, error_code& ec)
```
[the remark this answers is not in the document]
paullouisageneau: That would be great, sadly it's not possible since `SSL_CTX_set_tlsext_servername_callback` expects a C-style callback, not a C++ callable.

src/ssl.cpp
```diff
@@ -0,0 +1,257 @@
...
+{
+#if defined TORRENT_USE_OPENSSL
+#if OPENSSL_VERSION_NUMBER >= 0x90812f
```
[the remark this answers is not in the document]
paullouisageneau: OK great!

Jamfile
arvidn: I can try this on mac tomorrow (if I remember)

src/ssl.cpp
paullouisageneau: Oh, that's true. Other gnutls x509 functions require it though, like [gnutls_certificate_set_x509_key_mem](https://www.gnutls.org/manual/html_node/Core-TLS-API.html#index-gnutls_005fcertificate_005fset_005fx509_005fkey_005fmem)

I'm pretty sure the fact that it can be a list in the case of PEM encoding only means you can list different CA in the same string.

src/session_impl.cpp
```diff
@@ -452,51 +452,45 @@ void apply_deprecated_dht_settings(settings_pack& sett, bdecode_node const& s)
...
 
-#if defined TORRENT_SSL_PEERS
+#ifdef TORRENT_SSL_PEERS
 	namespace {
 	// when running bittorrent over SSL, the SNI (server name indication)
 	// extension is used to know which torrent the incoming connection is
 	// trying to connect to. The 40 first bytes in the name is expected to
 	// be the hex encoded info-hash
-	int servername_callback(SSL* s, int*, void* arg)
+	bool ssl_server_name_callback(ssl::stream_handle_type stream_handle, std::string const& name, void* arg)
```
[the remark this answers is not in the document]
paullouisageneau: OK, works for me!

src/ssl.cpp
arvidn: I see. Right, I think I was in C++ mode thinking that there must be a reason it was a pointer instead of being passed by value or reference :)

paullouisageneau:
I fixed the Jamfile, the new travis build with `crypto=gnutls` seems to work.

src/utp_stream.cpp
```diff
@@ -651,10 +651,7 @@ void utp_socket_impl::issue_write()
 	m_written = 0;
 	if (test_socket_state()) return;
 
-	// try to write. send_pkt returns false if there's
-	// no more payload to send or if the congestion window
-	// is full and we can't send more packets right now
-	while (send_pkt());
+	write_some(false);
```
arvidn, line 654: if you revert this, then `write_some()` wouldn't need to take the `bool` parameter, right?
it's not clear that reusing this function here is worth it. it's a single line of code, and a `bool` to disable two other lines of code in there.

test/test_ssl.cpp
```diff
@@ -476,11 +469,11 @@ bool try_connect(lt::session& ses1, int port
 			name += hex_alphabet[rand() % 16];
 
 		std::printf("SNI: %s\n", name.c_str());
-		aux::openssl_set_tlsext_hostname(ssl_sock.native_handle(), name.c_str());
+		ssl::set_host_name(ssl::get_handle(ssl_sock), name, ec);
```
arvidn, line 472:
```suggestion
		ssl::set_host_name(ssl::get_handle(ssl_sock), name, ec);
		TEST_CHECK(!ec);
```

test/test_ssl.cpp
```diff
@@ -465,7 +458,7 @@ bool try_connect(lt::session& ses1, int port
 	{
 		std::string name = aux::to_hex(t->info_hash().v1);
 		std::printf("SNI: %s\n", name.c_str());
-		aux::openssl_set_tlsext_hostname(ssl_sock.native_handle(), name.c_str());
+		ssl::set_host_name(ssl::get_handle(ssl_sock), name, ec);
```
arvidn, line 461:
```suggestion
		ssl::set_host_name(ssl::get_handle(ssl_sock), name, ec);
		TEST_CHECK(!ec);
```

src/utp_stream.cpp
paullouisageneau: Sure, actually I only added it so it is symmetrical with `read_some`.

src/torrent.cpp
```diff
@@ -1384,16 +1382,87 @@ bool is_downloading_state(int const st)
...
+#ifndef TORRENT_DISABLE_LOGGING
+		std::string names;
+		bool match = false;
+#endif
+
+#ifdef TORRENT_USE_OPENSSL
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#pragma clang diagnostic ignored "-Wused-but-marked-unused"
```
arvidn, line 1463: I think you also need:

```
#pragma clang diagnostic ignored "-Wold-style-cast"
#pragma clang diagnostic ignored "-Wzero-as-null-pointer-constant"
```

src/ssl.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+void set_host_name(stream_handle_type s, std::string const& name, error_code& ec)
+{
+#if defined TORRENT_USE_OPENSSL
+	TORRENT_UNUSED(ec);
+#if defined __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wold-style-cast"
+#endif
+	SSL_set_tlsext_host_name(s, name.c_str());
```
arvidn, line 130: I'm getting this warning:
```
../src/ssl.cpp:130:30: warning: cast from 'const char *' to 'char *' drops const qualifier [-Wcast-qual]
```

I think you also need:

```
#pragma clang diagnostic ignored "-Wcast-qual"
```

before this patch, all warnings are suppressed when using these openssl macros, by `#include "libtorrent/aux_/disable_warnings_push.hpp"`

src/session_impl.cpp
```diff
@@ -61,7 +61,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <boost/asio/ts/executor.hpp>
 #include "libtorrent/aux_/disable_warnings_pop.hpp"
 
-#include "libtorrent/aux_/openssl.hpp"
+#include "libtorrent/ssl.hpp"
```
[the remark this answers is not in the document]
paullouisageneau: Oh, indeed. It seems I was quite inattentive when rebasing :sweat_smile:

paullouisageneau:
Hum, this seems platform-specific though, so it's probably linked to how error codes of the socket layer are handled. I'm having a look.

arvidn:
Here's what I know about the test failure:

`boost::asio::read()` fails ([here](https://github.com/arvidn/libtorrent/pull/4588/files#diff-ae80e01a9156b9c10eb22a8606c6fc57R520)) with error "Resource temporarily unavailable" (i.e. `EAGAIN`).

That's strange because the wrapper `boost::asio::read()` is supposed to keep reading until the requested number of bytes have been received, blocking. So I would expect it to *never* fail with this error.

arvidn:
looking into the gnutls `stream.hpp` wrapper. I notice that `gnutls_record_recv()` returns `GNUTLS_E_AGAIN` and you translate that to the error_code counterpart. What's strange about this is that the SSL socket (i.e. gnutls::stream) isn't supposed to have been set to non-blocking mode. (by default sockets are blocking in asio), so the call to `gnutls_record_recv` really should have blocked until some data become available. (but maybe all gnutls APIs are nob-blocking and it expects you to feed it more data in this case).

Looking around a bit in `stream.hpp`, I see a fair number of calls to `m_next_layer.non_blocking()`, which makes me wonder. Does your gnutls wrapper not support blocking mode?
it seems like it, and maybe it doesn't make much sense to support it since libtorrent itself doesn't use blocking mode. This test does though :/

arvidn:
it looks like `read_some()` might be missing some `gnutls_record_get_direction()` and `async_schedule()` in case you get `GNUTLS_E_AGAIN`. But I don't fully understand how the stream class works

src/ssl.cpp
```diff
@@ -0,0 +1,211 @@
...
+#endif
+}
+
+void set_context(stream_handle_type s, context_handle_type c)
+{
+#if defined TORRENT_USE_OPENSSL
+	SSL_set_SSL_CTX(s, c);
+	SSL_set_verify(s
+			, SSL_CTX_get_verify_mode(c)
+			, SSL_CTX_get_verify_callback(c));
```
arvidn, line 147: sorry for nitpick; these two lines are indented one level too many

src/torrent.cpp
```diff
@@ -1384,16 +1382,89 @@ bool is_downloading_state(int const st)
...
+	{
+		// create the SSL context for this torrent. We need to
+		// inject the root certificate, and no other, to
+		// verify other peers against
+#if BOOST_VERSION >= 106400
+		std::unique_ptr<ssl::context> ctx(new ssl::context(ssl::context::tls));
+#else
+		std::unique_ptr<ssl::context> ctx(new ssl::context(ssl::context::tlsv12));
+#endif
+
```
arvidn, line 1395: In the existing `torrent::init_ssl()` there was some error checking here if `!ctx`, which you removed. It doesn't seem like it would make sense to check that here though, so it should be fine to remove it. If the `context` constructor fails, it will throw, to make `operator new` return `nullptr`.

src/torrent.cpp
```diff
@@ -1384,16 +1382,89 @@ bool is_downloading_state(int const st)
 #endif
 
 #ifdef TORRENT_SSL_PEERS
+	void torrent::init_ssl(string_view cert)
+	{
+		// create the SSL context for this torrent. We need to
+		// inject the root certificate, and no other, to
+		// verify other peers against
+#if BOOST_VERSION >= 106400
+		std::unique_ptr<ssl::context> ctx(new ssl::context(ssl::context::tls));
```
arvidn, line 1391: this use `std::make_unique<>` now

src/torrent.cpp
```diff
@@ -1384,16 +1382,89 @@ bool is_downloading_state(int const st)
...
+#else
+		std::unique_ptr<ssl::context> ctx(new ssl::context(ssl::context::tlsv12));
```
arvidn, line 1393: support for old boost is not necessary in `master`

src/torrent.cpp
```diff
@@ -1384,16 +1382,89 @@ bool is_downloading_state(int const st)
...
+		{
+			set_error(ec, torrent_status::error_file_ssl_ctx);
+			pause();
+			return;
+		}
+
+		// if all went well, set the torrent ssl context to this one
+		m_ssl_ctx = std::move(ctx);
+		// tell the client we need a cert for this torrent
+		alerts().emplace_alert<torrent_need_cert_alert>(get_handle());
```
arvidn, line 1437: these 4 lines used to be at the end of the function (after the `#if 0`-block)

src/torrent.cpp
```diff
@@ -1424,13 +1491,12 @@ bool is_downloading_state(int const st)
...
 #endif
 			if (std::strncmp(torrent_name, "*", name_length) == 0
-				|| std::strncmp(torrent_name, m_torrent_file->name().c_str(), name_length) == 0)
+				|| std::strncmp(torrent_name, expected.c_str(), name_length) == 0)
 			{
 #ifndef TORRENT_DISABLE_LOGGING
 				match = true;
 				// if we're logging, keep looping over all names,
 				// for completeness of the log
-				continue;
```
arvidn, line 1433: what happened to this `continue`?

src/torrent.cpp
```diff
@@ -1384,16 +1382,89 @@ bool is_downloading_state(int const st)
...
-	bool torrent::verify_peer_cert(bool const preverified, boost::asio::ssl::verify_context& ctx)
+	bool torrent::verify_peer_cert(std::string const& expected, bool preverified, ssl::verify_context& ctx)
 	{
 		// if the cert wasn't signed by the correct CA, fail the verification
 		if (!preverified) return false;
 
+#ifndef TORRENT_DISABLE_LOGGING
+		std::string names;
+		bool match = false;
+#endif
```
arvidn, line 1457: these 4 lines used to be right before the loop with `sk_GENERAL_NAME_num`. It seems unnecessary to make the patch larger by moving them

src/torrent.cpp
```diff
@@ -1456,130 +1522,80 @@ bool is_downloading_state(int const st)
...
 #endif
-
 			if (std::strncmp(torrent_name, "*", name_length) == 0
-				|| std::strncmp(torrent_name, m_torrent_file->name().c_str(), name_length) == 0)
+				|| std::strncmp(torrent_name, expected.c_str(), name_length) == 0)
 			{
-#ifdef TORRENT_DISABLE_LOGGING
-				return true;
-#else
+#ifndef TORRENT_DISABLE_LOGGING
```
arvidn, line 1528: this was inverted, I don't think that's necessary, or necessarily an improvement

paullouisageneau:
> Looking around a bit in `stream.hpp`, I see a fair number of calls to `m_next_layer.non_blocking()`, which makes me wonder. Does your gnutls wrapper not support blocking mode?
> it seems like it, and maybe it doesn't make much sense to support it since libtorrent itself doesn't use blocking mode. This test does though :/

Thanks, I found the issue. The gnutls wrapper should work fine in blocking mode too, only the underlying socket is set to non-blocking mode for async operations. However, after the handshake I set the socket non-blocking by mistake. For some reason `test_ssl` still passed on my machine without hitting EAGAIN. It should be OK now.

arvidn:
> For some reason test_ssl still passed on my machine without hitting EAGAIN. It should be OK now.

Works on MacOS for me now too!

src/torrent.cpp
paullouisageneau: I put them here rather than putting them twice, since they are shared between OpenSSL and GnuTLS.

arvidn: ah, of course. makes sense

src/torrent.cpp
arvidn: i.e. move `torrent::init_ssl()` down right after `torrent::verify_peer_cert`

paullouisageneau: Oh yes sorry!

arvidn:
I think this looks good. A few things:

* There should probably be an entry in `ChangeLog` saying GnuTLS is supported
* The `Makefile` is used to build a tarball, I think it should pull in the files from the `deps/asiognutls` submodule as well, to make sure that makes it into a source distribution. Similar to how `deps/try_signal` is pulled in.

paullouisageneau:
@arvidn Great! I updated the Makefile to pull files from `deps/asio-gnutls`, added a line to the changelog, and squashed some commits. Is it expected that The Doozer MacOS build does not run?

arvidn:
yeah, the mac on doozer have had problems for a long time (I'm still hopeful it will come back to life, but eventually I will have to remove the doozer file)

CMakeLists.txt
```diff
@@ -651,35 +652,57 @@ if(Iconv_FOUND)
...
+				TORRENT_USE_OPENSSL
+				TORRENT_SSL_PEERS
+				OPENSSL_NO_SSL2)
 	endif()
+endif()
 
-	target_link_libraries(torrent-rasterbar PUBLIC OpenSSL::SSL)
-	target_compile_definitions(torrent-rasterbar
-		PUBLIC
-			TORRENT_USE_LIBCRYPTO
```
arvidn, line 673: I just realized this was removed. It looks like it was redundant though, since it's added again below, if `OPENSSL_FOUND`

CMakeLists.txt
```diff
@@ -651,35 +652,57 @@ if(Iconv_FOUND)
...
+	find_public_dependency(GnuTLS)
+	set_package_properties(GnuTLS
+		PROPERTIES
+			URL "https://www.gnutls.org/"
+			DESCRIPTION "GnuTLS is a free software implementation of the TLS and DTLS protocols"
+			TYPE RECOMMENDED
+			PURPOSE "Provides HTTPS support to libtorrent"
+	)
+	if(GNUTLS_FOUND)
+		target_link_libraries(torrent-rasterbar PUBLIC GnuTLS::GnuTLS)
```
arvidn, line 689: it just occurred to me that `libgcrypt` is a dependency of `libgnutls`, isn't it? Unless I've misunderstood something, it might make sense to "enable" LibGcrypt when building in gnutls mode, like how `libcrypto` is "enabled" when building in openssl mode. This is a refinement that could be done later though, as it's not obvious whether it's a good idea or what the consequences might be.

src/torrent.cpp
```diff
@@ -60,11 +60,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <functional>
 
 #ifdef TORRENT_SSL_PEERS
+#include "libtorrent/ssl.hpp"
```
arvidn, line 63: ssl.hpp is included twice in this file. Once conditionally (here), and then unconditionally a bit further down. I imagine this include is redundant.

src/torrent.cpp
```diff
@@ -1384,16 +1382,25 @@ bool is_downloading_state(int const st)
 #endif
 
 #ifdef TORRENT_SSL_PEERS
-
-	bool torrent::verify_peer_cert(bool const preverified, boost::asio::ssl::verify_context& ctx)
+	bool torrent::verify_peer_cert(std::string const& expected, bool preverified, ssl::verify_context& ctx)
```
arvidn, line 1385: nit: `preverified` lost its `const`. it doesn't look like it needs to be mutable.

src/torrent.cpp
```diff
@@ -1479,37 +1538,20 @@ bool is_downloading_state(int const st)
...
-			return;
-		}
+		std::unique_ptr<ssl::context> ctx(std::make_unique<ssl::context>(ssl::context::tls));
 
-		ctx->set_options(context::default_workarounds
-			| context::no_sslv2
-			| context::single_dh_use);
+		ctx->set_options(ssl::context::default_workarounds
+			| ssl::context::no_sslv2
+			| ssl::context::no_sslv3
```
arvidn, line 1548: SSLv3 is outdated an insecure by now I assume?

src/torrent.cpp
```diff
@@ -1520,54 +1562,25 @@ bool is_downloading_state(int const st)
 		// the verification function verifies the distinguished name
 		// of a peer certificate to make sure it matches the info-hash
 		// of the torrent, or that it's a "star-cert"
-		ctx->set_verify_callback(std::bind(&torrent::verify_peer_cert, this, _1, _2), ec);
+		ctx->set_verify_callback(
+				std::bind(&torrent::verify_peer_cert, this, m_torrent_file->name(), _1, _2)
```
arvidn, line 1566: technically, since `verify_peer_cert` is still a member function now, `m_torrent_file->name()` doesn't need to be passed along here

CMakeLists.txt
paullouisageneau: I put it back here and remove it below.

src/torrent.cpp
paullouisageneau: Indeed, anyway this is somewhat redundant since now the method is set to `tls`.

CMakeLists.txt
paullouisageneau: `libgcrypt` was indeed the cryptographic backend for `libgnutls`, however they changed for `nettle` in the last versions, so there is no guarantee the libs will be present together.

paullouisageneau:
I addressed your few comments, thank you for the approval!

arvidn:
@paullouisageneau thanks! this is impressive work!

paullouisageneau:
@arvidn Thank you!

PR #4595, opened 2020-04-30
undeprecate the upnp_ignore_nonrouters setting

arvidn:
with a slightly different meaning. still default to false and make it mean to ignore routers outside of the local subnet

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* undeprecate upnp_ignore_nonrouters (but refering to devices on our subnet)
 	* increase default tracker timeout
 	* retry failed socks5 server connections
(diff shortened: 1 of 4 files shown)


iz8mbw:
Hi @arvidn 
I can confirm this patch works well both for local router and remote UPnP "server".
I have tested it with qBittorrent in these scenarios:

1. Local router in LAN (router/default gateway in LAN)
2. Remote UPnP "server" outside LAN (example a VPN scenario)

Thank you very much!

arvidn:
thanks for testing!

iz8mbw:
> thanks for testing!

does this patch will be included in libtorrent 1.2.7?
Thank you!

arvidn:
yes

PR #4606, opened 2020-05-02
use validate_hash_request in hash_picker::add_hashes

arvidn:
since there's a fair amount of overlap

### DIFF
--- a/src/hash_picker.cpp
+++ b/src/hash_picker.cpp
@@ -59,6 +59,11 @@ proof_layer  |    x                x
 
 bool validate_hash_request(hash_request const& hr, file_storage const& fs)
 {
+	// limit the size of the base layer to something reasonable
+	// Blocks are requested for an entire piece so this limit
+	// effectivly caps the piece size we can handle. A limit of 8192
+	// corresponds to a piece size of 128MB.
+
 	if (hr.file < file_index_t{0}
 		|| hr.file >= fs.end_file()
 		|| hr.base < 0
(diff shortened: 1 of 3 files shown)


src/hash_picker.cpp
```diff
@@ -241,26 +246,14 @@ bool validate_hash_request(hash_request const& hr, file_storage const& fs)
...
-		if (req.count > 8192)
-			return add_hashes_result(false);
-
-		int const num_layers = file_num_layers(req.file);
-
-		if (req.base >= num_layers || req.proof_layers >= num_layers)
-			return add_hashes_result(false);
-
-		if (req.index + req.count > (1 << (num_layers - req.base)))
+		if (!validate_hash_request(req, m_files))
```
ssiloti, line 249: I think this can be turned into an assert since this is currently only being called via `bt_peer_connection::on_hashes` which already validates the request.

arvidn: as far as I can tell, `bt_peer_connection` doesn't validate the number of hashes. But it *does* call `validate_hash_request()`. In a way it seems more reasonable for `hash_picker` to do the validation though, as it affects its invariant (I think). But on the other hand, an invalid hash request would be detected earlier in `bt_peer_connection`, since it wouldn't have to receive all the hashes. Actually, I imagine this is the main reason `bt_peer_connection` *does* validate it, since it could get stuck in an "infinite" loop receiving hashes otherwise.

arvidn: so, I think you're right. With this patch, the call to `validate_hash_request` could be an assert.

PR #4607, opened 2020-05-02
Make tests python3 compatible. Fixes #4597

### LINKED ISSUE #4597: Tests aren't python3 compatible
Version libtorrent-rasterbar-1.2.6 on FreeBSD 12.1

python-2.7 is EOLed.

Particularly, ```test_http_connection``` fails with python-3.7 seemingly because of python syntax incompatibility:
```
 ===== TESTING: https://non-existent-domain.se/non-existing-file =====
00:00:01.096 expecting: size: -1 status: -1 connected: 0 error: no error
00:00:01.139 < -1 
00:00:01.139 connect_handler_called: 0
00:00:01.139 handler_called: 1
00:00:01.139 status: 0
00:00:01.139 size: 0
00:00:01.139 expected-size: -1
00:00:01.139 error_code: Host not found (authoritative)
stopping web server
killing pid: 16910
cwd: /usr/ports/net-p2p/libtorrent-rasterbar/work/.build/test/test_tmp_16909_1
Unhandled exception in thread started by <class '__main__.ConnectionHandler'>
  File "../http.py", line 111, in __init__
Unhandled exception in thread started by <class '__main__.ConnectionHandler'>
  File "../http.py", line 111, in __init__
    self.method, self.path, self.protocol = self.get_base_header()
  File "../http.py", line 144, in get_base_header
    self.method, self.path, self.protocol = self.get_base_header()
  File "../http.py", line 144, in get_base_header
    self.client_buffer += self.client.recv(BUFLEN)
TypeError: can only concatenate str (not "bytes") to str
    self.client_buffer += self.client.recv(BUFLEN)
TypeError: can only concatenate str (not "bytes") to str
Unhandled exception in thread started by <class '__main__.ConnectionHandler'>
  File "../http.py", line 111, in __init__
    self.method, self.path, self.protocol = self.get_base_header()
  File "../http.py", line 144, in get_base_header
    self.client_buffer += self.client.recv(BUFLEN)
TypeError: can only concatenate str (not "bytes") to str
Unhandled exception in thread started by <class '__main__.ConnectionHandler'>
  File "../http.py", line 111, in __init__
    self.method, self.path, self.protocol = self.get_base_header()
  File "../http.py", line 144, in get_base_header
    self.client_buffer += self.client.recv(BUFLEN)
TypeError: can only concatenate str (not "bytes") to str
Unhandled exception in thread started by <class '__main__.ConnectionHandler'>
  File "../http.py", line 111, in __init__
    self.method, self.path, self.protocol = self.get_base_header()
  File "../http.py", line 144, in get_base_header
    self.client_buffer += self.client.recv(BUFLEN)
TypeError: can only concatenate str (not "bytes") to str
Unhandled exception in thread started by <class '__main__.ConnectionHandler'>
  File "../http.py", line 111, in __init__
    self.method, self.path, self.protocol = self.get_base_header()
  File "../http.py", line 144, in get_base_header
    self.client_buffer += self.client.recv(BUFLEN)
TypeError: can only concatenate str (not "bytes") to str
Unhandled exception in thread started by <class '__main__.ConnectionHandler'>
  File "../http.py", line 111, in __init__
    self.method, self.path, self.protocol = self.get_base_header()
  File "../http.py", line 144, in get_base_header
    self.client_buffer += self.client.recv(BUFLEN)
TypeError: can only concatenate str (not "bytes") to str
Unhandled exception in thread started by <class '__main__.ConnectionHandler'>
  File "../http.py", line 111, in __init__
    self.method, self.path, self.protocol = self.get_base_header()
  File "../http.py", line 144, in get_base_header
    self.client_buffer += self.client.recv(BUFLEN)
TypeError: can only concatenate str (not "bytes") to str
Unhandled exception in thread started by <class '__main__.ConnectionHandler'>
  File "../http.py", line 111, in __init__
    self.method, self.path, self.protocol = self.get_base_header()
  File "../http.py", line 144, in get_base_header
    self.client_buffer += self.client.recv(BUFLEN)
TypeError: can only concatenate str (not "bytes") to str
^CTraceback (most recent call last):
  File "../http.py", line 248, in <module>
  File "../web_server.py", line 211, in <module>
    httpd.handle_request()
  File "/usr/local/lib/python3.7/socketserver.py", line 294, in handle_request
    start_server(port=listen_port)
  File "../http.py", line 226, in start_server
    ready = selector.select(timeout)
  File "/usr/local/lib/python3.7/selectors.py", line 415, in select
    thread.start_new_thread(handler, soc.accept() + (timeout,))
  File "/usr/local/lib/python3.7/socket.py", line 212, in accept
    fd_event_list = self._selector.poll(timeout)
KeyboardInterrupt
[yuri@yv /usr/ports/net-p2p/libtorrent-rasterbar/work/.build/test]$     fd, addr = self._accept()
```

AllSeeingEyeTolledEweSew:
Making tests python3 compatible.

This is a pretty minimal change, basically to work around the bytes/str disparities in py3. Also treating py2 compatibility code as dead code.

A more complete fix would be to rewrite http.py as a BaseHTTPServer, which would work around encoding issues more cleanly, but maybe it's not worth that much effort for test infrastructure until it breaks.


FranciscoPombal:
> A more complete fix would be to rewrite http.py as a BaseHTTPServer, which would work around encoding issues more cleanly, but maybe it's not worth that much effort for test infrastructure until it breaks.

Since python2 is EOL'd anyway... why not? Assuming @arvidn is fine with that of course.

arvidn:
@FranciscoPombal the CI environments are finicky and fragile. Making sure all environments have a working python3 version installed and that it's preferred over a python2 installation and that all python code is python3 compatible is a non-trivial amount of work.

You end up with errors like this: https://travis-ci.org/github/arvidn/libtorrent/jobs/682400880#L5775

(but maybe that makes more sense to someone with more python experience than myself).

The good news is that the situation is better now than it was 6 months ago (or so) when I last attempted the transition of test scripts to python 3. The python binding and boost.python now all builds against python 3 (iirc).

arvidn:
and errors like this: https://ci.appveyor.com/project/arvidn/libtorrent/builds/32606474/job/521mw1rrjutp12ly#L1265

FranciscoPombal:
@arvidn 

> The python binding and boost.python now all builds against python 3 (iirc).

Not necessarily, unfortunately - that is part of the reason for my other PR.

Basically, it mostly depends on how you call the python executable. The best practice is to always use `python3` (e.g. `python3 script.py` or something like `#!/usr/bin/env python3` in scripts). The reason for this is because you never know what the unversioned `python` executable will point to. On modern distro releases, like the recently released Ubuntu 20.04, which have which have adapted to the post-python2 world, the policy is "`python` is deprecated, but points to `python3`". However, in not-so-old releases, like Ubuntu 18.04, `python` points to `python2`.

This bit me in the ass recently in PR I contributed to vcpkg, attempting to expose the python bindings feature of libtorrent. Supposedly the script was using `python3`, but `boost-python` was automatically linking against `python` dev libraries, which on my Ubuntu 18.04 system points to `python2` (and I happened to have the required packages). On other systems (clean installs of Ubuntu 18.04), it would of course fail. In the end I just patched the cmake file for the python bindings to work with python3 only - that way, even on systems where `python` points to `python2`, everything works correctly.

So, as long as every script, invocation, etc in the code uses `python3` explicitly, this:

>Making sure all environments have a working python3 version installed and that it's preferred over a python2 installation and that all python code is python3 compatible is a non-trivial amount of work.

should not be that hard.

It would of course be much harder if the goal was to support both python2 and python3 harmoniously, across systems where `python` may point to either. But given that python2 was EOL'd and is being abandoned fast(1), I think it's reasonable to only support python3.

(1) The number of python2 downloads on pypi is still significantly, which might make one think twice. However, even though I have no concrete evidence to back this claim up, I think that's entirely because until very recently, `python` in the mainstream distros (Ubuntu 16.04, 18.04, Debian stable, ...) points to `python2`, so users doing `pip install <package>` will grab python2 packages. Then they promptly realize their mistake and use `pip3`, or `pip` inside a python3 venv.
I would be very very surprised if any remotely significant percentage of users in 2020 is using a distro that does not have python3.

arvidn:
> > Making sure all environments have a working python3 version installed
> > and that it's preferred over a python2 installation and that all python code
> > is python3 compatible is a non-trivial amount of work.
> 
> should not be that hard.

The CI runs windows and MacOS as well, and travis runs old versions of MacOS and Linux. Appveyor has newer images with python 3. But sometimes newer images are also tied to newer other things, like compilers and other packages, which have ripple effects.

Anyway, I encourage trying to get this to work!

AllSeeingEyeTolledEweSew:
I had another motivation for this also, which was that the python binding was missing some python3 niceness (`torrent_handle.add_piece` only accepts `str`, not `bytes`). I wanted to fix this and add tests for it. PR for that incoming. I thought I'd get all python code running on python3 as a first step.

For another data point, it looks like deluge 2.0.x still supports python 2.7.

My feeling is it's probably reasonable that a project like libtorrent explicitly support both for the time being, so maybe the bindings tests should be made to run on both. The support scripts can be python3 only though.

So far it looks like python3 is installed and working on the linux and osx images, but python2.7 is the "default" one on the win32 image. I'll have to figure out how to invoke a python3 interpreter explicitly on win32.

FranciscoPombal:
@AllSeeingEyeTolledEweSew 
>For another data point, it looks like deluge 2.0.x still supports python 2.7.

Yikes, that's unfortunate. At least it works with python3. Maybe the deluge dev can be brought into the loop and drop python2 support for the 2.1.x series?

>The support scripts can be python3 only though.

Yep, this is reasonable in the meantime.

According to https://www.appveyor.com/docs/windows-images-software/#python, python 3.5.4 is supported in any image. Python 3.5 EOLs in a few months, but a transition to 3.6.x would still be smooth, since according to the same source, 3.6.8 is supported in all images except VS 2013, which is not a problem anyway since currently libtorrent already uses the VS 2017 image. In fact, even staying on the VS 2017 image, everything will be fine until python > 3.8.0.

So, to get appveyor to pass, just call python3 with the full path to those directories.

arvidn:
> So far it looks like python3 is installed and working on the linux and osx images, but python2.7 is the "default" one on the win32 image. I'll have to figure out how to invoke a python3 interpreter explicitly on win32.

iirc, I had the most issue with having the test executable run python 3 from mingw (because it picks up the non-mingw version of python). To work around it I added an environment variable `LIBTORRENT_PYTHON_INTERPRETER` that I set in the CI run script. It's possible all this is master only, but it might be worth pulling some of that in, if it isn't already in `RC_1_2`

arvidn:
actually, if all you're doing is to make the test scripts compatible with both python2 and python 3, I guess my previous comment doesn't apply.

arvidn:
> So, to get appveyor to pass, just call python3 with the full path to those directories.

Python is run from within the C++ unit tests, and those should not have hard coded paths specific to appveyor (or any other CI). Hence the environment variable I mentioned. The greater point is that it's not necessarily simple

FranciscoPombal:
@arvidn 
> Python is run from within the C++ unit tests, and those should not have hard coded paths specific to appveyor (or any other CI). Hence the environment variable I mentioned. The greater point is that it's not necessarily simple

Did not that, that certainly complicates things. Perhaps it is not so hard to change the default python version on appveyor (so that python->python3)? Will look into that.

FranciscoPombal:
Could this work? https://www.appveyor.com/docs/lang/python/#test-setup

AllSeeingEyeTolledEweSew:
I don't think tox helps us run the helper scripts, since they need to be invoked from C++, and they're "normal python programs", not test cases. I expect you could get it to work but you'd have to jerry rig it pretty hard.

It looks like there are python3 versions installed on the images in question, and I think it's sufficient to just tweak the invocation. I will do it in a second, I am working out a bug I put in the actual helper script.

Tox is definitely more compelling for invoking the binding tests, but I would actually like to refactor the binding tests to improve isolation, coverage, readability and debug-ability. I'll add tox to my to-do list.

arvidn:
it seems one problem on MacOS is that the web server closes the connection early, causing the web seed download test to fail: https://travis-ci.org/github/arvidn/libtorrent/jobs/683085767#L2196

and flake8 is complaining about formatting: https://travis-ci.org/github/arvidn/libtorrent/jobs/683085769#L977

on linux it looks like the web seed download test is having the same problem as on MacOS: https://travis-ci.org/github/arvidn/libtorrent/jobs/683085770#L2503
https://travis-ci.org/github/arvidn/libtorrent/jobs/683085772#L1834

AllSeeingEyeTolledEweSew:
The http.py proxy is giving me some trouble.

It doesn't seem terribly resilient. It's checking for EWOULDBLOCK on a blocking socket, and never checks for EOF, so whenever I run the tests they spin at 100% CPU on my machine. This is impeding testing.

I think it's also not terribly correct with the HTTP/1.1 spec. It connects to the host of the first request in the connection, then blindly copies everything. There are some subtleties to how proxies should pass headers, and a lot of subtlety to when a client should expect a connection to be closed, so I question if the http proxy tests are accurate now.

None of the third-party HTTP proxies I've found will fit in a single .py script. They'd all require some setup before running.

My understanding of the build system and test drivers is really weak, and my understanding of Python is better, so I'm taking a shot at rewriting the HTTP proxy, to be spec-compliant enough to test against, and run in a single file without dependencies. It's taking a while as Python's built-in http server and client provide less functionality than I thought. But I should be finished pretty soon.

arvidn:
yeah, shortcuts were made to just support the specific test cases I used it for. spinning with 100% CPU sounds pretty bad though, that sounds like a bug I have missed

AllSeeingEyeTolledEweSew:
I tried for a few days to make it work, but I couldn't figure out the right set of things to fix, versus leaving them shortcut-ed the way they were before. I'm finding it simpler to just rewrite it.

Far from the biggest yak I've ever shaved, but it's certainly up there

AllSeeingEyeTolledEweSew:
There's some complexity with making test.py compatible with both python2 and python3. It looks like the build system only builds against one of these, so maybe we'd have to duplicate the build logic to do both, if each is available. My understanding of the build system is still pretty flimsy, so I'm not going to try to do this here.

arvidn:
the build system is only concerned about python version for building the python bindings. I don't think that affects the unit tests at all, or which version of python they are running

AllSeeingEyeTolledEweSew:
Hm, I probably have some things

arvidn:
https://travis-ci.org/github/arvidn/libtorrent/jobs/685573305#L807

arvidn:
let's ignore guardrails for now. In practice that is a benign use of `system()`

AllSeeingEyeTolledEweSew:
Well, it looks like google's cpplint.py does not work with python 3, even at its most recent version.

It tests `isinstance(thing, unicode)`, which has never worked in any version of python 3. It also swallows the error with its overridden `sys.stderr`.

Someone should file a bug against their project for this. I'm at my limit for yak shaving at the moment. I put a comment in .travis.yml about this so when it breaks there'll be some pointer.

arvidn:
oh. `cpplint.py` is kind of old. I could probably remove it from the CI since there's `clang-tidy` now

AllSeeingEyeTolledEweSew:
I think it's finally done, assuming the doozer failures aren't my fault.

arvidn:
I take it this is not "WIP" anymore then, right?

AllSeeingEyeTolledEweSew:
Yes.

PR #4611, opened 2020-05-03
make torrent_info's merkle trees internal

arvidn:


arvidn:
What I'm working towards is to replace the `vector<sha256_hash>` as the representation for a merkle tree with an internal data structure. I want it to remain internal to allow for advances in it without breaking ABI in the future.

Making this function internal insulates clients from this internal type. The merkle trees can still be interacted with via the resume data.

The purpose of this (future) internal merkle tree data structure is to make it sparse. For instance, once we have the full tree, it would be sufficient to just store the block layer. Or if we have the piece layer (which is likely to be found in a .torrent file) we can just store that.

Given that we don't need to store padding either, this would likely be more than a 50% space saving, best case close to 75% space saving.

I'm working on this patch. It will already be fairly large, so I'm shaving smaller independent pieces off to land ahead of it.

PR #4612, opened 2020-05-04
Let add_piece python binding work with bytes.

AllSeeingEyeTolledEweSew:
Pretty straightforward.

I also added a test case for `add_piece`, with some synthetic data.


AllSeeingEyeTolledEweSew:
This won't collide with my other python3 PR.

PR #4619, opened 2020-05-08
GnuTLS prerequisite: TORRENT_USE_SSL compilation flag

paullouisageneau:
This PR introduces the compilation flag `TORRENT_USE_SSL`, defined to 1 iff `TORRENT_USE_OPENSSL` is defined else 0 in `config.hpp`.

This is a prerequisite for GnuTLS support https://github.com/arvidn/libtorrent/pull/4588

### DIFF
--- a/include/libtorrent/http_connection.hpp
+++ b/include/libtorrent/http_connection.hpp
@@ -36,7 +36,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #ifndef TORRENT_HTTP_CONNECTION
 #define TORRENT_HTTP_CONNECTION
 
-#ifdef TORRENT_USE_OPENSSL
+#if TORRENT_USE_SSL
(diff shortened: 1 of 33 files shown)


arvidn:
This is the benefit of using `#if` instead of `#ifdef`: https://travis-ci.org/github/arvidn/libtorrent/jobs/684626259#L371

It looks like perhaps there's a missing `config.hpp` include.

include/libtorrent/http_connection.hpp
```diff
@@ -36,7 +36,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #ifndef TORRENT_HTTP_CONNECTION
 #define TORRENT_HTTP_CONNECTION
```
arvidn, line 38: if you add `#include "libtorrent/config.hpp"` here, I think it will build on the MacOS CI as well

Issue #4529, opened 2020-04-11
 RC_1_2 tracker announce error using "127.0.0.1".


mhertz:
I make an unofficial deluge 2.0.3 windows installer(as not available for windows officially), and several users reported tracker connection errors sometime after libtorrent 1.2.3, in RC_1_2 branch builds. I asked for debug logs to forward here, from infected users, as I couldn't reproduce, but none replied with such unfortunetly. I last got a report not that long ago, with RC_1_2 build, so still an issue, though no logs provided again. Now all of a sudden i'm able to reproduce it for some reason, in my win10 VM which use bridged networking and host network nothing special and no VPN or anything(linux btw). In my installer I include an option for libtorrent 1.2.3 since fixes it for affected users I got reported, and it also fixes it for me I see now.

Deluge sets 0.0.0.0 as listen interface by default when you don't define anything yourself, and on RC_1_2 after 1.2.3, then I guess the multi-homed change, makes it so that I get announcess on localhost(127.0.0.1) and LAN IP, and the 127.0.0.1 always fails with error, so tracker shows in error status in deluge GTK-UI, and in the log I see the LAN IP works, and retrieves peers/seeds and scrapes etc.

Here are some parts of debug-logs I just obtained:

RC_1_2 today build:
````
15:40:05 [DEBUG   ][deluge.core.preferencesmanager    :65  ] Config key: listen_ports set to [6881, 6881]..
15:40:05 [DEBUG   ][deluge.core.preferencesmanager    :65  ] Listen Interface: 0.0.0.0, Ports: [61262, 61262] with use_sys_port: False
15:40:05 [DEBUG   ][deluge.core.preferencesmanager    :65  ] Config key: listen_interface set to ..
15:40:05 [DEBUG   ][deluge.core.preferencesmanager    :65  ] Listen Interface: 0.0.0.0, Ports: [61262, 61262] with use_sys_port: False
15:40:05 [DEBUG   ][deluge.core.preferencesmanager    :65  ] Config key: outgoing_interface set to ..
15:40:05 [DEBUG   ][deluge.core.preferencesmanager    :65  ] Config key: random_port set to True..
15:40:05 [DEBUG   ][deluge.core.preferencesmanager    :65  ] Listen Interface: 0.0.0.0, Ports: [61262, 61262] with use_sys_port: False
15:40:05 [DEBUG   ][deluge.core.preferencesmanager    :65  ] Config key: outgoing_ports set to [0, 0]..
[...]

PR #4620, opened 2020-05-08
limit concurrent http tracker announces

arvidn:


arvidn:
ideally this limit would be *per tracker*, but that would be a bit more complicated to implement

an0n666:
If I have lots of torrents then I'd have to bump the limits a bit higher right? Otherwise I just have to wait longer for client to close..

arvidn:
yes. And if you bump the limit higher, you may have to also raise your file descriptor limit (`ulimit -n`) for your process

arvidn:
@skuizy would this have to land in `RC_1_2` for you to be able to test it? Or can you build from another branch?

FranciscoPombal:
> yes. And if you bump the limit higher, you may have to also raise your file descriptor limit (`ulimit -n`) for your process

FYI qBittorrent >= 4.2.0 sets this to unlimited on startup.

FranciscoPombal:
@arvidn 
> ideally this limit would be per tracker, but that would be a bit more complicated to implement

What are your plans for this? Do you think it is worth it exposing this setting in qBittorrent or do you plan to rework this limit to work per tracker relatively soon?

arvidn:
I was hoping that someone could test this, at least. I improving this is a fair amount of work. The tracker code is some of the oldest in libtorrent and could really do with some rewriting (it's been patched so many times it's pretty brittle to be patched even more to fix this).

I envision that HTTP pipelining and per-host queuing would be the way to go eventually

an0n666:
Per host queue might not help if your system is limiting file descriptors. I think this type of limits should apply in general. One could have 500 torrents each for 4-5 trackers, individually limiting them to 50 would still push the actual concurrents over 200+. And individually using a lower limit would delay the client closing even more.

arvidn:
good point

skuizy:
> 
> 
> @skuizy would this have to land in `RC_1_2` for you to be able to test it? Or can you build from another branch?

Yes please. I never managed to build LT successfully :(

arvidn:
@skuizy ok, I've landed this in `RC_1_2` now. I'm not super confident in it, so please test it carefully.

skuizy:
OK after leaving deluge run with LT compiled with this patch and 2k+ torrents for 24h, I went through theses cases :
- 0 tracker error after installing the patch and restarting deluge deamon ! 
- All torrents are paused during the night and are resumed in the morning. After resuming I got ~600 torrents announces in connection timeout error that wouldn't have been retried...
- After forcing reanounce for all these torrents, I still get ~150 torrents in connection timeout error that won't get reannounced even manually.
- Failling torrents can be reanounced when manually paused and resumed individually.

So I'd say the issue in my case is not fully addressed despite some improvements that needs to be confirmed in the long run (see first point above).

arvidn:
maybe 50 concurrent announces is too many. you could try lowering it to, say 10. ([here](https://github.com/arvidn/libtorrent/pull/4620/files#diff-8e5f23e1a44ba74e70cdc5067a511ec3R355))

arvidn:
oh, right. you're not building yourself, never mind then.

skuizy:
I can play with max_concurrent_http_announces through ltConfig plugin for a bit.

an0n666:
> * After forcing reanounce for all these torrents, I still get ~150 torrents in connection timeout error that won't get reannounced even manually.
> * Failling torrents can be reanounced when manually paused and resumed individually.

Deluge force reannounce has issues in deluge 2.0.x with latest LT. I think that's why your force reannounce isn't working but pause/resume is. This is a problem that requires fixing on Deluge's end.

skuizy:
OK is there an issue open on their bug tracer ? The closest I can find is [3384](https://dev.deluge-torrent.org/ticket/3384) but it doesn't seems relevant as I'm only announcing through one interface...

an0n666:
> OK is there an issue open on their bug tracer ? The closest I can find is [3384](https://dev.deluge-torrent.org/ticket/3384) but it doesn't seems relevant as I'm only announcing through one interface...

You see, you may have one interface but there can be other local endpoints (i.e 127.0.0.1) as well.
There are couple of tickets open/closed regarding this issue.
One of them --> https://github.com/arvidn/libtorrent/issues/4529

skuizy:
I don't think this is the same as I have 3 intefaces : eth0, lo and tun0.
My deluge daemon is set to listen and emmit on tun0, so is lt.
Looking at the logs it produces when announcing to tracker, I can verify that other interfaces aren't used, only tun0.
If I don't set my outgoing interface, I can see in my logs that the announces are made on eth0, tun0 and lo (which is obviously timing out).

skuizy:
OK so this morning with max_concurrent_http_announces  set at 10, I was at ~400 announces errors.

After forcing tracker update, I went down to the same ~150 impossible to reannounce as yesterday.

I restarted my deluge daemon and I'm now at ~10 errors, but I'm OK with them as they are being reannounced periodically.

@an0n666 : to complete my previous point, I don't really care my torrents being in an error state as long as they will be reanounced later, even if it a false state because of the lo interface. Here they are simply failing and never retried until I manually reannounce them for some or fully discarded for others.

an0n666:
If your tracker is working in almost all torrents after a restart then the concurrent limit is working fine on client start. However it might not be working as expected when you’re resuming all torrents at once. Also I don’t know why some are not getting retried. Only possible explanation is the tracker or endpoint is getting disabled.

skuizy:
That's why I'm pointing it out :)

This morning it was in the same state with max_concurrent_http_announces still at 10. 

Looks like something lays in the resuming function.

>  Only possible explanation is the tracker or endpoint is getting disabled.

Well this same tracker is OK for other torrents that are being updated, is it handeled individually ? 
Which endpoint are you reffering to ?

arvidn:
I wonder if tracker requests that are queued could get stuck in the queue, and never actually announcing. Would that fit the symptoms?

skuizy:
Could be... How to check ? 🤔

arvidn:
it can really only be checked directly in a debugger. But, if you see any behavior that is *not* consistent with this theory, at least we can conclude this isn't it. For example, I would expect a stuck announce say "announcing" but never progress from there.

skuizy:
Oh right, so I think we can rule this out as stuck torrents are in "Error: Connection timed out" state.

skuizy:
> 
> 
> it can really only be checked directly in a debugger. But, if you see any behavior that is _not_ consistent with this theory, at least we can conclude this isn't it. For example, I would expect a stuck announce say "announcing" but never progress from there.

OK contradicting myself once agian, but I just got this today AFTER restarting deluge. Traker Status for almost all my torrents was "Announce Sent" and wasn't timing out, with the next announce in 18h 22m. This is hard to track as these torrents won't be shown by the error filter...

PR #4622, opened 2020-05-08
GnuTLS prerequisite: Implement utp_stream::write_some()

paullouisageneau:
This PR implements the previously unimplemented method `utp_stream::write_some()`.

Due to the lack of write buffer in the UTP socket, it is always non-blocking. It only tries to write as much as there is space available in the congestion window, and returns immediately.

This is a prerequisite for GnuTLS support #4588


paullouisageneau:
@arvidn Have you an idea how I should test this? `test_utp` is a full end-to-end test so it can't really be adapted.

arvidn:
yeah, unfortunately there's no good uTP test suite

PR #4626, opened 2020-05-09
merkle tree abstraction

arvidn:
This is a proposed abstraction for the per-file merkle tree. Currently this tree is a `std::vector<sha256_hash>` that's allocated to fit the full tree.

The goal of this abstraction is to enable a sparse representation of this tree, but this is not implemented yet. The `aux::merkle_tree` class does not allow direct write access to the tree, but the tree can only be mutated via a few high level operations. I believe these operations are at somewhat reasonable levels.

I've divided up the patch in bite-sized commits that can be reviewed individually.

@ssiloti I'm interested in whether you think this is a reasonable abstraction, or if you anticipate problems I may run into going further in this direction.

### DIFF
--- a/include/libtorrent/aux_/merkle.hpp
+++ b/include/libtorrent/aux_/merkle.hpp
@@ -101,6 +101,19 @@ namespace libtorrent {
 	TORRENT_EXTRA_EXPORT
 	std::pair<aux::vector<std::pair<sha256_hash, sha256_hash>>, sha256_hash>
 	merkle_check_proofs(sha256_hash to_validate, span<sha256_hash const> hashes, int index);
+
+	TORRENT_EXTRA_EXPORT
+	bool merkle_validate_node(sha256_hash const& left, sha256_hash const& right
+		, sha256_hash const& paremt);
+
+	// validates hashes from src and copies the valid ones to dst given root as
+	// the expected root of the tree (i.e. index 0)
+	// src and dst must be the same size. dst is expected to be initialized
+	// cleared (or only have valid hashes set), this function will not clear
+	// hashes in dst that are invalid in src.
+	TORRENT_EXTRA_EXPORT
+	void merkle_validate_copy(span<sha256_hash const> src, span<sha256_hash> dst
+		, sha256_hash const& root);
 }
 
 #endif
--- a/include/libtorrent/aux_/merkle.hpp
+++ b/include/libtorrent/aux_/merkle.hpp
@@ -103,6 +103,10 @@ namespace libtorrent {
 	merkle_check_proofs(sha256_hash to_validate, span<sha256_hash const> hashes, int index);
 
 	TORRENT_EXTRA_EXPORT
+	bool merkle_validate_proofs(int start_idx
+		, span<std::pair<sha256_hash, sha256_hash> const> proofs);
+
+	TORRENT_EXTRA_EXPORT
 	bool merkle_validate_node(sha256_hash const& left, sha256_hash const& right
 		, sha256_hash const& paremt);
 
--- a/include/libtorrent/aux_/merkle.hpp
+++ b/include/libtorrent/aux_/merkle.hpp
@@ -126,6 +126,18 @@ namespace libtorrent {
 
 	TORRENT_EXTRA_EXPORT
 	bool merkle_validate_single_layer(span<sha256_hash const> tree);
+
+	// given a leaf index (0-based index in the leaf layer) and a tree, return
+	// the leafs_start, leafs_size and root_index representing a subtree that
+	// can be validated. The leaf_index and leaf_size is the range of the leaf
+	// layer that can be verified, and the root_index is the node that needs to
+	// be known in (tree) to do so. The num_valid_leafs specifies how many of
+	// the leafs that are actually *supposed* to be non-zero. Any leafs beyond
+	// thses are padding and expected to be zero.
+	// The caller must validate the hash at root_index.
+	TORRENT_EXTRA_EXPORT
+	std::tuple<int, int, int> merkle_find_known_subtree(span<sha256_hash const> const tree
+		, int leaf_index, int num_valid_leafs);
 }
 
 #endif
--- a/include/libtorrent/aux_/merkle.hpp
+++ b/include/libtorrent/aux_/merkle.hpp
@@ -56,6 +56,10 @@ namespace libtorrent {
 	// returns the number of nodes in the tree, given the number of leaves
 	TORRENT_EXTRA_EXPORT int merkle_num_nodes(int);
 
+	// given the number of leafs in the tree, returns the index to the first
+	// leaf
+	TORRENT_EXTRA_EXPORT int merkle_first_leaf(int num_leafs);
+
 	// takes the number of leaves and returns the height of the merkle tree.
 	// does not include the root node in the layer count. i.e. if there's only a
 	// root hash, there are 0 layers. Note that the number of leaves must be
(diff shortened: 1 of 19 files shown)


include/libtorrent/aux_/merkle_tree.hpp
```diff
@@ -80,6 +84,14 @@ struct TORRENT_EXTRA_EXPORT merkle_tree
 
 	bool load_piece_layer(span<char const> piece_layer);
 
+	// inserts the hashes in "tree" at position "dest_start_idx"
+	// the proofs are the left and right hashes above the root of "tree",
+	// anchoring it in the existing tree. These are also inserted into the tree
```
ssiloti, line 85: The meaning of the return value should be explained since it's not obvious.

src/merkle_tree.cpp
```diff
@@ -89,6 +89,74 @@ namespace aux {
 		return true;
 	}
 
+	std::map<piece_index_t, std::vector<int>> merkle_tree::add_hashes(
+		int dest_start_idx
+		, int const blocks_per_piece
+		, span<sha256_hash const> tree
+		, span<std::pair<sha256_hash, sha256_hash> const> proofs)
+	{
+		std::map<piece_index_t, std::vector<int>> ret;
```
ssiloti, line 98: I think this should be given a more descriptive name since the meaning of the return value is not obvious from the function signature. Something like `failed_blocks`.

src/merkle_tree.cpp
```diff
@@ -157,6 +157,66 @@ namespace aux {
...
+					done = true;
+					break;
+				}
+			}
+			if (done) continue;
+
+			merkle_fill_tree(m_tree, num_leafs, file_first_leaf + first_leaf);
+			if (m_tree[base_layer_start + i] != hashes[i])
+			{
+				merkle_clear_tree(m_tree, num_leafs / 2, merkle_get_parent(file_first_leaf + first_leaf));
```
ssiloti, line 206: Did you mean to undo the changes made in previous commits to replace calls to `merkle_clear_tree`, `merkle_fill_tree` and `is_all_zeros` with the member functions `clear`, `fill`, and `has_node`?

ssiloti:
This seems like a reasonable approach to me. The one minor concern I have is that when the tree is sparse peers could craft a hash request which would require a large number of hash computations to generate proof hashes. E.g. request two leaf hashes and proofs all the way back to the file root.

I'm not sure if this would be a problem in practice. I don't think it would be very effective as a DOS attack. It may still be a good idea to rate limit hash requests which require a lot of hash (re)computation.

src/merkle_tree.cpp
arvidn: I did, yes, thinking that maybe those member functions won't be necessary anymore. I realize this thought wasn't completed though. I made them private, but the member functions probably don't need to exist anymore. The free functions do the same things and have unit tests already.

include/libtorrent/aux_/merkle_tree.hpp
arvidn: yeah, I need to spend some more time to fully grasp everything myself. I expect writing unit tests will make things more clear for me, hopefully I can come up with better names and comments

arvidn:
thanks for the review. Yeah, I should spend more time on improving the names and comments. I will also add unit tests for the `merkle_tree` type before making it support sparse modes.

Maybe a (future) approach to the DoS potential would be to keep score of how many hashes have been computed in response to a hash request (i.e. proofs higher up the tree have a higher cost) and rate limit based on that.

arvidn:
btw. To further justify this whole project; the reason I consider the memory usage a big deal is because I failed to run my performance test that I run every now and then (once a year or so). Where I generate 100'000 torrents and load them up in client_test under a profiler.

Running this experiment in master fails with out-of-memory, on my linux machine with 78.6 GiB of RAM.

src/merkle.cpp
```diff
@@ -208,5 +208,34 @@ namespace libtorrent {
...
 
+	std::pair<aux::vector<std::pair<sha256_hash, sha256_hash>>, sha256_hash>
+	merkle_check_proofs(int proof_leafs, sha256_hash tree_root, span<sha256_hash const> hashes
+		, int const index)
+	{
+		aux::vector<std::pair<sha256_hash, sha256_hash>> ret(int(hashes.size()));
+		auto ret_it = ret.begin();
+		for (auto const& proof : hashes)
+		{
+			proof_leafs *= 2;
```
arvidn, line 219: @ssiloti I'm working on unit tests for this function, and this line seems wrong.

my understanding of this snippet of code is that `tree_root` is the root of the sub tree we just inserted. We've validated that the inserted sub tree is internally consistent, now we want to make sure its root is valid in the full tree we're inserting it into.

In this context, `tree_root` is really at the bottom of part of the tree we're interested in. We'll iterate *up* the tree while using the uncle hashes passed in via `hashes`. At the end we have the root of the full tree (or at least the hash of a known-good node in the full tree) which is returned, along with the hashes of the nodes along the path.

The caller can then compare the returned hash against the known good one and determine whether `tree_node` was correct.

Given this understanding, it seems this line `proof_leafs *= 2` is wrong, and should be `proof_leafs /= 2`. Since we traverse *up* the tree, each layer has fewer nodes. In fact, I would expect a simpler way to implement this would be to skip the `proof_leafs` entirely, and just halve the `index` here (`index /= 2`) and `proof_right = (index & 1) == 0;`

Those changes are consistent with my understanding and it makes my test pass, but I'm worried I'm missing something.

arvidn: it's possible I had misunderstood the meaning of `index`. It seems to be the offset from the start of the layer of the first inserted *leaf*, not the layer offset for the *root*. This code is only concerned about the root though, so I think it would be a simplification to make it refer to the offset of `tree_root`.

ssiloti: D'oh, you've got it right. Not sure how this code worked at all in my testing. I agree it would be simpler to use the index of the root to determine the positions of the sibling hashes.

arvidn: I broke this out into a separate PR: https://github.com/arvidn/libtorrent/pull/4652

src/merkle.cpp
```diff
@@ -217,26 +224,115 @@ namespace libtorrent {
...
+		{
+			if (!merkle_validate_node(tree[idx], tree[idx + 1], tree[merkle_get_parent(idx)]))
+				return false;
+			idx += 2;
+		}
+		return true;
+	}
+
+	std::tuple<int, int, int> merkle_find_known_subtree(span<sha256_hash const> const tree
+		, int const block_index, int const num_valid_leafs)
```
arvidn, line 303: @ssiloti I'm impressed you got all this right without extensive unit tests :)

include/libtorrent/aux_/merkle.hpp
```diff
@@ -125,6 +125,18 @@ namespace libtorrent {
 
 	TORRENT_EXTRA_EXPORT
 	bool merkle_validate_single_layer(span<sha256_hash const> tree);
+
+	// given a leaf index (0-based index in the leaf leyer) and a tree, return
```
ssiloti, line 129: leyer -> layer

arvidn:
I think this patch is big enough to not build more onto. But the main missing part is still replacing (or at least replacing most use of) ``merkle_tree::operator[]()`` in favor of a higher level function to respond to a hash request. Right now there's potentially a lot of redundant memory allocation and hashing to respond to a request. Since each individual node is computed independently.

PR #4628, opened 2020-05-09
tweak rate based choker to increase rate threashold by 2 kiB/s.

arvidn:
Improve documentation

### DIFF
--- a/src/choker.cpp
+++ b/src/choker.cpp
@@ -307,8 +307,8 @@ namespace {
 		// intention is to not spread upload bandwidth too thin, but also to not
 		// unchoke few enough peers to not be able to saturate the up-link.
 		// this is done by traversing the peers sorted by our upload rate to
-		// them in decreasing rates. For each peer we increase our threshold
-		// by 1 kB/s. The first peer we get to whom we upload slower than
+		// them in decreasing rates. For each peer we increase the threshold by
+		// 2 kiB/s. The first peer we get to whom we upload slower than
 		// the threshold, we stop and that's the number of unchoke slots we have.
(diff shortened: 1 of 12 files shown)


src/choker.cpp
```diff
@@ -307,8 +307,8 @@ namespace {
 		// intention is to not spread upload bandwidth too thin, but also to not
 		// unchoke few enough peers to not be able to saturate the up-link.
 		// this is done by traversing the peers sorted by our upload rate to
-		// them in decreasing rates. For each peer we increase our threshold
-		// by 1 kB/s. The first peer we get to whom we upload slower than
+		// them in decreasing rates. For each peer we double the threshold
```
FranciscoPombal, line 310: shouldn't it be "increase the threshold by 2 KiB/s"? I though a multiplicative change to the threshold was being investigated in https://github.com/arvidn/libtorrent/pull/4417 instead.

arvidn: oh, you're right. I was looking through the docs for these kinds of issues, but must have glossed over this one.

I wanted to land a more conservative version of this patch first, and in `RC_1_2`. Especially the documentation updates are a clear improvement. #4417 is a bit more ambitious and I would like to have proper tests before landing it, I'm also starting to think that it probably should go into `master` instead.

This patch is picking the least controversial parts out of that patch to land them in `RC_1_2` at least.

arvidn: I think I missed this because it's internal documentation, not the one that makes it into the generated docs

arvidn:
Thanks for reviewing!

PR #4636, opened 2020-05-10
Add remove_flags_t to python bindings.

AllSeeingEyeTolledEweSew:
Add `remove_flags_t` to python bindings.

I guess `options_t` was sort of meant to be this, but the name is different and it never included `delete_partfile`. Maybe `options_t` could be a deprecated alias?

Nothing in test.py for this right now. I would like to refactor test.py once my other python3 PR is done.

### DIFF
--- a/bindings/python/src/session.cpp
+++ b/bindings/python/src/session.cpp
@@ -1138,6 +1138,9 @@ void bind_session()
     s.attr("local_peer_class_id") = session::local_peer_class_id;
 
     s.attr("reopen_map_ports") = lt::session::reopen_map_ports;
+
+    s.attr("delete_files") = lt::session::delete_files;
+    s.attr("delete_partfile") = lt::session::delete_partfile;
     }
 
 #if TORRENT_ABI_VERSION == 1


arvidn:
You accidentally changed the `libsimulator` submodule in this commit

arvidn:
you don't think it would make sense to put these directly into the `session` class/scope?

Like these flags: https://github.com/arvidn/libtorrent/pull/4636/files#diff-9a8153ff01a5fbbc984a9b19dcb49e22L1140

AllSeeingEyeTolledEweSew:
That submodule keeps tripping me up when managing branches.

I think it's fine to define them on the `session` scope as you suggest, if "fidelity" to the C++ interface is the standard.

Native python code pretty much never uses int bitfields so there's not a lot of precedent for what to do. Normally, a python developer would bind all the strong int types to objects with attributes which map to bits, like this: https://stackoverflow.com/a/11481471 . I'm not suggesting libtorrent do this, as it does compromise the goal of efficiency. I thought the separate `remove_flags_t` object was a good compromise between normal python code structure and reducing the mappings we need to do.

But it is easiest to adhere to the standard of "look just like the C++ interface", so I think that does make it a better standard.

arvidn:
yeah, right now there's a little bit of both. originally these were C++ enums and they were exposed to python as such (boost python supports enums specifically). The thing is that there's no specific documentation for the python bindings, so if things could look similar to C++ the documentation is easier to map to python.

When I converted plain enums into strong types on the C++ side, I took the opportunity to deprecate the previous enums and expose them as types in the same scope as they appear in C++. Except I took the shortcut of just converting them all to integers in the python interface. The syntax is the same, you just get less type-safety.

AllSeeingEyeTolledEweSew:
I also would definitely prefer some specific documentation on the python bindings, as "ABI consistency" in python pretty much happens through documentation. Right now it's hard to know which version of libtorrent my code would actually support unless I test them all.

arvidn:
The intention is that the API will stay stable withing minor versions. I try to keep it mostly stable across major versions too of course, as long as deprecated functions are included in the build (which they are by default).

I think most (if not all) original enums are still in the bindings.

PR #4641, opened 2020-05-12
improve CMake - sync documentation with #4627 and fixup #4547

FranciscoPombal:
About documentation: since #4627, `OPENSSL_USE_STATIC_LIBS` is fully independent of `static runtime`. `OPENSSL_USE_STATIC_LIBS` is automatically set to `TRUE` if `BUILD_SHARED_LIBS` is `FALSE`.

About fixing up #4547 - according to https://gitlab.kitware.com/cmake/cmake/issues/19263, `crypt32` should be linked against the project, not against the imported OpenSSL targets. With this fix, the generated target file `LibtorrentRasterbarTargets.cmake` look good (`crypt32` appears as a `LINK_ONLY` interface link library alongside other system libraries like `ws2_32`), and I now no longer get link errors when linking statically with dynamic runtime (previously I had only tested static runtime).

To reproduce the link errors on Windows (static libs+dynamic runtime) without this patch:
- Build libtorrent with static linkage but dynamic runtime (for example, using the `vcpkg` triplet `x64-windows-static-md`)
- Try to use libtorrent's imported CMake targets to build a project that would use crypto facilities in libtorrent. The `examples` directory modified to be a standalone project will do:

```
# examples/CMakeLists.txt
cmake_minimum_required(VERSION 3.12)
project(libtorrent-examples)

set(single_file_examples
    simple_client
    custom_storage
    stats_counters
    dump_torrent
    make_torrent
    connection_tester
    upnp_test)

find_package(LibtorrentRasterbar CONFIG REQUIRED)

foreach(example ${single_file_examples})
    add_executable(${example} "${example}.cpp")
    target_link_libraries(${example} PRIVATE LibtorrentRasterbar::torrent-rasterbar)
endforeach(example)

add_executable(client_test
    client_test.cpp
    print.cpp
    torrent_view.cpp
    session_view.cpp)
target_link_libraries(client_test PRIVATE LibtorrentRasterbar::torrent-rasterbar)
```

```
$ cd examples
$ cmake -S . -B build -G "Visual Studio 16 2019" -A x64 -DCMAKE_TOOLCHAIN_FILE=path/to/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET=x64-windows-static-md
$ cmake --build build --config RelWithDebInfo
```

- This should yield link errors related to `libcrypto`. The fix is either this PR, or to explicitly add a dependency on `crypt32` for every project that uses libtorrent (obviously not desirable). In this example, it would involve changing both the `target_link_libraries` lines to ` target_link_libraries(${example} PRIVATE LibtorrentRasterbar::torrent-rasterbar crypt32)` and `target_link_libraries(client_test PRIVATE LibtorrentRasterbar::torrent-rasterbar crypt32)`, respectively.

Curiously, with or without this patch or with or without `crypt32` as an explicit dependency of the executables, building libtorrent itself with `-Dbuild_examples` works fine. I assume because in this case the `examples` are in a subdirectory, and the required dependency was propagated from somewhere else. Same for qBittorrent, but in this latter case I suspect it comes from the fact that Qt also uses OpenSSL, and so the appropriately configured dependency on `crypt32` must get propagated from that somehow.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -636,7 +636,7 @@ if(OPENSSL_FOUND)
 
 	# TODO: needed until https://gitlab.kitware.com/cmake/cmake/issues/19263 is fixed
 	if(WIN32 AND OPENSSL_USE_STATIC_LIBS)
-		target_link_libraries(OpenSSL::Crypto INTERFACE crypt32)
+		target_link_libraries(torrent-rasterbar PRIVATE crypt32)
 	endif()
 
 	target_link_libraries(torrent-rasterbar PUBLIC OpenSSL::SSL)
--- a/docs/building.rst
+++ b/docs/building.rst
@@ -484,12 +484,6 @@ Other build options are:
 
 Options are set on the ``cmake`` command line with the ``-D`` option or later on using ``ccmake`` or ``cmake-gui`` applications. ``cmake`` run outputs a summary of all available options and their current values.
 
-.. note::
-
-	If you are linking statically against OpenSSL on Windows and not using ``-Dstatic_runtime=ON``,
-	you should additionally use the option ``-DOPENSSL_USE_STATIC_LIBS=ON``.
-	If you use ``-Dstatic_runtime=ON``, ``-DOPENSSL_USE_STATIC_LIBS=ON`` is implied.
-
 Step 2: Building libtorrent
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 


CMakeLists.txt
```diff
@@ -636,7 +636,7 @@ if(OPENSSL_FOUND)
 
 	# TODO: needed until https://gitlab.kitware.com/cmake/cmake/issues/19263 is fixed
 	if(WIN32 AND OPENSSL_USE_STATIC_LIBS)
-		target_link_libraries(OpenSSL::Crypto INTERFACE crypt32)
+		target_link_libraries(torrent-rasterbar PRIVATE crypt32)
```
arvidn, line 639: if it's private, does it still propagate to dependees?

arvidn:
looks good to me, assuming that it works :)

I normally ping @zeule when it comes to cmake changes (feel free to ask me to stop doing that btw)

CMakeLists.txt
zeule: No, it does not. If libtorrent is statically built, the PRIVATE will not work.

FranciscoPombal: @zeule I can't explain why either, but it does work with `PRIVATE` does work when building libtorrent statically (I tried with the vcpkg triplet `x64-windows-static-md` as mentioned above, which I verified passed `BUILD_SHARED_LIBS=OFF` to libtorrent).

It also works with `INTERFACE` (and thus, if it works with both, it works with `PUBLIC` for sure).

This aligns with the other Windows libraries which are linked as `PRIVATE` some lines above (such as `ws2_32`). Speaking of which, the CMake bug report mentions it is necessary to link both `ws2_32` and `crypt32`. Since we were only getting errors about `crypt32`, this suggests that `ws2_32` was working, despite it too being linked as `PRIVATE` the whole time.

Sadly, that CMake bug report uses the legacy form of `target_link_libraries`, so we can't look to it for guidance as to whether it is better to use `PRIVATE` or `INTERFACE` (or `PUBLIC`). In the end I went with `PRIVATE`, because as mentioned, that's what the other stuff above (`ws2_32` et al) has been using the whole time without issues.

zeule: I don't understand then how does static linking on Windows/CMake work.

FranciscoPombal: @zeule yeah. I don't don't know why it works in this case either. Logically it shouldn't, but hey we're working around a `FindOpenSSL` bug, as long as it works I'm happy. You can check the generated `LibtorrentRasterbarTargets.cmake`, `crypt32` will be in the `INTERFACE_LINK_LIBRARIES` property.

zeule: https://cmake.org/pipermail/cmake/2016-May/063400.html

FranciscoPombal: @zeule Thanks for the link.

I think these are the important bits for our case:

> If, however, A is a
static library, the situation changes. Static libraries do not carry
information about other libraries they depend on. For this reason, when A
links in B as PRIVATE and another target C links in A, CMake will still add
B to the list of libraries to be linked for C because parts of B are needed
by A, but A itself doesn't have that dependency encoded into it. So even
though B is an internal implementation detail of A, C still needs B added
to the linker command, which CMake conveniently handles for you.

and 

> If you were paying careful attention, you would have noticed that when A
links in B as PRIVATE, the include directories of B never propagate to
something linking to A, but if A is a static library, then the *linking* of
B behaves as though the relationship was PUBLIC. This
PRIVATE-becomes-PUBLIC behaviour for static libraries only applies to the
*linking*, not to the other dependencies (compiler options/flags and
include search paths).

So looks like `PRIVATE` is indeed correct and there is a logical reason for that :)

zeule: The it would be logical to link it to OpenSSL::Crypto

FranciscoPombal: @zeule No, in our case, in the first quote consider:
A=openssl
B=crypt32
C=libtorrent

Substituting in the last sentence of first quote above,

"So even though `crypt32` is an internal implementation detail of `openssl`, `libtorrent` still needs `crypt32` added
to the linker command, **which CMake conveniently handles for you.**" - the bug means that CMake is **not** doing this for us, so we need to do it ourselves.

The bug report links this mailing list message: https://cmake.org/pipermail/cmake-developers/2016-August/029210.html
The mailing list message also advises client targets to link directly in`crypt32`. The alternative is the attached patch. Yes, one of the things the patch does is `set_target_properties(OpenSSL::Crypto PROPERTIES INTERFACE_LINK_LIBRARIES crypt32)`. But just this is not enough for it to work, the patch also does other stuff. So, rather than essentially including the whole patch ourselves, we just link directly in `crypt32` until the bug is resolved.

arvidn: libtorrent directly uses libcrypto as well, so I think it's reasonable to have a direct dependency on it, even though it's also an openssl internal dependency.

FranciscoPombal: @arvidn 
> libtorrent directly uses libcrypto as well, so I think it's reasonable to have a direct dependency on it, even though it's also an openssl internal dependency.

You mean adding `target_link_libraries(torrent-rasterbar PRIVATE OpenSSL::Crypto)` as well? Supposedly this is not needed, since we already have `target_link_libraries(torrent-rasterbar PUBLIC OpenSSL::SSL)`, and `OpenSSL::SSL` in turn already has `OpenSSL::Crypto` as a usage requirement. Is it even possible to have `OpenSSL::SSL` without `OpenSSL::Crypto`? If so, then I guess the addition would make sense.

FranciscoPombal: @arvidn to answer my own question - yes of course libtorrent (optionally) uses libcrypto for things other than SSL - the best example is hashing. However, the way that the top-level CML is currently written, usage of libcrypto (via the `TORRENT_USE_LIBCRYPTO` define) is entirely dependent on OpenSSL as a whole. In other words - currently it is only possible to use libcrypto if the whole OpenSSL package is found (libssl + libcrypto).

I think it is possible to undo this dependency relatively easily, but that is another matter entirely, and it's orthogonal to this fix in particular.

FranciscoPombal: @arvidn something like this, just not sure if it needs to be `PUBLIC`, and whether or not `pe_crypto` also needs to be added to the sources.

```patch
diff --git a/CMakeLists.txt b/CMakeLists.txt
index a1dc22eb9..d815024fe 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -632,34 +632,34 @@ set_package_properties(OpenSSL
                PURPOSE "Provides HTTPS support to libtorrent"
 )
 
-if(OPENSSL_FOUND)
-
+if(TARGET OpenSSL::SSL)
        # TODO: needed until https://gitlab.kitware.com/cmake/cmake/issues/19263 is fixed
        if(WIN32 AND OPENSSL_USE_STATIC_LIBS)
                target_link_libraries(OpenSSL::Crypto INTERFACE crypt32)
        endif()
 
        target_sources(torrent-rasterbar PRIVATE src/pe_crypto)
 endif()
 
-if(OPENSSL_FOUND)
+if(TARGET OpenSSL::Crypto)
+       target_link_libraries(torrent-rasterbar PUBLIC OpenSSL::Crypto)
        target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_LIBCRYPTO)
-else(OPENSSL_FOUND)
+else()
        find_public_dependency(LibGcrypt)
        set_package_properties(LibGcrypt
                PROPERTIES
                        URL "https://www.gnupg.org/software/libgcrypt/index.html"
                        DESCRIPTION "A general purpose cryptographic library"
                        TYPE RECOMMENDED
                        PURPOSE "Use GCrypt instead of the built-in functions for RC4 and SHA1"
        )
        if (LibGcrypt_FOUND)
                target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_LIBGCRYPT)
                target_link_libraries(torrent-rasterbar PRIVATE LibGcrypt::LibGcrypt)
        endif()
-endif(OPENSSL_FOUND)
+endif()
 
 if (encryption)
        target_sources(torrent-rasterbar PRIVATE src/pe_crypto)
```

arvidn: seems reasonable. I think it would have to be public, as at least some types are exposed to clients via the `hasher` type. I consider this a nice-to-have, as my understanding is that the CMakeList essentially work already (or at least with this PR)

FranciscoPombal:
@arvidn 
> sorry, I lost track of this PR. @FranciscoPombal would you mind rebasing it so the tests can pass?

Rebased.

CMakeLists.txt
FranciscoPombal: @arvidn 

>seems reasonable. I think it would have to be public, as at least some types are exposed to clients via the `hasher` type.

Alright I'll open a new PR for this change.

>I consider this a nice-to-have, as my understanding is that the CMakeList essentially work already (or at least with this PR).

I have been using CMake exclusively to build libtorrent for months on Linux, and most recently under a variety of scenarios on Windows. This particular patch fixes building on Windows with (static linkage, dynamic runtime). The CMake build works just fine in many scenarios at this point, it's very convenient to use when other projects also use CMake. Not sure how the situation is with regards to full feature parity with the `b2` system, but if it isn't there already, it must be close.

PR #4650, opened 2020-05-14
improve CMake - allow using libcrypto even if libssl is not found

FranciscoPombal:
Continuation of https://github.com/arvidn/libtorrent/pull/4641, basically this patch: https://github.com/arvidn/libtorrent/pull/4641#discussion_r424124121

@arvidn Slight problem: I'm unsure if the line `target_sources(torrent-rasterbar PRIVATE src/pe_crypto)` needs to be moved from/copied from where it is now into `if(TARGET OpenSSL::Crypto) ...` or if it has nothing to do with it and is fine as it is.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -632,8 +632,7 @@ set_package_properties(OpenSSL
 		PURPOSE "Provides HTTPS support to libtorrent"
 )
 
-if(OPENSSL_FOUND)
-
+if(TARGET OpenSSL::SSL)
 	# TODO: needed until https://gitlab.kitware.com/cmake/cmake/issues/19263 is fixed
 	if(WIN32 AND OPENSSL_USE_STATIC_LIBS)
 		target_link_libraries(OpenSSL::Crypto INTERFACE crypt32)
@@ -641,12 +640,12 @@ if(OPENSSL_FOUND)
 
 	target_link_libraries(torrent-rasterbar PUBLIC OpenSSL::SSL)
 	target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_OPENSSL)
-	target_sources(torrent-rasterbar PRIVATE src/pe_crypto)
 endif()
 
-if(OPENSSL_FOUND)
+if(TARGET OpenSSL::Crypto)
+	target_link_libraries(torrent-rasterbar PUBLIC OpenSSL::Crypto)
 	target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_LIBCRYPTO)
-else(OPENSSL_FOUND)
+else()
 	find_public_dependency(LibGcrypt)
 	set_package_properties(LibGcrypt
 		PROPERTIES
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -644,9 +643,10 @@ if(OPENSSL_FOUND)
 	target_sources(torrent-rasterbar PRIVATE src/pe_crypto)
```
arvidn, line 643: my understanding is that this line is a mistake. It's repeated on line 665, which is where I think it belongs. I think someone at some point mixed up using libcrypto and bittorrent protocol encryption

arvidn: this line has been removed in master recently too

FranciscoPombal: I must admit my problem was that I did not really look at the file and couldn't figure out what the file was for just from its name, but now it makes sense, **p**rotocol **e**ncryption, of course :). I looked at it now and yeah it does not depend on `libcrypto` (or `libgcrypt`) in any way as far as I can tell.

Fixed.

CMakeLists.txt
```diff
@@ -632,21 +632,20 @@ set_package_properties(OpenSSL
...
 		target_link_libraries(OpenSSL::Crypto INTERFACE crypt32)
 	endif()
 
 	target_link_libraries(torrent-rasterbar PUBLIC OpenSSL::SSL)
 	target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_OPENSSL)
-	target_sources(torrent-rasterbar PRIVATE src/pe_crypto)
 endif()
 
-if(OPENSSL_FOUND)
+if(TARGET OpenSSL::Crypto)
```
arvidn, line 645: what does this mean exactly?
is it checking if a target *exists*?

arvidn: assuming that's what it does, whether to link libtorrent against openssl, and use openssl is a user option, right? Isn't it possible that the openssl target could *exist* even though libtorrent is not supposed to link against it?

arvidn: for example, suppose a project using libtorrent pulls it in as a git submodule, and adds a `add)subdirectory()` in their CMakeLists.txt, then adds a dependency to the libtorrent target.

In such scenario, the top level project *could* also pull in an openssl cmake target, without necessarily wanting to build libtorrent against it. Perhaps unlikely, but it seems like those are independent.

arvidn: oh, perhaps I'm wrong. it looks like the CMakeList currently decides whether to link against openssl based on whether it can find it or not...

FranciscoPombal: @arvidn 

The current design of libtorrent's CML is such that some features (such as OpenSSL) are enabled by default based on whether or not certain packages are found. In other words, the set of default features is not fixed for all users, it will vary between environments.

This PR does not change anything pertaining to that overall design choice.

Related: Recently, there were some somewhat heated discussions about this in qBittorrent. In the end, the majority decided to change towards a fixed set of default features (that can all be overriden by the user of course), not dependent on found packages in each system.

PR #4652, opened 2020-05-16
factor out computing proofs for a subtree root

arvidn:


src/hash_picker.cpp
```diff
@@ -294,31 +294,11 @@ bool validate_hash_request(hash_request const& hr, file_storage const& fs)
...
-			h.update(proof_iter->first);
-			h.update(proof_iter->second);
-			tree_root = h.final();
-			++proof_iter;
-		}
+		// TODO: use strucutured bindings here in C++17
+		aux::vector<std::pair<sha256_hash, sha256_hash>> proofs;
+		sha256_hash tree_root;
+		std::tie(proofs, tree_root) = merkle_check_proofs(
+			tree[0], uncle_hashes, req.index >> base_num_layers);
```
arvidn, line 301: I meant to highlight that `req.index` is converted to the index of the root node here by `>> base_num_layers`

PR #4653, opened 2020-05-16
Jamfile: Add win10 option for UWP development

NotTsunami:
Windows Store (UWP) apps require the latest required version to be set to Windows 10 (defined by _WIN32_WINNT).


NotTsunami:
https://github.com/arvidn/libtorrent/blob/RC_1_2/CMakeLists.txt#L551

Are you opposed to allowing this to be configured when building with CMake, defaulting to its current value and matching the b2 options?

arvidn:
@NotTsunami not at all. I think it would be good if this could be configured with cmake too

arvidn:
I noticed that XP is still on that list, which I don't think is supported anymore

NotTsunami:
Awesome I'll probably send a PR for CMake tomorrow(-ish). Thanks!

FranciscoPombal:
@arvidn
> I noticed that XP is still on that list, which I don't think is supported anymore

To be fair, neither is Vista or 7 (this is not an argument in favor of keeping XP support, rather one against Vista/7).

arvidn:
oh, I thought vista was still supported. but I think I'm just thinking about `enum_net`

PR #4655, opened 2020-05-16
Test merge

arvidn:


CMakeLists.txt
```diff
@@ -659,12 +659,12 @@ if(NOT gnutls)
 			TYPE RECOMMENDED
 			PURPOSE "Provides HTTPS support to libtorrent"
 	)
-	if(OPENSSL_FOUND)
+
+	if(TARGET OpenSSL::SSL)
```
arvidn, line 663: @FranciscoPombal would you mind making sure I merged this correctly from RC_1_2?
The new feature in master is support for using GnuTLS instead of OpenSSL

CMakeLists.txt
```diff
@@ -697,20 +697,23 @@ if(gnutls OR NOT OPENSSL_FOUND)
...
-		PROPERTIES
-			URL "https://www.gnupg.org/software/libgcrypt/index.html"
-			DESCRIPTION "A general purpose cryptographic library"
-			TYPE RECOMMENDED
-			PURPOSE "Use GCrypt instead of the built-in functions for RC4 and SHA1"
-	)
-	if (LibGcrypt_FOUND)
-		target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_USE_LIBGCRYPT)
-		target_link_libraries(torrent-rasterbar PRIVATE LibGcrypt::LibGcrypt)
+else()
```
arvidn, line 700: this changed indentation, so you may want to ignore white space changes

FranciscoPombal: Currently you have:

```
if(NOT gnutls)
# find and enable openssl....

....

if(gnutls OR NOT TARGET OpenSSL::SSL)
# find and enable gnutls, fail if not found
```

which means that if you pass `-Dgnutls=OFF` and `OpenSSL` is not found inside the first if condition, the second if condition will be satisfied (due to `OR NOT TARGET OpenSSL::SSL`), and thus `gnutls` will be silently used, or the build will fail with `GnuTLS library not found`, which is probably not expected.

CMakeLists.txt
```diff
@@ -697,20 +697,23 @@ if(gnutls OR NOT OPENSSL_FOUND)
...
+	if(TARGET OpenSSL::Crypto)
```
FranciscoPombal, line 701: You may want to use `libcrypto` (or `libgcrypt`, for that matter) even if `gnutls OR NOT TARGET OpenSSL::SSL`, right?

CMakeLists.txt
arvidn: so, are you suggesting, instead of an `else()` block here, restore the `if (NOT OPENSSL_FOUND)` below, but change it to `if (NOT TARGET OpenSSL AND NOT gnutls)`?

CMakeLists.txt
arvidn: I'm not sure `libcrypto` is independent of openssl actually. I think it openssl is enabled libcrypto may be implied. That's not the case with gnutls though

FranciscoPombal: >I think it openssl is enabled libcrypto may be implied. 

If OpenSSL is found it means, `libssl` (the `OpenSSL::SSL` target) and/or `libcrypto` (the `OpenSSL::Crypto` target) _may_ be available. If `libssl` is available, then `libcrypto` is also necessarily available, since `libssl` depends on it. But libcrypto can be available wihtout `libssl`.

Since it is my impression that there is code in libtorrent that can use `libcrypto` (or `libgcrypt`) functionality independently of the availability of `libssl` and `gnutls`, I think it makes sense to make those crypto features independent of the (non) availability of `libssl`/`gnutls`.

arvidn: right. I think this updated merge does that. But the longer term goal should be to make the configuration explicit, rather than depend on what libraries can be found

FranciscoPombal: To be clear: I think the outermost `if (NOT GNUTLS_FOUND AND NOT TARGET OpenSSL::SSL)` is not needed at all actually. This is because you may want to use `libcrypto`/`libgcrypt` (for `hasher.cpp`, for example), interdependently of `gnutls` or `libssl` being available.

arvidn: sure, but there is no way to express this wish in the CMakeList.txt right now, it's all prescribed based on what libraries are found. I think that should be fixed, but I don't think this merge changes that situation materially. Does it?

Also, I don't think libtorrent (currently) supports using OpenSSL *without* also using libcrypto. Maybe that should be changed, but it also doesn't seem reasonable to use some other crypto library since libcrypto is already linked in.

CMakeLists.txt
FranciscoPombal: @arvidn 

I think the problem is that the current code cannot distinguish between the user passing `-Dgnutls=OFF` explicitly and simply not being defined, which has the same effect.

I would expect that if the user passes `-Dgnutls=OFF` explicitly, they would expect the build system:
- to use `libssl` if it exists (which is currently the case)
- to fail if they their system has `gnutls` but not `libssl` (currently not the case - `gntls` is going to be used regardless)

So I think it should be:

```
if(NOT gnutls)
# find and enable OpenSSL if available. note that if it's available, it will always be used
else()
# find and enable gnutls, fail if not found
endif()
```

In the future, you may want to make the conditons of find/use of OpenSSL more explicit (for example, failing the build unless some kind of feature like `NO_TLS_LIBRARY` is passed).

arvidn: I would not expect the second bullet point. I would expect `gnutls=OFF` to mean "build without using gnutls" (i.e. not failing). The semantics are very confusing given that all other options, other than gnutls, are implicit. So, from the point of view that the implicit linking against openssl is reasonable, saying `gnutls=OFF` would still make openssl eligible to be linked against, but not required.

I think all these options should be only explicit, but I think that should be a separate patch, definitely not changed in a merge commit.

CMakeLists.txt
FranciscoPombal: @arvidn 

>I think that should be fixed, but I don't think this merge changes that situation materially. Does it?

yeah this is a change for another time.
 
> Also, I don't think libtorrent (currently) supports using OpenSSL _without_ also using libcrypto.

Yeah, but it does make sense to use `libcrypto` without `libssl`. Doesn't it?

>but it also doesn't seem reasonable to use some other crypto library since libcrypto is already linked in.

Even if you link against `gnutls`, wouldn't you maybe want to also link against `libcrypto` for the code in `hashser.cpp`, for example? Or can `gnutls` also handle that without `libcrypto` (or `libgcrypt`)?

arvidn: > Yeah, but it does make sense to use libcrypto without libssl. Doesn't it?

I agree.

> Or can gnutls also handle that without libcrypto (or libgcrypt)?

I used to be under the impression that gnutls was as tied to libgcrypt as openssl is to libcrypto, but that doesn't appear to be the case. So, yes, it might make sense to use gnutls and libcrypto, even though gnutls surely will also link in some other crypto library.

PR #4660, opened 2020-05-17
fix build with boost 1.73 (master)

### LINKED ISSUE #4658: boost 1.73 ssl::rfc2818_verification deprecated
Using master against boost 1.73 I get build errors on **using boost::asio::ssl::rfc2818_verification**. I switched to **host_name_verification** which seems to be the intended replacement. This is perhaps a boost bug since rfc2818_verification is only [supposed to be deprecated](https://www.boost.org/doc/libs/1_73_0/doc/html/boost_asio/reference/ssl__rfc2818_verification.html).

    diff --git a/include/libtorrent/ssl.hpp b/include/libtorrent/ssl.hpp
    index e3b385eb2..889f66cc1 100644
    --- a/include/libtorrent/ssl.hpp
    +++ b/include/libtorrent/ssl.hpp
    @@ -97,7 +97,7 @@ using error_code = boost::system::error_code;
            using boost::asio::ssl::stream;
     #endif
     using boost::asio::ssl::verify_context;
    -using boost::asio::ssl::rfc2818_verification;
    +using boost::asio::ssl::host_name_verification;
     
     using native_context_type = SSL_CTX*;
     using native_stream_type = SSL*;
    
    
    
    
    diff --git a/src/socket_type.cpp b/src/socket_type.cpp
    index d8051374c..a442989f1 100644
    --- a/src/socket_type.cpp
    +++ b/src/socket_type.cpp
    @@ -164,7 +164,7 @@ namespace aux {
                    template <typename T>
                    void operator()(ssl_stream<T>& s)
                    {
    -                       s.set_verify_callback(ssl::rfc2818_verification(hostname_), *ec_);
    +                       s.set_verify_callback(ssl::host_name_verification(hostname_), *ec_);
                            ssl_ = s.handle();
                            ctx_ = s.context_handle();
                    }

ssiloti:
Fixes #4658

### DIFF
--- a/include/libtorrent/ssl.hpp
+++ b/include/libtorrent/ssl.hpp
@@ -97,7 +97,11 @@ using error_code = boost::system::error_code;
 	using boost::asio::ssl::stream;
 #endif
 using boost::asio::ssl::verify_context;
-using boost::asio::ssl::rfc2818_verification;
+#if BOOST_VERSION >= 107300
+using boost::asio::ssl::host_name_verification;
+#else
+using host_name_verification = boost::asio::ssl::rfc2818_verification;
+#endif
 
 using native_context_type = SSL_CTX*;
 using native_stream_type = SSL*;
@@ -111,7 +115,7 @@ using boost::asio::gnutls::context;
 using boost::asio::gnutls::stream_base;
 using boost::asio::gnutls::stream;
 using boost::asio::gnutls::verify_context;
-using boost::asio::gnutls::rfc2818_verification;
+using boost::asio::gnutls::host_name_verification;
 
 using native_context_type = context::native_handle_type;
 using native_stream_type = stream_base::native_handle_type;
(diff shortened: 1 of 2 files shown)


include/libtorrent/ssl.hpp
```diff
@@ -111,7 +115,7 @@ using boost::asio::gnutls::context;
 using boost::asio::gnutls::stream_base;
 using boost::asio::gnutls::stream;
 using boost::asio::gnutls::verify_context;
-using boost::asio::gnutls::rfc2818_verification;
+using boost::asio::gnutls::host_name_verification;
```
ghbplayer, line 118: Does this work with older versions of boost?

arvidn: no, but master requires boost that isn't too old. I don't know the cut-off off hand

arvidn: actually, this specific line isn't boost, it's the gnutls socket wrappers, and they do support `host_name_verification`

PR #4663, opened 2020-05-18
add function to clear the peer list for a torrent

arvidn:


an0n666:
@arvidn 
I noticed an issue with this. If a torrent is in paused state and tracker is changed...peers are not cleaned.

arvidn:
I don't understand the second condition. "the tracker has changed". What do you mean by that exactly?

arvidn:
also, what symptom do you see to make you think the peer list isn't cleared?

arvidn:
and you also called `clear_peers()`, right? and it didn't work?

an0n666:
> and you also called `clear_peers()`, right? and it didn't work?

Yes. qBt calls that function now everytime changes are made to trackers. It works when a torrent is active. Doesn’t work when it’s paused.

an0n666:
@arvidn 
Extremely sorry for wasting your valuable time. Somehow my libtorrent got replaced with a older build :(

PR #4671, opened 2020-05-22
make the v2-only-magnet transfer test use a larger torrent

arvidn:
to exercise more parts of the hash_picker


arvidn:
I think this makes the test too expensive to run. I may look into optimizing it with a custom storage to avoid any disk I/O, but I suspect it's just the sheer number of invariant checks and asserts that cost the most.

PR #4672, opened 2020-05-24
add set_alert_fd in python binding, to supersede set_alert_notify

arvidn:


bindings/python/src/session.cpp
```diff
@@ -413,6 +413,17 @@ namespace
         s.set_alert_notify(std::bind(&alert_notify, cb));
     }
 
+    void alert_fd_notify(int const fd)
+    {
+        std::uint8_t dummy = 0;
+        ::write(fd, &dummy, 1);
```
WGH-, line 419: Unfortunately, `signal.set_wakeup_fd` is somewhat complicated for a reason. Although on UNIX-like systems `write` is ok for sockets, on Windows, you have to use `send` for sockets.

docs/python_binding.rst
```diff
@@ -117,9 +117,30 @@ To get a python dictionary of the settings, call ``session::get_settings``.
...
+``NAME_OF_METRIC`` is the name of a metric.
+
+set_alert_notify
+================
+
+The ``set_alert_notify()`` function is not compatible with python. Since it
+requires locking the GIL from within the libtorrent thread, to call the callback,
+it can cause a deadlock with the main thread.
+
+Instead, use the python-specific ``set_alert_fd()`` which takes a file descriptor
```
WGH-, line 135: I'd mention that the file descriptor must be non-blockable.

docs/python_binding.rst
```diff
@@ -117,9 +117,30 @@ To get a python dictionary of the settings, call ``session::get_settings``.
...
+that will have 1 byte written to it to notify the client that there are new
+alerts to be popped.
+
+This can be used with ``socket.socketpair()``, for exanple. The file descriptor
```
WGH-, line 139: I'd mention that `os.pipe` is also fine and even preferred for UNIX-like systems, due to being simpler and lower-overhead object. It's bad on Windows due to not being able to be configured non-blocking. though.

docs/python_binding.rst
```diff
@@ -117,9 +117,30 @@ To get a python dictionary of the settings, call ``session::get_settings``.
...
+set_alert_notify
+================
+
+The ``set_alert_notify()`` function is not compatible with python. Since it
+requires locking the GIL from within the libtorrent thread, to call the callback,
+it can cause a deadlock with the main thread.
+
+Instead, use the python-specific ``set_alert_fd()`` which takes a file descriptor
+that will have 1 byte written to it to notify the client that there are new
+alerts to be popped.
```
WGH-, line 137: I'd add here that the other end of the file descriptor can be easily integrated into existing event loop, like `asyncio`, `selectors`, `gevent`, etc.

bindings/python/src/session.cpp
arvidn: I see, so this patch doesn't support sockets on windows. Sockets on windows also aren't `int`s, they are `HANDLE`s. I wonder what `fileno()` returns on a socket in python on windows then..

arvidn:
does anyone have concrete suggestion on what to do instead of this?
it seems like it would be a safer API to ask for the file descriptor rather than passing one in. Since passing it in will have a bunch of (likely unintuitive) restrictions anyway.

I would like to keep this simple and practical, rather than generic.

it looks like socket handles may be usable with `WriteFile()` on windows ([here](https://docs.microsoft.com/en-us/windows/win32/winsock/socket-handles-2)). So maybe the `int` can be assumed to be a `HANDLE` on windows and a file descriptor otherwise. And use `WriteFile` and `write` respectively.

The problem is that this all depends on implementation details of python itself (afaiu).

WGH-:
> does anyone have concrete suggestion on what to do instead of this?

I think doing it in exactly the same way as in CPython's `signal.set_wakeup_fd` is a good idea.

https://github.com/python/cpython/blob/a59fc9160db50a5e20358303e7ddbda180a1b74f/Modules/signalmodule.c#L289-L327
https://github.com/python/cpython/blob/a59fc9160db50a5e20358303e7ddbda180a1b74f/Modules/signalmodule.c#L665-L775

bindings/python/src/session.cpp
WGH-: `HANDLE` is actually an integer, just even more opaque than fd on UNIX-like systems (the latter are guaranteed to be assigned sequentially, 0, 1 and 2 have predefined meaning, etc.).

arvidn: `HANDLE` is a void pointer, which is likely twice as wide as `int`

AllSeeingEyeTolledEweSew: https://github.com/python/cpython/blob/master/Modules/socketmodule.c#L3375
https://github.com/python/cpython/blob/master/Modules/socketmodule.h#L207

Looks like `socket.fileno()` is always a **python** `int`; on windows it's cast from `SOCKET`.

https://github.com/python/cpython/blob/master/Modules/_io/fileio.c#L382

For normal files python calls `_wopen` on windows, so you get file descriptor that fits in a C `int`.

https://github.com/python/cpython/blob/master/Modules/signalmodule.c#L670

Python's `signal.set_wakeup_fd` queries the fd with winsock, and if that fails, assumes it's a normal int file descriptor.

docs/python_binding.rst
arvidn: I don't think it has to be non-blocking. but not reading from it at sufficient frequency will block libtorrent's internal thread. Making it non-blocking won't really solve that problem, it will just make you miss events instead

docs/python_binding.rst
arvidn: I don't think the non-blocking is a serious issue, and I would think `os.pipe` would probably be a reasonable cross-platform solution

docs/python_binding.rst
WGH-: > I don't think it has to be non-blocking. but not reading from it at sufficient frequency will block libtorrent's internal thread.

That's true. "Pass a blocking FD at your own risk" :)

> Making it non-blocking won't really solve that problem, it will just make you miss events instead

Not really. If the fd is full, it means that the other end is already in "notified" state.

bindings/python/src/session.cpp
```diff
@@ -413,6 +413,39 @@ namespace
...
+    {
+        std::uint8_t dummy = 0;
+		::send(fd, &dummy, 1, 0);
+    }
+#endif
+
+    void alert_fd_notify(int const fd)
+    {
+        std::uint8_t dummy = 0;
+        while (::write(fd, &dummy, 1) < 0 && errno == EAGAIN);
```
WGH-, line 427: Did you mean to say `errno == EINTR`?

docs/python_binding.rst
arvidn: good point

bindings/python/src/session.cpp
arvidn: heh, yes :)

bindings/python/src/session.cpp
```diff
@@ -413,6 +413,39 @@ namespace
...
+    }
+#endif
+
+    void alert_fd_notify(int const fd)
+    {
+        std::uint8_t dummy = 0;
+        while (::write(fd, &dummy, 1) < 0 && errno == EINTR);
+    }
+
+    void set_alert_fd(lt::session& s, int fd)
```
AllSeeingEyeTolledEweSew, line 430: Don't we want the arg to be larger than `int`? I thought `SOCKET`s are pointers.

arvidn:
I think this patch is in a good state now. Would anyone mind trying it out?

arvidn:
since the unit test pass, I'm inclined to feel confident that it works

WGH-:
Seems to work fine, tested on Linux.

```python
async def await_for_alert(ses: lt.session):
    loop = asyncio.get_event_loop()
    event = asyncio.Event()
    r, w = os.pipe2(os.O_NONBLOCK | os.O_CLOEXEC)
    def on_ready():
        event.set()
        while True:
            try:
                os.read(r, 1)
            except BlockingIOError:
                return
    loop.add_reader(r, on_ready)
    ses.set_alert_fd(w)
    try:
        while True:
            event.clear()
            alerts = ses.pop_alerts()
            for alert in alerts:
                yield alert
            await event.wait()
    finally:
        loop.remove_writer(r)
        os.close(r)
        os.close(w)
```

Issue #2574, opened 2017-11-30
libtorrent and tcp::acceptor crashes with access violation on Windows  


0xAT:
**Please provide the following information**

libtorrent version (or branch):
master / libtorrent-1_1_5

platform/architecture:
Win32

compiler and compiler version:
msvc-14.0

boost version:
1.65.1

The code below always leads to crash.

```
#include "libtorrent/session.hpp"

#include <chrono>
#include <thread>

using namespace std;
int main()
{
	libtorrent::session session;

	boost::asio::io_service io_service;
        boost::asio::ip::tcp::acceptor acceptor(io_service);

	std::this_thread::sleep_for(chrono::seconds(15));
	return 0;
}
```

Call stack:

```
>	test-crow-libtorrent.exe!boost::asio::detail::win_iocp_socket_service_base::do_open(struct boost::asio::detail::win_iocp_socket_service_base::base_implementation_type &,int,int,int,class boost::system::error_code &)	C++
 	test-crow-libtorrent.exe!libtorrent::aux::session_impl::setup_listener(class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const &,class boost::asio::ip::tcp const &,int,int,class boost::system::error_code &)	C++
 	test-crow-libtorrent.exe!libtorrent::aux::session_impl::open_listen_port(void)	C++
 	test-crow-libtorrent.exe!libtorrent::aux::session_impl::init(class boost::shared_ptr<struct libtorrent::settings_pack>)	C++
 	test-crow-libtorrent.exe!boost::_mfi::mf1<void,struct libtorrent::aux::session_impl,class boost::shared_ptr<struct libtorrent::settings_pack> >::operator()(struct libtorrent::aux::session_impl *,class boost::shared_ptr<struct libtorrent::settings_pack>)	C++
[...]

PR #4682, opened 2020-05-27
Emphasizing the need to define matching macros

fmike7:
Based on this issue: https://github.com/arvidn/libtorrent/issues/2574


fmike7:
I also updated it to Visual Studio 2019, since this is what most people would be using now.

docs/building.rst
```diff
@@ -647,3 +654,18 @@ in a command shell::
 
 This will also install the headers and library files in the visual studio directories to
 be picked up by libtorrent.
+
+list of macros
+--------------------
+
+The following is a list of defines that libtorrent is built with: ``BOOST_ALL_NO_LIB``, ``BOOST_ASIO_ENABLE_CANCELIO``, ``BOOST_ASIO_HAS_STD_CHRONO``, ``BOOST_MULTI_INDEX_DISABLE_SERIALIZATION``, ``BOOST_NO_DEPRECATED``, ``BOOST_SYSTEM_NO_DEPRECATED``, ``TORRENT_BUILDING_LIBRARY``, ``TORRENT_USE_I2P=1``
```
arvidn, line 661: are you sure that `TORRENT_USE_I2P` affects the ABI? if so, would you mind pointing out where? I thought I had fixed all the `TORREN_*` macros to not affect ABI (except for `TORRENT_ABI_VERSION`).

`TORRENT_BUILDING_LIBRARY` is *not* supposed to be defined by clients. It's there to allow libtorrent headers to tell when they are built as part of libtorrent itself or when they are included. On windows this is an important distinction since it's the difference between `__declspec(dllimpoty)` and `__declspec(dllexport)`

docs/building.rst
```diff
@@ -647,3 +654,18 @@ in a command shell::
...
+
+Make sure you define the same at complile time for your code to avoid any runtime errors and other issues.
+
+These might change in the future, so it's always best to verify these every time you upgrade to
+a new version of libtorrent. The simplest way to see the full list of macros defined is to build
+libtorrent with "--debug-configuration -d2+2" switches added to b2 command line:
+
+``b2 --debug-configuration -d2+2 toolset=msvc-14.2 link=static runtime-link=static boost-link=static variant=release``
```
arvidn, line 669: I would prefer command lines to be in "literal blocks", i.e. put a "::", blank line and indentation (there are other examples in this file)

docs/building.rst
```diff
@@ -647,3 +654,18 @@ in a command shell::
...
+list of macros
+--------------------
+
+The following is a list of defines that libtorrent is built with: ``BOOST_ALL_NO_LIB``, ``BOOST_ASIO_ENABLE_CANCELIO``, ``BOOST_ASIO_HAS_STD_CHRONO``, ``BOOST_MULTI_INDEX_DISABLE_SERIALIZATION``, ``BOOST_NO_DEPRECATED``, ``BOOST_SYSTEM_NO_DEPRECATED``, ``TORRENT_BUILDING_LIBRARY``, ``TORRENT_USE_I2P=1``
+
+Make sure you define the same at complile time for your code to avoid any runtime errors and other issues.
+
+These might change in the future, so it's always best to verify these every time you upgrade to
+a new version of libtorrent. The simplest way to see the full list of macros defined is to build
+libtorrent with "--debug-configuration -d2+2" switches added to b2 command line:
```
arvidn, line 667: Unless I'm misunderstanding what you're going for, I think `-n` is simpler. That will just print the command lines instead of executing them. If combined with `-a`, all command lines will be printed.

arvidn:
I think I've tried to stick to 80 columns for the .rst files. Would you mind making breaking your lines around that mark to? (if you use vim you can `set textwidth=80`, select the paragraphs and type `gq`).

Also, there's a custom hunspell dictionary in `docs/hunspell/libtorrent.dic` which you need to update. The spell checker is failing the CI right now, and there's at least one typo :)

https://travis-ci.org/github/arvidn/libtorrent/jobs/691889428#L849

docs/building.rst
fmike7: TORRENT_USE_I2P - yeah, I wasn't too sure about it either. Ok, let's remove it.

TORRENT_BUILDING_LIBRARY - yes, makes sense, let's remove it.

docs/building.rst
fmike7: Ok, I did not know that. Let's change to the way you are suggesting.

docs/building.rst
fmike7: Yeah, I see. Indentation is what I missed in front of the command line, because I tried to use double colon and it wasn't working.

fmike7:
> I think I've tried to stick to 80 columns for the .rst files. Would you mind making breaking your lines around that mark to? (if you use vim you can `set textwidth=80`, select the paragraphs and type `gq`).
> 
> Also, there's a custom hunspell dictionary in `docs/hunspell/libtorrent.dic` which you need to update. The spell checker is failing the CI right now, and there's at least one typo :)
> 
> https://travis-ci.org/github/arvidn/libtorrent/jobs/691889428#L849

Got it, ok!

fmike7:
Sorry, I am still from the pre-GitHub dark ages - are you able to see my new commit 26249bd4c8057141021e8e7b2c2386fec6a8611e ?

arvidn:
yes. The pull request is pinned to the branch, and updates as you push more commits to it

docs/hunspell/libtorrent.dic
```diff
@@ -561,3 +561,6 @@ NOTSENT
 LOWAT
 Dstatic
 DOPENSSL
+d2
```
arvidn, line 564: you don't use "d2" anymore now

fmike7: ok

docs/building.rst
```diff
@@ -647,3 +655,22 @@ in a command shell::
 
 This will also install the headers and library files in the visual studio directories to
 be picked up by libtorrent.
+
+list of macros
+--------------------
```
arvidn, line 660: you use too many `-` characters here. To make it a header you need as many `-` as there are characters in the heading itself

arvidn:
would you mind squashing the commits to a single one? (you can use `git rebase RC_1_2 -i` for example)

docs/building.rst
```diff
@@ -524,7 +531,8 @@ build configurations
 By default libtorrent is built In debug mode, and will have pretty expensive
 invariant checks and asserts built into it. If you want to disable such checks
 (you want to do that in a release build) you can see the table below for which
-defines you can use to control the build.
+defines you can use to control the build. Make sure to define the same macros in your
+own code that compiles and links with bittorrent.
```
arvidn, line 535: I think you meant to say "libtorrent" here, not "bittorrent"

docs/building.rst
```diff
@@ -182,7 +189,7 @@ can set the ``runtime-link`` feature on the command line, either to ``shared``
 or ``static``. Most operating systems will only allow linking shared against
 the runtime, but on windows you can do both. Example::
 
-  b2 msvc-14.0 link=static runtime-link=static
+  b2 msvc-14.2 link=static runtime-link=static debug-symbols=on
```
arvidn, line 192: `debug-symbols=on` is on by default (since the default build variant is `debug`).

docs/building.rst
```diff
@@ -28,8 +28,11 @@ you can usually get help in the ``#libtorrent`` IRC channel on
 	libtorrent has some code in header files, that code will not be
 	compatible with the built library if they see different configurations.
 
-	Always make sure that the same TORRENT_* macros are defined when you
-	link against libtorrent as when you build it.
+	Always make sure that the same TORRENT_* and BOOST_* macros are defined
+	when you link against libtorrent as when you build it. The simplest way
+	to see the full list of macros defined is to build libtorrent with
+	``-n -a`` switches added to b2 command line, which output all compiler
```
arvidn, line 34: I think it would make sense to make "b2" -> "``b2``". Or would that be inconsistent with the rest of this file?

docs/building.rst
fmike7: Got it! Definitely need to ramp my knowledge of RST format.

docs/building.rst
fmike7: Yes, sorry about this.

docs/building.rst
fmike7: ok, I added variant=release, because I wanted to give an example of how to build with debug symbols (because I think this is what most people want to do - you always want to keep debug symbols to analyze crashes in production)

docs/building.rst
fmike7: sure, done

fmike7:
> would you mind squashing the commits to a single one? (you can use `git rebase RC_1_2 -i` for example)

This one I got stuck on. Not enough knowledge of GIT, which caused merge conflicts and and could not resolve these properly. I created a fresh full request with one commit, let's use that?

fmike7:
> > would you mind squashing the commits to a single one? (you can use `git rebase RC_1_2 -i` for example)
> 
> This one I got stuck on. Not enough knowledge of GIT, which caused merge conflicts and and could not resolve these properly. I created a fresh full request with one commit, let's use that?

Sorry, I really have to pump up my knowledge of GIT before volunteering to contribute!

Can you just copy and paste it from here to your branch: https://github.com/fmike7/libtorrent/blob/patch-2/docs/building.rst

arvidn:
if you have a new branch with a single commit, you can just do:

```
git checkout patch-2
git reset <new-branch> --hard
git push origin patch-2 -f
```

arvidn:
> Did git rebase --skip to skip it (assuming the change is already there)

That's not a safe assumption. There could have been some other change to building.rst. If you open the file there should be merge conflict markers (`<<<<<<`, `======`, and `>>>>>>`). You have to patch up the file to what its final state should be, `git add` it and then `git rebase --continue`

arvidn:
if a rebase turned out to be more complicated than you thought or if you change your mind, you can always do `git rebase --abort` to go back to the state before you started the rebase.

If the resulting commit has everything you want in it (do `git diff origin RC_1_2`), then you can force push it. If you lost anything along the way (which it sounds like you may have), you can point back your branch point to the first commit again by using `git reflog` and `git reset <commit-hash> --hard`

arvidn:
does this look good? https://github.com/arvidn/libtorrent/pull/4694

fmike7:
> does this look good? #4694

Yep, I think we're all set. Thanks for guiding on the usage of git, much appreciated! I definitely have to study this topic in more detail.

Issue #1437, opened 2016-12-21
incoming_connection_alert but for a torrent_handle


vktr:
libtorrent version (or branch): a2c4e7490e2ae7aecd512715d754202ec4f06995
platform/architecture: W10 x64
compiler and compiler version: MSVC 14.0 Win64

Is there a `torrent_alert` derived alert for an incoming peer connection? Something that would be posted when the handshake successfully completed and a peer is associated with a torrent. I'd like to keep a torrent peer list synced without having to poll the peers in a `torrent_handle` and check for changes.

**Edit** This is for *incoming* connections. I've seen `peer_connect_alert` but that is posted for outgoing connections only.

If nothing is there, is this something you would be open to accept as a PR?

aldenml:
The plugins framework could help you, or isn't that a good solution in your case?

vktr:
I didn't think about that - that should work, thanks! 😄

arvidn:
the plugin api is not quite as good as it may not be as stable of an API. maintaining a peer list that stays in sync with libtorrent is definitely something I think _should_ be possible.

vktr:
I'll write a POC and let you know. Thanks!

arvidn:
btw, the reason the ``incoming_connection_alert`` doesn't have a torrent associated with it is because it's posted before the handshake completes with the peer. There *could* be a separate alert posted when an incoming peer is associated/attached to a torrent also, but there isn't one now.

vktr:
Would you consider a PR for adding one?

arvidn:
sure

PR #4686, opened 2020-05-28
Post peer_connect_alert on both in/out

vktr:
Changes the peer_connect_alert to post on both incoming and outgoing connections and
also adds a field to specify the direction. This helps existing usages to still work
as they expect (ie. only react to outgoing connections).

I mentioned this a few years ago in #1437. Also, I'm not really sure if I'm posting the incoming alert at the right place, i.e if it has all the necessary info to be posted at that moment. So please have a look and see if it makes sense.

### DIFF
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -782,19 +782,27 @@ TORRENT_VERSION_NAMESPACE_2
 #endif
 	};
 
-	// This alert is posted every time an outgoing peer connect attempts succeeds.
+	// This alert is posted every time an incoming peer connection both
+	// successfully passes the protocol handshake and is associated with a
+	// torrent, or an outgoing peer connection attempt succeeds. For arbitrary
+	// incoming connections, see incoming_connection_alert.
 	struct TORRENT_EXPORT peer_connect_alert final : peer_alert
 	{
+		enum class direction_t { in, out };
+
 		// internal
 		TORRENT_UNEXPORT peer_connect_alert(aux::stack_allocator& alloc, torrent_handle h
-			, tcp::endpoint const& ep, peer_id const& peer_id, socket_type_t type);
+			, tcp::endpoint const& ep, peer_id const& peer_id, socket_type_t type, direction_t direction);
 
 		TORRENT_DEFINE_ALERT(peer_connect_alert, 23)
 
 		static constexpr alert_category_t static_category = alert_category::connect;
 		std::string message() const override;
 
-		socket_type_t const socket_type;
+		// Tells you if the peer was incoming or outgoing
+		direction_t direction;
+
+		socket_type_t socket_type;
 	};
 
 	// This alert is generated when a peer is disconnected for any reason (other than the ones
--- a/src/alert.cpp
+++ b/src/alert.cpp
@@ -1463,16 +1463,18 @@ namespace {
 	}
 
 	peer_connect_alert::peer_connect_alert(aux::stack_allocator& alloc, torrent_handle h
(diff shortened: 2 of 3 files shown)


include/libtorrent/alert_types.hpp
```diff
@@ -782,18 +782,23 @@ TORRENT_VERSION_NAMESPACE_2
...
-			, tcp::endpoint const& ep, peer_id const& peer_id, socket_type_t type);
+			, tcp::endpoint const& ep, peer_id const& peer_id, socket_type_t type, direction_t direction);
 
 		TORRENT_DEFINE_ALERT(peer_connect_alert, 23)
 
 		static constexpr alert_category_t static_category = alert_category::connect;
 		std::string message() const override;
 
+		// Tells you if the peer was incoming or outgoing
+		direction_t const direction;
```
arvidn, line 800: I think it's a mistake to have `const` members, I'm moving away from that in the alerts

include/libtorrent/alert_types.hpp
```diff
@@ -782,18 +782,23 @@ TORRENT_VERSION_NAMESPACE_2
 #endif
 	};
 
-	// This alert is posted every time an outgoing peer connect attempts succeeds.
+	// This alert is posted every time an incoming or outgoing peer connect attempts succeeds.
```
arvidn, line 785: this description could be more precise. Incoming connections aren't "attempts", and I think it's important to point out that this is not posted when we accept an incoming connection, but when an incoming connection handshakes and is associated with a torrent

include/libtorrent/alert_types.hpp
arvidn: I think `peer_connect_alert::message()` (in `alert.cpp`) should be updated to print this field as well

vktr: RE: `const`, do you think I should remove it for the direction field here, or is it something to keep just to stay in line with the rest of the alert? Or should I try to remove all the `const` from this alert?

arvidn: I think it's fine to remove the other `const` for this alert too

vktr:
@arvidn  I've implemented the requested changes - I'm a bit unsure how to best print the direction in the message in a somewhat pretty way. I realize it's mostly for debugging but no one likes non-pretty debug messages :)

arvidn:
well, right now it says: "connecting to peer". maybe you can change it to "%s connection to peer", where the `direction` translates to: "outgoing" and "incoming"

src/alert.cpp
```diff
@@ -1463,16 +1463,18 @@ namespace {
 	}
 
 	peer_connect_alert::peer_connect_alert(aux::stack_allocator& alloc, torrent_handle h
-		, tcp::endpoint const& ep, peer_id const& peer_id, socket_type_t const type)
+		, tcp::endpoint const& ep, peer_id const& peer_id, socket_type_t const type, direction_t const dir)
 		: peer_alert(alloc, h, ep, peer_id)
 		, socket_type(type)
+		, direction(dir)
```
arvidn, line 1469: this initialization list entry is not in the order of initialization: https://travis-ci.org/github/arvidn/libtorrent/jobs/692316456#L961

vktr:
I fixed the message and field order initialization. If all looks good, do you want me to squash the commits?

arvidn:
yes please

vktr:
Sweet, all cleaned up!

arvidn:
"handshaked" is failing spell-check. is there a better phrase to use? https://travis-ci.org/github/arvidn/libtorrent/jobs/692509992#L966

vktr:
Tried to work my way around that wording now.. :)

PR #4689, opened 2020-05-29
Restore info_hash() API

arvidn:
I think changing the return value of `info_hash()` is too intrusive of a change. Instead, add another member function, `info_hashes()`, which returns the new `info_hash_t` object.

Also, add back the old `storage.hpp` header with a friendly error message.


xavier2k6:
typo in `storage.hpp` `customization poin` -> `customization point`

arvidn:
@ssiloti before I cut the release branch for master, I tried to build qBT against it. This was the main issue causing source code incompatibility with current libtorrent 1.2. I think it makes sense to restore this to a more conservative/backwards compatible upgrade of the API.

include/libtorrent/torrent_info.hpp
```diff
@@ -378,8 +378,11 @@ TORRENT_VERSION_NAMESPACE_3
 		index_range<piece_index_t> piece_range() const
 		{ return m_files.piece_range(); }
 
-		// returns the info-hash of the torrent
-		const info_hash_t& info_hash() const { return m_info_hash; }
+		// returns the info-hash of the torrent. For BitTorrent v2 support, use
+		// ``info_hashes()`` to get an object that may hold both a v1 and v2
+		// info-hash
+		sha1_hash info_hash() const { return m_info_hash.get_best(); }
```
ssiloti, line 384: Since the goal here is to provide backwards compatibility with unmodified clients, I think this should return the v1 hash for hybrid torrents instead of the truncated v2 hash. Non-v2 aware clients are likely to display this to the user as a v1 hash which will cause confusion when it doesn't work with v1 only clients.

arvidn: makes sense. I was thinking that, as long as the info-hash is just used as a unique identifier, a client could support v2 torrents without knowing about it. but maybe that's risky

src/torrent_handle.cpp
```diff
@@ -205,7 +205,13 @@ namespace libtorrent {
 		return r;
 	}
 
-	info_hash_t torrent_handle::info_hash() const
+	sha1_hash torrent_handle::info_hash() const
+	{
+		std::shared_ptr<torrent> t = m_torrent.lock();
+		return t ? t->info_hash().get_best() : sha1_hash{};
```
ssiloti, line 211: Shouldn't this forward to `t->info_hash()`?

ssiloti, line 211: This should return `t->info_hash().v1` for consistency with `torrent_info`

PR #4704, opened 2020-06-01
use standard string_view

arvidn:


neheb:
Does this mean master is C++17?

arvidn:
@neheb yes, it is. although. it seems msvc still has some issues with this particular patch

PR #4707, opened 2020-06-01
Use time_duration instead of int for predictive pieces

vktr:
Scanning for some low hanging fruits - found this. Seemed kinda straightforward so made a best effort fix.

### DIFF
--- a/src/peer_connection.cpp
+++ b/src/peer_connection.cpp
@@ -3013,13 +3013,20 @@ namespace {
 						// do we have left to download?
 						std::int64_t const rate = peer->connection->statistics().download_payload_rate();
 						std::int64_t const bytes_left = std::int64_t(st.requested) * t->block_size();
-						// the settings unit is milliseconds, so calculate the
-						// number of milliseconds worth of bytes left in the piece
-						if (rate > 1000
-							&& (bytes_left * 1000) / rate < m_settings.get_int(settings_pack::predictive_piece_announce))
+
+						// calculate the eta for the piece
+						time_duration const eta = rate > 0
+							? milliseconds((bytes_left * 1000) / rate)
+							: milliseconds(0);
+
+						// the configured threshold for predictive piece announce
+						time_duration const threshold = milliseconds(
+							m_settings.get_int(settings_pack::predictive_piece_announce));
+
+						if (rate > 1000 && eta < threshold)
 						{
 							// we predict we will complete this piece very soon.
-							t->predicted_have_piece(piece, int((bytes_left * 1000) / rate));
+							t->predicted_have_piece(piece, eta);
 						}
 					}
 				}
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -4166,8 +4166,7 @@ namespace {
 	// announce it to peers ahead of time to eliminate the
 	// round-trip times involved in announcing it, requesting it
 	// and sending it
-	// TODO: 2 use chrono type for time duration
-	void torrent::predicted_have_piece(piece_index_t const index, int const milliseconds)
+	void torrent::predicted_have_piece(piece_index_t const index, time_duration const duration)
 	{
 		auto const i = std::lower_bound(m_predictive_pieces.begin()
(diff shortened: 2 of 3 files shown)


src/peer_connection.cpp
```diff
@@ -3013,13 +3013,19 @@ namespace {
...
-						if (rate > 1000
-							&& (bytes_left * 1000) / rate < m_settings.get_int(settings_pack::predictive_piece_announce))
+
+						// calculate the eta for the piece
+						milliseconds32 const eta = rate > 0
+							? milliseconds32((bytes_left * 1000) / rate)
+							: milliseconds32(0);
+
+						// the configured threshold for predictive piece announce
+						int const threshold = m_settings.get_int(settings_pack::predictive_piece_announce);
```
arvidn, line 3023: ideally threshold here should be a chrono type as well

src/torrent.cpp
```diff
@@ -4166,8 +4166,7 @@ namespace {
 	// announce it to peers ahead of time to eliminate the
 	// round-trip times involved in announcing it, requesting it
 	// and sending it
-	// TODO: 2 use chrono type for time duration
-	void torrent::predicted_have_piece(piece_index_t const index, int const milliseconds)
+	void torrent::predicted_have_piece(piece_index_t const index, time_duration const duration)
```
arvidn, line 4169: I'm a bit torn here. On one hand, using the "generic" `time_duration` here is nice, because it hides internals. But on the other hand, it does mean the `milliseconds32` will be converted to nanoseconds (or maybe microseconds) in an `in64_t` just to be converted back again.

If this function would take a `milliseconds32` directly, you would still have the type-safety, you would just leak to the caller what format you're using internally in the function

arvidn:
I don't have a strong opinion about my comments. If you don't think anything should change, I'll merge this

src/torrent.cpp
vktr: I'm not too sure about the internals and I'm not at my workstation so I don't have the code in front of me - but if we make the `threshold` a time_duration, then we can use the regular `milliseconds` type I think? I chose `milliseconds32` to be able to use `eta.count()` without a static_cast to int.

If that works, we would keep the time_duration in `torrent::predicted_have_piece` but won't have to to the conversion from `milliseconds32`.

That said, I'm not well-versed in these APIs :)

vktr:
I left a review comment and also have another thought which spawned on me - the documentation for `settings_pack::predictive_piece_announce` mentions that setting it to 0 will disable it. Is that comparison done elsewhere or should I add an early return for that condition?

arvidn:
that check is here: https://github.com/arvidn/libtorrent/pull/4707/files#diff-ef3200244e6b5d4205ddce84844adb7eR2994

src/torrent.cpp
arvidn: > but if we make the threshold a time_duration, then we can use the regular milliseconds type I think? I chose milliseconds32 to be able to use eta.count() without a static_cast to int.

That sounds right. I thought you picked `milliseconds32` to maintain the `int` width in the existing code :)

vktr:
I updated the code to reflect what we discussed. Let me know if you need further changes, otherwise I think this is ready for merging.

arvidn:
speaking of low hanging fruit. `master` is using C++17 now, so there are probably a bunch of opportunities to make things a bit nicer with that. structured bindings instead of `std::tie()` for instance. maybe limiting variable scopes with statement-if.

I'm working on some cleanups to use `string_view` and to rely on stronger noexcept guarantees.

vktr:
I'll have a look. C++ is not really my main language, but I know enough to be dangerous so I'll see if I can find something to do :)

vktr:
One (of 13) Travis builds failed - something to look into or ignore?

arvidn:
yeah, those tests have been flaky lately, especially on mac. I really should look into fixing that

vktr:
I've got a MacBook so I can see if it's something I can reproduce.

Issue #4721, opened 2020-06-03
checking-files and resuming does not work or v2 torrents


arvidn:
libtorrent version (or branch): RC_2_0

when checking a v2-only torrent that's partially downloaded, the torrent ends up in a state where it won't download anything from any peer.

PR #4722, opened 2020-06-04
fix checking files of v2-only torrent

arvidn:
related to this issue: https://github.com/arvidn/libtorrent/issues/4721

The symptom if this problem was that pieces whose *piece hash* failed, would instead turn into `download_piece`es in the piece picker, where every block was finished. But the piece was not marked as "have".

With this change, the piece is identified as failing the hash check immediately and puts the piece bitfield in a correct state after a full check.

@ssiloti please verify/confirm that my understanding is correct, and this was the intended check here

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -2328,6 +2328,7 @@ bool is_downloading_state(int const st)
 		{
 			if (error.ec == boost::system::errc::no_such_file_or_directory
 				|| error.ec == boost::asio::error::eof
+				|| error.ec == lt::errors::file_too_short
 #ifdef TORRENT_WINDOWS
 				|| error.ec == error_code(ERROR_HANDLE_EOF, system_category())
 #endif
@@ -4026,6 +4027,7 @@ namespace {
 		int const blocks_in_piece = torrent_file().orig_files().blocks_in_piece2(piece);
 		int const blocks_per_piece = torrent_file().orig_files().piece_length() / default_block_size;
 
+		// the blocks are guaranteed to represent exactly one piece
 		TORRENT_ASSERT(blocks_in_piece == int(block_hashes.size()));
 
 		TORRENT_ALLOCA(block_passed, bool, blocks_in_piece);
@@ -4095,6 +4097,13 @@ namespace {
 			}
 			else if (result.status == set_block_hash_result::result::block_hash_failed)
 			{
+				ret = false;
+			}
+			else if (result.status == set_block_hash_result::result::piece_hash_failed && i == blocks_in_piece - 1)
+			{
+				// only if the *last* block causes the piece to fail, do we know
+				// it actually failed. Otherwise it might have been failing
+				// because of other, previously existing block hashes.
 				ret = false;
 			}
 		}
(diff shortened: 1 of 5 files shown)


arvidn:
I'm having a hard time triggering this issue in the `test_checking` unit test. So I'm less confident about this change. Though, it does seem like an improvement for my test torrent.

src/torrent.cpp
```diff
@@ -4070,7 +4070,8 @@ namespace {
 					we_have(verified_piece);
 				}
 			}
-			else if (result.status == set_block_hash_result::result::block_hash_failed)
+			else if (result.status == set_block_hash_result::result::block_hash_failed
+				|| result.status == set_block_hash_result::result::piece_hash_failed)
```
ssiloti, line 4074: This is in a loop which sets each block hash individually so there may be a spurious `piece_hash_failed` in the middle of the loop if any of the yet-to-be-set block hashes are non-zero (e.g. after re-downloading a previously failed piece). Only a piece hash failure on the last block can be considered a definite failure.

ssiloti:
The piece state you describe is the intended state for a piece which is fully downloaded and waiting for its piece hash to be downloaded so that it can be checked. I think you're on the right track making sure `on_blocks_hashed` detects a piece hash failure and returns false in that case.

arvidn:
I think the `block_hashes` always represent a full piece, right? So changing this condition to check if the *last* result is a `piece_hash_failed` might do it.

The current logic doesn't seem right where, as long as no block hash fails, the piece is considered passing. The case I'm hittin (but not in the unit test) is where we have the piece-layer from the .torrent file, and we check all the files on disk.

src/torrent.cpp
arvidn: makes sense. given the `block_hashes` are guaranteed to cover exactly one piece, is it ever possible for the loop, in the case `result.status == set_block_hash_result::result::success`, to take more than one turn? and is it ever possible for `verified_piece != piece`?

src/torrent.cpp
```diff
@@ -2328,6 +2328,7 @@ bool is_downloading_state(int const st)
 		{
 			if (error.ec == boost::system::errc::no_such_file_or_directory
 				|| error.ec == boost::asio::error::eof
+				|| error.ec == lt::errors::file_too_short
```
arvidn, line 2331: this was an independent issue uncovered by extending `test_checking`

src/torrent.cpp
```diff
@@ -4074,6 +4076,13 @@ namespace {
 			{
 				ret = false;
 			}
+			else if (result.status == set_block_hash_result::result::piece_hash_failed && i == blocks_in_piece - 1)
+			{
+				// only if the *last* block causes the piece to fail, do we know
+				// it actually failed. Otherwise it might have been failing
+				// because of other, previously existing block hashes.
+				ret = false;
```
arvidn, line 4084: without this, the extended test fails because even though the piece hash failed, the data would be marked as "finished" in the piece picker. The test now checks for that by ensuring that file progress is not 100% in the case where we don't have the full torrent.

I must admit I don't fully understand the case where *other* pieces than this one can be completed (in the loop above). I have a feeling that this check with `i` may need to be more sophisticated.

arvidn: after having spent some more time understanding this, I feel more confident that this is correct

PR #4727, opened 2020-06-05
additional asserts and comments in torrent::on_blocks_hashed

arvidn:
This is me trying to understand the constraints on this function. As far as I can tell, these additional asserts should hold.

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -4040,22 +4040,45 @@ namespace {
 				ret = false;
 				break;
 			}
-			auto result = get_hash_picker().set_block_hash(piece, i * default_block_size, block_hashes[i]);
+			auto const result = get_hash_picker().set_block_hash(piece
+				, i * default_block_size, block_hashes[i]);
 
-			block_passed[i] = result.status == set_block_hash_result::result::success;
 			if (result.status == set_block_hash_result::result::success)
 			{
-				TORRENT_ASSERT(result.first_verified_block <= blocks_in_piece);
+				TORRENT_ASSERT(result.first_verified_block < blocks_in_piece);
+				TORRENT_ASSERT(blocks_in_piece <= blocks_per_piece);
+
+				// all verified ranges should always be full pieces or less
+				TORRENT_ASSERT(result.first_verified_block >= 0
+					|| (result.first_verified_block % blocks_per_piece) == 0);
+				TORRENT_ASSERT(result.num_verified < blocks_in_piece
+					|| (result.num_verified % blocks_per_piece) == 0);
+
+				// sometimes, completing a single block may "unlock" validating
+				// multiple pieces. e.g. if we don't have the piece layer yet,
+				// but we completed the last block in the whole torrent, now we
+				// can validate everything. For this reason,
+				// first_verified_block may be negative.
+
+				// In this call, we track the blocks in this piece though, in
+				// the block_passed array. For that tracking we need to clamp
+				// the start index to 0.
 				auto const first_block = std::max(0, result.first_verified_block);
-				std::fill_n(block_passed.begin() + first_block
-					, std::min(blocks_in_piece - first_block, result.num_verified), true);
+				auto const count = std::min(blocks_in_piece - first_block, result.num_verified);
+				std::fill_n(block_passed.begin() + first_block, count, true);
 
-				using piece_delta = piece_index_t::diff_type;
+				// the current block (i) should be part of the range that was
+				// verified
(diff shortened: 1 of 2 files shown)


src/torrent.cpp
```diff
@@ -4040,22 +4040,45 @@ namespace {
...
 				break;
 			}
-			auto result = get_hash_picker().set_block_hash(piece, i * default_block_size, block_hashes[i]);
+			auto const result = get_hash_picker().set_block_hash(piece
+				, i * default_block_size, block_hashes[i]);
 
-			block_passed[i] = result.status == set_block_hash_result::result::success;
 			if (result.status == set_block_hash_result::result::success)
 			{
-				TORRENT_ASSERT(result.first_verified_block <= blocks_in_piece);
```
arvidn, line 4048: my understanding is that `first_verified_block` should never be `== blocks_in_piece`. That would only make sense if `num_verified == 0`. I strengthened this assert by saying it's `< blocks_in_piece`

src/torrent.cpp
```diff
@@ -4040,22 +4040,45 @@ namespace {
...
+				TORRENT_ASSERT(result.first_verified_block < blocks_in_piece);
+				TORRENT_ASSERT(blocks_in_piece <= blocks_per_piece);
+
+				// all verified ranges should always be full pieces or less
+				TORRENT_ASSERT(result.first_verified_block >= 0
+					|| (result.first_verified_block % blocks_per_piece) == 0);
+				TORRENT_ASSERT(result.num_verified < blocks_in_piece
+					|| (result.num_verified % blocks_per_piece) == 0);
```
arvidn, line 4055: it seems to hold true, that if more than one piece was validated, we only get whole pieces.

src/torrent.cpp
```diff
@@ -4040,22 +4040,45 @@ namespace {
...
+				// first_verified_block may be negative.
+
+				// In this call, we track the blocks in this piece though, in
+				// the block_passed array. For that tracking we need to clamp
+				// the start index to 0.
 				auto const first_block = std::max(0, result.first_verified_block);
-				std::fill_n(block_passed.begin() + first_block
-					, std::min(blocks_in_piece - first_block, result.num_verified), true);
+				auto const count = std::min(blocks_in_piece - first_block, result.num_verified);
+				std::fill_n(block_passed.begin() + first_block, count, true);
```
arvidn, line 4068: the only change here is factoring out of the `count`, just to be able to use it in the assert below

src/torrent.cpp
```diff
@@ -4040,22 +4040,45 @@ namespace {
 				ret = false;
 				break;
 			}
-			auto result = get_hash_picker().set_block_hash(piece, i * default_block_size, block_hashes[i]);
+			auto const result = get_hash_picker().set_block_hash(piece
+				, i * default_block_size, block_hashes[i]);
 
-			block_passed[i] = result.status == set_block_hash_result::result::success;
```
arvidn, line 4045: when we allocate `block_passed`, we clear it to all be `false`. Setting it to false here should not be necessary. What's left is setting it to `true`, which we can do unconditionally in the if-block below. But, in fact, we already do a `fill_n()` to set fields to `true` inside the if-block, and the range we set should always include `i`, which means we don't actually need this assignment anymore.

src/torrent.cpp
```diff
@@ -4040,22 +4040,45 @@ namespace {
...
 
-				// if the hashes for more than one piece have been verified, check for any pieces which
-				// were already checked but couldn't be verified and mark them as verified
-				for (piece_index_t verified_piece = piece + piece_delta(result.first_verified_block / blocks_per_piece)
-					, end = verified_piece + piece_delta(result.num_verified / blocks_per_piece)
+				// if the hashes for more than one piece have been verified,
+				// check for any pieces which were already checked but couldn't
+				// be verified and mark them as verified
+				for (piece_index_t verified_piece = piece + delta(result.first_verified_block / blocks_per_piece)
+					, end = verified_piece + delta(result.num_verified / blocks_per_piece)
```
arvidn, line 4081: it's asserted above that `first_verified_block` and `num_verified` should be divisible by `blocks_per_piece` (unless we verified less than a full piece, in which case this loop isn't relevant anyway)

arvidn:
@ssiloti does my understanding seem correct?

src/torrent.cpp
ssiloti: Right. IIRC the verified block range and corresponding fill_n came much later than this line. I clearly didn't notice that the former obsoleted the latter.

PR #4728, opened 2020-06-06
clean up parse_session_stats.py

arvidn:
and remove some old unused counters (num_blocks_cache_hits, torrent_evicted_counter)

### DIFF
--- a/include/libtorrent/performance_counters.hpp
+++ b/include/libtorrent/performance_counters.hpp
@@ -144,10 +144,6 @@ namespace libtorrent {
 			on_disk_queue_counter,
 			on_disk_counter,
 
-#if TORRENT_ABI_VERSION == 1
-			torrent_evicted_counter,
-#endif
-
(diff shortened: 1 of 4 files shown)


AdamWill:
`num_blocks_cache_hits` was actually being used by Deluge, it seems. I just noticed that while Deluge is running it is constantly logging an unhandled error caused by it expecting to get this from libtorrent, and not.

arvidn:
that counter doesn't exist anymore though (in libtorrent 2.0). It only contained 0 until this commit removed it, so an error message in Deluge is probably better than having behavior depend on it (which becomes implicitly disabled because the counter is disconnected).

I would guess that Deluge only used this counter to display it to the user though.

AdamWill:
yeah, it was just for display, I think. anyway, I looked a bit more and found there's a downstream issue/PR already - https://github.com/deluge-torrent/deluge/pull/312

PR #4734, opened 2020-06-07
Tune file checking

arvidn:
various fixes and tuning of disk I/O


arvidn:
issues uncovered and results from running these experiments: https://github.com/arvidn/libtorrent/wiki/disk-I-O-settings-for-checking-files

Feedback welcome!

FranciscoPombal:
@arvidn 

> issues uncovered and results from running these experiments: https://github.com/arvidn/libtorrent/wiki/disk-I-O-settings-for-checking-files

Summary:

1. The following make no difference in any case:
    - Checking memory usage > 256
    - `madvise(MADV_SEQUENTIAL)`
    - number of disk I/O threads, beyond the effective minimum tested of 3.
2. Perhaps unsurprisingly, NVMe (and probably SATA as well) SSD performance scales "indefinitely" w.r.t. the number of hashing threads, presumably up to their own bandwidth limit, their IOPS limit, or CPU bottleneck, whichever comes first (most likely the CPU bottleneck).
3. HDD performance falls of a cliff with > 2 hashing threads

It would be interesting to find _why_ HDD performance gets killed like that at > 2 hashing threads, and not, say, at > 3 or > 4. But other than that, IMO these results strongly strengthen the case for separating the configuration for `(number of hashing threads, number of disk I/O threads)`.
With the current configuration method, the disparity between the ideal value for each setup is far too great to settle on a "sane default", in a world which hasn't fully moved away from HDDs, and will not do so anytime soon.

For example, for each N:4, N:3, N:2 case, the ideal max value for HDD (meaning `8` for N:4, for example), leaves ~80% (!) relative potential performance on the table for SSDs (comparing to `aio_threads` set to `64`, there is probably even more headroom still). On the other hand, using more than 2 hashing threads to benefit SDDs leaves ~60% of HDDs relative potential performance on the table.

In fact, since many systems actually use both SSDs and HDDs, do you think there is any way to go a bit further and have libtorrent use some kind of hint from the system to know how many threads to use for each storage device, depending on their type? For instance, if I am downloading 10 torrents, 8 to my SSD and 2 to my HDD, I would want libtorrent to use at most 2 hashing threads for the stuff going to the HDD, and the rest for the SSD.

arvidn:
> It would be interesting to find why HDD performance gets killed like that at > 2 hashing threads, and not, say, at > 3 or > 4.

Yes. I tried to dig into this a bit with `perf`, but I couldn't get good stack traces not any samples from kernel page fault handlers (where I suspect it all happens). I would probably have to spend some more time to get good data out of perf.

I suspect too many threads causes the read pattern to not be perfectly sequential (but just *mostly* sequential) which makes the kernel to disable read-ahead (despite using `MADV_SEQUENTIAL`). This is partly why I left the `MADV_SEQUENTIAL` patch in, since it didn't seem to make things worse, but there's a chance it might make things better in the future. Or if there's some kernel tunable to make `MADV_SEQUENTIAL` a "stronger" hint.

>  But other than that, IMO these results strongly strengthen the case for separating the configuration for (number of hashing threads, number of disk I/O threads).

I agree. It's still a bit of a cop-out, since most people probably won't touch such setting, so performance will be sub-optimal in the majority of cases. However, it's not feasible to solve this before the 2.0 release anyway. (by "solve", I mean make it perform optimally for both SSD and HDD cases).

FranciscoPombal:
@arvidn 

> I suspect too many threads causes the read pattern to not be perfectly sequential (but just mostly sequential) which makes the kernel to disable read-ahead

Yes. Though one has to wonder: does the read pattern _only_ stops being "perfectly sequential" from 2 -> 3 threads, or is it the case that it is already not "perfectly sequential" with 2 threads, but it's too small of an impact to be noticed? It could be the case that the impact is absorbed by the drive's small onboard cache.

> I agree. It's still a bit of a cop-out, since most people probably won't touch such setting, so performance will be sub-optimal in the majority of cases. However, it's not feasible to solve this before the 2.0 release anyway. (by "solve", I mean make it perform optimally for both SSD and HDD cases).

Yes, fully solving the issue (by your definition of "solving it", which I think is fine) will take time, and decoupling the tunables is only a part of the full fix. But I think it is a good starting point, from the perspective of a client application it is rather non-intuitive to have such an important parameter as the number of hasher threads configured indirectly and dependent on another tunable.

> The simplest (and most realistic) hint I can think of would be to say, serialize all hash jobs for this torrent. If the hashin is disk-bound, read-ahead should take care of hiding all the CPU cost of hashing anyway, and a single thread might be good enough.

> I don't think this is in scope for 2.0, but maybe 2.1. Can you think of a good mechanism to detect whether this should be enabled or not for a torrent?

To be honest, this is not my comfort zone (I had to go look up `madvise` earlier). But on Linux, for example, is it not possible to query `/sys/block/sdX/queue/rotational` to distinguish the type of storage medium? The hasher threads could be distributed accordingly. Assuming of course it is possible to establish the correspondence between each write job and the device they are writing to.

arvidn:
> Yes. Though one has to wonder: does the read pattern only stops being "perfectly sequential" from 2 -> 3 threads, or is it the case that it is already not "perfectly sequential" with 2 threads, but it's too small of an impact to be noticed? It could be the case that the impact is absorbed by the drive's small onboard cache.

my suspicion is that the read-ahead is triggered by "perfect" sequential reads, because it's simple and predictable. and presumably the common case is single-threaded reads, which would get you perfect back-to-back reads. It's probably a coincidence that me using two threads becomes perfect back-to-back requests the vast majority of times.

In my experiment there definitely seems to be a "cliff" or a step from 2 -> 3 hasher threads, which I imagine is because of read-ahead getting disabled. The completion time basically doubles.

arvidn:
@FranciscoPombal I replaced the commit that changed the hasher-thread ratio with one that separates the settings. does this look reasonable?

PR #4736, opened 2020-06-09
make create_torrent fail rather than produce invalid torrents

arvidn:
By default, create a torrent corresponding to which hashes were set

### DIFF
--- a/src/create_torrent.cpp
+++ b/src/create_torrent.cpp
@@ -409,8 +409,6 @@ namespace {
 		, m_v2_only(bool(flags & create_torrent::v2_only))
 		, m_v1_only(bool(flags & create_torrent::v1_only))
 	{
-		TORRENT_ASSERT_PRECOND(!(m_v2_only && m_v1_only));
-
 		// return instead of crash in release mode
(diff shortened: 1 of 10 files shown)


src/create_torrent.cpp
```diff
@@ -518,38 +507,94 @@ namespace {
...
+		// torrent
+		bool const make_v2 = validate_v2_hashes(m_files, m_file_piece_hash);
+		bool const make_v1 = validate_v1_hashes(m_files, m_piece_hash);
+
+		// if neither v1 nor v2 hashes were set, we can't create a torrent
+		if (!make_v1 && !make_v2)
+			aux::throw_ex<system_error>(errors::invalid_hash_entry);
+
+		if ((make_v1 && m_v2_only) || (make_v2 && m_v1_only))
+			aux::throw_ex<system_error>(errors::invalid_hash_entry);
```
arvidn, line 592: maybe this check is excessive. or maybe it would make more sense to have `set_hash()` and `set_hash2()` fail directly instead of deferring the failure until here.

ssiloti: I think it makes more sense to do this check in `set_hash()` and `set_hash2()`.

PR #4738, opened 2020-06-09
validate file index in reject_hash_request

arvidn:
if `file_index` is invalid, we reject the hash request, and we better make sure not to try to look up that file when we do.


arvidn:
@ssiloti perhaps the best would be to echo back the invalid file root hash, but the current `hash_request` structure holds an index, so it would complicate things a bit.

arvidn:
credit goes to OSS-Fuzz for discovering this btw!

ssiloti:
I think we should return the invalid file root, otherwise the peer can't correlate the reject with a request.

arvidn:
does this look good? it was simpler than I thought and it even saves a lookup

PR #4744, opened 2020-06-11
Remove Boost.system dependency and up minimum boost to 1.69

gtumanyan:
Boost.system is header-only since boost 1.69


arvidn:
it looks like `configure.ac` and cmake still looks for the library, causing these failures:
https://travis-ci.org/github/arvidn/libtorrent/jobs/697096303#L912
https://travis-ci.org/github/arvidn/libtorrent/jobs/697096304#L781

arvidn:
for some reason, travis isn't showing up in PRs. So, here it is: https://travis-ci.org/github/arvidn/libtorrent/builds/698567614

arvidn:
do you know which the oldest version of ubuntu is that has a boost-1.69 or later `apt` package? boost-1.71 is available in 20.04 at least

arvidn:
https://travis-ci.org/github/arvidn/libtorrent/jobs/698567625#L1239
https://travis-ci.org/github/arvidn/libtorrent/jobs/698567626#L777

petersasi:
> do you know which the oldest version of ubuntu is that has a boost-1.69 or later `apt` package? boost-1.71 is available in 20.04 at least
https://packages.ubuntu.com/search?searchon=sourcenames&keywords=boost
Xenial (16.04) has 1.58.0.1
Bionic (18.04) has 1.65.1.0
Eoan (19.10) has 1.67.0.2
Focal and Groovy (20.04 and 20.10) has 1.71.0.0

gtumanyan:
> do you know which the oldest version of ubuntu is that has a boost-1.69 or later `apt` package? boost-1.71 is available in 20.04 at least

Just installed libboost1.70-dev from [ppa:mhier/libboost-latest](https://launchpad.net/~mhier/+archive/ubuntu/libboost-latest) with 
sudo add-apt-repository ppa:mhier/libboost-latest
sudo apt-get install libboost1.70-dev
on Ubuntu 18.04
According to this repository libboost 1.70 is available in ubuntu14.04.1 and up

arvidn:
as far as I understand, that doesn't really help the maintainer of the libtorrent-rasterbar package in ubuntu. Although, I see that is a *very* old version of libtorrent anyway.

arvidn:
@gtumanyan Thinking about this some more, I'm a bit hesitant to put this in the current stable branch. Would you mind rebasing this on top of `RC_2_0`? The boost version requirement is already bumped for that version.

arvidn:
there are still failures on travis (for some reason travis has stopped showing up in PRs on github).

https://travis-ci.org/github/arvidn/libtorrent/builds/698795137

Specifically:

https://travis-ci.org/github/arvidn/libtorrent/jobs/698795146#L1367
https://travis-ci.org/github/arvidn/libtorrent/jobs/698795147#L777

gtumanyan:
> there are still failures on travis (for some reason travis has stopped showing up in PRs on github).
Fixed.
By the way [Martin Hierholzer](https://launchpad.net/~mhier) updated his [ppa](https://launchpad.net/~mhier/+archive/ubuntu/libboost-latest) with latest libboost	1.73-0~12 which installs in Ubuntu down to 14.04.1

arvidn:
This is looking like an increasingly risky patch. certainly too risky for the `RC_1_2` branch, maybe even the `RC_2_0` branch.
It doesn't seem the benefit of this outweighs the risk (at least not in a stable branch). Feel free to "sell" the benefits if you think they do.

As far as I understand, this patch will change not just how the CI is built, but also how *users* will link against boost, which means it's not obvious that someone successfully linking against libtorrent-1.2.7 would also successfully link against a future libtorrent-1.2.8

CMakeLists.txt
```diff
@@ -678,11 +678,6 @@ else()
 	target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_DISABLE_DHT)
 endif()
 
-# Boost
-find_public_dependency(Boost REQUIRED COMPONENTS system)
-target_include_directories(torrent-rasterbar PUBLIC ${Boost_INCLUDE_DIRS})
-target_link_libraries(torrent-rasterbar PUBLIC ${Boost_SYSTEM_LIBRARY})
```
arvidn, line 684: if someone is on a system that has boost.system as a compiled library, removing this will cause link errors, right? Is it important to remove this?
A system where boost system is used as header-only, presumably cmake would still create a target for it, just not link against a library.

gtumanyan: Thank you. I didn't got that (Boost REQUIRED COMPONENTS system) is a boost detector.

FranciscoPombal:
I don't think such a massive minimum version bump and change in the way boost is linked should target the mature `RC_1_2` branch. Unless of course there is a compelling technical reason, like a bug fix that really affects libtorrent. Of course it's always better to use the newest versions of everything, but one has to keep in mind the expectations set when a branch/series is first released. Since `RC_1_2` was released not released with the policy of "this will track and require recent versions of boost, and it's possible that the min requirements change between point releases", I don't think the precedent should be broken for no reason in the middle of a series.

I would say it this would be ok for `RC_2_0`. In fact, since it is technically not released yet, it could even be decided to change the boost version policy to be "the most recent version if possible, even between point releases", as long as such a statement were clearly present in the documentation, to correctly manage expectations of users of the library.

But back to the real world - I would expect `RC_2_0` to be released with a certain minimum required boost version, and to stick with it for the remainder of the 2.0.x series - that's just how it has been so far for this project. But, if you're going to choose a minimum boost version for 2.0.x, 1.69 seems like an odd choice:

- If you want to build OOTB with Debian Buster (released 2019), you would be limited to 1.67.
- If you don't care about Buster, and just want to guarantee OOTB build with Ubuntu 20.04/Debian Bullseye you could go all the way up to 1.71 (Ubuntu 20.04 being the limiting factor here, since Bullseye is not yet released and could still see a version bump).

I only mention these 2 because they are the most relevant "slow" distributions, and thus usually the limiting factor when deciding minimum supported dependency versions. Not sure how important it is to guarantee OOTB builds with "glacially slow" distors (Red Hat et. al.) - you would have to go with an even lower version. But then again, it is common for projects to not wait around for these ones. Everything else (rolling release distros, macOS and Windows) is largely irrelevant for these considerations because of the different way they handle package versions, "stability" and long term support.

@arvidn thoughts?

arvidn:
Yes, I've [suggested](https://github.com/arvidn/libtorrent/pull/4744#issuecomment-646539823) that this should not go into `RC_1_2` and probably not `RC_2_0` either.

PR #4747, opened 2020-06-11
Update ge.cpp

CaeaMadalina:
array aslide[256] was accessed at index -1


arvidn:
do you understand why address sanitizer did not catch this?
Your change seems like it easily could break the semantics of that function, are you confident it's correct? (and if so, how?)

arvidn:
the fact that the unit tests are failing suggests that this is not a correct solution

arvidn:
It's not obvious how the existing code could cause `aslide` to be accessed at index -1.

PR #4749, opened 2020-06-11
Update assert.cpp

CaeaMadalina:
Returning pointer to local variable 'demangled_name' that will be inavlid when returning


arvidn:
Can you explain the rationale behind this change?
The function returns a `std::string`, you're not changing the return type.
At first I thought the issue was the prevent a warning about array to pointer decay, but you still rely on that in your patch.

PR #4123, opened 2019-11-25
WebTorrent support

paullouisageneau:
This PR suggests an implementation of the WebTorrent protocol inside libtorrent. With this add-on, libtorrent can be used to run hybrid peers seamlessly.

WebTorrent peers use the same application protocol as classic peers but with WebRTC Data Channels as underlying transport instead of uTP or TCP, allowing to run peers in web browsers. Due to the necessary connection signaling, WebTorrent peers require a specific WebSocket tracker to connect. 

Since documentation seems to be lacking, I implemented the WebTorrent signaling protocol the best I could guess by looking at existing code. You can find useful information and the reference Javascript implementation here: https://github.com/webtorrent/webtorrent

The implementation requires two external dependencies added as submodules:
- [libdatachannel](https://github.com/paullouisageneau/libdatachannel), my lightweight WebRTC Data Channels implementation, which relies on my ad-hoc UDP Interactive Connectivity Establishment (ICE) library [libjuice](https://github.com/paullouisageneau/libjuice)
- [boost.JSON](https://github.com/vinniefalco/json), a header-only JSON library for the tracker protocol

**This PR does not add new external dependencies to libtorrent (it only requires OpenSSL or GnuTLS depending on the compilation option).**

To test, you need a torrent file with at least one WebSocket tracker, for instance `wss://tracker.openwebtorrent.com`.
You can find such a torrent file here: https://ageneau.org/upload/sintel.torrent
You can also run a web browser peer on this page: https://webtorrent.io/

PR #4750, opened 2020-06-11
[RC_2_0]clang-tidy fixes

neheb:


arvidn:
These patches introduce quite a lot of build and test failures. Take a look here: https://travis-ci.org/github/arvidn/libtorrent/builds/697374505

(I'm not sure why travis doesn't show up in this PR though).

I think it would be a better approach to have these commits as separate pull requests, I imagine some are more controversial than others.

neheb:
:).

Probably just the reference ones.

arvidn:
I'm a bit hesitant to put these in `RC_2_0` as well. would you mind rebasing on top of `master`? they are pretty similar, I would expect it to apply cleanly

src/upnp.cpp
```diff
@@ -691,13 +691,13 @@ void upnp::try_map_upnp()
 	TORRENT_ASSERT(is_single_thread());
 	if (m_devices.empty()) return;
 
-	for (auto i = m_devices.begin(), end(m_devices.end()); i != end; ++i)
+	for (const auto& m_device : m_devices)
```
arvidn, line 695: sorry for a nit-pick, but I use "east const" int this code base. i.e. `auto const&`

src/ut_pex.cpp
```diff
@@ -237,9 +237,9 @@ namespace libtorrent { namespace {
 			, m_first_time(true)
 		{
 			const int num_pex_timers = sizeof(m_last_pex) / sizeof(m_last_pex[0]);
```
arvidn, line 239: `num_pex_timers` is unused now and should probably be removed in this commit

src/disk_job_pool.cpp
```diff
@@ -44,11 +44,15 @@ namespace aux {
...
 
+#if 0
 	disk_job_pool::~disk_job_pool()
 	{
 // #error this should be fixed!
 //		TORRENT_ASSERT(m_jobs_in_use == 0);
 	}
+#else
+	disk_job_pool::~disk_job_pool() = default;
+#endif
```
arvidn, line 55: I don't think this is a reasonable change

arvidn:
I don't think the commit `[clang-tidy] pass by value` is good. the `address` type is essentially an `std::array<char, 16>`. There's no benefit in moving it, taking it by `const&` makes sense. I'm not sure what clang-tidy's heuristic is, perhaps 16 bytes is small enough where it thinks it's better to copy.

src/disk_job_pool.cpp
neheb: Interestingly, after this change clang-tidy complains that this is trivially destructible. The fix there being to move = default to the header.

neheb:
I'll post each commit separately and rebase against master. I only based my changes on RC_2_0 as that seems to be the most active branch.

src/upnp.cpp
```diff
@@ -277,7 +277,7 @@ void upnp::discover_device_impl()
 
 // returns a reference to a mapping or -1 on failure
 port_mapping_t upnp::add_mapping(portmap_protocol const p, int const external_port
-	, tcp::endpoint const local_ep)
+	, tcp::endpoint const& local_ep)
```
arvidn, line 280: this change makes sense.
Every other change in this commit (after a cursory scan) look wrong. I'm pretty sure I *want* to pass any type that can be efficiently moved, and that the function want a copy of, by value. this would include smart pointers and vectors for instance.

src/bt_peer_connection.cpp
```diff
@@ -182,13 +182,13 @@ namespace {
 #if !defined TORRENT_DISABLE_ENCRYPTION
 	void bt_peer_connection::switch_send_crypto(const std::shared_ptr<crypto_plugin>& crypto)
 	{
-		if (m_enc_handler.switch_send_crypto(std::move(crypto), send_buffer_size() - get_send_barrier()))
```
arvidn, line 185: I think a better fix would be to take the `crypto` argument by value

src/bt_peer_connection.cpp
```diff
@@ -182,13 +182,13 @@ namespace {
...
+		if (m_enc_handler.switch_send_crypto(crypto, send_buffer_size() - get_send_barrier()))
 			set_send_barrier(send_buffer_size());
 	}
 
 	void bt_peer_connection::switch_recv_crypto(const std::shared_ptr<crypto_plugin>& crypto)
 	{
-		m_enc_handler.switch_recv_crypto(std::move(crypto), m_recv_buffer);
```
arvidn, line 191: same here, `crypto` should probably be taken by value

src/create_torrent.cpp
```diff
@@ -372,7 +372,7 @@ namespace {
 			info_hash
 		};
 
-		storage_holder storage = disk_thread->new_torrent(std::move(params)
```
arvidn, line 373: this makes sense. and in fact, `new_torrent()` should probably take this by `const&` as well

src/disabled_disk_io.cpp
```diff
@@ -119,7 +119,7 @@ struct TORRENT_EXTRA_EXPORT disabled_disk_io final
 		, std::function<void(status_t, std::string const&, storage_error const&)> handler) override
 	{
 		post(m_ios, [h = std::move(handler), path = std::move(p)] () mutable
-			{ h(status_t::no_error, std::move(path), storage_error{}); });
```
arvidn, line 122: I don't see how this is useless. the lambda is `mutable`, `path` should be too.

src/disabled_disk_io.cpp
```diff
@@ -146,7 +146,7 @@ struct TORRENT_EXTRA_EXPORT disabled_disk_io final
 		, std::function<void(std::string const&, file_index_t, storage_error const&)> handler) override
 	{
 		post(m_ios, [h = std::move(handler), index, n = std::move(name)]
-			{ h(std::move(n), index, storage_error{}); });
```
arvidn, line 149: I think a better fix would be to make the lambda mutable

src/file_view_pool.cpp
```diff
@@ -172,7 +172,7 @@ namespace libtorrent { namespace aux {
 		auto& lru_view = m_files.get<1>();
 		if (lru_view.size() == 0) return {};
 
-		auto mapping = std::move(lru_view.back().mapping);
```
arvidn, line 175: It would be better to fix this to *actually* move the mapping. I imagine `lru_view.back()` returns an immutable reference.

src/disabled_disk_io.cpp
neheb: This was all automated. No idea what clang-tidy was thinking.

src/file_view_pool.cpp
```diff
@@ -221,7 +221,7 @@ namespace libtorrent { namespace aux {
 		auto const end = key_view.upper_bound(file_id{st, std::numeric_limits<file_index_t>::max()});
 
 		for (auto it = begin; it != end; ++it)
-			defer_destruction.emplace_back(std::move(it->mapping));
```
arvidn, line 224: I would prefer to figure out how to get to a mutable mapping, in order to actually make these moves

src/peer_connection_handle.cpp
```diff
@@ -53,7 +53,7 @@ void peer_connection_handle::add_extension(const std::shared_ptr<peer_plugin>& e
 #ifndef TORRENT_DISABLE_EXTENSIONS
 	std::shared_ptr<peer_connection> pc = native_handle();
 	TORRENT_ASSERT(pc);
-	pc->add_extension(std::move(ext));
```
arvidn, line 56: I think a better fix would be to take `ext` by value

src/peer_connection_handle.cpp
```diff
@@ -331,7 +331,7 @@ void bt_peer_connection_handle::switch_send_crypto(const std::shared_ptr<crypto_
 #if !defined TORRENT_DISABLE_ENCRYPTION
 	std::shared_ptr<bt_peer_connection> pc = native_handle();
 	TORRENT_ASSERT(pc);
-	pc->switch_send_crypto(std::move(crypto));
```
arvidn, line 334: I think a better fix would be to take `crypto` by value

src/peer_connection_handle.cpp
```diff
@@ -342,7 +342,7 @@ void bt_peer_connection_handle::switch_recv_crypto(const std::shared_ptr<crypto_
 #if !defined TORRENT_DISABLE_ENCRYPTION
 	std::shared_ptr<bt_peer_connection> pc = native_handle();
 	TORRENT_ASSERT(pc);
-	pc->switch_recv_crypto(std::move(crypto));
```
arvidn, line 345: ditto

src/posix_disk_io.cpp
```diff
@@ -347,7 +347,7 @@ namespace {
 			storage_error error;
 			st->rename_file(idx, name, error);
 			post(m_ios, [idx, error, h = std::move(handler), n = std::move(name)] () mutable
-				{ h(std::move(n), idx, error); });
```
arvidn, line 350: I don't see why this wouldn't be a move. n should be mutable here

src/session_impl.cpp
```diff
@@ -878,7 +878,7 @@ bool ssl_server_name_callback(ssl::stream_handle_type stream_handle, std::string
 			{
 				auto state = ext->save_state();
 				for (auto& v : state)
-					ret.ext_state[std::move(v.first)] = std::move(v.second);
+					ret.ext_state[v.first] = std::move(v.second);
```
arvidn, line 881: `v.first` is a mutable `std::string`, no? maybe this should be changed to an `insert()` call, where the key can be moved in

src/session_impl.cpp
```diff
@@ -2542,7 +2542,7 @@ namespace {
 
 		ADD_OUTSTANDING_ASYNC("session_impl::on_udp_packet");
 		s->sock.async_read(make_handler([this, socket, ls, ssl](error_code const& e)
-			{ this->on_udp_packet(std::move(socket), std::move(ls), ssl, e); }
+			{ this->on_udp_packet(socket, ls, ssl, e); }
```
arvidn, line 2545: I think a better fix would be to make the lambda mutable

src/session_impl.cpp
```diff
@@ -4705,7 +4705,7 @@ namespace {
 		{
 			std::shared_ptr<torrent_plugin> tp(e->new_torrent(
 				torrent_ptr->get_handle(), userdata));
-			if (tp) torrent_ptr->add_extension(std::move(tp));
```
arvidn, line 4706: I think moving is correct here. even if `add_extension` doesn't take the shared pointer by value, the call site should move if it can. I also think that `add_extension()` probably *should* take it by value

src/session_impl.cpp
```diff
@@ -5912,7 +5912,7 @@ namespace {
 	{
 		if (!m_dht) return;
 		m_dht->get_item(dht::public_key(key.data()), std::bind(&session_impl::get_mutable_callback
-			, this, _1, _2), std::move(salt));
```
arvidn, line 5913: I think a better fix would be to take `salt` by value

src/session_impl.cpp
```diff
@@ -5980,7 +5980,7 @@ namespace {
 		if (!m_dht) return;
 		m_dht->put_item(dht::public_key(key.data())
 			, std::bind(&on_dht_put_mutable_item, std::ref(m_alerts), _1, _2)
-			, std::bind(&put_mutable_callback, _1, std::move(cb)), std::move(salt));
```
arvidn: ditto

src/torrent_info.cpp
```diff
@@ -1117,7 +1117,7 @@ namespace {
 	{
 		INVARIANT_CHECK;
 
-		torrent_info tmp = std::move(ti);
```
arvidn, line 1120: this absolutely need to be a move. how is this redundant?

src/udp_socket.cpp
```diff
@@ -86,7 +86,7 @@ struct socks5 : std::enable_shared_from_this<socks5>
 		, m_timer(ios)
 		, m_retry_timer(ios)
 		, m_alerts(alerts)
-		, m_listen_socket(std::move(ls))
+		, m_listen_socket(ls)
```
arvidn, line 89: I think a better fix is to take `ls` by value

src/udp_socket.cpp
```diff
@@ -179,7 +179,7 @@ udp_socket::udp_socket(io_context& ios, const aux::listen_socket_handle& ls)
 	: m_socket(ios)
 	, m_ioc(ios)
 	, m_buf(new receive_buffer())
-	, m_listen_socket(std::move(ls))
+	, m_listen_socket(ls)
```
arvidn, line 182: `ls` should probably be passed in by value

src/path.cpp
```diff
@@ -976,8 +976,7 @@ namespace {
 			return true;
 		return false;
 #else
-		if (f[0] == '/') return true;
-		return false;
+		return f[0] == '/';
```
arvidn, line 979: this is a good change.

src/peer_connection.cpp
```diff
@@ -3614,7 +3614,7 @@ namespace libtorrent {
 		}
 
 		pending_block pb(block);
-		pb.busy = (flags & busy) ? true : false;
+		pb.busy = static_cast<bool>((flags & busy));
```
arvidn, line 3617: this is a good change

src/peer_connection.cpp
```diff
@@ -4476,8 +4476,7 @@ namespace libtorrent {
 
 		if (m_ses.ignore_unchoke_slots_set(*this)) return true;
 		std::shared_ptr<torrent> t = m_torrent.lock();
-		if (t && m_ses.ignore_unchoke_slots_set(*t)) return true;
-		return false;
+		return t && m_ses.ignore_unchoke_slots_set(*t);
```
arvidn, line 4479: +1

src/peer_list.cpp
```diff
@@ -440,15 +440,12 @@ namespace libtorrent {
...
+		return !(p.connection
 			|| p.banned
 			|| p.web_seed
 			|| !p.connectable
 			|| (p.seed && m_finished)
-			|| int(p.failcount) >= m_max_failcount)
-			return false;
-
-		return true;
+			|| int(p.failcount) >= m_max_failcount);
```
arvidn, line 448: +1

src/web_peer_connection.cpp
```diff
@@ -699,7 +699,7 @@ void web_peer_connection::handle_redirect(int const bytes_left)
 		// the new web seed we're adding only has this file for now
 		// we may add more files later
 		web->redirects[file_index] = redirect_path;
-		if (web->have_files.get_bit(file_index) == false)
+		if (!web->have_files.get_bit(file_index))
```
arvidn, line 702: I'm not convinced about everything this checker suggests. `a == false` can be a lot clearer and easier to read than `!a`, especially when `a` is a long expression

arvidn:
I will pull some of these changes that are borderline bugs into `RC_2_0`

arvidn:
I would really appreciate it if you would add the corresponding checkers to the clang-tidy command line in `,travis.yml` as well, so they are checked going forward.

neheb:
I think adding a clang-format file and a travis linter comes first. Especially since you mentioned the east const thing.

arvidn:
https://github.com/arvidn/libtorrent/pull/4758

arvidn:
there's a fairly large patch waiting to land in `master` ( https://github.com/arvidn/libtorrent/pull/4123 ) and I'm a bit hesitant to fix clang-tidy issues ahead if that, having to rebase that patch.

You'll find the current clang-tidy command line here: https://github.com/arvidn/libtorrent/blob/RC_1_2/.travis.yml#L152

It's currently just doing:
```
-checks=-clang-analyzer-core.*
```

regarding a clang-format file; I think it would be great to have one. I'm a bit worried about the transition to using one. My impression is that it doesn't support all the formatting conventions used in the code base right now, which means there would have to be a *big* patch changing virtually everything, which isn't great git-history-wise. It would also have to be carefully timed to happen when no large patches are outstanding.

neheb:
Not really. git clang-format only applies it to changes, not to the whole codebase.

src/web_peer_connection.cpp
FranciscoPombal: On the other hand, using `a == false` can lead to accidentally writing `a = false` sometimes, leading  to possibly very hard to track bugs, Unless one commits to using the `false == a` style.

PR #4752, opened 2020-06-11
[clang-tidy] use emplace_back

neheb:
Found with modernize-use-emplace

### DIFF
--- a/src/merkle_tree.cpp
+++ b/src/merkle_tree.cpp
@@ -220,7 +220,7 @@ namespace aux {
 			}
 			else
 			{
-				passed_pieces.push_back(piece_index_t{file_piece_offset + piece});
+				passed_pieces.emplace_back(file_piece_offset + piece);
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


arvidn:
```
'argument': conversion from '_Ty' to 'uint8_t', possible loss of data
        with
        [
            _Ty=int
        ]
    call "..\bin\standalone\msvc\17b09cf932770549210a83141190f492\msvc-setup.bat"  >nul
 cl /Zm800 -nologo @"..\bin\de4cc3570b8122ae43418614795d7a78\src\bdecode.obj.rsp" 
```

instantiated from here: `..\src\bdecode.cpp(882): note: see reference to function template instantiation`

arvidn:
https://ci.appveyor.com/project/arvidn/libtorrent/builds/33496100/job/megth2eqgovpinr3#L265

neheb:
No real idea how to fix.

src/bdecode.cpp
```diff
@@ -878,8 +878,8 @@ namespace aux {
 						start = int_start;
 						TORRENT_FAIL_BDECODE(e);
 					}
-					ret.m_tokens.push_back({int_start - orig_start
-						, 1, bdecode_token::integer, 1});
+					ret.m_tokens.push_back(int_start - orig_start
```
arvidn, line 881: I think you meant to say `emplace_back()` here.

arvidn:
if you were to add `modernize-use-emplace` to the `.clang-tyidy` file, would it find more places. or would it pass CI?

neheb:
It would. I intentionally left out one file since it doesn't compile with MSVC for some reason.

PR #4753, opened 2020-06-11
[clang-tidy] use raw strings

neheb:
Found with modernize-raw-string-literal

### DIFF
--- a/src/peer_list.cpp
+++ b/src/peer_list.cpp
@@ -597,7 +597,7 @@ namespace libtorrent {
 #ifndef TORRENT_DISABLE_LOGGING
 			if (i->connection != nullptr && c.should_log(peer_log_alert::info))
 			{
-				c.peer_log(peer_log_alert::info, "DUPLICATE PEER", "this: \"%s\" that: \"%s\""
+				c.peer_log(peer_log_alert::info, "DUPLICATE PEER", R"(this: "%s" that: "%s")"
 					, print_address(c.remote().address()).c_str()
 					, print_address(i->address()).c_str());
 			}
--- a/src/upnp.cpp
+++ b/src/upnp.cpp
@@ -254,7 +254,7 @@ void upnp::discover_device_impl()
 #ifndef TORRENT_DISABLE_LOGGING
 		if (should_log())
 		{
-			log("multicast send failed: \"%s\" and \"%s\". Aborting."
+			log(R"(multicast send failed: "%s" and "%s". Aborting.)"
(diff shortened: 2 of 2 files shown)


arvidn:
It's not completely obvious to me that this is an improvement. only one of the changes reduces the number of characters for instance, and only by 2 characters.

FranciscoPombal:
@arvidn Even though there might not be a reduction in the number of characters, IMO the benefit of using raw strings is to not have to worry about escaping anything or missing escapes.

arvidn:
> Even though there might not be a reduction in the number of characters, IMO the benefit of using raw strings is to not have to worry about escaping anything or missing escapes.

but with that logic *all* strings ought to be raw string literals, in case something need to be escaped, or if someone forgets to escape something that was meant to be. (I think it really only applies to `"`, since you don't really forget to escape `\n`).

I think the main benefit of raw string literals are when there is so much escaping that the "meaning" is obfuscated.

I'm not opposed to this change, but I also don't feel like it's particularly well justified. Should all string literals with a `"` it them be a raw string literal? Is it just to satisfy clang-tidy?

FranciscoPombal:
Good points. I suppose it's a matter of style/policy in the end then. You could decide "every string in the code base should be raw, for maximum consistency and `clag-tidy` statisfiability" or "every string in the code base should not be raw, for maximum consistency", or even "I'll know good candidates for raw strings when I see them, otherwise they should not be used". Up to you. Full disclosure, I'm more partial to the first option - but is there any particular disadvantage (in terms of performance or something) to using them?

arvidn:
I'm with Tony :) https://youtu.be/GPP64opjy_Y?t=1169

neheb:
There's no performance difference with raw strings. Not even compile time.

Let's look at a hilarious example: https://github.com/jbeder/yaml-cpp/pull/876/commits/cfe99dc1078cd772c6fc718d10690cfead0c9a73#diff-46e7b027d02f780467a7b7ed7b0a7cedR384

The original is unreadable to me.

arvidn:
I think it's obvious that there's no performance difference (unless we're measuring performance in legibility).
I'm suggesting that sometimes escaping a quote is more readable than using a raw string literal, and sometimes it isn't

neheb:
To each his own I guess...

arvidn:
in that commit, changing `"\x20"` -> `R(" ")` seems a bit odd and inconsistent. Why not change it to `" "`?

neheb:
Different project. If I were to speculate, \x20 is to prevent compilers from being clever.

PR #4754, opened 2020-06-11
[clang-tidy] use auto

neheb:
Found with modernize-use-auto

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -461,7 +461,7 @@ void apply_deprecated_dht_settings(settings_pack& sett, bdecode_node const& s)
 int ssl_server_name_callback(SSL* s, int*, void* arg)
 {
 	char const* name = SSL_get_servername(s, TLSEXT_NAMETYPE_host_name);
-	session_impl* si = reinterpret_cast<session_impl*>(arg);
+	auto* si = reinterpret_cast<session_impl*>(arg);
 	return ssl_server_name_callback_impl(s, name ? std::string(name) : "", si)
 			? SSL_TLSEXT_ERR_OK
 			: SSL_TLSEXT_ERR_ALERT_FATAL;
@@ -469,7 +469,7 @@ int ssl_server_name_callback(SSL* s, int*, void* arg)
 #elif defined TORRENT_USE_GNUTLS
 bool ssl_server_name_callback(ssl::stream_handle_type stream_handle, std::string const& name, void* arg)
 {
-	session_impl* si = reinterpret_cast<session_impl*>(arg);
+	auto* si = reinterpret_cast<session_impl*>(arg);
 	return ssl_server_name_callback_impl(stream_handle, name, si);
 }
 #endif
@@ -2262,7 +2262,7 @@ namespace {
 				, nullptr, nullptr, true, false));
 
 		ADD_OUTSTANDING_ASYNC("session_impl::on_i2p_accept");
-		i2p_stream& s = std::get<i2p_stream>(*m_i2p_listen_socket);
+		auto& s = std::get<i2p_stream>(*m_i2p_listen_socket);
 		s.set_command(i2p_stream::cmd_accept);
 		s.set_session_id(m_i2p_conn.session_id());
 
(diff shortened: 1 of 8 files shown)


src/session_impl.cpp
```diff
@@ -461,15 +461,15 @@ void apply_deprecated_dht_settings(settings_pack& sett, bdecode_node const& s)
...
+	auto si = reinterpret_cast<session_impl*>(arg);
 	return ssl_server_name_callback_impl(s, name ? std::string(name) : "", si)
 			? SSL_TLSEXT_ERR_OK
 			: SSL_TLSEXT_ERR_ALERT_FATAL;
 }
 #elif defined TORRENT_USE_GNUTLS
 bool ssl_server_name_callback(ssl::stream_handle_type stream_handle, std::string const& name, void* arg)
 {
-	session_impl* si = reinterpret_cast<session_impl*>(arg);
+	auto si = reinterpret_cast<session_impl*>(arg);
```
arvidn, line 472: I really prefer to preserve the "pointerness" in auto expressions. i.e. `auto* si = ` (applies to this change and the one above)

src/torrent_info.cpp
```diff
@@ -473,7 +473,7 @@ namespace {
 			return false;
 		}
 
-		std::time_t const mtime = std::time_t(dict.dict_find_int_value("mtime", 0));
+		auto const mtime = std::time_t(dict.dict_find_int_value("mtime", 0));
```
arvidn, line 476: it would be even shorter to remove the `std::time_t` on the right hand side, like this:
`std::time_t const mtime(dict.dict_find_int_value("mtime", 0));`
I'm not sure which one I prefer

src/session_impl.cpp
FranciscoPombal: There is a `clang-tidy` check for this, btw: [readability-qualified-auto](https://clang.llvm.org/extra/clang-tidy/checks/readability-qualified-auto.html)

neheb: Fixed.

neheb:
Changed the pointer thing. I hate that change to be honest. It's less readable. Example:

```
const int i = 5;
const *char c = "5";

i = 6; //ERROR
c = 6; //OK ???

const auto j = 5;
const auto c = "5";

j = 6; //ERROR
c = "6"; //ERROR
```

It's specifically less readable with const pointers.

arvidn:
> It's specifically less readable with const pointers.

I would say it's because you use "const west" instead of "east const".

> const *char c = "5";

That's not valid syntax. I think you meant to say:

`char const* c = "5";` or `const char* c = "5";`

neheb:
> > It's specifically less readable with const pointers.
> 
> I would say it's because you use "const west" instead of "east const".
Not really.

```
auto const* c = "5";
c = "6"; //OK

auto const* const d = "5";
d = "6"; //ERROR
```
> 
> > const *char c = "5";
> 
> That's not valid syntax. I think you meant to say:
> 
> `char const* c = "5";` or `const char* c = "5";`
Ah yeah I meant const char *c.

arvidn:
Well, in your updated example, everything is regular. It's obvious that `c` is not immutable, because it doesn't have a `const` (and `const` is always on the right side of the thing it applies to).

arvidn:
What I find more interesting though is that the existing `clang-tidy` call on travis *already has* the `modernize-use-auto` check. But yet it isn't failing on these cases. There might be something wrong with how clang-tidy is run.

arvidn:
oops, yes. I'll have to fix this:

https://travis-ci.org/github/arvidn/libtorrent/jobs/697835150#L452

arvidn:
@neheb So, it turns out that clang-tidy hasn't been running correctly in CI for a while, and most of the changes you made were not caught immediately as they appeared. Restoring clang-tidy in CI for `RC_2_0` bring up a lot of the issues you posted changes against (which I then suggested putting in master).

In order to bring back clang-tidy, I will have to address some of these in there.

arvidn:
clang-tidy is now running on travis, and I've moved the configuration into a `.clang-tidy` file in `master`

neheb:
This commit seems pointless now. Closing.

PR #4759, opened 2020-06-12
[clang-tidy] remove bad const

neheb:
Found with performance-no-automatic-move


arvidn:
does this fix all reports from `performance-no-automatic-move`? if so, would you mind adding it to the clang-tidy command line on travis CI as well?

neheb:
I think so.

neheb:
Actually why is clang-tidy ran with the options manually? It looks for a .clang-tidy file automatically.

arvidn:
no strong reason. it's simple and one less file in the root of the repository.

PR #4768, opened 2020-06-13
ensure we don't keep redundant copies of merkle trees around

arvidn:
in an add_torrent_params object by moving it all the way down into the torrent constructor. Clearly document which fields in add_torrent_alert's params field can be relied upon to be preserved

### DIFF
--- a/include/libtorrent/torrent_info.hpp
+++ b/include/libtorrent/torrent_info.hpp
@@ -590,7 +590,7 @@ TORRENT_VERSION_NAMESPACE_3
 
 		// internal
 		aux::vector<aux::merkle_tree, file_index_t>& internal_merkle_trees();
-		void internal_load_merkle_trees(std::vector<std::vector<sha256_hash>> const& t);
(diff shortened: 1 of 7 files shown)


include/libtorrent/torrent_info.hpp
```diff
@@ -590,7 +590,7 @@ TORRENT_VERSION_NAMESPACE_3
 
 		// internal
 		aux::vector<aux::merkle_tree, file_index_t>& internal_merkle_trees();
-		void internal_load_merkle_trees(std::vector<std::vector<sha256_hash>> const& t);
+		void internal_load_merkle_trees(std::vector<std::vector<sha256_hash>> t);
```
arvidn, line 593: even though this function doesn't need a copy of this vector (it already copies it unconditionally), it takes it by value just to let the caller give up ownership, because we don't it to hang around. Perhaps this is a bit subtle, but at least there is an assert to ensure it happens

Issue #4354, opened 2020-02-27
Possible to have a session-total `num_peers` that does not include peers who are queued or attempting to connect?


FranciscoPombal:
**Please provide the following information**

libtorrent version (or branch): RC_1_2

platform/architecture: Any/amd64

compiler and compiler version: GCC 8.3.0

please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

This came up in https://github.com/qbittorrent/qBittorrent/pull/9232#issuecomment-592145644. Basically, we want to display/provide via a web API the total number of actually connected peers. But to do this currently, we need to iterate through all the torrents and accumulate the `num_peers` we get from there, because the `num_peers` that can be obtained from status() also includes peers who are queued or attempting to connect. Relevant code: https://github.com/qbittorrent/qBittorrent/blob/9649c2a1638f7090d2c7dfa0e84fceb8d9256eb4/src/webui/api/synccontroller.cpp#L139

In terms of API this is also rather confusing as there is the following asymmetry:
- `status()` has `num_peers` which includes every peer (actually connected or just queued/attempting) from every torrent.
- `torrent_status()` has `num_peers` which only includes actually connected peers for a single torrent, and `num_connections` which behaves like status()'s `num_peers` but for a single torrent.

Instead I would expect `status()` to also have `num_connections` and for `num_connections` and `num_peers` in `status()` to behave analogously to their counterparts in `torrent_status()`.

This is what I understood from reading the relevant entries in https://libtorrent.org/manual.html.
Please correct me if I am looking at this the wrong way or if I am simply not seeing an existing easy way of obtaining the desired result.

EDIT: what am I even doing... I should be looking in https://libtorrent.org/single-page-ref.html instead. But please consider updating the other page.

FranciscoPombal:
Just looked into the peer gauges. The docs says every gauge includes half open connections, but is there a way I can derive the number of peers excluding the half-open connections e.g. by summing/subtracting certain gauges from one another?

FranciscoPombal:
@arvidn ping

arvidn:
"queued" is not a thing anymore. That was back when there was a half-open connection limit.

https://libtorrent.org/manual.html has a version number 0.16. it's extremely old. There's no link pointing to this file anymore, and it's just here because when I update the website I just copy the existing files over the old ones (and this file doesn't exist in 1.0 and later). I should remove it.

It does seem reasonable to address this, for sure. The asymmetry is especially bad.

[...]

PR #4772, opened 2020-06-14
fix peer counters documentation

FranciscoPombal:
Fix one of the remaining points of https://github.com/arvidn/libtorrent/issues/4354#issuecomment-599283250 that I had forgotten about.


arvidn:
I think the only possibility for this PR to fail CI would be if `make spell-check` fails

arvidn:
which it didn't

Issue #4731, opened 2020-06-07
upload_mode doesn't completely prevent files/directories from being created


WGH-:
libtorrent version (or branch): 1.2.6
platform/architecture: Linux, amd64
compiler and compiler version: gcc 9.3.0

Although the documentation states that "In order to download just the metadata (.torrent file) from a magnet link, set the torrent_flags::upload_mode flag in add_torrent_params before adding the it.", empty files and directories are still _occasionally_ created after metadata is downloaded.

here's a repro Python script: https://gist.github.com/WGH-/37ec99e090ab390f105ca7dd142955b6

arvidn:
empty files and the directories they are in are created on startup, when the storage is constructed. I don't think this logic carese about whether the torrent is in upload_mode or not.

This happens when the torrent enters the state "checking_fast_resume". I suppose your torrents are "auto_managed" and only some of them are started to enter this state, and that's why not all of them have these files created.

To prevent the creation of those files, you can set the file priorities to 0 on a large number of files (since you won't know how many files are in the torrent before you have the metadata).

If you can come up with improvements to the documentation, patches are welcome!

WGH-:
> I suppose your torrents are "auto_managed" and only some of them are started to enter this state, and that's why not all of them have these files created.

I create the torrents with flags `params["flags"] = lt.torrent_flags.upload_mode | lt.torrent_flags.update_subscribe | lt.torrent_flags.apply_ip_filter`. That should disable the auto-managed mode, right?

Besides, I create all torrents with the same flags.

> To prevent the creation of those files, you can set the file priorities to 0 on a large number of files (since you won't know how many files are in the torrent before you have the metadata).

I've seen this suggestion before, and it still doesn't look right :)

arvidn:
upload_mode still needs to create the storage, so it can upload. so to prevent the empty files and directories from being created, it has to be some other option.

MassaRoddel:
There is also the simple solution to use an invalid save path. That prevents the creation of anything.

WGH-:
Okay, I'll submit a documentation patch that clarifies it in a few days.

Still, what bothers me is that files and directories are only sometimes created. Is there some kind of race?

arvidn:
files and the directories those files are in, are created on demand. i.e. when the file is first written to.
Some files will never be written to; empty files. So, those files are created up-front. This means all directories those empty files are in will also be created up-front.

[...]

PR #4778, opened 2020-06-17
Clarify docs wrt upload_mode and downloading metadata

WGH-:
See #4731

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -198,6 +198,11 @@ still be running, but it will disconnect the majority of peers (since connection
 to peers that already have the metadata are redundant). It will keep seeding the
 *metadata* only.
 
+Note that this doesn't prevent empty files from being created, if the torrent
+contains any. If you need to prevent that, you can either
+set ``file_priority`` to a long list of zeros (since the number of files is not known
+in advance), or set ``save_path`` to an invalid path.
+
 queuing
 =======
 


docs/manual.rst
```diff
@@ -198,6 +198,11 @@ still be running, but it will disconnect the majority of peers (since connection
 to peers that already have the metadata are redundant). It will keep seeding the
 *metadata* only.
 
+Note that this doesn't prevent files from being created, which may happen
+especially if the torrent contains empty files. If you need to prevent that, you can either
```
arvidn, line 202: I think you can make this sentence more succinct and precise. Something like "This does not prevent empty files from being created, if the torrent contains any"

docs/manual.rst
```diff
@@ -198,6 +198,11 @@ still be running, but it will disconnect the majority of peers (since connection
...
+set ``file_priority`` to a long list of zeroes (since the number of files is not known
+in advance), or set ``save_path`` to some invalid path.
```
arvidn, line 204: I think "an" is better than "some"

docs/manual.rst
```diff
@@ -198,6 +198,11 @@ still be running, but it will disconnect the majority of peers (since connection
 to peers that already have the metadata are redundant). It will keep seeding the
 *metadata* only.
 
+Note that this doesn't prevent empty files from being created, if the torrent
+contains any. If you need to prevent that, you can either
+set ``file_priority`` to a long list of zeroes (since the number of files is not known
```
arvidn, line 203: hunspell complains about "zeroes" ([here](https://travis-ci.org/github/arvidn/libtorrent/jobs/699389447#L1099)). I think it prefers "zeros"

arvidn:
could you squash these commits to?

PR #4783, opened 2020-06-22
transition last remaining boost::bind to std::bind

arvidn:


FranciscoPombal:
@arvidn I have been experimenting with the various fixes `clang-tidy` offers - would you be interested in a PR targeting `RC_2_0` converting most of `std::bind` to lambdas, as per https://clang.llvm.org/extra/clang-tidy/checks/modernize-avoid-bind.html? In that branch, C++14 is the minimum required version, making stuff like this possible:

```cpp
lt::xml_parse({reinterpret_cast<char const*>(data), size}
    , [&s](auto&& PH1, auto&& PH2, auto&& /*unused*/) { lt::find_control_url(PH1, PH2, s); });
```

The automated "fixits" seem smart enough to produce good results with little or no further manual intervention.

arvidn:
@FranciscoPombal I would like such changes to target `master`. Ideally after the webtorrent patch lands (which seems fairly close now). It seems unfair to make sweeping changes to cause merge conflicts at this point.

Also, I would like a human to inspect and judge every automated change, to make sure it's an improvement. In my experience, only a handful of rules are so good that they can be enforced across the board. Changing *all* `std::bind()` to lambdas, may produce better machine code, so that's an up-side. But it may also produce quite a lot more verbose code, which isn't that great in places where better machine code isn't all that important (especially in tests for instance).

going a bit off-topic; I have a somewhat large patch locally where I enable `readability-qualified-auto` which suggests really good changes 90% of the time, but the remaining 10% it gives suggestions that break the build on other compilers. Specifically, iterators are guaranteed to be pointers in libc++, but not in general. Yet it suggests capturing iterators as `auto*`.

I don't know what to do about that, since it can't be enforced on CI for this reason.

arvidn:
```
lt::xml_parse({reinterpret_cast<char const*>(data), size}
    , [&s](auto&& PH1, auto&& PH2, auto&& /*unused*/) { lt::find_control_url(PH1, PH2, s); });
```

Btw, wouldn't those arguments have to be forwarded there? like: `std::forward<decltype(PH1)>(PH1)`

FranciscoPombal:
@arvidn 

> @FranciscoPombal I would like such changes to target `master`. Ideally after the webtorrent patch lands (which seems fairly close now). It seems unfair to make sweeping changes to cause merge conflicts at this point.

:+1:, that's definitely the most important PR right now.

> Also, I would like a human to inspect and judge every automated change, to make sure it's an improvement. In my experience, only a handful of rules are so good that they can be enforced across the board. Changing _all_ `std::bind()` to lambdas, may produce better machine code, so that's an up-side. But it may also produce quite a lot more verbose code, which isn't that great in places where better machine code isn't all that important (especially in tests for instance).

Agreed about human inspection - the automated changes are just a good starting point. About verbosity - it's about on par, except in cases where there are 3+ placeholders (because `auto&& PH1, auto&& PH2, auto&& PH3, ...` starts to really add up to line length vs. `_1, _2, _3, ...`) and/or in cases where perfect forwarding is required, like you mentioned (I'm not sure if it is actually required in the supplied example, that was just for illustrative purposes).

> going a bit off-topic; I have a somewhat large patch locally where I enable `readability-qualified-auto` which suggests really good changes 90% of the time, but the remaining 10% it gives suggestions that break the build on other compilers. Specifically, iterators are guaranteed to be pointers in libc++, but not in general. Yet it suggests capturing iterators as `auto*`.

> I don't know what to do about that, since it can't be enforced on CI for this reason.

Interesting, I did not know about that potential breakage. I've also seen that change suggested when running myself, but haven't tried the automated suggestions. Could you point me to concrete examples of this? What (versions of) compilers fail?

arvidn:
@FranciscoPombal see this: https://github.com/arvidn/libtorrent/pull/4789

Issue #4720, opened 2020-06-03
Large buffer in save_resume_data_alert for v2 torrents


vktr:
**Please provide the following information**

libtorrent version (or branch): master

platform/architecture: win64

compiler and compiler version: msvc

please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

PicoTorrent tracks the master branch, and I just got around to verifying v2 torrents. Using the test torrent here (https://libtorrent.org/bittorrent-v2-test.torrent).

When I recieve the `save_resume_data_alert` for that torrent, the buffer is ~8MB, which is a large increase from v1 torrents. I haven't checked the code, but is this expected?

I generate the alert with

```
handle.save_resume_data(
                lt::torrent_handle::flush_disk_cache
                | lt::torrent_handle::save_info_dict);
```

vktr:
Clarification - the buffer I recieve from `lt::write_resume_data` is big for v2 torrents.

arvidn:
I suspect this is the `trees` field, which contains the full merkle tree for each file. Could you confirm this?, `dump_bencode` should do it.

I should probably optimize that format a bit, similarly to what I did for the in-memory representation.

vktr:
Can confirm, the `trees` field contains a lot of entries, most of them empty (or null) (`\x00`).

arvidn:
Here are some ideas of how this could be addressed:

If there is a full tree, just store the leaf nodes (this cuts the size in half).
If the tree is full down to the piece-layer, only store the piece layer.

For the partial tree case, one of these could help:

1. Use a simple RLE compression of the tree in the resume data.
2. use a bitmask to indicate which hashes are "set" (i.e. non-zero) and just store those.
3. store the only the "frontier" nodes. i.e. the ones closest to the leaf layer in each  sub-tree. This is a bit complicated to extract and to reconstruct.

I think right now, the simplest approach is the most feasible. It can always be improved down the line.

Thoughts?

vktr:
I can't say I fully understand the merkle trees - I'll have to do a bit of reading on that. That said, I am in favor of simple solutions :) I'm also leaning towards the second suggestion on using a bitmask, mostly to avoid processing to stored data which I guess would make it easier to debug by eye.

arvidn:
@vktr would you mind testing that PR to see if it improves things sufficiently for you?

vktr:
[...]

PR #4788, opened 2020-06-24
Compress resume trees

arvidn:
This is meant to address https://github.com/arvidn/libtorrent/issues/4720

It does so by not storing the full merkle trees in the resume data. A new mask field is added that specifies which hashes are included in the tree, and the tree can then be reconstructed from the hashes that are included. If only the piece-layer or block layer is included, there's a short circuit to go directly into an internal sparse representation as well.

Since there's a fair amount of logic and opportunities to run into trouble when adding a torrent now (since `add_torrent_params` has more invariants) I also added a new fuzzer that adds torrents with random resume data.

I also took the opportunity to make the internal representation of `verified_leaf_hashes` match the one in `add_torrent_params`, so we don't have to copy it into a new type.

### DIFF
--- a/src/merkle_tree.cpp
+++ b/src/merkle_tree.cpp
@@ -67,6 +67,109 @@ namespace aux {
 		optimize_storage();
 	}
 
+	void merkle_tree::clear()
+	{
+		m_tree.clear();
+		m_tree.shrink_to_fit();
+		m_mode = mode_t::empty_tree;
+	}
+
+namespace {
+
+	// TODO: in C++20, use std::identity
+	struct identity
+	{
+		bool operator()(bool b) const { return b; }
+	};
+}
+
+	void merkle_tree::load_sparse_tree(span<sha256_hash const> t, std::vector<bool> const& mask)
+	{
+		TORRENT_ASSERT(mask.size() == size());
+		if (size() != mask.size()) return;
+
+		int const first_block = block_layer_start();
+		int const end_block = first_block + m_num_blocks;
+
+		TORRENT_ASSERT(first_block < int(mask.size()));
+		TORRENT_ASSERT(end_block <= int(mask.size()));
+
+		// if the mask covers all blocks, go straight to block_layer
+		// mode, and validate
+		if (std::all_of(mask.begin() + first_block, mask.begin() + end_block, identity()))
+		{
+			// the index in t that points to first_block
+			auto const block_index = std::count_if(mask.begin(), mask.begin() + first_block, identity());
+
+			// discrepancy
+			if (t.size() < block_index + m_num_blocks)
+				return clear();
+
+			m_tree.assign(t.begin() + block_index, t.begin() + block_index + m_num_blocks);
+			m_mode = mode_t::block_layer;
+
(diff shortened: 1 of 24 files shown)


src/merkle_tree.cpp
```diff
@@ -67,6 +67,109 @@ namespace aux {
...
+		TORRENT_ASSERT(mask.size() == size());
+		if (size() != mask.size()) return;
+
+		int const first_block = block_layer_start();
+		int const end_block = first_block + m_num_blocks;
+
+		TORRENT_ASSERT(first_block < int(mask.size()));
+		TORRENT_ASSERT(end_block <= int(mask.size()));
+
+		// if the mask convers all blocks, go straight to block_layer
```
arvidn, line 97: typo: covers

src/merkle_tree.cpp
```diff
@@ -423,6 +526,47 @@ namespace aux {
...
+		if (m_mode == mode_t::uninitialized_tree) return {{}, {}};
+
+		std::vector<bool> mask(size(), false);
+		std::vector<sha256_hash> ret;
+		switch (m_mode)
+		{
+			case mode_t::uninitialized_tree: break;
+			case mode_t::empty_tree: break;
+			case mode_t::full_tree:
+				for (int i = 0, end = int(m_tree.size()); i < end; ++i)
```
arvidn, line 540: I should use `m_tree.end_index()` instead of `int(m_tree.size())`

src/torrent.cpp
```diff
@@ -6797,7 +6793,14 @@ namespace {
...
 			ret.merkle_trees.reserve(m_torrent_file->internal_merkle_trees().size());
 			for (auto const& t : m_torrent_file->internal_merkle_trees())
-				ret.merkle_trees.emplace_back(t.build_vector());
+			{
+				// use stuctured binding in C++17
+				std::vector<bool> mask;
+				std::vector<sha256_hash> sparse_tree;
+				std::tie(sparse_tree, mask) = t.build_sparse_vector();
+				ret.merkle_trees.emplace_back(sparse_tree);
+				ret.merkle_tree_mask.emplace_back(mask);
```
arvidn, line 6802: `sparse_tree` and `mask` should be moved in here

src/merkle_tree.cpp
```diff
@@ -423,6 +526,47 @@ namespace aux {
 		return ret;
 	}
 
+	std::pair<std::vector<sha256_hash>, std::vector<bool>> merkle_tree::build_sparse_vector() const
```
arvidn, line 529: it should be possible to use `aux::vector<bool>` here, for `mask`

arvidn:
@ssiloti I've hit some edge cases that aren't working in here. I have a question. If I have a file that has two *blocks* (let's say it's 32 kiB) and a piece size of 1 MiB (i.e. normally 64 blocks per piece). This file will have a single piece, right?
Will the merkle tree have 64 leafs or will it have 2 leafs?

ssiloti:
I think it should have 64 leafs. Otherwise it creates an extra special case for not much benefit.

PR #4789, opened 2020-06-24
apply (most of) readability-qualified-auto changes

arvidn:


src/peer_list.cpp
```diff
@@ -98,15 +98,15 @@ namespace libtorrent {
 
 	void peer_list::clear()
 	{
-		for (auto const p : m_peers)
+		for (auto* p : m_peers)
```
FranciscoPombal, line 101: Why not `for (auto* const p : m_peers)` in these 2 cases?

arvidn: it didn't seem all that important to make `p` const, since it's scope is very limited

FranciscoPombal: Well, I don't really have any arguments against the change besides "moar `const` moar better", so yeah.

arvidn:
I guess clang-tidy 7 doesn't have this checker either

FranciscoPombal:
https://github.com/arvidn/libtorrent/pull/4790#issuecomment-649132669

> @arvidn You can check all supported checks with `clang-tidy -checks="*" --explain-config`
>
>Looks like for this one at least `clang-tidy` 8 is required. For `readability-qualified-auto` in https://github.com/arvidn/libtorrent/pull/4789, the minimum requirement is version 10.

`clang-tidy-10` is only available on Ubuntu >= 20.04. So for this one, either wait until Travis supports 20.04 images, or perhaps consider using LLVM's official apt repository: https://apt.llvm.org/

PR #4793, opened 2020-06-25
use weak pointers to hold http_tracker_connection

arvidn:
This is a bit of a hack, an experiment to narrow down the HTTPS tracker memory leak.


arvidn:
it seems like introducing a queue for issuing tracker requests solved (or at least mitigated) the issue

PR #4795, opened 2020-06-25
Raise default piece limit to meet the needs of larger torrents

xloem:
I was working with a tool that doesn't provide for configuring the max piece limit and found I couldn't download very large torrents at all.

This patch raises the limit to be less restrictive, providing compatibility with very large torrents.

### DIFF
--- a/include/libtorrent/torrent_info.hpp
+++ b/include/libtorrent/torrent_info.hpp
@@ -131,7 +131,7 @@ namespace aux {
 	{
 		int max_buffer_size = 6000000;
 		// the max number of pieces allowed in the torrent
-		int max_pieces = 0x100000;
+		int max_pieces = 0x1000000;
 		// the max recursion depth in the bdecoded structure
 		int max_decode_depth = 100;
 		// the max number of bdecode tokens
--- a/src/torrent_info.cpp
+++ b/src/torrent_info.cpp
@@ -80,10 +80,10 @@ namespace libtorrent {
 	// this is an arbitrary limit to avoid malicious torrents causing
 	// unreasaonably large allocations for the merkle hash tree
 	// the size of the tree would be max_pieces * sizeof(int) * 2
-	// which is about 8 MB with this limit
+	// which is about 256 MB on a 64-bit system with this limit
 	// TODO: remove this limit and the overloads that imply it, in favour of
 	// using load_torrent_limits
-	constexpr int default_piece_limit = 0x100000;
+	constexpr int default_piece_limit = 0x1000000;
 
 	bool valid_path_character(std::int32_t const c)
 	{
--- a/src/torrent_info.cpp
+++ b/src/torrent_info.cpp
@@ -80,7 +80,7 @@ namespace libtorrent {
 	// this is an arbitrary limit to avoid malicious torrents causing
 	// unreasaonably large allocations for the merkle hash tree
 	// the size of the tree would be max_pieces * sizeof(int) * 2
-	// which is about 256 MB on a 64-bit system with this limit
+	// which is about 128 MB with this limit
 	// TODO: remove this limit and the overloads that imply it, in favour of
 	// using load_torrent_limits
 	constexpr int default_piece_limit = 0x1000000;


include/libtorrent/torrent_info.hpp
```diff
@@ -131,7 +131,7 @@ namespace aux {
 	{
 		int max_buffer_size = 6000000;
 		// the max number of pieces allowed in the torrent
-		int max_pieces = 0x100000;
```
arvidn, line 134: Is there a reason the tool you use can't be changed instead, to use a higher limit?
That's the point of having `load_torrent_limits` in the first place, to let clients configure it. Whether that limit is exposed to the *user* is a different issue, if you just want to hard code it, it doesn't have to be.

arvidn:
in case the tool is written in python and so can't configure this at runtime, I'm fixing that here: https://github.com/arvidn/libtorrent/pull/4796

xloem:
The purpose is so that users of tools who don't know how to rewrite them
can download larger torrents.

include/libtorrent/torrent_info.hpp
xloem: The limit is still present and still defends against DOS.  The change provides default support for more torrents, for users unfamiliar with adjusting it.  Do you have a concern with it?

FranciscoPombal:
@xloem seems to me like the client/tool in question needs to be changed, not libtorrent. What is the client/tool? Are they accepting patches?

include/libtorrent/torrent_info.hpp
arvidn: right, it's not obvious that it "helps" the majority of clients though. It might "break" most of the users (who knows). Any computer that can't afford to allocate 30 megabytes for a .torrent file would have an issue with this patch.

xloem: what devices does libtorrent target that can't allocate 30MB?

Honestly the torrent I was worried about people not being able to access only needs a very small bump to the old value.  I tried to pick a value that seemed reasonable for use cases I was familiar with.  Would you accept a smaller increase?

xloem:
The tool that helped me discover this user experience is bcat, written in python: i was trying to stream a torrent larger than I had disk space for.  I'm assuming there are _other_ client applications that use libtorrent's defaults, and that this number can be increased _without_harm_: it looked arbitrary to me, to detect corruption.  What do you think a good value would be?

src/torrent_info.cpp
```diff
@@ -80,10 +80,10 @@ namespace libtorrent {
 	// this is an arbitrary limit to avoid malicious torrents causing
 	// unreasaonably large allocations for the merkle hash tree
 	// the size of the tree would be max_pieces * sizeof(int) * 2
-	// which is about 8 MB with this limit
+	// which is about 256 MB on a 64-bit system with this limit
```
FranciscoPombal, line 83: With`constexpr int default_piece_limit = 0x1000000;`, the max would be 128 MiB. `sizeof(int)` is still `4` on a 64-bit system (in practice, because in theory it can be anything).

include/libtorrent/torrent_info.hpp
arvidn: > what devices does libtorrent target that can't allocate 30MB?

I'm not convinced that that's the most relevant question. Any device will have degraded performance by allocating more memory than necessary. I would think a better question would be; "what is the 99th percentile of torrent sizes in the wild?", or maybe something along those lines.

basically, at what point is it likely the torrent is malicious. But also allocating a relatively large amount of memory is of course relevant too.

> Would you accept a smaller increase?

I would like a stronger argument why not each client can pick this limit themselves. It would seem the client is in 
a much better position to judge what reasonable limits are. Also, these limits only apply to certain "convenience" constructors of `torrent_info`, ones that loads from disk and parses bencoding. It's not limits that apply broadly to all aspects of libtorrent.

src/torrent_info.cpp
xloem: Thanks so much.  I fixed this.

xloem:
>
>
> I would think a better question would be; "what is the 99th percentile of
> torrent sizes in the wild?", or maybe something along those lines.
>
> basically, at what point is it likely the torrent is malicious. But also
> allocating a relatively large amount of memory is of course relevant too.
>
It seems to me a malicious torrent would have numbers that are beyond the
allocatable ram in a system, such that loading it freezes up the system and
force closes the process and likely other processes.  Hence, capacity of
the target device seems the discerning factor to me.  Facebook takes
hundreds of megabytes, so 30mb would seem fine to me.

> Would you accept a smaller increase?
>
> I would like a stronger argument why not each client can pick this limit
> themselves.
>
It's not that they can't theoretically pick it: it's that not all clients
will be maintained to do so, and not all users will know to configure it.
This then leaves a section of the population out of large torrents, if
these casual downloaders give up when encountering the error.

arvidn:
> It's not that they can't theoretically pick it: it's that not all clients
> will be maintained to do so, and not all users will know to configure it.
> This then leaves a section of the population out of large torrents, if
> these casual downloaders give up when encountering the error.

That's a fair point. Especially if the average size of torrents is increasing, it would definitely make sense to increase these limits too. I would like to gather some data to base a new limit on though

arvidn:
and I think it would be fine to increase this limit in the stable `RC_1_2` branch too

xloem:
If we come to agreement, would you like a separate PR for that branch?

I'm afraid I don't have the capacity to scrape torrents and calculate the distribution of their part counts, and I don't really see why that is needed to safely increase the limit when there is so much ram in computers now, but thank you for hearing me out on the issue.  I just felt so sad when my torrent wouldn't download, I wanted to try to prevent the situation for others in the future.

arvidn:
I think surveying the "available RAM" is more difficult and requires more unsafe assumptions about the kinds of devices that may be running libtorrent, and what other processes those devices may be running. Looking at torrents in the wild, I think, is a more concrete measurement of what is expected to work.

xloem:
Thanks.  We disagree.  I'm sad and defeated.  Thank you for a great library.

arvidn:
how would you survey how much RAM is reasonable to require? There must be something better than speculating based on one's own use case.

FranciscoPombal:
@xloem I can understand your argument about increasing RAM capacities, but consider that this does not necessarily mean that everyone is now running a single instance of a torrent client on machines with 32 GiB RAM. A lot of people took the opportunity to partition the total capacity to use multiple instances instead. Think several container instances each running a torrent client.

So, even though the overall capacity of physical machines has increased, the average effective RAM available in envrionments where people are running torrent clients might not have increased the same way. And it is probably really difficult to gather data about this.

arvidn:
This is what I've gathered on a short notice. I have used some data sets I already had available (of ~63000 and ~42000 torrents) as well as an early scrape program I threw together (but it's only gathered 4664 torrents so far).

So, the input data is not great. I'll try to improve this over time, to make sure I deduplicate and maybe only use recent torrents.
Another caveat is that the metadata size is just the info-dict. A full .torrent file would likely be a little bit larger to account for trackers, comment, and creator. Probably not significantly larger though.

These are cumulative distribution functions (CDF). The y-axis is the portion of all torrents that are smaller than the size on the x-axis. i.e. it's the integral of the frequency function.

image

image

image

xloem:
A raspberry pi has 2 GiB of ram.  Allocating data that fits within ram is generally harmless unless the data is frequently randomly seeked causing many swaps, and another process has also allocated a lot of ram and has similar behavior.  I would consider 1GiB total process allocation to be hands-down safe (1/2 a raspberry pi or tiny vps).  256 MiB/torrent would then be 4 or more active torrents, all of them the maximum size.  16MiB/torrent could be 64 maximum-sized torrents.

Thanks for making the chart up.  Looks like most torrents aren't very big.

I ended up buying a harddrive for other reasons and downloading my whole torrent at once with rtorrent.  The rtorrent version in ubuntu uses an older release of libtorrent, and I didn't run into the piece count error using it.

FranciscoPombal:
@arvidn 

Thanks for the data. A couple of questions and a suggestion for data processing:

What is the source of this data, and how are you collecting it? Are you scraping trackers? If so, which?

These graphs are interesting for a general overview, but I think it would also be very interesting and relevant to the conversation at hand to cluster the data according to torrent payload type. These graphs only tell us that ~99% of "all" torrents are < 10 GiB, and ~99 % have piece size < 8 MiB. But this is somewhat expected, given that it is obvious small torrent payloads (books/text, music, game ROMs from old consoles, etc) are always going to heavily outnumber larger payloads (like those of "large Linux ISOs read by 405 nm diode lasers", big ass datasets, etc), thus bringing the average payload/piece/metadata size down.

I would expect to see quite varied piece size/metadata size CDFs across different payload types, and quite a difference from the average of all torrents represented in the graph that you posted. In fact, I would probably be surprised if there were a single payload type whose characteristics followed closely to the average. For example, for torrents of "large Linux ISOs read by 405 nm diode lasers", I would expect near 0% to have a piece size of 2 MiB or less.

IMO it is important to cater to torrents with large payloads, since large payloads are the type of data that most benefits from, and depends on the BitTorrent network and ecosystem for easy distribution. Decisions taken with only the average metrics run the risk of making things worse for these large payloads.

FranciscoPombal:
> I ended up buying a harddrive for other reasons and downloading my whole torrent at once with rtorrent. The rtorrent version in ubuntu uses an older release of libtorrent, and I didn't run into the piece count error using it.

Note that `rtorrent` uses a different libtorrent, often called `libtorrent-rakshasa`: https://github.com/rakshasa/libtorrent

arvidn:
> What is the source of this data, and how are you collecting it? Are you scraping trackers? If so, which?

One data set was from BitSnoop, that they released when they went "belly up". The other I scraped from the DHT, probably almost 10 years ago. It remembered that I have the SkyTorrent data as well (I think that's what they're called, but not online so I can reach it right now). All of those are "old" though. They can't really be used to show the *current* distribution.

I threw together another DHT scraper just now, but it needs some tuning to work well.

FranciscoPombal:
@arvidn Right now the biggest public trackers I know of are `http://coppersurfer.tk/` and `https://opentrackr.org/` (both these URLs point to web pages with information; they serve on the order of 10^6~10^7 torrents/peers). This site might be useful to you for finding other trackers: https://newtrackon.com/.

arvidn:
I just realized this patch is proposing raising the limit of *number of pieces*. I've been thinking it was about the `max_buffer_size`.

I think that's reasonable to do. I've been analyzing the sky-torrent dump and so for I have encountered two torrents that failed to load because of the default limits.

1:
The .torrent file is 24 MiB (and the default limit is 6 MiB)
The bencoded structure has about 2269800 "tokens", where the default limit is 2000000.
Piece size: 2 MB
Number of pieces: just under 200000 (the default limit is 1048576).
Number of files: ~166000

2:
The .torrent file is about 16 MiB (and the default limit is 6 MiB)
The bencoded structure has about 2200000 "tokens", where the default limit is 2000000.
Piece size: 2 MB
Number of pieces: about 2500 (the default limit is 1048576).
Number of files: ~182000

In both cases, the majority of the size of the torrents is the file list, specifically long paths repeated for every file. They would most likely be significantly smaller as v2 torrents.

@xloem how many pieces does your torrent have, that failed to load?

arvidn:
I don't think that comment about merkle trees is particularly relevant anymore. Partly because nobody is really using those kinds of torrents anymore, and v2 will supersede it, and libtorrent has deprecated them.

I'm thinking that it might be reasonable to bump these limits:

.torrent size: 6 MB -> 10 MB
bdecode-tokens: 2'000'000 -> 3'000'000
max-pieces: 1048576 -> 2097152

arvidn:
In the sky-torrent dump, the torrent with the most pieces has just over 1'000'000 pieces, which is just below the current limit. Doubling the limit seems reasonable, to get some head room.

image

image

arvidn:
https://github.com/arvidn/libtorrent/pull/4815

arvidn:
closing, as I believe #4815 supersedes this change.

@xloem please check out another related patch! https://github.com/arvidn/libtorrent/pull/4820

PR #4802, opened 2020-06-27
don't leave lingering part files handles open

arvidn:


arvidn:
@xuefer could you try this updated patch?

xuefer:
> @xuefer could you try this updated patch?

better. no linger parts fd after restart (scanning torrents)

arvidn:
@FranciscoPombal @ssiloti Does this make sense?

The background is: xuefer highlighted that the file descriptors held by `part_file` are not subject to the limits (nor managed by) the `file_pool` which is problematic since just having a lot of torrents may bust the `file_pool_size` limit and trigger `EMFILE` errors.

This patch removes the persistent open file in `part_file` and instead opens and closes the file for each part file operation (read, write & flush header). My reasoning is that part file operations shouldn't be too frequent anyway, so the additional cost should be negligible.

FranciscoPombal:
@arvidn 

Couldn't this lead to performance problems in torrents with a lot of smaller files? In that case it is reasonable to assume most will be misaligned at piece boundary, meaning there will be a lot part files for which to "read, write & flush header", right? What about keeping a separate pool for these files (e.g. `partfile_pool_size`)? I can't really benchmark this in the near future though, sorry.

arvidn:
> Couldn't this lead to performance problems in torrents with a lot of smaller files? In that case it is reasonable to assume most will be misaligned at piece boundary, meaning there will be a lot part files for which to "read, write & flush header", right?

the part file is only used for files whose priority is 0, so in your scenario you also need to set every other file to priority 0, to maximize part file use. Yes, in that case it might lead to performance problems (but it's not all that likely to, opening files is surprisingly fast)

arvidn:
I think the long-term fix (probably something for 2.0.1 or something) would be to have the part file managed by the file pool

ssiloti:
In the past I once tried this open/access/close pattern and found it to be surprisingly slow on Windows. Perhaps it's not slow enough to matter in this case, but something to look out for.

arvidn:
and on windows there's virtually no limit on the number of open files instead.. maybe a more conservative change would be to utilize the periodic flushing of the header, to also close the file afterwards.

It would still break the case of starting up and checking 1000 torrents, all of which have a part file. Then you'll have 1000 extra file descriptors open, for 2 minutes until the storage is `tick()`ed.

arvidn:
I'm a bit torn, but I think I'm leaning towards the second (current) option, of opening and closing files on-demand. I think there's a risk that it becomes a performance issue on windows, especially with anti-virus hooking open and close calls.

However, I think you need a pretty unusual situation to actually suffer from it. Accessing part files shouldn't be very common under normal circumstances.

I'm also not very happy with how the part file header is delay flushed right now. With memory mapped files it would be a lot more efficient to keep the header mapped into RAM and let the kernel decide when to flush it.

Issue #4801, opened 2020-06-27
posix_fallocate detection (TORRENT_HAS_FALLOCATE)


libTorrentUser:
`TORRENT_HAS_FALLOCATE` is being "incorrectly" set to 0 on some systems where `posix_fallocate` is available. 

I confirmed the problem is happening on Alpine Linux, but I believe it will also happen on any system that uses `musl libc` because, although I believe it is supposed to conform to POSIX 2008 (`_POSIX_C_SOURCE=200809L`), it seems that the symbol is not automatically defined. Nor is `_XOPEN_SOURCE`. Therefore

[config.hpp line 179](https://github.com/arvidn/libtorrent/blob/55d9ee3451a480facb58935515236c79b1f9e097/include/libtorrent/config.hpp#L179)

    // posix_fallocate() is available under this condition
    #if _XOPEN_SOURCE >= 600 || _POSIX_C_SOURCE >= 200112L
    #define TORRENT_HAS_FALLOCATE 1
    #else
    #define TORRENT_HAS_FALLOCATE 0
    #endif

will set `TORRENT_HAS_FALLOCATE` to 0

This issue happens on the 1_2 branch. It was tested on 1.2.7 but I believe it also happens on 2_0 because the relevant parts of `config.hpp` remains the same.

Although I believe `musl libc` is the one to blame here (I think the library should be the one defining those two symbols in case they are not set by the user), since it does not, I think it would be safer to just surround the above with 

    #ifndef TORRENT_HAS_FALLOCATE
    ...
    #endif  // #ifndef TORRENT_HAS_FALLOCATE

this way, users building on platforms that do not define `_XOPEN_SOURCE` or `_POSIX_C_SOURCE` can simply define `TORRENT_HAS_FALLOCATE` in order to make `posix_fallocate` available. 

It is true that the user could simply define one of those two symbols to some appropriate value and we would not need to change a thing inside `config.hpp`, but I believe `TORRENT_HAS_FALLOCATE` has a very well defined and narrow scope, while defining any of those other symbols to wrong values may end up breaking way more stuff.

And last, but not least, maybe we could return an error or an alert or something to an user that sets `lt::storage_mode_t::storage_mode_allocate` in a system that cannot  (or was not configured to) use any of the currently available methods libtorrent uses to pre-allocate files. I believe this is particularly important on Linux, where `ls` doesn't usually report the real file size, unlike `du -h`.

arvidn:
[...]

PR #4804, opened 2020-06-27
detect posix_fallocate

arvidn:
instead of detecting whether posix_fallocate() is available, detect whether it's *not* available and otherwise always assume it is

### DIFF
--- a/include/libtorrent/config.hpp
+++ b/include/libtorrent/config.hpp
@@ -176,10 +176,10 @@ POSSIBILITY OF SUCH DAMAGE.
 #define TORRENT_USE_ICONV 0
 #else // ANDROID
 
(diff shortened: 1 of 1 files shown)


FranciscoPombal:
https://github.com/arvidn/libtorrent/issues/4801#issuecomment-651390944

Shouldn't this patch include the necessary change to the CML? I think you just need to add `target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_HAS_FALLOCATE)` pretty much anywhere.

arvidn:
@FranciscoPombal further down the `config.hpp` file. there is already a check to see if `TORRENT_HAS_FALLOCATE` is defined, and if it isn't, defines it to 1

PR #4808, opened 2020-06-28
clean up seed-mode with resume data

arvidn:
and ensure we leave seed-mode if there are missing files

### DIFF
--- a/src/disk_io_thread.cpp
+++ b/src/disk_io_thread.cpp
@@ -2588,8 +2588,6 @@ constexpr disk_job_flags_t disk_interface::cache_hit;
 		if (m_settings.get_bool(settings_pack::no_recheck_incomplete_resume))
 			return status_t::no_error;
 
-		// if the seed mode flag is not set, and the piece vector is empty, it
-		// suggests we're starting without any resume data
 		if (!aux::contains_resume_data(*rd))
 		{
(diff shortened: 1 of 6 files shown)


src/storage_utils.cpp
```diff
@@ -507,12 +507,54 @@ namespace libtorrent { namespace aux {
...
+				}
+
+				if (size != fs.file_size(file_index))
+				{
+					// the resume data indicates we're a seed, but this file has
+					// the wrong size. Reject the resume data
+					ec.ec = errors::mismatching_file_size;
+					ec.file(file_index);
+					ec.operation = operation_t::check_resume;
+					return false;
```
arvidn, line 549: there's a lot of duplicated code here. I should factor it out.

src/disk_io_thread.cpp
```diff
@@ -2582,7 +2582,8 @@ constexpr disk_job_flags_t disk_interface::cache_hit;
 		// the error message indicates that the fast resume data was rejected
 		// if 'fatal_disk_error' is returned, the error message indicates what
 		// when wrong in the disk access
-		if ((rd->have_pieces.empty() || !verify_success)
+		if (((!(rd->flags & torrent_flags::seed_mode)
```
arvidn, line 2585: this expression is becoming unwieldy, it should be broken up

PR #4809, opened 2020-06-28
fix race when two threads are opening the same file

arvidn:
one for reading and one for writing, and the writing thread loses the race

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -252,6 +252,7 @@ TOOLS_FILES= \
   Jamfile                \
   dht_put.cpp            \
   dht_sample.cpp         \
+  disk_io_stress_test.cpp\
   parse_dht_log.py       \
   parse_dht_rtt.py       \
   parse_dht_stats.py     \
--- a/src/file_view_pool.cpp
+++ b/src/file_view_pool.cpp
@@ -68,39 +68,25 @@ namespace libtorrent { namespace aux {
 		// this member to be destructed after we release the std::mutex. On some
 		// operating systems (such as OSX) closing a file may take a long
 		// time. We don't want to hold the std::mutex for that.
-		std::shared_ptr<file_mapping> defer_destruction;
+		std::shared_ptr<file_mapping> defer_destruction1;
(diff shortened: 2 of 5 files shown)


arvidn:
one concern is that; imagine the writer thread loses the race, the read-only file is left in the `file_view_pool`.

1. The reader reads some part of the file (we're guaranteed that it won't be reading anything overlapping what the writer is about to write).
2. The writer writes to its own file and closes it as it goes out of scope.
3. Imagine another job now comes in to read the bytes just written in step 2. It will pick up the read-only file mapping from the `file_view_pool` and look there. Will the bytes written in step 2 be visible in the read-only mapping? I don't think it would be on linux. The views would have different virtual addresses.

It seems safer to me to have the "stronger" file always win. i.e. if a read-write file loses, and the existing file is read-only, it overwrites the file in the pool.

arvidn:
I added a commit that implements a slightly different approach.

> the first lookup is only used if the open-mode is a match, otherwise we always move on to opening the file. The second lookup/insert replaces the entry if the existing one has the wrong open-mode.

I think this might be safer since subsequent *reads* will use the same mapping that was used to write.

I'm starting to think that there probably should be a stress test for the disk I/O, to catch things like this

MassaRoddel:
Just a thought ...
If several threads (running in parallel on different cores) use the same memory mapped files doesn't that have a bad impact on the cpu core caches because they have to be synchornized?

arvidn:
> If several threads (running in parallel on different cores) use the same memory mapped files doesn't that have a bad impact on the cpu core caches because they have to be synchornized?

I would not expect so. Threads never overlap *writes*, and threads should already be synchronizing from a write to reading the same block. Once a block is written, it's immutable, so from that point it's fine for it to exist in multiple caches, reading won't invalidate another core's cache.

arvidn:
@MassaRoddel can you still reproduce the issue with this patch?

MassaRoddel:
It seems to be fixed.

PR #4820, opened 2020-07-02
remove hard coded piece-count limits and expose them to clients

arvidn:
The main motivation of this patch is to remove the hard-coded constant that limits the number of pieces allowed in a torrent, and make sure clients are always able to raise the default limit.

torrents are not only parsed in `torrent_info`'s constructor, but also in the (publicly exposed) `torrent_info::parse_info_section()` which is used when loading just the info-section from a magnet link. It's also used when loading resume data that contains the info section (in `read_resume_data()`).

To enable loading large torrents via magnet links, I added a setting to `settings_pack`.

To enable loading large torrents from resume files, I added an extra parameter to all `load_resume_data()` overloads.

@FranciscoPombal @xloem does this seem reasonable? (not that I haven't merged the other patch in yet, it will come in a while)

### DIFF
--- a/bindings/python/src/session.cpp
+++ b/bindings/python/src/session.cpp
@@ -50,6 +50,9 @@ namespace boost
 using namespace boost::python;
 using namespace lt;
 
+// defined in torrent_info.cpp
+load_torrent_limits dict_to_limits(dict limits);
+
 namespace
 {
 #if TORRENT_ABI_VERSION == 1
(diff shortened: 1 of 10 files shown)


FranciscoPombal:
Having the possibility to load larger torrents via magnet links and resume files is nice :+1:

But how about having each field of `load_torrent_limits` exposed in the settings, and constructing the default `load_torrent_limits` based on that? Seems odd for the `max_pieces` setting to only affect magnet links and to have no effect on `load_torrent_limits` (in fact, the setting becomes something separate from `load_torrent_limits.max_pieces`).

arvidn:
> But how about having each field of load_torrent_limits exposed in the settings, and constructing the default load_torrent_limits based on that?

Right, it's not ideal to have this configuration in so many places, however, when constructing a `torrent_info`, or when calling `read_resume_data()`, there may not be a `settings_pack` anywhere. There may not be a session even, or there may be multiple sessions with different settings.

arvidn:
that said. I should make sure the documentation cross references the other places

PR #4822, opened 2020-07-02
rename error_message() to failure_reason() on tracker_error_alert

arvidn:
and improve documentation

### DIFF
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -516,11 +516,6 @@ TORRENT_VERSION_NAMESPACE_2
 	// This alert is generated on tracker time outs, premature disconnects,
 	// invalid response or a HTTP response other than "200 OK". From the alert
 	// you can get the handle to the torrent the tracker belongs to.
-	//
-	// The ``times_in_row`` member says how many times in a row this tracker has
-	// failed. ``status_code`` is the code returned from the HTTP server. 401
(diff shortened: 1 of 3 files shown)


include/libtorrent/alert_types.hpp
```diff
@@ -533,13 +528,23 @@ TORRENT_VERSION_NAMESPACE_2
 		static constexpr alert_category_t static_category = alert_category::tracker | alert_category::error;
 		std::string message() const override;
 
+		// This member says how many times in a row this tracker has failed.
 		int const times_in_row;
+
+		// the error code indicating why the tracker announce failed. If it is
+		// is ``lt::errors::tracker_failure`` the failure_reason() might contain
+		// a more detailed description of why the tracker rejected the request.
+		// HTTP status codes indicating errors are also set in this field.
```
FranciscoPombal, line 537: What about other kinds of errors? Like `Host not found (non-authoritative)`? They have to be checked elsewhere, right? That should probably be mentioned here.

arvidn: I don't really want to enumerate *all* categories of errors that can happen. But yes, all errors that can happen will be indicated by this error code

FranciscoPombal: Never mind, I misunderstood something, I also don't think all possible categories should be enumerated here.

PR #4824, opened 2020-07-03
[clang-tidy] use nodiscard

neheb:
Found with modernize-use-nodiscard


arvidn:
I'm a bit hesitant to mark all the `error_category` members nodiscard. They are never used directly, but only through the base class interface, so I don't think it makes a practical difference, but it adds some noise.

arvidn:
https://travis-ci.org/github/arvidn/libtorrent/builds/704511713

arvidn:
I'm pretty sure that call to `metadata()` should be removed

PR #4837, opened 2020-07-04
choker: reduce a bit of duplication with a lambda.

unaPoloGTIc:
Sorry for the extra PR.


unaPoloGTIc:
Not a big imptovement. But the logic was duplicated, any fix in one would have to be replicated in the other.
I'm not attchad to this, if you don't want it that's cool. I'll try to be more helpful in the next try. Only doing it to get involved.

PR #4846, opened 2020-07-05
fix issue when receiving have-all message before the metadata

arvidn:
for torrents with a lot of pieces


src/peer_connection.cpp
```diff
@@ -1917,8 +1921,7 @@ namespace libtorrent {
 
 		if (!t->valid_metadata() && index >= m_have_piece.end_index())
 		{
-			// TODO: 3 replace this magic number with something that makes sense
```
arvidn, line 1920: this is fixed in libtorrent-2.0

PR #4848, opened 2020-07-05
removed openssl related code to deal with old versions

aldenml:


arvidn:
maybe I misunderstood. I thought all the code you removed is no longer needed in 2.0, *because* older version of openssl are no longer supported. It looks like a good change to me

aldenml:
Well, I would say the issue is mostly in `master` with `webtorrent` on. I rather not change anything here in `RC_2_0`, but are you OK with removing support for openssl pre 1.1 in `master`?

arvidn:
oh, I see. I misread the version check in the code. Do you think it's reasonable to drop support for pre 1.1 in master?
it probably is. It would also simplify the Jamfile for windows, since the library names are unified on windows and unix then.

aldenml:
I think it's, they will be releasing a major version soon, so openssl 1.0.X is way in the past, and libtorrent `master` is way in the future :)

PR #4859, opened 2020-07-09
some fixes with export symbols

aldenml:


include/libtorrent/announce_entry.hpp
```diff
@@ -57,7 +57,7 @@ TORRENT_VERSION_NAMESPACE_2
 	struct TORRENT_EXPORT announce_infohash
 	{
 		// internal
-		TORRENT_UNEXPORT announce_infohash();
+		announce_infohash();
```
aldenml, line 60: this is needed because below in `announce_endpoint`
```
aux::array<announce_infohash, num_protocols, protocol_version> info_hashes;
```

include/libtorrent/piece_block.hpp
```diff
@@ -38,7 +38,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent {
 
-	struct TORRENT_EXTRA_EXPORT piece_block
+	struct TORRENT_EXPORT piece_block
```
aldenml, line 41: this is needed because
```
std::vector<piece_block> picker_log_alert::blocks() const;
```

include/libtorrent/socket_type.hpp
```diff
@@ -56,7 +56,7 @@ enum class socket_type_t : std::uint8_t {
 };
 
 // return a short human readable name for types of socket
-char const* socket_type_name(socket_type_t);
+TORRENT_EXPORT char const* socket_type_name(socket_type_t);
```
aldenml, line 59: well, this is a very very nice to have :)

include/libtorrent/announce_entry.hpp
arvidn: would it work to, instead, add a constructor to `announce_endpoint`, and define it as `= defailt` in announce_endpoint.cpp?

arvidn:
it would be great if you could add some code that depends on these symbols to a test, to make sure these fixes are covered by the CI as well. at least on windows tests link dynamically against the main library

include/libtorrent/socket_type.hpp
arvidn: it this should be public, maybe it should say it may return an empty string too

include/libtorrent/announce_entry.hpp
aldenml: yes, it did work, changed and squashed

include/libtorrent/socket_type.hpp
aldenml: done

aldenml:
OK, will add that code

arvidn:
since travis is broken right now, I'll cancel your builds there

arvidn:
you should be able to rebase this now

aldenml:
rebase and added test code

test/test_primitives.cpp
```diff
@@ -235,3 +235,12 @@ TORRENT_TEST(client_data_initialize)
...
 	TEST_CHECK(*v.get<int>() == 1337);
 }
+
+TORRENT_TEST(announce_endpoint_initialize)
+{
+	// announce_endpoint has an array of announce_infohash and
+	// announce_infohash has the constructor is marked TORRENT_UNEXPORT
+	// it's important that announce_endpoint provides a constructor
+	announce_endpoint ae;
+	TEST_EQUAL(ae.enabled, false);
```
arvidn, line 245: this defaults to `true`

aldenml: it defaults to `false` in my local env (macOS with clang) and for what I see, in a few other travis jobs. Do you want to make it `true` explicitly?

arvidn: oh, I was looking [here](https://github.com/arvidn/libtorrent/blob/RC_1_2/src/announce_entry.cpp#L58). but it seems the other constructors don't initialize it. I think it should be initialized in-class. where the member is declared.

arvidn: wait.. why is this default constructible? does it have to be because it's in a container?
If it doesn't have to be, I think it would be best to delete the default constructor. if it has to be, then a default constructor would have to be defined that initializes all these members. in-class initializers don't work on bitfields

arvidn: I'm also a bit confused because I was looking at `RC_1_2` instead of `RC_2_0` :)

arvidn: you know what. in `RC_2_0` I finally separated the public announce entry types from the internal ones. So there really isn't any good reason for `disabled` to be a single bit in the public struct. It could just be a `bool` and you could get a proper in-class initializer for it.

include/libtorrent/announce_entry.hpp
```diff
@@ -189,7 +191,7 @@ TORRENT_VERSION_NAMESPACE_2
 #endif
 
 		// set to false to not announce from this endpoint
-		bool enabled : 1;
+		bool enabled = true;
```
aldenml, line 194: is this what you meant?

include/libtorrent/socket_type.hpp
```diff
@@ -56,9 +56,9 @@ enum class socket_type_t : std::uint8_t {
 };
 
 // return a short human readable name for types of socket
+// TODO: move to aux
 char const* socket_type_name(socket_type_t);
```
aldenml, line 60: I decided not to export this instead (and not use it), it should be private unless really necessary

aldenml:
@arvidn maybe the deprecation of `travis-ci.org` is the reason we don't see the report in the checks, my recommendation is deleting the project from travis and rejoin with `travis-ci.com`

include/libtorrent/announce_entry.hpp
arvidn: yes

arvidn:
looks like there's an inline namespace missing: https://travis-ci.com/github/arvidn/libtorrent/jobs/359802518#L2882

aldenml:
do you have an idea where to look at?

arvidn:
make sure `TORRENT_VERSION_NAMESPACE_2` is used on the .cpp file as well (and the `_END`)

aldenml:
must be something else, still getting the same errors

arvidn:
The remaining errors are of the form:

```
undefined reference to `libtorrent::v1_2::announce_endpoint::announce_endpoint()'
```

So, this looks like there are tests relying on having access to that default constructor, and that it isn't exported (even with `EXPORT_EXTRA`)

arvidn:
I don't understand why it wouldn't be exported. Does it work locally for you, when you build with `link=shared`?

aldenml:
I finally was able to replicate it, the problem is with `deprecated-functions=off`

aldenml:
it should be fine now, rebased and squashed

Issue #4855, opened 2020-07-08
b2 builds usrsctp in debug mode even with variant=release


aldenml:
libtorrent version: `master`

I understand this is more about the way `libdatachannel` builds `usrsctp`, but it concerns the `libtorrent` build. Any hint and I can give it a try.

```
-- No build type selected, using DEBUG
```
See https://ci.appveyor.com/project/arvidn/libtorrent/branch/master/job/jt65wnmdicmmpnt6?fullLog=true#L409

CC @paullouisageneau

paullouisageneau:
I seems the `CMakeLists.txt` of `usrsctp` overrides the default build to debug. It should be solved by adding `-DCMAKE_BUILD_TYPE=Release` to the call from [libdatachannel's Jamfile](https://github.com/paullouisageneau/libdatachannel/blob/5825e44fc85056c74d85ceb93e8e354e4ac5c00c/Jamfile#L90).

arvidn:
ideally, the `build` directory used for `cmake` would also be `build-release` and `build-debug`, or something like that. I've noticed that if you build with `crypto=openssl` followed by a build of `crypto=gnutls`, some of the builds will not be rebuilt, but use the binary from the previous build. so, maybe the build directory should even include which ssl library is used. (I forget which target this happened to, it may not have been `usrsctp`)

paullouisageneau:
Indeed, cmake build directory should be different. It should be addressed by https://github.com/paullouisageneau/libdatachannel/pull/119

arvidn:
this can be closed, yes?

aldenml:
yes, thanks @paullouisageneau

PR #4867, opened 2020-07-10
Fix Jamfile in libdatachannel

paullouisageneau:
Fixes https://github.com/arvidn/libtorrent/issues/4855

The PR updates libdatachannel to build its dependencies usrsctp and libjuice with the correct variant and in different directories for different build configurations.


aldenml:
@paullouisageneau do you know how to pass `MINGW` when building with Jamfile? I'm having problems because  this line is not called
```
target_compile_definitions(usrsctp-static PUBLIC -DSCTP_STDINT_INCLUDE=<stdint.h>)
```

paullouisageneau:
@aldenml Setting `toolset=gcc-mingw` in the `b2` command should set `MINGW`.

aldenml:
I was using another toolset (btw, I'm compiling for windows with mingw-w64 under linux), I will try your suggestion and report back, thanks

arvidn:
I'm open to suggestions on how to avoid `Werror` to apply to dependencies. https://ci.appveyor.com/project/arvidn/libtorrent/builds/34029852/job/6quhfabxod3397e1#L381

paullouisageneau:
@arvidn Sadly I don't know how to do that in Jamfiles, I'm fixing the warning for now.

arvidn:
https://ci.appveyor.com/project/arvidn/libtorrent/builds/34037634/job/5ikp9etowhrfyahj#L1230

it looks like perhaps some mixup in where `libusrsctp.lib` is supposed to end up.

arvidn:
tested that this works now:

```
b2 -j30 -l250 crypto=openssl webtorrent=on variant=release,debug
```

However, this still fails:

```
b2 -j30 -l250 crypto=openssl,gnutls webtorrent=on
```

It would be nice if gnutls vs. openssl was also represented in the build directory. I haven't dug into which library is causing this, but I imagine it's one of the ones built outside of boost-build.

This is the error I get (when the initial build was with gnutls):

```
/usr/bin/ld: libtorrent.so.2.0.0: undefined reference to `nettle_hmac_sha1_set_key'
/usr/bin/ld: libtorrent.so.2.0.0: undefined reference to `nettle_hmac_sha1_digest'
/usr/bin/ld: libtorrent.so.2.0.0: undefined reference to `nettle_hmac_sha1_update'
```

paullouisageneau:
This seems to be caused by libjuice. Strangely, openssl/gnutls is now represented in the build directory, so I guess this is a problem with it not being registered as a different build by b2. I'll look into it.

PR #4869, opened 2020-07-10
fixup C-binding

arvidn:


thebluepotato:
@arvidn This is awesome! I've been able to get it to compile even for Swift use (wasn't trivial). I've added non-variadic versions of various functions with `va_list` so Swift could import them. I'll keep tinkering with it and suggest some changes in a new PR (disabling deprecated functions for example).

arvidn:
@thebluepotato patches to improve the C bindings are welcome! (against the `master` branch)

PR #4871, opened 2020-07-11
fix enumeration of network routes to support larger routing tables

arvidn:
by adapting the buffer size when necessary, rather than sticking to a fixed size buffer when reading from the netlink socket.

@ssiloti would you mind taking a look at this?

The patch is a little bit bigger than it needed to be because I wasn't excited about the way the address family was set on the netlink request messages, so I pulled some of that out of `nl_dump_request()`

### DIFF
--- a/src/enum_net.cpp
+++ b/src/enum_net.cpp
@@ -168,19 +168,25 @@ namespace {
 
 #if TORRENT_USE_NETLINK
 
-	int read_nl_sock(int sock, span<char> buf, std::uint32_t const seq, std::uint32_t const pid)
+	int read_nl_sock(int sock, std::vector<char>& buf, std::uint32_t const seq, std::uint32_t const pid)
 	{
 		nlmsghdr* nl_hdr;
 
-		int msg_len = 0;
-
+		int increment = 2048;
+		int const max_increment = 8192;
+		int const headroom = 2048;
+		int read_cursor = 0;
 		for (;;)
 		{
-			auto next_msg = buf.subspan(msg_len);
-			int const read_len = int(recv(sock, next_msg.data(), static_cast<std::size_t>(next_msg.size()), 0));
+			if (read_cursor + headroom > int(buf.size()))
+			{
+				buf.resize(buf.size() + increment);
+				if (increment < max_increment) increment *= 2;
+			}
+			int const read_len = int(recv(sock, buf.data() + read_cursor, buf.size() - read_cursor, 0));
 			if (read_len < 0) return -1;
 
-			nl_hdr = reinterpret_cast<nlmsghdr*>(next_msg.data());
+			nl_hdr = reinterpret_cast<nlmsghdr*>(buf.data() + read_cursor);
 
 #ifdef __clang__
 #pragma clang diagnostic push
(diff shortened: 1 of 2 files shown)


src/enum_net.cpp
```diff
@@ -200,41 +205,34 @@ namespace {
...
-		nl_msg->nlmsg_seq = seq;
+		request_msg->nlmsg_type = type;
+		request_msg->nlmsg_flags = NLM_F_DUMP | NLM_F_REQUEST;
+		request_msg->nlmsg_seq = seq;
 		// in theory nlmsg_pid should be set to the netlink port ID (NOT the process ID)
 		// of the sender, but the kernel ignores this field so it is typically set to
 		// zero
-		nl_msg->nlmsg_pid = 0;
-		// first byte of routing messages is always the family
-		msg[sizeof(nlmsghdr)] = family;
```
arvidn, line 224: I wanted to get rid of this line (and assumption). Instead, I pass in the netlink request message, and it's allocated on the stack by the caller. I don't think there's a good reason to reuse the buffer that we read the response into for the request.

src/enum_net.cpp
```diff
@@ -168,19 +168,24 @@ namespace {
...
+		int increment = 1024;
+		int const max_increment = 8192;
+		int read_cursor = 0;
 		for (;;)
 		{
-			auto next_msg = buf.subspan(msg_len);
-			int const read_len = int(recv(sock, next_msg.data(), static_cast<std::size_t>(next_msg.size()), 0));
+			if (read_cursor + (increment / 2) > int(buf.size()))
+			{
+				buf.resize(buf.size() + increment);
```
arvidn, line 182: this is the actual fix. instead of taking a `span` to a fixed size buffer, take a `std::vector` and resize it on demand.

src/enum_net.cpp
```diff
@@ -1185,18 +1193,26 @@ int _System __libsocket_sysctl(int* mib, u_int namelen, void *oldp, size_t *oldl
...
 		}
 
-		std::uint32_t seq = 0;
+		struct
+		{
+			struct nlmsghdr hdr;
+			struct rtmsg msg;
+		} request{};
+		request.hdr.nlmsg_len = std::uint32_t(NLMSG_LENGTH(sizeof(request.msg)));
+		request.msg.rtm_family = AF_UNSPEC;
```
arvidn, line 1202: moving the request out of the `nl_dump_request()` function also offers some additional flexibility, in case we want to specify which routing table we want to enumerate for instance, or setting other fields in `rtmsg`

src/enum_net.cpp
```diff
@@ -168,19 +168,24 @@ namespace {
...
-		int msg_len = 0;
-
+		int increment = 1024;
+		int const max_increment = 8192;
+		int read_cursor = 0;
 		for (;;)
 		{
-			auto next_msg = buf.subspan(msg_len);
-			int const read_len = int(recv(sock, next_msg.data(), static_cast<std::size_t>(next_msg.size()), 0));
+			if (read_cursor + (increment / 2) > int(buf.size()))
```
ssiloti, line 180: This assumes messages will never be larger than 512 bytes, which is probably fine in this case, but if you want to be really pedantic netlink messages can be up to 32KB on current kernels.

arvidn: good point. when I started this patch I was thinking netlink sockets were stream-like, but I realize they aren't.

arvidn: also, it seems a bit unnecessary to complicate this by making the headroom be `increment / 2`, instead of just a fixed number. The minimum number will be the bottleneck anyway

toasta:
just to mention this here as well: (from #4850)
while this does not return 0 routes if more than ~100 routes are installed (might be the 8k msgsize limit)
it seems it is getting only a tiny subset of all routes. (~65k routes set, libtorrent sees ~308 w/ test/enum_if)

does this need some sort of "pagnation" or different handling of multipart netlink-responses?

arvidn:
as ssiloti pointed out, it still assumes that a single netlink message is smaller than 1kiB (I updated it to 2 kiB). But if a single netlink message is larger than this, the function will fail.

I'm working on improving this by not allocating a buffer to fit all messages up-front, but to parse them as they're received here: https://github.com/arvidn/libtorrent/pull/4875 this also makes it easier to raise this per-message size limit. Would you mind testing that patch? (it's 2 commits, this one + one more)

toasta:
'll try. I saw you use the parse function in several places and already mentioned, that there's a bit of code duplication.

toasta:
i cherrypicked 64ab86d6de812b75300bcc503cc3ad52689ad16d and 95072f02cacb96162c3300f5490a7434ca628dd1
and this looks way better::

```
bf@bf-spielwiese:~/git/libtorrent/test$ ./enum_if | grep -B 999999 "^========= Interfaces"  | wc -l
65675
bf@bf-spielwiese:~/git/libtorrent/test$ ip r show | wc -l
65636
```

my testscript does the same...
```
[65671] route ff00:: mask ff00:: via gateway :: dev tun2 ?? ??
** GOT [65672] routes
```

gr8 work! thx!

arvidn:
thanks for testing! I will probably replace this patch with https://github.com/arvidn/libtorrent/pull/4875, but there's still some work left to do there. I'm working on getting more information about interfaces too, so I can ignore ones that are DOWN.

arvidn:
superseded by https://github.com/arvidn/libtorrent/pull/4875

PR #4879, opened 2020-07-14
export interface flags and state in enum_net_interface

arvidn:


src/enum_net.cpp
```diff
@@ -697,12 +846,39 @@ int _System __libsocket_sysctl(int* mib, u_int namelen, void *oldp, size_t *oldl
...
+					: (adapter->OperStatus == IfOperStatusNotPresent) ? if_state::notpresent
+					: (adapter->OperStatus == IfOperStatusLowerLayerDown) ? if_state::lowerlayerdown
+					: if_state::unknown;
+
+				r.flags = r.state != if_state::down ? if_flags::up : interface_flags{};
+				if (adapter->IfType == IF_TYPE_SOFTWARE_LOOPBACK)
+					r.flags |= if_flags::loopback;
+				if (adapter->IfType == IF_TYPE_PPP)
+					r.flags |= if_flags::pointopoint;
+				if (!(adapter->Flags & IP_ADAPTER_NO_MULTICAST))
```
ssiloti, line 864: This flag is actually the opposite meaning of the flag on other platform.

arvidn: oh, you mean on windows this means whether it's a multicast *recipient* whether on other platforms the multicast flag means it's multicast *capable*?

arvidn: it doesn't sound too different though. form msdn: "The adapter is not a multicast recipient."
versus the linux `IFF_MULTICAST` flag: "Supports multicast"

ssiloti: Oops my mistake I missed the exclamation point in front.

PR #4880, opened 2020-07-15
fix issue on NAT-PMP shutdown

arvidn:
where the periodic refresh timer would still run after initiating shutdown


arvidn:
I don't think this patch will suffice unfortunately.

This was based on speculation of why `m_socket` could be closed, and trying to prevent that to happen on shutdown. I would like to actually handle the error as well. And I have reasons to believe the initial patch won't prevent it from happening.

arvidn:
@fmike7 any results on testing this?

fmike7:
Have not seen any crashes, so it seems to be working. Thanks!

PR #4881, opened 2020-07-15
change has_default_route() to has_internet_route()

arvidn:
and make it not just look for a default route, but any route of a global internet IP


arvidn:
I should update documentation

PR #4885, opened 2020-07-17
moved bloom_filter.hpp and bt_peer_connection.hpp to aux

aldenml:


include/libtorrent/aux_/bloom_filter.hpp
```diff
@@ -39,7 +39,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <cmath> // for log()
 #include <cstdint>
 
-namespace libtorrent {
+namespace libtorrent::aux {
```
arvidn, line 42: nice touch. however, I believe this needs some updates to the `tools/gen_reference_doc.py` as well ([here](https://github.com/arvidn/libtorrent/blob/RC_1_2/docs/gen_reference_doc.py#L857) specifically)

arvidn: all of these changes are in files that were moved to the `aux_` directory though, so technically they are already correctly filtered by the documentation generation program

PR #4895, opened 2020-07-21
moved http_connection.hpp to aux

aldenml:


aldenml:
why do you think there are two travis checks here?

aldenml:
hi @arvidn, just to indicate that everything is green and that that second travis check in queued mode is really strange

PR #4902, opened 2020-07-24
moved ip_voter.hpp and io.hpp to aux

aldenml:


CMakeLists.txt
```diff
@@ -221,8 +219,10 @@ set(libtorrent_aux_include_files
 	invariant_check
 	instantiate_connection
 	io
+	io_bytes
```
aldenml, line 222: @arvidn naming might not be great, any suggestion?

arvidn: this seems fine

Issue #4899, opened 2020-07-24
Crash when using WebTorrent


aldenml:
libtorrent: `master`
platform/architecture: Android arm64
compiler: clang 9.0.8

When using a WebTorrent transfer, in just under a minute this crash happens
```
A/libc: FORTIFY: pthread_mutex_lock called on a destroyed mutex (0x7961d1dc20)
A/libc: Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) 
```
I compiled with all the variants trying to get the stacktrace/backtrace and nothing. I looked around and only `libjuice` seems to use pthread explicitly, and finally, compiled with no WebTorrent, and no crash, everything seems to work fine.

@arvidn @paullouisageneau, I don't mind moving the issue and discussion to `libdatachannel` if it's more appropriate, I also can troubleshoot, but I need some guidance.

arvidn:
If you run under a debugger, or enable coredumps and load it up in a debugger after the crash, you should be able to get a stack trace. You may want to build with `-fno-omit-frame-pointer`.

I'm pretty sure most implementations of `std::mutex` also use `pthread_mutex` under the hood, and this would point to a life-time issue, where a mutex is used after it's been destroyed.

paullouisageneau:
`libjuice` is actually a likely culprit because it'll fallback to mutexes if C11 atomics are not available (they might not be here), which could trigger such a bug. Did you use openssl for crypto?

aldenml:
Yes, I’m using openssl. Do you think that using nettle can improve the situation?

aldenml:
and what about nice? (but I fear the big binary size)

paullouisageneau:
I was asking just to try to narrow down the source of the issue. Did you have any luck getting a stack trace? Did you try with `ndk-stack`?

aldenml:
Nope, nothing, I will start hardcoding some (stack unwind related) flags in the inner CMakeFiles to see if that helps. Btw, I just checked and atomics seems to be present, at least `__STDC_NO_ATOMICS__` is not defined.

paullouisageneau:
A bit of a wild guess, but could you please try with this mitigation: https://github.com/paullouisageneau/libdatachannel/pull/133 ?

aldenml:
@paullouisageneau good news!! With that branch in `libdatachannel` I have no errors and no crash for a while now


[...]

PR #4907, opened 2020-07-25
Update libdatachannel to fix usrsctp callback crash

paullouisageneau:
Fixes https://github.com/arvidn/libtorrent/issues/4899


CMakeLists.txt
```diff
@@ -749,8 +749,7 @@ else()
 endif()
 
 if (webtorrent)
-	option(USE_JUICE "Use libjuice instead of libnice for libdatachannel" ON)
```
paullouisageneau, line 752: This is not needed anymore, libjuice is now the default.

Issue #4904, opened 2020-07-25
Proxy trackers, but keep UPNP and direct IP/DHT


iamtakingiteasy:
I am affected via qbittorrent issue https://github.com/qbittorrent/qBittorrent/issues/13182 which is dependent on libtorrent.

Apparently, libtorrent assumes having HTTP proxy set as grounds enough to disable UPNP, direct DHT and peer access.

Relevant changes seem to be https://github.com/arvidn/libtorrent/pull/4498/files#diff-e0ddc9bcdbc80d1a19c751515afc1ba0R287 / PR #4498

This is breaking my use-case, when local uplink/country may be blocking specific tracker (hence the need for HTTP proxy to report for tracker with pre-configured own external IP), while having no issue with the torrent traffic itself.

A way to proxy only tracker connections while retaining all direct IP/DHT/UPNP acceess is needed, to avoid need to constantly enable and disable HTTP proxy just to report to tracker while having DHT discovery ongoing.

iamtakingiteasy:
Fixed by #4904 , progress tracked in #4903

PR #4909, opened 2020-07-27
fix issue when configuring a proxy, but not proxying peer connections

arvidn:


arvidn:
this patch is believed to address https://github.com/arvidn/libtorrent/issues/4904

PR #4920, opened 2020-07-31
moved web_connection_base.hpp, web_peer_connection.hpp and http_seed_connection.hpp to aux

aldenml:


src/http_seed_connection.cpp
```diff
@@ -286,7 +286,7 @@ namespace libtorrent {
 
 					if (t->alerts().should_post<url_seed_alert>())
 					{
-						std::string const error_msg = to_string(m_parser.status_code()).data()
+						std::string const error_msg = lt::to_string(m_parser.status_code()).data()
```
arvidn, line 289: this call looks suspicious. why is `.data()` called on a temporary string there? just to then be copied again by the `std::string` constructor of `error_msg`. (I realize you didn't add this call).  but would you mind removing it? the call to `.data()`

aldenml: OK :)

PR #4928, opened 2020-08-01
Fix path duplication in .pc file installation

lgbaldoni:


arvidn:
@zeule does this look good?

zeule:
> @zeule does this look good?

Should make no difference, as far as I understand that.

Issue #4927, opened 2020-08-01
soversion goes backwards in 2.0.0RC2


lgbaldoni:
**Please provide the following information**

> libtorrent version (or branch):

2.0.0RC2

> platform/architecture:

openSUSE Tumbleweed x86_64

> compiler and compiler version:

gcc 10.1.1, cmake 3.18.0

> please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

Build with cmake creating shared libs.
libtorrent-rasterbar.so.2.0.0 will be produced, with a softlink to libtorrent-rasterbar.so.10: that's going to look like a downgrade on most linux distributions, are you sure it's the intended outcome?

Is there an API change with respect to the 1.2.x branch btw?

arvidn:
yes, not only API changes, there are ABI changes since 1.2.x. SO version need to be incremented for 2.0

arvidn:
As far as I understand, the main reason to use SOVERSION is to avoid subtle bugs because of ABI incompatibility.
It doesn't *really* matter whether it counts up or down, as long as each distinct ABI version is unique.

Would it not be possible to "start over" and just use the normal library version as the SOVERSION?

Also, the SOVERSION is really just a convention of the filename of the .so file, right? as far as I can tell, it's not encoded inside the library file itself. Is that right?

lgbaldoni:
> As far as I understand, the main reason to use SOVERSION is to avoid subtle bugs because of ABI incompatibility.
It doesn't really matter whether it counts up or down, as long as each distinct ABI version is unique.

It's true for the major version, but for minor and patch there is the assumption of non breaking changes.

> Would it not be possible to "start over" and just use the normal library version as the SOVERSION?

Since it's quite distant (2 vs. 10) there should be no version clash. Still, it's a break with tradition.

> Also, the SOVERSION is really just a convention of the filename of the .so file, right? as far as I can tell, it's not encoded inside the library file itself. Is that right?

Nope, see `objdump -p libname.so.<sover>|grep SONAME`.

arvidn:
this addresses the SONAME issue https://github.com/arvidn/libtorrent/pull/4930

PR #4929, opened 2020-08-01
Update soversion and set it correctly in cmake

lgbaldoni:
I wouldn't want to speak out of turn, but judging from the changelog for 2.0 there are breaking API changes, so a soversion bump is warranted.

This also solves https://github.com/arvidn/libtorrent/issues/4927

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,12 @@ project(libtorrent
 	DESCRIPTION "Bittorrent library"
 	VERSION ${VER_MAJOR}.${VER_MINOR}.${VER_TINY}
 )
-set (SOVERSION "10")
+set (LIBVERSION_MAJOR 11)
+set (LIBVERSION_MINOR 0)
+set (LIBVERSION_PATCH 0)
+
+set (LIBVERSION ${LIBVERSION_MAJOR}.${LIBVERSION_MINOR}.${LIBVERSION_PATCH})
+set (SOVERSION ${LIBVERSION_MAJOR})
 
 include(GNUInstallDirs)
 include(GeneratePkgConfig)
@@ -515,7 +520,7 @@ set_target_properties(torrent-rasterbar
 	PROPERTIES
 		CXX_VISIBILITY_PRESET "hidden"
 		VISIBILITY_INLINES_HIDDEN "true"
-		VERSION ${PROJECT_VERSION}
+		VERSION ${LIBVERSION}
 		SOVERSION ${SOVERSION}
 )
 
--- a/configure.ac
+++ b/configure.ac
@@ -44,7 +44,7 @@ dnl  increment AGE, Otherwise AGE is reset to 0. If CURRENT has changed,
 dnl  REVISION is set to 0, otherwise REVISION is incremented.
 dnl ---------------------------------------------------------------------------
 
-m4_define([VERSION_INFO_CURRENT],[10])
+m4_define([VERSION_INFO_CURRENT],[11])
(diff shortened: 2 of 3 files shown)


configure.ac
```diff
@@ -44,7 +44,7 @@ dnl  increment AGE, Otherwise AGE is reset to 0. If CURRENT has changed,
 dnl  REVISION is set to 0, otherwise REVISION is incremented.
 dnl ---------------------------------------------------------------------------
 
-m4_define([VERSION_INFO_CURRENT],[10])
```
arvidn, line 47: I don't know what happened here. This file does not exist in the RC_2_0 branch

CMakeLists.txt
```diff
@@ -515,7 +520,7 @@ set_target_properties(torrent-rasterbar
 	PROPERTIES
 		CXX_VISIBILITY_PRESET "hidden"
 		VISIBILITY_INLINES_HIDDEN "true"
-		VERSION ${PROJECT_VERSION}
+		VERSION ${LIBVERSION}
```
arvidn, line 523: can you explain this? Your title mentions "soversion", but it remain unchanged on the line below. Instead you change the "version".

CMakeLists.txt
```diff
@@ -9,7 +9,12 @@ project(libtorrent
 	DESCRIPTION "Bittorrent library"
 	VERSION ${VER_MAJOR}.${VER_MINOR}.${VER_TINY}
 )
-set (SOVERSION "10")
+set (LIBVERSION_MAJOR 11)
+set (LIBVERSION_MINOR 0)
+set (LIBVERSION_PATCH 0)
```
arvidn, line 14: This seems it would introduce a lot of confusion. The version is 2.0.0. the SOVERISON may be something else, because that's there for ABI compatibility, but elevating it up to these names seem it would cause confusion

arvidn:
the current state of libtorrent 2.0 (i.e. the `RC_2_0` branch) both the cmake file and the Jamfile will produce a library with the `.2.0.0' filename extension. If there's a good reason to change this to be just `.11`, it should be changed in both places. It should also be clear that the SOVERSION is separate from VERSION (the latter being 2.0.0 right now).

Since I only break ABI on updates to major and minor version updates, the patch-version should probably not be included in the SOVERSION actually. So that should be fixed at the very least.

CMakeLists.txt
lgbaldoni: I changed it in the variable on [top](https://github.com/arvidn/libtorrent/pull/4929/commits/b834883813fbd088073c3c978fc3a88a5d78dcf8#diff-af3b638bc2a3e6c650974192a53c7291R17).

Sorry about the rest, I've made a mess.
Will make a new PR when ready.

CMakeLists.txt
lgbaldoni: Having the SOVERSION at 10 and the VERSION at 2.0.0 will be very confusing on linux.
If you check any other library, you'll see libfoo.1.2.3 linked to libfoo.so.1 linked to libfoo.so.

Instead, with things as they are now, libtorrent-rasterbar.so.10 links to libtorrent-rasterbar.so.2.0.0 which links to libtorrent-rasterbar.so, which can't really be used in a package.

I had listed major/minor/patch for clarity, but if you decide to use VERSION, major can be the SOVERSION and those other variables will be supplied directly by cmake (CMAKE_PROJECT_VERSION_MAJOR and so on).

arvidn:
my understanding is that this PR is no longer relevant

lgbaldoni:
I would say the same.

PR #4930, opened 2020-08-01
change soname to use major.minor library version

arvidn:


lgbaldoni:
Sorry, but this is even worse than before :disappointed:. 

```
libtorrent-rasterbar.so -> libtorrent-rasterbar.so.11
libtorrent-rasterbar.so.11 -> libtorrent-rasterbar.so.2.0.0
libtorrent-rasterbar.so.2.0.0

> objdump -p libtorrent-rasterbar.so.2.0.0 |grep SONAME
  SONAME               libtorrent-rasterbar.so.11
```
In my opinion, the library ought to have a full version, with its major field coinciding with the soversion. Since soversion 2 was last used about 12 years ago, a count reset would be preferable to this.

Otherwise you could keep increasing the historical soversion and maybe take minor and patch from version if it's not confusing (it would still have to reflect the breaking changes / nonbreaking changes / fix convention).

Also see this [guideline](https://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html#AEN49).

arvidn:
in libtorrent I have broken ABI on minor version updates as well, so the SOVERSION would have to be 2.0 I think.

But I'm a little bit confused. I agree that it would be nice to have the SOVERSION the same as the library version. The original issue you raised, though, was that the SOVERSION was *unchanged* compared to 1.2.x. That's clearly bad. I don't see how this can worse. This patch changes the SOVERSION, and it follows the convention of incrementing it by one. So, granted, it's not as nice as using the same version, I would just classify it as the second best solution.

> a count reset would be preferable to this.

I'm curious. by "this", do you refer to using an integer version number for the ABI that's different from the library version number? or are you specifically referring to the fact that `libtorrent.so` traverses two symlinks? and that the concrete file still use the library version number?

None of those things are any different than 1.2.x though, are they? I mean, isn't it just as bad in the current release?

lgbaldoni:
> I'm curious. by "this", do you refer to using an integer version number for the ABI that's different from the library version number? or are you specifically referring to the fact that libtorrent.so traverses two symlinks? and that the concrete file still use the library version number?

I'm referring to the soversion/version commingling which, if left like this, will have to be patched out in every linux distribution.

> None of those things are any different than 1.2.x though, are they? I mean, isn't it just as bad in the current release?

On the other hand, in 1.2.7 versioning is consistent:
```
libtorrent-rasterbar.so -> libtorrent-rasterbar.so.10.0.0
libtorrent-rasterbar.so.10 -> libtorrent-rasterbar.so.10.0.0
libtorrent-rasterbar.so.10.0.0

$> objdump -p libtorrent-rasterbar.so.10.0.0 |grep SONAME
  SONAME               libtorrent-rasterbar.so.10
```

arvidn:
I see. thanks for explaining. I'll update this PR to got to soname `libtorrent.so.2.0` and going forward using that versioning scheme. I'm not super familiar with the cmake side of things, I'll try to fix that too

arvidn:
another benefit (apart from correctly distinguishing ABI incompatible versions) is that the the soname still remains unique, as I'm pretty sure the old `2` soname did not have the `2.0` name, but just `2`. So there's no conflict with the old version.

FranciscoPombal:
@arvidn In what concerns the CMake side of things, you might be interested in this talk by Craig Scott, https://youtu.be/m0DwB4OvDXk?t=921, specifically the part from 15:21 through 22:25 (the link already contains the correct start timestamp).

arvidn:
as far as I can tell, this patch honours all the conventions craig brings up.

arvidn:
cmake produce this:

```
$ ls -la libtorrent-rasterbar*
lrwxrwxrwx  1 arvid arvid      27 aug  5 11:38 libtorrent-rasterbar.so -> libtorrent-rasterbar.so.2.0
lrwxrwxrwx  1 arvid arvid      29 aug  5 11:38 libtorrent-rasterbar.so.2.0 -> libtorrent-rasterbar.so.2.0.0
-rwxrwxr-x  1 arvid arvid 7875768 aug  5 11:38 libtorrent-rasterbar.so.2.0.0
```

which looks exactly right to me.

PR #4933, opened 2020-08-02
the concrete values of stats counters are internal

arvidn:
and must be queried at runtime by clients

### DIFF
--- a/include/libtorrent/performance_counters.hpp
+++ b/include/libtorrent/performance_counters.hpp
@@ -46,6 +46,7 @@ namespace libtorrent {
 
 	struct TORRENT_EXPORT counters
 	{
+		// internal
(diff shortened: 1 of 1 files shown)


aldenml:
questions: what does "internal" mean in terms of ABI and API? aren't these values needed to index?

arvidn:
it means they are not part of the ABI nor the API. this whole class, `counters` was internal in 1.2.x. The only reason this was mad public was because `disk_interface` is now a new customization point, which needs to take a reference to these.

The concrete index values have never been guaranteed to be stable, but have to be queries at runtime (I think it's `stats_metrics()` and `find_stats_metric()`, or something like that).

aldenml:
right, I will refresh my memory looking at the code

PR #4935, opened 2020-08-02
fixing warnings of tests and logging=off

aldenml:


aldenml:
@arvidn, I discovered this while creating a travis with `arch=s390x toolset=gcc-s390x`, are you OK with all these in `master` and backport the fixes?

PR #4937, opened 2020-08-03
setup of travis tests for s390x architecture

aldenml:


.travis.yml
```diff
@@ -215,14 +231,23 @@ install:
...
     fi'
 
+  - 'if [[ "$arch" == "s390x" ]]; then
+      cd $TRAVIS_BUILD_DIR/test;
+      s390x-linux-gnu-g++-8 --version;
+      wget --no-verbose -O ubuntu-base.tar.gz http://cdimage.ubuntu.com/ubuntu-base/releases/19.10/release/ubuntu-base-19.10-base-s390x.tar.gz;
+      sudo mkdir rootfs;
+      sudo tar xf ubuntu-base.tar.gz -C rootfs;
+      sudo cp /usr/bin/qemu-s390x-static rootfs/usr/bin/;
+      sudo chroot rootfs mount -t proc none /proc;
```
arvidn, line 248: shouldn't all these commands be separated by `&&` instead of `;`? if one fails, will the error propagate?

.travis.yml
```diff
@@ -373,3 +398,13 @@ script:
       sudo cp test-launcher.sh rootfs;
       travis_retry ${B2} arm-tests warnings-as-errors=on warnings=all crypto=$crypto dht=off extensions=off logging=off deprecated-functions=off invariant-checks=off export-extra=on asserts=on $toolset target-os=linux link=static testing.launcher="sudo chroot rootfs /bin/sh test-launcher.sh";
     fi'
+  - 'if [[ "$arch" == "s390x" ]];
+    then
+      ${B2} s390x-tests warnings-as-errors=on warnings=all crypto=$crypto dht=on extensions=off logging=off deprecated-functions=off invariant-checks=off export-extra=on asserts=on $toolset target-os=linux link=static testing.execute=off;
```
arvidn, line 403: I can't remember, but is there a good reason to disable invariant-checks logging and extensions? Is the idea that the emulation will be so slow that we want to do as little as possible?

arvidn:
when I try to do this myself, on ubuntu, I just get this when I run the test:
```
terminate called:
<not supported>
```

any idea what I'm doing wrong?

.travis.yml
aldenml: it's probably a good idea, but if any of them fails, everything else will fail

.travis.yml
aldenml: yes, the idea is to do as little as possible and I doubt any "live network" test will work reliable, however that was the case for arm and I just copy-pasted :)

aldenml:
no idea about the `terminate called:` message

PR #4938, opened 2020-08-03
moved torrent_peer.hpp and torrent_peer_allocator.hpp to aux

aldenml:


include/libtorrent/aux_/torrent_peer_allocator.hpp
```diff
@@ -84,13 +84,13 @@ namespace libtorrent {
...
-		boost::pool<> m_ipv6_peer_pool{sizeof(libtorrent::ipv6_peer), 500};
+		boost::pool<> m_ipv4_peer_pool{sizeof(libtorrent::aux::ipv4_peer), 500};
+		boost::pool<> m_ipv6_peer_pool{sizeof(libtorrent::aux::ipv6_peer), 500};
 #if TORRENT_USE_I2P
-		boost::pool<> m_i2p_peer_pool{sizeof(libtorrent::i2p_peer), 500};
+		boost::pool<> m_i2p_peer_pool{sizeof(libtorrent::aux::i2p_peer), 500};
 #endif
 #if TORRENT_USE_RTC
-		boost::pool<> m_rtc_peer_pool{sizeof(libtorrent::rtc_peer), 500};
+		boost::pool<> m_rtc_peer_pool{sizeof(libtorrent::aux::rtc_peer), 500};
```
arvidn, line 93: couldn't you just remove the `libtorrent::` qualifier instead? since they are in the same namespace now

src/torrent_peer_allocator.cpp
```diff
@@ -129,8 +129,8 @@ namespace libtorrent {
             return;
         }
 #endif
-		TORRENT_ASSERT(m_ipv4_peer_pool.is_from(static_cast<libtorrent::ipv4_peer*>(p)));
-		static_cast<libtorrent::ipv4_peer*>(p)->~ipv4_peer();
+		TORRENT_ASSERT(m_ipv4_peer_pool.is_from(static_cast<libtorrent::aux::ipv4_peer*>(p)));
+		static_cast<libtorrent::aux::ipv4_peer*>(p)->~ipv4_peer();
```
arvidn, line 133: same here, I would expect you could remove `libtorrent::` instead

aldenml: I will do it

arvidn:
would you mind squashing these? (and maybe rebase on top of master now that I have merged a bunch of stuff from `RC_2_0`)

aldenml:
rebased and squashed :)

PR #4939, opened 2020-08-03
Have soversion match version major

lgbaldoni:


arvidn:
Surely it's important for soname to be different for versions of the library that are not ABI compatible. libtorrent's ABI changes with minor versions, so it I would expect the minor version would have to be part of the soname as well.

PR #4940, opened 2020-08-03
fix tests on big-endian

arvidn:


test/test_crc32.cpp
```diff
@@ -46,24 +45,24 @@ TORRENT_TEST(crc32)
 	std::uint32_t in1 = 0x5aa5feef;
 	out = aux::crc32c_32(in1);
 
-	TEST_EQUAL(out, aux::host_to_network(0xd5b9e35eU));
+	TEST_EQUAL(out, 0x5ee3b9d5);
```
aldenml, line 48: @arvidn, I hope you don't mind, can you elaborate a little bit here the reason? (it's just me learning)

arvidn: yeah, I'm pretty sure this was a mistake, to expect the result from the CRC32 function to be different depending on endianness. In the two places where this is used, it's used as an integer, not a sequence of bytes. (it's used in the DHT node ID generation and validation and the peer connection affinity).

As such, the result must be the same *integer*, the native endian of integers is not relevant.

aldenml: I see, thanks

arvidn:
@aldenml I'm not entirely convinced that this is working.

```
lscpu: failed to determine number of CPUs: /sys/devices/system/cpu/possible: No such file or directory
```

https://travis-ci.com/github/arvidn/libtorrent/builds/178277263#L618

but it also seems suspicious that these tests would pass here. I haven't attempted to fix them (just one of them, which was a simpler error in the test itself).

aldenml:
maybe `lscpu` is challenging to run with `qemu-s390x-static`, maybe something similar to `sudo chroot rootfs mount -t proc none /proc` is necessary for it.

As for the lack of errors, do you think that is possible that the cross-compilation toolchain is not setting the endianess correctly?

aldenml:
Seems to work in https://github.com/arvidn/libtorrent/pull/4941, but very "emulated" :)

Screen Shot 2020-08-03 at 5 37 10 PM

arvidn:
the fact that all the tests pass on s390x is really suspicious. I think I understand the CRC32 endian issue now though (but I find it hard to trust the CI results, especially if they are passing).

aldenml:
did you manage to get the tests to fail?

arvidn:
yes, the `#error` fired..

arvidn:
I think this is a step in the right direction. I would like to back-port this to RC_2_0 and then work on resolving the remaining issues there

PR #4948, opened 2020-08-05
Fix checking

arvidn:


arvidn:
this needs more tests to cover the failure

xavier2k6:
@arvidn do you need me to re-test with commit `a12212d` or is this close to being merged?

arvidn:
@xavier2k6 I'm only working on the tests, so I'm confident the fix is unchanged from when you tested it.

xavier2k6:
ok, does this currently include all other commits from RC_1_2?

arvidn:
yes

src/storage_utils.cpp
```diff
@@ -543,11 +543,19 @@ std::int64_t get_filesize(stat_cache& stat, file_index_t const file_index
 			{
 				if (fs.pad_file_at(file_index)) continue;
 
+				// files with priority zero may not have been saved to disk at their
+				// expected location, but is likely to be in a partfile. Just exempt it
+				// from checking
+				if (file_index < file_priority.end_index()
+					&& file_priority[file_index] == dont_download
```
glassez, line 550: I wonder what happens if user deselects some files (i.e. set download priority to 0) and chooses "skip hash checking" (i.e. enables "seed mode")? Can we consider this use case as valid? What behavior user expects to see?

xavier2k6: I just tried adding a torrent which contained 43 files.

1 movie file
1 .txt file
1 .exe file
40 .srt files

Deselected 38 .srt files/1 .exe file & 1 .txt file & checked `skip hash checking`

logs immediately showed `07/08/2020 12:24 - File sizes mismatch for torrent 'ABCD1234', pausing it.`

qBittorrent GUI status shows `missing files`

skip hash check missing files

arvidn:
Yeah, skip hash checking is called seed-mode in lt, and it's not compatible with file priorities. Seed mode will be disabled.

src/storage_utils.cpp
FranciscoPombal: @glassez In your scenario, does the user already have the selected files on their system or not?

If not, I would expect @xavier2k6's result. If they do, then I would not expect that; instead, I would expect the torrent to show up as completed without errors.

glassez: I myself never use this feature, but I assume that all the user wants from it is to add some torrent for seeding using existing files, but wants to skip the hash check (due to big file size, for example). It may never download all the garbage files contained in many torrents, so it may not have all the torrent files (as in @xavier2k6 example), but it still want to add corresponding torrent for seeding only existing part of files and skip hash checking.

FranciscoPombal: @glassez

>  but I assume that all the user wants from it is to add some torrent for seeding using existing files, but wants to skip the hash check (due to big file size, for example). 

Yes, I also believe this is the main use case for this feature. I would expect it to work (i.e., going straight to "Seeding") both when all files are selected and when only some files are selected, as long as said files exist on the system of course.

I think in https://github.com/arvidn/libtorrent/pull/4948#discussion_r466989767 @xavier2k6 did not have any of the files (so immediately transitioning to the "missing files" state is the expected result). @xavier2k6 can you please confirm this was the case?

xavier2k6: >but I assume that all the user wants from it is to add some torrent for seeding using existing files, but wants to skip the hash check (due to big file size, for example).

>Yes, I also believe this is the main use case for this feature. I would expect it to work (i.e., going straight to "Seeding") both when all files are selected and when only some files are selected, as long as said files exist on the system of course.

That is what most users would like to do alright, but..........checking this now shows missing files when adding a previously completed torrent via deselection of unwanted files/enabling `skip hash check` in GUI status/`File sizes mismatch for torrent 'ABCD1234', pausing it`

HOWEVER! - there seems to be slight **"DATA LOSS"**

I used the same torrent for testing purposes as previously. (43 files)

The torrent is no longer @ 100% (It's 99.8% in transfer list window but in contents view "movie file" shows at 99.9%)

complete torrent added


The previously selected .srt files x2 are 100% but the main "movie file" is showing now as having 84.6 KiB remaining.

contents tab

Pieces missing?

Test Info



@FranciscoPombal 
>I think in #4948 (comment) @xavier2k6 did not have any of the files (so immediately transitioning to the "missing files" state is the expected result). @xavier2k6 can you please confirm this was the case?

Yes, this was the case - confirm.

Issue #3622, opened 2019-02-11
File pre-allocation on Windows without admin privileges (SetFileValidData)


libTorrentUser:
There is just one problem with the current SetFileValidData solution. As you already know, the user may not have the required privileges for the call. When that happens the call will fail (as the code already expects) but nothing else is done.

The result, on Windows, is a useless sparse file. The user will think libTorrent is pre-allocating the file because he asked it do it, a not so tech savvy user will think the file was pre-allocated because windows explorer will "lie" and show the make believe size and the original problem pre-allocation was supposed to solve (disk fragmentation) will still happen.

I suggest falling back to the classic SetFilePointer,SetEndOfFile combo when that happens (when the user has no permission to call SetFileValidData). Yes, it will go back to the write zeros situation, but at least it will be pre-allocating instead of lying to the users.

arvidn:
Are sparse files on windows useless? (excuse my ignorance, I haven't used windows in a long time).

I would expect non tech savvy users to just suffer through windows being slow, not attributing it to files not being pre-allocated.

I imagine you're saying this is a problem on NTFS, right? FAT32 will just not do sparse files at all, but I suppose it could still be somewhat fragmented until a late piece arrives and the bulk of the file is allocated.

Which branch are you referring to? `RC_1_1`, `RC_1_2` or `master`?

afaict, all three branches do `SetFilePointerEx()`, `SetEndOfFile()` and then the (best effort) `SetFileValidData()`. What are you suggesting that's different from this?

My understanding is that the file won't be made sparse if the storage mode is set to `allocate`, isn't it enough to `SetEndOfFile` on a non-sparse file to get it to be pre-allocated?

libTorrentUser:
The problem is happening on 1.1.11, the one supposedly used by qBittorrent 4.1.5. So I would say RC_1_1. All this running on Windows 64 on what I believe is a regular `guest` account.

If libTorrent already does what you said, which is pretty much what I wanted it to do, then either something is failing when the user does not have elevated privileges, or sparse files really messes things up. Or, of course, the qBittorrent guys did something weird in their build. Because looking at their code they are definitely setting the storage mode to `allocate` and it is resulting in a sparse file.

I didn't notice the problem on 1.2, but then again, I only used that one while doing some tests using Visual Studio, meaning I was on an elevated account. As soon as I get home I will try to check and see if I can reproduce it using 1.2.

[...]

Issue #2372, opened 2017-09-26
Possible solution on fast preallocation on Windows?


sledgehammer999:
I had one user propose to use the [SetFileValidData](https://msdn.microsoft.com/library/aa365544.aspx) WINAPI to facilitate fast preallocation.
Basically he wants fast start on big files **and** avoidance of file fragmentation.
He writes that `aria2` uses this trick.

I am in no position to evaluate any of this. I just forward the request here. You decide what happens.
PS: He also mentions that the same trick can be done on linux via falloc.

sledgehammer999:
tagging the user: @pasha-zzz

sledgehammer999:
Also from a bit of searching on aria2 -> Possible security issue: https://github.com/aria2/aria2/issues/583

pasha-zzz:
About this security issue: user must have SE_MANAGE_VOLUME_NAME pivilege.

ssiloti:
libtorrent already tries to call `SetFileValidData`, but by default regular Windows users do not have the `SE_MANAGE_VOLUME_NAME` privilege so the call fails most of the time.

pasha-zzz:
User is in Administrators froup, "Perform volume maintenance tasks" in secpol.msc enabled for Administrators. But qBt allocates space slow.

pasha-zzz:
This code creates file of given size using SetFileValidData.
https://github.com/mzhou/setfilevaliddata/blob/master/setfilevaliddata/main.c

Required calls before calling this function are SetFilePointer or SetFilePointerEx and SetEndOfFile

ssiloti:
Did you try running that program? If you try to run it as a regular user The call to `SetFileValidData` fails with error code 1314 `A required privilege is not held by the client`.

While regular users _can_ launch processes with administrator privileges, that is not the default. A processes can request privilege elevation, which will cause a UAC prompt to appear to the user. If qBittorrent were so inclined they could make such a request, but I imaging most users do not want to get a UAC prompt when they launch the app.

pasha-zzz:
Run this program as admin - it rapidly create file 100 GB (trash content inside, no zeroes). Running qBT as admin - nothing changed.

ajax16384:
By default libtorrent (master) uses "storage_mode_sparse" flag, to use preallocate at start with possibility to use SetFileValidData technique "storage_mode_allocate" should be passed to "add_torrent" params.
As far as i understand qBT also disables preallocate by default:
````c++
[...]

PR #4958, opened 2020-08-09
move support for SetFileValidData() to the mmap file (where it matter…

arvidn:
…s now, on windows)

### DIFF
--- a/src/mmap.cpp
+++ b/src/mmap.cpp
@@ -46,6 +46,10 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include <cstdint>
 
+#ifdef TORRENT_WINDOWS
+#include "libtorrent/aux_/win_util.hpp"
+#endif
+
 #if TORRENT_HAVE_MMAP
 #include <sys/mman.h> // for mmap
 #include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


src/mmap.cpp
```diff
@@ -117,6 +177,29 @@ file_handle::file_handle(string_view name, std::int64_t
...
+		if (!(mode & aux::open_mode::sparse))
+		{
+			// Enable privilege required by SetFileValidData()
+			// https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-setfilevaliddata
+			std::call_once(g_once_flag, acquire_manage_volume_privs);
+
+			// if the user has permissions, avoid filling
+			// the file with zeroes, but just fill it with
+			// garbage instead
+			SetFileValidData(m_fd, size);
```
arvidn, line 200: @FranciscoPombal @ssiloti
do you think it's worth preserving this feature? I believe it only works if you run as admin, in which case it can significantly speed up allocating storage on disk, since it will skip zeroing out data (and instead leave old garbage in the new file, leading to a security risk).

src/mmap.cpp
```diff
@@ -117,6 +177,29 @@ file_handle::file_handle(string_view name, std::int64_t
...
 	}
+
+	if (mode & open_mode::truncate)
+	{
+		LARGE_INTEGER sz;
+		sz.QuadPart = size;
+		if (SetFilePointerEx(m_fd, sz, nullptr, FILE_BEGIN) == FALSE)
+			throw_ex<system_error>(error_code(GetLastError(), system_category()));
+
+		if (::SetEndOfFile(m_fd) == FALSE)
```
arvidn, line 188: calling `SetEndOfFile()` is redundant, as creating the file mapping will implicitly resize the file as well. This came along with the `SetFileValidData()` patch below. I think in order for that to work, the file needs to be extended first.

src/mmap.cpp
FranciscoPombal: @arvidn If this is about overwriting data of a deleted file:

I think the expectation of "_securely_ erasing files/data" is out-of-scope for libtorrent, so it is reasonable to keep this as a potential performance optimization.

arvidn: well, it's about how someone running a libtorrent process as admin may bring files into existence that have previously-deleted (and possibly sensitive) data already in them. To users not familiar with this, it can be quite surprising. But then again, these days, having a user run as admin would *also* be quite surprising.

Early versions of the Steam client used to create its cache files like this (back when everyone was admin on Win XP) and there was quite a stir in the forums with users thinking that the client was spying on them and storing unrelated personal data in its cache file.

FranciscoPombal: @arvidn ah, I see.

> But then again, these days, having a user run as admin would also be quite surprising.

:+1:

If someone is running software that uses libtorrent as admin, they either trust it, and realize this isn't a problem, or they don't, but all bets are off anyway since they are running as admin.

arvidn:
relevant previous issues:
https://github.com/arvidn/libtorrent/issues/2372
https://github.com/arvidn/libtorrent/issues/3622

So there are *some* people that care about this.

PR #4960, opened 2020-08-09
run all tests on windows in CI

arvidn:


arvidn:
mingw, again, is the straggler...

PR #4962, opened 2020-08-10
always check for duplicate peer IDs

ssiloti:
With multi-homed clients it is possible to have multiple connections to the
same peer with different IP addresses. This is becoming increasingly common
with the spread of dual-stack hosts.

### DIFF
--- a/src/bt_peer_connection.cpp
+++ b/src/bt_peer_connection.cpp
@@ -3188,29 +3188,26 @@ namespace {
 			peer_id pid;
 			std::copy(recv_buffer.begin(), recv_buffer.begin() + 20, pid.data());
 
-			if (t->settings().get_bool(settings_pack::allow_multiple_connections_per_ip))
+			// now, let's see if this connection should be closed
+			peer_connection* p = t->find_peer(pid);
+			if (p)
 			{
-				// now, let's see if this connection should be closed
-				peer_connection* p = t->find_peer(pid);
-				if (p)
+				TORRENT_ASSERT(p->pid() == pid);
+				// we found another connection with the same peer-id
+				// which connection should be closed in order to be
+				// sure that the other end closes the same connection?
+				// the peer with greatest peer-id is the one allowed to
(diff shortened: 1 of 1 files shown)


arvidn:
is it a problem to have multiple connections to the same peer, over separate networks? even if it's the same physical network and just IPv4 and IPv6, it's bounded by 2 connections.

src/bt_peer_connection.cpp
```diff
@@ -3188,29 +3188,26 @@ namespace {
 			peer_id pid;
 			std::copy(recv_buffer.begin(), recv_buffer.begin() + 20, pid.data());
 
-			if (t->settings().get_bool(settings_pack::allow_multiple_connections_per_ip))
-			{
-				// now, let's see if this connection should be closed
-				peer_connection* p = t->find_peer(pid);
-				if (p)
+			// now, let's see if this connection should be closed
+			peer_connection* p = t->find_peer(pid);
```
arvidn, line 3192: this is a bit problematic. This call is actually quite expensive, it's a linear search.
In fact, I'm under the impression that enabling this option (`allow_multiple_connections_per_ip`) doesn't work very well

ssiloti:
Having multiple connections doesn't break anything, it just multiplies the protocol overhead.

src/bt_peer_connection.cpp
ssiloti: I'm skeptical that the linear search is a significant performance problem. The number of connected peers in a torrent is typically less than 150. With an n that small even linear search isn't too expensive, especially for a relatively infrequent event like a new peer connection.

arvidn: that's reasonable, yes. Although, I sometimes have 1000-2000 peers, on large swarm and beefy machines and connections.

What makes me a bit uneasy is that people have been reporting issues with `allow_multiple_connections_per_ip`, and I'm worried this will introduce some of those problems when not enabling that setting as well. But I suppose this is only part of that feature.

PR #4966, opened 2020-08-10
Back port remove ConvertUTP.cpp

arvidn:


arvidn:
@FranciscoPombal would you be able to test this? it's a pretty large change, so ideally it would be well tested, both on windows, linux and OSX

FranciscoPombal:
@arvidn what kind of testing do you have in mind? Just building qBittorrent, or something else? I don't have access to a machine with macOS (only via GitHub Actions).

arvidn:
torrents with filenames with interesting characters in them. The UTF-8 -> native encoding functions have been touched. The validation of UTF-8 has also been touched. All semantics is supposed to be the same though (with possible exceptions for edge-cases that windows' built-in conversion functions treat differently than the previous)

FranciscoPombal:
@arvidn 

> torrents with filenames with interesting characters in them.

I'm assuming there's no need to test with known-invalid byte sequences right?

Any particular combinations you have in mind? It would be nice to have some sort of list or known "dataset", rather than just trying some emojis/chinese characters/etc at random.

I'm thinking of using this: https://www.w3.org/2001/06/utf-8-test/UTF-8-demo.html
Is it enough?

arvidn:
the conversion functions *do* have unit tests and fuzzers. However, there are no high-level tests for full torrents to make sure whatever ends up on the filesystem is reasonable.

FranciscoPombal:
@arvidn sorry for not following through with this, I won't be able to dig into this in the near future. But at least so far there have been no complaints :)

Issue #4965, opened 2020-08-10
Linking failure on macOS 10.15 (Xcode 11.6)


FranciscoPombal:
**libtorrent version (or branch)**

1.2.8

**platform/architecture**

macOS 10.15 x86_64

**compiler and compiler version**

Xcode 11.6, AppleClang 11.0.3.11030032

**please describe what symptom you see, what you would expect to see instead and
how to reproduce it**

Trying to build qBittorrent on macOS 10.15 (GitHub Actions). The build fails in the linking step with the following message:

```
[357/357] Linking CXX executable qbittorrent.app/Contents/MacOS/qbittorrent
FAILED: qbittorrent.app/Contents/MacOS/qbittorrent 
[...]

PR #4968, opened 2020-08-10
CMake: use target_compile_features() to automatically enforce minimum compiler version and C++ standard

FranciscoPombal:
Just a simple:

```cmake
target_compile_features(torrent-rasterbar
    PUBLIC
        cxx_std_11
)
```

provides equivalent functionality to the current mechanism, and would additionally yield the advantages listed in  https://github.com/arvidn/libtorrent/issues/4965#issuecomment-671322398:

> The benefit of using `target_compile_features()` this way is that all the important logic gets taken care of automagically, with the correct behavior in all cases:
>
> -   If the compiler is capable of supporting the minimum required standard, but defaults to an insufficient standard, CMake will append the appropriate `-std` flag to override the default.
> -    If the compiler defaults to a sufficiently recent standard, the compiler default is used.
> -    Even if the compiler default is sufficient, the user can easily override the compiler default (e.g., to select something more recent) by passing `-DCMAKE_CXX_STANDARD=FOO`; this is why this variable should be left alone in the script - so that it can have the correct behavior as a cache variable.

Plus the following point which I previously omitted by mistake:

- The correct minimum `-std` flag is used even if the user mistakenly passes an insufficient value via `-DCMAKE_CXX_STANDARD` at configure-time.

---

Having the other additional features increases resilience in cases where compilers advertise support for a standard, but not _full_ support:

> Experience with qBittorrent shows that only requiring `cxx_std_14`/`cxx_std_11`/etc is not enough to enforce a "true minimum standard", because, for example, some compilers (including some versions of GCC) will pass the `cxx_std_14` test while having incomplete C++14 support. To fix this, one must look at the language support status https://gcc.gnu.org/projects/cxx-status.html, compare with the CMake page, and add the strings corresponding to the appropriate proposal numbers.

This is the reason why this PR is in draft mode - is just using `cxx_std_11` preferable for this project?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -502,7 +502,49 @@ add_library(torrent-rasterbar
 	${libtorrent_aux_include_files}
 )
 
-select_cxx_standard(torrent-rasterbar 11)
+# C++ 11 support is required
+target_compile_features(torrent-rasterbar
+	PUBLIC
+		cxx_std_11
+		cxx_rvalue_references
+		cxx_reference_qualified_functions
+		cxx_nonstatic_member_init
+		cxx_variadic_templates
+		cxx_generalized_initializers
+		cxx_static_assert
+		cxx_auto_type
+		cxx_trailing_return_types
+		cxx_lambdas
+		cxx_decltype
+		cxx_right_angle_brackets
+		cxx_default_function_template_args
+		cxx_alias_templates
+		cxx_extern_templates
+		cxx_nullptr
+		cxx_strong_enums
+		cxx_enum_forward_declarations
+		cxx_attributes
+		cxx_constexpr
+		cxx_alignas
+		cxx_alignof
+		cxx_delegating_constructors
+		cxx_inheriting_constructors
+		cxx_explicit_conversions
+		cxx_raw_string_literals
+		cxx_user_literals
+		cxx_defaulted_functions
+		cxx_deleted_functions
+		cxx_extended_friend_declarations
+		cxx_sizeof_member
+		cxx_inline_namespaces
+		cxx_unrestricted_unions
+		cxx_local_type_template_args
+		cxx_range_for
+		cxx_final
+		cxx_override
+		cxx_noexcept
+		cxx_defaulted_move_initializers
+)
 
 if (BUILD_SHARED_LIBS)
 	target_compile_definitions(torrent-rasterbar
--- a/cmake/Modules/LibtorrentMacros.cmake
+++ b/cmake/Modules/LibtorrentMacros.cmake
@@ -54,47 +54,6 @@ macro(find_public_dependency _name)
 	endif()
 endmacro()
 
-function(_cxx_standard_to_year _yearVar _std)
-	if (${_std} GREATER 97)
-		math(EXPR _year "1900 + ${_std}")
-	else()
-		math(EXPR _year "2000 + ${_std}")
-	endif()
-	set(${_yearVar} ${_year} PARENT_SCOPE)
-endfunction()
-
-function(select_cxx_standard _target _minimal_supported_version)
-	message(STATUS "Compiler default is C++${CMAKE_CXX_STANDARD_DEFAULT}")
-	# make the CXX_STANDARD property public to ensure it makes it into the pkg-config file
-	get_target_property(_std ${_target} CXX_STANDARD)
-	# ${CMAKE_CXX_STANDARD_DEFAULT} could be 98, which is lower version than C++11. Thus we convert std
-	# version to year value and compare years
-	_cxx_standard_to_year(minimal_supported_version_year ${_minimal_supported_version})
-	# if it is unset, select the default if it is sufficient or the ${_minimal_supported_version}
-	if (NOT ${_std})
-		_cxx_standard_to_year(std_default_year ${CMAKE_CXX_STANDARD_DEFAULT})
-		if (${std_default_year} GREATER_EQUAL ${minimal_supported_version_year})
-			set(_std ${CMAKE_CXX_STANDARD_DEFAULT})
-		else()
-			set(_std ${_minimal_supported_version})
-		endif()
-	else()
-		_cxx_standard_to_year(std_year ${_std})
-		if (${std_year} LESS ${minimal_supported_version_year})
-			message(FATAL_ERROR "The minimal supported C++ standard version is C++${_minimal_supported_version}")
-		endif()
-	endif()
-
-	if (cxx_std_${_std} IN_LIST CMAKE_CXX_COMPILE_FEATURES)
-		# force this standard
-		target_compile_features(${_target} PUBLIC cxx_std_${_std})
-	else()
-		message(FATAL_ERROR "Requested C++ standard version (${_std}) is not supported by the compiler")
-	endif()
-
-	message(STATUS "Building in C++${_std} mode")
-endfunction()
-
 # function for parsing version variables that are set in version.hpp file
 # the version identifiers there are defined as follows:
 # #define LIBTORRENT_VERSION_MAJOR 1


CMakeLists.txt
```diff
@@ -502,7 +502,51 @@ add_library(torrent-rasterbar
 	${libtorrent_aux_include_files}
 )
 
-select_cxx_standard(torrent-rasterbar 11)
+# Full C++ 11 support is required
+target_compile_features(torrent-rasterbar
+	PUBLIC
+		cxx_std_11
```
arvidn, line 508: it seems risky to declare all of C++11 must be supported. If there's a pedantic interpretation of this, there are compilers that don't support the garbage collection support added to the standard library in C++11 (and nobody uses that anyway).

CMakeLists.txt
```diff
@@ -502,7 +502,51 @@ add_library(torrent-rasterbar
...
+		cxx_reference_qualified_functions
+		cxx_nonstatic_member_init
+		cxx_variadic_templates
+		cxx_generalized_initializers
+		cxx_static_assert
+		cxx_auto_type
+		cxx_trailing_return_types
+		cxx_lambdas
+		cxx_decltype
+		cxx_decltype_incomplete_return_types
```
arvidn, line 519: I don't think this needs to be here

CMakeLists.txt
```diff
@@ -502,7 +502,51 @@ add_library(torrent-rasterbar
...
+		cxx_sizeof_member
+		cxx_inline_namespaces
+		cxx_unrestricted_unions
+		cxx_local_type_template_args
+		cxx_range_for
+		cxx_final
+		cxx_override
+		cxx_noexcept
+		cxx_defaulted_move_initializers
+		cxx_thread_local
```
arvidn, line 548: there's actually a work-around in libtorrent when `thread_local` is not available, so this shouldn't be here

CMakeLists.txt
FranciscoPombal: @arvidn 

Well, then, other than `cxx_std_11`, do you think any of these is relevant enough on their own to be here? Like I said, just `cxx_std_11` will also do fine.

CMakeLists.txt
FranciscoPombal: fixed

CMakeLists.txt
FranciscoPombal: fixed

CMakeLists.txt
arvidn: I'm not sure I know what you mean. I think all of the features below are used (with a few exceptions I pointed out)

FranciscoPombal: @arvidn so, have you decided on whether requiring only `cxx_std_11` is sufficient?

arvidn: it would seem to me it would be safer to leave the other, individual, requirements instead. is `cxx_std_11` not an alias for all of the other ones?

FranciscoPombal: @arvidn 

> it would seem to me it would be safer to leave the other, individual, requirements instead.

I took it from the previous comments/reviews you would probably want to trim the list a little more still, but if that's not the case, then indeed it is safer this way. I will squash the fixup and mark the PR as ready for review.

> is `cxx_std_11` not an alias for all of the other ones?

Unfortunately, no. In practice it just means "_some_ kind of basic C++11 support".

arvidn: all of the remaining features are used in libtorrent right now, so I don't think it should be trimmed any more.

arvidn: I was under the impression that you would remove the `cxx_std_11` entry though. Did I misunderstand that?

FranciscoPombal: @arvidn 

I think it's best to leave it in. Aside from "potential bad side effects" that specifying the individual features but not the high-level meta requirements might have (don't know if any, but best not to risk it), there is this quote from Professional CMake 7th edition:

> In the early days of C++11 and C++14, working with compile features was potentially useful because
compiler support was often lagging behind. For later language releases, support by the mainstream
compilers has arrived much faster. Subsequently, CMake stopped providing fine-grained features
for language standards beyond C++14. For C++17 and later, only the high level meta features like `cxx_std_17` are provided.

So leaving the high level meta feature is more consistent with the future releases; it's only the individually specified features that will get removed.

arvidn:
I get the impression that this is idiomatic cmake use, so that's probably good enough reason to use this feature.

I think I understand why the cxx_std11 should be used as well, and I'm less sure about whether there's any point in leaving the other features in that list.

@zeule do you know of any reasons to stick with the previous code? I would worry that it increasingly would lead to incompatibilities with clients, as it's a bit off the beaten path.

The case I would worry about (that I don't think was mentioned in the bullet list); imagine using a compiler that supports c++11, but it's c++98 by default. A c++98 client links against libtorrent. Will the client be built with c++98 and libtorrent with c++11? (I think the ideal behaviour would be to either fail compilation or to propagate the c++11 requirement up the dependency tree).

zeule:
I guess something needs to be done to convert `cxx_stdXX` into compiler flags for pkg-config.

zeule:
> I guess something needs to be done to convert `cxx_stdXX` into compiler flags for pkg-config.

Oops, sorry, I already wrote that code.

arvidn:
ok, @FranciscoPombal this is good to go then?

FranciscoPombal:
@arvidn 

> The case I would worry about (that I don't think was mentioned in the bullet list); imagine using a compiler that supports c++11, but it's c++98 by default. A c++98 client links against libtorrent. Will the client be built with c++98 and libtorrent with c++11? (I think the ideal behaviour would be to either fail compilation or to propagate the c++11 requirement up the dependency tree).

The C++11 requirement will get propagated up the dependency tree because of the `PUBLIC` keyword, as long as everything else is done right with the pkg-config files and the CMake config files produced by the build (which is the case here).

I just tried with a simple program, passing `-DCMAKE_CXX_STANDARD=98` to its configure command line; the generated compile commands end up containing `-std=gnu++11`, which is the expected result (it demonstrates that the C++11 requirement was propagated up from libtorrent).

I did notice though, that if you build libtorrent with a later standard that required, say C++14, the propagated _requirement_ will still be C++11. Is this a problem? If so, we probably need to alter the requirement propagation logic a bit as a follow-up.

arvidn:
I don't think it's generally a safe assumption that object files built with different versions of C++ are link-compatible.

FranciscoPombal:
@arvidn 

> I don't think it's generally a safe assumption that object files built with different versions of C++ are link-compatible.

So you're saying that the propagated C++ version requirement for libtorrent 1.2.x should be `max(C++11, C++ version with which it was build)` instead of C++11 unconditionally?

arvidn:
no, I'm saying that object files built with different versions of C++ should not be linked together. so if the target linking against libtorrent is built with C++98, the build should fail. It's not the end of the world if cmake doesn't support this though.
And if the target linking against libtorrent is built with C++14, libtorrent need to be built with C++14 too.

FranciscoPombal:
@arvidn 

> no, I'm saying that object files built with different versions of C++ should not be linked together. 
> And if the target linking against libtorrent is built with C++14, libtorrent need to be built with C++14 too.

Right, but the logical consequence of this is that libtorrent's CMake build script:
- has to be aware of the C++ standard level actually used when building libtorrent, not just the bare minimum
- write that as a requirement (as opposed to just C++11 uncondtionally) in the resulting CMake configure files/pkgconfigs that other applications will use to link against it.

Hence my comment about setting `max(C++11, C++ version with which it was build)` as the requirement to propagate. Or am I missing something?

>so if the target linking against libtorrent is built with C++98, the build should fail. It's not the end of the world if cmake doesn't support this though.

I don't known whether or not you can make CMake fail right away in this case.

Currently, the _configure_ does not fail - the target is simply forced to used C++11 (i.e., the `-std=gnu++11` is passed to all the compiler command lines). Maybe the build could fail eventually, if the client is written in such a way that will make the build fail when using C++11.

arvidn:
> Right, but the logical consequence of this is that libtorrent's CMake build script:

ok. I take it, then, that cmake doesn't have built-in support for ensuring this level of link-compatibility.

> Hence my comment about setting max(C++11, C++ version with which it was build) as the requirement to propagate. Or am I missing something?

My understanding (which is likely wrong) is that the cmake-way of things is to have build configurations only propagate *down*, to dependencies, and only requirements like include paths and defines propagating up. So I wouldn't actually expect the C++ version to *ever* propagate up the tree. consider target A, linking against B and C. B requires C++11 and C requires C++14. If both C++11 and C++14 propagates up, you get a conflict (presumably). Hopefully A, B and C would still be built with the *same* version.

> Currently, the configure does not fail - the target is simply forced to used C++11 (i.e., the -std=gnu++11 is passed to all the compiler command lines). Maybe the build could fail eventually, if the client is written in such a way that will make the build fail when using C++11.

So it sounds like cmake doesn't consider different C++ versions as link-incompatible then.

FranciscoPombal:
@arvidn 

>  consider target A, linking against B and C. B requires C++11 and C requires C++14. If both C++11 and C++14 propagates up, you get a conflict (presumably). Hopefully A, B and C would still be built with the same version.

I assume in this example, the C++ version requirement for at least targets B and C is `PUBLIC`. In this case, if I am not mistaken, there will not be a conflict - `A` will simply be forced to build with C++14 (the appropriate `-std=` flag will be passed, if needed, to ensure this).

> Hopefully A, B and C would still be built with the same version.

The way this is written makes me think there is a tacit assumption that `A`, `B` and `C` are somehow built at the same time and are possibly part of the same project. I don't think this is the case we are dealing with here. Imagine you build libtorrent one day. This ultimately results in generating and installing the pkgconfig/CMake configuration files that other projects will consume when linking against it later, at a different time.
For a target that links against libtorrent, CMake will not rebuild libtorrent based on the C++ standard used by the target to make sure they are the same. Instead, it will just look at the minimum required by libtorrent (as specified in the aforementioned pkgconfig/configuration files) and make sure the application uses a standard version equal or greater to that one.

> So it sounds like cmake doesn't consider different C++ versions as link-incompatible then.

So long as the version used by your target is equal or greater that that used by all of its dependencies, CMake seems OK with it.

>ok. I take it, then, that cmake doesn't have built-in support for ensuring this level of link-compatibility.

It is probably possible for the target that links against libtorrent to purposely fail the build if it detects that the version used to build libtorrent is not the same as the selected version to build the target itself. But there is nothing else we can do on libtorrent's side, as far as the CMake scripts are concerned. In this particular case, the only additional you could do to improve the situation is to fix the issue identified in https://github.com/arvidn/libtorrent/issues/4965#issuecomment-671268442.

Overall, the most you can do is try to have as few of these incompatibilities as possible, and simply warn in the documentation that possible incompatibilities might arise if the target/application that links against libtorrent uses a different (more recent) C++ version than libtorrent (I believe this is already written somewhere in the docs).

arvidn:
right, pkg-config doesn't really support ensuring link-compatibility. I'm really just concerned with the case where libtorrent is a proper cmake target, part of a larger build. (I would say the pkg-config case doesn't really matter that much because if one care about link compatibility, all dependencies would be built from source)

FranciscoPombal:
@arvidn 

> I'm really just concerned with the case where libtorrent is a proper cmake target, part of a larger build.

If libtorrent is a dependency of something else, that something else will always be built _after_ libtorrent, and the dependent target will always be aware of the requirements when looking at the CMake configuration files generated by libtorrent's build, right? If you can import the `LibtorrentRasterbar::torrent-rasterbar` target (via `find_package()`), you know its C++ version requirement because is communicated via its target properties when you import it.

If someone just takes the whole libtorrent source and writes their own CML for it as part of a larger project, they are in full control anyway and will not have trouble enforcing whatever constraint they want to enforce.

 What am I missing here?

arvidn:
> If someone just takes the whole libtorrent source and writes their own CML for it as part of a larger project, they are in full control anyway and will not have trouble enforcing whatever constraint they want to enforce.

I would expect the most reasonable way to link against libtorrent would be to pull it in as a submodule and then say:

```
add_subdirectory(libtorrent)
```

followed by:

```
target_dependencies(my-target libtorrent)
```

So, whoever builds `my-target` ought to be in control over how libtorrent is built. For example, which C++ version is used. Because it better use the same. There are other build options that better be the same as well, for example `-fabi-version` or `-fshort-enums`, `-fno-exceptions`, `-fpack-struct` and the `_GLIBCXX_USE_CXX11_ABI` define.

That's what I mean by build configurations propagate down, not up, the dependency tree. If the propagate up, they would also have to propagate down again, to sibling targets.

Anyway, if cmake doesn't really work this way, it probably isn't worth trying to "fight" it by trying to implementing part of such logic in a CMakeLists.txt.

FranciscoPombal:
@arvidn 

> I would expect the most reasonable way to link against libtorrent would be to pull it in as a submodule and then say:
> 
> ```
> add_subdirectory(libtorrent)
> ```
> 
> followed by:
> 
> ```
> target_dependencies(my-target libtorrent)
> ```
> 
> So, whoever builds `my-target` ought to be in control over how libtorrent is built. For example, which C++ version is used. Because it better use the same. There are other build options that better be the same as well, for example `-fabi-version` or `-fshort-enums`, `-fno-exceptions`, `-fpack-struct` and the `_GLIBCXX_USE_CXX11_ABI` define.

I'm not too familiar with workflows that use external dependencies directly at build-time. Seems the modern way to do it is with `FetchContent`, which supersedes `ExternalProject_Add`. I have to learn more about it.

> That's what I mean by build configurations propagate down, not up, the dependency tree. If the propagate up, they would also have to propagate down again, to sibling targets.

My understanding of `PUBLIC` (or `INTERFACE`, since `PUBLIC` just means `PRIVATE` + `INTERFACE`) is that requirements propagate "up", without propagating "down" again.

Relevant quotes from the documentation of the `target_compile_features()` command...

>Specifies compiler features required when compiling a given target. If the feature is not listed in the CMAKE_C_COMPILE_FEATURES, CMAKE_CUDA_COMPILE_FEATURES, or CMAKE_CXX_COMPILE_FEATURES variables, then an error will be reported by CMake. If the use of the feature requires an additional compiler flag, such as -std=gnu++11, the flag will be added automatically.
>
>The INTERFACE, PUBLIC and PRIVATE keywords are required to specify the scope of the features. PRIVATE and PUBLIC items will populate the COMPILE_FEATURES property of \<target\>. PUBLIC and INTERFACE items will populate the INTERFACE_COMPILE_FEATURES property of \<target\>. (IMPORTED targets only support INTERFACE items.) Repeated calls for the same \<target\> append items.

...and the `INTERFACE_COMPILE_FEATURES` property:

> List of public compile features requirements for a library.
>
>Targets may populate this property to publish the compile features required to compile against the headers for the target. The target_compile_features() command populates this property with values given to the PUBLIC and INTERFACE keywords. Projects may also get and set the property directly.
>
>When target dependencies are specified using target_link_libraries(), CMake will read this property from all target dependencies to determine the build properties of the consumer.

Plus this whole section: https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#transitive-usage-requirements

Here's a simple example:

```cmake
cmake_minimum_required(VERSION 3.16 FATAL_ERROR) # Policies <= CMP0097 default to NEW

project(testProject
    LANGUAGES CXX
)

# libtorrent 1.2.x has a PUBLIC C++ standard requirement of cxx_std_11
find_package(LibtorrentRasterbar 1.2.9 CONFIG REQUIRED COMPONENTS torrent-rasterbar)

add_library(foo)
target_sources(foo PRIVATE foo.cpp)
target_compile_features(foo PRIVATE cxx_std_14)

add_library(bar)
target_sources(bar PRIVATE bar.cpp)
target_compile_features(bar PUBLIC cxx_std_17)

add_executable(test)
target_sources(test PRIVATE main.cpp)
target_compile_features(test PUBLIC cxx_std_98)

target_link_libraries(test
    PRIVATE
        LibtorrentRasterbar::torrent-rasterbar
        foo
        bar
)
```

Here's the corresponding generated `compile_commands.json`:

```json
[
{
  "directory": "/home/francisco/Desktop/test/test_project/build",
  "command": "/usr/bin/c++ -DBOOST_ALL_NO_LIB -DBOOST_ASIO_ENABLE_CANCELIO -DBOOST_SYSTEM_DYN_LINK -DTORRENT_LINKING_SHARED -DTORRENT_NO_DEPRECATE -DTORRENT_USE_ICONV -DTORRENT_USE_LIBCRYPTO -DTORRENT_USE_OPENSSL -isystem /opt/boost-1.73.0/include -fexceptions -std=gnu++17 -o CMakeFiles/test.dir/main.cpp.o -c /home/francisco/Desktop/test/test_project/main.cpp",
  "file": "/home/francisco/Desktop/test/test_project/main.cpp"
},
{
  "directory": "/home/francisco/Desktop/test/test_project/build",
  "command": "/usr/bin/c++ -std=gnu++14 -o CMakeFiles/foo.dir/foo.cpp.o -c /home/francisco/Desktop/test/test_project/foo.cpp",
  "file": "/home/francisco/Desktop/test/test_project/foo.cpp"
},
{
  "directory": "/home/francisco/Desktop/test/test_project/build",
  "command": "/usr/bin/c++ -std=gnu++17 -o CMakeFiles/bar.dir/bar.cpp.o -c /home/francisco/Desktop/test/test_project/bar.cpp",
  "file": "/home/francisco/Desktop/test/test_project/bar.cpp"
}
]
```

So, as you can see, `bar` propagated its `PUBLIC` C++17 version requirement up to the `test` executable, but that requirement was not propagated down again to the `foo` library target, which is still built with C++14.

If `bar`'s requirement were PRIVATE instead, `test` would be built with `-std=gnu++11`, since while its own compile features requirement is just C++98, it still has to respect the `PUBLIC` requirement of C++11 propagated up from its dependency on libtorrent 1.2.x. Additionally, in this case, `bar` would be built with `-std=gnu++17`.

I guess what this means is that for CMake, what matters is that the minimum standard version requirement (as determined by the `PUBLIC`/`PRIVATE`/`INTERFACE` propagation rules) is met; the onus of guaranteeing link compatibility (or warn of lack thereof) when targets compiled with a greater C++ version are linked against a library compiled with a previous one falls on the project(s). Alternatively, the onus of using the same C++ version for compiling all targets/dependencies falls on the user.

FranciscoPombal:
@arvidn 

> I'm really just concerned with the case where libtorrent is a proper cmake target, part of a larger build.

To address your main concern more directly and clearly than in https://github.com/arvidn/libtorrent/pull/4968#issuecomment-674589488 and other comments:

- if libtorrent is part of a larger build, I think the developer can link every target (including libtorrent, imported by `FetchContent`, for example) against an `INTERFACE` target that propagates a certain C++ version requirement to all other targets. The only thing to watch out for is to make sure that this "global" version requirement is greater than or equal to the greatest value found among all targets.

- If libtorrent is used from an installation (e.g. via `find_package()`), there is really no way to guarantee the user won't somehow use a more recent C++ version for other targets that link against libtorrent. So at most you can warn in the documentation not to do this, and try to minimize/fix these kinds of incompatibilities where possible.
  - My suggestion above about propagating  `max(C++11, C++ version with which it was built)` as the requirement instead of just C++11 is only useful to prevent the special case where libtorrent is built with a version higher than minimum (e.g. C++14), but targets built with a lower-but-still-higher-than-or-equal-to-the-minimum version (e.g. C++11) can still link against it. So I guess it's not that important.

PR #4970, opened 2020-08-11
fix torrent round-trip for v2 torrents in create_torrent

arvidn:


src/create_torrent.cpp
```diff
@@ -493,16 +493,26 @@ namespace {
...
+			auto const& file_trees = ti.internal_merkle_trees();
 			m_fileroots.resize(m_files.num_files());
 			m_file_piece_hash.resize(m_files.num_files());
 			for (auto const i : m_files.file_range())
 			{
 				// don't include merkle hash trees for pad files
 				if (m_files.pad_file_at(i)) continue;
-				m_file_piece_hash[i].resize(std::size_t(m_files.file_num_pieces(i)));
 
-				// TODO: copy piece layer from file_trees[i]
```
arvidn, line 505: this piece was missing. if one would create a torrent from a `torrent_info` (as qBittorrent does when it saves resume data), v2 torrents would fail to include the piece layers.

PR #4979, opened 2020-08-12
fix s390x tests and endian bugs

arvidn:


arvidn:
@aldenml after fixing the incorrect `test-launcher.sh`, I get the same thing on travis as I do at home, https://travis-ci.com/github/arvidn/libtorrent/builds/179460650#L848

```
terminate called:

<not supported>
```

aldenml:
interesting, I will run it in my VM and write back the result

arvidn:
I got it. https://travis-ci.com/github/arvidn/libtorrent/builds/179653600#L2500

aldenml:
@arvidn great! and the use of travis arch looks like a real win

PR #4991, opened 2020-08-14
fix incorrect mapping of error code to error string

tjps:
This fixes a slightly maddening issue where the enum of error codes didn't line up with the human readable error strings.

The map overhead is only incurred when converting to a human-readable string, so I assume it is acceptable.

It might not be apparent from the diff, but check out the location of the bittyrant message in the enum vs. the string list.

### DIFF
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -384,8 +384,10 @@ TORRENT_VERSION_NAMESPACE_2
 		error_code const error;
 	};
 
-	// This alert is generated when a limit is reached that might have a negative impact on
-	// upload or download rate performance.
+	// This alert is generated when a limit is reached that might have
+	// a negative impact on upload or download rate performance.
+	// NOTE: any additions to this enum should have a corresponding entry
+	//   added to the `warning_str` map in alert.cpp
 	struct TORRENT_EXPORT performance_alert final : torrent_alert
 	{
 		enum performance_warning_t
--- a/src/alert.cpp
+++ b/src/alert.cpp
@@ -278,23 +278,36 @@ namespace libtorrent {
 
 	std::string performance_alert::message() const
 	{
-		static char const* const warning_str[] =
+		static const std::unordered_map<performance_warning_t, char const*> warning_str =
 		{
-			"max outstanding disk writes reached",
-			"max outstanding piece requests reached",
-			"upload limit too low (download rate will suffer)",
-			"download limit too low (upload rate will suffer)",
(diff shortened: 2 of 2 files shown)


include/libtorrent/alert_types.hpp
```diff
@@ -384,8 +384,10 @@ TORRENT_VERSION_NAMESPACE_2
 		error_code const error;
 	};
 
-	// This alert is generated when a limit is reached that might have a negative impact on
-	// upload or download rate performance.
+	// This alert is generated when a limit is reached that might have
+	// a negative impact on upload or download rate performance.
+	// NOTE: any additions to this enum should have a corresponding entry
+	//   added to the `warning_str` map in alert.cpp
```
arvidn, line 390: I don't think this note quite belongs here. Putting it here will make it end up in the html documentation as well, but it's not targeting users, but implementors

src/alert.cpp
```diff
@@ -278,23 +278,36 @@ namespace libtorrent {
 
 	std::string performance_alert::message() const
 	{
-		static char const* const warning_str[] =
+		static const std::unordered_map<performance_warning_t, char const*> warning_str =
```
arvidn, line 281: is it really an improvement to use an unordered map here? It seems it would be a lot better to just enforce this mapping is correct with a test that samples a few errors

src/alert.cpp
```diff
@@ -278,23 +278,36 @@ namespace libtorrent {
...
+			{ download_limit_too_low, "download limit too low (upload rate will suffer)" },
+			{ send_buffer_watermark_too_low, "send buffer watermark too low (upload rate will suffer)" },
+			{ too_many_optimistic_unchoke_slots, "too many optimistic unchoke slots" },
+			{ too_high_disk_queue_limit, "the disk queue limit is too high compared to the cache size. The disk queue eats into the cache size" },
+			{ aio_limit_reached, "outstanding AIO operations limit reached" },
+			{
+#if TORRENT_ABI_VERSION == 1
+				bittyrant_with_no_uplimit,
+#else
+				deprecated_bittyrant_with_no_uplimit,
```
arvidn, line 295: I would expect to not have this string at all when deprecated functions are disabled. There's no way to get this error in this case, is there?

arvidn:
I'm suggesting this fix instead. https://github.com/arvidn/libtorrent/pull/4993

arvidn:
thanks for the report!

crbrayity132139:
https://github.com/arvidn/libtorrent/issues/#5167

Issue #4989, opened 2020-08-14
Optimise fast connections over slow connections for torrents with many seeds


Sophist-UK:
**Please provide the following information**

libtorrent version (or branch): 1.2.6.0 (in latest version of qBitTorrent)

platform/architecture: Windows 10 64-bit

compiler and compiler version: Not known

When downloading a torrent with a large number of seeds, I get some connections which download at several 100Kbps, and some that download at a few bytes per second - and I would like LibTorrent to drop these in the hope that it can find a significantly faster seed. I also get a lot of connections to seeds which wait some time before downloading. I would like LibTorrent to:

a. For a torrent with many more seeds than allowed connections (setting to define the multiple), drop any connection which downloads at a speed much slower than the average of the fastest 50% of all other connections (setting to define the %). So if I have 3 connections at > 100Kbps and two which are below 1Kbps then the average speed of the top 2 would be c. 100Kbs; the bottom one is < 5% of this (i.e. below 5Kbps) so it gets dropped.

b. Ideally I would like the BitTorrent protocol to allow seed to estimate the time before it starts downloading and to estimate the maximum upload speed, so that LibTorrent can decide whether to hold on to that connection or try to find another connection which will start faster and / or download faster (using the above algorithm for the latter - just without waiting for the download to actually start and stabilise).  But I appreciate that this might need enhancements to the BitTorrent protocol and that this might therefore need to be a long-term aspiration - in which case I would like LibTorrent to do the best it can to optimise connections which actually download over those which wait a long time (e.g. by caching the wait time and speed for each IP address).

Obviously, for torrents with a limited number of seeds, a slow download is better than none.

Sophist-UK:
To make it easier to understand, here is a screen shot. As you can see I have 4 seeds downloading at over 100KB/s, and 3 downloading at under 10KB/s. I would like LibTorrent to cancel the 3 slow ones and try to find 3 much faster ones instead.

![image](https://user-images.githubusercontent.com/3001893/90282161-2cf14680-de6e-11ea-9d85-db805df3c1a8.png)

arvidn:
I also notice that you only have 15 peers. The default limit is 50 peers per torrent. So I would expect libtorrent to *already* be trying to connect more peers, regardless of whether the existing slow ones are left connected or not.

Do you have a reason to believe that disconnecting the slow ones would improve your chances of finding faster seeds?

[...]

PR #4996, opened 2020-08-15
Tweak Peer Turnover prioritisation...

Sophist-UK:
... to prefer peers still waiting to download over the slowest actually downloading.

This is a partial solution to the issue discussed in #4989.

Peer Turnover only applies when most connection slots are in use, and the primary purpose is to free up the least beneficial connection slots in the hope that LibTorrent can find a faster download peer.

Previously a connection which has been made but which has not yet started downloading is preferred for disconnect over the slowest downloading connection.

The primary aim of this change is to prioritise connections that are waiting to start downloading over the slowest actually downloading on  the assumption that a connection that has already been made but is waiting to start downloading has the same chance of being faster than any new connection - and we should not lose the time already spent waiting by disconnecting and making a new connection that then waits. 

This is perhaps a bit simplistic, because it leaves LibTorrent open to keeping (potentially malicious) connections open that are (perhaps deliberately) never going to start downloading. To avoid this we will limit the above change to connections which have been waiting less than Peer_Turnover_Maximum_Wait and retain the current preference of disconnecting the waiting connection if it has been waiting longer than this, For this commit Peer_Turnover_Maximum_Wait is a hardcoded constant but this can be changed to a configurable setting in a future commit.

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -7429,23 +7429,51 @@ bool is_downloading_state(int const st)
 		if (lhs->on_parole() != rhs->on_parole())
 			return lhs->on_parole();
 
-		// prefer to disconnect peers that send data at a lower rate
-		std::int64_t lhs_transferred = lhs->statistics().total_payload_download();
-		std::int64_t rhs_transferred = rhs->statistics().total_payload_download();
-
-		time_point const now = aux::time_now();
-		std::int64_t const lhs_time_connected = total_seconds(now - lhs->connected_time());
-		std::int64_t const rhs_time_connected = total_seconds(now - rhs->connected_time());
-
-		lhs_transferred /= lhs_time_connected + 1;
-		rhs_transferred /= (rhs_time_connected + 1);
-		if (lhs_transferred != rhs_transferred)
-			return lhs_transferred < rhs_transferred;
-
 		// prefer to disconnect peers that chokes us
 		if (lhs->is_choked() != rhs->is_choked())
 			return lhs->is_choked();
 
+		// prefer to disconnect peers that send data at a lower rate over those waiting to download
+		// the assumption is that a connection waiting to start downloading has the same potential 
+		// to be faster than the current slowest downloading connection than a new connection that 
+		// replaces a waiting connection - and disconnecting a waiting connection starts a new wait
+		// however ... to avoid torrents which wait forever there needs to be a maximum wait time
+		// after which a waiting connection is disconnected
+		std::int64_t lhs_transferred = lhs->statistics().total_payload_download();
+		std::int64_t rhs_transferred = rhs->statistics().total_payload_download();
+		bool const is_waiting = lhs_transferred * rhs_transferred == 0;
+
+		// for this commit, the maximum time is a constant - but it could be replaced by a new 
+		// setting peer_turnover_maximum_wait in a future commit
+		std::int64_t const peer_turnover_maximum_wait = 60;
+		time_point const now = aux::time_now();
+		std::int64_t const lhs_time_connected = total_seconds(now - lhs->connected_time());
+		std::int64_t const rhs_time_connected = total_seconds(now - rhs->connected_time());
+		
+		// if one connection is waiting
+		if ((lhs_transferred != rhs_transferred) && (is_waiting))
+		{
+			// One of lhs_transferred & rhs_transferred is zero - get connected time for waiting connection
+			std::int64_t const waiting_time = 
+				((lhs_transferred * rhs_time_connected) + (rhs_transferred * lhs_time_connected)) 
+				/
+				(lhs_transferred + rhs_transferred);
+			if (waiting_time <= peer_turnover_maximum_wait)
+				return lhs_transferred != 0;
+			else return lhs_transferred == 0;
+		}
+
+		// when both have downloaded something prefer to disconnect peers that send data at a lower rate
+		else if (!is_waiting)
+		{
+			lhs_transferred /= lhs_time_connected + 1;
+			rhs_transferred /= (rhs_time_connected + 1);
+			if (lhs_transferred != rhs_transferred)
+				return lhs_transferred < rhs_transferred;
+		}
+
+		// if none of the above preferences apply then ...
+		// prefer to disconnect peers that have been silent longest
 		return lhs->last_received() < rhs->last_received();
 	}
 
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -7429,10 +7429,6 @@ bool is_downloading_state(int const st)
 		if (lhs->on_parole() != rhs->on_parole())
 			return lhs->on_parole();
 
-		// prefer to disconnect peers that chokes us
-		if (lhs->is_choked() != rhs->is_choked())
-			return lhs->is_choked();
-
 		// prefer to disconnect peers that send data at a lower rate over those waiting to download
 		// the assumption is that a connection waiting to start downloading has the same potential 
 		// to be faster than the current slowest downloading connection than a new connection that 
@@ -7471,6 +7467,11 @@ bool is_downloading_state(int const st)
 			if (lhs_transferred != rhs_transferred)
 				return lhs_transferred < rhs_transferred;
 		}
+
+		// prefer to disconnect peers that chokes us
+		// this code dropped with agreement from @arvidn
+		// if (lhs->is_choked() != rhs->is_choked())
+		//	return lhs->is_choked();
 
 		// if none of the above preferences apply then ...
 		// prefer to disconnect peers that have been silent longest


Sophist-UK:
Note to code reviewers:

1. I am not a C+ programmer - I have created the code by copying other similar code.
2. I do not have a test environment and have not compiled or tested this code.

Otherwise, typos excepted, this should work.

src/torrent.cpp
```diff
@@ -7429,23 +7429,51 @@ bool is_downloading_state(int const st)
...
+		if ((lhs_transferred != rhs_transferred) && (is_waiting))
+		{
+			// One of lhs_transferred & rhs_transferred is zero - get connected time for waiting connection
+			std::int64_t const waiting_time = 
+				((lhs_transferred * rhs_time_connected) + (rhs_transferred * lhs_time_connected)) 
+				/
+				(lhs_transferred + rhs_transferred);
+			if (waiting_time <= peer_turnover_maximum_wait)
+				return lhs_transferred != 0;
+			else return lhs_transferred == 0;
```
arvidn, line 7463: I think this is a really involved way of saying what you want to say. But it's also not clear that it makes sense to do this, in general. A peer that has never sent anything to you, will be prioritized. Over time you would presumably accrue many such peers.

src/torrent.cpp
```diff
@@ -7429,23 +7429,51 @@ bool is_downloading_state(int const st)
 		if (lhs->on_parole() != rhs->on_parole())
 			return lhs->on_parole();
 
-		// prefer to disconnect peers that send data at a lower rate
+		// prefer to disconnect peers that chokes us
+		if (lhs->is_choked() != rhs->is_choked())
+			return lhs->is_choked();
```
arvidn, line 7434: this is a bit risky. a peer can unchoke you but still not send you any data. A peer that has temporarily choked you, but otherwise is a good peer, would be penalized

src/torrent.cpp
Sophist-UK: In theory, you shouldn't accumulate many such peers waiting to download because over time they should start to actually download. But in reality:

a. Some peers will take a long time to start to download;
b. If a specially crafted malicious peer can connect and deliberately wait indefinitely, they could cause a DoS.  

And this is why I only prioritise disconnects of downloading connections over waiting connections if the wait time is < Peer_Turnover_Maximum_Wait - currently a constant, but it could be a setting. If a connection has waited longer than this then the waiting connection is prioritised for disconnection over the downloading connection, thus avoiding the accumulation of waiting connections.

src/torrent.cpp
Sophist-UK: In the existing code, it is unlikely that choked peers are ever disconnected because the previous check is on speed, and it is unlikely that the speed of two connections is the same esp. if one is choked.

However, I can see that a temporary choke on a fast connection might be preferable to a slower connection which is not choked.

I can either move this code back to where it was or drop it.

Sophist-UK:
@arvidn Did you cancel the checks? How will I know if there are further code issues if the checks were cancelled?

src/torrent.cpp
arvidn: I think it would make sense to drop this condition, if the aspiration is to come up with something better and less crude. In general it's probably a good idea to primarily look at actual useful data received, and no second or third order indicators.

src/torrent.cpp
arvidn: I think the other peer should be assumed to be malicious

src/torrent.cpp
Sophist-UK: Ok - will fix this tomorrow.

arvidn:
There is no unit test for this function, so no test would fail because of it (since it builds)

src/torrent.cpp
Sophist-UK: Sorry - I don't understand what you mean. This doesn't assume maliciousness as such, it just expects a waiting connection to start downloading in a reasonable timescale, and if it hasn't started downloading by Peer_Turnover_Maximum_Wait it prioritises the waiting connection to be disconnected over and above a downloading connection.

arvidn: yeah, that was a general comment. However, I think you can express this logic more straight-forward. For example, a peer that has been connected for less than 60 seconds could always be prioritized over one that has been connected longer. I think that could be done regardless of the number of bytes transferred.

Keep in mind that it can take a while for a TCP connection to ramp-up as well, especially if there's a lot of cross-traffic

Sophist-UK:
@arvidn I think this is ready for someone to test. 

Providing you are going to merge this, then if you would like me to code for adding the Peer_Turnover_Maximum_Wait setting as part of this PR before merging then let me know.

src/torrent.cpp
Sophist-UK: But that is not the logic. A connection <60s which is downloading slowly should NOT be prioritised over a connection that has been open longer but is faster.

Sophist-UK:
I just took a look at /test/test_torrent.cpp and put simply my c+ skills are not up to creating unit tests for this. (But see above about my reservations of the usefulness of such tests - esp. given that the previous algorithm had no unit tests either.)

Sophist-UK:
@arvidn Arvid - I am not trying to rush you, but is this going to be merged? If it is, I would be happy to start to look at the next PR as discussed in #4989

Sophist-UK:
@arvidn

> Arvid - I am not trying to rush you, but is this going to be merged? If it is, I would be happy to start to look at the next PR as discussed in #4989

arvidn:
oh, sorry. I still don't think there's compelling evidence that this is an improvement. But I recognize that this is a bit of a nebulous part of logic that can be tested against certain scenarios, without knowing whether those scenarios are representative of the real world.

I would like to see *some* attempt of showing that this is an improvement though. Have you tested it and observed an improvement? Could you measure it?

Sophist-UK:
As previously mentioned, I do not have a development environment in which to test it. However I have tested this approach in qbTorrent by manually killing off connections that are downloading very slowly and it does give me much faster download speeds.

arvidn:
it's not obvious that your updated logic in this function has the same behavior as you manually disconnecting peers. I don't think it's reasonable to merge untested code.

Sophist-UK:
I manually disconnect peers that are downloading at a low rate and leave peers that are connected but have not yet started downloading. This leads to MUCH faster downloads.

Additionally, the logic to this is clear

1. Where you have a lot of seeds, and you have slow downloads, it is significantly more likely that other seeds will download faster than existing slow connections.

2. You have no idea how fast a seed that is connected but has not started downloading will be when it downloads - what is the point in disconnecting a seed that you just connected to but which hasn't started downloading just to connect to a different seed?

Sophist-UK:
I manually disconnect peers that are downloading at a low rate and leave peers that are connected but have not yet started downloading. This leads to MUCH faster downloads. To be strictly accurate, when I do it manually I only disconnect downloading seeds in preference to waiting seeds if the downloading seeds are significantly slower than other downloading seeds and there are a large number of other seeds for the torrent i.e. there is a strong probability of other seeds being faster than the slow ones I am disconnecting. 

The logic to this is clear

1. Where you have a lot of seeds, and you have relatively slow downloads (compared to other downloads), it is significantly more likely that other seeds will download faster than existing slow connections.

2. You have no idea how fast a seed that is connected but has not started downloading will be when it downloads - what is the point in disconnecting a seed that you have already connected to and are part way through the wait period for downloading to start but which hasn't started downloading just to connect to a different seed and starting the wait for a download from the beginning? Obviously you don't want to prioritise seeds that have excessively long wait times hence only doing this for those that have been waiting a short time.

However, I think you are right to be cautious about potential downsides. This is, however, a completely separate discussion to having tested the algorithm before implementing it - without having a repeatable performance testing environment, you end up testing on live torrents, and then you are subject to what torrents and seeds are currently available, and (IMO) you are likely to end up testing on torrents which are "normal" and are going to play well with this algorithm, but not test on special cases for which this change would be detrimental.

Looking at potential issues, whilst the vast majority of normal torrents would benefit, I can identify two use cases where this change would be detrimental:

1. If you have fewer seeds downloading than seeds which have been connected for <60s but are not yet downloading, it will disconnect the seeds that are downloading in preference to those waiting regardless of how fast they are downloading - but it will disconnected the slowest of these. In a worst case scenario of a lot of seeds waiting for <60s and one seed actually downloading, it will disconnect the one downloading regardless of how fast it is downloading. In a normal torrent, I would expect this to resolve itself because by the time the next Peer Turnover happened, a lot of the pending seeds would have started downloading. But it would have short term detrimental impact.

2. In the event of a denial of service attack whereby a large number of fake seeds are created which connect but never start downloading in order to block genuine seeds from connecting and downloading, this change might make things worse. Without this change, the blocking fake seeds would not stop working seeds from downloading when they were selected, it would simply make connections to working seeds far less likely - and it would be the fake seeds which would be disconnected as part of Turnover, with downloading seeds staying connected. With this change, the downloading seeds would be disconnected in favour of newly connected fake seeds, thus making the denial of service worse. 

This is a consequence of using a sort comparison function to evaluate the relative priorities of connections to be turned over. Because the sort routine only has access to the two connections being compared it cannot determine how many seeds are downloading or how the speed of the two connections being evaluated compare to e.g. the fastest connection.

Sophist-UK:
It seems to me that the comparison function could be enhanced to know more about the overall position using the following technique:

a. You analyse the existing download state and store it in some sort of dict.

b. You create a python "partial" function where this analysis is pre-provided and where the partial function is used as a comparison function.

Depending on what information the comparison function would find useful, you might provide a full set of connection data, or specific data e.g. the number of seeds and fastest / average connection download speed for each torrent.

Sophist-UK:
P.S. Just to put this in context, for those torrents where there are large number of seeds, based on my experiences doing this manually I believe that this PR could **DOUBLE** the average download speed and halve the download time for these torrents.

The performance benefits are genuinely significant - so can we please, pretty please, try to find a way to try this out?

Sophist-UK:
Having had time to consider this further:

1. IMO, having a better algorithm for choosing connections is KEY to improving download speeds.
2. I am really unsure that tweaking the current mechanism with this PR is the right way to go, partly because of the above analysis of DoS attacks and partly because the current mechanism (simply sorting the choice of disconnects) is probably too simplistic.

Here are my current thoughts:

a. If libTorrent is already maxing out the maximum download speed, disconnecting downloading connections will do nothing. Ditto if maxing out on upload speeds. So you probably need to consider connections with blocks of interest separately from those without. But... if you are not maxed out...

b. We need to decide whether to prioritise torrents then connections within the priority torrents - or just connections in general. At present we just connections in general - but I am not sure whether this is too simplistic. But assuming we want to priritise torrents then what basis should we do so (and can we apply these rules during a connection-only prioritisation or does the code need to do torrent prioritisation explicitly):

* Torrents with a ratio of connections:(peers with blocks of interest/unknown) close to 1 will NOT benefit from Peer Turnover - there are few different seeds to connect to, so disconnecting existing downloading connections will not help this torrent - whilst it will free up connections for use by other torrents, it is not clear whether this is what the user wants (depends on relative user priorities for torrents and whether we want to try to even out download end times or focus on making the soonest torrent finish sooner). It seems to me that if we knew which torrents could be speeded up, we should focus on those with the minimum speed up first, because once they finish we will have more bandwidth / connections that can be used with other torrents which can be speeded up more. We might be able to guess which torrents could be speeded up more - those with many seeds, those with variable download speeds from those seeds we are connected to - but this might be simplistic and detrimental - I would suggest that this is left to a second phase.

* Equally, if all existing connections for a torrent are relatively slow compared to a different torrent, that should NOT (IMO) mean we should prioritise these slower connections for turnover. IMO we should prioritise connections for turnover based on relative speed compared to other connections for the same torrent.

* Looking within a torrent, then the thinking behind this PR still applies i.e. we should use Peer Turnover, but we should favour disconnecting connections which are downloading "slowly" (see below for definition) over those which are relatively new connections and not yet downloading. What we should NOT do (IMO) is disconnect connections which are downloading "fast" - because a new connection might be unlikely to be faster and we will slow down the download whilst we disconnect, make a new connection, negotiate and wait for the download to start.

(In addition, to prevent the DoS attack previously identified, we should: A) have a maximum wait time for downloads from a connection with blocks of interest, and disconnect it if this wait time is exceeded; and B) retain a record of such previous connections - along with download speed - which we use to prioritise / deprioritise future connection attempts.) 

* As previously discussed / discussed elsewhere, for those peers we have connected to we should retain knowledge of their block availability in order to be able to prioritise whether we try to reconnect to them at a later point.

So what makes a specific connection relatively "slow" / "fast" for a specific torrent?

1. You cannot know for certain. It is statistical. The comments below are based on my experience...

2. If you are getting some connections downloading at >100Kbps, then (if there are loads of other potential connections) any connection downloading at < 10Kps are contributing little to the download and a faster connection could potentially contribute a lot more. (IME some users configure their torrent clients with artificially slow upload speeds - in single digit bps - and these are to be avoided if there are faster alternative connections because they contribute little to the download speed.) The question is where to draw the line about expecting a replacement connection to download faster - and this is where statistics need to come into it.

A. If we assume a normal distribution on download speeds...

f(x)=e^(−x^2/√(2π))
image

What this means (assuming that the distribution is like this) is that lower throughput connections are equally likely as higher throughput connections.

B. But factoring in the speed gives:

f(x)=x * e^(−x^2/√(2π))

It is clear from this that this suggests favouring higher throughput connections by disconnecting lower throughput connections is beneficial.

(I will try to find a graph of this and then to determine a bit more precisely what it means.)

# SUMMARY #

1. The principle underlying this PR still holds true - that we should disconnect slowly downlaoding connections in preference to newly made connections still waiting to download.

2. This PR can be enhanced to disconnect connections which have been waiting for a long time (say > max(3min, peer turnover interval)) in preference to slowly downloading connections to avoid the risk of DoS attacks..

3. Really, the whole peer turnover approach needs to be revisited in order to implement a better algorithm based on the above.

Sophist-UK:
Here is a real life example for a torrent with 600 seeds:
image
As you can see I have 1/2 of the connections waiting for download.

I have a total download speed of over 2MB/s Of those actually downloading speeds vary between 186B/s and 800K+B/s - a huge variance. If I disconnect the one running at 186B/s I will slow down the total speed by a factor of 0.01% - or in other words, if the download takes 3 hours at 2MB/s, removing the one at 186B/s will take a further c. 1s.

I have no idea what speeds I am going to get from the connections waiting to download - and it doesn't make sense to disconnect these and replace them with others and starting a new wait for a download UNLESS the 

On the other hand the mean download speed is c. 300KB/s - so if I disconnect the one at 186B/s which is really not doing much, and replace it with one at 300KB/s, then I am going to have a significant increase in overall speed.

This is what I do by manually banning peers with consistently slow download speeds. And it makes a BIG difference to my download times.

arvidn:
I think we may have been speaking past each other a bit.

I do not doubt that the situation can and should be improved by disconnecting idle peers more aggressively. As you show in your screenshot, most of those peers probably should be disconnected, at least once they have been given a chance to prove themselves better. It's all the other cases that shouldn't be broken I'm concerned about.

What I doubt is that your code actually does what you want it to do and that it accurately identifies those peers that you want to be disconnected.

I think you hand wave some of these concerns.

for example, you say:

> But that is not the logic. A connection <60s which is downloading slowly should NOT be prioritised over a connection that has been open longer but is faster.

Every peer that was just connected will be downloading slowly. The challenge is to give every peer an opportunity to overtake another fast peer. If all peers that connected less than 60s ago, that you got a low download rate from were disconnected, *all* new peers would *always* be disconnected.

There are a few things contributing to the delay of ramping up download rates:

1. The handshake, not just the BitTorrent handshake, but also the extended handshake message, interested and unchoke messages. These require a few round-trips in the best case, but can easily take a few seconds.
2. If you rely on being optimistically unchoked by the peer, that could take minutes. If you unchoke the peer and wait for reciprocation, that can take up to 15 seconds.
3. Once data is flowing, the transport layer needs to ramp up its congestion window. Assuming an initial window of 1400 bytes and a 200 ms round-trip, it would take 8 round-trips to reach 1.79 MB/s, 1.6 seconds. That's under ideal conditions.
4. Once data is flowing and the transport layer has ramped up, the bittorrent request layer also needs to ramp up (this would most likely happen in parallel with the transport layer though, but imperfections may cause short stalls).

When disconnecting peers more aggressively, it's critical that you take many different scenarios into account, not *just* the mode where you're downloading from seeds and have half of your connection slots clogged by slow peers.

To be able to bootstrap a download for instance, you may need to rely on just hanging out being connected to peers until one of them optimistically unchokes you. It's critical to not break that case, or to make it take a lot longer because you disconnected idle peers.

The best way to build confidence in not breaking things, is to add tests. This function could really do with some tests.

arvidn:
As far as I can tell, this patch does not increase the aggressiveness of disconnecting peers. It just alters the ordering of which peers to pick for disconnection.

I don't see how this would solve the problem you posted in your screenshot. To solve that you'd need some logic that actually disconnects peers under certain circumstances.

src/torrent.cpp
```diff
@@ -7429,23 +7429,52 @@ bool is_downloading_state(int const st)
 		if (lhs->on_parole() != rhs->on_parole())
 			return lhs->on_parole();
 
-		// prefer to disconnect peers that send data at a lower rate
+		// prefer to disconnect peers that send data at a lower rate over those waiting to download
+		// the assumption is that a connection waiting to start downloading has the same potential 
+		// to be faster than the current slowest downloading connection than a new connection that 
+		// replaces a waiting connection - and disconnecting a waiting connection starts a new wait
+		// however ... to avoid torrents which wait forever there needs to be a maximum wait time
```
arvidn, line 7436: Could you spell out what you mean by "..." here?

src/torrent.cpp
```diff
@@ -7429,23 +7429,52 @@ bool is_downloading_state(int const st)
...
+		// after which a waiting connection is disconnected
 		std::int64_t lhs_transferred = lhs->statistics().total_payload_download();
 		std::int64_t rhs_transferred = rhs->statistics().total_payload_download();
+		bool const is_waiting = lhs_transferred * rhs_transferred == 0;
```
arvidn, line 7440: I think this name is confusing. `is_waiting` sounds like it would refer to a peer, but this means "is one of the peer waiting". I also think there might be a better name referring to "connected by haven't sent data yet".

src/torrent.cpp
```diff
@@ -7429,23 +7429,52 @@ bool is_downloading_state(int const st)
...
+			if (lhs_transferred != rhs_transferred)
+				return lhs_transferred < rhs_transferred;
+		}
 
 		// prefer to disconnect peers that chokes us
-		if (lhs->is_choked() != rhs->is_choked())
-			return lhs->is_choked();
+		// this code dropped with agreement from @arvidn
+		// if (lhs->is_choked() != rhs->is_choked())
+		//	return lhs->is_choked();
```
arvidn, line 7474: if this logic shouldn't be here, it shouldn't be left in a comment. Especially not without a motivation. If you think it's likely that someone in the future will read this code and think "oh, but we should sort by whether we're unchoked or not", it might make sense to add a comment explaining why that doesn't make sense. But I don't think the old code needs to stay

src/torrent.cpp
```diff
@@ -7429,23 +7429,52 @@ bool is_downloading_state(int const st)
...
+		std::int64_t const peer_turnover_maximum_wait = 60;
 		time_point const now = aux::time_now();
 		std::int64_t const lhs_time_connected = total_seconds(now - lhs->connected_time());
 		std::int64_t const rhs_time_connected = total_seconds(now - rhs->connected_time());
+		
+		// if one connection is waiting
+		if ((lhs_transferred != rhs_transferred) && (is_waiting))
+		{
+			// One of lhs_transferred & rhs_transferred is zero - get connected time for waiting connection
+			std::int64_t const waiting_time =
```
arvidn, line 7453: this computation seems a bit awkward. Is there a good reason to do this rather than the straigh-forward:

```
auto const waiting_type = (lhs_transferred == 0) ? lhs_time_connected : rhs_time_connected;
```

I think that's easier for a human to interpret and you don't need any multiplication or division.

src/torrent.cpp
```diff
@@ -7429,23 +7429,52 @@ bool is_downloading_state(int const st)
...
+		if ((lhs_transferred != rhs_transferred) && (is_waiting))
+		{
+			// One of lhs_transferred & rhs_transferred is zero - get connected time for waiting connection
+			std::int64_t const waiting_time = 
+				((lhs_transferred * rhs_time_connected) + (rhs_transferred * lhs_time_connected)) 
+				/
+				(lhs_transferred + rhs_transferred);
+			if (waiting_time <= peer_turnover_maximum_wait)
+				return lhs_transferred != 0;
+			else return lhs_transferred == 0;
```
arvidn, line 7459: if I understand this correctly, a "waiting" peer that connected < 60s ago is prioritized over a peer that's currently transferring in this case, right?

Clearly to get here the peers both need to be equal in a few other comparisons that happen first, like being a seed and interesting.

It seems like there's an edge case that could be problematic though. Imagine a swarm if only seeds. You join and immediately fill all your peer connection slots with seeds. One of them starts sending data but the other ones stall for some reasons.

If the optimistic disconnect logic is invoked within 60 seconds, the one transferring peer will be the top disconnect candidate and be disconnected.

I suppose that constant, `peer_turnover_maximum_wait` critically depends on how often the disconnect logic is invoked. I think there should be (at least) a 2x margin between those constants. And there should either be comment explaining that these constants are linked, or better yet, use the same constant.

arvidn:
I think this patch looks good. I have a few comments on the style though. Also, if you don't mind, I'd appreciate if the comments could be wrapped at 80 columns (it looks better on github among other things).

I would really prefer it if you (or someone else) would actually test this logic before landing it too. If not by writing unit tests, at least by building a client with it and observing its behavior manually.

arvidn:
Another possible challenge is that the distinction between a peer that's "waiting" and one that was waiting until 0.5 seconds ago may be a bit unfair. The non-waiting peer surely can't be expected to have ramped up yet, to prove its true worth.

Do you think it would make sense to have some slack where even peers transferring should be considered waiting if they have only been transferring for a few seconds?

src/torrent.cpp
Sophist-UK: I have no idea, the ellipsis seems superfluous.

src/torrent.cpp
Sophist-UK: This is set to true if one of the two connections being compared is waiting to download i.e. hasn't downloaded yet.

But we can certainly change it to something else if you can think of a better name.

src/torrent.cpp
Sophist-UK: Agreed.

src/torrent.cpp
Sophist-UK: I will need to check, but you are right - we should NOT prefer a connection which is waiting over one that has only just started downloading. So yes - we should prioritise the downloading peer for (say) twice the peer_turnover_maximum_wait,

src/torrent.cpp
Sophist-UK: Ok.

Sophist-UK (commented):
> I think this patch looks good. I have a few comments on the style though. Also, if you don't mind, I'd appreciate if the comments could be wrapped at 80 columns (it looks better on github among other things).

Can do.

> I would really prefer it if you (or someone else) would actually test this logic before landing it too. If not by writing unit tests, at least by building a client with it and observing its behavior manually.

Can't do - A) only client I use is qBitTorrent; B) no skills in building C+ libs; C) no development environment,

I only know that by when I manually disconnect the slowest actually downloading links, then my total download speed goes up massively.

This is because I get some downloads at several hundred kBps and some measured in single digit kBps or even single digit Bps. Clearly if you disconnect a client that is (say) 100x slower than the fastest, you don't really extend the download time by much, but if you get a significantly faster replacement connection, then you benefit significantly.

BUT, as I have said, I consider using a sort comparison function which can only compare two connections and has no knowledge of current total download speed vs. bandwidth available, nor an understanding of how many seeds there are nor an understanding of how this connection compares to the rest of the connections means that the decision capability is very restricted. We really need a better overall approach, and what this PR does is to tweak the existing algorithm rather than implement a completely different approach.

src/torrent.cpp
Sophist-UK: On reflection I am not sure that 2x maximum wait is long enough to allow a connection to reach its maximum speed.

Perhaps we should use 3x instead.

Sophist-UK:
I am still keen for this PR to be merged. Based on my experiences of manually banning slow IP addresses, I am convinced it will provide a significant performance boost to libtorrent clients.

Sophist-UK:
I am still keen for this PR to be merged. Based on my experiences of manually banning slow IP addresses, I am convinced it will provide a significant performance boost to libtorrent clients.

PR #5001, opened 2020-08-16
Update C++ compile features list

FranciscoPombal:
This is the full feature list that CMake exposes for C++14. Let me know what I should remove any. Fortunately, this is the last time we need to do this dance. For C++17 and beyond, there is only `cxx_std_17`, `cxx_std_20`, ...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -531,45 +531,15 @@ add_library(torrent-rasterbar
 target_compile_features(torrent-rasterbar
 	PUBLIC
 		cxx_std_14
-		cxx_rvalue_references
-		cxx_reference_qualified_functions
-		cxx_nonstatic_member_init
-		cxx_variadic_templates
-		cxx_generalized_initializers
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -531,45 +531,15 @@ add_library(torrent-rasterbar
...
-		cxx_extended_friend_declarations
-		cxx_sizeof_member
-		cxx_inline_namespaces
-		cxx_unrestricted_unions
-		cxx_local_type_template_args
-		cxx_range_for
-		cxx_final
-		cxx_override
-		cxx_noexcept
-		cxx_defaulted_move_initializers
```
arvidn, line 571: are all these removed ones implied by `cxx_std_14`?

FranciscoPombal: @arvidn 

I'd be very surprised if this wasn't the case, though I can't find an authoritative source to confirm for certain.

PR #5004, opened 2020-08-17
use a newer iconv.m4 script

arvidn:


Chocobo1:
Looks like you can get a more up-to-date version here: https://github.com/autotools-mirror/gettext/blob/master/gettext-runtime/m4/iconv.m4

Chocobo1:
BTW, other scripts in the m4 folder are also largely outdated, you mind updating them too, or do you mind I submit a pull request?
Also how long do you plan to keep autotools around since you have cmake build system already?

arvidn:
A pull request would be welcome!

arvidn:
autotools have been removed in 2.0 (so next major release)

PR #5010, opened 2020-08-18
add macro to pick C++11 or C++14 ABI from client

arvidn:


arvidn:
@FranciscoPombal I'm not sure about my cmake logic (just for the pkg-config file), but either way, it lets a *client* say that it knows the library was built with C++11 (by defining `TORRENT_CXX11_ABI`). Which should fix the issue (if set correctly).

In 2.0, where I can change the ABI, I will ensure to encode the C++11 or C++14 API in an inline namespace to improve the error message a bit.

CMakeLists.txt
```diff
@@ -781,6 +781,10 @@ if(developer-options)
 		ENABLED TORRENT_PROFILE_CALLS=1)
 endif()
 
+if (${CMAKE_CXX_STANDARD} STREQUAL "11")
```
FranciscoPombal, line 784: `CMAKE_CXX_STANDARD` does not contain the C++ version value that the compiler will use in all cases. It is only defined if the user defines it in the command line (or if its value is forced in the CML), and its value is used to initialize the `CXX_STANDARD` property of targets (i.e., if `CMAKE_CXX_STANDARD` is defined when a target is created, its `CXX_STANDARD` property is initialized to that value).

I assume what you want is "if building in C++11 mode, use this flag".

I'm lacking in imagination right now to figure out how to do this automatically, but you can always expose it as a feature:

```cmake
feature_option(cxx11_abi "ensure link compatibility with C++11" OFF)

# ....

if (cxx11_abi)
    target_compile_definitions(torrent-rasterbar PUBLIC TORRENT_CXX11_ABI)
endif()
```

FranciscoPombal (changes requested):

CMakeLists.txt
FranciscoPombal: ping @zeule ideas?

include/libtorrent/entry.hpp
```diff
@@ -87,7 +87,7 @@ namespace libtorrent {
 
 namespace aux {
 
-#if (__cplusplus > 201103) || (defined _MSC_VER && _MSC_VER >= 1900)
+#if !defined TORRENT_CXX11_ABI && (__cplusplus > 201103) || (defined _MSC_VER && _MSC_VER >= 1900)
```
FranciscoPombal, line 90: @arvidn 

Well, bad news, this doesn't work anyway, at least not without changes to client code:

```
[  8%] Building CXX object src/base/CMakeFiles/qbt_base.dir/bittorrent/tracker.cpp.o
[build] /home/francisco/Documents/qBittorrent/src/base/bittorrent/tracker.cpp: In member function ‘virtual Http::Response BitTorrent::Tracker::processRequest(const Http::Request&, const Http::Environment&)’:
[build] /home/francisco/Documents/qBittorrent/src/base/bittorrent/tracker.cpp:257:9: error: could not convert ‘{{{anonymous}::ANNOUNCE_RESPONSE_FAILURE_REASON, {(& error)->{anonymous}::TrackerError::<anonymous>.RuntimeError::<anonymous>.std::runtime_error::what()}}}’ from ‘<brace-enclosed initializer list>’ to ‘const dictionary_type’ {aka ‘const libtorrent::aux::map_string<libtorrent::entry>’}
[build]          };
[build]          ^
[build] /home/francisco/Documents/qBittorrent/src/base/bittorrent/tracker.cpp: In member function ‘void BitTorrent::Tracker::prepareAnnounceResponse(const BitTorrent::Tracker::TrackerAnnounceRequest&)’:
[build] /home/francisco/Documents/qBittorrent/src/base/bittorrent/tracker.cpp:400:5: error: no matching function for call to ‘libtorrent::aux::map_string<libtorrent::entry>::map_string(<brace-enclosed initializer list>)’
[build]      };
[build]      ^
[build] In file included from /home/francisco/Documents/qBittorrent/src/base/bittorrent/tracker.h:36,
[build]                  from /home/francisco/Documents/qBittorrent/src/base/bittorrent/tracker.cpp:31:
[build] /usr/local/include/libtorrent/entry.hpp:105:10: note: candidate: ‘libtorrent::aux::map_string<libtorrent::entry>::map_string()’
[build]    struct map_string : std::map<std::string, T>
[build]           ^~~~~~~~~~
[build] /usr/local/include/libtorrent/entry.hpp:105:10: note:   candidate expects 0 arguments, 4 provided
[build] /usr/local/include/libtorrent/entry.hpp:105:10: note: candidate: ‘libtorrent::aux::map_string<libtorrent::entry>::map_string(const libtorrent::aux::map_string<libtorrent::entry>&)’
[build] /usr/local/include/libtorrent/entry.hpp:105:10: note:   candidate expects 1 argument, 4 provided
[build] /usr/local/include/libtorrent/entry.hpp:105:10: note: candidate: ‘libtorrent::aux::map_string<libtorrent::entry>::map_string(libtorrent::aux::map_string<libtorrent::entry>&&)’
[build] /usr/local/include/libtorrent/entry.hpp:105:10: note:   candidate expects 1 argument, 4 provided
[build] /home/francisco/Documents/qBittorrent/src/base/bittorrent/tracker.cpp:438:17: error: could not convert ‘{{{anonymous}::ANNOUNCE_RESPONSE_PEERS_IP, peer.BitTorrent::Peer::address}, {{anonymous}::ANNOUNCE_RESPONSE_PEERS_PORT, peer.BitTorrent::Peer::port}}’ from ‘<brace-enclosed initializer list>’ to ‘libtorrent::entry::dictionary_type’ {aka ‘libtorrent::aux::map_string<libtorrent::entry>’}
[build]                  };
[build]                  ^
[build] src/base/CMakeFiles/qbt_base.dir/build.make:354: recipe for target 'src/base/CMakeFiles/qbt_base.dir/bittorrent/tracker.cpp.o' failed
[build] make[2]: *** [src/base/CMakeFiles/qbt_base.dir/bittorrent/tracker.cpp.o] Error 1
[build] make[2]: *** Waiting for unfinished jobs....
[build] CMakeFiles/Makefile2:236: recipe for target 'src/base/CMakeFiles/qbt_base.dir/all' failed
[build] make[1]: *** [src/base/CMakeFiles/qbt_base.dir/all] Error 2
[build] make: *** [all] Error 2
[build] Makefile:148: recipe for target 'all' failed
```

The offending snippets:

```cpp
        const lt::entry::dictionary_type bencodedEntry = {
            {ANNOUNCE_RESPONSE_FAILURE_REASON, {error.what()}}
        };
```

```cpp
    lt::entry::dictionary_type replyDict {
        {ANNOUNCE_RESPONSE_INTERVAL, ANNOUNCE_INTERVAL},
        {ANNOUNCE_RESPONSE_COMPLETE, torrentStats.seeders},
        {ANNOUNCE_RESPONSE_INCOMPLETE, (torrentStats.peers.size() - torrentStats.seeders)},

        // [BEP-24] Tracker Returns External IP (partial support - might not work properly for all IPv6 cases)
        {ANNOUNCE_RESPONSE_EXTERNAL_IP, toBigEndianByteArray(announceReq.socketAddress).toStdString()}
    };
```

```cpp
                lt::entry::dictionary_type peerDict = {
                    {ANNOUNCE_RESPONSE_PEERS_IP, peer.address},
                    {ANNOUNCE_RESPONSE_PEERS_PORT, peer.port}
                };
```

arvidn: I don't understand how that "offending code" would work with the C++14 type either (or does it only work with the C++11 type?)

When initializing something with a `{}`, normally you have to pass in the exact right types, since it's deducing the top-level type from those. In this case, I assume the all-caps names are `std::string`, but are all the intergers `std::int64_t`?
You may even have to say:
```
lt::entry::dictionary_type replyDict {
        {ANNOUNCE_RESPONSE_INTERVAL, entry(ANNOUNCE_INTERVAL)},
...
```

arvidn: @FranciscoPombal can you describe what you did in your test? I assume it was:

1. Use brew libtorrent binary built with C++11
2. build qBittorrent with C++14 against brew libtorrent headers but define the `TORRENT_CXX11_ABI` macro

The compilation failed, whereas if you had not defined `TORRENT_CXX11_ABI`, compilation would have worked but linking would have failed.

Is that right?

FranciscoPombal: @arvidn

> The compilation failed, whereas if you had not defined TORRENT_CXX11_ABI, compilation would have worked but linking would have failed.

Yes, basically. However, since I don't have local access to an Apple machine, and it's much easier to test this locally, I actually tested on my local machine running Ubuntu 18.04, GCC 8.4.0.

What I did (I fully uninstalled and deleted the build trees of both libtorrent and qBittorrent between tests):

### Test 1

1. Build libtorrent with C++11, without this patch
2. Build qBittorrent with C++14

Result: same linking failure as the Apple system run. This gave me confidence that these results with my local tests are relevant for the resolution of the problem in the Apple machine. After all, we're talking about modern versions of AppleClang and GCC, so they should fail in the same ways as they have identical C++14 support.

### Test 2

1. Build libtorrent with C++11, with an "alternative version" of this patch; basically, due to the concerns outlined in https://github.com/arvidn/libtorrent/pull/5013#issuecomment-675760232, I simply replaced `#if !defined TORRENT_CXX11_ABI && (__cplusplus > 201103) || (defined _MSC_VER && _MSC_VER >= 1900)` with `#if 0` and left the CML alone.
2. Build qBittorrent with C++14

Result: compilation failed, with the error mentioned in https://github.com/arvidn/libtorrent/pull/5010#discussion_r472537544

> I don't understand how that "offending code" would work with the C++14 type either (or does it only work with the C++11 type?)

The following has always worked fine:

1. Build libtorrent with C++14
2. Build qBittorrent with C++14

This "offending code" has been in qBittorrent for a long time and works fine with these steps. Plus, qBittorrent has required C++14 for a long time now (since at least December 2019).

arvidn:
Thanks. This was an old oversight. I wasn't expecting people to use the `dictionary_type` directly. Fixed now.

CMakeLists.txt
arvidn: I don't think making it an option makes sense, since the only thing it would affect would be the `.pc` file. @zeule is there a (better) way to detect whether the compiler is in C++11 mode, and if so, add this define to the `.pc` file?

FranciscoPombal:
@arvidn 

:+1:, these new changes to `entry.hpp` fixed the issue. I re-did "Test 2" above and this time it worked - qBittorrent now compiles and executes successfully. Now the only remaining issue is figuring out how to fix the CML.

CMakeLists.txt
FranciscoPombal: @arvidn 

This is not just important for the `.pc` file, but also for the CMake config files. The `TORRENT_CXX11_ABI` must be propagated up to enable the correct usage of the `entry.hpp` header when using both C++11 and C++14 or later.

arvidn: even if this check doesn't work, the patch still opens up the opportunity for the client to fix issues by defining this macro

FranciscoPombal: @arvidn That's true. I would suggest dropping the change to the CML, and just merge the rest. Or at least adding a comment like `# TODO: won't work properly unless user explicitly passes "-DCMAKE_CXX_STANDARD" to the configure command line`

FranciscoPombal: @arvidn

by the way, you must also enclose the variable in quotes:

`if ("${CMAKE_CXX_STANDARD}" STREQUAL "11")`

 or it will fail when the variable is undefined.

FranciscoPombal (approved):

PR #5011, opened 2020-08-18
optimize character conversions

arvidn:
if the local codepage/locale is already UTF-8, don't perform any character conversion

### DIFF
--- a/src/escape_string.cpp
+++ b/src/escape_string.cpp
@@ -466,6 +466,10 @@ namespace {
 
 	std::string convert_impl(std::string const& s, UINT from, UINT to)
 	{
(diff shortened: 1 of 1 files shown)


NotTsunami:
Is this something that can get backported to RC_1_2?

arvidn:
@NotTsunami It's likely to apply cleanly, do you want to give it a shot?
Do you notice a difference with this change?

NotTsunami:
At a glance it looks like it should apply cleanly. I can cherry-pick and test this patch on top of RC_1_2 and build with qBittorrent master on Windows later in the week, as I'm currently working on a few other projects.

Issue #5012, opened 2020-08-18
Wrong default C++ mode flag in libtorrent-rasterbar.pc


Chocobo1:
**Please provide the following information**

libtorrent version (or branch):
RC_1_2

platform/architecture:
linux/x64

compiler and compiler version:
gcc version 10.1.0
cmake version 3.18.1

please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

This bug is related to PR #4968.
Before that PR, cmake correctly detected the default C++ mode of the compiler (C++14) and generate the correct `Cflags: -fexceptions -std=c++14 ...` in libtorrent-rasterbar.pc (file generated by cmake) .
After PR #4968, it can be seen that libtorrent-rasterbar.pc contains the unexpected mode: `Cflags: -fexceptions -std=c++11 ...`.

The build script used:
```sh
  cd "libtorrent"
  mkdir -p "_build" && cd "_build"
  cmake \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_INSTALL_PREFIX="/usr" \
    -DCMAKE_INSTALL_LIBDIR="lib" \
    -Dpython-bindings=ON \
    -Dboost-python-module-name="python" \
    -Ddeprecated-functions=OFF \
    "../"
  make
```

arvidn:
@FranciscoPombal do you want to address this?

arvidn:
It's possible this is part of the solution too: https://github.com/arvidn/libtorrent/pull/5010
perhaps cmake really likes to build targets with the oldest possible version, regardless of what the dependee is building with. It does seem wrong to propagate C++11 up though..

arvidn:
pkg-config only has the concept of propagating compiler flags, nothing fancy

FranciscoPombal:
@Chocobo1 

Minimal example producing the expected results, using CMake config files instead of pkg-config:

```cmake
cmake_minimum_required(VERSION 3.16 FATAL_ERROR) # Policies <= CMP0097 default to NEW

project(testProject LANGUAGES CXX)

find_package(LibtorrentRasterbar 1.2.9 CONFIG REQUIRED COMPONENTS torrent-rasterbar)

add_executable(test)
target_sources(test PRIVATE main.cpp)
target_link_libraries(test PRIVATE LibtorrentRasterbar::torrent-rasterbar)
```

[...]

PR #5013, opened 2020-08-18
Don't force an -std= flag in the Cflags field of the generated pkg-config file

FranciscoPombal:


arvidn:
ping @zeule See linked ticket for context

zeule:
Why don't you check the compiler default prior to setting C++ mode to a value. lower than that? Why does the client put libtorrent flags after its own settings (otherwise qBt settings would override the C++ mode)?

FranciscoPombal:
@zeule Not sure if this is an issue with qBittorrent (and if so, with which buildsystem), Chocobo1 simply mentioned that it was unexpected that the generated pkgconfig always ends up forcing `-std=gnu++11`. Plus I don't think it's wise to rely on the order of these flags as a potential fix. The `-std` flags are supposed to be propagated in an order-agnostic way, no?

zeule:
Why don't you check the compiler default prior to setting C++ mode to a value, lower than that?

zeule:
Another question: why don't we choose the newest available if user did not request other version explicitly?

FranciscoPombal:
@zeule 

> Why don't you check the compiler default prior to setting C++ mode to a value, lower than that?

What is the context of this question? Building libtorrent, or building something that requires libtorrent? And how would you check the compiler default, without using `CMAKE_CXX_STANDARD_DEFAULT`?

FranciscoPombal:
@zeule 

> Another question: why don't we choose the newest available if user did not request other version explicitly?

I believe that was answered in the linked ticket: https://github.com/arvidn/libtorrent/issues/5012#issuecomment-675639644, https://github.com/arvidn/libtorrent/issues/5012#issuecomment-675706159, https://github.com/arvidn/libtorrent/issues/5012#issuecomment-675712519

zeule:
> What is the context of this question?  

libtorrent.

> And how would you check the compiler default, without using `CMAKE_CXX_STANDARD_DEFAULT`?

If you reject `CMAKE_CXX_STANDARD_DEFAULT`, you can write a simple test program to check the standard value.

FranciscoPombal:
@zeule 

> > What is the context of this question?
> 
> libtorrent.

When building libtorrent, CMake doesn't set the C++ standard to anything specific if it doesn't have to. It uses the compiler default if it can, otherwise it passes `-std=gnu++11` to satisfy the minimum requirement enforced by the call to `target_compile_features()`. The generated CMake config file propagates this minimum requirement up, so that a program that links against libtorrent will only be forced to use `-std=gnu++11` if it would otherwise use something lower.

The problem here is that it is not possible to convey these semantics with pkg-config.
With pkg-config, if you want to have `-std=` in the `Cflags`, you either pass it the value of the compile features (which is always C++11), or somehow determine the compiler default and pass that, which leads us to...

> > And how would you check the compiler default, without using `CMAKE_CXX_STANDARD_DEFAULT`?
> 
> If you reject `CMAKE_CXX_STANDARD_DEFAULT`, you can write a simple test program to check the standard value.

Why should we do that instead of what this PR does (previous justifications here: https://github.com/arvidn/libtorrent/pull/5013#issuecomment-675760232)?

zeule:
Because the libtorrent ABI depends on the C++ standard value used while building the library. That's unfortunate, but we have to export that standard value (and export it as the compiler flag like pkg-config does).

arvidn:
the only place that I know of (apart from possible changes in standard library types, but they try to avoid that) is that `entry` dictionaty type. Which I think should be communicated via pkg-configs cflags as `TORRENT_CXX11_ABI`. But I haven't looked at @FranciscoPombal's report on that patch yet

arvidn:
> Because the libtorrent ABI depends on the C++ standard value used while building the library. That's unfortunate, but we have to export that standard value (and export it as the compiler flag like pkg-config does).

Is the conclusion that @Chocobo1's [observation](https://github.com/arvidn/libtorrent/issues/5012) is an *improvement* over the previous behavior?

PR #5025, opened 2020-08-21
if the local codepage/locale is already UTF-8, don't perform any char…

NotTsunami:
…acter conversion

### DIFF
--- a/src/escape_string.cpp
+++ b/src/escape_string.cpp
@@ -564,6 +564,10 @@ namespace {
 
 	std::string convert_impl(std::string const& s, UINT from, UINT to)
 	{
+		// if the local codepage is already UTF-8, no need to convert
+		static UINT const cp = GetACP();
+		if (cp == 65001 || cp == CP_UTF8) return s;
+
 		std::wstring ws;
 		ws.resize(s.size() + 1);
 		int wsize = MultiByteToWideChar(from, 0, s.c_str(), -1, &ws[0], int(ws.size()));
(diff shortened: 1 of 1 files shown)


src/escape_string.cpp
```diff
@@ -591,8 +595,25 @@ namespace {
...
+
+	bool ends_with(string_view s, string_view suffix)
+	{
+		return s.size() >= suffix.size()
+			&& s.substr(s.size() - suffix.size()) == suffix;
+	}
+
+	bool need_conversion()
+	{
+		static bool const ret = ends_with(std::locale("").name(), ".UTF-8");
```
arvidn, line 608: I just recently discovered that this `std::locale` constructor may throw, which is likely to terminate the process. I will need to fix this in `RC_2_0`. I'm not sure how to fix it yet, since the default constructor seems to always give you the "C" locale.

arvidn:
@NotTsunami if you can figure out a reliable way of asking for the system locale on linux, I would appreciate it. it needs fixing in `RC_2_0` as I mentioned. The only thing this is used for in libtorrent is to find out the character map/code page used for filenames on the system. Most linux distros these days use UTF-8 I believe, so it would seem a bit wasteful to make everyone pay for this conversion when probably quite few use something other than UTF-8.

There are more than one kind of locale too, for different things, `LC_TELEPHONE`, `LC_MONETARY`, `LC_COLLATE` etc. It's not obvious to me which one is used to describe the system code page. By "system code page", I mean whatever `mbtowc()` use to interpret the input character.

Maybe the best fix is to just read the `LANG` environment variable.

What do you think?

arvidn:
there's this one too https://github.com/arvidn/libtorrent/pull/5063

NotTsunami:
Pushed

arvidn (approved):

Issue #5024, opened 2020-08-20
qbittorrent fallout from entry.hpp changes in 1.2.9


namtsui:
**Please provide the following information**

libtorrent version (or branch):

platform/architecture: openbsd amd64

compiler and compiler version: clang++ 10.0.1

please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

I get an error when compiling qbitorrent. I am guessing it is due to this commit because I diffed 1.2.8 and 1.2.9 entry.hpp:
https://github.com/arvidn/libtorrent/commit/0c0671987bbf615da94c8b3e759116d9f62e9f54

It works with 1.2.8 but not 1.2.9.

diff between entry.hpp in 1.2.8 and 1.2.9:
```
--- ../../libtorrent-rasterbar-1.2.8/libtorrent-rasterbar-1.2.8/include/libtorrent/entry.hpp    Tue Aug  4 09:21:21 2020
+++ ./include/libtorrent/entry.hpp      Thu Aug 20 14:04:27 2020
@@ -87,7 +87,7 @@
 
 namespace aux {
 
-#if (__cplusplus > 201103) || (defined _MSC_VER && _MSC_VER >= 1900)
+#if !defined TORRENT_CXX11_ABI && (__cplusplus > 201103) || (defined _MSC_VER && _MSC_VER >= 1900)
                // this enables us to compare a string_view against the std::string that's
                // held by the std::map
                // is_transparent was introduced in C++14
@@ -105,6 +105,9 @@
                struct map_string : std::map<std::string, T>
                {
                        using base = std::map<std::string, T>;
+                       using base::base;
+                       map_string() = default;
+                       map_string(base&& rhs) : base(std::move(rhs)) {}
 
                        typename base::iterator find(const string_view& key)
                        {
```

```
linking qbittorrent

ld: error: undefined symbol:
libtorrent::entry::entry(std::__1::map<std::__1::basic_string<char,
std::__1::char_traits<char>, std::__1::allocator<char> >,
libtorrent::entry, libtorrent::aux::strview_less,
std::__1::allocator<std::__1::pair<std::__1::basic_string<char,
std::__1::char_traits<char>, std::__1::allocator<char> > const,
libtorrent::entry> > >)

>>> referenced by tracker.cpp

>>>
tracker.o:(BitTorrent::Tracker::processRequest(Http::Request const&,
Http::Environment const&))

>>> referenced by tracker.cpp

>>>
[...]

Issue #4999, opened 2020-08-16
Updating iconv / gettext autoconf macros


brad0:
Please sync up to the latest iconv / gettext autoconf macros. The ancient macros shipped have a bug that prevent detection of the shared libs and proper linking on OpenBSD.

arvidn:
where do I find them?
a pull request against the `RC_1_2` branch would be very welcome!

arvidn:
does this look right? https://github.com/arvidn/libtorrent/pull/5004

namtsui:
Thank you for the fix. I am trying to update OpenBSD's port to 1.2.8. I will test out the latest commit with autoconf and report back to see if this goes away during configure:
```
Iconv library:
  Iconv Libs:           /usr/local/lib/libiconv.a
```

Hopefully, this shows up:
```
Iconv library:
  Iconv Libs:           -liconv

```

namtsui:
To test, I modified openbsd's port to run autoreconf on configure.ac. I was having some trouble with the bootstrap.sh script in calling aclocal-1.16.2 when only aclocal-1.16 is available.

It uses /usr/local/lib/libiconv.so.7.0 rather than -liconv, as done with cmake, which is probably correct. I was running autoreconf, and it errored. To resolve, I had to copy build-aux/config.rpath from m4/config.rpath.

This is something to look out for when creating a release tarball for 1.2.9. For now I will propose updating openbsd's port to 1.2.8 to prevent pulling in autotools.

from https://gist.github.com/namtsui/ac5045a6eebf25c98d820145128eb975
```
# error: required file 'build-aux/config.rpath' not found
post-extract:
        cd ${WRKSRC} && mkdir -p build-aux && cp m4/config.rpath build-aux

```

logs of `make configure` and `make update-plist` which worked after testing out autoreconf and this fix (search for libiconv):
https://gist.github.com/namtsui/8deae6bd3e0d143d18f27068eba5cee8
https://gist.github.com/namtsui/21cb5dd8a5d3231d3562bd28b9554b58
```
Iconv
 library:
  Iconv Libs:           /usr/local/lib/libiconv.so.7.0
```


```
c++ -pthread -shared -fPIC -L/usr/local/lib/
-L/usr/obj/ports/Python-3.8.5/Python-3.8.5 -L/usr/local/lib/ -O2
-pipe
[...]

PR #5026, opened 2020-08-21
Use ax_cxx_compile_stdcxx.m4 and move to gnu++14

namtsui:
Use ax_cxx_compile_stdcxx.m4 instead of ax_cxx_compile_stdcxx_11.m4. Before, it defaulted to -std=c++11 (conformance mode). Now, use extended mode (e.g., gnu++14) and newer c++14 standard like cmake.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -144,7 +144,49 @@ AS_ECHO "Checking for boost libraries:"
 
 AX_BOOST_BASE([1.58])
 
-AX_CXX_COMPILE_STDCXX_11([noext], [mandatory])
+AC_ARG_WITH(
+  [cxx-standard],
+  [AS_HELP_STRING(
+    [--with-cxx-standard[=ARG]],
+    [specify C++ standard (e.g., 11, 14 or 17) [default=11]])],
+  [[ARG_CXX_STANDARD=$withval]],
+  [[ARG_CXX_STANDARD=default]]
+)
+
+AC_MSG_CHECKING([which c++ standard to use])
+if test "x${ARG_CXX_STANDARD}" = "x17" ; then
+  AX_CXX_COMPILE_STDCXX(17, [noext], [mandatory])
+elif test "x${ARG_CXX_STANDARD}" = "x14" ; then
+  AX_CXX_COMPILE_STDCXX(14, [noext], [mandatory])
+elif test "x${ARG_CXX_STANDARD}" = "x11" ; then
+  AX_CXX_COMPILE_STDCXX_11([noext], [mandatory])
+elif test "x${ARG_CXX_STANDARD}" != "xdefault" -a "${ARG_CXX_STANDARD}" -lt "11" ; then
+  dnl force 11 if user specified out of bounds, regardless of compiler default
+  ARG_CXX_STANDARD="11"
+  AX_CXX_COMPILE_STDCXX_11([noext], [mandatory])
+elif test "x${ARG_CXX_STANDARD}" != "xdefault" -a "${ARG_CXX_STANDARD}" -gt "17" ; then
+  dnl force 11 if user specified out of bounds, regardless of compiler default
+  ARG_CXX_STANDARD="11"
+  AX_CXX_COMPILE_STDCXX_11([noext], [mandatory])
+else
+  dnl nothing specified so use compiler default
+
+  echo '#include <iostream>' > cpp_standard.cpp
+  echo 'using std::cout; using std::endl; int main() { switch (__cplusplus) { case 201103L: cout << 11 << endl; break; case 201402L: cout << 14 << endl; break; case 201703L: cout << 17 << endl; break; default: cout << 1 << endl; } }' >> cpp_standard.cpp
+  ${CXX} cpp_standard.cpp -o cpp_standard
+  DEFAULT_CXX_STANDARD=`./cpp_standard`
+  echo $DEFAULT_CXX_STANDARD
+  if test "x${DEFAULT_CXX_STANDARD}" = "x17" ; then
+    ARG_CXX_STANDARD="17"
+    AX_CXX_COMPILE_STDCXX(17, [noext], [mandatory])
+  elif test "x${DEFAULT_CXX_STANDARD}" = "x14" ; then
+    ARG_CXX_STANDARD="14"
+    AX_CXX_COMPILE_STDCXX(14, [noext], [mandatory])
+  else
+    ARG_CXX_STANDARD="11"
+    AX_CXX_COMPILE_STDCXX_11([noext], [mandatory])
+  fi
+fi
 
 AX_BOOST_SYSTEM()
 AS_IF([test -z "$BOOST_SYSTEM_LIB"],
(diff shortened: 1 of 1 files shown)


namtsui:
With the new m4 scripts in place, the configure will create a Makefile with -std=c++11 since it uses the older  ax_cxx_compile_stdcxx_11.m4 instead of ax_cxx_compile_stdcxx.m4.

Moving to -std=gnu++14 makes it more in line with cmake.

My concerns are:

- Is it a good idea to impose extended mode with [ext] (gnu++14) instead of conformance mode [noext] (c++14)?
- Is extended mode required by libtorrent-rasterbar?

related:
https://github.com/arvidn/libtorrent/issues/4999
https://github.com/arvidn/libtorrent/issues/5024

arvidn:
I'm not a fan of enabling the gnu extensions. libtorrent does not use any of them (deliberately), so I prefer `c++14` over `gnu++14`

arvidn:
as for the issue of building qBittorrent with C++14 and linking against libtorrent built with C++11. This is caused by an ABI issue I introduced (a bit recklessly), and there's now a patch in 1.2.9 where a client can define `TORRENT_CXX11_ABI` to still be able to link against a C++11-build of libtorrent.

It would probably be best to build libtorrent with C++14 as well, if available.

hhoffstaette:
Hi, it seems Gentoo just stumbled over this problem when trying to build with std=c++14 (see [here](https://bugs.gentoo.org/739654)) and I'd like to request some advice. Is there some quick intermediate fix we can do to make the Python binding observe the imposed C++ ABI until this PR is merged properly?

namtsui:
@hhoffstaette Similarly for my proposed OpenBSD port update I had to decide between:
- sed c++11 c++14 for every Makefile, including bindings/python/compile_cmd and bindings/python/Makefile. I saw that you had noted the compile_cmd file in your Gentoo bug report.
- introducing autotools with hardcoded 14 in configure.ac
```
-AX_CXX_COMPILE_STDCXX_11([noext], [mandatory])
+AX_CXX_COMPILE_STDCXX(14, [noext], [mandatory])
```
Either temporary fix works. I ended up going with introducing autotools. I would like to clean up this draft a bit since it isn't quite where I want. Ideally configure would be able to specify the C++ dialect.

arvidn:
this is still marked as a draft, but it sounds like it has important changes that ought to land for the 1.2.11. Is that right?

eli-schwartz:
The archlinux package maintainer has just imported this patch as-is to fix our distro package. Would be good to have it merged for the next release...

namtsui:
@arvidn Thanks for the ping. This is ready for review now. I vastly simplified it by keeping the m4 macro default. There is now an inline C++ program to test for __cplusplus.

The m4 macro already contains a test, but no harm in testing it twice:
```
// If the compiler admits that it is not ready for C++11, why torture it?
// Hopefully, this will speed up the test.
```

There are some limitations in the implementation though.

- User can't specify --with-cxx-standard=20. This is probably fine because ax_cxx_compile_stdcxx.m4 only supports 11, 14 and 17. In this case, specifying anything >17 or <11 results in 11, like point 4 above.
- Always prefers c++ over gnu++ even if the compiler default is gnu++. This is fine because no gnu extensions are needed.

Responding inline to the flowchart above:
-     If the compiler's default is equal to or greater than c++11/gnu++11, don't add anything, respect the compiler's default.

handled by the else case: `dnl nothing specified so use compiler default`. Still fails this point if given 20a.

-     If the compiler's default is less than c++11/gnu++11, use -std=c++11/-std=gnu++11
handled by the last `AX_CXX_COMPILE_STDCXX_11([noext], [optional])`

-     If the user manually overrides the C++ standard value with -DCMAKE_CXX_STANDARD at configure-time, and this value is greater than or equal to c++11/gnu++11, use it, regardless of compiler default.

handled by first 3 `AX_CXX_COMPILE_STDCXX(17` `AX_CXX_COMPILE_STDCXX(14` `AX_CXX_COMPILE_STDCXX_11`. Still fails this point if given 20a.

-     If the user manually overrides the value with -DCMAKE_CXX_STANDARD at configure-time, and this value is less than c++11/gnu++11, use -std=c++11/-std=gnu++11 instead, regardless of compiler default.

handled by two elif clauses around `dnl force 11 if user specified out of bounds, regardless of compiler default`

Overall, I think this is a good tradeoff between lowering maintenance for libtorrent-rasterbar (by not deviating too far from gnu m4 macros) and emulating cmake to an extent.

namtsui:
@eli-schwartz Can you let archlinux know that the pull request was updated? I am concerned that every time I commit it will break the archlinux package due to not matching checksum.

https://github.com/archlinux/svntogit-packages/blob/packages/libtorrent-rasterbar/trunk/PKGBUILD

namtsui:
C++ standard tester:
```
#include <iostream>

using std::cout;
using std::endl;

int main() {
	switch (__cplusplus) {
		case 201103L:
			cout << 11 << endl;
			break;
		case 201402L:
			cout << 14 << endl;
			break;
		case 201703L:
			cout << 17 << endl;
			break;
		default:
			cout << 1 << endl;
	}
}
```

eli-schwartz:
@felixonmars ^^

namtsui:
noticed a bug where it doesn't respect default compiler when no flag is specified because it falls into:
```
elif test "${ARG_CXX_STANDARD}" -lt "11" ; then
  dnl force 11 if user specified out of bounds, regardless of compiler default
  ARG_CXX_STANDARD="11"
  AX_CXX_COMPILE_STDCXX_11([noext], [mandatory])
elif test "${ARG_CXX_STANDARD}" -gt "17" ; then
  dnl force 11 if user specified out of bounds, regardless of compiler default
  ARG_CXX_STANDARD="11"
  AX_CXX_COMPILE_STDCXX_11([noext], [mandatory])
```

namtsui:
some unit tests:

no --with-cxx-standard specified:
cxx standard:         14
c++ -std=c++14

--with-cxx-standard=14:
cxx standard:         14
c++ -std=c++14

--with-cxx-standard=11:
cxx standard:         11
c++ -std=c++11

17:
cxx standard:         17
c++ -std=c++17

2a:
cxx standard:         14
c++ -std=c++14

98:
cxx standard:         11
c++ -std=c++11

03:
cxx standard:         11
c++ -std=c++11

20:
cxx standard:         11
c++ -std=c++11

emmercm:
I think this PR is still relevant, I don't believe v1.2.12 fixed it.

namtsui:
It is still ready for review. I can answer any questions.

arvidn:
@namtsui oh, I missed that. sorry

arvidn (commented):
works for me on ubuntu

arvidn (approved):

arvidn:
@namtsui Do you want to change the commits or should I just squash them all?
Do you think it makes sense to mention this in the Changelog?

namtsui:
> @namtsui Do you want to change the commits or should I just squash them all?
> Do you think it makes sense to mention this in the Changelog?

Hopefully I edited history correctly. Yeah it could be a good idea to let configure users know of this new --with-cxx-standard=14 flag.

PR #5041, opened 2020-08-22
moved stat.hpp and stat_cache.hpp to aux

aldenml:


arvidn (approved):

arvidn:
`stat` and `stat_cache` are independent classes (as I imagine you figured out). `stat_cache` is useful and `stat` is super old and I would like to remove it (eventually)

PR #5054, opened 2020-08-28
Alloca improvement

arvidn:
attempted allocations larger than 4kiB are made on the heap, instead of the stack


arvidn:
the first commit is just a back-port of a simplification of the code. The new feature is in the second commit

PR #5062, opened 2020-08-29
adriaandegroot's fix for generating pkg-config file in cmake

arvidn:


tambry:
LGTM. The `TARGET` argument based solution should still be implemented otherwise this may crop up again in newer CMake versions in the future.

arvidn:
thanks for your review! This is all above my head. I rely on people that know CMake well to help out.
Would you be up for a PR?

tambry:
@arvidn I'll give it a try.

PR #5069, opened 2020-08-30
GeneratePkgConfig: Handle generator expressions requiring a target

### LINKED ISSUE #4956: CMake 3.18 cannot configure
With CMake 3.18, libtorrent fails to configure:

```
CMake Error at cmake/Modules/GeneratePkgConfig.cmake:181 (file):
  Error evaluating generator expression:

    $<COMPILE_LANG_AND_ID:CUDA,NVIDIA>

  $<COMPILE_LANG_AND_ID:lang,id> may only be used with binary targets to
  specify include directories, compile definitions, and compile options.  It
  may not be used with the add_custom_command, add_custom_target, or
  file(GENERATE) commands.
Call Stack (most recent call first):
  CMakeLists.txt:747 (generate_and_install_pkg_config_file)
```

(on FreeBSD at least, but this looks like a generic CMake problem). There's an upstream report from me at https://gitlab.kitware.com/cmake/cmake/-/issues/21074 and a related report in a different project at https://github.com/facebook/folly/issues/1414

tambry:
CMake's find modules may have target-based generator expressions, which get propagated to us when generating the pkg-config file.
CMake 3.19 introduces the TARGET argument to file(GENERATE), which allows resolving such generator expressions and will avoid future issues.

Backport workaround for #4956 on CMake 3.18 from 1.2.

Fixes #4956.

### DIFF
--- a/cmake/Modules/GeneratePkgConfig.cmake
+++ b/cmake/Modules/GeneratePkgConfig.cmake
@@ -146,10 +146,13 @@ function(generate_and_install_pkg_config_file _target _packageName)
 	set(_generate_target_dir "${CMAKE_CURRENT_BINARY_DIR}/${_target}-pkgconfig")
 	set(_pkg_config_file_template_filename "${_GeneratePkGConfigDir}/pkg-config.cmake.in")
 
-	# Avoid things in the CFLAGS -- which can come from INTERFACE_OPTIONS
-	# of dependencies -- that don't work with file(GENERATE). CMake bug 21074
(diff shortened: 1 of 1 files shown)


arvidn:
I'd prefer to land this patch in `RC_1_2` and then merge it into `RC_2_0`. would that work?

tambry:
@arvidn Sure! Whatever fits your merge/branching strategy best.

arvidn:
the patch against `RC_1_2` is in `RC_2_0` now

PR #5073, opened 2020-08-30
Improve stat_file() performance on Windows.

kasper93:
Use more performant GetFileAttributesExW. Fallback to
GetFileInformationByHandle for symlink support.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* improve stat_file() performance on Windows
 	* fix issue with loading invalid torrents with only 0-sized files
 	* fix to avoid large stack allocations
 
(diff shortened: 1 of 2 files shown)


kasper93:
I noticed that starting is little bit slow. I did quick profile run and `CreateFileW` was taking most of the time. `GetFileAttributesExW` is faster, but doesn't follow symlinks, so fallback to `GetFileInformationByHandle` in case of symlink.

For reference, here is similar issue https://github.com/dotnet/msbuild/issues/2052

arvidn (approved):

arvidn:
I think this deserves an entry in the Chanelog too. I can add it later though, if you don't feel like updating the PR

kasper93:
Added, feel free to change wording if you like.

PR #5078, opened 2020-08-31
fix static linking patch for python binding Jamfile

arvidn:


arvidn:
@joriscarrier I think this additional patch is necessary to get it to work

PR #5089, opened 2020-09-01
modernize entry type to use std::variant

arvidn:


arvidn:
this works on msvc when I build it locally. perhaps the issue of supporting incomplete types in `std::map<>` was recently added nad I have a newer version. I don't see progressing on this until there's a way to have an incomplete type as value in `std::map<>`, which I don't believe the standard requires, it's just QoI

PR #5108, opened 2020-09-06
fix issue where pad blocks would skip the zero-hash check

arvidn:
and consider the last piece as passed, which in turn could cuase a hybrid magnet link to fail with inconsistent-hashes

### DIFF
--- a/src/merkle_tree.cpp
+++ b/src/merkle_tree.cpp
@@ -336,6 +336,7 @@ namespace {
 			int const num_leafs = 1 << base;
 
 			bool done = false;
+			if (first_leaf >= m_num_blocks) break;
 			for (int j = 0; j < std::min(num_leafs, m_num_blocks - first_leaf); ++j)
 			{
 				if (m_tree[file_first_leaf + first_leaf + j].is_all_zeros())


src/merkle_tree.cpp
```diff
@@ -336,6 +336,7 @@ namespace {
 			int const num_leafs = 1 << base;
 
 			bool done = false;
+			if (first_leaf >= m_num_blocks) break;
```
arvidn, line 339: @ssiloti does this look right to you?
What happened was that `first_leaf` > `m_num_blocks` (i.e. it pointed into the pad-blocks). The condition below (with `std::min(num_leafs, m_num_blocks - first_leaf)`) caused that loop to be skipped entirely, which in turn then continued down to the code where we assumed that we had all non-zero block hashes, and that we should add the piece to `passed_piece`. This despite not checking any blocks. This in turn caused a magnet link for a hybrid torrent to fail, because if we get a piece in `passed_pieces` but the piece hasn't passed, and we have v1 metadata, it's an inconsistency.

Anyway, I believe another fix would be to remove the clamping to `m_num_blocks` because at this point `m_tree` should be the *full* tree, including all padding blocks. Maybe that would be a more "regular" fix, but it seems like the optimization to skip the pad block was intended here already.

ssiloti (approved):
LGTM

PR #5110, opened 2020-09-07
fix error category in http_tracker_connection

kasper93:


kasper93:
I noticed that error is bogus when ipv6 endpoint cannot reach ipv4 only tracker. Turns out to be silly error category mismatch. Interestingly it is correct in udp code path. 

Alternatively we could steal ec from asio, but I don't think it is accurate, because it is our own generated error.
``` c++
fail(boost::asio::error::host_unreachable, operation_t::get_interface);
```

arvidn (approved):

PR #5116, opened 2020-09-08
fix msvc 64 bit build

arvidn:


arvidn:
@5523eeOMEGALUL I could reproduce the issue you posted and this patch fixes it for me now. release, 64 bit, C++17 and openssl now builds on msvc

PR #5129, opened 2020-09-10
disable use of SetFileValidData() by default (windows).

arvidn:
A new setting allows enabling it.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,6 @@
+	* disable use of SetFileValidData() by default (windows). A new setting
+	  allows enabling it
+
 2.0 release
 
 	* dropped depenency on iconv
--- a/docs/hunspell/libtorrent.dic
+++ b/docs/hunspell/libtorrent.dic
@@ -586,3 +586,4 @@ libgnutls
 0x200000
 golang
 img
+SetFileValidData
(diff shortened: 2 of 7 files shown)


MOZGIII (approved):
LGTM

arvidn:
@FranciscoPombal @cas-- 
Do you think it's reasonable to change this default?
Would you enable `SetFileValidData()` by default in qbt and Deluge?

MOZGIII:
For the record, I'm a qbt user too, and it looks like with my daily usage it's not active either way because I'm not running qbt with admin permissions.
From the perspective of the `libtorrent` user - as long as it's configurable I don't care what the default is :D

cas--:
@arvidn Considering the security issue, I think it's a wise move. Deluge would not expose or change the setting, instead would leave it to power users to optionally change in a 'settings pack' plugin

FranciscoPombal:
@arvidn 

I agree with leaving it OFF by default in libtorrent. qBittorrent is going to leave it hard coded to ON for now, to preserve the previous behavior for people who are running qBittorrent as administrator: https://github.com/qbittorrent/qBittorrent/pull/13355. I believe we are open to change this and/or expose it as an advanced setting in the future, though.

PR #5144, opened 2020-09-13
moved debug.hpp and file.hpp to aux

aldenml:


arvidn:
I'm working on removing `file.hpp` and `file.cpp` altogether

aldenml:
OK, closing this then

PR #5146, opened 2020-09-13
fix calculation in tracker_manager::num_requests when webtorrent is enable

aldenml:


aldenml:
@arvidn, I want to add a unit test for this, but didn't find tests that interact directly with `tracker_manager`, should I create a `test_tracker_manager.cpp`?

arvidn:
sure. A unit test would be ideal. It may be simpler to test in a simulation

aldenml:
Hi @arvidn, I decided to go with a regular test (with minimal hack to queue the udp request)

arvidn (approved):

Issue #5121, opened 2020-09-09
save_resume_data() should include pieces waiting to be written to disk, for paused torrents


AllSeeingEyeTolledEweSew:
**Please provide the following information**

libtorrent version (or branch): 1.2.8.0

platform/architecture: linux/x86_64

compiler and compiler version: gcc / unknown version, using alpine linux's edge testing build

I'm not sure if this is a bug report or clarification request.

The [docs](https://libtorrent.org/reference-Torrent_Handle.html) say:

>*flush_disk_cache*
>the disk cache will be flushed before creating the resume data. This avoids a problem with file timestamps in the resume data in case the cache hasn't been flushed yet.

Based on https://github.com/arvidn/libtorrent/blob/RC_1_2/src/torrent.cpp#L8560 , it seems like `save_resume_data(flush_disk_cache)` just asynchronously flushes the cache. It seems equivalent to
```
handle.save_resume_data()
handle.flush_cache()  # without generating cache_flushed_alert
```

Since the `add_torrent_params` is generated immediately after firing the flush, it's not clear to me the flag can affect the generated params at all.

In particular, I see that in the following sequence:
```
status = handle.status(flags=libtorrent.torrent_handle.query_pieces)
handle.save_resume_data(flags=libtorrent.torrent_flags.flush_disk_cache)
params = wait_for_save_resume_data_alert().params
```
If the torrent is actively downloading, there are cases where `states.pieces` has some "bits" set, but `params.have_piece` does **not** have those bits set.

The fields are built using piece picker code which is pretty hard for me to follow, but my understanding is that:
* `torrent_status::pieces` marks pieces which have been hashed
* `add_torrent_params::have_pieces` marks pieces which are hashed and written to disk

So it would be nice if `flush_disk_cache` behaved as described in the docs, but it doesn't do what I expect.

Is `flush_disk_cache` working as intended? If it is, what should I expect from it, and what's the best practice to ensure my resume data is up-to-date at shutdown?

arvidn:
this flag is an old left-over from the days when the resume data contained modification timestamps for all files. flushing the cache first would make it slightly less likely that the resume data would have a stale timestamp in it. That's what the flag would do, it would ensure that the flushing happened immediately before `stat()`ing the file to get the modification timestamp.

I don't think it is very useful anymore. As you point out, you can flush the cache manually, and it doesn't need to be synchronized with generating the resume data.

arvidn:
[...]

PR #5151, opened 2020-09-15
Update and clarify docs around torrent_handle::is_valid(). Closes #5112

### LINKED ISSUE #5112: save_resume_data_alert occurs after torrent_removed_alert

libtorrent version (or branch): 1.2.8.0

platform/architecture: linux/x86_64

compiler and compiler version: not sure; using alpine linux's edge testing binary

I'm seeing a case where if I do:
```
session.remove_torrent(handle)
handle.save_resume_data()
```
Then:
1. This succeeds without raising an exception (no `RuntimeError` in Python)
1. In the alert stream, I see `torrent_removed_alert` come before `save_resume_data_alert` for the same handle
1. When I receive `save_resume_data_alert`, then `alert.handle.is_valid()` is true, and `alert.handle.info_hash()` is non-all-zeros.

I sort of understand thing 1, as both calls are asynchronous, so they can both get queued before either is called.

Thing 2 follows from thing 1.

I am kind of confused by thing 3 though. My C++ is weak, but my understanding is that `handle::is_valid()` should return false once the torrent has been destructed. I don't understand exactly where it gets destructed in this case though.

The following python script triggers this condition. On my machine the last lines of output are like this:
```
torrent_removed_alert test.txt removed
torrent_log_alert test.txt: *** UPDATE LIST [ torrent_want_tick : 0 ]
torrent_log_alert test.txt: *** announce: no trackers
torrent_log_alert test.txt: aborting
torrent_log_alert test.txt: *** UPDATE LIST [ torrent_downloading_auto_managed
: 0 ]
save_resume_data_alert test.txt resume data generated
... info_hash = 9393ec7b02a264f71506301963429f064a472327 valid = True
```

[srda_timing.py.gz](https://github.com/arvidn/libtorrent/files/5181847/srda_timing.py.gz)

The big practical problem for me is figuring out how to serialize resume data correctly. If I write resume data when I see `save_resume_data_alert` and delete it when I see `torrent_removed_alert`, I end up with zombie data in this case. At the moment the only workaround I can think of is also watching `add_torrent_alert` and `torrent_removed_alert` to keep a list of all active handles, so I can properly ignore `save_resume_data_alert` after `torrent_removed_alert`. However this violates the "don't keep application state for every torrent" principle outlined in [your blog post](https://blog.libtorrent.org/2012/01/seeding-a-million-torrents/).

Do you have any guidance here? How do I figure out when the torrent is really removed?

AllSeeingEyeTolledEweSew:
Assuming my understanding in #5112 is right, I'd like to update the documentation around `torrent_handle::is_valid()` and `torrent_removed_alert`. If it's not right, please let me know :)

I wasn't sure if the current practice for things like this is to merge into RC_1_2 or RC_2_0. Let me know and I can fix

Closes: #5112.

### DIFF
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -109,6 +109,9 @@ TORRENT_VERSION_NAMESPACE_2
 
 	// This is a base class for alerts that are associated with a
 	// specific torrent. It contains a handle to the torrent.
+	//
+	// Note that by the time the client receives a torrent_alert, its
+	// ``handle`` member may be invalid.
 	struct TORRENT_EXPORT torrent_alert : alert
 	{
 		// internal
@@ -247,9 +250,17 @@ TORRENT_VERSION_NAMESPACE_2
 #endif
 
 	// The ``torrent_removed_alert`` is posted whenever a torrent is removed. Since
-	// the torrent handle in its base class will always be invalid (since the torrent
+	// the torrent handle in its base class will usually be invalid (since the torrent
 	// is already removed) it has the info hash as a member, to identify it.
 	// It's posted when the ``status_notification`` bit is set in the alert_mask.
+	//
+	// Note that the ``handle`` remains valid for some time after
+	// torrent_removed_alert is posted, as long as some internal libtorrent
+	// task (such as an I/O task) refers to it. Additionally, other alerts like
+	// save_resume_data_alert may be posted after torrent_removed_alert.
+	// To synchronize on whether the torrent has been removed or not, call
+	// torrent_handle::in_session(). This will return true before
+	// torrent_removed_alert is posted, and false afterward.
 	//
 	// Even though the ``handle`` member doesn't point to an existing torrent anymore,
 	// it is still useful for comparing to other handles, which may also no
@@ -1336,17 +1347,15 @@ TORRENT_VERSION_NAMESPACE_2
 	// .. code:: c++
 	//
 	//	torrent_handle h = alert->handle();
-	//	if (h.is_valid()) {
-	//		std::shared_ptr<torrent_info const> ti = h.torrent_file();
-	//		create_torrent ct(*ti);
-	//		entry te = ct.generate();
-	//		std::vector<char> buffer;
-	//		bencode(std::back_inserter(buffer), te);
-	//		FILE* f = fopen((to_hex(ti->info_hash().to_string()) + ".torrent").c_str(), "wb+");
-	//		if (f) {
-	//			fwrite(&buffer[0], 1, buffer.size(), f);
-	//			fclose(f);
-	//		}
+	//	std::shared_ptr<torrent_info const> ti = h.torrent_file();
+	//	create_torrent ct(*ti);
+	//	entry te = ct.generate();
+	//	std::vector<char> buffer;
+	//	bencode(std::back_inserter(buffer), te);
+	//	FILE* f = fopen((to_hex(ti->info_hash().to_string()) + ".torrent").c_str(), "wb+");
+	//	if (f) {
+	//		fwrite(&buffer[0], 1, buffer.size(), f);
+	//		fclose(f);
 	//	}
 	//
 	struct TORRENT_EXPORT metadata_received_alert final : torrent_alert
--- a/include/libtorrent/session_handle.hpp
+++ b/include/libtorrent/session_handle.hpp
@@ -811,12 +811,15 @@ namespace libtorrent {
 		// the swarm. This operation cannot fail. When it completes, you will
 		// receive a torrent_removed_alert.
 		//
+		// remove_torrent() is non-blocking, but will remove the torrent from the
+		// session synchronously. Calling session_handle::add_torrent() immediately
+		// afterward with the same torrent will succeed. Note that this creates a
+		// new handle which is not equal to the removed one.
+		//
 		// The optional second argument ``options`` can be used to delete all the
 		// files downloaded by this torrent. To do so, pass in the value
-		// ``session_handle::delete_files``. The removal of the torrent is asynchronous,
-		// there is no guarantee that adding the same torrent immediately after
-		// it was removed will not throw a system_error exception. Once
-		// the torrent is deleted, a torrent_deleted_alert is posted.
+		// ``session_handle::delete_files``. Once the torrent is deleted, a
+		// torrent_deleted_alert is posted.
 		//
 		// Note that when a queued or downloading torrent is removed, its position
 		// in the download queue is vacated and every subsequent torrent in the
@@ -824,6 +827,18 @@ namespace libtorrent {
 		// large state_update to be posted. When removing all torrents, it is
 		// advised to remove them from the back of the queue, to minimize the
 		// shifting.
+		//
+		// The torrent_handle remains valid for some time after remove_torrent() is
+		// called. It will become invalid only after all libtorrent tasks (such as
+		// I/O tasks) release their references to the torrent. Until this happens,
+		// torrent_handle::is_valid() will return true, and other calls such
+		// as torrent_handle::status() will succeed. Because of this, and because
+		// remove_torrent() is non-blocking, the following sequence usually
+		// succeeds (does not throw system_error):
+		// .. code:: c++
+		//
+		//	session.remove_handle(handle);
+		//	handle.save_resume_data();
 		void remove_torrent(const torrent_handle& h, remove_flags_t options = {});
 
 		// Applies the settings specified by the settings_pack ``s``. This is an
--- a/include/libtorrent/torrent_handle.hpp
+++ b/include/libtorrent/torrent_handle.hpp
@@ -545,12 +545,20 @@ namespace aux {
 
 		// Returns true if this handle refers to a valid torrent and false if it
 		// hasn't been initialized or if the torrent it refers to has been
-		// aborted. Note that a handle may become invalid after it has been added
-		// to the session. Usually this is because the storage for the torrent is
-		// somehow invalid or if the filenames are not allowed (and hence cannot
-		// be opened/created) on your filesystem. If such an error occurs, a
-		// file_error_alert is generated and all handles that refers to that
-		// torrent will become invalid.
+		// removed from the session AND destructed.
+		//
+		// To tell if the torrent_handle is in the session, use
+		// torrent_handle::in_session(). This will return true before
+		// session_handle::remove_torrent() is called, and false
+		// afterward.
+		//
+		// Clients should only use is_valid() to determine if the result of
+		// session::find_torrent() was successful.
+		//
+		// Unlike other member functions which return a value, is_valid()
+		// completes immediately, without blocking on a result from the
+		// network thread. Also unlike other functions, it never throws
+		// the system_error exception.
 		bool is_valid() const;
 
 		// will delay the disconnect of peers that we're still downloading
@@ -694,14 +702,17 @@ namespace aux {
 		//	extern int outstanding_resume_data; // global counter of outstanding resume data
 		//	std::vector<torrent_handle> handles = ses.get_torrents();
 		//	ses.pause();
-		//	for (torrent_handle const& h : handles)
+		//	for (torrent_handle const& h : handles) try
 		//	{
-		//		if (!h.is_valid()) continue;
 		//		torrent_status s = h.status();
 		//		if (!s.has_metadata || !s.need_save_resume_data()) continue;
 		//
 		//		h.save_resume_data();
 		//		++outstanding_resume_data;
+		//	}
+		//	catch (lt::system_error const& e)
+		//	{
+		//		// the handle was invalid, ignore this one and move to the next
 		//	}
 		//
 		//	while (outstanding_resume_data > 0)
@@ -1267,6 +1278,13 @@ namespace aux {
 		// i.e. the ``TORRENT_`` macros must match between libtorrent and the
 		// client builds.
 		std::shared_ptr<torrent> native_handle() const;
+
+		// Returns true if the torrent is in the session. It returns true before
+		// session::remove_torrent() is called, and false afterward.
+		//
+		// Note that this is a blocking function, unlike torrent_handle::is_valid()
+		// which returns immediately.
+		bool in_session() const;
 
 	private:
 
--- a/src/torrent_handle.cpp
+++ b/src/torrent_handle.cpp
@@ -863,6 +863,9 @@ namespace libtorrent {
 		return std::size_t(*reinterpret_cast<void* const*>(&th.m_torrent));
 	}
 
+	bool torrent_handle::in_session() const
+	{ return !sync_call_ret<bool>(false, &torrent::is_aborted); }
+
 	static_assert(std::is_nothrow_move_constructible<torrent_handle>::value
 		, "should be nothrow move constructible");
 	static_assert(std::is_nothrow_move_assignable<torrent_handle>::value
--- a/test/test_torrent.cpp
+++ b/test/test_torrent.cpp
@@ -891,3 +891,16 @@ TORRENT_TEST(redundant_add_piece)
 	h.add_piece(piece_index_t{0}, piece_data.data());
 	std::this_thread::sleep_for(lt::seconds(2));
 }
+
+TORRENT_TEST(test_in_session)
+{
+	lt::session ses(settings());
+	std::shared_ptr<torrent_info> ti = generate_torrent();
+	add_torrent_params p;
+	p.ti = ti;
+	p.save_path = ".";
+	torrent_handle h = ses.add_torrent(p);
+	TEST_CHECK(h.in_session());
+	ses.remove_torrent(h);
+	TEST_CHECK(!h.in_session());
+}


include/libtorrent/alert_types.hpp
```diff
@@ -247,9 +250,19 @@ TORRENT_VERSION_NAMESPACE_2
...
 	// is already removed) it has the info hash as a member, to identify it.
 	// It's posted when the ``status_notification`` bit is set in the alert_mask.
+        //
+        // Note that the ``handle`` may stay valid for an arbitrary time after
+        // ``torrent_removed_alert`` is posted, as long as some internal
+        // libtorrent task (such as an I/O task) refers to it. Additionally,
+        // other alerts like ``save_resume_data_alert`` may be posted after
+        // ``torrent_removed_alert``. To synchronize on whether the torrent has
+        // been removed or not, use the expression
+        // ``session.find_torrent(handle.info_hash()) == handle``. This
```
arvidn, line 263: Would it be better if the `torrent_handle` was always invalid after the torrent has been removed?
It would be quite simple to make that happen. Then this "hack" wouldn't be necessary, but would anything break in that case?

AllSeeingEyeTolledEweSew: Interesting. Would the session ignore calls for invalidated torrents?

What happens in the case of
```python
session.remove_torrent(handle)
handle.save_resume_data()
```
Presumably `save_resume_data()` doesn't throw because they're both asynchronous, but then how does `save_resume_data[_failed]_alert` get generated?

include/libtorrent/torrent_handle.hpp
```diff
@@ -542,12 +542,20 @@ namespace aux {
...
+                // valid until after the last task (such as an I/O task)
+                // referring to the torrent has completed.
+                //
+                // To synchronize on whether the handle is in the session, use
+                // the expression ``session.find_torrent(handle.info_hash()) ==
+                // handle``. This expression will return true before
+                // ``torrent_removed_alert`` is posted, and false afterward.
+                //
+                // Clients should only use ``is_valid()`` to determine if the
+                // result of ``session::find_torrent()`` was successful.
```
arvidn, line 558: the indentation isn't right here. Also, maybe it would be better to just make all handles in alerts posted after torrent removal be invalid, and document that. I think it would be a simpler behavior at least, and probably more intuitive. What do you think?

include/libtorrent/torrent_handle.hpp
```diff
@@ -693,7 +701,6 @@ namespace aux {
 		//	ses.pause();
 		//	for (torrent_handle const& h : handles)
 		//	{
-		//		if (!h.is_valid()) continue;
```
arvidn, line 696: the problem with removing this is that using an invalid handle later will throw an exception. Technically, this is a problem with the existing code too, since the handle may become invalid at any time. But maybe a better way to do this would be to add a try-catch block to the for-loop, like this:

```
for (torrent_handle const& h : handles) try
{
   torrent_status s = h.status();
   if (!s.has_metadata || !s.need_save_resume_data()) continue;
   ...
}
catch (lt::system_error const& e)
{
   // the handle was invalid, ignore this one and move on to the next
}
```

arvidn:
would you mind filing a ticket for the other issue you brought up? Where a paused torrent that still has outstanding blocks to be hash-checked should synchronize on checking (and maybe flushing) those blocks before generating the resume data, to not have to re-download anything. in the next session. It should also be fixed in `RC_1_2` I think

AllSeeingEyeTolledEweSew:
#5121 

I retitled it. Should I close and file a different one?

arvidn:
oh, great!

AllSeeingEyeTolledEweSew:
Pushing this discussion to the top level
>maybe it would be better to just make all handles in alerts posted after torrent removal be invalid, and document that. I think it would be a simpler behavior at least, and probably more intuitive. What do you think?

If you can invalidate all handles in `remove_torrent()` without changing any other behavior, I think that would be really useful and intuitive.

I'm trying to guess at implementation though my C++ isn't that strong, and I could see some of these behaviors changing:
* That `is_valid()` is non-blocking
* That the following sequence unconditionally produces `save_resume_data[_failed]_alert`, if it succeeds:
```python
session.remove_torrent(handle)
handle.save_resume_data()
```

arvidn:
what I meant was that all *new* handles would be invalid. i.e. the ones you get in alerts, after removing the torrent. Maybe that doesn't make a difference.

When do you need to synchronize and wait for a handle to become invalid?

arvidn:
`is_valid()` is already non-blocking. `handle` is technically not guaranteed to be valid right after it's removed (but in practice it always will be). Can you explain why it would be helpful to be guaranteed to be able to save resume data *after* removing it?

AllSeeingEyeTolledEweSew:
> what I meant was that all _new_ handles would be invalid. i.e. the ones you get in alerts, after removing the torrent. Maybe that doesn't make a difference.

Ah, I see. I guess there's a benefit to an app that wants to figure out if an alert came before or after `torrent_removed_alert`, doesn't want to keep per-torrent state, and wants to minimize blocking calls (they would call `session.find_torrent(info_hash) == handle` today). The app would need to be aware of the libtorrent version where this behavior changed though.

I think this would affect single-threaded event loop apps more than concurrent apps though. A concurrent app either doesn't mind the blocking `find_torrent()` call, or is better set up to run it in a thread pool when necessary, and in any case needs to deal with its existing handles becoming invalid.

The docs would get more complicated, trying to explain the expected return value of `is_valid()` and why it is that way.

I think it's a pretty surprising behavior for a pretty narrow benefit, so I vote no.

> When do you need to synchronize and wait for a handle to become invalid?

I don't currently; I can't think of a case.

> `is_valid()` is already non-blocking. `handle` is technically not guaranteed to be valid right after it's removed (but in practice it always will be). Can you explain why it would be helpful to be guaranteed to be able to save resume data _after_ removing it?

I meant that unconditionally posting the alert is helpful today, so that the shutdown-time resume data code can synchronize. It can always expect to see N `save_resume_data[_failed]_alert`s after making N successful (i.e. didn't throw) calls to `handle.save_resume_data()`. I guess that doesn't change in your proposal though.

FWIW, if there was some other way to synchronize on "libtorrent is not going to post any more alerts" (true sometime after `session::pause()`), then counting resume data alerts probably wouldn't be necessary.

arvidn:
my understanding is that there are exactly 3 outcomes of this code:
```
session.remove_torrent(handle)
handle.save_resume_data()
```

1. you get a `save_resume_alert`
2. you get a `save_resume_failed_alert`
3. the call to `handle.save_resume_data()` throws an exception

I think case 3 would naturally not increment your counter of how many alerts you expect to receive. But is there some other outcome?

AllSeeingEyeTolledEweSew:
> my understanding is that there are exactly 3 outcomes of this code:
> 
> ```
> session.remove_torrent(handle)
> handle.save_resume_data()
> ```
> 
>     1. you get a `save_resume_alert`
> 
>     2. you get a `save_resume_failed_alert`
> 
>     3. the call to `handle.save_resume_data()` throws an exception
> 
> 
> I think case 3 would naturally not increment your counter of how many alerts you expect to receive. But is there some other outcome?

No other outcome. I agree with what you said. I was trying to make sure your proposal wouldn't _change_ that behavior. It doesn't.

arvidn:
ok, good. In what circumstance do you need to do `session.find_torrent(handle.info_hash()) == handle`?
That is to determine whether the torrent has been removed or not, right? even though your handle may still be valid

AllSeeingEyeTolledEweSew:
> In what circumstance do you need to do `session.find_torrent(handle.info_hash()) == handle`?

I currently have two uses:
1. Handling `save_resume_data_alert`. If the torrent was removed, I need to not persist the data. Actually, it would be nice if `save_resume_data()` posted `save_resume_data_failed_alert` after the torrent has been removed. This is pretty un-surprising, and current code should handle it. If this means that `session::remove_torrent()` would set some flag on the torrent, I think it's reasonable to have it exposed in `torrent_status::flags`, as this would cover other cases.
2. When creating alert iterators (as I mention in #5154), a common case is to create an iterator for only alerts having a particular `torrent_handle`. There's a race that the iterator will miss the `torrent_removed_alert` and will wait forever, so I check using the `find_torrent()` call. More code [here](https://github.com/AllSeeingEyeTolledEweSew/tvaf/blob/master/tvaf/driver.py). Currently I call `find_torrent()` in a thread pool so it doesn't block alert handling, and it would be nice to call it inline instead, but that would require #5154.

I can't think of an alternative to checking `session.find_torrent(handle.info_hash()) == handle` in case 2. Because of this, I'm starting to think of the check as a "weird idiom", and not as a "hack". It's just a blocking call to synchronize on whether the torrent was removed, and maybe we always need that.

I think a more readable form would be nice though, like `session::has_handle(torrent_handle)` or `handle.status().flags & torrent_flags::removed`.

arvidn:
I did not mean to pass judgment by calling it a "hack". I just meant that it's an indirect way of finding something out that maybe there should be a direct way to ask (has this torrent been removed).

It's not obvious to me that it is generally true that resume data should be discarded after a torrent has been removed.

I made this change, but it's not clear to me that it helps anyone. I don't think it helps with either of the two cases you bring up: https://github.com/arvidn/libtorrent/pull/5156

AllSeeingEyeTolledEweSew:
> I did not mean to pass judgment by calling it a "hack". I just meant that it's an indirect way of finding something out that maybe there should be a direct way to ask (has this torrent been removed).

Agree. A name like `torrent_handle::is_removed()` seems most readable for my scenarios above. What do you think?

> It's not obvious to me that it is generally true that resume data should be discarded after a torrent has been removed.

In this case, I thought an app should just keep the last resume data they got, and _not_ delete it on `torrent_removed_alert`.

A `save_resume_data_alert` occurring after `torrent_removed_alert` seems like something else. It's resume data generated for a torrent that's been aborted, and is on its way out. I assumed it was potentially stale or invalid. Is it guaranteed to still be meaningful? I don't see a test case for this.

> I made this change, but it's not clear to me that it helps anyone. I don't think it helps with either of the two cases you bring up: #5156

If you want feedback, I don't think it's worth making this change, due to the surprise of `(handle_a == handle_b) != (handle_a.is_valid() == handle_b.is_valid())` (where one is from an alert and one is not), and due to the limited upside. Since it's such a simple change we could just do it later when someone presents their own use case for it.

If you intend to proceed anyway, I'll update this PR to reflect the new use of `is_valid()`.

arvidn:
here's, what I believe to be, a fix for including "writing" blocks in the resume data (i.e. blocks that are waiting to be flushed to disk). My argument for why it's safe to include these, even if there's a disk failure after saving the resume data is that these blocks all have to be re-checked anyway, so there's no risk of data corruption.

https://github.com/arvidn/libtorrent/pull/5160

AllSeeingEyeTolledEweSew:
(assuming your comment was meant for #5121)

I see you deleted #5156, so I guess my docs on `torrent_handle::is_valid()` are accurate for now.

Should I implement `torrent_handle::is_removed()`, so the docs can reference that instead of `session.find_torrent(handle.info_hash()) == handle`? I think it's probably simple enough that I can write and test it. Do you have thoughts on the name?

arvidn:
please submit a PR for `is_removed()` on `torrent_handle`. I think you can use `m_abort` or `is_aborted()` on the `torrent` object. You still have to post the call to the network thread and do a blocking call though, since it's not an atomic variable.

It might be possible to *turn* it into an atomic to avoid that message-passing cost too,

AllSeeingEyeTolledEweSew:
> It might be possible to _turn_ it into an atomic to avoid that message-passing cost too,

I'd be curious to know how this would work.

As much as I love performance enhancements, I can't think of a case where this would be a high priority.

AllSeeingEyeTolledEweSew:
I merged it with this PR, so the doc updates all make sense together.

I ended up calling it `torrent_handle::is_in_session()`; the "positive" version was easier to explain in documentation, which I take as a hint that it's the right version of it.

include/libtorrent/torrent_handle.hpp
```diff
@@ -542,12 +542,25 @@ namespace aux {
...
-		// be opened/created) on your filesystem. If such an error occurs, a
-		// file_error_alert is generated and all handles that refers to that
-		// torrent will become invalid.
+		// removed from the session and destructed.
+		//
+		// A handle does not immediately become invalid after
+		// ``session::remove_torrent()`` completes, or after
+		// ``torrent_removed_alert`` is posted. It will remain valid until after
+		// the last task (such as an I/O task) referring to the torrent has
+		// completed.
```
arvidn, line 551: it seems this paragraph belongs to the `remove_torrent()` documentation.

include/libtorrent/torrent_handle.hpp
```diff
@@ -691,15 +704,18 @@ namespace aux {
...
 		//		torrent_status s = h.status();
 		//		if (!s.has_metadata || !s.need_save_resume_data()) continue;
 		//
 		//		h.save_resume_data();
 		//		++outstanding_resume_data;
 		//	}
+		//	catch (lt::system_error const& e)
+		//	{
+		//		// the handle was invalid, ignore this one and move to the next
+		//	}
```
arvidn, line 718: :+1:

include/libtorrent/torrent_handle.hpp
```diff
@@ -1265,6 +1281,24 @@ namespace aux {
...
+		// return true, and other operations such as ``torrent_handle::status()``
+		// will succeed.
+		//
+		// ``torrent_handle::is_in_session()`` can be used to figure out if
+		// the call to ``session::remove_torrent()`` has been made. It returns true
+		// before ``torrent_removed_alert`` is posted, and false afterward.
+		//
+		// Note that this function posts a message to the libtorrent control thread
+		// and waits for the result, unlike ``is_valid()`` which returns instantly.
+		bool is_in_session() const;
```
arvidn, line 1300: I think just `in_session()` would be a better name.

include/libtorrent/torrent_handle.hpp
```diff
@@ -1265,6 +1281,24 @@ namespace aux {
 		// client builds.
 		std::shared_ptr<torrent> native_handle() const;
 
+		// Returns true if the torrent is active in its session; returns false if
```
arvidn, line 1284: I think this sentence is a bit misleading. The torrent doesn't have to be "active". It just have to *not* have been removed

include/libtorrent/torrent_handle.hpp
```diff
@@ -1265,6 +1281,24 @@ namespace aux {
...
+		// it has been removed.
```
arvidn, line 1285: it's probably a good idea to specify that "removed" means `session::remove_torrent()` has been called

include/libtorrent/torrent_handle.hpp
```diff
@@ -1265,6 +1281,24 @@ namespace aux {
...
+		//
+		// A ``torrent_handle`` remains valid for some time after
```
arvidn, line 1287: if you don't put types like `torrent_handle` and functions like `session::remove_torrent()` in two quotes ""`", they will turn into links in the documentation, which I think is better.

include/libtorrent/torrent_handle.hpp
```diff
@@ -1265,6 +1281,24 @@ namespace aux {
...
+		// ``session::remove_torrent()`` is called. It will become invalid only
+		// after all libtorrent tasks (such as I/O tasks) release their references
+		// to the torrent. Until this happens, ``torrent_handle::is_valid()`` will
+		// return true, and other operations such as ``torrent_handle::status()``
+		// will succeed.
```
arvidn, line 1292: this paragraph also seem to belong in the documentation for `remove_torrent()`

include/libtorrent/torrent_handle.hpp
```diff
@@ -1265,6 +1281,24 @@ namespace aux {
...
+		//
+		// ``torrent_handle::is_in_session()`` can be used to figure out if
+		// the call to ``session::remove_torrent()`` has been made. It returns true
+		// before ``torrent_removed_alert`` is posted, and false afterward.
```
arvidn, line 1296: although this is correct, it might make the reader think that it should return false until I *receive* `torrent_removed_alert`. I think specifying that it will return false until the call to `session::remove_torrent()` is made is simpler and just as correct. Since `is_in_session()` still is synchronous, it will always synchronize with a previous call to remove the torrent

include/libtorrent/torrent_handle.hpp
```diff
@@ -1265,6 +1281,24 @@ namespace aux {
...
+		//
+		// Note that this function posts a message to the libtorrent control thread
+		// and waits for the result, unlike ``is_valid()`` which returns instantly.
```
arvidn, line 1299: the thread is referred to as the "network thread", but I think there's precedence for documenting these kinds of calls. I think they're just called "blocking"

include/libtorrent/alert_types.hpp
```diff
@@ -247,10 +250,19 @@ TORRENT_VERSION_NAMESPACE_2
...
+	// the torrent handle in its base class will usually be invalid (since the torrent
 	// is already removed) it has the info hash as a member, to identify it.
 	// It's posted when the ``status_notification`` bit is set in the alert_mask.
 	//
+	// Note that the ``handle`` may stay valid for an arbitrary time after
+	// ``torrent_removed_alert`` is posted, as long as some internal libtorrent 
+	// task (such as an I/O task) refers to it. Additionally, other alerts like
+	// ``save_resume_data_alert`` may be posted after ``torrent_removed_alert``.
+	// To synchronize on whether the torrent has been removed or not, check
+	// ``session.find_torrent(handle.info_hash()) == handle``. This expression
```
arvidn, line 262: this should reference the new function instead

arvidn:
types and functions should not be wrapped in "``" in the documentation, that prevents them from turning into links. However, variable names should still be literals.

AllSeeingEyeTolledEweSew:
https://github.com/arvidn/libtorrent/blob/867cf863f21747f2df7290df81a8d6a57a4d0992/include/libtorrent/session_handle.hpp#L808-L810

Is it true any more that `remove_torrent()` followed by `add_torrent()` of the same torrent may fail? It appears that only the data deletion is asynchronous.

arvidn:
If the torrent is actively downloaded, it's possible that the removal waits for those outstanding operations. But I haven't checked the code. It definitely delays deleting the torrent object, but it may still remove it from the session immediately.

AllSeeingEyeTolledEweSew:
Updated per your comments

include/libtorrent/session_handle.hpp
```diff
@@ -811,19 +811,33 @@ namespace libtorrent {
 		// the swarm. This operation cannot fail. When it completes, you will
 		// receive a torrent_removed_alert.
 		//
+		// remove_torrent() is non-blocking, but will remove the torrent from the
+		// session synchronously. Calling session_handle::add_torrent() immediately
+		// afterward with the same torrent will succeed.
```
arvidn, line 816: maybe it would make sense to mention here that adding the torrent back immediately will still return a different `torrent_handle` (i.e. not resurrect the previous one, even if it was still valid)

include/libtorrent/alert_types.hpp
```diff
@@ -247,10 +250,18 @@ TORRENT_VERSION_NAMESPACE_2
 #endif
 
 	// The ``torrent_removed_alert`` is posted whenever a torrent is removed. Since
-	// the torrent handle in its base class will always be invalid (since the torrent
+	// the torrent handle in its base class will usually be invalid (since the torrent
 	// is already removed) it has the info hash as a member, to identify it.
 	// It's posted when the ``status_notification`` bit is set in the alert_mask.
 	//
+	// Note that the ``handle`` may stay valid for an arbitrary time after
```
arvidn, line 257: I'm not sure about "arbitrary" here. Technically it isn't arbitrary, it's bounded by the session destruction for instance. I think the sentence in the other paragraph where you say this seem better:

"The torrent_handle remains valid for some time after remove_torrent() is called"

include/libtorrent/torrent_handle.hpp
```diff
@@ -1268,6 +1279,13 @@ namespace aux {
 		// client builds.
 		std::shared_ptr<torrent> native_handle() const;
 
+		// Returns true if the torrent is in the session. It returns true before
+		// session::remove_torrent() is called, and false afterward.
+		//
+		// Note that this is a blocking function, unlike torrent_handle::is_valid()
+		// which returns instantly.
```
arvidn, line 1286: I think "immediately" is a better fit than "instantly".

arvidn:
sorry for the delay. I think this looks good, but with a few comments

AllSeeingEyeTolledEweSew:
Updated per your comments.

arvidn:
thanks!

PR #5159, opened 2020-09-17
fix invariant when setting the piece deadline of a piece with priority 0

arvidn:


arvidn:
I should add a unit test for this case as well

PR #5173, opened 2020-09-20
add a new torrent flag to not validate resume data against files on disk

arvidn:


arvidn:
one problem with this patch is that if mutable torrents are used, the hard links will not be created with this flag.

arvidn:
on the other hand, that's really only something that's expected to happen the *first* time a torrent is added, in which case this flag is not expected to be set

PR #5175, opened 2020-09-21
post save_resume_data_failed_alert if torrent was removed. Closes #5174

### LINKED ISSUE #5174: save_resume_data() should post save_resume_data_failed_alert if the torrent is removed/aborted
Forking one of the topics from #5151 

Consider:
```python
session.remove_torrent(handle)
handle.save_resume_data()
```
This _usually_ succeeds, because the requests are asynchronous and the handle isn't invalidated by the time the second call is made. This results in `torrent_removed_alert` followed by `save_resume_data_alert`.

However, I'm struggling to think of a use case for the data in such a `save_resume_data_alert`. In the common case, an application deletes the resume data for any removed torrent. So a concurrent app must detect this sequence somehow, and _avoid_ persisting data in this case.

Even if the app _wants_ to persist data after removal, it would surprise me if it expected the above sequence to be valid.

In any case, it's not obvious the data generated by this sequence should be valid or meaningful today (it's generated for a torrent that's been aborted; what does that even mean?), or that we can guarantee going forward that it stays valid and meaningful.

It's essential for the above sequence to either fail or generate some alert, since the prescribed shutdown sequence looks like this:
```python
for handle in session.get_torrents():
    try:
        handle.save_resume_data()
        outstanding += 1
    except:
        pass

for alert in iter_alerts():
    if isinstance(alert, (save_resume_data_alert, save_resume_data_failed_alert)):
        outstanding -= 1
    if isinstance(alert, save_resume_data_alert):
        write_atp(alert.params)
    if outstanding == 0:
        break
```

So, I propose that the sequence in question generate `save_resume_data_failed_alert` instead.

I'll submit a PR shortly.

AllSeeingEyeTolledEweSew:
Closes: #5174.

### DIFF
--- a/test/test_resume.cpp
+++ b/test/test_resume.cpp
@@ -1712,3 +1712,30 @@ TORRENT_TEST(resume_data_have_pieces)
 	TEST_EQUAL(rs->params.unfinished_pieces.size(), 1);
 }
 
+// See https://github.com/arvidn/libtorrent/issues/5174
+TORRENT_TEST(removed)
+{
+	lt::session ses(settings());
+	std::shared_ptr<torrent_info> ti = generate_torrent();
+	add_torrent_params p;
+	p.ti = ti;
+	p.save_path = ".";
+	// we're _likely_ to trigger the condition, but not guaranteed. loop
+	// until we do.
+	bool triggered = false;
+	for (int i = 0; i < 10; i++) {
+		torrent_handle h = ses.add_torrent(p);
+		// this is asynchronous
+		ses.remove_torrent(h);
+		try {
+			h.save_resume_data();
+			triggered = true;
+		} catch (std::exception const&) {
+			std::printf("failed to trigger condition, retrying\n");
+		}
+	}
+	TEST_CHECK(triggered);
+	if (!triggered) return;
+	alert const* a = wait_for_alert(ses, save_resume_data_failed_alert::alert_type);
+	TEST_CHECK(a != nullptr);
+}
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -8566,6 +8566,13 @@ bool is_downloading_state(int const st)
 		TORRENT_ASSERT(is_single_thread());
 		INVARIANT_CHECK;
 
+		if (m_abort)
+		{
+			alerts().emplace_alert<save_resume_data_failed_alert>(get_handle()
+				, errors::torrent_removed);
+			return;
+		}
+
 		if ((flags & torrent_handle::only_if_modified) && !m_need_save_resume_data)
 		{
 			alerts().emplace_alert<save_resume_data_failed_alert>(get_handle()


test/test_resume.cpp
```diff
@@ -1690,3 +1690,42 @@ TORRENT_TEST(resume_data_have_pieces)
 	TEST_EQUAL(rs->params.unfinished_pieces.size(), 1);
 }
 
+// See https://github.com/arvidn/libtorrent/issues/5174
+TORRENT_TEST(removed)
+{
+	// we're _likely_ to trigger the condition, but not guaranteed. loop
+	// until we do.
+	while (true) {
```
arvidn, line 1698: it's risky to have an infinite loop in a test. What if you run this on a machine that never triggers the condition? Or under a build configuration that doesn't use a separate thread for disk I/O. Then everything is deterministic, and it may not trigger this condition.

test/test_resume.cpp
```diff
@@ -1690,3 +1690,42 @@ TORRENT_TEST(resume_data_have_pieces)
...
+		// this is asynchronous
+		ses.remove_torrent(h);
+		try {
+			h.save_resume_data();
+		} catch (std::exception const& e) {
+			std::printf("failed to trigger condition, retrying\n");
+			continue;
+		}
+		// we triggered the condition.
+		while (true)
```
arvidn, line 1714: I don't think this infinite loop is a good idea either. I think it's better to try for some amount of time, and failing if the condition doesn't happen within some reasonable timeout. (in cases where we expect to always see the condition that is)

test/test_resume.cpp
```diff
@@ -1690,3 +1690,42 @@ TORRENT_TEST(resume_data_have_pieces)
...
+		{
+			std::vector<alert*> alerts;
+			alert const* ap = ses.wait_for_alert(seconds(1));
+			TEST_CHECK(ap != nullptr);
+			if (ap == nullptr) return;
```
arvidn, line 1719: I suppose it's not infinite, since it exists here. I think it would be better to type this out as a loop with this as the termination condition, to make it easier to read.

test/test_resume.cpp
AllSeeingEyeTolledEweSew: In such a configuration, does the libtorrent network thread still exist? Are the `async_call()`s actually synchronous?

arvidn: the network thread exists, but no other thread (this is the case when building with posix disk I/O in 2.0). So anything that may be asynchronous/non-deterministic because of the disk I/O will be deterministic in that configuration.

AllSeeingEyeTolledEweSew: I actually thought the condition would still get triggered because the `torrent` is kept alive by a pointer somewhere in the session's `io_service`.

If the sequence _doesn't_ trigger the condition (`save_resume_data()` doesn't generate any alerts), how will the caller get notified of this? Didn't it already successfully submit the call?

arvidn: I would expect that any call to save_resume_data() would result in exactly one response alert or an exception. Still, if there's a bug, it would be better with an explicit test failure than a hung test (that won't produce any output on CI)

AllSeeingEyeTolledEweSew:
Updated per your comments.

arvidn (approved):

PR #5176, opened 2020-09-21
fix issue in add_piece() called multiple times

arvidn:


AllSeeingEyeTolledEweSew:
I tested this, it fixes my test case. Thanks!

Issue #5100, opened 2020-09-04
Assertion when using WebTorrent


vktr:
libtorrent version (or branch): master (ea183a16b1cb340235b3ff920aaed498bd667ebf)
platform/architecture: win64
compiler and compiler version: msvc-19

I'm trying to verify WebTorrent support in PicoTorrent and am hitting this assertions here,

```
TORRENT_ASSERT(i_ == m_connections.end()
	|| (*i_)->type() != connection_type::bittorrent);
```

This only happens for WebTorrents. When using regular torrents (with WebTorrent support enabled) I don't hit the assertion. I'm using the Sintel torrent from here, https://webtorrent.io/torrents/sintel.torrent)

Here's the call stack (MSVC style so frames are in descending order)

```
> PicoTorrent.exe!libtorrent::assert_fail(const char * expr, int line, const char * file, const char * function, const char * value, int kind) Line 362	C++
PicoTorrent.exe!libtorrent::aux::torrent::connect_to_peer(libtorrent::aux::torrent_peer * peerinfo, bool ignore_limit) Line 7056	C++
PicoTorrent.exe!libtorrent::aux::torrent::try_connect_peer() Line 10639	C++
PicoTorrent.exe!libtorrent::aux::session_impl::try_connect_more_peers() Line 4223	C++
PicoTorrent.exe!libtorrent::aux::session_impl::on_tick(const boost::system::error_code & e) Line 3524	C++
PicoTorrent.exe!libtorrent::aux::session_impl::wrap<void (__cdecl libtorrent::aux::session_impl::*)(boost::system::error_code const &),boost::system::error_code const &>(void(libtorrent::aux::session_impl::*)(const boost::system::error_code &) f, const boost::system::error_code & <a_0>) Line 530	C++
PicoTorrent.exe!libtorrent::aux::session_impl::on_tick::__l2::<lambda>(const boost::system::error_code & err) Line 3288	C++
PicoTorrent.exe!libtorrent::aux::allocating_handler<void <lambda>(const boost::system::error_code &),144,3>::operator()<boost::system::error_code const &>(const boost::system::error_code & <a_0>) Line 245	C++
PicoTorrent.exe!boost::asio::detail::binder1<libtorrent::aux::allocating_handler<void <lambda>(const boost::system::error_code &),144,3>,boost::system::error_code>::operator()() Line 66	C++
[...]

PR #5184, opened 2020-09-25
WebTorrent:  Fix peer list sorting

paullouisageneau:
This PR addresses a regression introduced by WebTorrent support https://github.com/arvidn/libtorrent/pull/4123: when running with both webRTC and normal peers, the peer list sorting is broken, because it mixes peers sorted by peer id and peers sorted by remote address. This PR resolves the issue by using the remote address for webRTC peers too.

It fixes the duplicate peer issue https://github.com/arvidn/libtorrent/issues/5100

### DIFF
--- a/include/libtorrent/aux_/torrent_peer.hpp
+++ b/include/libtorrent/aux_/torrent_peer.hpp
@@ -220,13 +220,13 @@ namespace libtorrent::aux {
 #if TORRENT_USE_RTC
 	struct TORRENT_EXTRA_EXPORT rtc_peer : torrent_peer
 	{
-		rtc_peer(string_view pid_, peer_source_flags_t src);
+		rtc_peer(tcp::endpoint const& ep, peer_source_flags_t src);
 		rtc_peer(rtc_peer const&) = delete;
 		rtc_peer& operator=(rtc_peer const&) = delete;
 		rtc_peer(rtc_peer&&) = default;
 		rtc_peer& operator=(rtc_peer&&) & = default;
 
-		aux::string_ptr pid;
+		tcp::endpoint endpoint;
 	};
 #endif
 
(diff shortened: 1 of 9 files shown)


arvidn:
there's another instance of libtorrent building with higher warning level + warnings-as-errors than the dependencies:

https://ci.appveyor.com/project/arvidn/libtorrent/builds/35414064/job/k41tld3xy116peyg#L355

arvidn:
it might actually be a good idea to treat updating of submodules as separate changes (unless there's a dependency on something new in the submodule).

include/libtorrent/aux_/torrent_peer.hpp
```diff
@@ -264,16 +264,14 @@ namespace libtorrent::aux {
...
-#endif
-			)
-				return lhs->dest().compare(rhs->dest()) < 0;
+			if (lhs->is_i2p_addr || rhs->is_i2p_addr)
+			{
+				if(lhs->is_i2p_addr && rhs->is_i2p_addr)
+					return lhs->dest().compare(rhs->dest()) < 0;
+				else
+					return lhs->is_i2p_addr;
+			}
```
arvidn, line 274: is this restoring this to its original form?

arvidn:
I suppose having access to the remote peer's IP address is a new feature in libdatachannel, and that's why the submodule needs to be updated. is that right?

paullouisageneau:
@arvidn No it's not a new feature, actually I should have done the change before but it slipped my mind. Let me open a separate PR for updating libdatachannel.

include/libtorrent/aux_/torrent_peer.hpp
paullouisageneau: It is now. It would break if there is a mix a I2P peers and normal peers for a torrent, but it seems that's never supposed to happen.

paullouisageneau:
I extracted the libdatachannel update in PR https://github.com/arvidn/libtorrent/pull/5188

arvidn (approved):

PR #5185, opened 2020-09-26
replaced boost::optional with std::optional

aldenml:


bindings/python/src/optional.hpp
```diff
@@ -6,19 +6,19 @@
...
     optional_to_python()
     {
         boost::python::to_python_converter<
-            boost::optional<T>, optional_to_python<T>
+            std::optional<T>, optional_to_python<T>
         >();
     }
 
-    static PyObject* convert(boost::optional<T> const& x)
+    static PyObject* convert(std::optional<T> const& x)
```
arvidn, line 21: these functions, taking `optional` are part of the boost.python interface. I don't think this would work

include/libtorrent/aux_/http_connection.hpp
```diff
@@ -27,7 +27,6 @@ see LICENSE file.
 #include "libtorrent/aux_/socket_type.hpp"
 #include "libtorrent/aux_/vector.hpp"
 #include "libtorrent/aux_/resolver_interface.hpp"
-#include "libtorrent/optional.hpp"
```
arvidn, line 30: you need to add `#include <optional>` here now, right?

include/libtorrent/aux_/http_parser.hpp
```diff
@@ -23,7 +23,6 @@ see LICENSE file.
 #include "libtorrent/span.hpp"
 #include "libtorrent/string_view.hpp"
 #include "libtorrent/time.hpp" // for seconds32
-#include "libtorrent/optional.hpp"
```
arvidn, line 26: here too, I think you'd need `#include <optional>`

bindings/python/src/optional.hpp
aldenml: looking at other converters, it looks to me that they don't depend on boost types, also I expect to have some unit test failing if it doesn't work, but I'm not familiar with the python binding, can you help me to get this one right?

arvidn: I would also expect a unit test failing. and I think I was wrong, the `optional` here is the type being converted

aldenml: Ok, I will wait until appveyor is done and squash

arvidn (approved):

PR #5186, opened 2020-09-26
add missing update_storage_node_ids

arvidn:
to keep the DHT storage up to date when the m_nodes list is updated.

@aldenml would you mind taking a look at this? I think this was an oversight. I ended up encountering an assert where `m_nodes` had 1 entry in it but the storage object (`dht_default_storage`) had an empty list of `m_node_ids`


aldenml (approved):
LGTM

aldenml:
I also looked for other places, these seem to be the only one missing.

PR #5196, opened 2020-10-03
using structured binding

aldenml:


arvidn:
Looks good. However, one of the benefits of structured bindings is that you can make the variables const. I think there may be a few missed opportunities for that.

aldenml:
agree, do you want me to review every `auto [` for adding a `const`?

arvidn:
At least the ones you just added. I'm pretty sure some of them can be const now (but couldn't earlier, when using tie)

aldenml:
OK, added `const` when possible

PR #5198, opened 2020-10-05
Windows Phone/UWP compilation fixes

tigerw:
+ Added a workaround for missing LoadLibraryA
+ Added missing network definitions
+ Use CreateFile2
- Pre-process out calls to unavailable functions

### DIFF
--- a/include/libtorrent/aux_/win_util.hpp
+++ b/include/libtorrent/aux_/win_util.hpp
@@ -38,6 +38,10 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent { namespace aux {
 
+#ifdef TORRENT_WINRT
+	using LoadLibraryASignature = HMODULE WINAPI(_In_ LPCSTR lpLibFileName);
+#endif
+
 	template <typename Library>
 	HMODULE get_library_handle()
 	{
@@ -46,8 +50,26 @@ namespace libtorrent { namespace aux {
 
 		if (!handle_checked)
 		{
-			handle = LoadLibraryA(Library::library_name);
 			handle_checked = true;
+
+#ifdef TORRENT_WINRT
+			MEMORY_BASIC_INFORMATION Information;
+
+			if (::VirtualQuery(&VirtualQuery, &Information, sizeof(Information)) == 0)
+			{
+				return nullptr;
+			}
+
+			const auto SyscallBegin = static_cast<HMODULE>(Information.AllocationBase);
+			const auto LoadLibraryA = reinterpret_cast<LoadLibraryASignature *>(::GetProcAddress(SyscallBegin, "LoadLibraryA"));
+
+			if (LoadLibraryA == nullptr)
+			{
+				return nullptr;
+			}
+#endif
+
+			handle = LoadLibraryA(Library::library_name);
 		}
 		return handle;
 	}
--- a/src/enum_net.cpp
+++ b/src/enum_net.cpp
@@ -83,6 +83,9 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/aux_/windows.hpp"
 #include <iphlpapi.h>
 #include <ifdef.h> // for IF_OPER_STATUS
+#ifdef TORRENT_WINRT
+#include <netioapi.h>
+#endif
 #endif
 
 #if TORRENT_USE_NETLINK
--- a/src/mmap.cpp
+++ b/src/mmap.cpp
@@ -181,6 +181,38 @@ namespace {
 		return (mode & open_mode::write) ? OPEN_ALWAYS : OPEN_EXISTING;
 	}
 
+#ifdef TORRENT_WINRT
+
+	DWORD file_flags(open_mode_t const mode)
+	{
+		return ((mode & open_mode::no_cache) ? FILE_FLAG_WRITE_THROUGH : 0)
+			| ((mode & open_mode::random_access) ? 0 : FILE_FLAG_SEQUENTIAL_SCAN)
+			;
+	}
+
+	DWORD file_attributes(open_mode_t const mode)
+	{
+		return (mode & open_mode::hidden) ? FILE_ATTRIBUTE_HIDDEN : FILE_ATTRIBUTE_NORMAL;
+	}
+
+	auto create_file(const native_path_string & name, open_mode_t const mode)
+	{
+		CREATEFILE2_EXTENDED_PARAMETERS Extended
+		{
+			sizeof(CREATEFILE2_EXTENDED_PARAMETERS),
+			file_attributes(mode),
+			file_flags(mode)
+		};
+
+		return CreateFile2(name.c_str()
+			, file_access(mode)
+			, FILE_SHARE_WRITE | FILE_SHARE_READ | FILE_SHARE_DELETE
+			, file_create(mode)
+			, &Extended);
+	}
+
+#else
+
 	DWORD file_flags(open_mode_t const mode)
 	{
 		// one might think it's a good idea to pass in FILE_FLAG_RANDOM_ACCESS. It
@@ -192,18 +224,25 @@ namespace {
 			;
 	}
 
+	auto create_file(const native_path_string & name, open_mode_t const mode)
+	{
+		return CreateFileW(name.c_str()
+			, file_access(mode)
+			, FILE_SHARE_WRITE | FILE_SHARE_READ | FILE_SHARE_DELETE
+			, nullptr
+			, file_create(mode)
+			, file_flags(mode)
+			, nullptr);
+	}
+
+#endif
+
 } // anonymous
 
 file_handle::file_handle(string_view name, std::int64_t const size
 	, open_mode_t const mode)
-	: m_fd(CreateFileW(convert_to_native_path_string(name.to_string()).c_str()
-		, file_access(mode)
-		, FILE_SHARE_WRITE | FILE_SHARE_READ | FILE_SHARE_DELETE
-		, nullptr
-		, file_create(mode)
-		, file_flags(mode)
-		, nullptr))
-		, m_open_mode(mode)
+	: m_fd(create_file(convert_to_native_path_string(name.to_string()), mode))
+	, m_open_mode(mode)
 {
 	if (m_fd == invalid_handle) throw_ex<system_error>(error_code(GetLastError(), system_category()));
 
@@ -226,6 +265,8 @@ file_handle::file_handle(string_view name, std::int64_t const size
 
 		if (::SetEndOfFile(m_fd) == FALSE)
 			throw_ex<system_error>(error_code(GetLastError(), system_category()));
+
+#ifndef TORRENT_WINRT
 		// Enable privilege required by SetFileValidData()
 		// https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-setfilevaliddata
 		std::call_once(g_once_flag, acquire_manage_volume_privs);
@@ -234,6 +275,7 @@ file_handle::file_handle(string_view name, std::int64_t const size
 		// the file with zeroes, but just fill it with
 		// garbage instead
 		SetFileValidData(m_fd, size);
+#endif
 	}
 }
 
--- a/src/path.cpp
+++ b/src/path.cpp
@@ -165,6 +165,44 @@ namespace {
 		// convert to seconds
 		return time_t(ft / 10000000 - posix_time_offset);
 	}
+
+	void fill_file_status(file_status & s, LARGE_INTEGER file_size, DWORD file_attributes, FILETIME creation_time, FILETIME last_access, FILETIME last_write)
+	{
+		s.file_size = file_size.QuadPart;
+		s.ctime = file_time_to_posix(creation_time);
+		s.atime = file_time_to_posix(last_access);
+		s.mtime = file_time_to_posix(last_write);
+
+		s.mode = (file_attributes & FILE_ATTRIBUTE_DIRECTORY)
+			? file_status::directory
+			: (file_attributes & FILE_ATTRIBUTE_DEVICE)
+			? file_status::character_special : file_status::regular_file;
+	}
+
+	void fill_file_status(file_status & s, DWORD file_size_low, DWORD file_size_high, DWORD file_attributes, FILETIME creation_time, FILETIME last_access, FILETIME last_write)
+	{
+		LARGE_INTEGER file_size;
+		file_size.HighPart = static_cast<LONG>(file_size_high);
+		file_size.LowPart = file_size_low;
+
+		fill_file_status(s, file_size, file_attributes, creation_time, last_access, last_write);
+	}
+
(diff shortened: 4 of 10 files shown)


include/libtorrent/aux_/win_util.hpp
```diff
@@ -38,16 +38,40 @@ POSSIBILITY OF SUCH DAMAGE.
...
+#ifdef TORRENT_WINRT
+	using LoadLibraryAPrototype = HMODULE WINAPI(_In_ LPCSTR lpLibFileName);
+#endif
+
 	template <typename Library>
 	HMODULE get_library_handle()
 	{
 		static bool handle_checked = false;
-		static HMODULE handle = nullptr;
+		static HMODULE handle;
```
arvidn, line 49: this doesn't look like an improvement. is there a reason not to initialize this?

include/libtorrent/aux_/win_util.hpp
```diff
@@ -38,16 +38,40 @@ POSSIBILITY OF SUCH DAMAGE.
...
+
+#ifdef TORRENT_WINRT
+			MEMORY_BASIC_INFORMATION Information;
+
+			if (::VirtualQuery(&VirtualQuery, &Information, sizeof(Information)) == 0)
+			{
+				return nullptr;
+			}
+
+			const auto SyscallBegin = static_cast<HMODULE>(Information.AllocationBase);
```
arvidn, line 63: so this is, essentially, the handle to `kernel32` then?

include/libtorrent/aux_/win_util.hpp
```diff
@@ -38,16 +38,40 @@ POSSIBILITY OF SUCH DAMAGE.
...
+			const auto LoadLibraryA = reinterpret_cast<LoadLibraryAPrototype *>(::GetProcAddress(SyscallBegin, "LoadLibraryA"));
+
+			if (LoadLibraryA == nullptr)
+			{
+				return nullptr;
+			}
+
+			handle = LoadLibraryA(Library::library_name);
```
arvidn, line 71: did you observe this working in a WinRT build? Do `advapi32.dll`, `kernel32.dll` and `iphlpapi.dll` exist in that environment?

If not, and this will always fail, it might make sense to `#ifdef` this away higher up, where this utility is called

src/enum_net.cpp
```diff
@@ -83,6 +83,11 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/aux_/windows.hpp"
 #include <iphlpapi.h>
 #include <ifdef.h> // for IF_OPER_STATUS
+#ifdef TORRENT_WINRT
+using NDIS_MEDIUM = int;
+using NDIS_PHYSICAL_MEDIUM = int;
```
arvidn, line 88: if these don't exist, does it work to enumerate routes and adapters at all?

src/mmap.cpp
```diff
@@ -230,10 +268,12 @@ file_handle::file_handle(string_view name, std::int64_t const size
 		// https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-setfilevaliddata
 		std::call_once(g_once_flag, acquire_manage_volume_privs);
 
+#ifndef TORRENT_WINRT
```
arvidn, line 271: you should probably move this `#ifndef` above the call to `acquire_manage_volume_privs` too, since that's only there to support `SetFileValidData()`

src/path.cpp
```diff
@@ -206,6 +206,14 @@ namespace {
 		// Fallback to GetFileInformationByHandle for symlinks
 		if (!(flags & dont_follow_links) && (data.dwFileAttributes & FILE_ATTRIBUTE_REPARSE_POINT))
 		{
+#ifdef TORRENT_WINRT
+
+			ec.assign(ERROR_FILE_SYSTEM_LIMITATION, system_category());
```
arvidn, line 211: is it better to hard-code this error than to let `CreateFile()` fail? Are you guaranteed that this will fail in WinRT?

arvidn (commented):
looks good. just a few comments. I think it would be great to rebase this on top of https://github.com/arvidn/libtorrent/pull/5199 (once it has ran and we can observe the expected failures)

arvidn:
it seems asio doesn't build as a store or phone app (or maybe both). Do you use a patched boost for this to work?

tigerw:
Thanks for the reviews!

I haven't patched Boost. I noticed the AppVeyor build is trying to build with the C++/CX dialect? The `/ZW` flag is set and there's output like (with the carets):
```
client_test.cpp(1129): warning C4447: 'main' signature found without threading model. Consider using 'int main(Platform::Array<Platform::String^>^ args)'.
```

I'm building with vanilla C++17 on Visual Studio 2017, flags below:

Compiler
```
/JMC /permissive- /MP /GS /analyze- /W3 /Zc:wchar_t /I"LibTorrent\include" /I"LibTorrent\deps\try_signal" /I"Boost" /I"...Native\" /I"Generated Files\" /I"...Debug\" /ZW:nostdlib /ZI /Gm- /Od /sdl /Fd"...Debug\vc141.pdb" /Zc:inline /fp:precise /D "WIN32" /D "BOOST_ASIO_ENABLE_CANCELIO" /D "BOOST_ASIO_NO_DEPRECATED" /D "TORRENT_BUILDING_LIBRARY" /D "_FILE_OFFSET_BITS=64" /D "BOOST_EXCEPTION_DISABLE" /D "BOOST_ASIO_HAS_STD_CHRONO" /D "WIN32_LEAN_AND_MEAN" /D "BOOST_ALL_NO_LIB" /D "_SCL_SECURE_NO_DEPRECATE" /D "_CRT_SECURE_NO_DEPRECATE" /D "TORRENT_ABI_VERSION=3" /D "TORRENT_DISABLE_ENCRYPTION" /D "_SILENCE_ALL_CXX17_DEPRECATION_WARNINGS" /D "_WIN32_WINNT=0x0A00" /D "_WINDLL" /D "_UNICODE" /D "UNICODE" /D "_DEBUG" /D "WINAPI_FAMILY=WINAPI_FAMILY_APP" /D "__WRL_NO_DEFAULT_LIB__" /errorReport:prompt /WX- /Zc:forScope /RTC1 /Gd /Oy- /FU"D:\Program Files (x86)\Microsoft Visual Studio\VC\Tools\MSVC\14.16.27023\lib\x86\store\references\platform.winmd" /MDd /std:c++latest /FC /Fa"...Debug\" /EHsc /nologo /Fo"...x86\Debug\" /Fp"...Native.pch" /diagnostics:classic 
```

Linker
```
/OUT:"...Native.dll" /MANIFEST:NO /NXCOMPAT /PDB:"...Native.pdb" /DYNAMICBASE "bcrypt.lib" "Iphlpapi.lib" "WindowsApp.lib" /DEBUG /DLL /MACHINE:X86 /WINMD:NO /APPCONTAINER /INCREMENTAL /PGD:"...Native.pgd" /WINMDFILE:"...Native.winmd" /SUBSYSTEM:WINDOWS /MANIFESTUAC:NO /ManifestFile:"...Debug\Native.dll.intermediate.manifest" /ERRORREPORT:PROMPT /NOLOGO /TLBID:1 
```

include/libtorrent/aux_/win_util.hpp
tigerw: Yeah, the pointer ends up in KernelBase.dll

include/libtorrent/aux_/win_util.hpp
tigerw: Can confirm that the loads were successful, Desktop 19041 and Mobile 15254

src/enum_net.cpp
tigerw: Surprisingly it does work, on Desktop and Mobile; the `ret` vector in `enum_net_interfaces` got filled with something reasonable-looking. (it even found cellular devices!)

src/path.cpp
tigerw: I've tried to implement symlinks with CreateFile2 and GetFileInformationByHandleEx. It seems to behave well with a symlinked file.

include/libtorrent/aux_/win_util.hpp
tigerw: Thanks! I've unchanged it.

tigerw: Aren't statics zero-initialised? I suppose you would prefer something more explicit?

arvidn: global variables are, and it's possible this one will be too, as it's probably allocated in the BSS segment. But it's technically initialized at runtime, the first time this function is called.

arvidn:
https://ci.appveyor.com/project/arvidn/libtorrent/builds/35594002/job/lttw7hv3cyy534y9#L106

arvidn:
the problem I was having building for windows store was that boost-build automatically sets the windows version to an older version. When I set it back to windows 10, it works. However, when building your branch I get these errors:

```
compile-c-c++ bin\msvc-14.2\debug\cxxstd-14-iso\threading-multi\windows-api-store\windows-version-win10\src\ip_notifier.obj
ip_notifier.cpp
C:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\shared\ntddndis.h(2306): error C2371: 'NDIS_MEDIUM': redefinition; different basic types
src\ip_notifier.cpp(52): note: see declaration of 'NDIS_MEDIUM'
C:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\shared\ntddndis.h(2336): error C2371: 'NDIS_PHYSICAL_MEDIUM': redefinition; different basic types
src\ip_notifier.cpp(53): note: see declaration of 'NDIS_PHYSICAL_MEDIUM'
```

```
compile-c-c++ bin\msvc-14.2\debug\cxxstd-14-iso\threading-multi\windows-api-store\windows-version-win10\src\enum_net.obj
enum_net.cpp
C:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\shared\ntddndis.h(2306): error C2371: 'NDIS_MEDIUM': redefinition; different basic types
src\enum_net.cpp(87): note: see declaration of 'NDIS_MEDIUM'
C:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\shared\ntddndis.h(2336): error C2371: 'NDIS_PHYSICAL_MEDIUM': redefinition; different basic types
src\enum_net.cpp(88): note: see declaration of 'NDIS_PHYSICAL_MEDIUM'
```

```
src\mmap_disk_io.cpp(1385): warning C4459: declaration of 'MEMORY_PRIORITY_INFORMATION' hides global declaration
C:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\um\processthreadsapi.h(996): note: see declaration of 'MEMORY_PRIORITY_INFORMATION'
```

Does this indicate that I end up using the wrong SDK? If so, it would be nice if that could be detected and explicitly fail the build with a `#error`

arvidn:
building the current `RC_2_0` for windows store on appveyor gives [these](https://ci.appveyor.com/project/arvidn/libtorrent/builds/35614685/job/6uvvxtqdp8x00w7u) errors. I think your patch addresses them all, except the one in assert.h. Presumably stack traces should just be disabled on WinRT.

src/enum_net.cpp
arvidn: I get the impression that the existence of these types is related to the version of the SDK, perhaps in addition to whether it's a WinRT build.

src/path.cpp
```diff
@@ -165,6 +165,42 @@ namespace {
...
+
+		s.mode = (file_attributes & FILE_ATTRIBUTE_DIRECTORY)
+			? file_status::directory
+			: (file_attributes & FILE_ATTRIBUTE_DEVICE)
+			? file_status::character_special : file_status::regular_file;
+	}
+
+	void fill_file_status(file_status & s, LARGE_INTEGER file_size, DWORD file_attributes, LARGE_INTEGER creation_time, LARGE_INTEGER last_access, LARGE_INTEGER last_write)
+	{
+		fill_file_status(s, file_size, file_attributes, large_integer_to_file_time(creation_time), large_integer_to_file_time(last_access), large_integer_to_file_time(last_write));
```
arvidn, line 193: I think this line is a bit too long

src/path.cpp
```diff
@@ -165,6 +165,42 @@ namespace {
 		// convert to seconds
 		return time_t(ft / 10000000 - posix_time_offset);
 	}
+
+	FILETIME large_integer_to_file_time(LARGE_INTEGER i)
```
arvidn, line 169: since `LARGE_INTEGER` is already part of the signature (you can't pass anything else to it), I think it would be fine for this function to just be called `to_file_time()`.

src/mmap.cpp
```diff
@@ -192,17 +224,23 @@ namespace {
...
+#endif
+
 } // anonymous
 
 file_handle::file_handle(string_view name, std::int64_t const size
 	, open_mode_t const mode)
+#ifdef TORRENT_WINRT
+	: m_fd(create_file(name, mode))
+#else
 	: m_fd(CreateFileW(convert_to_native_path_string(name.to_string()).c_str()
```
arvidn, line 236: I wonder if it would  be an improvement to move this call to `CreateFileW()` into a `create_file()` function as well, with the same signature as the one for WinRT. That way you could remove the `#ifdef` here, and instead just rely on the `create_file()` having the right definition based on build mode

src/enum_net.cpp
tigerw: Yep, might actually be a bugfix in a newer version, I'm using 17763.

arvidn:
would you mind rebasing your branch on top of `windows-phone`? That way, your fixes would get tested on appveyor as well, and we can land it once everything is green.

arvidn:
seems to work! would you mind squashing into reasonable commits?

tigerw:
yay! Does this repo have "squash and merge" when merging?

arvidn:
well, I'd prefer my commits to stay as they are

src/mmap.cpp
```diff
@@ -204,7 +204,7 @@ namespace {
 			file_flags(mode)
 		};
 
-		return CreateFile2(convert_to_native_path_string(name.to_string()).c_str()
+		return CreateFile2(name.data()
```
arvidn, line 207: I just noticed this. This is problematic. a `string_view` (or `wstring_view` for that matter) is not guaranteed to be 0-terminated. Just passing in the pointer to the string here is not safe, since `CreateFile2()` expects a 0-terminated string

src/mmap.cpp
```diff
@@ -224,24 +224,25 @@ namespace {
 			;
 	}
 
+	auto create_file(const wstring_view name, open_mode_t const mode)
+	{
+		return CreateFileW(name.data()
```
arvidn, line 229: same thing applies here. I think it would be better to take the `std::wstring` here instead, to have it moved in. Then you can call `.c_str()` on that

arvidn (approved):

arvidn:
thanks!

tigerw:
Thank you for the reviews!

PR #5200, opened 2020-10-06
WebTorrent: Prevent protocol encryption over WebRTC

paullouisageneau:
This PR fixes the assert `m_recv_buffer.packet_size() == dh_key_len` from `aux::bt_peer_connection::on_receive_impl`  in https://github.com/arvidn/libtorrent/issues/5100

WebTorrent connections are always made over DTLS and therefore should never use protocol encryption.

### DIFF
--- a/src/bt_peer_connection.cpp
+++ b/src/bt_peer_connection.cpp
@@ -200,6 +200,12 @@ namespace {
 		if (is_ssl(get_socket()))
 			out_policy = settings_pack::pe_disabled;
 #endif
+#if TORRENT_USE_RTC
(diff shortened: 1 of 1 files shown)


arvidn (approved):

arvidn:
do you think the issue happened when libtorrent attempted to connect to Webtorrent peers that don't support protocol encryption?
ideally this assert wouldn't fire even in that case though, but if that's the problem, it's probably unrelated to webtorrent support anyway.

paullouisageneau:
Yes, this is the situation that seem to trigger the issue. It's not perfectly clear to me what happens but PE should definitely not be used for WebTorrent, and the Javascript client don't support it.

PR #5206, opened 2020-10-11
Use boost::system in header-only form when compiling with boost >= 1.69

Chocobo1:
https://www.boost.org/doc/libs/1_72_0/libs/system/doc/html/system.html#changes_in_boost_1_69

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -719,9 +719,12 @@ else()
 endif()
 
 # Boost
-find_public_dependency(Boost REQUIRED COMPONENTS system)
+find_public_dependency(Boost REQUIRED)
 target_include_directories(torrent-rasterbar PUBLIC ${Boost_INCLUDE_DIRS})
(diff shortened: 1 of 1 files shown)


arvidn:
I would have expected to also have to define a macro to configure boost.system to be in header-only mode. Is it the case that boost.system is header only unconditionally after 1.69? and the `.so`/`.dll` is just a stub for backwards compatibility?

Chocobo1:
From the link in the opening post:
>Boost.System is now header-only. A stub library is still built for compatibility, but linking to it is no longer necessary.

>Is it the case that boost.system is header only unconditionally after 1.69? and the .so/.dll is just a stub for backwards compatibility?

So I suppose the answer is yes and yes.

>I would have expected to also have to define a macro to configure boost.system to be in header-only mode. 

I've built this way (no additional macro definition) for a long time and it worked fine.

PR #5209, opened 2020-10-11
Make boost version detection robust against various cmake versions

Chocobo1:
`CMP0093` old behavior seems to be not upheld in recent cmake versions, so I had to resort to more primitive variables to handle the boost version detection.
Another way is to raise the minimum cmake version but I'm inclined not touching this.
Fix up 012cf7d9fc7cdb4976d68d59fe0cde757672d9df.

Related downstream issue: https://github.com/qbittorrent/qBittorrent/issues/13493


zywo:
The issue is fixed.
Thank you.

examples/cmake/FindLibtorrentRasterbar.cmake
```diff
@@ -118,12 +118,14 @@ function(_try_generic_mode)
...
 	endif()
 
-	if(NOT Boost_SYSTEM_FOUND)
-		find_package(Boost QUIET REQUIRED COMPONENTS system)
+	find_package(Boost QUIET REQUIRED)
+	if (Boost_MAJOR_VERSION LESS_EQUAL 1 AND Boost_MINOR_VERSION LESS 69)
+		if (NOT Boost_SYSTEM_FOUND)
+			find_package(Boost QUIET REQUIRED COMPONENTS system)
+		endif()
+		set(LibtorrentRasterbar_LIBRARIES Boost::system ${CMAKE_THREAD_LIBS_INIT})
```
arvidn, line 126: it looks suspicious to move this whole `set()` call to within the if-block. with boost >= 1.69, do we not need to include `${CMAKE_THREAD_LIBS_INIT}`?

Chocobo1: you're right, now I've separated `${CMAKE_THREAD_LIBS_INIT}` out.

arvidn (approved):

PR #5215, opened 2020-10-14
fix issue with pausing a session

arvidn:
which would cause tracker_manager to enter shutdown-mode and never come back

### DIFF
--- a/include/libtorrent/tracker_manager.hpp
+++ b/include/libtorrent/tracker_manager.hpp
@@ -371,6 +371,7 @@ namespace libtorrent {
 			, std::weak_ptr<request_callback> c
 				= std::weak_ptr<request_callback>()) = delete;
 		void abort_all_requests(bool all = false);
+		void stop();
 
 		void remove_request(http_tracker_connection const* c);
 		void remove_request(udp_tracker_connection const* c);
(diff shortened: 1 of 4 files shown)


src/tracker_manager.cpp
```diff
@@ -425,14 +426,19 @@ namespace libtorrent {
...
+	}
+
 	void tracker_manager::abort_all_requests(bool all)
 	{
 		// this is called from the destructor too, which is not subject to the
 		// single-thread requirement.
 		TORRENT_ASSERT(all || is_single_thread());
 		// removes all connections except 'event=stopped'-requests
 
-		m_abort = true;
```
arvidn, line 435: this is the fix

include/libtorrent/tracker_manager.hpp
```diff
@@ -371,6 +371,7 @@ namespace libtorrent {
 			, std::weak_ptr<request_callback> c
 				= std::weak_ptr<request_callback>()) = delete;
 		void abort_all_requests(bool all = false);
+		void cancel();
```
glassez, line 374: I would rename them:
1. abort_all_requests => cancel_requests
2. cancel => abort

PR #5216, opened 2020-10-14
torrent_info::hash2_for_piece

joriscarrier:
untested request for advice

### DIFF
--- a/include/libtorrent/torrent_info.hpp
+++ b/include/libtorrent/torrent_info.hpp
@@ -493,6 +493,23 @@ TORRENT_VERSION_NAMESPACE_3
 			return &m_info_section[std::ptrdiff_t(m_piece_hashes) + idx * 20];
 		}
 
+		// ``hash2_for_piece()`` takes a piece-index and returns the 32-bytes
+		// sha256-hash for that piece.
+		// ``hash2_for_piece_ptr()`` returns a pointer to the 32 byte sha256 digest
+		// for the piece. Note that the string is not 0-terminated.
+		sha256_hash hash2_for_piece(piece_index_t index) const;
+		char const* hash2_for_piece_ptr(piece_index_t const index) const
+		{
+			TORRENT_ASSERT_PRECOND(index >= piece_index_t(0));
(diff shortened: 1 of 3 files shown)


include/libtorrent/torrent_info.hpp
```diff
@@ -493,6 +493,23 @@ TORRENT_VERSION_NAMESPACE_3
...
+		char const* hash2_for_piece_ptr(piece_index_t const index) const
+		{
+			TORRENT_ASSERT_PRECOND(index >= piece_index_t(0));
+			TORRENT_ASSERT_PRECOND(index < m_files.end_piece());
+			TORRENT_ASSERT(is_loaded());
+			int const idx = static_cast<int>(index);
+			TORRENT_ASSERT(m_piece_hashes > 0);
+			TORRENT_ASSERT(m_piece_hashes < m_info_section_size);
+			TORRENT_ASSERT(idx < int((m_info_section_size - m_piece_hashes) / 32));
+			return &m_info_section[std::ptrdiff_t(m_piece_hashes) + idx * 32];
```
arvidn, line 510: v2 hashes are not located at this location. in fact, for a v2-only torrent, there may not be any `m_piece_hashes`. In v2-torrents each file has a merkle hash tree. The root hash for a file can be retrieved from the `file_storage`.

```
ti->files().root(file_index);
ti->files().root_ptr(file_index);
```

Any other hashes *may* not be available. However, if the `torrent_info` object was loaded from a `.torrent` file, there should be *piece layers*. i.e. the merkle node hashes at the layer where each hash corresponds to one piece size in the file.

It's a bit complicated to expose these as they are owned and modified by the internal bittorrent engine once the torrent is added, so the client can't easily access them without introducing a mutex.

One idea I am planning to implement is to expose the piece layers in the `torrent_info` object and as soon as the torrent is added to as session, those piece layers disappear and are moved into the internal torrent object. i.e. no data sharing issues.

Would that work for your use case?

joriscarrier: Yes, Thank you for the answer and your time

PR #5220, opened 2020-10-15
support constructing sessions in paused mode via new session_flag

arvidn:


include/libtorrent/session.hpp
```diff
@@ -184,14 +184,22 @@ namespace aux {
...
+		// paused mode (by passing in ``session::paused``). Note that
+		// ``add_default_plugins`` do not have an affect on constructors that
+		// take a session_params object. It already contains the plugins to use.
 		explicit session(session_params const& params)
-		{ start(session_params(params), nullptr); }
+		{ start({}, session_params(params), nullptr); }
 		explicit session(session_params&& params)
-		{ start(std::move(params), nullptr); }
+		{ start({}, std::move(params), nullptr); }
+		explicit session(session_params const& params, session_flags_t const flags)
```
glassez, line 195: Just to be sure... Is it impossible to implement it via session_params?

arvidn: no, and I could do that in `master`. But the layout of `session_params` is locked in the ABI for 1.2 and 2.0

glassez (approved):

PR #5221, opened 2020-10-15
fix CMake config file version compatibility

FranciscoPombal:
Make it the application's responsibility to manage their supported
versions.

This can be done with CMake find_package() version ranges, for instance.

---

Context: https://github.com/qbittorrent/qBittorrent/issues/13510#issuecomment-709279090 and the following two comments below it. An application like qBittorrent might want to support two or more versions at the same time that differ in major and minor version. It is difficult to accomplish this with the `SameMinorVersion` restriction in the generated CMake config files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -818,7 +818,7 @@ string(REPLACE ";" "\n" _find_dependency_calls "${_find_dependency_calls}")
 write_basic_package_version_file(
 	"${CMAKE_CURRENT_BINARY_DIR}/LibtorrentRasterbar/LibtorrentRasterbarConfigVersion.cmake"
 	VERSION ${libtorrent_VERSION}
-	COMPATIBILITY SameMinorVersion
+	COMPATIBILITY AnyNewerVersion
 )
 
 export(EXPORT LibtorrentRasterbarTargets


FranciscoPombal:
This is meant to be merged into `RC_2_0` and `master` as well, of course.

zywo:
@FranciscoPombal 
Just tested and works as expected with this command:
```cmake -DLibtorrentRasterbar_DIR=~/build/rc_2_0/LibtorrentRasterbar/ ~/qbittorrent/```

**Comment by zywo** [comment]
Yes, but I avoid doing ```sudo make install``` every time I compile libtorrent.

File: CMakeLists.txt [hunk 1]
```diff
@@ -818,7 +818,7 @@ string(REPLACE ";" "\n" _find_dependency_calls "${_find_dependency_calls}")
 write_basic_package_version_file(
 	"${CMAKE_CURRENT_BINARY_DIR}/LibtorrentRasterbar/LibtorrentRasterbarConfigVersion.cmake"
 	VERSION ${libtorrent_VERSION}
-	COMPATIBILITY SameMinorVersion
```
glassez, line 821: It's confusing...
Why does CMake provide this feature at all, so that the package can somehow affect its supporting? Shouldn't it be entirely the user's responsibility to determine which versions it supports?

FranciscoPombal: ¯\\\_(ツ)_/¯, I'm sure some projects make good use of it.

FranciscoPombal:
Big oof, I just realized version ranges are only available in CMake >= 3.19... they already updated the docs pages with the 3.19 RC, but I thought I was looking at some earlier version.

So I guess the ranges will have to wait, but that's not too bad. The main purpose of this PR is simply about raising minimum versions anyway.

arvidn:
looks reasonable. @zeule you agree?

CMakeLists.txt
arvidn: does `COMPATIBILITY` mean ABI or API compatibility?
without knowing really what it means, it seems reasonable to document it in some way, like this. Is the problem that a project that supports multiple versions have no way of expressing all the versions it supports when looking for a dependency?

zywo: @arvidn 
https://github.com/qbittorrent/qBittorrent/issues/13510#issuecomment-709322633

FranciscoPombal: @arvidn It's purely a "dumb" version number comparison policy. It's too inflexible when set to anything other than `AnyNewerVersion`, downstream projects should be the ones that what versions they support above a certain minimum (and they have the means to do this).

arvidn (approved):

PR #5225, opened 2020-10-18
fix incorrectly inlined move-assignment of file_storage

arvidn:


include/libtorrent/file_storage.hpp
```diff
@@ -220,7 +220,7 @@ namespace libtorrent {
 		file_storage(file_storage const&);
 		file_storage& operator=(file_storage const&);
 		file_storage(file_storage&&) noexcept;
-		file_storage& operator=(file_storage&&) = default;
+		file_storage& operator=(file_storage&&);
```
arvidn, line 223: the definition in `file_storage.cpp` as already there

PR #5227, opened 2020-10-18
copy torrent_info when adding to session

arvidn:


arvidn:
the fundamental problem is that the internal `torrent` objects keeps (and mutates) state inside the `torrent_info` object. If the client (thread) also has a shared_ptr to the *same* `torrent_info` objects, and looks at this state, there's a race condition. This is the main reason to make a copy. It's still possible to get the internal torrent_info object, but by then most mutations have happened.

As a separate fix, the amount of mutable state kept in the torrent_info object should be minimized.

PR #5228, opened 2020-10-19
if pre-allocated handler buffer is too small, fall back to malloc/free

arvidn:


sledgehammer999:
FYI, tt seems that these handlers were deprecated in boost 1.74: https://www.boost.org/doc/libs/1_74_0/doc/html/boost_asio/reference/asio_handler_allocate.html

arvidn:
yeah, the new ones are used in libtorrent 2.0

PR #5259, opened 2020-11-03
add install rule for python module to Jamfile

arvidn:


bindings/python/Jamfile
```diff
@@ -232,6 +232,18 @@ my-python-extension libtorrent
 	<suppress-import-lib>false
 	;
 
+rule python-install-dir ( properties * )
+{
+	local python_path = [ SHELL "python -c \"import sys; print(sys.path[4])\"" ] ;
+	return <location>$(python_path) ;
+}
+
+install install_module
```
arvidn, line 241: this target must be made explicit

userdocs:
I am hitting an error.

```
building boost from source directory:  /root/libtorrent-python/boost
Jamfile:265: Unescaped special character in argument $([ feature.get-values python : $(properties) ] )
Jamfile:273: Unescaped special character in argument $([ feature.get-values python : $(properties) ] )
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: malformed subscript
Jamfile:265: unexpected text following []
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: malformed subscript
Jamfile:273: unexpected text following []
Jamfile:280: in Jamfile</root/libtorrent-python/libtorrent/bindings/python>.python-install-dir
*** argument error
* rule SHELL ( command : * )
* called with: (  )
* missing argument command
(builtin):see definition of rule 'SHELL' being called
/root/libtorrent-python/boost/tools/build/src/kernel/modules.jam:107: in modules.call-in
/root/libtorrent-python/boost/tools/build/src/util/indirect.jam:105: in indirect.call
/root/libtorrent-python/boost/tools/build/src/build/property.jam:132: in property.evaluate-conditionals-in-context
/root/libtorrent-python/boost/tools/build/src/build/targets.jam:1088: in evaluate-requirements
/root/libtorrent-python/boost/tools/build/src/build/targets.jam:1122: in common-properties2
/root/libtorrent-python/boost/tools/build/src/build/targets.jam:1018: in targets.common-properties
/root/libtorrent-python/boost/tools/build/src/build/targets.jam:1314: in class@basic-target.generate
/root/libtorrent-python/boost/tools/build/src/build/targets.jam:812: in generate-really
/root/libtorrent-python/boost/tools/build/src/build/targets.jam:784: in class@main-target.generate
/root/libtorrent-python/boost/tools/build/src/build-system.jam:797: in load
/root/libtorrent-python/boost/tools/build/src/kernel/modules.jam:295: in import
/root/libtorrent-python/boost/tools/build/src/kernel/bootstrap.jam:139: in boost-build
/root/libtorrent-python/boost/boost-build.jam:17: in module scope
```

I have tried 

```
b2 install_module
b2 install_module python-install-scope=system
```

userdocs:
## error info

The problem for me appears to be around the this coming back null but nothing happens.

Specifically this `python.interpreter` whereas just `python` give a result.

```
local python-interpreter = [ feature.get-values python.interpreter : $(properties) ] ;
if ( $(python-interpreter) = "" )
{
	python-interpreter = "python" ;
}

```

If remove that part and specify `[ SHELL "python3 -c ...` it worked.

```
rule python-install-dir ( properties * )
{
	local site-selector = 4 ;
	if <python-install-scope>system in $(properties)
	{
		switch $([ feature.get-values python : $(properties) ] )
		{
			case 2.* : site-selector = 7 ;
			case 3.* : site-selector = 5 ;
		}
	}
	else
	{
		switch $([ feature.get-values python : $(properties) ] )
		{
			case 2.* : site-selector = 6 ;
			case 3.* : site-selector = 4 ;
		}
	}

	local python-path = [ SHELL "python3 -c \"import sys; sys.stdout.write(sys.path[$(site-selector)])\"" ] ;

	ECHO "python install directory:" $(python-path) ;
	return <location>$(python-path) ;
}
```

## python path detection 

I also think i have found a very consistent way to get the path from here reading here

https://github.com/boostorg/build/blob/develop/Jamroot.jam#L118#L143 and going here https://docs.python.org/3/library/sysconfig.html

So how does it work? this command `.get_config_vars` gives out a serious list of variables.

```bash
python2 -c "import sysconfig; print(sysconfig.get_config_vars())"
python3 -c "import sysconfig; print(sysconfig.get_config_vars())"
```

This slightly different command `.get_config_var` reads a specific variable and give the exact same result across Debian/Ubuntu/Alpine of the `.../lib-dynload` directory.

```bash
python2 -c "import sysconfig; print(sysconfig.get_config_var('DESTSHARED'))"
python3 -c "import sysconfig; print(sysconfig.get_config_var('DESTSHARED'))"
```

Sadly it does not help with the local path but the other command is consistent anyway.

userdocs:
I got his working but i had to make some changes. Perhaps understand why will help see if there are useful tweaks needed to work process platforms and environments.

```
rule python-install-dir ( properties * )
{
	local install-dir = [ feature.get-values python-install-path : $(properties) ] ;
	if ( $(install-dir)  != "" )
	{
		# if the user has provided an install location, use that one
		return <location>$(install-dir) ;
	}

	local python-version = [ feature.get-values python : $(properties) ] ;
	if ( $(python-version) = "" )
	{
		python-interpreter = "python" ;
	}
    else
    {
        python-interpreter = "python$(python-version)" ;
    }

	# sys.path are defined differently between python2 and python3

	local site-selector = 0 ;
    
	if <python-install-scope>system in $(properties)
	{
		switch $(python-version)
		{
			case 2.* : site-selector = 5 ;
			case 3.* : site-selector = 3 ;
		}
	}
	else
	{
		switch $(python-version)
		{
			case 2.* : site-selector = 6 ;
			case 3.* : site-selector = 4 ;
		}
	}

	local python-path = [ SHELL "$(python-interpreter) -c \"import sys; sys.stdout.write(sys.path[$(site-selector)])\"" ] ;

	ECHO "python install directory:" $(python-path) ;
	return <location>$(python-path) ;
}
```

One main change i don't understand was this,

```
local python-version = [ feature.get-values python.interpreter : $(properties) ] ;
```

to 

```
local python-version = [ feature.get-values python : $(properties) ] ;
```

Another was this `switch $(python-version)`. 

I sort of understand and follow what you are doing but i think this is outdated with thse changes i made to clone your method but using newer commands.

https://github.com/arvidn/libtorrent/pull/5262/files#diff-6abbccdb045e12134229c5dba31ef1783bb3b7301ccaa42c11f8855b4572a278

arvidn:
```
local python-version = [ feature.get-values python.interpreter : $(properties) ] ;
```

This is looking for the python interpreter, as configured by the `using python ;` statement in `user-config.jam`. We want the python interpreter to know which binary to run the little python script in, that returns the path.

```
local python-version = [ feature.get-values python : $(properties) ] ;
```

This is looking for the python *version*. It's not trivial to go from the *version* to the *binary*. You assume there's a binary in the current `PATH` called `python2` or `python3`. I don't think that's a safe assumption. On windows, `python.exe` may not even be available in `PATH`.

arvidn:
> Specifically this python.interpreter whereas just python give a result.

Can you elaborate on that? The logic is supposed to set the variable to `python` in case `python.interpreter` isn't set, you'd get `python`. Is the issue that `python` isn't available on your system, causing this to fail? The error makes it sound like the `python-interpreter` variable is empty though.

userdocs:
This is my `user-config.jam` 

```
using python : 3.7 : /usr/bin/python3.7 : /usr/include/python3.7 : /usr/lib/python3.7 ;
using gcc : : : <cxxflags>-std=c++14 ;
```

Why did it fail to work for me? 

```
local python-version = [ feature.get-values python.interpreter : $(properties) ] ;
```

Maybe because i used this `/usr/bin/python3.7` ?

I'm not saying it's wrong. I am just not sure why it returns a null value, which in turn broke the script.

userdocs:
Sorry, i just saw your second comment.

On Ubuntu 20.04 and Alpine Linux dockers, installing `python2` or `python3` will lead to a error when using `python`. I need to use `python2` and `python3`

```
root@eaffbe6de78b:~# python
bash: python: command not found
root@eaffbe6de78b:~# python3
Python 3.7.3 (default, Jul 25 2020, 13:03:44)
```

Currently you have this 

```
	local python-interpreter = [ feature.get-values python.interpreter : $(properties) ] ;
	if ( $(python-interpreter) = "" )
	{
		python-interpreter = "python" ;
	}

	local python-interpreter = [ feature.get-values python.interpreter : $(properties) ] ;
```

I think the second instance of this line is overriding the first check breaking following part? https://github.com/arvidn/libtorrent/pull/5259/files#diff-6abbccdb045e12134229c5dba31ef1783bb3b7301ccaa42c11f8855b4572a278R252-R258

```
	local python-interpreter = [ feature.get-values python.interpreter : $(properties) ] ;
```

If i test this in the `jamfile`

```
rule python-install-dir ( properties * )
{
	local python-interpreter = [ feature.get-values python.interpreter : $(properties) ] ;
	if ( $(python-interpreter) = "" )
	{
		python-interpreter = "python" ;
	}

	ECHO "python-interpreter:" $(python-interpreter) ;
    exit ;
}
```

I get this (which is the else condition):

```
OS = LINUX
building boost from source directory:  /root/libtorrent-python/boost
python-interpreter: python
```

So either way. this check never works for me.

```
local python-path = [ SHELL "$(python-interpreter) -c \"import sys; sys.stdout.write(sys.path[$(site-selector)])\"" ] ;
```

arvidn:
I think the issue you experienced is because of a bug I had. I don't really understand why it didn't fail for me. It failed for me right now though, so I must have made some mistake.

It should work now though.

userdocs:
I have tested it and the main errors are fixes but there is still 2 problems.

**1:** the result is just `python` and i need to also include `python2` or `python3` othwise it is always checking for `python` and version specific checks are not being made, specifically an issue here https://github.com/arvidn/libtorrent/pull/5259/files#diff-6abbccdb045e12134229c5dba31ef1783bb3b7301ccaa42c11f8855b4572a278R260-R276

I test with a jamfile to do `using python3 : 3.7 :` to see if this was configuration thing i could solve but i just get this when using `b2`

```
python3.jam: No such file or directory
```

In my approach i believed i was not assuming the version, but taking the version set in the jamfile and using `python2` or `python3` which should always work, whereas i thought `python` was a symlink created by the deb package rather than the actual python binary for the relevant version. You can see that here https://github.com/arvidn/libtorrent/pull/5262/files#diff-6abbccdb045e12134229c5dba31ef1783bb3b7301ccaa42c11f8855b4572a278R254-R256

Which in turn generated this variable `python$(python_v)`. I had believed the safest method way to call the relevant binary directly based on the check from the `user-config.jam`

**2:** if we use this command https://github.com/arvidn/libtorrent/pull/5262/files#diff-6abbccdb045e12134229c5dba31ef1783bb3b7301ccaa42c11f8855b4572a278R266

```
python_system_path = [ SHELL "python$(python_v) -c \"import sys; import sysconfig; sys.stdout.write(sysconfig.get_config_var('DESTSHARED'))\"" ] ;
```

it should remove the need for this entire section https://github.com/arvidn/libtorrent/pull/5259/files#diff-6abbccdb045e12134229c5dba31ef1783bb3b7301ccaa42c11f8855b4572a278R260-R276

Can you have another look over what i changed  in my pull to see what i mean?

https://github.com/arvidn/libtorrent/pull/5262/files#diff-6abbccdb045e12134229c5dba31ef1783bb3b7301ccaa42c11f8855b4572a278

The only thing i have in addition to your version is the `PYTHONPATH`check. if i removed that they are functional identically, i thought?

arvidn:
`python` does not work on windows. my recollection is that the default install path for python is `C:\Python3.8\bin` or something along those lines. The `PATH` environment variable.

My point wasn't that you assumed the version of the python interpreter, but the *name* and *path* of the python interpreter. namely, that `python` + major version becomes a valid and reachable executable for the python interpreter. `python2` isn't necessarily valid even on linux systems.

arvidn:
```
sys.stdout.write(sysconfig.get_config_var('DESTSHARED'))
```

Is `DESTSHARED` documented anywhere? And is this the system python-site (it sounds like it), if so, what's the name of the variable for user python-site?

userdocs:
Regarding python. I don't know about windows but specific to linux and Debain based these tests might be helpful.

Using these commands on a docker.

```
docker run -it debian:jessie /bin/bash
docker run -it debian:buster /bin/bash
docker run -it debian:sid /bin/bash
```

Using this installation setup

```
apt-get update
apt-get -y install python python3
```

Followed by these test commands

```
cd /usr/bin/
ls -la | grep " python"
```

I get theses results

Debian Buster

```
lrwxrwxrwx  1 root root         9 Mar 16  2015 python -> python2.7
lrwxrwxrwx  1 root root         9 Mar 16  2015 python2 -> python2.7
-rwxr-xr-x  1 root root   3785992 Sep 14  2019 python2.7
lrwxrwxrwx  1 root root         9 Nov 30  2014 python3 -> python3.4
-rwxr-xr-x  2 root root   4480600 Sep 14  2019 python3.4
-rwxr-xr-x  2 root root   4480600 Sep 14  2019 python3.4m
lrwxrwxrwx  1 root root        10 Nov 30  2014 python3m -> python3.4m
```

Debian Stable

```
lrwxrwxrwx  1 root root         7 Mar  4  2019 python -> python2
lrwxrwxrwx  1 root root         9 Mar  4  2019 python2 -> python2.7
-rwxr-xr-x  1 root root   3689352 Oct 10  2019 python2.7
lrwxrwxrwx  1 root root         9 Mar 26  2019 python3 -> python3.7
-rwxr-xr-x  2 root root   4861504 Jul 25 13:03 python3.7
-rwxr-xr-x  2 root root   4861504 Jul 25 13:03 python3.7m
lrwxrwxrwx  1 root root        10 Mar 26  2019 python3m -> python3.7m
```

Debian Sid

```
lrwxrwxrwx  1 root root         7 Nov  3 08:20 python -> python2
lrwxrwxrwx  1 root root         9 Aug  4 08:22 python2 -> python2.7
-rwxr-xr-x  1 root root   3672400 Apr 20  2020 python2.7
lrwxrwxrwx  1 root root         9 Oct 13 19:25 python3 -> python3.8
-rwxr-xr-x  1 root root   5233216 Sep 25 09:36 python3.8
```

This shows to me that `python` is a symlink and the real binary is always in the format `pythonMAJOR.MINOR` so testing for this should always work.

Currently i was just looking at `pythonMAJOR`. So can i conclude this?

-  you suggest we look for the `pythonMAJOR.MINOR` instead of `pythonMAJOR`.

arvidn:
> Currently i was just looking at pythonMAJOR. So can i conclude this?
> 
> you suggest we look for the pythonMAJOR.MINOR instead of pythonMAJOR.

No, I'm suggesting using the python interpreter that's configured in `user-config.jam`. i.e. the second parameter passed to `using python : <version> : <interpreter> : <include-path> : <library-path> ;`.

Relying on that configuration is more likely (I claim) to use the correct interpreter, that's compatible with the include path and libraries found in those paths that are also configured in `user-config.jam`.

userdocs:
Regarding, DESTSHARED

I cannot find it specifically documented. The main commands to check are documented though -https://docs.python.org/3/library/sysconfig.html#configuration-variables

But it is in the python source code from 2.7 to 3.9.

```
DESTSHARED=	$(BINLIBDEST)/lib-dynload
```

https://github.com/python/cpython/blob/2.7/Makefile.pre.in#L127

https://github.com/python/cpython/blob/3.9/Makefile.pre.in#L162

userdocs:
Ok, so i think i see where i got confused but not why.

With this `$HOME/user-config.jam`. 

```
using python : 3.7 : /usr/bin/python3.7 : /usr/include/python3.7 : /usr/lib/python3.7 ;
using gcc : : : <cxxflags>-std=c++14 ;
```

Docs example for reference.

```
using python : 3.5 : C:/Users/<UserName>/AppData/Local/Programs/Python/Python35 : C:/Users/<UserName>/AppData/Local/Programs/Python/Python35/include : C:/Users/<UserName>/AppData/Local/Programs/Python/Python35/libs ;
```

And using this test code.

```
rule python-install-dir ( properties * )
{
	local python-interpreter = [ feature.get-values python.interpreter : $(properties) ] ;
	if ( $(python-interpreter) = "" )
	{
		python-interpreter = "python" ;
	}

	ECHO $(python-interpreter) ;
    exit ;
}
```

I am getting this result.

building boost from source directory:  /root/libtorrent-python/boost
python

If i don't generate the `$HOME/user-config.jam` the build errors out.

-  I get a null value from this `python.interpreter` when it is set in my `$HOME/user-config.jam`. How can i test this to understand why, i should be getting `/usr/bin/python3.7`?

userdocs:
There problems i experienced (especially with testing) seem so be related this. It appears to loop 3 times and does not load the data from the `user-config.jam` the first time. I don't know why it does this.

```
OS = LINUX
building boost from source directory:  /root/libtorrent-python/boost
python
sh: 1: python: not found
python install directory: .
/usr/bin/python3.7
python install directory: /root/.local/lib/python3.7/site-packages
/usr/bin/python3.7
python install directory: /root/.local/lib/python3.7/site-packages
Performing configuration checks
```

userdocs:
Here is a debugged configuration output.

```
notice: Loading user-config configuration file 'user-config.jam' from '/root'.
notice: will use 'g++' for gcc, condition <toolset>gcc-8
notice: using gcc libraries :: <toolset>gcc-8 :: /usr/bin /usr/lib /usr/lib32 /usr/lib64
notice: using gcc archiver :: <toolset>gcc-8 :: /usr/bin/ar
notice: using gcc ranlib :: <toolset>gcc-8 :: /usr/bin/ranlib
warning: toolset gcc initialization: can not find tool windres
warning: initialized from /root/user-config.jam:1
notice: using rc compiler :: <toolset>gcc-8 :: /usr/bin/as
notice: [python-cfg] Configuring python...
notice: [python-cfg]   user-specified version: "3.7"
notice: [python-cfg]   user-specified cmd-or-prefix: "/usr/bin/python3.7"
notice: [python-cfg]   user-specified includes: "/usr/include/python3.7"
notice: [python-cfg]   user-specified libraries: "/usr/lib/python3.7"
notice: [python-cfg] Details of this Python configuration:
notice: [python-cfg]   interpreter command: "/usr/bin/python3.7"
notice: [python-cfg]   include path: "/usr/include/python3.7"
notice: [python-cfg]   library path: "/usr/lib/python3.7"
notice: [python-cfg] Checking for NumPy...
notice: [python-cfg] running command '/usr/bin/python3.7 -c "import sys; sys.stderr = sys.stdout; import numpy; print(numpy.get_include())"'
notice: [python-cfg] NumPy enabled
CXXFLAGS = -std=c++14 -fPIC
LDFLAGS = -L/root/libtorrent-python/lib
OS = LINUX
building boost from source directory:  /root/libtorrent-python/boost
notice: [zlib] Using pre-installed library
notice: [zlib] Condition
notice: [bzip2] Using pre-installed library
notice: [bzip2] Condition
notice: [lzma] Using pre-installed library
notice: [lzma] Condition
notice: [zstd] Using pre-installed library
notice: [zstd] Condition
python
sh: 1: python: not found
python install directory: .
/usr/bin/python3.7
python install directory: /usr/local/lib/python3.7/dist-packages
/usr/bin/python3.7
python install directory: /usr/local/lib/python3.7/dist-packages
Performing configuration checks
```

The warning `sh: 1: python: not found` is not something i think should be there.

What i don't really get it why i can use something like this that works on first load with no loop.

```
local python-interpreter = [ feature.values python ] ;
```

But i can't use something like this, which would essentially be perfect.

```
local python-interpreter = [ feature.values python.interpreter ] ;
```

userdocs:
If we could load this `cmd-or-prefix` it would be ideal?

https://github.com/boostorg/build/blob/5faca0dfca1de3d2297aed437a85754a1fbf5089/src/tools/python.jam#L104#L121

I have no idea how that could be done though. jamfiles are a recent mystery to me i'm trying to decipher, as you might have noticed.

arvidn:
I also don't understand why the rule invoked 3 times. But it seems like it settles on all the correct usage-requirements and that it's the last invocation that counts. It works for me on ubuntu and windows. I made a change to not try to execute "python", but just treat it as one of the initial invocation that can't succeed.

This current patch, does it work for you? If not, what's the error? Does it not discover the correct python interpreter, even after one or two invocations?

userdocs:
## Installation using the current jamfile

I tested it i now see this

```
OS = LINUX
building boost from source directory:  /root/libtorrent-python/boost
/usr/bin/python3.7
python install directory: /usr/local/lib/python3.7/dist-packages
/usr/bin/python3.7
python install directory: /usr/local/lib/python3.7/dist-packages
Performing configuration checks
```

### The good.

It loops x2 but both results are the same and no `sh` warning. i can live with that if it means a better cross platform solution as the real problem was the first loop, which seems to be resovled.

### the bad

That was using `install_module` with no scope specified, so it defaulted to `user`.

This is the path. This path is completely wrong for that scope.

```
python install directory: /usr/local/lib/python3.7/dist-packages
```

This is the final problem to address.

Any thoughts on the links i provided  here? https://github.com/arvidn/libtorrent/pull/5259#issuecomment-723360011

arvidn:
is `user` the correct default, or do you think it should default to `system`?

sys, is `sys.path[4]` not consistently returning the local user's python path? For me, in python3, it returns:

```
/home/arvid/.local/lib/python3.8/site-packages
```

Which looks like the per-user directory. Do you know how `sys.path` array is supposed to be interpreted?

userdocs:
I think user is the correct default. It's safe and built into python site so the result will always be the same.

The problems comes with the `sys.path` method. This is the correct way to view paths but as it is dynamic, hard coding the result to `4` or `6` is going present the chance the wrong dirt is selected by the installer. It is supposed to show you the paths but i am not aware of how to specific select the result i want. Unless u was using bash to grep/sed the result.

I'd say if you are unsure which system method to implement. remove it for now and stick with custom or user. But as it is the result is only certain on a new system install, like a docker.

I will look into `sys.path` to see there is a way to search/refine the result.

userdocs:
Oh, what about this?

Back to `sysconfig` but a different method

https://docs.python.org/3/library/sysconfig.html#sysconfig.get_paths

##System Python 2

```
python2 -c 'import sysconfig; print(sysconfig.get_paths())'

{'platinclude': '/usr/local/include/python2.7', 'platstdlib': '/usr/lib/python2.7', 'platlib': '/usr/local/lib/python2.7/dist-packages', 'purelib': '/usr/local/lib/python2.7/dist-packages', 'stdlib': '/usr/lib/python2.7', 'scripts': '/usr/local/bin', 'include': '/usr/local/include/python2.7', 'data': '/usr/local'}
```

Selective result

```
python2 -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])'
/usr/local/lib/python2.7/dist-packages
```

## System python3

```
python3 -c 'import sysconfig; print(sysconfig.get_paths())'
{'stdlib': '/usr/lib/python3.7', 'platstdlib': '/usr/lib/python3.7', 'purelib': '/usr/lib/python3.7/site-packages', 'platlib': '/usr/lib/python3.7/site-packages', 'include': '/usr/include/python3.7m', 'platinclude': '/usr/include/python3.7m', 'scripts': '/usr/bin', 'data': '/usr'}
```

Selective result

```
python3 -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])'
/usr/lib/python3.7/site-packages
```

## User

https://docs.python.org/3/library/site.html#site.USER_SITE

```
python2 -c "import site; print(site.USER_SITE)"
/root/.local/lib/python2.7/site-packages
```

```
python3 -c "import site; print(site.USER_SITE)"
/root/.local/lib/python3.7/site-packages

```

## Conclusion

Both methods are documented and provide a consistent result.

arvidn:
does this look right?

userdocs:
Some updates.

It works but i came across a problem + possible solutions.

So it turns out we need to make sure that the path we chose is also listed by this output as these are library paths, like `LD_LIBRARY_PATH` paths.

```
python3 -c "import sys; print (sys.path)"
['', '/usr/lib/python37.zip', '/usr/lib/python3.7', '/usr/lib/python3.7/lib-dynload', '/usr/local/lib/python3.7/dist-packages', '/usr/lib/python3/dist-packages']
```

Unfortunately it seems `purelib` give result not in this path.

```
python3 -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])'
/usr/lib/python3.7/site-packages
```

The only overlapping path on my docker tests was `stdlib`

So this works to install a library deluge can load..

```
python2 -c 'import sysconfig; print(sysconfig.get_paths()["stdlib"])'
python3 -c 'import sysconfig; print(sysconfig.get_paths()["stdlib"])'
```

Also we can just use this version.

```
python2 -c "import sysconfig; print(sysconfig.get_path('stdlib'))"
python3 -c "import sysconfig; print(sysconfig.get_path('stdlib'))"
```

`stdlib` seems like a higher level path `/usr/lib` but it works and solves the problem. Here are the definitions.

-  stdlib: directory containing the standard Python library files that are not platform-specific.
-  platstdlib: directory containing the standard Python library files that are platform-specific.
-  platlib: directory for site-specific, platform-specific files.
-  purelib: directory for site-specific, non-platform-specific files.
-  include: directory for non-platform-specific header files.
-  platinclude: directory for platform-specific header files.
-  scripts: directory for script files.
-  data: directory for data files.

https://docs.python.org/3/library/sysconfig.html#installation-paths

## Alternative solution revisit DESTSHARED

Why? The method to read the variables is document but the vast options of variables available isn't. But we can view them using this command.

```
python2 -m sysconfig
python3 -m sysconfig
```

There is a lot to chose from here and the method to get a particular variable is this.

```
python2 -c "import sysconfig; print(sysconfig.get_config_var('DESTSHARED'))"
python3 -c "import sysconfig; print(sysconfig.get_config_var('DESTSHARED'))"
```

I think it's the only option if we decided using a lower lever lib path is required.

arvidn:
> So it turns out we need to make sure that the path we chose is also listed by this output as these are library paths, like LD_LIBRARY_PATH paths.

Why is that?

> platlib: directory for site-specific, platform-specific files.

This seems like the best fit (of that list) to me. A compiled module surely is "platform-specific", and presumably the "system" install directory is "site-specific". But the question then is which path is the user-specific install directory.

arvidn:
surely python will look for python modules in specific directories, not what `LD_LIBRARY_PATH` points to

userdocs:
let me test with `platlib` and get back to you with examples.

arvidn:
I think the first question is the most important though. If you don't know *why* python modules only work if they're in an `LD_LIBRARY_PATH` directory, how can you know changing the install path is right?

userdocs:
Sorry, perhaps i misused that `LD_LIBRARY_PATH`. I was just trying to explain the problem with what i thought was a similar reference. I don't actually know if it is related, overlapping or totally irrelevant. Please look at the tests i did to see what i mean.

userdocs:
## The problem

Install path scope system (platlib)

```
common.copy /usr/lib/python3.7/site-packages/libtorrent.so
```

Test the `sys.path` to see where libraries are loaded from. `site-packages` is missing.

```
python3 -c "import sys; print(sys.path)"
['', '/usr/lib/python37.zip', '/usr/lib/python3.7', '/usr/lib/python3.7/lib-dynload', '/usr/local/lib/python3.7/dist-packages', '/usr/lib/python3/dist-packages']
```

Test deluge2

```
  File "/usr/local/lib/python3.7/dist-packages/deluge/_libtorrent.py", line 25, in <module>
    import libtorrent as lt
ModuleNotFoundError: No module named 'libtorrent'
```

Move lib to `/usr/lib/python3/dist-packages` in `sys.path`

```
mv /usr/lib/python3.7/site-packages/libtorrent.so /usr/lib/python3/dist-packages/libtorrent.so
```

Test deluge again.

```
deluged -V
deluged 2.0.3
libtorrent: 1.2.10.0
Python: 3.7.3
```

## Debian Stable py3 example

```
python3 -c "import sysconfig; print(sysconfig.get_path('stdlib'))"
/usr/lib/python3.7
python3 -c "import sysconfig; print(sysconfig.get_path('platlib'))"
/usr/lib/python3.7/site-packages
python3 -c "import sys; print(sys.path)"
['', '/usr/lib/python37.zip', '/usr/lib/python3.7', '/usr/lib/python3.7/lib-dynload', '/usr/local/lib/python3.7/dist-packages', '/usr/lib/python3/dist-packages']
```

## Ubuntu 20 py3 example

```
python3 -c "import sysconfig; print(sysconfig.get_path('stdlib'))"
/usr/lib/python3.8
python3 -c "import sysconfig; print(sysconfig.get_path('platlib'))"
/usr/lib/python3.8/site-packages
python3 -c "import sys; print(sys.path)"
['', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/usr/local/lib/python3.8/dist-packages', '/usr/lib/python3/dist-packages']
```

## Conclusion

The output of `python3 -c "import sys; print(sys.path)"` gives us a list of places where the module must be to be automatically loaded. I use docker official images of Debain, Ubuntu, Alpine so we are looking at what a default environment will be like.

Sure, we add to it and modify the host system to extend the `sys.path`, but i don't think that intended idea?

arvidn:
do you know where setup tools install modules to?

userdocs:
Using this

```
apt-get install python3-setuptools
python3 -m easy_install deluge
```

I got modules installed here

```
... to /usr/local/lib/python3.8/dist-packages
```

userdocs:
But now things are different,

```
python3 -c "import sys; print(sys.path)"
```

```
['', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/usr/local/lib/python3.8/dist-packages', '/usr/local/lib/python3.8/dist-packages/deluge-2.0.3-py3.8.egg', '/usr/local/lib/python3.8/dist-packages/zope.interface-5.2.0-py3.8-linux-x86_64.egg', '/usr/local/lib/python3.8/dist-packages/Twisted-20.3.0-py3.8-linux-x86_64.egg', '/usr/local/lib/python3.8/dist-packages/six-1.15.0-py3.8.egg', '/usr/local/lib/python3.8/dist-packages/setproctitle-1.1.10-py3.8-linux-x86_64.egg', '/usr/local/lib/python3.8/dist-packages/rencode-1.0.6-py3.8-linux-x86_64.egg', '/usr/local/lib/python3.8/dist-packages/pyxdg-0.27-py3.8.egg', '/usr/local/lib/python3.8/dist-packages/pyOpenSSL-19.1.0-py3.8.egg', '/usr/local/lib/python3.8/dist-packages/pyasn1-0.4.8-py3.8.egg', '/usr/local/lib/python3.8/dist-packages/Pillow-8.0.1-py3.8-linux-x86_64.egg', '/usr/local/lib/python3.8/dist-packages/Mako-1.1.3-py3.8.egg', '/usr/local/lib/python3.8/dist-packages/chardet-3.0.4-py3.8.egg', '/usr/local/lib/python3.8/dist-packages/service_identity-18.1.0-py3.8.egg', '/usr/local/lib/python3.8/dist-packages/idna-2.10-py3.8.egg', '/usr/local/lib/python3.8/dist-packages/incremental-17.5.0-py3.8.egg', '/usr/local/lib/python3.8/dist-packages/hyperlink-20.0.1-py3.8.egg', '/usr/local/lib/python3.8/dist-packages/constantly-15.1.0-py3.8.egg', '/usr/local/lib/python3.8/dist-packages/attrs-20.3.0-py3.8.egg', '/usr/local/lib/python3.8/dist-packages/PyHamcrest-2.0.2-py3.8.egg', '/usr/local/lib/python3.8/dist-packages/Automat-20.2.0-py3.8.egg', '/usr/local/lib/python3.8/dist-packages/cryptography-3.2.1-py3.8-linux-x86_64.egg', '/usr/local/lib/python3.8/dist-packages/MarkupSafe-2.0.0a1-py3.8-linux-x86_64.egg', '/usr/local/lib/python3.8/dist-packages/pyasn1_modules-0.2.8-py3.8.egg', '/usr/local/lib/python3.8/dist-packages/cffi-1.14.3-py3.8-linux-x86_64.egg', '/usr/local/lib/python3.8/dist-packages/pycparser-2.20-py3.8.egg', '/usr/lib/python3/dist-packages']
```

userdocs:
And then we have Alpine linux, being the sensible one.

```
python2 -c "import sysconfig; print(sysconfig.get_path('stdlib'))"
/usr/lib/python2.7

python2 -c "import sysconfig; print(sysconfig.get_path('platlib'))"
/usr/lib/python2.7/site-packages

python2 -c "import sys; print(sys.path)"
['', '/usr/lib/python27.zip', '/usr/lib/python2.7', '/usr/lib/python2.7/plat-linux2', '/usr/lib/python2.7/lib-tk', '/usr/lib/python2.7/lib-old', '/usr/lib/python2.7/lib-dynload', '/usr/lib/python2.7/site-packages']

python3 -c "import sysconfig; print(sysconfig.get_path('stdlib'))"
/usr/lib/python3.8

python3 -c "import sysconfig; print(sysconfig.get_path('platlib'))"
/usr/lib/python3.8/site-packages

python3 -c "import sys; print(sys.path)"
['', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/usr/lib/python3.8/site-packages']
```

userdocs:
OK, turns out this is a Debian platform specific issue. https://wiki.debian.org/Python

```
dist-packages instead of site-packages. Third party Python software installed from Debian packages goes into dist-packages, not site-packages. This is to reduce conflict between the system Python, and any from-source Python
```

userdocs:
So this is the way i understand the problem:

Debian/Ubuntu - no real way to check other than test OS. Maybe something like `cat /etc/os-release` but it must be done as these are two hugely popular distros

```
/usr/local/lib/python2.7/dist-packages
/usr/local/lib/python3.8/dist-packages
```

Everyone else using `platlib` which is pretty easy.

```
/usr/local/lib/python2.7/site-packages
/usr/local/lib/python3.8/site-packages
```

userdocs:
I think it's a good test.

```
source /etc/os-release && echo $ID
```

Via dockers.

```
debian
ubuntu
alpine
arch
fedora
```

arvidn:
@userdocs I squashed and added a new update. I don't think there needs to be any special case for debian and ubuntu. in the latest commit I use `distutils.sysconfig.get_python_lib()`, which points to the `dist-packages` on ubuntu.

userdocs:
ok, i'll test it now

userdocs:
Tested on Alpine, Debian, Ubuntu with system and user. Works fine!

I think this was a very good find `distutils.sysconfig.get_python_lib()` and it does exactly what is needed and is in the `sys.path` and i am glad we can put this particular issue behind us.

So overall i think it's complete now. I have had no bugs in all my messing around today.

Thank you for all the help for making the patch.

👍

PR #5262, opened 2020-11-05
Modify Jamfile

userdocs:
Add 3 new installation options with some logic to determine the version of python to use and where to install the libtorrent libraries.

## install_module_custom

This allows the installation of the libtorrent.so library to a specified location by exporting the custom `PYTHON_LIB_PATH` variable. If the variable is null and this flag is used a warning is printed and the jamfile exits.

## install_module_local

This will check for the major version for python used and use this command to get the `.local` lib path relative to the user's home directory for that version of python.

**Note:** `python_v` is the automatically detected major version of python.

```
[ SHELL "python$(python_v) -m site --user-site" ]
```

## install_module_system

This will check for the major version for python used and use this command to get the system lib path for global installation for that version of python.

**Note:** `python_v` is the automatically detected major version of python

`env -i` is used to ignore the env settings so we always get the same result we need.

```
[ SHELL "env -i python$(python_v) -c \"import sys; print(sys.path)\"" ] ;
```

If the variable `PYTHONPATH` is exported this will override the checks and this path will be used and `env -i` is not used

### DIFF
--- a/bindings/python/Jamfile
+++ b/bindings/python/Jamfile
@@ -20,6 +20,14 @@ feature libtorrent-link : shared static : composite propagated ;
 feature libtorrent-python-pic : off on : composite propagated link-incompatible ;
 feature.compose <libtorrent-python-pic>on : <cflags>-fPIC ;
 
+# when invoking the install_module target, this feature can be specified to
+# install the python module to a specific directory
+feature python-install-path : : free path ;
+
+# when not specifying a custom install path, this controls whether to install
+# the python module in the system directory or user-specifc directory
+feature python-install-scope : user system : ;
+
 # this is just to force boost build to pick the desired python target when using LIBTORRENT_PYTHON_INTERPRETER
 feature libtorrent-python : on ;
 
@@ -236,8 +244,8 @@ my-python-extension libtorrent
 rule python-install-dir-custom ( properties * )
 {
     local python_path = [ os.environ PYTHON_LIB_PATH ] ;
-    
-    if ! [ on $(<) return $(python_path) ] 
+
+    if ! [ on $(<) return $(python_path) ]
     {
         Echo ;
         Echo "Warning: The PYTHON_LIB_PATH cannot be empty when used with install_module_custom" ;
@@ -255,8 +263,8 @@ rule python-install-dir-local ( properties * )
     local pythons_versions = [ feature.values python ] ;
     local split_python_versions = [ SPLIT_BY_CHARACTERS $(pythons_versions) : "." ] ;
     local python_v = $(split_python_versions[1]) ;
-    local python_path = [ SHELL "python$(python_v) -c \"import site; print(site.USER_SITE, end = '')\"" ] ;
-    
+    local python_path = [ SHELL "python$(python_v) -c \"import sys; import site; sys.stdout.write(site.USER_SITE)\"" ] ;
+
 	return <location>$(python_path) ;
 }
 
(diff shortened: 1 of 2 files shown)


arvidn:
does this patch satisfy all your requirements?
For clarity, my understanding of those requirements are:

* support both python2 and python3, based on which version the module is built for (i.e. the `using python ;` statement and possibly `python=` feature specified on the build command line)
* support a manual override of the install path, regardless of which version of python is targeted.
* the ability to pick system- or user- install directory

instead of using `python -m site --user-site`, one could use python3 -c "import sys; print(sys.path[4])"`` instead, right? That might keep things a bit more uniform.

userdocs:
-  In line with the [docs]( https://libtorrent.org/python_binding.html) i use the `user-config.jam` to define python versions and this version is detected and used, for example.

```
using python : 2.7 : /usr/bin/python2.7 : /usr/include/python2.7 : /usr/lib/python2.7 ;
```
```
using python : 3.7 : /usr/bin/python3.7 : /usr/include/python3.7 : /usr/lib/python3.7 ;
```

-  Yes, `install_module_custom` is a user defined location. I think this just covers all bases.


-  Regarding the commands, i found this https://docs.python.org/3/library/site.html

The problem was that `import sys; print(sys.path)` was not giving the `USER_SITE` variable info required. It is now OK as I fund that I can use `import site; print(site.USER_SITE)` to get this exact information instead of using `python -m site --user-site`.

This also allows me to fix the new line problem in a more uniform way with a recommended method to remove python print new lines from printed strings:  `, end = ''`

This gave problematic new lines

```
import site; print(site.USER_SITE)
```

This strips the trailing new line

```
import site; print(site.USER_SITE, end = '')
```

arvidn:
for me: `python -c "import sys; print(sys.path[4])"` prints the exact same thing as `import site; print(site.USER_SITE)`. Is there a meaningful difference?

userdocs:
Yes, because it must exist first or `print(sys.path[4])` will return a different result. So I could assume it exists in the environment you are testing the command.

I am testing on new docker containers where this directory doesn't exist until the `install_module_local` command creates the directory structure. Therefore using `print(sys.path[4])` is not going to give me a meaningful result for this path.

Here are some examples to demonstrate the issue.

## Alpine 3.12 

Use `import sys; print(sys.path)` to see the paths. `USER_SITE` is missing.

```
python3 -c "import sys; print(sys.path)"
['', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/usr/lib/python3.8/site-packages']
```

Get the path of `USER_SITE` using `import site; print(site.USER_SITE)`

```
python3 -c "import site; print(site.USER_SITE)"
/home/username/.local/lib/python3.8/site-packages
```

Create this directory.

```
mkdir -p "$(python3 -c "import site; print(site.USER_SITE)")"
```

Use `import sys; print(sys.path)` to see the paths. `USER_SITE` is present.

```
python3 -c "import sys; print(sys.path)"
['', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/home/username/.local/lib/python3.8/site-packages', '/usr/lib/python3.8/site-packages']
```

## Ubuntu 20.04

Use `import sys; print(sys.path)` to see the paths. `USER_SITE` is missing.

```
python3 -c "import sys; print(sys.path)"
['', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/usr/local/lib/python3.8/dist-packages', '/usr/lib/python3/dist-packages']
```

Get the path of `USER_SITE` using `import site; print(site.USER_SITE)`

```
mkdir -p "$(python3 -c "import site; print(site.USER_SITE)")"
```

Create this directory.

```
python3 -c "import site; print(site.USER_SITE)"
/home/username/.local/lib/python3.8/site-packages
```

Use `import sys; print(sys.path)` to see the paths. `USER_SITE` is present.

```
python3 -c "import sys; print(sys.path)"
['', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/home/username/.local/lib/python3.8/site-packages', '/usr/local/lib/python3.8/dist-packages', '/u
```

Does that help?

PR #5264, opened 2020-11-08
Francisco pombal gha windows cmake rc 1 2 pr

arvidn:


FranciscoPombal:
Hey @arvidn, I haven't abandoned the other PR. I see you are trying to see what happens by always building with tests, but there is a problem - static debug builds with tests will always fail, because there is not enough space. One way I thought of working around this would be to keep the `[tests, notests]` distiction, but manually exclude all the `notests` cases.

For example, here is an attempt to always build with python bindings: https://github.com/FranciscoPombal/libtorrent/actions/runs/342725904/workflow. Since the `[static, py]` combination is already excluded, no `static` variant gets built at all. I will try with the exclusions.

arvidn:
yeah, I'm thinking; if there's a way to enumerate all the combinations to include instead, that might be simpler. i.e. a whitelist rather than a blacklist

arvidn:
@FranciscoPombal it looks like python3 is installed on the runners. do you know which path `python.exe` is installed to?

FranciscoPombal:
@arvidn I managed to trim the matrix down with a mix of include and exclude: https://github.com/FranciscoPombal/libtorrent/runs/1370605586?check_suite_focus=true

It appears that multiple values for keys in the `include` stanza don't work as expected (for example, `foo: [bar, baz]` won't unfold into 2 build variants). So it becomes really impractical to manually specify all whitelisted configs.

That run I linked also includes setting up python (2 different versions, as an example). Now the only thing left to figure out is what is causing all of the `shared` tests to fail.

arvidn:
https://github.com/FranciscoPombal/libtorrent/runs/1370605620?check_suite_focus=true#step:14:546

Those failures are because `python` can't be found. It could be resolved by setting the `PYTHON_INTERPRETER` environment variable to include the full path to `python.exe`

PR #5268, opened 2020-11-10
add attr v2 metadata when making a single file torrent

hyqhyq3:
This is a bug fix.
When making a single executable file torrent in linux. The torrent file's v2 metadata lack of 'attr' , cause libtorrent cannot load the torrent. libtorrent will report "v1 v2 metadata mismatch".

Step:
1. on linux platform
2. touch test_file
3. chmod +x test_file
4. make_torrent test_file -o test.torrent
5. dump_torrent test.torrent  # error happened

* make_torrent and dump_torrent is the example of official site. https://libtorrent.org/examples.html

### DIFF
--- a/src/create_torrent.cpp
+++ b/src/create_torrent.cpp
@@ -712,22 +712,23 @@ namespace {
 		if (!m_multifile)
 		{
 			file_index_t const first(0);
+			std::string attr;
+			file_flags_t const flags = m_files.file_flags(first);
+			if (flags & (file_storage::flag_pad_file
+				| file_storage::flag_hidden
+				| file_storage::flag_executable
+				| file_storage::flag_symlink))
+			{
+				if (flags & file_storage::flag_pad_file) attr += 'p';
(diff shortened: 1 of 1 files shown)


arvidn (approved):

arvidn:
There should be a test covering this case too. I can add one after this lands

arvidn:
I'm a bit tempted to solve this by unifying the multi-file and single-file cases for v2. Right now the same thing is done in 4 places, essentially. single-file v1, multi-file v1, single-file v2 and multi-file v2. the v2 format is more regular and could use the same logic for both single and multi-file.

arvidn:
@hyqhyq3 I really appreciate your contribution, but I may end up landing my alternative fix instead ([here](https://github.com/arvidn/libtorrent/pull/5270/files?diff=split&w=1)).

It fixes the issue by reducing the number of cases in `torrent_creator` from 4 to 3 (two v1 cases and now one, unified, v2 case).

Issue #5125, opened 2020-09-09
Path duplication in libtorrent-rasterbar.pc


lgbaldoni:
**Please provide the following information**

> libtorrent version (or branch):

2.0

>platform/architecture:

openSUSE Tumbleweed x86_64

> compiler and compiler version:

gcc 10

> please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

Build with:
```
[...]

PR #5271, opened 2020-11-11
fix cmake install directory for absolute paths

arvidn:


arvidn:
@krop @zeule is this an improvement?

zeule:
> @krop @zeule is this an improvement?

What about other `CMAKE_INSTALL_xxxDIR` variables?

arvidn:
well, I don't really know what I'm doing. I'm just trying to interpret some conclusion from https://github.com/arvidn/libtorrent/issues/5125

zeule:
CMake defines [variables](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) for all the common destination dirs. Correcting one of them and leaving others as is might be surprising. We use two of the variables (LIB and INCLUDE) and `CMAKE_INSTALL_PREFIX` to generate compiler flags. Maybe we'd better somehow utilize `GNUInstallDirs_get_absolute_install_dir()` to get the absolute paths?

PR #5291, opened 2020-11-18
fix concurrent access to part files

arvidn:
by using pread/overlapped I/O instead of seek+read. This used to work by simply restricting multi-threading when pread wasn't supported


arvidn:
`pread()` and `pwrite()` are part of posix, so it should be pretty safe to rely on them for non-windows, and overlapped structure for win32.

PR #5312, opened 2020-11-29
Extend std::hash for lt::digest32<N> type

Chocobo1:


include/libtorrent/sha1_hash.hpp
```diff
@@ -298,13 +298,13 @@ namespace libtorrent {
...
+	template <std::ptrdiff_t N>
+	struct hash<libtorrent::digest32<N>>
 	{
-		std::size_t operator()(libtorrent::sha1_hash const& k) const
+		std::size_t operator()(libtorrent::digest32<N> const& k) const
 		{
 			std::size_t ret;
-			// this is OK because sha1_hash is already a hash
+			// this is OK because digest32<N> is already a hash
 			std::memcpy(&ret, k.data(), sizeof(ret));
```
arvidn, line 308: this is only correct as long as `sizeof(ret) <= k.size()`. If `N == 1` and `sizeof(std::size_t) == 8`, this will copy 4 bytes past the end of the `digest32` object.

The simplest way to address this is probably to stick a
```
static_assert(N * 4 >= sizeof(ret), "hash is not defined for small digests");
```

in there. `digest32` (as the name implies) is sized in increments of 32bit words, i.e. 4 bytes. It's possible to extend this in the future with a template specialization.

Chocobo1: What about this slightly fancier version?
```c++
namespace std {
	template <std::ptrdiff_t N>
	struct hash<libtorrent::digest32<N>>
	{
		std::size_t operator()(libtorrent::digest32<N> const& k) const
		{
			std::size_t ret = 0;
			// this is OK because digest32<N> is already a hash
			std::memcpy(&ret, k.data(), std::min(N / 8, sizeof(ret)));
			return ret;
		}
	};
}
```

>The simplest way to address this is probably to stick a
`static_assert(N * 4 >= sizeof(ret), "hash is not defined for small digests");`
in there. 

I suppose you meant `static_assert(N >= sizeof(ret) * 8, "hash is not defined for small digests");` right? Such that `N` and `sizeof(ret) * 8` are in bits unit.

arvidn: ah, right. `N` is number of *bits*, not words. you can even use `CHAR_BIT` instead of 8

Chocobo1: PR updated.

> you can even use CHAR_BIT instead of 8

I see you defined `digest32::bits_in_byte` but it is a private member. I'll just use `8` here.

arvidn (approved):

arvidn:
The benefit of using `CHAR_BIT` instead of 8 is not to support C++ implementations where `char` does not have 8 bit, but to make it clear to the reader that 8 means the number of bits in a `char`.

PR #5314, opened 2020-11-29
validate HTTPS tracker certificates by default

arvidn:


arvidn:
@FranciscoPombal @cas-- would this have serious negative impact you think?
I'm ignorant about how common it is for HTTPS trackers to use self-signed certificates, do you have a sense?
There would probably need to be a switch to disable this by users, I imagine.

FranciscoPombal:
@arvidn 

>I'm ignorant about how common it is for HTTPS trackers to use self-signed certificates, do you have a sense?

From this list extracted from https://newtrackon.com/list at time of writing, the vast majority (10/11) are using valid HTTPS certs:

```
https://tracker.sloppyta.co:443/announce
https://tracker.nitrix.me:443/announce
https://tracker.imgoingto.icu:443/announce
https://tracker.lilithraws.cf:443/announce
https://tracker.foreverpirates.co:443/announce
https://tracker.tamersunion.org:443/announce
https://w.wwwww.wtf:443/announce
https://trakx.herokuapp.com:443/announce
https://tracker.cyber-hub.net:443/announce
https://1337.abcvg.info:443/announce
https://tracker.hama3.net:443/announce
```

IIRC most/all major private trackers are also using valid certs. It's easier than ever to get valid HTTPS certs nowadays thanks to Let's Encrypt - plus the other vendors are now forced to offer a competing alternative - so I would expect the trend to be for more and more trackers to go with valid certs by default, right off the bat.

>There would probably need to be a switch to disable this by users, I imagine.

For sure. Perhaps `"also allow self-signed certificates" | "only allow valid certs" | "don't validate"`, as opposed to the all-or-nothing `"validate" | "don't validate"`? This way, there can also be an alert to inform users which HTTPS trackers are using self signed certs, without completely disallowing them.

ArcticGems:
> 
> 
> @arvidn
> 
> > I'm ignorant about how common it is for HTTPS trackers to use self-signed certificates, do you have a sense?
> 
> From this list extracted from https://newtrackon.com/list at time of writing, the vast majority (10/11) are using valid HTTPS certs:
> 

That list must change allot since just now when I checked, there are allot of udp/http **public** trackers and that seems to match what my torrents has as well (allot of udp/http **public** trackers).


> 
> > There would probably need to be a switch to disable this by users, I imagine.
> 
> For sure. Perhaps `"also allow self-signed certificates" | "only allow valid certs" | "don't validate"`, as opposed to the all-or-nothing `"validate" | "don't validate"`? This way, there can also be an alert to inform users which HTTPS trackers are using self signed certs, without completely disallowing them.

Definitely needs a three choice options as FranciscoPombal suggested if this PR gets merged.

arvidn:
I'm not sure how "also allow self-signed certificates" would differ from "don't validate". There wouldn't be a root certificate to validate the self-signed cert against anywhere.

As a side note, I'm tempted to implement an extension where the root cert for trackers can be embedded in the torrent, to further strengthen the validation.

FranciscoPombal:
> I'm not sure how "also allow self-signed certificates" would differ from "don't validate". There wouldn't be a root certificate to validate the self-signed cert against anywhere.

"also allow self-signed certificates" provides the flexibility of accepting everything while also allowing valid/invalid certs to be distinguished. That information can then be posted as an alert, for example, as mentioned.

> As a side note, I'm tempted to implement an extension where the root cert for trackers can be embedded in the torrent, to further strengthen the validation.

An extension/derivative of https://www.bittorrent.org/beps/bep_0035.html ?

imgastro:
Hello how can I track someone

arvidn:
> An extension/derivative of https://www.bittorrent.org/beps/bep_0035.html ?

Yes. Specifically picking up an x509 cert from the .torrent and requiring it as the root (when available) would strengthen authentication of trackers. I still think this would be a good idea, but I've not made any progress on this feature.

PR #5325, opened 2020-12-02
Better integration of setup.py to invoke b2

AllSeeingEyeTolledEweSew:
Hi,

I rewrote `setup.py` in the python bindings to be better-integrated with the b2 build system.

My use case is that I want to more easily build [wheels](https://realpython.com/python-wheels/) of libtorrent. Specifically, I want to use [tox](https://tox.readthedocs.io/en/latest/) and [pre-commit](https://pre-commit.com) in my libtorrent-based python project. These (and other tools) set up isolated environments and try to install dependencies, and since libtorrent isn't published on pypi, the easiest way to make these tools work is run a local pypi instance which provides access to libtorrent.

It was possible to make wheels before, but this integration fix makes it much easier. I now can build wheels with custom b2 compilation, like so:
```shell
$ python setup.py build_ext -j9 --debug --libtorrent-link=static bdist_wheel
```

I deleted all the existing build infrastructure from `setup.py` as I don't believe it was very useful: the `--bjam` mode didn't seem to actually integrate with distutils. It just invoked b2 from the global level, but didn't allow for passing options. The non-`--bjam` mode seemed very limited and wasn't referenced at all in [the build docs](http://libtorrent.org/python_binding.html).

I can't find any reference that `setup.py` is actually used by anyone today; debian and alpine's build scripts just invoke autotools and don't use `setup.py`. This is appropriate for them, but setuptools is a better choice for packaging in the distro-agnostic python world.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -1,0 +1,44 @@
+name: Python bindings
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: build
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, macos-latest]
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: recursive
+
+    - name: dependencies (macos)
+      if: runner.os == 'macOS'
+      run: |
+        brew install boost-build boost boost-python3 python@3.9
+
+    - name: update package lists
+      if: runner.os == 'Linux'
+      run: |
+        sudo apt update
+
+    - name: dependencies (linux)
+      if: runner.os == 'Linux'
+      run: |
+        sudo apt install libboost-tools-dev libboost-python-dev libboost-dev libboost-system-dev python3 python3-setuptools
+
+    - name: build/install
+      run: |
+        cd bindings/python
+        # Homebrew's python "framework" sets a prefix via distutils config.
+        # --prefix conflicts with --user, so null out prefix so we have one
+        # command that works everywhere
+        python3 setup.py build_ext -j3 --libtorrent-link=static install --user --prefix=
+
+    - name: tests
+      run: |
+        cd bindings/python
+        python3 test.py
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -1,195 +1,219 @@
 #!/usr/bin/env python3
 
-
-from distutils.core import setup, Extension
-from distutils.sysconfig import get_config_vars
+import distutils.debug
 import os
+import pathlib
 import platform
 import sys
-import shutil
-import multiprocessing
+import sysconfig
+import tempfile
+
+import setuptools
+import setuptools.command.build_ext as _build_ext_lib
 
 
-class flags_parser:
-    def __init__(self):
-        self.include_dirs = []
-        self.library_dirs = []
-        self.libraries = []
-
-    def parse(self, args):
-        """Parse out the -I -L -l directives
-
-        Returns:
-            list: All other arguments
-        """
-        ret = []
-        for token in args.split():
-            prefix = token[:2]
-            if prefix == '-I':
-                self.include_dirs.append(token[2:])
-            elif prefix == '-L':
-                self.library_dirs.append(token[2:])
-            elif prefix == '-l':
-                self.libraries.append(token[2:])
-            else:
-                ret.append(token)
-        return ret
+def get_msvc_toolset():
+    # Reference: https://wiki.python.org/moin/WindowsCompilers
+    major_minor = sys.version_info()[0:2]
+    if major_minor in ((2, 6), (2, 7), (3, 0), (3, 1), (3, 2)):
+        return "msvc-9.0"
+    if major_minor in ((3, 3), (3, 4)):
+        return "msvc-10.0"
+    if major_minor in ((3, 5), (3, 6)):
+        return "msvc-14.1"  # libtorrent requires VS 2017 or newer
+    # unknown python version
+    return "msvc"
 
 
-def arch():
-    if platform.system() == 'Darwin':
-        __, __, machine = platform.mac_ver()
-        if machine.startswith('ppc'):
-            return ['-arch', machine]
-    return []
+def b2_bool(value):
+    if value:
+        return "on"
+    return "off"
 
 
-def target_specific():
-    if platform.system() == 'Darwin':
-        # On mavericks, clang will fail when unknown arguments are passed in.
-        # python distutils will pass in arguments it doesn't know about.
-        return ['-Wno-error=unused-command-line-argument-hard-error-in-future']
-    return []
+# Frustratingly, the "bdist_*" unconditionally (re-)run "build" without
+# args, even ignoring "build_*" earlier on the same command line. This
+# means "build_*" must be a no-op if some build output exists, even if that
+# output might have been generated with different args (like
+# "--define=FOO"). b2 does not know how to be "naively idempotent" like
+# this; it will only generate outputs that exactly match the build request.
+#
+# It doesn't work to short-circuit initialize_options() / finalize_options(),
+# as this doesn't play well with the way options are externally manipulated by
+# distutils.
+#
+# It DOES work to short-circuit Distribution.reinitialize_command(), so we do
+# that here.
 
 
-try:
-    with open('compile_flags') as _file:
-        extra_cmd = _file.read()
-except Exception:
-    extra_cmd = None
+class B2Distribution(setuptools.Distribution):
+    def reinitialize_command(self, command, reinit_subcommands=0):
+        if command == "build_ext":
+            return self.get_command_obj("build_ext")
+        return super().reinitialize_command(
+            command, reinit_subcommands=reinit_subcommands
+        )
 
-try:
-    with open('link_flags') as _file:
-        ldflags = _file.read()
-except Exception:
-    ldflags = None
 
-# this is to pull out compiler arguments from the CXX flags set up by the
-# configure script. Specifically, the -std=c++11 flag is added to CXX and here
-# we pull out everything starting from the first flag (i.e. something starting
-# with a '-'). The actual command to call the compiler may be more than one
-# word, for instance "ccache g++".
-try:
-    with open('compile_cmd') as _file:
-        cmd = _file.read().split(' ')
-        while len(cmd) > 0 and not cmd[0].startswith('-'):
-            cmd = cmd[1:]
-        extra_cmd += ' '.join(cmd)
-except Exception:
-    pass
+# Various setuptools logic expects us to provide Extension instances for each
+# extension in the distro.
+class StubExtension(setuptools.Extension):
+    def __init__(self, name):
+        # An empty sources list ensures the base build_ext command won't build
+        # anything
+        super().__init__(name, sources=[])
 
-ext = None
-packages = None
 
-if '--bjam' in sys.argv:
-    del sys.argv[sys.argv.index('--bjam')]
+def write_b2_python_config(config):
+    write = config.write
+    # b2 normally keys python environments by X.Y version, but since we may
+    # have a duplicates, we also key on a special property. It's always-on, so
+    # any python versions configured with this as a condition will always be
+    # picked.
+    # Note that python.jam actually modifies the condition for the build
+    # request, so that <define>TORRENT_FOO becomes something like
+    # <define>TORRENT_FOO,<python>3.7,<target-os>linux:... which causes chaos.
+    # We should always define a custom feature for the condition.
+    write("import feature ;\n")
+    write("feature.feature libtorrent-python : on ;\n")
 
-    if '--help' not in sys.argv \
-            and '--help-commands' not in sys.argv:
+    # python.jam tries to determine correct include and library paths. Per
+    # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=691378 , include
+    # detection is broken, but debian's fix is also broken (invokes a global
+    # pythonX.Y instead of the passed interpreter)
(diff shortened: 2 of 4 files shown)


AllSeeingEyeTolledEweSew:
I'm only now just seeing #4218 ...

arvidn:
it would be great if you could extend the GH action for building the python binding to use this, and maybe even upload the package as an artifact

AllSeeingEyeTolledEweSew:
> it would be great if you could extend the GH action for building the python binding to use this, and maybe even upload the package as an artifact

Interesting. I'd love to achieve the same goals as #4218, but it's a big PR, and I'd at least like to split out the artifact building part. I think it's going to get hairy.

Given that you have this:

https://github.com/arvidn/libtorrent/blob/f726d1089147dd5fdf7674782bb6fac8cd3e508f/bindings/python/Jamfile#L137-L140

If this is true (though [the boost-python docs](https://www.boost.org/doc/libs/1_74_0/libs/python/doc/html/building/choosing_a_boost_python_library_.html) don't mention it), then to build an appropriate linux wheel we need boost-python 1.74, which isn't in ubuntu yet, so we need a local download-and-install script as in #4218 . It just seems a bit much for this PR

arvidn:
sure, I don't want to bloat a single PR. But, if this code you wrote isn't running on CI, how do I know it works?

.github/workflows/macos.yml
```diff
@@ -93,19 +93,16 @@ jobs:
...
           brew install boost-build boost boost-python3 python@3.9
-          export PYTHON_INCLUDE=`echo /usr/local/Cellar/python@3.9/*/Frameworks/Python.framework/Versions/3.9/include/python3.9` ;
-          echo "using darwin ;" >>~/user-config.jam
-          echo "using python : 3.9 : /usr/local/opt/python@3.9/bin/python3 : $(PYTHON_INCLUDE) : /usr/local/opt/python@3.9/Frameworks/Python.framework/Versions/3.9/lib/python3.9/config-3.9-darwin : <target-os>darwin ;" >> ~/user-config.jam
 
-      - name: build python module
+      - name: build and install python module
         run: |
           cd bindings/python
-          b2 -j2 stage_module stage_dependencies cxxstd=11
```
arvidn, line 106: the build is failing be cause `cxxstd=11` isn't passed to `b2` right now it seems

.github/workflows/linux.yml
```diff
@@ -177,25 +177,19 @@ jobs:
         with:
            submodules: recursive
 
-      - name: install boost
-        run: |
-          sudo apt install libboost-tools-dev libboost-python-dev libboost-dev libboost-system-dev
-          echo "using gcc ;" >>~/user-config.jam
-          echo "using python ;" >>~/user-config.jam
```
arvidn, line 184: Is there a good reason not to configure boost build? It seems risky to rely on its default heuristic to pick gcc

.github/workflows/macos.yml
```diff
@@ -93,19 +93,16 @@ jobs:
         with:
            submodules: recursive
 
-      - name: install boost
+      - name: install dependencies
         run: |
           brew install boost-build boost boost-python3 python@3.9
-          export PYTHON_INCLUDE=`echo /usr/local/Cellar/python@3.9/*/Frameworks/Python.framework/Versions/3.9/include/python3.9` ;
-          echo "using darwin ;" >>~/user-config.jam
-          echo "using python : 3.9 : /usr/local/opt/python@3.9/bin/python3 : $(PYTHON_INCLUDE) : /usr/local/opt/python@3.9/Frameworks/Python.framework/Versions/3.9/lib/python3.9/config-3.9-darwin : <target-os>darwin ;" >> ~/user-config.jam
```
arvidn, line 101: Are you confident boost build will find the correct python headers and library without configuring it here?

arvidn: in fact, it looks like it's building against python 2.6 in the github action now, since you removed this configuration step

.github/workflows/macos.yml
AllSeeingEyeTolledEweSew: oh, is `cxxstd=11` required on macos right now?

`setup.py` could detect and pass this, but should the jamfile enforce it instead? :thinking:

.github/workflows/linux.yml
AllSeeingEyeTolledEweSew: This was intentional; the typical expectation with setuptools is that it should be able to pick sane defaults without special configuration (except for installing dependencies to build extensions).

I could add a separate test to pass `toolset=gcc`. I'm still trying to get a feel for what other specific feature properties to test.

.github/workflows/macos.yml
AllSeeingEyeTolledEweSew: I'm confident that I want this to be the test :)

Another expectation with setuptools is that it builds the package for the running interpreter, so ideally `setup.py` or the jamfile should be able to work out these paths. I don't have that much experience with macos though, so I'm not sure if this expectation is reasonable.

arvidn: ok, makes sense. I would *also* like to test that the python binding builds using `b2` using the boost build user configuration.

.github/workflows/macos.yml
arvidn: in later version (`RC_2_0` and `master`) the Jamfile defaults to cxxstd=14 and cxxstd=17 respectively. I just haven't made that change to RC_1_2 because it supports being built with an old version of boost where the `cxxstd` option didn't exist yet

arvidn:
I think [this](https://github.com/arvidn/libtorrent/pull/5325/checks?check_run_id=1508247987#step:4:4463) error looks like it's caused by setuptools trying to build the extension. That's not the intention, right? it should use the binary built by `b2`

AllSeeingEyeTolledEweSew:
> I think [this](https://github.com/arvidn/libtorrent/pull/5325/checks?check_run_id=1508247987#step:4:4463) error looks like it's caused by setuptools trying to build the extension. That's not the intention, right? it should use the binary built by `b2`

Yes. Looks like I actually needed to pass the extension suffix to boost so its output is recognized by setuptools; the debian version shadowed this problem.

I'm doing most of my testing on a debian-based arch and I'm finding that debian has patched the daylights out of `python.jam` from boost-build, with prety significant behavior changes. It makes me think that all the CI should be run on non-ubuntu linux as well as ubuntu, or else install boost from source.

AllSeeingEyeTolledEweSew:
I made `setup.py` default to passing `cxxstd=11`, only on osx. Hopefully this is appropriate.

I put `setup.py` invocation on osx in a separate github action, which is passing now.

It's not currently invoked on windows. Should I create a separate appveyor matrix entry to invoke `setup.py`, separate from one that builds the binding with b2? Appveyor already takes forever.

I doubt the linux CI failure is a result of this PR.

AllSeeingEyeTolledEweSew:
I rebased to make sure the CI failure wasn't mine.

@arvidn could you advise on whether you'd like to use appveyor or github actions for new windows CI?

FranciscoPombal:
@AllSeeingEyeTolledEweSew 

> I rebased to make sure the CI failure wasn't mine.
> 
> @arvidn could you advise on whether you'd like to use appveyor or github actions for new windows CI?

I was working on GH Actions CI for Windows + CMake here https://github.com/arvidn/libtorrent/pull/5197, I should finish it relatively soon.

.github/workflows/linux.yml
arvidn: I would like to preserve the test of `b2` in the `bindings/python` directory. would you mind restoring this and adding another job that runs `setup.py`?

.github/workflows/macos.yml
```diff
@@ -93,7 +93,31 @@ jobs:
...
+        run: |
+          cd bindings/python
+          /usr/local/opt/python@3.9/bin/python3.9 setup.py build_ext -j3 --libtorrent-link=static install
+
+      - name: run python tests
+        run: |
+          cd bindings/python
+          /usr/local/opt/python@3.9/bin/python3.9 test.py
+
+   python_b2:
```
arvidn, line 110: it looks like you preserved the existing test here. would you mind re-ordering these jobs to make the diff smaller, and reflect the fact that the existing test is unchanged?

arvidn:
lgtm, except the github workflow changes

.github/workflows/python.yml
```diff
@@ -0,0 +1,36 @@
...
+
+    - name: install dependencies (mac)
+      if: runner.os == 'macOS'
+      run: |
+        brew install boost-build boost boost-python3 python@3.9
+
+    - name: install dependencies (linux)
+      if: runner.os == 'Linux'
+      run: |
+        sudo apt install libboost-tools-dev libboost-python-dev libboost-dev libboost-system-dev python3 python3-setuptools
```
arvidn, line 26: you should probably rebase now that https://github.com/arvidn/libtorrent/pull/5749 landed, as well as add those `sudo apt update` here as well

.github/workflows/linux.yml
arvidn: I would like to keep this `python` job. would you mind just reverting this file?

bindings/python/setup.py
```diff
@@ -1,195 +1,219 @@
...
+    boolean_options = BuildExtBase.boolean_options + ["pic", "hash"]
+
+    def initialize_options(self):
+        self.libtorrent_link = None
+        self.boost_link = None
+        self.toolset = None
+        self.pic = None
+        self.optimization = None
+        self.hash = None
+        if platform.system() == "Darwin":
```
arvidn, line 141: I don't think this distinction is important. You just happen to have an older compiler on MacOS that doesn't default to C++11. The more interesting question is, which ABI is python itself use? perhaps C++11 is too old.

AllSeeingEyeTolledEweSew: > I don't think this distinction is important. You just happen to have an older compiler on MacOS that doesn't default to C++11.

I see. Do you have opinions on whether it would be better to default `toolset=gcc` on darwin, rather than default `cxxstd=11`? I'll play with this.

I guess what's really needed is a default of `cxxstd=11` if `toolset=darwin`, but `setup.py` doesn't know if `toolset=darwin` will be autodiscovered in b2. IMO would be much better to just apply this in some `Jamfile` in RC_1_2, rather than here.

> The more interesting question is, which ABI is python itself use? perhaps C++11 is too old.

cpython's ABI is C only. So `cxxstd`'s ABI just matters for the extension's downstream dynamic linkage. Ideally, when packaging a wheel for pypi, the choice of `cxxstd` should never matter to consumers.

(There are various C ABIs defined, but `setup.py`'s scope is to only know about the ABI of the running interpreter, and configure boost to build for that)

AllSeeingEyeTolledEweSew:
I think it's ready now.

Note that I don't currently have a windows workflow to invoke `setup.py`, because figuring out how to install boost is daunting. Once #5197 lands, I'll reuse its setup methods to extend my `python.yml`.

tools/set_version.py
```diff
@@ -48,9 +48,9 @@ def substitute_file(name):
         elif 'VERSION=' in line and name.endswith('build_dist.sh'):
             line = 'VERSION=%d.%d.%d\n' % (version[0], version[1], version[2])
         elif 'version=' in line and name.endswith('setup.py'):
-            line = "    version='%d.%d.%d',\n" % (version[0], version[1], version[2])
+            line = "    version=\"%d.%d.%d\",\n" % (version[0], version[1], version[2])
         elif "version = '" in line and name.endswith('setup.py'):
-            line = "    version='%d.%d.%d',\n" % (version[0], version[1], version[2])
+            line = "    version=\"%d.%d.%d\",\n" % (version[0], version[1], version[2])
```
arvidn, line 53: it looks like this line is duplicated. I should remove one of them once this lands

arvidn (approved):

arvidn:
great! are these commits squashed to reasonable changes? (if not, please do that and I'll land it afterwards)

AllSeeingEyeTolledEweSew:
> great! are these commits squashed to reasonable changes? (if not, please do that and I'll land it afterwards)

I tried to keep them clean, but failed by the end. I squashed to one as it's pretty straightforward.

bindings/python/setup.py
arvidn: > IMO would be much better to just apply this in some Jamfile in RC_1_2, rather than here.

It is in RC_2_0 and later

> cpython's ABI is C only. So cxxstd's ABI just matters for the extension's downstream dynamic linkage.

Which C++ standard is used affects which C standard the compiler is using as well. The C standard doesn't define an ABI, just like C++ doesn't. It may be easier for platforms to maintain a stable ABI in C, but there's no guarantee. windows/MSVC is a good example.

I imagine specifying `stdcxx=11` here is safe, but with MSVC, the version of the compiler depends on the version of python.

AllSeeingEyeTolledEweSew: In truth I don't think we need to work too hard to solve this here.

I don't think invoking `setup.py` with no args needs to produce _the_ perfect most-compatible build. When packaging wheels for pypi, I expect we'd do things like `--libtorrent-link=static --boost-link=static` which wouldn't be appropriate in most other cases.

I just wanted `setup.py` with no args to produce _something_ useful, both to help users, and to trim down os-specific overrides in CI. I expect the impact of changing this default later will be low.

AllSeeingEyeTolledEweSew: For what it's worth, I haven't seen evidence that python is required to be compiled with any particular C standard on mac/linux.

On linux particularly, I struggle to think of any problems that would be caused for an executable linking to a library, if they differed in C standard alone. I would expect versioned symbols in libc to resolve the differences (in a perfect world). I don't have much macos experience so I don't know if they have something similar.

PR #5341, opened 2020-12-11
fix _WIN32_WINNT minimum version checking

bemaru:
- OnLinkPrefixLength is supported from vista(_WIN32_WINNT==0x0600) and above.
- IP_ADAPTER_UNICAST_ADDRESS is IP_ADAPTER_UNICAST_ADDRESS_XP in Windows XP, but it does not have OnLinkPrefixLength.
- Microsoft also recommends checking NTDDI_VERSION >= NTDDI_VISTA, _WIN32_WINNT >= 0x0600, or WINVER >= 0x0600 when using IP_ADAPTER_UNICAST_ADDRESS.
- ref : https://docs.microsoft.com/en-us/windows/win32/api/iptypes/ns-iptypes-ip_adapter_unicast_address_lh

### DIFF
--- a/src/enum_net.cpp
+++ b/src/enum_net.cpp
@@ -848,7 +848,7 @@ int _System __libsocket_sysctl(int* mib, u_int namelen, void *oldp, size_t *oldl
 
 #elif TORRENT_USE_GETADAPTERSADDRESSES
 
-#if _WIN32_WINNT >= 0x0501
+#if _WIN32_WINNT >= 0x0600
 		using GetAdaptersAddresses_t = ULONG (WINAPI *)(ULONG,ULONG,PVOID,PIP_ADAPTER_ADDRESSES,PULONG);
 		// Get GetAdaptersAddresses() pointer
 		auto GetAdaptersAddresses =


arvidn:
I suppose the way the code is written now allows you to build on vista and run on XP, but not build on XP

ssiloti:
Yes, the indent of the comparison of length to the offset of `OnLinkPrefixLength` is to detect at runtime if the code is running on Windows XP. With this updated version check we can remove the runtime check since the code will now only run on Vista and newer anyways.

Or the compile time check could be removed too and the documentation updated to explicitly disclaim support for XP.

bemaru:
As far as I know, the latest version of libtorrent no longer supports Windows XP.
So it doesn't seem to be necessary to check for Windows XP.

Not only that, like ip_notifier.cpp, libtorrent is already using winapi above vista without checking _WIN32_WINNT.
(Like NotifyUnicastIpAddressChange, CancelMibChangeNotify2)

If libtorrent does not support XP, I think it is correct to use it without checking like in ip_notifier.cpp.

bemaru:
The code location of _WIN32_WINNT check is as follows.

>   D:\git\torrent\libtorrent_v2.0.1\include\libtorrent\config.hpp(206):# if _WIN32_WINNT >= 0x0600
  D:\git\torrent\libtorrent_v2.0.1\include\libtorrent\config.hpp(221):# if _WIN32_WINNT >= 0x0600
  D:\git\torrent\libtorrent_v2.0.1\include\libtorrent\config.hpp(251):# if _WIN32_WINNT >= 0x0600
  D:\git\torrent\libtorrent_v2.0.1\include\libtorrent\config.hpp(266):# if _WIN32_WINNT >= 0x0600
  D:\git\torrent\libtorrent_v2.0.1\src\enum_net.cpp(851):#if _WIN32_WINNT >= 0x0501
  D:\git\torrent\libtorrent_v2.0.1\src\enum_net.cpp(1305):#if _WIN32_WINNT >= 0x0600

Before doing the refactoring to remove the _WIN32_WINNT check (I don't know if it will be accepted),
Separately, shouldn't we unify the part that checks as 0x0501 for _WIN32_WINNT as 0x0600?

arvidn:
in my mind, having all parts of libtorrent agree on which is the oldest windows version they support *agree* is not in itself terribly important. I think it's more important that each piece of code makes sense with itself, regarding version checks and the code it uses. Also, anything to simplify the code by raising the assumed version for instance, is an improvement of course.

In this patch for example, it would make sense to either raise the version limit and remove the run-time check could be removed *or* the runtime check could be fixed to work on the oldest version.

PR #5348, opened 2020-12-13
Include necessary header

Chocobo1:
Because `operator<<()` requires it


arvidn:
interesting. On CI I build all headers as independent object files specifically to ensure they all are self-contained with regards to includes. maybe I don't build with `TORRENT_USE_IOSTREAM`

PR #5741, opened 2020-12-14
Bump minimum required boost version in master to 1.75.0

FranciscoPombal:
Allows for some `#if` cleanups + some cleanups in the build scripts; this is sort of a follow-up to d3216f63aeea558ed54c8bcb2dc9ab112db46577.

Let me know if there is anything wrong with the changes to the Jamfile, I'm not to familiar.

Also, I'm a little unsure about the change from `m_read_buffer.consume(m_read_buffer.size());` to `m_read_buffer.clear();` in `src/websocket_tracker_connection.cpp`.The docs for the former and latter method read, respectively (emphasis mine):

>Remove bytes from beginning of the **readable bytes**.
Removes n bytes from the beginning of the readable bytes.
All buffers sequences previously obtained using
`data` or `prepare` become invalid.
>
>`n`: The number of bytes to remove. If this number is greater than the number of readable bytes, all readable bytes are removed

>Set the size of the **readable and writable bytes** to zero.
This clears the buffer without changing capacity.
Buffer sequences previously obtained using `data` or
`prepare` become invalid.
>
>No-throw guarantee.

I did not study the code in detail, so I don't know if this makes any difference. But from a cursory look, it makes sense and seems to be OK, since in the libtorrent code I only see this buffer being read from, not written to, so the part about "and writable..." in the documentation should have no effect on the current behavior.

### DIFF
--- a/bindings/python/Jamfile
+++ b/bindings/python/Jamfile
@@ -31,41 +31,20 @@ feature python-install-path : : free path ;
 # the python module in the system directory or user-specifc directory
 feature python-install-scope : user system : ;
 
-# copied from boost 1.63's boost python jamfile
-rule find-py3-version
+# this is just to force boost build to pick the desired python target when using LIBTORRENT_PYTHON_INTERPRETER
+feature libtorrent-python : on ;
+
+if $(LIBTORRENT_PYTHON_INTERPRETER)
 {
-	local BOOST_VERSION_TAG = [ modules.peek boostcpp : BOOST_VERSION_TAG ] ;
-	if $(BOOST_VERSION_TAG) >= 1_67
-	{
-		# starting with boost 1.67.0 boost python no longer define a separate
-		# target for python3 (boost_python3) so then we just use the regular
-		# boost_python target
-		return ;
-	}
-	local versions = [ feature.values python ] ;
-	local py3ver ;
-	for local v in $(versions)
-	{
-		if $(v) >= 3.0
-		{
-			py3ver = $(v) ;
-		}
-	}
-	return $(py3ver) ;
+	echo "using python interpreter at: " $(LIBTORRENT_PYTHON_INTERPRETER) ;
+	using python : : "$(LIBTORRENT_PYTHON_INTERPRETER)" : : : <libtorrent-python>on ;
 }
 
 if $(BOOST_ROOT)
 {
 	use-project /boost : $(BOOST_ROOT) ;
 	alias boost_python : /boost/python//boost_python : : : <include>$(BOOST_ROOT) ;
-	if [ find-py3-version ]
-	{
-		alias boost_python3 : /boost/python//boost_python3 : : : <include>$(BOOST_ROOT) ;
-	}
-	else
-	{
-		alias boost_python3 : boost_python ;
-	}
+	alias boost_python3 : boost_python ;
 }
 else
 {
@@ -175,12 +154,6 @@ rule libtorrent_linking ( properties * )
 		ECHO "WARNING: you probably want to specify libtorrent-link=static rather than link=static" ;
 	}
 
-	local BOOST_VERSION_TAG = [ modules.peek boostcpp : BOOST_VERSION_TAG ] ;
-	if <boost-link>static in $(properties) && $(BOOST_VERSION_TAG) < 1_74 && <target-os>linux in $(properties)
-	{
-		ECHO "WARNING: you cannot link statically against boost-python on linux before version 1.74.0, because it links against pthread statically in that case, which is not allowed" ;
-	}
-
 	local boost_python_lib ;
 
 	for local prop in $(properties)
@@ -198,10 +171,7 @@ rule libtorrent_linking ( properties * )
 		boost_python_lib = boost_python ;
 	}
 
-	# linux must link dynamically against boost python because it pulls
-	# in libpthread, which must be linked dynamically since we're building a .so
-	# (the static build of libpthread is not position independent)
-	if <boost-link>shared in $(properties) || ( <target-os>linux in $(properties) && $(BOOST_VERSION_TAG) < 1_74 )
+	if <boost-link>shared in $(properties)
 	{
 		result += <library>$(boost_python_lib)/<link>shared/<warnings>off ;
 	}
@@ -353,4 +323,3 @@ install stage_dependencies
 
 explicit stage_module ;
 explicit stage_dependencies ;
-
--- a/include/libtorrent/aux_/utp_stream.hpp
+++ b/include/libtorrent/aux_/utp_stream.hpp
@@ -433,9 +433,6 @@ struct TORRENT_EXTRA_EXPORT utp_stream
 		issue_write();
 	}
 
-#if BOOST_VERSION >= 106600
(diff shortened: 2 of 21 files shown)


include/libtorrent/aux_/utp_stream.hpp
```diff
@@ -434,7 +434,6 @@ struct TORRENT_EXTRA_EXPORT utp_stream
 		issue_write();
 	}
 
-#if BOOST_VERSION >= 106600
 	// Compatiblity with the async_wait method introduced in boost 1.66
```
arvidn, line 437: this comment should be removed too

bindings/python/Jamfile
```diff
@@ -33,41 +33,11 @@ if $(LIBTORRENT_PYTHON_INTERPRETER)
...
 	alias boost_python : /boost/python//boost_python : : : <include>$(BOOST_ROOT) ;
-	if [ find-py3-version ]
-	{
-		alias boost_python3 : /boost/python//boost_python3 : : : <include>$(BOOST_ROOT) ;
-	}
-	else
-	{
-		alias boost_python3 : boost_python ;
-	}
+	alias boost_python3 : boost_python ;
```
arvidn, line 40: I think the `boost_python3` target can be removed altogether with this change. I'll try out an update to this PR

aldenml:
It's my impression that for some platforms and the most common way to install boost, the version could be behind one or two. For example, today `brew info boost` reports `1.74.0`, some of the CI taks are green just because `webtorrent` is `off`. Maybe is better to wait a few months.

Btw, this should be removed from the root `Jamfile`
```
<include>./deps/json/include
<define>BOOST_JSON_HEADER_ONLY
```

arvidn:
there should probably be a check in some header, causing an `#error` in case the boost version is too old. And as Alden points out, it might be tricky to get boost-1.75 on all CI just yet.

xavier2k6:
Building docs need to be updated too to reflect this change.

include/libtorrent/aux_/utp_stream.hpp
FranciscoPombal: fixed.

FranciscoPombal:
@aldenml 

> It's my impression that for some platforms and the most common way to install boost, the version could be behind one or two. For example, today `brew info boost` reports `1.74.0`, some of the CI taks are green just because `webtorrent` is `off`. Maybe is better to wait a few months.

Hmm, I see that historically, Homebrew apparently takes a bit longer than I expected to include newer boost versions. Let's see what happens.

I've tested with CMake and `-Dwebtorrent=ON` on Ubuntu 18.04 and 20.04 thus far. I see that CI for Linux currently uses boost from the repos, but it shouldn't be too difficult to set up boost from upstream into `/opt/boost-x.xx.xx` (some sort of caching should be used in order to not incur the delay of building it every time). This is what I do locally and it "just works" because `/opt` is one of the locations that CMake searches by default when trying to find a suitable version of a library. Hopefully the same can work with BBv2 without much hassle.

> Btw, this should be removed from the root `Jamfile`
> 
> ```
> <include>./deps/json/include
> <define>BOOST_JSON_HEADER_ONLY
> ```

Thanks, fixed.

FranciscoPombal:
> Building docs need to be updated too to reflect this change.

Thanks, fixed. Updated remaining docs + bumped boost on appveyor (1.75 does not exist there yet, so it's obviously not expected to pass).

FranciscoPombal:
@arvidn 

> there should probably be a check in some header, causing an `#error` in case the boost version is too old.

This is already done in this PR in CMake. Tips on how to do the same in BBv2 are welcome.

arvidn:
>> there should probably be a check in some header, causing an #error in case the boost version is too old.
> 
> This is already done in this PR in CMake. Tips on how to do the same in BBv2 are welcome.

I was imagining a check in some header file. It would be more reliable and portable

FranciscoPombal:
@arvidn 

> I was imagining a check in some header file. It would be more reliable and portable

By having the check in the build system, you allow the build to fail the fastest in case the necessary minimum requirements are not met (see the logs of the failed CMake builds of this PR as an example). By putting the check only where it's used, you have no real guarantees of how much time will be wasted running a build that will eventually fail anyway (okay, admittedly this should always be bounded by the total time taken by the preprocessing stage, so it can't be _that_ bad).
Otherwise, you have to either include that check everywhere or go out of your way to make sure that hypothetical header file is the first to be included in the build process and depended on by everything else, at which point it's just like putting the check in the build system with extra steps. Furthermore, it can lead to undesirable consequences: off the top of my head, this would at the very least least force a full recompilation every time the minimum required dependency version is changed.

Do you expect people to build libtorrent by rolling their own build system, or "manually"? I guess someone would do that if they had a very specialized need - but in that case, I would expect them to have combed thoroughly through the documentation and probably the source code, so they would be aware of the minimum requirements anyway.

I'm curious about what you think of this. I think of the build system as a core part of the project - just like any other component, it has a set of design goals it must meet, a proper way it's meant to be used, separation of concerns relative to other components (for example, it's its job to ensure all dependencies are the appropriate versions); it can ossify, decay etc, so it must be tended to every once in a while so it doesn't get in the way in the future. It's part an parcel; as an additional example, I consider supporting more than one build system a commitment akin to supporting more than one crypto library, for example (I don't mean this is necessarily a bad thing).

From the assumption that build dependency version checking is a job for the buildsystem (as opposed to a job for the compiler's preprocessor, indirectly), saying one wants such a check to be "reliable an portable" is really saying that they want the build system to be "reliable and portable" in the first place. Of course, even if the buildsystem tool itself has such potential, it is always possible to use it in a very non-portable and unreliable way, much the same way it is possible to write very non-portable and unreliable C, for instance.

But I'm curious about other perspectives or compelling use cases that go against this "philosophy", or are otherwise not feasible to achieve simultaneously.

userdocs:
1_75 is going to report 1_74 until they fix this.

https://github.com/boostorg/boost/issues/446

Considering they use the bintray archive in the formula 

https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/boost.rb#L4

It might be confusing the issue.

FranciscoPombal:
Most things should pass now. But Appveyor is still only on Boost 1.73, and on Ubuntu 20.04 the latest available version is 1.71. Perhaps it would be better to always get the latest headers from source as it is already being done for some Windows jobs.

FranciscoPombal:
@arvidn To be honest, I don't know enough about b2 to make this work. Can you make a PR against this branch that fixes the CI?

xavier2k6:
@arvidn now that Boost 1.77 is available in AppVeyor & there's a PR open to raise it to that, does this PR still have relevance/need to be reworked?!

(Also going forward, their plan is to have two latest versions available e.g. 1.77 & 1.78 when released etc etc.) 

Will `RC_2_0` & `RC_1_2` Minimum Boost requirements also be bumped?

arvidn:
> Will RC_2_0 & RC_1_2 Minimum Boost requirements also be bumped?

I wasn't going to do that until appveyor either drops support for the older images, or updates those images to only have newer boost versions.

arvidn:
I'm still hesitant to require a newer version of boost than what's available in a distro's package manager.

PR #5745, opened 2020-12-15
CMake: Fix static_runtime compile option on MSVC

Chocobo1:
Some recent changes broke my static compilation on MSVC and this patch fixed it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 3.15.0 FATAL_ERROR) # Configurable policies: <= CMP0094
 
+cmake_policy(SET CMP0091 NEW)
 cmake_policy(SET CMP0092 NEW)
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules)
@@ -486,8 +487,12 @@ elseif(MSVC)
 endif()
 
 if(static_runtime)
-	include(ucm_flags)
-	ucm_set_runtime(STATIC)
+	if (MSVC)
+		set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+	else()
+		include(ucm_flags)
+		ucm_set_runtime(STATIC)
+	endif()
 	set(Boost_USE_MULTITHREADED ON)
 	set(Boost_USE_STATIC_RUNTIME ON)
 	set(OPENSSL_MSVC_STATIC_RT ON)


arvidn:
what error does this fix? would you mind adding a comment explaining the special case for msvc?

Chocobo1:
>what error does this fix?

It was `error LNK2038: mismatch detected for 'RuntimeLibrary'` when linking with libtorrent, I don't have the full error message now.

>would you mind adding a comment explaining the special case for msvc?

I not sure it warrants a comment. As you can see in the original code, `ucm_set_runtime` is supposed to handle it, however it doesn't work anymore.
Its upstream repo has little activity: https://github.com/onqtam/ucm and it doesn't seem to have the fix.

FranciscoPombal (approved):
@arvidn 

> what error does this fix? would you mind adding a comment explaining the special case for msvc?

`ucm_flags` is quite outdated, it does things in ways that should not be done in modern CMake (e.g. manipulating/setting `CMAKE_C*_FLAGS`, especially to make important decisions about the build process; these should considered reserved for developers to override locally at will; and in toolchain files, the `CMAKE_C*_FLAGS_INIT` variables should be used for initialization instead).

This is the correct way to select the runtime library flags now on MSVC. The rest (handling the flags for GCC and other compilers) can be improved later, and then `ucm_flags` can be removed.

FranciscoPombal:
Curious as to why and how this regression happened in the first place, though. I suspect the bump to 3.15.0 in minimum required version (695d367a588eb9352965d941edfedd83f46c1d67) might have been it, but it doesn't really make sense - https://cmake.org/cmake/help/git-stage/policy/CMP0091.html defaults to `OLD`, so the behavior should have remained the same... Well, regardless, this is the correct change.

CMakeLists.txt
```diff
@@ -486,8 +487,12 @@ elseif(MSVC)
 endif()
 
 if(static_runtime)
-	include(ucm_flags)
-	ucm_set_runtime(STATIC)
+	if (MSVC)
+		set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
```
arvidn, line 491: so, cmake just requires you to set this variable when building static libraries with msvc?
does it not affect MingW? (it's not obvious that `MSVC` is the correct test).

is `ucm_set_runtime()` still required on other compilers? (still, it seems like it's more of a target platform distinction, rather than compiler). It seems a bit strange that cmake couldn't figure out which runtime library to link against on its own.

I think these things are a bit mysterious, at least to someone who isn't a cmake expert. A comment could explain some of this, or maybe provide a link to an explanation of what this does.

What runtime library does cmake link against if this is *not* set?

Chocobo1: >so, cmake just requires you to set this variable when building static libraries with msvc?

It appears so.

>does it not affect MingW? (it's not obvious that MSVC is the correct test).

The variable `CMAKE_MSVC_RUNTIME_LIBRARY` is explicitly for MSVC as the name is in it.

>What runtime library does cmake link against if this is not set?

See the link in the commit, I'll quote it for you:
> If they are not set then CMake uses the default value MultiThreaded$<$<CONFIG:Debug>:Debug>DLL which is equivalent to the original flags.

>is ucm_set_runtime() still required on other compilers? 

Seems yes, but I don't build statically on other platforms. This PR only aim to fix compilation with MSVC.

arvidn: > The variable CMAKE_MSVC_RUNTIME_LIBRARY is explicitly for MSVC as the name is in it.

good point

PR #5748, opened 2020-12-16
CMake: check if linking with libatomic is needed on non MSVC for atom…

### LINKED ISSUE #5117: Linking failure on ARM (Raspberry Pi 4, Raspbian 32-bit)
**libtorrent version (or branch)**

`1.2.10`

**platform/architecture**

ARM (Raspberry Pi 4b)
Rasbian Linux 32-bit

**compiler and compiler version**

- GCC 8.3.0

Libraries/tools:

- OpenSSL 1.1.1d
- Boost 1.74.0
- CMake 3.18.2

**please describe what symptom you see, what you would expect to see instead and
how to reproduce it.**

libtorrent builds successfully, but linking always fails with errors relating to `libatomic`. The simplest way to reproduce is to attempt to build `examples` and or `tools` (`-Dbuild_examples=ON`, `-Dbuild_tools=ON`).

This seems to be a widespread problem with Raspbian in general (just search for `raspberrypi libatomic` on the web for other examples).

Relevant snippet of the error:

```
[160/179] Linking CXX executable tools/session_log_alerts                                                                                                                                                                                                                
FAILED: tools/session_log_alerts                                                                                                                                                                                                                                 
: && /usr/bin/c++ -O3 -DNDEBUG  tools/CMakeFiles/session_log_alerts.dir/session_log_alerts.cpp.o -o tools/session_log_alerts  -Wl,-rpath,/home/pi/libtorrent/cmake-build-dir:/opt/boost-1.74.0/lib  libtorrent-rasterbar.so.1.2.10  -lpthread  
/usr/lib/arm-linux-gnueabihf/libssl.so  /usr/lib/arm-linux-gnueabihf/libcrypto.so  /opt/boost-1.74.0/lib/libboost_system.so.1.74.0 && :                                                                                                                                  
/usr/bin/ld: libtorrent-rasterbar.so.1.2.10: undefined reference to `__atomic_compare_exchange_8'                                                                                                                                                                        
/usr/bin/ld: libtorrent-rasterbar.so.1.2.10: undefined reference to `__atomic_load_8'                                                                                                                                                                                    
/usr/bin/ld: libtorrent-rasterbar.so.1.2.10: undefined reference to `__atomic_store_8'                                                                                                                                 
/usr/bin/ld: libtorrent-rasterbar.so.1.2.10: undefined reference to `__atomic_fetch_add_8'                                                                                                                            
```

What would be the best way for libtorrent to deal with this? So far I have found the following workarounds, to at least build the examples and tools:

- Passing `-DCMAKE_EXE_LINKER_FLAGS="-latomic"` to the configure command line
- This patch:

    ```diff
    diff --git a/examples/CMakeLists.txt b/examples/CMakeLists.txt
    index d35d54152..e49b048c2 100644
    --- a/examples/CMakeLists.txt
    +++ b/examples/CMakeLists.txt
    @@ -12,6 +12,8 @@ set(single_file_examples
     foreach(example ${single_file_examples})
         add_executable(${example} "${example}.cpp")
         target_link_libraries(${example} PRIVATE torrent-rasterbar)
    +    # XXX: raspberry pi workaround
    +    target_link_libraries(${example} PRIVATE atomic)
     endforeach(example)
     
     add_executable(client_test
    @@ -20,3 +22,6 @@ add_executable(client_test
         torrent_view.cpp
         session_view.cpp)
     target_link_libraries(client_test PRIVATE torrent-rasterbar)
    +# XXX: raspberrry pi workaround
    +target_link_libraries(client_test PRIVATE atomic)
    +
    diff --git a/tools/CMakeLists.txt b/tools/CMakeLists.txt
    index 1bb1b5ea7..21344c799 100644
    --- a/tools/CMakeLists.txt
    +++ b/tools/CMakeLists.txt
    @@ -3,3 +3,7 @@ target_link_libraries(dht PRIVATE torrent-rasterbar)
     
     add_executable(session_log_alerts session_log_alerts.cpp)
     target_link_libraries(session_log_alerts PRIVATE torrent-rasterbar)
    +
    +# XXX: raspberrry pi workaround
    +target_link_libraries(dht PRIVATE atomic)
    +target_link_libraries(session_log_alerts PRIVATE atomic)
    ```

Example of another solution: https://github.com/pothosware/SoapyRTLSDR/pull/42

FranciscoPombal:
…ics support

Fixes compilation of examples/tools on systems such as the Rasberry Pi 4b.

Based on: https://github.com/llvm/llvm-project/blob/f3e0431b763979c373258f7222668e87bb5d28cb/llvm/cmake/modules/CheckAtomic.cmake

Closes #5117.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -629,6 +629,38 @@ if (APPLE)
 	target_link_libraries(torrent-rasterbar PRIVATE "-framework CoreFoundation" "-framework SystemConfiguration")
 endif()
 
+# check if we need to link with libatomic (not needed on MSVC)
+if (NOT Windows)
+	# TODO: migrate to CheckSourceCompiles in CMake >= 3.19
+	include(CheckCXXSourceCompiles)
+
+	set(ATOMICS_TEST_SOURCE [=[
+		#include <atomic>
+		#include <cstdint>
+		std::atomic<int> x{0};
+		int main() {
+			x.fetch_add(1, std::memory_order_relaxed);
+			return 0;
+		}
+	]=])
+	string(REPLACE "std::atomic<int>" "std::atomic<std::int64_t>" ATOMICS64_TEST_SOURCE "${ATOMICS_TEST_SOURCE}")
+
+	check_cxx_source_compiles("${ATOMICS_TEST_SOURCE}" HAVE_CXX_ATOMICS_WITHOUT_LIB)
+	check_cxx_source_compiles("${ATOMICS64_TEST_SOURCE}" HAVE_CXX_ATOMICS64_WITHOUT_LIB)
+	if((NOT HAVE_CXX_ATOMICS_WITHOUT_LIB) OR (NOT HAVE_CXX_ATOMICS64_WITHOUT_LIB))
+		set(CMAKE_REQUIRED_LIBRARIES "atomic")
+		check_cxx_source_compiles("${ATOMICS_TEST_SOURCE}" HAVE_CXX_ATOMICS_WITH_LIB)
+		check_cxx_source_compiles("${ATOMICS64_TEST_SOURCE}" HAVE_CXX_ATOMICS64_WITH_LIB)
+		if ((NOT HAVE_CXX_ATOMICS_WITH_LIB) OR (NOT HAVE_CXX_ATOMICS64_WITH_LIB))
+			message(FATAL_ERROR "No native support for std::atomic, or libatomic not found!")
+		else()
+			message(STATUS "Linking with libatomic for atomics support")
+			unset(CMAKE_REQUIRED_LIBRARIES)
+			target_link_libraries(torrent-rasterbar PUBLIC atomic)
+		endif()
+	endif()
+endif()
+
 feature_option(build_tests "build tests" OFF)
 feature_option(build_examples "build examples" OFF)
 feature_option(build_tools "build tools" OFF)


FranciscoPombal:
Note: the `CMAKE_REQUIRED_*` variables only affect the `check_.*` macros/functions. Still, I reset the one that is used before and after the test; a consistent policy of resetting such variables between tests leads to fewer surprises when performing multiple checks.

CMakeLists.txt
```diff
@@ -629,6 +629,38 @@ if (APPLE)
 	target_link_libraries(torrent-rasterbar PRIVATE "-framework CoreFoundation" "-framework SystemConfiguration")
 endif()
 
+# check if we need to link with libatomic (not needed on MSVC)
+if (NOT MSVC)
```
arvidn, line 633: `libatomic` is not a thing on windows. It seems like this check should really check for windows, not which compiler is being used (i.e. clang and mingw on windows probably shouldn't do this either)

CMakeLists.txt
```diff
@@ -629,6 +629,38 @@ if (APPLE)
...
+	set(ATOMICS_TEST_SOURCE [=[
+		#include <atomic>
+		#include <cstdint>
+		std::atomic<int> x{0};
+		int main() {
+			x.fetch_add(1, std::memory_order_relaxed);
+			return 0;
+		}
+	]=])
+	string(REPLACE "std::atomic<int>" "std::atomic<int64_t>" ATOMICS64_TEST_SOURCE "${ATOMICS_TEST_SOURCE}")
```
arvidn, line 646: technically, `int64_t` needs `std::` in front of it. If the header `<stdint.h>` had been included, it wouldn't be needed, but this program includes the C++ version of that header.

FranciscoPombal: fixed.

CMakeLists.txt
FranciscoPombal: fixed.

arvidn (approved):

PR #5749, opened 2020-12-16
Fix GitHub Actions workflows running on Ubuntu

FranciscoPombal:
Always update package lists before attempting to install any packages.

This prevents update failures due to attempted downloads of packages that no longer exist.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,6 +15,10 @@ jobs:
       - name: checkout
         uses: actions/checkout@v2.3.3
 
+      - name: update package lists
+        run: |
+          sudo apt update
+
       - name: install dependencies
         run: |
           sudo apt install python-docutils python-pygments python-pil gsfonts inkscape icoutils graphviz hunspell imagemagick
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -22,6 +22,10 @@ jobs:
         with:
            submodules: recursive
 
+      - name: update package lists
+        run: |
+          sudo apt update
+
       - name: install boost
         run: |
           sudo apt install libboost-tools-dev libboost-dev libboost-system-dev
@@ -57,6 +61,10 @@ jobs:
         with:
            submodules: recursive
 
+      - name: update package lists
+        run: |
+          sudo apt update
+
       - name: install boost
         run: |
           sudo apt install libboost-tools-dev libboost-dev libboost-system-dev
@@ -79,6 +87,10 @@ jobs:
         with:
            submodules: recursive
 
+      - name: update package lists
+        run: |
+          sudo apt update
+
       - name: install boost
         run: |
           sudo apt install libboost-tools-dev libboost-dev libboost-system-dev
@@ -99,6 +111,10 @@ jobs:
         uses: actions/checkout@v2.3.3
         with:
            submodules: recursive
+
+      - name: update package lists
+        run: |
+          sudo apt update
 
       - name: install clang-tidy
         run: sudo apt install clang-tidy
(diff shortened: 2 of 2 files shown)


FranciscoPombal:
I noticed the problem due to this error: https://github.com/arvidn/libtorrent/runs/1564773398?check_suite_focus=true

> Err:61 http://security.ubuntu.com/ubuntu focal-updates/universe amd64 python-lxml amd64 4.5.0-1ubuntu0.1
  404  Not Found [IP: 52.252.75.106 80]
Fetched 30.8 MB in 1s (49.1 MB/s)
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/universe/l/lxml/python-lxml_4.5.0-1ubuntu0.1_amd64.deb  404  Not Found [IP: 52.252.75.106 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Error: Process completed with exit code 100.

As you can see in http://security.ubuntu.com/ubuntu/pool/universe/l/lxml/, at the time of writing, the requested one is no longer available, only `python-lxml_4.5.0-1ubuntu0.2_amd64.deb`. We also had this exact problem in `qBittorrent` :).

FranciscoPombal:
Oops, forgot about a bunch of jobs, fixed now.

AllSeeingEyeTolledEweSew:
Per [actions/virtual-environments](https://github.com/actions/virtual-environments):

```
The Ubuntu 20.04 virtual environment is currently provided as a preview only.
The "ubuntu-latest" YAML workflow label still uses the Ubuntu 18.04 virtual environment.
```

I wonder if it's better to solve this by pinning to `ubuntu-18.04` or `ubuntu-latest`.

I made #5750 to see what happens.

FranciscoPombal:
@AllSeeingEyeTolledEweSew 

Ubuntu 20.04 is not the issue and is more than mature enough by now for these kinds of tasks.

This was recommended to me a while ago by the Actions folks a while ago when I ran into the exact same issue with qBittorrent: https://github.com/actions/virtual-environments/issues/1757

AllSeeingEyeTolledEweSew:
> EDIT: this issue is 100% expected if you don't update your package lists; it is not specific to a CI environment. You can try it yourself in a machine running Ubuntu, provided you disable auto package list updates for a while. The reason most people never actually run into this in the wild is because Ubuntu's auto updater does the `update` first, and those who update their system manually have `sudo apt update && sudo apt upgrade` ingrained in their muscle memory.

Dang. I'd expect that the pre-supplied package index would just reflect the contents of their mirrors, and the two are updated together whenever they deploy, and `apt update` would be a no-op. That seems like the best way to provide a stable CI environment. I don't expect CI env maintenance to look very much like maintenance of a normal linux install.

Running `apt update` on every workflow seems to introduce chances to barf on short-lived problems, especially inconsistencies between mirrors, since they're not fetched transactionally.

`apt update` does seem like the best practice, but it makes me sad.

FranciscoPombal:
@AllSeeingEyeTolledEweSew 

> Dang. I'd expect that the pre-supplied package index would just reflect the contents of their mirrors, and the two are updated together whenever they deploy, and `apt update` would be a no-op. That seems like the best way to provide a stable CI environment. I don't expect CI env maintenance to look very much like maintenance of a normal linux install.
> 
> Running `apt update` on every workflow seems to introduce chances to barf on short-lived problems, especially inconsistencies between mirrors, since they're not fetched transactionally.
> 
> `apt update` does seem like the best practice, but it makes me sad.

Indeed. But it appears there is nothing else we can do about it:

https://github.com/actions/virtual-environments/issues/1757#issuecomment-705608329

arvidn:
I just fixed this in a less elegant way when merging RC_1_2 -> RC_2_0. I'll have to revert that at the next merge

PR #5750, opened 2020-12-16
Pin linux workflows to ubuntu-latest

AllSeeingEyeTolledEweSew:
CIs are failing because something appears to be broken with the `ubuntu-20.04` action runner.

There's #5749, but per [actions/virtual-environments](https://github.com/actions/virtual-environments), I see that:
```
The Ubuntu 20.04 virtual environment is currently provided as a preview only.
The "ubuntu-latest" YAML workflow label still uses the Ubuntu 18.04 virtual environment.
```

If we intended to pin a version, maybe `ubuntu-20.04` is unstable now. Perhaps `ubuntu-18.04` is better, or maybe `ubuntu-latest` is a nice way to express that things should build on latest stable ubuntu.

Anyway, this PR is really to check if reverting to 18.04 fixes the current problems, so I can work on my other PR without the CI going crazy.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -9,7 +9,7 @@ jobs:
 
    docs:
       name: Docs
-      runs-on: ubuntu-20.04
+      runs-on: ubuntu-latest
 
       steps:
       - name: checkout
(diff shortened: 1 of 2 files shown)


AllSeeingEyeTolledEweSew:
Looks like a no-go!

FranciscoPombal:
The issue is not 20.04 vs. 18.04. See https://github.com/arvidn/libtorrent/pull/5749#issuecomment-747103347.

PR #5759, opened 2020-12-20
Clean up setup.py's fixups

AllSeeingEyeTolledEweSew:
(Retroactive edit)

I fell behind in the race to fix `setup.py` on windows. This cleans up the fixes a bit:
 * Added comments according to some of our discussions, to clarify the intent of various hacks
 * Grouped some logic closer together for readability
 * The platform-specific logic *should* be fairly robust (fewer `os.name` tests)

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+from distutils import log
 import distutils.debug
 import distutils.sysconfig
 import multiprocessing
@@ -237,8 +238,8 @@ class LibtorrentBuildExt(BuildExtBase):
         try:
             write_b2_python_config(config)
             config.seek(0)
-            self.announce("project-config.jam contents:")
-            self.announce(config.read())
+            log.info("project-config.jam contents:")
+            log.info(config.read())
             config.close()
             args.append(f"--project-config={config.name}")
             self.spawn(["b2"] + args)
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -1,13 +1,14 @@
 #!/usr/bin/env python3
 
 import distutils.debug
+import distutils.sysconfig
+import multiprocessing
 import os
 import pathlib
 import platform
 import sys
 import sysconfig
 import tempfile
-import multiprocessing
 
 import setuptools
 import setuptools.command.build_ext as _build_ext_lib
@@ -65,18 +66,22 @@ class StubExtension(setuptools.Extension):
         super().__init__(name, sources=[])
 
 
-def escape(string):
-    return '"' + string.replace('\\', '\\\\') + '"'
+def b2_escape(value):
+    value = value.replace("\\", "\\\\")
+    value = value.replace('"', '\\"')
+    return f'"{value}"'
 
 
-def write_b2_python_config(target, config):
+def write_b2_python_config(config):
     write = config.write
-    # b2 normally keys python environments by X.Y version, but since we may
-    # have a duplicates, we also key on a special property. It's always-on, so
-    # any python versions configured with this as a condition will always be
-    # picked.
-    # Note that python.jam actually modifies the condition for the build
-    # request, so that <define>TORRENT_FOO becomes something like
+    # b2 keys python environments by X.Y version, breaking ties by matching
+    # a property list, called the "condition" of the environment. To ensure
+    # b2 always picks the environment we define here, we define a special
+    # feature for the condition and include that in the build request.
+
+    # Note that we might try to reuse a property we know will be set, like
+    # <define>TORRENT_FOO. But python.jam actually modifies the build request
+    # in this case, so that <define>TORRENT_FOO becomes something like
     # <define>TORRENT_FOO,<python>3.7,<target-os>linux:... which causes chaos.
     # We should always define a custom feature for the condition.
     write("import feature ;\n")
@@ -89,20 +94,31 @@ def write_b2_python_config(target, config):
     paths = sysconfig.get_paths()
     includes = [paths["include"], paths["platinclude"]]
 
-    # Note that on debian, the extension suffix is overwritten, but it's
-    # necessary everywhere else, or else b2 will just build "libtorrent.so".
-    # python.jam appends SHLIB_SUFFIX on its own.
-    target = target.name.split("libtorrent")[1]
-    if os.name == "nt":
-        suffix = sysconfig.get_config_var("EXT_SUFFIX")
-    else:
-        suffix = sysconfig.get_config_var("SHLIB_SUFFIX")
-    if suffix != None and target.endswith(suffix):
-        target = target[:-len(suffix)]
+    write("using python")
+    write(f" : {sysconfig.get_python_version()}")
+    write(f" : {b2_escape(sys.executable)}")
+    write(" : ")
+    write(" ".join(b2_escape(path) for path in includes))
+    write(" :")  # libraries
+    write(" : <libtorrent-python>on")
 
-    using = f'using python : {sysconfig.get_python_version()} : {escape(sys.executable)} : {" ".join(escape(path) for path in includes)} : : <libtorrent-python>on : "{target}" ;\n'
-    print(using)
-    write(using)
+    # Note that sysconfig and distutils.sysconfig disagree here, especially on
+    # windows.
+    ext_suffix = distutils.sysconfig.get_config_var("EXT_SUFFIX")
+
+    # Note that debian currently disregards ext_suffix, but it overrides it
+    # using the same logic we use. Hopefully any system-specific patches would
+    # do the same thing.
+
+    # python.jam appends the platform-specific final suffix on its own. I can't
+    # find a consistent value from sysconfig or distutils.sysconfig for this.
+    for plat_suffix in (".pyd", ".dll", ".so", ".sl"):
+        if ext_suffix.endswith(plat_suffix):
+            ext_suffix = ext_suffix[: -len(plat_suffix)]
+            break
+    write(f" : {b2_escape(ext_suffix)}")
+    write(" ;\n")
+
 
 BuildExtBase = _build_ext_lib.build_ext
 
@@ -187,24 +203,38 @@ class LibtorrentBuildExt(BuildExtBase):
         if self.cxxstd:
             args.append(f"cxxstd={self.cxxstd}")
 
-        # Jamfile hack to copy the module to our target directory
-        target = pathlib.Path(self.get_ext_fullpath("libtorrent"))
-        args.append(f"python-install-path={target.parent}")
+        # Jamfile hacks to ensure we select the python environment defined in
+        # our project-config.jam
         args.append("libtorrent-python=on")
+        args.append(f"python={sysconfig.get_python_version()}")
+
+        # Our goal is to produce an artifact at this path. If we do this, the
+        # distutils build system will skip trying to build it.
+        target = pathlib.Path(self.get_ext_fullpath("libtorrent"))
+        self.announce(f"target: {target}")
+
+        # b2 doesn't provide a way to signal the name or paths of its outputs.
+        # We try to convince python.jam to name its output file like our target
+        # and copy it to our target directory.
+
+        # Jamfile hack to copy the module to our target directory
+        args.append(f"python-install-path={target.parent}")
         args.append("install_module")
 
         # We use a "project-config.jam" to instantiate a python environment
         # to exactly match the running one.
         config = tempfile.NamedTemporaryFile(mode="w+", delete=False)
         try:
-            write_b2_python_config(target, config)
+            write_b2_python_config(config)
             config.seek(0)
             self.announce("project-config.jam contents:")
             self.announce(config.read())
-            args.append(f"python={sysconfig.get_python_version()}")
+            config.close()
             args.append(f"--project-config={config.name}")
             self.spawn(["b2"] + args)
         finally:
+            # If we errored while writing config, windows may complain about
+            # unlinking a file "in use"
             config.close()
             os.unlink(config.name)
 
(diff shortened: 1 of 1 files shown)


arvidn:
https://github.com/arvidn/libtorrent/pull/5758/commits/4e601e8cf7f846fc4e68e2b48ebd37f0cdca2e35

bindings/python/setup.py
```diff
@@ -172,8 +186,6 @@ def build_extension_with_b2(self):
 
         if self.parallel:
             args.append(f"-j{self.parallel}")
-        else:
-            args.append(f"-j{multiprocessing.cpu_count()}")
```
arvidn, line 176: is there a good reason not to do this?

AllSeeingEyeTolledEweSew: I just think it's standard / best practice to leave this sort of thing up to invocation.

Have you ever been watching a video that you care about more than your build, but the build makes the video stutter? This happens to me a lot :)

I pushed the `-j` invocation to `appveyor.yml`, with the intent to leave it there when i un-draft the PR.

arvidn: yes, but much more often I've built stuff with a single CPU instead of 30

arvidn: I think `ninja` is a good model

AllSeeingEyeTolledEweSew: Indeed. And I know newer golang defaults to using all cores. Maybe this preference is generational. I'll revert it.

AllSeeingEyeTolledEweSew:
Marked ready and updated the description.

This isn't actually invoking `setup.py` on windows, still, as it wasn't before. I can backport the invocation from RC_2_0 to avoid shenanigans like this in the future.

bindings/python/setup.py
```diff
@@ -89,20 +94,31 @@ def write_b2_python_config(target, config):
...
+    # windows.
+    ext_suffix = distutils.sysconfig.get_config_var("EXT_SUFFIX")
+
+    # Note that debian currently disregards ext_suffix, but it overrides it
+    # using the same logic we use. Hopefully any system-specific patches would
+    # do the same thing.
+
+    # python.jam appends the platform-specific final suffix on its own. I can't
+    # find a consistent value from sysconfig or distutils.sysconfig for this.
+    for plat_suffix in (".pyd", ".dll", ".so", ".sl"):
```
arvidn, line 115: what's ".sl"?

bindings/python/setup.py
```diff
@@ -89,20 +94,31 @@ def write_b2_python_config(target, config):
...
-    target = target.name.split("libtorrent")[1]
-    if os.name == "nt":
-        suffix = sysconfig.get_config_var("EXT_SUFFIX")
-    else:
-        suffix = sysconfig.get_config_var("SHLIB_SUFFIX")
-    if suffix != None and target.endswith(suffix):
-        target = target[:-len(suffix)]
-
-    using = f'using python : {sysconfig.get_python_version()} : {escape(sys.executable)} : {" ".join(escape(path) for path in includes)} : : <libtorrent-python>on : "{target}" ;\n'
-    print(using)
```
arvidn, line 104: I like being able to see this part of the configuration in the CI log

bindings/python/setup.py
```diff
@@ -65,18 +66,22 @@ def __init__(self, name):
         super().__init__(name, sources=[])
 
 
-def escape(string):
-    return '"' + string.replace('\\', '\\\\') + '"'
+def b2_escape(value):
+    value = value.replace("\\", "\\\\")
+    value = value.replace('"', '\\"')
```
arvidn, line 71: did you try this with a path containing a double quote, or is this encoding an assumption? I'm not 100% sure this wouldn't need to be double-escaped, but probably not.

bindings/python/setup.py
```diff
@@ -187,24 +203,38 @@ def build_extension_with_b2(self):
...
+        args.append(f"python={sysconfig.get_python_version()}")
+
+        # Our goal is to produce an artifact at this path. If we do this, the
+        # distutils build system will skip trying to build it.
         target = pathlib.Path(self.get_ext_fullpath("libtorrent"))
+        self.announce(f"target: {target}")
+
+        # b2 doesn't provide a way to signal the name or paths of its outputs.
+        # We try to convince python.jam to name its output file like our target
+        # and copy it to our target directory.
```
arvidn, line 218: it seems like this comment would be more appropriate at the place where we produce the `using` statement in the temporary project config file. Since that's where we actually produce the filename for the module.

bindings/python/setup.py
```diff
@@ -187,24 +203,38 @@ def build_extension_with_b2(self):
...
         # to exactly match the running one.
         config = tempfile.NamedTemporaryFile(mode="w+", delete=False)
         try:
-            write_b2_python_config(target, config)
+            write_b2_python_config(config)
             config.seek(0)
             self.announce("project-config.jam contents:")
             self.announce(config.read())
-            args.append(f"python={sysconfig.get_python_version()}")
+            config.close()
```
arvidn, line 232: do you need to close the file here too? It definitely needs to be closed in the finally block, in case an exception is thrown

bindings/python/setup.py
AllSeeingEyeTolledEweSew: Something used on hpux apparently. This list is from here:
https://github.com/boostorg/build/blob/e0870c43214a1d5e4c476eacf9f67dae90ec0b56/src/tools/python.jam#L1094-L1110

bindings/python/setup.py
AllSeeingEyeTolledEweSew: This is still done here:

https://github.com/AllSeeingEyeTolledEweSew/libtorrent/blob/3e2c28b7bb609e8ae5c2086566c0aabb725794b8/bindings/python/setup.py#L230-L231

It's just only output when debugging. I was finding that I really needed to see all b2's debug output to understand how it was digesting the config, for example whether it was getting overridden by something in `user-config.jam`. Should I make this an unconditional print? Or bump some debug level in CI? Bumping to the max looks like this right now:

```bash
DISTUTILS_DEBUG=1 python setup.py -v
```

bindings/python/setup.py
AllSeeingEyeTolledEweSew: I didn't test it yet. This is per https://boostorg.github.io/build/manual/develop/index.html#jam.language.lexical :

>B2 treats its input files as whitespace-separated tokens, with two exceptions: double quotes (") can enclose whitespace to embed it into a token, and everything between the matching curly braces (\{}) in the definition of a rule action is treated as a single string. A backslash (\) can escape a double quote, or any single whitespace character.

I originally read it such that `"` needs to be escaped as `\"` within a `"quoted string"`, but actually I guess it's ambiguous.

Although, it doesn't even say that `\` needs to be escaped as `\\`, though it occurs in their examples. Am I using the right documentation for B2? I found this documentation to be patchy and convoluted. I don't understand why the basic syntax is covered at the end of the manual, among other things

bindings/python/setup.py
AllSeeingEyeTolledEweSew: Doesn't Windows barf reading a file if another process has it open for writing? I didn't test with this removed.

double-closing is a no-op in python, so I thought this was a good defense.

bindings/python/setup.py
AllSeeingEyeTolledEweSew: It's relevant to the next lines: this is why we invoke the `install_module` rule and set the target directory.

It _would_ be relevant to the next part too, invoking `write_b2_python_config()`, if I'd done what you'd done earlier to pass the target filename and de-mangle it to find the right suffix to configure in boost.

But `write_b2_python_config()` can't force b2 to use an arbitrary filename:
 * `Jamfile` configures it to be `libtorrent.*`
 * `python.jam` configures it to be `*.{so,pyd,dll,sl}`
 * Debian actually patches `python.jam` to always enforce it to be exactly the name expected by distutils

So the best guarantee we can offer is that `write_b2_python_config()` ensures b2 will output a name expected by distutils, on all systems.

Moreover, I found that if we pass a target filename to `write_b2_python_config()` and de-mangle it to extract a suffix, it seems like more lines of code, more brittle and less clear than just grabbing the right `EXT_SUFFIX` sysconfig value.

So ... there's no reason to pass the target filename to `write_b2_python_config()`.

I updated some comments to clarify this.

bindings/python/setup.py
AllSeeingEyeTolledEweSew: If I stick a `using python ` line in my `user-config.jam` with paths like `"/home/foo/test \"test\"/python"` and enable debug output, it looks like b2 parses the config in the expected way.

bindings/python/setup.py
arvidn: when I tested this on windows 10, I did not need to close it here, before invoking `b2`

bindings/python/setup.py
```diff
@@ -187,24 +212,39 @@ def build_extension_with_b2(self):
...
 
         # We use a "project-config.jam" to instantiate a python environment
         # to exactly match the running one.
         config = tempfile.NamedTemporaryFile(mode="w+", delete=False)
         try:
-            write_b2_python_config(target, config)
+            write_b2_python_config(config)
             config.seek(0)
             self.announce("project-config.jam contents:")
             self.announce(config.read())
```
arvidn, line 241: what does `announce()` do? I thought initially that it would print it to the console, but this is not printed. (that's why I added the print of the config)

bindings/python/setup.py
arvidn: I suppose the pythonic way of doing this would be to introduce a scope:

```
with tempfile.NamedTemporaryFile(mode="w+", delete=False) as config:
   ...
```

bindings/python/setup.py
arvidn: well, it would be useful to see this config on the CI. But I don't think the verbose `b2` debug output is very useful there.

AllSeeingEyeTolledEweSew: Sounds good. I'll add it as an unconditional log

bindings/python/setup.py
AllSeeingEyeTolledEweSew: > when I tested this on windows 10, I did not need to close it here, before invoking `b2`

Great, that makes things simpler.

> I suppose the pythonic way of doing this would be to introduce a scope:
> 
> ```
> with tempfile.NamedTemporaryFile(mode="w+", delete=False) as config:
>    ...
> ```

There's not much benefit to this since nothing is done when the scope exits (because `delete=False`; which we have to override the behavior that the file is deleted when we close it).

However if we don't need to close the file before use, then you're exactly right.

bindings/python/setup.py
AllSeeingEyeTolledEweSew: It's a debug logger; it prints either with `DISTUTILS_DEBUG=1` or `setup.py -v`, I'm not sure which.

arvidn (approved):
are these commits squashed as you'd like them to be?

AllSeeingEyeTolledEweSew:
> are these commits squashed as you'd like them to be?

Yes, I think they're alright as they are.

arvidn:
actually, this fails for me, on Win10. It looks like the root cause is "Permission denied" to the temporary file.

arvidn:
also, isn't python 64 bits these days? If so, I would expect the module to be built as a 64 bit binary (`address-model=64`)

arvidn:
it looks like you need the `delete=False`, `try`-`finally` and manual `unlink()` after all

arvidn:
this patch makes it work again:

```
diff --git a/bindings/python/setup.py b/bindings/python/setup.py
index 0564b558b..c5225c534 100644
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -234,13 +234,18 @@ class LibtorrentBuildExt(BuildExtBase):
 
         # We use a "project-config.jam" to instantiate a python environment
         # to exactly match the running one.
-        with tempfile.NamedTemporaryFile(mode="w+") as config:
+        config = tempfile.NamedTemporaryFile(mode="w+", delete=False)
+        try:
             write_b2_python_config(config)
             config.seek(0)
             log.info("project-config.jam contents:")
             log.info(config.read())
+            config.close()
             args.append(f"--project-config={config.name}")
             self.spawn(["b2"] + args)
+        finally:
+            config.close()
+            os.unlink(config.name)
 
 
 setuptools.setup(
```

AllSeeingEyeTolledEweSew:
> it looks like you need the `delete=False`, `try`-`finally` and manual `unlink()` after all

Should I rebase on RC_2_0, or invoke `setup.py` on appveyor here in RC_1_2, so this actually gets tested before committing?

I'll apply your patch. It's basically reverting 9b97591d1d7239d28756a8ce60b67d8f6da87697 though, so you could squash if you like for simplicity.

> also, isn't python 64 bits these days? If so, I would expect the module to be built as a 64 bit binary (`address-model=64`)

I believe you can still find 32-bit python in various places. `setup.py` ought to build to match the running python, choosing 64 or 32 bits as appropriate. Per https://github.com/arvidn/libtorrent/runs/1606772053?check_suite_focus=true#step:6:14 and https://github.com/arvidn/libtorrent/runs/1606772061?check_suite_focus=true#step:6:14 , it appears to be doing 64-bit builds on CI as appropriate. Are you seeing something to the contrary?

arvidn:
> Should I rebase on RC_2_0, or invoke setup.py on appveyor here in RC_1_2, so this actually gets tested before committing?

I think this patch should go into RC_1_2. I will merge it into RC_2_0 later. I suppose it would be nice to test this on appveyor too. I think we can punt on that though. I'll test it locally, and we'll see if it still works in CI when I merger it to RC_2_0.

> I'll apply your patch. It's basically reverting 9b97591 though, so you could squash if you like for simplicity.

Would you mind amending that patch instead?

>  it appears to be doing 64-bit builds on CI as appropriate. Are you seeing something to the contrary?

No, it seems correct. I didn't see it in setup.py and got worried.

arvidn:
cool, this works on Win10 for me

PR #5776, opened 2020-12-25
CMake: Set Windows link libraries public.

Biswa96:


arvidn:
it would seem a better solution would be to make `ws2_32` a `PUBLIC` dependency of libtorrent itself. That way all clients of the library would automatically get this dependency. Is there a reason not to solve it that way?

Biswa96:
I have a superstition with making variables public. I will try to compile with that.

arvidn:
How come the build on windows CI (appveyor) builds work without this change? Does this only apply to static builds of libtorrent? If so, perhaps there should be another condition. I normally invoke @zeule as the cmake expert, but maybe @FranciscoPombal would like to weigh in too.

Biswa96:
FYI, I am compiling this project in MSYS2/Mingw-w64 environment with CMake and Ninja. I don't know what the case in Visual Studio.

zeule:
Yes, @arvidn, for static libraries all dependencies are promoted to the public type by cmake.

arvidn:
@Biswa96 could you share a bit on the actual error you're encountering? I assume that you're building libtorrent as a shared library, on windows with MingW and you get linker errors for missing symbols in `ws2_32` (or `mswsock`).

Do you get these link errors in the examples or tests that come with libtorrent?

Or are you getting these errors in your client when linking against libtorrent? If so, perhaps your client has a direct dependency on the sockets library. I suppose it could have acquired this dependency via some libtorrent header file.

Biswa96:
> could you share a bit on the actual error you're encountering?

Absolutely. I should have done that at first 😭 

> Do you get these link errors in the examples or tests that come with libtorrent?

Yes, the linking errors are with examples, tools and tests. Here is the CMake command I have used:

```sh
MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
${MINGW_PREFIX}/bin/cmake.exe \
-G "Ninja" \
-DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
-DCMAKE_BUILD_TYPE=Release \
../ \
-DBUILD_SHARED_LIBS=ON \
-Dbuild_tests=ON \
-Dbuild_examples=ON \
-Dbuild_tools=ON
```

Some variables are just MSYS2 specific but you could ignore it here. And here is the some snippet of the linking error:

```c
ld.exe: examples/CMakeFiles/dump_torrent.dir/dump_torrent.cpp.obj:dump_torrent.cpp:(.text+0xda): undefined reference to `__imp_WSACleanup'
ld.exe: examples/CMakeFiles/dump_torrent.dir/dump_torrent.cpp.obj:dump_torrent.cpp:(.text+0x123): undefined reference to `__imp_WSAStartup'
collect2.exe: error: ld returned 1 exit status
```

If you need I could set up a GitHub Actions to build in MSYS2/Mingw-w64 environment.

arvidn:
> If you need I could set up a GitHub Actions to build in MSYS2/Mingw-w64 environment.

If you're up for it, I think that would be great. I abandoned the mingw build on appveyor after spending too much time trying to get it to work with c++14 (iirc, or maybe it was C++17 in master)

Biswa96:
> If you're up for it

I mean just for testing. And how about this PR?

Biswa96:
> If you're up for it

I mean just for testing this patch. Is this patch OK?

arvidn:
> Is this patch OK?

I would like to understand why this is a problem for mingw and not msvc, before judging whether this solution is appropriate. The `WSACleanup()` and `WSAStartup()` functions are called by asio. It's called from the initialization of a global variable in `asio/detail/winsock_init.hpp`. I suppose this file ends up being included by the tests and examples. I wonder why it differs though. Perhaps msvc (somehow) exports this global variable from the library, preventing clients to also initialize winsock.

It seems like a harmless change and the Jamfile declares `ws2_32`, `wsock32` and `iphlpapi` all as usage requirements for libtorrent.

I'll land this unless anyone objects.

PR #5778, opened 2020-12-25
fix cmake atomic detection with clang and freebsd

aldenml:


CMakeLists.txt
```diff
@@ -635,7 +635,7 @@ if (NOT Windows)
 	set(ATOMICS_TEST_SOURCE [=[
 		#include <atomic>
 		#include <cstdint>
-		std::atomic<int> x{0};
+		std::atomic<int> x(0);
```
aldenml, line 638: hi @FranciscoPombal, there is something slightly different with AppleClang 12.0.0.12000032, however, I can't find any reference to it.

aldenml:
Interesting it fails with `Clang 8.0.1`

aldenml:
looks like a case of a missing `-std=c++17`

aldenml:
Something is wrong with FreeBSD, maybe these are relevant:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=230888
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=243364

arvidn:
The cirrus CI broke boost-build a few days ago too. Did they update their image, or was there something else that caused this to start failing?

aldenml:
I detected the problem when this PR (https://github.com/arvidn/libtorrent/pull/5748), the cmake atomic detection change landed in `master`. The problem is only for FreeBSD, because macOS is fine. I tried to change the image and other things with no positive result.

arvidn:
@FranciscoPombal any idea?It looks like the logic to find libatomtc fails on FreeBSD.

aldenml:
@arvidn, I see that you fixed freebsd in `RC_2_0`, I will wait for the merge in `master` to rebase this PR

arvidn:
@aldenml the cirrus CI build still fails in the cmake step failing to find libatomic though, with this patch

aldenml:
oh, I see that with your change to pin to `boost-1.74` in `RC_2_0` cirrus CI is green, I assumed that was it.

arvidn:
yeah, for some reason this `libatomic` issue doesn't manifest in `RC_2_0`. Not sure why.

PR #5784, opened 2020-12-27
shortcuts to ask a torrent_info for v1 and v2 metadata

arvidn:
allow not setting a creation date in create_torrent (but default to current timestamp).
Provide more prominent shortcuts to ask a torrent_info about whether it has v1 and v2 metadata

### DIFF
--- a/src/torrent_info.cpp
+++ b/src/torrent_info.cpp
@@ -1465,7 +1465,8 @@ namespace {
 			}
 
 			auto const hashes = piece_layer->second;
-			if ((hashes.size() % sha256_hash::size()) != 0) {
+			if ((hashes.size() % sha256_hash::size()) != 0)
+			{
 				ec = errors::torrent_invalid_piece_layer;
 				return false;
 			}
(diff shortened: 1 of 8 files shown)


src/torrent_info.cpp
```diff
@@ -1481,6 +1481,8 @@ namespace {
 	{
 		TORRENT_ASSERT_PRECOND(f >= file_index_t(0));
 		if (f >= m_piece_layers.end_index()) return {};
+		if (m_files.file_size(f) <= piece_length())
+			return {m_files.root_ptr(f), lt::sha256_hash::size()};
```
arvidn, line 1485: this should have a test

arvidn: and have a special case for pad files

src/torrent_info.cpp
```diff
@@ -1481,6 +1481,13 @@ namespace {
 	{
 		TORRENT_ASSERT_PRECOND(f >= file_index_t(0));
 		if (f >= m_piece_layers.end_index()) return {};
+		if (m_files.pad_file_at(f)) return {};
+
+		if (m_files.file_size(f) <= piece_length()) {
```
aldenml, line 1486: I know it's not the only one in this file, trailing `{` in the same line :)

arvidn: oops :)

PR #5793, opened 2020-12-28
support setup.py being invoked from any CWD

arvidn:
@AllSeeingEyeTolledEweSew what do you think about this?

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -10,6 +10,7 @@ import platform
 import sys
 import sysconfig
 import tempfile
+import subprocess
 
 import setuptools
 import setuptools.command.build_ext as _build_ext_lib
@@ -228,7 +229,7 @@ class LibtorrentBuildExt(BuildExtBase):
 
         # Our goal is to produce an artifact at this path. If we do this, the
         # distutils build system will skip trying to build it.
-        target = pathlib.Path(self.get_ext_fullpath("libtorrent"))
+        target = pathlib.Path(self.get_ext_fullpath("libtorrent")).absolute()
         self.announce(f"target: {target}")
 
         # b2 doesn't provide a way to signal the name or paths of its outputs.
@@ -250,7 +251,10 @@ class LibtorrentBuildExt(BuildExtBase):
             log.info(config.read())
             config.close()
             args.append(f"--project-config={config.name}")
-            self.spawn(["b2"] + args)
+
+            python_binding_dir = pathlib.Path(__file__).parent.absolute()
+            log.info(" ".join(["b2"] + args))
+            subprocess.run(["b2"] + args, cwd=python_binding_dir, check=True)
         finally:
             # If we errored while writing config, windows may complain about
             # unlinking a file "in use"
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -163,13 +163,13 @@ class LibtorrentBuildExt(BuildExtBase):
 
     def initialize_options(self):
 
-        default_cxxstd = '11'
+        self.cxxflags = None
 
         try:
             with open('compile_flags') as f:
                 opts = f.read()
                 if '-std=c++' in opts:
-                    default_cxxstd = opts.split('-std=c++')[-1].split()[0]
+                    self.cxxflags = '-std=c++' + opts.split('-std=c++')[-1].split()[0]
         except:
             pass
 
@@ -179,7 +179,10 @@ class LibtorrentBuildExt(BuildExtBase):
         self.pic = None
         self.optimization = None
         self.hash = None
-        self.cxxstd = default_cxxstd
+        if platform.system() == "Darwin":
+            self.cxxstd = '11'
+        else:
+            self.cxxstd = None
         return super().initialize_options()
 
     def run(self):
@@ -221,6 +224,8 @@ class LibtorrentBuildExt(BuildExtBase):
             args.append("--hash")
         if self.cxxstd:
             args.append(f"cxxstd={self.cxxstd}")
+        if self.cxxflags:
+            args.append(f"cxxflags={self.cxxflags}")
 
         # Jamfile hacks to ensure we select the python environment defined in
         # our project-config.jam
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -140,7 +140,7 @@ class LibtorrentBuildExt(BuildExtBase):
         (
             "libtorrent-link=",
             None,
-            "how to link to libtorrent ('static' or 'shared')",
+            "how to link to libtorrent ('static', 'shared' or 'prebuilt')",
         ),
         (
             "boost-link=",
@@ -164,6 +164,7 @@ class LibtorrentBuildExt(BuildExtBase):
     def initialize_options(self):
 
         self.cxxflags = None
+        self.linkflags = None
 
         try:
             with open('compile_flags') as f:
@@ -173,7 +174,19 @@ class LibtorrentBuildExt(BuildExtBase):
         except:
             pass
 
-        self.libtorrent_link = None
+        try:
+            with open('link_flags') as f:
+                opts = f.read().split(' ')
+                opts = [x for x in opts if x.startswith('-L')]
+                if len(opts):
+                    self.linkflags = opts
+        except:
+            pass
+
+        if os.name == "nt":
+            self.libtorrent_link = 'static'
+        else:
+            self.libtorrent_link = 'prebuilt'
         self.boost_link = None
         self.toolset = None
         self.pic = None
@@ -225,7 +238,14 @@ class LibtorrentBuildExt(BuildExtBase):
         if self.cxxstd:
             args.append(f"cxxstd={self.cxxstd}")
         if self.cxxflags:
-            args.append(f"cxxflags={self.cxxflags}")
+            args.append(f"cxxflags=\"{self.cxxflags}\"")
+        if self.linkflags:
+            # since b2 may be running with a different directory as cwd, relative
+            # paths need to be converted to absolute
+            for lf in self.linkflags:
+                if lf[2] != '/':
+                    lf = '-L' + str(pathlib.Path(lf[2:]).absolute())
+                args.append(f"linkflags=\"{lf}\"")
 
         # Jamfile hacks to ensure we select the python environment defined in
         # our project-config.jam
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -204,12 +204,13 @@ class LibtorrentBuildExt(BuildExtBase):
         return super().run()
 
     def build_extension_with_b2(self):
+        args = []
+
         if os.name == "nt":
             self.toolset = get_msvc_toolset()
             self.libtorrent_link = "static"
             self.boost_link = "static"
-
-        args = []
+            args.append('--abbreviate-paths')
 
         if distutils.debug.DEBUG:
             args.append("--debug-configuration")
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -268,23 +268,25 @@ class LibtorrentBuildExt(BuildExtBase):
 
         # We use a "project-config.jam" to instantiate a python environment
         # to exactly match the running one.
-        config = tempfile.NamedTemporaryFile(mode="w+", delete=False)
+        python_binding_dir = pathlib.Path(__file__).parent.absolute()
+        config = open(python_binding_dir / 'project-config.jam', 'w+')
+#        config = tempfile.NamedTemporaryFile(mode="w+", delete=False)
         try:
             write_b2_python_config(config)
             config.seek(0)
             log.info("project-config.jam contents:")
             log.info(config.read())
             config.close()
-            args.append(f"--project-config={config.name}")
+#            args.append(f"--project-config={config.name}")
 
-            python_binding_dir = pathlib.Path(__file__).parent.absolute()
             log.info(" ".join(["b2"] + args))
             subprocess.run(["b2"] + args, cwd=python_binding_dir, check=True)
         finally:
             # If we errored while writing config, windows may complain about
             # unlinking a file "in use"
             config.close()
-            os.unlink(config.name)
+#            os.unlink(config.name)
+            os.unlink(python_binding_dir / 'project-config.jam')
 
 
 setuptools.setup(
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -219,7 +219,7 @@ jobs:
         run: |
           sudo apt install libboost-tools-dev libboost-python-dev libboost-dev libboost-system-dev
           sudo apt install python-docutils python-pygments python-pil gsfonts inkscape icoutils graphviz hunspell imagemagick python3-setuptools
-          python -m pip install aafigure
+          python3 -m pip install aafigure
           echo "using gcc ;" >>~/user-config.jam
 
       - name: build tarball
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -198,7 +198,11 @@ jobs:
 
    dist:
       name: build dist
-      runs-on: ubuntu-20.04
+      runs-on: ${{ matrix.os }}
+
+      strategy:
+         matrix:
+            os: [ubuntu-20.04, ubuntu-18.04]
 
       steps:
       - name: checkout
@@ -230,8 +234,19 @@ jobs:
           name: tarball
           path: libtorrent-rasterbar-*.tar.gz
 
-      - name: test-tarball
+      - name: test-tarball (make)
         run: |
           tar xvzf libtorrent-rasterbar-*.tar.gz
-          cd libtorrent-rasterbar-*/test
-          b2 link=static -j2 cxxstd=11
+          cd libtorrent-rasterbar-*/
+          ./configure --enable-python-binding --enable-examples=yes --enable-encryption --enable-tests=yes
+          make -j2
+
+      - name: test-tarball (bjam)
+        uses: nick-invision/retry@v2
+        with:
+          timeout_minutes: 30
+          retry_wait_seconds: 1
+          max_attempts: 3
+          command: |
+            cd libtorrent-rasterbar-*/test
+            b2 link=static -j2 cxxflags=-std=c++11
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -211,10 +211,10 @@ jobs:
         run: |
           sudo apt update
 
-      - name: install boost
+      - name: install dependencies
         run: |
           sudo apt install libboost-tools-dev libboost-python-dev libboost-dev libboost-system-dev
-          sudo apt install python-docutils python-pygments python-pil gsfonts inkscape icoutils graphviz hunspell imagemagick
+          sudo apt install python-docutils python-pygments python-pil gsfonts inkscape icoutils graphviz hunspell imagemagick python3-setuptools
           python -m pip install aafigure
           echo "using gcc ;" >>~/user-config.jam
 
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-20.04, macos-latest, windows-2019 ]
+        os: [ubuntu-20.04, ubuntu-18.04, macos-latest, windows-2019 ]
 
     steps:
     - uses: actions/checkout@v2
--- a/bindings/python/Jamfile
+++ b/bindings/python/Jamfile
@@ -80,8 +80,37 @@ else
 
 	rule boost_python_version ( name : type ? : properties * )
 	{
+		# examples of names for the boost_python library:
+		# ubuntu bionic (1.65.1): libboost_python3-py36.so.1.65.1
+		# ubuntu bionic (1.65):   libboost_python-py36.so
+		#                         libboost_python3-py36.so
+		#                         libboost_python3.so
+		# ubuntu bionic (1.62.0): libboost_python-py36.so.1.62.0
+		# ubuntu focal (1.67.0):  libboost_python38.so.1.67.0
+		# ubuntu focal (1.71.0):  libboost_python38.so
+		# ubuntu groovy (1.71.0): libboost_python38.so
+		# ubuntu hirsute(1.71.0): libboost_python39.so
+		# debian buster (1.67.0): libboost_python37.so
+		#                         libboost_python3.so
+		#                         libboost_python3-py37.so
+		# debian sid (1.74.0):    libboost_python39.so
+		# debian sid (1.71.0):    libboost_python39.so
+		# debian bullseye (1.71): libboost_python39.so
+		# devian buster-backports (1.71): libboost_python37.so
+		# debian buster (1.67):   libboost_python37.so.1.67.0
+		# debian stretch (1.62.0): libboost_python-py35.so.1.62.0
+		# debian stretch (1.62):  libboost_python-py35.so
+		# debian jessie (1.55.0): libboost_python-py34.so.1.55.0
+		# boost Jamfile:          libboost_python38.so.1.73.0
+
 		local py-version = [ $(properties).get <python> ] ;
-		local boost-python-lib = "boost_python" [ SPLIT_BY_CHARACTERS $(py-version) : "." ] ;
+
+		local infix = "" ;
+		if $(py-version) < 3.7 && [ $(properties).get <target-os> ] = linux
+		{
+			infix = "-py" ;
+		}
+		local boost-python-lib = "boost_python" $(infix) [ SPLIT_BY_CHARACTERS $(py-version) : "." ] ;
 
 		if $(type) in SEARCHED_LIB
 		{
--- a/bindings/python/Jamfile
+++ b/bindings/python/Jamfile
@@ -14,7 +14,7 @@ BOOST_ROOT = [ modules.peek : BOOST_ROOT ] ;
 feature lt-visibility : default hidden : composite propagated ;
 feature.compose <lt-visibility>hidden : <cflags>-fvisibility=hidden <cxxflags>-fvisibility-inlines-hidden ;
 
-feature libtorrent-link : shared static : composite propagated ;
+feature libtorrent-link : shared static prebuilt : composite propagated ;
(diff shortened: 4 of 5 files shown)


sledgehammer999:
I think this fixed the previous problem about `Jamfile` and now it fails for yet another reason. `b2` doesn't recognize one cmd option.

```
b2 release address-model=64 -j4 cxxstd=14 libtorrent-python=on python=3.6 python-install-path=/build/libtorrent-rasterbar-1.2.12/build/bindings/python3.6/build/lib.linux-x86_64-3.6 install_module --project-config=/tmp/tmp1wlxz97p
CXXFLAGS = -g -O2 -fdebug-prefix-map=/build/libtorrent-rasterbar-1.2.12=. -fstack-protector-strong -Wformat -Werror=format-security -std=c++14 -ftemplate-depth=512 -Wno-format-zero-length -fvisibility=hidden -fvisibility-inlines-hidden
LDFLAGS = -L/usr/lib/x86_64-linux-gnu -Wl,-Bsymbolic-functions -Wl,-z,relro -Wl,--as-needed -fvisibility=hidden -fvisibility-inlines-hidden
OS = LINUX
warning: No toolsets are configured.
warning: Configuring default toolset "gcc".
warning: If the default is wrong, your build may not work correctly.
warning: Use the "toolset=xxxxx" option to override our guess.
warning: For more configuration options, please consult
warning: http://boost.org/boost-build2/doc/html/bbv2/advanced/configuration.html
/usr/share/boost-build/src/build/property.jam:293: in validate1 from module property
error: Invalid property '<cxxstd>14': unknown feature 'cxxstd'.
/usr/share/boost-build/src/build/property.jam:302: in property.validate from module property
/usr/share/boost-build/src/build/build-request.jam:228: in convert-command-line-element from module build-request
/usr/share/boost-build/src/build/build-request.jam:179: in build-request.convert-command-line-elements from module build-request
/usr/share/boost-build/src/build-system.jam:691: in load from module build-system
/usr/share/boost-build/src/kernel/modules.jam:295: in import from module modules
/usr/share/boost-build/src/kernel/bootstrap.jam:139: in boost-build from module
/usr/share/boost-build/boost-build.jam:8: in module scope from module

  File "../../../bindings/python/setup.py", line 278, in <module>
    distclass=B2Distribution,
  File "/usr/lib/python3/dist-packages/setuptools/__init__.py", line 129, in setup
    return distutils.core.setup(**attrs)
  File "/usr/lib/python3.6/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/usr/lib/python3.6/distutils/dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python3.6/distutils/dist.py", line 974, in run_command
    cmd_obj.run()
  File "/usr/lib/python3.6/distutils/command/build.py", line 135, in run
    self.run_command(cmd_name)
  File "/usr/lib/python3.6/distutils/cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File "/usr/lib/python3.6/distutils/dist.py", line 974, in run_command
    cmd_obj.run()
  File "../../../bindings/python/setup.py", line 187, in run
    self.build_extension_with_b2()
  File "../../../bindings/python/setup.py", line 257, in build_extension_with_b2
    subprocess.run(["b2"] + args, cwd=python_binding_dir, check=True)
  File "/usr/lib/python3.6/subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['b2', 'release', 'address-model=64', '-j4', 'cxxstd=14', 'libtorrent-python=on', 'python=3.6', 'python-install-path=/build/libtorrent-rasterbar-1.2.12/build/bindings/python3.6/build/lib.linux-x86_64-3.6', 'install_module', '--project-config=/tmp/tmp1wlxz97p']' returned non-zero exit status 1.
Makefile:536: recipe for target 'all-local' failed
make[2]: *** [all-local] Error 1
```

[Attached build log](https://github.com/arvidn/libtorrent/files/5749138/buildlog.txt)

arvidn:
@sledgehammer999 I've updated the patch

sledgehammer999:
And it keeps exposing incompatibilities:

```
b2 release address-model=64 -j4 cxxflags=-std=c++14 libtorrent-python=on python=3.6 python-install-path=/build/libtorrent-rasterbar-1.2.12/build/bindings/python3.6/build/lib.linux-x86_64-3.6 install_module --project-config=/tmp/tmp2es2k0uy
CXXFLAGS = -g -O2 -fdebug-prefix-map=/build/libtorrent-rasterbar-1.2.12=. -fstack-protector-strong -Wformat -Werror=format-security -std=c++14 -ftemplate-depth=512 -Wno-format-zero-length -fvisibility=hidden -fvisibility-inlines-hidden
LDFLAGS = -L/usr/lib/x86_64-linux-gnu -Wl,-Bsymbolic-functions -Wl,-z,relro -Wl,--as-needed -fvisibility=hidden -fvisibility-inlines-hidden
OS = LINUX
warning: No toolsets are configured.
warning: Configuring default toolset "gcc".
warning: If the default is wrong, your build may not work correctly.
warning: Use the "toolset=xxxxx" option to override our guess.
warning: For more configuration options, please consult
warning: http://boost.org/boost-build2/doc/html/bbv2/advanced/configuration.html
/usr/share/boost-build/src/build/feature.jam:494: in feature.validate-value-string from module feature
error: "3.6" is not a known value of feature <python>
error: legal values:
/usr/share/boost-build/src/build/property.jam:276: in validate1 from module property
/usr/share/boost-build/src/build/property.jam:302: in property.validate from module property
/usr/share/boost-build/src/build/build-request.jam:228: in convert-command-line-element from module build-request
/usr/share/boost-build/src/build/build-request.jam:179: in build-request.convert-command-line-elements from module build-request
/usr/share/boost-build/src/build-system.jam:691: in load from module build-system
/usr/share/boost-build/src/kernel/modules.jam:295: in import from module modules
/usr/share/boost-build/src/kernel/bootstrap.jam:139: in boost-build from module
/usr/share/boost-build/boost-build.jam:8: in module scope from module
```

[Attached build log](https://github.com/arvidn/libtorrent/files/5749251/buildlog.txt)

arvidn:
it's as if the command line argument `--project-config=/tmp/tmp2es2k0uy` is ignored. Or that the configuration generated into that file isn't taking effect. That's where python 3.6 is defined.

arvidn:
what version of boost and boost-build is default on debian?

sledgehammer999:
>what version of boost and boost-build is default on debian?

The particular build logs are for ubuntu 18.04 and it seems it has boost 1.65

AllSeeingEyeTolledEweSew:
> @AllSeeingEyeTolledEweSew what do you think about this?

Seems good.

Should we just move `bindings/python/setup.py` to the root then? I originally wanted to do this but punted. As you can see, a lot of build processes just invoke `setup.py` at the root.

Currently the root `setup.py` has a relative `chdir()` so maybe we need to modify it anyway for the goal of the PR.

I also like removing calls to `exec()` in python. Linters will complain about it in any case.

> it's as if the command line argument `--project-config=/tmp/tmp2es2k0uy` is ignored. Or that the configuration generated into that file isn't taking effect. That's where python 3.6 is defined.

Yeah, I found that b2 is finicky about which python config it picks, and it only gives you clues in debug prints. (this is why I questioned whether we need to always print the contents of `project-config.jam`, as I think it's hard to tell what b2's going to do based on config alone)

We can either always enable the `--debug-foo` args to `b2`, or `DISTUTILS_DEBUG=1` and `setup.py -v` if it's simpler to tweak it at the build system level.

bindings/python/setup.py
```diff
@@ -250,7 +256,10 @@ def build_extension_with_b2(self):
             log.info(config.read())
             config.close()
             args.append(f"--project-config={config.name}")
-            self.spawn(["b2"] + args)
+
+            python_binding_dir = pathlib.Path(__file__).parent.absolute()
```
AllSeeingEyeTolledEweSew, line 260: I think you want `Path.resolve()`, not `Path.absolute()`. Not actually sure what `.absolute()` seems undocumented.

bindings/python/setup.py
```diff
@@ -250,7 +256,10 @@ def build_extension_with_b2(self):
...
+            log.info(" ".join(["b2"] + args))
+            subprocess.run(["b2"] + args, cwd=python_binding_dir, check=True)
```
AllSeeingEyeTolledEweSew, line 262: Note that there's a lot of logic in `spawn()`: https://github.com/pypa/distutils/blob/35e1c9d4d788094c254485dfdefa0c24cd5820ee/distutils/spawn.py#L23

I'm not sure what it all does.

To keep it intact, I think it's also reasonable to do

```python
cwd = os.getcwd()
try:
    os.chdir(python_binding_dir)
    self.spawn(["b2"] + args)
finally:
    os.chdir(cwd)
```

I don't see a builtin `pushd`-like contextmanager.

arvidn: I don't think anything `spawn()` does is important. Most of it is logging and checking environment variables for MacOS.
I looked around for `pushd` or directory stack solutions for python, but I think it's a lot more reliable to leave cwd untouched, as my current patch does. It's a global variable that anything can depend on, so I think it's best to leave it alone.

arvidn:
the "shortcut" `setup.py` in the root doesn't look correct either. The call to `chdir()` assumes the current working directory is already the same as the python script itself. I think it could either be turned into a symlink or removed.

arvidn:
> The particular build logs are for ubuntu 18.04 and it seems it has boost 1.65

In my testing with boost-1.65, it seems like the `--project-config` option hadn't been added back then. I'm thinking I'll add a work-around for `RC_1_2` but leave the current implementation of `setup.py` as is for `RC_2_0`, making it require a newer version of boost.

Specifically, I'll name the file `project-config.jam` rather than creating a temporary file.

bindings/python/setup.py
arvidn: actually, `resolve()` is problematic on windows. It doesn't only resolve symlinks, but also mapped drive letters, to their canonical UNC path. When running this in `cmd.exe` that fails because it doesn't support UNC paths.

sledgehammer999:
@arvidn tag me when you're done and you want me to test it. I tested it with previous commits you had here (which were 6 in number) and it got another error:

```
gcc.link.dll bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/libtorrent.cpython-36m-x86_64-linux-gnu.so
/usr/bin/x86_64-linux-gnu-ld: cannot find -lboost_python36
/usr/bin/x86_64-linux-gnu-ld: cannot find -ltorrent-rasterbar

    "g++" -L"../../src/.libs"   -o "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/libtorrent.cpython-36m-x86_64-linux-gnu.so" -Wl,-h -Wl,libtorrent.cpython-36m-x86_64-linux-gnu.so -shared -Wl,--start-group "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/module.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/sha1_hash.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/converters.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/create_torrent.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/fingerprint.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/utility.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/session.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/entry.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/torrent_info.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/string.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/torrent_handle.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/torrent_status.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/session_settings.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/version.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/alert.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/datetime.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/peer_info.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/ip_filter.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/magnet_uri.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/error_code.o"  -Wl,-Bstatic  -Wl,-Bdynamic -lboost_python36 -ltorrent-rasterbar -ldl -lpthread -lutil -Wl,--end-group  -Wl,-Bsymbolic

```

Whole build log: [buildlog.txt](https://github.com/arvidn/libtorrent/files/5750920/buildlog.txt)


List of file in Ubuntu 18.04 for packages:
libboost-python: https://packages.ubuntu.com/bionic/amd64/libboost-python1.65.1/filelist
libboost-python-dev: https://packages.ubuntu.com/bionic/amd64/libboost-python1.65-dev/filelist

arvidn:
@sledgehammer999 could you give it another try?

sledgehammer999:
Now it can find libboost_python but can't find it's own library:
```
gcc.link.dll bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/libtorrent.cpython-36m-x86_64-linux-gnu.so
/usr/bin/x86_64-linux-gnu-ld: cannot find -ltorrent-rasterbar

    "g++" -L"../../src/.libs"   -o "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/libtorrent.cpython-36m-x86_64-linux-gnu.so" -Wl,-h -Wl,libtorrent.cpython-36m-x86_64-linux-gnu.so -shared -Wl,--start-group "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/module.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/sha1_hash.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/converters.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/create_torrent.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/fingerprint.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/utility.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/session.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/entry.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/torrent_info.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/string.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/torrent_handle.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/torrent_status.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/session_settings.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/version.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/alert.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/datetime.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/peer_info.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/ip_filter.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/magnet_uri.o" "bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/src/error_code.o"  -Wl,-Bstatic  -Wl,-Bdynamic -lboost_python-py36 -ltorrent-rasterbar -ldl -lpthread -lutil -Wl,--end-group  -Wl,-Bsymbolic

...failed gcc.link.dll bin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden/libtorrent.cpython-36m-x86_64-linux-gnu.so...
...skipped <p/build/libtorrent-rasterbar-1.2.12/build/bindings/python3.6/build/lib.linux-x86_64-3.6>libtorrent.cpython-36m-x86_64-linux-gnu.so for lack of <pbin/gcc-7/release/address-model-64/libtorrent-link-prebuilt/lt-visibility-hidden>libtorrent.cpython-36m-x86_64-linux-gnu.so...
...failed updating 1 target...
...skipped 1 target...
...updated 29 targets...
```

Log: [buildlog.txt](https://github.com/arvidn/libtorrent/files/5752549/buildlog.txt)

arvidn:
it's a bit mysterious. You can clearly see that `libtorrent-rasterbar.so` is created in the `src/.libs` directory earlier in the log. And that the library search path `-L../../src.libs` is passed to the link command line.

It works locally for me on ubuntu-20.4.

arvidn:
I'm testing this on ubuntu-18.04 in CI now, and it passes. Is perhaps the debian build script installing and cleaning the main library before invoking `setup.py`. I still don't see how that could fail, because then it would presumably just pick it up from the system.

sledgehammer999:
@arvidn I won't be able to retest for a few hours. I use a local chroot environment based on pbuilder/cowbuilder which kinda mimics the PPA environment (aka totally clean builds).

The basic script is [here](https://salsa.debian.org/debian/libtorrent-rasterbar/-/blob/master/debian/rules). Read it as a makefile. I think the relevant bit is this:
```
override_dh_auto_configure-nopy:
	dh_auto_configure --builddirectory build -- $(CONFIGURE_ARGS)

override_dh_auto_configure-%:
	CPPFLAGS="$(CPPFLAGS) $(shell python$*-config --includes)" \
	PYTHON_LDFLAGS="$(shell python$*-config --libs)" \
	dh_auto_configure --builddirectory build-py$* -- \
		PYTHON=/usr/bin/python$* \
		PYTHON_INSTALL_PARAMS="--root=$(CURDIR)/debian/tmp \
		--install-layout=deb" \
		--enable-python-binding \
		--with-boost-python=boost_python$(subst .,,$(subst -dbg,,$*)) \
		$(CONFIGURE_ARGS)
	mv build-py$*/bindings/python build/bindings/python$*
	cp -r bindings/python/* build/bindings/python$*
	sed s/-std=c++11//g < build/bindings/python$*/compile_cmd > build/bindings/python$*/compile_cmd.new && \
		mv -f build/bindings/python$*/compile_cmd.new build/bindings/python$*/compile_cmd

override_dh_auto_configure: override_dh_auto_configure-nopy $(ALLPY:%=override_dh_auto_configure-%)
```

It builds it twice. Once without bindings and once with the bindings. (in older version once per python version too). As you can see the build directory is different for each build. Maybe that's the reason the bindings build can't find the *.so. It is produced in another build directory.
**NOTE**: This is pure speculation. In a few hours I will try again and inspect the chroot filesystem to see where things are produced and if they exist.

arvidn:
ah, yes, the `--builddirectory` definitely seems like the cause.

arvidn:
@sledgehammer999 ok, I think I've fixed it now!

sledgehammer999:
It still fails to find the libtorrent .so file.
I inspected the chroot environment and the .so is in a different build directory from the bindings.
(root libtorrent folder)->build->src->.libs->*.so file
(root libtorrent folder)->build-py3.6->src->(empty, only a Makefile exists)

Should the build script be changed to use the same build directory or is this a bug of libtorrent's build system changes?

sledgehammer999:
Well, I re-read the build script, and it only configures the bindings in a different build directory. Then it copies it back into the main build directory and issues the build command based on that. So the libtorrent script is looking at the wrong place?

```
override_dh_auto_build-%:
	dh_auto_build --builddirectory build/bindings/python$*

```

arvidn:
My most recent change was to look at the `link_flags` file, which the autotools script puts in the bindings/python directory, specifically to propagate things like this, the library directory. So I would expect it to behave (essentially) the same as it used to now. Whatever directory the makefile places the library in ought to be propagated via that file.

Where can I find the build log?

arvidn:
the `-L` command line arguments should be on the `b2` command line in the log

arvidn:
I think there are enough fixes in here to land these changes. I'll keep trying to address the build issue.

sledgehammer999:
Sorry for the delay in answering. If you want, give me 1-2 hours and I'll test these last patches and provide a build log.

sledgehammer999:
Here is the [build log](https://github.com/arvidn/libtorrent/files/5759503/buildlog.txt).

arvidn:
It turns out I don't actually understand how this used to work before the `setup.py` patch either.

The way build- and link flags were propagated from the autoconf files (at the time `./configure` runs) was by creating two files, `bindings/python/compile_flags` and `bindings/python/link_flags`. The `setup.py` script would look for these files and add those flags to the distutils compiler invocation. (hoping/assuming distutils would use the same compiler as automake).

In your most recent build log, those files are created still:

```
Generating Makefiles:
checking that generated files are newer than configure... done
configure: creating ./config.status
config.status: creating Makefile
config.status: creating src/Makefile
config.status: creating include/libtorrent/Makefile
config.status: creating examples/Makefile
config.status: creating test/Makefile
config.status: creating tools/Makefile
config.status: creating bindings/Makefile
config.status: creating bindings/python/Makefile
config.status: creating bindings/python/link_flags
config.status: creating bindings/python/compile_flags
config.status: creating bindings/python/compile_cmd
config.status: creating libtorrent-rasterbar.pc
```

It's a bit unclear what the current working directory is at this point, since the `configure` script is invoked like this:

```
cd build-py3.6 && ../configure --build=x86_64-linux-gnu --prefix=/usr --includedir=\${prefix}/include --mandir=\${prefix}/share/man --infodir=\${prefix}/share/info --sysconfdir=/etc --localstatedir=/var --disable-silent-rules --libdir=\${prefix}/lib/x86_64-linux-gnu --libexecdir=\${prefix}/lib/x86_64-linux-gnu --disable-maintainer-mode --disable-dependency-tracking PYTHON=/usr/bin/python3.6 "PYTHON_INSTALL_PARAMS=--root=/build/libtorrent-rasterbar-1.2.12/debian/tmp 	--install-layout=deb" --enable-python-binding --with-boost-python=boost_python36 --with-libiconv --with-boost-libdir=/usr/lib/x86_64-linux-gnu
```

i.e. with "builddir" as the current directory.

I made a change recently (in this patch-set) to always look for `compile_flags` and `link_flags` *next* to `setup.py`, but perhaps that's wrong, if those files are created outside the source tree. (I've reverted that change now).

I can't make `./configure` work in a separate build directory, at least not trivially. It fails to find `src/Makefile.in`.

However, I also discovered that the `link_flags.in` file (the one used to generate the `link_flags` file in the configure step) looks like this:

```
-L@top_builddir@/src/.libs @LDFLAGS@ @LIBS@ @BOOST_LDFLAGS@ @BOOST_SYSTEM_LIB@ @BOOST_PYTHON_LIB@ @PTHREAD_LIBS@ @OPENSSL_LIBS@ @OPENSSL_LDFLAGS@
```

So to some extent, it also seem to hard-code the `src/.libs` location for the final library. Perhaps `@top_builddir@` points to the *build* directory though, not the source tree.

I'm sure how to proceed. I have this one patch that might be worth testing the build against. With the debug logging it might provide some insight too.

sledgehammer999:
I can try your latest patch in a few hours.
Also, for my purposes, I could just revise the official debian script to configure the main lib + python bindings in one place. I assume that wouldn't break the final build of the lib+bindings. Then, debian can take care of it by itself when they get around to it. (they are still stuck to 1.2.9)

sledgehammer999:
Here is a new [build log((https://github.com/arvidn/libtorrent/files/5761114/buildlog.txt).

arvidn:
I think I understand what's going on now. The library paths added via `link_flags` are relative. They're relative to current working directory of the bindings/python directory in the *build* directory, not the bindings/python in the *source* directory. So, when the python script (`setup.py`) invokes `b2` with the *source* directory as current working directory, those relative paths become invalid.

I can think of two options to solve this:

1. rewrite all the relative library paths found in `link_flags` to be relative the *source* directory instead (based on current working directory)
2. copy the `Jamfile` from the source directory into the build directory and invoke `b2` without changing current working directory.

(2) seems simplest

arvidn:
although, I managed to `./configure` in a separate build directory and build everything in there, and the python bindings worked as well. and the main library only showed up under the build directory.. So I don't fully understand why that worked.

arvidn:
another thing I noticed is that bjam will use its normal build directory (`./bin`) which means the bindings do not use the intended build directory for the debian build. It may not matter anyway, since `b2` separates different build configurations anyway, so there won't be any conflicts or cross-overs.

But it should be easy to fix with `--builddir=` as well

arvidn:
at a second though, (1) seemed simpler

arvidn:
@sledgehammer999 please try again (or give me some pointers on how I can build a package)

sledgehammer999:
Success! Now the packages are being built without any error!

arvidn:
great! I'll rebase and land this

PR #5800, opened 2020-12-31
moved vector_utils.hpp to aux

aldenml:


arvidn:
I'm a bit surprised that no place where these functions are used had to update. Perhaps all that code is already in the `aux` namespace, but I would have at least expected a test to have to update.

arvidn:
I see, there are no unit tests for these functions :/

aldenml:
I will try to add tests later

PR #5805, opened 2021-01-02
Fix redundant and failed bytes overflow check

xNihil0:
The current overflow check caps `m_total_redundant_bytes` to 2 GiB, while this value can exceed 2 GiB while `strict_end_game_mode` is set to `false` on large torrents, resulting in an incorrect `downloaded` value. 
Since this value is very unlikely to reach int64 max, I don't think there's a need to perform an overflow check in `inc_stats_counter()`.

### DIFF
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -1466,7 +1466,7 @@ namespace libtorrent {
 
 		// the number of bytes that has been
 		// downloaded that failed the hash-test
-		std::int32_t m_total_failed_bytes = 0;
+		std::int64_t m_total_failed_bytes = 0;
 		std::int64_t m_total_redundant_bytes = 0;
 
 		// the sequence number for this torrent, this is a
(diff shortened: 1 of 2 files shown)


arvidn:
you said "redundant" bytes, but in the patch you're changing `m_total_failed_bytes`. I assume you meant to say "failed bytes".

How does clamping failed bytes to 2GiB cause the "downloaded" counter, reported to the tracker, become incorrect? Is it that the internal "downloaded" counter keeps incrementing for failed bytes, and what's reported to trackers is "downloaded" - "failed" ?

xNihil0:
https://github.com/arvidn/libtorrent/blob/0ad0f29ab9ae5fae83a9cbb74f5a518521df5a40/src/torrent.cpp#L2901
https://github.com/arvidn/libtorrent/blob/0ad0f29ab9ae5fae83a9cbb74f5a518521df5a40/src/torrent.cpp#L2915

I only fixed the part for redundant bytes when I opened this PR, but since `downloaded = m_stat.total_payload_download() - m_total_failed_bytes - m_total_redundant_bytes` while `report_true_downloaded` is set to `true`, I thought I might as well fix them both.

arvidn:
this patch looks right. It seems quite unlikely to exceed 2 GiB of hash failures (given peers are banned when identified as sending corrupt data). Reporting the correct "failed" number doesn't seem as important as reporting the correct "downloaded" number.

PR #5809, opened 2021-01-02
moved peer_connection.hpp to aux

aldenml:


include/libtorrent/aux_/peer_connection.hpp
```diff
@@ -246,13 +246,6 @@ namespace aux {
 		bool m_ignore_stats:1;
 	};
 
-	enum class connection_type : std::uint8_t
```
aldenml, line 249: this was moved to `peer_connection_interface.hpp`

arvidn (approved):

PR #5822, opened 2021-01-05
moved tracker_manager.hpp to aux

aldenml:


include/libtorrent/alert_types.hpp
```diff
@@ -694,7 +694,7 @@ TORRENT_VERSION_NAMESPACE_3
 		// 2. Started
 		// 3. Stopped
 		// 4. Paused
-		event_t const event;
+		aux::event_t const event;
```
aldenml, line 697: hi @arvidn, this one here, I don't know how do you want to proceed. new header for that enum only? remove this? `event_t` is "documented" as internal

arvidn: is `event_t` a public type?

arvidn: oh, I see. It's public right here.. I think this type should be public, and properly documented. I think a new header for just this enum, or some other appropriate header would be fine. There's some precedence for putting types like these in its own header, such as `socket_type_t` and `operations_t`

aldenml:
created new `tracker_event.hpp`

arvidn (approved):

arvidn:
I suppose there's a `tcp` in `aux` as well, or something: https://github.com/arvidn/libtorrent/pull/5822/checks?check_run_id=1653354992#step:4:885

aldenml:
fixed, I didn't squash to easy the review but feel free to squash while merging the PR if you want to save some CI minutes, if not, I will do it when everything is green

arvidn:
I think the fuzzer needs updating too: https://github.com/arvidn/libtorrent/runs/1647248877?check_suite_focus=true#step:5:224
and this test: https://github.com/arvidn/libtorrent/runs/1647248835?check_suite_focus=true#step:7:1627

arvidn (approved):

arvidn:
The reason I'm hesitant to use github's squash-feature is that then it will also create a merge commit. I really prefer to have linear history and github doesn't have the option to rebase-squash-and-merge.

aldenml:
I'm pretty sure the github web ui `squash-and-merge` do exactly that, or at least it is what I see when I use it, I'm putting an image to make sure we are talking about the same
Screen Shot 2021-01-06 at 8 22 57 AM

arvidn:
yeah, I read up on it before I pressed the button. I've always thought that option did not rebase, since it doesn't mention it. Even that description doesn't say it will rebase. However, I changed the settings to require fast-forward merges, and that option was still available. That gave it away

PR #5838, opened 2021-01-08
moved platform_util.hpp, puff.hpp and union_endpoint.hpp to aux

aldenml:


include/libtorrent/aux_/platform_util.hpp
```diff
@@ -1,11 +1,13 @@
 #ifndef TORRENT_PLATFORM_UTIL_HPP
 #define TORRENT_PLATFORM_UTIL_HPP
 
+#include "libtorrent/config.hpp"
```
arvidn, line 4: if all you need is the `TORRENT_EXTRA_EXPORT` it's sufficient to include `aux_/export.hpp`

aldenml: right, now the long wait of appveyor :)

include/libtorrent/aux_/platform_util.hpp
```diff
@@ -1,11 +1,13 @@
...
 
+#include "libtorrent/aux_/export.hpp"
+
 #include <cstdint>
 
-namespace libtorrent {
+namespace libtorrent::aux {
 
-	int max_open_files();
+	TORRENT_EXTRA_EXPORT int max_open_files();
```
arvidn, line 10: what prompted this? technically it's only necessary once there's a test that needs access to the function.
I think it's a good change, I'm just curious if you experienced an error without it.

arvidn (approved):

include/libtorrent/aux_/platform_util.hpp
aldenml: no error, was just mechanical thinking on my part

PR #5853, opened 2021-01-12
moved socket_io.hpp to aux

aldenml:


aldenml:
a long one :)

arvidn:
do you understand why the file shows up as deleted and added, rather than renamed? I'm a bit worried about potential changes to the file in RC_2_0 not merging cleanly into master.

arvidn (approved):

aldenml:
Yes, it’s because I indented to the left most of the file, git rename threshold seems higher. If that’s a problem, let me know and I add the `\t`s back

arvidn:
I'm sure it's fine. hopefully changes in `RC_2_0` will have tests anyway, to fail if they don't merge correctly

PR #5857, opened 2021-01-12
moved deadline_timer.hpp and debug.hpp to aux

aldenml:


arvidn:
it looks like something is exposed to clients and are now causing link errors: https://github.com/arvidn/libtorrent/pull/5857/checks?check_run_id=1692881436#step:7:194

arvidn:
actually, not exposed to clients, but probably had the declaration moved into `aux` but not the definition: https://github.com/arvidn/libtorrent/pull/5857/checks?check_run_id=1692881403#step:4:363

arvidn:
the debug stuff is a bit messy when enabled. There's no clean counterpart to `debug.hpp` where the definitions are.

arvidn:
I think [these](https://github.com/arvidn/libtorrent/blob/master/src/session_impl.cpp#L160) need to me moved to `libtorrent::aux`

aldenml:
thanks @arvidn, I think that is the problem (moving the definitions to aux)

Issue #5340, opened 2020-12-11
compile error in master branch with web sockets


MassaRoddel:
I try to build the master branch with websockets enabled. But there are two compiler errors in websocket_tracker_connection.cpp.
I use VS2019 (v142) 64bit address model.
These two lines fail to compile:

json::string const data = json::to_string(payload);

with error: connot convert "boost::json::object" to "boost::json::kind"

any ideas?

MassaRoddel:
I changed it to:

json::string const data = json::serialize(payload);

is that ok?

aldenml:
Hi @MassaRoddel, what boost version are you using? I found the same error but with `1.75.0`

aldenml:
See https://github.com/arvidn/libtorrent/pull/5738

MassaRoddel:
1.73.0

aldenml:
I see, then you are in a similar situation as if using `1.75.0`, I imagine your boost clone is pulling Boost.JSON and overriding the include of the local one inside `deps`. If that's the case, I would just remove the one from github.

MassaRoddel:
switched to boost 1.75.0 and removed boost.json

paullouisageneau:
#5738 breaks compilation with boost 1.75.0 if you don't manually remove the `deps/json` submodule, as tries to include `deps/json/include/boost/json/src.hpp` which fails with `error: #error Do not compile library source with the header only macro defined`.

aldenml:
@paullouisageneau do you have a CI job log with the problem that we can take a look at?

paullouisageneau:
No it's only local. Don't worry, I'll update the submodule and push a fix.

PR #5869, opened 2021-01-15
Fix Boost.JSON submodule

paullouisageneau:
Re-fixes https://github.com/arvidn/libtorrent/issues/5340

#5738 breaks compilation with boost 1.75.0 if the compiler tries to include `deps/json/include/boost/json/src.hpp` which fails with the error `error: #error Do not compile library source with the header only macro defined`.

This PR updates the `deps/json` submodule and links it only as fallback if it's missing from Boost.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,9 +80,7 @@ set(libtorrent_include_files
 	session_status
 	session_types
 	settings_pack
-	sha1
 	sha1_hash
-	sha256
 	socket
 	socket_type
 	socks5_stream
@@ -231,6 +229,9 @@ set(libtorrent_aux_include_files
 	session_settings
 	session_udp_sockets
 	set_socket_buffer
+	sha1
+	sha256
+	sha512
 	sliding_average
 	socket_io
 	socket_type
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -880,9 +880,17 @@ include(CheckCXXCompilerFlag)
 add_subdirectory(bindings)
 
 if(webtorrent)
-target_link_libraries(torrent-rasterbar PRIVATE LibDataChannel::LibDataChannelStatic)
-target_include_directories(torrent-rasterbar PRIVATE deps/json/include)
-target_compile_definitions(torrent-rasterbar PRIVATE BOOST_JSON_HEADER_ONLY)
+	target_link_libraries(torrent-rasterbar PRIVATE LibDataChannel::LibDataChannelStatic)
+
+	# Boost.JSON was added to Boost in version 1.75
+	find_package(Boost OPTIONAL_COMPONENTS json)
+	if(Boost_JSON_FOUND)
+		target_link_libraries(torrent-rasterbar PUBLIC ${Boost_JSON_LIBRARY})
+	else()
+		# Fallback to the submodule
+		target_compile_definitions(torrent-rasterbar PRIVATE BOOST_JSON_HEADER_ONLY)
+		target_include_directories(torrent-rasterbar PRIVATE deps/json/include)
+	endif()
 endif()
 
 if(IS_ABSOLUTE "${CMAKE_INSTALL_LIBDIR}")
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,6 +17,9 @@ environment:
       ssl_include: /usr/local/lib
       lib: 1
     - cmake: 1
+      crypto: openssl
+      ssl_root_dir: c:\OpenSSL-v111-Win64
+      webtorrent: on
     - variant: release
       compiler: msvc
       model: 64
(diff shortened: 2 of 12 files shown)


paullouisageneau:
Does someone know how to properly check for Boost components or version in Jamfile?

arvidn:
I'll fix it. hold on

CMakeLists.txt
```diff
@@ -880,9 +880,17 @@ include(CheckCXXCompilerFlag)
...
-target_compile_definitions(torrent-rasterbar PRIVATE BOOST_JSON_HEADER_ONLY)
+	target_link_libraries(torrent-rasterbar PRIVATE LibDataChannel::LibDataChannelStatic)
+
+	# Boost.JSON was added to Boost in version 1.75
+	find_package(Boost OPTIONAL_COMPONENTS json)
+	if(Boost_JSON_FOUND)
+		target_link_libraries(torrent-rasterbar PUBLIC ${Boost_JSON_LIBRARY})
+	else()
+		# Fallback to the submodule
+		target_compile_definitions(torrent-rasterbar PRIVATE BOOST_JSON_HEADER_ONLY)
```
arvidn, line 891: when we're using it as header-only, we need to include `src.hpp` too, don't we?

arvidn: I have a feeling that it might be easier to use boost.json as header-only until libtorrent can require boost >= 1.75, and unconditionally rely on it from boost. But I could be wrong.

paullouisageneau: Oh, good point, the way to use the library header-only changed in the doc from defining `BOOST_JSON_HEADER_ONLY` to including `boost/json/src.hpp`. I'll fix that.

paullouisageneau: Using the library all the time would simplify the compilation setup a bit but I'm afraid I'll conflict with the one present in Boost.

arvidn: right. I posted this PR against your PR. I may have misunderstood the header-only use then.

https://github.com/paullouisageneau/libtorrent/pull/1

(feel free to squash my commit in btw)

arvidn: I think the reason this worked in CI was because there is no cmake build where webtorrent is enabled. feel free to enable it. I think maybe appveyor is the only remaining cmake coverage on CI right now

paullouisageneau: Actually, webtorrent is already enabled for the msvc build on appveyor.

arvidn: this is the command line I see executed on appveyor:

```
cmake -DBOOST_LIBRARYDIR=%BOOST_ROOT%\stage\lib -DCMAKE_CXX_STANDARD=17 -Dbuild_tests=ON -Dbuild_examples=ON -Dbuild_tools=ON -Dpython-bindings=%python% -Dboost-python-module-name="python" -Dskip-python-runtime-test=true -DPython_ADDITIONAL_VERSIONS="2.7" -G "Visual Studio 16 2019" -A x64 .. && cmake --build . --config Release --parallel %NUMBER_OF_PROCESSORS% -- -verbosity:minimal
```

And in the cmake output, I see:

```
-- The following features have been disabled:
 * static_runtime, build libtorrent with static runtime
 * python-bindings, build python bindings
 * python-egg-info, generate python egg info
 * python-install-system-dir, Install python bindings to the system installation directory rather than the CMake installation prefix
 * gnutls, build using GnuTLS instead of OpenSSL
 * webtorrent, build with WebTorrent support
```

paullouisageneau: Indeed, the option is passed to b2 but not to cmake.

appveyor.yml
```diff
@@ -94,7 +94,7 @@ build_script:
     cd %ROOT_DIRECTORY% &&
     mkdir build &&
     cd build &&
-    cmake -DBOOST_LIBRARYDIR=%BOOST_ROOT%\stage\lib -DCMAKE_CXX_STANDARD=17 -Dbuild_tests=ON -Dbuild_examples=ON -Dbuild_tools=ON -Dpython-bindings=%python% -Dboost-python-module-name="python" -Dskip-python-runtime-test=true -DPython_ADDITIONAL_VERSIONS="2.7" -G "Visual Studio 16 2019" -A x64 .. &&
+    cmake -DBOOST_LIBRARYDIR=%BOOST_ROOT%\stage\lib -DCMAKE_CXX_STANDARD=17 -Dbuild_tests=ON -Dbuild_examples=ON -Dbuild_tools=ON -Dwebtorrent=%webtorrent% -Dpython-bindings=%python% -Dboost-python-module-name="python" -Dskip-python-runtime-test=true -DPython_ADDITIONAL_VERSIONS="2.7" -G "Visual Studio 16 2019" -A x64 .. &&
```
arvidn, line 97: I don't think this is enough. `webtorrent` defaults to `off`. I think you have to add it to the build matrix too

paullouisageneau: It's already on in the build matrix:
https://github.com/arvidn/libtorrent/blob/ee7bd3b109d2d63d662486d707c33b5777717ed3/appveyor.yml#L26

arvidn: but that's not for the `cmake` build.
right under:
```
- cmake: 1
```

add:

```
  webtorrent: on
```

paullouisageneau: Oh, yes...

arvidn (approved):

PR #5875, opened 2021-01-16
moved lsd.hpp to aux

aldenml:


aldenml:
@arvidn this is just moving what was in `libtorrent/lsd.hpp` into the existing file `libtorrent/aux_/lsd.hpp`

arvidn (approved):

PR #5877, opened 2021-01-17
Fix possible overflow when setting connection limit

JaySandhu:
For large values of `max_files` the expression `(max_files - 20) * 8 / 10` overflows and becomes negative, when this happens `connections_limit` is set to 5, which is very low. It is common for `max_open_files()` to return `INT_MAX` so overflow handling is needed. An easy fix is to change it so that the division happens first, `((max_files - 20) / 10) * 8`. That's what this PR does.

Ref: https://github.com/qbittorrent/qBittorrent/issues/13747#issuecomment-761716512

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -668,7 +668,7 @@ bool ssl_server_name_callback(ssl::stream_handle_type stream_handle, std::string
 		// 80% of the available file descriptors should go to connections
 		m_settings.set_int(settings_pack::connections_limit, std::min(
 			m_settings.get_int(settings_pack::connections_limit)
-			, std::max(5, (max_files - 20) * 8 / 10)));
+			, std::max(5, ((max_files - 20) / 10) * 8)));
 		// 20% goes towards regular files (see disk_io_thread)
 #ifndef TORRENT_DISABLE_LOGGING
 		if (should_log())


arvidn:
good catch! It's worse than `connections_limit` being set to 5, there's undefined behavior.
However, it's also common for `max_open_files()` to return 256, which would truncate poorly with the division first.

arvidn:
I prefer solving this problem this way: https://github.com/arvidn/libtorrent/pull/5878
I think it's generally risky to work with limits in the billions.

JaySandhu:
> However, it's also common for `max_open_files()` to return 256, which would truncate poorly with the division first.

For what it's worth, 256 would evaluate to 184 with the division first, vs 188 with it after.

> I prefer solving this problem this way: #5878
> I think it's generally risky to work with limits in the billions.

Yup, that solution works too 👍. Probably is a good idea to keep this number somewhat sane.

JaySandhu:
Closing in favor of #5878.

arvidn:
> For what it's worth, 256 would evaluate to 184 with the division first, vs 188 with it after.

Yes, I admit that the uncertainty around random file descriptors that aren't "tracked" by any limit is probably greater than 4 (things like tracker connections, DNS lookups, futexes, reading from /dev/urandom etc.). But on the other hand, at small numbers, 4 more connections have a larger marginal effect.

Issue #5831, opened 2021-01-07
[WebTorrent] instant.io has trouble downloading from libtorrent seeds


lucybythec:
Hi! Thank you for this project. I discovered the following issue. When libtorrent is built with `-Dwebtorrent=on`, it can successfully and quickly download from magnet links created on instant.io. However, it doesn't work very well in reverse:

- Seed with instant.io, download with instant.io: Success
- Seed with libtorrent, download with libtorrent: Success
- Seed with instant.io, download with libtorrent: Success
- Seed with libtorrent, download with instant.io: Failure

In the last case, instant.io either forever stays at "Downloading torrent", or succeeds after a long time (over 10 minutes). Perhaps there is a compatibility issue?

Here is the test program for seeding:

```python
import os
import os.path
import time

import libtorrent as lt

lt_session = lt.session()

fn = 'testfile.bin'
with open(fn, 'wb') as f:
    f.write(os.urandom(1024 * 1024))

lt_file_storage = lt.file_storage()
lt.add_files(lt_file_storage, fn)

lt_create_torrent = lt.create_torrent(lt_file_storage)
# Same tracker list as used by instant.io
lt_create_torrent.add_tracker('wss://tracker.btorrent.xyz')
lt_create_torrent.add_tracker('wss://tracker.openwebtorrent.com')
lt_create_torrent.add_tracker('udp://tracker.leechers-paradise.org:6969')
lt_create_torrent.add_tracker('udp://tracker.coppersurfer.tk:6969')
lt_create_torrent.add_tracker('udp://tracker.opentrackr.org:1337')
lt_create_torrent.add_tracker('udp://explodie.org:6969')
lt_create_torrent.add_tracker('udp://tracker.empire-js.us:1337')

lt.set_piece_hashes(lt_create_torrent, '.')

lt_torrent_info = lt.torrent_info(lt_create_torrent.generate())
lt_add_torrent_params = {
    'ti': lt_torrent_info,
    'save_path': '.',
}

lt_torrent_handle = lt_session.add_torrent(lt_add_torrent_params)

torrent_magnet_uri = lt.make_magnet_uri(lt_torrent_handle)
print('* Seeding: ' + torrent_magnet_uri)

while True:
    s = lt_torrent_handle.status()

    state_str = ['queued', 'checking', 'downloading metadata',
                 'downloading', 'finished', 'seeding', 'allocating',
                 'checking fastresume']

[...]

PR #5882, opened 2021-01-19
WebTorrent: Add keepalive with ping on WebSocket

paullouisageneau:
This PR adds regular pings on the WebSocket to keep the connection to the tracker alive.

It should help with https://github.com/arvidn/libtorrent/issues/5831

### DIFF
--- a/src/websocket_stream.cpp
+++ b/src/websocket_stream.cpp
@@ -288,7 +288,7 @@ void websocket_stream::on_close(error_code)
 
 void websocket_stream::on_keepalive(error_code ec)
 {
-	if (ec) return;
+	if (ec || !m_open) return;
 
 	ADD_OUTSTANDING_ASYNC("websocket_stream::on_ping");
 	std::visit([&](auto& stream)
--- a/src/websocket_stream.cpp
+++ b/src/websocket_stream.cpp
@@ -15,6 +15,7 @@ see LICENSE file.
 #include "libtorrent/aux_/debug.hpp"
 #include "libtorrent/error.hpp"
 #include "libtorrent/io_context.hpp"
+#include "libtorrent/time.hpp"
 #include "libtorrent/aux_/parse_url.hpp"
 #include "libtorrent/aux_/random.hpp"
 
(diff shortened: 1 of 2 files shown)


src/websocket_stream.cpp
```diff
@@ -277,6 +286,35 @@ void websocket_stream::on_close(error_code)
 	COMPLETE_ASYNC("websocket_stream::on_close");
 }
 
+void websocket_stream::on_keepalive(error_code ec)
+{
+	if (ec) return;
+
```
arvidn, line 292: I think there's a risk for this to be called with `m_open` = `false`. It seems like that should be an early exit here.

src/websocket_stream.cpp
```diff
@@ -277,6 +286,35 @@ void websocket_stream::on_close(error_code)
...
+}
+
+void websocket_stream::on_ping(error_code ec)
+{
+	COMPLETE_ASYNC("websocket_stream::on_ping");
+
+	if (ec) return;
+
+	arm_keepalive();
```
arvidn, line 309: This looks suspicious to me. Just because the other end is sending me a ping, does that mean I don't have to send one? This seems similar to calling `arm_keepalive()` in `on_read()`

paullouisageneau: Yes, the naming is a bit misleading, but other callbacks follow the same naming pattern.

arvidn (approved):

src/websocket_stream.cpp
```diff
@@ -26,6 +27,8 @@ see LICENSE file.
 #include <algorithm>
 #include <tuple>
 
+#define WEBSOCKET_KEEPALIVE_PERIOD 10 // seconds
```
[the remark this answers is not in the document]
arvidn: I guess `seconds` isn't in scope. this constant should probably be moved into the `libtorrent` namespace.

https://github.com/arvidn/libtorrent/pull/5882/checks?check_run_id=1729261604#step:4:155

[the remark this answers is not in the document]
arvidn: (or `libtorrent::aux`)

arvidn:
did you consider using the built-in support for keepalives btw?
https://www.boost.org/doc/libs/1_74_0/libs/beast/doc/html/beast/using_websocket/timeouts.html

PR #5887, opened 2021-01-20
Adding sequential_start to piece_picker to speed up sequential downloads

master255:


src/piece_picker.cpp
```diff
@@ -2078,7 +2083,7 @@ namespace {
 		{
 			if (m_dirty) update_pieces();
 			TORRENT_ASSERT(!m_dirty);
-
+            //The piece priority logic may be removed in the future.
```
arvidn, line 2086: this should be removed

src/torrent_handle.cpp
```diff
@@ -691,6 +691,11 @@ namespace libtorrent {
 		return sync_call_ret<storage_interface*>(nullptr, &torrent::get_storage_impl);
 	}
 
+    void torrent_handle::set_sequential_start(piece_index_t piece) const
+    {
+        async_call(&torrent::set_sequential_start, piece);
```
arvidn, line 696: It would seem more appropriate to set the start and end at the same time as sequential mode is enabled. It really should be just one round-trip to the network thread too.

What do you think of a function that enables sequential mode and takes a start- and end piece as well?

include/libtorrent/piece_picker.hpp
```diff
@@ -850,6 +852,7 @@ namespace libtorrent {
 		// the number of pieces we have that also are filtered
 		int m_num_have_filtered = 0;
 
+		piece_index_t sequential_start {0};
```
arvidn, line 855: member variables have a `m_` prefix

src/piece_picker.cpp
```diff
@@ -367,6 +367,11 @@ namespace libtorrent {
 		*zero_prio = int(m_downloads[piece_pos::piece_zero_prio].size());
 	}
 
+    void piece_picker::set_sequential_start(piece_index_t const piece)
+    {
+        sequential_start = piece;
```
arvidn, line 372: there should be a bounds-check to ensure this always is set to a valid piece index

src/piece_picker.cpp
```diff
@@ -2117,7 +2122,7 @@ namespace {
 				}
 				else
 				{
-					for (piece_index_t i = m_cursor; i < m_reverse_cursor; ++i)
+					for (piece_index_t i = m_cursor > sequential_start ? m_cursor : sequential_start; i < m_reverse_cursor; ++i)
```
arvidn, line 2125: Wouldn't it be better to set `m_cursor` instead of having this check every time?

src/torrent.cpp
```diff
@@ -1293,6 +1293,12 @@ bool is_downloading_state(int const st)
 		return m_ses.disk_thread().get_torrent(m_storage);
 	}
 
+    void torrent::set_sequential_start(piece_index_t piece)
+    {
+        if (!m_picker) return;
+        m_picker->set_sequential_start(piece);
```
arvidn, line 1299: if the torrent doesn't have a piece picker at this point (say, because it's a magnet link that hasn't received metadata), this setting will silently be ignored. That doesn't seem right.

src/piece_picker.cpp
arvidn: some consideration should be given the case where sequential mode is enabled, and `m_cursror == m_reverse_cursor` but not all pieces have been downloaded. It seems suspicious if the sequential start would implicitly set some pieces to "do not download" priority.

arvidn (changes requested):
there are some formatting issues.
I think new features should target `master`

src/piece_picker.cpp
master255: Ok. Don't forget about it in the future.

arvidn: if you think the piece priority feature should be removed, please start a separate thread/ticket where you present your arguments

src/torrent_handle.cpp
master255: No. You don't understand.
The start piece must be set asynchronously at any point in time.
And the sooner it applies, then the better.
The start can change regularly while the data is being downloaded.
So combining this with the sequential mode enable function is not possible.

Specifying the end piece is not necessary.

arvidn: setting it at the same time as you set sequential mode is as soon as it can set, so, the best, right?

arvidn: specifying end is necessary if you don't want sequential mode past a certain point

src/piece_picker.cpp
master255: If it's not the right piece index, will the download just stop? There will be no error?
I don't think we need to add a check here.

arvidn: whatever behavior you think is the most reasonable should be documented.

arvidn:
there needs to be new tests covering this as well

src/torrent_handle.cpp
master255: > setting it at the same time as you set sequential mode is as soon as it can set, so, the best, right?

No. The start of data downloading in sequential mode can start as usual. There is no need to set 0. The starting piece will need to be set some time after the start of downloading.

> specifying end is necessary if you don't want sequential mode past a certain point

There is no such scenario. This function is definitely not needed.

arvidn: > No. The start of data downloading in sequential mode can start as usual.

That's suboptimal. Why create an API that invites suboptimal behavior?

> There is no such scenario. This function is definitely not needed.

How about, I want to download file 2 sequentially, the other files I don't care about. (i.e. they should be rarest first)

src/piece_picker.cpp
master255: > Wouldn't it be better to set m_cursor instead of having this check every time?

I thought about it.
It works for me like this. Since the starting piece can change at any time. It would be best to check this every time. This will increase the dynamics of loading the necessary data.

I leave it up to you to choose how to do this.

> some consideration should be given the case where sequential mode is enabled, and m_cursror == m_reverse_cursor but not all pieces have been downloaded. It seems suspicious if the sequential start would implicitly set some pieces to "do not download" priority.

Maybe that's really the problem. What do you suggest as a solution?

arvidn: > Maybe that's really the problem. What do you suggest as a solution?

I think disabling sequential mode and resetting the boundaries would be appropriate.

src/torrent.cpp
master255: @arvidn This is all very right. There is no way in the library to check if m_picker is created. So sometimes calls happen at the wrong time. In order not to get a crash - this check.

arvidn: well, it's not documented to have this limitation

src/piece_picker.cpp
master255: Let me leave the documentation to you. I have a different style of documentation. I don't want to spoil your style of documentation that you already have.

Besides, I don't really understand what needs to be documented.

src/torrent_handle.cpp
master255: @arvidn 

> setting it at the same time as you set sequential mode is as soon as it can set, so, the best, right?

Okay. I looked at my code. Indeed, sometimes there is a need to specify the start before the start of loading.

Let's add this function in the next pull request.

src/piece_picker.cpp
master255: @arvidn Argument 1: no usage scenarios.

arvidn: please start a separate thread/ticket where you present your arguments

src/piece_picker.cpp
master255: @arvidn Okay. This can be done by other methods. There is no need to do this inside the library.

src/torrent.cpp
master255: @arvidn Yes. You're right. We should document that. Where to document this?

arvidn: add or update comments above public functions in public headers. In this case, `include/libtorrent/torrent_handle.hpp`. It might also make sense to update the piece picker documentation in `docs/manual.rst`

master255:
@arvidn I can't solve the formatting problems. Correct them yourself, please. I have tried removing spaces and adding tabs, but nothing changes.

I will leave it to you to write the tests for this. I don't really understand what to test here. But you, as a knowledgeable man can write them.

master255:
> I think new features should target master

I see you regularly merge changes between the master and this branch. I need these changes in this branch first of all. Until FrostByte moves to branch 2.0.

src/torrent.cpp
master255: @arvidn Will you add the documentation yourself to docs/manual.rst ? Okay? 
I tried to describe set_sequential_start correctly in torrent_handle.hpp. You can take a look, adjust and copy it where you need it.

master255:
@arvidn I answered all the questions. Anything else?

arvidn:
> Anything else?

* formatting
* rebasing on top of master
* tests, including negative tests and the case of all sequential pieces being downloaded, but there are still pieces left
* documenting the edge cases.
* provide an end-piece index, or a really *good* reason not to
* tie setting the piece index and the sequential flag together in some reasonable way, and clearing them. The piece index should probably be reset when clearing the sequential flag

include/libtorrent/piece_picker.hpp
```diff
@@ -850,6 +852,7 @@ namespace libtorrent {
 		// the number of pieces we have that also are filtered
 		int m_num_have_filtered = 0;
 
+		piece_index_t m_sequential_start {0};
```
arvidn, line 855: if `m_cursor` and `m_reverse_cursor` are only used for sequential download, I think it makes sense to use those instead of adding a new field. Their semantics would have to be altered a bit though, the comments wouldn't necessarily be true

master255: @arvidn Okay. This is a difficult question for me. It has to be tested. Do you want to assign m_cursor?

master255:
> The piece index should probably be reset when clearing the sequential flag

If it will be a separate variable, it is not necessary.
If it is m_cursor, then yes. It needs to be reset, but to what value?

> provide an end-piece index, or a really good reason not to

In what scenario would that be needed?
To download a file from a set, I prioritize the file. The last piece is set automatically. What is the reason to set it manually?

include/libtorrent/piece_picker.hpp
master255: @arvidn Many factors affect m_cursor. Are you sure there won't be a problem with asynchronous assignment? Perhaps a problem with concurrency?

arvidn: no, I'm not sure. but iirc, the only purpose that member exists is for implementing sequential download. I would expect it make for simpler code, overall, to tweak the semantics of this rather than adding another member

master255:
> rebasing on top of master

To do this, it is necessary to make a new Pull Request.

arvidn:
> If it will be a separate variable, it is not necessary.

Whether it's necessary or not is independent on how it's implemented. I think it's unintuitive and error prone to leave this state after disabling sequential download. I'm open to be convinced otherwise.

> If it is m_cursor, then yes. It needs to be reset, but to what value?

I would think it would make sense that enabling sequential download would have the same semantics as it has today, regardless of any previous sequential mode with a specific range. i.e. reset to the whole torrent again. But maybe sequential mode without a range should be discouraged by the API generally.

> In what scenario would that be needed?

Wanting to download one file sequentially, and all other files with rarest first.

> To do this, it is necessary to make a new Pull Request.

You can change the target for a PR too, but either way is fine I think

master255:
@arvidn 

> Whether it's necessary or not is independent on how it's implemented. I think it's unintuitive and error prone to leave this state after disabling sequential download. I'm open to be convinced otherwise.

Okay. I added the zeroing of the piece.

> Wanting to download one file sequentially, and all other files with rarest first.

No. This is not a real scenario. You are trying to move the logic of the application to the functionality of the library. This is not right. The application keeps track of the downloaded files and controls how they are downloaded. It sets the priorities and the type of loading data. The library is not designed to do that. Other than that, it's complicated to calculate the final piece.

> You can change the target for a PR too, but either way is fine I think

I changed the branch to master, but now we have a lot of problems and conflicts. What to do?

arvidn:
> The application keeps track of the downloaded files and controls how they are downloaded. It sets the priorities and the type of loading data. The library is not designed to do that. Other than that, it's complicated to calculate the final piece.

The application can't control this without inherent race conditions. The network thread won't stop and wait for the client to decide whether to keep downloading or not. (and it shouldn't, it would poorly impact performance)

> I changed the branch to master, but now we have a lot of problems and conflicts. What to do?

Did you rebase your branch? If not, do that.

markmdscott:
@master255 maybe I can be of service with some of the changes @arvidn is requesting if you're not able to do them? I'm a casual developer but I think I can try.

master255:
@markmdscott You can write tests and test them. I'm not sure you can do it. It's not an easy task.
If you can write the tests, that would be very good.

AllSeeingEyeTolledEweSew:
Drive-by two cents here, but what's the use case? Would `set_piece_deadline` be a better fit?

arvidn:
the current `set_piece_deadline()` implementation has performance issues. Specifically, it scales poorly with many peers and frequent piece-requests.

master255:
@arvidn Well, you need to optimize that. Because:
1. it's already working.
2. it reduces the number of method calls (and the amount of data passed to the library) to the library. Which, in practice, significantly improves performance over piece prioritization.

markmdscott:
@master255 did you rebase back to the 1.2 branch? (It says: "master255 changed the base branch from master to RC_1_2")

Should this not be build on top of the master branch?

Also people just mentioned ```set_piece_deadline()``` up above. Should I continue with trying to work on this patch or wait until a decision is resolved on the direction to go?

master255:
@markmdscott Yes. I'm starting to do it right now.
You don't have to wait for anything. We have a direction.

master255:
@arvidn 

> The application can't control this without inherent race conditions. The network thread won't stop and wait for the client to decide whether to keep downloading or not. (and it shouldn't, it would poorly impact performance)

This is how my player works. I don't see any particular reason to set the end.
Let's go in order:
1. We have an end of file loading event in the application.
2. When the application knows the end of a sequential file download, it can change the file download priorities and download type to non-sequential.
3. All of these actions are done as quickly as possible, but if we constantly set the end of sequential loading, we need to regularly calculate the last piece needed by a number of parameters.
4. In my player, playing torrents online works without specifying the end piece. So it doesn't need to do that.

include/libtorrent/piece_picker.hpp
master255: @arvidn I'm going to test it now and let you know the results.

arvidn:
> In my player, playing torrents online works without specifying the end piece. So it doesn't need to do that.

I don't think that's a reasonable benchmark. By that standard sequential download isn't necessary either. A client could just constantly adjust piece priorities to almost get the same behavior.

I think a more reasonable benchmark is whether it can be done *efficiently*. If you don't care about the end of sequential download, I would argue that's less efficient than if you would. If you care about it but you implement it by having the client react as fast as it can, I would argue that that's also less efficient than if it would happen immediately by the piece picker instead (without race condition).

master255:
@arvidn I wrote a torrent player with the ability to download any torrent files. It is much more complex than just a file downloader like frost wire or qbittorent. I've tested all possible scenarios, torrent downloads and playback, and I can't understand what you're talking about. 
Perhaps you need to talk more simply. You write a lot of text that I don't understand.
Maybe you need to show me the scenario you want to improve in my player? That would make as much sense as possible.

Okay. Do you probably have an example of a program where this would have to be done? (continuing to download data chaotically, after the sequential download is over).

Arvin, people need software, not philosophy.

AllSeeingEyeTolledEweSew:
> Okay. Do you probably have an example of a program where this would have to be done? (continuing to download data chaotically, after the sequential download is over).

I thought [the argument against sequential downloading](https://wiki.vuze.com/w/Sequential_downloading_is_bad) was that the bias in requests leads to seeds quitting when there's a bias in availability, which leads to less swarm health. If you believe the argument, you'd at want to at least _minimize_ sequential downloading, and switch to rarest-first as soon as your player's buffer is full.

Do you have a link to your player? I'm interested in torrent video playback.

master255:
@AllSeeingEyeTolledEweSew [Yes, of course.](https://play.google.com/store/apps/details?id=com.media.library) Sequential downloading has been used in torrents for a long time. Your information is not actual.

include/libtorrent/piece_picker.hpp
master255: @arvidn I just checked it out.
If I remove the m_sequential_start variable and use m_cursor instead, the data download fading starts to appear. Only the data to which deadline is assigned gets downloaded.

master255:
@arvidn One more thing. I haven't written about this, but it is indirectly related to sequential data loading.
In file_storage we need to add the second file names. Without this it is impossible to understand which file was loaded. Maybe in another PR I'll talk about this in more detail.

arvidn:
there's also an "auto-sequential" feature that need to be considered. It's probably a good idea to clear any user-set boundaries when the user clears the sequential flag too, to avoid breaking it.

https://github.com/arvidn/libtorrent/blob/RC_2_0/src/torrent.cpp#L3486

master255:
@arvidn All very bad.
1. Version 2.0 of the library is not sufficiently tested. It is not compatible with some versions of Android.
2. To make a rebase to version 2.0 I need to test version 2.0. For this I need to compile it for android. But I can not do this because there is no script that can do this.
3. @aldenml wrote [some script](https://github.com/aldenml/libtorrent4j/issues/155) to compile it, but to test it I need to do merge in master libtorrent 2.0... It's a vicious circle.

markmdscott:
@master255 I've been trying to figure out how to rebase your patch onto 2.0. I guess you've been trying also. My question is that should the version 2.0 be tested before rebasing your patch? Because my thinking is that - and admittedly you know about this more than me - Arvid also has tests for version 2.0, so it's already been tested. The tests are needed after applying your patch, after rebasing your patch onto version 2.0. Mobile work on this can be done later, after the desktop testing of the patch as part of version 2.0 is complete and confidence is assured then it can be tested on mobile as part of the overall version 2.0 tests on mobile.

master255:
@markmdscott Testing in real situations reveals more problems than unit tests.
Version 2.0 of the library is very different from version 1.2, so it is better to test on a real device in real battle situations first.
This requires a compilation script. It's in development right now.

master255:
@arvidn Libtorrent 2.0 does not work on Android TV and Android api 19. This needs to be fixed. Without it, torrent streaming is useless.
https://github.com/aldenml/libtorrent4j/issues/155#issuecomment-772955983

master255:
@arvidn Since version 2.0 has a lot of bugs:
https://github.com/arvidn/libtorrent/issues/6033
https://github.com/arvidn/libtorrent/issues/6031
https://github.com/arvidn/libtorrent/issues/6030
https://github.com/arvidn/libtorrent/issues/6023

 and can not be used, I think merging streaming in it makes no sense.
Therefore, my merge is correct. You should approve it.

arvidn:
> Since version 2.0 has a lot of bugs

those tickets are mostly based on your misreading of the documentation (it almost seems you haven't read the documentation).

> and can not be used, I think merging streaming in it makes no sense.
> Therefore, my merge is correct.

None of my comments towards this patch have been addressed and none of them are affected by your claim (even if it were correct).

markmdscott:
> @arvidn Since version 2.0 has a lot of bugs:
> #6033
> #6031
> #6030
> #6023
> 
> and can not be used, I think merging streaming in it makes no sense.
> Therefore, my merge is correct. You should approve it.

Oleg I'm making good progress on the PR, and will look to backport it to 1.2, so you needn't worry. If you have any comments on the proposed functionality or suggestions for improvements let me know.

master255:
@markmdscott I see the following problems:
1. Libtorrent 2.0 is too unstable to implement streaming. The bugs I listed need to be fixed first, otherwise it will work even more unstable.
2. I don't understand why you added the final piece. You don't have a ready-made app and you're inventing, while I have a ready-made player and I really know what's needed and what's not.
3. You're not trying to argue with Arvin, who doesn't know anything about streaming at all.
4. You and Arvin have a poor understanding of which side to perform operations on. You're trying to implement application functionality into a library and library functionality into an application. This isn't right.
5. Your code is not professional enough and will run slowly. Performance in streaming is more important than just downloading files.
6. I won't be able to use this in my player.
7. Do you check what you write? I tested my code a long time ago and I use it on my production.

master255:
@arvidn 

> those tickets are mostly based on your misreading of the documentation (it almost seems you haven't read the documentation).

So why are the bugs still open? And you haven't given answers to them.
Arvin, torrents 2.0 is still in beta. I haven't created bugs yet, but they are there. I had a case where the metadata downloaded, but the files did not start downloading.
We will fix bugs in version 2.0 but it cannot be used now.

arvidn:
> Your code is not professional enough and will run slowly. Performance in streaming is more important than just downloading files.

@master255 it's one thing that you're rude to me, but if you're going to be rude to other people in this forum I will have to put an end to it. Your communication is not professional.

arvidn:
> So why are the bugs still open? And you haven't given answers to them.

Your statement is non-sensical. All of those tickets have responses from me, and *you* have closed half of them.

> I haven't created bugs yet, but they are there.

When you say "bugs", do you mean software faults, or do you mean github tickets? Typically the term "bug" is used to refer to the former.

> I had a case where the metadata downloaded, but the files did not start downloading.
> We will fix bugs in version 2.0 but it cannot be used now.

Given your track record so far, these are likely your bugs. But of course I'm open to the possibility there are bugs in libtorrent-2.0. None of those tickets demonstrate that though.

arvidn:
> I don't understand why you added the final piece. You don't have a ready-made app and you're inventing, while I have a ready-made player and I really know what's needed and what's not.

This has been explained several times. Please counter those arguments and explain why they are wrong instead of pretending they don't exist. What you're doing is not constructive.

master255:
@arvidn Okay. I won't write anymore. Do whatever you want. Everything works for me and there is no point in me updating it or changing anything.
I understand that you take help for hostile behavior.

PR #5892, opened 2021-01-21
Build as release

danger89:
By default it was building a debug build. I think this is not wanted when running in production.

Regards,
Melroy vd Berg

### DIFF
--- a/docs/building.rst
+++ b/docs/building.rst
@@ -72,13 +72,13 @@ Linux::
 
 	sudo apt install libboost-tools-dev libboost-dev libboost-system-dev
 	echo "using gcc ;" >>~/user-config.jam
-	b2 crypto=openssl cxxstd=14 variant=release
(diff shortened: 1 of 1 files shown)


docs/building.rst
```diff
@@ -72,21 +72,21 @@ Linux::
 
 	sudo apt install libboost-tools-dev libboost-dev libboost-system-dev
 	echo "using gcc ;" >>~/user-config.jam
-	b2 crypto=openssl cxxstd=14
+	b2 crypto=openssl cxxstd=14 variant=release
```
arvidn, line 75: `variant=` is not needed, you can just say `release`. I would prefer that.

arvidn (changes requested):
looks good apart from the comment

docs/building.rst
danger89: if you say so.

danger89:
reworked

PR #5874, opened 2021-01-16
generate and install pkg-config file when invoking the install rule

arvidn:

PR #5894, opened 2021-01-22
fix version checking for using package.paths in Jamfile

arvidn:


emmercm:
Waiting on the build from the v1.2.12 tag with patches from #5026, #5874, and this one.

I'm hoping to receive the same fix for v1.2.11 and v2.0.1 (having separate issues with v2.0.2) - I'm assuming the version tags are treated as immutable, is there a good way to port some of these changes elsewhere? Somewhat naive question, I'm not the strongest with C/C++/Boost/Make/etc.

emmercm:
This worked for me with v1.2.12, `pkg-config --list-all` found `libtorrent-rasterbar` like qBittorrent needs!

arvidn:
I will merge these into 2.0.x as well

PR #5896, opened 2021-01-22
fix load_file() to support non-ascii characters on windows

arvidn:
used to load files in the torrent_info constructor and the posix part file


arvidn:
@Gallardo994 I added a unit test that passes on windows for me. Are you sure you ended up testing the freshly built version?

PR #5902, opened 2021-01-23
[5/?] Add pre-commit hooks for python code style, and clean up setup.py

AllSeeingEyeTolledEweSew:
This is based on #5970.

This PR:
* introduces some python auto-formatting and type checking
* applies this to `setup.py`
* adds some minor readability fixes to `setup.py`

The pre-commit config excludes all existing python files except `setup.py` to prevent bloating the PR.

The intent is to illustrate what my proposed python auto-formatting would look like for discussion, and later apply it to other files.

I tried to pick this code style as a reasonable selection of practices in the open source python world. The code style is based on [black](https://github.com/psf/black). It's not very configurable, by design. I added some other auto-formatters to improve source code stability.

I also enforce python type annotations with mypy. In my experience, mypy runs very quickly and is quite good at catching python bugs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,5 +2,5 @@
 import os
 import runpy
 
-os.chdir('bindings/python')
-runpy.run_path('setup.py')
+os.chdir("bindings/python")
+runpy.run_path("setup.py")
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -62,6 +62,11 @@ repos:
             tools/set_version.py|
             tools/update_copyright.py
           )$
+- repo: https://github.com/myint/autoflake
+  rev: v1.4
+  hooks:
+    - id: autoflake
+      args: [--in-place, --remove-unused-variables, --remove-all-unused-imports, --remove-duplicate-keys]
 - repo: https://gitlab.com/pycqa/flake8.git
   rev: 3.8.4
   hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,11 +25,43 @@ repos:
     -   id: check-xml
     -   id: debug-statements
     -   id: check-symlinks
+    -   id: check-toml
 - repo: https://github.com/pre-commit/pygrep-hooks
   rev: v1.7.0
   hooks:
   - id: rst-directive-colons
   - id: rst-inline-touching-normal
+- repo: https://github.com/PyCQA/isort
+  rev: 5.7.0
+  hooks:
+    - id: isort
+      exclude: |
+          (?x)^(
+            # Enable these later, avoid bloating this PR
+            bindings/python/client.py|
+            bindings/python/dummy_data.py|
+            bindings/python/make_torrent.py|
+            bindings/python/simple_client.py|
+            bindings/python/test.py|
+            docs/gen_reference_doc.py|
+            examples/run_benchmarks.py|
+            fuzzers/tools/generate_initial_corpus.py|
+            fuzzers/tools/unify_corpus_names.py|
+            test/socks.py|
+            test/web_server.py|
+            test/websocket_server.py|
+            tools/clean.py|
+            tools/copyright.py|
+            tools/dht_flood.py|
+            tools/parse_dht_log.py|
+            tools/parse_dht_rtt.py|
+            tools/parse_dht_stats.py|
+            tools/parse_session_stats.py|
+            tools/parse_utp_log.py|
+            tools/run_benchmark.py|
+            tools/set_version.py|
+            tools/update_copyright.py
+          )$
 - repo: https://gitlab.com/pycqa/flake8.git
   rev: 3.8.4
   hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -67,6 +67,51 @@ repos:
   hooks:
     - id: autoflake
       args: [--in-place, --remove-unused-variables, --remove-all-unused-imports, --remove-duplicate-keys]
+- repo: https://github.com/python/black
+  rev: 20.8b1
+  hooks:
+    - id: black
+      # Avoiding PR bloat
+      exclude: |
+          (?x)^(
+               bindings/c/tools/gen_alert_header.py|
+               bindings/c/tools/gen_header.py|
+               bindings/python/client.py|
+               bindings/python/dummy_data.py|
+               bindings/python/make_torrent.py|
+               bindings/python/simple_client.py|
+               bindings/python/test.py|
+               docs/filter-rst.py|
+               docs/gen_reference_doc.py|
+               docs/gen_settings_doc.py|
+               docs/gen_stats_doc.py|
+               docs/gen_todo.py|
+               docs/join_rst.py|
+               examples/run_benchmarks.py|
+               fuzzers/tools/generate_initial_corpus.py|
+               fuzzers/tools/unify_corpus_names.py|
+               test/http_proxy.py|
+               test/socks.py|
+               test/web_server.py|
+               test/websocket_server.py|
+               tools/benchmark_checking.py|
+               tools/clean.py|
+               tools/copyright.py|
+               tools/dht_flood.py|
+               tools/gen_convenience_header.py|
+               tools/gen_fwd.py|
+               tools/parse_dht_log.py|
+               tools/parse_dht_rtt.py|
+               tools/parse_dht_stats.py|
+               tools/parse_lookup_log.py|
+               tools/parse_peer_log.py|
+               tools/parse_sample.py|
+               tools/parse_session_stats.py|
+               tools/parse_utp_log.py|
+               tools/run_benchmark.py|
+               tools/set_version.py|
+               tools/update_copyright.py
+          )$
 - repo: https://gitlab.com/pycqa/flake8.git
   rev: 3.8.4
   hooks:
@@ -76,7 +121,27 @@ repos:
             # Enable these later, avoid bloating this PR
             bindings/c/tools/gen_alert_header.py|
             bindings/c/tools/gen_header.py|
+            bindings/python/client.py|
+            bindings/python/test.py|
+            docs/gen_todo.py|
+            docs/gen_reference_doc.py|
+            docs/gen_settings_doc.py|
+            docs/gen_stats_doc.py|
+            examples/run_benchmarks.py|
             fuzzers/tools/generate_initial_corpus.py|
+            test/socks.py|
             test/websocket_server.py|
-            tools/benchmark_checking.py
+            test/web_server.py|
+            tools/benchmark_checking.py|
+            tools/dht_flood.py|
+            tools/gen_fwd.py|
+            tools/parse_dht_stats.py|
+            tools/parse_dht_log.py|
+            tools/parse_lookup_log.py|
+            tools/parse_peer_log.py|
+            tools/parse_session_stats.py|
+            tools/parse_utp_log.py|
+            tools/run_benchmark.py|
+            tools/set_version.py|
+            tools/update_copyright.py
         )$
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -112,6 +112,48 @@ repos:
                tools/set_version.py|
                tools/update_copyright.py
           )$
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: v0.800
+  hooks:
+    - id: mypy
+      # Avoiding PR bloat
+      exclude: |
+          (?x)^(
+              bindings/c/tools/gen_alert_header.py|
+              bindings/c/tools/gen_header.py|
+              bindings/python/client.py|
+              bindings/python/dummy_data.py|
+              bindings/python/make_torrent.py|
+              bindings/python/test.py|
+              docs/filter-rst.py|
+              docs/gen_reference_doc.py|
+              docs/gen_settings_doc.py|
+              docs/gen_stats_doc.py|
+              docs/gen_todo.py|
+              examples/run_benchmarks.py|
+              fuzzers/tools/generate_initial_corpus.py|
+              setup.py|
+              test/http_proxy.py|
+              test/socks.py|
+              test/web_server.py|
+              test/websocket_server.py|
+              tools/benchmark_checking.py|
+              tools/clean.py|
+              tools/copyright.py|
+              tools/dht_flood.py|
+              tools/gen_convenience_header.py|
+              tools/gen_fwd.py|
+              tools/parse_dht_log.py|
+              tools/parse_dht_stats.py|
+              tools/parse_lookup_log.py|
+              tools/parse_peer_log.py|
+              tools/parse_sample.py|
+              tools/parse_session_stats.py|
+              tools/parse_utp_log.py|
+              tools/run_benchmark.py|
+              tools/set_version.py|
+              tools/update_copyright.py
+          )$
 - repo: https://gitlab.com/pycqa/flake8.git
   rev: 3.8.4
   hooks:
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -2,7 +2,9 @@
 
 import contextlib
 from distutils import log
+import distutils.cmd
 import distutils.debug
+import distutils.errors
 import distutils.sysconfig
 import os
 import pathlib
@@ -12,13 +14,19 @@ import subprocess
 import sys
 import sysconfig
 import tempfile
+from typing import cast
+from typing import IO
+from typing import Iterator
+from typing import List
+from typing import Optional
+from typing import Set
 import warnings
 
 import setuptools
-import setuptools.command.build_ext as _build_ext_lib
+import setuptools.command.build_ext as build_ext_lib
 
 
-def b2_bool(value):
+def b2_bool(value: bool) -> str:
     if value:
         return "on"
     return "off"
@@ -40,30 +48,35 @@ def b2_bool(value):
 
 
 class B2Distribution(setuptools.Distribution):
-    def reinitialize_command(self, command, reinit_subcommands=0):
+    def reinitialize_command(
+        self, command: str, reinit_subcommands: int = 0
+    ) -> distutils.cmd.Command:
         if command == "build_ext":
-            return self.get_command_obj("build_ext")
-        return super().reinitialize_command(
-            command, reinit_subcommands=reinit_subcommands
+            return cast(distutils.cmd.Command, self.get_command_obj("build_ext"))
+        return cast(
+            distutils.cmd.Command,
+            super().reinitialize_command(
+                command, reinit_subcommands=reinit_subcommands
+            ),
         )
 
 
 # Various setuptools logic expects us to provide Extension instances for each
 # extension in the distro.
 class StubExtension(setuptools.Extension):
-    def __init__(self, name):
+    def __init__(self, name: str):
         # An empty sources list ensures the base build_ext command won't build
         # anything
         super().__init__(name, sources=[])
 
 
-def b2_escape(value):
+def b2_escape(value: str) -> str:
     value = value.replace("\\", "\\\\")
     value = value.replace('"', '\\"')
     return f'"{value}"'
 
 
-def write_b2_python_config(config):
+def write_b2_python_config(config: IO[str]) -> None:
     write = config.write
     # b2 keys python environments by X.Y version, breaking ties by matching
     # a property list, called the "condition" of the environment. To ensure
@@ -109,6 +122,7 @@ def write_b2_python_config(config):
     # Note that sysconfig and distutils.sysconfig disagree here, especially on
     # windows.
     ext_suffix = distutils.sysconfig.get_config_var("EXT_SUFFIX")
+    ext_suffix = str(ext_suffix or "")
 
     # python.jam appends the platform-specific final suffix on its own. I can't
     # find a consistent value from sysconfig or distutils.sysconfig for this.
@@ -120,16 +134,13 @@ def write_b2_python_config(config):
     write(" ;\n")
 
 
-BuildExtBase = _build_ext_lib.build_ext
-
-
-class LibtorrentBuildExt(BuildExtBase):
+class LibtorrentBuildExt(build_ext_lib.build_ext):
 
     CONFIG_MODE_DISTUTILS = "distutils"
     CONFIG_MODE_B2 = "b2"
     CONFIG_MODES = (CONFIG_MODE_DISTUTILS, CONFIG_MODE_B2)
 
-    user_options = BuildExtBase.user_options + [
+    user_options = build_ext_lib.build_ext.user_options + [
         (
             "config-mode=",
             None,
@@ -195,28 +206,28 @@ class LibtorrentBuildExt(BuildExtBase):
         ),
     ]
 
-    boolean_options = BuildExtBase.boolean_options + ["pic", "hash"]
+    boolean_options = build_ext_lib.build_ext.boolean_options + ["pic", "hash"]
 
-    def initialize_options(self):
+    def initialize_options(self) -> None:
 
         self.config_mode = self.CONFIG_MODE_DISTUTILS
         self.b2_args = ""
         self.no_autoconf = ""
 
-        self.toolset = None
-        self.libtorrent_link = None
-        self.boost_link = None
-        self.pic = None
-        self.optimization = None
-        self.hash = None
-        self.cxxstd = None
+        self.toolset: Optional[str] = None
+        self.libtorrent_link: Optional[str] = None
+        self.boost_link: Optional[str] = None
+        self.pic: Optional[bool] = None
+        self.optimization: Optional[str] = None
+        self.hash: Optional[bool] = None
+        self.cxxstd: Optional[str] = None
 
-        self._b2_args_split = []
-        self._b2_args_configured = set()
+        self._b2_args_split: List[str] = []
+        self._b2_args_configured: Set[str] = set()
 
-        return super().initialize_options()
+        super().initialize_options()
 
-    def finalize_options(self):
+    def finalize_options(self) -> None:
         super().finalize_options()
 
         if self.config_mode not in self.CONFIG_MODES:
@@ -274,7 +285,7 @@ class LibtorrentBuildExt(BuildExtBase):
             warnings.warn("--cxxstd is deprecated; use --b2-args=cxxstd=...")
             self._maybe_add_arg(f"cxxstd={self.cxxstd}")
 
-    def _should_add_arg(self, arg):
+    def _should_add_arg(self, arg: str) -> bool:
         m = re.match(r"(-\w).*", arg)
         if m:
             name = m.group(1)
@@ -282,18 +293,18 @@ class LibtorrentBuildExt(BuildExtBase):
             name = arg.split("=", 1)[0]
         return name not in self._b2_args_configured
 
-    def _maybe_add_arg(self, arg):
+    def _maybe_add_arg(self, arg: str) -> bool:
         if self._should_add_arg(arg):
             self._b2_args_split.append(arg)
             return True
         return False
 
-    def run(self):
+    def run(self) -> None:
         # The current jamfile layout just supports one extension
         self._build_extension_with_b2()
-        return super().run()
+        super().run()
 
-    def _build_extension_with_b2(self):
+    def _build_extension_with_b2(self) -> None:
         python_binding_dir = pathlib.Path(__file__).parent.absolute()
         with self._configure_b2():
             command = ["b2"] + self._b2_args_split
@@ -301,7 +312,7 @@ class LibtorrentBuildExt(BuildExtBase):
             subprocess.run(command, cwd=python_binding_dir, check=True)
 
     @contextlib.contextmanager
-    def _configure_b2(self):
+    def _configure_b2(self) -> Iterator[None]:
         if self.config_mode == self.CONFIG_MODE_DISTUTILS:
             # If we're using distutils mode, we'll auto-configure a lot of args
             # and write temporary config.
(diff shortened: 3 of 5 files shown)


.github/workflows/linux.yml
```diff
@@ -7,6 +7,26 @@ on:
...
+      - uses: actions/setup-python@v2
+        with:
+           python-version: "3.7"
+      - uses: actions/setup-python@v2
+        with:
+           python-version: "3.8"
+      - uses: actions/setup-python@v2
+        with:
+           python-version: "3.9"
+      - uses: pre-commit/action@v2.0.0
```
arvidn, line 28: why are so many versions of python required for this?

arvidn:
I'm a bit nervous about making large changes like this in `RC_1_2`. I would not look forward to merging them into `RC_2_0`. Making them in `RC_2_0` would make it easier since `master` is very similar and I would expect it to merge cleanly.

arvidn:
In fact, it might make the most sense to put changes like these in `master`. Obviously if you encounter a bug, that should be back-ported, but it doesn't seem urgent to get this into 1.2.x or 2.0.x

.github/workflows/linux.yml
AllSeeingEyeTolledEweSew: There are subtle differences between versions. This is a way to definitively test against all supported python versions. Examples of open source python projects doing this: [youtube-dl](https://github.com/ytdl-org/youtube-dl/blob/master/.github/workflows/ci.yml), [fastapi](https://github.com/tiangolo/fastapi/blob/master/.github/workflows/test.yml), [black](https://github.com/psf/black/blob/master/.github/workflows/test.yml).

Most of the differences between python versions are differences in the standard library, not language features. So this becomes more important as we write more pure python code, rather than just native-code extensions.

Although now that I look more closely at the way open-source projects do this, it looks like they matrix-config the python version. The way I have the mypy pre-config config set up actually requires all python versions to be installed, even when running locally. I'll change that.

AllSeeingEyeTolledEweSew:
> In fact, it might make the most sense to put changes like these in `master`. Obviously if you encounter a bug, that should be back-ported, but it doesn't seem urgent to get this into 1.2.x or 2.0.x

I can rebase on another branch. But I wanted to do a lot of work on `test.py`, and it's going to be extremely helpful to have python linting and formatting active on all branches.

Is there a way I can help with branch merges? Should I do separate PRs on each branch, then when you do regular merges you can skip commits?

.github/workflows/linux.yml
arvidn: the linter itself behaves differently on different versions of python?
Still, it looks like the way you invoke it is to just run the liter once though, with one of the python versions

AllSeeingEyeTolledEweSew: yes. `flake8` and `mypy` generally check the target code against the running environment. You can do some cross-version checking with `mypy`, but it's incomplete (if you want to type-check against a third-party package, you need to install the package in a python environment of the desired version)

The multi-version invocation actually happens in `.pre-commit-config.yml`, where there are multiple entries for `mypy`. Hence you need four python installations even to run pre-commit locally. I will change this.

arvidn:
I think these changes are generally reasonable. I can't help this feeling though, of the benefit having to carry the cost. These are the costs I see:

* This PR adds quite a few new files to the root of the repo. I would like to keep it as tidy as possible. Is there any way to move some of these into a directory? Something like `.pre-commit/` or something like that? Or maybe provide the configurations on the command lines.
* when flake8 was running on the python code on travis, I never found it highlight any actual bugs or issues, it just had opinions about formatting that (a) quite often I disagreed with and (b) could be quite annoying to get right, requiring several round-trips to the CI. I can think of a few ways to mitigate this:
   1. make sure it's easy to just run some pre-commit command to fix all formatting in-place.
   2. make flake8 only cover python files that "matter". Which I think primarily are the ones under `bindings/python/` The other python scripts are simple helpers, essentially just better shell scripts. I imagine that would also cut down on the long exclusions lists that are repeated right now too.
* running linters under multiple versions of python, and growing the number of CI checks that are run seems overkill. It seems unlikely that doing so would provide much value. (but slowing down CI is a big cost).
* isort seems also to add marginal value. I can see it being helpful in a large python project, but this project is mostly not python. It doesn't seem all that important to enforce import order.

The things I definitely see valuable are mypy and the rst linter. All the built-in checkers in pre-commit look good actually.

As you may have figured by now, I'm a bit ignorant when it comes to python. I was under the impression `flake8` was a formatter for python. How is it compatible with `black`?

I think the `LICENSE` file should be excluded as well, as I collect licenses verbatim in there.

arvidn:
I think I may have mixed up this and the other PR. Both seem very similar and both are against `master` now. Is one sitting on top of the other, so I should review them in a specific order?

I think my comment above largely applies to both either way.

One more thing I meant to mention. Once I install pre-commit, will it actually hook up to git to run when I run `git commit`? or is there some command I should run manually? Will it fix-up formatting of python files when I commit for instance?
It might be worth explaining this a bit in the `CONTRIBUTING` file.

One thing I do sometimes is to commit changes on disk, while holding unsaved changes in my editor. That wouldn't work so well if files are changed by committing them.

AllSeeingEyeTolledEweSew:
> This PR adds quite a few new files to the root of the repo. I would like to keep it as tidy as possible. Is there any way to move some of these into a directory? Something like `.pre-commit/` or something like that? Or maybe provide the configurations on the command lines.

If you're referring to having `.flake8` and `.isort.cfg` as separate files, this is intentional to make sure tools run with consistent config. I find it helpful that running `flake8` does the same thing as `pre-commit run flake8`. My goal is to try to minimize the number of "wrong ways" to run a tool, so I'm quite opposed to having tool configuration only found on a CI command line.

There is a "movement" to centralize python tool configuration, into one of `setup.cfg` or `pyproject.toml`. Most of these tools can read config from one of those. I can do this if it's more pleasing.

> when flake8 was running on the python code on travis, I never found it highlight any actual bugs or issues, it just had opinions about formatting that (a) quite often I disagreed with and (b) could be quite annoying to get right, requiring several round-trips to the CI. I can think of a few ways to mitigate this:
>       1. make sure it's easy to just run some pre-commit command to fix all formatting in-place.
>       2. make flake8 only cover python files that "matter". Which I think primarily are the ones under `bindings/python/` The other python scripts are simple helpers, essentially just better shell scripts. I imagine that would also cut down on the long exclusions lists that are repeated right now too.

See below for some discussion about what the different tools do, and the different options for this PR and down the road.

> running linters under multiple versions of python, and growing the number of CI checks that are run seems overkill. It seems unlikely that doing so would provide much value. (but slowing down CI is a big cost).

I definitely appreciate optimizing CI time, but right now the `pre-commit` jobs run in under a minute. I think long-running tasks aren't appropriate for `pre-commit`, so it should always run fast.

> isort seems also to add marginal value. I can see it being helpful in a large python project, but this project is mostly not python. It doesn't seem all that important to enforce import order.

> As you may have figured by now, I'm a bit ignorant when it comes to python. I was under the impression `flake8` was a formatter for python. How is it compatible with `black`?

`black`, `isort` and `autoflake` are auto-formatters. `black`'s design goal is like `gofmt`: it automatically applies one authoritative format, basically chosen to not be too offensive to existing format philosophies. I add `isort` and `autoflake` to help with the same goal. In practice after adopting these changes, I find I don't *think* about formatting as much as I used to. I don't need to muck with editor settings almost at all when writing python, diffs are much cleaner when they don't have trivial formatting changes (`isort` is actually crucial for this), and of course code is more readable when formatting is consistent. I assume that with the amount of git merging you do, cleaner diffs are a big win.

`flake8` and `mypy` are linters; they just read code and print errors. `flake8` complains about some formatting errors, but `black` is almost 100% compatible with `flake8` (the one exception is noted in the `.flake8` config). `flake8` also has a robust plugin ecosystem of checkers that go beyond formatting. [flake8-bugbear](https://github.com/PyCQA/flake8-bugbear) has some good ones.

`mypy` is a type-checking linter. It's very advanced, but very complex. In practice I find `mypy` catches many more bugs than `flake8`, so much so that I normally require type annotations in all my code. I have in mind a project to add type annotations to the libtorrent extension as well, but I'll do this after I update the python tests.

Aside from the big `exclude:` lists in `.pre-commit-config.yaml`, we can also make linters skip files with directive comments. If you put `# type: ignore` at the top of a file then `mypy` will ignore it. Similarly `# flake8: noqa` will make `flake8` ignore it. I don't think that any of the autoformatters are sensitive to directives though.

As I mentioned, I set up most of these tools in anticipation of some big changes to the python test suite. I don't think e.g. type checking is important in helper scripts, but I really do think autoformatting and linting are helpful here, *if* they achieve the stated goal of "they help me think less about formatting".

I think a good end state is where *all* python code is auto-formatted and linted, except that helper scripts do not require type annotations. What do you think about this? I didn't do it here, because auto-formatting the world would bloat the PR so much, and it would be kind of senseless to enable linting without autoformatting. I thought it was simpler to *just* apply this final scheme to `setup.py` as an example and leave the rest in exclusion lists for now, but maybe I was wrong.

> I think the `LICENSE` file should be excluded as well, as I collect licenses verbatim in there.

I'll add that.

> I think I may have mixed up this and the other PR. Both seem very similar and both are against `master` now. Is one sitting on top of the other, so I should review them in a specific order?

I created a few PRs based on each other. I wanted to have the python formatting and linting in place before making more changes to python code. I had hoped to separate them so you could review `pre-commit`, and my `pre-commit` python config and example changes that use that config, as separate change sets.

I haven't found a brilliant way to stack dependent PRs like this in github. As a reviewer you can manually select a range of commits to view in the diff, but as the author I'd like to be able to tell github to diff relative to a PR, not relative to the base branch. Let me know if I can do this in a nicer way somehow.

> One more thing I meant to mention. Once I install pre-commit, will it actually hook up to git to run when I run `git commit`? or is there some command I should run manually? Will it fix-up formatting of python files when I commit for instance?
> It might be worth explaining this a bit in the `CONTRIBUTING` file.

Running `pre-commit install`  in a repo will install it as a git pre-commit hook for that repo. So whenever you commit, git will run `pre-commit`. If any auto-formatter makes a change, it *does* modify the file on disk (it becomes an unstaged change), *and* counts as a failure of the hook, so the commit fails. When writing python it's helpful to just run `pre-commit` after saving and before committing.

> One thing I do sometimes is to commit changes on disk, while holding unsaved changes in my editor. That wouldn't work so well if files are changed by committing them.

I don't know what to tell you about that. If you're modifying a file that's under management by an autoformatter, you should probably save, then run the formatter, then reload. You can run `pre-commit` at any time without having to commit. You might also be able to get various editors to automatically apply `black`'s code style so it wouldn't make changes. Personally I'm very happy to forget about my editor's format settings, write sloppily-formatted code, and have autoformatters do the work for me.

arvidn:
> If you're referring to having .flake8 and .isort.cfg as separate files, this is intentional to make sure tools run with consistent config.

Would it work if those files are put in `bindings/python/`?

> I definitely appreciate optimizing CI time, but right now the pre-commit jobs run in under a minute. I think long-running tasks aren't appropriate for pre-commit, so it should always run fast.

The total runtime of the new jobs is 363 seconds, granted, not a huge amount. I would still prefer a single virtual machine, which would run in about 90 seconds.

I'm still surprised that you want to run the tools under different versions of python. Are you sure they behave differently under different versions? And is it deliberate? It seems really strange. You would expect at least you could pick the behavior you want with a command line switch, rather than starting 4 different virtual machines with different versions of python just to get a certain check run by the linter.

Additionally, I question the marginal value of those additional checks you get by running more than one version, but I don't know what they are. I assume the majority of checks are identical for all versions of python. Is it documented anywhere which checks are run for which version of python?

> black, isort and autoflake are auto-formatters.

If they always agree, why do you need more than one? I suppose perhaps isort doesn't overlap with the others, as it seems only concerned about import order, and perhaps the other auto-formatters don't care about that. But there's no conflicts between autoflake and black? If there is, the order you run them matters, as the last one will be the authoritative.

Can you explain why you would need more than one?

setup.py
```diff
@@ -2,5 +2,5 @@
 
 import os
 
-os.chdir('bindings/python')
-exec(open('setup.py').read())
+os.chdir("bindings/python")
+exec(open("setup.py").read())
```
arvidn, line 6: unrelated, but this turned out to be problematic btw. see https://github.com/arvidn/libtorrent/commit/e7e5805c56873ddfdcc09c407251185234caafbc#r46496265

AllSeeingEyeTolledEweSew: I think your fix there looks reasonable enough, but in reality I think `setup.py` executing a different file is a brittle approach. I wanted to experiment with moving all of `bindings/python/setup.py` into the root `setup.py`. I think it's less brittle but haven't experimented with this yet so I don't know the tradeoffs.

AllSeeingEyeTolledEweSew:
> Would it work if those files are put in `bindings/python/`?

Sure, but then the tools need to always be invoked with `--config=./bindings/python/.foo`. This hurts my goal of minimizing "wrong" invocation patterns.

I also don't think it's appropriate, as it's a surprising requirement if we want to apply autoformatters on files outside of `bindings/python`.

Generally we're working against the grain when keeping such configuration files outside of root (including `setup.py`, as I mentioned inline).

Am I missing something about why the dotfiles are a burden? Is it that they create more clutter in windows?

> I'm still surprised that you want to run the tools under different versions of python. Are you sure they behave differently under different versions? And is it deliberate? It seems really strange. You would expect at least you could pick the behavior you want with a command line switch, rather than starting 4 different virtual machines with different versions of python just to get a certain check run by the linter.
>
> Additionally, I question the marginal value of those additional checks you get by running more than one version, but I don't know what they are. I assume the majority of checks are identical for all versions of python. Is it documented anywhere which checks are run for which version of python?

I believe `flake8` is the long pole here. It only supports linting on the python version where it's installed. At a high level I'd say it's an outcome of the pattern that python tools often only work with the python version in which they're installed and invoked, but there's good reason for this pattern.

[This search](https://github.com/python/typeshed/search?q=sys.version_info&type=) shows all the places where the official type stubs are sensitive to `sys.version_info`. You can also search for `sys.platform`. The TL;DR is that there are many subtle differences in the standard libraries across versions and platforms.

So for certain checks, any linter tool has to either maintain a reference for all python versions, or introspect installed libraries; the latter means you need a full python installation in the target version.

`mypy` actually does the first one. It vendors the official type stubs, and it can be configured to check variations of `sys.version_info` or `sys.platform` with a command line switch. This has a lot of downsides; I've been stung many times because `mypy`'s reference doesn't always match reality, and its configuration switch support is narrow (it doesn't understand that `os.name` varies with `sys.platform`). Up until recently, `mypy` had a hard requirement to work this way. The standard library hasn't included inline type annotations, because it would've been necessary to maintain branches with and without annotations while python 3.5 was actively supported, and this would've been onerous. I strongly suspect in the future (years) the standard library will adopt more inline annotations, and type checkers will introspect all type information from the installed environment, just like `flake8` does today.

Similar to what I mentioned in #5905, I think there are two strategies: either to test on all python versions or test on the oldest-supported one and try to test more aggressively for use of deprecated features.

I protest that this is a potential sacrifice in correctness, that doesn't balance the gain in CI time. The new 363 seconds are parallelized on GH actions. We can also use `action/cache` to save bootstrapping time for the `pre-commit` environments, which is the majority of this runtime. I do concede there's not *yet* enough python code to fight hard about this. If we want to only enable python 3.6 for now, I will likely come back to fight for enabling more versions once I've written more python code.

> If they always agree, why do you need more than one? I suppose perhaps isort doesn't overlap with the others, as it seems only concerned about import order, and perhaps the other auto-formatters don't care about that. But there's no conflicts between autoflake and black? If there is, the order you run them matters, as the last one will be the authoritative.
> 
> Can you explain why you would need more than one?

We use more than one because they do different things.

`black` fixes whitespace and newlines to a standard format. I think it does some trailing-comma manipulation in some cases. It doesn't make any semantic code changes.

`isort` reorders and groups imports. It's a much-requested feature of `black`, but hasn't been done yet. This is kind of like sorting `#include` lines: imports *can* have side effects so reordering is technically a semantic change, but this is very rare, and `isort` can be configured to be aware of it anyway.

In practice `isort` is really important in the face of type annotations. You end up importing a lot of names from the `typing` module: `from typing import List`, etc. This can become a mess, especially across multiple authors using different styles, like `from typing import List` vs `from typing import (List, Dict)` in the same file. `isort` can vastly clean up diffs here.

`autoflake` just removes a few cases of unused code. In practice this is mainly helpful for removing unused imports, which dovetails well with `isort`, again especially in the face of type annotations.

As `black` lets me not think about spacing or line breaks or configuring my editor, `isort`/`autoflake` let me not think about imports. I just write new import lines as I need them somewhere around the top of the file, and the tools give me a clean diff.

I really want to emphasize that I picked this configuration because I *don't* have strong opinions about code style and formatting in python. I'm willing to sacrifice the opinions I *do* have in favor of automation and consistency. I think *consistent* formatting leads to readability and clean diffs, but I don't care about particulars. The goal of this configuration is to maximize automation, consistency, and support from tools. The combination of `black`, `isort` and `autoflake` is pretty widely used among open source python projects that have this same goal.

It's not impossible that the autoformatters can conflict with each other, but I haven't seen it with this configuration yet. I run `black` first because it has the most development effort in it, and its syntax analyzer seems the most advanced, so I figure its standardization gives the best chance for `isort` and `autoflake` to do consistent things. My `isort` config is targeted to break up long lines and so shouldn't conflict with `black`. I suppose there are some cases where `autoflake` removing a line could create a double-newline that would be fixed by earlier tools. We could guard against this by configuring `pre-commit` to run the whole tool sequence multiple times, but I've never seen anyone else do this. I would probably wait until it actually happens before I worry about it.

Trying to capture open issues:
* Where should config live? (stylistic choice)
  * `.flake8`/`.isort.cfg`
  * `setup.cfg` and `pyproject.toml` (I propose this)
  * Somewhere under `bindings/python`
  * On tool command lines
* Should we run python 3.6 only or multiple versions? (I propose 3.6 now, I'll enable more versions later)
* What are the target standards for auto-formatting and linting? Current proposal:
  * Auto-format and lint all python code, including helper tools
  * Type annotations are required everywhere *except* helper tools

AllSeeingEyeTolledEweSew:
You've mentioned a few times that libtorrent doesn't have much python code or many python commits.

In the first place, I just don't think that's true:

```
$ wc -l $(find . -name '*.py' | grep -v /deps/)
   438 ./examples/run_benchmarks.py
    24 ./fuzzers/tools/unify_corpus_names.py
   214 ./fuzzers/tools/generate_initial_corpus.py
   190 ./test/web_server.py
   317 ./test/socks.py
   550 ./test/http_proxy.py
    57 ./test/websocket_server.py
  1562 ./docs/gen_reference_doc.py
    46 ./docs/filter-rst.py
   136 ./docs/gen_settings_doc.py
   133 ./docs/gen_stats_doc.py
   166 ./docs/gen_todo.py
     6 ./setup.py
   650 ./tools/parse_session_stats.py
    88 ./tools/clean.py
    64 ./tools/parse_dht_stats.py
    58 ./tools/update_copyright.py
   162 ./tools/parse_sample.py
    75 ./tools/parse_peer_log.py
   137 ./tools/copyright.py
   122 ./tools/run_benchmark.py
    80 ./tools/set_version.py
    89 ./tools/benchmark_checking.py
   131 ./tools/parse_lookup_log.py
   341 ./tools/parse_dht_log.py
   140 ./tools/gen_fwd.py
    55 ./tools/parse_dht_rtt.py
    78 ./tools/dht_flood.py
   417 ./tools/parse_utp_log.py
   109 ./bindings/c/tools/gen_header.py
    31 ./bindings/c/tools/gen_alert_header.py
   381 ./bindings/python/client.py
    34 ./bindings/python/simple_client.py
    34 ./bindings/python/dummy_data.py
   269 ./bindings/python/setup.py
   836 ./bindings/python/test.py
    61 ./bindings/python/make_torrent.py
  8281 total
```

In the second place, I claim the test coverage on the python bindings is really lacking, so there's much more code to write, now and in the future.

I really think these changes will make a lot more sense once you see what I'm doing with the python tests.

arvidn:
> Sure, but then the tools need to always be invoked with --config=./bindings/python/.foo. This hurts my goal of minimizing "wrong" invocation patterns.

Yeah, I agree. I think in my mind, these are kind of separate projects, libtorrent and the python bindings. The main reason to keep them in the same repository is that changes can be made in lock-step, to ensure the bindings always build and work. In separate repos that would be more complicated and they would likely diverge at times.

Perhaps the bindings should have in its own repo (with its own python-oriented dotfiles) and it's pulled in as a submodule. It would still make it more complicated to make changes that affect both. One benefit would be that CI would be more focused and probably run quicker, since it would only have to cover a smaller part of the code.

> Am I missing something about why the dotfiles are a burden? Is it that they create more clutter in windows?

They create clutter generally, on perhaps primarily on github.

> I protest that this is a potential sacrifice in correctness, that doesn't balance the gain in CI time. The new 363 seconds are parallelized on GH actions.

GH actions only run so many actions in parallel, and that limit has already been hit by the existing checks. So every one of those 363 seconds eat into the total number of CI seconds available. It may not be 363 wall-clock seconds though.

> We can also use action/cache to save bootstrapping time for the pre-commit environments, which is the majority of this runtime. I do concede there's not yet enough python code to fight hard about this. If we want to only enable python 3.6 for now, I will likely come back to fight for enabling more versions once I've written more python code.

yeah, let's start with baby steps.

> I really want to emphasize that I picked this configuration because I don't have strong opinions about code style and formatting in python. I'm willing to sacrifice the opinions I do have in favor of automation and consistency.

I don't have a problem with auto formatters. What I have had problems with in the past is very particular rules that I have to manually apply (e.g. vertical alignment). Perhaps there's a way to ask flake8 to change files in-place so I don't have to apply them manually, but that doesn't really matter, because I don't know about that switch, and I can't really be bothered because 99% of the time I don't change any python code. I will have forgotten about it until the next time anyway.

So, as long is it's obvious what to do when the pre-commit formatting check fails in CI, I'm fine with it (assuming it's easy to address it). It sounds like it is in this case, since it's installed as a git commit hook.

> setup.cfg and pyproject.toml (I propose this)

Sounds good.

> Should we run python 3.6 only or multiple versions? (I propose 3.6 now, I'll enable more versions later)

Sounds good.

> What are the target standards for auto-formatting and linting? Current proposal:
>     Auto-format and lint all python code, including helper tools
>     Type annotations are required everywhere except helper tools

Type annotations are the ones I personally see the most value in and would prefer be checked everywhere. But when you say "require" type annotations, you mean a plain old untyped python function would be disallowed? I imagine code would have to be transitioned first, right? I don't think any python code would pass such check.

Sounds good though.

arvidn:
> 8281 total

Which still is a little bit over 4% of all the code, 91% being C++.

https://www.openhub.net/p/rasterbar-libtorrent/analyses/latest/languages_summary

I think the ratio of actual *commits* modifying C++ vs. python is an even starker contrast. At least until recently with your contributions!

> In the second place, I claim the test coverage on the python bindings is really lacking, so there's much more code to write, now and in the future.

I definitely agree with that!

AllSeeingEyeTolledEweSew:
> Yeah, I agree. I think in my mind, these are kind of separate projects, libtorrent and the python bindings. The main reason to keep them in the same repository is that changes can be made in lock-step, to ensure the bindings always build and work. In separate repos that would be more complicated and they would likely diverge at times.
> 
> Perhaps the bindings should have in its own repo (with its own python-oriented dotfiles) and it's pulled in as a submodule. It would still make it more complicated to make changes that affect both. One benefit would be that CI would be more focused and probably run quicker, since it would only have to cover a smaller part of the code.

Personally I'm not too motivated to strip the python bindings out to a separate project. The standard git commands are always tripping on submodules; I've never got the hang of working with them.
 
> I don't have a problem with auto formatters. What I have had problems with in the past is very particular rules that I have to manually apply (e.g. vertical alignment). Perhaps there's a way to ask flake8 to change files in-place so I don't have to apply them manually, but that doesn't really matter, because I don't know about that switch, and I can't really be bothered because 99% of the time I don't change any python code. I will have forgotten about it until the next time anyway.
> 
> So, as long is it's obvious what to do when the pre-commit formatting check fails in CI, I'm fine with it (assuming it's easy to address it). It sounds like it is in this case, since it's installed as a git commit hook.

With the config as written, `flake8` *should* never complain about formatting (whitespace, line breaks or parentheses), as the autoformatters *should* take care of it all.

If `flake8` does complain about formatting after autoformatters have run, *and* there's no way to remedy this via config, *then* I think that's a good reason to just permanently ignore that `flake8` error (this is already done for `E203`).

`flake8` and `mypy` may complain about stuff other than formatting, like syntax errors or type mismatches, but there's no quick fix for such complaints.

> Type annotations are the ones I personally see the most value in and would prefer be checked everywhere. But when you say "require" type annotations, you mean a plain old untyped python function would be disallowed? I imagine code would have to be transitioned first, right? I don't think any python code would pass such check.

Ah, I didn't explain this.

`mypy`'s will type check functions *only if* they have type annotations. So here
```python
def f(a: str, b: int) -> None: ...

def g(a): ...
```
`f` would get type-checked, and `g` would be ignored, *even if* `g` calls `f`, or calls another function with type annotations.

This behavior is intended to make it easier to adopt type checking incrementally.

If `g` is *called from* a type-checked function, `g`'s return type is treated as `Any`. This matches any type, *but* it means that this will fail type checking:
```python
def f(a: str, b: int) -> None:
    f(g(a), g(b))

def g(a): ...
```

because `g` returns `Any`, which doesn't match `int` or `str`.

So typically, type annotations must be adopted on innermost functions first, and outermost functions last.

(you can add external type annotations, even to code you don't own, via type stub files. That's a story for another time)

I assumed you wouldn't want to write type annotations in every one of your helper scripts. So eventually (later PR) we would configure `mypy` to require type annotations on all new code in `bindings/python`, but not elsewhere. So type checks would be honored in helper scripts, but not enforced.

However I strongly support if you want to enforce type annotations everywhere, I think they're great.

arvidn:
I imagine you'll rebase this, right?

AllSeeingEyeTolledEweSew:
> I imagine you'll rebase this, right?

Yes, I suppose I need to rebase every time one of the "dependent" PRs is merged. It's a fine tradeoff for me to be able to continue working with dependent work already in place. I wish I could do something less confusing for reviewers.

arvidn:
Hey, I just ran into this, while switching between various branches:

```
No .pre-commit-config.yaml file was found
- To temporarily silence this, run `PRE_COMMIT_ALLOW_NO_CONFIG=1 git ...`
- To permanently silence this, install pre-commit with the --allow-missing-config option
- To uninstall pre-commit run `pre-commit uninstall`
```

I imagine it would be reasonable to use the `--allow-missing-config` option when installing. Specifically to support older branches. Is that right?

AllSeeingEyeTolledEweSew:
> Hey, I just ran into this, while switching between various branches:
> 
> ```
> No .pre-commit-config.yaml file was found
> - To temporarily silence this, run `PRE_COMMIT_ALLOW_NO_CONFIG=1 git ...`
> - To permanently silence this, install pre-commit with the --allow-missing-config option
> - To uninstall pre-commit run `pre-commit uninstall`
> ```
> 
> I imagine it would be reasonable to use the `--allow-missing-config` option when installing. Specifically to support older branches. Is that right?

Ahh, that seems right.

arvidn:
is `mypy.ini` relevant?

AllSeeingEyeTolledEweSew:
Great catch! It should've been deleted.

arvidn:
With the addition of `pre-commit` I'm struggling with `flake8`. As far as I can tell, there is no way to have it fix the formatting, it has to be done by hand.

I'd rather have random formatting than having to fix it my hand. Is `autoflake` a tool to fix it? If so, is `flake8` necessary?
I'd prefer there to not be a risk that `flake8` disagrees with `autoflake` and something has to be fixed by hand

AllSeeingEyeTolledEweSew:
`black` fixes the formatting.

`black` and `flake8` *should* be mutually configured such that after `black` runs, `flake8` never complains about formatting.

(`flake8` may complain about other things, like referencing names that don't exist).

If you're doing `pre-commit run` and `flake8` has formatting complaints, then maybe I got the config wrong. Can you show me an example of what's going on?

AllSeeingEyeTolledEweSew:
I think I just tripped on the problem while updating #5948. Let me take a look...

AllSeeingEyeTolledEweSew:
So, if you're referring to errors like https://github.com/arvidn/libtorrent/pull/5948/checks?check_run_id=1847492176 , it looks like `gen_convenience_header.py` got added to the repo and pushed without running `pre-commit` and without a PR (so CI didn't run), unless I missed something.

In that, note that `flake8` isn't the one complaining. The `black` hook is failing, because it fails whenever its auto-formatting has to make a change. `pre-commit` then renders a diff of what was changed.

arvidn:
I was merging RC_2_0 into master in a repo where I had not installed pre-commit. But I always run my merges through CI, which caught it. I installed pre-commit to run it locally, but it didn't fix the issue. IIRC, one of them was "visual indent" being wrong. Black did not correct that.

arvidn:
So maybe I wasn't running black with the correct config then, since I was running it current master

AllSeeingEyeTolledEweSew:
*This* PR is the one that *adds* `black`. I got confused previously and said wrong things.

The commit that got merged was the one that added `flake8` to master. Right now master has `flake8` but no `black`.

That happened mainly because I originally authored the PR chain against `RC_1_2` which still runs `flake8` in appveyor, so it was just a lateral move and improvement. When I rebased to `master` I left `flake8` in part because I didn't expect too many *new* python scripts to be added between that PR and this one.

TL;DR we should merge this PR, it will make our lives better.

arvidn:
Now that the changes to `setup.py` have bubbled up from `RC_1_2` to `master`, I rebased this patch, here: https://github.com/arvidn/libtorrent/pull/5966

Feel free to rebase this PR or comment on my rebase.

AllSeeingEyeTolledEweSew:
I rebased this one

AllSeeingEyeTolledEweSew:
I actually had some more fixes, this is now rebased on #5970

AllSeeingEyeTolledEweSew:
Closing in favor of #5966

PR #5905, opened 2021-01-24
[6/?] Rework python include/library detection. Use tox for running tests

AllSeeingEyeTolledEweSew:
[tox](https://tox.readthedocs.io/en/latest/) is designed to run tests against a python project. It's widely used in the python world. It does:
* packaging the project
* installing the project in a virtual environment
* invoking a set of tests

This allows you to just run `tox` in an environment, and the right thing will happen to run tests against the project. This should simplify CI config, as well as simplify the process of running tests locally.

This patch ended up making some changes to the way we manage b2's python configuration.

It's based on #5902.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -11,18 +11,43 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     strategy:
+      # on linux and mac, the boost libraries are fixed to a python version.
+      # since we install boost from repos on those systems, we must use the
+      # matching python versions
       matrix:
-        os: [ubuntu-20.04, macos-latest, windows-2019 ]
+        include:
+          - os: ubuntu-20.04
+            python-version: 3.8
+          - os: macos-latest
+            python-version: 3.9
+          - os: windows-2019
+            python-version: 3.6
 
     steps:
     - uses: actions/checkout@v2
       with:
         submodules: true
 
+    # Note:
+    #  - on mac and linux images, 'python' is python2 and 'python3' is
+    #    python3
+    #  - on windows, neither 'python' nor 'python3' is in PATH by default
+    #  - setup-python sets up PATH so 'python' and 'python3' point to the
+    #    requested version on mac and linux, but on windows it only sets up
+    #    'python'.
+    - uses: actions/setup-python@v2
+      with:
+         python-version: ${{ matrix.python-version }}
+
+    - name: install tox
+      run: |
+        python3 -m pip install --upgrade pip
+        python3 -m pip install --upgrade tox
+
     - name: dependencies (MacOS)
       if: runner.os == 'macOS'
       run: |
-        brew install boost-build boost boost-python3 python@3.9
+        brew install boost-build boost boost-python3
 
     - name: update package lists (linux)
       if: runner.os == 'Linux'
@@ -33,7 +58,7 @@ jobs:
     - name: dependencies (linux)
       if: runner.os == 'Linux'
       run: |
-        sudo apt install libboost-tools-dev libboost-python-dev libboost-dev libboost-system-dev python3 python3-setuptools
+        sudo apt install libboost-tools-dev libboost-python-dev libboost-dev libboost-system-dev
 
     # there appears to be a bug in boost-build where referring to a
     # dependency that's on a different drive fails, deep in boost-build
@@ -42,40 +67,16 @@ jobs:
       if: runner.os == 'Windows'
       shell: cmd
       run: |
+        echo on
         echo using msvc ; >%HOMEDRIVE%%HOMEPATH%\user-config.jam
         mklink /J boost %BOOST_ROOT_1_72_0%
         cd boost
         b2 headers
+        cd ..
+        echo BOOST_ROOT=%CD%\boost>>%GITHUB_ENV%
+        echo BOOST_BUILD_PATH=%CD%\boost\tools\build>>%GITHUB_ENV%
+        echo %CD%\boost>>%GITHUB_PATH%
 
-    - name: build/install (windows)
-      if: runner.os == 'Windows'
-      shell: cmd
+    - name: build/test with tox
       run: |
-        set BOOST_ROOT=%CD%\boost
-        set BOOST_BUILD_PATH=%BOOST_ROOT%\tools\build
-        set PATH=%BOOST_ROOT%;%PATH%
-
-        cd bindings\python
-        python setup.py build_ext install --user --prefix=
-
-    - name: tests (windows)
-      if: runner.os == 'Windows'
-      shell: cmd
-      run: |
-        cd bindings\python
-        python test.py
-
-    - name: build/install
-      if: runner.os != 'Windows'
-      run: |
-        cd bindings/python
-        # Homebrew's python "framework" sets a prefix via distutils config.
-        # --prefix conflicts with --user, so null out prefix so we have one
-        # command that works everywhere
-        python3 setup.py build_ext --b2-args="libtorrent-link=static" install --user --prefix=
-
-    - name: tests
-      if: runner.os != 'Windows'
-      run: |
-        cd bindings/python
-        python3 test.py
+        tox -e py
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -6,6 +6,7 @@ import distutils.cmd
 import distutils.debug
 import distutils.errors
 import distutils.sysconfig
+import functools
 import os
 import pathlib
 import re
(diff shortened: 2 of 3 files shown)


arvidn:
I don't think it's a good idea to make major changes to `RC_1_2`.
You removed a few github action workflows, I assume that was by mistake.
You still need to test `setup.py` in CI, right?

arvidn:
It probably makes sense to pull out the part that extends the existing CI to test the bindings with more versions of python. i.e. extending matrix. It's possible there are existing problems with some python versions on windows

AllSeeingEyeTolledEweSew:
tl;dr it's still a draft :)

> I don't think it's a good idea to make major changes to `RC_1_2`.

Currently like 70% of the changes are just debugging.

tox whitelists environment variables by default. I'm having trouble figuring out how to get boost's `msvc.jam` to find `msvc-14.2` predictably, and can't easily add debug statements to it, and can't easily reproduce the tools cache layout in a local environment, so I'm iterating a lot on CI via force push.

I'm confident that when I'm done it'll be a simple change that merges between branches easily. If I end up being wrong I'll rebase then. I think this one _should_ be on all branches if for no other reason than to trim one line from appveyor...

> You removed a few github action workflows, I assume that was by mistake.
> You still need to test `setup.py` in CI, right?

I removed all the other actions purely to iterate faster. There are enough actions now, and they take long enough, that the work queue would stay full for a while by the time I push the next commit.

> It probably makes sense to pull out the part that extends the existing CI to test the bindings with more versions of python. i.e. extending matrix. It's possible there are existing problems with some python versions on windows

I wanted to start off with the matrix just representing what works with the current testing strategy. I don't intend to get any _new_ matrix entries working in this PR.

I believe all python versions actually work on windows; I did get them to all successfully build and test once. The trouble is `setup.py` doesn't work right now _when run inside a python virtualenv_ on windows.

arvidn:
> The trouble is setup.py doesn't work right now when run inside a python virtualenv on windows.

That seems like its own PR. Do you have a sense of why? Do virtual environments typically come with headers and libraries to link against?

AllSeeingEyeTolledEweSew:
> > The trouble is setup.py doesn't work right now when run inside a python virtualenv on windows.
> 
> That seems like its own PR. Do you have a sense of why? Do virtual environments typically come with headers and libraries to link against?

It looks like `python.jam`'s automagic doesn't account for https://bugs.python.org/issue16116. `distutils` was patched to account for it.

I agree this is a separate issue than using tox. But a separate PR would have to run `setup.py` in a virtualenv on CI to test the fix anyway, and we may as well use tox to do this.

AllSeeingEyeTolledEweSew:
You were right; I had to make major changes to get `setup.py` to work reliably in virtual environments especially on windows, hence I rebased to `master`.

`python.jam`'s include/library search path autodetection never did the right thing in virtual environments. Previously my approach had been to supplement its autodetection with a little of our own, but I got fed up and changed the approach to aggressively override its autodetection, in favor of `distutils`'s autodetection.

I added a pretty aggressive matrix of python installation types to CI. It's probably overkill to keep this going forward, and we could just test with `tox`, but I wanted to see what you thought.

...

Stepping back, I'm going to say `b2`'s python support sucks. It can build python extensions, but this is useless without packaging and testing, and `b2` makes this hard.

`b2` has clearly has never been tested in virtual environments and really lacks design support for them (hence our `<libtorrent-python>on` hack). It treats python like `gcc`, as a tool that's installed globally by a supervisor in one or more versions, but that's not how python works.

Python virtual environments are not just convenient hacks for users who lack root. `pyenv` is essential for using multiple python versions on most linux distros. Build and test isolation in virtual environments with `tox` or similar has been the standard for years. Today the standard is to run apps (including `pre-commit`'s hooks) in stabilized virtual environments with `pipenv`.

I'm frustrated that `b2` is incompatible with a huge chunk of the python ecosystem, provides nothing to replace its functionality, and isn't well-designed enough that we can just extend the functionality we need (unless you want to fork `python.jam`).

I'm also frustrated that I can't find any resources about these issues. `b2` doesn't seem widely used enough that someone else has solved these problems; we're on our own.

AllSeeingEyeTolledEweSew:
Could you especially take a look at my approach of passing `b2 library-path=x`? Do you think this is brittle?

AllSeeingEyeTolledEweSew:
Again, I'm assuming these CI failures are not due to my changes.

AllSeeingEyeTolledEweSew:
I rebased this on #5902, because I want to continue to work on `test.py` with all these changes in place.

.github/workflows/python.yml
```diff
@@ -42,40 +103,25 @@ jobs:
...
       run: |
+        echo on
         echo using msvc ; >%HOMEDRIVE%%HOMEPATH%\user-config.jam
         mklink /J boost %BOOST_ROOT_1_72_0%
         cd boost
         b2 headers
+        cd ..
+        echo BOOST_ROOT=%CD%\boost>>%GITHUB_ENV%
+        echo BOOST_BUILD_PATH=%CD%\boost\tools\build>>%GITHUB_ENV%
+        echo %CD%\boost>>%GITHUB_PATH%
```
arvidn, line 114: what's this for?

.github/workflows/python.yml
```diff
@@ -11,18 +11,79 @@ jobs:
...
+          python-env: global
+        - os: macos-latest
+          python-version: 3.9
+          python-env: tox
+        - os: macos-latest
+          python-version: 3.9
+          python-env: global
+        - os: macos-latest
+          python-version: 3.9
+          python-env: homebrew
```
arvidn, line 37: is it important to build for both python 3.9 and 3.8? it seems it bloats this workflow file quite a lot, and I imagine it'll make CI take longer too.

I wonder the same thing about `tox`. What's the benefit of using `tox` here?
I get the impression a lot of it boils down to making sure it works to install the python module in a virtual environment. That could be done by *just* testing `setup.py` in one, right? If it works in a venv, presumably it works without one as well.

.github/workflows/python.yml
AllSeeingEyeTolledEweSew: This is github actions magic to persist the environment variables and `PATH` to the next step. Otherwise the only way to do it seems to be to do `set BOOST_ROOT=...` which is only effective for that step, so we must duplicate all later commands in the step.

.github/workflows/python.yml
AllSeeingEyeTolledEweSew: In practice, it is probably okay to just run tests against python 3.6. It's possible for language features to change between actively-supported python versions, and testing against all of them is the brute-force way to make sure we don't get stung. The other strategy is to test against the oldest version and maintain discipline to be aware of deprecations. We can tweak the tests and linters to help with this discipline, but it is less foolproof.

As much as I think reducing overall CI time is crucial, I think there are much bigger targets. Getting rid of appveyor is probably the most important. It might also be worth it to build a `setup-boost` github action analogous to `setup-python`, so we can run more piecemeal jobs at less cost, using consistent boost versions.

It's also interesting to note that the `cibuildwheel` process (as in #4218) does separate builds on all combinations of platform * python version, which would be even more combinations than I've listed here (we invoke `cibuildwheel` from one arbitrary python version, and it dispatches other jobs). We should probably dig into this process and make sure there's not too much waste (e.g. the artifacts of the main libtorrent library should probably be reused on a given platform), or else don't run `cibuildwheel` on every PR.

The benefits of `tox` are
* being a python-based test script, so test invocation is more portable
* isolation and reproducibility of the test environment
* running python packaging as part of the test script, to expose packaging issues
* a well-supported ecosystem to do basic stuff with minimal overhead
* a pluggable environment to do fancy stuff (tox testenvs can be run in docker, to make the build process even more isolated and reproducible)

I've never seen another python project with venv-specific issues. Even python extensions normally just use the `distutils` build system so they work fine in venvs. It was a real surprise to find this setup exposed so many issues with `b2`.

What do you think about: python 3.6 only + tox only? (No testing against global python installations)

arvidn: > I've never seen another python project with venv-specific issues. Even python extensions normally just use the distutils build system so they work fine in venvs. It was a real surprise to find this setup exposed so many issues with b2.

Yeah, `b2` is primarily a build system for C++, so I'm not too surprised its python support isn't particularly sophisticated (similar to how distutils C++ support isn't particularly sophisticated). I haven't taken a deep look at this yet, I will see if I can come up with any simplifications.

> What do you think about: python 3.6 only + tox only? (No testing against global python installations)

Sounds reasonable to me.

AllSeeingEyeTolledEweSew:
I rebased this, but it actually requires #5969 and #5970 to work. Once those land, I'll rebase, and I claim the CI will then succeed.

AllSeeingEyeTolledEweSew:
Should be good to go now.

arvidn (approved):

PR #5918, opened 2021-01-27
add -mbig-obj compiler flag to GCC on windows

arvidn:


latot:
Hihi, is working! thx :D

PR #5930, opened 2021-01-30
remove strview_less and just use standard maps

arvidn:


arvidn:
I remember now why I used `boost.container`, it's guaranteed to be nothrow move constructible

PR #5934, opened 2021-02-01
increase the default setting for max_allowed_in_request_queue

arvidn:
from 500 to 2000. The entries in this queue are relatively small, and allowing too few may negatively impact high-throughput peers


xavier2k6:
@arvidn Is there a reason that new setting of `2000` wasn't backported in to `RC_1_2`?

https://github.com/arvidn/libtorrent/blob/c5ff6c3186a92ddec01f6f0a8146aaedb4a1c3f9/src/settings_pack.cpp#L227

PR #5935, opened 2021-02-01
remove misleading (and redundant) peer log message

arvidn:


aldenml:
should similar log a few lines above (inside `if (!t->valid_metadata())`) also be removed?

arvidn:
yes, good spot

arvidn:
this will snow-ball a bit, there are more places.

Issue #5932, opened 2021-01-31
1.2.12 breaks cross compiling for python binding


xuefer:
**Please provide the following information**

libtorrent version (or branch):
1.2.12 (RC_1_2)

platform/architecture:
openwrt/x86_64

compiler and compiler version:
gcc 10.2.0

please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

```
In file included from src/magnet_uri.cpp:5:
src/boost_python.hpp:10:10: fatal error: boost/python.hpp: No such file or directory
```

unexpectedly: host compiler/linker and flags are used to compile target
```
warning: No toolsets are configured.
warning: Configuring default toolset "gcc".
warning: If the default is wrong, your build may not work correctly.
warning: Use the "toolset=xxxxx" option to override our guess.
warning: For more configuration options, please consult
warning: http://boost.org/boost-build2/doc/html/bbv2/advanced/configuration.html
.....
    "g++"   -fPIC -O3 -finline-functions -Wno-inline -Wall -Wno-deprecated-declarations -fvisibility-inlines-hidden -fvisibility=hidden -DNDEBUG -DTORRENT_USE_I2P=1 -D_WIN3
2_WINNT=0x0600  -I"../../include" -I"/openwrt/source/staging_dir/hostpkg/include/python3.9" -I"/opt/local/include" -I"/usr/local/include" -I"/usr/sfw/includ
e" -I"src"  -c -o "bin/gcc-9.3.0/release/address-model-64/lt-visibility-hidden/python-3.9/src/module.o" "src/module.cpp"
```

"g++" is host compiler which is unexpected, expecting target compiler "ccache_cxx" (providing CXX=ccache_cc by openwrt)
cflags like "-I" flags missing quite a lot for target include dirs, unexpected "/openwrt/source/staging_dir/hostpkg/include/python3.9"  which is from host python sysconfig, expecting target include directory "/openwrt/source/staging_dir/target-x86_64_glibc/usr/include/python3.9”

reason: bindings/python/setup.py does not pass all necessary flags decided by configure like openwrt or gentoo do
see:
https://github.com/openwrt/packages/blob/master/libs/boost/Makefile#L405
```
[...]

PR #5939, opened 2021-02-01
add option to setup.py to opt-out of python auto-configuration

arvidn:
@AllSeeingEyeTolledEweSew I don't think this is quite right, but I don't know why it doesn't work.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -157,9 +157,10 @@ class LibtorrentBuildExt(BuildExtBase):
             "property subdirectories",
         ),
         ("cxxstd=", None, "boost cxxstd value (11, 14, 17, etc)"),
+        ("dont-configure-python", None, "use python configuration from Jamroot.jam or user-config.jam"),
     ]
 
-    boolean_options = BuildExtBase.boolean_options + ["pic", "hash"]
+    boolean_options = BuildExtBase.boolean_options + ["pic", "hash", "dont-configure-python"]
 
     def initialize_options(self):
 
@@ -196,6 +197,7 @@ class LibtorrentBuildExt(BuildExtBase):
             self.cxxstd = '11'
         else:
             self.cxxstd = None
+        self.dont_configure_python = False
         return super().initialize_options()
 
     def run(self):
@@ -219,8 +221,10 @@ class LibtorrentBuildExt(BuildExtBase):
 
         variant = "debug" if self.debug else "release"
         args.append(variant)
-        bits = 64 if sys.maxsize > 2 ** 32 else 32
-        args.append(f"address-model={bits}")
+
+        if not self.dont_configure_python:
+            bits = 64 if sys.maxsize > 2 ** 32 else 32
+            args.append(f"address-model={bits}")
 
         if self.parallel:
             args.append(f"-j{self.parallel}")
@@ -248,10 +252,11 @@ class LibtorrentBuildExt(BuildExtBase):
                     lf = '-L' + str(pathlib.Path(lf[2:]).absolute())
                 args.append(f"linkflags=\"{lf}\"")
 
-        # Jamfile hacks to ensure we select the python environment defined in
-        # our project-config.jam
-        args.append("libtorrent-python=on")
-        args.append(f"python={sysconfig.get_python_version()}")
+        if not self.dont_configure_python:
+            # Jamfile hacks to ensure we select the python environment defined in
+            # our project-config.jam
+            args.append("libtorrent-python=on")
+            args.append(f"python={sysconfig.get_python_version()}")
 
         # Our goal is to produce an artifact at this path. If we do this, the
         # distutils build system will skip trying to build it.
@@ -270,24 +275,28 @@ class LibtorrentBuildExt(BuildExtBase):
         # We use a "project-config.jam" to instantiate a python environment
         # to exactly match the running one.
         python_binding_dir = pathlib.Path(__file__).parent.absolute()
-        config = open(python_binding_dir / 'project-config.jam', 'w+')
-#        config = tempfile.NamedTemporaryFile(mode="w+", delete=False)
+
+        if not self.dont_configure_python:
+            config = open(python_binding_dir / 'project-config.jam', 'w+')
+#            config = tempfile.NamedTemporaryFile(mode="w+", delete=False)
         try:
-            write_b2_python_config(config)
-            config.seek(0)
-            log.info("project-config.jam contents:")
-            log.info(config.read())
-            config.close()
-#            args.append(f"--project-config={config.name}")
+            if not self.dont_configure_python:
+                write_b2_python_config(config)
+                config.seek(0)
+                log.info("project-config.jam contents:")
+                log.info(config.read())
+                config.close()
+#                args.append(f"--project-config={config.name}")
 
             log.info(" ".join(["b2"] + args))
             subprocess.run(["b2"] + args, cwd=python_binding_dir, check=True)
         finally:
             # If we errored while writing config, windows may complain about
             # unlinking a file "in use"
-            config.close()
-#            os.unlink(config.name)
-            os.unlink(python_binding_dir / 'project-config.jam')
+            if not self.dont_configure_python:
+                config.close()
+#                os.unlink(config.name)
+                os.unlink(python_binding_dir / 'project-config.jam')
 
 
 setuptools.setup(


AllSeeingEyeTolledEweSew:
I haven't tested this yet.

Is this just to provide a path to cross-compiling as mentioned in #5932? I'm not super convinced this is the best approach.

There's not much precedent for cross-compiling python extensions. I found [crossenv](https://pypi.org/project/crossenv/). I looked through numpy and found they have a totally custom `distutils`-based build system that can cross-compile. Both tools take the approach of customizing `distutils`' behavior, so we could take an approach to maximize passing information from `distutils` to `b2`. (`distutils` already ingests `CFLAGS`, so maybe this is what we should be doing anyway)

This PR takes the other approach, where we get out of the way and try to let `b2` pick up the right stuff from its ecosystem. Maybe that's the right approach, but it bugs me that I can only find `b2` configurations that are configured by hand. Is the right answer for us to provide a `--python` arg for `setup.py` to pass through as `b2 python=`, or to not pass `python=` at all? Do we only need to care about hand-configured environments, or are there instrumented environments we should care about? I don't know how to answer these questions. If the answer to this is obvious to you, I've missed it after trying hard to answer it.

Since I have a lot of uncertainty and since the OP of #5932 isn't actually cross-compiling, I really want to punt on this until we have a real use case. `RC_1_2`'s `setup.py` is already messy.

arvidn:
> This PR takes the other approach, where we get out of the way and try to let b2 pick up the right stuff from its ecosystem. Maybe that's the right approach, but it bugs me that I can only find b2 configurations that are configured by hand. Is the right answer for us to provide a --python arg for setup.py to pass through as b2 python=, or to not pass python= at all?

not passing all the build configuration *through* `setup.py` (via environment variables, say) doesn't necessarily mean that the configuration is made by hand. Clearly the build script examples, ebuild and openwrt, suggest that they need ways of configuring many aspects of the build. Traditionally this is configured via environment variables, which I think most people agrees these days is not a very robust way of building things. This approach would allow a build script to affect all aspects of building without having `setup.py` be aware of it. It's not just passing the right `python=` argument. It could be passing the right `architecture=` or `target-os=` argument too.

> Since I have a lot of uncertainty and since the OP of #5932 isn't actually cross-compiling, 

The headline of that ticket says he cannot cross compile anymore. Why do you say he's not cross compiling?

AllSeeingEyeTolledEweSew:
> not passing all the build configuration _through_ `setup.py` (via environment variables, say) doesn't necessarily mean that the configuration is made by hand. Clearly the build script examples, ebuild and openwrt, suggest that they need ways of configuring many aspects of the build. Traditionally this is configured via environment variables, which I think most people agrees these days is not a very robust way of building things. This approach would allow a build script to affect all aspects of building without having `setup.py` be aware of it. It's not just passing the right `python=` argument. It could be passing the right `architecture=` or `target-os=` argument too.

Environment variables are still the gold standard for configuring Linux builds. Gentoo bases everything around them for sure.

Even in this ticket in fact, the actual content of the complaint is that `setup.py` doesn't honor `CFLAGS` or `CXX`.

They may not be robust for all cases, but they definitely offer certain features robustly. If I set `CXX`, I can be sure any `libtool` project is going to honor it.

What can I do to get an arbitrary `b2` project to use a particular compiler binary? In your example in the ticket, you translate a compiler binary into some `user-config.jam` by hand, and then need to make sure `toolset=gcc-aarch` (I think?) is passed to all `b2` invocations. Passing it through still needs to be done ad-hoc, and must be explicitly supported by any scripts wrapping `b2`.

> The headline of that ticket says he cannot cross compile anymore. Why do you say he's not cross compiling?

In the first post, his complaint is that `setup.py` ends up invoking `g++` when he has `CXX=ccache_cc`.

In a more recent post he says "both host and target are x86_64".

I assume by "cross-compiling" he really meant selecting a compiler and flags.

My trouble isn't that `b2` has *no obvious way* of passing build configuration; it's that it has *many ways* that might require us to do different things in `setup.py`, and not a lot of clear standards or precedent to decide between them.

One thing we could do is instead of a lot of individual args like `build_ext --toolset=gcc --no-configure-python`, we could have some meta-args like `build_ext --arg=toolset=gcc --no-arg=python` (to indicate we don't autoconfigure `python=`), etc. That would let users build any arbitrary `b2` command line they like. They could try to pass their `CFLAGS` via args to `b2`, or dump it into a `user-config.jam`, or change it to a `--arg=toolset=gcc-aarch`, depending on whatever does make sense.

I don't advocate doing that here. I'm not sure what answer is right. I just think we don't know enough, and any quick fix we apply now is something that needs to be supported later, I guess for `RC_1_2`'s lifetime.

arvidn:
> They may not be robust for all cases, but they definitely offer certain features robustly. If I set CXX, I can be sure any libtool project is going to honor it.

Perhaps, but it may pass flags to your `CXX` that don't work.

> What can I do to get an arbitrary b2 project to use a particular compiler binary? 

(I think that's retorical, since you answer it in the next sentence).

> In your example in the ticket, you translate a compiler binary into some user-config.jam by hand, and then need to make sure toolset=gcc-aarch (I think?) is passed to all b2 invocations. Passing it through still needs to be done ad-hoc, and must be explicitly supported by any scripts wrapping b2.

files aren't necessarily made "by hand", just like environment variables aren't necessarily automated. Both humans and computers can deal with both.

> I don't advocate doing that here. I'm not sure what answer is right. I just think we don't know enough, and any quick fix we apply now is something that needs to be supported later, I guess for RC_1_2's lifetime.

I would like to have some way for a build script to specify exactly what it wants to have built. Perhaps this patch is not it, perhaps an even dumber switch saying "here is the binary python module, now make a package for it". In essence, `setup.py` right now, does two things: builds + package, perhaps there could be a way to just get access to the "package" part.

AllSeeingEyeTolledEweSew:
> I would like to have some way for a build script to specify exactly what it wants to have built. Perhaps this patch is not it, perhaps an even dumber switch saying "here is the binary python module, now make a package for it". In essence, `setup.py` right now, does two things: builds + package, perhaps there could be a way to just get access to the "package" part.

It would definitely be nice to have the option to do this. The killer feature is building python wheels, and sadly the *only* supported tool to build these is `python setup.py bdist_wheel`. There is no standalone tool. That was my whole motivation to refactor `setup.py` in this way in the first place.

Technically the wheel format is just a zipfile layout and we could roll one ourselves, but I expect all the subtleties would be high maintenance and brittle. I, uh, don't want to reinvent the wheel...

I think a lot of awkwardness here is best described as: `b2` and `distutils` are really different, and `setup.py` has behaviors from both, as well as a little `libtool`. Maybe the best thing is to develop a guideline for how to resolve the differences.

I propose:
* By default, `setup.py`'s behavior is "as expected for distutils"
  * The default build configuration is suitable for uploading to pypi
    * Probably defaults to `libtorrent-link=static boost-link=static`
    * (I've been thinking this should also include `deprecated-functions=on`, but that's for another ticket)
  * Builds for the invoking python interpreter
  * Honors `CFLAGS` and other environment variables, or any other behavior that distutils currently has
* `setup.py` is capable of constructing an arbitrary `b2` command, and just packaging the result
  * Probably have something like `--manual-mode` to block all autoconfiguration
  * `setup.py bdist_ext --manual-mode` behaves "as expected for b2"
    * Expects toolsets and python to be configured in `*-config.jam`
    * Doesn't honor `CFLAGS` unless some `*-config.jam` does
    * If any build requirements are necessary to get `b2` to build, they're also necessary here
  * Example: `setup.py bdist_ext --manual-mode --arg=toolset=gcc-custom --arg=address-model=64 --arg=python=x.y bdist_wheel`
* `setup.py` can honor mix-and-matching the two systems in limited ways
  * `setup.py bdist_ext --arg=my-custom-feature=on` would still apply other autoconfiguration
  * `setup.py bdist_ext --arg=python=3.6` would skip python autoconfiguration
  * `setup.py bdist_ext --arg=toolset=gcc` would skip any toolset guesses

What do you think?

arvidn:
> Technically the wheel format is just a zipfile layout and we could roll one ourselves, but I expect all the subtleties would be high maintenance and brittle. I, uh, don't want to reinvent the wheel...

I agree.

> I propose:
> ...

linking statically against boost is only possible with boost-1.74 or later on linux, so that might be a bit of a hurdle. I suppose python packages are expected to be independent from system package managers, like `apt`, so it probably makes sense to link as much as possible statically, to be as self-sufficient as possible. Though, packagers of system packages would probably link shared against both libtorrent and boost.

> What do you think?

I think that sounds good. I'm not sure auto-configuration of python necessarily needs to be disabled even, it could just be disabled via a custom `python=` command line option. It would mean a `project-config.jam` file would never be picked up, but I think that's fine. (i.e. I don't think `--manual-mode` would be necessary)

AllSeeingEyeTolledEweSew:
> linking statically against boost is only possible with boost-1.74 or later on linux, so that might be a bit of a hurdle. I suppose python packages are expected to be independent from system package managers, like `apt`, so it probably makes sense to link as much as possible statically, to be as self-sufficient as possible. Though, packagers of system packages would probably link shared against both libtorrent and boost.

Right. I thought a default of `boost-link=static` would be appropriate since `Jamfile` currently overrides this where it's not supported. A recent boost is required to *actually* build for pypi, but it should degrade gracefully to match whatever the builder's system can do.

I'm not totally sure what's best to support distro package maintainers. The `distutils` path will have the most support but the `b2` path has flexibility. All the existing libtorrent distro build scripts I've seen just use the `libtool` path. I think my #5948 provides good flexibility to do either `distutils` or `b2` though.

> I think that sounds good. I'm not sure auto-configuration of python necessarily needs to be disabled even, it could just be disabled via a custom `python=` command line option. It would mean a `project-config.jam` file would never be picked up, but I think that's fine. (i.e. I don't think `--manual-mode` would be necessary)

Check out my #5948. I think it provides a good framework around the complexity here.

PR #5942, opened 2021-02-02
Ensure we don't attempt to make outgoing connections to RTC peers

arvidn:
@paullouisageneau does this look reasonable?

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -7159,7 +7159,7 @@ namespace {
 #if TORRENT_USE_RTC
 		if (peerinfo->is_rtc_addr)
 		{
-			// unsollicited connection is not possible
+			// unsolicited connection is not possible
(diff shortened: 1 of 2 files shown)


src/torrent.cpp
```diff
@@ -7159,7 +7159,7 @@ namespace {
 #if TORRENT_USE_RTC
 		if (peerinfo->is_rtc_addr)
 		{
-			// unsollicited connection is not possible
+			// unsolicited connection is not possible
 			TORRENT_ASSERT_FAIL();
```
arvidn, line 7163: my test peer hit this assert. I think it's because RTC peers are allowed to be considered connection candidates in the `peer_list` class

paullouisageneau:
Yes it looks good! WebRTC peers can only be connected when randomly put in relation via the tracker, never via the peer list.

paullouisageneau (approved):

PR #5948, opened 2021-02-04
[4/?] Add setup.py build_ext --config-mode

AllSeeingEyeTolledEweSew:
This is an implementation of my suggestion from #5939.

The PR is based on #5905.

This adds `setup.py build_ext --config-mode=[b2|distutils]`. The intent is to have a clear modal deliniation between when users should expect `distutils`-like behavior, when they should expect `b2`-like behavior, and a structured framework for merging the two behaviors.

---

`setup.py bdist_ext --config-mode=b2` is intended for the "just run b2 and package the result" use case, as you mentioned in #5939. It gives full control over the `b2` environment. The usage looks like:

```
python setup.py bdist_ext --config-mode=b2 --b2-args="release address-mode=64 python=3.6 ..."
```

This won't do any auto-configuration. It will just run invoke `b2` with the given command line, and it will expect that the command will output an artifact at the location expected by `distutils`. If it does, it will continue processing the artifact (it can package it with `bdist_wheel`, etc). If the command doesn't produce the expected output, `setup.py` will just fail.

One thing to note is that this doesn't enable cross-compiling wheels. The `bdist_wheel` command will still write wheel metadata appropriate for the running python. Any cross-compiling should probably just use `b2` alone and use the artifact directly.

---

`setup.py bdist_ext --config-mode=distutils` is intended to give the expected behavior of `distutils`. It will build against the running python, doesn't require any special configuration, and by default will produce a build suitable for uploading to pypi.

This mode is close to the current behavior, but to make it pypi-appropriate I've changed it so the default build has `deprecated-functions=on libtorrent-link=static boost-link=static`.

I've also changed the way that `--config-mode=distutils` constructs its `b2` command line, so it can be overridden in a more structured way. It will use the `--b2-args` command line as a starting point, and won't auto-configure any features or arguments that appear in it.

For example, this will add `--hash` to the auto-configured command line:
```
python setup.py bdist_ext --b2-args=--hash
```

This will override the auto-configured default and build without deprecated functions:
```
python setup.py bdist_ext --b2-args=deprecated-functions=off
```

Either of these will block the python auto-configuration:
```
python setup.py bdist_ext --b2-args=--project-config=/path/to/config
python setup.py bdist_ext --b2-args=python=3.6
```

There's a special pattern for "just don't autoconfigure" an argument, because not all arguments have replacements or negatives. For example, we auto-configure `--abbreviate-paths` on windows, but this will negate that:
```
python setup.py bdist_ext --no-autoconf=--abbreviate-paths
```

---

This makes the interaction between automagic and user overrides easier to understand (for `b2` users, anyway), and less limited. I think anyone caught by limitations in `setup.py` will now be able to un-stick themselves.

I've also tried here to draw a brighter line around the scope of the "make b2 behave like distutils" idea. This is a complex project in its own right. I think this PR clarifies where all that complexity should live, both in terms of user expectations and the codebase.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -6,6 +6,7 @@ import distutils.cmd
 import distutils.debug
 import distutils.errors
 import distutils.sysconfig
+import functools
 import os
 import pathlib
 import re
@@ -14,6 +15,7 @@ import subprocess
 import sys
 import sysconfig
 import tempfile
+from typing import Callable
 from typing import cast
 from typing import IO
 from typing import Iterator
@@ -25,19 +27,6 @@ import warnings
 
 import setuptools
 import setuptools.command.build_ext as build_ext_lib
-
-
-def get_msvc_toolset() -> str:
-    # Reference: https://wiki.python.org/moin/WindowsCompilers
-    major_minor = sys.version_info[0:2]
-    if major_minor in ((2, 6), (2, 7), (3, 0), (3, 1), (3, 2)):
-        return "msvc-9.0"
-    if major_minor in ((3, 3), (3, 4)):
-        return "msvc-10.0"
-    if major_minor in ((3, 5), (3, 6)):
-        return "msvc-14.1"  # libtorrent requires VS 2017 or newer
-    # unknown python version
-    return "msvc"
 
 
 def b2_bool(value: bool) -> str:
@@ -91,7 +80,9 @@ def b2_escape(value: str) -> str:
 
 
 def write_b2_python_config(
-    config: IO[str], include_dirs: Sequence[str], library_dirs: Sequence[str]
+    include_dirs: Sequence[str],
+    library_dirs: Sequence[str],
+    config: IO[str],
 ) -> None:
     write = config.write
     # b2 keys python environments by X.Y version, breaking ties by matching
@@ -239,30 +230,6 @@ class LibtorrentBuildExt(build_ext_lib.build_ext):
         self.b2_args = ""
         self.no_autoconf = ""
 
-        self.cxxflags = None
-        self.linkflags = None
-
-        # TODO: this is for backwards compatibility
-        # loading these files will be removed in libtorrent-2.0
-        try:
-            with open("compile_flags") as fp:
-                opts = fp.read()
-                if "-std=c++" in opts:
-                    self.cxxflags = "-std=c++" + opts.split("-std=c++")[-1].split()[0]
-        except OSError:
-            pass
-
-        # TODO: this is for backwards compatibility
-        # loading these files will be removed in libtorrent-2.0
-        try:
-            with open("link_flags") as fp:
-                opt_list = fp.read().split(" ")
-                opt_list = [x for x in opt_list if x.startswith("-L")]
-                if opt_list:
-                    self.linkflags = opt_list
-        except OSError:
-            pass
-
         self.toolset: Optional[str] = None
         self.libtorrent_link: Optional[str] = None
         self.boost_link: Optional[str] = None
@@ -270,8 +237,6 @@ class LibtorrentBuildExt(build_ext_lib.build_ext):
         self.optimization: Optional[str] = None
         self.hash: Optional[bool] = None
         self.cxxstd: Optional[str] = None
-        if os.name == "nt":
-            self.toolset = get_msvc_toolset()
 
         self._b2_args_split: List[str] = []
         self._b2_args_configured: Set[str] = set()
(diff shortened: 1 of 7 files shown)


AllSeeingEyeTolledEweSew:
I think this setup clarifies whether, and where, we should support environment variables like `CFLAGS` and `CXX`. I think both of these things were questionable before.

bindings/python/setup.py
```diff
@@ -153,32 +167,85 @@ def write_b2_python_config(
...
+            "distutils (libtorrent will be built against the invoking "
+            "python, etc; note not all behaviors are currently supported). "
+            "The feature set will match the version found on pypi. "
+            "You can selectively override the auto-configuration in "
+            "this mode with --no-autoconf or --b2-args. For example "
+            "--b2-args=python=x.y or --no-autoconf=python will prevent "
+            "python from being auto-configured",
+        ),
+        (
+            "b2-args=",
```
arvidn, line 193: shouldn't `b2-args` be singular? you can only provide one argument to b2 at a time, right? i.e. `b2-arg`

bindings/python/setup.py
```diff
@@ -82,8 +79,25 @@ def b2_escape(value: str) -> str:
     return f'"{value}"'
 
 
+# Used for understanding a b2 command line. We could get more introspective here,
+# and use python to parse the b2 environment to get all the implicit features
+# from the b2 environment, but this is good enough for now
+IMPLICIT_FEATURE_MAP = {
```
arvidn, line 85: this approach worries me a bit. It would be more reliable to require explicit disabling of certain features (I understand why you want to do that now). As long as the defaults are what most people want, I think it's fine to require some additional arguments to get more advanced uses to work.

arvidn: actually, even better, we could require the custom `b2` arguments to always specify the feature name, and not rely on implicit features at all.

bindings/python/setup.py
AllSeeingEyeTolledEweSew: I couldn't find a reasonable way to pass multiple args here. The only way to override the option parsing mechanism requires overriding some large functions from `distutils` in a subclass, and changing one line out of a few dozen.

So `--b2-args` takes the whole argument sequence as a single string. We parse it with `shlex`, which allows spaces within args to b2, like: `--b2-args 'my-feature="a value with spaces"'`.

I'll update its documentation string to make that a little more clear.

bindings/python/setup.py
AllSeeingEyeTolledEweSew: Sounds good to me, I'll remove this. For `variant` in particular we consider `build_ext --debug` to mean `b2 variant=debug`, so the most-common case is pretty user friendly.

arvidn:
I think I hadn't read your description very carefully. I think this approach is a good one, though I still think it would be fine to reject use of implicit b2 features.

I think any to back port the essence of this to RC_1_2 as well, to get an escape hatch in there.

arvidn:
what's up with this? https://github.com/arvidn/libtorrent/pull/5948/checks?check_run_id=1847631704#step:4:98

AllSeeingEyeTolledEweSew:
> what's up with this? https://github.com/arvidn/libtorrent/pull/5948/checks?check_run_id=1847631704#step:4:98

If I understand right, I think the `pre-commit` github action uses `actions/cache` internally to cache the isolation environments.

Prior to seeing this failure, I rebased my whole chain of PRs and force-pushed them all at nearly the same time. It seems like this cache mechanism can actually collide with itself when storing the cache.

It looks like there are various failure modes that can get the cache stuck in this state for some period of time, but it's a known issue and being worked on. https://github.com/actions/cache/issues/485

I was curious to try to re-run this but I don't know how to force it without making a spurious change to the PR. Can you force a re-run as the repo owner?

arvidn:
in the description of this PR, you said:

```
python setup.py bdist_ext --config-mode=b2 --b2-args="release address-mode=64 python=3.6 ..."
```

You meant to say `build_ext` there, right? I think that's the only option that works with the custom b2 args. Is that right?

AllSeeingEyeTolledEweSew:
> You meant to say `build_ext` there, right? I think that's the only option that works with the custom b2 args. Is that right?

You're right, sorry. No idea how I did that repeatedly. Updated.

AllSeeingEyeTolledEweSew:
I want to clean up some stuff that got changed in the path from this PR -> `RC_1_2` -> `master`, but I wanted to keep all the above discussion above, so I'll close this PR and open a new one.

Issue #2539, opened 2017-11-21
Question about LTO and macOS slow binaries


aldenml:
libtorrent: `master`

I'm asking here to see if anyone has experience compiling `libtorrent` as a shared library for macOS + clang with LTO enabled.

My results is that with Xcode 7.3, the binary is extremely slow (downloading, destructing the session, everything). With Xcode 9.1 it's like 5X faster, but still very slow. Thanks

arvidn:
it's slow relative a shared library on macOS built without LTO?

aldenml:
yes, without LTO, it behaves like I always expected (fast and achieving high transfer speeds), but with LTO, it's like unusable, it takes ages to even destruct the session.

aldenml:
Closing this until I have more information about it

PR #5951, opened 2021-02-05
make release builds with LTO

arvidn:


milahu:
why was this closed?

same issue as #2539?

> clang with LTO enabled ... the binary is extremely slow (downloading, destructing the session, everything)

LTO got better in recent years
but still, probably the gains will be small (5%?) (even with PGO)
because libtorrent is mostly IO-bound (except encryption)

PR #5956, opened 2021-02-07
[5.1.1/?] Apply python formatting and linting to a selection of tools

AllSeeingEyeTolledEweSew:
This lifts some of the `exclude:` lines from `pre-commit` config, and applies our formatting and linting standards.

It's possible to remove them all at once, but the change list is huge and it seems unreasonable to do in a single review.

This PR is based on #5902.

Some things to note:
 * The tools didn't like the hard tabs which occurred in the code-generating scripts. I replaced the hard tabs in strings with `\t`, which has the side effect that my editor had an easier time.
 * You'll notice in some cases I annotate some variables as libtorrent types, like `lt.session`. This is still the same as `Any`, which means it isn't type-checked, and I could annotate these things as `Any`. But I plan to add type annotations to the libtorrent module in the future which will make these annotations relevant, and it will help track down problems in the future to annotate them correctly now.


bindings/c/tools/gen_header.py
```diff
@@ -12,98 +15,106 @@
...
+\t\t\tcase sp::int_type_base:
+\t\t\t\tret.set_int(key, va_arg(lp, int));
+\t\t\t\tbreak;
+\t\t\tcase sp::bool_type_base:
+\t\t\t\tret.set_bool(key, va_arg(lp, int));
+\t\t\t\tbreak;
+\t\t}
+\t\ttag = va_arg(lp, int);
+\t}
+\treturn ret;
```
arvidn, line 109: what happened here?

AllSeeingEyeTolledEweSew: As I mentioned in the description, tools don't like hard tabs in python files very much.

Here, I believe `black` converted all the hard tabs into spaces in the indented python code. Then, `flake8` interpreted the hard tabs in the multiline string as being "indentation", and complained you shouldn't mix hard tabs and spaces for indentation.

It's also burdensome to figure out how to configure an editor to work with hard tabs in just this one file.

`\t` works fine as a tab character for multiline strings, so I replaced the hard tabs by hand.

Other options:
 * We could leave this file on the exclusion lists, so `black` won't autoformat it (I'm not sure `flake8` can easily be configured to ignore mixed-space warnings in this specific file, but I'll see what I can do)
 * We could turn the hard tabs in generated files into spaces

arvidn: so, github prints tab characters as "\t"? All C++ code use tabs, I don't think they are printed as "\t", that's why I thought these were literally backslash-t.

arvidn: You can also replace tabs with `%s/\t/    /g`

AllSeeingEyeTolledEweSew: They *are* literally backslash-t

AllSeeingEyeTolledEweSew: But the backslash-t is in a multi-line quoted string

arvidn: ah, I see. flake8 has opinions about indentation in strings too :)

AllSeeingEyeTolledEweSew: Right. See PyCQA/pycodestyle#45 for some more info on it

arvidn: that thread suggests putting `# noqa` after the string would also fix it. I would prefer that if it works

AllSeeingEyeTolledEweSew:
Rebased

bindings/c/tools/gen_header.py
AllSeeingEyeTolledEweSew: Ah, just saw this comment. I applied this change, narrowly filtering just the whitespace complaints. What do you think?

AllSeeingEyeTolledEweSew:
Oh, is `gen_header.py` used to generate files to be committed to the repo? If so we can get pre-commit to run it, so those headers always stay up-to-date.

For some reason I thought it was used by clients to generate customized headers, I never actually looked at the content and thought about it

arvidn (approved):

PR #5958, opened 2021-02-07
support passing through b2 args through setup.py

arvidn:
@AllSeeingEyeTolledEweSew this is a (slightly) simplified version of your patch, back-ported to `RC_1_2`.
Does it look OK?

I will update the documentation as well

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -11,6 +11,10 @@ import sys
 import sysconfig
 import tempfile
 import subprocess
+import contextlib
+import warnings
+import re
+import shlex
 
 import setuptools
 import setuptools.command.build_ext as _build_ext_lib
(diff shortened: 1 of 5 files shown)


arvidn:
@AllSeeingEyeTolledEweSew one issue with this patch is that the `--b2-args` arguments are not recognized when issuing `install`, just `build_ext`

arvidn:
I see, the order of arguments matter..

arvidn:
I tried to attribute this to you by adding the `Co-authored-by:` to the commit message, but it doesn't look like github picked it up

bindings/python/setup.py
```diff
@@ -264,29 +398,24 @@ def build_extension_with_b2(self):
...
         config = open(python_binding_dir / 'project-config.jam', 'w+')
-#        config = tempfile.NamedTemporaryFile(mode="w+", delete=False)
         try:
             write_b2_python_config(config)
             config.seek(0)
             log.info("project-config.jam contents:")
             log.info(config.read())
             config.close()
-#            args.append(f"--project-config={config.name}")
+            yield
```
AllSeeingEyeTolledEweSew, line 413: This version still unconditionally updates `project-config.jam`. I think this can still collide with the user's intent, if they want to use their own.

Even if the user just wants to use their own `using python : 3.6 : ...` line from their own config, we may override their intent by writing `using python : 3.6 : ...` in our local `project-config.jam`. The `libtorrent-python=on` feature was meant to force this to happen, but it could happen accidentally.

I know it's a little different here because we write to `project-config.jam` instead of adding an arg pointing to a tempfile path. But I think we can check `self._should_add_arg("--project-config")` to gate whether or not we write `project-config.jam`. This way users can do `setup.py build_ext --no-autoconf=--project-config`, if they want to use their own, but we won't actually add an arg if they want it to be autoconfigured.

docs/python_binding.rst
```diff
@@ -11,86 +11,140 @@ libtorrent python binding
...
 
-1. BOOST_BUILD_PATH: "c:/Libraries/boost_1_73_0/tools/build/"
-2. BOOST_ROOT: "c:/Libraries/boost_1_73_0/"
+You can customize the build by passing options to the ``build_ext`` step of
+``setup.py`` by passing arguments directly to ``b2`` via ``--b2-args=``::
 
-Navigate to ``BOOST_ROOT``, execute "bootstrap.bat" and add to the path "c:/Libraries/boost_1_73_0/"
-	
-Create a file ``user-config.jam`` in tour home directory and add this::
+	python setup.py build_ext --b2-args="toolset=msvc-14.2"
```
AllSeeingEyeTolledEweSew, line 86: We should probably have examples of the `shlex` syntax, like
```
python setup.py build_ext --b2-args "toolset=msvc-14.2 libtorrent-python-pic=on"
python setup.py build_ext --b2-args "toolset=msvc-14.2 'my-feature=a space separated value'"
```

docs/python_binding.rst
```diff
@@ -11,86 +11,140 @@ libtorrent python binding
...
-``address-model=64``
+Here, it's important to note that ``build_ext`` has no "memory" of the build
+config and arguments you passed to it before. This is *different* from the way
+``distutils`` normally works. Consider::
 
-This will create the file libtorrent.pyd inside build/lib/ that contains the binding.
-	
-building using boost build (others)
------------------------------------
+	python setup.py build_ext --b2-args"optimization=space"
```
AllSeeingEyeTolledEweSew, line 96: `=`

bindings/python/setup.py
```diff
@@ -184,74 +240,152 @@ def initialize_options(self):
             pass
 
         if os.name == "nt":
-            self.libtorrent_link = 'static'
+            self.toolset = get_msvc_toolset()
```
AllSeeingEyeTolledEweSew, line 243: I started doubting whether `get_msvc_toolset()` is necessary. It always returns `msvc-14.1` or `msvc` on supported python versions.

I couldn't determine from the referenced wiki page whether there was a meaningful difference among `14.x` versions for python extensions.

Also my CI for #5948 broke in windows when I passed `b2 toolset=msvc-14.1`. It worked when I left `toolset` unconfigured. I don't remember the error now. `msvc-14.2` is installed on the CI image.

If autodetection is very broken on older supported boost versions, can we get away with a default of `msvc` instead of `msvc-14.1`?

bindings/python/setup.py
```diff
@@ -136,36 +140,86 @@ def write_b2_python_config(config):
...
 
+        self.config_mode = self.CONFIG_MODE_DISTUTILS
+        self.b2_args = ""
+        self.no_autoconf = ""
+
         self.cxxflags = None
         self.linkflags = None
 
+        # TODO: this is for backwards compatibility
+        # loading these files will be removed in libtorrent-2.0
```
AllSeeingEyeTolledEweSew, line 222: Based on the number of `RC_1_2` python build-related tickets in the past few weeks, it might help to reference that this is specifically to pick up configuration from `libtool`, and that all of `libtool` goes away in 2.0

bindings/python/setup.py
```diff
@@ -136,36 +140,86 @@ def write_b2_python_config(config):
...
-        ("optimization=", None, "b2 optimization mode"),
         (
-            "hash",
+            "cxxstd=",
             None,
-            "use a property hash for the build directory, rather than "
-            "property subdirectories",
+            "(DEPRECATED; use --b2-args=cxxstd=...) "
+            "boost cxxstd value (14, 17, 20, etc.)",
         ),
```
AllSeeingEyeTolledEweSew, line 207: Do we not want to keep `--hash`? I figured this might get used on windows

AllSeeingEyeTolledEweSew (commented):
I assume it was intentional to not add the default of `libtorrent-link=static boost-link=static` here

arvidn:
> I assume it was intentional to not add the default of libtorrent-link=static boost-link=static here

Yes. I want this to remain compatible with `./configure && make` in this branch.

bindings/python/setup.py
arvidn: > This version still unconditionally updates project-config.jam. I think this can still collide with the user's intent, if they want to use their own.

Right, but this branch supports an old enough version of boost, before the `--project-config` argument was a thing.

> 
I know it's a little different here because we write to project-config.jam instead of adding an arg pointing to a tempfile path. But I think we can check self._should_add_arg("--project-config") to gate whether or not we write project-config.jam

makes sense

bindings/python/setup.py
arvidn: yeah, My understanding is that with msvc committing to longer term ABI stability, new versions of python no longer has this additional burden.

arvidn: I think it's fine to remove in later branches

bindings/python/setup.py
arvidn: yeah, I didn't mean to drop this

bindings/python/setup.py
AllSeeingEyeTolledEweSew: Perhaps we could strip down the function to remove checks for older unsupported pythons, and document that it's a deprecated default

bindings/python/setup.py
```diff
@@ -264,30 +411,27 @@ def build_extension_with_b2(self):
...
-            write_b2_python_config(config)
-            config.seek(0)
-            log.info("project-config.jam contents:")
-            log.info(config.read())
-            config.close()
-#            args.append(f"--project-config={config.name}")
-
-            log.info(" ".join(["b2"] + args))
-            subprocess.run(["b2"] + args, cwd=python_binding_dir, check=True)
+            if self._should_add_arg("--project-config"):
```
AllSeeingEyeTolledEweSew, line 420: `... and self._should_add_arg("python")`

bindings/python/setup.py
arvidn: yes, that makes sense. in future versions, like `RC_2_0` or later

AllSeeingEyeTolledEweSew:
Looks good

PR #5960, opened 2021-02-08
add work-around for systems without fseeko() (such as Android)

arvidn:


include/libtorrent/config.hpp
```diff
@@ -175,6 +175,10 @@ POSSIBILITY OF SUCH DAMAGE.
 #if defined __ANDROID__
 #define TORRENT_ANDROID
 #define TORRENT_HAS_FALLOCATE 0
+#if __ANDROID_API__ < 21
```
aldenml, line 178: sorry I didn't mention the right api level at first. This needs to be `24`.

aldenml (commented):
looks good

arvidn:
@aldenml is there an easy-to-install android cross-compiler I could run in CI?

aldenml:
not that I'm aware of, but it's easy to put together another GitHub action similar to what I have. Problem is that I don't think it's possible to run executables. Are you open to just compile and perform no test for android?

arvidn:
I think just compiling is better than nothing

aldenml:
good, I will help with that. If you want to support lower APIs (< 24), you also need hacks around ifaddrs and random. Do you have an opinion already?

arvidn:
> good, I will help with that. If you want to support lower APIs (< 24), you also need hacks around ifaddrs and random. Do you have an opinion already?

I don't have a strong opinion. Do you have patches already?

aldenml:
I will try to send a PR to build with no hacks (API 28), then start lowering with workarounds :)

PR #5962, opened 2021-02-09
add GH action workflow to build android arm 32 bits

aldenml:


.github/workflows/android.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      steps:
+      - name: checkout
+        uses: actions/checkout@v2.3.3
+        with:
+           submodules: true
+
+      - name: install boost
+        run: |
+          wget -nv -O boost.tar.gz https://dl.bintray.com/boostorg/release/${BOOST_DOT_VERSION}/source/boost_${BOOST_VERSION}.tar.gz
```
arvidn, line 29: I had an issue with this earlier, when doing this on travis. bintray has a quota for boost and the download would start to fail until the quota was replenished. iirc, that's when I changed to download from github instead.

.github/workflows/android.yml
```diff
@@ -0,0 +1,73 @@
...
+          export CXX=${ANDROID_TOOLCHAIN}/bin/arm-linux-androideabi-clang++
+          export CC=${ANDROID_TOOLCHAIN}/bin/arm-linux-androideabi-clang
+          echo "boost-build ${BOOST_ROOT}/tools/build/src ;" > boost-build.jam
+          echo "using clang-linux : arm : ${ANDROID_TOOLCHAIN}/bin/arm-linux-androideabi-clang++
+          <cxxflags>-fPIC
+          <cxxflags>-march=armv7-a
+          <cxxflags>-mfpu=neon
+          <cxxflags>-DANDROID
+          <cxxflags>-D__STDC_FORMAT_MACROS
+          <cxxflags>-D_FILE_OFFSET_BITS=64 ;" >>~/user-config.jam;
```
arvidn, line 72: I'm pretty sure these two defines are set by libtorrent already. Are you sure you need them?

.github/workflows/android.yml
```diff
@@ -0,0 +1,73 @@
...
+          ${BOOST_ROOT}/b2 warnings-as-errors=on cxxstd=14 toolset=clang-linux-arm target-os=android link=static crypto=openssl openssl-include=${OPENSSL_ROOT}/include openssl-lib=${OPENSSL_ROOT}/lib
```
arvidn, line 73: if you build the examples instead, you don't have to specify `cxxstd=14`, and merging this (and possible future updates) up to master becomes simpler

arvidn: on the other hand, if you would do a `b2 install --prefix=./package` you could tar it up and upload it as an artifact

.github/workflows/android.yml
aldenml: I remember, also, there is some noise about Bintray going down: https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

I think I can use the one from the system, will try it.

.github/workflows/android.yml
aldenml: need to check again, but it was needed some time ago (years)

.github/workflows/android.yml
aldenml: didn't know, will look into it

.github/workflows/android.yml
arvidn: if only there was some way to crowd-source bandwidth... :)

aldenml:
@arvidn I made the main changes. I decided to keep the build of the library only because I don't think `examples` should be "android compatible". Also, do you think we can leave the artifact package for another time?

arvidn:
yeah, I'm not uploading any other builds either. it would be nice to do, but unrelated to this PR

arvidn (approved):

PR #5968, opened 2021-02-10
[2/?] Fix setup.py's --project-config bypass

AllSeeingEyeTolledEweSew:
This fixes the code for when setup.py's `--project-config` override is *not* used. The `config` variable is undefined in the `finally:` clause.

I'm quite surprised that `flake8` didn't detect this. `mypy` doesn't either, when run locally. They usually catch variable scope problems.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -411,24 +411,25 @@ class LibtorrentBuildExt(BuildExtBase):
         self._maybe_add_arg(f"python-install-path={target.parent}")
         self._maybe_add_arg("install_module")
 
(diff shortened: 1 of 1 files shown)


arvidn:
this is a bug in `RC_1_2` as well. I'll back port this fix

arvidn:
https://github.com/arvidn/libtorrent/pull/5973

arvidn:
I think this can be closed now, right?

AllSeeingEyeTolledEweSew:
Right, looks good.

PR #5969, opened 2021-02-10
Remove msvc toolset default, and libtool support

AllSeeingEyeTolledEweSew:
This removes the `toolset=msvc-X.Y` default on windows, which seems to not be effective these days.

It also removes the libtool support, which I don't think was intended to be in master.

This is based on #5968.

PR #5970, opened 2021-02-10
Default setup.py to do static linking on all platforms

AllSeeingEyeTolledEweSew:
As we discussed before, `setup.py` should default to do static linking.

It won't do this with `setup.py build_ext --config-mode=b2`.

This is based on #5969.


arvidn:
looks good. This top commit is independent of the other ones though, and could land directly into master

AllSeeingEyeTolledEweSew:
Sounds good. I rebased and changed it to just the single commit.

arvidn (approved):

PR #5978, opened 2021-02-13
compiling android arm 32 bits with api level 19

aldenml:


src/mmap.cpp
```diff
@@ -416,14 +414,11 @@ file_handle::file_handle(string_view name, std::int64_t const size
 	}
 #endif
 
-#ifdef POSIX_FADV_RANDOM
+#if (TORRENT_HAS_FADVISE && defined POSIX_FADV_RANDOM)
```
arvidn, line 417: interesting. android has the `POSIX_FADV_RANDOM` macro defined, but not the `posix_fadvice()` function?

arvidn (approved):

src/mmap.cpp
aldenml: that's correct

PR #5979, opened 2021-02-15
Fix error messages related to opening files

arvidn:
specifically truncate and allocate failures (which currently are incorrectly attributed to open file)


aldenml:
LGTM

PR #5990, opened 2021-02-18
[7/?] Expand test coverage of the python bindings

AllSeeingEyeTolledEweSew:
This PR has three commits:
* Use `pytest` to run python bindings tests, instead of directly running a `test.py`
* Split `test.py` into multiple files
* Radically expand test coverage

I came up with some guidelines for what we should test, and added these at `guidelines.txt`. TL;DR: we don't care about testing libtorrent's correctness, we just care about testing data access, data round-trips, correctness of complex binding-specific code (e.g. `parse_magnet_uri_dict`), as well as some tightly-coupled stuff like memory management and filesystem access.

I recognize this PR is massive. It's not complex per se, just a lot of standalone functions that ought to conform to the guidelines. So I thought it might be easier to review as single PR instead of "chained" PRs I've made in the past, but if you'd rather see it split up, let me know and I'll do so.

There are many TODOs in this PR. Some refer to other issues I've filed; some are just things I couldn't figure out on my own. I don't think any of the TODOs are crucial to this PR though, and can be addressed later.


arvidn:
This is great!

what is this file for? `bindings/python/url_seed_multi.torrent`
Also, are you confident that none of the recent tests added to `test.py` may have been lost? (I don't know of a simple way of verifying that)

AllSeeingEyeTolledEweSew:
> what is this file for? `bindings/python/url_seed_multi.torrent`

This is used by `client_test.py`, `simple_client_test.py` and `make_torrent_test.py`. Those tests have been copied verbatim from `test.py`, where there's currently "setup" code like:
```python
if __name__ == '__main__':
    try: 
        shutil.copy(os.path.join('..', '..', 'test', 'test_torrents',
                                 'url_seed_multi.torrent'), '.') 
    except shutil.SameFileError:
        pass 
```

I'll clean up and expand those tests in a different PR. For now I just made `url_seed_multi.torrent` a symlink for sanity.

> Also, are you confident that none of the recent tests added to `test.py` may have been lost? (I don't know of a simple way of verifying that)

I'm "pretty confident." I've tried to ensure that all the recent changes got captured every time I rebase.

I could make the changes easier to see, by splitting this into more PRs. The first could copy lines verbatim from `test.py` into multiple files; subsequent PRs could expand each `*_test.py` file.

That might be a lot of work for nothing, though. Once this lands, I'm going to spam you with PRs to fix all the tests marked with `@unittest.skip(...)`, so I'll be manually combing through all the python bindings again, and I'll be sure to add any tests currently missing. IMO, this process will be at least as thorough as the thoroughness of the current `test.py`, so it's worth landing this PR quickly, so I can start applying fixes earlier.

PR #6000, opened 2021-02-21
pkg config fixes

arvidn:


arvidn:
changing the openssl library names on windows to only apply to msvc caused this failure on mingw: https://ci.appveyor.com/project/arvidn/libtorrent/builds/37938140/job/t1j8dv069f8a0cp5

arvidn:
@emmercm does this patch still work for you? I made a few fixes since you tested it. I would think for the better.

emmercm:
@arvidn I'll try to check this out early this week.

emmercm:
This appeared to work building qBittorrent v4.2.5, when patching against `RC_1_2`. Thank you!

Issue #6004, opened 2021-02-22
crash in parse_websocket_tracker_response


arvidn:
libtorrent version (or branch): master
platform/architecture: linux/x86
compiler and compiler version: GCC 9.2

This happened on my test of webtorrent seeding (using libdatachannel 10.2), but it doesn't seem related to libdatachannel:

```
==182004==ERROR: AddressSanitizer: unknown-crash on address 0x7fcde35fc4c0 at pc 0x0000015d8e6b bp 0x7fcde35fb7d0 sp 0x7fcde35fb7c0
READ of size 20 at 0x7fcde35fc4c0 thread T1
    #0 0x15d8e6a in libtorrent::digest32<160l>::assign(char const*) ../include/libtorrent/sha1_hash.hpp:108
    #1 0x15d8e6a in libtorrent::digest32<160l>::digest32(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ../include/libtorrent/sha1_hash.hpp:95
[...]

PR #6005, opened 2021-02-22
fix issue in parsing invalid webtorrent tracker responses

arvidn:
this is to address https://github.com/arvidn/libtorrent/issues/6004

### DIFF
--- a/src/websocket_tracker_connection.cpp
+++ b/src/websocket_tracker_connection.cpp
@@ -420,6 +420,8 @@ try {
 		auto sdp = payload_offer["sdp"].as_string();
 		auto id = utf8_latin1(payload["offer_id"].as_string());
 		auto pid = utf8_latin1(payload["peer_id"].as_string());
+		if (pid.size() != 20)
+			throw std::invalid_argument("invalid peer_id size " + std::to_string(pid.size()));
 
 		aux::rtc_offer_id oid{span<char const>(id)};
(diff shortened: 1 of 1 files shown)


arvidn:
@paullouisageneau I think we need a unit test for this function, `parse_websocket_tracker_response()`, similar to the parser for regular tracker responses ([here](https://github.com/arvidn/libtorrent/blob/master/test/test_tracker.cpp)).

Ideally it should include malicious responses (like a `peer_id` that has the wrong size).

It would probably also make sense to add a fuzzer for this function, like [this](https://github.com/arvidn/libtorrent/blob/master/fuzzers/src/http_tracker.cpp) one.

arvidn:
it also seems a bit excessive to use exceptions internally in there

paullouisageneau (approved):

paullouisageneau:
> @paullouisageneau I think we need a unit test for this function, `parse_websocket_tracker_response()`, similar to the parser for regular tracker responses ([here](https://github.com/arvidn/libtorrent/blob/master/test/test_tracker.cpp)).
> 
> Ideally it should include malicious responses (like a `peer_id` that has the wrong size).
> 
> It would probably also make sense to add a fuzzer for this function, like [this](https://github.com/arvidn/libtorrent/blob/master/fuzzers/src/http_tracker.cpp) one.

Yes, makes sense, I'll add the tests.

> it also seems a bit excessive to use exceptions internally in there

The reason was at first the exception needed to be handled for nlohmann::json. However, I think with Boost.JSON error_codes could be used instead.

PR #6010, opened 2021-02-24
fix python binding for get_torrent_status()

arvidn:
where a predicate is passed in


arvidn:
@categorical I updated this patch (force-pushed) to ensure only a *reference* to the lambda is passed into the libtorrent thread. I could reproduce the crash in a virtual machine before this change, but not after.

Could you give it another try?

PR #6015, opened 2021-02-28
fix need_save_resume_data() for renaming files

arvidn:
share-mode, upload-mode, disable- pex, lsd, and dht.

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -770,6 +770,9 @@ bool is_downloading_state(int const st)
 		{
 			inc_stats_counter(counters::non_filter_torrents);
 		}
+
+		set_need_save_resume();
+
 		m_apply_ip_filter = b;
 		ip_filter_updated();
 		state_updated();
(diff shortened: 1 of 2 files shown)


src/torrent.cpp
```diff
@@ -1028,11 +1031,23 @@ bool is_downloading_state(int const st)
 		if (mask & torrent_flags::stop_when_ready)
 			stop_when_ready(bool(flags & torrent_flags::stop_when_ready));
 		if (mask & torrent_flags::disable_dht)
-			m_enable_dht = !bool(flags & torrent_flags::disable_dht);
+		{
+			bool const new_value = !bool(flags & torrent_flags::disable_dht);
+			if (m_enable_dht != new_value) set_need_save_resume();
```
glassez, line 1036: Maybe compare old and new flags at once?

arvidn: what do you mean? You mean also move the assignment `m_enable_dht = new_value` into the if-block?

glassez: No. I mean something like `if (flags != m_flags) set_need_save_resume();` once per `set_flags()` instead of checking each flag independently.

arvidn: ah, I'll take a closer look at that. I don't think all flags affect the resume data (although, most do).

arvidn: one problem with that is that the flags aren't stored as flags internally, they are individual bits in a bitfield. It would be an improvement and simplification to store them as flags internally too. I think I'll make that change in `master` though.

The reason for this is that these features were originally controlled by individual API calls (as you may recall), and using flags was a way to simplify the API, and also make it more extendable without breaking ABI. The internal representation hasn't caught up yet.

PR #6018, opened 2021-02-28
prevent duplicate args being passed to b2 in setup.py

arvidn:
@AllSeeingEyeTolledEweSew Does this look right?
I was surprised to realize duplicate arguments could be passed to `b2`. This fixes it.

The command line that failed for me was:

```
BOOST_ROOT="" python3 setup.py build_ext --libtorrent-link=prebuilt
```

The `b2` command line would end up with *both* 'libtorrent-link=prebuilt' and 'libtorrent-link=static'. It would work to pass this as `--b2-args=` because it sets the `configured` set. But this hasn't been released yet, and I think it would make sense for both to be supported for now.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -302,27 +302,32 @@ class LibtorrentBuildExt(BuildExtBase):
                 "--libtorrent-link is deprecated; use --b2-args=libtorrent-link=..."
             )
             self._maybe_add_arg(f"libtorrent-link={self.libtorrent_link}")
+            self._b2_args_configured.add("libtorrent-link")
         if self.boost_link:
             warnings.warn("--boost-link is deprecated; use --b2-args=boost-link=...")
             self._maybe_add_arg(f"boost-link={self.boost_link}")
+            self._b2_args_configured.add("boost-link")
         if self.toolset:
             warnings.warn("--toolset is deprecated; use --b2-args=toolset=...")
             self._maybe_add_arg(f"toolset={self.toolset}")
+            self._b2_args_configured.add("toolset")
         if self.pic:
             warnings.warn("--pic is deprecated; use --b2-args=libtorrent-python-pic=on")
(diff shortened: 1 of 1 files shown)


yurivict:
Somehow https://github.com/arvidn/libtorrent/pull/6018 doesn't apply to 2.0.2 - there is no ```_should_add_arg```.

arvidn:
I see. It applies cleanly to `RC_2_0`, but there's been some changes there since 2.0.2. Hold on.

arvidn:
I don't think you need this patch for 2.0.2 actually. This command works on the `v2.0.2` tag for me (on ubuntu):

```
python3 setup.py build_ext --libtorrent-link=prebuilt
```

AllSeeingEyeTolledEweSew:
What about free args? `--b2-args="library-path=foo library-path=bar"`

arvidn:
good point. user-provided arguments should probably not suppress each other, but it would make sense to have user-supplied arguments suppress built-in defaults. None of the defaults are free features afaik. Having multiple values of any of them causes multiple targets to be built.

arvidn:
@AllSeeingEyeTolledEweSew actually, your example is the *only* case the current logic applies to. the `_b2_args_configured` set is only updated by arguments passed in via `--b2-args`, so if you pass in the same twice, only one will take effect. Am I reading this wrong?

I updated the patch to more precisely address the issue. But I still think your example won't work currently

AllSeeingEyeTolledEweSew:
> @AllSeeingEyeTolledEweSew actually, your example is the _only_ case the current logic applies to. the `_b2_args_configured` set is only updated by arguments passed in via `--b2-args`, so if you pass in the same twice, only one will take effect. Am I reading this wrong?

`--b2-args` gets split into `_b2_args_split`, so any free-feature properties passed in `--b2-args` are all still honored. `_b2_args_configured` is just used to note which features have been specified by the user, to gate auto-configuration.

Patch looks good to me

PR #6020, opened 2021-03-02
improve comment in piece-picker

arvidn:


arvidn:
@markmdscott how does this look?

markmdscott:
> @markmdscott how does this look?

Yes this is much improved! It would not have confused a neophyte like me.

PR #6042, opened 2021-03-07
add size_on_disk() to file_storage and torrent_info

arvidn:
reporting the total size of non-pad files


src/file_storage.cpp
```diff
@@ -740,6 +740,8 @@ namespace aux {
 		}
 
 		m_total_size += e.size;
+		m_size_on_disk += e.size;
```
master255, line 743: Where is the logic?

src/file_storage.cpp
```diff
@@ -1223,13 +1227,16 @@ namespace {
 			TORRENT_ASSERT(off < max_file_offset - static_cast<std::int64_t>(file.size));
 			file.offset = static_cast<std::uint64_t>(off);
 			off += file.size;
+			on_disk += file.size;
```
master255, line 1230: Where is the logic?

test/test_file_storage.cpp
```diff
@@ -52,6 +52,7 @@ void setup_test_storage(file_storage& st)
 	TEST_EQUAL(st.file_offset(file_index_t{3}), 60000);
 
 	TEST_EQUAL(st.total_size(), 100000);
+	TEST_EQUAL(st.size_on_disk(), 100000);
```
master255, line 55: For what purpose make another variable, if it will also be equal to total_size

arvidn: As I've already explained, the total size and size_on_disk are the same for torrents with no pad files.

```
total_size == size_on_disk + pad_file_size
```

master255: @arvidn Maybe I'm not seeing something. Where do you filter .pad files?

master255: @arvidn Okay. I found where the pad files are filtered. But you have to test my .torrent file. The information in it doesn't match. Maybe there are some other problems.

PR #6051, opened 2021-03-10
build android arm 32 bits with webtorrent enabled

aldenml:


aldenml:
hi @arvidn, do I need to do something different for `Linux/pre-commit`?

arvidn:
hm. I thought I had fixed that.

arvidn:
https://github.com/arvidn/libtorrent/pull/6052

src/torrent_info.cpp
```diff
@@ -344,7 +344,7 @@ namespace {
 			return false;
 		}
 
-		std::time_t const mtime(dict.dict_find_int_value("mtime", 0));
+		std::time_t const mtime(static_cast<std::time_t>(dict.dict_find_int_value("mtime", 0)));
```
arvidn, line 347: is `time_t` a 32 bit type on android 32bit by any chance? If so, maybe this need to be fixed in a non-truncating way. Or maybe we just don't care. Will anyone have sympathy for people still running android 32bit in 2038?

aldenml: yes, it is 32 bit in android 32 bit. I don't think it's important (as long as it was stored correctly), and there are similar truncations (for `time_t`) in other places. 2038? who knows :)

arvidn: I still feel a bit uncomfortable with this. I haven't thought it through yet. I used `time_t` here to make it clear what the type of the value was (posix-time, seconds since epoch). But maybe it's better to make this `std::int64_t` unconditionally.

aldenml: well, looking at this, the good news is that `add_file_borrow` receives a `std::int64_t`, so we can just change the type of `mtime` and call it a day, but, `add_file` still receives `std::time_t`, not clear to me if they should have the same type. My proposal is:

1. change `mtime` to `std::int64_t`
2. if you feel like it, change `add_file` to receive  `std::int64_t`

let me know.

PR #6054, opened 2021-03-11
fix issue where peers entering upload-mode cannot leave upload-mode

arvidn:
don't mark upload-only peers as seeds, as it breaks the num_seeds accounting

### DIFF
--- a/examples/client_test.cpp
+++ b/examples/client_test.cpp
@@ -315,7 +315,7 @@ int print_peer_info(std::string& out
 	int pos = 0;
 	if (print_ip) out += "IP                             ";
 	if (print_local_ip) out += "local IP                       ";
-	out += "progress        down     (total | peak   )  up      (total | peak   ) sent-req tmo bsy rcv flags         dn  up  source  ";
+	out += "progress        down     (total | peak   )  up      (total | peak   ) sent-req tmo bsy rcv flags          dn  up  source  ";
 	if (print_fails) out += "fail hshf ";
 	if (print_send_bufs) out += "rq sndb (recvb |alloc | wmrk ) q-bytes ";
 	if (print_timers) out += "inactive wait timeout q-time ";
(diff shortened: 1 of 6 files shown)


Sophist-UK:
@arvidn Thanks - it was a long road getting here, and the patch hits more lines of code than I might have imagined, but thanks for bearing with us.

arvidn:
While testing this I realized there's a brief window where a peer connection will count as a seed even though we're not 100% sure it is one.

This how it works. The peer-list (of peers we know about, not necessarily connected to) records information about all peers we know about as part of the swarm. Peers received from trackers, PEX, LSD, DHT as well as incoming connections we haven't heard about before. All peers that are connected have an entry in this list, but we also have entries for peers we used to be connected to and peers we've never connected to.

This list records information like the number of bytes we've uploaded and downloaded from the peer, whether it supports uTP, protocol encryption and also whether it's a seed. If we have been connected to a peer that was a seed, it's recorded. We use this to avoid even attempting to connect to such peers once we become a seed ourselves.

When we initiate an outgoing connection, we create a peer connection object which is part of the peer *connections* list. i.e. connection attempts count as connected peers, and in this mode the seed flag will also consider this peer connection a seed, counted against our internal connected-seeds counter.

However, the `torrent_status::num_seeds` counter is not supposed to include *potential* seeds, it's supposed to only count actually connected seeds. For this reason, there's a separate (internal) counter for the number of seeds that are currently in the "connecting" state, i.e. we haven't connected to the peers yet. This counter is deducted from the internal seed counter to produce `torrent_info::num_seeds`.

Here's the issue: when a potential seed completes, i.e. it becomes connected, we will start counting it immediately against the `num_seeds` counter. However, at that point it may not have completed its bittorrent *handshake* with the peer. Only after completing the bittorrent handshake will we know which pieces the peer has. Additionally, the peer may decide to disconnect us immediately after the handshake, if it already has a full set of peers.

This leaves a window of up to a second or two where potential seeds count as seeds in the seed counter.

PR #6055, opened 2021-03-12
restore setup.py to support out-of-tree build directory

arvidn:


sledgehammer999:
I am checking it now. My local VM is a bit slow.

sledgehammer999:
I had to restart the build because it crashed. Now it finished and this patch works.

PR #6058, opened 2021-03-13
return nullptr instead of crashing in get_storage_impl

arvidn:


src/torrent.cpp
```diff
@@ -1307,6 +1307,7 @@ bool is_downloading_state(int const st)
 
 	storage_interface* torrent::get_storage_impl() const
 	{
+		if (m_storage) return nullptr;
```
pavel-pimenov, line 1310: !m_storage ?

arvidn: doh!

PR #6062, opened 2021-03-15
allow no-interest timeouts of peer connections before all connections…

arvidn:
… slots are full

### DIFF
--- a/src/peer_connection.cpp
+++ b/src/peer_connection.cpp
@@ -1754,9 +1754,10 @@ namespace libtorrent {
 		peer_log(peer_log_alert::incoming_message, "INTERESTED");
 #endif
 		if (m_peer_interested == false)
+		{
 			m_counters.inc_stats_counter(counters::num_peers_up_interested);
-
-		m_peer_interested = true;
+			m_peer_interested = true;
+		}
 		if (is_disconnecting()) return;
 
 		// if the peer is ready to download stuff, it must have metadata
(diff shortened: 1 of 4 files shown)


src/peer_connection.cpp
```diff
@@ -1841,15 +1842,16 @@ namespace libtorrent {
...
 
-		m_became_uninterested = aux::time_now();
-
 #ifndef TORRENT_DISABLE_LOGGING
 		peer_log(peer_log_alert::incoming_message, "NOT_INTERESTED");
 #endif
 		if (m_peer_interested)
+		{
 			m_counters.inc_stats_counter(counters::num_peers_up_interested, -1);
+			m_became_uninterested = aux::time_now();
```
arvidn, line 1851: this is the main change, to not restart the timestamp on duplicate not-interested messages

src/peer_connection.cpp
```diff
@@ -4954,6 +4962,13 @@ namespace libtorrent {
 		time_duration const time_limit = seconds(
 			m_settings.get_int(settings_pack::inactivity_timeout));
 
+		// if we are close enough to the limit, consider the peer connection
+		// list full. This will enable the inactive timeout
+		bool const max_session_conns = m_ses.num_connections()
+			>= m_settings.get_int(settings_pack::connections_limit) - 5;
+		bool const max_torrent_conns = t && t->num_peers()
+			>= t->max_connections() - 5;
```
arvidn, line 4970: this is the second part of the logic change, to treat the connection slots at limit slightly before reaching the absolute limit

PR #6074, opened 2021-03-19
Don't set need_save_resume_data when update piece priorities on init

### LINKED ISSUE #6071: Incorrect behavior of need_save_resume_data on torrent startup
The issue is confirmed with libtorrent RC_2_0 but RC_1_2 is most likely affected too.

When torrent is restored (i.e. added using previously stored resume data) it still has `need_save_resume_data` set unexpectedly. As I can see the wrong place is `update_piece_priorities()` that is called from `init()`. If resume data provides file priorities it calls `prioritize_pieces()` that sets `need_save_resume_data` to `true`.

glassez:
Dunno if other code paths in `torrent::init()` can also lead to unexpected changes of `need_save_resume_data`...
Closes #6071.

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -1778,7 +1778,12 @@ bool is_downloading_state(int const st)
 		// in case file priorities were passed in via the add_torrent_params
 		// and also in the case of share mode, we need to update the priorities
 		// this has to be applied before piece priority
-		if (!m_file_priority.empty()) update_piece_priorities(m_file_priority);
+		if (!m_file_priority.empty())
+		{
+			bool const need_save_resume_data = m_need_save_resume_data;
+			update_piece_priorities(m_file_priority);
+			m_need_save_resume_data = need_save_resume_data;
+		}
 
 		if (m_add_torrent_params)
 		{


arvidn:
good catch! I'm a bit tempted to instead solve this by removing the `set_need_save_resume()` in `prioritize_pieces()`, and make that a lower level function instead.
This overlaps a bit with the patch I'm working on for the other ticket. I'd like to hold off on this one until I've explored that solution a bit more

glassez:
> I'd like to hold off on this one until I've explored that solution a bit more

It would be highly desirable if it was fixed in one way or another for the release of version 1.2.13/2.0.3.

arvidn:
yes, I will hold off 1.2.13 for this

glassez:
> yes, I will hold off 1.2.13 for this

I hope it won't last long after all.

arvidn:
I would also like to add some tests where torrents are added with a bunch of different `add_torrent_params` settings, and ensure `need-save-resume` isn't set immediately after.

arvidn:
I think this is the simplest solution. but I think it should go into `RC_1_2`, and I think there should be a comment explaining why `m_need_save_resume` shouldn't be affected

glassez:
> but I think it should go into `RC_1_2`

I hope you don't intend to release v2.0.3 w/o any fix for this issue.


> I think there should be a comment explaining why `m_need_save_resume` shouldn't be affected

Can you help with exact wording of such a comment?

glassez:
> I would also like to add some tests where torrents are added with a bunch of different `add_torrent_params` settings, and ensure `need-save-resume` isn't set immediately after.

You can also add some assertion, can't you?
No matter how hard you try, you may not be able to check all possible cases in the tests.

arvidn:
I folded your patch into this, as part of testing the original fix too: https://github.com/arvidn/libtorrent/pull/6076

arvidn:
incorporated in https://github.com/arvidn/libtorrent/pull/6076

PR #6076, opened 2021-03-19
fix issue with setting file priorities and saving resume data

arvidn:
The need-save-resume flag should not be set until *after* the new file priorities have been applied. Don't set the need-save-resume flag just by loading the resume data (when applying the file priorities into the piece priorities)

### DIFF
--- a/simulation/test_transfer.cpp
+++ b/simulation/test_transfer.cpp
@@ -399,7 +399,8 @@ TORRENT_TEST(is_finished)
 				TEST_EQUAL(is_finished(ses), false);
 				std::vector<download_priority_t> prio(4, dont_download);
 				ses.get_torrents()[0].prioritize_files(prio);
-				TEST_EQUAL(is_finished(ses), true);
+				// applying the priorities is asynchronous. the torrent may not
+				// finish immediately
 			}
 		},
 		[](std::shared_ptr<lt::session> ses[2]) {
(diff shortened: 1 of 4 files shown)


glassez (approved):

simulation/test_transfer.cpp
```diff
@@ -399,7 +399,8 @@ TORRENT_TEST(is_finished)
 				TEST_EQUAL(is_finished(ses), false);
 				std::vector<download_priority_t> prio(4, dont_download);
 				ses.get_torrents()[0].prioritize_files(prio);
-				TEST_EQUAL(is_finished(ses), true);
+				// applying the priorities is asynchronous. the torrent may not
+				// finish immediately
```
arvidn, line 403: this was an unintended consequence of this change. it's making setting the *piece* priorities asynchronous too. Which in some sense is more correct, to keep file- and piece priorities in sync. But it also makes me a bit worried about ease of use and efficiency. It would be good if this could be more immediate.

Perhaps one approach would be to disconnect the file priorities from whether the pieces are stored in the part file or not. That way, the priorities could be updated immediately, and only after the files have moved is the "use-partfile" bit updated.

To some extend that state is already kept in the disk thread, to support adding an existing torrent that didn't use a part file.

arvidn: although, making that change would definitely be too big for `RC_1_2`.

PR #6078, opened 2021-03-20
prefer __builtin_popcountl() over portable C implementation of popcnt

arvidn:
This enables the ARM64 SIMD instruction on Apple M1

### DIFF
--- a/src/bitfield.cpp
+++ b/src/bitfield.cpp
@@ -109,6 +109,9 @@ namespace libtorrent {
 
 		for (int i = 1; i < words + 1; ++i)
 		{
+#if defined __GNUC__ || defined __clang__
+			ret += __builtin_popcountl(m_buf[i]);
+#else
 			std::uint32_t const v = m_buf[i];
 			// from:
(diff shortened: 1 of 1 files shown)


aldenml:
this makes me realize that the code path for "arm neon" does not support arm64, and with this patch, the optimization should come out of the box for android 64 bits.

arvidn:
I was looking at that too. I think it depends on the compiler flags, and that clang on M1 *always* builds with the necessary arch flags to let the compiler know which hardware features exist. I would think, generically, this wouldn't be the case just by building on aarch64 linux for instance.

I intend to test this on a RPi 4 to see what happens there.

arvidn:
as far as I can tell, without specifying simd support on the GCC command line, RPi 4 doesn't turn the `__builtin_popcount` into the `cnt` instruction. but I may be missing something. I also can't figure out how to get an assembly version to work for aarch64

aldenml:
next time I compile `master` with this, I will look at the generated code in android

PR #6079, opened 2021-03-20
add a file_prio_alert, posted when file priorities are applied

arvidn:


src/torrent.cpp
```diff
@@ -5347,6 +5347,12 @@ namespace {
 #endif
 		}
 
+		if (!err)
+		{
+			if (alerts().should_post<file_prio_alert>())
+				alerts().emplace_alert<file_prio_alert>(get_handle());
+		}
+
 		if (err)
```
glassez, line 5356: `else`?

include/libtorrent/alert_types.hpp
```diff
@@ -2934,6 +2934,27 @@ TORRENT_VERSION_NAMESPACE_3
 		aux::noexcept_movable<tcp::endpoint> ip;
 	};
 
+	// posted when a prioritize_files() or file_priority() update of the file
+	// priorities complete, which requires a round-trip to the disk thread.
+	//
+	// If the disk operation fails, ``operation_t`` and ``error`` indicates what
```
glassez, line 2940: Seems it isn't implemented...

arvidn: oops. my first pass had error states, but there's already a `file_error_alert` being posted in this case

src/torrent.cpp
```diff
@@ -5347,6 +5347,12 @@ namespace {
 #endif
 		}
 
+		if (!err)
+		{
+			if (alerts().should_post<file_prio_alert>())
+				alerts().emplace_alert<file_prio_alert>(get_handle());
```
glassez, line 5353: Can correct priorities be saved in resume data at this point?

arvidn: the whole reason to delay applying the file priorities is to *not* save the new file priorities in this case, since they failed to be applied. The problem with saving them here is that the next time the torrent is loaded, the location on disk for blocks will be based on the priorities, and blocks will go missing and there's a risk a full re-check is required .

As I've been thinking about this, I think it was a mistake to conflate the location on disk with the priorities, and those should probably be saved as separate entries in the resume data.

arvidn: Sorry, I thought this comment was on the error path, but it isn't. The answer to your question is "yes". and my previous comment applies to a few lines down, in `if (err) {`

src/torrent.cpp
arvidn: there is an else attach to this `if` already, it wouldn't work unfortunately.

src/torrent.cpp
arvidn: also, as I look at this. It doesn't look right. the `m_file_priority` member is updated before checking the error.

arvidn: oh, never mind. the disk thread will send back the new file priorities based on which ones succeeded.

glassez:
Will it be ported to RC_1_2 branch?

glassez (approved):

arvidn:
> Will it be ported to RC_1_2 branch?

I'm a bit hesitant to that since I'm not sure I could assign the same identifier to the alert as in 2.0. If you think it would be really helpful I could look into it further. It's possible it won't be a problem.

glassez:
> > Will it be ported to RC_1_2 branch?
> 
> I'm a bit hesitant to that since I'm not sure I could assign the same identifier to the alert as in 2.0. If you think it would be really helpful I could look into it further. It's possible it won't be a problem.

Seeing libtorrent-1.2 as a thing of the past, I wouldn't insist on it if it's burdensome for you. Given the previous fixes, we can now at least hope that priorities will be saved by timeout. And linking with libtorrent-2.0, we will save it by this alert.
Anyway thank you!

glassez:
BTW, how about fix `need_save_resume_data`  on torrent initialization for libtorrent-2.0?

arvidn:
> how about fix need_save_resume_data on torrent initialization for libtorrent-2.0?

Anything I land in `RC_1_2` gets merged into `RC_2_0` as well. And I think the test part of the previous patch tests this (at least a smoke test, not exhaustive). [here](https://github.com/arvidn/libtorrent/pull/6076/files#diff-ece1890f8d1b129158c2dc140889f4c57280060d996e61525f2121ef731b984eR653).

glassez:
@arvidn, seems you forgot to add `file_prio_alert` forward declaration in `libtorrent/fwd.hpp`.

arvidn:
yeah, I just noticed when I merged into master

PR #6093, opened 2021-03-27
peer connection sims

arvidn:
* Raise the severity of peers sending invalid messages to 'peer_error'.
* Simplify m_upload_only to just record the explicit upload-only state communicated via the extension message. Implied upload only is moved to the upload_only() query.
* Add simulations for peer connection tests. Mostly for have-all, have-none and bitfields, and some simple invalid requests. This will serve as a platform to add more sophisticated tests.

### DIFF
--- a/include/libtorrent/peer_connection.hpp
+++ b/include/libtorrent/peer_connection.hpp
@@ -427,7 +427,7 @@ namespace aux {
 #endif
 
 		void set_upload_only(bool);
-		bool upload_only() const { return m_upload_only; }
+		bool upload_only() const { return m_upload_only || is_seed() || m_have_all; }
 
 		void set_holepunch_mode() override;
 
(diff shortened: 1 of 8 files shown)


include/libtorrent/peer_connection.hpp
```diff
@@ -1143,7 +1143,9 @@ namespace aux {
 		bool m_share_mode:1;
 #endif
 
-		// set to true when this peer is only uploading
+		// set to true when this peer has told us explicitly that it is onlu
```
aldenml, line 1146: typo in `onlu`

simulation/fake_peer.hpp
```diff
@@ -180,6 +172,17 @@ struct fake_peer
 			});
 	}
 
+private:
+
+	void send_simple(std::uint8_t msg)
```
aldenml, line 177: may I suggest a different name? like `send_simple_message` or `send_small_message`? also is `msg` a "message code"?

arvidn: yes, makes sense

PR #6106, opened 2021-04-01
fix identify Azureus style without trailing

SeaHOH:
There are some mistakes in #6098, fixed here.

### DIFF
--- a/src/identify_client.cpp
+++ b/src/identify_client.cpp
@@ -217,7 +217,6 @@ namespace {
 		, {"SD", "Xunlei"}
 		, {"SK", "spark"}
 		, {"SN", "ShareNet"}
-		, {"SP", "BitSpirit"}
 		, {"SS", "SwarmScope"}
 		, {"ST", "SymTorrent"}
 		, {"SZ", "Shareaza"}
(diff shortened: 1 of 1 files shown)


src/identify_client.cpp
```diff
@@ -71,7 +71,9 @@ namespace {
 		if (id[0] != '-' || !is_print(char(id[1])) || (id[2] < '0')
 			|| (id[3] < '0') || (id[4] < '0')
 			|| (id[5] < '0') || (id[6] < '0')
-			|| id[7] != '-')
+			|| (id[7] != '-'
+				&& !std::equal(id.begin() + 1, id.begin() + 3, "FG")
+				&& !std::equal(id.begin() + 1, id.begin() + 3, "SP")))
```
arvidn, line 76: it doesn't quite seem appropriate to have this exception here. If people are sloppy with how they encode the Azureus-style, perhaps this function should just be more forgiving about it in general

arvidn:
I think this needs to be rebased on top of `RC_2_0` as well

src/identify_client.cpp
SeaHOH: Ok, "SP" should be moved, not "FG", I will fix it later.

arvidn: why does "FG" belong here?

SeaHOH: I am not good at English, it means this commit, not the next fix.

SeaHOH:
It has been rebased now.

arvidn (approved):

PR #6109, opened 2021-04-03
CMake: fix compile test on AppleClang

FranciscoPombal:
Upstream patch from https://github.com/microsoft/vcpkg/commit/eb419063843d0e370214307820589db8108f142c

Back in the saddle.


CMakeLists.txt
```diff
@@ -645,6 +645,9 @@ if (NOT Windows)
 	]=])
 	string(REPLACE "std::atomic<int>" "std::atomic<std::int64_t>" ATOMICS64_TEST_SOURCE "${ATOMICS_TEST_SOURCE}")
 
+	if(APPLE)
+		set(CMAKE_REQUIRED_FLAGS "-std=c++11")
```
arvidn, line 649: how does this coexist with `CXX_STANDARD`? it seems like the right way to do this is to set that macro instead.

FranciscoPombal: @arvidn The `CMAKE_REQUIRED_FLAGS` variable is only used in the `Check.*Compiles` modules, it does not affect the actual project targets at all: https://cmake.org/cmake/help/latest/search.html?q=CMAKE_REQUIRED_FLAGS. Granted, its name is unfortunate.

arvidn: I see. So check compiles may say yes, even though it fails with the configuration the target is built with, and vice versa.

FranciscoPombal: In this case, there is no circumstance in which the compile check succeeds but the actual compilation of libtorrent fails, since `std` >= 11 is a requirement.

arvidn: perhaps, although there are examples of code that's valid in older versions of C++ and invalid in newer versions. However, I mostly had other build configuration features in mind. Say, windows version or MacOS version for example.

PR #6121, opened 2021-04-09
WebTorrent: Update libdatachannel

paullouisageneau:
This PR updates libdatachannel to v0.12.1. One notable change in v0.12 impacting the use case here is the generation of ECDSA certificates instead of RSA ones by default, which reduces CPU load. There are also a few compatibility fixes.


arvidn:
there is this warning, which is treated as an error in CI. I'm open to suggestions for how to not treat warnings as errors in dependencies only. But you might also want to fix this:

https://ci.appveyor.com/project/arvidn/libtorrent/builds/38640040/job/0kvpg9g7xtyu6hxk#L564

paullouisageneau:
@arvidn Indeed, it should be fixed now.

markmdscott:
> This PR updates libdatachannel to v0.12.1. One notable change in v0.12 impacting the use case here is the generation of ECDSA certificates instead of RSA ones by default, which reduces CPU load. There are also a few compatibility fixes.

Paul-Louis, why don't you use EdDSA instead of ECDSA?

Better for security and performance.

paullouisageneau:
> Paul-Louis, why don't you use EdDSA instead of ECDSA?
> 
> Better for security and performance.

The reason is that ECDSA is specified for certificates in the WebRTC standard, while EdDSA is not, probably because its standardization is a bit too recent. Ed25519 could be added as an option though.

arvidn (approved):

PR #6123, opened 2021-04-09
report deprecation warnings in python bindings

arvidn:


AllSeeingEyeTolledEweSew:
Awesome!

Note that python can be configured so warnings raise an exception instead of just printing a message. So `PyErr_WarnEx` returns `int` to indicate whether it actually threw an exception. It's probably best to check the return value and return early.

You can use `python -W error` to have warnings raise exceptions to test the behavior.

AllSeeingEyeTolledEweSew:
We should also add to `test.py` for this. You can use `unittest.TestCase.assertWarns`

bindings/python/client.py
```diff
@@ -301,9 +301,9 @@ def main():
                     out = '\n'
                     fp = h.file_progress()
                     ti = t.torrent_file
-                    for f, p in zip(ti.files(), fp):
+                    for idx, p in zip(range(ti.num_files()), fp):
```
AllSeeingEyeTolledEweSew, line 304: Assuming `len(h.file_progress()) == ti.num_files()` always, then this is simpler:
```python
for idx, p in enumerate(fp):
```

AllSeeingEyeTolledEweSew:
:+1:

Issue #5988, opened 2021-02-18
Python bindings: deprecate some usage of str


AllSeeingEyeTolledEweSew:
Currently, the python bindings generally allow `str` as input where `std::string` or `char *` is expected internally. It attempts to convert `str` to `bytes` with the default encoding.

In some cases, I think this does more harm than good. I think we should deprecate this functionality (and fire `DeprecationWarning`).

* `bencode(...)` should not allow `str` in its input, including `dict` keys: this causes unstable round-trips (`v != bdecode(bencode(v))`) and I can't think of any benefit from it
* `torrent_info({"info": ...})` should not allow `str` in its input: as above
* `torrent_handle.add_piece(i, "value")` should not accept `str`: any use of this is likely unintentional
* `generate_fingerprint()` should be deprecated, in favor of a version that returns `bytes`: I think it's sensible to accept `str` here, but the return value is intended for use in binary protocols, and client code should handle it as such

I'll add more as I find them.

AllSeeingEyeTolledEweSew:
There's a solid argument that lots of `str` accessors should be deprecated, like `file_storage.file_path()` or even `torrent_info.name()`. These functions access encoding-agnostic byte strings, which may be created with unknown encodings on other systems, so *it's usually a mistake to access them as `str`*

However *there are cases where `str` access is appropriate*, such as an app creating `torrent_info` with synthetic data and filenames (*not* sourced from the filesystem). If the data was sourced from python `str`, it's totally fair to access it as `str`. Deprecating `str` access would be pretty unfriendly to such a case.

I am on the fence about whether we should be unfriendly to a fringe use case, in favor of preventing the huge majority of use cases from doing something wrong.

arvidn:
when libtorrent loads a torrent, it sanitizes file paths/names as well as comment and creator strings. They *should* be valid UTF-8 strings (possibly with replacement characters for invalid sequences).

Is this not your experience?

Perhaps I'm a bit too ignorant of python, but `str` is always a unicode string, right? or is there a new type for that?

I agree that `sha1_hash()`, `add_piece()` and `sha1_hash.to_string()` should not use `str`.

How come `str` inputs to `bencode()` causes round-trip issues? You mean because when you decode it, all strings are bytes?
I'm not sure I'm convinced of that. If you have a unicode string it seems convenient to not have to encode it explicitly.

AllSeeingEyeTolledEweSew:
[...]

PR #6125, opened 2021-04-10
deprecate invalid inputs to bencode()

arvidn:
everything but int, list, dict, string, bytes and tuples (tuples are for preformatted buffers)

### DIFF
--- a/bindings/python/src/entry.cpp
+++ b/bindings/python/src/entry.cpp
@@ -158,6 +158,13 @@ struct entry_from_python
 
             return entry(preformatted);
         }
+        else
+        {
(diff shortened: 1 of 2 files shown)


AllSeeingEyeTolledEweSew:
Looks pretty good, and tests look good. Still think `str` should be deprecated, see #5988

arvidn:
you might be right about `str`. I think it's less urgent than these, and maybe it can be deprecated in 2.0.x instead of 1.2.x.
I think I'll land this for now.

What kind of test do you envision for the `str` patch? that all valid "entry" objects round-trip through `bencode()` -> `bdecode()`?

Issue #5995, opened 2021-02-18
Python bindings: various apparent gaps in type signature mappings


AllSeeingEyeTolledEweSew:
There are a few functions in the python bindings that seem like they're intended to work given the source code, but invoking the bindings doesn't work or doesn't produce the expected result.

* `create_torrent.set_comment(b"bytes arg")`: does not match a type signature, but probably should
* `create_torrent.set_creator(b"bytes arg")`: does not match a type signature, but probably should
* `create_torrent.add_url_seed("str arg")`: produces unexpected results in output of `create_torrent.generate()`. maybe the input got mapped incorrectly?
* `create_torrent.add_http_seed("str arg")`: produces unexpected results in output of `create_torrent.generate()`. maybe the input got mapped incorrectly?
* `create_torrent.add_collection("str arg")`: produces unexpected results in output of `create_torrent.generate()`. maybe the input got mapped incorrectly?
* `create_torrent.set_root_cert("str arg")`: produces unexpected results in output of `create_torrent.generate()`. maybe the input got mapped incorrectly? I'm not sure if this function should be deprecated, or if it should be allowed, so as to accept PEM data as a `str`.
* `fingerprint.name`: accessing the attribute raises `TypeError`
* `ip_filter.export_filter()`: raises `TypeError`
* `torrent_info(bytearray(bencode({b"info": {...}})))`: raises `RuntimeError("torrent file is not a dictionary")`, though the source code implies this should work
* `torrent_info(memoryview(bencode({b"info": {...}})))`: raises `RuntimeError("torrent file is not a dictionary")`, though the source code implies this should work
* `torrent_info(info_hash_t(sha1_hash(b"a" * 20)))`: raises `RuntimeError("torrent file is not a dictionary")`, though this should probably work
* `torrent_info.add_tracker("str arg")` does not match a type signature, though source code indicates it should allow the other arguments to be defaults
* `torrent_info.add_url_seed("str arg")` does not match a type signature, though source code indicates it should allow the other arguments to be defaults
* `torrent_info.add_url_seed("str arg", "str arg", [])` does not match a type signature, though source code indicates it should work
* `torrent_info.add_http_seed("str arg")` does not match a type signature, though source code indicates it should allow the other arguments to be defaults
* `torrent_info.add_http_seed("str arg", "str arg", [])` does not match a type signature, though source code indicates it should work
* `torrent_info.file_at_offset(0)` raises `TypeError`

I'll add more as I find them.

[...]

PR #6126, opened 2021-04-10
fix bug in python converter from PyUnicode -> string_view

arvidn:


AllSeeingEyeTolledEweSew:
Could you add a bit more to `test.py`? I'm not totally sure which items from #5995 this is meant to fix

arvidn:
the one new test that caught this bug was [this](https://github.com/arvidn/libtorrent/pull/6126/files#diff-b81174e110bef18e29b0788e332ae9574f6353b58a1cc97cdfbb072b8c4ab0d4R64), which you demonstrated.

It should cover all functions that take a `string_view. It doesn't affect *returning* a `string_view`, that case should work fine already.

arvidn:
as far as I can tell, boost.python (by default) accepts `bytes` -> `std::string` conversion. Which means the overloads of the `torrent_info` constructor fail. Even when passing in `bytes`, it's interpreted as a filename, even though it's supposed to be interpreted as a bencoded buffer.

I'm adding another commit to this PR to improve handling of `string_view` to not mimic this behavior as well an attempt to convert `PyUnicode` -> `std::string`, but that doesn't seem to get hit in the normal `std::string` conversion path.

arvidn:
I think this is in a good shape as it is now and I intend to land it soon.

bindings/python/src/sha1_hash.cpp
```diff
@@ -33,7 +33,7 @@ void bind_sha1_hash()
         .def(init<std::string>())
         .def("clear", &sha1_hash::clear)
         .def("is_all_zeros", &sha1_hash::is_all_zeros)
-        .def("to_string", &sha1_hash::to_string)
+        .def("to_string", sha1_hash_bytes)
```
AllSeeingEyeTolledEweSew, line 36: This would break any current callers, which is probably fine since it rarely worked. But why not just delete it instead?

AllSeeingEyeTolledEweSew:
Otherwise seems good

bindings/python/src/sha1_hash.cpp
arvidn: I think leaving it in preserves fidelity with the C++ interface. since there isn't really any python specific API documentation, staying similar to C++ is helpful.

both removing and changing the return type would break current callers.

AllSeeingEyeTolledEweSew: I see.

If I said I had some thoughts on "documenting" the python API in the future, would it sway you? :) This is a minor point and I'm not willing to die on this hill, but I'm compelled to _try_, since the upside is avoiding extraneous APIs.

Check this out:

https://github.com/AllSeeingEyeTolledEweSew/tvaf/blob/master/typeshed/libtorrent.pyi

Modern python (3.6+) supports type annotations and type checking. For extensions like libtorrent, you can write "type stub" files which just have type annotations. They're header files for python.

I wrote a local `libtorrent.pyi` (against RC_1_2) to help with type-checking my own code. My end goal is to have it incorporated into libtorrent.

But in using type-checked python, I have found it extremely useful to treat type stubs _as documentation_. The type stubs of the standard libraries are an excellent quick reference, and half the time I just refer to those instead of referring to the official documentation pages.

I don't expect libtorrent's python bindings will ever get official documentation, but if they included official type stubs I would strongly argue we don't need strong fidelity with the C++ interface.

arvidn: > but if they included official type stubs I would strongly argue we don't need strong fidelity with the C++ interface.

sounds reasonable

PR #6130, opened 2021-04-11
remove python binding for (deprecated) fingerprint::name member

arvidn:
since it's broken right now. There's no converter from a plain C char-array


AllSeeingEyeTolledEweSew:
:+1:

PR #6147, opened 2021-04-15
info_hash related fixes

arvidn:
add `sha256_hash` type
add `__hash__` member to `info_hash_t`
add `torrent_info` constructor overload for a `sha256_hash`

@AllSeeingEyeTolledEweSew looks good?

### DIFF
--- a/bindings/python/src/sha256_hash.cpp
+++ b/bindings/python/src/sha256_hash.cpp
@@ -12,35 +12,33 @@ namespace {
 
 using namespace lt;
 
-long get_hash(sha1_hash const& s)
+long get_hash(sha256_hash const& s)
 {
-    return std::hash<sha1_hash>{}(s);
+    return std::hash<sha256_hash>{}(s);
 }
 
-bytes sha1_hash_bytes(const sha1_hash& bn) {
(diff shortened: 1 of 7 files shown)


bindings/python/src/sha256_hash.cpp
```diff
@@ -0,0 +1,46 @@
...
+        .def(self_ns::str(self))
+        .def(init<std::string>())
+        .def("clear", &sha256_hash::clear)
+        .def("is_all_zeros", &sha256_hash::is_all_zeros)
+        .def("to_string", sha256_hash_bytes)
+        .def("__hash__", get_hash)
+        .def("to_bytes", sha256_hash_bytes)
+        ;
+
+    scope().attr("peer_id") = scope().attr("sha256_hash");
```
AllSeeingEyeTolledEweSew, line 44: Does this mean `libtorrent_.peer_id` now points to `sha256_hash` instead of `sha1_hash`? I think this makes this a breaking change. Looks like it's still `sha1_hash` in C++ AFAICT.

arvidn: oops. that was not meant to make it in there

PR #6149, opened 2021-04-15
add converter for dht_lookup

arvidn:


AllSeeingEyeTolledEweSew:
Looks good :+1:

bindings/python/test.py
```diff
@@ -737,6 +737,21 @@ def test_add_torrent(self):
...
+        try:
+            st = s.status()
+            print(st)
+            print(st.active_requests)
+            print(st.dht_nodes)
+            print(st.dht_node_cache)
+            print(st.dht_torrents)
+            print(st.dht_global_nodes)
+            print(st.dht_total_allocations)
+        except AttributeError:
```
AllSeeingEyeTolledEweSew, line 751: This is trouble. It doesn't actually test whether the attributes are bound. If the binding is broken as it was before, then a `TypeError` would get raised here, but this still weakens the test.

One way to solve this would be to expose compile-time flags in python. But that rabbit hole goes deep. The tests get a lot more complicated (should the tests also gate based on `TORRENT_DISABLE_DHT`? etc).

There's generally broad support in python for modules/extensions having a variable "feature set" where attributes may exist or not. Generally, either a whole module exists or it doesn't. The main exception is `sys.version_info` and `sys.platform`. In particular, the type checker mypy [*only* supports those variables for conditional type declarations](https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks). (It supports conditionals based on types like `if isinstance(...)`, but other runtime value checks like `if value == 3` only work with a narrow set). I'm not saying the python bindings should *only* support a single configuration, but we should be aware that multiple configurations is kind of an impedence mismatch.

Personally I think that, at least for now, all python code including the tests should just assume the "pypi configuration", with everything enabled including deprecated functions. CI should only run the tests in this configuration.

Thoughts?

AllSeeingEyeTolledEweSew (changes requested):

bindings/python/test.py
arvidn: well, the missing `.def(init<>())` was only discovered when building the binding without deprecated functions. But generally I think it's reasonable to always assume (recently) deprecated functions are there.

PR #6153, opened 2021-04-17
fix issue creating and loading v2 torrents with empty files

arvidn:


arvidn:
this changes the pad-file algorithm to add pad-files *after* files rather than before them. This leads to tail-padding. It's what the reference implementation does, so this is better in that it's likely to be compatible with other clients.

In order to be backwards compatible with libtorrent 2.0.0-2.0.3, there's some extra logic to also accept hybrid torrents without tail-padding.

PR #6154, opened 2021-04-17
fix (deprecated) use of add_torrent_params::info_hash

arvidn:


AllSeeingEyeTolledEweSew:
The following segfaults:
```python
session.add_torrent({"info_hash": b"a" * 20, "info_hashes": b"a" * 32})
```

The `{"info_hash": ..., "info_hashes": ...}` form is typical of the output of `parse_magnet_uri_dict("magnet:?xt=urn:btmh:1220...")`

Sorry I did apparently a bad job of testing the other PRs you've pointed me to. I cherry-picked this one on top of the test suite I'm building in #5990 and ran all my tests there. This crash was the only error, otherwise it fixed the case it was intended to fix

arvidn:
if you have unit tests, would you mind submitting a PR with them? Even if they fail, it would make it easier for me to fix the underlying issues.

arvidn:
that segfault you mention, was it introduced by this PR? I would think the issue is that constructing an `info_hash_t` from bytes doesn't work (yet). I'm not sure it makes sense for it to work either, would it default to interpret it as sha256 or sha1? should it depend on the length maybe?
I don't think there's a C++ counterpart for this constructor.

arvidn:
I cannot reproduce the segfault

arvidn:
the issue with adding a torrent with the same `sha1-hash` as `sha256-hash` is an issue in the `torrent_list`, which assumes that those will have different values, and fail to maintain its invariant when they hash to the same value.

I'll fix that in a separate PR

AllSeeingEyeTolledEweSew:
> if you have unit tests, would you mind submitting a PR with them? Even if they fail, it would make it easier for me to fix the underlying issues.

I'm staging such a PR in #5990. It made sense to develop as one big change, but I might try to split it up for submission.

> that segfault you mention, was it introduced by this PR? I would think the issue is that constructing an `info_hash_t` from bytes doesn't work (yet). I'm not sure it makes sense for it to work either, would it default to interpret it as sha256 or sha1? should it depend on the length maybe?
> I don't think there's a C++ counterpart for this constructor.

> I cannot reproduce the segfault

I forgot to say that #5990 is on master. So I cherry-pick #6154 onto master, and got the segfault there.

In particular, this segfaulted:

```python
session.add_torrent(parse_magnet_uri_dict("magnet:?xt=urn:btmh:1220..."))
```

Without #6154 it doesn't segfault.

arvidn:
segfault should be fixed here: https://github.com/arvidn/libtorrent/pull/6159

Issue #5891, opened 2021-01-21
set_piece_deadline causes major speed regressions


jagannatharjun:
**Please provide the following information**

libtorrent version (or branch):
libtorrent 1.2.12

platform/architecture:
Windows x64

compiler and compiler version:
default CMake, msvc

please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

I am trying to implement an HTTP file serve on the top of libtorrent in qbittorrent but using set_piece_deadline is causing major speed regression while using sequential download doesn't cause any such regressions. I set-piece a deadline in the next 32 MiB file from the current range. but I am noticing very large speed drops because of it.
If interested here is the code https://github.com/jagannatharjun/qBittorrent/blob/stream-final/src/base/streaming/streamfile.cpp#L175

arvidn:
the current algorithm for piece deadline does unfortunately not scale very well with many peers and/or small pieces. This is something I hope to have time to improve by keeping more state in between piece-picks, similar to the regular piece picker, but optimized for the deadline ordering, rather than rarest first.

arvidn:
fewer peers and larger pieces help

master255:
@jagannatharjun I think your algorithm should be much more complex. Study in more detail how it works.

jagannatharjun:
using larger deadline time seems to improve the situation a bit but this increases wait times on the receiver side.

master255:
@jagannatharjun 

> I think your algorithm should be much more complex.

jagannatharjun:
>  improve by keeping more state in between piece-picks, similar to the regular piece picker, but optimized for the deadline ordering, rather than rarest first.

if someone wanted to work on this, can you point a good starting point?

elgatito:
@arvidn Also encountered this. 

I test on a highly crowded torrent. 
Downloading 75 pieces, 4MB each (300MB in total). 
`strict_end_game_mode = true`.

If I just use priorities to mark which pieces to download - I got these times:
```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 1544M  100 1544M    0     0  8222k      0  0:03:12  0:03:12 --:--:-- 13.7M
```

When I call `set_piece_deadline` on the currently read piece and the next one, then I get this:
```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
[...]

PR #6156, opened 2021-04-19
removed queue size limiter for time critical mode.

elgatito:
Does think make sense? 

When we have time critical pieces - other pieces are skipped inside `request_blocks` and queue is also limited.

Related to #5891

### DIFF
--- a/src/request_blocks.cpp
+++ b/src/request_blocks.cpp
@@ -83,8 +83,7 @@ namespace libtorrent {
 
 		// in time critical mode, only have 1 outstanding request at a time
 		// via normal requests
-		int const desired_queue_size = time_critical_mode
-			? 1 : c.desired_queue_size();
+		int const desired_queue_size = c.desired_queue_size();
 
 		int num_requests = desired_queue_size
 			- int(c.download_queue().size())
@@ -206,8 +205,7 @@ namespace libtorrent {
 		bool const dont_pick_busy_blocks
 			= ((ses.settings().get_bool(settings_pack::strict_end_game_mode)
 				&& p.get_download_queue_size() < p.num_want_left())
-				|| dq.size() + rq.size() > 0)
-				&& !time_critical_mode;
+				|| dq.size() + rq.size() > 0);
 
 		// this is filled with an interesting piece
 		// that some other peer is currently downloading
@@ -216,13 +214,6 @@ namespace libtorrent {
 		for (piece_block const& pb : interesting_pieces)
 		{
 			if (prefer_contiguous_blocks == 0 && num_requests <= 0) break;
-
-			if (time_critical_mode && p.piece_priority(pb.piece_index) != top_priority)
-			{
-				// assume the subsequent pieces are not prio 7 and
-				// be done
-				break;
-			}
 
 			int num_block_requests = p.num_peers(pb);
 			if (num_block_requests > 0)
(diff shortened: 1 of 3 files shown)


arvidn:
looks reasonable. did you experiment with setting it to half, or some other size proportional to the normal queue length, rather than 1?

elgatito:
@arvidn 

> looks reasonable. did you experiment with setting it to half, or some other size proportional to the normal queue length, rather than 1?

No, not yet, was also thinking about that.
I have noticed, when new time critical piece is added in `torrent.cpp`, it's blocks are moved to the beginning of the queue, so thought it would be picked anyway, when there is free space in the queue.

elgatito:
@arvidn I have tested with queue / 2, but it looks pretty much like when it was `1`. 
Instead, I have noticed that fix is not helping without changes from #6157. I will try more tests.

elgatito:
@arvidn I have tested it several time in different combinations.

Two things make good download speeds:
1) In `request_blocks`:
```
		int const desired_queue_size = time_critical_mode
			? 1 : c.desired_queue_size();
```
change to 
```
		int const desired_queue_size = c.desired_queue_size();
```

2) In `peer_connection`:
Comment this block:
```
		if (t->num_time_critical_pieces() > 0)
		{
		 	ret |= piece_picker::time_critical_mode;
		}
```

Changing only one thing does not help to speed up at all.
Probably I had good result because of using own `pick_pieces` code, which had modified time critical mode clauses.

elgatito:
@arvidn I have removed usage of `time_critical_mode` from `peer_connection`, `piece_picker::pick_pieces` and `request_blocks`. 
Looks better regarding speed.

My usual test with forcing few piece with deadlines show this:
```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 1544M  100 1544M    0     0  8211k      0  0:03:12  0:03:12 --:--:-- 27.5M
```

src/request_blocks.cpp
```diff
@@ -83,8 +83,7 @@ namespace libtorrent {
 
 		// in time critical mode, only have 1 outstanding request at a time
 		// via normal requests
-		int const desired_queue_size = time_critical_mode
-			? 1 : c.desired_queue_size();
+		int const desired_queue_size = c.desired_queue_size();
```
arvidn, line 86: does it still work if you make this use `c.desired_queue_size() / 2` in `time_critical_mode`? If so, I think that would be a bit safer

src/request_blocks.cpp
```diff
@@ -206,8 +205,7 @@ namespace libtorrent {
 		bool const dont_pick_busy_blocks
 			= ((ses.settings().get_bool(settings_pack::strict_end_game_mode)
 				&& p.get_download_queue_size() < p.num_want_left())
-				|| dq.size() + rq.size() > 0)
-				&& !time_critical_mode;
+				|| dq.size() + rq.size() > 0);
```
arvidn, line 208: this change isn't obvious. Do you measure a difference in your tests with this change and without it?

arvidn:
sorry for the delay. I think this is looking pretty good. Just two questions

jagannatharjun:
don't work well with torrents with pieces of small size otherwise, I can see nice improvements. 👍

arvidn:
I rebased this onto `RC_2_0`, and fixed the test here: https://github.com/arvidn/libtorrent/pull/6262

elgatito:
@arvidn looks nice, I see you've added some simplifications for loops as well.
Should I perform some testing for this PR in 1_2 ?

PR #5996, opened 2021-02-18
Adding sequential_start to piece_picker to speed up sequential downloads

markmdscott:
PR includes Oleg's changes rebased on master as a first step. Ready to work on any additional changes as necessary.

PR #6157, opened 2021-04-19
Piece picker changes with priority based selector

elgatito:
PR does these changes:
- New setting in `settings_pack` that enables priority-based piece selection.
- New `piece_picker::priority_order` bit to transfer configuration to `piece_picker`.

In `peer_connection::picker_options()`:
- `piece_picker::priority_order` placed above `piece_picker::sequential`, because basically priority_order is the same, but first sorted with priority.

In `piece_picker::pick_pieces`:
- `if (options & time_critical_mode)` check is moved above, to make sure time critical pieces are picked before any other selector mode.
- Other conditions starting with `if`, instead of `else if`, because all time critical pieces can be already requested, usually it is not much of them.
- `for (auto i = m_pieces.begin(); i != m_pieces.end() && piece_priority(*i) == top_priority; ++i)` - this was leaving the loop if first piece is not `top_priority`, no?
- Added new functionality to iterate pieces with specific priorities. Not sure it there is a better (faster) way of that, to keep the order.

What do you think?

Related to #5891

### DIFF
--- a/src/piece_picker.cpp
+++ b/src/piece_picker.cpp
@@ -119,6 +119,7 @@ namespace libtorrent {
 	constexpr picker_options_t piece_picker::time_critical_mode;
 	constexpr picker_options_t piece_picker::align_expanded_pieces;
 	constexpr picker_options_t piece_picker::piece_extent_affinity;
+	constexpr picker_options_t piece_picker::priority_order;
 
 	constexpr download_queue_t piece_picker::piece_pos::piece_downloading;
 	constexpr download_queue_t piece_picker::piece_pos::piece_full;
@@ -2246,6 +2247,29 @@ namespace {
 					, prefer_contiguous_blocks, peer, suggested_pieces
 					, options);
 				if (num_blocks <= 0) return ret;
+			}
+		}
+		else if (options & priority_order)
+		{
+			if (m_dirty) update_pieces();
+			TORRENT_ASSERT(!m_dirty);
+
+			for (uint8_t prio = 7; prio > 0; --prio)
+			{
+				for (auto i : m_pieces)
+				{
+					if (piece_priority(i) != download_priority_t(prio)) continue;
+					if (!is_piece_free(i, pieces)) continue;
+
+					ret |= picker_log_alert::random_pieces;
+
+					num_blocks = add_blocks(i, pieces
+						, interesting_blocks, backup_blocks
+						, backup_blocks2, num_blocks
+						, prefer_contiguous_blocks, peer, suggested_pieces
+						, options);
+					if (num_blocks <= 0) return ret;
+				}
 			}
 		}
 		else
--- a/include/libtorrent/piece_picker.hpp
+++ b/include/libtorrent/piece_picker.hpp
@@ -151,6 +151,9 @@ namespace libtorrent {
 		// small)
 		static constexpr picker_options_t piece_extent_affinity = 7_bit;
 
+		// pick pieces in priority descending order
+		static constexpr picker_options_t priority_order = 8_bit;
+		
 		struct downloading_piece
 		{
 			downloading_piece()
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -872,6 +872,10 @@ namespace aux {
 			// happen at the application level.
 			allow_idna,
 
+			// when enabled, pieces are picked, based on their priority value. 
+			// Same priorities will follow sequential order.
+			piece_priority_order,
+
 			max_bool_setting_internal
 		};
 
--- a/src/peer_connection.cpp
+++ b/src/peer_connection.cpp
@@ -921,7 +921,11 @@ namespace libtorrent {
 			ret |= piece_picker::time_critical_mode;
 		}
 
-		if (t->is_sequential_download())
+		if (m_settings.get_bool(settings_pack::piece_priority_order))
+		{
+			ret |= piece_picker::priority_order;
+		}
+		else if (t->is_sequential_download())
 		{
 			ret |= piece_picker::sequential;
 		}
--- a/src/settings_pack.cpp
+++ b/src/settings_pack.cpp
@@ -213,6 +213,7 @@ constexpr int CLOSE_FILE_INTERVAL = 0;
 		SET(validate_https_trackers, true, &session_impl::update_validate_https),
 		SET(ssrf_mitigation, true, nullptr),
 		SET(allow_idna, false, nullptr),
+		SET(piece_priority_order, false, nullptr),
 	}});
 
 	aux::array<int_setting_entry_t, settings_pack::num_int_settings> const int_settings


elgatito:
@arvidn 

```
src/piece_picker.cpp:2164:44: note: candidate: ‘operator>(int, int)’ <built-in>
 2164 |    for (download_priority_t prio = 7; prio > 0; --prio)
      |                                       ~~~~~^~~
src/piece_picker.cpp:2164:44: note:   no known conversion for argument 1 from ‘libtorrent::download_priority_t’ {aka ‘libtorrent::aux::strong_typedef<unsigned char, libtorrent::download_priority_tag>’} to ‘int’
```

How can I check that locally? Gcc-6 is not complaining on that..

arvidn:
I'm not convinced the piece picker flag `time_critical_mode` is useful. I get the impression you've misunderstood what it's supposed to do. What it's doing is to *prevent* piece from being picked. As far as I can tell, you're changing that behavior. That's probably why the tests are failing.

I would think removing this flag and all associated logic would make more sense,

elgatito:
@arvidn 

> I'm not convinced the piece picker flag time_critical_mode is useful. I get the impression you've misunderstood what it's supposed to do. What it's doing is to prevent piece from being picked.

In my understanding, when piece is marked a time critical, it should be picked, if possible. So when `piece_picker::pick_pieces` is called and there is time critical piece to get - it should be got, and it there is space for other pieces - we should also take them. 
That is wrong understanding?

arvidn:
> In my understanding, when piece is marked a time critical, it should be picked, if possible. So when piece_picker::pick_pieces is called and there is time critical piece to get - it should be got, and it there is space for other pieces - we should also take them.
> That is wrong understanding?

in the context of piece picker flags; yes, that is a misunderstanding. That flag is intended to prevent the piece picker from interfering with the time-critical piece picker.

elgatito:
@arvidn 
> I would think removing this flag and all associated logic would make more sense,

So what then you mean with that? Remove time critical options from `piece_picker::pick_pieces` and `peer_connection::picker_options`?

markmdscott:
> I'm not convinced the piece picker flag `time_critical_mode` is useful. I get the impression you've misunderstood what it's supposed to do. What it's doing is to _prevent_ piece from being picked. As far as I can tell, you're changing that behavior. That's probably why the tests are failing.
> 
> I would think removing this flag and all associated logic would make more sense,

Arvid you've said in the past that time critical mode was added as part of the streaming (time critical deadlines) piece picker logic. Without this mode what becomes of the streaming piece picking implementation?

markmdscott:
> I'm not convinced the piece picker flag `time_critical_mode` is useful. I get the impression you've misunderstood what it's supposed to do. What it's doing is to _prevent_ piece from being picked. As far as I can tell, you're changing that behavior. That's probably why the tests are failing.
> 
> I would think removing this flag and all associated logic would make more sense,

Arvid you've said in the past that time critical mode was added as part of the streaming (time critical deadlines) piece picker logic. Without this mode what becomes of the streaming piece picking implementation?

arvidn:
> Arvid you've said in the past that time critical mode was added as part of the streaming (time critical deadlines) piece picker logic. Without this mode what becomes of the streaming piece picking implementation?

I'm assuming that by "this mode" you mean the `piece_picker::time_critical_mode` flag that can be passed to `piece_picker`.

Without that mode, peers that are not considered the best peers will still be allowed to pick rarest-first pieces (to the same extent as they normally would) when there are time-critical pieces. i.e. pieces with a deadline set on them.

the streaming piece picking isn't affected, this is a mode of the *normal* piece picker, for picking rarest first pieces.

arvidn:
> So what then you mean with that? Remove time critical options from piece_picker::pick_pieces and peer_connection::picker_options?

Yes, that's what I mean.

elgatito:
@arvidn I have removed everything related to `time_critical_mode` from this PR. 
This one is only adding a priority_based selector to piece_picker.

src/piece_picker.cpp
```diff
@@ -2248,6 +2249,29 @@ namespace {
...
 			}
 		}
+		else if (options & priority_order)
+		{
+			if (m_dirty) update_pieces();
+			TORRENT_ASSERT(!m_dirty);
+
+			for (uint8_t prio = 7; prio > 0; --prio)
+			{
+				for (auto i : m_pieces)
```
arvidn, line 2259: this is extremely inefficient. To the point where your download speed may be CPU bound. This is not the way to do this.

If you think priorities should strictly take precedence over rarity, change the order in the priority boundary to put all prio-7 pieces first, then prio-6 and so on.
The main challenge with that is that the availability of a piece is virtually unbounded, so you don't know what to multiply the priority by. You may have to set a cap to 100 or something.

markmdscott:
> > Arvid you've said in the past that time critical mode was added as part of the streaming (time critical deadlines) piece picker logic. Without this mode what becomes of the streaming piece picking implementation?
> 
> I'm assuming that by "this mode" you mean the `piece_picker::time_critical_mode` flag that can be passed to `piece_picker`.
> 
> Without that mode, peers that are not considered the best peers will still be allowed to pick rarest-first pieces (to the same extent as they normally would) when there are time-critical pieces. i.e. pieces with a deadline set on them.
> 
> the streaming piece picking isn't affected, this is a mode of the _normal_ piece picker, for picking rarest first pieces.

Ok that's odd, for if time critical mode was added as part of the streaming implementation - as you have affirmed[^](https://github.com/arvidn/libtorrent/pull/5996#issuecomment-785136372) - then *by definition* it cannot be a mode of the normal piece picker! How do you square you past and present statements?

arvidn:
> Ok that's odd, for if time critical mode was added as part of the streaming implementation - as you have affirmed^ - then by definition it cannot be a mode of the normal piece picker!

Why not?

> How do you square you past and present statements?

Both are true. Assuming "time critical mode" refers to the `time_critical_mode` flag to the normal `piece_picker`. Also assuming "it" in your sentence refers to that same flag. However, if "it" refers to the streaming implementation as a whole, then you're right. The streaming implementation has a few parts to it, and this flag to `piece_picker` is one of those parts, it's not all of it.

src/piece_picker.cpp
elgatito: > If you think priorities should strictly take precedence over rarity, change the order in the priority boundary to put all prio-7 pieces first, then prio-6 and so on.

You mean to change  `m_priority_boundaries` fill-up logic? 
Trying to understand how it is done currently with rarest pick mode.

elgatito:
@arvidn If doing logic, like it is currently with `rarest_first`, then I do need to make own `m_priority_boundaries` and `m_pieces` that will reflect priority sorted `m_pieces` and it's `start/end` in `m_priority_boundaries`. Right?

markmdscott:
> @arvidn If doing logic, like it is currently with `rarest_first`, then I do need to make own `m_priority_boundaries` and `m_pieces` that will reflect priority sorted `m_pieces` and it's `start/end` in `m_priority_boundaries`. Right?

That would seem to be the case Denis.

Though thinking carefully, this streaming priority piece problem is similar yet somewhat different than the rarest first availability problem tackled by the existing piece picking algorithm and data structures. Rarest first is tougher in that it has to pick (and keep track) among all outstanding pieces, and all of them have to potentially be updated when peers join or leave. Whereas with streaming you only have to consider the next sequential meg or two's worth of pieces (~30 sec) which are critical for playback, and their priorities have to be updated whenever the user seeks the media. It is as though any peer you come across, even a seed, only has very few pieces you're actually interested in. Therefore, "In situations like these, it would be more efficient to simply build a list of all pieces [of interest] the peer has and order those by availability [priority]." This way, rather than recreating the same algorithm and data structures of rarest first, you can write a very simple streaming piece picking algorithm that complements the rarest first piece picker.

markmdscott:
> @arvidn If doing logic, like it is currently with `rarest_first`, then I do need to make own `m_priority_boundaries` and `m_pieces` that will reflect priority sorted `m_pieces` and it's `start/end` in `m_priority_boundaries`. Right?

That would seem to be the case @elgatito

But thinking carefully Denis, this streaming priority piece problem is similar yet somewhat different than the rarest first availability problem tackled by the existing piece picking algorithm and data structures. Rarest first is tougher in that it has to pick (and keep track) among all outstanding pieces, and all of them have to potentially be updated when peers join or leave. Whereas with streaming you really only have to consider the next sequential meg or two's worth of pieces (~30 sec) which are critical for playback, and their priorities have to be updated whenever the user seeks the media. It is as though any peer you come across, even a seed, only has very few pieces you're actually interested in. Therefore, "In situations like these, it would be more efficient to simply build a list of all pieces [of interest] the peer has and order those by availability [priority]." This way, rather than naively recreating/copying the same algorithm and data structures of rarest first, you can write a very simple streaming piece picking algorithm that complements the rarest first piece picker.

arvidn:
The current piece picker is essentially "one dimensional", in that it only considers rarity of pieces. The priorities are tacked on to this by recomputing the effective rarity based on true rarity and piece priority. one consequence of this is that a very rare piece with low priority may be considered before a very common piece with high priority. I think this is not intuitive, and most people would expect priority to trump rarity.

The logic to alter the effective rarity can be found [here](https://github.com/arvidn/libtorrent/blob/RC_2_0/include/libtorrent/piece_picker.hpp#L679).

However, to make priority always trump rarity, this mechanism isn't quite viable. At least not as far as I can see. The number of peers that can have a piece is unbounded, so if you combine availability and priority into a single dimension, the availability can always be higher. I think you would have to factor out the `m_priority_boundaries`, `m_pieces` members and have a copy per priority level. However, it's not so simple since `m_pieces` maps piece index to priority level, so there probably would still have to be just one of that, but the mapping would have to specify which `m_piece_boundary` to map into.

elgatito:
@arvidn That `priority` makes more sense.

```
return availability * int(priority_levels - piece_priority)
					* prio_factor + adjustment;
```

But here `prio_factor` is `3`, and `availability` can be easily > 10, `priority_levels - piece_priority = [1 .. 7]`. 

So with:
```
piece_priority = 1, availability = 3, result = 63
piece_priority = 1, availability = 10, result = 210
piece_priority = 1, availability = 20, result = 420

piece_priority = 3, availability = 3, result = 45
piece_priority = 3, availability = 10, result = 150
piece_priority = 3, availability = 20, result = 300

piece_priority = 5, availability = 3, result = 27
piece_priority = 5, availability = 10, result = 90
piece_priority = 5, availability = 20, result = 180

piece_priority = 7, availability = 3, result = 9
piece_priority = 7, availability = 10, result = 30
piece_priority = 7, availability = 20, result = 60

```

So it looks almost fine. 
But I have a feeling that usage of `availability` should be non-linear. 
When getting `availability  > 10` it is almost always a good sign and it you don't select the whole torrent for downloading, then you can prefer priority over rarity.

One more thing is to not multiply by `int(priority_levels - piece_priority) * prio_factor` but instead do ` pow(int(priority_levels - piece_priority), prio_factor)`. 
With that approach we get:
```
piece_priority = 1, availability = 3, result = 1029
piece_priority = 1, availability = 10, result = 3430
piece_priority = 1, availability = 20, result = 6860

piece_priority = 3, availability = 3, result = 375
piece_priority = 3, availability = 10, result = 1250
piece_priority = 3, availability = 20, result = 2500

piece_priority = 5, availability = 3, result = 81
piece_priority = 5, availability = 10, result = 270
piece_priority = 5, availability = 20, result = 540

piece_priority = 7, availability = 3, result = 3
piece_priority = 7, availability = 10, result = 10
piece_priority = 7, availability = 20, result = 20

```

Still not great.

elgatito:
When I do calculation of priority groups, to put N pieces into group with top priority, then increasing geometrically for each next group with less priority, I do get group size with:
```
inline int progression_result(int a, double r, int N) {
    // finding the Nth term in:
    // TN = a1 * r^(N-1)
    return int(a * (double)(pow(r, N - 1))); 
}
```

In my case `a = 2, r = 1.7`.

PR #6159, opened 2021-04-20
expand test for torrent_list. fix issue when adding a hybrid torrent

arvidn:
with a sh1-hash the same as truncated sha256 hash

### DIFF
--- a/bindings/python/src/session.cpp
+++ b/bindings/python/src/session.cpp
@@ -301,6 +301,10 @@ namespace
                     p.info_hashes = info_hash_t(sha256_hash(
                             bytes(extract<bytes>(value)).arr.data()));
                 }
+                else
+                {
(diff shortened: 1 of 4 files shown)


AllSeeingEyeTolledEweSew:
I tried cherry-picking this onto master to run my tests from #5990, but fd05f3be9039c31aee36cc381ef49902fe463eb1 didn't apply trivially. So I checked out this branch and ran my tests on it.

It looks like this fixes the segfault I was seeing earlier.

One other test fails on this branch:
```python
ti = torrent_info(...)
ti.set_web_seeds({"url": "http://example.com", "auth": ""})
"""
KeyError: 'type'
"""
```

AllSeeingEyeTolledEweSew:
(I don't yet have any tests for `sha256_hash`, like the one failing)

PR #6166, opened 2021-04-23
fix bindings for dht alert fields

arvidn:


AllSeeingEyeTolledEweSew:
* `dht_log_alert.module` still raises `TypeError: No to_python (by-value) converter found for C++ type: libtorrent::dht_log_alert::dht_module_t`
 * `dht_immutable_item_alert.target` works, but it worked before
 * `dht_immutable_item_alert.item` still raises `RuntimeError: invalid type requested from entry`
 * `dht_mutable_item_alert.key` and `dht_mutable_item_alert.signature` are fixed
 * `dht_put_alert.target` works, but it worked before
 * `dht_put_alert.public_key` and `dht_put_alert.signature` are fixed
 
I updated the checklist in #5995 with what works and doesn't

PR #6187, opened 2021-05-06
[8/?] Add type stubs to the python bindings

AllSeeingEyeTolledEweSew:
This adds type stubs to the python bindings. The type stubs are packaged with the python extension, so a user can just install `python-libtorrent` to type-check their code.

If the type stubs have errors, the remedy is for users to apply workarounds in their code (typically `# type: ignore` or `cast(...)`), and wait for a new version of libtorrent. Even the standard library's type stubs have errors and gaps, so this is a relatively common practice. When new code is released, mypy has a feature to warn on unused `# type: ignore` or `cast(...)`, so unused workarounds don't accumulate forever.

There currently isn't a way to "test" type stubs, or to measure coverage. But since we are type-checking the python bindings tests, those tests also now exercise the type stubs.

This adds a maintenance burden that the type stubs must be kept in sync with the python bindings. Type stubs can be generated from an extension with `stubgen`, but apparently its output from a boost.python project is very poor. I expect the best maintenance process for the type stubs is to ensure very good coverage of the python bindings *tests*; and type-checking the tests will uncover gaps in the type stubs.

## Packaging

mypy only recognizes type stubs *included in a python package*, not *adjacent to a module*.

This means I had to change `libtorrent.so` to `libtorrent/__init__.so`, and include type stub files in `libtorrent/`.

This is only effective for `setup.py`. So building dirctly with `b2` still builds a `libtorrent.so` in the b2 output directory; but `setup.py` (with both `--config-mode=b2` or `--config-mode=distutils`) will produce `libtorrent/__init__.so`, with type stubs alongside.

This change should be transparent for all usage in python. It may change some introspection output in some cases, but I don't think it changes anything that counts as a stable interface.

## On deprecating functions

Reading the type stubs, you can see various places where two functions do exactly the same thing but one is deprecated (e.g. `msg` vs `log_message()` on many alerts), just because a type changed on the C++ side.

You've indicated this is so the python bindings match the C++ code closely.

If the main motivation for this is that the C++ docs are the only docs, I ask that you consider my thought that type stubs serve double-duty as documentation. It may be easier for python users to read the type stubs to see available functions, rather than try to apply the C++ docs to python. This would make it easier to maintain a world where the names in python diverge from names in C++.

Personally, I would rather deal with names in python that don't quite match C++, rather than deal with deprecations which have no meaning in python.

### DIFF
--- a/bindings/python/client.py
+++ b/bindings/python/client.py
@@ -13,7 +13,9 @@ import time
 from typing import Any
 from typing import Dict
 from typing import List
+from typing import Mapping
 from typing import Optional
+from typing import Sequence
 from typing import Union
 
 import libtorrent as lt
@@ -99,7 +101,7 @@ def progress_bar(progress: float, width: int) -> str:
     return progress_chars * "#" + (width - progress_chars) * "-"
 
 
-def print_peer_info(console: ConsoleType, peers: List[Any]) -> None:
+def print_peer_info(console: ConsoleType, peers: List[lt.peer_info]) -> None:
 
     out = " down    (total )   up     (total )" "  q  r flags  block progress  client\n"
 
@@ -132,7 +134,10 @@ def print_peer_info(console: ConsoleType, peers: List[Any]) -> None:
         elif p.flags & lt.peer_info.connecting:
             id = "connecting to peer"
         else:
-            id = p.client
+            try:
+                id = p.client.decode()
+            except ValueError:
+                id = p.client.hex()
 
         out += "%s\n" % id[:10]
 
@@ -140,7 +145,7 @@ def print_peer_info(console: ConsoleType, peers: List[Any]) -> None:
 
 
 def print_download_queue(
-    console: ConsoleType, download_queue: List[Dict[str, Any]]
+    console: ConsoleType, download_queue: Sequence[Mapping[str, Any]]
 ) -> None:
 
     out = ""
@@ -342,9 +347,12 @@ def main() -> None:
                     out = "\n"
                     fp = h.file_progress()
                     ti = t.torrent_file
-                    for idx, p in enumerate(fp):
-                        out += progress_bar(p / float(ti.files().file_size(idx)), 20)
-                        out += " " + ti.files().file_path(idx) + "\n"
+                    if ti is not None:
+                        for idx, p in enumerate(fp):
+                            out += progress_bar(
+                                p / float(ti.files().file_size(idx)), 20
+                            )
+                            out += " " + ti.files().file_path(idx) + "\n"
                     console.write(out)
                 except Exception:
                     pass
@@ -373,8 +381,8 @@ def main() -> None:
         if len(alerts_log) > 20:
             alerts_log = alerts_log[-20:]
 
-        for a in alerts_log:
-            console.write(a + "\n")
+        for line in alerts_log:
+            console.write(line + "\n")
 
         c = console.sleep_and_input(0.5)
 
--- a/bindings/python/install_data/libtorrent/__init__.pyi
+++ b/bindings/python/install_data/libtorrent/__init__.pyi
@@ -1,0 +1,1976 @@
+import datetime
+from typing import Any
+from typing import Callable
+from typing import Dict
+from typing import Iterable
+from typing import Iterator
+from typing import List
+from typing import Mapping
+from typing import Optional
+from typing import overload
+from typing import Tuple
+from typing import Union
+
+from typing_extensions import TypedDict
+
+# I created this file by starting with stubgen and modifying its output.
+
+# I recommend that maintainers occasionally re-run stubgen, and compare it to
+# this file. This diff will be too verbose to use for small changes, but it's the
+# only way to ensure coverage.
+
+# To help with the above process, please keep everything in sorted order, to
+# match stubgen's output.
+
+# Typing notes:
+# - This project has a very particular approach to typed enum values. Pay
+#   special attention to which are typed, and which are just ints.
+# - Some attributes like add_torrent_params.save_path ought to be read as one
+#   type (str) and set as another (Union[str, bytes]), but mypy doesn't support
+#   this.
+# - We use strict primitive types (Dict and List) for function args, because
+#   boost strictly requires these, and will not understand duck-typed
+#   abstractions like Mapping or Sequence.
+# - I chose to make the "names" and "values" enum class attributes typed as
+#   Mapping, specifically to block modification. I coludn't think of a case
+#   where these would need to be used in a Dict context.
+# - Boost mangles function signatures to all look like (*args, **kwargs) in
+#   python. This allows for "anonymous" positional args which python doesn't
+#   normally support. We want the type stubs to match the *usage* of a function,
+#   not this mangled version, so for anonymous positional args, we use a name
+#   like "_name" (these names might fairly all be "_", but python doesn't allow
+#   duplicates). PEP 570 provides positional-only argument syntax but mypy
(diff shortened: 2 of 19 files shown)


bindings/python/install_data/libtorrent/__init__.pyi
```diff
@@ -0,0 +1,1954 @@
...
+# I recommend that maintainers occasionally re-run stubgen, and compare it to
+# this file. This diff will be too verbose to use for small changes, but it's the
+# only way to ensure coverage.
+
+# To help with the above process, please keep everything in sorted order, to
+# match stubgen's output.
+
+# Typing notes:
+# - This project has a very particular approach to typed enum values. Pay
+#   special attention to which are typed, and which are just ints.
```
arvidn, line 27: I (think I)  made all enums convert to plain `int` on the python side. Mostly for simplicity, and because type checking isn't very strong in python anyway. I'd be open to better ways of doing it. Same thing with flag types.

bindings/python/install_data/libtorrent/__init__.pyi
```diff
@@ -0,0 +1,1954 @@
...
+# - Some attributes like add_torrent_params.save_path ought to be read as one
+#   type (str) and set as another (Union[str, bytes]), but mypy doesn't support
+#   this.
+# - We use strict primitive types (Dict and List) for function args, because
+#   boost strictly requires these, and will not understand duck-typed
+#   abstractions like Mapping or Sequence.
```
arvidn, line 33: I would think it's *possible* to essentially take `Any` by accepting `boost::python::object` on the C++ side. But it might be a bit more cumbersome to call the `__X__` functions to treat it as a dict or list.

arvidn:
what does `bindings/python/install_data/libtorrent/py.typed ` do?

arvidn:
This seems like a very good change. I especially appreciate the splitting up and additional unit tests.
Would I need to rerun `stubgen` every time I updated the API? say, adding a function or overload?
Where does `stubgen` get its type information from?

I haven't reviewed everything carefully yet, I will do that later.

bindings/python/install_data/libtorrent/__init__.pyi
```diff
@@ -0,0 +1,1954 @@
...
+
+# To help with the above process, please keep everything in sorted order, to
+# match stubgen's output.
+
+# Typing notes:
+# - This project has a very particular approach to typed enum values. Pay
+#   special attention to which are typed, and which are just ints.
+# - Some attributes like add_torrent_params.save_path ought to be read as one
+#   type (str) and set as another (Union[str, bytes]), but mypy doesn't support
+#   this.
```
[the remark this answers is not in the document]
AllSeeingEyeTolledEweSew: Seems like a fine thing to support, but I believe `add_torrent_params.save_path` doesn't actually support being set from a `pathlib.Path` instance today, so it would require a code change.

The type stubs should (mostly) faithfully reflect the implemented functionality.

In this comment, I was referring to this situation:
```python
atp.save_path = "str"  # supported, should type-check
atp.save_path = b"bytes"  # supported, should type-check
x = atp.save_path  # type-checker should know this is a *str*
```

But, mypy just doesn't support this today. Best we can do is make `save_path` a `str` since that's the most common usage.

bindings/python/install_data/libtorrent/__init__.pyi
AllSeeingEyeTolledEweSew: Where the enum values are *declared* in python, I found there's an even split between strongly-typed enums (members of a class descending from `int`, like `operation_t`) and weakly-typed (just `int`s, like `torrent_flags`). I think this is the difference between being declared with boost.python's `enum` versus not.

Where enum values are *accepted*, I think they're almost always weakly-typed (accepts any `int`). I found one exception in `torrent_info.add_tracker()` which requires a strongly-typed `tracker_source` enum value and won't accept an `int`.

IMO, accepting weakly-typed `int` everywhere is a good thing.

I also think we could convert `torrent_flags` and others to boost.python `enum`s. I don't think it affects any usage and it would help users write strongly-typed code of their own

bindings/python/install_data/libtorrent/__init__.pyi
AllSeeingEyeTolledEweSew: That would be cool. There's a downside of writing a lot of bespoke conversion code; which can become a new vector for bugs. I don't think it's a pressing issue right now.

AllSeeingEyeTolledEweSew:
(to clarify: you will need to manually update the type stubs whenever adding a new function or overload)

AllSeeingEyeTolledEweSew:
@arvidn ping

bindings/python/client.py
```diff
@@ -132,15 +134,18 @@ def print_peer_info(console: ConsoleType, peers: List[Any]) -> None:
         elif p.flags & lt.peer_info.connecting:
             id = "connecting to peer"
         else:
-            id = p.client
+            try:
+                id = p.client.decode()
```
arvidn, line 138: `client` is always supposed to be a valid UTF-8 string. have you seen this not be a valid string?

bindings/python/tests/session_test.py
```diff
@@ -957,9 +957,14 @@ def test_deprecated(self) -> None:
...
         with self.assertWarns(DeprecationWarning):
-            self.session.dht_get_mutable_item("a" * 32, "salt")
+            self.session.dht_get_mutable_item("a" * 32, "salt")  # type: ignore
         with self.assertWarns(DeprecationWarning):
-            self.session.dht_put_mutable_item("a" * 64, "b" * 32, "data", "salt")
+            self.session.dht_put_mutable_item(
+                "a" * 64,  # type: ignore
+                "b" * 32,  # type: ignore
+                "data",  # type: ignore
+                "salt",  # type: ignore
```
arvidn, line 966: I suppose all of these should technically be `bytes`, right?

setup.cfg
```diff
@@ -12,3 +12,4 @@ warn_unused_configs = True
 #disallow_any_unimported = True
 disallow_untyped_defs = True
 disallow_incomplete_defs = True
+mypy_path = bindings/python/install_data
```
arvidn, line 15: will this be included in the package?

arvidn (approved):

arvidn:
Thanks! I think this is good to land. You're good with that, yeah?

bindings/python/client.py
AllSeeingEyeTolledEweSew: The conversion is just here because adding type stubs caused type checking to fail, so it needs to be converted.

I've used transmission a lot in the past, and using that, I've seen a lot of random data in the user agent field. I wasn't sure if libtorrent sanitizes the string, so I did a defensive conversion.

bindings/python/tests/session_test.py
AllSeeingEyeTolledEweSew: Right, the type stubs declare these ase `bytes`, but I wanted to add a test with `str` as a deprecated usage.

setup.cfg
AllSeeingEyeTolledEweSew: This config is only picked up when running `mypy` directly in the repo, or via `pre-commit`.

The python packaging includes `bindings/python/install_data` such that when libtorrent is installed, `mypy` would find the type stubs automatically.

AllSeeingEyeTolledEweSew:
> Thanks! I think this is good to land. You're good with that, yeah?

Yes! :+1:

AllSeeingEyeTolledEweSew:
Thanks!!

Issue #5333, opened 2020-12-07
Recover "python-libtorrent" package on pypi.org


AllSeeingEyeTolledEweSew:
It would be nice to write python projects that depend on libtorrent. Today that's difficult, partly because libtorrent doesn't exist on pypi.org in any meaningful form. #5325 and #4218 are moving to help with this.

Historically, libtorrent's `setup.py` has historically used the name `python-libtorrent`. This name (not `libtorrent`) would be used to declare a dependency on libtorrent; and python tools (`pip`, etc) would search pypi for this name.

Someone published unofficial builds of libtorrent on pypi under this name, at https://pypi.org/project/python-libtorrent .

Also, linux distros have packaged libtorrent with python packaging metadata that declares the `python-libtorrent` name. See [alpine](https://pkgs.alpinelinux.org/contents?branch=edge&name=py3-libtorrent-rasterbar&arch=x86_64&repo=testing) and [ubuntu](https://packages.ubuntu.com/groovy/amd64/python3-libtorrent/filelist).

This means that there may be python projects that depend on `python-libtorrent` and do work in some environments. So changing this name is not ideal.

(for what it's worth, I don't believe deluge has ever depended on `python-libtorrent`. It just includes a vendored copy of libtorrent and a custom build system)

It would be nice if arvid could take over the project from pypi, so we can publish wheels there. There's a process for doing this: https://www.python.org/dev/peps/pep-0541/#how-to-request-a-name-transfer

filips123:
In which cases is `python-libtorrent` as a package metadata actually used? I think this is only used for installing from PyPI (which doesn't really matter because there are no official releases on PyPI) and checking which packages are installed (using `pip list`/`pip freeze` or maybe some runtime checks but I doubt anyone is using this because they are using custom build system in first place).

Also, it seems that Ubuntu, Debian and Alpine are currently still using 1.x version of libtorrent, so this change might fall into "big 2.0 update" (although it is few months late).

I personally think just `libtorrent` would be better because it makes sense that library published to PyPI is probably a Python library.

AllSeeingEyeTolledEweSew:
> In which cases is `python-libtorrent` as a package metadata actually used?

I don't have evidence anything depends on `python-libtorrent`, but it's also hard to _disprove_. pypi.org doesn't seem to have a way to browse "packages that depend on this package", and there could always be a PPA that depends on the debian/ubuntu packages.

[...]

PR #6188, opened 2021-05-07
Add a cibuildwheel workflow

AllSeeingEyeTolledEweSew:
This adds a `cibuildwheel` workflow.

## Features
 * Builds and tests wheels in 28 flavors
 * Can be run manually, with an option to upload to pypi
 * ~~Will run automatically when a release is published, **unconditionally** uploads to pypi~~
 * If a PR touches cibuildwheel infrastructure (`cibuildwheel.yml` or `tools/cibuildwheel/**`), we will test by building a *representative sample* of wheels (the full suite of wheels takes hours)
 * Caches build artifacts, so re-runs are fast

## Changes
* ~~The python project name is changed from `python-libtorrent` to just `libtorrent`. Closer to common practice in python, plus the name was available on pypi~~
* ~~Other changes to `setup.py`, following the principle that `--config-mode=distutils` should create a wheel-suitable build by default~~

## Tests
I patched this (and other dependent build system changes) on the `v1.2.14` and `v2.0.4` tags. I uploaded the wheels to test-pypi:
* https://test.pypi.org/project/libtorrent/2.0.4.post3/
* https://test.pypi.org/project/libtorrent/1.2.14.post3/

## Release procedures and notes

* @arvidn should create a pypi API token restricted to the "libtorrent" project, and add it as a github actions secret (`PYPI_API_TOKEN`)
* EDIT: also create an API token on test-pypi restricted to the separate "libtorrent" project there, and add it as a github actions secret `TEST_PYPI_API_TOKEN`
* I already registered the project on pypi (EDIT: and test-pypi), and invited @arvidn as an owner

Proposed release procedure:
* before you create a release, manually run the `cibuildwheel` workflow without publishing to pypi, to make sure everything builds
* If everything builds, run the workflow again and publish to pypi. ~~Or just create the release, and the workflow will publish automatically.~~

## Release procedure caveats
**NOTE**: pypi uploads are **immutable**, and **filenames cannot be reused even after files are deleted**. Publish with care!!

* I already uploaded some files for libtorrent 2.0.4, then deleted them, not realizing those filenames can never be used again
  * If desired, this could be remedied by publishing a 2.0.4.1 release
* It's good practice to upload to https://test.pypi.org/ before uploading the "production" index at https://pypi.org/
* ~~But "libtorrent" at test-pypi is held by someone else, I believe by github user @mayli~~ 
* ~~I have contacted @mayli by email to transfer ownership of the test-pypi project~~ EDIT: mayli transferred ownership of the project, thanks!

## Build Flavors
* `cp37-cp37m-macosx_10_9_x86_64`
* `cp37-cp37m-manylinux_2_12_x86_64.manylinux2010_x86_64`
* `cp37-cp37m-manylinux_2_17_aarch64.manylinux2014_aarch64`
* `cp37-cp37m-musllinux_1_1_aarch64`
* `cp37-cp37m-musllinux_1_1_x86_64`
* `cp37-cp37m-win32`
* `cp37-cp37m-win_amd64`
* `cp38-cp38-macosx_10_9_x86_64`
* `cp38-cp38-manylinux_2_12_x86_64.manylinux2010_x86_64`
* `cp38-cp38-manylinux_2_17_aarch64.manylinux2014_aarch64`
* `cp38-cp38-musllinux_1_1_aarch64`
* `cp38-cp38-musllinux_1_1_x86_64`
* `cp38-cp38-win32`
* `cp38-cp38-win_amd64`
* `cp39-cp39-macosx_10_9_x86_64`
* `cp39-cp39-manylinux_2_12_x86_64.manylinux2010_x86_64`
* `cp39-cp39-manylinux_2_17_aarch64.manylinux2014_aarch64`
* `cp39-cp39-musllinux_1_1_aarch64`
* `cp39-cp39-musllinux_1_1_x86_64`
* `cp39-cp39-win32`
* `cp39-cp39-win_amd64`
* `cp310-cp310-macosx_10_9_x86_64`
* `cp310-cp310-manylinux_2_12_x86_64.manylinux2010_x86_64`
* `cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64`
* `cp310-cp310-musllinux_1_1_aarch64`
* `cp310-cp310-musllinux_1_1_x86_64`
* `cp310-cp310-win32`
* `cp310-cp310-win_amd64`

### Flavors we'd like to build but can't
* macos arm64
  * the x86 macos runners *can* cross-compile, but I can't seem to make it work. We are invoking `b2 architecture=arm address-model=64` which should be right, but there are still link errors about mixing `x86_64` objects.
  * macos on x86_64 cannot run or emulate arm64 code, so these builds can't be tested until there are m1 mac runners 
* macos `universal2`
  * `b2 architecture=combined address-model=64` does not build this (should pick `-arch x86_64 -arch arm64`), even in boost 1.77.0. I tried `cxxflags="-arch x86_64 -arch arm64"`, but the compiler says these flags are unused.
  * as above, these can't be tested until there are m1 mac runners. apparently mac on m1 *can* emulate x86_64, so universal binaries *can* be tested on a single runner, just not the runners we have
* ~~`musllinux`~~
  * ~~this is new, and `cibuildwheel` doesn't support it yet~~ EDIT: the latest cibuildwheel supports this!
* pypy
  * the python bindings don't build with this. I'll investigate

### Other flavors
The build times already take hours, and the artifacts are large due to static linking, and pypi has finite hosting space. We should add flavors as requested, rather than building everything.

[Python 3.6 will reach end of life in 2021-12](https://www.python.org/dev/peps/pep-0494/), so I've preemptively removed it from the list to shorten build times. [numpy has already done this in their most recent release](https://pypi.org/project/numpy/#files)

## Future optimizations

The builds currently take hours on `aarch64` and other emulated architectures. It would be nice to reduce this.

### `abi3`
The biggest win would be to use `abi3` which would let us produce one build per platform rather than building for every python version. However
  * [boost-python doesn't support `abi3`](https://github.com/boostorg/python/issues/221)
  * [pybind11 (the main boost-python alternative) doesn't support `abi3`](https://github.com/pybind/pybind11/issues/1755)
  * [cibuildwheel is janky to use with `abi3`](https://github.com/pypa/cibuildwheel/issues/78)

### Prebuild dependencies

* openssl and boost-python (in all python versions) are built from source every time. We could cache them somewhere. On linux, using `actions/cache` is difficult  because the builds run in docker. The right approach is probably to build a custom docker image with prebuilt dependencies. This would especially cut down build times for `aarch64` and other emulated architectures.

### macos `ccache`

We build artifacts for a given platform sequentially in a single job, which lets us use `ccache`. On the Linux builds this works well today.

On mac, `ccache` reports 0% hit rates between builds for some reason. The only thing that changes is the python version. We should investigate this.

### windows `ccache` equivalent

On Windows, there is no build caching. It would be nice to have. However the Windows builds already complete in ~40 minutes, so this is a smaller gain compared to others.

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -1,0 +1,142 @@
+name: cibuildwheel
+
+# Note: We use a dynamic matrix to build different sets of wheels under
+# different conditions. On workflow_dispatch, we build the full suite of
+# wheels. This takes hours, so on pull_request, we just build a representative
+# sample.
+
+# The full list of cibuildwheel's build targets can be found here:
+# https://github.com/pypa/cibuildwheel/blob/v2.2.0a1/cibuildwheel/resources/build-platforms.toml
+
+# Notes on build targets we (don't) support:
+# - pypy: libtorrent doesn't build with pypy as of writing
+# - macos_arm64: can be cross-compiled from x86_64, but not run, so can't be
+#   tested. Build output indicates it isn't building correctly
+# - macos_universal2: b2 / setup.py doesn't have a straightforward way to build
+#   this as of writing
+# - abi3: Not supported by boost-python (or pybind11) or cibuildwheel as of
+#   writing
+
+on:
+  workflow_dispatch:
+    inputs:
+      publish:
+        description: Write 'PUBLISH' to publish to pypi. BEWARE! ARTIFACTS ARE IMMUTABLE AND CANNOT BE REPLACED ONCE PUBLISHED!
+      publish_test:
+        description: Write 'PUBLISH_TEST' to publish to test-pypi. BEWARE! ARTIFACTS ARE IMMUTABLE AND CANNOT BE REPLACED ONCE PUBLISHED!
+
+  pull_request:
+    paths:
+      - .github/workflows/cibuildwheel.yml
+      - tools/cibuildwheel/**
+      - pyproject.toml
+
+jobs:
+  configure_matrix:
+    runs-on: ubuntu-latest
+    env:
+      # github actions syntax doesn't allow us to have yaml structures as
+      # an input to a job. These environment variables are literal json strings
+      MATRIX_PULL_REQUEST: |
+        {
+          "include": [
+            {"os": "ubuntu-20.04", "CIBW_BUILD": "cp37-manylinux_*", "CIBW_ARCHS": "x86_64"},
+            {"os": "ubuntu-20.04", "CIBW_BUILD": "cp37-musllinux_*", "CIBW_ARCHS": "x86_64"},
+            {"os": "macos-10.15", "CIBW_BUILD": "cp37-*", "CIBW_ARCHS": "x86_64"},
+            {"os": "windows-2019", "CIBW_BUILD": "cp37-*", "CIBW_ARCHS": "AMD64"}
+          ]
+        }
+      MATRIX_WORKFLOW_DISPATCH: |
+        {
+          "include": [
+            {"os": "ubuntu-20.04", "CIBW_BUILD": "cp*-manylinux_*", "CIBW_ARCHS": "x86_64"},
+            {"os": "ubuntu-20.04", "CIBW_BUILD": "cp*-manylinux_*", "CIBW_ARCHS": "aarch64"},
+            {"os": "ubuntu-20.04", "CIBW_BUILD": "cp*-musllinux_*", "CIBW_ARCHS": "x86_64"},
+            {"os": "ubuntu-20.04", "CIBW_BUILD": "cp*-musllinux_*", "CIBW_ARCHS": "aarch64"},
+            {"os": "macos-10.15", "CIBW_BUILD": "cp*", "CIBW_ARCHS": "x86_64"},
+            {"os": "windows-2019", "CIBW_BUILD": "cp*", "CIBW_ARCHS": "x86"},
+            {"os": "windows-2019", "CIBW_BUILD": "cp*", "CIBW_ARCHS": "AMD64"}
+          ]
+        }
+
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+
+    steps:
+    - id: set-matrix
+      run: |
+        if [ $GITHUB_EVENT_NAME == "pull_request" ]; then
+          echo ::set-output name=matrix::$(echo $MATRIX_PULL_REQUEST | jq -c)
+        else
+          echo ::set-output name=matrix::$(echo $MATRIX_WORKFLOW_DISPATCH | jq -c)
+        fi
+
+  build_wheels:
+    needs: configure_matrix
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJSON(needs.configure_matrix.outputs.matrix) }}
+
+    env:
+      CIBW_BUILD_VERBOSITY: 1
+      CIBW_BUILD: ${{ matrix.CIBW_BUILD }}
+      CIBW_ARCHS: ${{ matrix.CIBW_ARCHS }}
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+
+    - uses: actions/cache@v2
+      id: cache-wheel
+      with:
+        path: wheelhouse
+        key: wheel-${{ matrix.CIBW_BUILD }}-${{ matrix.CIBW_ARCHS }}-${{ github.sha }}
+
+    - uses: docker/setup-qemu-action@v1
+      if: steps.cache-wheel.outputs.cache-hit != 'true' && runner.os == 'Linux'
+
+    - uses: pypa/cibuildwheel@v2.2.2
+      if: steps.cache-wheel.outputs.cache-hit != 'true'
+
+    - uses: actions/upload-artifact@v2
+      with:
+        path: wheelhouse/*.whl
+        name: wheels
+
+  upload_pypi:
+    needs: build_wheels
+    runs-on: ubuntu-latest
+    if: needs.build_wheels.result == 'success' && github.event.inputs.publish == 'PUBLISH'
+
+    steps:
+    - uses: actions/download-artifact@v2
+      with:
+        name: wheels
+        path: wheelhouse
+
+    - uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        password: ${{ secrets.PYPI_API_TOKEN }}
+        packages_dir: wheelhouse
+        skip_existing: true
+
+  upload_pypi_test:
+    needs: build_wheels
+    runs-on: ubuntu-latest
+    if: needs.build_wheels.result == 'success' && github.event.inputs.publish_test == 'PUBLISH_TEST'
+
+    steps:
+    - uses: actions/download-artifact@v2
+      with:
+        name: wheels
+        path: wheelhouse
+
+    - uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        packages_dir: wheelhouse
+        skip_existing: true
+        repository_url: https://test.pypi.org/legacy/
--- a/tools/cibuildwheel/manylinux/build_utils.sh
+++ b/tools/cibuildwheel/manylinux/build_utils.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+# Helper utilities for build
+
+
+# use all flags used by ubuntu 20.04 for hardening builds, dpkg-buildflags --export
+# other flags mentioned in https://wiki.ubuntu.com/ToolChain/CompilerFlags can't be
+# used because the distros used here are too old
+MANYLINUX_CPPFLAGS="-Wdate-time -D_FORTIFY_SOURCE=2"
+MANYLINUX_CFLAGS="-g -O2 -Wall -fdebug-prefix-map=/=. -fstack-protector-strong -Wformat -Werror=format-security"
+MANYLINUX_CXXFLAGS="-g -O2 -Wall -fdebug-prefix-map=/=. -fstack-protector-strong -Wformat -Werror=format-security"
+MANYLINUX_LDFLAGS="-Wl,-Bsymbolic-functions -Wl,-z,relro -Wl,-z,now"
+
+
+function check_var {
+    if [ -z "$1" ]; then
+        echo "required variable not defined"
+        exit 1
+    fi
+}
+
+
+function fetch_source {
+    # This is called both inside and outside the build context (e.g. in Travis) to prefetch
+    # source tarballs, where curl exists (and works)
+    local file=$1
+    check_var ${file}
+    local url=$2
+    check_var ${url}
+    if [ -f ${file} ]; then
+        echo "${file} exists, skipping fetch"
+    else
+        curl -fsSL -o ${file} ${url}/${file}
+    fi
+}
+
+
+function check_sha256sum {
+    local fname=$1
+    check_var ${fname}
+    local sha256=$2
+    check_var ${sha256}
+
+    echo "${sha256}  ${fname}" > ${fname}.sha256
+    sha256sum -c ${fname}.sha256
+    rm -f ${fname}.sha256
+}
+
+
+function do_standard_install {
+    ./configure "$@" CPPFLAGS="${MANYLINUX_CPPFLAGS}" CFLAGS="${MANYLINUX_CFLAGS}" "CXXFLAGS=${MANYLINUX_CXXFLAGS}" LDFLAGS="${MANYLINUX_LDFLAGS}" > /dev/null
+    make > /dev/null
+    make install > /dev/null
+}
+
+function strip_ {
+	# Strip what we can -- and ignore errors, because this just attempts to strip
+	# *everything*, including non-ELF files:
+	find $1 -type f -print0 | xargs -0 -n1 strip --strip-unneeded 2>/dev/null || true
+}
+
+function clean_pyc {
+	find $1 -type f -a \( -name '*.pyc' -o -name '*.pyo' \) -delete
+}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,6 +26,11 @@ repos:
     -   id: debug-statements
     -   id: check-symlinks
     -   id: check-toml
+- repo: https://github.com/pappasam/toml-sort
+  rev: v0.19.0
+  hooks:
+  - id: toml-sort
+    args: [--all, --in-place]
 - repo: https://github.com/pre-commit/pygrep-hooks
   rev: v1.7.0
   hooks:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,88 @@
+[project]
+requires-python = ">=3.7"
+
+[tool.cibuildwheel]
+skip = "pp*"
+
+[tool.cibuildwheel.macos]
+before-all = [
+    "./tools/cibuildwheel/setup_boost.sh $BOOST_VERSION $BOOST_ROOT",
+    "brew install openssl",
+]
+test-command = [
+    "cd {project}/bindings/python",
+    "python test.py",
+]
+
+[tool.cibuildwheel.macos.environment]
+BOOST_BUILD_PATH = "/tmp/boost/tools/build"
+BOOST_ROOT = "/tmp/boost"
+BOOST_VERSION = "1.76.0"
+MACOSX_DEPLOYMENT_TARGET = "10.9"  # required for full C++11 support
+PATH = "/tmp/boost:$PATH"
+
+[[tool.cibuildwheel.overrides]]
+before-all = [
+    "./tools/cibuildwheel/setup_boost.sh $BOOST_VERSION $BOOST_ROOT",
+    "yum install -y glibc-static",  # needed for libutil.a and libdl.a
+    "./tools/cibuildwheel/setup_ccache_on_manylinux.sh",
+    "./tools/cibuildwheel/setup_openssl.sh",
+]
+before-test = "ccache -s"
+select = "*-manylinux_*"
+test-command = [
+    "cd {project}/bindings/python",
+    "python test.py",
+]
+
+[tool.cibuildwheel.overrides.environment]  # sub-table of previous block!
+BOOST_BUILD_PATH = "/tmp/boost/tools/build"
+BOOST_ROOT = "/tmp/boost"
+BOOST_VERSION = "1.76.0"
+PATH = "/usr/local/ccache/bin:/tmp/boost:$PATH"
+
+[[tool.cibuildwheel.overrides]]
+before-all = [
+    "./tools/cibuildwheel/setup_boost.sh $BOOST_VERSION $BOOST_ROOT",
+    "apk add ccache openssl-dev openssl-libs-static",
+    "./tools/cibuildwheel/setup_openssl.sh",
+]
+before-test = "ccache -s"
+select = "*-musllinux_*"
+test-command = [
+    "cd {project}/bindings/python",
+    "python test.py",
+]
+
+[tool.cibuildwheel.overrides.environment]  # sub-table of previous block!
+BOOST_BUILD_PATH = "/tmp/boost/tools/build"
+BOOST_ROOT = "/tmp/boost"
+BOOST_VERSION = "1.76.0"
+PATH = "/usr/lib/ccache/bin:/tmp/boost:$PATH"
+
+[[tool.cibuildwheel.overrides]]
+before-all = [
+    "bash -c './tools/cibuildwheel/setup_boost.sh $BOOST_VERSION $BOOST_ROOT'",
+    "bash -c 'choco install --no-progress --x86 openssl'",  # choco only allows EITHER 32 OR 64-bit version of a package
+]
+select = "*-win32"
+
+[[tool.cibuildwheel.overrides]]
+before-all = [
+    "bash -c './tools/cibuildwheel/setup_boost.sh $BOOST_VERSION $BOOST_ROOT'",
+    "bash -c 'choco install --no-progress openssl'",        # choco only allows EITHER 32 OR 64-bit version of a package
+]
+select = "*-win_amd64"
+
+[tool.cibuildwheel.windows]
+test-command = '''bash -c "cd '{project}/bindings/python' && python test.py"'''
+
+[tool.cibuildwheel.windows.environment]
+BOOST_BUILD_PATH = 'c:/boost/tools/build'
+BOOST_ROOT = 'c:/boost'
+BOOST_VERSION = "1.76.0"
+PATH = 'c:/boost;$PATH'
+
 [tool.isort]
 profile = "google"
 single_line_exclusions = []
--- a/tools/cibuildwheel/manylinux/build-openssl.sh
+++ b/tools/cibuildwheel/manylinux/build-openssl.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+# Top-level build script called from Dockerfile
+
+# Stop at any error, show all commands
+set -exuo pipefail
+
+# Get script directory
+MY_DIR=$(dirname "${BASH_SOURCE[0]}")
+
+# Get build utilities
+source $MY_DIR/build_utils.sh
+
+# Install a more recent openssl
+check_var ${OPENSSL_ROOT}
+check_var ${OPENSSL_HASH}
+check_var ${OPENSSL_DOWNLOAD_URL}
+
+OPENSSL_VERSION=${OPENSSL_ROOT#*-}
+OPENSSL_MIN_VERSION=1.1.1
+
+INSTALLED=$(openssl version | head -1 | awk '{ print $2 }')
+SMALLEST=$(echo -e "${INSTALLED}\n${OPENSSL_MIN_VERSION}" | sort -t. -k 1,1n -k 2,2n -k 3,3n -k 4,4n | head -1)
+if [ "${SMALLEST}" == "${OPENSSL_MIN_VERSION}" ]; then
+	echo "skipping installation of openssl ${OPENSSL_VERSION}, system provides openssl ${INSTALLED} which is newer than openssl ${OPENSSL_MIN_VERSION}"
+	exit 0
+fi
+
+if which yum; then
+	yum erase -y openssl-devel
+else
+	apt-get remove -y libssl-dev
+fi
+
+fetch_source ${OPENSSL_ROOT}.tar.gz ${OPENSSL_DOWNLOAD_URL}
+check_sha256sum ${OPENSSL_ROOT}.tar.gz ${OPENSSL_HASH}
+tar -xzf ${OPENSSL_ROOT}.tar.gz
+pushd ${OPENSSL_ROOT}
+./config no-shared --prefix=/usr/local/ssl --openssldir=/usr/local/ssl CPPFLAGS="${MANYLINUX_CPPFLAGS}" CFLAGS="${MANYLINUX_CFLAGS} -fPIC" CXXFLAGS="${MANYLINUX_CXXFLAGS} -fPIC" LDFLAGS="${MANYLINUX_LDFLAGS} -fPIC" > /dev/null
+make > /dev/null
+make install_sw > /dev/null
+popd
+rm -rf ${OPENSSL_ROOT} ${OPENSSL_ROOT}.tar.gz
+
+
+/usr/local/ssl/bin/openssl version
--- a/tools/cibuildwheel/manylinux/openssl-version.sh
+++ b/tools/cibuildwheel/manylinux/openssl-version.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+export OPENSSL_ROOT=openssl-1.1.1l
+export OPENSSL_HASH=0b7a3e5e59c34827fe0c3a74b7ec8baef302b98fa80088d7f9153aa16fa76bd1
+export OPENSSL_DOWNLOAD_URL=https://www.openssl.org/source
--- a/tools/cibuildwheel/setup_boost.sh
+++ b/tools/cibuildwheel/setup_boost.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# This script is meant to be called by cibuildwheel. It should run on github
+# actions Linux, Mac and Windows.
+
+set -ex
+
+VERSION="$1"
+BOOST_ROOT="$2"
+
+VUNDER="${VERSION//./_}"
+
+TEMP_ARCHIVE="$(mktemp)"
+
+mkdir -p "$BOOST_ROOT"
+
+curl -L -o "$TEMP_ARCHIVE" "https://boostorg.jfrog.io/artifactory/main/release/${VERSION}/source/boost_${VUNDER}.tar.gz"
+
+tar -z -x -C "$BOOST_ROOT" -f "$TEMP_ARCHIVE" --strip-components 1
+rm "$TEMP_ARCHIVE"
+
+cd "$BOOST_ROOT"
+
+./bootstrap.sh
+
+cat ./project-config.jam
+
+./b2 headers
--- a/tools/cibuildwheel/setup_ccache_on_manylinux.sh
+++ b/tools/cibuildwheel/setup_ccache_on_manylinux.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+set -xe
+
+if [ $(uname -m) != x86_64 -a $(uname -m) != aarch64 ]
+then
+	echo "ccache isn't known to exist on $(uname -m). skipping ccache setup"
+	exit 0
+fi
+
+yum install -y epel-release  # overlay containing ccache
+yum install -y ccache
+
+# The symlinks in /usr/lib64/ccache are auto-managed by rpm postinstall
+# scripts. They are only created if appropriate packages are installed. However
+# this management only knows about the standard gcc* packages, not the
+# devtoolset packages, so they don't get created correctly on manylinux. We try
+# to create them ourselves.
+mkdir -p /usr/local/ccache/bin
+for path in /opt/rh/devtoolset-*/root/usr/bin/*cc /opt/rh/devtoolset-*/root/usr/bin/*cc-[0-9]* /opt/rh/devtoolset-*/root/usr/bin/*++ /opt/rh/devtoolset-*/root/usr/bin/*++-[0-9]* /opt/rh/devtoolset-*/root/usr/bin/*cpp /opt/rh/devtoolset-*/root/usr/bin/*cpp-[0-9]*
+do
+	ln -s /usr/bin/ccache "/usr/local/ccache/bin/$(basename "$path")"
+done
--- a/tools/cibuildwheel/setup_openssl.sh
+++ b/tools/cibuildwheel/setup_openssl.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+set -xe
+
+TOOLS=$(dirname "$(readlink -f "$0")")
+
+. "$TOOLS/manylinux/openssl-version.sh"
+manylinux-entrypoint "$TOOLS/manylinux/build-openssl.sh"
+
+# If the build script finds a new enough openssl on the system, it will skip building.
+
+if [ -d /usr/local/ssl ]
+then
+	ln -s /usr/local/ssl/include/openssl /usr/local/include
+	ln -s /usr/local/ssl/lib/libcrypto.a /usr/local/lib
+	ln -s /usr/local/ssl/lib/libssl.a /usr/local/lib
+fi


AllSeeingEyeTolledEweSew:
@arvidn could you take a look at the linux and mac workflows? They have compile errors but I don't understand why.

arvidn:
```
    ../../src/disk_io_job.cpp:111:8: error: 'visit<libtorrent::aux::(anonymous namespace)::caller_visitor, std::__1::variant<...> &>' is unavailable: introduced in macOS 10.14
```

Apparently `std::variant` isn't part of the runtime on MacOS until 10.14. So there needs to be a requirement for at least that version. It's something I need to fix in the Jamfile (and cmake file) to specify the oldest version to support.

I think the `Jamfile` and `CMakeList.txt` both need to specify `MACOSX_DEPLOYMENT_TARGET=10.14` as a macro.

arvidn:
I'm not sure why this shows up in this PR though.

AllSeeingEyeTolledEweSew:
> I'm not sure why this shows up in this PR though.

I think I needed to configure cibuildwheel for this, but didn't. https://cibuildwheel.readthedocs.io/en/stable/cpp_standards/#macos-and-deployment-target-versions

Is something similar happening with the Linux error? I'll try to configure a newer `manylinux` image and see if it gets fixed.

AllSeeingEyeTolledEweSew:
@arvidn could you take a look at the latest linux build? I can't explain why I'm getting `cannot find -ldl`. The centos docker image is missing `glibc-static`, but if I install it, I still get the same error.

`-Wl,-Bstatic -ldl -lutil -Wl,-Bdynamic -ldl -lutil -lpthread` seems odd to me. I'm not sure why it wants static `-ldl -lutil`, or why it seems to request them as both static and dynamic.

AllSeeingEyeTolledEweSew:
@arvidn ping

AllSeeingEyeTolledEweSew:
FYI: with this PR, plus #5333, we'll be able to post stuff on pypi. This will be really huge for python usability!

AllSeeingEyeTolledEweSew:
For uploading to pypi:

cibuildwheel has [some general info](https://cibuildwheel.readthedocs.io/en/stable/deliver-to-pypi/).

I'm not super familiar with github releases. I expect the best process would be something like:
 * when a release (or version tag?) is created, run the `cibuildwheel` jobs
 * when a release is published, upload the artifacts to pypi

Offhand I think I understand how to write github actions to do this the naive way, such that the "publish" step would re-build the artifacts and include the upload step.

It looks like there's [wheelhouse-uploader](https://github.com/ogrisel/wheelhouse-uploader) which helps this out by stashing build artifacts on cloud storage before they're ready to be uploaded when the release is published.

We could also just build artifacts when you create a release, use `actions/upload-artifact` to upload them to github, then you can manually push them to pypi when you create a release. We could figure out automation later.

AllSeeingEyeTolledEweSew:
If you wanted to stick something on pypi now to claim the name, you could:
 - rebase this on `v2.0.4`
 - create a PR to let the jobs run
 - download the artifacts and manually publish

AllSeeingEyeTolledEweSew:
**TL;DR**: I think c++17 is too new for the current state of python linux wheel packaging.

Details:

 * Python defines [platform tags](https://www.python.org/dev/peps/pep-0425/) to specify binary compatibility. There's a non-specific `linux`, and a newer `manylinux` ([1](https://www.python.org/dev/peps/pep-0513/), [2](https://www.python.org/dev/peps/pep-0600/)) which restricts the allowed set of dependent libraries, and *symbol versions* for `libc.so`, `libgcc_s.so` and `libstdc++.so`.
 * Cross-referencing current [defined manylinux tags](https://github.com/mayeut/pep600_compliance/blob/f86a7d7c153cc45aa3f2add6ffcf610c80501657/pep600_compliance/tools/policy.json) with [gcc symbol versions](https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html) and [libstdc++'s language support](https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html), we have
   * `manylinux_2_5`: gcc 4.1.1
   * `manylinux_2_12`: gcc 4.4.2
   * `manylinux_2_17`: gcc 4.8.3
   * `manylinux_2_24`: gcc 6.1.0 (stable c++14 ABI)
   * `manylinux_2_27`: gcc 7.2.0 (stable c++14 ABI)
   * (note: gcc 9.1 is the first version with stable c++17 ABI)

I've rebased this on `RC_2_0`, where we still have c++14. It builds now! But if we merge c++17 support, it will break.

It *is possible* to use c++17, it would just mean our wheels would have the unconstrained `linux` tag, which would constrain which distros/versions can install them, and users on the wrong distro would get runtime `ImportError`. To implement this, we should be able to just change `CIBW_MANYLINUX_*_IMAGE` to something newer of our choice (probably debian for best compatibility?) and ensure `auditwheel` uses `--plat linux`.

AllSeeingEyeTolledEweSew:
@arvidn could you take another look before I un-draft? My current plan is to change these workflows to `on: release: types: [created]`, unless you want something else.

AllSeeingEyeTolledEweSew:
Per https://github.com/pypa/manylinux/issues/118, looks like some people just use `-static-libstdc++`. Is there a downside if we do this, except code size? Then we can package much more broadly-compatible linux wheels.

AllSeeingEyeTolledEweSew:
After speaking with the `cibuildwheel` maintainers, my understanding is that C++14 should be usable on their `manylinux2010` and C++17 on their `manylinux2014`. Those images use selective static linking to make modern C++ link against ancient `glibc`.

Currently doing either of those, I run into the bug that `-ldl` and `-lutil` are not found, but that seems to be an independent bug, *not* related to version requirements.

The newer `manylinux_2_24` image has a plain install of gcc 6.x which does no static linking, but is suitable for C++14 builds, and doesn't have the `-ldl`/`-lutil` bug. Rather than chase the bug, it's easiest to just put this in `RC_2_0` for now.

I also changed the python distribution name from `python-libtorrent` to `libtorrent`, per some consensus from #5333.

Still waiting on @arvidn to take a look at the workflows, before I change them to be `on: release`. At that point, the only way to test the changes will be to create a release, so it's better to test what I have now.

jkterry1:
If I'm reading this PR correctly, you're only distributing wheels for x86_64 systems. Could you build wheels for ARM as well? ARM macbooks, windows machines and raspberry pis are all pretty common.

AllSeeingEyeTolledEweSew:
I'm trying to get it working first.

Before that, I'm trying to simplify it a bit. It's a fairly complex feature that arvid hasn't even had time to look at in months.

AllSeeingEyeTolledEweSew:
I fixed the `-lutil`/`-ldl` problem. `glibc-static` needs to be installed on the centos images, presumably for `libutil.a` and `libdl.a`. I don't fully understand this, since we do end up with dynamic linkage to `libutil` and `libdl`. But it works! The workflow now uses `manylinux2010` for x86 and `manylinux2014` for non-x86, which maximizes compatibility.

AllSeeingEyeTolledEweSew:
@arvidn I think this is ready now, please review when you have a chance. It would be really great to land this before the next release

jkterry1:
@AllSeeingEyeTolledEweSew were you able to sort out wheel ARM support?

mayli:
> @AllSeeingEyeTolledEweSew were you able to sort out wheel ARM support?

Those aarch64 should work for modern ARM system.

AllSeeingEyeTolledEweSew:
> @AllSeeingEyeTolledEweSew were you able to sort out wheel ARM support?

I enabled everything that was easy to enable with `cibuildwheel`.

See the description. `aarch64` works, which seems like it may cover one or more raspberry pi models. I can't really tell which ones.

macos arm64 can be cross-compiled but can't be tested right now. Based on a spot-check, doesn't *look* like it's building correctly.

It's pretty important to only release tested builds, since you can't replace uploads on pypi. And the testing should be automated, because arvid has enough work to do without manual checks.

It looks like there aren't currently any hosted m1 mac CI systems. AFAICT anyone doing m1 mac CI is using self-hosted runners.

Maybe macos arm64 isn't crucial, since it seems m1 macs can run x86 code in emulation.

AllSeeingEyeTolledEweSew:
@arvidn, with your permission, I'd like to publish wheels for 1.2.14 and 2.0.4 to pypi, so people can start using it today.

I applied changes from this PR on top of the release tags `v2.0.4` and `v1.2.14`. For 2.0.4, I also changed the python module version to `2.0.4.post1`, to avoid conflict with my erroneous uploads of `2.0.4`.

So, those branches contain *only* commits from their respective releases, plus build system changes to build wheels.

I ran these workflows on those branches:
 - [2.0.4](https://github.com/AllSeeingEyeTolledEweSew/libtorrent/actions/runs/1188388156)
 - [1.2.14](https://github.com/AllSeeingEyeTolledEweSew/libtorrent/actions/runs/1187897245)

IMO these wheels are appropriate for release today.

mayli:
> @arvidn, with your permission, I'd like to publish wheels for 1.2.14 and 2.0.4 to pypi, so people can start using it today.
> 
> I applied changes from this PR on top of the release tags `v2.0.4` and `v1.2.14`. For 2.0.4, I also changed the python module version to `2.0.4.post1`, to avoid conflict with my erroneous uploads of `2.0.4`.
> 
> So, those branches contain _only_ commits from their respective releases, plus build system changes to build wheels.
> 
> I ran the cibuildwheel workflow on those branches:
> 
> * [2.0.4](https://github.com/AllSeeingEyeTolledEweSew/libtorrent/actions/runs/1188388156)
> * [1.2.14](https://github.com/AllSeeingEyeTolledEweSew/libtorrent/actions/runs/1187897245)
> 
> IMO these wheels are appropriate for release today.

@AllSeeingEyeTolledEweSew  Do you want to release to test.pypi first? I tried to invite arvidn as owner, but i guess he is too busy on python binding. I can either transfer the owner to you, or delete the project so you can create one on your own.

https://test.pypi.org/manage/project/libtorrent/releases/

AllSeeingEyeTolledEweSew:
> @AllSeeingEyeTolledEweSew Do you want to release to test.pypi first? I tried to invite arvidn as owner, but i guess he is too busy on python binding. I can either transfer the owner to you, or delete the project so you can create one on your own.
> 
> https://test.pypi.org/manage/project/libtorrent/releases/

Yes, releasing to test.pypi first would be ideal.

It would also be ideal for arvid to own and manage the projects, but he seems very busy these days.

Even before he was busy, he prioritizes core development over python integration.

I've been doing a lot of python work on libtorrent, and I would be happy to take ownership of libtorrent on test.pypi, if you had no other plans for it.

AllSeeingEyeTolledEweSew:
That is: having both me and arvid as owners on the project

mayli:
> That is: having both me and arvid as owners on the project

Yep, I've released the project name, and you can grab it now.

AllSeeingEyeTolledEweSew:
Wonderful, thanks!

I've uploaded those release-tag wheels to https://test.pypi.org/project/libtorrent/

AllSeeingEyeTolledEweSew:
@cas-- any chance you could help test the release wheels from testpypi?

AllSeeingEyeTolledEweSew:
@filips123 also

cas--:
@AllSeeingEyeTolledEweSew absolutely :+1: 

I've setup a branch to switch Deluge github actions to use the new repo with 1.2.14. but I am still encountering #6437 so test runs are failing

AllSeeingEyeTolledEweSew:
> @AllSeeingEyeTolledEweSew absolutely +1
> 
> I've setup a branch to switch Deluge github actions to use the new repo with 1.2.14. but I am still encountering #6437 so test runs are failing

I'll see if I can help with that. Note though that my intent is to release wheels for the *existing* 1.2.14 release. That bug would only be fixed in some later release.

cas--:
There might be an issue with existing `python-libtorrent` installed packages alongside `libtorrent` installed. e.g. in a virtualenv with system-site-packages enabled: 
```
$ pip list
...
libtorrent                    2.0.4.post1
python-libtorrent             1.2.14
...
```
There are scenarios where the wrong package is picked by the Python import system. I know I suggested we use `libtorrent` but now I can see this could have unintended consequences. Can we try to reclaim `python-libtorrent` anyway as per #5333 ?

filips123:
> I've never heard of system-site-packages. What's it for?

This is venv which also includes/uses all globally installed Python packages ("Give the virtual environment access to the system site-packages dir").

> any chance you could help test the release wheels from testpypi?

I tested it on 64-bit Windows 10 with Python 3.8 and ["official" Python test](https://github.com/arvidn/libtorrent/blob/RC_2_0/bindings/python/test.py) fails because of

```
======================================================================
FAIL: test_add_torrent_info_hash (__main__.test_session)
  File "C:\Users\filips\AppData\Local\Temp\libtorrent\bindings\python\test.py", line 901, in test_add_torrent_info_hash
    self.assertEqual(h.status().info_hashes, lt.info_hash_t(lt.sha256_hash(b'a' * 32)))
AssertionError: <libtorrent.info_hash_t object at 0x0000027529D27FA0> != <libtorrent.info_hash_t object at 0x0000027529C78680>
```

With this versions in venv:

```
Package    Version
---------- -----------
libtorrent 2.0.4.post1
pip        21.1.3
setuptools 56.0.0
```

I also tested it on almost fresh Windows 11 VM (dev channel in VirtualBox, so this might have caused problems) with 32-bit Python 3.9.7 and it fails to load DLL when importing `libtorrent`:

```
Python 3.9.7 (tags/v3.9.7:1016ef3, Aug 30 2021, 20:05:41) [MSC v.1929 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import libtorrent
  File "<stdin>", line 1, in <module>
ImportError: DLL load failed while importing libtorrent: The specified module could not be found.
```

With this versions:

```
Package    Version
---------- -----------
libtorrent 2.0.4.post1
pip        21.2.3
setuptools 57.4.0
```

On 64-bit Python 3.9.7 in the same VM, importing worked, but had the same problems as in Windows 10:

```
======================================================================
FAIL: test_add_torrent_info_hash (__main__.test_session)
  File "C:\Users\User\Downloads\libtorrent-64\libtorrent-RC_2_0\bindings\python\test.py", line 901, in test_add_torrent_info_hash
    self.assertEqual(h.status().info_hashes, lt.info_hash_t(lt.sha256_hash(b'a' * 32)))
AssertionError: <libtorrent.info_hash_t object at 0x0000022FEA88BFA0> != <libtorrent.info_hash_t object at 0x0000022FE8FC7E50>

Ran 88 tests in 14.394s

FAILED (failures=1)
```

AllSeeingEyeTolledEweSew:
It looks like @arvidn changed the behavior of `info_hashes` in b93a740d92065aa3c24adb058d05f562604a3c8c and changed `test.py` accordingly. This appears in `RC_2_0` *after* `v2.0.4`. 

@arvidn could you explain this change? It seems to break functionality.

@filips123 if you try `test.py` from `v2.0.4`, does it work?

Not sure why 32-bit python doesn't load. [Here](https://github.com/AllSeeingEyeTolledEweSew/libtorrent/runs/3478043967?check_suite_focus=true)'s the workflow where the wheel was built. It seems to successfully run the test suite.

filips123:
> if you try `test.py` from `v2.0.4`, does it work?

Yes, it works with `v2.0.4`.

> Not sure why 32-bit python doesn't load.

I had to install the 32-bit Microsoft Visual C++ Redistributable for Visual Studio 2015, 2017 and 2019, and now it imports correctly. Test script also works.

AllSeeingEyeTolledEweSew:
> I had to install the 32-bit Microsoft Visual C++ Redistributable for Visual Studio 2015, 2017 and 2019, and now it imports correctly. Test script also works.

Ah, did I misunderstand how windows/python work here? My impression was python 3.5+ depends on, and ships with, the 14.x (2015/2017/2019) redistributable.

AllSeeingEyeTolledEweSew:
Cross-referencing [the distributed python.exe / Visual Studio version matrix](https://pythondev.readthedocs.io/windows.html#python-and-visual-studio-version-matrix) with [the C++ runtime feature / Visual Studio version matrix](https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-160#c-standard-library-features), I would expect that the Windows installers for python 3.5+ include a Visual Studio redistributable package that supports extensions using C++14. Apparently that's not the case, according to @filips123's experience. Can anyone with more Windows experience shed some light on this?

I see [tensorflow just declares that their pypi wheel requires the redistributable to be manually installed by users](https://www.tensorflow.org/install/pip), so maybe this issue is unavoidable, but I'd really like to understand this better to convince myself there's nothing we can do.

AllSeeingEyeTolledEweSew:
I just realized this build uses `crypto=built-in`, which doesn't support https trackers. We almost certainly want this in the pypi wheel.

I believe most (all?) official python builds link openssl either statically or dynamically, so building with `crypto=openssl` is probably the right way to go. I'll start experimenting.

AllSeeingEyeTolledEweSew:
Notes so far:
* Python includes [builtin openssl support](https://docs.python.org/3/library/ssl.html), and [cpython's build system doesn't support statically linking openssl](https://bugs.python.org/issue38794), so most python installations *should* include openssl suitable for dynamic linking
* openssl 1.0 and 1.1 are ABI-incompatible. [1.0 is EOL as of Jan 2020](https://www.openssl.org/policies/releasestrat.html). They also have 3.0, but it's still in alpha.
* Very few projects want to statically link openssl, out of fear they'll be on the hook to build new releases in the dead of night when the next heartbleed hits. https://github.com/pyca/cryptography does statically link, but they are a well-supported crypto project dedicated to giving python users access to newest openssl.

TL;DR dynamic linking to openssl 1.1.x will probably work in most environments, but probably break a few.

I would feel better if there were fallbacks to breakage. In a perfect world, openssl would be an optional library opened with `dlopen()`, and would fall back to built-in crypto functions and just fail on https trackers.

Maybe there's a clever solution with python packaging, like publishing a `libtorrent-nossl` package alternative. This would defeat dependency declarations, but maybe there's some other way.

Anyone have thoughts or suggestions?

arvidn:
> I would feel better if there were fallbacks to breakage. In a perfect world, openssl would be an optional library opened with dlopen(), and would fall back to built-in crypto functions and just fail on https trackers.
> 
> Maybe there's a clever solution with python packaging, like publishing a libtorrent-nossl package alternative. This would defeat dependency declarations, but maybe there's some other way.
> 
> Anyone have thoughts or suggestions?

Silently falling back to not supporting SSL is also problematic. I can envision cases where you'd rather fail hard immediately, than limping along without SSL support.

AllSeeingEyeTolledEweSew:
> Silently falling back to not supporting SSL is also problematic. I can envision cases where you'd rather fail hard immediately, than limping along without SSL support.

Right, there should be an explicit signal. In any case this would be a significant change, not part of this PR.

mayli:
For a commonly non-mission critical client application, I am fine to have it working even with a stale statically linked openssl. Bundling all external dependencies is also one of the common practice of building wheels with the help of auditwheel.

AllSeeingEyeTolledEweSew:
> For a commonly non-mission critical client application, I am fine to have it working even with a stale statically linked openssl. Bundling all external dependencies is also one of the common practice of building wheels with the help of auditwheel.

The problem is openssl bugs turn non-mission-critical apps into vulnerabilities. As I understand it, heartbleed could be used to attack peers connected over ssl.

mayli:
Let's go back to the original reason to provide Python wheels for libtorrent: that's to reduce the complexity of user using this lib, without requiring them to have the full toolchain to rebuild the binary extensions.
As a potential user, I'd like to sacrifice the possible security issue in openssl to avoid rebuild or unable to use it. That's one of a common issue of those universal wheels, that ships all dependencies risking one of them being out dated.
For people who wants dynamic linking or want a third-party, they can stay with the libtorrent from their distro or rebuild on their own.

Ideally the pypi package could contain all source code that a user could rebuild libtorrent locally on their own, and that would solve the concerns on static linking.

AllSeeingEyeTolledEweSew:
pypi isn't only used by non-mission-critical users. Reducing complexity isn't its sole purpose.

On static linking, I'm more concerned about the excess burden on arvid than I am about putting the security risk on a subset of users. Even if there are no mission-critical users, arvid will be on the hook to update the wheels after critical releases of openssl for the users that do exist. libtorrent isn't supported by any foundations or corporations (I think), just arvid's spare time, and it's a precious resource, and he is not that connected to the python ecosystem, which is how we come to be here. The less work for arvid, the healthier the python libtorrent ecosystem will be.

All that said, maybe static linking is the right choice in the short term, because it certainly has upsides. I think it depends on what the long-term plan is, and how we transition to the new plan. I don't feel settled on the long term plan just yet.

AllSeeingEyeTolledEweSew:
> Ideally the pypi package could contain all source code that a user could rebuild libtorrent locally on their own, and that would solve the concerns on static linking.

Compiling libtorrent requires boost-build, which isn't part of the the usual toolchain for building from an sdist.

Also, compiling libtorrent from sdist would require openssl plus a bunch of build tools, whereas if they just want to fix libtorrent's dynamic linkage against openssl, they could just install openssl alone.

I'll reiterate my claim that *most* python distributions should include openssl 1.1 already.

jkterry1:
Hey, I'm the maintainer of Gym, one of the largest machine learning libraries in the world (~33 million installs on PyPI- https://github.com/openai/gym, https://pepy.tech/project/gym). We're about to upgrade one of it's biggest pieces of functionality in a few days (https://github.com/openai/gym/pull/2348), and in doing so everyone will have to start using this package of ours https://github.com/PettingZoo-Team/AutoROM.

For a variety of important reasons that should be obvious upon briefly looking at AutoROM (that I don't want to admit to in writing if lawyers see this in the future) we would very like to switch to depending on libtorrent for torrent based distribution of Atari ROMs to tens of thousands of researchers around the world. This means that we would likely be the biggest end user of libtorrent on PyPI by far. With this in mind, let me offer a few comments:

- I am uncomfortable depending on system libraries or requesting users build things themselves. Users will would be installing libtorrent on all manner of devices (e.g. supercomputers, docker containers, shared clusters where users don't have admin rights, etc.). Because of this requiring any dependencies other than PyPI is onerous and is why we aren't using libtorrent now. Relatedly, CI for instance would be our biggest user, and building libtorrent from source is a complete nonstarter there for time reasons.
- At least for us, using libtorrent via PyPI would absolutely be mission critical.
- I at least would have no objection with this not working on old Linux/Python distributions if that's needed for OpenSSL to function in a reasonable manner. We already do similar things ourselves.
- Including binary wheels for as many platforms and architectures as possible is a huge boost for our use case. Especially ARM CPUs.

AllSeeingEyeTolledEweSew:
@jkterry1 Do you have a survey of what CI systems would be used?

jkterry1:
@AllSeeingEyeTolledEweSew what do you mean by survey? It's just the ones everyone uses (Jenkins, GitHub actions, Travis, etc.). A ton of packages would depend on this. On a given service you can configure different Linux setups though.

AllSeeingEyeTolledEweSew:
@jkterry1 sounds good. I was just curious if you knew of usage data to drive decisions about how to proceed. I'm still getting a handle on what python distributions' openssl setups look like in practice.

jkterry1:
@AllSeeingEyeTolledEweSew We don't really have any data (I don't actually know how you'd collect it), but I can at least give you this for our users:

-We don't support pip versions that doesn't have manylinux2010 (and accordingly Linux distros that old)
-The oldest CPython version we support is 3.6. We do not officially support any version of Python other than CPython, and it is fairly unlikely to come up in official use.
-The oldest versions of Python really are used by our users. E.g. 3.6 is likely still heavily used and we won't be removing support for it until it's fully end-of-lifed at the end of this year, at which point Python 3.7 will still be heavily used. Also new versions are very slowly adopted for stupid reasons (e.g. Python 3.9 is super rarely used).
-Weird architectures are meaningfully used, and a ton of users are using ARM based Macbooks
-We don't officially support Windows but we do try to make sure binaries generally work for it

Another future thing is that a bunch of our users use Conda (a popular alternative to pip that many people think is better for a host of reasons). We currently don't have full Conda support but want to add it in the next 6-12 months.

AllSeeingEyeTolledEweSew:
All that closely aligns with what I expect we can build. The linux "weird architecture" builds are wonderfully easy with `cibuildwheel` + github actions today. I haven't found a way to do mac arm64 yet. It's high on my list though.

I've only skimmed conda, but I appreciate that it's popular.

AllSeeingEyeTolledEweSew:
@jkterry1 given libtorrent would be mission-critical for you, how do you feel about including a statically-linked openssl in its build? Do you foresee any use of SSL torrents, described [here](https://blog.libtorrent.org/2012/01/bittorrent-over-ssl/)?

jkterry1:
SSL torrents are in no way required for what we want to do.

Regarding SSL linking, I am at best an amateur in the C, networking and security spaces, and after staring at your discussion and Googling things for 10 minutes I don't have a highly informed opinion. What I can say is that, if I'm understanding the matter correctly though, is that dynamically linking is the good thing, however it may not be possible with all CPython distributions. If you could elaborate on which CPython distributions this is an issue in, I could have a more informed opinion, but in general I am have no major objections to not supporting bizarre CPython distributions that are not commonly used. Other comments:

- I don't know enough to comment on if no SSL would be an issue here or how to handle that. The data we'd be transmitting is in no way sensitive or anything though.
- I have major objections to depending on system level packages, and would greatly prefer using the openssl that comes with Python or something to avoid that.
- Please keep in mind that these comments are all coming someone who has no serious appreciation for the nuances or severity of your concerns regarding future heartbeat type vulnerabilities or the risks of SSL malfunctioning or not being enabled.

arvidn:
Openssl is required￼ for HTTPS trackers, which isn't exactly a fringe feature.

AllSeeingEyeTolledEweSew:
For sanity, let's spin off the discussion of openssl linking strategy to #6448. I still need to discuss the implementation of the strategy in this thread.

I'll draft this PR until after we land on a strategy and I've actually implemented it in this PR.

AllSeeingEyeTolledEweSew:
I rebased this on `RC_1_2` to better integrate with arvid's merge process.

I will un-draft once dependent PRs are merged:
* #6485
* #6493

AllSeeingEyeTolledEweSew:
Per #6448, this now sets up openssl for static vendoring.
* On Windows, we use openssl from chocolatey
* On macos, we use openssl from homebrew
* On Linux, if the system openssl is new enough we use it, otherwise we build openssl from source.
  * To help with this, I stole some openssl setup scripts from the manylinux build repo

AllSeeingEyeTolledEweSew:
Previously I was building one flavor per CI job. It took about 9 hours to build all flavors, with the CI job queue being full for most of that time so nothing else could run. The artifacts were about 450mb in size for a given version.

I decided to
* Not build all flavors: `ppc64le` and `s390x` in particular are CPU hogs due to running in emulation, and no one has asked for these. If someone does ask, we can build then.
* Not build for python 3.6: It's end-of-life in 2 months. `numpy` has already dropped support for the latest release. If someone really needs it, I suggest we only build for the platform they need.
* Build sequentially (for a given platform/arch) rather than in parallel. This lets us conserve the number of times we build dependencies from source, and lets us `ccache` between builds, which drastically reduces the amount of CPU time on emulated architectures.

AllSeeingEyeTolledEweSew:
I think we're ready for review.

I posted updated builds at https://test.pypi.org/project/libtorrent/, please help test if you can!

AllSeeingEyeTolledEweSew:
@arvidn I have a lot of PRs open right now, but if you only have time for one, this is definitely the most important.

AllSeeingEyeTolledEweSew:
@tribler Do you think you guys could help test? https://test.pypi.org/project/libtorrent

AllSeeingEyeTolledEweSew:
@arvidn Any chance this could land ahead of the 1.2.15 release? It would be nice to iron out any issues with this PR when doing the real-world release process.

arvidn:
sure. I tried running:
```
$ python tools/cibuildwheel/run_cibuildwheel.py --output-dir wheelhouse --platform linux
```

and it's *really* slow. Do you know why it takes such a long time? Clearly it's doing a lot more work than just building the python bindings.

arvidn:
it just failed with:
```
    + pip install /tmp/cibuildwheel/repaired_wheel/libtorrent-1.2.15-cp37-cp37m-manylinux_2_12_x86_64.manylinux2010_x86_64.whl
WARNING: You are using pip version 21.2.4; however, version 21.3.1 is available.
You should consider upgrading via the '/tmp/tmp.M2qaFo0dTz/venv/bin/python -m pip install --upgrade pip' command.
Processing /tmp/cibuildwheel/repaired_wheel/libtorrent-1.2.15-cp37-cp37m-manylinux_2_12_x86_64.manylinux2010_x86_64.whl
Installing collected packages: libtorrent
Successfully installed libtorrent-1.2.15
    + sh -c 'bash -c '"'"'cd '"'"'"'"'"'"'"'"'/project/bindings/python'"'"'"'"'"'"'"'"' && python test.py'"'"''
  File "test.py", line 5, in <module>
    import libtorrent as lt
ImportError: libasan.so.5: cannot open shared object file: No such file or directory

```

Why is it building with address sanitizer enabled?

arvidn:
is there a simpler way for me to smoke-test this PR?
I take it the workflow is skipped because it's intended to be run manually, is that right?

AllSeeingEyeTolledEweSew:
> sure. I tried running:
> 
> ```
> $ python tools/cibuildwheel/run_cibuildwheel.py --output-dir wheelhouse --platform linux
> ```
> 
> and it's _really_ slow. Do you know why it takes such a long time? Clearly it's doing a lot more work than just building the python bindings.

That invocation will build for pythons 3.7 through 3.10, for `musllinux` and `manylinux`, and for 32- and 64-bit. So 16 different builds, and they'll build sequentially.

The script is a wrapper which configures a subset of `cibuildwheel`'s options. See https://cibuildwheel.readthedocs.io/en/stable/options/ for a full set.

You could try something like this:

```shell
$ CIBW_BUILD=cp37-manylinux_x86_64 python ./tools/cibuildwheel/run_cibuildwheel.py --platform linux
```

> Why is it building with address sanitizer enabled?

That's unexpected. It should just invoke `setup.py` with no special arguments. Does it build with asan if you use the limited invocation above? If it does, can you paste the output?

> is there a simpler way for me to smoke-test this PR?

`cibuildwheel` uses docker for Linux builds, so you can run them locally. When running on mac or windows though, `cibuildwheel` will stomp on the system, e.g. installing packages globally. It's designed to run on CI for those platforms.

The PR workflows ([1](https://github.com/arvidn/libtorrent/runs/4026939681?check_suite_focus=true), [2](https://github.com/arvidn/libtorrent/runs/4026939814?check_suite_focus=true), [3](https://github.com/arvidn/libtorrent/runs/4026939979?check_suite_focus=true), [4](https://github.com/arvidn/libtorrent/runs/4026940071?check_suite_focus=true)) are the only effective smoke test.

> I take it the workflow is skipped because it's intended to be run manually, is that right?

The workflow does run on PRs (it ran on this PR). It only runs if `cibuildwheel` infrastructure is modified though; for all other PRs the workflow is expected to be redundant with the other workflows that build the python bindings.

When run on a PR, it does a limited selection of builds. When run manually, it does the full suite of builds (this takes hours, due to building on emulated architectures).

Only the `upload_pypi(_test)` steps were skipped. Those steps will only run when the workflow is run manually, and then only if explicitly specified. Sadly with the workflow written this way, it always shows `cibuildwheel / upload_pypi (Pull request) Skipped` in the PR checks list. But this only shows up for PRs which actually trigger the workflow (i.e. those that modify `cibuildwheel` infrastructure).

arvidn:
The reason it takes a long time to run this is not because it's building multiple configurations. I'm not entirely sure I'm reading the log right, but these are the highlights:

```
Copying project into Docker...

    + mkdir -p /project
tar: ./test/rootfs/root: Cannot open: Permission denied
tar: Exiting with failure status due to previous errors

                                                            ✓ 367.46s
```

```
Running before_all...
... (building boost-build, ccache and OpenSSL)
                                                            ✓ 260.88s
```

```
Building wheel...

    + rm -rf /tmp/cibuildwheel/built_wheel
    + mkdir -p /tmp/cibuildwheel/built_wheel
    + python -m pip wheel /project --wheel-dir=/tmp/cibuildwheel/built_wheel --no-deps
    + /opt/python/cp38-cp38/bin/python -c 'import sys, json, glob; json.dump(glob.glob('"'"'/tmp/cibuildwheel/built_wheel/*.whl'"'"'), sys.stdout)'
    + rm -rf /tmp/cibuildwheel/repaired_wheel
    + mkdir -p /tmp/cibuildwheel/repaired_wheel

                                                            ✓ 559.96s
```

```
Testing wheel...

    + pip install virtualenv -c /constraints.txt
    + mktemp -d
    + python -m virtualenv --no-download /tmp/tmp.qTa6PemMSR/venv
created virtual environment CPython3.7.12.final.0-64 in 347ms
  creator CPython3Posix(dest=/tmp/tmp.qTa6PemMSR/venv, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/root/.local/share/virtualenv)
    added seed packages: pip==21.2.4, setuptools==58.1.0, wheel==0.37.0
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator
    + sh -c 'bash -c '"'"'(! which ccache || ccache -s)'"'"''
/usr/bin/ccache
cache directory                     /root/.ccache
cache hit (direct)                     0
cache hit (preprocessed)               0
cache miss                          1142
called for link                      194
called for preprocessing               3
compile failed                         1
preprocessor error                     6
unsupported source language           29
no input file                         24
files in cache                      3243
cache size                          99.5 Mbytes
max cache size                       1.0 Gbytes
    + pip install /tmp/cibuildwheel/repaired_wheel/libtorrent-1.2.15-cp37-cp37m-manylinux_2_12_x86_64.manylinux2010_x86_64.whl
WARNING: You are using pip version 21.2.4; however, version 21.3.1 is available.
You should consider upgrading via the '/tmp/tmp.qTa6PemMSR/venv/bin/python -m pip install --upgrade pip' command.
Processing /tmp/cibuildwheel/repaired_wheel/libtorrent-1.2.15-cp37-cp37m-manylinux_2_12_x86_64.manylinux2010_x86_64.whl
Installing collected packages: libtorrent
Successfully installed libtorrent-1.2.15
    + sh -c 'bash -c '"'"'cd '"'"'"'"'"'"'"'"'/project/bindings/python'"'"'"'"'"'"'"'"' && python test.py'"'"''
  File "test.py", line 5, in <module>
    import libtorrent as lt
ImportError: libasan.so.5: cannot open shared object file: No such file or directory

                                                             ✕ 28.02s
```

AllSeeingEyeTolledEweSew:
That looks mostly right (compare to [the PR check](https://github.com/arvidn/libtorrent/runs/4026939681?check_suite_focus=true)), except this:

> ```
> Copying project into Docker...
> 
>     + mkdir -p /project
> tar: ./test/rootfs/root: Cannot open: Permission denied
> tar: Exiting with failure status due to previous errors
> 
>                                                             ✓ 367.46s
> ```

That's orders of magnitude longer than it should be. What's `./test/rootfs`? Is your checkout clean?

The `auditwheel` step should vendor any nonstandard libraries, or error if it can't, so it shouldn't be possible to even produce a wheel with link errors.

I can only guess the test is picking up a build artifact from your repo.

Can you extract the wheel (it's a zip file) and `ldd` the extension to confirm it's not really linked to `libasan.so.5`? You might need to skip the test step (`CIBW_TEST_SKIP=*`) to get it to copy the artifact out of docker.

Do you still get these issues after a `git clean -fxd`?

AllSeeingEyeTolledEweSew:
I don't think I really answered your comment: it is indeed slow. It needs to build dependencies from source (openssl and boost). We don't really have a choice in that, as the CentOS build images are (intentionally) ancient.

I applied all the straightforward optimizations I could think of. There's probably more time to gain (curiously the most time-consuming part of setup is extracting the boost source).

But these run times shouldn't block PRs. The workflow should really only run once per release.

arvidn:
> That's orders of magnitude longer than it should be. What's ./test/rootfs? Is your checkout clean?

No, I have lots of stuff in my repo, and I would like to keep it. I take it this will pull in everything in the libtorrent directory then, whether it's needed or not.

AllSeeingEyeTolledEweSew:
> No, I have lots of stuff in my repo, and I would like to keep it. I take it this will pull in everything in the libtorrent directory then, whether it's needed or not.

Yes. On Linux it creates a docker container, copies the whole directory, and builds in isolation. I don't think cibuildwheel has options to limit what gets copied.

cas--:
Normally would use a `.dockerignore` file, wonder if that would work here?

AllSeeingEyeTolledEweSew:
> With Docker, normally would use a `.dockerignore` file, wonder if that would work here?

I believe `.dockerignore` is only used for a `COPY` in a `Dockerfile`, but *this* copy is done by `cibuildwheel` to copy stuff into a running container.

The copy takes less than a second on a fresh checkout (see [the PR check](https://github.com/arvidn/libtorrent/runs/4026939681?check_suite_focus=true#step:6:211)), and we don't expect `cibuildwheel` to be run locally except when debugging Linux builds, so I don't think we need to optimize copying, or do work to defend against interference from local build artifacts.

AllSeeingEyeTolledEweSew:
I changed the implementation from the `run_cibuildwheel.py` wrapper script to just using declarative config in `pyproject.toml`.  Their config system is now robust enough to support everything we were doing before, and I think it's slightly easier to read.

Full workflow (all build flavors) [here](https://github.com/AllSeeingEyeTolledEweSew/libtorrent/actions/runs/1497916100)

.github/workflows/cibuildwheel.yml
```diff
@@ -0,0 +1,142 @@
...
+      publish:
+        description: Write 'PUBLISH' to publish to pypi. BEWARE! ARTIFACTS ARE IMMUTABLE AND CANNOT BE REPLACED ONCE PUBLISHED!
+      publish_test:
+        description: Write 'PUBLISH_TEST' to publish to test-pypi. BEWARE! ARTIFACTS ARE IMMUTABLE AND CANNOT BE REPLACED ONCE PUBLISHED!
+
+  pull_request:
+    paths:
+      - .github/workflows/cibuildwheel.yml
+      - tools/cibuildwheel/**
+      - pyproject.toml
```
arvidn, line 32: this is assuming that anything that would break cibuildwheel would also break the existing builds of the python bindings, right?
Isn't this the only place where the binding is built in an anylinux environment though?

arvidn:
I checked out your branch in a clean repo, but without `run_cibuildwheel.py`, how do I build this locally now? It seems the commands are now hidden inside github actions

tools/cibuildwheel/manylinux/build_utils.sh
```diff
@@ -0,0 +1,63 @@
...
+    local sha256=$2
+    check_var ${sha256}
+
+    echo "${sha256}  ${fname}" > ${fname}.sha256
+    sha256sum -c ${fname}.sha256
+    rm -f ${fname}.sha256
+}
+
+
+function do_standard_install {
```
arvidn, line 49: I don't see this function being used anywhere

.github/workflows/cibuildwheel.yml
AllSeeingEyeTolledEweSew: Changes to the referenced files could break `cibuildwheel` without breaking builds of the python bindings.

`cibuildwheel` is the easiest (or only) way to use `manylinux`/`musllinux` environments, yes.

Not sure if this answers what you're asking, but I expect `cibuildwheel` to be redundant with the other python build workflows, except when we want the specialized environments for building release wheels.

tools/cibuildwheel/manylinux/build_utils.sh
AllSeeingEyeTolledEweSew: These scripts are vendored from https://github.com/pypa/manylinux/tree/main/docker/build_scripts. I took them as a semi-official reference for how to build up-to-date openssl. (`manylinux`'s build scripts build openssl as a dependency for other build tools, then delete openssl, because they don't want to be on the hook to deploy updates for heartbleed 2.0).

I thought it would be nice to add a cron workflow to automatically pull these scripts from their repo, but I wanted to keep this PR simple

AllSeeingEyeTolledEweSew:
> I checked out your branch in a clean repo, but without `run_cibuildwheel.py`, how do I build this locally now? It seems the commands are now hidden inside github actions

Now it's just `python -m cibuildwheel`.

I used the `pypa/cibuildwheel` action in the workflow definition because
* It shortens the workflow definition for readability (we don't even need `actions/setup-python`)
* Automatic updates of `cibuildwheel` are very simple with dependabot, per https://cibuildwheel.readthedocs.io/en/stable/faq/#option-1-github-action

.github/workflows/cibuildwheel.yml
AllSeeingEyeTolledEweSew: In theory we could remove the existing python bindings workflows and just use `cibuildwheel` with limited flavor selection instead for all PRs, but I think there are a few disadvantages:
* The `cibuildwheel` workflow builds openssl from source on Linux, which takes longer
* The `cibuildwheel` workflow wouldn't test against distro versions of openssl, which can be nice

arvidn:
I get this build failure. I'll investigate:

```
  src/module.cpp: In function ‘void init_module_libtorrent()’:
  src/module.cpp:34:5: error: ‘Py_Initialize’ was not declared in this scope
       Py_Initialize();
       ^~~~~~~~~~~~~
  src/module.cpp:34:5: note: suggested alternative: ‘PyPy_IsInitialized’
       Py_Initialize();
       ^~~~~~~~~~~~~
       PyPy_IsInitialized

      "g++"   -fPIC -O3 -finline-functions -Wno-inline -Wall -Wno-deprecated-declarations -fvisibility-inlines-hidden -fPIC -fvisibility=hidden -DBOOST_ALL_NO_LIB -DBOOST_ASIO_ENABLE_CANCELIO -DBOOST_ASIO_HAS_STD_CHRONO -DBOOST_MULTI_INDEX_DISABLE_SERIALIZATION -DBOOST_PYTHON_STATIC_LIB -DBOOST_SYSTEM_NO_DEPRECATED -DBOOST_SYSTEM_STATIC_LINK=1 -DNDEBUG -DOPENSSL_NO_SSL2 -DTORRENT_USE_I2P=1 -DTORRENT_USE_LIBCRYPTO -DTORRENT_USE_OPENSSL -D_FILE_OFFSET_BITS=64 -D_WIN32_WINNT=0x0600  -I"../../include" -I"../../include/libtorrent" -I"/opt/python/pp37-pypy37_pp73/include" -I"/tmp/boost" -I"/usr/sfw/include" -I"src"  -c -o "bin/gcc-8/release/address-model-64/crypto-openssl/libtorrent-python-pic-on/lt-visibility-hidden/python-3.7/src/module.o" "src/module.cpp"
```

arvidn:
interestingly, with this patch it builds:

```
diff --git a/bindings/python/src/boost_python.hpp b/bindings/python/src/boost_python.hpp
index ed636e4b2..2abda563f 100644
--- a/bindings/python/src/boost_python.hpp
+++ b/bindings/python/src/boost_python.hpp
@@ -7,6 +7,7 @@
 
 #include <cstdio>
 #include <libtorrent/aux_/disable_warnings_push.hpp>
+#include <Python.h>
 #include <boost/python.hpp>
 
 #include <boost/bind/placeholders.hpp>
diff --git a/bindings/python/src/module.cpp b/bindings/python/src/module.cpp
index 305f4872f..89837aaa3 100644
--- a/bindings/python/src/module.cpp
+++ b/bindings/python/src/module.cpp
@@ -7,6 +7,7 @@
 #endif
 
 #include <boost/python/module.hpp>
+#include "boost_python.hpp"
 
 void bind_utility();
```

but cibuildwheel still fails with:

```
Successfully installed libtorrent-1.2.15
    + sh -c 'cd /project/bindings/python && python test.py'
  File "test.py", line 5, in <module>
    import libtorrent as lt
ImportError: libtorrent-rasterbar.so.10.0.0: cannot open shared object file: No such file or directory

                                                              ✕ 5.83s
Error: Command ['sh', '-c', 'cd /project/bindings/python && python test.py'] failed with code 1. 
```

AllSeeingEyeTolledEweSew:
That `PyPy_IsInitialized` indicates it's trying to build wheels for pypy. I had disabled that in the `run_cibuildwheel.py` wrapper but apparently forgot to port that config to `pyproject.toml`. I've updated that now.

The workflows specifically select cpython currently, so it didn't get caught there.

It might be easy to get pypy to build and work, but I didn't want to do it in this PR.

arvidn:
ok, that seems to fix the build failure. I'm still seeing this though:

```
Processing /tmp/cibuildwheel/repaired_wheel/libtorrent-1.2.15-cp37-cp37m-manylinux_2_12_x86_64.manylinux2010_x86_64.whl
Installing collected packages: libtorrent
Successfully installed libtorrent-1.2.15
    + sh -c 'cd /project/bindings/python && python test.py'
  File "test.py", line 5, in <module>
    import libtorrent as lt
ImportError: libtorrent-rasterbar.so.10.0.0: cannot open shared object file: No such file or directory

                                                              ✕ 5.89s
Error: Command ['sh', '-c', 'cd /project/bindings/python && python test.py'] failed with code 1. 
```

arvidn:
I found that you can see the build log with:

```
CIBW_BUILD_VERBOSITY="3" python -m cibuildwheel --platform linux
```

Which reveals that the main libtorrent library is *not* built, just the python bindings. So it's really no surprise that running the tests fail saying the library isn't found.

I suppose it's reasonable to just build the binding, but I don't understand why the `pyptoject.toml` then is also trying to run the tests. That won't ever work.

Am I misunderstanding something? Is it supposed to also build the main library?

arvidn:
I see, it's linking the python binding statically against both libtorrent and boost:

```
b2 boost-link=static libtorrent-link=static crypto=openssl deprecated-functions=on variant=release address-model=64 python=3.7
```

clearly the python module that's actually loaded by the test wasn't linked statically

arvidn:
is the virtual machine torn down and deleted every time? I'm interested in being able to look around where files end up and analyze the build output. Is there a way to do that?

cas--:
@arvidn You can use upload artifact action to store output from a job and download it 

https://docs.github.com/en/actions/advanced-guides/storing-workflow-data-as-artifacts#downloading-or-deleting-artifacts

arvidn:
@cas-- as far as I understand, it's working in CI. I'm interested in being able to reproduce the build locally.

AllSeeingEyeTolledEweSew:
I can't think of why the linking wouldn't work. I suspect the test is picking up the wrong build artifact again, somehow.

You could use `CIBW_TEST_SKIP="*"`, which will mean the (broken) wheel will get copied out of the container, and can be inspected. If that looks intact (no missing link to `libtorrent-rasterbar.so`), I'd be more confident a build artifact is getting picked up.

Unfortunately it looks like `cibuildwheel` unconditionally removes the container, so you can't inspect intermediary files used in building.

You could start the docker container manually, and run each step manually in the container.

I'm not sure why I haven't been able to reproduce this failure locally. It should be exactly the same.

AllSeeingEyeTolledEweSew:
:tada:

AllSeeingEyeTolledEweSew:
FYI: I believe manually-triggered workflows must exist in the default branch (`RC_2_0`) before they can be run against any branch, so you may not see an option to run this until it's merged through

jkterry1:
@AllSeeingEyeTolledEweSew I'm finally trying to fully put this into production for the project I was telling you about and me and my collaborators had a few questions I wanted to ask you. Could you please email me at jkterry0@gmail.com?

PR #6191, opened 2021-05-08
Restrict some github actions by path

AllSeeingEyeTolledEweSew:
This turns on path-based restrictions for some github actions.

This adds a risk of missing some builds/tests if we don't get the paths right, but the upside is hopefully faster development speed.

If you think this is a good idea, I think we should also refactor `{linux|macos|windows}.yml` such that e.g. the build jobs are all in one workflow file, the simulation jobs in another, and the "dist" job in another.

### DIFF
--- a/.github/workflows/c.yml
+++ b/.github/workflows/c.yml
@@ -3,7 +3,11 @@ name: C bindings
 on:
    push:
       branches: [ RC_1_2 RC_2_0 master ]
+      paths-ignore:
+         - bindings/python/**
    pull_request:
+      paths-ignore:
+         - bindings/python/**
 
 jobs:
   test:
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -3,7 +3,11 @@ name: Android
 on:
    push:
       branches: [ RC_2_0 master ]
+      paths-ignore:
+         - bindings/python/**
    pull_request:
+      paths-ignore:
+         - bindings/python/**
 
 env:
   NDK_VERSION: "r21d"
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -3,7 +3,11 @@ name: MacOS
 on:
    push:
       branches: [ RC_1_2 RC_2_0 master ]
+      paths-ignore:
+         - bindings/python/**
    pull_request:
+      paths-ignore:
+         - bindings/python/**
 
 jobs:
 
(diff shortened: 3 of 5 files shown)


AllSeeingEyeTolledEweSew:
It's come up a lot for me since I spend so much time on the python bindings :) There's no need to re-run the C++ tests and simulations, if I only touch stuff in `bindings/python`. I plan to make a lot of small PRs to fix some of the issues I've filed, so that could translate to a lot of wasted cpu time.

It's also possible to use `paths-ignore`, for the specific case of skipping C++ tests/simulations when I only touch stuff in `bindings/python`

arvidn:
I see. I suppose it would make sense to just run the python tests when just changing the python bindings.

.github/workflows/docs.yml
```diff
@@ -3,7 +3,19 @@ name: Documentation
...
+      - examples/**
+      - tools/**
+      - bindings/python/simple_client.py
    pull_request:
+      paths:
+      - docs/**
+      - include/**
+      - examples/**
+      - tools/**
+      - bindings/python/simple_client.py
```
arvidn, line 18: I don't think `simple_client.py` affects the documentation. Neither does `examples` or `tools`

.github/workflows/c.yml
```diff
@@ -3,7 +3,23 @@ name: C bindings
...
+         - "!bindings/c/tools/**"
    pull_request:
+      paths:
+         - src/**
+         - include/**
+         - deps/**
+         - Jamfile
+         - Jamroot.jam
+         - bindings/c/**
+         - "!bindings/c/tools/**"
```
arvidn, line 22: I take it these are python tools. I think this might be a bit too granular. I'm worried about there being a mistake in these cut-outs, so the simpler the better I think

arvidn (commented):
I think this looks good, except for the comments

.github/workflows/docs.yml
AllSeeingEyeTolledEweSew: The docs include other files, including `simple_client.py` and various stuff from `examples`. I thought something included `tools` but I can't find it now, removed.

.github/workflows/c.yml
AllSeeingEyeTolledEweSew: Hm, these are all files to build the c bindings (I believe).

I think you may have read this as a disallow-list ("don't run workflow if only these files modified"), when it's actually an allow-list ("only run workflow if any of these files modified").

The disallow-list is probably the safer option. I'll change it to this.

arvidn: oh, yeah. I assumed the `!`-prefix meant "not"

arvidn (approved):

tesla1605 (commented):
b

PR #6193, opened 2021-05-09
CMake: misc improvements

FranciscoPombal:
- Fix CMP0115 warnings on recent CMake versions.
- Make minimum required CMake version more consistent across library, 
bindings and tests. It can't quite be the same due to CI constraints.
- Sort compile features/compile options/sources listings.
- Move a couple of sources to the correct target_sources() call.
- Minor indentation fixes.

---

Targets `RC_1_2`, let me know if you'd like me to repeat for `RC_2_0` or if you'll do it in one of your `test_merge`s.


FranciscoPombal:
Depends on https://github.com/arvidn/libtorrent/pull/6195.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.15.0 FATAL_ERROR) # Configurable policies: <= CMP0094
+cmake_minimum_required(VERSION 3.17.0 FATAL_ERROR) # Configurable policies: <= CMP0094
```
arvidn, line 1: the LGTM service uses a somewhat old cmake, and its analysis of libtorrent depends on it (this is why `master` isn't analyzed right now). I think this is risky, but I can't recall which version LGTM requires

FranciscoPombal: @arvidn I also could not find information about this. However, I forced LGTM to run and it succeeded, so it seems to be fine.

FranciscoPombal:
Force-pushed to correct `# Configurable policies: <= CMP0102` comment.

arvidn:
appveyor is stil failing with cmake: https://ci.appveyor.com/project/arvidn/libtorrent/builds/39099491/job/xiuvdih46bnt2dgh

FranciscoPombal:
@arvidn 

> appveyor is stil failing with cmake: https://ci.appveyor.com/project/arvidn/libtorrent/builds/39099491/job/xiuvdih46bnt2dgh

I know, that's why I submitted https://github.com/arvidn/libtorrent/pull/6195. No idea what the error that that particular PR is running into is about though.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.15.0 FATAL_ERROR) # Configurable policies: <= CMP0094
+cmake_minimum_required(VERSION 3.17.0 FATAL_ERROR) # Configurable policies: <= CMP0102
```
arvidn, line 1: I don't quite get this comment. What part of this patch requires a new cmake?

FranciscoPombal: @arvidn

The minimum version specified for `test/CMakeLists.txt` was extremely outdated. As a result, useless noisy deprecation warnings were always printed.

`bindings/python/CMakeLists.txt` requires at least 3.17, so that was chosen as the target for all CMLs (in the commit message: `Unify minimum required versions across library, bindings and tests`). The minimum version for `test/CMakeLists.txt` could have been bumped to an earlier version just for silencing the deprecation warnings, but since we're bumping it, might as well go for consistency. The top-level CML doesn't actually require anything from 3.17 and again was done for consistency.

An additional benefit is that potential future improvements are less constrained by availability of features. If it were 100% up to me I would just bump the minimum to 3.20, since that is also available on Appveyor :).

FranciscoPombal:
@arvidn OK, I dialed the minimum version bump back to 3.16, so that now all CI should pass without needing any changes.

FranciscoPombal:
`Linux / Tests (toolset=clang)` also seems to failing for other PRs.

What about `Windows / Test (pull_request)`? Is it acceptable to fail? None of the present changes should cause it to fail in comparison to other PRs. Maybe it's spurious? Or maybe it was previously passing due to relying on some very old CMake 2.6 behavior?

CMakeLists.txt
arvidn: updating the appveyor image is non-trivial though, so bumping it to the point where that has to happen will expand the scope of this change.

FranciscoPombal: @arvidn Yes, see my most recent comments on both PRs, I have made it so that updating the appveyor image is not needed to merge the improvements in this PR.

arvidn:
> Linux / Tests (toolset=clang) also seems to failing for other PRs.

Yeah, I haven't figured out what's going on there yet. It works fine for me locally with clang-9 on ubuntu.

> What about Windows / Test (pull_request)? Is it acceptable to fail? None of the present changes should cause it to fail in comparison to other PRs. Maybe it's spurious?

Yes, that failure is an old issue. It's so benign that I haven't bothered addressing it.

arvidn (approved):

PR #6195, opened 2021-05-10
AppVeyor: switch the CMake build to Ninja Multi-Config generators for easier build parallelization

FranciscoPombal:
- About `/p:UseMultiToolTask=true`: see https://discourse.cmake.org/t/parallel-does-not-really-enable-parallel-compiles-with-msbuild/964 and https://devblogs.microsoft.com/cppblog/improved-parallelism-in-msbuild/ linked at the bottom of that discourse thread.

- Removed the `RC_*` branches, but let me know if they should still be there.

---

PR now targets `master` with only a minor change.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,9 +2,6 @@ version: "{build}"
 branches:
   only:
     - master
-    - RC_2_0
-    - RC_1_2
-    - RC_1_1
 image: Visual Studio 2019
 clone_depth: 1
 environment:
@@ -93,13 +90,12 @@ build_script:
   # we need to build the boost libraries we use with C++17
   # and stage it for cmake to pick up
   - if defined cmake (
+    call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars64.bat" &&
     cd %BOOST_ROOT% &&
     b2.exe cxxstd=17 release --with-python --with-system --layout=system address-model=64 link=shared stage &&
     cd %ROOT_DIRECTORY% &&
-    mkdir build &&
-    cd build &&
-    cmake -DBOOST_LIBRARYDIR=%BOOST_ROOT%\stage\lib -DCMAKE_CXX_STANDARD=17 -Dbuild_tests=ON -Dbuild_examples=ON -Dbuild_tools=ON -Dwebtorrent=%webtorrent% -DOPENSSL_ROOT_DIR=%ssl_root_dir% -Dpython-bindings=%python% -Dboost-python-module-name="python" -Dskip-python-runtime-test=true -DPython_ADDITIONAL_VERSIONS="2.7" -G "Visual Studio 16 2019" -A x64 .. &&
-    cmake --build . --config Release --parallel %NUMBER_OF_PROCESSORS% -- -verbosity:minimal
+    cmake -B build -DBOOST_LIBRARYDIR=%BOOST_ROOT%\stage\lib -DCMAKE_CXX_STANDARD=17 -Dbuild_tests=ON -Dbuild_examples=ON -Dbuild_tools=ON -Dwebtorrent=%webtorrent% -DOPENSSL_ROOT_DIR=%ssl_root_dir% -Dpython-bindings=%python% -Dboost-python-module-name="python" -Dskip-python-runtime-test=true -DPython_ADDITIONAL_VERSIONS="2.7" -G "Ninja Multi-Config" &&
+    cmake --build build --config Release
     )
 
 test_script:


FranciscoPombal:
@arvidn 

> I think this looks good, but it should target `master`.

Any particular reason why it can't target both `RC_1_2` and `RC_2_0`? This is a prerequisite for https://github.com/arvidn/libtorrent/pull/6193, unless you would really like for me to lower the CMake version in that PR.

> Also, it's odd that this error showed up: https://ci.appveyor.com/project/arvidn/libtorrent/builds/39087199/job/jxa4ue5oguhg39nj#L80

Any idea why it did and how it can be fixed?

FranciscoPombal:
Force pushed, rebasing on the latest `RC_1_2`. Also removed the override of `CMAKE_CXX_STANDARD` on the CMake build. CMake will figure out all the correct flags it needs to pass to the MSVC tools based on the `target_compile_features()` we now provide in the main CML. Let's see if we still get the same error.

FranciscoPombal:
Alright, for now I'm simply disabling just enough `warnings-as-errors` for it to pass.

FranciscoPombal:
@arvidn 

I have disabled `warnings-as-errors` where necessary for the CI to pass. Is it acceptable, can the resolution of these warnings be left for later? I guess these problems were always there, but now the upgrade to the MSVC toolchain has uncovered them.

However, now a new error has occurred: https://ci.appveyor.com/project/arvidn/libtorrent/builds/39292599/job/jbdmxquafa72c8h9#L2100. Do you think it is spurious?

Also, I noticed a bunch of these messages: https://ci.appveyor.com/project/arvidn/libtorrent/builds/39292599/job/61u1e3ti12vgcca9#L341. Could these be caused by the way AppVeyor compiles their boost binaries, or is there something to worry about here? I think this is the case, since the GCC version used for building libtorrent is new enough https://ci.appveyor.com/project/arvidn/libtorrent/builds/39292599/job/61u1e3ti12vgcca9#L32, and it is compiled with at least `-std=c++11`.

arvidn:
> Any particular reason why it can't target both RC_1_2 and RC_2_0? This is a prerequisite for #6193, unless you would really like for me to lower the CMake version in that PR.

The main reason I'm hesitant is that `RC_1_2` is committed to work on old compilers (C++11) and a relatively old boost. Likewise `RC_2_0` is committed to only require C++14. The boost versions are a bit looser I suppose. But the risk of updating the compiler in older branches is that they will silently accept code changes that require that new version. (I've made the mistake of including C++14 code in RC_1_2 for instance).

Most of this is caught by the `cxxstd=11` and `cxxstd=14` build flags, but in practice it's not the C++ version that matters, it's the compiler version.

Bumping the compiler version also involves dealing with new warnings. Since warnings are treated as errors. Some of these may have been fixed in more recent branches, if they run a newer version.

e.g. https://ci.appveyor.com/project/arvidn/libtorrent/builds/39087199/job/jxa4ue5oguhg39nj#L79

I understand the motivation for this is to be able to use a newer version of cmake. In your other PR, that bumps cmake version, I don't see any other changes in there that look like they would require a new version. Is it just to avoid bugs (or unfortunate behavior) in older versions of cmake?

FranciscoPombal:
@arvidn 

> The main reason I'm hesitant is that `RC_1_2` is committed to work on old compilers (C++11) and a relatively old boost. Likewise `RC_2_0` is committed to only require C++14. The boost versions are a bit looser I suppose. But the risk of updating the compiler in older branches is that they will silently accept code changes that require that new version. (I've made the mistake of including C++14 code in RC_1_2 for instance).
> 
> Most of this is caught by the `cxxstd=11` and `cxxstd=14` build flags, but in practice it's not the C++ version that matters, it's the compiler version.
> 
> Bumping the compiler version also involves dealing with new warnings. Since warnings are treated as errors. Some of these may have been fixed in more recent branches, if they run a newer version.
> 
> e.g. https://ci.appveyor.com/project/arvidn/libtorrent/builds/39087199/job/jxa4ue5oguhg39nj#L79
> 
> I understand the motivation for this is to be able to use a newer version of cmake. In your other PR, that bumps cmake version, I don't see any other changes in there that look like they would require a new version. Is it just to avoid bugs (or unfortunate behavior) in older versions of cmake?

OK, I understand your reasoning. I will change the other PR to only bump CMake to 3.16 on everything except the python bindings (which actually require 3.17). The rest was being bumped to 3.17 as well for consistency. However, it seems that the cascade of changes that such a bump requires (especially pertaining to CI) is more trouble than it's worth for `RC_1_2` and `RC_2_0`.

FranciscoPombal:
@arvidn PR now targets `master` with a simple change.

arvidn:
appveyor skipped building this, with this message:

> Commit "b91c7383" skipped as branch "RC_1_2" is not in white-list

I wonder if appveyor doesn't support updating the target branch for a PR. I'll try close and re-open to re-trigger CI.

FranciscoPombal:
@arvidn success :)

arvidn:
it doesn't seem to have improved the run-time on CI though...

arvidn:
are you sure `--parallel %NUMBER_OF_PROCESSORS%` doesn't work on windows?

FranciscoPombal:
@arvidn 

> it doesn't seem to have improved the run-time on CI though...

Yeah, I noticed that. How many CPU cores are appveyor runners supposed  to have? No speed up at all would suggest they only have 1.

> are you sure `--parallel %NUMBER_OF_PROCESSORS%` doesn't work on windows?

It does not work correctly with VS generators. With VS generators, `--parallel` affects MSBuild's `/m` flag, which just tells it "compile multiple projects at the same time" (which is useless, because there is only 1 "project"), not "compile multiple translations units at the same time" (which woud be controlled by the`/p:CL_MPcount` flag, which in turn is simply a frontend for # `cl.exe`'s `/MP` flag).

Most often, especially with CMake-based builds, you have/want a single project with multiple translation units compiled simultaneously, so you'd want to set `/p:CL_MPcount` to the number of cores in the machine. However, a while ago the new multi-tool task flag was introduced recently to auto-manage everything (i.e., ensure `/m * /MP = nproc`, guaranteeing resource usage is maximized while not submitting more simultaneous jobs than there are cores), so that is what should be used. It is possible that in the future CMake will set it natively when using the VS 2019 generator or a later VS generator. See https://discourse.cmake.org/t/parallel-does-not-really-enable-parallel-compiles-with-msbuild/964 

Another valid workaround is to just use the ninja generstor. It automatically uses all cores in the machine, you don't even to manually set `--parallel`.

arvidn:
> Yeah, I noticed that. How many CPU cores are appveyor runners supposed to have? No speed up at all would suggest they only have 1.

The virtual machine says it has 2 CPUs, but I wouldn't be surprised if it's effectively simulated on a single CPU. I can't recall ever having seen speedups when compiling in parallel on appveyor.

FranciscoPombal:
@arvidn In any case, this is still better.

FranciscoPombal:
@arvidn You could always switch to using the Ninja generator, which automatically uses all cores by default. Any particular reason to stick with VS generators?

arvidn:
Not that I can think of. I think Ninja is the best part of cmake :)

FranciscoPombal:
@arvidn 

> Not that I can think of. I think Ninja is the best part of cmake :)

So should I change this PR to implement such a change?

arvidn:
yeah, I think using ninja would be fine. I'm not too optimistic it will actually make a difference in CI runtime though

FranciscoPombal:
@arvidn done, commit message updated.

arvidn:
https://ci.appveyor.com/project/arvidn/libtorrent/builds/39399704/job/5pdk1x29vim112dq#L211

FranciscoPombal:
@arvidn

It was a small stupid mistake, fixed now.

arvidn:
I'm not sure what might have changed to cause this: https://ci.appveyor.com/project/arvidn/libtorrent/builds/39401528/job/0iyu9l2guts7ns2c#L8169

arvidn:
could it be that `_MSC_VER` doesn't get defined, so `intrin.h` isn't included? it seems odd since that macro should be defined by the compiler afaik.

FranciscoPombal:
@arvidn It's picking up the wrong compiler - https://ci.appveyor.com/project/arvidn/libtorrent/builds/39401528/job/0iyu9l2guts7ns2c?fullLog=true#L211

arvidn (approved):

arvidn:
interesting. presumably building with clang on windows is broken then...

Issue #6201, opened 2021-05-12
Active+Finished duration get doubled on queued torrents


sledgehammer999:
libtorrent version (or branch):
RC_1_2 22af326f6c063af519e1a7fd02a9070c255a759c
(it has been observed with older code too)

platform/architecture:
Windows 7 sp1 x64

compiler and compiler version:
mingw 9.2.0 / msvc2017

**Terminology used**
queued == auto_managed && paused

This issue is observed only when a torrent enters the queued state upon finishing downloading. This can be easily achieved by setting active_seeds to 0.

While a torrent is downloading the `torrent_status::active_duration` counter updates normally. When the torrent finishes and enters the queued state the value of the counter abruptly doubles. This can happen either immediately or even up to ~80 seconds after the torrent is queued. Sometimes it doesn't happen at all, but this is rare. So maybe a race issue somewhere?

Moreover, if a torrent after finishing continues to seed without queuing this doubling doesn't seem to happen. Both `torrent_status::active_duration` and `torrent_status::finished_duration` continue to update correctly. But if I change active_seeds to 0, the torrent is queued and these 2 counters get their valued doubled. <--- this doesn't happen consistently though

Backstory: In qbt we present to users the average down speed using the equation `avg = all_time_download / (active_duration - finished_duration)`. I had observed that the avg speed was instantly halved upon torrent finished. And up until now I didn't bother investigating further into the matter.

arvidn:
I've been trying to reproduce this in the `test_torrent_status` simulation. However, it doesn't happen for me.

I suspect what's going on is that `is_paused()`, for some reason, returns `false`, despite the `do_pause()` function having been called, moving the time (current_time - m_started) into `m_active_time`. That would make that time count twice.

arvidn:
when this doubling of `active_duration` happens, is it persistent? or does it come back down to the value you would expect?

arvidn:
I believe this fixes it. https://github.com/arvidn/libtorrent/pull/6205
This issue highlights that the logic around pausing, and especially graceful-pause is unnecessarily complex

sledgehammer999:
>when this doubling of active_duration happens, is it persistent? or does it come back down to the value you would expect?

I have never seen it go back to its proper value. Even after a program restart.

PR #6205, opened 2021-05-17
active_duration

arvidn:
fix incorrect reporting of active_duration when entering graceful-pause


sledgehammer999:
It seems I can't reproduce #6201 with this PR. So it probably fixes that issue.

PR #6206, opened 2021-05-17
Windows XP compatible branch based on libtorrent 1.2.3 version

gtumanyan:
Libtorrent 1.2.3 adopted to boost >1.75 and cmake >3.17
Vista checks removed to work on Windows XP
Restored the ability to download a torrent file via URL
Tested to build with cmake 3.20 and visual studio 2019
Works on Windows XP SP3

### DIFF
--- a/Jamfile
+++ b/Jamfile
@@ -265,7 +265,18 @@ rule building ( properties * )
 		result += <build>no ;
 	}
 
+<<<<<<< HEAD
 	if <toolset>msvc in $(properties)
+=======
+	local VERSION = [ feature.get-values <cxxstd> : $(properties) ] ;
+	if ! $(VERSION) || $(VERSION) < 14
+	{
+		ECHO "libtorrent requires at least C++14. Specify cxxstd=14 or higher" ;
+		result += <build>no ;
+	}
+
+	if <toolset>msvc in $(properties) || <toolset>intel-win in $(properties)
+>>>>>>> 80affa04f (make the default build for libtorrent be cxxstd=14)
 	{
 		# allow larger .obj files (with more sections)
 		result += <cflags>/bigobj ;
@@ -815,6 +826,7 @@ lib torrent
 
 	: # default build
 	<threading>multi
+	<cxxstd>14
 	<c++-template-depth>512
 
 	: # usage requirements
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -549,7 +549,7 @@ if (WIN32)
 			debug dbghelp
 	)
 
-	add_definitions(-D_WIN32_WINNT=0x0600) # target Windows Vista or later
+	add_definitions(-D_WIN32_WINNT=0x0500) # target Windows XP or later
 
 	target_compile_definitions(torrent-rasterbar
 		PUBLIC WIN32_LEAN_AND_MEAN # prevent winsock1 to be included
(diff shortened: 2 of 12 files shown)


arvidn:
is your branch rebased on top of `RC_1_2`? it looks like you have some mysterious commits in there. Did you merge?
If so, please undo the merge and rebase instead.

Jamfile
```diff
@@ -265,7 +265,18 @@ rule building ( properties * )
...
+=======
+	local VERSION = [ feature.get-values <cxxstd> : $(properties) ] ;
+	if ! $(VERSION) || $(VERSION) < 14
+	{
+		ECHO "libtorrent requires at least C++14. Specify cxxstd=14 or higher" ;
+		result += <build>no ;
+	}
+
+	if <toolset>msvc in $(properties) || <toolset>intel-win in $(properties)
+>>>>>>> 80affa04f (make the default build for libtorrent be cxxstd=14)
```
arvidn, line 279: this isn't right

arvidn:
also, make sure you don't change the submodule

FranciscoPombal:
I strongly disagree with this. Why should everyone have to put up with compatibility changes made to support a proprietary OS that has been unsupported for almost 8 years now (no, the hacks made to enable receiving "extended support updates" for consumers don't count when counting the amount of years since support was officially dropped)?

At the very least, this should be gated behind a feature flag in the build system, disabled by default.

The defines and checks should be bumped to >=Windows 8.1, too, or at the very least, >= Windows 7 (but that's a discussion for another PR).

arvidn:
most of this patch is just broken merge/rebase issues. It's hard to know which changes are intended. Some changes seem reasonable (like `#include <windows.h>` -> `#include "libtorrent/_aux/windows.h"`). I think I made those changes in more recent versions. But it also seems of questionable value to back-port them.

The change to `client_test` to use a deprecated feature (i.e. download torrents via http) is not appropriate. It's deprecated because it has been removed.

I agree with @FranciscoPombal. The default build configuration should assume a more recent version of windows. If the changes to support building for older versions can be small, I'm OK with landing those. I would prefer to just cut features not supported by older operating systems (when building in this mode), rather than implementing the feature in libtorrent.

Anyway, this may still be premature, because we don't really know what change you're proposing, @gtumanyan

gtumanyan:
The patch is based on the 1_2_3, not the RC_1_2 head. There are too many changes in RC_1_2 since 1_2_3 that break XP support.
I'm new to git and didn't commit every change made, that's my fault.
Nonetheless, the client runs on XP and builds with the latest tools.
> The change to `client_test` to use a deprecated feature (i.e. download torrents via http) is not appropriate. It's deprecated because it has been removed.
The server that hosts the .torrent file doesn't support web-seed yet, download torrents via URL will be removed when web-seed be implemented on the server.

arvidn:
> The patch is based on the 1_2_3, not the RC_1_2 head. There are too many changes in RC_1_2 since 1_2_3 that break XP support.
> Maybe it would be better to create a dedicated XP-legacy branch?

I see, there is no branch for 1.2.3, that's just a tag on an previous version of `RC_1_2` (which is the head of the 1.2.x version).
In order for any patch to be included in a release it has to go into one of the branches that releases are cut from. those are `master`, `RC_2_0` or `RC_1_2`, although that last one probably won't have that many more releases from it.

I'm not sure what to do with a legacy branch, it there's no release.

gtumanyan:
There is no branch for those who need XP support. Maybe create a new one?

PR #6208, opened 2021-05-18
DRAFT: Questions for asynchronous IO

micvbang:
Hello,

As noted in the title this PR is a draft and is not meant to be merged. I'm posting my progress in adding asynchronous IOs to libtorrent and want to give some context for a few questions that I have in regards to whether and how to continue this work.

For context, I saw your [blog post from 2012 about async IOs](https://blog.libtorrent.org/2012/10/asynchronous-disk-io/) and got interested in trying to improve the way that IOs are done in libtorrent. I know I'm almost 10 years late to the party, but it seems to me that the implementation of IOs in libtorrent hasn't architecturally changed a lot from what you described in that post.

I'm working on the open source project [xNVMe](https://github.com/OpenMPDK/xNVMe) which provides a simple API that abstracts over underlying IO paths such as io_uring, libaio, and posix aio to provide all of the benefits of truly asynchronous IOs while requiring the user (libtorrent in this case) to implement only a single API. The xNVMe API then makes it possible to change the actual IO path at runtime.

This PR contains a simple xNVMe `disk_interface` implementation that is mostly identical to the existing posix disk interface, except that `async_read` and `async_write` have been implemented using xNVMe. This implementation is not yet "truly asynchronous" because IOs are reaped immediately after they are posted.

I tried briefly to implement this in a "truly asynchronous" fashion, but my lacking understanding of the libtorrent architecture made it difficult to come up with a good design. Instead of hacking on it for too long, I thought I'd start out by asking you if this is something that you think is worthwhile to work on and might be interested in merging into libtorrent when it's fully functional, or if the current thread pool model is fine as it is?
If you think it's worthwhile, I'd love to ask for some pointers in how you think this could fit in with the existing architecture:

The main problem I've had in implementing truly asynchronous IOs so far, has been to identify when it's sensible to reap IOs.

Other than that, I've struggled a bit with identifying which existing functionality can be reused in a non-blocking context (e.g. excellent helpers like `readwritev`) as many of them (naturally) assume a blocking behavior.

Do you have any insights or code from your previous experiments with this that could make my life easier?

Thanks!


Jamfile
```diff
@@ -213,6 +213,8 @@ rule linking ( properties * )
 		result += <library>/try_signal//try_signal/<link>static ;
 	}
 
+	result += <linkflags>"-lxnvme -laio -luuid -lnuma -pthread -lrt" ;
```
elgatito, line 216: What about android/windows/darwin? Will it build there?

micvbang: We are currently working on Windows support for xNVMe and expect a release in couple of months. It'll be using I/O Control Ports for the async IO.

xNVMe is currently untested on Darwin, but does run on FreeBSD. Android is also untested, but at least the POSIX interfaces should work, and then depending on the kernel it should support libaio and io_uring. I expect that we'll get test coverage for both Darwin and Android :)

arvidn (commented):
This is quite exciting!
I haven't had time to go through everything in detail yet, and I will have more questions later, but I figured I shouldn't delay posting these too long.

I'm curious about your take on async disk I/O. I kind of dismissed it for these reasons:

1. open()/close()/stat() are all synchronous in all async I/O models (iirc)
2. I expect storage to be moving towards Direct Access, i.e. where it's treated more like persistent RAM, mapped into the address space. With this in mind, multi-threaded mmap seems to be somewhat future proof.

micvbang:
First of all, thanks for the great comments on the code! I'll make sure to address everything noted.

In regards to your questions about async disk I/O:

> 1. open()/close()/stat() are all synchronous in all async I/O models (iirc)

Yep, that's pretty much the case, except for Linux that can do some of these things via io_uring. But nothing is defined in POSIX or otherwise generally available in a portable fashion.

I'm wondering why the lack of asynchronous versions of said calls is enough to dismiss async I/O? There must be some part of the workflow that I'm missing - e.g., is torrenting largely bottlenecked by metadata I/O operations?

My gut feeling would be that caching of file descriptors could amortize the cost of blocking metadata operations to be small enough that the increased performance from async read/write calls would more than make up for it. But I guess I'm basing this on the assumption that many torrents contain files that are big enough to benefit from a higher I/O depth.

This gut feeling is based on nothing, so some of your insight on this would be much appreciated!

> 2. I expect storage to be moving towards Direct Access, i.e. where it's treated more like persistent RAM, mapped into the address space. With this in mind, multi-threaded mmap seems to be somewhat future proof.

What do you mean by Direct Access? Do you mean bypassing the file system using raw block devices? Or do you mean new hardware such as persistent memory? Or something else?

Generally, I think it will be an excellent long-term solution to provide an IO path that utilizes async I/Os because the $/GB of HDD and SSDs is likely to stay much lower than hardware that natively supports byte-addressing/unaligned access.

src/xnvme_disk_io.cpp
```diff
@@ -0,0 +1,439 @@
...
+			// TODO: 3 this const_cast can be removed once iovec_t is no longer a
+			// thing, but we just use plain spans
+			iovec_t const b = {const_cast<char *>(buf), r.length};
+
+			time_point const start_time = clock_type::now();
+
+			storage_error *error = new storage_error();
+
+					// TODO: this function isn't being free()d anywhere right now
+			std::function<void()> *whandler = new std::function<void()>([=, handler = std::move(handler)]() {
```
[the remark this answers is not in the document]
micvbang: I attempted to do something like this myself, but failed. This works great!

src/xnvme_disk_io.cpp
```diff
@@ -0,0 +1,439 @@
...
+
+
+		void async_move_storage(storage_index_t const storage, std::string p
+			, move_flags_t const flags
+			, std::function<void(status_t, std::string const&, storage_error const&)> handler) override
+		{
+			xnvme_storage* st = m_torrents[storage].get();
+			storage_error ec;
+			status_t ret;
+			std::tie(ret, p) = st->move_storage(p, flags, ec);
```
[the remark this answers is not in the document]
micvbang: Yep, agreed. I was considering doing the read/write IOs asynchronously and using a thread pool to perform the file system operations

src/xnvme_disk_io.cpp
```diff
@@ -0,0 +1,439 @@
...
+
+			for (int i = 0; i < blocks_to_read; ++i)
+			{
+				bool const v2_block = i < blocks_in_piece2;
+
+				auto const len = v1 ? std::min(default_block_size, piece_size - offset) : 0;
+				auto const len2 = v2_block ? std::min(default_block_size, piece_size2 - offset) : 0;
+
+				iovec_t b = {buffer.data(), std::max(len, len2)};
+				int const ret = st->readv(m_settings, b, piece, offset, error);
```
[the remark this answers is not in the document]
micvbang: I didn't work on this yet, but it definitely should be possible

arvidn:
> I'm wondering why the lack of asynchronous versions of said calls is enough to dismiss async I/O? There must be some part of the workflow that I'm missing - e.g., is torrenting largely bottlenecked by metadata I/O operations?

If I would make blocking calls in the network thread, yes it would be.

`stat()` is probable the most notable example. When starting a torrent with 1000 files in it, the first thing we need to do is to make 1000 calls to `stat()` to see if one exists.

It's not a show stopper, you just have to make those calls in a thread pool. However, once you have a thread pool, it's natural to use that for the actual I/O as well. Synchronizing opening and closing files with issuing async reads and writes becomes complicated.

Additionally, hashing pieces ought to be done in separate threads too, so it's hard to escape the thread pool.

> My gut feeling would be that caching of file descriptors could amortize the cost of blocking metadata operations to be small enough that the increased performance from async read/write calls would more than make up for it. But I guess I'm basing this on the assumption that many torrents contain files that are big enough to benefit from a higher I/O depth.

Any blocking call is problematic in the main thread, since it would prevent all sockets to respond to events, or peer requests to be responded to. I think most torrents probably do have a few large files, but there are so many torrents in the world that even the minority of many-small-files torrents is still a large number. Both extremes need to be handled well.

> What do you mean by Direct Access? Do you mean bypassing the file system using raw block devices? Or do you mean new hardware such as persistent memory? Or something else?

What linux calls DAX: see [here](https://www.kernel.org/doc/Documentation/filesystems/dax.txt) and [here](https://lwn.net/Kernel/Index/#DAX).

arvidn:
By the way, please don't interpret any of my comments as discouragement!
I think it would be awesome to get this to work and see how it does.

micvbang:
> By the way, please don't interpret any of my comments as discouragement!
> I think it would be awesome to get this to work and see how it does.

Definitely not, your honest feedback is **much** appreciated! 
I think async I/O has got potential to improve the I/O path of libtorrent. But we'll have to test if that's true or not :)

src/xnvme_storage.cpp
```diff
@@ -0,0 +1,806 @@
...
+
+		auto search = m_file_handles->find(fname);
+		if (search != m_file_handles->end()) {
+			return search->second;
+		}
+
+		int xnvme_mode = XNVME_FILE_OFLG_CREATE | XNVME_FILE_OFLG_RDWR;
+		auto xnvme_fname = fname;
+		xnvme_fname.append("?async=" + m_xnvme_backend);
+		xnvme_dev *dev = xnvme_file_open(xnvme_fname.c_str(), xnvme_mode);
```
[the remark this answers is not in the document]
micvbang: Great point. We'll look at redesigning the interface in xNVMe

src/xnvme_storage.cpp
```diff
@@ -0,0 +1,806 @@
...
+	}
+
+	xnvme_file_queue* xnvme_storage::open_file_xnvme(file_index_t idx)
+	{
+
+		std::string fname = files().file_path(idx, m_save_path);
+
+		auto search = m_file_handles->find(fname);
+		if (search != m_file_handles->end()) {
+			return search->second;
```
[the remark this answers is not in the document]
micvbang: > are all file handles opened in both read and write mode? 

Currently they are

> If files are always opened in read/write mode, you may fail to just seed files from read-only media. or volumes your user only has read-access to.

Great point! I'll look into the `file_view_pool` cache

include/libtorrent/aux_/xnvme_storage.hpp
```diff
@@ -0,0 +1,148 @@
...
+		void use_partfile(file_index_t index, bool b);
+
+		file_storage const& m_files;
+		std::unique_ptr<file_storage> m_mapped_files;
+		std::string m_save_path;
+		stat_cache m_stat_cache;
+
+		aux::vector<download_priority_t, file_index_t> m_file_priority;
+
+		std::unordered_map<std::string, xnvme_file_queue*> *m_file_handles;
```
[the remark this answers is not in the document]
micvbang: > this should probably not be a pointer to a heap allocated map, but just a member directly.
Fixed.

> I notice that files are never evicted from this map, until destruction. Keep in mind that there are torrents with tens of thousands of small files. But I suppose this part isn't meant to be "production ready"
Yeah, agreed. This was just to get things moving :)

src/xnvme_disk_io.cpp
```diff
@@ -0,0 +1,439 @@
...
+			, std::function<void(storage_error const&)> handler
+			, disk_job_flags_t) override
+		{
+			// TODO: 3 this const_cast can be removed once iovec_t is no longer a
+			// thing, but we just use plain spans
+			iovec_t const b = {const_cast<char *>(buf), r.length};
+
+			time_point const start_time = clock_type::now();
+
+			storage_error *error = new storage_error();
```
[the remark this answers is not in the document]
micvbang: Definitely

src/xnvme_storage.cpp
```diff
@@ -94,7 +94,7 @@ namespace aux {
 		, m_part_file_name("." + to_hex(p.info_hash) + ".parts")
 	{
 		if (p.mapped_files) m_mapped_files.reset(new file_storage(*p.mapped_files));
-		m_file_handles = new std::unordered_map<std::string, xnvme_file_queue*>();
+		m_file_handles = std::unordered_map<std::string, xnvme_file_queue*>();
```
arvidn, line 97: you don't need this assignment. It will be default constructed already

src/xnvme_storage.cpp
```diff
@@ -707,7 +707,7 @@ namespace aux {
 		}
 
 		xnvme_file_queue *fq = new xnvme_file_queue(dev, queue);
-		m_file_handles->insert(std::make_pair(fname, fq));
+		m_file_handles.insert(std::make_pair(fname, fq));
```
arvidn, line 710: I think you can just say: `m_file_handles.emplace(fname, fq);` here instead.

src/xnvme_storage.cpp
```diff
@@ -272,7 +272,7 @@ namespace aux {
 	, span<iovec_t const> bufs
 	, piece_index_t const piece, int const offset
 	, storage_error &error
-	, std::function<void()> *handler)
+	, std::function<void(storage_error)> handler)
```
arvidn, line 275: `storage_error` is typically passed around as `const&`

src/xnvme_disk_io.cpp
```diff
@@ -120,15 +120,12 @@ namespace {
 			iovec_t buf = {bufz, r.length};
 
 			// TODO: this function isn't being free()d anywhere right now
-			std::function<void()> *whandler = new std::function<void()>([=, handler = std::move(handler)]() {
+			auto whandler = [=, handler = std::move(handler)](storage_error error) mutable {
```
arvidn, line 123: this `=` capture makes me a bit nervous. Does it capture a lot? If you were to explicitly list everything it captures, would that be an unreasonably long list? it seems like it captures `this`

micvbang: Nice catch -- it captures `this`, `bufz, and `start_time`, so I agree it's better to write it out completely.

The capture of `this` gives access to `m_ios` and `m_stats_counters`, and the `libtorrent::buffer_allocator_interface` implemented by `xnvme_disk_io`. 

> it seems like it captures this

I can see it causing trouble if the lifetime of `xnvme_disk_io` is not guaranteed to be longer than the invocation of the callback being declared. Is this (or something else) going to be a potential problem?

arvidn: > I can see it causing trouble if the lifetime of xnvme_disk_io is not guaranteed to be longer than the invocation of the callback being declared. Is this (or something else) going to be a potential problem?

No, I think that's fine

micvbang: Cool!

micvbang:
I'm currently working towards making reads and writes truly asynchronous, i.e. moving the current usage of `xnvme_queue_wait` to somewhere that we can reap IO completions. Is there a main loop where stuff like this can be done? Do you have a suggestion for a good place to do this?

arvidn:
> I'm currently working towards making reads and writes truly asynchronous, i.e. moving the current usage of xnvme_queue_wait to somewhere that we can reap IO completions. Is there a main loop where stuff like this can be done? Do you have a suggestion for a good place to do this?

I see two approaches:

1. Spawn a thread that reaps completions and post callbacks to the io context
2. Somehow integrate with the io context message loop. This is probably a bit complicated to do in a platform independent manner, but something like an `eventfd` could probably be used on linux. To get a callback when a file descriptor becomes  readable, you can use [stream_descriptor](https://www.boost.org/doc/libs/1_76_0/doc/html/boost_asio/reference/posix__stream_descriptor.html) and wait on it by calling [async_wait()](https://www.boost.org/doc/libs/1_76_0/doc/html/boost_asio/reference/posix__basic_stream_descriptor/async_wait.html). On windows I would expect you could integrate with a HANDLE (say, a win32 Event) by using [object_handle](https://www.boost.org/doc/libs/1_76_0/doc/html/boost_asio/reference/windows__object_handle.html).

micvbang:
I've been wondering - how do you do performance testing of the `disk_interface`s? 
I first tried to set up a tracker locally and try to move data between two clients, but that doesn't seem to give me very consistent numbers. I've been trying out `connection_tester` as well, which is a lot easier to get started.

What would you recommend?

arvidn:
There's this python script that runs transfer benchmarks. This covers more than just the disk I/O, but I would expect that to be an important part of it. This script is not run regularly and may have experienced some bit rot. I would also expect you'd want to tweak certain aspects of it for your specific setup.

https://github.com/arvidn/libtorrent/blob/RC_2_0/tools/run_benchmark.py

There's also a test that benchmarks checking of files. this is much more disk I/O centric.

https://github.com/arvidn/libtorrent/blob/RC_2_0/tools/benchmark_checking.py

micvbang:
I'll try to play around with this - thank you!

micvbang:
Hey again!

I've been attempting to make the IOs "truly" async by not forcing IO reaping inside `async_read` and `async_write`. This has brought progress in both the right and wrong direction. The IOs are asynchronous now, which is awesome, but I've messed something up in my implementation that I've had a hard time to track down.

To get the async IOs started, I went with your suggestion of having a separate thread doing the polling. As xNVMe IO queues aren't thread safe, this has required me to add some locking. I think the locking currently is a bit too pessimistic, but an iteration or two should make it clear if that's the case. Overall I think this works well.

I've messed around with the implementation enough to get the `test_storage` tests passing. That helped me to find a few bugs, which was awesome. There's still a bunch of other tests that don't pass, so there's an obvious avenue for future work.

My current problem (or symptom of my problem) is that when I run `client_test` I can get it to actually download torrent data -- it progresses through the torrent (slowly), but there's a lot of bytes ending up in "fail" because piece hashing fails, I think. At least I'm often getting logs like these `ubuntu-20.04.2.0-desktop-amd64.iso hash for piece xxx failed`, which leads me to believe that I screwed up `async_hash` somehow. My implementation of `async_hash` is a bit complicated, so it's definitely very likely that there's something there I'm missing.

The `readv2` and `writev` implementations are so simple that I'm pretty confident that they aren't the issue. The only exception to this statement is that `readv2` and `writev` both utilize my `prepare_ios` function which mimics and replaces `readwritev`, and I'm not still not entirely certain if what I'm doing is safe. I think it is, though.

Ah, and I definitely see how messed up my implementation of `open_file_xnvme` is -- the implementation has gotten even more hacky since you saw it last, so a next step definitely is to look into the `file_view_pool`. I didn't look into this yet as I don't think this relates to the problem I'm currently focusing on.


My current focus is to try to get torrents downloading without seeing any data in the "fail" bucket. My theory is that the problem is with `async_hash`.
Would you have time to do a review of `async_hash` and `prepare_ios` to see if there's something obvious that I'm messing up? Or just a general review to spot the parts that I haven't realized myself are problematic

arvidn:
one thing that makes `async_hash` complicated is that it's expected to "synchronize" with any previous writes that have been posted. So, if writes are issued and started as async tasks, then an async_hash arrives, it's supposed to see all previous writes, even if they haven't completed yet.

In the `mmap_disk_io` implementation I use a store_buffer (see `include/libtorrent/aux_/store_buffer.hpp`) where all pending writes are stored while waiting for them to complete. Any read or hash operation always looks there first, before issuing reads from disk.

It's similar to store buffers in CPUs.

micvbang:
Ahhh, this makes a lot of sense! That definitely could explain why I'm seeing random fails.

Thanks for the hint - I'll look into this!

PR #6227, opened 2021-05-24
bump version to 1.2.14

arvidn:


FranciscoPombal:
@arvidn 

It would be nice if https://github.com/arvidn/libtorrent/pull/6193 could make it in time for 1.2.14. I just need a bit more of your help investigating the problems in its prerequisite PR: https://github.com/arvidn/libtorrent/pull/6195

FranciscoPombal:
@arvidn 

Seems you (perhaps accidentally) merged this before https://github.com/arvidn/libtorrent/pull/6193. Which commit will you tag as version 1.2.14.

arvidn:
I'm fine with any order. the commit hash in the header will be wrong either way..

PR #6231, opened 2021-05-28
Fix C compilation with CMake

paullouisageneau:
This PR fixes a regression introduced by https://github.com/arvidn/libtorrent/pull/6230 breaking the compilation of C dependencies with CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -479,54 +479,6 @@ feature_option(static_runtime "build libtorrent with static runtime" OFF)
 
 find_public_dependency(Threads REQUIRED)
 
-if(CMAKE_CXX_COMPILER_ID MATCHES Clang)
-	add_compile_options(
-		-Weverything
-		-Wno-documentation
-		-Wno-c++98-compat-pedantic
-		-Wno-c++11-compat-pedantic
-		-Wno-padded
-		-Wno-alloca
-		-Wno-global-constructors
-		-Wno-exit-time-destructors
-		-Wno-weak-vtables
-		-Wno-return-std-move-in-c++11
-		-Wno-unknown-warning-option
-	)
-elseif(CMAKE_CXX_COMPILER_ID MATCHES GNU)
-	add_compile_options(
-		-Wall
-		-Wextra
-		-Wpedantic
-		-Wparentheses
-		-Wvla
-		-Wno-c++11-compat
-		-Wno-format-zero-length
-		-Wno-noexcept-type
-		-ftemplate-depth=512
-	)
-elseif(MSVC)
-	add_compile_options(
-		# https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
-		/Zc:__cplusplus
-		/W4
-		# C4251: 'identifier' : class 'type' needs to have dll-interface to be
-		#        used by clients of class 'type2'
-		/wd4251
-		# C4268: 'identifier' : 'const' static/global data initialized
-		#        with compiler generated default constructor fills the object with zeros
-		/wd4268
-		# C4275: non DLL-interface classkey 'identifier' used as base for
-		#        DLL-interface classkey 'identifier'
-		/wd4275
-		# C4373: virtual function overrides, previous versions of the compiler
-		#        did not override when parameters only differed by const/volatile qualifiers
-		/wd4373
-		# C4503: 'identifier': decorated name length exceeded, name was truncated
-		/wd4503
-	)
-endif()
-
 if(static_runtime)
 	if (MSVC)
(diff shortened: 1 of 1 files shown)


arvidn:
I don't understand the distinction between those two ways of setting flags. I would have thought that the existing way would be preferred.

Is the issue that one propagates to dependencies and the other one doesn't?
It's also odd that these flags don't apply to a target, but presumable all targets.

@FranciscoPombal Do you have any thoughts?

arvidn:
is there a way to add *just* c++ options? or is setting variables the way you're supposed to do this? (I thought "modern cmake" had more reasonable ways)

paullouisageneau:
> Is the issue that one propagates to dependencies and the other one doesn't?
> It's also odd that these flags don't apply to a target, but presumable all targets.

Actually, both indeed propagates to dependencies. However, `CXX` flags are only for C++, whereas `add_compile_options` sets both C and C++.

paullouisageneau:
> is there a way to add _just_ c++ options? or is setting variables the way you're supposed to do this? (I thought "modern cmake" had more reasonable ways)

I'm not a huge CMake Guru but I don't think so. It seems you have to set `CMAKE_C_FLAGS` or `CMAKE_CXX_FLAGS` to set options respectively only for C or only for C++.

arvidn:
this looks good to me. I feel like this problem has come up before. There should probably be a comment in there explaining why we have to do it this way

FranciscoPombal:
There is a better way to fix this. It is to use target_compile_options for these options, and a different set for the things that need C. add_compile_options simply affectes all suddirectories recursively from the place it is set.

Messing with CMAKE_CXX_FLAGS within the script itself should never be done, unless as an absolutely last resort.

I can peobably cook something up tomorrow

arvidn:
and `target_compile_options()` propagates to dependencies too, right?

FranciscoPombal:
> and `target_compile_options()` propagates to dependencies too, right?

Yes, in a better way - i.e., controlled by the`PRIVATE`/`INTERFACE`/`PUBLIC` keywords, does not depend on the directory hierarchy. `add_compile_options` is the equivalent of adding all options as `PUBLIC` indiscriminately to everything at the level it is set and below.

paullouisageneau:
@FranciscoPombal You mean `target_compile_options()` propagates *from* dependencies, right? For instance, if you set the target options `PUBLIC` for a dependency, the options are also added when compiling any target importing it.

FranciscoPombal:
@paullouisageneau 

> @FranciscoPombal You mean `target_compile_options()` propagates _from_ dependencies, right? For instance, if you set the target options `PUBLIC` for a dependency, the options are also added when compiling any target importing it.

Here is an example to make things clear:

```cmake
add_library(Foo)
# ...
target_compile_options(Foo PUBLIC "-x")
# ...
```

All source files for `Foo` will be compiled with `-x` flag, **and** the `-x` flag will be propagated as a _usage requirement_ to anything that depends on it.

Now suppose there is a library `Bar` that depends on `Foo`:

```cmake
add_library(Bar)
# ...
target_compile_options(Foo PRIVATE "-y")
# ...
# Establishing the dependency on Foo, option A
target_link_libraries(Bar PRIVATE Foo)
# Establishing the dependency on Foo, option B
target_link_libraries(Bar PUBLIC Foo)
```

`Bar`'s source files will be compiled with both the `-x` (propagated from `Foo`) and `-y`. Neither `-x` nor `-y` will be propagated to targets that depend on `Bar` in the case of option A, while **only** `-x` will be propagated in the case of option B.

`INTERFACE` is not mentioned in this example, as its relationship to the other keywords is simple to understand - `PUBLIC = PRIVATE + INTERFACE`.

paullouisageneau:
Actually, I was under the impression @arvidn wanted the flags to propagate to submodules, but maybe I'm mistaken here. 

@arvidn is that the case? Would it be OK to set the flags only for the libtorrent target? If so, I can make the change to `target_link_libraries`.

arvidn:
as an example; if you build `libtorrent` with `-fshort-enums`, you better build everything `libtorrent` links against with `-fshort-enums` too, otherwise you'll violate ODR and have ABI incompatibility issues, typically resulting in subtle memory corruption.

Perhaps a better example is `debug-iterators=on`, which affects the ABI of the standard library types. Another example would be thread-sanitizer.

arvidn:
as long as it's just warning flags, I would think we're good. But selecting the C++ version *may* affect ABI.

arvidn:
as long as we stick to just enabling warnings, it might be desirable to not have the flags propagate to dependencies.

paullouisageneau:
@arvidn OK, I see. Since current flags control only warning policies, I'm changing to `target_link_libraries`.

FranciscoPombal (approved):

PR #6232, opened 2021-05-29
Peer list seeds

arvidn:


src/peer_list.cpp
```diff
@@ -79,7 +79,64 @@ namespace {
...
+
+		if (lhs->last_connected != rhs->last_connected)
+			return lhs->last_connected < rhs->last_connected;
+
+		if (lhs->maybe_seed != rhs->maybe_seed)
+		{
+			// if we're finished, de-prioritze peers we think may be seeds
+			// otherwise, prioritize peers we think may be seeds
+			if (finished) return rhs->maybe_seed;
+			else return lhs->maybe_seed;
```
arvidn, line 127: given that the spec says that this may also include upload-only peers, this inverse prioritization should probably not be made

PR #6250, opened 2021-06-07
allow to set the ca certificate when loading a magnet link

thierryba:
This means that when you load a magnet link from an SSL torrent,
you can now join the swarm by calling torrent_handle::set_ssl_ca_for_metadata
when you receive the add_torrent_alert alert.
That way you can have a fully SSL controlled environment and use magnet links.

Any comment is appreciated.

### DIFF
--- a/include/libtorrent/add_torrent_params.hpp
+++ b/include/libtorrent/add_torrent_params.hpp
@@ -365,6 +365,12 @@ TORRENT_VERSION_NAMESPACE_3
 		error_code internal_resume_data_error;
 #endif // TORRENT_ABI_VERSION
 
+		// the root ca certificate to use in case of a magnet link that refers
+        // to an SSL Torrent. It will allow the client to connect to the corresponding swarm
+        // and downloadd the metadata. If it is not the cirrect certificate, the torrent
+        // will stay in downloading metadata state because it cannot be part of the swarm.
+        // This corresponds to the 'ssl-cert' field found on an SSL torrent.
+		std::string ca_certificate;
 	};
 
 TORRENT_VERSION_NAMESPACE_3_END
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -536,6 +536,14 @@ bool is_downloading_state(int const st)
 					? m_peer_list->num_connect_candidates() : -1);
 			}
 #endif
+       
+			if (!p.ca_certificate.empty()) {
+				m_ssl_torrent = true;
+#ifdef TORRENT_SSL_PEERS
+				init_ssl(p.ca_certificate);
+#endif
+			}
+
 		}
 
 #ifndef TORRENT_DISABLE_LOGGING
--- a/test/test_ssl.cpp
+++ b/test/test_ssl.cpp
@@ -16,8 +16,10 @@ see LICENSE file.
 #include "libtorrent/session_status.hpp"
 #include "libtorrent/torrent_info.hpp"
 #include "libtorrent/hex.hpp" // for to_hex
+#include "libtorrent/magnet_uri.hpp" // for parse_magnet_uri
 #include "libtorrent/time.hpp"
 #include "libtorrent/aux_/ssl.hpp"
+#include "libtorrent/aux_/torrent.hpp"
 
 #include "test.hpp"
 #include "test_utils.hpp"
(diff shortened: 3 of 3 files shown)


arvidn:
I don't think this would work. The certificate authority need to be tied to the torrent from the beginning. Either passed in via the link itself, or at least via `add_torrent_params`.

With your patch, the torrent will start out as *open*, anyone can connect. And then close down to require authentication. That window where it's open is a security issue.

I think it would make more sense to pass this in via `add_torrent_params`.

Also keep in mind that "normal" SSL torrents also authenticate with the tracker, so the peers it tries to connect to are also signed off on by the authority. I believe DHT is disabled for SSL torrents.

Before a feature like this could land you would need to extend `test/test_ssl.cpp` to cover the new feature also.

thierryba:
@arvidn Thank you for the quick reply. I'll try to propose a patch that does set the ca certificate earlier.
About the tests, I was first waiting to get that early feedback but I do understand that quality is very important.
Stay tuned.

include/libtorrent/magnet_uri.hpp
```diff
@@ -83,6 +83,7 @@ namespace libtorrent {
 	TORRENT_EXPORT add_torrent_params parse_magnet_uri(string_view uri, error_code& ec);
 	TORRENT_EXPORT add_torrent_params parse_magnet_uri(string_view uri);
 	TORRENT_EXPORT void parse_magnet_uri(string_view uri, add_torrent_params& p, error_code& ec);
+	TORRENT_EXPORT void parse_magnet_uri(string_view uri, add_torrent_params& p, error_code& ec, string_view ca_certificate);
```
arvidn, line 86: this looks like it has limited value. You could just set the member on the `add_torrent_params` later anyway

include/libtorrent/add_torrent_params.hpp
```diff
@@ -388,6 +388,8 @@ TORRENT_VERSION_NAMESPACE_3
 		error_code internal_resume_data_error;
 #endif // TORRENT_ABI_VERSION
 
+        // the ca certificate to use in case of a magnet link
```
arvidn, line 391: I forgot to mention that new fields to `add_torrent_params` is an ABI break, and can't be done in 2.0. It has to go into the `master` branch.

Also, this documentation need to be more thorough. Maybe with a reference to the field that may exist in a .torrent file.

include/libtorrent/torrent_handle.hpp
```diff
@@ -842,6 +842,11 @@ namespace aux {
 		void set_ssl_certificate_buffer(std::string const& certificate
 			, std::string const& private_key
 			, std::string const& dh_params);
+        
+        
+        // For SSL torrents added via magnet-link, use this to specify the ca certificate to connect to the swarm
+        // ``ca_certificate`` is the pem encoded CA certificate used in the swarm
+        void set_ssl_ca_for_metadata(std::string const& ca_certificate);
```
arvidn, line 849: I think this should be removed now.

src/torrent.cpp
```diff
@@ -564,6 +564,15 @@ bool is_downloading_state(int const st)
 					? m_peer_list->num_connect_candidates() : -1);
 			}
 #endif
+            
+            if (!p.ca_certificate.empty()) {
+                m_ssl_torrent = true;
+    #ifdef TORRENT_SSL_PEERS
+                init_ssl(p.ca_certificate);
```
arvidn, line 571: I think it would be better to use the same path as when the cert is part of the torrent file itself. This looks shady.

For example, what happens if you load an SSL torrent (with an embedded CA cert) and also set this `add_torrent_params` field?

src/torrent.cpp
```diff
@@ -5878,6 +5887,13 @@ namespace {
 		}
 	}
 
+    void torrent::set_ssl_ca_cert_buffer(std::string const& certificate)
```
arvidn, line 5890: again, I don't see how it can be safe to *change* this on a torrent.

arvidn (changes requested):

src/torrent.cpp
thierryba: Sorry my code was not finalised yet...

include/libtorrent/torrent_handle.hpp
thierryba: indeed

include/libtorrent/add_torrent_params.hpp
thierryba: Done.

src/torrent.cpp
thierryba: that is a tricky one. I'll see what I can do.
BTW thanks a lot for the feedback.

PR #6268, opened 2021-06-14
Refactor benchmark tools

micvbang:
Using the benchmark scripts you mentioned, I found a few things that I personally see as improvements.

Most of the diff of this PR is a refactoring away from executing everything in the global scope, but does include some minor changes to usage, e.g. taking `toolset` as an input flag instead of using the first argument of argv, and checking the event log in benchmark_checking for a different event than the existing one, which did not appear in all tests I ran.

Let me know if I there's anything I should change, split the changes into smaller commits, or if the global scope execution is there on purpose.

### DIFF
--- a/tools/benchmark_checking.py
+++ b/tools/benchmark_checking.py
@@ -1,70 +1,64 @@
 #!/usr/bin/env python3
 # vim: tabstop=8 expandtab shiftwidth=4 softtabstop=4
 
+import argparse
 import os
 import shutil
 import subprocess
 import sys
 
-toolset = ''
-if len(sys.argv) > 1:
-    toolset = sys.argv[1]
 
-ret = os.system('cd ../examples && b2 release %s stage_client_test stage_connection_tester'
-                % toolset)
-if ret != 0:
-    print('ERROR: build failed: %d' % ret)
-    sys.exit(1)
+def main():
+    args = parse_args()
 
-try:
-    os.remove('.ses_state')
-except Exception:
-    pass
-try:
-    shutil.rmtree('.resume')
-except Exception:
-    pass
-
-if not os.path.exists('checking_benchmark.torrent'):
-    ret = os.system('../examples/connection_tester gen-torrent -s 10000 -n 15 -t checking_benchmark.torrent')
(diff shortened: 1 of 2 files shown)


arvidn (approved):
looks great!

tools/benchmark_checking.py
```diff
@@ -1,70 +1,64 @@
...
-                 '-1 %s -s . -f %s/events.log --alert_mask=all' \
-                 % (client_arg, output_dir)
+    client_cmd = ('../examples/client_test checking_benchmark.torrent '
+                  '--enable_dht=0 --enable_lsd=0 --enable_upnp=0 --enable_natpmp=0 '
+                  '-1 %s -s . -f %s/events.log --alert_mask=all'
+                  ) % (client_arg, output_dir)
 
     client_out = open('%s/client.out' % output_dir, 'w+')
-    print(client_cmd)
+    print('client_cmd: "{cmd}"'.format(cmd=client_cmd))
```
arvidn, line 61: I think this could even be simplified to:
```
print(f'client_cmd: "{client_cmd}"')
```

FranciscoPombal: You can also further simply it to:

```python3
print(f"{client_cmd=}")
```

With python >= 3.8. Optionally, you can have spaces around the `=`, which will also appear in the output.

Reference: https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging

micvbang: Yeah, good point @arvidn :) The primary reason I did not use literal string interpolation is that I didn't want to impose a python version (python 3.5 future or 3.6).

I did not know about the syntax you just posted at all, @FranciscoPombal, so thanks for that :)

arvidn: I'll merge it and fix it up to pass CI in master then

arvidn:
@micvbang did you see my PR against your PR? https://github.com/micvbang/libtorrent/pull/1

arvidn:
in my tests, locally, there still seem to be an issue in running `gprof2dot`. did you get that to work?

arvidn (approved):

PR #6275, opened 2021-06-18
Clarify docs on streaming strategies

AllSeeingEyeTolledEweSew:
Per https://github.com/arvidn/libtorrent/discussions/6272


arvidn (approved):
thanks!

PR #6278, opened 2021-06-18
defend against microsoft's max macro

arvidn:


include/libtorrent/file_storage.hpp
```diff
@@ -235,7 +235,7 @@ namespace aux {
 		// tree.
 		static constexpr std::int64_t max_file_size = std::min(
```
3unjee, line 236: We need to defend this one too: `static constexpr std::int64_t max_file_size = (std::min)(`

3unjee: After that, I can confirm it works in a x86_64 context on MSVC (haven't tested x86 but shouldn't be an issue either).

PR #6290, opened 2021-06-22
V2 hash picker fixes

arvidn:


glassez:
It works better with the latest changes. But I'm away now, so I can't test it in more detail.

Issue #4408, opened 2020-03-09
Clarification about Sequential mode


elgatito:
I'm trying to understand how sequential piece picker works in libtorrent.

For example consider this flow:
1) Adding torrent, selecting file to download.
2) User reads file from the beginning immediatelly. I see the piece is not available yet, so I set highest priority to the first piece. Pieces are downloaded from the 1st one sequentially now?
3) Now user wants to seek the file to the middle. I see the piece is not available and set the highest priority to that piece. How should piece picker work now?

What I'm trying to make:
Reads from user is coming through my code, so I know what it is reading now. So I want to make sure that next coming pieces are requested next and not randomly.

Currently I manually update pieces priority each second, so set kind of priority chain, like 7-6-6-5-5-5-5.... to pieces, and when user seeks I update priorities to new starting place.

But maybe there are much better options. Because on not-so-fast devices and torrents with big number of pieces I see it takes more time to update priorities (which is logical of course).

Thanks

ssiloti:
Sequential mode always downloads sequentially from the beginning of the torrent. For what you describe `torrent_handle::set_piece_deadline` should work better than changing piece priorities. Setting a piece deadline of zero will ensure that piece is downloaded as quickly as possible.

elgatito:
@ssiloti consider this pseudo-code:

`
startPiece = 10
endPiece = 60
currentReadPiece = 30

for (index in (currentReadPiece, currentReadPiece+10)) {
    t.set_piece_deadline(index, (index - currentReadPiece) * 50)
}
`

So  I need to download 50 pieces, but user currently reads piece number 30, so I need to prioritize 10 pieces next it, but make sure they are downloaded kind of sequentially. `currentReadPiece + 1` with the highest priority, so calling `set_piece_deadline` with `0` and the next one with `50` and so on.

Should I expect pieces to be downloaded in that order?

elgatito:
@ssiloti @arvidn 

I have found this test: https://github.com/arvidn/libtorrent/blob/bdee0b563cf0c6f6bbb265752b9d80768528fd12/test/test_piece_picker.cpp#L782

It looks for me that if I set all priorities to 1, and the ones I need urgently with 2 (or any bigger than 1), then picker will request pieces in my desired order?

arvidn:
[...]

PR #6293, opened 2021-06-23
Elaborate on streaming API and strategy

AllSeeingEyeTolledEweSew:
I've been reading a lot trying to get a handle on torrent streaming, how libtorrent supports it and how its support should be used.

I updated `streaming.rst` with my findings, both to validate them and to help anyone else in this space (it seems there are a couple people working actively on streaming right now).

Please let me know if anything's inaccurate or incomplete.

In particular, I'm boldly assuming the best growth-friendly API usage is that users should call `set_piece_deadline()` for a whole file, and use other mechanisms to dynamically constrain the time-critical set. I lean this way because I can think of some other such mechanisms to try, but I doubt I've thought of everything. Please let me know if you have different ideas.

### DIFF
--- a/docs/streaming.rst
+++ b/docs/streaming.rst
@@ -6,6 +6,52 @@ Streaming implementation
 This documents describes the algorithm libtorrent uses to satisfy time critical
 piece requests, i.e. streaming.
 
+considerations
+--------------
+
+Streaming data over bittorrent presents several new issues.
+
+1. Sequential downloading hurts download capacity
+	Bittorrent encourages downloaders to do tit-for-tat data transfers between each
+	other. This only works if there is mutual *interest* (each has data the other
+	wants). But if two peers are streaming the torrent in *strictly* sequential
+	piece order, then the "younger" peer has no data to offer the "older" peer.
+	These peers would depend on the upload capacity of any seeds in the swarm.
+2. Full request queues hurt interactivity
+	Whenever the set of time-critical pieces changes (the stream "seeks" to a
+	new position), the optimal move to minimize delay would be to request the
+	newly-critical data from a fast peer *with an empty request queue*.
+3. Non-full request queues hurt throughput
+	Converse to the above, if we want to maximize throughput, we should keep
+	peer request queues as full as possible. In addition to maximizing transfer
+	with a single peer, it's also good to exercise transfer with *all* peers, to discover
+	the *fastest ones*.
+
+Dealing with these issues is an open question.
+
+critical window
+---------------
+
+A common approach to streaming is to use a *critical window* of data ahead of
+the stream position. There are many ways to define such a window.
+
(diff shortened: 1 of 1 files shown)


docs/streaming.rst
```diff
@@ -6,6 +6,52 @@ Streaming implementation
...
+the "rate" of the stream. For media files, the rate could be the bit rate of the
+media. Otherwise it could be dynamically measured by how fast the consumer
+consumes data.
+
+If the stream consumer consumes data slower than the torrent's download rate
+(this should usually be the case when streaming media), we can decide what to
+do with the "leftover" bandwidth once the critical window is fulfilled. One idea
+is to use classic rarest-first picking to download the rest of the torrent, as
+this can mitigate the problems of sequential downloading, and keeps request
+queues full.
```
arvidn, line 53: what's described as a "one idea is to" here is in fact what libtorrent does. I'm worried that phrasing it like this may suggest to the reader that this is not necessarily what libtorrent does.

arvidn: I suppose this is clarified further down

AllSeeingEyeTolledEweSew: I'm using a first-principles approach to documenting streaming logic and functions, because it seems to be a cutting-edge area. I think documenting our basis of understanding is crucial. In contrast, we don't need to document *why* libtorrent uses rarest-first picking, because there's plenty of information available on the topic.

I believe the lack of clarity has left users confused, and having wrong expectations of libtorrent.

Recent user discussions end up disregarding time-critical logic in favor of `sequential_download` ([[1]](https://github.com/arvidn/libtorrent/issues/4408), [[2]](https://github.com/arvidn/libtorrent/pull/5996)). Since time-critical logic is intended to be a better version of `sequential_download`, I conclude either these users don't understand it, or couldn't figure out how to make `set_piece_deadline` work in a context that lacks well-defined deadlines.

Code examples I can find in the wild which use `set_piece_deadline` ([[1]](https://github.com/animeshkundu/pyflix/blob/master/torrent/strategy.py#L28), [[2]](https://github.com/AlexCovizzi/torrenttv/blob/master/torrenttv/bittorrent/file.py#L75-L93), [[3]](https://github.com/FLYKIS2/pelisalacarta-Cmos/blob/master/core/mct.py#L184-L187)) seem to use it wrong. Judging by that code, I *suspect* there's a prevailing belief that the intent of `set_piece_deadline()` is to download a piece *within the next N milliseconds*. Said another way, I think a lot of users are confused between *deadline* and *timeout*.

I think it would be really helpful to link to a project that uses `set_piece_deadline` in the intended way, where deadlines are well-defined, e.g. in a video player. I couldn't find any though.

AllSeeingEyeTolledEweSew:
Unstale. I'd still like to figure this out.

PR #6296, opened 2021-06-25
added missed sources to Python bindings CMakeLists.txt

Kolcha:
fixes libtorrent module loading issues:
```
libtorrent.cpython-39-x86_64-linux-gnu.so: undefined symbol: _Z16bind_sha256_hashv
```
found this by myself thanks to own Debian packages nightly builds (cmake is used for building) and a lot of scripts using python bindings. no reported issues were found.

master branch is not tested, so that's why this PR to RC_2_0 branch

### DIFF
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -49,6 +49,7 @@ Python3_add_library(python-libtorrent MODULE WITH_SOABI
 	src/entry.cpp
 	src/error_code.cpp
 	src/fingerprint.cpp
+	src/info_hash.cpp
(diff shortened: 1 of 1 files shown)


arvidn:
good catch! ideally there would be a python tests that uses these functions to catch it in CI.

arvidn:
oh, I don't think there's any CI coverage for cmake builds of the python binding..

Kolcha:
no specific tests are required to detect mentioned issue, it happens even on `import libtorrent` statement. so, any existing tests must catch such issues. only one condition must be met - python bindings were built using cmake

PR #6310, opened 2021-07-08
use can_disconnect in bt_peer_connection upload-to-upload connection

arvidn:


arvidn:
@joriscarrier would you mind testing this patch to make sure it works for you?

joriscarrier:
> @joriscarrier would you mind testing this patch to make sure it works for you?

i will do that

joriscarrier:
test in progress (i think is missing an other can_disconnect: #6318)

joriscarrier:
@arvidn for me it is good, I do not see any other case.

Issue #6302, opened 2021-07-02
Seed rank oscillates when we are the only seed causing queued seeds to repeatedly pause/resume


thrnz:
This is regarding libtorrent 1.2.14, although 2.0 appears to use the same code.

When seeding a number of torrents with 0 other seeds, torrents appear to regularly oscillate between seeding and queued.

If I understand [this code](https://github.com/arvidn/libtorrent/blob/33a10d5a723a6c27d9baee7d9bf2028eb81c88ed/src/torrent.cpp#L8575-L8636) right, a torrent with 0 seeds gets a boost to it's seed ranking causing it to generally get priority over other torrents. However, once seeding begins, the torrent appears to no longer be considered 'seedless' as the tracker will now report a single seed (ie us), and will no longer get a boost to its seed rank when its next recalculated (I assume after `auto_manage_interval`), causing it to become queued again. It will then be considered seedless again after the next `auto_manage_interval` and so on.

Ideally libtorrent might consider a torrent to have 0 seeds for the purpose of calculating the seed rank if the tracker reports a single seed and it is currently seeding (ie. we are that single seed).

C++ isn't my thing, so forgive the psuedocode, but this
https://github.com/arvidn/libtorrent/blob/33a10d5a723a6c27d9baee7d9bf2028eb81c88ed/src/torrent.cpp#L8625
Might be changed to something like:
```if (seeds == 0) or (seeds == 1 && currently_seeding)```

A semi-workaround I've been using for the time being is to set `stop_tracker_timeout` to 0, so the tracker still reports a single seed when scraped even once we've stopped. So after a while, all 'seedless' torrents will be considered to have a single seed and will no longer oscillate until the tracker no longer considers us a seed.

wrtorrent:
try set 'active_seeds' to -1 (assume you not exceed 500 of `active_limit`)

thrnz:
Uncapping the number of active seeds kind of defeats the purpose of queueing and [seeding based on demand](https://www.libtorrent.org/manual-ref.html#seeding-queue).

I rebuilt libtorrent having changed 
https://github.com/arvidn/libtorrent/blob/33a10d5a723a6c27d9baee7d9bf2028eb81c88ed/src/torrent.cpp#L8625

to:

```if (seeds == 0 || (seeds == 1 && !is_paused())```

and it no longer flip-flops between queued and seeding when it's the only seed.

wrtorrent:
not work like utorrent
[...]

PR #6311, opened 2021-07-08
Fix seed rank calculation when we're the only known seed

thrnz:
When seeding, torrents with no other known seeds receive a boost to their seed rank so as to take priority. Currently this boost is only applied when there are no known seeds (ie `seeds == 0`). However, once we begin seeding, `seed_rank()` now sees that `seeds > 0` and no longer considers the torrent seedless, even though that single seed is itself. This causes queued seeds to repeatedly alternate between seeding/queued whenever the seed rank is recalculated.

See #6302

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -9181,7 +9181,9 @@ namespace {
 		if (m_incomplete != 0xffffff) downloaders = m_incomplete;
 		else downloaders = m_peer_list ? m_peer_list->num_peers() - m_peer_list->num_seeds() : 0;
 
-		if (seeds == 0)
+		// If there is a single known seed and we're currently seeding then it's
+		// likely us, so consider the torrent otherwise seedless
+		if (seeds == 0 || (seeds == 1 && !is_paused()))
 		{
 			ret |= no_seeds;
 			ret |= downloaders & prio_mask;
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -9177,9 +9177,9 @@ namespace {
 
 		// If we're currently seeding and using tracker supplied scrape
 		// data, we should remove ourselves from the seed count
-		int const self_seed = is_seed() && !is_paused() ? 1 : 0;
+		std::uint32_t const self_seed = is_seed() && !is_paused() ? std::uint32_t(1) : std::uint32_t(0);
 
-		if (m_complete != 0xffffff) seeds = std::max(0, m_complete - self_seed);
+		if (m_complete != 0xffffff) seeds = std::max(std::uint32_t(0), m_complete - self_seed);
 		else seeds = m_peer_list ? m_peer_list->num_seeds() : 0;
 
 		if (m_incomplete != 0xffffff) downloaders = m_incomplete;
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -9177,9 +9177,9 @@ namespace {
 
 		// If we're currently seeding and using tracker supplied scrape
 		// data, we should remove ourselves from the seed count
-		std::uint32_t const self_seed = is_seed() && !is_paused() ? std::uint32_t(1) : std::uint32_t(0);
+		int const self_seed = is_seed() && !is_paused() ? 1 : 0;
 
-		if (m_complete != 0xffffff) seeds = std::max(std::uint32_t(0), m_complete - self_seed);
+		if (m_complete != 0xffffff) seeds = std::max(0, int(m_complete) - self_seed);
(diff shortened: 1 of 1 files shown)


src/torrent.cpp
```diff
@@ -9181,7 +9181,9 @@ namespace {
 		if (m_incomplete != 0xffffff) downloaders = m_incomplete;
 		else downloaders = m_peer_list ? m_peer_list->num_peers() - m_peer_list->num_seeds() : 0;
 
-		if (seeds == 0)
+		// If there is a single known seed and we're currently seeding then it's
+		// likely us, so consider the torrent otherwise seedless
+		if (seeds == 0 || (seeds == 1 && !is_paused()))
```
arvidn, line 9186: there is one issue here. We can actually get here without being a seed, in which case `!is_paused()` cannot mean that we are the one seed.

on line 9146 we exit this function with 0 in case we aren't *finished* (i.e. we have downloaded all the files we want to download, but we may have de-selected some).

then we check `is_seed()`. I think this should probably take `is_seed()` into account as well.

It might be simpler to create a new variable, something like this:

```
int const self_seed = is_seed() && !is_paused() ? 1 : 0;
```

We can then compute `seeds` by subtracting this. Specifically, we really risk seeing ourself in the scrape data from the tracker, that's `m_complete`. If we happen to compute `seeds` by looking at our own peer list, we won't have ourself in there. So, rather than putting this logic here, I think it would be more accurate to change line 9178 to be:

```
if (m_complete != 0xffffff) seeds = std::min(0, m_complete - self_seed);
```

And move the comment there too.

Would that work?

thrnz: That looks good, though should that `min` be a `max`? Otherwise we'd never end up with more than 0 seeds.

arvidn: heh, yes :)

arvidn (approved):

src/torrent.cpp
```diff
@@ -9175,7 +9175,11 @@ namespace {
 		int seeds = 0;
 		int downloaders = 0;
 
-		if (m_complete != 0xffffff) seeds = m_complete;
+		// If we're currently seeding and using tracker supplied scrape
+		// data, we should remove ourselves from the seed count
+		int const self_seed = is_seed() && !is_paused() ? 1 : 0;
```
arvidn, line 9180: I think this should be `std::uint32_t` rather than `int`. to be compatible with `m_complete`.

https://github.com/arvidn/libtorrent/pull/6311/checks?check_run_id=3032982089#step:6:202

```
..\src\torrent.cpp(9182): error C2672: 'std::max': no matching overloaded function found
203
..\src\torrent.cpp(9182): error C2780: '_Ty std::max(std::initializer_list<_Elem>)': expects 1 arguments - 2 provided
204
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30037\include\utility(52): note: see declaration of 'std::max'
205
..\src\torrent.cpp(9182): error C2784: '_Ty std::max(std::initializer_list<_Elem>,_Pr)': could not deduce template argument for 'std::initializer_list<_Elem>' from 'int'
206
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30037\include\utility(49): note: see declaration of 'std::max'
207
..\src\torrent.cpp(9182): error C2782: 'const _Ty &std::max(const _Ty &,const _Ty &) noexcept(<expr>)': template parameter '_Ty' is ambiguous
208
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30037\include\utility(41): note: see declaration of 'std::max'
209
..\src\torrent.cpp(9182): note: could be 'uint32_t'
210
..\src\torrent.cpp(9182): note: or       'int'
```

arvidn: oh, now the issue is that `0` is `int`. You might have to say `std::uint32_t(0)`

thrnz: Third time lucky...

thrnz: Or maybe even fourth

thrnz: I went the other way and converted `m_complete` into an `int` instead. The other relevant vars (`downloaders`, `scale`, `seeds`) are also `int`, and it seems easier than trying to get my head around implicit conversion and signedness.

Looks like I have some reading to do.

arvidn (approved):

arvidn:
thanks!

PR #6314, opened 2021-07-09
Add `torrent_status::total` field to python bindings

### LINKED ISSUE #6313: `torrent_status::total` is missing in the python bindings
libtorrent version (or branch): RC_2_0 branch

The `total` field of `torrent_status` defined [here](https://libtorrent.org/reference-Torrent_Status.html#torrent_status) is missing in the python bindings.

aykxt:
fixes #6313


arvidn:
this field should probably be accessed by some test as well, to ensure it actually works

arvidn:
I believe this is a problem in `RC_1_2` as well.

PR #6316, opened 2021-07-12
fix issue when asking for file progress

arvidn:
right when starting a force-recheck


fergus-dall:
Can this be backported to the 1.2 branch? I ran into this crash today using deluge on archlinux, which is still using libtorrent 1.2

arvidn:
sure, do you want to give it a try? I would expect it to apply cleanly to `RC_1_2`. Otherwise I can do it.

PR #6326, opened 2021-07-21
allow loading v2 .torrent files without piece layers

arvidn:


src/torrent_info.cpp
```diff
@@ -1576,11 +1576,19 @@ namespace {
...
-			// mark the torrent as invalid
-			m_files.set_piece_length(0);
-			return false;
+			// allow torrent files without piece layers, just like we allow magnet
+			// links. However, if there are piece layers, make sure they're
+			// valid
+			bdecode_node const& e = torrent_file.dict_find_dict("piece layers");
+			if (e && !parse_piece_layers(e, ec))
+			{
+				TORRENT_ASSERT(ec);
```
glassez, line 1587: It seems that you are asserting on something coming from outside (a valid torrent file). This doesn't look like a good idea. This should be treated as a regular, although not valid, runtime case.

glassez (approved):
LGTM

arvidn:
@glassez does that last commit make sense too?

PR #6327, opened 2021-07-21
add write_torrent_file()

arvidn:
a function to save a .torrent file from an add_torrent_params

### DIFF
--- a/src/write_resume_data.cpp
+++ b/src/write_resume_data.cpp
@@ -42,8 +42,33 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/aux_/numeric_cast.hpp"
 #include "libtorrent/torrent.hpp" // for default_piece_priority
 #include "libtorrent/aux_/numeric_cast.hpp" // for clamp
+#include "libtorrent/aux_/merkle.hpp" // for merkle_
 
 namespace libtorrent {
+namespace {
+	entry build_tracker_list(std::vector<std::string> const& trackers
+		, std::vector<int> const& tracker_tiers)
+	{
+		entry ret;
+		entry::list_type& tr_list = ret.list();
+		if (trackers.empty()) return ret;
+
+		tr_list.emplace_back(entry::list_type());
+		std::size_t tier = 0;
+		auto tier_it = tracker_tiers.begin();
+		for (std::string const& tr : trackers)
+		{
+			if (tier_it != tracker_tiers.end())
+				tier = aux::clamp(std::size_t(*tier_it++), std::size_t{0}, std::size_t{1024});
+
+			if (tr_list.size() <= tier)
+				tr_list.resize(tier + 1);
+
+			tr_list[tier].list().emplace_back(tr);
+		}
+		return ret;
+	}
+}
 
 	entry write_resume_data(add_torrent_params const& atp)
 	{
@@ -176,25 +201,11 @@ namespace libtorrent {
 		}
 
 		// save trackers
-		entry::list_type& tr_list = ret["trackers"].list();
-		if (!atp.trackers.empty())
-		{
-			tr_list.emplace_back(entry::list_type());
-			std::size_t tier = 0;
-			auto tier_it = atp.tracker_tiers.begin();
-			for (std::string const& tr : atp.trackers)
-			{
-				if (tier_it != atp.tracker_tiers.end())
-					tier = aux::clamp(std::size_t(*tier_it++), std::size_t{0}, std::size_t{1024});
-
-				if (tr_list.size() <= tier)
-					tr_list.resize(tier + 1);
-
-				tr_list[tier].list().emplace_back(tr);
-			}
-		}
+		ret["trackers"] = build_tracker_list(atp.trackers, atp.tracker_tiers);
 
 		// save web seeds
+		// if we removed the web seeds, make sure to record that in the resume
+		// data
 		entry::list_type& url_list = ret["url-list"].list();
 		std::copy(atp.url_seeds.begin(), atp.url_seeds.end(), std::back_inserter(url_list));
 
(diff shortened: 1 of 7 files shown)


src/write_resume_data.cpp
```diff
@@ -154,9 +109,161 @@ namespace libtorrent {
...
+						{
+							layer += h.to_string();
+						}
+					}
+				}
+			}
+
+			if (!piece_layers.empty())
+			{
+				ret["piece layers"] = piece_layers;
```
glassez, line 143: Can you confirm that if `piece_layers` is not empty at this point, then it contains complete/valid "piece layers" data?

src/write_resume_data.cpp
```diff
@@ -42,66 +42,20 @@ POSSIBILITY OF SUCH DAMAGE.
...
 
-		// cast to seconds in case that internal values doesn't have ratio<1>
-		ret["active_time"] = atp.active_time;
-		ret["finished_time"] = atp.finished_time;
-		ret["seeding_time"] = atp.seeding_time;
-		ret["last_seen_complete"] = atp.last_seen_complete;
-		ret["last_download"] = atp.last_download;
-		ret["last_upload"] = atp.last_upload;
+	// only writes fields to create a .torrent file, no other resume data
+	constexpr write_resume_flags write_torrent_only = 0_bit;
```
glassez, line 53: IMO, `_only` suffix is redundant/meaningless. It looks completely out of place in an expression like `write_torrent_only | write_resume_only`.

arvidn: perhaps these should be inverted, to mean *do* write torrent data, and *do* write resume data, and remove the `_only`

src/write_resume_data.cpp
arvidn: it may not be complete. There may be files missing piece layers, but the piece layers that are there are correct.

I suppose that's a problem since the loading of a torrent requires all files to be there if it's present

glassez: I mean you should write "piece layers" field only in the case when it is fully valid (i.e. it produces valid torrent file so it can be correctly loaded later).

arvidn: I was thinking of relaxing the loading function instead, to accept some files not having the piece layer

arvidn: see most recent commit to https://github.com/arvidn/libtorrent/pull/6326

src/write_resume_data.cpp
glassez: Otherwise they shouldn't be flags:
1. Write everything
2. Write torrent only
3. Write resume data only

src/write_resume_data.cpp
```diff
@@ -154,9 +109,161 @@ namespace libtorrent {
...
+				}
+			}
+
+			if (!(flags & write_resume_only) && atp.trackers.size() > 1)
+				ret["announce-list"] = announce_list;
+		}
+
+		// .torrent file fields above
+
+		if (flags & write_torrent_only) return ret;
```
glassez, line 195: Nitpick
I would implement it in a reversed way:
1. Function that (really) produces torrent data
2. Function that (really) produces resume data
3. Function that produces merged (or one of, depending of parameter) data above.

src/write_resume_data.cpp
glassez: >I was thinking of relaxing the loading function instead, to accept some files not having the piece layer

You didn't understand me.
You should either produce fully valid "piece layers" field or produce no such field either.
IIRC, "loading function" still fails if there is invalid "piece layers" that is correct behavior.

arvidn: the benefit of allowing incomplete piece layers when loading torrents is that some hashes are better than none, and not producing a piece layer here means that if any file is "deselected", it won't ever save any hashes when creating a .torrent file.

I think it's reasonable to accept that some files don't have hashes when loading a torrent, I don't see a strong reason to implement "all or nothing" behavior.

glassez: 1. There is a "valid torrent file" requirements in BitTorrent v2 protocol specifications so the function that is intended to create real torrent files should follow it (i.e. create v2/hybrid torrent only with fully valid "piece layers") so that any compliant client can load it.
2. There is nothing wrong to allow load torrent files w/o "piece layers" since libtorrent can process it anyway.
3. We need to produce "incomplete" torrent files only for internal use (to split resume data). Since resume data stores merkle tree data anyway we don't need to store "piece layers" separately, isn't it? So if you provide separate function to produce such "internal purpose" torrent files you shouldn't either care to store "piece layers" there. I wouldn't even call it `write_torrent_file` to eliminate confusion.

arvidn: > There is a "valid torrent file" requirements in BitTorrent v2 protocol specifications so the function that is intended to create real torrent files should follow it (i.e. create v2/hybrid torrent only with fully valid "piece layers") so that any compliant client can load it.

That's reasonable. It's a larger issue than producing incomplete piece layers though. Omitting the piece layer would be equally invalid. This function would need a failure path, either through exception or an `error_code`

glassez: So how about par.3 above?

src/write_resume_data.cpp
```diff
@@ -285,6 +296,88 @@ namespace libtorrent {
...
+			std::copy(atp.url_seeds.begin(), atp.url_seeds.end(), std::back_inserter(url_list));
+		}
+
+		if (!atp.http_seeds.empty())
+		{
+			entry::list_type& httpseeds_list = ret["httpseeds"].list();
+			std::copy(atp.http_seeds.begin(), atp.http_seeds.end(), std::back_inserter(httpseeds_list));
+		}
+
+		if (!atp.name.empty()) ret["name"] = atp.name;
```
glassez, line 370: This looks suspicious. Is this field used in a torrent file?

arvidn: hm. no, probably not. "name" is pulled out of the info dictionary

src/write_resume_data.cpp
arvidn: right, if you were to just include the info-dict in the resume data, you wouldn't need to create .torrent files at all. (i.e. specify that flag to `save_resume_data()`)

glassez: >if you were to just include the info-dict in the resume data, you wouldn't need to create .torrent files at all.

We found it irrational to overwrite immutable information (info-dict) every time we save resume data.

Personally, I recognize this problem even deeper.
We are dealing with two classes of information about the torrent that we want to store:
1. Immutable or rarely changed (usually changed by event/action) information, e.g. info-dict, save path, file names, etc.
2. Frequently/continuously changed information (usually regardless of events or actions on the user side) information, e.g. data of downloaded pieces, time counters (active, downloading, seeding, etc.).

It is obvious (at least to me) that:
1. It would be more convenient to store these two classes of information separately;
2. It would be more convenient to receive information of different classes in different ways:
2.1 Each item of information of the first class should be available both on demand and in a notifying manner (i.e., when an event occurs, as a result of which the information changes, libtorrent should send a notification),
2.2 Information of the second class should be available on demand, so that the user side can determine the desired frequency of updates.

PR #6342, opened 2021-07-26
Improve superseeding performance by scaling the number of superseeded pieces to a peer's upload speed

thrnz:
To-do:

- [ ] Possibly improve handling of peers that filter redundant `HAVE` messges
- [x] Simulations
- [ ] Unit tests

---

In order to improve transfer speed a peer can queue up a number of blocks for download. However when superseeding, peers will only be able to queue data from the one or two pieces we've said that we have, in some cases limiting the achievable upload speed. This is more of an issue with a faster upload capacity, higher latency, and for torrents with a smaller piece size.

For instance, for a torrent with an 8MB piece size, a peer can potentially request up to 16MB (depending on `max_allowed_in_request_queue` and their clients behavior) of data, as it believes that we only have those 2 pieces. In the case of a 256KB piece size, the peer could only request up to a max of 512KB though. Note these are maximums, and deplete as the pieces download until we tell the peer we `HAVE` another piece.

In order to have upload speeds to peers on superseeded torrents up to par with normally seeded torrents, we would need to increase the number of pieces we make available to each peer from the current hard limit of 2.

For testing I reworked the superseed behavior to make use of a new option `max_superseed_bytes`. The number of pieces to offer a peer when superseeding would then be this divided by the torrent's piece size. These are the results I got from a test torrent with 256KB sized pieces with a single tcp peer with ~150ms latency and ~2.5MB/s bandwidth:

| `max_superseed_bytes` | Equiv number of 256KB pieces offered| Observed speed |
|---|---|---|
|524288|2 (current behavior)|~0.35MB/s|
|1048576|4|~0.4MB/s|
|2097152|8|~0.8MB/s|
|4194304|16|~1.2MB/s|
|8388608|32|~1.8MB/s|
|10485760|40|~2.1MB/s|
|12582912|48|~2.3MB/s|
|16777216|64|~2.3MB/s|
|x|All - Standard seeding|~2.3MB/s|

While this seems to be enough to improve performance, it is still a bit blunt and requires manually configuring to suit the upload speed. Also there is no point, for instance, announcing 8MB (32x256KB) worth of pieces to a peer that's only downloading from us at 50KB/s. With a large number of slower peers this may impact superseeding efficiency.

A better solution, and the one in this PR, is to dynamically increase/decrease the number of pieces superseeded to a peer depending on our upload speed to them, similar to the way `request_queue_time` is used for sizing the download request queue. Having added a `superseed_scale_time` setting I reran the previous test, and the number of pieces superseeding automatically increased to 53 as the speed increased and stabilized, which is about the sweet spot in this case and on par with standard seeding. That was with `superseed_scale_time=6` - ie. offering as many pieces at a time as we expect the peer to download from us in 6s.

Running some further tests involving higher latency (~550ms) it looks like a higher default value may be worthwhile, so maybe 8-10s or so. I don't think a higher value is wasteful, as actual requests is still limited by `max_allowed_in_request_queue` - it'll just give the peer a larger range of pieces to choose from. I wouldn't imagine it's high enough to effect superseeding efficiency at this level either.

I'm reasonably confident in the logic behind this PR - I don't think superseeding efficiency would be impacted, and the performance improvements seem easy enough to demonstrate. It does stray from the [BEP](https://www.bittorrent.org/beps/bep_0016.html) somewhat, though I see that 'strict' superseeding as the BEP decsribes has been deprecated since 1.2.6, so hopefully trying to 'improve' on the BEP in this way isn't an issue. 

As I'm still relatively new to c++ stuff I can't rule out amateur errors in the code. That said, I'm pleased with how it's been working so far, and I think its time to get some feedback on it and whether this behavior, or something similar, might be worth implementing.


src/peer_connection.cpp
```diff
@@ -1973,7 +1974,17 @@ namespace libtorrent {
...
-				superseed_piece(index, t->get_piece_to_super_seed(m_have_piece));
+#ifndef TORRENT_DISABLE_LOGGING
+				if (should_log(peer_log_alert::incoming))
+					peer_log(peer_log_alert::incoming, "HAVE"
+						, "got HAVE message for superseeded piece %d"
+						, static_cast<int>(index));
+#endif
+				// Don't replace the piece if we're above the current target
+				if (target_superseed_pieces() < m_superseed_pieces.size())
+					m_superseed_pieces.erase(std::remove(m_superseed_pieces.begin(), m_superseed_pieces.end(), index)
```
arvidn, line 1985: `std::remove()` was introduced in C++17, and this branch of libtorrent is locked in to C++11. `RC_2_0` is C++14 and `master` is C++17.

However, it seems the `m_superseed_pieces` my be better suited as a `set`. As far as I can tell, we only use it to insert piece indices, and look up whether an index is in there, and then remove an index. (it could also be a bitmask, but that would most likely be less efficient)

src/bt_peer_connection.cpp
```diff
@@ -1999,12 +1999,6 @@ namespace {
 			// if we are super seeding, pretend to not have any piece
 			// and don't send a bitfield
 			m_sent_bitfield = true;
-
-			// bootstrap super-seeding by sending two have message
-			piece_index_t piece = t->get_piece_to_super_seed(get_bitfield());
-			if (piece >= piece_index_t(0)) superseed_piece(piece_index_t(-1), piece);
-			piece = t->get_piece_to_super_seed(get_bitfield());
-			if (piece >= piece_index_t(0)) superseed_piece(piece_index_t(-1), piece);
```
arvidn, line 2007: how does the "bootstrapping" happen now? what triggers us to send the first `HAVE` message?

src/peer_connection.cpp
```diff
@@ -4692,14 +4707,31 @@ namespace libtorrent {
 		write_have(new_piece);
 
 		if (replace_piece >= piece_index_t(0))
+			m_superseed_pieces.erase(std::remove(m_superseed_pieces.begin(), m_superseed_pieces.end(), replace_piece), m_superseed_pieces.end());
+		m_superseed_pieces.insert(m_superseed_pieces.begin(), new_piece);
+
+		// Check we haven't exceeded the piece limit. This may happen
+		// if a peer filters sending redundant HAVE messsages
```
arvidn, line 4714: wouldn't it be simpler to avoid adding the piece rather than reverting it?

src/peer_connection.cpp
```diff
@@ -4692,14 +4707,31 @@ namespace libtorrent {
...
-		m_superseed_piece[0] = new_piece;
+	}
+
+	uint32_t peer_connection::target_superseed_pieces() const
+	{
+		std::shared_ptr<torrent> t = m_torrent.lock();
+		torrent_info const& ti = t->torrent_file();
+		int const ps = int(ti.piece_length());
+		TORRENT_ASSERT(ps > 0);
+		return std::max(m_settings.get_int(settings_pack::superseed_scale_time) * m_statistics.upload_payload_rate() / ps, 2);
```
arvidn, line 4734: I think it's worth a comment here, explaining that we need at least `2` here, in order to bootstrap and offer *any* pieces before we have started uploading (and the upload rate is 0).

In fact, it could probably be 1

src/settings_pack.cpp
```diff
@@ -355,6 +355,7 @@ constexpr int CLOSE_FILE_INTERVAL = 0;
 		SET(rate_choker_initial_threshold, 1024, nullptr),
 		SET(upnp_lease_duration, 3600, nullptr),
 		SET(max_concurrent_http_announces, 50, nullptr),
+		SET(superseed_scale_time, 10, nullptr),
```
arvidn, line 358: 10 second seems like a pretty high default. This represents the expected round-trip time from making a request and receiving it. I would think 3 or 4 seconds would seem a bit more reasonable. Did you measure poor performance with setting this lower?

arvidn (commented):
Thanks for your PR! I have 3 comments:

1. I'm curious of what your use case of super-seeding is. I was under the impression that it's not very commonly used anymore. And you're right that I deprecated the strict implementation of the BEP as it's essentially useless in practice. I'm worried about the increase in complexity vs. benefit with improving super-seeding.

2. Putting a change like this into `RC_1_2` is a bit risky. I'd prefer to put it in `master`. The release branches primarily receive fixes and minor updates.

3. There would need to be tests for this. Ideally the logic could be isolated into free functions or a separate class, that can be unit tested. It also shouldn't be too much trouble to write a simulation that exercises the super seeding transfer (see `simulation/test_transfer.cpp`)

arvidn:
I think the spell-checker would be happier if you turned "superseeded" into "super-seeded"

thrnz:
1. I have a number of otherwise seedless torrents with only a limited upload speed. I've found that using superseeding makes more efficient use of what I've got to spare compared to standard seeding.

2. Sure. I'll rebase this to master. 

3. I'll look into this and see if I can sort something. It might above my level at this stage but keen to learn.

src/bt_peer_connection.cpp
thrnz: I found that bootstrapping here could fail. `superseed_piece()` requires `in_handshake()` to be false, whereas it can still be true here when `write_bitfield()` is called from `on_receive_impl()`.

'Bootstrapping' now happens in `second_tick()` when we decide whether to increase the number of pieces (ie. from 0 to the minimum of 2) once we're no longer in_handshake().

1265578519:
你可以在制作种子的时候设置较小的区块就可以解决这个问题，更改这个些代码可能会导致其他客户端不适配引起BUG，对方客户端区块未下载完成无法尝试上传

arvidn:
> You can solve this problem by setting a smaller block when making the seed. Changing these codes may cause other clients to not adapt and cause BUG, and the other client’s block cannot be uploaded without downloading.

The piece size is immutable, you can't change it without changing the info-hash

src/settings_pack.cpp
thrnz: I did find I had to set this higher than seemed intuitive to match standard seeding speeds, especially when testing with a higher latency. ~6s seemed about right when initially testing at ~150ms and 8-10s at ~550ms. It didn't seem to scale as I'd expect.

I do plan on doing more thorough testing though and recording data for various rtts to see what the best default might be more for the most common conditions.

src/peer_connection.cpp
thrnz: I left the minimum at 2 as that's the current behavior, but 1 probably makes more sense to start with, especially now that its able to grow.

src/peer_connection.cpp
thrnz: I could probabaly rework it as a `set` instead, though might need to change how we handle peers that filter sending redundant `HAVE` messages.

Previously we dealt with this by waiting for a peer to be uninterested for 10 seconds, and then offering new pieces until the peer becomes interested again. I adapted this for use with multiple pieces by using a `vector` - adding the newly offered piece to the beginning, and resizing off the oldest piece/s at the end when it grows too big. 

Using a `set`, I guess we would then clear it whenever we try offering a new piece to such a peer, as this would be closest to the current behavior. Such a peer wouldn't be able to benefit from having multiple pieces to choose from anyway. Thankfully, so far in testing, I've noticed very few peers that seem to behave this way anwyay.

There also might be some weirdness when a peer only has a few pieces to go. If I'm reading it correctly, `torrent::get_piece_to_super_seed()` can still return a piece that is already being superseeded to the peer if there are no other pieces left to pick from. We would have to take this into account when calculating a target number of pieces to superseed, otherwise we would keep adding duplicate pieces, but the set size would never grow to the target as there aren't enough unique needed pieces left. This is probabaly something that needs addressing anyway. As it stands, while functional, the `vector` would be filling with duplicate pieces as the peer nears completion and we would be re-sending out `HAVE` messages for the last few pieces that we've already made available.

arvidn:
Another, pre-existing, issue is that this logic can easily be gamed. You just need to send a HAVE message back to get another super seeded piece to open up. You can do that until the "right" piece opens up.

I think that mitigation would be to keep track of bytes we upload, and not always trust HAVE messages. It's a tricky balance though, since there may be other seeds sending data too.

thrnz:
I've rebased to master and changed m_superseed_pieces from `vector` to `set`.

`target_superseed_pieces()` also now takes into account the number of pieces the peer has remaining, and makes sure it never returns a target higher than that. That should prevent any weirdness as the upload nears completion.

Dealing with peers that filter redundant `HAVE` messages remains unchanged at this stage.

I did some experimenting with different values of `superseed_scale_time` with high latencies and got the following. This was with a torrent with 256KB pieces and a single TCP peer.

Peer latency | `superseed_scale_time` | Observed speed
--- | --- | ---
300ms | 2s | 0.1-0.2MB/s
300ms | 3s | 1.0-1.3MB/s (slow to increase)
300ms | 4s | 1.6-1.8MB/s
300ms | 6s | 2.1-2.3MB/s
300ms | 8s | 2.1-2.3MB/s
300ms | normal seeding | 2.1-2.3MB/s
600ms | 3s | 0.1MB/s
600ms | 4s | 0.1MB/s
600ms | 5s | 1.3-1.7MB/s (slow to increase)
600ms | 6s | 2.0-2.3MB/s (slow to increase)
600ms | 8s | 2.0-2.3MB/s
600ms | normal seeding | 2.1-2.3MB/s

From that I guess 6s should be enough for most cases to reach parity with standard seeding.

thrnz:
The current `super_seeding` and `strict_super_seeding` tests in `simulation/test_super_seeding.cpp` appear to always pass even if super seeding is intentionally broken. If I'm reading things correctly, it appears that both tests are finishing before the downloads have a chance to complete/fail.

I've changed them so they now wait until the downloaders either successfully become seeds or timeout/fail before the tests terminate.

src/peer_connection.cpp
thrnz: If we only offer a single piece to a peer, then they'll probably send us a 'not interested' after each piece, causing us to prematurely choke them and sending the to the back of the queue. Offering at least 2 pieces may help to prevent this.

thrnz:
I've done some further testing and got the following `bmon` graphs showing network utilization for various values of `superseed_scale_time`. These were done while super-seeding multiple torrents with varying piece sizes to a mix of both TCP and uTP peers of various latencies, so is more 'real world' than the previous tests which were only to a single peer.


```
                           superseed_scale_time=0
                  (ie. single piece offered/current behavior)
     MiB                      (TX Bytes/second)
    2.63 ................................|..|........................
    2.19 |.........|..||.................|..|...|....................
    1.75 |..|...|..||.||.........|||...|.|..||..|...||...|.......||..
    1.31 |..|..|||.||.||.||.|.||||||||||||.|||.||...|||||||..||.|||||
    0.88 ||||||||||||||||||.||||||||||||||||||.||||||||||||.|||||||||
    0.44 ||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
         1   5   10   15   20   25   30   35   40   45   50   55   60

                            superseed_scale_time=2
     MiB                      (TX Bytes/second)
    2.77 ...|......................................|.................
    2.31 ...||..||........|.|.||..|...|.|......|..|||...|......|...||
    1.84 .|.||..||...|.|..|||.||..|..||.|.|||..|..||||.||.|||.|||.|||
    1.38 |||||..|||||||||.||||||.||||||||.|||.|||||||||||||||||||||||
    0.92 ||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
    0.46 ||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
         1   5   10   15   20   25   30   35   40   45   50   55   60

                            superseed_scale_time=4
     MiB                      (TX Bytes/second)
    2.88 ||||.|||.|||...|..||||..|.|...||..|....||.....|....|.|||||.|
    2.40 |||||||||||||||||.|||||||||||||||||||.|||||||.|....|||||||||
    1.92 |||||||||||||||||.||||||||||||||||||||||||||||||||||||||||||
    1.44 ||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
    0.96 ||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
    0.48 ||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
         1   5   10   15   20   25   30   35   40   45   50   55   60

                            superseed_scale_time=6
     MiB                      (TX Bytes/second)
    2.88 ||||||||||||||....||||||||||||||||.|||.||||||||||||||||||.||
    2.40 ||||||||||||||||..||||||||||||||||||||||||||||||||||||||||||
    1.92 ||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
    1.44 ||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
    0.96 ||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
    0.48 ||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
         1   5   10   15   20   25   30   35   40   45   50   55   60

                            superseed_scale_time=8
     MiB                      (TX Bytes/second)
    2.88 |||||||||||||||||||||||||||||||||||||.|||||||.||||||||||.|||
    2.40 ||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
    1.92 ||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
    1.44 ||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
    0.96 ||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
    0.48 ||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
         1   5   10   15   20   25   30   35   40   45   50   55   60
```

PR #6349, opened 2021-07-30
Add generate_fingerprint_bytes

### LINKED ISSUE #5985: Python bindings: add bytes accessors for data
A number of fields in the torrent protocol (or libtorrent) are encoding-agnostic byte strings.

We should ensure there's always a way to access `bytes`, if `bytes` may be passed as input, or if the underlying data is encoding-agnostic.

One pattern in python for this is "runtime templating", where the type of output changes with the type of input. For instance, `os.listdir(".")` returns a list of `str`, while `os.listdir(b".")` returns a list of `bytes`. However, the python bindings have historically allowed either `bytes` or `str` as input for `char *` or `std::string`, but tend to always return `str`. If we make output type vary with input, we'll probably break a lot of clients.

The only alternative I can think of is have `*_bytes` versions of functions.

Here's a proposed list of functions returning `str`, which should have pair functions returning `bytes`:

* `file_storage.symlink()`: `symlink_bytes()`
* `file_storage.file_path()`: `file_path_bytes()`
* `file_storage.file_name()`: `file_name_bytes()`
* `file_storage.name()`: `name_bytes()`
* `generate_fingerprint()`: `generate_fingerprint_bytes()`
* `torrent_info.web_seeds()`: `web_seeds_bytes()` (the `"auth"` field may be binary data)
* `torrent_info.name()`: `name_bytes()`
* `torrent_info.creator()`: `creator_bytes()`
* `torrent_info.comment()`: `comment_bytes()`
* `torrent_info.collections()`: `collections_bytes()`

I'll add more as I find them.

Adding these new functions should be a non-breaking change.

---

## Thoughts on file_storage

There's an argument that the fields of `file_storage` should be given "pathname treatment", and use the filesystem encoding as detailed in #5984. This would make sense when creating torrents off the local filesystem.

However, currently `file_storage` is used in lots of different contexts. A torrent found randomly on the internet and converted into a `torrent_info` has a `file_storage` whose bytes may represent an encoding from some other system.

So I think the best thing is to add `*_bytes()` accessors to `file_storage`, and clients can apply whatever decoding (or guesses, with fallbacks) as necessary.

Notably, that means that the proper usage for creating files from the filesystem is a little funky:

```python
fs = lt.file_storage()
lt.add_files(fs, my_root_dir)
for i in fs.num_files():
    # file_storage is abstract, but we know the encoding came from the filesystem
    print(os.fsdecode(fs.file_path_bytes(i)))
```

The only alternative I can think of here is that `file_storage` could hold a flag to remember how its inputs are encoded, such that `fs.file_path(i)` is the same as `os.fsdecode(fs.file_path_bytes(i))` if `fs` was populated from the filesystem, but I can't see any way to add this feature without making it confusing and error prone.

AllSeeingEyeTolledEweSew:
Closes #5985


bindings/python/tests/fingerprint_test.py
```diff
@@ -4,37 +4,34 @@
...
+            lt.generate_fingerprint_bytes(b"ABCD", 1, 2, 3, 4),
             b"-AB1234-",
         )
 
         # short name
         self.assertEqual(
-            lt.generate_fingerprint_bytes(b"A", 1, 2, 3, 4),  # type: ignore
-            b"-A\x001234-",
+            lt.generate_fingerprint_bytes(b"A", 1, 2, 3, 4),
+            b"---1234-",
```
arvidn, line 17: this doesn't look right. what happened to the `A`?

AllSeeingEyeTolledEweSew: The original test was wrong. The behavior of `generate_fingerprint_bytes()` is analogous to `generate_fingerprint()` as expected.

The tests I wrote with `@unittest.skip()` were not all correct (because they wouldn't run, so I couldn't verify the desired condition would be correct until the fix was made).

I expect that here:
 * I wrote and validated the test for `fingerprint()` (`str(fingerprint("A", 1, 2, 3, 4)) == "-A\x001234-"`)
 * I copied that output into some test cases for `generate_fingerprint()`
 * I realized we should have `generate_fingerprint_bytes()`
 * I split the functions into `test_generate()` (with `@unittest.skip()`) and `test_generate_str()`
 * I validated `test_generate_str()` (`generate_fingerprint("A", 1, 2, 3, 4) == "---1234-"`, unlike `fingerprint()`)
 * I forgot to analogously update `test_generate()`

arvidn: I expect the "A" that's passed it would be used in the final fingerprint. It looks like a bug that it turns into `--` instead. I can take a closer look at what's going on

arvidn: I see, there's a check and the `name` *has* to be two bytes long

arvidn (approved):

Issue #5989, opened 2021-02-18
Python bindings: various segfaults, should raise exception instead


AllSeeingEyeTolledEweSew:
The following code snippets cause segfaults. Segfaults shouldn't ever happen in python.

* `fs = file_storage(); fs.at(0)`: should raise `IndexError`
* `fs = file_storage(); fs.at(-1)`: should raise `IndexError`
* `fs = file_storage(); fs.symlink(0)`: should raise `IndexError`
* `fs = file_storage(); fs.symlink(-1)`: should raise `IndexError`
* `fs = file_storage(); fs.file_path(0)`: should raise `IndexError`
* `fs = file_storage(); fs.file_path(-1)`: should raise `IndexError`
* `fs = file_storage(); fs.file_name(0)`: should raise `IndexError`
* `fs = file_storage(); fs.file_name(-1)`: should raise `IndexError`
* `fs = file_storage(); fs.file_entry(0)`: should raise `IndexError`
* `fs = file_storage(); fs.file_entry(-1)`: should raise `IndexError`
* `fs = file_storage(); fs.file_offset(0)`: should raise `IndexError`
* `fs = file_storage(); fs.file_offset(-1)`: should raise `IndexError`
* `fs = file_storage(); fs.file_flags(0)`: should raise `IndexError`
* `fs = file_storage(); fs.file_flags(-1)`: should raise `IndexError`
* `fs = file_storage(); fs.rename_file(0, "file.txt")`: should raise `IndexError`
* `fs = file_storage(); fs.rename_file(-1, "file.txt")`: should raise `IndexError`
* `fs = file_storage(); ct = create_torrent(fs); ct.set_hash(0, b"a" * 20)`: should raise `IndexError`
* `fs = file_storage(); ct = create_torrent(fs); ct.set_hash(-1, b"a" * 20)`: should raise `IndexError`
* `fs = file_storage(); ct = create_torrent(fs); ct.set_file_hash(0, b"a" * 20)`: should raise `IndexError`
* `fs = file_storage(); ct = create_torrent(fs); ct.set_file_hash(-1, b"a" * 20)`: should raise `IndexError`

PR #6352, opened 2021-07-31
Check bounds on file_storage and create_torrent

AllSeeingEyeTolledEweSew:
Progress toward #5989


AllSeeingEyeTolledEweSew:
I guess there's an argument for moving these exceptions to the C++ code. I don't know if that's appropriate, but I can do that instead of you like.

PR #6363, opened 2021-08-01
Make dht mutable item keys be bytes-only

AllSeeingEyeTolledEweSew:
Progress toward #5988

### DIFF
--- a/bindings/python/src/session.cpp
+++ b/bindings/python/src/session.cpp
@@ -703,12 +703,12 @@ namespace
 	}
 
 #ifndef TORRENT_DISABLE_DHT
-    void dht_get_mutable_item(lt::session& ses, std::string key, std::string salt)
+    void dht_get_mutable_item(lt::session& ses, bytes key, bytes salt)
     {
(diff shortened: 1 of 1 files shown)


bindings/python/src/session.cpp
```diff
@@ -728,20 +728,20 @@ namespace
         sig = sign.bytes;
     }
 
-    void dht_put_mutable_item(lt::session& ses, std::string private_key, std::string public_key,
-        std::string data, std::string salt)
+    void dht_put_mutable_item(lt::session& ses, bytes private_key, bytes public_key,
+        bytes data, bytes salt)
     {
         TORRENT_ASSERT(private_key.size() == 64);
         TORRENT_ASSERT(public_key.size() == 32);
```
arvidn, line 735: these asserts should probably be run-time checks

arvidn (approved):

bindings/python/src/session.cpp
AllSeeingEyeTolledEweSew: Did that in #6355

PR #6365, opened 2021-08-01
Deprecate add_piece(i, 'str')

AllSeeingEyeTolledEweSew:
Progress toward #5988

### DIFF
--- a/bindings/python/src/torrent_handle.cpp
+++ b/bindings/python/src/torrent_handle.cpp
@@ -418,6 +418,7 @@ std::shared_ptr<const torrent_info> get_torrent_info(torrent_handle const& h)
 void add_piece_str(torrent_handle& th, piece_index_t piece, char const *data
     , add_piece_flags_t const flags)
 {
+    python_deprecated("add_piece with str is deprecated");
     th.add_piece(piece, data, flags);
 }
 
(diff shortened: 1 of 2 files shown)


arvidn:
It looks like the `bytes`-version is a bit odd too:

```
void add_piece_bytes(torrent_handle& th, piece_index_t piece, bytes data
    , add_piece_flags_t const flags)
{
    th.add_piece(piece, data.arr.c_str(), flags);
}
```

there should probably be a run-time check that `data.arr` has the right length (and fail if it doesn't). Also, there's really no need to get a `c_str()` there, `data()` seems more correct

arvidn (approved):

AllSeeingEyeTolledEweSew:
> there should probably be a run-time check that `data.arr` has the right length (and fail if it doesn't). Also, there's really no need to get a `c_str()` there, `data()` seems more correct

Since `torrent_handle::add_piece()` returns `void`, it's currently asynchronous. I'd be reluctant to make it synchronous by fetching the piece length for the check.

Maybe `add_piece(piece_index_t, char const*, add_piece_flags_t={})` should be deprecated in favor of a new version which takes a `size_t`

I'm not that worried about it since `add_piece()` is increasingly targeted to unit tests

arvidn:
Right, we don't know the piece size at that point

PR #6366, opened 2021-08-01
Deprecate use of bytes with paths in file_storage

AllSeeingEyeTolledEweSew:
Progress toward #5988

### DIFF
--- a/bindings/python/src/create_torrent.cpp
+++ b/bindings/python/src/create_torrent.cpp
@@ -137,10 +137,24 @@ namespace
         add_files(fs, file, [&](std::string const& i) { return cb(i); }, flags);
     }
 
-    void add_file(file_storage& fs, std::string const& file, std::int64_t size
+    void add_file0(file_storage& fs, string_view const file, std::int64_t size
+       , file_flags_t const flags, std::time_t md, string_view const link)
+    {
+       fs.add_file(std::string(file), size, flags, md, std::string(link));
+    }
+
+    void add_file1(file_storage& fs, bytes const& file, std::int64_t size
        , file_flags_t const flags, std::time_t md, std::string link)
     {
-       fs.add_file(file, size, flags, md, link);
+       python_deprecated("add_file with bytes is deprecated");
+       fs.add_file(file.arr, size, flags, md, link);
+    }
+
+    void add_file2(file_storage& fs, string_view const file, std::int64_t size
+       , file_flags_t const flags, std::time_t md, bytes link)
+    {
+       python_deprecated("add_file with bytes is deprecated");
+       fs.add_file(std::string(file), size, flags, md, link.arr);
     }
 
     void add_tracker(create_torrent& ct, std::string url, int tier)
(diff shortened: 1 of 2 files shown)


AllSeeingEyeTolledEweSew:
The windows errors make no sense. Looks like the network changed radically in the middle of the test. Can you re-run the workflow?

bindings/python/src/create_torrent.cpp
```diff
@@ -129,10 +129,24 @@ namespace
         add_files(fs, file, [&](std::string const& i) { return cb(i); }, flags);
     }
 
-    void add_file(file_storage& fs, std::string const& file, std::int64_t size
+    void add_file0(file_storage& fs, string_view const& file, std::int64_t size
```
arvidn, line 132: the `string_view` should be passed by value, since it already acts as a pointer. (it should also be made `const`)

bindings/python/src/create_torrent.cpp
```diff
@@ -129,10 +129,24 @@ namespace
...
+
+    void add_file1(file_storage& fs, bytes const& file, std::int64_t size
        , file_flags_t const flags, std::time_t md, std::string link)
     {
-       fs.add_file(file, size, flags, md, link);
+       python_deprecated("add_file with bytes is deprecated");
+       fs.add_file(file.arr, size, flags, md, link);
+    }
+
+    void add_file2(file_storage& fs, string_view const& file, std::int64_t size
```
arvidn, line 145: same comment applies to this `string_view`

bindings/python/src/create_torrent.cpp
```diff
@@ -179,10 +193,17 @@ namespace
         return fs.file_flags(index);
     }
 
-    void rename_file0(file_storage& fs, file_index_t index, std::string const& path)
+    void rename_file0(file_storage& fs, file_index_t index, string_view const& path)
```
arvidn, line 196: and here

bindings/python/src/create_torrent.cpp
```diff
@@ -192,12 +213,21 @@ namespace
         return fs.at(index);
     }
 #endif
+
+    void set_name0(file_storage& fs, string_view const& name)
```
arvidn, line 217: and here

bindings/python/src/create_torrent.cpp
```diff
@@ -129,10 +129,24 @@ namespace
...
 
-    void add_file(file_storage& fs, std::string const& file, std::int64_t size
+    void add_file0(file_storage& fs, string_view const& file, std::int64_t size
+       , file_flags_t const flags, std::time_t md, string_view link)
+    {
+       fs.add_file(std::string(file), size, flags, md, std::string(link));
+    }
+
+    void add_file1(file_storage& fs, bytes const& file, std::int64_t size
        , file_flags_t const flags, std::time_t md, std::string link)
```
arvidn, line 139: I notice you left this `link` parameter as `std::string`. is it the case that `std::string` accepts both bytes and unicode strings (converted to utf-8), whereas `string_view` only accepts unicode strings?

AllSeeingEyeTolledEweSew: Yep

Let me know if there's some other way of indicating `str`-only input, other than `string_view`. It's a bit weird to me to use it this way since it's meant as a pointer, unlike `std::string` and `bytes`

arvidn (approved):

bindings/python/src/create_torrent.cpp
AllSeeingEyeTolledEweSew: Right. In particular the builtin `std::string` conversion comes from boost-python, not python itself. It's not clear there's an option to disable it. (IMO it's an impedence mismatch with python 3, and adds to my gripes with boost-python).

I meant I wasn't sure whether the `string_view` converter was *intended* to match `str` only, or just intended for convenience wrapping functions that take `string_view`.

In the latter case, someone might decide to make it match `bytes` one day, to match the non-optional behavior with `std::string`. Then maybe it's better to make a custom C++ `str` class analogous to our custom `bytes` class, specifically intended to only match `str`s from python.

PR #6369, opened 2021-08-01
Fix binding for file_progress_flags_t

AllSeeingEyeTolledEweSew:
This is the same as #6350, but it updates the new python test suite in master.

Progress toward #5995


arvidn:
@cas-- this should be back-ported `RC_2_0`, right?

cas--:
Yes at least this one

arvidn:
https://github.com/arvidn/libtorrent/pull/6492

PR #6371, opened 2021-08-01
Fix bindings for members of picker_log_alert

AllSeeingEyeTolledEweSew:
Progress toward #5995

### DIFF
--- a/bindings/python/src/alert.cpp
+++ b/bindings/python/src/alert.cpp
@@ -1027,8 +1027,28 @@ void bind_alert()
 
     class_<picker_log_alert, bases<peer_alert>, noncopyable>(
        "picker_log_alert", no_init)
-        .add_property("picker_flags", &picker_log_alert::picker_flags)
+        .add_property("picker_flags", make_getter(&picker_log_alert::picker_flags, by_value()))
         .def("blocks", &picker_log_alert::blocks)
+        ;
+
+    enum_<picker_flags_t>("picker_flags_t")
+        .value("partial_ratio", picker_log_alert::partial_ratio)
+        .value("prioritize_partials", picker_log_alert::prioritize_partials)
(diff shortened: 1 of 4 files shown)


arvidn (approved):

AllSeeingEyeTolledEweSew:
I think one of the existing merges got messed up. There are build errors in master now. I had to correct it in all of my rebases.

Just merge one, and I'll rebase the rest again

bindings/python/src/alert.cpp
```diff
@@ -1017,10 +1017,30 @@ void bind_alert()
 
     class_<picker_log_alert, bases<peer_alert>, noncopyable>(
        "picker_log_alert", no_init)
-        .add_property("picker_flags", &picker_log_alert::picker_flags)
+        .add_property("picker_flags", make_getter(&picker_log_alert::picker_flags, by_value()))
         .def("blocks", &picker_log_alert::blocks)
         ;
 
+    enum_<picker_flags_t>("picker_flags_t")
```
arvidn, line 1024: iirc, for flags in other alerts, I put them as constant members of those alert classes, to mimic how they are accessed in C++. Did you turn this into an enum to deliberately strengthen type-safety? (since previously it was treated like all other strongly-typed flags, and just converted to `int` on the python side

AllSeeingEyeTolledEweSew: Yes, this is just for type safety on the python side.

I think there are probably more-pythonic and more-type-safe options than even this, but I think `enum_` is better than static `int` members on the class, and follows our existing patterns.

IIUC, the main benefit of python access patterns strictly following C++ ones is just as pseudo-documentation, in lieu of full docs for the python bindings. I'm considering that the type stubs (`__init__.pyi`) does this job now, so we can make the python access patterns whatever we want.

arvidn (approved):

PR #6374, opened 2021-08-02
Fix dht_[im]mutable_item_alert.item to just be the item

AllSeeingEyeTolledEweSew:
Progress toward #5995

This is a radical change of the binding in the alerts, but I think it makes sense. It looks like the existing binding hasn't really worked for some time, or at least I haven't been able to make it work. Anyway this makes much more sense than having `alert.item` return a dict of all the other members of the alert.

My current tests are only generating "failure" versions of `dht_[im]mutable_item_alert`, where the item is unset. I need to redo the tests to generate real data. I expect I need to change dht bootstrapping, which will probably change *all* the dht tests, so I will do that in a later PR after the current ones are done.

### DIFF
--- a/bindings/python/install_data/libtorrent/__init__.pyi
+++ b/bindings/python/install_data/libtorrent/__init__.pyi
@@ -479,7 +479,7 @@ class dht_get_peers_reply_alert(alert):
     info_hash: sha1_hash
 
 class dht_immutable_item_alert(alert):
-    item: _Entry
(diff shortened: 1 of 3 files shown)


AllSeeingEyeTolledEweSew:
What's up with Windows workflows?

AllSeeingEyeTolledEweSew:
Closed in favor of #6391

PR #6409, opened 2021-08-19
Fixed last_seen_complete not being applied to added torrents

thrnz:
I've noticed that the [`last_seen_complete`](https://www.libtorrent.org/reference-Add_Torrent.html#last_seen_complete) paramater doesn't appear to be applied as expected when adding a torrent to a session. This is most noticeable when adding torrents from resume data (ie. when a torrent client is restarted) as all torrents will report having never seen a complete copy of the torrent before, despite previously having done so. 

If I'm reading `src/torrent.cpp` correnctly, it appears that `m_last_seen_complete`  isn't being set to the paramater pack's value at creation and remains at the default.

After this change, `last_seen_complete` appears to work as expected, and is no longer lost when a torrent client is restarted.


arvidn:
this looks good. I bet this is a problem in earlier versions as well. I'll take a look

arvidn:
how does this look? https://github.com/arvidn/libtorrent/pull/6411

thrnz:
Looks good to me

PR #6415, opened 2021-08-21
add invariant check to merkle_tree

arvidn:
and ensure it can maintain its invariant by making the API such that any hashes to be added must be proven to be valid in the same call


arvidn:
test_merkle_tree is incomplete..

PR #6422, opened 2021-08-23
fix for building with boost-1.77 on windows

arvidn:


sledgehammer999:
This works! I can compile the 64bit version on Windows now.

PR #6440, opened 2021-08-31
Add a cibuildwheel workflow to RC_1_2

AllSeeingEyeTolledEweSew:
This is a port of #6188 to RC_1_2.

No major differences between #6188 and this PR.

[Test workflow run here](https://github.com/AllSeeingEyeTolledEweSew/libtorrent/actions/runs/1184562489)

### DIFF
--- a/tools/run_cibuildwheel.py
+++ b/tools/run_cibuildwheel.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python
+
+import logging
+import os
+import runpy
+from shlex import quote
+import sys
+
+_LOG = logging.getLogger(__name__)
+
+
+def set_env(name: str, value: str) -> None:
+    _LOG.info("set %s = %s", name, value)
+    os.environ[name] = value
+
+
+def append_inner_env(name: str, value: str) -> None:
+    existing = os.environ.get("CIBW_ENVIRONMENT", "")
+    set_env("CIBW_ENVIRONMENT", f"{existing} {name}={value}")
+
+
+def setup_env() -> None:
+    # Per https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html,
+    # we need gcc 5.1 for stable C++11 ABI.
+    # manylinux1 has gcc 4.8.2.
+    # manylinux2010 and manylinux2014 have a backported gcc at version
+    # 8+. Their glibcs are old; they achieve support for newer gcc by
+    # statically linking backported newer glibc functions.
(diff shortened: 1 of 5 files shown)


tools/setup_boost.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+set -e
```
arvidn, line 3: I think it would be helpful with a comment here, describing what this script does and what it's used for. I don't think it's obvious that it's used for CI

tools/run_cibuildwheel.py
```diff
@@ -0,0 +1,101 @@
...
+    boost_version = os.environ["BOOST_VERSION"]
+
+    if sys.platform == "win32":
+        boost_root = "c:/boost"
+    else:
+        boost_root = "/tmp/boost"
+    boost_build_path = f"{boost_root}/tools/build"
+
+    before_all_commands = [
+        f"./tools/setup_boost.sh {quote(boost_version)} {quote(boost_root)}",
```
arvidn, line 67: do we download boost every time this is run? even if we've run it before?

arvidn (commented):
it makes me nervous to make large changes like these to the build files, in the stable release. This would have to be merged into `RC_2_0` and `master` too..

AllSeeingEyeTolledEweSew:
> it makes me nervous to make large changes like these to the build files, in the stable release. This would have to be merged into `RC_2_0` and `master` too..

Per the description this is a rebase of #6188 from `RC_2_0` to `RC_1_2`.

I started a rebase on `master` as well but it fails due to #6447.

tools/run_cibuildwheel.py
AllSeeingEyeTolledEweSew: On linux, `cibuildwheel` spawns a docker container and does the builds there. On other platforms it runs build commands nakedly.

I thought I'd written that in comments, but I guess it got lost in refactors.

AllSeeingEyeTolledEweSew:
Would you mind putting further code comments in #6188? It's going to be confusing to jump between the rebase PRs, trying to figure out which discussion is where.

AllSeeingEyeTolledEweSew:
Drafting until I update #6188 per #6448

AllSeeingEyeTolledEweSew:
Dropping this in favor of #6188. I expect it will merge cleanly onto `RC_2_0`, and work without changes. It will need some changes on `master` to work though, due to C++17.

PR #6446, opened 2021-09-02
Loosen restrictions on error categories

AllSeeingEyeTolledEweSew:
This is required for `cibuildwheel` in master. The current tests don't seem to work on 32-bit linux. The storage implementation returns different error codes for the tested conditions.


arvidn (approved):

arvidn:
ideally the test would ensure the error code is one of the ones we expect, either from `mmap` or `posix`

AllSeeingEyeTolledEweSew:
Given the error codes aren't consistent for a given condition, an app probably shouldn't have conditional logic on the error code. It should just check the code is nonzero. I changed the tests to do this.

PR #6450, opened 2021-09-07
Migrate to OpenSSL EVP interface for hashing

Chocobo1:
The low-level interface is deprecated since OpenSSL v3.0 [1].

[1] https://www.openssl.org/docs/manmaster/man7/migration_guide.html "Deprecated low-level digest functions"

### DIFF
--- a/include/libtorrent/aux_/hasher512.hpp
+++ b/include/libtorrent/aux_/hasher512.hpp
@@ -56,9 +56,9 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #elif defined TORRENT_USE_LIBCRYPTO
 
-	extern "C" {
-	#include <openssl/sha.h>
-	}
+extern "C" {
+#include <openssl/evp.h>
+}
 
 #else
 #include "libtorrent/aux_/sha512.hpp"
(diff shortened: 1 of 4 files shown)


include/libtorrent/aux_/hasher512.hpp
```diff
@@ -122,7 +122,7 @@ namespace aux {
 #elif TORRENT_USE_CRYPTOAPI_SHA_512
 		aux::crypt_hash<CALG_SHA_512, PROV_RSA_AES> m_context;
 #elif defined TORRENT_USE_LIBCRYPTO
-		SHA512_CTX m_context;
+		EVP_MD_CTX *m_context;
```
Chocobo1, line 125: It seems openssl has intentionally made these structures opaque as it forward declares it...
https://wiki.openssl.org/index.php/OpenSSL_1.1.0_Changes#Compatibility_Layer

Chocobo1: @arvidn 
This changes the structure size so it breaks ABI compatibility right?
Should I retarget to master branch instead?

arvidn: yes, ABI changes need to go into `master`. Alternatively, it could be a new build configuration with its own inline namespace, separating it from previous symbols. But I suspect that's overkill. The lead time for this API to go away is probably long

arvidn:
it looks like openssl allocates the state on the heap now, hence the need to delete it. this means you need to implement the assignment and copy constructor as well, for `hasher` and `hasher512`.

These types should probably be movable as well, but I'm not sure any users of these classes attempt to move them.

Chocobo1:
> it looks like openssl allocates the state on the heap now, hence the need to delete it. this means you need to implement the assignment and copy constructor as well, for `hasher` and `hasher512`.
> 
> These types should probably be movable as well, but I'm not sure any users of these classes attempt to move them.

OK I addressed those in #6455 and it targets master branch now.

PR #6455, opened 2021-09-09
Migrate to OpenSSL EVP interface for hashing

Chocobo1:
The low-level interface is deprecated since OpenSSL v3.0 [1].

[1] https://www.openssl.org/docs/manmaster/man7/migration_guide.html "Deprecated low-level digest functions"

### DIFF
--- a/src/ed25519/hasher512.cpp
+++ b/src/ed25519/hasher512.cpp
@@ -32,7 +32,10 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "libtorrent/aux_/hasher512.hpp"
 #include "libtorrent/assert.hpp"
-#include "libtorrent/aux_/ssl.hpp"
+
+#if defined TORRENT_USE_LIBCRYPTO
+#include <openssl/opensslv.h> // for OPENSSL_VERSION_NUMBER
+#endif
 
 namespace libtorrent::aux {
 
@@ -45,7 +48,12 @@ namespace libtorrent::aux {
 #elif TORRENT_USE_CNG
 #elif TORRENT_USE_CRYPTOAPI_SHA_512
 #elif defined TORRENT_USE_LIBCRYPTO
-		SHA512_Init(&m_context);
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+		m_context = EVP_MD_CTX_new();
+#else
+		m_context = EVP_MD_CTX_create();
+#endif
+		EVP_DigestInit_ex(m_context, EVP_sha512(), nullptr);
 #else
 		SHA512_init(&m_context);
 #endif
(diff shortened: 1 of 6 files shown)


src/ed25519/hasher512.cpp
```diff
@@ -127,19 +172,23 @@ namespace libtorrent::aux {
...
 	{
+#if defined TORRENT_USE_LIBGCRYPT
 		gcry_md_close(m_context);
-	}
+#elif defined TORRENT_USE_LIBCRYPTO
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+		EVP_MD_CTX_free(m_context);
 #else
-	hasher512::~hasher512() = default;
+		EVP_MD_CTX_destroy(m_context);
```
arvidn, line 189: the documentation says:

> EVP_MD_CTX_destroy() cleans up digest context ctx and frees up the space allocated to it, it should be called only on a context created using EVP_MD_CTX_create().

But you may end up calling it on `nullptr` here. I think there needs to be a check, like:

```
if (m_context) EVP_MD_CTX_destroy(m_context);
```

arvidn: I don't see the `_free()` version being documented, but might have the same restriction

arvidn (commented):
I think some of these comment apply to all 3 versions

src/ed25519/hasher512.cpp
Chocobo1: They do guard against it:
https://github.com/openssl/openssl/blob/d270a6c9ea9f240f653a98c7eb0a5c2134a5d63a/crypto/evp/digest.c#L118-L121

nonetheless I'll still add the check.

src/hasher.cpp
```diff
@@ -58,13 +67,50 @@ TORRENT_CRYPTO_NAMESPACE
...
+	}
+
+	hasher& hasher::operator=(hasher const& h) &
+	{
+		if (this == &h) return *this;
+
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+		m_context = EVP_MD_CTX_new();
+#else
+		m_context = EVP_MD_CTX_create();
```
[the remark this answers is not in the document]
Chocobo1: Right, i got mixed up with something else. Fixed.

src/ed25519/hasher512.cpp
```diff
@@ -70,11 +78,40 @@ namespace libtorrent::aux {
 		gcry_md_copy(&m_context, h.m_context);
 		return *this;
 	}
+#elif defined TORRENT_USE_LIBCRYPTO
+	hasher512::hasher512(hasher512 const& h)
+		: hasher512()
+	{
+		EVP_MD_CTX_copy_ex(m_context, h.m_context);
```
arvidn, line 85: this is a bit of a nit-pick, but it seems `EVP_MD_CTX_copy()` is a better match here, since it's both initializes *and* copies the context. then you don't need to also invoke the default constructor.

arvidn (approved):

src/ed25519/hasher512.cpp
Chocobo1: I had the same thought until I noticed it is viewed as obsoleted:

https://www.openssl.org/docs/manmaster/man3/EVP_MD_CTX_copy.html
>The functions EVP_DigestInit(), EVP_DigestFinal() and EVP_MD_CTX_copy() are obsolete but are retained to maintain compatibility with existing code. New applications should use EVP_DigestInit_ex(), EVP_DigestFinal_ex() and EVP_MD_CTX_copy_ex() because they can efficiently reuse a digest context instead of initializing and cleaning it up on each call and allow non default implementations of digests to be specified.

PR #6459, opened 2021-09-11
don't open the piece in the piece picker if there still are outstanding hash jobs

arvidn:
to handle write failures

### DIFF
--- a/src/piece_picker.cpp
+++ b/src/piece_picker.cpp
@@ -2977,7 +2977,7 @@ get_out:
 		{
 			new_state = piece_pos::piece_zero_prio;
 		}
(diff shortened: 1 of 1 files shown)


elgatito:
@arvidn is this relevant to 1_2?

arvidn:
I don't know for sure. This came up in tests that I'm working on in 2.0 when a write to disk fails and the piece needs to be "reset". i.e. the piece picker need to record that it may be available to be picked again, but if there are outstanding disk operations against it, it can't be opened immediately, but only after those disk operations complete.

so, the short answer is that I have a test exercising this in 2.0, but not in 1.2. (but not pushed yet)

PR #6471, opened 2021-09-19
buffer.hpp: Fix build on NetBSD and OpenBSD

brad0:
OpenBSD, NetBSD and DragonFly have had the same patches in our ports / pkgsrc trees to fix the build on each respective OS.

### DIFF
--- a/include/libtorrent/buffer.hpp
+++ b/include/libtorrent/buffer.hpp
@@ -49,8 +49,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <malloc.h>
 #elif defined __FreeBSD__
 #include <malloc_np.h>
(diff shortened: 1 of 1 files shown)


arvidn:
I don't know where I got the impression from that `malloc_size()` was a thing on OpenBSD and NetBSD. Perhaps just assumed because it's in OSX.

Is there no counterpart for those flavors of BSD?

arvidn (approved):

arvidn:
I'm tempted to go with something like this instead: https://github.com/arvidn/libtorrent/pull/6472

brad0:
> I'm tempted to go with something like this instead: #6472

Ah, yes. That does look more appropriate. macOS does have malloc_size().

PR #6473, opened 2021-09-21
Allow windows openssl paths to default on CI

AllSeeingEyeTolledEweSew:


AllSeeingEyeTolledEweSew:
@arvidn I'm trying to understand your merge process -- if a change ought to be pushed to all branches (`RC_1_2`, `RC_2_0` and `master`), is it best to stick it in `RC_1_2` first, then merge into "newer" branches?

arvidn:
Yes, that's how I normally does it. One exception is if the fix has to be significantly different in older branches than newer ones.

AllSeeingEyeTolledEweSew:
OK. I'll rebase this, and some of my other changes, on `RC_1_2`.

AllSeeingEyeTolledEweSew:
never mind; we use `crypto=openssl` in `RC_2_0` but not `RC_1_2`, so we ought to test our default paths work in `RC_2_0`.

This PR should stay intact. I rebased on `RC_1_2` in #6480.

I removed a mistaken debug option in this PR though.

AllSeeingEyeTolledEweSew:
Trying to get the hang of the merge process: I guess if a change ought to exist on `RC_1_2` but be tweaked in `RC_2_0`, the tweaks should be in a separate commit for cleaner merges.

This is a tweak for #6480 on `RC_2_0` and later.

PR #6476, opened 2021-09-26
Remove msvc toolset pinning

AllSeeingEyeTolledEweSew:
The current logic breaks with boost 1.76 on windows github actions runners. The problem appears to be that `msvc-14.2` is the "correct" choice for the environment and boost-build scripts, but `get_msvc_toolset()` returns `msvc-14.1`, which fails with some error that `cl.exe` can't be found.

Allowing boost-build to autoselect the toolset appears to work better than `get_msvc_toolset()`. Ideally we want logic that selects something like `>=msvc-14.1`, but I don't see a way to do that.

The older python versions don't matter, since we only support python 3.6+.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -18,19 +18,6 @@ import setuptools
 import setuptools.command.build_ext as _build_ext_lib
 
 
-def get_msvc_toolset():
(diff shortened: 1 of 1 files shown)


AllSeeingEyeTolledEweSew:
#6482

PR #6477, opened 2021-09-26
Change default to boost-link=static

AllSeeingEyeTolledEweSew:
This is in line with the principle that `setup.py` should produce "pypi wheel-ready" builds by default.

This is a prerequisite for #6188.


AllSeeingEyeTolledEweSew:
#6486

PR #6479, opened 2021-09-26
Change python package name to libtorrent

AllSeeingEyeTolledEweSew:
Per discussions in #5333 and #6188.


AllSeeingEyeTolledEweSew:
#6485

PR #6481, opened 2021-09-26
Add macOS multi-arch awareness to setup.py

AllSeeingEyeTolledEweSew:
macOS uses multi-architecture binaries, and some architectures can run others in emulation (I believe M1 arm macs can run x86_64 binaries). This change makes `setup.py` attempt to build for the architecture (or arch combination) of the running python binary, rather than the architecture of the system.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -4,6 +4,7 @@ from distutils import log
 import distutils.debug
 import distutils.sysconfig
 import multiprocessing
+import distutils.util
 import os
 import pathlib
 import platform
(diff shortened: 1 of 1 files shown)


arvidn:
does this have implications for cross compiling? Or perhaps that isn't a thing for `setup.py`

arvidn:
I suppose `setup.py` is already pretty committed to building for the specific version of python that's invoking it though.

AllSeeingEyeTolledEweSew:
This **is** cross-compiling. It's not **generic** cross-compiling; we're just keeping the goal that `setup.py` builds for the invoking python. The invoking python on mac might be a different arch than the machine's native arch, or might be a universal binary, so we may need cross-compilation to match it.

arvidn:
sure. does `setup.py` / distutils normally not do (normal) cross compilation, where you produce binaries for a different architecture than what you're running on?

AllSeeingEyeTolledEweSew:
`distutils` has only narrow support for cross-compiling: https://docs.python.org/3/distutils/builtdist.html#cross-compiling-on-windows

Otherwise, distutils generally attempts to build for "the platform of the running python". I can't find documentation for what platforms are supposed to work under what conditions, but given my experience and perusing the `distutils` source, there seems to be a general rule that if python for platform A can **run on** platform B, then `distutils` will cross-compile from B to A in that environment.

In practice, `cibuildwheel` relies on cross-compiling like the above. It produces builds by running 32-bit python on 64-bit windows, and `universal2` python on x64 macos. It must rely on this, because those are the only CI runner architectures available. (I believe all linux builds are produced in native environments though, via docker and qemu).

For libtorrent's `setup.py`, I've just tried to follow `distutils`' apparent principle of "build for the running python, cross-compiling if necessary".

arvidn (approved):

PR #6482, opened 2021-09-26
Remove msvc toolset pinning

AllSeeingEyeTolledEweSew:
The current logic breaks with boost 1.76 on windows github actions runners. The problem appears to be that `msvc-14.2` is the "correct" choice for the environment and boost-build scripts, but `get_msvc_toolset()` returns `msvc-14.1`, which fails with some error that `cl.exe` can't be found.

Allowing boost-build to autoselect the toolset appears to work better than `get_msvc_toolset()`. Ideally we want logic that selects something like `>=msvc-14.1`, but I don't see a way to do that.

The older python versions don't matter, since we only support python 3.6+.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -20,19 +20,6 @@ import setuptools
 import setuptools.command.build_ext as _build_ext_lib
 
 
-def get_msvc_toolset():
-    # Reference: https://wiki.python.org/moin/WindowsCompilers
-    major_minor = sys.version_info[0:2]
-    if major_minor in ((2, 6), (2, 7), (3, 0), (3, 1), (3, 2)):
-        return "msvc-9.0"
-    if major_minor in ((3, 3), (3, 4)):
-        return "msvc-10.0"
-    if major_minor in ((3, 5), (3, 6)):
-        return "msvc-14.1"  # libtorrent requires VS 2017 or newer
-    # unknown python version
-    return "msvc"
-
-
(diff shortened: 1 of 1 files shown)


arvidn (approved):

arvidn:
makes sense. msvc maintains ABI compatibility these days anyways

AllSeeingEyeTolledEweSew:
I do wish we could have a rule that says "use `msvc-14.1` if it works, otherwise autodetect"

My understanding is that [if you use a newer msvc, your build **might** depend on a newer runtime](https://docs.microsoft.com/en-us/cpp/c-runtime-library/crt-library-features?view=msvc-160#c-standard-library-lib-files). I **hope** it's the case that you only depend on the 2019 runtime if you use 2019-specific features, but I can't find documentation that guarantees this. I **fear** our build having an "unnecessary" dependency on the 2019 runtime, which might've been avoided by building with 2017 / 14.1.

[Windows python 3.7 - 3.10 are built with 2017](https://pythondev.readthedocs.io/windows.html#python-and-visual-studio-version-matrix), so ships with that runtime. But when a module needs a newer runtime than the one that ships with python, it looks like [the only answer is to tell users to install it manually](https://github.com/tensorflow/tensorflow/issues/18116), which is awkward.

Aside: I believe that once we cut a release from `master`, since it requires 2019, it will mean that **all** windows python users will need to manually install the 2019 runtime.

PR #6484, opened 2021-09-26
Use ccache on github actions

AllSeeingEyeTolledEweSew:
@Chocobo1 created a github action to set up ccache. I've applied it to github actions.

When it gets a cache hit, it appears to shave minutes off of CI runtime.

We could probably fine-tune cache keys, and maybe we should set eviction config, but I think this is a good start.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -31,6 +31,12 @@ jobs:
       run: |
         sudo apt update
 
+    - uses: Chocobo1/setup-ccache-action@v1
+      # the ccache version on ubuntu-18.04 is too old
+      if: matrix.os != 'ubuntu-18.04'
+      with:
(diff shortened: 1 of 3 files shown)


.github/workflows/python.yml
```diff
@@ -31,6 +31,12 @@ jobs:
       run: |
         sudo apt update
 
+    - uses: Chocobo1/setup-ccache-action@v1
```
arvidn, line 34: what happens on windows? ccache doesn't support windows, right?

arvidn (approved):

.github/workflows/python.yml
AllSeeingEyeTolledEweSew: Right, it's a no-op.

PR #6485, opened 2021-09-26
Change python package name to libtorrent

AllSeeingEyeTolledEweSew:
Per discussions in #5333 and #6188.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -448,7 +448,7 @@ class LibtorrentBuildExt(BuildExtBase):
 
 
 setuptools.setup(
-    name="python-libtorrent",
+    name="libtorrent",
     version="1.2.14",
     author="Arvid Norberg",
     author_email="arvid@libtorrent.org",


arvidn:
can you really have two different targets with the same name in cmake? I don't quite understand how or why this works. I would have thought the python binding target would clash with the libtorrent library target name. How are they kept separate?

AllSeeingEyeTolledEweSew:
It might not be valid. I just did find/replace and assumed it was valid since the tests pass, but maybe it's a false positive. My knowledge of cmake is minimal.

arvidn:
@zeule or @FranciscoPombal is it the case that `Python3_add_library()` uses a different namespace than `add_library()` for the target name?

AllSeeingEyeTolledEweSew:
Superseding question: do I even need to change the target names? I guess the artifact name of the extension was `libtorrent.so` already, and maybe the cmake system *only* builds artifacts and has no integration with packaging.

arvidn:
> Superseding question: do I even need to change the target names?

I don't know :) Did you see the target file on disk change name when you change the name of the target?

zeule:
I don't understand what it all is about, sorry. 

1. There are no namespaces in CMake.
2. The python target sets [OUTPUT_NAME](https://cmake.org/cmake/help/latest/prop_tgt/OUTPUT_NAME.html) to 'libtorrent' already.
3. The libtorrent target name is [torrent-rasterbar](https://github.com/arvidn/libtorrent/blob/RC_2_0/CMakeLists.txt#L544).

bindings/python/setup.py
```diff
@@ -448,7 +448,7 @@ def _configure_b2_with_distutils(self, python_binding_dir):
 
 
 setuptools.setup(
-    name="python-libtorrent",
```
arvidn, line 451: @AllSeeingEyeTolledEweSew I think this may be the only important part of this patch then

arvidn:
@AllSeeingEyeTolledEweSew I'm having a hard time finding (good) documentation for `setuptools.setup()`. Specifically, what the `name` parameter means.

Is this the name that users need to say in the `import` statements?
Is this a backwards compatible change for programs using the python module?

AllSeeingEyeTolledEweSew:
TL;DR: It doesn't change the import name in `import libtorrent`. It changes the distribution name. This is technically not backward compatible, but I believe that it won't break anybody, as I cannot find any way someone could meaningfully depend on the name today.

Some terms:
* **module**: a python file like `module.py`, meant to be imported with `import module`
* **extension**: a shared library of native code that can be imported like a module
* **package**: a directory of modules, which works like a namespace. If there's `a/b.py`, then `a` is a package containing module `a.b`, allowing `import a.b` or `from a import b`.
* **distribution**: a collection of modules / extensions / packages, to be versioned and distributed as a unit.

Distributions are used by `pip` to manage the modules, extensions and packages installed in a python environment, in the `site-packages` directory. A distribution's metadata gets put in `site-packages/*.dist-info`.

The import system doesn't know about distributions.

They are distributed as `*.whl`, `*.tar.gz` or (deprecated) `*.egg` archives. pypi is a big index of these files.

A distribution can contain any number of modules / extensions / packages, and there are no constraints on the names. A distribution named `foo` can contain a module `bar` and package `baz.bat`. The most common case is that a distribution `foo` contains exactly one module, extension or package named `foo`.

A distribution includes a version and dependencies as part of its metadata. Dependencies refer to other distribution names and versions, not module / extension / package names.

In the example of `setuptools.setup(name="foo")`, `foo` is the name of the distribution. This would cause the name to show up in the following places:
* wheels will be named like `foo-1.2.3-cp36-abi3-win32.whl`
* the pypi project will be hosted at `https://pypi.org/project/foo`
* other python distributions will declare dependencies like `foo>=1.2.3`

Linux distro package managers respect python packaging by convention only. They don't programatically depend on python packaging metadata. Distro package names are usually different than python distribution names; e.g. `requests` is `py-requests` or `py3-requests` on alpine. Distro package builder systems appear to always list dependencies manually, rather than parsing metadata programatically.

On alpine, the package `deluge` depends on `py3-libtorrent-rasterbar`. If you install these, they do install python distribution metadata globally to `/usr/lib/pythonX.Y/site-packages/*.dist-info`. This metadata is vestigial, and not used by anything. deluge's dependencies in this metadata do not include libtorrent. The dependencies are all controlled by the alpine package system, not the python package system.

This vestigial metadata is the *only* place I can find where the existing `python-libtorrent` name occurs in the wild. I believe pypi and `pip` is the only (open) python packaging system which directly respects python distribution metadata. And today, the `python-libtorrent` project on pypi is empty, not even containing any out-of-date builds.

Therefore, I believe it's safe to change the distribution name to `libtorrent`.

PR #6486, opened 2021-09-26
Default to libtorrent-link=static boost-link=static

AllSeeingEyeTolledEweSew:
This is in line with the principle that `setup.py` should produce "pypi wheel-ready" builds by default.

This is a prerequisite for #6188.


arvidn (approved):

AllSeeingEyeTolledEweSew:
Updated CI to not set `libtorrent-link=static`

PR #6487, opened 2021-09-26
All seeing eye tolled ewe sew dep/openssl windows/rc 1 2

arvidn:


AllSeeingEyeTolledEweSew:
could also set `ssl_lib`/`ssl_include` matrix parameters like `RC_2_0` does today

arvidn:
yeah, it's probably better to mirror how this is done in `RC_2_0`.

Issue #6448, opened 2021-09-05
Decide on a policy for openssl linkage in pypi wheels


AllSeeingEyeTolledEweSew:
@arvidn @jkterry1 @cas-- @mayli @filips123

Let's spin off the discussion about openssl linkage.

I'll try to summarize the ideas, and any conclusions, in the top comment here.

## Impact of openssl bugs on libtorrent
Assuming something like "heartbleed 2.0" happens:
* Users of SSL torrents would be at risk of attack by peers
  * These seem rare
* Users could be at risk when connecting to a malicious HTTPS tracker

## Risks of statically linking openssl
* **Developer maintenance**: Arvid may be on the hook to rapidly integrate new versions of openssl for critical bug fixes, and publish new wheels
  * **Mitigation**: we can roll our own dependabot for our openssl dependency. [manylinux does this](https://github.com/pypa/manylinux/blob/c009bd6f75c5a69fe1962d7a7b297d5351a898a6/.github/workflows/update-dependencies.yml). The outcome is a bot makes a PR against libtorrent when openssl needs to be updated; Arvid just needs to accept it and run the cibuildwheel workflow to publish it.
* **Disharmony**: There could be differences between libtorrent's openssl and other ssl contexts, even in the same app, yet both must be configured by the same global certificate store and environment variables
* **User surprise**: users may expect that updating their system openssl will be effective for all SSL connections on their machine
* **Development lock-in**: static linking will satisfy basic use cases in all environments. If we ever change to dynamic linking we will definitely break some users whose installations previously worked.

## Risks of dynamically linking openssl
* **Breakage**: Users with the wrong openssl version can't load libtorrent, because we can only link against one of a few ABI-incompatible versions
  * Impact: 1.1 is popularly installed; 1.0 reached EOL 18 months ago but is probably installed widely still
  * Impact gets worse once openssl 3.0 is more widely installed
  * Also impacts python distributions which link openssl statically (at least the build pythons found on `manylinux` centos images do this)
* **User surprise**: users may expect that updating their python packages is sufficient to fix security issues in those packages

## Risks of using `dlopen()` for openssl
* **Development risk**: I've only seen this method discussed in blog posts, not done in practice. There are many unknowns.
  * **High investment**: I expect this would require some significant work to research.
  * **Bad signs**: openssl has an infamously unstable API and ABI. Reading their blog posts, they expect you to recompile when they release a new version.

## Conclusions

* We should vendor openssl with static linking.
[...]

PR #6493, opened 2021-09-27
Default python bindings to crypto=openssl

AllSeeingEyeTolledEweSew:
Per discussion in #6188 and #6448.

Draft until there's consensus that this is a good plan.

The `test.py` changes backport a lot of infrastructure I added in `master`. I'll make a separate PR for `master` for a cleaner merge.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -40,7 +40,7 @@ jobs:
     - name: dependencies (linux)
       if: runner.os == 'Linux'
       run: |
-        sudo apt install libboost-tools-dev libboost-python-dev libboost-dev libboost-system-dev python3 python3-setuptools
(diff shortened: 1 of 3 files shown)


arvidn:
I've been thinking that it would probably make sense to set `crypto=openssl` as the default for the main target as well.

AllSeeingEyeTolledEweSew:
After the investigations I did in #6448, I would think Apple's crypto APIs would be preferable to openssl. I haven't looked into it, but in theory it should be one less dependency to install, for both build and runtime.

AllSeeingEyeTolledEweSew:
@arvidn do you have an opinion on shipping libtorrent with openssl vendored? Would you mind responding in #6448?

AllSeeingEyeTolledEweSew:
As #6448 has your approval, I marked this ready for review

PR #6499, opened 2021-09-28
Reference major version for actions on Github CI

Chocobo1:
* Reference major version for actions on Github CI
  https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsuses
  > Using the specific major action version allows you to receive critical fixes and security patches while still maintaining compatibility. It also assures that your workflow should still work.
* Remove ccache action conditional
  The latest release added support to Ubuntu-18.04:
  https://github.com/Chocobo1/setup-ccache-action/releases/tag/v1.1.1
* Use latest OS for CI jobs that are agnostic to the underlying platform 

I would advise to apply them to other branches (RC_2_0, master) as well.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-20.04, ubuntu-18.04, macos-latest, windows-2019 ]
+        os: [ubuntu-20.04, ubuntu-18.04, macos-latest, windows-latest ]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -32,8 +32,7 @@ jobs:
         sudo apt update
 
     - uses: Chocobo1/setup-ccache-action@v1
-      # the ccache version on ubuntu-18.04 is too old
-      if: matrix.os != 'ubuntu-18.04'
+      if: runner.os != 'Windows'
       with:
         update_packager_index: false
 
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -18,7 +18,7 @@ jobs:
 
       steps:
       - name: checkout
-        uses: actions/checkout@v2.3.3
+        uses: actions/checkout@v2
         with:
            submodules: recursive
 
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -13,7 +13,7 @@ jobs:
 
    tests:
       name: Test
-      runs-on: windows-2019
+      runs-on: windows-latest
       continue-on-error: true
 
       steps:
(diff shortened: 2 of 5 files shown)


.github/workflows/python.yml
```diff
@@ -32,8 +32,7 @@ jobs:
         sudo apt update
 
     - uses: Chocobo1/setup-ccache-action@v1
-      # the ccache version on ubuntu-18.04 is too old
-      if: matrix.os != 'ubuntu-18.04'
+      if: runner.os != 'Windows'
```
arvidn, line 35: I was under the impression that your ccache action would just be a no-op on windows, and that it's not necessary to have a condition to only run it on non-windows.

I would have expected this condition to just be removed instead.

.github/workflows/windows.yml
```diff
@@ -13,12 +13,12 @@ jobs:
 
    tests:
       name: Test
-      runs-on: windows-2019
+      runs-on: windows-latest
```
arvidn, line 16: in my mind it's preferable to pin a specific version of the machine image, to make the CI jobs deterministic. Using the "latest" the CI jobs can start failing without reflecting a change to the code itself, just that the machine image was updated. Is there a good reason to use "latest" instead of the specific version that currently is the latest one?

arvidn:
I will merge this into `RC_2_0` and `master` too, once it lands.

.github/workflows/python.yml
AllSeeingEyeTolledEweSew: the action outputs a warning about being a no-op on windows. This gets surfaced in the "PR checks" UI like an error on the workflow. See https://github.com/Chocobo1/setup-ccache-action/issues/2

Chocobo1: > I was under the impression that your ccache action would just be a no-op on windows, and that it's not necessary to have a condition to only run it on non-windows.
> 
> I would have expected this condition to just be removed instead.

Personally I would like to have a conditional, but ultimately it is up to you whether you are ok with a warning...

.github/workflows/windows.yml
Chocobo1: I was being conservative with this commit (see commit message).

>Is there a good reason to use "latest" instead of the specific version that currently is the latest one?

Less maintenance in case when the OS is announced obsolete from github actions.
And another viewpoint is: for those using `*-latest`, it shows the OS is not a requirement (nor related) for the CI jobs; and for those pinned to a specific version, OS version is tied to the jobs/steps in some way.

AllSeeingEyeTolledEweSew: FWIW: I noticed that currently using `macos-latest` yields a warning because it will be changed to `macos-11` soon. And presumably you'd also get a warning if you use an os that would be removed soon. I guess the key is to stay on top of warnings in either case

PR #6502, opened 2021-09-28
Use job-specific ccaches

AllSeeingEyeTolledEweSew:
The ccache action prints stats at the end of the job. I notice that our cache hit rates are low, even when re-running a workflow without changes. I believe that matrix'ed jobs are stomping on each other's caches:
* multiple jobs start at the same time
* they load the same cache
* they modify the cache, each differently
* they finish at different times, and store their updated caches
* on the next workflow run, all the jobs only see the cache of the *last job to finish* from the previous workflow

I gave each ccache a separate key. This resulted in much higher hit rates, until we spill out of the cache within a single job.

I don't believe this strategy results in cache data being duplicated between isolated job caches, because that would mean two different jobs were are doing identical builds, and workflow jobs are generally written to avoid this.

I also separated caches by base ref (i.e. target branch). There's some chance that builds are partially identical across branches, but it should still be fairly low.

The default cache size of 500mb is too small for some jobs, especially the linux dist jobs. To pick cache sizes, I arbitrarily decided that our target workload is 5 active pull requests on a given branch. I ran the pull request workflows with cold caches, measured the cache sizes when each job was done, multiplied by 5 and rounded up a bit.

Note that caches are compressed ~75% when stored, so ccache with `max_size=400M` will only use about 100mb of `actions/cache` storage for libtorrent.

You can test cache hit rates by re-running checks after they've completed the first time.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -30,6 +30,9 @@ jobs:
       - uses: Chocobo1/setup-ccache-action@v1
         with:
           update_packager_index: false
+          override_cache_key: ccache-linux-build-${{ github.base_ref }}-${{ matrix.config }}
+          ccache_options: |
+            max_size=5G
 
       - name: install boost
         run: |
@@ -74,6 +77,9 @@ jobs:
       - uses: Chocobo1/setup-ccache-action@v1
         with:
           update_packager_index: false
+          override_cache_key: ccache-linux-fuzzers-${{ github.base_ref }}
+          ccache_options: |
+            max_size=500M
 
       - name: install clang-9
         continue-on-error: true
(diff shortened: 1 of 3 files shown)


arvidn:
> I also separated caches by base ref (i.e. target branch). There's some chance that builds are partially identical across branches, but it should still be fairly low.

So, this is the branch a PR is being merged into, right? so multiple PRs against `master` will all share cache.

arvidn (approved):

AllSeeingEyeTolledEweSew:
> So, this is the branch a PR is being merged into, right? so multiple PRs against `master` will all share cache.

That's right.

PR #6503, opened 2021-09-29
Test merge 2

arvidn:


AllSeeingEyeTolledEweSew:
relevant: https://github.com/actions/virtual-environments/pull/4154

it looks like `brew install openssl` now installs openssl 3.0, FYI

the macos runner base image still does `brew install openssl@1.1`, but I guess that puts files in some versioned dir that isn't picked up by our build scripts

arvidn:
yeah, I noticed that. I think my approach right now, to suppress the openssl deprecation warnings, is a reasonable way forward for 2.0.x

PR #6510, opened 2021-10-03
Backport d6b9d03a9 (remove since-long deprecated wstring overloads)

AllSeeingEyeTolledEweSew:
This has the same intent as #6509: to backport changes from `master`.

I thought this change deserved discussion on its own.

In python 3, the default boost.python converters for `std::string` and `std::wstring` both match `str` and `bytes`. So the `std::wstring` bindings here are always shadowed by the `std::string` functions in python 3.

In python 2, the `std::string` converter matches `str` but not `unicode`. So the `std::wstring` binding is functional there, and required to use `unicode` args.

Removing this binding could break users, but they would need to be
* using python 2 (EOL 2020-01-01)
* using `unicode` args to these functions (`unicode` won't work most places in the API, since it only matches `std::wstring` functions in python2)
* either Windows-only, or else using code like
```python
if sys.platform == "win32":
    handle.rename_file(0, u"file.txt")
else:
    handle.rename_file(0, "file.txt")
```

The only thing to gain from this PR is cleaner merges from `RC_1_2` to `master`. But the only thing to lose is the narrow case above.

Personally I think smoother development is the path to good code quality, so I favor clean merges.

### DIFF
--- a/bindings/python/src/create_torrent.cpp
+++ b/bindings/python/src/create_torrent.cpp
@@ -105,14 +105,6 @@ namespace
 
     FileIter end_files(file_storage const& self)
     { return FileIter(self, self.end_file()); }
-
-#ifdef TORRENT_WINDOWS
-    void add_file_wstring(file_storage& fs, std::wstring const& file, std::int64_t size
-       , file_flags_t const flags, std::time_t md, std::string link)
-    {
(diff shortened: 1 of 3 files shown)


arvidn:
oh, I see. This is just affecting the python binding. That makes it more reasonable I think.

PR #6511, opened 2021-10-03
Backport 2f9ed031d (disable deprecated boost.asio APIs)

AllSeeingEyeTolledEweSew:
This has the same intent as #6509: to backport changes from `master`.

I don't know if this change is appropriate. It breaks on `ubuntu-18.04`.

### DIFF
--- a/bindings/python/src/converters.cpp
+++ b/bindings/python/src/converters.cpp
@@ -58,7 +58,7 @@ struct tuple_to_endpoint
         extract<std::uint16_t> port(object(borrowed(PyTuple_GetItem(x, 1))));
         if (!port.check()) return nullptr;
         lt::error_code ec;
-        lt::address::from_string(ip, ec);
+        lt::make_address(ip, ec);
         if (ec) return nullptr;
         return x;
     }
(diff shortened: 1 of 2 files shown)


userdocs:
@AllSeeingEyeTolledEweSew Ubuntu bionic (18.04) defaults to gcc 7.5 but defualt repos actually include [gcc 8.4](https://packages.ubuntu.com/bionic/gcc-8)

I would suggest you try that to see if there was a compiler related issue as using gcc-8 i have no issues building libttorrrent

https://github.com/userdocs/libtorrent-crossbuild/runs/4130353848?check_suite_focus=true

arvidn:
@AllSeeingEyeTolledEweSew sorry, somehow I missed your message. I think the problem is that the `Extract<std::string>` type is a proxy that defers the actual conversion until it's *used* as a string. The function signature, on the other hand, expects to convert *from* a `std::string` (or something similar) into a `string_view`, but there is no `std::string` anywhere.

I think if you add an explicit cast to `std::string`, you'll force it to happen, and work. something like this:

```
lt::make_address(static_cast<std::string>(ip), ec);
```

But if the current code works in master without the cast, maybe it was solved in some other way there.

AllSeeingEyeTolledEweSew:
How's this?

arvidn (approved):

PR #6495, opened 2021-09-27
cmake: install pdb files

jagannatharjun:

PR #6515, opened 2021-10-05
cmake: fix PDB installing

jagannatharjun:
fixes https://github.com/arvidn/libtorrent/pull/6495#issuecomment-932743810


CMakeLists.txt
```diff
@@ -931,9 +931,18 @@ install(
 )
 
 if (MSVC)
+    set_target_properties(
```
Chocobo1, line 934: Minor comment, the file used tabs and here are spaces.

Chocobo1 (approved):
It works for me, no more error messages. Note that it leaves an empty bin folder at installation site.

jagannatharjun:
> Note that it leaves an empty bin folder at installation site.

should be fixed

PR #6523, opened 2021-10-11
use ifaddrs for android api >= 24, implement enum_routes with operation_not_supported

aldenml:
Second PR for Android 11 fix. There is one more remaining change that I think could be a little contentious.


aldenml:
Hi @arvidn, I'm waiting to have this PR merged before moving to the next one. Do you prefer to have everything in this PR?

PR #6528, opened 2021-10-18
fix issue with stalled pieces on disk full errors

arvidn:


an0n666:
Thank you! This fixes the issue on RC1_2.

PR #6529, opened 2021-10-18
explicitly FlushViewOfFile() on windows when closing a file

arvidn:
This is to address the issue of windows not pre-emptively flushing dirty pages to disk in memory mapped files.

### DIFF
--- a/src/mmap_disk_io.cpp
+++ b/src/mmap_disk_io.cpp
@@ -314,10 +314,6 @@ private:
 
 	settings_interface const& m_settings;
 
-	// we call close_oldest_file on the file_pool regularly. This is the next
-	// time we should call it
-	time_point m_next_close_oldest_file = min_time();
-
 	// LRU cache of open files
 	aux::file_view_pool m_file_pool;
 
@@ -1510,6 +1506,14 @@ TORRENT_EXPORT std::unique_ptr<disk_interface> mmap_disk_io_constructor(
 		++m_num_running_threads;
 		m_stats_counters.inc_stats_counter(counters::num_running_threads, 1);
 
+		// we call close_oldest_file on the file_pool regularly. This is the next
+		// time we should call it
+		time_point next_close_oldest_file = min_time();
+
+#if TORRENT_HAVE_MAP_VIEW_OF_FILE
+		time_point next_flush_file = min_time();
+#endif
+
 		for (;;)
 		{
 			aux::disk_io_job* j = nullptr;
@@ -1538,20 +1542,30 @@ TORRENT_EXPORT std::unique_ptr<disk_interface> mmap_disk_io_constructor(
 					}
 				}
 
-				if (now > m_next_close_oldest_file)
+				if (now > next_close_oldest_file)
 				{
 					seconds const interval(m_settings.get_int(settings_pack::close_file_interval));
 					if (interval <= seconds(0))
 					{
 						// check again in one minute, in case the setting changed
-						m_next_close_oldest_file = now + minutes(1);
+						next_close_oldest_file = now + minutes(1);
 					}
 					else
 					{
-						m_next_close_oldest_file = now + interval;
+						next_close_oldest_file = now + interval;
 						m_file_pool.close_oldest();
 					}
 				}
+
+#if TORRENT_HAVE_MAP_VIEW_OF_FILE
+				if (now > next_flush_file)
+				{
+					// on windows we need to explicitly ask the operating system to flush
+					// dirty pages from time to time
+					m_file_pool.flush_next_file();
+					next_flush_file = now + seconds(30);
+				}
+#endif
 			}
 
 			execute_job(j);
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* on windows, explicitly flush memory mapped files periodically
 	* fix build with WolfSSL
 	* fix issue where incoming uTP connections were not accepted over SOCKS5
 	* fix several issues in handling of checking files of v2 torrents, esp. from magnet links
--- a/Jamfile
+++ b/Jamfile
@@ -309,7 +309,7 @@ rule building ( properties * )
 	if <toolset>msvc in $(properties) || <toolset>intel-win in $(properties)
 	{
 		# allow larger .obj files (with more sections)
-		result += <cflags>/bigobj ;
+		result += <cxxflags>/bigobj ;
 	}
 
 	if <toolset>gcc in $(properties) && <target-os>windows in $(properties)
--- a/include/libtorrent/aux_/file_view_pool.hpp
+++ b/include/libtorrent/aux_/file_view_pool.hpp
@@ -110,6 +110,12 @@ namespace aux {
 
 		void close_oldest();
 
+#if TORRENT_HAVE_MAP_VIEW_OF_FILE
+		void flush_next_file();
+		void record_file_write(storage_index_t st, file_index_t file_index
+			, uint64_t pages);
+#endif
+
 	private:
 
 		std::shared_ptr<file_mapping> remove_oldest(std::unique_lock<std::mutex>&);
@@ -140,6 +146,9 @@ namespace aux {
 			file_id key;
 			std::shared_ptr<file_mapping> mapping;
 			time_point last_use{aux::time_now()};
+#if TORRENT_HAVE_MAP_VIEW_OF_FILE
+			std::uint64_t dirty_bytes;
+#endif
 			open_mode_t mode{};
 		};
 
@@ -150,6 +159,10 @@ namespace aux {
 			mi::ordered_unique<mi::member<file_entry, file_id, &file_entry::key>>,
 			// look up files by least recently used
 			mi::sequenced<>
+#if TORRENT_HAVE_MAP_VIEW_OF_FILE
+			// look up files with dirty pages
+			, mi::ordered_non_unique<mi::member<file_entry, std::uint64_t, &file_entry::dirty_bytes>>
+#endif
 			>
 		>;
 
--- a/include/libtorrent/aux_/mmap.hpp
+++ b/include/libtorrent/aux_/mmap.hpp
@@ -122,6 +122,10 @@ namespace aux {
 #endif
 			);
 
+#if TORRENT_HAVE_MAP_VIEW_OF_FILE
+		void flush();
+#endif
+
 		// non-copyable
 		file_mapping(file_mapping const&) = delete;
 		file_mapping& operator=(file_mapping const&) = delete;
--- a/src/file_view_pool.cpp
+++ b/src/file_view_pool.cpp
@@ -269,6 +269,36 @@ namespace libtorrent { namespace aux {
 		std::unique_lock<std::mutex> l(m_mutex);
 		deferred_destruction = remove_oldest(l);
 	}
+
+#if TORRENT_HAVE_MAP_VIEW_OF_FILE
+	void file_view_pool::flush_next_file()
+	{
+		std::shared_ptr<file_mapping> mapping;
+		{
+			std::unique_lock<std::mutex> l(m_mutex);
+			auto& flush_view = m_files.get<2>();
+			if (flush_view.size() == 0) return;
+
+			auto it = std::prev(flush_view.end());
+			if (it->dirty_bytes == 0) return;
+			mapping = it->mapping;
+			flush_view.modify(it, [](file_entry& e) { e.dirty_bytes = 0; });
+		}
+
+		// we invoke flush after we release the mutex
+		mapping->flush();
+	}
+
+	void file_view_pool::record_file_write(storage_index_t const st
+		, file_index_t const file_index, uint64_t const bytes)
+	{
+		std::unique_lock<std::mutex> l(m_mutex);
+		auto& key_view = m_files.get<0>();
+		auto i = key_view.find(file_id{st, file_index});
+		if (i == key_view.end()) return;
+		key_view.modify(i, [bytes](file_entry& e) { e.dirty_bytes += bytes; });
+	}
+#endif
 }
 }
 
--- a/src/mmap.cpp
+++ b/src/mmap.cpp
@@ -638,9 +638,18 @@ file_mapping::file_mapping(file_handle file, open_mode_t const mode
 		throw_ex<storage_error>(error_code(GetLastError(), system_category()), operation_t::file_mmap);
 }
 
+void file_mapping::flush()
+{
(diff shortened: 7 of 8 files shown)


sledgehammer999:
@arvidn any news on this? It is kinda a blocker item for qbittorrent.
IMO, we can't do an official release based on RC_2_0 when this "bug" is happening.

arvidn:
@sledgehammer999 I don't have a very good windows environment to test this on. I think this might work, but I can't say for sure. Would you mind giving it a try?

sledgehammer999:
Do I need to define `TORRENT_HAVE_MAP_VIEW_OF_FILE` somewhere? Your diff doesn't seem to define it.
Moreover, from a fast check I did, RAM usage keeps climbing up. I stopped when it reached 13.5GB out of the 16GB available.

But now the file size on disk seems to be updating. This is strange.

arvidn:
> Do I need to define TORRENT_HAVE_MAP_VIEW_OF_FILE somewhere?

`TORRENT_HAVE_MAP_VIEW_OF_FILE` is defined in [config.hpp](https://github.com/arvidn/libtorrent/blob/RC_2_0/include/libtorrent/config.hpp#L224).

>  from a fast check I did, RAM usage keeps climbing up. I stopped when it reached 13.5GB out of the 16GB available.
> 
> But now the file size on disk seems to be updating. This is strange.

That suggests that the flushing is working. I suppose the non-dirty pages are kept in RAM but would be the first to be evicted under memory pressure.

sledgehammer999:
>I suppose the non-dirty pages are kept in RAM but would be the first to be evicted under memory pressure.

I suppose I can try to let it exhaust the whole RAM and see what happens...

Does the view map the whole file or parts of it(based on the down/up requests)?

sledgehammer999:
>I suppose I can try to let it exhaust the whole RAM and see what happens...

I let it download a ~20GB file. Eventually the RAM was exhausted hovering between 15.8 and 15.9GB usage in the Task Manager tab.
**However,** regular PC usage didn't seem to be impacted. Along with the download I was browsing, opening files, watching youtube, listening to music. During the RAM exhaustion I didn't see any slowdown to the above. Only twice at random the streaming music from youtube did a hiccup/freeze.
I lack the imagination on how to benchmark this under more load of the PC.

Lastly, I am a bit skeptical of the regular user reaction when they see full RAM usage even though in actuality it won't impact them. In the future, you might want to periodically unmap the view so the OS reclaims instantly the memory. The criteria could be huge files and long running operations on them.

arvidn:
There is a feature to periodically close the least recently used file in the file pool. For an actively seeded large file for though, the large file won't likely be the least recently used.

Libtorrent does map the entire files. I was considering mapping smaller views (to support 32 bit machines). But the complexity of that is quite high, only to gain 32 bit support.

arvidn:
There's also this classic :)

https://www.linuxatemyram.com

sledgehammer999:
> I was considering mapping smaller views (to support 32 bit machines). But the complexity of that is quite high, only to gain 32 bit support.

I understand. I want to ask this: Does this mean that RC_2_0 isn't actually suitable for 32bit binaries? Will it fail if it tries to map a big file? Or does it have a fallback mechanism?
This is an important detail to know if I should provide 32bit binaries or not.

>There's also this classic :)

At least I hope that on linux/macos the file maps don't need manual flushing like on Windows to write data to disk. I haven't tested this situation on those platforms.

arvidn:
There is a fallback mechanism for 32 but systems, and systems that don't support memory mapped files. However, it's single-threaded. Also, people are reporting that the fopen-family of functions also fail intermittently on windows.

arvidn:
this is the option to close files periodically. It's supposed to be set to 120 seconds on windows. http://libtorrent.org/reference-Settings.html#close_file_interval

sledgehammer999:
>this is the option to close files periodically. It's supposed to be set to 120 seconds on windows. http://libtorrent.org/reference-Settings.html#close_file_interval

Eureka!
I think you have a bug in your sources. I grepped them and you set that setting to zero unconditionally. See: https://github.com/arvidn/libtorrent/blob/139131ee13ece244891aa4aaa7b82bd8cbca44a1/src/settings_pack.cpp#L359

I set it explicitly to 120 seconds in qbt. Now the RAM usage doesn't take all available RAM.
Moreover, I used it with vanilla RC_2_0 and it does flush to disk after the file is closed. This PR is kinda unneeded. However, I am not positive that a 2 minute* window is good for data to reside in RAM. 

*plus the time it takes the system manager to lazily commit it. Supposedly it commits 1/8 of cached data every second.

sledgehammer999:
> This PR is kinda unneeded. However, I am not positive that a 2 minute* window is good for data to reside in RAM.

Another sideeffect is that it tanks the download speed when closing the file. Does this mean it is a synchronous call?
For example I might be downloading a popular torrent at 9MB/sec which after 2 minutes has about 1GB cached. During the close these data committing to HDD (slow). And for 2-3 seconds the download speed is tanked to 1-3MB/sec.
I don't know if this PR suffers from the same thing because the 30seconds gap doesn't let it accumulate much data.

arvidn:
interesting. the `FlushViewOfFile()` is documented to be asynchronous. So the back-pressure to the network thread is probably happening in the threads faulting in new pages of the memory mapped file. Presumably the speed drops every 30 seconds are shorter though, right?

arvidn:
could you try increasing the number of disk threads, to see if that makes a difference?
If it's some lock contention in the memory manager that causes the slow-downs, it probably won't.

arvidn:
I still think this PR serves a purpose. The flushing of file views in this patch round-robins the files, whereas the file close interval closes the *least recently used* file. so, if you have one file that's more commonly accessed, it may never be closed, but in this patch it will at least be flushed periodically.

arvidn:
in fact, I wonder if the 30 seconds flush interval that's hard coded here should be proportional to the size of the file view pool. The large the pool, the more rare each individual file will be flushed.

sledgehammer999:
>could you try increasing the number of disk threads, to see if that makes a difference?

I've set `aio_threads` to 16 but I still see speed dropdowns during flush. It is random but is is always at least 2MB/sec dropdown from whatever speed I am at.

>Presumably the speed drops every 30 seconds are shorter though, right?

Personally, I don't observe a difference. I am not measuring this via code. Rough estimation of seeing the values fluctuate in the GUI.

>I still think this PR serves a purpose. The flushing of file views in this patch round-robins the files, whereas the file close interval closes the least recently used file. so, if you have one file that's more commonly accessed, it may never be closed, but in this patch it will at least be flushed periodically.

:+1: 


>The large the pool, the more rare each individual file will be flushed.

Do you mean the opposite or did I misunderstand you?

>in fact, I wonder if the 30 seconds flush interval that's hard coded here should be proportional 

Maybe a combination of "data received since last flush" and a hard time limit since last flush?

arvidn:
The way this patch works is that it flushes the next file in the file_view_pool every 30 seconds. If there are a lot of files in the pool, one specific file will be flushed less often than every 30 seconds. The larger the file pool is, the longer it will take to rotate back to the same file again.

glassez:
> The way this patch works is that it flushes the next file in the file_view_pool every 30 seconds. If there are a lot of files in the pool, one specific file will be flushed less often than every 30 seconds. The larger the file pool is, the longer it will take to rotate back to the same file again.

I'm sorry, I haven't followed this topic in great detail, so my question may be a bit silly... What about downloading one large file, the size of which exceeds the amount of RAM? As far as I understand, the main problem is with such torrents, which contain large files, and not just have a total large size. Will such a file be flushed to disk regularly so as not to clog RAM with dirty pages, as is happening now?

arvidn:
Yes. Fundamentally it's windows responsibility to flush dirty pages to disk and evict clean pages from the disk cache. If you leave all this to the OS, I don't think windows will completely lock up or break in some way, but it seems it's not very good at it, hence the "manual" periodic flushing of the file cache.

So, to answer your question; yes it will work and this patch will also ask windows to flush dirty pages periodically, to prevent it from just accruing GBs of them.

arvidn:
That said, I only have windows in a VM, so I don't have a good sense of how the system actually performs. But above is my understanding at least.

glassez:
> The way this patch works is that it flushes the next file in the file_view_pool every 30 seconds.

I just want to make sure that any file will be flushed periodically, even if there are only one file in the pool.

arvidn:
> I just want to make sure that any file will be flushed periodically, even if there are only one file in the pool.

yes. in that scenario the file will be flushed every 30 seconds. If there are two files in the pool, one specific file will be flushed every 60 seconds. If the file isn't in the pool anymore, it has been closed, which also implies a flush.

arvidn:
the worst-case scenario I can think of is having a very large file pool size, say several thousand file handles. A torrent with one very large file and hundreds of tiny files. Most disk activity will happen in the large file, but the small files wil "dilute" the round-robin flushing, possibly making the large file only be flushed very rarely, and the "flushing" of the tiny files ending up being no-ops, because essentially no activity is happening on those.

sledgehammer999:
Would it be costly to just flush every open file every 30secs?

arvidn:
I don't know. in that case it would probably make sense to make this interval configurable.

The way I would implement that would be to make the interval be `30 / file_pool.size()`

glassez:
> Would it be costly to just flush every open file every 30secs?

I don't think any time interval is an appropriate measure for this. It seems to me that it would be better to perform FlushViewOfFile() for file after writing a certain amount of data to it.

src/mmap_disk_io.cpp
```diff
@@ -1538,20 +1542,35 @@ TORRENT_EXPORT std::unique_ptr<disk_interface> mmap_disk_io_constructor(
...
+				{
+					// on windows we need to explicitly ask the operating system to flush
+					// dirty pages from time to time
+					int const pool_size = m_file_pool.flush_next_file();
+
+					// aim at flushing every file every 30 seconds, without
+					// flushing a file more often that every other second.
+					auto const interval = std::max(
+						seconds(30) / std::max(1, pool_size), seconds(2));
+					next_flush_file = now + ;
```
sledgehammer999, line 1571: Typo: You forgot to add `interval`

AllSeeingEyeTolledEweSew:
Has anyone experimented to see if large *reads* on a file mapping cause the same kind of memory management problems as large *writes*? I wouldn't expect problems, but it seems you never know with windows.

IIUC from the docs, `FlushViewOfFile` *should* just flush dirty pages. It looks like the code currently flushes all files. Could it instead only flush known-dirty ones?

FWIW, [this article](https://ayende.com/blog/163298/with-a-little-knowledge-and-a-profiler-let-us-optimize) claims `FlushViewOfFile` is O(N) in the size of the range being referred to, and can be optimized by flushing only known-dirty ranges.

arvidn:
> IIUC from the docs, FlushViewOfFile should just flush dirty pages. It looks like the code currently flushes all files. Could it instead only flush known-dirty ones?

I consider this a future improvement. It's not obvious that it would be more efficient to track every 16 kiB block that has been written since the last flush and flush just those.

arvidn:
but maybe it is a hint that `FlushViewOfFile()` shouldn't be called this frequently. every 2 seconds.

AllSeeingEyeTolledEweSew:
> > IIUC from the docs, FlushViewOfFile should just flush dirty pages. It looks like the code currently flushes all files. Could it instead only flush known-dirty ones?
> 
> I consider this a future improvement. It's not obvious that it would be more efficient to track every 16 kiB block that has been written since the last flush and flush just those.

sorry; I meant: could we flush only known-dirty *files*? We could still flush the whole range of a file; I agree tracking ranges more granularly is complex

glassez:
>Has anyone experimented to see if large reads on a file mapping cause the same kind of memory management problems as large writes? 

What exactly is bothering you about reading? Can it cause the same problems as writing? Isn't the main problem of writing that we are discussing is the accumulation of a large number of dirty pages in RAM, which leads either to performance problems when they all start being written to disk when the file is closed, or to data loss in the event of a sudden system crash (or power loss)?

AllSeeingEyeTolledEweSew:
It itches my brain that Windows appears to implement file mappings as a second layer above their page cache, and that some believe this is done because their page cache has bad performance characteristics.

If those things are true, then maybe they also thought it would be "helpful" to cache reads as well as writes in that second layer.

arvidn:
I'm having second thoughts about this patch. Considering that `FlushViewOfFile()` might traverse all pages of a mapping, to find the dirty ones, it might be extremely inefficient to make these calls unconditionally once a torrent is seeding. We know there are no dirty pages, yet we ask windows to traverse them all.

The reason I've been hesitant to flush only files with dirty pages, or just the dirty parts of those files is the additional complexity of tracking those things. Right now, the concurrency considerations are quite simple. The main file handle container is protected by one mutex and (in this patch) encodes the round-robin order of flushing the files.

In order to track files that have been written to, one has to take into account that the writing happens in separate threads, with no other synchronization. There would have to be a separate index of references to these files, protected by its own mutex.

arvidn:
this patch will, every 30 seconds, flush the file with the most bytes written to it since the last flush.
The benefit compared to the previous version is that a client that is only seeding will never invoke the (redundant) flush calls.
A possible drawback is that if a single file is the most actively written to, other files with dirty pages won't be considered (until the active files is completely downloaded and writes cease).

glassez:
> this patch will, every 30 seconds, flush the file with the most bytes written to it since the last flush. The benefit compared to the previous version is that a client that is only seeding will never invoke the (redundant) flush calls. A possible drawback is that if a single file is the most actively written to, other files with dirty pages won't be considered (until the active files is completely downloaded and writes cease).

Along with this, you could also provide an API method so that application developers can apply their own flushing optimization algorithms. Then you also need to provide the ability to disable the built-in flushing mechanism (or even better, configure it - maybe it will be enough for someone to change its timeout).

arvidn:
@sledgehammer999 could you give this new patch a try?
my testing suggests that it works as intended

sledgehammer999:
@arvidn give me ~5-6hours when I'll be home to test it.

sledgehammer999:
**Test case**: Torrent with a 20GB file downloading
**Result:**
1. Filesize on disk seems to be updated regularly. Probably every 30secs.
2. System RAM usage seems to be under control. It gets "released" after ~2GBs are used every few minutes.
3. I didn't observe any speed slowdown during flushing. But don't take my word for it. It was a simple test. It may have been just luck.
4. Not important: File timestamps for accessed/modified time don't get updated after flushing.

arvidn:
thanks for testing!

PR #6538, opened 2021-10-27
fix issue where incoming uTP connections were not accepted over SOCKS5

arvidn:


arvidn:
back-ported to 1.2 here: https://github.com/arvidn/libtorrent/pull/6545

PR #6549, opened 2021-11-02
use cached DNS resolver for socks5 proxy

arvidn:


tjhorner:
(Posting here before opening a new issue in order to not spam the issue tracker.)

Has this bug resurfaced? I am seeing a deluge (heh) of DNS queries coming from Deluge using libtorrent 2.0.11. Rate of about 20/sec.

Version info:

```
deluge 2.1.1
libtorrent: 2.0.11.0
Python: 3.12.10
OS: Linux
```

Specifically, I am using [this Docker image](https://github.com/linuxserver/docker-deluge) which uses the Alpine distribution of [Deluge](https://pkgs.alpinelinux.org/package/edge/community/x86_64/deluge) (which in turn uses the Alpine-packaged [libtorrent-rasterbar](https://pkgs.alpinelinux.org/package/edge/community/x86_64/libtorrent-rasterbar) and [python bindings](https://pkgs.alpinelinux.org/package/edge/community/x86_64/py3-libtorrent-rasterbar)).

Let me know if you need any more information to dig into this.

Issue #6548, opened 2021-10-31
qBittorrent makes way too many DNS queries when using a SOCKS5 proxy


MarkWieczorek:
**Please provide the following information**

libtorrent version (or branch): Installed via qbittorent (libtorrent-rasterbar-dbg/focal 1.2.14+git20211025.eb4bbfd49c-1ppa1~20.04 amd64)

platform/architecture:  Ubuntu server 20.04.3

I am connecting to a VPN by a SOCKS5 proxy server (proxy.torguard.org) with qBittorrent. The problem is that when using a proxy connection, qBittorrent makes alot of DNS queries to the proxy server. The number of queries is so large that it sometimes exceeds 1000 queries per minute. When this happens, my pihole DNS resolver exceeds the default number of allowed queries per minute, and hence stops responding to future requests from qBittorrent. This behavior has been [confirmed](https://github.com/qbittorrent/qBittorrent/issues/15652).

I have tried a few things to help fix this issue, but none of them really worked.

    I have disabled "Enable Local Peer Discovery to find more peers"
    I have disabled "Resolve peer countries:"
    I have reduced the number of connections to a maximum of 200.

Though these changes might affect the number of queries, the number of queries is still way higher than what is normal. A possible work around would be to turn off DNS rate limits on pihole, but there is obviously a problem with qBittorrent, so maybe this wouldn't be a good idea. I note that I am only downloading a single file.

I note that the same behavior occurs with the Deluge client on Ubuntu.

arvidn:
interesting, I'll take a look at this

arvidn:
could someone give this patch a try? https://github.com/arvidn/libtorrent/pull/6549

MarkWieczorek:
Thanks for looking into this!

arvidn:
that patch has landed

Issue #6467, opened 2021-09-15
bdecode token_limit argument unattainable internally


Lymkwi:
**Required information**

libtorrent version (or branch): 2.0.4 (current `RC_2_0`)

platform/architecture: x86_64-linux-gnu

compiler and compiler version: GCC 11.1.0

**Issue**

Over the past couple months I've frequently had my torrent client silently (and yet messily) fail to add torrents due to `"bencoded item limit exceeded"`. After messing around in libtorrent, I worked out a way to expose `token_limit` and `depth_limit` in the Python bindings used by my client.
However, while testing, I noticed that some calls to `bdecode` would still fail while adding those large torrents, and I traced those back to internal calls made by methods within libtorrent to the default definitions of `bdecode` written [here](https://github.com/arvidn/libtorrent/blob/2e6f8f68d983bd9d050664d3f8537c72ba6d2c2d/include/libtorrent/bdecode.hpp#L455) with default arguments. For example, while my client managed to decode a torrent's data to a `bdecode_node`, building a `torrent_info` class with it would immediately fail with the "token limit exceeded" error. The torrent info class does not have a parameter to expose the values it uses for `token_limit` and `depth_limit` in its [internal call](https://github.com/arvidn/libtorrent/blob/2e6f8f68d983bd9d050664d3f8537c72ba6d2c2d/src/torrent_info.cpp#L936). Similarly, [other](https://github.com/arvidn/libtorrent/blob/2e6f8f68d983bd9d050664d3f8537c72ba6d2c2d/src/torrent.cpp#L7477) [places](https://github.com/arvidn/libtorrent/blob/2e6f8f68d983bd9d050664d3f8537c72ba6d2c2d/src/http_tracker_connection.cpp#L512) seem to use it with default values without exposing it (and my guess is that these could fail as well in case of a large torrent).

The binding fix manages to preserve backwards compatibility, but I do not see the point of even exposing these parameters if the internal code still uses the default, lower hardcoded limit. Would it be possible to either expose those limits everywhere they are used internally, or provide another mechanism to provide them when using libtorrent? My current hotfix (alonside some additions to the Python binding) is to change the default value for `token_limit` in the three definitions to be a compile-time constant that defaults to `2000000` (its current value), but can be redefined by whoever compiles it. This is not great in my opinion and could be done way better.

arvidn:
> The torrent info class does not have a parameter to expose the values it uses for token_limit and depth_limit in its internal call.

[...]

PR #6550, opened 2021-11-02
Test merge

arvidn:


.pre-commit-config.yaml
```diff
@@ -48,6 +48,7 @@ repos:
             tools/clean.py|
             tools/copyright.py|
             tools/dht_flood.py|
+            tools/libtorrent_lldb.py|
```
AllSeeingEyeTolledEweSew, line 51: This shouldn't be masked

It looks like `tools/libtorrent_lldb.py` is auto-formatted and linted in `master`, but not in `RC_2_0`.

MRSCARTER06 (changes requested):
> #6551 #6550 #6467 #6548 #6535

arvidn:
@MRSCARTER06 I don't know what you're trying to communicate

PR #6560, opened 2021-11-13
fix build with WolfSSL

arvidn:


arvidn:
@userdoc So, this is the fix, right?

userdocs:
@arvidn I just tested this branch and it built successfully and loaded in deluge.

```
common.copy /root/.local/lib/python3.9/site-packages/libtorrent.so
...updated 270 targets...

 Python binding file is located at: /root/.local/lib/python3.9/site-packages/libtorrent.so

 Python binding version is: 2.0.4.0
```

The only difference behaviour I observed was that in deluge using an ubuntu torrent https seems not resolve properly.

image

```
0
https://torrent.ubuntu.com/announce
1
https://ipv6.torrent.ubuntu.com/announce
```

With openssl I know why this is and it works with system openssl or locally built if I specify the correct prefixes. I am having difficulty running deluge using a python binding built against `libwolfssl` to test. 

Aside from this it builds ok

PR #6568, opened 2021-11-27
allow unspecified address for android api >= 24

aldenml:
Hi @arvidn, this is the remaining change for the android fix. I kind of understand the implications, but I can't gauge the actual negative effect, since I don't know what's the likelihood of a phone having multiple routable IPv6 addresses (I have never seen one).

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -2076,7 +2076,19 @@ namespace {
 #endif
 			}
 
+#if defined TORRENT_ANDROID && __ANDROID_API__ >= 24
+			// For Android API >= 24, enum_routes with the current NETLINK based
+			// implementation is unsupported (maybe in the future the operation
+			// will be restore using another implementation). If routes is empty,
+			// allow using unspecified address is a best effort approach that
+			// seems to work. The issue with this approach is with the DHTs,
+			// because for IPv6 this is not following BEP 32 and BEP 45. See:
+			// https://www.bittorrent.org/beps/bep_0032.html
+			// https://www.bittorrent.org/beps/bep_0045.html
+			if (!routes.empty()) expand_unspecified_address(ifs, routes, eps);
+#else
 			expand_unspecified_address(ifs, routes, eps);
+#endif
 			expand_devices(ifs, eps);
 		}
 


arvidn:
there are non-phones running Android, I'm pretty sure, and it's not really a matter of multiple routable IPv6 addresses. Any kind of multi-homing is affected. Phones often have both cell-network and wifi connectivity for instance.

Regardless, that doesn't really matter for the purposes of fixing this issue.

If we can know at compile-time that `enum_routes()` won't work, do you think it would make sense to just not make that function available at all? That would move run-time errors into compile- or link errors (which I think is an improvement in general).

AllSeeingEyeTolledEweSew:
This is a drive-by comment, but I've been staring at `expand_unspecified_address` recently and I'm skeptical that it achieves its goal, at least on Linux.

On Linux, the relationship between addresses, routes and reachability isn't straightforward. If you configure multiple WAN interfaces with standard software, you'll get multiple default routes but only one will be picked. Binding a socket to either address doesn't guarantee traffic goes over its interface. The only way to have "true" multi-homing is with the help of high-level scripts to manage source routing with `ip rule`, and these tools never gained mainstream traction (I'm only aware of distro-specific [pingu](https://github.com/ncopa/pingu) for alpine, and [mwan3](https://openwrt.org/docs/guide-user/network/wan/multiwan/mwan3) for openwrt). [BEP 45](https://www.bittorrent.org/beps/bep_0045.html) is written in a way I would call misleading -- it presumes that reachability is implied by multiple interfaces, and that reachability is achieved by binding to a particular address. IMO, `expand_unspecified_address` and BEP 45 compliance is a job for the user or at least the app, not for libtorrent.

I don't have enough Android/Mac/Windows experience to know if this comment applies to those platforms.

aldenml:
@AllSeeingEyeTolledEweSew your comment is interesting (my view was very simplistic in comparison).

@arvidn as far as I know, android phones don't have active routes for cell and wifi at the same time. They can quickly switch back and forth, but only one is "active" from the application perspective.

> If we can know at compile-time that enum_routes() won't work, do you think it would make sense to just not make that function available at all?

I don't know how that would improve the situation.

What I'm looking for is essentially a workaround. Android phones connectivity is very dynamic, and for that reason is common to just keep the unspecified address in the settings. But the current code expands these addresses to nothing, and then libtorrent is not able to work.

What about this: if `routes` is empty, then filter out unspecified IPv6 and allow `0.0.0.0` to work?

arvidn:
> I don't know how that would improve the situation.

Not by very much. It would trigger a link error if any code ever tried to use the `enum_routes()` function on a system where we know it doesn't work.

In your patch, you check whether `routes` is empty, is this a proxy for determining whether `enum_routes()` works or not?
Just in case your running on on older phone?

aldenml:
The check on `routes` is for the case `enum_routes()` works in the future. With this merge, I assume you concluded that this is better than nothing. Let me know if you have additional ideas to improve the situation.

arvidn:
one (minor) thing I was thinking about was that maybe it would be better to look at the error code from `enum_routes()` rather than assuming 0 routes meaning it failed. But I haven't thought that through.

PR #6573, opened 2021-11-28
new setting to allow disabling DHT implied-port

arvidn:


an0n666:
@arvidn IPv6 doesn’t use NAT. So when libtorrent is announcing IPv6 to DHT, it shouldn’t use implied port feature.

arvidn:
@an0n666 true, there's no value to use if for IPv6. I don't fully understand how it could fail though. Even for IPv6, the DHT and peer socket is the same, and should share port.

alannz42:
> @arvidn IPv6 doesn’t use NAT. So when libtorrent is announcing IPv6 to DHT, it shouldn’t use implied port feature.

IPV6 can use NAT. It is used as a security feature to hide the topology of your internal network from the public Internet. In this case the IPV6 network can be hidden behind a single external address.

arvidn:
I don't think this change is important. My understanding of the issue this was meant to mitigate has evolved.

PR #6581, opened 2021-12-04
add project-config.jam to the root directory

arvidn:


AllSeeingEyeTolledEweSew:
In the interest of understanding b2 better -- what's the goal of this change?

arvidn:
`b2` searches parent directories to find the "root" of a project. This is to ensure that sub-projects that belong together use a uniform configuration. The documentation says that a directory with a `Jamroot.jam` file will be considered the root and the search will stop at that level. However, this does not seem to be the case in practice. In my experiments the search is terminated with a `project-config.jam` though.

I wouldn't expect this to cause problems normally, but in my current `build_dist.sh` script, I pack up a tarball, unzip it, cd into it to test it. This means that a libtorrent source directory will be nested inside of another one. This is how I discovered that the search isn't terminated, both the `Jamfile` in the freshly unzipped tarbar *and* the `Jamfile` in the directory below it are loaded. Since they are the same, the both declare build features that clash, causing the build to fail.

This fixes that failure.

AllSeeingEyeTolledEweSew:
I think `setup.py` autogenerates `bindings/python/project-config.jam` and deletes it afterward. (We initially had it generate a temp file and invoke `b2 --project-config=/foo` but some older `b2` versions don't support this). Could that clash with this?

arvidn:
I don't think expect that to be an issue

PR #6583, opened 2021-12-05
Fix master merge

AllSeeingEyeTolledEweSew:
The test config is definitely inelegant. There's probably a better way but I'll figure it out later.

There were some other build environment versions to bump for C++17 support.

I found that `export_filter()`'s output is normalized differently on alpine linux only. See [this workflow run](https://github.com/AllSeeingEyeTolledEweSew/libtorrent/runs/4423515749?check_suite_focus=true#step:5:2453). From what I can tell this normalization comes from boost, and we always build boost from source, so I'm very surprised.

### DIFF
--- a/bindings/python/tests/ip_filter_test.py
+++ b/bindings/python/tests/ip_filter_test.py
@@ -1,6 +1,24 @@
+import ipaddress
+from typing import List
+from typing import Tuple
 import unittest
 
 import libtorrent as lt
+
+
+# boost's address formatting is currently inconsistent between alpine linux and
+# other environments, even when boost is built from source
+def normalize(
+    exported_filter: Tuple[List[Tuple[str, str]], List[Tuple[str, str]]]
+) -> Tuple[List[Tuple[str, str]], List[Tuple[str, str]]]:
+    def norm(addr: str) -> str:
+        return str(ipaddress.ip_address(addr))
+
+    def norm_list(filter_list: List[Tuple[str, str]]) -> List[Tuple[str, str]]:
+        return [(norm(lo), norm(hi)) for lo, hi in filter_list]
+
+    v4, v6 = exported_filter
+    return (norm_list(v4), norm_list(v6))
 
 
 class IpFilterTest(unittest.TestCase):
@@ -12,7 +30,7 @@ class IpFilterTest(unittest.TestCase):
         self.assertEqual(ipf.access("::123"), 0)
 
         self.assertEqual(
-            ipf.export_filter(),
+            normalize(ipf.export_filter()),
             (
                 [("0.0.0.0", "255.255.255.255")],
                 [("::", "ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff")],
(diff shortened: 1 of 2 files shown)


bindings/python/tests/ip_filter_test.py
```diff
@@ -1,8 +1,26 @@
+import ipaddress
+from typing import List
+from typing import Tuple
 import unittest
 
 import libtorrent as lt
 
 
+# boost's address formatting is currently inconsistent between alpine linux and
+# other environments, even when boost is built from source
```
arvidn, line 10: I don't think boost defines an address format. I'm pretty sure asio is just deferring to the system API.
could we just make the test compare address objects instead of strings? I would think that would make it clearer.

Btw, is this a new issue in `master`?

pyproject.toml
```diff
@@ -11,14 +17,16 @@ before-all = [
...
+    "python -X dev -m pytest --log-level=DEBUG --log-cli-level=DEBUG",
+    "python -m mypy --config-file mypy-tox.ini tests",
 ]
 
 [tool.cibuildwheel.macos.environment]
 BOOST_BUILD_PATH = "/tmp/boost/tools/build"
 BOOST_ROOT = "/tmp/boost"
 BOOST_VERSION = "1.76.0"
-MACOSX_DEPLOYMENT_TARGET = "10.9"  # required for full C++11 support
+MACOSX_DEPLOYMENT_TARGET = "10.14"  # required for full C++17 support
```
arvidn, line 29: in `RC_2_0` we need C++14 support too...

pyproject.toml
```diff
@@ -75,7 +87,7 @@ before-all = [
 select = "*-win_amd64"
 
 [tool.cibuildwheel.windows]
-test-command = '''bash -c "cd '{project}/bindings/python' && tox -e"'''
+test-command = '''bash -c "cd '{project}/bindings/python' && python -m pip install -r test-requirements.txt && python -X dev -m pytest --log-level=DEBUG --log-cli-level=DEBUG && python -m mypy --config-file mypy-tox.ini tests"'''
```
arvidn, line 90: once installing the test dependencies, would it not work to just run `pytest tests`?

pyproject.toml
AllSeeingEyeTolledEweSew: [10.9 is sufficient for C++14](https://cibuildwheel.readthedocs.io/en/stable/cpp_standards/#macos-and-deployment-target-versions), but I guess the comment is out of sync with the requirement.

pyproject.toml
AllSeeingEyeTolledEweSew: This copies the testing invocation from `tox`. It's not sufficient to invoke `tox` because it creates an isolated environment, and we want to test in the outer `cibuildwheel` environment. I'll factor these invocations out somewhere so it can be shared.

This config block is only duplicated in `pyproject.toml` several times due to shell quoting issues on windows. If I factor it out into a script, that can probably be avoided.

I wrote some justification for why we test what we test in `tox.ini`. The highlights are that `python -X dev` has been super useful in the past for catching bugs and producing good stack traces, and that this particular `mypy` run is crucial for testing that the type stubs got *installed* correctly from the build artifact. Running `mypy` in the source tree just tests the type stubs in the tree.

bindings/python/tests/ip_filter_test.py
AllSeeingEyeTolledEweSew: Libtorrent's API returns the inconsistently-normalized strings, we can't compare to address objects directly.

We could make `normalize()` return address objects, but that would make the comparisons more verbose due to constructing lots of objects versus just strings. I prefer to keep test comparisons succinct and readable when possible.

I haven't tested, but I don't expect it's an issue in `master`, just something that gets uncovered by the expanded test suite, and only now uncovered due to running that on alpine for the first time.

PR #6585, opened 2021-12-06
fix python version comparison

arvidn:


bindings/python/Jamfile
```diff
@@ -110,14 +112,18 @@ else
...
+		if $(py-version-str) {
+			py-version = [ SPLIT_BY_CHARACTERS $(py-version-str) : "." ] ;
+
+			if [ version.version-less $(py-version) : 3 7 ] && [ $(properties).get <target-os> ] = linux
+			{
+				infix = "-py" ;
+			}
 		}
-		local boost-python-lib = "boost_python" $(infix) [ SPLIT_BY_CHARACTERS $(py-version) : "." ] ;
+		local boost-python-lib = "boost_python" $(infix) $(py-version-str) ;
```
sergiodj, line 126: I think this should be:

```suggestion
		local boost-python-lib = "boost_python" $(infix) $(py-version) ;
```

Otherwise I get the following error:

```
/usr/bin/ld: cannot find -lboost_python3.10
```

sergiodj (changes requested):
Thanks for the quick PR, @arvidn!

This almost works; there's a small change (typo?) that needs to addressed in order to make it fully build here.

Other than that, it fixes the problem and LGTM :-).

arvidn:
right, of course. I force pushed that change.

sergiodj:
Thanks, LGTM.

sergiodj (approved):

Issue #6447, opened 2021-09-02
Segfaults in 32-bit linux while running master python test suite, maybe allocation related


AllSeeingEyeTolledEweSew:
I'm trying to port #6188 to master. I'm getting segfaults running the python test suite, only on 32-bit linux.

```
#0  listen_socket_t (this=0xf69f25a0 <std::string::_Rep::_S_empty_rep_storage>) at ../../include/libtorrent/aux_/session_impl.hpp:155
#1  construct<libtorrent::aux::listen_socket_t> (__p=0xf69f25a0 <std::string::_Rep::_S_empty_rep_storage>, this=<optimized out>)
    at ../../include/libtorrent/aux_/session_impl.hpp:155
#2  construct<libtorrent::aux::listen_socket_t> (__p=0xf69f25a0 <std::string::_Rep::_S_empty_rep_storage>, __a=...)
    at /opt/rh/devtoolset-10/root/usr/include/c++/10/bits/alloc_traits.h:512
#3  _Sp_counted_ptr_inplace<> (__a=..., this=0xf5d022c8) at /opt/rh/devtoolset-10/root/usr/include/c++/10/bits/shared_ptr_base.h:551
#4  __shared_count<libtorrent::aux::listen_socket_t, std::allocator<libtorrent::aux::listen_socket_t> > (__a=..., 
    __p=@0x8468be0: 0xf7065558 <vtable for libtorrent::aux::session_impl+8>, this=0x8468be4) at /opt/rh/devtoolset-10/root/usr/include/c++/10/bits/shared_ptr_base.h:682
#5  __shared_ptr<std::allocator<libtorrent::aux::listen_socket_t> > (__tag=..., this=0x8468be0) at /opt/rh/devtoolset-10/root/usr/include/c++/10/bits/shared_ptr_base.h:1371
#6  shared_ptr<std::allocator<libtorrent::aux::listen_socket_t> > (__tag=..., this=0x8468be0) at /opt/rh/devtoolset-10/root/usr/include/c++/10/bits/shared_ptr.h:408
#7  allocate_shared<libtorrent::aux::listen_socket_t, std::allocator<libtorrent::aux::listen_socket_t> > (__a=...)
    at /opt/rh/devtoolset-10/root/usr/include/c++/10/bits/shared_ptr.h:860
#8  make_shared<libtorrent::aux::listen_socket_t> () at /opt/rh/devtoolset-10/root/usr/include/c++/10/bits/shared_ptr.h:876
[...]

PR #6588, opened 2021-12-07
Dummy change to trigger #6447

AllSeeingEyeTolledEweSew:
Dummy change to trigger #6447


AllSeeingEyeTolledEweSew:
Bump, this should be reopened, it's still an issue

PR #6591, opened 2021-12-08
Revert "Change python package name to libtorrent"

AllSeeingEyeTolledEweSew:
This reverts commit 8ea0e60fbee8c594b7aa01fbb2445afa5c0414b3.

Per https://github.com/arvidn/libtorrent/issues/5333#issuecomment-986308800

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -451,7 +451,7 @@ class LibtorrentBuildExt(BuildExtBase):
 
 
 setuptools.setup(
-    name="libtorrent",
+    name="python-libtorrent",
     version="1.2.15",
     author="Arvid Norberg",
     author_email="arvid@libtorrent.org",


AllSeeingEyeTolledEweSew:
Will need the following in `master`

```patch
diff --git a/tox.ini b/tox.ini
index 852fccf33..ea97cda07 100644
--- a/tox.ini
+++ b/tox.ini
@@ -42,7 +42,7 @@ deps =
 
 commands =
     {envpython} setup.py bdist_wheel
-    {envpython} -m pip install --upgrade --force-reinstall --ignore-installed --find-links=dist --no-index libtorrent
+    {envpython} -m pip install --upgrade --force-reinstall --ignore-installed --find-links=dist --no-index python-libtorrent
     {envpython} -m pip install -r test-requirements.txt
     {envpython} -X dev -m pytest --log-level=DEBUG --log-cli-level=DEBUG
     {envpython} -m mypy --config-file mypy-tox.ini tests
```

arvidn:
I would like to understand this a bit better. It seems to me that "libtorrent" is a much better name for the python package. The fact that it's python is implied.

I seem to recall "packagers" being a problem? The idea is that packagers for system packages have script that make system packages that depend on this name, and they would have to update their scripts to be able to package this now. Is that right?

AllSeeingEyeTolledEweSew:
I'll try to recap everything:

The name from `setup(name="python-libtorrent")` is the *distribution* name. It's the name used with `pip install python-libtorrent`.

A distribution is a collection of python modules, packages and extensions. At install time, info  about the distribution will live in `.../site-packages/python-libtorrent-1.2.3.*-info/`, which references all the files that belong to it.

The distribution name is unrelated to names of *modules* (`*.py`), *extensions* (`*.so`) or *packages* (directories). Python's import machinery doesn't (*can't*) know about distributions.

Ingesting distribution info at runtime historically required [`pkg_resources`](https://setuptools.pypa.io/en/latest/pkg_resources.html) which is part of `setuptools`, or the newer [`importlib`](https://docs.python.org/3/library/importlib.metadata.html), which has only been in the standard library since 3.8.

Distributions contain some metadata, particularly entry points, that is used by many python projects. However as far as I'm aware, the use cases for this metadata universally look like `do x for all distributions`, rather than looking up distributions by name.

Dependency metadata requires looking up distributions by name, but as far as I'm aware, `pip` is the only tool that has ever cared about dependency metadata, or otherwise doing lookups by name.

Linux package managers use their own dependency systems, unrelated to python. When a Linux packager installs a python distribution, it will end up installing `/usr/lib/python*/site-packages/python-libtorrent-1.2.3.*-info/*`.

Using `pip` as root *outside* a virtualenv will install stuff to `/usr/lib/python*/site-packages`, and will look for existing dependencies there. This opens up some possibility of conflict between the old and new distribution names. Here are some scenarios I can think of:

1. Clients with a metadata dependency on the old name:
   1. `sudo apt install python3-libtorrent=2.0.5` (uses dist name `libtorrent`)
   2. `sudo pip install my-torrent-client` (depends on dist named `python-libtorrent`)
   3. Outcome: worked prior to 1.2.15 / 2.0.5, now install will fail
2. Users mixing Linux packagers and `pip`
   1. `sudo apt install python3-libtorrent=2.0.4` (uses dist name `python-libtorrent`)
   2. `sudo pip install libtorrent==2.0.5` (uses dist name `libtorrent`)
   3. Outcome: two distributions will try to manage `/usr/lib/python*/site-packages/libtorrent.*.so`, install will fail

However I believe these scenarios are implausible in practice. In scenario 1, there are no python packages that depend on `python-libtorrent` because it's never been (reliably) published to pypi (note that `deluge` doesn't depend on `python-libtorrent`; its setup script tries to import the extension, and installs from source if not found). In scenario 2, the user has been warned about breaking stuff, so they should not be surprised when stuff breaks (although the stuff that breaks is different than expected).

It is possible that Linux packager scripts (debhelper and cousins) may ingest the distribution name, but I can't find any that do.
* Debian has [`python3-libtorrent`](https://packages.debian.org/source/sid/libtorrent-rasterbar)
* Alpine has [`py3-libtorrent-rasterbar`](https://pkgs.alpinelinux.org/package/edge/community/x86_64/py3-libtorrent-rasterbar)
* Arch has [`libtorrent-rasterbar`](https://archlinux.org/packages/extra/x86_64/libtorrent-rasterbar/)
* Gentoo has [`libtorrent-rasterbar`](https://packages.gentoo.org/packages/net-libs/libtorrent-rasterbar)

I see some of these packagers are already using 2.0.5, which uses the dist name `libtorrent`, so perhaps we'll know soon if it causes trouble.

I think the concerns raised by @cas-- are about the "unknown unknowns". It's hard to prove the negative that there are no bad interactions to changing the name.

In trying to strawman my own argument, I can't think of any scenarios that suffer significantly from using `python-libtorrent`. It's slightly less straightforward for users to find when installing libtorrent ad-hoc, but this is likely rare.

cas--:
Let's just leave it as `libtorrent` and the fallout is what it is... The linux package managers can handle the change, I just would not have arbitrarily changed the name and I was airing my opinion that it was changed too quickly without fully understanding the potential impact.

The worst problem I can see is on the Linux distributions where mixing system installed python-libtorrent with pip installed libtorrent can cause unexpected versions to be loaded in Python. 

For reference scenario 1 is not an excepted outcome and in scenario 2 the pip package manager would be unaware that python-libtorrent provides the same library as libtorrent so both would be installed thus causing potential confusion if one is loaded instead of the other. I have actually encountered this when using the testing Pypi...

AllSeeingEyeTolledEweSew:
> The worst problem I can see is on the Linux distributions where mixing system installed python-libtorrent with pip installed libtorrent can cause unexpected versions to be loaded in Python.

Indeed, all the bad cases I can think of come when mixing dependency managers. But all these scenarios, as I understand them, range between perilous and unsupported.

> in scenario 2 the pip package manager would be unaware that python-libtorrent provides the same library as libtorrent so both would be installed thus causing potential confusion if one is loaded instead of the other. I have actually encountered this when using the testing Pypi...

Ah, you're right. I thought I tested this and found that `pip` would detect conflicting files and error out, but I was wrong. It looks like it actually silently overwrites files (https://github.com/pypa/pip/issues/4625). I bet overwriting is more likely than dual versions and one shadowing another, assuming both the wheel and packager use consistent ABI tags, which they did in my case.

We will be able to cover these scenarios in the future, when the packaging team provides a better package conflict / rename story (https://github.com/pypa/packaging-problems/issues/154).

@arvidn it seems like there's consensus to leave the dist name as `libtorrent`. If you agree, would you mind closing this PR and publishing the wheels to pypi production?

AllSeeingEyeTolledEweSew:
Ah the wheels were already on pypi production.

https://pypi.org/project/libtorrent/2.0.5/
https://pypi.org/project/libtorrent/1.2.15/

:tada:

PR #6233, opened 2021-05-29
WebTorrent: Update libdatachannel to v0.13.3

paullouisageneau:
This PR updates libdatachannel to fix https://github.com/paullouisageneau/libdatachannel/issues/427. Breaking changes in v0.13 affect mostly the media API. As libdatachannel does not expose plog anymore, I refactored debug logging.

PR #6602, opened 2021-12-20
WebTorrent: Update libdatachannel to v0.16.1

paullouisageneau:
This PR updates libdatachannel to v0.16.1. It introduces stability fixes (in particular for a deadlock in usrsctp), as well as support for more platforms (especially FreeBSD and Microsoft UWP). The other changes in libdatachannel since the last version are mostly about media and websockets handling, which are not used here.


arvidn:
it looks like the failures exist in master already.. I'll fix those separately

arvidn:
I'm hitting this now:

```
../deps/libdatachannel/src/impl/icetransport.cpp:290:35: runtime error: member call on address 0x000001cd7510 which does not point to an object of type 'IceTransport'
0x000001cd7510: note: object has invalid vptr
 0f 1f 40 00  55 48 89 e5 41 57 41 56  41 55 41 54 49 89 f4 53  48 81 ec 48 02 00 00 48  85 f6 0f 84
              ^~~~~~~~~~~~~~~~~~~~~~~
              invalid vptr
```

I'll try to get better diagnostics

arvidn:
looking around a bit, the `IceTransport` destructor looks suspicious. You call `stop()` and then you tear down the `mAgent` object. It seems to me quite likely you may get callback after destroying the `IceTransport` object itself, which would cause the UB above.

arvidn:
actually, it seems I hit this issue because `libdatachannel` and its dependencies weren't all rebuilt properly. after `git clean -f` in the dependencies, it seems to work.

paullouisageneau:
> actually, it seems I hit this issue because `libdatachannel` and its dependencies weren't all rebuilt properly. after `git clean -f` in the dependencies, it seems to work.

This already happened after the previous update: https://github.com/arvidn/libtorrent/pull/6233#issuecomment-852388139

I'll look into why external targets are not properly rebuilt.

PR #6626, opened 2021-12-31
keep trying to announce to trackers even when all fail

arvidn:


src/torrent.cpp
```diff
@@ -3175,7 +3175,8 @@ bool is_downloading_state(int const st)
 
 				aep.updating = true;
 				aep.next_announce = now;
-				aep.min_announce = now;
+				aep.min_announce = now
+					+ settings().get_int(settings_pack::min_announce_interval);
```
an0n666, line 3179: This is causing the build to fail.
`libtorrent/src/torrent.cpp: In member function ‘void libtorrent::torrent::announce_with_tracker(uint8_t)’:
libtorrent/src/torrent.cpp:3179:41: error: no match for ‘operator+’ (operand types are ‘const time_point32’ {aka ‘const std::chrono::time_point<std::chrono::_V2::system_clock, std::chrono::duration<int> >’} and ‘int’)
 3178 |                                 aep.min_announce = now
      |                                                    ~~~
      |                                                    |
      |                                                    const time_point32 {aka const std::chrono::time_point<std::chrono::_V2::system_clock, std::chrono::duration<int> >}
 3179 |                                         + settings().get_int(settings_pack::min_announce_interval);
      |                                         ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                             |
      |                                                             int
`

an0n666:
Now I can force re-announce without passing the ignore_minimum_interval flag but force-reannounce still doesn't work. It just resets the timer back to 1 min.

arvidn:
hm, so there's still something setting the `min_interval` too high then. I have one more thing to try.

arvidn:
@an0n666 try now!

an0n666:
@arvidn 
Now tracker starts working as soon as I reconnect. No need to re-announce.
Also previously the tracker status would display incorrectly in qBt when you force reannounce in disconnected state.
Now that too seem to have been resolved!

PR #3845, opened 2019-05-31
Replace unicode symbol by ascii

Chocobo1:
MSVC was emitting warning C4819 before this commit.
https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4819?view=vs-2019

Please cherry-pick it to master branch too!

PR #6627, opened 2022-01-01
Set source and executable character sets to UTF-8

Chocobo1:
By default msvc assume the source file is encoded using the current user code page.
https://docs.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-170

Reason I'm submitting this: I still see compiler warnings like in #3845 occasionally and this flag would eliminate them all.


arvidn (approved):

arvidn:
I posted this PR too, for the Jamfile: https://github.com/arvidn/libtorrent/pull/6628

PR #6629, opened 2022-01-02
Clean up warning flags

Chocobo1:
In addition, I would like to suggest the use of `/guard:cf` flag when compiling with msvc. AFAIK it inject runtime checks to guard against buffer overruns. I didn't add it in this PR because I couldn't determine the performance impact, however this flag is [recommended](https://github.com/qbittorrent/qBittorrent/blob/dd76525372c2ce0afed1c636d391eb687d7c5801/conf.pri.windows#L55-L56) in qbittorrent project and we haven't seen noticeable drawbacks of it.

https://docs.microsoft.com/en-us/cpp/build/reference/guard-enable-control-flow-guard?view=msvc-170
https://devblogs.microsoft.com/cppblog/visual-studio-2015-preview-work-in-progress-security-feature/


CMakeLists.txt
```diff
@@ -450,9 +450,9 @@ find_public_dependency(Threads REQUIRED)
 
 if(CMAKE_CXX_COMPILER_ID MATCHES Clang)
 	add_compile_options(
-		-Weverything
```
arvidn, line 453: the idea of using `-Weverything` is to have new warnings enabled, and then disable ones that aren't important or helpful. So generally I would prefer to disable warnings rather than enabling them. Is there a good reason to remove this?

CMakeLists.txt
```diff
@@ -461,19 +461,14 @@ if(CMAKE_CXX_COMPILER_ID MATCHES Clang)
 elseif(CMAKE_CXX_COMPILER_ID MATCHES GNU)
 	add_compile_options(
 		-Wall
-		-Wc++11-compat
```
arvidn, line 464: the `RC_1_2` branch is still meant to build in C++11 mode, so I think this is a good warning to have.

CMakeLists.txt
```diff
@@ -461,19 +461,14 @@ if(CMAKE_CXX_COMPILER_ID MATCHES Clang)
...
 		-Wextra
-		-Wno-format-zero-length
-		-Wparentheses
```
arvidn, line 467: why disable `Wparentheses`?

Jamfile
```diff
@@ -276,10 +273,12 @@ rule building ( properties * )
...
 	if <toolset>msvc in $(properties) || <toolset>intel-win in $(properties)
 	{
-		# allow larger .obj files (with more sections)
-		result += <cflags>/bigobj ;
 		# https://docs.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-170
 		result += <cflags>/utf-8 ;
+		# https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
+		result += <cxxflags>/Zc:__cplusplus ;
+		# allow larger .obj files (with more sections)
+		result += <cflags>/bigobj ;
```
arvidn, line 281: the order here doesn't matter, right?

CMakeLists.txt
Chocobo1: The link (https://quuxplusone.github.io/blog/2018/12/06/dont-use-weverything/) from the relevant commit quote clang developers stance on this flag. TLDR, the flag is for "It is intended strictly for Clang developers or for exploring what warnings exist" and not daily use.

CMakeLists.txt
Chocobo1: it is already covered by `-Wall`.

CMakeLists.txt
Chocobo1: it is not disabled, it is already covered by `-Wall`.

Jamfile
Chocobo1: the order is only for syncing between build scripts for easier maintenance, nothing else.

Jamfile
```diff
@@ -276,10 +273,12 @@ rule building ( properties * )
 
 	if <toolset>msvc in $(properties) || <toolset>intel-win in $(properties)
 	{
-		# allow larger .obj files (with more sections)
-		result += <cflags>/bigobj ;
 		# https://docs.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-170
 		result += <cflags>/utf-8 ;
+		# https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
+		result += <cxxflags>/Zc:__cplusplus ;
```
Chocobo1, line 279: I don't know why it is unrecognized:
https://ci.appveyor.com/project/arvidn/libtorrent/builds/42059904/job/e9gkxsmfgwujpxy0#L125
Any ideas?

Jamfile
```diff
@@ -207,9 +207,9 @@ rule warnings ( properties * )
 	if <toolset>clang in $(properties)
 		|| <toolset>darwin in $(properties)
 	{
-		result += <cflags>-Weverything ;
-		result += <cflags>-Wno-documentation ;
-		result += <cxxflags>-Wno-c++98-compat-pedantic ;
+		result += <cflags>-Wall ;
```
Chocobo1, line 210: I think it would be better if those `cflags` were actually `compileflags` as the first is for C compiler only and the latter is for both C and C++ compiler.

https://www.boost.org/doc/libs/1_78_0/tools/build/doc/html/index.html

I'm unfamiliar with boost build system and I probably should stop here...

Jamfile
Chocobo1: It is because of `:` and it need double quotes around it.

CMakeLists.txt
arvidn: It makes me a bit nervous since it's hard to know if there are any warnings that are disabled by this change. The code honours the existing warnings, so disabling them won't be noticed by CI

arvidn (approved):

Jamfile
Chocobo1: unfortunate that appveyor CI still don't recognize it:
https://ci.appveyor.com/project/arvidn/libtorrent/builds/42060994/job/owm1w4ublhnvlyw0#L124
while Github Actions have no problem:
https://github.com/arvidn/libtorrent/runs/4685279536?check_suite_focus=true#step:5:36

CMakeLists.txt
aldenml: A lot of effort has been put in `libtorrent` to not have warnings and I don't recall a problem due to the use of `Weverything`. My opinion is that at this moment there is no harm in keeping it, and a lot to lose by not knowing what "interesting warnings" can come up in new clang versions.

Chocobo1: Well @arvidn do you really want to keep it?

Jamfile
Chocobo1: I think I found the root cause. 
Appveyor is using msvc `14.0` which translate to "Visual Studio 2015 version 14.0" but `/Zc:__cplusplus` only has support starting in "Visual Studio 2017 version 15.7".
Now libtorrent RC_1_2 supports C++11 which also includes "Visual Studio 2015" and that means `/Zc:__cplusplus` isn't available. So none of the build system should have `/Zc:__cplusplus`. I'll make the change soon.

CMakeLists.txt
Chocobo1: Anyway, the change is dropped.

Jamfile
Chocobo1: On the other hand, code is already using `__cplusplus` and removing the flag may cause trouble. Forget it, I'll leave this absurd state as is...

arvidn:
it seems this didn't merge right into `RC_2_0`

PR #6647, opened 2022-01-09
set DSCP_TRAFFIC_TYPE socket option on windows

arvidn:


src/peer_connection.cpp
```diff
@@ -355,6 +355,18 @@ namespace libtorrent {
 					peer_log(peer_log_alert::outgoing, "SET_TOS", "tos: %d e: %s"
 						, m_settings.get_int(settings_pack::peer_tos), ec.message().c_str());
 				}
+#endif
+#ifdef IP_DSCP_TRAFFIC_TYPE
+				ec.clear();
+				m_socket->set_option(dscp_traffic_type(char(m_settings.get_int(settings_pack::peer_tos) >> 2)), ec);
```
arvidn, line 361: note the `>> 2` here

include/libtorrent/socket.hpp
```diff
@@ -175,6 +175,22 @@ namespace libtorrent {
 		tos_t m_value;
 	};
 
+#ifdef IP_DSCP_TRAFFIC_TYPE
+	struct dscp_traffic_type
+	{
+		explicit dscp_traffic_type(char val) : m_value(DWORD(val)) {}
+		template<class Protocol>
+		int level(Protocol const&) const { return IPPROTO_IP; }
```
kjbracey, line 183: The docs seem to be saying (at least to me) that level should be `IP_DSCP_TRAFFIC_TYPE` and name should be `DSCP_TRAFFIC_TYPE`, and this would apply to either IPv4 or IPv6 sockets.

arvidn: ah, yes

arvidn:
should the `IPV6_TCLASS` value also be right-shifted by 2?

kjbracey:
> should the `IPV6_TCLASS` value also be right-shifted by 2?

No, both `IPPROTO_IP | IP_TOS` and `IPPROTO_IPV6 | IPV6_TCLASS` are setting the whole byte, so both should have the same value with the bottom 2 bits clear. They're semantically the same, it's just that IPv4 and IPv6 have different names for what is now the DSCP+ECN byte.

src/peer_connection.cpp
```diff
@@ -355,6 +355,18 @@ namespace libtorrent {
 					peer_log(peer_log_alert::outgoing, "SET_TOS", "tos: %d e: %s"
 						, m_settings.get_int(settings_pack::peer_tos), ec.message().c_str());
 				}
+#endif
+#ifdef IP_DSCP_TRAFFIC_TYPE
```
kjbracey, line 359: This wants to be restructured to

```
    if (m_settings.get_int(settings_pack::peer_tos) != 0)
    {
        if (is_v4())
            type_of_service(val)
#ifdef IPV6_TCLASS
        else if (is_v6())
            traffic_class(val)
#endif
#ifdef IP_DSCP_TRAFFIC_TYPE
        dscp_traffic_type(val >> 2)
#endif
    }
```

Or maybe an if/else on the `IP_DSCP_TRAFFIC_TYPE`? Only attempt one form?

ChangeLog
```diff
@@ -1,3 +1,5 @@
+	* use DSCP_TRAFFIC_TYPE socket option on windows
+	* update defaule ToS setting according to RFC 8622
```
kjbracey, line 2: "default"

src/peer_connection.cpp
kjbracey: Anal nit - you're quietly clamping the other socket options to 8 bits. To match, this one should be quietly clamped to 6 bits, so `dscp_traffic_type(char(m_settings.get_int(settings_pack::peer_tos)) >> 2)`. As it is, you're passing bits 2-9 of the settings value.

arvidn: I'm uncomfortable to right-shift a signed type, but I can truncate it with `&`

kjbracey: Good catch. Yes, I was assuming char was unsigned.

How's about clamping the TOS values to 6 bits with `&` too?

`traffic_class(m_settings.get_int(settings_pack::peer_tos) & 0xfc)`

Save the network from people getting confused and putting a DSCP code into peer_tos.

kjbracey: Actually, the `char` is wrong even as it is. If `char` is signed, that will turn `0xc0` into int `-0x40` (`0xFFFFFFC0`), and that will get passed as the socket option. Kernel may well reject that. Certainly `-1` has a special meaning for `IPV6_TCLASS`, so it won't just be looking at the bottom 8 bits.

Probably haven't hit that because the top-bit-set DSCPs are higher priority and probably not going to be selected for torrents.

ChangeLog
```diff
@@ -1,3 +1,5 @@
+	* use DSCP_TRAFFIC_TYPE socket option on windows
+	* update default ToS setting according to RFC 8622
```
NotTsunami, line 2: I don't see this change in RC_2_0 or in this PR

arvidn: it's from a PR landed earlier today. it hasn't been merged into RC_2_0 yet.

PR #6659, opened 2022-01-10
fix const_cast in create_torrent

arvidn:


src/create_torrent.cpp
```diff
@@ -405,17 +405,17 @@ namespace {
...
 		if (flags & (optimize_alignment | mutable_torrent_support))
-			m_files.optimize(pad_file_limit, alignment, bool(flags & mutable_torrent_support));
+			fs.optimize(pad_file_limit, alignment, bool(flags & mutable_torrent_support));
 
-		m_files.set_num_pieces(static_cast<int>(
-			(m_files.total_size() + m_files.piece_length() - 1) / m_files.piece_length()));
-		m_piece_hash.resize(m_files.num_pieces());
+		fs.set_num_pieces(static_cast<int>(
+			(fs.total_size() + fs.piece_length() - 1) / fs.piece_length()));
+		m_piece_hash.resize(fs.num_pieces());
```
glassez, line 414: Shouldn't you store `fs` only after it is modified?

glassez (approved):

src/create_torrent.cpp
glassez: Sorry, I've misunderstood it previously. Both they are references to the same data so it's ok.

PR #6665, opened 2022-01-13
Use UTF-8 encoding for messages coming from boost on Windows 

Chocobo1:
* Use UTF-8 encoding for messages coming from boost on Windows
  Previously it was using the system default locale encoding.
  https://www.boost.org/doc/libs/1_78_0/libs/system/doc/html/system.html#ref_macros
* Remove redundant compiler warning flag
  The flag is already implied by `-Wall`.


arvidn (approved):

arvidn:
this patch enables https://github.com/arvidn/libtorrent/pull/6669

stalkerok:
> * Use UTF-8 encoding for messages coming from boost on Windows
>   Previously it was using the system default locale encoding.

This is a bad change. If you can do without it, it's best not to use it.

Chocobo1:
> This is a bad change. If you can do without it, it's best not to use it.
> 

Is it a screenshot of qbt?
If so, then it is better to patch up qbt instead IMO.

stalkerok:
Yes, this is a screenshot from qbt.
If this can be done in qbt, then I'll close the PR.

PR #6674, opened 2022-01-18
cleanup CMakeLists.txt for python bindings

Kolcha:
see commits messages for detailed description

### DIFF
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -116,7 +116,6 @@ if (python-egg-info)
 	configure_file(${SETUP_PY_IN} ${SETUP_PY} @ONLY)
 
 	add_custom_command(OUTPUT ${OUTPUT}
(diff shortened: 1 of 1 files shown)


arvidn (approved):
the only way I can judge the correctness of this is by the tests, but it seems reasonable

arvidn:
I don't think there are any tests exercising this path thought, are there?

Kolcha:
from my point of view there is nothing to test, these changes affect only python-specific metadata which can be used by `pip` and maybe some other tools (at least `pip list` can display `libtorrent  2.0.5` if this metadata present alongside library itself).
I decided to fix this just because it waas annoying to version like this `2.0.5-usr-src-packages-BUILD-bindings-python` (real example from build server I'm using) and I noticed the warning from python tools noticed on one of commits

arvidn:
building the python bindings using cmake is a feature that isn't exercised by the CI (as far as I can tell), and if this breaks it, I can't tell. If it was exercised by CI, I *could* tell. That's what I mean by not being covered by a test.

PR #6679, opened 2022-01-22
measure memory usage in benchmarks (on linux)

arvidn:


arvidn:
Example output on a Raspberry PI:

download benchmark:

memory_stats log

seeding benchmark:

memory_stats log

PR #6684, opened 2022-01-23
explicitly run homebrew's python version on MacOS CI

arvidn:


arvidn:
this makes the installation step work, but the run-step still can't find the libtorrent module.

@AllSeeingEyeTolledEweSew I don't know what else to try now.

Issue #6578, opened 2021-11-30
Loading from another swarm


master255:
@arvidn I am using the 1.2.13 version of the library.

On the first computer I have files distributed in alignment by pieces.
On the second computer I download them.

When I change at least one file on the first computer (and recreating the torrent file.), the second computer stops downloading.
I tried adding the same line in the collection - did not help.

What do I need to do to keep downloading? After all, the hashes of unchanged files are the same.
Perhaps another version of the library can do this? Or did I use the collections incorrectly somehow?

arvidn:
> When I change at least one file on the first computer (and recreating the torrent file.), the second computer stops downloading.
> I tried adding the same line in the collection - did not help.

Why does it stop downloading? Is it because you change the file from under the feet of the seeder and it ends up sending corrupt data to the downloader?

If you don't want the download to stop, don't change the file. If you want the downloader to download the new file, add the new torrent to its session

arvidn:
> Why doesn't it start?

I was going to ask you that. Do you have any peers?

see this trouble shooting guide: https://libtorrent.org/troubleshooting.html

> How do the collections work? I couldn't find any details about this anywhere. What data should be put in the collections and how will it work?

Collections are arbitrary names you can associate with torrents. Torrents belonging to the same collection (i.e. having a matching string in the collections list) will be assumed to share many piece hashes. It's a hint to libtorrent to go compare them and possibly reuse files.

iirc, files are only reused from seeding torrents, this is to avoid having to coordinate which pieces are downloaded across swarms.

master255:
@arvidn 

> I was going to ask you that. Do you have any peers?

1) Probably not. Are the peers computers with the same hash I'm looking for?
2) Or are peers computers with the same or similar hash mentioned in the info section of the torrent file?
3) Or are peers computers with the same hash or similar collections mentioned in the info section of the torrent file?

arvidn:
peers are other clients part of the same swarm as you. A swarm is identified by its info-hash.

arvidn:
the "mutable torrents" feature is purely local to your machine. Think of it as a way to hint that you may have large portions of a torrent already. i.e. if you have downloaded version 1 of a torrent, downloading v2 may reuse most of the files you already have from v1.

master255:
@arvidn Okay. I got it.
Why aren't similar hashes used to find pieces of files?
It's an obvious feature and should be easy to add to the library functionality.

arvidn:
[...]

PR #6686, opened 2022-01-24
fix similar torrents feature for v2 torrents

arvidn:


master255:
@arvidn I don't know, does that apply to [my question? ](https://github.com/arvidn/libtorrent/issues/6578#issuecomment-1024952740)

arvidn:
@master255 I don't think it does

master255:
@arvidn This is very bad.

arvidn:
how so?

master255:
@arvidn Maybe I don't understand something?
I asked - it's relevant to my question
You answered: No, it doesn't.
I wrote: it's very bad.
Unexpected response: how so? - What do you mean by that?

arvidn:
You're saying there's something wrong (or "bad") with this patch. I'm curious what, specifically, is bad about it.

master255:
@arvidn I'm saying it's too bad this patch doesn't solve my issue. I haven't looked at the patch code in detail and don't understand what it does.

PR #6703, opened 2022-01-31
add option to flush pieces to disk as soon as they are validated

arvidn:
This is an attempt to address issues raised (mostly on windows, but not exclusively) where the operating system does not flush dirty pages of memory mapped files early enough, and doesn't prioritize flushing them when the system is running low on physical RAM.

I can't measure a difference on windows running the benchmark (but it's a VM).

Before (with `disk_io_write_mode` = `enable_os_cache`):
memory_stats log-memory

after (with `disk_io_write_mode` = `write_through`):
memory_stats log-memory

Similarly, on Ubuntu, no obvious difference from my measurements:
Before (with `disk_io_write_mode` = `enable_os_cache`):
memory_stats log-memory

after (with `disk_io_write_mode` = `write_through`):
memory_stats log-memory


arvidn:
this patch also enables this option by default on windows. i.e. `disk_io_write_mode` = `write_through`

jagannatharjun:
now and then drops 12gig but then grows up again
image

the screenshot was taken while checking 150 gig torrent with ~2 gig files

arvidn:
@jagannatharjun I don't consider it a problem that memory is being used (alone). If you have some problem where your system slows down, or other applications suffer, *that* would be a problem.

Second, this patch doesn't affect checking torrents, it affects downloading. I don't know of a way, on windows, to specify that (non-dirty) pages should be given a low priority, and first in line to be evicted. There is a process-wide memory priority that could have an effect.

arvidn:
@jagannatharjun would you mind testing this patch? https://github.com/arvidn/libtorrent/pull/6711

jagannatharjun:
>  don't consider it a problem that memory is being used (alone). If you have some problem where your system slows down, or other applications suffer, that would be a problem.

It does seem to affect system performance somewhat. Plus for a non-technical user (majority of users) a program filling their  RAM to 98% is not a good UX (IMHO)

> would you mind testing this patch? https://github.com/arvidn/libtorrent/pull/6711

doesn't seem to have any effect 
image

arvidn:
@an0n666 Anything is possible. The disk I/O in 1.2 is quite a bit more complex though and also had the problems of the OS disk cache growing too large and evicting other processes (on windows). The solution there was to regularly close files, which is not a great solution. If you have one large file, the disk cache will grow by the amount you can download in 2 minutes, before the file is closed. If you have many small files, and a relatively small file pool, then you'll be OK either way, because only so many files will be open at one time.

That said, someone brought up an issue with network shares, where a memory mapped file will end up having to *read* pages from the disk, in order to map them, even if the purpose is to write to those pages and have them flushed. This may affect non-network shares as well, but especially filesystems that don't suppose sparse files (because most pages we write to are holes normally).

To address this I started working on a relatively simple win32 overlapped I/O backend. The idea is to still not have an application layer cache, but just a store buffer (similar to the mmap backend) and issue asynchronous read and write calls. That way the filesystem driver will (probably) have an easier time turning that into requests for a network share.

jagannatharjun:
> I don't consider it a problem that memory is being used (alone). If you have some problem where your system slows down, or other applications suffer, that would be a problem.

there is qbittorrent bug report regarding this https://github.com/qbittorrent/qBittorrent/issues/16166

an0n666:
@arvidn its been a long time since the last version bump. So once you merge patches like these please bump the version so we can get the latest commits in qbittorrent action builds.

arvidn:
yeah. I'm hoping to get the benchmark to log memory stats on windows before I land this, to validate it does something.
it was also pointed out that I probably should use `msync()` instead of `madvise(MADV_PAGEOUT)`.

xavier2k6:
>latest commits in qbittorrent action builds.

@an0n666 You don't need `arvidn` to bump versions to get latest commits (they just need to be merged) in GitHub Actions builds to use with `qBittorrent`

The workflow files can easily be modified on your own `forked` repo of qBittorrent to build with `RC_2_0/RC_1_2` Branches/Boost 1.78/Qt 6.2.3 etc.

Windows example:
https://github.com/xavier2k6/qBittorrent/blob/CI_latest_lt_qbt_Qt_Boost/.github/workflows/ci_windows.yaml

xavier2k6:
@arvidn Boost 1.78 (`Asio`) recently introduced an `io_uring` backend, would this be of benefit at all to `libtorrent`?

Reference:
https://www.boost.org/doc/libs/1_78_0/doc/html/boost_asio/history.html

https://www.boost.org/users/history/version_1_78_0.html

PR #6707, opened 2022-02-03
Add workaround for windows shared drive directory issue

FletcherD:
I've noticed an issue with the way file handles on shared drives are handled than can prevent libtorrent from downloading torrents to Windows shared drives. I also created a workaround that fixes this and a pull request.

The issue is this:
When starting to write a file in a torrent libtorrent checks if it can open the file using CreateFileW. If the directory doesn't exist then Windows responds with "ERROR_FILE_NOT_FOUND". Libtorrent then tries to create the directory tree and tries again.

However when we try to create a file on a Windows network drive, there can be some odd behavior. If the directory doesn't exist Windows instead returns error code 1117, "ERROR_IO_DEVICE". This doesn't really make sense to me because there is no real I/O error with the network drive, the directory just doesn't exist.

This causes libtorrent to fail instead of trying to create the directory.
As a workaround I just check for this code too on Windows platforms and also try to create the directory in this case. 

It's possible this could also occur when a real I/O device error occurs, however I don't see this having much impact as if there is a real I/O error, the subsequent directory creation will fail too.

### DIFF
--- a/src/mmap_storage.cpp
+++ b/src/mmap_storage.cpp
@@ -903,17 +903,15 @@ error_code translate_error(std::system_error const& err, bool const write)
 #endif
 
 		boost::optional<aux::file_view> h = open_file_impl(sett, file, mode, ec);
-		
-		bool should_retry = (mode & aux::open_mode::write)
-			&& ec.ec == boost::system::errc::no_such_file_or_directory;
+		if ((mode & aux::open_mode::write)
+			&& (ec.ec == boost::system::errc::no_such_file_or_directory
 #ifdef _WIN32
-		// this is a workaround for improper handling of files on windows shared drives.
-		// if the directory on a shared drive does not exist, 
-		// windows returns ERROR_IO_DEVICE instead of ERROR_FILE_NOT_FOUND
-		should_retry |= (ec.ec.value() == ERROR_IO_DEVICE);
+				// this is a workaround for improper handling of files on windows shared drives.
+				// if the directory on a shared drive does not exist, 
+				// windows returns ERROR_IO_DEVICE instead of ERROR_FILE_NOT_FOUND
+				|| ec.ec == error_code(ERROR_IO_DEVICE, boost::asio::error::system_category)
 #endif
-
-		if (should_retry)
+		))
 		{
 			// this means the directory the file is in doesn't exist.
 			// so create it
--- a/src/mmap_storage.cpp
+++ b/src/mmap_storage.cpp
@@ -903,8 +903,17 @@ error_code translate_error(std::system_error const& err, bool const write)
 #endif
 
 		boost::optional<aux::file_view> h = open_file_impl(sett, file, mode, ec);
-		if ((mode & aux::open_mode::write)
-			&& ec.ec == boost::system::errc::no_such_file_or_directory)
+		
+		bool should_retry = (mode & aux::open_mode::write)
+			&& ec.ec == boost::system::errc::no_such_file_or_directory;
+#ifdef _WIN32
+		// this is a workaround for improper handling of files on windows shared drives.
+		// if the directory on a shared drive does not exist, 
+		// windows returns ERROR_IO_DEVICE instead of ERROR_FILE_NOT_FOUND
+		should_retry |= (ec.ec.value() == ERROR_IO_DEVICE);
+#endif
+
+		if (should_retry)
 		{
 			// this means the directory the file is in doesn't exist.
 			// so create it


src/mmap_storage.cpp
```diff
@@ -903,8 +903,17 @@ error_code translate_error(std::system_error const& err, bool const write)
...
-		if ((mode & aux::open_mode::write)
-			&& ec.ec == boost::system::errc::no_such_file_or_directory)
+		
+		bool should_retry = (mode & aux::open_mode::write)
+			&& ec.ec == boost::system::errc::no_such_file_or_directory;
+#ifdef _WIN32
+		// this is a workaround for improper handling of files on windows shared drives.
+		// if the directory on a shared drive does not exist, 
+		// windows returns ERROR_IO_DEVICE instead of ERROR_FILE_NOT_FOUND
+		should_retry |= (ec.ec.value() == ERROR_IO_DEVICE);
```
arvidn, line 913: this may set `should_retry` even in the case where the `write` flag is not set in `mode`, which doesn't seem correct.
Additionally, it's not safe to *just* check the error code, you need to match the error category as well. Something like this:

```
ec == error_code(ERROR_IO_DEVICE, system_category())
```

If you think it makes sense to break out the condition into a bool rather than placing it directly in the if-condition, that bool should be made `const` as well.

```
bool const should_retry = ... ;
```

src/mmap_storage.cpp
```diff
@@ -903,8 +903,17 @@ error_code translate_error(std::system_error const& err, bool const write)
...
+#endif
+
+		if (should_retry)
```
arvidn, line 916: perhaps something like this:

```
    if ((mode & aux::open_mode::write)
        && (ec.ec == boost::system::errc::no_such_file_or_directory
#ifdef _WIN32
            || ec == error_code(ERROR_IO_DEVICE, system_error())
#endif
    ))
```

arvidn:
I suspect this is a problem in `RC_1_2` as well. it might make sense to back-port

src/mmap_storage.cpp
FletcherD: That looks better. This seems to work for me:
`		if ((mode & aux::open_mode::write)
			&& (ec.ec == boost::system::errc::no_such_file_or_directory
#ifdef _WIN32
				// this is a workaround for improper handling of files on windows shared drives.
				// if the directory on a shared drive does not exist, 
				// windows returns ERROR_IO_DEVICE instead of ERROR_FILE_NOT_FOUND
				|| ec.ec == error_code(ERROR_IO_DEVICE, boost::asio::error::system_category)
#endif
		))`

src/mmap_storage.cpp
```diff
@@ -904,7 +904,14 @@ error_code translate_error(std::system_error const& err, bool const write)
 
 		boost::optional<aux::file_view> h = open_file_impl(sett, file, mode, ec);
 		if ((mode & aux::open_mode::write)
-			&& ec.ec == boost::system::errc::no_such_file_or_directory)
+			&& (ec.ec == boost::system::errc::no_such_file_or_directory
+#ifdef _WIN32
+				// this is a workaround for improper handling of files on windows shared drives.
+				// if the directory on a shared drive does not exist, 
+				// windows returns ERROR_IO_DEVICE instead of ERROR_FILE_NOT_FOUND
+				|| ec.ec == error_code(ERROR_IO_DEVICE, boost::asio::error::system_category)
```
arvidn, line 912: I would prefer if this was unqualified, to pick up `libtorrent::system_category()`, and actually calling the function.  `system_category()`

I'm not sure why this compiles at all.

arvidn:
I would think this same problem exist in `posix_storage.cpp` as well, right?

src/mmap_storage.cpp
FletcherD: There must be something weird with my compile configuration such as boost version, because this compiles for me but yours didn't.

If it's ok with you why don't you implement it in a way you prefer, because I am not really knowledgeable enough about the code base to do it in a way I'm confident in

FletcherD:
I can attempt to check whether this exists on a POSIX machine.

arvidn:
actually, the posix disk backend is used on windows as well, in 32 bit mode. if you build libtorrent with `address-model=32` it will be used.

src/mmap_storage.cpp
arvidn: how does this look? https://github.com/arvidn/libtorrent/pull/6718

arvidn:
thanks!

PR #6709, opened 2022-02-03
Fix #6683 macOS python binding build

### LINKED ISSUE #6683: fix python build on MacOS CI
@AllSeeingEyeTolledEweSew I'm trying to figure out what the best solution to this failure is. But I'm not familiar enough with python to judge what the best fix would be:

https://github.com/arvidn/libtorrent/runs/4907293463?check_suite_focus=true#step:12:630

> 2022-01-22T14:04:28.2245340Z TEST FAILED: /Users/runner/Library/Python/3.9/lib/python3.9/site-packages/ does NOT support .pth files
> 2022-01-22T14:04:28.2245790Z bad install directory or PYTHONPATH
> 2022-01-22T14:04:28.2245930Z 
> 2022-01-22T14:04:28.2246170Z You are attempting to install a package to a directory that is not
> 2022-01-22T14:04:28.2246490Z on PYTHONPATH and which Python does not read ".pth" files from.  The
> 2022-01-22T14:04:28.2246940Z installation directory you specified (via --install-dir, --prefix, or
> 2022-01-22T14:04:28.2247230Z the distutils default setting) was:
> 2022-01-22T14:04:28.2247390Z 
> 2022-01-22T14:04:28.2247740Z     /Users/runner/Library/Python/3.9/lib/python3.9/site-packages/
> 2022-01-22T14:04:28.2247910Z 
> 2022-01-22T14:04:28.2248140Z and your PYTHONPATH environment variable currently contains:
> 2022-01-22T14:04:28.2248320Z 
> 2022-01-22T14:04:28.2248510Z     ''
> 2022-01-22T14:04:28.2248620Z 
> 2022-01-22T14:04:28.2248830Z Here are some of your options for correcting the problem:
> 2022-01-22T14:04:28.2248990Z 
> 2022-01-22T14:04:28.2249220Z * You can choose a different installation directory, i.e., one that is
> 2022-01-22T14:04:28.2249490Z   on PYTHONPATH or supports .pth files
> 2022-01-22T14:04:28.2249640Z 
> 2022-01-22T14:04:28.2249860Z * You can add the installation directory to the PYTHONPATH environment
> 2022-01-22T14:04:28.2250170Z   variable.  (It must then also be on PYTHONPATH whenever you run
> 2022-01-22T14:04:28.2250460Z   Python and want to use the package(s) you are installing.)
> 2022-01-22T14:04:28.2250630Z 
> 2022-01-22T14:04:28.2250850Z * You can set up the installation directory to support ".pth" files by
> 2022-01-22T14:04:28.2251130Z   using one of the approaches described here:
> 2022-01-22T14:04:28.2251280Z 
> 2022-01-22T14:04:28.2251750Z   https://setuptools.pypa.io/en/latest/deprecated/easy_install.html#custom-installation-locations
> 2022-01-22T14:04:28.2251980Z 
> 2022-01-22T14:04:28.2252080Z 
> 2022-01-22T14:04:28.2252310Z Please make the appropriate changes for your system and try again.

Trying to reproduce it locally, I either get a failure that `pyconfig.h` cannot be found. The system-installed `python3` that comes with MacOS doesn't seem to come with header files. When I try to run it with homebrew's `python3` it just works.

Perhaps MacOS could be split out as its own build step and explicitly run it with homebrew.

Samfun75:
Install libtorrent in Homebrew's python site-packages by passing the location with `--install-lib` when running `setup.py`

Runs on branch:
- [RC_1_2 run](https://github.com/Samfun75/libtorrent/actions/runs/1789100101)
- [RC_2_0 run](https://github.com/Samfun75/libtorrent/actions/runs/1789100117)

Fixes #6683

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -106,9 +106,7 @@ jobs:
     - name: build/install (MacOS)
       if: runner.os == 'macOS'
       run: |
-        # Add Homebrew's python site-packages to PYTHONPATH and
-        # install to it. no need for --user and --prefix
-        export PYTHONPATH=$(brew --prefix)/lib/python3.9/site-packages
+        # Install to Homebrew's python site-packages. no need for --user and --prefix
         cd bindings/python
         python3 setup.py build_ext install --install-lib $(brew --prefix)/lib/python3.9/site-packages
 
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -106,6 +106,8 @@ jobs:
     - name: build/install (MacOS)
       if: runner.os == 'macOS'
       run: |
+        # Add Homebrew's python site-packages to PYTHONPATH and
+        # install to it. no need for --user and --prefix
         export PYTHONPATH=$(brew --prefix)/lib/python3.9/site-packages
         cd bindings/python
         python3 setup.py build_ext install --install-lib $(brew --prefix)/lib/python3.9/site-packages
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -97,19 +97,21 @@ jobs:
         cd bindings\python
         python test.py
 
-    - name: build/install
-      if: runner.os != 'Windows'
+    - name: build/install (Linux)
+      if: runner.os == 'Linux'
       run: |
         cd bindings/python
-        # Homebrew's python "framework" sets a prefix via distutils config.
-        # --prefix conflicts with --user, so null out prefix so we have one
-        # command that works everywhere
         python3 setup.py build_ext install --user --prefix=
 
-    # TODO: we currently don't run the tests on MacOS, because something fails
-    # to install correctly. It needs to be resolved
+    - name: build/install (MacOS)
+      if: runner.os == 'macOS'
+      run: |
+        export PYTHONPATH=$(brew --prefix)/lib/python3.9/site-packages
+        cd bindings/python
+        python3 setup.py build_ext install --install-lib $(brew --prefix)/lib/python3.9/site-packages
+
     - name: tests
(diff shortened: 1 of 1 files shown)


.github/workflows/python.yml
```diff
@@ -97,19 +97,21 @@ jobs:
...
         python3 setup.py build_ext install --user --prefix=
 
-    # TODO: we currently don't run the tests on MacOS, because something fails
-    # to install correctly. It needs to be resolved
+    - name: build/install (MacOS)
+      if: runner.os == 'macOS'
+      run: |
+        # Install to Homebrew's python site-packages. no need for --user and --prefix
+        cd bindings/python
+        python3 setup.py build_ext install --install-lib $(brew --prefix)/lib/python3.9/site-packages
```
arvidn, line 111: there's no way to ask python for this directory, rather than hard coding `python3.9`? This will break every time python is updated.

arvidn:
thanks for your contribution!

.github/workflows/python.yml
Samfun75: Sorry for late reply, we could write a python 1 liner for it but idk if that would behave always the same on Mac (I don't own a Mac) hence why we are here in the first place. Or put the 3.9 on a job env and use it with brew install and here so always only update the env and everything should be fine... I am no expert so someone else could definitely weigh in on this.

arvidn: ok, thanks! Maybe @AllSeeingEyeTolledEweSew has some ideas

Samfun75: Yeah, Let's see what he can conjure up. let me know if I could assist more.

PR #6711, opened 2022-02-04
set lower memory priority for disk I/O threads on windows

arvidn:


jagannatharjun:
still happens, while checking 150g torrent
image

> Do you experience any problems, apart from task manager showing all your memory being used?

the system becomes a bit slow like chrome won't retain cache

glassez:
>set lower memory priority for disk I/O threads 

Looks like that's not what you need. IIRC, it only make the pages of thread "less important" than other pages of process so the memory manager will prefer to delete pages of this thread when it is necessary to trim the working set of the process. It seems that we will still have to somehow limit the maximum working set (at least provide such a setting). And it seems that this is not the task of libtorrent (or some other libraries), but of the application itself.
Ref: https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-setprocessworkingsetsizeex
@Chocobo1, @sledgehammer999, ping!

glassez:
> It seems that we will still have to somehow limit the maximum working set (at least provide such a setting). And it seems that this is not the task of libtorrent (or some other libraries), but of the application itself.

I'll try to figure it out.
But this PR still makes sense, so that when trimming the process Working Set, the memory manager prefers to delete pages of IO threads that are less priority than the main libtorrent thread, GUI thread, etc.

arvidn:
yeah, assuming most processes have the privilege to set the desired working set size, then the question is; what's a good value? It surely depends on the user.

Chocobo1:
>But this PR still makes sense, so that when trimming the process Working Set, the memory manager prefers to delete pages of IO threads that are less priority than the main libtorrent thread, GUI thread, etc.

qbt currently has its own code to set it (process wide). If thread handles can be obtained from libtorrent, then users can fine tune it. Or another way, perhaps libtorrent can provide setting for it?

>yeah, assuming most processes have the privilege to set the desired working set size, then the question is; what's a good value? It surely depends on the user.

I think for applications in qbt, there can be a special value `auto` (like the removed disk cache size setting from RC_1_2) and otherwise let user decide for themselves (in MB units).

glassez:
> let user decide for themselves (in MB units).

Using a fraction (as a percentage) of the total amount of physical memory looks better to me.

PR #6722, opened 2022-02-06
CMake: prefer `-pthread` flag

r-value:
* Moved from qbittorrent/qBittorrent#16389.

ref: https://cmake.org/cmake/help/latest/module/FindThreads.html#variable:THREADS_PREFER_PTHREAD_FLAG

> Use of both the imported target as well as this switch is highly recommended for new code.

Note that this change also fixes build on RISC-V architecture, which needs `-pthread` flag for libpthread to function correctly.

See also: [Stack Overflow: Difference between -pthread and -lpthread](https://stackoverflow.com/questions/23250863/difference-between-pthread-and-lpthread-while-compiling)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -476,6 +476,7 @@ list(TRANSFORM try_signal_sources PREPEND "deps/try_signal/")
 feature_option(BUILD_SHARED_LIBS "build libtorrent as a shared library" ON)
 feature_option(static_runtime "build libtorrent with static runtime" OFF)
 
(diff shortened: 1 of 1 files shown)


arvidn (approved):

Chocobo1:
I think RC_1_2 would benefit from this patch too.

arvidn:
https://github.com/arvidn/libtorrent/pull/6728

zywo:
`-l-pthread` is added to `libtorrent-rasterbar.pc` instead of `-lpthread`

PR #6732, opened 2022-02-10
amend move and file priority documentation

arvidn:


cheater:
Thank you. I suggest adding the following warning that further explains the disk space usage:

Moved sparse files that have a singular part downloaded at their end will take up their full file size on the destination file system, instead of just taking up space necessary to store the part. This almost always always happens if you only elect to download a single file, because the preceding file's last part will be necessary for the file that you elected to download.

arvidn:
"Almost always" would be more accurate as either "almost never" or "when a file is downloaded to before setting its priority to 0". Don't you think?

PR #6759, opened 2022-02-27
Always send a User-Agent field in HTTP tracker requests

arvidn:
even when it's set to an empty string. It appears some trackers require this field and will fail with 'Forbidden' otherwise.

### DIFF
--- a/src/http_tracker_connection.cpp
+++ b/src/http_tracker_connection.cpp
@@ -233,8 +233,11 @@ namespace libtorrent {
 		// in anonymous mode we omit the user agent to mitigate fingerprinting of
 		// the client. Private torrents is an exception because some private
 		// trackers may require the user agent
-		std::string const user_agent = settings.get_bool(settings_pack::anonymous_mode)
-			&& !tracker_req().private_torrent ? "" : settings.get_str(settings_pack::user_agent);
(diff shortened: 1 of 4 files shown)


src/http_tracker_connection.cpp
```diff
@@ -233,8 +233,11 @@ namespace libtorrent {
 		// in anonymous mode we omit the user agent to mitigate fingerprinting of
 		// the client. Private torrents is an exception because some private
 		// trackers may require the user agent
```
an0n666, line 235: Please update these documentations and the ones found in your website.

an0n666:
‘Trackers will only be used if they are using a proxy server.’

I do not understand the meaning of this. I tried without using any proxy and trackers are still being used.

arvidn:
@an0n666 I fixed the documentation. I don't think the intention of anonymous mode ever was to "force tracker" and disallow incoming connections. That's not what it ever did at least. I suspect there was some conflation happening when `force_proxy` was removed.

an0n666:
@Chocobo1 @xavier2k6 @thalieht
If possible please update the current qBt documentation regarding anonymous mode as the current one has many misinformations.
https://github.com/qbittorrent/qBittorrent/wiki/Anonymous-Mode
I would've updated it myself but seems like the edit option is missing in qBt wiki.
Sorry for the ping!

PR #6775, opened 2022-03-11
Add a pre-commit autoupdate cron job

AllSeeingEyeTolledEweSew:
This will auto-generate PRs to update our pre-commit hooks.

On my fork, it auto-generated this PR:
 - https://github.com/AllSeeingEyeTolledEweSew/libtorrent/pull/48

I believe the scheduled workflow will only run once this change is merged to `RC_2_0`.

### DIFF
--- a/.github/workflows/pre_commit_autoupdate.yml
+++ b/.github/workflows/pre_commit_autoupdate.yml
@@ -1,0 +1,26 @@
+name: Pre-commit auto-update
+
+on:
+  # every day at midnight
+  schedule:
+    - cron: "0 0 * * *"
+  # on demand
+  workflow_dispatch:
+
+jobs:
+  auto-update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-python@v2
+
+      - uses: browniebroke/pre-commit-autoupdate-action@main
+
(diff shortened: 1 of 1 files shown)


AllSeeingEyeTolledEweSew:
bump

arvidn:
what does it mean to update a "pre commit hook"?
is it about modifying the pre commit config to specify new version numbers?

AllSeeingEyeTolledEweSew:
Yes. It's analogous to updating pinned versions of github actions, e.g. `actions/setup-python@v3` to `actions/setup-python@v4`.

In practice the most common side effect of these updates is new complaints from mypy, where there was a legitimate typing error in our code but mypy only notices it on a version bump.

AllSeeingEyeTolledEweSew:
@arvidn FYI: libtorrent's pre-commit config currently breaks if run with python 3.7+. Currently the CI job uses python 3.6 which happens to work, but that version is EOL and should be bumped.

We need a `pre-commit autoupdate` run to fix it. Or merge this PR, and it will automatically generate PRs with `pre-commit autoupdate`.

IMO, this kind of thing is exactly the reason to auto-update dependencies, even for CI. It seems much better to make incremental updates as parts of the environment change over time, rather than try to update a bunch of stuff at once when something breaks.

Issue #6531, opened 2021-10-19
Make python bindings source more consistent across branches


AllSeeingEyeTolledEweSew:
`bindings/python/**` has become quite inconsistent across branches, and getting worse. The differences are mainly build/test infrastructure and bugfixes. The existing inconsistencies are making it harder to do *new* changes consistently, due to messy merges (accruing technical debt).

I'd like to do a little cleanup to get `bindings/python/**` consistent across branches, within reason.

This is mainly motivated by #6532. But more than that, it looks like `RC_1_2` and `RC_2_0` will be supported for the foreseeable future, so it seems worthy to backport fixes and infrastructure improvements.

It might also be worthy to clean up non-python parts of the source, but I don't understand C++ sufficiently to do this. :)

I don't want to backport functionality (except type stubs), or expose anything in the python bindings not already exposed.

 * [ ] Backport pre-commit infrastructure
   * #6505
     * #6518
 * [ ] Backport all non-feature-related changes
   * #6509
   * #6510
   * #6511
 * [ ] Backport `setup.py` to be more textually consistent
 * [ ] Backport the python bindings tests (`bindings/python/tests/**` and related)
 * [ ] Backport all python bindings fixes (all above tests should pass on `RC_1_2`)
 * [ ] Optional: #6532

AllSeeingEyeTolledEweSew:
I'm still working on this. Let's see if this bump reopens the issue

PR #6776, opened 2022-03-11
Backport setup.py formatting from master (RC_1_2 only)

AllSeeingEyeTolledEweSew:
Progress toward #6531 

The intent is to backport fixes, formatting and type annotations. The larger goal is to backport the test suite and type stubs from `master`, as well as make it easier to do other backports, and fixes that should be merged on all branches.

I copied `setup.py` from `master`, and manually added the following logic, currently present in `RC_1_2`:
1. Configure the build based on autotools output
2. Always use `cxxstd=11` on mac
3. Write `project-config.jam` to cwd, rather than use `b2 --project-config=<tempfile>`

AIUI, 1 and 2 **must not** be merged to `RC_2_0`, and 3 **must** be merged to `RC_2_0`.

I created a separate PR to backport `setup.py` from `master` to `RC_2_0`.

This PR's diff is complex. It's also helpful to diff this version of `setup.py` against `master`. In theory I should be able to provide a github link to do this [comparison across forks](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/viewing-and-comparing-commits/comparing-commits#comparing-across-forks), but this seems broken, and I can't make the appropriate diff appear.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -46,7 +46,6 @@ repos:
             bindings/python/client.py|
             bindings/python/dummy_data.py|
             bindings/python/make_torrent.py|
-            bindings/python/setup.py|
             bindings/python/simple_client.py|
             bindings/python/test.py|
             docs/gen_reference_doc.py|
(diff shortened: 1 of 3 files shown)


AllSeeingEyeTolledEweSew:
Diff of `setup.py` from `master` to this PR: https://gist.github.com/AllSeeingEyeTolledEweSew/37d85e5a2937ae622063c5846ed3c20e

PR #6795, opened 2022-04-05
cmake: refine atomics check

### LINKED ISSUE #6794: Linking failure on RISC-V with GCC 11.2.1
libtorrent version (or branch): `2.0.5`

Platform/Architecture: RISC-V, Gentoo Linux, with `lp64d systemd` profile

Compiler and compiler version: `gcc (Gentoo 11.2.1_p20220115 p4) 11.2.1 20220115`

Encountered the problem while trying to compile the qBittorrent 4.4.2 ebuild with latest portage tree. `libtorrent-rasterbar` was built successfully, but during linking of qBittorrent it would fail due to missing `__atomic_*` symbol from `libtorrent`.

```

/usr/lib/gcc/riscv64-unknown-linux-gnu/11.2.1/../../../../riscv64-unknown-linux-gnu/bin/ld: /usr/lib64/libtorrent-rasterbar.
so.2.0.5: undefined reference to `__atomic_exchange_1'                                                                      
distcc[2276441] ERROR: compile (null) on localhost failed
ninja: build stopped: subcommand failed. 

```

AFAIK, this is because CMake script fails to detect missing subword atomic, as RISC-V doesn't have 1/2 byte atomic instruction.

- GCC should provide subword atomic inling IMO, but currently they don't.
- This was bypassed by prefer `pthread`, hints linking to `latomic` previously (#6722). However, after glibc 2.34, `-pthread` is nolonger needed to use `pthread` libraries, hence no more linking to `latomic`.

Since replying on `-pthread` for atomic is sementically incorrect, plus we already have a detection mechanism here, a possible fix could be enhance the detection.

Build log:
[libtorrent-rasterbar-2.0.5-r1.log](https://github.com/arvidn/libtorrent/files/8411284/libtorrent-rasterbar-2.0.5-r1_20220403-153624.log)
[qbittorrent-4.4.2.log](https://github.com/arvidn/libtorrent/files/8411287/qbittorrent-4.4.2_20220403-161029.log)

infiWang:
Check for 8-bit atomic operation.
Fixes linking on RISC-V systems.

Closes #6794


arvidn:
sorry for the delay. does the Risc-V toolchain also require explicitly linking against libatomic to get access to standard library facilities? I was under the impression this was a bug in some version of the Raspberry PI toolchains.

infiWang:
> does the Risc-V toolchain also require explicitly linking against libatomic to get access to standard library facilities?

Yes, indeed.
It was concealed by `-pthread`, which implies linking to libatomic currently. However, keep in mind that this behavior is undocumented and may change in the future. It's already troublesome with glibc 2.34+ and CMake as the related issue points out.

arvidn (approved):

xavier2k6:
@arvidn Any ETA for `2.0.6 / 1.2.16` release?

arvidn:
soon. I'm aiming for this week

xavier2k6:
> soon. I'm aiming for this week

Thanks - I was just asking out of curiosity & not necessity.....

infiWang:
Sorry for interrupting, but I just realized that RC_1_2 would also gain from this.

arvidn:
https://github.com/arvidn/libtorrent/pull/6835

infiWang:
> #6835

Much appreciated.

PR #6817, opened 2022-04-17
fix bdecode support for large strings (>= 100 MB)

arvidn:


include/libtorrent/bdecode.hpp
```diff
@@ -166,15 +166,18 @@ void escape_string(std::string& ret, char const* str, int len);
 struct bdecode_token
 {
 	// the node with type 'end' is a logical node, pointing to the end
-	// of the bencoded buffer.
+	// of the bencoded buffer. THe `long_string` type is for strings that are so
```
glassez, line 169:
```suggestion
	// of the bencoded buffer. The `long_string` type is for strings that are so
```

include/libtorrent/bdecode.hpp
```diff
@@ -166,15 +166,18 @@ void escape_string(std::string& ret, char const* str, int len);
...
 	enum type_t : std::uint8_t
-	{ none, dict, list, string, integer, end };
+	{ none, dict, list, string, integer, long_string, end };
 
 	enum limits_t
 	{
 		max_offset = (1 << 29) - 1,
 		max_next_item = (1 << 29) - 1,
-		max_header = (1 << 3) - 1
+		short_string_max_header = (1 << 3) - 1 + 2,
```
glassez, line 179:
```suggestion
		short_string_max_header = (1 << 3) + 1,
```

include/libtorrent/bdecode.hpp
```diff
@@ -166,15 +166,18 @@ void escape_string(std::string& ret, char const* str, int len);
...
+		long_string_max_header = 8 + (1 << 3) - 1 + 2
```
glassez, line 180:
```suggestion
		long_string_max_header = 8 + (1 << 3) + 1
```

include/libtorrent/bdecode.hpp
```diff
@@ -190,25 +193,35 @@ struct bdecode_token
...
-	bdecode_token(std::ptrdiff_t off, std::uint32_t next
-		, bdecode_token::type_t t, std::uint8_t header_size = 0)
+	bdecode_token(std::ptrdiff_t const off, std::uint32_t const next
+		, bdecode_token::type_t const t, std::uint32_t const header_size = 0)
 		: offset(std::uint32_t(off))
 		, type(t)
 		, next_item(next)
-		, header(type == string ? std::uint32_t(header_size - 2) : 0)
+		, header(type == string ? std::uint32_t(header_size - 2)
+			: type == long_string ? std::uint32_t(header_size - 8 - 2) : 0)
```
glassez, line 202:
```suggestion
			: type == long_string ? std::uint32_t(header_size - 10) : 0)
```

include/libtorrent/bdecode.hpp
```diff
@@ -190,25 +193,35 @@ struct bdecode_token
...
 	}
 
-	int start_offset() const { TORRENT_ASSERT(type == string); return int(header) + 2; }
+	int start_offset() const
+	{
+		TORRENT_ASSERT(type == string || type == long_string);
+		if (type == string)
+			return int(header) + 2;
+		else
+			return int(header) + 8 + 2;
```
glassez, line 223:
```suggestion
			return int(header) + 10;
```

src/bdecode.cpp
```diff
@@ -933,13 +941,25 @@ namespace detail {
...
+
 					// the bdecode_token only has 8 bits to keep the header size
 					// in. If it overflows, fail!
-					if (start - str_start - 2 > detail::bdecode_token::max_header)
+					if (header > detail::bdecode_token::long_string_max_header)
 						TORRENT_FAIL_BDECODE(bdecode_errors::limit_exceeded);
 
-					ret.m_tokens.push_back({str_start - orig_start
-						, 1, bdecode_token::string, std::uint8_t(start - str_start)});
+					if (header <= detail::bdecode_token::short_string_max_header)
```
glassez, line 952:
```suggestion
					bdecode_token::type_t type = (header <= detail::bdecode_token::short_string_max_header) ? bdecode_token::string : bdecode_token::long_string;
					ret.m_tokens.push_back({str_start - orig_start
							, 1, type, std::uint32_t(header)});
```

include/libtorrent/bdecode.hpp
arvidn: the 8 and the 2 are there for different reasons. I don't want to obfuscate how this offset is computed

include/libtorrent/bdecode.hpp
arvidn: I changed (in my mind, I simplified) the check against these upper limits, to make them also include the implied + 2

arvidn: see this check: https://github.com/arvidn/libtorrent/pull/6817/files#diff-e0c2c30bf10517b10c7e267b7717b23cbad276da4c395ab4eeaf6ba46a2f00b6R949

glassez: -1 + 2 = + 1

include/libtorrent/bdecode.hpp
glassez: > the 8 and the 2 are there for different reasons. I don't want to obfuscate how this offset is computed

They have little meaning for reader without some explanation. But having explanation they can be easily "obfuscated".
Otherwise you could use some named constants to make code self descriptive.

include/libtorrent/bdecode.hpp
arvidn: ah, this is the same reason as the other comment. the largest value stored in 3 bits is one part of the expression (`(1 << 3) - 2`), and + 2 is separate.

glassez:
I wonder if we will face exceeding other limits in the near future? For example, max offset?

master255:
@arvidn Is it to support large torrent files?

arvidn:
> @arvidn Is it to support large torrent files?

It's to support bencoded files with large strings, which can be found in resume data for v2 torrents for very large files. Specifically while they are being downloaded.

arvidn:
good poinr

PR #6824, opened 2022-04-19
Improve compliant with C++20

Chocobo1:
Suppresses some minor compiler warnings when compiling in C++20 mode.
Compiler is gcc version 11.2.0.


Chocobo1:
Great, seems those changes aren't welcomed in earlier C++ modes. I give up.

PR #6840, opened 2022-05-03
removed unused variable in torrent.cpp

aldenml:


arvidn:
I would like to look at the history of this to ensure the bug isn't that it should be used somewhere instead

arvidn:
it looks like this was orphaned when I introduced `file_piece_range_inclusive()`

arvidn (approved):

PR #6847, opened 2022-05-07
Back-port cmake pkg-config fix

arvidn:
@Kolcha how does this look?
What's an easy way to test this?


Kolcha:
looks fine (the same as for 2.0, very likely that files were not changed for a long time)
tested these changes with 1.2.x - pkg-config fine is generated correctly (and incorrectly without these changes)

tested with
```bash
mkdir build && cd $_
cmake -G Ninja -D CMAKE_BUILD_TYPE=Release -D CMAKE_INSTALL_PREFIX=/usr ..
ninja
DESTDIR=/tmp/lt ninja install/strip
```
and just looked into generated .pc file in install directory (`/tmp/lt/usr/lib/x86_64-linux-gnu/pkgconfig/libtorrent-rasterbar.pc` in my case)

PR #6850, opened 2022-05-09
add settings_for_path() utility function

arvidn:
to optimize disk I/O settings based on the save path. This patch looks at these properties of the drive of the specified path:

1. Is it a spinning disk or SSD (windows boils this down to, is there a penalty to seek)
2. If it's an SSD, is it DAX? (which is really where memory mapped files would shine)
3. Is it a network share (where memory mapped files probably won't to so well)

This is still work in progress and I'm interested in feedback.

- [ ] python binding
- [ ] consider feasibility to also hint on what disk I/O back-end to use
- [ ] consider feasibility to also provide per-torrent settings

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,7 @@ set(libtorrent_include_files
 	session_stats.hpp
 	session_status.hpp
 	session_types.hpp
+	settings_for_path.hpp
 	settings_pack.hpp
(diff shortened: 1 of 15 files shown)


glassez:
@arvidn 
IIRC, you want to determine drive types and use different disk IO backends for different drives?

PR #6851, opened 2022-05-09
suppress tracker_error_alert for announce_skipped to reduce alert spam

arvidn:


arvidn:
I don't think this should land

glassez:
"to reduce alert spam" you could provide compound alerts in cases when several identical alerts are posted obviously sequentially.
E.g. in `announce_with_tracker()` you traverse through all the trackers/endpoints and post `tracker_announce_alert` for several of them within the same loop. It would be very nice to have single `tracker_announce_alert` with aggregated info. Even single `tracker_announce_alert` per `announce_entry` would reduce alert spam. As a bonus we will know about the fact that endpoint list of `announce_entry` has refreshed.

PR #6853, opened 2022-05-09
improve backwards compatibility with loading magnet link resume files

arvidn:


thewb:
my issue was i couldn't just add a magnet link, last guy's fix worked for me. ill try this.

PR #6855, opened 2022-05-09
Fix V1004 The 't' pointer was used unsafely after it was verified against nullptr

pavel-pimenov:


src/peer_connection.cpp
```diff
@@ -4760,10 +4760,11 @@ namespace {
 		INVARIANT_CHECK;
 
 		auto t = m_torrent.lock();
-
+		if (t)
+		{
 		std::uint8_t warning = 0;
 		// drain the IP overhead from the bandwidth limiters
-		if (m_settings.get_bool(settings_pack::rate_limit_ip_overhead) && t)
+			if (m_settings.get_bool(settings_pack::rate_limit_ip_overhead))
```
glassez, line 4767: I see no point in this change. Besides, you broke the indentations.

arvidn:
I don't understand how the existing code is using `t` unsafely. do you?

pavel-pimenov:
crash t->alerts()
if warning != 0 and t == nullptr

https://github.com/arvidn/libtorrent/blob/master/src/peer_connection.cpp#L4776

arvidn:
but warning is initialized to 0. It can never be != 0 (unless `t` is set)

arvidn:
did you experience or record a crash at `t->alert()`?

pavel-pimenov:
> did you experience or record a crash at `t->alert()`?

no. report static analyser PVS Studio (false positive)

Alternative fix (small optimisation):
<pre>
- if (m_settings.get_bool(settings_pack::rate_limit_ip_overhead) && t)
+ if (t && m_settings.get_bool(settings_pack::rate_limit_ip_overhead))
</pre>

Issue #6198, opened 2021-05-10
Python bindings: proposal for enhanced exception translation


AllSeeingEyeTolledEweSew:
# Current state

Currently the python bindings translate all exceptions to `RuntimeError`.

The only way to apply special handling to error conditions is to check the exception message. This makes code awkward, verbose and error-prone.

In particular: in concurrent usage of libtorrent, the `invalid torrent handle used` error can happen *on most lines of code*, so special error handling is essential.

# Desired state

The python bindings should raise exceptions from a meaningful type hierarchy.

# Proposal

I propose the following exception type hierarchy:
```python
Error  # subclass of RuntimeError
 +-- LibtorrentError
 |    +-- DuplicateTorrentError
 |    +-- InvalidTorrentHandleError
 +-- BDecodeError
 +-- UPNPError
 +-- SOCKSError
 +-- I2PError
_OSError  # subclass of builtin OSError and RuntimeError
 +-- _BlockingIOError    # subclass of builtin BlockingIOError
 +-- _ChildProcessError  # subclass of builtin ChildProcessError
 +-- _ConnectionError    # subclass of builtin ConnectionError
 |    +-- _BrokenPipeError         # subclass of builtin BrokenPipeError
 |    +-- _ConnectionAbortedError  # subclass of builtin ConnectionAbortedError
 |    +-- _ConnectionRefusedError  # subclass of builtin ConnectionRefusedError
 |    +-- _ConnectionResetError    # subclass of builtin ConnectionResetError
 +-- _FileExistsError      # subclass of builtin FileExistsError
 +-- _FileNotFoundError    # subclass of builtin FileNotFoundError
 +-- _InterruptedError     # subclass of builtin InterruptedError
 +-- _IsADirectoryError    # subclass of builtin IsADirectoryError
 +-- _NotADirectoryError   # subclass of builtin NotADirectoryError
 +-- _PermissionError      # subclass of builtin PermissionError
 +-- _ProcessLookupError   # subclass of builtin ProcessLookupError
 +-- _TimeoutError         # subclass of builtin TimeoutError
 +-- CanceledError        # no matching builtin exception
```

In a future version, we should remove `RuntimeError` from the exception ancestry.

The `_`-.prefixed `_OSError` tree is intended as a deprecated bridge to `RuntimeError`.

The exceptions should have the following features:
* Each exception type should correspond to errors of a particular `error_category`.
* `Error` and its subclasses should have a `value` attribute with the error value.
[...]

Issue #5154, opened 2020-09-15
Feature request: heap-allocated alerts


AllSeeingEyeTolledEweSew:
## TL;DR
I'd like to request that alerts be allocated on the heap, or have the option to be. This is to support concurrency, simplicity and safety of client code.

I assume the main advantage of the specialized memory allocation is speed of the main libtorrent thread. I'd like to hear if there are any others.

I have a lot of thoughts on why the specialized alert memory management is a disadvantage. Apologies for the wall of text; it's a significant change so I thought it important to be thorough.

## Argument summary
* Currently, responsive apps _shouldn't_ make blocking calls while handling alerts (so the next `pop_alerts()` can be called ASAP). This can lead to awkward workarounds in code trying to avoid calls to `torrent_handle::status()` etc.
* One natural pattern is to have multiple tasks (threads, coroutines, etc) each processing the alert stream. With current memory management, this pattern is difficult to build, and has caveats.
* The main alternative to the task pattern is an app with limited or no concurrency. This makes simple use cases complex, and limits scalability and flexibility.
* Apps are incentivized to restrict the `alert_mask` as much as possible, which should minimize the allocations done in the libtorrent thread
* It would be nice to eliminate hard crashes via invalid memory access. It's possible to hold references to alert-managed memory in really surprising ways.

I'll try to motivate my request by explaining my use case, and exploring the limits of how I can use libtorrent to implement my use case, and how those limits relate to alert memory management.

## My use case
The goal of my project [tvaf](https://github.com/AllSeeingEyeTolledEweSew/tvaf) is to provide access to a library of torrents as HTTP resources, downloading data on the fly. It must be highly responsive and concurrent.

This leads to a pretty complex workflow:
 * When a request comes in, look up the torrent by infohash
 * If it's not found, add it to the session
 * Prioritize different pieces of the torrent, according to all active requests
 * As pieces are downloaded, vend them via HTTP
 * Ensure errors are passed through
 * When all requests are finished, optionally clean up the torrent
 * Meanwhile, still function as a normal torrent client, allowing the user to add/remove torrents at any point

Keeping in mind that new requests may come in or be canceled at any point in the workflow.
## Event loop vs iterators
Every libtorrent-based app I can find uses a single-threaded event loop which fires a set of callbacks on user action or alert from libtorrent.

In my experience, single-threaded event loop apps either
 * are tightly constrained in scope and complexity,
 * have some blocking calls or long-running handlers which limit performance, or
[...]

PR #6858, opened 2022-05-11
update stale bot config

arvidn:


AllSeeingEyeTolledEweSew:
This doesn't address my largest complaint about stalebot, which is that it has apparent bugs, and will close an issue despite activity. See #5154 and #6198

AllSeeingEyeTolledEweSew:
This doesn't address my largest complaint about stalebot, which is that it has apparent bugs, and will close an issue despite activity. See #5154 and #6198

.github/stale.yml
```diff
@@ -1,11 +1,12 @@
...
-daysUntilStale: 90
+daysUntilStale: 120
 # Number of days of inactivity before a stale issue is closed
-daysUntilClose: 20
+daysUntilClose: 30
 # Issues with these labels will never be considered stale
 exemptLabels:
   - pinned
   - security
+  - todo
```
AllSeeingEyeTolledEweSew, line 9: the effect of this is that I'll just prompt you to add `todo` to any issue I open

AllSeeingEyeTolledEweSew, line 9: the effect of this is that I'll just prompt you to add `todo` to any issue I open

arvidn: that's not an obvious problem to me.

arvidn:
I should research whether there are some configuration options of stale bot to improve or fix this. Perhaps it just has the wrong default.

.github/stale.yml
AllSeeingEyeTolledEweSew: it seems backwards that high quality reports should carry the burden of requesting the "good issue" label, arguing why they deserve it, knowing to do so in the first place, re-requesting it if you forget to apply the label when responding to the ticket, and dealing with the additional noise.

I make high quality reports and I want all the burden to be on low-quality reporters :)

summerqB:
428 issues closed by stale bot 😮

userdocs:
Most of my issues closed by stale bot have been generic build stuff I forgot to close or was never resolved but became obsolete from using newer versions and stuff so I can't say it caused me any drama but I think it should work like this:

Stale bot should message the issue and ask the op if they want to keep it open and if not can they please tag it can close it manually.

So basically interact with the OP act in response to that outcome.

Maybe after x amount of non responses default to a behaviour.

I think using it to encourage the OP manage the issue is the best first step. At least that way they wont just get their issue closed?

Maybe stale bot forces someone to tag the issue with a easily filterable tag before closing so relevant issues can be filtered when searching? Good for python binding specific issues for example.

Dead stuff with no response will get closed for lack of reply and

PR #6867, opened 2022-05-19
Handle all "fastresume rejected" cases consistently

glassez:
Issue a "fastresume rejected" alert in all the cases. Otherwise, it's damn hard to determine that an I/O error occurred precisely during "fastresume data" checking. Also, I don't see the point in completely discarding "fastresume data" (e.g. peers), but just continue processing it the same way as in other cases.


arvidn (approved):

glassez:
Will it be backported into RC_1_2?

PR #6869, opened 2022-05-20
openssl: Replace SHA* family by EVP_*

toooonyy:
Built against openssl 1.1.1o and 3.0.3, which are both compiled with `no-deprecated`.
https://github.com/toooonyy/libtorrent/commit/1bfc03bb0616aa851db58c4238d2f5a288ffe649 this change might lead to recession(memory leaks?), but so far my qbit seems normal.

### DIFF
--- a/include/libtorrent/aux_/hasher512.hpp
+++ b/include/libtorrent/aux_/hasher512.hpp
@@ -58,6 +58,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 	extern "C" {
 	#include <openssl/sha.h>
+	#include <openssl/evp.h>
 	}
 
 #else
@@ -122,7 +123,7 @@ namespace aux {
 #elif TORRENT_USE_CRYPTOAPI_SHA_512
 		aux::crypt_hash<CALG_SHA_512, PROV_RSA_AES> m_context;
 #elif defined TORRENT_USE_LIBCRYPTO
-		SHA512_CTX m_context;
+		EVP_MD_CTX *m_context = EVP_MD_CTX_new();
 #else
 		sha512_ctx m_context;
 #endif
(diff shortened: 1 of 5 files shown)


Chocobo1:
FYI, related work on master branch:
212cdfc9cde240cb6ca2164d4eeb2a381c30eba8
e7fd8ae2e3a41b8f8dc668f4f280a42381954c37

toooonyy:
> FYI, related work on master branch:
> [212cdfc](https://github.com/arvidn/libtorrent/commit/212cdfc9cde240cb6ca2164d4eeb2a381c30eba8)
> [e7fd8ae](https://github.com/arvidn/libtorrent/commit/e7fd8ae2e3a41b8f8dc668f4f280a42381954c37)

Ah I didn't know it's already done in the master branch, guess I'll close this PR wait and it merges into release.

arvidn:
master doesn't merge into `RC_2_0`. It might make sense to back-port this to `RC_2_0`

arvidn:
using the new API affects ABI and should use a separate define and separate inline namespace. The existing code needs to still be there to be backwards compatible with existing 2.0.x releases

toooonyy:
> using the new API affects ABI and should use a separate define and separate inline namespace. The existing code needs to still be there to be backwards compatible with existing 2.0.x releases
Screen Shot 2022-05-21 at 17 57 42

From the manual this `EVP*` function family is not that new? So unless this project still cares about the compatibility for `openssl 1.0.2`, this should work for both 1.1.1 and 3.0.*.

arvidn:
my concern is the ABI between a client of libtorrent and libtorrent itself.

libtorrent commits to maintaining ABI compatibility within major releases. e.g. a client built against libtorrent 2.0.0 should be able to link (at run-time) against any libtorrent 2.0.x version.
This change breaks that ABI. Therefore, in libtorrent 2.0.x, the new OpenSSL API needs to be treated as a separate build configuration, and preserve the previous one also.

This is similar to how there are different build configuration for selecting crypto library. This selection is then encoded in the inline namespace name to ensure that a client linking against an incompatible build of libtorrent fail hard rather than limps along with memory corruption.

arvidn:
`master` doesn't have this concern, because it will break ABI when it's released anyway.

PR #6872, opened 2022-05-21
V2 gcc49 c++1y

th0ma7:
Convert constexpr to const for gcc49 compatibility using c++1y


th0ma7:
Wrong PR sorry for that. closing.

PR #6873, opened 2022-05-21
Unique trackers

arvidn:
This patch enforces unique URLs for trackers. It also factors out the tracker list into its own class to make it clearer what its invariants are.

With a clearer separation of torrent.cop and tracker_list.cpp I could make the data structure more sophisticated without complexity going out of control.

The new data structure allocates the announce_entry objects in a pool allocator (instead of a vector) and inserts them into an intrusive linked list. This means their addresses are stable and can be referenced by an index mapping URL to announce entry.

This makes it more efficient to map a response from a tracker to its announce_entry. It also makes it more efficient to reorder trackers when they fail (to deprioritise) as well as prioritise (when preferring udp trackers).

As a result, handling torrents with a large number of trackers is more efficient with this patch. (But there's still room for improvements).

### DIFF
--- a/include/libtorrent/aux_/tracker_list.hpp
+++ b/include/libtorrent/aux_/tracker_list.hpp
@@ -1,0 +1,124 @@
+/*
+
+Copyright (c) 2022, Arvid Norberg
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in
+      the documentation and/or other materials provided with the distribution.
+    * Neither the name of the author nor the names of its
+      contributors may be used to endorse or promote products derived
+      from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+
+*/
+
+#ifndef TORRENT_TRACKER_LIST_HPP_INCLUDED
+#define TORRENT_TRACKER_LIST_HPP_INCLUDED
+
+#include <string>
+#include <unordered_map>
+
+#include "libtorrent/aux_/announce_entry.hpp"
+#include "libtorrent/announce_entry.hpp"
+#include "libtorrent/aux_/vector.hpp"
+#include "libtorrent/aux_/invariant_check.hpp"
+
+#include "libtorrent/aux_/disable_warnings_push.hpp"
+#include <boost/intrusive/list.hpp>
+#include <boost/pool/object_pool.hpp>
+#include "libtorrent/aux_/disable_warnings_pop.hpp"
+
+
+namespace libtorrent::aux {
+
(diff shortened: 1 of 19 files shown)


include/libtorrent/aux_/tracker_list.hpp
```diff
@@ -0,0 +1,115 @@
...
+	using trackers_t = aux::vector<aux::announce_entry>;
+	using iterator = typename trackers_t::iterator;
+	using const_iterator = typename trackers_t::const_iterator;
+
+	aux::announce_entry* find_tracker(std::string const& url);
+
+	// returns true if the tracker was added, and false if it was already
+	// in the tracker list (in which case the source was added to the
+	// entry in the list)
+	bool add_tracker(announce_entry const& ae);
```
glassez, line 58: I haven't had time to review it completely yet. But I remembered one thing that I noticed several times, but all the time I forgot to report. I wonder what is the use of having `announce_entry::source` field? In any case, it seems to be handled incorrectly. Due to the fact that it is not stored in the "fastresume data", it is reset to "magnet" for all trackers after the session is restarted.

arvidn: the only reason is to report it to the client. I would have expected it to say "resume_data" when you restart.

It's similar to reporting whether a torrent has received any incoming connections. It's ephemeral state to indicate whether things appear to be working right. Although, I believe the order in which peers are attempted to be connected to may be affected by which sources we've received them from.

glassez: > I would have expected it to say "resume_data" when you restart.

This will not make it more useful, since all trackers will have such a source after the first session restart.

arvidn: I was mixing it up with `source` on peers btw. Only trackers that were loaded from resume data will have the source be "magnet", not trackers added manually, nor trackers received via tracker-exchange (but that's not a thing anymore).

Just like whether you've received incoming connections or not isn't saved, this state isn't meant to be saved either. Similar to the source field for peers.

Granted, since tracker-exchange isn't a thing anymore, it's not terribly useful. And since magnet links are deconstructed into the resume data class (`add_torrent_params`), there's no distinction between trackers originating from a magnet link and the resume data.

You can tell whether a tracker was added manually, was part of the .torrent file or came from resume data/magnet link.

glassez: > You can tell whether a tracker was added manually, was part of the .torrent file or came from resume data/magnet link.

Until the session restarts. Then all the trackers will be "come from resume data". So I think it is useless. Otherwise you should store it in resume data.

arvidn: just as useless as "received incoming connection" is. or the `source` field for peers. The point isn't to remember where it came from, the point is to register that features are still operational.

src/torrent.cpp
```diff
@@ -651,8 +647,9 @@ bool is_downloading_state(int const st)
 		if (m_trackers.empty()) return true;
 		if (!settings().get_bool(settings_pack::use_dht_as_fallback)) return true;
 
-		return std::none_of(m_trackers.begin(), m_trackers.end()
-			, [](aux::announce_entry const& tr) { return bool(tr.verified); });
+		using std::begin;
```
glassez, line 650: Is it needed?

src/tracker_list.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+	m_trackers.clear();
+	m_url_index.clear();
+
+	// insert unordered
+	m_trackers.reserve(aes.size());
+	int idx = 0;
+	for (auto const& ae : aes)
+	{
+		auto const [iter, added] = m_url_index.insert(std::make_pair(ae.url, idx));
```
glassez, line 164:
```suggestion
        if (ae.url.empty())
            continue;

		auto const [iter, added] = m_url_index.insert(std::make_pair(ae.url, idx));
```

src/torrent.cpp
arvidn: ah, no. those are left-overs from an earlier implemetnation

arvidn:
I'm not quite happy with this patch. using a `boost::multi_index` container would be ideal. My first attempt was to do that, but the current abstraction of trackers makes it very difficult since all fields become immutable then.

The benchmark I use is to handle .torrent files with many trackers, and currently a failing tracker that's de-prioritized is very inefficient to move past all the other trackers.

glassez:
> currently a failing tracker that's de-prioritized is very inefficient to move past all the other trackers.

Maybe store some kind of pointers so that moving a tracker in the list does not require rebuilding the index?

arvidn:
@AllSeeingEyeTolledEweSew I'm having a hard time reproducing the failing python tests. When I try to run the tests manually it seems as if DYLD_LIBRARY_PATH is not propagated to the sub process (and it fails to load libtorrent.so). When I run the programs being tested directly (i.e. simple_client and make_torrent) they work fine.

So, I think there's something about how those programs are invoked.

arvidn:
Oh, maybe the GitHub page hadn't updated the test status. It looks like they all passed now.

src/tracker_list.cpp
```diff
@@ -0,0 +1,300 @@
...
+		// TODO: if this throws, new_ae needs to be freed
+		m_url_index.insert(std::make_pair(string_view(new_ae->url), new_ae));
+		return true;
+	}
+
+	auto iter = m_trackers.end();
+	while (iter != m_trackers.begin() && std::prev(iter)->tier > new_ae->tier)
+		--iter;
+
+	m_trackers.insert(iter, *new_ae);
```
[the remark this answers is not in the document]
arvidn: the intrusive list does not copy the element. The `insert()` function takes a reference though, not a pointer. That's why the pointer dereferenced.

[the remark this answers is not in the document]
glassez: Ok. I didn't noticed that you changed the type of `m_trackers`.

PR #6882, opened 2022-05-24
add asynchronous overload of torrent_handle::add_piece()

arvidn:


arvidn:
Good point

arvidn:
https://github.com/arvidn/libtorrent/pull/6885

PR #6886, opened 2022-05-26
add new overload to make_magnet_uri that takes an add_torrent_params

arvidn:


AllSeeingEyeTolledEweSew:
Can we add python bindings for this?

PR #6892, opened 2022-05-31
don't request the same metadata block unless the source is disconnected.

joriscarrier:


src/ut_metadata.cpp
```diff
@@ -500,8 +500,11 @@ namespace {
 		int const piece = int(i - m_requested_metadata.begin());
 
 		// don't request the same block more than once every 3 seconds
+		// unless the source is disconnected
 		time_point const now = aux::time_now();
 		if (m_requested_metadata[piece].last_request != min_time()
+			&& !m_requested_metadata[piece].source.expired()
+			&& !m_requested_metadata[piece].source.lock()->m_pc.is_disconnecting()
```
arvidn, line 507: technically, I don't think this is safe. `expired()` may return `false` and the last reference may go away before we call `lock()`.

An alternative would be to lock the pointer before the `if` condition, and instead of checking `expired()`, just check if the pointer is set or not.

joriscarrier: since the extensions are in a single thread I thought it wouldn't cause a problem but I agree I just changed that

arvidn: that might be true. I just get nervous about this code depending on no other thread releasing the last strong reference to the peer.

arvidn (approved):

PR #6897, opened 2022-06-03
Post add_torrent_alert after torrent is fully initialized

### LINKED ISSUE #6896: add_torrent_alert is posted too early
It should be really posted after torrent is fully initialized (e.g. torrent extensions should be created and added to torrent as well) and inserted in torrent list. Otherwise it can lead to undefined behavior. At least we can't rely on some actions to be done by torrent plug-in when handling added torrent.

I would move alert posting just below the following line:
https://github.com/arvidn/libtorrent/blob/9c87dd77be4e8477334047cd97cf20740eeddc8c/src/session_impl.cpp#L4906

glassez:
Closes #6896.

### DIFF
--- a/simulation/test_torrent_status.cpp
+++ b/simulation/test_torrent_status.cpp
@@ -384,6 +384,9 @@ TORRENT_TEST(finish_time_shift_paused)
 // first
 TORRENT_TEST(alert_order)
 {
+#if TORRENT_ABI_VERSION == 1
+	bool received_torrent_add_alert = false;
+#endif
 	bool received_add_torrent_alert = false;
 	int num_torrent_alerts = 0;
 
@@ -398,14 +401,30 @@ TORRENT_TEST(alert_order)
 		, [](lt::add_torrent_params ) {}
 		// on alert
 		, [&](lt::alert const* a, lt::session&) {
+#if TORRENT_ABI_VERSION == 1
+			if (alert_cast<torrent_added_alert>(a))
+			{
+				TEST_EQUAL(received_torrent_add_alert, false);
+				received_torrent_add_alert = true;
+			}
+			else
+#endif
 			if (auto ta = alert_cast<add_torrent_alert>(a))
 			{
+#if TORRENT_ABI_VERSION == 1
+				TEST_EQUAL(received_torrent_add_alert, true);
+#endif
 				TEST_EQUAL(received_add_torrent_alert, false);
 				received_add_torrent_alert = true;
 				handle = ta->handle;
 			}
-
-			if (auto ta = dynamic_cast<torrent_alert const*>(a))
+			else if (dynamic_cast<torrent_log_alert const*>(a))
+			{
+				// it's acceptable to receive torrent_log_alert before
+				// add_torrent_alert, since they are for debugging
+				return;
+			}
+			else if (auto ta = dynamic_cast<torrent_alert const*>(a))
 			{
 				TEST_EQUAL(received_add_torrent_alert, true);
 				TEST_CHECK(handle == ta->handle);
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -4870,9 +4870,12 @@ namespace {
 		alert_params.info_hashes = info_hash;
 
 		torrent_handle handle(torrent_ptr);
-		m_alerts.emplace_alert<add_torrent_alert>(handle, std::move(alert_params), ec);
 
-		if (!torrent_ptr) return handle;
+        if (!torrent_ptr)
+        {
+            m_alerts.emplace_alert<add_torrent_alert>(handle, std::move(alert_params), ec);
+            return handle;
+        }
 
 		TORRENT_ASSERT(info_hash.has_v1() || info_hash.has_v2());
 
@@ -4886,6 +4889,7 @@ namespace {
 		if (!added)
 		{
 			abort_torrent.disarm();
+            m_alerts.emplace_alert<add_torrent_alert>(handle, std::move(alert_params), ec);
 			return handle;
 		}
 
@@ -4904,6 +4908,8 @@ namespace {
 
 		TORRENT_ASSERT(info_hash == torrent_ptr->torrent_file().info_hashes());
 		insert_torrent(info_hash, torrent_ptr);
+
+        m_alerts.emplace_alert<add_torrent_alert>(handle, std::move(alert_params), ec);
 
 		// once we successfully add the torrent, we can disarm the abort action
 		abort_torrent.disarm();


arvidn:
it looks like `test_torrent_status.cpp.alert_order` needs to be updated as well.

arvidn:
I agree that it seems reasonable to have added the torrent to the torrent list before posting the alert.
However, I also think it's reasonable to expect a certain chronological order torrent alerts. The test that's failing is checking that `add_torrent_alert` is always the first.

arvidn:
I think this would do it: https://github.com/glassez/libtorrent/pull/1

glassez:
> I think this would do it: [glassez#1](https://github.com/glassez/libtorrent/pull/1)

Thank you! Added it.

arvidn (approved):

glassez:
Will it be backported into RC_1_2?

arvidn:
Iirc, this isn't so easy to back-port. Especially with confidence, since I don't believe there is a test

glassez:
> Especially with confidence, since I don't believe there is a test

What test do you mean?
Besides, your answer is late, and libtorrent-1.2.17 was released without this change...
Now I have provided the backport PRs myself. Let's continue there.

AllSeeingEyeTolledEweSew:
> I also think it's reasonable to expect a certain chronological order torrent alerts

this PR broke one of my tests for this reason. The system under test is my alert dispatching code; I was testing that alerts are dispatched in the expected order. How do I write that test when the expected order may change in a patch release of libtorrent? IMO alert order ought to be considered part of the public API.

glassez:
> > I also think it's reasonable to expect a certain chronological order torrent alerts
> 
> this PR broke one of my tests for this reason.

Alerts notify about some events happening in a running system, don't they? This PR fixes a clear bug when alert is posted before corresponding event can be considered as happened, making behavior of libtorrent unpredictable.
If this PR changes the order of some alerts then maybe they were ordered incorrectly before? If it makes the order unpredictable then it most likely indicates the problems in another part of code.

glassez:
@AllSeeingEyeTolledEweSew 
What alerts exactly do you talk about?



> IMO alert order ought to be considered part of the public API.

I don't think of it that way. Otherwise, the initially incorrect order will prevent fixes from being applied.
The main thing is that the system should have predictable behavior. Then the order of notifications will be stable. Unfortunately, this is not the case now, hence this patch and the other one (which fixes the problem when the fastresume data is actually rejected but no `fastresume_rejected_alert` is posted).

AllSeeingEyeTolledEweSew:
@glassez your issue was about the ordering of `add_torrent_alert` relative to other conditions, especially relative to extensions being initialized. My issue is with the ordering of one alert versus another.

In 2.0.6, `add_torrent_alert` is posted first. In 2.0.7 the deprecated `torrent_added_alert` is posted first. If I want to test "does my alert dispatch code dispatch alerts in the right order", this kind of reordering is a real issue.

glassez:
> In 2.0.6, `add_torrent_alert` is posted first. In 2.0.7 the deprecated `torrent_added_alert` is posted first.

`torrent_added_alert` isn't part of current API so I don't consider its reordering relative to alerts from current API a problem.

AllSeeingEyeTolledEweSew:
But an alert isn't like a function. A client can't just "not touch" a deprecated alert. It's in the alert stream, so it has to be processed. And pypi releases will always have deprecated functions enabled, so most python clients will always see deprecated alerts.

glassez:
> A client can't just "not touch" a deprecated alert.

Why not?

>It's in the alert stream, so it has to be processed.

Updated client (current API compliant) can just skip any deprecated alerts. Why should it process them?

AllSeeingEyeTolledEweSew:
Well, to figure out whether an alert is a desired type, you need to inspect it. That includes all the deprecated alerts.

The test that fails isn't that alert filtering works, it's that my low-level dispatch code was seeing an unexpected alert.

glassez:
@AllSeeingEyeTolledEweSew 
The productive application should use a single alternative of API. Even with a "smooth" transition from the old API to the new one, it can use some functions from one and some from the other, but it clearly should not use both interchangeable functions (alerts in this case, i.e. add_torrent_alert and torrent_added_alert). So if your program fully supports the current API, then I don't see any problems just skipping deprecated alerts. Do you have to expect a certain alert type to realize that this is something you can skip? You just get next alert from the queue and check whether it is included in the set you are interested in. If yes, then you process it (for example, you test whether it follows the order relative to other alerts from this set), otherwise you just move on to the next one.

As soon as qBittorrent gets full compatibility with the current API, I build libtorrent only with deprecated functions disabled. But I really don't care if this is enabled or disabled when someone producing releases for different platforms and distributions. qBittorrent just doesn't use the deprecated API, it doesn't matter if it's regular functions or alerts.

@arvidn 
I am honestly surprised why deprecated functions are not removed after several releases of the branch.

@AllSeeingEyeTolledEweSew 
If you are really interested in an deprecated API (or even in a mixed API, as follows from your comments), then why don't you offer to fix the error (that I fixed in this PR) for an deprecated API, instead of trying to undo this fix?

AllSeeingEyeTolledEweSew:
> If you are really interested in an deprecated API (or even in a mixed API

I'm not.

> I build libtorrent only with deprecated functions disabled

As I mentioned, python doesn't (really) have that option.

> trying to undo this fix?

I'm not. This PR has a valid purpose. It broke stuff, but fixing that should be an additional PR.

> offer to fix the error

I will if I have time. It's honestly more important that I complain here about the breakage, to avoid the same class of breakage in the future.

Let me try to illustrate what my code looks like, and the test that broke.

```python
# simplified version. the real version supports multiple concurrent callers
def iter_alerts(desired_types=None):
    while True:
        for alert in session.pop_alerts():
            if (not desired_types) or isinstance(alert, desired_types):
                yield alert

def test_iter_alerts():
    session.apply_settings({"alert_mask": status})
    session.add_torrent_async(...)
    it = iter_alerts(desired_types=None)
    first_alert = next(it)
    assert isinstance(first_alert, add_torrent_alert)
```

All python alert handling code will always see deprecated alerts. I can't filter them by changing build settings. I have to filter them by touching them.

My test may look so simple as to be degenerate. But the real `iter_alerts()` is complex due to libtorrent's alert memory management. There's a real risk of messing up alert processing order or filtering.

I should also clarify that unlike qbittorrent, python clients do not vendor libtorrent. I can't just adapt to the new alert order like this
```python
    assert isinstance(first_alert, torrent_added_alert)
```
I have to be flexible with expected alert order, like this
```python
    assert isinstance(first_alert, (add_torrent_alert, torrent_added_alert))
```
which weakens the test. In general, the more "flexible" the alert order is, the harder it is to write strong tests. I want my tests to be as strong as possible.

> why deprecated functions are not removed

Removing functions is a breaking change. That should only happen in a major release. As much as I think they're a burden, I think it's valid to have them until libtorrent 3.0 (or 2.1, or whatever).

glassez:
> Removing functions is a breaking change.

I agree if we're talking about removing functions from regular API. But isn't deprecated API is only intended to simplify transition of existing clients to new API?

glassez:
> All python alert handling code will always see deprecated alerts. I can't filter them by changing build settings. I have to filter them by touching them.

This shouldn't be a problem either, the list of deprecated alerts is obviously known. You could filter them out within `iter_alerts()` so main code works only with current API.


> I have to be flexible with expected alert order, like this

You have to anyway, don't you? The order of alerts is not so strict. For example, what alert would you expect after `add_torrent_alert`?

AllSeeingEyeTolledEweSew:
> You have to anyway, don't you? The order of alerts is not so strict. For example, what alert would you expect after `add_torrent_alert`?

There are good reasons to make alert ordering *more* strict though. Enabling clients to write strong tests is a good goal, and at some point that depends on alerts being delivered in some reliable order.

That said, in the course of discussing this I realized that I don't understand the extensions space very well. Reordering `torrent_added_alert` to mate with this PR could break some legacy clients, for all I know. I withdraw the suggestion to change it.

I'll just have to test some different condition, maybe "`add_torrent_alert` always comes before `torrent_removed_alert`"

PR #6901, opened 2022-06-04
interrupt http_connection::on_resolve() for closed commection

AlexeyKhrolenko:
Do not try to connect() in http_connection::on_resolve if connection is closed. Otherwise, there is a delay about 30 seconds on libtorrent session close if on_resolve() runs after http_connection::close()

### DIFF
--- a/src/http_connection.cpp
+++ b/src/http_connection.cpp
@@ -494,6 +494,9 @@ void http_connection::on_resolve(error_code const& e
 {
 	COMPLETE_ASYNC("http_connection::on_resolve");
 	m_resolving_host = false;
+
+	if (m_abort) return;
+
 	if (e)
 	{
 		callback(e);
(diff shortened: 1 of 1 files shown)


src/http_connection.cpp
```diff
@@ -508,6 +508,9 @@ void http_connection::on_resolve(error_code const& e
 		m_endpoints.emplace_back(addr, m_port);
 
 	if (m_filter_handler) m_filter_handler(*this, m_endpoints);
+
+	if (m_abort) return;
```
arvidn, line 512: is the reason you check `m_abort` here to catch it when the filter handler function triggers the abort?
It would seem reasonable to check it at the top of the function otherwise.

arvidn: from your description, I would also have expected this check to happen *before* `m_start_time` being set, since that resets the timeout.

AlexeyKhrolenko: fixed

PR #6907, opened 2022-06-07
Ignore visual studio files and folders in git

mohammad-ghasemi-dev:


.gitignore
```diff
@@ -84,3 +84,8 @@ examples/upnp_test
 # binaries in the tests directory
 tools/dht_put
 tools/session_log_alerts
+
+# visual studio files and folders
+/out
+/CMakeSettings.json
+/.vs
```
arvidn, line 91: these paths look suspicious. in posix these would be root paths, not relative. But maybe this is some syntax of `.gitignore` files I'm unfamiliar with.

mohammad-ghasemi-dev: You're right. I used Visual Studio's GUI to ignore these folders. These lines are created by VS. I just assumed they were correct.

PR #6908, opened 2022-06-07
load_torrent

arvidn:


AllSeeingEyeTolledEweSew:
Can we add python bindings please? :)

src/create_torrent.cpp
```diff
@@ -643,12 +643,12 @@ namespace {
 				if (url.second != current_tier)
 				{
 					current_tier = url.second;
-					trackers.push_back(std::move(tier));
+					trackers.emplace_back(std::move(tier));
```
glassez, line 646: Could you change this in the next suitable update (2.1 or 3.0) so that the trackers are in the same array as some structures (even as `announce_entry`), and not different arrays for trackers and tiers. In addition, you would then be able to store the original sources of trackers in the resume data, which may be useful to someone.

include/libtorrent/load_torrent.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include "libtorrent/span.hpp"
+#include "libtorrent/bdecode.hpp"
+#include "libtorrent/torrent_info.hpp" // for load_torrent_limits
+#include "libtorrent/aux_/export.hpp"
+
+namespace libtorrent {
+
+	// These functions load the content of a .torrent file into an
+	// add_torrent_params object.
+	// The immutable part of a torrent file (the info-dictionary) is stored in
```
glassez, line 46: I hope you haven't given up on the idea of making `torrent_info` completely immutable in the future in order to represent only original torrent metadata through it.

arvidn: that is precisely the goal this patch is making moves towards. That's why, when loading a torrent via `load_torrent()`, things like trackers, web seeds v2 merkle hashes etc. won't be set on the `torrent_info` object, but just in the `add_torrent_params` object.

glassez: And I would also rename `add_torrent_params` struct to something like `torrent_data`.

test/test_torrent.cpp
```diff
@@ -240,18 +242,15 @@ TORRENT_TEST(total_wanted)
 
 	std::vector<char> tmp;
 	bencode(std::back_inserter(tmp), t.generate());
-	auto info = std::make_shared<torrent_info>(tmp, from_span);
+	add_torrent_params p = load_torrent_buffer(tmp);
```
glassez, line 245: One more question for the future. Does it make sense to continue using `std::shared_ptr<torrent_info>` in `add_torrent_params`? Why not switch to `std::optional<torrent_info>`?

glassez: Is `master` branch the one that can accept such breaking changes I mentioned here? Maybe it makes sense to open the PR, so as not to forget all this?

arvidn: I think it makes sense to use `shared_ptr` because it allows adding it to the session without copying it. Once `torrent_info` is fully immutable, the same copy of it can be used internally as is returned by `torrent_file()`

glassez: > Once `torrent_info` is fully immutable, the same copy of it can be used internally as is returned by `torrent_file()`

Well, it will make sense.

PR #6919, opened 2022-06-11
ref sanitize_symlinks: wo goto

k-morozov:
It seems to me that sanitize_symlinks is too large. I started to refactor it. I also got away from using goto.

### DIFF
--- a/include/libtorrent/file_storage.hpp
+++ b/include/libtorrent/file_storage.hpp
@@ -716,6 +716,12 @@ namespace aux {
 	TORRENT_EXTRA_EXPORT
 	std::int64_t size_on_disk(file_storage const& fs);
 
+	TORRENT_EXTRA_EXPORT
+	bool check_valid_symlink(std::string&& target
+				, const std::unordered_set<std::string>& dir_map
+				, const std::unordered_map<std::string, std::string>& dir_links
+				, const std::unordered_map<std::string, file_index_t>& file_map);
+
 } // namespace aux
 } // namespace libtorrent
 
--- a/include/libtorrent/file_storage.hpp
+++ b/include/libtorrent/file_storage.hpp
@@ -718,9 +718,9 @@ namespace aux {
 
 	TORRENT_EXTRA_EXPORT
 	bool check_valid_symlink(std::string&& target
-				, const std::unordered_set<std::string>& dir_map
-				, const std::unordered_map<std::string, std::string>& dir_links
-				, const std::unordered_map<std::string, file_index_t>& file_map);
+				, std::unordered_set<std::string> const& dir_map
+				, std::unordered_map<std::string, std::string> const& dir_links
+				, std::unordered_map<std::string, file_index_t> const& file_map);
 
 } // namespace aux
 } // namespace libtorrent
--- a/src/file_storage.cpp
+++ b/src/file_storage.cpp
@@ -1543,9 +1543,9 @@ namespace aux {
 	}
 
 	bool check_valid_symlink(std::string&& target
-			, const std::unordered_set<std::string>& dir_map
-			, const std::unordered_map<std::string, std::string>& dir_links
-			, const std::unordered_map<std::string, file_index_t>& file_map)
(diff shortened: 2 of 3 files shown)


src/file_storage.cpp
```diff
@@ -1445,49 +1444,59 @@ namespace {
...
-			// among the concrete paths, it may be a symlink, which is also OK
-			// note that we won't iterate through this for the last step, where the
-			// filename is included. The filename is validated after the loop
-			for (string_view branch = lsplit_path(target).first;
-				branch.size() < target.size();
-				branch = lsplit_path(target, branch.size() + 1).first)
-			{
-				// this is a concrete directory
-				if (dir_map.count(branch.to_string())) continue;
+	bool file_storage::check_valid_symlink(const aux::file_entry& fe,
```
arvidn, line 1454: it's not obvious to me that this change can carry its own weight unless this can be made a free function. I can't easily tell whether this still depends on `this` or not. If it can be made a free function it can be unit tested, which would be a benefit.

include/libtorrent/file_storage.hpp
```diff
@@ -635,6 +635,11 @@ namespace aux {
 
 		aux::path_index_t get_or_add_path(string_view path);
 
+		bool check_valid_symlink(const aux::file_entry& fe,
+								 const std::unordered_set<std::string>& dir_map,
```
arvidn, line 639: this should be one additional level of indentation, not 6

src/file_storage.cpp
```diff
@@ -1445,49 +1444,59 @@ namespace {
...
+										   const std::unordered_set<std::string>& dir_map,
```
arvidn, line 1455: this is indented too much

include/libtorrent/file_storage.hpp
k-morozov: sure.
my idea works works strangely)
done.

src/file_storage.cpp
k-morozov: done

arvidn (commented):
I'm not sure this is an appropriate place to factor out a function. I think good criteria are:

1. a minimal amount of input and output state is required (which simplifies reading the code but also simplifies unit tests and possibly fuzzing)
2. not be pessimizing (at least not materially)
3. making it simpler to unit test

In this case, since `check_valid_symlink()` is still a member function, the input and output state includes the full state of the `file_storage` object, which is a lot. That's what I meant about making it a free function.

Additionally, looking at this code it looks like there are a few opportunities to make things more efficient. e.g.

1. the `traversed` set could be constructed outside of the loop, and just `clear()`ed rather than destructed and reconstructed for every iteration.
2. the `std::set<std::string>` could probably be a set of `string_view`

maybe a few other places where strings are copied could be changed to use `string_view` or at least `std::move()` the string object at its last use.

This change makes some of those optimizations more difficult, which I also think is a sign of this not being quite the right place to cut out a new function.

That said, I definitely agree that the existing function *is* too large and needs to be split up somehow.

src/file_storage.cpp
k-morozov: +

src/file_storage.cpp
```diff
@@ -1445,49 +1444,59 @@ namespace {
...
-			for (string_view branch = lsplit_path(target).first;
-				branch.size() < target.size();
-				branch = lsplit_path(target, branch.size() + 1).first)
-			{
-				// this is a concrete directory
-				if (dir_map.count(branch.to_string())) continue;
+	bool file_storage::check_valid_symlink(const aux::file_entry& fe
+		, const std::unordered_set<std::string>& dir_map
+		, const std::unordered_map<std::string, std::string>& dir_links
+		, const std::unordered_map<std::string, file_index_t>& file_map)
```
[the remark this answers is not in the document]
k-morozov: done

k-morozov:
It takes time to understand the structure of symlinks for writing check_valid_symlink unit-tests.

Issue #6917, opened 2022-06-11
"infohash merge conflict" leads to removing handles without `torrent_removed_alert`


AllSeeingEyeTolledEweSew:
libtorrent version (or branch): `libtorrent==2.0.6` from pypi

platform/architecture: Linux x86_64

Consider a v1-only magnet link, and a v2-only magnet link, that actually refer to the hashes of the same hybrid torrent. These magnet links may be added to a session, and libtorrent will create separate handles for them. When either torrent receives metadata, libtorrent realizes it conflicts with the other torrent, and puts it in an error state. I call this an "infohash merge conflict".

This is tricky. Many different behaviors by libtorrent might be valid. I couldn't find documentation on what to expect. I'll file a separate issue to narrow down what guarantees libtorrent should provide, and to document these.

This issue is about a specific issue with this case that is definitely wrong: if both handles are in a "merge conflict" error state, then removing either one from the session implicitly removes the other one, but `torrent_removed_alert` is only fired for one. This will definitely lead to bad state in the app.

Here's a python script that generates this case. I create a merge conflict, test some properties of each handle, then try to remove them both. Note that removing one handle invalidates the other one, and `torrent_removed_alert` is never fired for the second handle.

```python
import libtorrent as lt
import sys
import traceback
import requests


def main() -> int:
    # Use the official hybrid torrent
    resp = requests.get("https://libtorrent.org/bittorrent-v2-hybrid-test.torrent")
    ti = lt.torrent_info(lt.bdecode(resp.content))
    # Create separate magnet links for the v1 and v2 info hashes
    atp1 = lt.parse_magnet_uri(f"magnet:?xt=urn:btih:{ti.info_hashes().v1}")
    atp2 = lt.parse_magnet_uri(f"magnet:?xt=urn:btmh:1220{ti.info_hashes().v2}")

    # Add the magnet links as separate torrent_handles
    session = lt.session({"alert_mask": lt.alert_category.status})
    handle1 = session.add_torrent(atp1)
    handle2 = session.add_torrent(atp2)

    # Inject the metadata (rather than waiting for ut_metadata)
    handle1.set_metadata(ti.info_section())
    handle2.set_metadata(ti.info_section())

    def ih_str(ih: lt.info_hash_t) -> str:
        return f"[{ih.v1},{ih.v2}]"

    # Print some features about each handle
    print("after set_metadata(), ...")
    print(f"handle1 has info_hashes(): {ih_str(handle1.info_hashes())}")
[...]

Issue #6918, opened 2022-06-11
"infohash merge conflict" should be considered and documented


AllSeeingEyeTolledEweSew:
Consider a v1-only magnet link, and a v2-only magnet link, that actually refer to the hashes of the same hybrid torrent. These magnet links may be added to a session, and libtorrent will create separate handles for them. When either torrent receives metadata, libtorrent realizes it conflicts with the other torrent, and puts it in an error state. I call this an "infohash merge conflict".

I couldn't find specific documentation about this so I did some testing:
* The error is set to `duplicate_torrent`
* `metadata_received_alert` is **not** fired
* `torrent_handle::info_hashes()` does **not** change value
* `torrent_handle::torrent_file()` **is** populated with a valid `torrent_info`
* `torrent_handle::save_resume_data()` results in `save_resume_data_failed_alert` with error `torrent_removed`
*  #6917

Except for the last point, I think the current behavior is mostly good. However without documentation, it's hard to know whether good behavior is intentional or may change. I think we should use this edge case to help come up with some guarantees about libtorrent's behavior, and document them.

Our goal should be that either apps can automatically resolve a merge conflict, or else can present users something meaningful and they can resolve it manually.

I suggest the following invariants:
* [X] `torrent_handle::info_hashes()` may change, but will never collide with another torrent in the session
* [X] Any change of `torrent_handle::info_hashes()` will be indicated by a `metadata_received_alert` (#6913)
* [ ] `torrent_handle.info_hashes()` shall match `torrent_handle.torrent_file().info_hashes()` (if metadata exists)
* [ ] `metadata_received_alert` is fired if and only if `torrent_handle::torrent_file()` becomes non-NULL
* [X] `save_resume_data_alert`'s resume data should always be valid, if it is fired
* [ ] A call to `torrent_handle::set_metadata()` causing a merge conflict should fire a `merge_conflict_alert` which contains both conflicting handles. Since I think we shouldn't populate `torrent_file()`, this will be the only way to gather the conflicting handles. There should only be one `merge_conflict_alert` per conflict, not two.

Here's a python program that triggers a merge conflict, and probes some information about the handles:
```python
import libtorrent as lt
import sys
import traceback
import requests


def main() -> int:
    resp = requests.get("https://libtorrent.org/bittorrent-v2-hybrid-test.torrent")
    ti = lt.torrent_info(lt.bdecode(resp.content))
[...]

PR #6923, opened 2022-06-13
fix issue with merging two magnet link torrents into one hybrid

arvidn:
Fix case where two magnet links (v1-only and v2-only) end up referring to the same hybrid torrent.

- [x] documentation

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -213,6 +213,29 @@ The resume data format is very similar to the .torrent file format, and when
 including the info-dict in the resume data, the resume file can be used as a
 .torrent file (with just a few minor exceptions).
 
+BitTorrent v2 torrents
+======================
+
+BitTorrent v2 introduces a number of features outlined in `this blog post`_ as
+well as `BEP 52`_. The v2 protocol introduces the possibility to use a merkle
+hash tree instead of a flat list of piece hashes. It also supports *hybrid* torrents,
+that are both valid classing torrents (v1) as well as valid v2 torrents. Hybrid
+torrents contain both a flat list of piece hashes as well as a merkle hash tree.
+
+.. _`this blog post`: https://blog.libtorrent.org/2020/09/bittorrent-v2/
+.. _`BEP 52`: https://www.bittorrent.org/beps/bep_0052.html
+
+This introduces a few new error cases. A hybrid torrent may have mismatching v1
+and v2 hashes. Since v2 torrents use SHA-256 and v1 uses SHA-1 the fact that the
+hashes are mismatching won't be detectable until the piece has been downloaded.
+It results in a ``torrent_inconsistent_hashes`` error.
+
+A magnet link may contain just a v1 info-hash or a v2 info-hash. If two separate
+magnet links, one v1-only and one v2-only, end up resolving to the same hybrid torrent,
+both torrent_handle objects are put into an error state of ``duplicate_torrent``.
+In this state, one of them has to be removed, and the other one can be resumed,
+in order to download the metadata again.
+
 queuing
 =======
 
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -236,6 +236,18 @@ both torrent_handle objects are put into an error state of ``duplicate_torrent``
 In this state, one of them has to be removed, and the other one can be resumed,
 in order to download the metadata again.
 
+When a conflict between two torrents occur, a torrent_conflict_alert is posted.
+This alert derives from torrent_alert, so is associated with a torrent_handle.
+It contains a second torrent_handle referring to the other torrent in the
+conflict as well as the metadata that was downloaded. One way to resolve the
+conflict is to remove both torrents and add it back using the metadata supplied
+in the torrent_conflict_alert.
+
+When a v1-only or v2-only magnet link resolves to a hybrid torrent, the
+info_hash_t object associated with the torrent will be updated to include both
+the v1 and v2 info hash. This applies both to torrent_handle::info_hashes() as
+well as torrent_info::info_hashes().
+
 queuing
 =======
 
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -1985,6 +1985,9 @@ TORRENT_VERSION_NAMESPACE_3
 	};
 
 	// This is posted whenever a torrent is transitioned into the error state.
+	// If the error code is duplicate_torrent (error_code_enum) error, it suggests two magnet
+	// links ended up resolving to the same hybrid torrent. For more details,
+	// see BitTorrent-v2-torrents_.
 	struct TORRENT_EXPORT torrent_error_alert final : torrent_alert
 	{
 		// internal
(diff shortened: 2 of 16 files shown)


arvidn:
@AllSeeingEyeTolledEweSew @glassez how does this look?

There are tests ensuring:

* both torrents are put in an error state (`duplicate_torrent`).
* `clear_error()` + `resume()` will keep failing on such torrent
* removing one and `clear_error()` + `resume()` will work and complete the download (albeit, with the 60 second reconnect delay)

AllSeeingEyeTolledEweSew:
:+1: this at least resolves #6917 and makes recovery possible.

I still think `merge_conflict_alert` will be very helpful for automated recovery (https://github.com/arvidn/libtorrent/issues/6918#issuecomment-1155780225), but that could be a separate PR.

glassez:
@arvidn
Maybe I didn't see it in the code... After receiving `torrent_error` alert with `duplicate_torrent` code, how will the application be able to find the conflicting counterpart of this handle (in general, when there may be several such pairs)?

arvidn:
> After receiving torrent_error alert with duplicate_torrent code, how will the application be able to find the conflicting counterpart of this handle (in general, when there may be several such pairs)?

I don't believe that's possible with this patch. Adding a new alert would be one way of making that possible.

glassez:
> I don't believe that's possible with this patch. Adding a new alert would be one way of making that possible.

Then I don't think recovery is currently possible. We still have no way to find out which torrents are duplicates of each other in order to delete one of them. And we don't have access to the received metadata to use it for recovery.

arvidn:
I added a `torrent_conflict_alert`, including both `torrent_handle`s as well as the downloaded metadata.

include/libtorrent/alert_types.hpp
```diff
@@ -3000,6 +3003,30 @@ TORRENT_VERSION_NAMESPACE_3_END
 		file_index_t reserved;
 	};
 
+	// this alert is posted when two separate torrents (magnet links) resolve to
```
AllSeeingEyeTolledEweSew, line 3006: It may be helpful to specify that the "first" or "owner" `torrent_handle` is the one that received the metadata. This is "obvious" but may be helpful to have guaranteed in documentation, in case the app wants to use this information when resolving the conflict.

docs/manual.rst
```diff
@@ -213,6 +213,36 @@ The resume data format is very similar to the .torrent file format, and when
...
+BitTorrent v2 introduces a number of features outlined in `this blog post`_ as
+well as `BEP 52`_. The v2 protocol introduces the possibility to use a merkle
+hash tree instead of a flat list of piece hashes. It also supports *hybrid* torrents,
+that are both valid classing torrents (v1) as well as valid v2 torrents. Hybrid
+torrents contain both a flat list of piece hashes as well as a merkle hash tree.
+
+.. _`this blog post`: https://blog.libtorrent.org/2020/09/bittorrent-v2/
+.. _`BEP 52`: https://www.bittorrent.org/beps/bep_0052.html
+
+This introduces a few new error cases. A hybrid torrent may have mismatching v1
```
AllSeeingEyeTolledEweSew, line 228: It's not in the scope of the bug, but I think this section also ought to mention how `torrent_handle::info_hashes()` can change when metadata is received. It's an important consideration for any code that indexes torrents

src/torrent.cpp
```diff
@@ -7543,22 +7542,38 @@ namespace {
...
+
 			auto t = m_ses.find_torrent(info_hash_t(ih)).lock();
 			if (t && t != shared_from_this())
 			{
-				// TODO: if the existing torrent doesn't have metadata, insert
-				// the metadata we just downloaded into it.
+				TORRENT_ASSERT(!t->valid_metadata());
+
+				// if we get a collision, both torrents fail and have to be
+				// removed.
```
AllSeeingEyeTolledEweSew, line 7557: It may be helpful for future debuggers to link to the github issue to figure out the author's thought process

include/libtorrent/alert_types.hpp
arvidn: Yeah, that's true. Since that torrent doesn't have the metadata assigned to it, but left in its original state, I figured that it won't matter. But I suppose it may convey information about which hash has a swarm for instance

AllSeeingEyeTolledEweSew:
Also can we have python bindings for `torrent_conflict_alert`?

arvidn:
@AllSeeingEyeTolledEweSew good now?

src/torrent.cpp
```diff
@@ -7527,7 +7525,8 @@ namespace {
 		bdecode_node const metadata = bdecode(metadata_buf, ec, &pos, 200
 			, settings().get_int(settings_pack::metadata_token_limit));
 
-		if (ec || !m_torrent_file->parse_info_section(metadata, ec
```
glassez, line 7530: This causes https://github.com/qbittorrent/qBittorrent/issues/17595.
Previously, metadata merged into an existing `torrent_info` that contained some data from a magnet link (for example, trackers), so we just used it to (re)add a torrent.

glassez: It could be fixed by replacing the following line
```c++
auto info = std::make_shared<torrent_info>(*m_torrent_file);
```
with the following one
```c++
auto info = std::make_shared<torrent_info>(*m_torrent_file);
```

xavier2k6: > It could be fixed by replacing the following line
> 
> ```c++
> auto info = std::make_shared<torrent_info>(*m_torrent_file);
> ```
> 
> with the following one
> 
> ```c++
> auto info = std::make_shared<torrent_info>(*m_torrent_file);
> ```

they are the same??

glassez: > Did you mean `replace`
> 
> ```c++
> auto info = std::make_shared<torrent_info>(old_ih);
> ```

Exactly!
Thanks.

arvidn: the trackers part of `m_torrent_file` have (presumably) already been parsed into `torrent::m_trackers`, haven't they?
I'm hoping that any non-info-section fields can be removed from `torrent_info` in the long-run.

arvidn: I see a comment on line 332 suggesting the trackers are also saved in `m_torrent_file` in order to propagate to `create_torrent()`. Is that the path where you get the discrepancy?

arvidn: https://github.com/arvidn/libtorrent/pull/7031

glassez: > I'm hoping that any non-info-section fields can be removed from `torrent_info` in the long-run.

Yes. I agree. I believe `torrent_info` should become an immutable representation of info section in future. But until then, such changes are still unexpected.

glassez: > I see a comment on line 332 suggesting the trackers are also saved in `m_torrent_file` in order to propagate to `create_torrent()`. Is that the path where you get the discrepancy?

No. We use it differently.
When user adds magnet interactively we add it to libtorrent session behind the scenes in order to obtain metadata and allow user to select needed files. Once metadata is received we remove the torrent from libtorrent session and then add it using obtained metadata as a regular torrent file.

glassez: Of course the scenario above will be adjusted when `torrent_info` becomes "info section" only.

PR #6939, opened 2022-06-27
fix python test on Mac

arvidn:


arvidn:
@AllSeeingEyeTolledEweSew even though this didn't fix the issue I was hoping it would, it seems like a good change still, right?

AllSeeingEyeTolledEweSew:
:+1: seems good

AFAICT this didn't change the path of the invoked python or the install location, but I guess this invocation makes more sense than what was there

Samfun75:
The failure might have occurred because 2 different pythons are being used to install and run the tests. I have tried running the tests with brew python and that might have fixed it.

You can check the run [here](https://github.com/Samfun75/libtorrent/runs/7210269152?check_suite_focus=true) and the changes on top of this PR can be found [here](https://github.com/Samfun75/libtorrent/pull/13/files).

PR #6955, opened 2022-07-06
Prevent 'slack' connections triggering a disconnect when falling back under the peer limit

thrnz:
When a 'slack' connection is accepted it gets tagged so as to force a disconnection - of either itself or another peer - once it gets attached to a torrent. It's possible to drop below the connection limit while waiting for that to happen though, meaning we could end up disconnecting a peer needlessly.

This change adds a check to see if we're still over the connection limit before choosing which peer to disconnect.


arvidn (approved):

arvidn:
thanks!

PR #6969, opened 2022-07-17
Fix missing macro definition when compiling in shared library mode

Chocobo1:
When compiling in shared library mode with MSVC, the `TORRENT_LINKING_SHARED` macro is expected to be defined but it wasn't. This commit fixes it.

@arvidn 
Other branches (RC_1_2, master) also has this defect, please port it over.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -569,7 +569,7 @@ target_compile_features(torrent-rasterbar
 if (BUILD_SHARED_LIBS)
 	target_compile_definitions(torrent-rasterbar
 		PRIVATE TORRENT_BUILDING_SHARED
-		INTERFACE TORRENT_LINKING_SHARED
+		PUBLIC TORRENT_LINKING_SHARED
 	)
 endif()
 


Chocobo1:
On a second thought, I don't think this fix is in the right direction.

I might have accidentally sidestepped the issue with it. My issue was: when using libtorrent as a shared library (built with cmake), qbittorent will crash in a libtorrent function (`bdecode()`) due to dereferencing a null pointer. But the strange part was the pointer shouldn't be null in any case and it really isn't by looking at one frame above the crashing function. The compiling and linking stages had no errors nor warnings...

glassez:
@Chocobo1
I have been using libtorrent as shared library for many years (I previously built it using Boost.Build, now using CMake) and have never had such a problem. Are you sure you're doing everything right?

Chocobo1:
>Are you sure you're doing everything right?

My own static linking was working fine but not dynamic linking (on qbt appveyor). Compiler doesn't emit anything... I was using the following flags for shared lib:
```cmake
-DBUILD_SHARED_LIBS=ON
-Ddeprecated-functions=OFF
-Dstatic_runtime=OFF
```

glassez:
>My own static linking was working fine but not dynamic linking (on qbt appveyor). 

@Chocobo1 
Do you talk about build artifact from AppVeyor?
Since AppVeyor uses QMake to build qBittorrent (unless I'm wrong) I suppose there are conflicts between build options between libtorrent and qBittorrent (e.g. different compiler or linker flags, or even different c++ standard).

> I was using the following flags for shared lib:
> 
> ```cmake
> -DBUILD_SHARED_LIBS=ON
> -Ddeprecated-functions=OFF
> -Dstatic_runtime=OFF
> ```

What c++ standard is used for building libtorrent?
And what one for qBittorrent?

arvidn:
I would expect the failure when this macro isn't defined would be a link error. it controls wither symbols are marked as `__declspec(dllimport)` or `__declspec(dllexport)` on msvc.

The crash sounds similar to the mistake I made in `RC_1_2` where libtorrent's ABI depends on the C++ version ([here](https://github.com/arvidn/libtorrent/blob/RC_1_2/include/libtorrent/entry.hpp#L90)), but in `RC_2_0` and later, C++14 is required so it shouldn't be a problem anymore.

Another possibility is that you have installed libtorrent (old) headers somewhere that get picked up instead of the ones that came with the shared library you just built.

glassez:
> in `RC_2_0` and later, C++14 is required so it shouldn't be a problem anymore.

So do you believe that it isn't a problem if libtorrent is compiled as c++14 and the application that linked with it is compiled as c++17?

glassez:
> > in `RC_2_0` and later, C++14 is required so it shouldn't be a problem anymore.
> 
> So do you believe that it isn't a problem if libtorrent is compiled as c++14 and the application that linked with it is compiled as c++17?

Although to be honest I don't recall having to take care of the C++ standard when I build libtorrent lately, but it works without problems with qBittorrent that is built in C++17 mode.

Chocobo1:
>Do you talk about build artifact from AppVeyor?

Yes and it only crashes at run time.
Although, I have somehow work around the problem, so there is no crash now. I hope we get to decommission appveyor before requiring another update..

>What c++ standard is used for building libtorrent?

The default for msvc. I suppose it is c++14. I tried changing https://github.com/arvidn/libtorrent/blob/722d78250a30c89c92970a78be970269a395be36/CMakeLists.txt#L557 to `cxx_std_17` but it still crash.

>And what one for qBittorrent?

As usual, qbt master branch requires C++17.

>The crash sounds similar to the mistake I made in RC_1_2 w[here](https://github.com/arvidn/libtorrent/blob/RC_1_2/include/libtorrent/entry.hpp#L90) libtorrent's ABI depends on the C++ version (here), but in RC_2_0 and later, C++14 is required so it shouldn't be a problem anymore.
>
>Although to be honest I don't recall having to take care of the C++ standard when I build libtorrent lately, but it works without problems with qBittorrent that is built in C++17 mode.

For my own local builds, I never need to meddle with C++ standard when building with MSVC (on both libtorrent and qbttorrent) they are left at the default value.

Probably worth noting that (when compiling shared library for appveyor) besides the change in this PR, the only other change is manually setting to [`/std:c++17`](https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170) at msvc project files. However I find it hard to believe that it could make a difference..

PR #6978, opened 2022-07-22
remove unnecessary oig_files() on storage classes

arvidn:


xavier2k6:
small thing.....typo, `oig_files()` in title & commit should be `orig_files()`

PR #7003, opened 2022-08-10
fix MacOS python CI build

arvidn:


.github/workflows/python.yml
```diff
@@ -103,6 +105,7 @@ jobs:
       run: |
         # Install to Homebrew's python site-packages. no need for --user and --prefix
         cd bindings/python
+        export PATH=$(brew --prefix)/opt/python@3.10/bin:$PATH
         python3 setup.py build_ext install --install-lib $(brew --prefix)/lib/python3.9/site-packages
```
Samfun75, line 109:
```suggestion
        python3 setup.py build_ext install --install-lib $(brew --prefix)/lib/python3.10/site-packages
```

Samfun75: The install location should be updated too

PR #7018, opened 2022-08-18
Increase the version of upcoming branch

glassez:
Otherwise it is hard (or even impossible) to maintain compatibility of applications with release branches and master branch simultaneously.

### DIFF
--- a/Jamfile
+++ b/Jamfile
@@ -12,7 +12,7 @@ import cast ;
 
 # we need version numbers in the form X.Y.Z in order to trigger the built-in
 # support for generating symlinks to the installed library
-VERSION = 2.0.7 ;
+VERSION = 2.1.0 ;
 
 BOOST_ROOT = [ modules.peek : BOOST_ROOT ] ;
 ECHO "OS =" [ os.name ] ;
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-VERSION=2.0.7
+VERSION=2.1.0
 
 BUILD_CONFIG=release link=shared crypto=openssl warnings=off address-model=64
 
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -470,7 +470,7 @@ def find_all_files(path: str) -> Iterator[str]:
 
 setuptools.setup(
     name="libtorrent",
-    version="2.0.7",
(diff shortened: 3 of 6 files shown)


arvidn:
there's a script called `set_version.py` which updates the version in a few other places also.

I'm interested in keeping `master` as API compatible as possible with 2.0.x, so please let me know if you come across some incompatibility you think should be fixed. (accepting session flags in the constructor to be able to start in paused mode is on my list already)

glassez:
>I'm interested in keeping master as API compatible as possible with 2.0.x

Do you talk about full compatibility, i.e. with `deprecated-functions=OFF`?

In any case, as long as `master` has differences from other branches, it should have a different version (or some other convenient way to distinguish it).

glassez:
> please let me know if you come across some incompatibility you think should be fixed. (accepting session flags in the constructor to be able to start in paused mode is on my list already)

Other than session flags in the session constructor (and new disk I/O type, of course) it is currently only the following:
https://github.com/arvidn/libtorrent/blob/283cea5981b07923b14247fc7be6f68392778c0e/include/libtorrent/torrent_info.hpp#L60-L65

>I'm interested in keeping master as API compatible as possible with 2.0.x

I doubt this is really a good idea. We don't have the ability to make changes affecting the API/ABI inside the branch, and now we won't even be able to do it with larger updates?

arvidn:
There's one place in `docs/hunspell/libtorrent.dic` where the version number is mentioned, and it's not updated automatically by the script. That's why the spell check is failing right now.

glassez:
> > I doubt this is really a good idea. We don't have the ability to make changes affecting the API/ABI inside the branch, and now we won't even be able to do it with larger updates?
> 
> I don't know what you mean.

From time to time, there are some improvements/fixes are offered (and maybe you yourself have some ideas) that could change the API or ABI. You can't accept them into the current release branches (RC_1_2 and RC_2_0), so we're either left with nothing or forced to use some kind of workarounds on the application side. But we still hope that these improvements/fixes will be accepted into the `master` branch, so that they will get into the next minor/major update.

glassez:
@arvidn 
How about type of `web_seed_entry`? Is it even useful in current API?

PR #7022, opened 2022-08-20
remove unused variable in torrent::recalc_share_mode

aldenml:
I found this with the linter and looks a little suspicious


arvidn (approved):
this variable has been unused since this function was introduced

PR #7025, opened 2022-08-22
Optimize loading merkle tree mask

glassez:
For some reason loading data into `std::vector<bool>` using `reserve()` + `emplace_back()` is much less efficient, at least under Windows + MSVC (maybe other compilers provide more optimal implementation). According to my observations with the naked eye (tested with fully checked dummy v2 torrent of size 35+ GB), this patch works an order of magnitude faster, provided that the entire source array is filled with 1 (which differs from the default value used when initializing the vector), and it works faster the more the source data matches the default value.

### DIFF
--- a/src/read_resume_data.cpp
+++ b/src/read_resume_data.cpp
@@ -4,7 +4,7 @@ Copyright (c) 2015-2020, Arvid Norberg
 Copyright (c) 2016-2018, Alden Torres
 Copyright (c) 2017, Pavel Pimenov
 Copyright (c) 2017, Steven Siloti
-Copyright (c) 2020, Vladimir Golovnev (glassez)
+Copyright (c) 2020, 2022, Vladimir Golovnev (glassez)
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


glassez:
@qbittorrent/bug-handlers
Could you test it as well (especially under non-Windows OS)?

glassez:
[@qbittorrent/bug-handlers](https://github.com/orgs/qbittorrent/teams/bug-handlers)
Could you test it as well (especially under non-Windows OS)?

arvidn (approved):

arvidn:
good catch! I've realized that the resume-file format really should be fixed here to store the bitfields as bits rather than bytes. That's a bigger change though, that needs a transition plan.

glassez:
> I've realized that the resume-file format really should be fixed here to store the bitfields as bits rather than bytes. That's a bigger change though, that needs a transition plan.

Initially, I thought about it, but `std::vector<bool>` does not allow you to simply take an internal array of data and convert it into a binary string (by standard, it may not store all the data in a contiguous memory area). So I tried changing it to use a `libtorrent::bitfield` instead of `std::vector<bool>`. The results amazed me, even without changing the bencoded data format. But since it can only be merged into `master`, I decided to try the same approach in loading data to apply to `std::vector<bool>` in order to affect the `RC_2_0` branch.

I can post PR to `master` if you are really interested in optimizing it even more by storing the data in binary form.

arvidn:
> I can post PR to master if you are really interested in optimizing it even more by storing the data in binary form

please do

glassez:
> > I can post PR to master if you are really interested in optimizing it even more by storing the data in binary form
> 
> please do

#7028

PR #7027, opened 2022-08-22
make the SOCKS5 client tolerate a domain name in the response to UDP ASSOCIATE

arvidn:
Patch by Andrei Borzenkov

### DIFF
--- a/src/udp_socket.cpp
+++ b/src/udp_socket.cpp
@@ -5,6 +5,7 @@ Copyright (c) 2015, Thomas Yuan
 Copyright (c) 2016-2018, 2020, Alden Torres
 Copyright (c) 2016, Andrei Kurushin
 Copyright (c) 2016, Steven Siloti
+Copyright (c) 2022, Andrei Borzenkov
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
@@ -111,6 +112,8 @@ private:
 	void socks_forward_udp();
 	void connect1(error_code const& e);
 	void connect2(error_code const& e);
+	void read_bindaddr(error_code const& e);
+	void read_domainname(error_code const& e);
 	void hung_up(error_code const& e);
 	void on_retry_socks_connect(error_code const& e);
 
@@ -902,7 +905,7 @@ void socks5::connect1(error_code const& e)
 	}
 
 	ADD_OUTSTANDING_ASYNC("socks5::connect2");
-	boost::asio::async_read(m_socks5_sock, boost::asio::buffer(m_tmp_buf.data(), 10)
+	boost::asio::async_read(m_socks5_sock, boost::asio::buffer(m_tmp_buf.data(), 5)
 		, std::bind(&socks5::connect2, self(), _1));
 }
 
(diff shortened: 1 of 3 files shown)


src/udp_socket.cpp
```diff
@@ -932,8 +935,19 @@ void socks5::connect2(error_code const& e)
...
-		m_udp_proxy_addr.port(read_uint16(p));
+		ADD_OUTSTANDING_ASYNC("socks5::read_bindaddr");
+		m_tmp_buf[0] = *p;
+		boost::asio::async_read(m_socks5_sock, boost::asio::buffer(m_tmp_buf.data() + 1, 5)
+			, std::bind(&socks5::read_bindaddr, self(), _1));
+	}
+	else if (atyp == 3)
+	{
+		// we need to skip DOMAINNAME to imitate Python socks client
+		std::size_t const len = read_uint8(p);
```
arvidjaar, line 946: After this read `p` advances past the last read (5th) byte and points at the 6th byte which is the first hostname character.

src/udp_socket.cpp
```diff
@@ -932,8 +935,19 @@ void socks5::connect2(error_code const& e)
...
+		ADD_OUTSTANDING_ASYNC("socks5::read_domainname");
+		m_tmp_buf[0] = *p;
```
arvidjaar, line 948: If I interpret this correctly, it moves the beginning of `m_tmp_buf`. Which means after several invocations it will point near the end of available space. In the best case there will be not enough space to process requests, in the worst case there will be buffer overflow.

src/udp_socket.cpp
```diff
@@ -942,6 +956,58 @@ void socks5::connect2(error_code const& e)
...
+			m_alerts.emplace_alert<socks5_alert>(m_proxy_addr, operation_t::handshake, e);
+		++m_failures;
+		retry_connection();
+		return;
+	}
+
+	using namespace libtorrent::aux;
+
+	char* p = m_tmp_buf.data();
+	int const len = read_uint8(p);
```
arvidjaar, line 1007: No, at this point `p` points at the first hostname character, not length byte. In our case this character is `\0` ...

src/udp_socket.cpp
```diff
@@ -942,6 +956,58 @@ void socks5::connect2(error_code const& e)
...
+	p += len;
```
arvidjaar, line 1008: ... so we remain at the same place and miss the second byte of port. Because in my case port number is 7497 (0x1d49), it always attempts to send datagram to port 29 (0x001d).

src/udp_socket.cpp
arvidn: that's right. We get to this callback after having read 5 bytes, so the length (the 5th byte) is valid here. this is the last valid byte of the read buffer.

This was an optimization over your original patch where you would read up to the `ATYP` as the first read operation, instead I read up to the byte following `ATYP`. That way, if it's a hostname, I know how many more bytes to read exactly, and only have to issue a single additional read operation.

src/udp_socket.cpp
arvidn: I see the error. this was meant to write `len` to this position in the buffer. Because it's passed along to the next callback.

arvidn: I don't understand what you mean by "after several invocations". But `*p` is uninitialized and UB (or at least just corrupt)

src/udp_socket.cpp
arvidn: actually, it points to uninitialized memory (because of the bug I mentioned earlier). Note that the `async_read()` calls start at position 1 in the read buffer.

arvidn:
the test didn't catch the typo because it was an empty hostname. I should make it send some random string instead

arvidn:
https://github.com/arvidn/libsimulator/pull/69

arvidn:
I force pushed an update to fix the bug (and use the updated `libsimulator` which sends a non-empty hostname)

src/udp_socket.cpp
arvidjaar: Sorry, my bad, I misunderstood what it does.

arvidjaar:
I confirm that it works. But honestly, there is very little remained from my patch so I do not think I can be attributed authorship much less the copyright. If you feel it is necessary, I am fine with "reported by". Thank you!

P.S. now if we could fix the second issue with UDP trackers ... :)

PR #7028, opened 2022-08-22
Use lt::bitfield to represent merkle tree mask

glassez:


include/libtorrent/bitfield.hpp
```diff
@@ -201,6 +206,30 @@ namespace libtorrent {
...
+			{
+				TORRENT_ASSERT(buf);
+				if (n == 0)
+					return;
+
+				if (n < 0)
+					sub(-n);
+
+				buf += (n % 32);
+				for (int i = 0; i < (n / 32); ++i) inc();
```
glassez, line 219: Have no clear idea how to make it better...

glassez:
I'll try to figure out the tests later.

glassez:
> I'll try to figure out the tests later.

@arvidn
I've done everything I could. I don't really know why some tests fail on macOS.

include/libtorrent/bitfield.hpp
```diff
@@ -201,6 +206,32 @@ namespace libtorrent {
...
+				TORRENT_ASSERT(buf);
+				if (n == 0)
+					return;
+
+				if (n < 0)
+					return sub(-n);
+
+				buf += (n / 32);
+				int const remaining_bits = (n % 32);
+				for (int i = 0; i < remaining_bits; ++i) inc();
```
glassez, line 220: This loop can be avoided if change iterator implementation to store buffer start + position instead of current word + bitmask.

PR #7033, opened 2022-08-27
use pwrite() for writing to files in mmap_storage

arvidn:
This is an experiment to use `pwrite()` instead of writing into a memory mapped file. This can improve performance when downloading to a network share, since a write operations can be sent directly to the remove machine without first reading a full page (or edges of the write operation that don't fully cover a page).

This appears to be working on windows, linux and MacOS. It requires that the operating system synchronizes the two ways of writing to files.

Perhaps there should be a configuration option to pick which mechanism to use when writing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -218,10 +218,12 @@ set(libtorrent_aux_include_files
 	disk_io_thread_pool.hpp
 	disk_job_fence.hpp
 	disk_job_pool.hpp
+	drive_info.hpp
 	ed25519.hpp
 	escape_string.hpp
 	export.hpp
 	ffs.hpp
+	file_descriptor.hpp
 	file_progress.hpp
 	file_view_pool.hpp
 	has_block.hpp
@@ -275,6 +277,7 @@ set(libtorrent_aux_include_files
 	utp_stream.hpp
 	vector.hpp
 	win_crypto_provider.hpp
+	win_file_handle.hpp
 	win_util.hpp
 )
 
@@ -314,6 +317,7 @@ set(sources
 	disk_io_thread_pool.cpp
 	disk_job_fence.cpp
 	disk_job_pool.cpp
+	drive_info.cpp
 	entry.cpp
 	enum_net.cpp
 	error_code.cpp
--- a/Jamfile
+++ b/Jamfile
@@ -727,6 +727,7 @@ SOURCES =
 	disabled_disk_io
 	disk_job_fence
 	disk_job_pool
+	drive_info
 	entry
 	error_code
 	file_storage
--- a/Makefile
+++ b/Makefile
@@ -315,6 +315,7 @@ SOURCES = \
   disk_io_thread_pool.cpp         \
   disk_job_fence.cpp              \
   disk_job_pool.cpp               \
+  drive_info.cpp                  \
   entry.cpp                       \
   enum_net.cpp                    \
   error_code.cpp                  \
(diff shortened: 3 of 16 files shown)


glassez:
Until it has a configurable/manageable memory consumption, I'm afraid it won't be a real alternative to the libtorrent-1.2 disk I/O subsystem.

arvidn:
is the bulk of the memory used by clean pages (i.e. read-cache) then?

glassez:
> is the bulk of the memory used by clean pages (i.e. read-cache) then?

I suppose it is still a problem of system performance/responsiveness when most of the physical memory is occupied (even by clean pages).

KyleSanderson:
Indeed. Mmap is a well documented mistake for performance over ease of use. This is why sendfile is still so important, with io_uring taking over both storage and networking bulking operations like this. Page faulting is simply expensive, and even with madvise hints you're making those same syscalls that you would normally make.

The "right thing" for systems on 5.4 or greater is io_uring, and would be a very welcomed improvement for the single threaded network stack as well.

arvidn:
yes, page faults are expensive. It is a concern, but it's not the main concern. The problem with a user-space disk cache is that it's always either too small or too large (because making it dynamic is hard [1]). Perhaps it's not very *much* too small or too large, and perhaps that efficiency loss is smaller than the efficiency loss of the kernel not managing the page cache very well. But I don't believe throughput it the main problem people have had (except perhaps when running a complete hash check of all files).

My understanding of the main issue people are reporting fall into a few different buckets:
1. The disk read cache balloons and infringes more important pages
2. The dirty pages aren't flushed in a timely manner and infringe on more important pages.
3. Checking files, i.e. sequentially reading all files, isn't as fast.
4. Downloading to a remote mount is slow and inefficient.

I think there may be another bucket that has emerged recently where just having a lot of file mappings in a process might be expensive, so torrents with many small files are inefficient.

It's not so easy to measure and quantify these problems. I believe my benchmarks suffer from artificially non-fragmented files being checked. Another difficulty in pin pointing these problems is that problem (1) also exists when reading with normal pread-like calls, especially on windows.

This patch attempts to address (4). The theory being that:
* when writing via a memory mapped file, the page fault needs to *read* the page content from disk first, which is especially inefficient over a network mounted drive (at least if the protocol has poor support for sparse files, and doesn't know whether we're writing to a hole or not). 
* filesystems on local drives are more likely to know whether a page is a hole or not, in which case there won't be a need to read the page from disk first.

[1]: MacOS supports "purgable" memory, where you can put things like a cache, and if the kernel needs more RAM, that's one of the first things that it reclaims. When experimenting with this, it was not a clear win, it also still required to keep a copy of what was in the page cache.

glassez:
> yes, page faults are expensive. It is a concern, but it's not the main concern. The problem with a user-space disk cache is that it's always either too small or too large (because making it dynamic is hard [1]). Perhaps it's not very _much_ too small or too large, and perhaps that efficiency loss is smaller than the efficiency loss of the kernel not managing the page cache very well. But I don't believe throughput it the main problem people have had

The general unresponsiveness of the system as a result of the fact that qBittorrent occupies almost all physical memory was one of the first problems reported by users when we started using libtorrent 2.0 as the main one. Because of this, we had to experiment with limiting the memory working set for entire application, but it seems to be a bad workaround, besides it does not work on all systems.

glassez:
>The problem with a user-space disk cache is that it's always either too small or too large

At least, it can be limited by the user so that it obviously does not take up all system resources, and leave enough space in memory to perform other regular tasks (web surfing, watching videos, typing essays, etc.).

arvidn:
> At least, it can be limited by the user so that it obviously does not take up all system resources, and leave enough space in memory to perform other regular tasks (web surfing, watching videos, typing essays, etc.).

There are a lot of reports from windows users that the disk cache still grows indefinitely, even when using regular file I/O functions. This is mitigated by regularly closing file handles on windows. It seems like it might be fewer people experiencing it compared to with memory mapped files though.

PR #6777, opened 2022-03-11
Backport setup.py from master (RC_2_0 only)

AllSeeingEyeTolledEweSew:
Progress toward #6531 

The intent is to backport fixes, formatting and type annotations. The larger goal is to backport the test suite and type stubs from `master`, as well as make it easier to do other backports, and fixes that should be merged on all branches.

I copied `setup.py` from `master`, and manually added the following logic, currently present in `RC_2_0`:
1. Write `project-config.jam` to cwd, rather than use `b2 --project-config=<tempfile>`

I created a separate PR to backport `setup.py` from `master` to `RC_1_2`.

See also: [Diff from `master` to this PR](https://github.com/arvidn/libtorrent/compare/master..AllSeeingEyeTolledEweSew:backport/10-setup.py/RC_2_0#diff-34cc0f25ff048b0f9e6a88e67b0860c45056874287c359c1454ae32edd8826ab)

PR #7040, opened 2022-09-04
Backport setup.py formatting from master (RC_1_2 only)

AllSeeingEyeTolledEweSew:
Progress toward #6531

Companion to #6777

The intent is to backport fixes, formatting and type annotations. The larger goal is to backport the test suite and type stubs from `master`, as well as make it easier to do other backports, and fixes that should be merged on all branches.

I copied `setup.py` from `master`, and manually added the following logic, currently present in `RC_1_2`:
1. Write `project-config.jam` to cwd, rather than use `b2 --project-config=<tempfile>`
2. Honor `compile_flags`/`link_flags` generated by autotools
3. Default `cxxstd=11` on mac

See also: [Diff from `master` to this PR](https://github.com/arvidn/libtorrent/compare/master..AllSeeingEyeTolledEweSew:backport/10-setup.py/RC_1_2#diff-34cc0f25ff048b0f9e6a88e67b0860c45056874287c359c1454ae32edd8826ab)


AllSeeingEyeTolledEweSew:
NB: pre-commit fails because we're using an outdated `black` pre-commit hook. The current `black` hook fails on python 3.7+. This didn't show up in #7041 because we did not actually enable the hook for any files.

The remedy is to update the hook. I did this in #7045. I have confirmed that if #7045 is merged, the pre-commit hook will succeed in this PR. However I also think #6775 should be merged to avoid problems like this, or it should be closed if you don't agree.

AllSeeingEyeTolledEweSew:
@arvidn I rebased this, now that #7045 landed. Please take another look

arvidn:
I hop I won't regret this when merging forward again :)

AllSeeingEyeTolledEweSew:
> I hop I won't regret this when merging forward again :)

I really tried to make it simple: `setup.py` shouldn't change in the next merge. This PR is what it should look like in `RC_1_2`, and #6777 is what it should look like in `RC_2_0`. It shouldn't change in `master`. Is there some better way to do this?

arvidn:
how do you suggest I merge `RC_1_2` into `RC_2_0` now?
If I merge and then just undo the changes to `bindings/python/setup.py`, the change to default to `toolset=darwin` is lost.
There are quite a few conflicts that are non-trivial to fix in order to perform the merge properly

arvidn:
does this look right? https://github.com/arvidn/libtorrent/pull/7062

AllSeeingEyeTolledEweSew:
> how do you suggest I merge `RC_1_2` into `RC_2_0` now? If I merge and then just undo the changes to `bindings/python/setup.py`, the change to default to `toolset=darwin` is lost. There are quite a few conflicts that are non-trivial to fix in order to perform the merge properly

The way I thought it might work is you'd merge #6777 first, merge `RC_2_0` to `master` in which `setup.py shouldn't change, then I'd create this PR later and repeat the process. That's why I wrote #6777 first, months ago.

I wasn't sure what was going on with #6777 -- whether you thought there was some better way to achieve the goal -- so I just started working on this PR and other changes that needed to happen in all branches, like `toolset=darwin`.

#7062 looks good.

AllSeeingEyeTolledEweSew:
I am going to have to continue to make changes to `setup.py`, such that it has different behavior in different branches. Merges are going to continue to be tricky this way.

Instead of the current state, what do you think about ensuring `setup.py` is identical in all branches, and having behavior controlled by feature flags in some external file?

PR #7041, opened 2022-09-04
bump python version to 3.7+ in CI and docs

AllSeeingEyeTolledEweSew:
[Python 3.6 is end-of-life as of 2021-12-23](https://peps.python.org/pep-0494/#lifespan).

In #6188 we added a requirement of python 3.7+, but only to the `setup.py` build path. The b2, autotools and cmake paths still require python 3.6+. Furthermore ubuntu-18.04 still only provides python 3.6, and [18.04 will be supported until 2023 or later](https://ubuntu.com/about/release-cycle). We probably must wait until then to drop 3.6 support, to avoid breaking too many people.

I thought we should at least bump the version in CI and documentation.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -20,7 +20,7 @@ jobs:
            submodules: recursive
       - uses: actions/setup-python@v2
         with:
-           python-version: 3.6
+           python-version: 3.7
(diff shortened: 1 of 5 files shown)


arvidn (approved):
excellent

PR #7048, opened 2022-09-07
don't memory map files smaller than 1 MiB

arvidn:
use pwrite/pread calls.

perhaps the file size cut-off should be configurable. That would enable not using memory mapped files at all.

### DIFF
--- a/include/libtorrent/aux_/file_view_pool.hpp
+++ b/include/libtorrent/aux_/file_view_pool.hpp
@@ -89,7 +89,8 @@ namespace aux {
 		// return an open file handle to file at ``file_index`` in the
 		// file_storage ``fs`` opened at save path ``p``. ``m`` is the
 		// file open mode (see file::open_mode_t).
-		file_view open_file(storage_index_t st, std::string const& p
+		std::shared_ptr<file_mapping>
+		open_file(storage_index_t st, std::string const& p
 			, file_index_t file_index, file_storage const& fs, open_mode_t m
 #if TORRENT_HAVE_MAP_VIEW_OF_FILE
 			, std::shared_ptr<std::mutex> open_unmap_lock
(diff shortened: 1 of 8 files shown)


glassez:
>don't memory map files smaller than 1 MiB

Wouldn't it be useful to also add `don't memory map files larger than <size>`?

arvidn:
> Wouldn't it be useful to also add `don't memory map files larger than <size>`?

The overhead of the mapping would be amortized over a larger file, so it isn't as costly for larger files.

glassez:
> The overhead of the mapping would be amortized over a larger file, so it isn't as costly for larger files.

Apparently, I don't understand what you're actually doing with file mapping (I haven't looked at its code in detail). That's why I'm worried about keeping the whole files mapped into memory (especially the big ones). It seems to me that in any case, sooner or later it leads to the use of all physical memory (unless, of course, you close files periodically, as is done for downloaded files).

arvidn:
The concern of using too much virtual address space could be addressed by having a limit and only map views of files. to support files larger than the limit. This would also open up the possibility to use mmap on 32 bit systems.

The concern of the mappings never being evicted would also be addressed in some other way. Perhaps by closing files periodically (which there is support for already I believe), but that might not make any difference. I believe using `pwrite()` to write might alleviate this issue to some extent. But it's kind of a nebulous issue.

Having an upper size limit on when to use mmap wouldn't really address it since you can still have a torrent with a lot of files just under that limit. You would probably want something like a limit on the virtual address space in that case.

This patch is meant to address the overhead of mapping small files, which is disproportionately expensive (as far as I understand).

glassez:
> Having an upper size limit on when to use mmap wouldn't really address it since you can still have a torrent with a lot of files just under that limit.

You're right.


> You would probably want something like a limit on the virtual address space in that case.

No, the limit exactly of what is loaded into physical memory.

Issue #6935, opened 2022-06-25
Nagle's algorithm and uTP


thrnz:
I've noticed that the way libtorrent handles undersized uTP packets is different than that of ~libutp~ and the original [RFC](https://datatracker.ietf.org/doc/html/rfc896) introducing Nagle's algorithm, and introduces a potential delay of up to an RTT.

If I'm reading the code correctly, currently libtorrent holds back sending an undersized uTP packet until its either no longer undersized, or until there are no more unack'd packets in flight. As a result, the packet can be held back for up to an entire RTT waiting for acks to arrive before being transmitted.

https://github.com/arvidn/libtorrent/blob/084219e75175b4f3ab096aed83f32d70ef1f2ee4/src/utp_stream.cpp#L1514-L1520

~libutp appears to only hold back an undersized packet until the next ack is received.~ The RFC also only mentions holding back the packet until something arrives from the other end (though it is from 1984, and I don't know if things have changed). As a result, there is much less likely to be any significant extra delay.

I'm not sure what the effects of this extra delay might be - if any - or whether it's been intentionally done that way or was an oversight. One thing that does come to mind though is when a peer is choked, where there will be a large stream of outgoing data that ends in a tiny choke message. The choke message could potentially be held back as undersized until everything previously in-flight is ack'd, causing the peer to continue requesting new pieces, and - if the extra RTT delay on top of the existing send buffers and underlying latency is high enough - potentially be disconnected due to sending requests while choked.

I'd be surprised if this alone has any meaningful effect on #3542.

arvidn:
hopefully I'll find some time to look into this.
There are some simulations that might be useful in investigating the exact behavior, although, I don't think there are any tests for nagel's algorithm specifically.

thrnz:
I've been using [this](https://github.com/thrnz/libtorrent/commit/4d55a1169ef322f5de51f4ba467f2c2ca70bd199) for the last month or so to potentially work around it, though it's not been tested beyond the 'it still compiles and doesn't outright break uTP' stage.

arvidn:
interesting. My understanding of nagle's algorithm is that only one under-sized packet is supposed to be sent per RTT. Perhaps I misread it. I think it makes sense, though, to try really hard to coalesce sent data into as large packets as possible.

thrnz:
I was basing that on what was in the original [RFC](https://datatracker.ietf.org/doc/html/rfc896) from 1984:

```
[...]

PR #7055, opened 2022-09-08
Change the Nagle algorithm to allow a single undersized packet in flight at a time

thrnz:
Currently, libtorrent will hold back an undersized uTP packet until all previously transmitted packets have been ack'd. This can add an extra RTT of delay when sending the last few bytes of a 'big' send (eg. a stream of data followed by a choke).

This suggested change will only hold back the undersized packet if there is currently an undersized packet in flight, essentially allowing one undersized packet to be sent per RTT. This would allow the tail of a 'big' send to go out immediatley without any extra delay.

This is done by noting the sequence number of the last undersized packet sent, as suggested [here
](https://datatracker.ietf.org/doc/html/draft-minshall-nagle-01)

See #6935

### DIFF
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -1523,11 +1523,12 @@ bool utp_socket_impl::send_pkt(int const flags)
 
 		// did we fill up the whole mtu?
 		// if we didn't, we may still send it if there's
-		// no bytes in flight
+		// no undersized packet currently in flight
 		if (m_bytes_in_flight > 0
 			&& int(p->size) < std::min(int(p->allocated), effective_mtu)
 			&& !force
-			&& m_nagle)
+			&& m_nagle
+			&& compare_less_wrap(m_acked_seq_nr, m_nagle_seq_nr, ACK_MASK))
 		{
 			// the packet is still not a full MSS, so put it back into the nagle
 			// packet
@@ -1559,11 +1560,12 @@ bool utp_socket_impl::send_pkt(int const flags)
 	if (m_bytes_in_flight > 0
 		&& int(p->size) < std::min(int(p->allocated), effective_mtu)
 		&& !force
-		&& m_nagle)
+		&& m_nagle
+		&& compare_less_wrap(m_acked_seq_nr, m_nagle_seq_nr, ACK_MASK))
 	{
 		// this is nagle. If we don't have a full packet
 		// worth of payload to send AND we have at least
-		// one outstanding packet, hold off. Once the
+		// one outstanding undersized packet, hold off. Once the
 		// outstanding packet is acked, we'll send this
 		// payload
 		UTP_LOGV("%8p: NAGLE not enough payload send_buffer_size:%d cwnd:%d "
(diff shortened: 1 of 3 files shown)


src/utp_stream.cpp
```diff
@@ -1513,11 +1513,12 @@ bool utp_socket_impl::send_pkt(int const flags)
...
 		// did we fill up the whole mtu?
 		// if we didn't, we may still send it if there's
-		// no bytes in flight
+		// no undersized packet currently in flight
 		if (m_bytes_in_flight > 0
 			&& int(p->size) < std::min(int(p->allocated), effective_mtu)
 			&& !force
-			&& m_nagle)
+			&& m_nagle
+			&& !compare_less_wrap(m_nagle_seq_nr, m_acked_seq_nr, ACK_MASK))
```
arvidn, line 1521: i `m_nagle_seq_nr` == `m_acked_seq_nr`, that means we don't have an outstanding undersized packet, right?
I believe this logic handles that case incorrectly. `compare_less_wrap()` would return false and we would hold off the packet.

If I understand correctly, this last condition should be:
```
&& compare_less_wrap(m_acked_seq_nr, m_nagle_seq_nr, ACK_MASK))
```

src/utp_stream.cpp
```diff
@@ -1549,11 +1550,12 @@ bool utp_socket_impl::send_pkt(int const flags)
 	if (m_bytes_in_flight > 0
 		&& int(p->size) < std::min(int(p->allocated), effective_mtu)
 		&& !force
-		&& m_nagle)
+		&& m_nagle
+		&& !compare_less_wrap(m_nagle_seq_nr, m_acked_seq_nr, ACK_MASK))
```
arvidn, line 1554: I believe the comment aboc also applies here

arvidn:
this is an interesting result of this test: https://github.com/arvidn/libtorrent/runs/8244601393?check_suite_focus=true#step:5:723

If it's not caused by the comparison issue, it would be good to understand where the redundant packet is coming from

thrnz:
Well spotted. I'll fix that up and see how that effects that unexpected packet.

thrnz:
Looks like its still there. Might have to do some more digging to work out whats going on.

arvidn:
the redundant packet counter is incremented on `utp_stream.cpp:2367`.
We receive a packet that's ACKing a packet that hasn't been sent. The socket is in `fin_sent` state, and when sending a `FIN`, the seq_nr is not incremented. So it looks like this is the `FIN` being acked, but it's ignored as an invalid ack.
The incoming packet is also a `FIN`.

arvidn:
I believe this change should be made as well. It's unrelated to your change except that they affect the simulation outcome:

```
diff --git a/src/utp_stream.cpp b/src/utp_stream.cpp
index 4d142daac..18a134ceb 100644
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -2353,7 +2353,7 @@ bool utp_socket_impl::incoming_packet(span<char const> b
        // Note that when we send a FIN, we don't increment m_seq_nr
        std::uint16_t const cmp_seq_nr =
                ((state() == state_t::syn_sent || state() == state_t::fin_sent)
-                       && ph->get_type() == ST_STATE)
+                       && (ph->get_type() == ST_STATE || ph->get_type() == ST_FIN))
                ? m_seq_nr : (m_seq_nr - 1) & ACK_MASK;
 
        if ((state() != state_t::none || ph->get_type() != ST_SYN)
```

arvidn:
feel free to incorporate that patch in this PR, or just fix up the test and I can add it afterwards

thrnz:
> feel free to incorporate that patch in this PR, or just fix up the test and I can add it afterwards

Will do.

I'll fix the expected packet counts in `test_utp.cpp` once #7056 is merged as it will likely effect the results.

arvidn (approved):

PR #7056, opened 2022-09-08
Pick up and send undersized uTP packets when a packet is 'forced'

thrnz:
Undersized (<MTU) uTP packets are held back in order to reduce overhead (ie. large headers relative to actual payload). Currently, libtorrent can 'force' send a uTP header-only packet without a payload (such as a deferred ack) while an undersized packet is being held back, negating any gains made from holding it back.

This suggested change will pick up and send the undersized packet when a uTP packet is being 'forced' out if there is room in the congestion window to do so.

See #6935

### DIFF
--- a/simulation/test_utp.cpp
+++ b/simulation/test_utp.cpp
@@ -136,10 +136,10 @@ TORRENT_TEST(utp_pmtud)
 	// TODO: 3 This timeout happens at shutdown. It's not very clean
 	TEST_EQUAL(metric(cnt, "utp.utp_timeout"), 1);
 
-	TEST_EQUAL(metric(cnt, "utp.utp_packets_in"), 611);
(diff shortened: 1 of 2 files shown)


arvidn:
this looks reasonable

arvidn:
there is at least one simulation test that counts exactly how many packets are sent in a few scenarios. You'll need to update those numbers to make the test pass. (and make sure those changes matches your understanding of what aught to happen with this patch).

arvidn:
see this test:
https://github.com/arvidn/libtorrent/runs/8244652336?check_suite_focus=true#step:5:701

This is `simulations/test_utp.cpp`

thrnz:
Those numbers look about right. An increase in payload packets transferred as the undersized packets are now piggy-backing on the previously 'ack only' packets, and a decrease in total packets transferred. I'll update the expected test numbers to match.

arvidn (approved):

PR #7059, opened 2022-09-09
support incoming SOCKS5 packets with hostnames as source address

arvidn:
for UDP trackers


arvidjaar:
I built libtorrent with PR #7027 and this one on top of Ubuntu 22.04 package and I confirm that it works. With packet capture I also see correct SOCKS5 exchange with `ATYP == 3` and tracker FQDN. Thank you!

PR #7069, opened 2022-09-11
use retry for macOS simulations

AllSeeingEyeTolledEweSew:
The simulations on macOS are flaky. Let's use a retry loop for them.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -69,9 +69,13 @@ jobs:
           echo "using darwin ;" >>~/user-config.jam
 
       - name: build and run simulations
-        run: |
-          cd simulation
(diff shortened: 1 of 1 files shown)


arvidn:
The flakiness of this test on MacOS is caused by a problem in the simulator. Something is not deterministic. I believe I've fixed these issues in RC_2_0 and later. I would like those more recent branches to fail if some non-determinism is introduced again.

arvidn:
so, I'm hesitant to land this. It would be better to back-port the determinism fixes

AllSeeingEyeTolledEweSew:
IMO this isn't any worse than ignoring a CI failure, but sure, backporting is great.

I'll take a crack at backporting these fixes, if you can tell me how I can identify them.

PR #7070, opened 2022-09-11
use retry for windows simulations

AllSeeingEyeTolledEweSew:
The simulations on Windows are flaky. Let's use a retry loop for them.

Same as #7069 but for Windows simulations on `RC_2_0`

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -97,6 +97,11 @@ jobs:
           cd boost
           .\b2 headers
 
+      - shell: pwsh
+        run: |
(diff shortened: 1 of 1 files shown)


arvidn:
the reason the simulations on windows fail sometimes in CI is a bit more mysterious. they appear to hang. So I suppose this makes more sense than the other patch, but it would also be really nice to understand why it hangs.

AllSeeingEyeTolledEweSew:
My understanding of Windows is poor, so I don't have much hope of tracking mysterious issues.

Until someone else does it, this patch will literally save me a lot of time scrolling through the list of CI checks to ensure the windows simulation one is the only one that failed

Issue #7068, opened 2022-09-11
uTP selective ACK payload size differs from BEP


thrnz:
[BEP 29](https://www.bittorrent.org/beps/bep_0029.html) says that the selective ACK 'payload is a bitmask of at least 32 bits, in multiples of 32 bits.' If I'm reading the code correctly, libtorrent appears to use an 8 bit minimum with 8 bit multiples when sending selective ACKs:

https://github.com/arvidn/libtorrent/blob/d5f74cb7edb489de5931f81f8269f4e718237949/src/utp_stream.cpp#L1370

Note that `sack` is the selective ACK payload size in bytes (the extension's `len` field in the BEP). `m_inbuf.span()` is in bits, so `sack` would need to be a multiple of 4. Maybe something like this:

`sack = ((m_inbuf.span() + 31) / 32) * 4;`

I'm not sure if non 32-bit multiples are an issue at all in practice though. I guess it depends on how strictly other clients might have been coded to handle it.

PR #7071, opened 2022-09-12
Modify uTP selective ACKs to conform to BEP

thrnz:
The size of libtorrent's uTP selective ACK payload appears to be rounded up to the nearest byte, however BEP 29 defines it as 'a bitmask of at least 32 bits, in multiples of 32 bits.'

This potential change conforms to the BEP by rounding up to the nearest 4 byte multiple. Note that `sack` is the SACK payload size in bytes (ie `len` from the BEP which 'must be at least 4, and in multiples of 4').

I'm not sure if this is actually an issue in practice at all, or if there are any clients out there in the wild that might have issues receiving SACKs that don't strictly conform.

See #7068

### DIFF
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -1357,6 +1357,9 @@ bool utp_socket_impl::send_pkt(int const flags)
 
 	auto const close_reason = static_cast<std::uint32_t>(m_close_reason);
 
+	// size in bytes of the sack payload
+	// according to BEP 29, when sending a sack this
+	// 'must be at least 4, and in multiples of 4'
 	int sack = 0;
 	if (m_inbuf.size())
 	{
@@ -1367,8 +1370,8 @@ bool utp_socket_impl::send_pkt(int const flags)
 
 		// the SACK bitfield should ideally fit all
 		// the pieces we have successfully received
-		sack = (m_inbuf.span() + 7) / 8;
-		if (sack > max_sack_size) sack = max_sack_size;
+		sack = ((m_inbuf.span() + 31) / 32) * 4;
+		if (sack > max_sack_size) sack = (max_sack_size / 4) * 4;
 	}
 
 	int const header_size = int(sizeof(utp_header))
@@ -1548,6 +1551,10 @@ bool utp_socket_impl::send_pkt(int const flags)
 		write_sack(ptr, sack);
 		ptr += sack;
 		TORRENT_ASSERT(ptr <= p->buf + p->header_size);
+#if TORRENT_UTP_LOG
+		UTP_LOGV("%8p: Including SACK payload size: %d bytes\n"
+			, static_cast<void*>(this), sack);
+#endif
 	}
 
 	if (close_reason)


arvidn:
uTorrent always send exactly 4 bytes: https://github.com/bittorrent/libutp/blob/master/utp_internal.cpp#L815
this is where it parses incoming SACK fields (also called EACK in uTorrent): https://github.com/bittorrent/libutp/blob/master/utp_internal.cpp#L1441

As far as I can tell, it doesn't require the SACK bitfield to be a multiple of 4 bytes.

simulation/test_utp.cpp
```diff
@@ -139,7 +139,7 @@ TORRENT_TEST(utp_pmtud)
 	TEST_EQUAL(metric(cnt, "utp.utp_packets_in"), 593);
 	TEST_EQUAL(metric(cnt, "utp.utp_payload_pkts_in"), 66);
 
-	TEST_EQUAL(metric(cnt, "utp.utp_packets_out"), 603);
+	TEST_EQUAL(metric(cnt, "utp.utp_packets_out"), 602);
```
arvidn, line 142: it's not clear to me why this change would result in one less packet to be sent. If anything, a larger SACK header would eat into the size of the payload that would fit in the packet

thrnz: I'd only updated that packet count in order to see if the other tests would still pass, which they seem to do.

I'm yet to work out why that particular test has one less packet, but that was my first thought as well - with the extra 3 bytes possible in the header, if anything it should be the other way round.

thrnz: There might be an issue when picking up a Nagled packet with a SACK already attached if the sack size has increased since it was held back. It appears to keep the old size, rather than growing it if needed. It looks like an existing issue that just happened to trigger on that one test with this PR compared to without.

In this case, the unpatched version misses out on acking a packet (the 1 byte SACK payload needs to grow to hold 9 bytes in this case but doesn't), while the patched version already has enough padding to grow (although it will be an issue growing a Nagled SACK from 32-33 bytes). As a result, the unpatched version sends an extra ack packet later on to catch up.

When comparing utp debug logfiles of the failing test using this PR against those without they appear to diverge right after that happens.

Hopefully if that gets fixed, then that test will pass without needing to update the packet count.

ValeZAA:
>uTorrent always send exactly 4 bytes:

Are you sure this is indeed code used by mutorrent? Also it is not like it is not full of bugs. Will this break compatibility with mutorrent?

ValeZAA:
Why is this bot closing stale pull requests and issues? That is not serious.

simulation/test_utp.cpp
ValeZAA: Can you attach the two pcap one with 603 packets, one with 602?

ValeZAA:
I just realised that BEP 29 was written by @arvidn probably by reading utorrent open source parts. So that means it is just a typo. Errata not possible in BEP?

PR #7074, opened 2022-09-13
bump boost version used in cibuildwheel for musllinux

AllSeeingEyeTolledEweSew:
This bumps the version of boost used when compiling python wheels on musllinux (alpine).

This segfaults when running the python tests. I've confirmed this happens with every boost version after 1.76.

@arvidn can you help me investigate?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ test-command = [
 [tool.cibuildwheel.overrides.environment]  # sub-table of previous block!
 BOOST_BUILD_PATH = "/tmp/boost/tools/build"
 BOOST_ROOT = "/tmp/boost"
-BOOST_VERSION = "1.76.0"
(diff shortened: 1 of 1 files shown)


AllSeeingEyeTolledEweSew:
I updated the boost version to 1.77 only to illustrate that's the version where it broke. Once fixed, it should be updated to 1.80.

arvidn:
it looks like there's still a segfault: https://github.com/arvidn/libtorrent/actions/runs/3033234023/jobs/4881323494#step:5:932

arvidn:
it looks like there's still a segfault: https://github.com/arvidn/libtorrent/actions/runs/3033234023/jobs/4881323494#step:5:932

arvidn:
oh, I see. this PR is meant to demonstrate the problem

PR #7075, opened 2022-09-13
bump boost versions and windows-* runner for cibuildwheel

AllSeeingEyeTolledEweSew:
* Bump the cibuildwheel windows runner environment to `windows-2022`
* Bump the boost version used in cibuildwheel to latest (1.80)
  * ...except for the musllinux wheel due to a segfault, See #7074

Bumping the boost version was necessary for `windows-2022`, to pick up support for msvc-14.3. I would have bumped it anyway since we should probably always be statically linking the wheel with the newest boost.

The random `HAVE_SNPRINTF` define is necessary with the newer boost version. See https://github.com/boostorg/system/issues/32#issuecomment-462912013. We never hit this previously this because it's triggered by a combination of newer boost *and* older python. Our regular windows python build CI currently uses python 3.9, whereas cibuildwheel always uses the oldest supported python (3.7)

### DIFF
--- a/bindings/python/src/boost_python.hpp
+++ b/bindings/python/src/boost_python.hpp
@@ -7,6 +7,8 @@
 
 #include <cstdio>
 #include <libtorrent/aux_/disable_warnings_push.hpp>
+// https://github.com/boostorg/system/issues/32#issuecomment-462912013
+#define HAVE_SNPRINTF
 #include <boost/python.hpp>
 
 #include <boost/bind/placeholders.hpp>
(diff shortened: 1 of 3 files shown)


bindings/python/src/boost_python.hpp
```diff
@@ -7,6 +7,8 @@
 
 #include <cstdio>
 #include <libtorrent/aux_/disable_warnings_push.hpp>
+// https://github.com/boostorg/system/issues/32#issuecomment-462912013
+#define HAVE_SNPRINTF
```
arvidn, line 11: I would be more comfortable with this being conditioned by an old version of `_MSC_VER`

arvidn (approved):

PR #7076, opened 2022-09-13
add a script to update openssl build scripts vendored from manylinux,…

AllSeeingEyeTolledEweSew:
… and run the script


AllSeeingEyeTolledEweSew:
That's right, for now. We could add some workflow to automatically run the update script and submit a PR on a cron job, but I thought I'd do that separately.

PR #7082, opened 2022-09-18
Fix uTP packets being timed out and resent before the minimum timeout has passed

thrnz:
Currently the time of the next uTP timeout is set on socket creation, when a packet is received, and every time the timeout itself happens. If a packet is sent just before the next timeout expires, which can happen on an otherwise idle connection with no regular incoming packets to reset it, it can be unnecessarily timed out and resent.

This patch fixes it by updating the next timeout when a packet is sent. It also moves updating the timeout in `incoming_packet()` to after sacks/acks have been read as this can effect the timeout duration.

The BEP has timeouts getting updated every time a packet is both sent and received:

> Every time a socket sends or receives a packet, it updates its timeout counter. If no packet has arrived within timeout number of milliseconds from the last timeout counter reset, the socket triggers a timeout.

### DIFF
--- a/simulation/test_torrent_status.cpp
+++ b/simulation/test_torrent_status.cpp
@@ -480,8 +480,8 @@ TORRENT_TEST(active_timer_no_seed)
 			{
 				// some part of the simulation is not deterministic, and causes this to vary
 				// between platforms/compilers
-				TEST_CHECK(active_time >= 10);
-				TEST_CHECK(active_time <= 14);
+				TEST_CHECK(active_time >= 6);
+				TEST_CHECK(active_time <= 9);
 			}
 
 			torrent_status st = handle.status();
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -1673,7 +1673,18 @@ bool utp_socket_impl::send_pkt(int const flags)
 	}
 
 	if (!m_stalled)
+	{
 		++p->num_transmissions;
+		// Only reset the timeout for the initial packet
+		if (m_bytes_in_flight == 0)
+		{
+			m_timeout = now + milliseconds(packet_timeout());
+#if TORRENT_UTP_LOG
+			UTP_LOGV("%8p: updating timeout to: now + %d\n"
+			, static_cast<void*>(this), packet_timeout());
+#endif
+		}
+	}
 
 	// Any queued up deferred ack is now redundant
 	if (m_deferred_ack)
@@ -2463,12 +2474,6 @@ bool utp_socket_impl::incoming_packet(span<char const> b
 
 	++m_in_packets;
 
-	// this is a valid incoming packet, update the timeout timer
-	m_num_timeouts = 0;
-	m_timeout = receive_time + milliseconds(packet_timeout());
-	UTP_LOGV("%8p: updating timeout to: now + %d\n"
-		, static_cast<void*>(this), packet_timeout());
-
 	// the test for INT_MAX here is a work-around for a bug in uTorrent where
 	// it's sometimes sent as INT_MAX when it is in fact uninitialized
 	const std::uint32_t sample = ph->timestamp_difference_microseconds == INT_MAX
@@ -2578,6 +2583,13 @@ bool utp_socket_impl::incoming_packet(span<char const> b
 		ptr += len;
 		extension = next_extension;
 	}
+
+	// this is a valid incoming packet, update the timeout timer
+	// do this after processing sacks/acks as that can effect packet_timeout()
+	m_num_timeouts = 0;
+	m_timeout = receive_time + milliseconds(packet_timeout());
+	UTP_LOGV("%8p: updating timeout to: now + %d\n"
+		, static_cast<void*>(this), packet_timeout());
 
 	// the send operation in parse_sack() may have set the socket to an error
 	// state, in which case we shouldn't continue
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -3102,18 +3102,15 @@ void utp_socket_impl::do_ledbat(const int acked_bytes, const int delay
 		, scaled_gain / double(1 << 16), int(m_cwnd >> 16)
 		, int(m_slow_start));
 
-	// if scaled_gain + m_cwnd <= 0, set m_cwnd to 0
-	if (-scaled_gain >= m_cwnd)
-	{
-		m_cwnd = 0;
-	}
+	// don't drop below 1*MSS. This behavior is from rfc6817 (LEDBAT). This differs
+	// from BEP 29 which allows cwnd to drop to 0, however this way avoids needing
+	// to wait until the next timeout to resume sending.
+	if ((m_cwnd + scaled_gain) >> 16 < m_mtu)
+		m_cwnd = std::int64_t(m_mtu) * (1 << 16);
 	else
-	{
 		m_cwnd += scaled_gain;
-		TORRENT_ASSERT(m_cwnd > 0);
-	}
 
-	TORRENT_ASSERT(m_cwnd >= 0);
+	TORRENT_ASSERT((m_cwnd >> 16) >= m_mtu);
 
 	int const window_size_left = std::min(int(m_cwnd >> 16), int(m_adv_wnd)) - in_flight + acked_bytes;
 	if (window_size_left >= m_mtu)


src/utp_stream.cpp
```diff
@@ -1673,7 +1673,14 @@ bool utp_socket_impl::send_pkt(int const flags)
 	}
 
 	if (!m_stalled)
+	{
 		++p->num_transmissions;
+		m_timeout = clock_type::now() + milliseconds(packet_timeout());
```
thrnz, line 1678: libutp appears to only update the timeout on sending when theres nothing already in flight, and also only for `ST_DATA` and `ST_FIN` packets. Adding a check for `m_bytes_in_flight == 0 && (p->size > p->header_size || (flags & pkt_fin))` before resetting the timeout could probabaly be done here if preferred.

https://github.com/bittorrent/libutp/blob/2b364cbb0650bdab64a5de2abb4518f9f228ec44/utp_internal.cpp#L992-L999

simulation/test_torrent_status.cpp
```diff
@@ -481,7 +481,7 @@ TORRENT_TEST(active_timer_no_seed)
 				// some part of the simulation is not deterministic, and causes this to vary
 				// between platforms/compilers
 				TEST_CHECK(active_time >= 10);
-				TEST_CHECK(active_time <= 14);
+				TEST_CHECK(active_time <= 15);
```
thrnz, line 484: This failed test seems to be one that expects to complete in ~14s, but now takes ~15s.

After some digging, it looks like at several stages do_ledbat() on the swarm peers sets `m_cwnd` to 0 due to excessive latency. It must then wait until the next timeout (1s) before its able to resume sending.

It's not uncommon for that to happen - it can be seen at multiple points in both the patched and unpatched test's `utp.log`. The test result logfiles become out of sync after a while when the unpatched one begins timing out and re-sending packets that the patched one no longer does, and by chance `m_cwnd` resetting to 0 appears to trigger one more time with this patch than without, delaying the test completing.

While pouring over the debug logs to figure that out I noted 12 lost packets and 5 resends in the unpatched test, though none in the patched one. So I guess that shows that it seems to be working as intended.

src/utp_stream.cpp
arvidn: I like adding that condition here, to only update the timeout if there are no outstanding packets already.

simulation/test_torrent_status.cpp
arvidn: you mean that `m_cwnd` is set to 1 MSS, right? (i.e. one packet size, not 0). That's supposed to happen when the send direction has become idle.

arvidn (commented):
I agree that the current logic is not great, only updating the timeout on incoming packets.

simulation/test_torrent_status.cpp
thrnz: https://github.com/arvidn/libtorrent/blob/b3d0148e641b94597c8ef132083e2226f9b0cefb/src/utp_stream.cpp#L3093-L3102

If latency is large enough to drop `m_cwnd` below 0, it looks like do_ledbat() sets it to 0. It then gets reset to the MSS on the next timeout (~1s). This wait for the timout appears to be whats causing the test to take 1s longer to complete.

From the BEP:
> If max_window becomes less than 0, it is set to 0. A window size of zero means that the socket may not send any packets. In this state, the socket will trigger a timeout and force the window size to one packet size, and send one packet. See the section on timeouts for more information.

Here it is in action. Note the first log message is printed just before its set to 0. The timestamps are microseconds:
```
[000013680048] 0x55f9ab26b940: do_ledbat delay:348978 off_target: -248978 window_factor:0.500000 target_factor:-2.489777 scaled_gain:-3734.664917 cwnd:2435 slow_start:0
[000013680048] 0x55f9ab26b940: no space in window send_buffer_size:423 cwnd:0 adv_wnd:1048576 in-flight:1451 mtu:1471
[000013680048] 0x55f9ab26b940: skipping send seq_nr:12652 ack_nr:30607 id:23865 target:50.0.0.1:6881 header_size:20 error:Success send_buffer_size:423 cwnd:0 adv_wnd:1048576 in-flight:1451 mtu:1471 effective-mtu:1471
...
[000013859928] 0x55f9ab26b940: do_ledbat delay:169458 off_target: -69458 window_factor:1.000000 target_factor:-0.694565 scaled_gain:-2083.694458 cwnd:0 slow_start:0
[000013859928] 0x55f9ab26b940: no space in window send_buffer_size:423 cwnd:0 adv_wnd:1048576 in-flight:0 mtu:1471
[000013859928] 0x55f9ab26b940: skipping send seq_nr:12652 ack_nr:30607 id:23865 target:50.0.0.1:6881 header_size:20 error:Success send_buffer_size:423 cwnd:0 adv_wnd:1048576 in-flight:0 mtu:1471 effective-mtu:1471
...
[000014500000] 0x55f9ab26b940: timeout num-timeouts: 0 max-resends: 3 confirmed: 1  acked-seq-num: 12651 mtu-seq: 0
[000014500000] 0x55f9ab26b940: resetting cwnd:1471
[000014500000] 0x55f9ab26b940: slow_start -> 1
[000014500000] 0x55f9ab26b940: sending packet seq_nr:12652 ack_nr:30607 type:ST_DATA id:23865 target:50.0.0.1:6881 size:443 error:Success send_buffer_size:0 cwnd:1471 adv_wnd:1048576 in-flight:0 mtu:1471 timestamp:15000000 time_diff:65120 mtu_probe:0 extension:0
[000014500000] 0x55f9ab26b940: updating timeout to: now + 500
```

Interestingly enough, I can't find this in the [LEDBAT rfc](https://datatracker.ietf.org/doc/rfc6817/) itself. There's a `MIN_CWND` floor with a suggested value of 2 (ie a 2*MSS floor) which it cannot fall below.

For curiosity I changed do_ledbat() to not lower `m_cwnd` below `m_mtu`, and that test now completes in 9.8s.

arvidn: that sounds like a very reasonable change (to clamp to 2 MSS)

src/utp_stream.cpp
thrnz: Do you think the timeout should only update on ST_DATA and ST_FIN packets as well? I'm not sure of the reasoning behind it yet, and would only be adding it as its what libutp does.

src/utp_stream.cpp
```diff
@@ -1673,7 +1673,18 @@ bool utp_socket_impl::send_pkt(int const flags)
 	}
 
 	if (!m_stalled)
+	{
 		++p->num_transmissions;
+		// Only reset the timeout for the initial packet
+		if (m_bytes_in_flight == 0 && (p->size > p->header_size || (flags & pkt_fin)))
+		{
+			m_timeout = clock_type::now() + milliseconds(packet_timeout());
```
arvidn, line 1681:
```suggestion
			m_timeout = now + milliseconds(packet_timeout());
```

arvidn: I think it would be better to use the timestamp we've already asked for a bit further up

src/utp_stream.cpp
```diff
@@ -3090,18 +3102,15 @@ void utp_socket_impl::do_ledbat(const int acked_bytes, const int delay
...
 
-	// if scaled_gain + m_cwnd <= 0, set m_cwnd to 0
-	if (-scaled_gain >= m_cwnd)
-	{
-		m_cwnd = 0;
-	}
+	// don't drop below 2*MSS. This behavior is from rfc6817 (LEDBAT). This differs
+	// from BEP 29 which allows cwnd to drop to 0, however this way avoids needing
+	// to wait until the next timeout to resume sending.
+	if ((m_cwnd + scaled_gain) >> 16 < 2 * m_mtu)
```
arvidn, line 3108: `m_cwnd` is `int64_t` fixed point with 16 bit fraction.
`scaled_gain` is also `int64_t` fixed point with 16 bits fraction.
`m_mtu` is `uint16_t` and will be promoted to `int`, which presumably is `int32_t`.
I don't see any risk of overflow here.

However, this does introduce an effective floor of the congestion window. I believe cwnd can still be set to 0 (or < 1) by normal packet loss logic, and the only way to recover is to wait for timeout, and reset cwnd to 1.

However, the odd thing with this logic is that it will immediately snap up cwnd to 2 MSS the first message we receive. Even if `scaled_gain` is positive, and is pushing cwnd *up*, this logic will push it all the way up to 2 MSS unconditionally. I initially got the impression this was meant to be a floor preventing adjusting cwnd *down* too far.

Perhaps it has to work this way, it just seems a bit inconsistent that a timeout would bring cwnd to 1 MSS but this would bring it to 2.

thrnz: I'll rework this into a hard floor for consistency throughout. This would bring the initial `m_cwnd` up to 2*MSS  as well.

LEDBAT clamps to `MIN_CWND` for both latency and loss, and uses `INIT_CWND` to begin with, both of which 'should' be 2*MSS.

simulation/test_torrent_status.cpp
```diff
@@ -480,8 +480,8 @@ TORRENT_TEST(active_timer_no_seed)
 			{
 				// some part of the simulation is not deterministic, and causes this to vary
 				// between platforms/compilers
-				TEST_CHECK(active_time >= 10);
-				TEST_CHECK(active_time <= 14);
+				TEST_CHECK(active_time >= 6);
+				TEST_CHECK(active_time <= 9);
```
thrnz, line 484: I guess the question now is whats a reasonable min and max time for this test to complete in now as the comment suggests that it may vary.

src/utp_stream.cpp
thrnz: On second thought, how about only using the floor if cwnd is shrinking. Otherwise, allow it to grow as normal. That way it won't get bumped up to 2*MSS prematurely, though it would if it were to shrink.

A simpler option might be to just use a minimum of 1*MSS here instead of 2, as is currently used elsewhere (on loss/timeout/initial).
	
This is the reasoning behind using a value of 2:

```
   INIT_CWND and MIN_CWND SHOULD both be 2.  An INIT_CWND of 2 should
   help seed FILTER() at the sender when there are no samples at the
   beginning of a flow, and a MIN_CWND of 2 allows FILTER() to use more
   than a single instantaneous delay estimate while not being too
   aggressive.  Slight deviations may be warranted, for example, when
   these values of INIT_CWND and MIN_CWND interact poorly with the
   framing protocol.  However, INIT_CWND and MIN_CWND MUST be no larger
   than the corresponding values specified for TCP [RFC5681].
```

Having a base of 2 for the INIT_CWND makes sense, to get more delay samples at the beginning, however as each uTP packet only contains a single one-way delay sample (LEDBAT seems to allow for combining multiple samples per packet), there may not be a benefit to using 2 instead of 1 for MIN_CWND.

arvidn (approved):
I think this is reasonable.

I was actually under the impression that packet loss could also bring cwnd to be less than 1 MSS, and force a delay of 1 timeout before resuming. It seems I was wrong.

Trying to make the floor only apply when shrinking would be subject to noise where the cwnd is trending up, but gets one sample ticking down slightly. That would still have caused it to instantly snap to 2 MSS. Or the logic could be more sophisticated to only apply when shrinking from > 2MMS to < 2MSS. Anyway, I think keeping this simple is good.

PR #7086, opened 2022-09-19
add upgrade_to_2.1 document

arvidn:


docs/upgrade_to_2.1.rst
```diff
@@ -0,0 +1,76 @@
...
+			* piece layers (v2 torrents)
+
+To load a torrent file from disk, use one of load_torrent_file(),
+load_torrent_buffer() or load_torrent_parsed(). Those functions return an
+add_torrent_params object containing all the information about the torrent.
+
+To save a torrent file from an active torrent session, first ask for the
+add_torrent_params object by calling save_resume_data() and pass in the
+save_info_dict flag. The resulting save_resume_data_alert will contain an
+add_torrent_params object which can be saved to disk using write_torrent_file().
```
glassez, line 55: It would still be nice to be able to serialize/parse torrent_info separately. This is useful for the purposes of storing internal application data to avoid rewriting immutable torrent_info. Do you mind providing such a way?

arvidn: `torrent_handle::torrent_file()` returns the `torrent_info` which can be serialized by `torrent_info::info_section()`. deserializing it can be done via `torrent_info::parse_info_section()`. That last function should probably be a free function and the main way to create a `torrent_info` object (other than parsing a full torrent file).

glassez: Ok 👍

docs/upgrade_to_2.1.rst
```diff
@@ -0,0 +1,76 @@
...
+To enable webtorrent support, build with ``webtorrent=on``.
+
+loading and saving torrents
+===========================
+
+The loading and saving of .torrent files and resume data has been simplified and
+responsibilities of classes have become clearer.
+
+	torrent_info
+		Represents the immutable info section of a torrent file. It includes:
```
glassez, line 37: I would also expect `torrent_info` to drop all unrelated functionality, e.g. providing information about effective (changed) filenames will be provided via a separate `file_layout` of torrent, and `torrent_info` will store only the original ones.

glassez: > I would also expect `torrent_info` to drop all unrelated functionality, e.g. providing information about effective (changed) filenames will be provided via a separate `file_layout` of torrent, and `torrent_info` will store only the original ones.

Ideally there should be no mutators in `torrent_info` interface.

arvidn: Yes, I intend to deprecate those in 2.1

PR #7093, opened 2022-09-20
add option to not memory map files below a certain size

arvidn:


AllSeeingEyeTolledEweSew:
@arvidn is there a way to test whether a file is mmapped? Should it be a `file_open_mode`?

arvidn:
Yeah, that might be a good idea.

Issue #6996, opened 2022-08-08
Inefficient synchronization way to obtain some data from libtorrent


glassez:
libtorrent version (or branch): both v1.2.x and v2.0.x

platform/architecture: any

compiler and compiler version: any

When receiving some data from libtorrent using synchronous calls (for example, using some kind of `torrent_handle` getter), too long delays are noticed from time to time, regardless of either the current system load or the parameters of the affected torrent.
It looks like either the synchronization code has some drawbacks, or it is implemented in a way that is unfriendly to the caller. Recently, we have tried to avoid synchronous calls as much as possible. Unfortunately, some of the mentioned data cannot be obtained asynchronously (for example, via `torrent_status`), so we still have to resort to synchronous way, which leads to GUI hangups.

glassez:
I tried to figure it out myself, but I'm not that familiar with the libtorrent codebase. Besides, I've never used Boost.Asio to understand it enough. I have done some educational work to understand its basics, but my judgments below may still be erroneous. Here's what I realized:

When some better of `torrent_handle` is called, it obtains a pointer to handled `torrent` instance and calls the corresponding method of it using `async_call_ret`:
https://github.com/arvidn/libtorrent/blob/40fc5ce16a522dbc67d383c239a7e960d6521c39/src/torrent_handle.cpp#L172
But, IIRC, it isn't called directly and just posted in some queue and then It'll invoked in libtorrent 
session thread:
https://github.com/arvidn/libtorrent/blob/40fc5ce16a522dbc67d383c239a7e960d6521c39/src/torrent_handle.cpp#L187
So, if I'm not mistaken, it will be executed only after all the previous callbacks from the queue have been executed. At the same time, the calling thread is blocked until the callback is completed, and if the libtorrent is currently busy with some heavy work, we get the delay mentioned above.
If it works as I described above, then this is clearly an inefficient and unfriendly way from the caller perspective. After all, it turns out that caller is forced to wait until all the current work is done, even if it doesn't affect the resources requested by caller. **Wouldn't it be more efficient to do this in a classic way, executing all the code in the caller thread and locking only the requested data using some kind of mutexes?**

Chocobo1:
>Wouldn't it be more efficient to do this in a classic way, executing all the code in the caller thread and locking only the requested data using some kind of mutexes?

I would guess, the work queue would be easier to implement/work correctly than to put mutex on the related data and debug when it dead locks...

glassez:
[...]

PR #7094, opened 2022-09-20
Update state when need to save resume data

glassez:
In #7063 I incorrectly recognized the place where `m_need_save_resume_data` is being set, so now I have to fix it.

### DIFF
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -581,7 +581,9 @@ namespace libtorrent {
 
 		void set_need_save_resume()
 		{
+			if (m_need_save_resume_data) return;
 			m_need_save_resume_data = true;
+			state_updated();
 		}
 
 		bool is_auto_managed() const { return m_auto_managed; }
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -10038,8 +10038,7 @@ namespace {
 
 		// these counters are saved in the resume data, since they updated
 		// we need to save the resume data too
-		m_need_save_resume_data = true;
-		state_updated();
+		set_need_save_resume();
 
 		// if the rate is 0, there's no update because of network transfers
 		if (m_stat.low_pass_upload_rate() > 0 || m_stat.low_pass_download_rate() > 0)


include/libtorrent/torrent.hpp
```diff
@@ -582,6 +582,7 @@ namespace libtorrent {
 		void set_need_save_resume()
 		{
 			m_need_save_resume_data = true;
+			state_updated();
```
arvidn, line 585: this makes me a bit nervous. Right now some care has been taken to not post updates for torrents willy nilly, but only on certain ("important") updates.

I suppose it's odd to consider the total upload/download counters important enough to mark the torrent as needing to have its resume data saved, but not important enough to be posted.

One alternative would be to save resume data for all torrents but set the flag to only actually do it if it needs the resume data saved. It would still be a problem that `torrent_status` includes the `need_save_resume` member though.

glassez: > I suppose it's odd to consider the total upload/download counters important enough to mark the torrent as needing to have its resume data saved, but not important enough to be posted.

But such behavior makes it confusing. It should be posted on any member of `torrent_status` updated.


> It would still be a problem that `torrent_status` includes the `need_save_resume` member though.

I would consider to remove any member that doesn't affect state update.

> One alternative would be to save resume data for all torrents but set the flag to only actually do it if it needs the resume data saved.

Apparently, we will have to turn to this method.

P.S. There wouldn't be such a big problem with it if `need_save_resume_data` could be obtained without waiting in queue until libtorrent done all current jobs.

arvidn: would it work to turn this around and set `m_need_save_resume_data = true;` in `state_update()`? that seems to make more sense

glassez: > that seems to make more sense

Sorry, I don't see the point. It makes the presence of `torrent_status::need_save_resume` meaningless (at least in a scenario avoiding blocking calls), because it will be `true` whenever `state_update_alert` is received (i.e., just always in the above scenario). But at the same time, we still will not be able to find out when it was really needed, because it does not respond to every change in this property.

IMO, the ideal logic is:
1. Set `need_save_resume_data` whenever some part of resume data is changed since last save/torrent adding. Seems it works correctly.
2. "Torrent status" should be marked as updated whenever some part of it (effectively some part of `torrent_status`) is changed. Currently it works inconsustently/incorrectly.

If you consider changing of some existing field of torrent_status shouldn't trigger it to be marked as "updated" then such a part clearly should be removed from it.

glassez: > > One alternative would be to save resume data for all torrents but set the flag to only actually do it if it needs the resume data saved.
> 
> Apparently, we will have to turn to this method.

The only thing that worries me is that even a non-blocking method of `torrent_handle` is more expensive than checking a boolean variable, so if there are many torrents (several thousand), this may have a significant overhead.

arvidn: right, `need_resume_data` indicates whether there would be any updates to the resume data itself, but `torrent_status` covers a sligthly different set of states, I suppose. Let's say there's some state that might change that only updates the `torrent_status`, but wouldn't affect the resume data; my patch would make that case worse.

The case that concerns you is where some state updates that affects the resume data, but doesn't affect any other aspect of `torrent_status` (and hence won't post it, despite `need_resume_data` was updated).

I think your patch make sense with that in mind. Do you know of any other inconsistencies to your description of how it should work (which sound right to me)?

> The only thing that worries me is that even a non-blocking method of torrent_handle is more expensive than checking a boolean variable, so if there are many torrents (several thousand), this may have a significant overhead.

I think the apples-to-apples comparison would be the cost of:

1. posting and copying a `torrent_info` object with `need_save_resume` set, and then check that bool.
2. posting an async save_resume message to the main thread, it checking `m_need_save_resume` and either posts it or posts a `save_resume_data_alert` or `save_resume_data_failed_alert`

It's not obvious which one would scale better. It probably depends on what the common case is, that thousands of torrents are continuously active (and need their resume data saved frequently) or whether the majority of the thousand torrents are inactive.

arvidn:
I think the function should be changed to something like this:

```
void set_need_save_resume()
{
	if (!m_need_save_resume_data) return;
	m_need_save_resume_data = true;
	state_updated();
}
```

include/libtorrent/torrent.hpp
glassez: > 1. posting and copying a `torrent_info` object with `need_save_resume` set, and then check that bool.

I don't quite understand what you mean.
In our scenario, we have torrent statuses updated periodically (with sufficient frequency to consider them constantly up-to-date). So if there were no problems with `torrent_status::need_save_resume`, then the cost comparison in our case would be "checking an existing boolean variable against a `torrent_handle::save_resume_data()` method call.


> It's not obvious which one would scale better. It probably depends on what the common case is, that thousands of torrents are continuously active (and need their resume data saved frequently) or whether the majority of the thousand torrents are inactive.

Given the negligible cost of checking a Boolean variable, we have the following:
1. Even if all N torrents are really active it adds N Boolean checking which is not so big addition to N `torrent_handle::save_resume_data()` calls
2. With a decrease in the number of active torrents compared to the total number, it begins to clearly win against the method with an unconditional `torrent_handle::save_resume_data()` call for all torrents (and the more it wins, the greater this difference is).

glassez: > I think your patch make sense with that in mind. Do you know of any other inconsistencies to your description of how it should work (which sound right to me)?

I haven't set myself the task of finding all of them yet. I came across this one in the process of performing other tasks.
As far as I remember, there was once a similar problem with `torrent_status::moving_storage`.

arvidn: > In our scenario, we have torrent statuses updated periodically (with sufficient frequency to consider them constantly up-to-date).

Assuming you do this by calling `post_torrent_updates()`, the copy I'm talking about is where you handle the `state_update_alert`. The trade-off being that with this patch (and the previous one) you'll now get updates for torrents where `need_save_resume` was the *only* thing that changed. That's the additional cost (the scenario where other states also changed aren't affected, since you would have received those updates regardless).

So, the additional cost of updating the bool that you're checking, is one more `torrent_status` in the list of updates (per torrent).

That said, I'm convinced that it's definitely better to make this work as expected and as regular as possible.

Also, and unrelated, the `need_save_resume` bool could probably be an atomic and be read from any thread lock-free.

arvidn:
oops, typo. You spotted it though

arvidn:
@glassez the submodule got updated accidentally

include/libtorrent/torrent.hpp
glassez: > Also, and unrelated, the `need_save_resume` bool could probably be an atomic and be read from any thread lock-free.

👍 
It is related to #6996.

glassez:
> @glassez the submodule got updated accidentally

Oops... I'm not so familiar with submodules. How to fix it easier?

arvidn:
@glassez 
if you diff against `RC_2_0` you should see the before and after commit hashes for the submodule.

```
git diff RC_2_0
```

Copy the before-commit hash and then do:

```
cd simulation/libsimulator
git checkout <hash>
cd ../..
git commit -a --amend
```

arvidn (approved):

include/libtorrent/torrent.hpp
glassez: > The trade-off being that with this patch (and the previous one) you'll now get updates for torrents where `need_save_resume` was the _only_ thing that changed.

This case seems doubtful to me. I have long wanted to understand the relationship between "torrent status" and "resume data", but I still did not get around to dealing with it in detail.
From an abstract point of view, there should just be some set of mutable properties/data of the torrent that represent its current "status". These properties/data can be used in the current session for some purposes (including the one of showing them to the user in the UI), or saved between sessions to "resume" the torrent the next time the application is launched. Yes, most likely there are some data that does not need to be saved between sessions, but I doubt that there should be some that only need to be saved without providing for other purposes as part of the torrent "status".
Of course, the above is an abstract point of view, so in reality there may be some exceptions made for optimization purposes, etc.


> I suppose it's odd to consider the total upload/download counters important enough to mark the torrent as needing to have its resume data saved, but not important enough to be posted.

What kind of counters exactly do you mean? Are they really not part of `torrent_status` struct as well?

glassez: Anyway, here are a couple of ideas:
1. Drop `torrent_status::need_save_resume` field and implement `need_save_resume_data_alert` that is posted once corresponding property of torrent becomes `true`.
2. Implement `session::save_resume_data()` that will do the job for all the torrents (to avoid multiple calls to `torrent_handle::save_resume_data()`. It may even have some optimizing options, for example, alternating resume data collection with the execution of other jobs.

PR #7099, opened 2022-09-22
fix flaky tests

arvidn:


AllSeeingEyeTolledEweSew:
Is this meant to fix the "Windows / Simulations" tests? If so, perhaps this PR could remove the CI retry.

arvidn:
> Is this meant to fix the "Windows / Simulations" tests? If so, perhaps this PR could remove the CI retry.

it's not

Issue #6532, opened 2021-10-19
Backport the python type stubs from master to RC_1_2 and RC_2_0


AllSeeingEyeTolledEweSew:
I'd like for https://github.com/AllSeeingEyeTolledEweSew/tvaf to depend on libtorrent's python type stubs.

But the type stubs are only in `master`, and it looks like there won't be releases from `master` for the foreseeable future. I think they're useful enough to be worthy of backporting.

Backporting the stubs is easy, but the existing `test.py` won't exercise them thoroughly. I would greatly prefer to backport the rest of the test infrastructure from `master` along with the type stubs.

I think this is a larger symptom of #6531. We should backport other python binding infrastructure changes, and just be sure to include the type stubs.

This is a separate issue from #6531 because the type stubs are technically new functionality.

AllSeeingEyeTolledEweSew:
@arvidn could you reopen this? it's been requested by other users (#6730)

arvidn:
is there a way to ensure type stubs stay in sync with the bindings? just a test that fails would be helpful.

AllSeeingEyeTolledEweSew:
The `mypy` config in `.pre-commit-config.yaml` type-checks the python test suite against the type stubs. That's as good as we can do right now.

This is an extra incentive to make the test suite extremely complete. All the lines in the `master` test suite like `assert isinstance(some_enum_value, int)` are there to exercise the type stubs more than to exercise the runtime types of every enum value.

The only way to do better would be to generate type stubs rather than handcraft them. There's an official `stubgen` tool that generates stubs from python extensions. But to work well it relies on some runtime introspection, and this introspection is particularly mangled in boost.python. I know pybind11 (a boost.python alternative) puts more effort into the introspection of its output, so I believe it will work better with stubgen, but I haven't tested this. I believe pybind11 has a number of advantages over boost.python so I want to explore it more, but this won't happen for a while.

Issue #5763, opened 2020-12-21
setup.py using b2 has broken make installs on RC_1_2


liaralabs:
**Please provide the following information**

libtorrent version (or branch):
RC_1_2 at b87d01292a11af2d3dad1c91dbd27bd90d04350a

platform/architecture:
amd64
Debian 10

compiler and compiler version:
gcc8.3

please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

I understand boost builds are the future of this project and I am working on migrating my build process over to `b2`; however, this is quite an undertaking and I haven't yet had the time to devote to rewriting the scripts to make this happen. Thus, my build scripts are currently utilizing `make` to build libtorrent as they have for some time.

The biggest issue I can see right now is that `make` produces a library called `libtorrent-rasterbar` and `b2` produces a library called `libtorrent`.

As it stands, if you configure and make libtorrent with the python binding while using `make`, the compiler will go through and make the `libtorrent-rasterbar` library then call `b2` once setup.py is called, which then goes through the trouble of recompiling the entire `libtorrent` library followed by the python shared object, but it will then link the shared object to `libtorrent.so.1.2.11` which will cause an import failure due to the installed library being called `libtorrent-rasterbar.so`.

Reproducible build method which is now failing causing an import error due to `libtorrent.so` not found:

```
./autotool.sh
PYTHON=python3 ./configure --enable-python-binding --with-libiconv --with-boost-python=boost_python3 CXXFLAGS=-std=c++14
make -j$(nproc)
make install
python3 -c "import libtorrent"
ImportError: libtorrent.so.1.2.11: cannot open shared object file: No such file or directory
```
As it stands, the python binding can no longer be created correctly while utilizing `make`, which seems like a very big change for a minor point release. Also, it seems like cxxflags and the `-j` make argument are not being passed to the python build command either.

Attached is the build log and final result/errors given by the steps above.

[make.log](https://github.com/arvidn/libtorrent/files/5726074/make.log)

arvidn:
I believe covering this in CI went away with travis. I forgot to convert this to github actions

arvidn:
does this fix it for you? https://github.com/arvidn/libtorrent/pull/5767

liaralabs:
Unfortunately, it looks like I'm still receiving the same error at the end of the steps above:

```
root@debian-4gb-hel1-1:/tmp/libtorrent# python3 -c "import libtorrent"
Traceback (most recent call last):
[...]

PR #7104, opened 2022-09-23
merge #7073 into `RC_2_0`

AllSeeingEyeTolledEweSew:
This merges #7073 into `RC_2_0`.

The intent of that change was that `setup.py` be unified across branches, and `setup.cfg` would contain branch-specific parts.

That's a bit complicated so I made this PR to make it clear how I thought this should work.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -7,8 +7,8 @@ import distutils.command.install_data as install_data_lib
 import distutils.debug
 import distutils.errors
 import distutils.sysconfig
-import distutils.util
 import functools
+import itertools
 import os
 import pathlib
 import re
@@ -16,6 +16,7 @@ import shlex
 import subprocess
 import sys
 import sysconfig
+import tempfile
 from typing import Callable
 from typing import cast
 from typing import IO
@@ -24,6 +25,7 @@ from typing import List
 from typing import Optional
 from typing import Sequence
 from typing import Set
+from typing import Tuple
 import warnings
 
 import setuptools
@@ -34,6 +36,25 @@ def b2_bool(value: bool) -> str:
     if value:
         return "on"
     return "off"
+
+
+def b2_version() -> Tuple[int, ...]:
+    # NB: b2 --version returns exit status 1
+    proc = subprocess.run(
+        ["b2", "--version"], stdout=subprocess.PIPE, universal_newlines=True
+    )
+    # Expected output examples:
+    #   Boost.Build 2015.07-git
+    #   B2 4.3-git
+    m = re.match(r".*\s([\d\.]+).*", proc.stdout)
+    assert m is not None, f"{proc.stdout} doesn't match expected output"
+    result = tuple(int(part) for part in re.split(r"\.", m.group(1)))
+    # Boost 1.71 changed from YYYY.MM to version 4.0. Return an "epoch" as the first
+    # part of the tuple to distinguish these version patterns.
+    if result[0] > 1999:
+        return (0, *result)
+    else:
+        return (1, *result)
 
 
 # Frustratingly, the "bdist_*" unconditionally (re-)run "build" without
@@ -219,8 +240,14 @@ class LibtorrentBuildExt(build_ext_lib.build_ext):
         (
             "cxxstd=",
             None,
-            "(DEPRECATED; use --b2-args=cxxstd=...) "
             "boost cxxstd value (14, 17, 20, etc.)",
+        ),
+        (
+            "configure-from-autotools",
+            None,
+            "(DEPRECATED) "
+            "when in --config-mode=distutils, also apply cxxflags= and linkflags= "
+            "based on files generated from autotools",
         ),
     ]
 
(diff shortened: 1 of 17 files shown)


bindings/python/setup.py
```diff
@@ -448,10 +497,43 @@ def _configure_b2_with_distutils(self) -> Iterator[None]:
...
+    def _find_project_config(self) -> Optional[pathlib.Path]:
+        for directory in itertools.chain(
+            (PYTHON_BINDING_DIR,), PYTHON_BINDING_DIR.parents
+        ):
+            path = directory / "project-config.jam"
+            if path.exists():
+                return path
+        return None
+
+    def _configure_from_autotools(self) -> None:
```
arvidn, line 514: did you intend to keep the autotools logic here? It was removed in `RC_2_0`

AllSeeingEyeTolledEweSew: I say leave it in. Then `setup.py` is identical across branches and any changes can be merged everywhere. There's not much downside to leaving it, since it spams a deprecation warning and doesn't have any effect in `RC_2_0` anyway.

arvidn: The cost of leaving it in is maintenance cost of recent branches, the value of leaving it in is maintenance cost in older branches.

It sounds like you anticipate more maintenance of setup.py in RC_1_2 than I do

AllSeeingEyeTolledEweSew: The driving force of the complexity of this file is the "quirks" in boost.build that vary across version, platform and distro. These apply to all branches.

AFAICT boost.build's design is highly ad-hoc so I see no way to be confident we've covered all relevant features, or all aspects of any given one. Git history shows I've iterated a lot on each feature.

So I anticipate endless maintenance of `setup.py` in `RC_1_2`, as long as boost.build is our primary build system.

Today with the branches out of sync, I have to figure out how a fix should apply in each branch, test each, and communicate to you how the `RC_1_2` PR should look when merged to later branches. Each of those steps is time-consuming and brittle. It's discouraging.

IMO branch-specific code is a small cost to gain easier maintenance across all branches.

I propose we follow the pattern I established here: branch-specific features get isolated in their own function, which is enabled by `setup.cfg` in its branch.

AllSeeingEyeTolledEweSew:
@arvidn: could you let me know whether you intend to merge this PR, or else use it as a template when doing the merge yourself, or else there's some other way I can make merges easier?

I am working on #6532. It's quite complicated, since it needs to start in `RC_1_2` then be merged with care to later branches, just like #7073.

I was hoping to use this PR as a template to inform how I can make merges easier.

My PRs against libtorrent tend to stay open for a long time. I assume I am making it difficult for you, since I do large changes which are difficult to split up.

I want to figure out what I can do to make this process easier, so these changes can happen more quickly.

bindings/python/setup.py
arvidn: > So I anticipate endless maintenance of setup.py in RC_1_2, as long as boost.build is our primary build system.

I don't intend to maintain `RC_1_2` indefinitely. (just like 1.1 and earlier aren't maintained)

> IMO branch-specific code is a small cost to gain easier maintenance across all branches.
> 
> I propose we follow the pattern I established here: branch-specific features get isolated in their own function, which is enabled by setup.cfg in its branch.

If the branch specific functions are separated out, it should be fairly easy to merge even if the outdated functionality is gone in future branches. I would expect any conflicts to be easily resolved at least.

I'm not excited about having autoconf logic remain in 2.0+, It's misleading. If you think it's unreasonable to only keep it in RC_1_2, would it be reasonable to remove it alltogether, from RC_1_2 also?

AllSeeingEyeTolledEweSew: > I don't intend to maintain `RC_1_2` indefinitely. (just like 1.1 and earlier aren't maintained)

The point is that most build infra changes should go on all active branches, so should start on `RC_1_2` and be merged through
 
> If the branch specific functions are separated out, it should be fairly easy to merge even if the outdated functionality is gone in future branches. I would expect any conflicts to be easily resolved at least.

Merge conflicts are a manual process. So far, they've been prone to creating spurious differences between branches.

We could audit the branches so they only contain expected changes between them, but that's a manual process.

The discipline of maintaining branch-specific code in isolated functions is a manual process. As soon as someone does a PR against `setup.py` without this discipline and you merge it, the discipline is lost.

Gaps in any of the manual processes create more spurious differences across branches, which create more merge conflicts, which make changes more expensive.

*Maybe* all the manual processes can be followed perfectly until `RC_1_2` is deprecated and autotools logic can be deleted. But it's hard to convince myself that's true when it hasn't been true so far.

Keeping `setup.py` identical on all branches allows for automatic merges in place of these manual processes.

I agree that intentionally keeping dead code is surprising, but it just seems like a small cost to pay, compared to how costly it is to do build system changes today. (#7073 was written 3 weeks ago).

Maintaining three active branches of such an enormous codebase is hard. I don't think it's crazy to pay some costs to make it easier.

All that said, if you insist on removing autotools logic from `RC_2_0`, I'll make that change. But only if you insist :)

> I'm not excited about having autoconf logic remain in 2.0+, It's misleading. If you think it's unreasonable to only keep it in RC_1_2, would it be reasonable to remove it alltogether, from RC_1_2 also?

I do think we should drop autotools from `setup.py`. I did that when I first started working on `setup.py`, it led to #5763

IMO, breaking changes to the build system in a point release is a defensible decision, but you're the one who would have to do the defending.

AllSeeingEyeTolledEweSew:
Closed in favor of #7163

PR #7107, opened 2022-09-25
Fix triggering loss for uTP packets that were never sent

thrnz:
https://github.com/arvidn/libtorrent/blob/8786d17e59c90aee2cb66db891d3a16c7f369ad7/include/libtorrent/aux_/utp_stream.hpp#L964-L969

A uTP packet that fails to send due to a 'stall' isn't marked as `need_resend`. The send is only retried once the packet has timed out or is picked up by a SACK, causing it to trigger loss and the `cwnd` to be halved despite never having been sent yet. This also means packets with a higher `seq_nr` may be sent out first if the socket becomes writable before the unsent packet has timed out/been picked up by SACK.

From this utp log note that 6542-6544 stall while 6545-6594 were sent as normal as the socket became writable again. 6542-6544 were only sent once they are noticed as missing from a SACK, and are then treated as lost packets:

```
[000026139017] 0000028257BE58C0: sending packet seq_nr:6542 ack_nr:27277 type:ST_DATA id:15964 target:193.176.124.10:37495 size:1457 error:The operation completed successfully send_buffer_size:16315 cwnd:82107 adv_wnd:78873 in-flight:76161 mtu:1457 timestamp:3248500762 time_diff:3078327667 mtu_probe:0 extension:0
[000026139091] 0000028257BE58C0: socket stalled
...
[000026139136] 0000028257BE58C0: sending packet seq_nr:6543 ack_nr:27277 type:ST_DATA id:15964 target:193.176.124.10:37495 size:1457 error:The operation completed successfully send_buffer_size:14878 cwnd:82107 adv_wnd:81747 in-flight:74724 mtu:1457 timestamp:3248500879 time_diff:3078328093 mtu_probe:0 extension:0
[000026139142] 0000028257BE58C0: socket stalled
...
[000026139221] 0000028257BE58C0: sending packet seq_nr:6544 ack_nr:27277 type:ST_DATA id:15964 target:193.176.124.10:37495 size:1457 error:The operation completed successfully send_buffer_size:13454 cwnd:82107 adv_wnd:81747 in-flight:76161 mtu:1457 timestamp:3248500967 time_diff:3078328093 mtu_probe:0 extension:0
[000026139228] 0000028257BE58C0: socket stalled
...
[...]

PR #7113, opened 2022-09-27
Resend uTP ACKs that fail to send due to stalling

thrnz:
Following #7107, uTP packets with payload that failed to send due to a stalled socket should now be sent once the socket becomes writable again. However, this doesn't apply to `ST_STATE` packets (ie payload-less ACKs), which would simply be lost. If another packet isn't sent out in time (updating our `ack_nr` at the other end) then this will be seen as loss and can trigger a needless resend by the other-end.

The idea behind this change is to note that an ack is still needed if the socket stalls while sending an empty payload ACK-only packet. Once the socket is writable again, the deferred ack is then sent out if its still needed.

I wonder if a simulation with a low enough `send_socket_buffer_size` might be able to trigger this and to confirm that this fix as well as #7107 work as intended.

### DIFF
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -962,7 +962,10 @@ void utp_socket_impl::writable()
 	m_stalled = false;
 	if (should_delete()) return;
 
-	while(send_pkt());
+	// if the socket stalled while sending an ack then there will be a
+	// pending deferred ack. make sure it gets sent out
+	if (!m_deferred_ack || send_pkt(pkt_ack))
+		while(send_pkt());
 
 	maybe_trigger_send_callback();
 }
@@ -1686,17 +1689,19 @@ bool utp_socket_impl::send_pkt(int const flags)
 			, static_cast<void*>(this), packet_timeout());
 #endif
 		}
+		// Any queued up deferred ack is now redundant
+		if (m_deferred_ack)
+		{
+	#if TORRENT_UTP_LOG
+			UTP_LOGV("%8p: Cancelling redundant deferred ack\n"
+				, static_cast<void*>(this));
+	#endif
+			m_deferred_ack = false;
+		}
 	}
-
-	// Any queued up deferred ack is now redundant
-	if (m_deferred_ack)
-	{
-#if TORRENT_UTP_LOG
-		UTP_LOGV("%8p: Cancelling redundant deferred ack\n"
-			, static_cast<void*>(this));
-#endif
-		m_deferred_ack = false;
-	}
+	// If this is an ack then defer it to when the socket becomes writable again
+	else if (p->size == p->header_size && (flags & pkt_ack))
+		defer_ack();
 
 	// if we have payload, we need to save the packet until it's acked
 	// and progress m_seq_nr


arvidn:
I've added it to my list to add a new uTP simulation where the send socket buffer is filled up. I'm not sure the current libsimulator models this yet.

src/utp_stream.cpp
```diff
@@ -963,6 +963,10 @@ void utp_socket_impl::writable()
 	if (should_delete()) return;
 
 	while(send_pkt());
+	// if the socket stalled while sending an ack, then send it
+	// now if needed
+	if (m_deferred_ack && !m_stalled)
+		send_ack();
```
arvidn, line 969: maybe I'm misunderstanding something here. but if the `send_pkt()`-loop above ended up sending *any* packets, we don't need to send a dedicated ACK packet. Since the ACK sequence number is always included in the header.

I would have expected this part of the patch to instead do something like: if `m_deferred_ack` is set, make the first call to `send_pkt()` be called with the `pkt_ack` flag. If it returns true, then move on to run the loop that attempts to drain the send buffer.

looking closer i see that `m_deferred_ack` is actually cleared by `send_pkt()` if it ends up sending a packet, so I suppose this also works. However, if `send_pkt()` doesn't send any packets (and doesn't clear `m_deferred_ack`), there shouldn't be a way for `m_stalled` to have been set either.

I'm leaning towards thinking that this code might be slightly simpler if it doesn't rely on`m_deferred_ack` and `m_stalled` being updated by the call to `send_pkt()`.

src/utp_stream.cpp
```diff
@@ -1687,9 +1691,12 @@ bool utp_socket_impl::send_pkt(int const flags)
 #endif
 		}
 	}
+	// If this is an ack then note that it still needs to be sent
+	else if (p->size == p->header_size && (flags & pkt_ack))
+		defer_ack();
 
 	// Any queued up deferred ack is now redundant
-	if (m_deferred_ack)
+	if (m_deferred_ack && !m_stalled)
```
arvidn, line 1699: this if-block should probably just be moved into the `if (!m_stalled)` block above

arvidn (commented):
looks good. just some thoughts

src/utp_stream.cpp
```diff
@@ -962,7 +962,10 @@ void utp_socket_impl::writable()
 	m_stalled = false;
 	if (should_delete()) return;
 
-	while(send_pkt());
+	// if the socket stalled while sending an ack then there will be a
+	// pending deferred ack. make sure it gets sent out
+	if (!m_deferred_ack || send_pkt(pkt_ack))
+		while(send_pkt());
```
thrnz, line 968: How about this: If there's no outstanding unsent ack then go straight to the `send_pkt()` loop like normal, otherwise use the `pkt_ack` flag for the first call, then proceed into the loop if needed. It's much simpler, and I think will achieve the same thing.

src/utp_stream.cpp
thrnz: > this if-block should probably just be moved into the `if (!m_stalled)` block above

Done

src/utp_stream.cpp
arvidn: that sounds good

PR #7114, opened 2022-09-27
No deprecated ci

arvidn:


AllSeeingEyeTolledEweSew:
Should probably be in `RC_1_2` as well

PR #7115, opened 2022-09-27
make python binding build without deprecated functions

arvidn:


bindings/python/tests/alert_test.py
```diff
@@ -2026,7 +2026,7 @@ def test_picker_log_alert(self) -> None:
         self.assert_peer_alert(
             alert, self.peer_endpoint, fingerprint=self.peer_fingerprint
         )
-        self.assertIsInstance(alert.picker_flags, lt.picker_flags_t)
```
AllSeeingEyeTolledEweSew, line 2029: Why does this type weakening happen? It's not obvious from the code change.

Is there no way to keep the strong type?

If it's really necessary to weaken it, it should change in the type stubs too

arvidn: generally, strong types aren't exposed to python. I'm not sure why this was in the first place, perhaps because the converters to and from `int` were missing.

arvidn: adding the converters was probably the change causing this. the change highlighting that the converters were missing was the change from using a boost.python `enum` to bind these flags to using a regular scope and adding them as variables to that scope.

AllSeeingEyeTolledEweSew: The converter change is necessary due to the fact that `static_cast<long>(...)` of a `bitfield_flag` doesn't work with `deprecated-functions=off`, and boost.python `enum` uses that cast internally, do I have that right?

Is there no better way to do this? It's unfortunate to cause a python API change over such an issue. The boost.python `enum` types has `names` and `values` dictionaries which are especially helpful in python, whereas their scope objects do not.

I understand the `deprecated-functions=off` build is important to check the path to eventual deprecation, but I expect real users of this build for python are rare or perhaps nonexistent, so it's hard to swallow this.

arvidn: > The converter change is necessary due to the fact that static_cast<long>(...) of a bitfield_flag doesn't work with deprecated-functions=off, and boost.python enum uses that cast internally, do I have that right?

that's my understanding, yes. This change also harmonizes these flags with how all other flags are treated

> Is there no better way to do this? It's unfortunate to cause a python API change over such an issue. The boost.python enum types has names and values dictionaries which are especially helpful in python, whereas their scope objects do not.

In an ideal world, the types would be preserved, with the risk of it being inconvenient on the python side. It would likely also mean all flag bit-wise operations would need to be bound.

These constants are really flags, binding them as enums is misleading. enumerating them as an enum isn't necessarily desirable. (they can still be enumerated by `dir()`).

> I understand the deprecated-functions=off build is important to check the path to eventual deprecation, but I expect real users of this build for python are rare or perhaps nonexistent, so it's hard to swallow this.

that's why this is going into master. API changes happen here, mostly slight ones, and I would consider these quite fringe. Nobody noticed that the converters were missing. I believe these flags could not be combined with bitwise operators.

AllSeeingEyeTolledEweSew: Hm, I think this comment didn't go through because it was on a "pending review".

To be clear, I dislike this change

arvidn: would you prefer to export flags as enums in general? (i.e. change the other flags to have a distinct type)
in your expression `lt.picker_flags_t.end_game | lt.picker_flags_t.rarest_first`, is the resulting type an `int` or a `picker_flags_t`? If it's the former, is that still compatible with assigning to a variable of type `picker_flags_t`?

AllSeeingEyeTolledEweSew: > would you prefer to export flags as enums in general? (i.e. change the other flags to have a distinct type)

Yes, IMO we should change them all.

This is an API change, so it must be in `master`. Converting a scope to `enum` adds a bunch of attributes (`.names`, `.values`, and all the methods of `int`), which derails client code currently using `dir()`. This is a great illustration of is why `dir()` shouldn't be the API of record.

I have been working on #6532. I would prefer to complete that before doing more `master` enhancements so it's easier to understand the backports.

> in your expression `lt.picker_flags_t.end_game | lt.picker_flags_t.rarest_first`, is the resulting type an `int` or a `picker_flags_t`? If it's the former, is that still compatible with assigning to a variable of type `picker_flags_t`?

The expression has type `int`. Assigning to a variable of type `picker_flags_t` needs a cast.

In practice, I haven't seen this come up too much. Types are usually implicit in python. Also any bound functions with strongly-typed int arguments will actually accept any `int` from the python side. So using or-expressions in a function argument works without a cast. All of this is reflected in the type stubs today.

It would be nice if boost.python's `enum` had strongly-typed operators. IIUC even pybind11's `enum` has similarly weakly-typed operators. The standard library's `enum.IntFlag` has strongly-typed operators, but the C++ binding libraries don't interface to it.

This situation at least matches the state of typing in python. There are lots of wrinkles. Most of the development effort today is put toward fixing common cases rather than ironing the edges.

IMO, `enum` is still a huge improvement over scopes due to `.names`, `.values`, and lack of `dir()`.

arvidn: I'm all for exposing the type-safety into python. The current approach was just chosen for simplicity (and before I realized python 3 has type-hints and that people actually use them these days).

I hesitate to expose flags as enums though, since they aren't meant to be enumerated, and they aren't mutually exclusive alternatives. Both of which are implied with enums.

I don't know what the best approach would be though. Since type checking only is done by mypy anyway, would it be reasonable to *just* expose types in the `.pyi` file, and still just use `int` everywhere? I would expect `int` to have a more efficient representation over a native type too, but that's speculation.

In an ideal world there would be actual runtime errors when trying to mix incompatible types too, but it seems like boost-python enum won't do that.

AllSeeingEyeTolledEweSew: > I hesitate to expose flags as enums though, since they aren't meant to be enumerated

I'm not sure what this means. It may be difficult to enumerate them in C++ land, but there are definitely good reasons to enumerate all possible values, as well as map flags to symbolic names at runtime. I cited some above.

> and they aren't mutually exclusive alternatives. Both of which are implied with enums.

This is a strict interpretation of the name "enum" that I have not heard before. As I mentioned, the python `enum` library provides strongly-typed integer flags. [The pybind11 enum also provides optional arithmetic operators for flags](https://pybind11.readthedocs.io/en/stable/classes.html#enumerations-and-internal-types). I've come to understand "enum" as "a related collection of constants".

On the contrary, the existence of "scopes" *is* confusing. Python generally does not have dummy classes that just act as namespaces for integer constants. Namespacing is done by putting things in separate modules. It would be nice to split up the gigantic libtorrent module for various reasons, but I think there would be a lot of interdependencies and I'm not sure how these work with boost.python.

> Since type checking only is done by mypy anyway, would it be reasonable to just expose types in the .pyi file, and still just use int everywhere?

Interesting idea. I think it would actually work better with flags bound as boost.python `enum`, rather than scopes. This is because an `enum` is really a subclass of `int` at runtime and the dummy scope isn't, so there are fewer gaps between type fiction and reality. I'll experiment with this.

> In an ideal world there would be actual runtime errors when trying to mix incompatible types too, but it seems like boost-python enum won't do that.

IMO this would be an impedence mismatch with python. The zeitgeist is still to use duck-typing at runtime.

Strongly-typed arithmetic would be interesting though (i.e. `a | b` is a strong type, whereas it's currently `int`). That might be possible with some modifications to boost.python's `enum`.

PR #7119, opened 2022-09-29
Fix bytes-in-flight accounting for uTP packets that were never sent

thrnz:
These potential fixes are related to each other and are a WIP.

First, `m_bytes_in_flight` shouldn't include packets marked as `need_resend`. Current behavior is to subtract it once the packet times out, though that doesn't happen with packets marked for resend straight away due to stalling. This change prevents such packets from being added to the total in the first place.

Second, I think `send_pkt()` and `resend_packet()` can still be called while the socket is marked as stalled. Some of their internal logic regarding re-sends expects `m_stalled` to be initially false and only set to true if that particular send fails. This changes it to instead use the result of the current send.

And finally, `m_out_packets` is currently increased regardless of whether any packets were actually sent or not. I'm not certain, but there might a case where this can prevent the `send_pkt()` loop for an incoming packet from deferring an ack when needed. To address it I've moved `++m_out_packets` to inside the stalled check. I'm not sure if the `utp_packets_out` stats counter should also be moved, or if that should remain counting packets that fail to send.

### DIFF
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -1318,6 +1318,20 @@ bool utp_socket_impl::send_pkt(int const flags)
 
 	bool const force = (flags & pkt_ack) || (flags & pkt_fin);
 
+	// If we're currently stalled then only proceeed if 'forced'
+	// If this is an ack, then defer it until writable()
+	// If this is a fin and it restalls, then it'll be
+	// stored in m_outbuf and resent once writable()
+	if (m_stalled)
+	{
+		if (!force) return false;
+		if (flags & pkt_ack)
+		{
+			defer_ack();
+			return false;
+		}
+	}
+
 //	TORRENT_ASSERT(state() != state_t::fin_sent || (flags & pkt_ack));
 
 	// first see if we need to resend any packets
@@ -1629,7 +1643,6 @@ bool utp_socket_impl::send_pkt(int const flags)
 		, reinterpret_cast<char const*>(h), p->size, ec
 		, p->mtu_probe ? udp_socket::dont_fragment : udp_send_flags_t{});
 
-	++m_out_packets;
 	m_sm.inc_stats_counter(counters::utp_packets_out);
 
 	if (ec == error::message_size)
@@ -1657,6 +1670,7 @@ bool utp_socket_impl::send_pkt(int const flags)
 			, reinterpret_cast<char const*>(h), p->size, ec, {});
 	}
 
+	bool stalled = false;
 	if (ec == error::would_block || ec == error::try_again)
 	{
 #if TORRENT_UTP_LOG
@@ -1667,6 +1681,7 @@ bool utp_socket_impl::send_pkt(int const flags)
 			m_stalled = true;
 			m_sm.subscribe_writable(this);
 		}
+		stalled = true;
 	}
 	else if (ec)
 	{
@@ -1677,8 +1692,9 @@ bool utp_socket_impl::send_pkt(int const flags)
 		return false;
 	}
 
-	if (!m_stalled)
+	if (!stalled)
 	{
+		++m_out_packets;
 		++p->num_transmissions;
 		// Only reset the timeout for the initial packet
 		if (m_bytes_in_flight == 0)
@@ -1719,7 +1735,7 @@ bool utp_socket_impl::send_pkt(int const flags)
 			|| m_seq_nr == 0);
 
 		// If we're stalled we'll need to resend
-		if (m_stalled)
+		if (stalled)
 			p->need_resend = true;
 
 		// If this packet is undersized then note the sequenece number so we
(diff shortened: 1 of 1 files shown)


arvidn:
> First, m_bytes_in_flight shouldn't include packets marked as need_resend. Current behavior is to subtract it once the packet times out, though that doesn't happen with packets marked for resend straight away due to stalling.

I think this has to be done carefully. If a packet cannot be sent because the socket isn't writable, if the packets-in-flight counter isn't incremented, is there *any* other back-pressure mechanism to prevent more packets being (attempted) sent, because there's still room in the congestion window?

The socket being stalled must cause all streams to stall too.

thrnz:
> > First, m_bytes_in_flight shouldn't include packets marked as need_resend. Current behavior is to subtract it once the packet times out, though that doesn't happen with packets marked for resend straight away due to stalling.
> 
> I think this has to be done carefully. If a packet cannot be sent because the socket isn't writable, if the packets-in-flight counter isn't incremented, is there _any_ other back-pressure mechanism to prevent more packets being (attempted) sent, because there's still room in the congestion window?
> 
> The socket being stalled must cause all streams to stall too.

I not sure there is any back-pressure apart from ending the currently running `send_pkt()` loop. It might still able to be called again though while stalled - ie. the second issue this tries to address.

We could stop any processing at all on `send_pkt()` and `resend_packet()` while `m_stalled` is `true` to address this - this would also have the benefit of solving the second issue entirely. Something like this near the start of `send_pkt()` maybe:
```
if (m_stalled)
{
	if (flags & pkt_ack) defer_ack();
	return false;
}
```
If the call to `send_pkt()` happens to be forcing an ack, then it can be deferred until the socket is writable again. A similar clause could be added to `resend_packet()`.

thrnz:
Just to clarify, the problem with `m_bytes_in_flight` at the moment is that there's no longer a point when then stalled packets bytes are removed from the total, causing it to keep growing every time a packet stalls.

https://github.com/arvidn/libtorrent/blob/92c86aab991ddf924a0d07a9187b4852ba2579fe/src/utp_stream.cpp#L3287-L3298

Previously, the bytes would be removed from `m_bytes_in_flight` here when the packet timed out and got marked for resending, and then re-added again when just before it got resent. However since the packet is now marked for resend straight after the stall this no longer happens, causing `m_bytes_in_flight` to keep growing for every stall, eventually getting to the stage where nothing else can be sent as it'll always be greater than the window size.

thrnz:
How about this:
 * Don't increment `m_bytes_in_flight` when a packet isn't sent due to stalling
 * Don't resend any packets while `m_stalled`. They'll get resent when `writable()` gets called.
 * Change `send_pkt()` to behave as if the congestion window is full while `m_stalled`, keeping the current behavior for sending 'forced' packets out.
 
If I've got this right then that should retain similar pressure to sending new packets as the current behavior, while fixing the `m_bytes_in_flight` accounting problem.

arvidn:
sorry for the delay. I've been quite busy lately, I'll try to do a thorough review soon.

thrnz:
> sorry for the delay. I've been quite busy lately, I'll try to do a thorough review soon.

No worries. I imagine this would need fixing before 2.0.8 and 1.2.18, but there's no hurry otherwise.

thrnz:
I wonder if I might be approaching this from the wrong angle.

I guess the key question is whether packets that fail to send due to 'stalling' should trigger loss or not. I was assuming that they probably shouldn't, as they're seemingly not 'lost', but perhaps they should.

Packet loss further along the network path happens due to various network buffers getting full and packets being dropped. In the case of 'stalling', the drops are happening right at the start due to the local network send buffer filling up. Perhaps this should be treated then the same as at any other point along the route and be considered as loss (as it previously was), otherwise the cwnd may continue to grow (assuming the stalls happen before the 100ms latency target kicks in) and the frequency of stalls may increase without the halving of the window to counter it.

The root cause of the stalls is a too shallow send buffer. It might be worth triggering `experienced_loss()` on stalls (when they happen, rather than waiting for the timeout), but also considering the network buffer sizes as well.

uTorrent appears to use 1MB network send/receive buffers by default (`bt.rcv_max_sockbuf`, `bt.send_max_sockbuf`), whereas libtorrent uses the OS default unless manually specified using `recv_socket_buffer_size`/`send_socket_buffer_size`. IIRC the default may be as low as 8K for UDP on Windows. This could explain some of the difference in performance between the two.

If my understanding is correct, every uTP stream shares the same UDP socket buffer as they all share the same socket, unlike TCP where each stream has its own. With a large enough number of active uTP peers and a high enough total speed, drops for both sending and receiving seem inevitable with too shallow buffers.

Perhaps it might be worth reconsidering the defaults. Increasing the network buffers would have the benefit of preventing incoming packets being dropped as well.

Have you got any thoughts on this @arvidn ?

arvidn:
My original stance has been "a packet loss is a packet loss", it doesn't matter what the underlying cause was.

However, I'm starting to think that maybe it *does* make sense to distinguish between causes, as you suggest. Filling the kernel send buffer could also be a sign of packets needing to be paced more evenly, in which case it shouldn't be considered a loss (probably).

I also agree that it most likely makes sense to set a higher default kernel send buffer for the one UDP socket.

arvidn:
@thrnz I wrote a test to cover the case of the UDP socket stalling, and discovered a few other cases that needed handling. So I ended up incorporating your patch.

I think it's a bit simpler to just have a check at the top of `send_pkt()` to exit early if the socket is stalled.

Would you mind taking a look? https://github.com/arvidn/libtorrent/pull/7145

PR #7122, opened 2022-09-29
introduce new API for create_torrent

arvidn:
to phase out the use of file_storage for creating torrents. Having a double-use for it complicates things

### DIFF
--- a/include/libtorrent/create_torrent.hpp
+++ b/include/libtorrent/create_torrent.hpp
@@ -29,6 +29,10 @@ see LICENSE file.
 #include <string>
 #include <utility>
 
+#if TORRENT_ABI_VERSION < 4
+#include <optional>
+#endif
+
 // OVERVIEW
 //
 // This section describes the functions and classes that are used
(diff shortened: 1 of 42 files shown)


include/libtorrent/create_torrent.hpp
```diff
@@ -73,9 +75,37 @@ see LICENSE file.
 //
 namespace libtorrent {
 
+TORRENT_VERSION_NAMESPACE_4
+
 	// hidden
 	using create_flags_t = flags::bitfield_flag<std::uint32_t, struct create_flags_tag>;
 
+	// This represents one file in the torrent file to be created
+	struct TORRENT_EXPORT create_file_entry
```
glassez, line 84: Always felt this name confusing since it isn't any kind of "callable".

arvidn: the underlying (msvc) feature `TORRENT_EXPORT` resolves to is `__declspec(dllexport)`, which means this type will be exported by the shared library. For types, in addition to exporting member functions, that means exporting the vtable and RTTI. (RTTI is used for `typeinfo` and `dynamic_cast`)

glassez: My comment refers to class names. And yes, I missed a bit (due to the inconvenience of navigating GitHub on a smartphone), so it had to be under the `create_torrent` name.

glassez: I would expect it to be something like `torrent_creator` since it isn't a "callable".

arvidn: ah, yes. That would be a better name probably

glassez: So can rename it (and others like it)? This could be done in the `master`, and in the current branches you could create aliases, like you have already done `file_layout` for `file_storage`.

PR #7132, opened 2022-10-06
Make sure boost-system is found when using old versions

lgbaldoni:
When using older boost versions, boost-system is a 'PUBLIC' link target.
Make sure it's treated as a required build dependency.
(courtesy of Christophe Giboudeaux <christophe@krop.fr>)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -808,7 +808,7 @@ endif()
 find_public_dependency(Boost REQUIRED)
 target_link_libraries(torrent-rasterbar PUBLIC Boost::headers)
 if (Boost_MAJOR_VERSION LESS_EQUAL 1 AND Boost_MINOR_VERSION LESS 69)
-	find_package(Boost REQUIRED COMPONENTS system)
+	find_public_dependency(Boost REQUIRED COMPONENTS system)
 	target_link_libraries(torrent-rasterbar PUBLIC Boost::system)
 endif()
 


arvidn:
what's the distinction between `find_target()` and `find_public_dependency()`?
I'm not a cmake expert, but it looks like the actual dependency is declared `PUBLIC` on the line below, so it's not obvious what difference this makes.

lgbaldoni:
> what's the distinction between find_target() and find_public_dependency()?
I'm not a cmake expert, but it looks like the actual dependency is declared PUBLIC on the line below, so it's not obvious what difference this makes.

The trouble with `find_package()` is that the dependency is added to `target_link_libraries()` but not searched for. So, for example, when one builds qBittorrent against libtorrent built against boost 1.66, the linking with boost_system fails because cmake doesn't know about it.

krop:
> what's the distinction between `find_target()` and `find_public_dependency()`? I'm not a cmake expert, but it looks like the actual dependency is declared `PUBLIC` on the line below, so it's not obvious what difference this makes.

You probably mean `find_package`. The purpose is different.

find_package finds dependencies needed to build your project.

find_public_dependency is a libtorrent function (https://github.com/arvidn/libtorrent/blob/RC_2_0/cmake/Modules/LibtorrentMacros.cmake#L42) used to add find_dependency entries for its public dependencies [LibtorrentRasterbarConfig.cmake.in](https://github.com/arvidn/libtorrent/blob/RC_2_0/LibtorrentRasterbarConfig.cmake.in)

and find_dependency tells projects using your code they have to find additional dependencies used for linking.

PR #7145, opened 2022-10-15
uTP stalled socket fixes

arvidn:
The bulk of this patch is based on https://github.com/arvidn/libtorrent/pull/7119

There are a few supporting patches in order to get test coverage, which uncovered a few other cases that needed fixing.

### DIFF
--- a/include/libtorrent/aux_/utp_stream.hpp
+++ b/include/libtorrent/aux_/utp_stream.hpp
@@ -798,6 +798,9 @@ private:
 	std::uint32_t m_adv_wnd = TORRENT_ETHERNET_MTU;
 
 	// the number of un-acked bytes we have sent
+	// This does not include packets that have been created but either failed to
+	// be sent or were lost. i.e. when a packet is lost, it's no longer
+	// considered in-flight.
 	std::int32_t m_bytes_in_flight = 0;
 
 	// the number of bytes read into the user provided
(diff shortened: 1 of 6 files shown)


arvidn:
This PR is suitable to review one commit at a time, to keep the scope a bit more focused

thrnz:
This looks pretty good to me.

Ending `send_pkt()` asap when stalled does simplify things, though it doesn't allow for sending an `ST_FIN` packet while stalled - `send_fin()` appears to progress to the fin-sent state regardless of a stall. Maybe it would be better for `send_fin()` to be more like `send_syn()` and handle sending the packet itself/adding it to `m_outbuf` as there's no need to have `send_pkt()` add any payload data.

There is some pre-existing fin wonkiness that causes the connection to time out rather than close cleanly anyway (it shows up in the test_utp simulations utp logiles as a single timeout right at the end), so it could probably be left as-is for now and addressed along with other fin related stuff. I should probably open a new PR with the fin tweaks I've got so far.

Nice work on the simulation too. I imaging running it against the 2.0.7 release, those 190 packet resends would have been previously treated as loss.

These fixes should probably also be backported to 1.2 along with #7113.

arvidn:
The test is tricky to back-port because of the libsimulator versions, but the fix would probably apply cleanly to 1.2

arvidn:
the tests should probably assert that there are 0 time-outs, once the FIN issue is fixed

arvidn:
back-ported here: https://github.com/arvidn/libtorrent/pull/7146

PR #7146, opened 2022-10-16
back-port bytes-in-flight uTP fix

arvidn:


arvidn:
@AllSeeingEyeTolledEweSew what's up with flake8 failing with:
```
AttributeError: 'EntryPoints' object has no attribute 'get'
```
[here](https://github.com/arvidn/libtorrent/actions/runs/3257849192/jobs/5349436336#step:4:130)?

PR #7153, opened 2022-10-16
Fix some issues finalizing uTP connections

thrnz:
This draft PR currently does the following:

* Prevents any more payload data being sent once we've either received or sent an `ST_FIN`. I've noticed in utp log files that it's currently possible for more payload data to be sent and for the sequence number to then exceed that of our sent `ST_FIN` packet. Ideally the other end would just discard such packets, by we probably shouldn't be sending them in the first place.

* Continue to ack incoming data packets after transitioning to the fin-sent state. The idea is to prevent the other side from potentially triggering timeouts/resends, especially if the fin packet itself needs resending.

* `m_seq_nr` is the next packet to be sent. Usually we don't allow `m_acked_seq_nr` to reach it as anything euqal to or above hasn't been sent yet. Once we've sent out the `ST_FIN` packet, `m_seq_nr` becomes the sequence number of the fin packet (it's not incremented when sending the fin), meaning we then need to allow `m_acked_seq_nr` to reach `m_seq_nr` in order to ack it. `maybe_inc_acked_seq_nr()` has been updated to allow this.

* Ensure that the fin packet is stored in `m_outbuf` so it can be resent if needed. Previously it was only stored if it happened to include payload (now it should never include payload). This is also needed for `maybe_inc_acked_seq_nr()` to correctly mark it as ack'd.

This appears to be enough to cleanly close connections that would previously have instead timed out. The results in the `test_utp.cpp` simulation no longer include unexpected timeouts.

* Handle calling `send_fin()` while the socket is stalled. It is re-called once the socket is writable again. If it re-stalls then it'll be in `m_sendbuf` and can be retried like any other packet. `send_fin()` progresses to the fin-sent state regardless, so other logic can still act as if it was sent.

### DIFF
--- a/include/libtorrent/aux_/utp_stream.hpp
+++ b/include/libtorrent/aux_/utp_stream.hpp
@@ -877,7 +877,9 @@ private:
 	std::uint16_t m_ack_nr = 0;
 
 	// the sequence number of the next packet
-	// we'll send
+	// we'll send. when we're closing the connection
+	// this becomes the sequence number of the ST_FIN packet
+	// and will no longer increase
 	std::uint16_t m_seq_nr = 0;
 
 	// this is the sequence number of the packet that
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -978,9 +978,12 @@ void utp_socket_impl::writable()
 	m_stalled = false;
 	if (should_delete()) return;
 
+	// this handles the case where send_fin() was called while stalled
+	if (state() == state_t::fin_sent && m_outbuf.at(m_seq_nr) == nullptr)
+		send_pkt(pkt_fin);
 	// if the socket stalled while sending an ack then there will be a
 	// pending deferred ack. make sure it gets sent out
-	if (!m_deferred_ack || send_pkt(pkt_ack))
+	else if (!m_deferred_ack || send_pkt(pkt_ack))
 		while(send_pkt());
 
 	maybe_trigger_send_callback();
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -3023,6 +3023,10 @@ bool utp_socket_impl::incoming_packet(span<char const> b
 			// After that has happened we know the remote side has all our
 			// data, and we can gracefully shut down.
 
+			// we should still ack any incoming data to prevent potential
+			// timeouts/resends at the other end
+			if (ph->get_type() == ST_DATA) defer_ack();
+
 			if (consume_incoming_data(ph, ptr, payload_size, receive_time))
 			{
 				break;
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -1409,7 +1409,10 @@ bool utp_socket_impl::send_pkt(int const flags)
 	// although, we may re-send packets, but those live in m_outbuf
 	TORRENT_ASSERT(state() != state_t::fin_sent || m_write_buffer_size == 0);
 
-	int payload_size = std::min(m_write_buffer_size
+	// If the connection is finalizing we no longer want to include any payload
+	bool const finalizing = state() == state_t::fin_sent || (flags & pkt_fin);
+
+	int payload_size = finalizing ? 0 : std::min(m_write_buffer_size
 		, effective_mtu - header_size);
 	TORRENT_ASSERT(payload_size >= 0);
 
@@ -1471,7 +1474,7 @@ bool utp_socket_impl::send_pkt(int const flags)
 	// payload size being zero means we're just sending
 	// an force. For efficiency, pick up the nagle packet
 	// if there's room
-	if (!m_nagle_packet || (payload_size == 0 && force
+	if (!m_nagle_packet || finalizing || (payload_size == 0 && force
 		&& m_bytes_in_flight + m_nagle_packet->size
 		> std::min(int(m_cwnd >> 16), int(m_adv_wnd))))
 	{
@@ -1768,6 +1771,13 @@ bool utp_socket_impl::send_pkt(int const flags)
 		TORRENT_ASSERT(payload_size >= 0);
 		if (!m_stalled) m_bytes_in_flight += new_in_flight;
 	}
+	else if (flags & pkt_fin)
+	{
+		TORRENT_ASSERT(payload_size == 0);
+		// If we're stalled we'll need to resend
+		if (m_stalled) p->need_resend = true;
+		m_outbuf.insert(m_seq_nr, std::move(p));
+	}
 	else
 	{
 		TORRENT_ASSERT(h->seq_nr == m_seq_nr);
@@ -1985,7 +1995,10 @@ void utp_socket_impl::maybe_inc_acked_seq_nr()
 	// supposed to be in m_outbuf
 	// if the slot in m_outbuf is 0, it means the
 	// packet has been ACKed and removed from the send buffer
-	while (((m_acked_seq_nr + 1) & ACK_MASK) != m_seq_nr
+	// once we're in the fin_sent state m_acked_seq_nr can equal
+	// m_seq_nr, but shouldn't reach m_seq_nr + 1
+	while (((m_acked_seq_nr + 1) & ACK_MASK) !=
+		(state() == state_t::fin_sent ? ((m_seq_nr + 1) & ACK_MASK) : m_seq_nr)
 		&& m_outbuf.at((m_acked_seq_nr + 1) & ACK_MASK) == nullptr)
 	{
 		// increment the fast resend sequence number
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,5 @@
+	* fix uTP streams timing out instead of closing cleanly
+
 2.0.8 released
 
 	* add write_torrent_file_buf() overload for generating .torrent files
--- a/simulation/test_utp.cpp
+++ b/simulation/test_utp.cpp
@@ -135,8 +135,7 @@ TORRENT_TEST(utp_pmtud)
 
 	TEST_EQUAL(metric(cnt, "utp.utp_packet_loss"), 0);
 
-	// TODO: 3 This timeout happens at shutdown. It's not very clean
-	TEST_EQUAL(metric(cnt, "utp.utp_timeout"), 1);
+	TEST_EQUAL(metric(cnt, "utp.utp_timeout"), 0);
 
 	TEST_EQUAL(metric(cnt, "utp.utp_packets_in"), 593);
 	TEST_EQUAL(metric(cnt, "utp.utp_payload_pkts_in"), 66);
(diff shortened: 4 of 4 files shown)


include/libtorrent/aux_/utp_stream.hpp
```diff
@@ -971,6 +973,10 @@ struct utp_socket_impl
 	// the socket being writable again
 	bool m_stalled:1;
 
+	// if the socket was 'stalled' when we wanted to send the ST_FIN packet
+	// then we retry once the socket is writable again
+	bool m_stalled_fin:1;
```
arvidn, line 978: it seems reasonable to keep this state, but I wonder if it's not possible to infer it from `state() == state_t::fin_sent`.

We the `fin_sent` state means we're not going to transmit any more (new) payload, but possibly resend lost packets. I suppose we can't necessarily know that if the socket is un-stalled, that we need to resend the FIN packet just because we're in the fin_sent state.

src/utp_stream.cpp
```diff
@@ -1360,6 +1375,9 @@ bool utp_socket_impl::send_pkt(int const flags)
 			m_fast_resend_seq_nr = (m_fast_resend_seq_nr + 1) & ACK_MASK;
 	}
 
+	if (state() == state_t::fin_sent && !force)
+		return false;
```
arvidn, line 1379: is this the case where we attempt to send more, new, payload *after* having closed the connection?

src/utp_stream.cpp
```diff
@@ -1394,7 +1412,10 @@ bool utp_socket_impl::send_pkt(int const flags)
 		+ (sack ? sack + 2 : 0)
 		+ (close_reason ? 6 : 0);
 
-	int payload_size = std::min(m_write_buffer_size
+	// If the connection is finalizing we no longer want to include any payload
+	bool const finalizing = state() == state_t::fin_sent || (flags & pkt_fin) || m_eof;
```
arvidn, line 1416: I don't think it's right to add `|| m_eof` at the end here. Just because the remote end is done sending payload doesn't mean we have sent all our payload yet.

src/utp_stream.cpp
```diff
@@ -2992,6 +3024,10 @@ bool utp_socket_impl::incoming_packet(span<char const> b
 			// After that has happened we know the remote side has all our
 			// data, and we can gracefully shut down.
 
+			// we should still ack any incoming data to prevent potential
+			// timeouts/resends at the other end
+			if (ph->get_type() == ST_DATA) defer_ack();
```
arvidn, line 3029: I imagine this is the key change for fixing the timeouts in the tests, right? since those tests never cause the socket to stall

thrnz: Deferring acks here is so we can continue to ack data packets when we're the first end to send the fin packet.

I think the key change for fixing the timeouts in the tests is in `maybe_inc_acked_seq_nr()`. From the test's utp log, prior to this PR we had this:
```
[000004732088] 0x562466e0ca60: incoming packet seq_nr:39341 ack_nr:54427 type:ST_STATE id:14063 size:0 timestampdiff:32623 timestamp:5199507 our ack_nr:39341 our seq_nr:54427 our acked_seq_nr:54426 our state:FIN_SENT
```
Note that the incoming packet is trying to ack the fin packet (54427), however our acked_seq_nr hasn't been increased.

`maybe_inc_acked_seq_nr()` now allows `acked_seq_nr` to reach `seq_nr` once we're in the fin-sent state. After this PR, for the same packet we now get the following. Note that our acked_seq_nr has now increased to where it should be.
```
[000004732088] 0x558860f04c20: incoming packet seq_nr:39341 ack_nr:54427 type:ST_STATE id:14063 size:0 timestampdiff:32623 timestamp:5199507 our ack_nr:39341 our seq_nr:54427 our acked_seq_nr:54427 our state:FIN_SENT
[000004732088] 0x558860f04c20: FIN acked

```
For `maybe_inc_acked_seq_nr()`, I think we also need to have the fin packet stored in `m_outbuf` to stop it being prematurely ack'd, which we also now do. The tests still timeout without both of these changes.

src/utp_stream.cpp
thrnz: I think there might have been a particular case where this actually happened, though I may have also just fluffed this bit up. I originally did this a few weeks ago, and vaguely recall solving something with this, but can't remember exactly. I'll see if I can work out how (if!) I got this to happen. I imagine I might have triggered it while testing with 'real-world' torrents, rather than the simulations.

src/utp_stream.cpp
thrnz: I agree. I'll remove it.

include/libtorrent/aux_/utp_stream.hpp
thrnz: I figure that if we reach `writable()` while in the fin-sent state, we can't really tell if we've actually sent the fin packet or not - we could be there after a stall from a resend or ack. I figured explicit noting whether we need to call send_fin() again would be the simplest way.

Although, maybe we could check `m_sendbuf` for the sent-but-unack'd fin packet instead?

thrnz: `(state() == state_t::fin_sent && m_outbuf.at(m_seq_nr) == nullptr)`

Would this be able to pick it up? Once we're in the fin_sent state, `m_seq_nr` is the fin packet itself. I think the only time it won't be present in `m_outbuf` while in the fin_sent state would be if the `send_pkt(pkt_fin)` within `send_fin()` happened while stalled. Once it gets ack'd and removed from `m_outbuf`, the state progresses beyond fin_sent.

src/utp_stream.cpp
thrnz:
```
        if (state() == state_t::fin_sent && !force)
	{
		UTP_LOGV("%8p: We can reach here\n"
			, static_cast<void*>(this));
		return false;
	}
```

It looks like it is possible for this to happen. It seems to come via `issue_write()`

```
[000325524325] 0x7f806a583d10: skipping send (no payload and no force) seq_nr:56054 ack_nr:47142 id:39595 target:xxx header_size:20 error:No error information send_buffer_size:0 cwnd:3634 adv_wnd:1048576 in-flight:238 mtu:1241
[000325524334] 0x7f806a583d10: defer ack
[000325524334] 0x7f806a583d10: subscribe drained
[000325524335] 0x7f806a583d10: incoming stream consumed
[000325524338] 0x7f806a583d10: sending packet seq_nr:56054 ack_nr:47142 type:ST_FIN id:39595 target:xxx size:20 error:No error information send_buffer_size:0 cwnd:3634 adv_wnd:1048576 in-flight:238 mtu:1241 timestamp:1462880446 time_diff:185335 mtu_probe:0 extension:0
[000325524373] 0x7f806a583d10: Cancelling redundant deferred ack
[000325524374] 0x7f806a583d10: state:FIN_SENT
[000325524572] 0x7f806a583d10: calling read handler read:41843
[000325524574] 0x7f806a583d10: calling read handler read:41843 ec:No error information shutdown:0
[000325524643] 0x7f806a583d10: add_read_buffer 327502 bytes
[000325525006] 0x7f806a583d10: Didn't fill entire target: 0 bytes left in buffer
[000325525008] 0x7f806a583d10: 240 packets moved from buffer to user space (327502 bytes)
[000325525830] 0x7f806a583d10: add_read_buffer 377133 bytes
[000325525832] 0x7f806a583d10: new read handler. 0 bytes in buffer
[000325525837] 0x7f806a583d10: add_write_buffer 128 bytes
[000325525838] 0x7f806a583d10: add_write_buffer 128 bytes
[000325525839] 0x7f806a583d10: add_write_buffer 128 bytes
[000325525840] 0x7f806a583d10: add_write_buffer 128 bytes
[000325525840] 0x7f806a583d10: add_write_buffer 128 bytes
[000325525841] 0x7f806a583d10: add_write_buffer 128 bytes
[000325525842] 0x7f806a583d10: add_write_buffer 14 bytes
[000325525843] 0x7f806a583d10: new write handler. 782 bytes to write
[000325525844] 0x7f806a583d10: We can reach here
```

arvidn: interesting. trying to write to a socket that we have closed should probably fail earlier, long before we add the data to the send buffer even. It also indicates that something fishy is going on a layer up. There shouldn't be any new data written to sockets after they're closed.

arvidn:
I imagine most (if not all) of this makes sense. I would like to better understand the various issues in more detail. I think it might make sense to split it up into smaller pieces to help in this understanding.

I can think of at least 3 somewhat independent issues:

1. `async_write()` on a socket that has been closed should probably fail immediately. And maybe it should be an assert at the `utp_stream` level, that we don't add more write buffers when in `fin_sent` mode.
2. The fix to make sure we honor an incoming FIN instead of timing out when closing sockets.
3. The ability to resend a FIN packet if we close a socket while the underlying socket is stalled

I'll start looking at these as well

thrnz:
I'll split what I've currently got into separate better-described commits to make it clearer what each change is addressing.

src/utp_stream.cpp
thrnz: Triggering more data being sent while the connection is closing seems to be a bit of a special case. Picking up existing write buffer or nagled packets when sending acks or the fin itself should already be prevented (`finalizing == true`), although in this case its when more data is being added to the write buffer. This seems to only sometimes happen.

My current understanding is that the other end needs to be the first to send the fin for this to occur. Maybe in this case the BT layer isn't aware that the connection is closing until it's actually deleted. Is `cancel_handlers()` when it finds out? Could this maybe be called earlier when the other end is the first to send a fin?

arvidn:
excellent, thanks! The patch to prevent sending data after the socket has been closed ([here](https://github.com/arvidn/libtorrent/pull/7153/commits/7e928115478a76cca50e008e440ce30f97df4343#diff-a9b0002dfebb81b315f1d19d346361c385368345d731117eefad87b4b9817ca7R1366)). I would prefer to have that check in `async_write_some()` and `write_some()` instead, and have your check be an assert.

Because of the type erasure between the `utp_socket` front-end and the `utp_stream` back-end, the `m_state` field isn't directly available. I'll propose a PR.

src/utp_stream.cpp
arvidn: consider the case where we have a bunch of packets in-flight. Before those packets are ACKed, we close the socket and send a FIN.

Some of those outstanding packets might need to be re-sent, which would take this path, even when we're in `fin_sent` mode. I think this invariant really should be to not send any packets with sequence numbers past `m_seq_nr`

arvidn: oh, I see. This is *after* the loop that's re-sending packets.

arvidn:
This is what I had in mind: https://github.com/arvidn/libtorrent/pull/7162

arvidn (approved):

thrnz:
It might be worth backporting this alongside #7162 to 1.2.

arvidn:
back-porting the main commit from this PR is non-trivial. i.e. fixing uTP streams to close gracefully rather than timing out.

It causes this assert in `RC_1_2`:

```
file: '../src/utp_stream.cpp'
line: 3775
function: tick
expression: m_bytes_in_flight == 0
```

arvidn:
https://github.com/arvidn/libtorrent/pull/7166

PR #7161, opened 2022-10-19
Fix for C bindings

ntfshard:
`torrent_status::error` is a 1024 bytes long. Still if `ts.error.size()` will be > 1024 then `\0` will not be added to the end of `error` field. Perhaps it should be added explicitly in any case as the last symbol of array (here and in other places too).

Also we should call [va_end](https://en.cppreference.com/w/c/variadic/va_end) for freeing resources.

Target branch is a RC_2_0 is correct?

### DIFF
--- a/bindings/c/library.cpp
+++ b/bindings/c/library.cpp
@@ -151,6 +151,7 @@ TORRENT_EXPORT void* session_create(int tag, ...)
 
 		tag = va_arg(lp, int);
 	}
+	va_end(lp);
 
 	if (listen_range.first != -1 && (listen_range.second == -1
 		|| listen_range.second < listen_range.first))
@@ -253,6 +254,7 @@ TORRENT_EXPORT int session_add_torrent(void* ses, int tag, ...)
 
 		tag = va_arg(lp, int);
 	}
+	va_end(lp);
 
 	if (!params.ti && torrent_data && torrent_size)
 		params.ti.reset(new (std::nothrow) torrent_info(torrent_data, torrent_size));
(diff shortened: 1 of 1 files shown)


arvidn (approved):

bindings/c/library.cpp
```diff
@@ -490,7 +493,7 @@ TORRENT_EXPORT int torrent_get_status(int tor, torrent_status* s, int struct_siz
 	s->state = (state_t)ts.state;
 	s->paused = ts.paused;
 	s->progress = ts.progress;
-	strncpy(s->error, ts.error.c_str(), 1025);
+	strncpy(s->error, ts.error.c_str(), 1024);
```
arvidn, line 496:
```suggestion
	strncpy(s->error, ts.error.c_str(), sizeof(s->error)-1);
	s->error[sizeof(s->error)-1] = '\0';
```

arvidn: I wonder if the original intention was to say 1023 instead of 1025, and assume the array was initialized to zeroes. Either way, I think you're right that it needs to be explicitly null terminated.

bindings/c/library.cpp
```diff
@@ -490,7 +493,7 @@ TORRENT_EXPORT int torrent_get_status(int tor, torrent_status* s, int struct_siz
...
 	s->next_announce = lt::total_seconds(ts.next_announce);
 	s->announce_interval = lt::total_seconds(ts.announce_interval);
 	strncpy(s->current_tracker, ts.current_tracker.c_str(), 512);
```
ntfshard, line 499: Also applying such change?)

bindings/c/library.cpp
ntfshard: Dangerous assumption)
Perhaps we could try to re-analyze some other places in code to fix potential memory overrun and integrate it in this change? Any suspects?

arvidn (approved):

PR #7162, opened 2022-10-20
prevent sending data on a uTP socket after closing it

arvidn:


thrnz:
This seems to do the job. I've not yet been able to trigger it happening again using this PR.

src/utp_stream.cpp
```diff
@@ -1394,6 +1405,10 @@ bool utp_socket_impl::send_pkt(int const flags)
 		+ (sack ? sack + 2 : 0)
 		+ (close_reason ? 6 : 0);
 
+	// once we're in fin-sent mode, the write buffer should not be re-filled
+	// although, we may re-send packets, but those live in m_outbuf
+	TORRENT_ASSERT(state() != state_t::fin_sent || m_write_buffer_size == 0);
```
arvidn, line 1410: this assert fails. The fundamental issue is that the utp implementation doesn't handle one-sided closing of the connection.

When receiving a FIN, we correctly note the sequence number. We reject more data past that sequence number but allow re-sends of previous packets. However, currently we also respond with a FIN immediately (without the consent of the upper layer). It's not great.

Ideally, an incoming FIN should be acked immediately, and only close the incoming stream, and produce an EOF when the upper layer reads up against it. The outgoing FIN should only be sent when the upper layer closes the socket. Which means the outgoing sequence number needs to be recorded when closing the socket, and then assert we don't try to send any data past that.

PR #7165, opened 2022-10-22
Minor fixes

ntfshard:
I guess I need some clarification for some other lines in code. Seems they should be changed, they looks suspicious a little bit

fuzzers/src/add_torrent.cpp:142, 146, 150: the same code in an 'if' and related 'else' branch
```
if (bits.read(1)) mask.set_bit(i); else mask.set_bit(i);
if (bits.read(1)) ret.have_pieces.set_bit(i); else ret.have_pieces.set_bit(i);
if (bits.read(1)) ret.verified_pieces.set_bit(i); else ret.verified_pieces.set_bit(i);
```

test/test_remap_files.cpp:112 the same expression on both sides of an operator &&
```
TEST_CHECK(st.state != torrent_status::checking_files
	&& st.state != torrent_status::checking_files);
```

### DIFF
--- a/examples/connection_tester.cpp
+++ b/examples/connection_tester.cpp
@@ -484,7 +484,7 @@ struct peer_conn
 		unsigned int length = read_uint32(ptr);
 		if (length > sizeof(buffer))
 		{
-			std::fprintf(stderr, "len: %d\n", length);
+			std::fprintf(stderr, "len: %u\n", length);
 			close("ERROR RECEIVE MESSAGE PREFIX: packet too big", error_code());
 			return;
 		}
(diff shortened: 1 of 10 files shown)


examples/connection_tester.cpp
```diff
@@ -484,7 +484,7 @@ struct peer_conn
 		unsigned int length = read_uint32(ptr);
 		if (length > sizeof(buffer))
 		{
-			std::fprintf(stderr, "len: %d\n", length);
+			std::fprintf(stderr, "len: %u\n", length);
```
arvidn, line 487: I'm surprised GCC didn't warn about this

arvidn:
thanks for the typos. I fixed them here: https://github.com/arvidn/libtorrent/pull/7167

arvidn (approved):

examples/connection_tester.cpp
ntfshard: Perhaps warning is disabled or GCC version is too low. Streams doesn't have such potential problem

PR #7166, opened 2022-10-22
Backport utp fixes

arvidn:


thrnz:
It looks like the socket state comparisons are being done against `ST_FIN` (the packet type) instead of `UTP_STATE_FIN_SENT` (the socket state).

arvidn:
oops

PR #7198, opened 2022-11-30
moved file_entry out of file_storage.hpp to aux

aldenml:


arvidn:
I'm a bit torn about this. the file_entry is really tightly coupled with the implementation details of `file_storage`

aldenml:
that's true, let me close this and revisit it in the future

PR #7199, opened 2022-12-01
some typos and minor refactor

aldenml:


aldenml:
Hi @arvidn, if I'm reading correctly, the errors in the CI are not due to these changes. Is that right?

include/libtorrent/torrent_handle.hpp
```diff
@@ -293,8 +293,8 @@ namespace aux {
 		// omitted then it is implicitly set to the end of the file the first
 		// piece is set to. If the torrent metadata has not been downloaded yet
 		// then the functions do nothing.
-		void set_sequential_range(piece_index_t const first_piece, piece_index_t const last_piece) const;
-		void set_sequential_range(piece_index_t const first_piece) const;
+		void set_sequential_range(piece_index_t first_piece, piece_index_t last_piece) const;
+		void set_sequential_range(piece_index_t first_piece) const;
```
sledgehammer999, line 297: Won't this break the API/ABI?

aldenml: the compiler ignores this particular use of `const` (to an `int`) in the declaration and the lint suggests removing it. It's still present in the definition.

arvidn (approved):

include/libtorrent/torrent_handle.hpp
arvidn: right. top level `const` is redundant in function declarations.

PR #7211, opened 2022-12-19
optimize resume data format to use less space

arvidn:


docs/manual.rst
```diff
@@ -479,11 +479,20 @@ The file format is a bencoded dictionary containing the following fields:
...
+|                          | can still be read.                                           |
+|                          | A string with piece flags, one character per piece.          |
 |                          | Bit 1 means we have that piece.                              |
 |                          | Bit 2 means we have verified that this piece is correct.     |
 |                          | This only applies when the torrent is in seed_mode.          |
 +--------------------------+--------------------------------------------------------------+
+| ``pieces2``              | A string where each bit indicates whether we have the piece  |
+|                          | or not.                                                      |
++--------------------------+--------------------------------------------------------------+
+| ``verified``             | A string where each bit indicates whether the piece has been |
```
glassez, line 492: Why not re-use `pieces2`. IIRC, it isn't used in `seed-mode`.

docs/manual.rst
```diff
@@ -479,11 +479,20 @@ The file format is a bencoded dictionary containing the following fields:
...
 +--------------------------+--------------------------------------------------------------+
-| ``pieces``               | A string with piece flags, one character per piece.          |
+| ``pieces``               | **deprecated**. This field is no longer written, but can     |
+|                          | can still be read.                                           |
+|                          | A string with piece flags, one character per piece.          |
 |                          | Bit 1 means we have that piece.                              |
 |                          | Bit 2 means we have verified that this piece is correct.     |
 |                          | This only applies when the torrent is in seed_mode.          |
 +--------------------------+--------------------------------------------------------------+
+| ``pieces2``              | A string where each bit indicates whether we have the piece  |
```
glassez, line 489: Wouldn't it be better to add format version and re-use existing names? Maybe you can even use libtorrent version for such a purpose.

arvidn: that's a good point. I already have a version field. I should use that.
I intend to back-port support for loading the new format, so libtorrent version wouldn't work.

docs/manual.rst
arvidn: I think it would be risky to make the semantics of this field depend on a different flag (i.e. seed mode)

glassez: >I think it would be risky to make the semantics of this field depend on a different flag (i.e. seed mode)

But in fact it is. Unless I'm wrong and there is existing case when both pieces and verified are supposed to be used simultaneously.
However, as you wish.

arvidn: yes, I don't think `pieces` and `verified` are used at the same time. At least it wouldn't make sense. however, there is some fault tolerance in giving them different names in case something goes wrong.

src/write_resume_data.cpp
```diff
@@ -141,24 +141,16 @@ namespace {
 				{
 					auto const& verified = atp.verified_leaf_hashes[f];
 					if (!verified.empty())
-					{
-						auto& ret_verified = ret_dict["verified"].string();
-						ret_verified.reserve(verified.size());
-						for (auto const bit : verified)
-							ret_verified.push_back(bit ? '1' : '0');
-					}
+						ret_dict["verified2"] = std::string(verified.data()
```
glassez, line 144:
```suggestion
						ret_dict["verified"] = std::string(verified.data()
```

src/write_resume_data.cpp
```diff
@@ -141,24 +141,16 @@ namespace {
...
+							, std::size_t(verified.size() + 7) / 8);
```
glassez, line 145: Wouldn't it be better to provide data size via `bitfield` interface?

docs/manual.rst
```diff
@@ -479,11 +482,26 @@ The file format is a bencoded dictionary containing the following fields:
...
+|                          |                                                              |
+|                          | In version 1 of this file format, this was a string with one |
+|                          | byte for every piece. Bit 0 meant we had the piece, bit 1    |
+|                          | meant the piece was verified (for seed-mode lazy checking)   |
++--------------------------+--------------------------------------------------------------+
+| ``verified``             | A string where each bit indicates whether the piece has been |
+|                          | verified to be correct. This is used in seed-mode, where     |
+|                          | pieces are checked lazily.                                   |
+|                          |                                                              |
+|                          | This field did not exist in version 1 of this file format.   |
```
glassez, line 503: Is it correct? Or is there another "verified" field in version 1?

arvidn: it is correct. in version 1, the `pieces` field used a full byte per piece. one of those bits meant "we have the piece" and another bit meant "the piece has been verified (in seed-mode". The remaining 6 bits were wasted space.

arvidn: but there's another `verified` field under the v2 merkle trees too, but that's separate

src/write_resume_data.cpp
arvidn: yeah, that would probably be better. there's already a member for the number of words

src/write_resume_data.cpp
arvidn: this typo exposed a hole in the tests. fixing the tests exposed an issue in this patch where the bitfield sizes are now rounded up to be divisible by 8, which causes failures in the round-trip tests.

Issue #6112, opened 2021-04-05
hashing using Intel ISA-L


fbdtemme:
I want to suggest implementing a new hashing backend using the [Intel ISA-L crypto library](https://github.com/intel/isa-l_crypto).

ISA-L uses multi-buffer hashing, which takes advantage of the parallelism of the execution pipeline to compute several hashes at once within a single core using SIMD instructions. It supports ARM NEON as well as all Intel SIMD instruction sets.

I assume that significant refactoring would be needed to support this library, but the increase in throughput is quite impressive.

I am the author of a metafile utility [torrenttools](https://github.com/fbdtemme/torrenttools) which has implemented supported for this library. A small benchmark shows the AVX-2 SHA1 implementation in ISA-l can increase the throughput by a factor of 2.5 compared to the OpenSSL SHA1 implementation.

userdocs:
How does it compare with wolfssl that can be built against libtorrent v2 (out of curiosity).

> wolfssl - links against wolfssl to use it for SHA-1 hashing and HTTPS tracker support.

`crypto=wolfssl`

https://libtorrent.org/building.html

fbdtemme:
I have not tested wolfSSL. 
According to the benchmarks available on their website, the AVX2 implementation of SHA256 is about equal to that of OpenSSL, so the speedup by using Intel ISA-L should be the same as well.

elgatito:
@fbdtemme can you shared a source code for your benchmark?

Do they target latest CPUs? What if you have to target systems with pretty old CPUs, will it still be that fast on both old and new CPUs?

fbdtemme:
The benchmark is quite basic and is done be simply comparing the runtime of various torrent creation tools.
You can find the notebook file used to generate the timings in the benchmark folder of my torrent utility tool [torrenttools](https://github.com/fbdtemme/torrenttools). Files are read from an in-memory filesystem to make sure the storage device is not the bottleneck.  

My cpu is an Intel i7-7700HQ so the AVX-2 implementation is used. ISA-L has various backends that are automatically selected at runtime according to the best available instruction set.

Intel has some benchmarks with different configurations here: https://01.org/sites/default/files/documentation/intel_isa-l_2.19_performance_report_0.pdf
The numbers to look for are "Multibuffer SHA-1" and "Multibuffer SHA-256"

[...]

PR #7214, opened 2022-12-21
add option to build against library isa-l_crypto (intel) no SSL support

joriscarrier:
see #6112

### DIFF
--- a/Jamfile
+++ b/Jamfile
@@ -99,6 +99,10 @@ rule linking ( properties * )
 	{
 		result += <library>wolfssl ;
 	}
+	else if <crypto>isal in $(properties)
+	{
+		result += <library>isal_crypto ;
+	}
 
 	if <target-os>windows in $(properties)
 		|| <target-os>cygwin in $(properties)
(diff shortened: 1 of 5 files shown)


arvidn:
this seems fine, but my understanding is that you won't see any speed-ups this way. iiuc, ISA-L gains a speed advantage when you can compute many hashes in parallel. There are a few places where that's somewhat easy to do (like in v2 merkle trees) but to do it for piece hashes would be quite complicated.

arvidn:
have you measured any improvements?

arvidn:
the spell checker caught one typo. some new words need to be added to the dictionary and a default for the preprocessor macro, to define it as 0 would need to be added

joriscarrier:
> have you measured any improvements?

No, I didn't measure any improvement, it would be the opposite

arvidn:
> > have you measured any improvements?
> 
> No, I didn't measure any improvement, it would be the opposite

what does "it" refer to?
Do you mean that using ISA-L causes a performance regression? If you haven't measured, we wouldn't know I'd think.

PR #7221, opened 2022-12-28
fix Mac python binding build on CI

arvidn:


Samfun75:
The error is coming when `brew update` finds new versions of python 3.10 & 3.11 and after installing the new versions the symlinks fail to link. The possible solution could be:
- Remove `brew update` or
- Pin `python@3.10` & `python@3.11` so they don't update or
- Pin only python@3.10
`brew pin python@3.10` and
remove the python@3.11 symlinks so that when 3.11 is update the symlinks don't conflict using either `rm -rf /usr/local/bin/2to3-3.11 /usr/local/bin/idle3.11 /usr/local/bin/pydoc3.11 /usr/local/bin/python3.11 /usr/local/bin/python3.11-config` or `brew unlink python@3.11` [Test Run](https://github.com/Samfun75/libtorrent/actions/runs/3792822389/jobs/6449359616)

arvidn:
@Samfun75 thanks!

PR #7232, opened 2023-01-02
Fix utp outgoing FIN

arvidn:
This is a series of patches to uTP cleaning up the handling of closing the streams. It addresses this comment: https://github.com/arvidn/libtorrent/pull/7162#discussion_r1045233041

The correct behavior when the other end sends a `FIN` packet is to record that the stream has been closed and at which sequence number it was closed. Then deliver any undelivered payload to the upper layer and finally deliver the `eof` error. The uTP layer should not respond with a `FIN` to close down the outgoing direction until the upper layer says to do so.

In the case of bittorrent, this will cause the upper layer to close its end of the connection as well.

These patches are separated to be somewhat independent. The last commit is the main one that makes the uTP layer not respond with a `FIN`, but instead let the upper layer close the socket before doing so.

With these patches, the socket is kept alive until un-acked data has been acked, and is kept open for one time-out period in order to ACK the other end's `FIN` packet as well, even after the socket has been closed by the upper layer.

The rules that have been made stricter are:

* we now correctly check to ensure the upper layer doesn't send more data after closing it
* the check for incoming payload at sequence numbers past the EOF seq nr was incorrect, and has been corrected
* FIN packets are not supposed to contain payload (this can be deduced from libutp, but it's not well documented). With this patch, FIN packets with payload is supported, but logs a warning.
* When closing a socket that has a pending nagle packet, that packet will be force-sent before the FIN packet. This ensures sequence numbers remain correct and that the FIN packet does not carry payload.

### DIFF
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -168,7 +168,7 @@ utp_socket_impl::utp_socket_impl(std::uint16_t const recv_id
 	, m_recv_id(recv_id)
 	, m_delay_sample_idx(0)
 	, m_state(static_cast<std::uint8_t>(state_t::none))
-	, m_eof(false)
+	, m_in_eof(false)
 	, m_attached(true)
 	, m_nagle(true)
 	, m_slow_start(true)
@@ -2179,7 +2179,7 @@ bool utp_socket_impl::consume_incoming_data(
 
 	if (ph->get_type() != ST_DATA) return false;
 
-	if (m_eof && m_ack_nr == m_eof_seq_nr)
+	if (m_in_eof && m_ack_nr == m_in_eof_seq_nr)
 	{
 		// What?! We've already received a FIN and everything up
 		// to it has been acked. Ignore this packet
@@ -2485,12 +2485,12 @@ bool utp_socket_impl::incoming_packet(span<char const> b
 	// if the socket is closing, always ignore any packet
 	// with a higher sequence number than the FIN sequence number
 	// ST_STATE messages always include the next seqnr.
-	if (m_eof && (compare_less_wrap(m_eof_seq_nr, ph->seq_nr, ACK_MASK)
-		|| (m_eof_seq_nr == ph->seq_nr && ph->get_type() != ST_STATE)))
+	if (m_in_eof && (compare_less_wrap(m_in_eof_seq_nr, ph->seq_nr, ACK_MASK)
+		|| (m_in_eof_seq_nr == ph->seq_nr && ph->get_type() != ST_STATE)))
 	{
 #if TORRENT_UTP_LOG
 		UTP_LOG("%8p: ERROR: incoming packet type: %s seq_nr:%d eof_seq_nr:%d (ignored)\n"
-			, static_cast<void*>(this), packet_type_names[ph->get_type()], int(ph->seq_nr), m_eof_seq_nr);
+			, static_cast<void*>(this), packet_type_names[ph->get_type()], int(ph->seq_nr), m_in_eof_seq_nr);
 #endif
 		return true;
 	}
@@ -2724,15 +2724,13 @@ bool utp_socket_impl::incoming_packet(span<char const> b
 			}
 		}
 
-		if (m_eof)
+		if (m_in_eof)
 		{
 			UTP_LOGV("%8p: duplicate FIN packet (ignoring)\n", static_cast<void*>(this));
 			return true;
 		}
-		m_eof = true;
-		m_eof_seq_nr = ph->seq_nr;
-
-		// we will respond with a fin once we have received everything up to m_eof_seq_nr
+		m_in_eof = true;
+		m_in_eof_seq_nr = ph->seq_nr;
 	}
 
 	switch (state())
@@ -2874,7 +2872,7 @@ bool utp_socket_impl::incoming_packet(span<char const> b
 
 			// Everything up to the FIN has been received, respond with a FIN
 			// from our side.
-			if (m_eof && m_ack_nr == ((m_eof_seq_nr - 1) & ACK_MASK))
+			if (m_in_eof && m_ack_nr == ((m_in_eof_seq_nr - 1) & ACK_MASK))
 			{
 				UTP_LOGV("%8p: incoming stream consumed\n", static_cast<void*>(this));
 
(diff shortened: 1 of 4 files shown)


arvidn:
@thrnz ping if you're interested in reviewing

glassez:
@arvidn 
Could it be related to fixing crash reported in https://github.com/qbittorrent/qBittorrent/issues/18157#issuecomment-1347030504?

arvidn:
While fixing this, I don't think I saw any misbehavior that could cause a crash or use-after-free. Just issues at the uTP protocol level.

thrnz:
I'm not too familiar with how the upper layer interacts with the uTP layer - the recent uTP performance improvements were mostly at the uTP level itself - but the reasoning behind these changes seems to make sense.

I can't spot anything obviously wrong. I might try throwing some live torrents at this later just to make sure there's no weirdness going on.

Seeker2:
I feel compelled to mention a scenario I've been painfully experiencing for a long time that may tie into this:
https://github.com/arvidn/libtorrent/issues/3542#issuecomment-1257038658

At the least, it should be one of the tests done against this.

markmdscott:
> I feel compelled to mention a scenario I've been painfully experiencing for a long time that may tie into this: [#3542 (comment)](https://github.com/arvidn/libtorrent/issues/3542#issuecomment-1257038658)
> 
> At the least, it should be one of the tests done against this.

#3542 was reportedly fixed by @thrnz 's recent set of uTP patches a little while ago.

thrnz:
> I feel compelled to mention a scenario I've been painfully experiencing for a long time that may tie into this: [#3542 (comment)](https://github.com/arvidn/libtorrent/issues/3542#issuecomment-1257038658)
> 
> At the least, it should be one of the tests done against this.

Prior to 2.0.8, uTP streams weren't closing cleanly and were instead timing out. That might help explain the issue of lingering libtorrent based uTP peers, and should now hopefully be fixed, though doesn't explain those other issues.

I know of one more issue related to #3542 that probably still needs addressing - it helps explain incoming packet loss on Windows 7 in particular - though would be surprised if it resolved everything.

Seeker2:
> Prior to 2.0.8, uTP streams weren't closing cleanly and were instead timing out. That might help explain the issue of lingering libtorrent based uTP peers, and should now hopefully be fixed, though doesn't explain those other issues.

The long-lived uTP seed-to-seed connections has another issue with the unknown (to me!) traffic about 5 seconds after the connection is made. TCP seed-to-seed connections typically disconnect in under a second. Cleaning up the closing may help prevent a 30+ second delay but not a 4+ second one.
 
> I know of one more issue related to #3542 that probably still needs addressing - it helps explain incoming packet loss on Windows 7 in particular - though would be surprised if it resolved everything.

I saw a test of uTP traffic both to and from Windows and Linux...and the results suggest only the source needed to be a Windows PC to have the extremely high packet loss result and low artificial speed limit on a per-peer basis. Even Windows 10 (or 11?) probably exhibits the same symptoms, but I haven't bothered to check it. The family PC that has Windows 10 on it...other family members may not appreciate me doing that level of testing on it!

arvidn:
I believe this patch is pretty solid now. Has anyone else tested it?

thrnz:
I've tried it out with live torrents as of the most recent push without noticing any obvious problems, though I guess any FIN issues might not be that noticeable to begin with.

Is the `num_fin_resends` setting still used after these changes? It looks like we're now always giving up after the first timeout after sending the FIN, so the `num_fin_resends` check might be redundant.

arvidn:
> Is the `utp_fin_resends` setting still used after these changes? It looks like we're now always giving up after the first timeout after sending the FIN, so the `utp_fin_resends` check might be redundant.

good point. I'll take a look at this. I should probably restore that behavior.

src/utp_stream.cpp
```diff
@@ -981,35 +1062,59 @@ void utp_socket_impl::writable()
...
 
-	maybe_trigger_send_callback();
+	// try to flush the last part of the send buffer before sending a FIN
+	if (m_out_eof && m_nagle_packet) send_pkt();
+
+	// this handles the case where send_fin() was called while stalled
+	if (m_out_eof
+		&& !m_nagle_packet
+		&& m_write_buffer_size == 0
+		&& state() == state_t::connected
```
thrnz, line 1080: When `send_fin()` fails due to stalling, it still progresses to `state_t::fin_sent`. Should this check be `state() == state_t::fin_sent` (as it was prior to this patch) in order to pick it up?

arvidn: good point. Maybe the state shouldn't even be transitioned in case the socket is stalled, failing to send the FIN.

arvidn: it's not obvious to me that checking the socket state here is important either.

thrnz: This bit might actually be able to be dropped entirely. If sending the FIN packet stalls, it gets marked for resend and ends up in `m_outbuf`, and I think should get resent in one of the previous `send_pkt()` calls above.

It looks like the second commit in #7153 might have been a bit wonky and tried to address something that was already solved by it's first commit, so the logic here no longer seems to makes sense.

arvidn: yeah. I hesitate dropping this until there are proper unit tests for these cases.

thrnz: If I'm understanding it correctly, the old conditions would never be met, so functionally nothing was happening here anyway. When `state() == state_t::fin_sent`, `m_outbuf.at(m_seq_nr)` would be the FIN packet. Once it's ACK'd and removed from `m_outbuf`, we'll no longer be in the `fin_sent` state. The assumption was that a stalled FIN packet wouldn't have made its way into `m_outbuf` (which now happens) and could be detected here and re-sent - it instead gets resent in the `send_pkt()` calls above along with any other packets needing to be resent.

I'm not sure what effect the new conditions here using the `connected` state instead of the `fin_sent` state might have.

Reproducing stalled FIN packets in a test would be ideal.

arvidn:
> How do you see/check for how many re-transmits are done? Maybe adding adebug/logged counter of those can be of help?

I think the next step in actually making this uTP implementation robust would be to add unit tests (as opposed to the existing end-to-end simulations). Where exact packets and their timings are crafted and injected by the test itself. The incoming packets callback and the send-packet callback could already be hooked into by a test, the main missing interface is the clock. The utp code currently calls into the `chrono` clock directly right now, to ask what time it is. If the time could be overridden, I think that would enable proper unit tests.

Perhaps the timestamp could be passed in to the incoming packet callback.

thrnz:
> How do you see/check for how many re-transmits are done? Maybe adding adebug/logged counter of those can be of help?

There's a `utp_packet_resend` performance counter that increments when a packet is resent.

I wonder if adding a 'utp stall' counter might be worthwhile. It could be handy for picking up too shallow buffers.

rafi-d:
Sounds like a good idea.

thrnz:
> Sounds like a good idea. Is this resend counter visible in the app?or only with the debugger?

There's some more info [here](https://www.libtorrent.org/manual-ref.html#session-statistics) about the available counters and gauges and how to access them.

PR #7234, opened 2023-01-02
fix cancelling deferred ACK

arvidn:
when cancelling a deferred ACK, also tell the socket manager about it (to prevent it from triggering it later). This is especially important when destructing, in case the socket manager still has a reference to the socket after it is destructed.

This could possibly fix this issue: https://github.com/qbittorrent/qBittorrent/issues/18157#issuecomment-1347030504

### DIFF
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -224,10 +224,10 @@ udp::endpoint utp_socket_impl::remote_endpoint() const
 	return {m_remote_address, m_port};
 }
 
-void utp_socket_impl::send_ack()
+void utp_socket_impl::send_deferred_ack()
 {
-	if (!m_deferred_ack)
-		return;
+	TORRENT_ASSERT(m_deferred_ack);
+	if (!m_deferred_ack) return;
 	m_deferred_ack = false;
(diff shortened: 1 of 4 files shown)


arvidn:
@thrnz this is amending your patch a bit

thrnz:
This looks good to me. Well spotted.

src/utp_stream.cpp
```diff
@@ -224,10 +224,10 @@ udp::endpoint utp_socket_impl::remote_endpoint() const
 	return {m_remote_address, m_port};
 }
 
-void utp_socket_impl::send_ack()
+void utp_socket_impl::send_deferred_ack()
 {
-	if (!m_deferred_ack)
-		return;
+	TORRENT_ASSERT(m_deferred_ack);
+	if (!m_deferred_ack) return;
```
thrnz, line 230: `if (!m_deferred_ack) return;`

This check might no longer be needed, as we shouldn't be able to reach here without a pending deferred ACK any more.

arvidn: right, hence the assert. I left it in as a precaution

Issue #7237, opened 2023-01-03
libtorrent webtorrent websocket tracker peer_id changing on announce answer


SilentBot1:
**Please provide the following information**

**libtorrent version (or branch):**
`main`/`2.1.0.0`

**Platform/architecture:**
`Windows 11 x64`, `Gentoo x64`, `Ubuntu 22.04 x64`

**Compiler and compiler version:**
 `gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0` on Ubuntu 22.04 x64
`Microsoft (R) C/C++ Optimizing Compiler Version 19.29.30145 for x64` on Windows 11 x64

**Please describe what symptom you see, what you would expect to see instead and how to reproduce it:**

When using a compiled version of libtorrent 2.1.0.0 ([libtorrent@e22f3ee](https://github.com/arvidn/libtorrent/commit/e22f3ee7d3dc87aea79c1c0ecacfeffaea17866d)) compiled for qBittorrent v4.6.0apha1, and standalone libtorrent 2.1.0.0 via [libtorrent4j](https://github.com/aldenml/libtorrent4j) in Java, in conjunction with bittorrent-tracker ([bittorrent-tracker@df39fe](https://github.com/webtorrent/bittorrent-tracker/commit/df39fed6faf20b6a345fe6136759294ddbbe7b65)) or aquatic-ws ([aquatic-ws@8bf0e12c](https://github.com/greatest-ape/aquatic/commit/8bf0e12cb4553fe0e9dc4985a42bda7fc56e5fd3)), I have noted that the `peer_id` provided by libtorrent webrtc peers via websockets is different per `announce` when `answers` are sent to the tracker.

This issue can be observed when there are two or more peers within a swarm, when a libtorrent client sends an announcement which contains answers, when this occurs, the announcement sent will contain a new `peer_id`, which is different than the `peer_id` provided in previous announcements.

While these `answers` are sent with a new `peer_id`, all other announcements, including `offers` are still sent from the client with it's original `peer_id`, which I believe is unique per swarm for privacy reasons.

In bittorrent-tracker, this is silently ignored by default and treated as an announcement update event, so peer discovery still functions, though with how the tracker keeps tracks of peers, this ends up with an artificially inflated peer count per libtorrent peer. When using other trackers, such as aquatic-ws, clients are disconnected with the error message `Only one peer id can be used per torrent`.

**Steps to reproduce:**
[...]

PR #7249, opened 2023-01-09
Fix WebSocket tracker peer id changing in answer

paullouisageneau:
The `rtc_signaling` implementation currently generates a new `peer_id` for each offer and answer, following an early attempt to mimic the per-connection `peer_id` mechanism in libtorrent. However, this is not possible with the WebTorrent tracker protocol and as a result answers announces are wrongly sent with changing peer ids while offers batches are correctly sent with the peer id generated for the torrent. This causes incompatibility with trackers and possible connection issues.

This PR fixes the behavior by always using the `peer_id` generated for the torrent in  `rtc_signaling`.

Fixes https://github.com/arvidn/libtorrent/issues/7237

### DIFF
--- a/include/libtorrent/aux_/torrent.hpp
+++ b/include/libtorrent/aux_/torrent.hpp
@@ -772,6 +772,9 @@ namespace libtorrent::aux {
 		// The key is passed to http trackers as ``&key=``.
 		std::uint32_t tracker_key() const;
 
+		// the peer id for tracker interaction only, generated when we add the torrent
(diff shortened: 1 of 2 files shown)


arvidn (approved):
generally I'd like to see less coupling between `rtc_signalling` and `torrent`. I'm envisioning perhaps passing in the alert_manager and settings. Maybe even in the long run have the websocket_tracker connection talk directly to `rtc_signalling` rather than going through `torrent`.

This is good for now though

PR #7253, opened 2023-01-09
update documentation for metadata_received_alert

arvidn:


include/libtorrent/alert_types.hpp
```diff
@@ -1329,28 +1329,37 @@ TORRENT_VERSION_NAMESPACE_3
...
+	// but also a copy of torrent file:
+	// .. code:: c++
+	//
+	//	save_resume_data_alert* alert = ...;
+	//	std::vector<char> buf = write_resume_data_buf(alert->params);
+	//	std::ofstream f("resume-data");
+	//	f.write(buf.data(), buf.size());
+	//
+	// If writing a .torrent file that may be a v2 torrent and may not yet have
+	// downloaded the piece layers from its peers, you need to specifu the
```
markmdscott, line 1354: minor typo

include/libtorrent/alert_types.hpp
```diff
@@ -1329,28 +1329,37 @@ TORRENT_VERSION_NAMESPACE_3
...
-	// only those that needs to download it from peers (when utilizing the libtorrent extension).
+	// only those that needs to download it from peers (via ut_metadata).
 	//
 	// There are no additional data members in this alert.
 	//
-	// Typically, when receiving this alert, you would want to save the torrent file in order
-	// to load it back up again when the session is restarted. Here's an example snippet of
-	// code to do that:
-	//
+	// Typically, when receiving this alert, you would want to trigger saving
```
markmdscott, line 1336: You might want to show the example of saving the torrent when receiving the metadata alert as it is a simpler case equivalent to the existing deprecated code and less convoluted than having to have other code in another part of the code to do this when the example could be self contained in one spot in the metadata alert.

arvidn: saving a file with just an info-dict (from the `torrent_info` obect) is not equivalent to the existing, deprecated, code. The orignal use of `torrent_info` was that it contained everything you find in a .torrent file. That's no longer the case, now it's just the (immutable) info-dict. no piece layers, no trackers no web seeds.

PR #7255, opened 2023-01-10
move towards an immutable torrent_info

arvidn:


docs/building.rst
```diff
@@ -392,12 +392,19 @@ Build features
...
+|                          | * ``3`` libtorrent-2.0                             |
+|                          | * ``4`` libtorrent-2.1                             |
+|                          | For backwards compatibility, it's also possible    |
+|                          | to specify:                                        |
+|                          | * ``on`` - Includes deprecated functions from the  |
+|                          |   oldest supported version.                        |
+|                          | * ``off`` - Disables all deprecated functions.     |
+|                          |   Same as specifying the newest version.           |
 +--------------------------+----------------------------------------------------+
 | ``i2p``                  | * ``on`` - default. build with I2P support         |
```
markmdscott, line 409: Is i2p support still current? I thought it bit rotted.

include/libtorrent/aux_/resolve_links.hpp
```diff
@@ -32,29 +32,29 @@ namespace libtorrent::aux {
 	{
 		struct TORRENT_EXTRA_EXPORT link_t
 		{
-			std::shared_ptr<const torrent_info> ti;
+			std::shared_ptr<torrent_info const> ti;
```
markmdscott, line 35: Is it stylistic best practice to prefer a const at the end now, instead of at the beginning?

src/load_torrent.cpp
```diff
@@ -16,9 +16,8 @@ see LICENSE file.
 namespace libtorrent {
 
 namespace {
-	void update_atp(add_torrent_params& atp)
+	void update_atp(std::shared_ptr<torrent_info> ti, add_torrent_params& atp)
```
markmdscott, line 19: As I understand it, a torrent_info is composed of the immutable info dict part, in addition to the trackers, comment, etc.

Would it be preferable to have a non const torrent info structure, that encapsulates a const info dict structure?

include/libtorrent/aux_/resolve_links.hpp
arvidn: I subscribe to "east const": https://hackingcpp.com/cpp/design/east_vs_west_const.html

src/load_torrent.cpp
arvidn: the mutable type that contains everything in a torrent file, including things outside of the info dictionary is `add_torrent_params`. `torrent_info` will be transitioned to only contain the immutable information from the info-dictionary.

src/torrent.cpp
```diff
@@ -304,11 +306,13 @@ bool is_downloading_state(int const st)
...
 				continue;
 
+#if TORRENT_ABI_VERSION < 4
 			// add the tracker to the m_torrent_file here so that the trackers
 			// will be preserved via create_torrent() when passing in just the
 			// torrent_info object.
 			if (!m_torrent_file->is_valid())
 				m_torrent_file->add_tracker(e.url, e.tier, lt::announce_entry::tracker_source(e.source));
+#endif
 		}
```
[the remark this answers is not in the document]
arvidn: it's not the checks that are gated, it's setting trackers and web seeds.

src/load_torrent.cpp
glassez: > `torrent_info` will be transitioned to only contain the immutable information from the info-dictionary.

👍

arvidn: I think you're absolutely right. One obstacle right now is that `file_storage` is used both for creating new torrents as well as representing existing (immutable) torrents. It makes that class much more complex and less efficient than it would need to. My plan is to shift away from using `file_storage` when creating new torrents, and instead use a separate class. The new API for creating torrents has been implemented in the `master` branch and documented here: https://github.com/arvidn/libtorrent/blob/master/docs/upgrade_to_2.1.rst#creating-torrents

Perhaps I misunderstand and what you're saying is that using `add_torrent_params` both for resume data and non-info-dict .torrent file fields is problematic. I agree with that the name of that class is not great when used to represent a .torrent file, but I think it's a pretty good fit functionality wise.

glassez: > Perhaps I misunderstand and what you're saying is that using `add_torrent_params` both for resume data and non-info-dict .torrent file fields is problematic.

I don't remember talking about it... But since we are talking about this, I believe that such reuse does not bring any real benefit, except for the momentary benefit of writing a slightly smaller amount of code. Even if these data coincide in some way, I would still make their difference explicit, i.e. I would have a separate class representing a torrent file (and another one for a magnet link), so that specific functions (e.g. parsing/generating routines)  accept parameters of the corresponding types, and a couple of helpers that would generate `add_torrent_params` from their data and vice versa (you could even store `add_torrent_params` as a private field of that classes, it is still better than re-use it directly).

>I agree with that the name of that class is not great when used to represent a .torrent file, but I think it's a pretty good fit functionality wise.

I would still use a name that reflects what this structure represents, and not what it is used for, especially since now it is used not only to add a torrent, but also to save resume data. IMO, even such a simple name as `torrent_data` is much better.

glassez:
@arvidn 
In addition to these and other related changes. Now no code (with the exception of the one left for backward compatibility and explicitly marked as deprecated) should refer to torrent_info as a "torrent file" (e.g. `torrent_handle::torrent_file()` should become `torrent_handle::torrent_info()` in new API).

arvidn:
yes, that's a good point

Issue #7248, opened 2023-01-07
[I2P] unknown error


Vort:
After #7247 fix applied, another error appears.
It happens because of incorrect initialization of `remaining` variable here:
https://github.com/arvidn/libtorrent/blob/690881fec9cc069976cd3d909254a87c22f43ce0/include/libtorrent/i2p_stream.hpp#L285
When such code is used, `remaining` also contains null character, which prevents this comparison from working correctly:
https://github.com/arvidn/libtorrent/blob/690881fec9cc069976cd3d909254a87c22f43ce0/include/libtorrent/i2p_stream.hpp#L309
When such line is used instead:
`string_view remaining(m_buffer.data(), m_buffer.size() - 1);`
unknown error disappears.
Related: #2530

_libtorrent version (or branch)_: 690881fec9cc069976cd3d909254a87c22f43ce0
_platform/architecture_: Windows x64
_compiler and compiler version_: clang 14.0.0

jiigen:
Did some tests too  and this is what i get:
`successfully listening on [uTP] 0.0.0.0:6881
I2P+ Universal Installer 2.0.0+ added
I2P+ Universal Installer 2.0.0+: state changed to: dl metadata
added torrent: I2P+ Universal Installer 2.0.0+
I2P+ Universal Installer 2.0.0+ resumed
I2P+ Universal Installer 2.0.0+ (http://tracker2.postman.i2p/announce.php)[0.0.0.0:6881] v1  "Waiting for i2p acceptor from SAM bridge" (1)
I2P+ Universal Installer 2.0.0+ (http://tracker2.postman.i2p/announce.php)[0.0.0.0:6881] v1  "Waiting for i2p acceptor from SAM bridge" (2)
I2P+ Universal Installer 2.0.0+ (http://tracker2.postman.i2p/announce.php)[0.0.0.0:6881] v1 Host not found (authoritative) "" (3)`

No more unknown error

PR #7256, opened 2023-01-10
fix I2P issue introduced in 2.0.0

arvidn:


arvidn:
@Vort would you mind testing this patch? I think the other `read_line()` -> `start_read_line()` was another bug

Vort:
Both places now correspond to the state of code before 23730776adc078c9b9f4b7b14672fcbfad4e966f:
https://github.com/arvidn/libtorrent/blob/80b9552feb31cbe060fea8ea5ae127a5bf320b14/src/i2p_stream.cpp#L300-L301
https://github.com/arvidn/libtorrent/blob/80b9552feb31cbe060fea8ea5ae127a5bf320b14/src/i2p_stream.cpp#L465
This is good.

I made tests with this PR and fix from #7248 and now I see `GET /announce.php?` request for the first time.
So it looks like second change is correct also from test results.

_(stream connection with GET request gets instantly disconnected after request is sent, but that is another story)_

Issue #7260, opened 2023-01-14
I2P: incorrect parsing of peer list from tracker response


Vort:
In `parse_tracker_response` function of `http_tracker_connection.cpp` `#if TORRENT_USE_I2P` section exists, which should trigger when response from i2p tracker is processed.
However, because of bug, this check fails and `peers` gets processed as regular IP addresses:
https://github.com/arvidn/libtorrent/blob/b17f6150ecb0cf5b335305124a4400fc81a03831/src/http_tracker_connection.cpp#L585
To fix it, I added to `torrent_info::add_tracker` function in `torrent_info.cpp` these lines:
```cpp
#if TORRENT_USE_I2P
		if (is_i2p_url(e.url)) m_flags |= i2p;
#endif
```
after this line:
https://github.com/arvidn/libtorrent/blob/b17f6150ecb0cf5b335305124a4400fc81a03831/src/torrent_info.cpp#L1748

With such change, I start to see naming lookup requests for peers:
`NAMING LOOKUP NAME=...`
`NAMING REPLY RESULT=OK NAME=...`

Related: https://github.com/arvidn/libtorrent/issues/2530

_libtorrent version (or branch)_: 690881fec9cc069976cd3d909254a87c22f43ce0 + fixes from #7247, #7248 and #7259
_platform/architecture_: Windows x64
_compiler and compiler version_: clang 14.0.0

Vort:
It turned out that changes in `torrent_info::add_tracker` fixed this problem only partially.
That's why in #7272 different soultion was implemented - fix is located in `torrent::announce_with_tracker` now.
Most likely, it solved this problem completely, that's why I'm closing this issue.
But if someone finds that addresses still parsed incorrectly in some cases - please let me know, I will reopen this bugreport.

Issue #7262, opened 2023-01-14
I2P: no outgoing connections to peers after resolve is complete


Vort:
There are two similar functions in `torrent.cpp`: `on_peer_name_lookup` and `on_i2p_resolve`.
One of them, `on_peer_name_lookup` contains `update_want_peers();` call at the end:
https://github.com/arvidn/libtorrent/blob/b17f6150ecb0cf5b335305124a4400fc81a03831/src/torrent.cpp#L3795
But other one, `on_i2p_resolve`, do not have it:
https://github.com/arvidn/libtorrent/blob/b17f6150ecb0cf5b335305124a4400fc81a03831/src/torrent.cpp#L3744
It looks like a bug.
When I added `update_want_peers();` call at the end of `on_i2p_resolve` function, my test program started working as expected: lots of outgoing connections start to spawn and download start going.

Related: https://github.com/arvidn/libtorrent/issues/2530

_libtorrent version (or branch)_: 690881fec9cc069976cd3d909254a87c22f43ce0 + fixes from #7247, #7248, #7259 and #7260
_platform/architecture_: Windows x64
_compiler and compiler version_: clang 14.0.0

jiigen:
Going to test these changes  and will share my result. Since I use a different OS and compiler could be useful I guess.

jiigen:
No luck this time...Added both your changes to the code and recompiled but I only seen the  naming lookup. After that it get stuck as before and no download nor connection to peers.
To be clear I've added: _update_want_peers();_  after _peers_erased(st.erased);_

Vort:
> To be clear I've added: update_want_peers(); after peers_erased(st.erased);

Right, at line # 3744.

> Added both your changes to the code and recompiled but I only seen the naming lookup.

To diagnose problem you can replace `lt::alert_category::status | lt::alert_category::error` with `lt::alert::all_categories` in test program.

If output have `I2P+ Universal Installer 2.0.0+ (http://tracker2.postman.i2p/announce.php)[0.0.0.0:6881] v1 received peers: 37`, then it confirms that previous fix (peer list / naming lookup) is working.
If after that several `I2P+ Universal Installer 2.0.0+: i2p_resolve error: invalid key` appear, then it is fine - for some reason not every name resolves correctly.

Without `update_want_peers();` nothing happens next.
[...]

Issue #7259, opened 2023-01-13
I2P: Host not found


Vort:
This problem happens because of incorrect refactoring of `http_connection.cpp` file in 592544f30f587294351bafce3dee8f53ff7cf03e commit.

This is how code should look like:
https://github.com/arvidn/libtorrent/blob/417405fc892903c8f850ca88d02bc045e56d2e19/src/http_connection.cpp#L387-L389
And this is what it was changed to:
https://github.com/arvidn/libtorrent/blob/592544f30f587294351bafce3dee8f53ff7cf03e/src/http_connection.cpp#L385-L388

It was still not possible for me to download file after I reverted this change, but
1. There were no `Host not found` errors anymore.
2. I can see `HTTP/1.1 200 OK` reply from tracker now.

Related: https://github.com/arvidn/libtorrent/issues/2530

_libtorrent version (or branch)_: 690881fec9cc069976cd3d909254a87c22f43ce0 + fixes from #7247 and #7248
_platform/architecture_: Windows x64
_compiler and compiler version_: clang 14.0.0

jiigen:
I can confirm same results on:
__libtorrent version (or branch): https://github.com/arvidn/libtorrent/commit/690881fec9cc069976cd3d909254a87c22f43ce0 + fixes_ from https://github.com/arvidn/libtorrent/issues/7247 and https://github.com/arvidn/libtorrent/issues/7248
platform/architecture:  Linux Mint 20.2 Uma x64
compiler and compiler version: gcc 9.4.0_

Vort:
> after a while it started the download and finished it

This is important result, thank you!
I started program once again and after 30 minutes of waiting I saw some activity.
For some reason test program just exited without any additional output, but looks like I've figured out what is happening.
Most likely, outgoing connections are broken, but incoming connections works somehow.
It explains both long waiting times and eventual success (after some remote peer decides to connect to our address).

Vort:
Fixed with #7272.

Issue #7258, opened 2023-01-13
[I2P] Update to SAM 3.1


zzzi2p:
ref: https://geti2p.net/en/docs/api/samv3

Supported since 2014 and the minimum recommended version.

Changes:

- HELLO VERSION MIN=3.1 MAX=3.1
- SESSION CREATE DESTINATION=TRANSIENT SIGNATURE_TYPE=7 ...

We also recommend specifying inbound.quantity=x outbound.quantity=x in the SESSION CREATE line, as i2pd and Java I2P have different defaults.

More background and guidance in the spec linked above.
Thanks

Vort:
Are changes in #7272 enough to close this issue?

PR #7272, opened 2023-01-16
improve I2P support

### LINKED ISSUE #7248: [I2P] unknown error

Vort:
Should fix #7248, #7258, #7259, #7260, #7262 and, probably, #2530.

### DIFF
--- a/include/libtorrent/i2p_stream.hpp
+++ b/include/libtorrent/i2p_stream.hpp
@@ -193,7 +193,7 @@ private:
 
 		// send hello command
 		m_state = read_hello_response;
-		static const char cmd[] = "HELLO VERSION MIN=3.0 MAX=3.0\n";
+		static const char cmd[] = "HELLO VERSION MIN=3.1 MAX=3.1\n";
 
 		ADD_OUTSTANDING_ASYNC("i2p_stream::start_read_line");
 		async_write(m_sock, boost::asio::buffer(cmd, sizeof(cmd) - 1), wrap_allocator(
@@ -277,7 +277,8 @@ private:
 				break;
 		}
 
-		string_view remaining(m_buffer.data(), m_buffer.size());
+		TORRENT_ASSERT(m_buffer[int(m_buffer.size()) - 1] == '\0');
+		string_view remaining(m_buffer.data(), m_buffer.size() - 1);
 		string_view token;
 
 		std::tie(token, remaining) = split_string(remaining, ' ');
@@ -427,8 +428,9 @@ private:
 		TORRENT_ASSERT(m_magic == 0x1337);
 		m_state = read_session_create_response;
 		char cmd[400];
-		int size = std::snprintf(cmd, sizeof(cmd), "SESSION CREATE STYLE=STREAM ID=%s DESTINATION=TRANSIENT\n"
-			, m_id);
+		int size = std::snprintf(cmd, sizeof(cmd),
+			"SESSION CREATE STYLE=STREAM ID=%s DESTINATION=TRANSIENT SIGNATURE_TYPE=7 "
+			"inbound.quantity=3 outbound.quantity=3 inbound.length=3 outbound.length=3\n", m_id);
 		ADD_OUTSTANDING_ASYNC("i2p_stream::start_read_line");
 		async_write(m_sock, boost::asio::buffer(cmd, std::size_t(size)), wrap_allocator(
 			[this](error_code const& ec, std::size_t, Handler hn) {
--- a/src/http_connection.cpp
+++ b/src/http_connection.cpp
@@ -368,11 +368,11 @@ void http_connection::start(std::string const& hostname, int port
 		}
 		else
 #endif
-		m_hostname = hostname;
 		if (ps && ps->proxy_hostnames
 			&& (ps->type == settings_pack::socks5
 				|| ps->type == settings_pack::socks5_pw))
 		{
+			m_hostname = hostname;
 			m_port = std::uint16_t(port);
 			m_endpoints.emplace_back(address(), m_port);
 			connect();
@@ -385,6 +385,7 @@ void http_connection::start(std::string const& hostname, int port
 				, std::bind(&http_connection::on_resolve
 				, me, _1, _2));
 		}
+		m_hostname = hostname;
 		m_port = std::uint16_t(port);
 	}
 }
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -3145,7 +3145,7 @@ namespace {
 					req.key = tracker_key();
 
 #if TORRENT_USE_I2P
-					if (is_i2p())
+					if (is_i2p_url(req.url))
 					{
 						req.kind |= tracker_request::i2p;
 					}
@@ -3741,6 +3741,8 @@ namespace {
 		if (m_peer_list->add_i2p_peer(dest, peer_info::tracker, {}, &st))
 			state_updated();
 		peers_erased(st.erased);
+
+		update_want_peers();
 	}
 	catch (...) { handle_exception(); }
 #endif


obscuratus-i2p:
I built and tested this pull request.  It was working fine.  Using a close adaption of the simple magnet client presented by Vort, I was able to successfully complete a torrent download.  I did not experience the delays noted elsewhere, the client promptly picked up my seeder.  Perhaps this was due to the tracker I used.  I didn't use postman's tracker, I used a ZZZOT open tracker.

include/libtorrent/i2p_stream.hpp
```diff
@@ -193,7 +193,7 @@ struct i2p_stream : proxy_base
 
 		// send hello command
 		m_state = read_hello_response;
-		static const char cmd[] = "HELLO VERSION MIN=3.0 MAX=3.0\n";
+		static const char cmd[] = "HELLO VERSION MIN=3.1 MAX=3.1\n";
```
arvidn, line 196: is there a new feature this update depend on that's only available since SAM version 3.1?

include/libtorrent/i2p_stream.hpp
```diff
@@ -277,7 +277,7 @@ struct i2p_stream : proxy_base
 				break;
 		}
 
-		string_view remaining(m_buffer.data(), m_buffer.size());
+		string_view remaining(m_buffer.data(), m_buffer.size() - 1);
```
arvidn, line 280: Fixing it this way makes me a but uneasy. Especially by not also ensuring that the character being truncated is in fact the new-line. It would seem a bit more logical to truncate the last character of `m_buffer` on line 241, instead of adding the null terminator. I believe this buffer used to be treated as null terminated.

include/libtorrent/i2p_stream.hpp
```diff
@@ -427,8 +427,8 @@ struct i2p_stream : proxy_base
 		TORRENT_ASSERT(m_magic == 0x1337);
 		m_state = read_session_create_response;
 		char cmd[400];
-		int size = std::snprintf(cmd, sizeof(cmd), "SESSION CREATE STYLE=STREAM ID=%s DESTINATION=TRANSIENT\n"
-			, m_id);
+		static const char options[] = "inbound.quantity=3 outbound.quantity=3 inbound.length=3 outbound.length=3";
```
arvidn, line 430: is it important to configure the inbound and outbound channels here? I would expect that by leaving these options out, there may be a global configuration that takes effect instead.

src/http_connection.cpp
```diff
@@ -385,6 +385,7 @@ void http_connection::start(std::string const& hostname, int port
 				, std::bind(&http_connection::on_resolve
 				, me, _1, _2));
 		}
+		m_hostname = hostname;
```
arvidn, line 388: I don't follow this. Why is this change necessary?

src/torrent_info.cpp
```diff
@@ -1747,6 +1747,10 @@ namespace {
 		e.source = source;
 		m_urls.push_back(e);
 
+#if TORRENT_USE_I2P
+		if (is_i2p_url(e.url)) m_flags |= i2p;
```
arvidn, line 1751: I believe it should be pretty easy to add a unit test for this, which I think would be good. Also, what do you think of checking `m_flags & i2p` as the first condition, to avoid calling `is_i2p_url()` when the flag has already been set.

arvidn (commented):
sorry for the delay in reviewing this.

include/libtorrent/i2p_stream.hpp
Vort: https://geti2p.net/en/docs/api/samv3:
_SAM 3.1 is the recommended minimum SAM implementation because of its support for better signature types than SAM 3.0._
I guess @zzzi2p can explain better why 3.1 is needed (#7258).

include/libtorrent/i2p_stream.hpp
Vort: Default values for Java I2P and i2pd are different.
So to have similar behaviour it is better to specify values directly.
https://geti2p.net/en/docs/api/samv3:
_Please specify the tunnel quantities in the SESSION CREATE message to get consistent performance with the Java I2P and i2pd routers, using the options e.g. inbound.length=3 outbound.length=3._

src/http_connection.cpp
Vort: This is how this block of code was implemented before commit 592544f30f587294351bafce3dee8f53ff7cf03e.
This commit broke I2P support so most safe change in my opinion was to just revert it.

Probably, there is better way of fixing this problem, but I don't perfectly understand how this block of code works to make it better than it was before.

include/libtorrent/i2p_stream.hpp
Vort: Ok, I will look if I can implement it better.

src/http_connection.cpp
arvidn: so moving the assignment to `m_hostname` breaks i2p support?
by manual inspection, they look like they ought to have identical behavior

include/libtorrent/i2p_stream.hpp
arvidn: I see. is there a good reason to not include these directly in the format string?

jiigen: Keeping those out of the format string has been a choice of mine when i first tried that. The reason is that those should be parameters that should be available to set when creating the connection by the client.

src/http_connection.cpp
Vort: Commit 592544f30f587294351bafce3dee8f53ff7cf03e puts assignment between `else` and `if`, which executes `if` regardless of what happens inside `#if TORRENT_USE_I2P` block.

include/libtorrent/i2p_stream.hpp
Vort: Yes, I made it like that because this is how it was suggested by @jigen.
Also I worried about too long strings not fitting in text editor.
It is possible to solve it by breaking string in two parts.
I was thinking about such feature as something non-standard, but I searched more about it, and now I know that adjacent string literals concatenation is perfectly fine method.
I will make this change.

jiigen: I even thought that separating made it more readable than a long string.

include/libtorrent/i2p_stream.hpp
zzzi2p: 3.1 has been supported since 2014 and is the only version supported by i2pd. 3.1 gives you SIGNATURE_TYPE=7 (Ed25519) which gets your users off of insecure DSA_SHA1 signatures. There is no risk of finding a pre-2014 router to talk to, they wouldn't work anyway.

include/libtorrent/i2p_stream.hpp
zzzi2p: My recommendations about the options, although perhaps better implemented by arvidn later than by Vort in this PR, are here: https://github.com/arvidn/libtorrent/issues/7262#issuecomment-1385362013

Vort: @jiigen look how I made it in 022b53cdc9b42884bd1fb2303be8f81cf5758472.
It is separated and at the same time it is single string.

include/libtorrent/i2p_stream.hpp
Vort: Looks like null is required at line 247.
I can change `m_dest = &m_buffer[0];` to `m_dest.assign(&m_buffer[0], m_buffer.size());` there.
And then at line 241 `m_buffer[read_pos - 1] = 0;` to `m_buffer.pop_back();`
@arvidn will it be better than ` - 1`?

include/libtorrent/i2p_stream.hpp
jiigen: Looks fine to me.

src/torrent_info.cpp
Vort: I made more tests and found that added check to `torrent_info::add_tracker` fixes #7260 only when fresh torrent is added.
When torrent is started from resume data, `torrent::is_i2p` function returns `false` because `m_torrent_file` have no `m_urls`.
Because of it, I propose to change line
https://github.com/arvidn/libtorrent/blob/32e36984c944e85eacc57c381db4605d6636aa84/src/torrent.cpp#L3148
to `if (is_i2p_url(req.url))` instead.
@arvidn can you check if it is better way to fix #7260 ?

include/libtorrent/i2p_stream.hpp
arvidn: that's a good point. So this is really just ignoring the null terminator in `m_buffer`.
would you mind adding a `TORRENT_ASSERT(m_buffer[m_buffer.size() - 1] == '\0');`

src/http_connection.cpp
arvidn: ouch. I see that now

src/torrent_info.cpp
arvidn: when I made the comment I mistakenly thought this was done in a loop.
I think setting the flag here makes more sense than checking tracker URLs in the `is_i2p()` function

include/libtorrent/i2p_stream.hpp
Vort: I added assert in f6d329a4c0df214053edee9d2a11457e89476bfe, but slightly modified it because of compiler complaints.

arvidn: at a second thought it would probably have been nicer to make it:
```
TORRENT_ASSERT(m_buffer.back() == '\0');
```
but what you have is fine.

arvidn (approved):

arvidn:
I intend to land this once CI is green. This should be mentioned in the ChangeLog, but I could add that later

arvidn:
@Vort @jiigen Thanks for your contributions!

PR #7280, opened 2023-01-21
Test merge

arvidn:


tox.ini
```diff
@@ -41,7 +41,7 @@ deps =
 #   in the python directory, only run mypy on the tests dir.
 
 commands =
-    {envpython} setup.py bdist_wheel
+    {envpython} setup.py build_ext --b2-args "asserts=on invariant-checks=full" bdist_wheel
```
arvidn, line 44: @AllSeeingEyeTolledEweSew is this reasonable? I just want to make the CI build with these options, not release build. This being under `[testenv]` makes me think it only affects tests.

PR #7288, opened 2023-01-26
Get close reason apply visitor types

cyrillicw:
get_close_reason in socket_type.cpp applies const reference to socket as a parameter. That's why visitor should expect const reference to utp_socket. Otherwise none is always returned


arvidn:
good catch!

PR #7292, opened 2023-01-30
Add macOS arm64 wheels build and tests to .cirrus.yml

### LINKED ISSUE #7131: Python binding for macOS aarch64 fails to install
Currently cannot install libtorrent on macOS running Apple Silicon (arm64)

```
$ pip install libtorrent==2.0.7
ERROR: Could not find a version that satisfies the requirement libtorrent==2.0.7 (from versions: none)
ERROR: No matching distribution found for libtorrent==2.0.7
```

moodyjon:
Fixes #7131

Does not have any automated pypi upload, but otherwise works for cp39 and cp310 wheels. The artifacts have to be downloaded from Cirrus-CI servers.

https://cirrus-ci.org/guide/writing-tasks/#latest-build-artifacts

```
Uploading 2 artifacts for /var/folders/76/zy5ktkns50v6gt5g8r0sf6sc0000gn/T/cirrus-ci-build/wheelhouse/*
Uploaded /var/folders/76/zy5ktkns50v6gt5g8r0sf6sc0000gn/T/cirrus-ci-build/wheelhouse/libtorrent-2.0.8-cp310-cp310-macosx_11_0_arm64.whl
Uploaded /var/folders/76/zy5ktkns50v6gt5g8r0sf6sc0000gn/T/cirrus-ci-build/wheelhouse/libtorrent-2.0.8-cp39-cp39-macosx_11_0_arm64.whl
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -38,7 +38,6 @@ build_and_store_wheels: &BUILD_AND_STORE_WHEELS
 
 build_macos_arm64_task:
   name: Build macOS arm64 wheels.
-  trigger_type: manual
 
   macos_instance:
     image: ghcr.io/cirruslabs/macos-monterey-xcode
@@ -63,7 +62,6 @@ build_macos_arm64_task:
 
 test_macos_arm64_task:
   name: Test macOS arm64.
-  trigger_type: manual
 
   macos_instance:
     image: ghcr.io/cirruslabs/macos-monterey-xcode
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -38,6 +38,8 @@ build_and_store_wheels: &BUILD_AND_STORE_WHEELS
 
 build_macos_arm64_task:
   name: Build macOS arm64 wheels.
+  trigger_type: manual
+
   macos_instance:
     image: ghcr.io/cirruslabs/macos-monterey-xcode
 
@@ -61,6 +63,8 @@ build_macos_arm64_task:
 
 test_macos_arm64_task:
   name: Test macOS arm64.
+  trigger_type: manual
+
   macos_instance:
     image: ghcr.io/cirruslabs/macos-monterey-xcode
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -34,11 +34,10 @@ build_macos_arm64_task:
     image: ghcr.io/cirruslabs/macos-monterey-xcode
 
   env:
-    PATH: /opt/homebrew/opt/python@3.9/bin:$PATH
+    PATH: /opt/homebrew/bin:$PATH
 
   install_pre_requirements_script:
     - brew install python@3.9
-    - ln -s python3 /opt/homebrew/opt/python@3.9/bin/python
   <<: *BUILD_AND_STORE_WHEELS
 
 test_macos_arm64_task:
@@ -47,11 +46,10 @@ test_macos_arm64_task:
     image: ghcr.io/cirruslabs/macos-monterey-xcode
 
   env:
-    PATH: /opt/homebrew/opt/python@3.9/bin:$PATH
+    PATH: /opt/homebrew/bin:$PATH
 
   install_pre_requirements_script:
     - brew install python@3.9
-    - ln -s python3 /opt/homebrew/opt/python@3.9/bin/python
     - brew install boost-build boost openssl@1.1
     - echo "using darwin ;" >>~/user-config.jam
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -25,7 +25,7 @@ build_and_store_wheels: &BUILD_AND_STORE_WHEELS
     - echo $PATH
     - ls -l /opt/homebrew/bin
   install_cibuildwheel_script:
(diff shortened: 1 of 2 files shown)


robd003:
@AllSeeingEyeTolledEweSew could you take a look at this?

Jamfile
```diff
@@ -396,7 +396,7 @@ rule openssl-lib-path ( properties * )
 		# on macOS, default to pick up openssl from the homebrew installation
 		# brew install openssl
 		# homebrew on M1 Macs install to /opt/homebrew
-		OPENSSL_LIB = /opt/homebrew/opt/openssl/lib /usr/local/opt/openssl/lib ;
+		OPENSSL_LIB = /opt/homebrew/opt/openssl@1.1/lib /opt/homebrew/opt/openssl/lib /usr/local/opt/openssl/lib ;
```
arvidn, line 399: I would prefer you put version specific paths in the workflow file rather than here.
you can specify this path on the `b2` command line by adding: `openssl-lib=/opt/homebrew/opt/openssl@1.1/lib`

Jamfile
```diff
@@ -423,7 +423,7 @@ rule openssl-include-path ( properties * )
 		# on macOS, default to pick up openssl from the homebrew installation
 		# brew install openssl
 		# homebrew on M1 Macs install to /opt/homebrew
-		OPENSSL_INCLUDE = /opt/homebrew/opt/openssl/include /usr/local/opt/openssl/include ;
+		OPENSSL_INCLUDE = /opt/homebrew/opt/openssl@1.1/include /opt/homebrew/opt/openssl/include /usr/local/opt/openssl/include ;
```
arvidn, line 426: here too, you can specify this path on the `b2` command line: `openssl-include=/opt/homebrew/opt/openssl@1.1/include`

.cirrus.yml
```diff
@@ -19,3 +19,76 @@ task:
...
+    - echo "using darwin ;" >>~/user-config.jam
+
+  debug_script:
+    - echo $PATH
+    - ls -l /opt/homebrew/bin
+    - find /opt/homebrew -name '*openssl*' -ls
+
+  test_script:
+    - cd test
+    - b2 crypto=built-in -l400 warnings-as-errors=on debug-iterators=on invariant-checks=full asserts=on deterministic-tests
```
arvidn, line 88: you're building with `crypto=built-in`, which means there's no dependency on openssl, yet you changed the paths in the `Jamfile`. Did you mean to build with `crypto=openssl` here?

arvidn:
the new tasks you added don't seem to be running right. At least they're not green.

robd003:
@arvidn it looks like tasks need approval before they'll run

arvidn:
usually github doesn't show tasks as grey before approved though, but maybe cirrus is different. I just triggered one manually, and now it's green. the `main` task still ran and was green.

arvidn:
"Build macOS arm64 wheels" did not seem to have triggered by me approving and running tests.

.cirrus.yml
```diff
@@ -19,3 +19,76 @@ task:
...
+
+  run_cibuildwheel_script:
+    - cibuildwheel
+
+  wheels_artifacts:
+    path: "wheelhouse/*"
+
+build_macos_arm64_task:
+  name: Build macOS arm64 wheels.
+  trigger_type: manual
```
arvidn, line 41: I see, this is why these workflows don't run. did you mean to make these manual/

moodyjon: I made it manual based on this comment:
https://github.com/arvidn/libtorrent/issues/7131#issuecomment-1310942640

The `trigger_type` seems to only support `manual`. There are also options to restrict it to monitor certain organizations, repositories, branches, etc. I haven't found an option that runs a task specifically on pull requests.

I want to be sure the setting doesn't end up triggering too many runs per day... wouldn't want to be the one that ruins the free M1 runner pool.

.cirrus.yml
moodyjon: I tried to duplicate more or less what worflows/macos.yml does in the "test" job. Wasn't aware of all the options. I can add it if you like.

Jamfile
moodyjon: Will try this.... My concern is that the `cibuildwheels` wraps and obscures the invocation of `b2`.

moodyjon: `cibuildwheels` seems to do the Python wrappers and `libtorrent-rasterbar` under openssl@3 ("brew install openssl" with no @<version>). The include & lib for openssl@3 are apparently found under `/opt/homebrew/opt/openssl`.

.cirrus.yml
arvidn: I think if you just remove the trigger type field, and have it behave like the other cirrus jobs, I believe it would only run on pushes.
if it doesn't run on every PR, it's likely to break at some point without anyone noticing.

moodyjon: Done.

arvidn:
looks good. the next step would be to hook this up to the `cibuildwheel` workflow that I trigger when releasing. That job currently uploads packages to pypi

Issue #7196, opened 2022-11-29
Use `std::chrono::system_clock` instead of `std::chrono::high_resolution_clock`


HanabishiRecca:
`high_resolution_clock` actually leads to major confusion. It counts time from unix epoch in Linux, but counts time from boot in Windows.
`system_clock` otherwise always has strict unix epoch behavior by C++20 spec.

Ref issues:
https://github.com/qbittorrent/qBittorrent/issues/13150
https://github.com/qbittorrent/qBittorrent/pull/14815#issuecomment-824000122
https://github.com/qbittorrent/qBittorrent/issues/8396
https://github.com/qbittorrent/qBittorrent/issues/8021
https://github.com/qbittorrent/qBittorrent/issues/7461

More details:
https://stackoverflow.com/questions/68906605/c-stdchronohigh-resolution-clock-time-since-epoch-returns-too-small-number/68907126#68907126

https://github.com/Iyengar111/NanoLog/issues/11#issuecomment-275015072

HanabishiRecca:
My solution so far:
https://github.com/HanabishiRecca/libtorrent/compare/RC_2_0...RC_2_0-time
https://github.com/HanabishiRecca/libtorrent/compare/RC_1_2...RC_1_2-time

ukoz:
You have to convert time under Windows via duration_cast
Here is example of converting lt::torrent_status::last_upload to Windows format

```
lt::time_point last_upload;
char cbuf[64];
struct tm tmtime;
std::time_t added_time;
...
last_upload = lt::torrent_status::last_upload;
std::chrono::system_clock::time_point now = std::chrono::system_clock::now();
std::chrono::steady_clock::time_point snow = std::chrono::steady_clock::now();
std::chrono::system_clock::time_point tp = now + std::chrono::duration_cast<std::chrono::system_clock::duration>(last_upload - snow);
std::time_t tmptime_t = std::chrono::system_clock::to_time_t(tp);
localtime_s(&tmtime, &tmptime_t);
if (strftime(cbuf, sizeof cbuf, "%Y %b %e %H:%M:%S\n", &tmtime)) {
	printf("formatted time: %s\n",cbuf);
} else {
	printf("fail cvt\n");
}
```

arvidn:
[...]

PR #7303, opened 2023-02-11
fix time_point conversion from/to time_t

jagannatharjun:
since high_resolution_clock epoch is implementation dependent and may not be same as UNIX time, direct conversion to time_t is not possible

for a clear fix time_point should be changed std::system_clock but for now to preserve ABI special functions are introduced to fix conversion of high_resolution_clock to time_t

ref #7196

### DIFF
--- a/include/libtorrent/aux_/time.hpp
+++ b/include/libtorrent/aux_/time.hpp
@@ -37,12 +37,17 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/config.hpp"
 #include "libtorrent/time.hpp"
 
+#include <ctime> // for std::time_t
+
 namespace libtorrent { namespace aux {
 
 	// returns the current time, as represented by time_point. The
 	// resolution of this timer is about 100 ms.
 	TORRENT_EXTRA_EXPORT time_point time_now();
 	TORRENT_EXTRA_EXPORT time_point32 time_now32();
+
+	TORRENT_EXTRA_EXPORT std::time_t to_time_t(time_point32 tp);
+	TORRENT_EXTRA_EXPORT time_point32 from_time_t(std::time_t t);
 } }
 
 #endif
--- a/src/time.cpp
+++ b/src/time.cpp
@@ -32,9 +32,36 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "libtorrent/aux_/time.hpp"
 
+#include <chrono>
+
 namespace libtorrent { namespace aux {
 
 	time_point time_now() { return clock_type::now(); }
 	time_point32 time_now32() { return time_point_cast<seconds32>(clock_type::now()); }
+
+	// for simplying implementation
+	using std::chrono::system_clock;
+
+
+	// consider using std::chrono::clock_cast on C++20
+	time_t to_time_t(const time_point32 tp)
+	{
+		const auto lt_now = clock_type::now();
+		const auto sys_now = system_clock::now();
+
+		const auto r = sys_now + std::chrono::duration_cast<system_clock::duration>(lt_now - tp);
+		return system_clock::to_time_t(r);
+	}
+
+	// consider using std::chrono::clock_cast on C++20
+	time_point32 from_time_t(const std::time_t t)
+	{
+		const auto tp = system_clock::from_time_t(t);
+		const auto lt_now = clock_type::now();
+		const auto sys_now = system_clock::now();
+
+		auto r = lt_now + std::chrono::duration_cast<clock_type::duration>(sys_now - tp);
+		return std::chrono::time_point_cast<seconds32>(r);
+	}
 
 } }
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -233,8 +233,8 @@ bool is_downloading_state(int const st)
 		, m_announce_to_dht(!(p.flags & torrent_flags::paused))
 		, m_ssl_torrent(false)
 		, m_deleted(false)
-		, m_last_download(seconds32(p.last_download))
-		, m_last_upload(seconds32(p.last_upload))
+		, m_last_download(aux::from_time_t(p.last_download))
+		, m_last_upload(aux::from_time_t(p.last_upload))
 		, m_userdata(p.userdata)
 		, m_auto_managed(p.flags & torrent_flags::auto_managed)
 		, m_current_gauge_state(static_cast<std::uint32_t>(no_gauge_state))
@@ -6863,8 +6863,8 @@ namespace {
 		ret.finished_time = static_cast<int>(total_seconds(finished_time()));
 		ret.seeding_time = static_cast<int>(total_seconds(seeding_time()));
 		ret.last_seen_complete = m_last_seen_complete;
-		ret.last_upload = std::time_t(total_seconds(m_last_upload.time_since_epoch()));
-		ret.last_download = std::time_t(total_seconds(m_last_download.time_since_epoch()));
+		ret.last_upload = aux::to_time_t(m_last_upload);
+		ret.last_download = aux::to_time_t(m_last_download);
 
 		ret.num_complete = m_complete;
 		ret.num_incomplete = m_incomplete;
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -11675,11 +11675,14 @@ namespace {
 		st->seeding_time = int(total_seconds(seeding_time()));
 
 		time_point32 const unset{seconds32(0)};
+		auto const last_upload_unix = aux::to_time_t(m_last_upload),
+			last_download_unix = aux::to_time_t(m_last_download);
 
 		st->time_since_upload = m_last_upload == unset ? -1
-			: static_cast<int>(total_seconds(aux::time_now32() - m_last_upload));
+			: static_cast<int>(std::time(nullptr) - last_upload_unix);
 		st->time_since_download = m_last_download == unset ? -1
-			: static_cast<int>(total_seconds(aux::time_now32() - m_last_download));
+			: static_cast<int>(std::time(nullptr) - last_download_unix);
+
 #endif
 
 		st->finished_duration = finished_time();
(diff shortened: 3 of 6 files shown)


include/libtorrent/time.hpp
```diff
@@ -87,6 +87,33 @@ namespace libtorrent {
 	std::int64_t total_microseconds(T td)
 	{ return duration_cast<microseconds>(td).count(); }
 
+	template <typename HighResTimePoint>
```
glassez, line 90: IIRC, these functions are clock type independent. So I would drop any mentions of "high resolution", e.g. `template <typename TimePoint>`, `unix_time_to_time_point` (or even `from_time_t` to be consistent with `to_time_t`).

src/time.cpp
```diff
@@ -37,4 +37,24 @@ namespace libtorrent { namespace aux {
 	time_point time_now() { return clock_type::now(); }
 	time_point32 time_now32() { return time_point_cast<seconds32>(clock_type::now()); }
 
+	time_t to_time_t(const time_point32 &tp)
+	{
```
arvidn, line 41: I think it might be nice to add:
```
using std::chrono::system_clock;
```
here, to make the names a bit shorter in this function. Also in the other function, `from_time_t`.

include/libtorrent/aux_/time.hpp
```diff
@@ -43,6 +43,9 @@ namespace libtorrent { namespace aux {
 	// resolution of this timer is about 100 ms.
 	TORRENT_EXTRA_EXPORT time_point time_now();
 	TORRENT_EXTRA_EXPORT time_point32 time_now32();
+
+	TORRENT_EXTRA_EXPORT std::time_t to_time_t(const time_point32 &tp);
+	TORRENT_EXTRA_EXPORT time_point32 from_time_t(std::time_t t);
```
arvidn, line 48: The names of these export macros aren't great. `EXTRA` refers to symbols that aren't part of the public API, but that need to be exported in order for unit tests to access them when building as a shared library. So, as this patch stands now, there are no unit tests accessing these functions, so they shouldn't be exported.

However, perhaps there should be a test that ensures that this conversion round-trips correctly. It would probably be somewhat flaky since it depends on the wall clock time. So some measure should be taken to limit flakiness.

Perhaps the round-trip can be run 10 times, and ensure it gets back to the same `time_t` at least 9 of the times.

include/libtorrent/aux_/time.hpp
```diff
@@ -43,6 +43,9 @@ namespace libtorrent { namespace aux {
 	// resolution of this timer is about 100 ms.
 	TORRENT_EXTRA_EXPORT time_point time_now();
 	TORRENT_EXTRA_EXPORT time_point32 time_now32();
+
+	TORRENT_EXTRA_EXPORT std::time_t to_time_t(const time_point32 &tp);
```
arvidn, line 47: I don't think there's a good reason to pass this parameter by reference. it's only a 32 bit integer.

arvidn (commented):
I think this looks a lot better than the initial version. I think it would make sense to have a round-trip test though

src/time.cpp
```diff
@@ -37,4 +37,24 @@ namespace libtorrent { namespace aux {
...
+	}
+
+	time_point32 from_time_t(std::time_t t)
+	{
+		auto tp = std::chrono::system_clock::from_time_t(t);
+		auto hc_now = std::chrono::high_resolution_clock::now();
+		auto sys_now = std::chrono::system_clock::now();
+
+		auto r = hc_now + std::chrono::duration_cast<std::chrono::system_clock::duration>(sys_now - tp);
+		return  std::chrono::time_point_cast<time_point32::duration>(r);
```
arvidn, line 57: this cast seems to be failing on CI

jagannatharjun:
> I think this looks a lot better than the initial version. 

I was trying to make chatgpt write the code. In the end, I've to rewrite it in the morning 😓

src/time.cpp
jagannatharjun: I can see one failling for missing std::time_t, can't see anything related to this.

arvidn: https://github.com/arvidn/libtorrent/actions/runs/4155202534/jobs/7188113893#step:6:278

arvidn: I wonder if this could be a limitation in the simulator, or maybe a missing include

jagannatharjun: is "simulations" compile differently? I can't figure out why it's failing.

jagannatharjun: hmm it seems something from libsimulation end, conversion between std::system_clock to simulated_clock

arvidn: I believe it's coming from here: https://github.com/arvidn/libsimulator/blob/master/include/simulator/simulator.hpp#L155

src/time.cpp
```diff
@@ -32,9 +32,33 @@ POSSIBILITY OF SUCH DAMAGE.
...
+		return system_clock::to_time_t(r);
+	}
+
+	time_point32 from_time_t(const std::time_t t)
+	{
+		const auto tp = system_clock::from_time_t(t);
+		const auto hc_now = std::chrono::high_resolution_clock::now();
+		const auto sys_now = system_clock::now();
+
+		auto r = hc_now + std::chrono::duration_cast<system_clock::duration>(sys_now - tp);
```
arvidn, line 60: shouldn't this `duration_cast<>` be casting to `lt::clock_type::duration`?

src/time.cpp
```diff
@@ -32,9 +32,33 @@ POSSIBILITY OF SUCH DAMAGE.
...
+		const auto sys_now = system_clock::now();
+
+		const auto r = sys_now + std::chrono::duration_cast<system_clock::duration>(hc_now - tp);
+		return system_clock::to_time_t(r);
+	}
+
+	time_point32 from_time_t(const std::time_t t)
+	{
+		const auto tp = system_clock::from_time_t(t);
+		const auto hc_now = std::chrono::high_resolution_clock::now();
```
arvidn, line 57: I believe this should use `lt::clock_type::now()`, right? To preserve the abstraction of `lt::clock_type`. It may be changed to a `monotonic_clock` in the future for instance

src/time.cpp
```diff
@@ -32,9 +32,33 @@ POSSIBILITY OF SUCH DAMAGE.
...
+
 namespace libtorrent { namespace aux {
 
 	time_point time_now() { return clock_type::now(); }
 	time_point32 time_now32() { return time_point_cast<seconds32>(clock_type::now()); }
 
+	// for simplying implementation
+	using std::chrono::system_clock;
+
+	time_t to_time_t(const time_point32 tp)
```
arvidn, line 45: would you mind adding a comment suggesting that in C++20 this can be replaced by `std::chrono::clock_cast`? (which I believe is the case)

arvidn:
looks like the only thing remaining is to fix up existing tests

jagannatharjun:
@arvidn do you need anything from my end?

arvidn:
@jagannatharjun would you mind taking a look at the tests that are failing because of this change and understand why, and possibly fixup the tests?

jagannatharjun:
@arvidn this will break reading last_upload and last_download from existing resume data, this also warrents changes in various tests. Do you know any other place I should look into that may require change?

src/torrent.cpp
```diff
@@ -11675,11 +11675,14 @@ namespace {
 		st->seeding_time = int(total_seconds(seeding_time()));
 
 		time_point32 const unset{seconds32(0)};
+		auto const last_upload_unix = aux::to_time_t(m_last_upload),
+			last_download_unix = aux::to_time_t(m_last_download);
```
arvidn, line 11679: I prefer declaring variables one at at time.

src/torrent.cpp
```diff
@@ -11675,11 +11675,14 @@ namespace {
...
 
 		st->time_since_upload = m_last_upload == unset ? -1
-			: static_cast<int>(total_seconds(aux::time_now32() - m_last_upload));
+			: static_cast<int>(std::time(nullptr) - last_upload_unix);
 		st->time_since_download = m_last_download == unset ? -1
-			: static_cast<int>(total_seconds(aux::time_now32() - m_last_download));
+			: static_cast<int>(std::time(nullptr) - last_download_unix);
```
arvidn, line 11684: This conversion seems unnecessary (and risky). The value we're computing is a time duration, there's no need to mix in posix time here.

test/test_resume.cpp
```diff
@@ -207,7 +207,8 @@ void default_tests(torrent_status const& s, lt::time_point const time_now)
 	TEST_CHECK(s.active_time >= 1339);
 	TEST_CHECK(s.active_time < 1339 + 10);
 
-	auto const now = duration_cast<seconds>(time_now.time_since_epoch()).count();
+	auto const now = time(NULL);
```
arvidn, line 210: this doesn't look right. `now` is only used to compute relative time durations, why would this make a difference?

arvidn (commented):
I believe you accidentally updated the libsimulator and try_signal submodules

arvidn:
also: https://github.com/arvidn/libtorrent/actions/runs/4215330112/jobs/7316388314#step:5:459

src/time.cpp
```diff
@@ -32,9 +32,36 @@ POSSIBILITY OF SUCH DAMAGE.
...
+	using std::chrono::system_clock;
+
+
+	// consider using std::chrono::clock_cast on C++20
+	time_t to_time_t(const time_point32 tp)
+	{
+		const auto lt_now = clock_type::now();
+		const auto sys_now = system_clock::now();
+
+		const auto r = sys_now + std::chrono::duration_cast<system_clock::duration>(lt_now - tp);
```
arvidn, line 52: this should be `tp - lt_now`. this difference has the same bug on line 63

arvidn:
@jagannatharjun I fixed up this patch and posted a new PR here: https://github.com/arvidn/libtorrent/pull/7328
please take a look

jagannatharjun:
looks good thanks 👍

PR #7304, opened 2023-02-12
simplify handling of privileged ports setting

arvidn:
by untangling it from the port filter

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -11246,9 +11246,31 @@ namespace {
 		peers_erased(st.erased);
 	}
 
+	// this call will disconnect any peers whose remote port is < 1024
+	void torrent::privileged_port_updated()
+	{
+		if (!m_peer_list) return;
+
+		port_filter ports;
+		ports.add_rule(0, 1024, port_filter::blocked);
+
+		torrent_state st = get_peer_list_state();
+		std::vector<address> banned;
+		m_peer_list->apply_port_filter(ports, &st, banned);
+
+		if (alerts().should_post<peer_blocked_alert>())
+		{
+			for (auto const& addr : banned)
+				alerts().emplace_alert<peer_blocked_alert>(get_handle()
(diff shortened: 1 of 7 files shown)


src/torrent.cpp
```diff
@@ -11246,9 +11246,31 @@ namespace {
 		peers_erased(st.erased);
 	}
 
+	// this call will disconnect any peers whose remote port is < 1024
+	void torrent::privileged_port_updated()
+	{
+		if (!m_peer_list) return;
+
+		port_filter ports;
+		ports.add_rule(0, 1024, port_filter::blocked);
```
HanabishiRecca, line 11255: Shouldn't it be `0, 1023`? Privileged ports are **below** 1024.

src/torrent.cpp
```diff
@@ -11246,9 +11246,31 @@ namespace {
...
+
+		torrent_state st = get_peer_list_state();
+		std::vector<address> banned;
+		m_peer_list->apply_port_filter(ports, &st, banned);
+
+		if (alerts().should_post<peer_blocked_alert>())
+		{
+			for (auto const& addr : banned)
+				alerts().emplace_alert<peer_blocked_alert>(get_handle()
+					, tcp::endpoint(addr, 0)
```
HanabishiRecca, line 11265: Can we preserve the port number for the alert?
I am actually pushing https://github.com/qbittorrent/qBittorrent/pull/18544 atm.

arvidn: yeah, that makes sense

HanabishiRecca: Thanks.

arvidn:
@HanabishiRecca does this patch fix the issue you experienced?

HanabishiRecca:
Yes.

image

Will it be backported to 1.x?

HanabishiRecca:
I have a tangential question: https://github.com/arvidn/libtorrent/issues/7315

Issue #7285, opened 2023-01-24
I2P Proxy settings


glassez:
Did I understand correctly that for an I2P proxy, not the usual settings (`proxy_hostname` and `proxy_port` are used, but separate ones: `i2p_hostname` and `i2p_port`? If so, why?

jiigen:
> Did I understand correctly that for an I2P proxy, not the usual settings (`proxy_hostname` and `proxy_port` are used, but separate ones: `i2p_hostname` and `i2p_port`? If so, why?

I don't know exaclty why, probacly cuz internally they are used differently by libtorrent. Maybe @Vort or @arvidn can have the answer.

Vort:
> probably cuz internally they are used differently by libtorrent

This is most likely the reason.
SAM protocol is similar to proxy, but unlike proxies, it gives access to different network, not just allows different route to the same network.
Let's wait for @arvidn answer.

arvidn:
I don't remember why the i2p hostname and port ended up as a separate setting. There are some remnants of a proxy type for i2p, and I think it may have worked that way at some point in the past. it seems to need some straightening out though.

maybe setting proxy type to i2p (and using proxy_hostname and proxy_port) could be made synonymous with i2p_hostname and i2p_port.

glassez:
The presence of these alternative settings (`i2p_hostname` and `i2p_port`) would be necessary only if they could be used simultaneously/in parallel with `proxy_hostname` and `proxy_port`. Otherwise (if i2p can only act as another type of proxy), they are confusing.

arvidn:
yes

zzzi2p:
Well, libtorrent and the downstreams need to decide whether you eventually want to support "mixed" networks or not, and if so what's the default.

Mixed = both I2P and clearnet at the same time, "bridging" clearnet torrents into I2P. This has obvious anonymity issues but BiglyBT supports it and lots of people run it that way. There's complexity with bandwidth and peer management if you do it that way because of the vastly different speeds.

Whether you support mixed or not, you have to make sure you're not trying to pump clearnet traffic or announce to clearnet trackers or do uTP through SAM, that's why it's not a real proxy and perhaps why it's a different set of configs. So think through where you're headed.

Speaking of, defaulting to localhost/7656 would help your users, as that's where the i2p router will be 99.9% of the time.

Vort:
I just want to note that right now `settings_pack::allow_i2p_mixed` already exists in code.
[...]

PR #7305, opened 2023-02-12
improve documentation clarity around i2p settings

arvidn:
Slightly simplify session_interface

### DIFF
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -729,7 +729,6 @@ namespace aux {
 
 #if TORRENT_USE_I2P
 			char const* i2p_session() const override { return m_i2p_conn.session_id(); }
-			proxy_settings i2p_proxy() const override;
 
 			void on_i2p_open(error_code const& ec);
 			void open_new_incoming_i2p_connection();
--- a/include/libtorrent/aux_/session_interface.hpp
+++ b/include/libtorrent/aux_/session_interface.hpp
@@ -213,7 +213,6 @@ namespace aux {
 		virtual proxy_settings proxy() const = 0;
 
 #if TORRENT_USE_I2P
-		virtual proxy_settings i2p_proxy() const = 0;
 		virtual char const* i2p_session() const = 0;
 #endif
 
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -363,7 +363,10 @@ namespace aux {
 			proxy_password,
 
 			// sets the i2p_ SAM bridge to connect to. set the port with the
-			// ``i2p_port`` setting.
+			// ``i2p_port`` setting. Unless this is set, i2p torrents are not
+			// supported. This setting is separate from the other proxy settings
+			// since i2p torrents and their peers are orthogonal. You can have
+			// i2p peers as well as regular peers via a proxy.
 			//
 			// .. _i2p: http://www.i2p2.de
 			i2p_hostname,
@@ -2206,8 +2209,13 @@ namespace aux {
 			// authorization. The username and password will be sent to the proxy.
 			http_pw,
 
-			// route through a i2p SAM proxy
+#if TORRENT_USE_I2P
+			// internal
+			// This is used internally to communicate with the
+			// http_tracker_connection. To configure an i2p SAM bridge, set
+			// i2p_hostname and i2p_port.
 			i2p_proxy
+#endif
 		};
 	private:
 
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -2354,17 +2354,6 @@ namespace {
 #endif
 
 #if TORRENT_USE_I2P
-
-	proxy_settings session_impl::i2p_proxy() const
-	{
-		proxy_settings ret;
-
-		ret.hostname = m_settings.get_str(settings_pack::i2p_hostname);
-		ret.type = settings_pack::i2p_proxy;
-		ret.port = std::uint16_t(m_settings.get_int(settings_pack::i2p_port));
-		return ret;
-	}
-
 	void session_impl::on_i2p_open(error_code const& ec)
 	{
 		if (ec)
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -7350,7 +7350,7 @@ namespace {
 #if TORRENT_USE_I2P
 		if (peerinfo->is_i2p_addr)
 		{
-			if (m_ses.i2p_proxy().hostname.empty())
+			if (settings().get_str(settings_pack::i2p_hostname).empty())
 			{
 				// we have an i2p torrent, but we're not connected to an i2p
 				// SAM proxy.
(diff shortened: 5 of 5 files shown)


Vort:
Not sure if this information is useful, but I made checks with client_test and I2P download still works.

jiigen:
> Not sure if this information is useful, but I made checks with client_test and I2P download still works.

Did you modify the code of test cliebt or left it unchanged? If changed can you post the modified code? Thanks

Vort:
> Did you modify the code of test client or left it unchanged?

I did not changed it.
Just launched with such `run.bat` file: `client_test -f log.txt --alert_mask=all --enable_dht=false --proxy_type=i2p_proxy --i2p_hostname=127.0.0.1 --i2p_port=7656 magnet:?xt=urn:btih:cb18974db461c640c0d938710203bb263a746694^&dn=I2P%%2B+Universal+Installer+2.0.0%%2B^&tr=http://tracker2.postman.i2p/announce.php`

jiigen:
@Vort uhm..so you still used the option "--proxy_type=i2p_proxy" that confuses me cuz according to https://github.com/arvidn/libtorrent/issues/7285#issuecomment-1427065452 which has been confirmed by @arvidn here https://github.com/arvidn/libtorrent/issues/7285#issuecomment-1427106205 it should work even without that setting...Am I correnct or I didn't understand??

jiigen:
@arvidn as suggestion wouldn't be useful to add an "i2p_simple_client" among the examples to test about the different i2p options? (disabled,alone,mixed...)

Vort:
> it should work even without that setting

I tested and it works without `--proxy_type=i2p_proxy`.
However, without `proxy_type`, `client_test` opens lots of TCP and UDP listeners and I'm not sure if it is correct way of interaction with I2P network.
Maybe I should add some different option to prevent it from happening? Why so many listeners are needed when connections should go only via I2P?

jiigen:
I dont see those listeners...anyway I must look more carefully.

jiigen:
However so far the client which connect faster and grabs more peers is Deluge, I can even see the peers list after my little patch: 
https://github.com/deluge-torrent/deluge/pull/412/commits/cae645470de7ab9c43f345bcfe50b449b37ee8b3
Would be able to modify the code to try without i2p_proxy setting and mixed mode but failed so far...

jiigen:
@Vort Using this plugin for deluge: https://github.com/ratanakvlun/deluge-ltconfig and manually setting i2p_hostname and i2p_port makes deluge work with i2p leaving proxy setting to none. 
I got a torrent from postman tracker which was even on clearnet trackers added those trackers to the tracker list and it connected to both i2p and cleanet peers.
Oddly if I don't disable dht and pex it doesn't connect to clearnet peers even if it gets them from the trackers.
Screenshot at 2023-03-04 09-54-19
Screenshot at 2023-03-04 10-00-14

Vort:
> I got a torrent from postman tracker which was even on clearnet trackers added those trackers to the tracker list and it connected to both i2p and cleanet peers.

Shouldn't it work this way only when `allow_i2p_mixed` is set?

jiigen:
> 
> Shouldn't it work this way only when `allow_i2p_mixed` is set?

Oh yes...sorry forgot to tell that with that same plugin I've enabled i2p_mixed too.

jiigen:
From the libtorrent reference:
```
allow_i2p_mixed
 "if this is true, i2p torrents are allowed to also get peers from other sources than the tracker, 
 and connect to regular IPs, not providing any anonymization."
```
So I'm wondering, if peers comes from a clearnet tracker they are used regardless of this setting?

arvidn:
@Vort 
> Maybe I should add some different option to prevent it from happening? Why so many listeners are needed when connections should go only via I2P?

This is because the libtorrent session supports adding multiple torrents, and opening listen sockets is preparing for supporting non-i2p torrents. The i2p torrents aren't supposed to accept connections via them.

@jiigen 
> So I'm wondering, if peers comes from a clearnet tracker they are used regardless of this setting?

My recollection is that traditional trackers shouldn't be used at all in i2p mode (unless mixed mode is enabled). But if they are, that might be a bug.

Vort:
> opening listen sockets is preparing for supporting non-i2p torrents. The i2p torrents aren't supposed to accept connections via them

Do it means that `allow_i2p_mixed` have no effect on clearnet-only torrents?
It makes some sense, however, I think with such approach it is too easy to leak i2p torrent to clearnet in case of user mistake.

arvidn:
I'm not sure I understand what the (potentially) problematic scenario is. You mean if someone accidntally sticks a normal (public) tracker into an i2p torrent?

Vort:
> what the (potentially) problematic scenario is

I may be wrong, but I think that user can accidentally delete i2p tracker from torrent, making it lose its i2p status and consequently telling whole world about BTIH -> IP relation.

jiigen:
> @jiigen
> 
> > So I'm wondering, if peers comes from a clearnet tracker they are used regardless of this setting?
> 
> My recollection is that traditional trackers shouldn't be used at all in i2p mode (unless mixed mode is enabled). But if they are, that might be a bug.

Do you mean a bug in the client (Deluge in thiscase)?

jiigen:
> > what the (potentially) problematic scenario is
> 
> I may be wrong, but I think that user can accidentally delete i2p tracker from torrent, making it lose its i2p status and consequently telling whole world about BTIH -> IP relation.

So there's an "i2p status" ? and how it is recognized? just if it has an i2p tracker? From what I've seen so far:
- if i2p is used as proxy only i2p peers/tracker can be reached.
- if I don't use the proxy option but only set the host/port of i2p router regardless of mixed mode if one adds a clearnet tracker it is contacted and peers retrived.
So probably only DHT and PEX needs mixed mode? IDK

Vort:
> This is problematic for i2p magnet links.

It may be also problematic when user want to download clearnet torrent and change tracker to i2p one with torrent client. Ideally this should convert torrent to "i2p mode" and block clearnet connections unless user enables mixed mode.

Another similar question is how DHT behaves when i2p torrent is added. It should not leak IP<->BTIH connection unless mixed mode is enabled. Ideally, it also should use i2p DHT modification in cases of i2p torrents, however it is not urgent, most important is to prevent leaks.

glassez:
What if some torrent contains both regular and i2p trackers?

arvidn:
then it's considered an i2p torrent, and will not allow regular peers unless `allow_i2p_mixed` is true (but it defaults to false).

Currently, it may still announce to the plain trackers, it just won't connect to any of the peers. I'm fixing that to not even announce to non-i2p trackers when in i2p mode in https://github.com/arvidn/libtorrent/pull/7362

glassez:
> Currently, it may still announce to the plain trackers, it just won't connect to any of the peers. I'm fixing that to not even announce to non-i2p trackers when in i2p mode in #7362

:+1:


> then it's considered an i2p torrent, and will not allow regular peers unless `allow_i2p_mixed` is true (but it defaults to false).

Does adding/removing trackers behave consistently? I.e.:
1. Adding i2p tracker. I expect that it should turn torrent to i2p mode and disconnect it from regular peers (stop announcing to regular trackers) in case of disabled `allow_i2p_mixed`
2. Removing last i2p tracker. I expect that it should turn torrent to non-i2p mode 

What if non-i2p torrent received i2p peers from regular tracker (is it possible?) or from user (manually added)?

I think if you really want to define a "strict i2p torrent" model, then all such questions should be answered so that it is consistent.

arvidn:
> Does adding/removing trackers behave consistently? I.e.:
>
> 1. Adding i2p tracker. I expect that it should turn torrent to i2p mode and disconnect it from regular peers (stop announcing to regular trackers) in case of disabled allow_i2p_mixed
> 2. Removing last i2p tracker. I expect that it should turn torrent to non-i2p mode

Currently, a torrent is only considered an i2p-torrent if you add it as a .torrent file, and it contains a tracker with an i2p URL. That's it.

With https://github.com/arvidn/libtorrent/pull/7362 the situation is improved. There's torrent flag added to indicate whether a torrent should be treated as an i2p torrent or not. It's set automatically when loading a .torrent file with an i2p tracker, or parsing a magnet URI including an i2p tracker. It can then be altered explicitly by the client, via the new torrent flag.

> What if non-i2p torrent received i2p peers from regular tracker (is it possible?) or from user (manually added)?

That's supposed to work, as long as the client has configured a SAM connection. The restriction introduced by `allow_i2p_mixed` only affects i2p torrents, not regular torrents.

I believe the documentation in #7362 is consistent with this description

PR #7314, opened 2023-02-13
fix build error on Windows ARM64 by add new define

wbsdty331:
support arm64 windows build.
image

### DIFF
--- a/src/assert.cpp
+++ b/src/assert.cpp
@@ -190,16 +190,21 @@ TORRENT_EXPORT void print_backtrace(char* out, int len, int max_depth
 	std::array<void*, 50> stack;
 
 	STACKFRAME64 stack_frame = {};
-#if defined(_WIN64)
-	int const machine_type = IMAGE_FILE_MACHINE_AMD64;
-	stack_frame.AddrPC.Offset = context_record.Rip;
-	stack_frame.AddrFrame.Offset = context_record.Rbp;
-	stack_frame.AddrStack.Offset = context_record.Rsp;
-#else
+#if defined(_M_IX86)
 	int const machine_type = IMAGE_FILE_MACHINE_I386;
 	stack_frame.AddrPC.Offset = context_record.Eip;
 	stack_frame.AddrFrame.Offset = context_record.Ebp;
 	stack_frame.AddrStack.Offset = context_record.Esp;
+#elif defined(_M_X64)
+	int const machine_type = IMAGE_FILE_MACHINE_AMD64;
+	stack_frame.AddrPC.Offset = context_record.Rip;
+	stack_frame.AddrFrame.Offset = context_record.Rbp;
+	stack_frame.AddrStack.Offset = context_record.Rsp;
+#elif defined(_M_ARM64)
+	int const machine_type = IMAGE_FILE_MACHINE_ARM64;
+	stack_frame.AddrPC.Offset = context_record.Pc;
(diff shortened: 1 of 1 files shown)


src/assert.cpp
```diff
@@ -190,16 +190,21 @@ TORRENT_EXPORT void print_backtrace(char* out, int len, int max_depth
...
-	stack_frame.AddrPC.Offset = context_record.Rip;
-	stack_frame.AddrFrame.Offset = context_record.Rbp;
-	stack_frame.AddrStack.Offset = context_record.Rsp;
-#else
+#ifdef _M_IX86
 	int const machine_type = IMAGE_FILE_MACHINE_I386;
 	stack_frame.AddrPC.Offset = context_record.Eip;
 	stack_frame.AddrFrame.Offset = context_record.Ebp;
 	stack_frame.AddrStack.Offset = context_record.Esp;
+#elif _M_X64
```
glassez, line 198: Shouldn't it be `#elif defined(_M_X64)`? Etc.

arvidn:
@wbsdty331 could you please provide some more details of what this patch does, why it's necessary and any other compelling arguments to land this patch.

Am I right in assuming that this only fixes debug builds, and that release builds already work for Windows ARM64?

Do you know if there's a way to build for Windows ARM64 on github actions? Without CI coverage, there's a good chance support will break again in the future.

wbsdty331:
@arvidn 
> could you please provide some more details of what this patch does, why it's necessary and any other compelling arguments to land this patch.

let me see the original code:
```cpp
#if defined(_WIN64)
	int const machine_type = IMAGE_FILE_MACHINE_AMD64;
	stack_frame.AddrPC.Offset = context_record.Rip;
	stack_frame.AddrFrame.Offset = context_record.Rbp;
	stack_frame.AddrStack.Offset = context_record.Rsp;
#else
	int const machine_type = IMAGE_FILE_MACHINE_I386;
	stack_frame.AddrPC.Offset = context_record.Eip;
	stack_frame.AddrFrame.Offset = context_record.Ebp;
	stack_frame.AddrStack.Offset = context_record.Esp;
#endif
```

from MSDN [Predefined preprocessor macros](https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170) we can know:

> _WIN64 Defined as 1 when the compilation target is 64-bit ARM or x64. Otherwise, undefined.

original code only focused on i386 and x86_64 platform, but stackframe have dependencies on different CPU registers (Rip rbp Ebp and so on),

It means arm64 architectures in this define is also true. for x86_64 cpu, the code is ok, but arm64 processor don't have `Rip` `Rbp` and `Rsp` registers, and `MACHINE_TYPE` is also wrong, so you must use arm64 CPU register and MACHINE_TYPE to use stackframe.

ref: https://learn.microsoft.com/en-us/cpp/build/arm64-windows-abi-conventions?view=msvc-170


> Am I right in assuming that this only fixes debug builds, and that release builds already work for Windows ARM64?

no, both debug and releases.

image


> Do you know if there's a way to build for Windows ARM64 on github actions? Without CI coverage, there's a good chance support will break again in the future.

You can use MSVC ARM64 toolchina to cross compile, reference: https://github.com/minnyres/qbittorrent-windows-arm64/blob/main/.github/workflows/ci_windows_arm64.yaml

arvidn:
sorry for the delay. thanks for the contribution!

PR #7339, opened 2023-03-05
instead of removing web seed with certain errors

arvidn:
just disable them and preserve them in the resume data. e.g. a web seed that redirects to other URLs are now preserved while the redirect targets are not


glassez:
BTW, how is it supposed to behave in case the user manually adds new web seed while there is some ephemeral web seed with the same address?

arvidn:
> BTW, how is it supposed to behave in case the user manually adds new web seed while there is some ephemeral web seed with the same address?

It should probably make it non-ephemeral

glassez:
> > BTW, how is it supposed to behave in case the user manually adds new web seed while there is some ephemeral web seed with the same address?
> 
> It should probably make it non-ephemeral

At least, I would expect it to behave that way. That's why I was interested in how it really is.

PR #7341, opened 2023-03-11
Improve save resume if modified

arvidn:
The main cases this fixes are things like:
* saving resume data whenever a torrent finishes downloading. This is a reasonable thing to do, but is not reasonable if you just started and added an existing torrent (just loading its resume data) and it was finished. You'll still get this alert to indicate that it's finished. saving with `only_if_modified` still triggers saving in this case, but not with this patch.
* saving resume data on `add_torrent_alert`, since maybe a new torrent was just added and it should be saved in the resume data so it will be restored in the next session. However, it the torrent was added from existing resume data, and we just started up a new session, it does not make sense to re-save the resume data. Saving with `only_if_modified` will trigger saving currently, but not with this patch.

Other minor improvements are:
* if you set per-torrent upload rate, download rate, connections limit, upload slots limit to the *same* value as was already set, it's not considered a "modification" for need-save-resume data purposes.
* if a web seed redirects and adds a new ephemeral web seed, it's not considered a modification for need-save-resume data purposes (since ephemeral web seeds aren't saved in the resume data anyway)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* improve save_resume behavior with only_if_modified flag
 	* fix issue with retrying trackers in tiers > 0
 	* fix last_upload and last_download resume data fields to use posix time
 	* improve error messages for no_connect_privileged_ports, by untangle it from the port filter
--- a/examples/client_test.cpp
+++ b/examples/client_test.cpp
@@ -1111,7 +1111,7 @@ bool handle_alert(client_state_t& client_state, lt::alert* a)
 		// the alert handler for save_resume_data_alert
 		// will save it to disk
 		torrent_handle h = p->handle;
-		h.save_resume_data(torrent_handle::save_info_dict);
+		h.save_resume_data(torrent_handle::save_info_dict | torrent_handle::only_if_modified);
 		++num_outstanding_resume_data;
 		if (exit_on_finish) quit = true;
 	}
--- a/examples/client_test.cpp
+++ b/examples/client_test.cpp
@@ -2148,7 +2148,7 @@ done:
 				std::vector<lt::open_file_state> file_status = h.file_status();
 				std::vector<lt::download_priority_t> file_prio = h.get_file_priorities();
 				auto f = file_status.begin();
-				std::shared_ptr<const lt::torrent_info> ti = h.torrent_file();
+				std::shared_ptr<const lt::torrent_info> ti = s.torrent_file.lock();
 
 				auto const& file_progress = client_state.file_progress;
 				int p = 0; // this is horizontal position
--- a/include/libtorrent/torrent.hpp
+++ b/include/libtorrent/torrent.hpp
@@ -873,7 +873,7 @@ namespace libtorrent {
 
 		// called when we learn that we have a piece
 		// only once per piece
-		void we_have(piece_index_t index);
+		void we_have(piece_index_t index, bool loading_resume = false);
 
 		// process the v2 block hashes for a piece
 		boost::tribool on_blocks_hashed(piece_index_t piece
@@ -1670,7 +1670,11 @@ namespace libtorrent {
 		// the number of unchoked peers in this torrent
 		unsigned int m_num_uploads:24;
 
-		// 4 unused bits
+		// 3 unused bits
+
+		// set to false when saving resume data. Set to true
+		// whenever some statistics that's saved in the resume data is updated
+		bool m_counters_updated:1;
 
 		// when this is true, this torrent supports peer exchange
 		bool m_enable_pex:1;
--- a/include/libtorrent/torrent_handle.hpp
+++ b/include/libtorrent/torrent_handle.hpp
@@ -683,10 +683,20 @@ namespace aux {
 		// if nothing significant has changed in the torrent since the last
 		// time resume data was saved, fail this attempt. Significant changes
 		// primarily include more data having been downloaded, file or piece
-		// priorities having changed etc. If the resume data doesn't need
-		// saving, a save_resume_data_failed_alert is posted with the error
+		// priorities, paused-state, transfer limit settings  having changed etc.
+		// Counters keeping statistics are notably *not* considered significant.
+		// Things like total_uploaded, total_downloaded, last_seen_complete,
+		// tracker scrape counters, are not considered significant enough
+		// changes for this flag to save the resume data.
+		// If the resume data doesn't need saving, a
+		// save_resume_data_failed_alert is posted with the error
 		// resume_data_not_modified.
 		static constexpr resume_data_flags_t only_if_modified = 2_bit;
+
+		// this flag amends the behavior of only_if_modified to include
+		// stats counter or scrape data updating to also count as an update, so
+		// it becomes more eager to save resume data.
+		static constexpr resume_data_flags_t save_counters = 3_bit;
 
 		// ``save_resume_data()`` asks libtorrent to generate fast-resume data for
 		// this torrent. The fast resume data (stored in an add_torrent_params
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -224,6 +224,7 @@ bool is_downloading_state(int const st)
 		, m_enable_lsd(!bool(p.flags & torrent_flags::disable_lsd))
 		, m_max_uploads((1 << 24) - 1)
 		, m_num_uploads(0)
+		, m_counters_updated(false)
 		, m_enable_pex(!bool(p.flags & torrent_flags::disable_pex))
 		, m_apply_ip_filter(p.flags & torrent_flags::apply_ip_filter)
 		, m_pending_active_change(false)
@@ -2170,7 +2171,7 @@ bool is_downloading_state(int const st)
 					m_picker->we_have(i);
 					inc_stats_counter(counters::num_piece_passed);
 					update_gauge();
-					we_have(i);
+					we_have(i, true);
 				}
 
 				// --- UNFINISHED PIECES ---
@@ -3363,7 +3364,7 @@ namespace {
 			update_auto_sequential();
 
 			// these numbers are cached in the resume data
-			set_need_save_resume();
+			m_counters_updated = true;
 		}
 	}
 
@@ -4056,7 +4057,7 @@ namespace {
 	// piece, it just does all the torrent-level accounting that needs to
 	// happen. It may not be called twice for the same piece (if it is,
 	// file_progress will assert)
-	void torrent::we_have(piece_index_t const index)
+	void torrent::we_have(piece_index_t const index, bool const loading_resume)
 	{
 		TORRENT_ASSERT(is_single_thread());
 		TORRENT_ASSERT(!has_picker() || m_picker->has_piece_passed(index));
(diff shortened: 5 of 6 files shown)


glassez:
>improve behavior of save_resume_data() only_if_modified flag to avoid re-saving the resume data too frequently

At first glance, it definitely does not look like an improvement. Isn't it the client-side concern to take care of the save resume data frequency?
With these changes in mind, how do you propose to request data retention if at least some part has been changed (insignificant according to this PR)?

arvidn:
it's an "improvement" in the sense that `only_if_modified` becomes useful. The statistics are essentially updated continuously, so all active torrents are essentially considered "modified" currently.

The resume data saving frequency should mostly not be affected. Saving resume data on a timer is likely not passing in `only_if_modified`. But if it does, there's a new flag `save_counters`, which will make counter updates considered cause for saving the resume data as well.

The main cases this fixes are things like:
* saving resume data whenever a torrent finishes downloading. This is a reasonable thing to do, but is not reasonable if you just started and added an existing torrent (just loading its resume data) and it was finished. You'll still get this alert to indicate that it's finished. saving with `only_if_modified` still triggers saving in this case, but not with this patch.
* saving resume data on `add_torrent_alert`, since maybe a new torrent was just added and it should be saved in the resume data so it will be restored in the next session. However, it the torrent was added from existing resume data, and we just started up a new session, it does not make sense to re-save the resume data. Saving with `only_if_modified` will trigger saving currently, but not with this patch.

Other minor improvements are:
* if you set per-torrent upload rate, download rate, connections limit, upload slots limit to the *same* value as was already set, it's not considered a "modification" for need-save-resume data purposes.
* if a web seed redirects and adds a new ephemeral web seed, it's not considered a modification for need-save-resume data purposes (since ephemeral web seeds aren't saved in the resume data anyway)

glassez:
> it's an "improvement" in the sense that `only_if_modified` becomes useful.

It was `useful` before! Its main use, IMO, was to avoid a blocking call to `torrent_handle::need_save_resume()`. And, by the way, it was consistent with each other. Now it seems that there is no analogue to the previous methods behavior (`torrent_handle::need_save_resume()` and appropriate field of `torrent_status`) at all.
In any case, such a behavior change, even if it is justified, is highly undesirable to apply by default in an intermediate release.

glassez:
Another important note is that all those resume data changes that come from the user side can be tracked there, i.e. if the user (effectively, the application) has added a web seed, or changed some limits, then it is aware that the resume data has been changed, so it can not to find out about it from libtorrent, but simply to request the data to be saved unconditionally.
But the changes that occur in libtorrent itself during the torrent operation (counters, progress, etc.) can only be asked from libtorrent itself. Therefore, hindering the existing possibilities of obtaining such information looks very unfriendly. This is a movement in the opposite direction from what I would expect in the development of libtorrent.

arvidn:
> It was useful before! Its main use, IMO, was to avoid a blocking call to torrent_handle::need_save_resume().

It still does. It's just that `need_save_resume()` had its meaning slightly changed.

> Now it seems that there is no analogue to the previous methods behavior (torrent_handle::need_save_resume() and appropriate field of torrent_status) at all.

There is. instead of `need_save_resume()` you could call `!is_paused()`. This is because the counters are updated continuously for non-paused torrents. And instead of calling `save_resume_data(torrent_handle::only_if_changed)` you can call `save_resume_data(torrent_handle::only_if_changed | torrent_handle::save_counters)`. However, if you find yourself wanting to do that, you're likely to really just want to call `save_resume_data()` (with no flags), to save resume data unconditionally.

> In any case, such a behavior change, even if it is justified, is highly undesirable to apply by default in an intermediate release.

I agree that ideally there wouldn't be subtle changes in semantics like this (at all, let alone in point-releases). However, the intention of `only_if_modified` was always to prevent superfluous saving of resume data, and it was documented as such (but not with great precision). I'm open to be convinced that I should restore the default behavior and require the new flag for the new behavior. I just don't think the existing behavior of `only_if_modified` is very useful, since it's essentially the same as not passing any flag at all (which still is the default, if nothing is specified).

> Another important note is that all those resume data changes that come from the user side can be tracked there, i.e. if the user (effectively, the application) has added a web seed, or changed some limits, then it is aware that the resume data has been changed, so it can not to find out about it from libtorrent, but simply to request the data to be saved unconditionally.

I think the most important state is blocks and pieces being downloaded. That was the original intention. Since the current behavior also considers state changes like paused, seed-mode, force-started, web seeds and trackers as need-save-resume data updates, I think it's reasonable to leave those there, even though (as you say) they don't necessarily provide all that much value, but some convenience.

> But the changes that occur in libtorrent itself during the torrent operation (counters, progress, etc.) can only be asked from libtorrent itself. 

You know that at least some counters change as long as the torrent is not paused. e.g. the counter of the number of seconds the torrent has been active for. The difficult part is really download progress. This change moves `only_if_modified` in that direction.

>  Therefore, hindering the existing possibilities of obtaining such information looks very unfriendly. This is a movement in the opposite direction from what I would expect in the development of libtorrent.

I think this is a misunderstanding. The counters updating is always true for a non-paused torrent. Therefore, including this in the need-save-resume makes it more-or-less meaningless, and not helpful.

glassez:
> And instead of calling `save_resume_data(torrent_handle::only_if_changed)` you can call `save_resume_data(torrent_handle::only_if_changed | torrent_handle::save_counters)`.

Even if it is a really good idea to separate them, I would rather expect it to look different, i.e. `only_if_changed` would remain as accurate as it was before, and an additional flag would change it by adding a new behavior (e.g. call it as `save_resume_data(torrent_handle::only_if_changed | torrent_handle::skip_counters)` for new behavior). And it would also be much better in terms of compatibility with existing applications.

glassez:
> There is. instead of `need_save_resume()` you could call `!is_paused()`. This is because the counters are updated continuously for non-paused torrents.

It looks like a workaround. Also standalone call to `is_paused()` means not the same thing as standalone call to `need_save_resume()` before this PR, isn't it? I mean if `is_paused()` returns `false` then we can treat it as `need_save_resume is true` but when it returns `true` it is undefined whether we need save resume data or not. Still, why not keeping its accurate meaning as before and just add some modifier if you need to alter it, e.g. `need_save_resume(torrent_handle::skip_counters)`

arvidn:
> Also standalone call to is_paused() means not the same thing as standalone call to need_save_resume() before this PR, isn't it?

If you negate `is_paused()` it will give you the same response as calling `need_save_resume()`. This is because the current meaning of need-save-resume makes it always true for any non-paused torrent.

> Still, why not keeping its accurate meaning as before and just add some modifier if you need to alter it, e.g. need_save_resume(torrent_handle::skip_counters)

calling `save_resume_data()` will save resume data unconditionally. Going out of ones way to pass in this flag to the call (i.e. calling `save_resume_data(torrent_handle::only_if_modified)`) suggests that the intention was to get different behavior. I have a hard time seeing the current behavior ever being desired. If it is, the flag should have a different name, something like `only_if_active`. Maybe that would be a better name than `save_counters` or `skip_counters`

glassez:
> If you negate `is_paused()` it will give you the same response as calling `need_save_resume()`. This is because the current meaning of need-save-resume makes it always true for any non-paused torrent.

But `need_save_resume()` wasn't always `false` for any paused torrents.

glassez:
> Since the current behavior also considers state changes like paused, seed-mode, force-started, web seeds and trackers as need-save-resume data updates, I think it's reasonable to leave those there, even though (as you say) they don't necessarily provide all that much value, but some convenience.

I don't ask you to exclude some changes from being affected to need_save_resume result. On the contrary, I would expect to see things simple and strict, i.e. `need_save_resume` should be true by default if **any** part of resume data is changed.

Note that when you provide some small, simple and understandable tools, while covering all the required functionality, the application can combine them as it likes to build some complex logic (even if in some cases it turns out to be cumbersome). But when you try to think for the developers of the application and provide them with some complex logic out of the box, then you risk making a mistake and instead of something useful, you only provide a pain in the ass.
(Of course, it is often really useful to have such complex utilities as part of the backend, because in some cases it is really easier, more convenient and more effective to implement it there.)

arvidn:
> As I said above its main benefit is that it avoids additional blocking calls to torrent_handle::need_save_resume().

It still does with this patch. But I should have been more precise; this patch alters the meaning of `need_save_resume()`.

> And it's not the same as not passing any flag at all. When I have hundreds or thousands of torrents and I want to save data only for those who really need it, it is really useful in any scenario, even if in some case all torrents would have changed data, it still saves me from hundreds/thousands of blocking calls.

Yes, I agree. The most important part here is defining "those who really need it". If you consider counters updating as "really needing" to be saved, you'll have a very close approximation to that behavior by not passing any flag. The only exception (afaict) are torrents that are paused (and have been paused since their resume data was saved last).

I would break down saving resume data into these main categories:

1. periodic saving of state just in case of a crash or other non-graceful termination
2. triggered by some event (becoming a seed, being added, receiving metadata etc.)
3. shutting down

In case (1), I don't think `only_if_modified` is necessary. Periodic saving is staggered and periodic anyway, it's unlikely to cause performance issues.
Likewise in case (3), unless a substantial number of torrents are paused, saving resume data for all torrents unconditionally is a pretty good approximation. I would also expect that, on shutdown, one would want to save resume data for every torrent unconditionally as a precaution of some other, edge-triggered, saving of it was missed.

The main case, that I experienced serious problems in, is (2). For example, when starting up and adding back the torrents from the previous session, every single one triggered another save resume data (triggered by both `add_torrent_alert` and `torrent_finished_alert`). Most of the torrents were seeding, so they would also immediately receive a `torrent_finished_alert`.
Why would I save resume data on those alerts? When adding a magnet link, it's really useful to save the metadata to disk as soon as it's received. Likewise, when a torrent becomes a seed, it's good to ensure that we save its state immediately, since it won't be downloading anymore. If we terminate, we'll at least be able to pick up all the pieces we downloaded.

> But need_save_resume() wasn't always false for any paused torrents.

I believe it only returns true if the state-change to paused hasn't been recorded in a resume data yet. There's another, rare, case where we might still be scraping the tracker and update scrape data, even for paused torrents.

> I don't ask you to exclude some changes from being affected to need_save_resume result. On the contrary, I would expect to see things simple and strict, i.e. need_save_resume should be true by default if any part of resume data is changed.

I agree that it would be ideal to have simple rules. However, consider a future where some new piece of information is included in the resume data. It's not necessarily in everyone's interest to save resume data more often because of this change. Certainly not for users who don't care about the new piece of information. This case is somewhat similar. Resume data was traditionally just the state needed to pick up the download where it was left off.

> As the creator of the backend library, you can't always make precise assumptions about what would be really more convenient and useful for applications using it. That's why we (as end-application developers) are trying to give you feedback to improve it.

Right, I'm very curious about your use case.
I imagine you're using `only_if_changed` in case (1) as well, as a slight performance improvement. But I don't know how you deal with case (2) without it being very expensive.
I believe you don't include the info-dict in your resume data, so that reduces cost of saving it, but I imagine it takes twice as many disk seeks to start up instead (loading both .torrent and resume data).

arvidn:
I'm experimenting with restoring the behavior of `only_if_modified` and moving towards separate flags for different kinds of changes. It might make sense to separate counters, download progress and state (paused, auto managed, trackers, web seeds, seed-mode etc.). That would make 3 flags that can be used to specify more precisely what kinds of changes are interesting.

That would also let me add more counters for more stats to the resume data without burdening clients that aren't interested in it.

glassez:
And I believe that the related aspects should also be adjusted accordingly, i.e. `torrent_handle::need_save_resume()` takes the same flags, and `torrent_status::need_save_resume` is supplemented with more detailed indicators like `need_save_counters`, `need_save_progress`, etc.

arvidn:
https://github.com/arvidn/libtorrent/pull/7343

PR #7343, opened 2023-03-16
Save resume data tweaks

arvidn:


include/libtorrent/torrent_handle.hpp
```diff
@@ -680,23 +680,51 @@ namespace aux {
...
+		// resume data was saved. This includes upload/download counters, active
+		// time counters and scrape data. A torrent that is not paused will have
+		// its active time counters incremented continuously.
+		static constexpr resume_data_flags_t if_counters_changed = 3_bit;
+
+		// save the resume data if any blocks have been downloaded since the
+		// last time resume data was saved. This includes:
+		// * checking existing files on disk
+		// * downloading a block from a peer
+		// * files renamed or moved
```
glassez, line 698: Does this refer to the download progress and not to the state?

include/libtorrent/torrent_handle.hpp
```diff
@@ -680,23 +680,51 @@ namespace aux {
...
+		// * upload mode
+		// * paused state
+		// * super-seeding
+		// * seed-mode
+		static constexpr resume_data_flags_t if_state_changed = 6_bit;
+
+		// save the resume data if any *metadata* changed since the last time
+		// resume data was saved. This includes:
+		// * add/remove web seeds
+		// * add/remove trackers
```
glassez, line 725: It looks confusing. I would only refer to "info section" as metadata. Otherwise, wouldn't it be unexpected for a client who wants to save the resume data only after receiving the metadata for the magnet link, if instead it is saved when adding the tracker?

src/torrent.cpp
```diff
@@ -247,6 +245,9 @@ bool is_downloading_state(int const st)
 		, m_outstanding_file_priority(false)
 		, m_complete_sent(false)
 	{
+		if (p.flags & torrent_flags::need_save_resume)
+			set_need_save_resume(torrent_handle::if_metadata_changed, true);
```
glassez, line 249: I would explicitly do `m_need_save_resume_data = torrent_handle::if_metadata_changed` instead of introducing such a "bad" parameter as `in_constructor`.

src/torrent.cpp
```diff
@@ -11821,7 +11830,7 @@ namespace {
 		// if we don't have any metadata, stop here
 
 		st->queue_position = queue_position();
-		st->need_save_resume = need_save_resume_data();
+		st->need_save_resume = bool(m_need_save_resume_data);
```
glassez, line 11833: So how about add yet another field to represent it in details?

arvidn: that would be an ABI break. I could do that in `master`

src/torrent.cpp
arvidn: yeah, good point

include/libtorrent/torrent_handle.hpp
arvidn: it's not "instead", it's in addition to. "metadata" has traditionally meant the whole .torrent file. If one wants to save the .torrent file when it's received, trackers and web seeds will be part of it, and I would expect those to count too. But the water gets a bit muddy when considering "nodes" for instance (DHT nodes), which can also be included in torrent files.

include/libtorrent/torrent_handle.hpp
arvidn: yeah, I'm unsure about this. "moved" should probably be considered the "save_path" state. but it's possible that when moving the storage new files will be picked up. However, I suppose that case would be covered already, by checking files.

glassez: >I suppose that case would be covered already, by checking files.

👍

src/torrent.cpp
glassez: > that would be an ABI break. I could do that in `master`

OK. Just don't delay doing it, otherwise it may be forgotten.

PR #7344, opened 2023-03-16
fix using implicit capture of this is deprecated in c++20

joriscarrier:


arvidn:
```
../src/torrent_handle.cpp:120:35: error: explicit capture of 'this' with a capture default of '=' is a C++20 extension [-Werror,-Wc++20-extensions]
```

my understanding is that implicit capture of `this` is only deprecated when the default capture is `=`. I think it's reasonable to not capture `this` implicitly, but the default `=` captures should probably be removed too, to capture everything explicitly

arvidn (approved):

PR #7348, opened 2023-03-19
add need_save_resume field to torrent_status

arvidn:
with the resume data flags indicating which aspects of a torrent have changed since the last resume data was saved


include/libtorrent/torrent_status.hpp
```diff
@@ -448,6 +448,12 @@ TORRENT_VERSION_NAMESPACE_3
 		// was saved.
 		bool need_save_resume = false;
 
+		// These are the flags indicating which aspects of this torrent have
+		// changed since the last time resume data was saved. See
+		// torrent_handle::save_resume_data(). It is a higher fidelity version
+		// of the need_save_resume field.
+		resume_data_flags_t need_save_resume_data;
```
glassez, line 455: Would you mind to remove/deprecate legacy `need_save_resume` field?

Issue #7351, opened 2023-03-21
I2P: acceptor is not recreated after connection is accepted


Vort:
When acceptor accepts incoming connection, new acceptor should be created instead of used one.
But currently, this mechanism is broken, probably because of commit b61c492b376b5b0d2f5037f8bf4d63f0c8d0cc7c.

I think this problem can be fixed by moving `open_new_incoming_i2p_connection()` after `m_i2p_listen_socket.reset()`, like that: https://github.com/Vort/libtorrent/commit/84f53312df29a9389c6196918a977f08727025e4.

_libtorrent version (or branch):_ 58170282b081027aaed5ee234f8ffa0ed8208104
_platform/architecture:_ Windows x64
_compiler and compiler version:_ clang 15.0.7

Vort:
Implemented with #7356

PR #7356, opened 2023-03-26
expose i2p destination in `peer_info`

arvidn:
The main feature is to add a function `i2p_destination()` to `peer_info`, which returns the base32 address (in binary form, i.e. `sha256(base64(destination))`) for i2p peers.

There's a number of other commits that fix and improve other aspects of i2p support as well

### DIFF
--- a/examples/client_test.cpp
+++ b/examples/client_test.cpp
@@ -330,6 +330,56 @@ int peer_index(lt::tcp::endpoint addr, std::vector<lt::peer_info> const& peers)
 	return int(i - peers.begin());
 }
 
+#if TORRENT_USE_I2P
+void base32encode_i2p(lt::sha256_hash const& s, std::string& out, int limit)
+{
+	static char const base32_table[] =
+	{
+		'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h',
+		'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p',
+		'q', 'r', 's', 't', 'u', 'v', 'w', 'x',
+		'y', 'z', '2', '3', '4', '5', '6', '7'
+	};
+
+	static std::array<int, 6> const input_output_mapping{{0, 2, 4, 5, 7, 8}};
+
+	std::array<std::uint8_t, 5> inbuf;
+	std::array<std::uint8_t, 8> outbuf;
+
+	TORRENT_ASSERT(s.size() % 5 );
+	for (auto i = s.begin(); i != s.end();)
+	{
+		int const available_input = std::min(int(inbuf.size()), int(s.end() - i));
+
+		// clear input buffer
+		inbuf.fill(0);
+
+		// read a chunk of input into inbuf
+		std::copy(i, i + available_input, inbuf.begin());
+		i += available_input;
+
+		// encode inbuf to outbuf
+		outbuf[0] = (inbuf[0] & 0xf8) >> 3;
+		outbuf[1] = (((inbuf[0] & 0x07) << 2) | ((inbuf[1] & 0xc0) >> 6)) & 0xff;
+		outbuf[2] = ((inbuf[1] & 0x3e) >> 1);
+		outbuf[3] = (((inbuf[1] & 0x01) << 4) | ((inbuf[2] & 0xf0) >> 4)) & 0xff;
+		outbuf[4] = (((inbuf[2] & 0x0f) << 1) | ((inbuf[3] & 0x80) >> 7)) & 0xff;
+		outbuf[5] = ((inbuf[3] & 0x7c) >> 2);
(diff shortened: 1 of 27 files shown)


Vort:
> (in binary form, i.e. sha256(base64(destination)))

I'm not sure about it, but it looks like base64 destination should be decoded before hashing.

glassez:
> > (in binary form, i.e. sha256(base64(destination)))
> 
> I'm not sure about it, but it looks like base64 destination should be decoded before hashing.

:+1
Destination is base64 encoded. So to produce sha256 hash it should be decided first.

src/bt_peer_connection.cpp
```diff
@@ -481,7 +481,13 @@ namespace {
 		if (support_extensions()) p.flags |= peer_info::supports_extensions;
 		if (is_outgoing()) p.flags |= peer_info::local_connection;
 #if TORRENT_USE_I2P
-		if (is_i2p(get_socket())) p.flags |= peer_info::i2p_socket;
+		if (is_i2p(get_socket()))
+		{
+			p.flags |= peer_info::i2p_socket;
+			auto const* pi = peer_info_struct();
+			sha256_hash const b32_hash = hasher256(pi->dest()).final();
```
glassez, line 488: `b32_` prefix confuses here.

glassez:
Also, please pay attention to the following code:
https://github.com/arvidn/libtorrent/blob/a175e62e0d89a17b8a8fe99b800d18360b8089f0/src/torrent.cpp#L3499-L3526

I doubt that it is correct. ` && string_ends_with(i.hostname, ".i2p")` should be removed, IMO.
IIRC, there can be two types of I2P addresses:
1. Base64 encoded I2P destination without any extension/suffix
2. Base32 encoded I2P address having `.b32.i2p` extension/suffix

src/bt_peer_connection.cpp
arvidn: my understanding is that this address is called "base32 address", even though the most interesting property of it is that it's the sha256 hash of the destination.

glassez: > my understanding is that this address is called "base32 address"

Yes. But here is no Base32 address. This is a preform for it.

jiigen:
Would be possible to add that to python bindings so I can test it with deluge?? Thank you very much in any case

arvidn:
> Would be possible to add that to python bindings too so I can test it with deluge?

done! thanks for reminding me

examples/client_test.cpp
```diff
@@ -82,6 +82,11 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "session_view.hpp"
 #include "print.hpp"
 
+#if TORRENT_USE_I2P
+// this is cheating, for base32encode_i2p()
```
Vort, line 86: Does not work for me.
_(However, I'm not sure if it's not because of my mistake)_

ld.lld: error: undefined symbol: libtorrent::base32encode_i2p(libtorrent::span<char const>)
&gt;>> referenced by examples/CMakeFiles/client_test.dir/client_test.cpp.obj...

Seems like adding
```
	../src/escape_string.cpp
	../src/hex.cpp
	../src/parse_url.cpp
	../src/string_util.cpp
	../src/random.cpp
```
to CMakeLists.txt helped.

arvidn: client_test.cpp is currently "cheating", it's using a non-public function (`base32encode_i2p`) from libtorrent. If you build libtorrent as a DLL on windows, that function won't be exported and you'll probably see a link error. It should probably be fixed before landing. maybe client_test doesn't need to print addresses as base32

arvidn:
> It is still crashing, like before, on exit, but it may be addressed later.

It's crashing?

It's working well for me on linux. Could you provide some more hints on how and when?

arvidn:
@Vort I updated the patch to make client_test not depend on a private function from the library. It should work with linking shared on windows now (without additional patches)

arvidn:
I folded https://github.com/arvidn/libtorrent/issues/7351 into this PR too

Vort:
> It should work with linking shared on windows now (without additional patches)

I confirm that CMakeLists.txt hack is not needed anymore.

Also it looks like crashes because of assertions are gone.

arvidn:
i2p is pretty busted without these last commits. it looks like it the i2p-torrent property failed to be restored from resume data, and DHT did not appear to be disabled for i2p torrents either.

arvidn:
I believe this is ready to land

Vort:
I made basic tests with `client_test` and results looks good.
However I'm worried about compatibility with other clients.
Are developers ready to make corresponding changes to their clients?
Do changes in this PR looks good to them too?

arvidn:
This patch is supposed to maintain, not only API compatibility, but ABI compatibility as well

glassez:
> Are developers ready to make corresponding changes to their clients?

👍

glassez:
@arvidn 
Do you intend to backport the latest I2P related changes to 1.2.x branch?

PR #7359, opened 2023-03-28
fix issue with read_piece() stopping torrent

arvidn:
on pieces not yet downloaded

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+	* fix issue with read_piece() stopping torrent on pieces not yet downloaded
 	* improve handling of allow_i2p_mixed setting to work for magnet links
 	* fix web seed request for renamed single-file torrents
 	* fix issue where web seeds could disappear from resume data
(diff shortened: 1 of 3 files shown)


arvidn:
oh, interesting, stopping the torrent *is* documented. I think this patch may not be right.
I think a better way to address this is to not issue a read for a piece we know we don't have.

cas--:
Hmm this change has broken how Deluge handles setting an error state for torrents that were seeding but the files were moved on OS therefore should not try to re-download them without user intervention. I will need to revisit this to understand what would work instead, any suggestions @arvidn? 

Deluge ticket: https://dev.deluge-torrent.org/ticket/1032

cas--:
It's been a while since I touched that code but the tests were using read_piece to force the error which seemed to simulate what was occurring with actual torrents correctly. 

I think there are two scenarios, one where re-adding a torrent to the session and the files on disk are missing and the other where a running session with seeding torrent have their file on disk moved or removed (e.g. unmounted disk)

I created a ticket with sample code to demo the difference in lt versions: https://dev.deluge-torrent.org/ticket/3603

PR #7372, opened 2023-04-08
V2 hash checking

arvidn:
add stricter checking of piece layers when loading torrents (`load_torrent_*()` functions)
add stricter checking of v1 and v2 hashes being consistent


src/torrent.cpp
```diff
@@ -4294,6 +4308,23 @@ namespace {
...
 			// because of other, previously existing block hashes.
 			ret = false;
+
+			// if the hashes for more than one piece have been verified,
+			// check for any pieces which were already checked but couldn't
+			// be verified and mark them as verified
+			using delta = piece_index_t::diff_type;
+			for (piece_index_t verified_piece = piece + delta(last_result.first_verified_block / blocks_per_piece)
+				, end = verified_piece + delta(last_result.num_verified / blocks_per_piece)
+				; verified_piece < end; ++verified_piece)
```
arvidn, line 4318: maybe the computation of this range could be factored out, to not be repeated

PR #7373, opened 2023-04-08
DNS cache improvements

arvidn:


nasbdh9:
Can this refresh the dns cache immediately when the tracker fails?

arvidn:
> Can this refresh the dns cache immediately when the tracker fails?
> such as when there is an error of skipping tracker announce (unreachable)

This patch is about caching DNS failures, to *avoid* "refreshing" the DNS lookup for failing hostnames.

What you're referring to is not related to hostname resolution, but related to there being a listen interface on the same network as the tracker.

glassez:
> What you're referring to is not related to hostname resolution, but related to there being a listen interface on the same network as the tracker.

I think @nasbdh9 is talking about the problem when the tracker's IP has changed, but `libtorrent` still uses cached one.

arvidn:
I believe the message: "skipping tracker announce (unreachable)" is reported when the tracker's IP is on the local network, and you're only listening on the internet (or vice versa)

PR #7374, opened 2023-04-08
fix openssl installation on windows CI

arvidn:


arvidn:
@Chocobo1 do you have any ideas of what might be going on here?
The difference I see from the logs between the (earlier) passing runs is that it used to say the `PATH` variable was being updated, but it doesn't anymore: https://github.com/arvidn/libtorrent/actions/runs/4646231504/jobs/8222533360?pr=7374#step:3:26

Chocobo1:
I took a look and sorry, no idea. 
Here is an earlier good run for comparison: https://github.com/arvidn/libtorrent/actions/runs/4581983414/jobs/8091891357#step:3:26
And the failed run: https://github.com/arvidn/libtorrent/actions/runs/4647235381/jobs/8224113830?pr=7374#step:3:28
Notice that the path for `openssl.cfg` is quite different. I doubt the package wasn't even installed correctly.

BTW here is the openssl page on choco: https://community.chocolatey.org/packages/openssl
And its package repo: https://github.com/chtof/chocolatey-packages/tree/master/automatic/openssl
The last update for openssl was feb 9.

PR #7377, opened 2023-04-10
Back port load_torrent_*()

arvidn:


include/libtorrent/torrent_info.hpp
```diff
@@ -292,6 +292,16 @@ namespace libtorrent {
 			, announce_entry::tracker_source source);
 		std::vector<announce_entry> const& trackers() const { return m_urls; }
 
+		// hidden
+		std::vector<announce_entry> drain_trackers() { return std::move(m_urls); }
```
glassez, line 296: Of course, I understand that you have to use some tricks until you can move all the parsing code to a new place. But if you choose between possible options, I would prefer to avoid adding such "hidden" methods to the interface. Maybe just declare `update_atp()` as `friend`?

arvidn: oh, is it because IDEs pick them up as public?

glassez: > oh, is it because IDEs pick them up as public?

It's because they are declared as public.

arvidn: now I'm not sure what your "it" refers to. My "it" referred to your request of changing this.
I *would* expect IDEs to autocomplete these functions. Is there another reason to not have them?

glassez: Well, since you've already merged something like this to 2.0 branch, there's no point in continuing this discussion. I still hope that this workaround will not get into the codebase of the next version (2.1 or whatever it will be).

arvidn: > I still hope that this workaround will not get into the codebase of the next version (2.1 or whatever it will be).

I hope so too. or at least 2.1.1. I'd like to release 2.1.0 as soon as the disk cache for `pread-disk-io` is done, and possibly some other ABI alterations.

PR #7378, opened 2023-04-10
Comments for clarity

Shreyan11:
Just added a few comments to understand what it actually does,took me quite some time to get it,i feel it'll be helpful to others as well. I'm open to all suggestions

### DIFF
--- a/src/bitfield.cpp
+++ b/src/bitfield.cpp
@@ -42,20 +42,24 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent {
 
-	bool bitfield::all_set() const noexcept
+	bool bitfield::all_set() const noexcept //all set 
 	{
-		if(size() == 0) return false;
+		if(size() == 0) return false; //checks if all bits in the bitfield are set to 1. If the size of the bitfield is zero, it returns false.
 
 		int const words = size() / 32;
 		for (int i = 1; i < words + 1; ++i)
 		{
 			if (m_buf[i] != 0xffffffff) return false;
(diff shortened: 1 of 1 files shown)


src/bitfield.cpp
```diff
@@ -42,20 +42,24 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent {
 
-	bool bitfield::all_set() const noexcept
+	bool bitfield::all_set() const noexcept //all set
```
glassez, line 45: 👎 
Totally meaningless comment, IMO.

src/bitfield.cpp
```diff
@@ -42,20 +42,24 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent {
 
-	bool bitfield::all_set() const noexcept
+	bool bitfield::all_set() const noexcept //all set 
 	{
-		if(size() == 0) return false;
+		if(size() == 0) return false; //checks if all bits in the bitfield are set to 1. If the size of the bitfield is zero, it returns false.
```
glassez, line 47: This comment is confusing in this place.

arvidn: yeah, this comment does not make sense. It would be more sensible to just indicate that it's on purpose that a bitfield with 0 length is not considered have any bits set

arvidn (changes requested):
These comments repeat the logic in the code in english. I don't think that's helpful, it's not supposed to be a tutorial on what various C++ operators do. Stick to the *reason* why these things are done (and stay succinct and precise).

Also, please stick to something closer to 80 columns, and a space after `//`.

PR #7379, opened 2023-04-10
I2p roundup

arvidn:
this cherry-picks a few of the recent i2p fixes from RC_2_0 into RC_1_2. i2p is not recommended to be used in libtorrent-1.2.x still


xavier2k6:
@arvidn You may want to stop the `18.04` runs & update accordingly due to the runner being deprectaed -> https://github.com/actions/runner-images/issues/6002

Issue #7387, opened 2023-04-13
I2P: crash while seeding in mixed mode


Vort:
Sometimes when torrents are seeding, assertion failure happens.
I was able to reproduce it with such settings and with 5 torrents seeding:
`--allow_i2p_mixed=true --i2p_hostname=127.0.0.1 --i2p_port=7656 --i2p_inbound_quantity=8 --i2p_outbound_quantity=8 --i2p_inbound_length=1 --i2p_outbound_length=1`
Most likely, mixed mode is important for reproduction and quantity/length are not, but who knows - bug happens randomly.

It happens on this line:
https://github.com/arvidn/libtorrent/blob/96d055bf5e9622ab4e0376d27a6b422c19c31443/src/hasher.cpp#L216

```text
libtorrent::assert_fail assert.cpp:362
libtorrent::lcrypto::hasher256::update hasher.cpp:216
libtorrent::lcrypto::hasher256::hasher256 hasher.cpp:181
libtorrent::bt_peer_connection::get_specific_peer_info bt_peer_connection.cpp:496
libtorrent::peer_connection::get_peer_info peer_connection.cpp:4613
libtorrent::torrent::get_peer_info torrent.cpp:7267
libtorrent::torrent::post_peer_info torrent.cpp:7274
libtorrent::torrent_handle::async_call<void (libtorrent::torrent::*)()>(void (libtorrent::torrent::*)()) const::{lambda()#1}::operator()() const torrent_handle.cpp:130
boost::asio::detail::binder0<libtorrent::torrent_handle::async_call<void (libtorrent::torrent::*)()>(void (libtorrent::torrent::*)()) const::{lambda()#1}>::operator()() bind_handler.hpp:60
[...]

PR #7390, opened 2023-04-14
accept remote i2p peers with an empty destination address

arvidn:


glassez:
@arvidn 
Do you mind providing some explanations?

arvidn:
@Vort reported an assertion failure (here: https://github.com/arvidn/libtorrent/issues/7387) indicating that the destination address of an i2p peer was empty. I don't know how or why that would happen, this patch just makes it be tolerated.

Vort:
Looks like I know where problem happens:
When I2P peer is accepted, execution goes this way:
```text
libtorrent::ipv4_peer::ipv4_peer torrent_peer.cpp:228
libtorrent::peer_list::new_connection peer_list.cpp:781
libtorrent::torrent::attach_peer torrent.cpp:7968
libtorrent::peer_connection::attach_to_torrent peer_connection.cpp:1355
libtorrent::bt_peer_connection::on_receive_impl bt_peer_connection.cpp:2972
libtorrent::bt_peer_connection::on_receive bt_peer_connection.cpp:2753
libtorrent::peer_connection::on_receive_data peer_connection.cpp:6177
...
```
Which results in creation of wrong object, `ipv4_peer` instead of `i2p_peer`:
https://github.com/arvidn/libtorrent/blob/96d055bf5e9622ab4e0376d27a6b422c19c31443/src/peer_list.cpp#L778-L781

PR #7392, opened 2023-04-16
fix peer_list to support incoming i2p connections

arvidn:
this fixes a few issues in the peer_list when putting i2p peers in it.
* incoming i2p connections did not work correctly
* detecting self connections did not work correctly
* resolving a double-connect (where two peers connect to each other at the same time), did not work correctly


Vort:
I tested seeding in mixed mode both with client_test and qBittorrent.
Result is good: during 1 hour test there were no crashes.

Issue #7405, opened 2023-05-04
Set `MADV_RANDOM` in `madvise` call


database64128:
**Please provide the following information**

libtorrent version (or branch): RC_2_0

platform/architecture: Linux version 6.2.9-arch1-1

compiler and compiler version: gcc 12.2.1

please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

Seeding at 2 MiB/s results in over 40 MiB/s of constant disk read. This is a symptom of excessive read-ahead. So I took a quick glance at the code. It seems that for opened files, `POSIX_FADV_RANDOM` is set:

https://github.com/arvidn/libtorrent/blob/272828e1cc37b042dfbbafa539222d8533e99755/src/file.cpp#L563-L569

But not `MADV_RANDOM`:

https://github.com/arvidn/libtorrent/blob/272828e1cc37b042dfbbafa539222d8533e99755/src/mmap.cpp#L187-L206

So the question becomes: Does setting `POSIX_FADV_RANDOM` has the same effect as setting `MADV_RANDOM` on `mmap(2)` calls? A quick search in the kernel source tree seems to suggest that they are mostly unrelated in terms of the logic they touch:

https://github.com/torvalds/linux/blob/1a5304fecee523060f26e2778d9d8e33c0562df3/mm/madvise.c#L1033-L1035

https://github.com/torvalds/linux/blob/1a5304fecee523060f26e2778d9d8e33c0562df3/mm/fadvise.c#L86-L90

https://github.com/torvalds/linux/blob/1a5304fecee523060f26e2778d9d8e33c0562df3/mm/filemap.c#L3155-L3157

If you think it's indeed a good idea to set the `MADV_RANDOM` flag, I can open a PR for it.

database64128:
I'll give it a go and see if it actually reduces disk read.

alanswanson:
I believe you're correct (that they're unrelated). Not that I suffer from this with my torrents, as I've set a maximum of ten upload slots but usually less are active, but it's a simple enough patch if you've not tried it yourself yet.

database64128:
> but it's a simple enough patch if you've not tried it yourself yet.

I did try it after posting that comment: database64128/libtorrent@1c9caddad12071129f8f9b5bc703cfe95bdc20d8

I wasn't able to observe a significant enough reduction in IO for me to immediately open an PR.

ValeZAA:
>This is a symptom of excessive read-ahead. 

[...]

PR #7409, opened 2023-05-10
use madvise(MADV_RANDOM) when torrenting mmaped files to reduce disk readahead

alanswanson:
For #7405. When torrenting (random_access) files, reduce excessive disk IO readahead with mmaped files by using MADV_RANDOM and hence matching normal file access using POSIX_FADV_RANDOM.

Could cause more IO calls though as unfortunately there's no reduced read ahead option to madvise.

### DIFF
--- a/src/mmap.cpp
+++ b/src/mmap.cpp
@@ -187,7 +187,7 @@ file_mapping::file_mapping(file_handle file, open_mode_t const mode, std::int64_
 #if TORRENT_USE_MADVISE
 	if (file_size > 0)
 	{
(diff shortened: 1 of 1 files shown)


arvidn:
Have you measured the performance impact of this change?

arvidn:
reading up on the issue, it sounds like you did not observer a clear improvement. Are you confident the excess reading from disk is caused by read-ahead?
I'm a bit hesitant to throw additional flags in without at least some evidence of it improving things (even though it seems very reasonable to hint random access).

I encountered a windows pitfall a long time ago, relating to a flag like this: https://blog.libtorrent.org/2012/05/windows-disk-cache/

PR #7410, opened 2023-05-11
fix gcc 13 build

battlmonstr:
include/libtorrent/aux_/numeric_cast.hpp:52:33: error:
    'int64_t' is not a member of 'std'; did you mean 'int64_t'?

see: https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes


arvidn (approved):

arvidn:
thanks!

PR #7416, opened 2023-05-14
fix function announce_with_tracker assert

pzc105:
When start a full check, should stop announcing that prevent assert in function announce_with_tracker.
`TORRENT_ASSERT(e == event_t::stopped || state() != torrent_status::checking_files)`


arvidn:
would you mind dropping the two first commits?

arvidn:
can you post the stack trace of when that assert fires?

arvidn (approved):

arvidn:
I made this PR instead: https://github.com/arvidn/libtorrent/pull/7418

Issue #7354, opened 2023-03-24
I2P: make session options configurable


Vort:
Right now code contains following line:
https://github.com/arvidn/libtorrent/blob/58170282b081027aaed5ee234f8ffa0ed8208104/include/libtorrent/i2p_stream.hpp#L433

These defaults are good, however, lots of users will prefer to change them.
That's why new configuration options are needed.

This is how I think such implementation can be done: https://github.com/Vort/libtorrent/commit/08041f267ba4a05d8c153b373c836aa59d10171e.

_libtorrent version (or branch):_ 58170282b081027aaed5ee234f8ffa0ed8208104
_platform/architecture:_ Windows x64
_compiler and compiler version:_ clang 15.0.7

Vort:
I made my change compatible with latest version: https://github.com/Vort/libtorrent/commit/d385e4c055719c063ec4d753dd2534dbd3aa8b4b.

arvidn:
@Vort would you mind making a PR with that change?

A few notes:

1. adding settings to a `settings_pack` enum has to be at the end, otherwise some settings will change value and that's incompatible with the existing ABI.
2. I think it would be good to omit options that haven't been set explicitly, when building the `SESSION CREATE` string. I'd like libtorrent to defer default values to the i2p router as much as possible.
3. it *might* be worth exploring putting these settings in their own class, to be able to pass them as one bundle to `open()`. `m_inbound_quantity`, `m_outbound_quantity`, `m_inbound_length`, `m_outbound_length`

Vort:
> would you mind making a PR with that change?

Done: #7419.

Vort:
Implemented in #7422.

PR #7419, opened 2023-05-14
make i2p session options configurable

Vort:
Implements #7354.

> 1. adding settings to a settings_pack enum has to be at the end, otherwise some settings will change value and that's incompatible with the existing ABI.

Ok, I made this change.

> 2. I think it would be good to omit options that haven't been set explicitly, when building the SESSION CREATE string. I'd like libtorrent to defer default values to the i2p router as much as possible.

Options are set explicitly intentionally.
It is needed for consistent behaviour among two I2P implementations (i2pd and Java I2P).

> 3. it might be worth exploring putting these settings in their own class, to be able to pass them as one bundle to open(). m_inbound_quantity, m_outbound_quantity, m_inbound_length, m_outbound_length

I don't have strong opinion on this.
But I made such change. Technically it should work, but I know little about style traditions in this project, so I'm not sure if my changes aesthetically good.

### DIFF
--- a/include/libtorrent/i2p_stream.hpp
+++ b/include/libtorrent/i2p_stream.hpp
@@ -101,6 +101,22 @@ namespace libtorrent {
 	{ return i2p_category(); }
 #endif
 
+struct i2p_session_options
+{
+	i2p_session_options(int inbound_quantity = 3,
+		int outbound_quantity = 3, int inbound_length = 3, int outbound_length = 3)
+		: m_inbound_quantity(inbound_quantity)
+		, m_outbound_quantity(outbound_quantity)
+		, m_inbound_length(inbound_length)
+		, m_outbound_length(outbound_length)
+	{}
+
+	int m_inbound_quantity;
+	int m_outbound_quantity;
+	int m_inbound_length;
+	int m_outbound_length;
+};
+
 struct i2p_stream : proxy_base
 {
 	explicit i2p_stream(io_context& io_context);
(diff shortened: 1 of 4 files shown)


arvidn:
is there a good reason to override the defaults of the i2p router?
as this patch is right now, you don't have the option to defer to the default

Vort:
> is there a good reason to override the defaults of the i2p router?

Regarding `quantity`:
From #7258:
> We also recommend specifying inbound.quantity=x outbound.quantity=x in the SESSION CREATE line, as i2pd and Java I2P have different defaults.

From https://geti2p.net/en/docs/api/samv3:
> The Java I2P and i2pd routers have different defaults for tunnel quantities. The Java default is 2 and the i2pd default is 5. For most low- to medium-bandwidth and low- to medium-connection counts, 2 or 3 is sufficient. Please specify the tunnel quantities in the SESSION CREATE message to get consistent performance with the Java I2P and i2pd routers _[rest of citation contains mistake, so I omitted it -- Vort]_

Regarding `length`:
It makes sense to use the same approach as for `quantity` because of consistency and because these defaults are very stable. From #7262:
> [length] 3 is both the min and the max required for anonymity, and it is the default in both routers.

---
> as this patch is right now, you don't have the option to defer to the default

Because of reasons mentioned above, I think users will never have need to prefer router defaults over libtorrent defaults.

Vort:
As example, this is how defaults are handled in I2PSnark:
https://github.com/i2p/i2p.i2p/blob/18a3893185e0a524069eb3a51b35823a8ff31540/apps/i2psnark/java/src/org/klomp/snark/SnarkManager.java#L847-L849
```java
_config.setProperty(PROP_I2CP_OPTS, "inbound.length=3 outbound.length=3" +
                                    " inbound.quantity=" + DEFAULT_TUNNEL_QUANTITY +
                                    " outbound.quantity=" + DEFAULT_TUNNEL_QUANTITY);
```

arvidn:
thanks! I will make some finishing touches to this patch.

arvidn:
replaced by https://github.com/arvidn/libtorrent/pull/7422

PR #7437, opened 2023-05-23
don't run win32 tests when building python wheel

arvidn:
when building the python wheel, the 32 bit windows tests seem to fail consistently. There's no output, so it's not so easy to understand why they fail. However, building and publishin the wheels only happen after CI is green anyway, it shouldn't be that important to run the tests as part of the publishing step as well. Especially if some tests are flaky.


arvidn:
testing here: https://github.com/arvidn/libtorrent/actions/runs/5054689811

arvidn:
@AllSeeingEyeTolledEweSew I'm considering landing this in `RC_2_0` in order to publish wheels to pypi

Issue #7265, opened 2023-01-15
[feature] add pypi for python 3.11


pseudo-rnd-thoughts:
With python 3.11 released, could pypi have pre build libtorrent for python 3.11 

I can't find a prior issue on this topic nor does the current change logs include this 

But looking through the .GitHub folder, I see that python 3.11 is used in the python.yml

malomehi:
Yes please. It would be great.
Some distros are now using 3.11 as the default python3 version.

cas--:
Looks like it's being currently skipped for cibuildwheel since boost-python >= 1.81 is required so should be relatively easy to resolve...

Skipped for builds: https://github.com/arvidn/libtorrent/blob/2b6360320b1b1bf3a28b72c9866071b2b155ac21/pyproject.toml#L5

Skipped for macos too: https://github.com/arvidn/libtorrent/blob/2b6360320b1b1bf3a28b72c9866071b2b155ac21/.cirrus.yml#L48

Solomon1732:
Is version 2.0.9 a stable release or is it, say, a release candidate (or beta or somesuch)? I'm asking here since it looks like it support python 3.11, but I'm not sure. I'm basing it on the fact `pyproject.toml` doesn't skip it anymore.

cas--:
Anything with a release version is stable

PR #7445, opened 2023-05-25
Bump boost version to enable Python 3.11 package builds

cas--:
Python 3.11 is not compatible with Boost Python <= 1.80 due to error with Py_TPFLAGS_HAVE_GC enum.

Ref: https://github.com/boostorg/python/issues/400
Fixes: https://github.com/arvidn/libtorrent/issues/7265

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 requires-python = ">=3.7"
 
 [tool.cibuildwheel]
-skip = "{pp*,cp311-*}"
+skip = "{pp*,}"
 
 [tool.cibuildwheel.macos]
 before-all = [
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -17,7 +17,7 @@ test-command = [
 [tool.cibuildwheel.macos.environment]
 BOOST_BUILD_PATH = "/tmp/boost/tools/build"
 BOOST_ROOT = "/tmp/boost"
-BOOST_VERSION = "1.80.0"
+BOOST_VERSION = "1.81.0"
```
arvidn, line 20: what does this affect?

cas--: Unsure what you mean? This is for the macos wheels

arvidn: I'm not very familiar with how the wheels are build, so I'm curious what logic looks at this constant and installs the corresponding version of boost.

cas--: I'm not a fan of toml... Check the previous table for the macos boost install, this table is nested env vars for it

PR #7456, opened 2023-06-20
Don't use visibility("hidden") on Windows

lazka:
clang v16 (in combination with mingw-w64) has gotten stricter and now fails with:

"error: hidden visibility cannot be applied to 'dllexport' declaration"

Since "visibility" isn't used on Windows anyway, just skip it there.

### DIFF
--- a/include/libtorrent/aux_/export.hpp
+++ b/include/libtorrent/aux_/export.hpp
@@ -100,7 +100,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #endif
 
 #if !defined TORRENT_EXPORT_EXTRA \
-  && ((defined __GNUC__ && __GNUC__ >= 4) || defined __clang__)
+  && ((defined __GNUC__ && __GNUC__ >= 4) || defined __clang__) && !defined _WIN32
 # define TORRENT_UNEXPORT __attribute__((visibility("hidden")))
 #else
 # define TORRENT_UNEXPORT


kasper93:
I don't think this patch do the right thing.

> error: hidden visibility cannot be applied to 'dllexport' declaration

It is clearly a bug to set visibility hidden on dllexport declaration. They contradict each other, so it should be fixes and diagnosed why both are set.

lazka:
ok, let me see if I can find more context

lazka:
Here is the full error:

```
C:/msys64/home/user/git/libtorrent/src/alert.cpp:104:9: error: hidden visibility cannot be applied to 'dllexport' declaration
        alert::alert() : m_timestamp(clock_type::now()) {}
               ^
C:/msys64/home/user/git/libtorrent/src/alert.cpp:108:17: error: hidden visibility cannot be applied to 'dllexport' declaration
        torrent_alert::torrent_alert(aux::stack_allocator& alloc
                       ^
C:/msys64/home/user/git/libtorrent/src/alert.cpp:154:14: error: hidden visibility cannot be applied to 'dllexport' declaration
        peer_alert::peer_alert(aux::stack_allocator& alloc
                    ^
C:/msys64/home/user/git/libtorrent/src/alert.cpp:176:17: error: hidden visibility cannot be applied to 'dllexport' declaration
        tracker_alert::tracker_alert(aux::stack_allocator& alloc
                       ^
C:/msys64/home/user/git/libtorrent/src/alert.cpp:201:20: error: hidden visibility cannot be applied to 'dllexport' declaration
        read_piece_alert::read_piece_alert(aux::stack_allocator& alloc
                          ^
C:/msys64/home/user/git/libtorrent/src/alert.cpp:201:20: error: hidden visibility cannot be applied to 'dllexport' declaration
C:/msys64/home/user/git/libtorrent/src/alert.cpp:210:20: error: hidden visibility cannot be applied to 'dllexport' declaration
        read_piece_alert::read_piece_alert(aux::stack_allocator& alloc
                          ^
C:/msys64/home/user/git/libtorrent/src/alert.cpp:210:20: error: hidden visibility cannot be applied to 'dllexport' declaration
C:/msys64/home/user/git/libtorrent/src/alert.cpp:242:24: error: hidden visibility cannot be applied to 'dllexport' declaration
        file_completed_alert::file_completed_alert(aux::stack_allocator& alloc
                              ^
C:/msys64/home/user/git/libtorrent/src/alert.cpp:242:24: error: hidden visibility cannot be applied to 'dllexport' declaration
C:/msys64/home/user/git/libtorrent/src/alert.cpp:263:22: error: hidden visibility cannot be applied to 'dllexport' declaration
        file_renamed_alert::file_renamed_alert(aux::stack_allocator& alloc
                            ^
C:/msys64/home/user/git/libtorrent/src/alert.cpp:263:22: error: hidden visibility cannot be applied to 'dllexport' declaration
C:/msys64/home/user/git/libtorrent/src/alert.cpp:302:28: error: hidden visibility cannot be applied to 'dllexport' declaration
        file_rename_failed_alert::file_rename_failed_alert(aux::stack_allocator& alloc
                                  ^
C:/msys64/home/user/git/libtorrent/src/alert.cpp:302:28: error: hidden visibility cannot be applied to 'dllexport' declaration
C:/msys64/home/user/git/libtorrent/src/alert.cpp:326:21: error: hidden visibility cannot be applied to 'dllexport' declaration
        performance_alert::performance_alert(aux::stack_allocator& alloc
                           ^
C:/msys64/home/user/git/libtorrent/src/alert.cpp:326:21: error: hidden visibility cannot be applied to 'dllexport' declaration
C:/msys64/home/user/git/libtorrent/src/alert.cpp:366:23: error: hidden visibility cannot be applied to 'dllexport' declaration
        state_changed_alert::state_changed_alert(aux::stack_allocator& alloc
                             ^
C:/msys64/home/user/git/libtorrent/src/alert.cpp:366:23: error: hidden visibility cannot be applied to 'dllexport' declaration
C:/msys64/home/user/git/libtorrent/src/alert.cpp:390:23: error: hidden visibility cannot be applied to 'dllexport' declaration
        tracker_error_alert::tracker_error_alert(aux::stack_allocator& alloc
                             ^
fatal error: too many errors emitted, stopping now [-ferror-limit=]
221 warnings and 20 errors generated.
```

I think the problem is that `struct TORRENT_EXPORT alert` exports the struct and all its members, and `TORRENT_UNEXPORT` then tries to revert it on some members, which doesn't work.

kasper93:
> I think the problem is that struct TORRENT_EXPORT alert exports the struct and all its members, and TORRENT_UNEXPORT then tries to revert it on some members, which doesn't work.

I see, so this patch cripples this whole mechanism. Proper solution if members are to be selectively exported then need to decorated individually, "unexporting" seems like asking for trouble as we can see.

Nevertheless, as a fix for immediate compilation errors it is fine, just doesn't really resolves the issue, just disables it.

lazka:
> I see, so this patch cripples this whole mechanism.

since the visibility attribute has no affect on Windows anyway this shouldn't change anything.

kasper93:
> since the visibility attribute has no affect on Windows anyway this shouldn't change anything (and I do wonder why clang has started to throw such errors given that...)

I think your assumption is not fully correct, this attribute is supported and does "things". I don't care enough to dig into details, but long story short this error message has a merit and as we see it discovered usage that is not working correctly as one would expect.

All I'm saying to be fully correct we would need more changes, but if we agree that those symbols are exported regardless of UNEXPORT, it is probably fine too.

FWIW, here is the change https://github.com/llvm/llvm-project/commit/6f9c4851ab7c729812bc8f48b19de3f84a64f6f0 and review with discussion about it https://reviews.llvm.org/D133266 .

include/libtorrent/aux_/export.hpp
```diff
@@ -100,7 +100,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #endif
 
 #if !defined TORRENT_EXPORT_EXTRA \
-  && ((defined __GNUC__ && __GNUC__ >= 4) || defined __clang__)
+  && ((defined __GNUC__ && __GNUC__ >= 4) || defined __clang__) && !defined _WIN32
```
arvidn, line 103: I would appreciate a comment explaining that the problematic case is clang on windows. I suppose `_WIN32` is a good proxy for this, although `_WIN32` indicates a platform, not a compiler. but the interaction with shared libraries is platform specific anyway

arvidn:
>  see, so this patch cripples this whole mechanism.

No, it does not cripple the *whole* mechanism. The types are still exported, it's just that the constructors (which are considered implementation details and not part of the public API (or ABI) are exported on windows. That was already case when building with MSVC, so this is not causing any regression as far as I can tell.

Issue #7398, opened 2023-04-24
thread stuck when destroy mmap_disk_job


Lwxiang:
**Please provide the following information**

libtorrent version (or branch): 2.0.8
boost version: 1.67.0

platform/architecture: centos7/x86_64

compiler and compiler version: gcc7.3.1

thread stuck in boost `find_prev`
```bash
#0  0x0000000000f01302 in boost::simple_segregated_storage<unsigned long>::find_prev(void*) ()
#1  0x0000000000f0117c in boost::simple_segregated_storage<unsigned long>::ordered_free(void*) ()
#2  0x0000000000f00d94 in boost::pool<libtorrent::aux::allocator_new_delete>::ordered_free(void*) ()
#3  0x0000000000f00d08 in boost::object_pool<libtorrent::aux::mmap_disk_job, libtorrent::aux::allocator_new_delete>::free(libtorrent::aux::mmap_disk_job*) ()
#4  0x0000000000f00c8d in boost::object_pool<libtorrent::aux::mmap_disk_job, libtorrent::aux::allocator_new_delete>::destroy(libtorrent::aux::mmap_disk_job*) ()
#5  0x0000000000f00999 in libtorrent::aux::disk_job_pool::free_jobs(libtorrent::aux::mmap_disk_job**, int) ()
#6  0x0000000000dbab79 in libtorrent::mmap_disk_io::call_job_handlers() ()
#7  0x0000000000dba60b in libtorrent::mmap_disk_io::add_completed_jobs_impl(libtorrent::tailqueue<libtorrent::aux::mmap_disk_job>, libtorrent::tailqueue<libtorrent::aux::mmap_disk_job>&)::{lambda()#1}::operator()() const ()
#8  0x0000000000dbc0f0 in void boost::asio::asio_handler_invoke<libtorrent::mmap_disk_io::add_completed_jobs_impl(libtorrent::tailqueue<libtorrent::aux::mmap_disk_job>, libtorrent::tailqueue<libtorrent::aux::mmap_disk_job>&)::{lambda()#1}>(libtorrent::mmap_disk_io::add_completed_jobs_impl(libtorrent::tailqueue<libtorr
ent::aux::mmap_disk_job>, libtorrent::tailqueue<libtorrent::aux::mmap_disk_job>&)::{lambda()#1}&, ...) ()
[...]

Issue #7386, opened 2023-04-12
crash in peer allocator


Lwxiang:
**Please provide the following information**

libtorrent version (or branch): 1.2.18

platform/architecture: centos7/x86_64

compiler and compiler version: gcc7.3.1

observed a crash in peer allocator when handling new connections
```bash
Program terminated with signal 11, Segmentation fault.
#0  0x0000000000b42ab8 in boost::simple_segregated_storage<unsigned long>::malloc() ()
Missing separate debuginfos, use: debuginfo-install glibc-2.17-323.tl2.x86_64 libgcc-4.8.5-39.tl2.1.x86_64 libstdc++-4.8.5-39.tl2.1.x86_64
(gdb) bt
#0  0x0000000000b42ab8 in boost::simple_segregated_storage<unsigned long>::malloc() ()
#1  0x0000000000b429fd in boost::pool<libtorrent::aux::allocator_new_delete>::malloc() ()
#2  0x0000000000b4254a in libtorrent::torrent_peer_allocator::allocate_peer_entry(int) ()
#3  0x0000000000d63ee9 in libtorrent::peer_list::new_connection(libtorrent::peer_connection_interface&, int, libtorrent::torrent_state*) ()
#4  0x0000000000a9309f in libtorrent::torrent::attach_peer(libtorrent::peer_connection*) ()
#5  0x0000000000d19aba in libtorrent::peer_connection::attach_to_torrent(libtorrent::digest32<160l> const&) ()
#6  0x0000000000c53c1c in libtorrent::bt_peer_connection::on_receive_impl(unsigned long) ()
#7  0x0000000000c51766 in libtorrent::bt_peer_connection::on_receive(boost::system::error_code const&, unsigned long) ()
#8  0x0000000000d2c5d1 in libtorrent::peer_connection::on_receive_data(boost::system::error_code const&, unsigned long) ()
#9  0x0000000000d57504 in void std::__invoke_impl<void, void (libtorrent::peer_connection::* const&)(boost::system::error_code const&, unsigned long), std::shared_ptr<libtorrent::peer_connection> const&, boost::system::error_code const&, unsigned long const&>(std::__invoke_memfun_deref, void (libtorrent::peer_connection::* const&)(boost::system::error_code const&, unsigned long), std::shared_ptr<libtorrent::peer_connection> const&, boost::system::error_code const&, unsigned long const&) ()
[...]

PR #7460, opened 2023-06-27
FIX(peer_list): clear the m_candidate_cache when clear peer_list, and…

Lwxiang:
… set m_num_seeds to 0

### DIFF
--- a/include/libtorrent/peer_list.hpp
+++ b/include/libtorrent/peer_list.hpp
@@ -155,6 +155,7 @@ namespace libtorrent {
 #endif
 
 		int num_peers() const { return int(m_peers.size()); }
+		int num_candidate_cache() const { return int(m_candidate_cache.size()); }
 
 		using peers_t = aux::deque<torrent_peer*>;
 		using iterator = peers_t::iterator;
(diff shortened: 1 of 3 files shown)


Lwxiang:
when user call `torrent_handle::clear_peers()`, it will trigger the `peer_list::clear()`.

then the `m_candidate_cache` which contains the pointers of peers should also be cleaned after the pointers being freed. I believe this issue causes some crash in #7386 or #7398.

also the `m_num_seeds` should be reset to 0.

arvidn:
great catch!
yes, I will have to update the test certificates.

Would you mind extending the unit test to cover this fix as well?

Lwxiang:
sure, I will add some test later.

Lwxiang:
already add the `clear_peers` unit test for peer_list.

before fix, it will not pass the test:
企业微信截图_549922df-747b-4b77-be20-55bf821b7133

after fix, it pass the test:
企业微信截图_b35c76e9-2e4c-450e-a8f6-c36d0cbc197f

Lwxiang:
@arvidn hi PTAL, can this be merged?

arvidn (approved):

Lwxiang:
@arvidn seems there are some mistakes with `test_resume`, the testcase compares the microsecond-duration with second-duration.

PR #7469, opened 2023-07-05
Allow to change save path without moving any files

glassez:


glassez:
@arvidn, ping!
Are you on vacation?

arvidn:
this makes sense. I'll add some tests too.

arvidn:
https://github.com/arvidn/libtorrent/pull/7498

PR #7488, opened 2023-08-19
Remove redundant move in return statement

Chocobo1:
This suppresses the following warning from gcc 13.2.1:
```
/home/user/libtorrent-rasterbar-2_0-git/src/libtorrent/bindings/python/src/entry.cpp:23:25: warning: redundant move in return statement [-Wredundant-move]
   23 |         return std::move(result);
      |                ~~~~~~~~~^~~~~~~~
```

Also refer to the manual:
https://gcc.gnu.org/onlinedocs/gcc-13.2.0/gcc/C_002b_002b-Dialect-Options.html#index-Wredundant-move

### DIFF
--- a/bindings/python/src/entry.cpp
+++ b/bindings/python/src/entry.cpp
@@ -20,7 +20,7 @@ struct entry_to_python
             result.append(*i);
         }
 
-        return std::move(result);
+        return result;
     }
 
     static object convert(entry::dictionary_type const& d)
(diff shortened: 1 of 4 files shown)


glassez (approved):

Chocobo1:
It seems clang-10 gives opposite warnings: https://github.com/arvidn/libtorrent/actions/runs/5909524042/job/16030198579?pr=7488#step:8:32 or perhaps clang 10 is simply outdated. 

This blog post actually suggest not using `std::move` in those cases (the paragraph about `Wpessimizing-move`): https://developers.redhat.com/blog/2019/04/12/understanding-when-not-to-stdmove-in-c
However this might be specific to gcc and clang thinks differently, I'm not sure.

Anyway I'll leave this PR open and let you (arvidn) decide.

arvidn:
I made this: https://github.com/arvidn/libtorrent/pull/7501

Chocobo1:
Superseded by #7501.

PR #7491, opened 2023-08-23
Allow to compile RC_1_2 branch in C++20 mode

Chocobo1:
Previously boost 1.81 reported some errors about parameters should be movable.
Inspired by: 417405fc892903c8f850ca88d02bc045e56d2e19

I suppose this might be breaking ABI. Anyway I'll leave it to arvidn to decide.


glassez (approved):

arvidn (approved):
none of these types are exported in the public API as far as I can tell, so it shouldn't affect the ABI

PR #7492, opened 2023-08-23
Add explicit `this` to lambda

Chocobo1:
Implicit capture of `this` via `[=]` is deprecated in C++20 [-Wdeprecated]

### DIFF
--- a/src/resolver.cpp
+++ b/src/resolver.cpp
@@ -111,7 +111,11 @@ namespace libtorrent {
 		address const ip = make_address(host, ec);
 		if (!ec)
 		{
+#if __cplusplus >= 202002L
+			m_ios.post([=, this]{ callback(h, ec, std::vector<address>{ip}); });
+#else
 			m_ios.post([=]{ callback(h, ec, std::vector<address>{ip}); });
+#endif
 			return;
 		}
 		ec.clear();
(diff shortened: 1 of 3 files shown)


glassez (approved):

arvidn:
clang-tidy flags this at: https://github.com/arvidn/libtorrent/actions/runs/5951657979/job/16141913252?pr=7492#step:7:100
as 
```
/home/runner/work/libtorrent/libtorrent/src/session_impl.cpp:5895:10: error: explicit capture of 'this' with a capture default of '=' is a C++20 extension [clang-diagnostic-c++20-extensions,-warnings-as-errors]
                        , [=, this](aux::listen_socket_handle const& sock
                              ^
```

Is explicit capture of `this` not supported before C++20? If so, it seems a bit drastic to require C++20 support, rather than suppressing the deprecation warning

Chocobo1:
>Is explicit capture of this not supported before C++20? 

Sadly this seems to be the case. From https://en.cppreference.com/w/cpp/language/lambda#Lambda_capture:
>```c++
>struct S2 { void f(int i); };
>void S2::f(int i)
>{
>    [=, this] {};  // until C++20: Error: this when = is the default
>                   // since C++20: OK, same as [=]
>}
>```

Although clang doesn't produce a hard error, there is still a warning: https://godbolt.org/z/o8Eh91Wx4

Now I've updated the PR to include a conditional to ensure compatibility with all cpp versions.

arvidn:
I think it might be a mistake that `this` is captured in some of those cases

Chocobo1:
Superseded by #7502.

PR #7509, opened 2023-08-28
add python bindings

rcarpa:


arvidn:
have you tried to use these functions from python as well?

rcarpa:
Good point. I actually didn't. For some reason I assumed it's as simple as just exposing these bindings and that's all. Is it more tricky than that ?

arvidn:
type conversions are performed at run-time in python, not compile-time. So it may still fail when running it in python

rcarpa:
Just tried the functions from python. Both work as expected.
- `ssl_listen_port` returns an integer (0 if not listening on the ssl socket)
- `set_ssl_certificate_buffer` works correctly when passing strings or bytes with pem-encoded content. I successfully transferred content between two instances of a (patched) Deluge using an encrypted connection. 
 
To be noted that `set_ssl_certificate_buffer` accepts (without any errors) strings containing invalid data. It only results in silent failures to establish the SSL connection. However, this is not something specific to the python bindings.

PR #7512, opened 2023-09-01
Allow loading .torrent files with 1 GB piece size

Chocobo1:
Previously the limit was 1073741823, however 1 GB = 1073741824 which is off by one (at least for x64 systems).
The reason to allow 1 GB piece size is because this is the maximum size that could be created by libtorrent. And it doesn't make sense if one can create it but cannot load it successfully.

~~RC_1_2 is not affected.~~ RC_1_2 can load it but it still has the piece picker problem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1012,6 +1012,7 @@ TEST_TORRENTS = \
   invalid_pieces.torrent \
   invalid_symlink.torrent \
   large.torrent \
+  large_piece_size.torrent \
   long_name.torrent \
   many_pieces.torrent \
   missing_path_list.torrent \
--- a/src/torrent_info.cpp
+++ b/src/torrent_info.cpp
@@ -1162,12 +1162,12 @@ namespace {
 
 		// extract piece length
 		std::int64_t const piece_length = info.dict_find_int_value("piece length", -1);
-		// limit the piece length at INT_MAX / 2 to get a bit of headroom. We
+		// limit the piece length at (INT_MAX / 2 + 1) to get a bit of headroom. We
 		// commonly compute the number of blocks per pieces by adding
 		// block_size - 1 before dividing by block_size. That would overflow with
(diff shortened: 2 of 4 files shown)


glassez (approved):

arvidn:
is the issue that `numeric_limits<int>::max() / 2` truncates the result to be off by one?
That makes sense and seems to be a legitimate problem. Given that some tests are still failing though, it might go a bit deeper than this. I'll take a look

arvidn:
maybe it's misguided to use `numeric_limits<int>` to begin with. if `int` is not 32 bits, a lot of things will probably break. (including this test :) )

arvidn:
@Chocobo1 is your objective to support 1 GiB piece sizes, or to fix an inconsistency/error message?

Right now, the limiting factor is the piece picker using 15 bits to count the various states of blocks in a piece that's being downloaded. Those counters cannot hold 32768, that would wrap. So the max is one less than that. The block size is 0x4000 (16 kiB), which yields a maximum piece size of 32767 * 0x4000 which is (512 MiB - 0x4000).

arvidn:
https://github.com/arvidn/libtorrent/pull/7514

Chocobo1:
>@Chocobo1 is your objective to support 1 GiB piece sizes, or to fix an inconsistency/error message?

It was mainly about trying to load a large piece size torrent and some fixes was done along the way. Originally 1 GiB was the target but later realized it wasn't possible due to piece picker.

>Right now, the limiting factor is the piece picker using 15 bits to count the various states of blocks in a piece that's being downloaded. 

I was thinking if it can utilize a larger data type which can hold more bits but I'm not familiar enough to make patches...

Pentaphon:
@arvidn should this be closed since the variant has been merged?

arvidn:
yeah, probably

PR #7514, opened 2023-09-03
improve checking and testing of max piece size

arvidn:
a variant of @Chocobo1's patch: https://github.com/arvidn/libtorrent/pull/7512


test/test_torrent_info.cpp
```diff
@@ -980,7 +984,7 @@ void sanity_check(std::shared_ptr<torrent_info> const& ti)
 	// for it, it's still no good.
 	piece_picker pp(ti->total_size(), ti->piece_length());
 
-	TEST_CHECK(ti->piece_length() < std::numeric_limits<int>::max() / 2);
+	TEST_CHECK(ti->piece_length() <= (std::numeric_limits<int>::max() / 2 + 1));
```
Chocobo1, line 987: seems you could remove this check?

PR #7545, opened 2023-11-14
ask msvc to report the correct C++ version in __cplusplus macro

arvidn:
context: https://github.com/arvidn/libtorrent/discussions/7534


arvidn:
it appears this is causing simulations to fail on windows. needs more investigation

xavier2k6:
`/Zc\:__cplusplus` - shouldn't this be `/Zc:__cplusplus` ??

arvidn:
> `/Zc\:__cplusplus` - shouldn't this be `/Zc:__cplusplus` ??

I believe `:` needs to be escaped in Jam strings

PR #7546, opened 2023-11-14
address issue in boost.asio building on iOS

arvidn:


xavier2k6:
typo: `keword` -> `keyword`?

PR #7552, opened 2023-11-15
Allow to change save path without moving any files

glassez:
Backport #7469 into RC_1_2 branch.


arvidn (approved):

glassez:
@arvidn 
Could the addition of similar flags to the file renaming function be accepted?

glassez:
Or (maybe even better) add a more fundamental fix so that `libtorrent` never touches those files that are not taken into account in the current progress. I.e. if the file does not intersect with any currently downloaded/checked data, then it should not be affected either when renaming the file or when moving the storage. This would help in resolving various file name conflicts (especially when working with magnet links, when we do not know about the contents of the torrent before receiving metadata).

glassez:
Created separate Issue for above: #7555.

PR #7558, opened 2023-11-18
WIP: add future on asynchronous methods that trigger alerts (should I continue?) 

joriscarrier:
# requires that the lifetime of alerts be guaranteed after session::pop_alerts is called, if they are still referenced.

this pull request allows you to have a future on asynchronous methods that trigger alerts:

### torrent_handle:

- [ ] add_piece
- [x] read_piece
- [ ] post_peer_info
- [ ] post_status
- [ ] post_download_queue
- [ ] post_trackers
- [ ] set_metadata
- [x] flush_cache
- [ ] force_recheck
- [ ] save_resume_data
- [ ] post_piece_availability
- [ ] prioritize_files
- [ ] file_priority
- [ ] scrape_tracker
- [x] move_storage
- [ ] rename_file

### session_handle:

- [ ] post_torrent_updates
- [ ] post_session_stats
- [ ] post_dht_stats
- [ ] async_add_torrent
- [ ] dht_live_nodes
- [ ] dht_sample_infohashes
- [ ] dht_direct_request
- [ ] remove_torrent

### DIFF
--- a/include/libtorrent/alert_types.hpp
+++ b/include/libtorrent/alert_types.hpp
@@ -79,6 +79,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <bitset>
 #include <cstdarg> // for va_list
 
+#include <boost/type_index.hpp>
+
 #if TORRENT_ABI_VERSION == 1
 #define PROGRESS_NOTIFICATION | alert::progress_notification
 #else
@@ -3114,6 +3116,27 @@ TORRENT_VERSION_NAMESPACE_3_END
 
 	// internal
 	TORRENT_EXTRA_EXPORT char const* performance_warning_str(performance_alert::performance_warning_t i);
+
+
+	template<typename T>
+	class alert_exception : public std::exception {
+	public:
+		alert_exception(const T* a) : m_alert(a) {}
+
+		const char* what() const noexcept override {
+			return m_alert->message().c_str();
+		}
+
+	private:
+		const T* m_alert;
+	};
+
+	template<typename T>
+	class alert_dont_post_exception: public std::exception {
+		const char* what() const noexcept override {
+			return std::string(boost::typeindex::type_id<T>().pretty_name() + " ignored").c_str();
+		}
+	};
 
 
 #undef TORRENT_DEFINE_ALERT_IMPL
--- a/include/libtorrent/aux_/alert_manager.hpp
+++ b/include/libtorrent/aux_/alert_manager.hpp
@@ -69,7 +69,7 @@ namespace aux {
 		~alert_manager();
 
 		template <class T, typename... Args>
-		void emplace_alert(Args&&... args) try
+		const T* emplace_alert(Args&&... args) try
 		{
 			std::unique_lock<std::recursive_mutex> lock(m_mutex);
 
@@ -82,13 +82,15 @@ namespace aux {
 			{
 				// record that we dropped an alert of this type
 				m_dropped.set(T::alert_type);
-				return;
+				return nullptr;
 			}
 
 			T& alert = queue.emplace_back<T>(
 				m_allocations[m_generation], std::forward<Args>(args)...);
 
 			maybe_notify(&alert);
+
+			return &alert;
 		}
 		catch (std::bad_alloc const&)
 		{
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -743,7 +743,7 @@ bool is_downloading_state(int const st)
 			m_ses.close_connection(p);
 	}
 
-	void torrent::read_piece(piece_index_t const piece)
+	void torrent::read_piece(piece_index_t const piece, std::shared_ptr<std::promise<const read_piece_alert*>> promise)
 	{
 		error_code ec;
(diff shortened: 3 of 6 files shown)


joriscarrier:
a simple example:
```c++
auto f = th.async_flush_cache();
f.wait()
```

and an other more complex:

```c++
std::future<const lt::save_resume_data_alert*> f = th.async_save_resume_data();

try {
    const auto* alert = f.get();
    std::vector<char> buf = lt::write_torrent_file_buf(alert->params);
} catch (lt::alert_exception<lt::save_resume_data_failed_alert> const& e) {
    const auto* alert = e.get_failed_alert();
}
```

include/libtorrent/aux_/alert_manager.hpp
```diff
@@ -82,13 +82,15 @@ namespace aux {
 			{
 				// record that we dropped an alert of this type
 				m_dropped.set(T::alert_type);
-				return;
+				return nullptr;
```
arvidn, line 85: you need a `return nullptr;` on lines 100 as well

include/libtorrent/aux_/alert_manager.hpp
```diff
@@ -69,7 +69,7 @@
 		~alert_manager();
 
 		template <class T, typename... Args>
-		void emplace_alert(Args&&... args) try
+		const T* emplace_alert(Args&&... args) try
```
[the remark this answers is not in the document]
arvidn: this looks like a parse failure in CodeQL..

include/libtorrent/alert_types.hpp
```diff
@@ -3116,6 +3118,27 @@ TORRENT_VERSION_NAMESPACE_3_END
...
+	class alert_exception : public std::exception {
+	public:
+		alert_exception(const T* a) : m_alert(a) {}
+
+		const char* what() const noexcept override {
+			return m_alert->message().c_str();
+		}
+
+	private:
+		const T* m_alert;
```
arvidn, line 3131: this looks a bit suspicious. alert allocated in the heterogeneous queue have their lifetimes tied to that queue. throwing exceptions with pointers into it seems like asking for accessing danging pointers

src/torrent.cpp
```diff
@@ -8684,26 +8690,35 @@ namespace {
...
 		if (m_abort)
 		{
-			if (alerts().should_post<storage_moved_failed_alert>())
-				alerts().emplace_alert<storage_moved_failed_alert>(get_handle()
+			if (alerts().should_post<storage_moved_failed_alert>()) {
+				auto* a = alerts().emplace_alert<storage_moved_failed_alert>(get_handle()
 					, boost::asio::error::operation_aborted
 					, "", operation_t::unknown);
+
+				if (promise) promise->set_exception(std::make_exception_ptr(alert_exception(a)));
```
arvidn, line 8705: `a` is a non-owning pointer to the alert, and the life-time of the alert is tied to when the user calls `pop_alerts()`. This seems really risky as the future may be checked *after* the alert object has been deleted.

src/torrent.cpp
```diff
@@ -8684,26 +8690,35 @@ namespace {
...
 
 		// if we don't have metadata yet, we don't know anything about the file
 		// structure and we have to assume we don't have any file.
 		if (!valid_metadata())
 		{
-			if (alerts().should_post<storage_moved_alert>())
-				alerts().emplace_alert<storage_moved_alert>(get_handle(), save_path, m_save_path);
+			if (alerts().should_post<storage_moved_alert>()) {
+				auto* a = alerts().emplace_alert<storage_moved_alert>(get_handle(), save_path, m_save_path);
+				if (promise) promise->set_value(a);
```
arvidn, line 8718: passing the pointer to the alert to the future is risky in the same way as the exception. Making sure that the client gets the future's value before calling `pop_alerts()` is not trivial and, I would expect, a source of errors.

arvidn (commented):
I would think a much simpler approach would be to add `client_data_t` to the alerts, and accept an optional `client_data_t` when making the calls.

That way, at least the alert can be associated with the corresponding call in a simple and cheap way. the `client_data_t` object could even point to a promise object to be invoked.

Although, having the promise be invoked directly by the main thread would save time in synchronizing the threads. It would be possible to do call the promise in a plugin, but it wouldn't be super elegant

glassez:
> I would think a much simpler approach would be to add `client_data_t` to the alerts, and accept an optional `client_data_t` when making the calls.

Offtopic:
Wouldn't it be a good idea to switch to use `std::any` instead of `client_data_t` (in future update)?

arvidn:
`std::any` has higher overhead and also requires RTTI, which some people like to disable

glassez:
Doesn't seem like it actually requires RTTI (except maybe `type()` method which can be easily avoided from being used). There are several confirmations of it on the Stackoverflow etc.
As for possible overhead I don't believe it is so significant that it could play much role in our scenarios.
But we could get advantages in the form of the ability to pass types by value including managed pointers like `unique_ptr`/`shared_ptr`.

PR #7559, opened 2023-11-19
Backport `file_renamed_alert::old_name()` to RC_1_2 branch

glassez:


arvidn:
I'm pretty sure this will break the ABI. i.e. a client built against this patch will break when linked against the current 1.2.x library, and vice versa

glassez:
>a client built against this patch will break when linked against the current 1.2.x library

Isn't it expected that an application using some new features of library is incompatible with a previous version of the library that does not have these features (unless it specifically takes care of compatibility with previous versions of the library by conditionally disabling support for these features)?
>vice versa

I think we should only talk about the opposite case, when the updated library remains compatible with applications linked to its previous versions. So what exactly do you think could break existing applications? It doesn't change existing thing and just something new.

glassez:
BTW, as for ABI compatibility.
It seems that we have come across some gap in the existing preconditions, which may negate all your efforts about it. Judging by some recent Issues with qBittorrent binaries distributed via some Linux distros, a simple rebuild of the same libtorrent version using a different Boost version can lead to an ABI compatibility break.

xavier2k6:
@arvidn going forward is focus primarily being given to RC_2_0 & newer branches??

PR #7569, opened 2023-12-05
Adds pread_disk_io.cpp and pread_storage.cpp to CMakeList sources

vktr:
Adding the pread sources to CMakeLists.txt to get the build working with CMake (i.e vcpkg).


arvidn:
I've incorporated this in the branch. thanks!

PR #7574, opened 2023-12-29
fix assert in uTP resend_packet()

arvidn:


PriitUring:
Backport it to RC_1_2?

PR #7576, opened 2024-01-03
Implement torrent synchronization and automatic notification on torrent alerts

joriscarrier:


arvidn:
what's the rationale for this change? it's not obvious from the documentation you added.
it looks like you also meant to expose the `wait` functions to `torrent_handle` as well, otherwise they're not usable by the client.

But more fundamentally, I believe this kind of mechanism would be implemented either by a plugin or by a `set_alert_notify()` function. It's not obvious why this would need to be built into libtorrent.

arvidn:
if you want a hook on when a torrent changes state, it would probably be better to do something in `torrent::set_state()`. Going via the alert mechanism won't work if the client clear the `alert_category::status` bit in the alert mask.

arvidn:
I also think there's a race condition in here. For example, if you call `handle.status()` followed by `handle.wait()`, it the torrent may have changed status before you call `wait()`

PR #7582, opened 2024-01-14
Migrate away from deprecated functions in C++23

Chocobo1:
https://stackoverflow.com/questions/71828288/why-is-stdaligned-storage-to-be-deprecated-in-c23-and-what-to-use-instead

### DIFF
--- a/include/libtorrent/aux_/aligned_storage.hpp
+++ b/include/libtorrent/aux_/aligned_storage.hpp
@@ -37,9 +37,10 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace libtorrent { namespace aux {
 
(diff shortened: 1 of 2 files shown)


Chocobo1:
@arvidn 
Please merge it into `RC_2_0` and `master`. They are affected too.
Also can we expect a new maintenance release for `RC_1_2`? There are some nice small fixes accumulated.

arvidn:
thanks! use of `aligned_union` and `aligned_storage` was removed in `master` a while ago. This still applies to `RC_1_2` and `RC_2_0` though

arvidn (approved):

Chocobo1:
>use of aligned_union and aligned_storage was removed in master a while ago. 

But I see you still have aligned_storage in master: https://github.com/arvidn/libtorrent/blob/master/include/libtorrent/aux_/aligned_storage.hpp

PR #7584, opened 2024-01-14
migrate awy from aligned_storage

arvidn:
forward-port of https://github.com/arvidn/libtorrent/pull/7582


glassez:
There is a typo in commit message: awy -> away.

PR #7594, opened 2024-01-30
Restore file priorities in POSIX storage

### LINKED ISSUE #7593: Partial downloads misbehave in POSIX-compliant mode
Partial downloads misbehave in POSIX-compliant mode. This happens because POSIX storage do not restore file priorities for some reason.

I discovered this bug using qBittorent, repro details are described in https://github.com/qbittorrent/qBittorrent/issues/20351

The fix seems to be simple:

```patch
diff --git a/src/posix_storage.cpp b/src/posix_storage.cpp
index 4ce84cf..cfed5da 100644
--- a/src/posix_storage.cpp
+++ b/src/posix_storage.cpp
@@ -61,6 +61,7 @@ namespace aux {
 	posix_storage::posix_storage(storage_params const& p)
 		: m_files(p.files)
 		, m_save_path(p.path)
+		, m_file_priority(p.priorities)
 		, m_part_file_name("." + to_hex(p.info_hash) + ".parts")
 	{
 		if (p.mapped_files) m_mapped_files.reset(new file_storage(*p.mapped_files));
```

HanabishiRecca:
POSIX storage do not restore file priorities for some reason, which leads to bugs with partial downloads.

Fixes #7593

### DIFF
--- a/src/posix_storage.cpp
+++ b/src/posix_storage.cpp
@@ -61,6 +61,7 @@ namespace aux {
 	posix_storage::posix_storage(storage_params const& p)
 		: m_files(p.files)
 		, m_save_path(p.path)
+		, m_file_priority(p.priorities)
 		, m_part_file_name("." + to_hex(p.info_hash) + ".parts")
 	{
 		if (p.mapped_files) m_mapped_files.reset(new file_storage(*p.mapped_files));


arvidn (approved):
thanks, good find!

PR #7596, opened 2024-01-30
Fix the wrong port is being announced to tracker

### LINKED ISSUE #7595: Wrong port is announced to tracker
libtorrent version (or branch): 2.0.x

https://github.com/qbittorrent/qBittorrent/issues/20006

When custom port is mapped using UPnP it overwrites the session one so it is announced to tracker.
This problem is quite serious for users, since some trackers block clients based on libtorrent 2.0 because of it.

glassez:
Closes #7595.

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -5697,8 +5697,11 @@ namespace {
 			listen_socket->external_address.cast_vote(external_ip, source_router, address());
 		}
 
-		if (proto == portmap_protocol::tcp) listen_socket->tcp_port_mapping[transport].port = port;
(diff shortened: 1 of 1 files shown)


tainewoo:
Thank you so much.
This issue blocks lots of users to move on to libtorrent v2.x.
With this fix, lots of them might try to move on to the newer versions.

Pentaphon:
@arvidn please make this a priority as this is a major issue according to many tracker users in the qBittorrent community who use private trackers.

arvidn (approved):
good find!

glassez:
@arvidn 
While dealing with this problem, I found that portmap alert message does not contain the source port (this information is generally not available through the alert interface), and this is an essential part of the information about established port mapping.

PR #7597, opened 2024-01-30
Reject resume data if it contains mismatching info hashes

glassez:
Otherwise, successfully loaded/parsed resume data could lead to an error when adding a torrent to session.

### DIFF
--- a/src/read_resume_data.cpp
+++ b/src/read_resume_data.cpp
@@ -135,6 +135,11 @@ namespace {
 				{
 					ec = err;
 				}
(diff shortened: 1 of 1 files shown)


arvidn (approved):
looks good. is there a way for this to happen without someone (maliciously) creates a corrupt resume data file?

arvidn:
I want to add a test as well. I'll land this once I have that ready

glassez:
> is there a way for this to happen without someone (maliciously) creates a corrupt resume data file?

In fact, I have not encountered exactly such a problem. I came across this code while dealing with a similar problem that has a different source.
But still, if the format allows for the possibility of containing inconsistent data, then it is better to have such a check when parsing it.

PR #7602, opened 2024-02-03
default built to require windows 10

xavier2k6:


src/drive_info.cpp
```diff
@@ -243,7 +243,7 @@ drive_info get_drive_info(std::string const& path)
 		if (dev.handle() == INVALID_HANDLE_VALUE)
 			continue;
 
-#if _WIN32_WINNT >= 0x601
```
arvidn, line 246: I don't understand this change. Is there anything in this block that requires windows 10?

arvidn (approved):

xavier2k6:
BTW, Boost are starting to `default` their target API's to Windows 10 as of [Boost 1.84.0](https://www.boost.org/users/history/version_1_84_0.html)

arvidn:
it appears that having the build system define `_WIN32_WINNT` will make boost.asio think it's building on windows, and try to include `<winapifamily.h>`.

arvidn:
here's my attempt: https://github.com/arvidn/libtorrent/pull/7662

PR #7610, opened 2024-02-17
add test coverage for read_resume_data()

arvidn:
when v1 or v2 hashes do not match the torrent metadata

### DIFF
--- a/src/read_resume_data.cpp
+++ b/src/read_resume_data.cpp
@@ -135,11 +135,6 @@ namespace {
 				{
 					ec = err;
 				}
-				else if (((ret.info_hashes.has_v1() && ret.info_hashes.v1 != ret.ti->info_hashes().v1)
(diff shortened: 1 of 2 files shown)


src/read_resume_data.cpp
```diff
@@ -135,11 +135,6 @@ namespace {
 				{
 					ec = err;
 				}
-				else if (((ret.info_hashes.has_v1() && ret.info_hashes.v1 != ret.ti->info_hashes().v1)
-					|| (ret.info_hashes.has_v2() && ret.info_hashes.v2 != ret.ti->info_hashes().v2)))
```
arvidn, line 139: @glassez this is the same check as on line 128 (but inverted), so I moved the block to its else-block

glassez: :+1:
I didn't even notice that info hash is calculated before torrent info is parsed.

PR #7615, opened 2024-02-18
Clean up MSVC compiler flags

Chocobo1:
* Specify standards conformance mode to MSVC
  https://learn.microsoft.com/en-us/cpp/build/reference/permissive-standards-conformance?view=msvc-170
* Remove unsuitable compiler flag
  Control of multi-core compilation should be set at the build tool (e.g. MSBuild, ninja, make) and not as a compiler flag.
  Also cmake already append it automatically:
  https://github.com/Kitware/CMake/blob/e75afeae5cc1b41e1fb97e1924e6dbcd1e39d91a/CompileFlags.cmake#L118-L134

This patch is for RC_1_2 and it should also be applied to RC_2_0 and master branch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -612,14 +612,14 @@ if (WIN32)
 		)
 		target_compile_options(torrent-rasterbar
 			PRIVATE
+				# allow larger .obj files (with more sections)
+				/bigobj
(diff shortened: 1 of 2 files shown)


arvidn:
forward-port here: https://github.com/arvidn/libtorrent/pull/7616

arvidn (approved):

PR #7619, opened 2024-02-19
bump version to 2.0.10

arvidn:


glassez:
@arvidn
BTW, we once discussed the issue that it is much more convenient for application developers you change the version in sources immediately after the release, rather than immediately before. That is, as soon as you release version 2.0.10, the sources should be changed to contain the next one.

arvidn:
https://github.com/arvidn/libtorrent/pull/7621

PR #7640, opened 2024-02-28
try encode web seed redirect location

ajax16384:
web server may redirect with non encoded paths (like "Location: http://newserver.net/with space/file.bin" instead of  "Location: http://newserver.net/with%20space/file.bin")
let's try encode such values as curl or wget

### DIFF
--- a/src/web_peer_connection.cpp
+++ b/src/web_peer_connection.cpp
@@ -666,6 +666,7 @@ void web_peer_connection::handle_redirect(int const bytes_left)
 		file_index_t const file_index = m_file_requests.front().file_index;
 
 		location = aux::resolve_redirect_location(m_url, location);
+		location = maybe_url_encode(location);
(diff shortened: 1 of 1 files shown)


src/web_peer_connection.cpp
```diff
@@ -666,6 +666,7 @@ void web_peer_connection::handle_redirect(int const bytes_left)
 		file_index_t const file_index = m_file_requests.front().file_index;
 
 		location = aux::resolve_redirect_location(m_url, location);
+		location = maybe_url_encode(location);
```
arvidn, line 669: I think there should be a comment here explaining that this is a work-around for buggy HTTP servers, that fail to encode the redirect URL.

arvidn (commented):
It would be good to have a test exercising this case as well.
you can add a new or extend an existing test for redirects here:
`simulation/test_web_seed.cpp`

arvidn:
re-running CI

PR #7642, opened 2024-03-02
Send read_piece_alert after piece passes

### LINKED ISSUE #7634: Reading a piece using set_piece_deadline fails because it's not finished yet

libtorrent version (or branch): `RC_2_0`

platform/architecture: Linux 6.7.5 x86\_64

compiler and compiler version: Clang 16.0.6

I'm developing a [GStreamer](https://gstreamer.freedesktop.org) BitTorrent streaming plugin based on libtorrent, and the implementation is as follows.

- Setup a session and a handle.
- Launch a task thread designated to handle alerts.
- Send `read_piece` requests one at a time using `torrent_handle::set_piece_deadline` with `alert_when_available` (I'll implement queuing later).
- Get piece data from the task thread.

The problem is that when `read_piece_alert` arrives it does it _before_ a piece is done checking, so it's going to be an error even if `disable_hash_checks` is set to `true`. I don't know yet how this happens, if it's either a data race or something wrong with the library logic. I can post more info if needed.

YakoYakoYokuYoku:
Fixes #7634. Per comment https://github.com/arvidn/libtorrent/issues/7634#issuecomment-1962734438 I've found that `lt::torrent::remove_time_critical_piece` gets called before the piece is set to having passed the check by `lt::torrent::we_have`, this change places the former to after the latter in `lt::torrent::piece_passed`.

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -4412,10 +4412,6 @@ namespace {
 
 		inc_stats_counter(counters::num_piece_passed);
 
-#ifndef TORRENT_DISABLE_STREAMING
-		remove_time_critical_piece(index, true);
-#endif
-
 		if (settings().get_int(settings_pack::suggest_mode)
 			== settings_pack::suggest_read_cache)
 		{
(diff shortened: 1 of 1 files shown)


arvidn (commented):
I think this is the right fix. I think there should be a test exercising the issue along with it though.

arvidn:
re-running CI

arvidn (approved):

PR #7644, opened 2024-03-10
improve const correctness of disk I/O code

arvidn:
by making readwrite() a template that can work with both span<char> and span<char const>.


glassez:
Isn't `span<char const>` constructible from `span<char>`?

arvidn:
I believe it is. the issue here is that it's being passed to a function, and that function then passes it back into a callback. So it's being "laundered" by the `readwrite()` function signature. By making it a template, it can either `char const` or `char` and the callback takes the same type

glassez:
> the issue here is that it's being passed to a function, and that function then passes it back into a callback.

Got it.

PR #7658, opened 2024-04-08
Make anti-leech choker never increase score for dishonesty

Artoria2e5:
The current anti-leech choker takes the maximum of client-reported have and our own session uploaded size as the progress of the client.  This may paradoxically inflate the score of dishonest clients if we've been uploading to it for long enough, that our uploaded size exceeds 50% of the total size. (That could in turn happen if we had nobody else to upload to for a while. That in turn screws over some honest client that then appears in our peer list, until the dishonest peer disconnects.)

This commit caps "given_size" to 50% of total_size, so that we never cause an increase of the score by taking into account the "given".

### DIFF
--- a/src/choker.cpp
+++ b/src/choker.cpp
@@ -167,7 +167,10 @@ namespace {
 
 		std::int64_t const total_size = t->torrent_file().total_size();
 		if (total_size == 0) return 0;
-		std::int64_t const have_size = std::max(peer->statistics().total_payload_upload()
+		// Cap the given_size so that it never causes the score to increase
+		std::int64_t const given_size = std::min(peer->statistics().total_payload_upload()
+			, total_size / 2);
+		std::int64_t const have_size = std::max(given_size
 			, std::int64_t(t->torrent_file().piece_length()) * peer->num_have_pieces());
 		return int(std::abs((have_size - total_size / 2) * 2000 / total_size));
 	}


arvidn:
this will penalise peers that are close to completing the torrent, but that we see for the first time, right?

Also, with this change, the comment above, with the ascii-art plot, would need to be updated. But I'm not sold on this being an improvement. The scenario you're talking about is that a "dishonest" peer is the only peer in the swarm, and you're uploading almost all of the torrent to it. In this scenario, there is no distinction between an honest and a dishonest peer.

If another peer joins, what is it that makes the old peer dishonest? If the new peer has not downloaded anything, that peer would have a higher anti-leech score than the old one (which seems right).

arvidn:
If you want to detect dishonesty, you could compare the number of bytes uploaded to the number of pieces the peer claims to have. But that's also risky; earlier versions of libtorrent would not send HAVE messages to peers that already have those pieces (as they are redundant from the download-protocol point of view).

Artoria2e5:
> this will penalise peers that are close to completing the torrent, but that we see for the first time, right?
> 
> 

It will not.

Let's say the total is 100 blocks worth of things and the client reports having 99. Our known uploaded amount is 0, so we take min(0, 50) which is still 0. We then take max(0, 99) for the amount that counts toward scoring, and that would be still 99.

arvidn:
I see how this works now. it makes sense

PR #7662, opened 2024-04-21
default target windows 10

arvidn:


arvidn:
@xavier2k6 I don't have a windows machine at the moment. Would you mind making sure this works as expected? (especially with `b2 windows-version=win7`)

xavier2k6:
I didn't have time to test with/compile for/on windows 7 but was able to compile qBittorrent successfully with this PR & it ran ok on Windows 10 22H2 system.

PR #7668, opened 2024-05-08
Add option to install python .dist-info.

Lord-Kamina:
This is just to make the bindings more pip-friendly when building from source.


arvidn:
it probably makes sense to enable this build option on CI as well

Lord-Kamina:
> it probably makes sense to enable this build option on CI as well

I'm gonna make a new PR soon, just straight-up replacing egg-info with dist-info. I'm not sure, however, where in the CI should this be included?

PR #7684, opened 2024-06-12
Test: run address sanitizer with Boost 1.85.0 on windows in CI

xavier2k6:


arvidn:
I suspect things might start working if we remove the `choco install openssl` step, and rely on the openssl already installed on the runner.

xavier2k6:
> #7685

Implemented those OpenSSL changes & as you can see with Boost 1.85.0, sanitizer & MingW builds are still failing
Example: [-Wstringop-overflow=] & [-Warray-bounds]

Cirrus CI is failing to locate boost for macos arm testing.

PR #7691, opened 2024-06-27
Moved type stubs into libtorrent directory

### LINKED ISSUE #7690: Move Python type stubs into `bindings/python/libtorrent`
**libtorrent version (or branch):** master

While creating the fixes of #7688, I spotted one optimization to the current directory structure: the Python type stubs can be moved from `bindings/python/install_data/libtorrent` into `bindings/python/libtorrent`.

This change would obsolete the following code in its entirety.

https://github.com/arvidn/libtorrent/blob/7c698d82636e667ad9fe80f629fd85bf7ef4314f/bindings/python/setup.py#L506-L513

Even though the code above can be removed, all paths that point to `install_data` would have to be updated.

qstokkink:
Fixes #7690

This new directory structure should get rid of some code.


qstokkink:
@arvidn I thought this PR could cause terrible issues but everything seems fine. Ready for merge 👍

arvidn (approved):

arvidn:
thanks!

PR #7699, opened 2024-07-01
GHA workflow update

arvidn:


xavier2k6:
@arvidn for pre-commit failure, quote the actual python version -> python-version: "3.10" & that should resolve failure.

CodeQL - Perhaps you could also use the CMake version that comes pre-installed on runner image rather than downloading/installing a specified version.

arvidn:
the cibuildwheel is targeting an older MacOS than what the image supports

xavier2k6:
> the cibuildwheel is targeting an older MacOS than what the image supports

You've set it to 10.9 below:

https://github.com/arvidn/libtorrent/blob/d4a9a0ce62878d101a8ad874fa4bee774e8ed13a/pyproject.toml#L21

I've been looking in to this via the macOS arm PR & newer actions which require different targets for x64/arm......I believe if the `MACOSX_DEPLOYMENT_TARGET` was not set by the user it is automatically set to the associated requirement through the cibuildwheel action.

xavier2k6:
What I meant was to **remove below line completely** & let the cibuildwheel action automatically set the correct "minimum target" 

https://github.com/arvidn/libtorrent/blob/d4a9a0ce62878d101a8ad874fa4bee774e8ed13a/pyproject.toml#L21

Look at Line 294 onwards in relation to `MACOSX_DEPLOYMENT_TARGET`
https://github.com/pypa/cibuildwheel/blob/main/cibuildwheel/macos.py

xavier2k6:
I wondering about current failure ......is it to do with OpenSSL being installed from brew rather than being installed from source?

Issue #7692, opened 2024-06-27
`libdatachannel submodule` (`0.18.1`) is outdated - Update to latest


xavier2k6:
**Please provide the following information**

libtorrent version (or branch):
`master`
platform/architecture:
N/A
compiler and compiler version:
N/A
please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

`libdatachannel submodule` (`0.18.1`) is outdated.

Latest is `0.21.1`
https://github.com/paullouisageneau/libdatachannel/releases

@paullouisageneau Can you create the relevant PR?

paullouisageneau:
Yes, I'll update the library.

xavier2k6:
Closed via 
- #7701

PR #7701, opened 2024-07-03
Update libdatachannel to v0.21.2

paullouisageneau:
This PR updates the libdatachannel submodule to v0.21.2. Main changes are related to media transport and MbedTLS support (which are not used here), but there are also a bunch of enhancements and fixes related to data channels.

Fixes https://github.com/arvidn/libtorrent/issues/7692

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -839,8 +839,7 @@ if (webtorrent)
 		target_compile_options(datachannel-static PRIVATE
 			-Wno-pedantic
 			-Wno-unused-parameter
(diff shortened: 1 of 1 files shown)


paullouisageneau:
@arvidn As libdatachannel is now packaged in a few distributions, I'm planning on removing the submodule in the long term. Would it be OK to depend on an external library for WebTorrent instead?

paullouisageneau:
I fixed the Jamfile build issue with MSVC, all good now ✅

PR #7708, opened 2024-07-17
announce port=1025 instead of port=1, when there is no listen port

Aurumh4x:
slight update from https://github.com/arvidn/libtorrent/pull/4400

this was causing issues with certain trackers which block advertised listen ports less than 1024 - https://github.com/HDInnovations/UNIT3D-Community-Edition/pull/3329

this was causing bugs with using libtorrent based clients (such as qbittorrent) when using socks proxy, as seen in following issues
https://github.com/HDInnovations/UNIT3D-Community-Edition/issues/3428
https://github.com/HDInnovations/UNIT3D-Community-Edition/issues/3847

please let me know if there are any other follow on effects from this that might require changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,6 +5,7 @@
 	* fix issue where set_piece_deadline() did not correctly post read_piece_alert
 	* fix integer overflow in piece picker
 	* torrent_status::num_pieces counts pieces passed hash check, as documented
+	* announce port=1025 instead of port=1, when there is no listen port
 
 2.0.10 released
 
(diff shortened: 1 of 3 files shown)


MiM-MiM:
This is just quite a terrible idea.  Faking a new port rather than fixing the issue will only lead to trackers that liked it blocking these users getting banned.  Cheating is not allowed on most sites.

arvidn:
it does seem like a stretch to do this. Even reporting `&port=` at all seems wrong. I think it would be more palatable if there would at least be a setting to enable this (or some compatible) behavior.

PR #7710, opened 2024-07-20
add option to send host header in HTTP proxy CONNECT command

arvidn:


xavier2k6:
**Cirrus CI - Failure:**
```
pkg install -y boost-build boost-libs openssl cmake ninja py39-pip
Updating FreeBSD repository catalogue...
FreeBSD repository is up to date.
All repositories are up to date.
Updating database digests format: . done
pkg: No packages available to install matching 'py39-pip' have been found in the repositories
```

See/read below:
https://lore.kernel.org/all/CAKBkRUzf1ZWahU3WrTxfM_fHPpxVBMC=YcjHu+evKB5PWdQGaQ@mail.gmail.com/T/

bearx3f:
I've tested qBittorrent 4.6.5 with my patch (forcing send_host_in_connect to 1) but unfortunately, the issue persists.

After careful investigation, I noticed that src/http_connection.cpp#590 seems never to be executed. Additionally, the condition in include/libtorrent/http_stream.hpp#132 appears to always evaluate to false.

I believe understanding why these code sections are not being reached might be crucial in resolving the underlying issue.

arvidn:
@bearx3f are you connecting to your proxy over SSL by chance? It looks like my patch failed to take into account HTTPS proxies.

bearx3f:
No my proxy server only handles HTTP

bearx3f:
I'm trying with your test_http_connection.cpp by modify proxy_settings to ps.send_host_in_connect = true; and tracker server protocol is https but `http_stream` never call `set_host`

arvidn:
@bearx3f I landed (and fixed up) your patch. did it make it work for you?

bearx3f:
@arvidn I can not make test for full stack (test with my proxy) but I try with modified http_proxy.py and it's works so maybe my proxy should work too

PR #7714, opened 2024-07-22
check settings_pack::max_out_request_queue before alert outstanding_r…

tkittich:
…equest_limit_reached

This should fix performance warning: max outstanding piece requests reached, outstanding_request_limit_reached

### DIFF
--- a/src/peer_connection.cpp
+++ b/src/peer_connection.cpp
@@ -5103,8 +5103,7 @@ namespace libtorrent {
 
 		update_desired_queue_size();
 
-		if (m_desired_queue_size == m_max_out_request_queue
-			&& m_desired_queue_size >= m_settings.get_int(settings_pack::max_out_request_queue)
+		if (m_desired_queue_size >= m_settings.get_int(settings_pack::max_out_request_queue)
 			&& t->alerts().should_post<performance_alert>())
 		{
 			t->alerts().emplace_alert<performance_alert>(t->get_handle()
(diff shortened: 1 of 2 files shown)


arvidn:
The existing check compares against `m_max_out_request_queue` which is set to this setting, [here](https://github.com/arvidn/libtorrent/blob/RC_2_0/src/peer_connection.cpp#L141):
```
m_max_out_request_queue(aux::clamp_assign<std::uint16_t(m_settings.get_int(settings_pack::max_out_request_queue)))
```

This limit can be adjusted if the peer indicates a `reqq` limit in the extended handshake, or if the peer is snubbed (or if the peer is identified as bitcomet).

So, I think you're right to check against the setting instead. I think the existing check can be removed however.

src/peer_connection.cpp
```diff
@@ -5104,6 +5104,7 @@ namespace libtorrent {
 		update_desired_queue_size();
 
 		if (m_desired_queue_size == m_max_out_request_queue
+			&& m_desired_queue_size >= m_settings.get_int(settings_pack::max_out_request_queue)
```
arvidn, line 5107:
```suggestion
		if (m_desired_queue_size >= m_settings.get_int(settings_pack::max_out_request_queue)
```

arvidn:
If you take my suggestion, please add a note to the `Changelog` as well, that this issue was addressed.

tkittich:
> If you take my suggestion, please add a note to the `Changelog` as well, that this issue was addressed.

Thank you for the review & suggestions.

arvidn (approved):

PR #7717, opened 2024-07-27
fix issue when closing a uTP connection

arvidn:
with a close reason and picking up a nagle packet. This would corrupt the payload.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 
 2.0.11 not released
 
+	* fix uTP issue where closing the connection could corrupt the payload
 	* apply DSCP/TOS to sockets before initiating the TCP connection
 	* assume copy_file_range() exists on linux (unless old glibc)
 	* fix issue where set_piece_deadline() did not correctly post read_piece_alert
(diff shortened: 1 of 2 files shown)


Seeker2:
Is this related to or explain why uTP seeds-to-seed connections don't close quickly on BT clients connected to qBitTorrent?:
https://github.com/arvidn/libtorrent/issues/3542#issuecomment-962175985

rafi-d:
Neah, no one will yell at you ..

arvidn:
> Is this related to or explain why uTP seeds-to-seed connections don't close quickly on BT clients connected to qBitTorrent?:
> 
> https://github.com/arvidn/libtorrent/issues/3542#issuecomment-962175985

It's possible, but I haven't been able to reproduce that issue. The symptom I saw, caused by this, was "invalid unchoke message" and "packet too large".

Seeker2:
> The symptom I saw, caused by this, was "invalid unchoke message" and "packet too large".

I saw that too, but not every time for "Invalid packet length":
https://github.com/arvidn/libtorrent/issues/3542#issuecomment-1257038658
That "Invalid packet length" event happened when the qBitTorrent client reached 100% and disconnected.

PR #7724, opened 2024-08-06
Add pre-commit typos & codespell hooks

xavier2k6:
Testing for typo's etc.

example:

`verisoning` -> `versioning`
https://github.com/arvidn/libtorrent/blob/2ab8fedabb7716947edc77958da6d0b5e1040df9/Jamfile#L190


arvidn:
there are a lot of good fixes here. if only it could be made to ignore string literals. ignoring the `.torrent` files is easy enough

xavier2k6:
certain words etc. can be excluded.....will be back to it...any suggetsions on what you'd like to exclude or want to actually change?

leafs -> leaves
file_prioritity0 -> file_priority0
file_prioritity1 -> file_priority1

arvidn:
the most obvious false positives were in string literals in the c++ code, in the utf-8 test file and in .torrent files. Maybe also in some `.xml` files.

PR #7726, opened 2024-08-08
Enabling listening on tcp port when proxy is used for peer connections

Masong19hippows:
qbittorrent disables the listening tcp port when the option "Use proxy for peer connections" is enabled. This is mentioned in [#21134](https://github.com/qbittorrent/qBittorrent/issues/21134) and [#8718](https://github.com/qbittorrent/qBittorrent/issues/8718). 

I don't believe this should be the case as its a fundamental misunderstanding of what the proxy should do. Proxies are only for outbound connections and should not change behavior of incoming connections. This is outlined in the [http proxy spec](https://www.ietf.org/rfc/rfc2068.txt#:~:text=An%20intermediary%20program%20which,requirements%20of%20this%20specification.) as defining a proxy as a service for making requests. 

I think the current implementation is an artificial limitation set by libtorrent rather than the protocol itself. There are plenty of reasons to still accept incoming connections while using the proxy for outbound connections. A very reason is for networks that require a proxy for internet. 

This pull request is heavily based on [this](https://xcel.me/qbittorrent-listening-on-port-while-using-proxy/) article.

### DIFF
--- a/include/libtorrent/aux_/proxy_settings.hpp
+++ b/include/libtorrent/aux_/proxy_settings.hpp
@@ -79,6 +79,10 @@ namespace aux {
 		// This is only supported by SOCKS5 and HTTP.
 		bool proxy_hostnames = true;
 
+		// defaults to false for integration compatibility. It means that the
+		// torrent port is listening with tcp and udp while there is a proxy configured.
+		bool listen_on_proxy = false;
+
 		// determines whether or not to exempt peer and web seed connections
 		// from using the proxy. This defaults to true, i.e. peer connections are
 		// proxied by default.
--- a/include/libtorrent/aux_/proxy_settings.hpp
+++ b/include/libtorrent/aux_/proxy_settings.hpp
@@ -81,7 +81,7 @@ namespace aux {
 
 		// defaults to false for integration compatibility. It means that the
 		// torrent port is listening with tcp and udp while there is a proxy configured.
-		bool listen_on_proxy = false;
+		bool proxy_accept_incoming = false;
 
 		// determines whether or not to exempt peer and web seed connections
 		// from using the proxy. This defaults to true, i.e. peer connections are
--- a/include/libtorrent/aux_/proxy_settings.hpp
+++ b/include/libtorrent/aux_/proxy_settings.hpp
@@ -86,6 +86,10 @@ namespace aux {
 
 		// if true, tracker connections are subject to the proxy settings
 		bool proxy_tracker_connections = true;
+
+		// when proxying non plain HTTP over an HTTP proxy (SSL or peer
+		// connections), also tell the proxy which hostname we're connecting to
+		bool send_host_in_connect = false;
 	};
 
 }}
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -880,9 +880,6 @@ namespace aux {
 			// any). This is only supported by SOCKS5 and HTTP.
 			proxy_hostnames,
 
-			// if true, listening on the defined port will be enabled while using proxy
-			proxy_accept_incoming,
-
 			// if true, peer connections are made (and accepted) over the
 			// configured proxy, if any. Web seeds as well as regular bittorrent
 			// peer connections are considered "peer connections". Anything
@@ -1029,7 +1026,19 @@ namespace aux {
 			// man-in-the-middle connections.
 			proxy_send_host_in_connect,
 
-			max_bool_setting_internal
+			// When using HTTP proxy (in proxy_type), libtorrent will connect
+			// to peers and trackers using the `CONNECT` proxy command. In this
+			// command it's possible to reveal the hostname of the server we're
+			// connecting to. When this option is true, the hostname will be
+			// sent. This feature can be useful if the proxy is used to
+			// man-in-the-middle connections.
+			proxy_send_host_in_connect,
+
+			max_bool_setting_internal,
+
+			// if true, listening on the defined port will be enabled while using proxy
+			proxy_accept_incoming
+
 		};
 
 		// hidden
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -881,7 +881,7 @@ namespace aux {
 			proxy_hostnames,
 
 			// if true, listening on the defined port will be enabled while using proxy
-			listen_on_proxy,
+			proxy_accept_incoming,
 
 			// if true, peer connections are made (and accepted) over the
 			// configured proxy, if any. Web seeds as well as regular bittorrent
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -1949,8 +1949,10 @@ namespace aux {
 			max_web_seed_connections,
 
 			// the number of seconds before the internal host name resolver
-			// considers a cache value timed out, negative values are interpreted
-			// as zero.
+			// considers a cache value out of date, and removes it. Negative
+			// values are interpreted as zero. The host name cache also caches
+			// failed DNS lookups, but only for 1/8th of the time of this
+			// timeout.
 			resolver_cache_timeout,
 
 			// specify the not-sent low watermark for socket send buffers. This
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -2086,6 +2086,10 @@ namespace aux {
 			//    instead of the actual local listening port.
 			announce_port,
 
+			// Configures the variance for I2P inbound and outbound tunnel lengths [-7..7]
+			i2p_inbound_length_variance,
+			i2p_outbound_length_variance,
+
 			max_int_setting_internal
 		};
 
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -880,6 +880,9 @@ namespace aux {
 			// any). This is only supported by SOCKS5 and HTTP.
 			proxy_hostnames,
 
+			// if true, listening on the defined port will be enabled while using proxy
+			listen_on_proxy,
+
 			// if true, peer connections are made (and accepted) over the
 			// configured proxy, if any. Web seeds as well as regular bittorrent
 			// peer connections are considered "peer connections". Anything
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -1018,6 +1018,14 @@ namespace aux {
 			// protocol may not be valid from the proxy's point of view.
 			socks5_udp_send_local_ep,
 
+			// When using HTTP proxy (in proxy_type), libtorrent will connect
+			// to peers and trackers using the `CONNECT` proxy command. In this
+			// command it's possible to reveal the hostname of the server we're
+			// connecting to. When this option is true, the hostname will be
+			// sent. This feature can be useful if the proxy is used to
+			// man-in-the-middle connections.
+			proxy_send_host_in_connect,
+
 			max_bool_setting_internal
 		};
 
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -4874,9 +4874,9 @@ namespace {
 		return ret;
 	}
 
-	torrent_handle session_impl::find_torrent_handle(sha1_hash const& info_hash)
+	torrent_handle session_impl::find_torrent_handle(info_hash_t const& info_hash)
 	{
-		return torrent_handle(find_torrent(info_hash_t(info_hash)));
+		return torrent_handle(find_torrent(info_hash));
 	}
 
 	void session_impl::async_add_torrent(add_torrent_params* params)
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -830,8 +830,8 @@ bool ssl_server_name_callback(ssl::stream_handle_type stream_handle, std::string
 					if (val) s.set_int(settings_pack::proxy_type, int(val.int_value()));
 					val = settings.dict_find_int("proxy_hostnames");
 					if (val) s.set_bool(settings_pack::proxy_hostnames, val.int_value() != 0);
-					val = settings.dict_find_int("listen_on_proxy");
-					if (val) s.set_bool(settings_pack::listen_on_proxy, val.int_value() != 0);
+					val = settings.dict_find_int("proxy_accept_incoming");
+					if (val) s.set_bool(settings_pack::proxy_accept_incoming, val.int_value() != 0);
 					val = settings.dict_find_int("proxy_peer_connections");
 					if (val) s.set_bool(settings_pack::proxy_peer_connections, val.int_value() != 0);
 					val = settings.dict_find_string("hostname");
@@ -2061,7 +2061,7 @@ namespace {
 		// proxies
 		if (m_settings.get_int(settings_pack::proxy_type) != settings_pack::none
 			&& m_settings.get_bool(settings_pack::proxy_peer_connections)
-			&& !m_settings.get_bool(settings_pack::listen_on_proxy))
+			&& !m_settings.get_bool(settings_pack::proxy_accept_incoming))
 		{
 			// we will be able to accept incoming connections over UDP. so use
 			// one of the ports the user specified to use a consistent port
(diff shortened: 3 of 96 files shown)


arvidn:
@Masong19hippows that link appears to be broken. Is that text posted anywhere else?

src/session_impl.cpp
```diff
@@ -5661,21 +5633,6 @@ namespace {
...
-		{
-			// the natpmp constructor may fail and call the callbacks
-			// into the session_impl.
-			s->natpmp_mapper = std::make_shared<natpmp>(m_io_context, *this, listen_socket_handle(s));
-			ip_interface ip;
-			ip.interface_address = s->local_endpoint.address();
-			ip.netmask = s->netmask;
-			std::strncpy(ip.name, s->device.c_str(), sizeof(ip.name) - 1);
-			ip.name[sizeof(ip.name) - 1] = '\0';
-			s->natpmp_mapper->start(ip);
```
arvidn, line 5677: it doesn't make any sense to disable NAT-PMP entirely. this seems unrelated to the intention of this patch

Masong19hippows:
> @Masong19hippows that link appears to be broken. Is that text posted anywhere else?

Unfortunately, it doesn't look like it. However, it still cached on my machine via google amp. I just coped the text article and here it is.

Patch qBittorrent: Listening on port while using proxy
Brandon - 05 11月 2022

First things first, to achieve that, we need a proxy where we can set up port forwarding on it, then we modify and compile qbittorrent ourselves. A VPN can work, but that requires UPnP/PMP support of the VPN, which many lacks in. And also, use with caution, if you do not want to expose your IP address accidentally, always use a VPN.

Recently, I came across a situation where I need to do some seeding, but the network I was on requires a proxy to gain access to the Internet. Naively, I set up the proxy settings in qbittorrent, quickly found out that it did not work at all.

After some long nights searching and digging, it turned out that the library qbittorrent (along with several other FOSS torrenting software) depends on, libtorrent, on some recent versions, completely disables port listening while Use proxy for peer connections is enabled, to avoid network traffic leakage, rendering it impossible to do seeding at all while using a proxy. But the authors never make this clear, and I suspect that qbittorrent authors haven't even realized this behavior. The setting page on qbittorrent doesn't reflect the change in port listening section at all when you enable proxy for peers.

So here come the easier parts. We find the corresponding conditional code, modify, patch, then recompile libtorrent. No changes in qbittorrent are needed, only libtorrent. Check out my patch file listen-when-proxy-on.patch based on v2.0.7 of libtorrent at end of this article.

Since I'm on archlinux (and yes, BTW, I'm using arch :D), patching software is relatively convenient. Go grab the PKGBUILD script from arch offical package page https://archlinux.org/packages/extra/x86_64/libtorrent-rasterbar/ (source files link), add our patch into prepare() section like this:

prepare() {
    makedir -p build
    cd "$pkgname-$pkgver"
    patch --forward --strip=1 --input="${srcdir}/listen-when-proxy-on.patch"
}
and do makepkg -Acs to compile the package. After it's done, sudo pacman -U libtorrent... to do a force install. Then we need to do some config stuff.

Fire up qbittorrent and set a fixed listening port, say 7890;
Then set up port forwarding on the proxy server we're gonna use. I'm using FRP for port forwarding on my proxy server, so I set up a tcp reverse tunnel with local_port as 7890 on my computer;
Go to qbittorrent advanced settings page, enable Allow multiple connections from the same IP address;
Finally, set proxy server for qbittorrent, and enable proxy for peer connections. Do not turn on UPnP/NAT-PMP. But if you know what you're doing, the patch ensures that they can be turned on along with proxy.
Now we can add a torrent and see, it should start listening for incoming connections from the reverse tunnel we just set on the proxy server. And we can also see all the incoming peers shown as localhost or 127.0.0.1. No worries, the bittorrent protocol doesn't care about peers' ip addresses, from everyone else's perspective, you're just running qbittorrent on 7890 on the proxy server, everything works fine.

Disclaimer: qBittorrent and libtorrent are FOSS, the patch shared here is also free from restrictions. Please comply with your local regulations, use with caution and at your own risks.

listen-when-proxy-on.patch:

diff --git a/src/session_impl.cpp b/src/session_impl.cpp
index e25434926..8f38a3344 100644
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -2016,7 +2016,7 @@ namespace {
 
   // if we don't proxy peer connections, don't apply the special logic for
   // proxies
- if (m_settings.get_int(settings_pack::proxy_type) != settings_pack::none
+ if (false && m_settings.get_int(settings_pack::proxy_type) != settings_pack::none
    && m_settings.get_bool(settings_pack::proxy_peer_connections))
   {
    // we will be able to accept incoming connections over UDP. so use
@@ -2774,7 +2774,7 @@ namespace {
   // directly.
   // This path is only for accepting incoming TCP sockets. The udp_socket
   // class also restricts incoming packets based on proxy settings.
- if (m_settings.get_int(settings_pack::proxy_type) != settings_pack::none
+ if (false && m_settings.get_int(settings_pack::proxy_type) != settings_pack::none
    && m_settings.get_bool(settings_pack::proxy_peer_connections))
    return;
 
@@ -5473,7 +5473,7 @@ namespace {
    // connections. Not even uTP connections via the port we know about.
    // The DHT may use the implied port to make it work, but the port we
    // announce here has no relevance for that.
- if (sock->flags & listen_socket_t::proxy)
+ if (false && sock->flags & listen_socket_t::proxy)
     return 0;
 
    if (!(sock->flags & listen_socket_t::accept_incoming))
@@ -5513,7 +5513,7 @@ namespace {
    return std::uint16_t(sock->tcp_external_port());
   }
 
- if (m_settings.get_int(settings_pack::proxy_type) != settings_pack::none
+ if (false && m_settings.get_int(settings_pack::proxy_type) != settings_pack::none
    && m_settings.get_bool(settings_pack::proxy_peer_connections))
    return 0;
 
@@ -5611,7 +5611,7 @@ namespace {
 
   if (!s->natpmp_mapper
    && !(s->flags & listen_socket_t::local_network)
- && !(s->flags & listen_socket_t::proxy))
+ && !(s->flags & listen_socket_t::proxy && false))
   {
    // the natpmp constructor may fail and call the callbacks
    // into the session_impl.
@@ -6833,8 +6833,9 @@ namespace {
   // there's no point in starting the UPnP mapper for a network that isn't
   // connected to the internet. The whole point is to forward ports through
   // the gateway
+ // Enable UPnP anyway, even when on proxy, we can deal with this ourselves.
   if ((s->flags & listen_socket_t::local_network)
- || (s->flags & listen_socket_t::proxy))
+ || (s->flags & listen_socket_t::proxy && false))
    return;
 
   if (!s->upnp_mapper)
小本本
✏️📖🖥️

arvidn (commented):
one possible way to implement this would be to add a new proxy type, like `http_with_local_listen`, or something like that. It would behave like `http` but also accept incoming connections

Masong19hippows:
>The Socks5 protocol (which is also called "proxy") _does_ in fact support incoming connections, both over TCP and UDP.

It does not. The socks4 protocol specifies bind and connect commands while the socks5 protocol specifies auth and the udp communication commands Bind in the socks4 protocol is defined as:

"The client connects to the SOCKS server and sends a BIND request when
it wants to prepare for an inbound connection from an application server.
This should only happen after a primary connection to the application". 

and in the socks5 spec:

"It is expected that the client side of an application protocol will
   use the BIND request only to establish secondary connections after a
   primary connection is established using CONNECT"

This is not facilitating inbound connections in the traditional sense, rather it facilitates inbound connections on an already existing outbound connection. In the spec, it used ftp as an example, and I think this is perfect because ftp needs at least 2 connections to the client, one inbound and one outbound. I think [this](https://stackoverflow.com/questions/7731870/socks-4-bind-request-explanation) explains it better than I could. If we were to hack the bind command in order to listen and forward ports, the qbittorrent host would first need to send an outbound connect command to all potential leachers. This would require massive changes and its more hacky than it should be. 

There are applications that extend/break the socks protocol (like ssh) to include the support of inbound connections. However, this is not something outlined in the socks protocol itself, and so it shouldn't be treated as such in my opinion. 
 
> I wouldn't consider that text a compelling argument to change libtorrents behavior. I think enumerating the legitimate use cases might be.

I was more just trying to point out that the proxy server itself should not facilitate inbound connections, because this was used in the past issues to justify libtorrent behavior. 

I have been using this PR with qbittorrent for awhile now and I honestly haven't noticed any issues. Before, the seeding wasn't working well. As for the use cases, I see this as removing an artificial limitation rather than adding a feature. The use case would be that the user would get full seeding/leaching potential. It honestly works so well that I had to limit the number of inbound connections because it overloaded my socks5 server and used all of the ports on my system.
 
> Either way, if there is a compelling argument, I think it would be best to guard it under a new setting, rather than changing behavior for everyone.

That makes sense. I just did a quick and dirty thing to try and get the conversation started, but I am willing to learn and expand this.

src/session_impl.cpp
Masong19hippows: I think this was just a mistake on my end. It was not my intention to disable nat-pmp entirely. Rather, it was to enable the ability to use it regardless of proxy.

Masong19hippows:
> We can setup reserve proxy/port forwarding on the same server with http/socks proxy.  
> That would be equivalent to wireguard+port forwarding setup, if port listening is not disabled.  

I'm unsure what you mean. Can you please elaborate? 

I get setting up a reverse proxy type thing  and forwarding the port to the socks server, but what would the socks server do with it once it gets that traffic? Socks servers do not accept inbound traffic to forward to a host unless otherwise defined by the socks bind protocol.

If I'm getting what you are putting down, I am actually already doing this right now. I am using the patch above to listen to the port while using a proxy, and then using socat to forward the port from the socks server to the qbittorrent host. 

Qbittorrent should not have anything to do with a reverse proxy though, because the socks host and the qbittorrent host might be on separate machines. Any way you look at it, you can't inheritly forward ports from the socks host to the qbittorrent host with just the socks protocol.

anonbaby:
Your setup is exactly what I mean.
I saw your discussions but I didn't understand why not just patch lt (keep listening with proxy set) and let other tools take care of port forwarding? (e.g. iptables/socat/frp/...)


________________________________

Masong19hippows:
@arvidn can you please review again? I changed this so that there is now a new setting to enable listening while using a proxy. The default value for this is false to maintain compatibility between versions. When this value is false, it follows the original logic before the PR. 

I also branched and made changes to qbittorrent in order to test these changes. This can be found in the listenOnProxy branch at https://github.com/Masong19hippows/qBittorrent. There are some bugs that I need to fix and I need to figure out the translations, but it should provide a good enough platform for testing the change. The most notable issue with the qbittorrent branch is that if you change the new setting, you will need to restart qbittorrent in order for it to be applied.

arvidn:
This test is ensuring some aspects of ABI stability and is failing. https://ci.appveyor.com/project/arvidn/libtorrent/builds/50811938/job/l22u63al4sr0ykyn#L1156

the `listen_on_proxy` enum value has to be added at the end of the enums, to ensure that all existing ones are unchanged.

include/libtorrent/settings_pack.hpp
```diff
@@ -876,6 +876,9 @@ namespace aux {
 			// any). This is only supported by SOCKS5 and HTTP.
 			proxy_hostnames,
 
+			// if true, listening on the defined port will be enabled while using proxy
+			listen_on_proxy,
```
arvidn, line 880: this enum value has to be added at the end, to prevent existing constants to change value.
I'm not super happy about this name. all other proxy settings have the `proxy_` prefix, maybe something like `proxy_accept_incoming`?

include/libtorrent/aux_/proxy_settings.hpp
```diff
@@ -79,6 +79,10 @@ namespace aux {
 		// This is only supported by SOCKS5 and HTTP.
 		bool proxy_hostnames = true;
 
+		// defaults to false for integration compatibility. It means that the
+		// torrent port is listening with tcp and udp while there is a proxy configured.
+		bool listen_on_proxy = false;
```
arvidn, line 84: This name seem to suggest we're listening for connections from the *proxy*. I have a feeling that `accept_incoming` would be a better name for this.

src/session_impl.cpp
```diff
@@ -6891,7 +6896,7 @@ namespace {
 		// connected to the internet. The whole point is to forward ports through
 		// the gateway
 		if ((s->flags & listen_socket_t::local_network)
-			|| (s->flags & listen_socket_t::proxy))
+			|| (s->flags & !m_settings.get_bool(settings_pack::listen_on_proxy)))
```
arvidn, line 6899: This is not right. the `&` is a bit-wise AND operation. you can mask the `flags` field against flags/bits set in this field. `get_bool()` returns a `bool`, not a flag value. I think you meant to preserve the existing check and add this additional one.

src/session_impl.cpp
```diff
@@ -5664,7 +5669,7 @@ namespace {
 
 		if (!s->natpmp_mapper
 			&& !(s->flags & listen_socket_t::local_network)
-			&& !(s->flags & listen_socket_t::proxy))
+			&& m_settings.get_bool(settings_pack::listen_on_proxy))
```
arvidn, line 5672: if you don't configure a proxy and leave `listen_on_proxy` as `false`; it looks like this change would disable NAT-PMP in that case. I think what you meant to do was to add another condition to the `flags & proxy` check.

src/session_impl.cpp
```diff
@@ -5566,9 +5574,6 @@ namespace {
 			return std::uint16_t(sock->tcp_external_port());
 		}
 
-		if (m_settings.get_int(settings_pack::proxy_type) != settings_pack::none
-			&& m_settings.get_bool(settings_pack::proxy_peer_connections))
-			return 0;
```
arvidn, line 5571: I would have expected this also depend on the `listen_on_proxy` setting, no?

src/session_impl.cpp
```diff
@@ -5522,11 +5525,11 @@ namespace {
 		if (m_listen_sockets.empty()) return 0;
 		if (sock)
 		{
-			// if we're using a proxy, we won't be able to accept any TCP
-			// connections. Not even uTP connections via the port we know about.
```
arvidn, line 5526: I think it would be better to update or extend this comment, rather than removing it

src/session_impl.cpp
arvidn: I think there might be a simpler way of doing this. Rather than checking the settings here, I think you should be able to check the `listen_socket_t::accept_incoming` flag. You'd have to make sure this flag is set correctly depending on the `listen_on_proxy` setting when the sockets are created though

arvidn (commented):
I think this could be simplified by using the existing listen socket flag `listen_socket_t::accept_incoming` instead of tacking on additional conditions

include/libtorrent/settings_pack.hpp
Masong19hippows: Thx! fixed in [91081ce](https://github.com/arvidn/libtorrent/pull/7726/commits/91081ce7eb5eff128df3f8fe1f8294f95c1675aa)

include/libtorrent/aux_/proxy_settings.hpp
Masong19hippows: Thx! fixed in [91081ce](https://github.com/arvidn/libtorrent/pull/7726/commits/91081ce7eb5eff128df3f8fe1f8294f95c1675aa)

src/session_impl.cpp
Masong19hippows: @arvidn would it be acceptable to just remove the proxy being a factor in whether the user can enable natpmp? Or is this something that is best left in another PR/Issue/Discussion?

Neustradamus:
Interesting PR :)

LowerHater:
Where did this get to; as this is exactly the issue I'm facing: socks5 proxy for outbound; reverse proxy forwarding my qt port to my client listening at on the qt port.

Masong19hippows:
Honestly, senior year finals hit and I've been swamped with school ever since. As far as I know, this pr is still functional and I have been using it since I put it in. I haven't noticed any weird behavior myself, it just works. I think the pr now just needs to be smoothed over to fit the rest of the code base and that build error needs to be fixed. I'll try to prioritize this soon.

Neustradamus:
.

Neustradamus:
@Masong19hippows: Why have you closed?

Masong19hippows:
Opening a new PR. After a year of updates to the repository, there were a
lot of changes. I tried to rebase but messed up and so it looked ugly.
Thought it would be better to create a new PR with reference to the old
one. Doing so now

PR #7733, opened 2024-09-14
Make setting the file priority early

### LINKED ISSUE #7732: File preallocation does not work after its priority is changed from "do not download"
libtorrent version (or branch): 2.0.10
platform/architecture: Linux

File preallocation does not work after its priority is changed from "do not download".
It is reproducible if you try to download additional files on a partial torrent.

I see that `open_file_impl` intentionally checks for `dont_download` priority.

https://github.com/arvidn/libtorrent/blob/2ab8fedabb7716947edc77958da6d0b5e1040df9/src/mmap_storage.cpp#L899-L901

But the problem is, `set_file_priority` opens the file **before** updating the priority.

https://github.com/arvidn/libtorrent/blob/2ab8fedabb7716947edc77958da6d0b5e1040df9/src/mmap_storage.cpp#L164-L172

---

This patch fixes the issue:

```patch
--- a/src/mmap_storage.cpp
+++ b/src/mmap_storage.cpp
@@ -164,6 +164,7 @@ error_code translate_error(std::error_code const& err, bool const write)
 			if (old_prio == dont_download && new_prio != dont_download)
 			{
 				// move stuff out of the part file
+				m_file_priority[i] = new_prio;
 				std::shared_ptr<aux::file_mapping> f = open_file(sett, i, aux::open_mode::write, ec);
 				if (ec)
 				{

```

But I'm not sure how appropriate it is. What do you think, @arvidn?

HanabishiRecca:
* Fix file preallocation after priority change from `dont_download`.
Fixes #7732.
* Ensure priority consistency to avoid such bugs in the future.

### DIFF
--- a/src/mmap_storage.cpp
+++ b/src/mmap_storage.cpp
@@ -160,8 +160,9 @@ error_code translate_error(std::error_code const& err, bool const write)
 			if (fs.pad_file_at(i)) continue;
 
 			download_priority_t const old_prio = m_file_priority[i];
-			download_priority_t new_prio = prio[i];
-			if (old_prio == dont_download && new_prio != dont_download)
+			m_file_priority[i] = prio[i];
+
+			if (old_prio == dont_download && m_file_priority[i] != dont_download)
 			{
 				// move stuff out of the part file
 				std::shared_ptr<aux::file_mapping> f = open_file(sett, i, aux::open_mode::write, ec);
@@ -216,7 +217,7 @@ error_code translate_error(std::error_code const& err, bool const write)
 					}
 				}
 			}
-			else if (old_prio != dont_download && new_prio == dont_download)
+			else if (old_prio != dont_download && m_file_priority[i] == dont_download)
 			{
 				// move stuff into the part file
 				// this is not implemented yet.
@@ -266,7 +267,6 @@ error_code translate_error(std::error_code const& err, bool const write)
 */
 			}
 			ec.ec.clear();
-			m_file_priority[i] = new_prio;
 
 			if (m_file_priority[i] == dont_download && use_partfile(i))
 			{


src/mmap_storage.cpp
```diff
@@ -164,6 +164,7 @@ error_code translate_error(std::error_code const& err, bool const write)
 			if (old_prio == dont_download && new_prio != dont_download)
 			{
 				// move stuff out of the part file
+				m_file_priority[i] = new_prio;
```
arvidn, line 167: I take it this needs to be set early, so that when files are created, the priority is non-zero, to trigger the file allocation.
That should be a comment here, otherwise someone reading this function in the future may think this is redundant and remove it. (unless there's a test covering this).

arvidn:
nice catch!

src/mmap_storage.cpp
HanabishiRecca: Sure.

HanabishiRecca: But now when I'm thinking about it, maybe it's better to move setting the priority up, instead of duplicating it?

I.e.
```patch
--- a/src/mmap_storage.cpp
+++ b/src/mmap_storage.cpp
@@ -161,6 +161,8 @@ error_code translate_error(std::error_code const& err, bool const write)
 
 			download_priority_t const old_prio = m_file_priority[i];
 			download_priority_t new_prio = prio[i];
+			m_file_priority[i] = new_prio;
+
 			if (old_prio == dont_download && new_prio != dont_download)
 			{
 				// move stuff out of the part file
@@ -266,7 +268,6 @@ error_code translate_error(std::error_code const& err, bool const write)
 */
 			}
 			ec.ec.clear();
-			m_file_priority[i] = new_prio;
 
 			if (m_file_priority[i] == dont_download && use_partfile(i))
 			{

```

This way looks cleaner. Idk why it was at the bottom in the first place, but behavior shouldn't change, as far as I can see.

HanabishiRecca: Ok, I pushed the new iteration, should make it better overall.

src/mmap_storage.cpp
```diff
@@ -160,8 +160,9 @@ error_code translate_error(std::error_code const& err, bool const write)
 			if (fs.pad_file_at(i)) continue;
 
 			download_priority_t const old_prio = m_file_priority[i];
-			download_priority_t new_prio = prio[i];
-			if (old_prio == dont_download && new_prio != dont_download)
+			m_file_priority[i] = prio[i];
```
arvidn, line 163: the issue of updating the priority this early is that if there's a failure, it will be left incorrect. This was (probably) the main reason it was done at the end earlier. Perhaps it doesn't matter so much. If it fails, ideally the user would be able to retry, but if we change the priority to non-zero, this logic won't re-trigger.

arvidn: Also, I think it's really important to document the reason for updating the priority earlier (just like it would have been good to document that updating it even in failure cases is problematic

HanabishiRecca: > the issue of updating the priority this early is that if there's a failure, it will be left incorrect.

I thought about that. But it seems unavoidable anyway, if we want `open_file` to work right. (Beside changing `open_file_impl` not to use `m_file_priority` somehow.)

Arguably we could also move all `prio = m_file_priority;` to the top. As I understand it, a caller could read the mutated state back.

arvidn:
I modified your patch here: https://github.com/arvidn/libtorrent/pull/7738

src/mmap_storage.cpp
HanabishiRecca: > Arguably we could also move all `prio = m_file_priority;` to the top.

Ok, I see you already came up with the same idea. 😅

arvidn: `prio` is owned by the disk job at this point. however, in the response, we pass back the *actual* priorities that ended up being set, and we use the same vector for that

PR #7736, opened 2024-09-29
raise the max piece size to 512 MiB and fix check when loading torrents

arvidn:


include/libtorrent/piece_picker.hpp
```diff
@@ -97,9 +97,9 @@ namespace libtorrent {
 			// priority factor
 			prio_factor = 3,
 			// max blocks per piece
-			// there are counters in downloading_piece that only have 15 bits to
+			// there are counters in downloading_piece that only have 16 bits to
 			// count blocks per piece, that's restricting this
-			max_blocks_per_piece = (1 << 15) - 1
```
arvidn, line 102: the reason this wasn't caught early was because this `15` should have been `14`. The bitfield was probably changed without this changing.

include/libtorrent/block_cache.hpp
```diff
@@ -235,21 +235,14 @@ namespace aux {
...
+		// the number of blocks in the cache for this piece
+		std::uint16_t num_blocks = 0;
 
 		// ---- 64 bit boundary ----
 
 		// the number of dirty blocks in this piece
 		std::uint32_t num_dirty:14;
 
-		// the number of blocks in the cache for this piece
-		std::uint32_t num_blocks:14;
```
arvidn, line 251: by bumping this to 16 bits, the max piece size goes to 512 MiB

glassez: How about `num_dirty`?

include/libtorrent/piece_picker.hpp
```diff
@@ -97,9 +97,9 @@ namespace libtorrent {
 			// priority factor
 			prio_factor = 3,
 			// max blocks per piece
-			// there are counters in downloading_piece that only have 15 bits to
+			// there are counters in downloading_piece that only have 16 bits to
```
glassez, line 100: I didn't see that the counter itself was changed. Just this comment... Or is it a 16-bit size already?

include/libtorrent/block_cache.hpp
arvidn: yeah, that should be wider too

include/libtorrent/piece_picker.hpp
arvidn: `num_blocks` went from 14 bits to a `uint16_t` (i.e. 16 bits)

glassez: > `num_blocks` went from 14 bits to a `uint16_t` (i.e. 16 bits)

Ah, I thought it was about something else.

include/libtorrent/block_cache.hpp
glassez: > How about `num_dirty`?

And `pinned`?

glassez: > > How about `num_dirty`?
> 
> And `pinned`?

I hope you haven't lost sight of it. It is also 15 bits.

arvidn: I'm tempted to just fix the check rather than mucking about with this struct. Do you think there's a legitimate use for torrents with piece sizes this large? It really seems questionable

arvidn:
alternative fix: https://github.com/arvidn/libtorrent/pull/7740

include/libtorrent/block_cache.hpp
glassez: > Do you think there's a legitimate use for torrents with piece sizes this large? It really seems questionable

I only know that the problem was reported by users... In addition, what seemed meaningless yesterday or today will become commonplace tomorrow.
Support for 512 MiB piece size does not require major changes, so why not add it?
Still, I don't insist. The correct processing of limits is good in itself. But I would still expand the potential capacity in upcoming major update.

arvidn: I think it would be much more reasonable to grow torrents in the number of pieces, rather than piece size.

PR #7737, opened 2024-09-29
don't discard netlink messages for previously seen addresses

ssiloti:
Addresses can come and go. Supressing notifications for previously seen addresses can cause the client to miss when the address is re-added. New address notifications also don't necessarily include an IFA_LOCAL attribute so don't filter on that either. Note that notifications don't cause all listen sockets to be closed and re-opened. Despite its name reopen_listen_sockets only closes/opens sockets as necessary to synchronize with the list of addresses enumerated from the OS.

### DIFF
--- a/src/ip_notifier.cpp
+++ b/src/ip_notifier.cpp
@@ -167,15 +167,6 @@ private:
 	netlink::socket m_socket;
 	std::array<char, 4096> m_buf;
 
-	struct local_address
-	{
-		int family;
-		std::array<char, 16> data;
-	};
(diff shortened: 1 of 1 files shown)


ssiloti:
For a while I've noticed that sometimes when my machine reboots the torrent client doesn't listen on IPv6. Restarting the client always fixes it so I suspected it had to do with address change notifications. I finally got around to investigating and found that all of the netlink notifications were being ignored due to these checks.

arvidn (approved):

PR #7738, opened 2024-09-29
fix pre-allocation when changing priority

arvidn:
based on @HanabishiRecca 's research and fix. set m_file_priority early, to make open_file() work as expected. If there's a failure, restore it.

https://github.com/arvidn/libtorrent/pull/7733

### DIFF
--- a/src/mmap_storage.cpp
+++ b/src/mmap_storage.cpp
@@ -63,6 +63,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/aux_/drive_info.hpp"
 #include "libtorrent/stat_cache.hpp"
 #include "libtorrent/hex.hpp" // to_hex
+#include "libtorrent/aux_/scope_end.hpp"
 
 #if TORRENT_HAVE_MMAP || TORRENT_HAVE_MAP_VIEW_OF_FILE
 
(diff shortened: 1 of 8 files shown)


src/mmap_storage.cpp
```diff
@@ -161,15 +162,22 @@ error_code translate_error(std::error_code const& err, bool const write)
 
 			download_priority_t const old_prio = m_file_priority[i];
 			download_priority_t new_prio = prio[i];
+
+			m_file_priority[i] = new_prio;
```
HanabishiRecca, line 166: Again, I'm not a C++ guy, but this looks sketchy to me. Also considering that `new_prio` is not const. Why we need intermediate variable?

arvidn: it's not necessary. but I like giving it an intuitive name. it probably should be `const`, but it's a copy either way.

arvidn:
while working on a test for this, I realized that the `allocate_files` feature doesn't really make any difference on posix systems where files are sparse by default. on windows it makes the files be created without the sparse flag, so they are allocated. Maybe it would make sense to have the same behavior on posix.

HanabishiRecca:
Posix storage is not affected because it does not use this pref in the first place, and `open_file` does not access `m_file_priority` anywhere.
But for the sake of consistency, I guess the same change could be useful there too.

arvidn:
I was wrong. files are actually allocated on non-windows when the sparse flag isn't set.
The test fails without the fix

HanabishiRecca:
I have completely lost the thread of discussion. What do you mean by posix systems?

File preallocation does work on Linux, that's the platform I use and test things on.

arvidn:
thanks @HanabishiRecca !

PR #7739, opened 2024-09-30
limit the piece size in create_torrent constructor

arvidn:


src/create_torrent.cpp
```diff
@@ -48,6 +48,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/aux_/directory.hpp"
 #include "libtorrent/aux_/time.hpp" // for posix_time
 #include "libtorrent/disk_interface.hpp"
+#include "libtorrent/piece_picker.hpp"
```
glassez, line 51: It looks strange that `create_torrent` depends on `piece_picker`. It would be better to define this limit outside to both `create_torrent` and `piece_picker` classes.

arvidn: yeah, that's a good point. This should be a well defined and documented limit

src/create_torrent.cpp
```diff
@@ -455,6 +456,12 @@ namespace {
 			aux::throw_ex<system_error>(errors::invalid_piece_size);
 		}
 
+		// this is an unreasonably large piece size. Some clients don't support
+		// pieces this large.
+		if (piece_size > 256 * 1024 * 1024) {
```
glassez, line 461: At least, I expect libtorrent to have the same limit for creating and using torrents.

src/create_torrent.cpp
```diff
@@ -455,6 +455,12 @@ namespace {
 			aux::throw_ex<system_error>(errors::invalid_piece_size);
 		}
 
+		// this is an unreasonably large piece size. Some clients don't support
+		// pieces this large.
+		if (piece_size > 128 * 1024 * 1024) {
```
glassez, line 460: Am I the only one who doesn't like to look for such fundamental limits among the code, rather than declared as constants in some more appropriate place?

PR #7748, opened 2024-10-07
Support non-default rtables in OpenBSD

lgv5:
OpenBSD supports multiple routing tables. The current code doesn't account for that and instead always request the default gateway for routing table 0. I did a similar treatment in the past to libnatpmp: https://github.com/miniupnp/libnatpmp/pull/46 . In that case, the issue was manifested in Transmission, which I'm user of and hence it was easy to test. I haven't been able to test potential issues for this project, but we might have a tester: https://marc.info/?l=openbsd-ports&m=172833602926653&w=2 . Nevertheless, the proposed patch is the right thing to do in OpenBSD. More details can be found at:

- https://man.openbsd.org/rtable.4 for the multiple routing tables mechanism
- https://man.openbsd.org/sysctl.2#PF_ROUTE~2 for the explanation of the seventh parameter.

Marking it as draft until I get some confirmation of runtime still working in the mailing list.

### DIFF
--- a/src/enum_net.cpp
+++ b/src/enum_net.cpp
@@ -1146,13 +1146,19 @@ int _System __libsocket_sysctl(int* mib, u_int namelen, void *oldp, size_t *oldl
 		}
 		::close(s);
 */
+#define nelems(_a)	(sizeof((_a)) / sizeof((_a)[0]))
+
+#ifdef __OpenBSD__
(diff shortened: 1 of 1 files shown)


lgv5:
This turned out to be an issue in the manpage for `sysctl(2)`, as `NET_RT_DUMP` uses the current rtable instead of 0. Closing. Sorry for the noise.

PR #7750, opened 2024-10-13
Fix some more typos

xavier2k6:


bindings/c/library.cpp
```diff
@@ -99,7 +99,7 @@ TORRENT_EXPORT void* session_create(int tag, ...)
 	va_list lp;
 	va_start(lp, tag);
 
-	fingerprint fing("LT", lt::version_major, lt::version_minor, lt::version_tiny, 0);
+	fingerprint find("LT", lt::version_major, lt::version_minor, lt::version_tiny, 0);
```
arvidn, line 102: I think `fing` is supposed to refer to "fingerprint"

fuzzers/src/add_torrent.cpp
```diff
@@ -58,9 +58,9 @@ std::vector<sha256_hash> g_tree;
 int const piece_size = 1024 * 1024;
 int const blocks_per_piece = piece_size / lt::default_block_size;
 int const num_pieces = 10;
-int const num_leafs = merkle_num_leafs(num_pieces * blocks_per_piece);
-int const num_nodes = merkle_num_nodes(num_leafs);
-int const first_leaf = merkle_first_leaf(num_leafs);
+int const num_leaves = merkle_num_leaves(num_pieces * blocks_per_piece);
```
arvidn, line 61: I'm not sure about this change. "leaf" is a term of art in computer science. Perhaps mostly used as adjective to describe the leaf nodes. I hesitate to call them "leaves" as that implies actual leaves on actual trees (not the computer science data structure "tree").

They are currently (as far as I know) consistently referred to as "leafs". I kind of see the "node" as implied in that name. Perhaps this is a controversial perspective. I would rather rename these names "leaf_nodes" over "leaves". But I also think the current acronym is quite reasonable.

I think the change is controversial mostly because it's already internally consistent, and a rename would have to make sure it's consistently renamed everywhere.

xavier2k6: `leaves` are already mentioned throughout some parts of code in repo:
https://github.com/search?q=repo%3Aarvidn%2Flibtorrent%20leaves&type=code

arvidn: I see. so some decision has to be made then. do you think "leaves" is preferred over "lefs"?

xavier2k6: > I see. so some decision has to be made then. do you think "leaves" is preferred over "lefs"?

I think changing to "leaves" would be ok, but it's up to you really.
____

https://en.wikipedia.org/wiki/Merkle_tree#Overview
>A hash tree is a [tree](https://en.wikipedia.org/wiki/Binary_tree) of [hashes](https://en.wikipedia.org/wiki/Hash_function) in which the leaves (i.e., leaf nodes, sometimes also called "leafs") are hashes of data [blocks](https://en.wikipedia.org/wiki/Blockchain) in, for instance, a file or set of files.

xavier2k6: awaiting feedback

arvidn: I think it's fine to rename "leafs" to "leaves". But it might be safer to do so in `master`, in case any of these names are visible to clients

xavier2k6: > I think it's fine to rename "leafs" to "leaves". But it might be safer to do so in `master`, in case any of these names are visible to clients

Will open a new PR targeting `master` in time..

PR #7754, opened 2024-10-17
allow connections while checking resume data if no_verify_files flag is set

e1y4s:
When adding a torrent with peers, if metadata is already available and the torrent is in `checking_resume_data` state, `torrent::want_peers()` currently returns `false`. This behavior prevents `torrent::do_connect_boost()` from initiating connections, leading to delays.  

This modification addresses the issue by permitting peer connections in this scenario when `no_verify_files` flag is set.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 
 2.0.11 not released
 
+	* allow boost connect while checking resume data if no_verify_files flag is set
 	* fix BEP-40 peer priority for IPv6
 	* limit piece size in torrent creator
 	* fix file pre-allocation when changing file priority (HanabishiRecca)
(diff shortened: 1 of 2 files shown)


arvidn (approved):
thanks! looks good. It's kind of subtle, it might warrant a comment in the code explaining this carve-out. Would you mind adding an entry to the `Changelog` as well?

joriscarrier:
@arvidn One more question: why don't you want to connect to peers while checking files (m_state == torrent_status::checking_files)?

e1y4s:
> thanks! looks good. It's kind of subtle, it might warrant a comment in the code explaining this carve-out. Would you mind adding an entry to the `Changelog` as well?

Thanks! Just added both. Let me know if anything else needs adjustment.

arvidn:
@joriscarrier 
> why don't you want to connect to peers while checking files (m_state == torrent_status::checking_files)?

Although it would be possible to check files while participating in the swarm it's more complicated and isn't supported by the torrent/piece-picker (and possibly not the disk I/O system). The unchecked pieces would need a new state to indicate that they can't be picked for downloading, even though their priority may be non-zero.

For historical reasons, checking has been a separate state where we, essentially, don't have a piece picker.

arvidn:
@e1y4s thanks!

PR #7758, opened 2024-10-20
don't hint `POSIX_FADV_RANDOM`

arvidn:
Normal piece sizes, of 2 or 4 MiB, probably warrants read-ahead. Some antecdotes suggests this improve read speed in some circumstances.

The use of the internal open mode flag `random_access` is really used to hint sequential access, for when checking files. This patch, therefore, renames the flag to `sequential_access` and makes it mean that instead.

This is an internal flag so doesn't affect the user facing API.

### DIFF
--- a/src/mmap_disk_io.cpp
+++ b/src/mmap_disk_io.cpp
@@ -142,7 +142,7 @@ namespace {
 	aux::open_mode_t file_mode_for_job(aux::mmap_disk_job* j)
 	{
 		aux::open_mode_t ret = aux::open_mode::read_only;
-		if (!(j->flags & disk_interface::sequential_access)) ret |= aux::open_mode::random_access;
+		if (j->flags & disk_interface::sequential_access) ret |= aux::open_mode::sequential_access;
 		return ret;
 	}
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 
 2.0.11 not released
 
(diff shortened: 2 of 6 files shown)


HanabishiRecca:
Just for the context, from [`posix_fadvise(2)`](https://www.man7.org/linux/man-pages/man2/posix_fadvise.2.html):

> Under Linux, `POSIX_FADV_NORMAL` sets the readahead window to the default size for the backing device; `POSIX_FADV_SEQUENTIAL` doubles this size, and `POSIX_FADV_RANDOM` disables file readahead entirely.

src/mmap_disk_io.cpp
```diff
@@ -142,7 +142,7 @@ namespace {
 	aux::open_mode_t file_mode_for_job(aux::mmap_disk_job* j)
 	{
 		aux::open_mode_t ret = aux::open_mode::read_only;
-		if (!(j->flags & disk_interface::sequential_access)) ret |= aux::open_mode::random_access;
+		if (!(j->flags & disk_interface::sequential_access)) ret |= aux::open_mode::sequential_access;
```
HanabishiRecca, line 145: You forgot to remove the `!` here. It makes logic to be reversed.
And confirmed by testing, I see regular reads using 2x readahead size, but torrent re-checking use 1x instead.

arvidn:
@HanabishiRecca thanks for your testing and review!

PR #7765, opened 2024-10-25
fix bug where file_progress could sometimes be reported as >100%

arvidn:


xavier2k6:
Does RC_1_2 require this fix to be backported?

PR #7768, opened 2024-10-25
Move construct `std::shared_ptr` objects

Chocobo1:


arvidn (approved):
looks good. I would have expected you would also need `emplace_back()` rather than `push_back()`, but maybe this is enough

Chocobo1:
I looked into two C++ libraries. Both of them (emplace_pack, push_back) seem to forward the `value` to the same internal function so I suppose there wasn't too much difference.

https://github.com/microsoft/STL/blob/ca1553d38480b1224cb5b36ebd755af17aec275b/stl/inc/list#L978-L980
https://github.com/microsoft/STL/blob/ca1553d38480b1224cb5b36ebd755af17aec275b/stl/inc/list#L997-L1006

https://github.com/gcc-mirror/gcc/blob/40fedaf35fa99a9728d5b84d47035f4c92e1ba90/libstdc%2B%2B-v3/include/bits/stl_list.h#L1309-L1311
https://github.com/gcc-mirror/gcc/blob/40fedaf35fa99a9728d5b84d47035f4c92e1ba90/libstdc%2B%2B-v3/include/bits/stl_list.h#L1319-L1325

PR #7771, opened 2024-10-26
Add announce_port support

0xThiebaut:
The `announce_port` setting permits to overwrite the port passed along to trackers as the `&port=` parameter. If left as the default, the listening port is used. This setting is only meant for very special cases where a seed's listening port differs from the effectively exposed port (e.g., through external NAT-PMP).

This PR would be complementary to the https://github.com/arvidn/libtorrent/pull/7726 `listen_on_proxy` setting.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,6 +13,7 @@
 	* fix integer overflow in piece picker
 	* torrent_status::num_pieces counts pieces passed hash check, as documented
 	* check settings_pack::max_out_request_queue before performance alert
+	* add announce_port setting to override the port announced to trackers
 
 2.0.10 released
 
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -2068,6 +2068,20 @@ namespace aux {
 			i2p_inbound_length,
 			i2p_outbound_length,
 
+			// ``announce_port`` is the port passed along as the ``port`` parameter
+			// to remote trackers such as HTTP or DHT. This setting does not affect
+			// the effective listening port nor local service discovery announcements.
+			// If left as zero (default), the listening port value is used.
+			//
+			// .. note::
+			//    This setting is only meant for very special cases where a
+			//    seed's listening port differs from the external port. As an
+			//    example, if a local proxy is used and that the proxy supports
+			//    reverse tunnels through NAT-PMP, the tracker must connect to
+			//    the external NAT-PMP port (configured using ``announce_port``)
+			//    instead of the actual local listening port.
+			announce_port,
+
 			max_int_setting_internal
 		};
 
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -1301,9 +1301,11 @@ namespace {
 #endif
 			req.ssl_ctx = &m_ssl_ctx;
 #endif
-
-		auto ls = req.outgoing_socket.get();
-		if (ls)
+		if (const auto announce_port = std::uint16_t(m_settings.get_int(settings_pack::announce_port)))
+		{
+			req.listen_port = announce_port;
+		}
+		else if (auto ls = req.outgoing_socket.get())
 		{
 			req.listen_port =
 #ifdef TORRENT_SSL_PEERS
--- a/include/libtorrent/session_handle.hpp
+++ b/include/libtorrent/session_handle.hpp
@@ -507,6 +507,7 @@ namespace libtorrent {
 		// specified info-hash, advertising the specified port. If the port is
 		// left at its default, 0, the port will be implied by the DHT message's
 		// source port (which may improve connectivity through a NAT).
+		// ``dht_announce()`` is not affected by the ``announce_port`` override setting.
 		//
 		// Both these functions are exposed for advanced custom use of the DHT.
 		// All torrents eligible to be announce to the DHT will be automatically,
--- a/simulation/test_tracker.cpp
+++ b/simulation/test_tracker.cpp
@@ -361,6 +361,90 @@ void on_alert_notify(lt::session* ses)
 	});
 }
 
+void test_announce()
+{
+	using sim::asio::ip::address_v4;
+	sim::default_config network_cfg;
+	sim::simulation sim{network_cfg};
+
+	sim::asio::io_context web_server(sim, make_address_v4("2.2.2.2"));
+
+	// listen on port 8080
+	sim::http_server http(web_server, 8080);
+
+	int announces = 0;
+
+	// expect announced IP & port
+	std::string const expect_port = "&port=1234";
+	std::string const expect_ip = "&ip=1.2.3.4";
+
+	http.register_handler("/announce"
+	, [&announces, expect_port, expect_ip](std::string method, std::string req
+		, std::map<std::string, std::string>&)
+	{
+		++announces;
+		TEST_EQUAL(method, "GET");
+		TEST_CHECK(req.find(expect_port) != std::string::npos);
+		TEST_CHECK(req.find(expect_ip) != std::string::npos);
+		char response[500];
+		int const size = std::snprintf(response, sizeof(response), "d8:intervali1800e5:peers0:e");
+		return sim::send_response(200, "OK", size) + response;
+	});
+
+	{
+		lt::session_proxy zombie;
+
+		std::vector<asio::ip::address> ips;
+		ips.push_back(make_address("123.0.0.3"));
+
+		asio::io_context ios(sim, ips);
+		lt::settings_pack sett = settings();
+		sett.set_str(settings_pack::listen_interfaces, "0.0.0.0:6881");
+		sett.set_str(settings_pack::announce_ip, "1.2.3.4");
+		sett.set_int(settings_pack::announce_port, 1234);
+
+		auto ses = std::make_unique<lt::session>(sett, ios);
+
+		ses->set_alert_notify(std::bind(&on_alert_notify, ses.get()));
+
+		lt::add_torrent_params p;
+		p.name = "test-torrent";
+		p.save_path = ".";
+		p.info_hashes.v1.assign("abababababababababab");
+
+		p.trackers.push_back("http://2.2.2.2:8080/announce");
+		ses->async_add_torrent(p);
+
+		// stop the torrent 5 seconds in
+		sim::timer t1(sim, lt::seconds(5)
+			, [&ses](boost::system::error_code const&)
+		{
+			std::vector<lt::torrent_handle> torrents = ses->get_torrents();
+			for (auto const& t : torrents)
+			{
+				t.pause();
+			}
+		});
+
+		// then shut down 10 seconds in
+		sim::timer t2(sim, lt::seconds(10)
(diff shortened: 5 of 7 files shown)


0xThiebaut:
#### Example use-case
To secure outbound connections, a proxy server on the same subnet can be used to tunnel requests through a VPN (e.g., using [pufferffish/wireproxy](https://github.com/pufferffish/wireproxy)). Some VPNs support NAT-PMP (e.g., [ProtonVPN](https://protonvpn.com/support/port-forwarding-manual-setup/#linux)) which could technically allow the proxy server to perform dynamic port forwarding back.

Through the `announce_port` setting, a tracker can be informed on which NAT-PMP port has been made available through the proxy while the proxy has been configured to direct the port to the actual listening port.
Announce Port

src/http_tracker_connection.cpp
```diff
@@ -143,7 +144,7 @@ namespace libtorrent {
 				, escape_string({tracker_req().pid.data(), 20}).c_str()
 				// the i2p tracker seems to verify that the port is not 0,
 				// even though it ignores it otherwise
-				, tracker_req().listen_port
+				, announce_port ? announce_port : tracker_req().listen_port
```
arvidn, line 147: I think a better place to override the listen port is in `session_impl.cpp` where the tracker request object is set up and queued.

https://github.com/arvidn/libtorrent/blob/RC_2_0/src/session_impl.cpp#L1306-L1319

Then you don't need to change the HTTP and UDP tracker logic (I don't think at least)

src/tracker_manager.cpp
```diff
@@ -278,8 +278,15 @@ constexpr tracker_request_flags_t tracker_request::i2p;
...
-			, req.listen_port);
+		if (cb) {
+			if (const auto announce_port = std::uint16_t(sett.get_int(settings_pack::announce_port))) {
+				cb->debug_log("*** QUEUE_TRACKER_REQUEST [ listen_port: %d, announce_port: %d ]"
+				, req.listen_port, announce_port);
+			} else {
+				cb->debug_log("*** QUEUE_TRACKER_REQUEST [ listen_port: %d ]"
+				, req.listen_port);
+			}
+		}
```
arvidn, line 289: I'm not sure this change is necessary.

include/libtorrent/settings_pack.hpp
```diff
@@ -2068,6 +2068,16 @@ namespace aux {
 			i2p_inbound_length,
 			i2p_outbound_length,
 
+			// ``announce_port`` is the port passed along to trackers as the
+			// ``&port=`` parameter. If left as the default, the listening port
+			// is used.
```
arvidn, line 2073: I think you need to mention that 0 means the listen port will not be overridden. I also think it would be nice to give an example of a setup where this would be useful.

arvidn (commented):
I think it would be good to have some test coverage of this feature. Maybe you could add a test case here:

https://github.com/arvidn/libtorrent/blob/RC_2_0/simulation/test_tracker.cpp#L384-L410

src/tracker_manager.cpp
0xThiebaut: As in the entire change is unnecessary or you would rather have the a single `debug_log` call where `announce_port` would be `0` if not set?

arvidn: I don't think it's important (nor necessary) to preserve the distinction between "listen_port" and "announce_port". We only use this for tracker announcements anyway.

0xThiebaut: Thanks! The change will be dropped anyway as `req.listen_port` will be overwritten in https://github.com/arvidn/libtorrent/blob/RC_2_0/src/session_impl.cpp#L1306-L1319

Waiting on the tests to pass to submit the change.

0xThiebaut:
> I think it would be good to have some test coverage of this feature. Maybe you could add a test case here:
> 
> https://github.com/arvidn/libtorrent/blob/RC_2_0/simulation/test_tracker.cpp#L384-L410

A simple test coverage has been added to ensure `announce_ip` and `announce_port` are properly reflected in the tracker URL.

arvidn (commented):
looks good. would you mind adding a line to `Changelog` too?

0xThiebaut:
> looks good. would you mind adding a line to `Changelog` too?

Done as well, thanks!

arvidn:
should this feature affect the port we announce to the DHT?
the current default is to use "implied port", i.e. the source port of the DHT message. This works since we accept uTP connections on the same port.

arvidn:
and I imagine it should *not* affect local service discovery announcements (i.e. multicast on the local network)

0xThiebaut:
The `announce_port` feature is intended for cases where additional appliances are located between local and remote peers. The local network discovery is by definition local so I agree it should not be affected by the `announce_port` setting. A future boolean setting could be introduced should anyone manage to find a use-case requiring `announce_port` overrides for local network discovery.

For DHT/uTP there can however be appliances between the local and remote peers; It would make sense to support overriding the announced port. I'm not familiar with DHT/uTP, do they currently use the same listening port as the TCP listener but for UDP?

arvidn:
> I'm not familiar with DHT/uTP, do they currently use the same listening port as the TCP listener but for UDP?

Yes, libtorrent attempts to open the listen socket for TCP and UDP on the same port. The UDP socket is used for DHT, uTP (peer connections) and UDP trackers.

0xThiebaut:
My current understanding would be as follow:

For UDP trackers I think we are covered as it uses `req.listen_port` which `announce_port` overrides.
https://github.com/arvidn/libtorrent/blob/6a2c1eef6f3ba6b0a9763f3968bb79ddef73a355/src/udp_tracker_connection.cpp#L732

For uTP (peer connections) I don't believe there is a notion of "announcing"; it is merely the transport protocol and does not define which ports are used.

For DHT we need to ensure that `announce_port` indeed overrides the announced port. I believe we would need to do so in the session implementation and, if `announce_port` is set, make sure to clear the `dht::announce::implied_port` flag while overriding the `port` argument.
https://github.com/arvidn/libtorrent/blob/6a2c1eef6f3ba6b0a9763f3968bb79ddef73a355/src/session_impl.cpp#L6217-L6221

Would that make sense?

0xThiebaut:
Another DHT option is to do it at node-level where the logic already replaces `0` with the listening port; Although doing it at session-level would match the TCP `announce_port` variant.
https://github.com/arvidn/libtorrent/blob/6a2c1eef6f3ba6b0a9763f3968bb79ddef73a355/src/kademlia/node.cpp#L458-L467

arvidn:
I can land this PR, and if you think the DHT behavior also should change, you could make another PR. I think having this logic in the session would make sense. It might also make sense to update the documentation for the setting to make it clear that it doesn't affect local service discovery and whether or not it affects DHT announces

0xThiebaut:
I'll update the PR today to reflect that `announce_port` does not affect local discovery. I'll also apply the DHT change as I would rather avoid complex documentation with non-intuitive edge-cases.

0xThiebaut:
I might be misunderstanding the relationship between a `session` and a `torrent` but wouldn't the following bypass the `announce_port` applied in the session's `session_impl::dht_announce`? Based on my limited understanding I would have expected the torrent to call `m_ses.dht_announce()` instead of `m_ses.dht()->announce()`.
https://github.com/arvidn/libtorrent/blob/e5fe95f92273bd775961e2f1f3b4ec45f8012d9e/src/torrent.cpp#L2816-L2820

arvidn:
there's only one DHT node (technically, one per network interface you have). And to smooth-out the network traffic, torrents announces are staggered. The session loops over all the torrents and announces them one at a time, at a pace making them all announce once per 15 minutes (or whatever the setting has been set to).

0xThiebaut:
So on top of applying `announce_port` in `session_impl::dht_announce` we also need to add it to `torrent::dht_announce` as both functions determine a port to use? Or we move the logic to the underlying `dht_tracker::announce` which is used by both the `session_impl` and `torrent`?

Apologies if I am missing something.

src/session_impl.cpp
```diff
@@ -6214,9 +6216,14 @@ namespace {
...
 
-	void session_impl::dht_announce(sha1_hash const& info_hash, int port, dht::announce_flags_t const flags)
+	void session_impl::dht_announce(sha1_hash const& info_hash, int port, dht::announce_flags_t flags)
 	{
 		if (!m_dht) return;
+		if (const auto announce_port = std::uint16_t(m_settings.get_int(settings_pack::announce_port)))
+		{
+			port = announce_port;
+			flags &= ~dht::announce::implied_port;
+		}
```
arvidn, line 6226: is this change really necessary? you already have logic to not set `implied_port` and to pass in the overridden announce port, in `torrent.cpp` line 2008

ChangeLog
```diff
@@ -13,6 +13,7 @@
 	* fix integer overflow in piece picker
 	* torrent_status::num_pieces counts pieces passed hash check, as documented
 	* check settings_pack::max_out_request_queue before performance alert
+	* add announce_port setting to overwrite the port announced to trackers
```
arvidn, line 16:
```suggestion
	* add announce_port setting to override the port announced to trackers
```

include/libtorrent/settings_pack.hpp
```diff
@@ -2068,6 +2068,20 @@ namespace aux {
 			i2p_inbound_length,
 			i2p_outbound_length,
 
+			// ``announce_port`` is the port passed along to trackers as the
+			// ``port`` parameter. This setting does not affect the effective
```
arvidn, line 2072: I think it would be good to mention that it affects DHT announcements as well

arvidn:
thanks for your patience. I just did one more review pass

src/session_impl.cpp
0xThiebaut: I believe so? Based on my understanding `dht_tracker::announce` is called directly from both `torrent::dht_announce` as well as `session_impl::dht_announce`. Hence, I added the `announce_port` in both of these places.

I don't believe there is a relationship between `torrent::dht_announce` and `session_impl::dht_announce` where one would affect the other?

arvidn: I see

arvidn: as far as I can tell, this function is only called from `session_handle.cpp`, which means the *user* is calling it. It might be a bit unintuitive to override the port and flag passed in directly from the client.

0xThiebaut: Alright so ideally we would have the `announce_port` setting only affect announcements which are not _client_-issued?

If so, would that mean I roll back this change and only keep `torrent::dht_announce` for DHT and `session_impl::queue_tracker_request` for HTTP/UDP?

In which case it might also be worth documenting the non-application of the `announce_port` on the `session_handle`.
https://github.com/arvidn/libtorrent/blob/7f691246aa58d756fbcdc2f3a03b1623055148ac/include/libtorrent/session_handle.hpp#L502-L517

Thanks for your patience!

arvidn: yeah, I think that sounds good

arvidn:
thanks!

PR #7775, opened 2024-10-28
requires openssl minimum version 1.1.0

aldenml:
Hi @arvidn, this fixes a compilation for Android with latest openssl.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,6 @@
 2.1.0 not released
 
+	* requires OpenSSL minimum version 1.1.0 with SNI support
 	* try harder to bind TCP and UDP sockets to the same port
 	* made disk_interface's status_t type a flags type
 	* optimize resume data format to use less space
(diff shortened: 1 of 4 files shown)


arvidn:
was `OPENSSL_VERSION_NUMBER` not set correctly?
which part of this is the fix?

aldenml:
> was `OPENSSL_VERSION_NUMBER` not set correctly? which part of this is the fix?

The macro `OPENSSL_API_COMPAT` is not set, and was failing to compile `ssl.cpp`. I could "fix" only that, but I think removing support for < 1.0 is a good idea. What do you think?

aldenml:
hmmm... I see Android failing with, but I will wait for your reply before addressing it.
```
src/ssl.cpp:162:3: error: unused member function '~lifecycle' [-Werror,-Wunused-member-function]
                ~lifecycle() = default;
```

aldenml:
Yes, `OpenSSL_add_all_algorithms` undefined,  and also `CRYPTO_cleanup_all_ex_data`. The problem is that `OPENSSL_API_COMPAT` is not defined when compiling with `OpenSSL 3.4.0` using `b2` all from source code. I don't know in which version that started to happen.

I think that removing `OpenSSL 1.0.x` that's is not longer supported is a good idea. I'm not aware of any serious problem tbh, but more than 8 years old, and not distributed anymore....also, less `#if`s

Issue #7778, opened 2024-10-28
File pool does not work as intended


HanabishiRecca:
I noticed strange behavior of file descriptors. Opened files stale indefinitely, regardless of the file pool size.

Some investigation with `TRACE_FILE_VIEW_POOL` enabled shows that libtorrent never rotates any descriptors in the pool, except constantly reusing the last one when the pool is filled up.

```
135805188216512 opening file: (18, 12)
135805188216512 opening file: (18, 6)
135805188216512 opening file: (0, 0)
135805188216512 opening file: (18, 10)
135805188216512 opening file: (18, 8)
135805188216512 opening file: (18, 4)
135805188216512 opening file: (18, 72)
135805188216512 opening file: (18, 73)
135805188216512 opening file: (18, 74)
135805188216512 removing: (18, 74)
135805188216512 opening file: (18, 47)
135805188216512 removing: (18, 47)
135805188216512 opening file: (18, 82)
135805188216512 removing: (18, 82)
135805188216512 opening file: (18, 83)
135805188216512 removing: (18, 83)
135805188216512 opening file: (18, 21)
135805188216512 removing: (18, 21)
135805188216512 opening file: (18, 48)
135805188216512 removing: (18, 48)
135805188216512 opening file: (18, 61)
135805188216512 removing: (18, 61)
135805188216512 opening file: (18, 21)
135805188216512 removing: (18, 21)
135805188216512 opening file: (27, 0)
```

Obviously, this is not how it intended to work, and actually could lead to severe performance degradation.

I assume this thing does not sort files as expected.

https://github.com/arvidn/libtorrent/blob/6a2c1eef6f3ba6b0a9763f3968bb79ddef73a355/include/libtorrent/aux_/file_view_pool.hpp#L165-L166

Replacing `back`s with `front`s inside the `file_view_pool::remove_oldest` function seems to mitigate the issue.

```diff
--- a/src/file_view_pool.cpp
+++ b/src/file_view_pool.cpp
@@ -351,11 +351,11 @@ namespace libtorrent { namespace aux {
 
 #if TRACE_FILE_VIEW_POOL
 		std::cout << std::this_thread::get_id() << " removing: ("
-			<< lru_view.back().key.first << ", " << lru_view.back().key.second << ")\n";
+			<< lru_view.front().key.first << ", " << lru_view.front().key.second << ")\n";
 #endif
 
-		auto mapping = std::move(lru_view.back().mapping);
-		lru_view.pop_back();
[...]

PR #7779, opened 2024-10-29
fix performance bug in the file pool

arvidn:
It was evicting the most recently used files instead of the least recently used. The use-order sequence index was not maintained correctly.
Discovered by HanabishiRecca https://github.com/arvidn/libtorrent/issues/7778

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 
 2.0.11 not released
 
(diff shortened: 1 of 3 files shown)


HanabishiRecca:
I have it running for a while an it appears to be working as intended. Thanks.

PR #7780, opened 2024-10-29
Test merge

arvidn:


arvidn:
it landed: https://github.com/arvidn/libtorrent/commits/master
I was also surprised github didn't pick this up. maybe I failed to push the version with the updated submodule.

PR #7805, opened 2024-12-02
cibuildwheel: update workflow

xavier2k6:
* Bumped cibuildwheel action
* Removed macos-12 runner (deprecated)
* Updated Boost Download URL
* Used more appropriate OS Runners for required jobs (architecure specific)
* Explicitly declare each cibuildwheel identifier

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -6,7 +6,7 @@ name: cibuildwheel
 # sample.
 
 # The full list of cibuildwheel's build targets can be found here:
-# https://github.com/pypa/cibuildwheel/blob/v2.21.3/cibuildwheel/resources/build-platforms.toml
+# https://github.com/pypa/cibuildwheel/blob/v2.22.0/cibuildwheel/resources/build-platforms.toml
 
 # Notes on build targets we (don't) support:
 # - pypy: libtorrent doesn't build with pypy as of writing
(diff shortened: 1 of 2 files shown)


xavier2k6:
@qstokkink I've done a dry-run for all python versions from python 3.9+ across all builds so that the tribler team can test/use them especially for requested macOS 12 prior to it's imminent deprecation/removal as a GH runner image. (Today - 3rd Dec is it's deadline....)

I plan on removing macos-12 when it's completely unavailable as a GH runner image from this PR.

xavier2k6:
I plan to do a re-base/one final change after your macos fixes are in-place.....unsure of why windows-2025 is failing & don't really have time to look into it. (will more than likely revert to windows-latest which is currently = windows-2022)

xavier2k6:
@arvidn should we sync below files with upstream??
https://github.com/arvidn/libtorrent/blob/RC_2_0/tools/cibuildwheel/manylinux/build-openssl.sh
https://github.com/arvidn/libtorrent/blob/RC_2_0/tools/cibuildwheel/manylinux/build_utils.sh

Upstream:
https://github.com/pypa/manylinux/blob/main/docker/build_scripts/build-openssl.sh
https://github.com/pypa/manylinux/blob/main/docker/build_scripts/build_utils.sh
____

Unsure, if below needs to be updated at some stage:
https://github.com/arvidn/libtorrent/blob/RC_2_0/tools/cibuildwheel/manylinux/openssl-version.sh

.github/workflows/cibuildwheel.yml
```diff
@@ -126,7 +160,7 @@ jobs:
         vcpkg install openssl:x86-windows
```
arvidn, line 160: I think the condition for this job needs to be updated to also cover `CIBW_BUILD` ending with "win_amd64"

arvidn: or maybe instead look at whether `os` starts with "windows-"

xavier2k6: The `win_amd64` build had no issues previously with `windows-2022`...I'll think I'll just revert to that as the `windows-2025` image is still classed as `beta` & there are still tools etc. been added/updated. (will look in to using it again in a few weeks when v3 of cibuildweel action is released.)

Is that ok with you??

arvidn: sure

xavier2k6: I'll rebase/make changes in about an hour or so.

xavier2k6: Thoughts on sync? -> https://github.com/arvidn/libtorrent/pull/7805#issuecomment-2613052694

arvidn:
> @arvidn should we sync below files with upstream?? https://github.com/arvidn/libtorrent/blob/RC_2_0/tools/cibuildwheel/manylinux/build-openssl.sh https://github.com/arvidn/libtorrent/blob/RC_2_0/tools/cibuildwheel/manylinux/build_utils.sh
> 
> Upstream: https://github.com/pypa/manylinux/blob/main/docker/build_scripts/build-openssl.sh https://github.com/pypa/manylinux/blob/main/docker/build_scripts/build_utils.sh

Yeah, that would seem reasonable. There's a `tools/cibuildwheel/manylinux/update-scripts.sh` that might do that

.github/workflows/cibuildwheel.yml
arvidn: https://github.com/arvidn/libtorrent/pull/7824

xavier2k6:
@arvidn re-based & as you can see both manylinux builds are passing.....should I add the manylinux script changes here too or do you want to do your own thing?

arvidn:
CI seems very finicky when changing these things. so any green CI I think should land without mixing in additional risks of all the errors I've seen in the other PR

arvidn (approved):

arvidn:
thanks!

arvidn:
so close: https://github.com/arvidn/libtorrent/actions/runs/13010473229/job/36289088640

it looks like the upload job is expecting to be parameterized by a build matrix. since it isn't, those variables expand to empty strings.

xavier2k6:
@arvidn I'm unsure if changing the 2x `path: wheelhouse` -> `path: wheelhouse/*.whl` under `upload_pypi / upload_pypi_test` will help.

BTW, It may be of benefit to rebase #7824

qstokkink:
Looks reasonably straightforward to solve. Let me know if you need any help.

arvidn:
I expect the best and simplest solution is to just pull all `*.whl` artifacts and submit those, right?

xavier2k6:
Ya, if there was a way to pull all artifacts etc without initiating the publish step in case there's any mistake......

@qstokkink any help at all will be appreciated....

arvidn:
https://github.com/arvidn/libtorrent/pull/7836

xavier2k6:
@arvidn Below might be of interest you for the future.....

https://learn.scientific-python.org/development/guides/repo-review/?repo=arvidn%2Flibtorrent&branch=RC_2_0

PR #7808, opened 2024-12-09
Fix mmap error when reading empty file during resume check on Linux

### LINKED ISSUE #7790: force recheck with zero byte files gives mmap error

```sh
$ rpm -q rb_libtorrent qbittorrent
rb_libtorrent-2.0.10-3.fc41.x86_64
qbittorrent-5.0.1-1.fc41.x86_64
```

how to reproduce it.
11/12/24 12:03 AM - File error alert. Torrent: "...". Reason: "... file_mmap (...) error: Invalid argument"
--- 

1. start a torrent that creates a several files **in a subdirectory**
2. start the torrent
3. stop the torrent
4. remove a file in  a subdir and touch it, so it exists with zero bytes
5. force recheck

I think I triggered this by using an older version of qbittorrent that would touch files at zero bytes, then switched to qbittorrent that used new mmap version of libtorrent which caused this error, but it's easy to reproduce using the procedure with just latest libtorrent.  I tried on a simple torrent that didn't use subdirectories, and the force recheck didn't have any issues, so I used a torrent that creates lots of subdirs and it triggered immediately when following my procedure.

This was reported as well in qbittorrent here, but I couldn't find if it was reported here:
https://github.com/qbittorrent/qBittorrent/issues/20398

Doekin:
This PR fixes #7790, where calling `mmap()` with length = 0 on Linux causes an `-EINVAL` error during the resume check process.

### DIFF
--- a/src/mmap.cpp
+++ b/src/mmap.cpp
@@ -172,7 +172,7 @@ file_mapping_handle::file_mapping_handle(file_handle file, open_mode_t const mod
 file_mapping::file_mapping(file_handle file, open_mode_t const mode, std::int64_t const file_size)
 	: m_size(memory_map_size(mode, file_size, file))
 	, m_file(std::move(file))
-	, m_mapping((mode & open_mode::no_mmap) ? nullptr
+	, m_mapping((mode & open_mode::no_mmap) || m_size == 0 ? nullptr
 		: mmap(nullptr, static_cast<std::size_t>(m_size)
 			, mmap_prot(mode), mmap_flags(mode), m_file.fd(), 0))
 {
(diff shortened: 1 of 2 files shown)


arvidn:
There are tests with torrents that have 0-size files. Is this issue specifically when files that (when completed) are not 0-sized, but happens to have 0-size on disk during checking?

We would need an automated test to trigger this as well

Doekin:
Yes, you're correct.
I'm not entirely sure under what circumstances these 0-size files are generated

arvidn:
sorry for the delay. thanks!

Doekin:
No problem, thanks for reviewing and merging!

arvidn:
Windows seemed to have a similar issue (at least in CI), addressed by this follow-up commit: https://github.com/arvidn/libtorrent/pull/7816/commits/e7ac22e1f1f66f98ddf2eafbbbe0f15624681086

PR #7815, opened 2025-01-20
bump version

arvidn:


xavier2k6:
Are you going to replace `aligned_union` (deprecated) before release?

xavier2k6:
Now may also be a good time to change `ubuntu-20.04` -> `ubuntu-22.04 or ubuntu-24.04/ubuntu-latest` where appropriate as `ubuntu-20.04` images are being deprecated/removed, see: https://github.com/actions/runner-images/issues/11101

arvidn:
> Are you going to replace `aligned_union` (deprecated) before release?

as far as I can tell it's only being used when building with older versions of C++. But I suppose it's unnecessary to keep this logic around rather than just using the local implementation unconditionally.

PR #7819, opened 2025-01-23
Test merge

arvidn:


xavier2k6:
`c.yml` will need the ubuntu image to be bumped from `ubuntu-20.04`

PR #7822, opened 2025-01-24
pkg-config file generation: resolve CMAKE_INSTALL_PREFIX at install time

Piraty:
currently it's not straightforward to install into a destdir without providing the `CMAKE_INSTALL_PREFIX` definition at configure stage 

```
cmake -B builddir ...
cmake --build builddir ... 
cmake --install builddir --prefix=destdir
```
this currently tries to install pc file into `/usr/local/lib/pkgconfig` and neither works nor makes sense

this PR makes it work

i didn't try to understand the rationale behind this extra level of "in"direction for generating the pkg-file, so i might have missed the spot where to clean up the string replacement

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,7 +34,6 @@ libtool
 *.exp
 
 *.m4
-*.in
 *.pc
 *.libs
 *.deps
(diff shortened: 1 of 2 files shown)


arvidn:
However, the change to `.gitignore` seem uncontroversial

arvidn:
I'm fairly ignorant of how cmake scripts are expected to handle these cases. It seems like something should be done, either 

* updating the cmake script to make it easy to install into a destdir, but still produce correct pkgconfig files
* update documentation or help output to make it simpler to "install into a destdir"

Piraty:
> Your change restores this to it was originally done, here: #2938

not quite. note that the variables are inserted as is instead of using `@...@` to subsitute them during template expansion (which happens at configure stage, thus doesn't know of any user-preferred prefix yet).

I don't have deep cmake wisdom to have any say in *what* variable to use, my fix intends to improve on the timing of variable resolution

Piraty:
are there strongs reasons not to use the classic `libfoo.pc.in` + `configure_file` approach like most cmake projects to?

PR #7824, opened 2025-01-24
update manylinux docker scripts

arvidn:
```
cd tools/cibuildwheel/manylinux
./update-scripts.sh
```


arvidn:
there are a lot of ripple effects of updating this. GCC ICE on the updated OpenSSL

xavier2k6:
Change all instances of `pypa/cibuildwheel@v2.21.3` -> `pypa/cibuildwheel@v2.22.0` to see if it helps with regard to giving more verbose details.

xavier2k6:
> It's not clear to me why `cibuildwheel` needs to build OpenSSL from source.

It may be to do with the `arch`, just like with windows where we had to install 32bit via vcpkg

xavier2k6:
upstream are now using OpenSSL 3.5.0, so re-sync/re-run is in order.

Issue #7789, opened 2024-11-11
IP filter seems not to be applied for DHT traffic


HanabishiRecca:
I noticed that libtorrent still tries to contact filtered IPs. It sends UDP packets as part of an **outgoing** connection, which appears to be DHT traffic.
Fun fact that it does alert about IP being filtered, but still sends the packets regardless. And amount of outgoing UDP packets on my firewall matches with amount of alerts. 

Seeking through the code, I see that at least here a DHT node is added **before** the filter:

https://github.com/arvidn/libtorrent/blob/bb08bdbd8c8ae2355ec79d55de090f894eb6e451/src/torrent.cpp#L11185-L11199

It's hard to track the logic though, so maybe I misunderstood something. @arvidn, could you clarify on that?

HanabishiRecca:
Some testing of my wild guess confirms that. Moving the `add_dht_node` part below all filters actually seems to prevent packages from being sent.

```patch
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -11182,17 +11182,6 @@ namespace {
 
 		TORRENT_ASSERT(info_hash().has_v2() || !(flags & pex_lt_v2));
 
-#ifndef TORRENT_DISABLE_DHT
-		if (source != peer_info::resume_data)
-		{
-			// try to send a DHT ping to this peer
-			// as well, to figure out if it supports
-			// DHT (uTorrent and BitComet don't
-			// advertise support)
-			session().add_dht_node({adr.address(), adr.port()});
-		}
-#endif
-
 		if (m_apply_ip_filter
 			&& m_ip_filter
 			&& m_ip_filter->access(adr.address()) & ip_filter::blocked)
@@ -11218,6 +11207,17 @@ namespace {
 			return nullptr;
 		}
 
+#ifndef TORRENT_DISABLE_DHT
+		if (source != peer_info::resume_data)
+		{
+			// try to send a DHT ping to this peer
+			// as well, to figure out if it supports
+			// DHT (uTorrent and BitComet don't
+			// advertise support)
+			session().add_dht_node({adr.address(), adr.port()});
+		}
+#endif
+
 #if TORRENT_USE_I2P
 		// if this is an i2p torrent, and we don't allow mixed mode
 		// no regular peers should ever be added!

```

arvidn:
https://github.com/arvidn/libtorrent/pull/7829

HanabishiRecca:
Keep

ValdikSS:
[...]

PR #7829, opened 2025-01-25
fix applying IP filter to DHT traffic (HanabishiRecca)

arvidn:
adresses https://github.com/arvidn/libtorrent/issues/7789

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 
 2.0.11 released
 
+	* fix applying IP filter to DHT traffic (HanabishiRecca)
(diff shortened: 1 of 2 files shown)


HanabishiRecca:
I want to mention again that this patch does not block the interaction with banned IPs completely. It's more a partial workaroud.

`dht_tracker` itself simply lacks any filtering logic. I.e. `incoming_packet()` still accepts **all** incoming traffic, `send_packet()` still communicates with a banned peer when it manages to make it into the DHT node list somehow.

arvidn:
right, makes sense

PR #7853, opened 2025-02-09
round-trip torrent via session without relying on torrent_info storin…

arvidn:
…g non-info-section parts


glassez:
I hope all the recent similar changes are aimed at making `torrent_info` immutable and representing only parsed data of torrent info section (as was discussed earlier). So renamed files will also be handled outside of it.

arvidn:
yes, that's exactly right

PR #7866, opened 2025-02-12
deprecate storing piece layers in `torrent_info`

arvidn:
in favor of `add_torrent_params`

### DIFF
--- a/include/libtorrent/torrent_handle.hpp
+++ b/include/libtorrent/torrent_handle.hpp
@@ -923,6 +923,13 @@ namespace aux {
 		// call save_resume_data() and write_torrent_file() the
 		// add_torrent_params object passed back in the alert.
 		//
+		// Note that a torrent added from a magnet link may not have the full
+		// merkle trees for all files, and hence not have the complete piece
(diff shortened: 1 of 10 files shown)


include/libtorrent/torrent_handle.hpp
```diff
@@ -923,6 +923,13 @@ namespace aux {
 		// call save_resume_data() and write_torrent_file() the
 		// add_torrent_params object passed back in the alert.
 		//
+		// Note that a torrent added from a magnet link may not have the full
+		// merkle trees for all files, and hence not have the complete piece
+		// layers. In that state, you cannot create a .torrent file from the
+		// torrent_info returned. Once the torrent completes downloading all
+		// files, becoming a seed, you can make a .torrent file from it.
+		std::shared_ptr<const torrent_info> torrent_file() const;
```
glassez, line 931: I wouldn't call it a `torrent_file` anymore. It's confusing.

arvidn: Yeah, I intend to introduce aliases.

Issue #7855, opened 2025-02-09
Add a direct way to get asynchronous results


glassez:
(It's something a bit like #7558. But unlike it, it should not depend on alerts.)

Now, to get some data from `libtorrent`, we have two options - either use a blocking call or receive the result asynchronously through alerts.
The first way is bad for interactive applications, because it either blocks the main thread, which can cause GUI to hang, or we have to workaround this by using some tricks with additional threads (as if making it asynchronous).
The second way is also inconvenient, as it makes the code confusing and requires additional tricks. IMO, alerts are good for use within the Observer pattern (for example, when tracking some internal events or when it makes sense to have a request and receive data in different piece of logic), but they look unsuitable when requesting data directly.
Earlier, I suggested simply implementing direct access to such data from app thread with mutex-based protection, but this was rejected due to the reluctance to mess with mutexes in `libtorrent` in order to ensure correct multithreaded access. I agree, it could look more complicated than the current event-based access way.
But nevertheless, I would highly like to get a friendlier way of getting the data.
Now I believe that futures could be a good option in this matter. Leaving `libtorrent` to its current method of implementing data access through an event queue, futures could provide the user side with a direct way to obtain the results of these access requests, deciding on its own whether to do it synchronously or asynchronously.
So all methods like `torrent_handle::get_peer_info()`, `torrent_handle::url_seeds()`, `session_handle::is_listening()`, `session_handle::listen_port()`, `torrent_handle::piece_availability()`, `torrent_handle::get_download_queue()`, `torrent_handle::trackers()` (and maybe some other similar ones not mentioned here) should return `std::future` instantiation with appropriate result. (This will also make previously added `post_*` methods and their corresponding alerts unnecessary.)
Some other methods that would make sense to use in different patterns (e.g. `torrent_handle::save_resume_data()`) could have both a return value of type `std::future` and a corresponding alert.

glassez:
Honestly, I was impressed by `QFuture` when I made my suggestions above. I thought `std::future` had comparable functionality. Unfortunately `std::future` turned out to suck completely. But maybe Boost can still provide what I need.

arvidn:
I imagine the main challenge with a future-like object (or any other kind of message/notification) is that it needs to play well with some message loop. I don't think future can easily be combined with other things to wait for. e.g. a socket, a Qt event, epoll, etc.

glassez:
@arvidn 
[...]

Issue #7854, opened 2025-02-09
Notify when all the piece hashes are available


glassez:
Users often want to save a .torrent file for torrents added via magnet link.
There is no problem to do this for v1 torrents once the metadata (info-section) is received.
But for v2/hybrid torrents, we have to wait until all the piece hashes are also available, and we don't have a clear indication of this event.

P.S. To be honest, I have no idea of any straightforward way to save .torrent file for v2 torrents. Right now we are relying on the torrent finished event, after which we request `torrent_file_with_hashes`, which I don't like because it is a blocking call. In addition, we use `create_torrent` in a deprecated way, which will become unavailable later (BTW, `torrent_file_with_hashes` should also be deprecated). Otherwise, we should wait for the torrent to finish, then request resume data, and wait for the appropriate alert (along the way, filtering out such alerts that may have been requested earlier than the torrent was finished).

glassez:
> In addition, we use `create_torrent` in a deprecated way, which will become unavailable later (BTW, `torrent_file_with_hashes` should also be deprecated).

Well, there's `torrent_handle::piece_layers()` that we could use instead of `torrent_handle::torrent_file_with_hashes()`. But the question remains, is there a simple and reliable way to determine when complete piece layers are available?

glassez:
> Well, there's `torrent_handle::piece_layers()` that we could use instead of `torrent_handle::torrent_file_with_hashes()`.

Damn it! Looks like it won't work. `write_torrent_file()` still need the full merkle tree representation passed within `add_torrent_params` and we still have no way to get it except through `add_torrent_params` requested through `torrent_handle::save_resume_data()` (or by manually filling it from result of `torrent_handle::piece_layers()` which is counter productive).

PR #7869, opened 2025-02-14
Add ability to get resume data synchronously

glassez:


glassez:
@arvidn
I don't know if you're notified of new Issues/PRs immediately, or if you review them from time to time, so I preferred to call you explicitly.
I would like to draw your attention to #7855 and get at least a general feedback on it as early as possible. I don't want to be a "train lagger" if it could have been approved, but it was noticed too late. I could contribute to its implementation myself, but I don't want to start work without knowing if it has a chance to be merged.

arvidn:
I think this PR looks good. It would be nice to expose it to python, add a test to exercise it and I don't think you need the `static` empty `add_torrent_params` object, since you're returning it by value. (had the return value been by reference on the other hand, you would have needed it).

glassez:
> Probing the ground in the direction of solving #7855...

Unfortunately, I got my cards mixed up in a hurry. These changes are not really an attempt to solve #7855, but just a quick workaround for some problems related to generating torrent file as mentioned in #7854.

include/libtorrent/torrent_handle.hpp
```diff
@@ -844,6 +845,8 @@ namespace aux {
 		bool need_save_resume_data() const;
 		bool need_save_resume_data(resume_data_flags_t flags) const;
 
+		add_torrent_params get_resume_data(resume_data_flags_t flags = {}) const;
```
glassez, line 848: @arvidn 
Does it make sense to support flags in this function? The only flag that makes sense when applied to it is `save_info_dict`. But the point is that for some time now it only assign a pointer to `add_torrent_params`. What could be the problem of doing this unconditionally? The caller could just clear it if it is not needed. (IMO, `save_info_dict` could be even deprecated at all.)

glassez:
> I don't think you need the `static` empty `add_torrent_params` object, since you're returning it by value. (had the return value been by reference on the other hand, you would have needed it).

Ok, changed.


> It would be nice to expose it to python

I don't know anything about it at all. At least where is it?

>add a test to exercise it

I could add something similar to `save_resume_data()`, but I found a lot of occurrences of it in the test sources. I doubt what exactly I should do about it.

arvidn (approved):

arvidn:
I can add documentation and a Changelog entry

arvidn:
https://github.com/arvidn/libtorrent/pull/7972

PR #7870, opened 2025-02-14
cirrus: Increment freebsd image

xavier2k6:
* Bumped freebsd image to 14.2 (confirmed in different PR that it's actually available/working)


arvidn:
I suspect these new warnings (treated as errors) will need to be dealt with too

PR #7873, opened 2025-02-15
Fix arguments of dispatched functions cannot be actually moved

glassez:
It seems that move-only arguments were never used, so the problem went unnoticed. I discovered it when I was experimenting with `boost::promise`.

### DIFF
--- a/src/session_handle.cpp
+++ b/src/session_handle.cpp
@@ -3,6 +3,7 @@
 Copyright (c) 2014-2018, Steven Siloti
 Copyright (c) 2015-2018, Alden Torres
 Copyright (c) 2015-2022, Arvid Norberg
+Copyright (c) 2025, Vladimir Golovnev (glassez)
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
@@ -92,12 +93,12 @@ namespace libtorrent {
 	{
 		std::shared_ptr<session_impl> s = m_impl.lock();
 		if (!s) aux::throw_ex<system_error>(errors::invalid_session_handle);
-		dispatch(s->get_context(), [=]() mutable
+		dispatch(s->get_context(), std::bind([s, f](auto&&... args) mutable
 		{
 #ifndef BOOST_NO_EXCEPTIONS
 			try {
 #endif
-				(s.get()->*f)(std::forward<Args>(a)...);
+				(s.get()->*f)(std::forward<Args>(args)...);
 #ifndef BOOST_NO_EXCEPTIONS
 			} catch (system_error const& e) {
 				s->alerts().emplace_alert<session_error_alert>(e.code(), e.what());
@@ -107,7 +108,7 @@ namespace libtorrent {
 				s->alerts().emplace_alert<session_error_alert>(error_code(), "unknown error");
 			}
 #endif
-		});
+		}, std::forward<Args>(a)...));
 	}
 
 	template<typename Fun, typename... Args>
@@ -122,12 +123,12 @@ namespace libtorrent {
 		bool done = false;
 
 		std::exception_ptr ex;
-		dispatch(s->get_context(), [=, &done, &ex]() mutable
+		dispatch(s->get_context(), std::bind([s, f, &done, &ex](auto&&... args) mutable
 		{
 #ifndef BOOST_NO_EXCEPTIONS
 			try {
 #endif
-				(s.get()->*f)(std::forward<Args>(a)...);
+				(s.get()->*f)(std::forward<Args>(args)...);
 #ifndef BOOST_NO_EXCEPTIONS
 			} catch (...) {
 				ex = std::current_exception();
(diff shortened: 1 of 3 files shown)


glassez:
Oh, damn it! `libtorrent` still supports C++17...

src/session_handle.cpp
```diff
@@ -426,8 +426,7 @@ namespace {
 		handle_backwards_compatible_resume_data(params);
 #endif
 		error_code ec;
-		auto ecr = std::ref(ec);
-		torrent_handle r = sync_call_ret<torrent_handle>(&session_impl::add_torrent, std::move(params), ecr);
+		torrent_handle r = sync_call_ret<torrent_handle>(&session_impl::add_torrent, std::move(params), std::ref(ec));
```
glassez, line 429:
```suggestion
		auto ecr = std::ref(ec);
		torrent_handle r = sync_call_ret<torrent_handle>(&session_impl::add_torrent, std::move(params), ecr);
```

src/torrent_handle.cpp
```diff
@@ -796,8 +795,7 @@ namespace libtorrent {
 	entry torrent_handle::write_resume_data() const
 	{
 		add_torrent_params params;
-		auto retr = std::ref(params);
-		sync_call(&torrent::write_resume_data, resume_data_flags_t{}, retr);
+		sync_call(&torrent::write_resume_data, resume_data_flags_t{}, std::ref(params));
```
glassez, line 798:
```suggestion
		auto retr = std::ref(params);
				sync_call(&torrent::write_resume_data, resume_data_flags_t{}, retr);
```

src/session_handle.cpp
```diff
@@ -92,12 +93,12 @@ namespace libtorrent {
 	{
 		std::shared_ptr<session_impl> s = m_impl.lock();
 		if (!s) aux::throw_ex<system_error>(errors::invalid_session_handle);
-		dispatch(s->get_context(), [=]() mutable
+		dispatch(s->get_context(), std::bind([=](auto&&... args) mutable
```
glassez, line 96: Note that trick with std::bind isn't required since c++20. It can be rewritten as the following:
```c++
dispatch(s->get_context(), [=, ...args = std::forward<Args>(a)]() mutable
{
    // do something
});
```

test/test_storage.cpp
```diff
@@ -928,11 +928,12 @@ void test_fastresume(bool const test_deprecated)
 		p.storage_mode = storage_mode_sparse;
 		error_code ignore;
 		torrent_handle h = ses.add_torrent(std::move(p), ignore);
-		TEST_CHECK(exists(combine_path(p.save_path, "temporary")));
```
glassez, line 931: It is noteworthy that this "use-after-move" bug also went unnoticed, as `p` was not actually moved.

src/session_handle.cpp
```diff
@@ -92,12 +93,12 @@ namespace libtorrent {
 	{
 		std::shared_ptr<session_impl> s = m_impl.lock();
 		if (!s) aux::throw_ex<system_error>(errors::invalid_session_handle);
-		dispatch(s->get_context(), [=]() mutable
+		dispatch(s->get_context(), std::bind([s, f](auto&&... args) mutable
```
Chocobo1, line 96: It would be beneficial to move function `f` into the lambda too if possible.

Chocobo1, line 96: Why not:
```suggestion
		dispatch(s->get_context(), std::bind([s, f](Args&&... args) mutable
```

glassez: > It would be beneficial to move function `f` into the lambda too if possible.

Generally, yes. But in this case, it is intended only for pointer to member function, so moving it will not add any advantages.

glassez: > Why not:

It is incorrect.

Chocobo1: I see. So it is a pointer rather than a function object...

arvidn, line 96: this is to emulate `[s, f, args = std::move(a)]`, or do you always need `std::bind()` since `a` is a parameter pack?

glassez: This is to emulate `[s, f, ...args = std::forward<Args>(a)]` since explicit capturing of parameters pack is available only since c++20.

glassez: @arvidn 
So what are you going to do about this PR?

arvidn: sorry for the delay. I'll take a deeper look today.

glassez:
> It might also be nice to accomodate the shadowing warnings by using a different name. It looks like some of the warnings are just bugs in the CodeQL C++ parser though

I can't find one that isn't "just bug in the CodeQL C++ parser", can you?

arvidn:
> I can't find one that isn't "just bug in the CodeQL C++ parser", can you?

I cannot.

arvidn (approved):

arvidn:
cirrus is building the branch directly (`cc0b207e9243f42263023e3072e9f77c1aaca01a`), rather than the branch merged with `RC_2_0` (like github does). It's a bit disappointing, but I suppose it'll be green once it lands

glassez:
@arvidn 
What is c++ requirements for master branch? Does it still support c++17 or below?

glassez:
> @arvidn What is c++ requirements for master branch? Does it still support c++17 or below?

Well, according to `CMakeLists.txt`, libtorrent master still supports c++17, so I ported this fix as it is (i.e. using `std::bind` as a workaround).
#7903

PR #7887, opened 2025-02-22
AppVeyor: Bump python to 3.8

xavier2k6:


xavier2k6:
@arvidn Why does AppVeyor choose `OpenSSL 1.0.2u` instead of latest available 1.1.1 version on image?, same happens with RC_1_2 Branch........

arvidn (approved):
even python 3.8 is EOL now though, isn't it?

Issue #7681, opened 2024-05-31
Add support for overriding the NAT-PMP gateway


farcaller:
libtorrent version (or branch): 1.2.19

platform/architecture: linux x86_64

compiler and compiler version: n/a

please describe what symptom you see, what you would expect to see instead and
how to reproduce it.

libtorrent always tries to resolve the default gateway via the routes available. If the nat-pmp provider isn't actually the direct gateway (in e.g. a containerized environment), then the port mapping fails.

I suggest to add a setting that allows a client to override the gateway. In fact, I made a test implementation of this (https://github.com/farcaller/libtorrent/tree/natpmp) and verified it working with a patched qBitTorrent version. I'm just not sure how to make a proper PR for this, as qBitTorrent doesn't seem to be ready for 2.0 yet and thus I wrote the diff against 1.2.19.

samvv:
Thank you very much @farcaller, I needed this.

I've made a commit that syncs with the latest commit on `RC_2_0`. Maybe this can become a pull request? @arvidn would you accept a PR?

rptaylor:
Looks like this was addressed by https://github.com/arvidn/libtorrent/pull/8154

PR #7890, opened 2025-02-24
Add support for overriding the NAT-PMP gateway with a nat_pmp_gateway config option

samvv:
As discussed in #7681, this pull request updates farcaller's changes to the latest commit on RC_2_0.

**This hasn't been thoroughly been tested yet.** This seems to fix users behind ProtonVPN not being able to seed due to the default gateway not being set. However, it's difficult to tell because I fail to find a good torrent to test (I used the Arch Linux image). I sometimes get speeds in the kilobytes instead of plain nothing, which is better than nothing but still not ideal.

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -5676,7 +5676,11 @@ namespace {
 			ip.netmask = s->netmask;
 			std::strncpy(ip.name, s->device.c_str(), sizeof(ip.name) - 1);
 			ip.name[sizeof(ip.name) - 1] = '\0';
-			s->natpmp_mapper->start(ip);
+
+			error_code ignored;
+			std::string gateway = m_settings.get_str(settings_pack::nat_pmp_gateway);
+			boost::optional<address> gateway_addr = make_address(gateway, ignored);
+			s->natpmp_mapper->start(ip, gateway == "" ? boost::none : gateway_addr);
 		}
 	}
 
--- a/include/libtorrent/natpmp.hpp
+++ b/include/libtorrent/natpmp.hpp
@@ -91,7 +91,7 @@ struct TORRENT_EXTRA_EXPORT natpmp final
 {
 	natpmp(io_context& ios, aux::portmap_callback& cb, aux::listen_socket_handle ls);
 
-	void start(ip_interface const& ip);
+	void start(ip_interface const& ip, boost::optional<address> const& gateway);
 
 	// maps the ports, if a port is set to 0
 	// it will not be mapped
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -391,6 +391,11 @@ namespace aux {
 			// ``router.bt.ouinet.work:6881``,
 			dht_bootstrap_nodes,
 
+			// Overrides the NAT-PMP service gateway. When set, libtorrent won't try
+			// to resolve the default gateway and instead will send the requests to
+			// the address specified.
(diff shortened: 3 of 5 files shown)


src/session_impl.cpp
```diff
@@ -5676,7 +5676,11 @@ namespace {
 			ip.netmask = s->netmask;
 			std::strncpy(ip.name, s->device.c_str(), sizeof(ip.name) - 1);
 			ip.name[sizeof(ip.name) - 1] = '\0';
-			s->natpmp_mapper->start(ip);
+
+			error_code ignored;
+			std::string gateway = m_settings.get_str(settings_pack::nat_pmp_gateway);
+			boost::optional<address> gateway_addr = make_address(gateway, ignored);
```
arvidn, line 5682: It would make sense to move the call to `make_address()` into the condition, to only make it in case the setting is set

arvidn:
This seems to warrant a note in the ChangeLog as well

src/session_impl.cpp
samvv: That gives the following error: `Incompatible operand types ('const none_t' and 'address')`. I'll make it an if-statement, because I don't see how to get `boost::none` working in a ternary.

samvv:
@arvidn Anything else that needs to happen? Let me know :)

samvv:
Just a friendly ping to ask for a merge :)

samvv:
I just tested it out and can confirm this works with ProtonVPN.

qBittUser:
Currently GitHub is showing that this PR's ChangeLog file has same change written twice in different lines (12 & 14).

https://github.com/arvidn/libtorrent/pull/7890/files

https://github.com/arvidn/libtorrent/blob/706849f4255a377c1a42f25545d629584da0e8e2/ChangeLog#L12-L14

samvv:
Oh damn good catch, sorry! I'll fix it when i have the time.

BenSmith:
I'd love to see this make it in.

Neustradamus:
@arvidn: Any progress on this PR?

samvv:
Sorry I suck at Git 🙃 I think this should do it?

I'm also creating a Docker container with this patch. I'll keep you posted.

samvv:
A temporary Docker image is now available to test it out: https://hub.docker.com/r/samvv/qbittorrent

qBittUser:
> I think this should do it?

There's still at least 1 commit in your RC_2_0 branch that seems totally unnecessary, because * https://github.com/arvidn/libtorrent/commit/4fc6e833c004f4aa7e1891655ba98ca41e43b19e was already merged long time ago, yet you somehow added the change again, which is why your PR [files changed](https://github.com/arvidn/libtorrent/pull/7890/files) page still shows a change being mentioned twice in changelog file.

https://github.com/arvidn/libtorrent/blob/5c6bfd47e0136150e5dab57aa17d208bef6109a7/ChangeLog#L12-L14

Line 12 and 14 are same, which is unnecessary.

Should've used GitHub Desktop software if possible, could've easily deleted the excess commits and then modified existing ones if see a need to avoid adding extra commits.

If difficult to delete unnecessary commits, then maybe wait for @arvidn to review and to specify if can include only necessary ones when merging?

samvv:
Sorry about that. If I find the time I'll try to fix it one of these days. I'll start with a clean branch and do a cherry-pick.

samvv:
I've created a fresh pull request at https://github.com/arvidn/libtorrent/pull/8144

Issue #7894, opened 2025-02-27
CPU usage/efficiency


bolshoytoster:
I'm a user of qBittorrent, and I've been experiencing high CPU usage (30-50%) recently, mostly from libtorrent's network thread.

I decided to investigate this with callgrind, and I found that ~17% of the total CPU usage was in [`parse_url_components`](https://github.com/arvidn/libtorrent/blob/c31d90b59ffb3910b6d36ba7ad914e0e489d2b6c/src/parse_url.cpp#L43-L148), specifically in the calls to `std::find`, which, it seems aren't being properly inlined/optimized (yes, I'm using `-O3`).

I modified this to use raw pointers and `memchr` instead of iterators and `std::find`, and I've been able to get a 1.6x speed improvement (according to my crude benchmarking):
```c++
std::tuple<std::string, std::string, std::string, int, std::string>
	parse_url_components(std::string url, error_code& ec)
{
	std::string hostname; // hostname only
	std::string auth; // user:pass
	std::string protocol; // http or https for instance
	int port = -1;

	const char *at;
	const char *colon;
	const char *slash;
	const char *question_mark;
	const char *hash;
	const char *port_pos;

	// PARSE URL
	const char *start = url.data();
	const char *str_end = start + url.length();
	// remove white spaces in front of the url
	while (start != str_end && is_space(*start))
		++start;
	const char *end = (const char *) memchr(start, ':', str_end - start);
	protocol.assign(start, end);

	if (!end)
	{
		ec = errors::unsupported_url_protocol;
		goto exit;
	}
	++end;
	if (end == str_end || *end != '/')
	{
		ec = errors::unsupported_url_protocol;
		goto exit;
	}
	++end;
	if (end == str_end || *end != '/')
	{
		ec = errors::unsupported_url_protocol;
		goto exit;
	}
	++end;
	start = end;

	colon = (const char *) memchr(start, ':', str_end - start);
	at = colon ? (const char *) memchr(colon, '@', str_end - colon) : 0;
	slash = (const char *) memchr(start, '/', str_end - start);
	question_mark = (const char *) memchr(start, '?', str_end - start);
	hash = (const char *) memchr(start, '#', str_end - start);
	end = slash
		? question_mark
			? hash
				? std::min({slash, question_mark, hash})
				: std::min(slash, question_mark)
			: hash
				? std::min(slash, hash)
				: slash
		: question_mark
			? hash
				? std::min(question_mark, hash)
				: question_mark
			: hash
[...]

Issue #7529, opened 2023-10-09
[Feature request] Introduce ECH (Encrypted Client Hello) support for HTTPS-enabled announcers


giantplaceholder:
A sort of follow-up of https://github.com/arvidn/libtorrent/issues/5222

### Suggestion

As far as I understand, right now libtorrent does support HTTPS-enabled announcers.

I'm wondering whether you would consider to introduce ECH support for them as well.

In some countries -- in this particular example, Russia -- government telecommunication wachdogs are enforcing numerous ways to block communication with certain web-sites, including torrent trackers and their infrastructure, leveraging the fact that SNI and ALPN during TLS handshake are left unencrypted:

![image](https://github.com/qbittorrent/qBittorrent/assets/4792863/7b197088-0eec-4c0f-82d2-50c5411f00a7)

This allows gov't firewalls and traffic sniffing solutions to effectively block communications with web-site or announcer even when they are hiding behind Cloudflare (or other domain-front or CDN).

ECH sole purpose is to rectify that, and it is currently being deployed en masse by Cloudflare, Mozilla and Google with their Chrome browser. It is also supported by various middleware, such as openssl, boringssl, nginx, Apache, haproxy and so on.

### Use case

Let's assume that third party can monitor our traffic on the go and intercept TLS handshakes:

0) We're using libtorrent-powered client, like qBittorrent, with some private torrent in it (no DHT, no LSD, no PeX)
1) Torrent client tries to communicate with HTTPS-enabled announcer currently "banned" by gov't: bt.announcer.tld
2) In this example, bt.announcer.tld is hiding behind Cloudflare (for whatever reasons -- for instance, Rutracker does this) and as such, it is ECH-enabled
3) DPI solution on the ISP side intercepts TLS handshake and parses SNI, extracting our destination: "bt.announcer.tld"
4) DPI solution drops all traffic to that CF endpoint from that specific client containing requests for bt.announcer.tld
4) Since this torrent is private, torrent client is unable to learn about any peers and download is effectively stalled

(this is very short and not entirely correct way of describing how Russia's country-wide filtering works for sites behind CF, Akamai and Fastly, without actually black-holing all traffic originating to and from those CDNs)

This might be an even worse issue for the end user if they are using a torrent site which keeps track of users ratio and limits user's ability to download should their ratio fall below some arbitrary threshold -- because torrent client if is not able to communicate with announcer, it won't be able to report user's stats back to the site.

[...]

PR #7898, opened 2025-03-01
A few optimizations to lower CPU usage

bolshoytoster:
As mentioned in #7894, libtorrent's CPU usage is higher than it could be. This PR will get rid of some inefficiencies:

[`is_i2p_url`](https://github.com/arvidn/libtorrent/blob/c31d90b59ffb3910b6d36ba7ad914e0e489d2b6c/src/string_util.cpp#L401-L409) is called thousands of times per second, with the same urls, mainly by [`torrent::announce_with_tracker`](https://github.com/arvidn/libtorrent/blob/c31d90b59ffb3910b6d36ba7ad914e0e489d2b6c/src/torrent.cpp#L2967-L3312) and [`refresh_endpoint_list`](https://github.com/arvidn/libtorrent/blob/c31d90b59ffb3910b6d36ba7ad914e0e489d2b6c/src/torrent.cpp#L2894-L2937). `is_i2p_url` parses the given url, just to check the end of the hostname. This PR speeds it up by ~70% for cached urls (most of them).

Note: `torrent::announce_with_tracker` calls `refresh_endpoint_list` right before calling `is_i2p_url` itself, essentially meaning `is_i2p_url` is always called twice. We could halve it's CPU usage again by just calling it once and passing the result to `refresh_endpoint_list`, although this would require adding an argument that only exists with `TORRENT_USE_I2P` (I'm not sure how this repo feels about that, but it would increase efficiency).

Speaking of `refresh_endpoint_list`, I noticed that [the first `std::swap`](https://github.com/arvidn/libtorrent/blob/c31d90b59ffb3910b6d36ba7ad914e0e489d2b6c/src/torrent.cpp#L2925) was often (pretty much all of the time) called to swap an element with itself. This caused a fairly large amount of CPU usage, since it doesn't look like `std::swap` checks for this case. I've added a simple check to prevent this.

Another thing mentioned in the issue is that `torrent::announce_with_tracker` and [`torrent::update_tracker_timer`](https://github.com/arvidn/libtorrent/blob/c31d90b59ffb3910b6d36ba7ad914e0e489d2b6c/src/torrent.cpp#L9878C1-L10040) call `session_settings::get_bool` many times, nested 3 loops deep. This was horrible, since `session_settings::get_bool` locks and unlocks a mutex on every call. I've moved these to before the loop to remedy this.

Finally, the biggest issue now is in `torrent::update_tracker_timer` and `torrent::announce_with_tracker`, which search through `listen_socket_states` for every endpoint of every tracker. I've made this slightly better by using a `std::map` (can't use a `std::unordered_map` since `std::weak_ptr` can't be hashed) instead of a `std::vector`. It's still the biggest issue, and I'd like to see if there's a way to reduce the amount of searching that goes on, since, at least on my machine, every tracker has the same 5 endpoints anyway.

### DIFF
--- a/src/string_util.cpp
+++ b/src/string_util.cpp
@@ -44,6 +44,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <cstdlib> // for malloc
 #include <cstring> // for strlen
 #include <algorithm> // for search
+#include <unordered_map>
 
 namespace libtorrent {
 
@@ -400,12 +401,21 @@ namespace libtorrent {
 
 	bool is_i2p_url(std::string const& url)
 	{
+		static std::unordered_map<std::string, bool> cache;
+		auto i2p = cache.find(url);
+		if (i2p != cache.end())
+		{
+			return i2p->second;
+		}
+
 		using std::ignore;
 		std::string hostname;
 		error_code ec;
 		std::tie(ignore, ignore, hostname, ignore, ignore)
 			= parse_url_components(url, ec);
-		return string_ends_with(hostname, ".i2p");
+		bool ends_with = string_ends_with(hostname, ".i2p");
+		cache.insert({url, ends_with});
+		return ends_with;
 	}
 
 #endif
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -2922,7 +2922,7 @@ namespace {
 			for (auto& aep : aeps)
 			{
 				if (aep.socket != s) continue;
-				std::swap(aeps[valid_endpoints], aep);
+				if (&aeps[valid_endpoints] != &aep) std::swap(aeps[valid_endpoints], aep);
 				valid_endpoints++;
 				return;
 			}
@@ -3126,6 +3126,8 @@ namespace {
 		}
 #endif
 
+		bool announce_to_all_tiers = settings().get_bool(settings_pack::announce_to_all_tiers);
+		bool announce_to_all_trackers = settings().get_bool(settings_pack::announce_to_all_trackers);
 		for (auto& ae : m_trackers)
 		{
 #ifndef TORRENT_DISABLE_LOGGING
@@ -3195,15 +3197,15 @@ namespace {
 					}
 #endif
 
-					if (settings().get_bool(settings_pack::announce_to_all_tiers)
-						&& !settings().get_bool(settings_pack::announce_to_all_trackers)
+					if (announce_to_all_tiers
+						&& !announce_to_all_trackers
 						&& state.sent_announce
 						&& ae.tier <= state.tier
 						&& state.tier != INT_MAX)
 						continue;
 
 					if (ae.tier > state.tier && state.sent_announce
-						&& !settings().get_bool(settings_pack::announce_to_all_tiers)) continue;
+						&& !announce_to_all_tiers) continue;
 					if (a.is_working()) { state.tier = ae.tier; state.sent_announce = false; }
 					if (!a.can_announce(now, is_seed(), ae.fail_limit))
 					{
@@ -3289,8 +3291,8 @@ namespace {
 
 					state.sent_announce = true;
 					if (a.is_working()
-						&& !settings().get_bool(settings_pack::announce_to_all_trackers)
-						&& !settings().get_bool(settings_pack::announce_to_all_tiers))
+						&& !announce_to_all_trackers
+						&& !announce_to_all_tiers)
 					{
 						state.done = true;
 					}
@@ -9901,7 +9903,7 @@ namespace {
 
 		time_point32 next_announce = time_point32::max();
 
-		std::vector<timer_state> listen_socket_states;
+		std::map<std::weak_ptr<aux::listen_socket_t>, timer_state, std::owner_less<std::weak_ptr<aux::listen_socket_t>>> listen_socket_states;
 
 #ifndef TORRENT_DISABLE_LOGGING
 		int idx = -1;
@@ -9914,6 +9916,8 @@ namespace {
 				, int(m_trackers.size()));
 		}
 #endif
+		bool announce_to_all_tiers = settings().get_bool(settings_pack::announce_to_all_tiers);
+		bool announce_to_all_trackers = settings().get_bool(settings_pack::announce_to_all_trackers);
 		for (auto const& t : m_trackers)
 		{
 #ifndef TORRENT_DISABLE_LOGGING
@@ -9921,14 +9925,12 @@ namespace {
 #endif
 			for (auto const& aep : t.endpoints)
 			{
-				auto aep_state_iter = std::find_if(listen_socket_states.begin(), listen_socket_states.end()
-					, [&](timer_state const& s) { return s.socket == aep.socket; });
+				auto aep_state_iter = listen_socket_states.find(aep.socket.get_ptr());
 				if (aep_state_iter == listen_socket_states.end())
 				{
-					listen_socket_states.emplace_back(aep.socket);
-					aep_state_iter = listen_socket_states.end() - 1;
+					aep_state_iter = listen_socket_states.insert({aep.socket.get_ptr(), timer_state(aep.socket)}).first;
 				}
-				timer_state& ep_state = *aep_state_iter;
+				timer_state& ep_state = aep_state_iter->second;
 
 				if (!aep.enabled) continue;
 				for (protocol_version const ih : all_versions)
@@ -9952,8 +9954,8 @@ namespace {
 					}
 #endif
 
-					if (settings().get_bool(settings_pack::announce_to_all_tiers)
-						&& !settings().get_bool(settings_pack::announce_to_all_trackers)
+					if (announce_to_all_tiers
+						&& !announce_to_all_trackers
 						&& state.found_working
 						&& t.tier <= state.tier
 						&& state.tier != INT_MAX)
@@ -9961,7 +9963,7 @@ namespace {
 
 					if (t.tier > state.tier)
 					{
-						if (!settings().get_bool(settings_pack::announce_to_all_tiers)) break;
+						if (!announce_to_all_tiers) break;
 						state.found_working = false;
 					}
 					state.tier = t.tier;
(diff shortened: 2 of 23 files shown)


bolshoytoster:
I can get a bigger boost in `torrent::update_tracker_timer` by sticking to a `std::vector` and playing on the fact that most trackers use the same list of endpoints in the same order, by replacing
https://github.com/arvidn/libtorrent/blob/c31d90b59ffb3910b6d36ba7ad914e0e489d2b6c/src/torrent.cpp#L9922-L9930

with
```c++
			std::vector<timer_state>::iterator start = listen_socket_states.begin();
			for (auto const& aep : t.endpoints)
			{
				std::vector<timer_state>::iterator aep_state_iter;
				if (start == listen_socket_states.end())
				{
					listen_socket_states.emplace_back(aep.socket);
					aep_state_iter = listen_socket_states.end() - 1;
					start = listen_socket_states.end();
				}
				else if (start->socket == aep.socket)
				{
					aep_state_iter = start++;
				}
				else
				{
					aep_state_iter = std::find_if(start + 1, listen_socket_states.end()
						, [&](timer_state const& s) { return s.socket == aep.socket; });
					if (aep_state_iter == listen_socket_states.end())
					{
						listen_socket_states.emplace_back(aep.socket);
						aep_state_iter = listen_socket_states.end() - 1;
						start = listen_socket_states.begin();
					}
				}
```

It works by shifting the search's starting point whenever an endpoint was found, meaning it doesn't keep searching the same endpoints each time. In the best case, the tracker's endpoints are in the same order as `listen_socket_states`, and only runs the else if block. In the worst case, the endpoints are different, and it falls back to `std::find__if`.

This assumes that a tracker will never have duplicate endpoints.

Should I add this to the PR instead of the change to `std::map`, or is the code too messy?

bolshoytoster:
Also, the calls to `listen_socket_handle::is_ssl()` in `refresh_endpoint_list` are quite expensive, due to the `std::weak_ptr::lock`. Perhaps `listen_socket_t::ssl` could be cached in `listen_socket_handle`, or maybe `listen_socket_handle` could store a `std::shared_ptr` instead.

bolshoytoster:
Another thing I've found is that [`torrent::tracker_request_error` calls `torrent::announce_with_tracker` then `torrent::update_tracker_timer`](https://github.com/arvidn/libtorrent/blob/c31d90b59ffb3910b6d36ba7ad914e0e489d2b6c/src/torrent.cpp#L12308-L12313) even though `torrent::announce_with_tracker` already calls `torrent::update_tracker_timer`. Effectively meaning it's called twice for no reason. I've prevented this by only calling `torrent::update_tracker_timer` if `torrent::announce_with_tracker` wasn't called.

src/string_util.cpp
```diff
@@ -400,12 +401,21 @@ namespace libtorrent {
 
 	bool is_i2p_url(std::string const& url)
 	{
+		static std::unordered_map<std::string, bool> cache;
```
Chocobo1, line 404: I'm not sure my concern is valid but anyway:
1. What do you think about putting an upper limit on how many entries in the cache? IMO it isn't ideal for it to be unlimited. I suppose `1024` would suffice for most cases. Also this implies it needs some kind of cache replacement policy, for example LRU. Maybe libtorrent/boost provides such container?
2. If I understand correctly, using `static` means it will be shared across all libtorrent sessions which isn't ideal either. It should be local to a specific torrent/session or whatever scope that make sense.
3. Maybe this function requires to be thread-safe.

src/torrent.cpp
```diff
@@ -3114,14 +3114,16 @@ namespace {
 		// so that each one should get at least one announce
 		std::vector<announce_state> listen_socket_states;
 
+		bool announce_to_all_tiers = settings().get_bool(settings_pack::announce_to_all_tiers);
+		bool announce_to_all_trackers = settings().get_bool(settings_pack::announce_to_all_trackers);
```
Chocobo1, line 3118: Not sure about this repo policy on `const` but I would suggest adding them anyway:
```suggestion
		bool const announce_to_all_tiers = settings().get_bool(settings_pack::announce_to_all_tiers);
		bool const announce_to_all_trackers = settings().get_bool(settings_pack::announce_to_all_trackers);
```

Chocobo1 (commented):
Some preliminary review comments. The project author may have different opinion though.

src/string_util.cpp
bolshoytoster: 1. It looks like libtorrent implements an LRU cache in [src/file_view_pool.cpp](https://github.com/arvidn/libtorrent/blob/RC_2_0/src/file_view_pool.cpp), but that's specific to the file pool. [Boost also has an implementation](https://www.boost.org/doc/libs/1_67_0/boost/compute/detail/lru_cache.hpp), but it's under ::compute, so it can't be used here. I could implement one, but I'm not sure if it's needed - my qbittorrent uses 242 trackers urls in total, being ~8kb in text. Round it up to 10kb to account for `std::unordered_map`'s overhead, and it shouldn't get that big.

2 & 3. I hadn't considered multiple threads, but I think I could just use `thread_local` instead of `static`.

src/torrent.cpp
bolshoytoster: Whoops, I'm used to rust, where things are const by default.

bolshoytoster:
Currently, I'm seeing a fair amount of CPU usage in `announce_infohash::can_announce`, mostly because it's having to convert between `time_point` and `time_point32`. This could be improved if we just chose one and switched all APIs to it, but I recognize that may not be possible.

I could also cache the return value between iterations in `torrent::announce_with_tracker`, since it shouldn't change.

src/torrent.cpp
```diff
@@ -12310,7 +12312,10 @@ namespace {
 		{
 			announce_with_tracker(r.event);
```
arvidn, line 12313: if this call fails, `update_tracker_timer()` won't be called, with your patch. I'm not sure exactly what the consequence of that is, but it's possible the tracker will be tried over and over again in quick succession.
I would feel better about this change if you would check the return value here and call `update_tracker_timer()` if it fails. Or perhaps even better, make `announce_with_tracker()` update the timer unconditionally.

src/string_util.cpp
arvidn: I think this cache is the only controversial part of this patch. It seems more natural to record whether a tracker is i2p or not once, and not call this function repeatedly. It's only 1 bit of information.
The ABI issue is a challenge though. in `master` the announce entries have been separated into internal and exported types, so it's easy to add internal fields without affecting the ABI

arvidn (commented):
Looks good apart from the `is_i2p_url()` cache, I think it needs some more considerations

src/string_util.cpp
bolshoytoster: Since it's only one bit, could it be put in padding, [like `verified`](https://github.com/arvidn/libtorrent/blob/9eaa8522cb6da2cb2ae8277d95b183bcfc5adbc9/include/libtorrent/aux_/announce_entry.hpp#L192)? Or could that cause some issues?

src/torrent.cpp
bolshoytoster: `torrent::announce_with_tracker` returns `void`, so there's no way to know if it fails.

In all cases where `torrent::announce_with_tracker` returns early (fails), there are either no trackers, or libtorrent isn't announcing. In both cases, `torrent::update_tracker_timer` would do nothing anyway, so I don't think it matters.

src/torrent.cpp
```diff
@@ -2915,21 +2915,36 @@ namespace {
 
 		// update the endpoint list by adding entries for new listen sockets
 		// and removing entries for non-existent ones
+		std::vector<aux::announce_endpoint>::iterator start = aeps.begin();
```
arvidn, line 2918:
```suggestion
		auto start = aeps.begin();
```

src/torrent.cpp
```diff
@@ -2915,21 +2915,36 @@ namespace {
...
+				if (start->socket == s)
+				{
+					if (&aeps[valid_endpoints] != &*start) std::swap(aeps[valid_endpoints], *start);
+					valid_endpoints++;
+					++start;
+					return;
+				}
+				else
+				{
+					for (std::vector<aux::announce_endpoint>::iterator aep = start + 1; aep != aeps.end(); ++aep)
```
arvidn, line 2934:
```suggestion
					for (auto aep = start + 1; aep != aeps.end(); ++aep)
```

arvidn (commented):
the last commit appears to be independent of the other changes. It should probably be a separate PR. Also, the same pattern repeats 3 times, it would be good to factor that out somehow

bolshoytoster:
> the last commit appears to be independent of the other changes. It should probably be a separate PR. Also, the same pattern repeats 3 times, it would be good to factor that out somehow

Good point, I've dropped the commit, I'll open a new PR with it when this one closes.

bolshoytoster:
After ignoring `lt::errors::announce_skipped`, the biggest hotspots are [the calls](https://github.com/arvidn/libtorrent/blob/0afd8a7a480902273dc6dd75744f36644988a376/src/pe_crypto.cpp#L103C24-L103C28) [to `mp::powm`](https://github.com/arvidn/libtorrent/blob/0afd8a7a480902273dc6dd75744f36644988a376/src/pe_crypto.cpp#L118) in the Diffie Hellman exchanges. This may be more of a general bittorrent issue, but this seems quite expensive. Couldn't bittorrent's encryption be replaced by an encrypted client hello?

arvidn:
I would land most of these changes right now, but they're being held up by the (small) controversial change of the i2p-cache. A good appraoch is to make smaller independent PRs, to allow fixes to land independently

arvidn:
and thanks for your contribution! I'm interested in experiencing (and being able to measure) these CPU hotspots myself. I take it you have many torrents, each with many trackers, most of them not being reachable. Is that right?

glassez:
> A good appraoch is to make smaller independent PRs, to allow fixes to land independently

👍

bolshoytoster:
> I take it you have many torrents

116

> each with many trackers

115 trackers

> most of them not being reachable

64 of those trackers give `Host not found (authoritative)`

ValeZAA:
> After ignoring `lt::errors::announce_skipped`, the biggest hotspots are [the calls](https://github.com/arvidn/libtorrent/blob/0afd8a7a480902273dc6dd75744f36644988a376/src/pe_crypto.cpp#L103C24-L103C28) [to `mp::powm`](https://github.com/arvidn/libtorrent/blob/0afd8a7a480902273dc6dd75744f36644988a376/src/pe_crypto.cpp#L118) in the Diffie Hellman exchanges. This may be more of a general bittorrent issue, but this seems quite expensive. Couldn't bittorrent's encryption be replaced by regular SSL and an encrypted client hello?

Yes!! ECH IS GREAT idea and an open bug https://github.com/arvidn/libtorrent/issues/7529 

The hardest part is implementing HTTPS RR without the use of DOH. Yet, it works in both firefox and Chrome just fine (not on Windows 10, there is a bug in DNSQuaryA function, only Windows 11). That being said: Openssl 3.0.0 and later now have a license that is allowed to be run on Windows. So would be nice. Do not bother with schannel API of windows. It is too buggy.

bolshoytoster:
Most of the small optimizations from here have made it into libtorrent in #7916 and b8289d2, so it's just `is_i2p_url`.

@arvidn, do you know if it would be possible to add an extra 1-bit-wide bool to `announce_entry` ([after `verified`](https://github.com/arvidn/libtorrent/blob/c48cdadcdb453b89c1f97bf7dd7485ec8bd376e9/include/libtorrent/aux_/announce_entry.hpp#L192)) without affecting the ABI?

PR #7912, opened 2025-03-17
[WIP] Add `root_certificate` to `add_torrent_params`.

### LINKED ISSUE #2192: SSL torrents + ut_metadata
If already have a valid keypair, could I connect to an SSL swarm and download the metadata?

Looking at the code, it appears that an ssl cert can't be set until `m_ssl_ctx` has been initialized; which currently requires valid metadata.

Is there any reason we couldn't initialize the ssl context earlier, maybe add `ssl_cert` to `add_torrent_params`?

jameskeane:
This is so that a peer can join an ssl swarm with only a signed cert and infohash.
Fixes #2192

@arvidn I couldn't get the simulator building with openssl, so I have no real way of testing this right now. Any guidance on how to solve that?


arvidn:
unfortunately the simulator doesn't support the SSL wrappers in boost.asio (yet).

arvidn:
There is a test for the SSL feature already, which seem the most natural one to extend (`test/test_ssl.cpp`).

What happens if this field is set on a magnet link but the torrent turns out not to be an SSL torrent? (it should ideally be documented in the comment).

Should the root certificate also be allowed to be passed on the magnet link URI as well?

include/libtorrent/add_torrent_params.hpp
```diff
@@ -370,6 +370,11 @@ TORRENT_VERSION_NAMESPACE_3
 		std::time_t last_download = 0;
 		std::time_t last_upload = 0;
 
+		// if this is an SSL torrent, specify the root certificate here.
+		// This is so that an ssl torrent can be added with only an infohash
+		// and certificate.
+		std::string root_certificate;
```
arvidn, line 376: unfortunately `add_torrent_params` has some restrictions to maintain ABI compatibility. If you move this field to the end of the class, I believe it's OK

src/torrent.cpp
```diff
@@ -577,6 +577,14 @@ bool is_downloading_state(int const st)
 				do_connect_boost();
 			}
 
+			if (!is_ssl_torrent() && !p.root_certificate.empty())
```
arvidn, line 580: it's not obvious that the `is_ssl_torrent()` check is necessary or correct. Do you not want it to be possible to override a root certificate if there already exists one in the .torrent file?

jameskeane: Yes that was my intention, but you're right it seems unnecessary.

simulation/test_metadata_extension_with_ssl.cpp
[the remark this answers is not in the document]
jameskeane: I agree, I am working on incorporating one into `test_ssl` as per your suggestion. Since the simulator is not compatible with ssl torrents, this wouldn't work anyways.

jameskeane:
Added a rudimentary test for success using metadata plugin in `test_ssl.cpp`.

jameskeane:
>What happens if this field is set on a magnet link but the torrent turns out not to be an SSL torrent?

I could see it going both ways:
 1. If the user specifies a root CA; their intent is to use an encrypted channel so it should error.
 2. The user intends to download a torrent at the specified infohash, download it even if the certificate is not necessary.

What is preferable to you?

jameskeane:
> Should the root certificate also be allowed to be passed on the magnet link URI as well?

Possibly, but knowing that with a max length of 2048 characters for a URL, embedding in the magnet will work for some certs but not all.

jameskeane:
> Should the root certificate also be allowed to be passed on the magnet link URI as well?

I don't see why not, the vast majority of root certs would fit in the 2048 character limit of a URL.

arvidn:
I think the safest approach is to require it to be an SSL torrent if the certificate is specified. i.e. failing closed.

arvidn:
I think this change warrants a note in `Changelog` too

test/test_ssl.cpp
```diff
@@ -611,13 +612,125 @@ void test_malicious_peer()
...
+	torrent_handle tor2 = ses2.add_torrent(addp2, ec);
+
+	tor2.set_ssl_certificate(
+		combine_path("..", combine_path("ssl", "peer_certificate.pem"))
+		, combine_path("..", combine_path("ssl", "peer_private_key.pem"))
+		, combine_path("..", combine_path("ssl", "dhparams.pem"))
+		, "test");
+
+	ses1.listen_port();
+	ses2.listen_port();
```
arvidn, line 701: these calls appear meaningless. they return the listen port, right? and the return value is discarded

jameskeane: Was copied from the [original test case](https://github.com/arvidn/libtorrent/blob/09754d7b627378cb1a215a2d964b4876ef20c35c/test/test_ssl.cpp#L236), I assume it's solving for some race condition; so I just cargo-culted it.

jameskeane: I will add a comment

jameskeane:
Added a testcase for when the seeding torrent does not have a certificate. It was already correct behaviour, as the handshake failed during peer connection.

Looks like support for magnet links will require:
 1. Storing the root certificate on `torrent` or exposing a method on `ssl_context` to return the root certificate in encoded form (PEM?).
 2. Modifying `torrent_handle` and `torrent` to expose that.
 3. Adding support to `make_magnet_uri`, which will need to check the URL length and fail if the certificate would exceed the max length. Also error if `is_ssl_torrent` and no root certificate is set.
 4. Adding support to `parse_magnet_uri`.
 5. Associated tests.

I'm happy to take a stab at that, but maybe in a separate PR? Let me know your thoughts.

arvidn (approved):

jameskeane:
Had to remove check on the `peer_disconnected_alert` as the handshake failed message is different under gnussl and windows.

arvidn:
Extending `add_torrent_params` affects ABI. From the library's point of view, there's no way of telling whether the new or the old version of the struct was passed. This change is safe to make in `master`

jameskeane:
Updated the target branch to master.

jameskeane:
@arvidn Should be good to go, fixed failing build and tests are passing locally.

arvidn:
sorry for dropping the ball on this

jameskeane:
All good, thanks!

PR #7913, opened 2025-03-20
fix: UDP tracker: accept responses from unexpected IPs

baka-gourd:
## Related issues

- https://github.com/qbittorrent/qBittorrent/issues/20882
- https://github.com/arvidn/libtorrent/issues/7679

## Related BEP

[UDP Tracker Protocol for BitTorrent](https://www.bittorrent.org/beps/bep_0015.html)

## Why are these issues closed?

In one of the issues the author of the issue just provided the question but didn't follow the contributing rule so it was closed, the other one I don't know. But the issue still exists.

## Changes

one line. do not return when ip not match.

I can give a short description of the problem. the BEP0015 only asks for verification of the txID, not the server IP consistency. So, limiting ip consistency is redundant and causes quite a few problems.

## Whats problem

TBA, which has only been massively reported in recent years. Because in the ipv4 era, a tracker generally did not have multiple ip addresses, this validation did not cause serious problems, however, in the ipv6 era, a tracker can have many ipv6 addresses, and which ip it uses for a response is not strictly limited in BEP, so there are many scenarios where it is not the same ip that is used for a response.

Also, since some proxies (ie Fake IP) may change the address to which UDP packets are sent, these packets will also be dropped, making the udp tracker completely inoperable after the proxy.

So I submitted this PR


arvidn (commented):
the `#ifndef TORRENT_DISABLE_LOGGING` should be extended to cover the outer if-statement, and probably merged with the inner if-block.

The comment should be extended with a justification to allow the response from a different IP.

Ideally there would be a test as well, a simulation would probably be the simplest way to do that.

baka-gourd:
> the `#ifndef TORRENT_DISABLE_LOGGING` should be extended to cover the outer if-statement, and probably merged with the inner if-block.

done

> The comment should be extended with a justification to allow the response from a different IP.

done

> Ideally there would be a test as well, a simulation would probably be the simplest way to do that.

I tried, but it's hard to simulate a tracker with two ip's in a test environment, I failed and my test samples never returned, sorry.

arvidn (dismissed):

src/udp_tracker_connection.cpp
```diff
@@ -338,23 +338,23 @@ namespace libtorrent {
 			return false;
 		}
 
-		// ignore packet not sent from the tracker
-		// if m_target is inaddr_any, it suggests that we
-		// sent the packet through a proxy only knowing
-		// the hostname, in which case this packet might be for us
-		if (!m_target.address().is_unspecified() && m_target != ep)
-		{
 #ifndef TORRENT_DISABLE_LOGGING
```
arvidn, line 341: you need `TORRENT_UNUSED(ep);` to indicate that we may not use this variable

arvidn:
this test is still failing `test_tracker.cpp.udp_tracker_v4`

arvidn:
interestingly, it works when I run the test locally

baka-gourd:
> interestingly, it works when I run the test locally

Yes, I suspect it's an issue with the github network stack, since macos runners seem to share the network

arvidn:
I don't think it's a coincidence that the udp tracker tests are failing, seemingly consistently

baka-gourd:
> I don't think it's a coincidence that the udp tracker tests are failing, seemingly consistently

at the moment I don't have a mac to test and develop on (I've only been able to test with other people's macs before, which obviously can't be occupied for long periods of time), so if this isn't a coincidence I probably don't have a way to fix it. Sorry

baka-gourd:
However, when a tracker has multiple different addresses, UDP packets are not necessarily sent from the address they were received at. I acknowledge the security concerns, so this might require a switch. After all, multi-address servers are quite common.

1265578519:
哇！突然看到有人推进解决这个问题了，我在来简短的说一下吧

```
qbittorrent、libtorrent和其它一些开源的BT客户端不遵守bep协议规范，请不要在tracker域名dns上解析多个ip地址
https://github.com/qbittorrent/qBittorrent/issues/20882

bep15协议地址
https://www.bittorrent.org/beps/bep_0015.html

这些客户端都没有遵守协议去判断transaction ID，如果客户端请求域名IP地址A，但是tracker通过IP地址B发生回应，那么就会导致客户端请求失败，报错skipping tracker announce （unreachable）
ipv4和ipv6都会发生这个现象，3年前和客户端作者反馈过，但是他们至今没有解决这个问题

我们作为tracker提供管理者，为了避免这个现象，请使用不同的二级域名来分配更多服务器ip，例如tracker1、tracker2、tracker3、tracker4，确保每个域名只负责一个ip地址，而且不要把多个ip地址绑定到一台相同的服务器中（极其重要）


```

PR #7915, opened 2025-03-23
Adds support for v2 infohash to `load_torrent`.

### LINKED ISSUE #7914: `find_torrent` does not support sha256 (v2) infohashes
The `find_torrent` method on Session currently only accepts a sha1 v1 infohash.

jameskeane:
Tried to keep it backwards compatible, so not the prettiest. Compiler couldn't deduce the overload between handle->impl, hence the new `find_torrent_handle_v2` method to be explicit.

Fixes #7914.

### DIFF
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -714,6 +714,7 @@ namespace aux {
 			void abort_stage2() noexcept;
 
 			torrent_handle find_torrent_handle(sha1_hash const& info_hash);
+			torrent_handle find_torrent_handle_v2(sha256_hash const& info_hash);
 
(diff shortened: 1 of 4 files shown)


arvidn:
I recognize that there's no unit test for `find_torrent()`, but it would be good to have one for this function to ensure it keeps working

arvidn:
the python binding will need to support the overloaded function as well

arvidn:
the title says "load_torrent", but the patch is about "find_torrent".

I made a variant of this here: https://github.com/arvidn/libtorrent/pull/7921

jameskeane:
> the title says "load_torrent", but the patch is about "find_torrent".
My mistake.

Feel free to close in favor of your variant! Thank you!

PR #7918, opened 2025-03-28
Add support for using `io_uring` backend in boost asio

Chocobo1:
Requires liburing library and a modern Linux kernel.
https://github.com/axboe/liburing

Tested with Linux 6.13.8, boost 1.87 and liburing 2.9.

Example with cmake:
cmake -Dio-uring=ON ./


Chocobo1:
Disclaimer: I have only ran some naive testing and it seems to be fine. Didn't benchmark it nor tested its stability. Since this new option is OFF by default I suppose it can be left to those willing to try.

@HanabishiRecca 
FYI, you might be interested in this one.

HanabishiRecca:
Hmm, interesting. I don't think it would actually change anything though, beside slightly reducing CPU load. But I'll try it anyway.

Chocobo1:
>beside slightly reducing CPU load

I would consider it a clear win if that is the case. Hopefully its overhead would be comparable to epoll/others.

HanabishiRecca:
FYI, it makes qBittorrent process to spin 100% of iowait on one CPU core.
But according to axboe/liburing#943 that appears to be normal behavior for apps constantly waiting for network.

HanabishiRecca:
One interesting consequence I observe, using our simple pread mode with multiple I/O threads there are no request merges anymore (0 rrqm). That means `io_uring` backend actually produces better disk access pattern.
Of course OS I/O schedulers are made to compensate for contention, thus request merges, but not having them in the first place is better.

xavier2k6:
It may be worth making a note of `minimum io_uring` support in the docs as each branch has different `min` requirements for building boost.
* [RC_2_0 = You'll need at least version 1.67 of the boost library in order to build libtorrent.](https://github.com/arvidn/libtorrent/blob/RC_2_0/docs/building.rst#step-1-download-boost)
* [RC_1_2 = You'll need at least version 1.58 of the boost library in order to build libtorrent.](https://github.com/arvidn/libtorrent/blob/RC_1_2/docs/building.rst#step-1-download-boost)
____

`BOOST_ASIO_HAS_IO_URING` was [added in Boost 1.78.0](https://www.boost.org/doc/libs/1_78_0/doc/html/boost_asio/history.html)

**Linux Kernel 5.1**
* [The io_uring kernel interface was adopted in Linux kernel version 5.1 to resolve the deficiencies of Linux AIO](https://en.wikipedia.org/wiki/Io_uring)
* [Prominent features](https://en.wikipedia.org/wiki/Linux_kernel_version_history)

HanabishiRecca:
> The backend is disabled by default, and must be enabled by defining both `BOOST_ASIO_HAS_IO_URING` and `BOOST_ASIO_DISABLE_EPOLL`.

The patch doesn't define the latter btw.

Chocobo1:
>It may be worth making a note of minimum io_uring support in the docs as each branch has different min requirements for building boost.

Good idea. I added a little bit of docs.

> > The backend is disabled by default, and must be enabled by defining both `BOOST_ASIO_HAS_IO_URING` and `BOOST_ASIO_DISABLE_EPOLL`.
> 
> The patch doesn't define the latter btw. But the output seems to be identical though.

I'm not sure if it really needs to disable epoll.
As for `BOOST_ASIO_HAS_IO_URING_AS_DEFAULT`, it came from here: https://www.boost.org/doc/libs/1_87_0/doc/html/boost_asio/using.html

HanabishiRecca:
> I'm not sure if it really needs to disable epoll.

Well, epoll and io_uring are kinda mutually exclusive, afaik. So I don't see reason not to do that explicitly.

Chocobo1:
>Well, epoll and io_uring are kinda mutually exclusive, afaik. So I don't see reason not to do that explicitly.

Alright, added. Lets be cautious about it as the defines looks like a maze in boost asio...

include/libtorrent/aux_/allocating_handler.hpp
```diff
@@ -119,17 +119,30 @@ namespace libtorrent { namespace aux {
 	constexpr std::size_t openssl_write_cost = 0;
 #endif
 
+#ifdef BOOST_ASIO_HAS_IO_URING
```
arvidn, line 122: I'm a bit hesitant of this. Does this macro just mean boost.asio *supports* iouring as a backend, or does it mean that it *is* using it?
I would expect (over time) everyone's asio will support it (on linux), but whether you want to (or benefit from) using it doesn't necessarily follow. However, this will make you allocate almost 1 kB extra for every write peer unconditionally, at compile time.

To me, it seems more reasonable to check `BOOST_ASIO_HAS_IO_URING_AS_DEFAULT`, and maybe `BOOST_ASIO_DISABLE_EPOLL` being defined. This would be reverse-engineering the logic used by boost.asio to pick a back-end. Disabling epoll could also make it chose select, or poll. If this is intended to track how boost asio selects back-end, I think it would be good to have a link to the asio source where this logic is defined

CMakeLists.txt
```diff
@@ -830,6 +832,28 @@ else()
...
+		set(_pkg_check_modules_calls
+			"find_dependency(PkgConfig)"
+			"pkg_check_modules(liburing REQUIRED QUIET IMPORTED_TARGET GLOBAL liburing)")
+		string(REPLACE ";" "\n" _pkg_check_modules_calls "${_pkg_check_modules_calls}")
+	endif()
+
+	target_compile_definitions(torrent-rasterbar PUBLIC
+		BOOST_ASIO_DISABLE_EPOLL
+		BOOST_ASIO_HAS_IO_URING
+		BOOST_ASIO_HAS_IO_URING_AS_DEFAULT)
```
arvidn, line 854: from the documentation "In general, BOOST_ASIO_HAS_ macros should not be explicitly defined by the user,"

https://www.boost.org/doc/libs/1_87_0/doc/html/boost_asio/using.html

CMakeLists.txt
```diff
@@ -830,6 +832,28 @@ else()
...
+		# propagate the dependency properly to help downstream project find liburing
+		set(_pkg_check_modules_calls
+			"find_dependency(PkgConfig)"
+			"pkg_check_modules(liburing REQUIRED QUIET IMPORTED_TARGET GLOBAL liburing)")
+		string(REPLACE ";" "\n" _pkg_check_modules_calls "${_pkg_check_modules_calls}")
+	endif()
+
+	target_compile_definitions(torrent-rasterbar PUBLIC
+		BOOST_ASIO_DISABLE_EPOLL
+		BOOST_ASIO_HAS_IO_URING
```
arvidn, line 853: this macro doesn't seem to be documented

include/libtorrent/aux_/allocating_handler.hpp
arvidn: I see, this is a *libtorrent* option. It should be named with the `TORRENT_` prefix. but it's still not obvious that just disabling epoll will make asio switch to io uring

HanabishiRecca: It was [linked](https://www.boost.org/doc/libs/1_78_0/doc/html/boost_asio/history.html) above, the official docs state

> The backend is disabled by default, and must be enabled by defining both `BOOST_ASIO_HAS_IO_URING` and `BOOST_ASIO_DISABLE_EPOLL`.

CMakeLists.txt
HanabishiRecca: Yeah, considering the above, the documentation simply contradicts itself.

include/libtorrent/aux_/allocating_handler.hpp
Chocobo1: >To me, it seems more reasonable to check BOOST_ASIO_HAS_IO_URING_AS_DEFAULT

Indeed. Changed.

>This would be reverse-engineering the logic used by boost.asio to pick a back-end. Disabling epoll could also make it chose select, or poll. 

Right and that is why I also defined `BOOST_ASIO_HAS_IO_URING_AS_DEFAULT` to force using io_uring.
In https://www.boost.org/doc/libs/1_87_0/doc/html/boost_asio/using.html it mentioned:
`Linux: io_uring is used instead of epoll.`

Also I don't think we need to replicate the logic. We let the user use `-Dio_uring` option which will define all the proper macros instead of detecting all the possible combinations.

>I see, this is a libtorrent option. It should be named with the TORRENT_ prefix. but it's still not obvious that just disabling epoll will make asio switch to io uring

I used to have that option but later I realized it is always synonymous to `BOOST_ASIO_HAS_IO_URING_AS_DEFAULT` so I think there is little value of having a `TORRENT_IO_URING` macro.

AFAIK here is how asio decide whether to really use io_uring:
https://github.com/boostorg/asio/blob/032545fc569553ff9a1f142dd9cec1a9e070a4f9/include/boost/asio/detail/config.hpp#L840-L844

CMakeLists.txt
Chocobo1: https://www.boost.org/doc/libs/1_78_0/doc/html/boost_asio/history.html

include/libtorrent/aux_/allocating_handler.hpp
Chocobo1: So basically `Dio_uring=ON` would add 3 additional macros:
* BOOST_ASIO_DISABLE_EPOLL
  Required as stated in https://www.boost.org/doc/libs/1_78_0/doc/html/boost_asio/history.html 
* BOOST_ASIO_HAS_IO_URING
  Required as stated in https://www.boost.org/doc/libs/1_78_0/doc/html/boost_asio/history.html 
* BOOST_ASIO_HAS_IO_URING_AS_DEFAULT
  This is a safety measure to force io_uring will be used. Even if libtorrent didn't define it, asio will still define it as seen in their source code.
  https://github.com/boostorg/asio/blob/032545fc569553ff9a1f142dd9cec1a9e070a4f9/include/boost/asio/detail/config.hpp#L840-L844

xavier2k6:
>Added a compile-time check for a minimum Linux kernel version of 5.10, when `io_uring` support is enabled.

This was added in Boost 1.83.0 (see `asio` full history below)
https://www.boost.org/doc/libs/1_87_0/doc/html/boost_asio/history.html

Chocobo1:
> > Added a compile-time check for a minimum Linux kernel version of 5.10, when `io_uring` support is enabled.
> 
> This was added in Boost 1.83.0 (see `asio` full history below) https://www.boost.org/doc/libs/1_87_0/doc/html/boost_asio/history.html

It sounded like previous versions of boost could compile with kernel < 5.10. I would prefer to let boost do the check by their own.

HanabishiRecca:
Ok, I have been using io_uring mode on my server since and don't see any issues.

HanabishiRecca:
No, as I said above:

> I don't think it would actually change anything though, beside slightly reducing CPU load.

A torrent client is not something that typically bottlenecked by CPU.

Chocobo1:
>Well then it's actually a net negative because Arvid mentioned this adds memory load of 1 KB for every write peer in your client.

When [zero copy](https://github.com/chriskohlhoff/asio/pull/1560) lands in boost asio then this shouldn't be an issue. Depending on how many data is pushed at once (and other factors), it might bring net positive results.

HanabishiRecca:
Also the recent kernel [tweak](https://www.phoronix.com/news/Linux-6.15-IO_uring) could potentially decrease amount of memory used by the network stack.

HanabishiRecca:
It's optional, disabled by default and takes no effort to add. I see no reason not to have it.

USBhost:
I have been running this for awhile now and it's working fine.

web0316:
So the file io was abandoned after all.

PR #7927, opened 2025-04-12
Utp resend optimization

bolshoytoster:
There has been a TODO message in `utp_socket_impl::send_pkt` for over 10 years now saying that the current implementation isn't very efficient, and it isn't.

This PR adds a new `std::vector` `m_needs_resend` for each utp connection to track the packets that need to be resent. I've kept the behaviour the same as the previous implementation - the packets at `m_acked_seq_nr` (last acked packet) and `m_seq_nr` (FIN packet) will never be added to `m_needs_resend`.

### DIFF
--- a/src/utp_stream.cpp
+++ b/src/utp_stream.cpp
@@ -206,6 +206,26 @@ void utp_socket_impl::release_packet(packet_ptr p)
 	m_sm.release_packet(std::move(p));
 }
 
+void utp_socket_impl::insert_packet(packet_ptr p)
+{
+	// If we're stalled we'll need to resend
+	if (m_stalled)
+	{
+		p->need_resend = true;
+		m_needs_resend.push_back(p.get());
+	}
+
+	packet_ptr old = m_outbuf.insert(m_seq_nr, std::move(p));
+	if (old)
+	{
+//		TORRENT_ASSERT(reinterpret_cast<utp_header*>(old->buf)->seq_nr == m_seq_nr);
+		if (old->need_resend)
+			m_needs_resend.erase(std::find(m_needs_resend.begin(), m_needs_resend.end(), &*old));
+		else m_bytes_in_flight -= old->size - old->header_size;
+		release_packet(std::move(old));
+	}
+}
+
 void utp_socket_impl::abort()
 {
 	m_error = boost::asio::error::connection_aborted;
@@ -1024,6 +1044,7 @@ void utp_socket_impl::send_syn()
 			m_sm.subscribe_writable(this);
 		}
 		p->need_resend = true;
+		m_needs_resend.push_back(p.get());
 	}
 	else if (ec)
 	{
@@ -1464,13 +1485,8 @@ bool utp_socket_impl::send_pkt(int const flags)
 
 	// first see if we need to resend any packets
 
-	// TODO: this loop is not very efficient. It could be fixed by having
-	// a separate list of sequence numbers that need resending
-	for (int i = (m_acked_seq_nr + 1) & ACK_MASK; i != m_seq_nr; i = (i + 1) & ACK_MASK)
+	for (packet *p: m_needs_resend)
 	{
-		packet* p = m_outbuf.at(aux::numeric_cast<packet_buffer::index_type>(i));
-		if (!p) continue;
-		if (!p->need_resend) continue;
 		if (!resend_packet(p))
 		{
 			// we couldn't resend the packet. It probably doesn't
@@ -1485,7 +1501,7 @@ bool utp_socket_impl::send_pkt(int const flags)
 		}
 
 		// don't fast-resend this packet
-		if (m_fast_resend_seq_nr == i)
+		if (m_fast_resend_seq_nr == reinterpret_cast<utp_header*>(p->buf)->seq_nr)
 			m_fast_resend_seq_nr = (m_fast_resend_seq_nr + 1) & ACK_MASK;
 	}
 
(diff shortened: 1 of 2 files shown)


bolshoytoster:
Something I noticed while doing this, is that in `utp_socket_impl::tick`, on timeout, libtorrent [marks every non-acked packet with `need_resend`](https://github.com/arvidn/libtorrent/blob/21cbbf74ee4cdf41c5877fddb84bde128d217914/src/utp_stream.cpp#L3466-L3477). The weird thing is that this includes the packet at `m_acked_seq_nr`, which, as I understand it, has already been acked (the previous resend loop started from `(m_acked_seq_nr + 1) & ACK_MASK`). Is this a mistake, or am I misunderstanding something?

The code has been there since utp support was merged (15 years ago), so it doesn't look like it causes any problems, it was just surprising.

bolshoytoster:
I've marked this as a draft, because I've just had a segfault in qbittorrent. I'm not sure if it's because of these changes, but it probably is.

bolshoytoster:
I caught the segfault, it was an issue with the FIN packet being marked as `need_resend` even though it's never added to `m_needs_resend`.

ValeZAA:
>The code has been there since utp support was merged (15 years ago), so it doesn't look like it causes any problems, it was just surprising.

... Erm, do you know that this bug is still open? A bot that closes the issues automatically. Sigh:  https://github.com/arvidn/libtorrent/issues/3542

src/utp_stream.cpp
```diff
@@ -2055,21 +2055,35 @@ bool utp_socket_impl::resend_packet(packet* p, bool fast_resend)
...
-	}
 	else
 	{
-		m_sm.inc_stats_counter(counters::utp_packets_out);
-		++m_out_packets;
-		++p->num_transmissions;
+		if (p->need_resend)
+		{
+			p->need_resend = false;
+			auto entry = std::find(m_needs_resend.begin(), m_needs_resend.end(), p);
```
arvidn, line 2070: this looks like it could be a problem under edge-case conditions. But probably not worse that the current implementation

src/utp_stream.cpp
```diff
@@ -206,6 +206,30 @@ void utp_socket_impl::release_packet(packet_ptr p)
...
+	if (m_stalled)
+	{
+		p->need_resend = true;
+		m_needs_resend.push_back(p.get());
+	}
+
+	packet_ptr old = m_outbuf.insert(m_seq_nr, std::move(p));
+	if (old)
+	{
+//		TORRENT_ASSERT(reinterpret_cast<utp_header*>(old->buf)->seq_nr == m_seq_nr);
```
arvidn, line 221: did you try un-commenting this? It looks like a reasonable thing to assert.

arvidn (commented):
looks good. Definitely an improvement. I think it would be even more efficient to use an intrusive linked list. Then both insert and removal is O(1). But that's a bigger change

src/utp_stream.cpp
bolshoytoster: Uncommented in f4fd64d, I just assumed there was a reason it was commented [9 years ago](https://github.com/arvidn/libtorrent/commit/103d574cb0d4cef4468586628bbfde42d3b6e66d#diff-a9b0002dfebb81b315f1d19d346361c385368345d731117eefad87b4b9817ca7).

src/utp_stream.cpp
bolshoytoster: Yeah, it could get bad if there are a lot of packets in flight, although I've only observed at most ~70 packets in the buffer at once due to the uTP max window size.

This could probably also be improved by using an intrusive linked list.

src/utp_stream.cpp
arvidn: I was just curious to know if you had tested it at all.

arvidn (approved):

src/utp_stream.cpp
```diff
@@ -1464,32 +1489,25 @@ bool utp_socket_impl::send_pkt(int const flags)
 
 	// first see if we need to resend any packets
 
-	if (!m_outbuf.empty())
+	while (!m_needs_resend.empty())
 	{
-		// TODO: this loop is not very efficient. It could be fixed by having
-		// a separate list of sequence numbers that need resending
-		for (int i = (m_acked_seq_nr + 1) & ACK_MASK; i != m_seq_nr; i = (i + 1) & ACK_MASK)
+		packet *p = m_needs_resend[0];
```
arvidn, line 1494:
```suggestion
		packet *p = m_needs_resend.front();
```

bolshoytoster:
I think I've fixed the tests. The issue was that I was temporarily invalidating a packet (adding its size to `m_bytes_in_flight` without unsetting `p->need_resend`), which caused some invariant checks to fail.

src/utp_stream.cpp
```diff
@@ -2004,6 +2005,7 @@ bool utp_socket_impl::resend_packet(packet* p, bool fast_resend)
 #if TORRENT_USE_ASSERTS
 	if (fast_resend) ++p->num_fast_resend;
 #endif
+	bool need_resend = p->need_resend;
```
arvidn, line 2008:
```suggestion
	bool const need_resend = p->need_resend;
```

arvidn (commented):
minor nit pick

arvidn (approved):

arvidn:
I don't understand why github is saying "This branch cannot be rebased due to conflicts"

arvidn:
maybe because of the merge commit in the middle

arvidn:
something is going on with github. I don't know why it's preventing the merge of this PR, or my duplicated (and rebased) version.

arvidn:
@ValeZAA 
I rebased all these commits and removed the merge commit here: https://github.com/arvidn/libtorrent/pull/7931
it has the same problem still. Is it important to just turn it into a single commit?

ValeZAA:
>has the same problem still. Is it important to just turn it into a single commit?

Yes, that is what I was saying.

arvidn:
landed my rebased and squashed version

arvidn:
thanks @bolshoytoster !

ValeZAA:
>The weird thing is that this includes the packet at m_acked_seq_nr, which, as I understand it, has already been acked (the previous resend loop started from (m_acked_seq_nr + 1) & ACK_MASK). Is this a mistake, or am I misunderstanding something?

What happened to this one?

PR #7937, opened 2025-05-01
Don't leak file descriptors to child processes

Chocobo1:
Without `O_CLOEXEC`, `SOCK_CLOEXEC` and `MADV_DONTFORK`, the file descriptors are available to the child process which would lead to leaking file descriptors and other unwanted behaviors.


arvidn (approved):

HanabishiRecca:
@Chocobo1, could this be related to qbittorrent/qBittorrent#22592?

Chocobo1:
>@Chocobo1, could this be related to https://github.com/qbittorrent/qBittorrent/issues/22592?

More or less. But the fd from boost asio is still leaking if I'm not mistaken: https://github.com/chriskohlhoff/asio/issues/523

glassez:
> @Chocobo1, could this be related to [qbittorrent/qBittorrent#22592](https://github.com/qbittorrent/qBittorrent/issues/22592)?

That is related to Qt, not to libtorrent.

PR #7940, opened 2025-05-04
fix building the python wheel on Mac

arvidn:


arvidn:
@AllSeeingEyeTolledEweSew Do you have any idea why setuptools is trying to build an egg?
I'm increasingly thinking that building the C++ code via `setup.py` is a mistake. It's incredibly complex and poorly documented.

I think it would be simpler to just have a python program that configures python for boost build and makes the binary. Then have some thin wrapper to turn a native python module into a wheel (it's not so easy to find out how to do that) and then invoke this for every python version and platform to build a wheel for.

arvidn:
maybe it's green now, or it's flaky. it used to fail like this:
https://github.com/arvidn/libtorrent/actions/runs/14721771065/job/41509430023?pr=7912#step:13:284
```
NotImplementedError: Support for egg-based install has been removed.
```

xavier2k6:
I think the command `python setup.py build_ext` should be replaced with `python pip3 build_ext`, if you google `NotImplementedError: Support for egg-based install has been removed.` you'll see other repo's have had same/similar problem recently

xavier2k6:
python-setuptools 1:80.0.0-1 dropped support for eggs. Upstream bug report https://dev.gnupg.org/T6784 and similar issue for meson https://github.com/mesonbuild/meson/issues/14522.

arvidn:
> I think the command python setup.py build_ext should be replaced with python pip3 build_ext

It's just not very clear how one would pass in custom build paramters. e.g. `python setup.py build_ext --b2-args "asserts=on invariant-checks=full" bdist_wheel --user --prefix=`

DUTNO1:
> > I think the command python setup.py build_ext should be replaced with python pip3 build_ext
> 
> It's just not very clear how one would pass in custom build paramters. e.g. `python setup.py build_ext --b2-args "asserts=on invariant-checks=full" bdist_wheel --user --prefix=`

I use this to build wheels for libtorrent: `python -m build -n -w -C--build-option=build_ext -C--build-option=--b2-args="optimization=speed crypto=openssl cxxstd=17 variant=release threading=multi libtorrent-link=static boost-link=static fpic=on lto=on logging=off alert-msg=off"`

arvidn:
@DUTNO1 that's also failing on MacOS CI, but in a different way:

https://github.com/arvidn/libtorrent/actions/runs/15241712705/job/42862602802?pr=7948#step:14:19
```
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/pyproject_hooks/_impl.py", line 402, in _call_hook
    raise BackendUnavailable(
    ...<4 lines>...
    )
pyproject_hooks._impl.BackendUnavailable: Cannot import 'setuptools.build_meta'

ERROR Backend 'setuptools.build_meta:__legacy__' is not available.
```

Issue #7535, opened 2023-10-23
100% of CPU core is used when torrent with mix of I2P and non-I2P trackers is added


Vort:
This problem can be reproduced with I2P mode **disabled**.
Just run `client_test` with default parameters, like this: `client_test --alert_mask=all` and add such magnet link:
`magnet:?xt=urn:btih:70260ca39b2867a59f46d81db85eecb256b9365c&dn=I2P%20Update%202.1.0&tr=http%3A%2F%2Ftracker2.postman.i2p%2Fannounce.php&tr=udp%3A%2F%2Ftracker.opentrackr.org%3A1337%2Fannounce`
Then observe CPU usage.

I suspect hang may be realted to function `libtorrent::torrent::update_tracker_timer`.

_libtorrent version (or branch):_ 5ce4c516479085eda9327d35891b7732d6d5037b
_platform/architecture:_ Windows x64
_compiler and compiler version:_ clang version 15.0.7

ukoz:
confirm 1 core 100% 2nd 50%, not going past 'no metadata'
 209 win64 MSVS 17.7.3

luzpaz:
soft bump

absolutep:
Here is the fix - it solved for me atleast in both "mixed mode" as well as "I2P-only mode"

https://github.com/qbittorrent/qBittorrent/discussions/21926

Vort:
> Here is the fix - it solved for me atleast in both "mixed mode" as well as "I2P-only mode"

This is not a fix, it is avoidance of the problem.

Neustradamus:
@arvidn has done a recent PR about I2P:
- https://github.com/arvidn/libtorrent/pull/7831

Neustradamus:
.

absolutep:
unstale this

milahu:
fixed by #7950 

the old version was bottlenecking at `is_i2p_url`

the new version is bottlenecking at

```
libtorrent::aux::torrent::update_tracker_timer
libtorrent::aux::torrent::announce_with_tracker
boost::asio::detail::timer_queue
libtorrent::aux::session_settings::get_bool
```

these functions are also mentioned in #7898

i get 100% cpu load seeding 4K torrents
quickfix: use multiple network threads https://github.com/arvidn/libtorrent/issues/7894#issuecomment-2893483860

edit: this was with the non-default [libtorrent settings](https://www.libtorrent.org/reference-Settings.html)
`announce_to_all_tiers = true`
[...]

PR #7950, opened 2025-05-27
Some optimizations

bolshoytoster:
See #7898 for details.

### DIFF
--- a/src/kademlia/routing_table.cpp
+++ b/src/kademlia/routing_table.cpp
@@ -1100,7 +1100,7 @@ std::vector<node_entry> routing_table::find_node(node_id const& target
 		if (int(l.size()) > count)
 		{
 			// sort the nodes by how close they are to the target
-			std::sort(l.begin() + unsorted_start_idx, l.end()
+			std::nth_element(l.begin() + unsorted_start_idx, l.begin() + unsorted_start_idx + count, l.end()
 				, [&target](node_entry const& lhs, node_entry const& rhs)
 				{ return compare_ref(lhs.id, rhs.id, target); });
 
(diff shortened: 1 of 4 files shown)


src/kademlia/routing_table.cpp
```diff
@@ -1100,7 +1100,7 @@ std::vector<node_entry> routing_table::find_node(node_id const& target
 		if (int(l.size()) > count)
 		{
 			// sort the nodes by how close they are to the target
-			std::sort(l.begin() + unsorted_start_idx, l.end()
+			std::nth_element(l.begin() + unsorted_start_idx, l.begin() + count, l.end()
```
arvidn, line 1103: I think the comment should be changed to indicate that we don't need the nodes to be sorted, we just want the `count` closest nodes.

src/kademlia/routing_table.cpp
```diff
@@ -1139,7 +1139,7 @@ std::vector<node_entry> routing_table::find_node(node_id const& target
 		if (int(l.size()) > count)
 		{
 			// sort the nodes by how close they are to the target
```
arvidn: this comment is also outdated now

arvidn (commented):
looks good, except a minor comment

src/kademlia/routing_table.cpp
bolshoytoster: I've updated the comments.

arvidn (approved):

arvidn:
thanks!

Neustradamus:
Thanks @bolshoytoster!

Linked to:
- https://github.com/arvidn/libtorrent/pull/7898
- https://github.com/arvidn/libtorrent/issues/7894
- https://github.com/arvidn/libtorrent/issues/7535

PR #7951, opened 2025-06-03
Fix wrong variable in CMake

Chocobo1:
The `Windows` variable is not defined in CMake. The correct variable should be `MSVC`.
https://cmake.org/cmake/help/latest/manual/cmake-variables.7.html#variables-that-describe-the-system

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -657,7 +657,7 @@ if (APPLE)
 endif()
 
 # check if we need to link with libatomic (not needed on MSVC)
-if (NOT Windows)
+if (NOT MSVC)
 	# TODO: migrate to CheckSourceCompiles in CMake >= 3.19
 	include(CheckCXXSourceCompiles)
 


xavier2k6:
@Chocobo1 Do you know why the Windows build in this repo (in-general) always show a green tick overall even when there's been a failure of say even 1 test?

Chocobo1:
>@Chocobo1 Do you know why the Windows build in this repo (in-general) always show a green tick overall even when there's been a failure of say even 1 test?

Not familiar with libtorrent CI workflow but it seems to be from this line: https://github.com/arvidn/libtorrent/actions/runs/15422437022/workflow?pr=7951#L21

petterreinholdtsen:
It look like your fix broke the Windows CI run.  Perhaps more fixes are needed?

Chocobo1:
>It look like your fix broke the Windows CI run. Perhaps more fixes are needed?

How do I fix it?

glassez:
> It look like your fix broke the Windows CI run.

Doesn't seem that they are related.

arvidn (approved):

Issue #7840, opened 2025-01-31
PEP 621 compatible Python build?


mxlei01:
libtorrent's pyproject.toml is missing version and name [PEP-621](https://peps.python.org/pep-0621/) and this causes package managers such as [uv](https://github.com/astral-sh/uv) for versions > 0.5.19 to fail.

If anybody uses `uv` and has `boost-python` installed, they can have a project pyproject.toml setup such as:
```
[project]
name = "test"
version = "0.1.0"
description = ""
readme = "README.md"
requires-python = ">=3.12"
dependencies = [
    "libtorrent @ git+https://github.com/arvidn/libtorrent.git@v2.0.10#subdirectory=bindings/python"
]
```
Then the command `uv sync` will automatically build libtorrent based on this git url + version.

With how libtorrent `pyproject.toml` is setup it will cause the build to fail on `uv`.
```
Build fails: `pyproject.toml` is using the [project] table, but the required project.name field is not set
```

The solution is pretty simple adding these two lines in `pyproject.toml`:
```
name = "libtorrent"
version = "2.0.10.1"
```
Example via: https://github.com/mxlei01/libtorrent/blob/v2.0.10.1/pyproject.toml

Just raising a potential issue, but the workaround is pretty simple by forking and editing the file.

xavier2k6:
@mxlei01 I've added `name` field in #7955 & hopefully this should be enough?!

mxlei01:
@xavier2k6  I think you might need a version as well?

You should definitely test this via:

```
dependencies = [
    "libtorrent @ git+https://github.com/<YOUR FORK HERE>/libtorrent.git@<YOUR BRANCH ID>#subdirectory=bindings/python",
]
```

And then run `uv sync`.

But if you do need a version this makes it a not so simple change because for every branch you'll need to make sure `version` is updated. Either you might rely during build and try to capture the version number for the change or you'll need to remember to do it manually.

xavier2k6:
>I think you might need a version as well?

I've looked at other projects & `version` field wasn't included in all cases, so it seems that it's not a hard requirement unlike the `name` field.

Edit.:
Version is bumped prior to new release in `bindings/python/setup.cfg` under `metadata`

Anyway....it's a step in the right direction & this repo's maintainer can weigh-in/refactor as they see fit.

xavier2k6:
repo-review seem to indicate it's required.....(will ammend my PR)

[...]

PR #7955, opened 2025-06-04
PEP 621 Compliance - Add `name` & `version` fields to `pyproject.toml`

xavier2k6:
* https://peps.python.org/pep-0621/
* Xref.: https://github.com/arvidn/libtorrent/issues/7840


arvidn (approved):
I'll update the `set_version.py` script to keep this updated as well

PR #7956, opened 2025-06-07
Improve loading sparse merkle tree

glassez:
Avoid memory allocations and data copying while adjusting size of mask when loading sparse merkle tree.

### DIFF
--- a/src/merkle_tree.cpp
+++ b/src/merkle_tree.cpp
@@ -119,18 +119,19 @@ namespace {
 		, std::vector<bool> const& verified)
 	{
 		INVARIANT_CHECK;
-		TORRENT_ASSERT(mask.size() == size());
-		if (size() != mask.size()) return;
+        TORRENT_ASSERT(mask.size() <= size());
+        if (size() < mask.size()) return;
 
 		int const first_block = block_layer_start();
 		int const end_block = first_block + m_num_blocks;
 
-		TORRENT_ASSERT(first_block < int(mask.size()));
-		TORRENT_ASSERT(end_block <= int(mask.size()));
+		TORRENT_ASSERT(first_block < int(size()));
+		TORRENT_ASSERT(end_block <= int(size()));
 
 		// if the mask covers all blocks, go straight to block_layer
 		// mode, and validate
-		if (std::all_of(mask.begin() + first_block, mask.begin() + end_block, identity()))
+		if ((first_block < int(mask.size())) && (end_block <= int(mask.size()))
+			&& std::all_of(mask.begin() + first_block, mask.begin() + end_block, identity()))
 		{
 			// the index in t that points to first_block
 			auto const block_index = std::count_if(mask.begin(), mask.begin() + first_block, identity());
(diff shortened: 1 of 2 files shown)


src/merkle_tree.cpp
```diff
@@ -119,18 +119,23 @@ namespace {
 		, std::vector<bool> const& verified)
 	{
 		INVARIANT_CHECK;
-		TORRENT_ASSERT(mask.size() == size());
-		if (size() != mask.size()) return;
```
arvidn, line 123: it looks like the main change is to tolerate a small (or large) mask. Is this assumed to be saved from an older version or is there some issue with saving these too?

src/merkle_tree.cpp
```diff
@@ -119,18 +119,23 @@ namespace {
...
 
-		TORRENT_ASSERT(first_block < int(mask.size()));
-		TORRENT_ASSERT(end_block <= int(mask.size()));
+		TORRENT_ASSERT(first_block < int(size()));
+		TORRENT_ASSERT(end_block <= int(size()));
 
 		// if the mask covers all blocks, go straight to block_layer
 		// mode, and validate
-		if (std::all_of(mask.begin() + first_block, mask.begin() + end_block, identity()))
+		if ((first_block < mask_size) && (end_block <= mask_size)
```
arvidn, line 137:
```suggestion
		if (end_block <= mask_size
```
We're guaranteed that `first_block` < `end_block`, right?

src/merkle_tree.cpp
```diff
@@ -156,14 +161,14 @@ namespace {
...
-			TORRENT_ASSERT(end_piece <= int(mask.size()));
+			TORRENT_ASSERT(first_piece < int(size()));
+			TORRENT_ASSERT(end_piece <= int(size()));
 
 			// if the mask covers all pieces, and nothing below that layer, go
 			// straight to piece_layer mode and validate
-			if (std::all_of(mask.begin() + first_piece, mask.begin() + end_piece, identity())
-
-				&& std::all_of(mask.begin() + end_piece, mask.end(), std::logical_not<>()))
+			if ((first_piece < mask_size) && (end_piece <= mask_size)
```
arvidn, line 169:
```suggestion
			if (end_piece <= mask_size
```
I would think this is an OK simplification

src/merkle_tree.cpp
```diff
@@ -183,12 +188,12 @@ namespace {
 		}
 
 		// if the mask has only zeros, go straight to empty tree mode
-		if (t.empty() || std::none_of(mask.begin(), mask.end(), identity()))
+		if (t.empty() || std::none_of(mask.begin(), mask.begin() + mask_size, identity()))
```
arvidn, line 191: ~would it make this simpler by truncating `mask` at the beginning, in case it's too long?~
never mind, it's immutable

src/torrent.cpp
```diff
@@ -437,7 +437,6 @@ bool is_downloading_state(int const st)
 			std::vector<bool> const& verified_bitmask = (i >= verified.end_index()) ? empty_verified : verified[i];
 			if (i < mask.end_index() && !mask[i].empty())
 			{
-				mask[i].resize(m_merkle_trees[i].size(), false);
```
arvidn, line 440: I see. this is where we save an allocation

arvidn (approved):

src/merkle_tree.cpp
glassez: > it looks like the main change is to tolerate a small (or large) mask. Is this assumed to be saved from an older version or is there some issue with saving these too?

The previous code allowed masks of a size other than the size of the tree. I only improved it so as to avoid reallocating memory and copying data (which can be quite large) during its processing.
As for the very fact of having such masks, I proceeded from the assumption that smaller masks are acceptable by design (to save space, avoiding keeping tails containing consecutive false values). I hardly called getting a larger mask expected, but as I mentioned in the code comment, they do occur in reality. I do not know how they appear, I just assumed that the only possibility for this is a bug in the generation code. I have not done any detailed research on this incident. I haven't found a better way to handle such a case other than ignoring the extra mask data. Besides, the previous code behaved exactly the same way.

src/merkle_tree.cpp
glassez: > We're guaranteed that `first_block` < `end_block`, right?

At least we assert on this indirectly:
```c++
TORRENT_ASSERT(first_block < int(size()));
TORRENT_ASSERT(end_block <= int(size()));
```

src/torrent.cpp
glassez: > I see. this is where we save an allocation

Yes.

PR #7960, opened 2025-06-21
Magnet-URI Webseeding

milahu:
implement https://github.com/bittorrent/bittorrent.org/pull/171

status: early draft

i need help to get from "parse magnet link with xs parameter" to "send http request to fetch metadata file"
should the http server be treated like a bittorrent peer?

so far, i tried to base `exact_source_connection` on `web_connection_base` (which is based on `peer_connection`)
but it should be based on `http_connection` (right?)
but i dont see where i should create the `http_connection`

i need "dont require files_checked to connect to webseeds" to reach my code
`files_checked` makes no sense when we have no metadata

currently the parser fails at

```
[1027] xxxxxxxxxxxxxxx peer [ 127.0.0.1:80 client: Unknown [ .]....S....i.|.....] ] [127.0.0.1:80] <<< exact_source_connection::on_receive [ loop step: m_parser.status_code() == -1 ]
```

this looks like `exact_source_connection::m_parser` is not an http parser...?
because `recv_buffer` contains a valid http response

```
HTTP/1.1 200 OK
Content-Length: 800
Connection: keep-alive

d8:announce30:udp://...
```

`exact_source_connection` examples:

```
./examples/exact_source_client "magnet:?xt=urn:btih:xxxxxxxxxxxxxxx&xs=http://127.0.0.1/cas/btih/xxxxxxxxxxxxxxx.torrent"
./examples/exact_source_client "magnet:?xt=urn:btih:xxxxxxxxxxxxxxx&cas=http://127.0.0.1/cas/"
```

this requires an http server on 127.0.0.1
serving metadata and content files under http://127.0.0.1/cas/

```
http://127.0.0.1/cas/btih/xxxxxxxxxxxxxxx.torrent
http://127.0.0.1/cas/btih/xxxxxxxxxxxxxxx/Some.Dir/Some.File.txt
```

### DIFF
--- a/src/web_peer_connection.cpp
+++ b/src/web_peer_connection.cpp
@@ -230,6 +230,7 @@ void web_peer_connection::on_connected()
 			}
 		}
 		*/
+		/*
 		peer_log(peer_log_alert::info, "web_peer_connection.cpp 304", "have.get_bit(i) = ...");
 		for (auto const i : fs.file_range())
 		{
@@ -238,6 +239,7 @@ void web_peer_connection::on_connected()
 				peer_log(peer_log_alert::info, "web_peer_connection.cpp 304", "file %d: have.get_bit(%d) = %d", int(i), int(k), have.get_bit(k));
 			}
 		}
+		*/
 		t->set_seed(peer_info_struct(), false);
 		if (have.none_set())
 		{


milahu:
feel free to ignore
currently this has zero priority for me

the goal here is to bridge the gap between magnet links and HTTP servers

assuming that most code needed for this is already implemented in libtorrent
(`http_connection`...)

assuming that HTTP servers are easier to deploy than bittorrent seeders

assuming we find a method to announce webseeds to trackers
for example by assuming that TCP port 80 | 443 = webseed
and that all metadata and content files are served from `/cas/`

> there's a much simpler way to make an HTTP request, just use `http_connection`

like i said:
i dont see where i should create the `http_connection`

> a lot of this code seems suspicious. Was it generated by an LLM?

yep, some changes are "inspired" by chatGPT
this was 6 months ago, so i dont remember details

PR #7963, opened 2025-06-27
add surpport for custom tracker proxy

Tearon2020:
i have lots of pt website,because of isp,some tracker only could connect successful by using proxy server.
befor, i use two qbittorrent client for my pt. one use proxy, one connect without proxy.
so i want to need custom proxy rule.
if the libtorrent  surpport, i  would only use one qbittorrent.


Tearon2020:
the code has some wrong

PR #7965, opened 2025-06-27
add surpport:not all tracker connection use proxy server

Tearon2020:
i have lots of pt website,because of isp,some tracker only could connect successful by using proxy server.
befor, i use two qbittorrent client for my pt. one use proxy, one connect without proxy.
so i want to need custom proxy rule.
if the libtorrent surpport, i would only use one qbittorrent.

### DIFF
--- a/include/libtorrent/aux_/proxy_settings.hpp
+++ b/include/libtorrent/aux_/proxy_settings.hpp
@@ -63,6 +63,15 @@ namespace aux {
 
 		// if true, tracker connections are subject to the proxy settings
 		bool proxy_tracker_connections = true;
+
+		//defaults to false, it means that all tracker connections are subject if proxy_tracker_connections is true;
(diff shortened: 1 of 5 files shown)


arvidn:
I'm a bit uneasy about this. It's a pretty bug feature for a (presumably) rare use case.
There's a global setting to `proxy_tracker_connections`, maybe it would be more appropriate to have a per-tracker setting to override this. I don't think there's any precedence for that though.

Tearon2020:
> proxy_tracker_connections

i known this option,but it doesn't useful for me.
example: i have 6 pt websites， A/B/C/D/E can  connect tracker direct， but F can not， i only use another client connect it with proxy setting.
although ,i could use only one client for all pt website with proxy, but some pt website will judge it seedbox, because the proxy ip is vps.
so, i priority use direct first.

PR #7969, opened 2025-06-28
cibuildwheel: Bump action to v3.0.0

xavier2k6:
Xref.: https://github.com/pypa/cibuildwheel/releases


arvidn:
maybe this fixes the ubuntu cibuildwheel. https://github.com/arvidn/libtorrent/pull/7970

xavier2k6:
> maybe this fixes the ubuntu cibuildwheel. #7970

Maybe or #7824 will help too.

>Warning: cibuildwheel: Invalid skip selector: 'pp*'. This selector matches a group that wasn't enabled. Enable it using the `enable` option or remove this selector. This selector will have no effect.

Well, we can remove all remnants of this since it's opt-in & it's not being used.

xavier2k6:
```
  /opt/_internal/cpython-3.9.23/include/python3.9/pyconfig.h:1009:9: warning: "HAVE_SNPRINTF" redefined
   1009 | #define HAVE_SNPRINTF 1
        |         ^~~~~~~~~~~~~
  src/boost_python.hpp:11:9: note: this is the location of the previous definition
     11 | #define HAVE_SNPRINTF
        |         ^~~~~~~~~~~~~
```

What about this?

arvidn:
> What about this?

I believe that's probably to handle issues with an old version of python or boost.python

xavier2k6:
>#if OPENSSL_VERSION_NUMBER < 0x30000000L

I changed this as a test, since OpenSSL 1.1.1 is EOL/INSECURE & shouldn't be used.

Making this change.....caused an awful lot more failures......(some of the runner images still don't have OpenSSL 3.x, I believe & the AppVeyor image being used only has OpenSSL 1.x)

baseplate-admin:
Hi, @xavier2k6,

Do you think we can drop cibuildwheel and do something like using github actions matrix to build the dependencies?

I already have a [CI/CD action](https://github.com/baseplate-admin/libtorrent-python/blob/main/.github/workflows/CI.yaml) that builds `libtorrent`  from master.

xavier2k6:
@baseplate-admin this is not up to me, I am not the maintainer of this repo - the wheels are uploaded to pypi for other projects to use.

baseplate-admin:
Okay, sorry for pinging you (i thought you were managing `cibuildwheels` for `libtorrent`),

@arvidn, can we take the approach i have mentioned here: https://github.com/arvidn/libtorrent/pull/7969#issuecomment-3083421385

I think that would make life a whole lot easier

xavier2k6:
No problem, I only do my best to keep them ticking over as such....

arvidn:
> Okay, sorry for pinging you (i thought you were managing `cibuildwheels` for `libtorrent`),
> 
> @arvidn, can we take the approach i have mentioned here: [#7969 (comment)](https://github.com/arvidn/libtorrent/pull/7969#issuecomment-3083421385)
> 
> I think that would make life a whole lot easier

I think that sounds good. I'm somewhat ignorant of the python world myself. I don't know what the best tools to build wheels are. But I would think fewer moving parts is better.

xavier2k6:
@arvidn It may be no harm in getting the opinion of tribler/deluge team members if moving away from wheels being published to pypi as it may effect their CI Builds etc. too. (unless I'm misinterpreting the approach by @baseplate-admin )

CC @qstokkink @zakkarry

zakkarry:
> @arvidn It may be no harm in getting the opinion of tribler/deluge team members if moving away from wheels being published to pypi as it may effect their CI Builds etc. too. (unless I'm misinterpreting the approach by @baseplate-admin )
> 
> CC @qstokkink @zakkarry

Thanks for the ping. I should have some time soon to look at this and possibly test it locally soon. I've also messaged a few team members about this to double-check things.

baseplate-admin:
Hi @xavier2k6, 

> if moving away from wheels being published to pypi as it may effect their CI Builds etc. too

My approach is not to move away from pypi. I am proposing using github action matrix to build the package, the using another action to publish it to pypi. This is the default way of handling things like maturin builds. 

This way we can remove both tox and cibuildwheel and use matrix python to build it and test it

qstokkink:
My thoughts on this matter:

 - AFAIK `cibuildwheel` is still the best way to (cross-)compile for all supported `libtorrent` wheels. Dropping it to do manual compilation in a GHA matrix for all supported ISAs and compiler flavors seems, frankly, insane to me. At best, I think this would turn the `libtorrent` GHA (CI) into a Dockerized hellscape.
 - Dropping PyPI would make it highly inconvenient to use `libtorrent` in Python. The alternative would be that every downstream project compiles its own `libtorrent`, which would make error-reporting even more of a pain (I'm eyeballing the 197 issues on GitHub here). I say: make an effort to keep it.
 - Updating OpenSSL to an actually secure version for all supported wheels seems like a good idea.

With that in mind, my conclusion is short: "update the OpenSSL version and fix the resulting issues."

zakkarry:
> * atrix for all supported ISAs and compiler flavors seems, frankly, insane to me. At best, I think this would turn the `libtorrent` GHA (CI) into a Dockerized hellscape.
> 
> * Dropping PyPI would make it highly inconvenient to use `libtorrent` in Python. The alternative would be that every downstream project compiles its own `libtorrent`, which would make error-reporting even more of a pain (I'm eyeballing the 197 issues on GitHub here). I say: make an effort to keep it.
> 
> * Updating OpenSSL to an actually secure version for all supported wheels seems like a good idea.

1 & 2 agreed, dropping pypi would break things for us as well (deluge)
3. Also agreed, it would be nice

building wheels internal for our projects independently was not something that sounded fun for us, and I'm sure you agree @qstokkink - its not an easy thing to get going or maintain, and it would be nice to have stable and consistent building (1.2.x isn't even being built currently, or wasn't last release...and I recommend it for deluge for a variety of reasons... it would be nice to have them up to date as well as 2.x)

I understand that this is kind of a headache sometimes, but the maintainers of the pypi and build process already have most of it done and separating it up into each project that utilizes python builds would segment development and your note about errors/bugs being all over the place for each project instead of consolidated would be the primary issue but hardly the only one.

<hr>

if possible, I will say this again with hopes that its replied to and addressed somehow (@arvidn) like to see reintroduction of 1.2.x builds, as 1.2.20+ is not available on pypi currently.

I realize that kind of expands the scope from bumping gha, but I think if we're talking about python builds and their effect downstream in multiple projects we should handle everything that is currently outstanding if possible (since we're dealing with the CI/CD).

I've made multiple issues in the last year over 1.2.x builds and at one point earliest to my GHI making builds were not being made for python at all.

@arvidn is 1.2.x building when released through the CI/CD and the openssl issue @qstokkink mentioned something we can accomplish? tbh I kinda felt like you were avoiding responding to my comments and issues in regards to it previously (1.2.x python builds) - and an issue or two of mine was closed without actually addressing the this and when I tagged you and asked about it I felt basically ghosted.

If you have no intention of maintaining 1.2.x going forward and not producing builds, that really sucks and although I am currently on 1.2.20 and no builds are available there that is only because I am using a LSIO built libtorrent for alpine in docker, many many users rely on 1.2.x and prefer it still.

I hope you'll comment back and give us (deluge) an idea where you are at with this.

Thank you for your continued work, we do appreciate and value it and you.

cas--:
I can see that newer comments have echoed my opinion that the discussion of dropping of cibuildwheel is a ridiculous idea when most of the issues here are changes to OpenSSL version and only two manylinux wheel builds are actually failing.

The rest of the CI failures due to @xavier2k6 mistaken change to the libtorrent ssl minimum version check which has a different purpose than perceived and now forces openssl 3. This is not something a library should be doing unless requires newer version functionality or causes breakage, as demonstrated by the build error message:
 
> #error OpenSSL too old, use a recent version with SNI support

I can see that arvid instigated the move to OpenSSL 3 in CI build scripts but in my opinion we should not be rush switching to OpenSSL 3 without careful consideration since the downstream effects can be numerous. Sure 1.1.1 is official EOL but certainly for [Ubuntu they will still be releasing security patches](https://ubuntu.com/blog/running-openssl-1-1-1-after-eol-with-ubuntu-pro), are there any [current vulnerabilities](https://www.cvedetails.com/version/1657305/Openssl-Openssl-1.1.1w.html) and are we aware of [Python performance degradation in OpenSsl 3.0](https://github.com/openssl/openssl/issues/17064)

I hope that arvid and the community can still help to provide official libtorrent wheels on PyPi since it's been a immense help for development and users to have easily available libtorrent python bindings.

xavier2k6:
>The rest of the CI failures due to @xavier2k6 mistaken change to the libtorrent ssl minimum version check which has a different purpose than perceived and now forces openssl 3

It was not a mistaken change but deliberate! see quote below from previous comment.

>**I changed this as a test**, since OpenSSL 1.1.1 is EOL/INSECURE & shouldn't be used.


Anyway, I think the consensus is to continue with wheels.

arvidn:
it seems reasonable to not *require* openssl3, but to support it. using openssl3 when testing builds and running unit tests seems reasonable, as it's probably more important to stay compatible with it.

In my mind, the main friction for building wheels is the current (very complex) `setup.py` which uses the (very complicated) distutils to build the wheel. `b2` builds the shared library that is the native python module just fine. I would think that creating the accompanying files and directories to turn that into a wheel is simple, but I haven't found any documentation for that. `distutils` makes it complicated because it also seems to want to actually trigger the build command now.

xavier2k6:
https://github.com/pypa/manylinux/issues/1818

cas--:
@arvidn I was going to reply here about replacing distutils but instead I've created a new issue: https://github.com/arvidn/libtorrent/issues/7993

xavier2k6:
Closing, I actually had another PR open & pending outcome of any further discussions/directions will either continue that one or scrap it.....time will tell.

PR #7972, opened 2025-06-29
add documentation and changelog entry for get_resume_data()

arvidn:


glassez:
Should it also be mentioned in the documentation that only `save_info_dict` flag has any meaning for `get_resume_data()`?

arvidn:
as you pointed out, maybe that should be enabled unconditionally. I'll take a look at that next

PR #7983, opened 2025-07-13
more immutable `file_storage`

arvidn:
Transition resolving duplicate and invalid filenames to the new structure where torrent_info and file_storage are immutable and renamed files are recorded in a separate structure.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -234,6 +234,7 @@ set(libtorrent_aux_include_files
 	readwrite.hpp
 	receive_buffer.hpp
 	request_blocks.hpp
+	resolve_duplicate_filenames.hpp
 	resolve_links.hpp
 	resolver.hpp
 	resolver_interface.hpp
@@ -391,6 +392,7 @@ set(sources
 	read_resume_data.cpp
 	receive_buffer.cpp
 	request_blocks.cpp
+	resolve_duplicate_filenames.cpp
 	resolve_links.cpp
 	resolver.cpp
 	rtc_signaling.cpp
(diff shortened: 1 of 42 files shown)


milahu:
how to replace `rename_file` in downstream code?

> `qBittorrent/src/base/bittorrent/customstorage.cpp:166:53: warning: 'void libtorrent::file_storage::rename_file(libtorrent::file_index_t, const std::string&)' is deprecated`

[qBittorrent/src/base/bittorrent/customstorage.cpp](https://github.com/qbittorrent/qBittorrent/blob/b819b9b7a6aa2b856b7d70e3de8a71a34b729f44/src/base/bittorrent/customstorage.cpp#L152)

```cc
void CustomDiskIOThread::async_rename_file(lt::storage_index_t storage, lt::file_index_t index, std::string name
                                           , std::function<void (const std::string &, lt::file_index_t, const lt::storage_error &)> handler)
{
    m_nativeDiskIO->async_rename_file(storage, index, name
            , [=, this, handler = std::move(handler)](const std::string &name, lt::file_index_t index, const lt::storage_error &error)
    {
        if (!error)
        {
            #if LIBTORRENT_VERSION_NUM < 20100
            m_storageData[storage].files.rename_file(index, name);
            #else
            // FIXME rename_file is deprecated since libtorrent 77eb55f31e53f798d9b88c946d434b8076310d00
            m_storageData[storage].files.rename_file(index, name);
            #endif
        }
        handler(name, index, error);
    });
}
```

arvidn:
@milahu One way would be to record renamed files in a separate object. I use [renamed_files](https://github.com/arvidn/libtorrent/blob/master/include/libtorrent/file_storage.hpp#L767), which has a similar interface as `file_storage`, but it wraps it.

You could just use a `std::map<lt::file_index, std::string>` as well, to just record the renamed files.

Issue #7693, opened 2024-06-27
32-bit signed int is not enough for transfer speeds


HanabishiRecca:
libtorrent version (or branch): 2.0.10

platform/architecture: Linux x86-64

compiler and compiler version: gcc 14.1.1

As the title says, I'm actually able to acheive transfers speeds more than 2 GiB/s in qBittorrent and exceed the max int32 value. This leads to weird results.

Notice the `Unknown` column value.

![screenshot](https://github.com/arvidn/libtorrent/assets/13597663/258188b6-45c7-4170-acf0-24aecefb56b0)

And even weirder on a seeding side. Alongside with `Unknown` upload speeds, I'm actually able to achieve **negative** share ratio. 🙃

![screenshot](https://github.com/arvidn/libtorrent/assets/13597663/ca10a0af-bbfa-49d8-86a2-28c9706f31df)

The roots of the problem is fairly straghtforward:

https://github.com/arvidn/libtorrent/blob/ba3a13c3413b52b78e0f23edc642690db1dfd848/include/libtorrent/torrent_status.hpp#L316-L317

https://github.com/arvidn/libtorrent/blob/ba3a13c3413b52b78e0f23edc642690db1dfd848/include/libtorrent/torrent_status.hpp#L323-L324

Seems like it's time to promote the values to `int64_t`.

glassez:
IMO, many of the data sizes used are already obsolete, so now problems related to speeds and sizes are becoming more common (e.g. #7735). Interestingly, in some such cases, it is clear that the algorithms used normally cope with the current needs of users, it is enough only to increase the data types used.

PR #7990, opened 2025-07-19
use 64 bit rate limits to support higher transfer rates

arvidn:
An experiment to address https://github.com/arvidn/libtorrent/issues/7693


xavier2k6:
@arvidn How likely is work to continue on this anytime soon & backported to `RC_2_0` prior to the upcoming `2.0.12` release or will it be `2.1.x` only?

PR #7992, opened 2025-07-20
add load torrent api that no throw exceptions

aldenml:


qBittUser:
Do consider writing PR and Commit title differently.

Maybe replace "`that no throw`" with:
"`that doesn't throw`" or "`that does not throw`".

src/torrent_info.cpp
```diff
@@ -51,7 +51,9 @@ see LICENSE file.
 
 namespace libtorrent {
 
+#if TORRENT_ABI_VERSION < 4
```
arvidn, line 54: removing `from_span` seems unrelated to the main patch, does it interact with the other changes?
CI doesn't test building all deprecation versions, so these kinds of changes need to be manually checked.

include/libtorrent/torrent_info.hpp
```diff
@@ -64,11 +64,15 @@ namespace aux {
...
 
 	// hidden
+#if TORRENT_ABI_VERSION < 4
 	class from_span_t {};
+#endif
 	class from_info_section_t {};
 
 	// used to disambiguate a bencoded buffer and a filename
+#if TORRENT_ABI_VERSION < 4
 	extern TORRENT_EXPORT from_span_t from_span;
```
arvidn, line 74:
```suggestion
	extern TORRENT_DEPRECATED_EXPORT from_span_t from_span;
```

include/libtorrent/load_torrent.hpp
```diff
@@ -28,14 +28,20 @@ namespace libtorrent {
...
 	TORRENT_EXPORT add_torrent_params load_torrent_file(
 		std::string const& filename);
+	TORRENT_EXPORT add_torrent_params load_torrent_buffer(
+		span<char const> buffer, error_code& ec, load_torrent_limits const& cfg);
 	TORRENT_EXPORT add_torrent_params load_torrent_buffer(
 		span<char const> buffer, load_torrent_limits const& cfg);
 	TORRENT_EXPORT add_torrent_params load_torrent_buffer(
 		span<char const> buffer);
+	TORRENT_EXPORT add_torrent_params load_torrent_parsed(
+		bdecode_node const& torrent_file, error_code& ec, load_torrent_limits const& cfg);
```
arvidn, line 49: these new overloads should be covered by `test/test_torrent_info.cpp` as well.

include/libtorrent/load_torrent.hpp
```diff
@@ -28,14 +28,20 @@ namespace libtorrent {
 	//   * added to a session via add_torrent() or async_add_torrent()
 	//   * saved as a .torrent_file via write_torrent_file()
 	//   * turned into a magnet link via make_magnet_uri()
+	TORRENT_EXPORT add_torrent_params load_torrent_file(
```
arvidn, line 36: I think the documentation (above) should be updated to describe the non-throwing versions. But keep in mind that they may still throw `std::bad_alloc`

src/torrent_info.cpp
aldenml: you are right

include/libtorrent/load_torrent.hpp
aldenml: will do

include/libtorrent/torrent_info.hpp
aldenml: hi @arvidn , I'm getting this error with clang macOS:
```
libtorrent/include/libtorrent/torrent_info.hpp:74:9: error: 'deprecated' attribute cannot be applied to types
   74 |         extern TORRENT_DEPRECATED_EXPORT from_span_t from_span;
```

include/libtorrent/load_torrent.hpp
aldenml: added only for `load_torrent_file` (similar to the existing), since it use the others.

arvidn:
thanks @aldenml !

PR #7997, opened 2025-07-25
Fix use after move

glassez:


src/session_impl.cpp
```diff
@@ -4922,7 +4922,7 @@ namespace {
 
 #ifndef TORRENT_DISABLE_EXTENSIONS
 		auto extensions = std::move(params.extensions);
-		auto const userdata = std::move(params.userdata);
+		auto const userdata = params.userdata;
```
glassez, line 4925: `add_torrent_params::userdata` is accessed in torrent constructor:
https://github.com/arvidn/libtorrent/blob/a98873873880b815e1d3275ad59cabe4d6d41125/src/torrent.cpp#L240

PR #8010, opened 2025-08-16
add test for torrent with two invalid directory names

arvidn:
merging two directories and still de-duplicating the name


glassez:
Opening this file in the current qBittorrent gives the following result:

glassez:
Well, I figured out what the problem is.
`libtorrent` substitutes "invalid" characters in file/folder names when it parses .torrent file so such "cleaned" paths are considered as "original filenames". Then it resolves duplicate filenames but applies this on top of existing "file storage" so renamed files don't considered as "original filenames". I feel this behavior inconsistent.

As for qBittorrent, we assume that just parsed `torrent_info` contains should not contain any "renamed" files and we ourselves  don't use `torrent_info` for renaming the files. We force use `torrent_info::orig_files()` when we want to be sure that we get "original filenames" rather than renamed by user.

arvidn:
in my effort to separate the immutable and the mutable parts of torrents, I've discovered this issue as well. In `master` I've moved all renaming of files into the same structure. So invalid filenames and user-renamed files are treated the same, and the original filenames are truly original.

I think the original thought was that invalid filenames should be rare and mostly indicative of an attack. It's not obvious what can be done about it in the 2.0 branch. I'm open to suggestions.

glassez:
> I think the original thought was that invalid filenames should be rare and mostly indicative of an attack.

I wouldn't go that far. It seems that the source of this issue is often that names that are valid on one platform may not be valid on another.

glassez:
> It's not obvious what can be done about it in the 2.0 branch. I'm open to suggestions.

If we don't go into details, then you need to do all the cleanups at the same step, either 
1. at the initial parsing step (and then treat the cleaned names as the original ones), or 
2. at the second pass (then the original names must remain really original, even if they are invalid on this system).

I would also like to add that I am a proponent of strict parsing, i.e. if we encounter real errors (for example, duplicate filenames among the original ones), then such a torrent should simply be rejected with an error.

glassez:
@arvidn 

Also I believe the cleanup algorithm should never put the files in the same folder if they were originally in different folders, i.e. different folders with invalid characters in their names should not be renamed into the same name.

arvidn:
I see what you mean about mixing parsing and sanitizing paths. However, it's not so simple. For example, path elements are allowed to be empty in `.torrent` files. I think it would be very misleading (and error prone) to represent that as a double slash, `//` in the `file_storage`. Also, path elements are allowed to contain any UTF-8 character, including `/`. It would be very misleading and error prone to include such directory name verbatim as well.

When a path is turned into a string, `/` must have special meaning, so names must at least be sanitized of those.

There are a few other important ones as well, e.g. `con` on windows is likely to hang the process if accessed. I think it's a net positive to sanitize such filenames.

The risk being that the client fails to sanitize the original filenames and creates a vulnerability.

Even in `master`, the sanitization happens very early, along with the parsing. But the de-duplication happens later and preserves the sanitized names.

glassez:
> Even in `master`, the sanitization happens very early, along with the parsing. But the de-duplication happens later and preserves the sanitized names.

I don't know how it behaves in the `master`, but the downside of the current branches is that it breaks the directory structure during name sanitization.

glassez:
> For example, path elements are allowed to be empty in `.torrent` files. I think it would be very misleading (and error prone) to represent that as a double slash, `//` in the `file_storage`. Also, path elements are allowed to contain any UTF-8 character, including `/`.

Don't tell me there are good reasons why such "flaws" are allowed by the BitTorrent protocol specifications.

PR #8017, opened 2025-08-23
Draft: Feat: Add I2P reconnection fix #7453

### LINKED ISSUE #7453: I2P SAM connection is not reestablished when closed
Tested with client_test and i2pd.
1. Start i2pd.
2. Start client_test and verify that SAM session is active at http://127.0.0.1:7070/?page=sam_sessions.
3. Stop i2pd _(Force shutdown)_.
4. Start i2pd again.

Expected behaviour: SAM session is recreated once i2pd is up again (after step 4).
Actual behaviour: no session is (re)created, SAM sessions page is empty.

Reason: i2pd crashes sometimes and it is more convenient for user to restart just i2pd and not both i2pd and torrent client.

libtorrent version (or branch): c2012b084c6654d681720ea0693d87a48bc95b14
platform/architecture: Windows
compiler and compiler version: clang 15.0.7

covert8:
This commit adds i2p sam re-connection with exponential backoff (max 5m)

### DIFF
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -715,6 +715,8 @@ namespace aux {
 			void on_i2p_open(error_code const& ec);
 			void open_new_incoming_i2p_connection();
 			void on_i2p_accept(error_code const& e);
+			void schedule_i2p_reconnect();
+			void on_i2p_reconnect_timer(error_code const& ec);
 #endif
 
 			void start_ip_notifier();
@@ -1018,6 +1020,13 @@ namespace aux {
 #if TORRENT_USE_I2P
 			i2p_connection m_i2p_conn;
 			std::optional<socket_type> m_i2p_listen_socket;
+			
+			// Timer for I2P reconnection attempts
+			deadline_timer m_i2p_reconnect_timer;
+			
+			// Exponential backoff state for I2P reconnection
+			int m_i2p_reconnect_delay = 1;  // seconds
+			bool m_i2p_reconnecting = false;
 #endif
 
 #if TORRENT_USE_SSL
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -507,6 +507,7 @@ bool ssl_server_name_callback(ssl::stream_handle_type stream_handle, std::string
 		, m_work(make_work_guard(m_io_context))
 #if TORRENT_USE_I2P
 		, m_i2p_conn(m_io_context)
+		, m_i2p_reconnect_timer(m_io_context)
 #endif
 		, m_created(clock_type::now())
 		, m_last_tick(m_created)
@@ -1030,6 +1031,7 @@ bool ssl_server_name_callback(ssl::stream_handle_type stream_handle, std::string
 		m_timer.cancel();
 
 #if TORRENT_USE_I2P
+		m_i2p_reconnect_timer.cancel();
 		m_i2p_conn.close(ec);
 #endif
 		stop_ip_notifier();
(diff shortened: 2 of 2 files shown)


include/libtorrent/aux_/session_impl.hpp
```diff
@@ -1018,6 +1020,15 @@ namespace aux {
...
 			std::optional<socket_type> m_i2p_listen_socket;
+			
+			// Timer for I2P reconnection attempts
+			deadline_timer m_i2p_reconnect_timer;
+			
+			// Exponential backoff state for I2P reconnection
+			int m_i2p_reconnect_delay = 1;  // seconds
+			bool m_i2p_reconnecting = false;
+			static constexpr int m_i2p_initial_reconnect_delay = 1;  // 1 second
+			static constexpr int m_i2p_max_reconnect_delay = 300;  // 5 minutes
```
arvidn: I don't think these should be members of this class. Would you mind moving them into the anonymous namespace (i.e. internal linkage) of `session_impl.cpp`?

(and remove the `m_`-prefix)

src/session_impl.cpp
```diff
@@ -2388,6 +2395,8 @@ namespace {
 #endif
 
 #if TORRENT_USE_I2P
+	
+
```
```suggestion
```

src/session_impl.cpp
```diff
@@ -2448,6 +2465,53 @@ namespace {
...
+		
+		m_i2p_reconnecting = true;
+		
+		// Cancel any existing timer
+		m_i2p_reconnect_timer.cancel();
+		
+		// Schedule reconnection with current delay
+		m_i2p_reconnect_timer.expires_after(seconds(m_i2p_reconnect_delay));
+		m_i2p_reconnect_timer.async_wait(
+			std::bind(&session_impl::on_i2p_reconnect_timer, this, _1));
```
arvidn: I think it might be safer to pass in a weak reference to the session impl here, rather than `this`. Something like this:

```
std::weak_ptr<session_impl> self = shared_from_this();
m_i2p_reconnect_timer.async_wait(
	std::bind(&on_i2p_reconnect_timer, std::move(self), _1));
```

The handler must then take a `weak_ptr`, promote it to a `shared_ptr`, and only if that's successful, actually call the member function `on_i2p_reconnect_timer()`

include/libtorrent/aux_/session_impl.hpp
```diff
@@ -1018,6 +1020,15 @@ namespace aux {
 #if TORRENT_USE_I2P
 			i2p_connection m_i2p_conn;
 			std::optional<socket_type> m_i2p_listen_socket;
+			
+			// Timer for I2P reconnection attempts
+			deadline_timer m_i2p_reconnect_timer;
+			
+			// Exponential backoff state for I2P reconnection
+			int m_i2p_reconnect_delay = 1;  // seconds
```
arvidn, line 1028:
```suggestion
			seconds m_i2p_reconnect_delay = seconds(1);
```

(and make sure `libtorrent/time.hpp` is included)

include/libtorrent/aux_/session_impl.hpp
arvidn:
```suggestion
			static constexpr seconds m_i2p_initial_reconnect_delay = seconds(1);
			static constexpr seconds m_i2p_max_reconnect_delay = minutes(5);
```

arvidn (commented):
looks good! I mostly have some style suggestsions. Did you test this? Does it work after hibernation?

covert8:
Thanks for the review, I'll get these sorted out asap

covert8:
Should be in order now; i'll be deploying this so perhaps hold off until tested

edit: as for suspend/hibernate this should work as the deadline_timer should just fire if it was scheduled after the suspend/hibernate

include/libtorrent/aux_/session_impl.hpp
arvidn: you marked this resolved without addressing it, as far as I can tell.

Neustradamus:
Linked to:
- https://github.com/arvidn/libtorrent/pull/8055

Neustradamus:
It has been merged into:
- https://github.com/arvidn/libtorrent/pull/8055

Thanks to @covert8 and @arvidn!

PR #8020, opened 2025-08-27
Debugging link local interfaces

tomleb:
Just wanted to debug why libtorrent isn't announcing when the only interfaces are the loopback or the link local.


tomleb:
Bleh was meant to open this on my own fork, not here.

Issue #7651, opened 2024-03-16
Using HTTP Proxy: Retrieved Host is IP Address instead of Domain in HTTPS Tracker URL


bearx3f:
**Please provide the following information**

libtorrent version (or branch): 2.0.10

platform/architecture: qbittorrent-static-nox x86_64

compiler and compiler version: GCC 13.2.0 -- x86_64-linux-musl

I'm using qbittorrent and having http proxy settings 

![image](https://github.com/arvidn/libtorrent/assets/9717561/699b2bfb-3448-47cb-9e12-1a221abe09cf)

And when announce to HTTPS Tracker libtorrent not sent Host name on CONNECT request

arvidn:
correct. the `CONNECT` command optionally includes `Proxy-Authorization:`, but not the hostname of the destination.

The `Host` header is sent inside the SSL tunnel, to the destination server.

Is there a problem? Does the proxy need to know the hostname (as opposed to just the IP address) of the destination server?

bearx3f:
there's a problem to sign the ssl connection in mitm mode proxy

arvidn:
I'm intrigued. What's the feature or utility that's enabled by the proxy man-in-the-middling the connection?

You would also have to install their SSL cert for libtorrent to pick up as well. Is that a certification signed for all hostnames?

bearx3f:
> What's the feature or utility that's enabled by the proxy man-in-the-middling the connection?

I'm tracking the download, upload and event of the torrent

> Is that a certification signed for all hostnames?

No, just sign for each hostname of each request

arvidn:
@bearx3f could you try this patch out?
https://github.com/arvidn/libtorrent/pull/7710

bearx3f:
Thank you, for your hard work I'll try

bearx3f:
I'm sorry I cannot test your code but after review your change I think may you change code at http_connection.cpp@591

```cpp
	if (m_proxy.send_host_in_connect)
	{
		if (boost::get<http_stream>(&*m_sock))
		{
			boost::get<http_stream>(*m_sock).set_host(m_hostname);
		}
#if TORRENT_USE_SSL
		else if (m_ssl && boost::get<ssl_stream<http_stream>>(&*m_sock))
		{
			boost::get<ssl_stream<http_stream>>(*m_sock).next_layer().set_host(m_hostname);
		}
#endif
	}
```

bearx3f:
[...]

PR #8021, opened 2025-08-29
Fix Handling of HTTP CONNECT Header in Proxy Connections

bearx3f:
From #7651 After I tested the latest pull request, which I'd previously said passed, the proxy server still couldn't sign the certificate when deployed with the actual application. This was because the server didn't receive the remote server in the correct `domain:port` format, such as `example.com:443`. In this patch, I've fixed the `CONNECT` request to be correct.

I noticed this by observing the request sent by Curl, which sends:

```
CONNECT example.com:443 HTTP/1.0
Host: example.com:443
```

Originally, Libtorrent was sending:

```
CONNECT 23.220.75.232:443 HTTP/1.0
Host: example.com
```

In this original request, the Host header was missing the port, and the **CONNECT** method was sending an IP address instead of a domain name.

### DIFF
--- a/simulation/test_http_connection.cpp
+++ b/simulation/test_http_connection.cpp
@@ -618,18 +618,27 @@ void test_connection_ssl_proxy(bool const with_hostname)
 	int client_counter = 0;
 	int proxy_counter = 0;
 
-	http_proxy.register_handler("10.0.0.2:8080"
-		, [&proxy_counter, with_hostname](std::string method, std::string req, std::map<std::string, std::string>& headers)
+	std::string expected_target = with_hostname ? "test-hostname.com:8080" : "10.0.0.2:8080";
+
+	http_proxy.register_handler(expected_target
+		, [&proxy_counter, with_hostname, expected_target](std::string method, std::string req, std::map<std::string, std::string>& headers)
 		{
 			proxy_counter++;
 			TEST_EQUAL(method, "CONNECT");
+			
+			// Host header is always sent to comply with RFC 9110 and RFC 9112 requirements.
+			// The send_host_in_connect setting controls the format:
+			// - true: Host header contains domain:port format
+			// - false: Host header contains ip:port format
 			if (with_hostname)
 			{
-				TEST_EQUAL(headers["host"], "10.0.0.2");
+				// When send_host_in_connect is true, Host header should contain domain:port
+				TEST_EQUAL(headers["host"], "test-hostname.com:8080");
 			}
 			else
 			{
-				TEST_CHECK(headers.empty());
+				// When send_host_in_connect is false, Host header should contain ip:port
+				TEST_EQUAL(headers["host"], "10.0.0.2:8080");
 			}
 			return sim::send_response(403, "Not supported", 1337);
 		});
@@ -655,7 +664,9 @@ void test_connection_ssl_proxy(bool const with_hostname)
 #endif
 		);
 
-	h->start("10.0.0.2", 8080, seconds(1), &ps, true /*ssl*/);
+	// Use hostname when testing with_hostname=true, IP when with_hostname=false
+	std::string target_host = with_hostname ? "test-hostname.com" : "10.0.0.2";
+	h->start(target_host, 8080, seconds(1), &ps, true /*ssl*/);
 
 	sim.run();
 
@@ -663,13 +674,17 @@ void test_connection_ssl_proxy(bool const with_hostname)
 	TEST_EQUAL(proxy_counter, 1);
 }
 
-// Requests a proxied SSL connection. This test just ensures that the correct CONNECT request
-// is sent to the proxy server.
+// Tests SSL proxy connection with send_host_in_connect=false.
+// Uses IP address for connection and Host header should contain ip:port format.
+// This verifies that the Host header is always present and contains the target IP:port.
 TORRENT_TEST(http_connection_ssl_proxy_no_hostname)
 {
 	test_connection_ssl_proxy(false);
 }
 
+// Tests SSL proxy connection with send_host_in_connect=true.
+// Uses hostname for connection and Host header should contain domain:port format.
+// This ensures proper hostname handling when send_host_in_connect is enabled.
 TORRENT_TEST(http_connection_ssl_proxy_hostname)
 {
 	test_connection_ssl_proxy(true);


include/libtorrent/http_stream.hpp
```diff
@@ -98,6 +99,80 @@ class http_stream : public proxy_base
...
+	// - if port == 0, return host unchanged
+	// - if host is bracketed IPv6: [addr] or [addr]:port, append port only if missing
+	// - if host contains no ':', append ":port"
+	// - if host contains colon(s):
+	//     * if suffix after last ':' is all digits, assume it's already a host:port -> leave unchanged
+	//     * otherwise treat it as an (unbracketed) IPv6 literal or hostname with colons and wrap
+	//       it in brackets and append :port -> [host]:port
+	static std::string format_host_for_connect(std::string host, unsigned short const port)
+	{
+		if (port == 0) return host;
```
arvidn:
```suggestion
		if (port == 80) return host;
```
I imagine this is what you meant, right?

include/libtorrent/http_stream.hpp
```diff
@@ -98,6 +99,80 @@ class http_stream : public proxy_base
...
 
+	// format a hostname (not a numeric IP) with a port for an HTTP CONNECT request.
+	// rules:
+	// - if port == 0, return host unchanged
+	// - if host is bracketed IPv6: [addr] or [addr]:port, append port only if missing
+	// - if host contains no ':', append ":port"
+	// - if host contains colon(s):
+	//     * if suffix after last ':' is all digits, assume it's already a host:port -> leave unchanged
+	//     * otherwise treat it as an (unbracketed) IPv6 literal or hostname with colons and wrap
+	//       it in brackets and append :port -> [host]:port
```
arvidn: rather than describing exactly what this function is doing, in detail, it would be much better to describy *why* it's doing it. Perhaps even with a reference to the appropriate RFC describing the host header format. And maybe with a quote from it

include/libtorrent/http_stream.hpp
```diff
@@ -98,6 +99,80 @@ class http_stream : public proxy_base
...
+	static std::string format_host_for_connect(std::string host, unsigned short const port)
+	{
+		if (port == 0) return host;
+
+		if (!host.empty() && host.front() == '[')
+		{
+			auto const rb = host.find(']');
+			bool const has_port = (rb != std::string::npos && rb + 1 < host.size() && host[rb + 1] == ':');
```
arvidn: does the host sometimes come with a port? that doesn't sound right.
I think it should either be an assert that it doesn't have a port, or a comment explaining why it (sometimes) has a port

include/libtorrent/http_stream.hpp
bearx3f: In some case port=0 there's no port to be append

include/libtorrent/http_stream.hpp
bearx3f: I will assume that host will not already has port and then assert there's port in host

include/libtorrent/http_stream.hpp
bearx3f: I will change it

include/libtorrent/http_stream.hpp
bearx3f: and then add ASSERT `host`

include/libtorrent/http_stream.hpp
```diff
@@ -98,6 +99,41 @@ class http_stream : public proxy_base
...
+	// Format a hostname with port for HTTP CONNECT request per RFC 9110 Section 9.3.6.
+	// The authority component must be in the form "host:port" where IPv6 literals
+	// are enclosed in square brackets as defined in RFC 3986 Section 3.2.2.
+	// Assumes host parameter contains no port suffix.
+	// Rules:
+	// - if port == 0, return host unchanged
+	// - if host is already bracketed IPv6: [addr], append ":port"
+	// - if host contains colons (IPv6), bracket and append ":port" -> [host]:port
+	// - otherwise append ":port" for regular hostnames/IPv4
+	static std::string format_host_for_connect(std::string host, unsigned short const port)
```
arvidn: I think this should be a free function (not a static member). It should be defined in a `.cpp` file. And it should have a unit test, to make sure it handles all the edge cases correctly?
Since `http_stream` doesn't have a `.cpp` file, maybe it would make sense to put this in `string_util.cpp`, with the test in `test_string.cpp`

test/http_proxy.py
```diff
@@ -209,15 +209,20 @@ def connect_request(self):
...
+                self.log_message("Host header matches CONNECT target")
+            else:
+                self.log_message("Host header differs from CONNECT target")
         else:
-            self.log_message("CONNECT request without Host header (target: %s:%s)",
+            self.log_message("CONNECT request: target=%s:%s, NO Host header", 
                            host, port)
 
         # If require_host_header is set, reject requests without Host header
-        if self.require_host_header and not host_header:
```
arvidn, line 219: you you remove the check for `self.require_host_header` (and make it mandatory), you should remove the comment above as well as *setting* this member to `True` at the top of the class

bearx3f: I'm really sorry.

arvidn:
it looks like the simulation error is legitimate and needs to be addressed

bearx3f:
> it looks like the simulation error is legitimate and needs to be addressed

The problem is my patch will always send Host header but simulation test check if send_host_in_connect=false host should be empty. I will fix.

include/libtorrent/http_stream.hpp
```diff
@@ -128,10 +128,24 @@ class http_stream : public proxy_base
 		// send CONNECT
 		std::back_insert_iterator<std::vector<char>> p(m_buffer);
 		std::string const endpoint = print_endpoint(m_remote_endpoint);
-		write_string("CONNECT " + endpoint + " HTTP/1.0\r\n", p);
+		// if we were given the original host (domain or IP), prefer using it (lets proxy resolve domains)
 		if (!m_host.empty())
 		{
-			write_string("Host: " + m_host + "\r\n", p);
+			std::string const remote_host = ::libtorrent::format_host_for_connect(
+				std::move(m_host),
```
arvidn, line 135: this `std::move()` looks really suspicious. It means this socket can't be reused, which may be an unexpected behavior. Are you certain that this is correct?

Since `m_host` is not passed in via `async_connect()`, I'm pretty sure you can end up setting the host only once, and the connect the socket multiple times, e.g. if the connection fails for some reason.

I don't think you should move `m_host` here.
```suggestion
				m_host, 
```

include/libtorrent/http_stream.hpp
```diff
@@ -128,10 +128,24 @@ class http_stream : public proxy_base
 		// send CONNECT
 		std::back_insert_iterator<std::vector<char>> p(m_buffer);
 		std::string const endpoint = print_endpoint(m_remote_endpoint);
-		write_string("CONNECT " + endpoint + " HTTP/1.0\r\n", p);
+		// if we were given the original host (domain or IP), prefer using it (lets proxy resolve domains)
 		if (!m_host.empty())
 		{
-			write_string("Host: " + m_host + "\r\n", p);
+			std::string const remote_host = ::libtorrent::format_host_for_connect(
```
arvidn, line 134:
```suggestion
			std::string const remote_host = format_host_for_connect(
```

arvidn:
sorry for dropping the ball on this one. I think it looks good. just a few small tweaks

test/http_proxy.py
```diff
@@ -152,7 +152,7 @@ class Handler(http.server.BaseHTTPRequestHandler):
     close_connection = True
     timeout = 30
     basic_auth = None
-    require_host_header = False  # Set to True to require Host header in CONNECT requests
+    require_host_header = True  # Always require Host header in CONNECT requests for proper HTTP compliance
```
arvidn, line 155: I believe I was thinking that this was part of the `BaseHTTPRequestHAndler`, but I don't think it is. Now you can probably just remove this field entirely

arvidn:
@bearx3f how does this look? https://github.com/arvidn/libtorrent/pull/8155

bearx3f:
Thank you, for you resopnse

I don't have much experience with C++, so I can't really provide feedback on your changes. Please feel free to use your best judgment. Also, just to mention, the code I submitted was written with the help of AI.

PR #8025, opened 2025-09-05
 HTTP/2 Tracker Support via libcurl

### LINKED ISSUE #4334: pipeline HTTP tracker requests over the same connection
1.2.4

how to reproduce it.

I have around 2500 torrents in qBt. If I start them all at once, the announces go through all at once and some 500-600 would fail(timed out). I think the tracker gets overwhelmed by the sudden influx of simultaneous announce requests. There should be a limit to max simultaneous announce requests. Maybe make a queue system which would send max "x" amount of announces/second.

TeholTheOnly:
Fixes #4334

Introduces native HTTP/2 support for tracker announces through libcurl integration, enabling connection multiplexing and improved performance while maintaining full backward compatibility.

#### Key Changes
- Added curl_tracker_client implementing tracker_client_interface for HTTP/2 announces
- Implemented dedicated curl_thread_manager for async libcurl operations
- Connection pooling and multiplexing for reduced overhead
- Opt-in feature via settings_pack::use_libcurl_for_trackers
- Comprehensive test suite including HTTP/2 test server

#### Performance
- Reduction in connection overhead through multiplexing
- SSL caching and reuse
- 30-40ms average latency improvement

#### Build
`b2 use-libcurl=on`

#### Breaking Changes
None - feature is disabled by default


TeholTheOnly:
## Dedicated Thread vs ASIO integration

After prototyping both approaches, I chose a single-threaded libcurl design over direct ASIO integration due to fundamental socket handling incompatibilities that would have required extensive platform-specific code.

## The Core Problem: Socket Descriptor Impedance Mismatch
### Platform-Specific Socket Types
```cpp
// Windows: SOCKET (UINT_PTR) vs ASIO's wrapped handles
curl_socket_t sock = ...; // Returns Windows SOCKET
// ASIO expects different wrapper, uses IOCP internally

// Unix: int fd vs ASIO's epoll/kqueue abstractions  
int fd = ...; // Raw file descriptor from curl
// ASIO manages its own epoll/kqueue state
```

### Event Model Incompatibility
libcurl's `curl_multi_socket_action` requires precise per-socket event control that doesn't map cleanly to ASIO:

```cpp
// libcurl needs: "Watch this socket for READ only, now WRITE only, now BOTH"
case CURL_POLL_IN:    // Register for read events only
case CURL_POLL_OUT:   // Register for write events only  
case CURL_POLL_REMOVE: // Unregister all events

// ASIO model: async operations own the socket
socket.async_read_some(...);  // Can't easily switch to write-only monitoring
```

### Platform Event Systems Don't Mix
Each platform would need custom bridging code:

- **Windows**: IOCP (ASIO) vs WSAEventSelect (curl) - incompatible event models
- **Linux**: epoll with ASIO's reactor pattern vs curl's edge-triggered needs  
- **macOS**: kqueue filter management conflicts between libraries
- **BSD**: Similar kqueue issues plus different socket option semantics

### Socket Ownership Conflicts
```cpp
// Race condition example:
// ASIO thinks it owns socket → schedules async_read
// libcurl closes socket → curl_easy_cleanup()  
// ASIO continues → crash on invalid socket
```

## Pragmatic Solution

I chose a dedicated thread with `curl_multi_wait/perform` because:
1. **Zero platform-specific code** - Works identically everywhere
2. **Clear ownership** - One thread owns all curl resources
3. **Proven Solution** - Mirroring transmissions implementation

## Trade-offs
- **One polling thread** - Negligible CPU impact (<0.1% in profiling)

## Conclusion
For tracker announces, the complexity of ASIO integration wasn't justified. The single-threaded design prioritizes **correctness**, **portability**, and **maintainability** over theoretical performance gains that might not materialize in practice.

arvidn:
The two first commits don't belong to this PR. Are they a prerequisite for this to work?

include/libtorrent/aux_/mock_tracker_client.hpp
```diff
@@ -0,0 +1,114 @@
...
+#define TORRENT_MOCK_TRACKER_CLIENT_HPP
+
+#include "libtorrent/aux_/tracker_manager.hpp"
+#include "libtorrent/aux_/deadline_timer.hpp"
+#include "libtorrent/time.hpp"
+
+namespace libtorrent::aux {
+
+// Mock implementation for testing
+class TORRENT_EXTRA_EXPORT mock_tracker_client {
```
arvidn, line 11: This doesn't look like it ought to be part of the production code. I would have thought this would live under `tests/`. Is there a good reason to put it here?

include/libtorrent/aux_/mock_tracker_client.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+    void scrape(
+        tracker_request const& req,
+        std::function<void(error_code const&, tracker_response const&)> handler
+    ) {
+        // Reuse announce logic for scrape
+        announce(req, handler);
+    }
+
+    [[nodiscard]] bool can_reuse() const {
```
arvidn, line 71: it really looks suspicious to have this feature *just* for tests. If the real `curl_tracker_connection` always is reusable, shouldn't the mock tracker connection mimic that behavior?
The production code where this returns true isn't necessarily well tested.

include/libtorrent/aux_/curl_thread_manager.hpp
```diff
@@ -0,0 +1,325 @@
...
+struct curl_request_context {
+    std::shared_ptr<curl_transfer_data> transfer_data;
+    curl_request request;
+};
+
+// Efficient incremental tracker host counting for dynamic connection pool scaling
+class tracker_host_counter {
+private:
+    std::unordered_map<std::string, int> m_tracker_ref_counts;
+    mutable std::mutex m_mutex;
```
arvidn, line 99: it doesn't look right that the counters have their own mutex. Shouldn't the counters be protected by the same mutex as the thing they are counting? i.e. the announce queue I assume.

src/curl_thread_manager.cpp
```diff
@@ -0,0 +1,1609 @@
...
+        // Post error to completion handler for each pending retry
+        boost::asio::post(m_ios, [handler = item.request.completion_handler]() {
+            handler(errors::session_is_closing, std::vector<char>{});
+        });
+    }
+    m_retry_queue.clear();
+}
+
+void curl_thread_manager::wakeup_curl_thread() {
+    // Simple delegation to perform_wakeup for backward compatibility
```
arvidn, line 555: backwards compatibility with what?

src/curl_thread_manager.cpp
```diff
@@ -0,0 +1,1609 @@
...
+    // The actual batching is now handled via process_queue_notification
+    perform_wakeup();
+}
+
+void curl_thread_manager::perform_wakeup() {
+    // FIX: Allow wakeup during shutdown to prevent deadlock
+    // The curl thread needs to wake up from curl_multi_poll() to check
+    // the shutdown flag and exit cleanly
```
arvidn, line 563: this sounds important

include/libtorrent/aux_/curl_thread_manager.hpp
```diff
@@ -0,0 +1,325 @@
...
+    // Wakeup the curl thread (requires libcurl 7.68.0+)
+    void wakeup_curl_thread();
+
+    void process_queue_notification();
+    void on_timer(boost::system::error_code const& ec);
+    void perform_wakeup();
+
+private:
+    // Memory pool for response buffers with fine-grained locking
+    class response_buffer_pool {
```
arvidn, line 172: did you benchmark to ensure this complexity is worth it? It's not obvious to me

include/libtorrent/aux_/curl_thread_manager.hpp
```diff
@@ -0,0 +1,325 @@
...
+    mutable std::mutex m_queue_mutex;
+    std::queue<curl_request> m_request_queue;
+
+    // Multi handle (thread-safe for curl_multi_wakeup)
+    std::atomic<CURLM*> m_multi_handle{nullptr};
+
+    std::atomic<bool> m_shutting_down{false};
+
+    // Timer-based wakeup batching mechanism
+    static constexpr auto WAKEUP_DELAY = std::chrono::milliseconds(5);
```
arvidn, line 281: I can't tell what this is for, by just reading this comment. Does it really need to be a member?

include/libtorrent/aux_/curl_thread_manager.hpp
```diff
@@ -0,0 +1,325 @@
...
+    std::condition_variable m_init_cv;
+    InitStatus m_init_status = InitStatus::Pending;
+
+    mutable std::mutex m_queue_mutex;
+    std::queue<curl_request> m_request_queue;
+
+    // Multi handle (thread-safe for curl_multi_wakeup)
+    std::atomic<CURLM*> m_multi_handle{nullptr};
+
+    std::atomic<bool> m_shutting_down{false};
```
arvidn, line 278: this looks suspicious. When you shut down, you also need to drain the request queue, to ensure all handlers are notified. So shutting down, ideally, should be part of the same message queue. Making it an atomic suggests that you're checking it out-of-band

include/libtorrent/aux_/curl_thread_manager.hpp
```diff
@@ -0,0 +1,325 @@
...
+        Pending,
+        Success,
+        Failed
+    };
+    std::mutex m_init_mutex;
+    std::condition_variable m_init_cv;
+    InitStatus m_init_status = InitStatus::Pending;
+
+    mutable std::mutex m_queue_mutex;
+    std::queue<curl_request> m_request_queue;
```
arvidn, line 273: I really think this would be a lot simpler (and maybe even perform better) if you'd just use a plain condition variable together with this queue and mutex.

include/libtorrent/aux_/curl_thread_manager.hpp
```diff
@@ -0,0 +1,325 @@
...
+    std::atomic<CURLM*> m_multi_handle{nullptr};
+
+    std::atomic<bool> m_shutting_down{false};
+
+    // Timer-based wakeup batching mechanism
+    static constexpr auto WAKEUP_DELAY = std::chrono::milliseconds(5);
+    deadline_timer m_wakeup_timer;              // Accessed ONLY on IO thread
+    bool m_timer_running = false;               // Accessed ONLY on IO thread
+    std::atomic<bool> m_notification_pending{false}; // Cross-thread notification
+    std::atomic<uint64_t> m_notification_version{0}; // Version counter for debugging
```
arvidn, line 285: Is this better than a condition variable?

include/libtorrent/aux_/curl_thread_manager.hpp
```diff
@@ -0,0 +1,325 @@
...
+    struct retry_item {
+        time_point scheduled_time;
+        curl_request request;
+
+        // Comparison operator for multiset ordering (earliest time first)
+        bool operator<(const retry_item& other) const {
+            return scheduled_time < other.scheduled_time;
+        }
+    };
+    std::multiset<retry_item> m_retry_queue;
```
arvidn, line 301: when and why are requests retried? Is the retry mechanism circumventing the higher level announce timers in the `torrent` object?

src/curl_thread_manager.cpp
```diff
@@ -0,0 +1,1609 @@
...
+        }
+    }
+}
+
+int curl_thread_manager::process_completions(CURLM* multi, curl_handle_pool& pool) {
+    CURLMsg* msg;
+    int msgs_left;
+    int completed_count = 0;
+
+    // CORRECT: Loop while reading messages from curl
```
arvidn, line 1363: what do you mean by "CORRECT: " in this comment? Is it the *verb* or the *adjective* "correct"?

src/curl_thread_manager.cpp
```diff
@@ -0,0 +1,1609 @@
...
+    while ((msg = curl_multi_info_read(multi, &msgs_left))) {
+        // DEBUG: Log any completed message
+        if (msg->msg == CURLMSG_DONE) {
+// Removed verbose debug logging
+        }
```
arvidn, line 1368: this looks like left-over debugging code

src/curl_thread_manager.cpp
```diff
@@ -0,0 +1,1609 @@
...
+    curl_easy_setopt(easy, CURLOPT_PROTOCOLS, CURLPROTO_HTTP | CURLPROTO_HTTPS);
+#pragma GCC diagnostic pop
+#endif
+
+    // Enable connection reuse (TCP Keepalive)
+    curl_easy_setopt(easy, CURLOPT_TCP_KEEPALIVE, 1L);
+    curl_easy_setopt(easy, CURLOPT_TCP_KEEPIDLE, 120L);
+    curl_easy_setopt(easy, CURLOPT_TCP_KEEPINTVL, 60L);
+
+    // HTTP/2 support if available
```
arvidn, line 768:
```suggestion
    // HTTP/2 support is available
```

include/libtorrent/aux_/curl_thread_manager.hpp
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    // Clear all counts (for shutdown)
+    void clear() {
+        std::scoped_lock<std::mutex> lock(m_mutex);
+        m_tracker_ref_counts.clear();
+    }
+};
+
+class TORRENT_EXPORT curl_thread_manager : public std::enable_shared_from_this<curl_thread_manager> {
```
arvidn, line 121: this name seems a bit misleading. This class, as I understand it, is the interface with curl, queuing new requests and dispatches response handlers.

It's more of a *request* manager than a *thread* manger, is it not?

include/libtorrent/aux_/curl_thread_manager.hpp
```diff
@@ -0,0 +1,325 @@
...
+public:
+    static std::shared_ptr<curl_thread_manager> create(
+        io_context& ios, session_settings const& settings);
+
+    ~curl_thread_manager();
+
+    void add_request(
+        std::string const& url,
```
arvidn, line 129: is it not possible to make more sophisticated requests with curl? e.g. affecting request headers, or using a proxy. The URL doesn't seem like enough to make a request.
Could `tracker_request` be used instead?

include/libtorrent/aux_/curl_thread_manager.hpp
```diff
@@ -0,0 +1,325 @@
...
+        std::function<void(error_code, std::vector<char>)> handler,
+        time_duration timeout = seconds(30));
+
+    void shutdown();
+
+    void tracker_added(const std::string& url);
```
arvidn, line 135: what's the difference between this and `add_request()`?

include/libtorrent/aux_/curl_thread_manager.hpp
```diff
@@ -0,0 +1,325 @@
...
+    void tracker_removed(const std::string& url);
```
arvidn, line 136: is this a way to cancel a request?

include/libtorrent/aux_/curl_thread_manager.hpp
```diff
@@ -0,0 +1,325 @@
...
+
+    [[nodiscard]] curl_thread_stats get_stats() const;
```
arvidn, line 138: is this used anywhere? If not, I think it should be removed.

src/curl_thread_manager.cpp
```diff
@@ -0,0 +1,1609 @@
...
+            #endif
+        }
+    }
+}
+
+std::shared_ptr<curl_thread_manager> curl_thread_manager::create(
+    io_context& ios, session_settings const& settings)
+{
+    auto manager = std::shared_ptr<curl_thread_manager>(
+        new curl_thread_manager(ios, settings));
```
arvidn, line 451: if you have to allocate objects in `shared_ptr<>`, please use `make_shared()`

src/curl_thread_manager.cpp
```diff
@@ -0,0 +1,1609 @@
...
+
+    // FIX Issue 1: Break shared_ptr reference cycle.
+    // The original code captured a shared_ptr in the thread lambda, causing a leak.
+    // We use the raw pointer instead. This is safe because the destructor joins the thread.
+    manager->m_curl_thread = std::thread(
+        &curl_thread_manager::curl_thread_func, manager.get());
```
arvidn, line 457: 2-phase initialization should typically be avoided. is there a good reason to initialize the thread after constructing the object?

src/curl_thread_manager.cpp
```diff
@@ -0,0 +1,1609 @@
...
+        // Check if initialization failed (only if status changed from Pending)
+        if (manager->m_init_status == InitStatus::Failed) {
+            // Join the failed thread before throwing
+            if (manager->m_curl_thread.joinable()) {
+                manager->m_curl_thread.join();
+            }
+            throw std::runtime_error("Failed to initialize curl multi handle");
+        }
+        // If still pending, that's OK - initialization will complete asynchronously
+    }
```
arvidn, line 478: this doesn't look right. `std::thread` will throw an exception if it fails. If it doesn't fail, the thread will be created. It seems quite unnecessary to handle the case where `std::thread` or the operating system isn't functioning properly.

src/curl_thread_manager.cpp
```diff
@@ -0,0 +1,1609 @@
...
+    if (m_notification_pending.compare_exchange_strong(expected, true,
+        std::memory_order_acq_rel, std::memory_order_acquire)) {
+        // Post notification with shared_from_this for safety
+        boost::asio::post(m_ios, [self = shared_from_this()]() {
+            self->process_queue_notification();
+        });
+    }
+}
+
+std::vector<curl_request> curl_thread_manager::swap_pending_requests() {
```
arvidn, line 678: this name does not describe what this function does.

I think what you want is, essentially:
```
std::unique_lock<std::mutex> lock(m_queue_mutex);
return std::exchange(m_request_queue, std::vector<curl_request>());
```

But it's not clear why.

src/curl_thread_manager.cpp
```diff
@@ -0,0 +1,1609 @@
...
+        #endif
+
+        // Store multi handle for wakeup mechanism
+        m_multi_handle = multi.get();
+
+        // Signal that initialization is successful
+        {
+            std::scoped_lock<std::mutex> lock(m_init_mutex);
+            m_init_status = InitStatus::Success;
+        }
```
arvidn, line 1026: why not just perform the initialization in the original thread instead?
Now you have to add synchronization just so the original thread can wait for the newly created one. I think it would be simpler if you just create the thread after the initialization is done.

and remove `m_init_mutex` and `m_init_cv`

test/test_tracker_interface.cpp
```diff
@@ -0,0 +1,198 @@
+#include "test.hpp"
+#include "libtorrent/aux_/mock_tracker_client.hpp"
```
arvidn, line 2: you should probably move the mock implementation into this test

test/test_tracker_interface.cpp
```diff
@@ -0,0 +1,198 @@
...
+    TEST_EQUAL(resp.interval.count(), expected_response.interval.count());
+    TEST_EQUAL(resp.complete, expected_response.complete);
+    TEST_EQUAL(resp.incomplete, expected_response.incomplete);
+}
+
+TORRENT_TEST(tracker_scrape_behavior)
+{
+    io_context ios;
+    settings_pack settings;
+    mock_tracker_client client(ios, settings);
```
arvidn, line 80: what's the point of these tests? They don't seem to actually test any production code, just the mock object.

arvidn (commented):
This is a big enough change to be making against the `master` branch, rather than `RC_2_0`. Please rebase it.

Also, it seems to be a lot more complex that it would need to be. I'm not done reviewing all of it yet though.

src/curl_tracker_connection.cpp
```diff
@@ -0,0 +1,190 @@
...
+}
+
+void curl_tracker_connection::close()
+{
+	if (m_client)
+	{
+		m_client->close();
+	}
+	cancel();
+	m_man.remove_request(static_cast<aux::http_tracker_connection const*>(this));
```
arvidn, line 135: will this make the handler be called? If not, there won't be a way to synchronize the threads, is there?

src/session_impl.cpp
```diff
@@ -494,6 +494,9 @@ bool ssl_server_name_callback(ssl::stream_handle_type stream_handle, std::string
 		, m_download_rate(peer_connection::download_channel)
 		, m_upload_rate(peer_connection::upload_channel)
 		, m_host_resolver(m_io_context)
+#ifdef TORRENT_USE_LIBCURL
+		, m_curl_thread_manager(curl_thread_manager::create(m_io_context, m_settings))
```
arvidn, line 498: how come this isn't a member of `tracker_manager` instead?

src/curl_tracker_connection.cpp
```diff
@@ -0,0 +1,190 @@
...
+	io_context& ios
+	, aux::tracker_manager& man
+	, aux::tracker_request req
+	, std::weak_ptr<aux::request_callback> c
+	, std::shared_ptr<aux::curl_thread_manager> curl_mgr)
+	: aux::http_tracker_connection(ios, man, std::move(req), c)
+	, m_curl_thread_manager(std::move(curl_mgr))
+{
+	aux::session_settings const& session_sett = m_man.settings();
+	settings_pack sett;
```
arvidn, line 52: why are you making a copy of the settings here?

include/libtorrent/settings_pack.hpp
```diff
@@ -373,6 +373,15 @@ namespace aux {
...
 			webtorrent_stun_server,
 
+			// ``tracker_ca_certificate`` specifies the path to a custom CA
+			// certificate bundle file (in PEM format) to use for verifying
+			// HTTPS tracker connections. If empty (the default), the system's
+			// default CA bundle will be used. This is particularly useful for
+			// testing with self-signed certificates or when using a private
+			// certificate authority. The file must contain one or more CA
+			// certificates in PEM format. Only applies when using libcurl.
+			tracker_ca_certificate,
```
arvidn, line 383: Does this only work for curl trackers? If so, that should be documented. If not, this seems like a great independent feature, that can be made as a separate PR.

include/libtorrent/aux_/ssl_session_cache_interface.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <string>
+
+#ifdef TORRENT_USE_OPENSSL
+typedef struct ssl_session_st SSL_SESSION;
+#endif
+
+namespace libtorrent::aux {
+
+// Non-template interface for SSL session caching
+// This allows C callbacks to interact with the templated connection pool
```
arvidn, line 46: isn't curl managing the connection pool?

test/test_idle_cpu_bug.cpp
arvidn, line 1: what's the point of this test?

test/web_server.py
arvidn, line 1: does it really take this many more lines of code to test http/2?

include/libtorrent/settings_pack.hpp
```diff
@@ -1008,6 +1017,45 @@ namespace aux {
...
+
+			// ``proxy_force_internal_addresses`` controls whether internal/localhost
+			// addresses bypass the proxy (default secure behavior) or are forced
+			// through the proxy. When false (default), localhost and private IP
+			// ranges bypass the proxy for security (prevents SSRF attacks).
+			// When true, ALL addresses including localhost are forced through the
+			// proxy. WARNING: Only enable if you trust your proxy server with
+			// internal traffic. This setting only applies when using libcurl for
+			// tracker connections. Default: false.
+			proxy_force_internal_addresses,
```
arvidn, line 1057: does this only work for curl tracker requests? If so, that must be documented, if not, it seems like an independent change. I'm a bit skeptical to add something like this just to support tests though. It seems this feature itself would be hard to test.

src/curl_thread_manager.cpp
```diff
@@ -0,0 +1,1609 @@
...
+
+    // Wake up the thread if it's waiting (direct wakeup for immediate shutdown)
+    perform_wakeup();
+
+    // Wait for thread to finish
+    if (m_curl_thread.joinable()) {
+        m_curl_thread.join();
+    }
+
+    // Process any remaining queued requests with error
```
arvidn, line 529: is there a good reason not to do this in the cancelled thread itself? It seems like it would make it a bit simpler and avoid possible race conditions

src/curl_thread_manager.cpp
```diff
@@ -0,0 +1,1609 @@
...
+    // Cancel timer synchronously (safe because we're about to join the thread)
+    // The timer will be destroyed automatically when this object is destroyed
+    m_wakeup_timer.cancel(); // cancel() returns number of cancelled operations
+    m_timer_running = false;
+
+    // Wake up the thread if it's waiting (direct wakeup for immediate shutdown)
+    perform_wakeup();
+
+    // Wait for thread to finish
+    if (m_curl_thread.joinable()) {
```
arvidn, line 525: it seems problematic to sometimes *not* join this thread.

include/libtorrent/aux_/tracker_manager.hpp
```diff
@@ -392,6 +401,9 @@ using tracker_request_flags_t = flags::bitfield_flag<std::uint8_t, struct tracke
 		aux::resolver_interface& m_host_resolver;
 		aux::session_settings const& m_settings;
 		counters& m_stats_counters;
+#ifdef TORRENT_USE_LIBCURL
+		std::shared_ptr<aux::curl_thread_manager> m_curl_thread_manager;
```
arvidn, line 405: is it  really necessary for this to be held by a `shared_ptr`? Why can't this class be the sole owner?

src/curl_tracker_client.cpp
```diff
@@ -0,0 +1,435 @@
...
+		}
+
+		return resp;
+	}
+
+curl_tracker_client::curl_tracker_client(
+	io_context&,
+	std::string const& url,
+	settings_pack const& settings,
+	std::shared_ptr<curl_thread_manager> curl_mgr)
```
arvidn, line 276: passing in ownership of this object here creates a circular reference. I think this should be a raw reference

src/curl_tracker_client.cpp
```diff
@@ -0,0 +1,435 @@
...
+	return m_tracker_url + "?" + build_tracker_query(req, false);
+}
+
+std::string curl_tracker_client::build_scrape_url(tracker_request const& req) const
+{
+	std::string scrape_url = scrape_url_from_announce(m_tracker_url);
+	return scrape_url + "?" + build_tracker_query(req, true);
+}
+
+std::string curl_tracker_client::build_tracker_query(tracker_request const& req, bool scrape) const
```
arvidn, line 349: this looks copy-pasted. it would be better to factor it out into a separate function

src/curl_tracker_client.cpp
```diff
@@ -0,0 +1,435 @@
...
+	if (!req.ipv6.empty()) {
+		query += "&ipv6=";
+		query += libtorrent::escape_string(req.ipv6[0].to_string());
+	}
+
+	return query;
+}
+
+
+std::string curl_tracker_client::scrape_url_from_announce(std::string const& announce) const
```
arvidn, line 412: this also looks like it should be a free function

src/session_impl.cpp
```diff
@@ -1079,6 +1085,21 @@ bool ssl_server_name_callback(ssl::stream_handle_type stream_handle, std::string
 #endif
 		m_tracker_manager.stop();
 
+#ifdef TORRENT_USE_LIBCURL
+		// CRITICAL FIX: Explicitly shut down curl thread manager
+		// This ensures the worker thread is stopped before any member destruction begins,
+		// preventing dangling reference to m_settings
+		// C++17 style with if-initializer for cleaner scope management
+		if (auto manager = std::move(m_curl_thread_manager); manager)
```
arvidn, line 1093:
```suggestion
		if (auto manager = std::move(m_curl_thread_manager))
```

This has the same behavior, right?

src/session_impl.cpp
```diff
@@ -1079,6 +1085,21 @@ bool ssl_server_name_callback(ssl::stream_handle_type stream_handle, std::string
...
+		{
+#ifndef TORRENT_DISABLE_LOGGING
+			session_log(" shutting down curl thread manager");
+#endif
+			manager->shutdown();
```
arvidn, line 1098: `session_impl` separates initiating shutdown from waiting for shutdown. I believe `manager->shutdown()` does both, right?

arvidn (commented):
I think it would be easier to review and land this in smaller pieces

TeholTheOnly:
> I think it would be easier to review and land this in smaller pieces

Would you prefer me to go through and address the comments that you've made on this PR, or close this one out and start the process of making smaller PRs to land the main functionality of this and then slowly expand out? Just want to be mindful of your time, and I appreciate you taking a look!

include/libtorrent/aux_/curl_handle_wrappers.hpp
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+private:
+    CURLM* m_handle;
+};
+
+// Deleter for unique_ptr with custom cleanup
+struct curl_easy_deleter {
+    void operator()(CURL* handle) const noexcept {
+        if (handle) curl_easy_cleanup(handle);
```
loskutov, line 183: `unique_ptr` doesn't invoke the deleter for `nullptr` anyways

seabashed:
I'm really excited about this pull request, but noticed progress seems to have stalled. Is there any work happening behind the scenes that isn't visible in the repo?

FYI: rTorrent (and its libtorrent) recently adopted libcurl's modern connection reuse practices, which significantly reduces bandwidth usage. It would be fantastic to see similar improvements implemented here.

I wish I had the technical skills to contribute directly, but unfortunately that's beyond my current abilities. Still very much hoping to see this move forward!

PR #8026, opened 2025-09-06
allow bypassing announce queue

HamletDuFromage:
When adding a torrent to a client with a lot of torrents, its announce (as far as I understand) is pushed to the end of a queue that could take a while to be processed.

As a workaround if I need to leech immediately, I'd just restart my client and the newer torrents would be added at the top of the queue, but it's rather inelegant. 

I'd like a flag for force re-announce to bypass that queue for the torrent to start faster. This PR compiles, but I'm not familiar with the codebase whatsoever so this might not be the appropriate approach. For instance, I'm not sure about how this may or may not be related to https://www.libtorrent.org/reference-Torrent_Handle.html#queue_position_top()

This is more of a feature request than a PR

### DIFF
--- a/include/libtorrent/torrent_handle.hpp
+++ b/include/libtorrent/torrent_handle.hpp
@@ -1165,6 +1165,11 @@ namespace aux {
 		// published by the tracker. If this flag is set, it will be ignored
 		// and the tracker is announced immediately.
 		static constexpr reannounce_flags_t ignore_min_interval = 0_bit;
+		
+		// by default, force-reannounce will queue the announce normally.
+		// If this flag is set, the announce will be put at the front of the
+		// tracker queue for immediate processing.
+		static constexpr reannounce_flags_t priority_queue = 1_bit;
 
 		// ``force_reannounce()`` will force this torrent to do another tracker
 		// request, to receive new peers. The ``seconds`` argument specifies how
@@ -1179,7 +1184,7 @@ namespace aux {
 		// If set to -1 (which is the default), all trackers are re-announce.
 		//
 		// The ``flags`` argument can be used to affect the re-announce. See
-		// ignore_min_interval.
+		// ignore_min_interval and priority_queue.
 		//
 		// ``force_dht_announce`` will announce the torrent to the DHT
 		// immediately.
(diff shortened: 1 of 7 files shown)


src/tracker_manager.cpp
```diff
@@ -291,7 +292,7 @@ constexpr tracker_request_flags_t tracker_request::i2p;
 #endif
 		{
 			auto con = std::make_shared<http_tracker_connection>(ios, *this, std::move(req), c);
-			if (m_http_conns.size() < std::size_t(sett.get_int(settings_pack::max_concurrent_http_announces)))
+			if (priority || m_http_conns.size() < std::size_t(sett.get_int(settings_pack::max_concurrent_http_announces)))
```
arvidn: this makes me feel a bit uneasy. It effectively removes this cap. How about just using `push_front()` if `priority` is set instead?

include/libtorrent/torrent_handle.hpp
```diff
@@ -1165,6 +1165,11 @@ namespace aux {
 		// published by the tracker. If this flag is set, it will be ignored
 		// and the tracker is announced immediately.
 		static constexpr reannounce_flags_t ignore_min_interval = 0_bit;
+		
+		// by default, force-reannounce will queue the announce normally.
+		// If this flag is set, the announce will be put at the front of the
+		// tracker queue for immediate processing.
+		static constexpr reannounce_flags_t priority_queue = 1_bit;
```
arvidn:
```suggestion
		static constexpr reannounce_flags_t high_priority = 1_bit;
```

include/libtorrent/torrent_handle.hpp
```diff
@@ -1179,7 +1184,7 @@ namespace aux {
 		// If set to -1 (which is the default), all trackers are re-announce.
 		//
 		// The ``flags`` argument can be used to affect the re-announce. See
-		// ignore_min_interval.
+		// ignore_min_interval and priority_queue.
```
arvidn:
```suggestion
		// ignore_min_interval and high_priority.
```

arvidn:
it should probably also be incorporated with "connection boost", so newly added torrents take priority over existing ones, for the first announce

HamletDuFromage:
I don't think I understand what connection boost is or does tbh

arvidn:
I made a slightly tweaked version of this idea, here: https://github.com/arvidn/libtorrent/pull/8054

HamletDuFromage:
neato!

arvidn:
https://github.com/arvidn/libtorrent/pull/8054 has landed now.

PR #8030, opened 2025-09-08
don't link against libboost_system explicitly

arvidn:
It's deprecated. It's a header-only library now.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -17,7 +17,7 @@ jobs:
 
    test:
       name: Tests
-      runs-on: macos-latest
+      runs-on: macos-14
       continue-on-error: true
 
       strategy:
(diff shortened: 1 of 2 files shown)


xavier2k6:
@arvidn What was the reason for going back to `macos-14`?

arvidn:
to avoid having CI (potentially) break when there's a new MacOS image

xavier2k6:
Only other image is `macos-26` (Tahoe) & it'll be a while before it's migrated to `macos-latest` label.

It might be no harm to create a "dry-run" with the `macos-26` runner image to see if there would be any upcoming breakages.....

same, could be done when a `ubuntu-26.04` runner image becomes available.

Also note, that Visual Studio 2026 will probably be coming to the `windows-2025/windows-latest` runner image at a later date, `windows-2022` will be sticking with Visual Studio 2022.

PR #8039, opened 2025-09-22
Add piece cache feature for fileless torrent operation

ismdevteam:
- Add PieceCacheManager for disk-based piece caching
- Add SimplePieceCache for libtorrent integration
- Enhance client_test with -Z fileless mode
- Add documentation and examples
- Support both torrent files and magnet links


USBhost:
Is the point of this to help seed if you do not care about the content?

ismdevteam:
> Is the point of this to help seed if you do not care about the content?

This pull request implements functionality similar to what qBittorrent offers with its "SQLite database (experimental)" resume data storage option, but directly within libtorrent itself and without the SQLite dependency.

**What it does:**

qBittorrent has an advanced setting where you can choose:
- **Fastresume files** (traditional `.fastresume` files per torrent)
- **SQLite database** (stores all torrent metadata in a single SQLite database)

**This PR implements a third approach within libtorrent:**
- **Piece cache storage** - stores both resume data AND torrent pieces as blob files in a cache directory
- **All done natively in libtorrent** without external database dependencies

**The main purpose:** Enable efficient seeding when you don't want to keep the actual content files - perfect for pure seeding scenarios or temporary downloads.

**How it works (vs. SQLite approach):**
Instead of using SQLite's structured database tables, this implements a simpler, native libtorrent solution:
- Resume metadata stored alongside cached piece blobs in regular files
- No SQLite dependency or overhead
- Direct piece access without database queries
- Enables fileless seeding from cache-only storage

arvidn:
this appears to be a feature of the `client_test` *example* only. It's not adding a feature to libtorrent, only to the example. It's not necessarily a problem, but it seems like quite a lot of code to be an *example* of how to use the disk io interface.

I think the main problem is that you clone the `client_test.cpp` rather than introducing this feature as an option to the existing program. Is there a reason for that? There's quite a lot of duplicated code in your PR.

Additionally, by introducing a new test, without updating the CI workflows, there's no CI coverage of this new code at all. Not building it, let alone running it.

Can you think of a way to test this feature on CI, to demonstrate it working?

arvidn:
this smells like AI slop

ismdevteam:
> this smells like AI slop

I mistakenly pushed several commits to the branch that was already pull requested. I apologize for any confusion this caused.

PR #8045, opened 2025-10-19
Python bindings: Add Python 3.14 to supported versions

xavier2k6:


chapmanjacobd:
I can confirm that the features that I use work with python3.14 on Fedora 43 x86_64 when built likeso:

```
sudo dnf install -y cmake gcc-c++ python3.14-devel boost-devel boost-python3-devel boost-system boost-build openssl-devel pkgconfig ninja

git clone --recursive https://github.com/arvidn/libtorrent
cd libtorrent
git checkout v2.0.11
mkdir build; cd build
cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_STANDARD=14 -Dpython-bindings=3.14 -Dpython-install-system-dir=1 -G Ninja ..
ninja
sudo ninja install
```

xavier2k6:
Might need to change python to 3.14 in ci.

xavier2k6:
Python 3.9 is EOL, should it be dropped here?

arvidn:
I'll incorporate this change in https://github.com/arvidn/libtorrent/pull/8075

xavier2k6:
Might bump below too, unless you want to have this done when `cibuildwheel` changes are made?!
https://github.com/arvidn/libtorrent/blob/262028c09d84ddab0aef4f4717c1fd113dc30f04/pyproject.toml#L3

arvidn:
apparently `windows-2025` has python 3.9 installed.

xavier2k6:
> apparently `windows-2025` has python 3.9 installed.

not for much longer, https://github.com/actions/runner-images/issues/13468

It is still available however in `actions/setup-python`
https://github.com/actions/python-versions/blob/main/versions-manifest.json

xavier2k6:
You can keep Python 3.9 support as `cibuildwheel` still supports it.

chapmanjacobd:
Just FYI, looks like PyPI hasn't been updated for a while

https://pypi.org/project/libtorrent/#history

on python3.14 it looks like this:

```sh
pip install libtorrent
Defaulting to user installation because normal site-packages is not writeable
ERROR: Could not find a version that satisfies the requirement libtorrent (from versions: none)
ERROR: No matching distribution found for libtorrent
```

edit:

ah... I just realized most people probably wouldn't use pip anyway as it should be packaged along with a system package manager to match the ABI version. sorry for the noise

edit2:

wait but it doesn't include python bindings by default so actually my loudness is perhaps warranted:

```sh
rpmquery -ql rb_libtorrent
/usr/lib/.build-id
/usr/lib/.build-id/8e
/usr/lib/.build-id/8e/216b1537e761d133dcabd63374e65e72b95d4f
/usr/lib64/libtorrent-rasterbar.so.2.0
/usr/lib64/libtorrent-rasterbar.so.2.0.11
/usr/share/doc/rb_libtorrent
/usr/share/doc/rb_libtorrent/AUTHORS
/usr/share/doc/rb_libtorrent/ChangeLog
/usr/share/licenses/rb_libtorrent
/usr/share/licenses/rb_libtorrent/COPYING
```

edit3: oh wait there is `rb_libtorrent-python3`

PR #8046, opened 2025-10-19
Update pre-commit-config workflow

xavier2k6:
* Bumped numerous hook revisions.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ default_language_version:
     python: python3
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v5.0.0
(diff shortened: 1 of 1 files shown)


arvidn:
these checks should not apply to `.torrent` files. I imagine it would be best to have a whitelist of file extensions to apply it to. something like: `.cpp`, `.py`, `.hpp`, `Jamfile`, `.txt`, `.rst`. Except, `test/utf8_test.txt` should probably be left as-is. It's kind of a test vector.

xavier2k6:
>* Enabled `trailing-whitespace` & `end-of-file-fixer` hooks

I'll disable these again & will leave it to you to handle these at a later stage, will bump to newer hooks on next push.

arvidn (approved):

PR #8074, opened 2025-12-26
go back to windows-2022 runners, that have openssl installed

arvidn:


xavier2k6:
OpenSSL (3.6.0 currently) is already on the runner images
https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#tools
https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md#tools

xavier2k6:
Use Boost 1.90.0?

PR #8076, opened 2025-12-27
search Program Files/OpenSSL/ for OpenSSL on windows runners

arvidn:


arvidn:
@xavier2k6 openssl appears to be installed in a new path

xavier2k6:
might've been https://github.com/actions/runner-images/pull/13203

PR #8082, opened 2025-12-28
add debug logging to windows python CI workflow

arvidn:


arvidn:
there's so many layers of python abstraction to pierce through. This used to be as simple as putting `libtorrent.pyd` in the current working directory and launch python.

qstokkink:
I'm going through some of `libtorrent`'s PRs to see how to best optimize the Python build process. In this case, the issue is:

```
LINK : fatal error LNK1181: cannot open input file 'libcrypto.lib'
```

OpenSSL again 👿. I've also seen (a few 100 PRs ago somewhere) that that there have been issues with finding DLLs. This really seems to be a thorn in the side for development and i.m.o. it could use a dedicated check.

For debugging purposes, it would be nice to have _small and fast-compiling_ b2 build target that only does some quick OpenSSL calls. From the Python side, we can then call this and see if it works. Both of these things can then be put in a separate GitHub Action workflow.

If all the other GitHub Actions then wait for this quick debug call/Action workflow to succeed, PRs that suffer from OpenSSL linking/path issues can be hopefully fixed in a matter of seconds or minutes, instead of the full 1.5 hours it now takes for a PR CI run.

PR #8083, opened 2025-12-29
docs: Use `ubuntu-slim` runner image

xavier2k6:
* Use `ubuntu-slim` runner image for **docs** workflow

From https://docs.github.com/en/actions/reference/runners/github-hosted-runners#single-cpu-runners :
>This type of runner is optimized for automation tasks, issue operations and short-running jobs. They are not suitable for typical heavyweight CI/CD builds.

* https://github.com/actions/runner-images?tab=readme-ov-file#available-images
* https://github.com/actions/runner-images/blob/main/images/ubuntu-slim/ubuntu-slim-Readme.md
* https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -13,7 +13,7 @@ jobs:
 
    docs:
       name: Docs
-      runs-on: ubuntu-24.04
+      runs-on: ubuntu-slim
 
       steps:
       - name: checkout


arvidn:
the failure is odd. "Umeå" is present in the `libtorrent.dic` file. Perhaps there's some UTF-8 feature omitted in the `slim` runner.

xavier2k6:
> the failure is odd. "Umeå" is present in the `libtorrent.dic` file. Perhaps there's some UTF-8 feature omitted in the `slim` runner.

This runner image is new enough, so may need to hold off on using it in `this repo` until more details can be provided in relation to installed software/dependencies & limitations..

xavier2k6:
`bsdextrautils` is currently not included in the `ubuntu-slim` image, unsure if it's related.

PR #8027, opened 2025-09-07
WIP - boost-1.87.0 api changes for RC_1_2 into RC_1_3

userdocs:
**Disclaimer:** These changes are made with me spamming claude.ai cli with prompts and build errors until `RC_1_2` successfully built against boost-1.89.0 - I have no idea what I'm doing with this codebase, all changes need thorough review and testing. Maybe it's good enough of a start to get `RC_1_3` going.

To build on this issue https://github.com/arvidn/libtorrent/issues/7796 and specifically @arvidn comment https://github.com/arvidn/libtorrent/issues/7796#issuecomment-2613849165

I used claude.ai cli (pro plan) to brute force these changes to allow me to successfully `build RC_1_2` against `boost-1.89.0` with `cmake` and `b2` with `qbittorrent-nox` 

So it works, in basic testing building against

Build tool tested:

- [x] cmake
- [x] b2 
 
OS tested :

- [x] Linux
- [ ] Windows
- [x] OSX

> [!NOTE]
> So I don't really know if the ai has done a good job, I just understand the outcome and that it works, I'm not going debate the merits of using it or not in this thread. I only care about getting us past the post so that @arvidn can do minor changes to get it into RC_1_3 and provide boost-1.87.0 support to libtorrent v1.3

Issue #7796, opened 2024-11-16
boost-1.87.0.beta1 libtorrent RC_1_2


userdocs:
Build error 

cmake
boost-1.87.0.beta1
libtorrent RC_1_2 

```
-- Configuring done (0.8s)
-- Generating done (0.0s)
Generate graphviz: /root/qbt-build/graphs/libtorrent/RC_1_2/dep-graph.dot
-- Build files have been written to: /root/qbt-build/libtorrent/build
[1/156] Building CXX object CMakeFiles/torrent-rasterbar.dir/src/assert.cpp.o
[2/156] Building CXX object CMakeFiles/torrent-rasterbar.dir/src/bandwidth_limit.cpp.o
[3/156] Building CXX object CMakeFiles/torrent-rasterbar.dir/src/bandwidth_queue_entry.cpp.o
[4/156] Building CXX object CMakeFiles/torrent-rasterbar.dir/src/bandwidth_manager.cpp.o
[5/156] Building CXX object CMakeFiles/torrent-rasterbar.dir/src/bitfield.cpp.o
[6/156] Building CXX object CMakeFiles/torrent-rasterbar.dir/src/bdecode.cpp.o
[7/156] Building CXX object CMakeFiles/torrent-rasterbar.dir/src/alert.cpp.o
FAILED: CMakeFiles/torrent-rasterbar.dir/src/alert.cpp.o 
[...]

PR #8085, opened 2025-12-29
Use Boost 1.90.0 in Windows CI

xavier2k6:
https://www.boost.org/releases/latest/

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -59,7 +59,7 @@ jobs:
       if: runner.os == 'Windows'
       shell: cmd
       run: |
-        git clone --depth=1 --recurse-submodules -j10 --branch=boost-1.88.0 https://github.com/boostorg/boost.git
+        git clone --depth=1 --recurse-submodules -j10 --branch=boost-1.90.0 https://github.com/boostorg/boost.git
         cd boost
         bootstrap.bat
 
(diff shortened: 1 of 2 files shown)


Neustradamus:
Linked to:
- https://github.com/arvidn/libtorrent/pull/8027
- https://github.com/arvidn/libtorrent/issues/7796

xavier2k6:
https://github.com/arvidn/libtorrent/blob/f59119648138d08c00de86a06bfd4c68933c5fe0/.github/workflows/windows.yml#L132

@arvidn Should above be changed to target Windows10 here since we target (Windows 10 by default for some time now) or another config option be added for Windows 10 or even Windows 11?

```
             - config: asio-debugging=on picker-debugging=on windows-version=vista
             - config: asio-debugging=on picker-debugging=on windows-version=win10
             - config: asio-debugging=on picker-debugging=on windows-version=win11
             - config: windows-api=store windows-version=win10
             - config: windows-api=store windows-version=win11
```

arvidn:
I think the idea of targeting an old version of windows is to ensure no (mandatory) dependency on a newer version is added. But it's probably a good idea to build and test against the latest version as well.

arvidn (approved):

xavier2k6:
>I think the idea of targeting an old version of windows is to ensure no (mandatory) dependency on a newer version is added.

Going back to `xp` is a no go anyway
https://github.com/arvidn/libtorrent/actions/runs/20619365589/job/59218180043

>But it's probably a good idea to build and test against the latest version as well.

I added Windows10, Windows 11 as well as other versions in #8092 for testing....you can have a look.

I can add `win10` to the config.

PR #8091, opened 2025-12-30
Testing matrix with `macos-26` (Tahoe) and `macos-15-intel`

xavier2k6:


xavier2k6:
@arvidn the GHA Runners have now moved to python 3.12 by default & dropped 3.9, this has changed the previous python bindings failure to below:

```
  set BOOST_ROOT=%CD%\boost
  set BOOST_BUILD_PATH=%BOOST_ROOT%\tools\build
  set PATH=%BOOST_ROOT%;%PATH%
  
  cd bindings\python
  python setup.py build_ext --b2-args "asserts=on invariant-checks=full" install --user --prefix=
  shell: C:\Windows\system32\cmd.EXE /D /E:ON /V:OFF /S /C "CALL "{0}""
  env:
    HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: 1
    HOMEBREW_NO_AUTO_UPDATE: 1
  File "D:\a\libtorrent\libtorrent\bindings\python\setup.py", line 26, in <module>
    import setuptools
ModuleNotFoundError: No module named 'setuptools'
Error: Process completed with exit code 1.
```

xavier2k6:
I'm testing `macos-26` (Tahoe) in CI as `macos-14` will be dropped & that will require going to `macos-15` or `macos-latest` the latter will roll-over to `macos-26` (Tahoe) further down the road.

arvidn:
I have a feeling the way the python bindings are built will need to change soon.
It's a bit frustrating that it's so complicated to turn a native python module (the output of `b2`) to a wheel or something that can be published.

xavier2k6:
@arvidn There are issues with the use of newer macos....which will have to be handled.......macos-14 will be deprecated/removed as a runner image in a few months.

PR #8092, opened 2025-12-30
Testing new windows config setup

xavier2k6:


xavier2k6:
@arvidn Are the `sanitizer` failures genuine?, can you handle them if they are?

arvidn:
> @arvidn Are the `sanitizer` failures genuine?, can you handle them if they are?

These crashes look more like symptoms of some issue in the build or linking. e.g. if the address sanitizer runtime library can't be found, or something. I would expect msvc's address sanitizer to also print helpful output if there's an actual sanitizer failure.

xavier2k6:
Might be time to update the copyright year, last update was 2022 via #6931

xavier2k6:
b2 5.4.2+ fixes autodetection of vs2026

https://github.com/bfgroup/b2/releases

PR #8103, opened 2026-01-03
add new setting disk_disable_copy_on_write

arvidn:
to set `FS_NOCOW_FL` on new files. This is enabled by default, Linux only and is expected to reduce fragmentation on btrfs filesystems.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,6 @@
 2.0.12
 
+	* add setting to set no-copy-on-write flag on new files
 	* add performance counters to file pool
 	* add high_priority flag to torrent_handle::force_reannounce()
 	* update default DSCP value and update docs
--- a/docs/hunspell/libtorrent.dic
+++ b/docs/hunspell/libtorrent.dic
@@ -637,3 +637,5 @@ WebDAV
 transmissionbt
 ouinet
 IANA's
+NOCOW
+btrfs
--- a/include/libtorrent/aux_/open_mode.hpp
+++ b/include/libtorrent/aux_/open_mode.hpp
@@ -54,6 +54,9 @@ namespace aux {
 		constexpr open_mode_t executable = 7_bit;
 		constexpr open_mode_t allow_set_file_valid_data = 8_bit;
 		constexpr open_mode_t no_mmap = 9_bit;
+		// sets the FS_NOCOW_FL flag on the file, when creating it.
+		// This is currently linux specific, for btrfs filesystems
+		constexpr open_mode_t no_cow = 10_bit;
 	}
 } // aux
 
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -1026,6 +1026,11 @@ namespace aux {
 			// man-in-the-middle connections.
 			proxy_send_host_in_connect,
 
+			// When set, downloaded files will have the no-copy-on-write flag
+			// (``FS_NOCOW_FL``) set on Linux. This mitigates heavy
+			// fragmentation on filesystems like btrfs.
+			disk_disable_copy_on_write,
+
 			max_bool_setting_internal
 		};
 
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -105,6 +105,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #ifdef TORRENT_LINUX
 // linux specifics
 
+#include <linux/fs.h>
 #include <sys/ioctl.h>
 #ifdef TORRENT_ANDROID
 #include <sys/syscall.h>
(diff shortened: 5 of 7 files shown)


database64128:
Sorry if I'm a bit late here. This new setting seems nice to have, but I'm not sure if having it enabled by default is a good idea. Personally I find this to be a breaking change for my use cases.

I run qBittorrent with libtorrent v2 on multiple Linux machines, all with btrfs as the underlying filesystem. I make extensive use of reflinks to bridge the gap between the same torrent content from different sources, so that I can seed all of them simultaneously without them taking up extra space. I had to use reflinks to copy some torrented files because I wanted to share them via SMB but ksmbd did not support symlinks.

I understand that it's still possible to use reflinks with `NOCOW` files, but this requires significant effort from the user to convert existing directories, which I personally am not interested in doing.

I also have zstd compression enabled on all btrfs filesystems. While most of my torrented files won't benefit from compression at all, I do expect the compressible ones to be properly compressed upon downloading. As far as I know, `NOCOW` files are not eligible for compression on btrfs.

I have a habit of running `btrfs -v filesystem defrag -rt 4G` on newly downloaded files that I intend to keep for a long period of time (maybe forever). But I never run defragmentation on play-once-and-delete type of files, which is most of them. Fragmentation with CoW is actually not that bad. On my 7200rpm HDDs, defragmenting torrented files increases the sequential read speed from ~200 MB/s to ~300 MB/s. My conclusion is that, unless you do "force rechecks" (in qBittorrent's terms) all the time, fragmentation under normal use with CoW doesn't matter at all.

tsweet64:
I agree that this should not be default. It would be a workflow-breaker for me too due to my use of reflinks. But even for those for whom that doesn't matter (i.e. average users), nocow eliminates some of the fundamental correctness-guarantees of btrfs in cases of power loss and disables data checksumming. These features are whole point of me using btrfs on my media drives. and I expect that (unless I specifically say otherwise) all non-temporary data I write on my disk would be CoW and summed. I work around the fragmentation issue by downloading to an ssd and then having qbittorrent move the files to hdd upon completion. But even still, in my testing recently the fragmentation issue with torrents and CoW on btrfs, while certainly worth some concern, is nowhere near as severe as it used to be a few years ago.

arvidn:
@database64128 I'm not an expert on btrfs. Originally, I was under the impression of CoW being used *only* for snapshots and reflink copies of files. In this case, you would primarily get fragmentation when you update a file that's part of a snapshot (which presumably would be relatively short-lived, during a back-up for example), or when you copy a file with reflink and then update one of the files. This seems like a very reasonable use of CoW.

However, what I've learned is that btrfs makes *every* update to a file (or at least nearly every) atomic, by making a copy of a block, mutating it and then linking it into the file. This form of CoW has a slightly different use case. It enables reliable file content at various kinds of outages for instance. This latter kind of CoW is what some people find problematic with torrents (or any file with random writes, presumably). Over time a file with random updates would be maximally fragmented.

Setting the `FL_NOCOW_FL` on a file, I would have thought it only disables the second use of CoW, but would still allow files to be part of snapshots or reflink-copied.

I understand your comment as the first kind of CoW is also disabled for files with this flag, is that right?

database64128:
> I understand your comment as the first kind of CoW is also disabled for files with this flag, is that right?

Yes, but not exactly. [Snapshotting still works on NOCOW files.](https://linux-btrfs.vger.kernel.narkive.com/7diJiBfC/are-nocow-files-snapshot-aware) You can still reflink NOCOW files, but [only to NOCOW directories](https://btrfs.readthedocs.io/en/latest/Reflink.html). Converting existing directories to NOCOW is an unrealistic ask, and we wouldn’t want to do that, because you lose checksumming and compression when you disable CoW. Both are important features we want out of the filesystem.

HanabishiRecca:
> However, what I've learned is that btrfs makes _every_ update to a file (or at least nearly every) atomic, by making a copy of a block, mutating it and then linking it into the file.

Yes, that's the whole point of CoW file systems actually. They all work that way: data is never overwritten/updated in-place. The goal is to guarantee a consistent state at any moment.

You totally can see how that does not align well with libtorrent's random 16K writes. Dirty pages help somewhat, but result is not guaranteed and depends on many conditions. Also btrfs actually cheats a little in default configuration, a really consistent fs state is achieved by enabling the `flushoncommit` mount flag, which in turn makes fragmentation even worse.

The real solution here would be to keep incomplete torrent pieces fully in RAM and only flush them to the disk in one write once they are fully complete / have the checksum verified. It has tradeoffs of course, significantly increased RAM usage at least, and most importantly will require fundamental code rewrite.

So as for now, I'm on the fence in this discussion. On one hand, some people want CoW regardless and setting NOCOW on your downloads dir when you want is not hard anyway. On the other hand, many people are simply not aware of such details of their file system operation.

arvidn:
> Yes, that's the whole point of CoW file systems actually. They all work that way: data is never overwritten/updated in-place. The goal is to guarantee a consistent state at any moment.

So people, presumably, have the same problem with APFS and ZFS too then, right?

HanabishiRecca:
Probably, as they all share the core idea https://en.wikipedia.org/wiki/Copy-on-write#In_computer_storage
But different file systems obviously have different underlying implementations, so it's hard to predict without direct testing.
E.g. ZFS has own aggressive memory caching on top of dirty pages, afaik. Some other file systems like Bcachefs have background garbage collection that reassembles data blocks on the fly etc.

PR #7971, opened 2025-06-28
update manylinux docker scripts

arvidn:
Updated version of https://github.com/arvidn/libtorrent/pull/7824

PR #8112, opened 2026-01-11
update cibuildwheel to build against python 3.14

arvidn:


xavier2k6:
@arvidn current cibuildwheel action doesn't have support for python 3.14, Can you take this opportunity to update the manylinux docker scripts again. xref: #7971

I'll submit a PR for the actions / invalid skip step & updated runners etc.

qstokkink:
> @arvidn current cibuildwheel action doesn't have support for python 3.14

Updating from `pypa/cibuildwheel@v2.22.0` to `pypa/cibuildwheel@v3.4.0` (`v3.3.1` or later) should fix this now, right?

xavier2k6:
@qstokkink ya I have a PR open but I believe the manylinux scripts may need to be updated again & i'm trying to add `windows-11-arm`

PR #8113, opened 2026-01-11
cibuildwheel: refactor workflow(s)

xavier2k6:


qstokkink:
@xavier2k6 I've been looking into OpenSSL build issues and this seems like another instance of the system install (for your new build targets) missing header files. Could you try commenting out the following lines?

https://github.com/arvidn/libtorrent/blob/061d430a81766167e6c4c6dda0084d114818c21f/tools/cibuildwheel/manylinux/build-openssl.sh#L28-L31

Maybe it's too soon for libtorrent to handle these new build targets. Short term, just updating `cibuildwheel` to `3.4.0` - without any new targets - might be a good idea. Just to get rid of all the annoying `urllib.error.HTTPError: HTTP Error 429: Too Many Requests` errors.

xavier2k6:
* The `manylinux` failures are due to the scripts needing to be synced with upstream again I believe even if I lower `cibuildwheel` version, they still failed previously even without adding in the new `22.04/22.04 arm` targets.

* `winows-11-arm` target can be removed & looked at/added after a `2.0.12` release, if issues can't be investigated/resolved currently.

qstokkink:
I doubt syncing will resolve the errors on this PR. I would not expect errors like `../../include/libtorrent/ssl.hpp:56:10: fatal error: openssl/opensslv.h: No such file or directory` to be solved by any change in `cibuildwheel` Action version. That's why I requested commenting out those particular lines.

I know how to fix the `winows-11-arm` target, but given the resolution, I'd advise not trying to do that without a dedicated OpenSSL setup action. That will likely turn into a big mess.

xavier2k6:
> I doubt syncing will resolve the errors on this PR. I would not expect errors like `../../include/libtorrent/ssl.hpp:56:10: fatal error: openssl/opensslv.h: No such file or directory` to be solved by any change in `cibuildwheel` Action version. That's why I requested commenting out those particular lines.

I'm on about syncing these upstream scripts:

https://github.com/pypa/manylinux/blob/main/docker/build_scripts/build_utils.sh

https://github.com/pypa/manylinux/blob/main/docker/build_scripts/build-openssl.sh

We updated these before due to `perl-core` issue but also had to provide a workaround, should we have to keep providing workarounds if a fix can be provided `upstream` to alleviate any other potential issues.

> I know how to fix the `winows-11-arm` target, but given the resolution, I'd advise not trying to do that without a dedicated OpenSSL setup action. That will likely turn into a big mess.

That target can be removed for now so.

qstokkink:
🤔 Those scripts don't look too different. That said, the core of the issue is that the Dockerized OSes of `manylinux` do ship OpenSSL but don't ship the (install_dev/install_sw) headers. Then `build-openssl.sh` doesn't build the version with the headers because it already found OpenSSL installed. I'll leave it there and make it your problem 🙂

xavier2k6:
hmmm....an official `window-11-arm64` runner image may be incoming, this can maybe be used instead of having to use the GitHub partner `window-11-arm` provided image.

PR #8114, opened 2026-01-11
when building against OpenSSL, require at least version 1.1.1

arvidn:


xavier2k6:
Boost 1.85.0 is minimum available version on 2022 image, while 1.89.0 is currently maximum.

https://www.appveyor.com/docs/windows-images-software/#boost

xavier2k6:
@arvidn  specifying the OpenSSL version via find_package was introduced in CMake 3.20, correct?! so would that mean that you'd also have to bump `cmake_minimum_required` in code?

* https://github.com/search?q=repo%3Aarvidn%2Flibtorrent%20cmake_minimum_required&type=code
* https://cmake.org/cmake/help/latest/module/FindOpenSSL.html

arvidn:
I see. and the appveyor image only has cmake 3.16.2. I may have to bump the image after all

xavier2k6:
@arvidn How come you don't use something like `set(minOpenSSLVersion 1.1.1)` in `CMakeLists.txt`?

Below could be discarded then couldn't it?
```
#if OPENSSL_VERSION_NUMBER < 0x1010100fL
#error OpenSSL too old, libtorrent requires least OpenSSL 1.1.1 or later
```

I got a bit further with `AppVeyor` config & OpenSSL 3
https://ci.appveyor.com/project/arvidn/libtorrent/builds/53357292

xavier2k6:
3x Python38 left to be changed....

xavier2k6:
@arvidn might be no harm to resolve below "TODO's" & bump `CMake` to `3.21.0`, add note in changelog.

https://github.com/arvidn/libtorrent/blob/4479071bf5d202b957f526c692573e5692fdf3d5/CMakeLists.txt#L661

https://github.com/arvidn/libtorrent/blob/4479071bf5d202b957f526c692573e5692fdf3d5/CMakeLists.txt#L736

arvidn:
@xavier2k6 I think that's big enough change to warrant a second PR

PR #8116, opened 2026-01-11
fix no-copy-on-write flag setting

arvidn:


HanabishiRecca:
Just in case, I have double checked that the patch works.

PR #8117, opened 2026-01-11
when building against OpenSSL, require at least version 3.0.0

xavier2k6:


xavier2k6:
Note to self: `Boost` needs to be updated in `android` for `OpenSSL 3.x` compatability & more than likely the same is required for `CIFuzz` workflow

xavier2k6:
Do we need to make use of `#define OPENSSL_API_COMPAT=0x30000000L` / `#define OPENSSL_API_COMPAT 30000` ?

Issue #7194, opened 2022-11-29
Location of *.parts files


SpL1Ne:
https://github.com/qbittorrent/qBittorrent/issues/13205
As I can see in [this comment](https://github.com/qbittorrent/qBittorrent/issues/13205#issuecomment-665686667) there is no way to define folder for parts file. Also I searched and didn't find issue about this question.
So, is there in plans to provide possibility to change location of *.parts file (instead near downloaded torrent files)?

libtorrent version (or branch):
qBittorrent 4.4.5

platform/architecture:
Windows 10x64 21H1

master255:
What is it for? A cosmetic effect? Or functional?

SpL1Ne:
Well, users literally has some hidden files inside download folder. I'm usually move it whole and capture batch of 100-1000mb file that don't need to. And regular user doesn't even know about them.

So yeah definitely functional. Keep regular files at one location and qbittorrent files at other and don't mix them like don't mix temp files and user's files in any other program.

master255:
@SpL1Ne 

> Well, users literally has some hidden files inside download folder. I'm usually move it whole and capture batch of 100-1000mb file that don't need to. And regular user doesn't even know about them.

I don't understand what you're writing about. Can you explain it in more detail?

SpL1Ne:
@master255 
1. Let's say you have like 5 torrents with same theme\subject. For example, pictures, because they have more parts to complete pieces, if you don't download whole torrent.
2. Since all of them have same theme, you create folder to store them together. In my case fore example `Pictures` folder inside `Download` folder.
3. You load torrent file into qbittorrent. Select let's say 1/5 of total files. Because you don't need whole torrent.
4. Do that with other 4 torrents.
5. Set for all of them location as `Pictures` folder inside `Download`
6. Wait torrents to complete.
7. Now, you have your `Pictures` folder with torrents what you want.
8. You don't store your media inside `Download` folder, so you select `Pictures` folder ctrl-x and ctrl-v inside `D:\Media`.
9. Seems like everything good, right?
10. Wrong.
![2023-03-19_00-50-45](https://user-images.githubusercontent.com/8836461/226144504-b0750726-f8ab-48e5-b5a1-479ee893ab72.png)
[...]

PR #8119, opened 2026-01-14
make the save path for part files configurable

arvidn:


arvidn:
Responding to https://github.com/arvidn/libtorrent/issues/7194#issuecomment-3754686904

@cheater

Yes, you make a good point. A relative path makes more sense.

cheater:
would it be possible for you to make it default to .parts ? if everyone has a different directory it'll be annoying when bug reports happen.

in fact i would honestly suggest just hardcoding it to .parts

arvidn:
> would it be possible for you to make it default to .parts ? if everyone has a different directory it'll be annoying when bug reports happen.
> 
> in fact i would honestly suggest just hardcoding it to .parts

Then it would break backwards compatibility. part files for torrents downloaded with the previous version would not be picked up.

cheater:
Oh, I think we misunderstood each other.

I would still suggest being able to turn the utilization of a subdir on and off, however:
- default the subdir path to '.parts', so that when it's turned on the first time, that's the value that it gets
- better yet, make the subdir path not configurable, and always make it '.parts', while also still allowing turning it on and off, and defaulting it to off

arvidn:
the way this feature is "turned off" right now (in this patch) is to leave the `part_file_dir` an empty string. I would imagine qBT could have a default set to `.parts`, or something like that. But they would also know whether a torrent is being added for the first time, or whether an existing torrent is loaded from resume data (in which case it would have to use an empty string, to be backwards compatible).

cheater:
I strongly suggest making part_file_dir non-configurable:
- not a string, but a constant
- not configurable in the UI
- the state for whether a part file is being used with a specific torrent would be a boolean, and not an empty check on the string
- instead of having a variable part_file_dir that contains the subdir name, have a constant PART_FILE_DIR = ".part" and a variable part_file_dir_used that's a boolean

The reasons for that are as follows:

What do we win by making this configurable? Honestly, nothing that I can see. One could say it's for customization but we weren't able to customize the names of the parts files either and they were just being spewed into the save directory, so this is better in every way even without customization.

What do we lose? 
- The code is more complicated (empty string checks, etc), more to go wrong and for bugs to happen
- What do you do if one torrent has one part_file_dir and another one has another part_file_dir? How do you import both into the same instance of qBt running? How do you represent the difference?
- People showing up with weird inscrutable bugs, and now we always have to ask "did you use a weird custom part_file_dir?"
- Another piece of customization to keep in mind when debugging
- What are the security implications here? I don't know, and we don't want to have to think about that
- probably a bunch more stuff but I'm on the run :)

cheater:
oh also, when initializing, checking a boolean is loads faster than checking if a string is empty... especially with thousands of torrents, that will help

arvidn:
> What do we win by making this configurable?

Backwards compatibility

> oh also, when initializing, checking a boolean is loads faster than checking if a string is empty

I don't believe you. But even if it's slightly faster, it makes no difference in this context.

> The code is more complicated (empty string checks, etc), more to go wrong and for bugs to happen

Very marginally. The complicated parts are not related to these checks. Can you point out some specific complexity in this patch that you're concerned about?

> What do you do if one torrent has one part_file_dir and another one has another part_file_dir?

Their respective part files are stored in different directories. The name of that directory is stored in the resume data though.

cheater:
> Backwards compatibility

I think you misunderstood me.

I suggest making it possible to _turn using subdirs on and off_ and default to off. That ensures backwards compatibility like you want.

Configuring the _name of the subdir_ does not provide any backwards compatibility, since currently there is no subdir at all.

I suggest making the _name of the subdir_ NOT configurable, hardcode it, and decide whether it's being used by using a boolean flag. The _flag_ is what makes it backwards compatible.

glassez:
@cheater 
So it turns out that you're just suggesting changing it from one hardcoded option that you don't like to another hardcoded option that you like.
Well, tomorrow there will be another "cheater" who will not like your option, and he will start all this nonsense from the beginning. It doesn't look so good.

I still think that since we've already gotten into this, we need to make it as flexible as possible (as I mentioned, just let you set both relative and absolute paths). This still does not oblige the frontend application to provide all these possibilities to the user, but it at least makes it easier to further modify this behavior if necessary (without having to make new changes to libtorrent, which is much more difficult to do).

cheater:
> So it turns out that you're just suggesting changing it from one hardcoded option that you don't like to another hardcoded option that you like.

What? No. You probably came up with this and then didn't read my messages properly. I explain it three times.

Let me explain it in another way that frames things the way that you frame it.

I suggest:

- make it possible to switch from using a subdir or not (boolean flag)
- make the subdir name hard coded to ".parts" if it is used
- use no subdir name if it is not used
- default to it not being used for compatibility
- let the user turn it on or off using a checkbox in the UI

I'm not asking to "change from one hardcoded setting to another".

> This still does not oblige the frontend application to provide all these possibilities to the user

Front-end validation is a bug introducing antipattern. I've been trying to explain this earlier on. The choices should be limited by the data structures in use, not by the front end UI. This means that data on disk cannot assume values that would be impossible to represent with the UI.

arvidn:
> > So it turns out that you're just suggesting changing it from one hardcoded option that you don't like to another hardcoded option that you like.

> What? No.

The ".parts" directory name you suggested is the hard coded option.

> I'm not asking to "change from one hardcoded setting to another".

You're suggesting adding another hard coded name, after saying you don't like the first hard coded name. I believe @glassez's point is that someone else may have an equally strong argument for a different name, and then we have 3 hardcoded options. and so on.

> Front-end validation is a bug introducing antipattern. 

You say this, but you don't really make a case for it. Compare it to other APIs. `open()` lets you specify the filename you want to open. Does that make it error prone, because you might chose the wrong file?

cheater:
Hey man, I don't think you understand what front-end validation means when talking about bad patterns ... anyways, don't worry about it. Do what you want. I don't care if the subdir is called ".parts", "smurfs", "vodka", or anything else. I was just trying to save people a future headache, but nvm.

arvidn:
a client of a shared library isn't necessarily a "front-end".

glassez:
> Hey man, I don't think you understand what front-end validation means when talking about bad patterns ... anyways, don't worry about it.

It's you who talk about "front-end validation" so you could clarify what you mean.

I just said that it's a good idea to have library flexibly configurable (in those aspects where it doesn't require unnecessary complexity).
On the one hand, it allows the library developer to operate only with basic abstract concepts, not caring about the tasks of the UI, and on the other hand, it allows developers of all kinds of UI and middleware using this library to implement the behavior that they consider necessary, and easily change it without requiring changes to the library (what to do much more difficult, longer, if at all acceptable in some cases due to compatibility issues with other software using it).
Having the limited capabilities provided by the lower level (library), you often end up with what you have without being able to do something more advanced at the application level.

@cheater 
As a qBittorrent user, the above should not concern you at all. After all, qBittorrent can implement such behavior as you suggest (just a checkbox in the settings), on top of the more advanced libtorrent feature. And I think this is a great pattern.

PR #8121, opened 2026-01-19
python bindings: Install `setuptools` on Windows

xavier2k6:

PR #8122, opened 2026-01-19
fix premature seed detection in is_seed()

### LINKED ISSUE #8118: Incorrect `last_seen_complete` update due to premature seed detection in `is_seed()`
**Problem Description**

  There is a bug where the last_seen_complete status for a torrent is being updated incorrectly. This appears to be
  caused by a peer being marked as a seed for a "split second" during the initial handshake, even when it is not a
  complete seed.

  **Root Cause**

  The issue lies in the peer_connection::is_seed() function. The original implementation was:

```
bool peer_connection::is_seed() const
{
    // ...
    return m_num_pieces == m_have_piece.size()
        && m_num_pieces > 0 && t && t->valid_metadata();
}
```

  This logic is flawed because it compares m_num_pieces (the total number of pieces in the torrent) with
  m_have_piece.size() (the total capacity of the peer's bitfield). During the initial handshake, a peer's bitfield might
  be initialized to its full size but with no pieces actually set. In this state, m_num_pieces == m_have_piece.size()
  can evaluate to true, incorrectly identifying the peer as a seed before it has reported having any pieces.

  **Proposed Fix**

  A more accurate way to determine if a peer is a seed is to check if it has all the pieces. This can be accomplished by
  using the bitfield::all_set() method, which verifies that every bit in the peer's bitfield is set to 1.

  The corrected implementation of peer_connection::is_seed() would be:

```
bool peer_connection::is_seed() const
{
    // ...
    return m_have_piece.all_set()
        && m_num_pieces > 0 && t && t->valid_metadata();
}
```

  This change ensures that a peer is only identified as a seed if it has explicitly indicated that it possesses all the
  pieces of the torrent. This prevents the premature seed detection during the handshake and ensures the
  last_seen_complete status is updated correctly.

peterdrier:
## Summary
- Use `m_have_piece.all_set()` instead of comparing `m_num_pieces` with `m_have_piece.size()` in `is_seed()`
- This prevents incorrectly identifying peers as seeds during handshake when the bitfield is initialized to full capacity but no actual pieces have been reported

Fixes #8118

## Test plan
- Verify peers are not marked as seeds until they have explicitly reported having all pieces via their bitfield
- Check that `last_seen_complete` is not updated prematurely during initial peer connections

### DIFF
--- a/src/peer_connection.cpp
+++ b/src/peer_connection.cpp
@@ -2070,7 +2070,10 @@ namespace libtorrent {
 			TORRENT_ASSERT(m_have_piece.count() == m_have_piece.size());
 			TORRENT_ASSERT(m_have_piece.size() == t->torrent_file().num_pieces());
 
-			t->seen_complete();
+			// only mark last-seen-complete if we've received actual payload data
+			// from this peer, to prevent lying peers from updating the timestamp
+			if (m_statistics.total_payload_download() > 0)
+				t->seen_complete();
 			t->set_seed(m_peer_info, true);
 			TORRENT_ASSERT(is_seed());
 
@@ -6255,7 +6258,9 @@ namespace libtorrent {
 		TORRENT_ASSERT(m_recv_buffer.pos_at_end());
 		TORRENT_ASSERT(m_recv_buffer.packet_size() > 0);
 
-		if (is_seed())
+		// only mark last-seen-complete if we've received actual payload data
+		// from this peer, to prevent lying peers from updating the timestamp
+		if (is_seed() && m_statistics.total_payload_download() > 0)
 		{
 			std::shared_ptr<torrent> t = m_torrent.lock();
 			if (t) t->seen_complete();
@@ -6863,7 +6868,7 @@ namespace libtorrent {
 		// if m_num_pieces == 0, we probably don't have the
 		// metadata yet.
 		std::shared_ptr<torrent> t = m_torrent.lock();
-		return m_have_piece.all_set()
+		return m_num_pieces == m_have_piece.size()
 			&& m_num_pieces > 0 && t && t->valid_metadata();
 	}
 


sysdump:
I'm not a C++ programmer, but I think I see a possible issue:
Have you made sure that m_have_piece.all_set() does not return true if m_have_piece.size() == 0? You might need to check that m_have_piece.empty() == false

arvidn:
It makes sense that if `is_seed()` incorrectly reports `true` during handshake it would cause last-seen-complete to be incorrect. However, I don't quite see how this can be a problem with the current implementation, nor how this change would fix it.

> This prevents incorrectly identifying peers as seeds during handshake when the bitfield is initialized to full capacity but no actual pieces have been reported

I don't believe the bitfield *is* initialized to full capacity. It would be if the peer reports `HAVE_ALL` or sends a bitfield with all bits set. Bit that would be part of the handshake itself.

`m_have_piece` is initialized here:

https://github.com/arvidn/libtorrent/blob/RC_2_0/src/peer_connection.cpp#L744

```
m_have_piece.resize(t->torrent_file().num_pieces(), m_have_all);
m_num_pieces = m_have_piece.count();
```

i.e. if `m_have_all` is `false`, it will be initialized to all zeroes.

`m_num_pieces` is always initialized to 0, here: https://github.com/arvidn/libtorrent/blob/RC_2_0/src/peer_connection.cpp#L140

The edge case of having 0 pieces, and not having initialized the `m_have_piece` bitfield, that I first assumed this change was about, is already handled here:

https://github.com/arvidn/libtorrent/blob/RC_2_0/src/peer_connection.cpp#L6867

as long as `m_num_pieces == 0`, the peer is never considered a seed.

Did you try this an experience a difference?

peterdrier:
Apologies, I didn't have a good way to test that before.  I've an active torrent now though stuck at 31% which is having this issue though, so with some prayers I'll be able to cobble together a proper test.

arvidn:
you're not connected to any seed, but the last-seen-complete keeps ticking up? Is that the symptom you see?

Is it possible that seeds connect and handshake with you, just to immediately disconnect?

peterdrier:
After further investigation, I've updated this PR with a different approach.

## What I found
The actual problem was malicious peers (observed: old Transmission 2.2* client from Russia) that:
1. Connect to a torrent
2. Send a bitfield claiming to have all pieces (100%)
3. Immediately disconnect without sending any data

This behavior legitimately triggers `is_seed()` (they ARE claiming to be a seed via their bitfield), so the original `is_seed()` implementation was correct. The bitfield is properly initialized to zeroes, not full capacity as I originally assumed.

## New approach
Instead of changing `is_seed()`, this PR now adds a check for `m_statistics.total_payload_download() > 0` before calling `t->seen_complete()`. This means:
- Peers must actually send us payload data before they can update `last_seen_complete`
- Lying peers that claim 100% and immediately disconnect won't pollute the timestamp
- Normal seed behavior is unaffected

@arvidn - you were right that the original `is_seed()` change wasn't addressing a real bug. This new approach directly addresses the bad actor behavior instead.

peterdrier:
This now changes the intent behind the "last_seen_complete" field though, so it becomes more of a question to you as to if you think this is a good/reasonable change?

on my side, I'll probably just block the whole 31.200.* network in the meantime.   Though watching different torrent just now, I saw more fishy behavior two other extremely large countries..  Seems the DHT network is their medium of exploration.  joy.

peterdrier:
> you're not connected to any seed, but the last-seen-complete keeps ticking up? Is that the symptom you see?
> 
> Is it possible that seeds connect and handshake with you, just to immediately disconnect?

To be specific, I have a few stalled downloads in my queue, with 1-6 other peers all stuck at the same level per torrent.  Every 1-5 minutes, a connection from Russia comes in, from a particular ip block, which states 100% and leaves after about 30 seconds.  Rotating through all of the stalled torrents in no particular order. It’s marked as transmission 2.2.?  A quite old version, so fishy on a few points.  

One of these torrents is at 31.1% for more than a week now, arguably only kept for testing this situation. 

So, yes on your first question. Last updated increases to now() due to this activity, yet no bytes provided. 

And likely yes on your second question. They’re connected long enough to see/screenshot. I don’t know the DHT bits well enough. Perhaps this activity is probing to see what other files I’m active with? 

I also noticed every 15-30 minutes a bunch of connections appeared for a hot second before leaving. I suspect this was some tracker related peer search cadence though.

arvidn:
looks reasonable

peterdrier:
Looks like the setuptools issue from #8121 is effecting the build here..  I'm presuming the fix there will fix this?

peterdrier:
Can this one get rebased/run?  the build error is external to the changes on the pr.

HanabishiRecca:
~~Am I understanding it right that now the last seen complete value will never update for torrents you seed?~~

Nevermind, testing reveals that it does update even when you seed.

PR #8136, opened 2026-01-31
Improve compatibility with C++23

Chocobo1:
This addresses the following warning when compiling with gcc c++23 mode:
warning: space between quotes and suffix is deprecated in C++23 [-Wdeprecated-literal-operator]

### DIFF
--- a/include/libtorrent/string_view.hpp
+++ b/include/libtorrent/string_view.hpp
@@ -98,7 +98,7 @@ namespace libtorrent {
 
 inline namespace literals {
 
-	constexpr string_view operator "" _sv(char const* str, std::size_t len)
+	constexpr string_view operator ""_sv(char const* str, std::size_t len)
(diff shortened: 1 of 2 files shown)


xavier2k6:
@arvidn It may be time to add C++23 to CI ensuring it builds like (C++17 and C++20) in #7501

https://github.com/arvidn/libtorrent/blob/f2ac61c9f6508904867a2aad0111f6d125133d30/.github/workflows/linux.yml#L35

https://github.com/arvidn/libtorrent/blob/f2ac61c9f6508904867a2aad0111f6d125133d30/bindings/python/setup.py#L235

Does (C++20 and C++23) needed to be added here or anywhere else in codebase?
https://github.com/arvidn/libtorrent/blob/f2ac61c9f6508904867a2aad0111f6d125133d30/clang_tidy.jam#L70-L75

PR #8137, opened 2026-01-31
Improve compatibility with C++23 (RC_1_2)

Chocobo1:
This addresses the following warning when compiling with gcc c++23 mode:
warning: space between quotes and suffix is deprecated in C++23 [-Wdeprecated-literal-operator]


xavier2k6:
@arvidn You might want to cancel workflow as it seems to be stuck on "check headers" for the last 13days & counting!

https://github.com/arvidn/libtorrent/actions/runs/21541237722/job/62075961508

PR #8007, opened 2025-08-06
Add RC_2_0 Python type stubs

qstokkink:
Fixes #7996

This PR cherry-picks the Python typing commits from `master` and fixes them up. There intent here is to NOT make any API changes, only to introduce typing. The first four commits are cherry-picked and only the second one (
965d9f2) is rewritten.

 - One GitHub Actions job is added:  `check_wheel_type_exposure` enforces both exposed type presence and type consistency.
 - One entry is added to `.pre-commit-config.yaml`: `flake8-pyi-run` enforces modern Python syntax in the stubs.
 - I updated some Python files in `bindings/python/` to verify the typing (it is impossible to properly test types using a file that does not use types itself).
 - One Python TestCase is added to somewhat match function stub signatures to their actual bindings.

### Backporting to the `master` branch

> The first 4 commits in this PR are picked from `master` and can be thrown out when backmerging to `master` (`Add type stubs`, `Add type stubs and tests for sha256_hash`, `fix wheel types not being found`, and `Moved type stubs into libtorrent directory`).

PR #8142, opened 2026-02-06
Fixed Python CI paths and deprecations

qstokkink:
This PR contains a small assortment of fixes:

 - `macos-13` jobs have been removed from the `cibuildwheel` matrix because GitHub auto-cancels them anyway.
 - `python.yml` installs the `setuptools` dependency on Windows, which is the only platform where it was not installed.
 - The `Jamfile` has its OpenSSL directory correctly set for where GitHub Actions has it installed.
 - `pyproject.toml` puts its OpenSSL dll's in a directory where all python installs can find it (`System32`). This is necessary due to an extremely frail construction in `bindings/python/test.py` that calls `subprocess.Popen([sys.executable, ...`: `sys.executable` is not the same executable as the virtual env one! The tests that use it are essentially using a randomly installed python executable. I'd recommend rewriting these tests.


xavier2k6:
@arvidn Can you review & merge asap, this PR fixes the python binding failures in CI & other PR's can progress now.

qstokkink:
@arvidn as already remarked, this PR allows other PRs to progress. However, the Python part of this repository is old and tired, running on patchwork like this PR. Both the path resolution here and the `ImportError: DLL load failed` repo issues are symptoms of this. In particular, the Python packaging and testing could really use a makeover. I already made #8007 and I could keep on going with modernization. That would be a pretty major effort, though. So, before I spend a lot of time I want to ask: are you interested in PRs on this or would you rather keep it as-is (and/or work on it yourself)?

arvidn:
personally, I was happy with the extent of my responsibility was to produce a `.pyd`/`.so` that you could `import` into python. But things are more complicated now. I would appreciate any effort towards addressing these issues, especially if they result in fewer moving parts.

pyproject.toml
```diff
@@ -74,6 +74,7 @@ before-all = [
 select = "*-win_amd64"
 
 [tool.cibuildwheel.windows]
+before-test = '''python -c "import glob,shutil;[shutil.copy(f,'C:/Windows/System32') for f in glob.glob('C:/Program Files/OpenSSL/bin/*.dll')]"'''
```
arvidn, line 78: this could probably use a comment. Is this something windows users usually do to install openssl?

qstokkink: I'll add a comment 👍 Your question has a bit of a multifaceted answer (please forgive my wall of text):

This particular construction is only for the tests. The "usual" expectation of PyPI/pip users is that they don't have to pull in any dlls themselves (which is why you get people on your repo complaining about Python ImportErrors).

The next best thing would be for users to drop the DLLs into their Python interpreter directory. However, even that is "impossible"/impractical with the way `cibuildwheel` virtualizes the environment in combination with libtorrent's `test.py`. So, for these tests, I opted for the last resort of `System32`.

This part of why I suggest to rework the Python part of libtorrent's python bindings. Updating `setup.py` would avoid this construction.

qstokkink:
This may sound insane, but I do enjoy implementing Python modernizations - including the build system. I'll spend some of my free time making changes (in future PRs) to libtorrent's Python bits then.

My battle plan would be this:
1. Get a grip on what the bindings/signatures are: what is _actually_ being bound/exported by the cpp code (i.e., finish #8007). Then I know what I should be testing in the first place.
2. Expand and modernize the Python tests so that I'm sure I don't break anything (in the next step).
3. Upgrade the build system to make libtorrent's Python part easier to build, maintain, and use.

PR #8144, opened 2026-02-10
Add support for overriding the NAT-PMP gateway with a nat_pmp_gateway config option

samvv:
As discussed in https://github.com/arvidn/libtorrent/issues/7681, this pull request updates farcaller's changes to the latest commit on RC_2_0.

This is a fresh pull request due to the issues @qBittUser made me aware of. I started off a fresh branch and cherry-picked the relevant commits.

**This hasn't been thoroughly been tested yet.** This seems to fix users behind ProtonVPN not being able to seed due to the default gateway not being set. However, it's difficult to tell because I fail to find a good torrent to test (I used the Arch Linux image). I sometimes get speeds in the kilobytes instead of plain nothing, which is better than nothing but still not ideal.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -19,6 +19,7 @@
 	* optimize tracker announcements
 	* fix handling of invalid filenames
 	* add support for I2CP inbound.lengthVariance and outbound.lengthVariance tunnel settings
+	* add nat_pmp_gateway setting to manually set the default gateway for the NAT-PMP protocol
 
 2.0.11 released
 


src/session_impl.cpp
```diff
@@ -5674,7 +5674,16 @@ namespace {
 			ip.netmask = s->netmask;
 			std::strncpy(ip.name, s->device.c_str(), sizeof(ip.name) - 1);
 			ip.name[sizeof(ip.name) - 1] = '\0';
-			s->natpmp_mapper->start(ip);
+
+			error_code ignored;
+			std::string gateway = m_settings.get_str(settings_pack::nat_pmp_gateway);
+			boost::optional<address> gateway_addr;
+			if (gateway == "") {
```
arvidn, line 5681:
```suggestion
			if (gateway.empty()) {
```

src/session_impl.cpp
```diff
@@ -5674,7 +5674,16 @@ namespace {
...
+				gateway_addr = boost::none;
+			} else {
+				gateway_addr = make_address(gateway, ignored);
```
arvidn, line 5684: if this call fails, it would be reasonable to have `gateway_addr` be `nullopt`. Otherwise any invalid address will break NAT-PMP.

i.e. the `ignored` error code should not be ignored.

arvidn:
Does this look good? https://github.com/arvidn/libtorrent/pull/8154

PR #8146, opened 2026-02-11
disable disk_disable_copy_on_write by default

database64128:
See https://github.com/arvidn/libtorrent/pull/8103#issuecomment-3867306591 and https://github.com/qbittorrent/qBittorrent/issues/23683#issuecomment-3873848891


arvidn:
those are compelling arguments. It is a more conservative approach to leave this off. But it would be good to capture some of the reasoning of how to use this settings could be captured in the documentation of it.

database64128:
> But it would be good to capture some of the reasoning of how to use this settings could be captured in the documentation of it.

Do you want me to update the comment in `include/libtorrent/settings_pack.hpp`, or would you prefer to do it yourself?

arvidn:
I think a separate PR makes sense either way. If you don't mind updating the docs, that would be greatly appreciated. I would have to rely on your answer on the other thread anyway :)

PR #8149, opened 2026-02-14
add dump_torrent.py - Python implementation of dump_torrent utility

ismdevteam:
This PR adds a Python implementation of the dump_torrent utility that produces identical output to the original C++ binary while adding useful features.

## Added Files
- `examples/dump_torrent.py` - Main Python script
- `examples/README_DUMP_TORRENT_PY.md` - Documentation

## Features
- Complete compatibility with original dump_torrent output (text mode)
- JSON output mode with additional metadata
- Support for both v1 and v2 torrents
- Optional pad file inclusion (`--show-padfiles`)
- Custom libtorrent path support for development (`--libtorrent-path`)
- Comprehensive file statistics in JSON mode
- Graceful handling of corrupt/invalid torrents

## Testing
- Tested against 60+ torrent files including edge cases
- Validated against original dump_torrent output
- Handles pad files, symlinks, web seeds, and trackers correctly
- Supports all original command-line options

## Example Usage
```bash
# Text output (compatible with original)
./examples/dump_torrent.py torrent.torrent --dump

# JSON output with statistics
./examples/dump_torrent.py torrent.torrent --pretty

# Use custom libtorrent build
./examples/dump_torrent.py torrent.torrent --libtorrent-path /path/to/bindings
```

This PR contains **only** these two new files and no other changes.

### DIFF
--- a/examples/dump_torrent.py
+++ b/examples/dump_torrent.py
@@ -1,0 +1,806 @@
+#!/usr/bin/env python3
+"""
+dump_torrent.py - Generate metadata.json by parsing torrent files directly
+Output matches dump_torrent format exactly
+"""
+import sys
+import json
+import re
+import binascii
+from pathlib import Path
+from typing import Dict, Any, List, Optional
+import argparse
+import datetime
+import os
+import hashlib
+import io
+
+# Global variable to track libtorrent availability
+LIBTORRENT_AVAILABLE = False
+lt = None
+
+def import_libtorrent(custom_path: Optional[str] = None) -> bool:
+    """Import libtorrent, optionally from a custom path"""
+    global lt, LIBTORRENT_AVAILABLE
+    
+    # If custom path is provided, try to import from there first
+    if custom_path:
+        try:
+            sys.path.insert(0, custom_path)
+            import libtorrent as lt
+            LIBTORRENT_AVAILABLE = True
+            print(f"Using libtorrent from custom path: {custom_path}", file=sys.stderr)
+            return True
+        except ImportError:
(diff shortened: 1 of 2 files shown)


arvidn:
this example should be moved to `bindings/python`. Perhaps all python examples should be moved into an examples folder, but that's not really related to this PR. The document describing this example seems a bit excessive. A lot of it should really just belong as comments in the python file.

This is not the place to document how to build the python bindings.

examples/dump_torrent.py
```diff
@@ -0,0 +1,806 @@
...
+    # Try system libtorrent
+    try:
+        import libtorrent as lt
+        LIBTORRENT_AVAILABLE = True
+        print("Using system libtorrent", file=sys.stderr)
+        return True
+    except ImportError:
+        print("Error: libtorrent not found in system paths.", file=sys.stderr)
+        LIBTORRENT_AVAILABLE = False
+        return False
```
arvidn, line 47: this does not belong in an example meant to demonstrate how to use libtorrent

examples/dump_torrent.py
arvidn, line 1: this program dumps information from a torrent file by parsing it and printing the information in it. It does not use libtorrent. This is not an example showing how to use libtorrent to dump information from a torrent. I don't see any use of including this file in the libtorrent repository.

arvidn (changes requested):
I think you may have misunderstood what "software engineering" entails. It's not just about asking your LLM to generate code, it's also to make judgment calls about which code *should* be generated and what it should look like, and why, and that it does the thing you want it to do.

PR #8159, opened 2026-02-20
(opt-in) use curl for tracker communication

TheBaronVladimirHarkonnen:
fixes https://github.com/arvidn/libtorrent/issues/4334
closes https://github.com/arvidn/libtorrent/pull/8025

### Goal
Connection reuse improves efficiency for both the client and server side. Coupled with support for SSL session reuse and advanced HTTP features, this implementation reduces the computing cost of implementing encryption in HTTP trackers.
Hopefully, this will stimulate more trackers to adopt encryption.

### Breaking changes
 - None; disabled by default. Compile time opt-in.

**Breaking changes (with respect to the reference http_connection_tracker implementation):**
 - `send_host_in_connect` is ignored because it is required for curl's HTTP/1.x proxy and cannot be turned off. I  suggest removing the setting and always turning it on.
 - `proxy_hostnames` is ignored for HTTP proxies, because it is required in curl's HTTP proxy implementation.

### Relevant curl issues:
 - Keep-alive pings for connections in the connection cache are not processed, causing idle connections to die (https://github.com/curl/curl/blob/master/docs/TODO.md#monitor-connections-in-the-connection-pool).
 - No port-hint support for HTTPS-rr DNS records feature advertisement (ECH, ALPN) (https://github.com/curl/curl/issues/17045):
   * HTTPS-rr is currently experimental in libcurl and can only be enabled/disabled at compile time. It should be disabled until a baseline of features (such as port hint) is supported. For example, trackers can use a port hint to upgrade from HTTP to TLS connections on a different port.
   * Only works with libcurl DoH and c-ares resolvers (c-ares does not support ECH a.t.m.)

### Questions:
 - It is unclear how to run all automatic test with two HTTP stacks. Should all tests be executed for both?
 - `libtorrent/aux_/deadline_timer.hpp` uses system_clock, which is not resistant to time jumps. Any chance of changing it to steady_clock?

### Unit test coverage:
- Building with `curl=on` allows running all current tests for the curl HTTP stack.
- New test are added for ssrf and ip_filter because the simulator tests are not run yet (maybe remove these in favor of running the simulator tests?).
- New test are added for intrusive_list, bitmask and curl.
- No test are added for curl specific features like HTTP/2. Proper functioning of curl internals is assumed.
- Simulator support is not added yet
  Can be implemented by stubbing the curl_pool, and redirecting the curl_easy_* calls to secretly use the http_connection class. Is this desired? should this be a separate PR to keep this one small?

### Testing:
- A rebased version has been tested with a custom qbittorrent docker image, based on libtorrent v2.0.11.
  https://github.com/TheBaronVladimirHarkonnen/docker-qbittorrent-nox
  + Compiled with `curl_debug=on` all the HTTP connection information is logged. Observed that the connections are being reused.
  + When the application is shutting down, the "stopped" events are sent. Good, the application exits cleanly.
  + Tested that curl errors are propagated properly to the UI.
- All testcases are passing with `curl=on`, no degradation in test running times.
- Did not run the simulator.
- Has not been tested on Windows yet, I will start working on setting up a windows box.

### Implementation design:
- Uses libcurl thread-unaware multi library with boost asio async io handlers (single threaded, on the main libtorrent thread)
- All the newly added algorithms are `O(1)` per HTTP request, using `O(N)` to process `N` requests. However, the libcurl multi interface uses naive algorithms with `O(N^2)` time complexity (inside libcurl for queuing and list operations).


arvidn:
please simplify this patch. also, review it to make sure you don't duplicate any functionality. also, make sure you have CI test coverage of the new functionality.

arvidn:
btw. I would think the main issue with tracker announces right now is that there's no support for pipelining. If multiple torrents use the same tracker, they shouldn't each have to connect and SSL handshake.

TheBaronVladimirHarkonnen:
> btw. I would think the main issue with tracker announces right now is that there's no support for pipelining. If multiple torrents use the same tracker, they shouldn't each have to connect and SSL handshake.

You make a good point, I agree that this is both easy to implement in boost and cleaner to integrate into the current code base. The reason why I'm more interested in curl is as follows:
- With HTTP/1.1, it is possible to have more announces than you can process using a single connection. For example, a tracker expects an announce every 30 minutes, when an announce takes 0.5 seconds, you can only support at most `30*60*2=3600` torrents. H2 or H3 can solve this with doing many parallel announces, and the server can choose how many streams it wants to use.

- With HTTP/3 and QUIC, a server is no longer limited by the number of available TCP ports (0-65536), allowing it to keep more persistent connections.

- Encrypted Hello support, while this is currently not working yet. This is a cool privacy upgrade.

- HTTPS-RR support, making clients manually upgrade their tracker urls to HTTPS is not feasible. HTTP redirect for upgrades are not feasible. I want a server to upgrade all my torrents to HTTPS by simply adding a HTTPS-RR DNS record.

- boost beast has no development roadmap and has been in maintenance mode for years. I doubt it will ever support more advanced HTTP features. Writing correct/good code is their main goal, not low level efficiency. 

I'm not happy with the current state of curl but it is the best library out there. Thinking about the future is important too, will the simple boost implementation be good enough 10 years from now? I would like the think we are all using the new advanced HTTP features by then.

Don't be afraid to reject the integration of curl if you don't like the idea, I would totally understand.

TheBaronVladimirHarkonnen:
just a status update, I'm still busy reworking this.

TheBaronVladimirHarkonnen:
needs more work, don't review yet.

seabashed:
> needs more work, don't review yet.

Looking forward to whenever you have time for this. The fact we need to create and tear down a connection with all the overhead (especially SSL announces) that entails is quite sad. Hope to see modern approaches adopted in our favorite tools! 👍

TheBaronVladimirHarkonnen:
@seabashed
Thank you for the motivation to work on this.

Currently I'm working on creating a custom HTTP request queuing system to avoid using the curl queuing system entirely. The curl queuing was not made for applications like Libtorrent and we are better of not using it.

This queuing creates a separate queue per tracker:
- HTTP server without connection reuse, then use up to `f(queue_size)` connections, e.g. 5 connections per 200 queued items would avoid overloading the server.
- HTTP1.X server with connection reuse, then use up to `g(queue_size)` connections, e.g. 1 connection per 200 queued items.
- HTTP2/HTTP3 server with connection reuse, then use up to `k(queue_size)` connections, e.g. 1 connection with many streams would work.

The balance is between not creating too many connections, but at the same time being able to send all announces at startup/shutdown in a reasonable time-frame.

seabashed:
Thanks @TheBaronVladimirHarkonnen, is this still worth pursuing or effectively replaced by @arvidn's work on https://github.com/arvidn/libtorrent/commit/9654c5184f9ab6321430125b4bb5aed9887fdfc1?

TheBaronVladimirHarkonnen:
@seabashed I think @arvidn prefers keeping things simple and was not going to accept/merge the complexity that comes along with curl and the advanced HTTP features, hence I stopped actively working on it.

His commit deals with HTTP1.1 connection reuse only. I would like to see the following additions:
- Use more than 1 connection per domain-key, because that impacts startup and shutdown times, especially if the connection for a domain can never be reused.
- Add SSL session reuse.

What that commit cannot address is HTTP2/HTTP3 support with encrypted SNI (via HTTPS-RR records). Features like 0-RTT would be great for announces that are spread out. These might sound like overkill, but they are a great way to lessen the burden of encryption and stimulate its adoption.

If he is open to the more complex curl implementation, I would finish this draft.

arvidn:
i don't mind curl specifically. But having an agent port it with no understanding of the challenges isn't really realistic.

curl does not integrate with the boost.asio io context, so it needs more threads (I think).

there's also the question of all other http uses in libtorrent. UPnP, web seeds and http proxies, do they still use the built-in http parser?

TheBaronVladimirHarkonnen:
>But having an agent port it with no understanding of the challenges isn't really realistic

No offense taken, I can assure you that my code is handwritten (I do try to use AI for code reviews and boilerplate). I had to overcome many challenges and learned a lot about curl, boost and async. Which includes reading a lot of source code from various libraries and reference implementations (libcurl,boost,transmission,rtorrent,QUIC/https-rr specifications) to figure out how they work internally.

> curl does not integrate with the boost.asio io context, so it needs more threads (I think).

It depends on the interface that is used. In my case everything runs on main async thread.

libcurl provides a threadless library, and uses callback functions to subscribe to IO events (wrapped in src/curl_boost_socket.cpp).  Libcurl does provide a threaded interface but I don't think that is appropriate for libtorrent. 

> UPnP, web seeds and http proxies, do they still use the built-in http parser?

Yes. Only the (proxified) http connections in tracker_manager would use libcurl. It is possible to expand this later if desired. I2P connections are also sent to the built-in http client.

I'm excited to hear that you're interested. I'll try to finish this over the weekend.

Can I put my code in a custom folder/namespace libtorrent::aux::curl or is that undesirable ?

arvidn:
I genuinely did not mean "agent" as an insult btw. There are many agent-written PRs that I've merged.

I think one of the most important objectives of incorporating libcurl would have to be to reduce complexity, and being able to fully replace existing HTTP code.

The flip-side, however, is that curl supports a *lot* of features that I don't think is necessary in libtorrent (but I'm open to be wrong about that), so on net it might still introduce more complexity (but maintained by someone else).

boost.beast isn't an alternative to curl, it's an HTTP layer on top of boost.asio sockets and some convenience functions and classes. I'm pretty sure QUIC or HTTP/3 is part of its scope.

One really nice property of sitting on top of boost.asio is that it supports simulations, and it works in deterministic, time-step simulated, end-to-end tests (see simulations/).

At a brief glance on this patch, it seems like a lot of it is a C++ wrapper on top of curl. I imagine most of that really isn't necessary. The part of curl that's used by libtorrent, I would expect to be quite small.

TheBaronVladimirHarkonnen:
> being able to fully replace existing HTTP code

curl does not support virtual or I2P sockets. There is no clean way to work around this (https://github.com/curl/curl/issues/5835). That closes any potential for future HTTP2/HTTP3 support unless a more flexible HTTP library comes along (which I would welcome).

PR #8168, opened 2026-02-24
clean ':' from filenames on macOS

bwrightkc:
libtorrent currently sanitizes ':' on Windows but not on macOS.
Since ':' is illegal on HFS+/APFS, this results in unusable filenames being passed to clients on macOS.
This patch adds ':' to invalid_chars for APPLE builds, consistent with Windows handling.

### DIFF
--- a/src/torrent_info.cpp
+++ b/src/torrent_info.cpp
@@ -101,6 +101,9 @@ namespace libtorrent {
 		// style filesystems, which also further restricts valid characters
 		// https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/os/FileUtils.java;l=997?q=isValidFatFilenameChar
 		static const char invalid_chars[] = "\"*:<>?|";
+		// macOS does not permit use of a colon in a file or folder name
+#elif defined __APPLE__
+    static const char invalid_chars[] = ":";
 #else
 		static const char invalid_chars[] = "";
 #endif
--- a/src/torrent_info.cpp
+++ b/src/torrent_info.cpp
@@ -103,7 +103,7 @@ namespace libtorrent {
 		static const char invalid_chars[] = "\"*:<>?|";
 		// macOS does not permit use of a colon in a file or folder name
 #elif defined __APPLE__
-    static const char invalid_chars[] = ":";
+		static const char invalid_chars[] = ":";
 #else
 		static const char invalid_chars[] = "";
 #endif
--- a/test/test_torrent_info.cpp
+++ b/test/test_torrent_info.cpp
@@ -725,7 +725,7 @@ TORRENT_TEST(sanitize_path)
 
 	path.clear();
 	sanitize_append_path_element(path, "dev:");
-#ifdef TORRENT_WINDOWS
+#if defined TORRENT_WINDOWS || defined __APPLE__
 	TEST_EQUAL(path, "dev_");
 #else
 	TEST_EQUAL(path, "dev:");
@@ -734,7 +734,7 @@ TORRENT_TEST(sanitize_path)
 	path.clear();
 	sanitize_append_path_element(path, "c:");
 	sanitize_append_path_element(path, "b");
-#ifdef TORRENT_WINDOWS
+#if defined TORRENT_WINDOWS || defined __APPLE__
 	TEST_EQUAL(path, "c_" SEPARATOR "b");
 #else
 	TEST_EQUAL(path, "c:" SEPARATOR "b");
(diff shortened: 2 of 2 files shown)


arvidn:
do you have a source for this?
https://en.wikipedia.org/wiki/HFS_Plus seems to contradict your assertion

bwrightkc:
> do you have a source for this? https://en.wikipedia.org/wiki/HFS_Plus seems to contradict your assertion

You're right -  I have to correct myself - after a quick search, I can't find anything authoritative to back me up on that. macOS does prohibit using a colon in the filename, but apparently that isn't so much a file system restriction as it is a restrictions based on how macOS used to use the colon to separate directories eg bwright:Downloads:My Torrent.torrent. 

You can, however, find quite a few people online discussing how macOS doesn't allow you to use a colon in a filename, however. Can also do `touch test:file.torrent` in Terminal and it will instead create a file with the name test/file.torrent.

xavier2k6:
https://www.computerhardwareinc.com/2022/06/04/the-best-characters-to-use-when-naming-files-and-folders/

xavier2k6:
Perhaps `/` (forward slash) should be added for Windows??

What is `\b` for?

https://github.com/arvidn/libtorrent/blob/163d36465fec31070c7717efb759507a177ec8a6/src/torrent_info.cpp#L96

bwrightkc:
> What is `\b` for? 
> https://github.com/arvidn/libtorrent/blob/163d36465fec31070c7717efb759507a177ec8a6/src/torrent_info.cpp#L96

\b is the backspace character.

xavier2k6:
> backspace character

👍 

Ya, googled it afterwards......

xavier2k6:
xref:
- https://github.com/qbittorrent/qBittorrent/pull/23386

arvidn (approved):

arvidn:
Ok. it seems like a reasonable change. The main thing I worry about with changing this filter is if someone has downloaded (and is seeding) a torrent with this filename in it. If that person upgrades it will no longer be seeding the file (but possible re-download it under the new name).

But if it hasn't been possible to create these files on mac anyway, it shouldn't be a problem.

arvidn:
the unit tests need to be updated

src/torrent_info.cpp
```diff
@@ -101,6 +101,9 @@ namespace libtorrent {
 		// style filesystems, which also further restricts valid characters
 		// https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/os/FileUtils.java;l=997?q=isValidFatFilenameChar
 		static const char invalid_chars[] = "\"*:<>?|";
+		// macOS does not permit use of a colon in a file or folder name
+#elif defined __APPLE__
+    static const char invalid_chars[] = ":";
```
arvidn:
```suggestion
		static const char invalid_chars[] = ":";
```

vafada:
> touch test:file.torrent

i tried this and it created `test:file.torrent` on my Mac

Screenshot 2026-03-06 at 9 34 54 PM

bwrightkc:
> > touch test:file.torrent
> 
> i tried this and it created `test:file.torrent` on my Mac
> Screenshot 2026-03-06 at 9 34 54 PM

Correct. Note, however, that when you go look for the file in Finder, it will be named test/file.torrent. My understanding of this is that classic macOS used : as the folder separator, unix uses /, and OS X at release wanted to keep the capability of referencing paths using : in applications so finder translates : to /.

vafada:
> > > touch test:file.torrent
> > 
> > 
> > i tried this and it created `test:file.torrent` on my Mac
> > Screenshot 2026-03-06 at 9 34 54 PM
> 
> Correct. Note, however, that when you go look for the file in Finder, it will be named test/file.torrent. My understanding of this is that classic macOS used : as the folder separator, unix uses /, and OS X at release wanted to keep the capability of referencing paths using : in applications so finder translates : to /.

Oh yes, I see now. Though it doesn't seems like `:` is an invalid character... Its just Finder being dumb 🤷

arvidn:
I think this makes my concern much stronger. Anyone, on a Mac, seeding a torrent where a filename has a `:` in it would break on an upgrade.

101Dude:
I think the issue can be summarized as follows:
```
Stored  ":"  → Finder displays "/"
Typed "/"   → Finder stores ":"
Typed ":"   → Finder rejects it and reverts to Stored  ":"  → Finder displays "/"
```
The only issue with colons in filenames is that Finder.app won't show them as colons. 

Therefore this PR would not make an otherwise unusable torrent usable. It would change a valid on-disk filename solely to work around how Finder displays it.

PR #8171, opened 2026-02-27
Test merge (external)

qstokkink:
In an act of "dogfooding", I ported my own Python type contributions back to `master`.

This PR is a full `RC_2_0` -> `master` merge. It can be merged, or it can be used as inspiration for a different PR (and then I'll close it later).

First run failures ([force-push](https://github.com/arvidn/libtorrent/compare/6a02bb2712557e6ecd93da55a9a8b5216b73ddc9..8c642f01aba42b4a0c551c549145ea8dc9beec26)):
 - [X] I forgot to run precommit for code styling. Not entirely sure why pre-commit modified `libtorrent_settings.h` and `settings.cpp`, but I guess that's fine?
 - [X] The Python bindings on Windows run Python 3.6, which reached EOL in 2021. This workflow has to be updated to Python 3.9.

PR #8175, opened 2026-03-03
Test merge

arvidn:


arvidn:
@qstokkink I'm about to land this. Thanks for your help!
There are some missing session settings missing from the type stubs. I'll follow up with a separate PR

qstokkink:
@arvidn No worries; I caused the problems, so I feel responsible for solving them. I'll close #8171 in a bit then.

Ideally, I would like to get the Python bindings back into pristine shape. It's going to be a bit of effort, though. Maybe getting OpenSSL packaged into the Python wheel would be a nice next step.

PR #8181, opened 2026-03-06
fix issue when saving resume data on a magnet link without metadata

arvidn:
the new `layout()` function requires that we have the metadata

### DIFF
--- a/src/torrent.cpp
+++ b/src/torrent.cpp
@@ -7392,7 +7392,7 @@ namespace {
 				ret.verified_leaf_hashes.emplace_back(t.verified_leafs());
 			}
 
-			if (!has_hash_picker() && !m_have_all)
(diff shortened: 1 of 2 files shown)


glassez:
Then how do you propose to save resume data for torrents that do not have the received metadata?

arvidn:
this was a regression. without this patch the test triggers an assert because it's trying to save parts of the resume data that requires iterating files, but we don't know about the files yet.

the test is ensuring that save_resume_data() succeeds

arvidn:
I see, the comment in the test was wrong, fixed now.

glassez:
I'm confusing by the following code comment:
>adding a magnet link (without metadata) and immediately calling save_resume_data() should fail with save_resume_data_failed_alert, because the torrent info has not been received yet.

I interpret it so that saving resume data for torrents without metadata is now impossible.

PR #8197, opened 2026-03-11
new fuzzers

arvidn:
depends on https://github.com/google/oss-fuzz/pull/15399

To get the boost-json package


arvidn:
it seems OSSFuzz has to be updated to install boost.json

PR #8215, opened 2026-03-16
Allow multiple connections from the same Peer ID

lzhzh1:
~~if the value of settings_pack::allow_multiple_connections_per_ip is true, then skip the check for the same pid~~
Add a new option allow_multiple_connections_per_pid. When set to true, it skips the duplicate Peer ID check, thereby allowing multiple connections from the same peer.
The issue #8214

### DIFF
--- a/include/libtorrent/peer_list.hpp
+++ b/include/libtorrent/peer_list.hpp
@@ -69,6 +69,7 @@ namespace libtorrent {
 	{
 		bool is_finished = false;
 		bool allow_multiple_connections_per_ip = false;
+		bool allow_multiple_connections_per_pid = false;
 
 		// this is set by peer_list::add_peer to either true or false
 		// true means the peer we just added was new, false means
--- a/include/libtorrent/peer_list.hpp
+++ b/include/libtorrent/peer_list.hpp
@@ -69,7 +69,6 @@ namespace libtorrent {
 	{
 		bool is_finished = false;
 		bool allow_multiple_connections_per_ip = false;
-		bool allow_multiple_connections_per_pid = false;
 
 		// this is set by peer_list::add_peer to either true or false
 		// true means the peer we just added was new, false means
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -412,6 +412,9 @@ namespace aux {
 			// likely to fail in some edge cases. It is not recommended to enable
 			// this feature.
 			allow_multiple_connections_per_ip = bool_type_base,
+			
+			//
+			allow_multiple_connections_per_pid,
 
 #if TORRENT_ABI_VERSION == 1
 			// if set to true, upload, download and unchoke limits are ignored for
--- a/include/libtorrent/settings_pack.hpp
+++ b/include/libtorrent/settings_pack.hpp
@@ -412,9 +412,6 @@ namespace aux {
 			// likely to fail in some edge cases. It is not recommended to enable
 			// this feature.
 			allow_multiple_connections_per_ip = bool_type_base,
-			
-			//
-			allow_multiple_connections_per_pid,
 
 #if TORRENT_ABI_VERSION == 1
 			// if set to true, upload, download and unchoke limits are ignored for
@@ -1049,6 +1046,14 @@ namespace aux {
 			// NOCOW flag on the download directory, which will cause all files
 			// created within it to have the NOCOW flag set.
 			disk_disable_copy_on_write,
+			
+			// determines if connections from the same Peer ID as existing
+			// connections should be rejected or not. Typically, we
+			// only establish a single connection with each peer. If
+			// a peer has multiple IP addresses, enabling this feature
+			// may improve transfer efficiency, but it may also
+			// increase network load.
+			allow_multiple_connections_per_pid,
 
 			max_bool_setting_internal
 		};
--- a/simulation/test_allow_multiple_connections_per_pid.cpp
+++ b/simulation/test_allow_multiple_connections_per_pid.cpp
@@ -1,0 +1,227 @@
+/*
+
+
+
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in
+      the documentation and/or other materials provided with the distribution.
+    * Neither the name of the author nor the names of its
+      contributors may be used to endorse or promote products derived
+      from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+
+*/
+
+#include <functional>
+
+#include "libtorrent/session.hpp"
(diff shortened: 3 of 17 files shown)


arvidn:
if you want to allow multiple connections to the same peer via different IP addresses, I don't think this is the right solution. you would still want to avoid multiple connections to the same IP

arvidn:
Maybe it's OK to accept multiple connections to the same peer as long as they use different address families. But you could also have multiple routes to the internet. But the common case is (most likely) that you don't want to have multiple connections to the same peer.

lzhzh1:
> Maybe it's OK to accept multiple connections to the same peer as long as they use different address families. But you could also have multiple routes to the internet. But the common case is (most likely) that you don't want to have multiple connections to the same peer.

Thank you so much for your comments! I've introduced a new option, `allow_multiple_connections_per_pid`, to toggle the duplicate Peer ID check, which ensures other functionalities remain unaffected.

Since I'm still a beginner, I haven't been able to create a proper standalone test file yet. However, I've manually verified the logic by compiling my forked libtorrent and a simply modified version of qBittorrent to control the value of `allow_multiple_connections_per_pid`. It seems to be working as expected. Thanks again!"

allow_multiple_connections_per_pid = true
allow_multiple_connections_per_pid_true
allow_multiple_connections_per_pid = false 
(The result of manually forcing Tixati to disconnect established connections and then retry the connection)
allow_multiple_connections_per_pid_false

src/bt_peer_connection.cpp
```diff
@@ -3534,28 +3534,31 @@ namespace {
...
+					}
+					else
+					{
+						disconnect(errors::duplicate_peer_id, operation_t::bittorrent);
+						return;
+					}
 				}
 			}
-
+
```
arvidn, line 3561:
```suggestion

```

test/settings.cpp
```diff
@@ -77,6 +77,7 @@ lt::settings_pack settings()
 
 #ifndef TORRENT_BUILD_SIMULATOR
 	pack.set_bool(settings_pack::allow_multiple_connections_per_ip, true);
+	pack.set_bool(settings_pack::allow_multiple_connections_per_pid, true);
```
arvidn: this doesn't seem like a good idea. This changes a lot of existing tests. It would make sense to have a separate test to exercise this new feature though

test/setup_transfer.cpp
```diff
@@ -1016,7 +1016,11 @@ setup_transfer(lt::session* ses1, lt::session* ses2, lt::session* ses3
 	if (ses3) ses3->set_peer_class_filter(f);
 
 	settings_pack pack;
-	if (ses3) pack.set_bool(settings_pack::allow_multiple_connections_per_ip, true);
+	if (ses3) 
+	{
+		pack.set_bool(settings_pack::allow_multiple_connections_per_ip, true);
+		pack.set_bool(settings_pack::allow_multiple_connections_per_pid, true);
```
arvidn: this doesn't seem right either. Is there a good reason to make this change?

test/swarm_suite.cpp
```diff
@@ -85,6 +85,7 @@ void test_swarm(test_flags_t const flags)
 
 	settings_pack pack = settings();
 	pack.set_bool(settings_pack::allow_multiple_connections_per_ip, true);
+	pack.set_bool(settings_pack::allow_multiple_connections_per_pid, true);
```
arvidn: I don't think it's a good idea to change these tests, or the tests in `test_auto_unchoke.cpp`.

lzhzh1:
My apologies for the initial misunderstanding regarding the test files, which led me to place `allow_multiple_connections_per_pid` alongside `allow_multiple_connections_per_ip`. I have now added a new dedicated test file `libtorrent/simulation/test_allow_multiple_connections_per_pid.cpp` for it. I would appreciate any further feedback or suggestions you may have. Thank you!
```
root@hello-world:~/libtorrent/simulation# b2 test_allow_multiple_connections_per_pid
CXXFLAGS =
LDFLAGS =
OS = LINUX
BOOST_ROOT =
OS = LINUX
warning: No toolsets are configured.
warning: Configuring default toolset "gcc".
warning: If the default is wrong, your build may not work correctly.
warning: Use the "toolset=xxxxx" option to override our guess.
warning: For more configuration options, please consult
warning: https://www.bfgroup.xyz/b2/manual/release/index.html#bbv2.overview.configuration
warning: non-free usage requirements <threading>multi ignored
warning: in main-target simulator at libsimulator/Jamfile:105
...patience...
...found 2238 targets...
...updating 22 targets...
gcc.compile.c++ bin/test_allow_multiple_connections_per_pid.test/gcc-14/debug/asserts-on/crypto-built-in/cxxstd-14-iso/export-extra-on/invariant-checks-full/picker-debugging-on/simulator-on/threading-multi/test_allow_multiple_connections_per_pid.o
gcc.compile.c++ bin/test_allow_multiple_connections_per_pid.test/gcc-14/debug/asserts-on/crypto-built-in/cxxstd-14-iso/export-extra-on/invariant-checks-full/picker-debugging-on/simulator-on/threading-multi/create_torrent.o
gcc.compile.c++ bin/test_allow_multiple_connections_per_pid.test/gcc-14/debug/asserts-on/crypto-built-in/cxxstd-14-iso/export-extra-on/invariant-checks-full/picker-debugging-on/simulator-on/threading-multi/disk_io.o
gcc.compile.c++ bin/test_allow_multiple_connections_per_pid.test/gcc-14/debug/asserts-on/crypto-built-in/cxxstd-14-iso/export-extra-on/invariant-checks-full/picker-debugging-on/simulator-on/threading-multi/setup_dht.o
gcc.compile.c++ bin/test_allow_multiple_connections_per_pid.test/gcc-14/debug/asserts-on/crypto-built-in/cxxstd-14-iso/export-extra-on/invariant-checks-full/picker-debugging-on/simulator-on/threading-multi/setup_swarm.o
gcc.compile.c++ bin/test_allow_multiple_connections_per_pid.test/gcc-14/debug/asserts-on/crypto-built-in/cxxstd-14-iso/export-extra-on/invariant-checks-full/picker-debugging-on/simulator-on/threading-multi/transfer_sim.o
gcc.compile.c++ bin/test_allow_multiple_connections_per_pid.test/gcc-14/debug/asserts-on/crypto-built-in/cxxstd-14-iso/export-extra-on/invariant-checks-full/picker-debugging-on/simulator-on/threading-multi/utils.o
gcc.link bin/test_allow_multiple_connections_per_pid.test/gcc-14/debug/asserts-on/crypto-built-in/cxxstd-14-iso/export-extra-on/invariant-checks-full/picker-debugging-on/simulator-on/threading-multi/test_allow_multiple_connections_per_pid
testing.capture-output bin/test_allow_multiple_connections_per_pid.test/gcc-14/debug/asserts-on/crypto-built-in/cxxstd-14-iso/export-extra-on/invariant-checks-full/picker-debugging-on/simulator-on/threading-multi/test_allow_multiple_connections_per_pid.run
**passed** bin/test_allow_multiple_connections_per_pid.test/gcc-14/debug/asserts-on/crypto-built-in/cxxstd-14-iso/export-extra-on/invariant-checks-full/picker-debugging-on/simulator-on/threading-multi/test_allow_multiple_connections_per_pid.test
...updated 22 targets...
```

simulation/test_allow_multiple_connections_per_pid.cpp
```diff
@@ -0,0 +1,227 @@
...
+	for (auto const& err : d.alerts)
+		if (err == lt::errors::duplicate_peer_id)
+			has_duplicate_error = true;
+
+	TEST_CHECK(has_duplicate_error);
+}
+
+// allow_multiple_connections_per_pid = true:
+// a second connection from a different IP but the same peer-id should be allowed
+TORRENT_TEST(allow_multiple_connections_per_pid_true)
```
arvidn: it looks like there are some opportunities to merge these two tests, or have a common core test function

include/libtorrent/peer_list.hpp
```diff
@@ -69,6 +69,7 @@ namespace libtorrent {
 	{
 		bool is_finished = false;
 		bool allow_multiple_connections_per_ip = false;
+		bool allow_multiple_connections_per_pid = false;
```
arvidn: this field is not used right now. I think you meant to update `peer_list.cpp` as well, to take this into account. However, I don't believe `torrent_peer` holds the peer-ID, so I don't think it can take this into account when filtering peers.

I think adding this is a mistake.

lzhzh1:
Thanks for your review! I removed the unnecessary code and merged the test code.

arvidn (approved):

include/libtorrent/settings_pack.hpp
```diff
@@ -1046,6 +1046,14 @@ namespace aux {
...
 			// created within it to have the NOCOW flag set.
 			disk_disable_copy_on_write,
+			
+			// determines if connections from the same Peer ID as existing
+			// connections should be rejected or not. Typically, we
+			// only establish a single connection with each peer. If
+			// a peer has multiple IP addresses, enabling this feature
+			// may improve transfer efficiency, but it may also
+			// increase network load.
+			allow_multiple_connections_per_pid,
```
milahu, line 1056: nitpick: this should be renamed

```diff
-allow_multiple_connections_per_pid
+allow_multiple_connections_per_peer_id

-simulation/test_allow_multiple_connections_per_pid.cpp
+simulation/test_allow_multiple_connections_per_peer_id.cpp
```

to make it consistent with the rest of the codebase

https://github.com/arvidn/libtorrent/blob/cb6fe6b9c28735b42edd77740a554d1709acad02/include/libtorrent/close_reason.hpp#L49

fixed in #8373

Rootax:
Was this merged into 2.1 rc2 ?

lzhzh1:
> Was this merged into 2.1 rc2 ?

Yes!

Rootax:
> > Was this merged into 2.1 rc2 ?
> 
> Yes!

Thx, I was not sure. it seems qbittorrent master doesn't expose de setting in the gui yet.

PR #8285, opened 2026-04-24
Add Python MacOS x86_64 builds to cibuildwheel

qstokkink:
I noticed there were no MacOS x86_64 builds for 2.1.0-rc1. Having at least one x86_64 build allows us to create a universal2 binary for many different MacOS x86_64 OS versions. For deployment, it would be nice to have an "official" wheel.

I made a test build to verify that this PR's targets work without any issues:
https://github.com/qstokkink/libtorrent/actions/runs/24876857318

---

`CIFuzz / Fuzzing` failure seems unrelated to this PR:
```
 error: Invalid property '<webtorrent>on': unknown feature 'webtorrent'.
```


arvidn:
this looks like a real issue though: https://github.com/qstokkink/libtorrent/actions/runs/24876857318/job/72836870868#step:4:38

```
Successfully installed librt-0.9.0 mypy-1.19.1 mypy_extensions-1.1.0 pathspec-1.1.0 tomli-2.4.1 typing_extensions-4.15.0
WARNING: Requirement './wheelhouse/*_x86_64.whl' looks like a filename, but the file does not exist
ERROR: Invalid wheel filename (wrong number of parts): '*_x86_64'
```

arvidn (approved):

qstokkink:
> this looks like a real issue though: https://github.com/qstokkink/libtorrent/actions/runs/24876857318/job/72836870868#step:4:38
> 

That's due to my testing setup. The type checking depends on a specific Linux build, which I stripped out in that run to avoid building the whole `cibuildwheel` tree.

PR #8295, opened 2026-04-27
fix flaky python test

arvidn:
the issue with `wait_for()` is if a test calls it twice, and both events are consumed by the first call, the test will fail.

### DIFF
--- a/bindings/python/tests/alert_test.py
+++ b/bindings/python/tests/alert_test.py
@@ -393,28 +393,6 @@ class EnumTest(unittest.TestCase):
 _A = TypeVar("_A", bound=lt.alert)
 
 
-def wait_for(
-    session: lt.session,
-    alert_type: Type[_A],
-    *,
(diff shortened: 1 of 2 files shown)


.github/workflows/python.yml
```diff
@@ -125,13 +125,42 @@ jobs:
...
+          echo "===== backtrace: $core ====="
+          gdb -batch \
+            -ex "info threads" \
+            -ex "thread apply all bt full" \
+            python3 "$core" 2>&1 || true
+        done
+
+    - name: upload core dump (Linux)
+      if: runner.os == 'Linux' && failure()
+      uses: actions/upload-artifact@v4
```
userdocs: https://github.com/actions/upload-artifact/releases/tag/v7.0.1

arvidn: I can't reproduce this segfault locally. this is just temporary attempt to diagnose it

PR #8344, opened 2026-05-09
improve sanitization of filenames, and symlinks in particular

arvidn:
if both a file and a symlink target pointing to the file become sanitized, make sure they are sanitized exactly the same way


tehcneko:
@arvidn the removal of U+200D breaks emojis like ❤️‍🩹,❤️‍🔥 and 🐈‍⬛ etc.
Also, the removal of U+200C breaks languages Persian and Arabic.

PR #8345, opened 2026-05-09
clear_piece

arvidn:
`async_clear_piece()` is currently not a proper fence job, and may not correctly synchronize with outstanding write and hash jobs.

this PR is work-in-progress


arvidn:
this has been superceded by https://github.com/arvidn/libtorrent/pull/8449

PR #8368, opened 2026-05-13
add python binding alert.timestamp()

milahu:
useful to restore the exact time order of alerts
from multiple libtorrent instances running in parallel

### time precision

before f71405d7a0669eef90530ba292ebc8541e16d3bd
the time precision was limited to seconds
but for ordering alerts by time, we need higher precision
so now `alert.timestamp()` returns times with microsecond precision

todo? change more `convert` functions in
[libtorrent/bindings/python/src/datetime.cpp](https://github.com/arvidn/libtorrent/blob/RC_2_0/bindings/python/src/datetime.cpp)
to preserve microsecond precision

### cost

probably the binding for `alert.timestamp()`
has not-yet been added
because most users dont need it

which begs the question:
what is the cost of adding this binding?

as far as i can tell, the cost is near zero
because the conversion to python's `datetime.datetime`
happens only when `alert.timestamp` is actually called

### ordering

i have tried to preserve the order from
[libtorrent/include/libtorrent/alert.hpp](https://github.com/arvidn/libtorrent/blob/RC_2_0/include/libtorrent/alert.hpp)

```cc
	struct TORRENT_EXPORT alert
	{
		time_point timestamp() const;
		virtual int type() const noexcept = 0;
		virtual char const* what() const noexcept = 0;
		virtual std::string message() const = 0;
		virtual alert_category_t category() const noexcept = 0;
	};
```

### tests

i hope we can add this without tests...?

tested manually with

```py
def process_alerts(session, prefix, state):
    alerts = session.pop_alerts()
    for a in alerts:
        if isinstance(a, lt.peer_log_alert):
            print("a.timestamp()", a.timestamp().strftime(r"%FT%T.%f"))
```

example output

```
a.timestamp() 2026-05-13T22:01:28.760659
```

### disclosure

the changes were made with the help of AI (chatGPT)
but were optimized for a minimal diff

### DIFF
--- a/bindings/python/libtorrent/__init__.pyi
+++ b/bindings/python/libtorrent/__init__.pyi
@@ -604,6 +604,11 @@ class alert(metaclass=_BoostBaseClass):
         tracker_notification: Literal[16]
         upload_notification: Literal[8388608]
 
+    def timestamp(self) -> datetime.datetime:
+        """
+        timestamp( (alert)arg1) -> datetime.datetime :
+        """
+
     def category(self) -> int:
         """
         category( (alert)arg1) -> object :
--- a/bindings/python/src/alert.cpp
+++ b/bindings/python/src/alert.cpp
@@ -309,6 +309,7 @@ void bind_alert()
 
 	{
 		scope alert_scope = class_<alert, noncopyable>("alert", no_init)
+								.def("timestamp", &alert::timestamp)
 								.def("message", &alert::message)
 								.def("what", &alert::what)
 								.def("category", &alert::category)
--- a/bindings/python/src/datetime.cpp
+++ b/bindings/python/src/datetime.cpp
@@ -65,13 +65,21 @@ struct time_point_to_python
 	static PyObject* convert(T const pt)
 	{
 		using std::chrono::duration_cast;
+		using std::chrono::microseconds;
+		using std::chrono::seconds;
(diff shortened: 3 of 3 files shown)


arvidn:
I would prefer fixes in the python binding be made against the `RC_2_1` branch. there's been quite a few changes and it's difficult to merge forward.

Also, last time chatGPT tried to convert times between different clocks it got it quite wrong (but that was 2 years ago) :)

milahu:
> `RC_2_1` branch

fixed

again, manually tested

arvidn (dismissed):

arvidn:
please run pre-commit to apply the clang format config

milahu:
> pre-commit

fixed

i was confused because [clang-format check (changed lines vs PR base)](https://github.com/arvidn/libtorrent/actions/runs/25828072466/job/76145746971#step:6:1) printed 64358 lines 0__o

here the diff is only 53 lines

```
nix-shell -p llvmPackages_18.clang-tools
git-clang-format --diff HEAD^1
git-clang-format --diff HEAD^1 | wc -l
git-clang-format --diff HEAD^1 | patch -p1
git commit -a --amend
```

arvidn (approved):

arvidn:
it seems like the windows python bindings test is more prone to failure with this change.

PR #8369, opened 2026-05-13
Fix quoted I2P SAM values

M-Hassan-Raza:
This fixes quoted SAM response value parsing.

split_string(..., '"') already returns the value without the closing quote, so remove_suffix() dropped a valid character. The parser also now accepts quoted empty values, which SAM 3.2 allows implementations to send.

Checked locally with quoted, unquoted, and quoted-empty response values, plus the i2p_stream.cpp CMake compile and pre-commit hooks for the touched file.

### DIFF
--- a/include/libtorrent/i2p_stream.hpp
+++ b/include/libtorrent/i2p_stream.hpp
@@ -314,8 +314,6 @@ private:
 			if (!remaining.empty() && remaining[0] == '"')
 			{
 				std::tie(value, remaining) = split_string(remaining.substr(1), '"');
-				if (value.empty()) { handle_error(invalid_response, h); return; }
-				value.remove_suffix(1);
 			}
 			else
 			{


include/libtorrent/i2p_stream.hpp
```diff
@@ -315,7 +315,6 @@ struct i2p_stream : proxy_base
 			{
 				std::tie(value, remaining) = split_string(remaining.substr(1), '"');
 				if (value.empty()) { handle_error(invalid_response, h); return; }
```
arvidn: I imagine this check is not correct either then. a quoted empty string is valid in principle, as far as I know.

M-Hassan-Raza: Thanks, agreed. I checked the SAM v3 spec, and the Empty Values section says `KEY=""` may be valid depending on the implementation: https://i2p.net/en/docs/api/samv3/#empty-values

So rejecitng quoted empty values in the shared parser was too strict. I pushed an update to allow that while keeping the original quoted-value truncation fix.

arvidn (approved):

PR #8370, opened 2026-05-13
Fix UPnP content-type checks

M-Hassan-Raza:
This changes the UPnP response content-type checks to look for full media type substrings instead of any matching character.

The old find_first_of() calls could accept unrelated types like text/html because they only checked whether any character from text/xml was present. Using find() keeps the intended XML/SOAP allowlist behavior.

### DIFF
--- a/include/libtorrent/upnp.hpp
+++ b/include/libtorrent/upnp.hpp
@@ -68,6 +68,8 @@ namespace aux {
 		std::array<char, 1500> buffer;
 		udp::endpoint remote;
 	};
+
+	TORRENT_EXTRA_EXPORT bool is_upnp_xml_content_type(string_view content_type);
 }
 
 namespace upnp_errors {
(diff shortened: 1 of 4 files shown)


arvidn (approved):

M-Hassan-Raza:
Hmm I think I'll patch the older flaky tests as well, they were hiding this bug. pushing and update shortly to wrap up this PR

arvidn:
can you explain what the changes to `test/web_server.py` achieve?

M-Hassan-Raza:
> can you explain what the changes to `test/web_server.py` achieve?

It makes the test fixture handle the UPnP control POSTs.

Before this, `web_server.py` only handled GET, so the UPnP POSTs got Python’s default 501 HTML response. The old `find_first_of()` bug accidentally accepted that `text/html` response, so the test still passed.

With the stricter content-type check, the fixture needs to return the intended XML response instead. So this adds POST handling for the UPnP control paths and sends the canned XML response with an XML content type.

arvidn (approved):

PR #8372, opened 2026-05-14
add setting allow_multiple_connections_per_peer

### LINKED ISSUE #8367: reconnect is rejected after unclean disconnect: duplicate peer-id
I found a reconnect issue in libtorrent
version cb6fe6b9c28735b42edd77740a554d1709acad02 (RC_2_0) and 0b3a049ea4e16ca1896c638a648b6c3210e54efe (RC_2_1)
when a peer process exits uncleanly

### Reproduction

* start a seeder
* start a leecher
* connect leecher to seeder
* abruptly kill the leecher process via `os._exit(1)`
  (or equivalent hard termination)
* immediately start a new leecher instance
* reconnect from the same IP and same listen port

Important details:

* the old leecher does not shut down cleanly
* the new leecher uses same listen port (the same peer_id?)
* the seeder still thinks the old connection is alive

seeder peer log:

```text
*** DUPLICATE PEER [ this: "127.0.0.1" that: "127.0.0.1" ]
*** CONNECTION_CLOSED [ op: 1 ERROR: (libtorrent:79) duplicate peer-id ]
```

The new connection is rejected
and the peer cannot reconnect until the old connection eventually times out.

### Python reproducer

here is a small Python reproducer
using separate libtorrent processes
and forced abrupt termination via `os._exit(1)`

```sh
git clone https://github.com/milahu/libtorrent-issue-8367
cd libtorrent-issue-8367
git checkout fcc32bcd740acf05bfcdf394e0c8b0a7b811437a
nix-shell
./repro.py
```

### Expected behavior

If the old connection is dead/stale
and a new connection from the same peer arrives,
libtorrent should close the stale connection and keep the new one
instead of rejecting the reconnect attempt.

This is especially problematic for applications that may terminate abruptly
(Ctrl+C, crash, SIGKILL, power loss, etc.)
and then restart quickly.

one goal of libtorrent is to also work under hostile network conditions
(packet loss, connection loss, reconnects, ...)
and this is an example of such network conditions

### Challenge

the main challenge in fixing this issue
is to decide when to close an old connection
in favor of a new connection from the same peer

we have no "hard evidence"
because the peer did not close the connection
so we need some heuristic
like "how much data was transferred over this connection in the last N seconds"
or "did the transfer rate over this connection drop to zero"

### Workaround

to avoid this issue, the leecher can pick a random listen port
then on reconnect, it is no longer counted as "duplicate peer"

### Related

this issue is **not** fixed by [setting](https://www.libtorrent.org/reference-Settings.html)
`allow_multiple_connections_per_ip = True`
`allow_multiple_connections_per_pid = True` (https://github.com/arvidn/libtorrent/pull/8215)
`ignore_limits_on_local_network = True` (default)
`anonymous_mode = True`

### Disclosure

this issue and the repro script
were created with the help of AI (chatGPT)
but the issue was discovered "by hand"

### Sources

this is where the seeder closes the new connection

[libtorrent/src/peer_list.cpp](https://github.com/arvidn/libtorrent/blob/cb6fe6b9c28735b42edd77740a554d1709acad02/src/peer_list.cpp#L735) (RC_2_0)
[libtorrent/src/peer_list.cpp](https://github.com/arvidn/libtorrent/blob/0b3a049ea4e16ca1896c638a648b6c3210e54efe/src/peer_list.cpp#L720) (RC_2_1)

```cc
	bool peer_list::new_connection(peer_connection_interface& c, int session_time
		, torrent_state* state)
	{
// ...
				// the new connection is a local (outgoing) connection
				// or the current one is already connected
				if (i->connection->is_outgoing() == c.is_outgoing())
				{
					// if the other end connected to us both times, just drop
					// the second one. Or if we made both connections.
					// FIXME drop the first connection if it is idle
					printf("libtorrent/src/peer_list.cpp 740\n");
					c.disconnect(errors::duplicate_peer_id, operation_t::bittorrent);
					return false;
				}
```

seeder log

```
libtorrent/src/peer_list.cpp 740
<<< INCOMING_CONNECTION [ ep: 127.0.0.1:36231 type: uTP seed: 0 p: (nil) local: 127.0.0.1:45219 ]
*** DUPLICATE PEER [ this: "127.0.0.1" that: "127.0.0.1" ]
*** CONNECTION_CLOSED [ op: 1 ERROR: (libtorrent:79) duplicate peer-id ]
```

this disconnect is **not** guarded by

```cc
if (!t->settings().get_bool(settings_pack::allow_multiple_connections_per_pid))
```

like in
[libtorrent/src/bt_peer_connection.cpp](https://github.com/arvidn/libtorrent/blob/cb6fe6b9c28735b42edd77740a554d1709acad02/src/bt_peer_connection.cpp#L3525)

milahu:
fix #8367


arvidn:
this seems like a bug in the recent `allow_multiple_connections_per_pid` feature. It's not clear why we would need another setting, or how it differs from peer_id.

what does "peer" refer to in that name, if it's not IP nor peer-ID?

milahu:
> what does "peer" refer to in that name, if it's not IP nor peer-ID?

peer = IP address + port

`include/libtorrent/settings_pack.hpp`

```cc
			// allow multiple connections
			// from the same IP address and port
			allow_multiple_connections_per_peer,
```

so we could rename

```diff
-allow_multiple_connections_per_peer
+allow_multiple_connections_per_ip_and_port
```


so in total, would have these settings

```
allow_multiple_connections_per_ip
allow_multiple_connections_per_ip_and_port
allow_multiple_connections_per_peer_id [*]
```

`[*]` with #8373

glassez:
> peer = IP address + port

`IP address + port` is usually called `endpoint`, no?

milahu:
that would give

```
allow_multiple_connections_per_ip
allow_multiple_connections_per_endpoint
allow_multiple_connections_per_peer_id
```

but `ip_and_port` is more explicit than `endpoint`
in effect, more user-friendly

also because setting
`allow_multiple_connections_per_ip_and_port = True`
would imply
`allow_multiple_connections_per_ip = True`
(TODO implement)

arvidn:
TCP/IP does not allow two TCP connections to have the same (src-ip, src-port, dst-ip, dst-port) tuple. I don't believe it's possible for an IP and port to connect multiple times to the same IP and port.

milahu:
> TCP/IP does not allow two TCP connections to have the same (src-ip, src-port, dst-ip, dst-port) tuple.

yes, this allows us to detect stale connections

in d95d00758efa13492a24774ba28ed6a5d9fd6a3a i have replaced the setting

```diff
-allow_multiple_connections_per_peer
+replace_stale_connections
```

for now this is experimental, so default false
but in the future, this should be changed to default true

the beauty is this:

1. seeder is listening for leechers
2. leecher1 connects to seeder
3. leecher1 crashes and cannot cleanup its connection to seeder
4. seeder thinks that leecher1 is still connected
5. leecher2 tries to re-use the listening port of leecher1
6. the operating system detects that leecher1 is no longer using its listening port, so leecher2 can successfully re-use the listening port
7. leecher2 connects to seeder
8. seeder in `peer_list::new_connection` receives a new connection from leecher2, using the exact same connection tuple `(src_ipaddr, src_port, dst_ipaddr, dst_port)` as in the old connection from leecher1, which is not possible in TCP/IP, so the old connection entry from leecher1 must be stale, therefore it is 100% safe to close the old connection from leecher1 and replace it with the new connection from leecher2

this may be an edge case, because all peers are running on localhost
but my hope is that also not-so-edge cases will profit from this change

currently the seeder log shows

```
*** DUPLICATE_PEER [ this: "127.0.0.1" that: "127.0.0.1" ]
*** CONNECTION_CLOSED [ op: 1 ERROR: (libtorrent:220) replaced stale connection ]
```

ideally we should remove the DUPLICATE_PEER message
because in this case, it is a "false duplicate" peer

todo: squash commits (or make a new PR)

PR #8373, opened 2026-05-14
rename setting to allow_multiple_connections_per_peer_id

milahu:
fix https://github.com/arvidn/libtorrent/pull/8215#discussion_r3228341100

not tested, feeling lucky : P

basically this is

```sh
rg -l allow_multiple_connections_per_pid |
xargs sed -i -E 's|(allow_multiple_connections_per)_pid|\1_peer_id|g'

git mv simulation/test_allow_multiple_connections_per_{pid,peer_id}.cpp
```

why?

- make it consistent with the rest of the codebase: `duplicate_peer_id` etc
- `pid` is okay for a local variable, but bad in a user-facing API


include/libtorrent/settings_pack.hpp
```diff
@@ -1053,7 +1053,7 @@ namespace aux {
 			// a peer has multiple IP addresses, enabling this feature
 			// may improve transfer efficiency, but it may also
 			// increase network load.
-			allow_multiple_connections_per_pid,
```
arvidn, line 1056: was this released in the last release from 2.0.x? I was released in 2.1rc1, but that's probably fine to change. This is user facing so if it has been released we need to just add an alias

arvidn:
if the existing name has not been included in a release yet, I think this change is reasonable.

PR #8381, opened 2026-05-15
Handle short reads from part files

M-Hassan-Raza:
Part-file reads could return fewer bytes than requested and still be used by hash/export paths.

This treats short reads as `file_too_short` and adds tests for truncated part files.

### DIFF
--- a/src/part_file.cpp
+++ b/src/part_file.cpp
@@ -71,6 +71,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/aux_/path.hpp"
 #include "libtorrent/aux_/storage_utils.hpp" // for iovec_t
 
+#include <boost/asio/error.hpp>
+
 #include <functional> // for std::function
 #include <cstdint>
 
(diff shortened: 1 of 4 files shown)


src/part_file.cpp
```diff
@@ -79,6 +79,24 @@ namespace {
...
+		, libtorrent::span<char> const buf
+		, std::int64_t const offset, libtorrent::error_code& ec)
+	{
+		int const ret = libtorrent::aux::pread_all(fd, buf, offset, ec);
+		if (ret >= 0 && ret == buf.size()) return ret;
+
+		if (ret >= 0)
+			set_file_too_short(ec);
+		return -1;
+	}
```
arvidn: would it not make more sense to just make `pread_all()` fail with `libtorrent::errors::file_too_short` if it gets a short read?

src/posix_part_file.cpp
```diff
@@ -80,6 +80,12 @@ namespace {
 	// round up to even kilobyte
 	int round_up(int n)
 	{ return (n + 1023) & ~0x3ff; }
+
+	void set_file_too_short(libtorrent::error_code& ec)
+	{
+		ec.assign(libtorrent::errors::file_too_short
+			, libtorrent::libtorrent_category());
+	}
```
arvidn: this function seems a bit unnecessary. I's only a single statement, and only called from one place

M-Hassan-Raza:
@arvidn the requested changes have been made! Hopefully the PR is good to go once the CI settles on green

arvidn (approved):

PR #8384, opened 2026-05-17
Add BEP 39 update-url and originator support

dan0kun:
## Summary

Implements [BEP 39](http://www.bittorrent.org/beps/bep_0039.html) update feed fields for torrents.

- Adds  and  setters to 
- Adds  and  const accessors to 
- Adds  /  internal setters
- Serializes  and  keys into the info dict when non-empty
- Parses both fields from the info dict during torrent loading
- Propagates both fields when constructing  from an existing 

## Test plan

- [ ]  — round-trips  through create/parse
- [ ]  — round-trips  through create/parse
- [ ] Existing  suite passes without regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/torrent_info.cpp
+++ b/src/torrent_info.cpp
@@ -1422,6 +1422,9 @@ namespace {
 					, str.string_length());
 			}
 		}
+
+		m_update_url = info.dict_find_string_value("update-url").to_string();
+		m_originator = info.dict_find_string_value("originator").to_string();
 #endif // TORRENT_DISABLE_MUTABLE_TORRENTS
 
 		if (info.dict_find_string("ssl-cert"))
@@ -1547,6 +1550,12 @@ namespace {
 
 	void torrent_info::internal_set_comment(string_view const s)
 	{ m_comment = std::string(s); }
+
+	void torrent_info::internal_set_update_url(string_view const u)
+	{ m_update_url = std::string(u); }
+
+	void torrent_info::internal_set_originator(string_view const o)
+	{ m_originator = std::string(o); }
 
 	bdecode_node torrent_info::info(char const* key) const
 	{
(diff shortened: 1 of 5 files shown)


arvidn:
is anyone using this feature?

dan0kun:
> is anyone using this feature?

I'm planning to use BEP 39 and BEP 46 in a project, and I thought it made sense to add it to your code, which I was planning to use for that project.

dan0kun:
I didn't review everything Claude modified, although it did pass the tests.

src/torrent_info.cpp
```diff
@@ -1548,6 +1551,12 @@ namespace {
 	void torrent_info::internal_set_comment(string_view const s)
 	{ m_comment = std::string(s); }
 
+	void torrent_info::internal_set_update_url(string_view const u)
+	{ m_update_url = std::string(u); }
+
+	void torrent_info::internal_set_originator(string_view const o)
+	{ m_originator = std::string(o); }
+
```
arvidn: these should be removed

test/test_create_torrent.cpp
```diff
@@ -619,10 +619,26 @@ TORRENT_TEST(creator)
...
+		t.set_originator("cert-bytes");
+	});
+	TEST_EQUAL(info.originator(), "cert-bytes");
+}
+
 TORRENT_TEST(dht_nodes)
 {
 	auto info = test_field([](lt::create_torrent& t){
-		t.add_node({"foobar"s, 1337});
+		t.add_node({std::string("foobar"), 1337});
```
arvidn: this change seems unrelated (and unjustified)

src/torrent_info.cpp
dan0kun: removed setters per review

test/test_create_torrent.cpp
dan0kun: reverted to C++14 explicit string cast

dan0kun:
Updated code per comments.

PR #8388, opened 2026-05-17
Fix peer class bandwidth channel capacity

M-Hassan-Raza:
This fixes a capacity mismatch in the bandwidth limiter peer-class path.

A bandwidth request can collect channels from peer classes, torrent peer classes, and the two global channels, but the request storage was still capped at 10. This makes the peer-class and bandwidth-channel limits share the same contract, fixes the peer-class set capacity check, and adds coverage for the max class/channel case.

Checks run:
git diff --check
pre-commit run --files on the touched files
b2 -q test_bandwidth_limiter test_peer_classes
full Boost.Build deterministic test lane


M-Hassan-Raza:
Also debated the new header a bit, but I think it’s useful because the limit is shared by both peer_class_set and bw_request. Keeping it in one small aux header avoids duplicating the capacity contract or making the bandwidth queue header depend on peer_class_set. Happy to revisit this (or any other part of the pr, as always)

include/libtorrent/aux_/peer_class_limits.hpp
arvidn, line 1: I don't think a new header is justified

include/libtorrent/aux_/bandwidth_queue_entry.hpp
```diff
@@ -66,8 +67,7 @@ struct TORRENT_EXTRA_EXPORT bw_request
 	// from the most limiting one
 	int assign_bandwidth();
 
-	static constexpr int max_bandwidth_channels = 10;
-	// we don't actually support more than 10 channels per peer
+	static constexpr int max_bandwidth_channels = max_peer_class_bandwidth_channels;
```
arvidn: I think 10 is already quite high, this raises it to 32.

src/peer_class_set.cpp
```diff
@@ -41,7 +41,7 @@ namespace libtorrent {
 	{
 		if (std::find(m_class.begin(), m_class.begin() + m_size, c)
 			!= m_class.begin() + m_size) return;
-		if (m_size >= int(m_class.size()) - 1)
```
arvidn, line 44: I think using `m_class.size()` here is better. There's no risk of divergence

src/bandwidth_manager.cpp
```diff
@@ -115,7 +117,10 @@ namespace aux {
 		for (int i = 0; i < num_channels; ++i)
 		{
 			if (chan[i]->need_queueing(blk))
+			{
+				TORRENT_ASSERT(k < bw_request::max_bandwidth_channels);
```
arvidn: I think the correct check here would be `bwr.channel.size()`. But that check is already built-in, especially with `debug-iterators=on`

M-Hassan-Raza:
@arvidn Thanks for the feedback! I've incorporated all of it. Hope its solid now. Removign teh header etc trimmed the LoC by quite a bit which is good, less bloat overall

src/peer_connection.cpp
```diff
@@ -5707,7 +5707,7 @@ namespace libtorrent {
 
 		int const priority = get_priority(channel);
 
-		int const max_channels = num_classes() + (t ? t->num_classes() : 0) + 2;
+		int const max_channels = aux::bw_request::max_bandwidth_channels;
```
arvidn: this looks like a pessimization. It used to be determined at run-time, based on the number of classes in use. Now it's a compile-time constant, presumably the max. Id this is in fact better, we shouldn't be using `alloca` anymore, it could just be a `std::array`. But I think the common case is that there are few channels.

The previous expression could use a comment explaining what it's doing though

arvidn (commented):
looking better

src/peer_connection.cpp
M-Hassan-Raza: Good point. I restored the runtime-sized allocation for the common case, kept it capped to bw_request’s storage limit, and added a short comment explainnig the sizing.

arvidn (approved):

PR #8389, opened 2026-05-17
Fix mmap disk jobs stranded behind fences

M-Hassan-Raza:
Fixes an mmap disk I/O scheduling case where jobs unblocked by a fence could be put back on the generic queue even when no generic disk threads were running.

This could stall configs with aio_threads=0 and hashing_threads>0 after a fenced job, because follow-up work was not always routed to an active queue or executed inline.

The change keeps job queue and thread-pool selection consistent, routes unblocked fence jobs back to the right queue, and executes unblocked jobs inline when their target pool has no threads.

Added a regression test for fenced mmap hashing with zero aio threads.

Tested with test_storage and test_fence.

### DIFF
--- a/src/mmap_disk_io.cpp
+++ b/src/mmap_disk_io.cpp
@@ -1368,7 +1368,7 @@ TORRENT_EXPORT std::unique_ptr<disk_interface> mmap_disk_io_constructor(
 			m_generic_io_jobs.m_queued_jobs.push_back(j);
 			l.unlock();
 
-			if (num_threads() == 0 && user_add)
+			if (pool_for_job(j).max_threads() == 0 && user_add)
 				immediate_execute();
 
 			return;
@@ -1432,9 +1432,12 @@ TORRENT_EXPORT std::unique_ptr<disk_interface> mmap_disk_io_constructor(
 
 	void mmap_disk_io::immediate_execute()
 	{
-		while (!m_generic_io_jobs.m_queued_jobs.empty())
+		for (;;)
 		{
+			std::unique_lock<std::mutex> l(m_job_mutex);
+			if (m_generic_io_jobs.m_queued_jobs.empty()) return;
 			aux::mmap_disk_job* j = m_generic_io_jobs.m_queued_jobs.pop_front();
+			l.unlock();
 			execute_job(j);
 		}
 	}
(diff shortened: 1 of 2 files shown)


test/test_storage.cpp
```diff
@@ -1731,6 +1730,93 @@ void test_unaligned_read(lt::disk_io_constructor_type constructor, Fun fun)
 	disk_io->abort(true);
 }
 
+#if TORRENT_HAVE_MMAP || TORRENT_HAVE_MAP_VIEW_OF_FILE
+TORRENT_TEST(mmap_fenced_hash_zero_aio_threads)
```
arvidn: I think it would be better to parameterize all other tests to include this configuration. Would that be practical you think?

src/mmap_disk_io.cpp
```diff
@@ -1432,9 +1432,12 @@ TORRENT_EXPORT std::unique_ptr<disk_interface> mmap_disk_io_constructor(
 
 	void mmap_disk_io::immediate_execute()
 	{
-		while (!m_generic_io_jobs.m_queued_jobs.empty())
+		for (;;)
 		{
+			std::unique_lock<std::mutex> l(m_job_mutex);
```
arvidn, line 1439: in this case, where we don't have a separate disk thread, are you sure we need to acquire the mutex? Is this because a hasher thread may access this job queue?

arvidn (commented):
this looks correct. I have a feeling there are ways to simplify this a bit though, and maybe use/modify some existing functionality.

test/test_storage.cpp
M-Hassan-Raza: Yeah that’s practical for the mmap check-files path. I locally folded the standalone regression into the existing mmap check_files coverage and run the same tests with aio_threads=0 and hashing_threads=1. Kept it scoped to mmap because the bug is specific to mmap job routing when there are no generic disk threads but hash threads still exist. Good suggsetion!

src/mmap_disk_io.cpp
M-Hassan-Raza: Imo (i could be wrong which is usually the case) Yes, the lock is still needed. With aio_threads=0 there is no generic disk thread, but a hasher thread can still complete work, lower a fence, and touch the generic queue. immediate_execute() can also pop from that queue, so it still needs to serialize with the same job mutex. What do you think?

M-Hassan-Raza:
the Appveyor failures seems unrelated to my changes, their service seemed to time out from what I could see. So that check shouldn't be a blocker for now

test/test_storage.cpp
```diff
@@ -765,6 +767,47 @@ void test_check_files(check_files_flag_t const flags
 		run_until(ios, done);
 	}
 
+	if (aio_threads == 0 && hashing_threads > 0)
+	{
```
arvidn: I think it makes sense to run this test unconditionally

```suggestion
```

arvidn (commented):
looks good, I'd prefer your new test is run for multiple threads as well

M-Hassan-Raza:
@arvidn Thanks for the feedback! Should be solid now hopefully. Please lmk if there's any blocker before merge so that I can remove that as well. Appreciate your time on the review :)

PR #8393, opened 2026-05-18
hash_picker: stop retrying accepted piece-layer hashes

rosesopranodesertbat:
## Summary

`hash_picker::add_hashes()` validates and inserts accepted BEP52 piece-layer hash responses into the file merkle tree, but it did not update the per-512-piece request bookkeeping used by `pick_hashes()`.

That leaves `m_piece_hash_requested[].have` false after a successful response. Once the retry interval expires, `pick_hashes()` may select the same piece-layer range again, causing repeated `HASH_REQUEST` / `HASHES` exchanges with peers that already served valid hashes.

This marks the corresponding piece-layer request bucket complete after a valid response is accepted.

## Test

Added `add_piece_hashes_marks_request_complete` to `test_hash_picker.cpp`. The test makes a previously requested piece-layer range eligible again, accepts a valid hash response, and verifies `pick_hashes()` does not select the same range again.

## Local validation

- `git diff --check`
- `clang++ -std=c++14 -Iinclude -I/opt/homebrew/include -DTORRENT_BUILDING_LIBRARY -DTORRENT_USE_OPENSSL=1 -fsyntax-only src/hash_picker.cpp`
- `clang++ -std=c++14 -Iinclude -Itest -I/opt/homebrew/include -DTORRENT_USE_OPENSSL=1 -fsyntax-only test/test_hash_picker.cpp`

`make check` could not run locally because Boost.Build (`b2`) is not installed in this environment.

### DIFF
--- a/src/hash_picker.cpp
+++ b/src/hash_picker.cpp
@@ -289,6 +289,27 @@ bool validate_hash_request(hash_request const& hr, file_storage const& fs)
 		ret.hash_failed = std::move(results->failed);
 		ret.hash_passed = std::move(results->passed);
 
+		if (req.base == m_piece_layer)
+		{
+			// Piece-layer hash requests are tracked in 512-piece buckets.
+			// Once the response has been validated and inserted into the merkle
+			// tree, mark the bucket as complete so pick_hashes() does not retry
+			// the same range after min_request_interval.
+			TORRENT_ASSERT(req.index % 512 == 0);
+			TORRENT_ASSERT(req.count <= 512);
+			int const request_index = req.index / 512;
+			if (req.index % 512 == 0
+				&& req.count <= 512
+				&& request_index >= 0
+				&& request_index < int(m_piece_hash_requested[req.file].size()))
+			{
+				auto& piece_hash_req = m_piece_hash_requested[req.file][request_index];
+				piece_hash_req.have = true;
+				piece_hash_req.last_request = min_time();
+				piece_hash_req.num_requests = 0;
+			}
+		}
+
 		return ret;
 	}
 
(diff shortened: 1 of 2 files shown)


rosesopranodesertbat:
Additional local validation of the new regression test:

- Built a focused `test_hash_picker` binary manually because Boost.Build (`b2`) is not installed in this environment.
- Rebuilt that binary with `src/hash_picker.cpp` from `HEAD^` and the new test from this PR.
- Result before this patch:

```text
***** test/test_hash_picker.cpp:308 "TEST_ERROR: picked2.count: 512 expected: 0" *****
test/test_hash_picker.cpp.add_piece_hashes_marks_request_complete: FAIL
```

- Result after this patch:

```text
test/test_hash_picker.cpp.add_piece_hashes_marks_request_complete: PASS
```

So the test fails against the old picker behavior and passes with the added bookkeeping update.

src/hash_picker.cpp
```diff
@@ -289,6 +289,27 @@ bool validate_hash_request(hash_request const& hr, file_storage const& fs)
...
+			if (req.index % 512 == 0
+				&& req.count <= 512
+				&& request_index >= 0
+				&& request_index < int(m_piece_hash_requested[req.file].size()))
+			{
+				auto& piece_hash_req = m_piece_hash_requested[req.file][request_index];
+				piece_hash_req.have = true;
+				piece_hash_req.last_request = min_time();
+				piece_hash_req.num_requests = 0;
+			}
```
arvidn, line 310: All of these conditions have already been checked, a few lines above: https://github.com/arvidn/libtorrent/pull/8393/changes#diff-6718543712851f5e6d8903943e592385b6d99081a575e837d9988390687eb6a9R251-R254


```suggestion
			auto& piece_hash_req = m_piece_hash_requested[req.file][request_index];
			piece_hash_req.have = true;
			piece_hash_req.last_request = min_time();
```

arvidn (commented):
looks good

arvidn:
my version of this is merged now

PR #8410, opened 2026-05-22
Match DHT RPC replies by full endpoint

M-Hassan-Raza:
rpc_manager was matching incoming DHT replies by IP address only. That means a packet from the same IP but a different UDP port could consume an outstanding transaction.

This makes incoming() use the same full-endpoint check that unreachable() already uses.

I also added a regression test that sends a reply from the wrong port, checks that it is ignored, then sends the reply from the original endpoint and checks that the transaction completes.

Tests:
b2 test_dht
b2 test_dht warnings-as-errors=on
b2 test_dht warnings-as-errors=on asserts=on invariant-checks=full

### DIFF
--- a/src/kademlia/rpc_manager.cpp
+++ b/src/kademlia/rpc_manager.cpp
@@ -276,7 +276,7 @@ bool rpc_manager::incoming(msg const& m, node_id* id)
 	auto range = m_transactions.equal_range(tid);
 	for (auto i = range.first; i != range.second; ++i)
 	{
-		if (m.addr.address() != i->second->target_addr()) continue;
+		if (m.addr != i->second->target_ep()) continue;
 		o = i->second;
 		i = m_transactions.erase(i);
 		break;
--- a/test/test_dht.cpp
+++ b/test/test_dht.cpp
@@ -187,6 +187,22 @@ void node_from_entry(entry const& e, bdecode_node& l)
 	TEST_CHECK(ret == 0);
 }
 
+bool incoming_rpc_message(dht::rpc_manager& rpc, entry const& e
+	, udp::endpoint const& source, node_id* id)
+{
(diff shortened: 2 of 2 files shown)


src/kademlia/rpc_manager.cpp
```diff
@@ -276,7 +276,7 @@ bool rpc_manager::incoming(msg const& m, node_id* id)
 	auto range = m_transactions.equal_range(tid);
 	for (auto i = range.first; i != range.second; ++i)
 	{
-		if (m.addr.address() != i->second->target_addr()) continue;
+		if (m.addr != i->second->target_ep()) continue;
```
arvidn, line 279: is it safe to assume that all (or the vast majority) of nodes in the wild respond on the same socket? I suppose there's a strong incentive to do so, to keep NAT pinholes open. But there's a small risk associated with this change.

I wonder if it makes sense to have an option for this. If a new setting turns out to be a lot of changes, maybe it's not worth it.

M-Hassan-Raza: Hmmm that's a pretty good point. I don't have hard numbers from the wild, so I shouldn't claim this is universally safe.

My reasoning was that the transactoin is sent to a specific DHT endpoint, and the rest of this path already treats the peer as address+port. `unreachable()` also matches `target_ep()`. If a reply from the same IP but another port can complete the transaction, that other port can consume it before the original endpoint replies.

But still I agree this has some compatibility risk. I don't love adding a public setting for such a narrow RPC matching rule unless you think the risk is real enough. If you prefer a setting, I can add one and default it however you think makes sense.

I think whatever shape we agree to land on for this pr in the end would most likely be a net positive for the core overall.

arvidn (approved):

PR #8414, opened 2026-05-24
I2p flag

arvidn:
perhaps this is better targeting `master`.

The idea is to simplify the control of i2p torrents with a single flag `allow_i2p_peers`, instead of the `allow_i2p_mixed` settings and the `i2p_torrent` flag.

### DIFF
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -1,0 +1,13 @@
+# Auto-generated source files; do not reformat them.
+# These are produced by code-generation scripts and must match the generator
+# output exactly, so clang-format must leave them untouched. This file is
+# honored by a direct `clang-format` / `clang-format -i` run (e.g. from an
+# editor). NOTE: git-clang-format does NOT understand .clang-format-ignore --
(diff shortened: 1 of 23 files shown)


kiragakiyumi:
It would be great if these changes could still make it into 2.1.x, if you think that’s feasible. I understand you mentioned that this might be better targeted at master, though.

glassez:
> The idea is to simplify the control of i2p torrents with a single flag `allow_i2p_peers`, instead of the `allow_i2p_mixed` settings and the `i2p_torrent` flag.

It seems that this goes against users’ expectations regarding the use of I2P. Rather, they may or may not need to prohibit regular peers when using I2P.

kiragakiyumi:
@glassez, do you mean two modes for I2P torrents: Mixed Mode, where both I2P and regular peers are allowed through the I2P network, and I2P-only mode, where only I2P peers are allowed to connect through the I2P network?

If so, that does indeed seem like a more correct approach.

PR #8423, opened 2026-05-27
Tighten HTTP tracker peer parsing

M-Hassan-Raza:
Reject malformed compact tracker peer lists instead of silently dropping trailing bytes.

Reject dictionary peer ports outside the uint16 range before they can wrap into valid endpoints.

Add focused tracker parser coverage for IPv4, IPv6, I2P, and dictionary peer ports.

Validated with git diff --check, pre-commit for touched files, Boost.Build test_tracker build, and focused parser test_tracker cases.

### DIFF
--- a/src/http_tracker_connection.cpp
+++ b/src/http_tracker_connection.cpp
@@ -46,7 +46,7 @@ namespace libtorrent::aux {
 		bool valid_compact_peer_list(int const len, int const entry_size)
 		{
 			TORRENT_ASSERT(entry_size > 0);
-			return len % entry_size == 0;
+			return len >= 0 && len % entry_size == 0;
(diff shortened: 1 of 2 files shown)


arvidn:
since this is not really a bugfix, but improvement in test coverage and stricter parsing, I would prefer this PR be made against `RC_2_1`

M-Hassan-Raza:
> since this is not really a bugfix, but improvement in test coverage and stricter parsing, I would prefer this PR be made against `RC_2_1`

Rebased against 2_1, some code had to be changed to accommodate for 2_1's not insignificantly different state. Please review whenever you get the time.

arvidn (approved):

PR #8434, opened 2026-05-30
Fix DH xor mask secret padding

M-Hassan-Raza:
Fixes DH shared secret normalization before deriving the req3 xor mask.

The shared secret is now exported through the existing 96-byte padded contract before hashing. The outgoing handshake path uses the same helper, and the test covers a short-secret padding vector.

### DIFF
--- a/src/pe_crypto.cpp
+++ b/src/pe_crypto.cpp
@@ -124,8 +124,7 @@ namespace libtorrent {
 		// shared_secret = (remote_pubkey ^ local_secret) % prime
 		m_dh_shared_secret = mp::powm(remote_pubkey, m_dh_local_secret, dh_prime);
 
-		std::array<char, 96> buffer;
-		mp::export_bits(m_dh_shared_secret, reinterpret_cast<std::uint8_t*>(buffer.data()), 8);
+		std::array<char, 96> const buffer = export_key(m_dh_shared_secret);
 
 		static char const req3[4] = {'r', 'e', 'q', '3'};
 		// calculate the xor mask for the obfuscated hash


arvidn:
it's not obvious to me if this is fixing a bug or cleaning up. I don't see any behavior change

M-Hassan-Raza:
> it's not obvious to me if this is fixing a bug or cleaning up. I don't see any behavior change

yeah good question, my bad the pr body wasn't thoroguh enough. I'll try my best to illustrate/explain below. The helper part is cleanup, but the actual behavior change is in compute_secret().

Before this, that code called export_bits() straight into a 96 byte array and then hashed the whole thing. export_bits() only writes the significant bytes though. so if the DH secret needs fewer than 96 bytes, the rest of the array is left as whatever was already there, and the value also isn’t left-padded into the fixed-width PE representation.

the rest of the PE code uses export_key() for DH values, which does that 96-byte normalization. e.g. secret = 1 should be hashed as:

req3 || 95 zero bytes || 0x01

but the old compute_secret() path hashed roughly:

req3 || 0x01 || uninitialized trailing bytes

that changes the xor mask used by get_hash_xor_mask(), so an incoming encrypted handshake can fail to match the torrent when the shared secret serializes shorter than 96 bytes.

so yeah, the actual fix is using export_key() before deriving the req3 mask. the bt_peer_connection bit is just deduping the same req3 formula on the outgoing path.

Does it make sense? It seems reasonable to me but i could be missing smth as well :D

src/pe_crypto.cpp
```diff
@@ -124,12 +133,10 @@ namespace libtorrent {
 		// shared_secret = (remote_pubkey ^ local_secret) % prime
 		m_dh_shared_secret = mp::powm(remote_pubkey, m_dh_local_secret, dh_prime);
 
-		std::array<char, 96> buffer;
-		mp::export_bits(m_dh_shared_secret, reinterpret_cast<std::uint8_t*>(buffer.data()), 8);
```
arvidn, line 128: As far as I can tell, this is the only change required to fix it. I'm hesitant to land the other changes

M-Hassan-Raza: Should be solid now! Trimmed the bloat

arvidn (approved):

PR #8436, opened 2026-05-31
Fix whole-piece prefetch overflow

M-Hassan-Raza:
Fixes signed overflow when sizing whole-piece prefetch requests.

The download rate window is now computed in 64-bit before applying the existing 8 MiB clamp.

Validated with test//test_piece_picker and pre-commit for the touched file.

### DIFF
--- a/src/request_blocks.cpp
+++ b/src/request_blocks.cpp
@@ -41,6 +41,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/aux_/alert_manager.hpp"
 #include "libtorrent/aux_/has_block.hpp"
 
+#include <cstdint>
 #include <vector>
 
 namespace libtorrent {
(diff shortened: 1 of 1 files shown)


src/request_blocks.cpp
```diff
@@ -112,10 +113,11 @@ namespace libtorrent {
 		interesting_pieces.reserve(100);
 
 		int prefer_contiguous_blocks = c.prefer_contiguous_blocks();
+		int const whole_pieces_threshold = t.settings().get_int(settings_pack::whole_pieces_threshold);
```
arvidn: the one thing I'm a little bit nervous about is that `aux::session_settings` is thread-safe, and has an internal mutex. With this change we always acquire that mutex to ask for `whole_pieces_threshold`, even when `prefer_contiguous_blocks != 0` or `time_critical_mode`.

It probably doesn't matter much in practice, and in the case where we enter the block, we acquire the mutex twice anyway (in the existing code)

M-Hassan-Raza: Yep, fair point. I pulled it into a local to avoid reading the setting twice and keep the multiply using the same value, but yeah, doing that before the gate was a bit too eager.

Changed it so whole_pieces_threshold is only read after the existing prefer_contiguous_blocks == 0 && !time_critical_mode check, then reused for the 64-bit multiply. So same overflow fix, but no extra settings mutex hit on paths that don’t need whole-piece prefetching.

arvidn (approved):

PR #8441, opened 2026-06-01
Fix SOCKS5 auth request bounds

M-Hassan-Raza:
Truncate SOCKS5 username/password values to the protocol's one-byte length limit before building auth requests.

TCP SOCKS5 now clamps credentials in set_username(). UDP proxy settings clamp credentials when they are applied, and the UDP scratch buffer is large enough for the max valid auth request.

Validation:
b2 -j8 cxxflags=-I/Users/smd/personal-work/libtorrent/include test//test_udp_socket
b2 -j8 undefined-sanitizer=norecover cxxflags=-I/Users/smd/personal-work/libtorrent/include test//test_udp_socket
SKIP=flake8-pyi-install,flake8-pyi-run pre-commit run --files include/libtorrent/socks5_stream.hpp src/udp_socket.cpp

### DIFF
--- a/include/libtorrent/socks5_stream.hpp
+++ b/include/libtorrent/socks5_stream.hpp
@@ -123,6 +123,8 @@ public:
 	{
 		m_user = user;
 		m_password = password;
+		if (m_user.size() > 255) m_user.resize(255);
+		if (m_password.size() > 255) m_password.resize(255);
 	}
 
 	void set_dst_name(std::string const& host)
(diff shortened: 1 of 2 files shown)


arvidn (commented):
I would prefer a much simpler approach. e.g. `set_username()` and `set_proxy_settings()` could just truncate username and password in case they are too long.

M-Hassan-Raza:
> I would prefer a much simpler approach. e.g. `set_username()` and `set_proxy_settings()` could just truncate username and password in case they are too long.

Trimmed down the PR size significantly, should be good to go now :)

arvidn (approved):

PR #8451, opened 2026-06-05
Validate resume files touched by completed pieces

hussainarslan:
What
Updates resume validation to check every wanted, non-pad file that overlaps a completed resume piece, instead of only checking the first file mapped from each completed piece.

Adds regression coverage for a completed piece spanning two files where the second file is missing.

Why
Fastresume data can mark a piece complete even when that piece spans multiple files. The previous validation only checked the first mapped file for each completed piece, so a missing later file in the same piece could be accepted and the torrent could resume from stale or incomplete data instead of falling back to a check.

How
The non-seed resume validation now walks the torrent files and stats only wanted, non-pad files that overlap at least one set bit in have_pieces. It uses aux::file_piece_range_inclusive() for the file overlap bounds and scans the resume bitfield in that range.

The regression builds two half-piece files, marks the spanning piece complete, creates only the first file, and asserts resume validation rejects the resume data with a missing-file error for the second file.

Validation
git diff --check origin/RC_2_0
b2 -a crypto=openssl warnings=off test//test_storage

### DIFF
--- a/src/storage_utils.cpp
+++ b/src/storage_utils.cpp
@@ -52,6 +52,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #endif
 
 #include <set>
+#include <tuple>
 
 namespace libtorrent { namespace aux {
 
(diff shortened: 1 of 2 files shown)


src/storage_utils.cpp
```diff
@@ -430,6 +430,25 @@ std::int64_t get_filesize(stat_cache& stat, file_index_t const file_index
...
+{
+	if (fs.file_size(file_index) == 0) return false;
+
+	peer_request const first = fs.map_file(file_index, 0, 1);
+	peer_request const last = fs.map_file(file_index
+		, fs.file_size(file_index) - 1, 1);
+
+	piece_index_t const end_piece = std::min(next(last.piece)
+		, rd.have_pieces.end_index());
+	for (piece_index_t piece = first.piece; piece < end_piece; ++piece)
```
arvidn: There's a function [here](https://github.com/arvidn/libtorrent/blob/RC_2_0/include/libtorrent/file_storage.hpp#L736) called `file_piece_range_inclusive()` that returns the range of pieces that overlap the specified file. I think using that function would make this a bit easier to read. In fact, maybe the whole `file_has_completed_resume_piece()` can be inlined back into where it's called from

arvidn (commented):
looks good. Since there's no unit test for the new function anyway, as it becomes smaller, I think the code would be simpler to return the logic into the loop body where it's called right now.

Also, the web server fix is not controversial and I would have landed that immediately if it was a separate PR

hussainarslan:
> Also, the web server fix is not controversial and I would have landed that immediately if it was a separate PR

No worries in that regard, moved that to the PR #8454 

Also applied the requested fixes

arvidn (approved):

PR #8458, opened 2026-06-07
Fix Merkle proof cleanup

M-Hassan-Raza:
Clear hashes inserted by failed short Merkle proofs without touching existing tree state.

### DIFF
--- a/src/merkle.cpp
+++ b/src/merkle.cpp
@@ -327,6 +327,7 @@ namespace libtorrent {
 
 		int cursor = target_node_idx;
 		target_tree[cursor] = node;
+		bool cursor_inserted = false;
 		for (auto const& proof : uncle_hashes)
 		{
 			int const proof_idx = merkle_get_sibling(cursor);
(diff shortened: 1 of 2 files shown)


src/merkle.cpp
arvidn, line 353:
```suggestion
		while (clear_cursor >= cursor)
```

This would be an alternative fix, right? The main difference is that we write zeroes to the slot that we just checked t0 already be zeros, but I think it's keeps the algorithm a lot easier to follow

arvidn: actually, maybe this won't work. if `cursor` is 0, `merkle_get_parent()` probably fails.

But the idea that we don't need `bool cursor_inserted`, but it's fine to clear `target_tree[cursor]` unconditionally still holds. I think it would be simpler

arvidn:
thanks for the patch! I'm landing a variant of this

M-Hassan-Raza:
> thanks for the patch! I'm landing a variant of this

Thanks for taking it forward! I was super busy all day (on the job) and I read your comment but haven't had time to sit down and actually review the code again to respond to your pr comment and only would've gotten to it tomorrow at the earliest. You can land the patch in any shape you feel confident in, this pr can be closed and referenced later if needed!

PR #8466, opened 2026-06-10
harden merkle_tree proof validation

arvidn:
follow-up on https://github.com/arvidn/libtorrent/pull/8461


qBittUser:
Does PR and commit title have typo: `markle_tree` instead of `merkle_tree`?

PR #8472, opened 2026-06-12
Test merge

arvidn:


glassez:
Well, on the one hand, I am glad that libtorrent has begun to actively develop again after a very long stagnation. But still, I would like to know what are the forecasts for the v2.1 release? Are there any known bugs that prevent it?

arvidn:
my main focus is on completing the pread disk I/O. i expect to make another RC this weekend. it's in feature freeze (so bug fixes are still going in)

PR #8479, opened 2026-06-13
simplify flush_storage() by not making it erase the piece

arvidn:
just flush dirty buffers. this makes pending_free_flag unnecessary which takes some logic with it as it goes


qBittUser:
Typo in PR & commit title: `makign` instead of `making`.

PR #8489, opened 2026-06-19
`torrent_info` is immutable and can be shared

arvidn:


include/libtorrent/aux_/peer_connection.hpp
```diff
@@ -442,6 +442,17 @@ namespace libtorrent::aux {
 		std::weak_ptr<aux::torrent> associated_torrent() const
 		{ return m_torrent; }
 
+		// the cached pointer to the associated torrent_info, used to
+		// resolve info-hash, piece info, file storage, etc. without
+		// reaching back through associated_torrent().lock(). null
+		// before the peer has attached to a torrent.
+		std::shared_ptr<torrent_info const> const& torrent_file() const { return m_ti; }
```
glassez, line 449: Torrent file and torrent info (aka info section) don't have a one-to-one correspondence, so I wouldn't call it "torrent file".

arvidn: yeah I agree that it's not the best name, however, this is the term used in `torrent` (`torrent::get_torrent_file()` and `torrent::torrent_file()`).

glassez: And it's been looking inappropriate for a long time. Why always follow a previously made mistake and breed it more and more? Shouldn't we clean up the codebase from time to time?

arvidn: cleanups are fine, but ideally not mixed in with other changes. sometimes future merge conflicts should be taken into account too.

PR #8538, opened 2026-07-05
Tracker pipelining

arvidn:
reuse tracker connections for announces to the same endpoint. Don't pipeline requests, since trackers are likely to not support that. Except when shutting down, pipeline maximally to speed up shutdown.

This is my current plan with regards to reusing tracker connections:

* tracker connections are *reused* but requests are not *pipelined*. i.e. requests are still made one at at time but over the same connection. This saves the TCP and SSL handshake round-trip (which is not cheap), but it doesn't maximize throughput. The reasoning is that (I believe) it to be common to not support pipelining, or at least not very deep pipeline depth. This keeps the requests separate and it's simple to know which ones were made and which ones need to re-connect.
* At shutdown, tracker `event=stop` announces are pipelined maximally. i.e. all stop announcements are sent immediately, and we don't even wait for the response. The reasoning is that it's mostly in the interest of the tracker that we announce `event=stop`, and when the client is shutting down, we want to minimize the waiting for trackers.


sheepish-ops:
I'd love to see this feature, thank you for tackling this. TLS overhead really is a big problem, at least on the tracker side.

Some thoughts from a PT operator and ocelot dev:
* regarding support for pipelining/http2: heavy pipelining is problem because of HOL blocking, but http2 multiplexing on the other hand should work well in most cases. All PT I know of run their tracker behind a reverse proxy such as nginx/envoy/haproxy/cloudflare. ocelot doesn't even natively support HTTPS. It is reasonable to assume that such technology is supported by a tracker, at least as long as the operators care.
* Using a single connection with reuse only can be a huge bottleneck. Users are seeding thousands of torrents, sometimes tens of thousands of torrents. A single connection will stall significantly in such cases. The least minimum in that scenario should be a connection pool in my opinion, possibly scaled to one connection per 500/1000 torrents and tracker target? Just some quick math: assuming a response time of 100ms for an announce (including latency and everything; overseas connections are common), we'll be looking at 10 announces per second. With 10000 active torrents it would require more than 15 minutes to announce each of those. Further assuming an announce interval of 30 minutes, the queue would be fully saturated at least 50% of the time, and that is if only there are no network irregularities.

arvidn:
thanks for your feedback. The main problem with head of line blocking is really if the server takes a long time with some requests, the cheap requests then have to wait. Especially when it's interactive traffic. I don't think it applies so much in this case. Unless you're referring to problems with TCP congestion control for high-latency and high-throughput links, but I don't really see that applying to tracker connections either.

However, your second point is really good. I would probably attribute the problem, not to a single connection, but to not actually pipelining the requests. If trackers generally allow a pipeline depth of 10 requests, and libtorrent would take advantage of that, your scenario would turn into 100 announces per second (assuming the tracker actually supports concurrent lookups from the same connection).

Perhaps the simplest implementation would be to create another connection when the queue grows too long on one. However, that also opens up ordering issues, announces for the same torrent better make it into the same connection, to be ordered correctly.

arvidn:
@sheepish-ops do you think most trackers would support pipelining for `event=stopped` according to this change? i.e. on shutdown, libtorrent would send *all* `event=stopped` announces down a single connection, and then just drain the receive channel of the socket, ignoring the responses.

It would mean, potentially, thousands of pipelined announces. On the other hand, trackers don't really need to perform any lookup before responding to the peer, so it could be very efficient.

sheepish-ops:
I don't currently have time to test this, but I assume that degree of pipelining may cause problems . Reverse proxies will likely limit the number of pipelined requests and close connections if there are too many, as a basic DOS prevention. Even if HOL blocking isn't a major concern for bittorrent traffic in general, the reverse proxies still need to assume they potentially have buffer hundreds of responses if they come in out of order. And request buffers may just overflow, too, on both reverse proxies or trackers, causing high TCP drop rates.

> Perhaps the simplest implementation would be to create another connection when the queue grows too long on one. 

That sounds elegant. But you're right to worry about sending announces out of order. That can cause grief. qBittorrent sometimes does this and it can really screw with stats accounting. Announces don't have sequence numbers after all. In that regard, for private torrents there should ideally only be a single queue, even if the torrent has multiple tracker targets. I don't know how libtorrent handles dual stack, but receiving ipv4/ipv6 announces out of order could become a problem.

arvidn:
I think this patch is in a good state now. Since it's a fairly significant new feature, with some risk associated with it, it will land in `master`.

I'd appreciate if anyone could test this to make sure it works in practice, without any issues.

The main risk, I think, is how this models web servers. The limit is not pipeline queue depth, but the total number of requests sent over a single connection. I believe that's what nginx limits. However, this also only matters on shutdown, when `event=stopped` announces are actually pipelined. Otherwise announces are reusing the connection, but not pipelining.

seabashed:
@arvidn thank you for implementing this. I have been running master with qBittorrent for some time now and it is beautiful to see connection reuse in practice!

PR #8565, opened 2026-07-08
Optimize CPU usage in the DH key exchange hot path

anthonyryan1:
This is a two commit series focused on reducing CPU usage during "connection storms" where a large number of peers are connecting at the same time. 

My measurements show this reduces CPU usage by 30% for all users, and 35% for users where libtorrent configured with openssl. Improvements translated into both qBittorrent and Deluge (as testable downstream consumers of this library).

Feedback welcome!

### DIFF
--- a/src/pe_crypto.cpp
+++ b/src/pe_crypto.cpp
@@ -92,7 +92,7 @@ namespace libtorrent {
 	// Set the prime P and the generator, generate local public key
 	dh_key_exchange::dh_key_exchange()
 	{
-		aux::array<std::uint8_t, 96> random_key;
+		aux::array<std::uint8_t, 20> random_key;
 		aux::random_bytes({reinterpret_cast<char*>(random_key.data())
 			, static_cast<std::ptrdiff_t>(random_key.size())});
 


src/pe_crypto.cpp
```diff
@@ -92,15 +147,19 @@ namespace libtorrent {
...
 	dh_key_exchange::dh_key_exchange()
 	{
-		aux::array<std::uint8_t, 96> random_key;
+		// MSE spec: "Minimum length [for the private exponent] is 128 bit.
+		// Anything beyond 180 bit is not believed to add any further
+		// security and only increases the necessary calculation time.
+		// You should use a length of 160bits whenever possible[.]"
+		aux::array<std::uint8_t, 20> random_key;
 		aux::random_bytes({reinterpret_cast<char*>(random_key.data())
 			, static_cast<std::ptrdiff_t>(random_key.size())});
```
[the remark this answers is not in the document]
anthonyryan1: It's technically correct that the random number generator could give us enough leading zeroes to that the exponent could come in under 128 bit. We have a 1 in 2^(160-128) probability of that happening which is large but will definitely happen in practice.

We still have 160 bits of entropy because those leading zeros were random. And we lose entropy by clamping a bit. I'm not certain this is as clean a win as the robot suggests.

Would value a human second opinion on this one.

src/pe_crypto.cpp
```diff
@@ -60,6 +68,53 @@ namespace libtorrent {
...
+				&& bn_result != nullptr && ctx != nullptr
+				&& BN_mod_exp(bn_result, bn_base, bn_exp, bn_prime, ctx) == 1)
+			{
+				int const len = BN_num_bytes(bn_result);
+				BN_bn2bin(bn_result, key_bytes.data());
+				mp::import_bits(result, key_bytes.data(), key_bytes.data() + len);
+			}
+			else
+			{
+				// avoid dereferencing a null BIGNUM if any allocation failed
```
arvidn: what does this comment mean?

src/pe_crypto.cpp
```diff
@@ -60,6 +68,53 @@ namespace libtorrent {
...
+			BN_CTX* ctx = BN_CTX_new();
+
+			key_t result;
+			if (bn_base != nullptr && bn_exp != nullptr && bn_prime != nullptr
+				&& bn_result != nullptr && ctx != nullptr
+				&& BN_mod_exp(bn_result, bn_base, bn_exp, bn_prime, ctx) == 1)
+			{
+				int const len = BN_num_bytes(bn_result);
+				BN_bn2bin(bn_result, key_bytes.data());
+				mp::import_bits(result, key_bytes.data(), key_bytes.data() + len);
```
arvidn: if this function throws an exception, you're leaking memory

src/pe_crypto.cpp
```diff
@@ -60,6 +68,53 @@ namespace libtorrent {
...
+				std::uint8_t* const begin = key_bytes.data();
+				std::uint8_t* const end = mp::export_bits(v, begin, 8);
+				return BN_bin2bn(begin, int(end - begin), nullptr);
+			};
+
+			BIGNUM* bn_base = to_bn(base);
+			BIGNUM* bn_exp = to_bn(exp);
+			BIGNUM* bn_prime = to_bn(dh_prime);
+			BIGNUM* bn_result = BN_new();
+			BN_CTX* ctx = BN_CTX_new();
```
arvidn: this is an awful lot of heap allocations, per operation

arvidn (commented):
I appreciate the idea of using libcrypto to speed up this operation. but you need to work in this patch to make it palatable.

anthonyryan1:
Your findings are all correct, and I've spent about 2 hours iterating on the OpenSSL offload and I've decided I'm on the wrong end of the pareto principle with that commit. I've pulled the OpenSSL commit from this PR.

We get 85% of the benefit of the 2 commit series, with a single line change as opposed to the messy OpenSSL addition being part of this change.

I did find a lead on potentially using the same "Montgomery modular exponentiation" in boost multi-precision using the optional GMP backend which may be a cleaner approach. With an configure flag that could swap cpp_int for mpz_int at compile time, but I'd rather not get the majority of the benefit lost trying to land technical details for an additional 15% speedup when the 85% is right here in one line.

I'll follow up if/when I find a cleaner approach I'm happy with for the second optimization.

src/pe_crypto.cpp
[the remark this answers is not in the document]
arvidn: i agree with you. 0 is a valid random bit. i think the comment is a bit excessive though. does it even need a comment?
it makes sense with an explanation (and reference) in the PR, justifying that 160 bits are enough, but not in the code

arvidn:
would you be able to share your measurements of before and after this change?

anthonyryan1:
I have amended the commit to remove the comment.

-----

I have a somewhat unsatisfying answer to that.

To find a client seeding to an encrypted storm of incoming leechers I screened a large number of qbittorrent and deluge instances for CPU outliers, and then ran perf analysis (attached) on one of outliers to understand why it was was taking up such a disproportionate amount of CPU time, compared to an median instance.

I've attached the perf output for one of the CPU outliers that I observed, prior to this change.

Lines 32, 112, 450, and 517 all sum up together to show compute_secret() totaling 36.9% of the CPU time in the outlier instance.

After I deployed (the two commit version) across all instances, I no longer observed qbittorrent and deluge CPU outliers. The elimination of these heavy outliers resulted in a very substantial decrease in the mean qbittorrent and deluge CPU time (both greater than 50%). By contrast, the median deluge and qbittorrent CPU time barely changed. This is exclusively an outlier problem, but it shows up quite visibly with enough busy instances.

The loss out outliers though, means I can no longer identify instances in the middle of an encrypted connection storm to profile them, and all profiles I gather are of clients experiencing a different set of network circumstance, which would not be a fair comparison.

I'm sorry this isn't what you asked for. I've also attached a benchmark showing how compute_secret() is speed up by ~5x by this one line change. My hope is that taking this together with the perf report is enough to justify my interest in optimizing this particular code path.

[perf-analysis.txt](https://github.com/user-attachments/files/29861503/perf-analysis.txt)
[exp_width_benchmark.cpp](https://github.com/user-attachments/files/29862119/exp_width_benchmark.cpp)

arvidn (approved):

PR #8574, opened 2026-07-09
Add alert for internally banned IPs

101Dude:
This adds `ip_ban_alert`, emitted when `session_impl::ban_ip()` bans an IP address.

The alert exposes the banned address so client applications can distinguish libtorrent-initiated IP bans from ordinary IP-filter matches.

Behavior impact:
- diagnostic alert only
- no change to banning behavior
- no change to IP filter behavior
- no change to peer selection or DHT behavior

Tested:
- `cmake -S . -B /private/tmp/libtorrent-ip-ban-rc21-build -G Ninja -Dbuild_tests=ON -Dbuild_examples=OFF -Dbuild_tools=OFF -Dpython-bindings=OFF -DBUILD_SHARED_LIBS=OFF -Dwebtorrent=OFF -DCMAKE_BUILD_TYPE=Debug`
- `cmake --build /private/tmp/libtorrent-ip-ban-rc21-build --target test_alert_types`
- `/private/tmp/libtorrent-ip-ban-rc21-build/test/test_alert_types`


ChangeLog
```diff
@@ -7,6 +7,7 @@
 	* add option to not validate v1 hashes for hybrid torrents
 	* add post_file_status() to query them asynchronously
 	* add post_file_priorities() to query them asynchronously
+	* add ip_ban_alert for IPs banned internally by libtorrent
```
arvidn: this should go above the 2.1.0 release line, since it will be in 2.1.1

include/libtorrent/alert_types.hpp
```diff
@@ -3285,6 +3285,21 @@ TORRENT_VERSION_NAMESPACE_4_END
 		std::vector<open_file_state> state;
 	};
 
+	// This alert is generated when libtorrent internally bans an IP address.
+	// The banned IP address can be accessed through the ``banned_address`` member.
+	struct TORRENT_EXPORT ip_ban_alert final : alert
```
arvidn, line 3290: did you consider using the `peer_ban_alert`?
If there's a good reason to use a new alert, you should also add python bindings for it. see bindings/python/src/alert.cpp

arvidn (dismissed):

include/libtorrent/alert_types.hpp
101Dude: Thanks. I moved the ChangeLog entry above the 2.1.0 release section and added the Python binding.

I kept `ip_ban_alert` separate from `peer_ban_alert` because `peer_ban_alert` is a `peer_alert` with torrent/peer context and is documented for peers banned for sending corrupt pieces. This path is `session_impl::ban_ip(address)`, which only has an IP address and represents a session-level IP ban, so reusing `peer_ban_alert` would require inventing or omitting peer/torrent context.

I also kept the alert minimal because `ban_ip(address)` only carries the address at this point. If you prefer a reason enum or a different shape for the alert, I’m happy to adjust it.

101Dude:
I reproduced the pre-commit failure locally with `prek`.

`gen_alert_header` rewrites `bindings/c/include/libtorrent_alerts.h`, and `git-clang-format` also touches the same generated header. After running the hooks locally, my working tree is clean, so I don’t have a stable diff to commit.

Should this generated C header be committed in the raw `gen_alert_header` output style, or in the clang-formatted style?

arvidn:
i saw this and addressed it in RC_2_1. i think it will resolve if you rebase

Issue #8583, opened 2026-07-11
suggestion: add modern algo as obfuscation method


trim21:
# Add AES-CTR as MSE stream cipher alongside RC4

## Motivation

RC4 has no hardware acceleration. On modern CPUs, AES-NI is dramatically faster. Measured on AMD Ryzen 7 255 (Zen 5, `vaes` + `avx512`):

| Implementation       | Throughput (single core) |
| -------------------- | ------------------------ |
| RC4 asm (current)    | ~900 MB/s                |
| AES-128-CTR (AES-NI) | ~13,500 MB/s             |

~15× throughput improvement. Even on older AES-NI-only CPUs (without VAES), the gap remains 3-5×.

At 100 MB/s download + 10 MB/s upload, the CPU cost of encryption alone:

| Cipher  | Download (100 MB/s) | Upload (10 MB/s) | Total     |
| ------- | ------------------- | ---------------- | --------- |
| RC4 asm | 11.1% of one core   | 1.1%             | **12.2%** |
| AES-CTR | 0.74%               | 0.07%            | **0.81%** |

AES-CTR saves ~11.4% single-core CPU at these speeds, effectively eliminating encryption as a visible cost.

The change is limited to MSE stream encryption only — the peer wire protocol is untouched.

## Background

MSE uses RC4 purely as obfuscation, not for confidentiality or authentication. Since RC4 offers no meaningful cryptographic security here, replacing it with AES-CTR — which is both faster (via AES-NI) and harder to crack — is a straightforward win.

### Interface

`rc4_handler` already implements the `crypto_plugin` interface ([extensions.hpp](include/libtorrent/extensions.hpp:514)):

```cpp
struct crypto_plugin {
    virtual void set_incoming_key(span<char const> key) = 0;
    virtual void set_outgoing_key(span<char const> key) = 0;
    virtual std::tuple<int, span<span<char const>>> encrypt(span<span<char>>) = 0;
    virtual std::tuple<int, int, int> decrypt(span<span<char>>) = 0;
};
```

A new `aes_ctr_handler` implements the same interface — same call sites in `bt_peer_connection.cpp`, no control flow changes.

### Why no IV in the handshake

RC4 has no IV. MSE sets up a fresh key per connection (via DH), so each key is unique. AES-CTR's uniqueness requirement is "same key + same nonce must not repeat." Since the key itself is per-connection unique, the nonce can be derived from key material without exchanging any additional bytes over the wire.

### Key mapping

[...]

PR #8585, opened 2026-07-12
[POC]: add AES-128-CTR as MSE stream cipher alongside RC4

trim21:
This is a **proof of concept** of https://github.com/arvidn/libtorrent/issues/8583. It works end-to-end (roundtrip test passes, crypto negotiation wired up), but some details need discussion before landing.

### Changes

- `pe_crypto.hpp`: add `aes_ctr_handler` implementing `crypto_plugin`, pimpl struct keeps OpenSSL out of header
- `pe_crypto.cpp`: AES-CTR via OpenSSL EVP (`EVP_aes_128_ctr`), internal counter/keystream buffering handled by EVP
- `settings_pack.hpp`: add `pe_aes_ctr = 4` crypto bit, `prefer_aes_ctr` bool setting
- `bt_peer_connection.cpp`: init AES-CTR handler, negotiate via existing `crypto_provide`/`crypto_select` — AES-CTR preferred over RC4 when both peers support it
- `test_pe_crypto.cpp`: roundtrip encrypt/decrypt test for AES-CTR handler
- `session_settings.cpp`: Python bindings for `prefer_aes_ctr`

### Performance

AES-CTR delivers ~15× throughput vs RC4 on AES-NI hardware (~13.5 GB/s vs ~900 MB/s single-core), eliminating encryption CPU cost at typical BT speeds.

### Key / nonce mapping

SHA-1 outputs 20 bytes → 16-byte AES-128 key + 4-byte nonce. Same SHA-1 key derivation as RC4, no extra exchange.

### Open questions

- **crypto bit value**: currently `pe_aes_ctr = 4` (bit 2). Should be discussed — other implementations may want a different bit assignment.
- **prefer_rc4**: should `prefer_rc4` be exposed alongside `prefer_aes_ctr` for symmetry?

### Backward compatibility

Unknown crypto bits are silently masked away by older clients — a pre-AES-CTR peer seeing bit 4 in `crypto_provide` will ignore it and fall back to RC4. No wire format changes, no version negotiation needed.


src/bt_peer_connection.cpp
```diff
@@ -499,7 +543,11 @@ namespace {
 #if !defined TORRENT_DISABLE_ENCRYPTION
 		if (m_encrypted)
 		{
-			p.flags |= m_rc4_encrypted
+			p.flags |= (m_rc4_encrypted
+#if TORRENT_HAS_MSE_AES_CTR
+						   || m_aes_ctr_encrypted
+#endif
+						   )
 				? peer_info::rc4_encrypted
```
arvidn, line 551: it seems like this would warrant a separate `peer_info` flag, no?

src/bt_peer_connection.cpp
```diff
@@ -618,10 +666,13 @@ namespace {
 		std::memcpy(ptr, obfsc_hash.data(), 20);
 		ptr += 20;
 
-		// Discard DH key exchange data, setup RC4 keys
+		// Discard DH key exchange data, setup RC4 and AES-CTR keys
 		m_rc4 = init_pe_rc4_handler(secret_key, info_hash, is_outgoing());
+#if TORRENT_HAS_MSE_AES_CTR
+		m_aes_ctr = init_pe_aes_ctr_handler(secret_key, info_hash, is_outgoing());
+#endif
```
arvidn: if the rationale for this is to improve performance, I would think you'd only want to initialize *one* per connection. Wouldn't this double the cost of initializing a new (encrypted) connection?

src/pe_crypto.cpp
```diff
@@ -395,6 +411,201 @@ std::size_t rc4_encrypt(unsigned char *out, std::size_t outlen, rc4 *state)
...
+		iv[1] = std::uint8_t(key[17]);
+		iv[2] = std::uint8_t(key[18]);
+		iv[3] = std::uint8_t(key[19]);
+		iv[15] = 64;
+		return iv;
+	}
+
+#if defined(TORRENT_USE_LIBCRYPTO) || defined(TORRENT_USE_OPENSSL)
+
+	static void init_ctr_state(aes_ctr_handler::aes_ctr_state*& state, span<char const> key)
```
arvidn: `static` is redundant here, it's already in the anonymous namespace

arvidn (commented):
I think there would need to be some consensus with other client developers about this new protocol to warrant implementing it in libtorrent.

I think one way to make a convincing argument would be benchmarks demonstrating a significant improvement. I recently added bencher to this repository (`tools/bencher.cpp`) where new micro benchmarks can be added.

Are there any surveys or data points on whether ISPs identify and throttle bittorrent traffic still? And perhaps more importantly, whether obfuscating the bytes makes a difference (packet sizes and traffic patterns may be sufficient to identify bittorrent traffic anyway)

There should also be end-to-end tests (transfer tests) that run over this new encryption algorithm, to ensure it works (and keeps working)

trim21:
> I think there would need to be some consensus with other client developers about this new protocol to warrant implementing it in libtorrent.

This PR is a proof of concept and I would not expected we will merge it after even all code looks good. I'm hoping to have some preliminary discussions with you first.

If this approach looks good to you, I plan to create issues in other major torrent clients as well — Transmission and rTorrent included.

> I think one way to make a convincing argument would be benchmarks demonstrating a significant improvement. I recently added bencher to this repository (`tools/bencher.cpp`) where new micro benchmarks can be added.

I write one and it looks like this, on my AMD r7 255

| Benchmark | 16kb chunk | byte rate |
|---|---|---|
| **rc4_encrypt_16KB** | 59,218 ns (~59 µs) | ~277 MB/s |
| **aes_ctr_encrypt_16KB** | 1,354 ns (~1.4 µs) | ~12.1 GB/s |

(In issue I test this using my implementation on rc4 with ASM, looks like rc4 in libtorrent is much slower, maybe this is causes by optimization flags)

> Are there any surveys or data points on whether ISPs identify and throttle bittorrent traffic still? And perhaps more importantly, whether obfuscating the bytes makes a difference (packet sizes and traffic patterns may be sufficient to identify bittorrent traffic anyway)

Regardless of whether obfuscation actually helps against ISP throttling, switching from RC4 to AES is worth it for the CPU savings alone.

> There should also be end-to-end tests (transfer tests) that run over this new encryption algorithm, to ensure it works (and keeps working)

I'll add it later

src/bt_peer_connection.cpp
trim21: yes we should use base class here, instead of creating 2 seprated encrypt handler

src/bt_peer_connection.cpp
trim21: I'm not sure here...

trim21:
@arvidn I wrote a RFC: https://github.com/bittorrent/bittorrent.org/pull/179

PR #8594, opened 2026-07-13
Fix missing includes

georgthegreat:
Fixes:
```
libtorrent/include/libtorrent/flags.hpp: error: invalid operands to binary expression ('std::ostream' (aka 'basic_ostream<char>') and 'ios_base &(ios_base &)')
  112 |         { return os << std::hex << std::uint64_t(static_cast<UnderlyingType>(val)); }
```

And

```
libtorrent/include/libtorrent/aux_/io.hpp:150:3: error: no member named 'copy' in namespace 'std'; did you mean 'bcopy'?
  150 |                 std::copy(str.begin(), str.end(), view.begin());
      |                 ^~~~~~~~~
```

### DIFF
--- a/include/libtorrent/aux_/io.hpp
+++ b/include/libtorrent/aux_/io.hpp
@@ -11,9 +11,11 @@ see LICENSE file.
 #ifndef TORRENT_AUX_IO_HPP_INCLUDED
 #define TORRENT_AUX_IO_HPP_INCLUDED
 
+#include <algorithm>
 #include <cstdint>
 #include <string>
 #include <type_traits>
+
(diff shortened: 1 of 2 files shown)


arvidn:
can you explain why you need `iostream` instead of `iosfwd`? And what requires `algorithm` in `io.hpp`?

georgthegreat:
@arvidn sorry, forgot the description.
We are building libtorrent against libcxx with 
```
#define _LIBCPP_REMOVE_TRANSITIVE_INCLUDES
```

hence the errors.

PR #8600, opened 2026-07-14
fix undefined std::toupper on signed char in base32decode

isl-Ramzi:
base32decode passes each input character straight to std::toupper, but the byte can come from an untrusted source such as the percent-decoded base32 info-hash in a magnet link (xt=urn:btih:). When char is signed a byte above 127 is negative, and std::toupper is only defined for arguments representable as unsigned char or EOF, so that call is undefined behavior, and it is locale dependent besides. Uppercase the ASCII range directly the way to_lower already does, which keeps decoding of valid input unchanged.

### DIFF
--- a/test/test_string.cpp
+++ b/test/test_string.cpp
@@ -301,6 +301,11 @@ TORRENT_TEST(base32)
 
 	// make sure invalid encoding returns the empty string
 	TEST_CHECK(base32decode("mZXw6yTBO1{#&*()=") == "");
+
+	// bytes with the high bit set (e.g. from a percent-encoded magnet link)
+	// are not valid base32 and decode to the empty string
+	TEST_CHECK(base32decode("MZXW6YT\x80") == "");
+	TEST_CHECK(base32decode(std::string("MZXW\xffYTBO")) == "");
 }
 
 TORRENT_TEST(escape_string)
--- a/src/escape_string.cpp
+++ b/src/escape_string.cpp
@@ -36,7 +36,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "libtorrent/config.hpp"
 
 #include <string>
-#include <cctype>
 #include <algorithm>
 #include <mutex>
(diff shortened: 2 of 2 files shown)


test/test_string.cpp
```diff
@@ -281,6 +281,12 @@ TORRENT_TEST(base32)
 
 	// make sure invalid encoding returns the empty string
 	TEST_CHECK(base32decode("mZXw6yTBO1{#&*()=") == "");
+
+	// bytes with the high bit set (e.g. from a percent-encoded magnet link)
+	// are not valid base32 and must be rejected without relying on std::toupper
+	// being called with a negative char value
```
arvidn: I think the comments should only describe the current state of the code, not previous behavior. There's git history (commit message), github PR description and Changelog for tracking fixes

arvidn (commented):
I'm surprised `undefined-sanitizer` hasn't flagged this. There is a fuzzer for `base32decode()`.

arvidn:
oh, would you mind rebasing (and retargeting) this PR against `RC_2_0`? I'll merge it forward into `RC_2_1` and `master` later.

test/test_string.cpp
isl-Ramzi: Good point. Reworded it to just say high-bit bytes decode to the empty string. I trimmed the comment in escape_string.cpp the same way so it describes what the code does rather than narrating the old behavior.

isl-Ramzi:
Rebased and retargeted onto RC_2_0, so it's a single commit now.

On the sanitizer: this is contractual UB, not the kind UBSan instruments. libc itself isn't compiled with the sanitizer, and on glibc toupper() is a lookup into a table that's offset to accept indices down to -128, so a negative char reads a valid slot instead of going out of bounds. There's nothing for UBSan to trip on at runtime even though the call is undefined per the standard, which is why the fuzzer stayed quiet.

arvidn (approved):

PR #8606, opened 2026-07-15
fix stack buffer overflow in dht canonical_string on large salt

isl-Ramzi:
Repro: sign or verify a BEP44 mutable item with a salt larger than canonical_string's internal buffer, e.g. dht_put_item()/dht_get_item() with a salt over ~1130 bytes. The added test signing_oversized_salt does this and segfaults without the patch.

Cause: canonical_string() builds the signing string into a fixed char str[1200] and advances its cursor with ptr += std::snprintf(...). snprintf returns the length it would have written, so once the salt fills the buffer that advance moves ptr past the end. left comes from span::size() (a signed ptrdiff_t) and goes negative, then std::min(size, left) hands the following std::copy a negative count, which walks off the stack buffer. The 64-byte salt cap only lives on the incoming DHT put path (node.cpp), not in this helper, and sign/verify are reachable straight from the public API with the caller's salt.

Fix: clamp each snprintf advance to the space remaining so ptr never passes the end of the buffer and left stays non-negative. Output for in-range inputs is unchanged, so existing signatures still verify.

Verified with a small ASan reproducer of the routine and end to end: signing_oversized_salt segfaults on the current code and passes after the change, and the existing signing_test1-3 and mutable_get/put tests still pass.


arvidn:
if the salt needs to be truncated, the result won't be valid anyway, and we might as well just discard the request or fail validation.

Also, how do you get past the check that salt is <= 64 bytes?
https://github.com/arvidn/libtorrent/blob/RC_2_1/src/kademlia/node.cpp#L1000

isl-Ramzi:
Agreed, truncating just yields a signature nobody can verify. I dropped the clamp-and-continue and instead reject: verify_mutable_item now returns false when the salt is over 64 bytes, so it fails validation the same way the incoming put path already does. canonical_string still stays in bounds so the signing side can't walk off the stack if it's ever misused.

On getting past the node.cpp check: that 64-byte cap only sits on the incoming put wire path. The sign/verify helpers are also reachable straight from the public API with the caller's salt uncapped: dht_put_item goes through item::assign -> sign_mutable_item, and dht_get_item with a salt runs the get response back through verify_mutable_item. The python sign_mutable_item binding passes the salt through as-is too. So an app that hands any of those a salt larger than 64 bytes reaches canonical_string without ever touching node.cpp.

Updated the test to expect the oversized salt to be rejected and a 64-byte salt to still round-trip. Full test_dht passes.

ChangeLog
```diff
@@ -1,5 +1,6 @@
 2.1.1 not released
 
+	* fix stack buffer overflow in DHT mutable item signing/verification with oversized salt
```
arvidn: this should probably be toned down a bit now too. it sounds like it's exploitable

isl-Ramzi: Fair point, it reads worse than it is now that the wire path was never the issue. Reworded it to:

  * fail validation of DHT mutable items whose salt exceeds the 64 byte limit

The title still says overflow, happy to rename that to something like "fail validation of DHT mutable items with oversized salt" if you want it consistent.

Issue #8041, opened 2025-10-06
`external_ip_alert` is never emitted, but it gets IP from router using UPnP just fine


MestreLion:
This is my first time fiddling with libtorrent, so apologies if I'm doing any blatant mistake. While using Deluge, I was puzzled its External IP in Statusbar was always `n/a`, so I tried to create a Proof of Concept just to fetch the external IP:

```python
# torrent.py
import time
import libtorrent as lt

session = lt.session()
session.apply_settings({'alert_mask': lt.alert.category_t.all_categories})
while True:
	assert session.is_listening()
	alerts = session.pop_alerts()
	for alert in alerts:
		if (
			alert.what() == "external_ip_alert" or  # never emitted!
			"router external IP address" in alert.message()  # but it works fine
		):
			print(f"{alert.what()}\t{alert.message()}")
	time.sleep(1)
```
Results in (actual IPs redacted):
```
$ python3 torrent.py 
portmap_log	UPnP [10.10.10.123]: got router external IP address 123.111.222.123
```

Per the `portmap_log` alert, the UPnP call was successful, but the expected `external_ip_alert` alert never happened. Is this a known issue?

Package: libtorrent-rasterbar2.0t64  (from Ubuntu's 24.04 offcial repository)
Version: 2.0.10-1.1build2

MestreLion:
Some additional notes:
- It seems the actual alert type expected to be returned from `what()` is `external_ip`, not `external_ip_alert`. The [docs on Alerts](https://www.libtorrent.org/reference-Alerts.html#overview) are somewhat misleading in this regard, and the [docs on python bindings](https://www.libtorrent.org/python_binding.html#using-libtorrent-in-python) does not mention the `_alert` suffix is dropped
- I was not able to find `.timestamp()` method on alerts. Is there any way to retrieve it in Python?

ilayanyatto72733:
I have no problem receiving `external_ip_alert` alert (not `external_ip` alert) in C++ in all versions I've tried.

MestreLion:
@ilayanyatto72733 
> I have no problem receiving `external_ip_alert` alert (not `external_ip` alert) in C++ in all versions I've tried.

Thanks for the feedback!

[...]

PR #8607, opened 2026-07-15
fix external_ip_alert not being posted when UPnP/NAT-PMP succeeds

rts-kotori:
## Bug

`external_ip_alert` is not posted when UPnP/NAT-PMP successfully maps a
port and reports an external IP.

## Cause

`on_port_mapping` called `cast_vote` directly with `address()` as the
voter, bypassing `set_external_address`, which is where
`external_ip_alert` is posted.

## Fix

Call `set_external_address` instead, with the router/gateway address as
the vote source.

## Test

Tested with qBittorrent (uses libtorrent), with UPnP/NAT-PMP enabled.

Before:

```
UPnP/NAT-PMP port mapping succeeded. Message: "successfully mapped port using NAT-PMP. local: 10.0.0.128 external port: TCP/47191"
UPnP/NAT-PMP port mapping succeeded. Message: "successfully mapped port using NAT-PMP. local: 10.0.0.128 external port: UDP/47191"
Detected external IP. IP: "2001:ac8:a:2::18"
```

Port mapping succeeds, but no IPv4 external IP from NAT-PMP.

After:

```
UPnP/NAT-PMP port mapping succeeded. Message: "successfully mapped port using NAT-PMP. local: 10.0.0.128 external port: TCP/63027"
Detected external IP. IP: "146.70.113.106"
UPnP/NAT-PMP port mapping succeeded. Message: "successfully mapped port using NAT-PMP. local: 10.0.0.128 external port: UDP/63027"
Detected external IP. IP: "2001:ac8:a:2::18"
```

Both IPv4 and IPv6 external addresses are reported.

Closes issue #8041

### DIFF
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -5740,7 +5740,10 @@ namespace {
 		{
 			// TODO: 1 report the proper address of the router as the source IP of
 			// this vote of our external address, instead of the empty address
-			listen_socket->external_address.cast_vote(external_ip, source_router, address());
+			auto sock = ls.m_sock.lock();
+			TORRENT_ASSERT(sock);
+			if (sock)
+				set_external_address(sock, external_ip, source_router, address());
 		}
 
 		// need to check whether this mapping is for one of session ports (it could also be a user mapping)
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,6 @@
 2.0.14 not released
 
+	* fix external_ip_alert not being posted under some conditions when UPnP/NAT-PMP is enabled
 	* fix issue with stuck trackers on session pause
 	* fix windows disk read, possibly causing short-read errors
 	* fix blocking UDP socket in LSD and NAT-PMP


src/session_impl.cpp
```diff
@@ -5738,9 +5738,23 @@ namespace {
...
 		{
-			// TODO: 1 report the proper address of the router as the source IP of
-			// this vote of our external address, instead of the empty address
-			listen_socket->external_address.cast_vote(external_ip, source_router, address());
+			address gateway;
+			switch (transport)
+			{
+			case portmap_transport::natpmp:
+				if (listen_socket->natpmp_mapper)
+					gateway = listen_socket->natpmp_mapper->get_natpmp_gateway();
```
arvidn: this is not a reliable (or clean) way of doing this. if we do indeed need the gateway here, it should be passed along in the `on_port_mapping()` call

src/session_impl.cpp
```diff
@@ -5738,9 +5738,23 @@ namespace {
...
+			case portmap_transport::upnp:
+				if (listen_socket->upnp_mapper)
+					gateway = listen_socket->upnp_mapper->get_upnp_gateway(external_ip);
+				break;
+			}
+
+			auto sock = ls.m_sock.lock();
+			TORRENT_ASSERT(sock);
+			if (sock)
+				set_external_address(sock, external_ip, source_router, gateway);
```
arvidn: as far as I can tell, this is the material change. whether we pass `gateway` or `address()` here is not so important.

Just this change is small, simple and gets to the point of this PR. everything else is complexity with questionable value

rts-kotori:
Thanks for the review. Dropped the gateway IP address related changes. The PR now only replaces the `cast_vote` call with `set_external_address` call, reverted back to using `address()` as the vote source.

arvidn (approved):

PR #8620, opened 2026-07-17
fix torrent updates for idle torrents

arvidn:
This should allow the last update, with 0 transfer rate, to be posted.


HanabishiRecca (approved):
Tested, it does indeed work.

PR #8621, opened 2026-07-18
reject out-of-range numbers instead of truncating them with atoi

isl-Ramzi:
Repro: `aux::parse_endpoint("127.0.0.1:4294967376", ec)` returns port 80 with no error, and a proxy answering `HTTP/1.1 4294967496 Forbidden` gets past the `code != 200` check in `http_stream::handshake2`.

Cause: `std::atoi` is undefined for values that don't fit in an int, and in practice truncates them, so the number wraps into exactly the range the check afterwards accepts. Both of those inputs land on 200 / 80 after truncation. The port one is reachable from a magnet `x.pe=` and from `dht_bootstrap_nodes`; the same construct is used for the URL port, the UPnP SOAP `errorCode`, the magnet select-only indices, and the deprecated bdecode string length.

Fix: added `aux::parse_decimal()` on top of `std::from_chars`, which reports out-of-range input rather than wrapping it, and switched every one of those call sites over. `src/http_parser.cpp` already parses HTTP status codes this way, so this just brings the rest in line. Valid input parses identically.

The new `parse_decimal` test and the two extra cases in `parse_invalid_ipv4_endpoint` both fail before the change. `test_string`, `test_socket_io`, `test_magnet`, `test_bencoding`, `test_http_parser`, `test_settings_pack`, `test_upnp`, `test_bdecode` and `test_torrent_info` pass after it, and the deprecated `bdecode_recursive` path still compiles under `TORRENT_ABI_VERSION=1`.

I targeted RC_2_1 rather than RC_2_0 this time because RC_2_0 is C++14 and has no `<charconv>`, so the helper there would need a hand-rolled parser instead. Happy to write that version too if you'd rather take it downstream first and merge forward.

### DIFF
--- a/src/string_util.cpp
+++ b/src/string_util.cpp
@@ -19,9 +19,18 @@ see LICENSE file.
 #include "libtorrent/assert.hpp"
 
 #include <algorithm> // for search
+#include <charconv> // for from_chars
 #include <utility> // for move
 
 namespace libtorrent::aux {
+
+	int parse_decimal(string_view const str)
+	{
+		int ret = 0;
+		auto const r = std::from_chars(str.data(), str.data() + str.size(), ret);
+		if (r.ec != std::errc{}) return -1;
+		return ret;
+	}
 
 	// We need well defined results that don't depend on locale
 	std::array<char, 4 + std::numeric_limits<std::int64_t>::digits10>
@@ -227,7 +236,7 @@ namespace libtorrent::aux {
 				continue;
 			}
 
-			iface.port = std::atoi(port_str.c_str());
+			iface.port = parse_decimal(port_str);
 			if (iface.port < 0 || iface.port > 65535)
 			{
 				err.emplace_back(element);
@@ -279,7 +288,8 @@ namespace libtorrent::aux {
 
 			if (colon != std::string::npos && colon > start)
 			{
-				int port = std::atoi(in.substr(colon + 1, end - colon - 1).c_str());
+				int const port = parse_decimal(strip_string(
+					string_view(in).substr(colon + 1, end - colon - 1)));
 
 				// skip trailing spaces
 				std::string::size_type soft_end = colon;
--- a/src/string_util.cpp
+++ b/src/string_util.cpp
@@ -28,7 +28,8 @@ namespace libtorrent::aux {
 	{
 		int ret = 0;
 		auto const r = std::from_chars(str.data(), str.data() + str.size(), ret);
-		if (r.ec != std::errc{}) return std::nullopt;
+		if (r.ec != std::errc{})
+			return std::nullopt;
 		return ret;
 	}
 
--- a/src/string_util.cpp
+++ b/src/string_util.cpp
@@ -24,11 +24,11 @@ see LICENSE file.
 
 namespace libtorrent::aux {
 
-	int parse_decimal(string_view const str)
+	std::optional<int> parse_decimal(string_view const str)
 	{
 		int ret = 0;
 		auto const r = std::from_chars(str.data(), str.data() + str.size(), ret);
-		if (r.ec != std::errc{}) return -1;
+		if (r.ec != std::errc{}) return std::nullopt;
 		return ret;
 	}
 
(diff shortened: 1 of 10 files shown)


src/string_util.cpp
```diff
@@ -19,10 +19,19 @@ see LICENSE file.
...
+#include <charconv> // for from_chars
 #include <utility> // for move
 
 namespace libtorrent::aux {
 
+	int parse_decimal(string_view const str)
+	{
+		int ret = 0;
+		auto const r = std::from_chars(str.data(), str.data() + str.size(), ret);
+		if (r.ec != std::errc{}) return -1;
```
arvidn: in a way, it's a shame to throw away the error here. we have the opportunity to not make the same mistake as `atoi()` and distinguish a successful parse of `-1` and an error.

arvidn (commented):
this is definitely an improvement. Do you think it would make the patch unreasonably large to have proper error reporting, instead of returning -1?

src/string_util.cpp
isl-Ramzi: Agreed, and it wasn't much extra. Pushed a commit that makes it return `std::optional<int>`, so a successful parse of -1 is now distinguishable from a failure.

Each call site handles the empty optional on its own terms: `invalid_port` in `parse_endpoint` and `parse_url_components`, a rejected CONNECT in `http_stream::handshake2`, a skipped entry for the magnet select-only indices, the dht node port and the listen interfaces, and the UPnP SOAP parser leaves `error_code` at its -1 default rather than inventing a code. That last one is the only place the old sentinel would have been ambiguous anyway.

The touched tests still pass (`test_string`, `test_socket_io`, `test_magnet`, `test_bencoding`, `test_upnp`, `test_settings_pack`), and the deprecated `bdecode_recursive` path still compiles under `TORRENT_ABI_VERSION=1`.

arvidn:
if you install `pre-commit` and git-clang-format, the formatting is handled for you

isl-Ramzi:
Set that up, thanks for the pointer. Pushed the formatting fix plus the MSVC build break that was the actual cause of the other red jobs: `string_view::iterator` is a class type on MSVC, so `parse_decimal({port_pos, len})` in `parse_url.cpp` didn't compile. It takes the substring off `url` directly now.

One thing worth a look in the format commit: I also moved the `parse_decimal` declaration down next to `strip_string`. Sitting directly under the `TORRENT_EXTRA_EXPORT` / `to_string` block, clang-format reads it as a macro continuation and re-indents `is_digit` and `ensure_trailing_slash` along with it, leaving the closing brace at the old level. Moving the declaration keeps the hook happy without touching unrelated lines, but say the word if you'd rather I put it back where it was.

`test_socket_io`, `test_string`, `test_magnet`, `test_upnp` and `test_bencoding` still pass.

arvidn:
```diff
diff --git a/src/string_util.cpp b/src/string_util.cpp
index 1f5af12..51b30a8 100644
--- a/src/string_util.cpp
+++ b/src/string_util.cpp
@@ -28,7 +28,8 @@ namespace libtorrent::aux {
 	{
 		int ret = 0;
 		auto const r = std::from_chars(str.data(), str.data() + str.size(), ret);
-		if (r.ec != std::errc{}) return std::nullopt;
+		if (r.ec != std::errc{})
+			return std::nullopt;
 		return ret;
 	}
 
diff --git a/src/upnp.cpp b/src/upnp.cpp
index f86874e..fec8a66 100644
--- a/src/upnp.cpp
+++ b/src/upnp.cpp
@@ -1220,7 +1220,8 @@ void find_error_code(int const type, string_view string, error_code_parse_state&
 	{
 		// leave error_code at -1 if the device sent something that isn't a
 		// number, or a number that doesn't fit in an int
-		if (auto const code = aux::parse_decimal(string)) state.error_code = *code;
+		if (auto const code = aux::parse_decimal(string))
+			state.error_code = *code;
 		state.exit = true;
 	}
 }
```

arvidn:
sorry, I tweaked the clang-format config in a PR that landed ahead of this

isl-Ramzi:
No worries, applied both hunks as-is. Nothing else in the diff hits the `AllowShortIfStatementsOnASingleLine` change, those two were the only ones.

Still open from my last comment if you have a preference: I'd moved the `parse_decimal` declaration in string_util.hpp down next to `strip_string` to keep clang-format from re-indenting neighbouring lines. Happy to move it back if you'd rather it sit with the other parsing helpers.

arvidn (approved):

arvidn (approved):

PR #8624, opened 2026-07-18
Glassez fix resume tracker tiers

arvidn:
https://github.com/arvidn/libtorrent/pull/8622

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -61,7 +61,7 @@ IndentCaseLabels: true
 AllowShortFunctionsOnASingleLine: All
 AllowShortLambdasOnASingleLine:   All
 # 'if (cond) return;' without braces is common in this codebase
-AllowShortIfStatementsOnASingleLine: WithoutElse
+AllowShortIfStatementsOnASingleLine: Never
 AllowShortBlocksOnASingleLine: Empty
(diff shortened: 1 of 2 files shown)


.clang-format
```diff
@@ -61,7 +61,7 @@ IndentCaseLabels: true
 AllowShortFunctionsOnASingleLine: All
 AllowShortLambdasOnASingleLine:   All
 # 'if (cond) return;' without braces is common in this codebase
-AllowShortIfStatementsOnASingleLine: WithoutElse
+AllowShortIfStatementsOnASingleLine: Never
```
glassez, line 64: I actually do not know how clang-format considers one-line code like `if (cond) return;` using this option. Perhaps it disallow it. So the question is, what do you really prefer to get in the end?

glassez: So if you still prefer one liners, I will fix my PR (despite I feel it a bad choice).

glassez: I feel it confusing that the option called Allow actually Forces it.

arvidn: I think this is a good start. it's already made some quite unreasonable reformats, I'm hoping to refine it over time

PR #8630, opened 2026-07-19
fix out of bounds write in hashes_rejected on block hash requests

isl-Ramzi:
hashes_rejected() indexes m_piece_hash_requested by req.index / 512, which is only a piece index when req.base is the piece layer. After a v2 piece fails its hash check, pick_hashes() issues a base 0 request whose index counts blocks instead, and a peer that rejects that request rather than answering it walks the bucket index past the end of the vector, since on_hash_reject() only matches the reject against an outstanding request without validating it. Added the runtime check add_hashes() already has for the same bookkeeping, since the assert that documented the invariant is compiled out in release builds.

### DIFF
--- a/src/hash_picker.cpp
+++ b/src/hash_picker.cpp
@@ -356,13 +356,7 @@ bool validate_hash_request(hash_request const& hr, file_storage const& fs)
 	void hash_picker::hashes_rejected(hash_request const& req)
 	{
 		// only requests at the piece layer are recorded in
-		// m_piece_hash_requested. The index of a request at any other layer
-		// counts nodes in that layer rather than pieces, so it would index the
-		// wrong bucket, or one past the end of the vector. pick_hashes() issues
-		// base 0 requests for the block hashes of a failed piece, and
-		// on_hash_reject() only matches the reject against an outstanding
-		// request, it does not validate it. This mirrors the check in
-		// add_hashes().
+		// m_piece_hash_requested.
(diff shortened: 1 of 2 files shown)


src/hash_picker.cpp
```diff
@@ -355,7 +355,17 @@ bool validate_hash_request(hash_request const& hr, file_storage const& fs)
...
 	{
-		TORRENT_ASSERT(req.base == m_piece_layer && req.index % 512 == 0);
+		// only requests at the piece layer are recorded in
+		// m_piece_hash_requested. The index of a request at any other layer
+		// counts nodes in that layer rather than pieces, so it would index the
+		// wrong bucket, or one past the end of the vector. pick_hashes() issues
+		// base 0 requests for the block hashes of a failed piece, and
+		// on_hash_reject() only matches the reject against an outstanding
+		// request, it does not validate it. This mirrors the check in
+		// add_hashes().
```
arvidn:
```suggestion
		// only requests at the piece layer are recorded in
		// m_piece_hash_requested.
```

arvidn (approved):
looks good, I'd like the comment trimmed though

src/hash_picker.cpp
isl-Ramzi: Done, trimmed it to those two lines and pushed.

arvidn (approved):

PR #8636, opened 2026-07-22
reject invalid chunk sizes in web_peer_connection::on_receive

isl-Ramzi:
parse_chunk_header() reports a bad chunk header by setting chunk_size to -1, and it otherwise accepts hex sizes all the way up to int64 max, but on_receive() adds the value straight into the int m_chunk_pos without looking at it. A web seed answering a range request with Transfer-Encoding: chunked and a header of "1z" or "80000000" drives m_chunk_pos negative, which is the same value this function uses to mean "response complete", so the chunk loop exits with the receive buffer still non-empty and the enclosing for(;;) spins doing no work. I reproduced it against the fuzzers' web seed harness: 5.1s of CPU burned in 5s of wall time with no disconnect, and 0.04s with the peer dropped as "Invalid HTTP header" afterwards. http_parser::incoming() already applies the same check to the chunk headers it parses itself, so this just gives the web seed path the matching guard; test_web_seed_chunked still passes.

### DIFF
--- a/src/web_peer_connection.cpp
+++ b/src/web_peer_connection.cpp
@@ -19,6 +19,7 @@ see LICENSE file.
 #include <cstdlib>
 #include <cstdio> // for snprintf
 #include <cinttypes> // for PRId64 et.al.
+#include <limits>
 
 #include "libtorrent/aux_/web_peer_connection.hpp"
 #include "libtorrent/session.hpp"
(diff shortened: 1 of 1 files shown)


arvidn:
please feel free to PR the fuzzer too!

arvidn (approved):

arvidn:
I think this fix applies to `RC_2_0` as well. I can back-port it later (unless you want to rebase it)

arvidn:
```diff
--- a/src/web_peer_connection.cpp
+++ b/src/web_peer_connection.cpp
@@ -1042,8 +1042,7 @@ void web_peer_connection::on_receive(error_code const& error
 				// all the way up to int64 max, which don't fit in m_chunk_pos.
 				// http_parser::incoming() applies the same check to the chunk
 				// headers it parses itself
-				if (chunk_size < 0
-					|| chunk_size > std::numeric_limits<int>::max() - m_chunk_pos)
+				if (chunk_size < 0 || chunk_size > std::numeric_limits<int>::max() - m_chunk_pos)
 				{
 					received_bytes(0, int(recv_buffer.size()));
 					disconnect(errors::http_parse_error, operation_t::bittorrent, peer_error);
```

arvidn:
I could fix up the PR if you'd want

isl-Ramzi:
Thanks for applying this and taking care of the back-port. I'll send the web seed fuzzer for the chunk path as a separate PR.

PR #8637, opened 2026-07-22
Use libcrypto for peer encryption key exchange

Tansien:
## Summary

This follows the 160-bit exponent optimization merged in #8565 and offloads MSE
modular exponentiation to libcrypto when OpenSSL is available.

`dh_key_exchange` now has two implementations selected at compile time: a
libcrypto BIGNUM one (when `TORRENT_USE_LIBCRYPTO` is defined and
`TORRENT_USE_WOLFSSL` is not) and the existing Boost.Multiprecision one for
every other build.

- keys are stored and exposed in wire format (96-byte big-endian arrays), so
  the boost-int -> bytes -> BIGNUM round-trips are gone. `key_t` and
  `export_key()` are now private to `pe_crypto.cpp`, and `pe_crypto.hpp` no
  longer includes `<boost/multiprecision/cpp_int.hpp>` at all
- uses `BN_mod_exp_mont()` with a per-thread cached `BN_CTX`, `BN_MONT_CTX`
  and prime. Constant-time exponentiation and `OPENSSL_cleanse()` are
  deliberately not used, per review -- PE is obfuscation, not confidentiality,
  and the exponent is single-use
- sticks to BN APIs that LibreSSL also declares (no `BN_mod_exp_mont_word`),
  and keeps `BN_bn2bin` + manual left-pad rather than `BN_bn2binpad`, so
  OpenSSL 1.0 still builds
- `BN_CTX_start`/`BN_CTX_end` and `ERR_set_mark`/`ERR_pop_to_mark` are paired
  with `aux::scope_end`, so the shared per-thread context cannot be left
  unbalanced by an early return
- the degenerate-key rejection (outside `[2, p-2]`) is preserved in both
  backends, and the prime now has a single byte-array definition shared by
  both
- there is no `powm()` fallback in libcrypto builds: a local crypto failure
  fails the handshake. `good()` reports it, and the caller distinguishes it
  from a hostile peer key so our own failure isn't attributed to the peer at
  `peer_error` severity

The OpenSSL error queue is preserved across the whole operation via
`ERR_set_mark()`/`ERR_pop_to_mark()`, since the network thread shares it with
SSL torrents and HTTPS trackers.

## Benchmarks

`tools/bencher`, Release, AppleClang, OpenSSL 3.6, Apple silicon, measured
against this branch's merge base:

| benchmark | RC_2_1 | this branch |
|---|---:|---:|
| dh_key_exchange | 166 us | 24 us |
| dh_compute_secret | 183 us | 26 us |
| dh_handshake | 339-409 us | 52 us |

That is roughly a 7x improvement on the full per-connection DH cost. The
baseline varied noticeably run to run on a busy laptop, so treat these as
indicative -- now that #8637's base is `RC_2_1` and bencher thresholds are
configured, CI should produce the authoritative numbers.

## Tests

- `test_pe_crypto` (6 tests) passes with `crypto=openssl` and with the
  built-in/Boost backend, under ASan+UBSan, and with asserts and
  invariant-checks enabled
- new tests: a failed exchange must not leave the previous shared secret
  readable, and the xor mask must be `hash('req3', S)` over exactly what
  `get_secret()` returns. The degenerate-key test gained a `p+1` case, since
  `p+1` is congruent to 1 and would otherwise fix the shared secret
- cross-backend check: a libcrypto-built and a Boost-built binary perform DH
  with each other and agree on the shared secret, which pins the prime,
  endianness and zero-padding against the previous implementation rather than
  only against itself
- both PE fuzzers and `tools/bencher` updated to the new interface

Caveat worth stating: all of the above was run through CMake. I don't have
`b2` available locally, so the boost-build path -- including its warning set --
is only covered by CI here.

## Open questions

- `dh_key_exchange` is `aux_`-internal but `TORRENT_EXTRA_EXPORT`; its layout
  changes here. I've assumed `aux_` types sit outside the `RC_*` ABI contract,
  but say the word if that should target `master` instead.
- `errors::no_memory` now covers any local libcrypto failure, not just
  allocation. `no_entropy = 200` looks like precedent for giving this its own
  code -- happy to add one if you'd prefer.

### DIFF
--- a/src/pe_crypto.cpp
+++ b/src/pe_crypto.cpp
@@ -16,94 +16,346 @@ see LICENSE file.
 #include <cstdint>
 #include <cstring>
 #include <algorithm>
-#include <random>
+#include <iterator>
+#include <optional>
 #include <utility>
 
+#if defined TORRENT_USE_LIBCRYPTO && !defined TORRENT_USE_WOLFSSL
 #include "libtorrent/aux_/disable_warnings_push.hpp"
+#include <openssl/bn.h>
+#include <openssl/err.h>
+#include "libtorrent/aux_/disable_warnings_pop.hpp"
+#else
+#include "libtorrent/aux_/disable_warnings_push.hpp"
+#include <boost/multiprecision/cpp_int.hpp>
 #include <boost/multiprecision/integer.hpp>
 #include "libtorrent/aux_/disable_warnings_pop.hpp"
+#endif
 
 #include "libtorrent/aux_/random.hpp"
 #include "libtorrent/aux_/alloca.hpp"
 #include "libtorrent/aux_/pe_crypto.hpp"
+#include "libtorrent/aux_/scope_end.hpp"
 #include "libtorrent/hasher.hpp"
 
 namespace libtorrent::aux {
 
 	namespace {
-		// TODO: it would be nice to get the literal working
-		key_t const dh_prime
-			("0xFFFFFFFFFFFFFFFFC90FDAA22168C234C4C6628B80DC1CD129024E088A67CC74020BBEA63B139B22514A08798E3404DDEF9519B3CD3A431B302B0A6DF25F14374FE1356D6D51C245E485B576625E7EC6F44C42E9A63A36210000000000090563");
+		// the prime P from the MSE spec, 768 bits big-endian. The
+		// generator is 2
+		// clang-format off
+		unsigned char const dh_prime_bytes[96] = {
+			0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+			0xc9, 0x0f, 0xda, 0xa2, 0x21, 0x68, 0xc2, 0x34,
+			0xc4, 0xc6, 0x62, 0x8b, 0x80, 0xdc, 0x1c, 0xd1,
+			0x29, 0x02, 0x4e, 0x08, 0x8a, 0x67, 0xcc, 0x74,
+			0x02, 0x0b, 0xbe, 0xa6, 0x3b, 0x13, 0x9b, 0x22,
+			0x51, 0x4a, 0x08, 0x79, 0x8e, 0x34, 0x04, 0xdd,
+			0xef, 0x95, 0x19, 0xb3, 0xcd, 0x3a, 0x43, 0x1b,
+			0x30, 0x2b, 0x0a, 0x6d, 0xf2, 0x5f, 0x14, 0x37,
+			0x4f, 0xe1, 0x35, 0x6d, 0x6d, 0x51, 0xc2, 0x45,
+			0xe4, 0x85, 0xb5, 0x76, 0x62, 0x5e, 0x7e, 0xc6,
+			0xf4, 0x4c, 0x42, 0xe9, 0xa6, 0x3a, 0x36, 0x21,
+			0x00, 0x00, 0x00, 0x00, 0x00, 0x09, 0x05, 0x63
+		};
+		// clang-format on
+
+		char const req3[4] = {'r', 'e', 'q', '3'};
 	}
 
-	std::array<char, 96> export_key(key_t const& k)
-	{
-		std::array<char, 96> ret;
-		auto* begin = reinterpret_cast<std::uint8_t*>(ret.data());
-		std::uint8_t* end = mp::export_bits(k, begin, 8);
+	void rc4_init(const unsigned char* in, std::size_t len, rc4* state);
+	std::size_t rc4_encrypt(unsigned char* out, std::size_t outlen, rc4* state);
 
-		// TODO: it would be nice to be able to export to a fixed width field, so
-		// we wouldn't have to shift it later
-		if (end < begin + 96)
+#if defined TORRENT_USE_LIBCRYPTO && !defined TORRENT_USE_WOLFSSL
+
+	namespace {
+
+		// per-thread libcrypto state: scratch space for BIGNUM math and
+		// constants derived from the prime, set up once and reused by
+		// every key exchange on the thread
+		struct openssl_context
 		{
-			int const len = int(end - begin);
-#if defined __GNUC__ && __GNUC__ == 12
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wstringop-overflow"
-#endif
-			std::memmove(begin + 96 - len, begin, aux::numeric_cast<std::size_t>(len));
-#if defined __GNUC__ && __GNUC__ == 12
-#pragma GCC diagnostic pop
-#endif
-			std::memset(begin, 0, aux::numeric_cast<std::size_t>(96 - len));
-		}
-		return ret;
-	}
+			openssl_context() = default;
+			~openssl_context()
+			{
+				BN_MONT_CTX_free(montgomery);
+				BN_CTX_free(ctx);
+				BN_free(prime_minus_1);
+				BN_free(prime);
+			}
+			openssl_context(openssl_context const&) = delete;
+			openssl_context& operator=(openssl_context const&) = delete;
 
-	void rc4_init(const unsigned char* in, std::size_t len, rc4 *state);
-	std::size_t rc4_encrypt(unsigned char *out, std::size_t outlen, rc4 *state);
+			// these are initialized in declaration order, so prime_minus_1
+			// sees an already-initialized prime, and valid sees all four
+			BIGNUM* const prime = BN_bin2bn(dh_prime_bytes, int(sizeof(dh_prime_bytes)), nullptr);
+			BIGNUM* const prime_minus_1 = prime != nullptr ? BN_dup(prime) : nullptr;
+			BN_CTX* const ctx = BN_CTX_new();
+			BN_MONT_CTX* const montgomery = BN_MONT_CTX_new();
+			bool const valid = prime != nullptr && prime_minus_1 != nullptr && ctx != nullptr
+				&& montgomery != nullptr && BN_sub_word(prime_minus_1, 1) == 1
+				&& BN_MONT_CTX_set(montgomery, prime, ctx) == 1;
+		};
+
+		// export a BIGNUM as a 96 byte big-endian string, padded with
+		// leading zeroes. On failure "out" is left zeroed rather than
+		// holding a partial or stale value. The size check is what keeps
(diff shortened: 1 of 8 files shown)


src/pe_crypto.cpp
```diff
@@ -47,6 +47,16 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "libtorrent/aux_/disable_warnings_pop.hpp"
 
+#if defined TORRENT_USE_LIBCRYPTO && !defined TORRENT_USE_WOLFSSL
```
arvidn, line 23: maybe it's a bit misleading to define `TORRENT_USE_LIBCRYPTO` when using wolfssl. Do you think this should be changed?

src/pe_crypto.cpp
```diff
@@ -60,6 +70,94 @@ namespace libtorrent {
...
+		key_t dh_mod_exp(key_t const& base, key_t const& exponent)
+		{
+#if defined TORRENT_USE_LIBCRYPTO && !defined TORRENT_USE_WOLFSSL
+			bool const had_error_mark = ERR_set_mark() == 1;
+#if defined BOOST_NO_CXX11_THREAD_LOCAL
+			openssl_context storage;
+#else
+			thread_local openssl_context storage;
+#endif
+			if (storage.context != nullptr && storage.montgomery != nullptr)
```
arvidn: I think it might be a bit easier to read by inverting this condition, and handling the failure case directly. Then the bulk of the function doesn't need to be indented as deeply

arvidn (dismissed):
looks good! I think the code could be organized a bit better though

arvidn:
I think it would make sense to make this PR against `RC_2_1`, to run the benchmarks in CI. I can back-port it to RC_2_0 if it lands.

src/pe_crypto.cpp
Tansien: Agreed, it reads oddly, but defining `TORRENT_USE_LIBCRYPTO` for wolfSSL predates this PR. The Jamfile sets it alongside `TORRENT_USE_OPENSSL` and `OPENSSL_ALL` so the hashers build against wolfSSL compatibility headers, and several files key off it. wolfSSL compatibility does not provide the BIGNUM and error APIs this change needs, including `BN_MONT_CTX`, `BN_mod_exp_mont_consttime`, and `ERR_set_mark`, so this path keeps the explicit opt-out. I would rather handle a broader macro rename separately.

src/pe_crypto.cpp
Tansien: Done. Failure now returns early and the main path loses an indent level. I also dropped `had_error_mark`: when no mark is set because the queue is empty, `ERR_pop_to_mark()` clears the queue, matching the previous `ERR_clear_error()` branch.

Tansien:
Rebased onto `RC_2_1` and changed the PR base. The conflict was limited to the `aux_` namespace and include changes. I also applied the requested early-return layout. Local `test_pe_crypto` runs pass with OpenSSL 3.6.3, built-in crypto, and OpenSSL with AddressSanitizer plus UndefinedBehaviorSanitizer. The `RC_2_1` `dh_*` benchmark jobs should now capture the effect directly.

src/pe_crypto.cpp
```diff
@@ -34,6 +44,93 @@ namespace libtorrent::aux {
...
+#endif
+
+		key_t dh_mod_exp(key_t const& base, key_t const& exponent)
+		{
+#if defined TORRENT_USE_LIBCRYPTO && !defined TORRENT_USE_WOLFSSL
+			ERR_set_mark();
+#if defined BOOST_NO_CXX11_THREAD_LOCAL
+			openssl_context storage;
+#else
+			thread_local openssl_context storage;
```
arvidn: this variable is implicitly `static`. Perhaps that's obvious, it looked odd to me at first

src/pe_crypto.cpp
```diff
@@ -34,6 +44,93 @@ namespace libtorrent::aux {
...
+
+			BN_CTX_start(storage.context);
+			BIGNUM* const bn_base = BN_CTX_get(storage.context);
+			BIGNUM* const bn_exponent = BN_CTX_get(storage.context);
+			BIGNUM* const bn_modulus = BN_CTX_get(storage.context);
+			BIGNUM* const bn_result = BN_CTX_get(storage.context);
+
+			if (bn_result != nullptr
+				&& BN_bin2bn(reinterpret_cast<unsigned char const*>(base_bytes.data())
+					, int(base_bytes.size()), bn_base) != nullptr
```
arvidn: you're using `bn_base` here without checking it for `nullptr` first

src/pe_crypto.cpp
```diff
@@ -34,6 +44,93 @@ namespace libtorrent::aux {
...
+				&& BN_bin2bn(reinterpret_cast<unsigned char const*>(exponent_bytes.data())
+					, int(exponent_bytes.size()), bn_exponent) != nullptr
+				&& BN_bin2bn(reinterpret_cast<unsigned char const*>(modulus_bytes.data())
+					, int(modulus_bytes.size()), bn_modulus) != nullptr)
+			{
+				if (!storage.montgomery_initialized)
+					storage.montgomery_initialized = BN_MONT_CTX_set(storage.montgomery
+						, bn_modulus, storage.context) == 1;
+				BN_set_flags(bn_exponent, BN_FLG_CONSTTIME);
```
arvidn: I don't think this makes sense. protocol encryption doesn't provide any confidentiality at the level where constant time matters

src/pe_crypto.cpp
```diff
@@ -34,6 +44,93 @@ namespace libtorrent::aux {
...
+				if (storage.montgomery_initialized
+					&& BN_mod_exp_mont_consttime(bn_result, bn_base, bn_exponent
```
arvidn:
```suggestion
					&& BN_mod_exp_mont(bn_result, bn_base, bn_exponent
```

src/pe_crypto.cpp
```diff
@@ -34,6 +44,93 @@ namespace libtorrent::aux {
...
+			auto exponent_bytes = export_key(exponent);
+			auto const modulus_bytes = export_key(dh_prime);
+			std::array<unsigned char, 96> result_bytes{};
+			bool success = false;
+
+			BN_CTX_start(storage.context);
+			BIGNUM* const bn_base = BN_CTX_get(storage.context);
+			BIGNUM* const bn_exponent = BN_CTX_get(storage.context);
+			BIGNUM* const bn_modulus = BN_CTX_get(storage.context);
+			BIGNUM* const bn_result = BN_CTX_get(storage.context);
```
arvidn: only `bn_exponent` and `bn_result` are freed. it looks like the others leak

arvidn (commented):
I can't tell for sure, but it looks like this implementation can be much cleaner (more efficient and fewer reinterpret-cast) if it was implemented as a whole alternative `dh_key_exchange` class.

You could then store the constants and bignums in the native libcrypto format without converting back to `array<char const, 96>`

src/pe_crypto.cpp
arvidn: I see, you call `BN_CTX_end()` to free all of them before the end of the function.

src/pe_crypto.cpp
```diff
@@ -34,6 +44,93 @@ namespace libtorrent::aux {
...
+			ERR_pop_to_mark();
+
+			if (success)
+			{
+				key_t result;
+				mp::import_bits(result, result_bytes.begin(), result_bytes.end());
+				OPENSSL_cleanse(result_bytes.data(), result_bytes.size());
+				return result;
+			}
+			OPENSSL_cleanse(result_bytes.data(), result_bytes.size());
```
arvidn: I think all calls to `OPENSSL_cleanse()` are unnecessary and wasteful too

src/pe_crypto.cpp
```diff
@@ -34,6 +44,93 @@ namespace libtorrent::aux {
...
+#endif
+			if (storage.context == nullptr || storage.montgomery == nullptr)
+			{
+				ERR_pop_to_mark();
+				return mp::powm(base, exponent, dh_prime);
+			}
+
+			auto const base_bytes = export_key(base);
+			auto exponent_bytes = export_key(exponent);
+			auto const modulus_bytes = export_key(dh_prime);
```
arvidn: I think this is an example of how you end up doing unnecessary work. All of these calls turn a boost.multi-precision bigint into bytes, then you turn them into libcrypto bigints below (`BN_bin2bn()`).

`m_dh_local_secret` is a boost.multi-precision integer of random numbers that we use as the exponent. We turn it into a boost bigint, then to bytes, to pass it in here, then to `BIGNUM*` bigint.

If `dh_key_exchange` would contain the definition of `key_t` your class could define it as `BIGNUM*`, and save conversions and allocations. Or maybe even better, wrap it in a class that allocates and frees the `BIGNUM`, and maybe make it non-copyable to keep it simple.

In this scenario, `export_key()` would have to be a static member function of `dh_key_exchange` as well.

arvidn:
the `.clang-format` file may have changed since you first branched `RC_2_1`. If you rebase and re-run `pre-commit` on the commit, it should fix the formatting

arvidn:
I prefer this version: https://github.com/arvidn/libtorrent/pull/8660
I don't think two-phase initialization is a good idea, it gets messy

Tansien:
Rebased onto `183897fde` and force-pushed.

On the formatting: `.clang-format` hasn't changed since this branched — the only update to it is `d1d9a6d7c` (Jul 18), which was already in the tree at the branch point. The pre-commit failure was just this PR's own lines. They've been run through clang-format 18.1.8 now, converged.

Everything else in the review is addressed: native key storage, so the boost → bytes → `BIGNUM` round-trips are gone and `key_t`/`export_key()` are private to the .cpp; constant-time and `OPENSSL_cleanse()` dropped; `BN_mod_exp_mont()`; explicit null checks on every `BN_CTX_get()`; and the `BN_CTX_start`/`BN_CTX_end` and error-mark pairs are now RAII via `aux::scope_end`.

Two things I'd like your call on, also at the bottom of the description: is an `aux_`-internal `TORRENT_EXTRA_EXPORT` layout change acceptable on a `RC_*` branch? And should a local libcrypto failure get its own error code rather than reusing `errors::no_memory` — `no_entropy = 200` looks like precedent.

Issue #6812, opened 2022-04-16
TOS isn't being applied


tiagogaspar8:
libtorrent version (or branch): 2.0.5 (in qbittorrent)

platform/architecture: Windows 11

compiler and compiler version: N/A

please describe what symptoms you see, what you would expect to see instead and
how to reproduce it.

The TOS doesn't seem to be applying correctly, it is configured to 4 by default but in wireshark the DSCP isn't changed and stays at CS0 (0).
I'd also like to suggest the change from the name TOS to DSCP since TOS is deprecated.

Thanks!

arvidn:
did you look at all traffic or just some connections?
I believe DSCP is only meant to apply to peer connections, not DHT or tracker connections for instance.

arvidn:
are you looking at both TCP and UDP peer connections? The DSCP is set via slightly different code paths for those.

arvidn:
do you see the `SET_TOS` messages in the log?

arvidn:
https://github.com/arvidn/libtorrent/pull/6822

tiagogaspar8:
Hi,

My setup only has μtp enabled, and that doesn't have any dscp, I can't say about TCP, but I believe I can try if you'd like!

cas--:
@arvidn This rename has broken compatibility and generating errors in Deluge with lt 1.2, was this applied to the wrong branch and missing setting deprecated in settings_pack.cpp?

```
self.session.apply_settings(settings)
builtins.KeyError: 'unknown name in settings_pack: peer_tos'
```

arvidn:
good point. I need to fix the binding to support the old name as well.

arvidn:
https://github.com/arvidn/libtorrent/pull/6874

tiagogaspar8:
Hi @arvidn , I'd just like to let you know that TCP also suffers from the same issue, dscp  isn't marked

arvidn:
is it possible the process need a special privilege to set it?
Do you have any error messages in the log for setting DSCP?

tiagogaspar8:
I have tried to run it as an admin but it didn't help

arvidn:
@tiagogaspar8 can you get the log? Do you see entries saying `SET_TOS` or `SET_DSCP`? If so, they will contain error codes.

tiagogaspar8:
I'm using qbittorrent and I have no idea how to enable extra debug logging 😞 
Do you have any idea?

mzz2017:
[...]

PR #8652, opened 2026-07-25
fix DSCP marking on outgoing TCP peers

jiaming-shi:
## Problem

PR #7696 moved `peer_dscp` setup before `connect()` so that the TCP SYN would carry the configured traffic class. However, on the current `RC_2_1` branch the call is made before `m_socket.open()`.

An outgoing Boost.Asio TCP socket is only constructed at that point; it has no native descriptor yet. On Linux, `set_traffic_class()` therefore fails with `bad_descriptor`. `open()` subsequently creates the descriptor with the default traffic class (0), so the SYN and the rest of the connection remain unmarked.

This addresses the outgoing TCP part of #6812 and is a follow-up to #7696.

## Change

- Keep the existing single `peer_dscp` setup for incoming sockets, which have already been opened by `accept()`, and report any failure as incoming.
- For outgoing TCP peers, apply it after `open()` and before `bind()` / `async_connect()`.
- Use a separate error code so a DSCP failure cannot affect the bind path.
- Add a libsimulator regression test and a ChangeLog entry.

This leaves one DSCP setup site for each connection direction. The regression test disables outgoing uTP and specifically covers an ordinary outgoing TCP peer connection.

## Standalone POC

The relevant POC sequence is:

```cpp
boost::asio::io_context io;
boost::asio::ip::tcp::socket socket(io); // constructed, but not open
boost::system::error_code ec;

// Current RC_2_1 order
libtorrent_like_set_traffic_class(socket, 0x04, ec);
std::cout << ec.value() << " " << ec.message() << "\n";
socket.open(boost::asio::ip::tcp::v4(), ec);
socket.connect(server, ec);

// Correct order
boost::asio::ip::tcp::socket marked_socket(io);
marked_socket.open(boost::asio::ip::tcp::v4(), ec);
libtorrent_like_set_traffic_class(marked_socket, 0x04, ec);
marked_socket.connect(server, ec);
```

`libtorrent_like_set_traffic_class()` in the POC mirrors the relevant `local_endpoint()` and `set_option(IP_TOS)` flow in `aux::set_traffic_class()`.

Tested on Debian 13, Linux 6.12.90, Boost 1.83:

```text
[BAD: set DSCP -> open -> connect]
  initially is_open=0
  set_traffic_class before open: ec=9 ("Bad file descriptor")
  open: ec=0 ("Success")
  IP_TOS after open: raw ToS=0x0, decoded DSCP=0
  connect: ec=0 ("Success")

[GOOD: open -> set DSCP -> connect]
  initially is_open=0
  open: ec=0 ("Success")
  set_traffic_class after open: ec=0 ("Success")
  IP_TOS before connect: raw ToS=0x4, decoded DSCP=1
  connect: ec=0 ("Success")
```

The corresponding SYN packets captured with tcpdump were:

```text
IP (tos 0x0, ...) 127.0.0.1.35872 > 127.0.0.1.45432: Flags [S]
IP (tos 0x4, ...) 127.0.0.1.35882 > 127.0.0.1.45432: Flags [S]
```

## Regression test

Before the production change, the new test fails with:

```text
SET_DSCP [ value: 4 e: Bad file descriptor ]
OPEN [ protocol: IPv4 ]
ASYNC_CONNECT
TEST_ERROR: dscp_error: 1 expected: 0
```

With this change:

```text
$ cd simulation
$ b2 test_outgoing_interfaces -j2
**passed** .../test_outgoing_interfaces.test
```

Additional checks:

```text
git diff --check
git-clang-format-18 --binary clang-format-18 --diff HEAD -- \
  src/peer_connection.cpp simulation/test_outgoing_interfaces.cpp
clang-format did not modify any files
```

### DIFF
--- a/simulation/test_outgoing_interfaces.cpp
+++ b/simulation/test_outgoing_interfaces.cpp
@@ -29,8 +29,8 @@ namespace {
 	// src/enum_net.cpp).
 	constexpr char const* device_name = "eth0";
 
-	template <typename Settings, typename Test>
-	void run_test(Settings const& make_settings, Test const& test)
+	template <typename Settings, typename Alert, typename Test>
+	void run_test(Settings const& make_settings, Alert const& on_alert, Test const& test)
 	{
 		// setup the simulation
 		sim::default_config network_cfg;
@@ -56,12 +56,13 @@ namespace {
 		params.flags &= ~lt::torrent_flags::paused;
 		ses->async_add_torrent(params);
 
-		print_alerts(*ses, [](lt::session&, lt::alert const* a) {
+		print_alerts(*ses, [&](lt::session&, lt::alert const* a) {
 			if (auto at = lt::alert_cast<lt::add_torrent_alert>(a))
 			{
 				lt::torrent_handle h = at->handle;
 				add_fake_peer(h, 0);
 			}
+			on_alert(a);
 		});
 
 		sim::timer t(sim, lt::seconds(60), [&](boost::system::error_code const&) {
@@ -74,6 +75,12 @@ namespace {
 		});
 
 		sim.run();
+	}
+
+	template <typename Settings, typename Test>
+	void run_test(Settings const& make_settings, Test const& test)
+	{
+		run_test(make_settings, [](lt::alert const*) {}, test);
 	}
 
 } // anonymous namespace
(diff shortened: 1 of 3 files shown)


simulation/test_outgoing_interfaces.cpp
```diff
@@ -123,3 +130,29 @@ TORRENT_TEST(outgoing_interfaces_matching_device_allowed_with_proxy)
...
+	run_test(
+		[](lt::settings_pack& pack) {
+			pack.set_int(lt::settings_pack::peer_dscp, 0x04);
+			pack.set_bool(lt::settings_pack::enable_outgoing_utp, false);
+		},
+		[&](lt::alert const* a) {
+#ifndef TORRENT_DISABLE_LOGGING
+			auto const* pl = lt::alert_cast<lt::peer_log_alert>(a);
+			if (pl != nullptr && pl->event_type == lt::peer_log_alert::set_dscp)
+				dscp_error = true;
```
arvidn, line 157: this test will pass unconditionally if built with `TORRENT_DISABLE_LOGGING`, as it depends on receiving these log alerts. Ideally the test would look for some other signal, but I can't think of what it could be.

I think it would make sense to make this whole test be conditional on `TORRENT_DISABLE_LOGGING`, and not run it instead of reporting success unconditionally.

simulation/test_outgoing_interfaces.cpp
```diff
@@ -123,3 +130,29 @@ TORRENT_TEST(outgoing_interfaces_matching_device_allowed_with_proxy)
...
+	bool dscp_error = false;
+	run_test(
+		[](lt::settings_pack& pack) {
+			pack.set_int(lt::settings_pack::peer_dscp, 0x04);
+			pack.set_bool(lt::settings_pack::enable_outgoing_utp, false);
+		},
+		[&](lt::alert const* a) {
+#ifndef TORRENT_DISABLE_LOGGING
+			auto const* pl = lt::alert_cast<lt::peer_log_alert>(a);
+			if (pl != nullptr && pl->event_type == lt::peer_log_alert::set_dscp)
```
arvidn: I think, to make this test meaningful you'd need to also look for `peer_log_alert::open` and ensure it arrives before `peer_log_alert::set_dscp`

simulation/test_outgoing_interfaces.cpp
jiaming-shi: Thanks for catching this. I wrapped the whole test in the TORRENT_DISABLE_LOGGING guard, so it is not registered in logging-disabled builds.

simulation/test_outgoing_interfaces.cpp
jiaming-shi: Thanks, good point. The test now records the open alert, requires it to arrive before any set_dscp error alert, and verifies that open was observed.

arvidn (dismissed):

arvidn:
```
src/peer_connection.cpp:388:15: error: declaration shadows a local variable [-Werror,-Wshadow]
                        error_code ec;
                                   ^
src/peer_connection.cpp:377:14: note: previous declaration is here
                error_code ec;
                           ^
```

jiaming-shi:
Thanks for catching this. I’ve changed the local error code variable to `err` to avoid shadowing the outer `ec`. The compilation issue is fixed in the latest revision.

arvidn (approved):

PR #8660, opened 2026-07-26
use libcrypto's bignum in DH key exchange

arvidn:
inspired by https://github.com/arvidn/libtorrent/pull/8637

| benchmark | before | after | speed-up |
| --- | --- | --- | --- |
| dh_key_exchange | 133636 | 30991 | 4.31 |
| dh_compute_secret | 141014 | 34868 | 4.04 |
| dh_handshake | 274607 | 67748 | 4.05 |

### DIFF
--- a/fuzzers/src/pe_conn.cpp
+++ b/fuzzers/src/pe_conn.cpp
@@ -68,9 +68,8 @@ peer_fuzz_session g_fz;
 
 // Build RC4 keys for an outgoing (client-initiating) PE connection.
 // The client's send key is hash("keyA", S, SKEY); receive key is hash("keyB", S, SKEY).
-static std::shared_ptr<rc4_handler> make_rc4(lt::aux::key_t const& secret, sha1_hash const& skey)
+static std::shared_ptr<rc4_handler> make_rc4(span<char const> secret_buf, sha1_hash const& skey)
 {
-	std::array<char, 96> const secret_buf = export_key(secret);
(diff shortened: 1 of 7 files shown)


Tansien:
### Minor possible improvement

I tried one small follow-up on `fe68561c1`: cache a per-thread `BN_MONT_CTX` beside the existing `BN_CTX`, then use `BN_mod_exp_mont()` at both exponentiation sites. Since the DH modulus is fixed, this avoids rebuilding Montgomery state for every operation. OpenSSL documents that providing a preinitialized Montgomery context saves setup time: https://docs.openssl.org/3.4/man3/BN_mod_exp_mont/

The patch is 23 additions and 4 deletions. Five alternating Release A/B runs with OpenSSL 3.6.3 improved median `dh_handshake` from 32.89 us to 31.24 us, about 5%. `dh_compute_secret` improved about 7.8%.

The existing OpenSSL tests pass and the `BOOST_NO_CXX11_THREAD_LOCAL` fallback compiles. Would you be open to folding this into #8660?

arvidn:
Yes, I'm open to that. I think it comes down to a judgement call of performance improvement vs. added complexity (i.e. risk of additional issues).
@Tansien Do you have a PR against this branch?

Tansien:
Here it is: #8680

It targets `dh-key-exchange` and contains only the Montgomery context cache. I also reran the peer-crypto tests with OpenSSL, built-in crypto, and the no-thread-local fallback.

PR #8664, opened 2026-07-27
Fix Fast Extension piece validation after metadata

M-Hassan-Raza:
Preserve valid allowed-fast and suggested piece indices when metadata arrives for a sparse peer.

The metadata transition now validates these indices against the torrent piece range instead of the peer's piece count. The regression tests cover allowed-fast requests while choked and suggested-piece ordering over the wire.

Validated with test_fast_extension, predictive-pieces disabled, and the related metadata, Fast Extension, and peer-connection simulations.

### DIFF
--- a/src/peer_connection.cpp
+++ b/src/peer_connection.cpp
@@ -716,7 +716,7 @@ namespace {
 		m_have_piece.resize(t->torrent_file().num_pieces(), m_have_all);
 		m_num_pieces = m_have_piece.count();
 
-		piece_index_t const limit(m_num_pieces);
+		piece_index_t const limit = t->torrent_file().end_piece();
 
 		// now that we know how many pieces there are
 		// remove any invalid allowed_fast and suggest pieces
--- a/test/test_fast_extension.cpp
+++ b/test/test_fast_extension.cpp
@@ -193,6 +193,16 @@ void send_unchoke(tcp::socket& s)
 	if (ec) TEST_ERROR(ec.message());
 }
 
+void send_choke(tcp::socket& s)
+{
+	log("==> choke");
+	char msg[] = "\0\0\0\x01\0";
+	error_code ec;
+	boost::asio::write(s, boost::asio::buffer(msg, 5)
+		, boost::asio::transfer_all(), ec);
+	if (ec) TEST_ERROR(ec.message());
+}
+
 #ifndef TORRENT_DISABLE_PREDICTIVE_PIECES
 void send_interested(tcp::socket& s)
 {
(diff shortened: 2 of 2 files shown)


src/peer_connection.cpp
```diff
@@ -716,7 +716,7 @@ namespace {
 		m_have_piece.resize(t->torrent_file().num_pieces(), m_have_all);
 		m_num_pieces = m_have_piece.count();
 
-		piece_index_t const limit(m_num_pieces);
+		piece_index_t const limit = t->torrent_file().end_piece();
```
arvidn, line 719: as far as I can tell, this is the only non-test change. It suggests that `m_num_pieces` != `t->torrent_file().end_piece()`.

but `m_num_pieces = m_have_piece.size()`, which is determined by `m_have_piece.resize(t->torrent_file().num_pieces())`.

So, only if `t->torrent_file().num_pieces()` != `t->torrent_file().end_piece()`, does this make a difference, right?

https://github.com/arvidn/libtorrent/blob/RC_2_1/include/libtorrent/torrent_info.hpp#L401-L405
```
piece_index_t end_piece() const
{
	TORRENT_ASSERT(m_files.num_pieces() > 0);
	return m_files.end_piece();
}
```

https://github.com/arvidn/libtorrent/blob/RC_2_1/include/libtorrent/torrent_info.hpp#L388
```
int num_pieces() const { return m_files.num_pieces(); }
```

So it comes down to `file_storage::num_pieces()` vs. `file_storage::end_piece()`.

https://github.com/arvidn/libtorrent/blob/RC_2_1/include/libtorrent/file_storage.hpp#L447-L451
```
int num_pieces() const { TORRENT_ASSERT(m_piece_length > 0); return m_num_pieces; }
```

```
piece_index_t end_piece() const
{ return piece_index_t(m_num_pieces); }
```

I don't see how this change makes a difference.

M-Hassan-Raza: @arvidn Afaik `m_num_pieces` is set to `m_have_piece.count()` here, not `.size()`, so it counts pieces the peer has. The tests have 13 pieces, but the peer advertises only piece 12 (count 1), or pieces 0 and 12 (count 2), so the old limit wrongly removse piece 12. `num_pieces()` and `end_piece()` are equivilant here. So the relevant difference is `m_num_pieces` vs. the torrent’s piece range.

Pls lmk if this makes sense

arvidn: heh, I totally read `count()` as `size()`. you're right

arvidn (approved):

src/peer_connection.cpp
M-Hassan-Raza: Honestly I almost always mix and match len(), len, size() etc all the time. When you brought up the point I took a real hard look to ensure I wasn't hallucinating 😁

PR #8675, opened 2026-07-29
Normalize IPv4-mapped addresses in SSRF checks

acts-1631:
IPv4-mapped loopback addresses were classified as IPv6 and could bypass the tracker and web-seed SSRF restrictions. Web seeds also checked an IPv4-only accounting field instead of the selected endpoint, which skipped local-address checks for IPv6 destinations.

This normalizes IPv4-mapped addresses before scope classification, uses the normalized loopback check for trackers, and checks the actual web-seed endpoint.

### DIFF
--- a/src/ip_helpers.cpp
+++ b/src/ip_helpers.cpp
@@ -17,6 +17,13 @@ see LICENSE file.
 namespace libtorrent {
 namespace aux {
 
+	address normalize_address(address const& a)
+	{
+		if (a.is_v6() && a.to_v6().is_v4_mapped())
+			return make_address_v4(v4_mapped, a.to_v6());
+		return a;
+	}
+
 	bool is_ip_address(std::string const& host)
 	{
 		error_code ec;
@@ -24,8 +31,9 @@ namespace aux {
 		return !ec;
 	}
 
-	bool is_global(address const& a)
+	bool is_global(address const& addr)
 	{
+		address const a = normalize_address(addr);
 		if (a.is_v6())
 		{
 			// https://www.iana.org/assignments/ipv6-address-space/ipv6-address-space.xhtml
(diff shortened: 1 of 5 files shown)


src/ip_helpers.cpp
```diff
@@ -17,15 +17,26 @@ see LICENSE file.
...
 namespace aux {
 
+	namespace {
+
+		address normalize_address(address const& a)
+		{
+			if (a.is_v6() && a.to_v6().is_v4_mapped())
+				return make_address_v4(v4_mapped, a.to_v6());
+			return a;
+		}
```
arvidn: can you add unit tests to demonstrate it working correctly?

I think it should have external linkage too

src/ip_helpers.cpp
```diff
@@ -39,8 +50,11 @@ namespace aux {
 		}
 	}
 
-	bool is_link_local(address const& a)
+	bool is_loopback(address const& addr) { return normalize_address(addr).is_loopback(); }
```
arvidn, line 50: I think it would be good to have a unit test demonstrating this working correctly as well

src/ip_helpers.cpp
```diff
@@ -52,10 +66,11 @@ namespace aux {
 		return (ip & 0xffff0000) == 0xa9fe0000; // 169.254.x.x
 	}
 
-	bool is_local(address const& a)
+	bool is_local(address const& addr)
```
arvidn, line 66: the unit test for this function should be extended to include mapped addresses

arvidn (commented):
looks good in general, I think it should come with unit tests though

acts-1631:
Updated. `normalize_address()` now has external linkage, and `test_enum_net` covers mapped-address normalization, mapped loopback checks, and mapped private/public addresses in `is_local()`. Is that okay?

arvidn (approved):

PR #8676, opened 2026-07-29
Bound incoming TLS handshakes

acts-1631:
Incoming TLS peer sockets were retained outside the global connection count while their handshakes were pending, and the handshake operation had no timeout. A remote client could open connections without sending a ClientHello and keep listener resources occupied indefinitely.

This counts pending TLS handshakes against the incoming connection allowance and gives each TCP and uTP handshake a timer based on `handshake_timeout`. Completion cancels the timer, expiry closes the socket, and session shutdown cancels all pending timers.

### DIFF
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -436,8 +436,10 @@ namespace aux {
 
 			void incoming_connection(socket_type);
 			std::int64_t num_connections_with_pending() const;
+			std::int64_t connection_limit(
+				tcp::endpoint const& endp, socket_type_t type);
 			void reject_incoming_connection(
-				tcp::endpoint const& endp, socket_type_t socket_type, std::int64_t limit);
+				tcp::endpoint const& endp, socket_type_t type, std::int64_t limit);
 
 			std::weak_ptr<torrent> find_torrent(info_hash_t const&) const override;
 #if TORRENT_ABI_VERSION == 1
@@ -1086,7 +1088,7 @@ namespace aux {
 #endif
 #ifdef TORRENT_SSL_PEERS
 			void on_incoming_utp_ssl(socket_type s);
-			bool can_accept_peer() const;
+			bool can_accept_peer(tcp::endpoint const& endp, socket_type_t type);
 			void arm_ssl_handshake_timer(
 				socket_type* s, std::shared_ptr<deadline_timer> const& timer);
 			void ssl_handshake(error_code const& ec, socket_type* s);
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -77,12 +77,12 @@ see LICENSE file.
 #include <algorithm>
 #include <vector>
 #include <set>
+#include <map>
 #include <list>
 #include <deque>
 #include <condition_variable>
 #include <mutex>
 #include <cstdarg> // for va_start, va_end
-#include <unordered_map>
 
 namespace libtorrent {
 
@@ -435,6 +435,9 @@ namespace aux {
 				, std::weak_ptr<tcp::acceptor>, transport);
 
 			void incoming_connection(socket_type);
+			std::int64_t num_connections_with_pending() const;
+			void reject_incoming_connection(
+				tcp::endpoint const& endp, socket_type_t socket_type, std::int64_t limit);
 
 			std::weak_ptr<torrent> find_torrent(info_hash_t const&) const override;
 #if TORRENT_ABI_VERSION == 1
@@ -1016,11 +1019,12 @@ namespace aux {
 			connection_map m_connections;
 
 #ifdef TORRENT_SSL_PEERS
-			// this list holds incoming connections while they
+			// this map holds incoming connections while they
 			// are performing SSL handshake. When we shut down
 			// the session, all of these are disconnected, otherwise
 			// they would linger and stall or hang session shutdown
-			std::set<std::unique_ptr<socket_type>, unique_ptr_less> m_incoming_sockets;
+			std::map<std::unique_ptr<socket_type>, std::shared_ptr<deadline_timer>, unique_ptr_less>
+				m_incoming_sockets;
 #endif
 
 			// maps IP ranges to bitfields representing peer class IDs
@@ -1082,7 +1086,11 @@ namespace aux {
 #endif
 #ifdef TORRENT_SSL_PEERS
 			void on_incoming_utp_ssl(socket_type s);
+			bool can_accept_peer() const;
+			void arm_ssl_handshake_timer(
+				socket_type* s, std::shared_ptr<deadline_timer> const& timer);
 			void ssl_handshake(error_code const& ec, socket_type* s);
+			void ssl_handshake_timeout(error_code const& ec, socket_type* s);
 #endif
 
 			// round-robin index into m_outgoing_interfaces
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -77,12 +77,12 @@ see LICENSE file.
 #include <algorithm>
 #include <vector>
 #include <set>
+#include <map>
 #include <list>
 #include <deque>
 #include <condition_variable>
 #include <mutex>
 #include <cstdarg> // for va_start, va_end
-#include <unordered_map>
 
 namespace libtorrent {
 
@@ -435,6 +435,9 @@ namespace aux {
 				, std::weak_ptr<tcp::acceptor>, transport);
 
 			void incoming_connection(socket_type);
+			std::int64_t num_connections_with_pending() const;
+			void reject_incoming_connection(
+				tcp::endpoint const& endp, socket_type_t socket_type, std::int64_t limit);
 
 			std::weak_ptr<torrent> find_torrent(info_hash_t const&) const override;
 #if TORRENT_ABI_VERSION == 1
@@ -1012,11 +1015,12 @@ namespace aux {
 			connection_map m_connections;
 
 #ifdef TORRENT_SSL_PEERS
-			// this list holds incoming connections while they
+			// this map holds incoming connections while they
 			// are performing SSL handshake. When we shut down
 			// the session, all of these are disconnected, otherwise
 			// they would linger and stall or hang session shutdown
-			std::set<std::unique_ptr<socket_type>, unique_ptr_less> m_incoming_sockets;
+			std::map<std::unique_ptr<socket_type>, std::shared_ptr<deadline_timer>, unique_ptr_less>
+				m_incoming_sockets;
 #endif
 
 			// maps IP ranges to bitfields representing peer class IDs
@@ -1078,7 +1082,11 @@ namespace aux {
 #endif
 #ifdef TORRENT_SSL_PEERS
 			void on_incoming_utp_ssl(socket_type s);
+			bool can_accept_peer() const;
+			void arm_ssl_handshake_timer(
+				socket_type* s, std::shared_ptr<deadline_timer> const& timer);
 			void ssl_handshake(error_code const& ec, socket_type* s);
+			void ssl_handshake_timeout(error_code const& ec, socket_type* s);
 #endif
 
 			// round-robin index into m_outgoing_interfaces
--- a/include/libtorrent/aux_/session_impl.hpp
+++ b/include/libtorrent/aux_/session_impl.hpp
@@ -436,8 +436,7 @@ namespace aux {
 
 			void incoming_connection(socket_type);
 			std::int64_t num_connections_with_pending() const;
-			std::int64_t connection_limit(
-				tcp::endpoint const& endp, socket_type_t type);
+			std::int64_t connection_limit(tcp::endpoint const& endp, socket_type_t type);
 			void reject_incoming_connection(
 				tcp::endpoint const& endp, socket_type_t type, std::int64_t limit);
 
--- a/src/session_impl.cpp
+++ b/src/session_impl.cpp
@@ -1075,7 +1075,8 @@ bool ssl_server_name_callback(ssl::stream_handle_type stream_handle, std::string
 			auto const sockets = std::move(m_incoming_sockets);
 			for (auto const& s : sockets)
 			{
-				s->close(ec);
+				s.second->cancel();
+				s.first->close(ec);
 				TORRENT_ASSERT(!ec);
 			}
 		}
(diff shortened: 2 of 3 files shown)


include/libtorrent/aux_/session_impl.hpp
```diff
@@ -1017,6 +1017,8 @@ namespace aux {
 			// the session, all of these are disconnected, otherwise
 			// they would linger and stall or hang session shutdown
 			std::set<std::unique_ptr<socket_type>, unique_ptr_less> m_incoming_sockets;
```
arvidn: It would seem more straight-forward to bundle the timer with each item here. Once the handshake completes, the socket is removed from this set, right? at that point we won't need the timer anymore either.

So, maybe something like:

```
std::unordered_map<std::unique_ptr<socket_type>, std::shared_ptr<deadline_timer>> m_incoming_sockets;
```

The  `unique_ptr` would need to support hash though.

src/session_impl.cpp
```diff
@@ -2848,6 +2852,11 @@ namespace {
 		if (listen != m_listen_sockets.end())
 			(*listen)->incoming_connection = true;
 
+#ifdef TORRENT_SSL_PEERS
+		if (ssl == transport::ssl && !has_ssl_handshake_capacity())
+			return;
```
arvidn: I think this case warrants a session log too

acts-1631:
Each incoming socket and timer now share one `std::map` entry. The existing transparent `unique_ptr_less` comparator supports lookup by raw socket pointer, avoiding a custom hash. Timed-out sockets remain owned until the handshake handler runs. I also added session logging when TCP or uTP SSL handshakes are rejected at the connection limit.

src/session_impl.cpp
```diff
@@ -2917,6 +2958,26 @@ namespace {
...
+	}
+
+	void session_impl::arm_ssl_handshake_timer(
+		socket_type* const sock, std::shared_ptr<deadline_timer> const& timer)
+	{
+		timer->expires_after(seconds(m_settings.get_int(settings_pack::handshake_timeout)));
+
+		ADD_OUTSTANDING_ASYNC("session_impl::ssl_handshake_timeout");
+		timer->async_wait([this, sock, timer](error_code const& ec) {
+			TORRENT_UNUSED(timer);
```
arvidn, line 3010: if `timer` is unused, does it need to be captured by the lambda?

src/session_impl.cpp
```diff
@@ -2898,17 +2920,36 @@ namespace {
 	{
 		TORRENT_ASSERT(is_ssl(s));
 
+		if (!has_ssl_handshake_capacity())
+		{
```
arvidn, line 2966: in the other case where a peer is rejected, we post a `peer_disconnected_alert` (line 3208). I think it might make sense to do that here too.

In fact, both this new rejection and the new place as well, where you call `has_ssl_handshake_capacity()` share some code with the existing one.

There are some noteworthy differences though. Like using `has_ssl_handshake_capacity()` rather than computing `num_connections_with_pending`.

This also means the `session_log` logs `num_connections()`, instead the effective limit.

src/session_impl.cpp
```diff
@@ -3131,9 +3215,11 @@ namespace {
 #ifndef TORRENT_DISABLE_LOGGING
 			if (should_log())
 			{
-				session_log("<== INCOMING CONNECTION [ connections limit exceeded, conns: %d, limit: %d, slack: %d ]"
-					, num_connections(), m_settings.get_int(settings_pack::connections_limit)
-					, m_settings.get_int(settings_pack::connections_slack));
+				session_log("<== INCOMING CONNECTION [ connections limit exceeded, conns: %d, "
+							"limit: %d, slack: %d ]",
+					int(num_connections_with_pending),
```
arvidn: in the two new rejections, you print num connections separately from the number of pending connections. I think it would be good to be consistent.

src/session_impl.cpp
```diff
@@ -2917,6 +2958,26 @@ namespace {
 	//   -CAfile <torrent-cert>.pem  -debug -connect 127.0.0.1:4433 -tls1
 	//   -servername <hex-encoded-info-hash>
 
+	bool session_impl::has_ssl_handshake_capacity() const
```
arvidn: this function name is a bit misleading. It means something like `can_accept_peer()`.

acts-1631:
Kept the timer capture intentionally: after `ssl_handshake()` cancels the timer and erases its map entry, the capture keeps it alive until the queued cancellation handler completes.

Renamed the capacity check to `can_accept_peer()` and factored the pending-handshake-inclusive connection count and rejection reporting. Early TCP and uTP TLS rejections now use the same alert and log path as post-handshake rejections, with the total connection count and the limit used for the decision.

arvidn:
```
src/session_impl.cpp:2859:15: error: declaration shadows a local variable [-Werror,-Wshadow]
 2859 |                         error_code ec;
      |                                    ^
src/session_impl.cpp:2774:14: note: previous declaration is here
 2774 |                 error_code ec;
      |                            ^
1 error generated.
```

arvidn:
```
src/session_impl.cpp:3046:96: error: unused parameter ‘limit’ [-Werror=unused-parameter]
 3046 |                 tcp::endpoint const& endp, socket_type_t const socket_type, std::int64_t const limit)
      |                                                                             ~~~~~~~~~~~~~~~~~~~^~~~~
cc1plus: all warnings being treated as errors
```

src/session_impl.cpp
```diff
@@ -2921,6 +2948,28 @@ namespace {
 	//   -CAfile <torrent-cert>.pem  -debug -connect 127.0.0.1:4433 -tls1
 	//   -servername <hex-encoded-info-hash>
 
+	bool session_impl::can_accept_peer() const
```
arvidn: this function doesn't take the peer_class' connection limit into account. It's probably not possible to apply per-torrent rules at this point, but It would expect you can look up the peer class based on IP address at least.

Specifically `connection_limit_factor` could be used here.

acts-1631:
Anything else you want done on this one or is it ok to go?

src/session_impl.cpp
```diff
@@ -2992,8 +3010,60 @@ namespace {
 		incoming_connection(std::move(s));
 	}
 
+	void session_impl::ssl_handshake_timeout(error_code const& ec, socket_type* sock)
+	{
+		COMPLETE_ASYNC("session_impl::ssl_handshake_timeout");
+
+		if (ec)
+			return;
```
arvidn, line 3066: I'm wondering if it would make sense to issue a session_log that the connection failed here, if `ec` != `operation_aborted` of course.

When we get here, have we already posted the alert for the incoming connection?

src/session_impl.cpp
```diff
@@ -2992,8 +3010,60 @@ namespace {
...
+	{
+		std::int64_t ret = num_connections();
+#ifdef TORRENT_SSL_PEERS
+		ret += std::int64_t(m_incoming_sockets.size());
+#endif
+		return ret;
+	}
+
+	void session_impl::reject_incoming_connection(
+		tcp::endpoint const& endp, socket_type_t const socket_type, std::int64_t const limit)
```
arvidn: the `socket_type` variable name shadows the `socket_type` *type* name here

arvidn (commented):
sorry for the delay. just a few comments. I think it would be really great to have some test cases in simulations/test_ssl.cpp 

At least the main DoS case

acts-1631:
Okay, have a look at this one please.

src/session_impl.cpp
```diff
@@ -2884,6 +2849,22 @@ namespace {
...
+			error_code endpoint_ec;
+			tcp::endpoint const endp = s.remote_endpoint(endpoint_ec);
+			if (endpoint_ec) return;
+
+			bool const reject = !can_accept_peer(endp, socket_type_t::tcp_ssl);
+			if (reject)
+				reject_incoming_connection(endp,
+					socket_type_t::tcp_ssl,
+					connection_limit(endp, socket_type_t::tcp_ssl));
+			if (reject) return;
```
```suggestion
			if (reject)
			{
				reject_incoming_connection(endp,
					socket_type_t::tcp_ssl,
					connection_limit(endp, socket_type_t::tcp_ssl));
				return;
			}
```

arvidn (commented):
tou need to apply the `clang-format` diff, printed here: https://github.com/arvidn/libtorrent/actions/runs/31333885163/job/93296390497?pr=8676 (or install `pre-commit`).

cppcheck is saying this:
```
simulation/test_ssl.cpp:401:3: warning: Member variable 'malicious_peer::m_read_buf' is not initialized in the constructor. [uninitMemberVar]
  malicious_peer(sim::simulation& sim,
  ^
```

acts-1631:
Should be addressed now

arvidn:
@acts-1631 you agent needs to monitor the CI status

arvidn:
I was thinking it would be good to have a test for this as well, something like this:

```diff
diff --git a/simulation/test_ssl.cpp b/simulation/test_ssl.cpp
index ff20479c9..47a052edf 100644
--- a/simulation/test_ssl.cpp
+++ b/simulation/test_ssl.cpp
@@ -630,6 +630,126 @@ namespace {
                sim.run();
        }
 
+       // pending SSL handshakes must be bounded by connections_limit (+
+       // connections_slack), the same way incoming_connection() already bounds
+       // plain TCP accepts synchronously, at accept time. on_accept_connection()
+       // stashes an accepted SSL socket in m_incoming_sockets and only runs the
+       // connections_limit check, in incoming_connection(), after its TLS
+       // handshake completes, so a peer that never starts the handshake
+       // occupies a slot forever, see the "TODO: this size need to be capped"
+       // comments on m_incoming_sockets in src/session_impl.cpp.
+       //
+       // this test currently fails: every attacker below connects and stays
+       // connected, none get rejected, despite far exceeding connections_limit.
+       // it should start passing, unmodified, once pending SSL handshakes are
+       // capped and timed out.
+       void run_ssl_pending_handshake_limit()
+       {
+               address const peer0 = addr("50.0.0.1");
+               int const port = 6881;
+               int const connections_limit = 5;
+               int const connections_slack = 0;
+               int const allowed = connections_limit + connections_slack;
+               int const num_attackers = 4 * connections_limit;
+
+               sim::default_config network_cfg;
+               sim::simulation sim{network_cfg};
+               sim::asio::io_context ios0{sim, peer0};
+
+               lt::session_proxy zombie;
+
+               lt::session_params params;
+               lt::settings_pack& pack = params.settings;
+               pack = settings();
+               pack.set_int(settings_pack::connections_limit, connections_limit);
+               pack.set_int(settings_pack::connections_slack, connections_slack);
+
+               // this session only ever listens on its SSL port, like
+               // run_malicious_peer_attack() above
+               char iface[50];
+               std::snprintf(iface, sizeof(iface), "50.0.0.1:%ds", port);
+               pack.set_str(settings_pack::listen_interfaces, iface);
+
+               params.disk_io_constructor = test_disk().set_seed(true);
+               auto ses = std::make_shared<lt::session>(params, ios0);
+
+               std::string const root_cert = read_file("root_ca_cert.pem");
+               lt::add_torrent_params const atp = make_ssl_test_torrent(root_cert);
+
+               // attacker sockets: connect to the SSL port, then never send the
+               // TLS ClientHello ssl_stream::async_accept_handshake() is waiting
+               // for. reserved up front so these vectors never reallocate and
+               // invalidate async_connect's captured reference below
+               std::size_t const num_attackers_sz = std::size_t(num_attackers);
+               std::vector<std::unique_ptr<sim::asio::io_context>> attacker_ios;
+               std::vector<lt::tcp::socket> attacker_socks;
+               attacker_ios.reserve(num_attackers_sz);
+               attacker_socks.reserve(num_attackers_sz);
+               int connected = 0;
+
+               print_alerts(*ses, [&](lt::session&, lt::alert const* a) {
+                       if (auto* ta = alert_cast<add_torrent_alert>(a))
+                       {
+                               torrent_handle h = ta->handle;
+                               h.set_ssl_certificate(ssl_fixture_path("peer_certificate.pem"),
+                                       ssl_fixture_path("peer_private_key.pem"),
+                                       ssl_fixture_path("dhparams.pem"),
+                                       "test");
+                       }
+                       else if (alert_cast<torrent_finished_alert>(a))
+                       {
+                               // note: i starts at 1, since make_io_context(sim, 0) would
+                               // collide with the seed's own address, 50.0.0.1
+                               for (int i = 1; i <= num_attackers; ++i)
+                               {
+                                       attacker_ios.push_back(make_io_context(sim, i));
+                                       attacker_socks.emplace_back(*attacker_ios.back());
+                                       attacker_socks.back().async_connect(
+                                               lt::tcp::endpoint(peer0, std::uint16_t(port)),
+                                               [&](error_code const& ec) { TEST_CHECK(!ec); if (!ec) ++connected; });
+                               }
+                       }
+               });
+
+               ses->async_add_torrent(atp);
+
+               sim::timer give_up(sim, lt::seconds(30), [&](boost::system::error_code const&) {
+                       // a non-blocking read directly tells us whether the seed closed
+                       // each connection, without leaving an async operation pending
+                       // to cancel at teardown
+                       int closed = 0;
+                       for (lt::tcp::socket& sock : attacker_socks)
+                       {
+                               error_code ec;
+                               sock.non_blocking(true, ec);
+                               std::array<char, 1> buf{};
+                               sock.read_some(boost::asio::buffer(buf), ec);
+                               if (ec && ec != boost::asio::error::would_block)
+                                       ++closed;
+                       }
+
+                       std::printf("attackers: %d allowed: %d connected: %d closed: %d\n"
+                               , num_attackers, allowed, connected, closed);
+
+                       TEST_CHECK(num_attackers > allowed);
+
+                       // the TCP accept itself always succeeds, rejection happens
+                       // afterwards by closing the socket, so every attacker connects
+                       // regardless of the limit
+                       TEST_EQUAL(connected, num_attackers);
+
+                       // only `allowed` pending SSL handshakes may occupy a slot at
+                       // once; the rest must be closed, just like plain TCP accepts
+                       // beyond the limit already are in incoming_connection()
+                       TEST_EQUAL(closed, num_attackers - allowed);
+
+                       zombie = ses->abort();
+                       ses.reset();
+               });
+
+               sim.run();
+       }
+
 } // anonymous namespace
 
 TORRENT_TEST(ssl_transfer_matrix)
@@ -661,6 +781,11 @@ TORRENT_TEST(ssl_malicious_peer)
        }
 }
 
+TORRENT_TEST(ssl_pending_handshake_limit)
+{
+       run_ssl_pending_handshake_limit();
+}
+
 #else
 
 TORRENT_TEST(disabled) {}
```

arvidn:
also, a heads-up. this fix will bubble up from RC_2_0 -> RC_2_1 -> master

https://github.com/arvidn/libtorrent/pull/8835

PR #8677, opened 2026-07-29
Bound WebSocket RTC offers

acts-1631:
An untrusted WebSocket tracker could send unlimited unique WebRTC offers. Each offer created a PeerConnection and timer retained for up to `webtorrent_connection_timeout`, while offer IDs, SDP, and message sizes were not bounded.

This validates 20-byte offer IDs, limits SDP to 64 KiB and tracker messages to 128 KiB, grants only `numwant` inbound-offer tokens per announce, and caps concurrent inbound negotiations at 10 per torrent. Invalid remote SDP now removes the allocated negotiation immediately.


src/websocket_tracker_connection.cpp
```diff
@@ -446,6 +466,16 @@ parse_websocket_tracker_response(span<char const> message, error_code& ec) try
 		auto sdp = payload_answer["sdp"].as_string();
 		auto id = utf8_latin1(payload["offer_id"].as_string());
 		auto pid = utf8_latin1(payload["peer_id"].as_string());
+		if (id.size() != RTC_OFFER_ID_LEN)
```
arvidn: I understand that you want to cap this to some reasonable size, but are they really specified to be exactly 20 bytes? This is JSON, right? So if it's HEX, it can't fit a full info-hash. I would have expected something like

```
if (id.size() > RTC_OFFER_ID_LEN)
```

and making `RTC_OFFER_IF_LEN = 40`

src/rtc_signaling.cpp
```diff
@@ -215,22 +223,37 @@ void rtc_signaling::process_offer(rtc_offer const& offer)
...
 
 void rtc_signaling::process_answer(rtc_answer const& answer)
 {
+	if (answer.offer_id.size() != RTC_OFFER_ID_LEN || answer.sdp.size() > RTC_MAX_SDP_SIZE)
+		return;
+
 	auto it = m_connections.find(answer.offer_id);
-	if (it == m_connections.end()) return;
+	if (it == m_connections.end())
+		return;
```
arvidn: these cases could probably use a `debug_log()` too

src/tracker_manager.cpp
```diff
@@ -450,9 +450,7 @@ namespace libtorrent::aux {
 			std::shared_ptr<request_callback> cb = c.lock();
 			if (!cb) return;
 
-			// TODO: introduce a setting for max_offers
```
arvidn, line 331: I think an `int` setting in `settings_pack` would make sense for this

acts-1631:
Added `max_webtorrent_offers` as an integer setting, defaulting to 10, and use it for both the advertised offer count and concurrent incoming negotiations. I also added debug logs for rejected offers and answers.

I kept `RTC_OFFER_ID_LEN` at 20 for generated binary IDs and added `RTC_MAX_OFFER_ID_SIZE` at 40 for accepted input. The reference `bittorrent-tracker` implementation generates 20 random bytes, represents them as 40 hexadecimal characters internally, then sends `hex2bin(offerId)`. Changing the existing constant to 40 would make libtorrent generate 40 binary bytes instead. This accepts IDs up to 40 bytes without changing outbound IDs. The formatting failure is also fixed.

include/libtorrent/aux_/rtc_signaling.hpp
```diff
@@ -48,6 +49,9 @@ namespace aux {
 struct rtc_stream_init;
 
 constexpr int RTC_OFFER_ID_LEN = 20;
+constexpr std::size_t RTC_MAX_OFFER_ID_SIZE = 40;
```
arvidn: After having done some research on this, my understanding is that there is no size constrain on the rtc offer id length, but people generally seem to expect it to be 20 bytes. There's no protocol specification.

https://github.com/webtorrent/bittorrent-tracker/issues/525

I think it would be reasonable to require it to be exactly 20 bytes. That let's us use the `RTC_OFFER_ID_LEN` constant, instead of introducing a new one.

This constant should probably be removed and use `RTC_OFFER_ID_LEN` instead.

Does that seem reasonable?

src/rtc_signaling.cpp
```diff
@@ -190,11 +191,33 @@ void rtc_signaling::generate_offers(int count, offers_handler handler)
 
 void rtc_signaling::process_offer(rtc_offer const& offer)
 {
-	if (m_connections.find(offer.id) != m_connections.end()) {
+	if (offer.id.size() > RTC_MAX_OFFER_ID_SIZE || offer.sdp.size() > RTC_MAX_SDP_SIZE)
```
arvidn: as an anti-DoS measure, I think it may be too late to check the length here. We've already allocated the memory for `offer.id` here. I think it should be checked where the `offer` object is created, or maybe in `rtc_offer_id` constructor.

arvidn (commented):
I think this change should be made in `RC_2_1`, and included in the next release. I have some comments on its actual shape though.

src/rtc_signaling.cpp
arvidn: same thing for `offer.sdp`

arvidn:
what do you think of rebasing this PR on top of https://github.com/arvidn/libtorrent/pull/8685 ?

acts-1631:
Thanks for the comments. I'll have a look after this weekend and get back with you on it.

arvidn:
I landed my PR, so it should be rebased on top of `RC_2_1`. My PR restricts the size of the offer id, so that part of this PR is no longer needed.

acts-1631:
Rebased onto `RC_2_1` and removed the offer-ID changes covered by #8685. The SDP limit is now checked on the parsed JSON string before copying it into an `rtc_offer`. The WebSocket message cap and the remaining quota, connection-limit, and cleanup changes are unchanged.

arvidn:
```diff
diff --git a/src/tracker_manager.cpp b/src/tracker_manager.cpp
index 2c73cc6..f79ac6a 100644
--- a/src/tracker_manager.cpp
+++ b/src/tracker_manager.cpp
@@ -328,8 +328,7 @@ namespace libtorrent::aux {
 			std::shared_ptr<request_callback> cb = c.lock();
 			if (!cb) return;
 
-			int const max_offers = std::max(
-				sett.get_int(settings_pack::max_webtorrent_offers), 0);
+			int const max_offers = std::max(sett.get_int(settings_pack::max_webtorrent_offers), 0);
 			req.num_want = std::min(req.num_want, max_offers);
 			if (req.num_want == 0)
 			{
diff --git a/src/websocket_tracker_connection.cpp b/src/websocket_tracker_connection.cpp
index 703f782..916ed42 100644
--- a/src/websocket_tracker_connection.cpp
+++ b/src/websocket_tracker_connection.cpp
@@ -442,10 +442,11 @@ void websocket_tracker_connection::on_read(error_code ec, std::size_t /* bytes_r
 				--quota->second;
 
 				response.offer->answer_callback = [info_hash = response.info_hash,
-					self = shared_from_this(), id = response.offer->id,
-					pid = response.offer->pid](peer_id const& local_pid,
-					aux::rtc_answer const& answer)
-				{
+													  self = shared_from_this(),
+													  id = response.offer->id,
+													  pid = response.offer->pid](
+													  peer_id const& local_pid,
+													  aux::rtc_answer const& answer) {
 					self->queue_answer(
 						{std::move(info_hash), std::move(local_pid), std::move(answer)});
 					self->start();
```

acts-1631:
It says `RC_2_1` on my end.

arvidn:
> It says `RC_2_1` on my end.

yeah, I needed to reload the page. somehow your other updates made it, just not the target branch

PR #8680, opened 2026-07-30
cache Montgomery context for DH key exchange

Tansien:
## Summary

- Cache one `BN_MONT_CTX` beside the existing per-thread `BN_CTX`.
- Use `BN_mod_exp_mont()` for both DH exponentiations.
- Preserve the existing shared-context fallback and lock when thread-local storage is unavailable.

The DH modulus is fixed, so reusing its Montgomery context avoids rebuilding modulus-specific state for each operation.

## Performance

Five alternating Release A/B runs with OpenSSL 3.6.3 improved median `dh_handshake` from 32.89 us to 31.24 us, about 5%. `dh_compute_secret` improved about 7.8%.

## Tests

- OpenSSL `test_pe_crypto`: 4 passed
- Built-in crypto `test_pe_crypto`: 4 passed
- OpenSSL with `BOOST_NO_CXX11_THREAD_LOCAL`: 4 passed
- `git diff --check`

### DIFF
--- a/include/libtorrent/aux_/pe_crypto.hpp
+++ b/include/libtorrent/aux_/pe_crypto.hpp
@@ -120,6 +120,7 @@ namespace libtorrent::aux {
 		// from multiple threads. Falls back to a single instance guarded by a
 		// mutex where thread_local isn't available.
 		static ::BN_CTX* ctx();
+		static ::BN_MONT_CTX* mont();
 #else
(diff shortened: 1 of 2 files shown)


arvidn:
looks good. could you fix the `clang-format` issue? either install `pre-commit` or apply the diff printed in https://github.com/arvidn/libtorrent/actions/runs/30588411160/job/91043461888?pr=8680

arvidn:
you can re-target this to `RC_2_1`

PR #8707, opened 2026-08-05
Disable strict use of UTF-8

stalkerok:
Rollback of this change https://github.com/arvidn/libtorrent/pull/6665


HanabishiRecca:
I don't think it gonna work without reverting #6669 as well.

stalkerok:
I only see the broken encoding in the tracker's responses. Perhaps that will be enough, but I'm not sure.

PR #8714, opened 2026-08-08
fix UPnP hostname SSRF bypass

acts-1631:
UPnP validates literal Location and control URL addresses but accepted
IPv4-mapped loopback addresses and DNS names resolving to forbidden
endpoints. A forged local SSDP response could cause a request to a
loopback or link-local service.

Normalize mapped addresses and filter forbidden resolved endpoints on all
UPnP HTTP connections, including description fetches, control requests,
and redirects.

Also, do you want all PRs against RC_2_1 or master?

### DIFF
--- a/src/upnp.cpp
+++ b/src/upnp.cpp
@@ -43,6 +43,7 @@ see LICENSE file.
 #include <cstdlib>
 #include <cstdio> // for snprintf
 #include <cstdarg>
+#include <algorithm>
 #include <functional>
 
 using namespace std::placeholders;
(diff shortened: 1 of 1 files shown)


src/upnp.cpp
```diff
@@ -158,17 +159,31 @@ namespace {
...
+	// reject Location/control-URL addresses a real IGD would never use, to block
 	// SSRF via forged SSDP replies or device descriptions. RFC1918 stays
-	// allowed since that is where real devices live. only IP literals are
-	// checked here, not DNS names.
+	// allowed since that is where real devices live.
+	bool invalid_upnp_address(address const& addr)
+	{
+		address const normalized = (addr.is_v6() && addr.to_v6().is_v4_mapped())
+			? address(make_address_v4(v4_mapped, addr.to_v6()))
+			: addr;
```
arvidn, line 169: if you rebase this on top of `master`, you can use `normalize_address()` here

arvidn (commented):
looks good, but I think it should be rebased and re-targeted on top of `master`.

acts-1631:
Rebased. Sorry, still don't quite understand your preference between the two branches.

arvidn:
I think this patch is a bit in the grey zone. It is technically a fix, but the other related fixes have gone into master now (with mapped IPv6 addresses).

arvidn:
fixed now

PR #8716, opened 2026-08-09
Fix short writes in copy file fallbacks

M-Hassan-Raza:
The copy file fallback loops advanced the buffer offset while also shrinking the byte count. A short write could therefore skip data or underflow the next write size.

Use the existing pwrite_all helper for both POSIX and Windows fallback paths and preserve file_write error reporting. Add a kernel-backed regression test covering partial positional writes and the resulting file size.

Tests: pre-commit, clang-format 18, CMake build and test suite, installed-header check, and B2 test_file/test_copy_file with WebTorrent enabled and disabled.

### DIFF
--- a/test/test_file.cpp
+++ b/test/test_file.cpp
@@ -78,11 +78,7 @@ TORRENT_TEST(pwrite_all_short_write)
 		return;
 	}
 
-	struct sigaction ignore_action
-	{};
-	ignore_action.sa_handler = SIG_IGN;
-	sigemptyset(&ignore_action.sa_mask);
-	if (::sigaction(SIGXFSZ, &ignore_action, nullptr) != 0)
+	if (::signal(SIGXFSZ, SIG_IGN) == SIG_ERR)
 	{
 		TEST_ERROR("could not ignore SIGXFSZ");
 		return;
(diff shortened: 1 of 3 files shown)


arvidn:
```
test_file.cpp:83:16: error: disabled expansion of recursive macro [-Werror,-Wdisabled-macro-expansion]
        ignore_action.sa_handler = SIG_IGN;
                      ^
/usr/include/x86_64-linux-gnu/bits/sigaction.h:39:41: note: expanded from macro 'sa_handler'
# define sa_handler     __sigaction_handler.sa_handler
```

test/test_file.cpp
```diff
@@ -20,13 +20,21 @@ see LICENSE file.
...
 #include "test_utils.hpp"
+#include <algorithm>
+#include <cerrno>
 #include <vector>
 #include <set>
 #include <thread>
 #include <iostream>
 
+#ifndef TORRENT_WINDOWS
+#include <signal.h>
```
arvidn:
```suggestion
#include <csignal>
```

arvidn (dismissed):
just one minor suggestion

arvidn (approved):

